PR #1465, opened 2024-04-16
Unify prep.vm, deletevm.rb, and recreate-unpersisted into a single executable

pykello:
Merges `prepvm.rb`, `deletevm.rb`, and `recreate-unpersisted` into one executable.

These are in preparation to add a 4th action to `setup-vm` which will reinstall systemd units after an SPDK used for a VM is updated.


fdr (commented):
-84!? :heart:

pykello:
This needs more thought about error cases of vm purge, so converted to draft to revise it.

rhizome/host/bin/setup-vm
```diff
@@ -0,0 +1,70 @@
...
+
+  vm_setup.prep(unix_user, ssh_public_key, nics, gua, ip4,
+    local_ip4, boot_image, max_vcpus, cpu_topology, mem_gib,
+    ndp_needed, storage_volumes, storage_secrets, swap_size_bytes)
+when "recreate-unpersisted"
+  secrets = JSON.parse($stdin.read)
+  unless (storage_secrets = secrets["storage"])
+    puts "need storage secrets in secrets json"
+    exit 1
+  end
```
pykello, line 63: Although `secrets` fetch could be factored out of the conditionals if we only have `prep` and `recreate-unpersisted`, but I'm doing this way since future verbs don't need it.

rhizome/host/bin/setup-vm
```diff
@@ -0,0 +1,70 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require "json"
+require_relative "../lib/vm_setup"
+
+unless (verb = ARGV.shift)
```
enescakir, line 7: The `verb` seems quite unusual to me. Additionally, the error message is `"expected action as argument"`. Should we rename the variable to `action`?

prog/vm/nexus.rb
```diff
@@ -395,43 +395,46 @@ def before_run
...
       secrets_json = JSON.generate({
         storage: storage_secrets
       })
 
       # Enable KVM access for VM user.
       host.sshable.cmd("sudo usermod -a -G kvm #{q_vm}")
 
-      # put prep.json
-      host.sshable.cmd("sudo -u #{q_vm} tee #{params_path.shellescape}", stdin: params_json)
+      write_params_json
```
enescakir, line 405: For learning purpose, why do we write to a file instead of previous way or passing along with the secrets via stdin? You might want to write a sentence about it to the commit message.

enescakir: Also, where do we delete this file?

fdr (approved):
I see how these commits map nicely to deploys and roll-back points, but it does seem like the commit messages don't line up to what's going on in each patch.

If you ran "git blame" on some of the affected lines, you'd be stuck doing additional research on the second patch for sure.  And, you'd be confused about the first patch, which doesn't unify so much as add the new entry point in its own phase.

But, separating it out as you have did make it quick to review (even though I wasn't quick to get to it :see_no_evil: ), thank you for that.

velioglu (approved):

PR #1467, opened 2024-04-17
Increase nap time from 0 to 1 in donate

bsatzger:
With a nap time of 0 in `donate`, a parent strand executes its children in a loop until they exit, without taking a nap. This supports children making progress to unblock the parent.

However, an issue with this is that children are not always able to immediately make progress and nap 0 becomes wasteful. In production, we see instances where child strands are executed in a tight loop, hogging resources and starving other strands.

This change increases nap time in donate to 1. That way, a parent needs to take a short nap after running its children and cannot immediately re-execute them.

### DIFF
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -175,7 +175,7 @@ end
 
   def donate
     strand.children.map(&:run)
-    nap 0
+    nap 1
   end
 
   def reap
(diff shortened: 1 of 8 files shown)


furkansahin (approved):
I recently realised this in my development environment, too. This makes sense. Thank you.

enescakir (approved):

fdr (approved):
> That way, a parent needs to take a short nap

great prose

PR #1468, opened 2024-04-17
Increase max page limit for fetching failed GitHub deliveries

enescakir:
GitHub's app delivery API lacks adequate filtering features and only permits a maximum of 100 items per page. I've introduced a 'max_page' safety measure to prevent endless pagination.

As the count of delivered webhook events has risen, it appears we need to raise the 'max_page' limit to retrieve all failed deliveries within the specified time range.

### DIFF
--- a/lib/github.rb
+++ b/lib/github.rb
@@ -34,7 +34,7 @@ module Github
     @@runner_labels ||= YAML.load_file("config/github_runner_labels.yml").to_h { [_1["name"], _1] }
   end
 
-  def self.failed_deliveries(since, max_page = 20)
+  def self.failed_deliveries(since, max_page = 50)
     client = Github.app_client
     all_deliveries = client.get("/app/hook/deliveries?per_page=100")
     page = 1


fdr (approved):
huh. Better than before of course, I wonder if there's a way the system can detect it requires further adjustments and let us know. But that's no reason to get in the way of adjustment.

enescakir:
> huh. Better than before of course, I wonder if there's a way the system can detect it requires further adjustments and let us know. But that's no reason to get in the way of adjustment.
> 
> Possibly flagging some condition if the last page in the pagination is too close to the present date.

We already log the last page. I have also a graph at Mezmo.
https://github.com/ubicloud/ubicloud/blob/846ba9d4d24845100c26ca64369769f24bdb8687/lib/github.rb#L50

We might create a low level page, or Mezmo has PagerDuty and email integration. We can say if you see this log more than N times in M minutes, trigger the integration.
https://github.com/ubicloud/ubicloud/blob/846ba9d4d24845100c26ca64369769f24bdb8687/lib/github.rb#L43

PR #1469, opened 2024-04-17
Do not fail if the runner does not have a strand in serializer

enescakir:
Usually, we don't expect that the runner model will lack a strand. However, due to eager loading sending multiple queries, there may be brief periods when the runner doesn't have a strand in the loaded list. This PR ensures that the serializer doesn't fail if the runner is without a strand.

With recent changes, this state isn't only the state of virtual machine. So I moved to the runner model as "display_state" similar to the other models.

I also rename variables to make the test more readable.

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -25,6 +25,15 @@ class GithubRunner < Sequel::Model
     "http://github.com/#{repository_name}/settings/actions/runners/#{runner_id}" if runner_id
   end
 
+  def display_state
+    return vm.display_state if vm
+    case strand&.label
+    when "wait_vm_destroy" then "deleted"
+    when "wait_concurrency_limit" then "reached_concurrency_limit"
+    else "not_created"
+    end
+  end
+
   def init_health_monitor_session
     {
       ssh_session: vm.sshable.start_fresh_session
--- a/serializers/web/github_runner.rb
+++ b/serializers/web/github_runner.rb
@@ -18,15 +18,7 @@ class Serializers::Web::GithubRunner < Serializers::Base
         workflow_name: runner.workflow_job["workflow_name"],
         head_branch: runner.workflow_job["head_branch"]
       } : nil,
-      vm_state: if runner.vm
-                  runner.vm.display_state
-                elsif runner.strand.label == "wait_vm_destroy"
-                  "deleted"
-                elsif runner.strand.label == "wait_concurrency_limit"
-                  "reached_concurrency_limit"
-                else
-                  "not_created"
-                end
+      state: runner.display_state
     }
   end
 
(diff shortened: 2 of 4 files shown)


serializers/web/github_runner.rb
```diff
@@ -20,9 +20,9 @@ def self.base(runner)
       } : nil,
       vm_state: if runner.vm
                   runner.vm.display_state
-                elsif runner.strand.label == "wait_vm_destroy"
+                elsif runner.strand&.label == "wait_vm_destroy"
```
fdr, line 23: It's a bit smelly to have multiple "safe navigation" like this. Should `runner` learn a `display_state`, or any other refactoring you can think of? Or is the cure worse than the disease?

A local variable might also be fine.

fdr (commented):
.

serializers/web/github_runner.rb
enescakir: Makes sense. I focused too much on these lines.

I refactored now. With recent changes, this state isn't only the state of virtual machine. So I moved to the runner model as "display_state" similar to the other models.

model/github_runner.rb
```diff
@@ -25,6 +25,15 @@ def runner_url
     "http://github.com/#{repository_name}/settings/actions/runners/#{runner_id}" if runner_id
   end
 
+  def display_state
+    return vm.display_state if vm
+    if strand
+      return "deleted" if strand.label == "wait_vm_destroy"
+      return "reached_concurrency_limit" if strand.label == "wait_concurrency_limit"
+    end
+    "not_created"
```
fdr, line 34:
```suggestion 
    case strand&.label
    when "wait_vm_destroy"
      "deleted"
    when "wait_concurrency_limit"
      "reached_concurrency_limit"
    else
      "not_created"
    end
```

It blows up the line count, but probably has fewer expressions and definitely less flow control. Not entirely coincidentally, it's half the width.

fdr (changes requested):
presuming I retained all the semantics correctly

model/github_runner.rb
enescakir: Another alternative which doesn't blow up line counts

```
    case strand&.label
    when "wait_vm_destroy" then "deleted"
    when "wait_concurrency_limit" then "reached_concurrency_limit"
    else "not_created"
    end
```

fdr: Sure :)

bsatzger (approved):

views/project/github.erb
```diff
@@ -32,7 +32,7 @@
               <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">Runner</th>
               <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Repository</th>
               <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Label</th>
-              <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Vm</th>
+              <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">State</th>
```
fdr, line 35: is this an intentional renaming? It seems orthogonal from patch purpose ("don't crash")

fdr (approved):
maybe there's a bit of debris here you could separate or something, but no fundamental issues.

views/project/github.erb
enescakir: This is another outcome of moving it to model as `display_state`

From the commit message:
> With recent changes, this state isn't only the state of virtual machine. So I moved to the runner model as "display_state" similar to the other models

PR #1472, opened 2024-04-17
Update 2 serialization keys on vm and postgres apis

velioglu:
Updating serialization key display_size to size of vm and primary? to primary of postgres.


fdr (approved):
very reasonable.

enescakir (approved):

PR #1474, opened 2024-04-18
Make rodauth errors' scheme consistent with API's common error scheme

velioglu:
When requests fail at the authentication step, rodauth was generating it's own error messages in its own format. Converting those errors to the common scheme for the sake of consistency of the API errors.

### DIFF
--- a/clover_api.rb
+++ b/clover_api.rb
@@ -34,6 +34,42 @@ class CloverApi < Roda
     only_json? true
     use_jwt? true
 
+    # Converting rodauth error response to the common error format of the API
+    json_response_body do |hash|
+      # In case of an error, rodauth returns the error in the following format
+      # {
+      #   (required) "error": "There was an error logging in"
+      #   (optional) "field-error": [
+      #     "password",
+      #     "invalid password"
+      #   ]
+      # }
+      if json_response_error?
+        error_message = hash["error"]
+        type, code = case error_message
+        when "There was an error logging in"
+          ["InvalidCredentials", 401]
+        when "invalid JWT format or claim in Authorization header"
+          ["InvalidRequest", 400]
+        when "Please login to continue"
+          ["LoginRequired", 401]
+        else
+          # :nocov:
+          ["AuthenticationError", 401]
+          # :nocov:
+        end
+
+        hash.clear
+        hash["error"] = {
+          "code" => code,
+          "type" => type,
+          "message" => error_message
+        }
+      end
+
+      hash.to_json
+    end
+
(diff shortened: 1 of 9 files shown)


clover_api.rb
```diff
@@ -34,6 +34,42 @@ class CloverApi < Roda
...
+    json_response_body do |hash|
+      # In case of an error, rodauth returns the error in the following format
+      # {
+      #   (required) "error": "There was an error logging in"
+      #   (optional) "field-error": [
+      #     "password",
+      #     "invalid password"
+      #   ]
+      # }
+      if hash["error"]
```
enescakir, line 47: nit: Rodauth has `json_response_error?` helper. http://rodauth.jeremyevans.net/rdoc/files/doc/json_rdoc.html
```suggestion
      if hash["error"]
```

clover_api.rb
```diff
@@ -34,6 +34,42 @@ class CloverApi < Roda
...
+          ["Authentication Error", 401]
+          # :nocov:
+        end
+
+        hash.clear
+        hash["error"] = {
+          "code" => code,
+          "type" => type,
+          "message" => error_message
+        }
```
enescakir, line 67: Why do we clear hash instead of return a new one?
```suggestion
        return {
            "error" => {
                "code" => code,
                "type" => type,
                "message" => error_message
            }
        }.to_json
```

clover_api.rb
```diff
@@ -34,6 +34,42 @@ class CloverApi < Roda
...
+      #   (optional) "field-error": [
+      #     "password",
+      #     "invalid password"
+      #   ]
+      # }
+      if hash["error"]
+        error_message = hash["error"]
+        type, code = case error_message
+        when "There was an error logging in"
+          ["Invalid Credentials", 401]
```
enescakir, line 51: I was expecting error type to be more computer friendly such as `invalid_credentials` vs error message is human friendly.

enescakir (approved):

clover_api.rb
velioglu: I was also checking the same to have a consistent type. There is no common way exist, `invalid_credentials`, `INVALID_CREDENTIALS`, `InvalidCredentials` and `Invalid Credentials` are the examples I've seen. I've decided to go with `InvalidCredentials` as it doesn't look like a variable name or a constant value in the client codebase.

clover_api.rb
velioglu: I wanted to explicitly show that we are overwriting the hash returned by rodauth, let me keep it as it is.

PR #1478, opened 2024-04-18
Add Vm.update_spdk_version(...) 

pykello:
Previously to change SPDK version of a VM we needed to do multiple weird manual steps. This patch automates some of them and what we need to do is:

1. Install the new SPDK version on the host:

```
> Prog::Storage::SetupSpdk.assemble(
     vm_host.id, "some-spdk-version",
     start_service: true, allocation_weight: 100)
```

2. Update all VMs we want to migrate to this SPDK version. Note that this will just update related configs and won't take effect until host reboot:

```
> vm.update_spdk_version("some-spdk-version")
```

3. Reboot host:

```
> vm_host.incr_reboot
```

4. Optionally, remove the old SPDK installation:

```
> Prog::Storage::RemoveSpdk.assemble(old_spdk_installation.id)
```

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -21,7 +21,7 @@ class Vm < Sequel::Model
   include ResourceMethods
   include SemaphoreMethods
   include HealthMonitorMethods
-  semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup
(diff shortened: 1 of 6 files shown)


fdr (commented):
.

fdr (approved):
alright, though regenerating the unit file is a step into a different world just a little bit, where can't get some implied backwards compatibility anymore. We'll need to keep the re-generation in order with all extant VMs.

PR #1454, opened 2024-04-11
Usage alerts

byucesoy:
**Move base_url to config**
Current way of generating base_url depends on the request object. However, we
need base_url in contexts where request object is not available, such as while
sending e-mails from background jobs. With this commit, we make base_url a
config variable. Its default value as http://localhost:9292, which is what we
use for development environment and for production environment we can set it to
https://console.ubicloud.com/.

With this change, we lose the ability to dynamically set the url scheme, but
we redirect http to https in production anyway.

**Move send_email to util**
Previously send_email was part of CloverWeb. This makes it difficult to send an
email in contexts outside of the web application. Sending mail requires setting
up a whole web app, which is not a lightweight operation. The reason of putting
send_email in CloverWeb was to have access to render plugin, because we need to
render html pages as a part of email content. render plugin is part of roda, so
it is not available outside of roda app.

In this commit, I implemented a simple EmailRenderer. It is a wrapper around
Tilt::ErubiTemplate, which is the same as what is used in roda's render plugin.
With the new renderer, we can render templates without having to set up a roda
app and send emails from any context.

This also unblocks sending monthly emails for invoices.

**Add usage alerting based on the total consumption**
This commit adds feature of sending alerts when the total consumption exceeds
a user-defined threshold. We associate each alert to a user who created it and
send the email to that user when the alert is triggered. An alternative would
be allowing users to specify an email address to send the alert to, but that
opens an attack vector where a user could spam another user with alerts. This
vector can be mitigated by requiring the user to confirm the email address, but
implementing a whole confirmation workflow would significantly increase the
effort required to implement this feature.

Screenshot from 2024-04-14 19-59-58

Screenshot from 2024-04-15 13-15-28

PR #1479, opened 2024-04-18
Use Regexp literals in validations

fdr:
We used to store regular expressions as strings in the validations.rb which causes them to be compiled every time the validation is run. With this commit we are changing them to be stored as Regexp objects which are compiled only once at the load time.

### DIFF
--- a/lib/validation.rb
+++ b/lib/validation.rb
@@ -16,19 +16,19 @@ module Validation
   # - Not start or end with a hyphen
   # Adapted from https://stackoverflow.com/a/7933253
   # Do not allow uppercase letters to not deal with case sensitivity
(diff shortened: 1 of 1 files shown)


enescakir:
Does this PR conflict with this commit https://github.com/ubicloud/ubicloud/pull/1454/commits/f6b45993d6dcf5b7e5d70fd389e0d3ecb96170b6

fdr:
It's actually based on and cherry picked :) (I even kept the author)

fdr:
I just figured using the appropriate ruby literal had Less Going On (layers of escaping, even amount of evaluation on file load, though it's immaterial)

byucesoy (approved):

furkansahin (approved):

PR #1480, opened 2024-04-18
Add migration to index semaphore.strand_id

fdr:
Looking at our hot queries (i.e. frequency * duration = long), it appears we were seq scanning and disqualifying nearly all `semaphore` records running a query like:

     SELECT * FROM "semaphore" WHERE ("strand_id" = $1)

The result was excessive consumed CPU time.

The use of `CREATE INDEX CONCURRENTLY` is excessive, but it makes one of our inherited RuboCop rules happy.  Maybe I'll disable that rule later.


enescakir (approved):
LGTM

PR #1481, opened 2024-04-18
Allow reserving some cores per host for non-runner purposes.

pykello:
We currently use the whole available capacity of a host for github-runners. This can cause capacity issues for regular VMs. This PR 
allows reserving some cores for regular VMs by setting a threshold.

E.g. if we do the following:

```
vm_host.update(github_runner_allocation_core_threshold: 4)
```

Then it won't allocate a runner if number of available cores after provisioning the runner is below 4.

(I will add tests if the idea makes sense)


fdr:
it's pretty brutal, can we fix it by not uniformly distributing allocations? that seems more adaptive.

PR #1483, opened 2024-04-19
Fix Nexus.destroy volume asynchronous update issue

furkansahin:
We were using the available_storage_gib value we had in memory in the process when updating the storage device available_storage_gib. This caused async updates to overwrite each other's updates. This commit fixes the issue by using the valus in the database when updating.


byucesoy (approved):

enescakir (approved):
I hope to run E2E pipeline on this branch, but unfortunately E2E is down now

bsatzger (approved):

PR #1484, opened 2024-04-19
Add constraints to StorageDevice and not null VmVolume.storage_device_id

furkansahin:
**StorageDevice VmHost and VmStorageVolume schema updates** 
We have 3 main changes here.
1. Drop VmHost.available_storage_gib column;
This is not used anywhere. We solely depend on
StorageDevice.available_storage_gib during allocation. Therefore,
removing this column.

2. Drop VmHost.total_storage_gib column;
Not used anywhere. Removed to not cause confusion and async issues. We
added a wraaper function around storage devices to be safer.

3. Make VmStorageVolume.storage_device_id column not nullable;
We occasionally see a drift of StorageDevice.available_storage_gib vs
vm.vm_storage_volume used sizes. This is happening because of bugs and
not maintaining the relationship in between volume and the device
properly. In result, they cause negative available_storage_gib. This
commit introduces new constraints to block these cases so that we can be
aware of the situation.

4. Introduce non zero and smaller than total_storage_gib constraints to
the StorageDevice.available_storage_gib column.

**Replace VmHost.available_storage_gib column with a function** 
The availablae_storage_gib column in the VmHost comes from old times and
have no use in the codebase. It causes confusion and this is cleaning up
that column. Still for the sake of reaching the available storage size
in the host, I added a new function which is a wrapper around
storage_devices.

**Remove VmHost.total_storage_gib** 
This is not used anywhere, so, we are dropping the column. For the sake
of operator tooling, we add a wrapper function in the host model to find
out the total_storage_gib from the storage devices.

### DIFF
--- a/prog/vm/host_nexus.rb
+++ b/prog/vm/host_nexus.rb
@@ -79,14 +79,6 @@ class Prog::Vm::HostNexus < Prog::Base
           total_cores: total_cores,
           total_cpus: total_cpus
         }
-
-        vm_host.update(**kwargs)
-      when "LearnStorage"
-        kwargs = {
-          total_storage_gib: st.exitval.fetch("total_storage_gib"),
-          available_storage_gib: st.exitval.fetch("available_storage_gib")
-        }
-
         vm_host.update(**kwargs)
       end
     end
--- a/spec/prog/vm/nexus_spec.rb
+++ b/spec/prog/vm/nexus_spec.rb
@@ -393,7 +393,8 @@ RSpec.describe Prog::Vm::Nexus do
               arch: "x64"}.merge(args)
       sa = Sshable.create_with_id(host: "127.0.0.#{@host_index}")
       @host_index += 1
-      host = VmHost.create(**args) { _1.id = sa.id }
+
+      host = VmHost.create(**args.except(:available_storage_gib, :total_storage_gib)) { _1.id = sa.id }
       StorageDevice.create_with_id(
         name: "DEFAULT",
         available_storage_gib: args[:available_storage_gib],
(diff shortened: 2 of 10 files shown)


prog/vm/host_nexus.rb
```diff
@@ -83,8 +83,7 @@ def before_run
         vm_host.update(**kwargs)
       when "LearnStorage"
         kwargs = {
-          total_storage_gib: st.exitval.fetch("total_storage_gib"),
```
enescakir, line 86: Some learn progs, like LearnNetwork, directly update the VM host within the prog itself. Others, like LearnMemory, simply return the learned value, allowing the parent to perform the update. When I asked @fdr about this inconsistency in the past, he said, "We're unsure which method is superior. Let's maintain both styles and decide in the future." LearnStorage introduced a new pattern as well. It creates StorageDevices but returns the total storage for the VmHost model. Therefore, if we run LearnStorage but don't utilize the returned value, it could lead to some inconsistencies. IMHO, we need to decide whether to persist data within the prog or outside of it. In this context, we could also consider moving the update of vm host total_storage_gib into the LearnStorage prog itself.

furkansahin: Removed it.

migrate/20240419_remove_avail_st_gib.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:vm_host) do
+      drop_column :available_storage_gib
```
enescakir, line 6: Should we run dropping operations as the latest commit?

spec/prog/vm/nexus_spec.rb
```diff
@@ -393,7 +393,12 @@ def new_host(**args)
               arch: "x64"}.merge(args)
       sa = Sshable.create_with_id(host: "127.0.0.#{@host_index}")
       @host_index += 1
-      host = VmHost.create(**args) { _1.id = sa.id }
+      host_args = args.dup.tap {
+        _1.delete(:available_storage_gib)
+        _1.delete(:total_storage_gib)
+      }
+
+      host = VmHost.create(**host_args) { _1.id = sa.id }
```
enescakir, line 401:
```suggestion
      host = VmHost.create(**args.except([:available_storage_gib, :total_storage_gib])) { _1.id = sa.id }
```

enescakir (approved):
LGTM

PR #1486, opened 2024-04-19
Don't accept non-numeric value for page size

velioglu:
We've decided to not support non-numeric value for page size, adding a validation for that.


fdr (approved):
Phew was wondering if we'd forget about that one.

PR #1490, opened 2024-04-22
Track hugepages used for SPDK installations in the database. 

pykello:
We use number of hugepages used when installing & removing an SPDK installation. Although this is mostly static, but it is possible that we decide to change number of hugepages used when installing a new SPDK in future.

In which case, we should remember what value used for older installations to update VmHost records correctly when removing them.

This mismatch already happened in production last week, so making this change to avoid it in future.

### DIFF
--- a/migrate/20240423_add_spdk_hugepages.rb
+++ b/migrate/20240423_add_spdk_hugepages.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:spdk_installation) do
+      add_column :hugepages, :int, default: 2, null: false
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


rhizome/host/lib/spdk_setup.rb
```diff
@@ -85,6 +85,10 @@ def install_package
     end
   end
 
+  def hugepage_count
+    2
+  end
```
fdr, line 90: This seems like a lot of indirection to insert "2" somewhere, do you have a follow-up intended that makes this dynamic in some interesting way?

pykello: This will most likely become dynamic in future as max number of volumes per host will become non-constant. However, thinking more about it, I think we don't know what the correct change for those cases before we have an example. So, I removed the indirection to keep this change as simple as possible & defer that to future.

fdr (approved):
that does slim the patch quite a bit

PR #1492, opened 2024-04-23
Add check for fraud payment method

velioglu:
Adding a check whether the added payment method is tagged as fraud, if so user won't be able to add it. Payment methods will be tagged as fraud operationally.

### DIFF
--- a/migrate/20240423_add_fraud_columns.rb
+++ b/migrate/20240423_add_fraud_columns.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:payment_method) do
+      add_column :card_fingerprint, :text, null: true
+      add_column :fraud, :boolean, default: false, null: false
+    end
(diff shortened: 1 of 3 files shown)


fdr (approved):
neat feature. I wonder if it shows up in other payment processors, seems like it might.

migrate/20240423_add_fraud_columns.rb
```diff
@@ -0,0 +1,17 @@
...
+      add_column :fraud, :boolean, default: false, null: false
+    end
+  end
+
+  down do
+    alter_table(:payment_method) do
+      drop_column :card_fingerprint
+      drop_column :fraud
+    end
+  end
```
enescakir, line 16: nit: `change` automatically handles down
```suggestion
  change do
    alter_table(:payment_method) do
      add_column :card_fingerprint, :text, null: true
      add_column :fraud, :boolean, default: false, null: false
    end
  end
```

PR #1493, opened 2024-04-23
Quick and dirty suspend feature

fdr:
The case of re-allocating if CPU and/or memory resources have since allocated is left unsolved.

Also seen is grotesque copies of bookkeeping of VmHost core and memory counts, as well as some tests I declined to copy and paste for "systemctl stop" idempotency.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -9,6 +9,7 @@ require "base64"
 class Prog::Vm::Nexus < Prog::Base
   subject_is :vm
   semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup
+  semaphore :suspend, :unsuspend
 
   def self.assemble(public_key, project_id, name: nil, size: "standard-2",
     unix_user: "ubi", location: "hetzner-hel1", boot_image: "ubuntu-jammy",
@@ -489,6 +490,11 @@ WHERE (SELECT max(available_storage_gib) FROM storage_device WHERE storage_devic
   end
 
   label def wait
+    when_suspend_set? do
+      register_deadline(:suspended, 5 * 60)
+      hop_suspending
+    end
+
     when_start_after_host_reboot_set? do
       register_deadline(:wait, 5 * 60)
       hop_start_after_host_reboot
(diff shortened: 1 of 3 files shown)


furkansahin:
We can either exclude `suspending` and `suspended` states from the `before_run` and handle `destroy` path in these labels OR we need to be careful transitioning the VMs back to not suspended and then `destroy`. Otherwise, we will cause `used_cores` value to drift from reality.

prog/vm/nexus.rb
```diff
@@ -545,6 +551,44 @@ def before_run
...
+      host.sshable.cmd("sudo systemctl start #{q_vm} #{q_vm}-dnsmasq")
+      VmHost.dataset.where(id: vm.vm_host_id).update(
+        used_cores: Sequel[:used_cores] + vm.cores,
+        used_hugepages_1g: Sequel[:used_hugepages_1g] + vm.mem_gib
+      )
+      decr_unsuspend
+      hop_wait
+    end
+
+    nap 2**35
```
furkansahin, line 589: Do we have to nap this long? It simply means the unsespending will happen manually. Considering the semaphore increment will be done by the operator, I'm OK with this. But, maybe we can reduce it?

fdr: well, strange thing, last system routinely napped for a thousand years and was more event-driven. This system is actually regressing in reducing spurious polling-like traffic relative to its predecessor.

So it's odd it shows in one place, yet, we're kinda doing something wrong everywhere else where we could fall asleep for a long time.

byucesoy: Maybe we should schedule the strand when semaphore is incremented. It would definitely push things more towards being event driven and would encourage longer naps.

bsatzger: I like the idea of being more event driven, and incrementing semaphores could be one of these events. In this case, someone needs to log into the console already and increment the `unsuspend` semaphore. It's probably ok to then also `run` the strand to immediately perform the unsuspend action.

fdr: > Maybe we should schedule the strand when semaphore is incremented. It would definitely push things more towards being event driven and would encourage longer naps.

don't we do that? It tweaks sch to now.

```
  def self.incr(strand_id, name)
    DB.transaction do
      Strand.dataset.where(id: strand_id).update(schedule: Sequel::CURRENT_TIMESTAMP)
      Semaphore.create_with_id(strand_id: strand_id, name: name)
    end
  end
  ```

byucesoy: Ah cool. I did not realize that.

furkansahin:
Guess this is not gonna get merged.

PR #1495, opened 2024-04-23
Use location.display_name at POST

furkansahin:
With the renaming of the location, we made changes in the way we display the location to the user. However, we should also make use of that in the POST requests while creating or modifying resources. Otherwise, it is not consistent with what the user sees and the endpoints accept. It also creates an issue with choosing PrivateSubnets at the time of VM creation. Because, the location of the subnet is the display_location and the value of the location boxes in the create page is the internal location names. Since they do not much, the customer cannot pick the subnet in the region they are trying to provision the VM in.

### DIFF
--- a/config/providers.yml
+++ b/config/providers.yml
@@ -1,0 +1,11 @@
+[
+  {
+    provider_display_name: Hetzner,
+    provider_internal_name: hetzner,
+    locations: [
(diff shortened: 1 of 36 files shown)


velioglu:
I've included the change in the #1421 , we can close this one.

PR #1497, opened 2024-04-23
Fix redirect after usage alert deletion

byucesoy:
We used to redirect project/<ubid>/usage-alert after deleting a usage alert, however, there is no such page. We should redirect to project/<ubid>/billing instead.


fdr (approved):
Oops. I wonder how plausible it is to write link checking code.

PR #1498, opened 2024-04-23
Update error response title with PascalCase type

velioglu:
In the RFC https://datatracker.ietf.org/doc/html/rfc7807, type is
the only key required as a part of the error response. It is used
to decide the kind of the error obtained from the API. Client codes
will be using that type to understand the kind of the error. Note that,
same error code (which is basically the http error code) might have
different error type. To show that it is a constant text returned from an
API, value of the type key converted to the PascalCase.

As a side effect of that change, texts shown as title on UI also became
PascalCase. As it still informs the user on the UI as it was informing
before that commit, it is acceptable.

### DIFF
--- a/clover_api.rb
+++ b/clover_api.rb
@@ -16,7 +16,7 @@ class CloverApi < Roda
     {
       error: {
         code: 404,
-        title: "Resource not found",
+        type: "ResourceNotFound",
         message: "Sorry, we couldn’t find the resource you’re looking for."
       }
     }.to_json
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -44,7 +44,7 @@ class CloverWeb < Roda
   plugin :not_found do
     @error = {
       code: 404,
-      title: "Resource not found",
+      type: "ResourceNotFound",
       message: "Sorry, we couldn’t find the resource you’re looking for."
     }
 
--- a/lib/validation.rb
+++ b/lib/validation.rb
@@ -6,7 +6,7 @@ require "netaddr"
 module Validation
   class ValidationFailed < CloverError
     def initialize(details)
-      super(400, "Invalid request", "Validation failed for following fields: #{details.keys.join(", ")}", details)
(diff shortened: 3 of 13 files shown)


fdr (approved):
is it really a "title" anymore if so?

I still don't know what that field is for: short descriptions in a table-like context (whether html or cli), or to be useful for code generation (like the AWS's `code` attribute if memory serves, which is not the same as our `code`, which is perhaps redundant HTTP information...).

So, no real problem moving it around, but "we decided" is one of those things that'll make someone slap their head if they dig up the origin of the change in the log.

velioglu:
> is it really a "title" anymore if so?
> 
> I still don't know what that field is for: short descriptions in a table-like context (whether html or cli), or to be useful for code generation (like the AWS's `code` attribute if memory serves, which is not the same as our `code`, which is perhaps redundant HTTP information...).
> 
> So, no real problem moving it around, but "As we've decided" is one of those things that'll make someone slap their head if they dig up the origin of the change in the log.

I was planning to update title to type with a subsequent PR, but decided to include it in this one after your commit. 

Code and type will be different than each other as the code means the http error code of the response but type shows the kind of it. Same code might have different types. Also same type might have different messages.

I've said as we've already decided thinking about https://github.com/ubicloud/ubicloud/pull/1474#discussion_r1571343924 but I've updated the PR description to show our reasoning better.

fdr (approved):
okay, interesting, TIL about the RFC, I remember discussing `type` with you but I was always a bit fuzzy about title and the extent of their substitution. 

It's a bit ugly to leak the PascalCase directly to the customer in titles, but that can be smoothed out with More Programming at our option, a bit like `strerror`. No need to inline it into the API I don't think, as you have it here

fdr:
Do you have an idea of how to do nested/chained exception rendering here? I suspect we could wiggle it in (even the ruby-esque `cause`) and repeating the fields, if we really, really wanted to (perhaps never, though)

PR #1502, opened 2024-04-24
Optimize wait_sshable and firewall rule dance

furkansahin:
**Add tcp dport 22 allow by default at the provisioning time**
We add a new rule to be used at the provisioning time to verify ssh
connectivity to all VMs. This rule is immediately removed and the
customer firewall rules are applied the moment VM is declared running.

**Optimize wait_sshable and firewall rule dance**
In this commit, we make a couple of changes to optimize the wait-sshable
and firewall rule application logic at the provisioning time.
1. Immediately run the vm and verify ssh connectivity before firewall
rules are applied.
2. Use push to update firewall rules instead of bud and dummy wait.
3. For the first time we get into wait_sshable, we immediately nap for
15 seconds. This is done with the assumption of wait-sshable will never
take less than 15 seconds and we can just sleep instead of using
compute. It should optimize the amount of compute usage especially at
the busy time.

With the help of these three changes, we are optimizing the amount of time
spent on firewall rule application and the sshability verification.

**Remove verify-sshable script**


prog/vm/nexus.rb
```diff
@@ -432,33 +431,23 @@ def before_run
...
   label def run
     host.sshable.cmd("sudo systemctl start #{q_vm}")
     hop_wait_sshable
   end
 
   label def wait_sshable
-    case host.sshable.cmd("common/bin/daemonizer --check wait_sshable_#{q_vm}")
-    when "Succeeded"
-      host.sshable.cmd("common/bin/daemonizer --clean wait_sshable_#{q_vm}")
+    vm.incr_update_firewall_rules
```
bsatzger, line 440: only increment if not incremented already?

rhizome/host/lib/vm_setup.rb
```diff
@@ -364,6 +364,12 @@ def build_nftables_config(gua, nics, ip4)
       }
       # NAT4 rules
       #{generate_nat4_rules(ip4, nics.first.net4)}
+      table inet fw_table {
+        chain forward_ingress {
+          type filter hook forward priority filter; policy drop;
+          ip saddr 0.0.0.0/0 tcp dport 22 ip daddr #{nics.first.net4} counter accept
```
bsatzger, line 370: port 22 is always open? would this be closed after sshable is confirmed?

prog/vm/nexus.rb
```diff
@@ -432,33 +431,23 @@ def before_run
...
+    addr = vm.ephemeral_net4
+    create_billing_record unless addr
```
bsatzger, line 442: why call this here and hop to `create_billing_record` later?

rhizome/host/lib/vm_setup.rb
furkansahin: It is removed immediately after we switch to `wait` state because we apply firewall rules, there.

bsatzger (approved):

PR #1503, opened 2024-04-24
US Region Prep

furkansahin:
Introduces new billing rates, provider and makes changes to dataplane to support a US region.


fdr (approved):
Might want to include on that last one that it happened to be included on the hetzner image but not leaseweb's. While the title is useful (for what program are we installing this) it raises the question "uhhh how did anything work?

config/providers.yml
```diff
@@ -7,5 +7,12 @@
       {display_name: eu-north-h1, internal_name: hetzner-hel1, visible: true },
       {display_name: github-runners, internal_name: github-runners, visible: false }
     ]
+  },
+  {
+    provider_display_name: Leaseweb,
+    provider_internal_name: leaseweb,
+    locations: [
+      {display_name: us-central-l1, internal_name: leaseweb-wdc02, visible: true },
```
enescakir, line 15: Why isn't it `us-central-l2` instead of `us-central-l1`, given that it's `wdc02`?

Is Washington DC more central US than east US?"

fdr:
woops shouldn't it be east?

config/providers.yml
enescakir: Yes, I mixed them :D the correct one "Is Washington DC more east US than central US?"

furkansahin: Updated the names after checking with Umur.

PR #1504, opened 2024-04-24
Prioritize "github-runners" location for github runners.

pykello:
We had several capacity incidents for VM or PG allocations, because github-runners used capacity of all hosts. This patch will make sure that we use non "github-runners" locations for runners only when we have to. Therefore, leaving more capacity for VM and PG.

Note that we might not need special ordering for github-runners location after completing the fleet unification. This is a temporary solution to reduce capacity incidents for VM and PG allocations.


fdr (approved):
I dig it. clever, and given what we know about not only having to test new product lines from vendors, but perhaps even cohorts of servers....timely.

PR #1506, opened 2024-04-24
Add more log to monitor runner related metrics

enescakir:
We strive to actively monitor metrics like runner provisioning duration and job assignment, as they significantly affect our customers' experience. We don't require all properties from the `GithubRunner` and `Vm` model, so I've created a custom hash instead of using `model.values`.  These fields assist us in filtering our runner metrics based on various parameters.

I wrote the `log_duration` function to avoid repeatedly testing same condition branches and I don't want to add `:nocov`., I considered adding this helper function to the prog, but since it's also used in the web route to log the job's starting duration, I decided to include it in the model.

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -34,6 +34,13 @@ class GithubRunner < Sequel::Model
     end
   end
 
+  def log_duration(message, duration)
+    values = {ubid: ubid, label: label, repository_name: repository_name, duration: duration, vm_ubid: vm.ubid, arch: vm.arch, cores: vm.cores}
+    values[:vm_host_ubid] = vm.vm_host.ubid if vm.vm_host
+    values[:vm_pool_ubid] = VmPool[vm.pool_id].ubid if vm.pool_id
+    Clog.emit(message) { {message => values} }
+  end
+
   def init_health_monitor_session
     {
       ssh_session: vm.sshable.start_fresh_session
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -40,7 +40,6 @@ class Prog::Vm::GithubRunner < Prog::Base
     ).first
 
     if (picked_vm = pool&.pick_vm)
-      Clog.emit("Pool is used") { {github_runner: {label: github_runner.label, repository_name: github_runner.repository_name, cores: picked_vm.cores}} }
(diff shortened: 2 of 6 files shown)


prog/vm/github_runner.rb
```diff
@@ -154,6 +152,7 @@ def before_run
     picked_vm = pick_vm
     github_runner.update(vm_id: picked_vm.id)
     picked_vm.update(name: github_runner.ubid.to_s)
+    github_runner.reload.log_duration("runner allocated", Time.now - github_runner.created_at)
```
bsatzger, line 155: do we have to reload here?

enescakir: We updated `vm_id` just two lines ago, but it returns `nil` without reloading. I could attempt to use the `github_runner.add_vm(picked_vm)` method from Sequel. This might update the `.vm` association.

model/github_runner.rb
```diff
@@ -34,6 +34,13 @@ def display_state
     end
   end
 
+  def log_duration(message, duration)
+    values = {ubid: ubid, label: label, repository_name: repository_name, duration: duration, vm_ubid: vm.ubid, arch: vm.arch, cores: vm.cores}
+    values[:vm_host_ubid] = vm.vm_host.ubid if vm.vm_host
+    values[:pool_ubid] = VmPool[vm.pool_id].ubid if vm.pool_id
+    Clog.emit(message) { {message.tr(" ", "_") => values} }
```
byucesoy, line 41: Why do we do replacement here instead of sending message with underscores in the first place?

prog/vm/github_runner.rb
byucesoy: `github_runner.add_vm(picked_vm)` might have other unintended side effects (need to check the sequel code)

byucesoy (approved):

prog/vm/github_runner.rb
byucesoy: `github_runner.add_vm(picked_vm)` might have other unintended side effects (need to check sequel code)

enescakir: Since we don't have example of `add_*` in our codebase, I'll leave it as is.

furkansahin (approved):

PR #1508, opened 2024-04-25
Prioritize monitoring of non-ephemeral resources

byucesoy:
We have limited number of threads to monitor all resources. We want to use them to monitor non-ephemeral resources first. This PR changes the resources order while adding resources to the monitor queue.

### DIFF
--- a/bin/monitor
+++ b/bin/monitor
@@ -3,7 +3,7 @@
 
 require_relative "../loader"
 
-MONITORABLE_RESOURCE_TYPES = [PostgresServer, MinioServer, GithubRunner, VmHost, Vm.where(~Sshable.where(id: Sequel[:vm][:id]).exists)]
+MONITORABLE_RESOURCE_TYPES = [VmHost, PostgresServer, Vm.where(~Sshable.where(id: Sequel[:vm][:id]).exists), MinioServer, GithubRunner]
 
 sessions = {}
 ssh_threads = {}


bin/monitor
```diff
@@ -3,7 +3,7 @@
 
 require_relative "../loader"
 
-MONITORABLE_RESOURCE_TYPES = [PostgresServer, MinioServer, GithubRunner, VmHost, Vm.where(~Sshable.where(id: Sequel[:vm][:id]).exists)]
+MONITORABLE_RESOURCE_TYPES = [VmHost, PostgresServer, Vm.where(~Sshable.where(id: Sequel[:vm][:id]).exists), MinioServer, GithubRunner]
```
enescakir, line 6: I think MinIO server is less ephemeral than a VM
```suggestion
MONITORABLE_RESOURCE_TYPES = [VmHost, MinioServer, PostgresServer, Vm.where(~Sshable.where(id: Sequel[:vm][:id]).exists), GithubRunner]
```

enescakir (approved):

bin/monitor
enescakir: And it's also a prerequisite for PG :D So IMHO, VmHost > MinIO > PG > Vm > Runner

byucesoy: I put the MinIo later in the list because if it is down, it doesn't immediately create a customer impact.

Backup/Restore and new host initialization would not work, but user workloads are still running.

PR #1514, opened 2024-04-26
Do not use the same runner when the script failed

enescakir:
### Add a helper to create spare runner

Our runners are job agnostic, meaning they can run any job with the matched label. This enables us to create a spare runner with the same label if the initial one doesn't function properly.

This helper is also useful for on-call engineers.

### Do not use the same runner when the script failed

Currently, when the script failed, we assume it failed because of an initialization error, and we try to register the same runner again. This is not always true. The script might be "failed" while running the workflow. We should create a new spare runner and destroy the failed one.

I can implement additional checks such as verify if the runner has completed the job, and avoid creating a spare runner if the job is completed. However, runner script failures are uncommon. Even when some errors occur, they exit with a zero exit code, not a non-zero one generally. Therefore, I believe it's currently unnecessary to add more checks. If script failures increase, we can reconsider adding them.

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -41,6 +41,10 @@ class GithubRunner < Sequel::Model
     Clog.emit(message) { {message => values} }
   end
 
+  def provision_spare_runner
+    Prog::Vm::GithubRunner.assemble(installation, repository_name: repository_name, label: label).subject
+  end
+
   def init_health_monitor_session
     {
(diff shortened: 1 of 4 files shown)


bsatzger (approved):
Thanks! Would be great if you could update the wiki on how to properly use `provision_spare_runner` as on-call.

PR #1515, opened 2024-04-26
Change password changed email with our template

enescakir:
We've replaced Rodauth's default text-based emails with our own HTML emails.

It seems we forgot to replace the `send_password_changed_email` in the `change_password_notify` plugin.

### DIFF
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -184,6 +184,14 @@ class CloverWeb < Roda
       remove_all_active_sessions_except_current
     end
 
+    send_password_changed_email do
+      user = Account[account_id]
+      Util.send_email(email_to, "Ubicloud Account Password Changed",
+        greeting: "Hello #{user.name},",
+        body: ["Someone has changed the password for the account associated to this email address.",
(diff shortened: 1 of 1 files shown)


byucesoy (approved):

clover_web.rb
```diff
@@ -184,6 +184,14 @@ def assets
       remove_all_active_sessions_except_current
     end
 
+    send_password_changed_email do
+      user = Account[account_id]
+      Util.send_email(email_to, "Ubicloud Account Password Changed",
```
byucesoy, line 189: Is there any difference between using `email_to` vs `user.email`

byucesoy (approved):

clover_web.rb
enescakir: No difference for this example. https://github.com/jeremyevans/rodauth/blob/b44b4412fae55b9d97abaa34ab1e08434b161b90/lib/rodauth/features/email_base.rb#L30

For certain situations, like email change verification notifications, we must be cautious. We should also verify the value in other scenarios.

PR #1520, opened 2024-04-29
Add connection and channel initialize durations to sshable log

enescakir:
We record the total time the `sshable.cmd` takes, which includes the time needed to establish a connection and initialize the channel. We also maintain an SSH connection cache, reducing connection times when there is an existing cache. This PR logs the time taken to connect and initialize the channel, allowing us to see the actual time spent on command execution.


bsatzger (approved):
thanks!

byucesoy (approved):

PR #1524, opened 2024-04-30
Refactor image download in rhizome side & add optional image version.

pykello:
This is in preparation to make boot images versioned. Some of the reasons to do this includes:

* Allow installing a newer version of an image without breaking existing VMs which have references to the older versions.
* Make storage snapshots easier: we know exactly which base image is being used, and we can just snapshot the diff.
* Making correlation of failures to bad OS images easier.

Current patch:
* Moves almost all of the image download logic to the standalone class BootImage.
* Adds optional version and sha256sum parameters to images. If these are null or empty, the behaviour is exactly same as before.
* Uses BootImage in VmSetup and `download-boot-image`.

If an image is versioned, it'll be stored as `/var/storage/images/$name-$version.raw`. Legacy unversioned images are still supported and are stored at `/var/storage/images/$name.raw`.

### DIFF
--- a/rhizome/host/lib/boot_image.rb
+++ b/rhizome/host/lib/boot_image.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+require "digest"
+require "fileutils"
+require "uri"
+require_relative "../../common/lib/arch"
+
+class BootImage
+  def initialize(name, version)
+    @name = name
+    @version = version
+  end
+
+  def image_path
+    @image_path ||= if @version.nil?
+      "#{image_root}/#{@name}.raw"
+    else
+      "#{image_root}/#{@name}-#{@version}.raw"
+    end
+  end
+
+  def image_root
+    "/var/storage/images"
+  end
+
+  def download(url: nil, ca_path: nil, sha256sum: nil)
+    return if File.exist?(image_path)
+
+    url ||= get_download_url
+
+    fail "Must provide url for #{@name} image" if url.nil?
+    FileUtils.mkdir_p image_root
+
+    # If image URL has query parameter such as SAS token, File.extname returns
+    # it too. We need to remove them and only get extension.
+    ext = image_ext(url)
+    init_format = initial_format(ext)
+
+    # Use of File::EXCL provokes a crash rather than a race
+    # condition if two VMs are lazily getting their images at the
+    # same time.
+    temp_file_name = @version.nil? ? @name : "#{@name}-#{@version}"
+    temp_path = File.join(image_root, "#{temp_file_name}#{ext}.tmp")
+    curl_image(url, temp_path, ca_path)
+    verify_sha256sum(temp_path, sha256sum)
+    convert_image(temp_path, init_format)
+
+    rm_if_exists(temp_path)
+  end
+
+  def image_ext(url)
+    File.extname(URI.parse(url).path)
+  end
+
+  def initial_format(ext)
+    case ext
+    when ".qcow2", ".img"
+      "qcow2"
+    when ".vhd"
+      "vpc"
+    when ".raw"
+      "raw"
+    else
+      fail "Unsupported boot_image format: #{ext}"
+    end
+  end
+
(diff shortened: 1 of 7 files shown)


enescakir:
I was planning to implement fixed versioning for the runner image. This PR will be very helpful. Thank you.

rhizome/host/lib/boot_image.rb
```diff
@@ -0,0 +1,86 @@
...
+    @version = version || ""
+  end
+
+  def image_path
+    @image_path ||= if @version.empty?
+      "#{image_root}/#{@name}.raw"
+    else
+      "#{image_root}/#{@name}-#{@version}.raw"
+    end
+  end
```
enescakir, line 19: `nil` is a better indicator of a value's nonexistence.
```suggestion
  def initialize(name, version = nil)
    @name = name
    @version = version
  end

  def image_path
    @image_path ||= if @version
      "#{image_root}/#{@name}.raw"
    else
      "#{image_root}/#{@name}-#{@version}.raw"
    end
  end
```

rhizome/host/lib/vm_setup.rb
```diff
@@ -490,44 +491,8 @@ def download_boot_image(boot_image, custom_url: nil, ca_path: nil)
     }
 
     download = urls.fetch(boot_image) || custom_url
```
enescakir, line 493: What do you think about moving `urls` hash to the `BootImage` class too?

rhizome/host/lib/boot_image.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def image_root
+    "/var/storage/images"
+  end
+
+  def download(url:, ca_path: "", sha256sum: "")
```
enescakir, line 25: Why don't we use `nil` instead of `""`?

rhizome/host/lib/vm_setup.rb
pykello: I thought more about this & decided to keep URLs & sha256 sums in the rhizome side. So, I also did (a version of) what you suggested.

rhizome/host/lib/boot_image.rb
pykello: I did as you suggested, but didn't provide a default for `version`, at won't be optional after the transition to the new system.

enescakir (approved):
LGTM

byucesoy (commented):
Overall looks good. My only concern is the difficulty of updating hashes. I think it would be good to have a simpler way to store the list of hashes; such as passing them during first download.

rhizome/host/lib/boot_image.rb
```diff
@@ -0,0 +1,124 @@
...
+    }
+    urls.fetch(@name)
+  end
+
+  def get_sha256_sum
+    hashes = {
+      ["ubuntu-jammy", "x64", "20240319"] => "304983616fcba6ee1452e9f38993d7d3b8a90e1eb65fb0054d672ce23294d812",
+      ["ubuntu-jammy", "arm64", "20240319"] => "40ea1181447b9395fa03f6f2c405482fe532a348cc46fbb876effcfbbb35336f",
+      ["almalinux-9.3", "x64", "20231113"] => "6bbd060c971fd827a544c7e5e991a7d9e44460a449d2d058a0bb1290dec5a114",
+      ["almalinux-9.3", "arm64", "20231113"] => "a064715bc755346d5a8e1a4c6b1b7abffe4de03f1b0584942d5483ed32aafd67"
```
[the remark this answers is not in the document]
pykello: I'll move these & the URLs to the rhizome side in a later commit and `download_image` will pass all of these values. Since this PR is meant to not change the previous behaviour (where we don't use `download_image` for ubuntu images at all), I'm keeping the URLs and hashes in the clover side.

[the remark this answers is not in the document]
pykello: After some thinking, I think this map isn't needed in this PR at all. I removed it & will add it in the next PR when we do the clover changes.

PR #1526, opened 2024-05-01
Download & use versioned images

pykello:
In https://github.com/ubicloud/ubicloud/pull/1524 we changed rhizome to be able to download versioned boot images. In this change, we actually use that infrastructure.

To download a versioned boot image, one can do:

```
> vmh.download_boot_image("ubuntu-jammy", version: "20240319")
```

After this has been done, all VMs on that host which want to use the `ubuntu-jammy` image will use the latest boot image.

This allows us to be able to download new image versions without impacting previous VMs at all. Previous VMs will continue to use the older images.

As follow up items:
* VmHost setup will include downloading some default images (e.g. `ubuntu-jammy`) with explicit. vm_setup won't download the images automatically.
* We will enforce to use versioned images always
* Add a program to remove unused images

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -13,8 +13,9 @@ class VmHost < Sequel::Model
   one_to_many :spdk_installations, key: :vm_host_id
   one_to_many :storage_devices, key: :vm_host_id
   one_to_many :pci_devices, key: :vm_host_id
+  one_to_many :boot_images, key: :vm_host_id
 
-  plugin :association_dependencies, assigned_host_addresses: :destroy, assigned_subnets: :destroy, hetzner_host: :destroy, spdk_installations: :destroy, storage_devices: :destroy
+  plugin :association_dependencies, assigned_host_addresses: :destroy, assigned_subnets: :destroy, hetzner_host: :destroy, spdk_installations: :destroy, storage_devices: :destroy, boot_images: :destroy
 
   include ResourceMethods
   include SemaphoreMethods
@@ -183,9 +184,8 @@ class VmHost < Sequel::Model
   end
 
   # Introduced for downloading a new boot image via REPL.
-  # Use with caution as the vm_host will not accept a new vm during the image download.
-  def download_boot_image(image_name, custom_url = nil)
-    Strand.create_with_id(schedule: Time.now, prog: "DownloadBootImage", label: "start", stack: [{subject_id: id, image_name: image_name, custom_url: custom_url}])
+  def download_boot_image(image_name, version:, custom_url: nil)
+    Strand.create_with_id(schedule: Time.now, prog: "DownloadBootImage", label: "start", stack: [{subject_id: id, image_name: image_name, custom_url: custom_url, version: version}])
   end
 
   def hetznerify(server_id)
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -36,7 +36,7 @@ class Prog::DownloadBootImage < Prog::Base
     case sshable.cmd("common/bin/daemonizer --check download_#{image_name.shellescape}")
     when "Succeeded"
       sshable.cmd("common/bin/daemonizer --clean download_#{image_name.shellescape}")
-      hop_learn_storage
+      hop_update_available_storage_space
     when "NotStarted"
       url = custom_url || blob_storage_client.get_presigned_url("GET", Config.ubicloud_images_bucket_name, "#{image_name}-#{vm_host.arch}.raw", 60 * 60).to_s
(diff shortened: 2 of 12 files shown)


prog/download_boot_image.rb
```diff
@@ -1,14 +1,25 @@
...
-    @custom_url ||= frame["custom_url"]
+  def version
+    @version ||= frame.fetch("version")
+  end
+
+  def url
+    @url ||=
+      if frame["custom_url"]
+        frame["custom_url"]
+      elsif image_name.start_with?("github")
```
enescakir, line 20: We also have custom PG images.
```suggestion
      elsif image_name.start_with?("github") || image_name.start_with?("postgres")
```

byucesoy: ah, and actually we were considering using different prefixes for partnerships, such as `companyname-postgres`, but that is OK, we can switch to `postgres-companyname`

prog/vm/nexus.rb
```diff
@@ -240,6 +241,16 @@ def allocate
     vm_host_id
   end
 
+  def allocate_boot_image(vm_host, storage_volume, boot_image_name)
```
enescakir, line 244: This part conflicts with @bsatzger's allocator improvement: https://github.com/ubicloud/ubicloud/pull/1551. Whoever merges first, the other will need to incorporate these changes. fyi

pykello:
adding @bsatzger for allocator changes

migrate/20240503_boot_image.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:boot_image) do
```
enescakir, line 5: Should we add a `created_at` column to this table? It could be useful to know when we downloaded the image to the host.
```
      column :created_at, :timestamptz, null: false, default: Sequel.lit("now()")
```

migrate/20240503_boot_image.rb
```diff
@@ -0,0 +1,18 @@
...
+      column :id, :uuid, primary_key: true
+      foreign_key :vm_host_id, :vm_host, type: :uuid, null: false
+      column :name, :text, collate: '"C"', null: false
+      column :version, :text, collate: '"C"', null: false
+      column :active, :boolean, null: false, default: false
```
enescakir, line 10: What do you think about converting this column a timestamp as `activated_at`? It could be useful to know when we activated this version on the host. `nil` can indicate not activated.

prog/download_boot_image.rb
```diff
@@ -21,26 +49,32 @@ def blob_storage_client
...
-    nap 15 unless vm_host.vms_dataset.where(boot_image: image_name).empty?
-
-    sshable.cmd("sudo rm -f /var/storage/images/#{image_name.shellescape}.raw")
+    fail "Image already exists on host" if vm_host.boot_images_dataset.where(name: image_name, version: version).count > 0
+    BootImage.create(
+      vm_host_id: vm_host.id,
+      name: image_name,
+      version: version,
+      active: false
+    ) { _1.id = UBID.generate(UBID::TYPE_ETC).to_uuid }
```
enescakir, line 58: If you override `self.ubid_type` method, you can use `create_with_id` directly.

https://github.com/ubicloud/ubicloud/blob/f5cea35e33df427dbe3402b19b31dba7039ade4d/model/storage_device.rb#L10-L12

model/vm_host.rb
```diff
@@ -13,6 +13,7 @@ class VmHost < Sequel::Model
   one_to_many :spdk_installations, key: :vm_host_id
   one_to_many :storage_devices, key: :vm_host_id
   one_to_many :pci_devices, key: :vm_host_id
+  one_to_many :boot_images, key: :vm_host_id
 
   plugin :association_dependencies, assigned_host_addresses: :destroy, assigned_subnets: :destroy, hetzner_host: :destroy, spdk_installations: :destroy, storage_devices: :destroy
```
enescakir, line 18: Can you add `boot_images` here? They should be deleted when the vm host deleted.
https://github.com/ubicloud/ubicloud/blob/f5cea35e33df427dbe3402b19b31dba7039ade4d/model/vm_host.rb#L17

bsatzger (approved):

enescakir (approved):

enescakir (approved):
After all hosts have been associated with boot images, we'll force the allocator to select a host with this image, correct?

pykello:
> After all hosts have been associated with boot images, we'll force the allocator to select a host with this image, correct?

yes

PR #1531, opened 2024-05-02
Use private network for communication of PostgreSQL primary and standby

byucesoy:
**Provision standbys in the primary's private subnet**

Previously, we created separate subnets for the primary and standby and used
public network for all communication between them. This change provisions the
standby node in the primary's private subnet, which is more secure. It also
enables the standby to use the primary's private IP address for replication,
thus circumventing the user firewalls.

**Use private IPs in /etc/hosts file of PostgresSQL servers**
We use /etc/hosts for domain name resolution for the internal communication
between the primary and standby servers. This change ensures that the domain
name resolved to the private IP address, so that the all communication between
the primary and standby servers happens over the private network.

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -88,7 +88,7 @@ class PostgresServer < Sequel::Model
         }
       },
       identity: resource.identity,
-      hosts: "#{resource.representative_server.vm.ephemeral_net4} #{resource.identity}"
+      hosts: "#{resource.representative_server.vm.nics.first.private_ipv4.network} #{resource.identity}"
     }
   end
 
--- a/prog/postgres/postgres_resource_nexus.rb
+++ b/prog/postgres/postgres_resource_nexus.rb
@@ -49,9 +49,9 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
 
       PostgresFirewallRule.create_with_id(postgres_resource_id: postgres_resource.id, cidr: "0.0.0.0/0")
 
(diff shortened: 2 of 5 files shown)


model/postgres/postgres_server.rb
```diff
@@ -88,7 +88,7 @@ def configure_hash
         }
       },
       identity: resource.identity,
-      hosts: "#{resource.representative_server.vm.ephemeral_net4} #{resource.identity}"
```
enescakir, line 91: What happens if we reconfigure existing resources? They're not on the same subnet

furkansahin: What do you mean? PrivateSubnets span the whole region. We always put them into the same subnet.

enescakir: Weren't the primary and standby in different subnets before this PR? So primary and standy aren't in the same subnet for the existing PG resources.

furkansahin: Oh, now I understand. Yes, this is a problem after migration for old resources.

byucesoy: There are only 2 such resources. I can migrate them after merging this PR.

furkansahin: The migration would involve creating new standby instances and deprovisioning the previous one, I suppose? Just wanted to make sure.

furkansahin (approved):

model/postgres/postgres_server.rb
byucesoy: Yep. Actually just deprovisioning old one. New one would get created automatically. There would be no customer impact.

enescakir (approved):

PR #1534, opened 2024-05-02
Send email when the invoice processed

enescakir:
## Improve invoice charging function

We manually charge invoices. I've resolved some issues we encountered during the charging process.

Firstly, I switched the logging method from 'puts' to 'Clog'.

We attempted to charge each payment method listed in the billing information sequentially. Sometimes, the Stripe Ruby SDK throws an error when a payment fails. In such cases, we need to catch the error and proceed to charge the next payment method.

If we successfully charge an invoice, there's no need to continue trying other payment methods, so we break the loop.

If all payment methods fail, we leave the invoice unpaid.

## Add cc parameter to the email sending function

We need to cc the support email when sending a failed payment email to the customer. This way, the support team can follow up with the customer if necessary.

## Send an email when the invoice is successfully charged

We email the customer once the invoice is successfully processed.

We plan to include the invoice as a PDF attachment in future emails. However, generating a PDF on the backend is complex, so we'll add this feature in the next update.

## Send an email when the invoice is failed to charge

We automate the process of sending emails when an invoice fails to charge, including error details from the Stripe API.

We also add the support email as a CC, allowing the entire team to follow up with the customer if needed.

### DIFF
--- a/model/invoice.rb
+++ b/model/invoice.rb
@@ -15,58 +15,66 @@ class Invoice < Sequel::Model
   end
 
   def charge
+    reload # Reload to get the latest status to avoid double charging
     unless (Stripe.api_key = Config.stripe_secret_key)
-      puts "Billing is not enabled. Set STRIPE_SECRET_KEY to enable billing."
-      return
+      Clog.emit("Billing is not enabled. Set STRIPE_SECRET_KEY to enable billing.")
+      return true
     end
 
     if status != "unpaid"
-      puts "Invoice[#{ubid}] already charged: #{status}"
-      return
+      Clog.emit("Invoice already charged.") { {invoice_already_charged: {ubid: ubid, status: status}} }
+      return true
     end
 
-    if content["cost"] < Config.minimum_invoice_charge_threshold
+    amount = content["cost"].to_f.round(2)
+    if amount < Config.minimum_invoice_charge_threshold
       update(status: "below_minimum_threshold")
-      puts "Invoice[#{ubid}] cost is less than minimum charge cost: $#{content["cost"]}"
-      return
+      Clog.emit("Invoice cost is less than minimum charge cost.") { {invoice_below_threshold: {ubid: ubid, cost: amount}} }
+      return true
     end
 
-    unless (billing_info = BillingInfo[content.dig("billing_info", "id")])
-      puts "Invoice[#{ubid}] doesn't have billing info"
-      return
+    if (billing_info = BillingInfo[content.dig("billing_info", "id")]).nil? || billing_info.payment_methods.empty?
+      Clog.emit("Invoice doesn't have billing info.") { {invoice_no_billing: {ubid: ubid}} }
+      return false
     end
 
-    payment_methods = billing_info.payment_methods_dataset.order(:order).all
-
-    payment_methods.each do |payment_method|
-      amount = content["cost"].to_f.round(2)
-      payment_intent = Stripe::PaymentIntent.create({
-        amount: (amount * 100).to_i, # 100 cents to charge $1.00
-        currency: "usd",
-        confirm: true,
-        off_session: true,
-        customer: billing_info.stripe_id,
-        payment_method: payment_method.stripe_id
-      })
-
-      if payment_intent.status == "succeeded"
-        puts "Invoice[#{ubid}] charged with PaymentMethod[#{payment_method.ubid}] for $#{amount}"
-        self.status = "paid"
-        content.merge!({
-          "payment_method" => {
-            "id" => payment_method.id,
-            "stripe_id" => payment_method.stripe_id
-          },
-          "payment_intent" => payment_intent.id
+    errors = []
+    billing_info.payment_methods_dataset.order(:order).each do |pm|
+      begin
+        payment_intent = Stripe::PaymentIntent.create({
+          amount: (amount * 100).to_i, # 100 cents to charge $1.00
+          currency: "usd",
+          confirm: true,
+          off_session: true,
+          customer: billing_info.stripe_id,
+          payment_method: pm.stripe_id
         })
-        save(columns: [:status, :content])
-        return payment_intent.id
+      rescue Stripe::CardError => e
+        Clog.emit("Invoice couldn't charged.") { {invoice_not_charged: {ubid: ubid, payment_method: pm.ubid, error: e.message}} }
+        errors << e.message
+        next
       end
 
-      puts "Invoice[#{ubid}] couldn't charge with PaymentMethod[#{payment_method.ubid}]: #{payment_intent.status}"
+      unless payment_intent.status == "succeeded"
+        Clog.emit("BUG: payment intent should succeed here") { {invoice_not_charged: {ubid: ubid, payment_method: pm.ubid, intent_id: payment_intent.id, error: payment_intent.status}} }
+        next
+      end
+
+      Clog.emit("Invoice charged.") { {invoice_charged: {ubid: ubid, payment_method: pm.ubid, cost: amount}} }
+      self.status = "paid"
+      content.merge!({
+        "payment_method" => {
+          "id" => pm.id,
+          "stripe_id" => pm.stripe_id
+        },
+        "payment_intent" => payment_intent.id
+      })
+      save(columns: [:status, :content])
+      return true
     end
 
-    puts "Invoice[#{ubid}] couldn't charge with any payment method"
+    Clog.emit("Invoice couldn't charged with any payment method.") { {invoice_not_charged: {ubid: ubid}} }
+    false
   end
 end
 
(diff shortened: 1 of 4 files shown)


model/invoice.rb
```diff
@@ -76,9 +79,26 @@ def charge
...
+  def send_success_email(below_threshold: false)
+    ser = Serializers::Web::Invoice.new(:detailed).serialize(self)
+    message = if below_threshold
+      "Since the invoice total of #{ser[:total]} is below our minimum charge threshold, there will be no charges for this month."
+    else
+      "The invoice amount of #{ser[:total]} will be debited from your credit card on file."
+    end
+    Util.send_email(ser[:billing_email], "Ubicloud #{ser[:name]} Invoice ##{ser[:invoice_number]}",
+      greeting: "Dear #{ser[:billing_name]},",
+      body: ["Please find your current invoice ##{ser[:invoice_number]} at the link.",
```
byucesoy, line 97: How difficult is it to make this attachment?

serializers/web/invoice.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "countries"
```
byucesoy, line 3: Why do we need this?

model/invoice.rb
```diff
@@ -15,58 +15,69 @@ def name
...
-      puts "Invoice[#{ubid}] cost is less than minimum charge cost: $#{content["cost"]}"
-      return
+      Clog.emit("Invoice cost is less than minimum charge cost.") { {invoice_below_threshold: {ubid: ubid, cost: amount}} }
+      return true
     end
 
-    unless (billing_info = BillingInfo[content.dig("billing_info", "id")])
-      puts "Invoice[#{ubid}] doesn't have billing info"
-      return
+    unless (billing_info = BillingInfo[content.dig("billing_info", "id")]) && billing_info.payment_methods.count > 0
```
byucesoy, line 39: It is confusing to read unless statements with multiple conditions. How about;

```
if (billing_info = BillingInfo[content.dig("billing_info", "id")]).nil? || billing_info.payment_methods.count == 0
```

model/invoice.rb
```diff
@@ -15,58 +15,69 @@ def name
...
+        break
+      else
+        Clog.emit("Invoice couldn't charged.") { {invoice_not_charged: {ubid: ubid, payment_method: pm.ubid, error: payment_intent.status}} }
+        errors << "Unexpected payment status: #{payment_intent.status}"
       end
-
-      puts "Invoice[#{ubid}] couldn't charge with PaymentMethod[#{payment_method.ubid}]: #{payment_intent.status}"
+    rescue Stripe::CardError => e
+      Clog.emit("Invoice couldn't charged.") { {invoice_not_charged: {ubid: ubid, payment_method: pm.ubid, error: e.message}} }
+      errors << e.message
```
byucesoy, line 75: It seems we push things `errors` but never read from it. Do we need it?

model/invoice.rb
```diff
@@ -15,58 +15,69 @@ def name
...
-        return payment_intent.id
+        success = true
+        break
+      else
+        Clog.emit("Invoice couldn't charged.") { {invoice_not_charged: {ubid: ubid, payment_method: pm.ubid, error: payment_intent.status}} }
+        errors << "Unexpected payment status: #{payment_intent.status}"
       end
-
-      puts "Invoice[#{ubid}] couldn't charge with PaymentMethod[#{payment_method.ubid}]: #{payment_intent.status}"
+    rescue Stripe::CardError => e
```
byucesoy, line 73: Instead of rescueing big piece of code, can we put this just after the `PaymentIntent.create`call?

model/invoice.rb
```diff
@@ -15,58 +15,69 @@ def name
...
-            "stripe_id" => payment_method.stripe_id
+            "id" => pm.id,
+            "stripe_id" => pm.stripe_id
           },
           "payment_intent" => payment_intent.id
         })
         save(columns: [:status, :content])
-        return payment_intent.id
+        success = true
+        break
```
byucesoy, line 68: Instead of break, you can return `true` here. In that way, you wouldn't need to keep a flag for `success`

model/invoice.rb
byucesoy: And I think we should distinguish this type of error from the other one. In the other case, we cannot charge the invoice because something is wrong with the payment information (i.e. most likely user error). In this case, something is wrong either in our code or in Stripe APIs. So we shouldn't do certain things we would do such as sending service termination mail.

I'm not entirely sure, but maybe this function should return 3 state enum (success, system_failure, payment_method_failure) ?

byucesoy (commented):
Thanks a lot for separate commits. It made reviewing significantly easier.

model/invoice.rb
enescakir: We have several options, but haven't decided what to do yet. I didn't want to spend time on this feature for now. After this PR, only the attachment part remains.
**Possible solutions:**
1. Add a new dependency such as `wkhtmltopdf` and convert our HTML invoices to PDF and send it. We are somewhat conservative about adding new dependencies to our app hosts.
2. (Daniel said it's what Heroku does) Send invoice as an HTML file. We already have an HTML invoice, we can send it. I'm unsure if customers will appreciate HTML invoices
3. Use a pure Ruby library to generate PDF invoices. So we will not add a new dependency, only a new Ruby package

model/invoice.rb
enescakir: We use it in the next commit. The failed payment notification shows the error messages.

serializers/web/invoice.rb
enescakir: We use country codes from it in this serializer. https://github.com/ubicloud/ubicloud/blob/4408a5184298c3f215849a75c8a9fa2d43c38e47/serializers/web/invoice.rb#L43

The reason it worked before is that it was imported at the billing route, and the invoice serializers were only used there.
https://github.com/ubicloud/ubicloud/blob/4408a5184298c3f215849a75c8a9fa2d43c38e47/routes/web/project/billing.rb#L4

model/invoice.rb
enescakir: I changed it

model/invoice.rb
enescakir: Thanks. Makes sense

model/invoice.rb
byucesoy: 3

model/invoice.rb
```diff
@@ -15,58 +15,67 @@ def name
...
+          confirm: true,
+          off_session: true,
+          customer: billing_info.stripe_id,
+          payment_method: pm.stripe_id
         })
-        save(columns: [:status, :content])
-        return payment_intent.id
+      rescue Stripe::CardError => e
+        Clog.emit("Invoice couldn't charged.") { {invoice_not_charged: {ubid: ubid, payment_method: pm.ubid, error: e.message}} }
+        errors << e.message
```
byucesoy, line 57: Is this system error or user error. If it is system, then maybe we shouldn't put it to the errors array that we eventually send to user.

byucesoy, line 57: Is this system error or user error. If it is system, then maybe we shouldn't put it to the errors array that we eventually send to user.

byucesoy, line 57: Is this system error or user error. If it is system, then maybe we shouldn't put it to the errors array that we eventually send to user.

model/invoice.rb
```diff
@@ -15,58 +15,67 @@ def name
...
+        next
       end
 
-      puts "Invoice[#{ubid}] couldn't charge with PaymentMethod[#{payment_method.ubid}]: #{payment_intent.status}"
+      unless payment_intent.status == "succeeded"
+        Clog.emit("Invoice couldn't charged.") { {invoice_not_charged: {ubid: ubid, payment_method: pm.ubid, error: payment_intent.status}} }
+        errors << "Unexpected payment status: #{payment_intent.status}"
```
byucesoy, line 63: Is there any extra information we can share with the user? Such as error message or which payment method failed.

model/invoice.rb
enescakir: I made rescue block smaller.

We haven't encountered a system error yet. We only handle `Stripe::CardError`, which likely results from user errors. I hope the Stripe SDK would throw a different exception other than `Stripe::CardError`, when there's a system failure. As we don't handle other exceptions, the operation will crash, preventing customer notifications. We can think more about system failure scenarios when we encounter an issue.

model/invoice.rb
enescakir: Ideally, the Stripe SDK should throw an exception different from `Stripe::CardError` such as `Stripe::APIError` when a system failure occurs.

https://docs.stripe.com/error-handling?lang=ruby#error-types

model/invoice.rb
enescakir: We can add the last four digit and the card brand such as Master Card, Visa etc.

enescakir: I checked the Ruby SDK once more, and we shouldn't get to this point. Their documentation also depends on the exceptions.
https://docs.stripe.com/payments/save-and-reuse?platform=web&ui=elements#charge-saved-payment-method

It should throw an exception if there are any failures. I've updated this section. It now logs a `BUG` here

byucesoy (approved):

PR #1535, opened 2024-05-02
Add accounts association to the project model

enescakir:
Earlier, I implemented `user_ids` method to retrieve the list of user ids associated with a project. This was during my early days. Now, we can use the Sequel method to obtain project users, which is more practical.

This method is also beneficial for operational processes. I often found myself using `Account[project.user_ids.first]` to access the users.

### DIFF
--- a/routes/api/project.rb
+++ b/routes/api/project.rb
@@ -36,7 +36,7 @@ class CloverApi
         r.halt
       end
 
-      unless @project.user_ids.include?(@current_user.id)
+      unless @project.accounts.any? { _1.id == @current_user.id }
         fail Authorization::Unauthorized
       end
 
--- a/model/project.rb
+++ b/model/project.rb
@@ -9,6 +9,7 @@ class Project < Sequel::Model
   one_to_many :usage_alerts
   one_to_many :github_installations
 
+  many_to_many :accounts, join_table: AccessTag.table_name, left_key: :project_id, right_key: :hyper_tag_id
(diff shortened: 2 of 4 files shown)


byucesoy (approved):
YES! :heart:

routes/api/project.rb
```diff
@@ -36,7 +36,7 @@ class CloverApi
         r.halt
       end
 
-      unless @project.user_ids.include?(@current_user.id)
+      unless @project.accounts.include?(@current_user)
```
byucesoy, line 39: `include?` or more specifically equality is not safe for object comparisons, because they compare the objects not their ids. Here is a scary example;

```ruby
project = udec "...";
vm = project.vms.first;
project.vms.include?(vm)
=> true
project.reload.vms.include?(vm)
=> false
```

So we should stick with comparing ids.

byucesoy (changes requested):

routes/api/project.rb
byucesoy: Here is a simpler example;

```ruby
Vm["b15336fb-7925-8374-962d-c1d746c7c7e1"] == Vm["b15336fb-7925-8374-962d-c1d746c7c7e1"]
=> false
```

byucesoy: Then, why these work?

```ruby
1 == 1.0
=> true
"test" == "test"
=> true
```

Because their respective classes override `==` operator to make it more useful. It seems Sequel does not do that for its models, probably rightfully so.

enescakir: I changed it

byucesoy (approved):

PR #1544, opened 2024-05-03
Do not show usage alerts when no billing info

enescakir:
Usage alerts are not effective when the project doesn't have any billing info. The customer should provide a billing info to create a resource first. The usage alerts might distract the customer from providing billing info.

Screenshot 2024-05-03 at 18 37 19


byucesoy:
Can we show at least something about existence of usage alerts. For example a message something like "To be able to add usage alerts, you first need to enter billing information", wording is up to you. I'm afraid, some people might be hesitant to add billing information and seeing existence of usage alerts might alleviate some of their concerns.

byucesoy (commented):
Can we show at least something about existence of usage alerts. For example a message saying something like "To be able to add usage alerts, you first need to enter billing information", wording is up to you. I'm afraid, some people might be hesitant to add billing information and seeing existence of usage alerts might alleviate some of their concerns.

enescakir:
We'll probably review our UX in the future

PR #1546, opened 2024-05-03
Fix display state of PostgreSQL resources

byucesoy:
We used to just check the label of PostgresResource's own strand. However, the state of representative server is also important, especially at the time of it being unavailable. This commit fixes the display state of PostgreSQL resources by checking the state of the representative server as well.

While doing so, I found another case of wrong display state, while refreshing the dns records, which is also fixed in this commit.

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -43,7 +43,8 @@ class PostgresResource < Sequel::Model
   end
 
   def display_state
-    return "running" if ["wait", "refresh_certificates"].include?(strand.label)
+    return "unavailable" if representative_server&.strand&.label == "unavailable"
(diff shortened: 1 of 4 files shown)


model/postgres/postgres_resource.rb
```diff
@@ -43,7 +43,8 @@ def hyper_tag_name(project)
   end
 
   def display_state
-    return "running" if ["wait", "refresh_certificates"].include?(strand.label)
+    return "unavailable" if representative_server&.strand&.label == "unavailable"
+    return "running" if ["wait", "refresh_certificates"].include?(strand.label) || strand.label == "refresh_dns_record" && !initial_provisioning_set?
```
bsatzger, line 47: nit, feel free to ignore:
```suggestion
    return "running" if ["wait", "refresh_certificates", "refresh_dns_record"].include?(strand.label) && !initial_provisioning_set?
```

bsatzger (approved):

PR #1547, opened 2024-05-04
Add helper method for decoding deleted record

byucesoy:
While working on DeletedRecord, we frequently need to find the record from its ubid. This commit adds a helper method for that.

### DIFF
--- a/bin/pry
+++ b/bin/pry
@@ -7,6 +7,8 @@ require_relative "../loader"
 
 require "pry"
 
+Sequel.extension :pg_json_ops
+
 def dev_project
   return unless Config.development?
   ac = Account[email: "dev@ubicloud.com"] || Account.create_with_id(email: "dev@ubicloud.com")
@@ -15,6 +17,10 @@ end
 
 def udec(*)
   UBID.decode(*)
+end
+
+def udec_deleted(ubid)
+  DeletedRecord.where(Sequel.pg_json(:model_values)["name"] => Sequel.pg_jsonb_wrap(UBID.parse(ubid).to_uuid)).first
 end
 
 opts = Pry::CLI.parse_options


bin/pry
```diff
@@ -17,6 +19,10 @@ def udec(*)
   UBID.decode(*)
 end
 
+def udec_deleted(ubid)
+  DeletedRecord.where(Sequel.pg_json(:model_values)["name"] => Sequel.pg_jsonb_wrap(UBID.parse(ubid).to_uuid)).first
+end
```
enescakir, line 24: What do you think about adding it to the `udec` as a second option?
Probably we will run `udec` first and it will return `nil`, then we will run `udec_deleted`
```suggestion
  UBID.decode(*) || DeletedRecord.where(Sequel.pg_json(:model_values)["name"] => Sequel.pg_jsonb_wrap(UBID.parse(ubid).to_uuid)).first
end
```

bin/pry
```diff
@@ -17,6 +19,10 @@ def udec(*)
   UBID.decode(*)
 end
 
+def udec_deleted(ubid)
+  DeletedRecord.where(Sequel.pg_json(:model_values)["name"] => Sequel.pg_jsonb_wrap(UBID.parse(ubid).to_uuid)).first
```
furkansahin, line 23: I think we must add a gin index for this field. Simply querying deleted records in production is almost unbearable at this point.

fdr: it's also using almost all our space. Should we consider partitioning as well? How often are we digging in the ancient past? Not often, right?

GIN is perhaps the first thing to try, though the idea of a GIN over all (ever growing) history that has to get updated with every deletion makes me a little bit suspenseful. Just a bit.

enescakir, line 23: Shouldn't we filter by `id` instead of `name`?
```suggestion
  DeletedRecord.where(Sequel.pg_json(:model_values)["id"] => Sequel.pg_jsonb_wrap(UBID.parse(ubid).to_uuid)).first
```

fdr:
Adding @jeremyevans on this old patch. things have changed quite bit ever since we did partitioning, which kinda solves many usability problems by making lots of stuff less expensive, and it seems up his alley.

bin/pry
```diff
@@ -7,6 +7,8 @@ require_relative "../loader"
 
 require "pry"
 
+Sequel.extension :pg_json_ops
```
jeremyevans, line 10: This is now unneeded as it is included db.rb.

jeremyevans (commented):
We definitely need a gin or similar index on the archived_record table before we do this.  This also needs to be updated to use ArchivedRecord instead of DeletedRecord.

byucesoy:
Not an immediate priority right now.

PR #1550, opened 2024-05-04
Allow only available postgres location while creating it on API

velioglu:
Allowing locations available for postgres only while creating a postgres database via API.


routes/api/project/location/postgres.rb
```diff
@@ -28,6 +28,7 @@ class CloverApi
       r.post true do
         Authorization.authorize(@current_user.id, "Postgres:create", @project.id)
         fail Validation::ValidationFailed.new({billing_info: "Project doesn't have valid billing information"}) unless @project.has_valid_payment_method?
+        Validation.validate_postgres_location(@location)
```
byucesoy, line 31: Should we put the same check to web endpoint as well?

furkansahin (approved):

byucesoy (approved):

PR #1551, opened 2024-05-05
Allocator improvements

bsatzger:
The allocator is the component that assigns a VM to a host, as part of VM provisioning. This commit brings the following main changes:

* Carve allocation logic out of Vm::Nexus into a separate file.
* Assign a VM to the most suitable host, instead of a random valid host. Suitability is defined by a score. An allocation score is a relative measure that allows to rank allocation candidates. An allocation with a low score is "more suitable" than an allocation with a high score.
* Allow concurrent allocations if their result is valid. Validity is enforced by DB constraints.

### DIFF
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -1,0 +1,252 @@
+# frozen_string_literal: true
+
+module Scheduling::Allocator
+  def self.allocate(vm, storage_volumes, target_host_utilization: 0.65, distinct_storage_devices: false, allocation_state_filter: ["accepting"], host_filter: [], location_filter: [], location_preference: [])
+    request = Request.new(
+      vm.id,
+      vm.cores,
+      vm.mem_gib,
+      storage_volumes.map { _1["size_gib"] }.sum,
+      storage_volumes.size.times.zip(storage_volumes).to_h.sort_by { |k, v| v["size_gib"] * -1 },
+      distinct_storage_devices,
+      vm.ip4_enabled,
+      target_host_utilization,
+      vm.arch,
+      allocation_state_filter,
+      host_filter,
+      location_filter,
+      location_preference
+    )
+    allocation = Allocation.best_allocation(request)
+    fail "#{vm} no space left on any eligible host" unless allocation
+
+    allocation.update(vm)
+    Clog.emit("vm allocated") { {allocation: allocation.to_s, duration: Time.now - vm.created_at} }
+  end
+
+  Request = Struct.new(:vm_id, :cores, :mem_gib, :storage_gib, :storage_volumes, :distinct_storage_devices, :ip4_enabled, :target_host_utilization,
+    :arch_filter, :allocation_state_filter, :host_filter, :location_filter, :location_preference)
+
+  class Allocation
+    attr_reader :score
+
+    def self.best_allocation(request)
+      candidate_hosts(request).map { Allocation.new(_1, request) }
+        .select { _1.is_valid }
+        .min_by { _1.score }
+    end
+
+    def self.candidate_hosts(request)
+      ds = DB[:vm_host]
+        .join(:storage_devices, vm_host_id: Sequel[:vm_host][:id])
+        .join(:total_ipv4, routed_to_host_id: Sequel[:vm_host][:id])
+        .join(:used_ipv4, routed_to_host_id: Sequel[:vm_host][:id])
+        .select(:vm_host_id, :total_cores, :used_cores, :total_hugepages_1g, :used_hugepages_1g, :location, :num_storage_devices, :available_storage_gib, :total_storage_gib, :storage_devices, :total_ipv4, :used_ipv4)
+        .where(allocation_state: request.allocation_state_filter)
+        .where(arch: request.arch_filter)
+        .where { (total_hugepages_1g - used_hugepages_1g >= request.mem_gib) }
+        .where { (total_cores - used_cores >= request.cores) }
+        .with(:total_ipv4, DB[:address]
+          .select_group(:routed_to_host_id)
+          .select_append { round(sum(power(2, 32 - masklen(cidr)))).cast(:integer).as(total_ipv4) }
+          .where { (family(cidr) =~ 4) })
+        .with(:used_ipv4, DB[:address].left_join(:assigned_vm_address, address_id: :id)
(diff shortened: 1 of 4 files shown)


scheduling/allocator.rb
```diff
@@ -0,0 +1,239 @@
...
+        .select_append { sum(available_storage_gib).as(available_storage_gib) }
+        .select_append { sum(total_storage_gib).as(total_storage_gib) }
+        .select_append { json_agg(json_build_object(Sequel.lit("'id'"), Sequel[:storage_device][:id], Sequel.lit("'total_storage_gib'"), total_storage_gib, Sequel.lit("'available_storage_gib'"), available_storage_gib)).order(available_storage_gib).as(storage_devices) }
+        .where(enabled: true)
+        .where(allocation_state: request.allocation_state_filter)
+        .where(arch: request.arch_filter)
+        .where { (total_hugepages_1g - used_hugepages_1g >= request.mem_gib) }
+        .where { (total_cores - used_cores >= request.cores) }
+        .having { sum(available_storage_gib) >= request.storage_gib }
+        .having { count.function.* >= (request.distinct_storage_devices ? request.storage_volumes.count : 1) }
```
furkansahin, line 51: I'm asking because this is the first time I see this usage. Once we define the `count.function.*.as(num_storage_devices)`, is there a way to use its name in the `having` clause? It was kind of unintuitive to figure this out.

scheduling/allocator.rb
```diff
@@ -0,0 +1,239 @@
...
+    def self.update_vm(vm_host, vm)
+      ip4, address = vm_host.ip4_random_vm_network if vm.ip4_enabled
+      fail "no ip4 addresses left" if vm.ip4_enabled && !ip4
+      vm.update(
+        vm_host_id: vm_host.id,
+        ephemeral_net6: vm_host.ip6_random_vm_network.to_s,
+        local_vetho_ip: vm_host.veth_pair_random_ip4_addr.to_s,
+        allocated_at: Time.now
+      )
+      AssignedVmAddress.create_with_id(dst_vm_id: vm.id, ip: ip4.to_s, address_id: address.id) if ip4
```
furkansahin, line 67: I think this is good as is, but as a note to future, we might want to move this also into the host filtering query. The logic mainly comes from the times that we did not have the guest assignable `Address` entity.

scheduling/allocator.rb
```diff
@@ -0,0 +1,239 @@
...
+      vm.sshable&.update(host: vm.ephemeral_net4 || vm.ephemeral_net6.nth(2))
```
furkansahin, line 68: For now, updating the sshable with `ephemeral_net6` is useless since the production environment does not support `IPv6`. Therefore, any service that fully depends on IPv6 will not be able to make use of `sshable`. This is not a problem of this PR but writing for information sharing purposes. You may also consider removing it, it's upto you.

furkansahin (commented):
thank you so much

scheduling/allocator.rb
```diff
@@ -0,0 +1,239 @@
+# frozen_string_literal: true
+
+module Scheduling::Allocator
+  def self.allocate(vm, storage_volumes, target_utilization: 0.65, distinct_storage_devices: false, allocation_state_filter: ["accepting"], host_filter: [], location_filter: [], location_preference: [])
```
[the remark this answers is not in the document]
bsatzger: sure! will change.

scheduling/allocator.rb
bsatzger: I don't think that's supported in SQL

```
# select id, count(*) as c from vm_host group by id having count(*) > 1;
 id | c
----+-----
(0 rows)
```

```
# select id, count(*) as c from vm_host group by id having c > 1;
ERROR:  column "c" does not exist
LINE 1: ...ect id, count(*) as c from vm_host group by id having c > 1;
```

scheduling/allocator.rb
bsatzger: I can add this. To avoid misunderstandings, what would that additional filter look like?

scheduling/allocator.rb
bsatzger: let's leave it for now then

furkansahin (approved):

enescakir (approved):
It's great to see powerful use of Sequel library 🚀 @pykello has experience with our allocator and may have some insights to share.

scheduling/allocator.rb
```diff
@@ -0,0 +1,253 @@
...
+
+    def is_valid
+      @allocations.all? { _1.is_valid }
+    end
+
+    def update(vm)
+      vm_host = VmHost[@host[:vm_host_id]]
+      DB.transaction do
+        Allocation.update_vm(vm_host, vm)
+        VmHost.dataset.where(id: @host[:vm_host_id]).update(@allocations.map { _1.update(vm, vm_host) }.reduce(&:merge))
```
pykello, line 104: I misunderstood `@allocations.map { _1.update(vm, vm_host) }.reduce(&:merge)` initially. Just by looking at how `allocation::update` is used, I thought it returns columns in the VmHost to be updated, which was not accurate since it also has some side-effects in `StorageAllocation::update`.

What do you think breaking `allocation::update` into two methods? Like:
* `update` (which creates necessary database records)
* `vm_host_changes`

I think this way it might be a bit easier to understand without looking at the definitions of methods being called.

scheduling/allocator.rb
```diff
@@ -0,0 +1,253 @@
...
+      # utilization score
+      avg_util = util.sum.fdiv(util.size)
+      utilization_score = @request.target_host_utilization - avg_util
+      utilization_score = utilization_score.abs + 1 if utilization_score < 0
+
+      # imbalance score
+      imbalance_score = util.max - util.min
+
+      # location preference
+      location_preference_score = (@request.location_preference.empty? || @request.location_preference.include?(@host[:location])) ? 0 : 10
```
pykello, line 126: A comment with ranges of each of other scores might be helpful. Then, it will be clearer that 10 is chosen in a way that we give highest priority to location preference.

bsatzger: added comments

scheduling/allocator.rb
bsatzger: split this up into two different methods

PR #1553, opened 2024-05-06
Add more logging to monitor

byucesoy:
**Don't exit if the pulse check is stuck**
Normally, if the pulse check is stuck, the monitor will exit. This is to ensure
that the pulse checking is not stuck indefinitely. It is similar to apoptosis
logic in respirate. However, it happens more frequently than expected and thus
causing stability problems. While investigating the issue, it would be better
to not exit the monitor if the pulse check is stuck. Instead, we will just log
a message and manually cleanup stuck processes for a while.

**Add more logging to monitor**
We are investigating frequent stuck pulse checks problem. This commit will help
us to understand what is going on at the time of the stuck pulse checks.

### DIFF
--- a/bin/monitor
+++ b/bin/monitor
@@ -28,6 +28,8 @@ end
 pulse_checker = Thread.new do
   loop do
     mutex.synchronize do
+      # Deduct 3 for the threads that are always running: main, resource_scanner, pulse_checker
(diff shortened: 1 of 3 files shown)


furkansahin (approved):
I understand this is temporary but I would still prefer not to deploy commented out code. Let's clean it up and 🚢

PR #1558, opened 2024-05-07
Add firewall and nic vm name to private subnet serialization

velioglu:
Improving private subnet serialization on API by adding firewall details and nic vm name.


furkansahin (approved):

serializers/api/nic.rb
```diff
@@ -7,7 +7,7 @@ def self.base(nic)
       name: nic.name,
       private_ipv4: nic.private_ipv4.network.to_s,
       private_ipv6: nic.private_ipv6.nth(2).to_s,
-      subnet_name: nic.private_subnet.name
+      vm_name: nic.vm&.name
```
enescakir, line 10: If we serialize the list of NICs at somewhere, it might good to load vm eagerly.

PR #1560, opened 2024-05-07
PCI device pass through

bsatzger:
Detach PCI devices during VM destruction:
In `destroy`, VM::Nexus disassociates all PCI devices from the VM.

Add assigned PCI devices to VM:
Pass PCI devices that are assigned to a VM model to the VM on the host as part of VM preparation.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -434,7 +434,8 @@ WHERE (SELECT max(available_storage_gib) FROM storage_device WHERE storage_devic
       "mem_gib" => vm.mem_gib,
       "ndp_needed" => host.ndp_needed,
       "storage_volumes" => storage_volumes,
-      "swap_size_bytes" => frame["swap_size_bytes"]
+      "swap_size_bytes" => frame["swap_size_bytes"],
+      "pci_devices" => vm.pci_devices.map { [_1.slot, _1.iommu_group] }
     })
 
     host.sshable.cmd("sudo -u #{q_vm} tee #{params_path.shellescape}", stdin: params_json)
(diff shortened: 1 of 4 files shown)


prog/vm/nexus.rb
```diff
@@ -594,6 +595,8 @@ def write_params_json
         used_hugepages_1g: Sequel[:used_hugepages_1g] - vm.mem_gib
       )
 
+      PciDevice.dataset.where(vm_id: vm.id).update(vm_id: nil)
```
enescakir, line 598: Is it possible to use `association_dependencies` plugin here?

https://github.com/ubicloud/ubicloud/blob/9efe7f8c3889e3f0f6f0cad1a2997005c124d805/model/vm.rb#L16

enescakir, line 598: nit:
```suggestion
      vm.pci_devices_dataset.update(vm_id: nil)
```

bsatzger: we don't destroy the pci_device, just update it, in contrast to these entities. the update we do to pci devices is similar to what we do to the vm_host or storage_devices, where we don't use `association_dependencies`, either.

rhizome/host/bin/setup-vm
```diff
@@ -39,6 +39,7 @@ begin
   ndp_needed = params.fetch("ndp_needed", false)
   storage_volumes = params.fetch("storage_volumes")
   swap_size_bytes = params["swap_size_bytes"]
+  pci_devices = params["pci_devices"] || []
```
pykello, line 42: can also be `params.fetch("pci_devices", [])`

prog/vm/nexus.rb
enescakir: `association_dependencies` also has `:nullify` option instead of `:destroy`. But yes we didn't use `:nullify` before.
https://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/AssociationDependencies.html

enescakir (approved):

furkansahin (approved):

PR #1563, opened 2024-05-08
Tag related payment methods while suspending an account

velioglu:
We are suspending an account if we think it is a fraud. With this commit we are getting all projects of which the account is member of, then tag all payment methods of those projects as fraud.

### DIFF
--- a/model/account.rb
+++ b/model/account.rb
@@ -27,5 +27,7 @@ class Account < Sequel::Model(:accounts)
   def suspend
     update(suspended_at: Time.now)
     DB[:account_active_session_keys].where(account_id: id).delete(force: true)
+
+    projects.each { _1.billing_info&.payment_methods_dataset&.update(fraud: true) }
   end
 end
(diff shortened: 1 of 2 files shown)


model/account.rb
```diff
@@ -27,5 +27,8 @@ def create_project_with_default_policy(name, policy_body: nil)
   def suspend
     update(suspended_at: Time.now)
     DB[:account_active_session_keys].where(account_id: id).delete(force: true)
+
+    associated_billing_info_ids = Project.where(id: AccessTag.where(hyper_tag_id: id).map(:project_id)).map(:billing_info_id).compact
+    PaymentMethod.where(billing_info_id: associated_billing_info_ids).each { |pm| pm.update(fraud: true) }
```
byucesoy, line 32: AFAIR, you can run `.update` on the dataset directly, no need to iterate.

enescakir (approved):

furkansahin (approved):

PR #1566, opened 2024-05-08
Enable allocation of GPUs

bsatzger:
The allocator now checks if a VM requires a GPU. If it does, it schedules it on a host with available GPUs. It randomly selects an available GPU and then assigns that GPU together with other devices in the same IOMMU group to the VM.

The allocator penalizes assigning VMs that do not require a GPU to a host that has GPUs.


furkansahin (approved):

furkansahin:
Thank you so much!! LGTM

PR #1570, opened 2024-05-09
Download custom images with fixed version

enescakir:
### Download custom images with fixed version
When we download a custom image without specifying a version, defaults to the latest one, making it difficult to identify which version is in use on the host.

Moreover, when we release a new version of an image, we don't immediately want to download the latest version. Instead, we prefer to test the new version before deploying it on the some hosts first.

Using a fixed version also simplifies the verification of SHA256 checksums.

### Add SHA256 checksums for custom images

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -128,6 +128,10 @@ module Config
   optional :ubicloud_images_blob_storage_secret_key, string, clear: true
   optional :ubicloud_images_blob_storage_certs, string
 
+  override :github_ubuntu_2204_version, "20240422.1.0", string
+  override :github_ubuntu_2004_version, "20240422.1.0", string
+  override :postgres_ubuntu_2204_version, "20240226.1.0", string
+
   # Allocator
   override :allocator_target_host_utilization, 0.55, float
   override :allocator_max_random_score, 0.1, float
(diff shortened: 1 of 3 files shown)


pykello (approved):

model/vm_host.rb
```diff
@@ -184,7 +184,7 @@ def install_rhizome(install_specs: false)
   end
 
   # Introduced for downloading a new boot image via REPL.
-  def download_boot_image(image_name, version:, custom_url: nil)
+  def download_boot_image(image_name, version: nil, custom_url: nil)
```
pykello, line 187: I prefer to not have a default for `version` & pass `nil` explicitly when we intend to, to reduce operator error of unintended unversioned image downloads.

byucesoy (approved):

model/vm_host.rb
enescakir: I did `version` mandatory again. As we can still set `version: nil`, and might need to download unversioned images, I removed `default_version = Config.send(:"#{image_name.tr("-", "_")}_version")` helper. It overwrites `nil` version anyway even we want download unversioned one. We can add it when we decide to disallow unversioned images. For now, the operator can run download image method as `vm_host.download_image("github-ubuntu-2204", Config.github_ubuntu_2204_version)`

pykello (approved):

PR #1572, opened 2024-05-09
Allocator: fix for host forcing and test improvement

bsatzger:
Fix allocator query for empty allocation state (e.g. used when forcing a host) and make allocator tests deterministic.


furkansahin (approved):
Thank you!

enescakir (approved):

PR #772, opened 2023-10-23
Moves boot image download to host_nexus for predefined boot images

furkansahin:
With this commit, the predefined boot images are downloaded in the host_nexus. The main reason I wanted to do this was that for a fresh host, we had race condition while downloading images for the first multiple concurrent VM provisioning. This way, the host comes with images and we don't need to download in conflicting ways.

PR #1578, opened 2024-05-10
Download the default boot image in host setup

pykello:
**UPDATE**: Alternatively, we can merge https://github.com/ubicloud/ubicloud/pull/1597

Since clover is going to only allocate boot images that have already been downloaded to storage volumes, we need at least one boot image when host setup finishes. This commit downloads the default boot image (configurable, currently ubuntu-jammy) when setting up the host.

In addition, we don't need image download in vm_setup. So, we remove it.

### DIFF
--- a/prog/vm/host_nexus.rb
+++ b/prog/vm/host_nexus.rb
@@ -95,7 +95,7 @@ class Prog::Vm::HostNexus < Prog::Base
       spdk_installation = vm_host.spdk_installations.first
       spdk_cores = (spdk_installation.cpu_count * vm_host.total_cores) / vm_host.total_cpus
       vm_host.update(used_cores: spdk_cores)
-      hop_prep_reboot
+      hop_download_boot_image
     end
 
     push Prog::Storage::SetupSpdk, {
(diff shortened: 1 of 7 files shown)


rhizome/host/lib/vm_setup.rb
```diff
@@ -47,7 +47,6 @@ def vp
   def prep(unix_user, public_key, nics, gua, ip4, local_ip4, boot_image, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices)
     setup_networking(false, gua, ip4, local_ip4, nics, ndp_needed, multiqueue: max_vcpus > 1)
     cloudinit(unix_user, public_key, nics, swap_size_bytes)
-    download_boot_image(boot_image)
```
enescakir, line 50: We can also deprecate `boot_image` parameter. It's not used in this file anymore. The `boot_image` info passed via `storage_params`

prog/vm/host_nexus.rb
```diff
@@ -117,11 +117,22 @@ def before_run
...
       spdk_cores = (spdk_installation.cpu_count * vm_host.total_cores) / vm_host.total_cpus
       vm_host.update(used_cores: spdk_cores)
-      hop_prep_reboot
+      hop_download_boot_image
     end
     donate
   end
 
+  label def download_boot_image
+    bud Prog::DownloadBootImage, {"image_name" => Config.default_boot_image_name, "version" => Config.default_boot_image_version}
```
enescakir, line 126: @fdr expressed concerns about automatically downloading images during development previously in the one of @furkansahin PRs.
https://github.com/ubicloud/ubicloud/pull/772#issuecomment-1780427420

enescakir, line 126: What if we dynamically determine the images needed to download to the host? It's just an idea.

An example:

```ruby
def default_boot_images
    return [] if Config.development?
    # Should we download runner images to the all hosts?
    images = ["ubuntu-jammy", "github-ubuntu-2204", "github-ubuntu-2004"]
    
    # Germany hosts should have PG image too
    images << "postgres-ubuntu-2204" if vm_host.location == "hetzner-fsn1"

    # GPU hosts should have runner image too
    images << "github-gpu-ubuntu-2204" if vm_host.pci_devices_dataset.count > 0

    images
end
```

rhizome/host/lib/vm_setup.rb
pykello: I removed it.

prog/vm/host_nexus.rb
pykello: Makes sense, but this requires https://github.com/ubicloud/ubicloud/pull/1570 to be merged first.

Also, I think we should have `ubuntu-jammy` for development environment.

pykello: What do you think about https://github.com/ubicloud/ubicloud/pull/1597?

PR #1582, opened 2024-05-11
Don't send a certificate for non-blob-storage image downloads

pykello:
Non-blob-storage images don't require a certificate. This caused problems in the E2E tests where `ubicloud_images_blob_storage_certs` wasn't set.

### DIFF
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -13,6 +13,10 @@ class Prog::DownloadBootImage < Prog::Base
     @version ||= frame.fetch("version")
   end
 
+  def download_from_blob_storage?
+    image_name.start_with?("github", "postgres")
+  end
+
   def url
     # YYY: Should we get ubuntu & almalinux urls here? Since we might start
     # putting all images into the blob storage in future, we're postponing the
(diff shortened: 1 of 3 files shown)


prog/download_boot_image.rb
```diff
@@ -72,7 +76,10 @@ def blob_storage_client
         version: version,
         sha256sum: sha256_sum
       }.to_json
-      sshable.cmd("common/bin/daemonizer 'host/bin/download-boot-image #{params_json.shellescape}' #{q_daemon_name}", stdin: Config.ubicloud_images_blob_storage_certs)
+      sshable.cmd(
+        "common/bin/daemonizer 'host/bin/download-boot-image #{params_json.shellescape}' #{q_daemon_name}",
+        stdin: download_from_blob_storage? ? Config.ubicloud_images_blob_storage_certs : nil
```
enescakir, line 81: What do you think about moving `certs` property to `params_json` and pass whole json via stdin?

prog/download_boot_image.rb
```diff
@@ -70,9 +74,13 @@ def blob_storage_client
...
         version: version,
-        sha256sum: sha256_sum
+        sha256sum: sha256_sum,
+        certs: download_from_blob_storage? ? Config.ubicloud_images_blob_storage_certs : nil
       }.to_json
-      sshable.cmd("common/bin/daemonizer 'host/bin/download-boot-image #{params_json.shellescape}' #{q_daemon_name}", stdin: Config.ubicloud_images_blob_storage_certs)
+      sshable.cmd(
+        "common/bin/daemonizer 'host/bin/download-boot-image' #{q_daemon_name}",
+        stdin: params_json
+      )
```
enescakir, line 83: nit: IMHO, it's not that long. We have the lines longer than it.
```suggestion
      sshable.cmd("common/bin/daemonizer 'host/bin/download-boot-image' #{q_daemon_name}", stdin: params_json)
```

enescakir (approved):

PR #1583, opened 2024-05-12
Mark virtual machines as tainted and pick untainted machines from the vm pool

enescakir:
### Mark virtual machines as tainted
At a certain stage in a virtual machine's lifecycle, it may become tainted with user data.

From the moment we start the hypervisor for the first time, it's uncertain when the user logged in and began processing. After this, the virtual machine is deemed tainted if it wasn't created for a VM pool. Virtual machines from the pool are considered tainted once they are selected from the pool. Customers cannot access them until they are chosen from the pool. If the virtual machine hasn't been marked as tainted before its destroy, we mark it as tainted during the destroy process.

When a virtual machine is considered tainted, it theoretically contains irreplaceable and private data.

We can develop various logics based on the taint value.

For instance, we discussed some 'recreate' logics for virtual machines. We can easily recreate untainted virtual machines, but we need to exercise caution with tainted ones.

Furthermore, the taint value can be utilized in our pool logic. We can select only untainted virtual machines from the pool.

### Use tainted value to pick a vm from the pool

We can utilize the tainted value to choose an idle virtual machine from the pool. This approach allows us to lock only the virtual machines that haven't been picked yet. Moreover, we refrain from joining with the runner table, a practice that previously led to an incident. This incident occurred when I modified the runner to not wait for the destruction of the VM, resulting in virtual machines being selected by two different runners.

We don't disassociate the virtual machines from the pool when picked because we aim to maintain a fixed size of virtual machines provisioned by the pool. We join the VM dataset with the runner table to identify idle virtual machines. However, since we lock all virtual machines in the pool while picking one, it inadvertently locks the picked virtual machine that's in use by a runner, leading to deadlocks. We shouldn't lock the virtual machines that have already been picked.

### Use provisioned at instead of display state to pick a vm from the pool

The 'display_state' function is used to show the current state of the virtual machines to the customer, and it's dynamically calculated in certain cases. The 'display' prefix implies that we shouldn't use it into our backend logics. 'provisioned_at' serves as a more reliable indicator to determine whether the virtual machine has been provisioned.

### Pick the oldest virtual machine from the pool

At present, we select a virtual machine from the pool without any specific order. By choosing the oldest virtual machine first, we ensure our pool doesn't contain outdated machines.


migrate/20240510_add_tainted_at_vm.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:vm) do
+      add_column :tainted_at, :timestamptz
```
bsatzger, line 6: tainted typically means something became bad/unusable. And maybe also that it should be destroyed. Also, I'm wondering if we should distinguish between the time we take a vm from a pool and the time it becomes available to a customer. Potentially something like this:

`allocated_from_pool_at`:  the time we took the vm from the pool
`supplied_at`: the time it became available to the user

model/vm_pool.rb
```diff
@@ -13,13 +14,9 @@ class VmPool < Sequel::Model
...
     DB.transaction do
-      # first lock the whole pool in the join table so that no other thread can
-      # pick a vm from this pool
-      vms_dataset.for_update.all
-      pick_vm_id_q = vms_dataset.left_join(:github_runner, vm_id: :id)
-        .where(Sequel[:github_runner][:vm_id] => nil, Sequel[:vm][:display_state] => "running")
-        .select(Sequel[:vm][:id])
-      Vm.where(id: pick_vm_id_q).first
+      # We lock the idle vms in the pool so that no other thread can pick the same
+      locked_vms = idle_vms_dataset.for_update.order(:created_at).all
```
furkansahin, line 18: I'm thinking if we should perform the `name` and `tainted_at` update here and `rescue` in case `.update().zero?` and recursively call pick_vm. This would also mean we will need to add another line to check if pool is empty and return `nil` in that case.

furkansahin (commented):
Overall I like the approach but maybe we can make it better?

migrate/20240510_add_tainted_at_vm.rb
enescakir: I borrowed the term 'tainted' from @fdr, who suggested to mark virtual machines. We can use more positive work. `tainted` might have little negative impact.

The time we take a vm from a pool and the time it becomes available to a customer have distinct implications for different resources. For the standard VM product, these two times are same. However, for the runner product, we use the `ready_at` column to differentiate the latter. Thus, the final product should include a column to distinguish this. If we consider the runner service as a customer for the standard VM service, we have a time when the VM becomes available to the runner (which is the same as when it is picked from the pool), and when the runner becomes available to the end customer.

model/vm_pool.rb
enescakir: Thanks for recommendation. I refactored it. We avoid locking with the new version 🎉 

I tested the new approach with the following snippet and the first pry raised an exception.

```
# pry 1st
p = VmPool.first
DB.transaction do
    vm = p.idle_vms_dataset.order(:created_at).first
    puts "picked #{vm.name}"
    sleep 20
    fail "concurrent pick" if p.idle_vms_dataset.where(id: vm.id).update(tainted_at: Time.now).zero?
    puts "selected #{vm.name}"
end


# pry 2nd
p = VmPool.first
DB.transaction do
    vm = p.idle_vms_dataset.order(:created_at).first
    puts "picked #{vm.name}"
    fail "concurrent pick" if p.idle_vms_dataset.where(id: vm.id).update(tainted_at: Time.now).zero?
    puts "selected #{vm.name}"
end
```

migrate/20240510_add_tainted_at_vm.rb
fdr: yeah, fair, taint is often with some kind of badness, though per computer science tradition badness can be pretty abstract, e.g. in garbage collectors. In our case, the badness is increasing restriction on how we can manipulate the VM, as it could plausibly have data that we absolutely must not leak between customers, and not delete in a casual manner. In this respect it has some similarity how the jargon is used in garbage collectors (or even virtual memory, "dirty pages" is a related concept), where objects are often "tainted" as somehow being more complex to collect.

We could keep this wording in that tradition or have a new one.

As an aside, something I did previously was the inverse, and not at all for wording reasons, which was more like a "purity" (or, to get outside that nomenclature, a "flexibility") flag that objects started with and then was removed the moment it was plausible a customer might have put their irreplaceable data in it.

This took the form of a `can_recreate` semaphore that relevant objects started with and would be decr'd at some point during their provisioning process, when it was plausible irreplaceable user data might have entered the system. If `recreate` was incremented and `can_recreate` was still present, then the object could be re-provisioned. This "double-trigger" style provided very precise results in not permitting mistakes, and being concurrency safe. I'm not saying it's necessarily superior to what this might do, it's more just for reflection by the author.

I suppose the nice thing about that system is that the overwhelmingly common case results in `can_recreate` disappearing, i.e. there is zero bloat in the common case, provided one things that knowing the far-distant `tainted_at` time is extraneous...which it may not be

fdr: thinking it over, how many times would we reference this symbol? maybe three, no more than six times? it is okay to avoid a jargon at expense of length. `has_customer_data` or something like that might work.

prog/vm/nexus.rb
```diff
@@ -176,6 +176,7 @@ def clear_stack_storage_volumes
   def before_run
     when_destroy_set? do
       if strand.label != "destroy"
+        vm.update(tainted_at: Time.now) unless vm.tainted_at
```
fdr, line 179: I suppose one advantage of a "purity flag removal" idea is that the code to remove it is branchless, i.e. multiple unguarded updates do not mislead.

But a way to mitigate this would be to put the branch into a special method.

prog/vm/nexus.rb
```diff
@@ -291,6 +292,7 @@ def write_params_json
 
   label def run
     host.sshable.cmd("sudo systemctl start #{q_vm}")
+    vm.update(tainted_at: Time.now) unless vm.pool_id
```
fdr, line 295: You probably need this in the prior label, since you have to consider that `start` runs but the database cannot commit. At that point, I presume in principle irreplacable/sensitive user data may be present.

fdr: ...or perhaps you don't.  I guess pooled stuff is special. But fair question if that's the case: could tainting happen even later?

model/vm_pool.rb
```diff
@@ -5,21 +5,22 @@
...
+  def pick_vm(new_name)
+    vm = idle_vms_dataset.order(Sequel.lit("RANDOM()")).first
+    return if vm.nil?
+
+    # If we can successfully update the VM, we can be certain that no other
+    # thread has selected it. If the update fails, we continue to pick different
+    # VMs until we find an idle one or exhaust the pool of runners.
+    return pick_vm(new_name) if idle_vms_dataset.where(id: vm.id).update(name: new_name, tainted_at: Time.now).zero?
+
+    vm
```
fdr, line 24: Something about this code seems a little weird, it's odd to:

1. Run a query that qualifies a tuple
2. Run it again, but this time with update

Particularly when there's not much going on between steps (1) and (2).

How about `update` with `returning`? You can use `Model#call` to make model objects with any hash.

model/vm_pool.rb
```diff
@@ -5,7 +5,7 @@
 class VmPool < Sequel::Model
   one_to_one :strand, key: :id
   one_to_many :vms, key: :pool_id
-  one_to_many :idle_vms, key: :pool_id, class: Vm, conditions: {tainted_at: nil, display_state: "running"}
+  one_to_many :idle_vms, key: :pool_id, class: Vm, conditions: Sequel.&({tainted_at: nil}, Sequel.~(provisioned_at: nil))
```
fdr, line 8: This is a Good Change and commit message.

model/vm_pool.rb
enescakir: The challenge is updating a single idle vm without an ID for filtering in the update query. Using `limit` with `update` results in `Sequel::InvalidOperation: Dataset#update not supported on datasets with limits or offsets`.
A potential solution is to incorporate the first query as a subquery in the second one.

An example:
```
p.idle_vms_dataset
    .where(id: p.idle_vms_dataset.order(Sequel.lit("RANDOM()")).limit(1).select(:id))
    .update(name: "new-name", tainted_at: Time.now)
```

model/vm_pool.rb
enescakir: Thanks 🚀

model/vm_pool.rb
bsatzger: you could also use a query with a CTE, in SQL it'd look like:

```sql
with candidate as (
  select id from vm where ... limit 1
)
update vm set ... from candidate where vm.id = candidate.id
returning ...
```

prog/vm/nexus.rb
enescakir: I may refactor it to `vm.this.where(tainted_at: nil).update(tainted_at: Time.now)` to make it branchless, if you believe this is not an anti-pattern

prog/vm/nexus.rb
enescakir: I moved it to the previous label.

migrate/20240510_add_tainted_at_vm.rb
bsatzger: I'm personally ok if you decide to go with "tainted". Just wanted to bring this up.

bsatzger (approved):

migrate/20240510_add_tainted_at_vm.rb
byucesoy: IMO, both are fine, but I'd prefer something similar to `has_customer_data`, which requires less thinking about its function if you don't have the context.

model/vm_pool.rb
```diff
@@ -5,21 +5,22 @@
...
-      Vm.where(id: pick_vm_id_q).first
-    end
+  def pick_vm(new_name)
+    vm_hash = DB[:vm]
+      .returning
+      .with(:candidate, idle_vms_dataset.limit(1))
+      .from(:vm, :candidate)
+      .where(Sequel[:vm][:id] => Sequel[:candidate][:id])
+      .update(name: new_name, tainted_at: Time.now)
+      .first
```
byucesoy, line 22: Personally, I am having difficulty to visualize what kind of SQL query this would generate when `returning`, `with`, `update` and `first` are combined, hence I'm not confident on its correctness. It feels like if we plainly write the SQL query here, it would be easier to understand and reason about it.

BTW, do we have a convention about whether we prefer Sequel or SQL for non-trivial stuff? Personally I use both without much thinking, so I don't have a strong opinion.

prog/vm/nexus.rb
```diff
@@ -152,6 +152,7 @@ def before_run
       if strand.label != "destroy"
         vm.active_billing_records.each(&:finalize)
         vm.assigned_vm_address&.active_billing_record&.finalize
+        vm.update(has_customer_data: true)
```
furkansahin, line 155: Well, this is a little bit weird to mark it at the destroy time. Why is this necessary?

furkansahin (commented):
This whole PR and its content seems a little bit disconnected to me. We are adding a new feature to mark a VM "tainted" by setting "has_customer_data", however, we set it to true the moment we pick the VM from the pool or when we are destroying it. Essentially, the feature at hand simply marks the VM as picked or not, it has nothing to do with the customer data or being tainted.

I think the PR description and commit messages should be changed in a way that this simply is a vm_pool feature as it is. We can simply add a column to the VM as "picked" to avoid confusion. 

Another alternative is to create a join table between vm and vm_pool, and adding the `picked` column there and performing the `select for update` on that table..

prog/vm/nexus.rb
enescakir: We don't want to pick a vm which is already deleting

enescakir: Yeah, it looks weird changed it from tainted to has_customer_data

enescakir:
> This whole PR and its content seems a little bit disconnected to me. We are adding a new feature to mark a VM "tainted" by setting "has_customer_data", however, we set it to true the moment we pick the VM from the pool or when we are destroying it. Essentially, the feature at hand simply marks the VM as picked or not, it has nothing to do with the customer data or being tainted.
> 
> I think the PR description and commit messages should be changed in a way that this simply is a vm_pool feature as it is. We can simply add a column to the VM as "picked" to avoid confusion.
> 
> Another alternative is to create a join table between vm and vm_pool, and adding the `picked` column there and performing the `select for update` on that table..

Yes, it appears that this PR didn't achieve the desired state and ended up in a confusing state. The `has_customer_data` change further complicates some parts. I'm somewhat hesitant to use `select for update` on high-frequency code paths. At present, 42% of our total database execution time is consumed by this query. A separate table might reduce it, but we'd still retain the lock until the end of the transaction, causing some VMs to lock more.

What are your thoughts on renaming this column to `picked_from_pool`, `used`, `tainted`, or any other suggestions? Then do you think updating commit messages to reflect this new name could work? cc @fdr and @bsatzger for their input on naming suggestions

model/vm_pool.rb
fdr: Personally I use either. Evans, for his part, always uses the DSL. 

There are some interesting composition problems with using both that I just not thought about. One particularly ugly one is how datasets with raw SQL will silently accept, but not do anything with, additional dataset methods. There's a whole plugin for this:

https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/implicit_subquery_rb.html

I thought about applying it, but decided the bloating was not worth it...at the time.

Consider this:

```
[4] clover-development(main)> DB["select * from pg_class"].where { oid > 7 }.sql
=> "select * from pg_class"
```

I'd actually want a plugin that errored whenever such a ignored composition was done...I will write that into the mailing list.

fdr:
I didn't really intend tainting (or has_user_data) to be exclusively pool-oriented, though. I also thought it'd be useful for safety in VM recreate, which is something we're missing right now (...right?). It's also useful for operations sometimes: if the process is stuck, a quick look at that bit figures out how flexible our actions can be.

fdr:
I also was a bit puzzled why the query ended up like it did. It's not impermissible, but doesn't a straight join (that is, an inner join with conditions) do the job?

fdr:
so, in light of our recent incident, https://status.ubicloud.com/incidents/57rzzq3tglsy, could we have done better with a generic "recreate" feature capable of safely tearing down and re-allocating VMs when they're stuck? This was a standard mitigating feature of the previous control planes. I kinda think it would have helped us here. Maybe while the memories are still sharp for this incident, we can re-evaluate this patch.

PR #1586, opened 2024-05-13
Change index pages to use resource name to show

furkansahin:
## Change private subnets page to use name to show instead of Show button 
Having a separate Show button is meaningless when customers can click on
the name of the subnet. This is also more consistent with the Firewalls.

## Change PostgreSQL page to use name to show instead of Show button

### DIFF
--- a/views/postgres/index.erb
+++ b/views/postgres/index.erb
@@ -29,21 +29,19 @@
             <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">Name</th>
             <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Location</th>
             <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">State</th>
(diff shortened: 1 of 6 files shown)


views/postgres/index.erb
```diff
@@ -37,13 +37,16 @@
...
+                  <%= pg[:name] %>
+                </a>
+              </td>
               <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500"><%= pg[:location] %></td>
               <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
                 <%== render("components/pg_state_label", locals: { state: pg[:state] }) %>
               </td>
               <td class="relative whitespace-nowrap py-4 pl-3 pr-4 text-right text-sm font-medium sm:pr-6">
-                <a href="<%= @project_data[:path] %><%= pg[:path] %>" class="text-orange-600 hover:text-orange-700">Show</a>
               </td>
```
enescakir, line 50: Since the column is empty, you can remove this `<td></td>` tag. You also need to remove it from the header row.

furkansahin: missed it, thanks!

PR #1589, opened 2024-05-13
Program to remove a boot image.

pykello:
Once we install a newer version of an image, we might want to remove the older versions. This program enables that.

To use it, an operator is expect to run the following in the REPL:

```
> boot_image = BootImage[... id ...]
> st = boot_image.remove_boot_image
```

Then it'll wait for storage volumes using that boot image to drain and after that it'll remove the image files and db records.

### DIFF
--- a/migrate/20240516_boot_image_size_gib.rb
+++ b/migrate/20240516_boot_image_size_gib.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:boot_image) do
+      add_column :size_gib, Integer, null: true
+    end
+
+    # Update existing boot images
+    run <<-SQL
+    UPDATE boot_image
+    SET size_gib = CASE
+        WHEN boot_image.name = 'ubuntu-jammy' THEN 3
+        WHEN boot_image.name = 'almalinux-9.1' THEN 10
+        WHEN vm_host.arch = 'x64' AND boot_image.name LIKE 'github-ubuntu-%' THEN 75
+        WHEN vm_host.arch = 'arm64' AND boot_image.name LIKE 'github-ubuntu-%' THEN 50
+        WHEN boot_image.name = 'github-gpu-ubuntu-2204' THEN 31
+        WHEN boot_image.name LIKE 'postgres-%' THEN 8
+        ELSE 0
+    END
+    FROM vm_host
+    WHERE boot_image.vm_host_id = vm_host.id;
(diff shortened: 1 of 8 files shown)


prog/remove_boot_image.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  label def remove
+    # YYY: version will be enforced in future.
+    image_path =
+      boot_image.version ?
+          "/var/storage/images/#{boot_image.name}-#{boot_image.version}.raw" :
+          "/var/storage/images/#{boot_image.name}.raw"
+
+    boot_image.vm_host.sshable.cmd("sudo rm -rf #{image_path.shellescape}")
```
enescakir, line 23: When we download a new boot image, we decreases the available_storage_gib. Should we increase it when we remove the image?

migrate/20240516_boot_image_size_gib.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:boot_image) do
+      add_column :size_gib, Integer, default: 0
```
enescakir: Why don't we set `default: nil` initially? I didn't get why do we set it as `0`?

migrate/20240516_boot_image_size_gib.rb
```diff
@@ -3,7 +3,7 @@
 Sequel.migration do
   change do
     alter_table(:boot_image) do
-      add_column :size_gib, Integer, default: 0
+      add_column :size_gib, Integer, default: 0, null: false
```
enescakir, line 6: Should we move it to the first commit? Why does it change in the second commit?

prog/remove_boot_image.rb
```diff
@@ -0,0 +1,35 @@
...
+    nap 30 if boot_image.vm_storage_volumes.count > 0
+    hop_remove
+  end
+
+  label def remove
+    # YYY: version will be enforced in future.
+    image_path =
+      boot_image.version ?
+          "/var/storage/images/#{boot_image.name}-#{boot_image.version}.raw" :
+          "/var/storage/images/#{boot_image.name}.raw"
```
enescakir, line 21: The `DownloadBootImage` also has the same path check. Should we consider moving it to the `BootImage` model as a helper, like `boot_image.file_path`, `boot_image.image_path`?

enescakir (approved):
LGTM

PR #1595, opened 2024-05-14
Refactor IPv6 generator to support hosts with smaller than prefix 64

furkansahin:
Some providers place multiple hosts behind the same /64. Therefore, the
gateway is the same for both servers. We then try to use this whole /64
in Ubicloud and carve out /80 addresses to assign to VMs. However, when
a single server claims the full /64, we cannot add the second host. Here
in this PR we solve this problem by supporting subnets that are smaller
than /64. This way, when there are 2 servers behind the same subnet, we
can selectively assign a (let's say) /68 subnet to one and another /68
to another. After that, we can still carve out /80 addresses from those.

One important change we are making is in the ip6_random_vm_network. That
is needed because, the rounding up causes the `fail` line to be
triggered. Here is a scenario;
- The host_prefix is 68.
- The prefix is 79
With the rounding up, we produce a lower_bit that is 2 bytes. These 2
bytes before the /79, doesn't fit together with /68 prefix, we may reach
an address that is not directly a child address of the host_prefix.
Therefore, we keep that at minimum 1 and produce a number that is a
single byte which easily fits under the /68 prefix.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -73,16 +73,16 @@ class VmHost < Sequel::Model
   # Generate a random network that is a slice of the host's network
   # for delegation to a VM.
   def ip6_random_vm_network(prefix = 79)
-    subnet_bits = prefix - host_prefix
-
-    # Perform integer division, rounding up the number of random bytes
-    # needed.
-    bytes_needed = ((subnet_bits - 1) / 8) + 1
-
-    # Generate bits to sit between the host_prefix and the vm network.
-    #
-    # Shift them into the right place for a 128 bit IPv6 address
-    lower_bits = SecureRandom.bytes(bytes_needed).unpack1("n") << (128 - prefix - 1)
+    # Generates bits worth of a single byte of entropy and then shifts it to the
+    # correct position in the host's network. The bits sits in between the
+    # host's network and the network that is delegated to the VM. We use the
+    # range 2...256 to avoid the case where the lower_bits is 0 or 1, which
+    # would result in the host's network being returned. We also make sure the
+    # higest possible value is 255, so that the lower_bits is never greater than
+    # the host's network. This would be the case if the difference between the
+    # host's network and the network that is delegated to the VM is a single
+    # byte.
+    lower_bits = SecureRandom.random_number(2...256) << (128 - prefix - 1)
 
     # Combine it with the higher bits for the host.
     proposal = NetAddr::IPv6Net.new(
--- a/prog/learn_network.rb
+++ b/prog/learn_network.rb
@@ -34,7 +34,7 @@ class Prog::LearnNetwork < Prog::Base
     case JSON.parse(s)
     in [iface]
       case iface.fetch("addr_info").filter_map { |info|
-             if (local = info["local"]) && (prefixlen = info["prefixlen"]) && prefixlen <= 64
+             if (local = info["local"]) && (prefixlen = info["prefixlen"]) && prefixlen <= 68
                Ip6.new(local, prefixlen)
              end
            }
--- a/spec/model/vm_host_spec.rb
+++ b/spec/model/vm_host_spec.rb
@@ -59,8 +59,8 @@ RSpec.describe VmHost do
   end
 
   it "tries to get another random network if the proposal matches the reserved nework" do
-    expect(SecureRandom).to receive(:bytes).and_return("\0\0")
-    expect(SecureRandom).to receive(:bytes).and_call_original
+    expect(SecureRandom).to receive(:random_number).and_return(0)
+    expect(SecureRandom).to receive(:random_number).and_call_original
     expect(vh.ip6_random_vm_network.to_s).not_to eq(vh.ip6_reserved_network)
   end
 


model/vm_host.rb
```diff
@@ -72,17 +72,16 @@ def ip6_reserved_network(prefix = 79)
 
   # Generate a random network that is a slice of the host's network
   # for delegation to a VM.
-  def ip6_random_vm_network(prefix = 79)
-    subnet_bits = prefix - host_prefix
```
fdr, line 76: Do we always want fifteen bits of entropy, or perhaps use fewer bits, e.g. 12 to keep the customer subnet size the same?

I kind of think the latter would be better...

I also am a bit confused as to how this code matches the commit message. It makes reference to maintaining a /80 for customers, but it looks like here the prefix would always be fifteen bits smaller than the host prefix, so how is that maintained?

furkansahin: I'm having trouble getting this out. My problem is, if the provided 2 hosts have, let's say "2a01:4f8:10a:128b::/64" as the subnet network address. I want to configure them in this way;
1. First host: "2a01:4f8:10a:128b:1::/68"
2. Second host: "2a01:4f8:10a:128b:2::/68"

The above is not really possible with /68 subnet. Because, the network address of the /68 subnet will be the same as its /64 parent. The next hextet start at /80 which we give to the customer. Here in this PR, I tried to find a solution but failed. If we start the host with /80, we cannot allocate /80 to the customers. Is it important to be able to allocate? I'm not sure, because even if we allocate /80, today, they get /128 in their VM because of the dnsmasq DHCP limitations. (prefix delegation :/ )

Any suggestions?

fdr: Can it be divided in half? Instead of using ::1 and ::2, you'd want more like `128b::2` and `128c::2` or something of the sort (possibly adding more bits), rather mixing things together. It seems like doing things with overlaps like that would also make routing really hard. Having little pinholes go to one host or another is somewhat anti-routing (which likes prefixes the most)

One thing I find puzzling and never resolved is how Comcast gives my laptop a /64: I presumed this was prefix delegation but now I'm not sure given our experience earlier. I never did anything special either. It was this ignorance that made me think it would have come together more naturally than things did. Perhaps I should break open tcpdump again and try to resolve what my client software is doing and the router is saying to me.

furkansahin: Alright, I think I figured this out. So, we would divide it to 16 for /68. The child subnets would look like;
1. 2a01:4f8:10a:128b:0000::2/68
2. 2a01:4f8:10a:128b:1000::2/68
3. 2a01:4f8:10a:128b:2000::2/68
4. 2a01:4f8:10a:128b:3000::2/68
5. ....
16. 2a01:4f8:10a:128b:f000::2/68

This way, we would still be able to allocate /80 that belongs to that specific prefix in that specific host. Of course, the prefix will be set manually in case we get hosts behind the same /64 prefix.

fdr: That looks pretty good. Maybe back it off to a size where we can fit 42 computers, so, steal six bits. A standard rack is 42U, is why. 

And indeed comcast uses slaac in addition to a dhcp /128, and after doing it I remember having done this before and concluding we had nothing to do until https://datatracker.ietf.org/doc/draft-mishra-6man-variable-slaac/ gets anywhere (at present, it don't look good)

model/vm_host.rb
```diff
@@ -75,14 +75,13 @@ def ip6_reserved_network(prefix = 79)
   def ip6_random_vm_network(prefix = 79)
     subnet_bits = prefix - host_prefix
 
-    # Perform integer division, rounding up the number of random bytes
-    # needed.
-    bytes_needed = ((subnet_bits - 1) / 8) + 1
+    # Perform integer division
```
fdr, line 78:
```suggestion
```

fdr (commented):
I'm a bit puzzled by this patch. It appears it changes how many random bits are used in all cases, e.g. a reduction in precision from `n` to `C` that is unconditional. Is that what we want? 

Also, is `bytes_needed = [((subnet_bits - 1) / 8), 1].max` fixing a bug?

also, does this patch more or less do everything necessary to support smaller prefixes?

furkansahin:
So, when we go to the longer prefixes landscape, we're talking about > 64 and < 80. If we have the example of 68, 
```
bytes_needed = ((subnet_bits - 1) / 8) + 1
```
would return 2 bytes which is longer than we needed, and in return we hit the line for most of the proposals;
```
fail "BUG: host should be supernet of randomized subnet" unless net6.rel(proposal) == 1
```

By adding `[].max`, we also make sure we won't get `nil` from `SecureRandom.bytes(bytes_needed).unpack1("C")` for cases `bytes_needed=0`.

So, without the changes in this PR, we would either hit `nil ref` in line 84 or hit the `fail` in line 92 for hosts with longer prefixes.

furkansahin:
@fdr waiting for your input here

fdr (approved):
I would rewrite the commit message, but I'm out of time. I think it is somehow repetitive and could be condensed. 

And, it omits that ALL random subnets, even for hosts with shorter prefixes, will use fewer random bits as a side effect.

Secondarily, that the previous code would crash if host_prefix was too large, rounding the number of random bytes generated to zero.

But it's better to commit in a timely manner than not, so, here's the approval bit.

fdr:
Actually, now that I think of it, why even bother with the bytes needed with the "C" unpack? Can you avoid all that and generate it byte and the code can stop pretending it handles more or less random bytes?

furkansahin:
I have updated the commit message, thanks!

PR #1597, opened 2024-05-14
Remove implicit download in vm_setup, and add the option to download default image in HostNexus

pykello:
### Remove image download from vm_setup
DownloadBootImage is now how we manage image downloads, so we can remove implicit image download from here.

See https://github.com/ubicloud/ubicloud/pull/1526

### Add the option to download the default images on host setup
This is mostly a convenience for developers who always provision VMs after setting up a host. Without this, they'll need to manually do an image download after host setup is done. To use this, one needs to do:

```
> HostNexus.assemble(...., default_boot_images: ["ubuntu-jammy", "github-ubuntu-2204"])
```

### DIFF
--- a/prog/vm/host_nexus.rb
+++ b/prog/vm/host_nexus.rb
@@ -4,7 +4,7 @@ class Prog::Vm::HostNexus < Prog::Base
   subject_is :sshable, :vm_host
   semaphore :checkup, :reboot, :destroy
 
-  def self.assemble(sshable_hostname, location: "hetzner-hel1", net6: nil, ndp_needed: false, provider: nil, hetzner_server_identifier: nil, spdk_version: Config.spdk_version)
+  def self.assemble(sshable_hostname, location: "hetzner-hel1", net6: nil, ndp_needed: false, provider: nil, hetzner_server_identifier: nil, spdk_version: Config.spdk_version, default_boot_images: [])
     DB.transaction do
       ubid = VmHost.generate_ubid
 
@@ -23,7 +23,7 @@ class Prog::Vm::HostNexus < Prog::Base
       Strand.create(
         prog: "Vm::HostNexus",
         label: "start",
-        stack: [{"spdk_version" => spdk_version}]
+        stack: [{"spdk_version" => spdk_version, "default_boot_images" => default_boot_images}]
       ) { _1.id = vmh.id }
     end
   end
@@ -95,7 +95,8 @@ class Prog::Vm::HostNexus < Prog::Base
       spdk_installation = vm_host.spdk_installations.first
       spdk_cores = (spdk_installation.cpu_count * vm_host.total_cores) / vm_host.total_cpus
       vm_host.update(used_cores: spdk_cores)
-      hop_prep_reboot
+
+      hop_download_boot_images
     end
 
     push Prog::Storage::SetupSpdk, {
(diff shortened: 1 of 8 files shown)


prog/vm/host_nexus.rb
```diff
@@ -105,6 +110,33 @@ def before_run
...
+    when "postgres-ubuntu-2204"
+      Config.postgres_ubuntu_2204_version
+    else
+      fail "Unknown boot image: #{image_name}"
+    end
+  end
+
+  label def download_boot_image
+    # check if download is complete
+    hop_prep_reboot if /image=.* version=.* downloaded/.match?(retval&.dig("msg"))
```
fdr, line 132: retval can contain structured data, being a full JSON, I'm a bit puzzled why we have a structured format in a structured format, only to have to re-parse it. But perhaps that's merely path dependency.

rhizome/host/bin/setup-vm
```diff
@@ -32,7 +32,6 @@ begin
   unix_user = params.fetch("unix_user")
   ssh_public_key = params.fetch("ssh_public_key")
   nics = params.fetch("nics").map { |args| VmSetup::Nic.new(*args) }.freeze
-  boot_image = params.fetch("boot_image")
```
fdr, line 35: this commit message would benefit backreference to the commit that has rendered it obsolete. On its own, it is rather mysterious: a deletion of code that looks fundamental without replacement that is immediate adjacent to it in commit history (the latter being a special case where I think backreferences can be burdensome, because trival rebasing breaks the backreference).

 > Clover needs to track images available on each host, so we shouldn't download images implicitly in vm_setup.

I don't understand what's mutually exclusive between implied downloads and tracking images that are available; in principle the system might have been designed to read-back the result of the implied image download to update the `BootImage` records.

It may suffice to write that "DownloadBootImage" is how images are managed now because [commit X Y Z], without implying some intrinsic connection between laziness of download and the new regime.

fdr (approved):

prog/vm/host_nexus.rb
fdr: I can prepare a patch to rectify this after the fact though, unless you wanted to throw it in.

```diff
diff --git a/prog/download_boot_image.rb b/prog/download_boot_image.rb
index 849004f4..a28c0f1d 100644
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -112,6 +112,6 @@ def blob_storage_client
       name: image_name,
       version: version
     ).update(activated_at: Time.now)
-    pop "image=#{image_name} version=#{version} downloaded"
+    pop(msg: "image downloaded", image: image_name, version: version)
   end
 end
```

and running the tests suggests that the interface width is not so large...

fdr: ...though I guess if the interface is _so_ thin I'm wondering why retval includes dynamic content.

prog/vm/host_nexus.rb
```diff
@@ -105,6 +110,33 @@ def before_run
...
+      Config.github_ubuntu_2204_version
+    when "github-ubuntu-2004"
+      Config.github_ubuntu_2004_version
+    when "github-gpu-ubuntu-2204"
+      Config.github_gpu_ubuntu_2204_version
+    when "postgres-ubuntu-2204"
+      Config.postgres_ubuntu_2204_version
+    else
+      fail "Unknown boot image: #{image_name}"
+    end
```
enescakir, line 127:
```suggestion
    fail "Unknown boot image: #{image_name}" unless Config.respond_to?(:"#{image_name.tr("-", "_")}_version")
    Config.send(:"#{image_name.tr("-", "_")}_version") : nil
```

prog/vm/host_nexus.rb
```diff
@@ -23,7 +23,7 @@ def self.assemble(sshable_hostname, location: "hetzner-hel1", net6: nil, ndp_nee
       Strand.create(
         prog: "Vm::HostNexus",
         label: "start",
-        stack: [{"spdk_version" => spdk_version}]
+        stack: [{"spdk_version" => spdk_version, "download_default_boot_image" => download_default_boot_image}]
```
enescakir, line 26: What if we change `Config.default_boot_image_name` to an array, and pass a lis of image names for `download_default_boot_image` instead of a boolean? We could still keep the default list with a single item: "ubuntu-jammy". If the list is empty or nil, the download can skipped. This would also allow operators to download images while setting up hosts. For example, they could pass a list to `Vm::HostNexus.assemble(....., default_boot_images: ["ubuntu-jammy", "github-gpu-ubuntu-2204", "github-ubuntu-2004"])` and no need for manual downloads after setup.

prog/vm/host_nexus.rb
furkansahin: No. Let's not do this. It just isn't clear.

pykello: I agree. I am not a fan of magic.

prog/vm/host_nexus.rb
pykello: https://github.com/ubicloud/ubicloud/pull/1601

PR #1599, opened 2024-05-15
Allow to exclude hosts during allocation

bsatzger:
When assembling a VM, a list of hosts can be provided, which the allocator will not consider as valid candidates for that VM. This can be leveraged by Postgres to guarantee that a standby is placed onto a different host than the primary.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -14,10 +14,13 @@ class Prog::Vm::Nexus < Prog::Base
     unix_user: "ubi", location: "hetzner-hel1", boot_image: Config.default_boot_image_name,
     private_subnet_id: nil, nic_id: nil, storage_volumes: nil, boot_disk_index: 0,
     enable_ip4: false, pool_id: nil, arch: "x64", allow_only_ssh: false, swap_size_bytes: nil,
-    distinct_storage_devices: false, force_host_id: nil)
+    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [])
 
     unless (project = Project[project_id])
       fail "No existing project"
+    end
+    if exclude_host_ids.include?(force_host_id)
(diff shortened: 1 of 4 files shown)


prog/vm/nexus.rb
```diff
@@ -188,6 +189,7 @@ def before_run
     queued_vms = Vm.join(:strand, id: :id).where(:location => vm.location, :arch => vm.arch, Sequel[:strand][:label] => "start")
     begin
       distinct_storage_devices = frame["distinct_storage_devices"] || false
+      host_exclusion_filter = frame["exclude_host_ids"] || []
```
furkansahin, line 192: Should we add a check and error out if `exclude_host_ids` cover `force_host_id`?

furkansahin (approved):

prog/vm/nexus.rb
fdr: you think maybe `fail BUG:` type fidelity of error?

fdr (approved):

prog/vm/nexus.rb
bsatzger: If `exclude_host_ids` and `force_host_id` intersect, we'd get a "no space left on any eligible host" error right now. We can fail earlier with such a specific error.

spec/prog/vm/nexus_spec.rb
```diff
@@ -352,6 +352,7 @@ def requested_disk_size(st)
         allocation_state_filter: ["accepting"],
         distinct_storage_devices: false,
         host_filter: [],
+        host_exclusion_filter: [],
```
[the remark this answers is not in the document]
bsatzger: this basically checks that by default `host_exclusion_filter` is empty, and only set if `exclude_host_ids` is set on the stack, see `it "can exclude hosts"`. this can probably be improved, though.

PR #1600, opened 2024-05-15
Make Dallas public

furkansahin:
**Update US region billing rates**

**Make Dallas visible**

### DIFF
--- a/config/billing_rates.yml
+++ b/config/billing_rates.yml
@@ -1,8 +1,8 @@
 - { id: 139d9a67-8182-8578-a303-235cabd5161c, resource_type: VmCores,                 resource_family: standard,        location: hetzner-fsn1,   unit_price: 0.000622768 }
 - { id: 08c502f7-df5d-8978-9896-feafa0ec5c40, resource_type: VmCores,                 resource_family: standard,        location: hetzner-hel1,   unit_price: 0.000595982 }
 - { id: 2acc2bbf-d451-49c6-93de-db8d15311042, resource_type: VmCores,                 resource_family: standard,        location: github-runners, unit_price: 0.000595982 }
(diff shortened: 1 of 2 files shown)


fdr (approved):
Hmm, no explanation why at all? I presume WDC has something to do with it not selling computers at present & maybe for a while...I suppose the inputs into pricing are often more sophisticated and easily blur into sensitive data about the company cost structure and can remain un-elaborated.

config/billing_rates.yml
```diff
@@ -1,8 +1,8 @@
 - { id: 139d9a67-8182-8578-a303-235cabd5161c, resource_type: VmCores,                 resource_family: standard,        location: hetzner-fsn1,   unit_price: 0.000622768 }
 - { id: 08c502f7-df5d-8978-9896-feafa0ec5c40, resource_type: VmCores,                 resource_family: standard,        location: hetzner-hel1,   unit_price: 0.000595982 }
 - { id: 2acc2bbf-d451-49c6-93de-db8d15311042, resource_type: VmCores,                 resource_family: standard,        location: github-runners, unit_price: 0.000595982 }
-- { id: 8cc76122-1463-8178-a89a-3b74434d1943, resource_type: VmCores,                 resource_family: standard,        location: leaseweb-wdc02, unit_price: 0.000622768 }
-- { id: 0f189c1f-b182-8978-bc1c-d1e0412718c3, resource_type: VmCores,                 resource_family: standard,        location: leaseweb-dal13, unit_price: 0.000622768 }
+- { id: 8cc76122-1463-8178-a89a-3b74434d1943, resource_type: VmCores,                 resource_family: standard,        location: leaseweb-wdc02, unit_price: 0.000849950 }
```
byucesoy, line 4: I thought we were setting this to $0.00085/min

furkansahin: With one more check, you're right. Updated it to $0.00085/min. Thanks

byucesoy (approved):

PR #1604, opened 2024-05-16
Get the URL for ubuntu and almalinux images in Prog::DownloadBootImage

pykello:
### Get the URL for ubuntu and almalinux images in Prog::DownloadBootImage

For all other images we created the URL in the clover side and passed it
as an argument to rhizome. It makes sense if we do the same for ubuntu
and almalinux too, so we have the entire URL logic in one place.

### Enforce URL and Version in download-boot-image.

We now send URL and version for all images, so we can enforce it in the
rhizome side

### DIFF
--- a/rhizome/host/lib/boot_image.rb
+++ b/rhizome/host/lib/boot_image.rb
@@ -12,6 +12,9 @@ class BootImage
   end
 
   def image_path
+    # YYY: Support for unversioned images is still required in StorageVolume
+    # code when we want to recreate storage. We can remove this check once we
+    # have removed all unversioned images from production.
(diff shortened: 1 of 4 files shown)


rhizome/host/lib/boot_image.rb
```diff
@@ -26,9 +29,6 @@ def image_root
   def download(url: nil, ca_path: nil, sha256sum: nil)
```
enescakir, line 29:
```suggestion
  def download(url:, ca_path: nil, sha256sum: nil)
```

enescakir (approved):

PR #1605, opened 2024-05-16
Do not select unversioned image if there is an active versioned image

enescakir:
When ordering by version, PostgreSQL defaults to returning NULLs first. This implies that even if we download the most recent version of an image, an available unversioned image will be chosen first. However, as we're beginning to disallow unversioned images, we can consider them outdated and avoid selecting them if an active versioned image exists.

### DIFF
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -253,7 +253,7 @@ module Scheduling::Allocator
       boot_image = BootImage.where(
         vm_host_id: vm_host.id,
         name: boot_image_name
-      ).exclude(activated_at: nil).order_by(Sequel.desc(:version)).first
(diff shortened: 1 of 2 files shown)


scheduling/allocator.rb
```diff
@@ -253,7 +253,7 @@ def allocate_boot_image(vm_host, boot_image_name)
       boot_image = BootImage.where(
         vm_host_id: vm_host.id,
         name: boot_image_name
-      ).exclude(activated_at: nil).order_by(Sequel.desc(:version)).first
+      ).exclude(activated_at: nil).order_by(Sequel.desc(:version, nulls: :last)).first
```
pykello, line 256: Maybe add a test.

pykello (approved):

fdr (approved):

PR #1606, opened 2024-05-16
Make GPU instance not selectable to regular VM customers

furkansahin:
## Make GPU instance not selectable to regular VM customers via APIs 
The APIs allow all vm size family to be selectable. In the web, we don't
visualize them anyway, however, in the APIs, they were selectable. With
this commit, we perform size validation in the API and only allow
visible vm_sizes to be picked.

## Validate boot-image in web
Without this check, customers can edit the parameter via inspect element
and use postgres base image as a base image for a VM. We do the same for
APIs.

### DIFF
--- a/lib/validation.rb
+++ b/lib/validation.rb
@@ -60,9 +60,10 @@ module Validation
     fail ValidationFailed.new({location: msg}) unless available_pg_locs.include?(location)
   end
 
-  def self.validate_vm_size(size)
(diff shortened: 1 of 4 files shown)


byucesoy:
Even if we don't visualize them, people can send requests to our web endpoints. We should add the check there too.

enescakir (approved):

PR #1607, opened 2024-05-16
Make sha256 sum calculation faster.

pykello:
1. Use `tee` to calculate SHA-256 checksum incrementally during the download process.
2. Switch to `openssl dgst -sha256` for checksum calculation. This change significantly reduces the processing time for large files. For example, on my system, a 6 GB file's checksum calculation time decreased from 15 seconds with the old method to 3.7 seconds with the new method.


fdr:
Nice and concise. I had a prototype  manipulating all the pipes in Ruby, it ended up long, with process redirection in the shell it comes out nicer.

fdr (approved):

PR #1613, opened 2024-05-17
Reinstate AlmaLinux images with a different versioning strategy

fdr:
This is sort of a revert of
69bf5c861656dc4423c5772ffd1747651661f307. But it also modifies the versioning to be less specific: AlmaLinux is now requested only by its major version, e.g. "9", not "9.3"

This method is the most like Ubuntu even though it includes one less version number component.  Ubuntu has releases like "22.04.3", where we never displayed the last version component.  The "3" in AlmaLinux "9.3" is a similar minor version that we don't need to display while remaining consistent with Ubuntu.

That I included the second digit of AlmaLinux before was a mistake I made in the most ancient era: March 2023, whereas the entire code base began in January 2023, adc109d2ea570a170e71e339304c502b7d03c798.

### DIFF
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -21,6 +21,10 @@ class Prog::DownloadBootImage < Prog::Base
     case image_name
     when "ubuntu-jammy"
       Config.ubuntu_jammy_version
+    when "almalinux-9"
+      Config.almalinux_9_version
+    when "almalinux-8"
+      Config.almalinux_8_version
     when "github-ubuntu-2204"
       Config.github_ubuntu_2204_version
     when "github-ubuntu-2004"
@@ -43,9 +47,12 @@ class Prog::DownloadBootImage < Prog::Base
       elsif image_name == "ubuntu-jammy"
         arch = (vm_host.arch == "x64") ? "amd64" : "arm64"
         "https://cloud-images.ubuntu.com/releases/jammy/release-#{version}/ubuntu-22.04-server-cloudimg-#{arch}.img"
-      elsif image_name == "almalinux-9.3"
+      elsif image_name == "almalinux-8"
+        fail "Only x64 is supported for almalinux-8" unless vm_host.arch == "x64"
+        "https://repo.almalinux.org/almalinux/8/cloud/x86_64/images/AlmaLinux-8-GenericCloud-#{version}.x86_64.qcow2"
+      elsif image_name == "almalinux-9"
         arch = (vm_host.arch == "x64") ? "x86_64" : "aarch64"
-        "https://vault.almalinux.org/9.3/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-9.3-#{version}.#{arch}.qcow2"
+        "https://repo.almalinux.org/almalinux/9/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-#{version}.#{arch}.qcow2"
       else
         fail "Unknown image name: #{image_name}"
       end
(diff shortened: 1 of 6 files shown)


prog/download_boot_image.rb
```diff
@@ -34,8 +34,8 @@ def sha256_sum
     hashes = {
       ["ubuntu-jammy", "x64", "20240319"] => "304983616fcba6ee1452e9f38993d7d3b8a90e1eb65fb0054d672ce23294d812",
       ["ubuntu-jammy", "arm64", "20240319"] => "40ea1181447b9395fa03f6f2c405482fe532a348cc46fbb876effcfbbb35336f",
-      ["almalinux-9.3", "x64", "20231113"] => "6bbd060c971fd827a544c7e5e991a7d9e44460a449d2d058a0bb1290dec5a114",
-      ["almalinux-9.3", "arm64", "20231113"] => "a064715bc755346d5a8e1a4c6b1b7abffe4de03f1b0584942d5483ed32aafd67",
+      ["almalinux-9", "x64", "20231113"] => "6bbd060c971fd827a544c7e5e991a7d9e44460a449d2d058a0bb1290dec5a114",
```
pykello, line 37: 1. the `url` function needs to be updated (it seems your PR is not based on the up to date `main` branch)
2. What do you think about changing the version from `20231113` to `9.3-20231113`? Otherwise we need to make sure minor versions don't conflict in release dates.

pykello (changes requested):

prog/download_boot_image.rb
pykello: For 2, I think date is ok and every new release increases the minor & there is 1-to-1 mapping between minors and dates.

Also, just date is preferred since it is easier to sort & select the latest version.

fdr:
> If an existing customer has the boot image 'almalinux-9.3' when we deploy this PR, we need to update the respective columns.

good catch.

prog/download_boot_image.rb
```diff
@@ -26,9 +26,9 @@ def url
       elsif image_name == "ubuntu-jammy"
         arch = (vm_host.arch == "x64") ? "amd64" : "arm64"
         "https://cloud-images.ubuntu.com/releases/jammy/release-#{version}/ubuntu-22.04-server-cloudimg-#{arch}.img"
-      elsif image_name == "almalinux-9.3"
+      elsif image_name == "almalinux-9"
         arch = (vm_host.arch == "x64") ? "x86_64" : "aarch64"
-        "https://vault.almalinux.org/9.3/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-9.3-#{version}.#{arch}.qcow2"
+        "https://repo.almalinux.org/almalinux/9/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-#{version}.#{arch}.qcow2"
```
pykello, line 31: They delete old images from `https://repo.almaunix.org/`, so this will give 404 after their next release. Not sure if this is okay or not. However, they keep the images in `https://valut.almalinux.org/` for longer.

pykello:
> If an existing customer has the boot image 'almalinux-9.3' when we deploy this PR, we need to update the respective columns.

I am not sure if I understand this comment correctly. But note that updating the name in the database will break ability to remove the image it from the host, since the path in rhizome is a function of the name.

fdr:
> > If an existing customer has the boot image 'almalinux-9.3' when we deploy this PR, we need to update the respective columns.
> 
> I am not sure if I understand this comment correctly. But note that the path in the rhizome side is a function of the name, and is important when operations like removing the image.

if we have existing AlmaLinux VMs, the boot_image field may need a change, it's currently like `almalinux-9.3`. Granted, we've been somewhat fast and loose here, including removing support altogether, given that it sees little use and was even broken at one time.

Do you see a problem with this conflicting with legacy VMs that pre-date BootImage? Did you do some kind of backport that will be broken by it?

pykello:
> Do you see a problem with this conflicting with legacy VMs that pre-date BootImage? Did you do some kind of backport that will be broken by it?

Updating `Vm.boot_image` columns is ok. So, go ahead with that.

I was more concerned with updating `BootImage.name` fields. Note that we also created `BootImage` records for legacy almalinux installations & updated legacy VM storage volumes to point to them. Updating `BootImage.name` for legacy `almalinux-9.3` needs more care, as `BootImage.name` determines the image path.

prog/download_boot_image.rb
fdr: Looks like new versions are not in vault. I complained a bit about this making permalinks difficult, but between "lagging a version behind" and "need to figure out something to deal with this archiving problem," perhaps we should do "need to figure something out" (or change links to vault as they age out)

fdr:
> If an existing customer has the boot image 'almalinux-9.3' when we deploy this PR, we need to update the respective columns.

...hmm, well, maybe we should, but do we? aren't they busted right now, where there's no support? what does that accomplish? actually, we're a bit fuzzy on what the column on Vm does now that BootImage also exists and there's a foreign key to it.

fdr:
Requesting re-review, I didn't change a whole lot, but added almalinux-8 and phased the patches for deploy

pykello (approved):

enescakir:
Can you add default versions here?

https://github.com/ubicloud/ubicloud/blob/main/prog/download_boot_image.rb#L20

fdr:
@enescakir alright, done per the usual. Seems like there's a suboptimal amount of shotgun surgery here, but since we expect major fluctuation here as custom image support rolls in...I think I'm content to let a sleeping dog lie.

enescakir:
> @enescakir alright, done per the usual. Seems like there's a suboptimal amount of shotgun surgery here, but since we expect major fluctuation here as custom image support rolls in...I think I'm content to let a sleeping dog lie.

We might reduce the number of areas we need to modify when introducing a new boot image
https://github.com/ubicloud/ubicloud/pull/1597#discussion_r1601548640

enescakir (approved):

fdr:
> > @enescakir alright, done per the usual. Seems like there's a suboptimal amount of shotgun surgery here, but since we expect major fluctuation here as custom image support rolls in...I think I'm content to let a sleeping dog lie.
> 
> We might reduce the number of areas we need to modify when introducing a new boot image [#1597 (comment)](https://github.com/ubicloud/ubicloud/pull/1597#discussion_r1601548640)

Wew, don't think the juice is worth that squeeze.

PR #1614, opened 2024-05-17
Github E2E Tests

velioglu:
Adding github E2E tests. Following items are completed in the PR

- Moving default image version handling from host_nexus to download_boot_image. So, no need to pass version on tests.
- Imstall ngrok on test host to expose local server to public. Test app created for e2e tests will send hooks to the static domain provided by ngrok.
- Add github runner e2e tests. It will download github runner boot images and trigger test worklows passed in the github test config file. 
- Run e2e test in parallel.
- Update tests to cover all the changes above.


config.rb
```diff
@@ -95,6 +95,8 @@ def self.test?
   optional :github_app_webhook_secret, string, clear: true
   optional :vm_pool_project_id, string
   optional :github_runner_service_project_id, string
+  optional :github_test_installation_id, string
+  optional :github_test_access_token, string
```
enescakir, line 99: We also have `ci_hetzner_sacrificial_server_id` for E2E tests, denoted by the `ci` prefix. As the required configurations for E2E tests are growing, it's becoming difficult to distinguish between those for E2E and those for production. Could we relocate all three configurations to distinct section with a prefix for clarity such as:

```
# E2E Tests
e2e_hetzner_server_id
e2e_github_installation_id
e2e_github_access_token
```

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+  def test_runs
+    @@test_runs ||= frame["test_cases"].flat_map { tests[_1]["runs"] }
+  end
+
+  def vm_host_id
+    @@vm_host_id ||= frame["vm_host_id"] || VmHost.first.id
+  end
+
+  def client
+    @@client ||= Octokit::Client.new(access_token: Config.github_test_access_token)
```
enescakir, line 148: Do we need an additional access token here? I believe `Github.installation_client(installation_id)` should be able to access these workflows.

bin/ci
```diff
@@ -19,6 +19,12 @@ def main(options)
     wait_until(unencrypted_vms_st)
   end
 
+  github_runner_test_cases = options[:test_cases].select { _1.include?("github_runner") }
+  if github_runner_test_cases
```
enescakir, line 23:
```suggestion
    if options[:test_cases].include?("github")
```

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
+# frozen_string_literal: true
+
+require "octokit"
+require "yaml"
+
+class Prog::Test::GithubRunner < Prog::Test::Base
```
enescakir, line 6: What do you think about splitting this prog into two, similar to the VmGroup and Vm? The current version of `Prog::Test::GithubRunner` maintains a list of test cases and manages all workflow cases. We could move the operations for individual workflow cases to another prog, while the parent program simply waits for the child progs to complete. The names are just examples.

- `Prog::Test::GithubRunner`: It's the parent prog. We assemble this one. It buds `Prog::Test::GithubWorkflow` for each test cases, and waits for children be done.
- `Prog::Test::GithubWorkflow`: It manages steps for a single test case. The current `trigger_test_run` and `test_run_conclusion` helpers are actually a label in it. It triggers and checks only the specified test case, without iterating over multiple cases. Additionally, it downloads the image for the given test case, enabling to push instead of bud `Prog::DownloadBootImage`

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+
+  def self.assemble(vm_host_id, test_cases)
+    github_service_project = Project.create(name: "Github Runner Service Project") { _1.id = Config.github_runner_service_project_id }
+    github_service_project.associate_with_project(github_service_project)
+
+    github_test_project = Project.create_with_id(name: "Github Runner Test Project")
+    GithubInstallation.create_with_id(
+      installation_id: Config.github_test_installation_id,
+      name: "GithubTestInstallationName",
+      type: "GithubTestInstallationType",
```
enescakir, line 18: nit: more natural
```suggestion
      name: "TestUser",
      type: "User",
```

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+      project_id: github_test_project.id
+    )
+    github_test_project.associate_with_project(github_test_project)
```
enescakir, line 21: nit: Associate after creation

```suggestion
    github_test_project = Project.create_with_id(name: "Github Runner Test Project")
    github_test_project.associate_with_project(github_test_project)
    GithubInstallation.create_with_id(
      installation_id: Config.github_test_installation_id,
      name: "GithubTestInstallationName",
      type: "GithubTestInstallationType",
      project_id: github_test_project.id
    )
```

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+    run_id = get_latest_run_id(repo_name, workflow_name, branch_name)
+    begin
+      client.cancel_workflow_run(repo_name, run_id)
+    rescue
+      puts "Workflow run #{run_id} for #{repo_name} has already been finished"
+    end
+  end
+
+  def tests
+    @@tests ||= YAML.load_file("config/github_runner_e2e_tests.yml").to_h { [_1["name"], _1] }
```
enescakir, line 136: I still have doubts about putting test cases in the YAML instead of a constant variable. It seems too fragmented to me.

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+  end
+
+  def test_runs
+    @@test_runs ||= frame["test_cases"].flat_map { tests[_1]["runs"] }
+  end
+
+  def vm_host_id
+    @@vm_host_id ||= frame["vm_host_id"] || VmHost.first.id
```
enescakir, line 144: Why not?
```suggestion
    @vm_host_id ||= frame["vm_host_id"] || VmHost.first.id
```

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+    pop "GithubRunnerGroup tests are finished!"
+  end
+
+  label def failed
+    cleanup
+    nap 15
+  end
+
+  def trigger_test_run(repo_name, workflow_name, branch_name)
+    url = "repos/#{repo_name}/actions/workflows/#{workflow_name}/dispatches"
```
enescakir, line 88: nit: Octokit.rb has a method for it
https://github.com/octokit/octokit.rb/blob/c976852a4b54a967f26b5b943f7bc983ad2e3514/lib/octokit/client/actions_workflows.rb#L41

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+  end
+
+  def get_run_conclusion(repo_name, run_id)
+    run = client.workflow_run(repo_name, run_id)
+    run[:conclusion]
+  end
+
+  def cleanup
+    Project[frame["github_service_project_id"]]&.destroy
+    Project[frame["github_test_project_id"]]&.destroy
```
enescakir, line 116: Interesting. GithubInstallation has a `project_id` foreign key. How does it allow destroy without removing the installation?

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+  end
+
+  label def check_test_runs
+    test_runs.each do |test_run|
+      conclusion = test_run_conclusion(test_run["repo_name"], test_run["workflow_name"], test_run["branch_name"])
+      if FAIL_CONCLUSIONS.include?(conclusion)
+        fail_test "Test run for #{test_run["repo_name"]}, #{test_run["workflow_name"]}, #{test_run["branch_name"]} failed with conclusion #{conclusion}"
+      elsif IN_PROGRESS_CONCLUSIONS.include?(conclusion) || conclusion.nil?
+        nap 15
+      end
```
enescakir, line 71: The workflow run has `status` field with fewer possible options: requested, in_progress, completed, queued, pending, waiting. The `completed` is important one.

We might simplify this part as following. So we don't need to keep the list of conclusions.

- Nap unless `status == "completed"`
- If `status == "completed"`
  - fail unless `conclusion == "success"`

prog/test/github_runner.rb
```diff
@@ -0,0 +1,150 @@
...
+    end
+
+    hop_finish
+  end
+
+  label def finish
```
enescakir, line 77: Should we also verify that the GithubRunner has been successfully destroyed? This will ensure the destruction workflow is functioning correctly.

prog/test/github_runner.rb
enescakir: It also helps us track each test case separately and identify the failed ones more easily.

prog/test/github_runner.rb
velioglu: Yes, makes sense. I was thinking about having multiple installations for e2e tests but changed my mind afterwards, it is leftover.

bin/ci
```diff
@@ -19,6 +19,12 @@ def main(options)
     wait_until(unencrypted_vms_st)
   end
 
+  github_runner_test_cases = options[:test_cases].select { _1.include?("github_runner") }
```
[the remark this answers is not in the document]
velioglu: I think we need to have a way to trigger github runner cases separately, so we need to pass that information to e2e tests somehow. Given that we expect tests will be running automatically, development experience overweigh the complexity.

prog/test/github_runner.rb
velioglu: I find separating vm and vmgroup useful as we are doing multiple tests in vm after creating it. Having them in the vmgroup would make it more complex. 

Though, in that case we only trigger the run and check the result. If we implement it in the way you mentioned I would still have loops over all the child strands to check and cancel them. I also find propagating failure from child to parent as overhead. So, I'll be keeping it as it is.

prog/test/github_runner.rb
velioglu: I initially implemented it in this way though the status returned by github is flaky. Few cases, I got status nil although test case was still running. I don't want to add a check whether the status is empty as it is not intuitive. Conclusion state is more reliable.

prog/test/github_runner.rb
velioglu: I was thinking we had foreign key "on delete cascade" as destroying project also deletes GithubInstallation. Let me explicitly delete github installation then delete project afterwards. Let me also check why we can delete projects like that with a subsequent PR.

prog/test/github_runner.rb
velioglu: I though about it and I slightly prefer having it in config file instead of having it defined on the code itself as I find understanding and updating config easier. Though, if you have strong opinion I'm ok to change it as well.

config.rb
velioglu: I've moved github related one to E2E tests part in the config, I'll do it for sacrificial server with subsequent PR

bin/ci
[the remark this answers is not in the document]
velioglu: Within my tests preparing host takes ~12-15 mins. Since I've changed tests to run in parallel, github runner tests add ~12 mins (downloading images takes almost all) on top of it. I'm planning to keep E2E test runs periodic as we have right now. We can also trigger if we have changes related to github runner. Agree to moving larger host with faster network.

config/github_runner_e2e_tests.yml
```diff
@@ -0,0 +1,2 @@
+- {name: github_runner_ubuntu_2204, image_name: github-ubuntu-2204, runs: [{repo_name: velioglu/github_e2e, workflow_name: test_2204.yml, branch_name: main}]}
```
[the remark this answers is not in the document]
velioglu: That's not the name of the branch on the ubicloud repo but the branch on the e2e repo on which the github test workflow run. So, you can run e2e test for a branch on ubicloud repo by selecting it from the action page.

bin/ci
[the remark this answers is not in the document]
enescakir: I agree with the idea of parallelization, but we need to be cautious. To my knowledge, our E2E host has a capacity of 4 VMs (though I could be mistaken). The VmGroup test provisions 3 virtual machines. If we run two VmGroup tests in parallel, each might allocate 2 VMs and then wait for a third, potentially leading to a deadlock.

config/github_runner_e2e_tests.yml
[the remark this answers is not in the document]
enescakir: Sorry, my bad :( I shouldn't be thinking about work on the weekend :D

prog/download_boot_image.rb
```diff
@@ -63,9 +80,6 @@ def blob_storage_client
   end
 
   label def start
-    # YYY: we can remove this once we enforce it in the database layer.
-    fail "Version is required" if version.nil?
```
enescakir, line 67: Someone can still execute `vm_host.download_boot_image("image-name", version: nil)`. Let's maintain this additional check until we impose a constraint on the database.

bin/ci
```diff
@@ -9,14 +9,28 @@ def main(options)
...
+    tests_to_wait << [encrypted_vms_st]
 
     unencrypted_vms_st = Prog::Test::VmGroup.assemble(storage_encrypted: false, test_reboot: false)
     log(unencrypted_vms_st, "storage_encrypted: false")
-    wait_until(unencrypted_vms_st)
+    tests_to_wait << [unencrypted_vms_st]
+  end
+
+  github_runner_test_cases = options[:test_cases].select { _1.include?("github_runner") }
+  if github_runner_test_cases
```
enescakir, line 24: It returns `true` even the array is empty.
```suggestion
  unless github_runner_test_cases.empty?
```

```
[47] clover-development(main)> puts "array is not empty" if [1,2];
array is not empty
[48] clover-development(main)> puts "array is not empty" if [];
array is not empty
```

prog/test/github_runner.rb
```diff
@@ -0,0 +1,148 @@
...
+  label def trigger_test_runs
+    test_runs.each do |test_run|
+      unless trigger_test_run(test_run["repo_name"], test_run["workflow_name"], test_run["branch_name"])
+        update_stack({"fail_message" => "Can not trigger workflow for #{test_run["repo_name"]}, #{test_run["workflow_name"]}, #{test_run["branch_name"]}"})
+        hop_clean_resources
+      end
+    end
+
+    # To make sure that test runs are triggered
+    sleep 30
```
enescakir, line 62: 30 seconds may not be enough, especially during frequent GitHub incidents. To ensure safety, we can check the creation date of the latest run to confirm it was created after our trigger. If a new run isn't created, it will use the previous one and assume success. The project's creation time can serve as an example. You can either save the timestamp to the frame or use the one from the project which it's not perfect one. This check can also be performed under the `check_test_runs` label.

prog/test/github_runner.rb
```diff
@@ -0,0 +1,148 @@
...
+  def test_runs
+    @test_runs ||= frame["test_cases"].flat_map { tests[_1]["runs"] }
+  end
+
+  def vm_host_id
+    @vm_host_id ||= frame["vm_host_id"] || VmHost.first.id
+  end
+
+  def client
+    @client ||= Octokit::Client.new(client_id: Config.github_app_client_id, client_secret: Config.github_app_client_secret)
```
enescakir, line 146:
```suggestion
    @client ||= Github.app_client
```

spec/prog/download_boot_image_spec.rb
```diff
@@ -5,11 +5,13 @@
 RSpec.describe Prog::DownloadBootImage do
```
enescakir, line 5: I'd like to see this change in one commit. If you prefer two commits, please ensure they're consecutive. Same for the other test commit too.

```
Move default image versions from host_nexus to download_boot_image
Add tests moving image version from host_nexus to download_boot_image
```

bin/ci
```diff
@@ -9,20 +9,28 @@ def main(options)
...
   if github_runner_test_cases
     runner_st = Prog::Test::GithubRunner.assemble(options[:vm_host_id], github_runner_test_cases)
-    wait_until(runner_st)
+    tests_to_wait << [runner_st]
+  end
+
+  # Although wait_until will be blocked while checking the first one
+  # it won't affect the total time as other strands will continue in parallel.
+  # No need to make it parallel.
+  tests_to_wait.map do |strand, label|
```
enescakir, line 32: I'm confused here. I thought you wanted to add a feature that waits for different labels for various tests. However, right now, all of them are set to wait until they exit. If you plan to retain this feature for future use, it could be beneficial to explicitly push `nil` for exiting tests, like `tests_to_wait << [runner_st, nil]`. Currently, no one else knows about this feature.

prog/download_boot_image.rb
velioglu: Even if it is passed as nil we'll still get the default version since it will be nil

prog/download_boot_image.rb
```diff
@@ -10,13 +10,30 @@ def image_name
   end
 
   def version
-    @version ||= frame.fetch("version")
+    @version ||= frame.fetch("version") { default_boot_image_version(image_name) }
```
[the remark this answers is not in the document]
velioglu: Ruby calculates the default value even if the fetch returns some value. When the block is passed it only calculates the value if the fetch doesn't return. I wanted to keep the tests as it is, so block is used instead of default value directly.

bin/ci
velioglu: I have passed it as `tests_to_wait << [unencrypted_vms_st]` since the label would be nil if there is no label passed while adding tests to tests_to_wait. Let me add nil for being more explicit.

prog/test/github_runner.rb
```diff
@@ -0,0 +1,147 @@
...
+    runs = client.workflow_runs(repo_name, workflow_name, {branch: branch})
+    latest_run = runs[:workflow_runs].first
+    latest_run[:conclusion]
+  end
+
+  def get_latest_run_id(repo_name, workflow_name, branch)
+    runs = client.workflow_runs(repo_name, workflow_name, {branch: branch})
+    latest_run = runs[:workflow_runs].first
+    latest_run[:id]
+  end
```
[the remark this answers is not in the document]
velioglu: I'm keeping two seperate function to be more explicit but created a function for common part.

bin/ci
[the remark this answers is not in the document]
velioglu: As runner tests are starting after downloading image, it doesn't coincidence with vm tests. So, it is ok for now. We should also use bigger host for tests. I'll work on that in parallel.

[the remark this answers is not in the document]
enescakir: We have 2 vm group tests, they coincidence with each other 😄

prog/download_boot_image.rb
enescakir: Are you sure? It think `fetch` only checks the existing of the key. If it exits with `nil` value, it will continue with `nil` value.

```
    @version ||= frame.fetch("version") { default_boot_image_version(image_name) }
```

bin/ci
enescakir: Do we fix that?

prog/download_boot_image.rb
velioglu: You're right, I mistested it apparently. I've added it back and updated the tests accordingly.

enescakir (approved):

PR #1615, opened 2024-05-17
Loadbalancer MVP

furkansahin:


enescakir:
"Load Balancer" is typically two words, not a single one like "Loadbalancer". Is there a specific reason for using it as a single word?

furkansahin:
nah, this is mere a draft pr. needs a lot of clean-up, name is one of them.

prog/vnet/update_load_balancer.rb
```diff
@@ -0,0 +1,98 @@
...
+table inet nat {
+  set neighbor_ips_v4 {
+    type ipv4_addr;
+    flags interval;
+    #{neighbor_ips_v4_line}
+  }
+
+  set neighbor_ips_v6 {
+    type ipv6_addr;
+    flags interval;
```
fdr, line 55: btw why `flags interval`? do we use any ranges (a.k.a. intervals) here?

furkansahin:
moved to #1737

PR #1616, opened 2024-05-18
Firmware download improvement

bsatzger:
With this PR, we add the ability to download firmware from https://github.com/ubicloud/build-edk2-firmware and store it on the host in `/opt/fw` as files named 'CLOUDHV-`version`.fd'. New hosts will start to download the firmware (in addition to legacy firmware) during `prep_host`. The PR also adds capabilities to download specified versions to hosts on demand. With this, new firmware versions can be deployed to and enabled for all hosts by performing these steps:

1. Execute new method `download_firmware` on each host to download the new version
2. Update the version in `rhizome/host/lib/cloud_hypervisor.rb` to the new version

In order to switch from legacy to new firmware, instead of step 2, we need to rename `FIRMWARE` to `FIRMWARE_LEGACY` and `NEW_FIRMWARE` to `FIRMWARE`, to enable the new firmware for VMs.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -188,6 +188,14 @@ class VmHost < Sequel::Model
     Strand.create_with_id(schedule: Time.now, prog: "DownloadBootImage", label: "start", stack: [{subject_id: id, image_name: image_name, custom_url: custom_url, version: version}])
   end
 
+  # Introduced for downloading firmware via REPL.
+  def download_firmware(version_x64: nil, version_arm64: nil, sha256_x64: nil, sha256_arm64: nil)
+    version, sha256 = (arch == "x64") ? [version_x64, sha256_x64] : [version_arm64, sha256_arm64]
+    raise ArgumentError, "No version provided" if version.nil?
+    raise ArgumentError, "No SHA-256 digest provided" if sha256.nil?
+    Strand.create_with_id(schedule: Time.now, prog: "DownloadFirmware", label: "start", stack: [{subject_id: id, version: version, sha256: sha256}])
+  end
+
   def hetznerify(server_id)
     DB.transaction do
       HetznerHost.create(server_identifier: server_id) { _1.id = id }
--- a/rhizome/common/lib/util.rb
+++ b/rhizome/common/lib/util.rb
@@ -57,11 +57,16 @@ def sync_parent_dir(f)
   }
 end
 
-def safe_write_to_file(filename, content)
+def safe_write_to_file(filename, content = nil)
+  raise ArgumentError, "must provide either content or block" if (content.nil? && !block_given?) || (!content.nil? && block_given?)
   temp_filename = filename + ".tmp"
   File.open(temp_filename, File::RDWR | File::CREAT) do |f|
     f.flock(File::LOCK_EX)
-    f.puts(content)
+    if block_given?
+      yield f
+    else
+      f.puts(content)
+    end
     File.rename(temp_filename, filename)
   end
 end
--- a/rhizome/host/lib/cloud_hypervisor.rb
+++ b/rhizome/host/lib/cloud_hypervisor.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
+require "fileutils"
 require_relative "../../common/lib/arch"
 
 module CloudHypervisor
-  FirmwareClass = Struct.new(:version, :name) {
+  FirmwareClassLegacy = Struct.new(:version, :name) {
     def url
       "https://github.com/fdr/edk2/releases/download/#{version}/#{name}"
     end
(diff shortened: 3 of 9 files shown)


rhizome/host/lib/cloud_hypervisor.rb
```diff
@@ -13,6 +14,34 @@ def path
...
+    def path
+      "#{firmware_root}/CLOUDHV-#{arch}-#{version}.fd"
+    end
+
+    def download
+      return if File.exist?(path)
+      FileUtils.mkdir_p(firmware_root)
+      temp_path = path + ".tmp"
+      curl(temp_path, url)
+      FileUtils.mv(temp_path, path)
```
fdr, line 35: I suppose we should flush this (see safe_write_to_file, though it may need a variant that accepts a block to let a subprocess do the writing.

I suppose also given the size, it'd be nice (but not *necessary* if too bloated) to digest the file similar to recent changes to images, now that the standards of adequacy are going up.

rhizome/host/lib/cloud_hypervisor.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
+require "fileutils"
 require_relative "../../common/lib/arch"
 
 module CloudHypervisor
-  FirmwareClass = Struct.new(:version, :name) {
+  FirmwareClassLegacy = Struct.new(:version, :name) {
```
fdr, line 7: Seems like this the previous commit references this symbol, no? (Not sure how much of this came from my old prototype, which if it did this, can be called a concession to being only a prototype...)

model/vm_host.rb
```diff
@@ -188,6 +188,11 @@ def download_boot_image(image_name, version:, custom_url: nil)
     Strand.create_with_id(schedule: Time.now, prog: "DownloadBootImage", label: "start", stack: [{subject_id: id, image_name: image_name, custom_url: custom_url, version: version}])
   end
 
+  # Introduced for downloading a new firmware.
```
fdr, line 191: A somewhat out of character comment. Is it trying to say that it is intended for console use?

fdr:
It seems like the patches need some squashing. It'd also be nice to have a second patch set showing an unwinding of the legacy stuff, to show how much net code we're adding for this. This particular change is more phasing-intensive than program-complexity-intensive.

rhizome/host/lib/cloud_hypervisor.rb
bsatzger: oops, my bad. this was supposed to be in the first commit. I thought it already was ... but it isn't ... as you're pointing out.

rhizome/host/lib/cloud_hypervisor.rb
```diff
@@ -13,10 +14,38 @@ def path
...
+    def url
+      "https://github.com/ubicloud/build-edk2-firmware/releases/download/edk2-stable#{version}-#{arch}/CLOUDHV-#{arch}.fd"
+    end
+
+    def firmware_root
+      "/opt/fw"
+    end
+
+    def path
+      "#{firmware_root}/CLOUDHV-#{arch}-#{version}.fd"
```
enescakir, line 27: Since the VM host's architecture doesn't change and it's a single architecture, should we keep `arch` in the firmware name? Why not just use `"#{firmware_root}/CLOUDHV-#{version}.fd"` similar to our boot image names?

rhizome/host/lib/cloud_hypervisor.rb
```diff
@@ -13,10 +14,38 @@ def path
     end
   }
 
+  FirmwareClass = Struct.new(:arch, :version) {
```
enescakir, line 17: Should we pass `arch` here? It's rhizome code and only has a single architecture.

rhizome/host/lib/cloud_hypervisor.rb
fdr: Yeah, there is no great reason that I know of yet, since we'll won't have multiple architectures present simultaneously.

rhizome/host/lib/cloud_hypervisor.rb
bsatzger: Extended `safe_write_to_file` and using it to produce the firmware file. Also, it is now digested.

rhizome/host/lib/cloud_hypervisor.rb
bsatzger: I also went back and forth between these two alternatives. In the latest iteration, we omit the architecture from the filename. We have one more safeguard in place as we now check the firmware's digest. But I'm happy to add it back if there are concerns.

bsatzger:
> It'd also be nice to have a second patch set showing an unwinding of the legacy stuff, to show how much net code we're adding for this. 

See https://github.com/ubicloud/ubicloud/commits/ben/firmware_remove_legacy/

rhizome/common/lib/util.rb
```diff
@@ -58,10 +58,16 @@ def sync_parent_dir(f)
 end
 
 def safe_write_to_file(filename, content)
+  safe_create_file_with_content(filename) do |f|
+    f.puts(content)
+  end
+end
+
+def safe_create_file_with_content(filename, &block)
```
byucesoy, line 66: By just reading the function names, I'd expect `safe_create_file_with_content` to get `content` and `safe_write_to_file` to  get `block`. I guess you kept it this way to avoid changing existing references to `safe_write_to_file`.

One alternative you can do is having just one function, which can accept both content and block and its behavior would change based on whether block is passed or not. You can check if block is passed with `block_given?` helper. This is common pattern in ruby. We even have an example in our codebase at lib/clog.rb

model/vm_host.rb
```diff
@@ -188,6 +188,14 @@ def download_boot_image(image_name, version:, custom_url: nil)
     Strand.create_with_id(schedule: Time.now, prog: "DownloadBootImage", label: "start", stack: [{subject_id: id, image_name: image_name, custom_url: custom_url, version: version}])
   end
 
+  # Introduced for downloading firmware via REPL.
+  def download_firmware(version_x64: nil, version_arm64: nil, sha256_x64: nil, sha256_arm64: nil)
```
byucesoy, line 192: It looks like at a given time, only one pair of `version`, `sha256` would be used, right? If so, why do we get 4 parameters? Can we just get `version` and `sha256`? Then we wouldn't need to pick based on `arch`, it would be caller's responsibility to provide correct `version`/`sha256` pair (which kinda already is).

spec/model/vm_host_spec.rb
```diff
@@ -82,6 +82,35 @@
...
   end
 
+  it "has a shortcut to download a new firmware for x64" do
+    vh.id = "46683a25-acb1-4371-afe9-d39f303e44b4"
+    vh.arch = "x64"
+    expect(Strand).to receive(:create) do |args|
+      expect(args[:prog]).to eq("DownloadFirmware")
+      expect(args[:stack]).to eq([subject_id: vh.id, version: "202405", sha256: "sha-1"])
+    end
+    vh.download_firmware(version_x64: "202405", version_arm64: "202406", sha256_x64: "sha-1", sha256_arm64: "sha-2")
```
byucesoy, line 92: Do we expect/want caller to pass version info for the non-used arch as well?

rhizome/common/lib/util.rb
bsatzger: ok, if it's confusing I'll collapse it into a single function.

model/vm_host.rb
bsatzger: The way it is right now can help to avoid mistakes when downloading firmware for a single host. This would work on x64 but fail on arm64.
`vmh.download_firmware(version_x64: "202402", sha256_x64: "somesha")`

When downloading new version for the whole fleet, it makes things a little easier, as you can do something like:
`VmHost.all.map { _1.download_firmware(version_x64: "202402", sha256_x64: "somesha", version_arm64: "202211", sha256_arm64: "anothersha")}`

spec/model/vm_host_spec.rb
bsatzger: We don't.

fdr (approved):
although you can finish hashing out things with @byucesoy, I think it's in pretty good shape. looking forward to closing this long overdue weakness.

byucesoy (approved):

PR #1618, opened 2024-05-20
Decouple VM compute and storage

byucesoy:
**Add active_from field to billing rates**
We sometimes need to update billing rates, however, directly changing the rates
would affect the historical data. To avoid this, we are adding an active_from
field. Thanks to this field, we can create a new billing rate with the updated
data and more recent active_from date. The old billing rate will still be used
for the historical data and existing resources, but new resources will use the
new billing rate.

**Add billing rates for VM storage**
We used to have a single billing rate for VMs, which was combination of the
amount we would want to charge for both compute and storage. We are planning to
decouple storage and compute, thus we need to have separate billing rates for
each. This commit adds the billing rates for VM storage and also adds new rates
for VM compute, which is lower than the combined rate we used to have.

**Add backend code allowing picking storage independently from VM size**
We are expanding the VmSize option to include minimum and maximum storage sizes
as well as step size per each instance size. These will be passed to frontend
to dynamically update the user interface.

We are also adding storage_size option to API, which needs to be added to our
API documentation.

**Add frontend code allowing picking storage independently from VM size**
This frontend code is admittedly a bit of patch work on top of our existing UI
code. So far frontend code was not our primary focus and as a result of that,
it become a bit difficult to extend and test. I initially wanted to clean the
UI code up a bit, but that effort quickly exploded into a big refactor touching
lots of different areas, such as adding javascript tests, modifying the way we
render the forms, etc. I decided to postpone that work for now and make minimal
set of changes to get the storage decoupling feature working.

I'm planning to come back to form rendering part of the UI code in the future
and refactor it to make it more maintainable and testable.

**Remove unnecessary expectation in tests**

**Automatically select first instance size in VM creation form**
This ensures that the javascript code depending on instance size works even at
the initial load. It also makes the form more user-friendly by requiring one
less click.

**Increase default VM storage size to 40GB**
We decided to increase default VM storage size to 40GB.

image

### DIFF
--- a/config/billing_rates.yml
+++ b/config/billing_rates.yml
@@ -1,13 +1,19 @@
-- { id: 139d9a67-8182-8578-a303-235cabd5161c, resource_type: VmCores,                 resource_family: standard,        location: hetzner-fsn1,   unit_price: 0.0006227680, active_from: 2023-01-01T00:00:00Z }
-- { id: 08c502f7-df5d-8978-9896-feafa0ec5c40, resource_type: VmCores,                 resource_family: standard,        location: hetzner-hel1,   unit_price: 0.0005959820, active_from: 2023-01-01T00:00:00Z }
-- { id: 2acc2bbf-d451-49c6-93de-db8d15311042, resource_type: VmCores,                 resource_family: standard,        location: github-runners, unit_price: 0.0005959820, active_from: 2023-01-01T00:00:00Z }
-- { id: 8cc76122-1463-8178-a89a-3b74434d1943, resource_type: VmCores,                 resource_family: standard,        location: leaseweb-wdc02, unit_price: 0.0008500000, active_from: 2023-01-01T00:00:00Z }
-- { id: 0f189c1f-b182-8978-bc1c-d1e0412718c3, resource_type: VmCores,                 resource_family: standard,        location: leaseweb-dal13, unit_price: 0.0008500000, active_from: 2023-01-01T00:00:00Z }
-- { id: 8c82e087-10aa-8d78-876f-88afcc162d4c, resource_type: VmCores,                 resource_family: standard-gpu,    location: hetzner-fsn1,   unit_price: 0.0006227680, active_from: 2023-01-01T00:00:00Z }
-- { id: 80c193bc-528a-8d78-9377-17523b162c58, resource_type: VmCores,                 resource_family: standard-gpu,    location: hetzner-hel1,   unit_price: 0.0005959820, active_from: 2023-01-01T00:00:00Z }
-- { id: c1eb36d9-6b46-8578-869a-92e64646b547, resource_type: VmCores,                 resource_family: standard-gpu,    location: github-runners, unit_price: 0.0005959820, active_from: 2023-01-01T00:00:00Z }
-- { id: b44a8854-f7fb-8578-b1b8-f38b218b52e3, resource_type: VmCores,                 resource_family: standard-gpu,    location: leaseweb-wdc02, unit_price: 0.0006227680, active_from: 2023-01-01T00:00:00Z }
-- { id: 88dcb876-0a21-8978-bc58-558b1755ad29, resource_type: VmCores,                 resource_family: standard-gpu,    location: leaseweb-dal13, unit_price: 0.0006227680, active_from: 2023-01-01T00:00:00Z }
+# Active billing rates
+- { id: 8a890156-96cb-4087-878c-51dc4ed2ad8a, resource_type: VmCores,                 resource_family: standard,        location: hetzner-fsn1,   unit_price: 0.0005525180, active_from: 2024-05-27T00:00:00Z }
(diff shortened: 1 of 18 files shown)


prog/vm/nexus.rb
```diff
@@ -334,6 +334,14 @@ def write_params_json
       amount: vm.cores
     )
 
+    BillingRecord.create_with_id(
+      project_id: project.id,
+      resource_id: vm.id,
+      resource_name: vm.name,
+      billing_rate_id: BillingRate.from_resource_properties("VmStorage", vm.family, vm.location)["id"],
+      amount: storage_volumes.sum { _1["size_gib"] }
```
enescakir, line 342: nit: 
```suggestion
      amount: vm.storage_size_gib
```

prog/vm/nexus.rb
```diff
@@ -334,6 +334,14 @@ def write_params_json
       amount: vm.cores
     )
 
+    BillingRecord.create_with_id(
+      project_id: project.id,
+      resource_id: vm.id,
```
enescakir, line 339: Just an idea. Should we create individual billing records for each VmStorageVolume? They may use different storage backends and have varying prices in the future.

spec/routes/web/vm_spec.rb
```diff
@@ -71,6 +71,7 @@
         uncheck "Enable Public IPv4"
         choose option: "ubuntu-jammy"
         choose option: "standard-2"
+        find(".storage-slider").set(25)
```
enescakir: It's not crucial, but I believe this should be included in the next frontend commit, as I didn't see `.storage-slider` until this one.

enescakir (commented):
Thanks for the separate commits. They make reviews much easier.

config/billing_rates.yml
```diff
@@ -1,45 +1,45 @@
...
-- { id: 078d2f47-5827-49fa-9d13-bee6d1d8cadd, resource_type: GitHubRunnerMinutes,     resource_family: standard-4-arm,  location: global,         unit_price: 0.0016 }
-- { id: 0fc7f264-87f0-446d-bee7-c0454149b1bc, resource_type: GitHubRunnerMinutes,     resource_family: standard-8-arm,  location: global,         unit_price: 0.0032 }
-- { id: 995ab760-5d97-4a7b-981e-c9947233c78e, resource_type: GitHubRunnerMinutes,     resource_family: standard-16-arm, location: global,         unit_price: 0.0064 }
-- { id: 27d84309-e39d-8d78-b603-7632dafe374e, resource_type: GitHubRunnerMinutes,     resource_family: standard-gpu-6,  location: global,         unit_price: 0.0320 }
-- { id: c31b3640-92fa-49c3-bae8-c1011cc5f917, resource_type: GitHubRunnerConcurrency, resource_family: standard,        location: global,         unit_price: 0.0000248016 }
+- { id: 139d9a67-8182-8578-a303-235cabd5161c, resource_type: VmCores,                 resource_family: standard,        location: hetzner-fsn1,   unit_price: 0.0006227680, active_from: 2023-01-01t00:00:00Z }
```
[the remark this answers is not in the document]
byucesoy: No specific reason. Both are accepted, but I changed to uppercase since that is more common.

routes/clover_base.rb
```diff
@@ -65,10 +65,10 @@ def fetch_location_based_prices(*resource_types)
     # that users won't see higher price in their invoice compared
     # to price calculator and also we charge same amount no matter
     # the number of days in a given month.
-    BillingRate.rates.filter { resource_types.include?(_1["resource_type"]) }
+    BillingRate.rates.filter { resource_types.include?(_1["resource_type"]) }.sort_by { _1["active_from"] }
```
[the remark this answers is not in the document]
byucesoy: We already return the latest billing rate, because duplicate entries overwrite each other while generating the hash below and thanks to sort, latest billing rate performs the latest overwrite.

However, I understand this is not obvious at a first glance. So I'm adding an explicit filter here.

prog/vm/nexus.rb
byucesoy: Makes sense.

prog/vm/nexus.rb
byucesoy: This is not needed anymore since we are creating separate billing records for each volume

prog/vm/nexus.rb
```diff
@@ -336,6 +336,16 @@ def write_params_json
       amount: vm.cores
     )
 
+    storage_volumes.each do |vol|
+      BillingRecord.create_with_id(
+        project_id: project.id,
+        resource_id: vm.id,
+        resource_name: vm.name,
+        billing_rate_id: BillingRate.from_resource_properties("VmStorage", vm.family, vm.location)["id"],
+        amount: vol["size_gib"]
```
enescakir, line 345: Should we change `resource_id` to storage volume ID, or should we rename `resource_name`? At present, if a VM has two disk volumes of the same size, they are indistinguishable.

enescakir (approved):

prog/vm/nexus.rb
byucesoy: I changed the resource_name and added disk_index to be able to distinguish different disks.

PR #1620, opened 2024-05-21
Merge some Web and API serializers

byucesoy:
Web and API serializers are quite similar to each other, especially Nic, FirewallRule and PostgresFirewallRule. The only difference in these serializers is that API serializers use `ubid` as `id` and web serializers use `id` as `id`. If we use `ubid` as `id` in our front end code, there wouldn't be any difference and both serializers can be merged into one, which is what we do in this PR. Since the API serializers are the external facing ones, I primarily used them as the base for new common serializers.

I'm planning to merge other serializers as well, but they are bit more diverged, so I'm leaving them to a separate PR. The primary difference there is that web serializers has extra `path` field, which is heavily used in front end code. I think we can add `path` to external APIs as well, because that information can be useful for users too, but I didn't want to do that without discussing it with @velioglu first.

Finally this PR nets -50 lines :partying_face:  and more to come :slightly_smiling_face:


enescakir (approved):
nets -50 🚀 huh

furkansahin (approved):

PR #1621, opened 2024-05-21
Inline serializers in roda branches

byucesoy:
We used to define a global serializer at the top of the roda branch, and then calls to `serialize` would use that serializer. This was a bit confusing for few reasons;

1. In some cases, we use different serializers in same branch. To be able to use second serializer, we need to explicitly call define which serializer to use, however for the default case we don't need to do that. At a glance, it's hard to understand which serializer would be used while resolving `serialize` calls.
2. In most cases, we define global serializer but use it only once.

Overall, I think it is better to explicitly set the serializer at the time of the serialization, not at the top of the branch.

As an extra benefit, this also reduces the line count by 42.

### DIFF
--- a/routes/api/project.rb
+++ b/routes/api/project.rb
@@ -2,8 +2,6 @@
 
 class CloverApi
   hash_branch("project") do |r|
-    @serializer = Serializers::Api::Project
-
     r.get true do
       result = Project.authorized(@current_user.id, "Project:view").where(visible: true).paginated_result(
         start_after: r.params["start_after"],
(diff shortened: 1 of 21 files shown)


fdr (approved):
I do like shorter. I don't think there's a deep reason to use the instance variable pattern that *I* know about. We can always reintroduce when there's something tangible.

I suppose it moves us away from coupling the serialized representation from a branch to one another, but as-is we're getting negative line/expression economy. Maybe later we can write a plugin that will offer more, such as schema compliance checking as well as cleaning up some expressions.

PR #1626, opened 2024-05-22
Merge web and api serializers

byucesoy:
Polymorphic authorization checks (see first commit) and options based serializers made serializers powerful enough to format slightly different outputs based on the needs of web and api endpoints. We can finally merge them. We also won't need to define two serializers per each new resource in the future.

Also, this cuts yet another 170 lines :slightly_smiling_face:

I've been working on improving serializer code at #1620, #1621, #1622 and #1626 (this PR)  and as a result of that, we deleted 385 lines :partying_face:


lib/authorization.rb
```diff
@@ -37,7 +37,11 @@ def self.expand_actions(actions)
 
   def self.matched_policies(subject_id, actions = nil, object_id = nil)
     object_filter = if object_id
-      Sequel.lit("AND object_applied_tags.tagged_id = ?", object_id)
+      begin
+        Sequel.lit("AND object_applied_tags.tagged_id = ?", UBID.parse(object_id).to_uuid)
+      rescue UBIDParseError
+        Sequel.lit("AND object_applied_tags.tagged_id = ?", object_id)
+      end
```
enescakir, line 44: I liked the overall concept and motivation. Previously, we discussed this topic in the context of adding `Model[UBID]` support to our models. But the tricky part is determining whether the given argument is ubid or not. We can try to parse and assume it's not ubid if fails like in this commit.
Since authorization check by UUID is also very popular operation in the codebase, I'm not sure about adding a new branch here for UBID as first option in the case of performance. We might assume all 26 character long strings are UBID, since we use UUID as primary keys only and they are longer. cc @fdr and @bsatzger they had some thought about this topic.

byucesoy: Initially, I also had similar concern and thought about checking length as you suggested, but this is already what `UBID.parse` does at the first step. So I don't think we lose much cycles here with `UBID.parse`.

enescakir:
Can you also remove the serializer subfolders from the loader?
https://github.com/ubicloud/ubicloud/blob/82fc2280d2411bc9af9d0a95904e88eabcb0a0f7/loader.rb#L82

lib/authorization.rb
fdr: IMO checking for 26 chars is pretty darn cheap. I doubt it even allocates. I don't think we even do `Model#[]` *that* often.

fdr: But if you want to make sure....run `hyperfine` on the tests

fdr: If that doesn't work, we have one more trick we can pull: dynamically override `#[]` to add the branch only when the REPL is loaded, since it's mostly for better annotation/transcripts of our operations actions. I think the benefit here is, in fact, material.

byucesoy:
> Can you also remove the serializer subfolders from the loader?
> 
> https://github.com/ubicloud/ubicloud/blob/82fc2280d2411bc9af9d0a95904e88eabcb0a0f7/loader.rb#L82

Done :+1:

lib/authorization.rb
fdr: Yeah, that's right, w.r.t. [ubid]

fdr (approved):
I like the code reduction, @enescakir @velioglu , any more pointed ideas about this? I'm not sure if my sensitivity to subtle effects on how the programming works is particularly good here.

enescakir (approved):

PR #1630, opened 2024-05-23
Introduce standard-30 VM size and surface "out of capacity" issues to users

bsatzger:
A VM's display state is set to `waiting for capacity` if allocation is not able to find any eligible host. It's switched back to `creating` when a suitable host was found.

Add `standard-30` VM size as well as these new runner labels:
* ubicloud-standard-30
* ubicloud-standard-30-ubuntu-2204
* ubicloud-standard-30-ubuntu-2004
* ubicloud-standard-30-arm
* ubicloud-standard-30-arm-ubuntu-2204
* ubicloud-standard-30-arm-ubuntu-2004

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -8,7 +8,7 @@ require "base64"
 
 class Prog::Vm::Nexus < Prog::Base
   subject_is :vm
-  semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup, :update_spdk_dependency
+  semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup, :update_spdk_dependency, :waiting_for_capacity
 
   def self.assemble(public_key, project_id, name: nil, size: "standard-2",
     unix_user: "ubi", location: "hetzner-hel1", boot_image: Config.default_boot_image_name,
@@ -216,6 +216,7 @@ class Prog::Vm::Nexus < Prog::Base
     rescue RuntimeError => ex
       raise unless ex.message.include?("no space left on any eligible host")
 
+      incr_waiting_for_capacity unless vm.waiting_for_capacity_set?
       queued_vms = queued_vms.all
       Prog::PageNexus.assemble("No capacity left at #{vm.location} for #{vm.arch}", queued_vms.first(25).map(&:ubid), "NoCapacity", vm.location, vm.arch)
       Clog.emit("No capacity left") { {lack_of_capacity: {location: vm.location, arch: vm.arch, queue_size: queued_vms.count}} }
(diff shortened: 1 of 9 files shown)


byucesoy (approved):

migrate/20240523_vm_display_state_waiting_for_capacity.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    add_enum_value(:vm_display_state, "waiting for capacity")
```
enescakir, line 5: I have a plan to remove `display_state` column and make it dynamic similar to the `display_state` of `PostgresResource` https://github.com/ubicloud/ubicloud/blob/82fc2280d2411bc9af9d0a95904e88eabcb0a0f7/model/postgres/postgres_resource.rb#L45.

Is there an easy way to add this state here: https://github.com/ubicloud/ubicloud/blob/82fc2280d2411bc9af9d0a95904e88eabcb0a0f7/model/vm.rb#L54

bsatzger: that should be possible. one way would be to introduce a semaphore that is incremented when we cannot allocate a host for the vm, and decremented again after successful allocation.

enescakir (approved):

migrate/20240523_vm_display_state_waiting_for_capacity.rb
enescakir: I don't think it's necessary for this PR. It's good to see it's doable. We can do it when we remove the `display_state` column.

prog/vm/nexus.rb
```diff
@@ -428,6 +432,9 @@ def write_params_json
       hop_prevent_destroy
     end
 
+    when_waiting_for_capacity_set? do
```
enescakir, line 435: We already clean up semaphores while destroying the strand, so there's no need to manually decrease it here.

enescakir: I see, you did it because semaphore check has more precedence than the column value

enescakir: We might change "deleting" state similar to the PG resource

```ruby
return "deleting" if destroy_set? || strand.label == "destroy"
```

prog/vm/nexus.rb
```diff
@@ -216,13 +216,17 @@ def before_run
...
 
+      incr_waiting_for_capacity unless vm.waiting_for_capacity_set?
       queued_vms = queued_vms.all
       Prog::PageNexus.assemble("No capacity left at #{vm.location} for #{vm.arch}", queued_vms.first(25).map(&:ubid), "NoCapacity", vm.location, vm.arch)
       Clog.emit("No capacity left") { {lack_of_capacity: {location: vm.location, arch: vm.arch, queue_size: queued_vms.count}} }
 
       nap 30
     end
 
+    when_waiting_for_capacity_set? do
```
enescakir, line 227: Should we decrement it without `when_waiting_for_capacity_set?` check?

enescakir (approved):

PR #1633, opened 2024-05-24
Update a VMs IPv6 prefix

furkansahin:
## Add IPv6 reassignment functionality 
This is required because Hetzner decided to renew the IPv6 Prefixes
assigned to our hosts. Therefore, we need a way to update the IPv6
prefix of the VMs running on the impacted hosts. The action plan is
available at
https://github.com/ubicloud/ubinest/wiki/Hetzner-IPv6-Prefix-Change-%E2%80%90-Action-Plan

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -136,38 +136,8 @@ class Prog::Vm::Nexus < Prog::Base
     @host ||= vm.vm_host
   end
 
-  def local_ipv4
-    vm.local_vetho_ip&.to_s&.shellescape || ""
-  end
-
   def params_path
     @params_path ||= File.join(vm_home, "prep.json")
-  end
-
-  def storage_volumes
-    @storage_volumes ||= vm.vm_storage_volumes.map { |s|
-      {
-        "boot" => s.boot,
-        "image" => s.boot ? vm.boot_image : nil,
-        "image_version" => s.boot_image&.version,
-        "size_gib" => s.size_gib,
-        "device_id" => s.device_id,
-        "disk_index" => s.disk_index,
-        "encrypted" => !s.key_encryption_key_1.nil?,
-        "spdk_version" => s.spdk_version,
-        "use_bdev_ubi" => s.use_bdev_ubi,
-        "skip_sync" => s.skip_sync,
-        "storage_device" => s.storage_device.name
-      }
-    }
-  end
-
-  def storage_secrets
-    @storage_secrets ||= vm.vm_storage_volumes.filter_map { |s|
-      if !s.key_encryption_key_1.nil?
-        [s.device_id, s.key_encryption_key_1.secret_key_material_hash]
-      end
-    }.to_h
   end
 
   def clear_stack_storage_volumes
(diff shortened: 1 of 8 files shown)


prog/update_vm_ipv6.rb
```diff
@@ -0,0 +1,86 @@
...
+      "nics" => vm.nics.map { |nic| [nic.private_ipv6.to_s, nic.private_ipv4.to_s, nic.ubid_to_tap_name, nic.mac] },
+      "boot_image" => vm.boot_image,
+      "max_vcpus" => topo.max_vcpus,
+      "cpu_topology" => topo.to_s,
+      "mem_gib" => vm.mem_gib,
+      "ndp_needed" => vm_host.ndp_needed,
+      "storage_volumes" => storage_volumes,
+      "swap_size_bytes" => nil,
+      "pci_devices" => vm.pci_devices.map { [_1.slot, _1.iommu_group] }
+    })
```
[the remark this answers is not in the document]
furkansahin: I was actually thinking this would either be used a single time by an operator and then get cleaned up, or as an example script. If you think we can keep it around and productionize, I can work on it more.

prog/vm/update_ipv6.rb
```diff
@@ -0,0 +1,44 @@
...
+    vm_host.sshable.cmd("sudo host/bin/setup-vm reassign-ip6 #{vm.inhost_name}", stdin: JSON.generate({storage: vm.storage_secrets}))
+    hop_start_vm
+  end
+
+  label def start_vm
+    nic = vm.nics.first
+    addr = nic.private_subnet.net4.nth(1).to_s + nic.private_subnet.net4.netmask.to_s
+
+    vm_host.sshable.cmd("sudo ip -n #{vm.inhost_name.shellescape} addr replace #{addr} dev #{nic.ubid_to_tap_name}")
+    vm_host.sshable.cmd("sudo systemctl start #{vm.inhost_name}.service")
```
byucesoy, line 34: Do we need to start dnsmasq service as well?

prog/vm/nexus.rb
```diff
@@ -272,29 +242,7 @@ def before_run
...
-      "cpu_topology" => topo.to_s,
-      "mem_gib" => vm.mem_gib,
-      "ndp_needed" => host.ndp_needed,
-      "storage_volumes" => storage_volumes,
-      "swap_size_bytes" => frame["swap_size_bytes"],
-      "pci_devices" => vm.pci_devices.map { [_1.slot, _1.iommu_group] }
-    })
-
-    host.sshable.cmd("sudo -u #{q_vm} tee #{params_path.shellescape}", stdin: params_json)
+    host.sshable.cmd("sudo -u #{q_vm} tee #{params_path.shellescape}", stdin: vm.params_json(frame["swap_size_bytes"]))
```
byucesoy, line 245: safe_write_to_file?

byucesoy (approved):

prog/vm/update_ipv6.rb
furkansahin: Nope, since VM service depends on it, it auto starts if not already started.

PR #1636, opened 2024-05-25
Add FF to allow some projects create Github Installation wo billing info

furkansahin:
We would like to allow list some projects to use github runners without entering billing info. This PR introduces a new feature flag to handle that.

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -94,5 +94,5 @@ class Project < Sequel::Model
     end
   end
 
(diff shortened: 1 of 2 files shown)


byucesoy (commented):
What happens when we try to generate invoices for these customers? We should still generate invoices if we get paid eventually either by project owner or their sponsor. I'm not sure what needs to be put on the invoice as customer information though.

PR #1637, opened 2024-05-26
Allocate primaries and standbys at different hosts 

byucesoy:
**Create standbys in PostgresResource's wait state**
We have very similar standby creation logic in two places: assemble and wait.
When we need to make a change on the standby creation logic, we need to update
two places. We can consolidate the standby creation logic into the wait state
since all creations would eventually reach there. One drawback of this is we
would create standbys ~2-3 minutes later than before, but it is a small time
window and it is at the beginning of the resource's lifecycle, hence not very
critical.

**Create standbys at different hosts than primary**
We recently added ability to exclude hosts during allocation. This change uses
that new ability to create primary and standbys at different hosts. This is to
ensure if a host goes down, we still have either standby or primary available.

### DIFF
--- a/prog/postgres/postgres_resource_nexus.rb
+++ b/prog/postgres/postgres_resource_nexus.rb
@@ -158,9 +158,10 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
   end
 
   label def wait
-    # Create missing standbys
-    (postgres_resource.required_standby_count + 1 - postgres_resource.servers.count).times do
-      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", private_subnet_id: representative_server.vm.private_subnets.first.id)
+    # Only create one standby at a time to ensure that they are allocated on different hosts
+    if postgres_resource.required_standby_count + 1 > servers.count && servers.none? { _1.vm.vm_host.nil? }
+      exclude_host_ids = Config.development? ? [] : servers.map { _1.vm.vm_host.id }
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", private_subnet_id: representative_server.vm.private_subnets.first.id, exclude_host_ids: exclude_host_ids)
     end
 
     when_refresh_dns_record_set? do
--- a/prog/postgres/postgres_resource_nexus.rb
+++ b/prog/postgres/postgres_resource_nexus.rb
@@ -49,10 +49,7 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
 
       PostgresFirewallRule.create_with_id(postgres_resource_id: postgres_resource.id, cidr: "0.0.0.0/0")
 
-      primary = Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: timeline_access, representative_at: Time.now).subject
-      postgres_resource.required_standby_count.times do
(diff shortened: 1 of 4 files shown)


prog/postgres/postgres_server_nexus.rb
```diff
@@ -31,7 +31,8 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
         boot_image: postgres_resource.project.get_ff_postgresql_base_image || "postgres-ubuntu-2204",
         private_subnet_id: private_subnet_id,
         enable_ip4: true,
-        allow_only_ssh: true
+        allow_only_ssh: true,
+        exclude_host_ids: exclude_host_ids
```
bsatzger, line 35: In the future, should we also exclude the host running `postgres-monitoring`?

bsatzger (approved):

prog/postgres/postgres_server_nexus.rb
byucesoy: I think that makes sense. Though the problem is that `postgres-monitoring` DB also has HA, which means it would block 2 hosts for everyone and the hosts its servers are in would change after a failover. Maybe we should pin the monitoring DB to two specific hosts. We can do that after we have more hosts available for PG use. Currently we are running near capacity for PG/VM workloads. Since this is not meant to be blocker, I'll resolve this and return back in the future.

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -49,10 +49,7 @@ def self.assemble(project_id:, location:, name:, target_vm_size:, target_storage
 
       PostgresFirewallRule.create_with_id(postgres_resource_id: postgres_resource.id, cidr: "0.0.0.0/0")
 
-      primary = Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: timeline_access, representative_at: Time.now).subject
-      postgres_resource.required_standby_count.times do
-        Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: "fetch", private_subnet_id: primary.vm.private_subnets.first.id)
-      end
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: timeline_access, representative_at: Time.now).subject
```
enescakir, line 52: nit: no need to subject
```suggestion
      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: timeline_access, representative_at: Time.now)
```

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -161,9 +158,9 @@ def before_run
   end
 
   label def wait
-    # Create missing standbys
-    (postgres_resource.required_standby_count + 1 - postgres_resource.servers.count).times do
-      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", private_subnet_id: representative_server.vm.private_subnets.first.id)
+    # Only create one standby at a time to ensure that they are allocated on different hosts
+    if postgres_resource.required_standby_count + 1 > servers.count && servers.none? { _1.vm.vm_host.nil? }
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", private_subnet_id: representative_server.vm.private_subnets.first.id, exclude_host_ids: servers.map { _1.vm.vm_host.id })
```
enescakir, line 163: Should we add an exclusion for the development? If we only have one host, can we still provision a DB with HA?

PR #1638, opened 2024-05-27
Decouple PG compute and storage

byucesoy:
**Update PostgreSQL storage prices for US regions**
The cost of storage in the US regions is higher than EU regions, so we are
increasing the price of PostgreSQL storage in the US regions to reflect this.
Since these rates are not used previously, we can update them directly without
adding a new rate version.

**Add backend code allowing picking storage independently from PG size**

**Automatically select first options in PostgreSQL creation form**
This ensures that the javascript code depending on instance size works even at
the initial load. It also makes the form more user-friendly by requiring one
less click.

**Add frontend code allowing picking storage independently from PG size**

**Reorder form updates to make HA calculations at the end**
Since HA price depends on other selections, it should be calculated after
calculations based on previous selections are done. Otherwise, the price we
list in the form would be incorrect.

**Cache and set the value of slider while changing its properties**
Value in the slider might be invalid temporarily while changing min/max/step.
In such cases browser assigns a valid value automatically, which is not
necessarily the value we would pick. To prevent this, we read the value of the
slider before changing its properties and set it back after changing them.

image

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -272,10 +272,11 @@ function setupInstanceSizeBasedOptions() {
 
       let monthlyPrice = $("input[name=location]:checked").data("details")[storage_resource_type]["standard"]["monthly"]
       $(this).data("monthly-price", monthlyPrice);
-      $(this).next().find(".absolute").each(function(idx) {
-        storage_amount = min_storage_size_gib + idx * storage_size_step_gib;
-        $(this).text(storage_amount + "GB +$" + (storage_amount * monthlyPrice).toFixed(2));
-      });
+      $(this).next().html("");
+      for(var storage_amount = min_storage_size_gib; storage_amount <= max_storage_size_gib; storage_amount += storage_size_step_gib) {
+        let content = storage_amount + "GB<br/>+$" + (storage_amount * monthlyPrice).toFixed(2);
+        $(this).next().append("<li class='flex justify-right sm:justify-center relative items-center rotate-90 sm:rotate-0 mb-6 sm:mb-0'><span class='absolute text-center'>" + content + "</span></li>");
+      }
     }
   });
 }
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -4,7 +4,6 @@ $(function() {
   setupLocationBasedOptions();
   setupInstanceSizeBasedOptions();
   setupLocationBasedPostgresHaPrices();
-  setupLocationBasedPostgresHaPrices();
   setupAutoRefresh();
   setupPrint();
   setupDatePicker();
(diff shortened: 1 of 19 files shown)


lib/option.rb
```diff
@@ -51,11 +51,11 @@ def self.postgres_locations
     VmSize.new("standard-gpu-#{_1}", "standard-gpu", _1, (_1 * 5.34).to_i, (_1 / 2) * 60, _1 * 60, (_1 / 2) * 60, false, true)
   }).freeze
 
-  PostgresSize = Struct.new(:name, :vm_size, :family, :vcpu, :memory, :storage_size_gib) do
+  PostgresSize = Struct.new(:name, :vm_size, :family, :vcpu, :memory, :min_storage_size_gib, :max_storage_size_gib, :storage_size_step_gib) do
     alias_method :display_name, :name
   end
   PostgresSizes = [2, 4, 8, 16].map {
-    PostgresSize.new("standard-#{_1}", "standard-#{_1}", "standard", _1, _1 * 4, (_1 / 2) * 128)
+    PostgresSize.new("standard-#{_1}", "standard-#{_1}", "standard", _1, _1 * 4, (_1 / 2) * 128, _1 * 256, (_1 / 2) * 128)
```
enescakir, line 58: I prefer to standardize the styles for minimum and maximum values

```ruby
(_1 / 2) * 128, (_1 * 2) * 128
```
or
```ruby
_1 * 64, _1 * 256
```

assets/js/app.js
```diff
@@ -3,6 +3,8 @@ $(function() {
   setupLocationBasedPrices();
   setupLocationBasedOptions();
   setupInstanceSizeBasedOptions();
+  setupLocationBasedPostgresHaPrices();
+  setupLocationBasedPostgresHaPrices();
```
enescakir, line 7: Why do we call it twice? 😄

enescakir:
Could you please add a screenshot for future reference to document what we've done here?

byucesoy:
> Could you please add a screenshot for future reference to document what we've done here?
Done :+1:

lib/option.rb
```diff
@@ -51,11 +51,11 @@ def self.postgres_locations
     VmSize.new("standard-gpu-#{_1}", "standard-gpu", _1, (_1 * 5.34).to_i, (_1 / 2) * 60, _1 * 60, (_1 / 2) * 60, false, true)
   }).freeze
 
-  PostgresSize = Struct.new(:name, :vm_size, :family, :vcpu, :memory, :storage_size_gib) do
+  PostgresSize = Struct.new(:name, :vm_size, :family, :vcpu, :memory, :min_storage_size_gib, :max_storage_size_gib, :storage_size_step_gib) do
```
furkansahin, line 54: Instead of min, max and step, should we simply list the available sizes in an array? I think it's cleaner

lib/validation.rb
```diff
@@ -113,6 +113,14 @@ def self.validate_postgres_size(size)
     postgres_size
   end
 
+  def self.validate_postgres_storage_size(size, storage_size)
+    storage_size = storage_size.to_i
+    pg_size = validate_postgres_size(size)
+    allowed_sizes = (pg_size.min_storage_size_gib..pg_size.max_storage_size_gib).step(pg_size.storage_size_step_gib)
```
furkansahin, line 119: if we create the PostgresSize struct with this array provided already, we don't need to programmatically calculate this here.

furkansahin (approved):

enescakir (approved):

PR #1639, opened 2024-05-27
Enable standard-60 vm size with 60 vCPUs

bsatzger:
This PR enables standard-60 vm size, which comes with 60 vCPUs.

### DIFF
--- a/spec/model/vm_spec.rb
+++ b/spec/model/vm_spec.rb
@@ -22,22 +22,12 @@ RSpec.describe Vm do
   end
 
   describe "#mem_gib" do
-    it "handles standard-2" do
+    it "handles standard family" do
       vm.family = "standard"
-      vm.cores = 1
-      expect(vm.mem_gib).to eq 8
-    end
-
-    it "handles standard-16" do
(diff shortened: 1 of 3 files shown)


spec/model/vm_spec.rb
```diff
@@ -40,6 +40,12 @@
       expect(vm.mem_gib).to eq 120
     end
 
+    it "handles standard-60" do
```
byucesoy, line 43: Not sure if we need 4 tests for the same functionality as they are coming from same formula, but it you want to keep the tests in case for possible formula split in the future you can do this instead.

```ruby
it "handles mem_gib calculation" do
  vm.family = "standard"
  [1, 8 , 15, 30].each do |cores|
    expect(vm).to receive(:cores).and_return(cores)
    expect(vm.mem_gib).to eq cores * 8
  end
end
```

byucesoy (approved):

enescakir (approved):

enescakir (approved):
Adding a new vm size while reducing the code 🚀  Fantastic.

PR #1640, opened 2024-05-27
Start using new firmware

bsatzger:
Start using `NEW_FIRMWARE` in vm setup. The new firmware was downloaded to all existing hosts. I plan to `install_rhizome` first to one `x64` and one `arm64` runner host. If VM provisioning looks fine, I'll `install_rhizome` to the rest of the fleet. If all looks good for at least a week, I'll remove the legacy firmware code.


fdr (approved):
Nice, although, in retrospect, I wish that we renamed CloudHypervisor::FIRMWARE to LEGACY_FIRMWARE (as with the class names), to skip some renaming churn when removing the legacy code. No big deal though.

PR #1642, opened 2024-05-28
Add Ubuntu 24.04 support

enescakir:
### Add Ubuntu 24.04 LTS support

Ubuntu 24.04 (Noble Numbat) was launched in April 2024 as the latest LTS release following Ubuntu 22.04. It comes with 5 years of support, ending in April 2029.

https://ubuntu.com/about/release-cycle

Like other cloud providers, we have added support for Ubuntu 24.04 LTS to our platform.

I tested the new image and it works as expected.

    ubi@vmgsy6jv:~$ lsb_release -a
    No LSB modules are available.
    Distributor ID: Ubuntu
    Description:    Ubuntu 24.04 LTS
    Release:        24.04
    Codename:       noble

    ubi@vmgsy6jv:~$ uname -r
    6.8.0-31-generic
    

~~### Test the boot images exposed to customers~~

~~Since we disabled Alma Linux 9, our customers can only provision Ubuntu 22.04 virtual machines. However, with the launch of Ubuntu 24.04, customers can now provision virtual machines with two different images. We need to test all the boot images exposed to customers to ensure they function as expected. In the past, we've encountered issues with various boot images that we couldn't detect in time.  Therefore, it's crucial to include them in our E2E tests.~~

~~The VmGroup E2E test provisions 3 virtual machines and checks basic features like network connectivity. I've updated it to provision virtual machines using a boot image randomly selected from the available options. This way, we can test different combinations of boot images.~~

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -131,6 +131,7 @@ module Config
   optional :ubicloud_images_blob_storage_secret_key, string, clear: true
   optional :ubicloud_images_blob_storage_certs, string
 
+  override :ubuntu_noble_version, "20240523.1", string
   override :ubuntu_jammy_version, "20240319", string
   override :almalinux_9_version, "9.4-20240507", string
   override :almalinux_8_version, "8.10-20240530", string
--- a/lib/option.rb
+++ b/lib/option.rb
@@ -39,6 +39,7 @@ module Option
 
   BootImage = Struct.new(:name, :display_name)
   BootImages = [
+    ["ubuntu-noble", "Ubuntu Noble 24.04 LTS"],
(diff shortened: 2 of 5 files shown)


byucesoy (approved):

furkansahin (approved):

fdr (approved):
nice commit message (that includes the uname output for handy reference)

enescakir:
In another PR, we're enabling AlmaLinux 8 and 9. Selecting random boot images for E2E tests can cause some commands to fail for AlmaLinux boot images. Therefore, I've moved the E2E test modifications to a different PR. This current one is solely for enabling Ubuntu 24.04.

PR #1644, opened 2024-05-28
Fix broken billing record index

byucesoy:
We have a partial unique constraint on the resource_id of the billing_record table. However, the index was created with wrong condition. Its condition is `WHERE span(upper) = NULL``, which is always false, because in the SQL standard comparison against NULL always returns NULL which is evaluated as false. The correct form should have been `WHERE span(upper) is NULL`. This means that the index is not used at all. We didn't realized this, because we stopped caring about uniqueness of resource_id in the billing_record table for a while as one resource can have multiple active billing records (for different parts of the resource such as compute and storage).

As a fix, I'm recreating the index without the condition. Also I put noop for down migration, because the previous state of the DB also had the index (alas with wrong condition).


migrate/20240529_fix_broken_index.rb
```diff
@@ -0,0 +1,17 @@
...
+    alter_table(:billing_record) do
+      drop_index :resource_id, concurrently: true
+      add_index :resource_id, concurrently: true
+    end
+  end
+
+  down do
+    alter_table(:dns_record) do
+    end
+  end
```
enescakir, line 16: I think you can remove it at all

```suggestion
```

migrate/20240529_fix_broken_index.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
+
+  up do
+    alter_table(:billing_record) do
+      drop_index :resource_id, concurrently: true
+      add_index :resource_id, concurrently: true
```
furkansahin, line 9: Whoah, this looks very weird :D Can we at least add a comment here about the old index's format being different?

byucesoy: migration files are very rarely read and I added a detailed explanation to commit message. Is it OK to keep this as is?

furkansahin: sure

enescakir (approved):

PR #1645, opened 2024-05-28
Do not delete PrivateSubnet during deletion of PostgresServer

byucesoy:
The PrivateSubnet is a shared resource and should not be deleted when the PostgresServer is deleted. It can only be deleted when parent resource, PostgresResource is deleted.

### DIFF
--- a/prog/postgres/postgres_resource_nexus.rb
+++ b/prog/postgres/postgres_resource_nexus.rb
@@ -187,6 +187,7 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
 
     strand.children.each { _1.destroy }
     unless servers.empty?
+      servers.first.vm.private_subnets.each(&:incr_destroy)
       servers.each(&:incr_destroy)
       nap 5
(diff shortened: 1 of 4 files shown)


prog/postgres/postgres_resource_nexus.rb
```diff
@@ -187,6 +187,7 @@ def before_run
 
     strand.children.each { _1.destroy }
     unless servers.empty?
+      servers.first.vm.private_subnets.each(&:incr_destroy)
```
bsatzger, line 190: is it enough in all circumstances to delete private subnets of the first server?

bsatzger (approved):

prog/postgres/postgres_resource_nexus.rb
byucesoy: Yes, because they all share same private subnet. Also the first server is not special here, for all purposes we could use `servers.sample.vm.private_subnets.each(&:incr_destroy)`

PR #1651, opened 2024-05-30
Increase the IPv6 entropy to 2 bytes for hosts with /64

furkansahin:
The upper range is updated to increase the entropy if the host has a big enough prefix. This way, we avoid collisions. Keeping the range to a single byte caused elevated number of collisions.


fdr (approved):
I was wondering when that would become a problem, and if we would have to something smart to identify free space among 256 elements or so, which could be seen as an alternative approach.

We may have to do that smart thing some day, but for today, this nicely avoids the problem when it doesn't need to be a problem. Thank you.

enescakir:
A unit test failed with this error once 🤯 

https://github.com/ubicloud/ubicloud/actions/runs/9346099210/job/25720248705?pr=1642

```
  1) Scheduling::Allocator update fails concurrent allocations if memory constraints are violated
     Failure/Error: expect { al2.update(vm2) }.to raise_error(Sequel::CheckConstraintViolation, /hugepages_allocation_limit/)

       expected Sequel::CheckConstraintViolation with message matching /hugepages_allocation_limit/, got #<Sequel::ValidationFailed: ephemeral_net6 is already taken> with backtrace:
         # ./vendor/bundle/ruby/3.2.0/gems/sequel-5.80.0/lib/sequel/model/base.rb:1497:in `save'
         # ./vendor/bundle/ruby/3.2.0/gems/sequel-5.80.0/lib/sequel/model/base.rb:1513:in `save_changes'
         # ./vendor/bundle/ruby/3.2.0/gems/sequel-5.80.0/lib/sequel/model/base.rb:2104:in `update_restricted'
         # ./vendor/bundle/ruby/3.2.0/gems/sequel-5.80.0/lib/sequel/model/base.rb:1614:in `update'
         # ./scheduling/allocator.rb:105:in `update_vm'
         # ./scheduling/allocator.rb:133:in `block in update'

```

PR #1652, opened 2024-05-30
Reapply "Start using new firmware"

bsatzger:
We ran into an issue with the x64 version of the new firmware. VMs with GPU passthrough wouldn't boot with edk2-stable202402. We changed the x64 version to edk2-stable202311 in https://github.com/ubicloud/ubicloud/commit/d3b06b58b799a166c71c2f0f9a9f41397be23bba as we tracked the issue down to this commit https://github.com/tianocore/edk2/commit/e51965ddd1816a0506818e66644a635ae830e20f, and edk2-stable202311 is the most recent release without it.

We should be good to start using the new firmware.

This reverts commit 8090df77b4fd85f8a4cf174ac1392d2dac5af640.


byucesoy (approved):
I think it would be good to explain the problem we hit in the commit message.

PR #1656, opened 2024-05-31
Some cleanups to introduce a new roda app

enescakir:
### Clean up invalid CSRF exception handling

The `parse_error` in `clover_base` expects the `Roda::RodaPlugins::RouteCsrf::InvalidToken` to be defined. This is true if the `route_csrf` plugin is enabled, but it might not be in some cases. Only `CloverWeb` uses CSRF protection and has an additional check for it. Thus, we can combine them and pass it during the plugin configuration.

### Clean up routes spec helpers

- We had a similar setup and required the same libraries in the sub-routes spec helpers. I've now moved them to the main routes spec helper.
- In the API test, we log in using query parameters. Who put the password to the query parameters? I've moved it to the body, aligning with how it's done in production.
- When running tests, Octokit displays a warning: "To use retry middleware with Faraday v2.0+, install `faraday-retry` gem". It seems there's a design oversight in the library. I added a warning ignore for it. For more information https://github.com/octokit/octokit.rb/discussions/1486

### DIFF
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -30,7 +30,11 @@ class CloverWeb < Roda
     csp.frame_ancestors :none
   end
 
-  plugin :route_csrf
+  plugin :route_csrf do |token|
+    flash["error"] = "An invalid security token submitted with this request, please try again"
+    return redirect_back_with_inputs
+  end
+
   plugin :disallow_file_uploads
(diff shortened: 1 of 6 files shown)


fdr (approved):
thank you for these fine adjustments.

PR #1657, opened 2024-05-31
Introduce runtime API endpoints

enescakir:
Generally, the controlplane has been passing necessary data to the dataplane at the start of operations. However, this approach can't pre-determine the information needed for future operations. There are instances where the control plane sends additional data to the dataplane during later stages, once it identifies the needed data. However, this polling method introduces delays. In some scenarios, the dataplane should have the capability to directly request this information from the controlplane during runtime, eliminating any delay. I needed this feature for implementing GitHub Actions Cache integration. The GitHub runner requires pre-signed blob storage URLs for uploading and downloading cache, a decision made at runtime. It calculates the required cache keys and versions. Rather than developing this feature solely for GitHub integration, I opted to create a more generic solution to identify the requesting virtual machine. This can be utilized for other integrations as well. The dataplane may need information from the controlplane or may need to trigger an operation on the controlplane.

When a request arrives at the controlplane, we must verify that it originates from our virtual machines and authenticate it. I selected JWT tokens for this authentication process, the same method we use for our customer API. We embed the UBID of the virtual machines into the JWT token. Thus, if the token signature is valid, we can identify the source virtual machines. One common criticism of JWT tokens is their statelessness; they can't be revoked by default. However, there are workarounds. Firstly, the token is only valid for the lifetime of the virtual machine. Once the runner is destroyed, the token is no longer valid. Additionally, we include a creation date in the token, allowing us to set an expiration time. I haven't yet implemented this expiration feature since the runners are typically short-lived. However, it can be easily added when we introduce runtime API endpoints for long-lived resources.

### DIFF
--- a/clover_runtime.rb
+++ b/clover_runtime.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require "jwt"
+
+class CloverRuntime < Roda
+  include CloverBase
+
+  plugin :default_headers, "Content-Type" => "application/json"
+
+  plugin :hash_branches
+  plugin :json
+  plugin :all_verbs
+  plugin :json_parser
+
+  autoload_routes("runtime")
+
+  plugin :error_handler do |e|
+    error = parse_error(e)
+
+    {error: error}.to_json
+  end
+
+  def get_jwt_payload(request)
+    return unless (v = request.env["HTTP_AUTHORIZATION"])
+    jwt_token = v.sub(%r{\ABearer:?\s+}, "")
+    begin
+      JWT.decode(jwt_token, Config.clover_runtime_token_secret, true, {algorithm: "HS256"})[0]
+    rescue JWT::DecodeError
+    end
+  end
+
+  route do |r|
+    if (jwt_payload = get_jwt_payload(r)).nil? || (@vm = Vm.from_ubid(jwt_payload["sub"])).nil?
+      fail CloverError.new(400, "InvalidRequest", "invalid JWT format or claim in Authorization header")
+    end
+
+    r.hash_branches("")
+  end
+end
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "jwt"
 require_relative "../model"
 
 class Vm < Sequel::Model
@@ -49,6 +50,10 @@ class Vm < Sequel::Model
 
   def ip4
     assigned_vm_address&.ip
+  end
+
+  def runtime_token
+    JWT.encode({sub: ubid, iat: Time.now.to_i}, Config.clover_runtime_token_secret, "HS256")
   end
 
   def display_state
--- a/spec/routes/runtime/spec_helper.rb
+++ b/spec/routes/runtime/spec_helper.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require_relative "../spec_helper"
+
+RSpec.configure do |config|
+  config.before {
+    allow(Config).to receive(:clover_runtime_token_secret).and_return(Config.clover_session_secret)
+  }
+end
+
+def login_runtime(vm)
+  header "Authorization", "Bearer #{vm.runtime_token}"
+end
--- a/clover.rb
+++ b/clover.rb
@@ -30,6 +30,10 @@ class Clover < Roda
     end
     # :nocov:
 
+    r.on "runtime" do
+      r.run CloverRuntime
+    end
+
     r.run CloverWeb
   end
 end
--- a/config.rb
+++ b/config.rb
@@ -47,6 +47,7 @@ module Config
   optional :versioning_default, string
   optional :versioning_app_name, string
   optional :clover_session_secret, base64, clear: true
+  optional :clover_runtime_token_secret, base64, clear: true
   optional :clover_column_encryption_key, base64, clear: true
   optional :stripe_public_key, string, clear: true
   optional :stripe_secret_key, string, clear: true
--- a/loader.rb
+++ b/loader.rb
@@ -78,7 +78,7 @@ autoload_normal = ->(subdirectory, include_first: false, flat: false) do
 end
 
 autoload_normal.call("model", flat: true)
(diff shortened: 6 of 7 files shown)


fdr:
It is good that you sign the principal involved, the implementation needs careful design to make it hard for a compromised VM or even host to influence others (as well as making an attacker stumbling around more noisy) 

Doesn't this impose severe development issues, though? It seems like anything that uses this channel is going to have to bring in ngrok or something like that. Do we already have an obligate SSH tunnel we could piggyback on? (cc: @byucesoy). Though, it may need work, to deal with thread limits and partitioning. Though we will have to do that anyway, the timing is not something I'm sure about: {now, soon, later}.

fdr (commented):
. (sorry for these, I sometimes write a comment, but want to wipe the review bit, so I need to leave another comment but differently)

furkansahin (approved):

spec/routes/runtime/spec_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+  config.before {
+    allow(Config).to receive(:clover_runtime_token_secret).and_return(Config.clover_session_secret)
+  }
+end
+
+def login_runtime
+  vm = Vm.create_with_id(display_state: "running", unix_user: "x", public_key: "x",
+    name: "x", family: "standard", cores: 2, location: "x", boot_image: "x")
+  header "Authorization", "Bearer #{vm.runtime_token}"
+  vm
```
velioglu, line 15: What about passing vm to this function and login with the passed vm's runtime_token (by also changing the function's name). Returning vm from login_runtime looks counter-intuitive to me.

spec/routes/runtime/spec_helper.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative "../spec_helper"
+
+RSpec.configure do |config|
+  config.before {
+    allow(Config).to receive(:clover_runtime_token_secret).and_return(Config.clover_session_secret)
```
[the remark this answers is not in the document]
enescakir: During project setup, new developers execute `rake overwrite_envrb` to generate `.env.rb`. This sets the `CLOVER_SESSION_SECRET` for local testing.
 
https://github.com/ubicloud/ubicloud/blob/d899b2a1719bdd973e11394733d79640c0b22364/Rakefile#L128

I didn't want to disrupt the local environment, so I used the same value for `Config.clover_runtime_token_secret` and `Config.clover_session_secret`.

spec/routes/runtime/spec_helper.rb
enescakir: Fair point. I change it

model/vm.rb
```diff
@@ -51,6 +52,10 @@ def ip4
     assigned_vm_address&.ip
   end
 
+  def runtime_token
+    JWT.encode({sub: ubid, iat: Time.now.to_i}, Config.clover_runtime_token_secret, "HS256")
```
bsatzger, line 56: If we add future use-cases, we could consider extending the subject (or use a different claim) to embed the use-case this claim is good for. For example, a claim with subject `vmabc123...-runnercache` would only be good for that use-case but not for anything else.

clover_runtime.rb
```diff
@@ -0,0 +1,39 @@
...
+    jwt_token = v.sub(%r{\ABearer:?\s+}, "")
+    begin
+      JWT.decode(jwt_token, Config.clover_runtime_token_secret, true, {algorithm: "HS256"})[0]
+    rescue JWT::DecodeError
+    end
+  end
+
+  route do |r|
+    if (jwt_payload = get_jwt_payload(r)).nil? || (@vm = Vm.from_ubid(jwt_payload["sub"])).nil?
+      fail CloverError.new(401, "InvalidCredentials", "There was an error logging in")
```
bsatzger, line 34: should we return `invalid JWT format or claim in Authorization header` instead of login error?

bsatzger: see, spec/routes/api/auth_spec.rb

```rb
  it "wrong jwt" do
    header "Authorization", "Bearer wrongjwt"
    get "/api/project"

    expect(last_response.status).to eq(400)
    expect(JSON.parse(last_response.body)["error"]["message"]).to eq("invalid JWT format or claim in Authorization header")
    expect(JSON.parse(last_response.body)["error"]["type"]).to eq("InvalidRequest")
  end
```

bsatzger (approved):

model/vm.rb
enescakir: Indeed, we can also include the scope value in the payload for potential future use cases.

clover_runtime.rb
enescakir: Should we return distinct messages for an invalid JWT and a non-existent VM, or should we use the same message? I tend not to be overly explicit in error messages to prevent data exposure. I could change the error message to  `invalid JWT format or claim in Authorization header` for the both cases.

Interestingly, our APIs return an HTTP 400 for an invalid JWT instead of a 401. I expected a 401 response for all JWT-related failures.

fdr (commented):
so one thing about the commit message doesn't hold up: we deal with this problem (stuff that isn't known at provision time) all the time, we do it via some kind of polling, e.g. when writing routinely run-time calculated STS tokens for database backups.

So while I think I'm okay with this approach in general, it doesn't really follow from the justification, do we really have zero post-provision branches anywhere else in the code base? That can't be right, surely? Normally  I think of doing something like this as more of a speedup. Or maybe a simple-up, in this case?

enescakir:
> so one thing about the commit message doesn't hold up: we deal with this problem (stuff that isn't known at provision time) all the time, we do it via some kind of polling, e.g. when writing routinely run-time calculated STS tokens for database backups.
> 
> So while I think I'm okay with this approach in general, it doesn't really follow from the justification, do we really have zero post-provision branches anywhere else in the code base? That can't be right, surely? Normally I think of doing something like this as more of a speedup. Or maybe a simple-up, in this case?

@fdr I've made a slight modification to the commit message. What do you think about the updated version? Do you have any recommendations?

fdr (approved):

PR #1659, opened 2024-06-03
Add retry logic for processes in E2E tests

byucesoy:
We realized that in some cases, respirate exits due to a crash or apoptosis, which causes foreman to exit and E2E tests to fail. Foreman is programmed to exit when one of its child processes exits. It doesn't have any retry logic. Recommended way of adding retry logic is combining it with a monitoring tool such as systemd or upstart. However, we don't need such a complex system for our E2E tests. Instead I opted to add retry logic with simple `until` loop.

I didn't add the retry logic for `bin/ci` process we are adding in next step of the workflow file, because tests themselves might not be written in a way that allows them to be executed second time.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -67,6 +67,10 @@ jobs:
     - name: Install foreman
       run: gem install foreman
 
+    - name: Modify Procfile to add retry logic
+      run: |
+        sed -i "s/^\([^:]*\): \(.*\)/\1: bash -c 'for i in {1..4}; do \2 \&\& break || echo \"Attempt \$i failed, retrying...\"; sleep 1; done'/" Procfile
+
     - name: Add e2e script to Procfile
       run: |
         echo "e2e: bin/ci --test-cases=${{ github.event_name != 'workflow_dispatch' && 'vm' || inputs.test_cases }}" >> Procfile


enescakir:
Do we always want to restart `respirate`? What happens if we introduce a bug to `respirate` that causes constant crashes, yet somehow the tests still manage to complete?

byucesoy:
> Do we always want to restart `respirate`? What happens if we introduce a bug to `respirate` that causes constant crashes, yet somehow the tests still manage to complete?

Hmm, fair. What do you think about adding a retry number. For example it restarts up to 3 times.

byucesoy:
I added retry limit.

enescakir (commented):
I agree that we should implement a retry mechanism, but I'm unsure if we should apply it to the entire respirate process. I believe the respirate process should never crash. I think we have about 2-3 labels that cause crashes. What's your opinion on adding a retry or exception handling mechanism to these steps? 

I'm concerned about obscuring the label that shouldn't crash behind this general retry mechanism.

byucesoy:
But respirate is designed to crash, due to its apoptosis logic. Crashing is its escape hatch. I think it is designed in a way to assume external process would restart it. In fact, some of the failures in E2E tests are due to apoptosis, so we have to add retry mechanism for entire respirate.

enescakir (approved):

PR #1661, opened 2024-06-03
Add support to upgrade Cloud Hypervisor

bsatzger:
**Introduce utility function to download a file**
`curl_file` downloads a file from a `url` and writes it to a `path` on
the local file system. It returns the SHA-256 digest of the downloaded
file.

**Leverage utility function to download firmware**
Use a utility function from rhizome lib to download firmware instead of
a custom `curl_firmware` method.

**Introduce new CloudHypervisor version class**
The new implementation supports downloading cloud hypervisor assets from
https://github.com/cloud-hypervisor/cloud-hypervisor. It downloads
`cloud-hypervisor-static` and `ch-remote-static` for x64 and
`cloud-hypervisor-static-aarch64` and `ch-remote-static-aarch64` for
arm64. It verifies the SHA-256 digest of these files during download.
The destination and naming pattern of the downloaded files does not
change, i.e. they are stored as:
`/opt/cloud-hypervisor/v#{version}/cloud-hypervisor`
`/opt/cloud-hypervisor/v#{version}/ch-remote`

**Download new Cloud Hypervisor version during host prep**
Start downloading new cloud hypervisor versions in addition to
legacy versions during host preparation. When legacy cloud
hypervisor has become obsolete, we will stop downloading it.

**Standalone executable to download Cloud Hypervisor**
`download-cloud-hypervisor` can be run independent of preparing
a host to download cloud hypervisor.

**Add Prog to download Cloud Hypervisor on-demand**
Introduce a `DownloadCloudHypervisor` prog that facilitates
downloading new versions of cloud hypervisor from the control
plane. This functionality depends on the
`host/bin/download-cloud-hypervisor` script. A new method,
`download_cloud_hypervisor`, has been added to `VmHost`,
enabling the download of new firmware versions to hosts via REPL.

Usage examples:

To download a new version 35.1 to a single host:
```ruby
vmh.download_cloud_hypervisor(
  version_x64: "35.1",
  sha256_ch_bin_x64: "e842...",
  sha256_ch_remote_x64: "337b..."
)
```

To download a new version to all hosts:
```ruby
VmHost.all.map { _1.download_cloud_hypervisor(
  version_x64: "35.1",
  sha256_ch_bin_x64: "e842...",
  sha256_ch_remote_x64: "337b...",
  version_arm64: "38.0",
  sha256_ch_bin_arm64: "c1c0...",
  sha256_ch_remote_arm64: "008f..."
)}
```

For knows versions (see `Prog::DownloadCloudHypervisor.sha_256`), the sha256 digest can be omitted:

```ruby
vmh.download_cloud_hypervisor(version_x64: "35.1")

VmHost.all.map { _1.download_cloud_hypervisor(version_x64: "35.1", version_arm64: "35.1") }
```

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -200,6 +200,19 @@ class VmHost < Sequel::Model
     Strand.create_with_id(schedule: Time.now, prog: "DownloadFirmware", label: "start", stack: [{subject_id: id, version: version, sha256: sha256}])
   end
 
+  # Introduced for downloading cloud hypervisor via REPL.
+  def download_cloud_hypervisor(version_x64: nil, version_arm64: nil, sha256_ch_bin_x64: nil, sha256_ch_bin_arm64: nil, sha256_ch_remote_x64: nil, sha256_ch_remote_arm64: nil)
+    version, sha256_ch_bin, sha256_ch_remote = if arch == "x64"
+      [version_x64, sha256_ch_bin_x64, sha256_ch_remote_x64]
+    elsif arch == "arm64"
+      [version_arm64, sha256_ch_bin_arm64, sha256_ch_remote_arm64]
+    else
+      fail "BUG: unexpected architecture"
+    end
+    fail ArgumentError, "No version provided" if version.nil?
+    Strand.create_with_id(schedule: Time.now, prog: "DownloadCloudHypervisor", label: "start", stack: [{subject_id: id, version: version, sha256_ch_bin: sha256_ch_bin, sha256_ch_remote: sha256_ch_remote}])
+  end
+
   def hetznerify(server_id)
     DB.transaction do
       HetznerHost.create(server_identifier: server_id) { _1.id = id }
--- a/rhizome/common/lib/util.rb
+++ b/rhizome/common/lib/util.rb
@@ -74,3 +74,7 @@ def safe_write_to_file(filename, content = nil)
     File.rename(temp_filename, filename)
   end
 end
+
+def curl_file(url, path)
+  r("bash -c 'curl -f -L3 #{url.shellescape} | tee >(openssl dgst -sha256) > #{path.shellescape}'").split(" ").last
+end
--- a/rhizome/host/bin/prep_host.rb
+++ b/rhizome/host/bin/prep_host.rb
@@ -53,12 +53,12 @@ r "chmod +x /etc/update-motd.d/99-clover-motd"
 r "timedatectl set-timezone UTC"
 
 # Download cloud hypervisor binaries.
-ch_dir = CloudHypervisor::VERSION.dir
+ch_dir = CloudHypervisor::VERSION_LEGACY.dir
 FileUtils.mkdir_p(ch_dir)
 FileUtils.cd ch_dir do
-  r "curl -L3 -o ch-remote #{CloudHypervisor::VERSION.ch_remote_url.shellescape}"
+  r "curl -L3 -o ch-remote #{CloudHypervisor::VERSION_LEGACY.ch_remote_url.shellescape}"
   FileUtils.chmod "a+x", "ch-remote"
-  r "curl -L3 -o cloud-hypervisor #{CloudHypervisor::VERSION.url.shellescape}"
+  r "curl -L3 -o cloud-hypervisor #{CloudHypervisor::VERSION_LEGACY.url.shellescape}"
   FileUtils.chmod "a+x", "cloud-hypervisor"
 end
 
--- a/rhizome/host/bin/prep_host.rb
+++ b/rhizome/host/bin/prep_host.rb
@@ -61,6 +61,7 @@ FileUtils.cd ch_dir do
   r "curl -L3 -o cloud-hypervisor #{CloudHypervisor::VERSION_LEGACY.url.shellescape}"
   FileUtils.chmod "a+x", "cloud-hypervisor"
 end
+CloudHypervisor::VERSION.download
 
 # Download firmware binaries.
 fw_dir = File.dirname(CloudHypervisor::FIRMWARE.path)
--- a/rhizome/host/lib/cloud_hypervisor.rb
+++ b/rhizome/host/lib/cloud_hypervisor.rb
@@ -50,7 +50,7 @@ module CloudHypervisor
   NEW_FIRMWARE = FirmwareClass.new(Arch.render(x64: "202311", arm64: "202211"),
     Arch.render(x64: "e31738aacd3d68d30f8f9a4d09711cca3dfb414e8910dc3af90c50f36885380a", arm64: "482f428f782591d7c2222e0bc8240d25fb200fb21fd984b3339c85979d94b4d8"))
 
-  VersionClass = Struct.new(:version) {
+  VersionClassLegacy = Struct.new(:version) {
     def ch_remote_url
       "https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v#{version}/ch-remote" + Arch.render(x64: "", arm64: "-static-aarch64")
     end
@@ -72,11 +72,55 @@ module CloudHypervisor
     end
   }
 
-  VERSION = if Arch.arm64?
-    VersionClass.new("35.0")
+  VERSION_LEGACY = if Arch.arm64?
+    VersionClassLegacy.new("35.0")
   elsif Arch.x64?
-    VersionClass.new("31.0")
+    VersionClassLegacy.new("31.0")
   else
     fail "BUG: unexpected architecture"
+  end
+
+  VersionClass = Struct.new(:version, :sha256_ch_bin, :sha256_ch_remote) {
+    def ch_remote_url
+      "https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v#{version}/ch-remote" + Arch.render(x64: "-static", arm64: "-static-aarch64")
+    end
+
+    def cloud_hypervisor_url
+      "https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v#{version}/cloud-hypervisor" + Arch.render(x64: "-static", arm64: "-static-aarch64")
+    end
+
+    def dir
+      "/opt/cloud-hypervisor/v#{version}"
+    end
+
+    def ch_remote_bin
+      File.join(dir, "ch-remote")
+    end
+
+    def bin
+      File.join(dir, "cloud-hypervisor")
+    end
+
+    def download
+      download_file(ch_remote_url, ch_remote_bin, sha256_ch_remote)
+      download_file(cloud_hypervisor_url, bin, sha256_ch_bin)
+    end
+
+    def download_file(url, path, sha256)
+      return if File.exist?(path)
+      FileUtils.mkdir_p(dir)
+      sha256_curl = nil
+      safe_write_to_file(path) do |f|
+        sha256_curl = curl_file(url, f.path)
+        fail "Invalid SHA-256 digest" unless sha256 == sha256_curl
+      end
+      FileUtils.chmod "a+x", path
+    end
+  }
+
+  VERSION = if Arch.arm64?
+    VersionClass.new("35.1", "071a0b4918565ce81671ecd36d65b87351c85ea9ca0fbf73d4a67ec810efe606", "355cdb1e2af7653a15912c66f7c76c922ca788fd33d77f6f75846ff41278e249")
+  elsif Arch.x64?
+    VersionClass.new("35.1", "e8426b0733248ed559bea64eb04d732ce8a471edc94807b5e2ecfdfc57136ab4", "337bd88183f6886f1c7b533499826587360f23168eac5aabf38e6d6b977c93b0")
   end
 end
--- a/rhizome/host/lib/cloud_hypervisor.rb
+++ b/rhizome/host/lib/cloud_hypervisor.rb
@@ -40,14 +40,10 @@ module CloudHypervisor
       FileUtils.mkdir_p(firmware_root)
       sha256_curl = nil
       safe_write_to_file(path) do |f|
-        sha256_curl = curl_firmware(f)
+        sha256_curl = curl_file(url, f.path)
         fail "Invalid SHA-256 digest" unless sha256 == sha256_curl
       end
       sha256_curl
-    end
-
-    def curl_firmware(file)
-      r("bash -c 'curl -f -L3 #{url.shellescape} | tee >(openssl dgst -sha256) > #{file.path.shellescape}'").split(" ").last
     end
   }
 
--- a/rhizome/host/spec/cloud_hypervisor_spec.rb
+++ b/rhizome/host/spec/cloud_hypervisor_spec.rb
@@ -39,4 +39,60 @@ RSpec.describe CloudHypervisor do
       end
     end
   end
+
+  context "when downloading cloud-hypervisor" do
+    subject(:ch) { CloudHypervisor::VersionClass.new("35.1", "sha_ch", "sha_remote") }
+
+    let(:gh_ch_remote_url) { "https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v35.1/ch-remote-static" }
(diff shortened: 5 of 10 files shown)


rhizome/common/lib/util.rb
```diff
@@ -70,3 +70,7 @@ def safe_write_to_file(filename, content = nil)
     File.rename(temp_filename, filename)
   end
 end
+
+def curl_file(url, path)
```
enescakir, line 74: We might also use this helper in the `BootImage` class in the future

rhizome/host/lib/cloud_hypervisor.rb
```diff
@@ -77,10 +73,54 @@ def ch_remote_bin
...
+      FileUtils.chmod "a+x", path
+    end
+  }
+
+  NEW_VERSION = VersionClass.new(Arch.render(x64: "35.1", arm64: "35.1"),
+    Arch.render(x64: "e8426b0733248ed559bea64eb04d732ce8a471edc94807b5e2ecfdfc57136ab4",
+      arm64: "071a0b4918565ce81671ecd36d65b87351c85ea9ca0fbf73d4a67ec810efe606"),
+    Arch.render(x64: "337bd88183f6886f1c7b533499826587360f23168eac5aabf38e6d6b977c93b0",
+      arm64: "355cdb1e2af7653a15912c66f7c76c922ca788fd33d77f6f75846ff41278e249"))
 end
```
enescakir, line 126: IMHO, this version is more readable
```suggestion
  NEW_VERSION = if Arch.arm64?
    VersionClass.new("35.1", "071a0b4918565ce81671ecd36d65b87351c85ea9ca0fbf73d4a67ec810efe606"), "355cdb1e2af7653a15912c66f7c76c922ca788fd33d77f6f75846ff41278e249")
  elsif Arch.x64?
    VersionClass.new("35.1", "e8426b0733248ed559bea64eb04d732ce8a471edc94807b5e2ecfdfc57136ab4", "337bd88183f6886f1c7b533499826587360f23168eac5aabf38e6d6b977c93b0")
  end
```

fdr:
will be reading this today fyi

rhizome/host/spec/cloud_hypervisor_spec.rb
```diff
@@ -13,23 +13,72 @@
...
+    subject(:ch) { CloudHypervisor::VersionClass.new("35.1", "sha_ch", "sha_remote") }
+
+    it "can use an existing version of cloud hypervisor" do
+      expect(File).to receive(:exist?).with("/opt/cloud-hypervisor/v35.1/cloud-hypervisor").and_return(true)
+      expect(File).to receive(:exist?).with("/opt/cloud-hypervisor/v35.1/ch-remote").and_return(true)
+      expect(ch).not_to receive(:curl_file)
+      ch.download
+    end
+
+    it "can download cloud hypervisor" do
```
fdr, line 48: Man, these tests are pretty brutal, I've gotta think about this one a bit.

rhizome/host/spec/cloud_hypervisor_spec.rb
```diff
@@ -13,23 +13,72 @@
...
+
+    it "fails if sha is incorrect" do
+      allow(Arch).to receive(:sym).and_return(:x64)
+      expect(File).to receive(:exist?).with("/opt/cloud-hypervisor/v35.1/ch-remote").and_return(false)
+      expect(File).to receive(:exist?).with("/opt/cloud-hypervisor/v35.1/cloud-hypervisor").and_return(false)
+      expect(FileUtils).to receive(:mkdir_p).with("/opt/cloud-hypervisor/v35.1").twice
+      expect(FileUtils).to receive(:chmod).with("a+x", "/opt/cloud-hypervisor/v35.1/ch-remote")
+      expect(FileUtils).not_to receive(:chmod).with("a+x", "/opt/cloud-hypervisor/v35.1/cloud-hypervisor")
+      f_remote = instance_double(File)
+      f_ch = instance_double(File)
```
fdr, line 74: You could use `instance_double(File, path: 'blah', some_other_attr: 'whatever')`

rhizome/host/bin/prep_host.rb
```diff
@@ -61,6 +61,7 @@
   r "curl -L3 -o cloud-hypervisor #{CloudHypervisor::VERSION.url.shellescape}"
   FileUtils.chmod "a+x", "cloud-hypervisor"
 end
+CloudHypervisor::NEW_VERSION.download
```
fdr, line 64: In general...if we're going through the trouble to name things `fooLegacy`, can't you spare yourself trouble by _not_ naming the new version `new`? Anything without `*Legacy` in it can presumed to be current, right?

model/vm_host.rb
```diff
@@ -196,6 +196,15 @@ def download_firmware(version_x64: nil, version_arm64: nil, sha256_x64: nil, sha
     Strand.create_with_id(schedule: Time.now, prog: "DownloadFirmware", label: "start", stack: [{subject_id: id, version: version, sha256: sha256}])
   end
 
+  # Introduced for downloading cloud hypervisor via REPL.
+  def download_cloud_hypervisor(version_x64: nil, version_arm64: nil, sha256_ch_bin_x64: nil, sha256_ch_bin_arm64: nil, sha256_ch_remote_x64: nil, sha256_ch_remote_arm64: nil)
+    version, sha256_ch_bin, sha256_ch_remote = (arch == "x64") ? [version_x64, sha256_ch_bin_x64, sha256_ch_remote_x64] : [version_arm64, sha256_ch_bin_arm64, sha256_ch_remote_arm64]
+    raise ArgumentError, "No version provided" if version.nil?
+    raise ArgumentError, "No SHA-256 digest provided for cloud-hypervisor" if sha256_ch_bin.nil?
+    raise ArgumentError, "No SHA-256 digest provided for ch-remote" if sha256_ch_remote.nil?
```
fdr, line 204: usually `fail` is used outside re-raise situations, I think.

model/vm_host.rb
```diff
@@ -196,6 +196,15 @@ def download_firmware(version_x64: nil, version_arm64: nil, sha256_x64: nil, sha
...
+    Strand.create_with_id(schedule: Time.now, prog: "DownloadCloudHypervisor", label: "start", stack: [{subject_id: id, version: version, sha256_ch_bin: sha256_ch_bin, sha256_ch_remote: sha256_ch_remote}])
```
fdr, line 205: Aside, not for this patch, long standing problem: it's rather gross we don't have a `Prog::DownloadCloudHypervisor.start` type of thing that gets rid of some of the free-handing of the record we see here.

model/vm_host.rb
```diff
@@ -196,6 +196,15 @@ def download_firmware(version_x64: nil, version_arm64: nil, sha256_x64: nil, sha
     Strand.create_with_id(schedule: Time.now, prog: "DownloadFirmware", label: "start", stack: [{subject_id: id, version: version, sha256: sha256}])
   end
 
+  # Introduced for downloading cloud hypervisor via REPL.
+  def download_cloud_hypervisor(version_x64: nil, version_arm64: nil, sha256_ch_bin_x64: nil, sha256_ch_bin_arm64: nil, sha256_ch_remote_x64: nil, sha256_ch_remote_arm64: nil)
+    version, sha256_ch_bin, sha256_ch_remote = (arch == "x64") ? [version_x64, sha256_ch_bin_x64, sha256_ch_remote_x64] : [version_arm64, sha256_ch_bin_arm64, sha256_ch_remote_arm64]
```
fdr, line 201: So uh, can we have some defaulting here? It's not like the artifacts ought to change after their release, we can build a little database in a hash literal here, right?

fdr (changes requested):
I guess also, do you have a prototype and/or strategy of how the upgrade process goes? There's two paths: generating systemd start units against the new C-H while leaving old ones intact (this is obliged I think), and then perhaps as something of an option that we definitely have to do sometime, some routine to force a systemd unit file rewrite and a VM restart to get a VM off an old hypervisor.

bsatzger:
> I guess also, do you have a prototype and/or strategy of how the upgrade process goes? There's two paths: generating systemd start units against the new C-H while leaving old ones intact (this is obliged I think), and then perhaps as something of an option that we definitely have to do sometime, some routine to force a systemd unit file rewrite and a VM restart to get a VM off an old hypervisor.

When this change goes in, we are able to download version `35.1` to all hosts. Then we would make a simple change to vm preparation to cut over to the new version. At that point, all newly created vms would use the new version.

We don't have that many long-lived vms, but I agree that we need a future change that rewrites systemd units. That would let vms pick up the new version after the next restart. We could backup the original unit file and create a new one based on the current logic. 

Vm restarts initiated by us might need some sort of maintenance window. Maybe we could introduce that concept, too, at some point. Customers could choose their time window, and our system would trigger customer impacting maintenance tasks (like a VM restart) during that window.

rhizome/host/spec/cloud_hypervisor_spec.rb
bsatzger: made some minor changes. did that make the tests a little more gentle?

rhizome/host/bin/prep_host.rb
bsatzger: renamed the original version (as well as existing references) to `VERSION_LEGACY`. `NEW_VERSION` was only meant to be a temporary name until the legacy version is removed.

model/vm_host.rb
bsatzger: Added defaulting to the Prog. You can now use `download_cloud_hypervisor(version_x64: "35.1")` (for an x64 host)

fdr: ...oof, but if you typo, it assumes things are arm, because its not x64? Can it be better still?

rhizome/common/lib/util.rb
fdr: yes, there should be de-duplication at some point.

rhizome/host/spec/cloud_hypervisor_spec.rb
fdr: Not...terribly, hah. I think I would consider isolating a subroutine and writing one brutal test for it, and then having the tests theoretically about sha checking be shorter.

For example, although `safe_write_for_file` is not tested (but perhaps it should be), it is partially mocked out in a couple of places that, if not for that, would unroll into a thicket of partial mocks in code that can do something like this:

```
rhizome/host/spec/cloud_hypervisor_spec.rb:      expect(fw).to receive(:safe_write_to_file).with("/opt/fw/CLOUDHV-202402.fd").and_yield(f)
```

model/vm_host.rb
```diff
@@ -195,11 +195,18 @@ def download_boot_image(image_name, version:, custom_url: nil)
   # Introduced for downloading firmware via REPL.
   def download_firmware(version_x64: nil, version_arm64: nil, sha256_x64: nil, sha256_arm64: nil)
     version, sha256 = (arch == "x64") ? [version_x64, sha256_x64] : [version_arm64, sha256_arm64]
-    raise ArgumentError, "No version provided" if version.nil?
-    raise ArgumentError, "No SHA-256 digest provided" if sha256.nil?
+    fail ArgumentError, "No version provided" if version.nil?
+    fail ArgumentError, "No SHA-256 digest provided" if sha256.nil?
```
fdr, line 199: oh huh, did I make this comment about a block not changed by your commits? oops.

fdr:
> > I guess also, do you have a prototype and/or strategy of how the upgrade process goes? There's two paths: generating systemd start units against the new C-H while leaving old ones intact (this is obliged I think), and then perhaps as something of an option that we definitely have to do sometime, some routine to force a systemd unit file rewrite and a VM restart to get a VM off an old hypervisor.
> 
> When this change goes in, we are able to download version `35.1` to all hosts. Then we would make a simple change to vm preparation to cut over to the new version. At that point, all newly created vms would use the new version.

Unless I need to gather data from code that is exercised but theoretically latent, or avoid merge conflicts, I generally would prefer to have the commit...at least the prototype of it...in hand, as to reduce the chance of miscalculation and another round of deploys. 

It does *seem* remote, I agree, but why not find out? Is there much gain by pipelining this patch  set a tiny bit sooner?

fdr (changes requested):
I think something can be about the large and repetitive test fixture: to bring it to more like one copy of brutal stuff, and increase focus in the tests that factor out the brutal stuff into a separate test.

As usual, I am not a fan of "else" meaning "the less common option", regardless of input. Even if it saves an expression or three.

model/vm_host.rb
bsatzger: `download_cloud_hypervisor(version_x64: "35.1")` works for x64 but fails for arm with ArgumentError.
`download_cloud_hypervisor(version_arm64: "35.2")` works for arm but fails for x64 with ArgumentError.
`download_cloud_hypervisor(version_x64: "35.1, version_arm64: "35.2")` works for both and can be used to download to all hosts in the fleet.

rhizome/host/spec/cloud_hypervisor_spec.rb
bsatzger: tests less brutal now?

model/vm_host.rb
bsatzger: I feel it's good that `download_cloud_hypervisor(version_x64: "35.1")` fails on an arm host but you can still use a single command like `VmHost.all.map { _1.download_cloud_hypervisor(version_x64: "35.1, version_arm64: "35.2") }` to download the appropriate version to the whole fleet.

bsatzger:
> > > I guess also, do you have a prototype and/or strategy of how the upgrade process goes? There's two paths: generating systemd start units against the new C-H while leaving old ones intact (this is obliged I think), and then perhaps as something of an option that we definitely have to do sometime, some routine to force a systemd unit file rewrite and a VM restart to get a VM off an old hypervisor.
> > 
> > 
> > When this change goes in, we are able to download version `35.1` to all hosts. Then we would make a simple change to vm preparation to cut over to the new version. At that point, all newly created vms would use the new version.
> 
> Unless I need to gather data from code that is exercised but theoretically latent, or avoid merge conflicts, I generally would prefer to have the commit...at least the prototype of it...in hand, as to reduce the chance of miscalculation and another round of deploys.
> 
> It does _seem_ remote, I agree, but why not find out? Is there much gain by pipelining this patch set a tiny bit sooner?

See https://github.com/ubicloud/ubicloud/commit/d0aafd744c916284fd26706670184011a7b13ed1 for how to start using the new C-H version.

fdr:
> > > > I guess also, do you have a prototype and/or strategy of how the upgrade process goes? There's two paths: generating systemd start units against the new C-H while leaving old ones intact (this is obliged I think), and then perhaps as something of an option that we definitely have to do sometime, some routine to force a systemd unit file rewrite and a VM restart to get a VM off an old hypervisor.
> > > 
> > > 
> > > When this change goes in, we are able to download version `35.1` to all hosts. Then we would make a simple change to vm preparation to cut over to the new version. At that point, all newly created vms would use the new version.
> > 
> > 
> > Unless I need to gather data from code that is exercised but theoretically latent, or avoid merge conflicts, I generally would prefer to have the commit...at least the prototype of it...in hand, as to reduce the chance of miscalculation and another round of deploys.
> > It does _seem_ remote, I agree, but why not find out? Is there much gain by pipelining this patch set a tiny bit sooner?
> 
> See [d0aafd7](https://github.com/ubicloud/ubicloud/commit/d0aafd744c916284fd26706670184011a7b13ed1) for how to start using the new C-H version.

well how about that.

model/vm_host.rb
bsatzger: I think we have the assumption that arch is either `x64` or `arm64` all over the place, but I'm happy to extend this check.

fdr: do we? This is kinda why I made `Arch.render`, to avoid this kind of thing

bsatzger: updated this to fail for unexpected architectures. 

we do this check in `LearnArch`: `fail "BUG: unexpected CPU architecture" unless ["arm64", "x64"].include?(arch)`, which provides some guarantees about valid architectures. I think we do have a couple of places in the code where we check for x64/arm64 and if it's not that one we assume it's the other. but I agree that's not optimal.

fdr (approved):

PR #1664, opened 2024-06-04
Download boot images from blob storage

byucesoy:
**Download boot images from blob storage for E2E tests**
There are several benefits of downloading boot images from blob storage:

- It is faster as most of the data transfer would be done in same region or in
the network of same provider.
- It is more reliable because there is no risk of source to be deleted or moved
without our knowledge.

The motivation trigger for this change was to reduce the time taken to download
the boot images, because our E2E tests are timing out due to slow download, but
the change is useful even out of E2E tests. So currently, I enabled this only
for E2E tests, but we can enable this for production as well in the future. We
would probably keep existing way for development because it doesn't require
setting up blob storage, thus it keeps the development setup simple.

### DIFF
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -14,7 +14,7 @@ class Prog::DownloadBootImage < Prog::Base
   end
 
   def download_from_blob_storage?
-    image_name.start_with?("github", "postgres")
+    image_name.start_with?("github", "postgres") || Config.e2e_test?
   end
 
   def default_boot_image_version(image_name)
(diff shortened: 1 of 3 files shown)


prog/download_boot_image.rb
```diff
@@ -41,26 +37,7 @@ def default_boot_image_version(image_name)
...
-      elsif image_name == "almalinux-8"
-        fail "Only x64 is supported for almalinux-8" unless vm_host.arch == "x64"
-        "https://repo.almalinux.org/almalinux/8/cloud/x86_64/images/AlmaLinux-8-GenericCloud-#{version}.x86_64.qcow2"
-      elsif image_name == "almalinux-9"
-        arch = (vm_host.arch == "x64") ? "x86_64" : "aarch64"
-        "https://repo.almalinux.org/almalinux/9/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-#{version}.#{arch}.qcow2"
-      else
-        fail "Unknown image name: #{image_name}"
-      end
+    @url ||= frame["custom_url"] || blob_storage_client.get_presigned_url("GET", Config.ubicloud_images_bucket_name, "#{image_name}-#{vm_host.arch}-#{version}.raw", 60 * 60).to_s
```
enescakir, line 40: Do we store all images in `.raw` format? The original Ubuntu images are in `.img` format, and AlmaLinux images are in `.qcow2` format. The `.raw` output for some images is 3-4 times larger than the original format.

byucesoy: I initially thought that we should converge to a single format in our blob storage, but it would require more effort to update the images. It would also make SHA256 verification logic more complex. Finally, as you noted, .raw output is bigger than the other formats. So I changed it to use original format of the image.

prog/download_boot_image.rb
```diff
@@ -44,8 +44,16 @@ def url
     @url ||=
       if frame["custom_url"]
         frame["custom_url"]
-      elsif download_from_blob_storage?
-        blob_storage_client.get_presigned_url("GET", Config.ubicloud_images_bucket_name, "#{image_name}-#{vm_host.arch}-#{version}.raw", 60 * 60).to_s
+      elsif download_from_blob_storage? || Config.e2e_test?
```
enescakir, line 47: Should we download from the blob storage in production to make production and E2E environments more similar? I prefer to use the same file for both production and E2E.
maybe
```suggestion
      elsif download_from_blob_storage? || !Config.development?
```

byucesoy: That is my intention as well, but I want to use this for E2E tests for a while (~1 week) before using it for prod.

enescakir (approved):
Did you upload images to the blob storage? Otherwise E2E will start to fail 😄

PR #1668, opened 2024-06-05
Introduce GitHub Actions Cache endpoints

enescakir:
### Add base of GitHub runtime API

The VM runtime token is a JWT token used for authenticating virtual machines with our runtime API, and it remains valid while the VM is running. We utilize the runner assigned to the VM and the repository associated with this runner in the "/runtime/github" endpoints. Subsequent commits introduce these new endpoints.

We pass the runtime token and URL to the runner as environment variables during the setup process.

### Add github cache entry migration

We introduced `github_cache_entry` table to track cache entries for "ubicloud/cache" package. Additionally, we added blob storage credentials to the `github_repository` table because we create a new bucket for each repository. The subsequent PRs provide a detailed explanation of these changes.

### Add AWS S3 SDK gem

The beast is back. We initially used the aws-s3-sdk gem to manage blob storage operations when setting up our MinIO cluster with our PostgreSQL service. However, considering its large size and our limited use of its operations, we removed it at 44a92e0d and created our own MinIO client. This has worked well until now, but I've encountered an issue: our client's presigned URLs aren't compatible with Cloudflare R2. Despite my attempts, I couldn't resolve this. So, for now, I'm reintroducing the aws-s3-sdk gem to our project until we can fix our client. Once we've resolved the issue, we'll remove the aws-s3-sdk gem again.

### Introduce GitHubCacheEntry model and add blob to repository

The cache entry contains metadata about the available caches for a specific repository. The unique identifiers for the cache, `key` and `version`, are used by the action client to either retrieve or upload a cache. The `scope` aids in maintaining the branch where the cache was created, as each branch cannot access the cache of another. The `upload_id` is generated when creating a multipart upload request on blob storage, which is then used to upload cache parts and complete the upload. Additionally, we keep track of the cache's last access time to clear old caches that haven't been accessed for a considerable period.

Each repository has its unique cache, necessitating a separate bucket for each repository. This separation naturally demarcates the caches of different repositories. Services like S3 and MinIO, which are S3 compatible, allow for the creation of access policies based on the object key prefix. Thus, we can limit access to a repository's cache by using the repository name as the object key prefix within the same bucket. However, Cloudflare R2 doesn't currently support this feature, requiring us to create a separate bucket and access token for each repository.

Since these credentials can only access the repository's bucket, we use them for object-level operations like generating presigned URLs and managing multipart uploads.

`url_presigner` helps to generate presigned URLs. We don't do any call to generate resigned URLs, it's generatable on the client side using its credentials.

### Manage setup/destroy of the repository's blob storage.

The blob storage credentials stored in github_repository table only have access to the repository's blob storage, so we need more powerful credentials to manage buckets. We supply these administrator credentials, which can create and delete buckets, through a Config value. We then use these credentials to initialize the admin client.

Cloudflare R2 markets itself as S3-compatible, but it doesn't support all S3 operations. Its authentication service is distinct from the storage service. Since it doesn't have ACL endpoints, we can't use the S3 client to create users and access keys. Instead, we must use the Cloudflare API for these operations.

When we receive the first cache request for a repository, we create a new bucket and an access token for it if it doesn't already have one. We store this access token in an encrypted database column, which we then use for generating presigned URLs for the specific bucket.

Alternatively, we could create a bucket at the time of repository creation. However, since not all repositories will use the cache feature, I prefer to only create a bucket when the repository actually needs it.

### Add reserveCache endpoint

The client reserves a cache entry before beginning to upload content to the blob storage. S3 compatible blob storage necessitates sending a `CreateMultipartUpload` request prior to the content upload. This returns an `UploadId`.

The 'POST /caches' endpoint, ('reserveCache' [^1]), creates a cache entry using the given "key" and "version" parameters. This prevents any other runner from starting to upload the same cache. It returns an `upload_id` from the blob storage and a list of presigned URLs for each 32MB part, which the client will use to upload each chunk.

The cache client uses the `cacheId` in some logics. This is unnecessary for our forked version since we already have `uploadId`. However, to avoid extensive modifications to the client code, a fixed `cacheId` is returned for each request.

### Add commitCache endpoint

The client commit the cache entry after uploading content to the blob storage. S3 compatible blob storage requires sending a `CompleteMultipartUpload` request following the content upload. This necessitates an `UploadId` and `ETags` from each part of the upload.

The 'POST /caches/commit' endpoint ('commitCache' [^2]) sends a `CompleteMultipartUpload` request to the blob storage using the etags provided by the client. Afterward, it marks the cache entry as committed, allowing other runners to access it.

### Add getCacheEntry endpoint

We've implemented caching endpoints based on GitHub's official cache client. We also have a forked version, which gives us flexibility, but we've tried to stay as close as possible to the official client, maintaining similar endpoints and parameters.

The official GitHub cache client is open-source and can be accessed here [^3]

The 'GET /cache' endpoint ('getCacheEntry' [^4]) retrieves cache entry using comma-separated 'keys' and 'version', returning a pre-signed URL to download the cache. If no cache entry is found, it returns a 204 status code.

Therefore, careful selection of cache entries is crucial.

Initially, we only consider committed caches for the repository of the authenticated runner. The runner can only access caches for the same branch or the default branch, but not for other branches.

The client can send multiple keys, and we should return the first matching cache entry in the same order as the keys are provided. The `.min_by { keys.index(_1.key) }` function assists with this. The `(scopes.index(_1.scope) * keys.size)` part helps us sort the caches by scope. If the same cache exists for the head_branch and the default branch, we should return the cache for the head_branch.

### Add listCache endpoint

The 'GET /caches' endpoint, ('listCache' [^5]), retrieves a list of cache entries for a single 'key' parameter. It does not provide presigned URLs. This endpoint is used by the client when no cache entry matches the 'getCacheEntry' request, and when debugging is turned on.

[^1]: https://github.com/ubicloud/toolkit/blob/d1df13e178816d69d96bdc5c753b36a66ad03728/packages/cache/src/internal/cacheHttpClient.ts#L204
[^2]: https://github.com/ubicloud/toolkit/blob/d1df13e178816d69d96bdc5c753b36a66ad03728/packages/cache/src/internal/cacheHttpClient.ts#L339
[^3]: https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/cacheHttpClient.ts
[^4]: https://github.com/ubicloud/toolkit/blob/d1df13e178816d69d96bdc5c753b36a66ad03728/packages/cache/src/internal/cacheHttpClient.ts#L103
[^5]: https://github.com/ubicloud/toolkit/blob/d1df13e178816d69d96bdc5c753b36a66ad03728/packages/cache/src/internal/cacheHttpClient.ts#L146

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,3 +58,5 @@ group :test do
 end
 
 gem "webauthn", "~> 3.1"
+
+gem "aws-sdk-s3", "~> 1.151"
--- a/clover_runtime.rb
+++ b/clover_runtime.rb
@@ -17,7 +17,7 @@ class CloverRuntime < Roda
   plugin :error_handler do |e|
     error = parse_error(e)
 
-    {error: error}.to_json
+    {error: error}.to_json unless error[:code] == 204
   end
 
   def get_jwt_payload(request)
--- a/config.rb
+++ b/config.rb
@@ -102,6 +102,10 @@ module Config
   optional :github_runner_service_project_id, string
   override :enable_github_workflow_poller, true, bool
 
+  # GitHub Cache
+  optional :github_cache_blob_storage_endpoint, string
+  optional :github_cache_blob_storage_region, string
+
   # Minio
   override :minio_host_name, "minio.ubicloud.com", string
   optional :minio_service_project_id, string
--- a/config.rb
+++ b/config.rb
@@ -105,6 +105,10 @@ module Config
   # GitHub Cache
   optional :github_cache_blob_storage_endpoint, string
   optional :github_cache_blob_storage_region, string
+  optional :github_cache_blob_storage_access_key, string, clear: true
+  optional :github_cache_blob_storage_secret_key, string, clear: true
+  optional :github_cache_blob_storage_account_id, string
+  optional :github_cache_blob_storage_api_key, string, clear: true
 
   # Minio
   override :minio_host_name, "minio.ubicloud.com", string
--- a/migrate/20240529_github_cache.rb
+++ b/migrate/20240529_github_cache.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:github_cache_entry) do
+      column :id, :uuid, primary_key: true
+      foreign_key :repository_id, :github_repository, type: :uuid, null: false
+      column :key, :text, collate: '"C"', null: false
+      column :version, :text, collate: '"C"', null: false
+      column :scope, :text, collate: '"C"', null: false
+      column :size, :bigint
+      column :upload_id, :text, collate: '"C"', unique: true
+      column :created_at, :timestamptz, null: false, default: Sequel.lit("now()")
+      column :created_by, :uuid, null: false
+      column :last_accessed_at, :timestamptz
+      column :last_accessed_by, :uuid
+      column :committed_at, :timestamptz
+      unique [:repository_id, :scope, :key, :version]
+    end
+
+    alter_table(:github_repository) do
+      add_column :access_key, :text, collate: '"C"'
+      add_column :secret_key, :text, collate: '"C"'
+    end
+  end
+end
--- a/model/github/github_repository.rb
+++ b/model/github/github_repository.rb
@@ -35,4 +35,39 @@ class GithubRepository < Sequel::Model
   def url_presigner
     @url_presigner ||= Aws::S3::Presigner.new(client: blob_storage_client)
   end
+
+  def admin_client
+    @admin_client ||= Aws::S3::Client.new(
+      endpoint: Config.github_cache_blob_storage_endpoint,
+      access_key_id: Config.github_cache_blob_storage_access_key,
+      secret_access_key: Config.github_cache_blob_storage_secret_key,
+      region: Config.github_cache_blob_storage_region
+    )
+  end
+
+  def destroy_blob_storage
+    admin_client.delete_bucket(bucket: bucket_name)
+    CloudflareClient.new(Config.github_cache_blob_storage_api_key).delete_token(access_key)
+    update(access_key: nil, secret_key: nil)
+  end
+
+  def setup_blob_storage
+    DB.transaction do
+      admin_client.create_bucket({
+        bucket: bucket_name,
+        create_bucket_configuration: {location_constraint: Config.github_cache_blob_storage_region}
+      })
+
+      policies = [
+        {
+          "effect" => "allow",
+          "permission_groups" => [{"id" => "2efd5506f9c8494dacb1fa10a3e7d5b6", "name" => "Workers R2 Storage Bucket Item Write"}],
+          "resources" => {"com.cloudflare.edge.r2.bucket.#{Config.github_cache_blob_storage_account_id}_default_#{bucket_name}" => "*"}
+        }
+      ]
+
+      token_id, token = CloudflareClient.new(Config.github_cache_blob_storage_api_key).create_token("#{bucket_name}-token", policies)
+      update(access_key: token_id, secret_key: Digest::SHA256.hexdigest(token))
+    end
+  end
 end
--- a/model/github/github_repository.rb
+++ b/model/github/github_repository.rb
@@ -1,14 +1,38 @@
 # frozen_string_literal: true
 
+require "aws-sdk-s3"
+
 require_relative "../../model"
 
 class GithubRepository < Sequel::Model
   one_to_one :strand, key: :id
   many_to_one :installation, key: :installation_id, class: :GithubInstallation
   one_to_many :runners, key: :repository_id, class: :GithubRunner
+  one_to_many :cache_entries, key: :repository_id, class: :GithubCacheEntry
 
   include ResourceMethods
   include SemaphoreMethods
 
   semaphore :destroy
+
+  plugin :column_encryption do |enc|
+    enc.column :secret_key
+  end
+
+  def bucket_name
+    ubid
+  end
+
+  def blob_storage_client
+    @blob_storage_client ||= Aws::S3::Client.new(
+      endpoint: Config.github_cache_blob_storage_endpoint,
+      access_key_id: access_key,
+      secret_access_key: secret_key,
+      region: Config.github_cache_blob_storage_region
+    )
+  end
+
+  def url_presigner
+    @url_presigner ||= Aws::S3::Presigner.new(client: blob_storage_client)
+  end
 end
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -7,5 +7,41 @@ class CloverRuntime
     end
 
     repository.setup_blob_storage unless repository.access_key
+
+    r.on "caches" do
+      # reserveCache
+      r.post true do
+        key = r.params["key"]
+        version = r.params["version"]
+        size = r.params["cacheSize"].to_i
+        fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if key.nil? || version.nil? || size == 0
+
+        unless (scope = runner.workflow_job&.dig("head_branch"))
+          # YYYY: If the webhook not delivered yet, we can try to get the branch from the API
+          Clog.emit("The runner does not have a workflow job") { {no_workflow_job: {ubid: runner.ubid, repository_ubid: repository.ubid}} }
+          fail CloverError.new(400, "InvalidRequest", "No workflow job data available")
+        end
+
+        if size > 10 * 1024 * 1024 * 1024
+          fail CloverError.new(400, "InvalidRequest", "The cache size is over the 10GB limit")
+        end
+
+        entry = GithubCacheEntry.create_with_id(repository_id: runner.repository.id, key: key, version: version, size: size, scope: scope, created_by: runner.id)
+
+        upload_id = repository.blob_storage_client.create_multipart_upload(bucket: repository.bucket_name, key: entry.blob_key).upload_id
+        entry.update(upload_id: upload_id)
+
+        max_chunk_size = 32 * 1024 * 1024 # 32MB
+        presigned_urls = (1..size.fdiv(max_chunk_size).ceil).map do
+          repository.url_presigner.presigned_url(:upload_part, bucket: repository.bucket_name, key: entry.blob_key, upload_id: upload_id, part_number: _1, expires_in: 900)
+        end
+
+        {
+          uploadId: upload_id,
+          presignedUrls: presigned_urls,
+          chunkSize: max_chunk_size
+        }
+      end
+    end
   end
 end
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class CloverRuntime
+  hash_branch("github") do |r|
+    if (runner = GithubRunner[vm_id: @vm.id]).nil? || (repository = runner.repository).nil?
+      fail CloverError.new(400, "InvalidRequest", "invalid JWT format or claim in Authorization header")
+    end
+  end
+end
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -5,5 +5,7 @@ class CloverRuntime
     if (runner = GithubRunner[vm_id: @vm.id]).nil? || (repository = runner.repository).nil?
       fail CloverError.new(400, "InvalidRequest", "invalid JWT format or claim in Authorization header")
     end
+
+    repository.setup_blob_storage unless repository.access_key
   end
 end
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -42,6 +42,29 @@ class CloverRuntime
     end
 
     r.on "caches" do
+      # listCache
+      r.get true do
+        key = r.params["key"]
+        fail CloverError.new(204, "NotFound", "No cache entry") if key.nil?
+
+        scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
+        entries = repository.cache_entries_dataset
+          .exclude(committed_at: nil)
+          .where(key: key, scope: scopes).all
+
+        {
+          totalCount: entries.count,
+          artifactCaches: entries.map do
+            {
+              scope: _1.scope,
+              cacheKey: _1.key,
+              cacheVersion: _1.version,
+              creationTime: _1.created_at
+            }
+          end
+        }
+      end
+
       # reserveCache
       r.post true do
         key = r.params["key"]
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -42,6 +42,33 @@ class CloverRuntime
           chunkSize: max_chunk_size
         }
       end
+
+      # commitCache
+      r.post "commit" do
+        etags = r.params["etags"]
+        upload_id = r.params["uploadId"]
+        size = r.params["size"].to_i
+        fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if etags.nil? || etags.empty? || upload_id.nil? || size == 0
+
+        entry = GithubCacheEntry[repository_id: repository.id, upload_id: upload_id, committed_at: nil]
+        fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil? || entry.size != size
+
+        begin
+          repository.blob_storage_client.complete_multipart_upload({
+            bucket: repository.bucket_name,
+            key: entry.blob_key,
+            upload_id: upload_id,
+            multipart_upload: {parts: etags.map.with_index { {part_number: _2 + 1, etag: _1} }}
+          })
+        rescue Aws::S3::Errors::InvalidPart, Aws::S3::Errors::NoSuchUpload => ex
+          Clog.emit("could not complete multipart upload") { {failed_multipart_upload: {ubid: runner.ubid, repository_ubid: repository.ubid, exception: Util.exception_to_hash(ex)}} }
+          fail CloverError.new(400, "InvalidRequest", "Wrong parameters")
+        end
+
+        entry.update(committed_at: Time.now)
+
+        {}
+      end
     end
   end
 end
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -8,6 +8,39 @@ class CloverRuntime
 
     repository.setup_blob_storage unless repository.access_key
 
+    # getCacheEntry
+    r.get "cache" do
+      keys, version = r.params["keys"]&.split(","), r.params["version"]
+      fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if keys.nil? || keys.empty? || version.nil?
+
+      # Clients can send multiple keys; we return the first matching cache in
+      # incoming key order. The function `.min_by { keys.index(_1.key) }` helps
+      # us achieve this by ordering entries based on the index of key in the
+      # given order. If the same cache exists for both the head_branch and the
+      # default branch, we prioritize and return the cache for the head_branch.
+      # The part `(scopes.index(_1.scope) * keys.size)` assists in sorting the
+      # caches by scope, pushing entries for later scopes to the end of the
+      # list.
+      scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
+      entry = repository.cache_entries_dataset
+        .exclude(committed_at: nil)
+        .where(key: keys, version: version, scope: scopes).all
+        .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
+
+      fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil?
+
+      entry.update(last_accessed_at: Time.now, last_accessed_by: runner.id)
+      signed_url = repository.url_presigner.presigned_url(:get_object, bucket: repository.bucket_name, key: entry.blob_key, expires_in: 900)
+
+      {
+        scope: entry.scope,
+        cacheKey: entry.key,
+        cacheVersion: entry.version,
+        creationTime: entry.created_at,
+        archiveLocation: signed_url
+      }
+    end
+
     r.on "caches" do
       # reserveCache
       r.post true do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -34,6 +34,22 @@ GEM
     argon2-kdf (0.2.0)
     ast (2.4.2)
     awrence (1.2.1)
+    aws-eventstream (1.3.0)
+    aws-partitions (1.947.0)
+    aws-sdk-core (3.198.0)
+      aws-eventstream (~> 1, >= 1.3.0)
+      aws-partitions (~> 1, >= 1.651.0)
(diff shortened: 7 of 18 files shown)


enescakir:
This PR lack tests and one small part where we create a bucket. I'm working on them currently.

routes/runtime/github.rb
```diff
@@ -0,0 +1,115 @@
...
+      fail CloverError.new(204, "NotFound", "No cache entry") if repository.access_key.nil?
+
+      keys, version = r.params["keys"]&.split(","), r.params["version"]
+      fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if keys.nil? || keys.empty? || version.nil?
+
+      scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
+      entry = repository.cache_entries_dataset
+        .exclude(committed_at: nil)
+        .where(key: keys, version: version, scope: scopes).all
+        .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
```
furkansahin, line 19: Can you add a comment here? I couldn't really understand how the minimum elements are fetched according to what.

routes/runtime/github.rb
```diff
@@ -0,0 +1,115 @@
...
+          Clog.emit("The runner does not have a workflow job") { {no_workflow_job: {ubid: runner.ubid, repository_ubid: repository.ubid}} }
+          fail CloverError.new(400, "InvalidRequest", "No workflow job data available")
+        end
+
+        entry = GithubCacheEntry.create_with_id(repository_id: runner.repository.id, key: key, version: version, size: size, scope: scope, created_by: runner.id)
+
+        upload_id = repository.blob_storage_client.create_multipart_upload(bucket: repository.bucket_name, key: entry.blob_key).upload_id
+        entry.update(upload_id: upload_id)
+
+        max_chunk_size = 32 * 1024 * 1024 # 32MB
```
furkansahin, line 77: I left a comment regarding this in other repository as well. Should we use a static chink size or somehow calculate a dynamic chunk size?

routes/runtime/github.rb
```diff
@@ -54,6 +54,37 @@ class CloverRuntime
...
+        presigned_urls = (1..size.fdiv(max_chunk_size).ceil).map do
+          repository.url_presigner.presigned_url(:upload_part, bucket: repository.bucket_name, key: entry.blob_key, upload_id: upload_id, part_number: _1)
+        end
+
+        {
+          cacheId: 1,
```
byucesoy, line 83: Should this be 1 in all cases? Seems like a weird protocol to return fixed value from endpoint.

byucesoy (commented):
Thanks for the PR. I think it looks good overall and I only have minor comments. Reviewing was easy thanks to nice split of commits. I'd only maybe put the reserve/commit commits before get/list commits because there were few areas I got confused first and then they become clear after reading the reserve/commit commits. Also it feels more natural to first "create" then "read".

byucesoy:
One question;

How can we limit potentially abusive very big file uploads, hopefully before completing the upload?

routes/runtime/github.rb
```diff
@@ -0,0 +1,115 @@
...
+      scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
+      entry = repository.cache_entries_dataset
+        .exclude(committed_at: nil)
+        .where(key: keys, version: version, scope: scopes).all
+        .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
+
+      fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil?
+
+      entry.update(last_accessed_at: Time.now, last_accessed_by: runner.id)
+      signed_url, _ = repository.url_presigner.presigned_request(:get_object, bucket: repository.bucket_name, key: entry.blob_key)
```
byucesoy, line 24: For how long this key is valid? I think we should set the validation duration to a minimum reasonable amount to prevent possibility of using our blob storage as file share by malicious actor.

model/github/github_repository.rb
```diff
@@ -1,14 +1,38 @@
...
+
+  def bucket_name
+    ubid
+  end
+
+  def blob_storage_client
+    @blob_storage_client ||= Aws::S3::Client.new(
+      endpoint: Config.github_cache_blob_storage_endpoint,
+      access_key_id: access_key,
+      secret_access_key: secret_key,
```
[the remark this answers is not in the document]
enescakir: > This PR lack some tests and one small part where we create a bucket
Sorry for the confusion. This relates to the part where we create a bucket.

The values from ﻿`Config` are the admin credentials, which enable us to create a bucket and generate an access key for it.

The values from the repository columns are the credentials needed to access a specific bucket. We set these when we actually create the bucket.

Gemfile
```diff
@@ -58,3 +58,5 @@ group :test do
 end
 
 gem "webauthn", "~> 3.1"
+
+gem "aws-sdk-s3", "~> 1.151"
```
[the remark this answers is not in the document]
enescakir: Good news, I believe I've identified the problem with the pre-signed URLs generated by our MinIO client. In the upcoming PR, I plan to implement the necessary changes and remove the S3 SDK.

routes/runtime/github.rb
enescakir: I've added some comments there.

routes/runtime/github.rb
enescakir: Yes, Users can extract it since we return it to the runner VM. The default time is set to 900 seconds (15 minutes), which I believe provides sufficient time for cache upload/download. This is based on our measurements that uploading or downloading a 10GB cache takes no more than 7 minutes with our current blob storage implementation. 

I also believe that a 15-minute limit is a good security measure. What are your thoughts on this? I've explicitly added this argument, even though it matches the default value. 

Given that the presigned URL is only valid for a specific bucket and prefix, the potential for attack is relatively small. However, there is always the risk of an attacker attempting to poison the caches. For this reason, GitHub does not permit branches to access caches from other branches.

enescakir:
> How can we prevent potentially abusive very big file uploads, hopefully before completing the upload?

The cache client checks the file size and doesn't allow uploads larger than 10GB. https://github.com/actions/toolkit/blob/c6b487124a61d7dc6c7bd6ea0208368af3513a6e/packages/cache/src/cache.ts#L205
However, someone could manipulate the client to try uploading larger files, as it runs within the user's environment. Since the client connects directly to the blob storage, limiting file size is challenging. I don't have an immediate solution.

routes/runtime/github.rb
enescakir: I've added it, ensuring not to unnecessarily alter the client code. It returns different values from the default GitHub Cache server, typically small integers. However, since we already use `uploadId` for S3 compatibility, we don't utilize `cacheId` at all. I found it weird too, so I decided to bite the bullet and refactor the client code.

> The cache client uses the cacheId in some logics. This is unnecessary for our forked version since we already have uploadId. However, to avoid extensive modifications to the client code, a fixed cacheId is returned for each request.

routes/runtime/github.rb
enescakir: The GitHub cache client includes a configuration value for setting the maximum chunk size, which defaults to 32MB. It remains unchanged in most customer workflows. Given the varying multipart upload workflows of Azure Blob Storage and S3 compatible storages, the server also needs to be aware of this maximum chunk size. 

I believe there's no need for a more complex chunk size calculation as it's also fixed for the default GitHub cache. However, I've added the maximum chunk size value to reserveCache response, enabling the backend to determine it. This logic can be enhanced in the future without causing any disruption to customers.

routes/runtime/github.rb
```diff
@@ -5,5 +5,31 @@ class CloverRuntime
...
       fail CloverError.new(401, "InvalidCredentials", "There was an error logging in")
     end
+
+    r.get "cache" do
+      fail CloverError.new(204, "NotFound", "No cache entry") if repository.access_key.nil?
+
+      keys, version = r.params["keys"]&.split(","), r.params["version"]
+      fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if keys.nil? || keys.empty? || version.nil?
+
+      scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
```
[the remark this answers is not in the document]
enescakir: That's a good question. As you pointed out, this is documented in GitHub: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache. However, I tested a few examples and was unable to access the caches in branch A. I plan to test this further in case I missed something.

Even though the actual implementation might differ, we should adhere to GitHub's documentation. We don't have insight into the implementation details, but we do have the documentation. I didn't include this scope because we lack the necessary data. The workflow_job payload doesn't contain this information, and we need to find a way to obtain it. Some other providers transmit this information from the client to the backend, but since we can't fully trust client data, I prefer not to source the information from the client. Making an API call to the GitHub API introduces additional latency and dependency just for obtaining caches. We may begin to store extra data about workflow runs with additional webhook events, similar to what we do with GitHub repositories, but it's unclear how much data we should keep.

I've decided to handle this part in a future, more isolated PR and have opted for a more restrictive version as the initial release.

routes/runtime/github.rb
bsatzger: In our case, is the concurrency defined by the size of the cache entry divided by 32MB? How many concurrent upload connections would we get for the maximum allowed cache entry size?

routes/runtime/github.rb
```diff
@@ -0,0 +1,128 @@
...
+        }
+      end
+
+      repository.setup_blob_storage unless repository.access_key
+
+      # reserveCache
+      r.post true do
+        key = r.params["key"]
+        version = r.params["version"]
+        size = r.params["cacheSize"].to_i
```
bsatzger, line 74: Should we limit the size and fail the request if it exceeds that limit?

routes/runtime/github.rb
```diff
@@ -0,0 +1,128 @@
...
+        end
+
+        entry = GithubCacheEntry.create_with_id(repository_id: runner.repository.id, key: key, version: version, size: size, scope: scope, created_by: runner.id)
+
+        upload_id = repository.blob_storage_client.create_multipart_upload(bucket: repository.bucket_name, key: entry.blob_key).upload_id
+        entry.update(upload_id: upload_id)
+
+        max_chunk_size = 32 * 1024 * 1024 # 32MB
+        presigned_urls = (1..size.fdiv(max_chunk_size).ceil).map do
+          repository.url_presigner.presigned_url(:upload_part, bucket: repository.bucket_name, key: entry.blob_key, upload_id: upload_id, part_number: _1, expires_in: 900)
```
bsatzger, line 90: Does this leave 15 minutes to start he upload or to finish the upload?

migrate/20240529_github_cache.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:github_cache_entry) do
```
bsatzger, line 5: Have you estimated very roughly how much traffic we'd route through our control plane and how many queries we'd make to Postgres if say 25% of our GH runners customers start to use our cache?

routes/runtime/github.rb
velioglu, line 74: Will there be a limit for account or repository level?

routes/runtime/github.rb
```diff
@@ -0,0 +1,128 @@
...
+
+        {
+          uploadId: upload_id,
+          presignedUrls: presigned_urls,
+          chunkSize: max_chunk_size
+        }
+      end
+
+      # commitCache
+      r.post "commit" do
```
velioglu, line 101: What happens if chunks can be uploaded partially? Does the client send a request to separate endpoint? How do we make sure that there is no dangling chunk.

routes/runtime/github.rb
enescakir: The cache client begins uploading all chunks simultaneously.

There was no mention of a maximum concurrency limit in the docs. The maximum number of upload parts allowed is 10,000. However, given our maximum cache size of 10GB/32MB, which equals 312 chunks, we can impose a concurrency limit on the client if necessary.

https://developers.cloudflare.com/r2/reference/limits/#account-plan-limits

routes/runtime/github.rb
```diff
@@ -42,6 +42,29 @@ class CloverRuntime
     end
 
     r.on "caches" do
+      # listCache
+      r.get true do
+        key = r.params["key"]
+        fail CloverError.new(204, "NotFound", "No cache entry") if key.nil?
```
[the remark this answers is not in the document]
enescakir: The official cache client uses `204` to indicate cache entries not found.

https://github.com/actions/toolkit/blob/361a115e538ac6d8eb06cc47f3fcecce557d04c8/packages/cache/src/internal/cacheHttpClient.ts#L122

routes/runtime/github.rb
enescakir: It leaves 15 minutes to start upload. Once the connection is open, the expiration doesn't matter. However, since we upload in chunks, we need to start uploading all chunks before the expiration.

clover_runtime.rb
```diff
@@ -17,7 +17,7 @@ class CloverRuntime < Roda
   plugin :error_handler do |e|
     error = parse_error(e)
 
-    {error: error}.to_json
+    {error: error}.to_json unless error[:code] == 204
```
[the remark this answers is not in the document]
enescakir: When we return content with HTTP 204, it fails with 204 should have no content. If we trigger an exception using `fail CloverError.new(204, "NotFound", "No cache entry")`, content is added by default. We could manually set the status and halt the request as an alternative, but using `CloverError` for HTTP errors consistently is more readable. The same check can be applied to API routes.

migrate/20240529_github_cache.rb
```diff
@@ -0,0 +1,26 @@
...
+      column :created_at, :timestamptz, null: false, default: Sequel.lit("now()")
+      column :created_by, :uuid, null: false
+      column :last_accessed_at, :timestamptz
+      column :last_accessed_by, :uuid
+      column :committed_at, :timestamptz
+      unique [:repository_id, :scope, :key, :version]
+    end
+
+    alter_table(:github_repository) do
+      add_column :access_key, :text, collate: '"C"'
```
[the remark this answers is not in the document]
enescakir: I copied them from the `postgres_timeline` table. I aim to maintain similar column names across different models to simplify the life of developers and on-call engineers.

https://github.com/ubicloud/ubicloud/blob/215626b6c305630ac1a414f056aca4c834ab057b/migrate/20231107_postgres_timeline.rb#L10

model/github/github_repository.rb
```diff
@@ -35,4 +35,42 @@ def blob_storage_client
   def url_presigner
     @url_presigner ||= Aws::S3::Presigner.new(client: blob_storage_client)
   end
+
+  def destroy_blob_storage
```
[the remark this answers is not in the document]
enescakir: Good catch. We should call it while destroying the repository. Let me make that change.

routes/runtime/github.rb
enescakir: The client uploads all chunks to the blob storage, each upload request returning an etag as a response. The client then sends these etags for each chunk as a list to the backend in the commit cache request. Finally, the backend sends a complete multipart request to the blob storage with these etags. If any etag is incorrect or missing, blob storage will return failure for complete multipart request.

routes/runtime/github.rb
enescakir: We will limit the total cache size for each repository. However, this PR doesn't include this feature

routes/runtime/github.rb
```diff
@@ -38,6 +38,33 @@ class CloverRuntime
...
       end
+
+      # commitCache
+      r.post "commit" do
+        etags = r.params["etags"]
+        upload_id = r.params["uploadId"]
+        size = r.params["size"].to_i
+        fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if etags.nil? || etags.empty? || upload_id.nil? || size == 0
+
+        entry = repository.cache_entries_dataset.where(upload_id: upload_id, committed_at: nil).first
```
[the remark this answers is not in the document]
enescakir: I changed it to `entry = GithubCacheEntry[repository_id: repository.id, upload_id: upload_id, committed_at: nil]`. What are your thoughts now? Does it feel like it's a single entry? I'm also planning to add a unique constraint to the `upload_id` column.

model/github/github_repository.rb
```diff
@@ -1,14 +1,76 @@
...
+      region: Config.github_cache_blob_storage_region
+    )
+  end
+
+  def url_presigner
+    @url_presigner ||= Aws::S3::Presigner.new(client: blob_storage_client)
+  end
+
+  def destroy_blob_storage
+    admin_client = Aws::S3::Client.new(
```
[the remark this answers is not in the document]
enescakir: Done

model/github/github_repository.rb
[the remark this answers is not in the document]
enescakir: I updated the commit messages. Hopefully, they're clearer now.

model/github/github_repository.rb
```diff
@@ -1,14 +1,76 @@
...
+      admin_client = Aws::S3::Client.new(
+        endpoint: Config.github_cache_blob_storage_endpoint,
+        access_key_id: Config.github_cache_blob_storage_access_key,
+        secret_access_key: Config.github_cache_blob_storage_secret_key,
+        region: Config.github_cache_blob_storage_region
+      )
+
+      admin_client.create_bucket({
+        bucket: bucket_name,
+        create_bucket_configuration: {location_constraint: Config.github_cache_blob_storage_region}
```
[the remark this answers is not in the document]
enescakir: Cloudflare's documentation recommends this method for creating a bucket in a specific region with AWS S3 SDK. https://developers.cloudflare.com/r2/reference/data-location/#set-hints-via-the-s3-api

I'm not sure they always respect the client's location

> Location Hints are optional parameters you can provide during bucket creation to indicate the primary geographical location you expect data will be accessed from.
Using Location Hints can be a good choice when you expect the majority of access to data in a bucket to come from a different location than where the create bucket request originates. Keep in mind Location Hints are a best effort and not a guarantee, and they should only be used as a way to optimize performance by placing regularly updated content closer to users.

config.rb
```diff
@@ -102,6 +102,12 @@ def self.e2e_test?
   optional :github_runner_service_project_id, string
   override :enable_github_workflow_poller, true, bool
 
+  # GitHub Cache
+  optional :github_cache_blob_storage_endpoint, string
+  optional :github_cache_blob_storage_access_key, string, clear: true
```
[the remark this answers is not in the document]
enescakir: I moved to the next commit

routes/runtime/github.rb
enescakir: The client has a limit, but it's still good to double-check here. I added that. However, the client might bypass this somehow since it connects directly to the blob storage. ref: https://github.com/ubicloud/ubicloud/pull/1668#issuecomment-2163187241

routes/runtime/github.rb
velioglu: My question was more about dangling chunks in case of partial upload. Your explanation offline (aborting after 7 days) answers my question.

velioglu (approved):
Thanks for the detailed commit messages.

model/github/github_repository.rb
```diff
@@ -1,14 +1,73 @@
...
+    DB.transaction do
+      admin_client.create_bucket({
+        bucket: bucket_name,
+        create_bucket_configuration: {location_constraint: Config.github_cache_blob_storage_region}
+      })
+
+      policies = [
+        {
+          "effect" => "allow",
+          "permission_groups" => [{"id" => "2efd5506f9c8494dacb1fa10a3e7d5b6", "name" => "Workers R2 Storage Bucket Item Write"}],
```
byucesoy, line 64: What is this id? Why do we hard code it?

model/github/github_repository.rb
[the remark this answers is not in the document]
byucesoy: Got it, but do you think we need separate secrets for each bucket, given that they are all generated, stored and used exclusively in the control plane. Only the presigned url is being passed to the data plane, which can be generated from single source.

routes/runtime/github.rb
byucesoy: Sounds good, maybe add a comment saying that this value is only there to conform to the protocol and is not being used actually.

model/github/github_repository.rb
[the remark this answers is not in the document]
enescakir: The pre-signed URL contains raw `access_key` (so this information is not private), and a `secret_key` is used to generate the signature. I'm not sure how it happens, but if the `secret_key` for a repository leak somehow, it would only affect that single repository, not all repositories.

model/github/github_repository.rb
enescakir: Cloudflare has predefined permissions. They suggest obtaining this ID by using the [List Token Permission Groups](https://developers.cloudflare.com/api/operations/permission-groups-list-permission-groups) endpoint.

https://developers.cloudflare.com/fundamentals/api/how-to/create-via-api/#permission-groups

byucesoy (approved):

migrate/20240529_github_cache.rb
enescakir: (_Sorry for the delayed reply. I thought I had already responded to this comment, but it seems I forgot to post it_)

Actually, I haven't done it yet, but let's do it here. I expect it will be very lower than our current traffic. GitHub sends all customer webhook events to our web server, regardless of whether they use ubicloud in their workflow or not. This results in high traffic on our web server. Usually, customer workflows are set up for a single language in their job, which means they download only one cache file. If the cache exists, it requires just one web request. If the cache doesn't exist, it requires two requests: reserve and commit. We use a single table for cache lookups, so the queries aren't complex. If we encounter performance issues, we can add indices to some columns.

routes/runtime/github.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class CloverRuntime
+  hash_branch("github") do |r|
```
fdr, line 4: question: sure, it makes sense to reject JWTs that are valid but refer to non-extant VMs (actually, the existence of these may suggest an attack, fair question if they should be logged or whether the logging opens a DOS vector). 

But do the JWTs also have a sensible expiry?

Finally, if the JWT is expired, vs. not-expired but carries a missing VM, do the error messages form any kind of meaningful information disclosure attack?

enescakir: If the request lacks a JWT, contains an invalid JWT, or references a non-existent VM/runner, we return the same response to prevent any potential information disclosure attacks.

https://github.com/ubicloud/ubicloud/blob/60e93097200eb1007445a84bd0b2e8622df7c2ed/clover_runtime.rb#L33
https://github.com/ubicloud/ubicloud/blob/60e93097200eb1007445a84bd0b2e8622df7c2ed/routes/runtime/github.rb#L5

Since runners are ephemeral, I didn't set an extra expiration for the JWT tokens. They expire automatically when the runner is deleted. However, the creation date is included in the payload. We can add a time limit to that whenever necessary.

PR #1672, opened 2024-06-06
Remove concurrency limit reached but allowed logs

furkansahin:


fdr:
Why? telemetry diet? what's the rough magnitude?

fdr (commented):
Why? telemetry diet? what's the rough magnitude?

furkansahin:
Talked with the Github Actions team in daily, this will stay for now

PR #1674, opened 2024-06-06
Reduce logging for up resources

furkansahin:
If the resource is up, there is no meaning to log it for every check. We can simply do it every 5th time. I am still keeping the logging for down resources at every repeat.


fdr:
Seems reasonable, but what's the rough reduction in log traffic?

byucesoy (approved):

furkansahin:
@fdr, I sampled it on the last week. I expect a 10% drop in usage;
Screenshot 2024-06-07 at 10 35 33

PR #1675, opened 2024-06-06
Run E2E tests in parallel

velioglu:
We currently have virtual machine and github runner e2e tests. We were running them sequentially as the previous test host machine was not big enough to have vms for both of those tests at the same time. Since we've increased the size of the host, running them in parallel.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -9,20 +9,28 @@ def main(options)
   hetzner_server_st = Prog::Test::HetznerServer.assemble(vm_host_id: options[:vm_host_id])
   wait_until(hetzner_server_st, "wait")
 
+  tests_to_wait = []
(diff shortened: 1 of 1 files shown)


fdr (approved):
Nice commit message, I didn't realize the constraint here. I guess it might be a nice-to-have to document how much faster it makes it, very roughly, at the end.

enescakir (approved):

byucesoy (approved):

enescakir:
Screenshot 2024-06-07 at 18 54 16
It improves by two minutes in the current situation because we spend most of our time rebooting three times. However, this will be more beneficial when we have additional test cases like GitHub, PG, etc.

PR #1678, opened 2024-06-07
Handle the deletion of the runner with running job smoother

enescakir:
Before we destroy the underlying virtual machine of the runner, we must ensure that it is not executing any job and has been deleted on GitHub's side. This precaution helps us avoid disrupting potentially active jobs.

Ideally, when we initiate the runner's deletion, it should have already completed its job. However, in some instances, such as when the runner script fails due to out-of-memory (OOM) errors, network issues, etc., it loses communication with GitHub. In such cases, GitHub delays marking the runner as idle to rule out intermittent issues. If the runner is still executing a job when we attempt its deletion, it raises an 'Octokit::UnprocessableEntity' exception. Instead of raising this exception and logging the entire stack trace, we can simply nap for a while and retry the deletion.

Additionally, I have extended the deadline from 10 to 15 minutes. Lately, we've encountered this situation due to some OOM errors. Typically, these are resolved between 10-15 minutes. There's no need for additional page.

### DIFF
--- a/spec/prog/vm/github_runner_spec.rb
+++ b/spec/prog/vm/github_runner_spec.rb
@@ -502,10 +502,16 @@ RSpec.describe Prog::Vm::GithubRunner do
 
   describe "#destroy" do
     it "naps if runner not deregistered yet" do
-      expect(client).to receive(:get)
+      expect(client).to receive(:get).and_return(busy: false)
       expect(client).to receive(:delete)
 
       expect { nx.destroy }.to nap(5)
+    end
+
+    it "naps if runner still running a job" do
+      expect(client).to receive(:get).and_return(busy: true)
+
+      expect { nx.destroy }.to nap(15)
     end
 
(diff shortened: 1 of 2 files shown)


spec/prog/vm/github_runner_spec.rb
```diff
@@ -502,12 +502,18 @@
 
   describe "#destroy" do
     it "naps if runner not deregistered yet" do
-      expect(client).to receive(:get)
+      expect(client).to receive(:get).and_return({busy: false})
```
fdr, line 505:
```suggestion
      expect(client).to receive(:get).and_return(busy: false)
```

I think this might work, but I'm ambivalent about this bit of syntactic simplification. ymmv.

enescakir: It worked. I don't have a strong preference either way. I just changed it.

bsatzger (approved):
LGTM

PR #1682, opened 2024-06-09
Retry downloading the boot image if it fails non-prod environment

byucesoy:
**Retry downloading the boot image if it fails non-prod environment**
The boot image download can fail due to various reasons, such as network. We
don't want it to cause failure in our E2E tests. This commit adds a retry logic
to the boot image download in the E2E tests. I didn't add the retry logic for
production case, because we might want to have a deeper look before blindly
retrying.

**Clean up the temp file if image download fails**
If the image download fails, the temp file is left behind, which breaks
the idempotency of the image download. This commit adds a cleanup step
to remove the temp file if the download fails.

### DIFF
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -136,7 +136,11 @@ class Prog::DownloadBootImage < Prog::Base
       }.to_json
       sshable.cmd("common/bin/daemonizer 'host/bin/download-boot-image' #{q_daemon_name}", stdin: params_json)
     when "Failed"
-      BootImage.where(vm_host_id: vm_host.id, name: image_name, version: version).destroy
+      if Config.production?
+        BootImage.where(vm_host_id: vm_host.id, name: image_name, version: version).destroy
+      else
+        sshable.cmd("common/bin/daemonizer --clean #{q_daemon_name}")
+      end
       fail "Failed to download '#{image_name}' image on #{vm_host}"
     end
 
--- a/rhizome/common/lib/util.rb
+++ b/rhizome/common/lib/util.rb
@@ -3,6 +3,7 @@
 require "bundler/setup"
 require "open3"
 require "shellwords"
+require "digest"
 
 class CommandFail < RuntimeError
   attr_reader :stdout, :stderr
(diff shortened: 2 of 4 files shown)


prog/download_boot_image.rb
```diff
@@ -136,8 +136,13 @@ def blob_storage_client
       }.to_json
       sshable.cmd("common/bin/daemonizer 'host/bin/download-boot-image' #{q_daemon_name}", stdin: params_json)
     when "Failed"
-      BootImage.where(vm_host_id: vm_host.id, name: image_name, version: version).destroy
-      fail "Failed to download '#{image_name}' image on #{vm_host}"
+      if !Config.production?
```
bsatzger, line 139: should we limit the number of retries?

byucesoy: We can, but the control plane is designed to retry failures indefinitely. It also has exponential backoff baked in, so it is not defenseless against truly problematic cases. I prefer to keep it as is.

As a tangential note, I added a retry limit for process restarts when respirate crashes, which I saw as a different situation (inside vs. outside of control plane).

enescakir: In the case of `!Config.production?`, we don't raise an exception, so exponential backoff isn't utilized. If there's an issue with the blob storage, it will repeatedly attempt to download and fail until the 40 minutes timeout.

rhizome/common/lib/util.rb
```diff
@@ -60,12 +60,14 @@ def sync_parent_dir(f)
 def safe_write_to_file(filename, content = nil)
   raise ArgumentError, "must provide either content or block" if (content.nil? && !block_given?) || (!content.nil? && block_given?)
   temp_filename = filename + ".tmp"
-  File.open(temp_filename, File::RDWR | File::CREAT) do |f|
-    f.flock(File::LOCK_EX)
+  File.open("#{temp_filename}.lock", File::RDWR | File::CREAT) do |lock_file|
```
enescakir, line 63: Does the lock file get deleted? If not, are lots of leftover lock files distracting?

prog/download_boot_image.rb
byucesoy: Converted `Clog.emit` into a `fail`

rhizome/common/lib/util.rb
byucesoy: It does not. I changed the code to create the lock file at /tmp instead to ensure its deletion eventually.

prog/download_boot_image.rb
enescakir: It won't back off for `!Config.production?` case. If it's `Failed`, it will clean up and raise an exception, increasing the try count. However, in the subsequent run, it will be 'NotStarted' due to the cleanup. It will then run successfully, decreasing the try count. However, if the issue persists, the next run will be `Failed` again. It will repeat until the 40 minutes timeout occurs

enescakir (approved):
Could you run E2E tests before merging, as `safe_write_to_file` is used in several parts of the provisioning process?

prog/download_boot_image.rb
byucesoy: Ah right... :slightly_smiling_face:

It seems backoff, retry and daemonized runs don't play nicely together. I can try to make them work, but then I'm not sure if we should spend effort there at this point. In development environment, we don't have backoff anyway.

I'm planning to resolve this comment thread and merge the PR after the tests passed.

PR #1692, opened 2024-06-13
Have e2e test environment loader similar to production environment

velioglu:
We've introduced a new e2e test environment recently, we haven't updated the loader code with the new environment which causes issues of updating frozen classes on e2e tests. With this commit having similar loader logic with production environment for e2e test environment.

### DIFF
--- a/loader.rb
+++ b/loader.rb
@@ -83,7 +83,7 @@ autoload_normal.call("model", flat: true)
 
 AUTOLOAD_CONSTANTS.freeze
 
-if Config.production?
+if Config.production? || Config.e2e_test?
   AUTOLOAD_CONSTANTS.each { Object.const_get(_1) }
 end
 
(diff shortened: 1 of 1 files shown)


fdr (approved):

fdr:
good catch

loader.rb
```diff
@@ -83,7 +83,7 @@
 
 AUTOLOAD_CONSTANTS.freeze
 
-if Config.production?
+if Config.production? || Config.e2e_test?
```
enescakir, line 86: I agree that we should keep the E2E and production environments as similar as possible. I believe that an extra environment with `Config.e2e_test?` can lead to divergence between the E2E and production environments, and encouraging people. If we decide to use blob storage for default images in production, I think we should remove `Config.e2e_test?` and strive to set `RACK_ENV=production` for E2E tests.

ref: https://github.com/ubicloud/ubicloud/pull/1664#discussion_r1630704271

enescakir (approved):

loader.rb
velioglu: Resolving it, as we'll be considering to do that with subsequent PR.

PR #1693, opened 2024-06-13
Start using new Cloud Hypervisor version

bsatzger:
With this change, newly created VMs will use a new version of Cloud Hypervisor: v35.1 for both x64 and arm64. Before this change, that version of Cloud Hypervisor was downloaded to all existing hosts.


fdr (approved):
:eyepop:

PR #1696, opened 2024-06-14
UI: Default to location with highest available capacity

bsatzger:
Before this change, during VM creation, the first element of of the "Location" radio button is selected by default.  After this change, the location with the highest available capacity is selected.

"Available capacity" is determined as the sum of available cores in a location. There are other options to calculate/interpret available capacity. The main drawback of this approach is that it doesn't consider fragmentation. Another option would be to pick the location with the host that has the max available capacity.  But in that case we'd currently default to our smallest location (US South).

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -37,6 +37,21 @@ class Project < Sequel::Model
     !!billing_info&.payment_methods&.any?
   end
 
+  def default_location
+    location_max_capacity = DB[:vm_host]
+      .where(location: Option.locations.map { _1.name })
+      .where(allocation_state: "accepting")
+      .select_group(:location)
+      .order { sum(Sequel[:total_cores] - Sequel[:used_cores]).desc }
+      .first
+
+    if location_max_capacity.nil?
+      Option.locations.first.name
+    else
+      location_max_capacity[:location]
+    end
+  end
+
   def path
     "/project/#{ubid}"
   end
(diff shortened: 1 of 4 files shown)


model/project.rb
```diff
@@ -37,6 +37,21 @@ def has_valid_payment_method?
     !!billing_info&.payment_methods&.any?
   end
 
+  def default_location
```
bsatzger, line 40: Note: Added `default_location` to project as it seams reasonable to let the project decide the default location and in the future there could be extensions, e.g., customers could pick their favorite location as default. But feel free to object if you feel this logic shouldn't go to project.

spec/model/project_spec.rb
```diff
@@ -57,4 +57,25 @@
       project.soft_delete
     end
   end
+
+  describe ".default_location" do
```
enescakir, line 61: We have some unit tests that directly use the database, and others that extensively mock DB requests. I'm unsure which approach we should adopt. We encountered a similar issue at my previous company. Some colleagues believe we should mock all external components to keep the unit tests highly isolated, while others think we should use a real database. The line between when to mock or not is not clear and thin for me, and I'm uncertain about where to draw the line.
For instance, in this particular test, we mock all DB requests. I'm wondering what benefits these tests offer, aside from increasing line coverage.

At times, I struggle to decide whether to mock a class or not. I'm posing this question to gain a better understanding of the topic.

@fdr might have some insights on this topic.

bsatzger: Fair point. I changed the tests to use and not mock the database. I agree that these tests cover a wider spectrum of the system under test.

furkansahin (approved):

PR #1697, opened 2024-06-14
Log github installation deletion for easy monitoring

velioglu:
To monitor github installation deletion via logs easily, emitting clog when the user deletes a github installation.

### DIFF
--- a/routes/web/webhook/github.rb
+++ b/routes/web/webhook/github.rb
@@ -48,6 +48,7 @@ class CloverWeb
       installation.runners.each(&:incr_destroy)
       installation.repositories.each(&:incr_destroy)
       installation.destroy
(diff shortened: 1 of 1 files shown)


routes/web/webhook/github.rb
```diff
@@ -48,6 +48,7 @@ def handle_installation(data)
       installation.runners.each(&:incr_destroy)
       installation.repositories.each(&:incr_destroy)
       installation.destroy
+      Clog.emit("GithubInstallation is deleted.") { {github_installation: {ubid: installation.ubid}} }
```
enescakir, line 51: Should we also include the installation type here?

enescakir (approved):

PR #1711, opened 2024-06-21
Preparations for Postgres metric export via prometheus

byucesoy:
**Configure TLS for prometheus web server**
Prometheus exposes a web server, which both provides a primitive UI and also
endpoints for scraping metrics. This commit configures the web server to use
TLS, which is done by providing a certificate and key file. We are reusing the
certificate we generated for the postgres as both postgres and prometheus
servers are exposed from the same URL, only the ports will differ.

**Add OS users to be used by the prometheus integration**
We added 2 new OS users; prometheus and ubi_monitoring, which are created while
generating the base PostgreSQL OS image. The prometheus user will be used to
run the prometheus server and node_exporter and the former means that it will
be exposed to outside. The ubi_monitoring user will be used to connect to the
database and scrape Postgres metrics. We are adding a special row for allowing
the ubi_monitoring OS user to connect to the database as the ubi_monitoring
database user. This user has pg_monitor rights, thus it is able to read/execute
various monitoring views and functions. I didn't want to allow prometheus to
connect to the database because it is exposed to the outside and potential
vulnerability on the prometheus server could lead to the database compromise
otherwise.

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -149,11 +149,21 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
     refresh_walg_credentials
 
     when_initial_provisioning_set? do
-      hop_configure
+      hop_configure_prometheus
     end
 
     vm.sshable.cmd("sudo -u postgres pg_ctlcluster 16 main reload")
     hop_wait
+  end
+
+  label def configure_prometheus
+    web_config = <<CONFIG
+tls_server_config:
+  cert_file: /dat/16/data/server.crt
+  key_file: /dat/16/data/server.key
+CONFIG
+    vm.sshable.cmd("sudo -u prometheus tee /home/prometheus/web-config.yml > /dev/null", stdin: web_config)
+    hop_configure
   end
 
   label def configure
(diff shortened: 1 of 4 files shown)


prog/postgres/postgres_server_nexus.rb
```diff
@@ -149,13 +149,28 @@ def before_run
...
+    vm.sshable.cmd("sudo adduser --disabled-password --gecos '' prometheus")
+    vm.sshable.cmd("sudo userdel -rf ubi_monitoring || true")
+    vm.sshable.cmd("sudo adduser --disabled-password --gecos '' ubi_monitoring")
+
+    web_config = <<CONFIG
+tls_server_config:
+  cert_file: /dat/16/data/server.crt
+  key_file: /dat/16/data/server.key
+CONFIG
+    vm.sshable.cmd("sudo -u prometheus tee /home/prometheus/web-config.yml > /dev/null", stdin: web_config)
```
bsatzger, line 170: how will you limit who can connect to the web server? will `web-config.yml` contain additional settings in the future?

byucesoy: Yes, that is correct. Subsequent commits will add additional configs for:
1. Setting up authentication
2. Lmiting the number of metrics we emit.

bsatzger (approved):

enescakir (approved):

PR #1712, opened 2024-06-21
User-provided firewall rules for VMs and private subnets

bsatzger:
### API: Add GET for firewall rules
This commit adds a GET endpoint for firewall rules. It allows to retrieve a single firewall rule by its ID.

Example:
```sh
curl "https://api.ubicloud.com/.../firewall-rule/$rule_id" \
  -H "Authorization: Bearer $jwt_token"
```

### API: Allow to create firewall with rules
Previously, it was only possible to create a firewall without rules.

```sh
curl "https://.../firewall" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
      "name": "api-firewall"
}'
```

This commit adds the possibility to specify the firewall rules when creating a firewall.

```sh
curl "https://.../firewall" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
      "name": "api-firewall",
      "firewall_rules": [
       {"cidr": "10.0.0.0/24", "port_range": "22"},
       {"cidr": "0.0.0.0/0", "port_range": "5432..5434"},
       {"cidr": "10.1.2.3/32"} 
  ]}'
```

It's still possible to create a firewall without rules.

```sh
curl "https://.../firewall" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
      "name": "api-firewall",
      "firewall_rules": []
}'
```

### Allow to specify default firewall rules for private subnets
When a private subnet is created, it also creates a firewall under the hood. This firewall is created with a default set of rules that allow all traffic. This is not always desirable, so this commit allows to specify the default rules when creating a private subnet.

The call to create a private subnet without firewall rules still works and continues to create a default firewall with "allow all" rules.

```sh
curl "https:/private-subnet/subnet-name" \
  -X POST \
  -H "Authorization: Bearer $jwt_token"
```

But now it's also possible to specify the rules of the default firewall when creating the subnet, like this:

```sh
curl "https:/private-subnet/subnet-name" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
    "firewall_rules": [
       {"cidr": "10.0.0.0/24", "port_range": "22"},
       {"cidr": "0.0.0.0/0", "port_range": "5432..5434"},
       {"cidr": "10.1.2.3/32"} 
  ]}'
```
This creates a private subnet with a default firewall that denies all incoming traffic:

```sh
curl "https:/private-subnet/subnet-name" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
      "firewall_rules": []
}'
```
### Allow to specify default firewall rules for a vm
When a VM is created without a private network, a private subnet is created and the VM is attached to it. This subnet is created with a default firewall that allows all traffic. This commit allows to specify the firewall rules when creating a VM.

Omitting firewall rules will continue to create a default subnet with a firewall that allows all traffic.

```sh
curl "https://.../vm/vm-name" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
	    "public_key": "the public ssh key"
}'
```

But now it's possible to specify the firewall rules:

```sh
curl "https://.../vm/vm-name" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
    "public_key": "the public ssh key",
    "firewall_rules": [
       {"cidr": "10.0.0.0/24", "port_range": "22"},
       {"cidr": "0.0.0.0/0", "port_range": "5432..5434"},
       {"cidr": "10.1.2.3/32"} 
  ]}'
```

It is also still possible to specify the private subnet to attach the VM to, in which case no additional firewalls or firewall rules are created:

```sh
curl "https://.../vm/vm-name" \
  -X POST \
  -H "Authorization: Bearer $jwt_token" \
  -H "Content-Type: application/json" \
  -d '{
       "public_key": "the public ssh key",
       "private_subnet_id": "the ps id"
  }'
```

### DIFF
--- a/routes/api/project/firewall/firewall_rule.rb
+++ b/routes/api/project/firewall/firewall_rule.rb
@@ -11,15 +11,8 @@ class CloverApi
       request_body_params = Validation.validate_request_body(request.body.read, required_parameters, allowed_optional_parameters)
 
       parsed_cidr = Validation.validate_cidr(request_body_params["cidr"])
-      port_range = if request_body_params["port_range"].nil?
-        [0, 65535]
-      else
-        request_body_params["port_range"] = Validation.validate_port_range(request_body_params["port_range"])
-      end
-
-      pg_range = Sequel.pg_range(port_range.first..port_range.last)
-
-      firewall_rule = @firewall.insert_firewall_rule(parsed_cidr.to_s, pg_range)
+      port_range = Validation.validate_port_range(request_body_params["port_range"])
+      firewall_rule = @firewall.insert_firewall_rule(parsed_cidr.to_s, Sequel.pg_range(port_range))
 
       Serializers::FirewallRule.serialize(firewall_rule)
     end
@@ -35,6 +28,13 @@ class CloverApi
 
         response.status = 204
         r.halt
+      end
+
+      request.get true do
+        if firewall_rule
+          Authorization.authorize(@current_user.id, "Firewall:view", @firewall.id)
+          Serializers::FirewallRule.serialize(firewall_rule)
+        end
       end
     end
   end
--- a/lib/validation.rb
+++ b/lib/validation.rb
@@ -158,6 +158,7 @@ module Validation
   end
 
   def self.validate_port_range(port_range)
+    return 0..65535 if port_range.nil?
     fail ValidationFailed.new({port_range: "Invalid port range"}) unless (match = port_range.match(ALLOWED_PORT_RANGE_PATTERN))
     start_port = match[1].to_i
 
@@ -170,7 +171,8 @@ module Validation
       fail ValidationFailed.new({port_range: "Port must be between 0 to 65535"}) unless (0..65535).cover?(start_port)
     end
 
-    end_port ? [start_port, end_port] : [start_port]
+    port_range = end_port ? [start_port, end_port] : [start_port]
+    port_range.first..port_range.last
   end
 
   def self.validate_request_body(request_body, required_keys, allowed_optional_keys = [])
--- a/lib/validation.rb
+++ b/lib/validation.rb
@@ -145,6 +145,14 @@ module Validation
     end
   end
 
+  def self.validate_firewall_rules(firewall_rules)
+    fail ValidationFailed.new({firewall_rules: "Missing array of firewall rules."}) unless firewall_rules.is_a?(Array)
+    firewall_rules.map do |rule|
+      fail ValidationFailed.new({firewall_rules: "Invalid firewall rule."}) unless rule.is_a?(Hash) && rule.key?("cidr")
+      [validate_cidr(rule["cidr"]).to_s, validate_port_range(rule["port_range"])]
+    end
+  end
+
   def self.validate_cidr(cidr)
     if cidr.include?(".")
       NetAddr::IPv4Net.parse(cidr)
--- a/model/firewall.rb
+++ b/model/firewall.rb
@@ -18,6 +18,16 @@ class Firewall < Sequel::Model
   dataset_module Pagination
   dataset_module Authorization::Dataset
 
+  def self.create_with_rules(name, description, rules)
+    DB.transaction do
+      firewall = Firewall.create_with_id(name: name, description: description)
+      rules.each do |rule|
+        firewall.insert_firewall_rule(rule[0], Sequel.pg_range(rule[1]))
+      end
+      firewall
+    end
+  end
+
   def path
     "/firewall/#{ubid}"
   end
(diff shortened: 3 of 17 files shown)


fdr:
@bsatzger and I had a call about this. We could, perhaps, abide this, although we would need new rules about when to break symmetry on what is POSTed and what is GETted: it's not reasonable to do joins for every correlation we seek to expose in this way, growing bigger no matter how limited the needed data requirements of the clients.

It's even worse in a list-type GET, where many VMs are rendered together....in the same subnet...with the same firewall rules (because it's actually normalized via Subnet).

The growth of such bloat would be a function of whatever "shortcuts" seemed convenient to push up into correlated objects at the time, conflicting with the desire to have fewer representations depending on exactly what context a Vm (or whatever) is rendered.

I proposed a general idea that instead of inlining little bits and pieces that may cut through data models, that instead the system be made somewhat recursive, for example:

```
POST vm
{..., private_subnet: {name: 'my-subnet', firewall_rules: [{name: 'my-rule', cidr: '0.0.0.0/0', etc}]} }
```

If `private_subnet` were a string, it's a reference to an existing private subnet. All GETs will return strings/references, never expanded output unless we 

If it's a dictionary, it's the same dictionary that would be passed as a provision call to creating a private_subnet independently, with more advanced defaults (at least to automatically choose a name based on the VM's name). The fields, however, would be the same. **It is presumed that there is no ambiguity between recursive object creations, and references as a string.** 

In doing so, the customer is taught something about the data model and the API. These rules can be explained in a manual (a provisioning shortcut that operates on the same representation as the stand-alone object's POST). If a user transitions to needing more power, they know they can transform to this:

``` POST subnet
{name: 'my-subnet', firewall_rules: [{...}]}
```

That is, the nested sub-document has the same semantics, though it may lose some defaults that we can calculate in the case of a correlated provision (most visibly: the name).

routes/api/project/firewall/firewall_rule.rb
```diff
@@ -36,6 +29,13 @@ class CloverApi
         response.status = 204
         r.halt
       end
+
+      request.get true do
+        if firewall_rule
```
enescakir, line 34: We check the existence of a single record after retrieving it in other endpoints. Should we follow the same procedure here as well?

https://github.com/ubicloud/ubicloud/blob/fc3cf1352ee5e88fddbadbca110209f758bcdfd3/routes/api/project/firewall.rb#L35

fdr:
Okay @bsatzger seems to think this might work (doesn't _not_ not work), but, we identified some reasons to hold off. The stripped down idea is to pursue a "zero candy" option for Terraform, where the number of API requests and knowledge of the data model and its dependency tree is absolutely brutal, building up a VM from atoms, explicitly. Now, granted, Terraform is a relatively "smart" client capable of defaulting internally, so it won't be brutal to use terraform itself, but it would be brutal for someone learning our API for programming directly.

The problem is that OpenAPI client generators we can grab off the internet look pretty awful with recursive data types: any kind of nested, type-defined sub-object/dictionary might cause severe indigestion. We'll want to survey these, that will take time, and we'd like Terraform to be in good shape sooner. So answering the question of exactly what API mechanism we want to use runs into how we anticipate customers using the API, whether with an HTTP client or OpenAPI generated SDKs. Answering that question is too big a pipeline stall to tolerate for gaining some feedback on using Ubicloud on Terraform.

It's my calculation that we'll have more Terraform users at first than API users, because Ubicloud is quite young, it seems less likely people are going to write large programs requiring API-levels of precision against Ubicloud for at least a little while.

ozgune:
Hey @bsatzger, @fdr - I have two clarifications on this. If we follow this "zero candy" API approach, could you share an example of Terraform and API requests to create a VM with firewall rules (only SSH open)?

I'm primarily trying to understand how our Terraform and API requests are going to compare to other cloud providers.

I think this is how Digital Ocean looks like.

You specific a droplet for your compute:
```
resource "digitalocean_droplet" "web" {
  name   = "web-1"
  size   = "s-1vcpu-1gb"
  image  = "ubuntu-18-04-x64"
  region = "nyc3"
}
```

And then add a firewall with rules:
```
resource "digitalocean_firewall" "web" {
  name = "only-22-80-and-443"

  droplet_ids = [digitalocean_droplet.web.id]

  inbound_rule {
    protocol         = "tcp"
    port_range       = "22"
    source_addresses = ["192.168.1.0/24", "2002:1:2::/48"]
  }
}
```

Under the hood, each of the two resources are created with a single CREATE request. Firewall CREATE allows to provide a list of droplet IDs.

bsatzger:
No candy in our case means to create the following resources (each with a single API call) 
- Firewall fw
- Firewall rule (for 192.168.1.0/24 and port 22, with firewall_id = fw.id)
- Firewall rule (for 2002:1:2::/48 and port 22, with firewall_id = fw.id)
- Subnet s (with firewall_id = fw.id)
- VM (with subnet_id = s.id)

In order to be able do that, we also need some minor changes to the API, as we need ways to
- Create a subnet without a default "allow all" firewall
- Associate a firewall to a subnet at subnet creation time

bsatzger:
Closing this for now in favor of https://github.com/ubicloud/ubicloud/pull/1724

PR #1715, opened 2024-06-24
Update PostgreSQL base image

byucesoy:
This image adds prometheus, node_exporter, and postgres_exporter to the base image. It also creates 2 new users; prometheus and ubi_monitoring to run the binaries. Finally, it also contains systemd unit files to allow managing these binaries with systemd.

### DIFF
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -90,7 +90,8 @@ class Prog::DownloadBootImage < Prog::Base
       ["github-ubuntu-2004", "arm64", "20240609.1.1"] => "458c66af10efca50c25efd2adaa8f97c402d4e3cd8e638d0b0503766248a4d89",
       ["github-gpu-ubuntu-2204", "x64", "20240422.1.0"] => "5bb0577f9aaeae298f5a66403ae55b2092e790eb98ea7ef5812892211a55a548",
       ["github-gpu-ubuntu-2204", "x64", "20240609.1.1"] => "bd1518b35ac5a848e67e2739ce6a9f715b2b97916621494348ad6af4de317c80",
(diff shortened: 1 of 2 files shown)


fdr:
Can you reference the image build source changes? They, too, need to be reviewed under the SOC rubric.

byucesoy:
The build source changes can be found here; https://github.com/ubicloud/postgres-images/pull/3

prog/download_boot_image.rb
```diff
@@ -85,7 +85,7 @@ def sha256_sum
       ["github-ubuntu-2004", "x64", "20240422.1.0"] => "cf4f3bd4fc43de5804eac32e810101fcfe078aafeb46cb5a34fff8f8f76b360d",
       ["github-ubuntu-2004", "arm64", "20240422.1.0"] => "3e34cf2cb05ff9ab8c915b556bf31f153e90b20de25551587fadbec81557204b",
       ["github-gpu-ubuntu-2204", "x64", "20240422.1.0"] => "5bb0577f9aaeae298f5a66403ae55b2092e790eb98ea7ef5812892211a55a548",
-      ["postgres-ubuntu-2204", "x64", "20240226.1.0"] => "f8a2b78189239717355b54ecf62a504a349c96b1ab6a21919984f58c2a367617"
+      ["postgres-ubuntu-2204", "x64", "20240624.1.0"] => "166f6fbc7b992fb38ea418ae6d49849166d94daffe5bbe0a5bab76eb8a767b37"
```
enescakir, line 88: Should we keep at least the latest version's checksum? If the new image encounters any issues, we can download the previous images to the new hosts without requiring a new deployment.

byucesoy: Makes sense, I added it back.

enescakir (approved):

PR #1717, opened 2024-06-25
Update rubocop-rspec to v3

enescakir:
rubocop-rspec had a major upgrade from v2 to v3, necessitating some modifications to our .rubocop.yml file.

- The `RSpec/FilePath` cop has been split into `RSpec/SpecFilePathFormat` and `RSpec/SpecFilePathSuffix`. We have already `RSpec/SpecFilePathFormat` cop enabled, but we had to disable `RSpec/FilePath` manually. But it's removed in the new version, so we don't need to disable it anymore.
- `RSpec/FactoryBot` and `RSpec/Rails` cops are extracted to separate gems. So we don't need to disable `RSpecRails/HaveHttpStatus` manually anymore which introduced at 68e40f32be0b28b53c94b38c440e05cb5add7a30.


fdr (approved):
I dig the elimination of some rule micromanagement. Also til anonymous `&`

PR #1720, opened 2024-06-25
Disable Octokit warnings in all environments

enescakir:
Octokit prints warnings upon loading if certain optional Faraday packages are not installed. We aren't planning to install these packages. We disabled these warnings using the `ruby-warning` package in tests. They become annoying in production, as they appear every time we initiate pry. They also introduce a new package warning [^1].

    ➜  ubicloud git:(main) heroku run ./bin/pry
    Running ./bin/pry on ⬢ ubicloud-console... up, run.1413 (Standard-1X)
    To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
    To use multipart middleware with Faraday v2.0+, install `faraday-multipart` gem; note: this is used by the ManageGHES client for uploading licenses
    [1] ⚠️  clover-production(main)>

There is no blocker to use `ruby-warning` in other environments too.

I have disabled these warnings in both production and testing. However, I have left them enabled in development as they also print some deprecation warnings, which we might find useful.

[^1]: https://github.com/octokit/octokit.rb/commit/7bc6dd50d9d82cd1ce1e025b8aaf34f9823846a2.
[^2]: https://github.com/octokit/octokit.rb/blob/10886bff72ce90621bbfa7863ce340991cef0851/lib/octokit/warnable.rb#L13.

### DIFF
--- a/loader.rb
+++ b/loader.rb
@@ -11,9 +11,11 @@ Bundler.setup
 
 require_relative "config"
 require "mail"
+require "warning"
 require "rack/unreloader"
 
 REPL = false unless defined? REPL
+Warning.ignore(/To use (retry|multipart) middleware with Faraday v2\.0\+, install `faraday-(retry|multipart)` gem/)
 
 Unreloader = Rack::Unreloader.new(reload: Config.development?, autoload: true) { Clover }
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -26,7 +26,6 @@ require "rspec"
 require "database_cleaner/sequel"
 require "logger"
 require "sequel/core"
-require "warning"
 require "webmock/rspec"
 
 # DatabaseCleaner assumes the usual DATABASE_URL, but the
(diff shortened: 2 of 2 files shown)


bsatzger:
There's also a short description of `OCTOKIT_SILENT` at https://github.com/octokit/octokit.rb?tab=readme-ov-file#using-env-variables .

fdr:
Do we want to ignore *all kinds* of deprecation warnings? Seems like the env-var over-includes. We do want some warning, maybe only not this one, right?

fdr:
I mean, I saw the trailing part about the commit message, I'm fairly whatever, but...do I want to see the annoying faraday-retry message every time? Not really in development, either. I guess...what was wrong with the old way?

enescakir:
The old way only worked in the test environment. After reading your messages, I wondered why we restricted its use to tests. I revisited the [`ruby-warning`](https://github.com/jeremyevans/ruby-warning) library and found no specific mention that it should be used in the test environment only. So, I believe we can use it in production, right? If we move these warning rules from `spec_helper` to `loader`, it will silence warnings across all environments. Do you foresee any problems with this approach?

fdr:
> The old way only worked in the test environment. After reading your messages, I wondered why we restricted its use to tests. I revisited the [`ruby-warning`](https://github.com/jeremyevans/ruby-warning) library and found no specific mention that it should be used in the test environment only. So, I believe we can use it in production, right? If we move these warning rules from `spec_helper` to `loader`, it will silence warnings across all environments. Do you foresee any problems with this approach?

Nope, I think we used warning in production in the past as well, because Some Gems are Bad (in at least this way).

enescakir:
I moved it from `spec_helper.rb` to `loader.rb`

bsatzger (approved):

fdr (approved):

PR #1722, opened 2024-06-26
Prometheus integration for PostgreSQL

byucesoy:
**Move Postgres server certs to /etc/ssl/certs**
We previously kept these certs at /dat/16/data, however, with the introduction
of the prometheus integration, now two services need access to these certs.
They are also managed by different users, so it makes sense to move the cert
files to a common location.

**Only access Postgres certs through cert_readers group**
We moved certs to /etc/ssl/certs in the previous commit, however files in this
directory can be read by anyone. This commit restricts access to the files to
the cert_readers group, which contains the postgres and prometheus user.

**Enable prometheus and helper services**
We want these services to automatically start after a reboot, thus we need to
enable them. We don't do this in the image generation, because at the first
boot, the services are not yet configured.

**Add Postgres metric destinations**
We are adding a new model; PostgresMetricDestination to keep track of each
Postgres resource's associated metric destinations. We then compile then and
write to prometheus.yml file so that Prometheus can write scraped metrics to
the destinations.

**Add UI and API endpoints for managing Postgres metric destinations**
We added capability to send metrics to remote destinations using prometheus'
remote write protocol at the backend. This commit adds the UI and API, so that
users can manage these destinations.

image


rhizome/postgres/bin/initialize-database-from-backup
```diff
@@ -15,6 +15,9 @@ r "chown postgres /dat"
 r "rm -rf /dat/16"
 r "rm -rf /etc/postgresql/16"
 
+# Required to read server certificate and key
+r "usermod --append --groups cert_readers postgres"
```
enescakir, line 19: Where do we add `prometheus` user to `cert_readers` group?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -166,10 +166,41 @@ def before_run
 CONFIG
     vm.sshable.cmd("sudo -u prometheus tee /home/prometheus/web-config.yml > /dev/null", stdin: web_config)
 
-    vm.sshable.cmd("sudo systemctl enable --now postgres_exporter")
-    vm.sshable.cmd("sudo systemctl enable --now node_exporter")
-    vm.sshable.cmd("sudo systemctl enable --now prometheus")
-    hop_configure
+    metric_destinations = postgres_server.resource.metric_destinations.map {
```
enescakir, line 169: What happens if prometheus can't connect to the remote destinations? Does it have a timeout?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -166,10 +166,41 @@ def before_run
...
+  basic_auth:
+    username: '#{_1.username}'
+    password: '#{_1.password}'
+METRIC_DESTINATION
+    }.prepend("remote_write:").join("\n")
+
+    prometheus_config = <<CONFIG
+global:
+  scrape_interval: 15s
+
```
enescakir, line 181: Should we add some labels to the metrics such as ubid of the PG resource and server? If the multiple PG servers write the same remote destination, the customer needs to filter metrics by label.

```
  # Attach these labels to any time series or alerts when communicating with
  # external systems (federation, remote storage, Alertmanager).
  external_labels:
    monitor: 'codelab-monitor'
```

prog/postgres/postgres_server_nexus.rb
```diff
@@ -166,10 +166,41 @@ def before_run
...
+      <<METRIC_DESTINATION
+- url: '#{_1.url}'
+  basic_auth:
+    username: '#{_1.username}'
+    password: '#{_1.password}'
+METRIC_DESTINATION
+    }.prepend("remote_write:").join("\n")
+
+    prometheus_config = <<CONFIG
+global:
```
enescakir, line 179: The default retention period is 15 days, with no limit on size. Do we have an estimate for disk usage? Since we're writing data remotely and no one will directly connect to this prometheus server and query metrics, should we consider reducing the retention time aggressively?

https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects

lib/validation.rb
```diff
@@ -208,4 +208,11 @@ def self.validate_short_text(text, field_name)
   def self.validate_account_name(name)
     fail ValidationFailed.new({name: "Name must only contain letters, numbers, spaces, and hyphens and have max length 63."}) unless name&.match(ALLOWED_ACCOUNT_NAME)
   end
+
+  def self.validate_url(url)
+    uri = URI.parse(url)
+    fail ValidationFailed.new({url: "Invalid URL"}) if !["http", "https"].include?(uri.scheme) || uri.host.nil? || uri.host.empty?
```
enescakir, line 214: Should we be more explicit if it doesn't start with "http/https"? It might be difficult for customers to understand that it's invalid due to the prefix.

serializers/postgres.rb
```diff
@@ -20,7 +20,8 @@ def self.serialize_internal(pg, options = {})
       base.merge!(
         connection_string: pg.connection_string,
         primary: pg.representative_server&.primary?,
-        firewall_rules: Serializers::PostgresFirewallRule.serialize(pg.firewall_rules.sort_by { |fwr| fwr.cidr.version && fwr.cidr.to_s })
+        firewall_rules: Serializers::PostgresFirewallRule.serialize(pg.firewall_rules.sort_by { |fwr| fwr.cidr.version && fwr.cidr.to_s }),
+        metric_destinations: pg.metric_destinations.map { {id: _1.ubid, url: _1.url} }
```
enescakir, line 24: Should we add a serializer for it?

views/postgres/show.erb
```diff
@@ -194,7 +194,75 @@
...
+                </td>
+              </form>
+            </tr>
+          <% end %>
+        </tbody>
+      </table>
+    </div>
+  <% end %>
+  <!-- Metric Destination Card -->
+  <% if Authorization.has_permission?(@current_user.id, "Postgres:edit", @pg[:id]) %>
```
enescakir, line 207: Could you please add a screenshot to show what it looks like?

views/postgres/show.erb
```diff
@@ -194,7 +194,75 @@
...
+              <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"></th>
+            <% end %>
+          </tr>
+        </thead>
+        <tbody class="divide-y divide-gray-200 bg-white">
+          <% @pg[:metric_destinations].each do |md| %>
+            <tr>
+              <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6" scope="row"><%= md[:url] %></td>
+              <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6" scope="row">-</td>
+              <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6" scope="row">-</td>
```
enescakir, line 232: We can use `*****` instead of `-`

lib/validation.rb
fdr: fair question: should the code support non-https? Should we let people complain and analyze their product problem to see if we should allow unencrypted communication?

rhizome/postgres/bin/initialize-database-from-backup
byucesoy: We add it during the [image generation](https://github.com/ubicloud/postgres-images/pull/3/files#diff-3a42b2ee5f81a94b1675a1297a2532d118cbcf129eab64618a1d58a2b7fffa1fR423).

prog/postgres/postgres_server_nexus.rb
byucesoy: Yes, it has 30 second timeout by [default](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write).

prog/postgres/postgres_server_nexus.rb
byucesoy: Based on my rought, back of the napkin calculations, 15 days retention would require ~1GB data. I'm also planning to reduce the number of metrics we emit, because by default prometheus emit lots of metrics that are not useful in our context. That would reduce the overall size further.

I added storage limit (to the image generation PR).

lib/validation.rb
byucesoy: I changed this to only support https. If we get enough complaints, we can consider supporting self-signed certificates (by asking user to upload their public CA cert)

serializers/postgres.rb
byucesoy: Since this is very small modification, I prefer to not add a serializer for it. If it becomes more complex, or needs to be used in multiple places in the future, I'd be fine to make this a serializer.

views/postgres/show.erb
byucesoy: Added

rhizome/postgres/bin/initialize-database-from-backup
byucesoy: On second thought, I'm removing this from here and add this to image generation as well.

prog/postgres/postgres_server_nexus.rb
byucesoy: Makes sense. Added.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -166,10 +166,44 @@ def before_run
...
+  scrape_interval: 15s
+  external_labels:
+    resource: #{postgres_server.resource.ubid}
+    role: #{postgres_server.primary? ? "primary" : "standby"}
```
enescakir, line 183: If there are multiple standbys, should we differentiate their metrics?

enescakir (approved):

prog/postgres/postgres_server_nexus.rb
```diff
@@ -166,10 +166,44 @@ def before_run
...
+    username: '#{_1.username}'
+    password: '#{_1.password}'
+METRIC_DESTINATION
+    }.prepend("remote_write:").join("\n")
+
+    prometheus_config = <<CONFIG
+global:
+  scrape_interval: 15s
+  external_labels:
+    resource: #{postgres_server.resource.ubid}
```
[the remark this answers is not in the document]
byucesoy: Makes sense. I added ubicloud_ prefix to the labels we set.

prog/postgres/postgres_server_nexus.rb
byucesoy: I didn't want to expose postgres_server's ubid, because so far, user's all interactions were against the resource, but it seems we have to for this case. I added an instance field to distinguish different standbys.

PR #1730, opened 2024-06-28
Fix Ruby cache in arm64 CI

enescakir:
I updated the cached Ruby path for arm64 runners from a hardcoded path to the $RUNNER_TOOL_CACHE environment variable at 69353ca. It appears this variable isn't set in the reusable workflows. Therefore, we retrieve the value from the `runner` context [^1] instead of directly from the environment variable.

[^1]: https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
       if: matrix.runs-on == 'ubicloud-arm'
       uses: actions/cache@v4
       with:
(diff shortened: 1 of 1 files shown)


enescakir:
Path is fixed
Before:
Screenshot 2024-06-28 at 15 14 36

After:
Screenshot 2024-06-28 at 15 14 54

furkansahin (approved):
🤞

PR #1731, opened 2024-06-28
Use Ubicloud cache instead of GitHub Cache in CI

enescakir:
We implement Ubicloud Cache as a drop-in replacement for GitHub Actions Cache. Ubicloud Cache improves on the GitHub Actions cache by storing cached files closer to Ubicloud runners, making cache downloading/uploading more reliable and up to 4x faster. As a first customer, we can use our own cache service within our CI.


velioglu (approved):

bsatzger (approved):
:rocket:

PR #1736, opened 2024-06-29
Send invoice failure notification to accounts with billing permissions

enescakir:
When an invoice fails to process, we send an email to the billing email address. Our experience shows that customers often provide service email addresses that aren't frequently checked. Consequently, the response rate from these emails is lower than that from users' registered email addresses. We will now also send an email to accounts with billing permissions for the project.

### DIFF
--- a/model/invoice.rb
+++ b/model/invoice.rb
@@ -100,7 +100,9 @@ class Invoice < Sequel::Model
 
   def send_failure_email(errors)
     ser = Serializers::Invoice.serialize(self, {detailed: true})
-    Util.send_email(ser[:billing_email], "Urgent: Action Required to Prevent Service Disruption",
+    receivers = [ser[:billing_email]]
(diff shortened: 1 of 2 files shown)


model/invoice.rb
```diff
@@ -100,7 +100,9 @@ def send_success_email(below_threshold: false)
 
   def send_failure_email(errors)
     ser = Serializers::Invoice.serialize(self, {detailed: true})
-    Util.send_email(ser[:billing_email], "Urgent: Action Required to Prevent Service Disruption",
+    receivers = [ser[:billing_email]]
+    receivers += project.accounts.select { Authorization.has_permission?(_1.id, "Project:billing", project.id) }.map(&:email)
+    Util.send_email(receivers, "Urgent: Action Required to Prevent Service Disruption",
```
velioglu, line 105: What about adding them to cc to have a single mail thread?

velioglu (approved):

byucesoy (approved):

enescakir:
I just added `.uniq` to `receivers`

PR #1737, opened 2024-07-01
Load Balancer Backend

furkansahin:
# How to test this PR locally?
It's a bit of a pain but you can do it. First and most important thing to have is a domain with access to dns management. Keep that at hand.
1. Run respirate and create a project.
2. Set environment variables;
```
ENV["LOAD_BALANCER_SERVICE_PROJECT_ID"] ||= "YOUR_PROJECT_ID"
ENV["LOAD_BALANCER_SERVICE_HOSTNAME"] ||= "lb.YOUR_DOMAIN"
```
3. Restart respirate.
4. Setup DNS zone for lb.YOUR_DOMAIN following this [wiki](https://github.com/ubicloud/ubinest/wiki/Setup-DNS-Server)
5. Once you have the dns server and zone setup, add the NS records for that dns server.
6. Provision 2 new VMs *in the same subnet* and start a web server in them (replace PONG_VM_NAME with the name of the VM. An example is this;
```python
from http.server import BaseHTTPRequestHandler, HTTPServer

class HealthCheckHandler(BaseHTTPRequestHandler):
    def do_GET(self):
        if self.path == '/up':
            self.send_response(200)
            self.send_header('Content-type', 'text/plain')
            self.end_headers()
            self.wfile.write(b'healthy')
        elif self.path == '/ping':
            self.send_response(200)
            self.send_header('Content-type', 'text/plain')
            self.end_headers()
            self.wfile.write(b'PONG_VM_NAME')
        else:
            self.send_response(404)
            self.end_headers()

def run(server_class=HTTPServer, handler_class=HealthCheckHandler, port=8000):
    server_address = ('', port)
    httpd = server_class(server_address, handler_class)
    print(f'Starting httpd server on port {port}...')
    httpd.serve_forever()

if __name__ == "__main__":
    run()
```
7. Provision a load balancer using a command like this;
```ruby
Prog::Vnet::LoadBalancerNexus.assemble(private_subnet.id, name: "my-first-lb", algorithm: "round_robin", src_port: 80, dst_port: 8000, health_check_endpoint: "/up")
```
8. Get the hostname of load balancer via;
```
LoadBalancer.first.hostname
```
9. Run this command locally, repetitively to see which server you get the answer back;
```
curl HOST_NAME/ping
```

After this initial setup, you can test different stuff like adding a new VM, removing one, shutting down one of the web servers, etc.

# Commits
## Remove flowtables from the veth interface
Keeping flowtables for the veth interface causes routing failure for
consecutive packages. After the initial packet hits the prerouting chain
and perform the dnat, we want it to continue with the post routing and
then resubmitted immediately to the kernel for ESP encapsulation.
However, with flowtable, the packets are immediately transformed to
their pre ESP phase (having private addresses in source and destination)
and then submitted back to veth interface. This causes packets to leak
out to vetho* interface, therefore, we miss the chance of creating an
ESP packet. This caused failures for balancing, therefore, we limit the
flowtables to the tap* device.

## Add Basic Load Balancing
This commit adds the basic Ubicloud load balancer. There are two main
parts of this commit:
1. Control plane implementation for VM <-> Load Balancer interactions.
2. Data plane implementation of the load balancer itself.

The first part involves managing the many-to-many entity relationship
between VMs and load balancers. This relationship is handled through
a join table called `load_balancers_vms`, which is automatically managed
by Sequel. Functions like `load_balancer.add_vm` create the intermediary
data in `DB[:load_balancers_vms]`.

The second part is more complex. The Ubicloud Load Balancer is
integrated into a VM's namespace. Each VM on a host has a dedicated
network namespace used for packet filtering, IPv4 NAT, and various
operational tasks. Our implementation uses nftables definitions within
the namespace. Based on packet descriptions and algorithms, we decide
whether to DNAT the packet to a neighboring VM or accept it on the
current one. We use private networking between VMs to transfer packets
when load balancing is required. Additionally, we support load balancing
for both IPv4 and IPv6. Customers can use the public IPv4 address of any
VM being load balanced. If there is only one VM, we perform port
remapping and leave the packet unchanged.

Our load balancer also supports 2 types of algorithms;
1. round_robin
2. source_hash

The important part to note for a load balancer with multiple VMs is
that, each load balancer node will keep their own round_robin counter or
the source_hash result. Therefore, if the client resolves to different
load balancer node at the DNS level, they may not see the behavior
exactly what the algorithm requires. For example, if they use
round_robin with 2 VMs;
1. The first connection is opened to vm1 which points the customer to
vm1.
2. The second connection  is opened to vm1 which points the customer to
vm2.
3. The third connection is opened to vm2 and it points the customer to
vm2.
As you can see, the expected behavior was vm1, vm2, vm1 but we broke it
because the 3rd time we went to a new load balancer.

## Load Balancer DNS integration
We need to provide a hostname to the customers to be able to face a
single address to provide continuous connectivity even when the
underlying IP addresses of the load balancer changes. We accomplish this
by integrating load balancer with our existing DNS service. We will
create a new dns zone and manage individual host names depending on the
load balancer activity.

## Add Load Balancer Health Probes
Health probes are an important part of a load balancer. The main
objective is to be able to find nodes that are down and remove them from
the load balancer so that the new connections are directed to the
healthy instances. In this commit, we implement health probes as a
separate Prog which is a child of the load balancer strand. This way,
the load balancer can update the state of individual nodes while the
health probes are individually performed. We create a single prog per
node in the load balancer. This way, a faulty node does not block the
health checks of other VMs. In case a status change is detected, we
start an update on the load balancer which reprograms all of the
nftables rules (including the failing one) to redirect the traffic to
only active VMs. In case the process is accessible again, we include it
in the load balancer.

## VM destroy that is part of a load balancer is handled
When we need to destroy a VM, we need to be able to graciously remove it
from the load balancer, give enough time to the clients of the load
balancer to start using other nodes. For that, we are removing most of
the entities at the VM destroy time but still keep the network namespace
running for 10 more minutes. That is more than enough for dns to expire
and all of the connections to be rerouted to the other load balancer
nodes.

## Load balancer tests
Since the project is very sizable and the commits tend to update similar
parts of the code to implement different features of the load balancer,
we had to write the tests at the very end to minimize the effort.

### DIFF
--- a/migrate/20240701_load_balancer.rb
+++ b/migrate/20240701_load_balancer.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:lb_algorithm, %w[round_robin hash_based])
+
+    create_table(:load_balancer) do
+      column :id, :uuid, primary_key: true
+      column :name, :text, null: false
+      column :algorithm, :lb_algorithm, null: false, default: "round_robin"
+      column :src_port, :integer, null: false
+      column :dst_port, :integer, null: false
+      foreign_key :private_subnet_id, :private_subnet, type: :uuid, null: false
+    end
+
+    create_table(:load_balancers_vms) do
+      foreign_key :load_balancer_id, :load_balancer, type: :uuid, null: false
+      foreign_key :vm_id, :vm, type: :uuid, null: false, unique: true
+      primary_key [:load_balancer_id, :vm_id]
+    end
+  end
+end
--- a/migrate/20240702_load_balancer_health_probes.rb
+++ b/migrate/20240702_load_balancer_health_probes.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:lb_node_state, %w[up down evacuating])
+
+    alter_table(:load_balancer) do
+      add_column :health_check_endpoint, :text, collate: '"C"', null: false
+      add_column :health_check_interval, :integer, null: false, default: 10
+      add_column :health_check_timeout, :integer, null: false, default: 5
+      add_column :health_check_up_threshold, :integer, null: false, default: 5
+      add_column :health_check_down_threshold, :integer, null: false, default: 3
+      add_constraint(:health_check_up_threshold_gt_0) { health_check_up_threshold > 0 }
+      add_constraint(:health_check_down_threshold_gt_0) { health_check_down_threshold > 0 }
+      add_constraint(:health_check_timeout_gt_0) { health_check_timeout > 0 }
+      add_constraint(:health_check_interval_gt_0) { health_check_interval > 0 }
+      add_constraint(:health_check_interval_lt_600) { health_check_interval < 600 }
+      add_constraint(:health_check_timeout_lt_health_check_interval) { health_check_timeout <= health_check_interval }
+    end
+
+    alter_table(:load_balancers_vms) do
+      add_column :state, :lb_node_state, null: false, default: "down"
+      add_column :state_counter, Integer, null: false, default: 0
+    end
+  end
+end
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -1,0 +1,37 @@
+#  frozen_string_literal: true
+
+require_relative "../model"
+
+class LoadBalancer < Sequel::Model
+  many_to_many :vms
+  one_to_one :strand, key: :id
+  many_to_one :private_subnet
+  one_to_many :projects, through: :private_subnet
+  one_to_many :load_balancers_vms, key: :load_balancer_id, class: LoadBalancersVms
+
+  plugin :association_dependencies, load_balancers_vms: :destroy
+
+  include ResourceMethods
+  include SemaphoreMethods
+  include Authorization::TaggableMethods
+  include Authorization::HyperTagMethods
+  semaphore :destroy, :update_load_balancer
+
+  def hyper_tag_name(project)
+    "project/#{project.ubid}/location/#{private_subnet.display_location}/load-balancer/#{name}"
+  end
+
+  def add_vm(vm)
+    DB.transaction do
+      super
+      incr_update_load_balancer
+    end
+  end
+
+  def detach_vm(vm)
+    DB.transaction do
+      DB[:load_balancers_vms].where(load_balancer_id: id, vm_id: vm.id).delete(force: true)
+      incr_update_load_balancer
+    end
+  end
+end
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -4,6 +4,7 @@ require_relative "../model"
 
 class LoadBalancer < Sequel::Model
   many_to_many :vms
+  many_to_many :active_vms, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: {state: ["up"]}
   many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: "evacuating")
   one_to_one :strand, key: :id
   many_to_one :private_subnet
@@ -33,6 +34,7 @@ class LoadBalancer < Sequel::Model
   def evacuate_vm(vm)
     DB.transaction do
       load_balancers_vms_dataset.where(vm_id: vm.id).update(state: "evacuating")
+      strand.children_dataset.where(prog: "Vnet::LoadBalancerHealthProbes").all.select { |st| st.stack[0]["subject_id"] == id && st.stack[0]["vm_id"] == vm.id }.map(&:destroy)
       incr_update_load_balancer
       incr_rewrite_dns_records
     end
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -4,6 +4,7 @@ require_relative "../model"
 
 class LoadBalancer < Sequel::Model
   many_to_many :vms
+  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: "evacuating")
   one_to_one :strand, key: :id
   many_to_one :private_subnet
   one_to_many :projects, through: :private_subnet
@@ -15,7 +16,7 @@ class LoadBalancer < Sequel::Model
   include SemaphoreMethods
   include Authorization::TaggableMethods
   include Authorization::HyperTagMethods
-  semaphore :destroy, :update_load_balancer
+  semaphore :destroy, :update_load_balancer, :rewrite_dns_records
 
   def hyper_tag_name(project)
     "project/#{project.ubid}/location/#{private_subnet.display_location}/load-balancer/#{name}"
@@ -25,13 +26,23 @@ class LoadBalancer < Sequel::Model
     DB.transaction do
       super
       incr_update_load_balancer
+      incr_rewrite_dns_records
     end
   end
 
-  def detach_vm(vm)
+  def evacuate_vm(vm)
     DB.transaction do
-      DB[:load_balancers_vms].where(load_balancer_id: id, vm_id: vm.id).delete(force: true)
+      load_balancers_vms_dataset.where(vm_id: vm.id).update(state: "evacuating")
       incr_update_load_balancer
+      incr_rewrite_dns_records
     end
+  end
+
+  def remove_vm(vm)
+    load_balancers_vms_dataset.where(vm_id: vm.id).all.map(&:destroy)
+  end
+
+  def hostname
+    "#{name}.#{ubid[-5...]}.#{Config.load_balancer_service_hostname}"
   end
 end
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -8,7 +8,7 @@ require "base64"
 
 class Prog::Vm::Nexus < Prog::Base
   subject_is :vm
-  semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup, :update_spdk_dependency, :waiting_for_capacity
+  semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup, :update_spdk_dependency, :waiting_for_capacity, :lb_expiry_started
 
   def self.assemble(public_key, project_id, name: nil, size: "standard-2",
     unix_user: "ubi", location: "hetzner-hel1", boot_image: Config.default_boot_image_name,
@@ -403,15 +403,13 @@ class Prog::Vm::Nexus < Prog::Base
         raise unless /Failed to stop .* Unit .* not loaded\./.match?(ex.stderr)
       end
 
-      host.sshable.cmd("sudo host/bin/setup-vm delete #{q_vm}")
+      # If there is a load balancer setup, we want to keep the network setup in
+      # tact for a while
+      action = vm.load_balancers.empty? ? "delete" : "delete_keep_net"
+      host.sshable.cmd("sudo host/bin/setup-vm #{action} #{q_vm}")
     end
 
     DB.transaction do
-      vm.nics.map do |nic|
-        nic.update(vm_id: nil)
-        nic.incr_destroy
-      end
-
       vm.vm_storage_volumes.each do |vol|
         vol.storage_device_dataset.update(available_storage_gib: Sequel[:available_storage_gib] + vol.size_gib)
       end
@@ -422,12 +420,42 @@ class Prog::Vm::Nexus < Prog::Base
       )
 
       vm.pci_devices_dataset.update(vm_id: nil)
+    end
 
+    hop_wait_lb_expiry unless vm.load_balancers.empty?
+
+    final_clean_up
+
+    pop "vm deleted"
+  end
+
+  label def wait_lb_expiry
+    unless vm.lb_expiry_started_set?
+      vm.incr_lb_expiry_started
+      vm.load_balancers.map { _1.evacuate_vm(vm) }
+      nap 10 * 60
+    end
+
+    vm.load_balancers.each do |lb|
+      lb.remove_vm(vm)
+    end
+
+    vm.vm_host.sshable.cmd("sudo host/bin/setup-vm delete_net #{q_vm}")
+
+    final_clean_up
+
+    pop "vm deleted"
+  end
+
+  def final_clean_up
+    DB.transaction do
+      vm.nics.map do |nic|
+        nic.update(vm_id: nil)
+        nic.incr_destroy
+      end
       vm.projects.map { vm.dissociate_with_project(_1) }
       vm.destroy
     end
-
-    pop "vm deleted"
   end
 
   label def start_after_host_reboot
--- a/prog/vnet/load_balancer_health_probes.rb
+++ b/prog/vnet/load_balancer_health_probes.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class Prog::Vnet::LoadBalancerHealthProbes < Prog::Base
+  subject_is :load_balancer
+
+  def vm
+    @vm ||= Vm[frame.fetch("vm_id")]
+  end
+
+  label def health_probe
+    response_code = nil
+    begin
+      endpoint = "#{vm.nics.first.private_ipv4.network}:#{load_balancer.dst_port}#{load_balancer.health_check_endpoint}"
+      response_code = vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} curl --max-time #{load_balancer.health_check_timeout} --silent --output /dev/null --write-out '%{http_code}' #{endpoint}")
+    rescue
+      response_code = "500"
+    end
+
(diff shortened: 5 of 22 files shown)


migrate/20240701_load_balancer.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:lb_algorithm, %w[round_robin source_hash])
```
fdr, line 5: this might be fine and all, I know there was some mailing list discussion, do we have a sense of what we'd put in the manual, or is this more an internal experimentation jig, or whatever? Not a blocker, it's not like a new lb algorithm is going to show up even once every two...four...six...years in nftables world, so agility to try them out is justifiable.

furkansahin: I was thinking in near future we start experimenting with least_conn.

prog/vnet/update_load_balancer.rb
```diff
@@ -0,0 +1,104 @@
...
+  label def remove_load_balancer
+    vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} nft --file -", stdin: generate_nat_rules(vm.ephemeral_net4.to_s, vm.nics.first.private_ipv4.network.to_s))
+    pop "load balancer is updated"
+  end
+
+  def generate_lb_based_nat_rules
+    public_ipv4 = vm.ephemeral_net4.to_s
+    public_ipv6 = vm.ephemeral_net6.to_s
+    private_ipv4 = vm.nics.first.private_ipv4.network
+    private_ipv6 = vm.nics.first.private_ipv6.nth(2)
```
fdr, line 29: This can't be sound, can it? since `nics` is unordered and you take first. I would check it for length one (i.e. `first` is unambiguous) and crash, because whenever someone (you?) comes along to add multiple-nic support to a VM, it's going to be hard to find/remember this code.

model/load_balancer.rb
```diff
@@ -4,6 +4,7 @@
 
 class LoadBalancer < Sequel::Model
   many_to_many :vms
+  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.lit("state NOT IN ('evacuating')")
```
fdr, line 8: Does this work instead of Sequel.lit?

```ruby
{state: Sequel.exclude(['evacuating'])}
```

model/load_balancer.rb
```diff
@@ -21,14 +22,24 @@ def hyper_tag_name(project)
...
       incr_update_load_balancer
+      incr_rewrite_dns_records
       super
     end
   end
 
   def detach_vm(vm)
     DB.transaction do
-      DB[:load_balancers_vms].where(load_balancer_id: id, vm_id: vm.id).delete(force: true)
+      DB[:load_balancers_vms].where(load_balancer_id: id, vm_id: vm.id).update(state: "evacuating")
```
fdr, line 33: Ah I see the delete just doesn't matter because it becomes update, ignore my previous comment I suppose; I don't think the intermediate state between the two patches being ever so slightly suboptimal is worth a change.

model/load_balancer.rb
```diff
@@ -21,14 +22,24 @@ def hyper_tag_name(project)
...
       incr_update_load_balancer
+      incr_rewrite_dns_records
     end
   end
+
+  def remove_vm(vm)
+    DB[:load_balancers_vms].where(load_balancer_id: id, vm_id: vm.id).delete(force: true)
```
fdr, line 41: ....but here is delete again, skipping auditing.

prog/vnet/load_balancer_nexus.rb
```diff
@@ -57,6 +62,7 @@ def before_run
   label def destroy
     decr_destroy
     strand.children.map { _1.destroy }
+    Prog::Vnet::LoadBalancerNexus.dns_zone&.delete_record(record_name: load_balancer.hostname)
```
fdr, line 82: Is there an economical way to tighten this up to only be okay with it being optional in development while keeping it tightly coupled? (as we gain less from two copies of code). I think you could do it with something like

```
if (binding = expr) && expr 
```

migrate/20240702_load_balancer_health_probes.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:lb_node_state, %w[up down evacuating])
```
fdr, line 5: So I don't quite understand the evacuating nomenclature, maybe I need to get used to it.

prog/vm/nexus.rb
```diff
@@ -403,15 +403,13 @@ def write_params_json
         raise unless /Failed to stop .* Unit .* not loaded\./.match?(ex.stderr)
       end
 
-      host.sshable.cmd("sudo host/bin/setup-vm delete #{q_vm}")
+      # If there is a load balancer setup, we want to keep the network setup in
+      # tact for a while
+      action = vm.load_balancers.empty? ? "delete" : "delete_wo_net"
```
fdr, line 408: maybe `delete_keep_net`? I presume that's what you mean here? like, delete, but *w*ith*o*ut deleting net?

fdr (approved):
Probably the thing I'm least happy about is the loss of 3% to non-lb VMs (we think?). It should at least go in the commit message. But are we just going to "never" get around to doing something about it?

The micro-thing that sticks out in my mind most is that `nics.first` thing.

But, I don't want to make my suggested areas for changes or more thought too inelastic, so, approve on your discretion of when to engage those, if at all. There is value in getting this integrated and making adjustments later.

prog/vnet/load_balancer_health_probes.rb
```diff
@@ -0,0 +1,31 @@
...
+  subject_is :load_balancer
+
+  def vm
+    @vm ||= Vm[frame.fetch("vm_id")]
+  end
+
+  label def health_probe
+    response_code = nil
+    begin
+      endpoint = "#{vm.nics.first.private_ipv4.network}:#{load_balancer.dst_port}#{load_balancer.health_check_endpoint}"
```
fdr, line 13: ditto on nics.first. Maybe should wrap them in something that complains.

model/load_balancer.rb
```diff
@@ -0,0 +1,34 @@
...
+  one_to_many :projects, through: :private_subnet
+
+  include ResourceMethods
+  include SemaphoreMethods
+  include Authorization::TaggableMethods
+  include Authorization::HyperTagMethods
+  semaphore :destroy, :update_load_balancer
+
+  def hyper_tag_name(project)
+    "project/#{project.ubid}/load-balancer/#{name}"
```
enescakir, line 20: Just to clarify, are load balancers location-specific or global resources?

prog/vnet/load_balancer_nexus.rb
```diff
@@ -0,0 +1,80 @@
...
+    load_balancer.vms.each do |vm|
+      bud Prog::Vnet::UpdateLoadBalancer, {"subject_id" => vm.id, "load_balancer_id" => load_balancer.id}, :update_load_balancer
+    end
+
+    hop_wait_update_vm_load_balancers
+  end
+
+  label def wait_update_vm_load_balancers
+    reap
+    if strand.children_dataset.where(prog: "Vnet::UpdateLoadBalancer").all? { _1.exitval == "load balancer is updated" } || strand.children.empty?
```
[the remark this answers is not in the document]
enescakir: Okay, the others are health probes. I see them in the next commits.

model/load_balancer.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def add_vm(vm)
```
bsatzger, line 23: should we name this `add_vm` and `remove_vm` or `attach_vm` and `detach_vm`?

model/load_balancer.rb
bsatzger: Subnets are location-specific, and load balancer seem to be children of subnets. It seems reasonable to make them location-specific. But their table doesn't have a location column.

prog/vnet/update_load_balancer.rb
furkansahin: I'm not sure what you mean by "for length one". You're right, using an unordered "nics" call is an invitation to failure the moment we add multiple nics support. I'll fix that in the next iteration in the whole code utilising the nic.created_at column. Since we have these "vm.nics.first" call everywhere in the code, I'll skip here, too.

prog/vnet/update_load_balancer.rb
```diff
@@ -0,0 +1,104 @@
...
+    neighbor_ips_v4_set, neighbor_ips_v6_set = generate_lb_ip_set_definition(neighbor_vms)
+    modulo = load_balancer.vms.count
+    ipv4_map_def, ipv6_map_def = generate_lb_map_defs
+
+    balance_mode_ip4, balance_mode_ip6 = if load_balancer.algorithm == "round_robin"
+      ["numgen inc", "numgen inc"]
+    elsif load_balancer.algorithm == "source_hash"
+      ["jhash ip saddr . tcp sport . ip daddr . tcp dport", "jhash ip6 saddr . tcp sport . ip6 daddr . tcp dport"]
+    else
+      raise "Unsupported load balancer algorithm: #{load_balancer.algorithm}"
```
bsatzger, line 40: `fail` instead of `raise`, see https://github.com/ubicloud/ubicloud/commit/aa6efb1bcfa20fb5e045ff241b2106f7891fe596

ubid.rb
```diff
@@ -69,6 +69,7 @@ class UBID
   TYPE_FIREWALL = "fw"
   TYPE_POSTGRES_FIREWALL_RULE = "pf"
   TYPE_GITHUB_REPOSITORY = "gp"
+  TYPE_LOAD_BALANCER = "1b"
```
bsatzger, line 72: this is `1b`, why not `lb`?

prog/vnet/load_balancer_nexus.rb
```diff
@@ -38,9 +43,21 @@ def before_run
...
     end
 
+    hop_create_new_health_probe if strand.children_dataset.count < load_balancer.vms_dataset.count
+
     nap 5
   end
 
+  label def create_new_health_probe
+    vms = load_balancer.vms
+    vms_getting_probbed = strand.children_dataset.where(prog: "Vnet::LoadBalancerHealthProbes").map { |st| st.stack[0]["subject_id"] }
```
bsatzger, line 53: `probbed` -> `probed`

model/load_balancer.rb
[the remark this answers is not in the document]
furkansahin: Considering this is the initial implementation and there won't be any load balancer initialization in the middle, consecutive commit integration to be broken is something I am OK with in this case. It's a big hassle to reshuffle the code only to be updated in the next commit as it is now.

prog/vnet/update_firewall_rules.rb
```diff
@@ -74,7 +74,7 @@ class Prog::Vnet::UpdateFirewallRules < Prog::Base
 
   flowtable ubi_flowtable {
     hook ingress priority filter
-    devices = { #{vm.nics.map(&:ubid_to_tap_name).join(",")}, vethi#{vm.inhost_name} }
```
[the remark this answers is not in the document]
furkansahin: Updated the commit message

model/load_balancer.rb
furkansahin: I would like to keep it. It makes reading the code easier.

model/load_balancer.rb
furkansahin: Audit still can be done through the archived health probe record. Destroy doesn't work here because I did not add a model for load_balancers_vms.

migrate/20240702_load_balancer_health_probes.rb
furkansahin: I'm using it for VMs to be destroyed and completely removed but simply not yet because the namespace is being used by active connections.

prog/vnet/load_balancer_health_probes.rb
furkansahin: answered in the other thread.

prog/vnet/update_load_balancer.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+class Prog::Vnet::UpdateLoadBalancer < Prog::Base
+  subject_is :vm
```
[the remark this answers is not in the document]
furkansahin: Noted. I will update this one later since now with multiple commits working on different parts of this file, renaming it will require a good amount of work. I'll do that in a secondary pr.

model/load_balancer.rb
furkansahin: I have named them as `add_vm`, `remove_vm`, `evacuate_vm`. Is this good?

ubid.rb
furkansahin: I had it `lb` first, but then the tests failed. I searched some and found out that the algorithm we use for UBID calculation maps the chars 1, l, i, all to 1. That's why we must use only 1 for all of those. https://github.com/ubicloud/ubicloud/blob/4ea8e725561fd35e74032a3f469e7ab7b97df9a6/spec/ubid_spec.rb#L133

model/load_balancer.rb
furkansahin:
```
    # In crockford's base32 multiple characters can map to a single number,
    # for example ["1","I","L"] all map to 1. However, the reverse (number
    # -> character) only uses one canonical one (e.g. 1 maps to "1").
    # This test ensures we only use canonical characters in our type
    # constants, so ids are actually prefixed by those constants.
    #
    # See https://www.crockford.com/base32.html
```

ubid.rb
furkansahin:
```
    # In crockford's base32 multiple characters can map to a single number,
    # for example ["1","I","L"] all map to 1. However, the reverse (number
    # -> character) only uses one canonical one (e.g. 1 maps to "1").
    # This test ensures we only use canonical characters in our type
    # constants, so ids are actually prefixed by those constants.
    #
    # See https://www.crockford.com/base32.html
```

model/load_balancer.rb
fdr: Okay, I see, it's a pure join table, with no model....hmm that's a little weird, but I can't deny (some formats of) economy here. I kinda wonder if it's going to lead you to a weird path dependency where people don't put stuff on that model that "ought" to exist because they don't feel like introducing it. 

I can go either way, I guess were it me, I'd make a model.

model/load_balancer.rb
furkansahin: started using private subnet's location 👍

prog/vnet/load_balancer_health_probes.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class Prog::Vnet::LoadBalancerHealthProbes < Prog::Base
```
byucesoy, line 3: I think you can defer most (all?) of this logic to monitoring by defining `check_pulse` and `init_health_monitor_session` for LoadBalancerVms.

model/load_balancer.rb
```diff
@@ -4,6 +4,7 @@
 
 class LoadBalancer < Sequel::Model
   many_to_many :vms
+  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: "evacuating")
```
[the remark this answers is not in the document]
furkansahin: `load_balancer.vms` are simply the VMs added to the load balancer but they do not have the state of the VM entity for the load balancer. `load_balancer.load_balancers_vms` have that `state` to be used to either add the VM to the load balancing or not. Therefore, we must use `load_balancer.load_balancers_vms`. You might argue, then, let's use directly that and filter it when we need to use in the code. You have a point, but, this way, much less filtering sneaks into the real code and I have a single relationship to make the decision to add that VM to the dns result or not. I would like to keep it this way.

model/load_balancer.rb
```diff
@@ -24,14 +25,25 @@ def hyper_tag_name(project)
...
       super
     end
   end
 
-  def detach_vm(vm)
+  def evacuate_vm(vm)
     DB.transaction do
-      DB[:load_balancers_vms].where(load_balancer_id: id, vm_id: vm.id).delete(force: true)
+      load_balancers_vms_dataset.where(vm_id: vm.id).update(state: "evacuating")
+      strand.children_dataset.where(prog: "Vnet::LoadBalancerHealthProbes").all.select { |st| st.stack[0]["subject_id"] == id && st.stack[0]["vm_id"] == vm.id }.map(&:destroy)
```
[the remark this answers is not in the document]
furkansahin: You're right, I'll move this.

prog/vnet/load_balancer_nexus.rb
```diff
@@ -57,6 +62,11 @@ def before_run
   label def destroy
     decr_destroy
     strand.children.map { _1.destroy }
+    # The following if statement makes sure that it's OK to not have dns_zone
+    # only if it's not in production.
+    if (dns_zone = Prog::Vnet::LoadBalancerNexus.dns_zone) && (Config.production? || dns_zone)
```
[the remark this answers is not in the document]
furkansahin: The reason the check is here is that Daniel wanted to make sure, in production, we fail. If we do it like it is in rewrite_dns_records, we'll omit the failure.

prog/vnet/load_balancer_health_probes.rb
furkansahin: Possible, but, initially, I wanted to keep this logic isolated as much as possible. It's a good addition for revision 2.

PR #1744, opened 2024-07-03
Put operator SSH keys to service VMs

byucesoy:
**Allow passing multiple ssh keys while setting up VM**
We used to allow passing only one ssh key while setting up a VM. For the VMs
that belongs to services, it would be useful to pass multiple ssh keys. For
example, we can pass array of operator keys to be put into the VMs.

An alternative approach would be adding SSH keys to the VM using the sshable
interface. However, that would happen after certain steps in the service setup,
thus wouldn't be useful for debugging bootstrapping issues. It would also
require each service to implement key addition logic.

**Add feature flag for passing SSH keys to VMs**
We are passing the content of this feature flag as authorized_keys to the VMs.
This allows us to operator SSH keys to the VMs that belongs to service
accounts.

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -202,6 +202,8 @@ class Vm < Sequel::Model
   def params_json(swap_size_bytes)
     topo = cloud_hypervisor_cpu_topology
 
+    project_public_keys = projects.first.get_ff_vm_public_ssh_keys || []
+
     # we don't write secrets to params_json, because it
     # shouldn't be stored in the host for security reasons.
     JSON.pretty_generate({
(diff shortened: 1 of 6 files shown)


model/vm.rb
```diff
@@ -210,7 +210,7 @@ def params_json(swap_size_bytes)
       "public_ipv4" => ip4.to_s || "",
       "local_ipv4" => local_vetho_ip.to_s.shellescape || "",
       "unix_user" => unix_user,
-      "ssh_public_key" => public_key,
+      "ssh_public_keys" => [public_key],
```
enescakir: Should we move the rhizome change to a prior commit and deploy it first?

Currently, when we deploy it, clover will send `ssh_public_keys` to the rhizome but it expects `ssh_public_key` until we run `install_rhizome` for all hosts.

byucesoy: Makes sense. I put the rhizome changes to the first commit.

fdr (approved):
nice commit separation. Would be nice to do something similar for VmHosts as well, our path dependency (from the prototype era) of using root because that's what bare metal providers tend to do should go by the wayside.

fdr:
You may also want to implement these recommendations starting with Postgres, with an eye towards adding it to all our services, and VmHosts.  https://infosec.mozilla.org/guidelines/openssh

enescakir (approved):

PR #1749, opened 2024-07-04
Load Balancer UI

furkansahin:
## Update Networking UI
To have a better UX and tell the story of networking entities that are
in close interraction with each other, we are rewriting the console UI.
Private Subnets and Firewalls are moved under a single Networking
sidebar. The view files are also moved to have a better folder
structure.

## Load Balancer UI
This commit includes load balancer UI pages and web endpoints.

Screenshot 2024-07-09 at 15 13 27

Screenshot 2024-07-09 at 15 13 49

Screenshot 2024-07-09 at 15 14 04

Screenshot 2024-07-10 at 09 19 47


views/networking/firewall/create.erb
```diff
@@ -4,7 +4,7 @@
   <%== render(
     "components/breadcrumb",
     locals: {
-      back: "#{@project_data[:path]}/firewall",
+      back: "#{@project_data[:path]}/networking?tab=firewalls",
```
enescakir, line 7: 🔫

views/networking/firewall/index.erb
```diff
@@ -0,0 +1,62 @@
...
+      <%== render(
+        "components/page_header",
+        locals: {
+          title: "",
+          right_items: has_project_permission("Firewall:create") ? [
+            render("components/button", locals: { text: "Create Firewall", link: "firewall/create" })
+          ] : []
+        }
+      ) %>
+    </div>
```
enescakir, line 47: We shouldn't use `components/page_header` component like that. You can use `components/button` component with right alignment.

views/networking/private_subnet/index.erb
```diff
@@ -49,17 +31,25 @@
...
+      locals: {
+        title: "",
+        right_items: has_project_permission("PrivateSubnet:create") ? [
+          render("components/button", locals: { text: "Create Private Subnet", link: "private-subnet/create" })
+        ] : []
+      }
+    ) %>
+  </div>
+  <% else %>
+    <div class="text-center py-6">
```
enescakir, line 46: Why not using `components/empty_state` component?

views/networking/tabbar.erb
```diff
@@ -0,0 +1,23 @@
+<div>
+  <div class="hidden">
+    <label for="tabs" class="sr-only">Select a tab</label>
+    <!-- Use an "onChange" listener to redirect the user to the selected tab URL. -->
+    <select id="tabs" name="tabs" class="block w-full rounded-md border-gray-300 focus:border-orange-600 focus:ring-orange-600">
+      <option selected>private_subnets</option>
+      <option>firewalls</option>
+    </select>
+  </div>
```
enescakir: This mobile select doesn't work without JavaScript code.

spec/routes/web/private_subnet_spec.rb
```diff
@@ -45,7 +45,7 @@
       it "can list no private subnets" do
         visit "#{project.path}/private-subnet"
 
-        expect(page.title).to eq("Ubicloud - Private Subnets")
+        expect(page.title).to eq("Ubicloud - Networking")
```
enescakir, line 48: Why don't we keep title as `Ubicloud - Private Subnets`? It is private subnets page  at all.

routes/web/project/load_balancer.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class CloverWeb
+  hash_branch(:project_prefix, "load-balancer") do |r|
+    r.get true do
+      lbs = @project.private_subnets.flat_map { |ps| ps.load_balancers_dataset.authorized(@current_user.id, "LoadBalancer:view").all }
```
enescakir, line 6: Why don't use `@project.load_balancers_dataset.authorized(@current_user.id, "LoadBalancer:view").all` directly? Since we associate load balancers with the project, and it has separate permissions than private subnets. I think it's better approach.

routes/web/project/load_balancer.rb
```diff
@@ -0,0 +1,44 @@
...
+      @lbs = Serializers::LoadBalancer.serialize(lbs, {include_path: true})
+
+      view "networking/load_balancer/index"
+    end
+
+    r.post true do
+      Authorization.authorize(@current_user.id, "LoadBalancer:create", @project.id)
+
+      ps = PrivateSubnet.from_ubid(r.params["private-subnet-id"])
```
enescakir, line 15: Should we check if this person has permission for this private subnet?

routes/web/project/load_balancer.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      lb = Prog::Vnet::LoadBalancerNexus.assemble(
+        ps.id,
+        name: r.params["name"],
+        algorithm: r.params["algorithm"],
+        src_port: r.params["src-port"],
```
enescakir, line 21: We are migrating to underscore in parameter names. Can you change them?
https://github.com/ubicloud/ubicloud/commit/6a9cc5cdb7c9b2b800e45f7122538c3ecb46aca6

views/networking/firewall/index.erb
furkansahin: Updated

routes/web/project/load_balancer.rb
furkansahin: updated

views/networking/tabbar.erb
furkansahin: Anything I can do at this point? I guess this is a general problem. Should I simply remove this?

routes/web/project/load_balancer.rb
furkansahin: Makes sense, updated.

routes/web/project/load_balancer.rb
furkansahin: Updated.

views/networking/firewall/index.erb
```diff
@@ -0,0 +1,57 @@
+<% @page_title = "Firewalls" %>
+  <%== render("networking/tabbar") %>
+    <% if @firewalls.count > 0 %>
```
enescakir, line 3: The head indentation is inconsistent between "firewall/index.erb" and "private_subnet/index.erb"

views/networking/tabbar.erb
enescakir: It's not a general problem. If you prefer to use this select, you need to write some JS code to https://github.com/ubicloud/ubicloud/blob/main/assets/js/app.js. Without this JS code, it's a dummy form input without any interaction.

Alternatively, if the tabbar looks good on smaller devices like mobiles, you can indeed remove this part.

routes/common/load_balancer_helper.rb
```diff
@@ -0,0 +1,94 @@
...
+    flash["notice"] = "'#{name}' is created"
+
+    @request.redirect "#{project.path}#{lb.path}"
+  end
+
+  def get
+    Authorization.authorize(@user.id, "LoadBalancer:view", @resource.id)
+    vms = @resource.private_subnet.vms
+    attached_vms = @resource.vms
+    @app.instance_variable_set(:@attachable_vms, Serializers::Vm.serialize(vms.reject { |vm| attached_vms.map(&:id).include?(vm.id) }))
```
enescakir, line 42: no need to call `.map` on `attached_vms` in all iterations
```suggestion
    attached_vm_ids = @resource.vms.map(&:id)
    @app.instance_variable_set(:@attachable_vms, Serializers::Vm.serialize(vms.reject { attached_vm_ids.include?(_1.id) }))
```

routes/common/load_balancer_helper.rb
```diff
@@ -0,0 +1,94 @@
...
+    @resource.incr_destroy
+    response.status = 204
+    @request.halt
+  end
+
+  def post_attach_vm
+    Authorization.authorize(@user.id, "LoadBalancer:edit", @resource.id)
+    required_parameters = %w[vm_id]
+    request_body_params = Validation.validate_request_body(params, required_parameters)
+    vm = Vm.from_ubid(request_body_params["vm_id"])
```
enescakir, line 60: Should we check permission on vm?

routes/web/project/location/load_balancer.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class CloverWeb
+  hash_branch(:project_location_prefix, "load-balancer") do |r|
+    r.on String do |lb_name|
+      pss = @project.private_subnets_dataset.where(location: @location).all
+      lb = pss.flat_map { _1.load_balancers_dataset.where { {Sequel[:load_balancer][:name] => lb_name} }.all }.first
```
enescakir, line 7: We already did similar change for the other part

```suggestion
      lb = @project.load_balancers_dataset.where(location: @location).where( {Sequel[:load_balancer][:name] => lb_name} ).first
```

views/networking/load_balancer/create.erb
```diff
@@ -0,0 +1,177 @@
+<% @page_title = "Create Load Balancer" %>
+
+<div class="space-y-1">
+  <%== render(
+    "components/breadcrumb",
+    locals: {
+      back: "#{@project_data[:path]}/networking?tab=load-balancers",
```
enescakir, line 7: 🔫

views/networking/load_balancer/create.erb
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                  }
+                ) %>
+              </div>
+              <div class="sm:col-span-1">
+                <%== render(
+                  "components/form/text",
+                  locals: {
+                    name: "health_check_down_threshold",
+                    label: "Health Check Unhealthy Threshold",
```
enescakir, line 144: nit: You can use `type: "number"` if the input is only integer. It has better UX.

views/networking/load_balancer/index.erb
```diff
@@ -0,0 +1,54 @@
...
+        "components/button", locals: { text: "Create Load Balancer", link: "load-balancer/create" })
+      %>
+    <% end %>
+  </div>
+<% else %>
+<div class="mt-6">
+  <%== render(
+    "components/empty_state",
+    locals: {
+      icon: "hero-circle-stack",
```
enescakir, line 44: not related icon for load balancers and private subnets

Screenshot 2024-07-11 at 14 34 15

Screenshot 2024-07-11 at 14 34 18

routes/web/project/location/load_balancer.rb
furkansahin: Yes but LoadBalancer does not have `location` it's location is resolved through private subnet.

views/networking/load_balancer/create.erb
```diff
@@ -0,0 +1,177 @@
...
+                      placeholder: "3"
+                    }
+                  }
+                ) %>
+              </div>
+            </div>
+            <div class="mt-6 grid grid-cols-1 gap-x-6 gap-y-8 sm:grid-cols-6">
+              <div class="sm:col-span-2">
+                <%== render(
+                  "components/form/select",
```
enescakir, line 156: Since most other inputs have default values, and this one is important, should we move it to the top, just below the name input?

views/networking/load_balancer/create.erb
```diff
@@ -0,0 +1,177 @@
...
+        [@project_data[:name], @project_data[:path]],
+        ["Load Balancers", "#{@project_data[:path]}/load-balancer"],
+        %w[Create #]
+      ]
+    }
+  ) %>
+  <%== render("components/page_header", locals: { title: "Create Load Balancer" }) %>
+</div>
+
+<div class="grid gap-6">
```
enescakir, line 19: Should we extend the inputs into the horizontal empty space?

Screenshot 2024-07-11 at 14 39 42

serializers/load_balancer.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class Serializers::LoadBalancer < Serializers::Base
+  def self.serialize_internal(lb, options = {})
+    base = {
+      id: lb.ubid,
+      name: lb.name,
+      hostname: lb.hostname,
+      ps: Serializers::PrivateSubnet.serialize(lb.private_subnet),
```
enescakir, line 9:
```suggestion
      subnet: Serializers::PrivateSubnet.serialize(lb.private_subnet),
```

serializers/load_balancer.rb
```diff
@@ -0,0 +1,27 @@
...
+      algorithm: (lb.algorithm == "round_robin") ? "Round Robin" : "Hash Based",
+      health_check_endpoint: lb.health_check_endpoint,
+      health_check_interval: lb.health_check_interval,
+      health_check_timeout: lb.health_check_timeout,
+      health_check_unhealthy_threshold: lb.health_check_down_threshold,
+      health_check_healthy_threshold: lb.health_check_up_threshold,
+      src_port: lb.src_port,
+      dst_port: lb.dst_port,
+      vms: lb.vms.map { |vm| Serializers::Vm.serialize(vm, {load_balancer: true}) }
```
enescakir, line 18: We don't use `vms` and `ps` in the load balancer index view. Can you move them to `options[:detailed]` similar to the pg serializer.

serializers/vm.rb
```diff
@@ -27,6 +27,10 @@ def self.serialize_internal(vm, options = {})
       )
     end
 
+    if options[:load_balancer]
+      base[:state] = DB[:load_balancers_vms].where(vm_id: vm.id).get(:state)
```
enescakir, line 31: It's not `vm[:state]`. We have already `state` property. I think `load_balancer_state` is more accurate for this case

views/networking/load_balancer/create.erb
furkansahin: The problem with this view is that I don't want to put so many items on the same row. I think that kills the usability. If we keep it at 2 items and make the overall view where the items listed, longer, then we will have 200 char long text boxes for just 2 char integer. So, I don't have a solution to this. I think, most important thing is to make sure the text boxes on the same row looks side by side, for most screens. Is this good enough for you?
Screenshot 2024-07-11 at 15 03 01

lib/validation.rb
```diff
@@ -173,6 +173,17 @@ def self.validate_port_range(port_range)
     end_port ? [start_port, end_port] : [start_port]
   end
 
+  def self.validate_port(port)
+    fail ValidationFailed.new({port: "Port must be between 0 to 65535"}) unless port.to_i.to_s == port.to_s
```
enescakir, line 177: The `validate_positive_int` validator and `validate_usage_limit` validator don't have the second `to_s`
```suggestion
    fail ValidationFailed.new({port: "Port must be between 0 to 65535"}) unless port.to_i.to_s == port
```

lib/validation.rb
```diff
@@ -173,6 +173,17 @@ def self.validate_port_range(port_range)
...
+    fail ValidationFailed.new({port: "Port must be between 0 to 65535"}) unless (0..65535).cover?(port.to_i)
+    port.to_i
+  end
+
+  def self.validate_positive_int(i)
+    fail ValidationFailed.new({i: "Value must be a positive integer."}) unless i.to_i.to_s == i && i.to_i > 0
```
enescakir, line 183: The key represents the name of input. Thus, when you define it as `i`, if I remember correctly, the error message won't appear below of the corresponding input.

routes/common/load_balancer_helper.rb
```diff
@@ -0,0 +1,99 @@
...
+    ).subject
+
+    flash["notice"] = "'#{name}' is created"
+
+    @request.redirect "#{project.path}#{lb.path}"
+  end
+
+  def get
+    Authorization.authorize(@user.id, "LoadBalancer:view", @resource.id)
+    vms = @resource.private_subnet.vms
```
enescakir, line 40: Should we check any vm permissions?

routes/web/project/location/load_balancer.rb
enescakir: nit: This version generates fewer database queries and is more intuitive. However, it's totally personal. The decision is up to you.

```suggestion      
      lb = @project.load_balancers_dataset.where( {Sequel[:load_balancer][:name] => lb_name} ).all.find { _1.private_subnet.location == @location}
```

serializers/load_balancer.rb
```diff
@@ -0,0 +1,30 @@
...
+      dst_port: lb.dst_port
+    }
+
+    if options[:include_path]
+      base[:path] = lb.path
+    end
+
+    if options[:detailed]
+      base[:subnet] = Serializers::PrivateSubnet.serialize(lb.private_subnet)
+      base[:vms] = lb.vms.map { |vm| Serializers::Vm.serialize(vm, {load_balancer: true}) }
```
enescakir, line 25: nit: serializers accept an array
```suggestion
      base[:vms] = Serializers::Vm.serialize(lb.vms, {load_balancer: true})
```

serializers/vm.rb
```diff
@@ -27,6 +27,10 @@ def self.serialize_internal(vm, options = {})
       )
     end
 
+    if options[:load_balancer]
+      base[:load_balancer_state] = vm.load_balancers_vms.first.state
```
enescakir, line 31: Can a virtual machine be associated with multiple load balancers? If so, I think should put this state here. Instead, we can generate a `{vm_ubid: lb_state}` dictionary in the endpoint and pass this map to the view.

views/layouts/sidebar/content.erb
```diff
@@ -28,21 +28,14 @@
             <%== render(
               "layouts/sidebar/item",
               locals: {
-                name: "Private Subnet",
+                name: "Networking",
                 url: "#{@project_data[:path]}/private-subnet",
-                is_active: request.path.start_with?("#{@project_data[:path]}/private-subnet"),
+                is_active:
+                  request.path.start_with?("#{@project_data[:path]}/private-subnet") ||
+                  request.path.start_with?("#{@project_data[:path]}/firewall"),
```
enescakir, line 35: Should we add load balancer here?

views/networking/load_balancer/show.erb
```diff
@@ -0,0 +1,136 @@
...
+  <%== render(
+    "components/kv_data_card",
+    locals: {
+      data: [
+        ["ID", @lb[:id]],
+        ["Name", @lb[:name]],
+        ["Connection String", @lb[:hostname], { copieble: true }],
+        [
+          "Private Subnet",
+          "<a href='#{@project_data[:path]}/location/#{@lb[:subnet][:location]}/private-subnet/#{@lb[:subnet][:name]}' class='text-rose-500 hover:underline'>#{@lb[:subnet][:name]}</a>",
```
enescakir, line 29: `Serializers::PrivateSubnet` has `include_path` path option available. You can use the path as `@lb[:subnet][:path]`

fdr (commented):
Oops sent an irrelevant draft here. Deleted.

(What is still true about that draft is I'm not skilled enough to offer mechanical remarks about page organization)

I do appreciate the ideas advancing cohesion in areas that grew one by one previously. Composition has been novel for us in the API and UI to date.

serializers/vm.rb
furkansahin: Nope, single load balancer per VM.

enescakir: Then I think we need to do `load_balancers_vms.first` relation singular  maybe `many_to_one`

furkansahin: Updated!

lib/validation.rb
```diff
@@ -173,6 +173,23 @@ def self.validate_port_range(port_range)
...
+  def self.validate_health_check_parameter(parameter_name, parameter_value)
+    unless validate_positive_int(parameter_value)
+      fail ValidationFailed.new({parameter_name => "Value must be a positive integer."})
+    end
+    parameter_value.to_i
+  end
+
+  def self.validate_positive_int(parameter)
+    parameter.to_i.to_s == parameter && parameter.to_i > 0
+  end
```
enescakir, line 191: Our validation methods raise exception if the validation failed. `validate_positive_int` looks like antipattern. Since it's used in a single place, we can it in that place. Or you can rename it as something like `self.positive_int?` 

```suggestion
  def self.validate_health_check_parameter(name, value)
    unless value.to_i.to_s == value && value.to_i > 0
      fail ValidationFailed.new({name => "Value must be a positive integer."})
    end
    value.to_i
  end
```

model/vm.rb
```diff
@@ -14,7 +14,7 @@ class Vm < Sequel::Model
   one_to_many :active_billing_records, class: :BillingRecord, key: :resource_id do |ds| ds.active end
   one_to_many :pci_devices, key: :vm_id, class: :PciDevice
   many_to_many :load_balancers, left_key: :vm_id, right_key: :load_balancer_id, join_table: :load_balancers_vms
```
enescakir, line 16: It should be singular too.

model/vm.rb
```diff
@@ -14,7 +14,7 @@ class Vm < Sequel::Model
...
-  one_to_many :load_balancers_vms, key: :vm_id, class: LoadBalancersVms
+  one_to_one :load_balancers_vms, key: :vm_id, class: LoadBalancersVms
```
enescakir, line 17: It's not a blocker and also I know the class name is plural. But vm has `one_to_one :load_balancers_vms` and load balancer has `one_to_many :load_balancers_vms, key: :load_balancer_id, class: LoadBalancersVms`. Their name are same but one returns an array others return a single object. It feels like counter-intuitive. Should we rename it?

furkansahin: You're right but the `load_balancer.load_balancers_vms` need to return multiple items but the `vm.load_balancers_vms` will have a single item. They are both resolved through the same join table. I would like to keep it as is for now.

enescakir (approved):

PR #1752, opened 2024-07-05
Assemble Pages with correct Prog

bsatzger:
Assume ProgA registers a deadline and then pushes another ProgB. If that deadline expires while ProgB is executing, the Page is created with a reference to ProgB instead of ProgA. As a result, the Page has misleading information and even if ProgA reaches the target state, the Page is never resolved. This commit fixes this issue by walking back the links on the stack to identify the Prog that has registered the deadline.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -92,13 +92,17 @@ SQL
       modified!(:stack)
     end
 
+    effective_prog = prog
     stack.each do |frame|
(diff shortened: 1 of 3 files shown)


fdr (approved):
good catch

byucesoy (approved):

enescakir (approved):

PR #1759, opened 2024-07-09
Making firewall and firewall rules location based resource

velioglu:
Firewalls were initially implemented as a global resource, though it causes a problem while creating them as a part of location based resource, like a default firewall for VMs, when the multi-region backend will be implemented. It would cause each VM creation request to go global database which significantly increases the dependency on global database.

With this change, firewalls are updated to be a location based resource instead of a global one. Since firewall rules are managed as a part of firewall, that change also updates firewall rules to be a location based one. Note that, location information is not added to firewall rule models since they are not managed separately but as a part of firewall.

While merging the PR, location information will be added to existing firewalls.


routes/api/project/location/private_subnet.rb
```diff
@@ -37,8 +37,8 @@ class CloverApi
           if request_body_params["firewall_id"]
             firewall_id = request_body_params["firewall_id"]
             fw = Firewall.from_ubid(firewall_id)
-            unless fw
-              fail Validation::ValidationFailed.new(firewall_id: "Firewall with id \"#{firewall_id}\" not found")
+            unless fw && fw.location == @location
+              fail Validation::ValidationFailed.new(firewall_id: "Firewall with id \"#{firewall_id}\" and location \"#{@location}\" is not found")
```
furkansahin, line 41:
```suggestion
              fail Validation::ValidationFailed.new(firewall_id: "Firewall with id \"#{firewall_id}\" is not found in location \"#{@location}\"")
```

model/firewall.rb
```diff
@@ -12,14 +12,18 @@ class Firewall < Sequel::Model
   include Authorization::TaggableMethods
   include Authorization::HyperTagMethods
   def hyper_tag_name(project)
-    "project/#{project.ubid}/firewall/#{ubid}"
+    "project/#{project.ubid}/location/#{display_location}/firewall/#{name}"
```
enescakir, line 15: You need to update existing hyper tags

migrate/20240701_add_firewall_location.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:firewall) do
+      add_column :location, :text, null: false
```
fdr, line 6: `collation '"C"'`, right? (as irritating as it is)...

fdr (approved):
It's a bit hard to review, but that seems to be hard to avoid given the movement of things between files.

What's your deploy idea for this? It looks pretty hairy.

migrate/20240701_add_firewall_location.rb
enescakir, line 6: Does it work with existing firewalls? IMHO, this migration should fail since the new column is `null: false` and the existing ones don't have a location

views/private_subnet/show.erb
```diff
@@ -96,7 +96,10 @@
           <tr>
             <td class="whitespace nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6" scope="row">
               <% if Authorization.has_permission?(@current_user.id, "Firewall:view", fw[:id]) %>
-                <a href="<%= "#{@project_data[:path]}/firewall/#{fw[:id]}" %>" class="text-orange-600 hover:text-orange-700"><%= fw[:name] %></a>
+                <a
+                  href="<%= "#{@project_data[:path]}/location/#{fw[:location]}/firewall/#{fw[:name]}" %>"
```
enescakir, line 100: Its serializer has `include_path` option. We can use `fw[:path]` for the second part

model/firewall.rb
velioglu: Yes, I'll be updating them after deploying the first commit and migrating existing firewalls, so it will be a part of migration.

routes/api/project/location/private_subnet.rb
velioglu: Since the suggestion implies that the firewall with that id exits but not in the given location, I'm keeping it as it is.

migrate/20240731_firewall_location_not_null.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:firewall) do
+      set_column_not_null :location
```
enescakir, line 6: Should we shift this migration to the latest commit in this PR? Even though you've updated the existing firewalls, new ones can still be created until all commits are deployed.

velioglu: Good point! I did it.

furkansahin (approved):

PR #1762, opened 2024-07-10
Fix strand delay calculation to use old schedule time instead of updated one

### LINKED ISSUE #1089: The respirate runs jobs from the future
The respirate runs jobs from the future, exponential backoff not working.

delay: -504

The current time is earlier than the scheduled time.

logs

```
Jan 10 11:44:57 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:307
Jan 10 11:45:02 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:308
Jan 10 11:45:04 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:309
Jan 10 11:45:09 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:310
Jan 10 11:45:12 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:311
Jan 10 11:45:15 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:312
Jan 10 11:45:18 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:313
Jan 10 11:45:21 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:314
Jan 10 11:45:25 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:315
Jan 10 11:45:29 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:316
Jan 10 11:45:32 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:317
Jan 10 11:45:36 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:318
Jan 10 11:45:39 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:319
Jan 10 11:45:42 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:320
Jan 10 11:45:46 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:321
Jan 10 11:45:50 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:322
Jan 10 11:45:55 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:323
Jan 10 11:45:59 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:324
Jan 10 11:46:02 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:325
Jan 10 11:46:05 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:326
Jan 10 11:46:09 ncaccnz49v9p0mcmxtma1zh9hb Vnet::NicNexus.destroy try:327
```

enescakir:
I've identified some inconsistencies between the VM creation time and the strand schedule time. There are instances where the initial strand schedule time and VM creation time differ significantly, yet the schedule delay is minimal. This led me to look deeper to our strand running code further.

We started to reload the strand at
bbf3e142bda5b408cd94150a435efca6bd40204c when we take the lease. This ensures we have the most recent version of the strand. However, we also update the try count and schedule time while taking the lease. Thus, upon taking the lease, we have a new schedule time for the strand in this context. We then calculate the delay based on strand's schedule time, so the new schedule time. This results in us observing a smaller delay than expected, or even a negative delay. As the try count increases, this negative delay also increases. This bug also provides an explanation for the the respirate picking jobs from the future mystery.

Fixes #1089

It seems the respirate may not be keeping up with the current load, but this issue has been overlooked due to this bug.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -29,7 +29,7 @@ SQL
     return false unless affected
     lease_time = affected.fetch(:lease)
 
-    Clog.emit("obtained lease") { {lease_acquired: {time: lease_time}} }
+    Clog.emit("obtained lease") { {lease_acquired: {time: lease_time, delay: Time.now - schedule}} }
     reload
 
     begin
@@ -79,7 +79,7 @@ SQL
   def unsynchronized_run
     start_time = Time.now
     prog_label = "#{prog}.#{label}"
-    Clog.emit("starting strand") { {strand: values, strand_started: {delay: start_time - schedule, prog_label: prog_label}} }
+    Clog.emit("starting strand") { {strand: values, strand_started: {prog_label: prog_label}} }
 
     if label == stack.first["deadline_target"].to_s
       if (pg = Page.from_tag_parts("Deadline", id, prog, stack.first["deadline_target"]))


model/strand.rb
```diff
@@ -79,7 +80,8 @@ def load(snap = nil)
   def unsynchronized_run
     start_time = Time.now
     prog_label = "#{prog}.#{label}"
-    Clog.emit("starting strand") { {strand: values, strand_started: {delay: start_time - schedule, prog_label: prog_label}} }
+    delay = @pre_lease_schedule ? start_time - @pre_lease_schedule : nil
+    Clog.emit("starting strand") { {strand: values, strand_started: {delay: delay, prog_label: prog_label}} }
```
bsatzger, line 84: great find! I'm wondering if we should log delay here
`Clog.emit("obtained lease") { {lease_acquired: {time: lease_time}} }` 

and potentially even get rid of this log line. 

that way we also wouldn't need `@pre_lease_schedule`. but it's definitely an improvement to the current broken log output, so I'm also ok to merge as is.

bsatzger (approved):

model/strand.rb
enescakir: Actually it's a good suggestion. I moved the delay property from 'strand_started' to 'lease_acquired'. We run the strand for 30 seconds when we acquire the lease. Therefore, delays become meaningful when we first start running the strand. It's not meaningful for consecutive runs since we only take schedule into account when take the lease.

fdr (approved):
I have also thought whether we should eliminate some of those log lines. They were super useful when the system had buggy synchronization problems (@furkansahin contributed that super-key and subtle `reload` in there), but....the code is quite "cool" now...perhaps the noise isn't worth it until we have problems?

bsatzger (approved):
thanks! we'll need to update Mezmo views after this change. Really curious to see the actual delays once this gets deployed.

PR #1767, opened 2024-07-11
Manage Postgres' subnet and firewalls through the parent entity

byucesoy:
**Allow setting firewall rules in bulk for firewalls**
Firewalls have helper methods to add and remove rules, but they increment the
update_firewall_rules semaphore for each rule added or removed. This is fine in
terms of correctness, because the operation is done in transaction and other
entities will saw the change atomically. However, it is slower, because we need
to go to database multiple times. It also adds unnecessary entries for each
incremented semaphore. Instead, I added a bulk firewall rule set method to
handle this case more efficiently.

**Attach Postgres' subnet to resource instead of servers**
We used to not keep track of the private subnet in the resource level. Instead,
we would access it through representative_server's VM. Since we are moving
towards centralizing management of firewalls and private subnets, it is better
to track the private subnet in the resource level.

**Manage PostgreSQL firewalls from the parent resource**
Each PostgresServer used to manage its own firewall and we used to attach each
firewall to the private subnet. However, we were not properly cleaning up the
firewalls when the PostgresServer was deleted, causing the old rules in the
firewall to be active for that particular subnet. We could fix that particular
bug by better managing firewall of each PostgresServer (e.g. properly cleaning
them up after the deletion of the PostgresServer). However, it is better to
centralize the management of the firewalls in the parent entity, that is
PostgresResource.

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -12,6 +12,7 @@ class PostgresResource < Sequel::Model
   one_through_one :timeline, class: PostgresTimeline, join_table: :postgres_server, left_key: :resource_id, right_key: :timeline_id
   one_to_many :firewall_rules, class: PostgresFirewallRule, key: :postgres_resource_id
   one_to_many :metric_destinations, class: PostgresMetricDestination, key: :postgres_resource_id
+  many_to_one :private_subnet
 
   plugin :association_dependencies, firewall_rules: :destroy, metric_destinations: :destroy
   dataset_module Authorization::Dataset
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -91,6 +91,13 @@ class PostgresResource < Sequel::Model
     required_standby_count_map[ha_type]
   end
 
+  def set_firewall_rules
+    vm_firewall_rules = firewall_rules.map { {cidr: _1.cidr.to_s, port_range: Sequel.pg_range(5432..5432)} }
+    vm_firewall_rules.push({cidr: "0.0.0.0/0", port_range: Sequel.pg_range(22..22)})
+    vm_firewall_rules.push({cidr: "::/0", port_range: Sequel.pg_range(22..22)})
+    private_subnet.firewalls.first.replace_firewall_rules(vm_firewall_rules)
+  end
+
   module HaType
     NONE = "none"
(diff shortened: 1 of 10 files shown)


prog/postgres/postgres_resource_nexus.rb
```diff
@@ -48,9 +48,12 @@ def self.assemble(project_id:, location:, name:, target_vm_size:, target_storage
       )
       postgres_resource.associate_with_project(project)
 
+      private_subnet_id = Prog::Vnet::SubnetNexus.assemble(Config.postgres_service_project_id, name: "#{postgres_resource.ubid}-subnet", location: location).id
+      postgres_resource.update(private_subnet_id: private_subnet_id)
+
       PostgresFirewallRule.create_with_id(postgres_resource_id: postgres_resource.id, cidr: "0.0.0.0/0")
 
-      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: timeline_access, representative_at: Time.now)
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: timeline_access, representative_at: Time.now, private_subnet_id: private_subnet_id)
```
enescakir, line 60: You already pass the resource to it. IMHO, it can use resource's private subnet directly. Do we have any case that we want to give another private subnet?

model/postgres/postgres_resource.rb
```diff
@@ -13,6 +13,7 @@ class PostgresResource < Sequel::Model
   one_to_many :firewall_rules, class: PostgresFirewallRule, key: :postgres_resource_id
   one_to_many :metric_destinations, class: PostgresMetricDestination, key: :postgres_resource_id
   many_to_one :private_subnet
+  many_to_one :firewall
```
enescakir, line 16: Do we really need to keep the firewall as seperate column? Every piece of data we add increases the cost of data tracking. Can't we directly use `private_subnet.firewalls.first` instead?

enescakir (approved):

enescakir:
Should we clean up leaked firewall? Also, will you backfill these values for the existing resources?

byucesoy:
> Should we clean up leaked firewalls? Also, will you backfill these values for the existing resources?

Yep, I'll do that.

PR #1768, opened 2024-07-11
Add more details to pages

enescakir:
### Add severity column to page table

Currently, we trigger all pages labeled with 'error' severity. However, certain pages, like capacity incidents, don't require immediate phone calls. Using the severity column, we can filter out non-urgent pages.

### Add extra data and severity to pages
The `severity` field allows us to designate the alert's severity level. PagerDuty uses this field to assess the alert's urgency. The severity field can be assigned one of the following values: `critical`, `error`, `warning`, `info` [^1]. In the past, we assigned the `error` value to all alerts.

Additionally, I added the `extra_data` parameter. This facilitates the addition of extra details to incidents, along with related resources. For instance, we can include utilization in capacity incidents.

While I usually appreciate the splat operator `*tag_parts`, its comprehension becomes challenging as the number of parameters increases. Therefore, I modified it to accept a list of tags.

[^1]: https://support.pagerduty.com/docs/dynamic-notifications#severity-and-urgency-mapping

###  Lower capacity incident severity and add utilization data

We don't have an immediate response for capacity incidents. They simply alert us when our utilization is higher than usual. If this persists, we can investigate for any unusual usage. Therefore, it's not necessary to make a phone call about it given its low severity. I've downgraded its severity to 'warning'. Additionally, I included utilization data in the incident details, which can be viewed on Slack messages.

### DIFF
--- a/migrate/20240711_add_severity_to_page.rb
+++ b/migrate/20240711_add_severity_to_page.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:page_severity, %w[critical error warning info])
+
+    alter_table(:page) do
+      add_column :severity, :page_severity, default: "error", null: false
+    end
+  end
+end
--- a/model/page.rb
+++ b/model/page.rb
@@ -29,7 +29,7 @@ class Page < Sequel::Model
     end
 
     incident = pagerduty_client.incident(OpenSSL::HMAC.hexdigest("SHA256", "ubicloud-page-key", tag))
-    incident.trigger(summary: summary, severity: "error", source: "clover", custom_details: details, links: links)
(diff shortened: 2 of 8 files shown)


migrate/20240711_add_severity_to_page.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:page) do
+      add_column :severity, :text, collate: '"C"', default: "error"
```
bsatzger, line 6: should we create an enum that only allows these values: `critical` , `error`, `warning`, `info`?

enescakir: Makes sense. Added

byucesoy (approved):

bsatzger (approved):
📟

PR #1770, opened 2024-07-12
Add custom inspect representation for UBID

enescakir:
When we get a UBID in pry, the displayed value is not useful at all. Usually, I prefer to view the UBID or UUID and use "to_s" or "to_uuid" to retrieve the value. I've updated the inspect method to display the UBID and UUID values by default. I believe this will be helpful for operators.

Before:

    [1] clover-development(main)> UBID.parse("vmy2mfzyj1yj2kv7rpevcp9afs")
    => #<UBID:0x0000000112e55610 @value=319892216702515363489663739989322781948>

After:

    [1] clover-development(main)> UBID.parse("vmy2mfzyj1yj2kv7rpevcp9afs")
    => #<UBID:Vm @ubid="vmy2mfzyj1yj2kv7rpevcp9afs" @uuid="f0a8fffa-41f4-8774-9ecf-8b3b6cb254fc">

### DIFF
--- a/ubid.rb
+++ b/ubid.rb
@@ -190,6 +190,10 @@ class UBID
     @value
   end
 
+  def inspect
+    "#<UBID:#{TYPE2CLASS[to_s[..1]] || "Unknown"} @ubid=#{to_s.inspect} @uuid=#{to_uuid.inspect}>"
+  end
+
   #
   # Utility functions
   #
--- a/spec/ubid_spec.rb
+++ b/spec/ubid_spec.rb
@@ -264,4 +264,8 @@ RSpec.describe UBID do
       described_class.decode("han2sefsk4f61k91z77vn0y978")
     }.to raise_error RuntimeError, "Couldn't decode ubid: han2sefsk4f61k91z77vn0y978"
   end
+
+  it "can be inspected" do
+    expect(described_class.parse("vmqsknkzw5164hkfnt6z6zgjps").inspect).to eq("#<UBID:Vm @ubid=\"vmqsknkzw5164hkfnt6z6zgjps\" @uuid=\"be6759ff-8509-8b74-8cdf-5d1be6fc256c\">")
+  end
 end


ubid.rb
```diff
@@ -190,6 +190,10 @@ def to_i
     @value
   end
 
+  def inspect
+    "#<UBID:#{TYPE2CLASS[to_s[..1]] || "Unknown"} @ubid=#{to_s.inspect} @uuid=#{to_uuid.inspect}>"
```
furkansahin, line 194: do you need `.inspect` here? Feels like `to_s` should be enough?

furkansahin (approved):

ubid.rb
enescakir: When we don't add `.inspect`, it prints the string without quotes. It makes hard to differentiate value and it's recommended to put all values in `.inspect` representation.

with `.inspect`: 

```ruby
[1] clover-development(main)> UBID.parse("vmy2mfzyj1yj2kv7rpevcp9afs")
=> #<UBID:Vm @ubid="vmy2mfzyj1yj2kv7rpevcp9afs" @uuid="f0a8fffa-41f4-8774-9ecf-8b3b6cb254fc">
```

without `.inspect`: 
```ruby
[1] clover-development(main)> UBID.parse("vmy2mfzyj1yj2kv7rpevcp9afs")
=> #<UBID:Vm @ubid=vmy2mfzyj1yj2kv7rpevcp9afs @uuid=f0a8fffa-41f4-8774-9ecf-8b3b6cb254fc>
```

Some recommendations:
 
> You’ll also note that we call inspect on each attribute, which ensures the value as a whole is returned as expected. If we did not do this, the name attribute would render without quotation marks. This is an issue because it makes it look like the name is a Ralph class, rather than a String.

```
Person.new("Ralph", 20)
# => #<Person @name="Ralph" @age=20>
Person.new("Ralph", 20).inspect
# => "#<Person @name=Ralph @age=20>" # <- Note the missing quotation marks on Ralph
```

furkansahin: I see, alright 🚢

bsatzger (approved):

fdr (approved):
I love it, I've been thinking about doing this a long time (so transcripts match slack searches)

PR #1771, opened 2024-07-12
Make setup/destroy cache blob storage operations idempotent

enescakir:
When a cache blob storage operation fails, it will also fail on subsequent attempts if the bucket has already been created or destroyed. This change ensures the operations are idempotent by disregarding the error when the bucket is already created or destroyed.

Idempotency is not a concern for Cloudflare API requests. We can create a new token with the same name. If the token has already been deleted, it returns a 404 error, but we accept both 200 and 404 as successful responses.


model/github/github_repository.rb
```diff
@@ -46,17 +46,24 @@ def admin_client
...
-      admin_client.create_bucket({
-        bucket: bucket_name,
-        create_bucket_configuration: {location_constraint: Config.github_cache_blob_storage_region}
-      })
+      begin
+        admin_client.create_bucket({
+          bucket: bucket_name,
+          create_bucket_configuration: {location_constraint: Config.github_cache_blob_storage_region}
+        })
+      rescue Aws::S3::Errors::BucketAlreadyOwnedByYou
```
bsatzger, line 65: from gut feeling I would have expected an error like `Aws::S3::Types::BucketAlreadyExists` in this case. but `Aws::S3::Errors::BucketAlreadyOwnedByYou` seems correct.

spec/model/github/github_repository_spec.rb
```diff
@@ -13,25 +13,44 @@
...
-    github_repository.destroy_blob_storage
+  describe ".destroy_blob_storage" do
+    it "deletes the bucket and token" do
+      expect(blob_storage_client).to receive(:delete_bucket).with(bucket: "gph0hh0ahdbj6ng2cc3rvdecr8")
+      expect(cloudflare_client).to receive(:delete_token).with(github_repository.access_key)
+      expect(github_repository).to receive(:update).with(access_key: nil, secret_key: nil)
+      github_repository.destroy_blob_storage
+    end
+
+    it "runs even if the bucket already deleted" do
```
bsatzger, line 24: maybe: succeeds if the bucket is already deleted

spec/model/github/github_repository_spec.rb
```diff
@@ -13,25 +13,44 @@
...
+          "permission_groups" => [{"id" => "2efd5506f9c8494dacb1fa10a3e7d5b6", "name" => "Workers R2 Storage Bucket Item Write"}],
+          "resources" => {"com.cloudflare.edge.r2.bucket.123_default_gph0hh0ahdbj6ng2cc3rvdecr8" => "*"}
+        }
+      ]
+      expect(cloudflare_client).to receive(:create_token).with("gph0hh0ahdbj6ng2cc3rvdecr8-token", expected_policy).and_return(["test-key", "test-secret"])
+      expect(github_repository).to receive(:update).with(access_key: "test-key", secret_key: Digest::SHA256.hexdigest("test-secret"))
+      github_repository.setup_blob_storage
+    end
+
+    it "runs even if the bucket already exists" do
```
bsatzger, line 48: maybe: succeeds if the bucket already exists

bsatzger (approved):

model/github/github_repository.rb
enescakir: It was a shock for me as well. I considered adding both exceptions, but bucket names in Cloudflare R2 are not globally unique. They are unique only within an account, so Cloudflare R2 never throws an `Aws::S3::Types::BucketAlreadyExists` error. However, since bucket names in AWS S3 are globally unique, I believe the `Aws::S3::Types::BucketAlreadyExists` error is thrown when a bucket with the same name exists in any other account.

spec/model/github/github_repository_spec.rb
enescakir: Thanks. Changed

spec/model/github/github_repository_spec.rb
enescakir: Thanks. Changed

velioglu (approved):

PR #1774, opened 2024-07-13
Put operator SSH keys to service VMs

byucesoy:
We deployed [this PR](https://github.com/ubicloud/ubicloud/pull/1744), but it caused failures in VM provisioning. This line was the culprit;

```
ssh_public_keys = params.fetch("ssh_public_keys", params.fetch("ssh_public_key"))
```

Here is what happens. Name of the variable sent from control plane was `ssh_public_key` (i.e. singular) but we are changing it to be `ssh_public_keys` (i.e plural). We added the second branch to fetch, where we fetch the `ssh_public_key` variable if `ssh_public_keys` is not found. In theory, we could deploy data plane that contains the above line first and install rhizome. This would ensure that no matter if the control plane code is old or new, the code would work. However, in the above line, right branch is evaluated first, without checking if `ssh_public_keys` exist or not. Then it throws error if `ssh_public_key` is not found. This makes sense in hindsight; after all variables of the function needs to be evaluated before the function. That means, in the most up-to-date control plane version of the code, where we use the variable name as `ssh_public_keys`, we always throw error due to lack of `ssh_public_key`. The fix is simple, we add a third branch:

```
ssh_public_keys = params.fetch("ssh_public_keys", params.fetch("ssh_public_key", ""))
```

This ensures that evaluation of second branch never throws error. However, in my opinion, having third branch has some problems;
- Three branches for fetching a simple variable is too much.
- We lose the ability to ever throw an error if neither `ssh_public_keys` nor `ssh_public_key` is sent.
- Branching is only useful during the deployment of this patch.

Thus, I added a final commit, that is supposed to be deployed last and remove the branching. The final version of the code looks like this;

```
ssh_public_keys = params.fetch("ssh_public_keys")
```

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -43,9 +43,9 @@ class VmSetup
     @vp ||= VmPath.new(@vm_name)
   end
 
-  def prep(unix_user, public_key, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices)
+  def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices)
     setup_networking(false, gua, ip4, local_ip4, nics, ndp_needed, multiqueue: max_vcpus > 1)
-    cloudinit(unix_user, public_key, nics, swap_size_bytes)
+    cloudinit(unix_user, public_keys, nics, swap_size_bytes)
     storage(storage_params, storage_secrets, true)
     hugepages(mem_gib)
     prepare_pci_devices(pci_devices)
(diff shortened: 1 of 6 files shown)


enescakir:
When you assign a default value using a block, it isn't evaluated upon calling. I learnt it from @velioglu https://github.com/ubicloud/ubicloud/pull/1614#discussion_r1608872614

```ruby
ssh_public_keys = params.fetch("ssh_public_keys") { params.fetch("ssh_public_key") }
```

rhizome/host/lib/vm_setup.rb
```diff
@@ -462,15 +462,15 @@ def generate_swap_config(swap_size_bytes)
...
+  def write_user_data(unix_user, public_keys, swap_size_bytes)
     vp.write_user_data(<<EOS)
 #cloud-config
 users:
   - name: #{yq(unix_user)}
     sudo: ALL=(ALL) NOPASSWD:ALL
     shell: /bin/bash
     ssh_authorized_keys:
-      - #{yq(public_key)}
+#{public_keys.map { "      - #{yq(_1)}" }.join("\n")}
```
fdr, line 473: Oh man, that we're still templating YAML (a prototype carryover) is such a bummer, but not for this patch to fix.

spec/prog/vm/nexus_spec.rb
```diff
@@ -227,6 +227,8 @@ def requested_disk_size(st)
       expect(vm).to receive(:nics).and_return([nic]).at_least(:once)
       expect(vm).to receive(:cloud_hypervisor_cpu_topology).and_return(Vm::CloudHypervisorCpuTopo.new(1, 1, 1, 1))
       expect(vm).to receive(:pci_devices).and_return([pci]).at_least(:once)
+      prj.set_ff_vm_public_ssh_keys(["operator_ssh_key"])
```
enescakir, line 230: The feature flag introduced in the next commit

Screenshot 2024-07-16 at 12 11 33

enescakir (approved):
You might write a brief explanation for the last commit detailing why we removed that part, for example we keep it at first place for zero downtime deployment etc. This will help clarify the reason behind the removal for anyone reviewing individual commits in the future via blame

spec/prog/vm/nexus_spec.rb
byucesoy: Moved the test changes to next commit.

byucesoy:
> You might write a brief explanation for the last commit detailing why we removed that part, for example we keep it at first place for zero downtime deployment etc. This will help clarify the reason behind the removal for anyone reviewing individual commits in the future via blame

Added a brief explanation. Running the E2E tests now. I'll deploy this after they pass.

PR #1779, opened 2024-07-16
Delete cache entries

enescakir:
### Delete blob storage object when cache entry destroyed

When the cache entry is destroyed, we need to delete the blob storage
object too. We clean up the object in the `after_destroy` hook. If the
cache object is already deleted, we don't need to do anything.

Also, the cache might not be committed yet. So if the multipart upload
is not completed, we abort the upload. If the upload is aborted already,
we don't need to do anything.

###  Delete all cache entries when repository delete

When the repository is deleted, we should delete all assosiated cache
entries. Also blob storage doesn't allow to delete bucket if it's not
empty, so we should delete all object before deleting the bucket. I
moved the bucket deletion to the `after_destroy` hook, so
`association_dependencies` plugin destroys all cache entries first, then
we delete the bucket.

### DIFF
--- a/model/github/github_cache_entry.rb
+++ b/model/github/github_cache_entry.rb
@@ -2,6 +2,8 @@
 
 require_relative "../../model"
 
+require "aws-sdk-s3"
+
 class GithubCacheEntry < Sequel::Model
   many_to_one :repository, key: :repository_id, class: :GithubRepository
 
@@ -13,5 +15,20 @@ class GithubCacheEntry < Sequel::Model
 
   def blob_key
     "cache/#{ubid}"
+  end
+
+  def after_destroy
+    super
+    if committed_at.nil?
+      begin
+        repository.blob_storage_client.abort_multipart_upload(bucket: repository.bucket_name, key: blob_key, upload_id: upload_id)
+      rescue Aws::S3::Errors::NoSuchUpload
+      end
+    end
+
+    begin
+      repository.blob_storage_client.delete_object(bucket: repository.bucket_name, key: blob_key)
+    rescue Aws::S3::Errors::NoSuchKey
+    end
   end
 end
--- a/model/github/github_repository.rb
+++ b/model/github/github_repository.rb
@@ -10,6 +10,8 @@ class GithubRepository < Sequel::Model
   one_to_many :runners, key: :repository_id, class: :GithubRunner
   one_to_many :cache_entries, key: :repository_id, class: :GithubCacheEntry
 
+  plugin :association_dependencies, cache_entries: :destroy
+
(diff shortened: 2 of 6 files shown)


model/github/github_cache_entry.rb
```diff
@@ -14,4 +16,19 @@ def self.ubid_type
...
+        repository.blob_storage_client.abort_multipart_upload(bucket: repository.bucket_name, key: blob_key, upload_id: upload_id)
+      rescue Aws::S3::Errors::NoSuchUpload
+      end
+    end
+
+    begin
+      repository.blob_storage_client.delete_object(bucket: repository.bucket_name, key: blob_key)
+    rescue Aws::S3::Errors::NoSuchKey
+    end
+  end
```
fdr, line 33: Oh man, complex side effects in model hooks...these tend to have a bad reputation. Is there a special reason you are doing this rather than a prog, which is designed around retries & idempotency?

fdr (commented):
I have Questions

model/github/github_cache_entry.rb
enescakir: Cache entries don't have a nexus program. Did you mean to suggest implementing a new prog such as "DestroyGithubCacheEntry"?

My primary motivation was to ensure that wherever a cache entry is destroyed (in the prog, pry, or web), the corresponding blob storage object is also deleted. I'm concerned that if it's deleted via `pry`, the operator might forget to delete the blob object.

At present, `github_repository` is destroyed in `GithubRepositoryNexus`, after which the `association_dependencies` plugin destroys `cache_entries`. This part is already executed within a prog, and it also runs in a transaction.

enescakir: The prog version may be more scalable when dealing with hundreds of cache entries.

enescakir: What do you think about something like this https://github.com/ubicloud/ubicloud/commit/379c48d9a0d3ae8586fa4b6c3ec18eca68d85192

fdr: I think they tend to be weak at composition and make error handling a lot harder. I suppose we have to think somewhat carefully about what happens if the side effect occurs but the transaction aborts, though I think in this case it's fine because it should get a retry from calling code & is idempotent.

Best I can tell, Sequel does auto-wrap the delete in a transaction, so mercifully there's no case where the record is deleted but the side effect fails.

So maybe it's okay? Cache invalidation is kinda-sorta one of the things that they are sometimes prescribed for.

Why don't we put it in and see how it feels?

fdr (approved):

velioglu (approved):

PR #1782, opened 2024-07-17
Use forwardable in the GithubRunner prog for easier access to associations

enescakir:
I didn't know about the 'forwardable' delegator when I first wrote this prog. I learned about it from PostgresResourceNexus.

It eliminates the need to write certain property accessors, and reduces line width significantly, and enhances code readability.


fdr (approved):
man, haven't thought about forwardable in a log time, probably since Heroku. No real problem with using it though.

byucesoy (approved):

PR #1783, opened 2024-07-17
Aggregate GitHub usage by installation not projects

enescakir:
We decided to aggregate daily GitHub usage to reduce the number of billing records during implementation. Currently, we aggregated them by project. However, since a project can have multiple installations, it's difficult to determine which installation is responsible for the usage. While the billing record already has a project ID column, it would be more efficient to use the installation ID to aggregate usage.


byucesoy (approved):

velioglu (approved):

furkansahin:
why is this waiting?

enescakir:
Each customer needs a breakdown of usage at different levels: repository, workflow, etc. Therefore, we may need different architecture to visualize GitHub Action usage. As the billing_record is used for invoice generation, we can keep it at a high level

PR #1785, opened 2024-07-18
Log how much time spent on each label

enescakir:
We log the duration once the strand has finished running, but this only pertains to a single run. Currently, we lack a metric to determine the time spent on each label, which is crucial for identifying the label that consumes most of our time. This PR adds the last label change timestamp to the stack and logs the duration when the strand is hopped or exited.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -106,6 +106,11 @@ SQL
       end
     end
 
+    unless stack.first["last_label_changed_at"]
+      stack.first["last_label_changed_at"] = Time.now.to_s
+      modified!(:stack)
+    end
+
     DB.transaction do
       SemSnap.use(id) do |snap|
         prg = load(snap)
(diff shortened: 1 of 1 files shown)


model/strand.rb
```diff
@@ -128,10 +133,18 @@ def unsynchronized_run
       changed_columns.delete(:schedule)
       e
     rescue Prog::Base::Hop => hp
+      last_changed_at = Time.parse(stack.first["last_label_changed_at"])
+      Clog.emit("hopped") { {strand_hopped: {duration: Time.now - last_changed_at, prog_label: prog_label}} }
```
bsatzger, line 137: should we also include the new label?

enescakir: Makes sense. I added it

bsatzger (approved):

fdr (commented):
does the metric do what you want when applied recursively, e.g. after a pop?

byucesoy (approved):

enescakir:
> does the metric do what you want when applied recursively, e.g. after a pop?

@fdr, currently, pushing from l1 to l2 results in three durations being printed. The first reflects the duration of the initial part of l1 when we push from l1 to l2. The second shows the duration of the l2 label when we pop from l2 back to l1. The final duration represents the second part of l1 when we move from l1 to any other label. I've considered whether we should only print two durations. While possible, this change would mean that any push operation would actually reflect the timing of the previous label, and complicating the code further. When I deploy my continuous push PR (it waited too long because of me :(), it will display two durations anyway: the duration of the l1 label before the push, and the duration of the l2 label before the pop.

PR #1786, opened 2024-07-18
Delete any cache that not accessed in the last 7 days and keep usage at 10GB

enescakir:
We offer 10GB of free cache storage per repository each week, similar to GitHub's approach.

Any cache that hasn't been accessed in the past 7 days is removed. Furthermore, when cache usage exceeds 10GB, we delete the oldest cache until the usage falls below 10GB.

GithubRepositoryNexus already checks for jobs every 5 minutes. I think it can check cache usage too. If we find that every 5 minutes is too frequent, we can adjust this timing in the future.

### DIFF
--- a/prog/github/github_repository_nexus.rb
+++ b/prog/github/github_repository_nexus.rb
@@ -80,6 +80,25 @@ class Prog::Github::GithubRepositoryNexus < Prog::Base
     @polling_interval = (remaining_quota < 0.5) ? 15 * 60 : 5 * 60
   end
 
+  def cleanup_cache
+    # Destroy cache entries not accessed in last 7 days or
+    # created more than 7 days ago and not accessed yet.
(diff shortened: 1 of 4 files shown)


prog/github/github_repository_nexus.rb
```diff
@@ -80,6 +80,25 @@ def check_queued_jobs
     @polling_interval = (remaining_quota < 0.5) ? 15 * 60 : 5 * 60
   end
 
+  def check_cache_usage
```
velioglu, line 83: Nit: Name of the function can be updated as we are not checking the cache usage but also destroy it if necessary. `cleanup_cache` might be a good candidate

velioglu (approved):

prog/github/github_repository_nexus.rb
enescakir: Makes sense. I renamed it

PR #1787, opened 2024-07-18
Correct associations in Address and AssignedVmAddress models

enescakir:
### Make one_to_many associations of addresses plural

`one_to_many` associations return an array of objects, so it makes sense
to name them in the plural form. This way, it's clear that the method
returns multiple objects.

### Fix vm association in AssignedVmAddress
The current configuration of the `vm` association in the
`AssignedVmAddress` is incorrect, leading to the below exception. This
issue arises because the `key` parameter corresponds to the column name
in the `Vm` table, while the `primary_key` refers to the column name in
the `AssignedVmAddress` table.

    [12] clover-development(main)> AssignedVmAddress.first.vm
    Sequel::DatabaseError: PG::UndefinedColumn: ERROR:  column vm.assigned_vm_address_id does not exist
    LINE 1: SELECT * FROM "vm" WHERE ("vm"."assigned_vm_address_id" = $1...
                                      ^
    from /Users/enescakir/.asdf/installs/ruby/3.2.4/lib/ruby/gems/3.2.0/gems/sequel-5.81.0/lib/sequel/adapters/postgres.rb:171:in `exec_params'
    Caused by PG::UndefinedColumn: ERROR:  column vm.assigned_vm_address_id does not exist
    LINE 1: SELECT * FROM "vm" WHERE ("vm"."assigned_vm_address_id" = $1...
                                      ^
    from /Users/enescakir/.asdf/installs/ruby/3.2.4/lib/ruby/gems/3.2.0/gems/sequel-5.81.0/lib/sequel/adapters/postgres.rb:171:in `exec_params'

### DIFF
--- a/model/address.rb
+++ b/model/address.rb
@@ -3,8 +3,8 @@
 require_relative "../model"
 
 class Address < Sequel::Model
-  one_to_many :assigned_vm_address, key: :address_id, class: :AssignedVmAddress
-  one_to_many :assigned_host_address, key: :address_id, class: :AssignedHostAddress
+  one_to_many :assigned_vm_addresses, key: :address_id, class: :AssignedVmAddress
+  one_to_many :assigned_host_addresses, key: :address_id, class: :AssignedHostAddress
 
   include ResourceMethods
 end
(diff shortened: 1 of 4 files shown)


furkansahin (approved):

fdr (commented):
typo: "assosiations"

fdr (approved):
spelling issue aside, seems sensible

PR #1788, opened 2024-07-18
Load balancer Refactor

furkansahin:
## Refactor Load Balancer Create Page
Screenshot 2024-07-18 at 16 55 44

## Refactor Load Balancer Show page
Screenshot 2024-07-18 at 16 55 57

## Update Load Balancer hostname to use subnet.ubid
Load Balancer names are unique in a given project. Furthermore, there
must be a private subnet that a load balancer is created under. Instead
of generating unique hostnames per load balancer, we are reusing the
private subnet ubid. This will be useful in future when we introduce the
TLS integration and manage certificates for the customer.

## Decrease deleted VMs' evacuation time to 30 seconds
We started with 10 minutes of wait time but this is not necessary. The
main reason is, the DNS TTL is set to 10 seconds. Therefore, 30 seconds
is more than enough to evacuate a VM.

### DIFF
--- a/views/networking/load_balancer/create.erb
+++ b/views/networking/load_balancer/create.erb
@@ -54,8 +54,8 @@
                 ) %>
               </div>
             </div>
-            <div class="mt-6 grid grid-cols-1 gap-x-6 gap-y-8 sm:grid-cols-6">
-              <div class="col-span-6 md:col-span-2 xl:col-span-1">
+            <div class="mt-6 grid grid-cols-2 gap-x-6 gap-y-8 sm:grid-cols-6">
+              <div class="col-span-6 md:col-span-4 xl:col-span-2">
                 <%== render(
                   "components/form/select",
                   locals: {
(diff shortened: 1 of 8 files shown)


bsatzger:
Around "Health Check Endpoint", should we somewhere mention HTTP? Like "Make sure it returns status code 200 in response to an HTTP request.". And/or we name it "HTTP Health Check Endpoint".

furkansahin:
@bsatzger updated accordingly. Thanks for the suggestion.

views/networking/load_balancer/create.erb
```diff
@@ -54,8 +54,8 @@
                 ) %>
               </div>
             </div>
-            <div class="mt-6 grid grid-cols-1 gap-x-6 gap-y-8 sm:grid-cols-6">
-              <div class="col-span-6 md:col-span-2 xl:col-span-1">
+            <div class="mt-6 grid grid-cols-2 gap-x-6 gap-y-8 sm:grid-cols-6">
```
enescakir, line 57: In this PR, I tried to locate where we remove health check input fields. However, I realized that this removal is in the API PR. Should we move that commit to this PR?

views/networking/load_balancer/create.erb
```diff
@@ -65,27 +65,22 @@
...
-              </div>
             </div>
+          </div>
+        </div>
+      </div>
+      <div class="px-4 py-5 mt-6 sm:p-6">
+        <div class="space-y-12">
+          <div>
+            <h2 class="text-base font-semibold leading-7 text-gray-900">Forwarding Rule</h2>
+            <p class="mt-1 text-sm leading-6 text-gray-600">Set how the traffic will be routed from the load balancer to your virtual machines.</p>
```
bsatzger, line 76: Maybe: Configures the routing of traffic from the load balancer to your virtual machines.

PR #1789, opened 2024-07-18
Load Balancer API Updates for the Terraform Provider

furkansahin:
This commit tweaks some of the endpoints to accommodate Terraform
Provider implementation in an easier way. It also contains an additional
PATCH endpoint which is a much better approach than implementing
individual resource types to handle /attach-vm /detach-vm endpoints.
Another change is using the vm ubids in an array form instead of
serializing multiple VMs in the response body for a load balancer. It is
still used in the WEB endpoints but not in the APIs.

### DIFF
--- a/routes/common/load_balancer_helper.rb
+++ b/routes/common/load_balancer_helper.rb
@@ -58,6 +58,45 @@ class Routes::Common::LoadBalancerHelper < Routes::Common::Base
     end
   end
 
+  def patch
+    Authorization.authorize(@user.id, "LoadBalancer:edit", @resource.id)
+    request_body_params = Validation.validate_request_body(params, %w[algorithm src_port dst_port health_check_endpoint vms])
+    @resource.update(
+      algorithm: request_body_params["algorithm"],
+      src_port: Validation.validate_port(:src_port, request_body_params["src_port"]),
+      dst_port: Validation.validate_port(:dst_port, request_body_params["dst_port"]),
+      health_check_endpoint: request_body_params["health_check_endpoint"]
+    )
+
+    request_body_params["vms"].each do |vm_id|
+      vm_id = vm_id.delete("\"")
+      vm = Vm.from_ubid(vm_id)
+      unless vm
+        response.status = 404
+        @request.halt
+      end
+
+      Authorization.authorize(@user.id, "Vm:view", vm.id)
+      if vm.reload.load_balancer && vm.load_balancer.id != @resource.id
+        fail Validation::ValidationFailed.new("vm_id" => "VM is already attached to a load balancer")
+      elsif vm.load_balancer && vm.load_balancer.id == @resource.id
+        next
+      end
+
+      @resource.add_vm(vm)
+    end
+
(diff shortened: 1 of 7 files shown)


views/networking/load_balancer/show.erb
```diff
@@ -26,10 +26,10 @@
         ["Connection String", @lb[:hostname], { copieble: true }],
         [
           "Private Subnet",
-          "<a href='#{@project_data[:path]}#{@lb[:subnet][:path]}' class='text-rose-500 hover:underline'>#{@lb[:subnet][:name]}</a>",
+          "<a href='#{@project_data[:path]}/location/#{@lb[:location]}/private-subnet/#{@lb[:subnet]}' class='text-rose-500 hover:underline'>#{@lb[:subnet]}</a>",
           { escape: false }
         ],
-        ["Algorithm", @lb[:algorithm]],
+        ["Algorithm", (@lb[:algorithm] == "round_robin") ? "Round Robin" : "Hash Based"],
```
bsatzger, line 32: in the provisioning drop down, it's called "round-robin" and "hash-based". Wondering if we should call it the same everywhere, maybe "Round-robin" and "Hash-based".

routes/common/load_balancer_helper.rb
```diff
@@ -58,6 +58,45 @@ def post(name: nil)
...
+      end
+
+      Authorization.authorize(@user.id, "Vm:view", vm.id)
+      if vm.reload.load_balancer && vm.load_balancer.id != @resource.id
+        fail Validation::ValidationFailed.new("vm_id" => "VM is already attached to a load balancer")
+      elsif vm.load_balancer && vm.load_balancer.id == @resource.id
+        next
+      end
+
+      @resource.add_vm(vm)
```
bsatzger, line 86: should we move this up to replace "next"?

routes/common/load_balancer_helper.rb
```diff
@@ -66,7 +105,7 @@ def get
       vms = @resource.private_subnet.vms_dataset.authorized(@user.id, "Vm:view").all
       attached_vm_ids = @resource.vms.map(&:id)
       @app.instance_variable_set(:@attachable_vms, Serializers::Vm.serialize(vms.reject { attached_vm_ids.include?(_1.id) }))
-      @app.instance_variable_set(:@lb, Serializers::LoadBalancer.serialize(@resource, {detailed: true}))
+      @app.instance_variable_set(:@lb, Serializers::LoadBalancer.serialize(@resource, {detailed: true, vms_serialized: true}))
```
bsatzger, line 108: what do we need serialized vms for?

bsatzger (approved):

views/networking/load_balancer/show.erb
furkansahin: I have updated the create experience to have "Round Robin" and "Hash Based"

routes/common/load_balancer_helper.rb
furkansahin: we are using location, vm name and the load balancer health check state of the vm. Instead of returning a custom map, I decided to serialize the whole VM. Those are not needed for APIs but just the Web endpoint.

PR #1792, opened 2024-07-19
Quotas

byucesoy:
**Add helper methods for quota calculations**
I was not entirely sure about whether to put default quotas as config file or
as a database table. I decided to go with a config file for now, because
- It is easier to change. Does not require a migration.
- It is easier to view. You can see all the quotas in one place.
- It seldom changes. Control plane does not need to edit default quotas as
part of its normal operation.

The drawback is that, we cannot join with it, so I had to write a helper
function to calculate effective_quota_value. Though, it is small enough that
I'm not sure if the join version would be any better.

**Add quota check for VM provisionings**
There are two important notes about this commit.
1. We don't differentiate between x64 and arm64 VMs for now, as it is not
possible to request arm64 VMs yet. In all cases, we calculate the core count as
vCPU / 2. The logic would need to change when we introduce arm64 VMs.

2. There is one weakness that I'd like to address, but it would touch APIs and
would require a deeper discussion, so I'm leaving that for future. In summary
when the vm size is not specified by user, we default to standard-2, however,
the resolution is done at the prog's assemble method. We are doing the quota
check at the API layer, thus we don't know the size of the VM at this point
(if it is not specified by user). I added a branch to calculate requested cores
as 1 if the size is not specified, but this means we leak information from prog
to API. I'd like to avoid this and I think cleanest solution would be making
VM size a required field. Apart from fixing this problem, I think it would make
more sense semantically as well. It feels bit weird that Vm size is optional
in Vm creation API.

**Add quota check for Github runner provisionings**
We are calling quota_available? method with 0 as requested_additional_usage.

`project.quota_available?("GithubRunnerCores", 0)`

This looks weird, but it is intentional. In existing Github quota calculations,
we compare the total allocated cores with the core limit and allow passing the
limit once. This is because we check quota and allocate VMs in different labels
(hence different transactions) and it is difficult to enforce the quotas in the
environment with lots of concurrent requests. There are some remedies, but it
would require some refactoring that I'm not keen to do at the moment. Passing 0
as requested_additional_usage keeps the existing behavior.

**Add quota check for Postgres provisionings**
Nothing fancy, we just check the current usage and allow provisioning if the
request would not cause quota to be exceeded.

### DIFF
--- a/config/default_quotas.yml
+++ b/config/default_quotas.yml
@@ -1,0 +1,3 @@
+- { id: c088f732-0a30-454b-aa6b-542ce1d67bfb, resource_type: VmCores,           value: 16  }
+- { id: 14fa6820-bf63-41d2-b35e-4a4dcefd1b15, resource_type: GithubRunnerCores, value: 150 }
+- { id: 91d616ea-a15b-4d54-90b7-aaa1e1bd2f19, resource_type: PostgresCores,     value: 64  }
--- a/migrate/20240720_project_quota.rb
+++ b/migrate/20240720_project_quota.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_table(:project_quota) do
+      column :project_id, :uuid, null: false
+      column :quota_id, :uuid, null: false
+      column :value, :integer, null: false
+      primary_key [:project_id, :quota_id]
+    end
+
+    run "INSERT INTO project_quota (SELECT id, '14fa6820-bf63-41d2-b35e-4a4dcefd1b15', runner_core_limit FROM project WHERE runner_core_limit != 150)"
+  end
+
+  down do
+    drop_table(:project_quota)
+  end
+end
--- a/model/project.rb
+++ b/model/project.rb
@@ -18,6 +18,7 @@ class Project < Sequel::Model
   many_to_many :load_balancers, join_table: AccessTag.table_name, left_key: :project_id, right_key: :hyper_tag_id
 
   one_to_many :invoices, order: Sequel.desc(:created_at)
+  one_to_many :quotas, class: ProjectQuota, key: :project_id
 
   dataset_module Authorization::Dataset
   dataset_module Pagination
@@ -96,6 +97,26 @@ class Project < Sequel::Model
     }
 
     Invoice.new(project_id: id, content: content, begin_time: begin_time, end_time: end_time, created_at: Time.now, status: "current")
+  end
+
+  def current_resource_usage(resource_type)
+    case resource_type
+    when "VmCores" then vms.sum(&:cores)
+    when "GithubRunnerCores" then github_installations.sum(&:total_active_runner_cores)
+    when "PostgresCores" then postgres_resources.flat_map { _1.servers.map { |s| s.vm.cores } }.sum
+    else
+      raise "Unknown resource type: #{resource_type}"
+    end
+  end
+
+  def effective_quota_value(resource_type)
+    default_quota = ProjectQuota.default_quotas[resource_type]
+    override_quota_value = quotas_dataset.first(quota_id: default_quota["id"])&.value
+    override_quota_value || default_quota["value"]
+  end
+
+  def quota_available?(resource_type, requested_additional_usage)
+    effective_quota_value(resource_type) >= current_resource_usage(resource_type) + requested_additional_usage
   end
 
   def self.feature_flag(*flags)
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -125,12 +125,12 @@ class Prog::Vm::GithubRunner < Prog::Base
   end
 
   label def start
-    hop_wait_concurrency_limit unless concurrency_available?
+    hop_wait_concurrency_limit unless quota_available?
     hop_allocate_vm
   end
 
   label def wait_concurrency_limit
-    hop_allocate_vm if concurrency_available?
+    hop_allocate_vm if quota_available?
 
     # check utilization, if it's high, wait for it to go down
     utilization = VmHost.where(allocation_state: "accepting", arch: label_data["arch"]).select_map {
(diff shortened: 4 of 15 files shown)


spec/model/project_spec.rb
```diff
@@ -74,4 +74,32 @@
       expect(project.default_location).to eq Option.locations.first.name
     end
   end
+
+  it "calculates current resource usage" do
```
enescakir, line 78: Could you add a test to confirm that all quota IDs are unique, similar to what we do in billing rates?

config/default_quotas.yml
```diff
@@ -0,0 +1,3 @@
+- { id: c088f732-0a30-454b-aa6b-542ce1d67bfb, resource_type: VmCores,           value: 8   }
+- { id: 14fa6820-bf63-41d2-b35e-4a4dcefd1b15, resource_type: GithubRunnerCores, value: 150 }
```
enescakir, line 2: Can we assert that all `id` and `resource_type` values are unique? Alternatively, could there be multiple instances of the same `resource_type` with different IDs? Is there any scenario for that?

model/project.rb
```diff
@@ -98,6 +99,26 @@ def current_invoice
...
+    when "GithubRunnerCores" then github_installations.sum(&:total_active_runner_cores)
+    when "PostgresCores" then postgres_resources.flat_map { _1.private_subnet.vms.map(&:cores) }.sum
+    else
+      raise "Unknown resource type: #{resource_type}"
+    end
+  end
+
+  def effective_quota_value(resource_type)
+    default_quota = ProjectQuota.default_quotas[resource_type]
+    overide_quota_value = quotas_dataset.where(quota_id: default_quota["id"]).first&.value
```
enescakir, line 114: nit
```suggestion
    overide_quota_value = quotas_dataset.first(quota_id: default_quota["id"])&.value
```

enescakir, line 114: Is it a typo? I forgot to correct one 😄 
```suggestion
    override_quota_value = quotas_dataset.where(quota_id: default_quota["id"]).first&.value
```

model/project.rb
```diff
@@ -98,6 +99,26 @@ def current_invoice
     Invoice.new(project_id: id, content: content, begin_time: begin_time, end_time: end_time, created_at: Time.now, status: "current")
   end
 
+  def current_resource_usage(resource_type)
+    case resource_type
+    when "VmCores" then vms.sum(&:cores)
+    when "GithubRunnerCores" then github_installations.sum(&:total_active_runner_cores)
+    when "PostgresCores" then postgres_resources.flat_map { _1.private_subnet.vms.map(&:cores) }.sum
```
enescakir, line 106: Can't we use `PostgresServer` relation? I find it more intuitive.
```suggestion
    when "PostgresCores" then postgres_resources.flat_map { _1.servers.sum { |ps| ps.vm.cores } }.sum
```

prog/vm/github_runner.rb
```diff
@@ -165,9 +165,16 @@ def before_run
...
     github_runner.installation.project_dataset.for_update.all
-    github_runner.installation.project.runner_core_limit > github_runner.installation.project.github_installations.sum(&:total_active_runner_cores)
+    # In existing Github quota calculations, we compare total allocated cores
+    # with the core limit and allow passing the limit once. This is because we
+    # check quota and allocate VMs in different labels hence transactions and it
+    # is difficult to enforce quotas ih the environment with lots of concurrent
+    # requests. There are some remedies, but it would require some refactoring
+    # that I'm not keen to do at the moment. Although it looks weird, passing 0
+    # as requested_additional_usage keeps the existing behavior.
+    github_runner.installation.project.quota_available?("GithubRunnerCores", 0)
```
enescakir, line 177: If I merge my pull request before yours. https://github.com/ubicloud/ubicloud/pull/1782

```suggestion
    project.quota_available?("GithubRunnerCores", 0)
```

prog/vm/github_runner.rb
```diff
@@ -165,9 +165,16 @@ def before_run
...
   end
 
-  def concurrency_available?
+  def quota_available?
     github_runner.installation.project_dataset.for_update.all
-    github_runner.installation.project.runner_core_limit > github_runner.installation.project.github_installations.sum(&:total_active_runner_cores)
+    # In existing Github quota calculations, we compare total allocated cores
+    # with the core limit and allow passing the limit once. This is because we
+    # check quota and allocate VMs in different labels hence transactions and it
+    # is difficult to enforce quotas ih the environment with lots of concurrent
```
enescakir, line 173: Typo `the quotas ih the environment`

prog/vm/github_runner.rb
```diff
@@ -165,9 +165,16 @@ def before_run
...
+    # requests. There are some remedies, but it would require some refactoring
```
enescakir, line 174: > it is difficult to enforce quotas ih the environment with lots of concurrent requests.

We have already locked the project using `for_update`. Isn't it enforced?

routes/common/postgres_helper.rb
```diff
@@ -34,6 +34,9 @@ def post(name: nil)
     request_body_params = Validation.validate_request_body(params, required_parameters, allowed_optional_parameters)
     parsed_size = Validation.validate_postgres_size(request_body_params["size"])
 
+    requested_postgres_core_count = parsed_size.vcpu / 2
+    Validation.validate_core_quota(project, "PostgresCores", requested_postgres_core_count)
```
enescakir, line 38: If HA is enabled, should we also consider the standby core counts?

migrate/20240720_project_quota.rb
```diff
@@ -0,0 +1,18 @@
...
+Sequel.migration do
+  up do
+    create_table(:project_quota) do
+      column :project_id, :uuid, null: false
+      column :quota_id, :uuid, null: false
+      column :value, :integer, null: false
+      primary_key [:project_id, :quota_id]
+    end
+
+    run "INSERT INTO project_quota (SELECT id, '14fa6820-bf63-41d2-b35e-4a4dcefd1b15', runner_core_limit AS quota_id FROM project WHERE runner_core_limit != 150)"
```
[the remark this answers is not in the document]
byucesoy: It was typo, removed it.

config/default_quotas.yml
byucesoy: In the future, we might have other fields to split resource types. For now they are unique, so I added a test case.

prog/vm/github_runner.rb
byucesoy: Yes, but after we process the label, we commit the transaction, which releases the lock. We pick the VM in another state. If quota is available, multiple progs can move to `allocate_vm`

prog/vm/github_runner.rb
```diff
@@ -125,12 +125,12 @@ def before_run
   end
 
   label def start
-    hop_wait_concurrency_limit unless concurrency_available?
+    hop_wait_concurrency_limit unless quota_available?
```
fdr, line 128: " Add quota check for Github runner provisionings" it seems like the commit previous to this one (that adds "dark" code uncalled), pause, and then deploying "Add quota check" would be a somewhat safe deploy (& rollback) strategy.

fdr (approved):
I thought it was informatively justified (reading patch by patch) and phased all around.

enescakir (approved):

PR #1794, opened 2024-07-20
Quotas UI

byucesoy:
**Disable VM options from UI if user has insufficient quota**
We still display the options, but they are disabled/grayouted if the user has
insufficient quota.

**Disable Postgres options from UI if user has insufficient quota**
We still display the options, but they are disabled/grayouted if the user has
insufficient quota.

**Add certain colors and width percentages to safelist**
While rendering the progress bar, we set the width property of it to percentage
value. We cannot do it via inline style because of the security policy. We need
to set it via tailwind class. However, since the value is dynamic and tailwind
class is not known at the compile time, thus not added to the generated css. To
solve this, we add certain colors and width percentages to the safelist.

**Add overall quota information to UI**

image
image

### DIFF
--- a/routes/common/postgres_helper.rb
+++ b/routes/common/postgres_helper.rb
@@ -227,6 +227,7 @@ class Routes::Common::PostgresHelper < Routes::Common::Base
     Authorization.authorize(@user.id, "Postgres:create", project.id)
     @app.instance_variable_set(:@prices, @app.fetch_location_based_prices("PostgresCores", "PostgresStorage"))
     @app.instance_variable_set(:@has_valid_payment_method, project.has_valid_payment_method?)
+    @app.instance_variable_set(:@enabled_postgres_sizes, Option::VmSizes.select { project.quota_available?("PostgresCores", _1.vcpu / 2) }.map(&:name))
     @app.view "postgres/create"
   end
 
--- a/routes/web/project.rb
+++ b/routes/web/project.rb
@@ -35,6 +35,13 @@ class CloverWeb
 
       @project_data = Serializers::Project.serialize(@project, {include_path: true})
       @project_permissions = Authorization.all_permissions(@current_user.id, @project.id)
+      @quotas = ["VmCores", "PostgresCores"].map {
+        {
+          resource_type: _1,
+          current_resource_usage: @project.current_resource_usage(_1),
+          quota: @project.effective_quota_value(_1) * 2
+        }
+      }
 
       r.get true do
         Authorization.authorize(@current_user.id, "Project:view", @project.id)
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -11,4 +11,10 @@ module.exports = {
     require('@tailwindcss/forms'),
     require('@tailwindcss/typography'),
   ],
+  safelist: [
+    ...[...Array(101).keys()].flatMap(i => `w-[${i}%]`),
+    {
+      pattern: /bg-[a-z]+-500/,
+    }
+  ]
 }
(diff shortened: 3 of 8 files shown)


furkansahin:
I'm not sure about the different colours. The problem is, what would we pick as colours when we have 2 more services? They will get mixed at some point. I feel like it may be a better idea to have a single colour.

furkansahin:
Oh, alright then!

tailwind.config.js
```diff
@@ -11,4 +11,10 @@ module.exports = {
     require('@tailwindcss/forms'),
     require('@tailwindcss/typography'),
   ],
+  safelist: [
+    ...[...Array(101).keys()].flatMap(i => `w-[${i}%]`),
```
fdr, line 15: Oh man, is that how it turns out in JavaScript? Well, could be worse I guess.

fdr (approved):
Interesting that you spare the effort to make a progress-bar-type-thing appear, but not opposed, it comes in light enough, it's a bit less brutal than the alternative.

I'm probably not able to give as good feedback as others on a lot of this though.

views/components/progress_bar.erb
```diff
@@ -0,0 +1,10 @@
+<% progress = [100, 100 * numerator / denominator].min %>
+<% color = (progress < 60) ? "bg-blue-500" : (progress < 80) ? "bg-yellow-500" : "bg-red-500" %>
```
enescakir, line 2: I was just as confused as Furkan. IMHO, it's better to use a single common orange color rather than multiple ones. However, this is purely subjective.

views/components/progress_bar.erb
```diff
@@ -0,0 +1,10 @@
...
+<div class="flex justify-between mb-1">
+  <span class="text-base font-medium text-blue-700 dark:text-white"><%= title %></span>
+  <span class="text-sm font-medium text-blue-700 dark:text-white"><%= numerator %>/<%= denominator %>
```
enescakir, line 5: Do we have a particular reason for making them blue? Blue gives the impression that they are clickable to me. I'd prefer gray, black, or orange instead. But it's totally personal

routes/common/postgres_helper.rb
```diff
@@ -227,6 +227,7 @@ def view_create_page
     Authorization.authorize(@user.id, "Postgres:create", project.id)
     @app.instance_variable_set(:@prices, @app.fetch_location_based_prices("PostgresCores", "PostgresStorage"))
     @app.instance_variable_set(:@has_valid_payment_method, project.has_valid_payment_method?)
+    @app.instance_variable_set(:@enabled_postgres_sizes, Option::VmSizes.select { project.quota_available?("PostgresCores", _1.vcpu / 2) }.map(&:name))
```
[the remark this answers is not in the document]
byucesoy: Yes, that is correct.

views/components/progress_bar.erb
byucesoy: Changed it to default text color.

views/components/progress_bar.erb
byucesoy: I think the reason of the confusion was that I tried to show-case all 3 possible states in the screenshot. In real world, most of them would be blue and 1 yellow/red would be less confusing. I prefer to keep it this way for now.

tailwind.config.js
byucesoy: yeah :frowning_face:

routes/web/project.rb
```diff
@@ -35,6 +35,14 @@ class CloverWeb
 
       @project_data = Serializers::Project.serialize(@project, {include_path: true})
       @project_permissions = Authorization.all_permissions(@current_user.id, @project.id)
+      @quotas = ["VmCores", "PostgresCores", "GithubRunnerCores"].map {
```
enescakir, line 38: We decided to hide `GithubRunnerCores` for now

PR #1795, opened 2024-07-22
Increase quotas after successful payment

byucesoy:
**Add different quotas for different reputation levels**
Currently we have 2 levels of reputation: new and verified. Each have different
default quotas. I think we can have more levels of reputation such as sponsored
or enterprise. Depending on the reputation level, we can have different default
quotas for each of them.

I'm not entirely sold on the naming "reputation", "type" feels more appropriate
but it is too generic, so I didn't want to use it.

**Update project's reputation after successful payment**
After a successful payment, the project becomes less risky against the fraud,
so we can increase its quotas, which is done by updating its reputation as
verified.

### DIFF
--- a/migrate/20240722_add_project_reputation.rb
+++ b/migrate/20240722_add_project_reputation.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:project_reputation, %w[new verified])
+
+    alter_table(:project) do
+      add_column :reputation, :project_reputation, default: "new", null: false
+    end
+
+    run "UPDATE project SET reputation = 'verified' WHERE id IN (SELECT project_id FROM invoice WHERE (content->'cost')::float > 5)"
+  end
+end
--- a/model/invoice.rb
+++ b/model/invoice.rb
@@ -73,6 +73,8 @@ class Invoice < Sequel::Model
         "payment_intent" => payment_intent.id
       })
       save(columns: [:status, :content])
+      project.update(reputation: "verified") if amount > 5
+
       send_success_email
       return true
     end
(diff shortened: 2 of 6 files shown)


fdr (approved):
I saw this, I don't see any structural issues, but also, would I really know?

I have not thought of any terminology better than "reputation," as you wrote, "type" is without enough meaning.

model/invoice.rb
```diff
@@ -73,6 +73,8 @@ def charge
         "payment_intent" => payment_intent.id
       })
       save(columns: [:status, :content])
+      project.update(reputation: "verified") if amount > 5
```
enescakir, line 76: Why did we choose $5 as threshold?

migrate/20240722_add_project_reputation.rb
```diff
@@ -0,0 +1,13 @@
...
+
+Sequel.migration do
+  change do
+    create_enum(:project_reputation, %w[new verified])
+
+    alter_table(:project) do
+      add_column :reputation, :project_reputation, default: "new", null: false
+    end
+
+    run "UPDATE project SET reputation = 'verified'"
```
enescakir, line 11: We have existing projects without any payment or accounts known for fraud. We can update this value using the `invoice` table.

model/invoice.rb
byucesoy: It was an arbitrary choice. $5 seemed like a good enough value. I'm open to different amounts as well.

migrate/20240722_add_project_reputation.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:project_reputation, %w[new verified])
```
enescakir, line 5: We don't have to do it in this PR. In future, We might add `fraud` option as well to block new resource creation in the time between we block the account and we suspect it's a fraud.

enescakir (approved):

PR #1797, opened 2024-07-23
ACME Certificate Integration

furkansahin:
## Add ACME Cert Migration
The cert table is added.

## Add ACME Cert provisioning
This commit adds the ACME certificate provisioning integration. We are
creating a generic cert nexus. This way, we will be able to utilize the
same Certificate provisioning experience for different resources. The
most important part of this commit is the acme-client integration.

The flow of the operations for provisioning acme cert is the following;
1. Login to one of the authorities such as ZeroSSL.
2. Put an order for authorization.
3. Create the necessary dns record to solve the dns challenge.
4. Perform the dns challenge validation.
5. Put a new order to provision a certificate.
7. Finalize the order and download certs.

After the provisioning, we simply start napping for a month and we check
the expiry date every month. If it has been more than 3 months, we
assume it's expired and get into the destroy path.

Refreshing the certificates should be managed by the user of this Cert
provisioner.

## Load Balancer's Cert integration
This commit simply implements the load balancer's certificate
integration by utilizing the FSMs implemented in the previous commit.

## Add Support to download certificates in the UI
Adds a new field to the Load Balancer show page to be able to download
the private_key and the certificate files.

## Load Balancer Health Probs Protocol Support
This commit adds the health check protocol support. This is needed once
we start providing certificates. If the customer implements an
application that runs HTTPS endpoints, we must use that. For that, we
already provide the certificate to them. It's also possible that they
don't want the health checks to be performed at the TCP level without
any specific endpoint. This commit supports that as well.

## ACME Cert integration tests

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -164,4 +164,12 @@ module Config
   # Load Balancer
   optional :load_balancer_service_project_id, string
   optional :load_balancer_service_hostname, string
+
+  # ACME
+  # The following are optional because they are only needed in production.
+  # They are not needed in development or test.
+  optional :acme_email, string
+  override :acme_directory, "https://acme.zerossl.com/v2/DV90", string
+  optional :acme_eab_kid, string, clear: true
+  optional :acme_eab_hmac_key, string, clear: true
 end
--- a/migrate/20240801_add_load_balancer_cert.rb
+++ b/migrate/20240801_add_load_balancer_cert.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:certs_load_balancers) do
+      foreign_key :load_balancer_id, :load_balancer, type: :uuid, null: false
+      foreign_key :cert_id, :cert, type: :uuid, null: false
+      primary_key [:load_balancer_id, :cert_id]
+    end
+
+    create_enum(:lb_hc_protocol, %w[tcp http https])
+    alter_table(:load_balancer) do
+      add_column :health_check_protocol, :lb_hc_protocol, null: false, default: "http"
+    end
+  end
+end
--- a/model/cert.rb
+++ b/model/cert.rb
@@ -1,0 +1,20 @@
+#  frozen_string_literal: true
+
+require_relative "../model"
+
+class Cert < Sequel::Model
+  one_to_one :strand, key: :id
+
+  include ResourceMethods
+  include SemaphoreMethods
+  semaphore :destroy
+
+  plugin :column_encryption do |enc|
+    enc.column :account_key
+    enc.column :csr_key
+  end
+
+  def self.redacted_columns
+    super + [:cert]
+  end
+end
--- a/model/cert.rb
+++ b/model/cert.rb
@@ -3,7 +3,11 @@
 require_relative "../model"
 
 class Cert < Sequel::Model
+  one_through_one :load_balancer, join_table: :certs_load_balancers, left_key: :cert_id, right_key: :load_balancer_id
+  one_to_one :certs_load_balancers, key: :cert_id, class: CertsLoadBalancers
   one_to_one :strand, key: :id
+
+  plugin :association_dependencies, certs_load_balancers: :destroy
 
   include ResourceMethods
   include SemaphoreMethods
--- a/model/certs_load_balancers.rb
+++ b/model/certs_load_balancers.rb
@@ -1,0 +1,15 @@
+#  frozen_string_literal: true
+
+require_relative "../model"
+
+class CertsLoadBalancers < Sequel::Model
+  many_to_one :cert
+  include ResourceMethods
+
+  def destroy
+    DB.transaction do
+      cert.incr_destroy
+      super
+    end
+  end
+end
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -10,8 +10,10 @@ class LoadBalancer < Sequel::Model
   many_to_one :private_subnet
   one_to_many :projects, through: :private_subnet
   one_to_many :load_balancers_vms, key: :load_balancer_id, class: LoadBalancersVms
+  many_to_many :certs, join_table: :certs_load_balancers, left_key: :load_balancer_id, right_key: :cert_id
+  one_to_many :certs_load_balancers, key: :load_balancer_id, class: CertsLoadBalancers
 
-  plugin :association_dependencies, load_balancers_vms: :destroy
+  plugin :association_dependencies, load_balancers_vms: :destroy, certs_load_balancers: :destroy
 
   include ResourceMethods
   include SemaphoreMethods
@@ -19,7 +21,7 @@ class LoadBalancer < Sequel::Model
   include Authorization::HyperTagMethods
   dataset_module Authorization::Dataset
   dataset_module Pagination
-  semaphore :destroy, :update_load_balancer, :rewrite_dns_records
+  semaphore :destroy, :update_load_balancer, :rewrite_dns_records, :refresh_cert
 
   def hyper_tag_name(project)
     "project/#{project.ubid}/location/#{private_subnet.display_location}/load-balancer/#{name}"
@@ -52,5 +54,15 @@ class LoadBalancer < Sequel::Model
 
   def hostname
     "#{name}.#{private_subnet.ubid[-5...]}.#{Config.load_balancer_service_hostname}"
+  end
+
+  def need_certificates?
+    return true if certs_dataset.empty?
+
+    certs_dataset.where { created_at > Time.now - 60 * 60 * 24 * 30 * 2 }.empty?
+  end
+
+  def active_cert
+    certs_dataset.where { created_at > Time.now - 60 * 60 * 24 * 30 * 3 }.order(Sequel.desc(:created_at)).first
   end
 end
--- a/prog/vnet/cert_nexus.rb
+++ b/prog/vnet/cert_nexus.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+
+require "acme-client"
+require "openssl"
+
+class Prog::Vnet::CertNexus < Prog::Base
+  subject_is :cert
+  semaphore :destroy
+
+  REVOKE_REASON = "cessationOfOperation" # unspecified
+
+  def self.assemble(hostname, dns_zone_id)
+    unless DnsZone[dns_zone_id]
+      fail "Given DNS zone doesn't exist with the id #{dns_zone_id}"
+    end
+
+    DB.transaction do
+      cert = Cert.create_with_id(hostname: hostname, dns_zone_id: dns_zone_id)
+
+      Strand.create(prog: "Vnet::CertNexus", label: "start") { _1.id = cert.id }
+    end
+  end
+
+  def before_run
+    when_destroy_set? do
+      hop_destroy unless %w[destroy].include?(strand.label)
+    end
+  end
+
+  label def start
+    register_deadline(:wait, 10 * 60)
+
+    account_key = OpenSSL::PKey::EC.generate("prime256v1")
+    client = Acme::Client.new(private_key: account_key, directory: Config.acme_directory)
+    account = client.new_account(contact: "mailto:#{Config.acme_email}", terms_of_service_agreed: true, external_account_binding: {kid: Config.acme_eab_kid, hmac_key: Config.acme_eab_hmac_key})
+    order = client.new_order(identifiers: [cert.hostname])
+    authorization = order.authorizations.first
+    cert.update(kid: account.kid, account_key: account_key.to_der, order_url: order.url)
+    dns_challenge = authorization.dns
+    dns_zone.insert_record(record_name: dns_record_name, type: dns_challenge.record_type, ttl: 600, data: dns_challenge.record_content)
+
+    hop_wait_dns_update
+  end
+
+  label def wait_dns_update
+    dns_record = DnsRecord[dns_zone_id: dns_zone.id, name: dns_record_name + ".", tombstoned: false, data: dns_challenge.record_content]
+    if DB[:seen_dns_records_by_dns_servers].where(dns_record_id: dns_record.id).empty?
+      nap 10
+    end
+
+    dns_challenge.request_validation
+
+    hop_wait_dns_validation
+  end
+
+  label def wait_dns_validation
+    case dns_challenge.status
+    when "pending"
+      nap 10
+    when "valid"
+      csr_key = OpenSSL::PKey::RSA.new(4096)
+      csr = Acme::Client::CertificateRequest.new(private_key: csr_key, common_name: cert.hostname)
+      acme_order.finalize(csr: csr)
+      cert.update(csr_key: csr_key.to_der)
+
+      hop_wait_cert_finalization
+    else
+      Clog.emit("DNS validation failed") { {order_status: dns_challenge.status} }
+      dns_zone.delete_record(record_name: dns_record_name)
+      hop_start
+    end
+  end
+
+  label def wait_cert_finalization
+    case acme_order.status
+    when "processing"
+      nap 10
+    when "valid"
+      cert.update(cert: acme_order.certificate, created_at: Time.now)
+
+      dns_zone.delete_record(record_name: dns_record_name)
+      hop_wait
+    else
+      Clog.emit("Certificate finalization failed") { {order_status: acme_order.status} }
+      dns_zone.delete_record(record_name: dns_record_name)
+      hop_start
+    end
+  end
+
+  label def wait
+    if cert.created_at < Time.now - 60 * 60 * 24 * 30 * 3 # 3 months
+      cert.incr_destroy
+      nap 0
+    end
+
+    nap 60 * 60 * 24 * 30 # 1 month
+  end
+
+  label def destroy
+    # the reason is chosen as "cessationOfOperation"
+    acme_client.revoke(certificate: cert.cert, reason: REVOKE_REASON) if cert.cert
+
+    dns_zone.delete_record(record_name: dns_record_name) if dns_challenge
+    cert.destroy
+    pop "certificate revoked and destroyed"
+  end
+
+  def acme_client
+    # If the private_key is not yet set, we did not start the communication with
+    # ACME server yet, therefore, we return nil.
+    Acme::Client.new(private_key: OpenSSL::PKey::EC.new(cert.account_key), directory: Config.acme_directory, kid: cert.kid) if cert.account_key
+  end
+
+  def acme_order
+    # If the order_url is set, acme_client cannot be nil, so, this is nullref safe
+    acme_client.order(url: cert.order_url) if cert.order_url
+  end
+
+  def dns_challenge
+    acme_order.authorizations.first.dns
+  end
+
+  def dns_record_name
+    dns_challenge.record_name + "." + cert.hostname
+  end
+
+  def dns_zone
+    @dns_zone ||= DnsZone[cert.dns_zone_id]
+  end
+end
--- a/prog/vnet/load_balancer_health_probes.rb
+++ b/prog/vnet/load_balancer_health_probes.rb
@@ -8,12 +8,16 @@ class Prog::Vnet::LoadBalancerHealthProbes < Prog::Base
   end
 
   label def health_probe
-    response_code = nil
-    begin
-      endpoint = "#{vm.nics.first.private_ipv4.network}:#{load_balancer.dst_port}#{load_balancer.health_check_endpoint}"
-      response_code = vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} curl --max-time #{load_balancer.health_check_timeout} --silent --output /dev/null --write-out '%{http_code}' #{endpoint}")
+    response_code = begin
+      cmd = if load_balancer.health_check_protocol == "tcp"
+        "sudo ip netns exec #{vm.inhost_name} nc -z -w #{load_balancer.health_check_timeout} #{vm.nics.first.private_ipv4.network} #{load_balancer.dst_port} && echo 200 || echo 400"
+      else
+        "sudo ip netns exec #{vm.inhost_name} curl --insecure --max-time #{load_balancer.health_check_timeout} --silent --output /dev/null --write-out '%{http_code}' #{load_balancer.health_check_protocol}://#{vm.nics.first.private_ipv4.network}:#{load_balancer.dst_port}#{load_balancer.health_check_endpoint}"
+      end
+
+      vm.vm_host.sshable.cmd(cmd)
     rescue
-      response_code = "500"
+      "500"
     end
 
     vm_state, vm_state_counter = load_balancer.load_balancers_vms_dataset.where(vm_id: vm.id).get([:state, :state_counter])
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
+require "acme-client"
+require "openssl"
+
 class Prog::Vnet::LoadBalancerNexus < Prog::Base
   subject_is :load_balancer
-  semaphore :destroy, :update_load_balancer, :rewrite_dns_records
+  semaphore :destroy, :update_load_balancer, :rewrite_dns_records, :refresh_cert
 
   def self.assemble(private_subnet_id, name: nil, algorithm: "round_robin", src_port: nil, dst_port: nil,
     health_check_endpoint: "/up", health_check_interval: 30, health_check_timeout: 15,
@@ -44,8 +47,20 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
     end
 
     hop_create_new_health_probe if strand.children_dataset.count < load_balancer.vms_dataset.count
+    hop_create_new_cert if load_balancer.need_certificates?
 
     nap 5
+  end
+
+  label def create_new_cert
+    cert = Prog::Vnet::CertNexus.assemble(load_balancer.hostname, Prog::Vnet::LoadBalancerNexus.dns_zone.id).subject
+    load_balancer.add_cert(cert)
+    hop_wait_cert_provisioning
+  end
+
+  label def wait_cert_provisioning
+    nap 60 if load_balancer.need_certificates?
+    hop_wait
   end
 
   label def create_new_health_probe
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -9,7 +9,7 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
 
   def self.assemble(private_subnet_id, name: nil, algorithm: "round_robin", src_port: nil, dst_port: nil,
     health_check_endpoint: "/up", health_check_interval: 30, health_check_timeout: 15,
-    health_check_up_threshold: 3, health_check_down_threshold: 2)
+    health_check_up_threshold: 3, health_check_down_threshold: 2, health_check_protocol: "http")
 
     unless (ps = PrivateSubnet[private_subnet_id])
       fail "Given subnet doesn't exist with the id #{private_subnet_id}"
@@ -19,10 +19,10 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
 
     DB.transaction do
       lb = LoadBalancer.create_with_id(
-        private_subnet_id: private_subnet_id, name: name, algorithm: algorithm,
-        src_port: src_port, dst_port: dst_port, health_check_endpoint: health_check_endpoint,
-        health_check_interval: health_check_interval, health_check_timeout: health_check_timeout,
-        health_check_up_threshold: health_check_up_threshold, health_check_down_threshold: health_check_down_threshold
+        private_subnet_id: private_subnet_id, name: name, algorithm: algorithm, src_port: src_port, dst_port: dst_port,
+        health_check_endpoint: health_check_endpoint, health_check_interval: health_check_interval,
+        health_check_timeout: health_check_timeout, health_check_up_threshold: health_check_up_threshold,
+        health_check_down_threshold: health_check_down_threshold, health_check_protocol: health_check_protocol
       )
       lb.associate_with_project(ps.projects.first)
 
--- a/routes/common/load_balancer_helper.rb
+++ b/routes/common/load_balancer_helper.rb
@@ -25,9 +25,10 @@ class Routes::Common::LoadBalancerHelper < Routes::Common::Base
   def post(name: nil)
     Authorization.authorize(@user.id, "LoadBalancer:create", project.id)
 
-    required_parameters = %w[private_subnet_id algorithm src_port dst_port health_check_endpoint]
+    required_parameters = %w[private_subnet_id algorithm src_port dst_port health_check_protocol]
     required_parameters << "name" if @mode == AppMode::WEB
-    request_body_params = Validation.validate_request_body(params, required_parameters)
+    optional_parameters = %w[health_check_endpoint]
+    request_body_params = Validation.validate_request_body(params, required_parameters, optional_parameters)
 
     ps = PrivateSubnet.from_ubid(request_body_params["private_subnet_id"])
     unless ps
@@ -47,7 +48,8 @@ class Routes::Common::LoadBalancerHelper < Routes::Common::Base
       algorithm: request_body_params["algorithm"],
       src_port: Validation.validate_port(:src_port, request_body_params["src_port"]),
       dst_port: Validation.validate_port(:dst_port, request_body_params["dst_port"]),
-      health_check_endpoint: request_body_params["health_check_endpoint"]
+      health_check_endpoint: request_body_params["health_check_endpoint"],
+      health_check_protocol: request_body_params["health_check_protocol"]
     ).subject
 
     if @mode == AppMode::API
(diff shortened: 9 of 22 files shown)


config.rb
```diff
@@ -163,4 +163,10 @@ def self.e2e_test?
   # Load Balancer
   optional :load_balancer_service_project_id, string
   optional :load_balancer_service_hostname, string
+
+  # Let's Encrypt
+  # The following are optional because they are only needed in production.
+  # They are not needed in development or test.
+  optional :lets_encrypt_email, string
```
fdr, line 170: pretty sure you mean "acme" here rather than let's encrypt, right? For one, production ubicloud probably can't use let's encrypt per se, though it's reasonable for developers or smaller installs of clover to use it, on rate limiting grounds.

config.rb
```diff
@@ -163,4 +163,10 @@ def self.e2e_test?
...
+  override :lets_encrypt_directory, "https://acme-staging-v02.api.letsencrypt.org/directory", string
```
fdr, line 171: Ditto: acme_directory

prog/vnet/load_balancer_nexus.rb
```diff
@@ -43,11 +46,72 @@ def before_run
...
 
+  label def refresh_cert
+    private_key = OpenSSL::PKey::RSA.new(4096)
+    client = Acme::Client.new(private_key: private_key, directory: Config.lets_encrypt_directory)
+    account = client.new_account(contact: "mailto:#{Config.lets_encrypt_email}", terms_of_service_agreed: true)
+    order = client.new_order(identifiers: [load_balancer.hostname])
+    authorization = order.authorizations.first
+    load_balancer.update(kid: account.kid, private_key: private_key, order_url: order.url)
+    dns_challenge = authorization.dns
+    Prog::Vnet::LoadBalancerNexus.dns_zone.insert_record(record_name: "#{dns_challenge.record_name}.#{load_balancer.hostname}", type: dns_challenge.record_type, ttl: 600, data: dns_challenge.record_content)
```
fdr, line 66: by the way, just ot be particular, I noticed you truncated the ubids for load balancer's entropic hostnames, do you have unique constraints to prevent Undefined Stuff from happening if a collision in the first N bits is generated? (we should probably do this for VM-to-Host assignment as well, but it kinda slid in from the prototype era)

prog/vnet/load_balancer_nexus.rb
```diff
@@ -43,11 +46,72 @@ def before_run
...
+      nap 10
+    end
+
+    dns_challenge.request_validation
+
+    hop_wait_dns_validation
+  end
+
+  label def wait_dns_validation
+    dns_challenge.reload
```
fdr, line 85: why?

prog/vnet/load_balancer_nexus.rb
```diff
@@ -43,11 +46,72 @@ def before_run
...
+      csr_key = OpenSSL::PKey::RSA.new(4096)
+      csr = Acme::Client::CertificateRequest.new(private_key: csr_key, common_name: load_balancer.hostname)
+      acme_order.finalize(csr: csr)
+      load_balancer.update(csr_key: csr_key.to_pem)
+
+      hop_wait_cert_finalization
+    else
+      fail "DNS validation failed"
+    end
+  end
```
fdr, line 98: might be a good application of `case`

prog/vnet/load_balancer_nexus.rb
```diff
@@ -43,11 +46,72 @@ def before_run
...
+
+  label def wait_cert_finalization
+    acme_order.reload
```
fdr, line 101: why?

prog/vnet/load_balancer_nexus.rb
```diff
@@ -43,11 +46,72 @@ def before_run
...
+    if acme_order.status == "processing"
+      nap 10
+    elsif acme_order.status == "valid"
+      load_balancer.update(cert: acme_order.certificate, cert_created_at: Time.now)
+      puts "DNS RECORD: #{dns_challenge.record_name}.#{load_balancer.hostname}"
```
fdr, line 106: superfluous `puts` I presume?

prog/vnet/load_balancer_nexus.rb
```diff
@@ -43,11 +46,72 @@ def before_run
...
+      decr_refresh_cert
+      hop_wait
+    else
+      fail "Certificate finalization failed"
+    end
+  end
+
   label def create_new_health_probe
     vms = load_balancer.vms
     vms_getting_probed = strand.children_dataset.where(prog: "Vnet::LoadBalancerHealthProbes").map { |st| st.stack[0]["vm_id"] }
```
fdr, line 117: just noticing this now, I know it's not part of this patch, but it's an oof given it'll break under `push`. Also oofy is crawling around in the child's data structure, it seems like any monotonic logic here is hard to analyze and keep intact.

fdr (changes requested):
some changes and should probably refer to acme-stuff as acme-stuff.

prog/vnet/load_balancer_nexus.rb
furkansahin: I have replaced it with private subnet's ubid. This way, we won't need to generate a specific cert for every load balancer in the same subnet. They can share. I was planning to start generating wildcard certs like `*.[private_subnet_ubid_last_5_chars].lb.ubicloud.com`

prog/vnet/load_balancer_nexus.rb
furkansahin: So, instead of creating health checks here, I can create them at the time of `add_vm`. I wanted to be able to backfill health checks just in case an operator (me), messes up.

furkansahin:
More tests will be written, but essentially, the PR is functional and reviewable.

migrate/20240801_add_load_balancer_cert.rb
```diff
@@ -0,0 +1,16 @@
...
+  change do
+    create_table(:certs_load_balancers) do
+      foreign_key :load_balancer_id, :load_balancer, type: :uuid, null: false
+      foreign_key :cert_id, :cert, type: :uuid, null: false
+      primary_key [:load_balancer_id, :cert_id]
+    end
+
+    create_enum(:lb_hc_protocol, %w[tcp http https])
+    alter_table(:load_balancer) do
+      add_column :health_check_protocol, :lb_hc_protocol, null: false, default: "http"
```
bsatzger, line 13: thanks!

migrate/20240801_add_load_balancer_cert.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:certs_load_balancers) do
```
bsatzger, line 5: a cert would probably always belong to a single load balancer. why do we need this association table? couldn't we just add a load_balancer_id to the cert table?

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+      fail "Certificate finalization failed"
+    end
+  end
+
+  label def wait
+    if cert.created_at < Time.now - 60 * 60 * 24 * 30 * 3
+      hop_destroy
+    end
+
+    nap 60 * 60 * 24 * 30 * 1
```
bsatzger, line 89: nit: do we need the `* 1`?

migrate/20240801_add_load_balancer_cert.rb
furkansahin: The reason I have implemented this way is that I wanted to make the Certs a separate product. This way, I am planning that in future, we will be able to provision authorized certs independently for Postgres or MinIO products as well. If we simply add a load balancer_id to the cert table, we'll need to add a new column for every other integration and I don't like it. Does this make sense?

prog/vnet/load_balancer_nexus.rb
furkansahin: I have implemented this in #1828

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+    if cert.created_at < Time.now - 60 * 60 * 24 * 30 * 3 # 3 months
+      hop_destroy
+    end
+
+    nap 60 * 60 * 24 * 30 # 1 month
+  end
+
+  label def destroy
+    begin
+      acme_client.revoke(certificate: cert.cert) if cert.cert
```
fdr, line 94: you could include the revocation "reason" as "cessation of operation" (5). Maybe you can be kind to the reader and bind (at least that one) to constants: https://datatracker.ietf.org/doc/html/rfc5280#section-5.3.1.

You might also find use for the renewal/superseded code.

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+
+    dns_zone.delete_record(record_name: "#{dns_challenge.record_name}.#{cert.hostname}") if dns_challenge
+    cert.destroy
+    pop "certificate revoked and destroyed"
+  end
+
+  def acme_client
+    Acme::Client.new(private_key: OpenSSL::PKey::RSA.new(cert.private_key), directory: Config.acme_directory, kid: cert.kid)
+  rescue
```
fdr, line 106: Blanket rescue is weird, why is it necessary?

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  label def start
+    register_deadline(:wait, 10 * 60)
+
+    private_key = OpenSSL::PKey::RSA.new(4096)
+    client = Acme::Client.new(private_key: private_key, directory: Config.acme_directory)
+    account = client.new_account(contact: "mailto:#{Config.acme_email}", terms_of_service_agreed: true, external_account_binding: {kid: Config.acme_eab_kid, hmac_key: Config.acme_eab_hmac_key})
+    order = client.new_order(identifiers: [cert.hostname])
+    authorization = order.authorizations.first
+    cert.update(kid: account.kid, private_key: private_key.to_s, order_url: order.url)
```
fdr, line 36: I suggest using the `der` binary format. Unlike, say, attr_vault (which I think is better in at least this one way), it will get base64 encoded anyway by Sequel column encryption. No need to bloat the representation by double-stacking things. You might notice the acme-client transmits everything in der as well (at least the revocation code did)

fdr (changes requested):
Pretty good, I like the decoupling from LBs, as you wrote, there are other uses of certificates, I think this will do nicely.

Some quibbles about encoding and cipher choice, RSA is pretty darn passe by this point: much slower as it requires huge keys to be secure. P-256 is likely adequate for all,~~ but in principle, the CSR key can be something even more obscure & efficient (ed25519) since client compatibility is not at issue.

On that, I appreciate that you separated the two. Many people reuse a single key, and that can make sense for the independent operator, but it's nice that we don't have to expose the csr_key to customer code at any point, unlike the private_key that accompanies their cert.~~ Not sure what I was thinking here

Also, what's the renewal situation? I thought it may come after this patch somewhere, and maybe we wouldn't do it since we don't know when the client fetches a new cert, and we'd let things expire naturally. Wanted to ask.

model/cert.rb
```diff
@@ -0,0 +1,18 @@
...
+  include ResourceMethods
+  include SemaphoreMethods
+  semaphore :destroy
+
+  plugin :column_encryption do |enc|
+    enc.column :cert
+    enc.column :private_key
+    enc.column :csr_key
+    enc.column :kid
+  end
```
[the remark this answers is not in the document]
furkansahin: Right, I don't think cert and kid need to be kept encrypted.

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    hop_wait_dns_validation
+  end
+
+  label def wait_dns_validation
+    case dns_challenge.status
+    when "pending"
+      nap 10
+    when "valid"
+      csr_key = OpenSSL::PKey::RSA.new(4096)
```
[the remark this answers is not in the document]
furkansahin: Ok, so, I'll re-iterate to see if I understand correctly, please correct me. The certificate is the public part, csr_key is the private part. CSR key is used to decrypt the data, locally, in the VM. Considering our VMs won't be connecting to internet via browsers (?) the browser compatibility of the csr key is not an issue. Therefore, we can use Ed25519.

config.rb
```diff
@@ -163,4 +163,12 @@ def self.e2e_test?
   # Load Balancer
   optional :load_balancer_service_project_id, string
   optional :load_balancer_service_hostname, string
+
+  # Let's Encrypt
+  # The following are optional because they are only needed in production.
+  # They are not needed in development or test.
+  optional :acme_email, string
+  override :acme_directory, "https://acme.zerossl.com/v2/DV90", string
```
enescakir, line 171: The comment mentions 'Let's Encrypt', but the directory is 'ZeroSSL'. Aren't they two different providers?

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+      csr = Acme::Client::CertificateRequest.new(private_key: csr_key, common_name: cert.hostname)
+      acme_order.finalize(csr: csr)
```
enescakir, line 61: Is this code idempotent? Can we finalize it multiple times?

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+      dns_zone.delete_record(record_name: "#{dns_challenge.record_name}.#{cert.hostname}")
+      hop_wait
+    else
+      fail "Certificate finalization failed"
+    end
+  end
+
+  label def wait
+    if cert.created_at < Time.now - 60 * 60 * 24 * 30 * 3 # 3 months
+      hop_destroy
```
enescakir, line 86: We should hop to destroy directly. It bypass the `before_run`. We should do it via semaphore `incr_destroy`

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+
+    nap 60 * 60 * 24 * 30 # 1 month
+  end
+
+  label def destroy
+    begin
+      acme_client.revoke(certificate: cert.cert) if cert.cert
+    rescue Sequel::Error
```
enescakir, line 95: `acme_client` raises what kind of `Sequel::Error`?

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+      nap 10
+    when "valid"
+      csr_key = OpenSSL::PKey::RSA.new(4096)
+      csr = Acme::Client::CertificateRequest.new(private_key: csr_key, common_name: cert.hostname)
+      acme_order.finalize(csr: csr)
+      cert.update(csr_key: csr_key.to_s)
+
+      hop_wait_cert_finalization
+    else
+      fail "DNS validation failed"
```
enescakir, line 66: If we put status to it, it might be useful

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  label def destroy
+    begin
+      acme_client.revoke(certificate: cert.cert) if cert.cert
+    rescue Sequel::Error
+      fail "Failed to revoke certificate"
+    end
+
+    dns_zone.delete_record(record_name: "#{dns_challenge.record_name}.#{cert.hostname}") if dns_challenge
```
enescakir, line 99: I'm not sure when `dns_challenge` can be null, but it seems we need to verify the existence of the DNS record and delete it regardless.

prog/vnet/cert_nexus.rb
furkansahin: It's not, removed.

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  def acme_order
+    acme_client&.order(url: cert.order_url)
+  end
+
+  def dns_challenge
+    acme_order&.authorizations&.first&.dns
+  end
+
+  def dns_zone
```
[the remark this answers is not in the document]
furkansahin: I didn't want to because it's possible the zone entity changes that manage the same sub domain name over time.

prog/vnet/cert_nexus.rb
furkansahin: no it is not, that's why we don't handle the failure. If it fails after `finalize` on-call needs to clean-up.

prog/vnet/cert_nexus.rb
furkansahin: Why?

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  label def wait_cert_finalization
+    case acme_order.status
+    when "processing"
+      nap 10
+    when "valid"
+      cert.update(cert: acme_order.certificate, created_at: Time.now)
+
+      dns_zone.delete_record(record_name: "#{dns_challenge.record_name}.#{cert.hostname}")
```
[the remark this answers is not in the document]
furkansahin: Does this refer to the line `dns_challenge.record_name`? What does this mean?

> Maybe push the anti-nil checks into a helper

Should I maybe simply store the record_name to the table? I'm not sure if it always has to be but also the record_name is always `_acme-challenge.#{cert.hostname}`. Maybe I don't have to pull it from dns_challenge.

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+    hop_wait_dns_update
+  end
+
+  label def wait_dns_update
+    dns_record = DnsRecord[dns_zone_id: dns_zone.id, name: "#{dns_challenge.record_name}.#{cert.hostname}.", tombstoned: false]
+    if DB[:seen_dns_records_by_dns_servers].where(dns_record_id: dns_record.id).all.empty?
+      nap 10
+    end
+
+    dns_challenge.request_validation
```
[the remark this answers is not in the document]
furkansahin: it won't return `nil`, I will actually make some changes regarding that.

model/certs_load_balancers.rb
```diff
@@ -0,0 +1,16 @@
+#  frozen_string_literal: true
+
+require_relative "../model"
+
+class CertsLoadBalancers < Sequel::Model
```
enescakir, line 5: Why did we create a model for this joint table?

model/load_balancer.rb
```diff
@@ -53,4 +55,14 @@ def remove_vm(vm)
   def hostname
     "#{name}.#{private_subnet.ubid[-5...]}.#{Config.load_balancer_service_hostname}"
   end
+
+  def need_certificates?
+    return true if certs_dataset.empty?
+
+    certs.all? { |lbc| lbc.created_at && lbc.created_at < Time.now - 60 * 60 * 24 * 30 * 2 }
```
enescakir, line 66: `created_at` column can't be null. So we don't need check `lbc.created_at` check

```ruby
column :created_at, :timestamp, null: false, default: Sequel.function(:now)
```

model/load_balancer.rb
```diff
@@ -53,4 +55,14 @@ def remove_vm(vm)
...
+  end
+
+  def active_cert
+    certs.sort_by(&:created_at).reverse.find { |lbc| lbc.created_at && lbc.created_at > Time.now - 60 * 60 * 24 * 30 * 3 }
```
enescakir, line 70: `created_at` column can't be null. So we don't need check `lbc.created_at` check

```ruby
column :created_at, :timestamp, null: false, default: Sequel.function(:now)
```

prog/vnet/load_balancer_nexus.rb
```diff
@@ -45,9 +48,27 @@ def before_run
...
+    load_balancer.add_cert(cert)
+    hop_wait_cert_provisioning
+  end
+
+  label def wait_cert_provisioning
+    if load_balancer.need_certificates?
+      nap 60
+    else
+      hop_wait
+    end
```
enescakir, line 69: nit: no need to else branch
```ruby
    nap 60  if load_balancer.need_certificates?
    
    hop_wait
```

prog/vnet/load_balancer_nexus.rb
```diff
@@ -45,9 +48,27 @@ def before_run
 
     hop_create_new_health_probe if strand.children_dataset.count < load_balancer.vms_dataset.count
 
+    if load_balancer.need_certificates?
+      hop_create_new_cert
+    end
```
enescakir, line 53: nit: it will have the same syntax with the previous line `hop_create_new_health_probe if strand.children_dataset.count < load_balancer.vms_dataset.count`

```ruby
hop_create_new_cert if load_balancer.need_certificates?
```

prog/vnet/cert_nexus.rb
enescakir: typo. `We should **not** hop to destroy directly`

prog/vnet/load_balancer_health_probes.rb
```diff
@@ -10,8 +10,13 @@ def vm
   label def health_probe
     response_code = nil
     begin
-      endpoint = "#{vm.nics.first.private_ipv4.network}:#{load_balancer.dst_port}#{load_balancer.health_check_endpoint}"
-      response_code = vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} curl --max-time #{load_balancer.health_check_timeout} --silent --output /dev/null --write-out '%{http_code}' #{endpoint}")
+      cmd = if load_balancer.health_check_protocol == "tcp"
+        "sudo ip netns exec #{vm.inhost_name} nc -z -w #{load_balancer.health_check_timeout} #{vm.nics.first.private_ipv4.network} #{load_balancer.dst_port} && echo 200 || echo 400"
+      else
+        "sudo ip netns exec #{vm.inhost_name} curl --insecure --max-time #{load_balancer.health_check_timeout} --silent --output /dev/null --write-out '%{http_code}' #{load_balancer.health_check_protocol}://#{vm.nics.first.private_ipv4.network}:#{load_balancer.dst_port}#{load_balancer.health_check_endpoint}"
```
enescakir, line 16: Should we use `--insecure` for `https` protocol?

prog/vnet/cert_nexus.rb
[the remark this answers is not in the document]
fdr: Okay I'm not sure what I was thinking, I guess my question should be: what is private_key for, versus csr_key?

prog/vnet/cert_nexus.rb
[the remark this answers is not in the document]
fdr: It's more about nil-to-empty expansion:

`"#{dns_challenge.record_name}.#{cert.hostname}"` can yield `.` if both fields are nil.

Compare to `dns_challenge.record_name + "." + cert.hostname`, which crashes if any of fields are nil. I used to occasionally confuse my reviewers as to why I would use this more verbose approach of what is a bit more natural string interpolation, but they tended to accept my reason when I explained it :)

config.rb
fdr: really it's anything ACME, but only total nerds know about ACME, which was developed by ISRG for the Let's Encrypt project.

prog/vnet/cert_nexus.rb
fdr: ooof can we avoid that? Like, what is the cleanup that we'd do by hand that cannot be automated?

model/load_balancer.rb
fdr: also, it would crash, right?

Though, this raises a question...why is this code toggling between using the dataset, and then using the orm array? It seems like this could all be one SQL query. Also any time I see this kind of mixed code, I have to think carefully about how it interacts with an already-cached `certs` array that may have read a different snapshot. That can be okay when there's a nice efficiency or clarity advantage to mixing and matching, but I think this is an anti-example.


You can probably do some variant of `certs_dataset.where { created_at < Time.now - 60 * 60 * 24 * 30 * 2 }.empty?` unless you are carefully loading a snapshot of the data into `certs` for some reason.

model/load_balancer.rb
fdr: This also can be done as a single SQL expression: `certs_dataset.where { created_at > Time.now - 60 * 60 * 24 * 30 * 3 }.order(Sequel.desc(:created_at)).first`

But, you may not want to do that, depending how the method is used and whether you want it to be consistent with an already-loaded `certs` array.

prog/vnet/load_balancer_health_probes.rb
fdr: I think for a health check it's okay, to allow for self-signed certificates or expired certificates. A "nice" feature would be to tell the customer about expiry :)

I am not a huge fan of using curl altogether though. Memory-unsafe C program that has fork overhead when we don't really need either demerit. But I think @furkansahin was thinking he could complicate things and make them more efficient via SSH tunnels in the monitor process later.

prog/vnet/load_balancer_health_probes.rb
```diff
@@ -10,8 +10,13 @@ def vm
...
+      end
+
+      response_code = vm.vm_host.sshable.cmd(cmd)
     rescue
```
fdr, line 20: blanket rescue, though from another patch. Consider the following:

```
[8] clover-development(main)> begin;
[8] clover-development(main)*   notbound
[8] clover-development(main)* rescue;  
[8] clover-development(main)* end;  
=> nil
```

prog/vnet/cert_nexus.rb
fdr: I think the more pressing concern is if record_name is ever nilable, or hostname ever becomes nilable, and how we supervise whether that happens and what the downside risk is if t does happen (whereas, some other programming styles, like using `+`, would crash)

fdr:
Question, I feel like there was a thread on this and I can't find it anymore, did I accidentally delete it?

So what's the deal with both `private_key` and `csr_key`? Is `csr_key` what's intended for the VM (eventually) and `private_key` something held in confidence to authenticate with ACME? I'm a bit puzzled at the two levels of asymmetric cryptography going on here.

model/certs_load_balancers.rb
furkansahin: to override the `destroy` method

prog/vnet/cert_nexus.rb
furkansahin: The `dns_challenge` can be `nil` if the cert starts to be destroyed before the `dns_challenge` is requested. But if the `dns_challenge` is not `nil`, we must have the `record_name`, I'll make the necessary changes.

prog/vnet/load_balancer_health_probes.rb
furkansahin: `--insecure` is OK because we're already inside network namespace. Even if the customer did not get the certificate, we want the load balancer to perform health checks and decide that this is healthy. This is the only way they can realise the connection errors are about the certificate but not the end to end connection they have from client to the server.

config.rb
furkansahin: removed the `Let's Encrypt` all together, variable names are clear enough.

model/certs_load_balancers.rb
enescakir: IMHO, it's too much code to just override `destroy`. I feel like we can make it cleaner without `CertsLoadBalancers` pivot model. Let's get call for it today

prog/vnet/cert_nexus.rb
```diff
@@ -0,0 +1,121 @@
...
+    register_deadline(:wait, 10 * 60)
+
+    private_key = OpenSSL::PKey::RSA.new(4096)
+    client = Acme::Client.new(private_key: private_key, directory: Config.acme_directory)
+    account = client.new_account(contact: "mailto:#{Config.acme_email}", terms_of_service_agreed: true, external_account_binding: {kid: Config.acme_eab_kid, hmac_key: Config.acme_eab_hmac_key})
+    order = client.new_order(identifiers: [cert.hostname])
+    authorization = order.authorizations.first
+    cert.update(kid: account.kid, private_key: private_key.to_s, order_url: order.url)
+    dns_challenge = authorization.dns
+    dns_zone.insert_record(record_name: "#{dns_challenge.record_name}.#{cert.hostname}", type: dns_challenge.record_type, ttl: 600, data: dns_challenge.record_content)
```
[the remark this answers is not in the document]
furkansahin: I have updated this.

prog/vnet/cert_nexus.rb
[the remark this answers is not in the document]
furkansahin: private_key is basically used to create a connection to the acme server for that cert specifically. And then, when we order a certificate, we send the csr_key in the payload which is then used to provision the cert. The customer gets the certificate and the csr_key to use.

prog/vnet/cert_nexus.rb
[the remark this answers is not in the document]
furkansahin: Updated this

prog/vnet/cert_nexus.rb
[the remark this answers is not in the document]
furkansahin: I have updated this.

prog/vnet/cert_nexus.rb
furkansahin: I have added paths to get back to start to retry the cert provisioning. If the deadline is reached because of an error that we did not handle, at that point the on-call will need to look into the logs and find a solution.

furkansahin:
> Question, I feel like there was a thread on this and I can't find it anymore, did I accidentally delete it?
So what's the deal with both private_key and csr_key? Is csr_key what's intended for the VM (eventually) and private_key something held in confidence to authenticate with ACME? I'm a bit puzzled at the two levels of asymmetric cryptography going on here. Some of my other comments suggest that at least one of these theoretically could use a state of the art cipher as it's not exposed to a web browser, i.e. compatibility considerations are lesser. But this gain might be pretty marginal.

I have responded this here https://github.com/ubicloud/ubicloud/pull/1797#discussion_r1711296158

prog/vnet/cert_nexus.rb
furkansahin: I have updated this

model/load_balancer.rb
furkansahin: Updated

model/load_balancer.rb
furkansahin: updated

prog/vnet/cert_nexus.rb
fdr: I think that's all we can ask for really, that at least we'd see something interesting if it doesn't quite work.

prog/vnet/cert_nexus.rb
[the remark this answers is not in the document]
fdr: > private_key is basically used to create a connection to the acme server for that cert specifically. And then, when we order a certificate, we send the csr_key in the payload which is then used to provision the cert. The customer gets the certificate and the csr_key to use.

let's call it "account key pair" (or account_key for short). This is what the RFC calls it. https://datatracker.ietf.org/doc/html/rfc8555/#section-3

Normally private key suffices, but now there are two of them.

enescakir (approved):

furkansahin:
@fdr review is appreciated

routes/common/load_balancer_helper.rb
```diff
@@ -119,6 +120,31 @@ def delete
     @request.halt
   end
 
+  def download_certificate_bundle
```
fdr, line 125: hmmm do we actually want to ship this feature, to get frequently-rotating private keys out via the web page? I understand why that may have been useful in development, but who is going to be clicking around every thirty or ninety days to do this?

prog/vnet/update_load_balancer.rb
```diff
@@ -18,7 +18,7 @@ def load_balancer
   end
 
   label def remove_load_balancer
-    vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} nft --file -", stdin: generate_nat_rules(vm.ephemeral_net4.to_s, vm.nics.first.private_ipv4.network.to_s))
+    vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} nft --file -", stdin: generate_nat_rules(vm.ephemeral_net4.to_s, vm.nics.first.private_ipv4.network.to_s)) if vm
```
fdr, line 21: I'd  rewrite this with:

if vm
 ...
end

It's easy to miss that "if" at the end of that huuuuge line

spec/model/load_balancer_spec.rb
```diff
@@ -64,4 +64,60 @@
       expect(lb.load_balancers_vms.count).to eq(0)
     end
   end
+
+  describe "need_certificates?" do
```
fdr, line 68: Commit message is: " ACME Cert integration tests ".... I think this is ambiguous because "integration test" has a conflicting meaning. I'm sure you can phrase-smith it to not make someone wonder if this contacts a live acme service, as "e2e" (which could be called an "integration test") does.

fdr (changes requested):
Main thing is I think we should keep the certificate download UI in our back pocket, because it seems of such marginal use given the rotation interval.

routes/common/load_balancer_helper.rb
furkansahin: You're right. At the time, I didn't think I would be this quick to implement the certificate server and I wanted to give a way to customer to get the certificates. However, considering we have the pr #1858 out, I'm removing this.

prog/vnet/update_load_balancer.rb
furkansahin: I removed the change all together, because I think #1828 is a better place to put this in.

spec/model/load_balancer_spec.rb
furkansahin: Updated

furkansahin:
@fdr , I updated the PR as you requested.

fdr (approved):

PR #1799, opened 2024-07-24
Combine create_unix_user commands to single command

enescakir:
### Combine create_unix_user commands to single command

We cache SSH connections to our host to run commands more quickly. Even
for short commands, the network adds ~0.4s of overhead. Previously, we
achieved a few seconds of gain by merging multiple commands into a
single one in GithubRunner, which makes a significant difference in
time-sensitive operations.

We can also consolidate the `create_unix_user` commands into a single
command. Additionally, I have relocated `sudo usermod -a -G kvm #{q_vm}`
from the `prep` label to this command, as it is more logical to execute
it here. I will remove it from the `prep` label in the next commit. If
there are any strands that have already passed `create_unix_user` but
haven't run `prep` yet. Since this operation is idempotent, it's safe to
run it multiple times.

### Remove duplicate command from prep label

I relocated it from `prep` to `create_unix_user` in the previous commit.
I removed it in this commit for a smoother deployment.

### Reduce sleep time in vm provisioning from 12 to 10 seconds

In the `wait_sshable` label, we initially set a long sleep time because
we expected the first run to take some time. The sleep time was
initially set at 12 seconds. However, we noticed that some duration
times were approaching this 12-second mark. As a result, we decided to
decrease the sleep time to 10 seconds.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -211,12 +211,19 @@ class Prog::Vm::Nexus < Prog::Base
   end
 
   label def create_unix_user
-    host.sshable.cmd("sudo userdel --remove --force #{q_vm} || true")
-    host.sshable.cmd("sudo groupdel -f #{q_vm} || true")
-
-    # create vm's user and home directory
     uid = rand(1100..59999)
-    host.sshable.cmd("sudo adduser --disabled-password --gecos '' --home #{vm_home.shellescape} --uid #{uid} #{q_vm}")
+    command = <<~COMMAND
+      set -ueo pipefail
+      # Make this script idempotent
+      sudo userdel --remove --force #{q_vm} || true
+      sudo groupdel -f #{q_vm} || true
+      # Create vm's user and home directory
+      sudo adduser --disabled-password --gecos '' --home #{vm_home.shellescape} --uid #{uid} #{q_vm}
+      # Enable KVM access for VM user
+      sudo usermod -a -G kvm #{q_vm}
+    COMMAND
+
+    host.sshable.cmd(command)
 
     hop_prep
   end
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -239,9 +239,6 @@ class Prog::Vm::Nexus < Prog::Base
         storage: vm.storage_secrets
       })
 
-      # Enable KVM access for VM user.
-      host.sshable.cmd("sudo usermod -a -G kvm #{q_vm}")
-
       write_params_json
 
       host.sshable.cmd("common/bin/daemonizer 'sudo host/bin/setup-vm prep #{q_vm}' prep_#{q_vm}", stdin: secrets_json)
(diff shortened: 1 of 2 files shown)


spec/prog/vm/nexus_spec.rb
```diff
@@ -476,7 +480,7 @@ def requested_disk_size(st)
     it "naps 12 second if it's the first time we execute wait_sshable" do
```
bsatzger, line 480: we should probably also change the test description to "naps 10 ..."

model/sshable.rb
```diff
@@ -98,6 +98,11 @@ def cmd(cmd, stdin: nil, log: true)
     stdout_str
   end
 
+  def cmd_multiline(cmd, stdin: nil, log: true)
```
byucesoy, line 101: Maybe use different name for the variable, so that it wouldn't get mixed up with the function. Maybe `command`?

enescakir: I've kept the same signature with `def cmd(cmd, stdin: nil, log: true)`. However, if you feel strongly about it, we can consider changing both

spec/prog/vm/nexus_spec.rb
enescakir: I didn't notice that. Thanks.

bsatzger (approved):

model/sshable.rb
enescakir: I removed it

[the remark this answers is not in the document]
enescakir: Fair enough, I've removed that. We might consider adding an extra parameter to the `cmd`, something like `strip_comments`, if we find it necessary in the future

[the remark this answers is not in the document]
fdr: Yeah, I think a parameter makes more sense: you can imagine someone switching from "cmd" to "multi line" when they change the program, but not think about how the latter messes with the bytes they put in...unless, we always mess with the bytes, in which case, there is no point in having two behaviors at all.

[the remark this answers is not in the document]
enescakir: @fdr Do you want to approve or reject this PR?

prog/vm/nexus.rb
```diff
@@ -239,9 +239,6 @@ def before_run
         storage: vm.storage_secrets
       })
 
-      # Enable KVM access for VM user.
-      host.sshable.cmd("sudo usermod -a -G kvm #{q_vm}")
-
```
fdr, line 237: reading this commit message, re: "I relocated it from `prep` to `create_unix_user` in the previous commit.
I removed it in this commit for a smoother deployment." 

Is the intermediate state (where we add the user to the KVM gruop twice) acceptable?

fdr (approved):
other than being careful about testing intermediate states if doing phased deployment, lgtm

prog/vm/nexus.rb
enescakir: We are moving adding the user to `kvm` group command to an earlier step. The second command is to ensure that all users are included in the `kvm` group if any strand has already passed the `create_unix_user` label but hasn't run `prep` label yet at the time of deployment.

enescakir: I run E2E pipeline on this middle step
https://github.com/ubicloud/ubicloud/actions/runs/10116708456

PR #1800, opened 2024-07-24
Do not invite deleted accounts to the project

enescakir:
Typically, user accounts are identified uniquely by their email. However, this only applies to active accounts. When an account is deleted, Rodauth changes its status to '3:deleted' and allows the creation of a new account with the same email. Therefore, when we select a user solely by email, we might retrieve a deleted account. To prevent this, we need to exclude deleted accounts. Some of our customers have already encountered this issue.

### DIFF
--- a/routes/web/project/user.rb
+++ b/routes/web/project/user.rb
@@ -12,7 +12,8 @@ class CloverWeb
 
     r.post true do
       email = r.params["email"]
-      user = Account[email: email]
+      # Don't invite deleted accounts
+      user = Account.exclude(status_id: 3)[email: email]
 
       user&.associate_with_project(@project)
 


byucesoy:
One thing I realized, when account become inactive, we don't delete access tags and applied tags. Then, it wouldn't be possible to associate new account with the project again. It errors out with the following message

```
Sequel::ValidationFailed: project_id and hyper_tag_id is already taken, project_id and name is already taken
```

Should we handle this case in this PR as well?

routes/web/project/user.rb
```diff
@@ -12,7 +12,8 @@ class CloverWeb
 
     r.post true do
       email = r.params["email"]
-      user = Account[email: email]
+      # Don't invite deleted accounts
+      user = Account.exclude(status_id: 3).first(email: email)
```
fdr, line 16: `Account.exclude(status_id: 3)[email: email]` ?

fdr (approved):
Probably need to sugar/desugar those policies into ubids, and/or rethink (or, more like, more-think) namespacing of accounts -- which are a bit special and "extraterrestrial" to the model.

enescakir:
> One thing I realized, when account become inactive, we don't delete access tags and applied tags. Then, we cannot associate new account with the project again (if the old account is already associated). It errors out with the following message
> 
> ```
> Sequel::ValidationFailed: project_id and hyper_tag_id is already taken, project_id and name is already taken
> ```
> 
> Should we handle this case in this PR as well?

I created a PR for it https://github.com/ubicloud/ubicloud/pull/1803

routes/web/project/user.rb
enescakir: I didn't realize that we could use `[]` with a dataset. I thought it could only be used directly with the class. Thanks for the new insight; I made the change.

byucesoy (approved):

PR #1805, opened 2024-07-25
Set Content-Type of not_found plugin for API

fdr:
Previously, this was returning JSON content with a text/html Content-Type.

I noticed this when trying to get our OpenAPI schema into shape, as to validate, the content type must match specification, and in the API, that is all `application/json` for now.

### DIFF
--- a/clover_api.rb
+++ b/clover_api.rb
@@ -24,6 +24,7 @@ class CloverApi < Roda
   end
 
   plugin :error_handler do |e|
+    response["Content-Type"] = "application/json"
     error = parse_error(e)
 
     {error: error}.to_json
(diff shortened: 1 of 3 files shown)


bsatzger (approved):

enescakir (approved):
Good finding! It's interesting that `plugin :default_headers, "Content-Type" => "application/json"
` plugin doesn't set default headers for `not_found` and `error_handler` handlers. It might be a bug or missing feature. I will looks its implementation details when I have time

velioglu (approved):

fdr:
> Good finding! It's interesting that `plugin :default_headers, "Content-Type" => "application/json" ` plugin doesn't set default headers for `not_found` and `error_handler` handlers. It might be a bug or missing feature. I will looks its implementation details when I have time

Interesting. Might be a report for The King.

PR #1812, opened 2024-07-27
Move the trigger for starting vm's systemd unit to rhizome

enescakir:
### Move the trigger for starting vm's systemd unit to rhizome

After the `prep` label is complete, the control plane run the systemd
unit in the `run` label. The `prep` script is daemonized, which causes
the control plane to poll too frequently to check its status, then it
starts the systemd unit. When we move this run command to the end of the
prep script at rhizome, the vm starts to boot immediately after it is
prepared. Thus, it will start booting even before the control plane
acknowledges that the prep is complete.

Our main goal is to minimize the interaction between the control plane
and the data plane to enhance our provisioning times. By removing the
`run` label and initiating the booting process earlier, I anticipate a
1-2 second improvement in provisioning time.

`VmSetup.recreate_unpersisted` is invoked in
`Vm::Nexus.start_after_host_reboot` and shares similar logic with
`VmSetup.prep`. Although it's not time-sensitive, I've changed it for
consistency.

I remove the `run` label in the next commit for smoother deployment.

### Skip the run label in nexus since it's moved to rhizome
We moved the run command of the vm's systemd unit to rhizome in the
previous commit. After installing rhizome to all hosts, we can start to
skip the `run` label.

### Remove the run label in nexus
It's skipped in the previous commit. So there is no nexus strand on the
`run` label. We can remove it now.

We also also moved start command in `start_after_host_reboot` to
rhizome.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -251,11 +251,6 @@ class Prog::Vm::Nexus < Prog::Base
     host.sshable.cmd("sudo -u #{q_vm} tee #{params_path.shellescape}", stdin: vm.params_json(frame["swap_size_bytes"]))
   end
 
-  label def run
-    host.sshable.cmd("sudo systemctl start #{q_vm}")
-    hop_wait_sshable
-  end
-
   label def wait_sshable
     unless vm.update_firewall_rules_set?
       vm.incr_update_firewall_rules
@@ -470,7 +465,6 @@ class Prog::Vm::Nexus < Prog::Base
     })
 
     host.sshable.cmd("sudo host/bin/setup-vm recreate-unpersisted #{q_vm}", stdin: secrets_json)
-    host.sshable.cmd("sudo systemctl start #{q_vm}")
     vm.nics.each { _1.incr_repopulate }
 
     vm.update(display_state: "running")
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -233,7 +233,7 @@ class Prog::Vm::Nexus < Prog::Base
     when "Succeeded"
       host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")
       vm.nics.each { _1.incr_setup_nic }
-      hop_run
+      hop_wait_sshable
     when "NotStarted", "Failed"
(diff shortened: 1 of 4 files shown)


byucesoy (approved):

fdr (approved):
Probably should, in one of the commits, likely the first one, mention the motive. I presume it makes things faster. Bonus points for some idea of how much faster.

I do appreciate that the deployment/patch division logic was very clear. That I think I understood completely on the first reading.

Although I don't anticipate a problem given my general (i.e., not at all to be trusted) understanding of the code, the main area I could see this having a conflict is with host reboots. I don't think it's likely since I see in `start_after_host_reboot` that there's a systemctl start in there as well, i.e. a duplicate.

enescakir:
> Probably should, in one of the commits, likely the first one, mention the motive. I presume it makes things faster. Bonus points for some idea of how much faster.

I added the following details to the first commit message:
```
Our main goal is to minimize the interaction between the control plane
and the data plane to enhance our provisioning times. By removing the
run label and initiating the booting process earlier, I anticipate a
1-2 second improvement in provisioning time.
```

> I do appreciate that the deployment/patch division logic was very clear. That I think I understood completely on the first reading.

I run E2E for each commits separately.
_Commit 1:_ https://github.com/ubicloud/ubicloud/actions/runs/10163851951
_Commit 2:_ https://github.com/ubicloud/ubicloud/actions/runs/10164492626
_Commit 3:_ https://github.com/ubicloud/ubicloud/actions/runs/10162391121


> Although I don't anticipate a problem given my general (i.e., not at all to be trusted) understanding of the code, the main area I could see this having a conflict is with host reboots. I don't think it's likely since I see in `start_after_host_reboot` that there's a systemctl start in there as well, i.e. a duplicate.

`start_after_host_reboot` calls `host/bin/setup-vm recreate-unpersisted`. I had changed `host/bin/setup-vm prep`, so it doesn't have a duplicate start. But `VmSetup.recreate_unpersisted` and `VmSetup.prep` shares similar logics. Furkan also experienced similar confusion when he reviewed the PR. Although `start_after_host_reboot` is not time-sensitive, I've changed it for consistency.

PR #1813, opened 2024-07-27
Remove unnecessary explicit schedule values

enescakir:
The `schedule` column already has a default value of `Sequel.lit("now()")`. Therefore, there's no need to explicitly set it to `Time.now` when creating a new strand.


byucesoy (approved):

fdr (approved):
Nice shortening in number of expressions

PR #1814, opened 2024-07-29
Run cache reserve operations in a transaction

enescakir:
Our progs run within a transaction by default, but web endpoints do not.
Therefore, we should wrap endpoint operations, such as cache reserve
operations, that could result in an inconsistent state, within a
transaction. If the `create_multipart_upload` operation fails, we're
left with a cache entry lacking an `upload_id`. Without an `upload_id`,
we're unable to abort the upload. We've encountered this issue in
production. The cache destroy fails with the following error:

    ArgumentError: missing required parameter params[:upload_id]

Furthermore, cache key and version are unique to a repository. Hence,
this inconsistency prevents customers from creating the same cache
again.

### DIFF
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -82,10 +82,12 @@ class CloverRuntime
           fail CloverError.new(400, "InvalidRequest", "The cache size is over the 10GB limit")
         end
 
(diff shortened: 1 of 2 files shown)


byucesoy (approved):

fdr (approved):
thanks for fixing, since it can page :)

PR #1817, opened 2024-07-30
Run networking and storage setup in parallel during vm prep

enescakir:
### Decouple cloudinit and networking setup in vm prepare

I'm planning to run a few vm preparation steps in parallel. To achieve
this, we need to decouple these steps as much as possible. The
`cloudinit` uses `guest_ephemeral` from the `setup_networking` step. In
reality, this is a subdivide of the gua and we've calculated it in
various places. Therefore, we can also calculate it here.

### Run networking and storage setup in parallel during vm prep

Setting up networking and storage are the most time-consuming steps in
our VM preparation script, taking up the majority of the process.
Ideally, each step takes 1 second, but when we have too many active vm
provisioning, each can take up to 5-8 seconds. However, as these steps
are independent, we can execute them in parallel to save time. This
approach could save us 1 second in the best-case scenario and 5-8
seconds in the worst-case scenario.

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -44,9 +44,14 @@ class VmSetup
   end
 
   def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices)
-    setup_networking(false, gua, ip4, local_ip4, nics, ndp_needed, multiqueue: max_vcpus > 1)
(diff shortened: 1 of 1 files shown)


fdr (approved):
I like the rough quantitative in the commit: {1, 5-8} seconds. Also the patch saying: "hey! at least I didn't make the duplication of a small logic *that* much worse."

We're probably short an abstraction for that...but it's not hurting us particularly much.

enescakir:
> We're probably short an abstraction for that...but it's not hurting us particularly much.

I considered a shorter abstraction, but I avoided complicating things by simply parallelizing these two steps. The basic `Thread.new` seems fine to me. If we identify more steps that can be parallelized, we can further explore parallelization within the rhizome

PR #1828, opened 2024-08-05
Load Balancer bug fixes and operational improvements

furkansahin:
## Rename UpdateLoadBalancer to UpdateLoadBalancerNode
This was a review in the initial PR, since the name is a little bit
confusing, we decided to make it more clear

## Separate VM Destroy logic from VM Detach for LBs
We used to manage VMs that are being destroyed or detached through the
same state "evacuating". That is causing issues since there are
different steps to take for both.
1. For VMs that are being destroyed, keep the load balancer definition
in place but remove self from the pool until DNS TTL is expired, and
then remove the VM <-> Load Balancer relationship and destroy the VM
entity.

2. For VMs that are being detached, we need to update DNS immediately,
remove the load balancer definition, update all of the nodes immediately
and remove the VM <-> load balancer relationship.

## Move Health Probe Lifecycle Management to LoadBalancer model
We used to create new health probes from nexus but destroy them through
model. This is not nice since they are two independent files. This
commit moves health probe creation to the model as well.

## Fix Load Balancer for IPv6
The Load Balancer was not working when there were only IPv6 only VMs in
the load balancing pool. The reason is, when there is only IPv6 only
VMs, the DNS resolves to an IPv6 address and the load balancer was
unable to route the traffic properly. Main reason of that bug is the
firewall rules implementation. When we started implementing firewalls,
we decided that the traffic in private network would be excempt from the
firewall rules. However, we made a concious decision to block packets
that blong to a NEW connection if they have the destination address of
the private ipv6. Because, it would be weird to have such a packet when
we are communicating in between different VMs. The whole private
networking logic depends on public IPs in pre forward hook + the IPSec
Tunnels.

Now, fast forward to the load balancer implementation, we DNAT the
packets destined to the public IPv6 of the VM and route them through the
Private IPv6 address. Therefore, when we evaluate the firewalls, just
before completing SNAT + IPSec Tunnelling, the firewalls see that there
is a NEW connection that is destined to the private IPv6 address of the
VM and drops the packet. With this commit, we change that logic to
accept such packets.

### DIFF
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -34,7 +34,7 @@ class LoadBalancer < Sequel::Model
   def add_vm(vm)
     DB.transaction do
       super
-      incr_update_load_balancer
+      Strand.create_with_id(prog: "Vnet::LoadBalancerHealthProbes", label: "health_probe", stack: [{subject_id: id, vm_id: vm.id}], parent_id: strand.id)
       incr_rewrite_dns_records
     end
   end
@@ -48,7 +48,7 @@ class LoadBalancer < Sequel::Model
   def evacuate_vm(vm)
     DB.transaction do
       load_balancers_vms_dataset.where(vm_id: vm.id, state: ["up", "down"]).update(state: "evacuating")
-      strand.children_dataset.where(prog: "Vnet::LoadBalancerHealthProbes").all.select { |st| st.stack[0]["subject_id"] == id && st.stack[0]["vm_id"] == vm.id }.map(&:destroy)
+      remove_health_probe(vm.id)
       incr_update_load_balancer
       incr_rewrite_dns_records
     end
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -5,7 +5,7 @@ require_relative "../model"
 class LoadBalancer < Sequel::Model
   many_to_many :vms
   many_to_many :active_vms, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: {state: ["up"]}
-  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: "evacuating")
+  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: ["evacuating", "detaching"])
   one_to_one :strand, key: :id
   many_to_one :private_subnet
   one_to_many :projects, through: :private_subnet
@@ -39,9 +39,15 @@ class LoadBalancer < Sequel::Model
     end
   end
 
+  def detach_vm(vm)
+    load_balancers_vms_dataset.where(vm_id: vm.id, state: ["up", "down", "evacuating"]).update(state: "detaching")
+    remove_health_probe(vm.id)
+    incr_update_load_balancer
+  end
+
   def evacuate_vm(vm)
     DB.transaction do
-      load_balancers_vms_dataset.where(vm_id: vm.id).update(state: "evacuating")
+      load_balancers_vms_dataset.where(vm_id: vm.id, state: ["up", "down"]).update(state: "evacuating")
       strand.children_dataset.where(prog: "Vnet::LoadBalancerHealthProbes").all.select { |st| st.stack[0]["subject_id"] == id && st.stack[0]["vm_id"] == vm.id }.map(&:destroy)
       incr_update_load_balancer
       incr_rewrite_dns_records
@@ -49,7 +55,12 @@ class LoadBalancer < Sequel::Model
   end
 
   def remove_vm(vm)
-    load_balancers_vms_dataset.where(vm_id: vm.id).all.map(&:destroy)
+    load_balancers_vms_dataset[vm_id: vm.id].destroy
+    incr_rewrite_dns_records
+  end
+
+  def remove_health_probe(vm_id)
+    strand.children_dataset.where(prog: "Vnet::LoadBalancerHealthProbes").all.select { |st| st.stack[0]["subject_id"] == id && st.stack[0]["vm_id"] == vm_id }.map(&:destroy)
   end
 
   def hostname
--- a/prog/vnet/cert_nexus.rb
+++ b/prog/vnet/cert_nexus.rb
@@ -7,7 +7,7 @@ class Prog::Vnet::CertNexus < Prog::Base
   subject_is :cert
   semaphore :destroy
 
-  REVOKE_REASON = "cessationOfOperation" # unspecified
+  REVOKE_REASON = "cessationOfOperation"
 
   def self.assemble(hostname, dns_zone_id)
     unless DnsZone[dns_zone_id]
@@ -98,7 +98,11 @@ class Prog::Vnet::CertNexus < Prog::Base
 
   label def destroy
     # the reason is chosen as "cessationOfOperation"
-    acme_client.revoke(certificate: cert.cert, reason: REVOKE_REASON) if cert.cert
+    begin
+      acme_client.revoke(certificate: cert.cert, reason: REVOKE_REASON) if cert.cert
+    rescue Acme::Client::Error::AlreadyRevoked => ex
+      Clog.emit("Certificate is already revoked") { {cert_revoke_failure: {ubid: cert.ubid, exception: Util.exception_to_hash(ex)}} }
+    end
 
     dns_zone.delete_record(record_name: dns_record_name) if dns_challenge
     cert.destroy
--- a/prog/vnet/update_firewall_rules.rb
+++ b/prog/vnet/update_firewall_rules.rb
@@ -88,9 +88,9 @@ table inet fw_table {
     ip saddr @private_ipv4_cidrs ct state established,related,new counter accept
     ip daddr @private_ipv4_cidrs ct state established,related counter accept
     ip6 saddr @private_ipv6_cidrs ct state established,related,new counter accept
-    ip6 daddr @private_ipv6_cidrs ct state established,related counter accept
+    ip6 daddr @private_ipv6_cidrs ct state established,related,new counter accept
     ip6 saddr #{guest_ephemeral} ct state established,related,new counter accept
-    ip6 daddr #{guest_ephemeral} ct state established,related counter accept
+    ip6 daddr #{guest_ephemeral} ct state established,related,new counter accept
     ip saddr @allowed_ipv4_cidrs ip daddr @private_ipv4_cidrs counter accept
     ip6 saddr @allowed_ipv6_cidrs ip6 daddr #{guest_ephemeral} counter accept
     ip saddr . tcp dport @allowed_ipv4_port_tuple ip daddr @private_ipv4_cidrs counter accept
--- a/prog/vnet/update_load_balancer_node.rb
+++ b/prog/vnet/update_load_balancer_node.rb
@@ -7,13 +7,25 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
     @load_balancer ||= LoadBalancer[frame.fetch("load_balancer_id")]
   end
 
-  label def update_load_balancer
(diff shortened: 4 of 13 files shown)


enescakir:
Screenshot 2024-08-08 at 12 59 09
I think it's a good commit title for our git history

enescakir:
I prefer to update tests on the related commits when the change done. Otherwise, it's really hard to review tests. I don't look big test commits generally.

model/load_balancer.rb
```diff
@@ -43,6 +43,12 @@ def add_vm(vm)
     end
   end
 
+  def detach_vm(vm)
+    load_balancers_vms_dataset.where(vm_id: vm.id).update(state: "detaching")
```
enescakir, line 47: nit: it may have same syntax with `remove_vm`
`load_balancers_vms_dataset[vm_id: vm.id].update(state: "detaching")`

prog/vnet/update_load_balancer.rb
```diff
@@ -7,13 +7,25 @@ def load_balancer
     @load_balancer ||= LoadBalancer[frame.fetch("load_balancer_id")]
   end
 
+  def vm_load_balancer_state
+    load_balancer.load_balancers_vms_dataset.where(vm_id: vm.id).first.state
```
enescakir, line 11: nit:
`load_balancer.load_balancers_vms_dataset[vm_id: vm.id].state`

model/load_balancer.rb
```diff
@@ -38,7 +38,7 @@ def certificate_bundle_path
   def add_vm(vm)
     DB.transaction do
       super
-      incr_update_load_balancer
+      Strand.create_with_id(schedule: Time.now, prog: "Vnet::LoadBalancerHealthProbes", label: "health_probe", stack: [{subject_id: id, vm_id: vm.id}])
```
enescakir, line 41: No need to give `schedule: Time.now` parameter. The default is already `Time.now` in DB

enescakir, line 41: In the removed code, we bud it. So it has `parent_id`. Should we set it here too?
```ruby
      bud Prog::Vnet::LoadBalancerHealthProbes, {"vm_id" => vm.id, "subject_id" => load_balancer.id}, :health_probe
```

prog/vnet/update_load_balancer_node.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class Prog::Vnet::UpdateLoadBalancer < Prog::Base
```
enescakir, line 3: Just need to be sure we don't have any existing strand for this prog before deployment

fdr (commented):
I imagine this has something in common with the other main ACME patch, which is currently getting a lot of edits, so the history is bifurcated, should I wait to review this until the other one is committed?

furkansahin:
> I imagine this has something in common with the other main ACME patch, which is currently getting a lot of edits, so the history is bifurcated, should I wait to review this until the other one is committed?

Yes

prog/vnet/update_load_balancer.rb
```diff
@@ -18,7 +18,10 @@ def load_balancer
   end
 
   label def remove_load_balancer
-    vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} nft --file -", stdin: generate_nat_rules(vm.ephemeral_net4.to_s, vm.nics.first.private_ipv4.network.to_s))
+    if vm
```
fdr, line 33: Nit: 
>  Load Balancer archival is fixed in case the VM is already archived 

Maybe like:

> Fix crash when VM is concurrently deleted

This is only to make it shorter and a bit clearer.

More generally: is it right for update load balancer to be orthogonal/not synchronized with the VM flow of control?

Granted, it's better to commit this given the status quo than not, but perhaps this prog ought to be triggered via VM semaphore unless it's necessary to the VM doing, possibly, anything while doing this.

model/load_balancer.rb
```diff
@@ -34,21 +34,21 @@ def path
...
       super
-      incr_update_load_balancer
+      Strand.create_with_id(prog: "Vnet::LoadBalancerHealthProbes", label: "health_probe", stack: [{subject_id: id, vm_id: vm.id}], parent_id: strand.id)
       incr_rewrite_dns_records
     end
   end
 
   def detach_vm(vm)
-    load_balancers_vms_dataset.where(vm_id: vm.id).update(state: "detaching")
+    load_balancers_vms_dataset[vm_id: vm.id].update(state: "detaching")
```
fdr, line 43: also I think given what you are doing here, the dataset version you had previously was better (it avoids loading the model). the square brackets are nice syntactic upgrade to avoid `.first` on a `where` dataset, a common but not exhaustive case.

Also, as an aside: what's the concurrency situation here? What threads are writers, what threads react to this data? I see it gets called directly through routes: does that cause any clobbering? It feels like you want a semaphore, but perhaps you have some monotonic logic that this follows that makes it safe to get clobbered at any time. Is it?

prog/vnet/cert_nexus.rb
```diff
@@ -98,7 +98,11 @@ def before_run
 
   label def destroy
     # the reason is chosen as "cessationOfOperation"
-    acme_client.revoke(certificate: cert.cert, reason: REVOKE_REASON) if cert.cert
+    begin
+      acme_client.revoke(certificate: cert.cert, reason: REVOKE_REASON) if cert.cert
+    rescue Acme::Client::Error::AlreadyRevoked
+      Clog.emit("Certificate is already revoked")
```
fdr, line 104: I suppose there is no useful structured data here, i.e we get enough for free from Clog.emit?

prog/vnet/update_load_balancer_node.rb
```diff
@@ -39,7 +39,7 @@ def before_run
 
   def generate_lb_based_nat_rules
     public_ipv4 = vm.ephemeral_net4.to_s
-    public_ipv6 = vm.ephemeral_net6.to_s
+    public_ipv6 = vm.ephemeral_net6.nth(2).to_s
```
fdr, line 42: Neat, although ... I'm a little concerned that in this tiny, tiny bit of code we ascribe special meaning to another address in the IPv6 space. I mean, it's great that we can, but probably need to follow up with some kind of abstraction that centralizes and describes how the address space is used.

fdr (commented):
Moves in a good direction. some nits, seems like " Load Balancer refactor test updates " ought to get squashed for example.

More puzzling but not *caused* by this patch is some of the concurrency thinking in the data/prog model, which is hard for me to assess right off.

fdr:
Also, another nit, regarding the PR's description: some of this is refactoring, but a lot of it is not: it has at least several bug fixes.

It may seem almost too conversational and vague, but if you have an "omnibus" set of adjustments...which I think is most common right after integrating an new code...you can have a super vague PR title, like "Fix some bugs" or "Make a some adjustments." So long as the size is reviewable, I don't mind. My general guidance here is commit messages are where we put our formality, PR aggregations can be a lot looser.

In "refactoring" I take a specific meaning, where `2 * 6`  is changed to `2 * 2 * 3`, but in the algebra of the code, that is intended to have no semantic effect, and is purely a change in expression.

prog/vnet/update_load_balancer.rb
furkansahin: I have initially designed it to be all under a single entity LoadBalancer, moving forward, it's possible that the VM sourced flow is better but my dilemma is, whenever we perform a vm specific load balancer update, we need to update the other load balancer nodes as well. I guess, for now, I will opt into watch and see.

model/load_balancer.rb
furkansahin: There are 2 writer paths;
1. Routes for adding/removing VMs from LBs.
2. VM destroy path.

Reader is the LoadBalancerNexus. 
I want a semaphore but that semaphore need to be able to tell for which VM this operation is triggered specifically. Otherwise, the synchronisation requires proper snapshotting and a whole dance. That's why, here I wanted to handle the join table entity creation directly at the time of request arrival.

prog/vnet/cert_nexus.rb
furkansahin: We can print the whole exception but, yeah, there is nothing useful in there.

prog/vnet/update_load_balancer_node.rb
furkansahin: `vm.ephemeral_net6.nth(2)` is the public ip that we assign to customer, so, we are not picking another address from the address space for load balancing. We can, but I did not do simply because of the reason you describe. I want to keep it as a future improvement.

fdr: oh what was I thinking. I was thinking you were working with the special internal half of the space. nevermind.

model/load_balancer.rb
fdr: Are the states monotonic, e.g. up => evacuating => detaching, in which case a conditional update can do the job (so you don't clobber `detaching` with `evacuating`). I think it'll be quite difficult to keep a correct program if the states get clobbered back and forth and are not in a monotonic progression.

prog/vnet/cert_nexus.rb
```diff
@@ -98,7 +98,11 @@ def before_run
 
   label def destroy
     # the reason is chosen as "cessationOfOperation"
-    acme_client.revoke(certificate: cert.cert, reason: REVOKE_REASON) if cert.cert
+    begin
```
fdr, line 101: commit message:

>  In case Cert is already revoked, simply omit the error message 

...I mean, seems reasonable, but..why?  By error message, do you mean the stack trace? It's pretty different to go from "exception raised from the method" to "the exception is *not* raised". Seems more significant than a matter of error messages.

fdr (approved):
More gain than pain, though I have questions about concurrency interleaving. But, they were pre-existing, no reason to block, and there are some desirable fixes here

prog/vnet/update_firewall_rules.rb
```diff
@@ -88,9 +88,9 @@ class Prog::Vnet::UpdateFirewallRules < Prog::Base
     ip saddr @private_ipv4_cidrs ct state established,related,new counter accept
     ip daddr @private_ipv4_cidrs ct state established,related counter accept
     ip6 saddr @private_ipv6_cidrs ct state established,related,new counter accept
-    ip6 daddr @private_ipv6_cidrs ct state established,related counter accept
```
fdr, line 91: typo in commit message

When we started implementing firewalls,
we decided that the traffic in private network would be **excempt**

fdr (commented):
> Main reason of that bug is the
firewall rules implementation. When we started implementing firewalls,
we decided that the traffic in private network would be excempt from the
firewall rules.

I actually forgot why/when we decided this. It seems bad that, say, an array of web servers would have full access to one another without recourse because they live in the same subnet.

prog/vnet/update_load_balancer.rb
enescakir, line 33: In the next commit you already add this check to `before_run`. Do we need this extra branch and test here?

```ruby
  def before_run
    pop "VM is destroyed" unless vm
  end
```

enescakir (approved):
I don't have no other comment except what Daniel said.

prog/vnet/cert_nexus.rb
furkansahin: Alright, that makes sense, I have added the exception and the stack trace into the log.

fdr:
> > I actually forgot why/when we decided this. It seems bad that, say, an array of web servers would have full access to one another without recourse because they live in the same subnet. Not immediately clear to me why being in the subnet is necessarily privileged in this way.
> > Maybe we came up with a scenario where this was the least cruel cut, but if so, I forgot it.
> > Can you remind me?
> 
> It is because we are applying firewalls to the whole subnet. Therefore, it's not possible to selectively allow a couple of resources to a single resource.

But that's not what I quoted. The fact the firewall rules are *uniform* make sense, I remember we (okay, mostly I) though that was okay.

But this is different:

> firewall rules implementation. When we started implementing firewalls,
we decided that the traffic in private network would be excempt from the
firewall rules.

Is that what you meant to write? e.g. that, if a subnet is created for web servers, that they could contact each other without the platform offering control to the customer ("exempt from firewall rules")? Most web server fleets don't have any lateral communication between them.

PR #1829, opened 2024-08-05
Fix flaky test failure caused by array order

enescakir:
The order of the array may vary due to the physical scan order, and it causes test failures. However the order isn't crucial for our use case. Therefore, I switched from `eq` to the `match_array` matcher. Rubocop then updated it from `match_array` to the `contain_exactly` matcher.


fdr (approved):
sure, that seems fine, though it's a little weird that the program's correct output as far as mail headers go is a bit arbitrary.

PR #1830, opened 2024-08-05
Suppress exception stack trace for a expected exception in the test output

enescakir:
Recently, I enhanced the robustness of our GitHub cache reserving endpoint in commit 3129af12. To verify this feature, I created a test case that triggers an exception. Our web routes print a stack trace for unknown exceptions [^1], so this test case was designed to expect an exception. While the case was successful, it also printed the stack trace of the exception, which cluttered our test output. This commit resolves the issue by suppressing the stack trace, changing the exception to an expected exception.

[^1]: https://github.com/ubicloud/ubicloud/blob/9a0ad0305f8d7d807baadc1b64004d35a5eee50a/routes/clover_base.rb#L35

### DIFF
--- a/spec/routes/runtime/github_spec.rb
+++ b/spec/routes/runtime/github_spec.rb
@@ -84,7 +84,7 @@ RSpec.describe Clover, "github" do
       end
 
       it "Rollbacks inconsistent cache entry if a failure occurs in the middle" do
-        expect(blob_storage_client).to receive(:create_multipart_upload).and_raise("error")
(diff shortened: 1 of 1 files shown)


enescakir:
Polluted output

Screenshot 2024-08-05 at 21 25 52

furkansahin (approved):

fdr (approved):

fdr:
I'm merging this! It's test-only and a bit obnoxious in the output without the fix. Thanks

PR #1839, opened 2024-08-08
Enable CI to run on macOS runners

enescakir:
I wasn't aware of that macOS runners are free for open source projects. Our development team is mostly using macOS, so it would be great to have CI running on macOS runners. Previously, we hit macOS specific package issues.


enescakir:
macOS runner images don't have preinstalled docker 🔫 
Screenshot 2024-08-08 at 15 13 17

PR #1853, opened 2024-08-12
Use existing ruby toolcache for arm CI

enescakir:
You may experience some slowness with our CI runs on arm64 runners recently. This is because we add more tools to our arm64 runner images and the preinstalled Ruby is the one of them. This preinstalled Ruby conflicts with our workaround for caching Ruby on arm64 as they exist on the same path. However, we no longer need this adhoc cache. I have kept the `ruby-build` for instances where the image lacks the exact required version. If the required version isn't preinstalled, `actions/setup-ruby` will install it for x64 runners but will fail for arm64 runners. Currently, our images come with preinstalled Ruby 3.2.5, which we use. If this changes in the future, `ruby-build` will install the required version until we update the image.

GitHub runner images come with preinstalled versions of various languages like Ruby, Python, etc. Our x64 images also include preinstalled Ruby. However, as GitHub didn't have official arm64 images, our arm64 images did not come with preinstalled Ruby. We used to install and cache the required Ruby version in the workflow. Recently, GitHub launched arm64 runners and released a list of installed tools, including Ruby. Therefore, we have added preinstalled Ruby to our arm64 images to maintain compatibility with official GitHub images. This eliminates the need for adhoc caching.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,24 +40,18 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v4
 
-    - name: Cache ruby for ARM runners
-      id: cache-ruby
-      if: matrix.runs-on == 'ubicloud-arm'
-      uses: ubicloud/cache@v4
-      with:
-        path: ${{ runner.tool_cache }}/Ruby
-        key: ${{ matrix.runs-on }}-ruby-${{ hashFiles('.tool-versions') }}
-
     - name: Install ruby for ARM runners if not cached
-      if: matrix.runs-on == 'ubicloud-arm' && steps.cache-ruby.outputs.cache-hit != 'true'
+      if: matrix.runs-on == 'ubicloud-arm'
       run: |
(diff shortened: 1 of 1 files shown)


fdr (approved):
nice. though, why is there any code at all conditional on arm64, given it's part of the image and ruby is pretty portable?

e.g.

```
if [ ! -f $RUNNER_TOOL_CACHE/Ruby/$RUBY_VERSION/arm64.complete ];
```

enescakir:
> nice. though, why is there any code at all conditional on arm64, given it's part of the image and ruby is pretty portable?
> 
> e.g.
> 
> ```
> if [ ! -f $RUNNER_TOOL_CACHE/Ruby/$RUBY_VERSION/arm64.complete ];
> ```

`ruby/setup-ruby` lacks arm64 packages, so it automatically downloads x64 packages even when the tool cache is not present on the machine. https://github.com/ruby/setup-ruby/issues/577

Of course, we can run this code for x64 and built our own Ruby, but it's not needed given the availability of prebuilt packages

PR #1858, opened 2024-08-14
LB Certificate Server

furkansahin:
## Add Ubicloud IPv6 address space to the namespace
This commit adds the new address space "oo-bic-lood-safe-ce" to the tap
devices.  This address space will be used for providing private endpoint
to the VMs so that we can serve certificate and key files.

In relation to that, we are making another change to add some nftables
rules to the cloudinit.img. With the help of new nftables rules inside
the VM, we are blocking access to the "oo-bic-lood-safe" address space
if the user is not root which has skuid 0.

## Add Cert Server initializer
We are adding new states to the UpdateLoadBalancerNode prog to;
1. Publish cert and key files to the /vm/vmxxxxx/cert directory
2. Create a new web server to serve the files to the VM.
3. Remove the files and service in case the node is removed from the
loadbalancer.

We are utilizing go as the web server. The server source code lives in
ubicloud/cert-server. We are also starting it on the private address
[FD00:0B1C:100D:5AFE:CE::]:8080 because it is cool and looks like
"oo-bic-lood-safe-ce".

## Add certificate server initialization to load_balancer.add_vm
We are simply creating a prog to initialize the certificate server and
publish the cert files to the vm folder at the time of add_vm.

## LoadBalancerNexus update to reshare certificates
This commit makes the changes necessary to update the certificates and
the web server in case we provision new certificates.

## Update health probes to use the hostname and check return type
We are updating the healthprobes to use the hostname to perform the
healthchecks. This is important for HTTPS healthchecks because it is
possible the customers will only accept the hostname, instead of raw ip.
Since we are in the network namespace, and we do not need to perform
lookup, we resolve the hostname automatically by using --resolve of
curl.

We also make a small fix here to transform response_code to integer and
perform the check. This is necessary in case the response_code has
spaces or \n characters at the start or end of the response code.

## Fix VM removal when its attached to a load balancer
We need to remove the vm user only after the cert server is removed
because it is used by cert server and deluser command fails otherwise.

The other case is that, we might start deleting VM and then the load
balancer is deleted before VM deletion completes. In that case, we
should perform a check at the last state to verify the load balancer is
still intact before removing VM from the load balancer.

### DIFF
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -35,12 +35,14 @@ class LoadBalancer < Sequel::Model
     DB.transaction do
       super
       Strand.create_with_id(prog: "Vnet::LoadBalancerHealthProbes", label: "health_probe", stack: [{subject_id: id, vm_id: vm.id}], parent_id: strand.id)
+      Strand.create_with_id(prog: "Vnet::CertServer", label: "put_certificate", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
       incr_rewrite_dns_records
     end
   end
 
   def detach_vm(vm)
     load_balancers_vms_dataset.where(vm_id: vm.id, state: ["up", "down", "evacuating"]).update(state: "detaching")
+    Strand.create_with_id(prog: "Vnet::CertServer", label: "remove_cert_server", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
     remove_health_probe(vm.id)
     incr_update_load_balancer
   end
@@ -49,6 +51,7 @@ class LoadBalancer < Sequel::Model
     DB.transaction do
       load_balancers_vms_dataset.where(vm_id: vm.id, state: ["up", "down"]).update(state: "evacuating")
       remove_health_probe(vm.id)
+      Strand.create_with_id(prog: "Vnet::CertServer", label: "remove_cert_server", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
       incr_update_load_balancer
       incr_rewrite_dns_records
     end
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -45,7 +45,11 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
       rewrite_dns_records
       decr_rewrite_dns_records
     end
-    hop_create_new_cert if load_balancer.need_certificates?
+
+    if load_balancer.need_certificates?
+      load_balancer.incr_refresh_cert
+      hop_create_new_cert
+    end
 
     nap 5
   end
@@ -59,9 +63,25 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
   label def wait_cert_provisioning
     if load_balancer.need_certificates?
       nap 60
-    else
+    elsif load_balancer.refresh_cert_set?
+      load_balancer.vms.each do |vm|
+        bud Prog::Vnet::CertServer, {"subject_id" => load_balancer.id, "vm_id" => vm.id}, :reshare_certificate
+      end
+
+      hop_wait_cert_broadcast
+    end
+
+    hop_wait
+  end
+
+  label def wait_cert_broadcast
+    reap
+    if strand.children.select { _1.prog == "Vnet::CertServer" }.all? { _1.exitval == "certificate is reshared" } || strand.children.empty?
+      decr_refresh_cert
       hop_wait
     end
+
+    nap 1
   end
 
   label def update_vm_load_balancers
@@ -92,7 +112,8 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
     end
 
     load_balancer.vms.each do |vm|
-      bud Prog::Vnet::UpdateLoadBalancerNode, {"subject_id" => vm.id, "load_balancer_id" => load_balancer.id}, :remove_load_balancer
+      bud Prog::Vnet::UpdateLoadBalancerNode, {"subject_id" => vm.id, "load_balancer_id" => load_balancer.id}, :update_load_balancer
+      bud Prog::Vnet::CertServer, {"subject_id" => load_balancer.id, "vm_id" => vm.id}, :remove_cert_server
     end
 
     hop_wait_destroy
--- a/prog/vnet/update_load_balancer_node.rb
+++ b/prog/vnet/update_load_balancer_node.rb
@@ -18,7 +18,6 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
   label def update_load_balancer
     if vm_load_balancer_state == "detaching"
       load_balancer.remove_vm(vm)
-      hop_remove_load_balancer
     end
 
     # if there is literally no up resources to balance for, we simply not do
@@ -31,7 +30,8 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
 
   label def remove_load_balancer
     vm.vm_host.sshable.cmd("sudo ip netns exec #{vm.inhost_name} nft --file -", stdin: generate_nat_rules(vm.ephemeral_net4.to_s, vm.nics.first.private_ipv4.network.to_s))
-    pop "load balancer is updated"
+
+    pop "load balancer is removed"
   end
 
   def generate_lb_based_nat_rules
--- a/rhizome/host/lib/cert_server_setup.rb
+++ b/rhizome/host/lib/cert_server_setup.rb
@@ -1,0 +1,120 @@
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require_relative "../../common/lib/arch"
+require_relative "vm_path"
+require "fileutils"
+require "json"
+
+class CertServerSetup
+  def initialize(vm_name)
+    @vm_name = vm_name
+  end
+
+  def vp
+    @vp ||= VmPath.new(@vm_name)
+  end
+
+  def cert_folder
+    vp.q_cert
+  end
+
+  def cert_path
+    "#{cert_folder}/cert.pem"
+  end
+
+  def key_path
+    "#{cert_folder}/key.pem"
+  end
+
+  def service_name
+    "#{@vm_name}-metadata-endpoint"
+  end
+
+  def service_file_path
+    "/etc/systemd/system/#{service_name}.service"
+  end
+
+  def server_main_path
+    File.join("", "opt", "metadata-endpoint")
+  end
+
+  def vm_server_path
+    File.join(cert_folder, "metadata-endpoint")
+  end
+
+  def package_url
+    Arch.render(
+      x64: "https://github.com/ubicloud/metadata-endpoint/releases/download/v0.1.3/metadata-endpoint_Linux_x86_64.tar.gz",
+      arm64: "https://github.com/ubicloud/metadata-endpoint/releases/download/v0.1.3/metadata-endpoint_Linux_arm64.tar.gz"
+    )
+  end
+
+  def copy_server
+    unless File.exist?(server_main_path)
+      download_server
+    end
+
+    r "cp #{server_main_path}/metadata-endpoint #{vm_server_path}"
+    r "sudo chown #{@vm_name}:#{@vm_name} #{vm_server_path}"
+  end
+
+  def download_server
+    temp_tarball = "/tmp/metadata-endpoint.tar.gz"
+    r "curl -L3 -o #{temp_tarball} #{package_url}"
+
+    FileUtils.mkdir_p(server_main_path)
+    FileUtils.cd server_main_path do
+      r "tar -xzf #{temp_tarball}"
+    end
+  end
+
+  def create_service
+    service = "#{service_name}.service"
+    File.write("/etc/systemd/system/#{service}", <<CERT_SERVICE
+[Unit]
+Description=Certificate Server
+After=network.target
+
+[Service]
+NetworkNamespacePath=/var/run/netns/#{@vm_name}
+ExecStart=#{vm_server_path}
+Restart=always
+Type=simple
+ProtectSystem=strict
+PrivateDevices=yes
+PrivateTmp=yes
+ProtectHome=yes
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+ProtectControlGroups=yes
+NoNewPrivileges=yes
+ReadOnlyPaths=#{cert_path} #{key_path}
+User=#{@vm_name}
+Group=#{@vm_name}
+Environment=VM_INHOST_NAME=#{@vm_name}
+Environment=IPV6_ADDRESS="FD00:0B1C:100D:5AFE:CE::"
+Environment=GOMEMLIMIT=9MiB
+Environment=GOMAXPROCS=1
+CPUQuota=50%
+MemoryLimit=10M
+
+CERT_SERVICE
+    )
+  end
+
+  def enable_and_start_service
+    r "systemctl enable #{service_file_path}"
+    r "systemctl start #{service_name}"
+  end
+
+  def stop_and_remove_service
+    r "systemctl stop #{service_name}"
+    r "systemctl disable #{service_name}"
+    FileUtils.rm_f(service_file_path)
+  end
+
+  def remove_paths
+    FileUtils.rm_rf(cert_folder)
+  end
+end
--- a/prog/vnet/cert_server.rb
+++ b/prog/vnet/cert_server.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+class Prog::Vnet::CertServer < Prog::Base
+  subject_is :load_balancer
+
+  def vm
+    @vm ||= Vm[frame.fetch("vm_id")]
+  end
+
+  def cert_folder
+    "/vm/#{vm.inhost_name}/cert"
+  end
+
+  def cert_path
+    "#{cert_folder}/cert.pem"
+  end
+
+  def key_path
+    "#{cert_folder}/key.pem"
+  end
+
+  label def before_run
+    pop "vm is destroyed" unless vm
+  end
+
+  label def reshare_certificate
+    put_cert_to_vm
+
+    pop "certificate is reshared"
+  end
+
+  label def put_certificate
+    nap 5 unless load_balancer.active_cert
+
+    put_cert_to_vm
+    hop_start_certificate_server
+  end
+
+  label def start_certificate_server
+    vm.vm_host.sshable.cmd("sudo host/bin/setup-cert-server setup #{vm.inhost_name}")
+    pop "certificate server is started"
+  end
+
+  label def remove_cert_server
+    vm.vm_host.sshable.cmd("sudo host/bin/setup-cert-server stop_and_remove #{vm.inhost_name}")
+    pop "certificate resources and server are removed"
+  end
+
+  def put_cert_to_vm
+    cert = load_balancer.active_cert
+
+    cert_payload = cert.cert
+    cert_key_payload = OpenSSL::PKey::RSA.new(cert.csr_key).to_pem
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} mkdir -p #{cert_folder}")
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} tee #{cert_path}", stdin: cert_payload)
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} tee #{key_path}", stdin: cert_key_payload)
+  end
+end
--- a/rhizome/host/bin/setup-cert-server
+++ b/rhizome/host/bin/setup-cert-server
@@ -1,0 +1,28 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require_relative "../lib/cert_server_setup"
+require "fileutils"
+
+unless (verb = ARGV.shift)
+  puts "expected verb as argument"
+  exit 1
+end
+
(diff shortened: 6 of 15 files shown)


prog/vnet/update_load_balancer_node.rb
```diff
@@ -29,6 +41,70 @@ def before_run
...
+    service_name = "#{vm.inhost_name}_cert_server"
+    service_file_path = "/vm/#{vm.inhost_name}/#{service_name}.service"
+    systemd_service = <<SYSTEMD
+[Unit]
+Description=Certificate Server
+After=network.target
+
+[Service]
+NetworkNamespacePath=/var/run/netns/#{vm.inhost_name}
+ExecStart=busybox httpd -f -p [FD00:0B1C:100D:CE::]:8080 -h #{cert_folder}
```
fdr, line 65: interesting choice. One thing I'm wondering is:

1) do we really want a memory-unsafe HTTP server
2) can whatever http server used protect the system from very large or other kinds of malicious requests?

prog/vnet/load_balancer_nexus.rb
```diff
@@ -60,8 +60,20 @@ def before_run
...
+      load_balancer.vms.each do |vm|
+        bud Prog::Vnet::UpdateLoadBalancerNode, {"subject_id" => vm.id, "load_balancer_id" => load_balancer.id}, :put_certificate
+      end
+      hop_wait_cert_broadcast
+    end
+  end
+
+  label def wait_cert_broadcast
+    reap
+    if strand.children_dataset.where(prog: "Vnet::UpdateLoadBalancerNode").all? { _1.exitval == "certificate server is started" } || strand.children.empty?
```
fdr, line 72: this seems a little weird because `reap` already deletes stuff. It seems like you'd always be falling into the `empty?` case? are there other children?

fdr (changes requested):
the `reap` thing is a bit weird. The other thing I'm wondering is if we should add a few more bytes to the private region, quite possibly we'll have endpoints that do _not_ have secret information, that we want to allow unprivileged users to access. Maybe. Or maybe not and we should swipe the entire /48.

Maybe like:

5afe aka "safe"

prog/vnet/update_load_balancer_node.rb
furkansahin: 1. I mainly chose this one because it was given by perplexity as the most lightweight web server. Maybe we can introduce the memory limit into the systemd file?
2. Considering the address we bind is completely blocked for non sudo user, the malicious actor would be the customer or a program they run in sudo. Still we need to rate-limit I guess. 

As an alternative, there is caddy from ZeroSSL https://github.com/caddyserver/caddy. It is written in GO. It also has a rate-limiting module but from a third party and it requires a re-build of caddy (https://github.com/RussellLuo/caddy-ext/tree/master/ratelimit) :/ 

I guess we can still use caddy (or if you have any other suggestion) + systemd configs.

prog/vnet/load_balancer_nexus.rb
furkansahin: The health probes are also a child strand of load balancer strand.

prog/vnet/update_load_balancer_node.rb
furkansahin: I found another way, we can still install caddy from the package but implement rate-limiting via nftables. This would give us static rate limiting. We can also configure caddy to limit the request size. How does it sound?

prog/vnet/update_load_balancer_node.rb
```diff
@@ -29,6 +41,68 @@ def before_run
...
+    cert_key_payload = OpenSSL::PKey::RSA.new(cert.csr_key).to_pem
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} mkdir -p #{cert_folder}")
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} tee #{cert_path}", stdin: cert_payload)
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} tee #{key_path}", stdin: cert_key_payload)
+    hop_start_certificate_server
+  end
+
+  label def start_certificate_server
+    service_name = "#{vm.inhost_name}_cert_server"
+    service_file_path = "/vm/#{vm.inhost_name}/#{service_name}.service"
```
enescakir, line 56: We put vm and dnsmasq service to `/etc/systemd/system` directory. Should put it to the same directory?
```ruby
  def dnsmasq_service
    "/etc/systemd/system/#{@vm_name}-dnsmasq.service"
  end

  def systemd_service
    File.join("/etc/systemd/system",
      IO.popen(["systemd-escape", @vm_name + ".service"]) { _1.read.chomp })
  end
```

prog/vnet/update_load_balancer_node.rb
```diff
@@ -29,6 +41,68 @@ def before_run
...
+    cert_payload = cert.cert
+    cert_key_payload = OpenSSL::PKey::RSA.new(cert.csr_key).to_pem
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} mkdir -p #{cert_folder}")
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} tee #{cert_path}", stdin: cert_payload)
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} tee #{key_path}", stdin: cert_key_payload)
+    hop_start_certificate_server
+  end
+
+  label def start_certificate_server
+    service_name = "#{vm.inhost_name}_cert_server"
```
enescakir, line 55: We use `-` in dnsmasq service. 
```suggestion
    service_name = "#{vm.inhost_name}-cert-server"
```

prog/vnet/update_load_balancer_node.rb
```diff
@@ -29,6 +41,68 @@ def before_run
...
+NoNewPrivileges=yes
+ReadOnlyPaths=#{cert_path} #{key_path}
+User=#{vm.inhost_name}
+Group=#{vm.inhost_name}
+SYSTEMD
+
+    vm.vm_host.sshable.cmd("sudo -u #{vm.inhost_name} tee #{service_file_path}", stdin: systemd_service)
+
+    start_server_cmds = <<CMD
+systemctl enable #{service_file_path}
```
enescakir, line 83: I know you don't like putting code to the dataplane. But I just want to say this changes remind me the `rhizome/host/lib/spdk_setup.rb` file. We pass secrets via stdin, then create service files, enable/start services, then delete them while removing etc.

prog/vnet/update_load_balancer_node.rb
furkansahin: I added caddy switch to this commit, it's still missing the rate-limiter, what do you think?
https://github.com/ubicloud/ubicloud/commit/f9daf854d9c32e510f54fcd3d37d630ee32e8e02

prog/vnet/load_balancer_nexus.rb
fdr: given `reap` loads `children` (right? I read the source code last night, but it's particular to use `delete_if` to keep it in sync), I'd probably use the in-memory data structure instead.

prog/vnet/update_load_balancer_node.rb
fdr: Okay, probably an improvement, of my informal survey of web servers a year ago, caddy gave some decent vibes (and has a request size limit, we can set it to, like 1-8K presumably). But uh, I kinda feel like we should do what @byucesoy and I did for DNS servers to choose the implementation, that is, identify a shortlist of projects and analyze their development process and code size. In my recollection, most can be dismissed out of hand: caddy is a decent bar to set. Notably absent from my survey for no particular reason was the rust and crystal server approaches.

fdr: I was doing a quick survey on the rust/crystal o'sphere and I'm not seeing anything super compelling, but I haven't done it systematically. I'd say 80% chance of caddy, but screw down the request size and take a look at the memory "RES" situation.

furkansahin:
I have handled "5AFE" and systemd control/dataplane related moves, also the cert broadcast to VMs moved into its own Prog. This will require further rewrite of tests + The server picking. I'll do that later, still operationally, this works and reviewable.

prog/vnet/update_load_balancer_node.rb
furkansahin: Looking into this further and without properly testing myself, here is AI supported analysis;
1. Caddy, written in GO, kind of large code base, actively developed, feature rich, possibly memory usage on the high end.
2. Warp, written in Rust, small code base (25k loc), actively developed, not so feature rich, memory usage is low. The problem is, we'll need to compile and run this ourselves, binary is not available.
3. Kemal, written in Crystal, very small code base (5k), not much development activity, not so feature rich, memory usage is expected to be higher than warp but lower than caddy. The problem is, we'll need to compile and run this ourselves, binary is not available.

There are some alternatives that are written in C but we avoid them. I also found some written in python but I assume we avoid that, too.

Giving it some more thought, I'm thinking, we already have ruby environment setup, we are already experienced in puma, it may require more memory consumption but, what about it? We can also consider puma for the sake of developer familiarity. 

I need to perform controlled tests to verify each of these memory consumptions.

furkansahin: Alright, here are my findings
1. busybox;
```
[root@vhgwk6zhtcqj4jh8f1we0wet1m vmv4xy42]# pidstat --human -r -u -p 993481 10
Linux 5.15.0-100-generic (vhgwk6zhtcqj4jh8f1we0wet1m) 	08/21/2024 	_x86_64_	(12 CPU)
^[
02:42:46 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
02:42:56 PM 33540    993481    0.0%    0.0%    0.0%    0.0%    0.0%    10  busybox

02:42:46 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
02:42:56 PM 33540    993481      0.00      0.00    2.4M    4.0k   0.0%  busybox

02:42:56 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
02:43:06 PM 33540    993481    0.0%    0.0%    0.0%    0.0%    0.0%    10  busybox

02:42:56 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
02:43:06 PM 33540    993481      0.00      0.00    2.4M    4.0k   0.0%  busybox

02:43:06 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
02:43:16 PM 33540    993481    0.1%    0.0%    0.0%    0.0%    0.1%    10  busybox
```
2. Warp;
```
^C[root@vhgwk6zhtcqj4jh8f1we0wet1m vmv4xy42]# pidstat --human -r -u -p 1005520 10
Linux 5.15.0-100-generic (vhgwk6zhtcqj4jh8f1we0wet1m) 	08/21/2024 	_x86_64_	(12 CPU)

04:01:57 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
04:02:07 PM     0   1005520    0.1%    0.0%    0.0%    0.0%    0.1%     3  warp_server

04:01:57 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
04:02:07 PM     0   1005520      0.50      0.00  870.0M    6.2M   0.0%  warp_server

04:02:07 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
04:02:17 PM     0   1005520    0.0%    0.0%    0.0%    0.0%    0.0%     3  warp_server

04:02:07 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
04:02:17 PM     0   1005520      0.10      0.00  870.0M    6.2M   0.0%  warp_server
```

3. Kemal
```
[root@vhgwk6zhtcqj4jh8f1we0wet1m ~]# pidstat --human -r -u -p 992684 10
Linux 5.15.0-100-generic (vhgwk6zhtcqj4jh8f1we0wet1m) 	08/21/2024 	_x86_64_	(12 CPU)

02:38:47 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
02:38:57 PM     0    992684    0.0%    0.0%    0.0%    0.0%    0.0%     8  crystal

02:38:47 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
02:38:57 PM     0    992684      0.00      0.00  647.7M  547.1M   0.4%  crystal

02:38:57 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
02:39:07 PM     0    992684    0.0%    0.0%    0.0%    0.0%    0.0%     8  crystal

02:38:57 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
02:39:07 PM     0    992684      0.00      0.00  647.7M  547.1M   0.4%  crystal

02:39:07 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
02:39:17 PM     0    992684    0.0%    0.0%    0.0%    0.0%    0.0%     8  crystal

02:39:07 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
02:39:17 PM     0    992684      0.00      0.00  647.7M  547.1M   0.4%  crystal
```

4. Caddy
```
[root@vhgwk6zhtcqj4jh8f1we0wet1m vmv4xy42]# pidstat --human -r -u -p 437776 10
Linux 5.15.0-100-generic (vhgwk6zhtcqj4jh8f1we0wet1m) 	08/21/2024 	_x86_64_	(12 CPU)

04:04:58 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
04:05:08 PM   998    437776    0.0%    0.0%    0.0%    0.0%    0.0%    10  caddy

04:04:58 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
04:05:08 PM   998    437776      0.00      0.00    1.2G   17.5M   0.0%  caddy

04:05:08 PM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
04:05:18 PM   998    437776    0.0%    0.0%    0.0%    0.0%    0.0%    10  caddy

04:05:08 PM   UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
04:05:18 PM   998    437776      0.00      0.00    1.2G   17.5M   0.0%  caddy
```

So, according to these results, if we only consider RSS, the order is;
1. busybox (4k)
2. Warp (6.2M)
3. Caddy (17.5M)
4. Kemal (547.1M)

if we consider VSZ;
1. busybox (2.4M)
2. Kemal (647.7M)
3. Warp (870M)
4. Caddy (1.2G)
Since busybox is not memory safe, I feel like we are picking either Warp or Caddy. Considering the activity, being backed by ZeroSSL itself, I am more inclined to go with Caddy but 1.2G Virtual memory is huge! Plus, the codebase is very big. So, I'm happy to go with warp, too. 

Another alternative, I feel like we can limit memory size via systemd and use busybox, too.

enescakir: I'm not sure if it can easily serve static files, but @bsatzger is using [cloudflare/pingora](https://github.com/cloudflare/pingora) for API gateway implementation. We could consider introducing a single dependency

fdr: We're not doing a proxy thing though, right? It's serving a small piece of text. And maybe other metadata in the future.

Here's a dumb thought: how about Go's standard library web server?

https://www.perplexity.ai/search/write-me-a-go-program-that-loa-Hi9Q69k1QoO7ARp8ZPeOwQ

```go
package main

import (
	"fmt"
	"io/ioutil"
	"log"
	"net/http"
	"time"

	"golang.org/x/time/rate"
)

// Define the rate limiter
var limiter = rate.NewLimiter(10, 1) // 10 requests per second

func certHandler(w http.ResponseWriter, r *http.Request) {
	// Check if the request can proceed
	if !limiter.Allow() {
		http.Error(w, "Too Many Requests", http.StatusTooManyRequests)
		return
	}

	// Limit the size of the request body
	r.Body = http.MaxBytesReader(w, r.Body, 1024) // Limit to 1KB

	// Read the cert.pem file
	data, err := ioutil.ReadFile("cert.pem")
	if err != nil {
		http.Error(w, "File not found", http.StatusNotFound)
		return
	}

	// Write the file contents to the response
	w.Header().Set("Content-Type", "application/x-pem-file")
	w.WriteHeader(http.StatusOK)
	w.Write(data)
}

func main() {
	// Set up the HTTP server
	http.HandleFunc("/cert.pem", certHandler)

	// Start the server
	port := ":8080"
	fmt.Printf("Starting server on port %s\n", port)
	if err := http.ListenAndServe(port, nil); err != nil {
		log.Fatalf("Could not start server: %s\n", err)
	}
}
```

fdr: But uh, basically, I had these general goal for the web server: rate limiting, small code, institutional stability, memory safety, maybe some graceful programming option.

Caddy has rate limiting, memory safety, and institutional stability.

I think the Rust options still kind of froth around, so without radically smaller sizes (much closer to busybox) it's harder to get excited about those.

But for an endpoint or five, perhaps straight up Go net/http will be a good set-and-forget option, even more so than a much larger and more complex program in Caddy.

fdr: I had Perplexity give some ideas on writing tests https://www.perplexity.ai/search/write-me-a-go-program-that-loa-Hi9Q69k1QoO7ARp8ZPeOwQ#2

model/load_balancer.rb
```diff
@@ -35,12 +35,14 @@ def add_vm(vm)
...
       super
       Strand.create_with_id(prog: "Vnet::LoadBalancerHealthProbes", label: "health_probe", stack: [{subject_id: id, vm_id: vm.id}], parent_id: strand.id)
+      Strand.create_with_id(prog: "Vnet::CertServer", label: "put_certificate", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
       incr_rewrite_dns_records
     end
   end
 
   def detach_vm(vm)
     load_balancers_vms_dataset.where(vm_id: vm.id, state: ["up", "down", "evacuating"]).update(state: "detaching")
+    Strand.create_with_id(prog: "Vnet::CertServer", label: "remove_cert_server", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
```
enescakir, line 45: Why don't we `bud` them since it has `parent_id`?

enescakir: Okay got it. It's in the `model` not a `prog`

prog/vnet/update_load_balancer_node.rb
furkansahin: I have done that. the new server lives in https://github.com/ubicloud/cert-server. We consume it here through github packages once and then copy the binary locally per VM.

fdr: should it be cert-server or should this be the metadata-endpoint server (or some other name less customary but also less...insipid?)

For example, finally, at some length, we could use `cloud-init` in a non-basic capacity. (does it matter? should we use `NoCloud` forever?!) 

I realize all we need is two HTTP endpoints today, but I suspect we're going to want to surface more stuff (should it be like `load-balancer/cert.pem`?)

I think it's very good that the simple case does not involve any dynamic text, like "load-balancer/#{lbidhere}`" because it's great to be able to hard code this. People with advanced use cases and multiple load balancers can deal with that in a more elaborate way.

All in all, these comments mean only two things:

1. Should we add an additional constant string to the URI to contain load-balancer related metadata data?
2. Maybe the project could be named differently, but that's not an interface with the user, so, kind of whatever, it's up to you if you feel like changing binary names later or whatnot.

fdr: I did what I could to make a new name, but none of the alternates I could get an LLM to generate really offer much in the way of improvement. Interesting quick overview of the various alternatives, though:

https://www.perplexity.ai/search/tell-me-about-metadata-endpoin-3ssyS24iRbWUqDUFB8XiNw

rhizome/host/lib/cert_server_setup.rb
```diff
@@ -0,0 +1,123 @@
...
+  end
+
+  def package_url
+    if Arch.arm64?
+      "https://github.com/ubicloud/metadata-endpoint/releases/download/v0.1.3/metadata-endpoint_Linux_arm64.tar.gz"
+    elsif Arch.x64?
+      "https://github.com/ubicloud/metadata-endpoint/releases/download/v0.1.3/metadata-endpoint_Linux_x86_64.tar.gz"
+    else
+      fail "BUG: unexpected architecture"
+    end
```
enescakir, line 53: nit: `Arch` has `.render` helper. It makes reduce the number of branches to tests. 
https://github.com/ubicloud/ubicloud/blob/a8952c1f15cba9a15f33d69b3f727494cab64a4b/rhizome/common/lib/arch_class.rb#L25

```suggestion
    Arch.render(
        x64: "https://github.com/ubicloud/metadata-endpoint/releases/download/v0.1.3/metadata-endpoint_Linux_x86_64.tar.gz",
        arm64: "https://github.com/ubicloud/metadata-endpoint/releases/download/v0.1.3/metadata-endpoint_Linux_arm64.tar.gz")
```

rhizome/host/lib/cert_server_setup.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  def download_server
+    temp_tarball = "/tmp/metadata-endpoint.tar.gz"
+    r "curl -L3 -o #{temp_tarball} #{package_url}"
+
+    FileUtils.mkdir_p(server_main_path)
+    FileUtils.cd server_main_path do
+      r "tar -xzf #{temp_tarball}"
+    end
+  end
```
enescakir, line 73: What happens when two virtual machines download the it at the same time on a fresh VM host?

PR #1864, opened 2024-08-17
Add cache proxy for projects with transparent cache flag

velioglu:
This commit introduces support for transparent cache, enabling the use of our own infrastructure for saving and restoring caches without changing anything on the test workflow files. The proxy is now run on runner VMs to intercept and handle all cache-related requests. The proxy will only be activated for projects where the transparent_cache flag is set to true.

Key changes include:

**Proxy Setup:** A proxy is deployed on the runner VMs to handle cache requests. This proxy intercepts requests intended for GitHub's cache API and redirects them to our infrastructure.

**Forked Runner Script:** A forked version of the runner script is downloaded and used instead of the pre-existing one on the VM. This modification is necessary because redirecting requests from GitHub's cache URI isn't straightforward. By setting CUSTOM_ACTIONS_CACHE_URL to localhost:8080, all cache requests are directed to the proxy running locally on port 8080.

**Blob Storage Redirection:** Requests to localhost.blob.core.windows.net are also redirected to the local proxy, as GitHub's toolkit directly accesses blob storage during file restoration rather than routing through the cache API.

Note that the current implementation downloads and runs the forked runner and proxy while initiating runner vm for ease of testing. In future iterations, both steps will be integrated into the image generation process to reduce VM setup time.

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -220,6 +220,69 @@ class Prog::Vm::GithubRunner < Prog::Base
     # Remove comments and empty lines before sending them to the machine
     vm.sshable.cmd(command.gsub(/^(\s*# .*)?\n/, ""))
 
+    if github_runner.installation.project.get_ff_transparent_cache
+      hop_setup_forked_runner
+    end
+
+    hop_register_runner
+  end
+
+  label def setup_forked_runner
+    command = <<~COMMAND
+      curl --output actions-runner.tar.gz -L #{Config.github_cache_forked_runner_tarball_uri}
+
+      rm -rf actions-runner
+      mkdir -p actions-runner
+      tar xzf actions-runner.tar.gz -C actions-runner
+
+      sudo chown -R runneradmin:runneradmin ./actions-runner
+      ./actions-runner/env.sh
+
+      cat <<EOT > ./actions-runner/run-withenv.sh
+      #!/bin/bash
+      mapfile -t env </etc/environment
+      exec env -- "\\${env[@]}" ./actions-runner/run.sh --jitconfig "\\$1"
+      EOT
+      chmod +x ./actions-runner/run-withenv.sh
+
+      jq '. += [#{setup_info.to_json}]' /imagegeneration/imagedata.json > ./actions-runner/.setup_info
+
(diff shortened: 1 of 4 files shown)


prog/vm/github_runner.rb
```diff
@@ -220,6 +220,71 @@ def setup_info
...
+      EOT
+      chmod +x ./actions-runner/run-withenv.sh
+
+      jq '. += [#{setup_info.to_json}]' /imagegeneration/imagedata.json | sudo -u runner tee ./actions-runner/.setup_info
+
+      echo "PATH=$PATH" >> ./actions-runner/.env
+
+      rm -rf /home/runner/actions-runner
+      sudo mv ./actions-runner /home/runner/
+      sudo chown -R runner:runner /home/runner/actions-runner
```
enescakir, line 256: What do you think about directly extracting to `/home/runner/actions-runner` using the `--overwrite` flag?

```suggestion
      sudo tar xzf actions-runner.tar.gz -C /home/runner/actions-runner --overwrite
```

velioglu: I've tried that but it couldn't set the permissions properly. Also, I wanted to be more explicit, so keeping it as it is.

prog/vm/github_runner.rb
```diff
@@ -220,6 +220,66 @@ def setup_info
...
+
+      echo "CUSTOM_ACTIONS_CACHE_URL=http://localhost:8080/spaceholder/" | sudo tee -a /etc/environment
```
enescakir, line 255: Should we use more randomized ports instead of `8080`? Since `8080` is commonly used for development and testing, it might already be part of the customer's workflow

prog/vm/github_runner.rb
```diff
@@ -220,6 +220,69 @@ def setup_info
...
+
+    vm.sshable.cmd(command)
+
+    hop_start_proxy
+  end
+
+  label def start_proxy
+    command = <<~COMMAND
+      export UBICLOUD_CACHE_URL=#{Config.base_url}/runtime/github/
+      export UBICLOUD_RUNTIME_TOKEN=#{vm.runtime_token}
```
enescakir, line 279:
```suggestion
      source /etc/environment
```

PR #1866, opened 2024-08-19
Use more descriptive region names in UI

byucesoy:
We have been using region names like "eu-central-h1" in the UI which is cryptic for users. We decided to simplify them and use more descriptive names referring to the actual location of the respective data centers. What we would do in case of having multiple data centers in the location is an open question currently, we would probably use the city name or a number suffix. However, it does not look like we would have that problem soon.

We are keeping the names in the API as is, because we don't want to introduce a breaking change at the moment.

image

### DIFF
--- a/config/providers.yml
+++ b/config/providers.yml
@@ -3,17 +3,17 @@
     provider_display_name: Hetzner,
     provider_internal_name: hetzner,
     locations: [
-      {display_name: eu-central-h1, internal_name: hetzner-fsn1, visible: true },
-      {display_name: eu-north-h1, internal_name: hetzner-hel1, visible: true },
-      {display_name: github-runners, internal_name: github-runners, visible: false }
+      {display_name: eu-central-h1, internal_name: hetzner-fsn1, ui_name: "Germany", visible: true },
+      {display_name: eu-north-h1, internal_name: hetzner-hel1, ui_name: "Finland", visible: true },
+      {display_name: github-runners, internal_name: github-runners, ui_name: "GithubRunner", visible: false }
     ]
   },
   {
     provider_display_name: Leaseweb,
     provider_internal_name: leaseweb,
     locations: [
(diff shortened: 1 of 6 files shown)


config/providers.yml
```diff
@@ -3,17 +3,17 @@
     provider_display_name: Hetzner,
     provider_internal_name: hetzner,
     locations: [
-      {display_name: eu-central-h1, internal_name: hetzner-fsn1, visible: true },
-      {display_name: eu-north-h1, internal_name: hetzner-hel1, visible: true },
-      {display_name: github-runners, internal_name: github-runners, visible: false }
+      {display_name: eu-central-h1, internal_name: hetzner-fsn1, ui_name: Germany, visible: true },
```
enescakir, line 6: Our regions currently have 3 different names: `internal_name`, `display_name`, and `ui_name`. 🤯  I wish to combine `internal` and `display` at some point

config/providers.yml
```diff
@@ -3,17 +3,17 @@
...
     ]
   },
   {
     provider_display_name: Leaseweb,
     provider_internal_name: leaseweb,
     locations: [
-      {display_name: us-east-a2, internal_name: leaseweb-wdc02, visible: false},
-      {display_name: us-south-a13, internal_name: leaseweb-dal13, visible: false},
+      {display_name: us-east-a2, internal_name: leaseweb-wdc02, ui_name: Washington D.C., US, visible: false},
+      {display_name: us-south-a13, internal_name: leaseweb-dal13,  ui_name: Dallas, US, visible: false},
```
enescakir, line 16: How `yml` support that? The name has `,` in it, also the delimiter is `,`. Should we put it into double quotes?

byucesoy: Good catch, updated.

enescakir (approved):

PR #1867, opened 2024-08-19
Centralize business logic of Vm web and api endpoints

byucesoy:
This is continuation from #1658, where we did a similar thing for Postgres endpoints. The goal is doing this for all resources, so that we wouldn't need to write a similar code twice. Like before, I used the API code as base to ensure that we don't introduce unintended breaking changes.

**Add helper class for Vm routes**
The business logic in both web and api endpoints are combined in the helper
class. Currently the combination is made simply adding a branch for each option
and copying the business logic from each endpoint to their respective branches.
Rubocop also moved some common lines out of the branches, but other than that
the logic is exactly same. There is no change in any functionality. We did not
try to converge to single logic and eliminate the branches in this commit. This
is intentional as I wanted to keep this commit as boring and add a separate
commit with the more interesting convergence stuff.

**Combine business logic for Vm resource creation endpoints**
Most of the code is straightforward changes to merge the logic in two branches
in the vm_helper. We had to rename the some fields in the UI to match to the
API.

**Combine business logic for Vm resource deletion endpoints**
For web, we used to return message in the response, however it is not used in
the frontend. We also returned 404 when the resource is not found, but that is
already handled as success case. This commit removes both thus makes overall
logic to more similar to API endpoints.

**Clean up Vm endpoint tests**
This is an opportunistic cleanup of some endpoint logic that were being tested
by both api and web tests. I think with a deeper inspection, we can clean the
tests up even more, but I only deleted most obvious duplicates for now.

### DIFF
--- a/routes/common/vm_helper.rb
+++ b/routes/common/vm_helper.rb
@@ -1,0 +1,138 @@
+# frozen_string_literal: true
+
+class Routes::Common::VmHelper < Routes::Common::Base
+  def list
+    if @mode == AppMode::API
+      dataset = project.vms_dataset
+      dataset = dataset.where(location: @location) if @location
+      result = dataset.authorized(@user.id, "Vm:view").paginated_result(
+        start_after: @request.params["start_after"],
+        page_size: @request.params["page_size"],
+        order_column: @request.params["order_column"]
+      )
+
+      {
+        items: Serializers::Vm.serialize(result[:records]),
+        count: result[:count]
+      }
+    else
+      vms = Serializers::Vm.serialize(project.vms_dataset.authorized(@user.id, "Vm:view").eager(:semaphores, :assigned_vm_address, :vm_storage_volumes).order(Sequel.desc(:created_at)).all, {include_path: true})
+      @app.instance_variable_set(:@vms, vms)
+
+      @app.view "vm/index"
+    end
+  end
+
+  def post(name)
+    Authorization.authorize(@user.id, "Vm:create", project.id)
+    fail Validation::ValidationFailed.new({billing_info: "Project doesn't have valid billing information"}) unless project.has_valid_payment_method?
+    if @mode == AppMode::API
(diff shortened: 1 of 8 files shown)


routes/common/vm_helper.rb
```diff
@@ -26,80 +26,56 @@ def list
...
-        name: name,
-        location: @location,
-        **request_body_params.except(*required_parameters).transform_keys(&:to_sym)
-      )
+    st = Prog::Vm::Nexus.assemble(
+      request_body_params["public_key"],
+      project.id,
+      name: name,
+      location: @location,
+      **request_body_params.except(*required_parameters).transform_keys(&:to_sym)
```
fdr, line 72: wew, these kinds of splats are quite dangerous. I realize there was pre-existence, so the onus is not on this refactoring, but on balance, I'd rather define and pack a struct and pass it, or something like that, given you already have a enumeration of field names.

fdr (approved):
nice that test coverage remains but execution time/number of reduces.

routes/common/vm_helper.rb
fdr: @geemus @enescakir got any thoughts or tips on this one?

byucesoy: This also caught my eye during refactoring. I left it as is because at line 33, we call `Validation.validate_request_body(params, required_parameters, allowed_optional_parameters)`, which ensures that we only accept predefined set of parameters.

Of course, future refactorings could endanger this. I think we can directly use `allowed_optional_parameters` to generate an approved list and use splat operator on that list.

enescakir (approved):
🔥

routes/common/vm_helper.rb
byucesoy: I'll fix this in a separate issue. It is also good opportunity to use new `security` tag.

PR #1874, opened 2024-08-21
Postgres US Prep to selectively pick the blob storage cluster

furkansahin:
## PostgresTimeline picks the blob store depending on the region
We are introducing a second blob store specifically for the US region
and we should be able to selectively use that. Therefore, we are making
changes to the PostgresTimeline to pick the blob store that is created
for that region. For that, we have to pass the location to the
PostgresTimelineNexus.assemble because there is simply no other way to
reach the location information of that resource. Furthermore, we are
removing the Config elements that are set to access to the MinIO cluster
to create new key pairs per bucket. Instead, we start to read those
values directly from the MinIO entity.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -128,8 +128,6 @@ module Config
   # Postgres
   optional :postgres_service_project_id, string
   override :postgres_service_hostname, "postgres.ubicloud.com", string
-  optional :postgres_service_blob_storage_access_key, string
-  optional :postgres_service_blob_storage_secret_key, string, clear: true
   optional :postgres_service_blob_storage_id, string
   override :postgres_monitor_database_url, Config.clover_database_url, string
   optional :postgres_monitor_database_root_certs, string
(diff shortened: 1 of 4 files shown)


byucesoy (commented):
How about instead of creating and managing extra database records, we get the responsible blob storage information by searching database, similar to what we do for DnsZone.

Something like (in `model/postgres/postgres_timeline.rb`)
```ruby
  def blob_storage
    @blob_storage ||= MinioCluster[project_id: Config.postgres_service_project_id, location: leader.resource.location]
  end
```

I see 3 benefits of this approach;
- No need to store credentials twice in the DB. (i.e. less data to maintain)
- No need for new model (i.e. less code to maintain)
- Re-uses a similar pattern (search like DnsZone) that we are using for both Postgres and Minio (i.e. less ways to do things)

furkansahin:
So, we need extra columns to store the credentials. We have an access_key, secret_key specifically created for postgres usage that has privileged access to manage bucket specific key pairs. We store them in the config at the moment, either we move them to minio table or we create a separate entity. The separate entity is mainly created for that purpose. So, the point 1 is invalid. 

Also, we can search for PgBlobStorageCredentials the same way you describe to keep using the same pattern? 

It's just that I don't want to add PG super user credentials to the minio_cluster table. I want to store them separately.

furkansahin:
We had an offline discussion and decided to use the MinIO entity directly since it already holds the `admin_user` and `admin_password`.

byucesoy (approved):

PR #1876, opened 2024-08-21
Generate parameters for Nexus::assemble from allowed_optional_parameters

byucesoy:
We used to directly use request_body_params with double splat operator to pass parameters to Nexus::assemble. This is not a good practice because it opens the door to pass any parameter to Nexus::assemble. With this change we start to generate the parameters for Nexus::assemble from allowed_optional_parameters.

### DIFF
--- a/routes/common/vm_helper.rb
+++ b/routes/common/vm_helper.rb
@@ -31,35 +31,36 @@ class Routes::Common::VmHelper < Routes::Common::Base
     required_parameters << "name" << "location" if @mode == AppMode::WEB
     allowed_optional_parameters = ["size", "storage_size", "unix_user", "boot_image", "enable_ip4", "private_subnet_id"]
     request_body_params = Validation.validate_request_body(params, required_parameters, allowed_optional_parameters)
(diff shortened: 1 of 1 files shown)


routes/common/vm_helper.rb
```diff
@@ -31,35 +31,36 @@ def post(name)
     required_parameters << "name" << "location" if @mode == AppMode::WEB
     allowed_optional_parameters = ["size", "storage_size", "unix_user", "boot_image", "enable_ip4", "private_subnet_id"]
     request_body_params = Validation.validate_request_body(params, required_parameters, allowed_optional_parameters)
+    assemble_params = allowed_optional_parameters.filter_map { |k| [k, request_body_params[k]] if request_body_params[k] }.to_h
```
fdr, line 34:
```suggestion
    assemble_params = request_body_params.slice(*allowed_optional_parameters).compact
```

I think that does the right thing/

fdr (approved):
much better, let's get this in pronto, then we can backburner thinking about if there's a good way to make this pattern a bit more clear, as I think this way of doing it is one think-o away from reopening a problem.

enescakir (approved):

PR #1877, opened 2024-08-21
Update billing rates for new VMs

byucesoy:
We reduced the rates for the storage as it is non-replicated and we increased the rates for compute. Overall, the VM prices are slightly reduced. Now, the standard-2 with 40GB storage costs $25.4 in Finland.

We also plan to combine storage and compute pricing in the future as a single rate, but it is a bigger change. I wanted to push this smaller change first.


fdr (commented):
sure

(but I am not an authority, mostly wanting to clear my code review query)

ozgune (approved):
Looks good to me

PR #1890, opened 2024-08-26
Calculate earliest_restore_time based on the backups in blob storage

byucesoy:
We used to cache the earliest_backup_completed_at in the database, with the
intention of updating it when we purge the backups. However, we decided to
purge backups with blob storage policy. Hence, there is no opportunity in the
code to update the earliest_backup_completed_at. This causes displaying wrong
potential restore times in the UI. To fix this, although it is slower, I'm
calculating the earliest_restore_time based on the backups in the blob storage.

### DIFF
--- a/model/postgres/postgres_timeline.rb
+++ b/model/postgres/postgres_timeline.rb
@@ -63,21 +63,11 @@ PGHOST=/var/run/postgresql
     backup.key.delete_prefix("basebackups_005/").delete_suffix("_backup_stop_sentinel.json")
   end
 
-  def refresh_earliest_backup_completion_time
-    update(earliest_backup_completed_at: backups.map(&:last_modified).min)
-    earliest_backup_completed_at
-  end
-
-  # The "earliest_backup_completed_at" column is used to cache the value,
-  # eliminating the need to query the blob storage every time. The
-  # "earliest_backup_completed_at" value can be changed when a new backup is
-  # created or an existing backup is deleted. It's nil when the server is
-  # created, so we get it from the blob storage until the first backup
-  # completed. Currently, we lack a backup cleanup feature. Once it is
-  # implemented, we can invoke the "refresh_earliest_backup_completion_time"
-  # method at the appropriate points.
+  # This method is called from serializer and needs to access our blob storage
+  # to calculate the answer, so it is inherently slow. It would be good if we
+  # can cache this somehow.
   def earliest_restore_time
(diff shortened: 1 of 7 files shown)


enescakir (approved):

model/postgres/postgres_timeline.rb
```diff
@@ -63,21 +63,11 @@ def latest_backup_label_before_target(target:)
...
-  # created, so we get it from the blob storage until the first backup
-  # completed. Currently, we lack a backup cleanup feature. Once it is
-  # implemented, we can invoke the "refresh_earliest_backup_completion_time"
-  # method at the appropriate points.
+  # This method is called from serializer and needs to access our blob storage
+  # to calculate the answer, so it is inherently slow. It would be good if we
+  # can cache this somehow.
   def earliest_restore_time
-    if (earliest_backup = earliest_backup_completed_at || refresh_earliest_backup_completion_time)
+    if (earliest_backup = backups.map(&:last_modified).min)
```
fdr, line 70: Any way to make backups an ordered list? IIRC the lexical scan order of the storage format ought to do that automagically.

fdr (approved):
I like the reduction in complexity here. I suppose it means the minio cluster has to be online for this to work, but really...if we can't do this, what's the odd we're not hair-on-fire from lack of archives anyway?

Maybe-opportunity for optimization, by reducing amount of paginated/processed data optional/can be follow-up

enescakir:
The only issue that slightly bothers me is that if MinIO goes down for some reason, the customer won't be able to access the PostgreSQL database details UI. We may catch the exception, and if Minio is inaccessible, we might hide the restore picker.

model/postgres/postgres_timeline.rb
byucesoy: They actually come sorted, but I'm not sure if ordering is guaranteed or a side effect, so I'm using `min`/`sort` where applicable.

byucesoy:
> The only issue that slightly bothers me is that if MinIO goes down for some reason, the customer won't be able to access the PostgreSQL database details UI. We may catch the exception, and if Minio is inaccessible, we might hide the restore picker.

Makes sense. Added a check.

fdr:
If it goes down...we can't archive, right?

byucesoy:
Yes, we can't. Though even in that case, we should be able to display database details in the UI. So I made a change that addresses Enes' comment.

PR #1894, opened 2024-08-27
Update reserve cache to return error if cache already exists

velioglu:
If the cache exists for the passed key and version, github itself returns "cache already exists" with a message including scope, key and version information. To mimic the behavior, returning and error including same information.

Note that, such a scenario might occur if explicit save action used on the workflow and that workflow is run multiple times.

### DIFF
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -78,6 +78,10 @@ class CloverRuntime
           fail CloverError.new(400, "InvalidRequest", "No workflow job data available")
         end
 
+        if repository.cache_entries_dataset[key: key, version: version, scope: scope]
+          fail CloverError.new(409, "AlreadyExists", "A cache entry for #{scope} scope already exists with #{key} key and #{version} version.")
+        end
+
         if size > GithubRepository::CACHE_SIZE_LIMIT
           fail CloverError.new(400, "InvalidRequest", "The cache size is over the 10GB limit")
         end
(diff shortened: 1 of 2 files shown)


routes/runtime/github.rb
```diff
@@ -78,6 +78,12 @@ class CloverRuntime
           fail CloverError.new(400, "InvalidRequest", "No workflow job data available")
         end
 
+        entry = repository.cache_entries_dataset
+          .exclude(committed_at: nil)
+          .where(key: key, version: version, scope: scope).all
+
+        fail CloverError.new(409, "AlreadyExists", "Scope: #{scope}, Key: #{key}, Version: #{version}") if entry.any?
```
enescakir, line 85: Since the current `ubicloud/cache` client also uses this API, I would prefer a more explanatory error message. Something like
```suggestion
        fail CloverError.new(409, "AlreadyExists", "A cache entry for #{scope} scope already exists with #{key} key and #{version} version.") if entry.any?
```

enescakir, line 85: Do we end up at `else` branch?

https://github.com/actions/toolkit/blob/6c4e082c181a51609197e536ef5255a0c9baeef7/packages/cache/src/cache.ts#L224C1-L238C1

velioglu: Yes

enescakir (approved):

PR #1899, opened 2024-08-29
Persist project invitation if account doesn't exists

enescakir:
### Persist project invitation if account doesn't exists

When a user is invited to a project, we associate them to it if they
already have an account. If they don't, we send an invitation email.
However, this approach often leads to a poor user experience, as the
admin must resend the invitation once the user creates an account. To
address this, I've implemented a fix that persists the invitation and
automatically associates the user to the project upon account creation.

I considered creating a temporary account in the existing accounts table
with a different status ID, but this would require extensive Rodauth
overrides. Therefore, I chose not to include non-existing users in the
accounts table, which serves as the backbone of our authentication
system.

Another alternative could involve not automatically associating the user
to the project upon account creation. Instead, users could explicitly
accept the invitation post-account creation. However, I found no
compelling reason to avoid automatic project-user association.

### Allow to remove user invitations
The admin might want to revoke the invitation if the user has not
created an account yet. This feature allows the admin to remove the
invitation.

### Expire project invitation after 7 days

Long-lasting project invitation links pose a security risk because email
address ownership can change over time. I've noticed other providers
expire these links after 7 days, which I believe is a good practice.
Therefore, I've implemented the same feature. This prog now expires
project invitation links every 7 days, with checks occurring every 6
hours.

### Add 50 pending invitations limit

Users can send invitations to any email address, which could potentially
allow a malicious actor to send a mass of invitations. To prevent this,
I added a limit of 50 pending invitations. This limit does not include
email addresses already registered in our system.

--------

_The last one is the user who was invited but doesn't have an account_

Screenshot 2024-08-29 at 10 05 25

### DIFF
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -156,7 +156,12 @@ class CloverWeb < Roda
       Validation.validate_account_name(account[:name])
     end
     after_create_account do
-      Account[account_id].create_project_with_default_policy("Default")
+      account = Account[account_id]
+      account.create_project_with_default_policy("Default")
+      ProjectInvitation.where(email: account.email).each do |inv|
+        account.associate_with_project(inv.project)
+        inv.destroy
+      end
     end
 
     reset_password_view { view "auth/reset_password", "Request Password" }
--- a/model/project_invitation.rb
+++ b/model/project_invitation.rb
@@ -1,0 +1,9 @@
+#  frozen_string_literal: true
+
+require_relative "../model"
+
+class ProjectInvitation < Sequel::Model
+  many_to_one :project
+end
+
+ProjectInvitation.unrestrict_primary_key
--- a/prog/expire_project_invitations.rb
+++ b/prog/expire_project_invitations.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class Prog::ExpireProjectInvitations < Prog::Base
+  label def wait
+    ProjectInvitation.where { _1.expires_at < Time.now }.all.each(&:destroy)
+
+    nap 6 * 60 * 60
+  end
+end
--- a/routes/web/project/user.rb
+++ b/routes/web/project/user.rb
@@ -19,6 +19,9 @@ class CloverWeb
       elsif ProjectInvitation[project_id: @project.id, email: email]
         flash["error"] = "'#{email}' already invited to join the project."
         r.redirect "#{@project.path}/user"
+      elsif @project.invitations_dataset.count >= 50
+        flash["error"] = "You can't have more than 50 pending invitations."
+        r.redirect "#{@project.path}/user"
       else
         @project.add_invitation(email: email, inviter_id: @current_user.id, expires_at: Time.now + 7 * 24 * 60 * 60)
       end
--- a/routes/web/project/user.rb
+++ b/routes/web/project/user.rb
@@ -6,6 +6,7 @@ class CloverWeb
 
     r.get true do
       @users = Serializers::Account.serialize(@project.accounts)
+      @invitations = Serializers::ProjectInvitation.serialize(@project.invitations)
 
       view "project/user"
     end
--- a/routes/web/project/user.rb
+++ b/routes/web/project/user.rb
@@ -12,10 +12,15 @@ class CloverWeb
 
     r.post true do
       email = r.params["email"]
-      # Don't invite deleted accounts
-      user = Account.exclude(status_id: 3)[email: email]
 
-      user&.associate_with_project(@project)
+      if (user = Account.exclude(status_id: 3)[email: email])
+        user.associate_with_project(@project)
+      elsif ProjectInvitation[project_id: @project.id, email: email]
+        flash["error"] = "'#{email}' already invited to join the project."
+        r.redirect "#{@project.path}/user"
+      else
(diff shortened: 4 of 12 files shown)


fdr (commented):
should there be a limit, too?

fdr:
> should there be a limit, too?

btw, the answer might be "not here, we're rife with needs for limits all over, there is probably at least some sharing/cohesion we want among those"

enescakir:
> should there be a limit, too?

What kind of limit are you thinking about? Are you referring to the number of users you can invite?

fdr:
> > should there be a limit, too?
> 
> What kind of limit are you thinking about? Are you referring to the number of users you can invite?

yeah, number of outstanding invitations, like, to prevent churning through 1,000,000 email addresses.

enescakir:
> > > should there be a limit, too?
> > 
> > 
> > What kind of limit are you thinking about? Are you referring to the number of users you can invite?
> 
> yeah, number of outstanding invitations, like, to prevent churning through 1,000,000 email addresses.

Makes sense. I added it.

routes/web/project/user.rb
```diff
@@ -6,16 +6,25 @@ class CloverWeb
...
-      # Don't invite deleted accounts
-      user = Account.exclude(status_id: 3)[email: email]
 
-      user&.associate_with_project(@project)
+      if (user = Account.exclude(status_id: 3)[email: email])
+        user.associate_with_project(@project)
+      elsif ProjectInvitation[project_id: @project.id, email: email]
+        flash["error"] = "'#{email}' already invited to join the project."
+        r.redirect "#{@project.path}/user"
+      elsif @project.invitations.count >= 50
```
fdr, line 22:
```suggestion
      elsif @project.invitations_dataset.count >= 50
```

Perhaps? No need to load the tuples.

enescakir: Makes sense. Done. I was expecting this comment from you 😄

prog/expire_project_invitations.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Prog::ExpireProjectInvitations < Prog::Base
+  label def wait
+    ProjectInvitation.all.select { _1.expired_at < Time.now }.each(&:destroy)
```
byucesoy, line 5: `.all.select` first fetches all records and then selects expired ones. Instead, we can do `ProjectInvitation.where { expired_at < Time.now }`, which would push filtering down to DB.

views/project/user.erb
```diff
@@ -83,7 +83,7 @@
           </tr>
         <% end %>
         <% @invitations.each do |invitation| %>
-          <tr id="invitation-<%= invitation[:email] %>" class="whitespace-nowrap text-sm font-medium">
+          <tr id="invitation-<%= invitation[:email].gsub(/\W+/, "") %>" class="whitespace-nowrap text-sm font-medium">
```
byucesoy, line 86: What is the purpose of `gsub` here?

model/project_invitation.rb
```diff
@@ -4,6 +4,10 @@
 
 class ProjectInvitation < Sequel::Model
   many_to_one :project
+
+  def expired_at
```
byucesoy, line 8: nit: Maybe rename this as `expires_at`, because for most active records expiration date will be in the future (thus it is not expire**d** yet.

clover_web.rb
```diff
@@ -156,7 +156,12 @@ def assets
       Validation.validate_account_name(account[:name])
     end
     after_create_account do
-      Account[account_id].create_project_with_default_policy("Default")
+      account = Account[account_id]
+      account.create_project_with_default_policy("Default")
+      ProjectInvitation.where(email: account.email).each do |inv|
+        account.associate_with_project(inv.project)
```
byucesoy, line 162: I'm not sure about auto-joining to invited projects. What if I'm invited to project that I don't want to join? Signing up would make me join those projects. We also don't have a way to exit from a project.

That said, I'm also not entirely sure about what should be the ideal experience here. Do you know how does other providers handle this case?

views/project/user.erb
enescakir: When we use `@` or `.` in the `id` attribute, it conflicts with CSS selectors.

prog/expire_project_invitations.rb
enescakir: We don't have `expired_at` column at DB. We have only `created_at`. `expired_at` is calculated dynamically. I can write a query like `created_at + 7 days` on the DB side.

clover_web.rb
enescakir: We checked a few other providers with Burak offline. It looks like they use auto-join. So we decided to keep it that way for now.

model/project_invitation.rb
enescakir: Makes sense

prog/expire_project_invitations.rb
enescakir: I decided to keep `expires_at` as column, instead of `created_at` to simplify it

enescakir: Changed it

byucesoy (approved):

PR #1909, opened 2024-09-03
Load balancer bug fixes

furkansahin:
## Update health probes to use the hostname and check return type
We are updating the health probes to use the hostname to perform the
health checks. This is important for HTTPS health checks because it is
possible the customers will only accept that virtual hosting endpoint,
instead of raw IP.

Since we are in the network namespace and do not need to perform
a lookup, we resolve the hostname automatically by using the --resolve
option of curl.

We also make a small fix here to transform response_code to an integer
and perform the check. This is necessary in case the response_code has
spaces or \n characters at the start or end of the response code.

## Fix VM removal when its attached to a load balancer
We need to remove the vm user only after the cert server is removed
because it is used by cert server and deluser command fails otherwise.

The other case is that, we might start deleting VM and then the load
balancer is deleted before VM deletion completes. In that case, we
should perform a check at the last state to verify the load balancer is
still intact before removing VM from the load balancer.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -437,7 +437,7 @@ class Prog::Vm::Nexus < Prog::Base
       nap 30
     end
 
-    vm.load_balancer.remove_vm(vm)
+    vm.load_balancer.remove_vm(vm) if vm.load_balancer
 
     vm.vm_host.sshable.cmd("sudo host/bin/setup-vm delete_net #{q_vm}")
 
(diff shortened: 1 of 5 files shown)


fdr (approved):
the missing jargon you are looking for in your commit message, re: hostnames is "virtual hosting." It took me a while to figure out what's what you were talking about, but I think that's what you mean? (we [I think reasonably] ignore the TLS certificate for health checking, and thus its hostname validation, though there are some pros and cons)

At some point...probably want to stop forking curl and instead do a ruby HTTP request via the monitor. Not sure when though.

enescakir:
The first commit message has some typo and duplicate paragraph

image

PR #1917, opened 2024-09-06
Update Ubicloud banner in README.md

ozgune:
Update README.md to display the latest Ubicloud banner


enescakir (approved):

fdr (approved):
huh okay seems weird, somehow in this format it comes off as too AWS-indexed, though verbally I think AWS comes across as an IaaS-in-general metonym, but whatever I guess?

PR #1920, opened 2024-09-09
Allow to pin the runners of a project to specific location

enescakir:
We use all locations for GitHub runners to increase our overall utilization. However, some customers experience job failures in certain locations and need to be pin them to a specific one. While we don't have a blanket policy to assign all customers to a specific location, we have decided to accommodate this if a customer has a specific requirement.

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -31,11 +31,12 @@ class Prog::Vm::GithubRunner < Prog::Base
   end
 
   def pick_vm
+    location = project.get_ff_runner_preferred_location || label_data["location"]
     skip_sync = true
     pool = VmPool.where(
       vm_size: label_data["vm_size"],
       boot_image: label_data["boot_image"],
-      location: label_data["location"],
+      location: location,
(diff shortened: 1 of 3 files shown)


prog/vm/github_runner.rb
```diff
@@ -31,11 +31,12 @@ def self.assemble(installation, repository_name:, label:, default_branch: nil)
   end
 
   def pick_vm
+    location = project.get_ff_runner_preferred_location || label_data["location"]
     skip_sync = true
     pool = VmPool.where(
       vm_size: label_data["vm_size"],
       boot_image: label_data["boot_image"],
-      location: label_data["location"],
+      location: location,
```
bsatzger, line 39: if we do this for more customers it might become a problem that we have to pin to `hetzner-fsn1` even though we probably would want to pin to 'github-runners' in fns1? maybe not now but in the future we could introduce an additional filter for data_center and keep 'github-runners' as location for this case but with an additional 'FSN1-%' filter for data center.

enescakir:
The job failures were due to intermittent network issues, which have now been resolved. We no longer need this feature.

PR #1923, opened 2024-09-10
Configure OpenSSH settings according to Mozilla guidelines

byucesoy:
Mozilla has a set of guidelines for secure SSH configuration, which can be found at https://infosec.mozilla.org/guidelines/openssh.

This commit creates a config file for OpenSSH that is based on the Mozilla's guidelines for secure SSH configuration. This configuration;
- Sets the preference order of HostKey algorithms
- Sets the allowed KexAlgorithms, Ciphers and MAC algorithms
- Disables the use of password authentication
- Increases logging level to VERBOSE

Final one also ensures that we are logging which SSH key is used to login to the server. This is useful for auditing purposes and SOC2.

Mozilla also recommends disabling root login, but this is not done in this commit, because implementing that requires a bit more work and some change in how we set the hosts up. I wanted to push this commit out quickly.

### DIFF
--- a/rhizome/host/bin/prep_host.rb
+++ b/rhizome/host/bin/prep_host.rb
@@ -99,3 +99,23 @@ CRON
 
 r "systemctl enable cron"
 r "systemctl start cron"
+
+# Taken from https://infosec.mozilla.org/guidelines/openssh
+safe_write_to_file("/etc/ssh/ssh_config.d/10-clover.conf", <<~SSHD_CONFIG)
+# Supported HostKey algorithms by order of preference.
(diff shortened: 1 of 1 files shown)


fdr (approved):
nice. maybe the next step can be to bring this into the services as well, e.g. in the `common` rhizome part, we pretty much want it everywhere

furkansahin (approved):

enescakir (approved):
Are you planning to apply this configuration to the existing hosts?

byucesoy:
> Are you planning to apply this configuration to the existing hosts?

Yes, I am.

PR #1937, opened 2024-09-13
Revert "Revert "Load balancer bug fixes""

furkansahin:
## Revert "Revert "Update health probes to use the hostname and check return type""
This reverts commit 3bf4312908631f68a19fa180a7d25d33fc6f1747.

## Fix "Fix VM removal when its attached to a load balancer""
This fixes commit bb3a0ec68544207e2d178bd85b79fc6b5d940b7e.

In the very first commit 4a00e905, we have introduced a change that
would remove the user home directory before the hugepages and spdk are
cleaned up. Because of active processes, that operation would not
complete and we had the stale vm directories backed up in the hosts.

We then reverted that commit. However, that meant the VM destroy
operation for a load balancer related VM was again, broken. Because,
during VM destroy that is part of a load balancer, we want to remove all
the resources but keep the network namespace and the VM user active so
that we can gracefully remove the node from the load balancer. However,
for regular VM removals, we were first calling remove networking + the
rest of the resources, AND the networking removal was trying to remove
the vm home directory. Since we had the SPDK resources and hugepages
setup, they would not be removed.

In this commit, we move the user removal to its own method, make sure it
is called as the last item for;
1. Regular VM removals from rhizome/host/bin/setup-vm.delete
(vm_setup.purge)
2. LB VM removal from rhizome/host/bin/setup-vm.delete_net
(vm_setup.purge_user)

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -437,7 +437,7 @@ class Prog::Vm::Nexus < Prog::Base
       nap 30
     end
 
-    vm.load_balancer.remove_vm(vm)
+    vm.load_balancer.remove_vm(vm) if vm.load_balancer
 
     vm.vm_host.sshable.cmd("sudo host/bin/setup-vm delete_net #{q_vm}")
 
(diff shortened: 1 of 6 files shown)


fdr (approved):
nice find.

bsatzger (approved):

PR #1943, opened 2024-09-17
Fix Almalinux nftables usage

furkansahin:
Commit 7303ec6240b8e62b5d2b25316ced5def4dff0057 made a change to how we boot VMs by adding new commands to the cloud-init. However, we realized that almalinux yaml files on cloud-init fail with ipv6 subnet. YAML files are sensitive to ":". Fixing that with escaping surfaced another issue. Almalinux doesn't come with nftables installed by default. Therefore, we are installing it at the time of provisioning.

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -43,8 +43,8 @@ class VmSetup
     @vp ||= VmPath.new(@vm_name)
   end
 
-  def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices)
-    cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes)
+  def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices, boot_image)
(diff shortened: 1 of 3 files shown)


rhizome/host/lib/vm_setup.rb
```diff
@@ -473,6 +474,17 @@ def generate_swap_config(swap_size_bytes)
   end
 
   def write_user_data(unix_user, public_keys, swap_size_bytes)
+    params = JSON.parse(File.read(vp.prep_json))
+    runcmd_nft_add = if params["boot_image"].include?("almalinux")
+<<NFT_SETUP
+  - [dnf, install, '-y', nftables]
```
enescakir, line 480: I feel like `bootcmd` runs in very early stage, and runs before `runcmd`. How `nft` commands work in `bootcmd`?

https://www.linode.com/docs/guides/run-shell-commands-with-cloud-init/#run-commands-with-bootcmd-directive

furkansahin: IF nftables already installed, they run properly in bootcmd.

bsatzger (approved):

rhizome/host/lib/vm_setup.rb
```diff
@@ -482,15 +488,17 @@ def write_user_data(unix_user, public_keys, swap_size_bytes)
     ssh_authorized_keys:
 #{public_keys.map { "      - #{yq(_1)}" }.join("\n")}
 
+packages:
+  - nftables
```
enescakir, line 492: https://cloudinit.readthedocs.io/en/latest/reference/examples.html#install-arbitrary-packages
> # if packages are specified, then package_update will be set to true
So it will increase provisioning times for runners 🙃

PR #1951, opened 2024-09-19
Fix GitHub installation configuration URL

enescakir:
We direct users to the installation configuration page on GitHub's UI when they want to add more repositories and so on. However, I discovered yesterday that non-admin users of an organization are led to a 404 page. I found a more generic URL that works for everyone. GitHub has decided to redirect users to the appropriate page.

Since we don't have separate branches for different installation types, I've moved it to the serializer as it's only used once.

### DIFF
--- a/model/github_installation.rb
+++ b/model/github_installation.rb
@@ -9,13 +9,6 @@ class GithubInstallation < Sequel::Model
 
   include ResourceMethods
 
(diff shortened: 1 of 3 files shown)


fdr:
Less code :eyes:

velioglu (approved):

PR #1953, opened 2024-09-19
Remove downloading cache proxy code

velioglu:
Since proxy code will be working via a service installed on the runner image, don't need to download and run the code directly anymore.

Updating CUSTOM_ACTIONS_CACHE_URL and manual resolvement of localhost.blob.core.windows.net are still required. First is to send cache api requests to local proxy and the second to redirect blob request to local proxy as well.

Once we'lll make sure to use transparent cache for all users, that part will be moved out of project flag check later.

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -222,87 +222,15 @@ class Prog::Vm::GithubRunner < Prog::Base
       UBICLOUD_CACHE_URL=#{Config.base_url}/runtime/github/" | sudo tee -a /etc/environment
     COMMAND
 
-    # Remove comments and empty lines before sending them to the machine
-    vm.sshable.cmd(command.gsub(/^(\s*# .*)?\n/, ""))
-
     if github_runner.installation.project.get_ff_transparent_cache
-      hop_setup_forked_runner
+      local_ip = vm.nics.first.private_ipv4.network.to_s
+      command += <<~COMMAND
+        echo "CUSTOM_ACTIONS_CACHE_URL=http://#{local_ip}:51123/random_token/" | sudo tee -a /etc/environment
+      COMMAND
     end
 
-    hop_register_runner
-  end
-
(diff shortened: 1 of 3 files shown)


prog/vm/github_runner.rb
```diff
@@ -234,60 +234,12 @@ def setup_info
...
-
-      jq '. += [#{setup_info.to_json}]' /imagegeneration/imagedata.json > ./actions-runner/.setup_info
-
-      echo "PATH=$PATH" >> ./actions-runner/.env
-
-      sudo rm -rf /home/runner/actions-runner
-      sudo mv ./actions-runner /home/runner/
-      sudo chown -R runner:runner /home/runner/actions-runner
-
       echo "CUSTOM_ACTIONS_CACHE_URL=http://localhost:51123/random_token/" | sudo tee -a /etc/environment
```
enescakir, line 237: Since the remaining part is too small, IMHO, you can append it to `command` in `setup_environment`. Something like that

```ruby
    if github_runner.installation.project.get_ff_transparent_cache
      command += <<~COMMAND
        echo "CUSTOM_ACTIONS_CACHE_URL=http://localhost:51123/random_token/" | sudo tee -a /etc/environment
        echo "127.0.0.1 localhost.blob.core.windows.net" | sudo tee -a /etc/hosts
      COMMAND
    end

```

spec/prog/vm/github_runner_spec.rb
```diff
@@ -398,7 +398,7 @@
       expect { nx.setup_environment }.to hop("register_runner")
     end
 
-    it "hops to setup_forked_runner" do
+    it "hops to register_runner with flag" do
```
enescakir, line 401:
```suggestion
    it "hops to register_runner after enabling transparent cache" do
```

enescakir (approved):

enescakir (approved):

PR #1954, opened 2024-09-19
Add option to bypass runner deregistration and destroy vm immediately

enescakir:
When we attempt to destroy the runner, we also deregister it from GitHub.  We wait to receive a 'not found' response for the runner. If the runner is still running a job and, due to stale data, it gets mistakenly hopped to destroy, this prevents the underlying VM from being destroyed and the job from failing. However, in certain situations like fraudulent activity, we might need to bypass this verification and immediately remove the runner.

First I named the semaphore as `force_destroy`, but then I renamed it as `skip_deregistration`.  The former name was misleading, because the operator should increment the `destroy` semaphore too.

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -12,7 +12,7 @@ class GithubRunner < Sequel::Model
   include ResourceMethods
   include SemaphoreMethods
   include HealthMonitorMethods
-  semaphore :destroy
+  semaphore :destroy, :skip_deregistration
 
   def run_url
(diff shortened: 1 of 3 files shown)


fdr (approved):
seems well-explained to me. Unfortunate that we have no block-related recipe  to do the `unless` case (only the `when_... do` case) but, tbf, we never actually used those blocks for the intended purpose, which was to annotate the transitions in telemetry in relation to a semaphore being set, so as-is, all our blocks doing conditional branching are more idiosyncratic bloat than delivering value.

velioglu (approved):

PR #1959, opened 2024-09-23
Add password requirements for user accounts

byucesoy:
Previously, we used the default password requirements that comes with rodauth, which are 6 characters minimum and no requirement on the type of characters.

The defaults are too permissive. With this commit we are increasing minimum length to 8 and also require passwords to contain at least one lowercase, one uppercase and one digit.

Finally, we also set a maximum length as well, this closes a low risk attack vector (DoS with long password). The length 72 is used becase it seems bcrypt only uses first 72 bytes of password while creating the hash. Source: https://github.com/jeremyevans/rodauth/blob/master/doc/login_password_requirements_base.rdoc

### DIFF
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -328,6 +328,17 @@ class CloverWeb < Roda
     auto_add_recovery_codes? true
     auto_remove_recovery_codes? true
     recovery_auth_view { view "auth/recovery_auth", "Recovery Codes" }
+
+    # Password Requirements
+    password_minimum_length 8
+    password_maximum_bytes 72
+    password_meets_requirements? do |password|
+      password.match?(/[a-z]/) && password.match?(/[A-Z]/) && password.match?(/[0-9]/)
+    end
+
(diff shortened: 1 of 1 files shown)


enescakir (approved):

clover_web.rb
```diff
@@ -328,6 +328,17 @@ def assets
     auto_add_recovery_codes? true
     auto_remove_recovery_codes? true
     recovery_auth_view { view "auth/recovery_auth", "Recovery Codes" }
+
+    # Password Requirements
+    password_minimum_length 8
+    password_maximum_bytes 72
+    password_meets_requirements? do |password|
+      return password.match?(/[a-z]/) && password.match?(/[A-Z]/) && password.match?(/[0-9]/)
```
fdr, line 336: do you really mean `return` here?

byucesoy: you are right, it is not necessary. removed it.

PR #1960, opened 2024-09-23
Start using dnsmasq as the dns server

furkansahin:
So far we have not been using dnsmasq for DNS queries, we just used it to configure public dns server addresses in the VM. Therefore, all the dns queries hit to a public dns server after the VM, for sure. Now, we configure dnsmasq to also run as a dns server. This way, if the address is cached locally in the dnsmasq, we skip going to a public server and respond locally from the host. To be able to do that, we assign a private ipv6 address (fd00:b1c:100d:53:: so, ubicloud-53) to the dnsmasq server, make it listen there and as part of the DHCP requests, we send the dnsmasq's address to the client as the dns server.

While doing above, I realized dnsmasq listens in all the available address space when we configure listen-address together with interfaces. For it to specifically bind to the addresses and interfaces, we are also adding "bind-interfaces".

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -280,6 +280,7 @@ add element inet drop_unused_ip_packets allowed_ipv4_addresses { #{ip_net} }
     end
 
     r "ip -n #{q_vm} addr replace fd00:0b1c:100d:5AFE:CE::/56 dev #{nics.first.tap}"
+    r "ip -n #{q_vm} addr replace fd00:0b1c:100d:53::/48 dev #{nics.first.tap}"
   end
 
   def parse_routes(routes)
@@ -421,18 +422,28 @@ DHCP
     end.join("\n")
 
     raparams = nics.map { "ra-param=#{_1.tap}" }.join("\n")
-
+    interfaces = nics.map { "interface=#{_1.tap}" }.join("\n")
+    dnsmasq_address_ip6 = NetAddr::IPv6.parse("fd00:0b1c:100d:53::")
     vp.write_dnsmasq_conf(<<DNSMASQ_CONF)
 pid-file=
 leasefile-ro
 enable-ra
 dhcp-authoritative
+domain-needed
+bogus-priv
+no-resolv
 #{raparams}
+#{interfaces}
 dhcp-range=#{guest_network.nth(2)},#{guest_network.nth(2)},#{guest_network.netmask.prefix_len}
 #{private_ip_dhcp}
-dhcp-option=option6:dns-server,2620:fe::fe,2620:fe::9
-dhcp-option=option:dns-server,149.112.112.112,9.9.9.9
+server=149.112.112.112
+server=9.9.9.9
+server=2620:fe::fe
+server=2620:fe::9
+dhcp-option=option6:dns-server,#{dnsmasq_address_ip6}
+listen-address=#{dnsmasq_address_ip6}
 dhcp-option=26,1400
+bind-interfaces
 DNSMASQ_CONF
 
     ethernets = nics.map do |nic|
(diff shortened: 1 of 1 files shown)


rhizome/host/lib/vm_setup.rb
```diff
@@ -280,6 +280,7 @@ def setup_taps_6(gua, nics)
     end
 
     r "ip -n #{q_vm} addr replace fd00:0b1c:100d:5AFE:CE::/56 dev #{nics.first.tap}"
+    r "ip -n #{q_vm} addr replace fd00:0b1c:100d:53::/48 dev #{nics.first.tap}"
```
fdr, line 283: I know it's kinda everywhere right now, but this whole `nics.first` business kinda worries me, I guess you're thinking it's just a dragon we'll have to slay later, in how to deal with multiple nics?

rhizome/host/lib/vm_setup.rb
```diff
@@ -421,18 +422,28 @@ def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image)
     end.join("\n")
 
     raparams = nics.map { "ra-param=#{_1.tap}" }.join("\n")
-
+    interfaces = nics.map { "interface=#{_1.tap}" }.join("\n")
+    dnsmasq_address_ip6 = NetAddr::IPv6.parse("fd00:0b1c:100d:53::")
```
fdr, line 426: I like how this looks, but wonder: is there any other use of all the bits in `fd00:0b1c:100d:53` that's non-zero?

I think we should probably keep it as-is, it's rather easy to read, and I can see the obvious problem with `::53` (in that the special base subnet of `fd00:0b1c:100d` gets crowded and services have no wiggle room), but, dare to dream I guess?

fdr (approved):
Might be pretty weird if we start doing something more advanced with DNS (e.g. internal service endpoint type stuff) and the answer you get depends on your IP stack, but for now, where at best we cache & test, seems tolerable.

rhizome/host/lib/vm_setup.rb
```diff
@@ -421,18 +422,28 @@ def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image)
...
-dhcp-option=option6:dns-server,2620:fe::fe,2620:fe::9
-dhcp-option=option:dns-server,149.112.112.112,9.9.9.9
+server=149.112.112.112
+server=9.9.9.9
+server=2620:fe::fe
+server=2620:fe::9
+dhcp-option=option6:dns-server,#{dnsmasq_address_ip6}
+listen-address=#{dnsmasq_address_ip6}
 dhcp-option=26,1400
+bind-interfaces
```
fdr, line 446: I do find the number of options being shuffled around here a bit confusing, it seems like you are making dnsmasq more selective from the commit message though, right?  What's dnsmasq's design theory here?

rhizome/host/lib/vm_setup.rb
furkansahin: For now, nics is a set with single element, the moment we introduce multiple nics, that will be a dragon to slay.

rhizome/host/lib/vm_setup.rb
furkansahin: Maybe later we will find another case but for now, I think this is good enough.

rhizome/host/lib/vm_setup.rb
furkansahin: Yes, so, the first 3 options are to handle dns lookups in a safe way, not forwarding dns lookups without dots, not leaking private ip addresses and not using resolve.conf for dns addresses. (domain-needed, bogus-priv, no-resolv)

`interface=`, `listen-address` and `bind-interfaces` part is where it gets interesting. We tell dnsmasq to listen in which interface by `interface` and which address by `listen-address`. However, it still binds itself to all of the interfaces and available addresses but filters the queries. Specifying `bind-interfaces` enforces dnsmasq to bind to the specified interface and address, eliminating the need to filter the queries.

PR #1965, opened 2024-09-24
Add common extensions to shared_preload_libraries

byucesoy:
Updating shared_preload_libraries requires server restart, which is doable but we can eliminate the extra work for some of the most common extensions such as pg_cron and pg_stat_statements by adding them to shared_preload_libraries by default.

pg_stat_statements allocate small amount of memory by default, which is not a big concern for regular servers. When we introduce smaller SKUs, we probably wouldn't add it to shared_preload_libraries for small SKUs by default.


fdr (approved):
they both seem unoffensive in terms of wide application to off-set their distributed costs.

PR #1966, opened 2024-09-24
ParadeDB Backend

byucesoy:
**Add migration file for flavor column in PostgresResource**

**Add ParadeDB as new PostgreSQL flavor**
As part of our partnership with ParadeDB, we are adding a new deployment option
for PostgreSQL.

**Pick base image based on the PostgreSQL flavor**
Currently we have 2 flavors. Their primary difference is the base image we use
for them. This commit adds logic to pick base image based on the PostgreSQL
database flavor.

**Send mail to partners when new database is created**
As our partnership discussion, partners will be responsible for the customer
journey and success. For effective communication, partners would need contact
information of the customers. This commit adds logic for sending e-mail to the
partners when a database managed by them is created.

In the UI, we will ask consent for sharing contact information.


routes/common/postgres_helper.rb
```diff
@@ -41,6 +41,8 @@ def post(name: nil)
     else 0
     end
 
+    flavor = request_body_params["flavor"] || PostgresResource::Flavor::POSTGRES
```
enescakir, line 44: Should we validate this, or does the database-level enum already handle it?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -18,6 +18,11 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
       ubid = PostgresServer.generate_ubid
 
       postgres_resource = PostgresResource[resource_id]
+      boot_image = case postgres_resource.flavor
+      when PostgresResource::Flavor::PARADEDB then "postgres-paradedb-ubuntu-2204"
+      else "postgres-ubuntu-2204"
```
enescakir, line 23: We could add `PostgresResource::Flavor::POSTGRES` as a case, and fails in the `else` branch. This approach helps guard against unwanted flavors, especially when we deprecate some. Otherwise, it will silently use the default one.

routes/common/postgres_helper.rb
```diff
@@ -253,4 +255,14 @@ def failover
 
     Serializers::Postgres.serialize(@resource, {detailed: true})
   end
+
+  def send_notification_mail_to_partners(resource, user_email)
+    if resource.flavor == PostgresResource::Flavor::PARADEDB
+      Util.send_email(Config.postgres_paradedb_notification_email, "New ParadeDB Postgres database has been created.",
+        greeting: "Hello ParadeDB team,",
+        body: ["New ParadeDB Postgres database has been created.",
```
enescakir, line 263: Should we include additional details such as VM size, storage, and HA, or should we share them in private?

enescakir (approved):

routes/common/postgres_helper.rb
byucesoy: Database level enum already handles that. I initially thought intercepting this here to provide a better error message but there is also parallel work going on API level type checking, which I assume handle this as well.

prog/postgres/postgres_server_nexus.rb
byucesoy: Makes sense, added additional check

routes/common/postgres_helper.rb
byucesoy: Makes sense, added additional details

routes/common/postgres_helper.rb
```diff
@@ -249,4 +255,20 @@ def failover
...
+          "Id: #{resource.id}",
```
enescakir, line 264: We don't show UUID anywhere. We can share UBID. It's more common in our team.

model/postgres/postgres_resource.rb
```diff
@@ -104,6 +104,11 @@ module HaType
     SYNC = "sync"
   end
 
+  module Flavor
+    POSTGRES = "standard"
+    PARADEDB = "paradedb"
```
fdr, line 109: Do you want to do this enumeration as a string, or something that is truly closest to a namespaced, singleton symbol, which the closest might be:

```ruby
[1] clover-development(main)> module Flavor
[1] clover-development(main)*   class ParadeDB; end;
[1] clover-development(main)* end;  
[2] clover-development(main)> Flavor::ParadeDB
=> Flavor::ParadeDB
```

routes/common/postgres_helper.rb
fdr: I agree, let's stick with ubid. I suspect email notification will be replaced with an API eventually.

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -154,12 +154,18 @@ def before_run
       billing_record_parts.push({resource_type: index.zero? ? "PostgresStorage" : "PostgresStandbyStorage", amount: postgres_resource.target_storage_size_gib})
     end
 
+    billing_rate_family = case postgres_resource.flavor
+    when PostgresResource::Flavor::POSTGRES then "standard"
+    when PostgresResource::Flavor::PARADEDB then "paradedb"
+    else raise "Unknown PostgreSQL flavor: #{postgres_resource.flavor}"
```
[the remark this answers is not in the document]
byucesoy: I tried this and unless I misunderstood your suggestion here is what I come up with;

```ruby
  module Flavor
    def self.from_string(flavor)
      case flavor
      when "standard"
        Standard
      when "paradedb"
        ParadeDB
      else
        raise ArgumentError, "Unknown flavor: #{flavor}"
      end
    end

    class Standard
      def self.boot_image
        "postgres-ubuntu-2204"
      end
    end

    class ParadeDB
      def self.boot_image
        "postgres-paradedb-ubuntu-2204"
      end
    end
  end
```
Then we can use this to pick correct boot image;

```ruby
boot_image = PostgresResource::Flavor.from_string(postgres_resource.flavor).boot_image
```

Personally, I found using the `case`/`when` directly to pick boot image shorter and more readable. Though it is possible that I misunderstood your suggestion. Did you had something else in your mind?

model/postgres/postgres_resource.rb
byucesoy: Keeping them string allows direct comparison with the value in the database. We can make them class but then we need to compare them to hardcoded strings somewhere and convert to class.

prog/postgres/postgres_resource_nexus.rb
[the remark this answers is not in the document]
fdr: So the general idea, which can be done a number of ways, is to make sure that the "case" statements are exhaustive and avoid the `else raise` boilerplate repeated, instead doing it but once.

The key in Arch.render is that keyword arguments take the place of conditionals:

```
  def render(x64: sym, arm64: sym)
    if x64?
      x64
    elsif arm64?
      arm64
    else
      fail "BUG: could not detect architecture"
    end.to_s
  end
```

That means that the callers have to exhaustively specify renderings for each keyword argument. the `else` case gets encapsulated.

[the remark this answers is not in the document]
byucesoy: Currently we only have one `case` block (we used to have two, but I managed to eliminate one of them). If we need to put more `case` blocks (which we might need, because different partners might come with different requirements), then centralizing them into one place makes more sense to me. For now I am inclined to keep this as-is unless you think we should centralize them now.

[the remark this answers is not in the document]
fdr: as-is is fine.

routes/common/postgres_helper.rb
enescakir: It has thumbs up emoji and resolved but it's still `resource.id`

enescakir (approved):

routes/common/postgres_helper.rb
byucesoy: Weird, I guess the change got lost while rebasing.

PR #1967, opened 2024-09-24
ParadeDB UI

enescakir:
### **Add ParadeDB flavor to the PostgreSQL create page**
  As part of our partnership with ParadeDB, we're introducing ParadeDB as
  a new flavor for our PostgreSQL database.
  
  The commit adds ParadeDB logo, along with a partnership notice, to
  enhance the visibility of our partnership to the create page.
  
  The customer have to accept the partnership details before creating a
  new database.
  
Screenshot 2024-09-24 at 18 25 11

Screenshot 2024-09-27 at 19 25 40

### **Pass ParadeDB billing rates to the create page**
  There were a few places on the creation page where the resource family
  field was hardcoded as `standard`. I've made these dynamic and added new
  `Option::PostgresSizes` options for ParadeDB.

### **Add ParadeDB details to the detail page**
  If it's not a partnership database, it doesn't show any details.
  
Screenshot 2024-09-27 at 19 26 08

### **Show PostgreSQL flavors as big buttons if there is no database already**
  If there is no database, we show possible PostgreSQL flavors as big
  create buttons. It helps to show available flavors to the customer
  clearly.
  
Screenshot 2024-09-27 at 19 26 50

### **Replace PostgreSQL create button with a dropdown**
  Since we have more than one flavor of PostgreSQL, we change the create
  button with a dropdown that allows the customer to select the PostgreSQL
  flavor they want to create.
  
Screenshot 2024-09-27 at 19 27 11

### **Add PostgreSQL flavor column to the resource table on UI**
  
Screenshot 2024-09-24 at 18 26 14


views/components/button.erb
```diff
@@ -3,6 +3,7 @@
 <% icon = (defined?(icon) && icon) ? icon : nil %>
 <% extra_class = defined?(extra_class) ? extra_class : nil %>
 <% attributes = (defined?(attributes) && attributes) ? attributes : {} %>
+<% rigth_icon = (defined?(rigth_icon) && rigth_icon) ? rigth_icon : nil %>
```
byucesoy, line 6: typo: right

byucesoy:
I think we need to pass flavor to backend (as a hidden input). We use it in the UI as a URL query parameter but we don't send it to backend.

views/postgres/create.erb
```diff
@@ -189,6 +200,28 @@
...
+                      This database managed by ParadeDB team. It's powered up with ParadeDB extension.<br>
+                      You can join ParadeDB's community on <a href="https://paradedb.com/community" target="_blank" class="text-orange-600 font-semibold">Slack</a> or ask for help at <a href="mailto:support@paradedb.com" class="text-orange-600 font-semibold">support@paradedb.com</a>.<br>
+                      We will share your details with ParadeDB team. You have to accept ParadeDB's <a href="https://paradedb.com/community" target="_blank" class="text-orange-600 font-semibold">Terms of Service</a> and <a href="https://paradedb.com/community" target="_blank" class="text-orange-600 font-semibold">Privacy Policy</a> to be able to create this database.
+                    </p>
+                    <%== render(
+                      "components/form/checkbox",
+                      locals: {
+                        name: "accept_partnership",
+                        options: [
+                          ["1", "Accept Terms of Service", {}, {required: true}],
```
byucesoy, line 217: It would be good to link to related policies here. Also I think we can combine them into one checkbox.

enescakir: Screenshot 2024-09-24 at 21 01 22

views/postgres/index.erb
```diff
@@ -59,6 +60,13 @@
                   <%= pg[:name] %>
                 </a>
               </td>
+              <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
+                <% (partner, _, _) = flavors[pg[:flavor]] %>
```
byucesoy, line 64: Maybe use some other name for this variable instead of partner, as Ubicloud is not technically a partner, but it is in this array.

byucesoy (approved):

routes/common/postgres_helper.rb
```diff
@@ -31,6 +31,7 @@ def post(name: nil)
     required_parameters = ["size"]
     required_parameters << "name" << "location" if @mode == AppMode::WEB
     allowed_optional_parameters = ["storage_size", "ha_type", "flavor"]
+    allowed_optional_parameters << "accept_partnership" if @mode == AppMode::WEB
```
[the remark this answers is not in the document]
enescakir: > you can do that by not setting the name field of the input

Good trick. I wasn't know that. I remove the name

views/postgres/create.erb
```diff
@@ -15,20 +21,26 @@
...
+      title:  @page_title,
+      right_items: [logo ? "<img src=\"/#{logo}\" class=\"h-6 object-contain\"/>" : nil]
+    }
+  ) %>
 </div>
 
 <div class="grid gap-6">
   <form action="<%= "#{@project_data[:path]}/postgres" %>" method="POST">
     <%== csrf_tag("#{@project_data[:path]}/postgres") %>
+    <%== render("components/form/hidden", locals: {name: "flavor", value: @flavor}) %>
```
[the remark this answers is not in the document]
enescakir: Good catch. I added a validation to the create page. So if the given flavor isn't valid, it rejects to render the create page

```
  def view_create_page
    Authorization.authorize(@user.id, "Postgres:create", project.id)
    flavor = @request.params["flavor"] || PostgresResource::Flavor::STANDARD
    Validation.validate_postgres_flavor(flavor)
    @app.instance_variable_set(:@flavor, flavor)

```

views/postgres/index.erb
enescakir: I renamed it as `brand`

PR #1969, opened 2024-09-25
Add support for partially matching github restore key

velioglu:
While restoring a cache, github cache support partial match by the prefix of existing keys. When a key doesn't match directly, the action searches for keys prefixed with the restore key and returns the most recently created cache. Adding support for that.

### DIFF
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -27,6 +27,20 @@ class CloverRuntime
         .where(key: keys, version: version, scope: scopes).all
         .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
 
+      # GitHub cache supports prefix match if the key doesn't match exactly.
+      # From their docs:
+      #   When a key doesn't match directly, the action searches for keys
+      #   prefixed with the restore key. If there are multiple partial matches
+      #   for a restore key, the action returns the most recently created cache.
+      if entry.nil?
+        entry = repository.cache_entries_dataset
+          .exclude(committed_at: nil)
+          .where { keys.map { |key| Sequel.like(:key, "#{key}%") }.reduce(:|) }
+          .where(version: version, scope: scopes)
+          .order(Sequel.desc(:created_at))
+          .first
+      end
+
       fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil?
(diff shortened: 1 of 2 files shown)


routes/runtime/github.rb
```diff
@@ -27,6 +27,18 @@ class CloverRuntime
         .where(key: keys, version: version, scope: scopes).all
         .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
 
+      # Github cache supports prefix match if the key doesn't match exactly. From the github docs:
+      # "When a key doesn't match directly, the action searches for keys prefixed with the restore key.
+      # If there are multiple partial matches for a restore key, the action returns the most recently created cache."
+      if entry.nil?
```
enescakir, line 33: Can you add a few test cases for it?

routes/runtime/github.rb
```diff
@@ -27,6 +27,18 @@ class CloverRuntime
         .where(key: keys, version: version, scope: scopes).all
         .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
 
+      # Github cache supports prefix match if the key doesn't match exactly. From the github docs:
+      # "When a key doesn't match directly, the action searches for keys prefixed with the restore key.
+      # If there are multiple partial matches for a restore key, the action returns the most recently created cache."
```
enescakir, line 32: nit: wrap by 80
```suggestion
      # GitHub cache supports prefix match if the key doesn't match exactly.
      # From their docs:
      #     When a key doesn't match directly, the action searches for keys
      #     prefixed with the restore key. If there are multiple partial matches
      #     for a restore key, the action returns the most recently created cache.
```

routes/runtime/github.rb
```diff
@@ -27,6 +27,18 @@ class CloverRuntime
...
+      if entry.nil?
+        entry = repository.cache_entries_dataset
```
enescakir, line 34: It fails with following error at my local
```
NoMethodError: undefined method `or' for #<Sequel::SQL::BooleanExpression @op=>:LIKE, @args=>[:key, "c%"]>
from (pry):86:in `inject'
```

enescakir (approved):

PR #1970, opened 2024-09-25
Support reserve cache without size information

velioglu:
It was assumed that reserve cache sends the size information, though it is not valid for all cases (docker layer cache is an example). In that case, reserve cache returns presigned urls for the biggest cache file size. Size of the entry is updated during commit cache as it is always sent with that request.

### DIFF
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -69,8 +69,8 @@ class CloverRuntime
       r.post true do
         key = r.params["key"]
         version = r.params["version"]
-        size = r.params["cacheSize"].to_i
-        fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if key.nil? || version.nil? || size == 0
+        size = r.params["cacheSize"]&.to_i
+        fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if key.nil? || version.nil?
 
         unless (scope = runner.workflow_job&.dig("head_branch"))
           # YYYY: If the webhook not delivered yet, we can try to get the branch from the API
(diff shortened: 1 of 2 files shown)


routes/runtime/github.rb
```diff
@@ -82,7 +82,7 @@ class CloverRuntime
           fail CloverError.new(409, "AlreadyExists", "A cache entry for #{scope} scope already exists with #{key} key and #{version} version.")
         end
 
-        if size > GithubRepository::CACHE_SIZE_LIMIT
+        if size&.> GithubRepository::CACHE_SIZE_LIMIT
```
enescakir, line 85: nit: It's very hard to read

```suggestion
        if size && size > GithubRepository::CACHE_SIZE_LIMIT
```

routes/runtime/github.rb
```diff
@@ -113,7 +118,12 @@ class CloverRuntime
         fail CloverError.new(400, "InvalidRequest", "Wrong parameters") if etags.nil? || etags.empty? || upload_id.nil? || size == 0
 
         entry = GithubCacheEntry[repository_id: repository.id, upload_id: upload_id, committed_at: nil]
-        fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil? || entry.size != size
+        fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil?
```
enescakir, line 121: Should we continue to check size if it's provided?
```suggestion
        fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil? || (entry.size && entry.size != size)
```

routes/runtime/github.rb
```diff
@@ -113,7 +118,12 @@ class CloverRuntime
...
+
+        # If the size can not be set with reserveCache, we set it here.
+        if entry.size.nil?
+          entry.update(size: size)
+        end
```
enescakir, line 126: nit: 
```suggestion
        entry.update(size: size) if entry.size.nil?
```

enescakir: btw we already send an update query to database in the next lines: `entry.update(committed_at: Time.now)`. Can you combine them?

velioglu: I prefer to keep it explicit after checking whether the size is nil.

enescakir: I agree with to keep it explicit. You can combine them like 
```ruby
updates = {committed_at: Time.now}
updates[:size] = size if entry.size.nil?
entry.update(updates)
```

PR #1971, opened 2024-09-25
Support leaseweb culprits ipv6

furkansahin:
## Remove the prefix to be shorter than 68 bits condition
We were expecting the prefix to be shorter than 68 bits but the latest
leaseweb hosts came with ipv6 blocks like:
2607:f5b7:1:64:1::_112/64
2607:f5b7:1:64:2::_112/64
2607:f5b7:1:64:3::_112/64

As you can see, they give /64 but the address space is not actually as
big as /64. Because, the :1, :2, :3 at the end of the prefix is actually
to 80th bit. We must see them like the following for it to make sense;
2607:f5b7:1:64:0001::/64
2607:f5b7:1:64:0002::/64

Here, the differentiating bit is in the residue part of a /64 network.
So, if we use them as they are, we need to actually map them to
2607:f5b7:1:64::/64 which would not work because then we would have
multiple hosts with the same prefix and we cannot assign ipv6 addresses
to the VMs correctly.

To solve this issue, we assign the prefixes correctly using /80. So, the
hosts become
2607:f5b7:1:64:0001::/80
2607:f5b7:1:64:0002::/80

This requires modification in learn_network prog because it was
previously expecting at least /68. We remove the condition.

The main idea of /68 was that so that we can delegeta prefixes at the
size of /80 which doesn't really work because dnsmasq cannot delegate
the prefixes anyway. So, in any case, we simply give /128 to anyone. So,
I remove that condition.

## Update vm_host to support allocation of longer prefixes to VMs
We always tried to allocate /79 to the VM to be used as 2 /80s.This
doesn't work with the new leaseweb hosts because we already get a /80
from them for the host. Therefore, we need to be able to allocate much
smaller. Keeping the entropy the same (2 bytes), we change the ipv6
allocation logic. This way, hetzner hosts will continue allocation /79
while leaseweb allocates /95.

## Use proxy ndp for both clover and guest ephemeral
With the recent change to use dnsmasq as the first level dns server, we
get out to the internet with the clover_ephemeral address. The existing
ndp proxy logic was only proxying the guest ephemeral which makes the VM
IPv6 stack working as expected. However, we realized that the dns
queries through dnsmasq was failing. Further investigation showed that
the dnsmasq was able to send packets out but no response was getting
back in. I realized the source address is the first half of the /95
prefix for dnsmasq. Therefore, we need to enable ndp_proxy for that
prefix, too. This way, our response packets started to flow back in.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -72,21 +72,32 @@ class VmHost < Sequel::Model
 
   # Generate a random network that is a slice of the host's network
   # for delegation to a VM.
-  def ip6_random_vm_network(prefix = 79)
-    # Generates bits worth of a single byte of entropy and then shifts it to the
-    # correct position in the host's network. The bits sits in between the
-    # host's network and the network that is delegated to the VM. We use the
-    # range 2...256 to avoid the case where the lower_bits is 0 or 1, which
-    # would result in the host's network being returned. We also make sure the
-    # higest possible value is 255, so that the lower_bits is never greater than
-    # the host's network. This would be the case if the difference between the
-    # host's network and the network that is delegated to the VM is a single
-    # byte.
-    # The upper range is updated later to increase the entropy if the host has
-    # a big enough prefix. This way we avoid collisions. Keeping the range to
-    # single byte caused elevated number of collisions.
-    upper_range = ((prefix - host_prefix) < 15) ? 2**8 : 2**16
-    lower_bits = SecureRandom.random_number(2...upper_range) << (128 - prefix - 1)
+  def ip6_random_vm_network
+    prefix = host_prefix + 15
+    # We generate 2 bytes of entropy for the lower bits
+    # and append them to the host's network. This way,
+    # the first X bits are the same as the host's network.
+    # And the next 16 bits are random. To achieve that, we shift
+    # the random number to the left by (128 - prefix - 1) bits.
+    # This way, the random number is placed in the correct position.
+    # With a simple example:
+    # Let's say we have a /64 network: 0:0:0:0::/64
+    # This can be expanded to: 0:0:0:0:0:0:0:0/64
+    # With even further expansion: 0000:0000:0000:0000:0000:0000:0000:0000/64
+    # Now, we can generate a random number between 2 and 2**16 (65536). This
+    # will be our next 2 bytes. Let's say the random number is 5001. In base 16,
+    # this is 0x1389.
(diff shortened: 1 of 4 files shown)


spec/model/vm_host_spec.rb
```diff
@@ -64,10 +64,10 @@
     expect(vh.ip6_random_vm_network.to_s).not_to eq(vh.ip6_reserved_network)
   end
 
-  it "can generate ipv6 for hosts with smaller than /64 prefix with single byte" do
+  it "can generate ipv6 for hosts with smaller than /64 prefix with two bytes" do
     vh.net6 = NetAddr.parse_net("2a01:4f9:2b:35a::/68")
-    expect(SecureRandom).to receive(:random_number).with(2...256).and_return(5)
-    expect(vh.ip6_random_vm_network.to_s).to eq("2a01:4f9:2b:35a:4::/79")
+    expect(SecureRandom).to receive(:random_number).with(2...2**16).and_return(5)
+    expect(vh.ip6_random_vm_network.to_s).to eq("2a01:4f9:2b:35a:0:4000:0:0/83")
```
fdr, line 70: I would expand this test to do some boundary conditions: the lowest value, the highest value. I guess a middle value for good measure, and freeze the result of that bit-bashing into the test.

model/vm_host.rb
```diff
@@ -85,8 +86,7 @@ def ip6_random_vm_network(prefix = 79)
     # The upper range is updated later to increase the entropy if the host has
```
fdr, line 86: Are these descriptions still legit? Seems like the algorithm/design theory has changed quite a bit?

spec/model/vm_host_spec.rb
fdr: I might also consider an all-zero address as the test fixture, to make sure the boundaries of your shifting are right, since a bitwise-or will make ambiguous whether a set bit is from the test fixture (is it set from the get-go, or did the `|` set it?)

model/vm_host.rb
furkansahin: Updated them

spec/model/vm_host_spec.rb
furkansahin: Added new tests

fdr (approved):

fdr:
@pykello you are a skilled bit-basher, could you double-check this?

PR #1972, opened 2024-09-25
Allow to set customer quota for GitHub cache storage

enescakir:
We allow 10GB of free cache storage per repository, a limit we've set in line with the official GitHub Actions cache.

However, some customers may need more than 10GB of cache storage. This PR enables us to assign a custom cache storage quota to all repositories within a project, though the quota still applies on a per-repository basis.

### DIFF
--- a/model/project_quota.rb
+++ b/model/project_quota.rb
@@ -10,3 +10,5 @@ class ProjectQuota < Sequel::Model
     end
   end
 end
+
+ProjectQuota.unrestrict_primary_key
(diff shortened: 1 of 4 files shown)


byucesoy (approved):

velioglu (approved):

model/project_quota.rb
```diff
@@ -10,3 +10,5 @@ def self.default_quotas
     end
   end
 end
+
+ProjectQuota.unrestrict_primary_key
```
fdr, line 14: Is the need compelling? It's there because of mass-assignment vulnerabilities. I think we should go over again the other models that do this and ask if it's really necessary: the gain has to be rather large.

enescakir: `:project_id` and `:quota_id` are primary keys for `ProjectQuota`.

It allows us to add a quota like that
```ruby
quota_id = ProjectQuota.default_quotas["GithubRunnerCacheStorage"]["id"]
project.add_quota(quota_id: quota_id, value: 20)
```

Otherwise, we'll need to do something like that
```ruby
quota_id = ProjectQuota.default_quotas["GithubRunnerCacheStorage"]["id"]
project.add_quota(ProjectQuota.new(value: 20).tap { _1.quota_id = quota_id })
```

I believe the first version is more readable. I unrestricted it because we already have it in other models. However, if you consider it a serious vulnerabilities, we can eliminate it from all models.

enescakir: I will refactor all instances of `unrestrict_primary_key` in the subsequent PR.

PR #1978, opened 2024-09-26
managed policies

enescakir:
- **Add migration for managed access policies**
  

- **Add managed access policies**
  One of the biggest issues in our user invitation flow is that invited
  users are not granted any permissions by default. The individual who
  invites the new user must manually assign permissions using our JSON
  policy editor. Often, users are not aware they need to do this.
  Moreover, writing a JSON policy can be challenging for newcomers.
  
  To streamline the user invitation process, we introduced managed
  policies for new users. These policies overlap with popular roles in
  other systems. Assigning roles to users is common practice in many
  applications.
  
  The ACLs of these policies are hidden from the user, who only sees the
  policy name. The controlplane manages the policy content.
  
  When the first user is added to this policy, we create a new access
  policy for the project. If the access policy already exists, we simply
  overwrite its body.
  

- **Allow to select a policy for the users of project**
  I've added a selection box to each row in the user list. Users can
  select a policy for their project and save it.
  
  To identify a user's current policy, we retrieve the managed access
  policies of the project and get the accounts for each policy using the
  'subjects' property of the policy via access tag.
  
  To set managed policies for users, we get the policies for all users of
  the project in a single HTML form request, formatted as
  `user_policies[<ACCOUNT_UBID>] = <POLICY_NAME>`. We simply replace all
  policy content with these users.
  
  We've decided to revise the user management page's design, resulting in
  a few more changes than a single picker would entail.
  

- **Allow to select a policy while inviting a user**
  If the user invited to the project, but doesn't have an account at
  Ubicloud, we persist the invitation until they create one.
  
  With this PR, the inviter can select a policy while inviting an user.
  Once the invited user creates an account, we append them to the existing
  policy's subjects list instead of replacing the entire list.
  
  Additionally, the policy for the invited user can be updated.
  

- **Relocate JSON policy editor to the bottom of the user management page**
  We decided to position JSON policy editor as an advanced configuration.
  
  If the predefined policies do not meet a customer's needs, they can
  grant more granular permissions by editing the custom policy.


fdr:
I have a general question, that may be answerable by carefully going through the patch, but I thought I'd ask at a high level to get your thoughts.

I think providing something like "RBAC sugar" is very cool and good, where it can gradually teach the user about the underlying notation, and handle most needs, especially in the beginning, or in development, where setting up an exact model of a locked down production policy may not always be attractive.

What's the thinking on versioning and changes to the roles we make over time? Are managed roles more statically bound (at time of application) or are they dynamic (where as we add services, people get more privileges).  Are there surprises that we need to accept, such as when re-applying a role, the version may get refreshed?

enescakir:
> I have a general question, that may be answerable by carefully going through the patch, but I thought I'd ask at a high level to get your thoughts.
> 
> I think providing something like "RBAC sugar" is very cool and good, where it can gradually teach the user about the underlying notation, and handle most needs, especially in the beginning, or in development, where setting up an exact model of a locked down production policy may not always be attractive.
> 
> What's the thinking on versioning and changes to the roles we make over time? Are managed roles more statically bound (at time of application) or are they dynamic (where as we add services, people get more privileges). Are there surprises that we need to accept, such as when re-applying a role, the version may get refreshed?

That's a good question. AccessPolicy is created at the time of application, but their content is overwritten when a new user is associated with them. Therefore, if we update the default permissions of the managed policy, users will gain more privileges. Alternatively, we could refresh the generated policies when we add new services, since these are managed by us.

We previously discussed whether or not we should grant more privileges to users when we introduce new services. Currently, we only have "Admin - Can perform all actions" and "Member - Can perform all actions except managing users and billing" as managed policies. Given the member's role, I believe it's acceptable to grant additional privileges, excluding administrative ones. Otherwise, it becomes challenging to enable new services for our customers.

views/project/user.erb
```diff
@@ -151,4 +151,44 @@
       </div>
     </form>
   </div>
+  <!-- Access Policy -->
+  <div id="policy-editor">
```
byucesoy, line 155: What do you think about making this collapsible and collapsed by default?

byucesoy (approved):
I think this PR is fine. Thanks a lot for separate commits. It helped me a lot during the review.

While reviewing it though, I felt like there are number of improvements we can do in our authorization model. These are bigger changes and not necessarily in scope of this PR. However, I'm adding my thoughts here while they are still fresh, so that they can be input for Authorization v2 (if we ever have that).

- We should decouple subject, object and actions. You can think this as normalization of policies. The problem with the coupling is that, we cannot re-use existing list of actions for different projects. Each project needs to have its own copy of the policy, thus when we need to update something we need to change all policies (i.e. more data to store (not that the storage is scarcity), more things to maintain etc.) Instead we can have something like;

```
Policies:
 id | policy_name | allowed_actions
----+-------------+----------------------
 1  | admin       | *
 2  | member      | Vm:*, Postgres:*, ...

Applied_Policies:
 policy_id | user_id | scope (i.e. object)
-----------+---------+---------------------
 1         | 1001    | pjxxx
 2         | 1002    | vmyyy
```

- We should enable exclusion rules and `*:view` type suffix match. I think the former is more important the latter. Exclusion rules would allow us to create a proper Member policy that doesn't need to be updated as we add new services. Initially I was against the exclusion rules, primarily due to it was not clear what to do in below case;

```
Policies:
 id | policy_name            | actions
----+------------------------+---------------------------
 1  | vm                     | Include: Vm:*
 2  | everything_but_vm      | Include: *, Exclude: Vm:*

Applied_Policies:
 policy_id | user_id | scope (i.e. object)
-----------+---------+---------------------
 1         | 1001    | pjxxx
 2         | 1001    | pjxxx
```

Should we allow VM access due to the first policy or not allow due to the second one? Then I checked, how hyperscalers handle this case. They allow Vm access and here is why. They expand actions text to set of all actions that text represents (i.e. Vm:* is expanded to Vm:create, Vm:view, ...) and take union of all applied policies to determine the final access policy. This approach resonated well with me. After all, it says `exclude` not `prevent`. I think if we had a `prevent` rules, that would indeed create the problem above, but with `include`/`exclude` it is OK to handle each actions as `set` and take union of them.

- This could be due to me being non-native speaker, but every time I saw subject and object, I had to stop for a few seconds to remember which one is which :D. Maybe we can rename `object` as `scope` and `subject` as `user`. I know theoretically a VM can be `subject` of a policy as well, but in that context, it is OK to think VM as `user`.

- While thinking about naming; I realized `policy` is an overloaded term. For example, in Azure, you can create a policy that prevents opening up port 22 or prevent deploying any resource to US regions etc. They are primarily used for compliance and security. In our case, it might be better to name this `access policy`/`access level`.

- I think a bit on our current authorization model, RBAC, ABAC and what kind of abstraction we would need eventually. I think at the end we will have at least 5 concepts:
  - `users`
  - `roles` (such as `Admin: {include: [*]}`, `Member: {include: [*], exclude: [billing:*]}`, etc.)
  - `attributes` (such as `has-2fa`, `in-eu`, `in-us`, `logged-in-with-phone` etc.)
  - `scope`
  - `access-policy` (this would tie everything together). Some examples
    - `Enes` should have `Admin` access while is `in-eu` and `logged-in-with-laptop` on `pj...`
    - `Enes` should have `Member` access while is `in-us` or `logged-in-with-phone` on `pj...`

We shouldn't avoid the term `role` due to us implementing ABAC. `role` is well established term and resonates well with people. It is easy to understand/easy to reason about it. Secondly ABAC is `roles` + `attributes`, so `role` concept is already part of ABAC. Finally, we don't have any `attributes` implemented at the moment. While we have the necessary infrastructure for ABAC, it is more closer to RBAC at the moment.

views/project/user.erb
enescakir: It makes sense. However, I'm considering delaying it. Given that our current users have only advanced policies for their projects, we can keep editor visible until we migrate them to managed policies. Otherwise, not being able to see their current policy by default scares them

views/project/user.erb
```diff
@@ -8,105 +8,135 @@
...
+  <div>
+    <div class="md:flex md:items-center md:justify-between pb-1 lg:pb-2">
+      <div class="min-w-0 flex-1">
+        <h3 class="text-2xl font-bold leading-7 text-gray-900 sm:truncate sm:text-2xl sm:tracking-tight">
+          Invite a new user
+        </h3>
+      </div>
+    </div>
+    <div class="overflow-hidden rounded-lg shadow ring-1 ring-black ring-opacity-5 bg-white divide-y divide-gray-200">
+      <!-- Invite user -->
```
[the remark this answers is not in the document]
enescakir: Yes, I moved it.

views/project/user.erb
byucesoy: That is fine by me.

PR #1979, opened 2024-09-26
Foundations of an Inference Endpoints feature

bsatzger:
This pull request introduces the groundwork for OpenAI compatible AI inferencing endpoints hosted on Ubicloud. Most important additions:

**New concepts**
The new ApiKey Sequel model and Nexus progs enables the creation and rotation of API keys, which can be linked to projects or specific inference endpoints. These are used to authenticate requests sent to the inference endpoint.

Inference endpoint and inference endpoint replica Sequel models and Nexus progs are introduced. An inference endpoint defines which AI model to serve and also how to serve it (e.g. which inference engine to use). It is backed by a private network, a load balancer, and multiple inference endpoint replicas. A replica is basically a VM attached to the load balancer running an inference gateway together with and inference engine serving the specified model. The inference gateway listens on the load balancer port, receives and authenticates requests, and forwards them to the inference engine. The inference gateway is discussed in a separate PR (in another repo).

**New Rhizomes**
Mainly create systemd units to setup the inference engine and inference gateway after VM provisioning. They also create a service/timer to periodically download the latest load balancer certs. 

**New Boot Image**
The PR adds support for new AI-specific boot images. An AI base boot images contains essential software (GPU drivers, inference engine, inference gateway) whereas AI model images contain the AI models themselves. The former is the actual VM boot image. The latter is mounted as a separate read-only disk to the VM.


migrate/20240926_inference_endpoint.rb
```diff
@@ -0,0 +1,33 @@
...
+      column :vm_size, :text, collate: '"C"', null: false
+      column :model_name, :text, collate: '"C"', null: false
+      column :storage_volumes, :jsonb, null: false
+      column :engine, :text, collate: '"C"', null: false
+      column :engine_params, :text, collate: '"C"', null: false
+      column :replica_count, :integer, null: false
+      foreign_key :project_id, :project, type: :uuid, null: false
+      foreign_key :load_balancer_id, :load_balancer, type: :uuid, null: false
+      foreign_key :private_subnet_id, :private_subnet, type: :uuid, null: false
+      index :is_public
```
fdr, line 23: Interesting, to index a binary value like this. Do you perhaps want a conditional index (WHERE is_public/not is_public) or a compound index?

lib/option.rb
```diff
@@ -27,8 +28,11 @@ module Option
     end
   end
 
+  AI_MODELS = ai_models.select { _1["enabled"] }
+
   PROVIDERS.freeze
   LOCATIONS.freeze
+  AI_MODELS.freeze
```
fdr, line 35: Since `freeze` returns self, you could condense this into one line:  `ai_models.select { _1["enabled"] }.freeze`

lib/validation.rb
```diff
@@ -100,7 +100,7 @@ def self.validate_storage_volumes(storage_volumes, boot_disk_index)
     end
     storage_volumes.each { |volume|
       volume.each_key { |key|
-        fail ValidationFailed.new({storage_volumes: "Invalid key: #{key}"}) unless allowed_keys.include?(key)
+        fail ValidationFailed.new({storage_volumes: "Invalid key: #{key}"}) unless allowed_keys.include?(key.to_sym)
```
fdr, line 103: Nice one, though I wonder if we should consider constraining the inputs, or if that is just too impractical (usually because keyword arguments or ruby program literals are easier as symbols, and sometimes we're working with data, which tends to come in as strings)

rhizome/host/bin/setup-vm
```diff
@@ -51,6 +51,7 @@ begin
   swap_size_bytes = params["swap_size_bytes"]
   pci_devices = params.fetch("pci_devices", [])
   boot_image = params["boot_image"]
+  readonly_image = params.fetch("readonly_image", "")
```
fdr, line 54: this is a worthy feature all on its own. the patch is not *so* large that it's a problem to inline it (but it would be even better to herald this enhancement in its own commit message), but support for read only volumes is definitely its own cool and good thing.

rhizome/inference_endpoint/bin/download-lb-cert
```diff
@@ -0,0 +1,30 @@
...
+timestamp=$(date +%s)
+new_ssl_folder="/ie/workdir/ssl_$timestamp"
+rm -rf $new_ssl_folder
+rm -rf /ie/workdir/ssl_latest
+mkdir $new_ssl_folder
+
+if [ "$1" = "self-signed" ]; then
+    openssl req -x509 -out $new_ssl_folder/ubi_cert.pem -keyout $new_ssl_folder/ubi_key.pem -newkey rsa:2048 -nodes -sha256 -days 7 -subj '/CN=localhost'
+else
+    curl -f -L3 [FD00:0B1C:100D:5afe:CE::]:8080/load-balancer/cert.pem > $new_ssl_folder/ubi_cert.pem
```
fdr, line 13: ah interesting, not sure how I memory-holed this, being on 8080. I can see why that's easier on the implementation, but I wonder if we should splice in an nftables rule to rewrite the port when dst=`[FD00:0B1C:100D:5afe:CE::].8080` to `80` to spare the customer the extra notation (cc @furkansahin ). Similar to the LB's own port rewriting feature for exactly the same reason (binding unprivileged ports, serving on a privileged port)

rhizome/inference_endpoint/bin/download-lb-cert
```diff
@@ -0,0 +1,30 @@
...
+set -ueo pipefail
+
+timestamp=$(date +%s)
+new_ssl_folder="/ie/workdir/ssl_$timestamp"
+rm -rf $new_ssl_folder
+rm -rf /ie/workdir/ssl_latest
+mkdir $new_ssl_folder
+
+if [ "$1" = "self-signed" ]; then
+    openssl req -x509 -out $new_ssl_folder/ubi_cert.pem -keyout $new_ssl_folder/ubi_key.pem -newkey rsa:2048 -nodes -sha256 -days 7 -subj '/CN=localhost'
```
fdr, line 11: This is fine, though, don't we have a bunch of self-signed certificate generation code in Ruby? Is there any benefit there to having our self-signed development regimes be systematic? Not a rhetorical question question, I haven't re-surveyed things; this may be a consistency without value (and may complicate things)

secondly, you could speed this up a by getting off RSA and using elliptic curves.  https://wiki.mozilla.org/Security/Server_Side_TLS 

Perplexity sez:

```bash
openssl req -x509 -out $new_ssl_folder/ubi_cert.pem -keyout $new_ssl_folder/ubi_key.pem -newkey ec -pkeyopt ec_paramgen_curve:P-256 -nodes -sha256 -days 7 -subj '/CN=localhost'
```

rhizome/inference_endpoint/lib/replica_setup.rb
```diff
@@ -0,0 +1,140 @@
...
+    r "systemctl start inference-engine.service"
+  end
+
+  def inference_gateway_service
+    "/etc/systemd/system/inference-gateway.service"
+  end
+
+  def write(path, s)
+    s += "\n" unless s.end_with?("\n")
+    File.write(path, s)
```
fdr, line 31: I think `puts` has these semantics "for free"

rhizome/inference_endpoint/lib/replica_setup.rb
```diff
@@ -0,0 +1,140 @@
...
+  def engine_start_command(inference_engine:, inference_engine_params:, model:)
+    case inference_engine
+    when "vllm"
+      "/opt/miniconda/envs/vllm/bin/vllm serve /ie/models/model --served-model-name #{model} --disable-log-requests --host 127.0.0.1 #{inference_engine_params}"
+    else
+      fail "BUG: unsupported inference engine"
+    end
+  end
+
+  def write_config_files(replica_ubid, ssl_crt_path, ssl_key_path, gateway_port)
```
fdr, line 71: with regard to systemd units: some things we've been somewhat uncommitted to, but probably would be served by being more committed to, is to run  `systemd-analyze  security` on the units. Our problem in other units, e.g. dnsmasq, is they are kind of complex in their interaction with system capabilities. With the exception of access to the GPU, both the vllm server and gateway may less resemble this, and you can harden them considerably by going down the list and being practical. Many are one-liners

fdr (commented):
not quite done but looks like pretty good work to me

model/ai/inference_endpoint_replica.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require_relative "../../model"
+
+class InferenceEndpointReplica < Sequel::Model
```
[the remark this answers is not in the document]
bsatzger: I see how this is confusing for DB folks. Kubernetes calls each identical Pod in a Deployment or ReplicaSet a replica. 
https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
```

migrate/20240926_inference_endpoint.rb
bsatzger: Actually I feel that index isn't that crucial atm. It anticipates we'll have many inference endpoints and we often need to distinguish between public vs non-public inference endpoints. Maybe let's not index this column right now at all and add one in the future if necessary.

lib/validation.rb
bsatzger: Fair point. I updated the PR to explicitly convert keys of storage volume definition (that we read from the ai_models.yml) to symbols. And got rid of this commit.

rhizome/inference_endpoint/lib/replica_setup.rb
bsatzger: thanks. changed to `puts`.

prog/ai/inference_endpoint_nexus.rb
```diff
@@ -0,0 +1,132 @@
...
+  def self.assemble(project_id:, location:, boot_image:, name:, vm_size:, storage_volumes:, model_name:,
+    engine:, engine_params:, replica_count:, is_public:)
+    unless (project = Project[project_id])
+      fail "No existing project"
+    end
+
+    Validation.validate_location(location)
+    Validation.validate_name(name)
+    Validation.validate_vm_size(vm_size)
+    fail "Invalid replica count" unless replica_count.is_a?(Integer) && replica_count > 0 && replica_count < 10
```
byucesoy, line 45: It becomes bit difficult to read unless statements with multiple conditions. How about;

```
fail if !replica_count.is_a?(Integer) || replica_count <= 0 || replica_count >= 10
```
or

```
fail if !replica_count.is_a?(Integer) || replica_count < 1 || replica_count > 9
```

model/vm.rb
```diff
@@ -15,6 +15,7 @@ class Vm < Sequel::Model
   one_to_many :pci_devices, key: :vm_id, class: :PciDevice
   one_through_one :load_balancer, left_key: :vm_id, right_key: :load_balancer_id, join_table: :load_balancers_vms
   one_to_one :load_balancers_vms, key: :vm_id, class: LoadBalancersVms
+  one_to_one :inference_endpoint_replica, key: :vm_id
```
byucesoy, line 18: Should VM, which is a more general concept refer to inference_endpoint_replica. I expect many VMs to not have inference_endpoint_replica.

Instead, maybe it would make more sense to pass readonly and image name as a fields in storage devices? (I'm not sure what would make sense, but overall I don't think lower level services like VM should refer to higher level ones if possible)

prog/ai/inference_endpoint_nexus.rb
byucesoy, line 45: It becomes bit difficult to read unless statements with multiple conditions. How about;

```
fail if !replica_count.is_a?(Integer) || replica_count <= 0 || replica_count >= 10
```
or

```
fail if !replica_count.is_a?(Integer) || replica_count < 1 || replica_count > 9
```

byucesoy (commented):
I didn't do a detailed review. I was just curious on what we are doing in AI front so I had a quick look and left 2 suggestions.

migrate/20240926_inference_endpoint.rb
bsatzger: removed the index

rhizome/host/bin/setup-vm
bsatzger: added separate commit.

rhizome/inference_endpoint/bin/download-lb-cert
bsatzger: we do have some code on the control plane side to create certificates for Postgres. This one-liner is purely to fill a gap in development, as the Load Balancer doesn't produce any certificate to download in dev environments (when no DNS server is available).

I changed it to use elliptic curves.

rhizome/inference_endpoint/lib/replica_setup.rb
bsatzger: inference-gateway  and inference-engine (vllm) run as a service users now. should we do hardening (identify minimum set of capabilities) in a subsequent pr?

prog/ai/inference_endpoint_nexus.rb
fdr: I think you can go shorter:

```ruby
fail unless replica_count.is_a?(Integer) && (1..9).cover?(replica_count)
```

While `cover?` handles most odd types, it doesn't handle, say, floating point, or rationals. So I left the type check in.

rhizome/inference_endpoint/bin/download-lb-cert
fdr: yeah, I'm fine with the one liner, more wondering if there's any development ease by, say, having a shared trust root even for development.  We can worry about it later, I don't think we have any established pattern for this (and maybe we never will)

rhizome/inference_endpoint/lib/replica_setup.rb
fdr: I dunno, why not get it over with? Can you dump systemd-analyze security on your two units? I can probably generate you a systemd fragment that gets you 80% of the way there in no time

model/ai/inference_endpoint_replica.rb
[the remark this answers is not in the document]
fdr: "Replica" is kind of growing on me, like: "AppReplica" says more than "Backend," which is more customary. OK, let's leave it.

prog/ai/inference_endpoint_nexus.rb
bsatzger: switched to 
```ruby
unless replica_count.is_a?(Integer) && (1..9).cover?(replica_count)
```

model/vm.rb
bsatzger: you are right. got rid of this. I now pass two volumes when creating vms for inference replicas: the boot volume and a new "read_only" volume.

rhizome/inference_endpoint/lib/replica_setup.rb
bsatzger: Added some sandboxing settings to the systemd units and tested that these don't break their functionality. With that in place it looks like below. 

```
systemd-analyze security
UNIT                                  EXPOSURE PREDICATE HAPPY
ModemManager.service                       6.3 MEDIUM    😐
cron.service                               9.6 UNSAFE    😨
dbus.service                               9.5 UNSAFE    😨
dm-event.service                           9.5 UNSAFE    😨
dmesg.service                              9.6 UNSAFE    😨
emergency.service                          9.5 UNSAFE    😨
getty@tty1.service                         9.6 UNSAFE    😨
inference-engine.service                   8.0 EXPOSED   🙁
inference-gateway.service                  7.3 MEDIUM    😐
iscsid.service                             9.5 UNSAFE    😨
lb-cert-download.service                   7.9 EXPOSED   🙁
lvm2-lvmpolld.service                      9.5 UNSAFE    😨
multipathd.service                         9.5 UNSAFE    😨
networkd-dispatcher.service                9.6 UNSAFE    😨
nvidia-persistenced.service                9.6 UNSAFE    😨
open-vm-tools.service                      9.5 UNSAFE    😨
plymouth-start.service                     9.5 UNSAFE    😨
polkit.service                             1.6 OK        🙂
rc-local.service                           9.6 UNSAFE    😨
rescue.service                             9.5 UNSAFE    😨
rsyslog.service                            6.3 MEDIUM    😐
serial-getty@ttyS0.service                 9.6 UNSAFE    😨
setup.service                              9.6 UNSAFE    😨
snapd.service                              9.8 UNSAFE    😨
ssh.service                                9.6 UNSAFE    😨
systemd-ask-password-console.service       9.4 UNSAFE    😨
systemd-ask-password-plymouth.service      9.5 UNSAFE    😨
systemd-ask-password-wall.service          9.4 UNSAFE    😨
systemd-bsod.service                       9.5 UNSAFE    😨
systemd-fsckd.service                      9.5 UNSAFE    😨
systemd-initctl.service                    9.4 UNSAFE    😨
systemd-journald.service                   4.3 OK        🙂
systemd-logind.service                     2.8 OK        🙂
systemd-networkd.service                   2.6 OK        🙂
systemd-resolved.service                   2.2 OK        🙂
systemd-rfkill.service                     9.4 UNSAFE    😨
systemd-timesyncd.service                  2.1 OK        🙂
systemd-udevd.service                      7.1 MEDIUM    😐
tpm-udev.service                           9.6 UNSAFE    😨
ubuntu-advantage.service                   9.6 UNSAFE    😨
udisks2.service                            9.6 UNSAFE    😨
unattended-upgrades.service                9.6 UNSAFE    😨
user@1000.service                          9.4 UNSAFE    😨
uuidd.service                              5.8 MEDIUM    😐
vgauth.service                             9.5 UNSAFE    😨

systemd-analyze security inference-engine.service
  NAME                                                        DESCRIPTION                                                             EXPOSURE
✗ RemoveIPC=                                                  Service user may leave SysV IPC objects around                               0.1
✗ RootDirectory=/RootImage=                                   Service runs within the host's root directory                                0.1
✓ User=/DynamicUser=                                          Service runs under a static non-root user identity
✗ CapabilityBoundingSet=~CAP_SYS_TIME                         Service processes may change the system clock                                0.2
✓ NoNewPrivileges=                                            Service processes cannot acquire new privileges
✓ AmbientCapabilities=                                        Service process does not receive ambient capabilities
✗ PrivateDevices=                                             Service potentially has access to hardware devices                           0.2
✗ ProtectClock=                                               Service may write to the hardware clock or system clock                      0.2
✗ CapabilityBoundingSet=~CAP_SYS_PACCT                        Service may use acct()                                                       0.1
✗ CapabilityBoundingSet=~CAP_KILL                             Service may send UNIX signals to arbitrary processes                         0.1
✗ ProtectKernelLogs=                                          Service may read from or write to the kernel log ring buffer                 0.2
✗ CapabilityBoundingSet=~CAP_WAKE_ALARM                       Service may program timers that wake up the system                           0.1
✗ CapabilityBoundingSet=~CAP_(DAC_*|FOWNER|IPC_OWNER)         Service may override UNIX file/IPC permission checks                         0.2
✗ CapabilityBoundingSet=~CAP_BPF                              Service may not load BPF programs                                            0.1
✗ CapabilityBoundingSet=~CAP_LINUX_IMMUTABLE                  Service may mark files immutable                                             0.1
✗ CapabilityBoundingSet=~CAP_IPC_LOCK                         Service may lock memory into RAM                                             0.1
✗ ProtectKernelModules=                                       Service may load or read kernel modules                                      0.2
✗ CapabilityBoundingSet=~CAP_SYS_MODULE                       Service may load kernel modules                                              0.2
✗ CapabilityBoundingSet=~CAP_SYS_TTY_CONFIG                   Service may issue vhangup()                                                  0.1
✗ CapabilityBoundingSet=~CAP_SYS_BOOT                         Service may issue reboot()                                                   0.1
✗ CapabilityBoundingSet=~CAP_SYS_CHROOT                       Service may issue chroot()                                                   0.1
✗ SystemCallArchitectures=                                    Service may execute system calls with all ABIs                               0.2
✗ CapabilityBoundingSet=~CAP_BLOCK_SUSPEND                    Service may establish wake locks                                             0.1
✗ MemoryDenyWriteExecute=                                     Service may create writable executable memory mappings                       0.1
✗ RestrictNamespaces=~user                                    Service may create user namespaces                                           0.3
✗ RestrictNamespaces=~pid                                     Service may create process namespaces                                        0.1
✗ RestrictNamespaces=~net                                     Service may create network namespaces                                        0.1
✗ RestrictNamespaces=~uts                                     Service may create hostname namespaces                                       0.1
✗ RestrictNamespaces=~mnt                                     Service may create file system namespaces                                    0.1
✗ CapabilityBoundingSet=~CAP_LEASE                            Service may create file leases                                               0.1
✗ CapabilityBoundingSet=~CAP_MKNOD                            Service may create device nodes                                              0.1
✗ RestrictNamespaces=~cgroup                                  Service may create cgroup namespaces                                         0.1
✗ RestrictSUIDSGID=                                           Service may create SUID/SGID files                                           0.2
✗ RestrictNamespaces=~ipc                                     Service may create IPC namespaces                                            0.1
✗ ProtectHostname=                                            Service may change system host/domainname                                    0.1
✗ CapabilityBoundingSet=~CAP_(CHOWN|FSETID|SETFCAP)           Service may change file ownership/access mode/capabilities unrestricted      0.2
✗ CapabilityBoundingSet=~CAP_SET(UID|GID|PCAP)                Service may change UID/GID identities/capabilities                           0.3
✗ LockPersonality=                                            Service may change ABI personality                                           0.1
✗ RestrictAddressFamilies=~AF_PACKET                          Service may allocate packet sockets                                          0.2
✗ RestrictAddressFamilies=~AF_NETLINK                         Service may allocate netlink sockets                                         0.1
✗ RestrictAddressFamilies=~AF_UNIX                            Service may allocate local sockets                                           0.1
✗ RestrictAddressFamilies=~…                                  Service may allocate exotic sockets                                          0.3
✗ RestrictAddressFamilies=~AF_(INET|INET6)                    Service may allocate Internet sockets                                        0.3
✗ CapabilityBoundingSet=~CAP_MAC_*                            Service may adjust SMACK MAC                                                 0.1
✗ RestrictRealtime=                                           Service may acquire realtime scheduling                                      0.1
✓ ProtectSystem=                                              Service has strict read-only access to the OS file hierarchy
✗ CapabilityBoundingSet=~CAP_SYS_RAWIO                        Service has raw I/O access                                                   0.2
✗ CapabilityBoundingSet=~CAP_SYS_PTRACE                       Service has ptrace() debugging abilities                                     0.3
✗ CapabilityBoundingSet=~CAP_SYS_(NICE|RESOURCE)              Service has privileges to change resource use parameters                     0.1
✓ SupplementaryGroups=                                        Service has no supplementary groups
✗ DeviceAllow=                                                Service has no device ACL                                                    0.2
✓ PrivateTmp=                                                 Service has no access to other software's temporary files
✓ ProtectHome=                                                Service has no access to home directories
✗ CapabilityBoundingSet=~CAP_NET_ADMIN                        Service has network configuration privileges                                 0.2
✗ ProtectProc=                                                Service has full access to process tree (/proc hidepid=)                     0.2
✗ ProcSubset=                                                 Service has full access to non-process /proc files (/proc subset=)           0.1
✗ CapabilityBoundingSet=~CAP_NET_(BIND_SERVICE|BROADCAST|RAW) Service has elevated networking privileges                                   0.1
✗ CapabilityBoundingSet=~CAP_AUDIT_*                          Service has audit subsystem access                                           0.1
✗ CapabilityBoundingSet=~CAP_SYS_ADMIN                        Service has administrator privileges                                         0.3
✗ PrivateNetwork=                                             Service has access to the host's network                                     0.5
✗ PrivateUsers=                                               Service has access to other users                                            0.2
✗ CapabilityBoundingSet=~CAP_SYSLOG                           Service has access to kernel logging                                         0.1
✓ KeyringMode=                                                Service doesn't share key material with other services
✓ Delegate=                                                   Service does not maintain its own delegated control group subtree
✗ SystemCallFilter=~@clock                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@cpu-emulation                            Service does not filter system calls                                         0.1
✗ SystemCallFilter=~@debug                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@module                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@mount                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@obsolete                                 Service does not filter system calls                                         0.1
✗ SystemCallFilter=~@privileged                               Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@raw-io                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@reboot                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@resources                                Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@swap                                     Service does not filter system calls                                         0.2
✗ IPAddressDeny=                                              Service does not define an IP address allow list                             0.2
✓ NotifyAccess=                                               Service child processes cannot alter service state
✓ ProtectControlGroups=                                       Service cannot modify the control group file system
✓ PrivateMounts=                                              Service cannot install system mounts
✓ ProtectKernelTunables=                                      Service cannot alter kernel tunables (/proc/sys, …)
✗ UMask=                                                      Files created by service are world-readable by default                       0.1

→ Overall exposure level for inference-engine.service: 8.0 EXPOSED 🙁
systemd-analyze security inference-gateway.service
  NAME                                                        DESCRIPTION                                                             EXPOSURE
✗ RemoveIPC=                                                  Service user may leave SysV IPC objects around                               0.1
✗ RootDirectory=/RootImage=                                   Service runs within the host's root directory                                0.1
✓ User=/DynamicUser=                                          Service runs under a static non-root user identity
✗ CapabilityBoundingSet=~CAP_SYS_TIME                         Service processes may change the system clock                                0.2
✓ NoNewPrivileges=                                            Service processes cannot acquire new privileges
✓ AmbientCapabilities=                                        Service process does not receive ambient capabilities
✗ ProtectClock=                                               Service may write to the hardware clock or system clock                      0.2
✗ CapabilityBoundingSet=~CAP_SYS_PACCT                        Service may use acct()                                                       0.1
✗ CapabilityBoundingSet=~CAP_KILL                             Service may send UNIX signals to arbitrary processes                         0.1
✗ ProtectKernelLogs=                                          Service may read from or write to the kernel log ring buffer                 0.2
✗ CapabilityBoundingSet=~CAP_WAKE_ALARM                       Service may program timers that wake up the system                           0.1
✗ CapabilityBoundingSet=~CAP_(DAC_*|FOWNER|IPC_OWNER)         Service may override UNIX file/IPC permission checks                         0.2
✗ CapabilityBoundingSet=~CAP_BPF                              Service may not load BPF programs                                            0.1
✗ CapabilityBoundingSet=~CAP_LINUX_IMMUTABLE                  Service may mark files immutable                                             0.1
✗ CapabilityBoundingSet=~CAP_IPC_LOCK                         Service may lock memory into RAM                                             0.1
✗ ProtectKernelModules=                                       Service may load or read kernel modules                                      0.2
✗ CapabilityBoundingSet=~CAP_SYS_MODULE                       Service may load kernel modules                                              0.2
✗ CapabilityBoundingSet=~CAP_SYS_TTY_CONFIG                   Service may issue vhangup()                                                  0.1
✗ CapabilityBoundingSet=~CAP_SYS_BOOT                         Service may issue reboot()                                                   0.1
✗ CapabilityBoundingSet=~CAP_SYS_CHROOT                       Service may issue chroot()                                                   0.1
✗ SystemCallArchitectures=                                    Service may execute system calls with all ABIs                               0.2
✗ CapabilityBoundingSet=~CAP_BLOCK_SUSPEND                    Service may establish wake locks                                             0.1
✗ MemoryDenyWriteExecute=                                     Service may create writable executable memory mappings                       0.1
✗ RestrictNamespaces=~user                                    Service may create user namespaces                                           0.3
✗ RestrictNamespaces=~pid                                     Service may create process namespaces                                        0.1
✗ RestrictNamespaces=~net                                     Service may create network namespaces                                        0.1
✗ RestrictNamespaces=~uts                                     Service may create hostname namespaces                                       0.1
✗ RestrictNamespaces=~mnt                                     Service may create file system namespaces                                    0.1
✗ CapabilityBoundingSet=~CAP_LEASE                            Service may create file leases                                               0.1
✗ RestrictNamespaces=~cgroup                                  Service may create cgroup namespaces                                         0.1
✗ RestrictSUIDSGID=                                           Service may create SUID/SGID files                                           0.2
✗ RestrictNamespaces=~ipc                                     Service may create IPC namespaces                                            0.1
✗ ProtectHostname=                                            Service may change system host/domainname                                    0.1
✗ CapabilityBoundingSet=~CAP_(CHOWN|FSETID|SETFCAP)           Service may change file ownership/access mode/capabilities unrestricted      0.2
✗ CapabilityBoundingSet=~CAP_SET(UID|GID|PCAP)                Service may change UID/GID identities/capabilities                           0.3
✗ LockPersonality=                                            Service may change ABI personality                                           0.1
✗ RestrictAddressFamilies=~AF_PACKET                          Service may allocate packet sockets                                          0.2
✗ RestrictAddressFamilies=~AF_NETLINK                         Service may allocate netlink sockets                                         0.1
✗ RestrictAddressFamilies=~AF_UNIX                            Service may allocate local sockets                                           0.1
✗ RestrictAddressFamilies=~…                                  Service may allocate exotic sockets                                          0.3
✗ RestrictAddressFamilies=~AF_(INET|INET6)                    Service may allocate Internet sockets                                        0.3
✗ CapabilityBoundingSet=~CAP_MAC_*                            Service may adjust SMACK MAC                                                 0.1
✗ RestrictRealtime=                                           Service may acquire realtime scheduling                                      0.1
✓ ProtectSystem=                                              Service has strict read-only access to the OS file hierarchy
✗ CapabilityBoundingSet=~CAP_SYS_PTRACE                       Service has ptrace() debugging abilities                                     0.3
✗ CapabilityBoundingSet=~CAP_SYS_(NICE|RESOURCE)              Service has privileges to change resource use parameters                     0.1
✓ SupplementaryGroups=                                        Service has no supplementary groups
✓ CapabilityBoundingSet=~CAP_SYS_RAWIO                        Service has no raw I/O access
✓ PrivateTmp=                                                 Service has no access to other software's temporary files
✓ ProtectHome=                                                Service has no access to home directories
✓ PrivateDevices=                                             Service has no access to hardware devices
✗ CapabilityBoundingSet=~CAP_NET_ADMIN                        Service has network configuration privileges                                 0.2
✗ ProtectProc=                                                Service has full access to process tree (/proc hidepid=)                     0.2
✗ ProcSubset=                                                 Service has full access to non-process /proc files (/proc subset=)           0.1
✗ CapabilityBoundingSet=~CAP_NET_(BIND_SERVICE|BROADCAST|RAW) Service has elevated networking privileges                                   0.1
✗ CapabilityBoundingSet=~CAP_AUDIT_*                          Service has audit subsystem access                                           0.1
✗ CapabilityBoundingSet=~CAP_SYS_ADMIN                        Service has administrator privileges                                         0.3
✗ PrivateNetwork=                                             Service has access to the host's network                                     0.5
✗ PrivateUsers=                                               Service has access to other users                                            0.2
✗ CapabilityBoundingSet=~CAP_SYSLOG                           Service has access to kernel logging                                         0.1
✓ DeviceAllow=                                                Service has a minimal device ACL
✓ KeyringMode=                                                Service doesn't share key material with other services
✓ Delegate=                                                   Service does not maintain its own delegated control group subtree
✗ SystemCallFilter=~@clock                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@cpu-emulation                            Service does not filter system calls                                         0.1
✗ SystemCallFilter=~@debug                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@module                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@mount                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@obsolete                                 Service does not filter system calls                                         0.1
✗ SystemCallFilter=~@privileged                               Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@raw-io                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@reboot                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@resources                                Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@swap                                     Service does not filter system calls                                         0.2
✗ IPAddressDeny=                                              Service does not define an IP address allow list                             0.2
✓ NotifyAccess=                                               Service child processes cannot alter service state
✓ ProtectControlGroups=                                       Service cannot modify the control group file system
✓ PrivateMounts=                                              Service cannot install system mounts
✓ CapabilityBoundingSet=~CAP_MKNOD                            Service cannot create device nodes
✓ ProtectKernelTunables=                                      Service cannot alter kernel tunables (/proc/sys, …)
✗ UMask=                                                      Files created by service are world-readable by default                       0.1

→ Overall exposure level for inference-gateway.service: 7.3 MEDIUM 😐
systemd-analyze security lb-cert-download.service
  NAME                                                        DESCRIPTION                                                             EXPOSURE
✗ RootDirectory=/RootImage=                                   Service runs within the host's root directory                                0.1
  SupplementaryGroups=                                        Service runs as root, option does not matter
  RemoveIPC=                                                  Service runs as root, option does not apply
✗ User=/DynamicUser=                                          Service runs as root user                                                    0.4
✗ CapabilityBoundingSet=~CAP_SYS_TIME                         Service processes may change the system clock                                0.2
✓ NoNewPrivileges=                                            Service processes cannot acquire new privileges
✓ AmbientCapabilities=                                        Service process does not receive ambient capabilities
✗ ProtectClock=                                               Service may write to the hardware clock or system clock                      0.2
✗ CapabilityBoundingSet=~CAP_SYS_PACCT                        Service may use acct()                                                       0.1
✗ CapabilityBoundingSet=~CAP_KILL                             Service may send UNIX signals to arbitrary processes                         0.1
✗ ProtectKernelLogs=                                          Service may read from or write to the kernel log ring buffer                 0.2
✗ CapabilityBoundingSet=~CAP_WAKE_ALARM                       Service may program timers that wake up the system                           0.1
✗ CapabilityBoundingSet=~CAP_(DAC_*|FOWNER|IPC_OWNER)         Service may override UNIX file/IPC permission checks                         0.2
✗ CapabilityBoundingSet=~CAP_BPF                              Service may not load BPF programs                                            0.1
✗ CapabilityBoundingSet=~CAP_LINUX_IMMUTABLE                  Service may mark files immutable                                             0.1
✗ CapabilityBoundingSet=~CAP_IPC_LOCK                         Service may lock memory into RAM                                             0.1
✗ ProtectKernelModules=                                       Service may load or read kernel modules                                      0.2
✗ CapabilityBoundingSet=~CAP_SYS_MODULE                       Service may load kernel modules                                              0.2
✗ CapabilityBoundingSet=~CAP_SYS_TTY_CONFIG                   Service may issue vhangup()                                                  0.1
✗ CapabilityBoundingSet=~CAP_SYS_BOOT                         Service may issue reboot()                                                   0.1
✗ CapabilityBoundingSet=~CAP_SYS_CHROOT                       Service may issue chroot()                                                   0.1
✗ SystemCallArchitectures=                                    Service may execute system calls with all ABIs                               0.2
✗ CapabilityBoundingSet=~CAP_BLOCK_SUSPEND                    Service may establish wake locks                                             0.1
✗ MemoryDenyWriteExecute=                                     Service may create writable executable memory mappings                       0.1
✗ RestrictNamespaces=~user                                    Service may create user namespaces                                           0.3
✗ RestrictNamespaces=~pid                                     Service may create process namespaces                                        0.1
✗ RestrictNamespaces=~net                                     Service may create network namespaces                                        0.1
✗ RestrictNamespaces=~uts                                     Service may create hostname namespaces                                       0.1
✗ RestrictNamespaces=~mnt                                     Service may create file system namespaces                                    0.1
✗ CapabilityBoundingSet=~CAP_LEASE                            Service may create file leases                                               0.1
✗ RestrictNamespaces=~cgroup                                  Service may create cgroup namespaces                                         0.1
✗ RestrictSUIDSGID=                                           Service may create SUID/SGID files                                           0.2
✗ RestrictNamespaces=~ipc                                     Service may create IPC namespaces                                            0.1
✗ ProtectHostname=                                            Service may change system host/domainname                                    0.1
✗ CapabilityBoundingSet=~CAP_(CHOWN|FSETID|SETFCAP)           Service may change file ownership/access mode/capabilities unrestricted      0.2
✗ CapabilityBoundingSet=~CAP_SET(UID|GID|PCAP)                Service may change UID/GID identities/capabilities                           0.3
✗ LockPersonality=                                            Service may change ABI personality                                           0.1
✗ RestrictAddressFamilies=~AF_PACKET                          Service may allocate packet sockets                                          0.2
✗ RestrictAddressFamilies=~AF_NETLINK                         Service may allocate netlink sockets                                         0.1
✗ RestrictAddressFamilies=~AF_UNIX                            Service may allocate local sockets                                           0.1
✗ RestrictAddressFamilies=~…                                  Service may allocate exotic sockets                                          0.3
✗ RestrictAddressFamilies=~AF_(INET|INET6)                    Service may allocate Internet sockets                                        0.3
✗ CapabilityBoundingSet=~CAP_MAC_*                            Service may adjust SMACK MAC                                                 0.1
✗ RestrictRealtime=                                           Service may acquire realtime scheduling                                      0.1
✓ ProtectSystem=                                              Service has strict read-only access to the OS file hierarchy
✗ CapabilityBoundingSet=~CAP_SYS_PTRACE                       Service has ptrace() debugging abilities                                     0.3
✗ CapabilityBoundingSet=~CAP_SYS_(NICE|RESOURCE)              Service has privileges to change resource use parameters                     0.1
✓ CapabilityBoundingSet=~CAP_SYS_RAWIO                        Service has no raw I/O access
✓ PrivateTmp=                                                 Service has no access to other software's temporary files
✓ PrivateDevices=                                             Service has no access to hardware devices
✗ CapabilityBoundingSet=~CAP_NET_ADMIN                        Service has network configuration privileges                                 0.2
✗ ProtectProc=                                                Service has full access to process tree (/proc hidepid=)                     0.2
✗ ProcSubset=                                                 Service has full access to non-process /proc files (/proc subset=)           0.1
✗ ProtectHome=                                                Service has full access to home directories                                  0.2
✗ CapabilityBoundingSet=~CAP_NET_(BIND_SERVICE|BROADCAST|RAW) Service has elevated networking privileges                                   0.1
✗ CapabilityBoundingSet=~CAP_AUDIT_*                          Service has audit subsystem access                                           0.1
✗ CapabilityBoundingSet=~CAP_SYS_ADMIN                        Service has administrator privileges                                         0.3
✗ PrivateNetwork=                                             Service has access to the host's network                                     0.5
✗ PrivateUsers=                                               Service has access to other users                                            0.2
✗ CapabilityBoundingSet=~CAP_SYSLOG                           Service has access to kernel logging                                         0.1
✓ DeviceAllow=                                                Service has a minimal device ACL
✓ KeyringMode=                                                Service doesn't share key material with other services
✓ Delegate=                                                   Service does not maintain its own delegated control group subtree
✗ SystemCallFilter=~@clock                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@cpu-emulation                            Service does not filter system calls                                         0.1
✗ SystemCallFilter=~@debug                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@module                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@mount                                    Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@obsolete                                 Service does not filter system calls                                         0.1
✗ SystemCallFilter=~@privileged                               Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@raw-io                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@reboot                                   Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@resources                                Service does not filter system calls                                         0.2
✗ SystemCallFilter=~@swap                                     Service does not filter system calls                                         0.2
✗ IPAddressDeny=                                              Service does not define an IP address allow list                             0.2
✓ NotifyAccess=                                               Service child processes cannot alter service state
✓ ProtectControlGroups=                                       Service cannot modify the control group file system
✓ PrivateMounts=                                              Service cannot install system mounts
✓ CapabilityBoundingSet=~CAP_MKNOD                            Service cannot create device nodes
✓ ProtectKernelTunables=                                      Service cannot alter kernel tunables (/proc/sys, …)
✗ UMask=                                                      Files created by service are world-readable by default                       0.1

→ Overall exposure level for lb-cert-download.service: 7.9 EXPOSED 🙁
```

fdr: The most crucial are the services handling (possibly malicious) user input.

Here's what I massaged Perplexity into saying:


```
[Service]
# User management
DynamicUser=yes
SupplementaryGroups=
PrivateUsers=yes # Low risk, but may cause issues if the service needs to interact with other users

# Capability and system call restrictions
CapabilityBoundingSet= # High risk, may be too restrictive for a web app
SystemCallFilter=@system-service
SystemCallFilter=~@privileged @resources @mount @debug @cpu-emulation @obsolete @raw-io @reboot @swap
SystemCallArchitectures=native # Low risk

# File system and device restrictions
ProtectSystem=strict # Medium risk, may be too restrictive if the app needs to write to system directories
ProtectHome=yes # Low risk
PrivateTmp=yes # Low risk
PrivateDevices=yes # Low risk for a web app
PrivateMounts=yes # Low risk
DeviceAllow= # Low risk for a web app

# Kernel and system protections
ProtectKernelTunables=yes # Low risk
ProtectKernelModules=yes # Low risk
ProtectKernelLogs=yes # Low risk
ProtectClock=yes # Low risk
ProtectHostname=yes # Low risk
ProtectControlGroups=yes # Low risk
ProtectProc=invisible # Medium risk, may interfere with monitoring or debugging
ProcSubset=pid # Medium risk, may interfere with monitoring or debugging

# Execution environment restrictions
NoNewPrivileges=yes # Low risk
RestrictNamespaces=yes # Low risk
RestrictRealtime=yes # Low risk
RestrictSUIDSGID=yes # Low risk
LockPersonality=yes # Low risk

# Network restrictions
RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6 # Low risk
IPAddressDeny=any # High risk, may prevent the web app from functioning
IPAddressAllow=localhost # High risk, may prevent the web app from functioning
PrivateNetwork=no # Low risk, appropriate for a web app

# Additional hardening
MemoryDenyWriteExecute=yes # Medium risk, may interfere with some web frameworks
RemoveIPC=yes # Low risk
UMask=0077 # Low risk
KeyringMode=private # Low risk
```

The "network restrictions" will definitely need work, you may even choose to leave them out. But if your inference server is only working by unix socket, or localhost binding, you can say so. The gateway would be less restricted.

The other area that is considered risky is dropping so many capabilities. I grilled my good pal perplexity some more, and the obvious one is CAP_NET_BIND_SERVICE, to bind to low ports in classic web servers like nginx that have a hyper-paranoid bit of code to bind to 80/443 and them drop to a less-privileged users. We shouldn't need that kind of thing, since the load balancer features rewriting to an unprivileged port from the get-go, e.g. 1443 and mapping packets to that.

fdr: You can use `include` features to have one copy, to factor out common stuff, e.g.


`IncludeDropins=hardening.conf.d`

fdr: You can use the "drop-in" functionality with symbolic links to copy the settings-in-common to multiple units:

https://stackoverflow.com/questions/59842743/what-is-a-drop-in-file-what-is-a-drop-in-directory-how-to-edit-systemd-service

I've found the easy way to do it is to use the `sudo systemctl edit` feature when developing, then taking note of the generated file paths and perhaps giving it a better name than "override.conf" and symlinking as necessary.

rhizome/inference_endpoint/bin/download-lb-cert
furkansahin: Alright, that makes sense, I'll create a PR on that.

furkansahin: https://github.com/ubicloud/ubicloud/pull/1993

rhizome/inference_endpoint/bin/download-lb-cert
```diff
@@ -0,0 +1,30 @@
...
+new_ssl_folder="/ie/workdir/ssl_$timestamp"
+rm -rf $new_ssl_folder
+rm -rf /ie/workdir/ssl_latest
+mkdir $new_ssl_folder
+
+if [ "$1" = "self-signed" ]; then
+    openssl req -x509 -out $new_ssl_folder/ubi_cert.pem -keyout $new_ssl_folder/ubi_key.pem -newkey ec -pkeyopt ec_paramgen_curve:P-256 -nodes -sha256 -days 7 -subj '/CN=localhost'
+else
+    curl -f -L3 [FD00:0B1C:100D:5afe:CE::]:8080/load-balancer/cert.pem > $new_ssl_folder/ubi_cert.pem
+    curl -f -L3 [FD00:0B1C:100D:5afe:CE::]:8080/load-balancer/key.pem > $new_ssl_folder/ubi_key.pem
```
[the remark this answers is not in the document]
bsatzger: there's a `download_lb_cert` label in inference_endpoint_replica_nexus.rb. it's responsible for the initial download of the cert. the strand will be stuck in that label until the cert is available. the timer that runs every hour is only required for cert rotation.

prog/vnet/cert_nexus.rb
```diff
@@ -131,4 +141,20 @@ def dns_record_name
   def dns_zone
     @dns_zone ||= DnsZone[cert.dns_zone_id]
   end
+
+  def create_self_signed_cert(duration)
```
bsatzger, line 145: we have similar functionality in our utils, which is used by Postgres and Minio, but that doesn't produce certs compatible for this use-case.

rhizome/inference_endpoint/lib/replica_setup.rb
fdr: sounds good, that's a lot more than most do, so it'll be easier to focus on the remaining for marginal gains (if they're even possible). Thanks for chopping through the list.

rhizome/inference_endpoint/bin/download-lb-cert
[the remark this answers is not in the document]
fdr: Interesting workaround that's fine for now, but @furkansahin may have to think about a good way to help the client spin more aggressively when the LB cert is imminently, but not yet, available. Maybe 503: temporarily unavailable with a Retry-After

rhizome/host/lib/vm_setup.rb
```diff
@@ -163,7 +163,7 @@ def purge_storage
     storage_roots = []
 
     params = JSON.parse(File.read(vp.prep_json))
-    params["storage_volumes"].each { |params|
+    params["storage_volumes"].reject { _1["read_only"] }.each { |params|
```
fdr, line 166: I know we're pretty unsystematic on rhizome testing, but is there anything cheap-and-easy we can do for now to exercise the read-only code a tiny bit? to make sure it generates the right kind of cloud-hypervisor line. If it's not so straightforward and we need to pencil in "bring rhizome up to speed on test model" for Soon, that's fine too.

fdr (approved):
I figure you probably want to get this in to ease integration and start finding/solving other bugs, and I don't have any pressing concerns right now worth holding that up. If any other team members have a little stitch in time they can suggest, then, by all means...

spec/prog/ai/inference_endpoint_replica_nexus_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+    end
+  end
+
+  describe "#setup" do
+    it "triggers setup if setup command is not sent yet or failed" do
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo inference_endpoint/bin/setup-replica' setup", {stdin: "{\"inference_engine\":\"vllm\",\"inference_engine_params\":\"--some-params\",\"model\":\"llama\",\"replica_ubid\":\"theubid\",\"ssl_crt_path\":\"/ie/workdir/ssl/ubi_cert.pem\",\"ssl_key_path\":\"/ie/workdir/ssl/ubi_key.pem\",\"gateway_port\":8443}"}).twice
+      expect(inference_endpoint).to receive(:engine).and_return("vllm").twice
+      expect(inference_endpoint).to receive(:engine_params).and_return("--some-params").twice
+      expect(inference_endpoint).to receive(:model_name).and_return("llama").twice
+      expect(inference_endpoint).to receive(:load_balancer).and_return(instance_double(LoadBalancer, id: "lb-id", dst_port: 8443)).twice
```
fdr, line 134: fyi, you may be able to condense this with instance_double (which doesn't care about `twice`, which I tend to view as overfitting unless there's something very special about the logic under test) or `at_least(:once)` (which I really wish was less wordy, but handles the case where one is indifferent to the number of times it is called)

rhizome/inference_endpoint/lib/replica_setup.rb
```diff
@@ -0,0 +1,203 @@
...
+PrivateUsers=yes
+CapabilityBoundingSet=
+SystemCallFilter=@system-service
+SystemCallFilter=~@privileged @resources @mount @debug @cpu-emulation @obsolete @raw-io @reboot @swap
+SystemCallArchitectures=native
+ProtectSystem=strict
+DeviceAllow=
+MemoryDenyWriteExecute=true
+RemoveIPC=true
+UMask=0077
```
fdr, line 171: fascinating, all/any of these break the inference engine? well, can look into that later, at least the first pass already does quite a bit

fdr (approved):

rhizome/host/lib/vm_setup.rb
bsatzger: Added some easy-to-make checks.

rhizome/inference_endpoint/bin/download-lb-cert
bsatzger: Made the switch to port 80 as the PR above was merged.

PR #1983, opened 2024-09-27
Add predefined address mapping for blob storage cache

velioglu:
When restoring a cache, the local cache server uses localhost.blob.core.windows.net as the host for file requests. This setup enables the toolkit’s concurrent download functionality (refer to the cacheHttpClient in the actions/toolkit repository for more details). The local cache server then expects to receive these requests. To ensure this, we map the VM's local IP so that any requests to that URL are always redirected back to the VM itself.

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -425,6 +425,9 @@ DHCP
     raparams = nics.map { "ra-param=#{_1.tap}" }.join("\n")
     interfaces = nics.map { "interface=#{_1.tap}" }.join("\n")
     dnsmasq_address_ip6 = NetAddr::IPv6.parse("fd00:0b1c:100d:53::")
+    address_mapping = boot_image.include?("github") ?
+      "address=/localhost.blob.core.windows.net/#{nics.first.net4.split("/").first}" :
+      ""
     vp.write_dnsmasq_conf(<<DNSMASQ_CONF)
 pid-file=
 leasefile-ro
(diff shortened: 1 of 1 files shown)


rhizome/host/lib/vm_setup.rb
```diff
@@ -445,6 +445,7 @@ def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image)
 listen-address=#{dnsmasq_address_ip6}
 dhcp-option=26,1400
 bind-interfaces
+address=/localhost.blob.core.windows.net/#{nics.first.net4.split("/").first}
```
fdr, line 448: whoa though, this affects every VM right? Can't you use /etc/hosts as a GHA special to do something similar?

velioglu: Yes, it affects every VM. We've discussed with Furkan whether it is safe for VMs and opened in that way but I can  check to do that for runner vms only. I can't use /etc/hosts as I need to have that mapping inside containers as well. I need to have that for containers because docker layer cache uses buildkit to build images and it caches files via container.

furkansahin: Upon further discussion, we think that we can get away by checking the boot_image so that it doesn't impact every VM.

fdr: Oof, containers, yes...I guess this the most well-sealed way to do it.

velioglu: With https://github.com/ubicloud/ubicloud/pull/2016 we are assigning ipv4 address to dnsmasq so it can used for dns resolution. I've also updated the code to have that address mapping only for github vms.

furkansahin (approved):

fdr (approved):

fdr:
Also  it would be good to have a test, since this introduces a branch. Alas, rhizome is one of the vexing weak points we have for no particularly good reason.

PR #1984, opened 2024-09-27
GitHub Runner UI Improvements

enescakir:
### **Extract tabbar to a separate component with small UI fixes**

I plan to refactor "GitHub Runners" page similar to "Networking" page, using the same tab bar component.
  
Additionally, there's a minor issue with border rounding. The middle corners shouldn't be rounded.
  
| Before | After |
|---|---|
| Screenshot 2024-09-27 at 20 15 48 | Screenshot 2024-09-27 at 20 19 36 |

### **Divide GitHub runners page into two subpages**

The GitHub runners page shows both active runners and connected accounts.
  
Additionally, I plan to add a cache list, which is becoming quite lengthy.
  
I've refactored it to resemble the 'Networking' page, splitting it into two sections with tabbar: 'Active Runners' and 'Settings'.
  
I'll add a 'Cache' subpage in the future. We could also consider adding an 'enable/disable' button for transparent cache on the 'Settings' page.
 
| Before | After |
|---|---|
| Screenshot 2024-09-27 at 20 18 38 | Screenshot 2024-09-27 at 20 19 01 |

### **Change no billing page for GitHub integration**

When a project lacks billing information, an error message is displayed at the top of the page. However, the "Connect Account" button remains visible, but clicking on it yields no results.
  
Users should navigate to the billing page to add new billing information, a process that requires two steps.
  
I've refactored the billing page for GitHub integration. Now, instead of the "Connect Account" button, a "New Billing Information" button is displayed. This allows users to add new billing information in a single step.
  
| Before | After |
|---|---|
| Screenshot 2024-09-27 at 20 16 13 | Screenshot 2024-09-27 at 20 20 11 |

### **Add documentation link to no runner page**

When there's no active runner for the project, we provide guidelines to the user.
  
We've updated our documentation to change the default label from `ubicloud` to `ubicloud-standard-2`.
  
Additionally, I've included a link to the documentation page to inform users about the different types of runners available.

| Before | After |
|---|---|
| Screenshot 2024-09-27 at 20 16 53 | Screenshot 2024-09-27 at 20 19 46 |


views/github/runner.erb
```diff
@@ -68,12 +68,20 @@
...
+                  To use Ubicloud runners, you need to change
+                  <span class="text-rose-500 text-sm font-medium bg-slate-100 py-0.5 px-2 rounded font-mono whitespace-nowrap">runs-on: ...</span>
+                  line to one of our labels such as
+                  <span class="text-rose-500 text-sm font-medium bg-slate-100 py-0.5 px-2 rounded font-mono whitespace-nowrap">runs-on: ubicloud-standard-2</span>
+                  in your workflow file. Then, trigger your workflow to start a runner.
+                  <br>
+                  Check out
+                  <a href="https://www.ubicloud.com/docs/github-actions-integration/quickstart" class="text-orange-500 font-medium">our documentation</a>
+                  for using larger runners, using arm64, enabling GPU.
+                </p>
```
furkansahin, line 84: should we move this outside of the tabs? So, we can remove the marked part:
371629827-4aed16e0-2049-4b41-a2df-f5dec669dac5

furkansahin (approved):

views/github/runner.erb
furkansahin: Actually, thinking about this more, considering we have the "No active runners" message, I think keeping this is the right way.

PR #1985, opened 2024-09-27
GitHub Cache UI

enescakir:
Screenshot 2024-10-01 at 14 50 20

### **Add cache entry list to UI**
  I added the list of cache entries as a subpage to the GitHub runners
  page. It shows data sizes and timestamps in humanized format. To
  accomplish this, I wrote two simple helper functions.
  
  GitHub Actions cache has similar page 
  (https://github.com/ubicloud/ubicloud/actions/caches). There are several
use cases for this UI:
  - We give 10GB of free cache storage and purge any caches that haven't
    been accessed for 7 days or that exceed the storage limit. This new
    UI allows users to monitor their cache storage usage and see future
    needs. They can choose to clear some caches to free up space and
    prevent cache eviction.
  - Sometimes, users might need to delete certain caches to force their
    regeneration, especially if the caches are corrupted.

### **Allow centered icon without text in the button component**
  When no text is provided in the button component, the icon should be
  centered within the button. Right now, the icon is not centered.
  
  Additionally, if no confirmation is provided for the delete button, the
  confirmation dialog shouldn't prompt for text input.

### DIFF
--- a/routes/web/project/github.rb
+++ b/routes/web/project/github.rb
@@ -39,5 +39,15 @@ class CloverWeb
         r.redirect "https://github.com/apps/#{Config.github_app_name}/installations/new", 302
       end
     end
+
+    r.on "cache" do
+      r.get true do
+        repository_id_q = @project.github_installations_dataset.join(:github_repository, installation_id: :id).select(Sequel[:github_repository][:id])
+        @entries = Serializers::GithubCacheEntry.serialize(GithubCacheEntry.where(repository_id: repository_id_q).exclude(committed_at: nil).eager(:repository).order(Sequel.desc(:created_at)).all)
+        @total_usage = Serializers::GithubCacheEntry.humanize_size(@entries.filter_map { _1[:size] }.sum)
+
+        view "github/cache"
+      end
+    end
   end
 end
--- a/routes/web/project/github.rb
+++ b/routes/web/project/github.rb
@@ -48,6 +48,22 @@ class CloverWeb
 
         view "github/cache"
       end
+
+      r.is String do |entry_ubid|
+        entry = GithubCacheEntry.from_ubid(entry_ubid)
+
+        unless entry
+          response.status = 404
+          r.halt
+        end
+
+        r.delete true do
+          entry.destroy
+          flash["notice"] = "Cache '#{entry.key}' deleted."
+          response.status = 204
+          request.halt
+        end
+      end
     end
   end
 end
--- a/serializers/github_cache_entry.rb
+++ b/serializers/github_cache_entry.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class Serializers::GithubCacheEntry < Serializers::Base
+  def self.serialize_internal(entry, options = {})
+    {
+      id: entry.ubid,
+      key: entry.key,
+      repository_name: entry.repository.name,
+      scope: entry.scope,
+      created_at: entry.created_at.strftime("%Y-%m-%d %H:%M UTC"),
+      created_at_human: humanize_time(entry.created_at),
+      last_accessed_at: entry.last_accessed_at&.strftime("%Y-%m-%d %H:%M UTC"),
+      last_accessed_at_human: humanize_time(entry.last_accessed_at),
+      size: entry.size,
+      size_human: humanize_size(entry.size)
+    }
+  end
+
+  def self.humanize_size(bytes)
+    return nil if bytes.nil? || bytes.zero?
+    units = %w[B KB MB GB]
+    exp = (Math.log(bytes) / Math.log(1024)).to_i
+    exp = [exp, units.size - 1].min
+
+    return "%d %s" % [bytes.to_f / 1024**exp, units[exp]] if exp == 0
+    "%.1f %s" % [bytes.to_f / 1024**exp, units[exp]]
+  end
+
+  def self.humanize_time(time)
+    return nil unless time
+    seconds = Time.now - time
+    return "just now" if seconds < 60
+    return "#{(seconds / 60).to_i} minutes ago" if seconds < 3600
+    return "#{(seconds / 3600).to_i} hours ago" if seconds < 86400
+    "#{(seconds / 86400).to_i} days ago"
+  end
+end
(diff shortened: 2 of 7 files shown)


routes/web/project/github.rb
```diff
@@ -39,5 +39,31 @@ class CloverWeb
         r.redirect "https://github.com/apps/#{Config.github_app_name}/installations/new", 302
       end
     end
+
+    r.on "cache" do
+      r.get true do
+        repository_ids = @project.github_installations_dataset.eager(:repositories).all.flat_map(&:repositories).map(&:id)
```
fdr, line 45: This fetches all the data (e.g. detoasting, busting potential index only scans, etc) in the record from Postgres and makes ruby objects with all of those, and then throws it away for the "id" strings.

```ruby
@project.github_installations_datase.dataset.join(:github_repository, installation_id: :id).select_map(Sequel[:github_repository][:id])
```

But perhaps you can get a similar effect from something like this?  I don't know "eager" as well as "join"

```ruby
@project.github_installations_dataset.eager(:repositories).select_map(Sequel[:github_repository][:id])
```

serializers/github_cache_entry.rb
```diff
@@ -0,0 +1,36 @@
...
+    }
+  end
+
+  def self.humanize_size(bytes)
+    return nil if bytes.nil? || bytes.zero?
+    units = %w[B KB MB GB]
+    exp = (Math.log(bytes) / Math.log(1024)).to_i
+    exp = [exp, units.size - 1].min
+
+    "%.2f %s" % [bytes.to_f / 1024**exp, units[exp]]
```
furkansahin, line 25: nit: do we need the .2f granularity? let's just make it easy to the eye

furkansahin (approved):

serializers/github_cache_entry.rb
enescakir: Do you prefer just integer or .1f granularity? I believe there's a difference between `3MB` and `3.8MB`.

routes/web/project/github.rb
```diff
@@ -39,5 +39,15 @@ class CloverWeb
         r.redirect "https://github.com/apps/#{Config.github_app_name}/installations/new", 302
       end
     end
+
+    r.on "cache" do
+      r.get true do
+        repository_id_q = @project.github_installations_dataset.join(:github_repository, installation_id: :id).select(Sequel[:github_repository][:id])
+        @entries = Serializers::GithubCacheEntry.serialize(GithubCacheEntry.where(repository_id: repository_id_q).exclude(committed_at: nil).eager(:repository).order(Sequel.desc(:created_at)).all)
```
fdr, line 46: radical, a dataset composition (passing one into another) to get a subquery? do I have that right?

fdr:
high level question: how does this look with some caches generated by common github actions. Mostly, wondering how many keys are involved. I imagine "not that many" if you opted to display in this way.

My second question is, what is the dominant use case here? One is figuring out how much you're spending on caches and why (maybe looking for excessively large objects), another may be deleting a cache entry if iterating on some build code that involves the cache, as an alternative to cache-busting in the build while iterating...anything else?

fdr:
Incidentally: would be good background for one/some of the commit messages (the "why")

routes/web/project/github.rb
enescakir: Yeah, it returns `Sequel::Dataset` instead of an array when you use `select` instead of `select_map`, allowing you to use it as a subquery in where filtering.

enescakir:
```ruby
[19] clover-development(main)> repository_id_q = @project.github_installations_dataset.join(:github_repository, installation_id: :id).select(Sequel[:github_repository][:id])
=> #<Sequel::Postgres::Dataset: "SELECT \"github_repository\".\"id\" FROM \"github_installation\" INNER JOIN \"github_repository\" ON (\"github_repository\".\"installation_id\" = \"github_installation\".\"id\") WHERE (\"github_installation\".\"project_id\" = $1); [\"ed0b26bf-53c4-82d2-9a00-21e5f05dc364\"]">

[20] clover-development(main)> GithubCacheEntry.where(repository_id: repository_id_q).exclude(committed_at: nil).eager(:repository).order(Sequel.desc(:created_at)).sql

SELECT * 
FROM "github_cache_entry" 
WHERE (
  (
    "repository_id" IN (
      SELECT "github_repository"."id" 
      FROM "github_installation" 
      INNER JOIN "github_repository" ON ("github_repository"."installation_id" = "github_installation"."id") 
      WHERE ("github_installation"."project_id" = $1)
    )
  ) 
  AND ("committed_at" IS NOT NULL)
) 
ORDER BY "created_at" DESC


```

enescakir: I learned this from @furkansahin with this line https://github.com/ubicloud/ubicloud/blob/3ed6a2fa4f0a4e7a389cadcdf3b66fd78584cac9/model/vm_pool.rb#L19

serializers/github_cache_entry.rb
furkansahin: .1f sounds good.

enescakir:
> high level question: how does this look with some caches generated by common github actions. Mostly, wondering how many keys are involved. I imagine "not that many" if you opted to display in this way.
> 
> My second question is, what is the dominant use case here? One is figuring out how much you're spending on caches and why (maybe looking for excessively large objects), another may be deleting a cache entry if iterating on some build code that involves the cache, as an alternative to cache-busting in the build while iterating...anything else?

1. `setup-*` actions generally generates a single cache with a very long key. Docker layer cache generates different entry for each layer.
Screenshot 2024-10-01 at 14 22 02

2. I updated the commit message

   GitHub Actions cache has similar page (https://github.com/ubicloud/ubicloud/actions/caches). There are several use cases for this UI:
    - We give 10GB of free cache storage and purge any caches that haven't been accessed for 7 days or that exceed the storage limit. This new UI allows users to monitor their cache storage usage and see future needs. They can choose to clear some caches to free up space and prevent cache eviction.
    - Sometimes, users might need to delete certain caches to force their regeneration, especially if the caches are corrupted.

PR #1992, opened 2024-09-30
Send customer invoices as PDF attachments

enescakir:
### **Generate invoice PDF at server side using prawnpdf**

In the early days, we chose to delegate the responsibility of generating invoice PDFs to the client side. We simply designed the invoice in HTML and utilized the customer's browser for PDF generation.
  
We adopted this approach because it was quick and straightforward to implement. However, it had its drawbacks. Primarily, it did not generate identical PDFs across all browsers as it heavily depended on the browser and its version. Secondly, we aimed to send invoices as PDFs via email, necessitating server-side PDF generation.
  
Given that we already had an HTML template for the invoice, our initial thought was to convert HTML to PDF. There are multiple tools available for this, such as wkhtmltopdf. However, we hesitated to add another dependency as it could complicate our deployment process.

We decided to use prawnpdf [^1] for server-side PDF generation. Prawn is a pure Ruby PDF generation library that offers a wide range of features for PDF generation. It is user-friendly and highly flexible.
  
Prawn provides excellent manuals for reference [^2] [^3]. You can check the details and its functionalities.

The design process was straightforward. The most significant challenge was dealing with encoding. Unfortunately, the default PDF fonts do not support all UTF-8 characters. If your text includes some special characters, it triggers the following error:
  
    Your document includes text that's not compatible with
    the Windows-1252 character set. (Prawn::Errors::IncompatibleStringEncoding)
    If you need full UTF-8 support, use external fonts instead of PDF's built-in fonts.
  
As the exception suggests, we should use external fonts for full UTF-8 support. I considered several fonts, such as "OpenSans" and "DejaVuSans", before deciding on the "BeVietnamPro" [^4] font. It's already in use on our website and might be useful in our console at some point. Thus, it seemed sensible to use it for PDF generation as well. It's open-source and licensed under the SIL Open Font License [^5].
  
I added the "pdf-reader" gem to read the generated PDF file in the tests, which allows us to verify its content.
  
[^1]: https://github.com/prawnpdf/prawn
[^2]: https://prawnpdf.org/docs/prawn/2.5.0/manual.pdf
[^3]: https://prawnpdf.org/prawn-table-manual.pdf
[^4]: https://fonts.google.com/specimen/Be+Vietnam+Pro/about
[^5]: https://openfontlicense.org

| **Old invoice:** | **New invoice:** |
|---|---|
| old-invoice | new-invoice-2 |

### **Remove code for client-side PDF generation**

Since we started to generate invoice PDFs on the server side, the code for client-side PDF generation is no longer necessary.


enescakir:
The tests are failing due to missing font files. After consulting with our lawyer, I'll add them to this PR.

lib/util.rb
```diff
@@ -77,14 +77,18 @@ def self.safe_write_to_file(filename, content)
...
+  def self.send_email(receiver, subject, greeting: nil, body: nil, button_title: nil, button_link: nil, cc: nil, attachments: [])
     html = EmailRenderer.new.render "email/layout", locals: {subject: subject, greeting: greeting, body: body, button_title: button_title, button_link: button_link}
     Mail.deliver do
       from Config.mail_from
       to receiver
       subject subject
       cc cc
 
+      attachments.each do |name, file|
+        add_file filename: name, content: file
```
byucesoy, line 89: I liked the fact that we don't need to save the PDF to disk and pass content directly. :+1:

model/invoice.rb
```diff
@@ -116,6 +118,105 @@ def send_failure_email(errors)
...
+        style(row(0..1).columns(0..1), padding: [2, 5, 2, 5], borders: [])
+        style(column(0), align: :right, font_style: :semibold, text_color: dark_gray)
+        style(column(1), align: :right)
+      end
+    end
+    pdf.move_down row.height.to_i - 20
+
+    # Row 3: Invoice items
+    items = [["RESOURCE", "DESCRIPTION", "USAGE", "AMOUNT"]]
+    # TODO: CHECK NO USAGE
```
byucesoy, line 193: What is this TODO?

model/invoice.rb
```diff
@@ -116,6 +118,105 @@ def send_failure_email(errors)
...
+    items += data[:items].map { [_1[:name], _1[:description], _1[:usage], _1[:cost_humanized]] }
+    pdf.table items, header: true, width: pdf.bounds.width, cell_style: {size: 9, border_color: "E5E7EB", borders: [], padding: 5} do
```
byucesoy, line 195: Is it possible to add small padding for cells? The old invoice looks bit nicer IMO.

byucesoy (approved):

model/invoice.rb
enescakir: I've added a bit more padding. What's your opinion on this?
new-invoice-2

enescakir: I didn't add much padding initially because the list could get very long if you're a heavy GitHub Actions user with runners of different sizes. i.e.:

long-invoice

model/invoice.rb
enescakir: Good catch. I had planned to add something like that but forgot. It's added now.

Screenshot 2024-09-30 at 16 33 41

fdr (approved):
I'll make my approval contingent on Burak's.

But, I wanted express appreciation for doing the work to doing lowish-level PDF generation, instead of bringing in a headless browser or something large like that, and devising a test model for it on the other end.

PR #1993, opened 2024-09-30
NAT MetadataEndpoint to port 8080 from 80

furkansahin:
Our metadata provider runs on port 8080. Here, we add a simple nat rule to dnat requests on that address and port 80 to 8080. This way, customer can simply run the following to get cert.pem
curl [FD00:0B1C:100D:5afe:CE::]/load-balancer/cert.pem

Before, this was:
curl [FD00:0B1C:100D:5afe:CE::]:8080/load-balancer/key.pem

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -389,6 +389,14 @@ add element inet drop_unused_ip_packets allowed_ipv4_addresses { #{ip_net} }
           #{generate_ip6_private_filter_rules(nics[1..])}
         }
       }
+
+      table ip6 nat_metadata_endpoint {
+        chain prerouting {
+          type nat hook prerouting priority dstnat; policy accept;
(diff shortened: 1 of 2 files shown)


enescakir:
Why don't we initially run the web server on port 80?

furkansahin:
@enescakir port numbers under 1024 are seen as privileged ports. Normal users must not run services on port number <= 1024. Though, it is HTTP port, so it's only normal that customers would directly connect to that port. That's why we NAT.

fdr (approved):
although probably fine, I wonder if it's possible to go simpler w.r.t. `ct state`. I leave it to you though, it's so very marginal. And it's not like people should be hammering this endpoint anyway (and in fact, it may reject them if they did so)

```
    ip6 daddr FD00:0B1C:100D:5AFE:CE:: tcp dport 80 dnat to [FD00:0B1C:100D:5AFE:CE::]:8080
```

fdr:
At some future point, could we also *not* allow :8080 ports, if nobody is using it?  Could leave two modes of address forever, but it'd be nice to have exactly one.

bsatzger (approved):
thanks!

PR #2002, opened 2024-10-01
Start versioning metadata-endpoint binary

furkansahin:
This commit does 2 things:
1. Starts properly versioning the metadata-endpoint binary used. This way, we will be able to identify which version is used per VM and migrate if required.
2. Upgrade the version from v0.1.3 to v0.1.4. With this version, we start allowing upto 4 requests for burst. The previous version only allowed 1 request which caused unexpected 429 errors.

### DIFF
--- a/rhizome/host/lib/cert_server_setup.rb
+++ b/rhizome/host/lib/cert_server_setup.rb
@@ -35,18 +35,22 @@ class CertServerSetup
     "/etc/systemd/system/#{service_name}.service"
   end
 
+  def server_version
+    "0.1.4"
+  end
+
   def server_main_path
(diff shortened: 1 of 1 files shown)


enescakir:
Does this require migration for existing cert servers?

furkansahin:
@enescakir , nope, because the only vm attached to a load balancer at the moment belongs to me.

enescakir (approved):

bsatzger:
and in the future, if you update from 0.1.4 to 0.1.5, is it guaranteed that 0.1.5 is downloaded before a path with 0.1.5 is used?

PR #2010, opened 2024-10-02
Load Aws::S3 module before freezing them

enescakir:
We freeze everything to catch malicious self modifying code from user input. Since the Aws SDK started to autoload modules when used, so we need to load them before freezing.

https://github.com/aws/aws-sdk-ruby/pull/3105

### DIFF
--- a/loader.rb
+++ b/loader.rb
@@ -128,6 +128,14 @@ def clover_freeze
   # this Ruby standard library method patches core classes.
   "".unicode_normalize(:nfc)
 
+  # Aws SDK started to autoload modules when used, so we need to load them
+  # before freezing. https://github.com/aws/aws-sdk-ruby/pull/3105
+  # rubocop:disable Lint/Void
+  Aws::S3::Client
+  Aws::S3::Presigner
+  Aws::S3::Errors
+  # rubocop:enable Lint/Void
+
   # A standard library method that edits/creates a module variable as
   # a side effect.  We encountered it when using rubygems for its tar
   # file writing.
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,7 @@ end
 
 gem "webauthn", "~> 3.1"
 
-gem "aws-sdk-s3", "~> 1.163"
+gem "aws-sdk-s3", "~> 1.167"
 
 gem "acme-client", "~> 2.0"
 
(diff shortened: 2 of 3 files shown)


model/github/github_repository.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
-require "aws-sdk-s3"
-
 require_relative "../../model"
 
+require "aws-sdk-s3"
+
 class GithubRepository < Sequel::Model
```
fdr, line 7: extra spurious change?

fdr (approved):
fine as it is, but I had some minor stuff you could streamline with.

loader.rb
```diff
@@ -128,6 +128,12 @@ def clover_freeze
   # this Ruby standard library method patches core classes.
   "".unicode_normalize(:nfc)
 
+  # Aws SDK started to autoload modules when used, so we need to load them
+  # before freezing. https://github.com/aws/aws-sdk-ruby/pull/3105
+  Aws::S3.const_get(:Client)
+  Aws::S3.const_get(:Presigner)
+  Aws::S3.const_get(:Errors)
```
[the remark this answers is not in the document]
enescakir: Screenshot 2024-10-09 at 22 28 49

Rubocop fails with the following error: `Lint/Void: Constant Aws::S3::Client used in void context.`

What do you think about something like that?
```ruby
[:Client, :Presigner, :Errors].each { Aws::S3.const_get(_1) }
```

model/github/github_repository.rb
enescakir: We generally require gems after requiring relative paths in our codebase. I just make it similar to others.

loader.rb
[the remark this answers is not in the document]
fdr: Sure I guess, I suppose it is admirably short, which of course sells it, though I guess it would be fine to also use a pragma comment to turn that off that linter constraint; this is a particular part of the code where we, in fact, do want to do weird stuff for side effects and do not want to be bothered about "void context"

model/github/github_repository.rb
fdr: ah okay, I'm going to be a bit of a pedant here, but it does make the patch confusing, because the subject is dealing with side effects of loading a library, so that load order may be implicated in something is not completely off-base. Although I'm not as pedantic about PRs glomming in a few unrelated changes, I would have understood that if it were committed separately as a cosmetic fix

enescakir: Fair point

fdr (approved):

PR #2012, opened 2024-10-02
Run E2E workflow with production environment settings

enescakir:
We have certain special optimizations, such as freezing, specifically for production. However, as these aren't present in our development environment, we sometimes fail to identify issues before deployment.

Running E2E tests with production environment settings could help us detect some of these issues earlier.

I've also removed the `def self.e2e_test?` helper to avoid future branching for E2E.

This PR only impacts E2E tests, excluding a single feature. We started to download default Ubuntu/Alma Linux images from our internal blob storage for E2E tests. This is because remote repositories aren't always reliable and don't consistently store older versions. This feature has been enabled in E2E for a while now, so we're ready to download these images from our internal blob storage for production as well.


fdr:
I'm actually not sure about the performance implications. Freeze can help for something as voluminous as data, and code is data in ruby, but I think it has more to do with software engineering (to avoid difficult changes in behavior depending on what had run so far) and security reasons.


 https://github.com/jeremyevans/ruby-refrigerator

fdr (approved):
My own unsureness of motive aside, I inherited this from sequel roda stack and believe making it noisier for malicious evals to go by undetected. But it's a good change. We could opt to say nothing about motive if we're not sure. I know I usually suggest including motives in commit messages, but when I'm not sure, I leave it out. I suppose pretty soon we can get the entire multi-faceted story from the author. (It may not be that long a story)

enescakir:
The word 'optimizations' for production was not the right choice. I changed it.

Issue #507, opened 2023-08-25
Docker containers can't resolve DNS addresses


enescakir:
When I tried to run GitHub action that builds docker image on our self-hosted runners, it failed with following error:

> ERROR: failed to solve: ruby:3.2.2-alpine3.17: failed to do request:
Head
"https://registry-1.docker.io/v2/library/ruby/manifests/3.2.2-alpine3.17":
dial tcp: lookup registry-1.docker.io on 192.168.0.1:53: read udp
172.17.0.2:48484->192.168.0.1:53: i/o timeout
https://github.com/ubicloud/ubicloud/actions/runs/5969312670/job/16194950769

Even simple `docker run -it alpine:3:17 ping google.com` can't resolve DNS.

`/etc/resolve.conf` of VmHost and Vm
```
nameserver 127.0.0.53
options edns0 trust-ad
search .
```

`/etc/resolve.conf` of docker container
```
nameserver 192.168.111.192
search .
```

We debug it with @furkansahin.
It has some possible solutions:
- Add DNS config to docker daemon config
    - Add following to `/etc/docker/daemon.json`, then restart docker `sudo systemctl restart docker`
```json
{
    "dns": ["9.9.9.9"]
}
```

- Docker gets `resolve.conf` content from `systemd-resolved` service. We can add additional DNS records to it. Also Digital Ocean has `/etc/systemd/resolved.conf.d/DigitalOcean.conf` config.
```bash
sudo mkdir /etc/systemd/resolved.conf.d
sudo sh -c 'echo "[Resolve]\nDNS=9.9.9.9" > /etc/systemd/resolved.conf.d/Ubicloud.conf'
sudo systemctl restart systemd-resolved.service
```

I added `Ubicloud.conf` to our self-hosted runners. But we need to fix this issue for new virtual machines too. Running docker containers on new created vm without any additional config is good user experience.

fdr:
what does `resolvectl` say?

fdr:
Interesting, so, on the VM, dnsmasq absorbs queries for ipv4 but not for ipv6. May be something to smooth out one way or another.

PR #2016, opened 2024-10-03
Assign private IPv4 address to Dnsmasq

furkansahin:
We had reserved the first 4 addresses from the private subnet for our own usage. They sit still and now we need one of them. Ubicloud VMs are totally fine using IPv6 for dns queries, however, the container landscape is not like that... The docker containers running on these VMs prefer (have to) use IPv4. This is why our dnsmasq based dns solution tend to break for containers. Here, in this commit, we are assigning the second address from the private subnet to dns. AWS does the same thing. More information available here:
https://docs.aws.amazon.com/vpc/latest/userguide/subnet-sizing.html

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -43,10 +43,10 @@ class VmSetup
     @vp ||= VmPath.new(@vm_name)
   end
 
-  def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices, boot_image)
-    cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image)
+  def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices, boot_image, dns_ipv4)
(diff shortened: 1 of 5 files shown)


velioglu (approved):

enescakir:
We can close this issue after deploying it. https://github.com/ubicloud/ubicloud/issues/507

rhizome/host/lib/vm_setup.rb
```diff
@@ -268,6 +268,9 @@ def setup_taps_6(gua, nics)
     nics.each do |nic|
       r "ip -n #{q_vm} addr replace #{guest_intrusion.shellescape} dev #{nic.tap}"
 
+      dns_ipv4 = NetAddr::IPv4Net.parse(nic.net4.split("/").first + "/26").nth(2).to_s
```
fdr, line 271: Seems a little weird to hard code as /26 here in duplication. I'm not sure why you need that to get the second address from the base? Could you even make it a /32?

fdr (commented):
Fundamentally I'm for it, but wondering if that ip addr can be streamlined.

rhizome/host/lib/vm_setup.rb
furkansahin: The address added is actually a /32. Since the subnets we allocate are /26 and I want to uniformly allocate the  second address, I'm finding the base first, and then the second address. Proper way would be to pass it with the prep.json. This was easier to do. I'll refactor this later since now this blocks @velioglu 's PR.

furkansahin: Actually, I refactored it already.

PR #2017, opened 2024-10-03
Remove Dnsmasq hard dependency to avoid VM reboot

furkansahin:
Using Requires together with After in systemd definition causes VM service to restart whenever we restart dnsmasq service, too. Replacing with Want still keeps the initial dependency of spin up dnsmasq before VM but restarts won't cause the VM restart as well. This way, we can add, remove arguments properly to dnsmasq; such as custom hostnames.


fdr (approved):
Nice, thought we already did something about that, but no time like the present.

PR #2023, opened 2024-10-04
Use local ip instead of localhost for cache endpoints

velioglu:
To access the local cache server on port 51123, localhost was used. It was ok to use that to access the server via processes running on the VM itself. Though, processes running in a container can not access the cache server with localhost. So, using local ip address instead to allow processes from both VM and container to access cache server.


prog/vm/github_runner.rb
```diff
@@ -235,6 +235,9 @@ def setup_info
   label def setup_forked_runner
     tarball_uri = (label_data["arch"] == "arm64") ? Config.github_cache_forked_runner_tarball_uri_arm64 : Config.github_cache_forked_runner_tarball_uri
 
+    ip_json = JSON.parse(vm.sshable.cmd("ip -j -4 addr show scope global ens3"))
```
enescakir, line 238: Should we use the data from controlplane such as `vm.nics.first.private_ipv4`?

enescakir (approved):

PR #2027, opened 2024-10-04
Fix the condition check during installation destroy

enescakir:
We ensure that the installation has no repositories or runners before destroy it.

I attempted to handle this check on the query side and used `.exists` with the dataset. However, it doesn't work as expected. Instead of returning a boolean, it returns `Sequel::SQL::PlaceholderLiteralString`. The `if` statement always evaluates as true because it's an object. I replaced it with `empty?`.

### DIFF
--- a/prog/github/destroy_github_installation.rb
+++ b/prog/github/destroy_github_installation.rb
@@ -36,8 +36,8 @@ class Prog::Github::DestroyGithubInstallation < Prog::Base
   end
 
   label def destroy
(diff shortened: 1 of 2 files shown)


furkansahin (approved):

fdr (approved):
That's gotten me a few times, though in this case it may suggest a dataset interesting enough that real queries should be considered to catch it in test, even with a slowdown. Alternatively you can factor the dataset to the method and test it in integration once, and then use that method as an injection point for mocking.

PR #2029, opened 2024-10-07
Add Load Balancer custom hostname support

furkansahin:
## Add Load Balancer custom hostname support
Before this commit, we would use the name of the lb, the first 5 chars
of the subnet ubid and .lb.ubicloud.com as the hostname of the
load balancer. With this commit, we start allowing first level
integrating customers (such as AI service) to specify their own dns zone
and the hostname. This way, our load balancers can have nicer names for
internal services.


prog/vnet/load_balancer_nexus.rb
```diff
@@ -9,7 +9,7 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
 
   def self.assemble(private_subnet_id, name: nil, algorithm: "round_robin", src_port: nil, dst_port: nil,
     health_check_endpoint: "/up", health_check_interval: 30, health_check_timeout: 15,
-    health_check_up_threshold: 3, health_check_down_threshold: 2, health_check_protocol: "http")
+    health_check_up_threshold: 3, health_check_down_threshold: 2, health_check_protocol: "http", custom_hostname: nil, custom_hostname_dns_zone_id: nil)
```
bsatzger, line 12: should we add a check that the `custom_hostname` and the provided dns zone match? I assume if the dns zone name was `ai.ubicloud.com` and the hostname was `host1.ai.ubicloud.com` all is good but if dns zone was `lb.ubicloud.com` and hostname  `host1.ai.ubicloud.com` it would be an issue. Or should we only provide the prefix here `host1` together with the dns zone and the full hostname would then be created as  `"#{custom_hostname_prefix}.#{dns_zone.name}"`?

furkansahin: That makes sense, I have updated as requested, please check it again.

bsatzger (approved):

fdr (approved):
today, for internal services on the ubicloud zone, tomorrow, the world (okay, not *actually* tomorrow)

PR #2034, opened 2024-10-08
Disable 4x storage for PG instances in the US region

furkansahin:


enescakir:
Should we write a few sentences explaining our reasoning for future reference?

byucesoy:
We also need to update UI accordingly and hide the 4TB options, however it requires digging into javascript to add/remove options dynamically. @furkansahin do you plan to do that as well? I initially put disabling 4TB options to my plate but then stopped working on them after seeing this PR. I can take it over if you want.

PR #2036, opened 2024-10-08
Apply firewall rules to private subnet addresses as well

furkansahin:
## Limit private_ipX_cidrs to the specific private ips of the resource
We were previously generating the private ip ranges of the whole subnet
for both ipv4 and ipv6. Using these sets later in the rules, we were hot
wiring all of the private subnet and avoiding them to get filtered by
firewall rules.

## Apply firewall rules to the private subnet as well
This commit makes 2 main modifications to apply firewall rules to the
private subnet ranges as well. Together with the previous commit, we are
reorganizing firewall rules to support that.
1. We add new set of rules that would allow connections for the same
lists in the previous commit but this time, when they are used in daddr.
2. Lastly, we get the clover_ephemeral address and apply new rules to
allow everything for them as very first rules.

## Additional reorganization and testing changes for FirewallRules
Here, we make further modifications to reorganize and clean-up the
firewall rules generation file.
1. We remove the sets allowed_ipv4_cidrs/allowed_ipv6_cidrs. These are
not used anymore. Those sets were added with the initial implementation
when our firewall rules did not have ports. However, now, even if we add
a rule without specifying a port, the default behavior is to set all the
ports. Therefore, these sets are not used, the rules that use them are
also removed in the previous commit.
2. We add comments to the necessary places.
3. We further increase the rules to allow pings for all addresses.
4. Modify tests to use the new firewall rules nftables definition.

## Firewalls E2E Tests
Here are the covered scenarios;
- When no firewall rules are set, ipv4 connectivity should fail
- When only public ipv4 of vm2 is allowed:
  - VM2 can connect to VM1
  - VM3 cannot connect to VM1
- When only public ipv6 of vm2 is allowed;
  - VM2 can connect to VM1 using IPv6
  - VM3 cannot connect to VM1 using IPv6
- When only private ipv4 is allowed:
  - VM2 can connect to VM1 using private IPv4
  - VM2 cannot connect to VM1 using public IPv4
- When only private ipv6 of VM2 is allowed:
  - VM2 can connect to VM1 using private IPv6
  - VM2 cannot connect to VM1 using public IPv6

This also includes the unit tests of the E2E tests.

### DIFF
--- a/prog/vnet/update_firewall_rules.rb
+++ b/prog/vnet/update_firewall_rules.rb
@@ -210,11 +210,11 @@ TEMPLATE
   end
 
   def generate_private_ip4_list
-    vm.nics.map { NetAddr::IPv4Net.parse(_1.private_ipv4.network.to_s + "/26").to_s }.join(",")
+    vm.nics.map { _1.private_ipv4.to_s }.join(",")
   end
 
   def generate_private_ip6_list
-    vm.nics.map { NetAddr::IPv6Net.parse(_1.private_ipv6.network.to_s + "/64").to_s }.join(",")
+    vm.nics.map { _1.private_ipv6.to_s }.join(",")
   end
 
   def subdivide_network(net)
--- a/prog/vnet/update_firewall_rules.rb
+++ b/prog/vnet/update_firewall_rules.rb
@@ -7,8 +7,6 @@ class Prog::Vnet::UpdateFirewallRules < Prog::Base
 
   label def update_firewall_rules
     rules = vm.firewalls.map(&:firewall_rules).flatten
-    allowed_ingress_ip4 = NetAddr.summ_IPv4Net(rules.select { !_1.ip6? && !_1.port_range }.map { _1.cidr }).map(&:to_s)
-    allowed_ingress_ip6 = NetAddr.summ_IPv6Net(rules.select { _1.ip6? && !_1.port_range }.map { _1.cidr }).map(&:to_s)
     allowed_ingress_ip4_port_set = consolidate_rules(rules.select { !_1.ip6? && _1.port_range })
     allowed_ingress_ip6_port_set = consolidate_rules(rules.select { _1.ip6? && _1.port_range })
     guest_ephemeral, clover_ephemeral = subdivide_network(vm.ephemeral_net6).map(&:to_s)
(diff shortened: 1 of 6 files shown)


fdr (commented):
feels like.....lot going on in the first commit, did you opportunistically clarify some stuff while also doing the key work for this?

fdr (commented):
pretty neat e2e test there, I have to read it more carefully, but wanted to recognize the good work on going outside the usual boundaries and getting an integration test in

furkansahin:
> feels like.....lot going on in the first commit, did you opportunistically clarify some stuff while also doing the key work for this?

Yeap, because otherwise it was a difficult time to remember all the rules. I did not want to go through the same again, in future.

fdr:
> > feels like.....lot going on in the first commit, did you opportunistically clarify some stuff while also doing the key work for this?
> 
> Yeap, because otherwise it was a difficult time to remember all the rules. I did not want to go through the same again, in future.

could you split the clarifications from the mechanical change? as-is the amount of hunks in the diff leave me wondering what's new and what's clarified.

prog/vnet/update_firewall_rules.rb
```diff
@@ -86,17 +86,26 @@ class Prog::Vnet::UpdateFirewallRules < Prog::Base
...
     ip6 daddr @globally_blocked_ipv6s drop
     ip saddr @private_ipv4_cidrs ct state established,related,new counter accept
-    ip daddr @private_ipv4_cidrs ct state established,related counter accept
+
+    # If we are using clover_ephemeral, that means we are using ipsec. We need
+    # to allow traffic for the private communication and block via firewall
+    # rules through @allowed_ipv4_port_tuple and @allowed_ipv6_port_tuple in the
+    # next section of rules.
+    ip6 daddr #{clover_ephemeral} ct state established,related,new counter accept
+    ip6 saddr #{clover_ephemeral} ct state established,related,new counter accept
```
fdr, line 94: can we simplify? IPsec isn't a connection tracked thing, it's not TCP. I'm actually surprised this does anything?

prog/vnet/update_firewall_rules.rb
```diff
@@ -79,12 +65,25 @@ class Prog::Vnet::UpdateFirewallRules < Prog::Base
 
   chain forward_ingress {
     type filter hook forward priority filter; policy drop;
+
+    # Offload to ubi_flowtable. This is used to offload already filtered
```
fdr, line 69: side note, not for now: I wonder if ipsec can be flow-tabled, to any positive effect.

spec/prog/vnet/update_firewall_rules_spec.rb
```diff
@@ -67,14 +55,14 @@
...
-      10.0.0.0/26
+      10.0.0.0/32
     }
   }
 
   set private_ipv6_cidrs {
     type ipv6_addr
     flags interval
-    elements = { fd00::/64 }
+    elements = { fd00::1/128 }
```
fdr, line 65: Interesting, what's going on here? May benefit from its own commit to explain this tiny change in constants

fdr (commented):
some questions, mostly around how ipsec relates to connection tracking, and some changes to constants.

spec/prog/vnet/update_firewall_rules_spec.rb
furkansahin: I made it the very first commit. Please re-check

fdr (approved):
I found some of the commit messages harder to follow, in the end I found the cumulative diff maybe told me a bit more. But it's a detailed change, watch out for regressions. You will need to check and backport a rule for people to give them the old perimeter model of things, where private traffic had special exception from firewalling.

PR #2045, opened 2024-10-10
Lantern backend

byucesoy:
**Add migration file for adding lantern as new PostgreSQL flavor**

**Add lantern as new PostgreSQL flavor**

**Pick appropriate base image for Lantern PostgreSQL**
For Lantern, we are generating a new base image that comes with pgvector that
is compiled from source along with lantern and lantern_extras extensions.

**Configure Postgres GUCs for Lantern**
Lantern team requested us to set these GUCs for best out of the box Lantern
experience.

**Add billing rates for Lantern PostgreSQL**
These billing rates are placeholder until we decide on the final pricing.

**Send mail to Lantern team upon new Lantern instance creation**
Lantern team is primary contact point for the success of new Lantern users.
They might need to reach out to users to provide support and potentially for
running e-mail campaigns. For these purposes, we share contact information and
basic instange information with Lantern team.

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -21,51 +21,51 @@ class PostgresServer < Sequel::Model
 
   def configure_hash
     configs = {
-      listen_addresses: "'*'",
-      max_connections: (vm.mem_gib * 25).to_s,
-      superuser_reserved_connections: "3",
-      shared_buffers: "#{vm.mem_gib * 1024 / 4}MB",
-      work_mem: "#{vm.mem_gib / 8}MB",
-      maintenance_work_mem: "#{vm.mem_gib * 1024 / 16}MB",
-      max_parallel_workers: "4",
-      max_parallel_workers_per_gather: "2",
-      max_parallel_maintenance_workers: "2",
-      min_wal_size: "80MB",
-      max_wal_size: "5GB",
-      random_page_cost: "1.1",
-      effective_cache_size: "#{vm.mem_gib * 1024 * 3 / 4}MB",
-      effective_io_concurrency: "200",
-      tcp_keepalives_count: "4",
-      tcp_keepalives_idle: "2",
-      tcp_keepalives_interval: "2",
-      ssl: "on",
-      ssl_min_protocol_version: "TLSv1.3",
-      ssl_ca_file: "'/etc/ssl/certs/ca.crt'",
-      ssl_cert_file: "'/etc/ssl/certs/server.crt'",
-      ssl_key_file: "'/etc/ssl/certs/server.key'",
-      log_timezone: "'UTC'",
-      log_directory: "'pg_log'",
-      log_filename: "'postgresql.log'",
-      log_truncate_on_rotation: "true",
-      logging_collector: "on",
-      timezone: "'UTC'",
-      lc_messages: "'C.UTF-8'",
-      lc_monetary: "'C.UTF-8'",
-      lc_numeric: "'C.UTF-8'",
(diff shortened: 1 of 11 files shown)


model/postgres/postgres_server.rb
```diff
@@ -58,6 +58,14 @@ def configure_hash
 
     if resource.flavor == PostgresResource::Flavor::PARADEDB
       configs[:shared_preload_libraries] = "'pg_cron,pg_stat_statements,pg_analytics,pg_search'"
+    elsif resource.flavor == PostgresResource::Flavor::LANTERN
+      configs[:shared_preload_libraries] = "'pg_cron,pg_stat_statements,lantern_extras'"
+      configs[:"lantern.external_index_host"] = "'external-indexing.cloud.lantern.dev'"
+      configs[:"lantern.external_index_port"] = "443"
+      configs[:"lantern.external_index_secure"] = "true"
```
enescakir, line 65: Do we have any specific reason for keeping the keys as symbols? They'll be converted to strings when generating JSON anyway.

routes/common/postgres_helper.rb
```diff
@@ -259,10 +259,12 @@ def failover
   end
 
   def send_notification_mail_to_partners(resource, user_email)
-    if resource.flavor == PostgresResource::Flavor::PARADEDB
-      Util.send_email(Config.postgres_paradedb_notification_email, "New ParadeDB Postgres database has been created.",
-        greeting: "Hello ParadeDB team,",
-        body: ["New ParadeDB Postgres database has been created.",
+    if [PostgresResource::Flavor::PARADEDB, PostgresResource::Flavor::LANTERN].include?(resource.flavor)
```
enescakir, line 262: maybe
```suggestion
    unless resource.flavor == PostgresResource::Flavor::STANDARD
```

routes/common/postgres_helper.rb
```diff
@@ -259,10 +259,12 @@ def failover
...
+      email = Config.send(:"postgres_#{resource.flavor}_notification_email")
+      flavor_name = resource.flavor.capitalize
+      Util.send_email(email, "New #{flavor_name} Postgres database has been created.",
+        greeting: "Hello #{flavor_name} team,",
+        body: ["New #{flavor_name} Postgres database has been created.",
           "ID: #{resource.ubid}",
```
enescakir, line 268: We might add the boot image version too

enescakir (approved):

routes/common/postgres_helper.rb
byucesoy: I wanted this list to be explicit, so that if/when we add a new flavor in the future, we wouldn't automatically try to send a mail.

routes/common/postgres_helper.rb
byucesoy: This is is bit tricky. We don't know which version will be used at this point. It is decided by allocator and host availability.

enescakir: Fair point

model/postgres/postgres_server.rb
byucesoy: No specific reason. I used symbols because existing configs were already symbols. I'll change them to strings.

enescakir: Just to confirm, we aren't accessing this config hash anywhere in the control plane, correct? It's solely for the data plane? If not, we'll need to update the locations where it's accessed as well.

byucesoy: No, we don't access it in the control plane. We only need to update some tests.

PR #2051, opened 2024-10-11
Enable tool support for inference endpoints and switch to ai.ubicloud.com DNS zone

bsatzger:
**Update AI base image to 20241011.1.0**
Job producing the image:
https://github.com/ubicloud/ai-images/actions/runs/11288392843
That image contains a version of vLLM that supports tool support for Llama 3.1+3.2.

**Enable tool support for llama 3.2 3b**
This enables tool support, which was recently added to vLLM in https://github.com/vllm-project/vllm/commit/344cd2b6f4c22bf278cff96066001d216ec1fe82

**Increase quota for public inference endpoints**
With tool support, interaction between client and the inference endpoint tends to be more chatty. Low quotas lead to frequent 429s even for benign interactions.

**Switch to ai.ubicloud.com DNS zone for inference**
Inference endpoints used to get URLs like https://1e00js9jec761m3g36cpbjahky.7msrb.lb.ubicloud.com/.
With this change, multi-tenant inference endpoints would get https://name.ai.ubicloud.com/ while single-tenant endpoints would get https://name-7msrb.ai.ubicloud.com/.

**Enable inference endpoint systemd services**
Enable inference endpoint systemd services by default so that they come up after a restart without control plane interaction.

**Prepare PCI devices during recreate_unpersisted**
This commit fixes an issue for VMs with GPUs whose host is rebooted. Without preparing PCI devices again, the VM won't be able to reboot properly.


prog/ai/inference_endpoint_nexus.rb
```diff
@@ -51,7 +51,13 @@ def self.assemble(project_id:, location:, boot_image:, name:, vm_size:, storage_
       firewall = internal_project.firewalls_dataset.where(location: location).where(Sequel[:firewall][:name] => "inference-endpoint-firewall").first
       fail "No firewall named 'inference-endpoint-firewall' configured for inference endpoints in #{location}" unless firewall
       subnet_s = Prog::Vnet::SubnetNexus.assemble(internal_project.id, name: ubid.to_s, location: location, firewall_id: firewall.id)
-      lb_s = Prog::Vnet::LoadBalancerNexus.assemble(subnet_s.id, name: ubid.to_s, src_port: 443, dst_port: 8443, health_check_endpoint: "/up", health_check_protocol: "https")
+
+      custom_dns_zone = DnsZone.where(project_id: Config.inference_endpoint_service_project_id).where(name: "ai.ubicloud.com").first
+      custom_hostname_prefix = if custom_dns_zone
+        name + (is_public ? "" : "-#{ubid.to_s[-5...]}")
```
furkansahin, line 57: nit: if it's private, maybe we can just use regular lb hostname?

furkansahin (approved):

prog/ai/inference_endpoint_nexus.rb
bsatzger: there's this problem: the name of the load balancer currently is the ubid of the inference endpoint. So the regular lb hostname is ugly, something like `1eprwnxxm9wt7gyhbm9az8sq1y.ycx1b.lb.ubicloud.com`. We cannot use `name.ycx1b.lb.ubicloud.com` because all load balancers for inference endpoints are in an internal project and we mandate that names in a project are unique. That means that a customer cannot use `name` if another customer is using that name already, even though these two load balancers would get different hostnames as they each are in a different subnet.

rhizome/inference_endpoint/lib/replica_setup.rb
```diff
@@ -18,8 +18,8 @@ def prep(inference_engine:, inference_engine_params:, model:, replica_ubid:, ssl
   end
 
   def start_systemd_units
-    r "systemctl start lb-cert-download.timer"
-    r "systemctl start inference-engine.service"
+    r "systemctl enable --now lb-cert-download.timer"
+    r "systemctl enable --now inference-engine.service"
```
enescakir, line 22: I didn't know about the `--now` flag for `enable`. I used to run `start` and `enable` separately. That's neat 🚀

enescakir (approved):

PR #2054, opened 2024-10-14
Subnet Peering

furkansahin:
Basic subnet peering for non conflicting subnets. It's as easy as creating a PeeredSubnet entity and performing ps.incr_add_new_nic in only 1 of the subnets. The rest is auto handled.


fdr (commented):
does nic deletion exist, and therefore unpeering, to remove extra xfrm states? 

I like that topologically defined uniqueness constraint

fdr:
> I like that topologically defined uniqueness constraint (with `<`)

...but surely you need a compound unique constraint too? and/or a primary key? I suspect we'd probably want a surrogate key here, unless you think there's a good reason to make exception and refer to everything by natural key. But I figured those might fall in the "itsaprototype" category.

furkansahin:
This needs more tests, I'll write them all soon, but the PR is in a reviewable state as is for the overall design.

model/private_subnet.rb
```diff
@@ -24,6 +24,16 @@ def hyper_tag_name(project)
...
   include Authorization::TaggableMethods
 
+  def connected_subnets
+    PrivateSubnet.where(
+      id: DB[:connected_subnet].where(subnet_id_1: id).or(subnet_id_2: id).select(Sequel.case({{subnet_id_1: id} => :subnet_id_2}, :subnet_id_1)).all.map(&:values).flatten
+    ).all
+  end
+
+  def all_nics
+    nics + connected_subnets.map(&:nics).flatten
```
enescakir, line 34: You can use `flat_map(&:nics)` instead of `map(&:nics).flatten`

migrate/20241014_add_connected_subnet.rb
```diff
@@ -0,0 +1,17 @@
...
+  change do
+    create_table :connected_subnet do
+      column :id, :uuid, primary_key: true, null: false
+      foreign_key :subnet_id_1, :private_subnet, type: :uuid, null: false
+      foreign_key :subnet_id_2, :private_subnet, type: :uuid, null: false
+
+      # Ensure no duplicate pairs regardless of order
+      constraint(:unique_subnet_pair) { (subnet_id_1 < subnet_id_2) }
+      unique [:subnet_id_1, :subnet_id_2]
+      index [:subnet_id_1]
```
fdr, line 13: you probably don't need this one, because the compound index can handle predicates on the left-most-side. It's not quite the same physical impact, since the locality of scanning on the compound index would be worse, but would make updates a bit cheaper.

model/private_subnet.rb
```diff
@@ -90,4 +100,37 @@ def random_private_ipv6
...
+    ConnectedSubnet.where(subnet_id_1: small_id_ps.id, subnet_id_2: large_id_ps.id).all.map(&:destroy)
+    subnet.incr_refresh_keys
+    incr_refresh_keys
+  end
+
+  def create_tunnels(nics, src_nic)
+    nics.each do |dst_nic|
+      next if src_nic == dst_nic
+      IpsecTunnel.create_with_id(src_nic_id: src_nic.id, dst_nic_id: dst_nic.id) unless IpsecTunnel[src_nic_id: src_nic.id, dst_nic_id: dst_nic.id]
+      IpsecTunnel.create_with_id(src_nic_id: dst_nic.id, dst_nic_id: src_nic.id) unless IpsecTunnel[src_nic_id: dst_nic.id, dst_nic_id: src_nic.id]
```
fdr, line 129: This is a bit smelly, considering that create_with_id is still generally entitled to fail with a unique constraint violation anyway. One option might be "insert...on conflict..."

model/private_subnet.rb
```diff
@@ -30,6 +30,16 @@ def hyper_tag_name(project)
...
 
+  def connected_subnets
+    PrivateSubnet.where(
+      id: DB[:connected_subnet].where(subnet_id_1: id).or(subnet_id_2: id).select(Sequel.case({{subnet_id_1: id} => :subnet_id_2}, :subnet_id_1)).all.map(&:values).flatten
+    ).all
+  end
+
+  def all_nics
+    nics + connected_subnets.flat_map(&:nics)
+  end
```
fdr, line 41: feels like this could be a join rather than detoasting a bunch of stuff. also, `flat_map` rather than map + flatten.

model/nic.rb
```diff
@@ -27,4 +27,8 @@ def self.ubid_to_name(ubid)
   def ubid_to_tap_name
     ubid.to_s[0..9]
   end
+
+  def unlock
+    Semaphore.where(strand_id: strand.id, name: "lock").all.map(&:destroy)
```
fdr, line 32: Does unlock ever has its value used? Isn't strand the strand for this nic? seems like two simplifications may be possible:

```ruby
 Semaphore.where(strand_id: id, name: "lock").delete
```

I had this as each(&:destroy) at first, but do we really want `DeletedRecord` entries for even semaphores? If not, straight `DELETE` might be fine, and much faster.

fdr (approved):
caveat: I haven't had time to think adequately about the concurrency algorithm. And I'm not sure about adding this as an orthogonal feature in the UI in the medium-term. As I mentioned earlier, if possible, I'd like to try de-orthogonalizing it as much as possible (to couple the tunneling, firewall, and dns).

But I also see this patch is prone to conflicts and given the preliminary situation, with regard to customer exposure, would like to get it in.

I suppose the UI could hide behind a feature flag.

model/private_subnet.rb
furkansahin: This was used as is before, too. I only changed its place, I'll create an item to look into it later.

PR #2059, opened 2024-10-15
Allow PrivateSubnet.[] to lookup by ubid

jeremyevans:
If PrivateSubnet.[] is given a UBID or UBID-looking string, convert it to a UUID for the lookup.  Otherwise, perform a standard lookup.


fdr:
Prototype right? I so imagine you wanted a draft. But yeah, something like that, but as a base class or include to hit all ubid'd models.

btw, bad thing about ubid.rb I just remembered:

```ruby
  TYPE2CLASS = constants.select { _1.start_with?("TYPE_") }.reject { _1.to_s == "TYPE_ETC" }
    .map { [const_get(_1), Object.const_get(camelize(_1.to_s).to_s)] }.to_h
```

The moment you touch it, it faults in a huge number of constants to build this dictionary. Okay for production, not so great for running a single test.

jeremyevans:
Ah, didn't catch that you wanted this for all models that support ubid.  Is that the same with the `#inspect` change you wanted?

Looks like the test failures are all rubocop related.  Probably need to add rubocop  as a pre-commit hook so I don't miss that in the future.

Fixing `TYPE2CLASS` to work with autoloading should be possible without too much work.

enescakir:
> Looks like the test failures are all rubocop related. Probably need to add rubocop as a pre-commit hook so I don't miss that in the future.

`rubocop` and `erb-formatter` might be good candidates for a pre-commit hook

jeremyevans:
> We have `ResourceMethods` that include some ubid helpers and included by most of the models
> 
> https://github.com/ubicloud/ubicloud/blob/388afff19678316739cdddf69e94a508af6d20d1/model.rb#L44

Thank you, I agree that looks like the best place for this code.

jeremyevans:
@fdr Can you take another look at this now that I think the issues have been fixed?

fdr (approved):

enescakir (approved):
Since it's for all models not just for PrivateSubnet, the PR title might be little outdated

jeremyevans:
> Since it's for all models not just for PrivateSubnet, the PR title might be little outdated

Good point, I have updated the title.

PR #2060, opened 2024-10-15
Add Disk IO to VM Host Healthchecks

aamederen:
The health checks in `VmHost#check_pulse` and `HostNexus#available?`
were simple SSH access attempts, but this was not sufficient for
detecting issues in the disks. This change adds a simple disk IO
to the health check commands.

The disk I/O added for the health check is a simple read from
/dev/zero into a random file in /tmp. This would be enough to detect
any obvious failures related to the disks. Note that, in cases of
multiple disks are mounted into the file system (`/`) via raid, this
approach may not detect a single disk failure as long as the system
works.

Note: Through the codebase, health checks for the monitor (`check_pulse`)
and progs (`available?`) are duplicated, which is also the case for
VM host. We can argue about unifying the health check and availability
detection in general as a next step. Still, I wanted to make some
deduplication by extracting the shell command to be used for health
check into a `VmHost` constant.

Future work: While some disk i/o would be enough to check for obvious problems,
as a future work we can consider monitoring SMART bits, preferably
with less frequency and less severity.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -22,6 +22,12 @@ class VmHost < Sequel::Model
   include HealthMonitorMethods
   semaphore :checkup, :reboot, :destroy
 
+  SSH_HEALTH_CHECK_COMMAND = <<EOS
+rm -f /tmp/ubiiotest-*; \
+ZFILE=/tmp/ubiiotest-$RANDOM; \
+timeout 10 dd if=/dev/zero of=$ZFILE bs=1M count=10 && echo "e5b844cc57f57094ea4585e235f36c78c1cd222262bb89d53c94dcb4d6b3e55d  $ZFILE" | sha256sum -c -;
+EOS
+
   def host_prefix
     net6.netmask.prefix_len
   end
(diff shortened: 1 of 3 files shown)


aamederen:

model/vm_host.rb
```diff
@@ -22,6 +22,8 @@ class VmHost < Sequel::Model
   include HealthMonitorMethods
   semaphore :checkup, :reboot, :destroy
 
+  SSH_HEALTH_CHECK_COMMAND = "RFILE=/tmp/ubiiotest-$RANDOM;timeout 10 dd if=/dev/urandom of=$RFILE bs=1M count=10;rm $RFILE"
```
jeremyevans, line 25: I recommend we check that we can read back what we wrote.  We can write a 10MB file from `/dev/zero`, and then check the SHA256 of the file is `e5b844cc57f57094ea4585e235f36c78c1cd222262bb89d53c94dcb4d6b3e55d`.

jeremyevans (commented):
Looks mostly good to me, but I think we should change the command slightly.

model/vm_host.rb
aamederen: I had given a thought to write something known (like a hard-coded string) to be able to read back, but that could be too short. Haven't considered the zeroes approach before. I have to admit, it sounds slightly awkward to write just zeroes (no real reason, just my hunch), but since you said it, let's go with it.

jeremyevans (approved):
I tested the command and it works as expected.  Not sure I'm qualified to review the control plane parts, but the changes look fine to me.

fdr:
I'm a little bit puzzled as to what problem this seeks to solve, mostly because it deals in terms of file systems, e.g. /tmp, and hosts have multiple disks. Would we really not get what we would want out of fishing around in procfs or sysfs or nvme cli or smartctl?

PR #2062, opened 2024-10-15
Billing for inference endpoints

bsatzger:
**Add billed_by to billing rates**
We bill some resources based on amount, others on duration. For 'VmCores', for instance, the core counts are stored in the amount column of billing records and charges are based on duration. 10 minutes of standard-4 vm usage would be calculated as `10 (duration) x 2 (amount) x rate_for_standard_core`. 'GitHubRunnerMinutes' billing records, on the other hand, store used minutes in the amount column and billing is based on that amount. For instance, 10 minutes of standard-4 runner usage would be calculated as `1 (duration) x 10 (amount) x rate_for_standard_4_runner_minute`. Currently, 'GitHubRunnerMinutes' is the only category we bill by amount, but 'InferenceTokens' are planned to be billed in the same way.

**Introduce billing for inference endpoints**
This commit introduces a new `InferenceTokens` billing resource type. Billing rates are added that define how much a user is billed for each inference token used. Billing depends on the amount of tokens and the type of model used for inference.

**Create billing records for inference endpoints**
We periodically collect information about inference tokens used by customers. This is done by periodically sending a request to inference gateway, which responds with tokens used since the last data exchange, at per project granularity. Token consumption data collected that way are then turned into `InferenceTokens` billing records.


model/billing_record.rb
```diff
@@ -14,20 +14,14 @@ def active
...
-    # data points before undertaking a more significant refactor.
-    return 1 if billing_rate["resource_type"] == "GitHubRunnerMinutes"
+    # Billing logic differs based on the resource type: some are billed by duration, others
+    # by amount. For 'VmCores' billing records, the core counts are stored in the amount
+    # column and charges are based on duration. For example, 10 minutes of standard-4 vm
+    # usage would be calculated as `10 (duration) x 2 (amount) x rate_for_standard_core`.
+    # 'GitHubRunnerMinutes' and 'InferenceTokens' records store the used minutes or tokens
+    # in the 'amount' column, and billing is based on that amount.
+    # For records billed by amount, the duration is always set to 1.
+    return 1 if billing_rate["billed_by"] == "amount"
```
enescakir, line 24: I like this `amount` vs `minutes` improvement. 🚀 cc @byucesoy Since he knows our billing system well too and we discussed it before.

enescakir: It could be a separate commit with more detailed commit message, though: introduce `billing_by` column, and introduce inference tokens billing rates.

prog/ai/inference_endpoint_replica_nexus.rb
```diff
@@ -154,9 +154,46 @@ def ping_gateway
...
+    end_time = begin_time + 24 * 60 * 60
+
+    project_usage.each do |usage|
+      tokens = usage["prompt_token_count"] + usage["completion_token_count"]
+      next if tokens.zero?
+      project = Project.from_ubid(usage["ubid"])
+
+      begin
+        begin_time = Time.now.to_date.to_time
+        end_time = begin_time + 24 * 60 * 60
```
enescakir, line 175: We defined `begin_time` and `end_time` seven lines above

prog/ai/inference_endpoint_replica_nexus.rb
```diff
@@ -154,9 +154,46 @@ def ping_gateway
...
+  end
+
+  def update_billing_records(project_usage)
+    rate = BillingRate.from_resource_properties("InferenceTokens", inference_endpoint.model_name, "global")
+    return if rate.nil? || rate["unit_price"].zero?
+    rate_id = rate["id"]
+    begin_time = Time.now.to_date.to_time
+    end_time = begin_time + 24 * 60 * 60
+
+    project_usage.each do |usage|
```
enescakir, line 168: When we receive this data from the gateway, does it reset the token counts? Otherwise, we might repeatedly count tokens that have already been accounted for.

prog/ai/inference_endpoint_replica_nexus.rb
bsatzger: sorry!

model/billing_record.rb
bsatzger: Split into separate commits.

prog/ai/inference_endpoint_replica_nexus.rb
```diff
@@ -154,9 +154,44 @@ def ping_gateway
...
-    # project_usage is a list of the following format:
-    # [{"ubid":"aprojectubid","request_count":0,"prompt_token_count":0,"completion_token_count":0}, ...]
-    # TODO: produce billing records for public endpoints based on project_usage
     Clog.emit("Successfully pinged inference gateway.") { {inference_endpoint: inference_endpoint.ubid, replica: inference_endpoint_replica.ubid, project_usage: project_usage} }
+    update_billing_records(project_usage)
+  end
+
+  def update_billing_records(project_usage)
+    rate = BillingRate.from_resource_properties("InferenceTokens", inference_endpoint.model_name, "global")
+    return if rate.nil? || rate["unit_price"].zero?
```
enescakir, line 163: Should we fail if we forget to add a billing rate for a model, instead of simply ignoring it?

prog/ai/inference_endpoint_replica_nexus.rb
```diff
@@ -154,9 +154,44 @@ def ping_gateway
...
+            project_id: project.id,
+            resource_id: inference_endpoint.id,
+            resource_name: "Inference tokens #{inference_endpoint.model_name} #{begin_time.strftime("%Y-%m-%d")}",
+            billing_rate_id: rate_id,
+            span: Sequel.pg_range(begin_time...end_time),
+            amount: tokens
+          )
+        end
+      rescue Sequel::Error => ex
+        Clog.emit("Failed to update billing record") { {billing_record_update_error: {project_ubid: project.ubid, model_name: inference_endpoint.model_name, replica_ubid: inference_endpoint_replica.ubid, tokens: tokens, exception: Util.exception_to_hash(ex)}} }
```
enescakir, line 193: Should we reraise the exception?

enescakir (approved):
LGTM in general. I just post two small questions but they are not blocker for me

prog/ai/inference_endpoint_replica_nexus.rb
bsatzger: Removed the `rate.nil?` check.

prog/ai/inference_endpoint_replica_nexus.rb
bsatzger: If we received usage data for 10 projects and fail to update the billing record for the first one we'd still like to update billing records for the remaining 9.

enescakir (approved):

PR #2064, opened 2024-10-15
Make small touches to DEVELOPERS.md

aamederen:
I felt like the developers document had some updates needed, mostly minor stuff, about versions, sentence structure, phrasing, etc. While rewriting, I got help from ChatGPT obviously. It also lacked the steps to prepare a host, so I added some high level steps for that as well.


DEVELOPERS.md
```diff
@@ -1,11 +1,10 @@
 # Clover
 
-Clover is a control plane and web console program for managing virtual
-machines and other programs.
+Clover is the codename for Ubicloud's software, which consists of the control plane, data plane, and web console program for managing virtual machines and other applications.
```
enescakir, line 3: I think `clover` is the only controlplane and web console. `rhizome` is the dataplane. cc @fdr

DEVELOPERS.md
```diff
@@ -51,79 +50,97 @@ three general steps:
...
+[There is documentation listing the commands you can use for each
 platform](https://github.com/rbenv/ruby-build/wiki#suggested-build-environment)
 (e.g. Macintosh Homebrew, or Ubuntu).
 
 After that, install Ruby.  `asdf` will consult the `.tool-versions`
 file to select the version.
 
-    asdf install ruby
+```sh
+$ asdf install ruby
```
enescakir, line 85: Can you remove the `$` signs? When the copy button is used, it also copies the `$`, which invalidates the command.

Screenshot 2024-10-15 at 15 28 36

DEVELOPERS.md
```diff
@@ -417,7 +479,54 @@ in. You'll see the "Getting Started" page.
...
 When you change any template file, format them with `erb-formatter`:
 
-    rake linter:erb_formatter
+```sh
+$ rake linter:erb_formatter
+```
+
+### Cloudifying a Host for Development
+
+We show cloudifying a host from Hetzner, but the principles should work everywhere. Make sure that the Hetzner instance has at least one `One additional subnet /29` ordered and `Ubuntu 20.04 LTS base` is installed.
```
enescakir, line 488: Our hosts are `Ubuntu 22.04 LTS base`

DEVELOPERS.md
aamederen: Well, I agree that it makes copy-paste a bit harder, but I think this style makes it more clear what's a shell command vs. what's a Ruby code piece or what's just output, hence I added them consistently everywhere.

Let me pick other thoughts on this.

DEVELOPERS.md
aamederen: Hmm, good point. Maybe it's just me, or maybe we can clarify that a little bit more then.

DEVELOPERS.md
aamederen: Good catch!

DEVELOPERS.md
fdr: yeah I guess that's a bit murky, because clover and rhizome (which is a type of root) are tightly coupled, and in the same repo, and clover uploads rhizome, but is it really a project apart? not so much.

I suppose I consider rhizome part of clover, a component.

fdr: but yes, it is a code name

DEVELOPERS.md
fdr: Here's a Terrible Idea :tm:

```sh
#!/usr/bin/sh
asdf install ruby
```

fdr: It's kinda wordy but maybe if we put it in the first time/once in a while it'll give people the idea.

DEVELOPERS.md
```diff
@@ -152,45 +173,56 @@ Node.js is required to work with frontend tooling such as tailwind-cli.
...
 
-    $ which node
-    /Users/enescakir/.asdf/installs/nodejs/20.2.0/bin/node
-    $ node --version
-    v20.2.0
+```sh
+$ which node
+/home/ubicloud/.asdf/installs/nodejs/20.2.0/bin/node
+$ node --version
+v22.9.0
```
fdr, line 187: Doesn't match the directory just above

fdr, line 187: Doesn't match the directory just above

fdr (commented):
With regard to $ prompt: it adds more to have it when the output is also present, which is useful for the reader to make sure they're on the right track. But not all shell  blocks have output (and that's fine)

DEVELOPERS.md
aamederen: What do you think about using `#!/usr/bin/sh` for _some_ single lines and throw `$` but it for multi-lines with output?

DEVELOPERS.md
aamederen: thanks, fixed it.

DEVELOPERS.md
aamederen: Since we couldn't get a strong signal, I will resolve this for now. @enescakir If you feel strong about it, I'm happy to change.

fdr (approved):

PR #2066, opened 2024-10-15
Add awesome_print for easier inspection

aamederen:
We used awesome_print gem [1] in predecessors of clover, and it proved a bit useful in inspecting objects in repl. This change adds it here too.

Usage:

```ruby
[9] clover-development(main)> s = Strand.first
=> #<Strand @values={:id=>"fea6d185-f903-82ac-96b7-25f5c833f220", :parent_id=>nil, :schedule=>2024-10-15 16:27:26.148985 +0300, :lease=>nil, :prog=>"Vnet::NicNexus", :label=>"wait", :stack=>[{"last_label_changed_at"=>"2024-10-15 14:50:28 +0300"}], :exitval=>nil, :retval=>nil, :try=>0}>
[10] clover-development(main)> ap s
{
           :id => "fea6d185-f903-82ac-96b7-25f5c833f220",
    :parent_id => nil,
     :schedule => 2024-10-15 16:27:26.148985 +0300,
        :lease => nil,
         :prog => "Vnet::NicNexus",
        :label => "wait",
        :stack => [{"last_label_changed_at"=>"2024-10-15 14:50:28 +0300"}],
      :exitval => nil,
       :retval => nil,
          :try => 0
}
=> nil
```

[1] https://github.com/awesome-print/awesome_print

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem "octokit"
 gem "argon2-kdf"
 
 group :development do
+  gem "awesome_print"
   gem "brakeman"
   gem "by", github: "jeremyevans/by", ref: "f022615e367da3a23f012c3ab06ff664fe3776d3"
   gem "erb-formatter", github: "ubicloud/erb-formatter", ref: "a9ff0001a1eb028e2186b222aeb02b07c04f9808"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,6 +46,7 @@ GEM
       ffi-compiler (~> 1.0)
     argon2-kdf (0.2.0)
     ast (2.4.2)
+    awesome_print (1.9.2)
     awrence (1.2.1)
     aws-eventstream (1.3.0)
     aws-partitions (1.983.0)
@@ -364,6 +365,7 @@ DEPENDENCIES
   acme-client (~> 2.0)
   argon2
   argon2-kdf
+  awesome_print
   aws-sdk-s3 (~> 1.167)
   bcrypt_pbkdf
   brakeman
--- a/bin/pry
+++ b/bin/pry
@@ -7,6 +7,10 @@ require_relative "../loader"
 
 require "pry"
 
+if Config.development?
+  require "awesome_print"
+end
+
 def dev_project
   return unless Config.development?
   ac = Account[email: "dev@ubicloud.com"] || Account.create_with_id(email: "dev@ubicloud.com")


fdr:
I like `awesome_print` personally, and would even consider making it the default print for pry. 

But haven't used it in a while because it seemed unmaintained. I suppose very occasionally in the past I had an issue. But why maintain something that basically works, is the question. I even thought of dumping `pry` which went through a period of low maintenance, but `irb` doesn't bind `_ex_` and maybe lacks on or two features I care about, so, I still use `pry`.

I'm most ambivalent about the array subscripting, which I like visually and when working with small arrays, but breaks pasting them back in as a literal, right? That is, however, a knob, maybe we should flip the switch and not subscript arrays.

We might want to bring awesomeprint back in and the remaining question is if it should be imposed as the default printer for pry right away. 

My other concern with bringing in awesomeprint is that it widens the base of code running in production, and vulnerable to supply chain attack, which is especially ungratifying given how little it changes. I think I'm the only one that does this, but if you can believe it, I often check the source code diffs for upgrades, when I process them. They are usually not so large, except the AWS gem, which we ought to get rid of.

fdr:
also ur a real one: "predecessor**s**", not too many are repeat-repeat customers. Sometimes the newer ones lose something, but for no good reason, awesomeprint may be one of these.

fdr:
also I just checked: you can guard it with an environment variable, e.g.

```
heroku run USE_AP ./bin/pry
```

then we can get it pronto, maybe play around with some settings, go all in and make it integrate with `pry` so it prints that way by default, and poll the team on whether to default it.

If we want to mitigate supply chain attack, we can use a fork, kinda like erb-formatter, but without any changes.

aamederen:
@fdr I'm also hesitating with the default, I don't know enough whether it would crash anything, hence I used it this way. 

re - guarding with env: IMO that wouldn't fix the issue completely. If it was a variable, I would probably almost always enable it when using pry for example.

re - forking: The tradeoff is: just another thing you kinda maintain + need to check regularly in case any new feature or bugfix lands on the original source.

My preference (in desc. order):
- Merge it as is
- Merge it only for dev
- Fork and merge

What do you think?

fdr:
Let's merge only for dev then. See what the popular reaction is

aamederen:
@fdr just done that, can you check if that's ok now?

fdr (approved):
Sure, okay, though I thought you might go a little further and make it the default pry printer in development (and turn off array index printing, which breaks paste-back)

PR #2069, opened 2024-10-16
Use require instead of autoload in production and when running all specs

jeremyevans:
This makes `rake spec` operate more similarly to the production environment,
loading all files up front.  Instead of using autoload, this uses require
to load all files in that case.  Using require is simpler than using
autoload, keeping an array of autoloaded constants, and then referencing
each autoloaded constant to force the loading. 

This required some changes:

* Use of `require_relative` in a number of files to specify dependencies
  explicitly instead of relying on autoload to load the dependency
  implicitly.

* Defining empty modules in loader.rb, instead of having loader.rb
  automatically define empty modules via metaprogramming.

* In Sequel models, reference other classes by name, avoiding
  unnecessary autoloads and stale references when Unreloader reloads
  dependencies.

This also changes ubid.rb so it doesn't load every class that can have a ubid.
It has ubid.rb keep a hash of prefixes to class names, and maps to
classes inside a method.  To avoid the performance hit in production
and when you are loading all classes anyway, this has loader.rb
switch back to the previous approach of mapping prefixes to classes,
since you know the classes will no longer change in that case.

This adds a couple rake tasks:

* spec_separate: Run all spec files in separate processes, to check
  that each spec file passes in isolation, and that there are no
  dependencies between the spec files.

* check_separate_requires: Check that each file that can be
  autoloaded by loader.rb can be required separately without
  relying on other autoloaded files already being loaded.

These rake tasks both found issues with missing requires.

Not using autoload when testing and in production is how roda-sequel-stack
is designed to work by default (since https://github.com/jeremyevans/roda-sequel-stack/commit/faf3844bc0982eac8cda4db5786bc2256bff6a53).

Note that this continues to use `Unreloader.autoload`, but if you initialize
`Unreloader` with `autoload: false`, then calls to `Unreloader.autoload`
actually call `require` instead of `autoload` (see https://github.com/jeremyevans/rack-unreloader?tab=readme-ov-file#label-Autoload).

See individual commits for more detail.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -127,10 +127,17 @@ end
 # Specs
 begin
   require "rspec/core/rake_task"
-  ENV["RACK_ENV"] = "test"
-  RSpec::Core::RakeTask.new(:spec)
-  task default: :spec
+  RSpec::Core::RakeTask.new(:_spec)
+  Rake::Task["_spec"].clear_comments
 rescue LoadError
+else
+  desc "Run specs"
+  task "spec" do
+    ENV["RACK_ENV"] = "test"
+    ENV["FORCE_AUTOLOAD"] = "1"
+    Rake::Task["_spec"].invoke
+  end
+  task default: :spec
 end
 
 desc "Run specs in parallel using turbo_tests"
@@ -139,7 +146,7 @@ task "pspec" do
   nproc = `(nproc 2> /dev/null) || sysctl -n hw.logicalcpu`.to_i
 
   # Limit to 6 processes, as higher number results in more time
-  system("bundle", "exec", "turbo_tests", "-n", nproc.clamp(1, 6).to_s)
+  system({"FORCE_AUTOLOAD" => "1"}, "bundle", "exec", "turbo_tests", "-n", nproc.clamp(1, 6).to_s)
 end
 
 # Other
--- a/Rakefile
+++ b/Rakefile
@@ -151,6 +151,12 @@ end
 
 # Other
 
+desc "Check that model files work when required separately"
+task "check_separate_requires" do
+  require "rbconfig"
+  system({"RACK_ENV" => "test", "LOAD_FILES_SEPARATELY_CHECK" => "1"}, RbConfig.ruby, "-r", "./loader", "-e", "")
+end
+
 desc "Annotate Sequel models"
 task "annotate" do
   ENV["RACK_ENV"] = "development"
--- a/Rakefile
+++ b/Rakefile
@@ -157,6 +157,22 @@ task "check_separate_requires" do
   system({"RACK_ENV" => "test", "LOAD_FILES_SEPARATELY_CHECK" => "1"}, RbConfig.ruby, "-r", "./loader", "-e", "")
 end
 
+desc "Run each spec file in a separate process"
+task :spec_separate do
+  require "rbconfig"
+
+  failures = []
+  Dir["spec/**/*_spec.rb"].each do |file|
+    failures << file unless system(RbConfig.ruby, "-S", "rspec", file)
+  end
+
+  if failures.empty?
+    puts "All files passed"
+  else
+    puts "Failures in:", failures
+  end
+end
+
 desc "Annotate Sequel models"
 task "annotate" do
   ENV["RACK_ENV"] = "development"
--- a/clover_api.rb
+++ b/clover_api.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require "roda"
+require_relative "model"
+
 class CloverApi < Roda
   include CloverBase
 
--- a/clover_runtime.rb
+++ b/clover_runtime.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "jwt"
+require "roda"
 
 class CloverRuntime < Roda
   include CloverBase
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
+require "roda"
 require "tilt"
 require "tilt/erubi"
+
+require_relative "model"
 
 class CloverWeb < Roda
   include CloverBase
--- a/lib/authorization.rb
+++ b/lib/authorization.rb
@@ -111,7 +111,7 @@ module Authorization
   module HyperTagMethods
     def self.included(base)
       base.class_eval do
-        many_to_many :projects, join_table: AccessTag.table_name, left_key: :hyper_tag_id, right_key: :project_id
+        many_to_many :projects, join_table: :access_tag, left_key: :hyper_tag_id, right_key: :project_id
       end
     end
 
@@ -154,7 +154,7 @@ module Authorization
   module TaggableMethods
     def self.included(base)
       base.class_eval do
-        many_to_many :applied_access_tags, class: AccessTag, join_table: AppliedTag.table_name, left_key: :tagged_id, right_key: :access_tag_id
+        many_to_many :applied_access_tags, class: :AccessTag, join_table: :applied_tag, left_key: :tagged_id, right_key: :access_tag_id
       end
     end
 
--- a/loader.rb
+++ b/loader.rb
@@ -91,6 +91,21 @@ autoload_normal.call("model", flat: true)
 %w[lib clover.rb clover_web.rb clover_api.rb clover_runtime.rb routes/clover_base.rb routes/clover_error.rb].each { autoload_normal.call(_1) }
 %w[scheduling prog serializers routes/common].each { autoload_normal.call(_1, include_first: true) }
 
+if ENV["LOAD_FILES_SEPARATELY_CHECK"] == "1"
+  files = %w[model lib scheduling prog serializers routes/common].flat_map { Dir["#{_1}/**/*.rb"] }
+  files.concat(%w[clover.rb clover_web.rb clover_api.rb clover_runtime.rb routes/clover_base.rb routes/clover_error.rb])
+  files.each do |file|
+    pid = fork do
+      require_relative file
+      exit(0)
+    rescue LoadError, StandardError => e
+      puts "ERROR: problems loading #{file}: #{e.class}: #{e.message}"
+    end
+    Process.wait(pid)
+  end
+  exit(0)
+end
+
 AUTOLOAD_CONSTANTS.freeze
 
 if force_autoload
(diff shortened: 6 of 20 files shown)


jeremyevans:
As an example of a problem with the current autoload approach, create a file named `model/bad.rb` with the content `raise`.  Run the tests, and they will still pass, since it never attempts to autoload the file.  However, start the application in production mode, and it won't even boot.  With this PR, the tests also fail immediately, matching the production behavior.

I grant you, it's unlikely a problem of that nature makes it through code review.  Still, I think that to the extent possible, we should make the test environment close to the production environment.

enescakir:
> I grant you, it's unlikely a problem of that nature makes it through code review. Still, I think that to the extent possible, we should make the test environment close to the production environment.

In the past, we've been tripped up by differences between the test and production environments a few times, particularly due to freezing issues. So, it's beneficial to keep them as similar as possible. I'm reviewing the PR, but it's slightly beyond my Ruby knowledge. @fdr might have more experience in this area.

fdr (approved):
I'm alright with this, though, I question what the main advantage of losing so much declarative autoloading is. There are two things I do not like about explicit enumeration of dependencies when I can avoid it:

1. When a require is no longer necessary because some references have disappeared, nobody ever figures out they don't need that `require`, so it lives on ~forever, bloating times to run a single test.

2. To keep the code legible, people require more than they need to, often indirectly, e.g. by requiring file (let's say, a hypothetical `api_base.rb`). People will promote a require into such a base file if, say, 80% of files tested would need it, but that does leave the 20%. And this proportion may change. But ~nobody is going to monitor the proportion.

The system as seen previously was neatly adaptive to loading as little as reasonably possible under change.

On the other hand, I don't have as good an instinct as someone that works on the Ruby interpreter on how much autoloading costs us, whether in reliability or overhead, even in the case where it excels, which is running a single test.

fdr:
Also...would production break? why would `bad` ever run, even in production, as it has no constants, and thus cannot be hit with autoload, whether eagerly (as seen in `loader.rb` today)

fdr:
Oh yes I'm sorry, thought it over. But I'd propose a different fix for that particular problem: if all tests are run, run the force-all-symbols code. As a test. The very lazy auto loading is then still active for single tests.

fdr:
> > I grant you, it's unlikely a problem of that nature makes it through code review. Still, I think that to the extent possible, we should make the test environment close to the production environment.
> 
> 
> 
> In the past, we've been tripped up by differences between the test and production environments a few times, particularly due to freezing issues. So, it's beneficial to keep them as similar as possible. I'm reviewing the PR, but it's slightly beyond my Ruby knowledge. @fdr might have more experience in this area.

With regard to freezing, our main problem is partial mocking. Autoload would also break stuff, but we can bypass that by running the production style force code of the entire suite is run (thus incremental loading doesn't help)

jeremyevans:
> Also...would production break? why would `bad.rb` ever run, even in production, as it has no constants, and thus cannot be hit with autoload, whether eagerly (as seen in `loader.rb` today). It would also nominally break the test coverage.

I tested this and production does break, because it loads the file.  Without this, tests do not .  It doesn't matter whether the file defines the constant (you cannot know that before loading the file), it only matters that Unreloader thinks `bad.rb` would define the `Bad` constant, so it would setup the autoload for `Bad` to load `bad.rb`, and then would later reference `Bad`, loading `bad.rb`.

You are correct that this would be caught by code coverage.

>  But I'd propose a different fix for that particular problem: if all tests are run, run the force-all-symbols code. As a test. The very lazy auto loading is then still active for single tests.

That's how it works in this PR.  `NO_AUTOLOAD` is set when running all tests using `rake`.  If you run individual specs using `rspec`, it isn't set, and normal autoloading is used.  I'm working on a separate approach to speed up running individual tests.

> The system as seen previously was neatly adaptive to loading as little as reasonably possible under change.

That is one of the advantages of the current approach, and I agree that it is a significant advantage.  So maybe we should keep the autoload and reference approach, but also force references when running full tests to mirror production.  I will adapt this PR to continue to use autoload.  We should still be able to keep the other advantages in this PR (ubid improvements, separate spec and file requiring, avoiding unnecessary autoloads).

fdr:
> > Also...would production break? why would `bad.rb` ever run, even in production, as it has no constants, and thus cannot be hit with autoload, whether eagerly (as seen in `loader.rb` today). It would also nominally break the test coverage.
> 
> I tested this and production does break, because it loads the file. Without this, tests do not . It doesn't matter whether the file defines the constant (you cannot know that before loading the file), it only matters that Unreloader thinks `bad.rb` would define the `Bad` constant, so it would setup the autoload for `Bad` to load `bad.rb`, and then would later reference `Bad`, loading `bad.rb`.

Yeah, I realized that the symbols actually defined in the file are irrelevant, since the symbols are assumed via reading the directory. My bad.

> You are correct that this would be caught by code coverage.
> 
> > But I'd propose a different fix for that particular problem: if all tests are run, run the force-all-symbols code. As a test. The very lazy auto loading is then still active for single tests.
> 
> That's how it works in this PR. `NO_AUTOLOAD` is set when running all tests using `rake`. If you run individual specs using `rspec`, it isn't set, and normal autoloading is used. I'm working on a separate approach to speed up running individual tests.

I'm okay with that. But here's a proposal to consider: allow the force-symbol-falting-code to be called, then call it in the test suite. The result is one fewer knob. You don't even had to add any CI definitions for it.

An example of such a test that doesn't really test any application code is in db_spec.rb:

```ruby
# frozen_string_literal: true

RSpec.describe DB do
  it "has no unexpectedly collated columns" do
    expect(described_class[<<SQL].all.map { _1[:name] }.join(", ")).to eq ""
SELECT quote_ident(nspname) || '.' || quote_ident(relname) || '.' || quote_ident(attname) AS name
FROM pg_class
JOIN pg_namespace ON pg_class.relnamespace = pg_namespace.oid AND
                     pg_namespace.nspname NOT IN ('pg_catalog', 'information_schema')
JOIN pg_attribute ON pg_class.oid = pg_attribute.attrelid
JOIN pg_collation ON pg_attribute.attcollation = pg_collation.oid AND
                     pg_collation.collcollate <> 'C'
SQL
  end
end
```

This ensures we manually specify "collation C" on text columns (unless we want to exempt one, which we haven't yet). It's a bit of a weird test, because no application code gets tested.

The upshot of doing it this way is, nobody runs this test unless they are running the whole suite (or something went wrong), so it mostly naturally does the right thing.

But the difference is *one* knob, so if you like the idea of having it be explicit for some reason, I'm not going to complain to the guy wrote/maintains Unreloader.  The magnitude of supervisory change is far less than having to supervise/profile a shifting landscape of `require`

Side-topic: I've thought it'd be nice -- in theory -- if I could get unreloader to integrate with Pry, to pick up changes right before I evaluate an expression, in development. But I never dug into how it worked with Rack, so have left it an idle thought.

> > The system as seen previously was neatly adaptive to loading as little as reasonably possible under change.
> 
> That is one of the advantages of the current approach, and I agree that it is a significant advantage. So maybe we should keep the autoload and reference approach, but also force references when running full tests to mirror production. I will adapt this PR to continue to use autoload. We should still be able to keep the other advantages in this PR (ubid improvements, separate spec and file requiring, avoiding unnecessary autoloads).

I think we should keep the super-declarative-almost-self-maintaining-laziness if we can, unless there are side effects.

The approach I took with Unreloader was to basically mutate the things I liked about Zeitwerk while getting enough flexibility to perform reopening-style reloading. You can see it in contemporary notes in the commit message at 998ef052242cee9f9cef6fec25ef61023ba43801, which references https://github.com/jeremyevans/roda-sequel-stack/issues/21#issuecomment-894669333

jeremyevans:
> I'm okay with that. But here's a proposal to consider: allow the force-symbol-falting-code to be called, then call it in the test suite. The result is one fewer knob. You don't even had to add any CI definitions for it.

Yep.  My plan is to remove 314a8ce and most of e64644a, and have the forced autoloads run for both the `Config.production?` and `NO_AUTOLOAD` case (maybe renamed to `EAGER_AUTOLOAD`).

> But the difference is _one_ knob, so if you like the idea of having it be explicit for some reason, I'm not going to complain to the guy wrote/maintains Unreloader. The magnitude of supervisory change is far less than having to supervise/profile a shifting landscape of `require`

I have what I guess is an irrational aversion to `autoload`, mostly from all of the pain it's caused me when trying to get my applications working with limited file system access (first with `chroot` and later with OpenBSD's `unveil`).  In CRuby, the `autoload` internals are very complex, much more than `require`, but by now I think most of the bugs have been ironed out.  So the benefits of `require` vs. `autoload` are more theoretical, while the issues you identified are more practical, and practicality should beat purity (to quote the zen of python).

> Side-topic: I've thought it'd be nice -- in theory -- if I could get unreloader to integrate with Pry, to pick up changes right before I evaluate an expression, in development. But I never dug into how it worked with Rack, so have left it an idle thought.

I've added that to my todo list.  It should be possible, I just have to figure out how to hook into Pry's REPL, so I can add the reload checks right before each statement is executed.

fdr:
Interesting. I never used any autoloading until system 3, at crunchy, and it was completely solid.

Likewise this zeitwerk-oid use of unreloader had never given a problem, and the flexibility to do reopen-reloads and even tweak the namespaces around a little bit have been useful.

fdr:
> > I'm okay with that. But here's a proposal to consider: allow the force-symbol-falting-code to be called, then call it in the test suite. The result is one fewer knob. You don't even had to add any CI definitions for it.
> 
> Yep. My plan is to remove [314a8ce](https://github.com/ubicloud/ubicloud/commit/314a8ce74730e75f38b7be0840ca44c551222297) and most of [e64644a](https://github.com/ubicloud/ubicloud/commit/e64644a6de64dfe53ba306b51c4f94d851d99287), and have the forced autoloads run for both the `Config.production?` and `NO_AUTOLOAD` case (maybe renamed to `EAGER_AUTOLOAD`).

I guess my point is, you could perhaps avoid any variable, if you include a test that calls the "touch all symbols" code, e.g. a `loader_spec.rb`. Then it would always run (but at an arbitrary point in execution, which is different) when running the whole suite, or someone could run just that test and get the same behavior as tweaking the environment variable, except without having to think about a separate system of administration/flagging in the test.

jeremyevans:
I've refactored this to continue to use autoload, but to support forcing autoloads in test as well as production, using FORCE_AUTOLOAD.  The default/spec/pspec task sets this environment variable, so that when running the full test suite, the test environment is more similar to the production environment.

fdr (approved):

jeremyevans:
I pushed a small commit to add nocov markers around the overridden code in UBID.  This currently is covered in CI, because CI does not force autoloads when running.  I'll be sending a pull request to do that shortly after this is merged.

PR #2071, opened 2024-10-16
Do not register runners for suspended accounts

enescakir:
- **Add active? helper to the project**
  We need to determine whether a project is active or not to allow certain
  actions, like installing our GitHub App. This is particularly important
  when we're investigating the project for any fraudulent activity. I've
  added two basic conditions to assess a project's activity status, and we
  can improve this in the future.
  
  At present, if any account within the project is suspended, we consider
  the project inactive. We also soft delete projects, preserving their
  billing information temporarily to finalize the billing. If a project is
  'soft deleted', we classify it as inactive.
  

- **Do not poll jobs if the project is inactive**
  We have a job poller that checks the repository for queued jobs via
  GitHub API, ensuring no webhook events are missed and that a runner is
  provisioned for each one.
  
  However, when we suspend an account, we intentionally avoid provisioning
  runners. Despite this, the poller might misinterpret this as the
  repository lacking runners and provision new ones.
  
  Ideally, we should destroy GithubInstallation when we suspend an account.
  If for some reason this hasn't been done, we shouldn't poll the jobs.
  

- **Do not register runner for inactive projects**
  In the last wave of fraud, we prevented new runner registrations for
  fraudulent accounts using our quota system. However, we can simplify
  this. If the project is inactive, we can terminate the runner prog. A
  project's activity status is determined by suspended accounts and soft
  project deletions. As the job poller also checks project activity, we
  can exit from this prog instead of taking a long nap. The poller will
  not a create new runners.

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -78,6 +78,10 @@ class Project < Sequel::Model
     end
   end
 
+  def active?
+    visible && accounts_dataset.exclude(suspended_at: nil).empty?
+  end
+
   def current_invoice
     begin_time = invoices.first&.end_time || Time.new(Time.now.year, Time.now.month, 1)
     end_time = Time.now
--- a/prog/github/github_repository_nexus.rb
+++ b/prog/github/github_repository_nexus.rb
@@ -32,6 +32,10 @@ class Prog::Github::GithubRepositoryNexus < Prog::Base
   end
 
   def check_queued_jobs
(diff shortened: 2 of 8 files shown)


fdr (commented):
not finished, but wanted to get my tiny microoptimzation out there :cool:

velioglu (approved):

PR #1233, opened 2024-02-12
Increase the amount of freezing

fdr:
Clover isn't intended to depend on class modifications after an
    initial load phase.  There's no reason to restrict freezing to core
    classes or models, and using `loader.rb`'s existing complications to
    itemize classes under our control to freeze is broad and economical.

PR #2074, opened 2024-10-16
Run finalize_associations on each Sequel::Model class when testing

jeremyevans:
This is something Sequel::Model.freeze_descendants does, which is what is run in production.  It would be best to be able to use Sequel::Model.freeze_descendants when testing, but due to the extensive use of rspec mocking, that isn't currently possible.  This makes the test environment closer to the production environment, and it speeds up the tests.

While here, change freeze_descendents to freeze_descendants. Both are currently supported, but descendents is a typo, and I plan to remove freeze_descendents in Sequel 6 (see https://github.com/jeremyevans/sequel/commit/185efe54047a6c912555651523726f9929451589)

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -7,8 +7,10 @@ require "roda"
 class Clover < Roda
   def self.freeze
     # :nocov:
-    unless Config.test?
-      Sequel::Model.freeze_descendents
+    if Config.test?
+      Sequel::Model.descendants.each(&:finalize_associations)
+    else
+      Sequel::Model.freeze_descendants
       DB.freeze
     end
     # :nocov:


enescakir (approved):

fdr (approved):

fdr:
I suspect we could freeze our stuff quite well, it's stuff like this that is brain-melting

4cfcceb22c85d914cd88e6e20af3d20ef01382e8

...and given we mock out certain subroutines wholesale (e.g. HTTP, per webmock), I suppose there is no perfect solution for exercise.

See https://github.com/ubicloud/ubicloud/pull/1233

jeremyevans:
Refrigerator is great at finding places where the core classes are modified after being loaded.  Sometimes these cases are annoying to diagnose and work around.  I'm not sure how much of a problem it has been in the past to keep things passing with refrigerator.  

I think mocking HTTP and anything using external I/O is a good idea, as you can never be sure that will be reliable when testing.  I'm guessing the reason that we are mocking Sequel model classes is for speed.  Certainly the test suite is quite fast, so we are successful on that front.  However, maybe we should consider a mockless approach for testing involving DB and Sequel models.  A mockless approach will make testing slower, but it gives more confidence that we are testing what we want to test, as it allows the test environment to better match the production environment.

enescakir:
I'm often uncertain about when to mock the database. This question comes up for me periodically. https://github.com/ubicloud/ubicloud/pull/1696#discussion_r1640116056

Currently, our approach depends on the initial author of the spec file. If they start with a mocked database, the rest of us follow suit. However, if they begin with a real database, we stick with that. This approach has led to several instances where our tests failed to detect bugs because we were mocking the database.

fdr:
Usually, if the queries are even a little interesting, I’ll run against the
database. Otherwise I try to take the speed up. You can split the
difference if necessary by having a method that returns a neatly mockable
representation, testing that with real database access, and then most tests
can be faster and avoid the database.

jeremyevans:
My personal preference is to never mock the database.  Fast tests are great, but tests that mock do not give me adequate confidence that the code will do what I expect.  We could make our test suite 3x slower and it would likely still finish in under a minute on most development laptops (in parallel mode). Granted, our test suite will grow slower over time as features are added, but I still think tradeoffs made by mocking the database are not worth it.  My opinion is we should only mock what we cannot control, such as external HTTP APIs.

PR #2076, opened 2024-10-16
Fix available storage bug on vm creation page

enescakir:
We recently started to passing available postgres storage sizes to the frontend with location details at 7b430520. However, we didn't apply this to available VM storage sizes, resulting in them lacking location keys. But they use same JS function, so it broke vm creation page. If the options are a single-dimension array and not a hash, there's no need for selecting by location. But if it's not an array, we should select by location.


bsatzger (approved):
This is getting a little hard to read but good to have a quick fix for the issue in production. Maybe add a comment why we have to do this?

PR #2077, opened 2024-10-16
Only run E2E GitHub Action when the variable RUN_E2E is set

fdr:
One (all?) of our contributors got annoying email once an hour on their fork of Ubicloud.

To change the default, while allowing downstream forks to in principle run E2E if they set everything up, I use a "variable" on the repository.  A fork can set the variable at a URL like this:

    https://github.com/$user-name/ubicloud/settings/variables/actions

To enable the hourly E2E runs, assign the text "true" to the `RUN_E2E` variable.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,6 +14,7 @@ on:
 
 jobs:
   run-ci:
+    if: vars.RUN_E2E == 'true'
     runs-on: ubicloud
     environment: E2E-CI
     timeout-minutes: 45


.github/workflows/e2e.yml
```diff
@@ -14,6 +14,7 @@ on:
 
 jobs:
   run-ci:
+    if: vars.RUN_E2E == 'true'
```
enescakir, line 17: You can also use something like 
```suggestion
    if: ${{ github.repository == 'ubicloud/ubicloud' }}
```

fdr: I thought it'd be nice if non-`ubicloud/ubicloud` repos could set up E2E tests without applying a patch

enescakir (approved):

PR #2078, opened 2024-10-16
Fix nondeterministic test when using invalid flavor

jeremyevans:
This spec has failed twice on GitHub Actions when submitting PRs, and once when I was doing parallel testing.

Reading through the code, the problem is Clover is redirecting based on the Referer header, but I think Referer header state may be leaking between test runs.  I'm not sure how to fix that, but this works around the problem by setting a specific Referer header in the spec.

Example failure: https://github.com/ubicloud/ubicloud/actions/runs/11368240212/job/31622834684?pr=2074#step:15:18


fdr (approved):
good catch.

enescakir (approved):

PR #2079, opened 2024-10-16
Replace database_cleaner-sequel with call to DB.transaction

jeremyevans:
I'm starting work on optimizing running single specs.  The TRUNCATE commands run by database_cleaner-sequel take far longer than the running a single spec.  Using a separate transaction per spec, which is rolled back after the spec completes, is simple to do by calling DB.transaction with the :rollback option.

With this, we can remove the GRANT TRUNCATE in the Rakefile, as well as a substantial amount of setup code in spec/spec_config.

On my laptop, this cuts the time to run a single spec by about 45%, from about 2.5 seconds to about 1.4 seconds (this will hopefully get to about 0.5 seconds with some additional changes).

Unfortunately, rspec does not support a global around(:context), so you cannot do a transaction-per-suite and savepoint-per-spec like you can when using minitest with minitest-hooks.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,7 +52,6 @@ group :development do
 end
 
 group :test do
-  gem "database_cleaner-sequel"
   gem "capybara"
   gem "rspec"
   gem "webmock"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -103,10 +103,6 @@ GEM
       bigdecimal
       rexml
     crass (1.0.6)
-    database_cleaner-core (2.0.1)
-    database_cleaner-sequel (2.0.2)
-      database_cleaner-core (~> 2.0.0)
-      sequel
     date (3.3.4)
     diff-lcs (1.5.1)
     docile (1.4.1)
@@ -374,7 +370,6 @@ DEPENDENCIES
   by!
   capybara
   countries
-  database_cleaner-sequel
   ed25519
   erb-formatter!
   erubi (>= 1.5)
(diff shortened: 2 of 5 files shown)


fdr (approved):
I'm all about eliminating both code and dependency (transitive code also counts!)

fdr:
Huh. I wonder why rspec doesn't support that, given it kinda does everything (where everything is sometimes "too much")

The current situation is nevertheless better and appealingly shorter and with fewer dependencies.  Unlike when this was written, we have quite a few tests now without such a rare case, so may as well go for it.

jeremyevans:
> Huh. I fwonder why rspec doesn't support that, given it kinda does everything (where everything is sometimes "too much")

This has been a long-running problem with rspec.  At least now it at least warns if you try to use `around(:all)`: https://github.com/rspec/rspec-core/issues/2486

There is this absolutely horrific hack for `around(:all)` support using fibers, but the repository has been archived for a while (no idea whether it actually works): https://github.com/seanwalbran/rspec_around_all/blob/master/lib/rspec_around_all.rb 

To be fair, minitest doesn't support all hooks for before, after, or around, you have to use a separate minitest-hooks library for that.  That's not because it's difficult for minitest, it's a deliberate decision by the minitest maintainer, who does not believe tests should have any shared state.  Personally, I view shared state in tests akin to database denormalization.  You generally should avoid it, but for performance reasons, it can be worth the tradeoffs.

fdr:
Oh, I realized something else: I turn fsync off for my development Postgres. It makes everything a bit faster. I was thinking about the magnitude of the different here and felt like something was bothering me. Maybe that's what it was. There's an old slack message somewhere that had a measurement

PR #2082, opened 2024-10-17
Move nic gateway setting to vm_setup

furkansahin:
## Move nic gateway setting to vm_setup
Nic gateway setup is the most crucial piece in ipv4 networking setup.
Most of that work already is happenning in vm_setup. Keeping this piece
in controlplane doesn't make sense. Therefore, I am moving it to the
proper place.

With this change, we simplify the controlplane operations for cases of
provisioning a VM and initially setting up its IPv4 networking or
repopulating networking after a host reboot. Without controlplane
intervention, we make sure the private subnet's first IP is added as the
gateway to the tap device during initialization.

This will require migration of prep.json files, which is fine and will
be handled at the time of deployment for existing VMs.

## Vm::Nexus.assemble clean-up for private networking setup
We just in-line some of the variable assignments and clean up some
scopes properly.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -64,29 +64,27 @@ class Prog::Vm::Nexus < Prog::Base
       # then we create a nic on that subnet.
       # - If the user provided neither nic_id nor private_subnet_id, that's OK, we create both.
       nic = nil
-      subnet = nil
-      if nic_id
+      subnet = if nic_id
         nic = Nic[nic_id]
         raise("Given nic doesn't exist with the id #{nic_id}") unless nic
         raise("Given nic is assigned to a VM already") if nic.vm_id
         raise("Given nic is created in a different location") if nic.private_subnet.location != location
         raise("Given nic is not available in the given project") unless project.private_subnets.any? { |ps| ps.id == nic.private_subnet_id }
 
-        subnet = nic.private_subnet
+        nic.private_subnet
       end
 
       unless nic
-        subnet = nil
-        if private_subnet_id
+        subnet = if private_subnet_id
           subnet = PrivateSubnet[private_subnet_id]
           raise "Given subnet doesn't exist with the id #{private_subnet_id}" unless subnet
           raise "Given subnet is not available in the given project" unless project.private_subnets.any? { |ps| ps.id == subnet.id }
+          subnet
         else
-          subnet_s = Prog::Vnet::SubnetNexus.assemble(project_id, name: "#{name}-subnet", location: location, allow_only_ssh: allow_only_ssh)
-          subnet = PrivateSubnet[subnet_s.id]
+          subnet = Prog::Vnet::SubnetNexus.assemble(project_id, name: "#{name}-subnet", location: location, allow_only_ssh: allow_only_ssh).subject
+          PrivateSubnet[subnet.id]
         end
-        nic_s = Prog::Vnet::NicNexus.assemble(subnet.id, name: "#{name}-nic")
-        nic = Nic[nic_s.id]
+        nic = Prog::Vnet::NicNexus.assemble(subnet.id, name: "#{name}-nic").subject
       end
 
       cores = if arch == "arm64"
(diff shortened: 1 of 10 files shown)


prog/vm/nexus.rb
```diff
@@ -239,7 +237,7 @@ def before_run
     case host.sshable.cmd("common/bin/daemonizer --check prep_#{q_vm}")
     when "Succeeded"
       host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")
-      vm.nics.each { _1.incr_setup_nic }
+      vm.private_subnets.map(&:incr_add_new_nic)
```
enescakir, line 240: nit: we don't use output
```suggestion
      vm.private_subnets.each(&:incr_add_new_nic)
```

prog/vm/nexus.rb
```diff
@@ -64,29 +64,27 @@ def self.assemble(public_key, project_id, name: nil, size: "standard-2",
...
       unless nic
-        subnet = nil
-        if private_subnet_id
+        subnet = if private_subnet_id
           subnet = PrivateSubnet[private_subnet_id]
           raise "Given subnet doesn't exist with the id #{private_subnet_id}" unless subnet
           raise "Given subnet is not available in the given project" unless project.private_subnets.any? { |ps| ps.id == subnet.id }
+          subnet
         else
           subnet_s = Prog::Vnet::SubnetNexus.assemble(project_id, name: "#{name}-subnet", location: location, allow_only_ssh: allow_only_ssh)
```
enescakir, line 84:
```suggestion
          Prog::Vnet::SubnetNexus.assemble(project_id, name: "#{name}-subnet", location: location, allow_only_ssh: allow_only_ssh).subject
```

fdr (commented):
perhaps re-add the tests (or reinterpret them) in rhizome

prog/vm/nexus.rb
```diff
@@ -64,29 +64,27 @@ def self.assemble(public_key, project_id, name: nil, size: "standard-2",
...
-        if private_subnet_id
+        subnet = if private_subnet_id
           subnet = PrivateSubnet[private_subnet_id]
           raise "Given subnet doesn't exist with the id #{private_subnet_id}" unless subnet
           raise "Given subnet is not available in the given project" unless project.private_subnets.any? { |ps| ps.id == subnet.id }
+          subnet
         else
-          subnet_s = Prog::Vnet::SubnetNexus.assemble(project_id, name: "#{name}-subnet", location: location, allow_only_ssh: allow_only_ssh)
-          subnet = PrivateSubnet[subnet_s.id]
+          subnet = Prog::Vnet::SubnetNexus.assemble(project_id, name: "#{name}-subnet", location: location, allow_only_ssh: allow_only_ssh).subject
```
fdr, line 84: This is a straight local respelling of some variables to clean up some scopes, right? Does it have something to do with the stated purpose of the patch?

prog/vm/nexus.rb
```diff
@@ -240,7 +238,7 @@ def before_run
     case host.sshable.cmd("common/bin/daemonizer --check prep_#{q_vm}")
     when "Succeeded"
       host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")
-      vm.nics.each { _1.incr_setup_nic }
```
fdr, line 243: Likewise these hunks are also not related to the stated patch purpose right?

fdr (changes requested):
It seems like the crux of this patch is far fewer hunks that is seen in the commit, and it's kinda sensitive and short while making its improvement, can the hairy part to check be more focused? I think I found it, but I tend to re-read the same code multiple times, so a bigger inner-loop of code churn tends to deteriorate my accuracy.

prog/vm/nexus.rb
furkansahin: You're right, I moved them to a separate commit

prog/vm/nexus.rb
furkansahin: I moved them

fdr:
You're going to have to break out the rhizome commit to go first right, in a somewhat backwards-compatible way? What's the phasing?

fdr (approved):

PR #2084, opened 2024-10-17
Avoid /api/ path mangling for testing purposes

fdr:
This makes test and production more similar.  By piggybacking on the existing rspec metadata used to add other common headers, it's possible to likewise set the Host header by default.

Most of the bulk of this patch is from running `sed`, but the interesting hunks are in:

* clover.rb: eliminate a branch to add `/api/` mangling

* routes/api/spec_helper.rb: add the Host header by default

* routes/api/project_spec.rb: remove a special test for the host calling convention, as it's now used in every test.

The motivation for this was to make the program easier to validate with OpenAPI and `committee`, where having two calling conventions (even for the one test that set `Host`) would need more workarounds.

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -476,18 +476,6 @@ class Clover < Roda
     if api?
       response.json = true
       response.skip_content_security_policy!
-      # To make test and development easier
-      # :nocov:
-      unless Config.production?
-        r.on("api") do
-          r.rodauth
-          rodauth.check_active_session
-          rodauth.require_authentication
-          r.hash_branches("")
-        end
-      end
-      # :nocov:
-
       r.rodauth
     else
       r.on "runtime" do
--- a/spec/model/private_subnet_spec.rb
+++ b/spec/model/private_subnet_spec.rb
@@ -164,18 +164,18 @@ RSpec.describe PrivateSubnet do
     }
 
     let(:ps1) {
-      Prog::Vnet::SubnetNexus.assemble(prj.id, name: "test-ps1", location: "hetzner-hel1").subject
+      Prog::Vnet::SubnetNexus.assemble(prj.id, name: "test-ps1", location: "hetzner-fsn1").subject
     }
 
     it ".connected_subnets" do
-      ps2 = Prog::Vnet::SubnetNexus.assemble(prj.id, name: "test-ps2", location: "hetzner-hel1").subject
+      ps2 = Prog::Vnet::SubnetNexus.assemble(prj.id, name: "test-ps2", location: "hetzner-fsn1").subject
       expect(ps1.connected_subnets).to eq []
 
       ps1.connect_subnet(ps2)
       expect(ps1.connected_subnets.map(&:id)).to eq [ps2.id]
       expect(ps2.connected_subnets.map(&:id)).to eq [ps1.id]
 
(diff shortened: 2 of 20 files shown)


fdr:
Also, @geemus

fdr:
okay just some dangling api references: `api/` rather than `/api/` that weren't caught by my sed. @jeremyevans do you have opinions about whether the (apparently optional) leading slash should be normal notation, or not?

jeremyevans (approved):
I would always include the leading slash in the paths for tests.  If it works without the leading slash, then some other library is inserting it automatically, but I think it unwise to rely upon that behavior.

clover.rb
```diff
@@ -21,15 +21,6 @@ def self.freeze
       r.run CloverApi
     end
 
-    # To make test and development easier
-    # :nocov:
-    unless Config.production?
```
enescakir, line 26: How can we test it in development?

fdr:
That is a bummer, I did not think of the interactive case. Will either
withhold that hunk or explore a viable alternative.

jeremyevans:
@enescakir You can try using `api.localhost` in the tests.  That works on my Ubuntu laptop with no configuration changes. If it doesn't work on yours, you can add an `127.0.0.1 api.localhost` (and/or `::1` for IPv6) entry to `/etc/hosts` .  Do you think that would be sufficient?

velioglu (approved):

fdr:
Yeah I think we're just going to have to eat this one, and perhaps tell macintosh people in DEVELOPERS.md to do something with /etc/hosts. I think mangling via path is just a bit too divergent to make it a good idea from what I got on tap for now...sorry folks.

This is truly the year of Linux on the Desktop,

jeremyevans:
@fdr I rebased this on main for you, to fix all of the conflicts (mostly with the ubid-name path merging, and also with the Clover* merging).

fdr:
@jeremyevans can you give this a quick once over one more time? I didn't do anything structurally, just a lot of hack and slash to get past the conflicts. And I will send mail to eng as well about the invasiveness to macintosh users in particular, since they would have to bind `api.localhost` to interact with the API outside rack if hosted locally

jeremyevans (approved):
The `hetzner-hel1` to `hetzner-fsn1` changes seem unrelated, but I'm not against including them in the same commit.

jeremyevans:
To make it easier for make users, we could have public DNS for `api.local.ubicloud.com` resolve to `127.0.0.1`/`::1`.  Not sure if we like that kind of hack, though.

fdr:
> The `hetzner-hel1` to `hetzner-fsn1` changes seem unrelated, but I'm not against including them in the same commit.

yeah, I thought about mentioning this debris in my sed processing, but I thought it subtracted more from clarity than it added in the commit message

fdr:
Actually? I'll mention it, to not leave future person confused about why it's there

fdr:
> To make it easier for make users, we could have public DNS for `api.local.ubicloud.com` resolve to `127.0.0.1`/`::1`. Not sure if we like that kind of hack, though.

We could do that, yeah, to spare occasional users who usually have internet access from having to hack /etc/hosts.

PR #2085, opened 2024-10-17
Validate Rhizome after install

aamederen:
We have seen incidents where some of the files in the rhizome
directory were empty. This is rare but also concerning.
We don't precisely know why that happened,
but it gave us an idea to improve our InstallRhizome prog.

This change adds a validation step to InstallRhizome prog
as a first step, maybe a 20/80 improvement. We checksum all the
files we pack into the rhizome package to be sent to the sshable
and add an extra file to the package with all the file
inventory along the hashes. Then, an additional script within
the package called `validate` is run by the prog, which goes
through the inventory and validates the hashes of all files.
If anything goes wrong, the InstallRhizome prog will fail and
we will know about it sooner so we can investigate more.

Some extra thoughts:
- Tried xxhash for being fast, but that required native compilation
so failed back to built-in md5, then finally SHA384.
- another method _could_ be: hash the tar and retar the untar-ed Rhizome to rehash
- Any harm of validation after the bundle step?
- Constructing the mode of a new file, which is an integer, seemed hard.
When I checked the files, non-executables were 33188 and executables were 33261.
Copied the stats of a non-executable (Gemfile):
`File.stat(base + "/Gemfile").mode == 33188`. Apparently
(thanks @jeremyevans for the pointer), it's the decimal value of the
octal representation 100644 or 100755.
- Ideally we would add the capability of syncing the deleted files as well.
Even better, atomic switch to a newer version of Rhizome folder,
maybe with symlinks. That's a future work.

### DIFF
--- a/prog/install_rhizome.rb
+++ b/prog/install_rhizome.rb
@@ -2,15 +2,17 @@
 
 require "rubygems/package"
 require "stringio"
+require "digest/sha2"
 
 class Prog::InstallRhizome < Prog::Base
   subject_is :sshable
 
   label def start
     tar = StringIO.new
+    file_hash_map = {} # pun intended
     Gem::Package::TarWriter.new(tar) do |writer|
       base = Config.root + "/rhizome"
-      Dir.glob(["Gemfile", "Gemfile.lock", "common/**/*", "#{frame["target_folder"]}/**/*"], base: base).map do |file|
+      Dir.glob(["Gemfile", "Gemfile.lock", "common/**/*", "#{frame["target_folder"]}/**/*"], base: base) do |file|
         next if !frame["install_specs"] && file.end_with?("_spec.rb")
         full_path = base + "/" + file
         stat = File.stat(full_path)
@@ -22,11 +24,16 @@ class Prog::InstallRhizome < Prog::Base
               IO.copy_stream(_1, tf)
             end
           end
+          file_hash_map[file] = Digest::SHA384.file(full_path).hexdigest
         else
           # :nocov:
           fail "BUG"
           # :nocov:
         end
+      end
+
+      writer.add_file("hashes.json", 0o100755) do |tf|
+        tf.write file_hash_map.to_json
       end
     end
 
@@ -37,8 +44,14 @@ class Prog::InstallRhizome < Prog::Base
   end
 
   label def install_gems
-    sshable.cmd("bundle config set --local path vendor/bundle")
-    sshable.cmd("bundle install")
+    sshable.cmd("bundle config set --local path vendor/bundle && bundle install")
+
+    hop_validate
+  end
+
+  label def validate
+    sshable.cmd("common/bin/validate")
+
     pop "installed rhizome"
   end
 end
--- a/rhizome/common/bin/validate
+++ b/rhizome/common/bin/validate
@@ -1,0 +1,20 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require "json"
+require "digest/sha2"
+
+hashes = JSON.load_file("hashes.json")
+
+violations = []
+
+hashes.each do |filename, expected_hash|
+  calculated_hash = Digest::SHA384.file(filename).hexdigest
+  next if calculated_hash == expected_hash
+
+  violations << {name: file, expected: expected_hash, calculated: calculated_hash}
+end
+
+puts violations
+
+fail "Error during rhizome validation: " + violations.to_s if violations.any?
--- a/spec/prog/install_rhizome_spec.rb
+++ b/spec/prog/install_rhizome_spec.rb
@@ -33,9 +33,15 @@ RSpec.describe Prog::InstallRhizome do
 
   describe "#install_gems" do
     it "runs some commands and exits" do
-      expect(sshable).to receive(:cmd).with("bundle config set --local path vendor/bundle")
-      expect(sshable).to receive(:cmd).with("bundle install")
-      expect { ir.install_gems }.to exit({"msg" => "installed rhizome"})
+      expect(sshable).to receive(:cmd).with("bundle config set --local path vendor/bundle && bundle install")
+      expect { ir.install_gems }.to hop
+    end
+  end
+
+  describe "#validate" do
+    it "runs the validate script" do
+      expect(sshable).to receive(:cmd).with("common/bin/validate")
+      expect { ir.validate }.to exit({"msg" => "installed rhizome"})
     end
   end
 end


jeremyevans:
> * Constructing the mode of a new file, which is an integer, seemed hard. When I checked the files, non-executables were 33188 and executables were 33261. Copied the stats of a non-executable (Gemfile): `File.stat(base + "/Gemfile").mode == 33188`

In general, you should use octal for the mode.  A common mode for a file would be 0100664.  The first three octal digits (100) indicate a regular file.  The last three octal digits (664) are owner, group, and world permissions.  Generally when you are setting permissions, you only set the last three (e.g. 0664).

aamederen:
For context, I talked with @bsatzger and @fdr about this PR and both thought this is a good first step.

prog/install_rhizome.rb
```diff
@@ -22,12 +24,17 @@ class Prog::InstallRhizome < Prog::Base
               IO.copy_stream(_1, tf)
             end
           end
+          file_hash_map[file] = Digest::MD5.file(full_path).hexdigest
```
fdr, line 27: While fine for this, any use of MD5 makes automatic checkers go berserk, needs SOC2 documentation, etc. unless the reasons are especially compelling, use an unremarkable digest. I like sha512-224 when I don't have compatibility issues.

aamederen: haven't thought about the auto-checks and other possible headache points about md5, it was just simple and fast without any security concerns. Switched to SHA 512 now (not sha512-224 because I wanted to stick to the stdlib and IMO anything is fine for this use case, but LMK if you think otherwise).

fdr: I think OpenSSL is not stdlib but stdlib-adjacent, and we have dependencies for it allllll over the place. Main advantage of 512-224: shorter. Secondary advantage: more resistant to length extension attacks, but who cares in this case I guess. See table at https://en.wikipedia.org/wiki/SHA-2#Comparison_of_SHA_functions. Also SHA512 is faster on modern architectures. This is all nerd-golf though relative to the purpose.

```ruby
require 'openssl'

# Create a SHA-512/224 digest object
digest = OpenSSL::Digest.new('SHA512-224')

# Compute the hash of a string
message = "Hello, world!"
hash = digest.digest(message)

# Convert the binary hash to a hexadecimal string
hex_hash = hash.unpack1('H*')

puts hex_hash

```

jeremyevans: The `openssl` ruby library is stdlib, there are no plans I am aware of to move it from stdlib to bundled gems. I don't think it will ever move to bundled gems, as rubygems depends on it.  That being said, I would recommend using SHA512 over SHA512-224 unless hash length is a significant issue.

jeremyevans: Because of ubiquity.  If you are worried about length extension attacks, I recommend we use SHA-384.  That should be more common than SHA-512/224.  SHA-256/224 is more resistant to length extension attacks, but SHA-384 is more resistant to collision attacks, while still providing sufficient security against length extension attacks.

fdr: SHA-384 suffices, then.

prog/install_rhizome.rb
```diff
@@ -22,12 +24,17 @@ class Prog::InstallRhizome < Prog::Base
               IO.copy_stream(_1, tf)
             end
           end
+          file_hash_map[file] = Digest::SHA256.file(full_path).hexdigest
```
fdr, line 27:
```suggestion
          file_hash_map[file] = Digest::SHA384.file(full_path).hexdigest
```

prog/install_rhizome.rb
```diff
@@ -22,12 +24,17 @@ class Prog::InstallRhizome < Prog::Base
...
         else
           # :nocov:
           fail "BUG"
           # :nocov:
         end
       end
+
+      writer.add_file("hashes.json", "100755".to_i(8)) do |tf|
```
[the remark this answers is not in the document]
fdr: to_i is always one typo away from trouble: 
```
[6] clover-development(main)> "hello".to_i(8)
=> 0
[7] clover-development(main)> "hello".to_i
=> 0
```

rhizome/common/bin/validate
```diff
@@ -0,0 +1,20 @@
...
+
+require "json"
+require "digest/sha2"
+
+hashes = JSON.load_file("hashes.json")
+
+violations = []
+
+hashes.each do |filename, expected_hash|
+  calculated_hash = Digest::SHA256.file(filename).hexdigest
```
fdr, line 12:
```suggestion
  calculated_hash = Digest::SHA384.file(filename).hexdigest
```

prog/install_rhizome.rb
```diff
@@ -22,12 +24,17 @@ class Prog::InstallRhizome < Prog::Base
...
           end
+          file_hash_map[file] = Digest::SHA384.file(full_path).hexdigest
         else
           # :nocov:
           fail "BUG"
           # :nocov:
         end
       end
+
+      writer.add_file("hashes.json", Integer("100755", 8) do |tf|
```
jeremyevans:
```suggestion
      writer.add_file("hashes.json", 0100755) do |tf|
```

prog/install_rhizome.rb
```diff
@@ -22,12 +24,17 @@ class Prog::InstallRhizome < Prog::Base
...
           end
+          file_hash_map[file] = Digest::SHA384.file(full_path).hexdigest
         else
           # :nocov:
           fail "BUG"
           # :nocov:
         end
       end
+
+      writer.add_file("hashes.json", Integer("100755", 8)) do |tf|
```
jeremyevans, line 35:
```suggestion
      writer.add_file("hashes.json", 0100755) do |tf|
```

Seems best to avoid unnecessary string to integer conversion.  If the `0` prefix is too obscure, you could use `0o100755`

jeremyevans (commented):
Looks good.  I recommend one optional change.

prog/install_rhizome.rb
jeremyevans: This was superseded by another change.

prog/install_rhizome.rb
aamederen: TIL `0o100755 == 33261` ℹ️

rhizome/common/bin/validate
```diff
@@ -0,0 +1,20 @@
+#!/bin/env ruby
```
jeremyevans, line 1: On my system `/bin` is a symlink to `usr/bin`. If that is true on the VM hosts, maybe 

```suggestion
#!/usr/bin/env ruby
```

jeremyevans (approved):
File mode change looks good to me.  I don't have experience with the data plane parts, but the changes make sense to me.

rhizome/common/bin/validate
aamederen: I did a quick grep and through the rhizome scripts, `/bin/env` is used 21 times and `/usr/bin/env` is used 2 times. For the consistency, let's leave it as is in this PR but discuss switching all?

fdr (approved):

PR #2090, opened 2024-10-19
Update openapi.yml with closer-to-correct in-progress version

fdr:
Daniel Farina <daniel@ubicloud.com> writing the commit message here.

Since the first prototypes of `openapi.yml`, there has been drift in the implementation while we work out the development tools to run the tests with the `committee` gem to keep it up to date.

But, as we're working with documentation, it'd be useful to have a recent snapshot of the in-progress work to review how it looks when generating API reference pages, so, copy in `geemus <geemus@gmail.com>`'s up to date copy of the converging `openapi.yml`. Since it's his in-progress work, I have listed him as the author.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -29,6 +29,7 @@ paths:
                 password:
                   type: string
                   example: password
+              additionalProperties: false
               required:
                 - login
                 - password
@@ -43,6 +44,9 @@ paths:
                   success:
                     type: string
                     example: You have been logged in
+                additionalProperties: false
+                required:
+                  - success
         default:
           $ref: '#/components/responses/Error'
       security: []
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -210,14 +224,299 @@ paths:
...
+                      $ref: '#/components/schemas/Firewall'
+                additionalProperties: false
+                required:
+                  - count
+                  - items
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Firewall
+  '/project/{project_id}/location/{location}/firewall/id/{firewall_id}':
```
jeremyevans, line 420: Shouldn't this be `firewall/{firewall_name_or_id}` instead of `firewall/id/{firewall_id}`, since a5be9194193bedd821d93d8f682ac64826a3b1f9 has been merged? Ditto for some other similar paths.  Or is that change expected to be merged later?

fdr: this predates that, but we'll probably snap the openapi.yml again. I suppose given the documentation repo is not  dependent on this copy, it can wait.

PR #2091, opened 2024-10-20
Avoid 4 allocations for every request

jeremyevans:
Instead of splitting the host, just check it starts with `api.`. The four allocations come from:

* 1 array allocation
* 3 string allocations (potentially more or less depending on the number of `.` in the Host header)

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -16,8 +16,7 @@ class Clover < Roda
   end
 
   route do |r|
-    subdomain = r.host.split(".").first
-    if subdomain == "api"
+    if r.host.start_with?("api.")
       r.run CloverApi
     end
 


clover.rb
```diff
@@ -16,8 +16,7 @@ def self.freeze
   end
 
   route do |r|
-    subdomain = r.host.split(".").first
-    if subdomain == "api"
+    if r.host.start_with?("api.")
```
enescakir, line 19: Should we use header_matcher or is it too much for just single usage?

jeremyevans: I think it's probably overkill for just this usage.

enescakir (approved):

furkansahin (approved):

PR #2096, opened 2024-10-22
Drop doc folder

aamederen:
Motivation: write better/more up-to-date versions on ubicloud.com/docs and avoid duplication.


furkansahin:
Can you move these to https://github.com/ubicloud/ubinest/wiki before completely removing?

enescakir (approved):

PR #2097, opened 2024-10-22
Add screenshots rake task

jeremyevans:
This task runs bin/regen_screenshots, which automates the generation of screenshots used on the documentation site.

The screenshot generation code is similar to the route testing code, but instead of using rack-test as the capybara driver, cuprite is used.  Cuprite is a capybara driver that uses the ferrum Ruby library to automate Chrome using the Chrome DevTools Protocol, which is substantially more reliable than Selenium/WebDriver/ChromeDriver.

Using an actual browser (Chrome) for the screenshot generation code requires using an actual server to handle the brower's requests. So this starts Puma as a separate thread in the screenshot generation process.

To avoid permanent database modification, this uses a transaction around the screenshot generation code.  As the server runs in a separate thread than the screenshot generation code, this uses the Sequel temporarily_release_connection extension to share a single connection between server and the main thread, so the main thread can open the transaction, all server actions use the same connection inside the transaction, and the main thread can rollback the transaction after generating all screenshots.

In order for this to work correctly, two things need to happen:

* The documentation site needs to be updated to store all screenshots in the screenshots directory

* bin/regen_screenshots needs to be modified to generate all screenshots stored in the screenshots directory.  Currently, the screenshots generated are just examples.  There are screenshots used in the documentation site that are not generated by bin/regen_screenshots, and bin/regen_screenshots currently generates screenshots not used in the documentation site.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,7 @@ group :development do
   gem "standard", ">= 1.24.3"
   gem "simplecov"
   gem "turbo_tests"
+  gem "cuprite"
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -105,6 +105,9 @@ GEM
       bigdecimal
       rexml
     crass (1.0.6)
+    cuprite (0.15.1)
+      capybara (~> 3.0)
+      ferrum (~> 0.15.0)
     date (3.3.4)
     diff-lcs (1.5.1)
     docile (1.4.1)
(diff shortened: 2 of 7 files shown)


fdr (approved):
I dig it.

jeremyevans:
I think this is ready for commit now.  I plan to merge as soon as https://github.com/ubicloud/documentation/pull/7 is merged.

jeremyevans:
I merged https://github.com/ubicloud/documentation/pull/7, so I'm going to merge this now.

PR #2099, opened 2024-10-22
Avoid private subnet conflicts at the project level, add dns ipv4 to the dhcp payload and remove static dns conf, add weighted subnet picking and ban some ranges

furkansahin:
## Avoid private subnet conflicts at the project level
We used to enforce subnet uniqueness at the location level. That's not
really necessary and there may be conflicts for subnets in different
projects. This opens up the possibility of conflicts for the subnet
peering but a pretty low possibility. We'll attack the conflict problem
later.

## Add dns ipv4 to the dhcp payload and remove static dns conf
This is necessary because if we don't add the ipv4 address to the dhcp
payload, systemd-resolved kind of resolves it from the interface it gets
the dhcp packs from and adds the first address from the subnet as the
dns ipv4 address because the first address is added to the tap device.

We also remove the static dns configuration because that's unnecessary,
the dns addresses must be configured as part of dhcp.

## Add weighted subnet picking and ban some ranges
This commit adds weighted private ipv4 class picking depending on the
size of the subnet. In addition, we ban 172.16.0.0/16, 172.17.0.0/16,
and 172.18.0.0/16 because these are commonly used by docker and
kubernetes networks.

### DIFF
--- a/model/private_subnet.rb
+++ b/model/private_subnet.rb
@@ -15,6 +15,12 @@ class PrivateSubnet < Sequel::Model
     "192.168.0.0/16"
   ].freeze
 
+  BANNED_IPV4_SUBNETS = [
+    NetAddr::IPv4Net.parse("172.16.0.0/16"),
+    NetAddr::IPv4Net.parse("172.17.0.0/16"),
+    NetAddr::IPv4Net.parse("172.18.0.0/16")
+  ].freeze
+
   dataset_module Pagination
   dataset_module Authorization::Dataset
   include Authorization::HyperTagMethods
@@ -53,7 +59,11 @@ class PrivateSubnet < Sequel::Model
   semaphore :destroy, :refresh_keys, :add_new_nic, :update_firewall_rules
 
   def self.random_subnet
-    PRIVATE_SUBNET_RANGES.sample
+    subnet_dict = PRIVATE_SUBNET_RANGES.each_with_object({}) do |subnet, hash|
+      prefix_length = Integer(subnet.split("/").last, 10)
+      hash[subnet] = (2**16 + 2**12 + 2**8 - 2**prefix_length)
+    end
+    subnet_dict.max_by { |_, weight| rand**(1.0 / weight) }.first
   end
 
   # Here we are blocking the bottom 4 and top 1 addresses of each subnet
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -179,7 +179,7 @@ class Prog::Vnet::SubnetNexus < Prog::Base
 
     selected_addr = addr.nth_subnet(26, SecureRandom.random_number(2**(26 - addr.netmask.prefix_len) - 1).to_i + 1)
 
-    selected_addr = random_private_ipv4(location, project) if banned_subnets.any? { _1.rel(selected_addr) } || project.private_subnets_dataset[Sequel[:net4] => selected_addr.to_s, :location => location]
+    selected_addr = random_private_ipv4(location, project) if PrivateSubnet::BANNED_IPV4_SUBNETS.any? { _1.rel(selected_addr) } || project.private_subnets_dataset[Sequel[:net4] => selected_addr.to_s, :location => location]
 
     selected_addr
   end
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -18,8 +18,8 @@ class Prog::Vnet::SubnetNexus < Prog::Base
     Validation.validate_name(name)
     Validation.validate_location(location)
 
-    ipv6_range ||= random_private_ipv6(location).to_s
-    ipv4_range ||= random_private_ipv4(location).to_s
+    ipv6_range ||= random_private_ipv6(location, project).to_s
+    ipv4_range ||= random_private_ipv4(location, project).to_s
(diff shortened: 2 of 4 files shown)


prog/vnet/subnet_nexus.rb
```diff
@@ -173,8 +173,7 @@ def self.random_private_ipv6(location)
   end
 
   def self.random_private_ipv4(location)
-    private_range = PrivateSubnet.random_subnet
-    addr = NetAddr::IPv4Net.parse(private_range)
+    addr = NetAddr::IPv4Net.parse(PrivateSubnet.const_get(:PRIVATE_SUBNET_RANGE))
```
enescakir, line 177: Why don't we call `PrivateSubnet::PRIVATE_SUBNET_RANGE `

prog/vnet/subnet_nexus.rb
```diff
@@ -163,31 +163,24 @@ def gen_reqid
...
   end
 
-  def self.random_private_ipv6(location)
+  def self.random_private_ipv6(location, project)
     network_address = NetAddr::IPv6.new((SecureRandom.bytes(7) + 0xfd.chr).unpack1("Q<") << 64)
     network_mask = NetAddr::Mask128.new(64)
     addr = NetAddr::IPv6Net.new(network_address, network_mask)
-    return random_private_ipv6(location) unless PrivateSubnet.where(net6: addr.to_s, location: location).first.nil?
+
+    addr = random_private_ipv6(location, project) if project.private_subnets.find { _1.net6.to_s == addr.to_s && _1.location == location }
```
fdr, line 171: `project.private_subnets_dataset.where[Sequel[:net6] => addr.to_s, location: location]` or something like that?

prog/vnet/subnet_nexus.rb
```diff
@@ -163,31 +163,24 @@ def gen_reqid
...
-    # subnets (e.g., 172.18.0.0/16). If a subnet is in use, Docker moves to the
-    # next available one. However, since dnsmasq assigns private IP addresses
-    # post-VM initialization, there's a race condition: Docker can assign its
-    # address before the VM is necessarily aware of its virtual networking
-    # address information, leading to routing clashes that render IPv4 internet
-    # access inoperable in some interleavings.
-    if PrivateSubnet[net4: selected_addr.to_s, location: location] || selected_addr.network.to_s == "172.17.0.0"
-      return random_private_ipv4(location)
+
+    if project.private_subnets.find { _1.net4.to_s == selected_addr.to_s && _1.location == location }
```
fdr: ditto, can use private_subnets_dataset.where...exist?

rhizome/host/lib/vm_setup.rb
```diff
@@ -456,6 +456,7 @@ def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image, dn
 server=2620:fe::fe
 server=2620:fe::9
```
fdr, line 457: >  Add dns ipv4 to the dhcp payload and remove static dns conf 

I would add to this commit message. You can mention the provenance (you found them by inspection) even if they solved no acute problem you know of at the time, and I think you documented at least some small changes to output to commands.

model/private_subnet.rb
```diff
@@ -53,7 +59,11 @@ def display_state
   semaphore :destroy, :refresh_keys, :add_new_nic, :update_firewall_rules
 
   def self.random_subnet
-    PRIVATE_SUBNET_RANGES.sample
+    subnet_dict = PRIVATE_SUBNET_RANGES.each_with_object({}) do |subnet, hash|
+      prefix_length = subnet.split("/").last.to_i
```
fdr, line 63: Consider `Integer(s, 10)` because `.to_i` is always one edit away from "0" given erroneous input. the `10` is necessary to avoid, say, octal conversion.

prog/vnet/subnet_nexus.rb
```diff
@@ -175,11 +175,12 @@ def self.random_private_ipv6(location, project)
 
   def self.random_private_ipv4(location, project)
     private_range = PrivateSubnet.random_subnet
+    banned_subnets = PrivateSubnet::BANNED_IPV4_SUBNETS.map { NetAddr::IPv4Net.parse(_1) }
     addr = NetAddr::IPv4Net.parse(private_range)
 
     selected_addr = addr.nth_subnet(26, SecureRandom.random_number(2**(26 - addr.netmask.prefix_len) - 1).to_i + 1)
 
-    if project.private_subnets.find { _1.net4.to_s == selected_addr.to_s && _1.location == location }
+    if project.private_subnets.find { _1.net4.to_s == selected_addr.to_s && _1.location == location } || banned_subnets.any? { _1.rel(selected_addr) }
```
fdr, line 183: push down into database again? "not exist" with a subquery. LLMs can usually get this right, I always use it rather than cracking open the manual again.

fdr (commented):
some useful database pushdowns (maybe you can also project, plus one to_i avoidance

rhizome/host/lib/vm_setup.rb
furkansahin: Do you like it now?

prog/vnet/subnet_nexus.rb
furkansahin: what about now?

fdr (approved):

PR #2101, opened 2024-10-22
Use Sequel schema_caching and index_caching extensions

jeremyevans:
With these caching extensions, Sequel no longer needs to run multiple queries for every model to get the schema and indexes. The schema and indexes are queried once in a Rake task, and stored in marshal format in the cache directory.  When models are loaded, if entries for the models are already present in the cache, Sequel uses the cached value instead of querying the database.

On my laptop, this shaves off about 0.2 seconds when loading, for about a 10% reduction in load time.

Using this cache reduces the number of database queries when loading the application and all models from 125 to 6.  The 6 remaining queries:

* 3 used for all new PostgreSQL connections
* 1 to get the PostgreSQL server version
* 2 to check whether the accounts table exists (probably in Rodauth)

This can add risk, because if the caches do not match the database, then the models will not have the correct schema or indexes.  To attempt to avoid this, this modifes the test_up and test_down rake tasks to refresh the Sequel caches.  However, developers must remember to check in the caches when committing migrations.

Given that the performance increase of doing this appears to be modest on modern computers, maybe the additional risk of this approach makes it not worth merging.  As Ubicloud grows in terms of number of models, the benefits of this approach increase, but computer performance also increases over time, and the higher the computer performance, the lower the benefit of caching.

Note that if we plan on using the pg_auto_constraint_validations plugin, that has a similar issue to a much greater degree than the queries to get the model schema and indexes. It also supports caching, and we would definitely want to use caching as the performance difference would be more significant.  Since we would be doing caching anyway at that point, we would want to do index and schema caching as well.


fdr:
Would CI be able to verify the up-to-datedness of these caches?

jeremyevans:
> Would CI be able to verify the up-to-datedness of these caches?

CI would run the specs with the cached index/schema, so if the cached values were out of sync (e.g. PR adding migration making schema changes without committing cache updates), CI would likely fail, assuming that there is any spec that depends on model having accurate schema information.

We could have CI also regenerate the schema and compare to the cached version.  However, I'm not sure the same schema will always generate the same cache file.  I have code in Sequel to try to make that more likely (mostly by sorting things when building the cache), but I'm not sure it's guaranteed.

fdr (approved):
100% branch coverage probably does the job re: old caches, as you say.

PR #2102, opened 2024-10-22
Add LB custom hostname feature to API

furkansahin:


routes/common/load_balancer_helper.rb
```diff
@@ -41,6 +41,16 @@ def post(name: nil)
       end
     end
     Authorization.authorize(@user.id, "PrivateSubnet:view", ps.id)
+    prefix, dns_zone_id = if (custom_hostname = request_body_params["custom_hostname"])
+      dns_zone_name = custom_hostname.split(".")[1..].join(".")
+      prefix = custom_hostname.split(".")[0]
+      dns_zone = DnsZone.where(name: dns_zone_name).first
```
fdr, line 47:
```suggestion
      dns_zone = DnsZone.where[name: dns_zone_name]
```

spec/routes/api/project/location/load_balancer_spec.rb
```diff
@@ -114,6 +114,33 @@
...
+        post "/api/project/#{project.ubid}/load-balancer/lb2", {
+          private_subnet_id: ps.ubid,
+          src_port: "80", dst_port: "80",
+          health_check_endpoint: "/up", algorithm: "round_robin",
+          health_check_protocol: "http", custom_hostname: "lb2.test.com"
+        }.to_json
+
+        expect(last_response.status).to eq(200)
+        expect(LoadBalancer.where(name: "lb2").first.hostname).to eq("lb2.test.com")
+        expect(LoadBalancer.where(name: "lb2").first.dns_zone.id).to eq(dz.id)
```
fdr, line 130:
```suggestion
        expect(last_response.status).to eq(200)
        
        lb = LoadBalancer.where[name: "lb2"]
        expect(lb.hostname).to eq("lb2.test.com")
        expect(lb.dns_zone.id).to eq(dz.id)
```

routes/common/load_balancer_helper.rb
```diff
@@ -41,6 +41,16 @@ def post(name: nil)
       end
     end
     Authorization.authorize(@user.id, "PrivateSubnet:view", ps.id)
+    prefix, dns_zone_id = if (custom_hostname = request_body_params["custom_hostname"])
+      dns_zone_name = custom_hostname.split(".")[1..].join(".")
+      prefix = custom_hostname.split(".")[0]
```
fdr, line 46:
```suggestion
      prefix, dns_zone_name = custom_hostname.split(".", 2)
```

routes/common/load_balancer_helper.rb
byucesoy, line 47: This allows user to modify arbitrary dns zone that they don't necessarily own.

Load balancer's custom_hostname_dns_zone logic is quite scary. (Sorry for just catching that, I missed the original PR). I'm not comfortable how it is possible to pass an arbitrary DnsZone id to LoadBalancerNexus and it takes control of the DnsZone. DnsZones are like the keys of the whole house. It is possible to do massive damage using them.

One proper way could be load balancer to create its own DnsZone and manage only that.

byucesoy (changes requested):

routes/common/load_balancer_helper.rb
fdr: Yeah, I was staring at that and asking furkan a few questions about the scoping of this. seems hairy as-is

furkansahin: Please recheck the commit message.

PR #2103, opened 2024-10-22
Postgres 17 Support

byucesoy:
Most changes in this PR are pretty straightforward and it is easy to follow the overall logic because I created micro commits. However, there are two non-obvious choices we made, so I want to write them down.

- We generate separate images for PG16 and PG17. The primary motivation is keeping the image size small and also allowing installing  (not just downloading) binaries during image generation. Both speeds up the provisioning time. Creating separate images increases the operational load a bit, but I think we can automate it end to end if it becomes too annoying. 
- We pass version from control plane to data plane when it is needed. Technically, data plane can deduce its version by checking the filesystem, but we might have both directories in the future (thanks to upgrades). It seems like passing this information from control plane is more future proof.


rhizome/postgres/bin/configure
```diff
@@ -4,6 +4,11 @@
 require "json"
 require_relative "../../common/lib/util"
 
+if ARGV.count != 1
+  fail "Wrong number of arguments. Expected 1, Given #{ARGV.count}"
+end
+
+v = ARGV[0]
```
enescakir, line 11: Should we cast it to an integer anyway?

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -40,13 +43,13 @@ def self.assemble(project_id:, location:, name:, target_vm_size:, target_storage
             parent.timeline.latest_restore_time && restore_target <= parent.timeline.latest_restore_time
           fail Validation::ValidationFailed.new({restore_target: "Restore target must be between #{earliest_restore_time} and #{parent.timeline.latest_restore_time}"})
         end
-        [parent.superuser_password, parent.timeline.id, "fetch"]
+        [parent.superuser_password, parent.timeline.id, "fetch", parent.version]
```
enescakir, line 46: Should we trigger an error when the parent and version are provided at the same time? Otherwise, the given version might be silently ignored, potentially leading to operator mistakes.

views/postgres/create.erb
```diff
@@ -174,6 +174,38 @@
                 </div>
               </div>
 
+              <div class="col-span-full">
+                <div class="space-y-2">
+                  <label for="size" class="text-sm font-medium leading-6 text-gray-900">Version</label>
+                  <fieldset class="radio-small-cards" id="size-radios">
+                    <legend class="sr-only">Version</legend>
+                    <div class="grid gap-3 grid-cols-1 md:grid-cols-2 xl:grid-cols-3">
+                      <% Option::PostgresVersionOptions.select { _1.family == @flavor }.map(&:version).each_with_index do |version, idx| %>
```
enescakir, line 183: What if we keep `Option::PostgresVersionOptions` as hash and select versions by O(1) with a key instead of select and map the list?

views/postgres/create.erb
```diff
@@ -174,6 +174,38 @@
                 </div>
               </div>
 
+              <div class="col-span-full">
```
enescakir, line 177: Why don't we use `radio-small-cards` component? Does it have any customization?

views/postgres/create.erb
```diff
@@ -174,6 +174,38 @@
...
+                <div class="space-y-2">
+                  <label for="size" class="text-sm font-medium leading-6 text-gray-900">Version</label>
+                  <fieldset class="radio-small-cards" id="size-radios">
```
enescakir, line 180: `id="version-radios"> `

views/postgres/create.erb
```diff
@@ -174,6 +174,38 @@
                 </div>
               </div>
 
+              <div class="col-span-full">
+                <div class="space-y-2">
+                  <label for="size" class="text-sm font-medium leading-6 text-gray-900">Version</label>
```
enescakir, line 179: `for="version"`

enescakir (approved):

rhizome/postgres/bin/configure
byucesoy: I think we shouldn't. Although we use only the major part currently, version numbers can contain non-digit characters such as "16.1" or "16-rc"

views/postgres/create.erb
byucesoy: We use it in a child element. Do you want to move it above?

prog/download_boot_image.rb
```diff
@@ -55,7 +55,7 @@ def url
       elsif download_from_blob_storage?
         suffixes = {
           "github" => "raw",
-          "postgres" => "raw",
+          "postgres16" => "raw",
```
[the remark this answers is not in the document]
byucesoy: Sounds good. I'll update this manually after the deployment.

enescakir (approved):

PR #2108, opened 2024-10-24
Fix suspend bypass via remember token

enescakir:
When we suspend an account, we set the `suspended_at` value and check it in the `before_login` method. However, the remember plugin doesn't go through this method, allowing the account to log in with the remember token even if it is suspended. @jeremyevans suggested using the `update_session` method so that every plugin eventually hits this method for login.

We call the `forget_login` method to remove the remember token; otherwise, Rodauth will keep redirecting back to the same page due to the remember token.


byucesoy:
Were you able to reproduce the behavior?

jeremyevans (approved):

PR #2114, opened 2024-10-24
Support running specs with frozen environment

jeremyevans:
This allows for running the specs with a frozen environment that should closely match the production environment.

This adds the following rake tasks:

* coverage: Runs specs in serial with coverage
* frozen_spec: Runs specs in serial in frozen environment
* frozen_pspec: Runs specs in parallel in frozen environment

It changes the default rake task to run both the coverage and frozen_spec tasks.  It also changes the GitHub CI workflow to use the default rake task.

When specs are run in the frozen environment they call clover_freeze before running the specs, after loading all other code.

There are currently 56 specs that do not work in the frozen environment.  They are skipped by including the skip_if_frozen method in each spec.


jeremyevans:
I should point out that this does not run the specs with frozen models, only with a frozen environment.  There are another 116 specs that need to be skipped for the specs to pass with frozen models.

fdr (approved):

PR #2119, opened 2024-10-25
De-duplicate one instance of self signed certificate code

fdr:
Use the `Util` library module to do this, which will both speed things up, and eliminate duplicate mechanism.

Elliptic curves require far less (very slow) entropy than RSA, and this was noticeable in the tests.


fdr:
Going to test it all in one shot with #2122

PR #2121, opened 2024-10-25
Remove use of Helsinki as default

byucesoy:
To complete the removal of the Helsinki region, we also need to create a new E2E tests host in hetzner-fsn1. However, it is not urgent as the current host would think that it is in hetzner-fsn1 anyway and continue to work.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -868,7 +868,7 @@ components:
       required: true
       schema:
         type: string
-        example: eu-north-h1
+        example: eu-central-h1
     order_column:
       description: Pagination - Order column
       in: query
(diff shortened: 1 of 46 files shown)


aamederen:
:shipit:

fdr:
Probably need to phase this patch, to ship
1) change test default fixture
2) prevent new allocations
3) <after no dependency> no more fsn1 code

fdr (commented):
clearing my review queue until we're ready to ship it in phases.

byucesoy:
> Probably need to phase this patch, to ship
> 
> 1. change test default fixture
> 2. prevent new allocations
> 3. after no dependency in prodcution, no more hel1 code

I changes this PR so that we only
- Start to default to hetzner-fsn1 in the places where we default to a region
- Updated tests to use non-helsinki regions.

We still display Helsinki as an option to user. I'll remove that option after #2/#3 is completed.

fdr (approved):

PR #2122, opened 2024-10-25
Switch certificate/acme code to Elliptic Curves for speedup (esp in test)

fdr:
This speedup is on one test suite alone, which was among the most expensive in this regard: 3.8x

```
Benchmark 1: bundle exec rspec ./spec/prog/vnet/cert_nexus_spec.rb
  Time (mean ± σ):      1.539 s ±  0.096 s    [User: 1.183 s, System: 0.174 s]
  Range (min … max):    1.434 s …  1.748 s    10 runs

Benchmark 1: bundle exec rspec ./spec/prog/vnet/cert_nexus_spec.rb
  Time (mean ± σ):      5.739 s ±  1.733 s    [User: 5.396 s, System: 0.181 s]
  Range (min … max):    2.959 s …  8.888 s    10 runs
```

But, it's somewhat bad to be using expensive entropy unnecessarily in tests: it likely be much faster still, with more work, by re-using key material.


prog/vnet/cert_nexus.rb
```diff
@@ -64,7 +64,7 @@ def before_run
     when "pending"
       nap 10
     when "valid"
-      csr_key = OpenSSL::PKey::RSA.new(4096)
+      csr_key = OpenSSL::PKey::EC.generate("prime256v1")
       csr = Acme::Client::CertificateRequest.new(private_key: csr_key, common_name: cert.hostname)
```
fdr, line 68: this does change our non-test behavior; P256 is common, is it okay with the acme certificaterequest?

prog/vnet/cert_server.rb
```diff
@@ -50,7 +50,7 @@ def put_cert_to_vm
     cert = load_balancer.active_cert
 
     cert_payload = cert.cert
-    cert_key_payload = OpenSSL::PKey::RSA.new(cert.csr_key).to_pem
+    cert_key_payload = OpenSSL::PKey::EC.new(cert.csr_key).to_pem
```
fdr, line 53: may have backwards compatibility issues here

jeremyevans:
Long term, it seems good to switch from RSA to EC. I'm not sure how the public certificate authorities handle EC keys, though.

As you mentioned, this changes production and not just test. For a lesser speedup without a behavior change, we could use lower-bit RSA keys in test and RSA 4096-bit keys in production.  How low we can go depends on the OpenSSL configuration.  The OpenSSL minimum (SSL_MIN_RSA_MODULUS_LENGTH_IN_BYTES) is 512 bit keys (scandalous).  I know the minimum allowed by LibreSSL is 2048. 

Here are the speedups:

```ruby
puts Benchmark.measure{3.times{OpenSSL::PKey::RSA.new(4096)}}
# 4.248120   0.000000   4.248120 (  4.248525)
                                                                    
puts Benchmark.measure{3.times{OpenSSL::PKey::RSA.new(2048)}}
# 0.680071   0.000000   0.680071 (  0.680102)
                                                            
puts Benchmark.measure{3.times{OpenSSL::PKey::RSA.new(1024)}}
# 0.035811   0.000000   0.035811 (  0.035820)
                                                   
 puts Benchmark.measure{3.times{OpenSSL::PKey::RSA.new(512)}}
# 0.018038   0.000000   0.018038 (  0.018034)

```

fdr:
> Long term, it seems good to switch from RSA to EC. I'm not sure how the public certificate authorities handle EC keys, though.

I think they support them, though it's a bit hard to try both ends of this without a live DNS server to complete the challenge (hello @furkansahin )

https://datatracker.ietf.org/doc/html/rfc8555 mandates P-256 (named as ES256) support at least for account keys. Note acme has both the account key and the customary certificate signing private key that are distinct keys, and could have distinct ciphers, too.

https://wiki.mozilla.org/Security/Server_Side_TLS has been my reference on issues of practical uses of ciphers like this, after many years, In recent years, P-256 has crept to their all-but-lowest/most desperate rung of support.

furkansahin (approved):
I have tested this locally with a dns setup + certificates + a simple web server. This works. 🚢

PR #2123, opened 2024-10-25
Add ruby-lsp to Gemfile

fdr:
Will Leinweber told me I was missing out if I wasn't using this.  I gave it a try, unless my previous (few years ago) experiences with lsp servers in emacs for ruby, this one worked immediately, having tested finding declarations and references for a couple of symbols.

So, make it easier for everyone by adding it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,7 @@ group :development do
   gem "rubocop-rake"
   gem "rubocop-rspec"
   gem "rubocop-sequel"
+  gem "ruby-lsp", require: false
   gem "standard", ">= 1.24.3"
   gem "simplecov"
   gem "turbo_tests"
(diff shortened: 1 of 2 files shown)


fdr:
I suppose in principle exact ruby-lsp version doesn't matter and perhaps people should `gem install` it on their own, rather than gemfile-ing it.

enescakir (approved):
I'm using it for several months with Zed. I didn't need to put it to Gemfile btw

https://zed.dev/docs/languages/ruby#language-servers

jeremyevans:
Late to respond here, but I agree that we shouldn't include it in `Gemfile` unless there is some code that actually uses it.  There is no reason to include it just to make it available.

PR #2128, opened 2024-10-28
Add Debian 12 support

fdr:


prog/download_boot_image.rb
```diff
@@ -34,23 +34,27 @@ def url
...
-        arch = (vm_host.arch == "x64") ? "amd64" : "arm64"
+        arch = vm_host.render_arch(arm64: "arm64", x64: "amd64")
         "https://cloud-images.ubuntu.com/releases/noble/release-#{version}/ubuntu-24.04-server-cloudimg-#{arch}.img"
       elsif image_name == "ubuntu-jammy"
-        arch = (vm_host.arch == "x64") ? "amd64" : "arm64"
+        arch = vm_host.render_arch(arm64: "arm64", x64: "amd64")
         "https://cloud-images.ubuntu.com/releases/jammy/release-#{version}/ubuntu-22.04-server-cloudimg-#{arch}.img"
+      elsif image_name == "debian-12"
+        arch = vm_host.render_arch(arm64: "arm64", x64: "amd64")
+        "https://cloud.debian.org/images/cloud/bookworm/#{version}/debian-12-genericcloud-#{arch}-#{version}.raw"
```
enescakir, line 52: Since E2E tests and production download images from the internal minio, should I put it to minio if you don't plan to change the version

model/vm_host.rb
```diff
@@ -272,4 +272,15 @@ def available_storage_gib
   def total_storage_gib
     storage_devices.sum { _1.total_storage_gib }
   end
+
+  def render_arch(arm64:, x64:)
```
enescakir, line 276: Since the rhizome version has default values, should we also include default values to keep consistency across the interfaces?

https://github.com/ubicloud/ubicloud/blob/4dc5c281c9a0c93ad551739d015b91f4aaa8e903/rhizome/common/lib/arch_class.rb#L25

prog/download_boot_image.rb
fdr: I don't plan to change it, please do feel free.

model/vm_host.rb
fdr: I think the defaults are never used in this case, so probably not (fair question: did the original ever use its defaults?)

prog/download_boot_image.rb
jeremyevans: I'm not sure what our past practice is, but I would think we would want to use the official source as the default, and support overriding via environment variables or in `.env.rb`.  Our hosted version of Ubicloud would want to use the internal minio to improve performance and reduce external bandwidth, but we wouldn't want that used by someone hosting their own version of Ubicloud (I assume).

fdr: yeah, it is the default, but our production and E2E environment is flagged into the accelerated variant. I develop using the source as-is..

fdr: btw @pykello
 
```
    # YYY: In future all images should be checked for sha256 sum, so the nil
    # default will be removed.
    hashes.fetch([image_name, vm_host.arch, version], nil)
```

is the future now? I noticed in development that my thing worked even though I hadn't encoded a hash yet and was surprised.

jeremyevans (commented):
I don't think I have the context yet to properly review this, but everything looks good from a code perspective.

prog/download_boot_image.rb
enescakir: It decides here: https://github.com/ubicloud/ubicloud/blob/4dc5c281c9a0c93ad551739d015b91f4aaa8e903/prog/download_boot_image.rb#L17. We may convert it to an environment variable flag.

I use this adhoc workflow to upload new versions from official sources to Minio: https://github.com/ubicloud/runner-images/blob/default-images/.github/workflows/ubicloud-image.yml

Screenshot 2024-10-27 at 20 10 46

enescakir: Here we go. I've verified the checksums and they match
Screenshot 2024-10-27 at 20 52 14

enescakir (approved):

PR #2129, opened 2024-10-28
Upload coverage report with different artifact names in CI

enescakir:
When the CI fails, we upload the coverage report as an artifact.

Previously, `actions/upload-artifact` allowed uploading artifacts with the same name in different jobs in v3, but v4 doesn't allow this [^1]. We need to upload artifacts with different names; otherwise, it fails with a 409 error. It seems we overlooked this change when upgrading from v3 to v4.

[^1]: https://github.com/actions/upload-artifact?tab=readme-ov-file#not-uploading-to-the-same-artifact

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,5 +101,5 @@ jobs:
       if: failure()
       uses: actions/upload-artifact@v4
       with:
-        name: code-coverage-report
+        name: code-coverage-report-${{ matrix.runs-on }}
         path: coverage/


prog/vm/github_runner.rb
```diff
@@ -8,7 +8,7 @@ class Prog::Vm::GithubRunner < Prog::Base
   semaphore :destroy, :skip_deregistration
 
   def self.assemble(installation, repository_name:, label:, default_branch: nil)
-    unless Github.runner_labels[label]
+    unless   Github.runner_labels[label]
```
fdr, line 11: seems weird (and surprised rubocop isn't modifying it...is it about to fail the linter in ci?)

enescakir: Yeah, I just tried to fail CI to see if uploading works. I'm uncommiting it now
Screenshot 2024-10-27 at 20 16 10

fdr (approved):

prog/vm/github_runner.rb
enescakir: Sorry for the confusion. I remembered to test it after creating the pull request :smile:

PR #2130, opened 2024-10-28
Ankush Disk Monitoring

furkansahin:
## Migration for block device serail number column in storage_device
The new column `blk_dev_serial_number` would contain the serial number
of the block storage device. The column has type `text[]` and not just
`text` to accomodate RAID setups.

## Add disk monitoring in VM host
This commit contains changes for disk monitoring for both RAID and
non-RAID systems, supporting NVME and SCSI storage controller
interfaces. The system is designed to identify and report disk failures
in a pull-based monitoring setup.

Storage Device Model

To identify which disks are down in a pull-based system, we first need
to maintain the record of existing disks. We introduced a column in the
`storage_device` model to store the list of physical disks mapped to
each logical disk.

Note: Entries in the storage_device table do not represent physical
storage. For example, a RAID1 setup would only have a single entry in
the table despite being backed by multiple physical storage devices. Due
to this, the newly added column is a collection, not a single entry.

Device Identification

We decided to store the device's serial number rather than using more
legible device file naming (e.g., /dev/nvme0n1). This choice was made
due to the non-permanent nature of device file naming. Device names tend
to change on reboot or when ports on which they are attached to the
machine are physically swapped.

Mountpoint and Physical Device Association

To find the associated physical device for each storage_device table
entry, we leverage mountpoint information. The mountpoint can be
inferred from the name column of storage_device:
- If the name is `DEFAULT`, the disk is mounted at `/`.
- Otherwise, it is mounted at `/var/storage/devices/<name>`.

Precedence for this can be found [in this
commit](https://github.com/ubicloud/ubicloud/commit/89f9dbd7f2b12c08eb993d947e64003c68c90606#diff-f336d7f0af53644f9dae8673dcebdaab01a2d3173a00ba648fc48d79aee2ba81R19).

After finding the associated mountpoint, we parse lsblk to get the
serial number of the physical device.

Health Check Process

On each check pulse run: 1. We fetch the device file name for each
serial number.  2. We run a health check using smartmontools.  3. If any
device fails, it fails the overall `check_pulse`.

We supplement smartmontools with another check to detect if the OS has
remounted the disk as read-only. This occurs when the OS observes too
many IO failures, among other reasons. To check if the device is
read-only, we rely on the `lsblk` command.

Smartmontools doesn't cache the health check result. It works for both
NVME and SCSI, with differences in output formatting:
- For NVME: We see "PASSED"
- For SCSI: We see "OK"

This is cross-checked from their source code on
[GitHub](https://github.com/smartmontools/smartmontools/blob/65707ca59285a4b2350fb5b30293398f17e868bd/smartmontools/scsiprint.cpp#L402).
Also, smartmontools are not part of linux distributions. It needs to be
installed. We do this in `rhizome/host/bin/prep_host.rb`.

Failure Modes

1. SSH fails:
   - If due to binary not found: The disk on which SSH resides gets
     detached. Systems are expected to survive such events.
   - If due to other reasons (e.g., machine busy or out of resources):
     We fail the check.

2. `smartctl` binary not found:
   - This could be due to disk detachment. We're currently ignoring this
     scenario.
   - Reasoning: The smartctl binary's page should be in memory since we
     run it every 30 seconds. However, we can't make assumptions about
client usage.
   - Potential solution: Pin the page to RAM using `vmtouch`.

3. `smartctl` command fails:
   - We fail the check in this case.

These checks are run per storage as part of the VmHost monitoring,
reusing the strand.

Other Considerations

- Populating serial numbers in the storage_device table:
  - Option 1: Manually run a script at new deployment time. This
    requires considering failure retry mechanisms and handling permanent
failures.
  - Option 2 (Chosen): Anchor population to the users of this column.
    When the storage_device model is asked for a serial number it
doesn't have, it runs an SSH command to find the serial number,
populates the table, and returns to the caller.

- Testing the mountpoint to serial number mapping:
  - We created loopback devices, which added more entries to the lsblk
    output.
  - Running the logic for fetching serial numbers from the control plane
    retrieved the expected serial numbers.
  - The script used to create loopback fs is [mentioned in this commit
    message](https://github.com/ubicloud/ubicloud/commit/89f9dbd7f2b12c08eb993d947e64003c68c90606#diff-f336d7f0af53644f9dae8673dcebdaab01a2d3173a00ba648fc48d79aee2ba81R19).

- LearnStorage update:
  - We updated LearnStorage to populate serial numbers in
    storage_device.
  - LearnStorage runs as part of the PrepHost program, which is executed
    on machine changes like new disk attachments.

- Disk CRUD:
  - Removal of disk and addition of new disk to existing RAID systems is
    still to be determined (TBD).
  - Addition of new disk is handled by `LearnStorage` prog.

- Transaction consideration:
  - A transaction is not required for populating serial numbers in
    storage_device as the race condition is benign.

## Added parallel health check in host_nexus to automatically recover

### DIFF
--- a/migrate/20241028_add_blk_dev_in_storage_devices.rb
+++ b/migrate/20241028_add_blk_dev_in_storage_devices.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    alter_table(:storage_device) do
+      add_column :blk_dev_serial_number, "text[]"
+    end
+  end
+
+  down do
+    alter_table(:storage_device) do
+      drop_column :blk_dev_serial_number
+    end
+  end
+end
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -251,14 +251,24 @@ class VmHost < Sequel::Model
 
   def check_pulse(session:, previous_pulse:)
     reading = begin
-      session[:ssh_session].exec!("true")
-      "up"
-    rescue
+      all_bdevs = storage_devices.flat_map(&:blk_dev_serial_number)
+      commands = all_bdevs.map do |serial|
+        "sudo smartctl -H /dev/disk/by-id/$(ls /dev/disk/by-id/ | grep -E '#{serial}$') | grep -qE 'OK|PASSED' && echo 'up' || echo 'down'"
+      end
+      stdout = session[:ssh_session].exec!(commands.join("; "))
+      all_up = stdout.split("\n").all? { |res| res.strip == "up" }
+      all_up ? "up" : "down"
+    rescue => ex
+      Clog.emit(ex)
       "down"
     end
+
     pulse = aggregate_readings(previous_pulse: previous_pulse, reading: reading)
 
-    if pulse[:reading] == "down" && pulse[:reading_rpt] > 5 && Time.now - pulse[:reading_chg] > 30 && !reload.checkup_set?
+    if pulse[:reading] == "down" &&
+        pulse[:reading_rpt] > 5 &&
+        Time.now - pulse[:reading_chg] > 30 &&
+        !reload.checkup_set?
       incr_checkup
     end
 
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -250,29 +250,30 @@ class VmHost < Sequel::Model
   end
 
   def check_pulse(session:, previous_pulse:)
-    reading = begin
-      all_bdevs = storage_devices.flat_map(&:blk_dev_serial_number)
-      commands = all_bdevs.map do |serial|
-        "sudo smartctl -H /dev/disk/by-id/$(ls /dev/disk/by-id/ | grep -E '#{serial}$') | grep -qE 'OK|PASSED' && echo 'up' || echo 'down'"
-      end
-      stdout = session[:ssh_session].exec!(commands.join("; "))
-      all_up = stdout.split("\n").all? { |res| res.strip == "up" }
-      all_up ? "up" : "down"
-    rescue => ex
-      Clog.emit(ex)
-      "down"
-    end
-
+    reading = run_health_checks(session) ? "up" : "down"
     pulse = aggregate_readings(previous_pulse: previous_pulse, reading: reading)
-
     if pulse[:reading] == "down" &&
         pulse[:reading_rpt] > 5 &&
         Time.now - pulse[:reading_chg] > 30 &&
         !reload.checkup_set?
       incr_checkup
     end
-
     pulse
+  end
+
+  def run_health_checks(session)
+    all_bdevs = storage_devices.flat_map(&:blk_dev_serial_number)
+    commands = all_bdevs.map do |serial|
+      # Health check for each storage device
+      "sudo smartctl -H /dev/disk/by-id/$(ls /dev/disk/by-id/ | grep -E '#{serial}$' | head -n1) | grep -qE 'OK|PASSED' && echo 'up' || echo 'down'"
+    end
+    begin
+      stdout = session[:ssh_session].exec!(commands.join("; "))
+      stdout.split("\n").all? { |res| res.strip == "up" }
+    rescue => ex
+      Clog.emit(ex)
+      false
+    end
   end
 
   def available_storage_gib
--- a/rhizome/host/bin/prep_host.rb
+++ b/rhizome/host/bin/prep_host.rb
@@ -89,6 +89,7 @@ r "apt-get -y install qemu-utils mtools acl"
 # debugging crashes.
 r "apt-get -y install nvme-cli systemd-coredump" if is_prod_env
 
+r "apt-get update && apt-get -y install smartmontools"
 SpdkSetup.prep
 
 # cron job to store serial.log files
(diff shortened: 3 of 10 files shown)


migrate/20241028_add_blk_dev_in_storage_devices.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
```
fdr, line 2: "serail" in commit

rhizome/host/bin/prep_host.rb
```diff
@@ -89,6 +89,7 @@
 # debugging crashes.
 r "apt-get -y install nvme-cli systemd-coredump" if is_prod_env
 
+r "apt-get update && apt-get -y install smartmontools"
```
fdr, line 92: should this be guarded by is_prod_env as well? I'm looking at `nvme-cli` and thinking, "but isn't that pretty similar?"

spec/prog/learn_storage_spec.rb
```diff
@@ -4,6 +4,10 @@
 
 RSpec.describe Prog::LearnStorage do
   describe "#start" do
+    before do
+      allow_any_instance_of(StorageDevice).to receive(:populate_blk_dev_serial_number).and_return(nil) # rubocop:disable RSpec/AnyInstance
+    end
+
```
fdr, line 10: Is there any way to rewrite this without it? it's pretty smelly.

model/vm_host.rb
```diff
@@ -251,14 +251,24 @@ def init_health_monitor_session
...
-    rescue
+      all_bdevs = storage_devices.flat_map(&:blk_dev_serial_number)
+      commands = all_bdevs.map do |serial|
+        "sudo smartctl -H /dev/disk/by-id/$(ls /dev/disk/by-id/ | grep -E '#{serial}$') | grep -qE 'OK|PASSED' && echo 'up' || echo 'down'"
+      end
+      stdout = session[:ssh_session].exec!(commands.join("; "))
+      all_up = stdout.split("\n").all? { |res| res.strip == "up" }
+      all_up ? "up" : "down"
+    rescue => ex
+      Clog.emit(ex)
```
fdr: I don't think that's how `Clog.emit` is used with exceptions. You banner it with a static string and then maybe use `Util.exception_to_hash(ex)`

model/vm_host.rb
```diff
@@ -250,31 +250,32 @@ def init_health_monitor_session
...
+    all_bdevs = storage_devices.flat_map(&:blk_dev_serial_number)
+    commands = all_bdevs.map do |serial|
+      # Health check for each storage device
+      "sudo smartctl -H /dev/disk/by-id/$(ls /dev/disk/by-id/ | grep -E '#{serial}$' | head -n1) | grep -qE 'OK|PASSED' && echo 'up' || echo 'down'"
+    end
+    begin
+      stdout = session[:ssh_session].exec!(commands.join("; "))
+      stdout.split("\n").all? { |res| res.strip == "up" }
+    rescue => ex
+      Clog.emit(ex)
```
fdr, line 274: ditto

aamederen (commented):
This is pretty good work overall, but more complex than necessary in my opinion. In summary:
- Checking Smart bits is a good practice nevertheless, but could be excessive for a frequent health pulse.
- SMART bits have false-positive false-negative scenarios. When it's positive, it's always a good idea to replace the host, maybe not as an immediate need. It's also possible that the drive goes bad but SMART is fine.
- I didn't get the reason for caching the serial numbers of the disks in DB. IMO it's an extra state that's not needed.

My preference would be to go with something like #2060 and add SMART check as a non-emergency-but-important (somehow like a Sev3 in IcM terms) data point.

model/vm_host.rb
```diff
@@ -251,14 +251,24 @@ def init_health_monitor_session
 
   def check_pulse(session:, previous_pulse:)
     reading = begin
```
fdr: re, commit message:

> 2. `smartctl` binary not found:
>    - This could be due to disk detachment. We're currently ignoring this
>      scenario.
>    - Reasoning: The smartctl binary's page should be in memory since we
>      run it every 30 seconds. However, we can't make assumptions about
> client usage.
>    - Potential solution: Pin the page to RAM using `vmtouch`.

Is that actually a failure mode that exists? Is a file system really going to detach references to files because of a storage device gone bad, rather than yield an EIO errno? I am somewhat skeptical, but I don't know for sure.

Is vmtouch really going to solve that problem? For starters, smartctl maps some libraries that'd need to be kept resident as well in whole:

```
	linux-vdso.so.1 (0x00007ffd737b6000)
	libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 (0x000077a8d16a8000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x000077a8d1400000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000077a8d13d2000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000077a8d1000000)
	libpcre2-8.so.0 => /lib/x86_64-linux-gnu/libpcre2-8.so.0 (0x000077a8d1338000)
	/lib64/ld-linux-x86-64.so.2 (0x000077a8d17fc000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000077a8d124f000)
```

Not to mention, the shell (bash) that this is probably run from, and swap, which can get involved for mutable memories, and that messes with a block device as well.

Could vmtouch make this somewhat better? Perhaps, but it seems incredibly marginal.

PR #2131, opened 2024-10-28
Download cache proxy source code and run it from source

velioglu:
As we are facing with an issue related to caching docker layers on prod hosts but couldn't reproduce it on local ones, updating control plane code to download source code of the cache proxy code and run it. It won't be affecting any customer workflow as transparant caching is behind a feature flag. Once the tests will be completed, logic will be reverted back.


furkansahin:
Isn't there a way to test this in development environment without merging?

furkansahin (approved):

enescakir (approved):

PR #2132, opened 2024-10-28
Update address for blob storage cache

velioglu:
While restoring a cache, we were using localhost.blob.core.windows.net to be able to concurrently download files. Though, with it's latest version, actions/cache repo adds a special check to hosts including `localhost` https://github.com/actions/cache/pull/1474/files and it broke our logic. To fix it and not face with a similar issue again updating address name.

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -437,7 +437,7 @@ DHCP
     interfaces = nics.map { "interface=#{_1.tap}" }.join("\n")
     dnsmasq_address_ip6 = NetAddr::IPv6.parse("fd00:0b1c:100d:53::")
     address_mapping = boot_image.include?("github") ?
-      "address=/localhost.blob.core.windows.net/#{nics.first.net4.split("/").first}" :
+      "address=/ubicloudhostplaceholder.blob.core.windows.net/#{nics.first.net4.split("/").first}" :
       ""
     vp.write_dnsmasq_conf(<<DNSMASQ_CONF)
 pid-file=


furkansahin:
This is interesting, it looks like they check suffix but we add it as a prefix, no?

velioglu:
> This is interesting, it looks like they check suffix but we add it as a prefix, no?

Yes, you are right, we are adding it as suffix. Though, it still, somehow, broke the restore part and fixed after changing it.

rhizome/host/lib/vm_setup.rb
```diff
@@ -437,7 +437,7 @@ def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image, dn
     interfaces = nics.map { "interface=#{_1.tap}" }.join("\n")
     dnsmasq_address_ip6 = NetAddr::IPv6.parse("fd00:0b1c:100d:53::")
     address_mapping = boot_image.include?("github") ?
-      "address=/localhost.blob.core.windows.net/#{nics.first.net4.split("/").first}" :
+      "address=/ubicloudhostplaceholder.blob.core.windows.net/#{nics.first.net4.split("/").first}" :
```
enescakir, line 440: maybe `ubicloudlocal`

enescakir (approved):

rhizome/host/lib/vm_setup.rb
velioglu: Keeping it as it is to not face any issues further in case "local" is checked somehow in future

PR #2137, opened 2024-10-28
Add CloverBase#current_user

jeremyevans:
Instead of setting @current_user twice in the main CloverWeb route block and once in the CloverApi route block, add a method that accesses it lazily, and sets it if it has not already been set.

This broke one spec that expected create_account when Account.[] raised an error.  Remove the mocking and define a route only in the test environment that can be used when you want to simulate an exception.

Changes to the routes and views made with:

```
sed -i -e 's/@current_user/current_user/' `git grep -l '@current_user' routes views`
```

This is potentially slower, as method calls are slower than instance variable accesses.  However, this avoids multiple queries for users logged in via the remember token, and queries are way more expensive than either method calls or instance variable accesses.

For CloverApi, since the user was only set once, there is no performance benefit from doing this.  However, since the long term plan is to merge CloverWeb and CloverApi, and this change would be necessary in that case, I took care of CloverApi in this commit.

### DIFF
--- a/clover_api.rb
+++ b/clover_api.rb
@@ -87,9 +87,6 @@ class CloverApi < Roda
     r.rodauth
     rodauth.check_active_session
     rodauth.require_authentication
-
-    @current_user = Account[rodauth.session_value]
-
     r.hash_branches("")
   end
 end
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -365,12 +365,20 @@ class CloverWeb < Roda
   end
 
   hash_branch("after-login") do |r|
-    if (project = @current_user.projects_dataset.order(:created_at).first)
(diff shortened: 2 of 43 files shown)


fdr (approved):

jeremyevans:
Rebased on master to fix the conflict.  This allows testing an additional spec in frozen mode.

enescakir:
We interchangeably use `user` and `account` in the codebase. However, it might be better to standardize everything to `account` because the model is `Account`.

jeremyevans:
@enescakir I'm OK with that change, and I can make it in a subsequent commit.  I think it's better to separate the commits, for for the instance variable to method, and one for the method renaming.

PR #2139, opened 2024-10-28
Rename current_user to current_account

jeremyevans:
This was recommended by Enes, and is more aligned with the model naming (the model is Account).

This was a completely mechanical change, using the following code:

```
sed -i -e 's/current_user/current_account/' `git grep -l 'current_user' clover*.rb routes/clover_base.rb routes views`
```

This makes some naming more consistent, such as `account_ubid: current_account.ubid`.

This makes some naming less consistent, such as `user: current_account`

We could fix the less consistent naming by changing `Routes::Common::Base#initialize` to use `account` instead of `user` as a keyword argument.  This commit doesn't make that change.  It's likely Routes::Common will go away when merging CloverWeb and CloverApi, so I'm not sure it's worth it.

### DIFF
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -365,7 +365,7 @@ class CloverWeb < Roda
   end
 
   hash_branch("after-login") do |r|
-    if (project = current_user.projects_dataset.order(:created_at).first)
+    if (project = current_account.projects_dataset.order(:created_at).first)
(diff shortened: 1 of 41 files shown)


enescakir (approved):
Thanks for doing this. IMHO, we can rename other `user` instances as `account` at some point

PR #239, opened 2023-07-03
Make all migrations reversible including "ph" ones

enescakir:
We have two different migration paths. Default one has reverting logic but "ph" migrations don't have that. Default migrations can't down to version zero without reverting "ph" migrations because some tables depend each other.

If we are reverting default migrations to version zero, it reverts "ph" migrations to zero first.

"009_ssh_generation" migration has "set_column_allow_null" and it's not auto reversible in change block. I divided it to up/down blocks.

Also, I added "rake test_down" to the end of CI pipeline, so we can be sure new migrations are reversible too.

PR #2144, opened 2024-10-29
Update rubocop and apply fixes

enescakir:
The new version fails with the following exception

    Error: The `Sequel/ColumnDefault` cop has been removed since ColumnDefault cop since it's mostly outdated advice. Details [#33](https://github.com/rubocop/rubocop-sequel/issues/32).
    (obsolete configuration found in .rubocop.yml, please update it)
    RuboCop failed!
    Running RuboCop...

Check the author of the linked issue [^1] (spoiler alert: @fdr)

Also I disabled new `Sequel/IrreversibleMigration` cop since we don't try to make our migrations reversible [^2].

[^1]: https://github.com/rubocop/rubocop-sequel/issues/32
[^2]: https://github.com/ubicloud/ubicloud/pull/239


jeremyevans (approved):
Looks good.  If you want my personal opinion, we should do:

```
AllCops:
  DisabledByDefault: true
```

and then only add the cops that we think are actually useful.

fdr:
Oof, I dunno about that one, I use standardrb to basically get rid of most of the (I think) insane rubocop default cops, so the number of times we're messing with cops is once every few months.

enescakir:
> Looks good. If you want my personal opinion, we should do:
> 
> ```
> AllCops:
>   DisabledByDefault: true
> ```
> 
> and then only add the cops that we think are actually useful.

How can we learn about new cops that we might really like if we disable by default and only enable some? Should we check rubocop and its plugins regularly?

jeremyevans:
> > Looks good. If you want my personal opinion, we should do:
> > ```
> > AllCops:
> >   DisabledByDefault: true
> > ```
> > and then only add the cops that we think are actually useful.
> 
> How can we learn about new cops that we might really like if we disable by default and only enable some? Should we check rubocop and its plugins regularly?

IMO, we don't need to learn about new cops, and we don't need to check Rubocop and its plugins regularly.  My recommendation would be that if we use Rubocop at all, we should only use it to prevent style that actually causes significant issues.  However, I place a very low value on stylistic consistency, and I think other programmers generally place a much higher value on it.

> Oof, I dunno about that one, I use standardrb to basically get rid of most of the (I think) insane rubocop default cops, so the number of times we're messing with cops is once every few months. the idea of adding new cops by construction seems like more accountancy than that. But probably essential to this is standardrb's cop settings.

I see `standard` in the Gemfile, but I don't see it actually used.  If we want to standardize on `standard`, shouldn't we remove `rubocop*` from the Gemfile and just rely on `standard` (`standard` still depends on `rubocop`)?  Alternatively, since we are not using `standard`, should we remove it from the Gemfile?

fdr:
> > I see `standard` in the Gemfile, but I don't see it actually used.
> 
> 
> 
> Turns out it is referenced in `.rubocop.yml`, which `standard` states is explicitly unsupported: https://github.com/standardrb/standard?tab=readme-ov-file#running-standards-rules-via-rubocop

Yeah but it works :)

fdr:
My opinion: let's keep the cops, or at least most of them. Though, it's quite possible that the Sequel cop (or any other cop suite) may cause more trouble than it is worth. Though, as of yet, it's not a _lot_ of trouble.

The reason is this: I probably would agree to go all anarchic on Rubocop (or have never adopted it) if my main interaction with it was largely complaining about my code, but with `rubocop -A`, it mostly fixes or makes consistent a bunch of stuff without me doing much of anything. Note -A is the "unsafe" one, and I have seen it generate bad code occasionally (mostly because statically analyzing code in complete rigor is impossible). So don't run it on large dirty trees, always do so in anything remotely sensitive after committing to do a quick inspection, otherwise you might be stuck picking through the wreckage to find a tiny behavioral change.

I use rubocop a bit more like "go fmt," whose stated philosophy of inflexibility has some obvious relation to standardrb (though I subvert it by using in the unsupported way). I've always thought `go fmt`'s concerns about bikeshedding on style were overblown in my experience (but I haven't worked on large C++ code bases), the main reason to be inflexible was to limit implementation complexity of `go fmt` itself, and beyond that, the main advantage was acceptance of wholesale code *rewriting*, not of a single style per se.

Secondarily, I do think this is a single-versus-some programmer thing, particularly in curtailing spurious diffs as people reformat code. A mostly-single programmer who "places a very low value on consistency" still has a high level of internal consistency. To some extent these problems are hidden because the current staff has worked in our rather strange little cult of control plane ruby, but I think that will break down with new inductees outside the cult, and it's just a lot faster and more efficient to let tools rewrite stuff.

fdr:
See also: the much more obscure erb-formatter project I went to the trouble of forking even, to change its rewrite behavior. I'm in general more interested in formatting than linting, though I'll take the latter as a snack or a side dish.

PR #2154, opened 2024-10-30
Run all specs in frozen mode

jeremyevans:
This introduces ThawedMock, which allows the specs to work in frozen
mode. With ThawedMock, you register all methods on the frozen objects
you plan to mock, and it redefines the methods to call methods on
an unfrozen mocked object.  That mocked object then calls the super
method of the method on the original object.  When you use:

```ruby
  expect(obj).to receive(:method)
```

ThawedMock replaces obj with the unfrozen mocked object, so it redefines
the method on the mocked object without raising an error.

If you want to mock additional methods on classes, you need to update
spec/thawed_mock.rb to add a line such as:

```ruby
  allow_mocking(Klass, :method1, :method2, ...)
```

If there is already a line for the class, you just add the method as
an additional argument on the line for the class.

Not only are all autoloaded classes now frozen, but I found numerous
additional classes and modules that are defined internally, which I
froze as well.

Related changes:

* Fix core freezing in tests, which I broke in
  66d5d629878590fff3c4aca9eaceff6f1d23483d.

* Remove all usage of class variables.  Switch to class instance
  variables (or a constant in one case).

* Remove use of expect_any_instance_of and stub_const in the specs,
  with changes to classes to move implementations to singleton
  methods that can be overridden with
  `expect(obj).to receive(:method)`.

* Make Project.feature_flag accept into keyword argument, so it can
  be tested when frozen.

* Remove all skip_if_frozen{,_models} calls in the specs, since they
  are no longer necessary.

* Remove the `frozen_{core,db_model}_{,s,p}spec` rake tasks, since
  `frozen_{,s,p}spec` now runs the full test suite.


spec/model/postgres/postgres_timeline_spec.rb
```diff
@@ -81,13 +81,13 @@
...
+      backup = Struct.new(:key, :last_modified)
       most_recent_backup_time = Time.now
       expect(postgres_timeline).to receive(:backups).and_return(
         [
-          instance_double(Backup, key: "basebackups_005/0001_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 200),
-          instance_double(Backup, key: "basebackups_005/0002_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 100),
-          instance_double(Backup, key: "basebackups_005/0003_backup_stop_sentinel.json", last_modified: most_recent_backup_time)
+          instance_double(backup, key: "basebackups_005/0001_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 200),
+          instance_double(backup, key: "basebackups_005/0002_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 100),
+          instance_double(backup, key: "basebackups_005/0003_backup_stop_sentinel.json", last_modified: most_recent_backup_time)
```
fdr, line 89: hm, if we're using a ginned up struct here, is verifying doubles (instance_double) really doing anything, or should we use regular "double"? Maybe I need to read more, but I'm a bit perplexed why the type value needs stubbing at all.

fdr (commented):
the promised land. You might want to deploy this in phases, e.g. change of class variables, etc, rather than all at once, so take advantage of temporal dispersion between refactorings (i.e. changes, that are not supposed to really change behavior)

I will re-review the complex parts later. But if you want to start the deploy-supervision pipeline, maybe chip out as many of those patches that gently touch production code as you want into a separate PR, which I can speedily approve.

spec/model/postgres/postgres_timeline_spec.rb
jeremyevans: I'm not sure, as I don't have much experience with rspec doubles.  I know the constant isn't needed.  I would guess that you could just return `backup` objects instead of `instance_double` or regular `double`, but I haven't tried.

jeremyevans:
I'll start splitting this into separate PRs for a phased deployment tomorrow.  Should be easy as the commits are already separated.

spec/model/postgres/postgres_timeline_spec.rb
fdr: instance_double checks the base type has those methods, a bit like how this affects expect...returns:


```ruby
    # Prevents you from mocking or stubbing a method that does not exist on
    # a real object. This is generally recommended, and will default to
    # `true` in RSpec 4.
    mocks.verify_partial_doubles = true
```

But when the type is being ginned up, it mostly defeats the point. I guess it prevents a typo in regular `double`. Incidentally....the verbosity of the two names would be better swapped, but, path dependency.

fdr: Okay, the reason for this is the backup is intended to the result object from an S3 query. The normal way to go about this is instance_double that kind of object (even though it would pull in the library at least a little). Unless it were too big and slow and we wanted to say "nah we're okay not checking the real implemention's methods"

fdr: Here's a patch that normalifies these things:

```diff
diff --git a/spec/model/postgres/postgres_timeline_spec.rb b/spec/model/postgres/postgres_timeline_spec.rb
index bd310f9b..d0174276 100644
--- a/spec/model/postgres/postgres_timeline_spec.rb
+++ b/spec/model/postgres/postgres_timeline_spec.rb
@@ -80,13 +80,12 @@
 
   describe "#latest_backup_label_before_target" do
     it "returns most recent backup before given target" do
-      backup = Struct.new(:key, :last_modified)
       most_recent_backup_time = Time.now
       expect(postgres_timeline).to receive(:backups).and_return(
         [
-          instance_double(backup, key: "basebackups_005/0001_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 200),
-          instance_double(backup, key: "basebackups_005/0002_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 100),
-          instance_double(backup, key: "basebackups_005/0003_backup_stop_sentinel.json", last_modified: most_recent_backup_time)
+          instance_double(Minio::Client::Blob, key: "basebackups_005/0001_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 200),
+          instance_double(Minio::Client::Blob, key: "basebackups_005/0002_backup_stop_sentinel.json", last_modified: most_recent_backup_time - 100),
+          instance_double(Minio::Client::Blob, key: "basebackups_005/0003_backup_stop_sentinel.json", last_modified: most_recent_backup_time)
         ]
       )
 
@@ -122,11 +121,10 @@
   end
 
   it "returns list of backups" do
-    backup = Struct.new(:key)
     expect(postgres_timeline).to receive(:blob_storage).and_return(instance_double(MinioCluster, url: "https://blob-endpoint", root_certs: "certs")).at_least(:once)
 
     minio_client = Minio::Client.new(endpoint: "https://blob-endpoint", access_key: "access_key", secret_key: "secret_key", ssl_ca_file_data: "data")
-    expect(minio_client).to receive(:list_objects).with(postgres_timeline.ubid, "basebackups_005/").and_return([instance_double(backup, key: "backup_stop_sentinel.json"), instance_double(backup, key: "unrelated_file.txt")])
+    expect(minio_client).to receive(:list_objects).with(postgres_timeline.ubid, "basebackups_005/").and_return([instance_double(Minio::Client::Blob, key: "backup_stop_sentinel.json"), instance_double(Minio::Client::Blob, key: "unrelated_file.txt")])
     expect(Minio::Client).to receive(:new).and_return(minio_client)
 
     expect(postgres_timeline.backups.map(&:key)).to eq(["backup_stop_sentinel.json"])
```

spec/thawed_mock.rb
```diff
@@ -0,0 +1,112 @@
...
+    [:allow, :expect].each do |method|
+      define_method(method) do |obj = nil, &block|
+        return super(&block) if obj.nil? && block
+
+        _, mock = OBJECTS[obj]
+        super(mock || obj, &block)
+      end
+    end
+  end
+  RSpec::Core::ExampleGroup.prepend(ExpectOverride)
```
fdr, line 39: Nice, I was thinking of messing with expect to do this, but honestly my Ruby language lawyering is not good enough, I use it like a C programmer who was taught lambda calculus in scheme that one time.

PR #1278, opened 2024-02-21
Backport roda view coverage testing

fdr:
From https://github.com/jeremyevans/roda-sequel-stack/commit/195dd40df7f440f56e3863aab4819078f92cdd38

PR #2157, opened 2024-10-30
Show total cache quota on the UI

enescakir:
### **Remove leftover text on the UI when no cache entries are available**

  When the project doesn't have any cache entries, the UI was showing a
  "used" text. I fixed this by hiding the header when there are no cache
  entries available.
  
| Before | After |
|--------|-------|
| Screenshot 2024-10-30 at 05 48 25 | Screenshot 2024-10-30 at 05 49 59 |

### **Show total cache quota on the UI**

  Currently, we display the total cache usage on the UI, but not the total
  cache quota. This leaves users uncertain about the remaining cache
  storage. Therefore, I've added the effective cache quota to the UI.

Screenshot 2024-10-30 at 05 56 29


fdr (approved):
conditionals on instance vars without a test makes me kinda jumpy typo-to-nil is scary but we haven't quite done #1278

velioglu (approved):

PR #2158, opened 2024-10-30
Run all specs in frozen mode - Phase 1

jeremyevans:
We are going to merge #2154 in phases, and this is the first phase.  This provides some setup:

* Remove all usage of class variables. Switch to class instance variables (or a constant in one case).  Class variables are a Ruby misfeature that should almost never be used.

* When switching away from class variables to class instance variables, add `freeze` methods to the classes that call the methods to set the instance variables, so calling the methods will not raise FrozenError.

* Fix core freezing in tests, which I broke in 66d5d62.

* Remove use of `expect_any_instance_of` in a spec (which I added last week in 0d3702f1), by adding a class method that can be mocked.

* Change Prog::Postgres::PostgresResourceNexus.dns_zone so it caches the value on first call, even if the value is nil.

### DIFF
--- a/lib/billing_rate.rb
+++ b/lib/billing_rate.rb
@@ -3,8 +3,15 @@
 require "yaml"
 
 class BillingRate
+  # :nocov:
+  def self.freeze
+    rates
+    super
+  end
+  # :nocov:
+
   def self.rates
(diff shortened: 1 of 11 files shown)


fdr (approved):
Let's get it in and supervise

PR #2160, opened 2024-10-30
Fix 2 nondeterministic tests by hardcoding expected value

jeremyevans:
Without this change, I see failures such as:

```
Failure/Error: expect(page.title).to eq("Ubicloud - #{Account[email: TEST_USER_EMAIL].projects.first.name} Dashboard")

expected: "Ubicloud - Default Dashboard"
     got: "Ubicloud - Invited project Dashboard"
```

This is because there are multiple projects for the account, and a dataset order was not enforced.  We could enforce a dataset order, but since the value we are expecting is constant, it seems simpler to hardcode it.

### DIFF
--- a/clover_web.rb
+++ b/clover_web.rb
@@ -366,7 +366,7 @@ class CloverWeb < Roda
   end
 
   hash_branch("after-login") do |r|
-    if (project = current_account.projects_dataset.order(:created_at).first)
+    if (project = current_account.projects_dataset.order(:created_at, :name).first)
       r.redirect "#{project.path}/dashboard"
     else
       r.redirect "/project"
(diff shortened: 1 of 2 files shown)


jeremyevans:
Well, this fails in CI, but passes locally.  Maybe the app behavior is non-deterministic, and not just the test?  We could expect either, or we could change the app behavior to be deterministic.  I'll look into it.

jeremyevans:
It turns out the application behavior is non-deterministic in the test environment since we switched to transactional testing, because it uses `order(:created_at)`, which is the same for all projects in the tests, as they are all inserted in the same transaction.  Switching to `order(:created_at, :name)` will allow for consistent behavior, so I'll add a commit for that.

fdr (approved):

enescakir (approved):

PR #2164, opened 2024-10-30
Support allocating multiple GPUs to a single VM

bsatzger:
**Create 'latitude-ai' location**
Create a new location `latitude-ai` for resources hosted on Latitude dedicated to AI.

**Migration to add gpu_count to inference_endpoint**
Add a column `gpu_count` to the `inference_endpoint` table. That column allows to specify how many gpus are to be assigned to each VM that runs a replica of this endpoint.

**Assemble VMs with specified number of GPUs**
Add a new parameter `gpu_count` that allows to specify the number of gpus that should be assigned to a vm. By default, that number is zero.

Allocator is extended to allocate gpus to vms based on requested gpu
count.

**Add gpu_count to inference endpoints**
Explicitly specify the number of gpus an inference endpoint requires to run a model.

**Add Llama 70B and 405B models**
Add `llama-3-1-nt-70b-it` and `llama-3-1-405b-it` to the list of available models. Each come in a flavor with and a flavor without tool support.


prog/vm/nexus.rb
```diff
@@ -14,7 +14,7 @@ def self.assemble(public_key, project_id, name: nil, size: "standard-2",
     unix_user: "ubi", location: "hetzner-hel1", boot_image: Config.default_boot_image_name,
     private_subnet_id: nil, nic_id: nil, storage_volumes: nil, boot_disk_index: 0,
     enable_ip4: false, pool_id: nil, arch: "x64", allow_only_ssh: false, swap_size_bytes: nil,
-    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [])
+    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [], gpu_count: 0)
```
fdr, line 17: would it be feasible to "learn" the gpu count, like cores and network?

fdr (approved):
basically fine, though I don't like the accountancy of having to etch the gpu count with a server by hand. This is a degradation from how a number of server dimensions are handled, i.e. they are measured and relieve the operator of cross reference (an override can be handy nevertheless, though)

prog/vm/nexus.rb
fdr: also, maybe you kinda did something about this implicitly?

```
commit 816d182735463338cf443c527cf456c6df3fd759
Author: Benjamin Satzger <ben@ubicloud.com>
Date:   Tue May 7 15:04:26 2024 +0200

    Learn about PCI devices of a host
    
    This commit adds the ability to HostNexus to learn about PCI devices.
    *For now, we only consider NVIDIA devices for which IOMMU is enabled.*
```

bsatzger: yes, we already do learn about gpus on hosts. this is for the vm, how many gpus we request for it, similar to how we request cores, memory, ipv4 addresses etc.

scheduling/allocator.rb
```diff
@@ -130,7 +130,7 @@ def self.update_vm(vm_host, vm)
         allocated_at: Time.now
       )
       AssignedVmAddress.create_with_id(dst_vm_id: vm.id, ip: ip4.to_s, address_id: address.id) if ip4
-      vm.sshable&.update(host: vm.ephemeral_net4 || vm.ephemeral_net6.nth(2))
+      vm.sshable&.update(host: vm.ephemeral_net4 || NetAddr.parse_net(vm.ephemeral_net6).nth(2))
```
furkansahin, line 133: The problem is... heroku doesn't support IPv6. So, this sshable wouldn't work in prod

furkansahin (approved):

bsatzger:
> basically fine, though I don't like the accountancy of having to etch the gpu count with a server by hand. This is a degradation from how a number of server dimensions are handled, i.e. they are measured and relieve the operator of cross reference (an override can be handy nevertheless, though)

we do have code to learn about gpus during host setup. gpu count is only specified for the vm. before this pr, you could say that a vm needs a gpu. now you can quantify that, i.e. say it needs N gpus.

prog/vm/nexus.rb
fdr: Of course, I'm not sure what I was thinking.

PR #2167, opened 2024-10-31
Add the amount of money saved compare to default runners to invoice emails

enescakir:
Our managed GitHub Actions runners are 10x cost-effective than GitHub-hosted runners. We decided to share the amount of money saved by just using our runners with customers in the invoice emails.

Screenshot 2024-10-31 at 17 57 27

### DIFF
--- a/model/invoice.rb
+++ b/model/invoice.rb
@@ -88,15 +88,21 @@ class Invoice < Sequel::Model
 
   def send_success_email(below_threshold: false)
     ser = Serializers::Invoice.serialize(self, {detailed: true})
-    message = if below_threshold
-      "Since the invoice total of #{ser[:total]} is below our minimum charge threshold, there will be no charges for this month."
+    messages = if below_threshold
+      ["Since the invoice total of #{ser[:total]} is below our minimum charge threshold, there will be no charges for this month."]
     else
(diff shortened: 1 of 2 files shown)


model/invoice.rb
```diff
@@ -88,15 +88,21 @@ def charge
...
+    saved_amount = 9 * github_usage
+    if saved_amount > 1
+      messages << "You saved $#{saved_amount.to_i} this month using managed Ubicloud runners instead of GitHub hosted runners!"
+    end
+
     Util.send_email(ser[:billing_email], "Ubicloud #{ser[:name]} Invoice ##{ser[:invoice_number]}",
       greeting: "Dear #{ser[:billing_name]},",
       body: ["Please find your current invoice ##{ser[:invoice_number]} below.",
-        message,
+        *messages,
```
fdr, line 105: I had no idea you could use a splat in an array literal like that.

fdr (approved):

velioglu (approved):

model/invoice.rb
enescakir: Yes, it's a cool one

```ruby
[16] clover-development(main)> [1,*[2,3,4],5,6]
=> [1, 2, 3, 4, 5, 6]
```

PR #2168, opened 2024-10-31
Merge CloverWeb, CloverApi, and CloverRuntime into Clover

jeremyevans:
This is a prerequisite for merging the api and web routing trees.

Much of this is just moving the bodies of clover_web.rb, clover_api.rb, and clover_runtime.rb into clover.rb.  However, to preserve differences in behavior, such as always using json responses for api and runtime, and only using sessions, content-security-policy, and HSTS for web, this adds the api?, runtime?, and web? methods, and overrides methods that check those methods to allow for conditional behavior. It also adds a response.json= method, which the api and runtime routes use to signal that responses should be json.

For api and runtime routes, this updates the hash_branch calls to use api and runtime specific namespaces.  The namespaces for the web routes are not modified.

This removes the freezing of Clover from spec/routes/spec_helper.rb. Now that Clover is the main app and autoloads hash branches, you can only freeze if after all hash branches has been loaded. So running an individual spec no longer freezes Clover.  However, the frozen specs will still test that everything works when Clover is frozen (those load all hash branches before freezing Clover).

This changes autoload_routes to load all routes up front when FORCE_AUTOLOAD=1.  This makes the test environment more like the production environment.  Running a single spec will still autoload only the branches necessary, to keep single spec runs fast.

I'm submitting this as a draft for now in case someone wants to provide early feedback.  Before removing the draft flag, I'd like to get one of the routing branches merged, to see what that would look like.


fdr:
@byucesoy added you as a reviewer since convergence work is something you have a lot of experience with, though it's one level lower than this (in the guts of the application behavior rather than how routes are organized)

and I'll add @enescakir since he's about as Roda-adept as anyone (and I think he organized the three apps at some point, coming from a single app in the beginning)

jeremyevans:
I've rebased this on main, and completed the merging of the project/location/vm and project/vm branches, and the elimination of Routes::Common::VmHelper.  I think the resulting routing tree is significantly simpler to understand, and it mostly avoids indirection.  The list and post handling still require indirection, but this changes them to handle them via a Clover method, so they run in the same context as the routing tree, and you don't have to context switch between routing tree context and helper context.

I am stopping here until I receive feedback regarding this approach.

routes/api/project/firewall.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-class CloverApi
-  hash_branch(:project_prefix, "firewall") do |r|
+class Clover
+  hash_branch(:api_project_prefix, "firewall") do |r|
```
fdr, line 4: This  `api_` practice is perhaps the weirdest part of the patch for me. Is this a durable practice or transitory? I kinda understand what you were getting at in general, to streamline the AppMode conditional branches.

jeremyevans: This is transitory until the routing trees are fully merged.  Once the routing trees are fully merged, there will be no need for separate prefixes.

fdr: can we see a spike of how this might work, perhaps choosing from one to three (but maybe staring with just one) of the more complicated convergences as a model? No need to make the commit history neat or do the tests, per prototyping. This way we can glimpse the future and put phasing in context.

jeremyevans: In the main route block, we currently have `r.hash_branches("api")` for the api routes, and `r.hash_branches("")` for the web routes.  This will combine and both will use `r.hash_branches("")` after all branches have finished merging.

In terms of how individual route files will work when merged, this PR has the project/location/vm and project/vm route blocks already merged.  The only change when this is finalized will be that there will single `hash_branch` call per file, instead of two calls (one for api and one for web).

routes/merged/project/vm.rb
```diff
@@ -2,14 +2,20 @@
...
-    vm_endpoint_helper = Routes::Common::VmHelper.new(app: self, request: r, user: current_account, location: nil, resource: nil)
-
     r.get true do
-      vm_endpoint_helper.list
+      dataset = vm_list_dataset
+
+      if api?
+        vm_list_api_response(dataset)
+      else
+        @vms = Serializers::Vm.serialize(dataset.eager(:semaphores, :assigned_vm_address, :vm_storage_volumes).reverse(:created_at).all, {include_path: true})
```
fdr, line 11: Huh, why is the eager load of semaphores interesting? Incidentally, loading of correlated objects (eager or not) is where we might get into a bit of trouble, e.g. if vm_storage_volumes has its own "view" capability or lack thereof. 

This is kind of the crux of the sensitivity we have to authorization bypasses for something as ABAC with a policy language. Often, dealing with correlated objects is implicated.

The AWS and other consoles tend to break (as in, spew errors sometimes) because they can't load certain data given the permutations of what their policy languages allow for, should one make a minimal set of capabilities is made available to a principal.

If I had to choose one, I'd still choose a flexible policy language and a weird looking or slightly busted console, hence, Ubicloud is rather similar in that regard. But, making you aware that this is where the hairy stuff comes from.

jeremyevans:
Rebased on main to fix merge conflicts.

routes/merged/project/vm.rb
enescakir: > Huh, why is the eager load of semaphores interesting

We use semaphores in the `display_state` https://github.com/ubicloud/ubicloud/blob/027c2c7d17d0ad266622d05bd612f9022b6defc1/model/vm.rb#L61

routes/merged/project/location/vm.rb
```diff
@@ -0,0 +1,61 @@
...
+        response.status = 204
+        nil
+      end
+    end
+
+    # 204 response for invalid names
+    r.is String do |vm_name|
+      r.post do
+        vm_post(vm_name)
+      end
```
enescakir, line 50: General Ruby styling question: What's your opinion on using one-line blocks for single-line codes? Are they popular in Ruby?

```suggestion
      r.post { vm_post(vm_name) }
```

clover.rb
```diff
@@ -3,38 +3,688 @@
...
+  end
+
+  def csrf_tag(*)
+    render("components/form/hidden", locals: {name: csrf_field, value: csrf_token(*)})
+  end
+
+  def redirect_back_with_inputs
+    flash["old"] = request.params
+    request.redirect env["HTTP_REFERER"] || "/"
+  end
```
enescakir, line 184: Helpers like `csrf_tag`, `redirect_back_with_inputs` specifically for web, and `get_jwt_payload` specifically for runtime. We could consider moving these helpers to the `routes/helpers` directory for better isolation.

clover.rb
```diff
@@ -3,38 +3,688 @@
...
+      .group_by { [_1["resource_type"], _1["resource_family"], _1["location"]] }
+      .map { |_, brs| brs.max_by { _1["active_from"] } }
+      .each_with_object(Hash.new { |h, k| h[k] = h.class.new(&h.default_proc) }) do |br, hash|
+      hash[br["location"]][br["resource_type"]][br["resource_family"]] = {
+        hourly: br["unit_price"].to_f * 60,
+        monthly: br["unit_price"].to_f * 60 * 672
+      }
+    end
+  end
+  plugin :not_found do
```
enescakir, line 246: General question for Roda: Can the not found exception be caught in the 'error_handler' when we aren't using the 'not_found' plugin? It might be more clear to handle all error cases in one place

clover.rb
```diff
@@ -3,38 +3,688 @@
...
+  autoload_routes("api", "api")
+  autoload_routes("runtime", "runtime")
+
+  route do |r|
+    if api?
+      response.json = true
+      # To make test and development easier
+      # :nocov:
+      unless Config.production?
+        r.on("api") do
```
enescakir, line 646: @fdr https://github.com/ubicloud/ubicloud/pull/2084 will remove some branches from this PR

clover.rb
```diff
@@ -3,38 +3,688 @@
...
+      rodauth.check_active_session
+      rodauth.require_authentication
+      r.hash_branches("api")
+    elsif runtime?
+      response.json = true
+      if (jwt_payload = get_jwt_payload(r)).nil? || (@vm = Vm.from_ubid(jwt_payload["sub"])).nil?
+        fail CloverError.new(400, "InvalidRequest", "invalid JWT format or claim in Authorization header")
+      end
+
+      r.on "runtime" do
```
enescakir, line 665: Since we're running the `runtime` check twice, is there a more elegant way to do it just once?

routes/merged/project/location/vm.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      r.get true do
+        Authorization.authorize(current_account.id, "Vm:view", vm.id)
+        @vm = Serializers::Vm.serialize(vm, {detailed: true})
+        if api?
+          @vm
+        else
+          view "vm/show"
+        end
```
enescakir, line 35: General Ruby styling question: What's your opinion on using ternary for single-line codes? Are they popular in Ruby?

```suggestion
        api? ? @vm : view "vm/show"
```

enescakir (commented):
I generally like the PR ❤️  I put a few questions specific to this PR, along with some general ones. My only concern is that it makes the `clover.rb` file too large (672 lines). Does this make it harder to modify certain sections or maintain it? Perhaps we should consider breaking it down a bit more.

clover.rb
```diff
@@ -3,38 +3,588 @@
...
+      def assets
+        exception_page_assets
+        super
+      end
+    end
+    # :nocov:
+  end
+
+  plugin :error_handler do |e|
+    # raise e
```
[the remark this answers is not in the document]
jeremyevans: Not sure if I should have left this in, but I find the current approach, where the application swallows all errors in testing, makes debugging much more difficult.  The roda-sequel-stack default is to override the error_handler in the specs to raise the original error, which makes debugging much easier.  However, we have a lot of specs that rely on using the production error handler, so we would need changes to use the roda-sequel-stack default.

Having this commented out makes it easy to uncomment when debugging.  Making it raise conditionally would be another approach.

clover.rb
```diff
@@ -3,38 +3,588 @@
...
+    error = parse_error(e)
+
+    if api? || runtime?
+      unless runtime? && error[:code] == 204
```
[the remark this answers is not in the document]
jeremyevans: That makes sense to me.  I only did it this way to match the previous behavior.    I'll add a commit to use empty responses for 204 requests.

clover.rb
jeremyevans: Yes.  We ran remove the `elsif`, and have `r.on "runtime"` after the `else`, and do this work inside that block.  I'll push a commit for that.

clover.rb
jeremyevans: The `error_handler` is only called for exceptions.  `not_found` handles cases where the response uses status code 404 and has not written anything to the body.  So you cannot use `error_handler` for `not_found` responses.

clover.rb
jeremyevans: OK, I'll push a commit to do that.  I'll move the custom methods to helpers, so that clover.rb just has the plugin configurations and the routes.

routes/merged/project/location/vm.rb
jeremyevans: I'm fine with using ternary operators.

routes/merged/project/location/vm.rb
jeremyevans: I think using single line blocks is fine for routes.

clover.rb
```diff
@@ -3,38 +3,588 @@
...
+      end
+    else
+      @error = error
+
+      case e
+      when Sequel::ValidationFailed
+        flash["error"] = @error[:message]
+        return redirect_back_with_inputs
+      when Validation::ValidationFailed
+        flash["errors"] = (flash["errors"] || {}).merge(@error[:details])
```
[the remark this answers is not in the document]
enescakir: Actually we do that. `redirect_back_with_inputs` puts old values to flash. Our form components check old value from flash

jeremyevans:
> I generally like the PR ❤️ I put a few questions specific to this PR, along with some general ones. My only concern is that it makes the `clover.rb` file too large (672 lines). Does this make it harder to modify certain sections or maintain it? Perhaps we should consider breaking it down a bit more.

Thank you for the detailed review.  I think I've made all of the changes you requested. Please let me know if I've missed anything.

clover.rb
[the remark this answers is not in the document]
jeremyevans: Ah.  The issue with that is that large form inputs will break things, since flash is stored in the session cookie, so a large form input would exceed the 4K cookie limit and and break the session cookie.

byucesoy (commented):
I didn't do line by line review. I think @enescakir already did that. Instead I looked each commit for general idea. I think this is a better model than having `Common` modules.

Not directly related with this PR, but what is your opinion on merging `project/location/{resource}` and `project/{resource}` files as well (just the content, the endpoints would still have `location` in them where necessary)? Keeping them in same file would simplify the development. There would be only one place to look for `{resource}` related endpoint logic.

jeremyevans:
> Not directly related with this PR, but what is your opinion on merging `project/location/{resource}` and `project/{resource}` files as well (just the content, the endpoints would still have `location` in them where necessary)? Keeping them in same file would simplify the development. There would be only one place to look for `{resource}` related endpoint logic.

Due to how Roda works, this mean mean having two separate routing branches in the same file.  That is possible.  However, in order for the autoloading to work, you would need a stub file, or extra code in `autoload_routes`.

For example, let's take VMs.  This currently has both `routes/merged/project/vm.rb` and `routes/merged/project/location/vm.rb`.  We could combine those in the same file.  Say we picked `routes/merged/project/vm.rb` to hold both routing branches.  We would still need a `routes/merged/project/location/vm.rb` that required `routes/merged/project/vm.rb` to support the autoloading of hash branches.  Otherwise, a request for a route in the location branch could fail if the non-location branch hadn't been loaded yet.

We could instead modify `autoload_routes` so that it would autoload the same file for both routing branches, but that would make that method more complicated.

My recommendation would be to keep the routing branches in separate files.

enescakir (approved):

clover.rb
jeremyevans: After I merge this, I'll see if I can update #2084 for this and other changes.

PR #2171, opened 2024-11-01
Display the list of fields that failed validation at the top

enescakir:
When a field in a form fails validation, we display the error at the bottom of the relevant field. However, this requires the customer to scroll down, which can be inconvenient for long pages. To improve this, I've added a list of failed fields to the top of the page in the error view. This way, customers can quickly see the names of the fields that failed at a glance.

Screenshot 2024-10-31 at 18 21 37


jeremyevans (approved):
This is definitely an improvement.

PR #2172, opened 2024-11-01
Add Cloudflare Turnstile captcha to account creation form

enescakir:
Our account creation form is publicly accessible, and open to abuse by spammers. The spam bots enter random email addresses and increase our bounce rate.

We decided to add a captcha to the form to prevent this abuse. I evaluated a few options and decided to use Cloudflare's Turnstile since it's free and privacy-friendly. It's also easy to implement. [^1]

We add their client-side widget to our form. This widget add token input and we use this token to verify the captcha on the server-side.

I added `cloudflare_turnstile.erb` component to add Cloudflare Turnstile to any form easily. External Cloudflare script is loaded only when the component is used.

There are 3 different modes for Turnstile: managed, non-interactive, and invisible [^2]. It's configurable in the Cloudflare UI. I think we can start with the invisible mode and see how it goes.

[^1]: https://developers.cloudflare.com/turnstile/get-started/
[^2]: https://developers.cloudflare.com/turnstile/concepts/widget/

### DIFF
--- a/lib/validation.rb
+++ b/lib/validation.rb
@@ -2,6 +2,7 @@
 
 require "time"
 require "netaddr"
+require "excon"
 
 module Validation
   class ValidationFailed < CloverError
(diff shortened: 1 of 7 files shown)


lib/validation.rb
```diff
@@ -238,4 +239,18 @@ def self.validate_core_quota(project, resource_type, requested_core_count)
       fail ValidationFailed.new({size: "Insufficient quota for requested size. Requested core count: #{requested_core_count}, currently used core count: #{current_used_core_count}, maximum allowed core count: #{effective_quota_value}, remaining core count: #{effective_quota_value - current_used_core_count}"})
     end
   end
+
+  def self.validate_cloudflare_turnstile(cf_response)
+    return unless !!Config.cloudflare_turnstile_site_key
```
jeremyevans, line 244: No need for `!!`, since you are not storing the return value.

views/components/form/cloudflare_turnstile.erb
```diff
@@ -0,0 +1,5 @@
+<% @enable_cloudflare_turnstile = !!Config.cloudflare_turnstile_site_key %>
```
jeremyevans, line 1: No need for `!!`.  While you are storing the value, you are only using it in boolean context in this template and the layout.

jeremyevans (approved):
Looks good.  I think the places you are using `!!` can be removed, but that's up to you.

PR #2173, opened 2024-11-01
Log web exceptions with Clog

enescakir:
When we encounter unexpected errors in our web application, we simply print the stack trace to stderr. This makes it challenging to investigate the error cause, as our log service can only parse JSON logs. Thus, we can log the exceptions with Clog, similar to how we handle exceptions in the respirate.

### DIFF
--- a/routes/clover_base.rb
+++ b/routes/clover_base.rb
@@ -37,8 +37,7 @@ module CloverBase
       message = e.message
       details = e.details
     else
-      $stderr.print "#{e.class}: #{e.message}\n"
-      warn e.backtrace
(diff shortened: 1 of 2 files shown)


fdr (approved):
overdue

jeremyevans (approved):
Clog.emit appears to be how we log everything, so this makes sense to me.  It has the added benefit of not flooding stderr if you make a change that breaks lots of things and then run the specs.

fdr:
@jeremyevans you can see the entire Clog story here. f2b128f972ce81847956e386d626bd9a94c0106e

PR #2179, opened 2024-11-03
Re-enable Rhizome Validation

aamederen:
Previously, we found that validations failed after the validation changes are deployed. In my tests, I realized that the validations failed due to Gemfile.lock changing after the bundle install step.

I decided to filter out the Gemfile.lock file. Alternatively, we could move the validation step before the gem install step, but making sure that `bundle install` runs before any ruby script is run sounded like a better idea.

### DIFF
--- a/prog/install_rhizome.rb
+++ b/prog/install_rhizome.rb
@@ -7,6 +7,8 @@ require "digest/sha2"
 class Prog::InstallRhizome < Prog::Base
   subject_is :sshable
 
+  SKIP_VALIDATION = ["Gemfile.lock"]
+
   label def start
     tar = StringIO.new
     file_hash_map = {} # pun intended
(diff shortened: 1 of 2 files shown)


prog/install_rhizome.rb
```diff
@@ -7,6 +7,8 @@
 class Prog::InstallRhizome < Prog::Base
   subject_is :sshable
 
+  SKIP_VALIDATION = ["Gemfile.lock"]
```
enescakir, line 10: What's changed in the `Gemfile.lock`? If it's changed, there might be something missing in our lock file

aamederen: We don't have a unified Gemfile for the whole Rhizome. Instead, we have one in rhizome/common, one in rhizome/host. Therefore, the deployed Gemfile.lock in the sshable can be different than what we have in the repo. The Gemfile.lock in the repo is valid for the VmHost, but not the PG VM for example.

byucesoy (approved):

PR #2182, opened 2024-11-04
Migrate adhoc delete buttons to the component

enescakir:
- **Add confirmation message argument to delete button component**
  JS code supports `confirmation-message` property for `.delete-btn`
  class.
  

- **Add method argument to delete button component**
  JS code supports `method` property for `.delete-btn` class.
  

- **Migrate all delete button instance to component**
  We have already `delete_button` component. But some delete button
  instances are not using this component. We can migrate all delete button
  instances to this component.
  

- **Add default url values to delete button component**
  Almost half of the `delete_button` instances use `request.path` for
  `url` or `redirect`. By adding default values to them, we can remove
  bunch of lines.

### DIFF
--- a/views/components/delete_button.erb
+++ b/views/components/delete_button.erb
@@ -1,6 +1,7 @@
 <% csrf_url = (defined?(csrf_url) && csrf_url) ? csrf_url : url %>
 <% text = (defined?(text) && text) ? text : "Delete" %>
 <% confirmation = (defined?(confirmation) && confirmation) ? confirmation : nil %>
+<% confirmation_message = (defined?(confirmation_message) && confirmation_message) ? confirmation_message : nil %>
 
 <%== render(
   "components/button",
@@ -13,6 +14,7 @@
       "data-url" => url,
       "data-csrf" => csrf_token(csrf_url, "DELETE"),
       "data-confirmation" => confirmation,
+      "data-confirmation-message" => confirmation_message,
       "data-redirect" => redirect
     }
   }
--- a/views/components/delete_button.erb
+++ b/views/components/delete_button.erb
@@ -1,7 +1,9 @@
+<% url = (defined?(url) && url) ? url : request.path %>
 <% csrf_url = (defined?(csrf_url) && csrf_url) ? csrf_url : url %>
 <% text = (defined?(text) && text) ? text : "Delete" %>
 <% confirmation = (defined?(confirmation) && confirmation) ? confirmation : nil %>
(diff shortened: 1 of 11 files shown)


fdr:
Seeing the forest for the trees in this kind of patch has been outside my experience, but, adding @jeremyevans as a reviewer... as it may be within his experience.

enescakir:
JS code as reference: https://github.com/ubicloud/ubicloud/blob/411664d71791670c2fd1d31d8341ea78b70ab9ea/assets/js/app.js#L55

fdr (commented):
.

views/components/delete_button.erb
```diff
@@ -1,6 +1,10 @@
+<% url = (defined?(url) && url) ? url : request.path %>
 <% csrf_url = (defined?(csrf_url) && csrf_url) ? csrf_url : url %>
 <% text = (defined?(text) && text) ? text : "Delete" %>
 <% confirmation = (defined?(confirmation) && confirmation) ? confirmation : nil %>
+<% confirmation_message = (defined?(confirmation_message) && confirmation_message) ? confirmation_message : nil %>
+<% redirect = (defined?(redirect) && redirect) ? redirect : request.path %>
+<% method = (defined?(method) && method) ? method : "DELETE" %>
```
jeremyevans, line 7: This code works, but it's more accidental than deliberate. In the code `url = defined?(url)` , `defined?(url)` always returns `'local-variable'`, since its on the RHS of the assignment, and the local variable is on the LHS.  The `defined?` calls could be eliminated.  That leaves you with `url = (url ? url : request.path)`.  The idiomatic way to do that in ruby is using `||=`:

```suggestion
<% 
  url ||= request.path
  csrf_url ||= url
  text ||= "Delete"
  confirmation ||= nil
  confirmation_message ||= nil
  redirect ||= request.path
  method ||= "DELETE"
%>
```

jeremyevans (approved):
I like it.  This DRYs up a lot of template code.

views/components/delete_button.erb
enescakir: I added `defined?(url)` because it was failing when we didn't provide this parameter to the component while rendering. It was sad to find out it worked by chance. Fortunately, it did work 😄 I tested the new method, and it also works. However, we use the old syntax in all components. I'll merge this PR as is, and create another one for refactoring the input variables of all components.

enescakir: https://github.com/ubicloud/ubicloud/pull/2200

furkansahin (approved):
I tested this locally, it works.

PR #2189, opened 2024-11-05
Relax time range when testing timestamp based ubids.

pykello:
In an rspec test for UBIDs we verified that timestamp part of sequentially created timestamps are close to each other. Previously, the allowed time range was < 8ms. This sporadically failed when that test took e.g. 10ms. In this change we relax the time range to < 128ms to reduce failures.

Also, I'm making the tests a little bit more readable.


jeremyevans (approved):
Thank you for fixing this.  I've hit this failure at least twice.

PR #2193, opened 2024-11-05
VM Recreate

furkansahin:
This pr adds a new logic to Vm::Nexus to be able to recreate a VM. For
now, this is only available to the on-call. In case we realize the VM is
stuck in the provisioning path, the on-call may simply incr_recreate and
force the VM to be recreated including new storage entities, ip
configuration, etc. In future, we may automate this process further and
even add features like excluding the previously tried host. For now,
this is a very basic implementation.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -107,7 +107,7 @@ class Prog::Vm::Nexus < Prog::Base
           "exclude_host_ids" => exclude_host_ids,
           "gpu_count" => gpu_count
         }]
-      ) { _1.id = vm.id }
+      ) { _1.id = vm.id }.tap { vm.incr_can_recreate }
     end
   end
 
@@ -157,6 +157,14 @@ class Prog::Vm::Nexus < Prog::Base
         hop_destroy
       end
     end
+
+    when_recreate_set? do
+      unless strand.label == "destroy"
+        when_can_recreate_set? do
+          hop_destroy
+        end
+      end
+    end
   end
 
   label def start
@@ -203,10 +211,6 @@ class Prog::Vm::Nexus < Prog::Base
     end
 
     register_deadline("wait", 10 * 60)
-
-    # We don't need storage_volume info anymore, so delete it before
-    # transitioning to the next state.
-    clear_stack_storage_volumes
 
     hop_create_unix_user
   end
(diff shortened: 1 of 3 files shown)


prog/vm/nexus.rb
```diff
@@ -399,25 +411,61 @@ def write_params_json
...
+
+    give_back_resources
+    vm.assigned_vm_address&.destroy
+    vm.vm_storage_volumes.each(&:destroy)
+    DB.transaction do
+      vm.nics.map do |nic|
+        ps = nic.private_subnet
+        nic.update(vm_id: nil)
+        nic.incr_destroy
+        new_nic = Prog::Vnet::NicNexus.assemble(ps.id, name: "#{vm.name}-nic").subject
```
jeremyevans, line 440: Are there any limits on naming?  If Vm and Nic names have the same limits, there would likely be problems if the Vm name was near the limit.

fdr: Definitely there are limits, though the vm name was chosen to be as short as possible while not engendering excessive collisions. Different systems in Linux have different limits as well (e.g. interface names, user names), so we kind of needed the minimum of all of them, plus some space for derivatives such as this one.

furkansahin: Also, we don't use the nic name directly in the server, rather make use of the first 7 chars of the ubid. So, this name is more for us or the customer when we expose them.

prog/vm/nexus.rb
```diff
@@ -399,25 +411,61 @@ def write_params_json
...
+  label def recreate
+    decr_recreate
+    clean_up_host(force_network_cleanup: true, omit_failure: true)
+    if vm.load_balancer
+      vm.load_balancer.evacuate_vm(vm)
+      vm.load_balancer.remove_vm(vm)
+    end
+
+    give_back_resources
+    vm.assigned_vm_address&.destroy
```
fdr, line 433: funny thing, the predecessor of this program actually would hop to code shared with destroy (in fact, it was destroy) but be spared by "recreate" and then hop back to the provisioning path. Does doing something like that avoid repetition?

furkansahin: Thanks for the suggestion, I restructured the code to use the `destroy` state which reduced the code footprint a low.

prog/vm/nexus.rb
```diff
@@ -162,6 +162,14 @@ def before_run
         hop_destroy
       end
     end
+
+    when_recreate_set? do
+      if strand.label != "recreate"
```
fdr, line 167: @furkansahin I know you like push, and probably not the time to do something about it, but it's in conflict with code like this.

fdr (commented):
that recreate method seems a little smelly, surely it must be duplicating a bunch of stuff as features are added. What's preventing us from looping back to the creation state and not doing quite so much? i.e. why is *re*-create different enough from "create" as it comes to those resources?

furkansahin:
@fdr , I have moved around some of the code, I think this makes more sense now. The problem I have is that some of the entities we generated need to be destroyed and recreated. We depend on plugins to destroy some of them. That's why the struggle.

prog/vm/nexus.rb
```diff
@@ -159,6 +159,14 @@ def before_run
         hop_destroy
       end
     end
+
+    when_recreate_set? do
+      if strand.label != "destroy"
```
jeremyevans, line 164:
```suggestion
      unless strand.label == "destroy"
```

Optional, but using unless with == is more idiomatic in Ruby than if with !=.

jeremyevans (commented):
I reviewed from a Ruby code perspective, and the changes seem fine to me.

furkansahin:
@fdr @jeremyevans , may I have a round of review?

jeremyevans (commented):
I don't see any issues with the Ruby code.

PR #2196, opened 2024-11-05
Merge private subnet routes

jeremyevans:


routes/merged/project/location/private_subnet.rb
```diff
@@ -0,0 +1,97 @@
...
+        Authorization.authorize(current_account.id, "PrivateSubnet:view", ps.id)
+        @ps = Serializers::PrivateSubnet.serialize(ps)
+        if api?
+          @ps
+        else
+          @nics = Serializers::Nic.serialize(ps.nics)
+          @connected_subnets = Serializers::PrivateSubnet.serialize(ps.connected_subnets)
+          connectable_subnets = ps.projects.first.private_subnets.select do |ps1|
+            ps1_id = ps1.id
+            ps1_id != ps.id && !ps.connected_subnets.find { |cs| cs.id == ps1_id }
```
enescakir, line 58: Why don't we call `ps1.id` directly?
```suggestion
            ps1.id != ps.id && !ps.connected_subnets.find { |cs| cs.id == ps1.id }
```

jeremyevans: It's a micro-optimization.  Using a local variable saves 1 method call per `find` iteration.

jeremyevans: Also, note that this is inside another `select` loop, so the number of method calls saved is `select` iterations * `find` iterations.

enescakir (approved):

PR #2199, opened 2024-11-06
Add load balancer stack option

furkansahin:
## Add load balancer stack option
Our load balancers are applied to both IPv4 and IPv6 addresses of the
underlying VMs. This is not very useful for the customer in such a
scenario;
1. Customer creates and apps and uses only IPv4 to server the app.
2. Creates a load balancer and attaches his VMs.
3. We create the dns records for both IPv4 and IPv6 addresses of these
VMs for the load balancer.
4. Client of the customer resolves to the IPv6 address of the load
balancer but the server is not responsive in that address.

This is why we are adding stack option to the load balancer. For now,
this is only added as a backend option and will be made available soon
to the customers as well.


prog/vnet/load_balancer_nexus.rb
```diff
@@ -139,9 +143,26 @@ def before_run
   def rewrite_dns_records
     load_balancer.dns_zone&.delete_record(record_name: load_balancer.hostname)
 
-    load_balancer.vms_to_dns.map do |vm|
-      load_balancer.dns_zone&.insert_record(record_name: load_balancer.hostname, type: "A", ttl: 10, data: vm.ephemeral_net4.to_s) if vm.ephemeral_net4
-      load_balancer.dns_zone&.insert_record(record_name: load_balancer.hostname, type: "AAAA", ttl: 10, data: vm.ephemeral_net6.nth(2).to_s)
+    load_balancer.vms_to_dns.each do |vm|
+      # Insert IPv4 record if stack is ipv4 or double, and vm has IPv4
```
bsatzger, line 147:
```suggestion
      # Insert IPv4 record if stack is ipv4 or dual, and vm has IPv4
```

prog/vnet/load_balancer_nexus.rb
```diff
@@ -139,9 +143,26 @@ def before_run
...
+      if load_balancer.ipv4_enabled? && vm.ephemeral_net4
+        load_balancer.dns_zone&.insert_record(
+          record_name: load_balancer.hostname,
+          type: "A",
+          ttl: 10,
+          data: vm.ephemeral_net4.to_s
+        )
+      end
+
+      # Insert IPv6 record if stack is ipv6 or double
```
bsatzger, line 157:
```suggestion
      # Insert IPv6 record if stack is ipv6 or dual
```

bsatzger (approved):

PR #2201, opened 2024-11-06
Add Latitude.sh Frankfurt as the new region

furkansahin:
## Move latitude-ai region under latitude provider

### DIFF
--- a/config/providers.yml
+++ b/config/providers.yml
@@ -7,7 +7,6 @@
       {display_name: eu-north-h1, internal_name: hetzner-hel1, ui_name: "Finland", visible: true },
       {display_name: github-runners, internal_name: github-runners, ui_name: "GithubRunner", visible: false },
       {display_name: hetzner-ai, internal_name: hetzner-ai, ui_name: "hetzner-ai", visible: false },
-      {display_name: latitude-ai, internal_name: latitude-ai, ui_name: "latitude-ai", visible: false }
     ]
   },
   {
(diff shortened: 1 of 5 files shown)


config/providers.yml
```diff
@@ -17,5 +17,12 @@
       {display_name: us-east-a2, internal_name: leaseweb-wdc02, ui_name: "Virginia, US", visible: true},
       {display_name: us-south-a13, internal_name: leaseweb-dal13,  ui_name: "Dallas, US", visible: false},
     ]
+  },
+  {
+    provider_display_name: Latitude,
+    provider_internal_name: latitude,
+    locations: [
+      {display_name: eu-central-l1, internal_name: latitude-fra, ui_name: "Frankfurt, Germany", visible: false}
```
enescakir, line 25: We use `a` such as `us-east-a2` for Leaseweb. Do you know the reason for that (I don't know)? Should we use `a` for Latitude too? Lower `l` is little hard to differentiate from numbers such as `1`?

enescakir, line 25: Currently, Hetzner regions are named with countries, while others use both city and country names. I recall Umur initiating a discussion on this topic earlier. I can't remember the outcome, but I believe we should adhere to one naming convention.

furkansahin: There is already an active discussion in slack about this :)

furkansahin: this makes sense. I'll update

bsatzger (approved):

lib/option.rb
```diff
@@ -32,8 +32,8 @@ module Option
   PROVIDERS.freeze
   LOCATIONS.freeze
 
-  def self.locations(only_visible: true)
-    Option::LOCATIONS.select { !only_visible || _1.visible }
+  def self.locations(only_visible: true, feature_flags: [])
+    Option::LOCATIONS.select { !only_visible || (_1.visible || feature_flags.include?("location_#{_1.name.gsub(/[-_]/, "_")}")) }
```
enescakir, line 36: Nit: no need to replace underscore with underscore
```suggestion
    Option::LOCATIONS.select { !only_visible || (_1.visible || feature_flags.include?("location_#{_1.name.tr("-", "_")}")) }
```

enescakir (approved):

PR #2202, opened 2024-11-06
Add pulse checks for ensuring disk health on VmHosts

mohi-kalantari:
Multiple checks using different linux tools are added to determine the health of the disks. The code also supports both raided and not raided VmHosts


prog/vm/host_nexus.rb
```diff
@@ -245,9 +245,26 @@ def get_boot_id
   end
 
   def available?
-    sshable.cmd("true")
-    true
-  rescue
+    devices = storage_devices.flat_map { |sd| sd.unix_devices.split(',') }
```
aamederen, line 248: I think you can extract this logic into a function within vmhost and use it in both HostNexus and VmHost

model/vm_host.rb
```diff
@@ -243,6 +244,56 @@ def reset
...
+      ssh_session.exec!("sudo bash -c \"head -c 1M </dev/zero > #{file_name}\"")
+      success = false unless ssh_session.exec!("sha256sum #{file_name}").strip == "30e14955ebf1352266dc2ff8067e68104607e750abb9d3b36582b8af909fcb58  #{file_name}"
+      ssh_session.exec!("sudo rm #{file_name}")
+    end
+    success
+
+  end
+
+  def check_storage_kernel_logs(ssh_session, devices)
+    kernel_logs = ssh_session.exec!("journalctl -kS -5min --no-pager").split("\n")
```
aamederen, line 281: `-5min` is somehow important here. That means, in case of a repair, we would detect unavailable for 5 more minutes. Maybe we can reduce this to 1min for now?

lib/filesystem_parser.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+class LsblkParser
```
aamederen, line 3: Specs for all these that show the expected input to be parsed would be nice.

aamederen:
Since this is also a warm-up, it could be nice to split this change into multiple commits and explain each change in the commit messages. The migration could be one commit, population of the UNIX devices could be another and finally the change in the health check logic, each with their own specs.

lib/filesystem_parser.rb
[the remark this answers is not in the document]
mohi-kalantari: done. created a single class named SystemParser containing all the parsers.

prog/vm/host_nexus.rb
```diff
@@ -245,9 +245,26 @@ def get_boot_id
...
+    is_available = false
+
+    begin
+      checks = [
+        -> { check_storage_smartctl(ssh_session, devices) },
+        -> { check_storage_nvme(ssh_session, devices) },
+        -> { check_storage_read_write(ssh_session, devices) },
+        -> { check_storage_kernel_logs(ssh_session, devices) }
+      ]
+      is_available = checks.all?(&:call)
```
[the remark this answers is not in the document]
mohi-kalantari: I did check how long it takes to check all. On many different tries, it takes around ~4.2s on average


I believe my internet connection is not very good and since these are run on top of the ssh connections, my numbers are not very reliable. Because it takes so long, sometimes I get a log from the monitor saying resource is locked but on its retries, it is resolved

prog/vm/host_nexus.rb
```diff
@@ -245,9 +245,26 @@ def get_boot_id
...
+    ensure
+      ssh_session.close if ssh_session
+    end
+
+    is_available
+
+  rescue => ex
+    Clog.emit("could not check availability of VmHost: #{ex}")
```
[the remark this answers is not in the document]
mohi-kalantari: removed it. done.

prog/vm/host_nexus.rb
mohi-kalantari: done

model/vm_host.rb
mohi-kalantari: good point. I reduced it to 1min.

lib/filesystem_parser.rb
mohi-kalantari: Specs are added for all the parsers.

mohi-kalantari:

mohi-kalantari:
recheck

model/vm_host.rb
```diff
@@ -243,16 +244,70 @@ def reset
...
+  def check_storage_nvme(ssh_session, devices)
+    devices.map { |device_name| ssh_session.exec!("sudo nvme smart-log /dev/#{device_name} | grep \"critical_warning\" | awk '{print $3}'").strip == "0" }.all?(true)
+  end
+
+  def check_storage_read_write(ssh_session, devices)
+    lsblk_json_info = ssh_session.exec!("lsblk --json")
+    device_mount_points = devices.map { |device| SystemParser.get_device_mount_points_from_lsblk(lsblk_json_info, device) }
+
+    all_mount_points = []
+    device_mount_points.each do |device_mount_point|
```
aamederen, line 260: nit: variable naming is a bit confusing in this block. Maybe change it to something like this:

```ruby
    devices_with_mount_points.each do |device_mount_points|
      device_mount_points.each do |device_name, mount_points|
        mount_points.each do |mount_point|
          all_mount_points.append(mount_point)
        end
      end
```

prog/vm/host_nexus.rb
```diff
@@ -245,8 +245,22 @@ def get_boot_id
...
+    ssh_session = sshable.start_fresh_session
+    is_available = false
+
+    begin
+      checks = [
+        -> { vm_host.check_storage_smartctl(ssh_session, vm_host.disk_devices) },
+        -> { vm_host.check_storage_nvme(ssh_session, vm_host.disk_devices) },
+        -> { vm_host.check_storage_read_write(ssh_session, vm_host.disk_devices) },
+        -> { vm_host.check_storage_kernel_logs(ssh_session, vm_host.disk_devices) }
+      ]
```
aamederen, line 257: overall, we do a lot of checks and raise a flag when anything fails. This works, but in case of a failure, the on-call engineer will need to spend some time to understand what was wrong. Maybe emitting logs about what failed inside the check_ functions can be useful.

model/vm_host.rb
```diff
@@ -243,16 +244,70 @@ def reset
...
+      success = false unless ssh_session.exec!("sha256sum #{file_name}").strip == "30e14955ebf1352266dc2ff8067e68104607e750abb9d3b36582b8af909fcb58  #{file_name}"
+      ssh_session.exec!("sudo rm #{file_name}")
+    end
+    success
+  end
+
+  def check_storage_kernel_logs(ssh_session, devices)
+    kernel_logs = ssh_session.exec!("journalctl -kS -1min --no-pager").split("\n")
+    error_count = Hash.new(0)
+    kernel_logs.reduce(error_count) do |error_count, line|
```
aamederen, line 281: Can we simplify this by just doing a regex search without reduce?

lib/system_parser.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+class SystemParser
+  DfRecord = Struct.new(:unix_device, :optional_name, :size_gib, :avail_gib)
+
+  def self.get_device_mount_points_from_lsblk(json_input, device_name)
```
aamederen, line 6: nit: maybe rename to get_device_mount_points_from_lsblk_json

mohi-kalantari: done

model/vm_host.rb
mohi-kalantari: done

prog/vm/host_nexus.rb
mohi-kalantari: you're right. I added logs to checkers

model/vm_host.rb
```diff
@@ -243,16 +244,79 @@ def reset
...
+  def disk_devices
+    storage_devices.map { |sd| sd.set_underlying_unix_devices if sd.unix_device_list.empty? }
+    storage_devices.flat_map { |sd| sd.unix_device_list.split(",") }
+  end
+
   def check_pulse(session:, previous_pulse:)
     reading = begin
-      session[:ssh_session].exec!("true")
-      "up"
+      checks = [
```
aamederen, line 313: I think we have room to increase reuse even further by moving all this into a single function too.

model/vm_host.rb
```diff
@@ -243,16 +244,79 @@ def reset
...
+      write_status && hash_status && delete_status
+    end
+  end
+
+  def check_storage_kernel_logs(ssh_session, devices)
+    kernel_logs = ssh_session.exec!("journalctl -kS -1min --no-pager")
+    return false unless kernel_logs.exitstatus == 0
+
+    error_count = kernel_logs.scan(/Buffer I\/O error on dev (\w+)/).tally
+    Clog.emit("found error on kernel logs. devices with error_count: #{error_count}")
```
aamederen, line 296: will this emit a message, even if error_count is 0?

model/vm_host.rb
```diff
@@ -243,16 +244,79 @@ def reset
...
+  end
+
   def init_health_monitor_session
     {
       ssh_session: sshable.start_fresh_session
     }
   end
 
+  def disk_devices
+    storage_devices.map { |sd| sd.set_underlying_unix_devices if sd.unix_device_list.empty? }
```
aamederen, line 307: can we use `each` instead of `map`  here?

mohi-kalantari: done.

model/vm_host.rb
mohi-kalantari: my bad :D fixed it. I also tested all scenarios manually to make sure logs are correct.

aamederen (approved):
LGTM, I would however like to have one more eye on this

fdr (changes requested):
at a minimum, two things:

* If storing an array, why not use either a text array, or even more likely, a jsonb
* The migration should be broken out from other application changes to not make rollbacks fussy.

prog/vm/host_nexus.rb
```diff
@@ -245,8 +245,16 @@ def get_boot_id
...
+    ssh_session = sshable.start_fresh_session
+    is_available = false
+
+    begin
+      vm_host.perform_health_checks(ssh_session)
+    ensure
+      ssh_session&.close
+    end
+
+    is_available
```
fdr, line 257: This is a big change in behavior, slowing down `available?` quite a bit, by losing connection caching.

model/vm_host.rb
```diff
@@ -243,16 +244,83 @@ def reset
     Hosting::Apis.reset_server(self)
   end
 
+  def check_storage_smartctl(ssh_session, devices)
+    devices.map do |device_name|
+      result = ssh_session.exec!("sudo smartctl -j -H /dev/#{device_name} | jq .smart_status.passed").strip == "true"
+      result.tap do |passed|
+        Clog.emit("Device #{device_name} failed smartctl check.") unless passed
+      end
```
jeremyevans, line 252:
```suggestion
      passed = ssh_session.exec!("sudo smartctl -j -H /dev/#{device_name} | jq .smart_status.passed").strip == "true"
      Clog.emit("Device #{device_name} failed smartctl check.") unless passed
      passed
```

`tap` generally makes sense if you are using it during a method chain.  If you already have a local variable, it's simpler to just use it.  Same for other methods in this file.

model/vm_host.rb
```diff
@@ -243,16 +244,83 @@ def reset
...
+  def check_storage_read_write(ssh_session, devices)
+    lsblk_json_info = ssh_session.exec!("lsblk --json")
+    devices_with_mount_points = devices.map { |device| SystemParser.get_device_mount_points_from_lsblk_json(lsblk_json_info, device) }
+
+    all_mount_points = []
+    devices_with_mount_points.each do |device_mount_points|
+      device_mount_points.each do |device_name, mount_points|
+        mount_points.each do |mount_point|
+          all_mount_points.append(mount_point)
+        end
```
jeremyevans, line 274:
```suggestion
        all_mount_points.concat(mount_points)
```

model/vm_host.rb
```diff
@@ -243,16 +244,83 @@ def reset
...
+    end.all?(true)
+  end
+
+  def check_storage_read_write(ssh_session, devices)
+    lsblk_json_info = ssh_session.exec!("lsblk --json")
+    devices_with_mount_points = devices.map { |device| SystemParser.get_device_mount_points_from_lsblk_json(lsblk_json_info, device) }
+
+    all_mount_points = []
+    devices_with_mount_points.each do |device_mount_points|
+      device_mount_points.each do |device_name, mount_points|
```
jeremyevans, line 271:
```suggestion
      device_mount_points.each_value do |mount_points|
```

model/vm_host.rb
```diff
@@ -243,16 +244,83 @@ def reset
...
+        mount_points.each do |mount_point|
+          all_mount_points.append(mount_point)
+        end
+      end
+    end
+
+    all_mount_points.uniq.all? do |mount_point|
+      file_name = (mount_point == "/") ? "/test-file" : "#{mount_point}/test-file"
```
jeremyevans, line 279: VmHost mount points should not contain attacker-controlled information. Even still, it may not be a bad idea to shell escape this (using shellwords in stdlib).

model/vm_host.rb
```diff
@@ -243,16 +244,83 @@ def reset
...
+      write_status = ssh_session.exec!("sudo bash -c \"head -c 1M </dev/zero > #{file_name}\"").exitstatus == 0
+      hash_status = ssh_session.exec!("sha256sum #{file_name}").strip == "30e14955ebf1352266dc2ff8067e68104607e750abb9d3b36582b8af909fcb58  #{file_name}"
+      delete_status = ssh_session.exec!("sudo rm #{file_name}").exitstatus == 0
+
+      if !(write_status && hash_status && delete_status)
+        Clog.emit("failed to perform read/write on mountpoint #{mount_point}")
+      end
+      write_status && hash_status && delete_status
```
jeremyevans, line 287:
```suggestion
      working = write_status && hash_status && delete_status
      Clog.emit("failed to perform read/write on mountpoint #{mount_point}") unless working
      working
```

model/vm_host.rb
```diff
@@ -243,16 +244,83 @@ def reset
...
+
   def init_health_monitor_session
     {
       ssh_session: sshable.start_fresh_session
     }
   end
 
+  def disk_devices
+    storage_devices.each { |sd| sd.set_underlying_unix_devices if sd.unix_device_list.empty? }
+    storage_devices.flat_map { |sd| sd.unix_device_list.split(",") }
```
jeremyevans, line 308:
```suggestion
    storage_devices.flat_map { |sd| sd.unix_device_list }
```

When using array column.

model/vm_host.rb
```diff
@@ -243,16 +244,83 @@ def reset
...
+  end
+
+  def perform_health_checks(ssh_session)
+    checks = [
+        -> { check_storage_smartctl(ssh_session, disk_devices) },
+        -> { check_storage_nvme(ssh_session, disk_devices) },
+        -> { check_storage_read_write(ssh_session, disk_devices) },
+        -> { check_storage_kernel_logs(ssh_session, disk_devices) }
+      ]
+      checks.all?(&:call)
```
jeremyevans, line 318:
```suggestion
    check_storage_smartctl(ssh_session, disk_devices) &&
      check_storage_nvme(ssh_session, disk_devices) &&
      check_storage_read_write(ssh_session, disk_devices) &&
      check_storage_kernel_logs(ssh_session, disk_devices)
```

Best to avoid unnecessary indirection.

prog/learn_storage.rb
```diff
@@ -1,51 +1,39 @@
...
+        unix_device_list: find_underlying_unix_devices(rec.unix_device)
       )
     end
 
     sds
   end
 
+  def find_underlying_unix_devices(unix_device)
+    return unix_device unless %r{^/dev/md}.match?(unix_device)
+    SystemParser.extract_unerlying_raid_devices_from_mdstat(sshable.cmd("cat /proc/mdstat"), unix_device).join(",")
```
jeremyevans, line 34:
```suggestion
    SystemParser.extract_underlying_raid_devices_from_mdstat(sshable.cmd("cat /proc/mdstat"), unix_device)
```

prog/vm/host_nexus.rb
```diff
@@ -245,8 +245,16 @@ def get_boot_id
...
   def available?
-    sshable.cmd("true")
-    true
+    ssh_session = sshable.start_fresh_session
+    is_available = false
+
+    begin
+      vm_host.perform_health_checks(ssh_session)
+    ensure
+      ssh_session&.close
```
jeremyevans, line 254:
```suggestion
      ssh_session.close
```

Unless `sshable.start_fresh_session` can return nil, by the time you get here, it should be set.

lib/system_parser.rb
```diff
@@ -0,0 +1,79 @@
...
+  #
+  # md0 : active raid1 nvme1n1p1[1] nvme0n1p1[0]
+  #       33520640 blocks super 1.2 [2/2] [UU]
+  #
+  # md1 : active raid1 nvme1n1p2[1] nvme0n1p2[0]
+  #       1046528 blocks super 1.2 [2/2] [UU]
+  #
+  # unused devices: <none>
+  #
+  def self.extract_unerlying_raid_devices_from_mdstat(mdstat_file_content, raid_device_name)
```
jeremyevans, line 46:
```suggestion
  def self.extract_underlying_raid_devices_from_mdstat(mdstat_file_content, raid_device_name)
```

jeremyevans (commented):
I don't have much of a background here, so my comments focus on general Ruby issues.  The most important part would be using a text array column in the database, instead of a comma separated list.

model/vm_host.rb
fdr: What do you think of using `flat_map`s to do the job?

jeremyevans: That approach is fine, but it likely results in more intermediate arrays, and with nested `flat_map`s, I think it's a little less clear.

prog/vm/host_nexus.rb
mohi-kalantari: fixed the issue thanks

lib/system_parser.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+class SystemParser
+  DfRecord = Struct.new(:unix_device, :optional_name, :size_gib, :avail_gib)
+
+  def self.get_device_mount_points_from_lsblk_json(json_input, device_name)
+    data = JSON.parse(json_input)
+    excluded_mounts = ["[SWAP]", "/boot"]
+
+    mount_points = Hash.new { |hash, key| hash[key] = [] }
```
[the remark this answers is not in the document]
mohi-kalantari: fixed the issue. it was also insightful thanks

model/vm_host.rb
mohi-kalantari: got it thanks. fixed it

mohi-kalantari:
I'll also fix the commits and separate the migration commit and others when review is done. thanks for pointing out. @fdr

lib/system_parser.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    mount_points = {}
+
+    data["blockdevices"].each do |device|
+      next unless device_name == device["name"]
+
+      get_mounts = lambda do |dev|
+        dev["mountpoints"]&.each do |mount|
+          mount_points[device["name"]] ||= []
+          mount_points[device["name"]] << mount unless excluded_mounts.include?(mount)
```
jeremyevans, line 18:
```suggestion
          (mount_points[device["name"]] ||= []) << mount unless excluded_mounts.include?(mount)
```

Alternatively, switch to having the unless around both lines.  The goal would be to not to have an empty array for an excluded mount.

jeremyevans (commented):
Looks good.  I only have one remaining comment.

lib/system_parser.rb
mohi-kalantari: done thanks

cache/schema.cache
jeremyevans, line 1: This looks like a bug, since the schema cache should not shrink when adding a column.  This file was likely created before the pg_schema_cache usage.  Try running `rake test_up` again. and committing the updated file.

jeremyevans (commented):
Previous comment was addressed.  One additional comment.

cache/schema.cache
mohi-kalantari: I fixed it thanks.

byucesoy:
Hey @mohi-kalantari, do we have any blocker for this PR, other than Daniel's comments?

byucesoy (commented):
Overall, looks OK IMO.

@mohi-kalantari, do we have any blocker for this PR, other than Daniel's comments?

fdr (approved):

PR #2209, opened 2024-11-06
Remove unnecessary to_json calls

byucesoy:
Roda's json plugin automatically converts Ruby hashes into `JSON` texts, so
these calls were unnecessary. Also, manually calling `to_json` requires setting
`Content-Type` header manually to `application/json`, which is different than
the default for our web endpoints and easy to forget. However Roda's json  
plugin also correctly sets the headers.

### DIFF
--- a/routes/web/project.rb
+++ b/routes/web/project.rb
@@ -64,13 +64,13 @@ class Clover
         # If it has some resources, do not allow to delete it.
         if @project.has_resources
           flash["error"] = "'#{@project.name}' project has some resources. Delete all related resources first."
-          return {message: "'#{@project.name}' project has some resources. Delete all related resources first."}.to_json
+          return {message: "'#{@project.name}' project has some resources. Delete all related resources first."}
         end
 
         @project.soft_delete
 
         flash["notice"] = "'#{@project.name}' project is deleted."
-        return {message: "'#{@project.name}' project is deleted."}.to_json
+        return {message: "'#{@project.name}' project is deleted."}
       end
 
       r.get "dashboard" do
--- a/routes/web/project/billing.rb
+++ b/routes/web/project/billing.rb
@@ -141,12 +141,12 @@ class Clover
         r.delete true do
           unless payment_method.billing_info.payment_methods.count > 1
             response.status = 400
-            return {message: "You can't delete the last payment method of a project."}.to_json
+            return {message: "You can't delete the last payment method of a project."}
           end
 
           payment_method.destroy
 
-          return {message: "Deleting #{payment_method.ubid}"}.to_json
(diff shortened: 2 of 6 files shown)


enescakir:
Since these are delete requests, we can change responses to `204 No Content` similar to other delete requests

byucesoy:
> Since these are delete requests, we can change responses to `204 No Content` similar to other delete requests

I think that makes sense. AFAIR, we don't use the response content in the frontend, only the response code.

jeremyevans (commented):
I think this is better fixed by removing the `to_json` calls.  This allows Roda's json plugin to automatically set the content type.  I've been doing that in the routes I've been merging.

byucesoy:
> I think this is better fixed by removing the `to_json` calls. This allows Roda's json plugin to automatically set the content type. I've been doing that in the routes I've been merging.

If we remove `to_json` calls but keep the response as Ruby hash, content-type would be set to `application/json` automatically, did I understand that correctly?

jeremyevans:
> > I think this is better fixed by removing the `to_json` calls. This allows Roda's json plugin to automatically set the content type. I've been doing that in the routes I've been merging.
> 
> If we remove `to_json` calls but keep the response as Ruby hash, content-type would be set to `application/json` automatically, did I understand that correctly?

Yes, you did: https://roda.jeremyevans.net/rdoc/classes/Roda/RodaPlugins/Json.html

byucesoy:
> > > I think this is better fixed by removing the `to_json` calls. This allows Roda's json plugin to automatically set the content type. I've been doing that in the routes I've been merging.
> > 
> > 
> > If we remove `to_json` calls but keep the response as Ruby hash, content-type would be set to `application/json` automatically, did I understand that correctly?
> 
> Yes, you did: https://roda.jeremyevans.net/rdoc/classes/Roda/RodaPlugins/Json.html

Cool. I was also testing the behavior and confirmed it, then I saw your reply. :+1:

In that case, I'm going with removing the `to_json` calls as it is less invasive. I removed all unnecessary `to_json` calls expect the one in `private_subnet.rb`, to prevent conflict with @jeremyevans' PR. In the future we can also check the each frontend call site to ensure if the response is not used and return `204 No-Content` in we want.

jeremyevans (approved):
This looks good.  Note that most of the return statement are also unnecessary.  In many cases `return` can be removed, but if not, changing them to `next` makes more sense.  `return` in these cases works by accident, because the `hash_branch` calls are turned into method definitions (which is an internal Roda implementation detail, that has changed before, and could potentially change back now that `instance_exec` no longer allocates a singleton class). However, the `return` removal could be a separate commit.

This is going to conflict with the firewall route merging I'm working, but that's OK, I can rebase on top of this after it is merged.

PR #2212, opened 2024-11-06
Remove explicit calls to Clover.freeze

jeremyevans:
In production mode and when running the frozen specs, clover_freeze now calls freeze on all autoloaded constants, and Clover is an autoloaded constant. Both the specs and config.ru call clover_freeze unconditionally.

This also removes the override of Clover.freeze.  In production and when running the frozen specs, it would call
Sequel::Model.freeze_descendants and DB.freeze.  The freezing of all autoloaded constants handles DB.freeze, and clover_freeze calls Sequel::Model.freeze_descendants before freezing all autoloaded constants.

This does remove the
`Sequel::Model.descendants.each(&:finalize_associations)` when running non-frozen specs, but that's solely an optimization and I don't think it is worth the complexity.

The `return self if frozen?` code was added to work around a bug in autoload_hash_branches, where Roda.freeze would break if you called it the second time on the same object. Now that Clover.freeze is only called once, it is not needed.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -4,6 +4,6 @@ require_relative "loader"
 
 clover_freeze
 
-run(Config.development? ? Unreloader : Clover.freeze.app)
+run(Config.development? ? Unreloader : Clover.app)
 
 Tilt.finalize! unless Config.development?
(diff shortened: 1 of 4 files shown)


fdr (commented):
lgtm

bsatzger (approved):

PR #2213, opened 2024-11-06
Apply order to runtime github/caches route query

jeremyevans:
Fixes nondeterministic test failure:

```
  1) Clover github cache endpoints lists cache entries returns the list of cache entries for the key
     Failure/Error: expect(response["artifactCaches"].map { [_1["cacheKey"], _1["cacheVersion"]] }).to eq([["k1", "v1"], ["k1", "v2"]])

       expected: [["k1", "v1"], ["k1", "v2"]]
            got: [["k1", "v2"], ["k1", "v1"]]
```

### DIFF
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -64,7 +64,8 @@ class Clover
         scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
         entries = repository.cache_entries_dataset
           .exclude(committed_at: nil)
-          .where(key: key, scope: scopes).all
+          .where(key: key, scope: scopes)
+          .order(:version).all
 
         {
           totalCount: entries.count,


fdr (approved):

routes/runtime/github.rb
```diff
@@ -64,7 +64,8 @@ class Clover
         scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
         entries = repository.cache_entries_dataset
           .exclude(committed_at: nil)
-          .where(key: key, scope: scopes).all
+          .where(key: key, scope: scopes)
+          .order(:version).all
```
enescakir, line 68: We can order by `Sequel.desc(:created_at)`

jeremyevans: Not in the specs, due to the use of transactions.  We could do `Sequel.desc(:created_at), :version`, would you prefer that?

jeremyevans: Also, FYI, `reverse(:created_at)` is the same as `order(Sequel.desc(:created_at))`

enescakir: Then let's keep it as `version`

enescakir (approved):

PR #2210, opened 2024-11-06
Merge firewall routes

jeremyevans:
@furkansahin I'm asking you to review this because I discovered when merging the routes that we are  using different authorization for the web and api routes.  This applies both authorizations when the specs pass when doing so.  The attach/detach api subnet routes failed if I applied the same authorization used on the web routes to them, so this currently has the api and web attach/detach subnet routes use different authorization.  We don't want that, but I don't know what authorization we do want.  Can you review these commits and determine what authorization we actually want:

* https://github.com/ubicloud/ubicloud/commit/d067c9db90debbd2ce834e6ed81e68e8374a8323
* https://github.com/ubicloud/ubicloud/commit/0322af05307926442a33b1d29796c7edac64e4d5
* https://github.com/ubicloud/ubicloud/commit/548b0a3f69ed5c4e576bdfbd06962c0091d4c412
* https://github.com/ubicloud/ubicloud/commit/5f4871437120f55878c25eb15602df88828dd6ae (didn't test if attach-subnet authorization issue also affects detach-subnet)

Additionally, I found a couple cases that did `response.status = 404` before redirecting.  This doesn't have an effect (redirecting overrides to a 30x status), so I removed it.  However, the flash messages say not found, so maybe it is better to treat these as 404 responses and not redirect.  Please let me know which approach is preferable:

* https://github.com/ubicloud/ubicloud/commit/a3dec8e18ee6e6a15c4ab781444cbe0cc498109b

PR #2216, opened 2024-11-07
Merge load balancer routes

enescakir:
Heavily inspired by #2196


routes/merged/project/load_balancer.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    r.on api? do
+      r.on String do |lb_name|
+        r.post true do
+          load_balancer_post(name: lb_name)
+        end
+      end
+    end
+
+    r.on web? do
```
jeremyevans, line 17: This isn't a problem, but the `r.on web?` call is unncessary. With the above `r.on api?` call, you know that everything after that until the end of the block must be a web route and not an api route.

routes/merged/project/load_balancer.rb
```diff
@@ -0,0 +1,33 @@
...
+    r.get true do
+      load_balancer_list
+    end
+
+    r.on api? do
+      r.on String do |lb_name|
+        r.post true do
+          load_balancer_post(name: lb_name)
+        end
+      end
```
jeremyevans, line 14:
```suggestion
      r.post String do |lb_name|
        load_balancer_post(name: lb_name)
      end
```

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+
+        if api?
+          Serializers::LoadBalancer.serialize(lb, {detailed: true})
+        else
+          flash["notice"] = "VM is attached"
+          r.redirect "#{@project.path}#{lb.path}"
+        end
+      end
+
+      r.post "detach-vm" do
```
jeremyevans, line 58: There is a lot of overlap with attach-vm, we could consider combining the two blocks, similar to https://github.com/ubicloud/ubicloud/pull/2210/commits/5f4871437120f55878c25eb15602df88828dd6ae

That doesn't need to be done before merging, though.

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+          dst_port: Validation.validate_port(:dst_port, request_body_params["dst_port"]),
+          health_check_endpoint: request_body_params["health_check_endpoint"]
+        )
+
+        request_body_params["vms"].each do |vm_id|
+          vm_id = vm_id.delete("\"")
+          vm = Vm.from_ubid(vm_id)
+          unless vm
+            response.status = 404
+            r.halt
```
jeremyevans, line 122:
```suggestion
            next
```

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+            fail Validation::ValidationFailed.new("vm_id" => "VM is already attached to a load balancer")
+          elsif vm.load_balancer && vm.load_balancer.id == lb.id
+            next
+          end
+
+          lb.add_vm(vm)
+        end
+
+        lb.vms.map { _1.ubid }.reject { request_body_params["vms"].map { |vm_id| vm_id.delete("\"") }.include?(_1) }.each do |vm_id|
+          vm = Vm.from_ubid(vm_id)
```
jeremyevans, line 136: Not an issue with this PR, since this was a problem with the previous code, but this section is an N+1 query issue.  Removing the `map { _1.ubid }` and updating the `reject` block would allow removing the `vm = Vm.from_ubid(vm_id)` inside the loop.

routes/helpers/load_balancer.rb
```diff
@@ -0,0 +1,62 @@
...
+        items: Serializers::LoadBalancer.serialize(result[:records]),
+        count: result[:count]
+      }
+    else
+      @lbs = Serializers::LoadBalancer.serialize(dataset.all, {include_path: true})
+      view "networking/load_balancer/index"
+    end
+  end
+
+  def load_balancer_post(name: nil)
```
jeremyevans, line 25: All calls provide this argument.  While it doesn't need to done in this PR, we should probably switch this to a required positional argument.

routes/helpers/load_balancer.rb
```diff
@@ -0,0 +1,62 @@
...
+    Authorization.authorize(current_account.id, "LoadBalancer:create", @project.id)
+
+    required_parameters = %w[private_subnet_id algorithm src_port dst_port health_check_protocol]
+    required_parameters << "name" if web?
+    optional_parameters = %w[health_check_endpoint]
+    request_body_params = Validation.validate_request_body(json_params, required_parameters, optional_parameters)
+
+    ps = PrivateSubnet.from_ubid(request_body_params["private_subnet_id"])
+    unless ps
+      response.status = 404
```
jeremyevans, line 35:
```suggestion
```

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+      r.post "attach-vm" do
+        Authorization.authorize(current_account.id, "LoadBalancer:edit", lb.id)
+        required_parameters = %w[vm_id]
+        request_body_params = Validation.validate_request_body(json_params, required_parameters)
+        vm = Vm.from_ubid(request_body_params["vm_id"])
+
+        unless vm
+          response.status = 404
+          if api?
+            r.halt
```
jeremyevans, line 33:
```suggestion
          if api?
            next
```

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+            flash["error"] = "VM not found"
+            r.redirect "#{@project.path}#{lb.path}"
+          end
+        end
+
+        Authorization.authorize(current_account.id, "Vm:view", vm.id)
+
+        if vm.load_balancer
+          flash["error"] = "VM is already attached to a load balancer"
+          response.status = 400
```
jeremyevans, line 44:
```suggestion
```

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+          else
+            flash["error"] = "VM not found"
+            r.redirect "#{@project.path}#{lb.path}"
+          end
+        end
+
+        Authorization.authorize(current_account.id, "Vm:view", vm.id)
+
+        if vm.load_balancer
+          flash["error"] = "VM is already attached to a load balancer"
```
jeremyevans, line 43: This `if vm.load_balancer` section seems web specific, when the rest of the route handles both api and web.  I'm guessing if any api request hits this, it does something we do not want.  This was a bug in the previous code, not in this PR.  So it isn't a blocker, but we should add api handling to this section.

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+        Authorization.authorize(current_account.id, "LoadBalancer:edit", lb.id)
+        required_parameters = %w[vm_id]
+        request_body_params = Validation.validate_request_body(json_params, required_parameters)
+
+        vm = Vm.from_ubid(request_body_params["vm_id"])
+        unless vm
+          response.status = 404
+
+          if api?
+            r.halt
```
jeremyevans, line 68:
```suggestion
          if api?
            next
```

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,158 @@
...
+          lb.add_vm(vm)
+        end
+
+        lb.vms.map { _1.ubid }.reject { request_body_params["vms"].map { |vm_id| vm_id.delete("\"") }.include?(_1) }.each do |vm_id|
+          vm = Vm.from_ubid(vm_id)
+          lb.evacuate_vm(vm)
+          lb.remove_vm(vm)
+        end
+
+        response.status = 200
```
jeremyevans, line 141:
```suggestion
```

I don't think this is needed, as 200 is the default status if there is a response (you would only need this if the `serialize` call below could return nil/false.

jeremyevans (changes requested):
Great work! I have some specific suggestions that should be easier to make before merging, along with some comments for things that could be made either before or after merging, whichever you prefer.

routes/merged/project/load_balancer.rb
enescakir: If I remember correctly, when we use `r.on`, it only applies to that specific branch. However, `r.is` checks all branches until it finds a match. Am I correct?

So, if it's `r.is api?`, we need to specify this line, right?

routes/merged/project/location/load_balancer.rb
enescakir: Merged at caf663f3945b4992a1498b44e7e102c872ea42f9

routes/merged/project/location/load_balancer.rb
enescakir: The routes aren't run in the database transactions, right? Should we put the patch route in to a transaction? If anything goes wrong, we could end up with a partially updated resource.

routes/merged/project/load_balancer.rb
jeremyevans: If we were using `r.is api?` above, then yes, but not for the reason you think.  The only difference between `r.is` and `r.on` is that `r.is` checks that the path has been fully consumed after all matchers are applied.  So with `r.on api?`, all api requests will go to that block, and you know the rest of the requests after `r.on api?` are web requests.  With `r.is api?`, only api requests were the remaining path is empty would go to that block.  After `r.is api?`, you would have all web requests, and also any api requests where the remaining path is not empty.

routes/merged/project/location/load_balancer.rb
jeremyevans: Understood.

routes/merged/project/location/load_balancer.rb
jeremyevans: Looks good!

jeremyevans: Correct.  Anytime we are doing multiple queries in a route, we should surround it in a transaction.  That can be done after the merge, though.

routes/merged/project/location/load_balancer.rb
```diff
@@ -0,0 +1,130 @@
...
+        Authorization.authorize(current_account.id, "Vm:view", vm.id)
+
+        if action == "attach-vm"
+          if vm.load_balancer
+            flash["error"] = "VM is already attached to a load balancer"
+            response.status = 400
+            r.redirect "#{@project.path}#{lb.path}"
+          end
+          lb.add_vm(vm)
+          actioned = "attached"
```
jeremyevans, line 49: This should probably be "attached to", and detach should use "detached from", and the flash notice should remove the " from".  This is because the flash notice should say "attached to the load balancer" and not "attached from the load balancer".

enescakir: Done

routes/merged/project/location/load_balancer.rb
enescakir: I removed the branch at 0b25439d6f0d2384902950edb8e320c913cfe9d5

routes/merged/project/location/load_balancer.rb
enescakir: Refactored at b2737fa95ca05eb7b59c28e52210ca894e9604d6

routes/merged/project/location/load_balancer.rb
enescakir: Removed

routes/merged/project/location/load_balancer.rb
enescakir: Refactored at b2737fa95ca05eb7b59c28e52210ca894e9604d6

routes/merged/project/location/load_balancer.rb
enescakir: Refactored at b2737fa95ca05eb7b59c28e52210ca894e9604d6

routes/merged/project/location/load_balancer.rb
enescakir: Refactored at b2737fa95ca05eb7b59c28e52210ca894e9604d6

jeremyevans (approved):
All changes look good!

PR #2220, opened 2024-11-07
Check if there are any missed dns records to add

furkansahin:


prog/vnet/load_balancer_nexus.rb
```diff
@@ -51,18 +51,35 @@ def before_run
...
-      decr_rewrite_dns_records
+      hop_rewrite_dns_records
     end
 
     if load_balancer.need_certificates?
       load_balancer.incr_refresh_cert
       hop_create_new_cert
     end
 
+    if need_to_rewrite_dns_records?
```
bsatzger, line 62: I see how this makes DNS record creation more robust but with this change we now perform somewhat costly checks every 5 seconds.

bsatzger: When we have more load balancers in production it would be good to revisit this.

bsatzger (approved):

PR #2221, opened 2024-11-07
Update erb-formatter version

enescakir:
- **Update erb-formatter version**
  We were using a forked version of nebula/erb-formatter due to these two
  commits:
  - https://github.com/ubicloud/erb-formatter/commit/05b30f687f78a78dd7d75760ca05da4a6355317c
  - https://github.com/ubicloud/erb-formatter/commit/a9ff0001a1eb028e2186b222aeb02b07c04f9808
  
  These issues have been fixed upstream, so I initially planned to switch
  to the official package. However, I encountered another problem and
  created two PRs to address them. Unfortunately, I didn't receive any
  response from the maintainer. Thus, I've decided to continue using our
  updated forked version.
  - https://github.com/nebulab/erb-formatter/pull/58
  - https://github.com/nebulab/erb-formatter/pull/59
  

- **Fix icon.erb ignoring for erb-formatter**
  

- **Increase erb-formatter line-width from 120 to 160**
  We format HTML/ERB files using the erb-formatter and Tailwind CSS,
  resulting in many long lines in our view files. After trying various
  line-widths, 160 appears to be the most suitable for our needs to me.
  

- **Fix unclosed </br> tag at otp_setup.erb**
  Both `<br>` and `</br>` are valid HTML tags. But erb-formatter couldn't
  format `</br>` and raises the following exception.
  
      ==> ERROR: Unmatched close tag, tried with ["br", "</br>"], but
      ["p", "<p class=\"mt-2 text-sm text-gray-500 text-center\">"] was on
      the stack
  

- **Format with new version of erb-formatter**
  ```
  rake linter:erb_formatter
  ```


enescakir:
erb-formatter and rubocop are conflicting. Let me check first

PR #2222, opened 2024-11-07
Update erb-formatter version

enescakir:
### **Fix unclosed `</br>` tag at otp_setup.erb**
  Both `<br>` and `</br>` are valid HTML tags. But erb-formatter couldn't
  format `</br>` and raises the following exception.
  
      ==> ERROR: Unmatched close tag, tried with ["br", "</br>"], but
      ["p", "<p class=\"mt-2 text-sm text-gray-500 text-center\">"] was on
      the stack
  

### **Fix icon.erb ignoring for erb-formatter**
  

### **Format mistakenly ignored files with erb-formatter**
  As expected all ERB files starting with the letters 'i', 'c', 'o', 'n'
  are formatted.
  
  One thing I'm not keen on is how the ERB formatter converts the ternary
  operator into an if-else statement. This makes the code longer and more
  difficult to read, and unfortunately, there's no option to disable this
  feature.
  

### **Update erb-formatter version**
  We were using a forked version of nebula/erb-formatter due to these two
  commits:
  - https://github.com/ubicloud/erb-formatter/commit/05b30f687f78a78dd7d75760ca05da4a6355317c
  - https://github.com/ubicloud/erb-formatter/commit/a9ff0001a1eb028e2186b222aeb02b07c04f9808
  
  These issues have been fixed upstream, so I initially planned to switch
  to the official package. However, I encountered another problem and
  created two PRs to address them. Unfortunately, I didn't receive any
  response from the maintainer. Thus, I've decided to continue using our
  updated forked version.
  - https://github.com/nebulab/erb-formatter/pull/58
  - https://github.com/nebulab/erb-formatter/pull/59
  

### **Format with new version of erb-formatter**
  The biggest change multiline classes and trailing commas.
  
    rake linter:erb_formatter
  

### **Disable Layout/AssignmentIndentation cop for erb files**
  erb-formatter and rubocop are conflicting with each other. Since
  erb-formatter formats the files, I disabled the cop.
  

### **Increase erb-formatter line-width from 120 to 160**
  We format HTML/ERB files using the erb-formatter and Tailwind CSS,
  resulting in many long lines in our view files. After trying various
  line-widths, 160 appears to be the most suitable for our needs to me.


jeremyevans (approved):
I reviewed this with the following, which I found much easier than the GitHub UI:

```
git diff --color-words --word-diff-regex='\\w+|[^[:space:]]' -bw
```

Most changes are comma additions, with some being ternary operator removal.

It's  a little weird that erb_formatter forces trailing commas, and Rubocop complains about trailing commas, but not a big deal.

fdr:
Ironically, I'm okay with that, because the tool being good at rewrites is
more important to me than absolute consistency, re: some previous
discussion of consistency.

enescakir:
> It's  a little weird that erb_formatter forces trailing commas, and Rubocop complains about trailing commas, but not a big deal.

If you prefer, we can revert this commit for our fork

https://github.com/nebulab/erb-formatter/commit/53ea6305ebc9808b0e5078ae601b22fcee18b7b8

jeremyevans:
I don't feel strongly enough about it.  Plus, it will be one more thing different in the fork.  Long term, hopefully upstream fixes the issues and we don't have to maintain our own fork.

enescakir:
We already upgraded it

PR #2224, opened 2024-11-07
Upgrade asdf .tool-versions Postgres to 15.8

fdr:
I noticed we haven't picked up point releases in development for a while, so I'm bumping this.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 ruby 3.2.6
-postgres 15.4
+postgres 15.8
 nodejs 22.9.0


jeremyevans (approved):
asdf's default of a minor version specific PGDATA makes this update more annoying than it needs to be (maybe there is a way to fix that?).

Up to you if you want to update DEVELOPERS.md to show this new PostgreSQL version.

fdr:
> asdf's default of a minor version specific PGDATA makes this update more annoying than it needs to be (maybe there is a way to fix that?).

Yeah, I think I "mv" the initdb.

> Up to you if you want to update DEVELOPERS.md to show this new PostgreSQL version.

Maybe I'll leave it until we change major versions, which we probably should once 17 has ripened a bit.

PR #2227, opened 2024-11-08
Use features that will be in Roda 3.86 to simplify app

jeremyevans:
This uses the new response.skip_content_security_policy! method and conditional_sessions plugin to reduce the amount of custom code.

I'm on the fence about whether we should merge this now, or wait until Roda 3.86.0 is released next week.


fdr:
How much conflict drag do you think there'd be? If more, I'd ship it, if less, I'd sit on it

jeremyevans:
I doubt this will conflict much.  I'll leave this here until Roda 3.86 is released, then update the Gemfile to use that.

fdr (approved):

PR #2229, opened 2024-11-08
Simplify Model inspect/to_s using Sequel 5.87 inspect_pk plugin

jeremyevans:


jeremyevans:
This no longer needs Gemfile changes, since we already updated the Gemfile in order to use the pg_schema_caching extension.

enescakir (approved):

PR #2232, opened 2024-11-08
Refactor Authorization.matched_policies

jeremyevans:
This series of commits will make it easier for us to support api key specific access policies.  It changes the SQL-based query to a Sequel-based query, and then speeds up other methods by calling dataset methods, instead of calling methods on array returned by `matched_policies`.

### DIFF
--- a/lib/authorization.rb
+++ b/lib/authorization.rb
@@ -22,11 +22,11 @@ module Authorization
   end
 
   def self.authorized_resources_dataset(subject_id, actions)
-    matched_policies_dataset(subject_id, actions).select(:tagged_id)
+    matched_policies_dataset(subject_id, actions).select(Sequel[:applied_tag][:tagged_id])
   end
 
   def self.authorized_resources(subject_id, actions)
-    matched_policies_dataset(subject_id, actions).select_map(:tagged_id)
+    matched_policies_dataset(subject_id, actions).select_map(Sequel[:applied_tag][:tagged_id])
   end
 
   def self.expand_actions(actions)
@@ -40,19 +40,17 @@ module Authorization
   end
 
   def self.matched_policies_dataset(subject_id, actions = nil, object_id = nil)
-    subject_dataset = DB[:access_tag]
-      .select(:project_id)
-      .join(:applied_tag, access_tag_id: :id)
-      .where(tagged_id: subject_id)
-      .where(Sequel.pg_jsonb_op(:subjects).has_key?(Sequel[:access_tag][:name]))
-
     dataset = DB.from { access_policy.as(:acl) }
-      .select(:tagged_id, :tagged_table, :subjects, :actions, :objects)
+      .select(Sequel[:applied_tag][:tagged_id], Sequel[:applied_tag][:tagged_table], :subjects, :actions, :objects)
       .cross_join(Sequel.pg_jsonb_op(Sequel[:acl][:body])["acls"].to_recordset.as(:items, [:subjects, :actions, :objects].map { |c| Sequel.lit("#{c} JSONB") }))
-      .join(:access_tag, project_id: Sequel[:acl][:project_id])
+      .join(:access_tag, project_id: Sequel[:acl][:project_id]) do
+        Sequel.pg_jsonb_op(:objects).has_key?(Sequel[:access_tag][:name])
+      end
       .join(:applied_tag, access_tag_id: :id)
-      .where(Sequel.pg_jsonb_op(:objects).has_key?(Sequel[:access_tag][:name]))
-      .where(Sequel[:acl][:project_id] => subject_dataset)
+      .join(:access_tag, {project_id: Sequel[:acl][:project_id]}, table_alias: :subject_access_tag) do
+        Sequel.pg_jsonb_op(:subjects).has_key?(Sequel[:subject_access_tag][:name])
+      end
+      .join(:applied_tag, {access_tag_id: :id, tagged_id: subject_id}, table_alias: :subject_applied_tag)
 
     if object_id
       begin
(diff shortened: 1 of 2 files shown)


furkansahin (approved):
I guess one dis/advantage of having a team of SQL brainers is that we use pure SQL just because it's easier :D

jeremyevans:
One thing I couldn't test in this commit is whether PostgreSQL was able to optimize the subquery to the same extent it was able to optimize the join.  What's the best way to test that?  Do we have a database similar to the production one that I could run some EXPLAIN ANALYZE on to compare the two approaches?  Is this something I should be worrying about currently?

jeremyevans:
Sorry to ask for another review for this.  I gave it some more thought, and couldn't convince myself that PostgreSQL would optimize the correlated subquery correctly.  The most recent switches to a pure-JOIN approach that is more similar to the original SQL query.  This approach should be easy to expand to authorize multiple subject access in a single query, using 2 joins per subject.  That's how I plan to handle authorization for personal access tokens (which must authorize both the account access and the token access).

fdr:
@jeremyevans I clicked the little spinny button next to furkan's face in the review section to change the review state for him. I'm not sure if he surveys a review query as I do (maybe he stays on top of all his email), but that's how I check [my queue](https://github.com/pulls?q=is%3Aopen+is%3Apr+review-requested%3Afdr+archived%3Afalse+user%3Aubicloud)

fdr (commented):
this transliteration is a bit brutal and obscures quite a bit IMO, but I can appreciate the increased compositional power of sequel objects, I just wonder if they are necessary at every internal level, rather than manipulating key queries in chunks. I kinda wonder if raw SQL + subquery wrapping would wind up nicer. But, all fine to me...I'll defer to @enescakir

furkansahin (approved):
Thank you!

lib/authorization.rb
```diff
@@ -38,10 +38,14 @@ def self.expand_actions(actions)
   def self.matched_policies(subject_id, actions = nil, object_id = nil)
     object_filter = if object_id
       begin
-        Sequel.lit("AND object_applied_tags.tagged_id = ?", UBID.parse(object_id).to_uuid)
+        ubid = UBID.parse(object_id)
       rescue UBIDParseError
-        Sequel.lit("AND object_applied_tags.tagged_id = ?", object_id)
+        # nothing
+      else
```
enescakir, line 60: I didn't know we could add an `else` clause to `rescue`. Cool!

lib/authorization.rb
```diff
@@ -58,14 +58,13 @@ def self.matched_policies(subject_id, actions = nil, object_id = nil)
 
     DB[<<~SQL, {subject_id: subject_id, actions_filter: actions_filter, object_filter: object_filter}].all
       SELECT object_applied_tags.tagged_id, object_applied_tags.tagged_table, subjects, actions, objects
-      FROM accounts AS subject
-        JOIN applied_tag AS subject_applied_tags ON subject.id = subject_applied_tags.tagged_id
+      FROM applied_tag AS subject_applied_tags
```
enescakir: Good improvements. The initial idea was using more resources as subject. API tokens are good example. Additionally, you can authorize virtual machines to access other resources on AWS, which is something we might consider doing in the future.

lib/authorization.rb
```diff
@@ -109,7 +113,7 @@ def authorized(subject_id, actions)
       # We need to determine table of id explicitly.
       # @opts is the hash of options for this dataset, and introduced at Sequel::Dataset.
       from = @opts[:from].first
-      where { {Sequel[from][:id] => Authorization.authorized_resources(subject_id, actions)} }
+      where { {Sequel[from][:id] => Authorization.authorized_resources_dataset(subject_id, actions)} }
```
enescakir, line 114: Less query count 🎉

lib/authorization.rb
```diff
@@ -18,7 +18,7 @@ def self.authorize(subject_id, actions, object_id)
   end
 
   def self.all_permissions(subject_id, object_id)
-    matched_policies(subject_id, nil, object_id).flat_map { _1[:actions] }
+    matched_policies_dataset(subject_id, nil, object_id).select_map(:actions).tap(&:flatten!)
```
enescakir, line 21: I didn't understand why `.tap(&:flatten!)` is used instead of simply `.flatten`

lib/authorization.rb
```diff
@@ -22,11 +22,11 @@ def self.all_permissions(subject_id, object_id)
   end
 
   def self.authorized_resources_dataset(subject_id, actions)
-    matched_policies_dataset(subject_id, actions).select(:tagged_id)
+    matched_policies_dataset(subject_id, actions).select(Sequel[:applied_tag][:tagged_id])
   end
 
   def self.authorized_resources(subject_id, actions)
-    matched_policies_dataset(subject_id, actions).select_map(:tagged_id)
+    matched_policies_dataset(subject_id, actions).select_map(Sequel[:applied_tag][:tagged_id])
```
enescakir, line 29: Can `authorized_resources` use `authorized_resources_datase` like `authorized_resources_dataset(subject_id, actions).all`?

enescakir (approved):
There are numerous improvements, but no changes required to the tests 😎  It's good to see that these modifications don't affect any high-level thing.

lib/authorization.rb
jeremyevans: `flatten` allocates an array, while `flatten!` modifies the receiver.  However, `flatten!` returns nil if there are no changes, so you can combine with to get the `flatten` behavior of always returning the array, but without the allocation.

lib/authorization.rb
jeremyevans: You want it to return an array of values, not a hash.  You could use `authorized_resources_dataset(subject_id, actions).map(:tagged_id)`, but this approach is more efficient as it avoids a dataset allocation, and `select_map` is better optimized than `map` in sequel_pg.

PR #2243, opened 2024-11-12
Add Prog to Create a new VM for a DNS Server

aamederen:
Add Prog to Create a new VM for a DNS Server

We didn't have any automation for creation of VMs to server DNS servers
and the process consisted of many manual error-prone steps.

In this change, we introduce SetupDnsServerVM prog, which makes it
straightforward to create fully-functional DNS servers from scratch
(which could be useful for development environments for now and prod in
the future) as well as adding VMs to existing DNS servers.

While working on this, I noticed that we don't have a config variable
for the project holding the DNS service resources we use in prod, so I
created that. You also need to populate that in the dev env. DnsServer
class itself doesn't hold that value (but DnsZone does) so that addition
could be another beginner-friendly task.

Usage is like this:
```ruby
# Get or create a DNS server
ds = DnsServer.first || DnsServer.create_with_id(name: "elephant")

# Create some zones if not exists
dz1 = DnsZone.create_with_id(
    project_id: Config.dns_service_project_id,
    name: "test1.ubicloud.com",
    last_purged_at: Time.now
)
dz2 = DnsZone.create_with_id(
    project_id: Config.dns_service_project_id,
    name: "test2.ubicloud.com",
    last_purged_at: Time.now
)

# Create corresponding DNS Zone Nexus Strands
Strand.create(prog: "DnsZone::DnsZoneNexus", label: "wait") { _1.id = dz1.id }
Strand.create(prog: "DnsZone::DnsZoneNexus", label: "wait") { _1.id = dz2.id }

# Associate the zones we have to the VM
dz1.add_dns_server DnsServer.first
dz2.add_dns_server DnsServer.first

# Create a few Records in each
["hebele", "hubele", "asdasd-asdasd-asd", "123123"].each do |r|
dz1.insert_record(record_name: "#{r}.#{dz1.name}", type: "A", ttl:10, data: "0.1.1.0")
dz2.insert_record(record_name: "#{r}.#{dz2.name}", type: "A", ttl:10, data: "0.1.1.0")
end

# Watch the VM getting created
st = Prog::DnsZone::SetupDnsServerVm.assemble(DnsServer.first.id)

# Validate after VM is created
`nslookup hebele.test1.ubicloud.com #{st.reload.vm.sshable.host}`

```

Note that concurrency between the VM creation, zone operations and
records are not handled in the best way. So it's advised to double-check
the results after the VM is created in prod environment.

We have a ton more to do for a proper DNS service lifecycle. A few
ideas:
- Health check for DNS servers and VMs
- Check-ups for consistency among the VMs of a DNS server
- Ability to add/remove DNS zones
- More validation about multiple DNS servers
- DnsServer belongs to a project

While working on this task, I thought that as we build more capabilities around DNS service, we will need the
purge logic to be called from various places, just to be sure
we can get the latest clean snapshot of the records.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -185,4 +185,7 @@ module Config
 
   # AI
   optional :inference_endpoint_service_project_id, string
+
+  # DNS
+  optional :dns_service_project_id, string
 end
--- a/model/dns_zone/dns_zone.rb
+++ b/model/dns_zone/dns_zone.rb
@@ -51,6 +51,40 @@ class DnsZone < Sequel::Model
     incr_refresh_dns_servers
   end
 
+  def purge_obsolete_records
+    DB.transaction do
+      # These are the records that are obsoleted by a another record with the
+      # same fields but newer date. We can delete them even if they are not
+      # seen yet.
+      obsoleted_records = records_dataset
+        .join(
+          records_dataset
+            .select_group(:dns_zone_id, :name, :type, :data)
+            .select_append { max(created_at).as(:latest_created_at) }
+            .as(:latest_dns_record),
+          [:dns_zone_id, :name, :type, :data]
+        )
+        .where { dns_record[:created_at] < latest_dns_record[:latest_created_at] }.all
+
+      # These are the tombstoned records, we can only delete them if they are
+      # seen by all DNS servers. We join with seen_dns_records_by_dns_servers
+      # and count the number of DNS servers to ensure that they are seen by all
+      # DNS servers.
+      dns_server_ids = dns_servers.map(&:id)
+      seen_tombstoned_records = records_dataset
+        .select_group(:id)
+        .join(:seen_dns_records_by_dns_servers, dns_record_id: :id, dns_server_id: dns_server_ids)
+        .where(tombstoned: true)
+        .having { count.function.* =~ dns_server_ids.count }.all
+
+      records_to_purge = obsoleted_records + seen_tombstoned_records
+      DB[:seen_dns_records_by_dns_servers].where(dns_record_id: records_to_purge.map(&:id).uniq).delete(force: true)
+      records_to_purge.uniq(&:id).map(&:destroy)
+
+      update(last_purged_at: Time.now)
+    end
+  end
+
   def add_dot_if_missing(record_name)
     (record_name[-1] == ".") ? record_name : record_name + "."
   end
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -139,9 +139,8 @@ class Sshable < Sequel::Model
   end
 
   def available?
-    cmd("true")
-    true
-  rescue
+    cmd("true") && true
+  rescue Net::SSH::Disconnect, Net::SSH::ConnectionTimeout, Errno::ECONNRESET, Errno::ECONNREFUSED, IOError
     false
   end
 
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -138,6 +138,13 @@ class Sshable < Sequel::Model
     Thread.current[:clover_ssh_cache]&.delete([host, unix_user])
   end
 
+  def available?
+    cmd("true")
+    true
+  rescue
+    false
+  end
+
   def self.reset_cache
     return [] unless (cache = Thread.current[:clover_ssh_cache])
 
--- a/prog/dns_zone/setup_dns_server_vm.rb
+++ b/prog/dns_zone/setup_dns_server_vm.rb
@@ -1,0 +1,144 @@
+# frozen_string_literal: true
+
+class Prog::DnsZone::SetupDnsServerVm < Prog::Base
+  subject_is :vm, :sshable
+
+  def self.assemble(dns_server_id, name: nil, vm_size: "standard-2", storage_size_gib: 30, location: "hetzner-fsn1")
+    unless (dns_server = DnsServer[dns_server_id])
+      fail "No existing Dns Server"
+    end
+
+    unless Project[Config.dns_service_project_id]
+      fail "No existing Project"
+    end
+
+    name ||= "#{dns_server.ubid}-#{SecureRandom.alphanumeric(8).downcase}"
+
+    DB.transaction do
+      vm_st = Prog::Vm::Nexus.assemble_with_sshable(
+        "ubi",
+        Config.dns_service_project_id,
+        location: location,
+        name: name,
+        size: vm_size,
+        storage_volumes: [
+          {encrypted: true, size_gib: storage_size_gib}
+        ],
+        boot_image: "ubuntu-jammy",
+        enable_ip4: true
+      )
+
+      Strand.create_with_id(prog: "DnsZone::SetupDnsServerVm", label: "start", stack: [{subject_id: vm_st.id, dns_server_id: dns_server_id}])
+    end
+  end
+
+  def ds
+    @ds ||= DnsServer[frame["dns_server_id"]]
+  end
+
+  label def start
+    nap 5 unless vm.strand.label == "wait"
+    register_deadline(nil, 15 * 60)
+    hop_prepare
+  end
+
+  label def prepare
+    sshable.cmd <<~SH
+sudo sed -i 's/#DNSStubListener=yes/DNSStubListener=no/' /etc/systemd/resolved.conf
+sudo sed -i ':a;N;$!ba;s/127.0.0.1 localhost\\n\\n#/127.0.0.1 localhost\\n127.0.0.1 #{vm.inhost_name}\\n\\n#/' /etc/hosts
+sudo ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
+sudo apt-get update
+sudo apt-get -y install apt-transport-https ca-certificates wget
+sudo wget -O /usr/share/keyrings/cznic-labs-pkg.gpg https://pkg.labs.nic.cz/gpg
+echo "deb [signed-by=/usr/share/keyrings/cznic-labs-pkg.gpg] https://pkg.labs.nic.cz/knot-dns jammy main" | sudo tee /etc/apt/sources.list.d/cznic-labs-knot-dns.list
+sudo apt-get update
+sudo systemctl reboot
+    SH
+
+    hop_setup_knot
+  end
+
+  label def setup_knot
+    nap 5 unless sshable.available?
+
+    sshable.cmd <<~SH
+sudo apt-get -y install knot
+echo "KNOTD_ARGS="-C /var/lib/knot/confdb"" | sudo tee -a /etc/default/knot
+    SH
+
+    knot_config = <<-CONF
+server:
+    rundir: "/run/knot"
+    user: "knot:knot"
+    listen: [ "0.0.0.0@53", "::@53" ]
+
+log:
+  - target: "syslog"
+    any: "info"
+
+database:
+    storage: "/var/lib/knot"
+
+acl:
+  - id: "allow_dynamic_updates"
+    address: "127.0.0.1/32"
+    action: "update"
+
+template:
+  - id: "default"
+    storage: "/var/lib/knot"
+    file: "%s.zone"
+    acl: "allow_dynamic_updates"
+    zonefile-sync: "60"
+    zonefile-load: "difference"
+    journal-content: "all"
+
+
+zone:
+  #{ds.dns_zones.map { |dz| "- domain: \"#{dz.name}.\"" }.join("\n  ")}
+    CONF
+
+    sshable.cmd("sudo tee /etc/knot/knot.conf > /dev/null", stdin: knot_config)
+
+    hop_sync_zones
+  end
+
+  label def sync_zones
+    nap 5 if ds.dns_zones.any?(&:refresh_dns_servers_set?)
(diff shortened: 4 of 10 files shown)


model/dns_zone/dns_zone.rb
```diff
@@ -52,6 +52,40 @@ def delete_record(record_name:, type: nil, data: nil)
     incr_refresh_dns_servers
   end
 
+  def purge
+    DB.transaction do
```
byucesoy, line 56: State machines already run in transaction. You don't need transaction block here.

config.rb
```diff
@@ -179,4 +179,7 @@ def self.test?
```
byucesoy, line 179: typo in the commit message; `ibi-cloud`.

model/sshable.rb
```diff
@@ -138,6 +138,13 @@ def invalidate_cache_entry
     Thread.current[:clover_ssh_cache]&.delete([host, unix_user])
   end
 
+  def available?
```
byucesoy, line 141: This feels like a weird way to check availability. It seems we use this to ensure the VM is up before running actual command through SSH. In that case, couldn't we just rescue the error from that command (which is what we do usually in similar scenarios)?

Since this falls into programming model bucket, I wonder what @fdr would suggest for this case.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -0,0 +1,137 @@
...
+    unless (dns_server = DnsServer[dns_server_id])
+      fail "No existing Dns Server"
+    end
+
+    project_id ||= (Config.production? || Config.dns_service_project_id) ? Config.dns_service_project_id : Project.order_by(:created_at).first.id
+    unless (project = Project[project_id])
+      fail "No existing Project"
+    end
+
+    name ||= dns_server.ubid.to_s + SecureRandom.alphanumeric(8).downcase
```
byucesoy, line 16: Maybe add a dash or underscore in between.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -0,0 +1,137 @@
...
+
+  label def sync_zones
+    nap 5 unless sshable.available?
+
+    nap 5 if ds.dns_zones.any?(&:refresh_dns_servers_set?)
+
+    ds.dns_zones.each do |dz|
+      zone_config = <<-CONF
+#{dz.name}.          3600    SOA     ns.#{dz.name}. #{dz.name}. 37 86400 7200 1209600 3600
+#{dz.name}.          3600    NS      elephant.ns.ubicloud.com.
```
byucesoy, line 105: `elephant.ns.ubicloud.com` part should come from dns_server's name.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -0,0 +1,137 @@
...
+
+  label def configure_knot
+    nap 5 unless vm.strand.label == "wait"
+
+    register_deadline(nil, 15 * 60)
+
+    sshable.cmd <<~SH
+sudo sed -i 's/#DNSStubListener=yes/DNSStubListener=no/' /etc/systemd/resolved.conf
+sudo sed -i ':a;N;$!ba;s/127.0.0.1 localhost\\n\\n#/127.0.0.1 localhost\\n127.0.0.1 #{vm.inhost_name}\\n\\n#/' /etc/hosts
+sudo ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
```
byucesoy, line 48: I think it would be better if we reboot after this, so knot would be installed cleaner without any errors in the systemd logs due to not being able to bind.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -0,0 +1,137 @@
...
+    @ds ||= DnsServer[frame["dns_server_id"]]
+  end
+
+  label def configure_knot
+    nap 5 unless vm.strand.label == "wait"
+
+    register_deadline(nil, 15 * 60)
+
+    sshable.cmd <<~SH
+sudo sed -i 's/#DNSStubListener=yes/DNSStubListener=no/' /etc/systemd/resolved.conf
```
byucesoy, line 46: AFAIR, we also need to specify alternative DNS resolvers such as Cloud9 or Cloudflare.

model/dns_zone/dns_zone.rb
```diff
@@ -52,6 +52,40 @@ def delete_record(record_name:, type: nil, data: nil)
     incr_refresh_dns_servers
   end
 
+  def purge
```
byucesoy, line 55: Maybe rename this to `purge_obsoleted_records`.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -0,0 +1,137 @@
...
+#{dz.name}.          3600    NS      elephant.ns.ubicloud.com.
+      CONF
+      sshable.cmd("sudo -u knot tee /var/lib/knot/#{dz.name}.zone > /dev/null", stdin: zone_config)
+    end
+
+    sshable.cmd "sudo systemctl restart knot"
+
+    ds.dns_zones.each(&:purge)
+
+    commands = ds.dns_zones.flat_map do |dz|
```
byucesoy, line 114: I don't think this is concurrency safe. What if something changes in the dns zone after we read the records?

byucesoy (commented):
Thanks a lot, this will simplify the development a lot.

model/dns_zone/dns_zone.rb
enescakir: Since it is in the model, we can run it through the model in `pry`. Therefore, having DB.transaction could be beneficial.

prog/dns_zone/setup_dns_server_vm.rb
enescakir: Underscore is not allowed in our names. Dash works

enescakir, line 16: nit: removes `.to_s`
```suggestion
    name ||= "#{dns_server.ubid}-#{SecureRandom.alphanumeric(8).downcase}"
```

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  def ds
+    @ds ||= DnsServer[frame["dns_server_id"]]
+  end
+
+  label def configure_knot
+    nap 5 unless vm.strand.label == "wait"
+
+    register_deadline(nil, 15 * 60)
```
enescakir, line 43: We generally use `start` as initial label. These two lines might extracted as `start`

config.rb
aamederen: That was intentional (FWIW ibi is used for Turkey, the animal in a local dialect)

model/sshable.rb
aamederen: Yeah, I was surprised with the lack of a general purpose `available?` method for sshable. In this change, since I do a reboot and wait for it to finish, I would need something like this anyway, so putting into the sshable felt better.

prog/dns_zone/setup_dns_server_vm.rb
aamederen: Yeah, it's not concurrency-safe, but I would prefer putting it as a future work. I've already added that to the backlog.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -0,0 +1,137 @@
...
+
+class Prog::DnsZone::SetupDnsServerVm < Prog::Base
+  subject_is :vm, :sshable
+
+  def self.assemble(dns_server_id, name: nil, project_id: nil, vm_size: "standard-2", storage_size_gib: 30, location: "hetzner-fsn1")
+    unless (dns_server = DnsServer[dns_server_id])
+      fail "No existing Dns Server"
+    end
+
+    project_id ||= (Config.production? || Config.dns_service_project_id) ? Config.dns_service_project_id : Project.order_by(:created_at).first.id
```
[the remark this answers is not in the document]
aamederen: Alright. I thought maybe this would make things easier for the dev envs, but fair assumption.

model/dns_zone/dns_zone.rb
aamederen: Exactly my thoughts

prog/dns_zone/setup_dns_server_vm.rb
aamederen: Good catch, thanks

prog/dns_zone/setup_dns_server_vm.rb
aamederen: Okay, made the change, although it cost a new state (or a side benefit?).

model/dns_zone/dns_zone.rb
aamederen: Renamed this and the state in the nexus

prog/dns_zone/setup_dns_server_vm.rb
aamederen: Thanks!

prog/dns_zone/setup_dns_server_vm.rb
aamederen: Good point, made the change

model/dns_zone/dns_zone.rb
byucesoy: Makes sense.

prog/dns_zone/setup_dns_server_vm.rb
byucesoy: Sorry, I meant Quad9. In any case, it seems this is not required. Normally after disabling systemd-resolved, DNS queries for external websites would stop working so we had to specify an external DNS server, but I just tested and it seems to work.

prog/dns_zone/setup_dns_server_vm.rb
byucesoy: For development, I think this is fine.

byucesoy (approved):
Apart from the new `available?` method, I think this looks good. For the `available?` method, I'm fine with whatever @fdr prefers.

model/sshable.rb
fdr: Not opposed in principle, but looking at how it is used here, I can see why it was not written earlier: why start by running the script you want rather than saying: "can I run my command?" first, then running it anyway? You always have to deal with a possible error for each cmd, why add another one? This reminds me of the unix classic, people that check: "stat" on a file before "unlink" (that is, `if [-f file] rm file`), only to then be subject to race conditions and required to handle exceptions from "unlink" anyway. The most common practice is to go ahead and try to do what you need to do and handle the exceptions once.

In theory I can imagine `available?` being useful if you have a number of complicated side effects outside/before the ssh, and you want to heuristically check whether you have a good chance at succeeding with the ssh command before going ahead. But in practice, given that every ssh command is liable to fail and needs idempotency handling, this seems to never quite happen.

model/sshable.rb
```diff
@@ -138,6 +138,13 @@ def invalidate_cache_entry
...
+    cmd("true")
+    true
+  rescue
```
fdr, line 144: also, blanket rescue is probably no what I'd do here, it can pick up a lot of stuff. 

We have a laundry list of ioerror-and-similar in host_nexus.rb:

`    rescue Net::SSH::Disconnect, Net::SSH::ConnectionTimeout, Errno::ECONNRESET, Errno::ECONNREFUSED, IOError`

So how about this: there's probably something here to make life a bit better, by encapsulating the "IOError and Friends" list, by using a block or possibly a flag to cmd. Maybe that can square the circle? The other reason I can think to do what you are doing is this `rescue` making the caller simpler.

aamederen: >The other reason I can think to do what you are doing is this rescue making the caller simpler.

That's correct, also I thought this was similar to `HostNexus#available`. 

I am taking a stab now to make it more elegant, but if I can't come up with an good solution to wrap "expected exceptions", I will revert this part and handle the exception in the caller.

model/sshable.rb
aamederen: In this case, we know the ssh command will fail, until the reboot finishes. There are 2 cases for it: in the host_nexus reboot and here. I think explicitly checking the availability in these cases makes the code a little bit more clear. 

about logistics: let me close this conversation and continue the discussion here: https://github.com/ubicloud/ubicloud/pull/2243#discussion_r1869512597

model/sshable.rb
aamederen: I made the rescue catch only the "IOerror and friends", also modified the code in hostnexus to separate availability check. Not sure if that's good enough, if not jut LMK and I will remove the available? function.

fdr (approved):
I think Burak is the authority here. I sort of don't like `available?` as a concept, since all this I/O can fail (it's a bit like `if File.exist? then File.unlink` the moment you hit the next command), but, I can appreciate one thing about it: it avoids the exponential backoff behavior in a common case, which may be case here. Rethinking this subtly is not a reason to block the patch though. I would just discourage use of `available?` except in very specific circumstances, it invites a kind of monotonic logic (if it works, the rest will work) that is false.

PR #2246, opened 2024-11-12
Fix two bugs in rake refresh_sequel_caches

jeremyevans:
Delete the existing cache files before rebuilding them.  Otherwise, new columns added to existing tables would not be picked up, because the model would use the cached schema instead of querying the database.

Always run using the test environment.  This rake task should not need to be run manually, as it is run automatically by the test_up and test_down tasks, so it should use the test environment.

This may result in `cache/schema.cache` showing as modified even if you made no changes.  I'm going to try to diagnose that issue in Sequel. Likely the schema is the same, but the marshalled version is different (different hash order seems likeliest).


jeremyevans:
Found the issue.  Ubicloud uses custom types, and oids for custom types are PostgreSQL cluster dependent.  Incorrect oids do not appear to cause any problems at present, but I think that's because the custom types we use are not typecast and left as strings.  I think if we were doing any custom types that needed to be typecast, this wouldn't work correctly.

I could work on support for caching without the custom type oids, with a single query to get the correct custom type oids at startup. Alternatively, we could just remove the caches and accept the startup hit.

jeremyevans:
> I could work on support for caching without the custom type oids, with a single query to get the correct custom type oids at startup. Alternatively, we could just remove the caches and accept the startup hit.

I went with the first option.

fdr (approved):

PR #2247, opened 2024-11-12
Disable ipv6 for runners

furkansahin:


fdr:
I would prefer to avoid something quite so extreme as eliminating an entire stack for docker quotas (which, either way, should definitely be in the commit message as motive)

Looking around, perhaps we can slice out the docker registry from AAAA:

> -A, --address=/<domain>[/<domain>...]/[<ipaddr>]
    Specify an IP address to return for any host in the given domains. A (or AAAA) queries in the domains are never forwarded and always replied to with the specified IP address which may be IPv4 or IPv6. To give multiple addresses or both IPv4 and IPv6 addresses for a domain, use repeated --address flags. Note that /etc/hosts and DHCP leases override this for individual names. A common use of this is to redirect the entire doubleclick.net domain to some friendly local web server to avoid banner ads. The domain specification works in the same way as for --server, with the additional facility that /#/ matches any domain. Thus --address=/#/1.2.3.4 will always return 1.2.3.4 for any query not answered from /etc/hosts or DHCP and not sent to an upstream nameserver by a more specific --server directive. As for --server, one or more domains with no address returns a no-such-domain answer, so --address=/example.com/ is equivalent to --server=/example.com/ and returns NXDOMAIN for example.com and all its subdomains. An address specified as '#' translates to the NULL address of 0.0.0.0 and its IPv6 equivalent of :: so --address=/example.com/# will return NULL addresses for example.com and its subdomains. This is partly syntactic sugar for --address=/example.com/0.0.0.0 and --address=/example.com/:: but is also more efficient than including both as separate configuration lines. Note that NULL addresses normally work in the same way as localhost, so beware that clients looking up these names are likely to end up talking to themselves.

fdr:
See this forum discussion as well. It may be necessary to tell it to upstream that domain for A records https://discourse.pi-hole.net/t/solved-disable-aaaa-response-for-a-given-domain/13143

fdr:
Maybe:

```
address=/registry.docker.com/::
```

fdr (approved):
Change the commit message, maybe

byucesoy (approved):

PR #2249, opened 2024-11-12
Use features in Roda 3.86 to simplify app

jeremyevans:
This uses the new response.skip_content_security_policy! method and conditional_sessions plugin to reduce the amount of custom code.

This is the same as #2227, but using a branch in the this repository instead of my forked repository, to see if that is the cause of the CI failures.


jeremyevans:
Looks like the #2227 CI failure was temporary, it eventually restarted by itself and passed.

PR #2255, opened 2024-11-13
Do not commit empty assets/css/app.css to repository

jeremyevans:
I think this is better than using
`git update-index --skip-worktree`.

This way, if a new user fails to generate the css file as described, they get an actual failure instead of a broken-looking page that has no CSS.

The only reason I think assets/css/app.css would have be committed empty previously is for the specs to run.  That's simple to fix by having the specs create the file if it doesn't exist, which is what this commit does.

Also, for some reason, there was an assets/css/.gitignore file that ignored app.css, even though it duplicated the entry in the main repo .gitignore.  Remove that file when readding the entry to the main .gitignore.


byucesoy (approved):

enescakir (approved):
> The only reason I think assets/css/app.css would have be committed empty previously is for the specs to run

Yes, you are write. We added it for solving this issue

PR #2264, opened 2024-11-14
Merge postgres routes

byucesoy:
This is continuation of the effort to merge different routes and inline `Routes::Common` modules. Earlier similar PRs are #2168, #2196, #2210, #2216


routes/helpers/postgres.rb
```diff
@@ -0,0 +1,81 @@
...
+      version: request_body_params["version"] || PostgresResource::DEFAULT_VERSION,
+      flavor: request_body_params["flavor"] || PostgresResource::Flavor::STANDARD
+    )
+    send_notification_mail_to_partners(st.subject, current_account.email)
+
+    if api?
+      Serializers::Postgres.serialize(st.subject, {detailed: true})
+    else
+      flash["notice"] = "'#{name}' will be ready in a few minutes"
+      request.redirect "#{@project.path}#{PostgresResource[st.id].path}"
```
jeremyevans, line 42: Is the PostgresResource retrievable from `st` without issuing an additional query? (I'm not sure)

routes/helpers/postgres.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+  end
+
+  def postgres_list
+    dataset = @project.postgres_resources_dataset.authorized(current_account.id, "Postgres:view").eager(:semaphores, :strand, :representative_server, :timeline)
```
jeremyevans, line 47: Only `:semaphores, :strand` were eagerly loaded for api first.  You can remove the `:representative_server, :timeline` here and add `dataset = dataset.eager(:representative_server, :timeline)` to the web branch.  Unless you have confirmed that representative server and timeline are currently lazily loaded in the api case?

jeremyevans (changes requested):
This looks really good.  I have a few requested changes, and few other questions/comments.

routes/helpers/postgres.rb
enescakir: I don't know a way to do it without an extra query, but we can access it through `st.subject.id`

byucesoy: I think it is not without changing the nexus to return the `PostgresResource` instead of `Strand`, which would have broader implications as many places expect it to return `Strand`. I can change this to `st.subject.path` if you think it is more readable, but it wouldn't help with the query count.

routes/merged/project/location/postgres.rb
```diff
@@ -0,0 +1,238 @@
...
+
+      r.delete true do
+        Authorization.authorize(current_account.id, "Postgres:delete", pg.id)
+        pg.incr_destroy
+        response.status = 204
+        ""
+      end
+
+      if web?
+        r.post "restart" do
```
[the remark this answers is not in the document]
byucesoy: I added 3 commits to apply this suggestion. Let me know if I missed any case.

jeremyevans (approved):
All changes look good!

PR #2267, opened 2024-11-15
Merge project routes

enescakir:
- **Merge the project api and web branches**
  

- **Move quota variable to the project detail branch.**
  We only use `@quota` in the `project/show` view, so there's no need for
  additional queries in all branches.
  

- **Perform same authorization for api and web requests in project list**
  The project is a special resource, distinct from others. When a user
  joins a project, we aim to display it in the project list, regardless of
  the user's permissions. The `Project:view` is only applied on the
  project detail page. While the web route has the correct query, the api
  route does not.
  

- **Combine unauthenticated api tests for project**
  

- **Fix rspec group name for project api**
  

- **Unify delete response for project**
  

- **Check account-project association with a query instead of Ruby**
  Account should be associated with the project to perform any operations
  on the project routes.
  
  We can use a database query instead of Ruby code to optimize
  performance.


spec/routes/api/project_spec.rb
```diff
@@ -2,28 +2,22 @@
...
-      post "/api/project"
-
-      expect(last_response).to have_api_error(401, "Please login to continue")
-    end
-
-    it "not delete" do
-      delete "api/project/#{project.ubid}"
-
-      expect(last_response).to have_api_error(401, "Please login to continue")
+    it "cannot perform authenticated operations" do
```
jeremyevans, line 11: I know we use this style in a lot of tests.  I don't think we should change it in this PR, but what are your thoughts on restructuring this and similar tests from:

```ruby
  it desc do
    array.each{|*args| do_something(*args)}
  end
```

to:

```ruby
  array.each do|*args| 
    it "#{desc} with #{args.inspect}" do
      do_something(*args)}
    end
  end
```

The proposed approached is slightly slower, but it has the following advantages:

* Makes each test case independent (separate transaction for each)
* Shows exactly which test cases fail (accelerates debugging if the failure only occurs in a subset of cases and not all cases)

routes/merged/project.rb
```diff
@@ -79,13 +79,8 @@ class Clover
 
         @project.soft_delete
 
-        if api?
-          response.status = 204
-          r.halt
-        else
-          flash["notice"] = "'#{@project.name}' project is deleted."
-          return {message: "'#{@project.name}' project is deleted."}
```
jeremyevans: I assume this request is submitted via javascript if we are returning 204 with an empty body. Does the UI still show an appropriate message in this case?

jeremyevans (approved):
Looks good! I have a question and general comment, but I think this can be merged as-is.

routes/merged/project.rb
enescakir: Yes, 204 is success for an ajax request

https://github.com/ubicloud/ubicloud/blob/ce499b0ff3f719d0cecc8ca65bff33171b88f3b3/assets/js/app.js#L79

spec/routes/api/project_spec.rb
enescakir: Several times, I enable cases individually to identify the ones that fail

PR #2269, opened 2024-11-16
WIP: IO Rate Limiting

pykello:


rhizome/host/lib/spdk_rpc.rb
```diff
@@ -93,6 +93,16 @@ def accel_crypto_key_destroy(name, if_exists = true)
...
+  def bdev_set_qos_limit(name, rw_ios_per_sec: nil, r_mbytes_per_sec: nil, w_mbytes_per_sec: nil)
+    params = {
+      name: name,
+      rw_ios_per_sec: rw_ios_per_sec || 0,
+      r_mbytes_per_sec: r_mbytes_per_sec || 0,
+      w_mbytes_per_sec: w_mbytes_per_sec || 0
+    }
+    call("bdev_set_qos_limit", params)
+  end
+
```
fdr, line 105:
```suggestion
  def bdev_set_qos_limit(name, rw_ios_per_sec: 0, r_mbytes_per_sec: 0, w_mbytes_per_sec: 0)
    params = {
      name:,
      rw_ios_per_sec:,
      r_mbytes_per_sec:,
      w_mbytes_per_sec:
    }
    call("bdev_set_qos_limit", params)
  end

```

fdr (approved):
nice e2e test there. @jeremyevans maybe you can take a look too?

prog/test/vm.rb
```diff
@@ -51,6 +53,61 @@ class Prog::Test::Vm < Prog::Test::Base
...
+      sudo fio --filename=./f --size=100M --direct=1 --rw=randrw --bs=4k --ioengine=libaio \\
+              --iodepth=256 --runtime=4 --numjobs=1 --time_based --group_reporting \\
+              --name=test-job --eta-newline=1 --output-format=json
+    CMD
+
+    fio_ios_output = JSON.parse(sshable.cmd(fio_ios_cmd))
+    fio_ios_output["jobs"].first["read"]["iops"] + fio_ios_output["jobs"].first["write"]["iops"]
+  end
+
+  def get_read_bw_mbytes
```
jeremyevans, line 70: I recommend having `get_read_bw_bytes` and `get_write_bw_bytes` (instead of mbytes).  Then having `verify_io_rates` do the calculation and comparison.  This issue with this approach is you are truncating down here, and comparing against a multiplication in `verify_io_rates`.

I'm guessing we don't care about exact numbers, since we are multiplying by `1.2`, but I think it is more flexible to expose the exact rates.  The truncation doesn't matter much for high rates (who cares between 473 and 474), but matters a lot for low rates (think about a value of 0, which could mean no IO, or anything less than 1MB).

We may currently only be using this to enforce maximum bandwidth limits, but if we later decide to use this for monitoring/logging, having it return the actual bytes is best.

jeremyevans (commented):
This isn't my area of familiarity, so please consider my comments as suggestions, and not educated ones at that.  Please let me know if you have questions.

rhizome/host/lib/spdk_rpc.rb
pykello: SPDK expects `0` for "unlimited", but I thought it is easier to understand if we use NULL for unlimited in our database records. So, I use NULL in our logic & just convert it properly when passing it to SPDK. I added a comment.

cache/schema.cache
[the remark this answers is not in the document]
fdr: you think so? that's probably fine, but also, the one after might suffice, because the moment the code goes out with the migration, it is not yet applied. The cache only becomes correct *after* application. As general rule, migrations have to be safe and latent in production, because a rollback across the migration will make sequel freak out about migration names in the database that are not in the file system.

jeremyevans (commented):
Looks good to me. The changes address all of my comments. However, someone with more knowledge of this area should approve.

migrate/20241120_add_storage_rate_limits.rb
macieksarnowicz, line 1: Does this need to be a separate PR?

pykello: Migrations don't need to be separate PRs. They need to be separate commits. We deploy commit by commit.

prog/test/vm.rb
```diff
@@ -24,7 +26,7 @@ class Prog::Test::Vm < Prog::Test::Base
   label def install_packages
     if /ubuntu|debian/.match?(vm.boot_image)
       sshable.cmd("sudo apt update")
-      sshable.cmd("sudo apt install -y build-essential")
+      sshable.cmd("sudo apt install -y build-essential fio")
     elsif vm.boot_image.start_with?("almalinux")
       sshable.cmd("sudo dnf check-update || [ $? -eq 100 ]")
       sshable.cmd("sudo dnf install -y gcc gcc-c++ make")
```
enescakir, line 32: Why don't we install `fio` on `almalinux`? Does it come by default?

pykello: No, I had forgotten. Thanks for reminding. Fixed.

fdr (approved):

PR #2272, opened 2024-11-18
Rewrite port for load balancer when balanced

furkansahin:
We were not rewriting the port when the load balancer redirects the connection to the neighbor node. This was not an issue with the initial implementation because the private networking was not impacted by firewalls. It is now. Therefore, we must rewrite the port in a firewall suitable way at the postrouting state. This commit adds that.


fdr (approved):
good catch

bsatzger (approved):

PR #2274, opened 2024-11-18
Update bin/regen_screenshots for documentation site changes

jeremyevans:
Additionally, make one screenshot generation idempotent by using a fixed id,
so it will not modify the screenshot if there has not been a visual change.


bin/regen_screenshots
```diff
@@ -146,8 +148,10 @@ class RegenScreenshots
     resize(1800)
     click_link "Compute"
     click_link "New Virtual Machine"
-    screenshot "quickstart/managed-services-2-screenshot.png"
+    screenshot "quick-start/managed-services-2-screenshot.png"
 
+# rubocop:disable Style/BlockComments
```
fdr, line 153: I endorse eliminating this cop. It's not a convenient rewrite one gets for free, even if it was, one may not want it, and block comments are not the kind of thing people do accidentally without a reason.

fdr:
general remark, although I've made some mistakes in this regard, binary names tend to use `-` separators in this project.

jeremyevans:
> general remark, although I've made some mistakes in this regard, binary names tend to use `-` separators in this project.

I'll add a commit to rename the binary.

bin/regen_screenshots
jeremyevans: I added a commit to eliminate the cop by default.

bin/regen_screenshots
```diff
@@ -103,7 +103,7 @@ class RegenScreenshots
     click_button "Verify Account"
 
     visit "/"
-    screenshot "quickstart/managed-services-1-screenshot.png"
+    screenshot "quick-start/managed-services-1-screenshot.png"
```
enescakir, line 106: Sorry for that. I forget to update here.

enescakir (approved):

PR #2275, opened 2024-11-18
Remove Routes::Common::Base class

enescakir:
Since we're merging the API and web routes, no one uses this class anymore.


jeremyevans (approved):
Looks good!

PR #2277, opened 2024-11-18
Final merging of api and web routes

jeremyevans:
This uses the same route namespace for both api and web routes.

This deletes the stub files previously in routes/api and routes/web, for routes that were already handled in routes/merged.

For the api-only or web-only routes that were still in routes/api and routes/web, move to routes/merged, and add appropriate guards. Unfortunately, review of these files is not easy with git, as git cannot ignore whitespace changes when doing rename detection. However, they are generally only indented, with an r.on web? block (the one api-specific route does not need a check in the route itself).

After the namespace change, Rubocop pointed out that the if/else branches in the main route had the same code, so this DRYs that up as well.

This fixes the dashboard and after-login branches in clover.rb to not accept trailing garbage, or request methods other than GET. Previously, a web request for /{dashboard,after_login}/some/crap/here would be treated the same as /{dashboard,after_login}.


routes/merged/github.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+class Clover
+  hash_branch("github") do |r|
+    r.on web? do
+      r.get "callback" do
```
enescakir, line 6: Roda also supports to combine matchers, right?

```suggestion
      r.get web?, "callback" do
```

jeremyevans: Yep.  I didn't check that this was the only route in the file.  I'll make this change.

enescakir (approved):
Should we also remove `merged` directory, and move route files to `routes` directory?

jeremyevans:
> Should we also remove `merged` directory, and move route files to `routes` directory?

I'm in favor of that, but it requires moving `clover_error.rb`, `helpers`, and `runtime`. I'll work on a pull request for that after merging this.    I'll plan to merge this early my time tomorrow.

byucesoy (approved):

PR #2278, opened 2024-11-18
Refactor load balancer post api

enescakir:
- **Refactor load balancer post api**
  We create/retrieve location-based resources using
  `project/<P>/location/<L>/vm/<NAME>` pattern [^1]. However, the load
  balancer deviates from this pattern. We retrieve the load balancer with
  `project/<P>/location/<L>/load-balancer/<NAME>`, but create it with
  `project/<P>/load-balancer/<NAME>`. This is because we get the location
  from the private subnet's location.  Also `attach-vm`/`detach-vm`
  endpoints have the location in the path.
  
  We should align it with other resources to maintain consistency for
  location-based resources. We use the location information from the path
  to validate the private subnet's location. It can't find private subnet
  if the locations don't match.
  
  [^1]: https://www.ubicloud.com/docs/api-reference/overview#creating-resources
  

- **Add load balancer post route for invalid names**
  If the resource name doesn't match the NAME_OR_UBID pattern, we should
  return an appropriate error. We have a similar route in other resources
  too.
  

- **Use TEST_LOCATION in load balancer API tests**
  It makes easier to keep locations same
  

- **Remove unnecessary load balancer test**
  It's already tested at
  spec/routes/api/project/location/load_balancer_spec.rb

### DIFF
--- a/spec/routes/api/project/load_balancer_spec.rb
+++ b/spec/routes/api/project/load_balancer_spec.rb
@@ -7,14 +7,6 @@ RSpec.describe Clover, "vm" do
 
   let(:project) { user.create_project_with_default_policy("project-1") }
 
-  describe "unauthenticated" do
-    it "not list" do
-      get "/api/project/#{project.ubid}/load-balancer"
-
-      expect(last_response).to have_api_error(401, "Please login to continue")
-    end
-  end
-
   describe "authenticated" do
     before do
       login_api(user.email)


enescakir:
@geemus What do you think about 828fd359ec4cd5c74c53ba6fda403f7b6e930198?

routes/merged/project/load_balancer.rb
```diff
@@ -6,21 +6,17 @@ class Clover
       load_balancer_list
     end
 
-    r.on api? do
-      r.post String do |lb_name|
-        load_balancer_post(lb_name)
+    r.on web? do
+      r.post do
```
jeremyevans, line 10:
```suggestion
      r.post true do
```

Otherwise the route would accept arbitrary garbage in the remaining path.

jeremyevans (changes requested):
Looks good!  My understanding is we haven't promised backwards compatibility for the API yet, so this should be fine after the one requested change.

fdr:
Well, approval modulo @jeremyevans 's spotted bug

enescakir:
> Looks good! My understanding is we haven't promised backwards compatibility for the API yet, so this should be fine after the one requested change.

Since it's not included in our API documentation, I think we haven't made any specific promises to anyone.
https://www.ubicloud.com/docs/api-reference/overview

geemus:
LGTM. Yeah, since you haven't explicitly promised anything yet, breaking changes are probably fine for the time being at least. Though it might be good to explicitly say that nothing is promised to be clear.

jeremyevans (approved):

PR #2279, opened 2024-11-18
Do not swallow unexpected errors when testing

jeremyevans:
Doing so complicates debugging.  Only swallow the following errors during tests:

* Sequel::ValidationFailed
* DependencyError
* Validation::ValidationFailed
* CloverError
* Authorization::Unauthorized

These are the errors that the test suite expects to be handled by the error handler.

Raise all other errors when testing, to make debugging easier.


clover.rb
```diff
@@ -99,6 +99,17 @@ def assets
     if error[:code] == 204
       # nothing
     elsif api? || runtime? || request.headers["Accept"] == "application/json"
+      # :nocov:
+      if Config.test?
+        case e
+        when Sequel::ValidationFailed, DependencyError, Validation::ValidationFailed, CloverError, Authorization::Unauthorized
```
enescakir, line 105: `DependencyError`, `Validation::ValidationFailed`, and `Authorization::Unauthorized` are all subclasses of `CloverError`. Does the `when` keyword match with the parent class?

enescakir:
Actually, the `else` case in the `parse_error` function within the `general.rb` file has a similar condition. Couldn't we add this raising there, instead of adding it in two different places, similar to this PR?

```ruby
  def parse_error(e)
    case e
    when Sequel::ValidationFailed
      code = 400
      type = "InvalidRequest"
      message = e.to_s
    when CloverError
      code = e.code
      type = e.type
      message = e.message
      details = e.details
    else
      Clog.emit("route exception") { Util.exception_to_hash(e) }

      #### START NEW LINES
      # :nocov:
      if Config.test?
        raise e if Config.test? && e.message != "test error"
      end
      # :nocov:
      #### END NEW LINES

      code = 500
      type = "UnexceptedError"
      message = "Sorry, we couldn’t process your request because of an unexpected error."
    end

    response.status = code

    {
      code: code,
      type: type,
      message: message,
      details: details
    }
  end
```

jeremyevans:
OK, I push a commit with all changes discussed.  I think it looks much nicer than before, thanks to inlining `parse_error`.

PR #2284, opened 2024-11-19
Add Clover instance methods related to authorization

jeremyevans:
These have the following advantages:

* DRYs up code in all of the routes
* Improves performance by not loading current_account unnecessarily
  (if only the id is needed, no reason to load the Account from the 
  database)
* Will allow transparent support for personal access token
  authorization when that is added.

Submitting as a draft, as this depends on #2277.  Will rebase and switch to non-draft after #2277 is merged.

### DIFF
--- a/routes/helpers/firewall.rb
+++ b/routes/helpers/firewall.rb
@@ -20,7 +20,7 @@ class Clover
   end
 
   def firewall_post(firewall_name)
-    Authorization.authorize(current_account.id, "Firewall:create", @project.id)
+    authorize("Firewall:create", @project.id)
     Validation.validate_name(firewall_name)
 
     optional_parameters = %w[description]
--- a/routes/helpers/firewall.rb
+++ b/routes/helpers/firewall.rb
@@ -2,7 +2,7 @@
 
 class Clover
   def firewall_list_dataset
-    @project.firewalls_dataset.authorized(current_account.id, "Firewall:view")
+    dataset_authorize(@project.firewalls_dataset, "Firewall:view")
   end
 
   def firewall_list_api_response(dataset)
--- a/routes/helpers/general.rb
+++ b/routes/helpers/general.rb
@@ -72,6 +72,10 @@ class Clover < Roda
     @is_web = !api? && !runtime?
   end
 
+  def authorize(actions, object_id)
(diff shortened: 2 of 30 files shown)


routes/helpers/firewall.rb
```diff
@@ -2,7 +2,7 @@
 
 class Clover
   def firewall_list_dataset
-    @project.firewalls_dataset.authorized(current_account.id, "Firewall:view")
+    dataset_authorize(@project.firewalls_dataset, "Firewall:view")
```
enescakir, line 5: I actually liked using chaining. Wrapping the entire dataset with another method like `dataset_authorize` doesn't seem quite intuitive. Can't we somehow support `rodauth.session_value` improvement with chaining?

jeremyevans: Would could do something like `@project.firewalls_dataset.with_account_id(current_account_id).authorized("Firewall:view")`.

Would you prefer that approach?

enescakir: If we pass the account id as `.with_account_id(current_account_id)`, we can also keep it in the current signature `.authorized(current_account_id, "Firewall:view")`. I think the issue is that we can't access `current_account_id` (`rodauth.session_value`) in the Dataset module. Therefore, it appears we need to pass it for chaining regardless.

jeremyevans: Correct.  I don't think the `.with_account_id(current_account_id)` chaining approach will work with the planned changes to support personal access tokens transparently.  Maybe we could do `@project.firewalls_dataset.authorized(self, "Firewall:view")`, and have the Dataset#authorize code deal with the Roda scope?  That looks nicer, but feels like a layer violation to me.

fdr (commented):
leaving my non-review review, as in, I glanced at it, and all the proposed solutions seem within a reasonable design envelope that I cannot assess better than @enescakir and @jeremyevans

enescakir (approved):

routes/helpers/firewall.rb
enescakir: Yeah, it mixups the layers. We can revisit it in the future if the new method becomes bothering.

PR #2285, opened 2024-11-19
Make inspect values use ubids instead of uuids

jeremyevans:
I think this makes the inspect output more helpful for models that use ResourceMethods.  We already display the ubid next to the main model name, but this makes the foreign key values also ubids.

Additionally, since the id is already included in ubid form, there is no reason to include it in the values, so this removes it from the values.

This also removes the allocation of many intermediate arrays (2 per column plus one for the map) for each inspect call.


spec/model/private_subnet_spec.rb
```diff
@@ -78,8 +78,8 @@
   describe "#inspect" do
     it "includes ubid if id is available" do
       ubid = described_class.generate_ubid
-      uuid = private_subnet.id = ubid.to_uuid.to_s
-      expect(private_subnet.inspect).to eq "#<PrivateSubnet[\"#{ubid}\"] @values={:net6=>\"fd1b:9793:dcef:cd0a::/64\", :net4=>\"10.9.39.0/26\", :location=>\"hetzner-fsn1\", :state=>\"waiting\", :name=>\"ps\", :id=>\"#{uuid}\"}>"
+      private_subnet.id = ubid.to_uuid.to_s
+      expect(private_subnet.inspect).to eq "#<PrivateSubnet[\"#{ubid}\"] @values={:net6=>\"fd1b:9793:dcef:cd0a::/64\", :net4=>\"10.9.39.0/26\", :location=>\"hetzner-fsn1\", :state=>\"waiting\", :name=>\"ps\"}>"
```
enescakir, line 82: We can also improve this test by adding a foreign key as ubid

enescakir (approved):
We've talked about it a lot, but we just couldn't pull it off as smoothly. Thanks for handling it elegantly.

enescakir:
It's not in the scope of this PR, we may also consider removing milliseconds and timezone details from dates to reduce noise in the output.

```ruby
:created_at=>2024-11-18 23:31:30.891319 +0000
```

spec/model/private_subnet_spec.rb
jeremyevans: I'll work on a PR today that adds a test for a foreign key as a ubid, as well as removing subseconds and timezone from timestamptz values.

PR #2286, opened 2024-11-19
Proof of concept that access_tag.hyper_tag_table is not needed

jeremyevans:
As the uuids encode type information, you can determine the table from the uuid, so the hyper_tag_table column is redundant.

To get the 10-bit type information from the uuid in the database:

* Turn it into a string
* Take characters 16-18 and convert the hex value to integer, modulo 0x400

To get the 10-bit type information from an existing uuid integer:

* Take bits 64-73 (UBID#type_number, though it is commented out)

To get the same 10-bit type information for a class:

* Get base32-encoded 2 byte type prefix from TYPE2CLASSNAME, convert to integer (UBID.type_number)

The hyper_tag_table column appears to be used only in a single place, and this converts that usage to using the hyper_tag_id value instead.

The same is true of applied_tag.tagged_table, as that is not even used (it is returned in an authorization query, but all specs pass when it is removed from the query).

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -62,7 +62,9 @@ class Project < Sequel::Model
   end
 
   def has_resources
-    access_tags_dataset.exclude(hyper_tag_table: [Account.table_name.to_s, Project.table_name.to_s, AccessTag.table_name.to_s]).count > 0 || github_installations.flat_map(&:runners).count > 0
+    hyper_tag_type_number = (Sequel.join(["0x", Sequel.function(:substr, Sequel[:hyper_tag_id].cast(String), 16, 3)]).cast(Float).cast_numeric % 0x400)
+    allowed_hyper_tag_type_numbers = [Account, Project, AccessTag].map { UBID.type_number(_1) }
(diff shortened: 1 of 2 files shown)


enescakir:
Yes, we used `hyper_tag_table` back when we didn't have UBIDs. We couldn't determine class from UUIDs these days, but we do have the option to remove it now

fdr (commented):
promising, I do wonder if we ever had IDs that are external that do not identify their class (e.g. the ETC type).

Even if it happens, I'm sure we can figure something out for that special case. I'll defer to @enescakir . Nice to eliminate something from the join planning.

jeremyevans:
This PR is no longer relevant as we dropped the columns in #2638 (which also dropped the applied_tag table, which had a similar column).  Note that we've added other polymorphic ubid foreign key columns recently without related table columns.  I think all of our existing polymorphic ubid foreign key columns no longer have an associated `*_table` column.  So we've already incorporated the idea in this PR.

While I haven't started work on it, one of my thoughts is creating triggers to emulate polymorphic ubid foreign keys, so we can still benefit from referential integrity for these columns (though it may not be as good as the built-in support).

PR #2289, opened 2024-11-19
Support Ubuntu 24.04 hosts

pykello:
The only blocker was that SPDK binaries on Ubuntu 22.04 didn't work on Ubuntu 24.04. This PR fixes that by downloading & using OS version specific binaries.

To do this, we add an `os_version` field to `vm_host`, learning it automatically when setting up the host, and use it when installing SPDK.

This PR also changes E2E tests to use Ubuntu 24.04.


fdr:
Is kernel_version too much? It's not really used for much, it's correlated to the os_version unless we have a very concrete and immediate need

pykello:
> Is kernel_version too much? It's not really used for much, it's correlated to the os_version unless we have a very concrete and immediate need

@fdr I removed `kernel_version`.

PR #2290, opened 2024-11-19
Adds connected subnets for Postgres

furkansahin:
Need to work on this more but here is the overall idea


byucesoy:
I like the overall idea, but I'll probably not have time to review this for a while. Is it OK if this waits for a while?

furkansahin:
I guess you don't have time :D

PR #2291, opened 2024-11-19
Destroy expired load balancer certificates

furkansahin:
The load balancer creates a new certificate when the time comes and distributes it to the metadata-endpoint instances but it misses to destroy the old certificates. This PR adds that.

### DIFF
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -104,6 +104,9 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
     reap
     if strand.children.select { _1.prog == "Vnet::CertServer" }.all? { _1.exitval == "certificate is reshared" } || strand.children.empty?
       decr_refresh_cert
+      load_balancer.certs_dataset.exclude(id: load_balancer.active_cert.id).all do |cert|
+        CertsLoadBalancers[cert_id: cert.id].destroy
+      end
       hop_wait
     end
 
--- a/spec/prog/vnet/load_balancer_nexus_spec.rb
+++ b/spec/prog/vnet/load_balancer_nexus_spec.rb
@@ -148,9 +148,21 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
       expect { nx.wait_cert_broadcast }.to nap(1)
     end
 
-    it "hops to wait if all children are done" do
+    it "hops to wait if all children are done and no certs to remove" do
       expect(nx).to receive(:reap)
       expect { nx.wait_cert_broadcast }.to hop("wait")
+    end
+
+    it "removes certs if all children are done and there are certs to remove" do
+      cert_to_remove = Prog::Vnet::CertNexus.assemble("cert-to-remove", dns_zone.id).subject
+      cert_to_remove.update(created_at: Time.now - 60 * 60 * 24 * 30 * 4)
+      nx.load_balancer.add_cert(cert_to_remove)
+
+      expect(nx).to receive(:reap)
+
+      expect { nx.wait_cert_broadcast }.to hop("wait")
+      expect(Semaphore[name: "destroy", strand_id: cert_to_remove.id]).not_to be_nil
+      expect(nx.load_balancer.reload.certs.count).to eq 1
     end
   end
 
--- a/prog/vnet/cert_nexus.rb
+++ b/prog/vnet/cert_nexus.rb
@@ -113,6 +113,14 @@ class Prog::Vnet::CertNexus < Prog::Base
       acme_client.revoke(certificate: cert.cert, reason: REVOKE_REASON) if cert.cert
     rescue Acme::Client::Error::AlreadyRevoked => ex
       Clog.emit("Certificate is already revoked") { {cert_revoke_failure: {ubid: cert.ubid, exception: Util.exception_to_hash(ex)}} }
+    rescue Acme::Client::Error::NotFound => ex
+      Clog.emit("Certificate is not found") { {cert_revoke_failure: {ubid: cert.ubid, exception: Util.exception_to_hash(ex)}} }
+    rescue Acme::Client::Error::Unauthorized => ex
+      if ex.message.include?("The certificate has expired and cannot be revoked")
+        Clog.emit("Certificate is expired and cannot be revoked") { {cert_revoke_failure: {ubid: cert.ubid, exception: Util.exception_to_hash(ex)}} }
+      else
+        raise ex
+      end
     end
 
     dns_zone.delete_record(record_name: dns_record_name) if dns_challenge
--- a/spec/prog/vnet/cert_nexus_spec.rb
+++ b/spec/prog/vnet/cert_nexus_spec.rb
@@ -238,6 +238,45 @@ RSpec.describe Prog::Vnet::CertNexus do
 
       expect { nx.destroy }.to exit({"msg" => "certificate revoked and destroyed"})
     end
+
+    it "emits a log and continues if the cert is not found" do
+      client = instance_double(Acme::Client)
+      expect(cert).to receive(:cert).and_return("test-cert").at_least(:once)
+      expect(nx).to receive(:acme_client).and_return(client)
+      expect(client).to receive(:revoke).and_raise(Acme::Client::Error::NotFound.new("not found"))
+
+      expect(Clog).to receive(:emit).with("Certificate is not found")
+      expect(nx).to receive(:dns_zone).and_return(dns_zone)
+      expect(dns_zone).to receive(:delete_record).with(record_name: "test-record-name.cert-hostname")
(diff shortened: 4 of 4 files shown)


prog/vnet/load_balancer_nexus.rb
```diff
@@ -104,6 +104,10 @@ def need_to_rewrite_dns_records?
     reap
     if strand.children.select { _1.prog == "Vnet::CertServer" }.all? { _1.exitval == "certificate is reshared" } || strand.children.empty?
       decr_refresh_cert
+      load_balancer.certs.reject { _1.id == load_balancer.active_cert.id }.each do |cert|
```
fdr, line 107: ? 
```suggestion
      load_balancer.certs_dataset.exclude(id: load_balancer.active_cert.id).each do |cert|
```

fdr (changes requested):
....that schema cache file is suspicious, since you have no migrations here. And also, can perhaps push down a thing, though I guess it only excludes a single record. Arguably, it comes out nicer: more characters but simpler subexpressions.

prog/vnet/load_balancer_nexus.rb
jeremyevans, line 107:
```suggestion
      load_balancer.certs_dataset.exclude(id: load_balancer.active_cert.id).all do |cert|
```

Sequel does not technically support queries inside a `Dataset#each` block.  It happens to work with the postgres adapter, unless you try using streaming, but it's best not to rely on that.

enescakir:
> that schema cache file is suspicious, since you have no migrations here

I suspect that upgrading Postgres from 15.4 to 15.8 caused this

jeremyevans:
> ....that schema cache file is suspicious, since you have no migrations here. And also, can perhaps push down a thing, though I guess it only excludes a single record. Arguably, it comes out nicer: more characters but simpler subexpressions.

Without migrations, the schema cache should not change, after the changes made in 0a6038e3eb849d2cac7e89729b17728ce0b89cc4 .  @furkansahin can you run refresh_schema_caches and see if you get the same cache as the one you committed here (or at least, a different one than in main)?  If so, please send it to me and I can compare it to the current cache and see what the differences are and try to figure out what might have caused it.

jeremyevans:
> > that schema cache file is suspicious, since you have no migrations here
> 
> I suspect that upgrading Postgres from 15.4 to 15.8 caused this. I had this issue before.

I didn't consider that, but it sounds reasonable.  I should be able to compare the schema cache in this PR to the main branch and see what the differences are.

jeremyevans:
> > > that schema cache file is suspicious, since you have no migrations here
> > 
> > 
> > I suspect that upgrading Postgres from 15.4 to 15.8 caused this. I had this issue before.
> 
> I didn't consider that, but it sounds reasonable. I should be able to compare the schema cache in this PR to the main branch and see what the differences are.

The schema cache in this PR matches the one in the main branch.  The one from the parent commit is different.  This just needs to be rebased after ea23875f72e6cf255e4ec3775fa8aba596700c19, the last change to `cache/schema.cache` in the main branch.

prog/vnet/load_balancer_nexus.rb
fdr: does it crash?

jeremyevans: On the postgres adapter, it does not by default.  This is because if you aren't using streaming, libpq will buffer all results in memory before iterating.  Many other Sequel adapters use a streaming approach by default, in which case, it can raise an exception (possibly crash depending on how the driver is programmed).  I consider it a bit habit to rely on this quirk of the postgres adapter, though.

furkansahin: Thanks for the help here :) I have updated the PR.

fdr: @jeremyevans yeah, but does it crash if doing the wrong thing? curious if I would have learned it in a graceful manner if not for the fact that I've been in libpq world for so long that it didn't wind up mattering

fdr (approved):

prog/vnet/load_balancer_nexus.rb
jeremyevans: With the way the postgres adapter works with libpq, it should not crash.  However, if we used sequel_pg's streaming support, or tried to support another database adapter, it could crash (or at least raise an exception).

Sequel::Dataset#each's documentation has a general warning against making queries inside an each block: https://sequel.jeremyevans.net/rdoc/classes/Sequel/Dataset.html#method-i-each

prog/vnet/load_balancer_nexus.rb
```diff
@@ -104,6 +104,10 @@ def need_to_rewrite_dns_records?
     reap
     if strand.children.select { _1.prog == "Vnet::CertServer" }.all? { _1.exitval == "certificate is reshared" } || strand.children.empty?
       decr_refresh_cert
+      load_balancer.certs_dataset.exclude(id: load_balancer.active_cert.id).all do |cert|
+        cert.incr_destroy
+        CertsLoadBalancers[cert_id: cert.id].destroy
```
jeremyevans, line 109: `CertsLoadBalancers#destroy` already calls `cert.incr_destroy`.  This results in `cert.incr_destroy` being called twice.  Is that what we want?

Additionally, `CertsLoadBalancers[cert_id: cert.id]` could return an entry for a different load balancer, if the certificate is associated to multiple load balancers.  This should probably be `CertsLoadBalancers[cert_id: cert.id, load_balancer_id: load_balancer.id]` instead.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -148,10 +148,22 @@
       expect { nx.wait_cert_broadcast }.to nap(1)
     end
 
-    it "hops to wait if all children are done" do
+    it "hops to wait if all children are done and no certs to remove" do
       expect(nx).to receive(:reap)
       expect { nx.wait_cert_broadcast }.to hop("wait")
     end
+
+    it "removes certs if all children are done and there are certs to remove" do
```
jeremyevans, line 156: I recommend a test where the certificate is associated with multiple load balancers, to ensure it is only removed from the expected load balancer.

jeremyevans (changes requested):
I don't have good knowledge of the domain here, but based on my guessing, I think this needs some changes.  I could be off-base, though.

prog/vnet/load_balancer_nexus.rb
fdr: IIRC there is no such thing as a shared certificate at present since they're keyed to the load balancer's DNS name, which is essential to the whole thing. But it may not hurt to program defensively.

jeremyevans: Interesting.  Looks like I misunderstood the purpose of `CertsLoadBalancers`?  It looks like a join table between certs and load balancers.  However, I see now that we have `Cert.one_through_one :load_balancer` and not many_to_many .

I guess its done this way because you can have multiple certificates to a load balancer, but not the other way around, and you don't want to have the load_balancer_id in the certificates table because not all certs are tied to a load balancer?

Is the double `cert.incr_destroy` still an issue?

fdr: It may be wasteful and unnecessary (depending if the other site for incr_destroy has other entrants), but it's not harmful. But if it's possible to eliminate one and be less confusing, that would be better.

What is harmful is an incr in a loop, because it generates one tuple per iteration

fdr: You can see the older system (now much older, I suppose) here: https://www.citusdata.com/blog/2016/08/12/state-machines-to-run-databases/

The current record-per-incr is effectively a unary pivot of what used to be an integer, to better be able to attach data to each increment, and to have finer grained locking data. But, it blows up the amount of state per increment by billions of times (not an exaggeration, considering previously, tweaking a bit), so it's definitely pushing an envelope of some kind.

spec/prog/vnet/load_balancer_nexus_spec.rb
furkansahin: That's not really a valid scenario, a cert can only be associated with a single load balancer.

prog/vnet/load_balancer_nexus.rb
furkansahin: The main reason CertsLoadBalancers table exist is that we would like to use Certs in future for other products as well, such as Postgres. In such a case, we would need to add a new columnt to the Cert table for every new service. So, I chose to implement a join table

jeremyevans: Even though it's not harmful, it seems best to remove the `cert.incr_destroy` in this code, since it's in `CertsLoadBalancers#destroy`.  However, as @fdr says it isn't harmful, it doesn't block approval.

jeremyevans (approved):

prog/vnet/load_balancer_nexus.rb
furkansahin: Oh, I have removed it already

enescakir (approved):

PR #2294, opened 2024-11-19
Exclude more lines from runner script logs

enescakir:
When a job fails, we log the runner script's output for any necessary further investigation. We have excluded the JIT token line and sudo lines.

But other logs include outputs from subprocesses like php, sudo, etc., which could contain sensitive data. `run-withenv` is the main process, while `systemd` lines provide valuable insights into the lifecycle of the runner script, including OOM issues. We exclude the 'Started' line to avoid exposing the JIT token.

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -351,8 +351,13 @@ class Prog::Vm::GithubRunner < Prog::Base
           serial_log_path = "/vm/#{vm.inhost_name}/serial.log"
           vm.vm_host.sshable.cmd("sudo ln #{serial_log_path} /var/log/ubicloud/serials/#{github_runner.ubid}_serial.log")
 
-          # Exclude the "Started" line because it contains sensitive information.
-          vm.sshable.cmd("journalctl -u runner-script --no-pager | grep -v -e Started -e sudo")
+          # We grep only the lines related to 'run-withenv' and 'systemd'. Other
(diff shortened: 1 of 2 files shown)


prog/vm/github_runner.rb
```diff
@@ -351,8 +351,13 @@ def setup_info
...
 
-          # Exclude the "Started" line because it contains sensitive information.
-          vm.sshable.cmd("journalctl -u runner-script --no-pager | grep -v -e Started -e sudo")
+          # We grep only the lines related to 'run-withenv' and 'systemd'. Other
+          # logs include outputs from subprocesses like php, sudo, etc., which
+          # could contain sensitive data. 'run-withenv' is the main process,
+          # while systemd lines provide valuable insights into the lifecycle of
+          # the runner script, including OOM issues.
+          # We exclude the 'Started' line to avoid exposing the JIT token.
+          vm.sshable.cmd("journalctl -u runner-script --no-pager | grep -e run-withenv -e systemd | grep -v -e Started")
```
fdr, line 360: double check this, but:

```suggestion
          vm.sshable.cmd("journalctl -u runner-script --no-pager -g 'run-withenv|systemd' | grep -Fv Started")
```

fdr (approved):
Maybe my proposal is slightly more efficient, if it does what you need.

velioglu (approved):

fdr:
`-t` is even better

fdr (approved):

PR #2295, opened 2024-11-19
Pykello/slice

macieksarnowicz:
DRAFT - not to be checked in'

Adding ResourceGroup object to track systemctl slice and control resources on the host

Adding a new model + prog for VmHostSlice. A slice controls the allocation of CPUs for now, and it can be expanded in the future to control other resources. When created in Clover and associated with a VmHost it creates a SystemCtl slice object (and underlying cgroup) on the host. It takes as an argument a set of CPUs to be affinitized to the group. The format we use is the same as used by the OS, with a ranges of CPUs separated by commas, for example: "2-3,6-7" or "4,6-7", etc. The Prog:Vm:VmHostSlice controls the lifecycle of a host-side slice, with prep, wait, and destroy states. Upon host restart, we perform the necessary steps to also restart the slice.
To create a ResourceGroup:
Prog::Vm::VmHostSlice.assemble_with_host("standard123", VmHost.first, allowed_cpus: "2-3", memory_1g: 8)

VmHost can list all available slices.

Added tests with 100% coverage.


macieksarnowicz:

PR #2296, opened 2024-11-19
Make cloudify_server work correctly with Hetzner servers

macieksarnowicz:
The prior version of this script did not pass `hetzner_server_identifier:` to the `Prog::Vm::HostNexus::assemble` method, which did not register the server as a Hetzner one. This resulted in not correctly recognizing IPv4 subnets attached to the server and making those addresses available for the VMs. This patch fixes this problem by asking a user to provide an optional server ID, which is passed to the assembly method.

I am also adding default_boot_image so the server is ready to accept VMs after this script completes.


macieksarnowicz:

pykello (approved):

PR #2298, opened 2024-11-20
Integrate Load Balancer Health Probes into monitoring

furkansahin:
## Prepare load_balancers_vms models for health probes
We are moving the health checks into load_balancers_vms. The main reason
these are not handled as part of VM monitoring is because the VM might
be up and running but the load balancer makes a decision based on the
application state. It is safer to keep this separate. This code will
also interract with load balancer and keeping a code path as part of a
VM that is not part of a load balancer doesn't seem clever. We are
making a couple of logic changes here.
1. The state of the load balancer node is only updated when the monitor
is sure their state is changed. This way, we don't have to look into
counters at the time of update.
2. IPv4 and IPv6 availability is handled properly. Before, these were
kind of mixed up and causing some logical errors. Now, they are
distinctly checked and properly declared.

This commit itself doesn't simply stop the health probe strands. We will
need to handle their migration manually by destroying the strands.

## Switch using monitor and remove health probes prog
With this commit, we are removing the prog and completely switching to
monitor for health probes.

## Inference endpoint doesn't use the counter anymore
We are removing the logic that checks the counter in inverence endpoint
availability check. The main reason is, now, if the state of the
load_balancers_vms is switched to "down" we are already sure that the
counter hit to the threshold.

## Remove state_counter and update keys in load_balancers_vms
The primary key is changed to id, state_counter is not needed and id
cannot be null anymore.


furkansahin:
Instead of spending time on https://github.com/ubicloud/ubicloud/pull/2658, I wanted to move this.

byucesoy (approved):
Looks good IMO apart from few minor comments.

model/load_balancers_vms.rb
```diff
@@ -3,16 +3,66 @@
...
+    else
+      "sudo ip netns exec #{vm.inhost_name} curl --insecure --resolve #{load_balancer.hostname}:#{load_balancer.dst_port}:#{(address.version == 6) ? "[#{address}]" : address} --max-time #{load_balancer.health_check_timeout} --silent --output /dev/null --write-out '%{http_code}' #{load_balancer.health_check_protocol}://#{load_balancer.hostname}:#{load_balancer.dst_port}#{load_balancer.health_check_endpoint}"
+    end
+  end
+
+  def check_pulse(session:, previous_pulse:)
+    reading_ipv4 = health_check(session:, ipv4: true)
+    reading_ipv6 = health_check(session:, ipv4: false)
+    reading = (reading_ipv4 == "up" && reading_ipv6 == "up") ? "up" : "down"
+    pulse = aggregate_readings(previous_pulse:, reading:, data: {ipv4: reading_ipv4, ipv6: reading_ipv6})
```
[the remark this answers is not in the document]
furkansahin: I was thinking it would be useful for debugging purposes.

migrate/20250128_make_lbvms_id_not_null.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    alter_table(:load_balancers_vms) do
+      set_column_not_null :id
```
bsatzger, line 6: how do you backfill the id of existing load_balancers_vms?

model/load_balancers_vms.rb
```diff
@@ -3,16 +3,75 @@
...
+
+    begin
+      (session[:ssh_session].exec!(health_check_cmd(type)).strip == "200") ? "up" : "down"
+    rescue
+      "down"
+    end
+  end
+
+  def health_check_cmd(type)
+    address = (type == :ipv4) ? vm.nics.first.private_ipv4.network : vm.ephemeral_net6.nth(2)
```
bsatzger, line 41: we have a number of calls throughout our codebase that gets addresses of a vm by performing calls like
`vm.nics.first.private_ipv4.network`. in the future, should we encapsulate this in a separate method?

migrate/20250128_make_lbvms_id_not_null.rb
furkansahin: After the first commit is deployed, I'll go to production and update `id`s

model/load_balancers_vms.rb
furkansahin: Makes sense, I'll look to implement that in a separate patch.

migrate/20250128_make_lbvms_id_not_null.rb
bsatzger: that's alright with me. a better approach might be to have another migration that executes those steps before setting the column to not null.

bsatzger (approved):

PR #2303, opened 2024-11-20
Move routes from routes/merged to routes

jeremyevans:
Recommended by @enescakir at https://github.com/ubicloud/ubicloud/pull/2277#pullrequestreview-2446492254

To prepare for this, move routes/helpers to helpers, routes/clover_error.rb to clover_error.rb, and routes/runtime to runtime-routes.

This also adds a couple cleanup commits for issues I discovered while working on this.


enescakir:
I'm not entirely sure why we can't keep the `helpers` and `runtime` directories in the `routes`. Is it related to the loader?

jeremyevans:
> I'm not entirely sure why we can't keep the `helpers` and `runtime` directories in the `routes`. Is it related to the loader?

It's due to how `Clover.autoload_routes` works, and also because I don't think it's a good idea to mix non-route files in the same directory as route files.

enescakir:
> > I'm not entirely sure why we can't keep the `helpers` and `runtime` directories in the `routes`. Is it related to the loader?
> 
> It's due to how `Clover.autoload_routes` works, and also because I don't think it's a good idea to mix non-route files in the same directory as route files.

Makes sense for `helpers` and `clover_error`. But since `runtime` is also routes, can't we keep it at `routes/runtime` and load as `autoload_routes("runtime", "routes/runtime")`

enescakir:
> > Makes sense for `helpers` and `clover_error`. But since `runtime` is also routes, can't we keep it at `routes/runtime` and load as `autoload_routes("runtime", "routes/runtime")`
> 
> Note without changes to `Clover.autoload_routes`. However, I can make those changes if you'd prefer that approach.

I'm not sure how, but I tried it locally and it worked without making any changes to `Clover.autoload_routes`.

jeremyevans:
> I'm not sure how, but I tried it locally and it worked without making any changes to `Clover.autoload_routes`.

"Worked" may mean the full test suite passes, which would be expected, as the full test suite and production take a different branch in `Clover.autoload_routes`. The problems caused would be in the other branch in `Clover.autoload_routes` that uses autoload_hash_branches.  Maybe the only problem is it would register the runtime github route using a namespace that isn't used in the application, but it still feels wrong to me to do that.

I'll work on fixing the issue cleanly so the runtime routes can remain at `routes/runtime`.

jeremyevans:
I was able to come up with a clean fix that avoids the problem and reduces overall complexity.

enescakir (approved):

PR #2304, opened 2024-11-20
DRY up r.rodauth call in main route, and move after check_active_session

jeremyevans:
This changes where r.rodauth is called for web routes, moving it after check_csrf!, rodauth.load_memory, and the root route.

To avoid double checking CSRF token after moving it after check_csrf!, set check_csrf? false in the Rodauth configuration.

The rodauth.load_memory change is probably better to have before r.rodauth than after, so if they logged in with the remember cookie, the rodauth pages will treat them requested as logged in.

For the root route, it just redirects to the login route, so it doesn't matter if it is before or after r.rodauth.  

For both the web routes and api routes, this moves r.rodauth after rodauth.check_active_session, so that if there is no active session, the rodauth routes will not treat the request as being authenticated.

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -203,6 +203,7 @@ class Clover < Roda
       :otp, :webauthn, :recovery_codes
 
     title_instance_variable :@page_title
+    check_csrf? false
 
     # :nocov:
     unless Config.development?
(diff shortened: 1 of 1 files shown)


fdr (commented):
Looks a little cleaner, but will defer to @byucesoy and/or @enescakir for a close read to make sure we're not losing anything on security.

enescakir (approved):
LGTM

PR #2307, opened 2024-11-21
Verify the purge logic in E2E tests.

pykello:
In the past, bugs in the purge logic went unnoticed, causing outages. This PR adds a step to the E2E tests to verify that resources are actually purged after VMs are destroyed.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -55,7 +55,7 @@ def main(options)
   # No need to make it parallel.
   tests_to_wait.each { |st| wait_until(st) }
 
-  Semaphore.incr(hetzner_server_st.id, "destroy")
+  Semaphore.incr(hetzner_server_st.id, "verify_cleanup_and_destroy")
   wait_until(hetzner_server_st)
 end
 
--- a/prog/test/hetzner_server.rb
+++ b/prog/test/hetzner_server.rb
@@ -3,7 +3,7 @@
 require_relative "../../lib/util"
 
 class Prog::Test::HetznerServer < Prog::Test::Base
-  semaphore :destroy, :allow_slices, :disallow_slices
+  semaphore :verify_cleanup_and_destroy, :allow_slices, :disallow_slices
 
   def self.assemble(vm_host_id: nil, default_boot_images: [])
     frame = if vm_host_id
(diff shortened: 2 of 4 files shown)


prog/test/vm_group.rb
```diff
@@ -135,6 +135,46 @@ def self.assemble(storage_encrypted: true, test_reboot: true)
       nap 5
     end
 
+    hop_verify_purge
+  end
+
+  def verify_vm_dir_purged
+    sshable = vm_host.sshable
+    vm_dir_content = sshable.cmd("sudo ls -1 /vm").split("\n")
```
enescakir, line 143: Since we run VM group tests and GitHub actions tests in parallel, even if this group doesn't have an active VM, others might. We might move this test to just before the Hetzner server is destroyed

fdr (commented):
This is a cursory review, so I'll defer to other reviewers, but: appreciate the enhancement, broadly, the form looks good.

prog/test/vm_group.rb
pykello: @enescakir done

prog/test/verify_purge.rb
```diff
@@ -0,0 +1,59 @@
...
+      prog: "Test::VerifyPurge",
+      label: "start",
+      stack: [{}]
+    )
+  end
+
+  label def start
+    verify_vm_dir_purged
+    verify_storage_files_purged
+    verify_spdk_artifacts_purged
```
enescakir, line 17: You could spread them across 3 labels instead of making 3 method calls in the same label.

prog/test/verify_purge.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    bdevs = JSON.parse(sshable.cmd("sudo #{rpc_py} -s #{rpc_sock} bdev_get_bdevs")).map { _1["name"] }
+    fail_test "SPDK bdevs not empty: #{bdevs}" unless bdevs.empty?
+
+    vhost_controllers = JSON.parse(sshable.cmd("sudo #{rpc_py} -s #{rpc_sock} vhost_get_controllers")).map { _1["ctrlr"] }
+    fail_test "SPDK vhost controllers not empty: #{vhost_controllers}" unless vhost_controllers.empty?
+  end
+
+  def vm_host
+    @vm_host ||= VmHost.first
```
enescakir, line 57: When you have multiple hosts and run e2e test on a local dev env (which I don't), it might cause failures.

enescakir (approved):
Since it has a single label, you could add this label between `wait` and `destroy` in the `HetznerServer` program instead of creating a separate one. `HetznerServer` has also correct vm host details.

PR #2308, opened 2024-11-21
Make clog.rb require sequel/model

jeremyevans:
It uses Sequel::Model internally, and without this require, spec/lib/clog_spec.rb fails if run in isolation.  Found by the spec_separate rake task.


fdr:
While permissible, it’s sort of a bummer for a logging utility procedure to
require a substantially larger program.

enescakir (approved):

PR #2309, opened 2024-11-21
Escape LIKE metacharacters when doing a prefix match in the runtime route

jeremyevans:
Not escaping LIKE metacharacters means that a prefix match isn't
forced if a user includes LIKE metacharacters in the key, which
can result in the index not being used.

While here, refactor runtime route entry setting:

* Use a dataset variable that can be shared by both cases.
* Switch from if entry.nil? to more idiomatic ||=.
* Use reverse instead of order with Sequel.desc.
* Use Dataset#grep to simplify prefix search.

### DIFF
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -22,9 +22,14 @@ class Clover
       # caches by scope, pushing entries for later scopes to the end of the
       # list.
       scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
-      entry = repository.cache_entries_dataset
+
+      dataset = repository.cache_entries_dataset
         .exclude(committed_at: nil)
-        .where(key: keys, version: version, scope: scopes).all
+        .where(version: version, scope: scopes)
+
+      entry = dataset
+        .where(key: keys)
+        .all
         .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
 
       # GitHub cache supports prefix match if the key doesn't match exactly.
@@ -32,14 +37,10 @@ class Clover
       #   When a key doesn't match directly, the action searches for keys
       #   prefixed with the restore key. If there are multiple partial matches
       #   for a restore key, the action returns the most recently created cache.
-      if entry.nil?
-        entry = repository.cache_entries_dataset
-          .exclude(committed_at: nil)
-          .where { keys.map { |key| Sequel.like(:key, "#{DB.dataset.escape_like(key)}%") }.reduce(:|) }
-          .where(version: version, scope: scopes)
-          .order(Sequel.desc(:created_at))
-          .first
-      end
+      entry ||= dataset
+        .grep(:key, keys.map { |key| "#{DB.dataset.escape_like(key)}%" })
+        .reverse(:created_at)
+        .first
 
       fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil?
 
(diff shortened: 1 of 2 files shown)


routes/runtime/github.rb
```diff
@@ -22,24 +22,25 @@ class Clover
...
-      if entry.nil?
-        entry = repository.cache_entries_dataset
-          .exclude(committed_at: nil)
-          .where { keys.map { |key| Sequel.like(:key, "#{DB.dataset.escape_like(key)}%") }.reduce(:|) }
-          .where(version: version, scope: scopes)
-          .order(Sequel.desc(:created_at))
-          .first
-      end
+      entry ||= dataset
+        .grep(:key, keys.map { |key| "#{DB.dataset.escape_like(key)}%" })
```
enescakir, line 41: I didn't know `dataset.grep`. Cool! https://www.rubydoc.info/github/jeremyevans/sequel/Sequel%2FDataset:grep

enescakir (approved):

routes/runtime/github.rb
```diff
@@ -26,11 +26,12 @@ class Clover
       dataset = repository.cache_entries_dataset
         .exclude(committed_at: nil)
         .where(version: version, scope: scopes)
+        .order(Sequel.case(scopes.map.with_index { |scope, idx| [{scope:}, idx] }.to_h, scopes.length))
 
       entry = dataset
         .where(key: keys)
-        .all
-        .min_by { keys.index(_1.key) + (scopes.index(_1.scope) * keys.size) }
+        .order_prepend(Sequel.case(keys.map.with_index { |key, idx| [{key:}, idx] }.to_h, keys.length))
```
enescakir, line 33: Should we use `order_append` instead of `order_prepend`?

Let's think a scenario like this:
```
keys = [a, b]
scopes = [dev, main]

cache_entries
1. {key: a, scope: main}
2. {key: b, scope: main}
3. {key: b, scope: dev}
```

The old code returns (3). If I'm not mistaken, the query returns (1) because it uses `ORDER BY key, scope`.

enescakir (commented):
Thanks for it. It's a very smart way of handling it in the database.

routes/runtime/github.rb
```diff
@@ -22,24 +22,26 @@ class Clover
       # caches by scope, pushing entries for later scopes to the end of the
```
enescakir, line 22: We need to update this comment to match with the query approach.

routes/runtime/github.rb
jeremyevans: Yep, you are correct.  I got confused because of the order of the `min_by`.  I'll fix this.

routes/runtime/github.rb
jeremyevans: I'll take care of that as well.

routes/runtime/github.rb
jeremyevans: I pushed a commit for this, with a new comment describing the prioritization.

routes/runtime/github.rb
```diff
@@ -13,33 +13,36 @@ class Clover
...
-      # default branch, we prioritize and return the cache for the head_branch.
-      # The part `(scopes.index(_1.scope) * keys.size)` assists in sorting the
-      # caches by scope, pushing entries for later scopes to the end of the
-      # list.
-      scopes = [runner.workflow_job&.dig("head_branch"), repository.default_branch].compact
-      entry = repository.cache_entries_dataset
+      # Clients can send multiple keys, and we look for caches in multiple scopes.
+      # We prioritize scope over key, returning the cache for the first matching
+      # key in the head branch scope, followed by the first matching branch in
+      # default branch scope.
```
enescakir, line 19:
```suggestion
      # key in the head branch scope, followed by the first matching key in
      # default branch scope.
```

enescakir (approved):

PR #2312, opened 2024-11-21
Committee 5

fdr:
Alright, here is @geemus's work, it's kind of a high-fidelity prototype that we need to chunk and commit. It adds systematic openapi validation of all requests and responses (some of our messages will change/deteriorate slightly from this), both in production and test [thus keeping it up to date], as well as a number of updates to the openapi.yml itself.


clover.rb
```diff
@@ -477,6 +513,22 @@ def assets
...
+      begin
+        schema_validator = SCHEMA_ROUTER.build_schema_validator(r)
+        schema_validator.request_validate(r)
+
+        raise Committee::NotFound, "That request method and path combination isn't defined." if !schema_validator.link_exist?
+      rescue JSON::ParserError => e
+        raise Committee::InvalidRequest.new("Request body wasn't valid JSON.", original_error: e)
+      end
+
+      r.hash_branches("api")
```
fdr, line 531: A mismerge on my part I think, probably needs to be reordered

clover.rb
```diff
@@ -107,6 +113,36 @@ def assets
...
+      when OpenAPIParser::NotExistPropertyDefinition
+        keys = e.original_error.instance_variable_get(:@keys)
+        message = "Validation failed for following fields: body"
+        details = {"body" => "Request body contains unrecognized parameters: #{keys.join(", ")}"}
+      when OpenAPIParser::NotExistRequiredKey
+        keys = e.original_error.instance_variable_get(:@keys)
+        message = "Validation failed for following fields: body"
+        details = {"body" => "Request body must include required parameters: #{keys.join(", ")}"}
+      else
+        raise e if Config.test?
```
jeremyevans, line 134: This and related branches need `# :nocov:` markings, since the implicit else branch cannot be hit in the tests.

clover.rb
jeremyevans, line 531:
```suggestion
      # Validate request against OpenAPI schema
      begin
        schema_validator = SCHEMA_ROUTER.build_schema_validator(r)
        schema_validator.request_validate(r)

        raise Committee::NotFound, "That request method and path combination isn't defined." if !schema_validator.link_exist?
      rescue JSON::ParserError => e
        raise Committee::InvalidRequest.new("Request body wasn't valid JSON.", original_error: e)
      end
```

clover.rb
```diff
@@ -477,6 +513,22 @@ def assets
...
+      rodauth.require_authentication
+
+      # Validate request against OpenAPI schema
+      begin
+        schema_validator = SCHEMA_ROUTER.build_schema_validator(r)
+        schema_validator.request_validate(r)
+
+        raise Committee::NotFound, "That request method and path combination isn't defined." if !schema_validator.link_exist?
+      rescue JSON::ParserError => e
+        raise Committee::InvalidRequest.new("Request body wasn't valid JSON.", original_error: e)
```
jeremyevans, line 528: Do we have to use `original_error` here?  Is Exception#cause not sufficient for this?

clover.rb
```diff
@@ -510,4 +562,18 @@ def assets
     rodauth.require_authentication
     r.hash_branches("")
   end
+
+  # Validate response against OpenAPI schema
+  after do |res|
+    status, headers, body = res
+    next unless api? && status && headers && body
+    schema_validator = SCHEMA_ROUTER.build_schema_validator(request)
```
jeremyevans, line 570: This duplicates what the main route block is doing.  I'm not sure how expensive `SCHEMA_ROUTER.build_schema_validator` is, but it seems like setting it to an instance variable in the main route, and using `||=` here, would avoid duplicate code.

jeremyevans (commented):
This should be rebased against main.  I did not review any of the yaml files, but the other changes look mostly good.  Please see inline comments.

clover.rb
geemus: Thanks, though I think it was my mistake in rebasing rather than yours, fwiw. It looks familiar when I was trying to collapse the api prefix special handling earlier today.

clover.rb
geemus: If I recall correctly, `original_error` is a reflection of how committee does some of this internally. Open to alternative ways to do it.

clover.rb
geemus: I don't think it's too bad expense-wise, but I agree that memoizing it would be worth doing.

clover.rb
jeremyevans: If that's how committee already works, it's fine to keep using it.  Sequel uses something similar, since that type of nested exception handling was supported before Exception#cause was added to Ruby.  I guess when committee was first being developed, it had to support Ruby versions that didn't have Exception#cause (it was added in Ruby 2.1).

clover.rb
fdr: I put in a rather brutal memoization. The brutality is there's an aspect of repeating myself. I played around with some of the conditionals a tiny bit, I was a bit puzzled how I could not assume `@schema_validator` set in some cases, rather than using `||=` at the last site. I did confirm that `@schema_validator` was set much of the time via print expressions, though.

clover.rb
fdr: Oh huh I guess I did something like that on my own in one of the revisions..

clover.rb
jeremyevans: You cannot assume it will be present, because it isn't set until fairly late in the main route block, after `r.rodauth` and some `rodauth` methods are called.   Setting `@schema_validator` would probably have to be the first thing in the route block (for api routes at least) for you to rely on it being present in the after hook.

fdr:
It's done! via #2356

PR #2317, opened 2024-11-22
Move openapi infrastructure to a directory, get a new openapi snapshot

fdr:
Some of the linting toolchain does not use hidden files, so they can
be annoying to mix with other top-level directory files.

This also picks up a new version of the openapi.yml from @geemus's
work that is more-correct, and far more complete: note that the length
has doubled.

It's not quite 100% correct, because at least a few hard-to-spec
features were added to the API -- probably accidentally -- that we'd
be better off removing, per daniel@ubicloud.com advisory.

These abnormalities are already identified in a development branch we
can pick through after, that has every single test passing the schema
verification.


fdr:
mostly this is to make sure the changes don't break the linter, which is part of CI

geemus (approved):
LGTM, thanks!

jeremyevans (approved):
I don't see anything objectionable, and this does pass CI.  The main change is in the openapi.yml file.  Might be easier to review if the move and update were in separate commits, but it wouldn't help my review, because even then I wouldn't know what to look for.

fdr:
> I don't see anything objectionable, and this does pass CI. The main change is in the openapi.yml file. Might be easier to review if the move and update were in separate commits, but it wouldn't help my review, because even then I wouldn't know what to look for.

yeah, I think we're still in the "inflationary" (to borrow a cosmology term) phase for the openapi file, since #2312  (or its constituent PRs as we go) changes our development obligations to incrementally update it. But for now, it drifts without convergence, so I think a big bang update is okay

PR #2318, opened 2024-11-22
Change error handling from redirecting to using an internal request

jeremyevans:
This changes redirect_back_with_inputs to use an internal request instead
of redirecting back.  The main advantage from doing this is you do not
need to store the submitted parameters in the flash, which avoids overflowing
the session (CookieTooLarge errors) and avoids JSON::GeneratorError if the
submitted inputs contain invalid UTF-8.

The internal requests use the rack API to Clover, so they appear to Clover
to be new GET requests from the server, for the referring page.  We rotate
the flash before the internal request, matching the previous redirect
behavior.

This changes some requests that result in error pages to use 400 responses
instead of 200 responses.  The 200 responses are not accurate, since the
request was a failure and not a success, they were an artifact of the
redirect implementation.  Other than this behavior improvement, there are
no changes to the specs, except for one new spec for a corner case (bad
Referer header value).

Note that redirects are still used inside of internal requests if the
original request is a GET request.  This is done to prevent internal
request loops, but the issue we are trying to avoid is not a significant
issue for GET requests, so it seems fine to limit it to non-GET requests.

For invalid or missing referer values, this redirects back to the root,
with no setting of old values.  The old values wouldn't be helpful for a
root request anyway.

The `flash["referrer"] = referrer` code is used to handle the following
case:

* User goes to path X
* User submits form to path Y
* Results in error, redisplaying form shown on path X, but browser
  referer is now Y and not X
* User resubmit form
* Which has the same or other error, and we want to redisplay form shown
  on path X, not on path Y

We may want to consider changing redirect_back_with_inputs to a different
name, now that the standard implementation does an internal request. But
that can be done later.

This also removes unnecessary returns when using redirect_back_with_inputs.


enescakir:
flash-error
The error appears again when we navigate to another page after a failure

jeremyevans:
> The error appears again when we navigate to another page after a failure

Thanks for finding that issue.  I pushed a commit that fixes it by discarding the flash after the internal request, and setting the referring in the flash after the discard.

helpers/web.rb
```diff
@@ -6,7 +6,27 @@ def csrf_tag(*)
...
+    end
+
+    request.redirect "/" unless uri
+
     flash["old"] = request.params
-    request.redirect env["HTTP_REFERER"] || "/"
+
+    if uri && env["REQUEST_METHOD"] != "GET"
+      # Force flash rotation, so flash works correctly for internal redirects
+      _roda_after_40__flash(nil)
```
enescakir, line 22: Only you can know this method 😄

enescakir (approved):

PR #2319, opened 2024-11-22
Postgres Endpoint Adjustments

fdr:
@geemus found these abnormalities while typing the entire schema in openapi that look like oversights:

```
Serialize the Postgres firewall rule when POSTing to the same

This looks like an oversight, to not return the rule, but the Postgres
resource instance related to the rule.
```

```
Nest postgres firewall rules into "items" key

Other routes use use "items" to list things, so do this here too.
```

### DIFF
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -74,7 +74,10 @@ class Clover
       r.on "firewall-rule" do
         r.get api?, true do
           authorize("Postgres:Firewall:view", pg.id)
-          Serializers::PostgresFirewallRule.serialize(pg.firewall_rules)
+          {
+            items: Serializers::PostgresFirewallRule.serialize(pg.firewall_rules),
+            count: pg.firewall_rules.count
+          }
         end
 
         r.post true do
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -87,16 +87,16 @@ class Clover
           request_body_params = Validation.validate_request_body(json_params, required_parameters)
           parsed_cidr = Validation.validate_cidr(request_body_params["cidr"])
 
-          DB.transaction do
+          firewall_rule = DB.transaction do
+            pg.incr_update_firewall_rules
             PostgresFirewallRule.create_with_id(
               postgres_resource_id: pg.id,
               cidr: parsed_cidr.to_s
             )
-            pg.incr_update_firewall_rules
           end
 
           if api?
-            Serializers::Postgres.serialize(pg, {detailed: true})
+            Serializers::PostgresFirewallRule.serialize(firewall_rule, {detailed: true})
           else
             flash["notice"] = "Firewall rule is created"
             r.redirect "#{@project.path}#{pg.path}"
--- a/spec/routes/api/project/location/postgres_spec.rb
+++ b/spec/routes/api/project/location/postgres_spec.rb
@@ -342,7 +342,8 @@ RSpec.describe Clover, "postgres" do
         get "/project/#{project.ubid}/location/#{pg.display_location}/postgres/_#{pg.ubid}/firewall-rule"
 
         expect(last_response.status).to eq(200)
-        expect(JSON.parse(last_response.body)[0]["cidr"]).to eq("0.0.0.0/0")
+        expect(JSON.parse(last_response.body)["items"][0]["cidr"]).to eq("0.0.0.0/0")
+        expect(JSON.parse(last_response.body)["count"]).to eq(1)
       end
     end
 
--- a/spec/routes/api/project/location/postgres_spec.rb
+++ b/spec/routes/api/project/location/postgres_spec.rb
@@ -167,6 +167,7 @@ RSpec.describe Clover, "postgres" do
           cidr: "0.0.0.0/24"
         }.to_json
 
+        expect(JSON.parse(last_response.body)["cidr"]).to eq("0.0.0.0/24")
         expect(last_response.status).to eq(200)
       end
 


fdr:
involving @furkansahin because he may be able to eliminate this kind of duplication (firewall rules data models unique to every service) in the future, but I don't want openapi maintenance to wait quite that long

routes/project/location/postgres.rb
```diff
@@ -84,16 +87,15 @@ class Clover
...
           parsed_cidr = Validation.validate_cidr(request_body_params["cidr"])
 
-          DB.transaction do
+          firewall_rule = DB.transaction do
             PostgresFirewallRule.create_with_id(
               postgres_resource_id: pg.id,
               cidr: parsed_cidr.to_s
-            )
-            pg.incr_update_firewall_rules
+            ).tap { pg.incr_update_firewall_rules }
```
fdr, line 94: paging @jeremyevans : this unusual use of tap I do once in a while,  because `create_with_id` is not powerful enough to allow subsequent incr of semaphores. Otherwise awkward and error-prone binding of symbols has to be done. But maybe the unusual pattern of using `tap` is a lesser defect than making a special device for our ubid'd model creations.

fdr: Possibly create_with_id ought to take a block that merely gets run after.

geemus:
An adjacent thing I noticed, fwiw, I believe there is a similar serialization gap around postgres metric destination, I think they return postgres serializations. It was less apparent to me what the right serialization was.

routes/project/location/postgres.rb
jeremyevans: Having `create_with_id` take a block seems like a good approach to me.

spec/routes/api/project/location/postgres_spec.rb
```diff
@@ -342,7 +342,7 @@
         get "/project/#{project.ubid}/location/#{pg.display_location}/postgres/_#{pg.ubid}/firewall-rule"
 
         expect(last_response.status).to eq(200)
-        expect(JSON.parse(last_response.body)[0]["cidr"]).to eq("0.0.0.0/0")
+        expect(JSON.parse(last_response.body)["items"][0]["cidr"]).to eq("0.0.0.0/0")
```
jeremyevans, line 345: Probably worth adding a line that checks for `count`, considering that's being added in the route.

routes/project/location/postgres.rb
```diff
@@ -84,16 +87,15 @@ class Clover
...
           end
 
           if api?
-            Serializers::Postgres.serialize(pg, {detailed: true})
+            Serializers::PostgresFirewallRule.serialize(firewall_rule, {detailed: true})
```
jeremyevans, line 98: Probably worth adding a test for this change.

jeremyevans (approved):
This change looks fine to me, though I think it would be worth it to add a couple tests/assertions.

routes/project/location/postgres.rb
geemus: I think it does get tested, albeit somewhat indirectly, due to the committee tests. ie there are tests that call this and if it didn't return what the spec asks for it would error. That said, it may be worth it to also be more explicit.

spec/routes/api/project/location/postgres_spec.rb
geemus: Similar here, I think absence of count would cause an error due to committee/schema, but more explicit tests would not go amiss (and would communicate intent more clearly).

routes/project/location/postgres.rb
fdr: Yes. Committee is quite strict and uniform. When doing the prototype, the cost was measurable but virtual. We can rebate some of that via fewer tests for form.

spec/routes/api/project/location/postgres_spec.rb
jeremyevans: I would guess that committee checks structure.  The advantage of a test here is we could test  the expected value of `count` (unless committee checks that as well?).

TBH, I'm not sure what the advantage of return `count` separately is, assuming it always matches the count of `items`.

routes/project/location/postgres.rb
jeremyevans: Likewise, I assume the committee tests for structure, whereas we would want a test for expected values.

spec/routes/api/project/location/postgres_spec.rb
fdr: Yeah, it doesn't check exact values, though it can constrain a number of other matters of format https://www.perplexity.ai/search/besides-nullity-what-kind-of-v-JGdB00QmS0qgHP.7S3KU0g

And true, not sure why there's a count in there. It's not in a place that can help the client or server do work

geemus: Fair question about count. In this case I was just following what appeared to be a relatively consistent example across existing list actions, rather than really thinking all that deeply about the particulars. It might be worth dropping to simplify, but probably should do it across the board to keep those kinds of actions consistent.

byucesoy (approved):

routes/project/location/postgres.rb
fdr: perhaps in this case I could actually reorder the increment before, though in others...with `assemble`...the strand is not yet created, so there's a foreign key problem. Maybe I'll look to tweak this one to get rid of tap.

spec/routes/api/project/location/postgres_spec.rb
fdr: Hmm okay, let's mark count for elimination unless there's an explicable countervailing reason

routes/project/location/postgres.rb
fdr: reordered it. it cost me a line, but I think is less subtle (usually tap uses the received object)

spec/routes/api/project/location/postgres_spec.rb
fdr: I went ahead and threw in an assertion until we do that though

routes/project/location/postgres.rb
fdr: I added one puny test for a value [not that the firewall rules for postgres have many values to set] as well, that should probably do the job with committee so far

fdr:
@jeremyevans @byucesoy re-requesting review for my tiny edits.

spec/routes/api/project/location/postgres_spec.rb
geemus: The only other thing I could think of would be pagination (which I don't recall the plan/pattern for). ie if count is say 100, but you only get 20 results at a time. There could be some value/reason there that wouldn't be apparent in the smaller sample responses we would otherwise be looking at.

PR #2323, opened 2024-11-22
Only serialize feature flags in the web serialization

fdr:
An opaque hash of feature flags is awkward to type in openapi schema, and was only used in the web views.  So limit it only to a "web" variant of the serializer, and leave it out of the API.

It raises at least one question: is what's going on for the web views in fact a "serialization" at all?

### DIFF
--- a/routes/project.rb
+++ b/routes/project.rb
@@ -17,7 +17,7 @@ class Clover
           count: result[:count]
         }
       else
-        @projects = Serializers::Project.serialize(dataset.all, {include_path: true})
+        @projects = Serializers::Project.serialize(dataset.all, {include_path: true, web: true})
         view "project/index"
       end
     end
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
This change looks fine.

I don't think any of the serializers actually do serialization, they merely format a hash or array of hashes that the Roda json plugin then serializes.  I'm not sure why they are used in the web routes, instead of accessing the related data directly on the model objects.

Almost all serializer calls are in the routes, with only `Serializers::Invoice` being used in `model/invoice.rb` and `Serializers::BillingInfo` being used in `lib/invoice_generator.rb`.  I can understand using serializers in the api routes, though having them being instance methods on `Clover` seems like a better approach.  It would avoid the need to pass the `:web` option in this PR, since the helper method could use `web?`.

fdr:
> This change looks fine.
> 
> I don't think any of the serializers actually do serialization, they merely format a hash or array of hashes that the Roda json plugin then serializes. I'm not sure why they are used in the web routes, instead of accessing the related data directly on the model objects.
>

That's just the thing, historically, in https://github.com/interagent/pliny, which was about making APIs only, they _were_ tantamount to serialization, now they're not, this is a strange evolution that may have drifted from the original purpose.

byucesoy (approved):
Looks good.

PR #2326, opened 2024-11-25
Add name validation for projects

jeremyevans:
Currently, even projects with empty names are allowed. The UI tries to avoid this by using a required attribute on the inputs, but it's trivial for users to work around that.

To not break existing projects with invalid names, only validate for new projects, and when the project's name has changed.

For the validation format, use a case-insensitive version of Validation::ALLOWED_NAME_PATTERN.

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -121,6 +121,13 @@ class Project < Sequel::Model
     ApiKey.create_with_id(owner_table: Project.table_name, owner_id: id, used_for: used_for)
   end
 
+  def validate
+    super
+    if new? || changed_columns.include?(:name)
+      validates_format(%r{\A[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?\z}i, :name)
+    end
+  end
+
   def self.feature_flag(*flags, into: self)
     flags.map(&:to_s).each do |flag|
       into.module_eval do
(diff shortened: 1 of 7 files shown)


fdr (approved):
I wonder if spaces should be allowed, though not-allowing them has some advantages as well. Would like second opinions from others on this one. Seems better than status quo, which allows very inconvenient inputs. @enescakir @byucesoy @geemus ?

geemus:
Broadly speaking, if it remains true that names can be a URL param, you will probably save headaches if you avoid allowing url unsafe characters. It would probably also be good to include this regex as a `format` value in the schema when we merge everything. Both because it should help document the expectation and because committee should then help enforce it as well.

enescakir:
It makes sense to ensure project names are URL safe. But it might requires small improvements on UI for project switcher.

| Before | After |
|--------|-------|
| Screenshot 2024-11-26 at 11 52 00 | Screenshot 2024-11-26 at 11 51 29 |
| Screenshot 2024-11-26 at 11 51 56 | Screenshot 2024-11-26 at 11 51 25 |

jeremyevans:
> It makes sense to ensure project names are URL safe. But it might requires small improvements on UI for project switcher.

The difference you showed seems acceptable to me.  I played around with Chrome Inspector and I couldn't get the same behavior as spaces.  Since the visual issue seems minor, and I think we will significantly benefit from some validation. I'm going to merge this.

PR #2329, opened 2024-11-25
Add support for personal access tokens for API authentication

jeremyevans:
This now implements all parts: authentication, authorization, and management UI.

### DIFF
--- a/model/api_key.rb
+++ b/model/api_key.rb
@@ -19,6 +19,10 @@ class ApiKey < Sequel::Model
     UBID::TYPE_ETC
   end
 
+  def self.create_personal_access_token(account, project: nil)
+    create_with_id(owner_table: "accounts", owner_id: account.id, used_for: "api")
+  end
+
   def self.create_with_id(owner_table:, owner_id:, used_for:)
     unless %w[project inference_endpoint accounts].include?(owner_table.to_s)
       fail "Invalid owner_table: #{owner_table}"
--- a/model/api_key.rb
+++ b/model/api_key.rb
@@ -19,7 +19,7 @@ class ApiKey < Sequel::Model
     end
 
     key = SecureRandom.alphanumeric(32)
-    super(owner_table: owner_table, owner_id: owner_id, key: key, used_for: used_for)
+    super(owner_table:, owner_id:, key:, used_for:)
   end
 
   def rotate
--- a/model/api_key.rb
+++ b/model/api_key.rb
@@ -4,9 +4,15 @@ require_relative "../model"
 
 class ApiKey < Sequel::Model
   include ResourceMethods
+  include Authorization::TaggableMethods
+  include Authorization::HyperTagMethods
 
   plugin :column_encryption do |enc|
     enc.column :key
+  end
+
+  def hyper_tag_name(project = nil)
+    "api-key/#{ubid}"
   end
 
   def self.ubid_type
--- a/model/api_key.rb
+++ b/model/api_key.rb
@@ -7,6 +7,9 @@ class ApiKey < Sequel::Model
   include Authorization::TaggableMethods
   include Authorization::HyperTagMethods
 
+  one_to_many :access_tags, key: :hyper_tag_id
+  plugin :association_dependencies, access_tags: :destroy
+
   plugin :column_encryption do |enc|
     enc.column :key
   end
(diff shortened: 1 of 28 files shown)


model/api_key.rb
```diff
@@ -4,11 +4,17 @@
...
   include ResourceMethods
+  include Authorization::TaggableMethods
+  include Authorization::HyperTagMethods
 
   plugin :column_encryption do |enc|
     enc.column :key
   end
 
+  def hyper_tag_name(project = nil)
+    "api_key/#{ubid}"
```
enescakir, line 18: We use `-` instead of `_` in hyper tags

jeremyevans: Fixed in 4544098

jeremyevans:
> What do you think about these UI changes? https://github.com/ubicloud/ubicloud/compare/jeremy-personal-access-tokens...pat-ui?expand=1
> 
> 1. We have `revealable_content` component, we can use it.
> 2. We can move create button to the right of title similar to other listing pages.

That looks much better, thank you.

jeremyevans:
> > What do you think about these UI changes? https://github.com/ubicloud/ubicloud/compare/jeremy-personal-access-tokens...pat-ui?expand=1
> > 
> > 1. We have `revealable_content` component, we can use it.
> > 2. We can move create button to the right of title similar to other listing pages.
> 
> That looks much better, thank you.

I merged in the changes and updated the new tests to work with them.

bsatzger (approved):
This looks great. I also feel it makes sense to reuse the ApiKey model.

enescakir:
I was planning to finish my review for this PR today, but couldn't find the time due to customer tickets. I'll review it tonight or tomorrow morning.

enescakir:
I'm a bit confused. We create PATs for accounts, and these accounts owns these PATs. However, we display them on the Project User page, which is essentially the access management page for the project. We only display the current account's PATs there. This means that each project user sees different content on the project's access management page, which I find slightly confusing.
If I'm not mistaken whole concept, I believe it would make more sense to display PATs on the 'My Account - /account' page.

something like that:
CleanShot 2024-11-29 at 19 25 02@2x

bsatzger:
Agreed, "My Account" seems to be a better home for PATs.

jeremyevans:
PATs are currently created per project. PATs are both project-specific and account-specific:

*  When creating, PATs are associated with the current project (https://github.com/ubicloud/ubicloud/pull/2329/files#diff-c55888791a06dd284d4094b34de7fda3ad2441b32286421f379160813d53528bR75).
* Updating policies for PATs affects the current project (https://github.com/ubicloud/ubicloud/pull/2329/files#diff-c55888791a06dd284d4094b34de7fda3ad2441b32286421f379160813d53528bR86)

Maybe we need a new page for PATs, but it should be a project-specific page and not an account-specific page.  What do you think?

enescakir:
We can still put it to 'My Account' and get the project as select input, similar to how it's done in Cloudflare.
CleanShot 2024-12-02 at 18 29 58@2x

But I'm also okay with creating separate project based for PATs.

I have another high level question. I may have missed the discussion, but why do we create API tokens for both the project and the account specific? Can't we just make them project-specific and share them among project members, like we do with other resources?

jeremyevans:
> We can still put it to 'My Account' and get the project as select input, similar to how it's done in Cloudflare. CleanShot 2024-12-02 at 18 29 58@2x

OK, I can look into switching the UI to that.

> I have another high level question. I may have missed the discussion, but why do we create API tokens for both the project and the account specific? Can't we just make them project-specific and share them among project members, like we do with other resources?

The way security for PATs currently works, it checks access both for the account and for the token itself.  That way an account cannot use a PAT to elevate their access.

Accountless access tokens, where access is just checked for for the account, would be equivalent to a PAT for a shared admin account.  I think other parts of Clover expect an account for requests, so a true accountless access token would be a more invasive change.

Perhaps we could support a shared admin account in the future, where the shared admin account does not have a password and cannot login, but other admin accounts can still create an access PATs for the shared account.  You would have to be sure that only other admin accounts could create PATs for a shared admin account, as otherwise it would allow for access elevation. Maybe we could build security so an admin account can set permissions on which other accounts can view which PATs for a shared admin account?

enescakir (approved):

PR #2334, opened 2024-11-26
Runner VM's networking reconfigurations

furkansahin:
## Allow /80 to be used
Since we are manually configuring the ipv6 network, we add the proper
route with "via". However, the addr replace also adds a route for the
same address without via which blocks the /80 prefix to be used by VM
freely.

## Enable Docker in runner with IPv6
Makes the configuration updates to assign a static cidr for docker ipv6
so that docker can use double stack networking

An important update is single-request-reopen.
Docker has issues with double stack dns resolution.  The
single-request-reopen option modifies the behavior of the DNS resolver
when handling IPv4 and IPv6 queries. By default, glibc sends both A
(IPv4) and AAAA (IPv6) DNS queries over the same socket. However, some
DNS servers or network appliances may not handle this properly, leading
to issues such as: Only one of the responses (either A or AAAA) being
returned. In this case, docker's internal dns resolver cannot handle the
responses that comes from dnsmasq. When single-request-reopen is
enabled: The resolver will close the socket after sending the first
query (e.g., an A query).  It will then open a new socket to send the
second query (e.g., an AAAA query). This ensures that each query is
sent using a separate connection, avoiding compatibility issues with
misbehaving DNS servers.

## Update Dnsmasq to v2.90

### DIFF
--- a/prog/install_dnsmasq.rb
+++ b/prog/install_dnsmasq.rb
@@ -30,10 +30,10 @@ class Prog::InstallDnsmasq < Prog::Base
   end
 
   label def git_clone_dnsmasq
-    q_commit = "5dc14b6e05f39a5ab0dc02e376b1d7da2fda5bc1".shellescape
+    q_commit = "b6769234bca9b0eabfe4768832b88d2cdb187092".shellescape
     sshable.cmd("git init dnsmasq && " \
                 "(cd dnsmasq && " \
-                "  git fetch https://github.com/fdr/dnsmasq.git #{q_commit} --depth=1 &&" \
+                "  git fetch https://github.com/ubicloud/dnsmasq.git #{q_commit} --depth=1 &&" \
                 "  git checkout #{q_commit} &&" \
                 "  git fsck --full)")
     pop "downloaded and verified dnsmasq successfully"
--- a/spec/prog/install_dnsmasq_spec.rb
+++ b/spec/prog/install_dnsmasq_spec.rb
@@ -55,7 +55,7 @@ RSpec.describe Prog::InstallDnsmasq do
     it "fetches a version and pops" do
       sshable = instance_double(Sshable)
       expect(sshable).to receive(:cmd).with <<CMD.rstrip
-git init dnsmasq && (cd dnsmasq &&   git fetch https://github.com/fdr/dnsmasq.git 5dc14b6e05f39a5ab0dc02e376b1d7da2fda5bc1 --depth=1 &&  git checkout 5dc14b6e05f39a5ab0dc02e376b1d7da2fda5bc1 &&  git fsck --full)
+git init dnsmasq && (cd dnsmasq &&   git fetch https://github.com/ubicloud/dnsmasq.git b6769234bca9b0eabfe4768832b88d2cdb187092 --depth=1 &&  git checkout b6769234bca9b0eabfe4768832b88d2cdb187092 &&  git fsck --full)
 CMD
       expect(idm).to receive(:sshable).and_return(sshable)
 


prog/vm/github_runner.rb
```diff
@@ -220,6 +220,8 @@ def setup_info
       # ubicloud/cache package which forked from the official actions/cache package, sends requests to UBICLOUD_CACHE_URL using this token.
       echo "UBICLOUD_RUNTIME_TOKEN=#{vm.runtime_token}
       UBICLOUD_CACHE_URL=#{Config.base_url}/runtime/github/" | sudo tee -a /etc/environment
+      echo '{"ipv6": true, "fixed-cidr-v6": "#{vm.ephemeral_net6.nth(2)}:dc/124", "experimental": false, "dns": ["#{vm.nics.first.private_subnet.net4.nth(2)}", "fd00:0b1c:100d:53::"]}' | sudo tee /etc/docker/daemon.json
```
fdr, line 223: I wonder if you should break out a JSON.generate(...).shellescape. I suppose we could go with this if we never anticipate ever wanting to change it again, because a typo would cause havoc in short order given where it is.

rhizome/host/lib/vm_setup.rb
```diff
@@ -275,6 +275,7 @@ def setup_taps_6(gua, nics, dns_ipv4)
       # Route ephemeral address to tap.
       r "ip -n #{q_vm} link set dev #{nic.tap} up"
       r "ip -n #{q_vm} route replace #{guest_ephemeral.to_s.shellescape} via #{mac_to_ipv6_link_local(nic.mac)} dev #{nic.tap}"
+      r "ip -n #{q_vm} route del #{guest_ephemeral.to_s.shellescape} dev #{nic.tap}"
```
fdr, line 278: good catch!

rhizome/host/lib/vm_setup.rb
```diff
@@ -440,10 +441,18 @@ def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image, dn
     runner_config = if boot_image.include?("github")
       <<~ADDRESSES
       address=/ubicloudhostplaceholder.blob.core.windows.net/#{nics.first.net4.split("/").first}
-      filter-AAAA
+      server=9.9.9.11
+      server=149.112.112.11
+      server=2620:fe::11
+      server=2620:fe::fe:11
```
fdr, line 447: very astute. I remember @enescakir was mentioning EDNS and I waffled on it, but if this casually helps with CDNs that serve a bunch of code, I think that's a lock: we should enable it. The privacy considerations are not the same for end users versus servers, I don't think.

rhizome/host/lib/vm_setup.rb
```diff
@@ -453,14 +462,12 @@ def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image, dn
 domain-needed
 bogus-priv
 no-resolv
+cache-size=0
```
fdr, line 465: this one is kind scary, caching is usually very good....do we absolutely need it? many operating systems even cache by default, CDNs have to deal with that.

rhizome/host/lib/vm_setup.rb
enescakir: https://en.wikipedia.org/wiki/Quad9

Screenshot 2024-11-26 at 22 28 37

enescakir: Screenshot 2024-11-26 at 22 29 43

prog/vm/github_runner.rb
```diff
@@ -220,6 +220,8 @@ def setup_info
...
+      sudo systemctl restart docker
```
enescakir, line 224: Do you have any idea how much this will increase our provisioning times?

enescakir: Runner script allows to run a hook script before the job starts. I'm adding a start hook script for another feature. We might move docker restart to there too.
https://github.com/ubicloud/runner-images/pull/8

enescakir: It adds a "Set up runner" step after "Set up job" step. So it runs before container operations anyway.

https://github.com/enescakir/ubicloud-test-dev/actions/runs/12038401443/job/33563995301

rhizome/host/lib/vm_setup.rb
furkansahin: Turned out that was not necessary, it was all about the docker + dns

rhizome/host/lib/vm_setup.rb
furkansahin: Removed this. Turns out, not necessary.

enescakir (approved):

PR #2335, opened 2024-11-27
Fix typo in explanation in views/project/user.erb

TimWolla:


jeremyevans (approved):
I included this typo fix in #2329, but there's no problem with merging this separately if the CLA is signed.

TimWolla:
Not signing the CLA for this change. I _believe_ it is non-copyrightable, but I would need to run the CLA through company legal nevertheless and that really isn't worth anyone's time and associated cost.

PR #2339, opened 2024-11-27
Return cache in incoming key order for partial match

enescakir:
We sorted cache entries according to the order of the incoming keys and returned the first match. However, this only works when the full key is provided. It seems that partial match doesn't follow the same ordering. Instead, it simply orders the entries by their creation time. We should order them by the incoming key order.

I used similar logic to the full match for this. The only difference is that I used `find_index` and `start_with?` instead of `index`.

### DIFF
--- a/routes/runtime/github.rb
+++ b/routes/runtime/github.rb
@@ -39,8 +39,7 @@ class Clover
       #   for a restore key, the action returns the most recently created cache.
       entry ||= dataset
         .grep(:key, keys.map { |key| "#{DB.dataset.escape_like(key)}%" })
-        .reverse(:created_at)
-        .first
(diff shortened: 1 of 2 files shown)


routes/runtime/github.rb
```diff
@@ -39,8 +39,7 @@ class Clover
       #   for a restore key, the action returns the most recently created cache.
       entry ||= dataset
         .grep(:key, keys.map { |key| "#{DB.dataset.escape_like(key)}%" })
-        .reverse(:created_at)
-        .first
+        .min_by { |e| keys.find_index { |k| e.key.start_with?(k) } + (scopes.index(e.scope) * keys.size) }
```
jeremyevans, line 42: I'm not sure how large the dataset is, and `min_by` would need to iterate over the entire set.  I'll merge this, but I'll try to get this back to the order+first approach.

PR #2341, opened 2024-11-27
Fix error handling for delete user

jeremyevans:
The hash with message key needs to be wrapped in an error key in order for the javascript to handle it.

While here, refactor:

* Only one route in the block, so combine `r.is String` and `r.delete true` to `r.delete String`.

* Use `next` instead of `response.status = 404; r.halt` for early exit.

* Avoid `return`, use `next` for early exit, and just have block return the hash implicitly.

### DIFF
--- a/routes/project/user.rb
+++ b/routes/project/user.rb
@@ -117,31 +117,28 @@ class Clover
         end
       end
 
-      r.is String do |user_ubid|
-        user = Account.from_ubid(user_ubid)
+      r.delete String do |user_ubid|
+        next unless (user = Account.from_ubid(user_ubid))
 
-        unless user
-          response.status = 404
-          r.halt
+        unless @project.accounts.count > 1
+          response.status = 400
+          next {error: {message: "You can't remove the last user from '#{@project.name}' project. Delete project instead."}}
         end
 
-        r.delete true do
-          unless @project.accounts.count > 1
-            response.status = 400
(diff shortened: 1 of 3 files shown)


jeremyevans:
While fixing the tests for this, I found a general bug in how 404 responses were handled for web requests that only accept JSON, so this includes a fix for that.

routes/project/user.rb
```diff
@@ -117,31 +117,25 @@ class Clover
...
-              policy.save_changes
-            end
-          end
-          user.dissociate_with_project(@project)
-
-          return {message: "Removing #{user.email} from #{@project.name}"}
         end
+        user.dissociate_with_project(@project)
+
+        {message: "Removing #{user.email} from #{@project.name}"}
```
enescakir, line 138: We don't use response https://github.com/ubicloud/ubicloud/blob/main/assets/js/app.js#L79-L81

We may return 204 no content similar to other deletes

enescakir (approved):

routes/project/user.rb
jeremyevans: Pushed a commit for that change.

enescakir (approved):

jeremyevans:
I updated this to set a flash notice, similar to the delete invited user handling, as the javascript will handle the redirect/refresh on success.

enescakir (approved):

PR #2347, opened 2024-11-28
Fix delete invited user route

jeremyevans:
This does not redirect, so it should not set flash.  Otherwise, the next request the user makes will display the set flash.

For consistency, use a 204 status, similar to other delete routes.

While here, avoid unnecessary nesting.

### DIFF
--- a/routes/project/user.rb
+++ b/routes/project/user.rb
@@ -107,14 +107,12 @@ class Clover
         end
       end
 
-      r.on "invitation" do
-        r.is String do |email|
-          r.delete true do
-            @project.invitations_dataset.where(email: email).destroy
-            flash["notice"] = "Invitation for '#{email}' is removed successfully."
-            r.halt
-          end
-        end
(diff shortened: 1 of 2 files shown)


enescakir:
> This does not redirect, so it should not set flash. Otherwise, the next request the user makes will display the set flash.

The frontend code refresh page when it completes the delete request

enescakir (approved):

jeremyevans:
> > This does not redirect, so it should not set flash. Otherwise, the next request the user makes will display the set flash.
> 
> The frontend code refresh page when it completes the delete request. So the flash message shown there. But yes, it's not very necessary

Ah, I was not aware of that. I think we should retain the flash message then.  We should probably also add one to the delete user route, assuming that is handled the same way.

jeremyevans:
Added back the flash notice, and added a test for it.

enescakir (approved):

PR #2350, opened 2024-11-29
Remove global dashboard page

enescakir:
- **Remove global dashboard page**
  We have two dashboards: one for global use and one for the project. The global dashboard isn’t used much and isn’t very helpful.
  
  It often confuses me, so I removed it.
  
  Less page to maintain is better.
  
| Global Dashboard | Project Dashboard |
|--------|-------|
| CleanShot 2024-11-29 at 21 46 29@2x | CleanShot 2024-11-29 at 21 45 47@2x |

- **Add link to Ubicloud logo on console**
  I instinctively clicked on the Ubicloud logo on the console multiple
  times, expecting it to take me to the home page. I believe this is a
  common user behavior.


jeremyevans (approved):
I don't think we have any screenshots in our docs that show non-project pages, so I don't think this requires regenerating screenshots.

PR #2351, opened 2024-11-29
Add default value to back field of breadcrumb component

enescakir:
We use the second the last as back value for breadcrumb component generally. So we can use it as default value.

### DIFF
--- a/views/components/breadcrumb.erb
+++ b/views/components/breadcrumb.erb
@@ -11,18 +11,13 @@
   </nav>
   <nav class="hidden sm:flex" aria-label="Breadcrumb">
     <ol role="list" class="flex items-center space-x-4">
-      <% parts.take(1).each do |name, url| %>
-        <li>
-          <div class="flex">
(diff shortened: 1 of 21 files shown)


jeremyevans (approved):
This looks good and simplifies things.  A couple issues I noted while review this, which could be fixed later:

* I think not displaying a Back link makes more sense then a link to `#`.

* I'm not sure why, but the Compute and Networking navigation links take you to pages that don't display breadcrumbs, which is inconsistent with other pages.

enescakir:
> This looks good and simplifies things. A couple issues I noted while review this, which could be fixed later:
> 
> * I think not displaying a Back link makes more sense then a link to `#`.
> * I'm not sure why, but the Compute and Networking navigation links take you to pages that don't display breadcrumbs, which is inconsistent with other pages.

1. Yes, we can hide back button.
2. I think we should add page header and breadcrumbs to them too. I will create a PR for that

PR #2356, opened 2024-12-02
Enable committee openapi schema checking in test and production

fdr:
Enable committee openapi schema checking in test and production

Committee is rack middleware that validates requests and responses for
adherence to an openapi specification.  It will limit the kind of
(possibly malicious) input to the API part of the program.  And, it
because it checks the test suite, and we have a 100% branch coverage
policy covering routes, fixing schema problems will keep the openapi
specification up to date as people add features.

committee already detected some mistakes, such as being too relaxed
about content types, in 793b61f260b872f79095ce73d95fd99789d672ad.

Applying committee obsoletes some validation code, but not a lot: at
least some validations are also called from the web application,
rather than the API.  It's not clear to me how well these can be
converged to reduce code.

However, it does obsolete several validation-related API *tests* that
were rolled in with API application testing: now, the entire API is
covered in a consistent way for that kind of thing: parsable JSON,
missing fields, extra fields, etc.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -47,6 +47,9 @@ Layout/SpaceInsideHashLiteralBraces:
   Exclude:
     - 'views/**/*.erb'
 
+RSpec/AnyInstance:
+  Enabled: false
+
 RSpec/ExampleLength:
   Enabled: false
 
--- a/clover.rb
+++ b/clover.rb
@@ -2,11 +2,16 @@
 
 require_relative "model"
 
+require "committee"
 require "roda"
 require "tilt"
 require "tilt/erubi"
 
 class Clover < Roda
+  OPENAPI = OpenAPIParser.load("openapi/openapi.yml", strict_reference_validation: true)
+  SCHEMA = Committee::Drivers::OpenAPI3::Driver.new.parse(OPENAPI)
+  SCHEMA_ROUTER = SCHEMA.build_router(schema: SCHEMA, strict: true)
+
   opts[:check_dynamic_arity] = false
   opts[:check_arity] = :warn
 
@@ -19,6 +24,7 @@ class Clover < Roda
   plugin :h
   plugin :hash_branches
   plugin :hash_branch_view_subdir
+  plugin :hooks
   plugin :Integer_matcher_max
   plugin :json
   plugin :invalid_request_body, :raise
@@ -115,15 +121,38 @@ class Clover < Roda
       type = e.type
       message = e.message
       details = e.details
+    when Committee::BadRequest, Committee::InvalidRequest
+      code = 400
+      type = "BadRequest"
+      message = e.message
+
+      case e.original_error
+      when JSON::ParserError
+        message = "Validation failed for following fields: body"
+        details = {"body" => "Request body isn't a valid JSON object."}
+      when OpenAPIParser::NotExistPropertyDefinition
+        keys = e.original_error.instance_variable_get(:@keys)
+        message = "Validation failed for following fields: body"
+        details = {"body" => "Request body contains unrecognized parameters: #{keys.join(", ")}"}
+      when OpenAPIParser::NotExistRequiredKey
+        keys = e.original_error.instance_variable_get(:@keys)
+        message = "Validation failed for following fields: body"
+        details = {"body" => "Request body must include required parameters: #{keys.join(", ")}"}
+      end
+    when Committee::InvalidResponse
+      code = 500
+      type = "InternalServerError"
+      message = e.message
     else
       raise e if Config.test? && e.message != "test error"
-
       Clog.emit("route exception") { Util.exception_to_hash(e) }
 
       code = 500
       type = "UnexceptedError"
       message = "Sorry, we couldn’t process your request because of an unexpected error."
     end
+
+    raise e if Config.test? && e.is_a?(Committee::Error)
 
     response.status = code
     next if code == 204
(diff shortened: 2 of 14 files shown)


.rubocop.yml
```diff
@@ -47,6 +47,9 @@ Layout/SpaceInsideHashLiteralBraces:
   Exclude:
     - 'views/**/*.erb'
 
+RSpec/AnyInstance:
```
fdr, line 50: there was not really any good way around this that I know of, so no need to say: "yeah break the rules" before breaking them, I instead remove the rule.

clover.rb
```diff
@@ -108,9 +114,33 @@ def assets
       type = e.type
       message = e.message
       details = e.details
-    else
-      raise e if Config.test? && e.message != "test error"
```
jeremyevans, line 112: We should keep this line in the else block, it makes debugging much easier.

jeremyevans (changes requested):

spec/routes/api/committee_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      def self.name
+        "CommitteeTestNestedException"
+      end
+
+      def original_error
+        Exception.new("unaddressed error value")
+      end
+    end.new("testing conversion of unenumerated nested exceptions")
+
+    allow_any_instance_of(Committee::SchemaValidator::OpenAPI3).to receive(:request_validate) do
```
jeremyevans, line 24: If you want to avoid `allow_any_instance_of`, you could add a custom hash branch only in test that raises the error, similar to what we are doing here: https://github.com/ubicloud/ubicloud/blob/main/clover.rb#L468

spec/routes/api/committee_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        boot_image: "ubuntu-jammy",
+        storage_size: "40"
+      }.to_json
+    }.to raise_error Committee::InvalidResponse, %r{#/components/schemas/Vm missing required parameters: .*}
+  end
+
+  it "fails when request is missing required keys" do
+    expect {
+      post "/project/#{project.ubid}/location/#{TEST_LOCATION}/vm/test-vm", {}.to_json
+    }.to raise_error Committee::InvalidRequest, "#/paths/~1project~1{project_id}~1location~1{location}~1vm~1{vm_name}/post/requestBody/content/application~1json/schema missing required parameters: public_key"
```
jeremyevans, line 60: I guess this `~1` stuff is committee's way of encoding `/`? I recommend switch to a regexp (ditto for next test):

```suggestion
    }.to raise_error Committee::InvalidRequest, /missing required parameters: public_key/
```

fdr:
Added a benchmark of overhead to commit message (it wasn't really measurable on a whole-suite basis)

spec/routes/api/committee_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  it "fails when request has extra unexpected keys" do
+    expect {
+      post "/project/#{project.ubid}/location/#{TEST_LOCATION}/vm/test-vm", {missing_stuff: true}.to_json
+    }.to raise_error Committee::InvalidRequest, "#/paths/~1project~1{project_id}~1location~1{location}~1vm~1{vm_name}/post/requestBody/content/application~1json/schema does not define properties: missing_stuff"
+  end
+
+  it "reports if response body is not valid JSON" do
+    allow_any_instance_of(Committee::SchemaValidator::OpenAPI3).to receive(:response_validate) do
```
jeremyevans, line 70: Avoiding `allow_any_instance_of` here could probably be done by overriding `Serializers::Vm.serialize` to return an invalid JSON string (maybe that triggers a different issue because the response header isn't set correctly).

jeremyevans (approved):
Looks good, and I'm OK with merging.  I do have some suggestions, but they are up to you.

clover.rb
```diff
@@ -2,11 +2,16 @@
...
 
+require "committee"
 require "roda"
 require "tilt"
 require "tilt/erubi"
 
 class Clover < Roda
+  OPENAPI = OpenAPIParser.load("openapi/openapi.yml", strict_reference_validation: true)
+  SCHEMA = Committee::Drivers::OpenAPI3::Driver.new.parse(OPENAPI)
+  SCHEMA_ROUTER = SCHEMA.build_router(schema: SCHEMA, strict: true)
```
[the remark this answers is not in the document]
fdr: Hmmm sometimes I like to `cd` into them with `pry`, so this is a bit easier for that, though it's definitely a minority case.

spec/routes/api/committee_spec.rb
fdr: well per minor abuse of the term, where serialize is more like "idiomatic downprojected representation of this rich object" (and leaves to_json to a final step) I don't think that'll work to get a invalid response output:

```ruby
    def self.serialize(object, options = {})
      return if object.nil?
      if object.respond_to?(:map) && !object.is_a?(Hash)
        object.map { |item| serialize_internal(item, options) }
      else
        serialize_internal(object, options)
      end
    end
```

spec/routes/api/committee_spec.rb
fdr: Maybe I can give that a try, although in some ways, I kinda like this, in the sense that the exactly one time such a test exception is injected, it's in the context of the single test that needs that; there are no others, there may never be another, and perhaps no more than one more other ever.

PR #2362, opened 2024-12-02
Have invalid project name error message explain why

jeremyevans:
Currently, the error message is the default "is invalid", which isn't that helpful to the user.

This is a bit explicit and technical, in that it mentions ASCII, but without ASCII, it implies that non-ASCII letters would be acceptable, which is incorrect.  I would guess that our users are technical enough to understand this message.

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -124,7 +124,7 @@ class Project < Sequel::Model
   def validate
     super
     if new? || changed_columns.include?(:name)
-      validates_format(%r{\A[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?\z}i, :name)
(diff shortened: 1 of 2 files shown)


model/project.rb
```diff
@@ -124,7 +124,7 @@ def create_api_key(used_for: "inference_endpoint")
   def validate
     super
     if new? || changed_columns.include?(:name)
-      validates_format(%r{\A[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?\z}i, :name)
+      validates_format(%r{\A[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?\z}i, :name, message: "must only include ASCII letters, numbers, and dashes, and must start and end with a ASCII letter or number")
```
bsatzger, line 127:
```suggestion
      validates_format(%r{\A[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?\z}i, :name, message: "must only include ASCII letters, numbers, and dashes, and must start and end with an ASCII letter or number")
```

bsatzger (approved):

PR #2367, opened 2024-12-02
Add vcpus & memory_gib to Vm; cores & memory_gib calculation to VmSize

pykello:
### Add vcpus & memory_gib to Vm 

Previously we used formulas to calculate number of vcpus & amount of memory for a VM based on number of cores it uses. Since changing formulas doesn't affect already created VMs, in this PR we materialize these values in the Vm model.

### Enhancements to VmSize.

* Add `cores` to VmSize, so its calculation is done only in one place.
* Rename `memory` to `memory_gib`, to make memory unit clear.
* Make memory calculation based on arch in `VmSize`, not in multiple
  places in code.
* Rename `vcpu` to `vcpus`.


prog/vm/nexus.rb
```diff
@@ -94,7 +94,7 @@ def self.assemble(public_key, project_id, name: nil, size: "standard-2",
       end
```
fdr: Although somewhat repetitive, the commit message

 >  Materialize Vm's memory_gib and vcpus.  

Should be expanded. It's chancy and annoying to navigate from a blame in the hope that the previous commits explain why this was done (though in this case, it does). 

An alternative: is hyperlink via commit hash, which is super awkward given rebasing. May work well if you merge the migration (giving it a permanent hash) and then edit this. That would be the most terse, though.

migrate/20241204_vm_enforce_vcpus_memory.rb
```diff
@@ -0,0 +1,26 @@
...
+        WHEN arch = 'arm64' THEN 3.2 * cores
+        WHEN family = 'standard-gpu' THEN 10.68 * cores
+        ELSE 8 * cores
+      END;
+
+    ALTER TABLE vm
+      ALTER COLUMN vcpus SET NOT NULL,
+      ALTER COLUMN memory_gib SET NOT NULL;
+
+    SQL
```
fdr, line 24: This one should change the cache a second time, right? That's where the NOT NULL comes in.

Also in this case you probably want an `up` block, `change` blocks are really for sequel auto-generating the up/down migration.

fdr (changes requested):
Minor edits, I appreciate the omnibus rename-stuff patch to keep hunks smaller/less intellectually interfering for future reviews.

migrate/20241204_vm_enforce_vcpus_memory.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    # Update the memory & vcpus for exisitng VMs, to match the logic from the
+    # code. Then enforce the NOT NULL constraint on the columns.
+    run <<-SQL
```
jeremyevans, line 7: Sequel does not support multiple queries inside a single `run` command.  Whether it works depends on the adapter (so I assume it works with the postgres adapter), but I would advise against such use.  Use separate `run`  methods for the two statements.

jeremyevans (commented):
Looks OK to me, other than the issues @fdr noted and an issue in the second migration.

migrate/20241204_vm_enforce_vcpus_memory.rb
macieksarnowicz: I think that means that those two migrations should be put as separate PRs.

model/vm.rb
```diff
@@ -67,17 +67,6 @@ def display_state
     super
   end
 
-  def mem_gib_ratio
```
macieksarnowicz, line 70: I think this is still used in later commits.

migrate/20241204_vm_enforce_vcpus_memory.rb
macieksarnowicz: @pykello - also, since you split it into separate transactions/commits, I think the UPDATE statement should only touch the columns that are NULL.

model/vm.rb
macieksarnowicz: Although, I was thinking of moving it to VmHost class, as it is no longer needed in Vm. It is used when allocating a VmHostSlice, as we cannot use @request.memory because that is a size of memory for a Vm not for the slice.

model/vm.rb
```diff
@@ -282,6 +282,8 @@ def storage_secrets
 #  arch           | arch                     | NOT NULL DEFAULT 'x64'::arch
 #  allocated_at   | timestamp with time zone |
 #  provisioned_at | timestamp with time zone |
+#  vcpus          | integer                  | NOT NULL
+#  memory_gib     | integer                  | NOT NULL
```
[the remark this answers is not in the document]
pykello: Now this "" in the first commit and "NOT NULL" in the 3rd commit.

prog/vm/nexus.rb
pykello: Expanded the commit message.

migrate/20241204_vm_enforce_vcpus_memory.rb
pykello: > Also in this case you probably want an up block, change blocks are really for sequel auto-generating the up/down migration.

@fdr Changed to `up`. Also included a `down` block.

> I think that means that those two migrations should be put as separate PRs.

@macieksarnowicz I'm not sure why? Migrations that can be deployed together are usually grouped in the same commit.

Applied the remaining points.

migrate/20241204_vm_enforce_vcpus_memory.rb
pykello: done

model/vm.rb
pykello: I saw your 2nd comment after I applied your 1st comment. For the 2nd comment, let's move it to `VmHost` in the other PR which allocates a VmHostSlice so we see how it's used.

jeremyevans (commented):
Changes look good to me.

macieksarnowicz:
Just a general comment to all this:
- cores is the unit of allocation at the host level (we add and subtract cores from the host)
- vcpus is the amount of CPUs (threads) allocated to a VM and specified to the cloudhypervisor
- cpu_percent_limit (that will be added later) is the CPU percentage cap put on top of the number of vcpus
- The cores to vcpus ratio is typically set by the architecture - 2:1 for x64, 1:1 for amd64, but because we specify it now in VmSize structure, it does not have to be fixed like that.
- in the future, we can have Vm families, defined in VmSize, where the number of cores will represent the cores allocated to a VmHostSlice (and removed from the VmHost), and the number of vcpus will be the number given to the Vm and they may have a different ratio than 2:1
- VmSize specifies the amount of memory given to the Vm
- The amount of memory for a Slice will be the memory_ratio * cores allocated for the Vm

migrate/20241204_vm_enforce_vcpus_memory.rb
macieksarnowicz: I am assuming the cache has to be updated for each migration separately, that's why.

migrate/20241204_vm_enforce_vcpus_memory.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+Sequel.migration do
```
fdr, line 3: commit message typo: `memroy_gib`  in ` Enforce vcpus & memroy_gib to be set for Vm. `

PR #2376, opened 2024-12-04
add pattern for project name to match #2326

geemus:


fdr (approved):

fdr:
Did some minor gussying up in #2659

PR #2379, opened 2024-12-04
Combine page_header and breadcrumb components

enescakir:
- **Combine page_header and breadcrumb components**
  We use them together in the same way in all pages, so it makes sense to
  combine them into a single component.
  

- **Use page_header component in PG list page**
  We used a customized version of the page_header component because it had
  a dropdown as the right item.
  
  However, we can use the standard page_header component with a dropdown
  as the right item.
  

- **Use new page_header component in PAT page**
  

- **Make all page headers sticky**
  Before, we only made page headers sticky for PostgreSQL pages. Now, we
  can make all page headers sticky.
  

- **Move GitHub page header to tabbar component**


views/components/page_header.erb
```diff
@@ -2,7 +2,7 @@
 breadcrumbs ||= []
 back ||= breadcrumbs.dig(-2, 1) || "#" %>
 
-<div class="space-y-1">
+<div class="space-y-1 sticky -mx-10 mb-0.5 px-10 top-[2.9rem] bg-gray-50 z-30">
```
jeremyevans, line 5: My preference would be to not use sticky page headers, but I don't feel strongly about it.

jeremyevans (approved):

views/components/page_header.erb
enescakir: We can decide to remove the sticky page headers after we unify them. As it stands, some pages are sticky while others are not.

PR #2381, opened 2024-12-04
Introduce new project dashboard

enescakir:
### Add github_runners association to project

The project has GithubRunner models through GitHubInstallation models.

It couldn't setup it with `one_through_many` association, so I used
`many_through_many`.

### Introduce new project dashboard

The project dashboard is the first thing users see when they create a
new account. It was designed in the early days and hasn't been updated
since.

We've overlooked the dashboard, despite it being one of the most crucial
pages.

Redesigning it with UX principles in mind is a significant task.

I've just introduced the new project dashboard. I haven't spent a lot of
time on it yet, but we can certainly enhance it over time.

Even though it's a preliminary version, I believe it's an improvement
over the previous one and a good starting point.

| Before | After |
|--------|-------|
| before | after |

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -8,6 +8,7 @@ class Project < Sequel::Model
   one_to_one :billing_info, key: :id, primary_key: :billing_info_id
   one_to_many :usage_alerts
   one_to_many :github_installations
+  many_through_many :github_runners, [[:github_installation, :project_id, :id], [:github_runner, :installation_id, :id]]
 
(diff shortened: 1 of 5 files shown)


jeremyevans (approved):
Looks great!

`many_through_many` is the correct choice. `one_through_many` would only be used if you were trying to return a single github runner per project, and not all github runners for the project.

enescakir:
I made some improvements based on feedback from our designer.

CleanShot 2024-12-13 at 11 43 35@2x

PR #2384, opened 2024-12-05
Refactor Authorization

jeremyevans:
At a high level, this replaces the AccessPolicy/AccessTag/AppliedTag authorization system with a AccessControlEntry/SubjectTag/ActionTag/ObjectTag/ActionType system.  Fundamental differences:

* Instead of AccessPolicy objects containing a JSON blob with multiple subjects, actions, and objects, AccessControlEntry objects are small, allow access for a single subject, action, and object.
* Each subject, action, and object in an AccessControlEntry can be a tag that references arbitrary subjects, actions, and objects.  AccessPolicy bodies can have multiple entries in them, but even the most complex AccessPolicy body entry can be replaced with a single AccessControlEntry, by creating a tag for the subjects, a tag for the actions, and a tag for the objects.
* AccessControlEntry action and object can be nil, to allow all actions and all objects, respectively. The default access control for a project is a currently a single Admin subject tag, in which the creating user is placed, and a single AccessControlEntry granting that subject tag access to all actions on all objects.
* Available actions are now fixed, and stored in an ActionType table.  The equivalent to `Vm:*` style permissions are global action tags (managed by Ubicloud).
* Tags can include other tags, recursively.
* Add action types for creating each tag type, and managing each tag type.
* There is a simple UI for users to manage the access control entries and tags.
* Use newly added access permissions/action types to authorize token/ACE/tag management, instead of using `Project:user` for everything.
* There is a simple UI for users to manage token permissions.
* Support converting AccessPolicy entries to *Tag and AccessControlEntry rows.

This removes AccessPolicy and AppliedTag as they are no longer needed.  AccessTag is still used, because it was doing double duty both as authorization, and for storing how objects are related to projects (the hyper_tag_id).  AccessTag is now only used for the hyper_tag parts.

Project invites now set the subject tag the user is placed in.  This is currently done by name and not UUID, but that can be changed later.

See individual commits for more detail.

All expected parts are now ready.  I'm keeping this draft for now, but I think it will need to be committed in a few separate chunks.


migrate/20241203_authorization_refactor.rb
```diff
@@ -0,0 +1,80 @@
...
+    # Action Type UUID primary keys created via:
+    # UBID.from_parts(i, UBID::TYPE_ACTION_TYPE, 0, i).to_uuid
+    from(:action_type).import([:id, :name],
+      [["00000000-0001-835a-8000-000000000001", "Project:delete"],          # 1
+        ["00000000-0002-835a-8000-000000000002", "Project:edit"],
+        ["00000000-0003-835a-8000-000000000003", "Project:view"],
+        ["00000000-0004-835a-8000-000000000004", "Project:user"],
+        ["00000000-0005-835a-8000-000000000005", "Project:billing"],
+        ["00000000-0006-835a-8000-000000000006", "Project:github"],
+        ["00000000-0007-835a-8000-000000000007", "Vm:create"],
```
fdr, line 18: oh man, it's going to bother me for no good reason in particular when we add another project action (or whatever) and it is numerically non-contiguous

Rakefile
```diff
@@ -100,14 +100,15 @@ end
 
 desc "Refresh schema and index caches"
 task :refresh_sequel_caches do
-  %w[schema index].each do |type|
+  %w[schema index static_cache].each do |type|
```
fdr, line 103: til

migrate/20241203_authorization_refactor.rb
jeremyevans: We don't have to use sequential integer values.  We could have "Project"` actions be 1xxx, "Vm" actions be 2xxx, etc.  Let me know if you prefer that approach.

byucesoy:
I didn't look into the PR yet, just wanted to leave [this](https://github.com/ubicloud/ubicloud/pull/1978#pullrequestreview-2347763318) here as some of my thoughts on the authorization. Feel free to incorporate the ideas that you find meaningful.

jeremyevans:
> I didn't look into the PR yet, just wanted to leave [this](https://github.com/ubicloud/ubicloud/pull/1978#pullrequestreview-2347763318) here as some of my thoughts on the authorization. Feel free to incorporate the ideas that you find meaningful.

I reviewed those comments before I started work on this PR, and they were helpful in formulating ideas.  This PR goes a bit further, it aims to be the Authorization v2 as mentioned in your comment.

For reusing lists of actions across projects, I plan to implement that via ActionTags with a NULL project_id (currently, action_tag.project_id is NOT NULL).  These will be static and managed by the application (i.e. Ubicloud developers), as opposed to the user.

I do not plan to add exclusion rules.  Users that want can add an ActionTag with all actions except the actions they do not want.  If we add new actions that they want to add to the action tag, they'll have to do so manually.  This is a fail-closed model.  If there is sufficient desire across our userbase for something similar to `everything_but_vm` rules, we could implement those as ActionTags with NULL project id, and manage them ourselves.

For naming, I kept the subject/action/object naming.  subject can be thought of as role/user/token . SubjectTag could be called Role, though I think Role would be overly specific, and for consistency, I named it SubjectTag for similarity to ActionTag and ObjectTag.  I did drop the policy naming, the rules are now access control entries.  The "access policy" for a project would be the list of its access control entries.

This doesn't implement attributes yet, by design. We don't currently have attributes, and I think it is best to wait to determine what attributes we actually need.  Attributes will likely be implemented via a single column added to the access_control_entry table, but I'm not yet sure what form that will take (could be `jsonb`, could be `text[]`, maybe something else?).

fdr:
@enescakir can you take pole position on reviewing this? I'll also review it, but mostly for my edification, but you can engage this at a deeper level than I could.

migrate/20241203_authorization_refactor.rb
fdr: How about encoding them look good in base32/ubid? like `vm0create` (other separator suggestions welcome, the 0, is an homage to null termination, but may also be one of the weird characters to prevent illegible ubids re: 0/O)

jeremyevans: That's a great idea, it will debugging cases with pry a lot easier.  I'll work on that.

fdr: Hmm I think I may have erred by thinking of  `0`, because of how "o" folds into it, and "o" is a vowel...that is, relatively common. Do any other characters stand out better as delimiters vs data? Say, a rare one, like "z". I'm also not sure what works best visually.

fdr: But possibly, it doesn't get better than this. As I'm squinting at it it occurred to me. I also wonder if trying to eliminate entropy by having exactly one copy of the in-band signaling is superior, e.g.

Original: 

`tt0ps0c0nn021g0ps0c0nnect1`

zeroed:

`tt0000000000000ps0c0nnect1`

For a quick preview of `z`
`ttzzzzzzzzzzzzzpszc0nnect1`

I suppose maybe `0` is good in spite of being seen in some data, because it's a tall character. It stands out.

And the couple of bytes that ubids needs to be uuids will not be quite as harmonious as all-zeros.

There's also the possibility of shifting the in-band data to the left to pad it from the parity bits at the end, e.g. more like:

`tt0ps0c0nnect000000000001`

What do you think? I think talked (or rather, read) myself out of z-padding but curious what you think of decreasing repetition and effect on legibility, and left-vs-right alignment.

jeremyevans: > But possibly, it doesn't get better than this. As I'm squinting at it it occurred to me. I also wonder if trying to eliminate entropy by having exactly one copy of the in-band signaling is superior, e.g.
> 
> Original:
> 
> `tt0ps0c0nn021g0ps0c0nnect1`
> 
> zeroed:
> 
> `tt0000000000000ps0c0nnect1`

This is an invalid ubid:

```ruby
udec("tt0000000000000ps0c0nnect1")
UBIDParseError: Invalid bottom bits parity
```

> For a quick preview of `z` `ttzzzzzzzzzzzzzpszc0nnect1`

This is also an invalid ubid:

```ruby
udec("ttzzzzzzzzzzzzzpszc0nnect1")
UBIDParseError: Invalid top bits parity
```

> I suppose maybe `0` is good in spite of being seen in some data, because it's a tall character. It stands out.
> 
> And the couple of bytes that ubids needs to be uuids will not be quite as harmonious as all-zeros.
> 
> There's also the possibility of shifting the in-band data to the left to pad it from the parity bits at the end, e.g. more like:
> 
> `tt0ps0c0nnect000000000001`
> 
> What do you think? I think talked (or rather, read) myself out of z-padding but curious what you think of decreasing repetition and effect on legibility, and left-vs-right alignment.

Personally, I don't have strong feelings either way.  However, there is less room at the left side than the right side, as the middle of the UBID requires certain things. So if you are going to only have it in a single place, it would be better to have it on the left side.  I did it in both places to make it possible to read from either side, though the 

Here's how it could look on the right hand side only:

```
tt00000000021g0ps0c0nnect1
ttzzzzzzzz021g0ps0c0nnect1
```

Another option would be putting part on the left and part on the right.  This would allow for  longer names for types, instead of the character abbreviation:

```
ttpr1vsvbn020g0zzzc0nnect0
ttpr1vsvbn020g0d1sc0nnect0
```

What would you prefer?

fdr: to me, the right-hand-side only  tracks better. Don't we have a list of idiomatic abbreviations (the first two bytes) for nearly everything in ubids anyway? This could be the rule for how to abbreviate the object type.

I kinda like your idea of a hybrid, with a bunch of "z"s for padding and `0` for delimiter:

> ttzzzzzzzz021g0ps0c0nnect1

jeremyevans: Change made: https://github.com/ubicloud/ubicloud/commit/c5ccc981bee4a9714704fcc8c583c778b905e14b#diff-b6e51bff148ac6a959b8da686a14246830c7f6f193704f16f848753cb2b088d0R12

A nice result of this is the UUIDs are also recognizable as different from random UUIDs (e.g. `ffffffff-ff00-835a-802d-903439602b50`).

I noticed the ActionType/ActionTag vanity prefixes do not match the UBID prefixes in some cases:

* Project: `pr` instead of `pj`
* SubjectTag: `st` instead of `ts`
* ActionTag: `at` instead of `ta`
* ObjectTag: `0t` instead of `t0`

If you'd like the prefixes to be made consistent, let me know and I can make that change.

fdr:
Is there a good reason to be inconsistent? Seems like two different two
byte abbreviations is the kind of thing that makes the next generation
wonder “what were they thinking?”

migrate/20241203_authorization_refactor.rb
jeremyevans: I also realized I only used `z` on the left side, the right side is still padded with `0` (or `1` if the first character in the vanity is `0`).  Please let me know if you would like to change that.

jeremyevans: Upon further review, I can convert two additional bytes from `0` to `z`:

```
tazzzzzzzzz20gzps0d1sc0nn1
tazzzzzzzzz20gzzzzps0v1ew0
```

Should I use that format?

jeremyevans: Or maybe it would be preferable to always have the type in the same position?:

```
tazzzzzzzzz20qzd1sc0nn0ps0
tazzzzzzzzz20qzzzzv1ew0ps1
```

fdr: Interesting idea. I think I'm mostly out of good ideas now, it does seem a bit weird to have to read right to left; the type bytes are themselves a bit familiar and help me track the progression.

I do think keeping it as `0` on the right-hand-side works pretty well, in that I think long runs of `0` dilute their value as a (unusually tall) separator, whereas `z` help you zoom past the boring parts (it's not beyond me that `z` is in `zoom` as well as `zzz` is sometimes a reaction to boring things)

fdr: that's to say, personally, I like `ttzzzzzzzz021g0ps0c0nnect1` (with the leading `0` before `ps` to separate)

jeremyevans: OK.  Changes made: https://github.com/ubicloud/ubicloud/commit/d3f0c03e40d68b07efdd49ed92038016e06e0f87#diff-b6e51bff148ac6a959b8da686a14246830c7f6f193704f16f848753cb2b088d0R13

So here are the rules as I understand them:

* First 2 characters: UBID type (required), either `tt` or `ta` currently
* Next 12 characters:  `zzzzzzzz021g`
* 0-9 padding characters of `z`
* `0` delimiter if there is a type
* 2 characters for embedded type (UBID type, e.g. `pj` for Project), if there is a type
* `0` delimiter
* 1-7 characters for action/tag name
* Last character: `0` or `1` (depending on parity)

Examples:

* PrivateSubnet:disconnect: `ttzzzzzzzz021g0ps0d1sc0nn0`
* Project:view: `ttzzzzzzzz021gzzz0pj0v1ew0`
* Postgres:Firewall:edit: `ttzzzzzzzz021gzzz0pf0ed1t1`
* Member global tag: `tazzzzzzzz021gzzzz0member0`

Note that `Postgres:Firewall` action types use the `pf` prefix, because they need to be different than `Postgres` actions (which use `pg`), but the UBID `pf` prefix is used for Postgres::FirewallRule, which is slightly different.  Authorizations for `Postgres:Firewall` actions appear to be against `pg.id`, not the id for specific firewall rules.  Maybe it would be better to change:

* `Postgres:Firewall:edit` to `Postgres:fwedit`
* `Postgres:Firewall:view` to `Postgres:fwview`

What do you think?

jeremyevans: I made the Postgres:Firewall permission name changes.  One of the commits changes the application to use the new names.

migrate/20241203_authorization_refactor.rb
```diff
@@ -0,0 +1,135 @@
...
+    # This shows the generated UBIDs, but those are removed before inserting.
+    # These non-project specific action tags will be managed by Ubicloud developers.
+    action_tags =
+      [["ffffffff-ff00-834a-87ff-ff82d2028210", "Project:all", "tazzzzzzzz021gzzzz0pj0a110"],
+        ["ffffffff-ff00-834a-87ff-ff8374028210", "Vm:all", "tazzzzzzzz021gzzzz0vm0a111"],
+        ["ffffffff-ff00-834a-87ff-ff82d9028210", "PrivateSubnet:all", "tazzzzzzzz021gzzzz0ps0a111"],
+        ["ffffffff-ff00-834a-87ff-ff81fc028210", "Firewall:all", "tazzzzzzzz021gzzzz0fw0a110"],
+        ["ffffffff-ff00-834a-87ff-ff802b028210", "LoadBalancer:all", "tazzzzzzzz021gzzzz01b0a111"],
+        ["ffffffff-ff00-834a-87ff-ff82d0028210", "Postgres:all", "tazzzzzzzz021gzzzz0pg0a111"],
+        ["ffffffff-ff00-834a-87ff-ff828ea2dd80", "Member", "tazzzzzzzz021gzzzz0member0"]].each(&:pop)
```
enescakir, line 90: We've `Admin` as the subject tag but `Member` as the action tag. It's little confusing for me. I believe we should also create `Member` as a subject tag and create it in the projects by default. Most users won't interact with our access control view or take the time to understand it. They just want to use it, which is why we introduced `admin` and `member` managed policies.

lib/authorization.rb
```diff
@@ -100,44 +100,17 @@ def hyper_tag(project)
     def associate_with_project(project)
```
enescakir, line 173: > Authorization::HyperTagMethods is staying, because that is how
objects are associated to projects, unrelated to
authorization/access control.

In version 1, we designed the project-resource association coupled with authorization. If version 2 doesn't accommodate this, I suggest we completely remove the `AccessTag`. It only seems to complicate matters. The `PostgresResource` has a `project_id` column, which is unique compared to other resources. @byucesoy wanted to experiment with having a `project_id` column instead of establishing an association through authorization. @fdr was willing to associating a single resource with multiple projects feature such as migration resources between projects. However, it appears we don't currently have a plan for a feature that leverages a single resource across multiple projects. Therefore, in my opinion, we can simplify the things by adding a `project_id` column to all resources and completely eliminate the `AccessTag`. This approach aligns with what @byucesoy implemented for `PostgresResources`, if @fdr has no objections.

migrate/20241203_authorization_refactor.rb
```diff
@@ -0,0 +1,135 @@
...
+        ["ffffffff-ff00-835a-87c0-15819872b4e0", "LoadBalancer:create", "ttzzzzzzzz021gz01b0create1"],
+        ["ffffffff-ff00-835a-87c0-1581ae0bb4e0", "LoadBalancer:delete", "ttzzzzzzzz021gz01b0de1ete0"],
+        ["ffffffff-ff00-835a-87ff-f005607343a0", "LoadBalancer:edit", "ttzzzzzzzz021gzzz01b0ed1t0"],
+        ["ffffffff-ff00-835a-87ff-f00560d85dc0", "LoadBalancer:view", "ttzzzzzzzz021gzzz01b0v1ew1"],
+        ["ffffffff-ff00-835a-87c1-68019872b4e0", "Postgres:create", "ttzzzzzzzz021gz0pg0create1"],
+        ["ffffffff-ff00-835a-87c1-6801ae0bb4e0", "Postgres:delete", "ttzzzzzzzz021gz0pg0de1ete0"],
+        ["ffffffff-ff00-835a-87ff-f05a007343a0", "Postgres:edit", "ttzzzzzzzz021gzzz0pg0ed1t0"],
+        ["ffffffff-ff00-835a-87ff-f05a00d85dc0", "Postgres:view", "ttzzzzzzzz021gzzz0pg0v1ew1"],
+        ["ffffffff-ff00-835a-87c1-6801fc7343a0", "Postgres:fwedit", "ttzzzzzzzz021gz0pg0fwed1t1"],
+        ["ffffffff-ff00-835a-87c1-6801fcd85dc0", "Postgres:fwview", "ttzzzzzzzz021gz0pg0fwv1ew0"],
```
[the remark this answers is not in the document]
enescakir: @byucesoy, what's your opinion on this?

migrate/20241203_authorization_refactor.rb
```diff
@@ -0,0 +1,135 @@
...
+        ["ffffffff-ff00-835a-87ff-ff8359029ad0", "SubjectTag:add", "ttzzzzzzzz021gzzzz0ts0add1"],
```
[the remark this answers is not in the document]
jeremyevans: This is different from `create`.  This is to add members to the tag, not to create the tag.

model/access_control_entry.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative "../model"
+
+class AccessControlEntry < Sequel::Model
```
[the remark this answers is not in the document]
jeremyevans: The reason for the combined commit is the ubid code is broken if you introduce the `TYPE_*` constants before the models, and in order to generate the static ubids in the migration, the `TYPE_*`  constants were added. It looks like the vanity ubid code doesn't need the type constant, so it's possible that commit could be split, even though it would look strange as the ubids generated in the migration would be using prefixes not yet added to the ubid code.  Do you want me to split that commit?

.by-session-setup.rb
```diff
@@ -12,20 +12,29 @@
 enc/trans/single_byte
 enc/trans/utf_16_32
 enc/utf_16be
+enc/utf_16le
 rubygems/openssl
 rubygems/package
 rubygems/security
+ripper
+coderay/scanners
+coderay/scanners/ruby
```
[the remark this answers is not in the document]
jeremyevans: It was added back in October: b3cf8d685c3bd70704893f454c1897a491e1177f

lib/authorization.rb
jeremyevans: I strongly feel that most things that most resources that currently hypertag projects should use a `project_id` column instead. Having anything other than 1 project associated for these resources seems wrong.  Having zero projects seems like a leak, and having multiple projects doesn't make sense to me. You can migrate/transfer a resource between projects by changing the `project_id` value.

The one except is users <-> projects, which is a many-to-many relationship.  For that reason, you would still need AccessTag, though maybe it could be renamed and use actual foreign keys.

This PR doesn't include such changes, because they are not technically related to authorization.

jeremyevans:
> Thank you for all your work. I am still reviewing them. In terms of general feedback, I appreciate individual commits that build upon each other. However, this PR has 71 commits, which makes it a bit difficult to review. Even go through each commit one by one without reviewing takes some time 😄

I understand that.  This is a large change, and touches many parts of the application.  We need to make sure that all parts work together, which is why this is in a single PR.  While I think it likely will be committed in 2-3 phases, it probably does need to be reviewed all at once, because if there are any changes that need to be made, we need to make sure that the entire system works correctly after the changes.

> Some commits merely address issues introduced in other commits within the same PR, while others simply refactor the code introduced in another commit within the same PR. This may be entirely personal, but I prefer to not see the initial version as I review each commit individually. In the first commit, I understand and accept the concept. Then, after several commits, it's completely refactored, and I have to familiarize myself with the new version.

I tried to rebase this continually as I was developing it, to mitigate this issue.  Maybe I didn't do a good enough job.  I can say this issue would be much worse if I didn't do that.

Sometimes, rebasing resulted in significant conflicts, so I didn't rebase in those cases.  I can give another shot at rebasing to combine related commits if you want.  However, I want to make sure I don't combine things you have already reviewed with things you haven't already reviewed, as otherwise, you would likely need to re-review things very similar to what you have already reviewed.

jeremyevans:
I rebased this against master to fix conflicts, then spent significant time rebasing to combine related commits, in order to make review easier.

I needed to modify the authorization migration to add new action types that were added since I started working on this.  As other migrations have been committed since I started work on this, I renumbered the migrations in this PR.  I put the simple migration to create an index first (20241214), and the larger migration second (20241215).  If you've already migrated using this PR, migrate down first to version 20241202 before switching branches, switch to this branch, then migrate back up.

All commits before the vanity UBID one are unrelated to the authorization refactor.  Some are independent on these changes, and are issues I found while developing this PR.  Others fix issues and this PR depends on those issues being fixed.  If it helps, I can submit a separate PR for those commits, which could be committed first.  That may make this PR slightly easier to review.  Please let me know if you would like me to do that.

jeremyevans:
I don't think there is one specific area causing conflicts that is likely to repeat going forward.  In terms of splitting this up for commit, currently, the first 14 commits could be committed without the rest. This would include the vanity UBID code and new access_tag index.  Everything else depends on the authorization migration.  While the authorization migration does not change tables, and therefore would be safe to run, I don't think it makes sense to commit it until this PR has been fully reviewed.

This did take a long while to rebase against main properly.  I ran into the following issues:

* Capybara page.refresh (which many of the new specs use) did not work with the Rack::Lint addition
* The removal of :selected support in the checkbox input took a while to discover.
* `test_up` in the Rakefile was causing problems because it was trying to use `shellwords` (I added `shellwords` to the Gemfile to work around)
* We've had 3 separate new calls to `Dataset#authorize` instead of using `Clover#dataset_authorize`, which needed to be fixed to ensure the token permissions are respected (caught due to the arity change in this commit).

fdr:
This one looks like the real bad one that'll introduce a lot of drag:

> We've had 3 separate new calls to Dataset#authorize instead of using Clover#dataset_authorize, which needed to be fixed to ensure the token permissions are respected (caught due to the arity change in this commit).

Is it practical to introduce a minimal patch to change the notation to `Clover#dataset_authorize` to be convergent with this patch, even while not swapping out the mechanism?

The main reason to not do it is if the programming methodology/interface is likely to change...but it looks pretty inoffensive to me at first glance (@enescakir should check). In principle, we might have written it that way to begin with, so you could see it as a value-neutral or even value-adding patch on its own.

fdr:
[In general, one of the reasons I wrote prototypes far ahead of the status quo (though, this is far more than a prototype quality) is so that a series of apparently value-neutral or minimum-value changes can be merged early once we have some confidence they're going somewhere.]

jeremyevans:
> This one looks like the real bad one that'll introduce a lot of drag:
> 
> > We've had 3 separate new calls to Dataset#authorize instead of using Clover#dataset_authorize, which needed to be fixed to ensure the token permissions are respected (caught due to the arity change in this commit).
> 
> Is it practical to introduce a minimal patch to change the notation to `Clover#dataset_authorize` to be convergent with this patch, even while not swapping out the mechanism?

Note that this issue is not related to this PR, it just happened to be caught by it.  To prevent this type of issuing going forward, we should make `Dataset#authorize` private (and possibly rename it).  No code in Clover should be calling Dataset#authorize directly, except the helper code, as otherwise it is very unlikely it will handle token permissions correctly (again, irrespective of this PR).

> The main reason to not do it is if the programming methodology/interface is likely to change...but it looks pretty inoffensive to me at first glance (@enescakir should check). In principle, we might have written it that way to begin with, so you could see it as a value-neutral or even value-adding patch on its own.

One of nice things is that all of the refactoring I did leading up to this PR means almost no code in the routes needed to be changed even though the authorization code was pretty much rewritten.  The only changes to routes in this PR, other than the project/user route:

* Add @project instance variable to github routes so the helpers know the related project
* Skip setting @project_permissions for API requests, which do not use it
* Rename `Postgres:Firewall:{view,edit}` permissions

jeremyevans:
>  No code in Clover should be calling Dataset#authorize directly, except the helper code, as otherwise it is very unlikely it will handle token permissions correctly (again, irrespective of this PR).

I added a commit to this PR to replace Dataset#authorized with Authorization.dataset_authorize. This should hopefully prevent issues where other developers use Dataset#authorized due to muscle memory, instead of Clover#dataset_authorize.

model/access_control_entry.rb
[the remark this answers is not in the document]
enescakir: Will this code fail when we deploy it, until we run the migration? If it will fail, yes we should split it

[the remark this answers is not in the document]
jeremyevans: OK.  I'll split the migration file into a separate commit.

migrate/20241203_authorization_refactor.rb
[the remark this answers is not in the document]
byucesoy: Yeah, it doesn't look good. `Postgres:firewall_edit` maybe?

Actually, thinking more on this... Do we really need separate permission for these? We can use `Postgres:edit` and `Postgres:view` directly. I don't think we need to provide access control in this granularity.

migrate/20241203_authorization_refactor.rb
jeremyevans: > 1. When inviting a new user, we should set 'admin' or preferably 'member' as the default role, similar to our previous system. Inviting a new user without any access is not a common practice.

The UI allows inviting a new user into any subject tag the inviting user has access to.  That would work for this, assuming a Member subject tag is added).

> 2. We could add a 'Subject/Role' selector to the 'User Management' page. This way, most users can choose from predefined roles (admin, member) without having to visit the access control page.

The issue with this is that accounts can be in multiple subject tags.  This type of selector only works if a user can only be in a single subject tag.  I'm fairly sure the current support for this with managed policies is broken, because changing a user from Admin to Member does not remove admin access, it just adds a new policy with member access.

> 3. Creating a 'Member' role for every project might be excessive. Instead, we could create it only when a customer assigns it to a project user, mirroring our current behavior.

Automatically creating a subject tag if it doesn't exist requires dealing with race conditions.  Having a choice between the two, I would prefer always generating a Member subject tag, and treat it like the Admin subject tag (a single ACE for Member subject tag, Member action tag, on all objects), not user-editable.

lib/authorization.rb
byucesoy: I'm supportive of having `project_id` column in the resources instead of having association. I think at some point, we were considering adding one resource to multiple projects to enable complex management options, but we didn't pursue that idea. Currently, we already have code in various places doing `projects.first`, which assumes there is only one project.

migrate/20241203_authorization_refactor.rb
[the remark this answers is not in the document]
jeremyevans: Combining these permissions would avoid the problem, so if there is no significant advantage to separate permissions, that is the approach I would recommend.  I'll update this PR to do that.

lib/authorization.rb
jeremyevans: OK.  I can work on a PR to do that after this PR is merged, unless someone else would prefer to work on it.

jeremyevans:
I split the migration into a separate commit, before the model commit.  I also made changes to remove the `Postgres:Firewall:{view,edit}` permissions and replace them with `Postgres:{view,edit}` permissions.

migrate/20241203_authorization_refactor.rb
enescakir: > The issue with this is that accounts can be in multiple subject tags. This type of selector only works if a user can only be in a single subject tag. I'm fairly sure the current support for this with managed policies is broken, because changing a user from Admin to Member does not remove admin access, it just adds a new policy with member access.

Actually, it should rewrites all managed policies and remove from the admin policy.
https://github.com/ubicloud/ubicloud/blob/ed2a77722924583ae1187eeb417f8c8169a75963/routes/project/user.rb#L119C1-L130C14

> Automatically creating a subject tag if it doesn't exist requires dealing with race conditions. Having a choice between the two, I would prefer always generating a Member subject tag, and treat it like the Admin subject tag (a single ACE for Member subject tag, Member action tag, on all objects), not user-editable.

When I looked into our database, we have 1.581 admin policies and 28 member policies. So, creating a member tag for every project could be excessive. Not sure should we have it at all or not.

lib/access_control_model_tag.rb
```diff
@@ -101,8 +101,10 @@ def check_members_to_add(to_add)
   end
 
   def before_destroy
+    meta_cond = {object_id: respond_to?(:metatag_uuid) ? metatag_uuid : id}
     applied_dataset.where(tag_id: id).or(applied_column => id).delete
-    AccessControlEntry.where(applied_column => id).destroy
+    DB[:applied_object_tag].where(meta_cond).delete
+    AccessControlEntry.where(applied_column => id).or(meta_cond).destroy
     super
   end
```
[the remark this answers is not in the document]
jeremyevans: You are correct.  The `ObjectTag#before_destroy` can be removed.  I'll push a commit for that.

migrate/20241203_authorization_refactor.rb
jeremyevans: > > The issue with this is that accounts can be in multiple subject tags. This type of selector only works if a user can only be in a single subject tag. I'm fairly sure the current support for this with managed policies is broken, because changing a user from Admin to Member does not remove admin access, it just adds a new policy with member access.
> 
> Actually, it should rewrites all managed policies and remove from the admin policy. https://github.com/ubicloud/ubicloud/blob/ed2a77722924583ae1187eeb417f8c8169a75963/routes/project/user.rb#L119C1-L130C14

Apologies, that does work correctly.  It's the token code I wrote that doesn't handle the situation correctly (looks like due to the `unless tokens.empty?` part).

> > Automatically creating a subject tag if it doesn't exist requires dealing with race conditions. Having a choice between the two, I would prefer always generating a Member subject tag, and treat it like the Admin subject tag (a single ACE for Member subject tag, Member action tag, on all objects), not user-editable.
> 
> When I looked into our database, we have 1.581 admin policies and 28 member policies. So, creating a member tag for every project could be excessive. Not sure should we have it at all or not.

I recommend not having it.  It wouldn't save much time for users that want to restrict access to projects.

views/project/user.erb
```diff
@@ -84,15 +81,7 @@
...
-                <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
-                  <%== render(
-                    "components/form/policy_select",
-                    locals: {
-                      name: "user_policies[#{user[:ubid]}]",
-                      selected: @user_policies[user[:ubid]]
-                    }
-                  ) %>
-                </td>
+                <td></td>
```
enescakir, line 84: `Subject Tag` column is empty

jeremyevans: This is expected.  You cannot modify the subject tags for existing users in this UI, only the subject tag an invited user will be placed in.  This is because existing users can be in multiple subject tags, and the UI doesn't offer a way to handle that correctly.

migrate/20241203_authorization_refactor.rb
jeremyevans: OK, then it makes sense to add the Member role by default for projects, and an access control entry for the Member subject tag, Member action tag, and NULL object tag (members can take all allowed member actions on any object in the project).  That won't cause significant complexity if we treat the subject tag and access control entry just as if the user created them (different from the Admin access control entry, which is not modifiable by the user).  I'll work on that.

Note that the previous flow (before managed policies) was too complex, because it required writing a JSON policy to express what you want.  This gives users a UI to manage the security permissions, which is much easier on the users (especially with #2458).

jeremyevans: @ozgune @enescakir I pushed a commit that adds a Member subject tag by default for projects, and added a default access control entry that allows members of that tag access to the same permissions as the current Member access policy.  This allows users to invite users directly into the Member tag, without having to use the access control system directly.

jeremyevans:
I've rebased on main to fix the conflicts and remove already merged commits.

jeremyevans:
I've rebased this, updated the converter to be better (still needs more work), and moved some things around.  I've broken out a set of commits that can be merged before the authorization system switchover in PR #2470.

migrate/20241203_authorization_refactor.rb
enescakir: @jeremyevans What do you think about adding subject selector to the user list if it has zero or one subject tag?

https://github.com/ubicloud/ubicloud/pull/2476

views/project/user.erb
```diff
@@ -77,54 +77,48 @@
           </thead>
           <tbody class="divide-y divide-gray-200 bg-white">
             <% @users.each do |user| %>
-              <tr id="user-<%= user[:ubid]%>">
+              <tr id="user-<%= user.ubid %>">
                 <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6" scope="row">
-                  <%= user[:email] %>
+                  <%= user.email %>
```
enescakir, line 82: We typically use serializers as a layer between `views` and `routes` to manage the data we send to the web or API. Directly passing models or logic to the view can complicate data management. Although it might violated a few times, it's the current pattern in UI/route relationships. 

What's your opinion? Should we continue with this method or ditch it?

jeremyevans: We should definitely ditch the use of serializers for the web routes, where they are an unnecessary and IMO negative abstraction. I think their use in the web routes makes the code worse in all cases where they are used.

I would recommend only using serializers in api routes, and only if multiple routes use the same serialization, where the abstraction makes sense.  Otherwise, the serialization code can be inlined into the single route that uses it.  Further, I recommend that even in that case, the serializer code be moved to Clover helper methods.

However, I'm not advocating we make this change all at once.  We can certainly do it piecemeal.

jeremyevans:
This is freshly rebased on master, and ready for final review.

PR #2386, opened 2024-12-05
Location specific default subnets

furkansahin:
## Introduce location specific default subnets
This way, customers will simply use their default subnet in that region
unless they choose otherwise. This will simply make the experience
easier for the most basic customer. Nothing changes for the ones who
care about proper grouping of resources in different subnets.

## Github Runners force different subnet usage
This is important because Github Runners simply live in the same project
and they leave private subnet provisioning to the Vm::Nexus.assemble
function. To make sure every runner is provisioned in a different
subnet, we pass the foce_new_subnet flag.

This is not a problem for the other services because they create their
individual subnets and pass their id.

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -128,6 +128,12 @@ class Project < Sequel::Model
     end
   end
 
+  def default_private_subnet(location)
+    name = "default-#{LocationNameConverter.to_display_name(location)}"
+    ps = private_subnets_dataset.first(:location => location, Sequel[:private_subnet][:name] => name)
+    ps || Prog::Vnet::SubnetNexus.assemble(id, name: name, location: location).subject
+  end
+
   def self.feature_flag(*flags, into: self)
     flags.map(&:to_s).each do |flag|
       into.module_eval do
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -41,6 +41,12 @@ class Prog::Vm::GithubRunner < Prog::Base
       return picked_vm
     end
 
+    ps = Prog::Vnet::SubnetNexus.assemble(
+      Config.github_runner_service_project_id,
+      location: label_data["location"],
+      allow_only_ssh: true
+    ).subject
+
     vm_st = Prog::Vm::Nexus.assemble_with_sshable(
       "runneradmin",
       Config.github_runner_service_project_id,
@@ -52,7 +58,8 @@ class Prog::Vm::GithubRunner < Prog::Base
       enable_ip4: true,
       arch: label_data["arch"],
       allow_only_ssh: true,
-      swap_size_bytes: 4294963200 # ~4096MB, the same value with GitHub hosted runners
+      swap_size_bytes: 4294963200, # ~4096MB, the same value with GitHub hosted runners
+      private_subnet_id: ps.id
     )
 
     vm_st.subject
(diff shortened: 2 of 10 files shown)


model/project.rb
```diff
@@ -128,6 +128,10 @@ def validate
     end
   end
 
+  def default_private_subnet(location)
+    private_subnets.find { |ps| ps.location == location && ps.name == "default-#{LocationNameConverter.to_display_name(location)}" }
```
enescakir, line 132:
```suggestion
    private_subnets_dataset.where(location: location, name: "default-#{LocationNameConverter.to_display_name(location)}").first
```

prog/vm/github_runner.rb
```diff
@@ -52,7 +52,8 @@ def pick_vm
       enable_ip4: true,
       arch: label_data["arch"],
       allow_only_ssh: true,
-      swap_size_bytes: 4294963200 # ~4096MB, the same value with GitHub hosted runners
+      swap_size_bytes: 4294963200, # ~4096MB, the same value with GitHub hosted runners
+      force_new_subnet: true
```
enescakir, line 56: Instead of introducing a new `force_new_subnet` parameter, what do you think about just assemble a new PS here and pass `private_subnet_id`. I think the runners are the only use case.

furkansahin: Updated.

model/project.rb
```diff
@@ -128,6 +128,10 @@ def validate
     end
   end
 
+  def default_private_subnet(location)
+    private_subnets_dataset.where(location: location).where(Sequel[:private_subnet][:name] => "default-#{LocationNameConverter.to_display_name(location)}").first
+  end
```
enescakir, line 133: nit: since we expect `default_private_subnet` to return the default private subnet, we can also assemble of it to there. It allows to use the same name variable too.
```suggestion
  def default_private_subnet(location)
    name = "default-#{LocationNameConverter.to_display_name(location)}"
    ps = private_subnets_dataset.where(location: location).where(Sequel[:private_subnet][:name] => name).first
    ps || Prog::Vnet::SubnetNexus.assemble(project_id, name: name, location: location).subject
  end
```

prog/vm/nexus.rb
```diff
@@ -81,7 +81,8 @@ def self.assemble(public_key, project_id, name: nil, size: "standard-2",
           raise "Given subnet is not available in the given project" unless project.private_subnets.any? { |ps| ps.id == subnet.id }
           subnet
         else
-          subnet = Prog::Vnet::SubnetNexus.assemble(project_id, name: "#{name}-subnet", location: location, allow_only_ssh: allow_only_ssh).subject
+          subnet = project.default_private_subnet(location)
+          subnet ||= Prog::Vnet::SubnetNexus.assemble(project_id, name: "default-#{LocationNameConverter.to_display_name(location)}", location: location, allow_only_ssh: allow_only_ssh).subject
           PrivateSubnet[subnet.id]
```
enescakir, line 86: We already have `PrivateSubnet` model, we don't need to fetch it from DB again

enescakir (approved):

enescakir (approved):
LGTM 🚀

model/project.rb
```diff
@@ -128,6 +128,12 @@ def validate
     end
   end
 
+  def default_private_subnet(location)
+    name = "default-#{LocationNameConverter.to_display_name(location)}"
+    ps = private_subnets_dataset.where(location: location).where(Sequel[:private_subnet][:name] => name).first
```
jeremyevans, line 133:
```suggestion
    ps = private_subnets_dataset.first(location: location, Sequel[:private_subnet][:name] => name)
```

PR #2387, opened 2024-12-05
Log runner job conclusion

enescakir:
It will help us determine decreases in the success rate in real time using our monitoring tools.

I also added a data center to identify data center-specific issues.

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -36,10 +36,13 @@ class GithubRunner < Sequel::Model
   end
 
   def log_duration(message, duration)
-    values = {ubid: ubid, label: label, repository_name: repository_name, duration: duration}
+    values = {ubid: ubid, label: label, repository_name: repository_name, duration: duration, conclusion: workflow_job&.dig("conclusion")}
     if vm
       values.merge!(vm_ubid: vm.ubid, arch: vm.arch, cores: vm.cores)
-      values[:vm_host_ubid] = vm.vm_host.ubid if vm.vm_host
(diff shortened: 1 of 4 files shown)


furkansahin (approved):

model/github_runner.rb
```diff
@@ -36,10 +36,13 @@ def display_state
   end
 
   def log_duration(message, duration)
-    values = {ubid: ubid, label: label, repository_name: repository_name, duration: duration}
+    values = {ubid: ubid, label: label, repository_name: repository_name, duration: duration, conclusion: workflow_job&.dig("conclusion")}
```
velioglu, line 39: Do we need conclusion info each point we log duration? We might have additional log_conclusion function and call it from the billing record. Also, logging conclusion from log_duration function is a bit unintuitive.

enescakir: Actually, the duration is the execution time for `runner_completed` log. So it's not just conclusion log.

enescakir: `workflow_job&.dig("conclusion")` is `nil` if the job not completed yet

enescakir: I can add a condition check if `{ conclusion: nil }` feels wrong such as

```ruby
if (conclusion = workflow_job&.dig("conclusion"))
  values[:conclusion] = conclusion
end
```

enescakir: @velioglu Any comment?

velioglu (approved):

PR #2388, opened 2024-12-05
Remove hard dependency to MinIO connection in Postgres serializer

byucesoy:
We need to connect to MinIO cluster to calculate earliest restore time, however this is not always possible and can error out due to various reasons. In such cases, instead of returning 500, it is better to just hide restore UI. This commit does that by surrounding the code that tries to connect MinIO with a rescue block.

### DIFF
--- a/serializers/postgres.rb
+++ b/serializers/postgres.rb
@@ -27,7 +27,11 @@ class Serializers::Postgres < Serializers::Base
       )
 
       if pg.timeline && pg.representative_server&.primary?
-        base[:earliest_restore_time] = pg.timeline.earliest_restore_time&.utc&.iso8601
+        begin
+          base[:earliest_restore_time] = pg.timeline.earliest_restore_time&.utc&.iso8601
+        rescue => ex
+          Clog.emit("Failed to get earliest restore time") { Util.exception_to_hash(ex) }
+        end
         base[:latest_restore_time] = pg.timeline.latest_restore_time&.utc&.iso8601
       end
     end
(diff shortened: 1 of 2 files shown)


serializers/postgres.rb
```diff
@@ -27,7 +27,10 @@ def self.serialize_internal(pg, options = {})
       )
 
       if pg.timeline && pg.representative_server&.primary?
-        base[:earliest_restore_time] = pg.timeline.earliest_restore_time&.utc&.iso8601
+        begin
+          base[:earliest_restore_time] = pg.timeline.earliest_restore_time&.utc&.iso8601
+        rescue
```
enescakir, line 32: Should we log the exception to figure out the cause in the future?

enescakir (approved):

serializers/postgres.rb
byucesoy: makes sense. added logging.

serializers/postgres.rb
```diff
@@ -27,7 +27,11 @@ def self.serialize_internal(pg, options = {})
       )
 
       if pg.timeline && pg.representative_server&.primary?
-        base[:earliest_restore_time] = pg.timeline.earliest_restore_time&.utc&.iso8601
+        begin
+          base[:earliest_restore_time] = pg.timeline.earliest_restore_time&.utc&.iso8601
+        rescue => ex
```
bsatzger, line 32: we probably need a test to cover this.

byucesoy: Yes, initially it was not required since the rescue part is empty, but now we need a test. I added it now.

PR #2389, opened 2024-12-05
Openapi responses components

geemus:
Consolidate/DRY responses by using components.

This helps to make it more obvious where serializations are being used consistently (or not) and should make it much easier to keep things in sync as things continue to evolve and more details continue to be added (like examples).

I also think, even in cases where responses are not reused (like some lists) it helps communicate intent, keeps similar things together, and helps things be more future proof.

Overall I think this should be a nice quality-of-life improvement, without materially changing how things function. Happy to discuss further or tweak if needed.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -209,11 +209,7 @@ paths:
                 - cidr
       responses:
         '200':
-          description: Firewall rule created successfully
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/FirewallRule'
+          $ref: '#/components/responses/FirewallRule'
         default:
           $ref: '#/components/responses/Error'
       tags:
@@ -228,7 +224,7 @@ paths:
       summary: Delete a specific firewall rule
       responses:
         '204':
-          description: Firewall rule deleted successfully
+          $ref: '#/components/responses/Delete'
         default:
           $ref: '#/components/responses/Error'
       tags:
@@ -238,11 +234,7 @@ paths:
       summary: Get details of a firewall rule
       responses:
         '200':
-          description: Details of the firewall rule
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/FirewallRule'
+          $ref: '#/components/responses/FirewallRule'
         default:
           $ref: '#/components/responses/Error'
       tags:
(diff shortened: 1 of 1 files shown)


fdr (commented):
seems good, I'd request a couple of elaborations in the commit messages, though, I think the structural nature of openapi stuff is a bit special, so normal approaches have to be tweaked.

Probably the key thing about commit messages, is that one runs "git blame" that the message stand alone. Secondarily, when doing `git log --oneline`, the subject line should make as much sense as possible given the format.

Here's a proposal:

`componentize project responses` => `openapi: componentize project responses`


We don't normally do `subsystem: description`, but I think openapi can be considered special enough. Alternatively, "Make project responses components in openapi" 

I think the transform of making things components can stand alone without elaboration; it's just part of what someone needs to know about OpenAPI generally. Not much point repeating justification case by case, particularly if our justification is extremely orthodox, which I think it is, here.

Moving on,

The patch `postgres -> postgres database in schema` could use some description. In some places, it change capitalization, and others, changes operation ids.

` fix some 204s I missed in the first pass at shared responses ` should probably say something else. It could reference the "first pass" by commit, and what is it to "fix" (the entire subject line may change)

` make response component descriptions more consistent `

consistent with what, is kind of the question I have reading the patch.

Structurally, though, the small commits are very useful at not having to thread a needle while reading a bigger diff, so, thank you very much for that.

geemus:
I have become lax in my commit messages as I work on things that tend to squash and merge. I'll take another pass through and fix this up (and keep those things in mind moving forward) in the next few days.

geemus:
@fdr Ok. I added the prefixes, squished the consistency/missed commits into the earlier relevant commits, and added more description around the postgres database renaming. I think all the componentizing should be pretty self-explanatory, the rename seems more a matter of taste so I'm certainly willing to discuss if there is contention. It still feels better to me this way, but I it's not a deal breaker or anything.

fdr:
Interesting, did 204 fix drop off or is my cramped environment on my phone causing me to misread?

geemus:
@fdr the changes that were in the 204 fix should have been part of a couple of the other commits anyway. So I broke them out into individual commits and then squashed them into the right places. The 204 conversion to using the shared delete response should now be part of the componentization of the relevant resources.

fdr (approved):

PR #2390, opened 2024-12-05
Ensure request body is rewindable

jeremyevans:
Committee requires rewindable input.  Input is not guaranteed to be rewindable in Rack 3, and Rack::Lint will complain about if you call rewind on input.

To work around this, use Rack::RewindableInput in development mode, which is designed to handle rewinding.

To make sure Clover is compatible with the current Rack version in use, use Rack::Lint when running in test mode, unless SimpleCov is defined.  Since the default rake task and CI test both with and without coverage, this ensures tests will cover both the Rack::Lint and non-Rack::Lint cases.

This should fix api use when running puma in development mode

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -8,6 +8,30 @@ require "tilt"
 require "tilt/erubi"
 
 class Clover < Roda
+  # :nocov:
+  linting = Config.test? && !defined?(SimpleCov)
+  use Rack::Lint if linting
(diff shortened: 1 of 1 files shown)


geemus (approved):

fdr (approved):
Maybe committee can be enhanced so that it both avoids this and we only parse the json but once.

geemus:
Yeah, I think there are improvements that can be made on the committee side, and this solves the immediate issue.

PR #2392, opened 2024-12-06
Log the system's overall utilization every minute

enescakir:
- **Remove vm host based utilization logs**
  We added these logs to monitor our utilization with our monitoring tool,
  but since they are too granular, it’s hard to keep track of them.
  
  Since we don't use them, there's no need to log each VM host every 30
  seconds.
  

- **Log the system's overall utilization every minute**
  Previously, we printed similar logs for each VM host, but they were hard
  to digest.
  
  Aggregated utilization by location, architecture, and allocation state
  is more useful for us.
  
  For utilization grouped by architecture, I log the utilization of
  accepting VM hosts; we don't need to consider draining VM hosts.

```json
{
  "location_utilization": {
    "allocation_state": "accepting",
    "location": "hetzner-fsn1",
    "arch": "x64",
    "host_count": 1,
    "used_cores": 1,
    "total_cores": 32,
    "core_utilization": 3.13,
    "used_hugepages_1g": 2,
    "total_hugepages_1g": 249,
    "hugepage_utilization": 0.8
  },
  "message": "location utilization",
  "time": "2024-12-10 19:47:31 +0000",
  "thread": "st10g0vmh0st0vt111zat10nzz"
}

{
  "arch_utilization": {
    "host_count": 1,
    "used_cores": 1,
    "total_cores": 32,
    "used_hugepages_1g": 2,
    "total_hugepages_1g": 249,
    "arch": "x64",
    "core_utilization": 3.13,
    "hugepage_utilization": 0.8
  },
  "message": "arch utilization",
  "time": "2024-12-10 19:47:31 +0000",
  "thread": "st10g0vmh0st0vt111zat10nzz"
}
```

### DIFF
--- a/bin/respirate
+++ b/bin/respirate
@@ -25,6 +25,10 @@ Strand.dataset.insert_conflict.insert(id: "645cc9ff-7954-1f3a-fa82-ec6b3ffffff5"
 
 # We always insert this strand using the same UBID ("stexp1repr0ject1nv1tat10na")
 Strand.dataset.insert_conflict.insert(id: "776c1c3a-d804-9f3a-6683-5d874ad04155", prog: "ExpireProjectInvitations", label: "wait")
+
+# We always insert this strand using the same UBID ("st10g0vmh0st0vt111zat10nzz")
+Strand.dataset.insert_conflict.insert(id: "08200dd2-20ce-833a-de82-10fd5a082bff", prog: "LogVmHostUtilizations", label: "wait")
+
 clover_freeze
 
 loop do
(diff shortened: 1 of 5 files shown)


fdr (approved):
maybe an example of the emitted record (which can be kind of read out of the tests) in the commit message would be nice.

enescakir:
> maybe an example of the emitted record (which can be kind of read out of the tests) in the commit message would be nice.

Makes sense. This actually helped me discover a bug. When setting up the VM host, the total values are either nil or zero. If the controlplane has a single host, like in E2E environment or development, it raises a divided by zero exception. To fix this, I added a filter: `.where { (total_cores > 0) & (total_hugepages_1g > 0) }`."

PR #2395, opened 2024-12-09
Set longer deadlines for restore operations

byucesoy:
Normally we set 10 minute deadline for PostgreSQL databases to be ready, but in case of restore, it takes more time until the database is ready. This commit increases the deadline to 2 hours for restore operations, which would be enough for most cases.

### DIFF
--- a/prog/postgres/postgres_resource_nexus.rb
+++ b/prog/postgres/postgres_resource_nexus.rb
@@ -90,8 +90,12 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
     nap 5 unless representative_server.vm.strand.label == "wait"
 
     postgres_resource.incr_initial_provisioning
-    register_deadline(:wait, 10 * 60)
-    bud self.class, frame, :trigger_pg_current_xact_id_on_parent if postgres_resource.parent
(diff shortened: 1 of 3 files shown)


prog/postgres/postgres_resource_nexus.rb
```diff
@@ -90,8 +90,12 @@ def before_run
     nap 5 unless representative_server.vm.strand.label == "wait"
 
     postgres_resource.incr_initial_provisioning
-    register_deadline(:wait, 10 * 60)
-    bud self.class, frame, :trigger_pg_current_xact_id_on_parent if postgres_resource.parent
+    if postgres_resource.parent
+      bud self.class, frame, :trigger_pg_current_xact_id_on_parent
+      register_deadline(:wait, 120 * 60)
```
bsatzger, line 95: At some point in the future, we could maybe set this adaptively, based on the storage size of the parent?

bsatzger (approved):

prog/postgres/postgres_resource_nexus.rb
byucesoy: Yep, makes sense.

PR #2403, opened 2024-12-11
Selected Github Runners switch to Cloudflare and Google DNS and add all-servers to Dnsmasq

furkansahin:
## Add feature flag to use all-server param for dnsmasq
We are enabling this in a controlled manner because all-servers
parameter will essentially amplify the number of dns queries we make.
That should not be a big problem since only the non-cached dns queries
will be sent out but still it is better to roll-out in a controlled
manner.

## Set all-servers and use Cloudflare and Google DNS for selected runners

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -65,6 +65,17 @@ class GithubRunner < Sequel::Model
     aggregate_readings(previous_pulse: previous_pulse, reading: reading, data: {available_memory: available_memory})
   end
 
+  def generate_cf_google_dns_dnsmasq_config
+    <<~COMMAND
+      sudo sed -i 's/^server=9.9.9.9@.*/server=2606:4700:4700::1111/' /vm/#{vm.inhost_name}/dnsmasq.conf
+      sudo sed -i 's/^server=149.112.112.112@.*//' /vm/#{vm.inhost_name}/dnsmasq.conf
+      sudo sed -i 's/^server=2620:fe::fe//' /vm/#{vm.inhost_name}/dnsmasq.conf
+      sudo sed -i 's/^server=2620:fe::9/server=2001:4860:4860::8888/' /vm/#{vm.inhost_name}/dnsmasq.conf
+      echo "all-servers" | sudo tee -a /vm/#{vm.inhost_name}/dnsmasq.conf
+      sudo systemctl restart #{vm.inhost_name}-dnsmasq
+    COMMAND
+  end
+
   def self.redacted_columns
     super + [:workflow_job]
   end
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -247,6 +247,8 @@ class Prog::Vm::GithubRunner < Prog::Base
     # Remove comments and empty lines before sending them to the machine
     vm.sshable.cmd(command.gsub(/^(\s*# .*)?\n/, ""))
 
+    vm.vm_host.sshable.cmd(github_runner.generate_cf_google_dns_dnsmasq_config) if github_runner.installation.project.get_ff_all_servers_dnsmasq
+
     hop_register_runner
   end
 
--- a/spec/prog/vm/github_runner_spec.rb
+++ b/spec/prog/vm/github_runner_spec.rb
@@ -386,7 +386,7 @@ RSpec.describe Prog::Vm::GithubRunner do
     it "hops to register_runner" do
       expect(vm).to receive(:vm_host).and_return(instance_double(VmHost, ubid: "vhfdmbbtdz3j3h8hccf8s9wz94", location: "hetzner-fsn1", data_center: "FSN1-DC8")).at_least(:once)
       expect(vm).to receive(:runtime_token).and_return("my_token")
-      expect(github_runner.installation).to receive(:project).and_return(instance_double(Project, ubid: "pjwnadpt27b21p81d7334f11rx", path: "/project/pjwnadpt27b21p81d7334f11rx")).at_least(:once)
+      expect(github_runner.installation).to receive(:project).and_return(instance_double(Project, ubid: "pjwnadpt27b21p81d7334f11rx", path: "/project/pjwnadpt27b21p81d7334f11rx", get_ff_all_servers_dnsmasq: false)).at_least(:once)
       expect(github_runner.installation).to receive(:cache_enabled).and_return(false)
       expect(sshable).to receive(:cmd).with(<<~COMMAND)
         set -ueo pipefail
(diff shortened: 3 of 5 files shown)


prog/vm/github_runner.rb
```diff
@@ -247,6 +247,18 @@ def setup_info
     # Remove comments and empty lines before sending them to the machine
     vm.sshable.cmd(command.gsub(/^(\s*# .*)?\n/, ""))
 
+    if github_runner.installation.project.get_ff_all_servers_dnsmasq
```
bsatzger, line 250: With this change we send every DNS query concurrently to 4 servers. Two might be already enough. 
```
Dec 11 10:32:59 vhxnr40hj3dp5kq5s49t28xth2 dnsmasq[1877858]: query[A] ab32a8b09b10c1392caed903.com from 10.193.172.228
Dec 11 10:32:59 vhxnr40hj3dp5kq5s49t28xth2 dnsmasq[1877858]: forwarded ab32a8b09b10c1392caed903.com to 2606:4700:4700::1111
Dec 11 10:32:59 vhxnr40hj3dp5kq5s49t28xth2 dnsmasq[1877858]: forwarded ab32a8b09b10c1392caed903.com to 2606:4700:4700::1001
Dec 11 10:32:59 vhxnr40hj3dp5kq5s49t28xth2 dnsmasq[1877858]: forwarded ab32a8b09b10c1392caed903.com to 2620:fe::fe
Dec 11 10:32:59 vhxnr40hj3dp5kq5s49t28xth2 dnsmasq[1877858]: forwarded ab32a8b09b10c1392caed903.com to 2620:fe::9
```

bsatzger: instead of rewriting the config it would be cleaner to write two different kinds of configs in vm_setup from the get go. If this is temporary maybe it's ok.

bsatzger: The good thing is that with this dnsmasq config change DNS resolution become much more stable and latency more predictable in a simple experiment I did.

prog/vm/github_runner.rb
```diff
@@ -247,6 +247,18 @@ def setup_info
...
+      # replace server=9.9.9.9@IP_ADDRESS with server=2606:4700:4700::1111
+      # and server=149.112.112.112@IP_ADDRESS with server=2606:4700:4700::1001
+      command = <<~COMMAND
+        sudo sed -i 's/^server=9.9.9.9@.*/server=2606:4700:4700::1111/' /vm/#{vm.inhost_name}/dnsmasq.conf
+        sudo sed -i 's/^server=149.112.112.112@.*/server=2606:4700:4700::1001/' /vm/#{vm.inhost_name}/dnsmasq.conf
+        echo "all-servers" | sudo tee -a /vm/#{vm.inhost_name}/dnsmasq.conf
+        sudo systemctl restart #{vm.inhost_name}-dnsmasq
+      COMMAND
+      vm.vm_host.sshable.cmd(command)
```
bsatzger, line 259: I'd try to avoid this additional `sshable.cmd(command)`. Similar to `github_runner.installation.cache_enabled`, why not just add these commands to `command` and then execute it all here in one go https://github.com/ubicloud/ubicloud/blob/f56ed7c3787d1e76c00c459a5c0c17a219c984bb/prog/vm/github_runner.rb#L248

furkansahin: but this one is run against vm_host

bsatzger: ah, I see, of course. then we can't combine the ssh calls.

prog/vm/github_runner.rb
furkansahin: You're right but I made a decision to completely remove quad9 and use CF and Google if we're gonna use only 2 endpoints. At this point, I don't trust them much.

enescakir:
How many days do you think we'll keep it like this behind a feature flag? It will increase queue times for these projects.

furkansahin:
> How many days do you think we'll keep it like this behind a feature flag? It will increase queue times for these projects.


I expect it to be around 2-3 days;

spec/prog/vm/github_runner_spec.rb
```diff
@@ -422,6 +422,36 @@
...
+        set -ueo pipefail
+        echo "image version: $ImageVersion"
+        sudo usermod -a -G sudo,adm runneradmin
+        jq '. += [{"group":"Ubicloud Managed Runner","detail":"Name: #{github_runner.ubid}\\nLabel: ubicloud-standard-4\\nArch: \\nImage: \\nVM Host: vhfdmbbtdz3j3h8hccf8s9wz94\\nVM Pool: \\nLocation: hetzner-fsn1\\nDatacenter: FSN1-DC8\\nProject: pjwnadpt27b21p81d7334f11rx\\nConsole URL: http://localhost:9292/project/pjwnadpt27b21p81d7334f11rx/github"}]' /imagegeneration/imagedata.json | sudo -u runner tee /home/runner/actions-runner/.setup_info
+        echo "UBICLOUD_RUNTIME_TOKEN=my_token
+        UBICLOUD_CACHE_URL=http://localhost:9292/runtime/github/" | sudo tee -a /etc/environment
+        ([ -f "/etc/docker/daemon.json" ] && sudo cat /etc/docker/daemon.json || echo '{}') | jq '. += #{JSON.generate({"experimental" => false, "dns-opts" => ["timeout:2", "attempts:3"]})}' | sudo tee /etc/docker/daemon.json
+        sudo systemctl restart docker
+        echo "CUSTOM_ACTIONS_CACHE_URL=http://10.0.0.1:51123/random_token/" | sudo tee -a /etc/environment
+      COMMAND
```
fdr, line 442: I might make this string generation a method and test that separately, then you don't need such a complex fixture/mocks, rather than inlining it.

So you might end up with a (very fast, in-memory) test that's like:

```ruby
github_runner.inhost_name = 'foo' # or an expect, or whatever
expect(obj.generate_cloudflare_sed).to eq(<<str)
the string
str
```

Then your test in integration can piggy-back on a different test that does all the fixture stuff, e.g.:

```ruby
expect(sshable).to receive(cmd).with(obj.generate_cloudflare_sed)
```

The net result is intended to be smaller and less intensive to read. As-is, there's a very long string that I take it you need because there are multiple calls to `cmd` and you can't reduce so easily to `receive(:cmd)` without `with`.

An alternative version you may prefer is put even the `cmd` into the helper method. With that approach, you can use a expect/with construction to make sure the string is what you expect, like so:

```ruby
github_runner.inhost_name = 'foo' # or an expect, or whatever
expect(sshable).to receive(:cmd).with('thestring')
obj.run_cloudflare_sed
```

This uglifies relative to a simple `eq`, but the payoff is in the test that integrates `run_cloudflare_sed`:

```ruby
expect(obj).to receive(:run_cloudflare_sed)
```

is all you need.

I realize that was a bit hand-wavey, but does it make sense? the general idea is to eliminate a lot of new-new fixture code and make the tests denser. Currently the ratio of interesting-to-not-very-interesting expressions in this test is rather low.

furkansahin: Thank you! I refactored the part that I added but the above CMD is unrelated from this PR. We should fix it in a different patch.

model/github_runner.rb
```diff
@@ -65,6 +65,17 @@ def check_pulse(session:, previous_pulse:)
     aggregate_readings(previous_pulse: previous_pulse, reading: reading, data: {available_memory: available_memory})
   end
 
+  def generate_cf_google_dns_dnsmasq_config
```
enescakir, line 68: Are these commands idempotent? If we run 'setup_environment' multiple times for reasons like deployment, will it work the second time?

enescakir (approved):

model/github_runner.rb
furkansahin: yes

PR #2404, opened 2024-12-11
Switch everyone to CF + Google DNS

furkansahin:
## Revert "NAT Dnsmasq IPv4 address properly"
This reverts commit 370eda80b6f49c5ba0c11781d7294756bd219af9.

## Revert "Add feature flag to use all-server param for dnsmasq"
This reverts commit 29417db8061e0969183354391fb36b6f1473c3a9.

## Revert "Set all-servers and use Cloudflare and Google DNS for selected runners"
This reverts commit b1465bf63075d2d7e7c9ca04a3ffcab6ee033e25.

## Switch to using Cloudflare and Google DNS with all-servers

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -68,17 +68,6 @@ class GithubRunner < Sequel::Model
     aggregate_readings(previous_pulse: previous_pulse, reading: reading, data: {available_memory: available_memory})
   end
 
-  def generate_cf_google_dns_dnsmasq_config
-    <<~COMMAND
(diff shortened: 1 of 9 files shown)


fdr:
I reviewed it anyway, but kinda see why you told everyone not to :) [not so much that the cumulative change as big as you are adjusting things back and forth]

fdr:
also you put in my name as a reviewer, which is why I did that

furkansahin:
@fdr now it's reviewable :)

PR #2406, opened 2024-12-11
openapi: add ubid pattern and example

geemus:
This adds the appropriate patterns to match ubid values across different resource types and either adds examples or corrects them to valid values. I also updated tests to use valid ubid values for not-found cases (as they were failing as invalid due to committee), for these I used the appropriate prefix plus all zeroes to hopefully still show at a glance that it isn't intended to be real (though technically it could be). I don't know off hand if this could create test flake, given the possibility that this ubid could really be generated, so others may want to chime in there.

This also highlighted that firewall rules and postgres firewall rules were reusing the same parameter definitions, and really shouldn't be. So I broke these out as well.

The documentation site also has some incorrect ubids in the overview, I'll try to fix that in an upcoming work session if someone doesn't beat me to it.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -934,7 +934,7 @@ paths:
       - $ref: '#/components/parameters/location'
       - $ref: '#/components/parameters/postgres_database_id'
       - $ref: '#/components/parameters/project_id'
-      - $ref: '#/components/parameters/firewall_rule_id'
+      - $ref: '#/components/parameters/postgres_firewall_rule_id'
     delete:
       operationId: deleteLocationPostgresFirewallRuleWithId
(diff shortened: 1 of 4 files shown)


openapi/openapi.yml
```diff
@@ -934,7 +934,7 @@ paths:
       - $ref: '#/components/parameters/location'
       - $ref: '#/components/parameters/postgres_database_id'
       - $ref: '#/components/parameters/project_id'
-      - $ref: '#/components/parameters/firewall_rule_id'
+      - $ref: '#/components/parameters/postgres_firewall_rule_id'
```
fdr, line 937: hah woops

fdr (approved):

PR #2408, opened 2024-12-11
Add vm restart to the UI

enescakir:
- **Move reset PG superuser password to danger zone**
  I moved dangerous actions to one place and added a title
  
CleanShot 2024-12-11 at 23 42 12@2x

- **Add vm restart semaphore**
  Several customers have already requested a method to force restart their
  VMs. Previously, we handled this through support tickets. However, we
  can easily integrate this functionality into the UI.
  

- **Add vm restart button to the UI**
  I did it very similar to the PostgreSQL restart functionality.
  
CleanShot 2024-12-11 at 23 42 20@2x

- **Change vm display state while restarting it**

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -9,6 +9,7 @@ require "base64"
 class Prog::Vm::Nexus < Prog::Base
   subject_is :vm
   semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup, :update_spdk_dependency, :waiting_for_capacity, :lb_expiry_started
+  semaphore :restart
 
   def self.assemble(public_key, project_id, name: nil, size: "standard-2",
     unix_user: "ubi", location: "hetzner-fsn1", boot_image: Config.default_boot_image_name,
@@ -323,6 +324,11 @@ class Prog::Vm::Nexus < Prog::Base
       hop_update_spdk_dependency
     end
 
+    when_restart_set? do
+      register_deadline(:wait, 5 * 60)
+      hop_restart
+    end
+
     when_checkup_set? do
       hop_unavailable if !available?
       decr_checkup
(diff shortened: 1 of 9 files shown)


views/vm/show.erb
```diff
@@ -89,23 +89,55 @@
...
+      <!-- Restart Card -->
+      <% if has_permission?("Vm:edit", @vm[:id]) %>
+        <div class="px-4 py-5 sm:p-6">
+          <form action="<%= "#{@project_data[:path]}#{@vm[:path]}/restart" %>" role="form" method="POST">
+            <%== csrf_tag("#{@project_data[:path]}#{@vm[:path]}/restart") %>
+            <div class="sm:flex sm:items-center sm:justify-between">
+              <div>
+                <h3 class="text-base font-semibold leading-6 text-gray-900">Restart virtual machine</h3>
+                <div class="mt-2 text-sm text-gray-500">
+                  <p>This action will restart the virtual machine. It will be offline momentarily, and all connections
```
bsatzger, line 111: Something like this might work better for vms.
```
This action will restart the virtual machine, causing it to be temporarily offline.
```

prog/vm/nexus.rb
```diff
@@ -9,6 +9,7 @@
 class Prog::Vm::Nexus < Prog::Base
   subject_is :vm
   semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup, :update_spdk_dependency, :waiting_for_capacity, :lb_expiry_started
+  semaphore :restart
```
bsatzger, line 12: why not just `, :restart`?

bsatzger: getting too long?

prog/vm/nexus.rb
```diff
@@ -329,6 +330,11 @@ def write_params_json
       hop_update_spdk_dependency
     end
 
+    when_restart_set? do
```
bsatzger, line 333: We can only restart if the vm is in wait state. I think it's fine for vms. if we had more sophisticated unavailability checks it might be desirable to still be able to restart if it's in unavailable state.

bsatzger (approved):

views/vm/show.erb
enescakir: Thanks

prog/vm/nexus.rb
enescakir: Yes, when we have more confidence, we can do it. When we do it, we can push to `restart` label similar to the PG restart. So when it exited from the restart, it will return the state that restart initiated.

byucesoy (approved):

PR #2411, opened 2024-12-12
Add io_limits to VmSize.

pykello:
We will put limits on IO rate burstable or shared instances can have. This PR prepares for that by adding `io_limits` to VmSize. Currently, IO performance of standard VMs are not limited. We might revisit this in future.

### DIFF
--- a/lib/option.rb
+++ b/lib/option.rb
@@ -48,17 +48,20 @@ module Option
     ["almalinux-9", "AlmaLinux 9"]
   ].map { |args| BootImage.new(*args) }.freeze
 
-  VmSize = Struct.new(:name, :family, :cores, :vcpus, :memory_gib, :storage_size_options, :visible, :gpu, :arch) do
+  IoLimits = Struct.new(:max_ios_per_sec, :max_read_mbytes_per_sec, :max_write_mbytes_per_sec)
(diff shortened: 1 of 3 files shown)


lib/option.rb
```diff
@@ -48,17 +48,20 @@ def self.postgres_locations
     ["almalinux-9", "AlmaLinux 9"]
   ].map { |args| BootImage.new(*args) }.freeze
 
-  VmSize = Struct.new(:name, :family, :cores, :vcpus, :memory_gib, :storage_size_options, :visible, :gpu, :arch) do
+  IoLimits = Struct.new(:max_ios_per_sec, :max_read_mbytes_per_sec, :max_write_mbytes_per_sec)
+  NO_IO_LIMITS = IoLimits.new(nil, nil, nil)
```
jeremyevans, line 52:
```suggestion
  NO_IO_LIMITS = IoLimits.new(nil, nil, nil).freeze
```

You generally want constant values to be frozen.  In this case, it is important, as the object is shared by multiple (all current?) VmSizes?

jeremyevans (approved):
This looks fine to me.  I have one suggestion, which I think should be merged first.

PR #2416, opened 2024-12-12
Create partitioned table to replace deleted_record

bsatzger:
**Migration: Add partitioned table archived_record**
This commit introduces the archived_record table, similar in schema to deleted_record but without an id column or a uniqueness constraint. The table is range-partitioned by the deleted_at column and includes an index on (model_name, deleted_at). This allows for efficient execution of queries that filter on both model_name and deleted_at.

Monthly partitions are pre-created until Jan 2026. A follow-up commit adds a test that starts failing when it's getting time to create new partitions.

**Create ArchivedRecord model**
ArchivedRecord is created with a structure similar to DeletedRecord but based on a partitioned table. A test is included that triggers a failure 45 days before the table's partitions are exhausted.

**Replace DeletedRecord with ArchivedRecord**
The DeletedRecord model is retired and replaced by ArchivedRecord, which is based on a partitioned table. This change improves the analysis of archived records and simplifies the process of deleting old data.


migrate/20241213_archived_record.rb
```diff
@@ -0,0 +1,56 @@
...
+
+Sequel.migration do
+  change do
+    create_table(:archived_record, partition_by: :deleted_at, partition_type: :range) do
+      column :deleted_at, :timestamptz, null: false, default: Sequel::CURRENT_TIMESTAMP
+      column :model_name, :text, null: false
+      column :model_values, :jsonb, null: false, default: "{}"
+    end
+
+    create_table(:archived_record_2024_10, partition_of: :archived_record) do
```
jeremyevans, line 11: Migrations are normal ruby code, I recommend using a loop here instead of manually unrolling it.

model.rb
```diff
@@ -97,7 +97,7 @@ def before_destroy
         end
       end
 
-      DeletedRecord.create(deleted_at: Time.now, model_name: model_name, model_values: model_values)
+      ArchivedRecord.create(deleted_at: Time.now, model_name: model_name, model_values: model_values)
```
jeremyevans, line 100:
```suggestion
      ArchivedRecord.create(model_name: model_name, model_values: model_values)
```

I realize `deleted_at` was set before, but it's better to have the database set the default timestamp value.

migrate/20241213_archived_record.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:archived_record, partition_by: :deleted_at, partition_type: :range) do
```
jeremyevans, line 5: This may be a dumb question, as I haven't used deleted_records myself, but isn't the biggest issue with deleted_records that you have to do sequential scans?  What are your thoughts on creating an index, maybe on deleted_at or model_name (depending on which is more valuable)?  I realize you give up cheaper inserts, but it should make querying the table faster.

jeremyevans (commented):
Using a partitioned table makes sense to me.

model.rb
bsatzger: Agree. Applied your suggested change.

migrate/20241213_archived_record.rb
bsatzger: Switched to creating partitions in a loop.

migrate/20241213_archived_record.rb
bsatzger: I believe the vast majority of queries would look at a single type of model in a given time interval. I added an index on (model_name, deleted_at), which should work well if there's a filter on both model_name and deleted_at or just model_name. If theres a filter on deleted_at only, which I believe would be rare, we'd still do a sequential scan, but at least only on relevant partitions. I'm fine with changing the index if I'm misjudging query access patterns (if we not almost always filter by model_name).

enescakir (approved):

migrate/20241213_archived_record.rb
```diff
@@ -0,0 +1,20 @@
...
```
byucesoy, line 5: Maybe use `:archived_at` as column name to match with the table name?

jeremyevans (approved):

migrate/20241213_archived_record.rb
bsatzger: I don't have a strong opinion on this. I felt that `deleted_at` still works, like soft deletion of a record puts it into the archive with a deleted_at timestamp on it. But `archived_at` might be more consistent. I can change the name.

PR #2417, opened 2024-12-12
Get semaphores from the subject of prog

enescakir:
Right now, we define semaphores in both the model and prog, but they are generally the same. We can retrieve the list of semaphores from the subject in prog.

We can still define semaphores in prog if it doesn't have a subject, like in Prog::Test::HetznerServer.

### DIFF
--- a/model.rb
+++ b/model.rb
@@ -28,7 +28,12 @@ module SemaphoreMethods
   end
 
   module ClassMethods
+    def semaphore_names
+      @semaphore_names || []
+    end
+
     def semaphore(*names)
+      (@semaphore_names ||= []).concat(names)
       names.each do |sym|
         name = sym.name
         define_method :"incr_#{name}" do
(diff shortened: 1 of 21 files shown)


model.rb
```diff
@@ -27,7 +27,12 @@ def self.included(base)
   end
 
   module ClassMethods
+    def semaphore_names
+      @semaphore_names || []
+    end
+
     def semaphore(*names)
+      (@semaphore_names ||= []).push(*names)
```
jeremyevans, line 35:
```suggestion
      (@semaphore_names ||= []).concat(names)
```

jeremyevans (approved):
Looks great!  I only have one suggestion, which is up to you.

fdr (approved):
I think this is an improvement, and low risk, most of the risk in manipulating semaphores would be in the snapshot code, since it can affect concurrency control (which this doesn't touch).

I mean, -24, who can turn that down?

PR #2420, opened 2024-12-13
Add rhizome side of SliceSetup.

pykello:
### Expand `r` in `util.rb` to get a list of expected exit statuses
Expand `r` in `util.rb` to get a list of expected exit statuses
    
By default, `r` threw exceptions whenever the exit status was not 0. However, in some scenarios the caller is ok with some exit statuses. For example, when the caller wants to stop a systemd unit if it exists. In which case, it is ok if the exit status is 5.
    
This change adds the `expect:` parameter to `r` to allow for that:
    
```
    def r(commandline, stdin: "", expect: [0])
```

### Add rhizome side of SliceSetup. 

This change introduces the rhizome library and executable to manage the creation and deletion of slices. Slices will form the foundation for controlling VM resource usage in future updates. This functionality will enable support for burstable VM instances.

### Add rspec tests for SliceSetup.
This change make sure we have code coverage SliceSetup.

### Add E2E rspec to actually try SliceSetup.

In this change we add an rspec test to actually try the SliceSetup library & verify that it works. This will be run as part of E2E tests.

### DIFF
--- a/rhizome/host/bin/setup-slice
+++ b/rhizome/host/bin/setup-slice
@@ -1,0 +1,42 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../lib/slice_setup"
+
+unless (action = ARGV.shift)
+  puts "expected action as argument"
+  exit 1
+end
+
+unless (slice_name = ARGV.shift)
+  puts "expected slice name as argument"
+  exit 1
+end
+
+slice_setup = SliceSetup.new(slice_name)
+
+case action
+when "delete"
+  slice_setup.purge
+
+when "prep"
+  unless (allowed_cpus = ARGV.shift)
+    puts "expected list of allowed cpus as argument"
+    exit 1
+  end
+  slice_setup.prep(allowed_cpus)
+
+when "recreate-unpersisted"
+  slice_setup.start_systemd_unit
+
+when "reinstall-systemd-units"
+  unless (allowed_cpus = ARGV.shift)
+    puts "expected list of allowed cpus as argument"
+    exit 1
+  end
+  slice_setup.install_systemd_unit(allowed_cpus)
+
+else
+  puts "Invalid action #{action}"
+  exit 1
+end
--- a/rhizome/host/lib/slice_setup.rb
+++ b/rhizome/host/lib/slice_setup.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require "fileutils"
+
+class SliceSetup
+  def initialize(slice_name)
+    @slice_name = slice_name
+  end
+
+  def systemd_service
+    @systemd_service ||= File.join("/etc/systemd/system", @slice_name)
+  end
+
+  def prep(allowed_cpus)
+    fail "BUG: invalid cpuset" unless valid_cpuset?(allowed_cpus)
+    install_systemd_unit(allowed_cpus)
+    start_systemd_unit
+  end
+
+  def purge
+    r("systemctl stop #{@slice_name.shellescape}", expect: [0, 5])
+    rm_if_exists systemd_service
+    r "systemctl daemon-reload"
+  end
+
+  def install_systemd_unit(allowed_cpus)
+    fail "BUG: unit name must not be empty" if @slice_name.empty?
+    fail "BUG: we cannot create system units" if @slice_name == "system.slice" || @slice_name == "user.slice"
+    fail "BUG: unit name cannot contain a dash" if @slice_name.include?("-")
+    fail "BUG: invalid allowed_cpus" if !valid_cpuset?(allowed_cpus)
+
+    # Only proceed if the slice has not yet been setup
+    unless File.exist? systemd_service
+      safe_write_to_file(systemd_service, <<SLICE_CONFIG)
+[Unit]
+Description=Restricting resouces for virtual machines
+Before=slices.target
+[Slice]
+AllowedCPUs=#{allowed_cpus}
+SLICE_CONFIG
+
+      r "systemctl daemon-reload"
+    end
+  end
+
+  def start_systemd_unit
+    r "systemctl start #{@slice_name}"
+    cpuset_path = File.join("/sys/fs/cgroup", @slice_name, "cpuset.cpus.partition")
+    File.write(cpuset_path, "root")
+  end
+
+  def valid_cpuset?(str)
+    return false if str.nil? || str.empty?
+    str.split(",").all? do |part|
+      if part.include?("-")
+        r = part.split("-")
+        r.size == 2 && r.all? { _1.to_i.to_s == _1 } && r[0].to_i <= r[1].to_i
+      else
+        part.to_i.to_s == part
+      end
+    end
+  end
+end
--- a/rhizome/common/lib/util.rb
+++ b/rhizome/common/lib/util.rb
@@ -24,9 +24,9 @@ class FsyncFail < Exception
 end
 # rubocop:enable Lint/InheritException
 
-def r(commandline, stdin: "")
+def r(commandline, stdin: "", expect: [0])
   stdout, stderr, status = Open3.capture3(commandline, stdin_data: stdin)
-  fail CommandFail.new("command failed: " + commandline, stdout, stderr) unless status.success?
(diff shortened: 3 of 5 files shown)


jeremyevans (commented):
I don't have the experience to review this from a Linux/control plane perspective, but from a Ruby perspective, it seems fine.  I added some small comments.

rhizome/host/bin/setup-slice
```diff
@@ -0,0 +1,43 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../lib/slice_setup"
+
+unless (action = ARGV.shift)
+  puts "expected action as argument"
```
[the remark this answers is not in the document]
macieksarnowicz: This was following what's in setup-vm. If there is a better way, I am happy to update both places.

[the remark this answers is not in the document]
jeremyevans: Considering this accepts a different number of arguments for different subcommands, a more correct approach in terms of arguments would add significant complexity, so I don't think it's worth it.

As for the stdout vs stderr difference, it would definitely be important if this was a command a user would run, but considering this is for Rhizome, it's fine.

In short, I don't think you need to make changes for this.

macieksarnowicz:
@pykello - in the commit description, maybe skip the part about 'shared' VMs, we are not going to have anything named like that. We can just say 'burstables' for now. Second, how about the changes in vm_setup.rb and setup-vm scripts. I think those could all go together.

pykello:
> @pykello - in the commit description, maybe skip the part about 'shared' VMs, we are not going to have anything named like that. We can just say 'burstables' for now.

done.

> Second, how about the changes in vm_setup.rb and setup-vm scripts. I think those could all go together.

They need a separate review & set of sets, so I'll put them in separate PR.

rhizome/host/lib/slice_setup.rb
```diff
@@ -0,0 +1,52 @@
...
+  def prep(allowed_cpus)
+    install_systemd_unit(allowed_cpus)
+    start_systemd_unit
+  end
+
+  def purge
+    if File.exist? systemd_service
+      r "systemctl stop #{@slice_name.shellescape}"
+      FileUtils.rm_f(systemd_service)
+
```
fdr, line 24: Perhaps this conditional can be made more robust. It tries to get something like `rm -f` semantics is how I'm reading it. It's the same as the anti-pattern of "File.exist?; File.rm", where it's almost always better to handle the rm exception.

In this case, I would try to stop the slice and absorb exit code 5, the "not found" exit status. Perhaps `r` can be expanded to support acceptable exit codes to avoid exception rigmarole, e.g. `expect: [0, 5]` as a keyword argument. The default would be `expect: 0` . Other nomenclature welcome, because `expect` looks a bit too much like rspec for comfort.

Another thing that has borrowed me is that `FileUtils.rm_f` (and indeed `rm -f`) is too relaxed about things like EPERM, which would be a reason to crash. We really only want to absorb ENOENT. Perhaps that can be a `common/lib/util` routine.

rhizome/host/lib/slice_setup.rb
```diff
@@ -0,0 +1,52 @@
...
+      r "systemctl daemon-reload"
+    end
+  end
+
+  def install_systemd_unit(allowed_cpus)
+    fail "BUG: unit name must not be empty" if @slice_name.empty?
+    fail "BUG: we cannot create system units" if @slice_name == "system.slice" || @slice_name == "user.slice"
+    fail "BUG: unit name cannot contain a dash" if @slice_name.include?("-")
```
fdr, line 32: as long as you are checking inputs, check allowed_cpus for nil in particular, since that will expand into a blank in the unit file and IIRC in the unit file such a blank expansion means "clear that setting" (i.e. it could be relatively silent kind of "oops")

rhizome/host/lib/slice_setup.rb
pykello: done

rhizome/host/lib/slice_setup.rb
pykello: done

rhizome/host/bin/setup-slice
```diff
@@ -0,0 +1,42 @@
...
+end
+
+slice_setup = SliceSetup.new(slice_name)
+
+case action
+when "delete"
+  slice_setup.purge
+
+when "prep"
+  unless (allowed_cpus = ARGV.shift)
```
furkansahin, line 23: maybe you should check the format of these parameters. I can see that you pass an `int-int` pair in the tests. Just in case, in future if the order of parameters are messed up, we can catch it here?

pykello: I added a check in `SliceSetup.prep` to fail if format is not correct.

rhizome/host/lib/slice_setup.rb
```diff
@@ -47,4 +48,15 @@ def start_systemd_unit
     r "systemctl start #{@slice_name}"
     r "echo \"root\" > /sys/fs/cgroup/#{@slice_name}/cpuset.cpus.partition"
   end
+
+  def valid_cpuset?(str)
+    str.split(",").all? do |part|
+      if part.include?("-")
+        r = part.split("-")
+        r.size == 2 && r.all? { _1.to_i.to_s == _1 } && r[0].to_i <= r[1].to_i
```
fdr, line 56: You sure about .to_i? If it gets any weird input, it'll give you zero, rather than raising. `Integer(x, 10)` is the most restrictive.

fdr (approved):
I know I had to read "Moby Dick" first, perhaps three times, but here's the approval

rhizome/host/lib/slice_setup.rb
fdr: Ah okay, this is among the shortest ways to test for reflexivity of the string representation plus inequality. Very slick, @jeremyevans you got something that's clearer (and maybe avoids some extra to_i conversion) but not longer? If not, I think its fine to keep.

jeremyevans: 3 characters shorter, and avoids duplicate `to_i` calls:

```ruby
r.size == 2 && r == (s = r.map(&:to_i)).map(&:to_s) && s[0] <= s[1]
```

Not sure that is clearer.  Maybe if you move the `s` assignment to the line above, but that's technically longer if you count the whitespace (5 characters shorter without the whitespace, since your can avoid the parentheses in that case).

I'm not requesting the code be changed, what is there seems fine.

furkansahin (approved):

rhizome/host/lib/slice_setup.rb
```diff
@@ -0,0 +1,62 @@
...
+    r("systemctl stop #{@slice_name.shellescape}", expect: [0, 5])
+    rm_if_exists systemd_service
+    r "systemctl daemon-reload"
+  end
+
+  def install_systemd_unit(allowed_cpus)
+    fail "BUG: unit name must not be empty" if @slice_name.empty?
+    fail "BUG: we cannot create system units" if @slice_name == "system.slice" || @slice_name == "user.slice"
+    fail "BUG: unit name cannot contain a dash" if @slice_name.include?("-")
+    fail "BUG: allowed_cpus cannot be nil or empty" if allowed_cpus.nil? || allowed_cpus.empty?
```
macieksarnowicz, line 31: Should this call valid_cpuset? as well?

rhizome/host/lib/slice_setup.rb
```diff
@@ -0,0 +1,62 @@
...
+AllowedCPUs=#{allowed_cpus}
+SLICE_CONFIG
+
+      r "systemctl daemon-reload"
+    end
+  end
+
+  def start_systemd_unit
+    r "systemctl start #{@slice_name}"
+    r "echo \"root\" > /sys/fs/cgroup/#{@slice_name}/cpuset.cpus.partition"
```
macieksarnowicz, line 49: There is a different way of constructing the path that is used in the other PR for VM files (File.join). Should that be used here as well? I am not sure if there is a pro/con for one way vs. the other.

pykello: done

rhizome/host/lib/slice_setup.rb
pykello: done

PR #2422, opened 2024-12-13
Fix incorrect comparision between symbols and strings in deadlines

byucesoy:
In register_deadline function, we have a logic where we don't allow extending the deadline if the target is same. However, we were comparing symbols with strings to decide if the target is same, which always results in us deciding the targets are different and extending the deadline. This commit fixes it by using strings for deadline targets.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -81,7 +81,7 @@ SQL
     prog_label = "#{prog}.#{label}"
     Clog.emit("starting strand") { [self, {strand_started: {prog_label: prog_label}}] }
 
-    if label == stack.first["deadline_target"].to_s
+    if label == stack.first["deadline_target"]
       if (pg = Page.from_tag_parts("Deadline", id, prog, stack.first["deadline_target"]))
         pg.incr_resolve
       end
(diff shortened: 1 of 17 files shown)


migrate/20241204_vm_add_vcpus_memory.rb
```diff
@@ -2,9 +2,5 @@
 
 Sequel.migration do
   change do
-    alter_table(:vm) do
```
enescakir, line 5: Why did you change migrations

byucesoy: Migrations failed on my setup because they are applied in the wrong order. I made this change to fix them in my environment, but didn't intend to commit the changes. I reverted it back now.

enescakir:
@jeremyevans has a whole chapter in his book about symbols and strings. He might have some reviews, but I believe we should use strings whenever possible, especially when saving to the database. The database always returns strings, and using symbols complicates things

jeremyevans (commented):
As @enescakir mentioned, you should use strings if these values are stored in the database.  This also applies to other external storage cases, such as the session or flash.

byucesoy:
Yep, they are used in the database, so this PR starts to use strings instead of symbols.

enescakir (approved):

PR #2425, opened 2024-12-13
Fix Bundler.setup calls

jeremyevans:
First, Bundler.setup after require "bundler/setup" does nothing.
If you require "bundler/setup", it calls Bundler.setup, and
once Bundler.setup is called once, all future Bundler.setup calls
are ignored.

This switches require "bundler/setup" to require "bundler", and
then updates the Bundler.setup calls to only load the expected
groups.

In the loader, only load the :default and either the :test or
:development groups, depending on the rack environment.

This adds :rubocop and :lint Gemfile groups, to allow only
loading those specific gems, since the related gems are never
loaded directly by our code, only by tools run by the Rakefile.

For the linter:rubocop task, run the rubocop command line tool.
This allows the use of a normal task, and means you don't need
to require rubocop every time you run the Rakefile (and handle
failures).  This runs rubocop with BUNDLE_WITH so it only loads
the rubocop related gems.

Have the brakeman and erb_formatter tasks load only the :lint
group.

Note that I am not that knowledgeable about bundler, so this
should be thoroughly scrutinized.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,13 +35,14 @@ gem "argon2-kdf"
 
 group :development do
   gem "awesome_print"
-  gem "brakeman"
   gem "by", ">= 1.1.0"
-  gem "erb-formatter", github: "ubicloud/erb-formatter", ref: "a9ff0001a1eb028e2186b222aeb02b07c04f9808"
   gem "foreman"
   gem "pry-byebug"
   gem "rackup"
-  gem "sequel-annotate"
+  gem "cuprite"
+end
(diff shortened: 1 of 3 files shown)


enescakir (approved):
I'm curious about how dependabot will group them. We'll likely find out in a few days.

https://github.com/ubicloud/ubicloud/blob/0be89a73578fd6d06bd544457f4b3f8cd840a124/.github/dependabot.yml#L9

PR #2433, opened 2024-12-16
Add Slice and CPU limit parameters to `VmSetup`.

pykello:
Adds the following parameters to `VmSetup`'s methods:

* `slice_name`: cgroup slice name the VM should be a member of. Default is `system.slice`, which will be used by default until we migrate to a system with stricter resource management. Note that this works also on Ubuntu 22.04.
* `cpu_percent_limit`: cpu quota the VM can use.
* `cpu_burst_percent_limit`: additional quota that a VM can burst into temporarily if it has build up enough credit.

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -43,7 +43,10 @@ class VmSetup
     @vp ||= VmPath.new(@vm_name)
   end
 
-  def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology, mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices, boot_image, dns_ipv4)
+  def prep(unix_user, public_keys, nics, gua, ip4, local_ip4, max_vcpus, cpu_topology,
+    mem_gib, ndp_needed, storage_params, storage_secrets, swap_size_bytes, pci_devices,
+    boot_image, dns_ipv4, slice_name, cpu_percent_limit, cpu_burst_percent_limit)
+
     cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image, dns_ipv4)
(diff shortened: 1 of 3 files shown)


jeremyevans (commented):
From a general review of the Ruby code, I don't see any issues.

rhizome/host/lib/vm_setup.rb
```diff
@@ -655,9 +671,20 @@ def install_systemd_unit(max_vcpus, cpu_topology, mem_gib, storage_params, nics,
 LimitNOFILE=500000
 #{limit_memlock}
 SERVICE
+    r "systemctl set-property #{q_vm_service} CPUQuota=#{cpu_percent_limit}%" unless cpu_percent_limit == 0
```
furkansahin, line 674: instead of running this as a separate command, can we build the string and put it into the systemd service like we do for `#{limit_memlock}`

furkansahin (approved):

rhizome/host/lib/vm_setup.rb
```diff
@@ -654,10 +671,21 @@ def install_systemd_unit(max_vcpus, cpu_topology, mem_gib, storage_params, nics,
...
   end
 
+  def enable_bursting(slice_name, cpu_burst_percent_limit)
+    # Convert cpu_burst_percent limit to a usec value. This is the additional quota every
+    # 100,000 use that the VM is occasionally allowed to burst to.
+    cpu_burst_limit = cpu_burst_percent_limit * 1000
+
+    cpu_max_burst_path = File.join("/sys/fs/cgroup", slice_name, "#{@vm_name}.service", "cpu.max.burst")
+
+    r "echo #{cpu_burst_limit} > #{cpu_max_burst_path.shellescape}"
```
fdr, line 686: Can't it be `File.write(path, limit.to_s)`?

fdr (approved):

PR #2439, opened 2024-12-17
Use Roda host_routing plugin

jeremyevans:
This plugin was designed to handle Ubicloud's use case of multiple subdomains served by the same routes.

This updates the routes to use `r.web` and `r.api`, which are faster and simpler versions of `r.on web?` and `r.on api?`.

This requires Roda 3.87.0, just released, so update Gemfile for that.


enescakir (approved):

enescakir (approved):
One of the benefits of having Jeremy Evan on the team 😎

PR #2440, opened 2024-12-17
Setup for authorization refactoring PR

jeremyevans:
This splits off the first series of commits in #2384.  These are all before the authorization migration, and most of these are issues I found and fixed while working on #2384, but do not depend on the authorization changes.  

The most important changes:

* Fix referrer handling for repeated failures (previously, the code worked with 2 successive failures, but not more)
* Fix Dataset#authorize calls to use dataset_authorize (direct use of Dataset#authorize is a security issue for API requests using personal access tokens, since it doesn't enforce token permissions).
* Replacing Postgres:Firewall:{view,edit} permissions with  Postgres:{view,edit} 
* Vanity UBID creation support (not used by anything in this commit series, just for setup for the authorization migration)

I'm submitting this as a separate PR so this can get merged more quickly, to make any future rebasing of the authorization commit easier.  Also, one of the issues is security-related, and we should prioritize such fixes (even if few if any users are using personal access tokens yet).


enescakir (approved):
These are the changes I've already reviewed in the other draft PR.

PR #2442, opened 2024-12-18
Fix email verification resend and disable button if resent recently

enescakir:
- **Add extra attributes to the submit button**
  

- **Add disabled state to the button component**
  

- **Add missing login field to verification resend view**
  We use our custom views instead of the built-in Rodauth views. It looks
  like I forgot to add the login field [^1] to the verification resend
  view while migrating to custom views, which causes the resend
  verification email to fail.
  
  [^1]: https://github.com/jeremyevans/rodauth/blob/5bd816adf6252dca28465d5030e2484e04ed6120/templates/verify-account-resend.str#L5
  

- **Disable the resend button if the email was recently sent.**
  Thanks to Jeremy, the new version of Rodauth will include the public
  `verify_account_email_recently_sent?` method. We can use this to disable
  the resend button and provide a different explanation if the email was
  sent recently.
  
- **Do not render nil attributes for button component**
  I used `compact` instead of `compact!` because `compact!` returns `nil`for `{}`.


views/components/form/submit_button.erb
```diff
@@ -1,14 +1,12 @@
...
 <%== render(
   "components/button",
   locals: {
     text: text,
     extra_class: extra_class,
-    attributes: {
-      "type" => "submit",
-      "name" => name
-    }
+    attributes: { "type" => "submit", "name" => name }.merge!(attributes)
```
jeremyevans, line 10: Not a bug in this commit, but I think it would be better to only add the `name` attribute if it is not `nil`.  I guess having `name=""` in the HTML isn't a violation of anything, and it doesn't result in an parameter with an empty name being submitted, but it would still be something I would like to avoid.

jeremyevans (approved):

views/components/form/submit_button.erb
enescakir: That makes sense. I've added it to the button component because the submit_button uses it.
https://github.com/ubicloud/ubicloud/pull/2442/commits/77475e01db471725a22863d18c91e116abf23fed

enescakir:
@jeremyevans Nearly a month has passed since the latest Rodauth release. Do you plan on releasing a new version soon, or should I stick with the commit hash?

jeremyevans:
> @jeremyevans Nearly a month has passed since the latest Rodauth release. Do you plan on releasing a new version soon, or should I stick with the commit hash?

Unlike Roda and Sequel, Rodauth isn't released every month.  I would stick with the commit hash for the time being.

PR #2443, opened 2024-12-18
Add almalinux-9's new release.

pykello:
The URLs for old releases has changed, so in development environment we
were getting 404 errors.

So, this version adds the new version.


prog/download_boot_image.rb
```diff
@@ -52,7 +52,12 @@ def url
         "https://cloud.debian.org/images/cloud/bookworm/#{version}/debian-12-genericcloud-#{arch}-#{version}.raw"
       elsif image_name == "almalinux-9"
         arch = vm_host.render_arch(arm64: "aarch64", x64: "x86_64")
-        "https://repo.almalinux.org/almalinux/9/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-#{version}.#{arch}.qcow2"
+        case version
+        when "9.4-20240507", "9.4-20240805"
+          "https://repo.almalinux.org/almalinux/9.4/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-#{version}.#{arch}.qcow2"
+        else
+          "https://repo.almalinux.org/almalinux/9/cloud/#{arch}/images/AlmaLinux-9-GenericCloud-#{version}.#{arch}.qcow2"
+        end
```
enescakir, line 60: I don't think we should download the old version onto our hosts. Is this branching necessary?

pykello: Until we upload the new version to our image repo & update `Config.almalinux_9_version`, having `9.4-20240805` is necessary to not break dev environment. I can remove the `9.4-20240507` though.

What do you think?

enescakir: I've just uploaded them to our blob storage. The checksums appear to be the same.
- x64: https://github.com/ubicloud/runner-images/actions/runs/12391060761/job/34587384050
- arm64: https://github.com/ubicloud/runner-images/actions/runs/12391097018/job/34587488422

pykello: Nice. I changed the PR accordingly to include just the new version.

enescakir (approved):

PR #2448, opened 2024-12-18
Lock only the selected row instead of whole pool

velioglu:
While allocating a vm from a pool, whole vms of the related pool were locked. This was causing unnecessary lock and increasing the run time of the query. So, locking the selected row instead to allow concurrent requests to continue with the next unlocked tuple.

### DIFF
--- a/model/vm_pool.rb
+++ b/model/vm_pool.rb
@@ -12,15 +12,14 @@ class VmPool < Sequel::Model
   semaphore :destroy
 
   def pick_vm
-    DB.transaction do
-      # first lock the whole pool in the join table so that no other thread can
-      # pick a vm from this pool
-      vms_dataset.for_update.all
-      pick_vm_id_q = vms_dataset.left_join(:github_runner, vm_id: :id)
-        .where(Sequel[:github_runner][:vm_id] => nil, Sequel[:vm][:display_state] => "running")
-        .select(Sequel[:vm][:id])
-      Vm.where(id: pick_vm_id_q).first
-    end
+    # Find an available VM in the "running" state and not associated with a GitHub runner,
+    # and lock it with FOR UPDATE SKIP LOCKED.
+    vms_dataset
+      .where(Sequel[:vm][:display_state] => "running")
+      .exclude(id: DB[:github_runner].exclude(vm_id: nil).select(:vm_id))
+      .for_update
+      .skip_locked
+      .first
   end
 end
 
--- a/spec/model/vm_pool_spec.rb
+++ b/spec/model/vm_pool_spec.rb
@@ -36,12 +36,9 @@ RSpec.describe VmPool do
 
     it "returns the vm if there is one in running state" do
       locking_vms = class_double(Vm)
-      expect(pool).to receive(:vms_dataset).and_return(locking_vms).at_least(:once)
-      expect(locking_vms).to receive_message_chain(:for_update, :all).and_return([])  # rubocop:disable RSpec/MessageChain
-      vms_dataset = [vm]
-      expect(pool).to receive_message_chain(:vms_dataset, :left_join, :where, :select).and_return(vm.id) # rubocop:disable RSpec/MessageChain
-      expect(Vm).to receive(:where).and_return(vms_dataset) # rubocop:disable RSpec/MessageChain
-      expect(pool.pick_vm.id).to eq(vm.id)
+      expect(pool).to receive(:vms_dataset).and_return(locking_vms)
+      expect(locking_vms).to receive_message_chain(:where, :exclude, :for_update, :skip_locked, :first).and_return(vm) # rubocop:disable RSpec/MessageChain
+      expect(pool.pick_vm).to eq(vm)
     end
   end
 end


jeremyevans (commented):
I think the changes are OK (they make the code more efficient).  However, I think this method is subject to race conditions.  I think in both the new code and previous code, you could have the same `Vm` selected for multiple runners.  This is likely testable by having a sleep after the transaction call in `VmPool#pick_vm`

`FOR UPDATE SKIP LOCKED` only locks until the end of the transaction, and this transaction only returns the row without updating it, so that's an indication something is not right.

From looking at callers, I think the transaction `VmPool#pick_vm` should be moved to `Prog::Vm::GithubRunner#allocate_vm`, around the `pick_vm` and `github_runner.update` calls (and preferably around the two lines following).  That should ensure the `Vm` row is locked until the `GitHubRunner` is updated to reference the `Vm` row, preventing that `Vm` from being picked up by another runner.

Additionally, I think the `Vm.where(id: pick_vm_id_q).first` is unneeded.  The `vms_dataset` method chain should call `first` instead of `select.limit(1)`.

I recommend another developer also review, because I don't have that much production experience with SKIP LOCKED.

enescakir:
> I think the changes are OK (they make the code more efficient). However, I think this method is subject to race conditions. I think in both the new code and previous code, you could have the same `Vm` selected for multiple runners. This is likely testable by having a sleep after the transaction call in `VmPool#pick_vm`
> 
> `FOR UPDATE SKIP LOCKED` only locks until the end of the transaction, and this transaction only returns the row without updating it, so that's an indication something is not right.
> 
> From looking at callers, I think the transaction `VmPool#pick_vm` should be moved to `Prog::Vm::GithubRunner#allocate_vm`, around the `pick_vm` and `github_runner.update` calls (and preferably around the two lines following). That should ensure the `Vm` row is locked until the `GitHubRunner` is updated to reference the `Vm` row, preventing that `Vm` from being picked up by another runner.
> 
> Additionally, I think the `Vm.where(id: pick_vm_id_q).first` is unneeded. The `vms_dataset` method chain should call `first` instead of `select.limit(1)`.
> 
> I recommend another developer also review, because I don't have that much production experience with SKIP LOCKED.

@jeremyevans I have a PoC for an alternative approach from the past, you can ignore `has_customer_data` column for now. What are your thoughts on this method?https://github.com/ubicloud/ubicloud/pull/1583/commits/fc1fff731da8bc3647c986daab4969d6f545edb7

model/vm_pool.rb
```diff
@@ -13,13 +13,14 @@ class VmPool < Sequel::Model
...
-      vms_dataset.for_update.all
-      pick_vm_id_q = vms_dataset.left_join(:github_runner, vm_id: :id)
-        .where(Sequel[:github_runner][:vm_id] => nil, Sequel[:vm][:display_state] => "running")
-        .select(Sequel[:vm][:id])
-      Vm.where(id: pick_vm_id_q).first
+      # Find an available VM in the "running" state and not associated with a GitHub runner,
+      # and lock it with FOR UPDATE SKIP LOCKED.
+      vms_dataset
+        .where(Sequel[:vm][:display_state] => "running")
+        .exclude(id: DB[:github_runner].select(:vm_id))
```
enescakir, line 20:
```suggestion
        .exclude(id: DB[:github_runner].exclude(vm_id: nil).select(:vm_id))
```

model/vm_pool.rb
```diff
@@ -13,13 +13,14 @@ class VmPool < Sequel::Model
 
   def pick_vm
     DB.transaction do
```
enescakir, line 15: @jeremyevans Actually, respirate runs all labels within a transaction, making the transaction here redundant and unnecessary. The `GithubRunner.allocate_vm` label and the `GitHubRunner.pick_vm` method both operate within a transaction.

https://github.com/ubicloud/ubicloud/blob/8731845a4b463eac77c28400eefc7d0aa4ef8f18/model/strand.rb#L43

jeremyevans: Understood, so this is safe in actual usage, it just looks unsafe because of this transaction call.

jeremyevans:
> @jeremyevans I have a PoC for an alternative approach from the past, you can ignore `has_customer_data` column for now. What are your thoughts on this method?[fc1fff7](https://github.com/ubicloud/ubicloud/commit/fc1fff731da8bc3647c986daab4969d6f545edb7)

That approach, which updates and returns the row in a single query, seems like a much better approach for this.  It should prevent race conditions even in the absence of transactions.

velioglu:
> > @jeremyevans I have a PoC for an alternative approach from the past, you can ignore `has_customer_data` column for now. What are your thoughts on this method?[fc1fff7](https://github.com/ubicloud/ubicloud/commit/fc1fff731da8bc3647c986daab4969d6f545edb7)
> 
> That approach, which updates and returns the row in a single query, seems like a much better approach for this. It should prevent race conditions even in the absence of transactions.

With that approach, as I understand, all the requests would wait for the lock on same row lock as we get the cte with `.with(:candidate, idle_vms_dataset.limit(1))` . I think, the advantage of having row level skip lock is allowing all concurrent requests to pick next available vm without waiting on the row lock.

model/vm_pool.rb
velioglu: Removed the DB.transaction, I agree that it is confusing

enescakir (approved):

jeremyevans:
> With that approach, as I understand, all the requests would wait for the lock on same row lock as we get the cte with `.with(:candidate, idle_vms_dataset.limit(1))` . I think, the advantage of having row level skip lock is allowing all concurrent requests to pick next available vm without waiting on the row lock.

I haven't tried this, but shouldn't `idle_vms_dataset.limit(1).for_update.skip_locked` handle that issue?

velioglu:
As discussing with Enes about two approach, we agreed that this approach is easier to implement/follow than the one Enes shared above. So, I'm merging this one.

PR #2450, opened 2024-12-19
Migrations & Prog for VmHostSlice.

pykello:
### Migrations for VmHostSlice
We'll put VMs in systemd slices. This change adds the migrations to
enable that.

Each slice reserves a subset of host's cpus & part of its memory. Then,
VMs can be allocated into the slice. `allowed_cpus` column maps directly
to the `AllowedCPUs` property of the corresponding systemd slice.

`total_memory_gib` & `used_memory_gib` are used to make sure that we
will have enough memory in the host for the VMs that'll be added to the
slice in future. When a VmHostSlice record is allocated, it will reserve
`total_memory_gib` from the host.

Each slice can be either be `dedicated` or `shared`, as specified in the
`type` column. We will use `shared` slices to host burstable instances.

All VMs in the host will have same number of `cores`, but in a `shared`
slice each VM will have a quota of how much of its cpu's time it can
use. To keep track of this, `total_cpu_percent` & `used_cpu_percent` are
used. For a dedicated slice, all of the quota will be used for a single
VM.

### Prog to manage lifetime of a VmHostSlice
Adds `Prog::Vm::VmHostSliceNexus` which owns managing lifetime of a
VmHostSlice. It communicates with the `setup-slice` in the rhizome side
to achieve this.


macieksarnowicz:
@pykello - this description is not entirely true, please update it. 

> All VMs in the host will have same number of cores, but in a shared
> slice each VM will have a quota of how much of its cpu's time it can
> use. To keep track of this, total_cpu_percent & used_cpu_percent are
> used. For a dedicated slice, all of the quota will be used for a single
> VM.
That is true for how Burstable family is configured now, but the implementation allows to specify the Slice size to be larger than the number of vCPUs visible to the VM. Those number are now decoupled, that is why we have both `cores` and `vcpus` values in the VmSize. A possible future family (`Basic`) could run on a 2-core slice (4 vCPUs) with 3x or 4x overcommit, and each VM would still be allowed to see only 2 vCPUs and capped at 100% or 200% of CPU capacity. All those are independent knobs to tune and can be used to define different families.

macieksarnowicz:
@pykello - after a discussion around @byucesoy change for billing rates, it might be better to rename all cpu_* fields to vcpu_* to keep it consistent with the general naming convention used elsewhere. It would be easier to do it now rather than later. 

What do you both think about it?

fdr:
> @pykello - after a discussion around @byucesoy change for billing rates, it might be better to rename all cpu_* fields to vcpu_* to keep it consistent with the general naming convention used elsewhere. It would be easier to do it now rather than later.
> 
> What do you both think about it?

Is it accurate? vCPU refers to a specific phenomena, the state of a virtual CPU. You can have thirty vCPUs and one Linux CPU/SMT thread executing on hardware at a time. Seems simply wrong in this context?

PR #2453, opened 2024-12-19
add integer to same-day migrations to enforce ordering

geemus:
Without this change, `dev_up` and `test_up` would consistently fail for me when it would try to run these migrations in the opposite order (as shown below). So this adds a number to clarify and enforce the expected order.

```
I, [2024-12-19T08:13:37.770018 #313906]  INFO -- : (0.000292s) SELECT NULL AS "nil" FROM "schema_migrations" LIMIT 1
I, [2024-12-19T08:13:37.770193 #313906]  INFO -- : (0.000070s) SELECT * FROM "schema_migrations" LIMIT 0
I, [2024-12-19T08:13:37.770714 #313906]  INFO -- : (0.000402s) SELECT "filename" FROM "schema_migrations" ORDER BY "filename"
I, [2024-12-19T08:13:37.787004 #313906]  INFO -- : Begin applying migration 20241204_vm_enforce_vcpus_memory.rb, direction: up
I, [2024-12-19T08:13:37.787342 #313906]  INFO -- : (0.000231s) BEGIN
E, [2024-12-19T08:13:37.787882 #313906] ERROR -- : PG::UndefinedColumn: ERROR:  column "vcpus" does not exist
LINE 12:     WHERE vcpus IS NULL OR memory_gib IS NULL;
                   ^:     UPDATE vm
    SET
      vcpus = CASE
        WHEN arch = 'arm64' THEN cores
        ELSE 2 * cores
      END,
      memory_gib = CASE
        WHEN arch = 'arm64' THEN 3.2 * cores
        WHEN family = 'standard-gpu' THEN 10.68 * cores
        ELSE 8 * cores
      END
    WHERE vcpus IS NULL OR memory_gib IS NULL;

I, [2024-12-19T08:13:37.788153 #313906]  INFO -- : (0.000099s) ROLLBACK
rake aborted!
Sequel::DatabaseError: PG::UndefinedColumn: ERROR:  column "vcpus" does not exist (Sequel::DatabaseError)
LINE 12:     WHERE vcpus IS NULL OR memory_gib IS NULL;
                   ^
migrate/20241204_vm_enforce_vcpus_memory.rb:7:in `block (2 levels) in <top (required)>'
/home/geemus/github/geemus/ubicloud/Rakefile:29:in `block in <top (required)>'
/home/geemus/github/geemus/ubicloud/Rakefile:88:in `block in <top (required)>'
/home/geemus/.asdf/installs/ruby/3.2.6/bin/bundle:25:in `load'
/home/geemus/.asdf/installs/ruby/3.2.6/bin/bundle:25:in `<main>'

Caused by:
PG::UndefinedColumn: ERROR:  column "vcpus" does not exist (PG::UndefinedColumn)
LINE 12:     WHERE vcpus IS NULL OR memory_gib IS NULL;
                   ^
migrate/20241204_vm_enforce_vcpus_memory.rb:7:in `block (2 levels) in <top (required)>'
/home/geemus/github/geemus/ubicloud/Rakefile:29:in `block in <top (required)>'
/home/geemus/github/geemus/ubicloud/Rakefile:88:in `block in <top (required)>'
/home/geemus/.asdf/installs/ruby/3.2.6/bin/bundle:25:in `load'
/home/geemus/.asdf/installs/ruby/3.2.6/bin/bundle:25:in `<main>'
Tasks: TOP => dev_up
(See full trace by running task with --trace)
```


enescakir:
It makes sense. However, we need to rename the migration that's already been applied in the `schema_migrations` table at production. If we don't, the production or any existing local development environment will encounter an error due to an unknown migration file exception.

geemus:
@enescakir yeah. I'm not sure what exactly the right fix is either, as this change made it work (from a fresh db) for me but it fails to work from a fresh db in CI. So there must be something that is causing the order to be reversed between the two places.

jeremyevans:
I fixed this non determinism issue in Sequel yesterday, updating the Gemfile to use Sequel master should fix things.

geemus:
Should be solved by updating sequel.

PR #2454, opened 2024-12-19
Dev db setup revisions

geemus:
An effort to simplify the setup process a bit. Also, I think setup_database can then be reused (since it will drop/recreate) whenever you would want/need.

It might make sense to still have some migration instructions in here somewhere, but it would no longer be required to get started (since setup_database includes migration). I dropped it for now since I wasn't surewhere to put it or how to frame it given what is there presently.


DEVELOPERS.md
```diff
@@ -371,22 +369,6 @@ Some useful corrections are only made with `bundle exec rubocop -A`
 (upper case `A`) which applies "unsafe" corrections that may alter the
 semantics of the program.
 
-### Running the migrations
-
-Empty databases will cause Clover to crash.  You can run database
```
jeremyevans, line 376: This section should be reworded, not removed.  Developers should not be running `setup_database` for every migration, it's unnecessary to rerun all migrations.  test_up/dev_up should still be recommended for use after initial setup.

jeremyevans (changes requested):
This looks good, except for the removal of the test_up/dev_up instructions.

DEVELOPERS.md
geemus: I rewrote the commit to include an updated version of this (see above).

DEVELOPERS.md
```diff
@@ -258,6 +256,19 @@ creating and destroying VMs this way.  Looking at the `clover_test`
 database is rare, usually when working on or debugging the testing
 infrastructure itself.
 
+### Running migrations
+
+The `setup_database` task will drop databases if they exist, create databases and then migrate them.
+
+You can run also run just the migrations to update to the latest schema with rake:
```
jeremyevans, line 263:
```suggestion
If you have already setup the databases, and you want to run new migrations to update them
to the latest schema:
```

jeremyevans (approved):
I think we should be more forceful in recommending `test_up`/`dev_up` for migrations, but otherwise looks looks good.

enescakir (approved):

PR #2457, opened 2024-12-20
Pass boot_images to `Prog::Test::VmGroup.assemble()`

pykello:
3rd party developers don't have permissions to download all boot images, so selecting images randomly from `Option::BootImages` doesn't work well for them.

This change allows a list of boot image names to be passed to workaround this problem:

```
 > st = Prog::Test::VmGroup.assemble(
    test_reboot: false,
    boot_images: vmh.boot_images.map { |x| x.name })
```

### DIFF
--- a/prog/test/vm_group.rb
+++ b/prog/test/vm_group.rb
@@ -2,15 +2,18 @@
 
 require "net/ssh"
 
+DEFAULT_BOOT_IMAGE_NAMES = Option::BootImages.map { _1.name }.freeze
+
 class Prog::Test::VmGroup < Prog::Test::Base
-  def self.assemble(storage_encrypted: true, test_reboot: true)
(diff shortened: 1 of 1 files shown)


prog/test/vm_group.rb
```diff
@@ -3,14 +3,15 @@
 require "net/ssh"
 
 class Prog::Test::VmGroup < Prog::Test::Base
-  def self.assemble(storage_encrypted: true, test_reboot: true)
+  def self.assemble(storage_encrypted: true, test_reboot: true, boot_images: Option::BootImages.map { _1.name })
```
jeremyevans, line 6: If the boot images don't change while the app is running, it would be good to store this `Option::BootImages.map { _1.name }` as a frozen constant, to avoid allocating every time this is called.

jeremyevans (approved):

PR #2459, opened 2024-12-20
Use vCPU in billing rates and quotas instead of cores

byucesoy:
Our customers are exposed to CPUs not cores. This will be more prominent after we introduce the burstable instances where we sell percentage of CPUs. We discussed this with @macieksarnowicz as well and it seems it would be better to use CPU as a common term in customer adjacent places.

This PR also opportunistically cleans up some unused billing rates and fixes a typo.


config/billing_rates.yml
```diff
@@ -12,7 +12,6 @@
 - { id: d31188b8-ced8-4bab-bb73-ee6b0c6a4806, resource_type: VmCores,                 resource_family: standard,            location: hetzner-hel1,   unit_price: 0.0005740000, billed_by: duration, active_from: 2024-08-22T00:00:00Z }
 - { id: a63fc107-176f-41e6-baf6-a1399446f616, resource_type: VmCores,                 resource_family: standard,            location: github-runners, unit_price: 0.0005257320, billed_by: duration, active_from: 2024-05-27T00:00:00Z }
 - { id: 724069e2-fa28-46d9-918c-666d1fbe80ba, resource_type: VmCores,                 resource_family: standard,            location: leaseweb-wdc02, unit_price: 0.0008000992, billed_by: duration, active_from: 2024-10-07T00:00:00Z }
-- { id: 534e85b6-2ba2-44c7-b799-8a1b99f8cd50, resource_type: VmCores,                 resource_family: standard,            location: leaseweb-dal13, unit_price: 0.0008000992, billed_by: duration, active_from: 2024-05-27T00:00:00Z }
```
enescakir, line 15: It's good to see no active billing records. However, we keep billing records indefinitely. What would happen if we wanted to calculate usage for these records again? It seems like a BillingRecord without a BillingRate is somewhat useless.

config/billing_rates.yml
```diff
@@ -41,7 +41,7 @@
 - { id: 159ffbe1-16ba-8978-aa38-164b7230d392, resource_type: PostgresCores,           resource_family: paradedb,            location: leaseweb-wdc02, unit_price: 0.0025448500, billed_by: duration, active_from: 2024-10-07T00:00:00Z }
 - { id: c7d24eb6-0c24-417d-8d81-f7558553f1a3, resource_type: PostgresCores,           resource_family: lantern,             location: hetzner-fsn1,   unit_price: 0.0018786380, billed_by: duration, active_from: 2024-10-10T00:00:00Z }
 - { id: df442872-a4a1-4937-9afe-7afc3f048d8b, resource_type: PostgresCores,           resource_family: lantern,             location: leaseweb-wdc02, unit_price: 0.0018786380, billed_by: duration, active_from: 2024-10-10T00:00:00Z }
-- { id: 46db9ff6-7f5a-4942-bd1a-6feccfe6c970, resource_type: PostgresStandbyCores,    resource_family: standard,            location: hetzner-fsn1,   unit_price: 0.0012524250, billed_by: duration, active_from: 2023-01-01T00:00:00Z }
```
enescakir, line 51: Should we introduce a new billing rate for this typo fix? It will change the total usage of the billing records, causing discrepancies between the invoices and the billing records.

lib/billing_rate.rb
```diff
@@ -26,15 +26,15 @@ def self.from_id(billing_rate_id)
 
   def self.line_item_description(resource_type, resource_family, amount)
     case resource_type
-    when "VmCores"
+    when "VmCores", "VmCpu"
       "#{resource_family}-#{(amount * 2).to_i} Virtual Machine"
```
enescakir, line 30: It shows `amount * 2`, it will show `standard-4` for 2 vCPUs machines

config/default_quotas.yml
```diff
@@ -1,4 +1,4 @@
-- { id: c088f732-0a30-454b-aa6b-542ce1d67bfb, resource_type: VmCores,                  new_value: 16,  verified_value: 128 }
-- { id: 14fa6820-bf63-41d2-b35e-4a4dcefd1b15, resource_type: GithubRunnerCores,        new_value: 150, verified_value: 150 }
-- { id: 0ac764de-48c5-4432-876d-389878da0885, resource_type: GithubRunnerCacheStorage, new_value: 30,  verified_value: 30 }
-- { id: 91d616ea-a15b-4d54-90b7-aaa1e1bd2f19, resource_type: PostgresCores,            new_value: 64,  verified_value: 128 }
+- { id: c088f732-0a30-454b-aa6b-542ce1d67bfb, resource_type: VmCpu,                    new_value: 32,   verified_value: 256 }
+- { id: 14fa6820-bf63-41d2-b35e-4a4dcefd1b15, resource_type: GithubRunnerCpu,          new_value: 300,  verified_value: 300 }
```
enescakir, line 2: Should we update the existing quota records?

lib/validation.rb
```diff
@@ -234,12 +234,12 @@ def self.validate_url(url)
...
-  def self.validate_core_quota(project, resource_type, requested_core_count)
-    if !project.quota_available?(resource_type, requested_core_count)
-      current_used_core_count = project.current_resource_usage(resource_type)
+  def self.validate_cpu_quota(project, resource_type, requested_cpu_count)
+    if !project.quota_available?(resource_type, requested_cpu_count)
+      current_used_cpu_count = project.current_resource_usage(resource_type)
       effective_quota_value = project.effective_quota_value(resource_type)
 
-      fail ValidationFailed.new({size: "Insufficient quota for requested size. Requested core count: #{requested_core_count}, currently used core count: #{current_used_core_count}, maximum allowed core count: #{effective_quota_value}, remaining core count: #{effective_quota_value - current_used_core_count}"})
+      fail ValidationFailed.new({size: "Insufficient quota for requested size. Requested CPU count: #{requested_cpu_count}, currently used CPU count: #{current_used_cpu_count}, maximum allowed CPU count: #{effective_quota_value}, remaining CPU count: #{effective_quota_value - current_used_cpu_count}"})
```
enescakir, line 242: If I remember correctly, @fdr thought we should stick to using `core` and `vCPU`. The `CPU` was making things more complicated. Why did we decide to use `CPU` instead of `vCPU`?

fdr: well vCPU is definitely unambiguously correct, it means one thread state regardless of how it is multiplexed onto hardware. We could make it a house rule that `CPU` = `vCPU`, I don't have any opinion about that.

Note that some of the newer Xeon chips (Sierra Forest) on x64 also do away with SMT, so life gets even harder or three years from now. But the conceit of the patch overall is basically right: my dogmatic adherence to cores only holds for when you want to compare _dedicated_ compute resources, eliminating the 2x noise from SMT that doesn't result in 2x performance, to help people compare things (e.g. an arm-1 roughly equal to a x64-1, now we have the problem all the other providers have: arm-2 is better compared to x64-4)

fdr: Anyway, this is my stance on core vs vcpu, for which I have thrown in the towel ages ago now:

image

config/billing_rates.yml
byucesoy: Makes sense. I only deleted the ones with no billing record (active or inactive) and moved the rest to the deprecated section.

config/default_quotas.yml
byucesoy: Added a commit for migration.

lib/validation.rb
byucesoy: On CPU vs vCPU; I changed this to vCPU to not introduce 3rd option into the mix.

On Core vs vCPU... yeah that is complicated.

enescakir (approved):

enescakir (approved):

PR #2463, opened 2024-12-21
Add social login

enescakir:
- ### Add social login

I used Janko's omniauth plugin for Rodauth.

Social Buttons

- ### Reduce duplication for before/after create account for standard and omniauth cases

- ### Ignore new warnings while running tests

    /Users/enescakir/.asdf/installs/ruby/3.2.6/lib/ruby/gems/3.2.0/gems/capybara-3.40.0/lib/capybara/session/config.rb:95: warning: URI::RFC3986_PARSER.make_regexp is obsolete. Use URI::RFC2396_PARSER.make_regexp explicitly.

    /Users/enescakir/.asdf/installs/ruby/3.2.6/lib/ruby/gems/3.2.0/gems/omniauth-2.1.2/lib/omniauth/strategy.rb:484: warning: URI::ABS_URI is obsolete. Use RFC2396_PARSER.regexp[:ABS_URI] explicitly.

- ### Add identities association to the account

I also refactored test to mock different omniauth providers.

- ### Do not link social identities to existing accounts without logging in

If an account already exists and it's being accessed for the first time via
social identity, we should not automatically link it to this identity. Existing
account holders should be able to link their social identites only after
they've logged in. This way, we can ensure that users are able to securely
connect their accounts. In the current situation, if I don't link my GitHub
account to Ubicloud and my GitHub account gets hacked, the attacker can log in
to Ubicloud. However, in the proposed situation, the attacker would also need
to know my Ubicloud password. It's not a case for newly created accounts.

- ### Allow existing users to connect their social accounts

In the previous commit, we prevented existing users from connecting
their social accounts without logging in first for security reasons.

This PR adds a new tab to the "My Account" page that allows existing
users to connect and disconnect their social accounts.

If the email of the current account does not match the social account,
we do not allow the operation.

We also do not allow disconnecting the last login method. However, we
can allow disconnecting the last omniauth provider if the user has a
password. The `has_password?` method is not yet public in Rodauth, so I
created a PR for it.

Login Methods

- ### Update Rodauth to the latest commit
  The new version has `has_password?` as public.
  

- ### Allow to disconnect the last omniauth provider if has password
  

- ### Allow to delete password if another login method is available
  If the user has another login method, they can delete their password.
  This is helpful for users who prefer to use only one login method.
  
  If the user doesn't have a password, they can create a new one using the
  "Change Password" feature. I changed "Change" to "Create" for clarity
  when the user doesn’t have a password.
  

- ### Do not allow to reset password if password is disabled


jeremyevans (commented):
Looks good.  For reducing duplication, I submitted #2468.

clover.rb
```diff
@@ -356,6 +356,35 @@ def assets
...
+    end
+    # :nocov:
+
+    before_omniauth_create_account do
+      account[:id] = Account.generate_uuid
+      account[:name] = omniauth_name
+      Validation.validate_account_name(account[:name])
+    end
+
+    after_omniauth_create_account do
```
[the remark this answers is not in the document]
jeremyevans: I decided to define a helper method in route scope for this, and have the rodauth hooks call it.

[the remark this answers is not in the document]
enescakir: Thanks for doing it

enescakir:
I rebased from main, and add 373dabef0f5d3c1e048896c3e5cce073c500c04b commit

jeremyevans (approved):
I tried this using the test apps you setup, and it works great and significantly reduces signup friction.

fdr:
If my github account is not secure -- bad password and/or no 2FA -- but my Google account (implied: connected to my email and passwords resets; that is, a compromised account is tantamount to game over), does this expose my account when it ships?

enescakir:
I added two new improvements, allowing users to choose which social login providers to use after logging in.

- Do not link social identities to existing accounts without logging in
- Allow existing users to connect their social accounts

clover.rb
```diff
@@ -323,7 +323,13 @@ def assets
     login_label "Email Address"
     two_factor_auth_return_to_requested_location? true
     already_logged_in { redirect login_redirect }
-    after_login { remember_login if request.params["remember-me"] == "on" }
+    after_login do
+      remember_login if request.params["remember-me"] == "on"
+      if omniauth_identity && (url = omniauth_params["redirect_url"])
+        flash["notice"] = "You have successfully connected your account with #{omniauth_provider.capitalize}."
```
jeremyevans, line 329: Is this message still accurate?  With the new code, you connect social identities after login and not during the login process, but maybe that connection process runs though this code?

clover.rb
```diff
@@ -342,19 +348,41 @@ def assets
...
+    before_omniauth_create_account do
+      scope.before_rodauth_create_account(account, omniauth_name)
+    end
+
+    after_omniauth_create_account do
+      scope.after_rodauth_create_account(account_id)
+    end
+
+    before_omniauth_callback_route do
+      account = Account[account_from_omniauth&.dig(:id)]
```
jeremyevans, line 377:
```suggestion
      account = Account[account_from_omniauth&.[](:id)]
```

It's not necessary to use `dig` for a single attribute.

clover.rb
```diff
@@ -342,19 +348,41 @@ def assets
...
+      if authenticated?
+        if (account && account.id != scope.current_account.id) || !omniauth_email.casecmp?(scope.current_account.email)
+          flash["error"] = "Your account's email address is different from the email address associated with the #{omniauth_provider.capitalize} account."
```
jeremyevans, line 380: If users are manually linking accounts and we are not linking them automatically, does it matter if the email address for the account is different?

clover.rb
```diff
@@ -342,19 +348,41 @@ def assets
...
+          redirect "/account/login-method"
         end
-        inv.destroy
+      elsif account && account.identities_dataset.where(provider: omniauth_provider.to_s).empty?
+        flash["error"] = "This account is already exists with another provider. Please login with the existing account and link the new provider from your account settings."
```
jeremyevans, line 384: I don't think the wording here is correct.  You get this message when there is an existing account for the email not tied to the social login, and you try to use social login for it.  Maybe:

```suggestion
        flash["error"] = "There is already an account with this email address, and it has not been linked to the #{omniauth_provider.capitalize} account. Please login to the existing account normally, and then link it to the #{omniauth_provider.capitalize} account from your account settings.  Then you can can login using the #{omniauth_provider.capitalize} account."
```

routes/account.rb
```diff
@@ -2,8 +2,37 @@
...
+
+          view "account/login_method"
+        end
+
+        r.post "disconnect" do
+          provider, uid = typecast_params.nonempty_str(["provider", "uid"])
+          identities = current_account.identities
+          unless identities.length > 1
+            # YYYY: We can allow to disconnect the last omniauth provider if the user has a password
+            # But need https://github.com/jeremyevans/rodauth/pull/461 to be merged first.
```
jeremyevans, line 22: It has been merged, so we can now allow disconnecting the last omniauth provider, though we'll need to update the rodauth checkout hash in the Gemfile.

jeremyevans (commented):
I agree we shouldn't automatically link social identities, and that the choice to link manually should be the user's.

I'm not a fan of how many dependencies this adds, but I suppose that is unavoidable.

clover.rb
enescakir: We only set `omniauth_params["redirect_url"]` on the Login Methods page. If it's initiated from the social login buttons on the login form, it won’t have this parameter.

https://github.com/ubicloud/ubicloud/blob/c151a223e464a5445773238119a4aa5856f0eb5a/views/account/login_method.erb#L31

clover.rb
enescakir: I remembered you used `&` and `[]` together in one of your PRs, but I couldn't find the example and couldn't use `[]` with `&`. Thanks for that; I applied it

clover.rb
enescakir: Thanks, applied it.

clover.rb
enescakir: The problem is that `rodauth-omniauth` initially attempts to find the account using the provider and UID, then tries using email. If it can't find an account, it simply creates a new one with the new email and logs into it. As a result, you might unintentionally switch to a different account.

https://github.com/janko/rodauth-omniauth/blob/3e8b627b8b3d21cc382f4a9d421fceef483b5d72/lib/rodauth/features/omniauth.rb#L98

I can override some methods, and for authenticated users, we can use the current user's account ID instead of creating a new one. However, to be on the safe side, I decided not to allow different emails. For example, I shouldn't connect my personal Gmail account to my Ubicloud account with my business email.


I've simplified this check a bit. Since `account_from_omniauth` finds the account by email, there's no need to separately check the email. If the account isn't found, it means the emails are different.

```
# before
if (account && account.id != scope.current_account.id) || !omniauth_email.casecmp?(scope.current_account.email)

# after
unless account && account.id == scope.current_account.id
```

I also added `omniauth_create_account? { !authenticated? }`. So it doesn't allow to create a new account accidentally, for `authenticated?` users.

routes/account.rb
enescakir: Since you've already reviewed this PR, I've created a new one for it.

https://github.com/ubicloud/ubicloud/pull/2485

enescakir:
> If my github account is not secure -- bad password and/or no 2FA, but is associated with a particular email -- but my Google account (implied: connected to my email and passwords resets; that is, a compromised account is tantamount to game over) is secure, does this nevertheless expose my account when it ships?

@fdr, with the last two commits, if the user hasn't connected their GitHub account to Ubicloud, the attacker can't log in. We don't link social identities to existing accounts without logging in, so this doesn't make existing accounts less secure when shipped.

If the user links their GitHub account, that's a different story. I will add MFA and SSO enforcement in a future PR after deployment.

fdr:
> > If my github account is not secure -- bad password and/or no 2FA, but is associated with a particular email -- but my Google account (implied: connected to my email and passwords resets; that is, a compromised account is tantamount to game over) is secure, does this nevertheless expose my account when it ships?
> 
> @fdr, with the last two commits, if the user hasn't connected their GitHub account to Ubicloud, the attacker can't log in. We don't link social identities to existing accounts without logging in, so this doesn't make existing accounts less secure when shipped.
> 
> If the user links their GitHub account, that's a different story. I will add MFA and SSO enforcement in a future PR after deployment.

I'm cool with doing improvements later so long as we don't link everything by email by default. We can muddle our way towards something balanced between customer learning curve and security by using email challenges to let them change their SSO preference, tinkering with the exact presentation of that choice over time. Thank you.

migrate/20241221_add_social_login.rb
jeremyevans, line 1: Doesn't the migration have to be added in its own commit first, before anything uses it, since we migrate after merging and not before?

jeremyevans (approved):
This looks good, other than the migration needing to be split into a separate commit, in order to allow migration before these code changes.  I'm not sure whether that is strictly necessary if the environment variables haven't been added to the application, but it seems a like a good idea to do so here, since we do so in other cases.

migrate/20241221_add_social_login.rb
enescakir: Yes, you're right. I've created a separate commit for it.

PR #2465, opened 2024-12-22
Remove 28 days monthly cap

byucesoy:
We cap the number of days in month at 28 for billing purposes. This ensures that users see same price every month in their bill but it complicates our billing calculations. This commit removes the logic so now the amount we would charge increase linearly till the end of the month. In previous PRs, we reduced the minutely rates to accommodate this change except for the IP address. This PR reduces the rate for IP address as well.


enescakir (approved):

byucesoy:
We decided to keep monthly cap.

PR #2468, opened 2024-12-23
Reduce duplication for before/after create account for standard and omniauth cases

jeremyevans:
Also, remove blank line at end of Gemfile to appease Rubocop.


enescakir (approved):

fdr:
I'm surprised things passed rubocop and then failed from these changes? Mostly I use rubocop for `-A` (but noting it is unsafe, I always have my changes at least staged if not committed), and only rarely for anything else (telling me about bad things I'm doing, which is rarely useful)

PR #2469, opened 2024-12-23
Return error json without error key for runtime endpoints

velioglu:
Returning an error json without an error key on top for runtime endpoints. It is required because docker layer caching expects error in this format to skip duplicate cache entries. Passing it in this way works properly with the cache proxy as we directly pass the error to the client without parsing it.

### DIFF
--- a/spec/routes/runtime/github_spec.rb
+++ b/spec/routes/runtime/github_spec.rb
@@ -14,14 +14,14 @@ RSpec.describe Clover, "github" do
     it "vm has no runner" do
       get "/runtime/github"
 
-      expect(last_response).to have_api_error(400, "invalid JWT format or claim in Authorization header")
+      expect(last_response).to have_runtime_error(400, "invalid JWT format or claim in Authorization header")
     end
 
     it "vm has runner but no repository" do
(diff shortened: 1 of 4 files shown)


spec/routes/runtime/github_spec.rb
```diff
@@ -67,35 +67,39 @@
...
       it "fails if the cache entry already exists" do
         GithubCacheEntry.create_with_id(key: "k1", version: "v1", scope: "dev", repository_id: repository.id, created_by: runner.id, committed_at: Time.now)
         post "/runtime/github/caches", {key: "k1", version: "v1", cacheSize: 100}
 
-        expect(last_response).to have_api_error(409, "A cache entry for dev scope already exists with k1 key and v1 version.")
+        expect(last_response.status).to eq(409)
+        response = JSON.parse(last_response.body)
+        expect(response["code"]).to eq(409)
+        expect(response["type"]).to eq("AlreadyExists")
+        expect(response["message"]).to eq("A cache entry for dev scope already exists with k1 key and v1 version.")
```
enescakir, line 95: nit: 

```suggestion
        expect(last_response).to have_runtime_error(409, "A cache entry for dev scope already exists with k1 key and v1 version.")
```

enescakir (approved):

PR #2470, opened 2024-12-23
Merge some code to prepare for new authorization system

jeremyevans:
This still uses the old authorization system, but merges changes required by the new authorization system that will not conflict.  This will allow breaking the switchover into smaller chunks.


model/project.rb
```diff
@@ -70,6 +72,9 @@ def soft_delete
     DB.transaction do
       access_tags_dataset.destroy
       access_policies_dataset.destroy
+      access_control_entries_dataset.destroy
+      DB[:applied_subject_tag].where(tag_id: subject_tags_dataset.select(:id)).delete
```
enescakir, line 76: Should we delete action and object tags too?

model/project.rb
```diff
@@ -5,6 +5,8 @@
 class Project < Sequel::Model
   one_to_many :access_tags
   one_to_many :access_policies
+  one_to_many :access_control_entries
+  one_to_many :subject_tags
```
enescakir, line 9: We might include the commit that adds also `object_tags`, `action_tags` to this PR too.

enescakir (approved):

model/project.rb
jeremyevans: Yes, that was an oversight, and looks like it wasn't in the main PR.  I updated this PR to include it, and will rebase the main PR on this PR.

model/project.rb
jeremyevans: That makes sense.  I updated the last commit in this PR to do that.

PR #2474, opened 2024-12-24
Use parallel var instead of raw arg in the setup_database rake task

enescakir:
We added a `false` option for `parallel` in
b9fa15e069fc7e825c3508f5cc823183c1e3ae78, but it seems we aren't using it everywhere. This causes `clover_test1` to be created instead of `clover_test` when running `rake setup_database\[test,false\]`


enescakir:
I can't add you, @geemus, as reviewer. jfyi

jeremyevans (approved):

PR #2476, opened 2024-12-24
Add policy selector to user list

enescakir:
- **Make "Member" policy default for user invitation**
  

- **Run policy updates in transaction**
  

- **Add subject tags of user as policy to serializer**
  It will be used in the next commit.
  

- **Show all subject tags in the policy selector**
  If the current user doesn't have permission for a subject tag, it won't
  appear in the policy selector.

  This is an issue because the selector doesn't provide an option for that
  tag, which is interpreted as no policy option. We should display all
  subject tags in the policy selector, but if the user lacks permission to
  add one of them, it should be disabled.

- **Do not allow to invite with unauthorized subject tag**

- **Add space between flash messages when multiple types are shown**

| Before | After |
|--------|-------|
| CleanShot 2024-12-25 at 16 20 30@2x | CleanShot 2024-12-25 at 16 23 38@2x |

- **Fix invitation policy update**
   While testing, I found two issues:
   -  Even if the current user doesn't have the "SubjectTag:remove"
   permission, they can still remove the subject tag from the
   invitation.
   -  The `allowed_tags[policy]` always returns nil for `No Policy`, so we
   can't update the invitation policy to `No Policy`.

- **Add policy selector to user list**
  Most users don't need to know about our access control system and
  usually don't visit the access control page.
  
  It's better to add a policy selector to the user list page. This way,
  most users can choose from predefined subject tags without needing to go
  to the access control page.
  
  If a user has more than one subject tag, we won't show the policy
  selector. If so, probably the user is aware of the access control system
  and manages access control via its page.
  
| Before | After |
|--------|-------|
| CleanShot 2024-12-25 at 15 21 28@2x | CleanShot 2024-12-25 at 15 21 51@2x |

### DIFF
--- a/routes/project/user.rb
+++ b/routes/project/user.rb
@@ -13,7 +13,7 @@ class Clover
         authorize("Project:user", @project.id)
 
         r.get do
-          @users = Serializers::Account.serialize(@project.accounts_dataset.order_by(:email).all)
+          @users = Serializers::Account.serialize(@project.accounts_dataset.order_by(:email).all, {project_id: @project.id})
           @invitations = Serializers::ProjectInvitation.serialize(@project.invitations_dataset.order_by(:email).all)
           view "project/user"
         end
--- a/routes/project/user.rb
+++ b/routes/project/user.rb
@@ -176,14 +176,15 @@ class Clover
       r.post "policy/managed" do
         authorize("Project:user", @project.id)
 
-        invitation_policies = r.params["invitation_policies"] || {}
-        allowed_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").select_hash(:name, Sequel.as(true, :v))
-        invitation_policies.each do |email, policy|
-          if allowed_tags[policy]
-            @project.invitations.find { _1.email == email }&.update(policy: policy)
+        DB.transaction do
+          invitation_policies = r.params["invitation_policies"] || {}
+          allowed_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").select_hash(:name, Sequel.as(true, :v))
+          invitation_policies.each do |email, policy|
+            if allowed_tags[policy]
+              @project.invitations.find { _1.email == email }&.update(policy: policy)
+            end
           end
         end
-
         flash["notice"] = "Subject tags for invited users updated successfully."
 
         r.redirect "#{@project.path}/user"
--- a/routes/project/user.rb
+++ b/routes/project/user.rb
@@ -184,18 +184,68 @@ class Clover
         DB.transaction do
           allowed_add_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").to_hash(:name)
           allowed_remove_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:remove").to_hash(:name)
+          additions = {}
+          removals = {}
+          issues = []
+          user_policies = r.params["user_policies"] || {}
+          user_policies.each do |ubid, policy|
+            policy = nil if policy == ""
+            user = Account.from_ubid(ubid)
+            existing_tags = user.subject_tags_dataset.where(project_id: @project.id).select_map(:name)
+            unless existing_tags.length < 2
+              issues << "Cannot change the policy for user, as they are in multiple subject tags"
+              next
+            end
+            next if existing_tags.include?(policy)
+            unless (added_tag = allowed_add_tags[policy]) || !policy
+              issues << "You don't have permission to add members to '#{policy}' tag"
+              next
+            end
+            if (tag = existing_tags.first)
+              unless (removed_tag = allowed_remove_tags[tag])
+                issues << "You don't have permission to remove members from '#{tag}' tag"
+                next
+              end
+              (removals[removed_tag.name] ||= []) << user.id
+            end
+            (additions[added_tag.name] ||= []) << user.id if added_tag
+          end
+
+          additions.each { |name, user_ids| @project.subject_tags_dataset.first(name:).add_members(user_ids) }
+          removals.each { |name, user_ids| @project.subject_tags_dataset.first(name:).remove_members(user_ids) }
+
+          if @project.subject_tags_dataset.first(name: "Admin").member_ids.empty?
+            flash["error"] = "The project must have at least one admin."
+            DB.rollback_on_exit
+            r.redirect "#{@project.path}/user"
+          end
 
           invitation_policies = r.params["invitation_policies"] || {}
           invitation_policies.each do |email, policy|
             policy = nil if policy == ""
-            next if policy && !allowed_add_tags[policy]
             next unless (inv = @project.invitations_dataset.where(email: email).first)
             old_policy = inv.policy
-            next if old_policy && !allowed_remove_tags[old_policy]
+            next if policy == old_policy
+            if policy && !allowed_add_tags[policy]
+              issues << "You don't have permission to add invitation to '#{policy}' tag"
+              next
+            end
+            if old_policy && !allowed_remove_tags[old_policy]
+              issues << "You don't have permission to remove invitation from '#{old_policy}' tag"
+              next
+            end
             inv.update(policy: policy)
+            (additions[inv.policy] ||= []) << 1 if inv.policy
+            (removals[old_policy] ||= []) << 1 if old_policy
           end
+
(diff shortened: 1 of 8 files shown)


routes/project/user.rb
```diff
@@ -171,14 +171,36 @@ class Clover
       r.post "policy/managed" do
         authorize("Project:user", @project.id)
 
-        invitation_policies = r.params["invitation_policies"] || {}
-        allowed_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").select_hash(:name, Sequel.as(true, :v))
-        invitation_policies.each do |email, policy|
-          if allowed_tags[policy]
-            @project.invitations.find { _1.email == email }&.update(policy: policy)
+        DB.transaction do
```
enescakir, line 174: Not sure if we can simplify this code

routes/project/user.rb
```diff
@@ -171,14 +171,36 @@ class Clover
...
+          allowed_add_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").all.to_h { [_1.name, _1] }
+          allowed_remove_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:remove").all.to_h { [_1.name, _1] }
+
+          user_policies = r.params["user_policies"] || {}
+          user_policies.each do |ubid, policy|
+            user = Account.from_ubid(ubid)
+            existing_tags = user.subject_tags_dataset.where(project_id: @project.id).select_map(:name)
```
jeremyevans, line 181: We need to take no action on the user here if the user is in more than one tag.

routes/project/user.rb
```diff
@@ -171,14 +171,36 @@ class Clover
...
-        end
 
+          invitation_policies = r.params["invitation_policies"] || {}
+          invitation_policies.each do |email, policy|
+            if allowed_add_tags[policy]
+              @project.invitations.find { _1.email == email }&.update(policy: policy)
+            end
+          end
+        end
         flash["notice"] = "Subject tags for invited users updated successfully."
```
jeremyevans, line 204: This notice may not be accurate, for any case that calls `next` in the loop above.  The code I wrote for tag membership tries to be specific about how many requested changes were applied successfully, and the reasons for any change not made.

jeremyevans (commented):
This seems like a good idea, as it keeps things simpler for users who only have basic access control needs.

routes/project/user.rb
enescakir: I added this check

routes/project/user.rb
enescakir: I added detailed notification in the latest commit

enescakir:
@jeremyevans I added 3 more commits for improvements

- **Show all subject tags in the policy selector**
  If the current user doesn't have permission for a subject tag, it won't
  appear in the policy selector.

  This is an issue because the selector doesn't provide an option for that
  tag, which is interpreted as no policy option. We should display all
  subject tags in the policy selector, but if the user lacks permission to
  add one of them, it should be disabled.

- **Add space between flash messages when multiple types are shown**

| Before | After |
|--------|-------|
| CleanShot 2024-12-25 at 16 20 30@2x | CleanShot 2024-12-25 at 16 23 38@2x |

- **Fix invitation policy update**
   While testing, I found two issues:
   -  Even if the current user doesn't have the "SubjectTag:remove"
   permission, they can still remove the subject tag from the
   invitation.
   -  The `allowed_tags[policy]` always returns nil for `No Policy`, so we
   can't update the invitation policy to `No Policy`.

routes/project/user.rb
```diff
@@ -31,6 +31,11 @@ class Clover
           end
 
           tag = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").first(name: policy)
+          unless tag || policy
```
jeremyevans, line 34: I'm not sure this condition is correct.  I think the correct condition is to show an error if a policy is selected, but the user does not have access to the subject tag:

```suggestion
          if policy && !tag
```

routes/project/user.rb
```diff
@@ -176,15 +181,67 @@ class Clover
...
+          removals = Hash.new(0)
+          additions = Hash.new(0)
+          changes = []
+          issues = []
+          user_policies = r.params["user_policies"] || {}
+          user_policies.each do |ubid, policy|
+            policy = nil if policy == ""
+            user = Account.from_ubid(ubid)
+            existing_tags = user.subject_tags_dataset.where(project_id: @project.id).select_map(:name)
+            next unless existing_tags.length < 2
```
jeremyevans, line 196: This case should add to `issues`, with a message like: "Cannot change policy for user, as they are in multiple subject tags"

routes/project/user.rb
```diff
@@ -176,15 +181,67 @@ class Clover
...
+            next if existing_tags.include?(policy)
+            existing_tags.each do |tag|
+              unless (removed_tag = allowed_remove_tags[tag])
+                issues << "You don't have permission to remove members from '#{tag}' tag"
+                next
+              end
+              removed_tag.remove_members(user.id)
```
jeremyevans, line 203: This approach uses a query per user.  A more optimized approach would be to collect all users to add and remove, and issue one add and/or delete query per tag to be modified, for all users that would be affected.  The tag code will insert/delete multiple rows in a single query if you call add_members/remove_members with an array of multiple arguments.

Additionally, I don't think we should remove the user from the tag unless we can add them to the new tag.

I recommend an approach where you only record the adds or removes that should be made during the loop, and then apply all changes after the loop.  You can make the `removals` and `additions` values arrays, append them with the member ids during the loop, and after the loop, loop over `removals` and call `remove_members` once for non-empty values, and them loop over `additions` and call `add_members` once for non-empty values.  You can get the counts by calling `size` on the arrays.

routes/project/user.rb
```diff
@@ -176,15 +181,67 @@ class Clover
...
         authorize("Project:user", @project.id)
 
-        invitation_policies = r.params["invitation_policies"] || {}
-        allowed_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").select_hash(:name, Sequel.as(true, :v))
-        invitation_policies.each do |email, policy|
-          if allowed_tags[policy]
-            @project.invitations.find { _1.email == email }&.update(policy: policy)
+        DB.transaction do
+          allowed_add_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").all.to_h { [_1.name, _1] }
+          allowed_remove_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:remove").all.to_h { [_1.name, _1] }
```
jeremyevans, line 186:
```suggestion
          allowed_add_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:add").to_hash(:name)
          allowed_remove_tags = dataset_authorize(@project.subject_tags_dataset, "SubjectTag:remove").to_hash(:name)
```

`Dataset#to_hash` is faster as it will avoid a bunch of array allocations.

jeremyevans (changes requested):
This looks pretty good.  I flagged some potential issues, but after those are addressed, I think this can be made non-draft and merged into the authorization refactor PR.

routes/project/user.rb
enescakir: When the user select `No policy`, the `policy` is `""`, so it doesn't falsy. So `tag` is `nil` for `No policy` too.

When the user select a disabled policy from selector, the `policy` is `nil`.

I tried to detect this condition in the old check. But after your comment, I found another issue with that.

So I think the correct version is `if !tag && policy != ""` or `unless tag || policy == ""`

routes/project/user.rb
enescakir: I didn't know that. Thank you

routes/project/user.rb
enescakir: I've refactored it an now writing tests to increase coverage

enescakir: The coverage is now 100% 🚀

jeremyevans:
I merged this into the `jeremy-authorization-refactor` branch, and added some followup commits to fix issues:

* bc0e7e31e6fcd9b700b3a013ed24fa1f8d2efe5e
* cb794eae94e7dccd02376ef0d62608cc196a589d
* 21f1964f675995271e93ec0e6ae9e60429e1282c
* 9a884cf5a6c03ccefc476d866622680e67d68a90

PR #2478, opened 2024-12-24
Use a template database for much faster migrating of test databases used for parallel testing

jeremyevans:
Previously, each parallel database was migrated separately from scratch.  This switches to only migrating the clover_test database, and using the createdb -T clover_test option to clone that to create the parallel test databases.  This is much faster, and doesn't clutter the screen with tons of unnecessary output, for easier debugging when testing migrations.

With this change, the migrate proc no longer needs a db keyword, so remove that and use DB instead.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,8 +13,22 @@ auto_parallel_tests = lambda do
   use_auto_parallel_tests
 end
 
+nproc = lambda do
+  # Limit to 6 processes, as higher number results in more time
+  `(nproc 2> /dev/null) || sysctl -n hw.logicalcpu`.to_i.clamp(1, 6).to_s
+end
+
+clone_test_database = lambda do
+  Sequel::DATABASES.each(&:disconnect)
+  nproc.call.to_i.times do |i|
+    database_name = "clover_test#{i + 1}"
+    sh "dropdb --if-exists -U postgres #{database_name}"
+    sh "createdb -U postgres -O clover -T clover_test #{database_name}"
+  end
+end
+
 # Migrate
-migrate = lambda do |env, version, db: nil|
+migrate = lambda do |env, version|
   ENV["RACK_ENV"] = env
   require "bundler"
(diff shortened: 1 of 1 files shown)


fdr (approved):
Neat. In previous projects we also collapsed migrations into a schema.sql using pg_dump with some options to, this much faster than processing intermediate stages. A bit of a similar methodology to the sequel schema caches.

Rakefile
```diff
@@ -13,42 +13,55 @@ auto_parallel_tests = lambda do
   use_auto_parallel_tests
 end
 
+nproc = lambda do
+  # Limit to 6 processes, as higher number results in more time
+  `(nproc 2> /dev/null) || sysctl -n hw.logicalcpu`.to_i.clamp(1, 6).to_s
```
fdr, line 18: You sure you want to make weird output yield a zero via to_i vs Integer()? It'll kinda work but a watchful eye would be required to see that one is subjected to a default/error default behavior. Were it me, I'd want it to crash, and then fix it.

jeremyevans: I'm fairly sure the command only works on Mac and Linux.  I don't think we should crash on other operating systems.  It's safe to default to a single CPU in that case, even if it is slower.  Maybe we can accept the value via an environment variable (e.g. `NCPU`) to allow parallel testing on other operating systems (or to allow a non-default concurrency limit on Mac or Linux).  What are your thoughts on that?  That should be done in a separate pull request.

fdr:
Perhaps print a warning? Alternatively, add the BSD one, and let it crash
otherwise. There are not many platforms. I guess I’d rather fix it on the
spot rather than limp along, even with warning that is easy to miss.

Rakefile
jeremyevans: I submitted pull request #2490 to switch the implementation to use `Etc.nprocessors` (from the `etc` standard library).

PR #2485, opened 2024-12-27
Allow to delete password if another login method is available

enescakir:
- **Update Rodauth to the latest commit**
  The new version has `has_password?` as public.
  

- **Allow to disconnect the last omniauth provider if has password**
  

- **Allow to delete password if another login method is available**
  If the user has another login method, they can delete their password.
  This is helpful for users who prefer to use only one login method.
  
  If the user doesn't have a password, they can create a new one using the
  "Change Password" feature. I changed "Change" to "Create" for clarity
  when the user doesn’t have a password.
  

- **Do not allow to reset password if password is disabled**


routes/account.rb
```diff
@@ -17,13 +17,15 @@ class Clover
...
-            # YYYY: We can allow to disconnect the last omniauth provider if the user has a password
-            # But need https://github.com/jeremyevans/rodauth/pull/461 to be merged first.
+          unless identities.length > (rodauth.has_password? ? 0 : 1)
             flash[:error] = "You must have at least one login method"
             r.redirect "/account/login-method"
           end
-          if (identity = identities.find { _1.provider == provider && _1.uid == uid })
+          if provider == "password"
+            DB[:account_password_hashes].where(id: current_account.id).delete(force: true)
+            DB[:account_previous_password_hashes].where(account_id: current_account.id).delete(force: true)
```
jeremyevans, line 26:
```suggestion
            DB[:account_password_hashes].where(id: current_account.id).delete
            DB[:account_previous_password_hashes].where(account_id: current_account.id).delete
```

`(force: true)` is only required for model datasets, not for plain datasets, since 9d81cc2a62fd3f7715dafc82281ed406c51628f9

views/account/change_password.erb
```diff
@@ -1,4 +1,6 @@
-<% @page_title = "Change Password" %>
+<% @page_title = "Change Password"
+  action = rodauth.has_password? ? "Change" : "Create"
```
jeremyevans, line 2:
```suggestion
<% action = rodauth.has_password? ? "Change" : "Create"
  @page_title = "#{action} Password"
```

routes/account.rb
```diff
@@ -17,13 +17,15 @@ class Clover
         r.post "disconnect" do
           provider, uid = typecast_params.nonempty_str(["provider", "uid"])
           identities = current_account.identities
-          unless identities.length > 1
-            # YYYY: We can allow to disconnect the last omniauth provider if the user has a password
-            # But need https://github.com/jeremyevans/rodauth/pull/461 to be merged first.
+          unless identities.length > (rodauth.has_password? ? 0 : 1)
```
jeremyevans, line 20:
```suggestion
          unless identities.length > (rodauth.has_password? && provider != "password") ? 0 : 1)
```

We should not allow deletion of password unless the user has at least one social login.

jeremyevans (approved):
This looks good.  I have a few suggestions, then it's fine to merge after.

routes/account.rb
enescakir: If the user has a password, `unless identities.length > 0`, so it doesn't allow to delete password if the user doesn't have any account identities.

routes/account.rb
enescakir: Cool. I didn't know that.

views/account/change_password.erb
enescakir: Good catch

routes/account.rb
jeremyevans: Understood.  If you don't have a password, you need at least 2 existing identities, because the assumption is the provider is an an identity provider, because you should not be deleting a password if you do not have a password.

spec/routes/web/auth_spec.rb
```diff
@@ -461,7 +475,21 @@ def mock_provider(provider, email = TEST_USER_EMAIL)
         expect(page).to have_content "Your account has been disconnected from Github"
       end
 
-      it "can not disconnect the last login method" do
+      it "can delete password if another login method is available" do
+        account.add_identity(provider: "google", uid: "123456790")
+        account.add_identity(provider: "github", uid: "123456790")
```
jeremyevans, line 480: May want to only have a single provider for this test, since you should be able to delete with a single provider.

jeremyevans (approved):

PR #2486, opened 2024-12-27
More preparation for new authorization system

jeremyevans:
This adds the tag member validation code and some other commits that do not conflict with the existing authorization system. We need to commit the tag member validation code early so that we can use it for the access policy to access control entry converter testing.

I've rebased #2384 on this, so these are now the first 5 commits in that PR.

### DIFF
--- a/lib/access_control_model_tag.rb
+++ b/lib/access_control_model_tag.rb
@@ -20,4 +20,86 @@ module AccessControlModelTag
   def add_member(member_id)
     applied_dataset.insert(:tag_id => id, applied_column => member_id)
   end
+
+  def member_ids
+    applied_dataset.where(tag_id: id).select_map(applied_column)
+  end
+
+  def add_members(member_ids)
+    applied_dataset.import([:tag_id, applied_column], member_ids.map { [id, _1] })
+  end
+
+  def remove_members(member_ids)
+    applied_dataset.where(:tag_id => id, applied_column => member_ids).delete
+  end
+
+  def currently_included_in
+    DB[:tag]
+      .with_recursive(:tag,
+        DB[applied_table].where(applied_column => id).select(:tag_id, 0),
+        DB[applied_table].join(:tag, tag_id: applied_column)
+          .select(Sequel[applied_table][:tag_id], Sequel[:level] + 1)
+          .where { level < Config.recursive_tag_limit },
+        args: [:tag_id, :level])
+      .select_map(:tag_id)
+  end
+
+  def check_members_to_add(to_add)
+    issues = []
+
+    current_members = member_ids
+
+    # Do not allow tag to include itself (prevent simple recursion)
+    if to_add.include?(id)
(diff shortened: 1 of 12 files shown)


lib/access_control_model_tag.rb
```diff
@@ -20,4 +20,86 @@ def applied_dataset
...
+  def member_ids
+    applied_dataset.where(tag_id: id).select_map(applied_column)
+  end
+
+  def add_members(member_ids)
+    applied_dataset.import([:tag_id, applied_column], member_ids.map { [id, _1] })
+  end
+
+  def remove_members(member_ids)
+    applied_dataset.where(:tag_id => id, applied_column => member_ids).delete
```
enescakir, line 33: Should we use `destroy` instead of `delete`? Otherwise, the data won't be archived to the ArchivedRecord table, which we use for auditing operations.

lib/access_control_model_tag.rb
```diff
@@ -20,4 +20,86 @@ def applied_dataset
...
+      issues << "#{proposed_additions.size - to_add.size} members not valid"
+    end
+
+    to_add.map!(&:id)
+    to_add.uniq!
+    [to_add, issues]
+  end
+
+  def before_destroy
+    applied_dataset.where(tag_id: id).or(applied_column => id).delete
```
enescakir, line 96: https://github.com/ubicloud/ubicloud/pull/2486/commits/359b564039e610b5ea6ccc4a8d2df1363b35b0d2#r1898699813

lib/access_control_model_tag.rb
jeremyevans: You cannot currently use `destroy`, because there is no model for the `applied_*_tag` tables.

If we want to copy deleted records for the `applied_*_tag` tables to the archive table, my recommendation would be to use database triggers and not model hooks.  That is faster and more robust.  `Dataset#destroy` is quite inefficient, because it retrieves all rows, and deletes each row in a separate query, running hooks separately for each row.  Please let me know if you would like me to work on a migration for that.

enescakir: Database triggers seem like a good idea to me. We may also get feedback from other team members too. What's your opinion on this? @byucesoy @fdr

jeremyevans: I pushed a commit with a migration that uses database triggers on `applied_{subject,action,object}_tag` to insert into archived_records, as well as a spec that shows that it works.

enescakir (approved):
LGTM

fdr:
Fine with them judiciously wielded. This crowd is well attuned for figuring
that out and adjusting.
I’ll review the specifics later,  but wanted to express my openness on this
sometimes hot topic.

fdr:
*checking the trigger*

Sure, seems fine, it's not that distinct from what, say, pg_repack (among others) does, and people tend to run pg_repack on somewhat higher volume databases.

PR #2487, opened 2024-12-27
Authorization tags UI improvements

enescakir:
# Disclaimer: Each commit has its own improvement. I don't feel very confident about all of them. They might be subjective. We can discuss and possibly discard some of them.

- **Do not render checkbox label if it is empty**
  

- **Relocate the tag creation form to the top of the list**
  Having lots of pages with limited functionality can complicate the user
  experience. We could merge some of these pages. Placing single input
  forms at the top of the listing page is a common web pattern. It is more
  user-friendly as it allows users to create a new tag without the need to
  navigate between pages.
  
  I've also repositioned the 'Membership' button from the left to the
  right, placing it next to the 'Remove' button. I believe grouping all
  action buttons in one column enhances usability. I've renamed it to
  'Manage' and will be improving this management page in upcoming commits.

| Before | After |
|--------|-------|
| Before List | After List |
| Before Create  | |

- **Wrap tables and titles in a div on the membership page**
  The container has the `grid gap-6` class, which adds gaps between the
  children. When the table and its title are not wrapped in a div, there
  is too much space between them. Wrapping them in a div will fix this
  issue.
  
| Before | After |
|--------|-------|
| Before | After |

- **Improve UI in tag membership page**
  I've made a few improvements:
  - We repeat the tag name three times in the page. It's enough to show it
    once in the page header.
  - I removed the extra space to the right of the buttons
  - I added a background color to the add group name rows

| Before | After |
|--------|-------|
| Before | After |  

- **Make the entire row clickable in the membership list**
  This won't alter the existing functionality. It simply expands the
  clickable area using JavaScript, so it won't impact any tests.
  

- **Consolidate tag detail operations into one page**
  I previously moved the create tag form to the top of the tag listing.
  
  We can also move the update tag form to the top of the detail page. In
  other words, I moved membership operations to the details page.
  
  I didn't realize the tag name is clickable. I might change its color,
  but the "Manage" button already has this link now.
  
CleanShot 2024-12-25 at 23 15 17@2x 2


jeremyevans:
I haven't reviewed the code, but from a review of the images, I think all of these UI improvements are good, except I would not recommend merging the current/available members tables.  If you have a lot of available members, but only a few current members (a very common scenario), it makes it much more difficult to see what members are currently in the tag.  I think the separation of current members and members available to add is friendlier to the user.

enescakir:
> except I would not recommend merging the current/available members tables. If you have a lot of available members, but only a few current members (a very common scenario), it makes it much more difficult to see what members are currently in the tag. I think the separation of current members and members available to add is friendlier to the user.

We may place selected items at the top of the list as a separate group.

CleanShot 2024-12-30 at 20 49 18@2x

But I don't feel too strong about it. I've excluded it from the current PR for now. We can revisit this topic if we choose to merge these list in a different PR.

jeremyevans:
Looks great, thanks!

PR #2498, opened 2024-12-31
Improve UI for token restriction

enescakir:
This improvement has similar characteristics to the previous one. I tried to reduce the number of pages.

I also moved the restrict and unrestrict actions to a separate card, so users can get more focused information about token restriction.

Additionally, I believe we don't need to show information about our token access policy feature if the token already has entries. If it has entries, it means the user is already aware of the feature. They can read more about it in the documentation.

| Before | After |
|--------|-------|
| before-token-list | after-token-list |
| before-restrict-token | after-restrict-token |
| before-no-policy | after-no-policy |
| before-policy | after-policy |


jeremyevans (approved):
Looks good.  I have a some small changes, but I'll make them after merging.

PR #2501, opened 2024-12-31
Fix the active tab on the user tab bar for nested pages


enescakir:
- **Allow to pass active property to tabbar item**
  

- **Fix the active tab on the user tab bar for nested pages**
  I just added the third item, but erb_formatter made it multi lines

### DIFF
--- a/views/components/tabbar.erb
+++ b/views/components/tabbar.erb
@@ -1,15 +1,13 @@
 <div class="sm:block mb-6">
   <nav class="isolate flex divide-x divide-gray-200 rounded-lg shadow" aria-label="Tabs">
-    <% tabs.each_with_index do |(label, url), i| %>
+    <% tabs.each_with_index do |(label, url, active), i| %>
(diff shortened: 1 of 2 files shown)


views/components/tabbar.erb
```diff
@@ -1,15 +1,13 @@
 <div class="sm:block mb-6">
   <nav class="isolate flex divide-x divide-gray-200 rounded-lg shadow" aria-label="Tabs">
-    <% tabs.each_with_index do |(label, url), i| %>
+    <% tabs.each_with_index do |(label, url, active), i| %>
+      <% active ||= (request.path == url) %>
```
jeremyevans, line 4: Seems like you would only want to do this if `active` is `nil`.  If you are explicitly passing `false`, it seems unnecessary to do the path check.

```suggestion
      <% active = (request.path == url) if active.nil? %>
```

jeremyevans (approved):
Looks good, I only recommend one small change, which I'll make after merging.

PR #2503, opened 2024-12-31
Fix HTML validation issues

jeremyevans:
This uses capybara-validate_html5 to validate HTML when parsing response bodies in the web route tests.  This found numerous HTML validation issues:

1. `<form>` directly inside `<tr>` (many cases)
2. `<p>` nested inside `<p>` (nesting not allowed, so second `<p>` implicitly closes first `<p>`, and second `</p>` is therefore invalid) (2 cases)
3. Use of `</br>` instead of `<br>` (1 case)
4. Missing `</div>` (1 case)
5. Plain text responses with text/html content-type (2 cases)

To fix the first case, the forms in question are moved inside `<td>`. If thtere are inputs in other `<td>` in the same `<tr>`, the `<form>` has an id attribute, an the other inputs use a form attribute to tie them to the correct `<form>`.


views/components/form/select.erb
```diff
@@ -58,6 +58,10 @@ attributes ||= {} %>
     <p class="text-sm text-red-600 leading-6" id="<%= name %>-error"><%= error %></p>
   <% end %>
   <% if description %>
-    <p class="text-sm text-gray-500 leading-6" id="<%= name %>-description"><%== description %></p>
+    <% if description.start_with?("<p") %>
```
enescakir, line 61: Do we have any select descriptions starting with `<p>`?

views/networking/firewall/show.erb
```diff
@@ -60,28 +60,44 @@
...
+            </td>
+            <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6" scope="row">
+              <%== render(
+                "components/form/text",
+                locals: {
+                  name: "port_range",
+                  type: "text",
+                  attributes: {
+                    placeholder: "0..65536",
+                    form: "form-fw-create-rule-#{@firewall[:id]}"
```
enescakir, line 85: I wasn't aware such a HTML feature. Cool 😎

views/components/form/select.erb
jeremyevans: Yes, in `views/vm/create.erb`

PR #2504, opened 2024-12-31
Hopefully final preparation for authorization system switchover

jeremyevans:
This commits the access policy converter and a few other small changes that the converter depends on to work.  This brings us to the commit directly before the switchover to using the new access control system.

One possible rollout approach is to commit this first, and just before pushing the authorization system changes (#2384) into production, run the access policy converter.


enescakir (approved):
I'll take another close look at the converter

PR #2510, opened 2025-01-02
Start charging for inference endpoints

bsatzger:
Start charging for inference endpoints serving these models:
- llama-3-2-3b-it: $0.1 per million tokens
- llama-3-3-70b-it: $0.9 per million tokens
- qwq-32b-preview: $0.9 per million tokens


config/billing_rates.yml
```diff
@@ -1,13 +1,13 @@
 # Active billing rates
 - { id: fc9877ec-131c-4572-a3f2-fd512d95b348, resource_type: InferenceTokens,         resource_family: test-model,          location: global,         unit_price: 0.0000000500, billed_by: amount,   active_from: 2024-10-14T00:00:00Z }
 - { id: 74eb807c-75dd-49e4-95f8-5de25163b845, resource_type: InferenceTokens,         resource_family: gemma-2-2b-it,       location: global,         unit_price: 0.0000000000, billed_by: amount,   active_from: 2024-10-14T00:00:00Z }
-- { id: 195bd918-8a0b-4c07-8b57-289de6d6163e, resource_type: InferenceTokens,         resource_family: llama-3-2-3b-it,     location: global,         unit_price: 0.0000000000, billed_by: amount,   active_from: 2024-10-14T00:00:00Z }
+- { id: 195bd918-8a0b-4c07-8b57-289de6d6163e, resource_type: InferenceTokens,         resource_family: llama-3-2-3b-it,     location: global,         unit_price: 0.0000001000, billed_by: amount,   active_from: 2024-10-14T00:00:00Z }
```
enescakir, line 4: Usually, when we adjust prices, we create a new billing rate and move the old ones to the deprecated section. However, since we currently have no paying customers and aren't charging anyone, I think it's fine to replace the rates inplace.

enescakir (approved):

config/billing_rates.yml
bsatzger: The implementation is such that for a unit price of zero, no billing records are produced. So I too think changing inference token cost in-place is fine in this case.

PR #2512, opened 2025-01-02
Reduce amount of respirate logs

bsatzger:
**Stop emitting "starting strand" log lines**
We produce a high volume of "starting strand" log lines, which are not useful enough to justify the downstream cost of processing and storing them.

**Emit "finished strand" logs only if duration > 1sec**
We produce a high volume of "finished strand" log lines. To reduce the volume, we emit the log line only if the duration is noteworthy, i.e. greater than 1s.

**Reduce amount of obtained/cleared lease log lines**
The system currently generates a large number of "obtained lease" and "lease cleared" log entries, which contributes to increased downstream costs. Since some views rely on these log entries, this update does not eliminate them entirely. Instead, it reduces their frequency by emitting them at a rate of approximately one per thousand.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -157,7 +157,8 @@ SQL
       fail "BUG: Prog #{prog}##{label} did not provide flow control"
     end
   ensure
-    Clog.emit("finished strand") { [self, {strand_finished: {duration: Time.now - start_time, prog_label: prog_label}}] }
+    duration = Time.now - start_time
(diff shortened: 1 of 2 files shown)


fdr (approved):
I'm a little puzzled what a stochastic process does for us here, but I would like to reduce our log levels quickly… relying on instincts of people that have used them carefully (which you did, for performance work)

If we want to be more radical and delete, it can come later.

enescakir (approved):

PR #2517, opened 2025-01-02
Enhance playground response formatting

bsatzger:
During the streaming of the model’s response, no formatting is applied; instead, the chunk contents are appended directly to
`inference_response_stream`. Once the entire response is received, the marked library is used to generate well-formatted HTML, which is then written to `inference_response_pretty`.

## Example 1:
While streaming the response (same behavior as before)
streaming_code

After the request has finished (new behavior)
pretty_code

## Example 2:

While streaming the response
streaming_table

After the request has finished
pretty_table


enescakir (commented):
Since we already import `dompurify` js library, should we use it?

CleanShot 2025-01-03 at 14 54 50@2x

enescakir (approved):

PR #2518, opened 2025-01-02
Remove bloated lease Clog messages

fdr:
These are too verbose relative to their value, imposing about 25% of log traffic by bytes.  But, they have not been useful in quite a while, at least, ever since the strand leasing/mutexing system has matured and not been changed much in a while.  They were useful assessing the crucial bug in bbf3e142bda5b408cd94150a435efca6bd40204c, though.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -29,7 +29,6 @@ SQL
     return false unless affected
     lease_time = affected.fetch(:lease)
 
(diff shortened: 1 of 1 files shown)


bsatzger:
Earlier today, I opened https://github.com/ubicloud/ubicloud/pull/2512 with similar intentions.

fdr:
Obsolete

PR #2519, opened 2025-01-02
Don't check specs not installed in E2E CI when the host is already setup.

pykello:
When using an already setup server, the `Prog::Test::HetznerServer` skips the `HostNexus.assemble`, so it doesn't make sense to check `HostNexus.assemble` didn't install specs.

This also makes life easier for developers who try running `bin/ci` with the `--vm-host-id` arg.

### DIFF
--- a/prog/test/hetzner_server.rb
+++ b/prog/test/hetzner_server.rb
@@ -10,11 +10,11 @@ class Prog::Test::HetznerServer < Prog::Test::Base
       vm_host = VmHost[vm_host_id]
       {
         vm_host_id: vm_host.id, server_id: vm_host.hetzner_host.server_identifier,
-        hostname: vm_host.sshable.host, destroy: false
+        hostname: vm_host.sshable.host, setup_host: false
       }
     else
       {
-        server_id: Config.ci_hetzner_sacrificial_server_id, destroy: true
+        server_id: Config.ci_hetzner_sacrificial_server_id, setup_host: true
       }
     end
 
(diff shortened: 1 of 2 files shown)


fdr:
What are the pros/cons of tying this to the `Config.development?` flag rather than vm-host-id being specified? It's a bit slick to tie the change in check to how explicit one is about the vm host, but maybe it works in practice.

prog/test/hetzner_server.rb
```diff
@@ -94,8 +94,8 @@ def self.assemble(vm_host_id: nil)
       hop_run_integration_specs
     end
 
-    # We shouldn't install specs by default
-    verify_specs_installation(installed: false)
+    # We shouldn't install specs by default when running Prog::Vm::HostNexus.assemble
+    verify_specs_installation(installed: false) unless frame["vm_host_id"]
```
macieksarnowicz, line 98: I think at this point vm_host_id is always present. Look at the line 78. I think the only parameter that you can check here is 'destroy'. It is set to 'false' when own host is used.

pykello: Thanks. I replaced `destroy` with `setup_host` & used it.

pykello:
> What are the pros/cons of tying this to the `Config.development?` flag rather than vm-host-id being specified? It's a bit slick to tie the change in check to how explicit one is about the vm host, but maybe it works in practice.

What's the relation for being `Config.development?` or not?

What we are verifying here is that the `HostNexus.assemble` didn't install the specs, so we do the check if `HostNexus.assemble` was run, otherwise we don't. It seems to be more logically related.

jeremyevans (commented):
I don't see any issues from a Ruby perspective.

enescakir (approved):

PR #2523, opened 2025-01-03
Update postgres image versions

byucesoy:


enescakir:
I think you need to rebase it @byucesoy

enescakir (approved):

PR #2531, opened 2025-01-06
Fix Load Balancer port rewriting

furkansahin:
## Do not overwrite the source port when load balancing
With the PR https://github.com/ubicloud/ubicloud/pull/2272, we have
introduced source port overwriting when load balancing to a neighbor
node. This has created confusion for connection tracking and simply
broke load balancer. In this commit, I am changing the logic back to not
overwriting, but introducing selective destination port overwriting.

This way, if the connection is forwarded to a neighbor node, we are
still using the load balancer public port and keep the source port same.
To be able to process that on the neighbor side, we are overwriting the
destination port from public to private upon arrival. For packs that are
coming from a neighbor node's private IP to the arrival private IP but
in public port.

This creates a problem for connections to be forwarded to the local VM.
The reason is, DNATing can only be done at the preforward hook which is
executed before applying the firewall rules. Therefore, we cannot leave
the public port, we must overwrite it to private port since the
application is running in the private port. Once we overwrite that, we
have to make sure firewall rules for the public port are still applied
to that connection. We will solve that in the next commit.

## Mark lb packets to be forwarded to local and apply FW rules
Since we are overwriting the destination port to the private port in the
preforward hook, we have to make sure the firewall rules for the public
port are applied. For that, when we are creating the firewall rule sets,
we generate a new copy of the sets and replace the public port with the
private port. It is not enough to do that because we only need to apply
the rules to the packets coming through the load balancer. For that, we
introduce marking in the load balancer and catch the marked packets in
the forward hook.

Since now load balancer and firewall need to work together, we have to
make sure firewall rules are updated when a node is removed or added. We
catch those in the dns update or destroy sequence.

### DIFF
--- a/prog/vnet/update_firewall_rules.rb
+++ b/prog/vnet/update_firewall_rules.rb
@@ -11,8 +11,8 @@ class Prog::Vnet::UpdateFirewallRules < Prog::Base
 
   label def update_firewall_rules
     rules = vm.firewalls.map(&:firewall_rules).flatten
-    allowed_ingress_ip4_port_set = consolidate_rules(rules.select { !_1.ip6? && _1.port_range })
-    allowed_ingress_ip6_port_set = consolidate_rules(rules.select { _1.ip6? && _1.port_range })
+    allowed_ingress_ip4_port_set, allowed_ingress_ip4_lb_dest_set = consolidate_rules(rules.select { !_1.ip6? && _1.port_range })
+    allowed_ingress_ip6_port_set, allowed_ingress_ip6_lb_dest_set = consolidate_rules(rules.select { _1.ip6? && _1.port_range })
     guest_ephemeral, clover_ephemeral = subdivide_network(vm.ephemeral_net6).map(&:to_s)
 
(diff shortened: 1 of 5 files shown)


prog/vnet/update_firewall_rules.rb
```diff
@@ -23,6 +23,12 @@ def before_run
       <<~LOAD_BALANCER_ALLOW_RULE
 #{allow_ipv4_lb_neigh_incoming}
 #{allow_ipv6_lb_neigh_incoming}
+
+# The traffic that is routed to the local VM from the load balancer
+# is marked with 0x00B1C100D. We need to allow this traffic to
+# the local VM.
+meta mark 0x00B1C100D ip saddr . tcp dport @allowed_ipv4_lb_dest_set ct state established,related,new counter accept
```
fdr, line 30: oobiclood shows up again! I was wondering if we were going to get into packet marking.

aamederen:
I tested this with my LB setup for k8s and it seems like the issue is fixed! Thanks.

aamederen (approved):

PR #2535, opened 2025-01-06
Improve billing info update

enescakir:
- **Show an error message when Stripe update fails**
  We don’t store any billing information or payment methods in our
  database. To update the billing method, we send a request to Stripe. I
  noticed a few failed requests that resulted in a 500 error. Instead of
  showing a 500 error, we can display the error message to the user.
  

- **Trim whitespace from billing email before sending**
  I noticed a few failed web requests caused by whitespace email address.
  
  Let's trim the email address before sending it to the API.
  

- **Add rspec helper for flash messages**
  We have many tests that check for flash messages. Currently, we search the whole page for the content or find it by a specific ID. To make it easier to test the exact content of flash messages, I added a spec helper.
  

Example failure message:

     Failure/Error: expect(page).to have_flash_notice("An email has been sent to you with a link to verify your account2")

             expected: notice - An email has been sent to you with a link to verify your account2
         actual error: (no error message)
        actual notice: An email has been sent to you with a link to verify your account

### DIFF
--- a/spec/routes/web/spec_helper.rb
+++ b/spec/routes/web/spec_helper.rb
@@ -39,6 +39,33 @@ RSpec.configure do |config|
     Capybara.reset_sessions!
     Capybara.use_default_driver
   end
+
+  def flash_message_matcher(expected_type, expected_message)
+    match do |page|
+      actual_message = page.find_by_id("flash-#{expected_type}").text
+      if expected_message.is_a?(String)
+        actual_message == expected_message
+      else
+        actual_message =~ expected_message
+      end
+    end
+
+    failure_message do |page|
+      <<~MESSAGE
+        #{"expected: ".rjust(16)}#{expected_type} - #{expected_message}
+        #{"actual error: ".rjust(16)}#{page.has_css?("#flash-error") ? page.find_by_id("flash-error").text : "(no error message)"}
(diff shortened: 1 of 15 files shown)


spec/routes/web/spec_helper.rb
```diff
@@ -39,6 +39,24 @@ def refresh
     Capybara.reset_sessions!
     Capybara.use_default_driver
   end
+
+  RSpec::Matchers.define :have_flash_message do |expected_type, expected_message|
```
jeremyevans, line 43: Can we have two matchers here, `have_flash_notice` and `have_flash_error`?  On failure, can we show the content of both the flash notice and error.  I think that will make debugging much easier.

jeremyevans (commented):
Billing info change looks good. I'm in favor of the flash notice/error support in the specs, but I have a couple of recommendations.  Please let me know what you think.

spec/routes/web/spec_helper.rb
enescakir: I refactored it. Also update failure message

Example failure message:

     Failure/Error: expect(page).to have_flash_notice("An email has been sent to you with a link to verify your account2")

             expected: notice - An email has been sent to you with a link to verify your account2
         actual error: (no error message)
        actual notice: An email has been sent to you with a link to verify your account

jeremyevans (approved):

PR #2539, opened 2025-01-07
Use fixed locals to simplify templates

jeremyevans:
By default, templates in Roda/Tilt do not use fixed locals. Templates support arbitrary locals passed to them, and a separate template method is compiled for each combination of locals.  This has numerous disadvantages if you are extensively using local variables in templates (as Clover does):

* It is inefficient, especially for large templates that are called with many combinations of locals.
* It hides issues if unused local variable names are passed to the template
* It does not support default values for local variables
* It does not support required local variables
* It does not support cases where you want to pass values via a keyword splat
* It does not support named blocks

With fixed locals, instead of the compiled template method accepting a hash positional argument, it accepts keyword arguments. Those keyword arguments can be required, if they are necessary for the template to work, or they can be optional and have default values.

This allows us to remove all of the:

```
local_var ||= nil
```

code in templates.

This set up clover to default to templates not accepting local variables.  If the template accepts local variables, you use a magic comment in the template to specify the local variables, such as this for 2 required arguments (from the partnership notice):

```
<%# locals: (description:, tos_text:) %>
```

Or this for all optional arguments, from the datepicker:

```
<%# locals: (name: nil, label: nil, default_date: nil, min_date: nil, max_date: nil) %>
```

This also specifies the :scope_class template option, which speeds up template lookup and reduces memory usage by avoiding an unnecessary template cache inside the Tilt::Template (with :scope_class and fixed locals, there is only a single compiled method per Tilt::Template).

This needs the current master branch of Tilt, Roda, and Rodauth to work. I expect new versions of all three libraries to be released next week.

Other minor changes in this commit:

* Remove unnecessary hash values
* Fix typo in views/components/empty_state that looks like it would result in links being created even if no link is specified.


fdr (approved):
pretty cool to see application streamlining show up in upstream projects with such fluency.

enescakir (approved):
Such a cool feature 🔥

PR #2541, opened 2025-01-07
Selectively run Load Balancer health probes for each stack

furkansahin:
We had introduced stack option at the backend but we forgot to update the health probes accordingly. This PR fixes that.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -282,6 +282,9 @@ paths:
                 algorithm:
                   description: Algorithm of the Load Balancer
                   type: string
+                stack:
+                  description: 'Networking stack of the Load Balancer (ipv4, ipv6, or dual)'
+                  type: string
                 dst_port:
                   description: Destination port for the Load Balancer
                   type: integer
@@ -714,6 +717,9 @@ paths:
                 algorithm:
                   description: Algorithm of the Load Balancer
                   type: string
+                stack:
+                  description: 'Networking stack of the Load Balancer (ipv4, ipv6, or dual)'
+                  type: string
                 dst_port:
                   description: Destination port for the Load Balancer
                   type: integer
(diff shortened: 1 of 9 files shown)


prog/vnet/load_balancer_health_probes.rb
```diff
@@ -8,18 +8,24 @@ def vm
   end
 
   label def health_probe
-    response_code = begin
+    address_ipv4 = load_balancer.ipv4_enabled? ? vm.nics.first.private_ipv4.network : nil
+    address_ipv6 = load_balancer.ipv6_enabled? ? vm.ephemeral_net6.nth(2).to_s : nil
```
bsatzger, line 12: I would either keep them both as String or both as NetAddr objects. Probably the latter:
```
    address_ipv4 = load_balancer.ipv4_enabled? ? vm.nics.first.private_ipv4.network : nil
    address_ipv6 = load_balancer.ipv6_enabled? ? vm.ephemeral_net6.nth(2) : nil
```

Then you can can later check if `address.version` is `4` or `6`, instead of checking whether the string includes a colon.

prog/vnet/load_balancer_health_probes.rb
```diff
@@ -8,18 +8,24 @@ def vm
...
+      rescue
+        "500"
       end
-
-      vm.vm_host.sshable.cmd(cmd)
-    rescue
-      "500"
     end
 
+    response_code = response_codes.include?("200") ? 200 : 400
```
bsatzger, line 27: nit: maybe remove this line and below replace `response_code.to_i == 200 with` `response_codes.include?("200")`.

openapi/openapi.yml
```diff
@@ -282,6 +282,9 @@ paths:
                 algorithm:
                   description: Algorithm of the Load Balancer
                   type: string
+                stack:
+                  description: Stack of the Load Balancer
```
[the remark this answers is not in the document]
bsatzger: On the UI that's less of a problem as it has some context, i.e. it shows ipv4, ipv6, dual.

bsatzger (approved):

PR #2544, opened 2025-01-07
Fix Subnet Nexus rekeying synchronisation issue

furkansahin:
A recent bug report showed that when multiple VMs are provisioned, sometimes, the private subnet gets stuck in wait_inbound_setup. The main reason is that, when we start rekeying nics, we include all of the nics in wait_setup label. The problem with that is, sometimes the VM is not allocated just yet. If it's not allocated, its public ipv6 address is not set. Therefore, for tunnels that target this nic, RekeyNicTunnel prog errors out in subdivide_network method. To solve this, we add a new state to NicNexus to wait VM allocation, we jump to the next label only after the allocation is completed.

### DIFF
--- a/prog/vnet/nic_nexus.rb
+++ b/prog/vnet/nic_nexus.rb
@@ -16,7 +16,7 @@ class Prog::Vnet::NicNexus < Prog::Base
 
     DB.transaction do
       nic = Nic.create(private_ipv6: ipv6_addr, private_ipv4: ipv4_addr, mac: gen_mac, name: name, private_subnet_id: private_subnet_id) { _1.id = ubid.to_uuid }
-      Strand.create(prog: "Vnet::NicNexus", label: "wait_setup") { _1.id = nic.id }
+      Strand.create(prog: "Vnet::NicNexus", label: "wait_allocation") { _1.id = nic.id }
     end
   end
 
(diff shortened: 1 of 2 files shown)


model/nic.rb
```diff
@@ -14,7 +14,7 @@ class Nic < Sequel::Model
   include SemaphoreMethods
 
   semaphore :destroy, :start_rekey, :trigger_outbound_update,
-    :old_state_drop_trigger, :setup_nic, :repopulate, :lock
+    :old_state_drop_trigger, :setup_nic, :repopulate, :lock, :allocated
```
enescakir, line 17: Do we really need a new semaphore? Can't we just use `vm.allocated_at`?

furkansahin: You're right, I updated it accordingly.

prog/vnet/nic_nexus.rb
```diff
@@ -26,6 +26,13 @@ def before_run
     end
   end
 
+  label def wait_allocation
+    if nic.vm.allocated_at
```
enescakir, line 30: I assume no nic would be in this state without a vm

enescakir (approved):

prog/vnet/nic_nexus.rb
furkansahin: yes

PR #2545, opened 2025-01-07
[DRAFT] Kubernetes Prototype

aamederen:
This is a Draft PR for viewing the overall thoughts on the managed kubernetes service design. Not intended for merge.

The whole package contains basic blocks for
1. Provisioning control plane and worker nodes, putting a LB for CP API servers
1. A homemade CNI for connectivity within a cluster
1. Handling upgrades (one-by-one, CP nodes first, by replacing)
1. Deprovisioning a cluster (not perfect, OK for dev)
1. UI for creating and viewing a cluster

image

image

image

Some questions about the implementation (comments are welcome)
1. Do we need a nexus for nodepools?
1. We do `push` a lot of progs (provision, upgrade, etc.). Would `bud` be preferable?
1. Shall we introduce a wrapper around VMs (aka `KubernetesNode` model)?

### DIFF
--- a/prog/kubernetes/kubernetes_nodepool_nexus.rb
+++ b/prog/kubernetes/kubernetes_nodepool_nexus.rb
@@ -80,7 +80,7 @@ class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
   end
 
   label def destroy
-    kubernetes_nodepool.vms.map(&:incr_destroy)
+    kubernetes_nodepool.vms.each(&:incr_destroy)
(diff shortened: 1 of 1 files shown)


config.rb
```diff
@@ -88,6 +88,8 @@ def self.test?
   override :minimum_invoice_charge_threshold, 0.5, float
   optional :cloudflare_turnstile_site_key, string
   optional :cloudflare_turnstile_secret_key, string
+  optional :management_k8s_url, string
+  optional :management_k8s_token, string
```
fdr, line 92: unused right? delete
```suggestion
```

aamederen: correct

PR #2548, opened 2025-01-07
Fix 500 for an invalid subnet for connect_subnet

furkansahin:
## Fix 500 for an invalid subnet for connect_subnet
In case the passed private subnet id doesn't exist, we crash with 500.
This is not a good user experience. While this should never be the case
if the private subnet is picked from the list, it can be done via
inspect elements. Here, I am handling this scenario, too.

## Fix Private Subnet disconnect
Using redirect doesn't really work since we did not submit a form. We
need to return.

### DIFF
--- a/routes/project/location/private_subnet.rb
+++ b/routes/project/location/private_subnet.rb
@@ -48,7 +48,7 @@ class Clover
           authorize("PrivateSubnet:disconnect", subnet.id)
           ps.disconnect_subnet(subnet)
           flash["notice"] = "#{subnet.name} will be disconnected in a few seconds"
-          ""
+          204
         end
       end
 
--- a/routes/project/location/private_subnet.rb
+++ b/routes/project/location/private_subnet.rb
@@ -26,6 +26,11 @@ class Clover
         r.post "connect" do
           authorize("PrivateSubnet:connect", ps.id)
           subnet = PrivateSubnet.from_ubid(r.params["connected-subnet-ubid"])
+          unless subnet
+            flash["error"] = "Subnet to be connected not found"
+            r.redirect "#{@project.path}#{ps.path}"
+          end
+
           authorize("PrivateSubnet:connect", subnet.id)
           ps.connect_subnet(subnet)
           flash["notice"] = "#{subnet.name} will be connected in a few seconds"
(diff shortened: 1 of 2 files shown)


spec/routes/web/private_subnet_spec.rb
```diff
@@ -76,7 +76,7 @@
         click_button "Create"
 
         expect(page.title).to eq("Ubicloud - #{name}")
-        expect(page).to have_flash_notice("'#{name}' will be ready in a few seconds")
+        expect(page).to have_content "'#{name}' will be ready in a few seconds"
```
jeremyevans, line 79: This seems unrelated.  Possible incorrect merge/rebase (same for other change in this file)?

spec/routes/web/private_subnet_spec.rb
```diff
@@ -186,6 +186,36 @@
...
+      it "cannot connect to a subnet when it does not exist" do
+        private_subnet
+        ps2 = Prog::Vnet::SubnetNexus.assemble(project.id, name: "dummy-ps-2", location: "hetzner-fsn1").subject
+        visit "#{project.path}#{private_subnet.path}"
+        ps2.strand.destroy
+        ps2.destroy
+        select "dummy-ps-2", from: "connected-subnet-ubid"
+        click_button "Connect"
+
+        expect(page).to have_content "Subnet to be connected not found"
```
jeremyevans, line 199:
```suggestion
        expect(page).to have_flash_error "Subnet to be connected not found"
```

jeremyevans (commented):
Looks good, other than a couple minor issues.

routes/project/location/private_subnet.rb
```diff
@@ -35,10 +40,15 @@ class Clover
...
+          unless subnet
+            response.status = 400
+            next {error: {message: "Subnet to be disconnected not found"}}
+          end
+
           authorize("PrivateSubnet:disconnect", subnet.id)
           ps.disconnect_subnet(subnet)
           flash["notice"] = "#{subnet.name} will be disconnected in a few seconds"
-          ""
+          r.redirect "#{@project.path}#{ps.path}"
```
enescakir: Since we're sending an AJAX request, the returned value isn't necessary. Therefore, we can return `204`.

enescakir: AJAX request just refreshes the page

furkansahin: Are the last changes OK?

enescakir (approved):

PR #2557, opened 2025-01-08
Fix default value for classes in the icon component

enescakir:
Recently, we started using fixed locals at 73149c4.

Interestingly, the current default value definition for the `classes` variable in `icon.erb` is broken.

If the page has multiple icons, the icons after the first one are not rendered correctly.

CleanShot 2025-01-08 at 18 32 47@2x

CleanShot 2025-01-08 at 18 33 24@2x

CleanShot 2025-01-08 at 18 35 08@2x

### DIFF
--- a/views/components/icon.erb
+++ b/views/components/icon.erb
@@ -1,6 +1,5 @@
-<%# locals: (name:, classes: "") %>
+<%# locals: (name:, classes: "w-6 h-6") %>
 
-<% classes ||= "w-6 h-6" %>
 <% case name %>
 <% when "hero-arrow-up-right" %>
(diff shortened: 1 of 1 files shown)


jeremyevans (approved):
Sorry about that.

enescakir:
> Sorry about that.

No problem. But I wonder what the root cause is. Why isn't the previous version working?

jeremyevans:
> No problem. But I wonder what the root cause is. Why isn't the previous version working?

Empty string is truthy, so `classes ||= "w-6 h-6"` has no effect.  If the default value were `nil` instead of the empty string, it would work.

PR #2558, opened 2025-01-08
Rename reset to reimage to avoid confusion with hetzner's own reset api.

0xjunhao:
Rename reset to factory_reset in hetzner api lib. This avoids confusion with the reset method in the hetzner REST api, which does hardware reset.

### DIFF
--- a/lib/hosting/apis.rb
+++ b/lib/hosting/apis.rb
@@ -10,9 +10,9 @@ class Hosting::Apis
     end
   end
 
-  def self.reset_server(vm_host)
+  def self.reimage_server(vm_host)
     if vm_host.provider == HetznerHost::PROVIDER_NAME
-      vm_host.hetzner_host.api.reset(vm_host.hetzner_host.server_identifier)
+      vm_host.hetzner_host.api.reimage(vm_host.hetzner_host.server_identifier)
     else
       raise "unknown provider #{vm_host.provider}"
     end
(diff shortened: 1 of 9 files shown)


0xjunhao:

bsatzger:
The `ubicloud` repo only allows `Rebase and merge`, i.e. the commits in your PR are not squashed but added as is onto `main` when merging. We try to avoid commits like `fix tests` that wouldn't give a lot of context when merged. Also, optimally, each individual commit should pass all tests.

lib/hosting/apis.rb
```diff
@@ -12,7 +12,7 @@ def self.pull_ips(vm_host)
 
   def self.reset_server(vm_host)
```
bsatzger, line 13: `reset_server` is somewhat ambiguous here as well, especially given Hetzner's terminology. I'm wondering if we should rename this method as well, something like `factory_reset_server`? Another option would be `reimage_server` for this method and `reimage` in the hosting apis, instead of `factory_reset`.

0xjunhao: Done. Changed to reimage.

prog/test/hetzner_server.rb
```diff
@@ -45,20 +45,20 @@ def self.assemble(vm_host_id: nil)
     hetzner_api.add_key("ubicloud_ci_key_#{strand.ubid}", keypair.public_key)
     update_stack({"hetzner_ssh_keypair" => Base64.encode64(keypair.keypair)})
 
-    hop_reset
+    hop_reimage
   end
 
-  label def reset
-    hetzner_api.reset(
+  label def reimage
```
bsatzger, line 51: usually we have to be careful when renaming labels to make sure that strands don't end up in labels that don't exist anymore. but since `reset` was a dev-only thing we should be fine.

bsatzger (approved):

prog/test/hetzner_server.rb
0xjunhao: Ack.

PR #2560, opened 2025-01-08
Automate host SSH key setup

shikharbhardwaj:
## Description

* Automate setting up the host SSH key by using a configured key to establish the initial connection and adding the generated VM host key before preparing the host.
* Update E2E tests to avoid setting up a new SSH key in hetzner on every run.

## Tests
- [x] Unit tests
- [x] HostNexus run

## Rollout
- [x] Generate and add a `ubicloud_ci_key_root` to the CI Hetzner account (key added to 1pass)
- [x] Update GH actions secret to add `HETZNER_SSH_PUBLIC_KEY` and `HETZNER_SSH_PRIVATE_KEY`
- [ ] Update docs to have devs setup their private key in `.env.rb`, and add operational public keys in dev/prod

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -92,7 +92,7 @@ module Config
   override :hetzner_connection_string, "https://robot-ws.your-server.de", string
   override :managed_service, false, bool
   override :sanctioned_countries, "CU,IR,KP,SY", array(string)
-  override :hetzner_ssh_key, string
+  override :hetzner_ssh_public_key, string
   override :minimum_invoice_charge_threshold, 0.5, float
   optional :cloudflare_turnstile_site_key, string
   optional :cloudflare_turnstile_secret_key, string
--- a/config.rb
+++ b/config.rb
@@ -54,6 +54,9 @@ module Config
   optional :omniauth_github_secret, string, clear: true
   optional :omniauth_google_id, string, clear: true
   optional :omniauth_google_secret, string, clear: true
+  optional :hetzner_ssh_private_key, string, clear: true
+  optional :hetzner_ssh_private_key_passphrase, string, clear: true
+  optional :operator_ssh_public_keys, string
 
   # :nocov:
   override :mail_driver, (production? ? :smtp : :logger), symbol
--- a/prog/test/hetzner_server.rb
+++ b/prog/test/hetzner_server.rb
@@ -37,21 +37,12 @@ class Prog::Test::HetznerServer < Prog::Test::Base
   label def fetch_hostname
     update_stack({"hostname" => hetzner_api.get_main_ip4})
 
-    hop_add_ssh_key
-  end
-
-  label def add_ssh_key
-    keypair = SshKey.generate
-    hetzner_api.add_key("ubicloud_ci_key_#{strand.ubid}", keypair.public_key)
-    update_stack({"hetzner_ssh_keypair" => Base64.encode64(keypair.keypair)})
-
     hop_reimage
   end
 
   label def reimage
     hetzner_api.reimage(
       frame["server_id"],
-      hetzner_ssh_key: hetzner_ssh_keypair.public_key,
       dist: "Ubuntu 24.04 LTS base"
     )
 
@@ -60,7 +51,7 @@ class Prog::Test::HetznerServer < Prog::Test::Base
 
   label def wait_reimage
     begin
-      Util.rootish_ssh(frame["hostname"], "root", [hetzner_ssh_keypair.private_key], "echo 1")
+      Util.rootish_ssh(frame["hostname"], "root", [Config.hetzner_ssh_private_key], "echo 1")
     rescue
       nap 15
     end
@@ -76,11 +67,6 @@ class Prog::Test::HetznerServer < Prog::Test::Base
       default_boot_images: Option::BootImages.map { _1.name }
     ).subject
     update_stack({"vm_host_id" => vm_host.id})
-
-    # BootstrapRhizome::start will override raw_private_key_1, so save the key in
-    # raw_private_key_2. This will allow BootstrapRhizome::setup to do a root
-    # ssh into the server.
-    vm_host.sshable.update(raw_private_key_2: hetzner_ssh_keypair.keypair)
 
     hop_wait_setup_host
   end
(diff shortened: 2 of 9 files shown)


config.rb
```diff
@@ -90,6 +90,8 @@ def self.test?
   override :managed_service, false, bool
   override :sanctioned_countries, "CU,IR,KP,SY", array(string)
   override :hetzner_ssh_key, string
```
shikharbhardwaj, line 92: Was wondering if i should rename this to `hetzner_ssh_public_key`

byucesoy: I think that makes sense. Also, it would make sense to put that to a separate commit, so that the "boring" refactoring code wouldn't be intertwined with more interesting logic changes.

config.rb
```diff
@@ -90,6 +90,8 @@ def self.test?
...
+  override :hetzner_ssh_private_key, "", string
+  override :hetzner_ssh_private_key_passphrase, "", string
```
enescakir, line 94: Instead of empty string, IMHO, we can do it optional

prog/vm/host_nexus.rb
```diff
@@ -38,6 +38,27 @@ def before_run
...
+
+    sshable.update(raw_private_key_1: SshKey.generate.keypair) if sshable.raw_private_key_1.nil?
+    public_key = sshable.keys.first.public_key
+
+    # Replace the key in production, append in other environments.
+    if Config.production?
+      init_sshable.cmd("echo '#{public_key}' > /root/.ssh/authorized_keys")
+    else
+      init_sshable.cmd("echo '#{public_key}' >> /root/.ssh/authorized_keys")
+    end
```
enescakir, line 56: We can use `Util.rootish_ssh` instead of new sshable
https://github.com/ubicloud/ubicloud/blob/2437c4c0fa564d227e61b0dd9ecbbfbcde61df0c/prog/bootstrap_rhizome.rb#L21

```suggestion
    if Config.hetzner_ssh_private_key
        sshable.update(raw_private_key_1: SshKey.generate.keypair) unless sshable.raw_private_key_1
        root_key = Net::SSH::Authentication::ED25519::PrivKey.read(Config.hetzner_ssh_private_key, Config.hetzner_ssh_private_key_passphrase).sign_key
        public_key = sshable.keys.first.public_key
        # We don't have `Config.operational_public_keys`, but it might beneficial to have to automate another step. Also it removes "Replace the key in production, append in other environments" step since developer can put his own SSH key to in the dev env
        public_key += "\n#{Config.operational_public_keys}" if Config.operational_public_keys
        Util.rootish_ssh(sshable.host, user, root_key.private_key, "echo '#{public_key.shellescape}' > /root/.ssh/authorized_keys")
    end
```

enescakir: I don't have context as much as @byucesoy, I wrote it as a prototype

shikharbhardwaj: This makes sense, removing `sshable` simplifies things a lot!
qq on the `operational_public_keys` config. Do we still need to wrap adding those keys under a `production?` check, to prevent adding operational keys to production hosts?

enescakir: We currently add whole team's ssh keys, which is protected with Yubikey, to production hosts manually during setup. Automating this process will help eliminate this step.

config.rb
```diff
@@ -90,6 +92,7 @@ def self.test?
   override :managed_service, false, bool
   override :sanctioned_countries, "CU,IR,KP,SY", array(string)
   override :hetzner_ssh_key, string
+  override :hetzner_ssh_private_key_passphrase, "", string
```
enescakir, line 95: We can make it optional and `clear: true` since it's a passphare

prog/bootstrap_rhizome.rb
```diff
@@ -10,11 +10,6 @@ def user
   end
 
   label def start
-    sshable.update(raw_private_key_1: SshKey.generate.keypair) if sshable.raw_private_key_1.nil?
```
enescakir, line 13: As I know some vm services such as PostgreSQL uses bootstrap rhizome to install packages to the vm. Does this affect these services? cc @byucesoy

config.rb
shikharbhardwaj: Done

prog/bootstrap_rhizome.rb
shikharbhardwaj: Yes had a brief chat about this with @byucesoy earlier. I just checked, it seems to have usage in PostgreSQL, MinIO and inference_endpoint as well. Wondering if we should defer this removal to a separate change.

shikharbhardwaj: Chatted with @byucesoy on this, going to revert this how it was before. We can take this one up in a separate change.

prog/test/hetzner_server.rb
```diff
@@ -37,21 +37,12 @@ def self.assemble(vm_host_id: nil)
...
-    hetzner_api.add_key("ubicloud_ci_key_#{strand.ubid}", keypair.public_key)
-    update_stack({"hetzner_ssh_keypair" => Base64.encode64(keypair.keypair)})
-
     hop_reimage
   end
 
   label def reimage
     hetzner_api.reimage(
       frame["server_id"],
-      hetzner_ssh_key: hetzner_ssh_keypair.public_key,
```
byucesoy, line 54: We still need to pass public key (of the temporary global ssh key), no?

shikharbhardwaj: `reimage` takes a default of `Config.hetzner_ssh_public_key` for that argument, so its passed

config.rb
```diff
@@ -54,6 +54,9 @@ def self.test?
   optional :omniauth_github_secret, string, clear: true
   optional :omniauth_google_id, string, clear: true
   optional :omniauth_google_secret, string, clear: true
+  optional :hetzner_ssh_private_key, string, clear: true
+  optional :hetzner_ssh_private_key_passphrase, string, clear: true
+  optional :operational_public_keys, string
```
byucesoy, line 59: Maybe rename this to `operator_ssh_public_keys`?

byucesoy (approved):

config.rb
shikharbhardwaj: Yes this reads better, done

PR #2562, opened 2025-01-08
Add hardware reset for vm host

0xjunhao:
Calls the Hetzner API to perform a hardware host reset and then reboot. This enhances remote server management and recovery, particularly when the host is unresponsive.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -20,7 +20,7 @@ class VmHost < Sequel::Model
   include ResourceMethods
   include SemaphoreMethods
   include HealthMonitorMethods
-  semaphore :checkup, :reboot, :destroy
+  semaphore :checkup, :reboot, :hardware_reset, :destroy
 
   def host_prefix
     net6.netmask.prefix_len
@@ -245,6 +245,14 @@ class VmHost < Sequel::Model
     end
 
     Hosting::Apis.reimage_server(self)
+  end
+
+  # Cuts power to a Server and starts it again. This forcefully stops it
+  # without giving the Server operating system time to gracefully stop. This
+  # may lead to data loss, it’s equivalent to pulling the power cord and
+  # plugging it in again. Reset should only be used when reboot does not work.
+  def hardware_reset
+    Hosting::Apis.hardware_reset_server(self)
   end
 
   def init_health_monitor_session
--- a/prog/vm/host_nexus.rb
+++ b/prog/vm/host_nexus.rb
@@ -156,6 +156,37 @@ class Prog::Vm::HostNexus < Prog::Base
     hop_verify_spdk
   end
 
+  label def prep_hardware_reset
+    register_deadline("wait", 20 * 60)
+    vm_host.vms_dataset.update(display_state: "rebooting")
+    decr_hardware_reset
+    hop_hardware_reset
+  end
+
+  # Cuts power to a Server and starts it again. This forcefully stops it
+  # without giving the Server operating system time to gracefully stop. This
+  # may lead to data loss, it’s equivalent to pulling the power cord and
+  # plugging it in again. Reset should only be used when reboot does not work.
+  label def hardware_reset
+    unless vm_host.allocation_state == "draining" || vm_host.vms_dataset.empty?
+      fail "Host has VMs and is not in draining state"
+    end
+
+    vm_host.hardware_reset
+
+    # Attempt to hop to reboot immediately after sending the hardware reset
+    # signal.
+    # The reboot may:
+    # 1. Fail, if the host is unreachable, which is a typical reason for a
+    #    hardware reset.
+    # 2. Succeed, if the host is reachable; however, the hardware reset may
+    #    interrupt the reboot.
+    # Regardless, the hardware reset will proceed, and upon completion, the
+    # host will receive a new boot id, allowing the sequence to continue
+    # without an additional reboot.
+    hop_reboot
+  end
+
   label def verify_spdk
     vm_host.spdk_installations.each { |installation|
       q_version = installation.version.shellescape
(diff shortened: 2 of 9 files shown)


model/vm_host.rb
```diff
@@ -247,6 +247,14 @@ def reimage
     Hosting::Apis.reimage_server(self)
   end
 
+  # Cuts power to a Server and starts it again. This forcefully stops it
+  # without giving the Server operating system time to gracefully stop. This
+  # may lead to data loss, it’s equivalent to pulling the power cord and
+  # plugging it in again. Reset should only be used when reboot does not work.
+  def hardware_reset
+    Hosting::Apis.hardware_reset_server(self)
```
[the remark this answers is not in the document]
enescakir: Makes sense. We need some form of protection; otherwise, an operator might accidentally reboot a host.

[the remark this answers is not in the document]
enescakir: We can also request the VM host's UBID for confirmation. It may seem overly cautious, but it helps the operator feel more comfortable.

```
    if vms.count > 0
      puts "#{self} #{sshable.host} #{location} #{data_center}"
      puts "Virtual machines:"
      vms.each { puts "#{_1} #{_1.name.ljust(24)} #{_1.display_state}" }

      print "\nEnter UBID of the vm host to continue: "
      fail "You entered UBID wrong" unless gets.chomp == ubid
    end
```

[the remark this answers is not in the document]
jeremyevans: Maybe only require entry of UBID if there are VMs or it is in accepting state?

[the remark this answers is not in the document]
fdr: do we do anything similar for the other (operating system based) reboot process? It seems like we're being too precious here relative to what's going on status quo there. If we need to be safer, I think we need to treat the two together rather than gold-plating this one, unless I am not fully aware of the status quo

spec/lib/hosting/hetzner_apis_spec.rb
```diff
@@ -38,6 +38,25 @@
     end
   end
 
+  describe "reset" do
+    it "can reset a server" do
+      expect(Config).to receive(:hetzner_ssh_key).and_return("ssh-rsa TestKey").at_least(:once)
```
fdr, line 43: can probably factor this into a `before`

fdr: saw the last one, which returns nil for the hetzner key, nvm

fdr: aha we found that the ssh key is a holdover from the reimage thing, can delete stuff (and @0xjunhao did)

model/vm_host.rb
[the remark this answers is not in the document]
0xjunhao: From what I understand, the intended way to use it is similar to rebooting a host, i.e., by increasing the semaphore. Therefore, if we prompt the user, the prompt only appears in respirate and gets stuck there, if respirate is running. When stuck in respirate, running vmh.strand.run in pry simply returns false.

If there have been no issues with reboot, how about we keep things consistent and simple by either:
1. Not adding any additional protections, or
2. Checking if the host is either not in the accepting state or has no VMs, for both cases.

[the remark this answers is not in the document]
bsatzger: I'm fine with just add a simple check. E.g. just wait until either `vm_host.vms_dataset` is empty or  `vm_host.allocation_state` is `draining` before doing the actual reset.

Could look something like this:

```ruby
  label def hardware_reset
    if vm_host.vms_dataset.empty? || vm_host.allocation_state == "draining"
      vm_host.hardware_reset
      hop_reboot
    end

    nap 30
  end
```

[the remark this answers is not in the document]
0xjunhao: Done. I set it to raise an error in this case, so the operator will get an error like "RuntimeError: Host has VMs and is not in draining state". This way, they will know that they need to change the allocation_state if they really intend to proceed with the reset.

prog/vm/host_nexus.rb
```diff
@@ -156,6 +156,37 @@ def before_run
...
+    decr_hardware_reset
+    hop_hardware_reset
+  end
+
+  # Cuts power to a Server and starts it again. This forcefully stops it
+  # without giving the Server operating system time to gracefully stop. This
+  # may lead to data loss, it’s equivalent to pulling the power cord and
+  # plugging it in again. Reset should only be used when reboot does not work.
+  label def hardware_reset
+    if vm_host.vms_dataset.count > 0 && vm_host.allocation_state != "draining"
```
bsatzger, line 171: Nit: `empty?` might be more efficient than `count`, maybe: `unless vm_host.vms_dataset.empty? || vm_host.allocation_state == "draining"`

prog/vm/host_nexus.rb
```diff
@@ -156,6 +156,37 @@ def before_run
...
+      raise "Host has VMs and is not in draining state"
```
bsatzger, line 172: nit: more of an FYI. There is `fail` and `raise` and I think we'd more commonly use `fail` in this case. But I don't think it really matters. For context:  https://github.com/ubicloud/ubicloud/commit/aa6efb1bcfa20fb5e045ff241b2106f7891fe596

bsatzger (approved):
thanks

prog/vm/host_nexus.rb
jeremyevans: `unless vm_host.allocation_state == "draining" || vm_host.vms_dataset.empty?` may be better as it avoids the query if the host is in not in draining state.  Not in this particular case, as an exception is being raised if the branch is true, but as a general practice.

prog/vm/host_nexus.rb
jeremyevans: I mentioned to Daniel a couple days ago that the only reason there isn't a Rubocop rule to tell people to use `raise` instead of `fail` is that nobody uses `fail` to begin with. :)  But yes, for consistency, best to use `fail` in the progs.

0xjunhao: Done.

prog/vm/host_nexus.rb
0xjunhao: Done.

PR #2563, opened 2025-01-08
Deduplicate connection code in Hetzner API client

0xjunhao:
Not entirely sure about the content type in the connection request. Do you think it will work if we always have that? @furkansahin


0xjunhao:
> Not entirely sure about the content type in the connection request. Do you think it will work if we always have that? @furkansahin

Nvm, I checked them one by one with my live machine, and they all works. @fdr

bsatzger (approved):
lgtm. thanks for reducing lines of code!

fdr:
:shipit:

fdr (approved):

furkansahin (approved):

PR #2564, opened 2025-01-08
Drop deleted record table

byucesoy:
This table is replaced by archived_records table.


enescakir (approved):

jeremyevans (approved):
You can simplify this using `revert do` instead of `down do`, and removing the `up do` branch.

bsatzger (approved):
thank you

PR #2565, opened 2025-01-09
Fixes to make LB cert-server safer against synchronisation issues

furkansahin:
## Add put-certificate functionality to setup-cert-server
We used to write certificate and the key remotely into the host.
However, this was an anti pattern because the cert-server functionality
was mostly left to cert_server_setup script. It also made it difficult
to write the cert and key files in a safe way into the correct paths. In
this commit, we are adding the necessary functionality to the
setup-cert-server script.

## Refactor setup and stop_and_remove steps into cert_server_setup
Instead of calling 3 functions independently through the
setup-cert-server, we create wrapper functions in cert_server_setup and
make use of them. This way, the setup-cert-server is easier to digest.

We also add a new parameter RestartSec to the systemd service. The main
reason is that, if the VM is attached to a load balancer way before the
necessary interfaces are setup, the metadata-endpoint was quickly
failing and hit the restart count limit. This way, we slow down the
restarts and let it quickly catch up with VM provisioning.

This commit also introduces the tests for the necessary scripts,
previously, there were no tests at all.

## Fix VM destroy for a Load Balancer detached VM
If the customer very quickly detaches the VM from the LB and immediately
starts the destroy sequence, because of the async nature of the
scheduling, we might end up in a state that vm.load_balancer returns
nil. That was causing nil ref for vm.load_balancer.evacuate_vm(vm) call.
We safeguard that by moving the code block into an if check.


rhizome/host/lib/cert_server_setup.rb
```diff
@@ -104,19 +116,20 @@ def create_service
...
-
 CERT_SERVICE
     )
+
+    r "systemctl daemon-reload"
   end
 
   def enable_and_start_service
-    r "systemctl enable #{service_file_path}"
+    r "systemctl enable #{service_name}"
```
bsatzger, line 126: I think these two lines could be combined as `systemctl enable --now ...`

rhizome/host/lib/cert_server_setup.rb
```diff
@@ -104,19 +116,20 @@ def create_service
...
     r "systemctl start #{service_name}"
   end
 
   def stop_and_remove_service
-    r "systemctl stop #{service_name}"
-    r "systemctl disable #{service_name}"
+    r "systemctl stop #{service_name} && systemctl disable #{service_name}" if File.exist?(service_file_path)
```
bsatzger, line 131: systemctl stop and disable could be combined as ``systemctl disable --now ...``

prog/vm/nexus.rb
```diff
@@ -457,14 +457,15 @@ def write_params_json
   end
 
   label def wait_lb_expiry
-    unless vm.lb_expiry_started_set?
-      vm.incr_lb_expiry_started
-      vm.load_balancer.evacuate_vm(vm)
-      nap 30
+    if vm.load_balancer
```
bsatzger, line 460: maybe: `if (lb = vm.load_balancer)` and then replace `vm.load_balancer` with `lb` inside `if`.

bsatzger (approved):

PR #2567, opened 2025-01-09
Map runner label aliases to the actual labels automatically

enescakir:
We have several labels that serve as an alias for another label. Currently, we keep the specifications of each label separately, which causes a lot of duplicate data in the YAML file and makes it hard to see the relationships between them. We can store the alias information in the YAML file and automatically map the aliases to the actual labels when loading the file.

### DIFF
--- a/lib/github.rb
+++ b/lib/github.rb
@@ -38,7 +38,10 @@ module Github
   # :nocov:
 
   def self.runner_labels
-    @runner_labels ||= YAML.load_file("config/github_runner_labels.yml").to_h { [_1["name"], _1] }
+    @runner_labels ||= begin
+      labels = YAML.load_file("config/github_runner_labels.yml").to_h { [_1["name"], _1] }
+      labels.transform_values { |v| (a = v["alias_for"]) ? labels[a] : v }
+    end
   end
 
   def self.failed_deliveries(since, max_page = 50)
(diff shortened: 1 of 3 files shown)


lib/github.rb
```diff
@@ -38,7 +38,10 @@ def self.freeze
   # :nocov:
 
   def self.runner_labels
-    @runner_labels ||= YAML.load_file("config/github_runner_labels.yml").to_h { [_1["name"], _1] }
+    @runner_labels ||= begin
+      labels = YAML.load_file("config/github_runner_labels.yml").to_h { [_1["name"], _1] }
+      labels.transform_values { |v| (a = v["alias_for"]) ? labels[a] : v }
```
bsatzger, line 43: we probably have a problem if a label is an alias pointing to another alias. should we add a test to make sure after transformation no label exists with `alias_for` anymore?

bsatzger (approved):

lib/github.rb
enescakir: Good idea. It also helpful if we make a typo in alias. I added following test.

```ruby
  it "can map all aliases to actual tag" do
    expect(described_class.runner_labels.values).to be_all
  end
```

velioglu (approved):

PR #2568, opened 2025-01-09
Support VmHost#incr_graceful_reboot to gracefully reboot a host

jeremyevans:
In a graceful reboot, the host is set to draining, and when it no longer has any VMs, it transitions to reboot.  After the reboot, it is automatically put back in accepting state.


fdr (approved):
I watched jeremy write this so I kinda know it pretty well.

fdr:
Added @bsatzger since his project with @0xjunhao is kinda in the same headspace

bsatzger (approved):

PR #2569, opened 2025-01-09
Rename InferenceToken to InferenceApiKey

bsatzger:
As the term `token` in the context of generative AI refers to text/data units, calling tokens that allow to access inference endpoints `inference token` might be confusing. For that reason, we consistently refer to such tokens as `api key`.

The renaming is performed at all levels. Most prominent change for users is on these views:
Screenshot 2025-01-09 at 22 07 10
Screenshot 2025-01-09 at 22 08 34
Screenshot 2025-01-09 at 22 09 03


routes/project/inference_api_key.rb
```diff
@@ -0,0 +1,28 @@
...
+      r.post true do
+        authorize("InferenceApiKey:create", @project.id)
+        it = DB.transaction { ApiKey.create_inference_api_key(@project) }
+        flash["notice"] = "Created Inference API Key with id #{it.ubid}"
+        r.redirect "#{@project.path}/inference-api-key"
+      end
+
+      r.delete String do |ubid|
+        if (it = ApiKey.from_ubid(ubid))
+          authorize("InferenceApiKey:delete", it.id)
```
jeremyevans, line 20: Not an issue with this change, but this technically leaks information about whether a ubid represents a valid ApiKey, even if the ApiKey is associated to a different project.  Not sure that matters, and Clover may leak the same information elsewhere.  To avoid this case leaking, you could switch to `it = inference_api_key_ds.with_pk(UBID.to_uuid(ubid))`.

jeremyevans (approved):
All changes outside the migration appear mechanical, which is good.  I manually checked the migration and it looks good.

If we assume nobody is currently using `INFERENCE_TOKEN` in the API (or we are OK breaking compatibility there), this looks good to me.

routes/project/inference_api_key.rb
bsatzger: thanks for the catch. applied your suggestion.

PR #2576, opened 2025-01-10
Start using encryption for MinIO drives

furkansahin:
At the time when we first introduced no-raid VM setup for MinIO servers, we did not have encryption support. Now, we do. So, we are starting to use the encryption for MinIO servers, too.


byucesoy (approved):

furkansahin:
Note to @pykello , we are setting up MinIO nodes with multiple drives without raid setup and distribute the VM disks to these devices. I had created a single instance MinIO node in my local and tested if it's working. It would be still good for you to check if this is a valid scenario with encryption.

pykello (approved):

PR #2577, opened 2025-01-10
Move tokens to a separate page and add to the sidebar

enescakir:
As a customer, I’m not likely to visit the `Users` page to create a personal access token.

Let’s make it easier to find by moving it to its own page and adding it to the sidebar.


jeremyevans (approved):
Looks good.  I should add some documentation about tokens to the documentation site.  Also, we should consider whether we want to drop the old API login approach and only allow personal access tokens in the API.

PR #2581, opened 2025-01-11
Update billing rate for qwq-32b-preview model

bsatzger:
We have decided on a pricing of `$0.60` per million tokens for the `Qwen/QwQ-32B-Preview` model for our upcoming announcements. As a result, we are lowering the rate from `$0.90` per million tokens to `$0.60`. Since no invoices have been generated with the previous rate, we update it in-place.


enescakir (approved):

byucesoy (approved):
This meter is not used in any invoice so far, right? Then it is safe to update directly.

PR #2583, opened 2025-01-12
Table component improvements

enescakir:
- **Unify empty states in tables**
  Some pages display a message in the table body when empty, while others
  show a full-page empty state. This PR unifies the empty states to
  consistently show a message in the table body. It removes the
  conditional wrapping the table. The new method of displaying the empty
  state is also more consistent with the rest of the app and includes the
  page header.
  
| Before | After |
|--------|-------|
| CleanShot 2025-01-13 at 00 02 40@2x | CleanShot 2025-01-13 at 00 02 30@2x |

- **Add right items argument to the table card**
  

- **Add ability to apply an extra class to the column in the row.**
  This feature lets you add a custom class to the column in the row, which
  is useful for applying specific styles, like right-aligning the last
  column.
  

- **Migrate more tables to the new table component**
  

- **Unify value content of kv and table card components**

### DIFF
--- a/views/components/value_content.erb
+++ b/views/components/value_content.erb
@@ -1,0 +1,15 @@
+<%# locals: (content:, opts: nil) %>
+<% opts ||= {} %>
+<% if opts[:copyable] && content %>
+  <%== render("components/copyable_content", locals: { content:, revealable: opts[:revealable] }) %>
+<% elsif opts[:link] %>
+  <a href="<%= opts[:link] %>" class="font-medium text-orange-600 hover:text-orange-700">
+    <%= content %>
+  </a>
+<% elsif opts[:component] %>
+  <%== render("components/#{content}", locals: opts[:component]) %>
+<% elsif opts[:escape] == false %>
+  <%== content %>
+<% else %>
+  <%= content || "-" %>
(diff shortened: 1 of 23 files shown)


views/components/value_content.erb
```diff
@@ -0,0 +1,17 @@
+<%# locals: (content:, opts: nil) %>
+<% opts ||= {} %>
+<% if opts[:copyable] && content %>
+  <%== render("components/copyable_content", locals: { content:, revealable: opts[:revealable] }) %>
+<% elsif opts[:revealable] && content %>
+  <%== render("components/revealable_content", locals: { content: }) %>
```
jeremyevans, line 6: I realize this just moves the logic, but in my work in #2551, I determined that we don't have any components that are revealable but not copyable, so this branch is never taken, and can be removed.

jeremyevans (approved):
Looks good.  Might conflict with some of my commits in #2551, but I'll handle that.

enescakir:
> Looks good. Might conflict with some of my commits in #2551, but I'll handle that.

Given that your PR is more complex than mine, I've rebased my PR to yours

PR #2589, opened 2025-01-13
Add VmHostSlice Migrations

pykello:
## Add VmHostSlice Migrations.
We'll put VMs in cgroup slices. This change adds the migrations to enable that. A VmHostSlice is used to track resource usage of a cgroup slice.

Each slice reserves a subset of host's cpus & part of its memory. Then, VMs can be allocated into the slice.

`total_memory_gib` & `used_memory_gib` are used to make sure that we will have enough memory in the host for the VMs that'll be added to the slice in future. When a VmHostSlice record is allocated, it will reserve `total_memory_gib` from the host.

Each slice can be either be `dedicated` or `shared`, as specified in the `is_shared` column. We will use `shared` slices to host burstable instances.

## Add a bare VmHostSlice model class.
A VmHostSlice is used to track resource usage of a cgroup slice. This change just adds a bare model class for it, and we will add more logic here when adding VmHostSlice lifetime management & monitoring methods later.


migrate/20250117_vm_host_slice.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_enum(:vm_host_slice_type, %w[dedicated shared])
```
jeremyevans, line 5: Personally, I dislike enums.  If you ever want to add to the enum, you are looking at an irreversible migration, since you cannot delete an enum member.  If you never want to add to the enum, then for this, a boolean column would work.  I generally recommend a foreign key to a static table instead of an enum.  @fdr may have different feelings, though.

migrate/20250117_vm_host_slice.rb
```diff
@@ -0,0 +1,33 @@
...
+      column :enabled, :bool, null: false, default: false
+      column :type, :vm_host_slice_type, default: "dedicated", null: false
+      column :cores, Integer, null: false
+      column :total_cpu_percent, Integer, null: false
+      column :used_cpu_percent, Integer, null: false
+      column :total_memory_gib, Integer, null: false
+      column :used_memory_gib, Integer, null: false
+      column :family, :text, null: false
+      column :created_at, :timestamptz, null: false, default: Sequel.lit("now()")
+      foreign_key :vm_host_id, :vm_host, type: :uuid
```
jeremyevans, line 19: Does it make sense to have a host slice not tied to a host?  Maybe it does, but if not, this should use `null: false`.

migrate/20250117_vm_host_slice.rb
```diff
@@ -0,0 +1,33 @@
...
+      column :name, :text, null: false
+      column :enabled, :bool, null: false, default: false
+      column :type, :vm_host_slice_type, default: "dedicated", null: false
+      column :cores, Integer, null: false
+      column :total_cpu_percent, Integer, null: false
+      column :used_cpu_percent, Integer, null: false
+      column :total_memory_gib, Integer, null: false
+      column :used_memory_gib, Integer, null: false
+      column :family, :text, null: false
+      column :created_at, :timestamptz, null: false, default: Sequel.lit("now()")
```
jeremyevans, line 18: If you want to avoid the `Sequel.lit`, you can use `Sequel::CURRENT_TIMESTAMP`

jeremyevans (commented):
No issues in the Ruby code that I can see.

Note that this appears to include the cache updates for `VmHostSlice` in the migration commit.  Otherwise, there would be separate cache updates when adding the `VmHostSlice` model.  That doesn't cause any problems, though.

migrate/20250117_vm_host_slice.rb
macieksarnowicz: @jeremyevans - good point. When I operate in a codebase new to me, I try to follow examples of other classes that exist so far, assuming that is the standard that the team follows. So in this case, I used enum, because VmHost also uses enum for allocation_state column. 
In this case I think the two states are what we will need in any foreseeable future, so we can change it to is_shared or is_dedicated boolean.

migrate/20250117_vm_host_slice.rb
macieksarnowicz: Good point. when we create a new slice we always pass in the host id.

migrate/20250117_vm_host_slice.rb
macieksarnowicz: OK. Again, this is following prior code example, but we can change. What is the downside of using Sequel.lit

migrate/20250117_vm_host_slice.rb
fdr: I also mostly have disliked enums for exactly this reason, other team member seem to like them more, yet, I cannot deny that even with the problem of not being able to scrub old values, it does reject _every other_ typo that could be made, e.g. code like: 

`foohost.update(allocation_state: "draiining")`

(typo intentional) crash.

I suppose the purer version would be a check constraint, but that blows up the implementation somewhat, whereas enum migrations are so terse.

fdr: another variant might be careful use of the `domain` feature in Postgres, I think you can attach a check constraint for those and any checking it does is lazy

fdr: IMO, probably not the domain (heh) of this patch to solve, but if we wanted to swap out all our enums for some other approach that doesn't, like, lose all of the typo checking advantages and doesn't blow up the implementation, I'd be down for that.

migrate/20250117_vm_host_slice.rb
jeremyevans: `Sequel::CURRENT_TIMESTAMP` is more portable as well as more descriptive.  I don't expect anyone to try to run Ubicloud on something other than PostgreSQL, but the more descriptive part is nice.  Additionally, Sequel.lit uses raw SQL, so it should be audited closely.  Obviously, with a static string here, the is no security issue, but as a good general practice, only use Sequel.lit when there is not a way Sequel offers to generate the SQL you need.  For generic function calls, `Sequel.function(:function_name)` can be used.

pykello:
> Note that this appears to include the cache updates for `VmHostSlice` in the migration commit. Otherwise, there would be separate cache updates when adding the `VmHostSlice` model. That doesn't cause any problems, though.

@jeremyevans I just edited the commits. When I did `rake test_up` in the first commit, it modified `cache/schema.cache` and then after 2nd commit it modified both `cache/index.cache` and `cache/schema.cache`. Does that sound about right or did I miss anything?

migrate/20250117_vm_host_slice.rb
pykello: I changed the column to boolean `is_shared`

fdr (approved):

jeremyevans:
> > Note that this appears to include the cache updates for `VmHostSlice` in the migration commit. Otherwise, there would be separate cache updates when adding the `VmHostSlice` model. That doesn't cause any problems, though.
> 
> @jeremyevans I just edited the commits. When I did `rake test_up` in the first commit, it modified `cache/schema.cache` and then after 2nd commit it modified both `cache/index.cache` and `cache/schema.cache`. Does that sound about right or did I miss anything?

That sounds correct.  The first commit changes a table for an existing model to add a column, the second adds a new model, which results in the schema and index for the model being cached.

PR #2597, opened 2025-01-14
ConnectedSubnets tests nap in case the listening service is not up yet

furkansahin:


prog/test/connected_subnets.rb
```diff
@@ -212,5 +212,7 @@ def test_connection(to_connect_ip, connecting, should_fail: false, ipv4: true)
   def start_listening(ipv4: true)
     vm_to_be_connected.sshable.cmd("sudo systemctl stop listening_ipv#{ipv4 ? "6" : "4"}.service")
     vm_to_be_connected.sshable.cmd("sudo systemctl start listening_ipv#{ipv4 ? "4" : "6"}.service")
+
+    nap 1 if vm_to_be_connected.sshable.cmd("sudo systemctl status listening_ipv#{ipv4 ? "4" : "6"}.service").include?("inactive")
```
enescakir, line 216: nit: we might use `is-active` similar to https://github.com/ubicloud/ubicloud/blob/009675777a60b492aaacd438487ae6da581c93e7/prog/vm/nexus.rb#L505

enescakir, line 216: nit: no need to check the same condition 3 times
```suggestion
    service = "listening_ipv#{ipv4 ? "6" : "4"}.service"
    vm_to_be_connected.sshable.cmd("sudo systemctl stop #{service}")
    vm_to_be_connected.sshable.cmd("sudo systemctl start #{service}")

    nap 1 if vm_to_be_connected.sshable.cmd("sudo systemctl status #{service}").include?("inactive")
```

enescakir, line 216: We stop, start, then check its status. If it hasn't started, we stop and start once more. If it doesn't start promptly, we repeat this process again and again. Should we consider adding a brief sleep between the start and the status check?

PR #2598, opened 2025-01-14
Kubernetes v0.1

aamederen:
### Add migration file for Kubernetes Service models

KubernetesCluster and KubernetesNodepool

### Introduce Kubernetes Cluster Provisioning Logic

KubernetesClusterNexus.assemble call creates a KubernetesCluster object
and manages the creation of control plane nodes using the
ProvisionKubernetesNode prog.

Nodes are created from plain images and k8s components are installed during
the provisioning, increasing the time to provision. Later, we will switch to
dedicated VM images.


migrate/20250110_add_kubernetes_models.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:kubernetes_cluster) do
+      column :id, :uuid, primary_key: true, default: Sequel.lit("gen_random_uuid()")
```
enescakir, line 6: Since we use UBIDs, we don't set a default UUID at the database level for `id` columns

https://github.com/ubicloud/ubicloud/blob/main/migrate/20230721_remove_default_ids.rb

migrate/20250110_add_kubernetes_models.rb
```diff
@@ -0,0 +1,46 @@
...
+      column :location, :text, collate: '"C"', null: false
+      column :created_at, :timestamp, null: false, default: Sequel::CURRENT_TIMESTAMP
+
+      check { (cp_node_count =~ [1, 3]) }
+      foreign_key :private_subnet_id, :private_subnet, type: :uuid, null: false
+      foreign_key :api_server_lb_id, :load_balancer, type: :uuid, null: true
+    end
+
+    create_table(:kubernetes_clusters_vms) do
+      primary_key :id, :uuid, default: Sequel.lit("gen_random_uuid()")
```
enescakir, line 19: Do we need the `id` column in the middle table? Isn't the `[:kubernetes_cluster_id, :vm_id]` pair enough?

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -0,0 +1,43 @@
...
+
+require_relative "../../model"
+
+class KubernetesCluster < Sequel::Model
+  one_to_one :strand, key: :id
+  one_to_one :api_server_lb, class: :LoadBalancer, key: :id, primary_key: :api_server_lb_id
+  many_to_one :private_subnet
+  many_to_many :vms do |ds|
+    ds.order_by(:created_at)
+  end
```
enescakir, line 11: nit: https://github.com/jeremyevans/sequel/blob/master/doc/association_basics.rdoc#label-3Aorder
```suggestion
  many_to_many :vms, order: :created_at
```

migrate/20250110_add_kubernetes_models.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:kubernetes_cluster) do
```
enescakir, line 5: I strongly suggest adding a `project_id` column and filling it.
https://github.com/ubicloud/ubicloud/pull/2594

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,135 @@
...
+    vm.sshable.cmd("echo '#{ip} #{kubernetes_cluster.endpoint}' | sudo tee -a /etc/hosts")
+  end
+
+  label def start
+    name = "#{kubernetes_cluster.name.downcase}-control-plane-#{SecureRandom.alphanumeric(5).downcase}"
+
+    vm_st = Prog::Vm::Nexus.assemble_with_sshable(
+      "ubi",
+      kubernetes_cluster.projects.first.id,
+      # we should reiterate how we name the vm. some how correlate it to the vm's info.
```
enescakir, line 24: Generally, we use the related object's UBID. For example, the name of the runner VM is the runner's UBID, and the name of the Postgres server's VM is its UBID

enescakir: We might use ubid + random key instead of name + random

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,135 @@
...
+    when "Succeeded"
+      pop vm_id: vm.id
+    when "NotStarted"
+      ps = kubernetes_cluster.private_subnet
+      vm.sshable.cmd("common/bin/daemonizer 'sudo /home/ubi/kubernetes/bin/init-cluster #{kubernetes_cluster.name} #{kubernetes_cluster.endpoint} 443 #{ps.net4} #{ps.net6} #{vm.nics.first.private_ipv4}' init_kubernetes_cluster")
+      nap 30
+    when "InProgress"
+      nap 10
+    when "Failed"
+      puts "INIT CLUSTER FAILED"
```
enescakir, line 105: You should use `Clog.emit` instead of `puts`

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  label def join_control_plane
+    case vm.sshable.cmd("common/bin/daemonizer --check join_control_plane")
+    when "Succeeded"
+      pop vm_id: vm.id
+    when "NotStarted"
+      join_token = kubernetes_cluster.vms.first.sshable.cmd("sudo kubeadm token create --ttl 24h --usages signing,authentication").tr("\n", "")
+      certificate_key = kubernetes_cluster.vms.first.sshable.cmd("sudo kubeadm init phase upload-certs --upload-certs")[/certificate key:\n(.*)/, 1]
+      discovery_token_ca_cert_hash = kubernetes_cluster.vms.first.sshable.cmd("sudo kubeadm token create --print-join-command")[/discovery-token-ca-cert-hash (.*)/, 1]
```
enescakir, line 121: nit: might have better name
```suggestion
      master_sshable = kubernetes_cluster.vms.first.sshable
      join_token = master_sshable.cmd("sudo kubeadm token create --ttl 24h --usages signing,authentication").tr("\n", "")
      certificate_key = master_sshable.cmd("sudo kubeadm init phase upload-certs --upload-certs")[/certificate key:\n(.*)/, 1]
      discovery_token_ca_cert_hash = master_sshable.cmd("sudo kubeadm token create --print-join-command")[/discovery-token-ca-cert-hash (.*)/, 1]
```

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,135 @@
...
+
+      vm.sshable.cmd("common/bin/daemonizer 'sudo kubernetes/bin/join-control-plane-node #{kubernetes_cluster.endpoint}:443 #{join_token} #{discovery_token_ca_cert_hash} #{certificate_key}' join_control_plane")
```
enescakir, line 123: Should we call `.shellescape` for these values?

migrate/20250110_add_kubernetes_models.rb
```diff
@@ -0,0 +1,46 @@
...
+      foreign_key :vm_id, :vm, type: :uuid, null: false
+
+      index [:kubernetes_cluster_id, :vm_id], unique: true
+    end
+
+    create_table(:kubernetes_nodepool) do
+      column :id, :uuid, primary_key: true, default: Sequel.lit("gen_random_uuid()")
+      column :name, :text, null: false
+      column :node_count, :integer, null: false
+      column :location, :text, collate: '"C"', null: false
```
enescakir, line 31: Can its value differ from `kubernetes_cluster.location`?

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -0,0 +1,43 @@
+#  frozen_string_literal: true
+
+require_relative "../../model"
+
+class KubernetesCluster < Sequel::Model
+  one_to_one :strand, key: :id
+  one_to_one :api_server_lb, class: :LoadBalancer, key: :id, primary_key: :api_server_lb_id
+  many_to_one :private_subnet
+  many_to_many :vms do |ds|
```
enescakir, line 9: k8s cluster has vms. Also nodepool has vms. I assume `cluster.vms` doesn't include node's vms, and these are cp ones. Should rename it as `cp_vms`?

prog/kubernetes/provision_kubernetes_node.rb
enescakir: Mezmo can't parse it

prog/kubernetes/provision_kubernetes_node.rb
fdr: Hmm this is also in a quote, so that means there's another layer of shell, which, fair question, should we...not do that? and rely on argv? is sudo even necessary given daemonizer implies root?

migrate/20250110_add_kubernetes_models.rb
aamederen: done!

prog/kubernetes/provision_kubernetes_node.rb
aamederen: I didn't understand this part:

> ... should we...not do that? and rely on argv?

Do you suggest pushing the info as a json in stdin for example? If not, an example would be awesome :)

prog/kubernetes/provision_kubernetes_node.rb
aamederen: This was a temporary line, but since we have Clog.emit, maybe it could be permanent now.

prog/kubernetes/provision_kubernetes_node.rb
fdr: well, it was more like...do you need to quote it, or do you need only to terminate the command's option parsing? 

e.g. 

```
common/bin/daemonizer join_control_plane -- kubernetes/bin/join-control-plane-node #{kubernetes_cluster.endpoint}:443 #{join_token} #{discovery_token_ca_cert_hash} #{certificate_key}
```

....I see the problem now, with `join_control_plane` as a positional argument, it's plausible daemonizer is better to be changed, to have a v2, to avoid nested argument parsing when unnecessary. Also probably to avoid writing to /var/log and instead rely on the systemd journal.

fdr: okay, I'm going to fool around with daemonizer, watch this space

fdr: okay, I think I can improve this a bit. We also have some patches in this patch to daemonzer that may be a bit too disruptive to do in-situ, we may need to make daemonizer2 and lump together a few simplifications.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,135 @@
...
+    hop_install_software
+  end
+
+  label def install_software
+    nap 5 unless vm.strand.label == "wait"
+    vm.sshable.cmd <<BASH
+  set -ueo pipefail
+  echo "net.ipv6.conf.all.forwarding=1\nnet.ipv6.conf.all.proxy_ndp=1\nnet.ipv4.conf.all.forwarding=1\nnet.ipv4.ip_forward=1" | sudo tee /etc/sysctl.d/72-clover-forward-packets.conf
+  sudo sysctl --system
+  sudo modprobe br_netfilter
```
fdr, line 49: what's this for?

rhizome/kubernetes/bin/init-cluster
```diff
@@ -0,0 +1,44 @@
...
+  extraArgs:
+    allocate-node-cidrs: "false"
+nodeRegistration:
+  kubeletExtraArgs:
+    pod-cidr: "#{vm_cidr}"
+etcd:
+  local:
+    dataDir: "/var/lib/etcd"
+YAML
+
```
fdr, line 39: should we consider a yaml dump here? this is asking for quoting issues. I know vmsetup does it this way (and I did it) for cloud-init but it's not to be repeated in a more civilized era

spec/prog/kubernetes/provision_kubernetes_node_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+  describe "#write_hosts_file_if_needed" do
+    it "exits early if the environment is not dev" do
+      expect(prog.vm).not_to receive(:sshable)
+      expect(Config).to receive(:development?).and_return(false)
+      prog.write_hosts_file_if_needed
+    end
+
+    it "exits early if /etc/hosts file contains an entry about the cluster endpoint already" do
+      sshable = instance_double(Sshable)
+      allow(prog.vm).to receive(:sshable).and_return(sshable)
```
fdr, line 42: why allow here? I only use allow in `before` blocks if it really helps with some difficult combination of methods to stub; I use `expect` so that when it goes obsolete, it complains and I can delete it. `allow` inevitably converges on plaque that accumulates.

fdr:
we probably want a kubernetes service project instead of squirting managed stuff into the customer's account (that's the status quo, right?). I realize that makes VM & subnet pages, etc invisible normally, and we may need to do something to make them visible...but not interact-able in every single possible way (as a managed service).

It otherwise seems like too much a burden: every time we add a feature to the compute/network/whatever layers, every single thing must compose with how the manage service works. seems a bit much?

prog/kubernetes/provision_kubernetes_node.rb
aamederen: >We also have some patches in this patch to daemonzer that may be a bit too disruptive to do in-situ, we may need to make daemonizer2 and lump together a few simplifications.

For clarification, if you're talking about [this](https://github.com/ubicloud/ubicloud/pull/2545/files#diff-5935f6c67f806cf538ca456dd1bf15e6a9f4be25b8e7498d3df7d28b4cf94124R68), I'm not planning to bring the daemonizer changes here.

migrate/20250110_add_kubernetes_models.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:kubernetes_cluster) do
```
enescakir, line 5: Jeremy's PR will likely be merged before this one. We should add a unique constraint for the project_id, location, and name.

https://github.com/ubicloud/ubicloud/pull/2602

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -0,0 +1,65 @@
...
+  one_to_one :api_server_lb, class: :LoadBalancer, key: :id, primary_key: :api_server_lb_id
+  many_to_one :private_subnet
+  many_to_one :project
+  many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
+
+  dataset_module Pagination
+
+  include ResourceMethods
+  include SemaphoreMethods
+  include Authorization::HyperTagMethods
```
enescakir, line 16: Jeremy's PR will likely be merged before this one. We should get rid of `Authorization::HyperTagMethods` and `hyper_tag_name`.

https://github.com/ubicloud/ubicloud/pull/2616

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def path
+    "/location/#{display_location}/kubernetes-cluster/#{name}"
+  end
+
+  def endpoint
+    api_server_lb.hostname
+  end
+
+  def disassociate_cp_vm(vm)
```
enescakir, line 36: When you define a `many_to_many` association, `remove_*` methods are added automatically. `remove_cp_vm` might perform the same function. cc @jeremyevans as an expert

https://github.com/jeremyevans/sequel/blob/master/doc/association_basics.rdoc#label-Methods+Added

migrate/20250110_add_kubernetes_models.rb
enescakir: Merged

model/kubernetes/kubernetes_cluster.rb
enescakir: Merged

prog/kubernetes/provision_kubernetes_node.rb
aamederen: Kubeadm required this for some time, but lately it was removed (https://github.com/kubernetes/kubernetes/pull/123464/files) it seems and pushed the requirement to CNIs. I'll try if we can get away without setting this now.

rhizome/kubernetes/bin/init-cluster
aamederen: image

prog/kubernetes/provision_kubernetes_node.rb
aamederen: Removed it and  things are fine

migrate/20250110_add_kubernetes_models.rb
aamederen: included

model/kubernetes/kubernetes_cluster.rb
aamederen: removeed

model/kubernetes/kubernetes_cluster.rb
aamederen: thanks!

aamederen:
> we probably want a kubernetes service project instead of squirting managed stuff into the customer's account (that's the status quo, right?). I realize that makes VM & subnet pages, etc invisible normally, and we may need to do something to make them visible...but not interact-able in every single possible way (as a managed service).
It otherwise seems like too much a burden: every time we add a feature to the compute/network/whatever layers, every single thing must compose with how the manage service works. seems a bit much?

@fdr I agree, though I intended to postpone it to the "Beta" stage. Maybe I can squeeze it in a separate PR.

prog/kubernetes/provision_kubernetes_node.rb
aamederen: I'm inclined towards postponing this decision to post-alpha as we have a few decisions to make
1. What will the Clover names of CP VMs be?
2. What about the worker VMs?
3. We also have cluster names of the nodes. Currently it's the VM inhost name.

prog/kubernetes/provision_kubernetes_node.rb
aamederen: so, what's the recommended change for this now? shellescape?

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+  sudo touch /etc/containerd/config.toml
+  containerd config default | sudo tee /etc/containerd/config.toml
+  sudo sed -i 's/SystemdCgroup = false/SystemdCgroup = true/' /etc/containerd/config.toml
+  sudo systemctl restart containerd
+  curl -fsSL https://pkgs.k8s.io/core:/stable:/#{kubernetes_cluster.kubernetes_version}/deb/Release.key | sudo gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
+  echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/#{kubernetes_cluster.kubernetes_version}/deb/ /' | sudo tee /etc/apt/sources.list.d/kubernetes.list
+  sudo apt update
+  sudo apt install -y kubelet kubeadm kubectl
+  sudo apt-mark hold kubelet kubeadm kubectl
+  sudo systemctl enable --now kubelet
```
furkansahin, line 67: I feel like it is better to move this into dataplane and use daemonizer to setup. Also, please make sure this is an idempotent state. A simple network connection issue may leave the strand infinitely running. Considering you `push` this prog, clean-up is a very hairy manual labor.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    nap 65536
+  end
+
+  label def join_control_plane
+    case vm.sshable.cmd("common/bin/daemonizer --check join_control_plane")
+    when "Succeeded"
+      pop vm_id: vm.id
+    when "NotStarted"
+      cp_sshable = kubernetes_cluster.cp_vms.first.sshable
```
furkansahin, line 117: is it only required to write this new config into a single cp vm? What if this array returns the newly added VM, does it cause any issues?

prog/kubernetes/provision_kubernetes_node.rb
aamederen: This is a temporary step that makes provisioning less resilient and more time-consuming. We are going to create VM images for k8s nodes but IMO this is OK for now to save time to start getting some feedback.

prog/kubernetes/provision_kubernetes_node.rb
aamederen: cp_vms are returned sorder by `created_at` so the `cp_vms.first` should already be a valid CP node. Running these commands against any CP node that's joined the cluster would be OK, but `cp_vms.first` is the safest one.

This design is not great when you think about concurrent operations. For future, instead of associating vms with clusters directly, we think about introducing a wrapper entity called KubernetesNode, that would act as a synchronization point as well.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+      private_subnet_id: kubernetes_cluster.private_subnet_id,
+      enable_ip4: true
+    )
+
+    current_frame = strand.stack.first
+    current_frame["vm_id"] = vm_st.id
+    strand.modified!(:stack)
+
+    kubernetes_cluster.add_cp_vm(vm_st.subject)
+    kubernetes_cluster.api_server_lb.add_vm(vm_st.subject)
```
enescakir, line 38: nit: each `st.subject` runs a query
```suggestion
    vm = Prog::Vm::Nexus.assemble_with_sshable(
      "ubi",
      kubernetes_cluster.project.id,
      # we should reiterate how we name the vm. some how correlate it to the vm's info.
      name: name,
      location: kubernetes_cluster.location,
      size: "standard-2",
      boot_image: "ubuntu-jammy",
      private_subnet_id: kubernetes_cluster.private_subnet_id,
      enable_ip4: true
    ).subject

    current_frame = strand.stack.first
    current_frame["vm_id"] = vm.id
    strand.modified!(:stack)

    kubernetes_cluster.add_cp_vm(vm)
    kubernetes_cluster.api_server_lb.add_vm(vm)
```

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+    name = "#{kubernetes_cluster.name.downcase}-control-plane-#{SecureRandom.alphanumeric(5).downcase}"
+
+    vm_st = Prog::Vm::Nexus.assemble_with_sshable(
+      "ubi",
+      kubernetes_cluster.project.id,
+      # we should reiterate how we name the vm. some how correlate it to the vm's info.
+      name: name,
+      location: kubernetes_cluster.location,
+      size: "standard-2",
+      boot_image: "ubuntu-jammy",
```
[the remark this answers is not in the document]
aamederen: That makes sense. However, for the "alpha" release we are targeting, migration may not be a concern and Ubuntu24 would require another round of revalidation. Therefore, I would put it as a early low-hanging-fruit task that's outside of the scope of this PR.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+    case vm.sshable.cmd("common/bin/daemonizer --check join_control_plane")
+    when "Succeeded"
+      pop vm_id: vm.id
+    when "NotStarted"
+      cp_sshable = kubernetes_cluster.cp_vms.first.sshable
+      join_token = cp_sshable.cmd("sudo kubeadm token create --ttl 24h --usages signing,authentication").tr("\n", "")
+      certificate_key = cp_sshable.cmd("sudo kubeadm init phase upload-certs --upload-certs")[/certificate key:\n(.*)/, 1]
+      discovery_token_ca_cert_hash = cp_sshable.cmd("sudo kubeadm token create --print-join-command")[/discovery-token-ca-cert-hash (.*)/, 1]
+
+      vm.sshable.cmd("common/bin/daemonizer 'kubernetes/bin/join-control-plane-node #{kubernetes_cluster.endpoint}:443 #{join_token.shellescape} #{discovery_token_ca_cert_hash.shellescape} #{certificate_key.shellescape}' join_control_plane")
```
jeremyevans, line 122: I'm not sure if it matters in this case, but these are invalid uses of `shellescape`.  `shellescape` should be used when the output is to an unquoted shell command, not in a single quoted shell command.  Inside of a single quoted shell command, I think you can escape with `.gsub("'", "'\\\\''")`.  You could continue to use `shellescape` if you remove the single quoting and manually escape the spaces.

I'm fairly sure if there are characters that actually need to be escaped, this breaks in various ways.  Most characters that `shellescape` will escape will result in unexpected inserted `\`, and `'` will break in a worse fashion, since it will close the single quote early.

Alternatively, and I'm not sure if it is an option, but if daemonizer could accept multiple arguments instead of the single argument, then you could just drop the single quoting and keep the `shellescape` as-is.

rhizome/kubernetes/bin/init-cluster
```diff
@@ -0,0 +1,54 @@
...
+if ARGV.count != 6
+  fail "Wrong number of arguments. Expected 6, Given #{ARGV.count}"
+end
+
+cluster_name = ARGV[0]
+load_balancer_host_name = ARGV[1]
+port = ARGV[2]
+private_subnet_cidr4 = ARGV[3]
+private_subnet_cidr6 = ARGV[4]
+vm_cidr = ARGV[5]
```
jeremyevans, line 16:
```suggestion
cluster_name, load_balancer_host_name, port, private_subnet_cidr4, private_subnet_cidr6, vm_cidr = ARGV
```

Similar change in join-control-plane-node

rhizome/kubernetes/bin/join-control-plane-node
```diff
@@ -0,0 +1,15 @@
...
+if ARGV.count != 4
+  fail "Wrong number of arguments. Expected 4, Given #{ARGV.count}"
+end
+
+cluster_endpoint = ARGV[0]
+token = ARGV[1]
+discovery_token_ca_cert_hash = ARGV[2]
+certificate_key = ARGV[3]
+
+r "kubeadm join #{cluster_endpoint} --control-plane --token #{token} --discovery-token-ca-cert-hash #{discovery_token_ca_cert_hash} --certificate-key #{certificate_key}"
```
jeremyevans, line 15: I think we should change `r` so it will splat an array when calling `Open3.capture3`.  Then we could pass an array here.  That avoids the use of a shell, avoiding shell interpolation issues.  What are your thoughts on that, @fdr?

spec/model/kubernetes/kubernetes_cluster_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+require_relative "../spec_helper"
+
+RSpec.describe KubernetesCluster do
+  subject(:kc) {
+    described_class.new(
+      name: "kc-name",
+      kubernetes_version: "v1.32.0",
+      location: "hetzner-fsn1",
+      cp_node_count: 3
+    ) { _1.id = "3ebf43d3-118a-4af7-87d2-59a3b9bf9116" }
```
jeremyevans, line 12: Is setting the `id` actually needed here?  You aren't saving the object, and it doesn't look like either of the two specs would need to access the id.

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -0,0 +1,126 @@
+# frozen_string_literal: true
+
+require_relative "../../model/spec_helper"
+
+RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
+  subject(:nx) { described_class.new(Strand.new(id: "8148ebdf-66b8-8ed0-9c2f-8cfe93f5aa77")) }
+
+  let(:project) { Project.create_with_id(name: "default") }
```
jeremyevans, line 8: You can replace `create_with_id` with `create` in all cases.  `create_with_id` is only for backwards compatibility.

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -0,0 +1,126 @@
+# frozen_string_literal: true
+
+require_relative "../../model/spec_helper"
+
+RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
+  subject(:nx) { described_class.new(Strand.new(id: "8148ebdf-66b8-8ed0-9c2f-8cfe93f5aa77")) }
```
jeremyevans, line 6: Is this hardcoded id needed?  I don't see it used anywhere else.  It appears to be for a `pg` UBID, which doesn't make much sense to me considering this is for k8s and not pg.

migrate/20250110_add_kubernetes_models.rb
```diff
@@ -0,0 +1,44 @@
...
+      column :id, :uuid, primary_key: true
+      column :name, :text, null: false
+      column :cp_node_count, :integer, null: false
+      column :kubernetes_version, :text, null: false
+      column :location, :text, collate: '"C"', null: false
+      column :created_at, :timestamp, null: false, default: Sequel::CURRENT_TIMESTAMP
+
+      check { (cp_node_count =~ [1, 3]) }
```
jeremyevans, line 13:
```suggestion
      constraint(:check_cp_node_count, cp_node_count: [1, 3])
```

First, it's best to always give a constraint a name so it can be deterministically dropped later.  Second, I don't think the virtual row block is needed here, so switch to the simpler hash format.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+    hop_join_control_plane
+  end
+
+  label def init_cluster
+    case vm.sshable.cmd("common/bin/daemonizer --check init_kubernetes_cluster")
+    when "Succeeded"
+      pop vm_id: vm.id
+    when "NotStarted"
+      ps = kubernetes_cluster.private_subnet
+      vm.sshable.cmd("common/bin/daemonizer '/home/ubi/kubernetes/bin/init-cluster #{kubernetes_cluster.name} #{kubernetes_cluster.endpoint} 443 #{ps.net4} #{ps.net6} #{vm.nics.first.private_ipv4}' init_kubernetes_cluster")
```
jeremyevans, line 99: This is similar to the daemonizer call below, but doesn't use `shellescape`.  It looks safe though, as I don't think any of the interpolated values can contain `'`.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+    nap 65536
+  end
+
+  label def join_control_plane
+    case vm.sshable.cmd("common/bin/daemonizer --check join_control_plane")
+    when "Succeeded"
+      pop vm_id: vm.id
+    when "NotStarted"
+      cp_sshable = kubernetes_cluster.cp_vms.first.sshable
+      join_token = cp_sshable.cmd("sudo kubeadm token create --ttl 24h --usages signing,authentication").tr("\n", "")
```
jeremyevans, line 118: Should there be multiple newlines in the output?  If there should only be the trailing newline, use `chomp` instead of `tr("\n", "")`

jeremyevans (commented):
I don't know anything about k8s, so my review is just from a general ruby/db perspective.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -0,0 +1,54 @@
+#  frozen_string_literal: true
+
+require_relative "../../model"
+
+class KubernetesCluster < Sequel::Model
+  one_to_one :strand, key: :id
+  one_to_one :api_server_lb, class: :LoadBalancer, key: :id, primary_key: :api_server_lb_id
```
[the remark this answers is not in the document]
jeremyevans: Actually, I think the issue is this should be `many_to_one` not `one_to_one`.  `one_to_one` is when the foreign key is the other table, referencing this table's primary key.  `many_to_one` is when the foreign key is in the current table, referencing the other table. See https://sequel.jeremyevans.net/rdoc/files/doc/association_basics_rdoc.html#label-The+Types+of+Associations

```suggestion
  many_to_one :api_server_lb, class: :LoadBalancer
```

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -0,0 +1,84 @@
...
+class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
+  subject_is :kubernetes_cluster
+
+  def self.assemble(name:, kubernetes_version:, private_subnet_id:, project_id:, location:, cp_node_count: 3)
+    DB.transaction do
+      unless (project = Project[project_id])
+        fail "No existing project"
+      end
+
+      unless ["v1.32", "v1.31"].include?(kubernetes_version)
```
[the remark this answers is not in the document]
aamederen: I'm not sure if we want to validate the model itself. We def. want to limit the versions selectable by the UI/API but for the operations that could be done via pry console, we may want to leave some flexibility.

A similar subject is PG version and there, we went further and added this as an enum on the DB, so would it be an even preferable option?

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
+
+class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
+  subject_is :kubernetes_cluster
+
+  def vm
+    @vm ||= Vm[frame.fetch("vm_id", nil)] || nil
```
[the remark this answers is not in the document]
aamederen: moved from https://github.com/ubicloud/ubicloud/pull/2621/commits/ab2b98a61f17b4b49e1995770e25523f347f0856

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -0,0 +1,134 @@
...
+  def write_hosts_file_if_needed(ip = nil)
+    return unless Config.development?
+    return if vm.sshable.cmd("sudo cat /etc/hosts").match?(/#{kubernetes_cluster.endpoint}/)
+    ip = kubernetes_cluster.cp_vms.first.ephemeral_net4 if ip.nil?
+
+    vm.sshable.cmd("echo '#{ip} #{kubernetes_cluster.endpoint}' | sudo tee -a /etc/hosts")
+  end
+
+  label def start
+    name = "#{kubernetes_cluster.name.downcase}-control-plane-#{SecureRandom.alphanumeric(5).downcase}"
```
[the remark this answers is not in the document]
aamederen: The naming is subject to change, but applied the name limit for safety anyway

prog/kubernetes/provision_kubernetes_node.rb
aamederen: shellescape seems to be working for now. We also had a discussion about this above: https://github.com/ubicloud/ubicloud/pull/2598#discussion_r1914948027

Overall, this seems like a smell of putting too much args into a script. Within the code, I think a better use is to dump a json like this one: https://github.com/ubicloud/ubicloud/blob/4766f573154fed17268ab92a3e4a9aa8969b90b1/prog/ai/inference_endpoint_replica_nexus.rb#L91

migrate/20250110_add_kubernetes_models.rb
aamederen: Actually, let me remove this constraint from the DB. IMO, it's a best practice to have odd number of CP nodes and having a DB constraint for that feels a bit too much.

prog/kubernetes/kubernetes_cluster_nexus.rb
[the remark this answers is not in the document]
aamederen: I kept the check in nexus and converted the column type to enum.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  label def start
+    register_deadline("wait", 120 * 60)
+    hop_create_load_balancer
+  end
+
+  label def create_load_balancer
+    load_balancer_st = Prog::Vnet::LoadBalancerNexus.assemble(
+      kubernetes_cluster.private_subnet_id,
+      name: "#{kubernetes_cluster.name}-apiserver",
```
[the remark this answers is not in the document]
aamederen: put a validation in assemble to cap it at 40 chars.

prog/kubernetes/kubernetes_cluster_nexus.rb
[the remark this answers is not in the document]
jeremyevans: I advise against using enums in PostgreSQL, as it results in irreversible migrations later if you want to add values to the enum.  I recommend using static lookup table (id, name) and a foreign key to that table.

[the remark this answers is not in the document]
aamederen: Alright. Reverted the enum and put the validation into #validate now.

spec/lib/validation_spec.rb
```diff
@@ -459,4 +459,49 @@
...
+  end
+
+  describe "#validate_kubernetes_cp_node_count" do
+    it "valid numers" do
+      [1, 3].each do |count|
+        expect(described_class.validate_kubernetes_cp_node_count(count)).to be_nil
+      end
+    end
+
+    it "invalid numers" do
```
fdr, line 501:
```suggestion
    it "valid numbers" do
      [1, 3].each do |count|
        expect(described_class.validate_kubernetes_cp_node_count(count)).to be_nil
      end
    end

    it "invalid numbers" do
```
?

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  label def wait
+    nap 65536
+  end
+
+  label def destroy
+    kubernetes_cluster.api_server_lb.incr_destroy
+    kubernetes_cluster.cp_vms.each(&:incr_destroy)
+    kubernetes_cluster.remove_all_cp_vms
```
fdr, line 83: is this necessary/desirable? it seems like cp_vms will disassociate themselves by virtue of their incr_destroy, when working properly. I guess the question is: do we want an unbounded, if theoretically transitory, leak?

fdr (approved):
seems like you have some squashing to do in a migration, make sure to do that, but no major objections, let's get this integrated.

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: >is this necessary/desirable

Someone will need to drop that association, whether the Kubernetes object or the VM itself. As far as I can see, we drop the association on the owning service side (eg. PG, AI, etc.). Not explicitly like this, but destroying the owning object (a postgres_server for example) right after vm.incr_destroy.

> it seems like cp_vms will disassociate themselves by virtue of their incr_destroy, when working properly. 

As far as I understand, the VM destroy process won't auto-disassociate themselves unless we associate VMs with their clusters or nodepools and use association_dependencies plugin. This feels less desirable to me.

> I guess the question is: do we want an unbounded, if theoretically transitory, leak?

IMO, leak is not an issue as long as we incr_destroy it.

Well, overall, in the not-so-long term, we may put a middle-man model between a cluster and a VM, something like a KubernetesNode. That would improve the association problems too.

PR #2602, opened 2025-01-15
Add a unique index on project_id and name to 5 tables

jeremyevans:
These tables previously used the unique index in access_tag to enforce a unique name per project.  When we stop using access_tag for these, we still need the unique name per project enforced, so these add the necessary indexes.

There are 5 additional tables that have project_id and name columns, but I don't believe they are created by users, so I'm not sure if a (project_id, name) unique index is appropriate (maybe only unique name index is appropriate, or something else):

* dns_zone
* github_installation
* inference_endpoint
* minio_cluster
* usage_alert

I noticed what I think is a bug, in that postgres_resource.name was unique by itself.  This means that two separate projects/users cannot create PostgreSQL databases with the same name.  I expect that was not desired, and stems from when postgres_resource was originally named postgres_server, so this drops the unique constraint on name when adding the unique index on (project_id, name).

Additionally, the migration to add the project_id columns did not make each column a foreign key, so this makes each project_id column a foreign key referencing project.


byucesoy (commented):
At the moment, users cannot create DnsZone, but they would be able at some point when we have DnsZone product and at that time, I think we should allow multiple DnsZone's with same name. So not putting unique key is OK for now and also OK for future.

Usage alert is created by users (from the billing page). I think it makes sense to add unique key for that as the primary purpose of the name in usage alert is giving way to distinguish different alerts. Multiple alerts with the same name would defeat the purpose.

I don't have strong opinion for others.

enescakir:
Names should be unique by project and location for location based resources. Our goal was to ensure the hyper_tag_name is unique for each resource
 
CleanShot 2025-01-15 at 16 46 10@2x

jeremyevans:
> Names should be unique by project and location for location based resources. Our goal was to ensure the hyper_tag_name is unique for each resource

Looking at load_balancer, this isn't currently possible, because it gets the location from the private subnet.  Should the load balancer constraint be `(project_id, name)` or `(project_id, private_subnet_id, name)`, or should we add a location column to load_balancer and do `(project_id, location, name)` ?

For the others, adding a unique index for `(project_id, location, name)` works fine.

jeremyevans:
I've pushed a new version that uses `(project_id, location, name)` where possible.  For load_balancer, it uses `(private_subnet_id, name)`.  If we want to add a `location` column to load_balancer, we can use `(project_id, location, name)` there.

I've also added `(project_id, name)` indexes for dns_zone and usage_alert.

migrate/20250115_project_id_name_indexes.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
+
+  up do
+    %i[postgres_resource firewall private_subnet vm].each do |table|
```
enescakir, line 7: Why isn't `minio_cluster` included in this list?

jeremyevans: No good reason.  I'll add it.

enescakir (approved):

PR #2603, opened 2025-01-15
Mark Finland as not visible

jeremyevans:
We plan on dropping support for the Helsinki region, so the first step is to stop advertising it as an option for new VMs.


byucesoy (approved):

enescakir (changes requested):
I think we keep Finland for standard vms with a few hosts. We will move 90% of its capacity to Germany. Lets asked to Ozgun.

jeremyevans:
Closing as we will continue to accept virtual machines in Finland.

PR #2612, opened 2025-01-15
Add E2E tests for PostgreSQL

shikharbhardwaj:
## Description

* Adds E2E tests for Postgres

## Tests
- [x] Unit tests
- [x] E2E tests run in GH CI from this branch


config/postgres_e2e_tests.yml
```diff
@@ -0,0 +1,5 @@
+- {name: "postgres16-standard-noha", postgres_version: "16", postgres_flavor: "standard", ha_type: "none"}
```
byucesoy, line 1: For the beginning, we can start with smaller test matrix. Just standard flavor is enough. We can add more later on.

prog/test/postgres_resource.rb
```diff
@@ -0,0 +1,152 @@
...
+
+  label def start
+    hop_download_boot_image
+  end
+
+  label def download_boot_image
+    flavor = "-#{tests[test_case]["postgres_flavor"]}"
+    flavor = "" if flavor == "-standard"
+    image_name = "postgres#{tests[test_case]["postgres_version"]}#{flavor}-ubuntu-2204"
+    vm_host = VmHost[vm_host_id]
```
byucesoy, line 28: I don't think we should pass around `vm_host_id`, just to use it to download boot image. There is already only one host in the E2E tests, we can directly download the boot image there. And even if we add more hosts, it is OK to download boot image to all of them.

I check the GithubRunner tests. It seems we also pass `vm_host_id` there. @enescakir, @velioglu any reason we decided to do it like that?

byucesoy: Actually, why don't we download all necessary images while setting up the host?

prog/test/testdata/order_analytics_data.sql
```diff
@@ -0,0 +1,14 @@
+-- Insert representative test data
```
byucesoy, line 1: Qq how did you generate these data set and the queries?

byucesoy (commented):
I understand that you are using GithubRunner tests as baseline, which would be a good strategy normally. However, for E2E tests, I think we should take a different approach. Here is why:

GithubRunner tests are about testing same thing with different configurations (i.e. different base image). Postgres tests would be about testing different things with one (or few) configurations. When we add HA, backup restore, scale up/down, etc. tests, this matrix approach wouldn't work well IMO.

My suggestion would be;
- Moving image download to host setup
- Getting rid of matrix.
- Merging two progs to have one prog that performs basic postgres e2e tests.

In the future when we add HA or backup/restore test, we can create another prog. Those tests would be pretty different than basic test and it would be difficult to reuse the second prog you are adding here. I'm sure there would be some common steps. We can extract them at that time. I don't think we should optimize for that at the moment though.

prog/test/postgres_resource.rb
shikharbhardwaj: @byucesoy do you mean download it in `bin/ci`after the `hetzner_server_st` is ready or within `prog/test/hetzner_server.rb` itself, since it is only used for E2E tests?

prog/test/testdata/order_analytics_data.sql
shikharbhardwaj: I used greptile (Claude) to generate these

prog/test/postgres_resource.rb
enescakir: I can't recall the exact reason right now. @velioglu chose to download the image during the test, rather than when setting up the host. He might know more about this.

byucesoy: I mean in `prog/test/hetzner_server.rb`

velioglu: Since downloading runner images takes few minutes, I downloaded it via test code to not download them if runner tests are not running. Do the same for specific host in case there will be more than 1 host in the future. If we will download them conditionally while setting up the host it is ok.

config/postgres_e2e_tests.yml
shikharbhardwaj: I've reduced this to only test for `standard`, Postgres 17

prog/test/hetzner_server.rb
```diff
@@ -99,7 +99,21 @@ def verify_specs_installation(installed: true)
     vm_host.sshable.cmd("sudo chmod a+rw #{tmp_dir}")
     vm_host.sshable.cmd("sudo RUN_E2E_TESTS=1 SPDK_TESTS_TMP_DIR=#{tmp_dir} bundle exec rspec host/e2e")
 
-    hop_wait
+    hop_download_boot_images
+  end
+
+  label def download_boot_images
```
enescakir, line 105: `Prog::Vm::HostNexus.assemble` has `default_boot_images` parameter, if you pass it, it will download images automatically.

prog/test/hetzner_server.rb
```diff
@@ -145,4 +159,8 @@ def hetzner_api
   def vm_host
     @vm_host ||= VmHost[frame["vm_host_id"]]
   end
+
+  def default_boot_images
+    @default_boot_images ||= YAML.load_file("config/e2e_boot_images.yml")
```
enescakir, line 164: We can determine which test cases to run, so we may not need all the images. Rather than using a static file, we can identify the necessary images based on the `options[:test_cases]` value and download only those.

prog/test/hetzner_server.rb
shikharbhardwaj: :ack: updated this

prog/test/hetzner_server.rb
shikharbhardwaj: Makes sense, added this

prog/test/postgres_resource.rb
shikharbhardwaj: I have added a new config called `e2e_boot_images.yml` which lists the boot images for different tests cases. In `./bin/ci`, this file is read and we pass whichever boot images are required for the list of test cases that are being run to `default_boot_images` arg in `Prog::Vm::HostNexus.assemble`.

prog/test/postgres_resource.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+
+require_relative "../../lib/util"
+
+class Prog::Test::PostgresResource < Prog::Test::Base
+  semaphore :destroy
+
+  def self.assemble
+    postgres_service_project = Project.create(name: "Postgres-Service-Project") { _1.id = Config.postgres_service_project_id }
+    postgres_service_project.associate_with_project(postgres_service_project)
```
enescakir, line 10: If Jeremy's PR https://github.com/ubicloud/ubicloud/pull/2616 merged first, we will not have `associate_with_project`, jfyi

prog/test/postgres_resource.rb
byucesoy: 99% of the time, we run all the tests, right? I think it would be better to download all the necessary images all the time for the sake of simplicity (no new configs, no extra logic to fetch additional boot images).

We don't need to refactor runner tests I think (I'd prefer to, but that would require buy-in from runner team and also I don't want to increase the scope of this PR), but for Postgres, I think we can download the image all the time.

prog/test/postgres_resource.rb
```diff
@@ -0,0 +1,132 @@
...
+  label def wait_postgres_resource
+    if postgres_resource.strand.label == "wait" &&
+        postgres_resource.representative_server.run_query("SELECT 1") == "1"
+      hop_test_basic_connectivity
+    else
+      nap 10
+    end
+  end
+
+  label def test_basic_connectivity
```
byucesoy, line 54: Just one label that tests everything is enough IMO. I think in a regular scenario, either all these test fail or all these tests pass. We don't gain any extra information with separate tests.

Let's have one label, that creates tables, loads data and run queries. In fact, we can have one sql file that does all these and we can only run that.

byucesoy (commented):
I think high level logic looks good. I added few comments. General theme in my comments is simplification. As a general philosophy, we value simplicity a lot in our team.

.github/workflows/e2e.yml
```diff
@@ -110,6 +110,7 @@ jobs:
         GITHUB_CACHE_BLOB_STORAGE_SECRET_KEY: ${{ secrets.GH_CACHE_BLOB_STORAGE_SECRET_KEY }}
         GITHUB_CACHE_BLOB_STORAGE_ACCOUNT_ID: ${{ secrets.GH_CACHE_BLOB_STORAGE_ACCOUNT_ID }}
         GITHUB_CACHE_BLOB_STORAGE_API_KEY: ${{ secrets.GH_CACHE_BLOB_STORAGE_API_KEY }}
+        POSTGRES_SERVICE_PROJECT_ID: "546a1ed8-53e5-86d2-966c-fb782d2ae5ab"
```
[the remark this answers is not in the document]
enescakir: Actually, I said to him hardcoded is okay.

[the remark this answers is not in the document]
enescakir: Why do you think we should put an arbitrary UUID to the secrets?

prog/test/postgres_resource.rb
enescakir: Merged

prog/test/postgres_resource.rb
byucesoy: We had a sync on this with @enescakir. Let's leave this as is. He will open a follow up PR that combines logic for all E2E tests.

.github/workflows/e2e.yml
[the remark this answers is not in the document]
byucesoy: It is weird to put arbitrary UUID to anywhere TBH, secrets or configs. I suggested secrets because we do the same for `GH_RUNNER_SERVICE_PROJECT_ID`.

Probably better way would be using something like `dev_project` for all tests, which would save us from any kind of hardcoding.

prog/test/postgres_resource.rb
enescakir: I created a draft PR for it.
https://github.com/ubicloud/ubicloud/pull/2642

prog/test/postgres_resource.rb
shikharbhardwaj: :ack: updated this

.github/workflows/e2e.yml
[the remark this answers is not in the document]
shikharbhardwaj: Chatted with @byucesoy on this, will leave this in the yaml
We can circle back to clean these up for all E2E tests in a separate change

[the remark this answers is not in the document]
enescakir: IMHO, we can also move the other service project IDs away from secrets and eventually use the `dev_project`.

prog/test/postgres_resource.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  label def destroy
+    postgres_service_project.destroy
+    postgres_test_project.destroy
+
+    frame["fail_message"] ? fail_test(frame["fail_message"]) : hop_finish
+  end
+
+  label def finish
```
byucesoy, line 64: No need for separate label for this, you can directly pop in the previous label.

byucesoy (approved):

PR #2616, opened 2025-01-16
Next steps in the switch from access_tag table to project_id columns

jeremyevans:
This builds on top of #2602.  After we have fully populated the newly added `project_id` columns, we can switch to using them for authorization, and mark them as NOT NULL.

This implements steps 4-6 and 9 of the 12-step plan discussed in #2594.  The remaining steps in the plan after this PR are just cleaning up unnecessary data.

### DIFF
--- a/lib/authorization.rb
+++ b/lib/authorization.rb
@@ -110,16 +110,6 @@ module Authorization
           .where { level < Config.recursive_tag_limit },
         args: [:object_id, :level]).select(:object_id)
 
-    # If the model includes HyperTagMethods, look in access_tag.
-    # This is a temporary change while we populate the project_id
-    # columns for the related models.  After population, only
-    # the Account model will use access_tag.
-    project_id_match = if dataset.model < HyperTagMethods
-      DB[:access_tag].select(:project_id).where(hyper_tag_id: Sequel[from][:id])
-    else
-      Sequel[from][:project_id]
-    end
-
     if dataset.model == ObjectTag
       # Authorization for accessing ObjectTag itself is done by providing the metatag for the object.
       # Convert metatag id returned from the applied_object_tag lookup into tag ids, so that the correct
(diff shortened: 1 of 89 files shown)


lib/authorization.rb
```diff
@@ -79,15 +79,27 @@ def self.matched_policies_dataset(project_id, subject_id, actions = nil, object_
         object_id = UBID.parse(object_id).to_uuid
       end
 
-      if (klass = UBID.class_for_ubid(ubid))
+      klass = UBID.class_for_ubid(ubid)
+      klass = ApiKey if ubid.start_with?("et")
```
enescakir, line 83: I recall seeing multiple special checks for `ApiKey`'s UBID while reviewing your authorization v2 PRs. Should we consider using a distinct UBID type for the ApiKey instead of etc, to eliminate these specific checks?

lib/authorization.rb
```diff
@@ -143,58 +145,12 @@ def self.dataset_authorize(dataset, project_id, subject_id, actions)
...
-
-      AccessTag.create_with_id(
-        project_id: project.id,
-        name: hyper_tag_name(project),
-        hyper_tag_id: id,
-        hyper_tag_table: self.class.table_name
-      )
-    end
-
     def before_destroy
```
enescakir, line 154: We'll get rid of it once we've deleted all the non-account tags from the `access_tag` table, correct?

model/account.rb
```diff
@@ -37,6 +36,34 @@ def create_project_with_default_policy(name, default_policy: true)
...
+  def associate_with_project(project)
+    return if project.nil?
+
+    # Temporary code to ensure that project_id column already set,
+    # except for Account which doesn't have the column
+    # :nocov:
+    case self
+    when Account
+      nil
+    when Project
```
enescakir, line 52: Since we moved it to the `Account` model, self can't be `Project`

lib/authorization.rb
jeremyevans: The issue there is all current ApiKey values use the ETC ubid type.  We could make new ApiKey values use a different ubid type, but unless we want to invalidate the current ones, we need to keep this support.

lib/authorization.rb
jeremyevans: Yes, this is only temporary.

model/account.rb
jeremyevans: Good point, I'll fix that.

jeremyevans:
@enescakir I pushed only the first commit to this PR, since that one should be deployed first.  This includes steps 5-6 of the plan in #2594.

The other commits in a previous version of this PR, which included steps 4 and 9 of the plan, are now included in #2624.

enescakir (approved):

lib/authorization.rb
jeremyevans: That approach would work.

PR #2620, opened 2025-01-16
Remove version from demo docker-compose template

enescakir:
It shows the following warning:

    WARN[0000] ./ubicloud/demo/docker-compose.yml: the attribute
    `version` is obsolete, it will be ignored, please remove it to avoid
    potential confusion


enescakir:
https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

PR #2617, opened 2025-01-16
[k8s] Introduce Ubicni for establishing intra-cluster network connectivity

aamederen:
Ubicni is our CNI plugin for kubernetes that allows intra-cluster network communication within a kubernetes cluster on top of Ubicloud network stack.

PR #2621, opened 2025-01-16
[k8s] Introduce Kubernetes Nodepools

aamederen:
This change adds the concept of Kubernetes Nodepools.

Each kubernetes nodepool is a set of VMs that act as the nodes that run the actual application workloads.

A cluster have multiple nodepools with different size and characteristics. The logic in this PR is also implemented this way. However, product-wise, we may hide this feature for now and default to a single nodepool.

A kubernetes cluster without a nodepool is kinda useless at this point.

Nodepools should be created separately from the cluster. However, a nodepool nexus strand would wait until its cluster is stable.

### DIFF
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -8,6 +8,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :private_subnet
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
+  one_to_many :nodepools, class: :KubernetesNodepool
 
   dataset_module Pagination
 
@@ -32,6 +33,10 @@ class KubernetesCluster < Sequel::Model
 
   def endpoint
     api_server_lb.hostname
+  end
+
+  def sshable
+    cp_vms.first.sshable
   end
 end
 
--- a/model/kubernetes/kubernetes_nodepool.rb
+++ b/model/kubernetes/kubernetes_nodepool.rb
@@ -1,0 +1,28 @@
+#  frozen_string_literal: true
+
+require_relative "../../model"
+
+class KubernetesNodepool < Sequel::Model
+  one_to_one :strand, key: :id
+  many_to_one :cluster, key: :kubernetes_cluster_id, class: :KubernetesCluster
+  many_to_many :vms, order: :created_at
+
+  include ResourceMethods
+  include SemaphoreMethods
+
+  semaphore :destroy
+end
+
+# Table: kubernetes_nodepool
+# Columns:
+#  id                    | uuid                     | PRIMARY KEY
+#  name                  | text                     | NOT NULL
+#  node_count            | integer                  | NOT NULL
+#  created_at            | timestamp with time zone | NOT NULL DEFAULT CURRENT_TIMESTAMP
+#  kubernetes_cluster_id | uuid                     | NOT NULL
+# Indexes:
+#  kubernetes_nodepool_pkey | PRIMARY KEY btree (id)
+# Foreign key constraints:
+#  kubernetes_nodepool_kubernetes_cluster_id_fkey | (kubernetes_cluster_id) REFERENCES kubernetes_cluster(id)
+# Referenced By:
+#  kubernetes_nodepools_vms | kubernetes_nodepools_vms_kubernetes_nodepool_id_fkey | (kubernetes_nodepool_id) REFERENCES kubernetes_nodepool(id)
--- a/prog/kubernetes/kubernetes_nodepool_nexus.rb
+++ b/prog/kubernetes/kubernetes_nodepool_nexus.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
+  subject_is :kubernetes_nodepool
+
+  def self.assemble(name:, node_count:, kubernetes_cluster_id:)
+    DB.transaction do
+      unless KubernetesCluster[kubernetes_cluster_id]
+        fail "No existing cluster"
+      end
+
+      kn = KubernetesNodepool.create(name:, node_count:, kubernetes_cluster_id:)
+
+      Strand.create(prog: "Kubernetes::KubernetesNodepoolNexus", label: "start") { _1.id = kn.id }
+    end
+  end
+
+  def before_run
+    when_destroy_set? do
+      if strand.label != "destroy"
+        hop_destroy
+      end
+    end
+  end
+
+  label def start
+    nap 30 unless kubernetes_nodepool.cluster.strand.label == "wait"
+    register_deadline("wait", 120 * 60)
+    hop_bootstrap_worker_vms
+  end
+
+  label def bootstrap_worker_vms
+    hop_wait if kubernetes_nodepool.vms.count >= kubernetes_nodepool.node_count
+    push Prog::Kubernetes::ProvisionKubernetesNode, {"nodepool_id" => kubernetes_nodepool.id, "subject_id" => kubernetes_nodepool.kubernetes_cluster_id}
+  end
+
+  label def wait
+    nap 30
+  end
+
+  label def destroy
+    kubernetes_nodepool.vms.each(&:incr_destroy)
+    kubernetes_nodepool.remove_all_vms
+    kubernetes_nodepool.destroy
+    pop "kubernetes nodepool is deleted"
+  end
+end
--- a/prog/kubernetes/provision_kubernetes_node.rb
+++ b/prog/kubernetes/provision_kubernetes_node.rb
@@ -7,16 +7,24 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
     @vm ||= Vm[frame["vm_id"]]
   end
 
+  def kubernetes_nodepool
+    @kubernetes_nodepool ||= KubernetesNodepool[frame["nodepool_id"]]
+  end
+
   def write_hosts_file_if_needed(ip = nil)
     return unless Config.development?
     return if vm.sshable.cmd("cat /etc/hosts").include?(kubernetes_cluster.endpoint.to_s)
-    ip ||= kubernetes_cluster.cp_vms.first.ephemeral_net4
+    ip ||= kubernetes_cluster.sshable.host
 
     vm.sshable.cmd("sudo tee -a /etc/hosts", stdin: "#{ip} #{kubernetes_cluster.endpoint}\n")
   end
 
   label def start
-    name = "#{kubernetes_cluster.name.downcase}-control-plane-#{SecureRandom.alphanumeric(5).downcase}"
+    name = if kubernetes_nodepool
+      "#{kubernetes_nodepool.name}-#{SecureRandom.alphanumeric(5).downcase}"
+    else
+      "#{kubernetes_cluster.name.downcase}-control-plane-#{SecureRandom.alphanumeric(5).downcase}"
+    end
 
     vm = Prog::Vm::Nexus.assemble_with_sshable(
       "ubi",
@@ -34,8 +42,12 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
     current_frame["vm_id"] = vm.id
     strand.modified!(:stack)
 
-    kubernetes_cluster.add_cp_vm(vm)
-    kubernetes_cluster.api_server_lb.add_vm(vm)
+    if kubernetes_nodepool
+      kubernetes_nodepool.add_vm(vm)
+    else
+      kubernetes_cluster.add_cp_vm(vm)
+      kubernetes_cluster.api_server_lb.add_vm(vm)
+    end
 
     hop_install_software
   end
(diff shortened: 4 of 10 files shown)


prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -0,0 +1,51 @@
...
+    push Prog::Kubernetes::ProvisionKubernetesNode, {"nodepool_id" => kubernetes_nodepool.id, "subject_id" => kubernetes_nodepool.kubernetes_cluster.id}
+  end
+
+  label def wait
+    nap 30
+  end
+
+  label def destroy
+    kubernetes_nodepool.vms.each(&:incr_destroy)
+    kubernetes_nodepool.vms.each { kubernetes_nodepool.remove_vm(_1) }
```
enescakir, line 47: It also adds `remove_all_*` helper
https://github.com/ubicloud/ubicloud/pull/2598#discussion_r1919142278
```suggestion
    kubernetes_nodepool.remove_all_vms
```

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -8,6 +8,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :private_subnet
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
+  one_to_many :kubernetes_nodepools
```
enescakir, line 11: Since both of them are in the `Kubernetes` context, `nodepools` might be more readable. 
I prefer `kubernetes_cluster.nodepools` instead of `kubernetes_cluster.kubernetes_nodepools`

similar discussion: https://github.com/ubicloud/ubicloud/pull/2628#discussion_r1920043782

model/kubernetes/kubernetes_nodepool.rb
```diff
@@ -0,0 +1,28 @@
+#  frozen_string_literal: true
+
+require_relative "../../model"
+
+class KubernetesNodepool < Sequel::Model
+  one_to_one :strand, key: :id
+  many_to_one :kubernetes_cluster
```
enescakir, line 7: Since both of them are in the `Kubernetes` context, `cluster` might be more readable. 
I prefer `kubernetes_nodepool.cluster` instead of `kubernetes_nodepool.kubernetes_cluster`

similar discussion: https://github.com/ubicloud/ubicloud/pull/2628#discussion_r1920043782

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -0,0 +1,51 @@
...
+    DB.transaction do
+      unless KubernetesCluster[kubernetes_cluster_id]
+        fail "No existing cluster"
+      end
+
+      kn = KubernetesNodepool.create(
+        name: name,
+        node_count: node_count,
+        kubernetes_cluster_id: kubernetes_cluster_id
+      )
```
enescakir, line 16: nit:
```suggestion
      kn = KubernetesNodepool.create(name:, node_count:, kubernetes_cluster_id:)
```

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -7,6 +7,10 @@ def vm
     @vm ||= Vm[frame.fetch("vm_id", nil)] || nil
   end
 
+  def kubernetes_nodepool
+    @kubernetes_nodepool ||= KubernetesNodepool[frame.fetch("nodepool_id", nil)] || nil
```
enescakir, line 11: Since you give `nil` as default value

```suggestion
    @kubernetes_nodepool ||= KubernetesNodepool[frame["nodepool_id"]] || nil
```

enescakir, line 11: Why do we have `|| nil` part 🤔

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -34,8 +42,12 @@ def write_hosts_file_if_needed(ip = nil)
...
     strand.modified!(:stack)
 
-    kubernetes_cluster.add_cp_vm(vm_st.subject)
-    kubernetes_cluster.api_server_lb.add_vm(vm_st.subject)
+    if kubernetes_nodepool
+      kubernetes_nodepool.add_vm(vm_st.subject)
+    else
+      kubernetes_cluster.add_cp_vm(vm_st.subject)
+      kubernetes_cluster.api_server_lb.add_vm(vm_st.subject)
+    end
```
enescakir, line 50: https://github.com/ubicloud/ubicloud/pull/2598#discussion_r1925373400

prog/kubernetes/kubernetes_nodepool_nexus.rb
aamederen: This is not one of my favorite features, but not opposed to using it.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -8,6 +8,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :private_subnet
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
+  one_to_many :nodepools, key: :kubernetes_cluster_id, class: :KubernetesNodepool
```
jeremyevans, line 11: Is `key` needed here?  The given value looks the same as what Sequel would use by default.

jeremyevans (commented):
Only one specific comment, but some of my general comments in #2598 and #2617 also apply to this.

rhizome/kubernetes/bin/ubicni
```diff
@@ -0,0 +1,295 @@
...
+require "json"
+require "logger"
+require "ipaddr"
+require "fileutils"
+require "securerandom"
+
+LOG = Logger.new("/opt/cni/bin/cni_plugin.log")
+LOG.level = Logger::INFO
+
+class UBICNI
```
fdr, line 14: weird that it's all upper case I think, more like UbiCNI right?

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -169,7 +166,12 @@ def write_hosts_file_if_needed(ip = nil)
     sshable.cmd("sudo tee /etc/cni/net.d/ubicni-config.json", stdin: cni_config)
 
     sshable.cmd("sudo chmod +x /opt/cni/bin/ubicni")
-    sshable.cmd("sudo iptables -t nat -A POSTROUTING -s #{vm.nics.first.private_ipv4} -o ens3 -j MASQUERADE")
+    vm.sshable.cmd <<~BASH
+      set -ueo pipefail
+      sudo nft add table ip nat
+      sudo nft add chain ip nat POSTROUTING { type nat hook postrouting priority 100';' }
+      sudo nft add rule ip nat POSTROUTING ip saddr #{vm.nics.first.private_ipv4} oifname ens3 masquerade
```
fdr, line 212: I suspect you really want to be using `nft -f` to get it done in one shot. 

```
flush table ip nat

table ip nat {
  chain POSTROUTING {
    type nat hook postrouting priority 100; 
    policy accept;
    ip saddr #{vm.nics.first.private_ipv4} oifname ens3 masquerade;
  }
}
```

You can pass it via stdin. The initial flush is to deal with repeat executions (e.g. from connection reset by peer or other error causing a retry), which tend to catenate directives, not upsert/replace them.

rhizome/kubernetes/bin/ubicni
```diff
@@ -69,11 +69,11 @@ class UBICNI
     container_id = ENV["CNI_CONTAINERID"]
     cni_netns = ENV["CNI_NETNS"].sub("/var/run/netns/", "")
 
-    inner_mac = gen_mac.shellescape
+    inner_mac = gen_mac
     inner_link_local = mac_to_ipv6_link_local(inner_mac)
     inner_ifname = ENV["CNI_IFNAME"]
 
-    outer_mac = gen_mac.shellescape
```
fdr: given cni is not committed seems like this is squashable no?

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -134,7 +148,32 @@ def write_hosts_file_if_needed(ip = nil)
       nap 10
     when "Failed"
       # TODO: Create a page
-      Clog.emit("JOIN CLUSTER FAILED")
+      Clog.emit("JOIN CP NODE TO CLUSTER FAILED")
```
fdr, line 151: all caps is kinda weird. also, would you want there to be anything as structured data, which is a hash returned from a block?

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -134,7 +148,32 @@ def write_hosts_file_if_needed(ip = nil)
...
+
+    nap 65536
+  end
+
+  label def join_worker
+    case vm.sshable.cmd("common/bin/daemonizer --check join_worker")
+    when "Succeeded"
+      hop_install_cni
+    when "NotStarted"
+      cp_sshable = kubernetes_cluster.cp_vms.first.sshable
```
fdr, line 163: I do feel like I see the `cp_vms.first` formation a lot, should it have a subexpression/method?

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -17,12 +17,13 @@
...
       project_id: project.id
     )
+    KubernetesNodepool.create(name: "k8stest-np", node_count: 2, kubernetes_cluster_id: kc.id)
 
     lb = LoadBalancer.create(private_subnet_id: subnet.id, name: "somelb", src_port: 123, dst_port: 456, health_check_endpoint: "/foo", project_id: project.id)
     kc.add_cp_vm(create_vm)
     kc.add_cp_vm(create_vm)
     kc.update(api_server_lb_id: lb.id)
-    kc
+    kc.reload
```
fdr, line 26: why?

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -44,7 +44,7 @@ def before_run
 
   label def destroy
     kubernetes_nodepool.vms.each(&:incr_destroy)
-    kubernetes_nodepool.vms.each { kubernetes_nodepool.remove_vm(_1) }
+    kubernetes_nodepool.remove_all_vms
```
fdr, line 43: seems like squash material

prog/kubernetes/provision_kubernetes_node.rb
mohi-kalantari: I fixed this issue in the main CNI PR. here's the change:
https://github.com/ubicloud/ubicloud/pull/2617/files#diff-03d16198cae3ba587e3977cec63261248a6af936cec60880873b90de07e12107R158

Also I'm not flushing here because at the same time, kube proxy is working on that table and flushing it would break the networking.

since we only run it once during the bootstrap, we do not really need to care about the idempotency

aamederen: >since we only run it once during the bootstrap, we do not really need to care about the idempotency

We need to care about idempotency because it's not guaranteed for a label to run only once.

mohi-kalantari: I assumed that label runs are atomic and we only run through the install_cni once. unless someone manually changes the label to go back. I thought we are not going to care about that that much because otherwise everything including the cni installation needs to be way more defensive

mohi-kalantari: Caring about it also requires some work. maybe we can come back later but we would need to rollout kube proxy everytime we flush that table.

aamederen: You're right, everything in the setup process (init, join, etc.) are designed to be run once. The problem is not specific to this step.

I'm personally OK for the alpha stage. For beta, we can introduce node-retry on fails

prog/kubernetes/provision_kubernetes_node.rb
aamederen: I made a change, but I'm not sure if this is more desirable than `cp_vms.first`.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -134,7 +148,32 @@ def write_hosts_file_if_needed(ip = nil)
...
+      hop_install_cni
+    when "NotStarted"
+      cp_sshable = kubernetes_cluster.sshable
+      params = {
+        endpoint: "#{kubernetes_cluster.endpoint}:443",
+        join_token: cp_sshable.cmd("sudo kubeadm token create --ttl 24h --usages signing,authentication").tr("\n", ""),
+        discovery_token_ca_cert_hash: cp_sshable.cmd("sudo kubeadm token create --print-join-command")[/discovery-token-ca-cert-hash (.*)/, 1]
+      }
+
+      vm.sshable.cmd("common/bin/daemonizer kubernetes/bin/join-worker-node join_worker", stdin: JSON.generate(params))
```
mohi-kalantari, line 170: I couldn't find where we use the extra join_worker in the rhizome script? do we need it?

mohi-kalantari (approved):

prog/kubernetes/provision_kubernetes_node.rb
aamederen: Resolving this for now, not a big deal to change later anyway.

PR #2627, opened 2025-01-16
Set docker mirror client code for the runners on the test host

velioglu:
In order to test docker registry mirror performance, registry mirror server is set up on one of the host manually. Updating docker and buildkit config here to use that server for the runners provisioned on the test host.

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -223,6 +223,32 @@ class Prog::Vm::GithubRunner < Prog::Base
       UBICLOUD_CACHE_URL=#{Config.base_url}/runtime/github/" | sudo tee -a /etc/environment
     COMMAND
 
+    if (mirror_vm = Vm[Config.docker_mirror_server_vm_id]) && vm.vm_host_id == mirror_vm.vm_host_id
+      mirror_address = "#{mirror_vm.load_balancer.hostname}:5000"
+      command += <<~COMMAND
+        # Configure Docker daemon with registry mirror
+        if [ -f /etc/docker/daemon.json ] && [ -s /etc/docker/daemon.json ]; then
+          sudo jq '. + {"registry-mirrors": ["https://#{mirror_address}"]}' /etc/docker/daemon.json | sudo tee /etc/docker/daemon.json.tmp
+          sudo mv /etc/docker/daemon.json.tmp /etc/docker/daemon.json
+        else
+          echo '{"registry-mirrors": ["https://#{mirror_address}"]}' | sudo tee /etc/docker/daemon.json
+        fi
+
+        # Configure BuildKit to use the mirror
+        sudo mkdir -p /etc/buildkit
+        echo '
+          [registry."docker.io"]
+            mirrors = ["#{mirror_address}"]
+
+          [registry."#{mirror_address}"]
+            http = false
+            insecure = false' | sudo tee -a /etc/buildkit/buildkitd.toml
+
+        sudo systemctl daemon-reload
+        sudo systemctl restart docker
+      COMMAND
+    end
+
     if github_runner.installation.cache_enabled
       local_ip = vm.nics.first.private_ipv4.network.to_s
       command += <<~COMMAND
--- a/config.rb
+++ b/config.rb
@@ -108,6 +108,9 @@ module Config
   optional :github_runner_service_project_id, string
   override :enable_github_workflow_poller, true, bool
 
+  # Docker Mirror Test
+  optional :docker_mirror_server_vm_id, string
+
   # GitHub Cache
   optional :github_cache_blob_storage_endpoint, string
   optional :github_cache_blob_storage_region, string
--- a/spec/prog/vm/github_runner_spec.rb
+++ b/spec/prog/vm/github_runner_spec.rb
@@ -377,7 +377,43 @@ RSpec.describe Prog::Vm::GithubRunner do
 
   describe "#setup_environment" do
     it "hops to register_runner" do
-      expect(vm).to receive(:vm_host).and_return(instance_double(VmHost, ubid: "vhfdmbbtdz3j3h8hccf8s9wz94", location: "hetzner-fsn1", data_center: "FSN1-DC8")).at_least(:once)
+      expect(Config).to receive(:docker_mirror_server_vm_id).and_return(vm.id).at_least(:once)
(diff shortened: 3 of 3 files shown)


prog/vm/github_runner.rb
```diff
@@ -223,6 +223,39 @@ def setup_info
       UBICLOUD_CACHE_URL=#{Config.base_url}/runtime/github/" | sudo tee -a /etc/environment
     COMMAND
 
+    if vm.vm_host.id == Vm[Config.docker_mirror_server_vm_id]&.vm_host&.id
+      mirror_server_address = "mirror.ubicloudregistry.com"
```
enescakir, line 227: Why don’t we use an existing domain? Since we don't own `ubicloudregistry.com` domain, I prefer to use a subdomain from `ubicloud.com`

prog/vm/github_runner.rb
```diff
@@ -223,6 +223,39 @@ def setup_info
...
+      mirror_server_port = 5000
+      mirror_server_vm = Vm[Config.docker_mirror_server_vm_id]
+      command += <<~COMMAND
+        echo "#{mirror_server_vm.ephemeral_net4} #{mirror_server_address}" | sudo tee -a /etc/hosts
+
+        sudo mkdir -p /etc/docker/certs.d/#{mirror_server_address}:#{mirror_server_port}/
+        echo "#{Config.docker_mirror_server_vm_ca_cert}" | sudo tee /etc/docker/certs.d/#{mirror_server_address}:#{mirror_server_port}/ca.crt
```
enescakir, line 234: Why do we use a self-signed certificate?

prog/vm/github_runner.rb
```diff
@@ -223,6 +223,33 @@ def setup_info
       UBICLOUD_CACHE_URL=#{Config.base_url}/runtime/github/" | sudo tee -a /etc/environment
     COMMAND
 
+    if vm.vm_host.id == Vm[Config.docker_mirror_server_vm_id]&.vm_host&.id
+      mirror_server_vm = Vm[Config.docker_mirror_server_vm_id]
```
enescakir, line 227: small nit: 3 queries to single query

```suggestion
    if (mirror_vm = Vm[Config.docker_mirror_server_vm_id]) && vm.vm_host_id == mirror_vm.vm_host_id
```

prog/vm/github_runner.rb
```diff
@@ -223,6 +223,33 @@ def setup_info
...
+        else
+          echo '{"registry-mirrors": ["https://#{mirror_server_address}:#{mirror_server_port}"]}' | sudo tee /etc/docker/daemon.json
+        fi
+
+        # Configure BuildKit to use the mirror
+        sudo mkdir -p /etc/buildkit
+        echo '[registry."docker.io"]
+          mirrors = ["#{mirror_server_address}:#{mirror_server_port}"]
+
+          [registry."#{mirror_server_address}:#{mirror_server_port}"]
```
enescakir, line 244: nit: we can make it little bit shorter

```suggestion
      mirror_address = "#{mirror_vm.load_balancer.hostname}:#{5000}"
      command += <<~COMMAND
        # Configure Docker daemon with registry mirror
        if [ -f /etc/docker/daemon.json ] && [ -s /etc/docker/daemon.json ]; then
          jq '. + {"registry-mirrors": ["https://#{mirror_address}"]}' /etc/docker/daemon.json | sudo tee /etc/docker/daemon.json.tmp
          sudo mv /etc/docker/daemon.json.tmp /etc/docker/daemon.json
        else
          echo '{"registry-mirrors": ["https://#{mirror_address}"]}' | sudo tee /etc/docker/daemon.json
        fi

        # Configure BuildKit to use the mirror
        sudo mkdir -p /etc/buildkit
        echo '[registry."docker.io"]
          mirrors = ["#{mirror_address}"]

          [registry."#{mirror_address}"]
```

prog/vm/github_runner.rb
```diff
@@ -223,6 +223,33 @@ def setup_info
       UBICLOUD_CACHE_URL=#{Config.base_url}/runtime/github/" | sudo tee -a /etc/environment
     COMMAND
 
+    if vm.vm_host.id == Vm[Config.docker_mirror_server_vm_id]&.vm_host&.id
+      mirror_server_vm = Vm[Config.docker_mirror_server_vm_id]
+      mirror_server_address = mirror_server_vm.load_balancer.hostname
+      mirror_server_port = 5000
+      command += <<~COMMAND
```
enescakir, line 230: When we run multiline commands, we usually add `set -ueo pipefail` at the beginning of the commands. Otherwise, the commands will keep running even if one of the middle ones fails.

prog/vm/github_runner.rb
```diff
@@ -223,6 +223,33 @@ def setup_info
...
+        # Configure Docker daemon with registry mirror
+        if [ -f /etc/docker/daemon.json ] && [ -s /etc/docker/daemon.json ]; then
+          jq '. + {"registry-mirrors": ["https://#{mirror_server_address}:#{mirror_server_port}"]}' /etc/docker/daemon.json | sudo tee /etc/docker/daemon.json.tmp
```
enescakir, line 233: We run `mv` and `tee` with `sudo`, so I assumed that accessing `/etc/docker/daemon.json` requires root permissions. Should we also add `sudo` to the `jq` command? Can it access the file without it?

prog/vm/github_runner.rb
```diff
@@ -223,6 +223,33 @@ def setup_info
...
+        fi
+
+        # Configure BuildKit to use the mirror
+        sudo mkdir -p /etc/buildkit
+        echo '[registry."docker.io"]
+          mirrors = ["#{mirror_server_address}:#{mirror_server_port}"]
+
+          [registry."#{mirror_server_address}:#{mirror_server_port}"]
+          http = false
+          insecure = false' | sudo tee /etc/buildkit/buildkitd.toml
```
enescakir, line 246: Should we check if this file exists? Since it's toml, we might want to append to it instead of overwriting it if it exists

prog/vm/github_runner.rb
velioglu: As I'm appending to the command, `set -ueo pipefail` is already there.

prog/vm/github_runner.rb
velioglu: Yes `jq` can read the file but there is no harm to do it with sudo, let me change it to not confuse.

enescakir (approved):

PR #2628, opened 2025-01-17
CPU allocation and tracking for hosts

macieksarnowicz:
We add a VmHostCpu to track all cpus on a host. A few are set aside for SPDK and the rest can be assigned to a VmHostSlice to be reserved for that slice. This provides a mechanism to select and track which slice owns which cpus and to make the assignment in a transaction which ensures the thread safety. 
This functionality will be expanded upon in subsequent commits.


migrate/20250116_insert_vm_host_cpus.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
pykello, line 1: I think the migration which inserts missing vm_host_cpu records should be a separate commit & after "Create VmHostCpu records when setting up a VmHost. ".

The reason is deployment related. With the current order of commits, if a host is setup after the migration commit is deployed but before the HostNexus commit is deployed, then that host won't have VmHostCpu records.

model/vm_host.rb
```diff
@@ -14,8 +14,10 @@ class VmHost < Sequel::Model
   one_to_many :storage_devices, key: :vm_host_id
   one_to_many :pci_devices, key: :vm_host_id
   one_to_many :boot_images, key: :vm_host_id
+  one_to_many :vm_host_slices, key: :vm_host_id
+  one_to_many :vm_host_cpus, key: :vm_host_id
```
enescakir, line 18: Since we will access them through the VM host model, like `vm_host.vm_host_cpus`, naming the relationships `slices` and `cpus` makes it more readable.

```suggestion
  one_to_many :slices, key: :vm_host_id, class: :VmHostSlice
  one_to_many :cpus, key: :vm_host_id, class: :VmHostCpu
```

migrate/20250116_insert_vm_host_cpus.rb
macieksarnowicz: OK, I will move it to a separate commit.

model/vm_host.rb
macieksarnowicz: I kind of assumed that the convention was to name the relation the same as the model/table it points to. If it is Ok to diverge from that convention, I am ok to change it. I have not seen an example in the codebase like what you suggest. However, I agree with you that it is more readable.

enescakir: Here's several examples:
- https://github.com/ubicloud/ubicloud/blob/3f0a635210d41b3e1c83375eed31652efdb53e5a/model/postgres/postgres_resource.rb#L10
- https://github.com/ubicloud/ubicloud/blob/3f0a635210d41b3e1c83375eed31652efdb53e5a/model/minio/minio_server.rb#L11
- https://github.com/ubicloud/ubicloud/blob/3f0a635210d41b3e1c83375eed31652efdb53e5a/model/github/github_repository.rb#L9

macieksarnowicz: OK, thank you. I can make that change.

jeremyevans (commented):
No issues from ruby perspective, beyond what other reviewed mentioned.

pykello (approved):

PR #2634, opened 2025-01-17
Set timeout of 5 seconds for the curl command used to communicate with inference gateway

bsatzger:
**Set max-time for curl command**

Set max-time to 5 seconds for the curl command that is used by clover to
communicate with and configure the inference gateway.

```
-m, --max-time <seconds>
    Maximum  time  in  seconds that you allow the whole operation to
    take.
```

**Ping inference gateway every other minute**
We used to ping inference gateway every minute. Reducing it to every
minute.


pykello:
> We used to ping inference gateway every minute. Reducing it to every
> minute.

In the commit message, should this be `every 2 minutes`?

pykello (approved):

bsatzger:
> > We used to ping inference gateway every minute. Reducing it to every
> > minute.
> 
> In the commit message, should this be `every 2 minutes`?

That was a typo and should say "every other minute". I fixed the message.

PR #2638, opened 2025-01-18
Drop columns from access_tag, and drop applied_tag and access_policy

jeremyevans:
This builds on #2637, and implements the final 2 steps of the plan in #2594.

This makes access_tag a pure project <-> accounts join table (modulo the created_at column to see when the two were associated).
    
While here, drop the applied_tag and access_policy tables, as they are no longer used.

### DIFF
--- a/migrate/20250120_access_tag_null_changes.rb
+++ b/migrate/20250120_access_tag_null_changes.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
+
+  up do
+    drop_table :applied_tag, :access_policy
+
+    # Remove non-account entries, otherwise the foreign key creation will fail
+    from(:access_tag).exclude(hyper_tag_id: from(:accounts).select(:id)).delete
+
+    alter_table(:access_tag) do
+      drop_index [:project_id, :name], name: :access_tag_project_id_name_index, concurrently: true
+      drop_column :id
+      drop_column :hyper_tag_table
+      drop_column :name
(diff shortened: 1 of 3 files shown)


migrate/20250119_access_tag_null_changes.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    # Remove non-account entries, otherwise the foreign key creation will fail
+    from(:access_tag).exclude(hyper_tag_id: from(:accounts).select(:id)).delete
+
+    alter_table(:access_tag) do
+      drop_index [:project_id, :name], name: :access_tag_project_id_name_index, concurrently: true
+      drop_column :id
+      drop_column :hyper_tag_table
+      drop_column :name
+      set_column_not_null :hyper_tag_id
```
enescakir, line 17: Given the significant operational work involved, we've decided against renaming the `access_tag` table.

What's your opinion on renaming `hyper_tag_id` to `account_id`? This way, the table would only contain `project_id` and `account_id` columns.

jeremyevans: I don't think think it is worth the effort.  If we are going to rename the column, we might as well rename the table at the same time.  I'm not against the change, though.

Note that the table still includes the `created_at` column.  I didn't remove that, since it might be useful to know when an account was associated with a project.

enescakir: Somehow I thought renaming a column would be easier than renaming a table, but they require similar effort. There's no need to do that right now.

enescakir (approved):

PR #2642, opened 2025-01-20
Merge e2e test cases

enescakir:
- **Deduplicate the client in GitHub runner E2E rspecs**
  

- **Merge e2e test case configs and image downloads**
  We are adding more e2e test cases for different services. Each service
  has its own configuration and requires different boot images. Some boot
  images are downloaded while setting up the VM host, while others are
  downloaded just before running the tests, like the GitHub E2E runner
  tests. This makes it hard to debug and maintain the test cases. This PR
  merges all the e2e test case configurations and image requirements into
  a single file. We will download all the required images at the beginning
  of the test during the VM host setup. This will make the test cases
  easier to maintain and debug.
  

- **Load default tests cases from config file**
  

- **Ensure all boot images are tested.**
  Previously, we randomly selected a boot image for VM tests. This means
  that if a specific boot image broke, we might miss it since we don’t
  test all boot images all the time.
  
  However, the green mark can mislead developers into thinking everything
  is fine.
  
  We should test all boot images to ensure they are functioning properly.
  This PR creates virtual machines to test at least a minimum number of
  boot images.
  
  If the number of provided boot images is less than 3, it will repeat the
  boot images to reach a total of 3. This way, we will create at least 3
  virtual machines for testing.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -82,7 +82,12 @@ jobs:
 
     - name: Add e2e script to Procfile
       run: |
-        echo "e2e: bin/ci --test-cases=${{ github.event_name != 'workflow_dispatch' && 'vm,github_runner_ubuntu_2204,github_runner_ubuntu_2004' || inputs.test_cases }}" >> Procfile
+        test_cases="$(yq -r '[.[].name] | join(",")' config/e2e_test_cases.yml)"
+        if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+          test_cases="${{ inputs.test_cases }}"
+        fi
+        echo "Running e2e test cases: $test_cases"
+        echo "e2e: bin/ci --test-cases=$test_cases" >> Procfile
 
     - name: Run tests
       env:
--- a/bin/ci
+++ b/bin/ci
@@ -8,44 +8,38 @@ require "optparse"
 @started_at = Time.now
 
 def main(options)
-  boot_images = get_additional_boot_images(options[:test_cases])
-  hetzner_server_st = Prog::Test::HetznerServer.assemble(vm_host_id: options[:vm_host_id], additional_boot_images: boot_images)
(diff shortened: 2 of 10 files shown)


enescakir:
@pykello What do you think about this PR since it overlaps with https://github.com/ubicloud/ubicloud/pull/2457 ?

.github/workflows/e2e.yml
```diff
@@ -82,7 +82,12 @@ jobs:
 
     - name: Add e2e script to Procfile
       run: |
-        echo "e2e: bin/ci --test-cases=${{ github.event_name != 'workflow_dispatch' && 'vm,github_runner_ubuntu_2204,github_runner_ubuntu_2004' || inputs.test_cases }}" >> Procfile
```
shikharbhardwaj, line 85: I missed updating this while adding a new test, thanks for simplifying this!

bin/ci
```diff
@@ -8,44 +8,38 @@ require "optparse"
 @started_at = Time.now
 
 def main(options)
-  boot_images = get_additional_boot_images(options[:test_cases])
-  hetzner_server_st = Prog::Test::HetznerServer.assemble(vm_host_id: options[:vm_host_id], additional_boot_images: boot_images)
+  all_test_cases = YAML.load_file("config/e2e_test_cases.yml").to_h { [_1["name"], _1] }
+  boot_images = all_test_cases.values_at(*options[:test_cases]).flat_map { _1["images"] }
```
shikharbhardwaj, line 12: jw: should we dedup here? does repeating boot images cause any issues/inefficiency? For eg. the postgres test I am adding will use the same image as the existing one

enescakir: Makes sense. I added `.uniq`. When we try to download the existing boot image to the host, it fails.

byucesoy (approved):

shikharbhardwaj (approved):

PR #2643, opened 2025-01-20
Add download button for Postgres CA certicates

shikharbhardwaj:
## Description
Add a download button on the postgres DB page to download the CA certificates as a `.pem` file.

## Tests
- [x] Local run
- [x] Unit tests

## Screenshots
While getting ready
image

After ready
image

Downloaded file
image

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -94,6 +94,10 @@ class PostgresResource < Sequel::Model
     private_subnet.firewalls.first.replace_firewall_rules(vm_firewall_rules)
   end
 
+  def ca_certificates
+    [root_cert_1, root_cert_2].join("\n") if root_cert_1 && root_cert_2
+  end
+
   module HaType
     NONE = "none"
     ASYNC = "async"
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -210,6 +210,16 @@ class Clover
           r.redirect "#{@project.path}#{pg.path}"
         end
       end
+
+      r.get "ca-certificates" do
+        authorize("Postgres:view", pg.id)
+
+        return 404 unless (certs = pg.ca_certificates)
+
+        response.headers["Content-Disposition"] = "attachment; filename=\"#{pg.name}.pem\""
+        response.headers["Content-Type"] = "application/x-pem-file"
+        certs
+      end
     end
 
     # 204 response for invalid names
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -938,6 +938,31 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/id/{postgres_database_id}/ca-certificates':
+    parameters:
+      - $ref: '#/components/parameters/project_id'
+      - $ref: '#/components/parameters/location'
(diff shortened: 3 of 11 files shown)


model/postgres/postgres_resource.rb
```diff
@@ -95,6 +95,14 @@ def set_firewall_rules
     private_subnet.firewalls.first.replace_firewall_rules(vm_firewall_rules)
   end
 
+  def ca_certificates
+    if root_cert_1.nil? || root_cert_2.nil?
```
enescakir, line 99: If one of them is not nil, don't we want to return one of them?

shikharbhardwaj: I think in this case we want to return both the certs or none as the customer needs to trust both of them for things to run smoothly. Based on my understanding of the certificate init/refresh code, both of them are set together so this essentially just checks if the certs are initialized or not.

model/postgres/postgres_resource.rb
```diff
@@ -95,6 +95,14 @@ def set_firewall_rules
...
+      nil
+    else
+      [root_cert_1, root_cert_2].join("\n")
+    end
+  end
```
enescakir, line 104:
```suggestion
  def ca_certificates
    [root_cert_1, root_cert_2].join("\n") if root_cert_1 && root_cert_2
  end
```

routes/project/location/postgres.rb
```diff
@@ -210,6 +210,20 @@ class Clover
           r.redirect "#{@project.path}#{pg.path}"
         end
       end
+
+      r.get "ca-certificates" do
+        authorize("Postgres:view", pg.id)
+
+        certs = pg.ca_certificates
+        if certs.nil?
```
enescakir, line 218:
```suggestion
        unless (certs = pg.ca_certificates)
```

routes/project/location/postgres.rb
```diff
@@ -210,6 +210,18 @@ class Clover
           r.redirect "#{@project.path}#{pg.path}"
         end
       end
+
+      r.get "ca-certificates" do
+        authorize("Postgres:view", pg.id)
+
+        unless (certs = pg.ca_certificates)
+          return 404
+        end
```
enescakir, line 219: nit:
```suggestion
        return 404 unless (certs = pg.ca_certificates)
```

furkansahin (commented):
Maybe let's find a way to disable the button if the certificate is not generated, yet?

shikharbhardwaj:
@furkansahin the other alternative I was thinking of if we want to show the field always was to keep it similar to the `Connection String` and show "Waiting for host to become ready" in case the certs are not initialized. wdyt?

model/postgres/postgres_resource.rb
```diff
@@ -95,6 +95,10 @@ def set_firewall_rules
     private_subnet.firewalls.first.replace_firewall_rules(vm_firewall_rules)
   end
 
+  def ca_certificates
+    [root_cert_1, root_cert_2].join("\n") if root_cert_1 && root_cert_2
```
byucesoy, line 99: We do similar operation in PostgresServerNexus as well. Could you please update it to use this function?

byucesoy (approved):
I think it makes sense to show the field always and show message "Waiting server to be ready" like we do for connection string.

model/postgres/postgres_resource.rb
shikharbhardwaj: Done

PR #2644, opened 2025-01-20
Introduce Leaseweb Client for network setup

furkansahin:
This is a very dirty PR.


rhizome/host/bin/setup-leaseweb-networking
```diff
@@ -0,0 +1,46 @@
...
+  gateway = subnet["gateway"].split("/").first
+  {
+    "to" => "default",
+    "via" => gateway,
+    "metric" => 100,
+    "on-link" => true
+  }
+end
+netplan_yaml["network"]["ethernets"][interface_name]["addresses"] = addresses
+netplan_yaml["network"]["ethernets"][interface_name]["routes"] = routes
```
fdr, line 42: maybe .dig would do something you want.

model/leaseweb_host.rb
```diff
@@ -0,0 +1,31 @@
...
+  def connection_string
+    Config.leaseweb_connection_string
+  end
+
+  def secret
+    Config.leaseweb_api_key
+  end
+end
+
+# Table: hetzner_host
```
enescakir, line 23: 🤔

lib/hosting/leaseweb_apis.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+require "excon"
+class Hosting::LeasewebApis
+  def initialize(leaseweb_host)
+    @host = leaseweb_host
+  end
+
+  def get_main_ip4
+    connection = Excon.new(@host.connection_string,
```
enescakir, line 10: We can do something similar there
https://github.com/ubicloud/ubicloud/pull/2563

PR #2645, opened 2025-01-20
Adding VmHostSlideNexus prog

macieksarnowicz:
Adding a new prog for VmHostSlice. A slice controls the allocation of CPUs and memory to one or more VMs. When created and associated with a VmHost, it creates a SystemCtl slice object (and underlying cgroup) on the host. It takes as an argument a set of CPUs to be affinitized to the group and an amount of memory to keep track of.

Also adding support for availability monitoring, following the standard pattern for ./bin/monitor script.

The VmHostSlice checks for three conditions:
- if the slice systemd unit is available
- if the allowed cpus is set to what is configured
- if the slice partition type is 'root' which indicates exclusive cpuset

### DIFF
--- a/model/vm_host_slice.rb
+++ b/model/vm_host_slice.rb
@@ -10,6 +10,7 @@ class VmHostSlice < Sequel::Model
 
   include ResourceMethods
   include SemaphoreMethods
+  include HealthMonitorMethods
   semaphore :destroy, :start_after_host_reboot, :checkup
 
   plugin :association_dependencies, cpus: :nullify
@@ -39,6 +40,47 @@ class VmHostSlice < Sequel::Model
     threads_per_core = vm_host.total_cpus / vm_host.total_cores
 
     update(cores: allocated_cpus / threads_per_core, total_cpu_percent: allocated_cpus * 100)
+  end
+
+  # Returns the name as used by systemctl and cgroup
+  def inhost_name
+    name + ".slice"
+  end
+
+  def init_health_monitor_session
+    {
+      ssh_session: vm_host.sshable.start_fresh_session
+    }
+  end
+
+  def up?(session)
+    # We let callers handle exceptions, as each calling method may have opt to handle them differently
+    session.exec!("systemctl is-active #{inhost_name}").split("\n").all?("active") &&
+      (session.exec!("cat /sys/fs/cgroup/#{inhost_name}/cpuset.cpus.effective").chomp == allowed_cpus_cgroup) &&
+      (session.exec!("cat /sys/fs/cgroup/#{inhost_name}/cpuset.cpus.partition").chomp == "root")
+  end
+
+  def check_pulse(session:, previous_pulse:)
+    reading = begin
+      up?(session[:ssh_session]) ? "up" : "down"
+    rescue
+      "down"
+    end
+    pulse = aggregate_readings(previous_pulse: previous_pulse, reading: reading)
+
+    if pulse[:reading] == "down" && pulse[:reading_rpt] > 5 && Time.now - pulse[:reading_chg] > 30 && !reload.checkup_set?
+      incr_checkup
+    end
+
+    pulse
+  end
+
+  def validate
+    super
+    errors.add(:name, "is not present") if name && name.empty?
+    errors.add(:family, "is not present") if family && family.empty?
+    errors.add(:name, "cannot be 'user' or 'system'") if name == "user" || name == "system"
+    errors.add(:name, "cannot contain a hyphen (-)") if name&.include?("-")
   end
 end
 
--- a/prog/vm/vm_host_slice_nexus.rb
+++ b/prog/vm/vm_host_slice_nexus.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+
+class Prog::Vm::VmHostSliceNexus < Prog::Base
+  subject_is :vm_host_slice
+
+  def self.assemble_with_host(name, vm_host, family:, allowed_cpus:, memory_gib:, is_shared: false)
+    DB.transaction do
+      vm_host_slice = VmHostSlice.create(
+        name: name,
+        is_shared: is_shared,
+        family: family,
+        cores: 0,
+        total_cpu_percent: 0,
+        used_cpu_percent: 0,
+        total_memory_gib: memory_gib,
+        used_memory_gib: 0,
+        vm_host_id: vm_host.id
+      )
+
+      # This will update the CPU allocation as well as total_cpu_percent and cores values
+      vm_host_slice.set_allowed_cpus(allowed_cpus)
+
+      Strand.create(prog: "Vm::VmHostSliceNexus", label: "prep") { _1.id = vm_host_slice.id }
+    end
+  end
+
+  def host
+    @host ||= vm_host_slice.vm_host
+  end
+
+  def before_run
+    when_destroy_set? do
+      hop_destroy if strand.label != "destroy"
+    end
+  end
+
+  label def prep
+    case host.sshable.cmd("common/bin/daemonizer --check prep_#{vm_host_slice.name}")
+    when "Succeeded"
+      host.sshable.cmd("common/bin/daemonizer --clean prep_#{vm_host_slice.name}")
+      vm_host_slice.update(enabled: true)
+      hop_wait
+    when "NotStarted", "Failed"
+      host.sshable.cmd("common/bin/daemonizer 'sudo host/bin/setup-slice prep #{vm_host_slice.inhost_name} \"#{vm_host_slice.allowed_cpus_cgroup}\"' prep_#{vm_host_slice.name}")
+    end
+
+    nap 1
+  end
+
+  label def wait
+    when_start_after_host_reboot_set? do
+      register_deadline(:wait, 5 * 60)
+      hop_start_after_host_reboot
+    end
+
+    when_checkup_set? do
+      hop_unavailable if !available?
+      decr_checkup
+    rescue Sshable::SshError
+      # Host is likely to be down, which will be handled by HostNexus. We still
+      # go to the unavailable state for keeping track of the state.
+      hop_unavailable
+    end
+
+    nap 30
+  end
+
+  label def unavailable
+    # If the slice becomes unavailable due to host unavailability, it first needs to
+    # go through start_after_host_reboot state to be able to recover.
+    when_start_after_host_reboot_set? do
+      incr_checkup
+      hop_start_after_host_reboot
+    end
+
+    begin
+      if available?
+        Page.from_tag_parts("VmHostSliceUnavailable", vm_host_slice.ubid)&.incr_resolve
+        decr_checkup
+        hop_wait
+      else
+        Prog::PageNexus.assemble("#{vm_host_slice.inhost_name} is unavailable", ["VmHostSliceUnavailable", vm_host_slice.ubid], vm_host_slice.ubid)
+      end
+    rescue Net::SSH::Disconnect, Net::SSH::ConnectionTimeout, Errno::ECONNRESET, Errno::ECONNREFUSED, IOError
+      # Host is likely to be down, which will be handled by HostNexus. No need
+      # to create a page for this case.
+    end
+
+    nap 30
+  end
+
+  label def destroy
+    decr_destroy
+
+    host.sshable.cmd("sudo host/bin/setup-slice delete #{vm_host_slice.inhost_name}")
+    vm_host_slice.destroy
+
+    pop "vm_host_slice destroyed"
+  end
+
+  label def start_after_host_reboot
+    host.sshable.cmd("sudo host/bin/setup-slice recreate-unpersisted #{vm_host_slice.inhost_name}")
+    decr_start_after_host_reboot
+
+    hop_wait
+  end
+
+  def available?
+    available = false
+    host.sshable.start_fresh_session do |session|
+      available = vm_host_slice.up? session
+    end
+
+    available
+  end
+end
--- a/spec/prog/vm/vm_host_slice_nexus_spec.rb
+++ b/spec/prog/vm/vm_host_slice_nexus_spec.rb
@@ -1,0 +1,221 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require_relative "../../../prog/vm/vm_host_slice_nexus"
+require_relative "../../../prog/vm/host_nexus"
+
+RSpec.describe Prog::Vm::VmHostSliceNexus do
+  subject(:nx) { described_class.new(Strand.create(id: "b231a172-8f56-8b10-bbed-8916ea4e5c28", prog: "Prog::Vm::VmHostSliceNexus", label: "create")) }
+
+  let(:sshable) {
+    Sshable.create
+  }
+
+  let(:vm_host) {
+    VmHost.create(
+      location: "x",
+      total_cores: 4,
+      used_cores: 1
+    ) { _1.id = sshable.id }
+  }
+
+  let(:vm_host_slice) {
+    VmHostSlice.create(
+      vm_host_id: vm_host.id,
+      name: "standard",
+      family: "standard",
+      is_shared: false,
+      cores: 1,
+      total_cpu_percent: 200,
+      used_cpu_percent: 0,
+      total_memory_gib: 4,
+      used_memory_gib: 0
+    )
+  }
+
+  before do
+    allow(nx).to receive_messages(vm_host_slice: vm_host_slice)
+    allow(vm_host_slice).to receive_messages(vm_host: vm_host)
+    allow(vm_host).to receive_messages(sshable: sshable)
+    (0..15).each { |i|
+      VmHostCpu.create(
+        spdk: i < 2,
+        vm_host_slice_id: (i == 2 || i == 3) ? vm_host_slice.id : nil
+      ) {
+        _1.vm_host_id = vm_host.id
+        _1.cpu_number = i
+      }
+    }
+  end
+
+  describe ".assemble_with_host" do
+    it "creates vm host slice" do
+      # prepare the host for the test
+      st_vh = Prog::Vm::HostNexus.assemble("1.2.3.4")
+      host = st_vh.subject
+      expect(host).not_to be_nil
+      host.update(total_cpus: 8, total_cores: 4)
+
+      (0..15).each { |i|
+        VmHostCpu.create(vm_host_id: host.id, cpu_number: i, spdk: i < 2)
+      }
+
+      # run the assemble test
+      st_rg = described_class.assemble_with_host("standard", host, family: "standard", allowed_cpus: [2, 3], memory_gib: 4)
+      rg = st_rg.subject
+      expect(rg).not_to be_nil
+      expect(rg.name).to eq("standard")
+      expect(rg.allowed_cpus_cgroup).to eq("2-3")
+      expect(rg.cores).to eq(1)
+      expect(rg.total_cpu_percent).to eq(200)
+      expect(rg.used_cpu_percent).to eq(0)
+      expect(rg.total_memory_gib).to eq(4)
+      expect(rg.used_memory_gib).to eq(0)
+      expect(rg.enabled).to be(false)
+      expect(rg.is_shared).to be(false)
+      expect(rg.id).to eq(st_rg.id)
+      expect(rg.ubid).to eq(st_rg.ubid)
+      expect(rg.ubid[..1] == "vs").to be true
+      expect(rg.vm_host).not_to be_nil
+      expect(rg.vm_host.id).to eq(host.id)
+    end
+  end
+
+  describe "#before_run" do
+    it "hops to destroy when needed" do
+      expect(nx).to receive(:when_destroy_set?).and_yield
+      expect { nx.before_run }.to hop("destroy")
+    end
+
+    it "does not hop to destroy if already in the destroy state" do
+      expect(nx).to receive(:when_destroy_set?).and_yield
+      expect(nx.strand).to receive(:label).and_return("destroy")
+      expect { nx.before_run }.not_to hop("destroy")
+    end
+  end
+
+  describe "#prep" do
+    it "starts prep on NotStarted" do
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check prep_standard").and_return("NotStarted")
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo host/bin/setup-slice prep standard.slice \"2-3\"' prep_standard")
+      expect(vm_host_slice).to receive(:inhost_name).and_return("standard.slice")
+
+      expect { nx.prep }.to nap(1)
+    end
+
+    it "starts prep on Failed" do
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check prep_standard").and_return("Failed")
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo host/bin/setup-slice prep standard.slice \"2-3\"' prep_standard")
+      expect(vm_host_slice).to receive(:inhost_name).and_return("standard.slice")
+
+      expect { nx.prep }.to nap(1)
+    end
+
+    it "hops to wait" do
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check prep_standard").and_return("Succeeded")
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --clean prep_standard")
+      expect(vm_host_slice).to receive(:update).with(enabled: true)
+
+      expect { nx.prep }.to hop("wait")
+    end
+
+    it "do nothing on random result" do
+      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check prep_standard").and_return("foobar")
+
+      expect { nx.prep }.to nap(1)
+    end
+  end
+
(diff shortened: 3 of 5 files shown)


prog/vm/vm_host_slice_nexus.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+
+class Prog::Vm::VmHostSliceNexus < Prog::Base
+  subject_is :vm_host_slice
+
+  semaphore :destroy, :start_after_host_reboot, :checkup
```
enescakir, line 6: Since the subject is `VmHostSlice`, and the model has semaphores. No need to define semaphores again at prog.

https://github.com/ubicloud/ubicloud/pull/2417

prog/vm/vm_host_slice_nexus.rb
```diff
@@ -0,0 +1,132 @@
...
+        name: name,
+        is_shared: is_shared,
+        family: family,
+        cores: 0,
+        total_cpu_percent: 0,
+        used_cpu_percent: 0,
+        total_memory_gib: memory_gib,
+        used_memory_gib: 0,
+        vm_host_id: vm_host.id
+      ) { _1.id = ubid.to_uuid }
```
enescakir, line 28:
```suggestion
    DB.transaction do
      vm_host_slice = VmHostSlice.create(
        name: name,
        is_shared: is_shared,
        family: family,
        cores: 0,
        total_cpu_percent: 0,
        used_cpu_percent: 0,
        total_memory_gib: memory_gib,
        used_memory_gib: 0,
        vm_host_id: vm_host.id
      )
```

There's no need to set the id explicitly; it will create the UBID automatically.

https://github.com/ubicloud/ubicloud/pull/2462

macieksarnowicz: I remember having some problems with that. Without this code, the id was created in the DB and did not have a format that converted to ubid with the right model class prefix. Can you elaborate how this supposed to work?

macieksarnowicz: Or, do I need to use .create_with_id?

macieksarnowicz: I will give it a try and see how it works.

enescakir: When did you try? It looks like it works with the latest code.

```ruby
[8] clover-development(main)> VmHostSlice.create(name: "test", cores: 2, total_cpu_percent: 0, used_cpu_percent:0, total_memory_gib: 1, used_memory_gib: 0, family: "standard", vm_host_id: VmHost.first.id)

=> #<VmHostSlice["vsxk5ra9fw2e5kwb6b1bq7da6h"] @values={:name=>"test", :enabled=>false, :is_shared=>false, :cores=>2, :total_cpu_percent=>0, :used_cpu_percent=>0, :total_memory_gib=>1, :used_memory_gib=>0, :family=>"standard", :created_at=>"2025-01-20 23:14:30", :vm_host_id=>"vhjab3390tfe1mt4h2cx53tkmt"}>

[9] clover-development(main)> _.id
=> "eccb8525-fc13-8b79-9f16-6585773b5468"

[10] clover-development(main)> udec "vsxk5ra9fw2e5kwb6b1bq7da6h"
=> #<VmHostSlice["vsxk5ra9fw2e5kwb6b1bq7da6h"] @values={:name=>"test", :enabled=>false, :is_shared=>false, :cores=>2, :total_cpu_percent=>0, :used_cpu_percent=>0, :total_memory_gib=>1, :used_memory_gib=>0, :family=>"standard", :created_at=>"2025-01-20 23:14:30", :vm_host_id=>"vhjab3390tfe1mt4h2cx53tkmt"}>
```

macieksarnowicz: Yes, I just pushed the change you suggested. It works. Thank you for the comment. 

It has been a while since this code was first written, and I must had it somewhat different, when it was not working. Anyway, the new version seems fine.

enescakir: Yeah, it merged recently. That’s probably the reason.

https://github.com/ubicloud/ubicloud/pull/2462

prog/vm/vm_host_slice_nexus.rb
```diff
@@ -0,0 +1,124 @@
...
+
+class Prog::Vm::VmHostSliceNexus < Prog::Base
+  subject_is :vm_host_slice
+
+  def self.assemble_with_host(name, vm_host, family:, allowed_cpus:, memory_gib:, is_shared: false)
+    fail "Must provide a VmHost." if vm_host.nil?
+    fail "Must provide slice name." if name.nil? || name.empty?
+    fail "Must provide family name." if family.nil? || family.empty?
+    fail "Slice name cannot be 'user' or 'system'." if name == "user" || name == "system"
+    fail "Slice name cannot contain a hyphen (-)." if name.include?("-")
```
fdr, line 11: What do you thin of some of these @jeremyevans ? Some of the checks seem overwrought by the normal standard, and yet not powerful enough to constrain program behavior in more advanced case (where something like `.is_a? VmHost` might be indicated...but that's a rare thing to have to do.

prog/vm/vm_host_slice_nexus.rb
```diff
@@ -0,0 +1,124 @@
...
+    host.sshable.cmd("sudo host/bin/setup-slice recreate-unpersisted #{vm_host_slice.inhost_name}")
+    decr_start_after_host_reboot
+
+    hop_wait
+  end
+
+  def available?
+    host.sshable.cmd("systemctl is-active #{vm_host_slice.inhost_name}").split("\n").all?("active") &&
+      (host.sshable.cmd("cat /sys/fs/cgroup/#{vm_host_slice.inhost_name}/cpuset.cpus.effective").chomp == vm_host_slice.allowed_cpus_cgroup) &&
+      (host.sshable.cmd("cat /sys/fs/cgroup/#{vm_host_slice.inhost_name}/cpuset.cpus.partition").chomp == "root")
```
fdr, line 122: I feel like I'm reading this code for the second time in this patch. Maybe it can be factored? I might have suggested encapsulting in rhizome, but I'm not sure that's an economy, given ruby startup time:

```
fdr@albatross ~/c/clover ((dcfc53f8)) [SIGINT]> hyperfine -- 'cat .tool-versions'
Benchmark 1: cat .tool-versions
  Time (mean ± σ):     861.6 µs ± 272.8 µs    [User: 422.4 µs, System: 609.4 µs]
  Range (min … max):     0.0 µs … 2331.3 µs    2080 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
fdr@albatross ~/c/clover ((dcfc53f8))> hyperfine -- 'ruby -e "puts File.read(%q(.tool-versions))"'
Benchmark 1: ruby -e "puts File.read(%q(.tool-versions))"
  Time (mean ± σ):      44.6 ms ±   5.6 ms    [User: 34.2 ms, System: 10.3 ms]
  Range (min … max):    37.6 ms …  66.0 ms    44 runs
```

@jeremyevans any of this track to you?

spec/prog/vm/vm_host_slice_nexus_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+      expect(Page).to receive(:from_tag_parts).and_return(nil)
+      expect { nx.unavailable }.to hop("wait")
+    end
+  end
+
+  describe "#available?" do
+    it "returns the available status" do
+      expect(sshable).to receive(:cmd).with("systemctl is-active standard.slice").and_return("active\nactive\n").once
+      expect(sshable).to receive(:cmd).with("cat /sys/fs/cgroup/standard.slice/cpuset.cpus.effective").and_return("2-3\n").once
+      expect(sshable).to receive(:cmd).with("cat /sys/fs/cgroup/standard.slice/cpuset.cpus.partition").and_return("root\n").once
```
fdr, line 232: encapsulating this in a method would let you factor out this triplet, seen multiple times, with one, shorter such assertion, e.g. `receive(:myhelper)`

spec/prog/vm/vm_host_slice_nexus_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+      used_cpu_percent: 0,
+      total_memory_gib: 4,
+      used_memory_gib: 0
+    )
+  }
+
+  before do
+    allow(nx).to receive_messages(vm_host_slice: vm_host_slice)
+    allow(vm_host_slice).to receive(:vm_host).and_return(vm_host)
+    allow(vm_host).to receive(:sshable).and_return(sshable)
```
fdr, line 39: Is it practical to reorganize the code so you don't need to do this for reference equality?  Given the cost of creating real persisted database records is sunk, it is somewhat preferable to limit mocking.

fdr (commented):
I think some stuff can be shorter via factoring perhaps? @byucesoy has the best handle on how a new monitor thing should look. That seems like the main substance here?

@jeremyevans may have comment on how to tweak some of the structural aspects of the code to reduce duplication and streamline the tests.

prog/vm/vm_host_slice_nexus.rb
jeremyevans: These seem like validations, and it would be best to have them or something similar in the `VmHostSlice#validate` method:

```ruby
  def validate
    super
    # no need for vm_host check, it's NOT NULL and auto validations or database constraints will catch it
    validates_presence([:name, :family])
    errors.add(:name, "cannot be 'user' or 'system'") if name == "user" || name == "system"
    errors.add(:name, "cannot contain a hyphen (-)") if name.include?("-")
  end
```

prog/vm/vm_host_slice_nexus.rb
jeremyevans: From what I can see, this duplicates the check in `VmHostSlice#check_pulse` .  I recommend adding `VmHostSlice#up?`:

```ruby
def up?(session:)
  session[:ssh_session].exec!("systemctl is-active #{inhost_name}").split("\n").all?("active") &&
    (session[:ssh_session].exec!("cat /sys/fs/cgroup/#{inhost_name}/cpuset.cpus.effective").chomp == allowed_cpus_cgroup) &&
    (session[:ssh_session].exec!("cat /sys/fs/cgroup/#{inhost_name}/cpuset.cpus.partition").chomp == "root")
rescue
  false
end
```

And calling it from both here and in `VmHostSlice#check_pulse`

spec/prog/vm/vm_host_slice_nexus_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+  let(:sshable) {
+    Sshable.create_with_id
+  }
+
+  let(:vm_host) {
+    VmHost.create(
+      location: "x",
+      total_cores: 4,
+      used_cores: 1
+    ) { _1.id = sshable.id }
```
jeremyevans, line 19: I'm not sure if this affects the test, but this seems backwards to me.  We set `sshable.id` to a VmHost id, we do not set `vm_host.id` to an Sshable id.  I'm not sure why Sshable has a ubid type, but maybe it is used for Sshables not associated with another existing class with a ubid type?

spec/prog/vm/vm_host_slice_nexus_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  let(:vm_host_slice) {
+    VmHostSlice.create_with_id(
```
jeremyevans, line 23:
```suggestion
    VmHostSlice.create(
```

We kept `create_with_id` for backwards compatibility, but there is no reason to use it in new code.

prog/vm/vm_host_slice_nexus.rb
macieksarnowicz: @jeremyevans - thanks! I was not aware of the validate method/pattern, I will covert to that.

prog/vm/vm_host_slice_nexus.rb
macieksarnowicz: The duplication is between VmHostSlice.check_pulse and VmHostSliceNexus.available? methods. The first one uses session.exec! to run the commands and the other uses vm_host.sshable.cmd. I could parameterize those calls and build an array of commands to run and expected responses and pass it to each of the methods, but I do not think this will improve the readability of the code.

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: Yes, however, if I get it correctly, the point of the tests is to make sure we are passing the right commands to the host in this case. And if I add a helper method, I still need to test that method, so this code just moves to another place.

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: Sorry, Daniel, can you please clarify. I am not following what are you suggesting to do here.

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: AFAIK, the sshable and vm_host have to share the ID value, they do not use PK/FK relationship. So I think it is easier to set up this way for tests. 
I will see if I can clean this up a bit.

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: OK, will change.

model/vm_host_slice.rb
```diff
@@ -40,6 +41,38 @@ def set_allowed_cpus(allowed_cpus)
...
+  def check_pulse(session:, previous_pulse:)
+    reading = begin
+      if session[:ssh_session].exec!("systemctl is-active #{inhost_name}").split("\n").all?("active") &&
+          (session[:ssh_session].exec!("cat /sys/fs/cgroup/#{inhost_name}/cpuset.cpus.effective").chomp == allowed_cpus_cgroup) &&
+          (session[:ssh_session].exec!("cat /sys/fs/cgroup/#{inhost_name}/cpuset.cpus.partition").chomp == "root")
+        "up"
+      else
+        "down"
+      end
+    rescue
```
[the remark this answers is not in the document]
macieksarnowicz: The other check_pulse methods do not do that, so I am not sure what the contract is between check_pulse and the `monitor` script. Would letting the exception pass through here break that script?

spec/prog/vm/vm_host_slice_nexus_spec.rb
fdr: Good question re: sshable ubid. My guess it is so ancient nobody really thought about it too much, it probably got processed in batch when ubids came into existence. I can't think of any reason it would stand alone.

prog/vm/vm_host_slice_nexus.rb
```diff
@@ -0,0 +1,124 @@
...
+  end
+
+  label def destroy
+    decr_destroy
+
+    vm_host_slice.update(enabled: false)
+
+    host.sshable.cmd("sudo host/bin/setup-slice delete #{vm_host_slice.inhost_name}")
+
+    vm_host_slice.destroy
```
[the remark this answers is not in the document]
macieksarnowicz: Good points, Hadi. I think this pertains more to the future code, so let me think about it and try what you suggest and we can continue this discussion with the next PR that I think will include the Allocator changes. Agree?

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: @jeremyevans - actually, what I said above is only partially correct. The vm_host has a FK to sshable table, so the sshable needs to exist first before vm_host can be created:
From vm_host.rb
# Foreign key constraints:
#  vm_host_id_fkey | (id) REFERENCES sshable(id)

prog/vm/vm_host_slice_nexus.rb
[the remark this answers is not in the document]
pykello: Sounds good. So, in this PR maybe just remove `vm_host_slice.update(enabled: false)` since the slice is destroyed in the same transaction and update's effect is never visible to others. Therefore, it's redundant.

model/vm_host_slice.rb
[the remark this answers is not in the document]
pykello: In that case, leave it as it is & this can be discussed later.

prog/vm/vm_host_slice_nexus.rb
macieksarnowicz: @jeremyevans - this is now refactored. It seems that the base class does a validation for basic checks expressed through the schema, as it checks for nil values of all properties if they have NOT NULL in the schema. So I could not use valdiates_presence because that resulted in duplicate errors. Instead, I just check for empty strings and the other two checks.

prog/vm/vm_host_slice_nexus.rb
macieksarnowicz: @jeremyevans , @fdr - I think I found a reasonable way to refactor this. Please take another look.

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: This is now refactored

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: I changed it use receive_messages, which is a bit less verbose.

spec/prog/vm/vm_host_slice_nexus_spec.rb
jeremyevans: The way this is handled in other parts of the code is the VmHost generates the UBID, inserts into `sshable`, then inserts into `vm_host`.

model/vm_host_slice.rb
```diff
@@ -40,6 +41,44 @@ def set_allowed_cpus(allowed_cpus)
...
+      incr_checkup
+    end
+
+    pulse
+  end
+
+  def validate
+    super
+    errors.add(:name, "is not present") if name && name.empty?
+    errors.add(:family, "is not present") if family && family.empty?
```
jeremyevans, line 78:
```suggestion
    errors.add(:name, "is not present") if name&.empty?
    errors.add(:family, "is not present") if family&.empty?
```

prog/vm/vm_host_slice_nexus.rb
jeremyevans: This seems fine.  We may want to consider using the `presence: true` option to the auto_validations plugin, which will add validations that `NOT NULL` text columns are not empty strings.

prog/vm/vm_host_slice_nexus.rb
jeremyevans: Looks good to me.

prog/vm/vm_host_slice_nexus.rb
```diff
@@ -0,0 +1,124 @@
...
+        Page.from_tag_parts("VmHostSliceUnavailable", vm_host_slice.ubid)&.incr_resolve
+        decr_checkup
+        hop_wait
+      else
+        Prog::PageNexus.assemble("#{vm_host_slice.inhost_name} is unavailable", ["VmHostSliceUnavailable", vm_host_slice.ubid], vm_host_slice.ubid)
+      end
+    rescue Sshable::SshError
+      # Host is likely to be down, which will be handled by HostNexus. No need
+      # to create a page for this case.
+    end
```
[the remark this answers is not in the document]
fdr: That's correct, we use a chunk of exceptions like this to deal with the host-down case:

```
  label def reboot
    begin
      q_last_boot_id = vm_host.last_boot_id.shellescape
      new_boot_id = sshable.cmd("sudo host/bin/reboot-host #{q_last_boot_id}").strip
    rescue Net::SSH::Disconnect, Net::SSH::ConnectionTimeout, Errno::ECONNRESET, Errno::ECONNREFUSED, IOError
      nap 30
    end
```

The laundry list of exceptions/rescue behavior should probably be encapsulated in some way. This is at least the second time I'm seeing this problem for a code off the top of my head, and I bet it happened more times and I forgot.

jeremyevans (commented):
Changes look good.

spec/prog/vm/vm_host_slice_nexus_spec.rb
macieksarnowicz: I think for the test code we are OK with how it is. Please let me know if you strongly disagree.

prog/vm/vm_host_slice_nexus.rb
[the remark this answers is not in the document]
macieksarnowicz: OK, I will update the list

spec/prog/vm/vm_host_slice_nexus_spec.rb
jeremyevans: I suppose for test code it should be OK.  We can change the test later if we want to tighten things up.

model/vm_host_slice.rb
macieksarnowicz: @jeremyevans - I wrote it that way, as it seems simpler, but rubocop converts it automatically to what got committed. Not sure how to force it to not do it or not complain. It does not even make sense to me for this to be a rule.

macieksarnowicz: Here is the rule: https://docs.rubocop.org/rubocop/cops_lint.html#lintsafenavigationwithempty

model/vm_host_slice.rb
```diff
@@ -40,6 +41,47 @@ def set_allowed_cpus(allowed_cpus)
...
+
+  def init_health_monitor_session
+    {
+      ssh_session: vm_host.sshable.start_fresh_session
+    }
+  end
+
+  def up?(session)
+    # We let callers handle exceptions, as each calling method may have opt to handle them differently
+    session.exec!("systemctl is-active #{inhost_name}").split("\n").all?("active") &&
```
byucesoy, line 58: One potential improvement here would be making just one call instead of 3 calls, which would speed up the health check, which is important for monitor. We can make this single call by;
- By concatenating the commands
- Or by creating a small script in rhizome that performs all these together and returns result.

byucesoy (approved):
Looks good in terms of monitoring.

model/vm_host_slice.rb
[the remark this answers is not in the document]
byucesoy: We are swallowing everything in other check_pulse's because we don't want an unexpected failure to cause crash in the monitor process. Instead we threat unexpected errors as unavailability and on-call can have a detailed look if needed.

model/vm_host_slice.rb
jeremyevans: In this case, it isn't the opposite of what is intended, though I can see that being true in other cases.  Just another case of rubocop making code worse.  It's fine to leave as is.

prog/vm/vm_host_slice_nexus.rb
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  label def start_after_host_reboot
+    host.sshable.cmd("sudo host/bin/setup-slice recreate-unpersisted #{vm_host_slice.inhost_name}")
+    decr_start_after_host_reboot
+
+    hop_wait
+  end
+
+  def available?
```
byucesoy, line 108: How about this

```
def available?
  vm_host_slice.up?(host.sshable.start_fresh_session)
end
```

macieksarnowicz: Will that auto-close the session? If a block is passed to `start_fresh_session`, the session will be auto-closed after the block is complete.

model/vm_host_slice.rb
macieksarnowicz: @fdr made a similar suggestion as well, but the showed this experiment, indicating that the cost of starting a Ruby script will likely outweight the savings. I can experiment with this if you'd like, though.

```
fdr@albatross ~/c/clover ((dcfc53f8)) [SIGINT]> hyperfine -- 'cat .tool-versions'
Benchmark 1: cat .tool-versions
  Time (mean ± σ):     861.6 µs ± 272.8 µs    [User: 422.4 µs, System: 609.4 µs]
  Range (min … max):     0.0 µs … 2331.3 µs    2080 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
fdr@albatross ~/c/clover ((dcfc53f8))> hyperfine -- 'ruby -e "puts File.read(%q(.tool-versions))"'
Benchmark 1: ruby -e "puts File.read(%q(.tool-versions))"
  Time (mean ± σ):      44.6 ms ±   5.6 ms    [User: 34.2 ms, System: 10.3 ms]
  Range (min … max):    37.6 ms …  66.0 ms    44 runs
```

prog/vm/vm_host_slice_nexus.rb
byucesoy: Probably not. Makes sense. Let's leave this as is then.

model/vm_host_slice.rb
byucesoy: We can combine them to a single call via concatenation then. Feel free to leave this to a separate PR though.

pykello (approved):

PR #2650, opened 2025-01-21
[k8s] Kubernetes Authorization & UI

aamederen:
### TODO
- [x] Tests
- [x] Split Auth changes commit
- [x] Kubeadm download (#2674)
- [x] Gated access to the feature (Feature Flag)
- [x] Display proper state
- [x] Better icon

Later
- Adapt according to #2652 
- Better Node naming


helpers/kubernetes_cluster.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    private_subnet = PrivateSubnet.from_ubid(request_body_params["private_subnet_id"])
+
+    unless private_subnet && private_subnet.location == @location
+      if api?
+        fail Validation::ValidationFailed.new({private_subnet_id: "Private subnet with the given id \"#{private_subnet_id}\" and the location \"#{@location}\" is not found"})
+      else
+        flash["error"] = "Private subnet not found"
+        r.redirect "#{@project.path}/kubernetes-cluster"
+      end
```
enescakir, line 18: We're returning back with an error message for `Validation::ValidationFailed` on the web. You might remove the branching here if that works for you.

https://github.com/ubicloud/ubicloud/blob/aaff966494396e31772d024f90749fb4732353ad/clover.rb#L225

```suggestion
      fail Validation::ValidationFailed.new({private_subnet_id: "Private subnet with the given id \"#{private_subnet_id}\" and the location \"#{@location}\" is not found"})
```

helpers/kubernetes_cluster.rb
```diff
@@ -0,0 +1,91 @@
...
+        kubernetes_cluster_id: st.subject.id
+      )
+
+      if api?
+        Serializers::KubernetesCluster.serialize(st.subject)
+      else
+        flash["notice"] = "'#{name}' will be ready in a few minutes"
+        request.redirect "#{@project.path}#{KubernetesCluster[st.id].path}"
+      end
+    end
```
enescakir, line 45: nit:
```suggestion
    DB.transaction do
      cluster = Prog::Kubernetes::KubernetesClusterNexus.assemble(
        name:,
        kubernetes_version: request.params["kubernetes_version"],
        private_subnet_id: private_subnet.id,
        project_id: @project.id,
        location: @location,
        cp_node_count: request.params["cp_nodes"].to_i
      ).subject

      Prog::Kubernetes::KubernetesNodepoolNexus.assemble(
        name: name + "-np",
        node_count: request.params["worker_nodes"].to_i,
        kubernetes_cluster_id: cluster.id
      )
    end

      if api?
        Serializers::KubernetesCluster.serialize(cluster)
      else
        flash["notice"] = "'#{name}' will be ready in a few minutes"
        request.redirect "#{@project.path}#{cluster.path}"
      end
```

routes/project/location/kubernetes_cluster.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      filter[:location] = @location
+      kc = @project.kubernetes_clusters_dataset.first(filter)
+
+      next (r.delete? ? 204 : 404) unless kc
+
+      r.get true do
+        authorize("KubernetesCluster:view", kc.id)
+        response.headers["Cache-Control"] = "no-store"
```
enescakir, line 27: We added this header only to the PostgreSQL detail page because it contains a confidential connection string. If you don't have anything like that, there's no need to add it.

serializers/kubernetes_cluster.rb
```diff
@@ -0,0 +1,22 @@
...
+  def self.serialize_internal(kc, options = {})
+    base = {
+      id: kc.ubid,
+      name: kc.name,
+      cp_node_count: kc.cp_node_count,
+      private_subnet_id: kc.private_subnet_id,
+      kubernetes_version: kc.kubernetes_version,
+      location: kc.display_location,
+      vms: kc.cp_vms.map { |vm| {name: vm.name, state: vm.display_state, hostname: vm.ephemeral_net4.to_s} },
+      nodepools: kc.kubernetes_nodepools.map { |np| {name: np.name, node_count: np.node_count, vms: np.vms.map { |vm| {name: vm.name, state: vm.display_state, hostname: vm.ephemeral_net4.to_s} }} }
```
enescakir, line 13: I highly recommend loading them eagerly; otherwise, you'll end up with a lot of N+1 queries.

helpers/kubernetes_cluster.rb
aamederen: applied. Also, changed all routes to web? only for now.

routes/project/kubernetes_cluster.rb
```diff
@@ -0,0 +1,24 @@
...
+class Clover
+  hash_branch(:project_prefix, "kubernetes-cluster") do |r|
+    r.get true do
+      kubernetes_cluster_list
+    end
+
+    r.web do
+      r.post true do
+        @location = LocationNameConverter.to_internal_name(r.params["location"])
+        kubernetes_cluster_post(r.params["name"])
```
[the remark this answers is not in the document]
aamederen: reverted to web-only for now

views/kubernetes-cluster/index.erb
```diff
@@ -0,0 +1,58 @@
...
+      "components/button", locals: { text: "Create Kubernetes Cluster", link: "kubernetes-cluster/create" })
+    %>
+    <% end %>
+  </div>
+<% else %>
+<div class="mt-6">
+  <%== render(
+    "components/empty_state",
+    locals: {
+      icon: "hero-arrows-pointing-out",
```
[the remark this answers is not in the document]
aamederen: #2693 #2696

fdr:
seems like " Make tests pass " commit should be squashed

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -40,6 +40,18 @@ def path
   def endpoint
     api_server_lb.hostname
   end
+
+  def kubeconfig
+    rbac_token = cp_vms.first.sshable.cmd("sudo kubectl --kubeconfig /etc/kubernetes/admin.conf -n kube-system get secret k8s-access -o jsonpath='{.data.token}' | base64 -d")
```
fdr: huh what is this doing with root exactly?

fdr: probably reading admin.conf I see. I wonder if you can down-scope the amount of root going on here with process redirection, i.e. `<(sudo cat /etc/kubernetes/admin.conf)`

rhizome/kubernetes/bin/init-cluster
```diff
@@ -56,3 +58,31 @@ config_path = "/tmp/kubeadm-config.yaml"
 safe_write_to_file(config_path, config.to_yaml)
 
 r "sudo kubeadm init --config #{config_path}"
+
+api_server_up = false
+5.times do
+  r("kubectl --kubeconfig=/etc/kubernetes/admin.conf get --raw='/healthz'", expect: [0])
```
fdr, line 64: expect [0] is redundant, no?

routes/project/location/kubernetes_cluster.rb
```diff
@@ -27,6 +27,17 @@ class Clover
...
+
+        r.on "kubeconfig" do
+          r.get do
+            authorize("KubernetesCluster:create", kc.id)
+
+            content = kc.kubeconfig
+            response["Content-Type"] = "text/plain"
+            response["Content-Disposition"] = "attachment; filename=\"#{kc.name}-kubeconfig.yaml\""
+            response.write(content)
+          end
```
fdr, line 39: huh, seems neat, but why? also I sort of loathe content-disposition attachment for something that is plain text, I kinda want to see it first pretty often rather than having it go straight to cluttering my file system when I click. But maybe I'm nuts.

spec/model/kubernetes/kubernetes_cluster_spec.rb
```diff
@@ -41,4 +41,31 @@
...
+          user:
+            client-certificate-data: "mocked_cert_data"
+            client-key-data: "mocked_key_data"
+    YAML
+    let(:sshable) { instance_double(Sshable) }
+    let(:vm) { instance_double(Vm, sshable: sshable) }
+    let(:cp_vms) { [vm] }
+
+    it "removes client certificate and key data from users and adds an RBAC token to users" do
+      expect(kc).to receive(:cp_vms).and_return(cp_vms).twice
```
fdr: while you can do .twice in some situations, usually `.at_least(:once)` is just a lot more robust and loses nothing. Too bad it's wordy. I wouldn't go so far to say that rspec should have made `at_least(:once)` the default behavior, but it's close. Surely it should have a shorter-hand.

model/kubernetes/kubernetes_cluster.rb
fdr: also, don't bother necessarily if you have a lot of other root kubectl calls that cannot be repair, no point gold-plating in one place.

routes/project/location/kubernetes_cluster.rb
aamederen: Normally, other cloud providers do that with CLI, and users don't see the kubeconfig at all. It's meant to be placed in it's default place in user home directory (by the CLI) and the user would use kubectl normally. Since we don't have those means yet, we will ask users to download and place it/or specify in their kubectl commands.

fdr:
okay probably my main question/grievance is around content disposition attachment because I loathe fishing around in my downloads directory to even take a glance at plain text as a programmer, but in general, I don't concretely understand the notably rigorous interest in exporting the yaml (though I have some guesses)

model/kubernetes/kubernetes_cluster.rb
aamederen: We run kubeadm with sudo during node provisioning, not sure if that's easy to escape.

Kubectl is used only here (for CP). And later with the upgrades.

Changing didn't hurt so just did it.

fdr:
feel like this is loaded with stuff in @jeremyevans area of expertise (incl. authorization)

routes/project/location/kubernetes_cluster.rb
aamederen: As discussed, left the download flow as is, but moved to inline with the rest of the cluster details

image

helpers/kubernetes_cluster.rb
```diff
@@ -0,0 +1,70 @@
...
+      ).subject
+
+      Prog::Kubernetes::KubernetesNodepoolNexus.assemble(
+        name: name + "-np",
+        node_count: request.params["worker_nodes"].to_i,
+        kubernetes_cluster_id: kc.id
+      )
+    end
+
+    flash["notice"] = "'#{name}' will be ready in a few minutes"
```
jeremyevans, line 37: You don't have to change things, but this can be inside the transaction block, so you don't need the earlier `kc = nil`.

helpers/kubernetes_cluster.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+class Clover
+  def kubernetes_cluster_post(name)
```
jeremyevans, line 4: In general, helper methods should only be used when they reduce redundancy (when they are called from multiple places).  If you are only calling these helper methods from a single place, inline the logic at the callsite.

serializers/kubernetes_cluster.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class Serializers::KubernetesCluster < Serializers::Base
```
jeremyevans, line 3: I recommend against the creation of new serializers, especially for use in the web routes. They add unnecessary and net-negative indirection.  I recommend directly calling methods on the model objects in the web views instead of using a serializer.

views/kubernetes-cluster/create.erb
```diff
@@ -0,0 +1,29 @@
...
+<%== render(
+  "components/page_header",
+  locals: {
+    breadcrumbs: [
+      %w[Projects /project],
+      [@project_data[:name], @project_data[:path]],
+      ["Kubernetes Clusters", "#{@project_data[:path]}/kubernetes-cluster"],
+      %w[Create #]
+    ]
+  }
```
jeremyevans, line 12:
```suggestion
<%== part(
  "components/page_header",
  breadcrumbs: [
    %w[Projects /project],
    [@project_data[:name], @project_data[:path]],
    ["Kubernetes Clusters", "#{@project_data[:path]}/kubernetes-cluster"],
    %w[Create #]
  ]
```

I recently added a commit to use `part` instead of `render` with `:locals`.  Above is an example of such a change.  Please change all other `render` calls with `:locals` to use `part`.

routes/project/kubernetes_cluster.rb
```diff
@@ -3,6 +3,8 @@
 class Clover
   hash_branch(:project_prefix, "kubernetes-cluster") do |r|
     r.web do
+      next unless @project.get_ff_kubernetes
```
jeremyevans, line 6: This check is missing in `routes/project/location/kubernetes_cluster.rb`.  Shouldn't it be there as well?

routes/project/location/kubernetes_cluster.rb
```diff
@@ -27,6 +27,17 @@ class Clover
           kc.incr_destroy
           204
         end
+
+        r.on "kubeconfig" do
+          r.get do
+            authorize("KubernetesCluster:create", kc.id)
```
jeremyevans, line 33: This permission seems incorrect.  For one this is a read-only request, not a request to change anything.  Second, in other places `KubernetesCluster:create` is checked against the project's id, not the id of the cluster.  `KubernetesCluster:view` is a read-only permission, but you may not want to grant everyone with access to view the cluster to download the config.  If you don't want to add a specific permission for this `KubernetesCluster:edit` or `KubernetesCluster:delete` is probably better, as that is designed to check against the cluster id.

routes/project/location/kubernetes_cluster.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        r.on "kubeconfig" do
+          r.get do
+            authorize("KubernetesCluster:create", kc.id)
+
+            content = kc.kubeconfig
+            response["Content-Type"] = "text/plain"
+            response["Content-Disposition"] = "attachment; filename=\"#{kc.name}-kubeconfig.yaml\""
+            response.write(content)
```
jeremyevans, line 38:
```suggestion
            content
```

routes/project/location/kubernetes_cluster.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        r.delete true do
+          authorize("KubernetesCluster:delete", kc.id)
+          kc.incr_destroy
+          204
+        end
+
+        r.on "kubeconfig" do
+          r.get do
```
jeremyevans, line 32:
```suggestion
        r.get "kubeconfig" do
```

With appropriate indentation fixes inside the block, and removal of extra `end`.  This isn't just for simplicity, the current code is incorrect as it will allow arbitrary segments after `kubeconfig` in the URL (since you did not pass an argument to `r.get`.

helpers/kubernetes_cluster.rb
```diff
@@ -0,0 +1,70 @@
...
+    required_parameters = ["name", "location", "version", "private_subnet_id", "cp_nodes", "worker_nodes"]
+    request_body_params = validate_request_params(required_parameters)
+
+    private_subnet = PrivateSubnet.from_ubid(request_body_params["private_subnet_id"])
+
+    unless private_subnet && private_subnet.location == @location
+      fail Validation::ValidationFailed.new({private_subnet_id: "Private subnet with the given id \"#{request_body_params["private_subnet_id"]}\" and the location \"#{@location}\" is not found"})
+    end
+
+    authorize("PrivateSubnet:edit", private_subnet.id)
```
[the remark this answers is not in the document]
aamederen: As far as I can see, the authorization logic would also handle the null check, i.e. if the given object is not found, then `authorize` will error out, right?

helpers/kubernetes_cluster.rb
aamederen: We have only a single place where we call the post for now, but in the upcoming weeks we will add the update paths for this resource, similar to other resources like postgres or VM. Therefore, I think we can keep it as is. 

On the other hand, as far as I can see, offloading route logic to a `helper` became a pattern in code. IMO, obeying a pattern would be preferable, even if there's no benefit for such a case.

What do you think?

serializers/kubernetes_cluster.rb
aamederen: when you mention "I recommend against the creation of **new** serializers", do you mean this is an anti-pattern that we should look for abandoning in the code? If so, we can do that, but if not, similar to the previous pattern discussion, we can let it stay for the sake of obeying the pattern + preparing for the API routes of the upcoming weeks.

helpers/kubernetes_cluster.rb
jeremyevans: If you are fairly sure you will be calling this from an additional place in the future, then it's OK to keep it in a helper.  However, be aware that similar helper methods in other helper files only exist because the methods are called from multiple places.  The general rule is that if you only need the logic in one place, put the logic in that place, do not add unnecessary indirection.

views/kubernetes-cluster/create.erb
aamederen: Wow :D I will need to do a bunch of rebases to get there

routes/project/kubernetes_cluster.rb
aamederen: I left it untouched on purpose, thinking that, if for some reason, someone is feature-blocked but has a provisioned cluster anyway, they should have some kind of access to that resource. I couldn't think of any security implications of it but if you think there is, we can add the check to location too.

What do you think?

serializers/kubernetes_cluster.rb
jeremyevans: Correct, I think it is an anti-pattern, at least for usage in web routes.  For web routes, we are using serializers as presenters, and a plain hash or array of hashes as a presenter object I think makes things worse.  I don't think we need such presenter objects at all, and can use the model objects directly.

For api routes, some form of serialization is necessary.  I think such serialization is probably best either inlined in the route (if only used in one route), or added as Clover helper method (if used in multiple routes).  However, I don't object to the current use of serializers for API routes.

helpers/kubernetes_cluster.rb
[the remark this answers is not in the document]
jeremyevans: The issue is if the private subnet exists, but the user does not have access to it, this leaks information about the private subnet.

helpers/kubernetes_cluster.rb
aamederen: Thank you, I will keep it then, as I'm sure we will have edits routes using this method.

migrate/20250121_kubernetes_tags.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    run <<~SQL
+      INSERT INTO
+        action_type (id, name)
+      VALUES
+        ('ffffffff-ff00-835a-87ff-f04d80d85dc0', 'KubernetesCluster:view'),
```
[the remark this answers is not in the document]
aamederen: Added the comments.

helpers/kubernetes_cluster.rb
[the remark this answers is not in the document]
aamederen: Thanks, I made a change assuming a passed `authorize` call would imply a null check

helpers/kubernetes_cluster.rb
aamederen: thanks, I moved them inside.

funny thing is, I vaguely remember moving those out of the transaction block due to another review comment, but I cannot find it :D

serializers/kubernetes_cluster.rb
aamederen: Alright, that's good to know. I will clean that up.

routes/project/location/kubernetes_cluster.rb
aamederen: Thanks for catching this. I think edit makes sense now, maybe we can include another permission specific for kubeconfig too..

routes/project/location/kubernetes_cluster.rb
aamederen: Thanks, I changed accordingly

migrate/20250121_kubernetes_tags.rb
```diff
@@ -0,0 +1,36 @@
...
+        ('ffffffff-ff00-834a-87ff-ff826c028210', 'ffffffff-ff00-835a-87ff-f04d807343a0'),
+        ('ffffffff-ff00-834a-87ff-ff826c028210', 'ffffffff-ff00-835a-87c1-3601ae0bb4e0');
+    SQL
+  end
+
+  down do
+    run <<~SQL
+      DELETE FROM applied_action_tag WHERE tag_id = 'ffffffff-ff00-834a-87ff-ff826c028210';
+      DELETE FROM action_type WHERE name LIKE 'KubernetesCluster%';
+      DELETE FROM action_tag WHERE name LIKE 'KubernetesCluster%';
```
jeremyevans, line 33: For safety, should probably add ` AND project_id IS NULL` here.  Wouldn't want to delete user action tags that start with that.

migrate/20250121_kubernetes_tags.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    run <<~SQL
```
jeremyevans, line 5: Just FYI, Sequel does not support multiple statements in a single `run` call.  This happens to work on the postgres adapter, but it can fail on other adapters.  It's not required, but its best if each statement has its own `run` method.

jeremyevans (commented):
All of my previous comments have been addressed.  Only a couple new minor comments, both optional.

migrate/20250121_kubernetes_tags.rb
aamederen: Thanks, applied that to action_tag.

migrate/20250121_kubernetes_tags.rb
aamederen: Good to know, made it all.

jeremyevans (commented):
Changes look good.

fdr (approved):

PR #2651, opened 2025-01-21
Enable transparent cache by default for new users

velioglu:
Enabling transparent cache by default for new users, as it is proved that it works reliably and improves performance significantly. For the existing users, it will be enabled separately.

### DIFF
--- a/migrate/20250121_enable_transparent_cache_by_default.rb
+++ b/migrate/20250121_enable_transparent_cache_by_default.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:github_installation) do
+      set_column_default(:cache_enabled, true)
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


migrate/20250121_enable_transparent_cache_by_default.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:github_installation) do
+      set_column_default(:cache_enabled, true)
+    end
+
+    run "UPDATE github_installation SET cache_enabled = true"
```
enescakir, line 9: Should we enable this only for new customers for a while and observe their reactions? Should we update it for existing customers at this time?

enescakir (commented):
Should we add a note to the documentation?

migrate/20250121_enable_transparent_cache_by_default.rb
velioglu: As we talked about sending informational email to existing customers in batch, updated PR to enable it only for new ones.

velioglu:
> Should we add a note to the documentation?

Yes, I'll be updating documentation as well.

PR #2652, opened 2025-01-21
Make Kubernetes Resources Created in Internal Project

aamederen:
Make Kubernetes Resources Created in Internal Project

This change makes the resources required for a kubernetes cluster get
created within a pre-created kubernetes service project.

Previously, the resources for a cluster were meant to live in project
belonging to the customer. Even the subnet was meant to be provided by
the customer once, but then we took the creation of that subnet in our hands.
Now, the subnet is created during the assemble phase in the
kubernetes service project and all VMs and LBs will also be created in that
project. The name of the subnet is also changed to a ubid-based one.

The KubernetesCluster entity is still tied to customer project and the
resources will be shown to the customer in the UI & API.

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -16,26 +16,24 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
       Validation.validate_kubernetes_name(name)
       Validation.validate_kubernetes_cp_node_count(cp_node_count)
 
+      ubid = KubernetesCluster.generate_ubid
       subnet = if private_subnet_id
-        project.private_subnets_dataset.first(id: private_subnet_id) || fail("Given subnet is not available in the given project")
+        PrivateSubnet[id: private_subnet_id, project_id: Config.kubernetes_service_project_id] || fail("Given subnet is not available in the k8s project")
       else
-        subnet_name = name + "-k8s-subnet"
-        ps = project.private_subnets_dataset.first(:location_id => location_id, Sequel[:private_subnet][:name] => subnet_name)
-        ps || Prog::Vnet::SubnetNexus.assemble(
-          project_id,
-          name: subnet_name,
+        Prog::Vnet::SubnetNexus.assemble(
+          Config.kubernetes_service_project_id,
+          name: "#{ubid}-subnet",
           location_id:,
           ipv4_range: Prog::Vnet::SubnetNexus.random_private_ipv4(Location[location_id], project, 18).to_s
         ).subject
       end
 
       # TODO: Validate location
-      # TODO: Move resources (vms, subnet, LB, etc.) into our own project
       # TODO: Validate node count
 
-      kc = KubernetesCluster.create_with_id(name:, version:, cp_node_count:, location_id:, target_node_size:, target_node_storage_size_gib:, project_id: project.id, private_subnet_id: subnet.id)
+      KubernetesCluster.create(name:, version:, cp_node_count:, location_id:, target_node_size:, target_node_storage_size_gib:, project_id: project.id, private_subnet_id: subnet.id) { _1.id = ubid.to_uuid }
 
-      Strand.create(prog: "Kubernetes::KubernetesClusterNexus", label: "start") { _1.id = kc.id }
+      Strand.create(prog: "Kubernetes::KubernetesClusterNexus", label: "start") { _1.id = ubid.to_uuid }
     end
   end
 
--- a/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
+++ b/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
@@ -5,8 +5,8 @@ require_relative "../../model/spec_helper"
 RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
   subject(:nx) { described_class.new(Strand.new) }
 
-  let(:project) { Project.create(name: "default") }
-  let(:subnet) { PrivateSubnet.create(net6: "0::0", net4: "127.0.0.1", name: "x", location_id: Location::HETZNER_FSN1_ID, project_id: project.id) }
+  let(:customer_project) { Project.create(name: "default") }
+  let(:subnet) { PrivateSubnet.create(net6: "0::0", net4: "127.0.0.1", name: "x", location_id: Location::HETZNER_FSN1_ID, project_id: Config.kubernetes_service_project_id) }
 
   let(:kubernetes_cluster) {
     kc = KubernetesCluster.create(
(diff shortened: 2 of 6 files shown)


fdr:
Looks good to me on the surface, but the most sophisticated appreciator of this kind of code is @byucesoy , so I'd let him have a crack at trying to see the forest for the trees...

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -13,6 +13,12 @@ def self.assemble(name:, kubernetes_version:, private_subnet_id:, project_id:, l
         fail "Invalid Kubernetes Version"
       end
 
+      subnet = Prog::Vnet::SubnetNexus.assemble(
+        Config.kubernetes_service_project_id,
+        location: location,
+        ipv4_range: "10.10.0.0/16" # TODO: a better way?
```
fdr, line 19: don't we have a deconfliction code of some sort? this guarantees any two k8s clusters can never peer via IPv**4** successfully (which, as you know, would be fine with me personally, but probably is not commercially ideal).

fdr (commented):
.

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: Summoning @furkansahin superpowers!

Maybe we can reuse `Prog::Vnet::SubnetNexus.random_private_ipv4` with some range customization modification?

byucesoy (approved):
From the project management standpoint, all looks good. I'll leave subnet management part to Furkan.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -13,6 +13,12 @@ def self.assemble(name:, kubernetes_version:, private_subnet_id:, project_id:, l
         fail "Invalid Kubernetes Version"
       end
 
+      subnet = Prog::Vnet::SubnetNexus.assemble(
```
enescakir, line 16: Should we name it as `#{CLUSTER_UBID/NAME}-subnet`

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -13,6 +13,12 @@ def self.assemble(name:, kubernetes_version:, private_subnet_id:, project_id:, l
...
+        Config.kubernetes_service_project_id,
+        location: location,
```
enescakir, line 18: nit:
```suggestion
        location:,
```

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -5,51 +5,57 @@
...
   before do
+    allow(Config).to receive(:kubernetes_service_project_id).and_return(Project.create(name: "UbicloudKubernetesService").id)
     allow(nx).to receive(:kubernetes_cluster).and_return(kubernetes_cluster)
   end
 
   describe ".assemble" do
     it "validates input" do
       expect {
-        described_class.assemble(project_id: SecureRandom.uuid, name: "k8stest", kubernetes_version: "v1.32", location: "hetzner-fsn1", cp_node_count: 3, private_subnet_id: subnet.id)
+        described_class.assemble(project_id: SecureRandom.uuid, name: "k8stest", kubernetes_version: "v1.32", location: "hetzner-fsn1", cp_node_count: 3)
```
enescakir, line 41: You might use hardcoded project ID

https://github.com/ubicloud/ubicloud/pull/799

prog/kubernetes/kubernetes_cluster_nexus.rb
fdr: Something like that, you'd have to filter against the customer's project, rather than within the project implied by the subnet, that may require some refactoring. But I think we can use the same routine.

prog/kubernetes/kubernetes_cluster_nexus.rb
fdr: motive? I guess to make it more legible for operations? I guess right now it's the default, right, unless I misread.

enescakir: > motive? 

Operational readability. When we have a private subnet in production, if it follows the `kc.....-subnet` pattern, we can easily identify it as a private subnet for a k8s cluster.

> I guess right now it's the default, right, unless I misread.

The default name of the private subnet is the first 8 characters of its UBID, so it's not very informative.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -16,26 +16,25 @@ def self.assemble(name:, project_id:, location_id:, version: "v1.32", private_su
...
+      ubid = KubernetesCluster.generate_ubid
       subnet = if private_subnet_id
-        project.private_subnets_dataset.first(id: private_subnet_id) || fail("Given subnet is not available in the given project")
+        PrivateSubnet[id: private_subnet_id, project_id: Config.kubernetes_service_project_id] || fail("Given subnet is not available in the k8s project")
       else
-        subnet_name = name + "-k8s-subnet"
-        ps = project.private_subnets_dataset.first(:location_id => location_id, Sequel[:private_subnet][:name] => subnet_name)
-        ps || Prog::Vnet::SubnetNexus.assemble(
-          project_id,
+        subnet_name = ubid.to_s + "-subnet"
```
enescakir, line 23: nit:
```suggestion
        subnet_name = "#{ubid}-subnet"
```

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -16,26 +16,25 @@ def self.assemble(name:, project_id:, location_id:, version: "v1.32", private_su
...
+        Prog::Vnet::SubnetNexus.assemble(
+          Config.kubernetes_service_project_id,
           name: subnet_name,
```
enescakir, line 26: nit: It has single usage. Do we really need a variable?
```suggestion
          name: "#{ubid}-subnet",
```

enescakir (approved):

spec/routes/web/kubernetes_cluster_spec.rb
```diff
@@ -157,7 +158,7 @@
         expect(new_kc.project_id).to eq(project.id)
         expect(new_kc.cp_node_count).to eq(3)
         expect(new_kc.nodepools.first.node_count).to eq(2)
-        expect(new_kc.private_subnet.name).to eq("k8stest-k8s-subnet")
+        expect(new_kc.private_subnet.name).to eq(new_kc.ubid.to_s + "-subnet")
```
enescakir, line 161: nit
```suggestion
        expect(new_kc.private_subnet.name).to eq("#{new_kc.ubid}-subnet")
```

PR #2658, opened 2025-01-22
Read Load Balancer Node state according to threshold

furkansahin:
We are updating load balancer node when the defined thresholds are passed, but a flipping health check, or an unrelated update may simply declare nodes unhealthy when they are just in the verification phase (state_count < threshold).

This commit adds the new node_state function to better understand the status. We basically smoothen the state read curve.

### DIFF
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -5,7 +5,6 @@ require_relative "../model"
 class LoadBalancer < Sequel::Model
   many_to_one :project
   many_to_many :vms
-  many_to_many :active_vms, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: {state: ["up"]}
   many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: ["evacuating", "detaching"])
   one_to_one :strand, key: :id
   many_to_one :private_subnet
@@ -24,6 +23,10 @@ class LoadBalancer < Sequel::Model
 
   def path
     "/location/#{private_subnet.display_location}/load-balancer/#{name}"
+  end
+
+  def active_vms
+    vms_to_dns.select { |vm| vm.load_balancers_vms.node_state == "up" }
   end
 
   def add_vm(vm)
--- a/model/load_balancers_vms.rb
+++ b/model/load_balancers_vms.rb
@@ -4,6 +4,18 @@ require_relative "../model"
 
 class LoadBalancersVms < Sequel::Model
   include ResourceMethods
+  many_to_one :load_balancer
+
+  def node_state
+    case state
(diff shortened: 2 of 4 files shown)


bsatzger (approved):

model/load_balancer.rb
```diff
@@ -26,6 +25,10 @@ def path
     "/location/#{private_subnet.display_location}/load-balancer/#{name}"
   end
 
+  def active_vms
+    vms_to_dns.select { |vm| vm.load_balancers_vms.node_state == "up" }
```
fdr, line 29: Does something like this work/make sense? Rather than loading up associations only to do filtering and find one record?

```ruby
VmToDns.join(:load_balancer_vm, {vm_id: :id, node_state: 'up'}).select_all(:vm_to_dns)
```

fdr:
Maybe I can't quite read this out of the patch and my context is too broken from previous patches, I'm a bit perplexed by the existence of multiple thresholds (up/down), rather than something like our old-school value/repeat/change paradigm.  Can you tell me about the design history of this? (or pass a link to a commit or whatever you think is most relevant)

model/load_balancer.rb
bsatzger: This is the issue we observed in prod multiple times: A health probe reports unavailable for a short duration and then available again with health probes succeeding continuously, but the load balancer doesn't route traffic. A manual `incr_update_load_balancer` on the load balancer fixes it. 

Is it possible that the logic in `Vnet::UpdateLoadBalancerNode` acts on a cached state where the value of column `state` is still `down` while in the DB it's already `up`?

bsatzger: [Here](https://github.com/ubicloud/ubicloud/blob/2c0ed37580380bc12779ce0f4a09b3f5596d3993/prog/vnet/load_balancer_health_probes.rb#L28) we set `incr_update_load_balancer` when the probe succeeds again. Threshold is 1 in this case.

Which triggers: https://github.com/ubicloud/ubicloud/blob/2c0ed37580380bc12779ce0f4a09b3f5596d3993/prog/vnet/update_load_balancer_node.rb#L25
https://github.com/ubicloud/ubicloud/blob/2c0ed37580380bc12779ce0f4a09b3f5596d3993/prog/vnet/update_load_balancer_node.rb#L37

fdr: thing is, I don't understand how this problem is different than how we smooth flaps for postgres (but maybe it is), so I wonder why we have a second approach, unless this the problem is different (possible) or the solution is superior (also possible).

It's also a little aberrant that the `monitor` daemon doesn't do the health checks, which is something we should correct. I'm not sure if we should correct it for _this_ patch, but, since Postgres has a prior art that comes from a very old and long-time adequate place (see `aggregate_readings`), I am wondering if the divergence is necessary.

furkansahin:
https://github.com/ubicloud/ubicloud/pull/2298 will fix this

PR #2661, opened 2025-01-22
Add option to specify custom range cidrs for IPV4

mohi-kalantari:
Currently subnets are created with /26 netmask and VMs get a /32 IP.
For kubernetes CNI, we need to have bigger subnets (perferably /16)
so each VM gets a /24 range to host pods and assign an IP from subnet
to them.

In this commit, based on the subnet mask, we always try to assign at most
2**8 ips to each vm. If the subnet mask is 26 min(32, 26+8) would be calculated
which is 32 so each VM gets a /32 ip. but for bigger subnets, VMs would get
at most 256 ips.

### DIFF
--- a/model/private_subnet.rb
+++ b/model/private_subnet.rb
@@ -80,10 +80,18 @@ class PrivateSubnet < Sequel::Model
   # use it in future for some other purpose. AWS also does that. Here
   # is the source;
   # https://docs.aws.amazon.com/vpc/latest/userguide/subnet-sizing.html
+
+  # Requirements:
+  # - The parent subnet mask can range from /16 to /26.
+  # - The VM's assigned subnet must allow:
+  #   - A maximum of 256 IPs (/24) for the largest parent subnet (/16).
+  #   - A minimum of 1 IP (/32) for the smallest parent subnet (/26).
   def random_private_ipv4
-    total_hosts = 2**(32 - net4.netmask.prefix_len) - 5
+    cidr_size = [32, (net4.netmask.prefix_len + 8)].min
+
+    total_hosts = 2**(cidr_size - net4.netmask.prefix_len) - 5
     random_offset = SecureRandom.random_number(total_hosts) + 4
-    addr = net4.nth_subnet(32, random_offset)
+    addr = net4.nth_subnet(cidr_size, random_offset)
     return random_private_ipv4 if nics.any? { |nic| nic.private_ipv4.to_s == addr.to_s }
 
     addr
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -177,11 +177,13 @@ class Prog::Vnet::SubnetNexus < Prog::Base
     selected_addr
   end
 
-  def self.random_private_ipv4(location, project)
+  def self.random_private_ipv4(location, project, cidr_size = 26)
+    raise ArgumentError, "CIDR size must be between 0 and 32" unless cidr_size.between?(0, 32)
+
     private_range = PrivateSubnet.random_subnet
     addr = NetAddr::IPv4Net.parse(private_range)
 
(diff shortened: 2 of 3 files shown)


prog/vm/nexus.rb
```diff
@@ -13,7 +13,8 @@ def self.assemble(public_key, project_id, name: nil, size: "standard-2",
     unix_user: "ubi", location: "hetzner-fsn1", boot_image: Config.default_boot_image_name,
     private_subnet_id: nil, nic_id: nil, storage_volumes: nil, boot_disk_index: 0,
     enable_ip4: false, pool_id: nil, arch: "x64", swap_size_bytes: nil,
-    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [], gpu_count: 0)
+    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [], gpu_count: 0,
+    ipv4_cidr_size: 32)
```
furkansahin, line 17: considering this is specific to kubernetes, let's not add this to VM arguments and leave the sizing to NicNexus. It should allocate a prefix size comparatively to private_subnet. So, if the private_subnet size is /26, Nic allocates a range of `26+6=32`. If the private_subnet size is `18`, the nic allocates a range of `18+6=24`. Does it make sense?

furkansahin (changes requested):

prog/vm/nexus.rb
mohi-kalantari: So I should create the NIC before creating the vm and pass the nic_id as an input arguement?

furkansahin: Or you can still leave it to NicNexus but not pass the argument. NicNexus should find the correct size by looking into the private_subnet's subnet size.

mohi-kalantari: I need NicNexus to add 8 bits. Is that configurable? 6 is less than what we want.

model/private_subnet.rb
```diff
@@ -80,10 +80,18 @@ def self.random_subnet
...
+  # Requirements:
+  # - The parent subnet mask can range from /16 to /26.
+  # - The VM's assigned subnet must allow:
+  #   - A maximum of 256 IPs (/24) for the largest parent subnet (/16).
+  #   - A minimum of 1 IP (/32) for the smallest parent subnet (/26).
   def random_private_ipv4
-    total_hosts = 2**(32 - net4.netmask.prefix_len) - 5
+    cidr_size = [32, (net4.netmask.prefix_len + 8)].min
+
+    total_hosts = 2**(cidr_size - net4.netmask.prefix_len) - 5
```
furkansahin, line 92: this `total_hosts` calculation was based on the fact that we would use at most 59 IP addresses in a /26 subnet. Now, you are allocating /24 subnets from a /16 subnet. You don't need to allocate 4 x /24 subnets for internal usage. This feels odd, we should be able to allocate enough IPs from the selected /24. Think about this and look into how we setup vm networking using these allocated IPs. This may not be a blocker but definitely worths spending some extra time to better understand and trying to solve.

mohi-kalantari: For the /24 ip allocation for each VM, in the CNI we are using two ips per Pod. the default pod number per VM is 110 so it's best for us to reserve 220 ips per vm.

If I understood correctly and you meant sharing these /24 ips with each other, I guess we do not want to do that. 🤔

furkansahin: No, what I meant was, this code, now, eliminates 5 x /24 automatically. It's OK to keep it as is but put a not here so that we come back.

furkansahin (approved):

model/private_subnet.rb
mohi-kalantari: oh now I understood. I'll add a task to come back and fix it later

PR #2662, opened 2025-01-22
Add ipv4 routes for cross vm packets

mohi-kalantari:
So far, VMs had a /32 ip range and VMs did not need to route packets for other entities with ip inside. For Kubernetes CNI, we will have pods which need to send packets out of the VM to another destination in the same private subnet. In that case we need to add routes to the VM's network namespace.


rhizome/host/lib/vm_setup.rb
```diff
@@ -335,6 +338,18 @@ def routes4(ip4, ip4_local, nics)
     end
   end
 
+  def update_via_routes(nics)
```
furkansahin, line 341: I don't think this is a good idea. If something goes wrong and the tap device is not coming up, our SSH connection will timeout and potentially cause apoptosis. Define a safe counter limit and early exit.

mohi-kalantari: Fixed it thanks.

fdr: Hmm the apoptosis probably only applies to things that have no timeouts. In-session timeouts are a bit trickier, e.g. running `ls` and not getting a response for ages, vs not able to negotiate the connection at all.

furkansahin (approved):

PR #2663, opened 2025-01-22
Decrease neg-ttl in dnsmasq config

mohi-kalantari:
Default neg-ttl is 300 seconds which is a lot to wait for retrying dns queries. For bootstrapping kubernetes, we have 4 minute timeout to discover the apiserver through the loadbalancer and DNS and if the first query fails, the bootstrap will fail and way more complex measurements should be added to fix that.

In this commit the neg-ttl is set to 10 seconds which is reasonable for retrying failed dns queries

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -484,6 +484,7 @@ dns-forward-max=10000
 dhcp-option=option6:dns-server,#{dnsmasq_address_ip6}
 listen-address=#{dnsmasq_address_ip6}
 all-servers
+neg-ttl=30
 DNSMASQ_CONF
 
     ethernets = nics.map do |nic|


furkansahin:
Considering this setting is applied to all of our VMs, and we want to be a good neighbour to our DNS providers, maybe 10 seconds is too low. Let's at least set it to a minute?

mohi-kalantari:
> Considering this setting is applied to all of our VMs, and we want to be a good neighbour to our DNS providers, maybe 10 seconds is too low. Let's at least set it to a minute?

I think it would be a bit risky. assuming that the actual bootstrap process takes 2 minutes and a half (my rough estimation based on numerous times I did the bootstrapping), we only have 1 retry left. I can't really do the math to prove my point but maybe 30 seconds? 

@furkansahin

furkansahin:
Alright, 30 secs it is

furkansahin (approved):

PR #2664, opened 2025-01-22
Remove automatically created routes after creating ipv6 link

mohi-kalantari:
When we create a new linux interface using ip link, a corresponding linux route is also created.

For IPV6, we create a new link in the VM's network namespace and a route is automatically added. Since that route does not route the packets to the gateway, containers inside with IPV6 ips would fail.

In this commit, we are deleting that route and rely on the route with via option.

to route the packets but that automatically created route will assume the destination is in the same subnet and ipv6 wouldn't work properly


mohi-kalantari:
how can I trigger the E2E tests?
@furkansahin

enescakir:
> how can I trigger the E2E tests? @furkansahin

Select your branch and run workflow. Since we have a single host to run E2E, it runs e2e workflow requests one by one. fyi

https://github.com/ubicloud/ubicloud/actions/workflows/e2e.yml

CleanShot 2025-01-23 at 21 18 56@2x

furkansahin:
🚢

PR #2666, opened 2025-01-23
Create a Standard Vm in a slice

macieksarnowicz:
Using the VmHostSlice introduced earlier to allocate a Standard VM. Slices allocate CPU and memory on a host and isolate it from other slices. VMs are then placed in a slice. We are starting with Standard VMs mapped 1:1 to a slice hosting them. The CPU and memory of a slice are fully dedicated to the VM. The main advantage is that CPUs are affinitized and not shared with any other slice and VMs they host. 

The VMs-in-slices are hosted on separate hosts from standalone VMs (the current code). Two flags guard this logic:
* VmHost.accept_slices - indicates host instances accepting VMs allocated in slices. Hosts with this flag set to `true` will not be used for standalone VMs
* Project feature flag - `use_slices_for_allocation` - only projects with this flag enabled will use the new logic of creating VMs in slices.
With those two flags, we can control the rollout of this feature. Note that if a project is marked with `use_slices_for_allocation` but there are no VmHosts that can accept_slices, the user will see a 'no capacity' error.

The allocator is modified to use a new VmHostSliceAllocation class that wraps and replaces VmHostAllocation when slice logic is enabled. It finds the capacity the same way as before, but when `update` is called, it creates a slice first, together with allocating VmHostCpus before starting the VM creation process. This building block will be expanded upon when we introduce 'burstable' instances later when multiple VMs can share a slice.

There is a commit of diagnostic code for the Allocator, helping troubleshoot the allocator query.


model/vm.rb
```diff
@@ -102,10 +96,11 @@ def cloud_hypervisor_cpu_topology
     total_packages = vm_host.total_sockets
 
     # Computed all-system statistics, now scale it down to meet VM needs.
-    proportion = Rational(cores) / vm_host.total_cores
```
fdr, line 105: btw the previous code was already weird (@pykello knows) with systems without uniform threads per core (E core and P cores on GEX), are there any interesting problems here for that?

prog/vm/nexus.rb
```diff
@@ -363,13 +397,13 @@ def write_params_json
 
   label def restart
     decr_restart
-    host.sshable.cmd("sudo systemctl restart #{vm.inhost_name}")
+    host.sshable.cmd("sudo systemctl restart #{q_vm}")
```
fdr, line 400: a spurious change? Should be broken out and with comment on why the alteration?

scheduling/allocator.rb
```diff
@@ -117,6 +156,15 @@ def self.candidate_hosts(request)
       ds = ds.exclude(Sequel[:vm_host][:id] => request.host_exclusion_filter) unless request.host_exclusion_filter.empty?
       ds = ds.where(location: request.location_filter) unless request.location_filter.empty?
       ds = ds.where(allocation_state: request.allocation_state_filter) unless request.allocation_state_filter.empty?
+      # Match the slice allocation to the hosts that can accept it
+      ds = ds.where(accepts_slices: request.use_slices)
+
+      # For debugging purposes, write the full SQL query, with expanded parameters, to a file,
+      # so it can be run directly against the DB server
+      # :nocov:
+      File.write(File.expand_path("~/allocator.sql"), ds.prepare(:select, :allocator_query).sql) if request.enable_diagnostics
```
fdr, line 165: This seems like weird code to commit, although I see the utility. @jeremyevans do you have a recommendation?

scheduling/allocator.rb
```diff
@@ -38,8 +41,27 @@ def self.allocate(vm, storage_volumes, distinct_storage_devices: false, gpu_coun
     Clog.emit("vm allocated") { {allocation: allocation.to_s, duration: Time.now - vm.created_at} }
   end
 
-  Request = Struct.new(:vm_id, :cores, :mem_gib, :storage_gib, :storage_volumes, :boot_image, :distinct_storage_devices, :gpu_count, :ip4_enabled,
-    :target_host_utilization, :arch_filter, :allocation_state_filter, :host_filter, :host_exclusion_filter, :location_filter, :location_preference)
+  Request = Struct.new(:vm_id, :family, :cores, :memory_gib, :storage_gib, :storage_volumes, :boot_image, :distinct_storage_devices, :gpu_count, :ip4_enabled,
+    :target_host_utilization, :arch_filter, :allocation_state_filter, :host_filter, :host_exclusion_filter, :location_filter, :location_preference,
+    :use_slices, :enable_diagnostics) do
```
fdr, line 46: This kind of diff make sit easy to miss  what actually happened (are there multiple?!): it seems a diagnostics feature is in this patch as well.

prog/vm/nexus.rb
```diff
@@ -466,8 +508,31 @@ def write_params_json
...
+    # Trigger the slice deletion if this is a
+    # dedicated slice for this VM or if this is
+    # a shared slice and there are no other VMs.
+    # We do not need to wait for this to complete
+    #
+    # We disable the slice to prevent another
+    # concurrent VM allocation from grabbing it
+    # while it is being destroyed
+    unless slice.nil?
+      slice.reload
```
fdr, line 529: why?

prog/vm/nexus.rb
```diff
@@ -363,13 +397,13 @@ def write_params_json
...
     decr_restart
-    host.sshable.cmd("sudo systemctl restart #{vm.inhost_name}")
+    host.sshable.cmd("sudo systemctl restart #{q_vm}")
     hop_wait
   end
 
   label def stopped
     when_stop_set? do
-      host.sshable.cmd("sudo systemctl stop #{vm.inhost_name}")
+      host.sshable.cmd("sudo systemctl stop #{q_vm}")
```
fdr, line 406: if there's a good reason for this, also seen in a few other places, let's separate it, it reduces my precision paging the other hunks

prog/vm/nexus.rb
```diff
@@ -225,6 +239,26 @@ def before_run
...
+        nap 1
+      else
+        # Once the slice is created update its utilization
+        # We make sure the slice is still enabled to avoid a race condition in the db
+        updated = VmHostSlice.dataset.where(id: vm.vm_host_slice_id, enabled: true).update(
+          used_cpu_percent: Sequel[:used_cpu_percent] + vm.cpu_percent_limit,
+          used_memory_gib: Sequel[:used_memory_gib] + vm.memory_gib
+        )
+
+        fail "failed to update slice utilization" unless updated == 1
```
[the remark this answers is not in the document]
pykello: Also, assuming we move the update command & its corresponding check from here to allocator, issues like following are still possible:

* Assume there's a shared slice S with one burstable A in it.
* Burstable B's `start` starts a new `transaction 1`: Burstable B is allocated into S & increases `used_cpu_percent` & `used_memory_gib` successfully. Because enabled is still true & slice has enough capacity.
* Burstable A's `destroy_slice` starts a new `transaction 2`
   * It destroys burstable A
   * Since `transaction 1` hasn't committed yet, `slice.is_shared && slice.vms.count == 0` is true; and `slice.update(enabled: false)` & `slice.incr_destroy` are executed (note that postgres executes transactions in `READ COMMITTED` mode by default)
* `transaction 2` commits successfully.
* `transaction 1` commits successfully.

Now, slice nexus will try to delete the slice but burstable B is in the slice.

model/vm.rb
macieksarnowicz: I was debating if to include this in the current PR or wait for the one that will bring the Burstable family. It is not per-se yet needed here, but it will be needed, so I wanted to have a discussion started. 

Answering your question:
When we are going to introduce an ability to overcommit a slice, meaning to put more vpcus on it than what would be expected from the calculation of 'cores / cpus_to_cores', we should calculate the topology starting from the vcpus. The number of 'cores' on the VM will indicate a size of a slice to allocate the VM in, while the number of vcpus will indicate the vcpu allocation to the VM. 
This also be needed when we introduce burstable-1, with 1 vcpu, which will require some more changes to this method.

prog/vm/nexus.rb
macieksarnowicz: Just clean up, some methods in this class use q_vm, while others use vm.inhost_name. I think q_vm is the correct to use here. I can remove this for now, and add later as a separate commit.

scheduling/allocator.rb
macieksarnowicz: I knew this will be discussed :) I created this early on in the prototyping phase when I needed to run this query on the server 'by hand' and make changes to it which later I'd translate to Ruby. The big query is not yet much changed, but it will be touched more when we add slice selection for Burstables. 
I thought you all may find this useful for troubleshooting, even in production, that's why I used a feature flag to control it. What I do not know is:
- do you have a precedent of this kind of diagnostic code in production code? I could not find any
- what is the recommended way of controlling this? Is it ok to use a per-project feature flag?
- should we define a better place to put the output file in?
- right now the file just gets overwritten every time this runs. Should I add some logic to keep the file in place once written? That was Ok for running in dev environment, but in production may not be ideal.

scheduling/allocator.rb
macieksarnowicz: It shows better in VS Code than in Git. What's changed is:
- mem_gib -> memory_gib
- added :use_slices and :enable_diagnostics

Would you prefer to pull out the diagnostics code into a separate commit in this PR or into a separate PR altogether?

prog/vm/nexus.rb
macieksarnowicz: Uhmmm... good catch. I think this was my misunderstanding of how Proc code works early in the project, and I missed catching it now before sending the PR. I will clean this up.

prog/vm/nexus.rb
macieksarnowicz: OK :)

prog/vm/nexus.rb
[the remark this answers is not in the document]
macieksarnowicz: Back to the drawing board :)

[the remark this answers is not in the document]
macieksarnowicz: I have been pondering this a bit. But first, let's clear up some understanding of READ COMMITTED isolation level. In that Tx mode, just having two update statements updating the same value will not automatically trigger a conflict. You have to have a WHERE clause that prevents one of them from happening if the other was committed during the first's transaction. In the READ COMMITTED, the second update will re-read the values on running the UPDATE and we can sue that to prevent the UPDATE. 
On the other hand, if we use READ REPEATABLE, then with two updates on the same row, the second one will fail and roll back. 

So, I think we have two options:
1. have a set of mutually exclusive update statements:
    on create - `UPDATE vm_host_slice SET used_cpu_percent = ... WHERE id = ... AND enabled = true;`
    on destroy 'UPDATE vm_host_slice SET enabled = false WHERE id = ... AND used_cpu_percent = 0;`

2. use READ REPEATABLE isolation level for some of those transactions.
    
I would like to hear what you guys think about #2 and if there is some code already using that? 

How is this handled so far for vm_host capacity updates? It is a similar problem there with multiple VMs being created and dropped on the same host.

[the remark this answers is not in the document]
pykello: > How is this handled so far for vm_host capacity updates? It is a similar problem there with multiple VMs being created and dropped on the same host.

Pretty much based on what you described for READ COMMITTED + using the constraints on the vm_host table.

For VM host update statements we issue a Ruby statement like:

```
VmHost.dataset.where(id: vm_host.id).update({
  :used_cores => Sequel[:used_cores] + requested_cores,
  :used_hugepages_1g => Sequel[:used_hugepages_1g] + requested_hugepages
})
```

which will issue a postgres query like:

```
UPDATE "vm_host"
SET "used_cores" = ("used_cores" + $1::int4), 
    "used_hugepages_1g" = ("used_hugepages_1g" + $2::int4)
WHERE ("id" = $3);
```

Postgres automatically acquires locks when updating the same rows, so if 2 transactions try updating the same vm_host row, one of them will block until the 2nd one commit. So, the 2nd commit will see the updated `used_cores` & `used_hugepages_1g` values & increase will be done correctly.

And since we have constraints like `"core_allocation_limit" CHECK (used_cores <= total_cores)`, if any update causes over-allocation, allocation transaction will fail and Clover will retry the label.

Note that if instead of reading `used_cores` in the UPDATE statement we had read the used_cores & used_hugepages in a Ruby variable and did something like:

```
used_cores = vm_host.used_cores
used_hugepages = vm_host.used_hugepages_1g

VmHost.dataset.where(id: vm_host.id).update({
  :used_cores => used_cores + requested_cores,
  :used_hugepages_1g => used_hugepages + requested_hugepages
})
```

Then we would have a concurrency problem because the read would have been done before the update statement which does the locking.

> have a set of mutually exclusive update statements:
on create - UPDATE vm_host_slice SET used_cpu_percent = ... WHERE id = ... AND enabled = true;
on destroy 'UPDATE vm_host_slice SET enabled = false WHERE id = ... AND used_cpu_percent = 0;`

I think this should work and looks more sensible than what I suggested in the other PR. I'll think more by Monday.

> use READ REPEATABLE isolation level for some of those transactions.
I would like to hear what you guys think about https://github.com/ubicloud/ubicloud/pull/2 and if there is some code already using that?

We're not doing that at the moment. I talked with Daniel about changing the isolation level long time ago, but we never considered it seriously. I think if we want to do it, it might be a bigger project than the scope of this PR, so I prefer to use alternative solutions. If we need to, we can think about it as a separate change. @fdr thoughts?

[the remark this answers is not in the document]
jeremyevans: I think it's fine to use higher isolation levels if the situation calls for it.  Sequel supports setting the isolation level for each transaction, and we do use that in at least one place already.

migrate/20250123_vm_add_cpulimits.rb
jeremyevans, line 1: Please give this migration a later version (e.g. 20250124).  With the same version, this migration would be executed before the previous migration if both are available migration candidates.  Note that migration versions are arbitrary, there is no requirement it be the date the migration was created.

model/vm.rb
```diff
@@ -220,7 +202,10 @@ def params_json(swap_size_bytes)
       "ndp_needed" => vm_host.ndp_needed,
       "storage_volumes" => storage_volumes,
       "swap_size_bytes" => swap_size_bytes,
-      "pci_devices" => pci_devices.map { [_1.slot, _1.iommu_group] }
+      "pci_devices" => pci_devices.map { [_1.slot, _1.iommu_group] },
+      "slice_name" => vm_host_slice.nil? ? "system.slice" : vm_host_slice.inhost_name,
+      "cpu_percent_limit" => cpu_percent_limit.nil? ? 0 : cpu_percent_limit,
+      "cpu_burst_percent_limit" => cpu_burst_percent_limit.nil? ? 0 : cpu_burst_percent_limit
```
jeremyevans, line 208:
```suggestion
      "slice_name" => vm_host_slice&.inhost_name || "system.slice",
      "cpu_percent_limit" => cpu_percent_limit || 0,
      "cpu_burst_percent_limit" => cpu_burst_percent_limit || 0
```

Same, just more idiomatic.

prog/vm/nexus.rb
```diff
@@ -225,6 +239,26 @@ def before_run
 
     host.sshable.cmd(command)
 
+    hop_wait_for_slice
+  end
+
+  label def wait_for_slice
+    unless vm.vm_host_slice.nil?
```
jeremyevans, line 246:
```suggestion
    if vm.vm_host_slice
```

Unless the value can be `false`, it's better to use it directly as opposed to `nil?`

prog/vm/nexus.rb
```diff
@@ -225,6 +239,26 @@ def before_run
...
+      if !vm.vm_host_slice.enabled
```
jeremyevans, line 247: Can we switch this to `if vm.vm_host_slice.enabled` and reverse the order of the branches?

prog/vm/nexus.rb
```diff
@@ -466,8 +508,31 @@ def write_params_json
...
 
+    # Trigger the slice deletion if this is a
+    # dedicated slice for this VM or if this is
+    # a shared slice and there are no other VMs.
+    # We do not need to wait for this to complete
+    #
+    # We disable the slice to prevent another
+    # concurrent VM allocation from grabbing it
+    # while it is being destroyed
+    unless slice.nil?
```
jeremyevans, line 528:
```suggestion
    if slice
```

scheduling/allocator.rb
```diff
@@ -210,6 +264,69 @@ def utilization
...
+  # for a place for a new slice and select a cpu set that will be used for that.
+  class VmHostSliceAllocation
+    def initialize(candidate_host, request, vm_host_allocations)
+      @candidate_host = candidate_host
+      @request = request
+      @vm_host_allocations = vm_host_allocations
+    end
+
+    def is_valid
+      @vm_host_allocations.all? { _1.is_valid }
```
jeremyevans, line 286:
```suggestion
      @vm_host_allocations.all?(&:is_valid)
```

scheduling/allocator.rb
```diff
@@ -210,6 +264,69 @@ def utilization
...
+    end
+
+    def utilization
+      # compute the score based on combined CPU and Memory utilization, as usual
+      util = @vm_host_allocations.map { _1.utilization }
```
jeremyevans, line 291:
```suggestion
      util = @vm_host_allocations.map(&:utilization)
```

scheduling/allocator.rb
```diff
@@ -210,6 +264,69 @@ def utilization
...
+      util.sum.fdiv(util.size)
+    end
+
+    def update(vm, vm_host)
+      DB.transaction do
+        # Update the host utilization
+        VmHost.dataset.where(id: vm_host.id).update(@vm_host_allocations.map { _1.get_vm_host_update }.reduce(&:merge))
```
jeremyevans, line 298:
```suggestion
        VmHost.dataset.where(id: vm_host.id).update(@vm_host_allocations.map(&:get_vm_host_update).reduce(&:merge))
```

scheduling/allocator.rb
macieksarnowicz: This is updated now and I pulled it out to a separate commit for clarity.

prog/vm/nexus.rb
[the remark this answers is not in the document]
macieksarnowicz: @pykello - take a look at this new version

migrate/20250123_vm_add_cpulimits.rb
macieksarnowicz: OK, I will update the versions right before we merge this, after all the discussions are resolved.

scheduling/allocator.rb
macieksarnowicz: I reformatted this whole thing. Does not help with reviewing this change, but it will help with any subsequent one. I also pulled out the diagnostics change into a separate commit, to make it more clear.

model/vm.rb
macieksarnowicz: OK, done.

scheduling/allocator.rb
macieksarnowicz: Just for my education - what is the difference?

prog/vm/nexus.rb
```diff
@@ -466,8 +499,36 @@ def write_params_json
...
+    # Trigger the slice deletion if this is a
+    # dedicated slice for this VM or if this is
+    # a shared slice but not used by any other VM.
+    # We do not need to wait for this to complete
+    #
+    # We disable the slice to prevent another
+    # concurrent VM allocation from grabbing it
+    # while it is being destroyed
+    if slice
+      # :nocov:
```
macieksarnowicz, line 520: Looking for suggestion on how to add mocking for Sequel .& and .| operators. I could not find a way, so for now put the :nocov: around it, but can update it.

scheduling/allocator.rb
jeremyevans: It's more idiomatic.  `{ _1.method }` and `&:method` are more or less the same, but if you are calling a method on the block argument, prefer `&:method`.  Use a block if you need to pass an argument or use the block argument in some other manner: `{ method(_1) }` or `{ _1.method(arg) }`

prog/vm/nexus.rb
jeremyevans: Can you explain why you would need to mock them?  Note that you likely could mock them just like any other method.  I'm not sure why these lines wouldn't be covered without any mocking.

prog/vm/nexus.rb
```diff
@@ -466,8 +499,36 @@ def write_params_json
...
+      updated = VmHostSlice.dataset
+        .where(id: slice.id)
```
jeremyevans, line 522:
```suggestion
      updated = slice.this
```

Since you already have the slice.

prog/vm/nexus.rb
macieksarnowicz: The nexus_spec.rb is mostly mocked, as is most of Prog tests, since they require respirate to run otherwise. 

The test for this code is:
```
    it "#destroy_slice with a slice" do
      vm_host_slice = instance_double(VmHostSlice, id: "9d487886-d167-4d00-8787-a746be0d4d9a")
      expect(vm).to receive(:vm_host_slice).and_return(vm_host_slice)
      expect(vm_host_slice).to receive(:incr_destroy)
      expect(vm).to receive(:destroy).and_return(true)

      vhs_dataset = instance_double(VmHostSlice.dataset.class)
      expect(VmHostSlice).to receive_messages(dataset: vhs_dataset)
      expect(vhs_dataset).to receive(:where).and_return(vhs_dataset).twice
      expect(vhs_dataset).to receive(:update).with(enabled: false).and_return(1)

      expect { nx.destroy_slice }.to exit({"msg" => "vm deleted"})
    end
```

I am already mocking the `:where` and `:update' dataset methods. However, with this code, if I remove the :nocov: it leaves this line without code coverage:
`.where { Sequel.|({enabled: true}, Sequel.&({used_cpu_percent: 0}, {used_memory_gib: 0})) }`
I am assuming that is because the Sequel methods are not mocked, but maybe I am wrong on this.

jeremyevans: Your mocking isn't yielding to the block passed to `where`.  It looks like your are using a virtual row block when you don't need to be.  Change to `.where(Sequel.|({enabled: true}, Sequel.&({used_cpu_percent: 0}, {used_memory_gib: 0})))`.  Only use a virtual row block if you are using features it provides (such as method_missing creating Sequel identifiers/functions).

macieksarnowicz: ugh! thank you. I should have caught that. I get tripped by {} in a code like this with a lot of different brackets.

scheduling/allocator.rb
pykello: > I thought you all may find this useful for troubleshooting, even in production

If this is going to be used in production, then we should use `Clog.emit(...)` instead of writing to filesystem. We use multiple heroku dynos for respirate, and as far as I know each of them gets its own ephemeral filesystem & figuring out where we should find `~/allocator.sql` is not trivial.

pykello:
@bsatzger could you please review the allocator changes?

prog/vm/nexus.rb
```diff
@@ -466,8 +499,33 @@ def write_params_json
...
+    # dedicated slice for this VM or if this is
+    # a shared slice but not used by any other VM.
+    # We do not need to wait for this to complete
+    #
+    # We disable the slice to prevent another
+    # concurrent VM allocation from grabbing it
+    # while it is being destroyed
+    if slice
+      updated = slice.this
+        .where(Sequel.|({enabled: true}, Sequel.&({used_cpu_percent: 0}, {used_memory_gib: 0})))
```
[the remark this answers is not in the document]
jeremyevans: That doesn't result in the same condition:

```ruby
Sequel.|({enabled: true}, Sequel.&({used_cpu_percent: 0}, {used_memory_gib: 0})
# ((enabled IS TRUE) OR ((used_cpu_percent = 0) AND (used_memory_gib = 0)))

Sequel[enabled: true, used_cpu_percent: 0, used_memory_gib: 0]
# ((enabled IS TRUE) AND (used_cpu_percent = 0) AND (used_memory_gib = 0))
```

I'm not sure which condition is correct, but the current mixing of `OR` and `AND` seems deliberate.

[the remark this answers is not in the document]
fdr: @jeremyevans really? to truncate/overwrite some random file in your home directory? furthermore, it cannot be useful in production (we don't have access to their file systems, there's a lot going on and it's going to always truncate the file anyway)

IMO: eliminate it from this patch, produce a separate one, but the file system mechanism looks damaged beyond repair to me.

scheduling/allocator.rb
fdr: @jeremyevans really? to truncate/overwrite some random file in your home directory? furthermore, it cannot be useful in production (we don't have access to their file systems, there's a lot going on and it's going to always truncate the file anyway)

IMO: eliminate it from this patch, produce a separate one, but the file system mechanism looks damaged beyond repair to me.

macieksarnowicz: I changed it to Clog.emit and tried with pry. It is controlled by a flag, so easy to enable/disable.
I have no insight into your prod environment, so sorry for using the file operation. This was intended as a proposal up for a discussion.

@fdr - a while ago you mentioned some diagnostics for this query would be useful. If you still think this is not helpful, I can remove it.

scheduling/allocator.rb
```diff
@@ -210,6 +273,75 @@ def utilization
...
       {@column => Sequel[@column] + @requested}
     end
+
+    def update(vm, vm_host)
+      VmHost.dataset.where(id: vm_host.id).update([get_vm_host_update].reduce(&:merge))
+    end
+  end
+
+  # The VmHostSliceAllocation is used when system is configured
+  # to allocate VMs inside VmHostSlice. It wrapps around the VmHostAllocation class
```
bsatzger, line 292:
```suggestion
  # to allocate VMs inside VmHostSlice. It wraps around the VmHostAllocation class
```

scheduling/allocator.rb
macieksarnowicz: I discussed this with @fdr and the conclusion is that we can proceed with this commit.

pykello:
@macieksarnowicz I think we should also add the code to start slices after reboot. I triggered E2E tests on a copy of this branch and it failed when doing tests after reboot.

I haven't looked deep yet, but this is the run output: https://github.com/ubicloud/ubicloud/actions/runs/13000779417/job/36258809262

pykello:
> @macieksarnowicz I think we should also add the code to start slices after reboot. I triggered E2E tests on a copy of this branch and it failed when doing tests after reboot.
> 
> I haven't looked deep yet, but this is the run output: https://github.com/ubicloud/ubicloud/actions/runs/13000779417/job/36258809262

Re-run succeeded & @macieksarnowicz said restart after reboot for slices shouldn't matter until burstables. Log for the 2nd run: https://github.com/ubicloud/ubicloud/actions/runs/13017172107

Then I'll just do few more manual tests before merging.

macieksarnowicz:
> > @macieksarnowicz I think we should also add the code to start slices after reboot. I triggered E2E tests on a copy of this branch and it failed when doing tests after reboot.
> > I haven't looked deep yet, but this is the run output: https://github.com/ubicloud/ubicloud/actions/runs/13000779417/job/36258809262
> 
> Re-run succeeded & @macieksarnowicz said restart after reboot for slices shouldn't matter until burstables. Log for the 2nd run: https://github.com/ubicloud/ubicloud/actions/runs/13017172107
> 
> Then I'll just do few more manual tests before merging.

Just to clarify - the code to set up slice correctly after the host reboot is already in place. What I meant is that there is some code change needed when VM is restarted, but that is only applicable to Burstables, and I will bring it with the next PR. Having said all that, the E2E tests are not yet running the 'slice' version of the code, so none of this matters. 
I have run E2E tests also on my own and they passed. I will check with @pykello how the runs look after this merges.

lib/option.rb
```diff
@@ -51,17 +51,17 @@ def self.postgres_locations
...
     storage_size_options = [_1 * 20, _1 * 40]
-    VmSize.new("standard-#{_1}", "standard", _1 / 2, _1, _1 * 4, storage_size_options, NO_IO_LIMITS, true, false, "x64")
+    VmSize.new("standard-#{_1}", "standard", _1 / 2, _1, _1 * 100, 0, _1 * 4, storage_size_options, NO_IO_LIMITS, true, false, "x64")
   }.concat([2, 4, 8, 16, 30, 60].map {
     storage_size_options = [_1 * 20, _1 * 40]
-    VmSize.new("standard-#{_1}", "standard", _1, _1, (_1 * 3.2).to_i, storage_size_options, NO_IO_LIMITS, false, false, "arm64")
+    VmSize.new("standard-#{_1}", "standard", _1, _1, _1 * 100, 0, (_1 * 3.2).to_i, storage_size_options, NO_IO_LIMITS, false, false, "arm64")
   }).concat([6].map {
-    VmSize.new("standard-gpu-#{_1}", "standard-gpu", _1 / 2, _1, (_1 * 5.34).to_i, [_1 * 30], NO_IO_LIMITS, false, true, "x64")
+    VmSize.new("standard-gpu-#{_1}", "standard-gpu", _1 / 2, _1, _1 * 100, _1, (_1 * 5.34).to_i, [_1 * 30], NO_IO_LIMITS, false, true, "x64")
```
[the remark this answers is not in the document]
macieksarnowicz: Thanks! Yes, it should be. Fixed now and I also added some tests to cover for this.

pykello (approved):

PR #2670, opened 2025-01-23
Enable prefix caching on Llama 3.2 3B model

0xjunhao:
Prefix caching allows for significant performance improvements when processing inputs that share a common prefix, which is common in conversation and multi-step tasks.

Here's the benchmark for Llama 3.2 3B with and without prefix caching: https://docs.google.com/spreadsheets/d/16gi4veItLB6UH9FppWjKIq10y-v-I8tp1vhlMsVuMVE/edit?usp=sharing

We can see that:
1. Enabling prefix caching has no noticeable performance impact when there are no shared prefixes but achieves several times higher throughput when there are.
2. VLLM automatically manages the normal KV cache and prefix cache using preallocated GPU memory.


bsatzger (approved):
👍

PR #2671, opened 2025-01-23
Add a CLI named ubi, using a thin-CLI approach

jeremyevans:
Maintaining a traditional CLI program for a web service can be
difficult, since any new CLI features are not available to users
unless the user updates their CLI program.  That defeats some of
the advantages of using a web service to being with, where new
features are immediately available for users to use when they
are deployed.

To avoid this issue, this uses an approach I'm dubbing thin-CLI,
where the CLI program is as a thin as possible.  With thin-CLI,
the CLI submits the argv to the web service, and the web service
parses the argv and takes appropriate action.  This allows the
CLI to support most new features as soon as they are deployed,
without the user having to update the CLI program.

See individual commits for more details.  Currently, the following
CLI commands are supported:

* vm list
* vm ssh
* vm sftp
* vm scp


fdr (approved):
I kinda wonder if:

```
  vm ssh location vm-name -A --
  vm sftp location vm-name -A
  vm scp location vm-name local-path :remote-path -A
```

should become:

```

  vm location vm-name ssh -A
  vm location vm-name sftp -A
  vm location vm-name scp local-path :remote-path -A
```

The general thinking I'm having on the reordering is how to toggle from one functionality to another without doing a lot of seeking around on the terminal line, e.g. to make `C-w` (backward-delete-word) more convenient.

We may want to think about how location is treated altogether: whether the radical proposal to make naming things global, though the ramifications for a multi-region system are severe, or we make it an option that can be applied to many commands with more relaxed ordering constraint, or something else. 

Regardless, let's get it in and be a bit more wild with this one and experiment. @geemus will be interested. Probably both in tweaking the feel, but also, the mechanism altogether, as he had an independent interest in the thin-cli approach.

Also, thank you for the excellent commit messages.

enescakir:
It looks good to me overall. It's an elegant approach for developing a CLI. We might turn `Rodish` into a separate package once it's stabilized.

I didn't review the code in detail; I just tested the CLI. I have some feedback on the UX. I think the default parameters should be for newbie users not advanced users.

CleanShot 2025-02-05 at 11 03 39@2x

- We might show headers by default.
- The widths of the headers and columns don’t match.  
- IMHO, adding a new parameter for each column isn’t very manageable. Instead of `ubi vm list --name --ip4 --ip6`, I think `ubi vm list --fields name,ip4,ip6` is a more common way to do it. Of course, we can change `fields` to something else.  
-  We could improve the error messages to indicate which parameters are missing.

fdr:
I like enes's proposal of a more compact field-projection notation, if we want a field projection thing.

jeremyevans:
> I kinda wonder if:
> 
> ```
>   vm ssh location vm-name -A --
>   vm sftp location vm-name -A
>   vm scp location vm-name local-path :remote-path -A
> ```
> 
> should become:
> 
> ```
> 
>   vm location vm-name ssh -A
>   vm location vm-name sftp -A
>   vm location vm-name scp local-path :remote-path -A
> ```

This requires changes to Rodish.  It currently supports both subcommands and an action for the current command, but there isn't currently the ability to restart the cli parsing once an action for a command is taken.  I don't think the changes required are large, probably just passing the subcommands to the `run` block as an additional parameter, and changing before blocks to be immediately executed when a subcommand is taken, instead of delayed until an action is taken.

I'm not sure if I've seen that style (arguments before subcommands) in other CLIs.  It's more similar to a routing tree style, and I could see it offering similar benefits as using a routing tree for http routes.  I'm certainly not opposed to it if you want to use that approach.

> We may want to think about how location is treated altogether: whether the radical proposal to make naming things global, though the ramifications for a multi-region system are severe, or we make it an option that can be applied to many commands with more relaxed ordering constraint, or something else.

One idea is to combine things in one argument for now `location/vm-name`.  Currently, the `location/` part would be required, but we could potentially loosen that later.  With the current two argument approach, dropping the location argument would not be possible in a backwards compatible manner.

> Regardless, let's get it in and be a bit more wild with this one and experiment. @geemus will be interested. Probably both in tweaking the feel, but also, the mechanism altogether, as he had an independent interest in the thin-cli approach.

Agreed, best to experiment with this in-tree.  I'll merge this in a bit. I can then work on a couple PRs for the changes discussed above, as well as PRs to address @enescakir's comments.

PR #2672, opened 2025-01-23
Add free inference tokens field

0xjunhao:


enescakir:
We have already implemented that future in https://github.com/ubicloud/ubicloud/pull/2818

PR #2673, opened 2025-01-24
Improve thread dump reporting

fdr:
Improve thread dump reporting

It's been a while since this has been revised.  This revision seeks to
make the following easier:

It's hard to find the beginning and end of thread dumps, particularly
if they are consecutive.  Solution: add `--BEGIN...` and `--END...`
framing, plus a disambiguating timestamp that matches between the two.
The rough inspiration is MIME format of email.

The thread dumps are often triggered by apoptosis, which needs to exit
the entire process in a timely manner to maintain mutual exclusion.
In doing so, it does collatoral damage to other threads that happen to
be running in the same process at the time.

The thread dump emitted likewise includes every thread, not just the
problematic one.  So, print the thread age, when available.  The
oldest are most likely to be implicated.

To make the thread age available in the case I care about most, the
`created_at` thread local variable is set in the dispatcher which runs
Strands.

### DIFF
--- a/lib/thread_printer.rb
+++ b/lib/thread_printer.rb
@@ -2,9 +2,17 @@
 
 module ThreadPrinter
   def self.run
+    now = Time.now
+    puts "--BEGIN THREAD DUMP, #{now}"
     Thread.list.each do |thread|
       puts "Thread: #{thread.inspect}"
+
+      if (created_at = thread[:created_at])
+        puts "Created at: #{created_at}, #{now - created_at} ago"
+      end
+
       puts thread.backtrace&.join("\n")
     end
+    puts "--END THREAD DUMP, #{now}"
   end
 end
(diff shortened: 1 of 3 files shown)


lib/thread_printer.rb
```diff
@@ -2,9 +2,17 @@
 
 module ThreadPrinter
```
enescakir, line 3: How can we find these logs in Mezmo? Should we print them using Clog?

fdr: Mezmo has stdout and stderr. This is one of the things I like about it: as opposed to using a tool that insists on API requests from within the process, we can catch weird stuff, like stderr messages in libraries or ruby's core dump text. Sure, it's ugly (and so is this), but you can get at it.

 The general idea was this was a similar kind of dump we want to get no matter what, it's intended to be exceptional. That's also why I don't frame it as JSON (also because: I _have_ framed it with JSON in the past only to find, reasonably, very large JSON messages with full backtraces are rejected)

enescakir (approved):

jeremyevans (approved):

PR #2675, opened 2025-01-24
MinIO Client cert chain update

furkansahin:
Some of the devs (including me) found out that the ca_bundle that doesn't have the server side cert fails to get validate the certificate. Here, we start passing the minio server cert together with the root certs.

### DIFF
--- a/model/minio/minio_server.rb
+++ b/model/minio/minio_server.rb
@@ -85,7 +85,7 @@ class MinioServer < Sequel::Model
       endpoint: server_url,
       access_key: cluster.admin_user,
       secret_key: cluster.admin_password,
-      ssl_ca_file_data: cluster.root_certs,
+      ssl_ca_file_data: cluster.root_certs + cert,
       socket: socket
     )
   end
--- a/spec/model/minio/minio_server_spec.rb
+++ b/spec/model/minio/minio_server_spec.rb
@@ -26,7 +26,8 @@ RSpec.describe MinioServer do
     described_class.create_with_id(
       minio_pool_id: mp.id,
       vm_id: vm.id,
-      index: 0
+      index: 0,
+      cert: "cert"
     )
   }
 
--- a/spec/prog/minio/minio_server_nexus_spec.rb
+++ b/spec/prog/minio/minio_server_nexus_spec.rb
@@ -424,6 +424,10 @@ RSpec.describe Prog::Minio::MinioServerNexus do
   end
 
   describe "#available?" do
+    before do
+      allow(nx.minio_server).to receive(:cert).and_return("cert")
+    end
+
     it "returns true if initial provisioning is set" do
       expect(nx.minio_server).to receive(:initial_provisioning_set?).and_return(true)
       expect(nx.available?).to be(true)


fdr:
How did anything work previously?

geemus:
Greetings all. Thanks @furkansahin for hopping on a call to share some further context. Just finding some time to return to this today, so I'll start by summarizing and then add what I've found so far.

The [code](https://github.com/ubicloud/ubicloud/blob/83644cc6cafcd9dcc8fa776cc5bab8213c69ff74/prog/minio/minio_cluster_nexus.rb#L21-L22) created two root certificates for each minio cluster, one with a 5 year expiry and one with a 10 year expiry. When servers boot, the [code](https://github.com/ubicloud/ubicloud/blob/83644cc6cafcd9dcc8fa776cc5bab8213c69ff74/prog/minio/minio_server_nexus.rb#L204-L209) uses the 5 year key if at least one year remains to expiry, and otherwise it uses the 10 year certificate.

The client [code]() then loads a bundle which is formed by concatenating these root certs and passing it as a ca_file to excon.

Apparently this did (and possibly does) work in some contexts, but now works inconsistently at best. We found that swapping the order of the root certs (so 10 year is first and 5 year is last) appears to fix it, as well a few other variations.

One thing to note is that there is no shared trust between the two certificates, so when the old one expires or is removed any related certificates will be invalid (maybe that's fine, given the one year overlap, presuming you notice and address it within that time). You may want to at least ensure there is monitoring or errors raised or something, which I didn't see.

Alternatively, how this would work outside self-signed contexts would be that you would have a CA that would mint certs and you could renew the cert as it neared expiry. The plus side to this is that existing certs wouldn't become invalid, the downside being running/maintaining a CA is more complicated (and can be more risky if done incorrectly).

Setting that aside, I was trying to dig into how to properly deal with setting multiple roots to trust for excon/ruby.

I had a hard time finding anything that was particularly satisfying/definitive, but a number of sources suggest that PEM/CRT files should represent a particular chain of trust starting with the narrowest cert and then proceeding with the certs that they trust up through the root. They also suggest that some software may be smart enough to disentangle this if you do it wrong, but no guarantees (which may be why we see what we do here). Here are some of the sources, fwiw: [1](https://knowledge.digicert.com/tutorials/create-pem-file-for-tls-ssl-certificate-installations) [2](https://stackoverflow.com/questions/72044547/how-is-an-ssl-certificate-chain-bundle-arranged) [3](https://www.howtogeek.com/devops/what-is-a-pem-file-and-how-do-you-use-it/)

So, it may be that there are some quirks relating to the implication that two roots are part of the same chain and/or that they relate to one another in a way that they don't (since these certs do not share trust in any way).

Given all that, one possibility there is a possibility that would allow us to avoid the file aspects of this (which seem unnecessary and also may be the root of our issues). Besides files/paths, excon also supports passing in a cert store, which would give us more levers. I think something like this in the minio client might be more successful:

```ruby
ssl_cert_store = OpenSSL::X509::Store.new
ssl_cert_store.set_default_paths # include system store certs
root_certs.each do |cert| # currently the client receives this as ssl_ca_file_data
  ssl_cert_store.add_cert(cert)
end
@client = Excon.new(endpoint, socket: socket, ssl_cert_store: ssl_cert_store)
```

I think that gets you a client that:
1. Trusts both of the root certificates.
2. Avoids any implied or intended relationship between the two root certs, which are independent.
3. Allows you to skip file writing/manipulation, since the certs appear to already be in memory.

I'll note, however, that I haven't tried this and the docs are spotty at best. I'm hopeful it may get you to a better place though. Please do let me know if you have questions and how I can further assist.

fdr:
Seems like the commit message could be a bit more robust, that _some clients_ validated the request, and some didn't. that, and the more firm findings e.g. in @geemus's post would make sense, such as it being sensitive to ordering. 

Background, because some aspects of this may be mysterious: the idea of two roots that are unrelated is to deal with rotation, since updating trust chains in every client endpoint is incredibly hard. This is exactly like how CAs retire their keys or even upgrade algorithms, it's a slow, many-years affair while rolling off customer signed certs and ancient trust stores.

So the period is chosen to be long enough where, for most purposes, the customer won't even have to be aware of the rotation interval, if they have even minimal contact with downloading the trust bundle once every _few years_. Our situation is more relaxed than the CA case, and a bit faster, but that's kind of the idea.

The 5/10 formation is also kind of weird, it's because it's a special case when a new cluster is made, but in the steady state, we would normally expect 10yr/10yr certs. Basically, we'd prepare for rotation years in advance by updating the trust chain and then not using that root of trust, so the customer could slowly get it to propagate before it ever was used.

In the fullness of time, Clover would store that newer cert's private key material somewhere orthogonal (and significantly less exposed) than the same database holding the current certificate in use. For now, it exists as something of a model to find a (hopefully small number of) unexpected problems...like this one.

byucesoy (approved):

jeremyevans:
When setting up the staging environment, I found this change broke things. It would only validate with the CA certificates, and not with the server certificate added. It's a CA bundle, it should only have CA certificates. Not sure why adding the server cert fixed earlier cases, but it seems wrong. So I propose to revert this (included in #3938).

jeremyevans:
Maybe the issue before was this wasn't using `ssl_cert_store`? Maybe after `ssl_cert_store`, the server cert is no longer necessary (and can cause problems)? The `ssl_cert_store` change wasn't made until 886eb0d59dc9f733974cd5ad57ff8581f5cbfa88, about 4 months after this was committed.

PR #2678, opened 2025-01-26
Fix ssh config for the server that were applied to the client

fdr:
There was a single-character miss here, that I only noticed upon trying to use SSSH from a host to do some diagnostics, as it causes the `ssh` client to crash complaining about incorrect options.

Conversely, it means these SSH server options were never applied...or tested, I guess (e.g. inspecting for verbose output that included SSH key fingerprints, per the comment on `LogLevel VERBOSE`)


shikharbhardwaj (approved):
Oof nice find, looks like this was broken for a while

shikharbhardwaj:
We'll need to fix this for all existing hosts I guess cc @byucesoy

furkansahin (approved):

PR #2686, opened 2025-01-28
Add more data to the deadline pages

enescakir:
- **Add create_vm_host spec helper**
  The VmHost model requires a Sshable record due to database constraints.
  Currently, we're creating VmHost models with dummy Sshable data across
  multiple test locations, leading to code duplication.
  
  Additionally, we're creating VmHost instances using `VmHost.create(...)
  { _1.id = Sshable.create.id }`, which results in VmHost having a Sshable
  UBID instead of its own UBID. While this doesn't break any test, it's
  conceptually incorrect and should be fixed.
  

- **Add more data to the deadline pages**
  Currently, the deadline pages only show the resource type, ID, and
  expected target label. It would be helpful to include additional data
  based on the resource type for the on-call engineer. For example, when a
  host crashes, we receive multiple deadline pages from virtual machines,
  making it easier for the on-call engineer to identify the host by
  reviewing the incident details in Slack.
  
  Additionally, while I haven't implemented this, we could choose whether
  or not to trigger the page based on the resource type to avoid page
  storms. For instance, when creating a new deadline page for a virtual
  machine, we can check if its host is unavailable. If the host is down,
  we can decide not to trigger the page for the virtual machine.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -96,7 +96,19 @@ SQL
     stack.each do |frame|
       if (deadline_at = frame["deadline_at"])
         if Time.now > Time.parse(deadline_at.to_s)
-          Prog::PageNexus.assemble("#{ubid} has an expired deadline! #{effective_prog}.#{label} did not reach #{frame["deadline_target"]} on time", ["Deadline", id, effective_prog, frame["deadline_target"]], ubid)
+          sbj = subject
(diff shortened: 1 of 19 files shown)


bsatzger (approved):
Very useful - thanks for doing this!

PR #2687, opened 2025-01-28
Add the target's ubid to SSH command logs

enescakir:
- **Omit unnecessary hash values in sshable log**
  

- **Add the target's ubid to SSH command logs**
  Our logs include the ubid of the strand that initiated the command as
  the `thread` property. However, this ubid refers to the initiator, not
  the target.
  
  For example, when a GitHub runner strand runs SSH commands on both the
  VM host and the VM, the `thread` for both is the same. If the host
  experiences SSH latencies, we can't aggregate the channel durations of
  the VM host by `thread`. Adding the sshable's `ubid` will help us
  aggregate the durations for the target. This way, if a specific host has
  SSH latencies, we can monitor it effectively.

### DIFF
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -82,9 +82,7 @@ class Sshable < Sequel::Model
     if log
       Clog.emit("ssh cmd execution") do
         finish = Time.now
-        embed = {start: start, finish: finish, duration: finish - start,
-                 cmd: cmd,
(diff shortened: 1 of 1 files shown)


jeremyevans (approved):
I made the same change in #2688.  This will cause conflicts with that, but if you'd like to merge it first, it's fine with me.

fdr (approved):

bsatzger (approved):

PR #2688, opened 2025-01-28
Use a timeout by default in Sshable#cmd

jeremyevans:
This takes the command line given, and runs it with:

  sudo timeout $Xs bash -c -- $ESCAPED_COMMAND

Where $X is the timeout in seconds.  The timeout can be given explicitly.  However, by default, it is slightly less than the apoptosis timeout, so that a command that failed should have enough time to result in the strand failing before apoptosis happens.

This should catch any shell commands that take too long to execute.  It will not catch timeouts as the SSH layer. That support could happen later.

This uses String#shellescape to escape the command, which should correctly escape all shell metacharacters

I'm including the timeout in the Clog emit, though it can also be determined by looking at the cmd.  Not sure if it would be better to emit the original command instead of the actual timeout command.

This switches the newly added `sudo timeout 10s` in VM nexus command with a `timeout: 10` argument.


fdr (commented):
IMO, I would not ship the `timeout` prefixing code in the general case for now, I think that's pretty invasive. 

> This is not a replacement for the timeout code in the previous
commit, because we still want commands that exceed the timeout
to be killed, which the timeout code should handle.

the fact it may terminate a process (it's hard to rigorously track all children, e.g. this would not work for daemonizer for sure) after *some* amount of time I think doesn't solve our problem (which we already have, right? upon customary a retry). 

Did we confirm the lingering process problem, even? We may want to tweak the KeepAlive settings to get the operating system to help SSH know to SIGHUP things more rapidly. We do some keepalive setting from the client that ought to suffice, but i suppose the server ought to better protect itself. 

All in all, I wouldn't try to heuristically address the already-existent problem of wild processes on the server in this way without desperation. I think we can solve it much more rigorously.

Some questions:

* Did you try this out with a `sleep` command (and maybe a short timeout). What does it do? Might be a good thing to paste into a commit message, to show the stack trace. (You might want to shorten the timeout for ginning this up)

* I notice session lost its .wait, does that hurt us at all?

jeremyevans:
> IMO, I would not ship the `timeout` prefixing code in the general case for now, I think that's pretty invasive.
> 
> > This is not a replacement for the timeout code in the previous
> > commit, because we still want commands that exceed the timeout
> > to be killed, which the timeout code should handle.
> 
> the fact it may terminate a process (it's hard to rigorously track all children, e.g. this would not work for daemonizer for sure) after _some_ amount of time I think doesn't solve our problem (which we already have, right? upon customary a retry).

It depends on what the problem is.  This is designed to solve the apoptosis problem, and it should do that.  However, it may well cause other issues for non-idempotent prog labels.

> Did we confirm the lingering process problem, even? We may want to tweak the KeepAlive settings to get the operating system to help SSH know to SIGHUP things more rapidly. We do some keepalive setting from the client that ought to suffice, but i suppose the server ought to better protect itself.

I didn't do any researching on the lingering process problem.  This handles things at the client level, closing the connection after timeout expires, regardless of the state of the connection.

> All in all, I wouldn't try to heuristically address the already-existent problem of wild processes on the server in this way without desperation. I think we can solve it much more rigorously.

Again, this is an attempt to solve the apoptosis issue.  If you think it will cause problems worse than the apoptosis problem it attempts to fix, it is probably not a good approach.

> Some questions:
> 
> * Did you try this out with a `sleep` command (and maybe a short timeout). What does it do? Might be a good thing to paste into a commit message, to show the stack trace. (You might want to shorten the timeout for ginning this up)

I tested this with a `sleep 600` command inside a label.  Without this PR, that results in apoptosis.  With `sudo timeout` (1st commit), no apoptosis, it would just keep retrying the label.  Without the `sudo timeout`, but with the ssh-level timeout handling (both commits, but removing the `sudo timeout $Xs` code) 

> * I notice session lost its .wait, does that hurt us at all?

This wasn't calling wait on the session, it was calling wait on the result of `connect.open_channel`, which returns the same channel it yields.  I kept both waits, but I think they are redundant.  Only the one after the `open_channel` call should be needed.

I did notice one problem here, which is that `exit_code` can be nil, resulting in the wrong error message being displayed.  I'll push another commit to improve the error handling.

fdr:
So I was playing around with SSH and signals and, in my setup, SSH does something expected: with keepalive set from the server (is it not symmetric with the one set in the client? that we theoretically do, but we do not do this from the server), a non-responsive peer will cause the ssh daemon to send a SIGHUP.

The test setup was ClientAliveInterval for sshd and using nftables (`nft`) do something like ` ip6 saddr 2a01:4f8:2b01:d86:4796::2 tcp dport 22 drop` to simulate a non-responsive situation.

On my laptop, running Ubuntu 24.10, I also confirmed that sending SIGHUP to `sudo` will propagate.

Maybe SIGHUP is good enough, maybe `timeout`'s SIGTERM is better, but usually SIGHUP is good enough for termination.  

I might try this with a client-side setting tomorrow, and maybe someone who has seen this lingering can make a few remarks about exactly what was seen (e.g. daemonize, where it's expected, or something else?) @bsatzger were you a reporter on this one?

bsatzger:
> So I was playing around with SSH and signals and, in my setup, SSH does something expected: with keepalive set from the server (is it not symmetric with the one set in the client? that we theoretically do, but we do not do this from the server), a non-responsive peer will cause the ssh daemon to send a SIGHUP. In doing so, cutting the TCP session from the client ought to be adequate to promote to a timeout/kill -SIGHUP on the server side in a fairly tight time boundary of our choosing.
> 
> The test setup was ClientAliveInterval (which I don't think we do at present) for sshd and using nftables (`nft`) do something like ` ip6 saddr 2a01:4f8:2b01:d86:4796::2 tcp dport 22 drop` to simulate a non-responsive situation. I had a bash script that dumped the signal it received running on the connection that was cut.*
> 
> On my laptop, running Ubuntu 24.10, I also confirmed that sending SIGHUP to `sudo` parent process will propagate.
> 
> Maybe SIGHUP is good enough, maybe `timeout`'s SIGTERM is better, but usually SIGHUP is good enough for termination.
> 
> I might try this with a client-side setting tomorrow, and maybe someone who has seen this lingering can make a few remarks about exactly what was seen (e.g. daemonize, where it's expected, or something else?) @bsatzger were you a reporter on this one?
> 
> * yes a LLM got this simple task right in one shot. Gotta do my part to make that 2% annualized year over year BLS productivity improvement you know?
> 
> ```shell
> #!/bin/bash
> 
> # Output file (modify as needed)
> output_file="/tmp/signal.log"
> 
> # Handler function
> handle_signal() {
>     local sig=$1
>     echo "Received signal: $sig" >> "$output_file"
>     exit 0
> }
> 
> # Trap signals
> trap 'handle_signal SIGHUP' HUP
> trap 'handle_signal SIGTERM' TERM
> 
> # Keep the script running
> echo "PID $$ waiting for signals (SIGHUP/HUP, SIGTERM/TERM)..."
> echo "Writing to: $output_file"
> 
> # Infinite wait using builtin (no CPU usage)
> tail -f /dev/null
> ```

The particular issue we encountered multiple times so far: `systemctl` commands, e.g., `systemctl daemon-reload` or `systemctl stop ...`, take a long time and cause apoptosis.

fdr:
fyi: i have done some work on this, not forgotten.

fdr:
okay *now* it was forgotten, and for a long time. Oops. Will renovate this one, mea culpa

fdr:
@jeremyevans I separated the vm nexus code (can be second, after this one proves no problem) and moved the apoptosis thread-local `apoptosis_at` setting around to adapt to the now much different dispatcher. Can you give those a look?

jeremyevans (commented):
Changes look good to me. Can't approve as I'm the submitter. :)

PR #2691, opened 2025-01-29
Allocator diagnostic improvements

fdr:
Improve naming with regard to allocator diagnostics

The previous naming introduced a "enable_diagnostics" feature flag
onto projects.  Firstly, a feature flag existing is to enable it, so
to say "enable_anything" is redundant.

Secondly, the flag controls one precise behavior, the printing of
allocator queries, not a catch-all omnibus diagnostic-mode.  I don't
think the intention was to construct such a device, so name things
more specifically and drop `enable_`.

I do pivot between `allocator_diagnostics` in global context
(e.g. project feature flag name) and simply `diagnostics` in the
allocator code context (e.g. the `Request` struct) to avoid
repetition, though the complication to grepping all uses of the symbol
may counter-indicate it: now a more complicated search pattern will be
required to see how it correlates through the entire program.  I am
ambivalent about my choice here, if later someone wishes to make it
correlate.

Change allocator diagnostics to use structured output

Normally, we do not encode dynamic information in the "banner string"
given to `Clog.emit`, it is static.  Visit
f2b128f972ce81847956e386d626bd9a94c0106e for the underpinnings of
Clog.

It was also unnecessary to introduce a coverage gap here: the result
is, once diagnostics is enabled, the process could run unchecked code
and crashed.  The structured output also makes it easy to check some
invariants about the printed data.

There is one regression, however: the old way of doing things would
look better in plain text, for copy-and-paste.  But the new way would
only look good in something that de-escapes JSON, in particular
double-quotes for SQL identifiers, and renders it, such as the log
analyzer in the production system.  I am not sure how to remediate
this.


fdr:
@macieksarnowicz (since I guess you are external and thus cannot be an approving reviewer per our SOC restrictions, but, FYI)

The ugly part is how this affects your copy-paste workflow in development because of the escaping. We may need a way out (perhaps to `puts` if `isatty` which is something we do with SSH channels, in fact, to emit stderr:

```ruby
          ch.on_data do |ch, data|
            $stderr.write(data) if repl?
            stdout.write(data)
          end
```

furkansahin (approved):

macieksarnowicz:
@fdr - the naming of the flag was intentionally broad - I expect there might be a need to introduce more diagnostics logging in other places and I think introducing a new flag for each of those would be an overkill. Having said that, it is hypothetical, and the name can be changed again if needed, so I agree the name change brings clarity. 

As for the formatting of the string - I don't have a good solution for this. I think this was why I was writing to a file in the original version of this code, but we discussed that already. There might be some VS Code helper/extension to remedy this, I will look for one.

pykello (approved):

PR #2694, opened 2025-01-29
CertNexus restart counter is added to not DOS ZeroSSL

furkansahin:
In case the passed DNSZone is not valid globally, DNS validation or cert generation may fail. In that case, CertNexus restarts to do a new acme dance. It goes on forever. Instead of this, we are introducing a restart logic. In case of restarts, we start napping aggressively up until 10 minutes.

Fixes https://linear.app/ubicloud/issue/UBI-33/do-we-back-off-acme-zerossl-if-customer-doesnt-control-ns

### DIFF
--- a/prog/vnet/cert_nexus.rb
+++ b/prog/vnet/cert_nexus.rb
@@ -16,7 +16,7 @@ class Prog::Vnet::CertNexus < Prog::Base
     DB.transaction do
       cert = Cert.create_with_id(hostname: hostname, dns_zone_id: dns_zone_id)
 
-      Strand.create(prog: "Vnet::CertNexus", label: "start") { _1.id = cert.id }
+      Strand.create(prog: "Vnet::CertNexus", label: "start", stack: [{"restarted" => 0}]) { _1.id = cert.id }
     end
   end
 
@@ -72,7 +72,7 @@ class Prog::Vnet::CertNexus < Prog::Base
     else
       Clog.emit("DNS validation failed") { {order_status: dns_challenge.status} }
       dns_zone.delete_record(record_name: dns_record_name)
-      hop_start
+      hop_restart
     end
   end
 
@@ -88,7 +88,7 @@ class Prog::Vnet::CertNexus < Prog::Base
     else
       Clog.emit("Certificate finalization failed") { {order_status: acme_order.status} }
       dns_zone.delete_record(record_name: dns_record_name)
-      hop_start
+      hop_restart
     end
   end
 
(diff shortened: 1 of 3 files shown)


prog/vnet/cert_nexus.rb
```diff
@@ -127,6 +138,13 @@ def before_run
     pop "certificate revoked and destroyed"
   end
 
+  def update_stack_restart_counter
+    current_frame = strand.stack.first
+    current_frame[:restarted] += 1
```
fdr, line 143: why a frame variable rather than something in an entity? Is this strand not attached to any entity?

furkansahin: It feels like the counter is very specific to the strand rather than entity. I didn't want to make a modification to the model to be honest.

enescakir, line 143: Since the stack is loaded from the database, it will have keys as strings, not symbols.
```suggestion
    current_frame["restarted"] += 1
```

prog/vnet/cert_nexus.rb
```diff
@@ -101,6 +101,17 @@ def before_run
...
 
+  label def restart
+    when_restarted_set? do
+      decr_restarted
+      update_stack_restart_counter
+      hop_start
+    end
+
+    cert.incr_restarted
+    nap [60 * (strand.stack.first[:restarted] + 1), 60 * 10].min
```
enescakir, line 112: nit:
```suggestion
    nap [60 * (frame["restarted"] + 1), 60 * 10].min
```

enescakir: Also, if you don't update the stack of the existing strands, it will fail. Please don't forget to update them.

prog/vnet/cert_nexus.rb
```diff
@@ -16,7 +16,7 @@ def self.assemble(hostname, dns_zone_id)
     DB.transaction do
       cert = Cert.create_with_id(hostname: hostname, dns_zone_id: dns_zone_id)
 
-      Strand.create(prog: "Vnet::CertNexus", label: "start") { _1.id = cert.id }
+      Strand.create(prog: "Vnet::CertNexus", label: "start", stack: [{restarted: 0}]) { _1.id = cert.id }
```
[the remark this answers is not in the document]
enescakir: https://github.com/ubicloud/ubicloud/pull/2694/files#r1935496533

prog/vnet/cert_nexus.rb
fdr: alright, I think that makes sense. So you don't think it fits with an existing data structure and doesn't justify a new one (not enough fields on the stack that are interesting enough for a schema?)

enescakir (approved):

PR #2698, opened 2025-01-29
Load available VM sizes from a config file

enescakir:
- **Load available VM sizes from a config file**
  As the number of VM size options and parameters in the VmSize struct has
  increased, it has become difficult to identify available VM sizes and
  their possible values from the code. I suggest loading available VM
  sizes from a config file, similar to how we manage runner labels, AI
  models, and so on.
  

- **Add gpu helper method to VmSize struct**
  GPUs are only available for "standard-gpu" family. No need to keep
  additional property in the config file


enescakir:
@fdr, @jeremyevans, @pykello What do you think about this PR?

jeremyevans (approved):

pykello:
Adding @macieksarnowicz since this is related to the work he's doing.

@macieksarnowicz what do you think about this change?

spec/lib/option_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+RSpec.describe Option do
+  it "can load io_limits to struct" do
```
enescakir, line 4: @jeremyevans, I added this test because of `row["io_limits"] = row["io_limits"].any? ? IoLimits.new(*row["io_limits"]) : NO_IO_LIMITS`, we don't have any VM size with an IO limit. Since we load `VmSizes` during the loading and freeze it, my YAML mocking isn't used. Do you have any solution to make this test work?

macieksarnowicz: FYI - I will be introducing VmSizes with IO Limits in my next PR, hopefully soon.

config/vm_sizes.yml
```diff
@@ -0,0 +1,13 @@
+- { name: "standard-2",     family: "standard",     arch: "x64",   cores: 1,  vcpus: 2,  cpu_percent_limit: 200,  cpu_burst_percent_limit: 0, memory_gib: 8,   storage_size_options: [40, 80],     io_limits: [null, null, null], visible: true }
```
macieksarnowicz, line 1: Do you find this format preferable to the computed model that was there before? Just curious, what is motivating this change.

FYI - I will be adding Burstable sizes soon, that will add 4 more entries.

spec/lib/option_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+RSpec.describe Option do
```
macieksarnowicz, line 3: I added a few tests for VmSizes in vm_spec.rb. In hindsight they should be in a new file, like you have it here. Please see if you can move those, or I can do it after you get this merged into main.

macieksarnowicz, line 3: I added a few tests for VmSizes in vm_spec.rb. In hindsight they should be in a new file, like you have it here. Please see if you can move those, or I can do it after you get this merged into main.

config/vm_sizes.yml
macieksarnowicz: Also, what then about PostgresSizes?

enescakir: > Do you find this format preferable to the computed model that was there before? 

Yes, I found it more readable.

> Just curious, what is motivating this change.

I want to make our available VM sizes and their properties clearer. In the previous method, it was hard to see them at a glance.

> Also, what then about PostgresSizes?

We can do them after this PR.

enescakir:
I talked with @byucesoy about this PR. He has some ideas for making it more dynamic and clearer. I'm closing this PR for now.

PR #2699, opened 2025-01-29
Adding E2E tests for slices

macieksarnowicz:
Updating E2E tests to validate VM creation inside a slice. The test switches the host's `accept_slices` flag to true and sets the project flag to use slices, then runs the standard VM test to create VMs and test them. Those VMs are creted inside slices on a host. An additional verification is added into VmGroup to check the validity of the slices - `VmHostSlices`.

In a separate commit, fixing a race condition in `Test::ConnectedSubnets` and `Test::FirewallRules`.

### DIFF
--- a/prog/test/connected_subnets.rb
+++ b/prog/test/connected_subnets.rb
@@ -5,7 +5,7 @@ require "uri"
 
 class Prog::Test::ConnectedSubnets < Prog::Test::Base
   label def start
-    unless frame["connected"]
+    unless frame["vm_to_be_connected_id"]
       pss.map(&:vms).flatten.each do |vm|
         vm.sshable.cmd("sudo yum install -y nc") if vm.boot_image.include?("almalinux")
         vm.sshable.cmd("sudo apt-get update && sudo apt-get install -y netcat-openbsd") if vm.boot_image.include?("debian")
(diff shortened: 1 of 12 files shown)


prog/test/connected_subnets.rb
```diff
@@ -189,7 +189,12 @@ def pss
   end
 
   def vm_to_be_connected
-    @vm_to_be_connected ||= ps_multiple.vms.first
+    connected_id = frame["vm_to_be_connected_id"]
+    @vm_to_be_connected ||= if connected_id
```
pykello, line 193:
```suggestion
    @vm_to_be_connected ||= unless connected_id
```

macieksarnowicz: @pykello - do you want me to swap the order of if/else branches? What's the benefit here?

pykello: No, this was a comment on the previous version which failed specs.

macieksarnowicz: Updated now.

pykello (approved):
Changes look good.

I also ran the E2E pipeline once and it took 39:36 minutes, up 3:46 minutes from the previous duration (35:50 minutes), which I think is acceptable.

enescakir:
> Changes look good.
> 
> I also ran the E2E pipeline once and it took 39:36 minutes, up 3:46 minutes from the previous duration (35:50 minutes), which I think is acceptable.

We have 40 minutes time limit on E2E tests, so 30:36 makes it a bit flaky. I increased the timeout.
https://github.com/ubicloud/ubicloud/pull/2712

PR #2700, opened 2025-01-29
Decrease neg-ttl in dnsmasq config

fdr:
Default neg-ttl is 300 seconds which is a lot to wait for retrying dns queries. For bootstrapping kubernetes, we have 4 minute timeout to discover the apiserver through the loadbalancer and DNS. If the first query fails, the bootstrap will fail and way more complex measurements should be added to resolve it.

In this commit the neg-ttl is set to 30 seconds which is reasonable for retrying failed dns queries

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -484,6 +484,7 @@ dns-forward-max=10000
 dhcp-option=option6:dns-server,#{dnsmasq_address_ip6}
 listen-address=#{dnsmasq_address_ip6}
 all-servers
+neg-ttl=30
 DNSMASQ_CONF
 
     ethernets = nics.map do |nic|


fdr:
@furkansahin, I cherry-picked it out of a k8s commit sequence, it could use your supervision and be baked in first, if you think it's a good idea.

furkansahin (changes requested):
Instead of setting this at the Dnsmasq level, we can configure it in our DnsZone. Dnsmasq actually caches the negative DNS according to the ttl value coming from NXDomain record, which is served from our own DNS Servers. The SOA record we add at the end here https://github.com/ubicloud/ubicloud/blob/main/prog/dns_zone/setup_dns_server_vm.rb#L136 can be reduced to a very low value which would simply only impact the dnszone we control and not all of the dns checks. In this way, we can be a very nice customer.

fdr:
Interesting, so maybe it can be a special on the zone. @mohi-kalantari what are the host names you compute these days for api servers, or other things? I guess it's in the code, but we never discussed how things are boxed exactly (e.g. how postgres and load balancer both have some theory and practice around how various dns labels are boxed)

mohi-kalantari:
@fdr We actually don't do any specific computations. We create a loadbalancer and based on the private subnet name (I guess) a domain is given to our LB and then we hand it over to kubernetes and the rest is typical k8s bootstrapping

1. sigining certs on the domain
2. listening on 0.0.0.0
3. trying to connect to itself through the LB which requires the DNS lookup

fdr:
@mohi-kalantari okay so it's getting default project-name nesting for LBs. Not sure if that's good enough or there's meaningful improvement in a `k8s.ubicloud.com` thing much as AI and PG have (special zones for managed services). Do you have thoughts about whether there is merit in replicating that pattern, and when to do it? 

You too, @furkansahin, as the one who has thought about load balancer, dns, certificates, inside and out quite extensively (and certainly the most as a combined package)

furkansahin:
It actually makes perfect sense for K8s service to own their own subdomain and the full control over DNS Zone. I guess they would want to do that in the next stage... Right now, what we can do is, reduce the neg-ttl value for the `lb.ubicloud.com` SOA record to 30. Later, we can increase it back up and specifically configure K8s DNSZone neg-ttl to be very low.

fdr:
what's your feel on when we should do it i.e. is it really cheap and we should do it avoid path dependency? Or run trials without and risk a handful of complicated "legacy" api server endpoints?

mohi-kalantari:
I'd say probably run trials without and risk a handful of complicated "legacy" api server endpoints

we should close this PR and come back later if we needed it.

@fdr

PR #2703, opened 2025-01-30
Attempt to recreate VM up to 2 times if it fails during prep

jeremyevans:
Previously, if a VM failed during Vm::Nexus#prep, it would continually be stuck in that step, because #prep treated Failed the same as NotStarted, and the steps to initially start prep are not the same as the steps to recover from failure.  I'm not sure what the steps are to recover from failure (since there are many different things that could fail), so this takes the hammer approach.  It sets a recreate semaphore and hops to destroy.

When destroy (#destroy_slice) is called, if the recreate semaphore is set, after destroying the VM, it assembles a new VM with the same ubid, and then hops to start instead of exiting.

To prevent infinite recreation, there is a hard coded limit of 3 attempts. After 3 failed attempts, the Strand is moved to prep_failed state, where it stays until manual intervention. This generates a page to ensure it isn't missed.

To implement this, I added the following keyword arguments to Vm::Nexus#assemble:

* ubid: ubid to set on the created Vm
* vm_size: a VmSize object, which can be used directly instead of determining it from the size and arch arguments
* attempt: The number of attempts made to create the VM, defaults to 1
* strand: Updates existing strand instead of creating a new strand

To hold more state, it adds some additional data to the strand's stack, so that the destroy_slice method can pass it back to assemble.

Other minor changes:

* Add Vm#vm_size. Extracted from Vm#display_size, this is used to set the vm_size argument to assemble.

* Do not show "deleting" as the display state for a Vm with recreate set, as showing that would look weird to the user.

* If requesting to create a Vm with a specific Nic (passing nic_id to assemble), when destroying and recreate is set, the related Nic is not destroyed when the Vm is destroyed, as it will be used for the newly created Vm.  I'm not sure this is used outside the tests, so it may be overkill.

* Remove transaction around Vm::Nexus#final_clean_up. This isn't needed, as it is only called inside a label, and labels have an implicit transaction around them.

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -26,7 +26,7 @@ class Vm < Sequel::Model
   include SemaphoreMethods
   include HealthMonitorMethods
   semaphore :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules, :checkup, :update_spdk_dependency, :waiting_for_capacity, :lb_expiry_started
-  semaphore :restart, :stop
+  semaphore :restart, :stop, :recreate
 
   include ObjectTag::Cleanup
 
@@ -63,7 +63,7 @@ class Vm < Sequel::Model
   end
 
   def display_state
-    return "deleting" if destroy_set? || strand&.label == "destroy"
+    return "deleting" if (destroy_set? && !recreate_set?) || strand&.label == "destroy"
     return "restarting" if restart_set? || strand&.label == "restart"
     return "stopped" if stop_set? || strand&.label == "stopped"
     if waiting_for_capacity_set?
@@ -136,13 +136,16 @@ class Vm < Sequel::Model
   # Reverse look-up the vm_size instance that was used to create this VM
   # and use its name as a display name.
   def display_size
-    vm_size = Option::VmSizes.find {
+    vm_size.name
+  end
+
+  def vm_size
+    Option::VmSizes.find {
       _1.family == family &&
         _1.arch == arch &&
         _1.vcpus == vcpus &&
         (cpu_percent_limit.nil? || _1.cpu_percent_limit == cpu_percent_limit)
     }
-    vm_size.name
   end
 
   # Various names in linux, like interface names, are obliged to be
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -13,7 +13,8 @@ class Prog::Vm::Nexus < Prog::Base
     unix_user: "ubi", location: "hetzner-fsn1", boot_image: Config.default_boot_image_name,
     private_subnet_id: nil, nic_id: nil, storage_volumes: nil, boot_disk_index: 0,
     enable_ip4: false, pool_id: nil, arch: "x64", swap_size_bytes: nil,
-    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [], gpu_count: 0)
+    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [], gpu_count: 0,
+    ubid: nil, vm_size: nil, attempt: 1, strand: nil)
 
     unless (project = Project[project_id])
       fail "No existing project"
@@ -22,8 +23,9 @@ class Prog::Vm::Nexus < Prog::Base
       fail "Cannot force and exclude the same host"
     end
     Validation.validate_location(location)
-    vm_size = Validation.validate_vm_size(size, arch)
+    vm_size ||= Validation.validate_vm_size(size, arch)
 
+    assemble_storage_volumes = storage_volumes
     storage_volumes ||= [{
       size_gib: vm_size.storage_size_options.first,
       encrypted: true
@@ -49,7 +51,7 @@ class Prog::Vm::Nexus < Prog::Base
 
     Validation.validate_storage_volumes(storage_volumes, boot_disk_index)
 
-    ubid = Vm.generate_ubid
+    ubid ||= Vm.generate_ubid
     name ||= Vm.ubid_to_name(ubid)
 
     Validation.validate_name(name)
@@ -108,18 +110,26 @@ class Prog::Vm::Nexus < Prog::Base
       nic.update(vm_id: vm.id)
 
       gpu_count = 1 if gpu_count == 0 && vm_size.gpu
-      Strand.create(
+
+      strand ||= Strand.new { _1.id = vm.id }
+      frame = strand.stack[-1] || {}
+      strand.update(
         prog: "Vm::Nexus",
         label: "start",
-        stack: [{
+        stack: [frame.merge(
+          "assemble_storage_volumes" => assemble_storage_volumes,
           "storage_volumes" => storage_volumes.map { |v| v.transform_keys(&:to_s) },
           "swap_size_bytes" => swap_size_bytes,
           "distinct_storage_devices" => distinct_storage_devices,
           "force_host_id" => force_host_id,
           "exclude_host_ids" => exclude_host_ids,
+          "private_subnet_id" => private_subnet_id,
+          "nic_id" => nic_id,
+          "pool_id" => pool_id,
+          "attempt" => attempt,
           "gpu_count" => gpu_count
-        }]
-      ) { _1.id = vm.id }
+        )]
+      )
     end
   end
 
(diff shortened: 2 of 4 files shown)


prog/vm/nexus.rb
```diff
@@ -258,8 +268,21 @@ def before_run
     when "Succeeded"
       host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")
       vm.private_subnets.each(&:incr_add_new_nic)
+      # To simulate failure case in development:
+      # if frame["attempt"] < 3
+      #   Clog.emit("VM prep forced fail for #{vm.ubid}, attempt: #{frame["attempt"]}")
```
fdr, line 273: Clog is really only useful in its structured form for log analysis, if doing this, maybe do `puts`.

prog/vm/nexus.rb
```diff
@@ -425,6 +448,10 @@ def write_params_json
     nap 30
   end
 
+  label def prep_failed
+    nap(5 * 60)
```
fdr, line 452: it can be much longer. A regression from system 3: naps were often 1000 years (I did that so the first digit was notable in dumps, but you got the lower bits to intuit what time the nap was done). _but_ if you look are the semaphore code, it wakes up a strand when it happens, that's how the whole system avoided polling. That this one does not do the same is vestigial/extraneous habits from system 2.

fdr (approved):
pretty nice, I was wondering how you were going to get the provisioning data structures to be persisted. Do you think a further/independent extension of this would be improved by using tables rather than loading so much stuff into the stack?

The stuff I was talking about, the double-trigger (can_recreate + recreate) is more applicable if recreate might apply to a number of possible labels, to bound the degree of preemption safely (before there is irreplaceable user data/the system is live). In this case, the problem is avoided because the recreate semaphore only gets incremented in one place, a prep failure.

model/vm.rb
```diff
@@ -121,13 +121,16 @@ def cloud_hypervisor_cpu_topology
   # Reverse look-up the vm_size instance that was used to create this VM
   # and use its name as a display name.
   def display_size
```
enescakir, line 123: It looks like `display_size` is only used in Vm serializers. We can remove it and use `vm.vm_size.name` directly.

prog/vm/nexus.rb
```diff
@@ -258,8 +268,21 @@ def before_run
...
       vm.private_subnets.each(&:incr_add_new_nic)
+      # To simulate failure case in development:
+      # if frame["attempt"] < 3
+      #   Clog.emit("VM prep forced fail for #{vm.ubid}, attempt: #{frame["attempt"]}")
+      #   incr_recreate
+      #   hop_destroy
+      # end
       hop_wait_sshable
-    when "NotStarted", "Failed"
+    when "Failed"
```
enescakir, line 278: We start the VM's systemd unit at the end of the prep script to avoid delaying VM provisioning. The dataplane starts it to prevent waiting for the control plane.

We noticed a few examples, especially around apoptosis times, where the prep status showed `Succeeded`, but after cleaning up the demonizer service, it crashed just before hopping to the `wait_sshable` label. In the next run, since the demonizer service was already cleaned up, the status returned `NotStarted`, and the control plane tried to rerun the prep script. However, the VM service is already running, so the new prep failed with a `Device or resource busy` exception from the spdk.

So, even though we ended up in a `Failed` state, the VM might have already started. IMHO, we should also check the status of the VM's systemd service, and if it's running, we shouldn't recreate it. We might hop to `wait_sshable` or stuck in the `prep`

```
prep_vmaywd8e.stderr
#<Thread:0x000055f5c9185eb0 /home/rhizome/host/lib/vm_setup.rb:51 run> terminated with exception (report_on_exception is true):
/home/rhizome/host/lib/spdk_rpc.rb:124:in `call': Device or resource busy (SpdkRpcError)
	from /home/rhizome/host/lib/spdk_rpc.rb:75:in `vhost_delete_controller'
	from /home/rhizome/host/lib/storage_volume.rb:93:in `purge_spdk_artifacts'
	from /home/rhizome/host/lib/storage_volume.rb:83:in `rescue in start'
	from /home/rhizome/host/lib/storage_volume.rb:74:in `start'
	from /home/rhizome/host/lib/vm_setup.rb:559:in `block in storage'
	from /home/rhizome/host/lib/vm_setup.rb:554:in `map'
	from /home/rhizome/host/lib/vm_setup.rb:554:in `storage'
	from /home/rhizome/host/lib/vm_setup.rb:52:in `block in prep'
/home/rhizome/host/lib/spdk_rpc.rb:124:in `call': failed to create DEK, rc -17 (SpdkExists)
	from /home/rhizome/host/lib/spdk_rpc.rb:87:in `accel_crypto_key_create'
	from /home/rhizome/host/lib/storage_volume.rb:252:in `setup_spdk_bdev'
	from /home/rhizome/host/lib/storage_volume.rb:75:in `start'
	from /home/rhizome/host/lib/vm_setup.rb:559:in `block in storage'
	from /home/rhizome/host/lib/vm_setup.rb:554:in `map'
	from /home/rhizome/host/lib/vm_setup.rb:554:in `storage'
	from /home/rhizome/host/lib/vm_setup.rb:52:in `block in prep'
#<Thread:0x000055f5c91861a8 /home/rhizome/host/lib/vm_setup.rb:48 run> terminated with exception (report_on_exception is true):
/home/rhizome/common/lib/util.rb:29:in `r': command failed: ip link add vethovmaywd8e addr 0a:93:7f:e5:ab:2d type veth peer name vethivmaywd8e addr 7e:92:82:8e:b3:43 netns vmaywd8e (CommandFail)

---STDOUT---
```

furkansahin:
I was actually waiting for an approval :D https://github.com/ubicloud/ubicloud/pull/2193

fdr:
> I was actually waiting for an approval :D #2193

oops :| sorry about that

prog/vm/nexus.rb
jeremyevans: @enescakir Your idea seems good.  I'm planning to merge this, since it should fix the VM stuck in prep issue, even if it is a heavier hand than is required.  Could you submit a PR to avoid the recreate if you can confirm the VM is working after this is merged?

model/vm.rb
jeremyevans: Good to know.  I can work on a follow-up PR that does this.

prog/vm/nexus.rb
jeremyevans: I'll add a PR to update the nap time after merging this.

prog/vm/nexus.rb
jeremyevans: I'll include a change to this comment in another PR.

prog/vm/nexus.rb
enescakir: If I'm not mistaken, we shouldn't merge this PR without the improvement I suggested. In the current state of this PR, we can destroy and recreate a customer VM that has already been started, right?

fdr:
@jeremyevans I think this code exacerbates an existing problem (per @enescakir remark)

```
when "Succeeded"
      host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")
      vm.private_subnets.each(&:incr_add_new_nic)
      # To simulate failure case in development:
      # if frame["attempt"] < 3
      #   Clog.emit("VM prep forced fail for #{vm.ubid}, attempt: #{frame["attempt"]}")
      #   incr_recreate
      #   hop_destroy
      # end
      hop_wait_sshable
```


`host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")` has an immediate side effect, but the transaction that does the hop may not commit. Thus, we could re-enter in the NotStarted state. Probably, "clean" (all cleans) should happen in a future label that cannot mis-interpret the state of whether the process was started or not. We may have to audit all daemonizer usage for this, and contemplate making it less error prone.

prog/vm/nexus.rb
jeremyevans: Well, this PR had to be reverted, but the cause was a different issue.  I don't think the approach should cause problems unless an administrator calls `incr_recreate` directly after the customer VM is running (which has basically the same issues as `incr_destroy` directly). `incr_recreate`  is only called automatically it prep fails, which is before the customer has any data on/access to the VM.

PR #2704, opened 2025-01-30
Do not create the automatically created route using noprefixroute

fdr:
When we create a new linux addr using ip addr, a corresponding linux route is also created. we will use the noprefixroute to skip the creation of that route. we will create another route using via in order to route packets using the gateway inside the subnet.


fdr:
Something feels vaguely fishy about this, to have to delete automatic routes rather than suppressing them to begin with, but I suppose it's possible it's not a thing iproute2 planned for. IIRC noprefixroute did not do the job for some reason? How was that?

fdr:
This is also a cherry-picked commit, so, "retweets are not endorsements"

mohi-kalantari:
Yes the noprefixroute didn't work. I'll provide more details why it didn't work

mohi-kalantari (approved):

PR #2705, opened 2025-01-30
Add option to specify custom range cidrs for IPV4

fdr:
Currently subnets are created with /26 netmask and VMs get a /32 IP. For kubernetes CNI, we need to have bigger subnets (perferably /16) so each VM gets a /24 range to host pods and assign an IP from subnet to them.

In this commit, based on the subnet mask, we always try to assign at most 2**8 ips to each vm. If the subnet mask is 26 min(32, 26+8) would be calculated which is 32 so each VM gets a /32 ip. but for bigger subnets, VMs would get at most 256 ips.

### DIFF
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -177,11 +177,13 @@ class Prog::Vnet::SubnetNexus < Prog::Base
     selected_addr
   end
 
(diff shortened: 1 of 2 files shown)


fdr:
This should be latent code we can bake a while ahead of time, plus pipeline the review.

furkansahin (approved):

PR #2710, opened 2025-01-30
Refactor providers

furkansahin:
## Replace references of HetznerHost with HostProvider
We are replacing the HetznerHost references with HostProvider so that we
can support multiple provider entities with a single reference. This
way, we will be able to implement further provider APIs without
requiring individual provider model references.

## Drop HetznerHost table
With the multi stage migration, we are making sure the deployment goes
through with zero downtime. With the 2nd commit, we are switching the
controlplane to start using the new model without destroying the
existing HetznerHost model. Now, with this commit, we can safely delete
the HetznerHost table.


furkansahin:
This is a preparation PR for https://github.com/ubicloud/ubicloud/pull/2644

migrate/20250130_create_host_provider_table.rb
```diff
@@ -0,0 +1,21 @@
...
+
+Sequel.migration do
+  up do
+    run <<-SQL
+      CREATE TABLE host_provider (LIKE hetzner_host INCLUDING ALL);
+    SQL
+
+    alter_table(:host_provider) do
+      add_column :provider_name, :text
+    end
```
enescakir, line 11: Since it only has 2 columns, I prefer the explicit option. In addition, It should be unique by both provider and identifier, not just by the identifier like `hetzner_host`.

```suggestion
    alter_table(:host_provider) do
      foreign_key :id, :vm_host, primary_key: true, type: :uuid
      column :provider_name, :text, null: false
      column :server_identifier, :text, null: false
      index [:provider_name, :server_identifier], unique: true
    end
```

model/hetzner_host.rb
```diff
@@ -31,5 +31,3 @@ def password
 # Indexes:
 #  hetzner_host_pkey                  | PRIMARY KEY btree (id)
 #  hetzner_host_server_identifier_key | UNIQUE btree (server_identifier)
-# Foreign key constraints:
-#  hetzner_host_id_fkey | (id) REFERENCES vm_host(id)
```
enescakir, line 35: I don't think the first commit should include these removals. It's just migration for new table

migrate/20250130_create_host_provider_table.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    run <<-SQL
+      CREATE TABLE host_provider (LIKE hetzner_host INCLUDING ALL);
+    SQL
+
+    alter_table(:host_provider) do
```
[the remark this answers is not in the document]
furkansahin: `HostProvider` model is used as an entrypoint to use which API client to use. I didn't want to introduce that into `vm_host` and make the file even longer. This feels like a niche purposed model.

migrate/20250130_create_host_provider_table.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_table(:host_provider) do
+      foreign_key :id, :vm_host, type: :uuid
+      column :server_identifier, :text, null: false, unique: true
+      column :provider_name, :text, null: false
+      index [:provider_name, :server_identifier], unique: true
```
jeremyevans, line 9:
```suggestion
      column :server_identifier, :text
      column :provider_name, :text
      primary_key [:provider_name, :server_identifier]
```

It's best it all tables have a primary key.  This removes the unique constraint on just server_identifier, since I think it's possible for separate providers to use the same server_identifier, even if it is very unlikely.

migrate/20250131_drop_hetzner_host.rb
```diff
@@ -0,0 +1,18 @@
...
+
+Sequel.migration do
+  up do
+    drop_table :hetzner_host
+  end
+
+  down do
+    alter_table(:host_provider) do
+      drop_column :provider_name
+    end
```
jeremyevans, line 11: The up block does not add a provider_name column to host_provider, so the down block should not drop the column.  Instead of using `LIKE host_provider INCLUDING ALL`, you need to recreate the hetzner host table as it was, and then populate it.

jeremyevans (commented):
Looks pretty good.  I do have some recommendations for the migrations.

model/hetzner_host.rb
jeremyevans: This is incorrect.  The migration changes the state of the database, and the model annotations need to match.  Otherwise, the commit would fail CI in isolation.

enescakir (approved):

PR #2711, opened 2025-01-30
Make K8s node compute and storage sizes be specifiable

aamederen:
---

**Make K8s node compute and storage sizes be specifiable**

This patch makes the Kubernetes node provisioner respect the specified
compute and storage sizes of kubernetes cluster or nodepool during VM
provisioning.

---

**Add migration for kubernetes vm size specification**

This patch adds new columns to kubernetes models so that we allow
specific compute and storage sizes on Kubernetes VMs.

This change applies to both KubernetesCluster model for control plane
nodes and KubernetesNodepool model for worker nodes.

For both models, the compute specification is not null and default
is standard-2. The theory is that, kubernetes community usually are
not interested in the actual size of the nodes, and slap cluster
autoscaler to handle the number of nodes.

Usually, even less important, is the storage size of the nodes, so
that's left as null-able, as usually clover picks the lowest available
storage size of the selected compute size by default, so can we.


fdr (approved):
admittedly, I cannot read it deeply, but it seems pretty straight forward to me.

Maybe the right person for more subtle thoughts is @byucesoy, possibly @enescakir, whose ideas on how to represent data tied with invoicing & product lines / "skus" are more/the most advanced.

mohi-kalantari (approved):

migrate/20250130_kubernetes_vm_size.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:kubernetes_nodepool) do
+      add_column(:target_node_size, :text, collate: '"C"', null: false, default: "standard-2")
```
enescakir, line 6: Other models set the default value through clover instead of at the database level. Should we follow that approach and not set the default value in the database?

migrate/20250130_kubernetes_vm_size.rb
```diff
@@ -0,0 +1,15 @@
...
+      add_column(:target_node_storage_size_gib, :bigint, null: true)
```
enescakir, line 7: If clover always picks the lowest option and this column has values, I prefer to set null: false for consistency.

enescakir (approved):

migrate/20250130_kubernetes_vm_size.rb
aamederen: Done!

PR #2713, opened 2025-01-30
Use vcpus for VM allocation and topology

macieksarnowicz:
This patch switches the VM allocation from Cores to VCpus when selecting a host. There are two use cases motivating this change:
- we have x64 hosts that have threads_per_cores ratio of 1 (GEX44). That breaks the assumption encoded in the VmSizes, per architecture type
- we are going to introduce Burstable family, where relation between number of CPUs allocated for a VM and number of Cores allocated to a slice hosting that VM may vary per VM instance, regardless of the architecture.

With this change, the number of cores is computed during the allocation, based on the actual architecture of the candidate host and then updated back to the VM. In case when the VM is allocated in a slice, the number of cores is left as 0 on the VM, and instead, the number of cores is saved in the VmHostSlice, and that is subtracted from the host. At any point in time this should be true: `vm_host.used_cores == SUM(vm_host_slice.cores) + SUM(vm.cores if vm.vm_host_slice_id.nil?)`

This logic also helps us indicate who is really controlling the cores - it is either the VmHostSlice or a Vm running without the slice. Vms inside the slice, do not control the cores and relay on the slice instead.

The special case for vcpus==1 in `cloud_hypervisor_cpu_topology` is needed for Burstables, where we will have Burstable-1 size. I wanted to include this in the review together with this patch for completeness.

There is also a separate commit now to Losen the restrictions on slice allocation. We allow now projects with `use_slice_for_allocation` flag enabled to create Standard VMs on all hosts. When the host is marked with `accept_slices`, the allocation will create a VM inside a slice. When `accepts_slices` is off, the VM creation will use the earlier allocation logic.

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -104,10 +104,17 @@ class Vm < Sequel::Model
     total_packages = vm_host.total_sockets
 
     # Computed all-system statistics, now scale it down to meet VM needs.
-    proportion = Rational(cores) / vm_host.total_cores
+    if vcpus == 1 && threads_per_core > 1
+      # special case for single-threaded VMs
+      cores_from_cpus = Rational(vcpus)
+      threads_per_core = 1
+    else
+      cores_from_cpus = Rational(vcpus) / threads_per_core
+    end
+    proportion = cores_from_cpus / vm_host.total_cores
     packages = (total_packages * proportion).ceil
     dies_per_package = (total_dies_per_package * proportion).ceil
-    cores_per_die = Rational(cores) / (packages * dies_per_package)
+    cores_per_die = cores_from_cpus / (packages * dies_per_package)
     fail "BUG: need uniform number of cores allocated per die" unless cores_per_die.denominator == 1
 
     topo = [threads_per_core, cores_per_die, dies_per_package, packages].map { |num|
@@ -118,7 +125,7 @@ class Vm < Sequel::Model
     }
 
     # :nocov:
-    unless topo.reduce(&:*) == threads_per_core * cores
+    unless topo.reduce(&:*) == vcpus
       fail "BUG: arithmetic does not result in the correct number of vcpus"
     end
     # :nocov:
--- a/prog/test/vm_group.rb
+++ b/prog/test/vm_group.rb
@@ -66,8 +66,18 @@ class Prog::Test::VmGroup < Prog::Test::Base
 
   label def wait_verify_vms
     reap
-    hop_verify_vm_host_slices if leaf?
+    hop_verify_host_capacity if leaf?
     donate
+  end
+
+  label def verify_host_capacity
+    vm_cores = vm_host.vms.sum(&:cores)
+    slice_cores = vm_host.slices.sum(&:cores)
+    spdk_cores = vm_host.cpus.count { _1.spdk } * vm_host.total_cores / vm_host.total_cpus
+
+    fail_test "Host used cores does not match the allocated VMs cores" if vm_cores + slice_cores + spdk_cores != vm_host.used_cores
+
+    hop_verify_vm_host_slices
   end
 
   label def verify_vm_host_slices
@@ -77,7 +87,7 @@ class Prog::Test::VmGroup < Prog::Test::Base
       hop_verify_firewall_rules
     end
 
-    slices = frame["vms"].map { Vm[_1].vm_host_slice.id }
+    slices = frame["vms"].map { Vm[_1].vm_host_slice&.id }.reject(&:nil?)
     push Prog::Test::VmHostSlices, {"slices" => slices}
   end
 
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -130,7 +130,8 @@ module Scheduling::Allocator
           Sequel.function(:coalesce, :num_gpus, 0).as(:num_gpus),
           Sequel.function(:coalesce, :available_gpus, 0).as(:available_gpus),
           :available_iommu_groups,
-          Sequel.function(:coalesce, :vm_provisioning_count, 0).as(:vm_provisioning_count)
+          Sequel.function(:coalesce, :vm_provisioning_count, 0).as(:vm_provisioning_count),
+          :accepts_slices
         )
         .where(arch: request.arch_filter)
         .where { (total_hugepages_1g - used_hugepages_1g >= request.memory_gib) }
@@ -179,8 +180,6 @@ module Scheduling::Allocator
       ds = ds.exclude(Sequel[:vm_host][:id] => request.host_exclusion_filter) unless request.host_exclusion_filter.empty?
       ds = ds.where(location: request.location_filter) unless request.location_filter.empty?
       ds = ds.where(allocation_state: request.allocation_state_filter) unless request.allocation_state_filter.empty?
-      # Match the slice allocation to the hosts that can accept it
-      ds = ds.where(accepts_slices: request.use_slices)
 
       # Emit the allocation query if the project is flagged for
       # diagnostics.
@@ -218,7 +217,7 @@ module Scheduling::Allocator
       @device_allocations = [StorageAllocation.new(candidate_host, request)]
       @device_allocations << GpuAllocation.new(candidate_host, request) if request.gpu_count > 0
 
-      if request.use_slices
+      if request.use_slices && candidate_host[:accepts_slices]
         # Wrap around and replace the host allocations. That way we can control that logic from the slice POV
         @vm_host_allocations = [VmHostSliceAllocation.new(candidate_host, request, @vm_host_allocations)]
       end
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -15,7 +15,7 @@ module Scheduling::Allocator
   def self.allocate(vm, storage_volumes, distinct_storage_devices: false, gpu_count: 0, allocation_state_filter: ["accepting"], host_filter: [], host_exclusion_filter: [], location_filter: [], location_preference: [])
     request = Request.new(
       vm.id,
-      vm.cores,
+      vm.vcpus,
       vm.memory_gib,
       storage_volumes.map { _1["size_gib"] }.sum,
       storage_volumes.size.times.zip(storage_volumes).to_h.sort_by { |k, v| v["size_gib"] * -1 },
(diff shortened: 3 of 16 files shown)


fdr:
note: GEX doesn't have one thread per core, it has two threads on some cores and one thread on another core. But you can degenerate the problem (with some loss in efficiency) by saying: all cores have *at least* one thread.

spec/prog/vm/allocator.sql
```diff
@@ -0,0 +1,140 @@
+WITH "total_ipv4" AS (
```
fdr, line 1: why is this file here?

macieksarnowicz: Please ignore it :)

macieksarnowicz:
There is one problem with tests that I need some help with. @pykello , @jeremyevans - maybe you can help

All specs are passing on the first run (COVERAGE=1) with 100% coverage. On the second run, with CLOVER_FREEZE=1 four tests are failing (see below). All four failures are due to the same reason - `vm.cores` value is not updated. That is related to my changes (change in allocator.rb:311).

Q: what does CLOVER_FREEZE=1 changes that would cause this specific DB update to not run? 
Q: how to run individual spec with this configuration? When I try, I get an error `uninitialized constant Aws1`

```
 1) Scheduling::Allocator update updates resources
     Failure/Error: expect(used_cores + vm.cores).to eq(vmh.used_cores)
     
       expected: 2
            got: 1
     
       (compared using ==)
     # ./spec/scheduling/allocator_spec.rb:556:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:51:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:50:in `block (2 levels) in <top (required)>'

  2) Scheduling::Allocator update allocates correctly on GEX44 host
     Failure/Error: expect(vm.cores).to eq(2)
     
       expected: 2
            got: 0
     
       (compared using ==)
     # ./spec/scheduling/allocator_spec.rb:757:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:51:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:50:in `block (2 levels) in <top (required)>'

  3) Scheduling::Allocator update allows concurrent allocations
     Failure/Error: expect(used_cores + vm1.cores + vm2.cores).to eq(vmh.used_cores)
     
       expected: 3
            got: 1
     
       (compared using ==)
     # ./spec/scheduling/allocator_spec.rb:593:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:51:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:50:in `block (2 levels) in <top (required)>'

  4) Scheduling::Allocator update updates pci devices
     Failure/Error: expect(used_cores + vm.cores).to eq(vmh.used_cores)
     
       expected: 2
            got: 1
     
       (compared using ==)
     # ./spec/scheduling/allocator_spec.rb:575:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:51:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:50:in `block (2 levels) in <top (required)>'
```

jeremyevans:
Use `FORCE_AUTOLOAD=1 CLOVER_FREEZE=1` when running individual specs.  Also, try running with `RUBYOPT=-d` to see if there is an exception being raised when running frozen that is being automatically rescued.

scheduling/allocator.rb
```diff
@@ -291,6 +299,19 @@ def update(vm, vm_host)
     end
   end
 
+  class VmHostCpuAllocation < VmHostAllocation
+    def initialize(column, total, used, requested)
+      super
+    end
+
```
jeremyevans, line 306:
```suggestion
```

A method that only calls super is unnecessary and only hurts performance.

macieksarnowicz:
> Use `FORCE_AUTOLOAD=1 CLOVER_FREEZE=1` when running individual specs. Also, try running with `RUBYOPT=-d` to see if there is an exception being raised when running frozen that is being automatically rescued.

Thanks @jeremyevans ! This helped a lot. Now that I can debug this, I can see that when running with CLOVER_FREEZE my new class VmHostCpuAllocation is not called at all. Instead, the base class VmHostAllocation is used. So, the vm.update method is never called and does not update the number of cores. When running without CLOVER_FREEZE, it all works as expected, with the subclass called. I feel like this is way beyond my knowledge of Ruby :)

jeremyevans:
> I have narrowed it down to the thawed_mock.rb:112 `allow_mocking(Scheduling::Allocator::VmHostAllocation, :new)` If that line is commented out, my tests are passing in the CLOVER_FREEZE=1 mode. However, that does not solve the problem, as it breaks the tests that actually use the mocking of `new` operator. My test does not use mocking. So, I am not clear why allowing mocking also prevents the subclassing.

The way thawed_mock is implemented is it basically makes static the current receiver of the method, which will negatively affect subclassing.  I think adding `allow_mocking(Scheduling::Allocator::VmHostCpuAllocation, :new)` above the existing `allow_mocking(Scheduling::Allocator::VmHostAllocation, :new)`  should fix it.

macieksarnowicz:
> The way thawed_mock is implemented is it basically makes static the current receiver of the method, which will negatively affect subclassing. I think adding `allow_mocking(Scheduling::Allocator::VmHostCpuAllocation, :new)` above the existing `allow_mocking(Scheduling::Allocator::VmHostAllocation, :new)` should fix it.

Thank you! This indeed fixed the problem. I had to update the tests that used mocking of the VmHostAllocaiton, but it is all working now.

scheduling/allocator.rb
macieksarnowicz: Fixed.

prog/test/vm_group.rb
```diff
@@ -66,10 +66,20 @@ def self.assemble(boot_images:, storage_encrypted: true, test_reboot: true, test
...
   label def wait_verify_vms
     reap
-    hop_verify_vm_host_slices if leaf?
+    hop_verify_host_capacity if leaf?
     donate
   end
 
+  label def verify_host_capacity
+    vm_cores = vm_host.vms.reduce(0) { |sum, vm| sum + vm.cores }
+    slice_cores = vm_host.slices.reduce(0) { |sum, slice| sum + slice.cores }
```
enescakir, line 75: nit:
```suggestion
    vm_cores = vm_host.vms.sum(&:cores)
    slice_cores = vm_host.slices.sum(&:cores)
```

scheduling/allocator.rb
```diff
@@ -339,6 +352,9 @@ def update(vm, vm_host)
           used_cpu_percent: Sequel[:used_cpu_percent] + vm.cpu_percent_limit,
           used_memory_gib: Sequel[:used_memory_gib] + vm.memory_gib
         )
+
+        # Update the host utilization
+        @vm_host_allocations.each { _1.update(vm, vm_host) }
```
bsatzger, line 357: The vm now needs to be updated before hosts, otherwise this won't be correct anymore. that is a little fragile.

scheduling/allocator.rb
```diff
@@ -79,6 +79,10 @@ def memory_gib_for_cores
 
       (cores * memory_gib_ratio).to_i
     end
+
+    def cores_for_vcpus(threads_per_core)
+      [1, vcpus / threads_per_core].max
```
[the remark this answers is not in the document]
macieksarnowicz: @bsatzger - I was looking at the vm_sizes that we have, which all are even numbers + the one I will introduce that is of 1 vcpu. I can try using something like `(Float(vcpus)/threads_per_core).ceil` here and look at the other places to make sure we handle the odd number of vcpus correctly. 
However it is not clear if there ever be a VM size of 3 or 5 or other odd numbers.

model/vm.rb
```diff
@@ -104,10 +104,17 @@ def cloud_hypervisor_cpu_topology
     total_packages = vm_host.total_sockets
 
     # Computed all-system statistics, now scale it down to meet VM needs.
-    proportion = Rational(cores) / vm_host.total_cores
+    if vcpus == 1 && threads_per_core > 1
```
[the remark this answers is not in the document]
macieksarnowicz: It will fail with an error: "BUG: need uniform number of cores allocated per die"
I added an extra test to make this more clear.

prog/test/vm_group.rb
macieksarnowicz: Fixed!

scheduling/allocator.rb
bsatzger: > Are you saying it is fragile because this order is required, which is not evident from reading this method?
Yes. Adding a comment would be good.

scheduling/allocator.rb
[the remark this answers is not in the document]
bsatzger: we could maybe just fail if we get an odd number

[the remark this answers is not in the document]
macieksarnowicz: @bsatzger - I am struggling on where to add that check. I first thought I will just put it in `Request.initialize` but that feels strange, as we have a large number of parameters there and only this one would be checked. We I think assume all the validation happens earlier. Then, if we move up the stack, we could include it in `Nexus.assemble`. The logical place seems to be `Validation.validate_vm_size`, but then again, why check a single parameters out of the whole vm_size structure. The structure is frozen and user cannot modify it. Since all creation of VMs happens with vm_size name, there is really no way for a user to specify an odd number of vcpus. 
If you have a suggestion on where such check would make sense, please let me know.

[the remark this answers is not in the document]
pykello: One other option (just a suggestion) is to add an RSpec test and iterate over all VmSizes & check their sizes are either 1 or even, with a comment on why we need this assumption. So, if someone adds a new size in future, they will need to think more if tests fail.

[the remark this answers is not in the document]
bsatzger: Agreed, a test that checks something like this:
```ruby
Option::VmSizes.all? { _1.vcpus == 1 || _1.vcpus.even? }
```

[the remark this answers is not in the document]
macieksarnowicz: sounds good, I can add that

pykello (approved):
* E2E succeeded: https://github.com/ubicloud/ubicloud/actions/runs/13166972069
* Manually tested for cpu/core cases of 96/48 and 14/14.

PR #2714, opened 2025-01-30
Catch unique constraint exception raised by database

enescakir:
We had a similar issue for GitHub cache entries and fixed at b73dd78.

If a customer tries to create multiple virtual machines with the same name at the same time, a race condition could cause two threads to try to create the same VM, resulting in an HTTP 500 error. We encountered this issue in production.

`Sequel::ValidationFailed` exception is raised when unique validation fails at the Sequel model level. `Sequel::UniqueConstraintViolation` exception is raised when uniqueness validation passes at the Sequel model level but fails at the database level due to concurrency.

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -161,7 +161,7 @@ class Clover < Roda
     end
 
     case e
-    when Sequel::ValidationFailed, Roda::RodaPlugins::InvalidRequestBody::Error
+    when Sequel::ValidationFailed, Sequel::UniqueConstraintViolation, Roda::RodaPlugins::InvalidRequestBody::Error
       code = 400
       type = "InvalidRequest"
       message = e.to_s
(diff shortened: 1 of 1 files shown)


jeremyevans (commented):
One issue with this it exposes the database error message directly to the user, which is not good.  I think we should consider using the Sequel pg_auto_constraint_validations plugin instead of this change (https://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/PgAutoConstraintValidations.html).  The plugin can catch additional constraint violations, and it associates the error with the specific fields related.  

Let me know what you think about that approach.  I can work on a pull request if you agree with it.

enescakir:
Let's use `pg_auto_constraint_validations`

PR #2716, opened 2025-01-31
Use Roda part plugin to simplify render calls with locals

jeremyevans:
This uses the new part plugin to simplify and optimize render
calls with locals.

```ruby
render(:template, locals: {foo: 'bar'})

part(:template, foo: 'bar')
```

This simplifies a large number of calls in Clover, since
rendering with locals is one of the most common method
calls in the templates.

The main advantage of this is simplicity, but the part method
is also more optimized, and will be even more optimized when
we upgrade to Ruby 3.4.

Diff best reviewed with:

```
git diff -b --color-words --word-diff-regex='\\w+|[^[:space:]]'
```

Split into 2 commits for easier review, the second commit contains
the ERBFormatter changes.  Will squash into a single commit before
merging.

I'm open to changing the name of the `part` method if anyone can
think of a better one that is also short (no longer than 6
characters).


enescakir (approved):
I love it 🚀

PR #2717, opened 2025-01-31
Remove extra bash indirection in nft invocation

fdr:
I don't think putting .shellescaped quoted text inside more quotes is always sound (it may go into the argument vector over-quoted), but besides that, we don't use any shell features in the invocation, so there's no need to start up bash at all.

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -429,8 +429,8 @@ add element inet drop_unused_ip_packets allowed_ipv4_addresses { #{ip_net} }
   end
 
   def apply_nftables
-    r "ip netns exec #{q_vm} bash -c 'nft flush ruleset'"
-    r "ip netns exec #{q_vm} bash -c 'nft -f #{vp.q_nftables_conf}'"
+    r "ip netns exec #{q_vm} nft flush ruleset"
+    r "ip netns exec #{q_vm} nft -f #{vp.q_nftables_conf}"
   end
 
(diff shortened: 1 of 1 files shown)


fdr:
another improvement here might be to work the flush into the same file, that makes it atomic, and probably faster, IIRC every `nft -f -` is pretty darn slow

furkansahin (approved):
Thanks!

jeremyevans:
> we don't use any shell features in the invocation, so there's no need to start up bash at all.

Note that I believe `r` always starts `bash` (or the default shell).  We could prevent this by having `r` take multiple arguments/array instead of a single argument, so it can run the program directly without spawning a shell.

fdr:
I amended the commit message so someone wouldn't over-interpret what was going on (the existence of zero shells, rather than the removal of one additional shell).

I used to be dogmatic about using argv rather than the shell when I used python, and maybe should be again, is there a version of `%w[]` that can also accept expressions for evaluation for a list element? Like the quasi-quote feature in lisp.

jeremyevans:
> I amended the commit message so someone wouldn't over-interpret what was going on (the existence of zero shells, rather than the removal of one additional shell).
> 
> I used to be dogmatic about using argv rather than the shell when I used python, and maybe should be again, is there a version of `%w[]` that can also accept expressions for evaluation for a list element? I found toggling from %w[] to ['quoted', 'quoted', 'quoted', expr] too burdensome. Something like the quasi-quote feature in lisp.

`%W[]` allows interpolation inside (operates like Ruby `""` string, while `%w[]` operates like Ruby `''` string).

PR #2718, opened 2025-01-31
Allow multiple ports per LoadBalancer

mohi-kalantari:
As a k8s requirement, we need to define multiple ports per LB so customers can expose multiple services to the internet.

We will introduce a new model named LoadBalancersPorts which allows defining multiple ports. Port definition fields will also be removed from the LoadBalancer model. We will also


fdr:
why?

mohi-kalantari:
This is WIP but the goal is introducing LoadBalancerPorts so each loadbalancer can support multiple ports. it's a k8s requirement.

Each VM can only be selected as a LB upstream once, so a customer cannot create two LBs selecting its worker nodes, meaning they can only expose one service to the internet which I'd say most of the time, they need to expose more than one service. 

We had two options, one would be multiple ports per LB and the other would be allowing VMs to be selected as upstream more than once. considering the time and complexity, we thought we should move towards multiple ports per LB

@fdr

enescakir:
We can keep `WIP` PRs as drafts. I assume Daniel just expects some commit messages since the PR was created and not drafted

fdr:
Taking from the Linux tradition, which birthed git, if you think maybe
you’d need an email to explain why you’re doing what you’re doing, stick
that in the commit (or if they need serious restructuring, into the PR text)

The draft status is useful so I can read the patch for strategy more than
tactics.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_table(:load_balancer_port) do
+      uuid :id, primary_key: true
+      uuid :load_balancer_id, null: false
+      Integer :src_port, null: false
+      Integer :dst_port, null: false
+      String :health_check_endpoint, text: true, null: false, default: "/up"
```
furkansahin, line 10: Are we going to have different endpoints per port mapping? At this point, we are going into simply having multiple load balancers per VM. If different set of VMs respond to different endpoints, they have different port mapping, I am not sure if this is the right way to go.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      check Sequel.lit("health_check_timeout > 0"), name: :chk_health_check_timeout
+      check Sequel.lit("health_check_interval > 0 AND health_check_interval < 600"), name: :chk_health_check_interval
+      check Sequel.lit("health_check_timeout <= health_check_interval"), name: :chk_timeout_interval
+      check Sequel.lit("health_check_up_threshold > 0"), name: :chk_health_check_up_threshold
+      check Sequel.lit("health_check_down_threshold > 0"), name: :chk_health_check_down_threshold
+    end
+
+    create_table(:load_balancer_vm_port) do
+      uuid :id, primary_key: true
```
furkansahin, line 28: are we adding different port assignments to different VMs in that same load balancer? Again, how different is this then to have all together multiple load balancers per vm?

helpers/load_balancer.rb
```diff
@@ -60,10 +57,10 @@ def generate_load_balancer_options
...
     options.add_option(name: "algorithm", values: ["Round Robin", "Hash Based"].map { {value: _1.downcase.tr(" ", "_"), display_name: _1} })
     options.add_option(name: "stack", values: [LoadBalancer::Stack::IPV4, LoadBalancer::Stack::IPV6, LoadBalancer::Stack::DUAL].map { {value: _1.downcase, display_name: _1.gsub("ip", "IP")} })
-    options.add_option(name: "src_port")
-    options.add_option(name: "dst_port")
-    options.add_option(name: "health_check_endpoint")
-    options.add_option(name: "health_check_protocol", values: ["http", "https", "tcp"].map { {value: _1, display_name: _1.upcase} })
+    # options.add_option(name: "src_port")
+    # options.add_option(name: "dst_port")
+    # options.add_option(name: "health_check_endpoint")
+    # options.add_option(name: "health_check_protocol", values: ["http", "https", "tcp"].map { {value: _1, display_name: _1.upcase} })
```
furkansahin: So, we are not getting load balancer port mapping option at the creation time? Why?

furkansahin (changes requested):
I feel like this is very complicated to solve the need. Looking into the implementation, I have more questions now regarding the real need. What if we implemented multiple load balancers per VM instead? Can you please show case a scenario where this actually helps with example ip addresses, port numbers, load balancing need. Maybe even with a drawing.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,77 @@
...
+      set_column_allow_null :health_check_endpoint
+      set_column_allow_null :health_check_interval
+      set_column_allow_null :health_check_timeout
+      set_column_allow_null :health_check_up_threshold
+      set_column_allow_null :health_check_down_threshold
+      set_column_allow_null :health_check_protocol
+    end
+
+    alter_table(:load_balancers_vms) do
+      drop_column :state
```
jeremyevans, line 54: This seems dangerous.  Isn't this column currently used at the time the migration will be run?  If so, you cannot drop it yet.  You need to wait until later, after it is no longer used, to drop it.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+  end
+
+  down do
+    drop_table(:load_balancer_vm_port)
+    drop_table(:load_balancer_port)
+
+    alter_table(:load_balancer) do
+      set_column_not_null :src_port
```
jeremyevans, line 63: This will fail if there any NULL values in the columns, making the migration irreversible.  You can set the column values to default values if they are NULL before this to avoid the problem.

migrate/20250225_add_loadbalancer_ports.rb
jeremyevans, line 1: The commit message for says this adds models, when this only adds a migration.

model/load_balancer.rb
```diff
@@ -6,15 +6,18 @@ class LoadBalancer < Sequel::Model
   many_to_one :project
   many_to_many :vms
   many_to_many :active_vms, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: {state: "up"}
-  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: Sequel.any_type(["evacuating", "detaching"], :lb_node_state))
+  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancer_vm_port, conditions: Sequel.~(state: Sequel.any_type(["evacuating", "detaching"], :lb_node_state)), select: Sequel.lit("DISTINCT vm.*")
```
jeremyevans, line 9: You can use `distinct: true` instead . `Sequel.lit` is a nice backdoor when Sequel cannot do what you need directly, but it's best to only use it in such circumstances.

prog/vnet/load_balancer_nexus.rb
```diff
@@ -27,14 +29,26 @@ def self.assemble(private_subnet_id, name: nil, algorithm: "round_robin", src_po
...
+          src_port: port["src_port"],
+          dst_port: port["dst_port"],
+          health_check_endpoint: port["health_check_endpoint"],
+          health_check_interval: port["health_check_interval"],
+          health_check_timeout: port["health_check_timeout"],
+          health_check_up_threshold: port["health_check_up_threshold"],
+          health_check_down_threshold: port["health_check_down_threshold"],
+          health_check_protocol: port["health_check_protocol"]
+        }.delete_if { |_, v| v.nil? }
+        LoadBalancerPort.create_with_id(attrs)
```
jeremyevans, line 49: `create_with_id` is only for backwards compatibility.  You can switch to `create`.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -7,23 +7,19 @@ def load_balancer
...
-    load_balancer.load_balancers_vms_dataset[vm_id: vm.id].state
-  end
-
   def before_run
     pop "VM is destroyed" unless vm
   end
 
   label def update_load_balancer
-    if vm_load_balancer_state == "detaching"
+    if vm.load_balancer_ports.any? { |lb_port| lb_port.state == "detaching" }
```
jeremyevans, line 15: If VM has two load balancer ports, and one is detaching and one is not, do we want to disassociate the vm from the load balancer and remove the load balancer?  What if the load balancer is servicing other VMs?  

Looking at the associations, it appears that `remove_vm` only affects `load_balancers_vms` and not `load_balancer_port`/`load_balancer_port_vm`, which seems odd considering the check.

views/networking/load_balancer/show.erb
```diff
@@ -25,9 +25,9 @@
       ],
       ["Algorithm", (@lb[:algorithm] == "round_robin") ? "Round Robin" : "Hash Based"],
       ["Stack", @lb[:stack].gsub("ip", "IP")],
-      ["Load Balancer Port", @lb[:src_port]],
-      ["Application Port", @lb[:dst_port]],
-      ["HTTP Health Check Endpoint", @lb[:health_check_endpoint]]
+      ["Load Balancer Port", @lb[:ports].first[:src_port]],
```
jeremyevans, line 28: I assume this should be changed to list all ports the load balancer is dealing with?

jeremyevans (commented):
I'll echo Furkan's concerns.  Is there a reason k8s couldn't use separate load balancers per port/VM?  Currently `load_balancers_vms` has a unique constraint on `vm_id`, but if we removed that constraint, and otherwise kept the schema the same, I think fewer code changes would be required.

views/networking/load_balancer/show.erb
mohi-kalantari: We wanted minimal changes in UI and didn't want to expose the new feature to the regular customers. that's the reason.

model/load_balancer.rb
```diff
@@ -29,28 +32,30 @@ def path
   def add_vm(vm)
     DB.transaction do
       LoadBalancersVms.create_with_id(load_balancer_id: id, vm_id: vm.id)
+      ports.each { |port| LoadBalancerVmPort.create_with_id(vm_id: vm.id, load_balancer_id: id, load_balancer_port_id: port.id) }
```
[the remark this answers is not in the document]
mohi-kalantari: we needed that table to store the state of this tuple (vm, loadbalancer, port) since each one of these tuples need their own healthcheck

helpers/load_balancer.rb
mohi-kalantari: I had fixed it before asking for review. the comment is on an outdated file. you can check again and we provide the mapping options.

migrate/20250225_add_loadbalancer_ports.rb
mohi-kalantari: It makes sense to assume the endpoint should be the same for all the endpoints most of the times. but there can be cases where the http endpoint for healthcheck is different from https. 

Where do you suggest placing the field if we don't want to store it in load_balancer_port? Keeping it inside the load balancer also seems off.

mohi-kalantari:
> I feel like this is very complicated to solve the need. Looking into the implementation, I have more questions now regarding the real need. What if we implemented multiple load balancers per VM instead? Can you please show case a scenario where this actually helps with example ip addresses, port numbers, load balancing need. Maybe even with a drawing.

There is a huge requirement in k8s which we cannot satisfy with multiple loadbalancers per vm and that is ingress

For ingress controller to work correctly, it must be reachable through an ip with both ports 80 and 443 open (80 for the acme http challenge and 443 for the actual applications)

we might say we can have multiple lbs for vms and create one for each port but that makes the dns routing strange because for each port of the lb, you need to have separate dns records.

@furkansahin

lib/validation.rb
```diff
@@ -193,6 +193,16 @@ def self.validate_port_range(port_range)
     end_port ? [start_port, end_port] : [start_port]
   end
 
+  def self.validate_load_balancer_ports(ports)
+    ports.map do |port|
+      port["src_port"] = validate_port(:src_port, port["src_port"])
+      port["dst_port"] = validate_port(:dst_port, port["dst_port"])
+      port["health_check_endpoint"] = port["health_check_endpoint"] || Prog::Vnet::LoadBalancerNexus::DEFAULT_HEALTH_CHECK_ENDPOINT
+      port["health_check_protocol"] ||= port["health_check_protocol"]
+      port
```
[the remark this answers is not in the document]
mohi-kalantari: fixed thanks.

migrate/20250225_add_loadbalancer_ports.rb
mohi-kalantari: you're right I missed that. fixed it. thanks.

migrate/20250225_add_loadbalancer_ports.rb
mohi-kalantari: fixed it thanks.

migrate/20250225_add_loadbalancer_ports.rb
mohi-kalantari: fixed it thanks.

prog/vnet/load_balancer_nexus.rb
mohi-kalantari: fixed it thanks.

prog/vnet/update_load_balancer_node.rb
mohi-kalantari: you're right.I updated the code to just remove the LoadBalancerVmPort we are working on. thanks.

mohi-kalantari:
> I'll echo Furkan's concerns. Is there a reason k8s couldn't use separate load balancers per port/VM? Currently `load_balancers_vms` has a unique constraint on `vm_id`, but if we removed that constraint, and otherwise kept the schema the same, I think fewer code changes would be required.

Yes. For ingress controllers we need to have multiple ports per lb. One would be used for http challenge (port 80) and another (port 443) for the actual application.

@jeremyevans

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    alter_table(:load_balancers_vms) do
+      set_column_allow_null :state
+    end
+
+    DB[:load_balancer_port].insert(
+      [:id, :load_balancer_id, :src_port, :dst_port],
+      DB[:load_balancer].select(
+        UBID.generate(UBID::TYPE_LOAD_BALANCER_PORT).to_uuid,
```
furkansahin, line 42: At the time of this migration, the `UBID::TYPE_LOAD_BALANCER_PORT` is not defined. So, this migration will fail. Can you please test the migration on a system with load balancers.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,84 @@
...
+    )
+    DB[:load_balancers_vms]
+      .join(:load_balancer_port, load_balancer_id: :load_balancer_id)
+      .select(
+        Sequel[:load_balancer_port][:id].as(:load_balancer_port_id),
+        Sequel[:load_balancers_vms][:id].as(:load_balancer_vm_id),
+        Sequel[:load_balancers_vms][:state]
+      ).each do |row|
+      DB[:load_balancer_vm_port].insert(
+        id: UBID.generate(UBID::TYPE_LOAD_BALANCER_VM_PORT).to_uuid,
```
furkansahin, line 56: same thing here.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,84 @@
...
+      set_column_default :dst_port, 80
+
+      set_column_not_null :src_port
+      set_column_not_null :dst_port
+    end
+
+    DB[:load_balancers_vms].where(state: nil).update(state: "down")
+    alter_table(:load_balancers_vms) do
+      set_column_not_null :state
+    end
```
furkansahin: We should update the `src_port`, `dst_port` columns using the `load_balancer_vm_port` table and then drop the tables.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -70,7 +70,7 @@
     DB[:load_balancer].where(dst_port: nil).update(dst_port: 8000)
     alter_table(:load_balancer) do
       set_column_default :src_port, 80
-      set_column_default :dst_port, 80
+      set_column_default :dst_port, 8000
```
furkansahin, line 73: this should be in the first commit. Also, when are you going to drop these columns? You need a third commit with column drops.

cache/index.cache
furkansahin, line 1: these changes should be appended to the migration commit.

model/load_balancer.rb
```diff
@@ -6,15 +6,34 @@ class LoadBalancer < Sequel::Model
...
   many_to_one :custom_hostname_dns_zone, class: :DnsZone, key: :custom_hostname_dns_zone_id
-
-  plugin :association_dependencies, load_balancers_vms: :destroy, certs_load_balancers: :destroy
+  many_to_many :vms_to_dns,
+    class: :Vm,
+    dataset: proc {
+      Vm.where(id: LoadBalancersVms.where(load_balancer_id: id)
+                                   .join(:load_balancer_vm_port, load_balancer_vm_id: :id)
+                                   .where(Sequel.~(Sequel[:load_balancer_vm_port][:state] => ["evacuating", "detaching"]))
+                                   .select(:vm_id))
```
furkansahin, line 22: we already have `load_balancers_vms`, why don't we reuse that?

model/load_balancer.rb
```diff
@@ -32,30 +51,48 @@ def path
 
   def add_vm(vm)
     DB.transaction do
-      LoadBalancersVms.create_with_id(load_balancer_id: id, vm_id: vm.id)
+      load_balancer_vm = LoadBalancersVms.create_with_id(load_balancer_id: id, vm_id: vm.id)
```
furkansahin, line 54: don't need to `create_with_id`

model/load_balancer.rb
```diff
@@ -32,30 +51,48 @@ def path
...
       Strand.create_with_id(prog: "Vnet::CertServer", label: "put_certificate", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
       incr_rewrite_dns_records
     end
   end
 
   def detach_vm(vm)
-    load_balancers_vms_dataset.where(vm_id: vm.id, state: Sequel.any_type(["up", "down", "evacuating"], :lb_node_state)).update(state: "detaching")
+    LoadBalancerVmPort.where(load_balancer_vm_id: LoadBalancersVms.where(load_balancer_id: id, vm_id: vm.id).select(:id))
+      .where(state: ["up", "down", "evacuating"])
+      .update(state: "detaching")
```
furkansahin, line 65: Refactor this using the `load_balancers_vms_dataset`

model/load_balancer.rb
```diff
@@ -32,30 +51,48 @@ def path
...
     Strand.create_with_id(prog: "Vnet::CertServer", label: "remove_cert_server", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
     incr_update_load_balancer
   end
 
   def evacuate_vm(vm)
     DB.transaction do
-      load_balancers_vms_dataset.where(vm_id: vm.id, state: Sequel.any_type(["up", "down"], :lb_node_state)).update(state: "evacuating")
+      LoadBalancerVmPort.where(load_balancer_vm_id: LoadBalancersVms.where(load_balancer_id: id, vm_id: vm.id).select(:id))
```
furkansahin, line 72: same here

furkansahin (changes requested):

migrate/20250225_add_loadbalancer_ports.rb
mohi-kalantari: I did try this on my local with load balancers on the old format and migrating it to the new model. since commits are both added, that worked fine. But I get the point thanks.

migrate/20250225_add_loadbalancer_ports.rb
mohi-kalantari: fixed. thanks.

migrate/20250225_add_loadbalancer_ports.rb
mohi-kalantari: fixed thanks.

serializers/vm.rb
```diff
@@ -29,7 +29,7 @@ def self.serialize_internal(vm, options = {})
     end
 
     if options[:load_balancer]
-      base[:load_balancer_state] = vm.load_balancers_vms.state
+      base[:load_balancer_state] = vm.load_balancer_vm_ports.first&.state
```
jeremyevans, line 32: Likewise, this seems odd.  We should show the state for all ports.  If we want to combine them, we should look at all ports and the state should reflect that (e.g. only show `up` if all ports are up, or do something like `vm.load_balancer_vm_ports.map(&:state).uniq.join(", ")`

routes/project/location/load_balancer.rb
```diff
@@ -74,12 +74,14 @@ class Clover
...
-          src_port: Validation.validate_port(:src_port, request_body_params["src_port"]),
-          dst_port: Validation.validate_port(:dst_port, request_body_params["dst_port"]),
-          health_check_endpoint: request_body_params["health_check_endpoint"]
-        )
+        DB.transaction do
+          lb.update(
+            algorithm: request_body_params["algorithm"],
+            health_check_endpoint: request_body_params["health_check_endpoint"]
+          )
+          lb.ports.first.update(src_port: Validation.validate_port(:src_port, request_body_params["src_port"]),
```
jeremyevans, line 82: If we support multiple ports per load balancer, why are we only updating the first?

Is the idea that customer load balancers only support a single port, but k8s load balancers support multiple ports?  If that's the case, maybe it would make more sense to have separate models for k8s load balancers compared to customer load balancers. Maybe the idea is that customer load balancers will eventually support multiple ports, and we'll update this later at that point?

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    alter_table(:load_balancers_vms) do
+      set_column_allow_null :state
+    end
+
+    DB[:load_balancer_port].import(
+      [:id, :load_balancer_id, :src_port, :dst_port],
+      DB[:load_balancer].select(
+        UBID.generate(TYPE_LOAD_BALANCER_PORT).to_uuid,
```
mohi-kalantari, line 46: I duplicated the Constants to here but what should I do with the `UBID.generate` funciton? Should I duplicate that too? @jeremyevans

prog/vnet/load_balancer_nexus.rb
```diff
@@ -8,37 +8,55 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
 
   subject_is :load_balancer
 
-  def self.assemble(private_subnet_id, name: nil, algorithm: "round_robin", src_port: nil, dst_port: nil,
+  def self.assemble_with_multiple_ports(private_subnet_id, name: nil, algorithm: "round_robin",
```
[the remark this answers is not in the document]
mohi-kalantari: I'm planning to use it in Kubernetes progs.

serializers/vm.rb
mohi-kalantari: We plan not to release this feature to customers and just use it in the k8s product for now. that's why we are always going to show the first port.

Load balancers created for k8s will also reside in an internal project customers cannot see.

We are aiming to keep the LB ui simple

prog/vnet/load_balancer_nexus.rb
```diff
@@ -8,37 +8,55 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
...
       Strand.create(prog: "Vnet::LoadBalancerNexus", label: "wait") { _1.id = lb.id }
     end
   end
 
+  def self.assemble(private_subnet_id, name: nil, algorithm: "round_robin",
+    health_check_endpoint: DEFAULT_HEALTH_CHECK_ENDPOINT, health_check_interval: 30, health_check_timeout: 15,
+    health_check_up_threshold: 3, health_check_down_threshold: 2, health_check_protocol: "http", src_port: nil, dst_port: nil,
+    custom_hostname_prefix: nil, custom_hostname_dns_zone_id: nil, stack: LoadBalancer::Stack::DUAL)
+
+    load_balancer_ports = [{"src_port" => src_port, "dst_port" => dst_port}]
```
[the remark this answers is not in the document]
mohi-kalantari: I'll need the multiple ports because It'll be used in k8s progs.

I was thinking about an API which gets boths ports list and src_port/dst_port and all of them are optional. There will be some checks making sure one of the is populated. Although I do not like it that much.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,93 @@
...
+        :src_port,
+        :dst_port
+      )
+    )
+    DB[:load_balancers_vms]
+      .join(:load_balancer_port, load_balancer_id: :load_balancer_id)
+      .select(
+        Sequel[:load_balancer_port][:id].as(:load_balancer_port_id),
+        Sequel[:load_balancers_vms][:id].as(:load_balancer_vm_id),
+        Sequel[:load_balancers_vms][:state]
```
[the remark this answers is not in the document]
mohi-kalantari: If I understood correctly, I need to do the migration in 3 different steps right?

At first I was being a bit reckless since we have few lbs in production and was counting on nothing happening during that gap. which is of course not optimal 😅

model/load_balancer.rb
```diff
@@ -5,16 +5,18 @@
...
-  many_to_many :vms_to_dns, class: :Vm, left_key: :load_balancer_id, right_key: :vm_id, join_table: :load_balancers_vms, conditions: Sequel.~(state: Sequel.any_type(["evacuating", "detaching"], :lb_node_state))
   one_to_one :strand, key: :id
   many_to_one :private_subnet
   one_to_many :load_balancers_vms, key: :load_balancer_id, class: :LoadBalancersVms
+  one_to_many :ports, key: :load_balancer_id, class: :LoadBalancerPort
   many_to_many :certs, join_table: :certs_load_balancers, left_key: :load_balancer_id, right_key: :cert_id
   one_to_many :certs_load_balancers, key: :load_balancer_id, class: :CertsLoadBalancers
   many_to_one :custom_hostname_dns_zone, class: :DnsZone, key: :custom_hostname_dns_zone_id
+  many_through_many :vm_ports, [[:load_balancer_port, :load_balancer_id, :id], [:load_balancer_vm_port, :load_balancer_port_id, :id]], class: :LoadBalancerVmPort, read_only: true
+  many_through_many :active_vm_ports, [[:load_balancer_port, :load_balancer_id, :id], [:load_balancer_vm_port, :load_balancer_port_id, :id]], class: :LoadBalancerVmPort, read_only: true, conditions: {Sequel.qualify(:load_balancer_vm_port, :state) => "up"}
```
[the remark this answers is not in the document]
jeremyevans: Yes, this looks good.

migrate/20250225_add_loadbalancer_ports.rb
jeremyevans: I don't think we need to duplicate that.  I think it is safe to assume the UBID.generate API will not change through the life of the application, and in the off chance it does, we could update the migration.

migrate/20250225_add_loadbalancer_ports.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+TYPE_LOAD_BALANCER_PORT = "1p"
```
jeremyevans, line 3: As you are referencing UBID below, you will need the `require_relative "../ubid"`.  That's the cause of the PR failing CI currently.

jeremyevans (commented):
A lot of the cleanups look good.  I still have a few more comments and recommendations.

prog/vnet/load_balancer_nexus.rb
```diff
@@ -24,21 +24,35 @@ def self.assemble(private_subnet_id, name: nil, algorithm: "round_robin", src_po
...
+        stack: stack, project_id: ps.project_id,
+        health_check_endpoint: health_check_endpoint,
+        health_check_interval: health_check_interval,
+        health_check_timeout: health_check_timeout,
+        health_check_up_threshold: health_check_up_threshold,
+        health_check_down_threshold: health_check_down_threshold,
+        health_check_protocol: health_check_protocol
       )
-
+      ports.map { |port| LoadBalancerPort.create(load_balancer_id: lb.id, src_port: port["src_port"], dst_port: port["dst_port"]) }
```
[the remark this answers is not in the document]
mohi-kalantari: changed it to arrays and it seems way neater. thanks.

spec/lib/validation_spec.rb
```diff
@@ -294,6 +294,12 @@
       end
     end
 
+    describe "#validate_load_balancer_ports" do
+      it "validates list of ports" do
+        expect { described_class.validate_load_balancer_ports([[80, 80]]) }.to raise_error described_class::ValidationFailed
```
furkansahin, line 299: so, we cannot use the load balancer without port rewriting, why?

model/load_balancer.rb
```diff
@@ -30,33 +32,59 @@ def path
...
       Strand.create_with_id(prog: "Vnet::CertServer", label: "remove_cert_server", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
       incr_update_load_balancer
       incr_rewrite_dns_records
     end
   end
 
   def remove_vm(vm)
-    load_balancers_vms_dataset[vm_id: vm.id].destroy
-    incr_rewrite_dns_records
+    DB.transaction do
```
furkansahin, line 73: shouldn't this call `remove_vm_port`?

model/load_balancer_vm_port.rb
```diff
@@ -0,0 +1,90 @@
...
+      (session[:ssh_session].exec!(health_check_cmd(type)).strip == "200") ? "up" : "down"
+    rescue
+      "down"
+    end
+  end
+
+  def health_check_cmd(type)
+    address = (type == :ipv4) ? vm.private_ipv4 : vm.ephemeral_net6.nth(2)
+    if load_balancer.health_check_protocol == "tcp"
+      "sudo ip netns exec #{vm.inhost_name} nc -z -w #{load_balancer.health_check_timeout} #{address} #{load_balancer_port.dst_port} && echo 200 || echo 400"
```
furkansahin, line 51: don't forget about this https://github.com/ubicloud/ubicloud/pull/3030/files

prog/vnet/load_balancer_nexus.rb
```diff
@@ -142,6 +155,9 @@ def need_to_rewrite_dns_records?
       dns_zone.delete_record(record_name: load_balancer.hostname)
     end
 
+    load_balancer.vm_ports.map(&:destroy)
+    load_balancer.ports.map(&:destroy)
```
furkansahin, line 159: there is already a plugin association for `ports` in `model/load_balancer.rb`, why do we need this?

prog/vnet/update_firewall_rules.rb
```diff
@@ -18,8 +18,15 @@ def before_run
...
-      allow_ipv4_lb_neigh_incoming = "ip saddr . tcp sport { #{vm.load_balancer.vms.reject { _1.id == vm.id }.map { "#{_1.private_ipv4} . #{vm.load_balancer.src_port}" }.join(", ")} } ct state established,related,new counter accept" if vm.load_balancer.vms.reject { _1.id == vm.id }.any?
-      allow_ipv6_lb_neigh_incoming = "ip6 saddr . tcp sport { #{vm.load_balancer.vms.reject { _1.id == vm.id }.map { "#{_1.private_ipv6} . #{vm.load_balancer.src_port}" }.join(", ")} } ct state established,related,new counter accept" if vm.load_balancer.vms.reject { _1.id == vm.id }.any?
+      neighbors = vm.load_balancer.vms.reject { _1.id == vm.id }
+      if neighbors.any?
+        allow_ipv4_lb_neigh_incoming = "ip saddr . tcp sport { #{neighbors.flat_map { |n|
+          vm.load_balancer.ports.map { |mapping| "#{n.private_ipv4} . #{mapping[:src_port]}" }
+        }.join(", ")} } ct state established,related,new counter accept"
+        allow_ipv6_lb_neigh_incoming = "ip6 saddr . tcp sport { #{neighbors.flat_map { |n|
+          vm.load_balancer.ports.map { |mapping| "#{n.private_ipv6} . #{mapping[:src_port]}" }
+        }.join(", ")} } ct state established,related,new counter accept"
```
furkansahin, line 28: how does this work? line 31 is not in the scope of this `if` and you define the variables inside?

model/load_balancer.rb
mohi-kalantari: it is doing it that but instead of a for loop, it queries directly on the database:

```
      vm_ports_by_vm(vm).destroy
```

prog/vnet/update_firewall_rules.rb
mohi-kalantari: you're right. it's strange it wasn't caught on the tests or my manual tests. will fix it.

furkansahin (approved):

PR #2722, opened 2025-02-03
Add destroying label to prevent hop to destroy again

enescakir:
- **Clean up vm pool nexus tests**
  

- **Fix destroy labels for the vm and inference endpoint**
  If the resource is already being destroyed, we don't want to hop to the
  `destroy` label again when the `destroy` semaphore is incremented.
  
  We check the current label to avoid hopping again. Some resources have
  labels after the `destroy` label, so we need to check those as well.
  
  We checked this for some resources, but it seems we missed it for the vm
  and inference endpoint.
  

- **Add destroying label to prevent hop to destroy again**
  If the resource is already being destroyed, we don't want to hop to the
  `destroy` label again when the `destroy` semaphore is incremented.
  
  We check the current label to avoid hopping again. Sometimes we forget
  to update the list of labels to check when we add new labels after a
  destroy.
  
  Instead of checking the labels separately, we can increase another
  semaphore, like `destroying`, along with the `destroy` semaphore and
  keep it incremented until the resource is destroyed.

### DIFF
--- a/prog/ai/inference_endpoint_nexus.rb
+++ b/prog/ai/inference_endpoint_nexus.rb
@@ -75,7 +75,7 @@ class Prog::Ai::InferenceEndpointNexus < Prog::Base
 
   def before_run
     when_destroy_set? do
-      if strand.label != "destroy"
+      unless ["destroy", "self_destroy"].include?(strand.label)
         hop_destroy
       end
     end
--- a/prog/ai/inference_endpoint_nexus.rb
+++ b/prog/ai/inference_endpoint_nexus.rb
@@ -75,7 +75,8 @@ class Prog::Ai::InferenceEndpointNexus < Prog::Base
 
   def before_run
     when_destroy_set? do
-      unless ["destroy", "self_destroy"].include?(strand.label)
+      unless inference_endpoint.destroying_set?
+        inference_endpoint.incr_destroying
         hop_destroy
       end
     end
--- a/prog/minio/minio_cluster_nexus.rb
+++ b/prog/minio/minio_cluster_nexus.rb
@@ -53,7 +53,8 @@ class Prog::Minio::MinioClusterNexus < Prog::Base
 
   def before_run
     when_destroy_set? do
-      unless ["destroy", "wait_pools_destroyed"].include?(strand.label)
+      unless minio_cluster.destroying_set?
+        minio_cluster.incr_destroying
         hop_destroy
       end
     end
--- a/prog/vm/vm_pool.rb
+++ b/prog/vm/vm_pool.rb
@@ -24,7 +24,8 @@ class Prog::Vm::VmPool < Prog::Base
 
   def before_run
     when_destroy_set? do
-      unless ["destroy", "wait_vms_destroy"].include?(strand.label)
+      unless vm_pool.destroying_set?
+        vm_pool.incr_destroying
         hop_destroy
       end
     end
--- a/model/ai/inference_endpoint.rb
+++ b/model/ai/inference_endpoint.rb
@@ -15,7 +15,7 @@ class InferenceEndpoint < Sequel::Model
   include SemaphoreMethods
   include ObjectTag::Cleanup
 
-  semaphore :destroy, :maintenance
+  semaphore :destroy, :maintenance, :destroying
 
   def display_location
     LocationNameConverter.to_display_name(location)
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -12,7 +12,7 @@ class GithubRunner < Sequel::Model
   include ResourceMethods
   include SemaphoreMethods
   include HealthMonitorMethods
-  semaphore :destroy, :skip_deregistration
+  semaphore :destroy, :skip_deregistration, :destroying
 
   def run_url
     "http://github.com/#{repository_name}/actions/runs/#{workflow_job["run_id"]}"
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -20,7 +20,7 @@ class LoadBalancer < Sequel::Model
   include SemaphoreMethods
   include ObjectTag::Cleanup
   dataset_module Pagination
-  semaphore :destroy, :update_load_balancer, :rewrite_dns_records, :refresh_cert
+  semaphore :destroy, :update_load_balancer, :rewrite_dns_records, :refresh_cert, :destroying
 
   def path
     "/location/#{private_subnet.display_location}/load-balancer/#{name}"
--- a/model/minio/minio_cluster.rb
+++ b/model/minio/minio_cluster.rb
@@ -16,7 +16,7 @@ class MinioCluster < Sequel::Model
   include ResourceMethods
   include SemaphoreMethods
 
-  semaphore :destroy, :reconfigure
+  semaphore :destroy, :reconfigure, :destroying
 
   plugin :column_encryption do |enc|
     enc.column :admin_password
--- a/model/minio/minio_pool.rb
+++ b/model/minio/minio_pool.rb
@@ -12,7 +12,7 @@ class MinioPool < Sequel::Model
   include ResourceMethods
   include SemaphoreMethods
 
-  semaphore :destroy, :add_additional_pool
+  semaphore :destroy, :add_additional_pool, :destroying
 
   def volumes_url
     return "/minio/dat1" if cluster.single_instance_single_drive?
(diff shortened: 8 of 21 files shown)


furkansahin (commented):
The last commit should be fixed like the following:
```
Add destroying semaphore to prevent hop to destroy again
```

Can we solve this differently instead of second semaphore? Maybe we can add a variable into the frame to better represent the state of the process instead of dancing with 2 semaphores? OR, we can make this part of the strand utility functions. If the hop_destroy is executed, we can immediately increment the `destroying` semaphore of that model, there. This way, if a model doesn't implement `destroying` semaphore, it will automatically fail during development.

prog/ai/inference_endpoint_nexus.rb
```diff
@@ -75,7 +75,8 @@ def self.assemble(project_id:, location:, boot_image:, name:, vm_size:, storage_
 
   def before_run
     when_destroy_set? do
-      unless ["destroy", "self_destroy"].include?(strand.label)
+      unless inference_endpoint.destroying_set?
+        inference_endpoint.incr_destroying
```
jeremyevans, line 79: Maybe we should have `incr_destroying` set at the top of the `destroy` method?  Having it set only in `before_run` will miss cases where we `hop_destroy` without `incr_destroy` being set (maybe there are no such cases currently, but it still seems safer).

prog/ai/inference_endpoint_nexus.rb
```diff
@@ -75,7 +75,8 @@ def self.assemble(project_id:, location:, boot_image:, name:, vm_size:, storage_
 
   def before_run
     when_destroy_set? do
-      unless ["destroy", "self_destroy"].include?(strand.label)
+      unless inference_endpoint.destroying_set?
```
jeremyevans, line 78: Maybe we should add `unless_*_set?` methods for semaphores?  The current code looks a bit inconsistent in terms of syntax, with `when_destroy_set? do` for the case where it is set, but `unless *.destroying_set?` in the case where it is not.  In addition to looking similar, this should make mocking in the specs easier.  Anyway, just an idea, up to you.

enescakir: I think we previously discussed the block method for checking the absence of the semaphores. If I remember correctly, @fdr might have some thoughts on it. @fdr, what do you think about the `unless_*_set?` methods?

Or should we run checks with `if inference_endpoint.destroyed_set? && ! inference_endpoint.destroying_set?` both with `subject.*_set? instead of mixing `when_*_set?` and `subject.*_set?`?

prog/ai/inference_endpoint_nexus.rb
enescakir: It makes sense, and it's also what Furkan suggested.
https://github.com/ubicloud/ubicloud/commit/b87897470cadb824be532e43fe62c17e9acd8987
I tried some PoC for the VM model, but I'm not sure about a few points. What do you think? Should we extract it to a separate module or implicitly use it as in the example

prog/ai/inference_endpoint_nexus.rb
byucesoy: Actually, all our uses of `when_*_set?` syntax can be replaced with if checks. I don't think we have a case where we don't inline the block, right? Should we remove `when_*_set?` completely? (Not necessarily in this PR)

prog/ai/inference_endpoint_nexus.rb
enescakir: @jeremyevans, @fdr What do you think about this approach? https://github.com/ubicloud/ubicloud/commit/b87897470cadb824be532e43fe62c17e9acd8987

jeremyevans: I like that approach.

enescakir: I refactored it to the new approach. @fdr -300 lines 🔥

prog/ai/inference_endpoint_nexus.rb
enescakir: @byucesoy As a first step I added `<SEM>_set?` helper to the base prog

enescakir:
> If the hop_destroy is executed, we can immediately increment the `destroying` semaphore of that model, there. This way, if a model doesn't implement `destroying` semaphore, it will automatically fail during development.

@furkansahin I implemented it

prog/base.rb
```diff
@@ -52,10 +57,17 @@ def self.label(label)
     (@labels ||= []) << label
 
     define_method :"hop_#{label}" do
+      incr_destroying if label == :destroy
       dynamic_hop label
```
jeremyevans, line 61: This adds an unnecessary conditional to every non-destroy hop.  You could avoid it by defining different methods depending on the conditional.  Up to you.

```suggestion
    if label == :destroy
      define_method :"hop_#{label}" do
        incr_destroying 
        dynamic_hop label
      end
     else
      define_method :"hop_#{label}" do
        dynamic_hop label
      end
```

prog/minio/minio_cluster_nexus.rb
```diff
@@ -51,14 +51,6 @@ def self.assemble(project_id, cluster_name, location, admin_user,
     end
   end
 
-  def before_run
-    when_destroy_set? do
-      unless ["destroy", "wait_pools_destroyed"].include?(strand.label)
```
jeremyevans, line 56: Just looking at the diff, it looks suspicious, but `wait_pools_destroyed` is only hopped to after `decr_destroy`, so this should work with the base `before_run`.  Similar for the diff in the minio_pool_nexus.

prog/minio/minio_server_nexus.rb
```diff
@@ -41,7 +41,7 @@ def self.assemble(minio_pool_id, index)
 
   def before_run
     when_destroy_set? do
-      if strand.label != "destroy"
+      if !destroying_set?
```
jeremyevans, line 44: Up to you, but this PR mixes `if !destroying_set?` and `unless destroying_set?`.

prog/vm/vm_pool.rb
```diff
@@ -22,14 +22,6 @@ def self.assemble(size:, vm_size:, boot_image:, location:, storage_size_gib:,
     end
   end
 
-  def before_run
-    when_destroy_set? do
-      unless ["destroy", "wait_vms_destroy"].include?(strand.label)
```
jeremyevans, line 27: Is this change safe?  `destroy` does not call `decr_destroy` in this prog, unlike other progs.

jeremyevans (commented):
I like how much code this removes.

prog/minio/minio_server_nexus.rb
enescakir: While refactoring, I used `unless destroying_set?` for everything, but since this condition tree has `elsif`, we can't mix `unless` and `elsif`. So, I changed them to `if !*_set?`. Since `hop_*` raises an exception, there's no need for `elsif`; we can just use `if` if you prefer.

prog/vm/vm_pool.rb
enescakir: I think it's safe. It's probably that `decr_destroy` was forgotten

prog/minio/minio_cluster_nexus.rb
enescakir: Yes, `wait_pools_destroyed` should only be hopped from `destroy`. Since the base `before_run` checks the `destroying` semaphore instead of `strand.label`, and `destroying` is incremented at `wait_pools_destroyed`, I think it's fine.

prog/base.rb
enescakir: I like it. It adds a few lines, but it's better to optimize runtime and make the decision during loading.

prog/ai/inference_endpoint_nexus.rb
```diff
@@ -75,7 +75,7 @@ def self.assemble(project_id:, location:, boot_image:, name:, vm_size:, storage_
 
   def before_run
     when_destroy_set? do
-      if strand.label != "destroy"
+      unless ["destroy", "self_destroy"].include?(strand.label)
```
bsatzger: good catch!

prog/base.rb
```diff
@@ -25,6 +25,7 @@ def #{name}
 
   def self.semaphore(*names)
     names.map!(&:intern)
+    names << :destroying if names.include?(:destroy)
```
bsatzger, line 28: this is great. another option without introducing another semaphore would be to mandate that all labels that come after destroy are prefixed with `destroy` (and no label before destroy is prefixed with `destroy`).
```
when_destroy_set? do
     hop_destroy unless @strand.label.start_with? "destroy"
end
```

bsatzger (approved):

fdr:
Huh this patch is from february and the feedback was positive, was there a reason it didn't make it? @jeremyevans @jeremyevans @furkansahin  [maybe the reason was me, but I don't think I objected, and I don't customarily gate _every_ patch]

jeremyevans:
I reviewed and approved, as did @bsatzger .  Looking at some comments above, @enescakir was asking for your feedback.

furkansahin:
we gotta get this in some form @enescakir .

PR #2725, opened 2025-02-03
Do not trigger a deadline page if the host is unavailable

enescakir:
When the host crashes, all resources on it trigger a deadline page, which creates a page storm. This is overwhelming for the on-call engineer and serves no purpose. If the VM host is unavailable, there’s no need to trigger a deadline page.

Some resources don't have separate unavailability page type and instead use deadlines to create them. Because of this, even if the VM host is unavailable, we create a deadlines page if the current label is "unavailable".

Our VM host monitoring isn't very sufficient right now. Adding disk monitoring and possibly SPDK service monitoring will make this PR more useful.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -108,7 +108,9 @@ SQL
             {}
           end
           extra_data.compact!
-          Prog::PageNexus.assemble("#{ubid} has an expired deadline! #{effective_prog}.#{label} did not reach #{frame["deadline_target"]} on time", ["Deadline", id, effective_prog, frame["deadline_target"]], ubid, extra_data:)
+          if Strand[extra_data[:vm_host]]&.label != "unavailable" || label == "unavailable"
+            Prog::PageNexus.assemble("#{ubid} has an expired deadline! #{effective_prog}.#{label} did not reach #{frame["deadline_target"]} on time", ["Deadline", id, effective_prog, frame["deadline_target"]], ubid, extra_data:)
+          end
           modified!(:stack)
         end
       end
--- a/prog/test.rb
+++ b/prog/test.rb
@@ -121,6 +121,11 @@ class Prog::Test < Prog::Base
   label def push_subject_id
     push Prog::Test, {"subject_id" => "70b633b7-1d24-4526-a47f-d2580597d53f"}
   end
+
+  label def unavailable
+    register_deadline("unavailable", -1)
+    nap(123)
+  end
 end
 
 class Prog::Test2 < Prog::Test
--- a/spec/prog/base_spec.rb
+++ b/spec/prog/base_spec.rb
@@ -296,6 +296,30 @@ RSpec.describe Prog::Base do
       )
     end
 
+    it "does not trigger deadline while the vm host is unavailable" do
+      vm = create_vm(vm_host: create_vm_host(data_center: "FSN1-DC1"))
+      st = Strand.create(prog: "Test", label: :napper, stack: [{"deadline_at" => Time.now - 1, "deadline_target" => "start"}]) { _1.id = vm.id }
+      vmh_st = Strand.create(prog: "Test", label: "unavailable") { _1.id = vm.vm_host_id }
+      expect {
+        st.unsynchronized_run
+      }.not_to change { Page.active.count }.from(0)
+
+      vmh_st.update(label: "available")
+      expect {
+        st.unsynchronized_run
+      }.to change { Page.active.count }.from(0).to(1)
+    end
+
+    it "triggers deadline while the vm host is unavailable if the current label is unavailable" do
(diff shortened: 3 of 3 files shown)


fdr (commented):
is it sensible to limit this to high-churn github actions only? these are the ones that cause storms, in some cases, like Postgres, they do have timely workarounds that could be obscured by a monolithic "host down" condition.

Secondly, we do tend to open a bunch of pages and then immediately close them again (e.g. if respirate stops processing), if we're doing something about storms, we could probably somehow be a bit smarter and absorb those rather than hitting the pager duty api with hundreds of requests.

enescakir:
> is it sensible to limit this to high-churn github actions only? these are the ones that cause storms, in some cases, like Postgres, they do have timely workarounds that could be obscured by a monolithic "host down" condition.

The current check is for deadline pages. If PostgreSQL/VM doesn't perform an operation during the VM host's unavailability, it won't trigger a deadline page anyway. If the VM host is down, we'll receive an unavailability page for PostgreSQL/VM, so it's not obscured. Since GitHub Actions are high-churn, they perform operations frequently, so they're more likely to trigger a deadline page.

> Secondly, we do tend to open a bunch of pages and then immediately close them again (e.g. if respirate stops processing), if we're doing something about storms, we could probably somehow be a bit smarter and absorb those rather than hitting the pager duty api with hundreds of requests.

Yes, we can also improve the page storms caused by respirate crashes. This PR aims to make host crashes less overwhelming. Should we address all cases at once? Or improve it over time

jeremyevans (commented):
I think this is a good change.  The only time I could see this causing an issue is if we were not generating a page for the VmHost when it became unavailable, because then we might miss actual issues.  Maybe we could also check whether there is an unresolved Page for the host?

fdr:
> > is it sensible to limit this to high-churn github actions only? these are the ones that cause storms, in some cases, like Postgres, they do have timely workarounds that could be obscured by a monolithic "host down" condition.
> 
> The current check is for deadline pages. If PostgreSQL/VM doesn't perform an operation during the VM host's unavailability, it won't trigger a deadline page anyway. If the VM host is down, we'll receive an unavailability page for PostgreSQL/VM, so it's not obscured. Since GitHub Actions are high-churn, they perform operations frequently, so they're more likely to trigger a deadline page.

For the case of Postgres, monitor failure yields an incr_checkup which yields unavailable which is:

```ruby
  label def unavailable
    register_deadline("wait", 10 * 60)

    nap 0 if postgres_server.trigger_failover

    reap
    nap 5 unless strand.children.select { _1.prog == "Postgres::PostgresServerNexus" && _1.label == "restart" }.empty?

    if available?
      decr_checkup
      hop_wait
    end

    bud self.class, frame, :restart
    nap 5
  end
```

The mechanism by which it itemizes itself for treatment is a deadline page.

I think it would be rather easy to miss a Postgres if we thought we had evacuated a host and were comfortable leaving it offline, but were mistaken. Given Postgres (or really, every other monitored VM type, such as MinIO) is so much rarer, I'd prefer a separate itemization, myself, if it does not come with too brutal a cost to the code. Otherwise, we have to substitute that with careful vigilance on any mysterious hosts we want to leave unavailable for a while, to make sure no extra-interesting resources are on them.

byucesoy (commented):
We had cases where we forgot to resolve unavailable VmHost pages multiple times. When we reimage a VmHost, we first make it unavailable, then manually change the state, which prevents control plane to auto resolve unavailable page. If operator forgets to resolve the page manually, then the page stays there. With the introduction of this PR, we would silence all future pages from that host.

I think the general idea on the PR is fine, but we should probably change the logic in HostNexus to resolve all unavailable pages in its wait label (if they exists).

enescakir:
> For the case of Postgres, monitor failure yields an incr_checkup which yields unavailable which is
@fdr I didn't realize that some resources, like Postgres, indicate unavailability through deadline pages. VM and VM host have separate pages, like `VmUnavailable`, so I assumed all resources had their own. I added another condition: `label == "unavailable"`. Even if the VM host is unavailable, we still create a deadlines page if the current label is "unavailable."

> We had cases where we forgot to resolve unavailable VmHost pages multiple times ...... With the introduction of this PR, we would silence all future pages from that host.

@byucesoy This PR checks the label of the VM host's strand not unavailability page of the VM host for the same reason. Even if the operator forgets to resolve the page, I hope they don't leave the VM host's strand label as `unavailable`. If so, we might have bigger problems.

fdr:
At least one problem we have: we don't log into pagerduty to check for non-alerting itemizations to solve, like we did with the web page we had in System 2. We have a somewhat crude coupling between alerting (event based) and itemization (poll based) where the general idea is we eventually drive the list to zero. The fact it showed on our diagnostic web page rather than pagerduty.com, which is rather out of the way, I think is not incidental to this.

PR #2726, opened 2025-02-03
Adding Burstable VM family

macieksarnowicz:
Adding a burstable VM family with the ability to place multiple VMs in one slice. The two VM sizes we are starting with are:
    - burstable-1 - 1 vcpu, 2GB of memory; internally, it is limited to 50% of 1 CPU with an ability to burst up to full 100%.
    - burstable-2 - 2 vcpus, 4GB of memory; internally it is limited to 100% of CPU with an ability to burst up to 200%.
    
Both sizes of Burstable VMs are placed in a VmHostSlice allocated for one core or 2 CPUs on x64, and 2 cores or 2 CPUs on amd64. Multiple VM instances can be placed inside that slice, up to the 200% CPU capacity. It could be 2 burstable-2s or 4 burstable-1s, or a combination. VMs from multiple customers can share a slice. The VMs running in the same slice are competing for the space to burst into, while having a guaranteed resources for their regular CPU limit. The memory is not shared.
    
The allocator is updated to select hosts that may be fully used but have some space in existing slices. The VmHostSliceAllocator decides if an existing slice can be reused or a new one needs to be created for each host. The hosts where an existing slice can be used are preferred.

We introduce a concept of VmFamilies in Options.rb, which will be expanded upon in subsequent commits to help with UI changes. The VmFamilies also include a concept of slice overcommit, which permits us to create VM families that place more VM instances into a slice than what a normal CPU limit would allow.


pykello (commented):
I think the changes looked good overall. I'll do some extra testing while waiting for feedback from others. Thanks.

scheduling/allocator.rb
```diff
@@ -31,7 +31,9 @@ def self.allocate(vm, storage_volumes, distinct_storage_devices: false, gpu_coun
       location_filter,
       location_preference,
       vm.family,
+      vm.cpu_percent_limit,
       vm.project.get_ff_use_slices_for_allocation || false,
+      Option::VmFamilies.find { _1.name == vm.family }.require_shared_slice || false,
```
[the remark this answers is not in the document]
macieksarnowicz: Changing this to `Option::VmFamilies.find { _1.name == vm.family }&.require_shared_slice || false` to make it a safe operation. I'd like to leave the `|| false` to make it consistent with the lines above and below.

scheduling/allocator.rb
```diff
@@ -81,7 +86,8 @@ def memory_gib_for_vcpus(cpus)
     end
 
     def cores_for_vcpus(threads_per_core)
-      [1, vcpus / threads_per_core].max
+      adjusted_vcpus = (vcpus.odd? && threads_per_core == 1) ? vcpus + 1 : vcpus
+      (Float(adjusted_vcpus) / threads_per_core).ceil
```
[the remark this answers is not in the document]
macieksarnowicz: Sure, I am adding a comment explaining this.

Gemfile
```diff
@@ -44,6 +44,7 @@ group :development do
   gem "pry-byebug"
   gem "rackup"
   gem "cuprite"
+  gem "text-table"
```
jeremyevans, line 47: Sequel ships an extension that appears to do the same thing.  `Sequel.extension :pretty_table`: https://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/PrettyTable.html

If that will work for you, it's better than using a gem that hasn't seen an update in over a decade.

lib/option.rb
```diff
@@ -52,6 +52,13 @@ def self.kubernetes_locations
     ["almalinux-9", "AlmaLinux 9"]
   ].map { |args| BootImage.new(*args) }.freeze
 
+  VmFamily = Struct.new(:name, :require_shared_slice)
+  VmFamilies = [
+    ["standard", false],
+    ["standard-gpu", false],
+    ["burstable", true]
+  ].map { |args| VmFamily.new(*args) }.freeze
```
jeremyevans, line 60:
```suggestion
  ].map { |args| VmFamily.new(*args).freeze }.freeze
```

If you are freezing the array, you probably want to freeze the elements of the array as well.  This is one way to do that, but you could also make `VmFamily` freeze automatically in `initialize`, or use `Data.define` instead of `Struct.new`.

Probably should do similar for the `VmSize`, but since that already exists, might make sense to do that as a separate commit.

scheduling/allocator.rb
```diff
@@ -163,6 +181,38 @@ def self.candidate_hosts(request)
...
+        # new one is created
+        ds.with(:slice_utilization, DB[:vm_host_slice]
+          .select_group(:vm_host_id)
+          .select_append { (sum(Sequel[:total_cpu_percent]) - sum(Sequel[:used_cpu_percent])).as(slice_cpu_available) }
+          .select_append { (sum(Sequel[:total_memory_gib]) - sum(Sequel[:used_memory_gib])).as(slice_memory_available) }
+          .where(enabled: true)
+          .where(is_shared: true)
+          .where(Sequel[:used_cpu_percent] + request.cpu_percent_limit <= Sequel[:total_cpu_percent])
+          .where(Sequel[:used_memory_gib] + request.memory_gib <= Sequel[:total_memory_gib]))
+          # end of 'with'
```
jeremyevans, line 199:
```suggestion
            .select_group(:vm_host_id)
            .select_append { (sum(Sequel[:total_cpu_percent]) - sum(Sequel[:used_cpu_percent])).as(slice_cpu_available) }
            .select_append { (sum(Sequel[:total_memory_gib]) - sum(Sequel[:used_memory_gib])).as(slice_memory_available) }
            .where(enabled: true)
            .where(is_shared: true)
            .where(Sequel[:used_cpu_percent] + request.cpu_percent_limit <= Sequel[:total_cpu_percent])
            .where(Sequel[:used_memory_gib] + request.memory_gib <= Sequel[:total_memory_gib])
          )
```

Seems better to use indentation to signal the end of `with`.

scheduling/allocator.rb
```diff
@@ -182,7 +232,14 @@ def self.candidate_hosts(request)
       ds = ds.where(allocation_state: request.allocation_state_filter) unless request.allocation_state_filter.empty?
       ds = ds.exclude(total_cores: 14, total_cpus: 14) unless request.family == "standard-gpu"
 
-      ds = ds.where(accepts_slices: false) if !request.use_slices
+      # If we dont's want to use slices, place those only on hosts that do not accept them
+      # If we require a shared slice (for burstable vm), allocate those only on hosts that accept slices
+      # In all other cases, the host's acceptance of slices will determine if the VM is created in a slice or not
+      if !request.use_slices
+        ds = ds.where(accepts_slices: false)
```
jeremyevans, line 239: If the request doesn't use slices, can it still be placed on a host that accepts slices?  If it can, this seems wrong.  If hosts that accept slices require that VMs on them use slices, it seems like `accept_slices` should be renamed to `require_slices`.

scheduling/allocator.rb
```diff
@@ -315,56 +375,106 @@ def update(vm, vm_host)
...
+  # All this logic is split between 'select_existing_slice' and 'update' methods. The first
+  # one checks if the candidate_host can host a slice or has one already.
+  # The second one creates a slice if needed, once the host candidate is selected.
   class VmHostSliceAllocation
     def initialize(candidate_host, request, vm_host_allocations)
       @candidate_host = candidate_host
       @request = request
       @vm_host_allocations = vm_host_allocations
+
+      @existing_slice = if @request.require_shared_slice then select_existing_slice end
```
jeremyevans, line 395:
```suggestion
      @existing_slice = select_existing_slice if @request.require_shared_slice
```

This is the preferred style for a single line `if`.  It's not technically the same, because it doesn't set `@existing_slice` in the `else` case, but that difference shouldn't affect this code.

scheduling/allocator.rb
```diff
@@ -315,56 +375,106 @@ def update(vm, vm_host)
...
-
-        st = Prog::Vm::VmHostSliceNexus.assemble_with_host(
-          "#{vm.family}_#{vm.inhost_name}",
-          vm_host,
-          family: vm.family,
-          allowed_cpus: cpus,
-          memory_gib: @request.memory_gib_for_vcpus(cpus.count),
-          is_shared: false
-        )
+      if @existing_slice.nil?
```
jeremyevans, line 419: In general, prefer `if @existing_slice` and switching the order of the `if`/`else` blocks.  Direct `nil?` calls in conditionals should be avoided unless the value can be `false` and you want to treat the `false` case differently.

scheduling/allocator.rb
```diff
@@ -315,56 +375,106 @@ def update(vm, vm_host)
...
+        DB.transaction do
```
jeremyevans, line 420: Probably best to move this around the preceding `if`, so you only need one call and not a separate call in both the `if` and `else` branches.

jeremyevans (commented):
Sorry for the delay in reviewing, been busy with cli stuff.

Gemfile
macieksarnowicz: Thanks for the suggestion. I swapped it for the pretty_table and the code is much simpler with that.

lib/option.rb
macieksarnowicz: Changed to Data.define. I agree all other structs in options.rb should be updated accordingly. I can send it later as a separate commit.

scheduling/allocator.rb
macieksarnowicz: I indented the lines, but the editor (Rubycop?) forces the closing parenthesis to be on the last where. I cannot make it go to the new line. So I left the comment in there as well, but hopefully it looks a bit better.

scheduling/allocator.rb
macieksarnowicz: The logic is as follows:
- if project does not use slices (`request.use_slices == false`) then we should only use hosts that do not accept slices
- if project wants to use slices (`request.use_slices == true`) AND the VM family requires slices (`request.require_shared_slice == true`) then we force the placement only on hosts that accept slices
- in other cases, meaning project uses slices but does not require them, we use all hosts and the VM will be created in a slice or not in the slice depending on the host placement. 

So to answer your question directly - No, if the request doesn't use slices it can only be placed on hosts where `accept_slices: false`.

prog/test/vm_group.rb
```diff
@@ -27,6 +27,7 @@ def self.assemble(boot_images:, storage_encrypted: true, test_reboot: true, test
     test_slices = frame.fetch("test_slices")
     project.set_ff_use_slices_for_allocation(test_slices)
 
+    size_options = test_slices ? ["standard-2", "burstable-1"] : ["standard-2"]
```
byucesoy, line 30: Should we include `size_options.size` to the vm_count calculation below?

config/billing_rates.yml
```diff
@@ -28,13 +28,27 @@
...
+- { id: 4809dd5b-1cdd-4e01-b587-4847a5565592, resource_type: VmVCpu,                  resource_family: burstable,           location: hetzner-fsn1,   unit_price: 0.0001525298, billed_by: duration, active_from: 2024-11-24T00:00:00Z }
+- { id: d43123cb-ae0f-48c6-acf9-2a7d240fa44c, resource_type: VmVCpu,                  resource_family: burstable,           location: hetzner-hel1,   unit_price: 0.0001450893, billed_by: duration, active_from: 2024-11-24T00:00:00Z }
+- { id: 082d275c-29c4-4f5e-b93d-81d2a0a38e02, resource_type: VmVCpu,                  resource_family: burstable,           location: github-runners, unit_price: 0.0001314330, billed_by: duration, active_from: 2024-11-24T00:00:00Z }
+- { id: 5561be8f-d21a-4caf-915e-4fa4584d07f0, resource_type: VmVCpu,                  resource_family: burstable,           location: leaseweb-wdc02, unit_price: 0.0001922123, billed_by: duration, active_from: 2024-11-24T00:00:00Z }
+- { id: edb31728-179a-4a68-b353-fb4d788380d1, resource_type: VmVCpu,                  resource_family: burstable,           location: hetzner-ai,     unit_price: 0.0001510000, billed_by: duration, active_from: 2024-11-24T00:00:00Z }
```
byucesoy, line 35: We shouldn't add billing rates for the regions we are not selling burstable family. We used to do that, but it just adds a line we need to maintain without providing any benefit, so we stopped adding billing rates for all regions. 

`hetzner-fsn1`, `hetzner-hel1` and `leaseweb-wdc02` is enough. Actually, I'm not sure if we want to enable burstable on `hetzner-hel1` even.

prog/test/vm_group.rb
macieksarnowicz: Added.

config/billing_rates.yml
macieksarnowicz: Reduced the billing rates to the three regions only.

scheduling/allocator.rb
pykello: I think the `accept_slices` -> `require_slices` rename should be addressed in a separate PR, so I'm resolving this.

pykello (approved):

pykello:
@macieksarnowicz this branch has conflicts with main branch, which prevents merging. Could you please resolve that?

macieksarnowicz:
> @macieksarnowicz this branch has conflicts with main branch, which prevents merging. Could you please resolve that?

@pykello - done.

PR #2730, opened 2025-02-04
Add query parameterization analysis

jeremyevans:
This allows you to analyze all queries issued during testing, and see which
ones were parameterized, which ones were not parameterized, and possible
queries where parameterization was missed.

To see this analysis, run "rake check_query_parameterization".  This will
run the specs logging all queries, then run a script that analyzes the
logged queries.  You can view the analysis in the created
sql_query_parameterization_analysis.txt.

The first line in this analysis has summary data:

```
Summary: Missed: 11|11, Parameterized: 1034|71461, Not Parameterized: 94|33003
```

`Missed: 11|11` means 11 total and 11 unique queries were issued that you would
expect to be parameterized but were not (missed parameterization).

`Parameterized: 1034|71461` means there were 71461 total and 1034 unique
parameterized queries.

`Not Parameterized: 94:33003` means there were 33003 total and 94 unique
queries that were not parameterized and not expected to be parameterized.

The rest of the file lists the unique queries for each of the three types.
It can be useful to review all sections to see the queries in use, in
case any appear odd and worthy of further review.

In order to get to the above results, this first adds a commit to make
the SQL generated by Sequel to not vary based on the number of elements
in an array, by switching `column IN (value_list)` to
`column = ANY(array_expr::type[])`.


fdr (approved):
People overuse the `.twice` or "exactly n times" feature, I harp on it but haven't really blocked commit for something like that.

There's only two useful modes in *almost* every circumstance. 100:1 or more: either `at_least(:once)` (ugh wordy, can we condense it with our own contraction?) or the default (which expects it exactly once). The "n times" is useful so rarely when it says something important about the algorithm.

Maybe I need to be stricter about that.

PR #2731, opened 2025-02-04
Support bdev_ubi 0.3 & increase iobuf pool capacity used for SPDK.

pykello:
### Rhizome: Support bdev_ubi 0.3 and change SPDK iobuf sizes.
Whenever an IO channel is made to a bdev, it gets some pre-allocated iobufs from SPDK's pre-allocated buffer pools. SPDK has 2 pools, one with small buffers (8kb per item), and one with large buffers (128kb per item).

This sizes of these pools are configurable at SPDK startup time.

We were getting crashes in production because of no free items in these pools.

To fix this, this change:
1. Increases the sizes of these pools.
2. Increases the number of 1g hugepages used for SPDK, to have enough  space for the new sizes of pools
3. Adds support for bdev_ubi-0.3 which will avoid avoid crash & will just error out in case were are out free pool items.

### Clover: Support bdev_ubi 0.3 & increase SPDK hugepages.
bdev_ubi 0.3 fixes several error handling issues, mainly the ones related to when we are out of iobuf free items which caused crashes in production.

When we added support for bdev_ubi 0.3 in the Clover side, we increased number of hugepages used for SPDK. So, this PR also increases that on the Clover side.

### Set used_hugepages_1g using total SPDK hugepages
SPDK allocates memory lazily as it needs, so although we mount 4G of hugepages to SPDK, `cat /proc/meminfo` shows only 3 hugepages used.

This PR changes fixes that by using the amount of hugepages reserved for SPDK on reboot.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -126,7 +126,7 @@ module Config
   override :minio_version, "minio_20240406052602.0.0_amd64"
 
   # Spdk
-  override :spdk_version, "v23.09-ubi-0.2"
+  override :spdk_version, "v23.09-ubi-0.3"
 
   # Boot Images
   override :default_boot_image_name, "ubuntu-jammy", string
--- a/prog/storage/setup_spdk.rb
+++ b/prog/storage/setup_spdk.rb
@@ -4,8 +4,8 @@ class Prog::Storage::SetupSpdk < Prog::Base
   subject_is :sshable, :vm_host
 
   SUPPORTED_SPDK_VERSIONS = [
-    ["v23.09-ubi-0.2", "x64"],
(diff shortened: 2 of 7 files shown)


.rubocop.yml
```diff
@@ -80,7 +80,7 @@ RSpec/SpecFilePathFormat:
     - 'spec/routes/**/*.rb'
 
 RSpec/MultipleMemoizedHelpers:
-  Max: 6
+  Max: 8
```
fdr, line 83: let's just eliminate this cop, it doesn't do rewrites, meh.

PR #2737, opened 2025-02-05
Improve Lb healthcheck logic to use the right ip and fix nic serializer

mohi-kalantari:
LB Healthcheck was not using the right private ipv4. now it supports VMs with bigger subnet than /32.

Also nic serializer had the same issue. now based on the provided netmask, it returns the right address.


serializers/nic.rb
```diff
@@ -5,7 +5,7 @@ def self.serialize_internal(nic, options = {})
     {
       id: nic.ubid,
       name: nic.name,
-      private_ipv4: nic.private_ipv4.network.to_s,
+      private_ipv4: (nic.private_ipv4.netmask.to_s == "/32") ? nic.private_ipv4.network.to_s : nic.private_ipv4.nth(1).to_s,
```
furkansahin, line 8: let's use `netmask.prefix_len==32`

rhizome/host/lib/vm_setup.rb
```diff
@@ -340,6 +340,8 @@ def routes4(ip4, ip4_local, nics)
   end
 
   def update_via_routes(nics)
+    return if NetAddr::IPv4Net.parse(nics.first.net4).netmask.to_s == "/32"
```
furkansahin, line 343: same here

serializers/nic.rb
mohi-kalantari: done

rhizome/host/lib/vm_setup.rb
mohi-kalantari: done

furkansahin:
Update the commit header as:
```
Fix LB healthchecks by finding the correct ip for subnet allocated K8s VMs
```

furkansahin (approved):

PR #2740, opened 2025-02-05
Remove Version and Subnet Selection From Kubernetes Cluster Creation Form

aamederen:
Remove the version selection from k8s creation form

    Since we are a new service, let's not deal with allowing an older
    version of k8s on the form. If someone asks for it, we can create
    v1.31 on the backend anyway. Otherwise, everything is v1.32.

    This is a leftover item that fell through the cracks during the
    lengthy review process.

    We expect v1.33 to be released in a short while and will bring this
    back again.

Remove Subnet Selection From k8s Creation Form

    With this patch, we remove the Subnet Selection from Kubernetes
    creation form. Instead, we create a default subnet with a larger
    ipv4 prefix (/16 instead of /26) with the name "cluster-k8s-subnet".

    This way, the flow is simplified, as we don't require a customer to create
    a subnet before creating a cluster. Also, the subnet would need to have
    a larger prefix and changing that is not easy at the moment.
    
Fix private subnet selection in a way to support larger subnets (/16)

    Previously, random_private_ipv4 function could randomly error out
    when requesting larger subnets, because one of the classes were not
    large enough to accomodate /16 and the code didn't handle this case.

    Now we retry the function if the initially selected family of large subnet
    cannot accomodate the requested cidr size.

    This is needed when we request bigger subnets for kubernetes clusters.

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -9,7 +9,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
         fail "No existing project"
       end
 
-      unless ["v1.32"].include?(version)
+      unless ["v1.32", "v1.31"].include?(version)
         fail "Invalid Kubernetes Version"
       end
 
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -3,7 +3,7 @@
 class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
   subject_is :kubernetes_cluster
 
-  def self.assemble(name:, version:, private_subnet_id:, project_id:, location:, cp_node_count: 3, target_node_size: "standard-2", target_node_storage_size_gib: nil)
+  def self.assemble(name:, project_id:, location:, version: "v1.32", private_subnet_id: nil, cp_node_count: 3, target_node_size: "standard-2", target_node_storage_size_gib: nil)
     DB.transaction do
       unless (project = Project[project_id])
         fail "No existing project"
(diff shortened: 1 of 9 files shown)


furkansahin (commented):
Let's not change the recursive logic to a loop in this patch. It should be handled in a different PR. The recursive function has been serving us just OK and we have no idea what is a safe loop time. 
Also, please update the commit message with a better explanation that `Fix some issues in random_private_ipv4`
It should be more like:
```
Fix private subnet selection in a way to support larger subnets (/16)
```

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -3,27 +3,38 @@
...
         fail "Invalid Kubernetes Version"
       end
 
       Validation.validate_kubernetes_name(name)
       Validation.validate_kubernetes_cp_node_count(cp_node_count)
 
-      # TODO: Validate subnet location if given
-      # TODO: Validate subnet size if given
-      # TODO: Create subnet if not given
+      subnet = if private_subnet_id && private_subnet_id != ""
```
enescakir, line 19: Do we really need to check for an empty string?

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -3,27 +3,38 @@
...
+        fail "Given subnet is not available in the given project" unless project.private_subnets.any? { |ps| ps.id == private_subnet_id }
+        PrivateSubnet[private_subnet_id]
```
enescakir, line 21: `project.private_subnets.find { |ps| ps.id == private_subnet_id` will return `PrivateSubnet` model.

maybe 🤔 

```suggestion
        project.private_subnets.find { _1.id == private_subnet_id } || fail "Given subnet is not available in the given project"
```

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -16,14 +16,25 @@ def self.assemble(name:, version:, private_subnet_id:, project_id:, location:, c
...
+      else
+        subnet_name = name + "-k8s-subnet"
```
enescakir, line 23: Can it exceed the length limit of a private subnet?

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: You're right, not anymore. I initially crafted the commit to allow empty string as the "default" subnet option in the UI, but later just removed the option altogether.

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: As far as I can see, `SubnetNexus.assemble` uses `validate_name`, which accepts names up to 63 characters. Kubernetes accepts names up to 40 characters.

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: neat!

prog/kubernetes/kubernetes_cluster_nexus.rb
enescakir: I didn't know there was a 40 character limit for the Kubernetes cluster name

furkansahin (approved):
I'm good with commit 311afcf1d19a6af92122031f5c33f83eed592abf

mohi-kalantari (approved):

PR #2746, opened 2025-02-05
Use a routing tree approach for cli vm ssh/sftp/scp subcommands

jeremyevans:
First three commits are support in Rodish for this style of subcommand, and final commit switches the commands to this.  So the commands now look like:

```
      vm location vm-name ssh
      vm location vm-name sftp
      vm location vm-name scp local-path :remote-path
    
      # ssh command with arguments
      vm location vm-name ssh cmd cmd-arg
    
      # With options
      vm location vm-name ssh -- -A --
      vm location vm-name ssh -- -A -- cmd cmd-arg
      vm location vm-name sftp -- -A
      vm location vm-name scp local-path :remote-path -r
```

The `--` is necessary to separate UbiCli options from ssh/sftp options, and two sets are needed ssh options, because `--` is used to separate ssh options from ssh arguments.


fdr (approved):

fdr:
That is pretty brutal I have to say, the obliged double-dashing to terminate option parsing, maybe unworkably so, but given most people apply no options most of the time, it may be moving in the right direction

jeremyevans:
> That is pretty brutal I have to say, the obliged double-dashing to terminate option parsing, maybe unworkably so, but given most people apply no options most of the time, it may be moving in the right direction

An alternative would to move the ubi cli options to the `vm` subcommand:

```
      vm -4 location vm-name ssh -A -- # still need -- for option/arg separation
      vm -6 location vm-name sftp -A
      vm -u username location vm-name scp local-path :remote-path -r
```

What I don't like about this is the options would also apply to `vm list` and other subcommands where they may not make sense.

The options could instead go after the vm reference:

```
      vm location vm-name -4 ssh -A --
      vm location vm-name -6 sftp -A
      vm location vm-name -u username scp local-path :remote-path -r
```

This avoids the options applying to `vm list` and others, but looks kind of weird.

Either of these options is possible, but would require changes to Rodish to support.

fdr:
This one makes the most sense to me:

```
      vm location vm-name -4 ssh -A --
      vm location vm-name -6 sftp -A
      vm location vm-name -u username scp local-path :remote-path -r
```

I think having `ssh`, `scp`, etc terminate option parsing (that is, `ssh` itself is an implied `--` to terminate option parsing)  makes the most sense to me, but I've hung around GNU getopt for longer than some practitioners have been alive.

jeremyevans:
> This one makes the most sense to me:
> 
> ```
>       vm location vm-name -4 ssh -A --
>       vm location vm-name -6 sftp -A
>       vm location vm-name -u username scp local-path :remote-path -r
> ```
> 
> I think having `ssh`, `scp`, etc terminate option parsing (that is, `ssh` itself is an implied `--` to terminate option parsing) makes the most sense to me, but I've hung around GNU getopt for longer than some practitioners have been alive.
> 
> It does fulfill my objective of good backspace (`C-w` for real ones) ergonomics.

I'll work on changes to Rodish and the vm cli command to support the above command format after merging this.

jeremyevans:
> I suspect the closer to a normal ssh/sftp/scp command (but with a prefix) that you can get, the easier it will be for people to pickup.

In terms of the proposed format:

```
      vm location vm-name -4 ssh -A --
      vm location vm-name -6 sftp -A
      vm location vm-name -u username scp local-path :remote-path -r
```

* The `ssh` command needs `--` if you are providing ssh options, in order to separate ssh options from ssh arguments, since the cli is not parsing the options and doesn't known which options accept/require arguments.

* The `sftp` command doesn't accept arguments, so it doesn't need `--` to separate the options.

* The `scp` command reverses the order of options and arguments. However, since the number of arguments are fixed, we could do `vm location vm-name -u username scp -r local-path :remote-path`.  I think that makes sense, so I'll switch it to that approach after merging.

> That said, I guess I'm still not entirely sure how I feel about `vm (subcommand|identity + subcommand)` as a pattern. It's probably workable, but if you do it here I would advocate trying to do it everywhere for the sake of consistency/transfer. And given that it might be good to work through a few known examples from other (probably non-vm) things to see if it breaks down in those contexts. That said, maybe it's worth the outlier status due to the advantages of ssh/sftp/scp transfer, but I'd be wary.

That makes sense.  I agree we should use the approach in other cases for consistency.  It also makes things simpler internally.

PR #2747, opened 2025-02-05
tmp

0xjunhao:


views/components/free_quota.erb
```diff
@@ -0,0 +1,24 @@
+<% warning_style = !@has_valid_payment_method %>
```
bsatzger, line 1: You can run this command to check if the formatting is compliant:
```bundle exec rake linter:erb_formatter```

File: prog/ai/inference_endpoint_replica_nexus.rb [hunk 2]
```diff
@@ -169,7 +169,12 @@ def ping_gateway
       .exists
 
     eligible_projects_ds = Project.where(api_key_ds)
-    eligible_projects_ds = eligible_projects_ds.where(id: inference_endpoint.project.id) unless inference_endpoint.is_public
+    free_quota_exhausted_projects_ds = FreeQuota.get_free_quota_exhausted_projects("inference-tokens")
+    unless inference_endpoint.is_public
```
bsatzger, line 173: today, all our inference endpoints are public. non-public inference endpoints would be where customers self-create an inference endpoint just for them. in that case they'd be billed by the resources they use and not on a per token basis.

model/billing_record.rb
```diff
@@ -46,13 +46,14 @@ def billing_rate
...
-#  amount          | numeric   | NOT NULL
-#  billing_rate_id | uuid      | NOT NULL
+#  id                | uuid      | PRIMARY KEY
+#  project_id        | uuid      | NOT NULL
+#  resource_id       | uuid      | NOT NULL
+#  resource_name     | text      | NOT NULL
+#  span              | tstzrange | NOT NULL DEFAULT tstzrange(now(), NULL::timestamp with time zone, '[)'::text)
+#  amount            | numeric   | NOT NULL
+#  billing_rate_id   | uuid      | NOT NULL
+#  free_quota_amount | numeric   | NOT NULL DEFAULT 0
```
bsatzger, line 56: do we really need to add this column? couldn't we just assume that the first 500,000 tokens for a project in a month are free tokens and deduct that for billing purposes. If a project has used >= 500,000 tokens in the current month and doesn't have billing info we exclude corresponding api keys.

views/components/free_quota.erb
0xjunhao: Seems there's no error. It says "Running ERB::Formatter..." and then just exits.

bsatzger: it might directly modify the file to make it compliant.

model/billing_record.rb
0xjunhao: I think although adding this columns seems to complicate the schema, it actually makes the logic simpler. Without this column, since the free quota applies to multiple billing rates, we will have to ensure the billing records is sorted by date and maintain an additional accumulated amount per project in order to know which amount is free quota and which is not when generating the invoices, and even with that, there is still the issue that on the day a customer exceed the free quota limit, we don't know how much usage on that day is free quota for each model, because we only have daily granularity on the billing records.

In addition, I think having this column tracks the free quota, and uses the same logic for displaying the free quota remaining and amending the billing records, and keeps the semantics of the amount field unchanged is also more robust to future billing changes.

prog/ai/inference_endpoint_replica_nexus.rb
0xjunhao: Done. Removed the logic on public.

model/billing_record.rb
bsatzger: Another idea that would help to simplify the whole logic without an additional bookkeeping column, however at the cost of less precise free inference token quotas:

Inference Endpoints: don't allow customers to further use API endpoints if their project consumed >= 500,000 inference tokens this month but doesn't have billing info. Similar to what is implemented in this PR.
Billing: statically apply a monthly credit worth of 500,000 tokens of the most expensive model, which currently amounts to $0.45. That logic would be similar to the $1 credit for GH runners we apply monthly.

This would be pretty bare bones (too bare bones? not sure) but easy to implement now and to extend in the future.

enescakir:
We have already implemented that future in https://github.com/ubicloud/ubicloud/pull/2818

PR #2751, opened 2025-02-06
Improve cli vm list command

jeremyevans:
This addresses the following issues mentioned in https://github.com/ubicloud/ubicloud/pull/2671#issuecomment-2636021590

* Use consistent widths
* Use headers by default
* Specify fields using single option

### DIFF
--- a/cli-commands/vm/list.rb
+++ b/cli-commands/vm/list.rb
@@ -2,7 +2,7 @@
 
 UbiRodish.on("vm", "list") do
   options("ubi vm list [options]", key: :vm_list) do
-    on("-h", "--headers", "show headers")
+    on("-N", "--no-headers", "do not show headers")
     on("-i", "--id", "show id")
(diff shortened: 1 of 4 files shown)


enescakir (approved):
Thanks. They look good to me.

```
➜  ubicloud git:(jeremy-cli-vm-list) ubi vm list
location      name    id                         ip4            ip6
eu-central-h1 test-vm vmnjrws8fcye7kmp1931ejxxye 136.243.203.76 2a01:4f8:252:5aa4:c696::2

➜  ubicloud git:(jeremy-cli-vm-list) ubi vm list -f name,ip4
name    ip4
test-vm 136.243.203.76
```

geemus:
It might make things a bit more readable if the columns had > 1 space between them. Probably 2 would be sufficient. I feel like I have to squint a bit presently to really find/see the breaks and think that could go a long way toward making it more readable at a glance.

PR #2752, opened 2025-02-06
Allow Rodish to accept a custom error message for invalid arguments

jeremyevans:
Previously, it could only report the number of arguments was not correct, with the expected number or range.  It couldn't display what the expected argument names were.

This adds underlying support needed to address the "We could improve the error messages to indicate which parameters are missing" request in https://github.com/ubicloud/ubicloud/pull/2671#issuecomment-2636021590 .  After it is merged, I can update the ubi cli commands to use it.  This isn't done in this pull request because it would cause additional conflicts (I think this already conflicts with other open PRs, though this conflict should be easy to resolve).

### DIFF
--- a/lib/rodish.rb
+++ b/lib/rodish.rb
@@ -71,8 +71,9 @@ module Rodish
       @command.before = block
     end
 
-    def args(args)
+    def args(args, invalid_args_message: nil)
       @command.num_args = args
+      @command.invalid_args_message = invalid_args_message
     end
 
     def autoload_subcommand_dir(base)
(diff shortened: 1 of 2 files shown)


fdr (approved):

geemus:
I've come to like offering feedback on what the valid options are and/or providing examples of their usage if possible. So you might consider something where, for instance, if location is omitted it would say something to the effect of: `key is required, it should be one of these values [x, y, z], passed in this way '--key=<value>``. I've often even done that as an interactive selection, but that's a whole other can of worms. Regardless, it's probably overkill for this at present, but I thought I'd bring it up as if you want to go that route it might be good to start thinking about how/where it would fit.

jeremyevans:
For the `ubi` cli, I think options should always be optional, anything required should be an argument. This PR should allow specifying what the correct arguments are (as opposed to the current behavior, which just tells you you are using the wrong number of arguments).

Maybe we could make this more flexible, so that if you pass 1 argument where 2 arguments are required, it could give you a different error message than if you passed 0 arguments, but I'm not sure whether that is worth the complexity.  It would be easy to support if we wanted it by making `invalid_args_message` be a proc in addition to a string.

PR #2753, opened 2025-02-06
Introduce pgbouncer

shikharbhardwaj:
## Description

Introduce pgbouncer to improve the cost of connection overhead,
improving the idle and short-lived connections at the database server.
The pgbouncer service runs as a systemd unit listening on port 6432,
relaying connections to the postgres server over a Unix domain socket.

## Limitations/TODOs

- [x] Plaintext password in userlist.txt
  * pgbouncer auth_type limitations: https://github.com/pgbouncer/pgbouncer/blob/master/doc/config.md#limitations
  * https://www.perplexity.ai/search/can-a-scram-secret-be-used-in-7ycLTyuDQMWP0McJPYkSdw

- [x] Warnings in pgbouncer logs

  ```
  Feb 05 20:56:39 vm2qhphx pgbouncer[589]: hba line 9: Ident map system2postgres is not found in ident config file
  Feb 05 20:56:39 vm2qhphx pgbouncer[589]: could not parse hba config line 9
  Feb 05 20:56:39 vm2qhphx pgbouncer[589]: hba line 36: Ident map standby2replication is not found in ident config file
  Feb 05 20:56:39 vm2qhphx pgbouncer[589]: could not parse hba config line 36
  ```

- [x] The get_auth function is in the public schema, should be moved to a dedicated pgbouncer schema

## Tests
- [x] Unit tests
- [x] Local run
- [x] E2E tests

## Rollout

- [x] Update all hosts to use the new postgres image with pgbouncer
- [x] Update default postgres image version in `download_boot_image.rb`
- [ ] Performance benchmarks
- [ ] Update documentation to reference availability of pgbouncer endpoint
- [ ] Update postgres details page to show pgbouncer connection string

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -160,6 +160,7 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
     end
 
     vm.sshable.cmd("sudo -u postgres pg_ctlcluster #{postgres_server.resource.version} main reload")
+    vm.sshable.cmd("sudo systemctl reload pgbouncer")
     hop_wait
   end
 
@@ -421,6 +422,7 @@ SQL
   label def restart
     decr_restart
     vm.sshable.cmd("sudo postgres/bin/restart #{postgres_server.resource.version}")
+    vm.sshable.cmd("sudo systemctl restart pgbouncer")
     pop "postgres server is restarted"
   end
 
--- a/rhizome/postgres/bin/configure
+++ b/rhizome/postgres/bin/configure
@@ -46,6 +46,7 @@ pg_hba_entries = <<-PG_HBA
 # TYPE  DATABASE        USER            ADDRESS                 METHOD
 # Database administrative login by Unix domain socket
 local   all             postgres                                peer map=system2postgres
+local   all             pgbouncer                               peer map=system2pgbouncer
 
 # "local" is for Unix domain socket connections only
 local   all             all                                     peer
@@ -86,6 +87,7 @@ pg_ident_entries = <<-PG_IDENT
 # Authentication" for a complete description.
 # MAPNAME             SYSTEM-USERNAME         PG-USERNAME
 system2postgres       postgres                postgres
+system2pgbouncer      postgres                pgbouncer
 system2postgres       ubi                     postgres
 standby2replication   #{identity}             ubi_replication
 PG_IDENT
@@ -93,3 +95,36 @@ safe_write_to_file("/etc/postgresql/#{v}/main/pg_ident.conf", pg_ident_entries)
 
 # Reload the postmaster to apply changes
 r "pg_ctlcluster #{v} main reload || pg_ctlcluster #{v} main restart"
+
+# PgBouncer config
+pgbouncer_config = <<-PGBOUNCER_CONFIG
+# PgBouncer configuration file
+# ============================
+[databases]
+; any db over Unix socket
+* =
+
+[pgbouncer]
+listen_port = 6432
+listen_addr = 0.0.0.0
+
+unix_socket_dir = /var/run/postgresql
+
+auth_type = hba
+auth_hba_file = /etc/postgresql/#{v}/main/pg_hba.conf
+auth_ident_file = /etc/postgresql/#{v}/main/pg_ident.conf
+auth_user = pgbouncer
+auth_query = SELECT p_user, p_password FROM pgbouncer.get_auth($1)
+
+client_tls_sslmode = require
+client_tls_protocols = tlsv1.3
+client_tls_ca_file = /etc/ssl/certs/ca.crt
+client_tls_cert_file = /etc/ssl/certs/server.crt
+client_tls_key_file = /etc/ssl/certs/server.key
+
+user = postgres
+
+max_client_conn = 5000
+max_db_connections = #{configure_hash["configs"]["max_connections"]}
+PGBOUNCER_CONFIG
+safe_write_to_file("/etc/pgbouncer/pgbouncer.ini", pgbouncer_config)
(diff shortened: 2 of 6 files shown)


prog/postgres/postgres_server_nexus.rb
```diff
@@ -279,6 +268,65 @@ def before_run
...
+      GRANT EXECUTE ON FUNCTION pgbouncer.get_auth(name) TO pgbouncer;
+    PGBOUNCER_SETUP
+
+    pgbouncer_setup_result = postgres_server.run_query(pgbouncer_setup_query).chomp
+
+    raise "Failed to set up pgbouncer" unless pgbouncer_setup_result.include?("CREATE FUNCTION")
+
+    userlist_entry = "\"pgbouncer\" \"#{pgbouncer_password}\"\n"
+    vm.sshable.cmd("sudo tee /etc/postgresql/#{postgres_server.resource.version}/main/userlist.txt > /dev/null", stdin: userlist_entry)
+    vm.sshable.cmd("sudo cp /etc/postgresql/#{postgres_server.resource.version}/pgbouncer.ini /etc/pgbouncer/pgbouncer.ini")
```
fdr, line 327: Is this a default ini file? normally I'd write the content wholesale for an application like this, but maybe my memory of the bulk in the pgbouncer.ini file is a bit hazy.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -279,6 +268,65 @@ def before_run
...
+        OUT   p_password text
+      ) RETURNS record
+        LANGUAGE sql SECURITY DEFINER SET search_path = pg_catalog AS
+      $$SELECT usename, passwd FROM pg_shadow WHERE usename = p_user$$;
+
+      REVOKE ALL ON FUNCTION pgbouncer.get_auth(name) FROM PUBLIC, pgbouncer;
+      GRANT EXECUTE ON FUNCTION pgbouncer.get_auth(name) TO pgbouncer;
+    PGBOUNCER_SETUP
+
+    pgbouncer_setup_result = postgres_server.run_query(pgbouncer_setup_query).chomp
```
fdr, line 321: is this transactional, e.g. is `-1` implied? If not, bracketing the whole text in BEGIN/COMMIT would be better if possible.

prog/postgres/postgres_server_nexus.rb
fdr: okay I see that you are templating the whole thing downpatch, but why is it bouncing around from one location to another?

fdr (commented):
It's not just adding configuration, right? it's tweaking services, it's adding a lot of the feature (the commit message does not yet say what the considerations are)

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: I generally wanted to keep all generated configs for a given postgres version in `/etc/postgresql/#{postgres_server.resource.version}`,  to track pgbouncer configs for different versions. In the current state we just run a singular process using the package-provided systemd unit, so here I just copy over the config to the path it looks at for the config.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -357,6 +421,11 @@ def before_run
       hop_configure_prometheus
     end
 
+    when_configure_pgbouncer_set? do
```
byucesoy, line 424: No one increments configure_pgbouncer, right? We actually want to set it up once at the initialization time. I think we can remove the semaphore and this check.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -279,6 +280,69 @@ def before_run
...
 
     vm.sshable.cmd(command)
+    hop_configure_pgbouncer
+  end
+
+  label def configure_pgbouncer
+    pgbouncer_password = SecureRandom.urlsafe_base64(15)
+
+    pgbouncer_setup_query = <<~PGBOUNCER_SETUP
+      BEGIN;
```
byucesoy, line 290: I think it would be better to move this to `initialize-empty-database` rhizome script

prog/postgres/postgres_server_nexus.rb
fdr: In my experience, pgbouncer more often stands alone and independent of a postgres version. I also have implemented the multi-pgbouncer approach with systemd templates, and it worked well, other than making metrics life way more complex. It's hard for me to remember exactly, but I think I worked off this one:

https://www.enterprisedb.com/blog/running-multiple-pgbouncer-instances-systemd

fdr: Probably worth more consideration: atomicity and syncing files, I don't know if `cp` does any of that, but `mv` is atomic and does. Anyway, I might in detail rework the config file handling, I think the coupling to postgres version might be spurious.

shikharbhardwaj: This makes sense, I switched it over to directly write the config to the final location.

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: I was wondering if this might come in handy for updating existing postgres servers to setup pgbouncer on them

rhizome/postgres/bin/configure
```diff
@@ -86,10 +87,44 @@ pg_ident_entries = <<-PG_IDENT
 # Authentication" for a complete description.
 # MAPNAME             SYSTEM-USERNAME         PG-USERNAME
 system2postgres       postgres                postgres
+system2postgres       postgres                pgbouncer
```
byucesoy, line 90: system2pgbouncer?

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Since the number of pg instances is small, we'll do this manually as needed

byucesoy (approved):
Could you please put rhizome changes to different (earlier) commit, so that we can deploy it first and install rhizome on existing Postgres databases. Otherwise, there is a risk of breaking existing databases if a database got restarted.

PR #2756, opened 2025-02-06
Introduce daemonizer2 for better log handling and supporting argv

mohi-kalantari:
With the new version, we will not store the systemd unit logs in file
and instead rely on journalctl. This way the logs of last tries would
not get deleted.

Locking system used in the version 1 is also deleted since the unit name
is enough for locking purposes.

Stdin is also written to /dev/shm which is a filesystem on top of memory
which is faster and files are deleted on reboot; increasing performance
and security

Also in the new version, we can specify argv arguments instead of
putting the whole command in single quotes.

Unit name is now the second argv instead of the last. it was done because
it was always a single word and we could position is somewhere static

This is an example of version 1:

vm.sshable.cmd("common/bin/daemonizer 'sleep infinity' test_daemonizer")

This is an example of version2:

vm.sshable.cmd("common/bin/daemonizer2 run test_daemonizer2 sleep infinity")

The first parameter is used as the command name. there will be no -- in
daemonizer2 API and instead "clean" or "check" would be used. Also in order
to keep the style consistent, "run" command will be used to spawn up new units.

the second parameter would be used as the name and all the other parameters
would be used in the original script.

### DIFF
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -112,6 +112,18 @@ class Sshable < Sequel::Model
     stdout_str
   end
 
+  def d_check(unit_name)
+    cmd("common/bin/daemonizer2 check #{unit_name}")
+  end
+
+  def d_clean(unit_name)
+    cmd("common/bin/daemonizer2 clean #{unit_name}")
+  end
+
+  def d_run(unit_name, *run_command, stdin: nil, log: true)
+    cmd("common/bin/daemonizer2 run #{unit_name} #{Shellwords.join(run_command)}", stdin:, log:)
+  end
+
   # A huge number of settings are needed to isolate net-ssh from the
   # host system and provide some anti-hanging assurance (keepalive,
   # timeout).
--- a/rhizome/common/bin/daemonizer2
+++ b/rhizome/common/bin/daemonizer2
@@ -1,0 +1,69 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require "fileutils"
+require "shellwords"
+require_relative "../lib/util"
+
+if ARGV.count < 2
+  fail "Wrong number of arguments. Expected at least 2, Given #{ARGV.count}"
+end
+
+STATE_MAP = {
+  "dead" => "NotStarted",
+  "running" => "InProgress",
+  "exited" => "Succeeded",
+  "failed" => "Failed"
+}
+
+def get_state(name)
+  state = r("systemctl show -p SubState --value #{name}.service").chomp
+  STATE_MAP[state] || "Unknown"
+end
+
+def clean(name, stdin_path)
+  state = get_state(name)
+  begin
+    case state
+    when "InProgress", "Unknown", "NotStarted"
+      fail "Cleanup is only allowed when the service is in a Succeeded, Failed, or NotStarted state"
+    when "Succeeded"
+      r "sudo systemctl stop #{name}.service"
+    when "Failed"
+      r "sudo systemctl reset-failed #{name}.service"
+    end
+  ensure
+    begin
+      File.unlink(stdin_path)
+    rescue Errno::ENOENT
+    end
+  end
+end
+
+command = ARGV[0]
+name = ARGV[1]
+stdin_path = "/dev/shm/#{name}.stdin"
+
+case command
+when "check"
+  print(get_state(name))
+when "clean"
+  clean(name, stdin_path)
+when "run"
+  command_parts = ARGV[2..]
+  state = get_state(name)
+  clean(name, stdin_path) if %w[Succeeded Failed].include?(state)
+
+  dir = Dir.pwd
+  stdin = $stdin.read.strip
+  if stdin.empty?
+    command = Shellwords.shelljoin(command_parts)
+  else
+    safe_write_to_file(stdin_path, stdin)
+    command = "cat #{stdin_path} | " + Shellwords.shelljoin(command_parts)
+  end
+
+  r "sudo systemd-run --working-directory #{dir} --unit #{name} --remain-after-exit /bin/bash -c #{Shellwords.escape(command)}"
+else
+  raise "Unknown command #{command}"
+end
--- a/spec/model/sshable_spec.rb
+++ b/spec/model/sshable_spec.rb
@@ -159,4 +159,30 @@ RSpec.describe Sshable do
       expect { sa.cmd("irrelevant") }.to raise_error err
     end
   end
+
+  describe "daemonizer methods" do
+    let(:unit_name) { "test_unit" }
+    let(:run_command) { "sudo host/bin/setup-vm prep test_unit" }
+    let(:stdin_data) { "secret_data" }
+
+    it "calls cmd with the correct check command" do
+      expect(sa).to receive(:cmd).with("common/bin/daemonizer2 check test_unit")
+      sa.d_check(unit_name)
+    end
+
+    it "calls cmd with the correct clean command" do
+      expect(sa).to receive(:cmd).with("common/bin/daemonizer2 clean test_unit")
+      sa.d_clean(unit_name)
+    end
+
+    it "calls cmd with the correct run command and no stdin" do
+      expect(sa).to receive(:cmd).with("common/bin/daemonizer2 run test_unit sudo\\ host/bin/setup-vm\\ prep\\ test_unit", stdin: nil, log: true)
+      sa.d_run(unit_name, run_command)
+    end
+
+    it "calls cmd with the correct run command and passes stdin" do
+      expect(sa).to receive(:cmd).with("common/bin/daemonizer2 run test_unit sudo\\ host/bin/setup-vm\\ prep\\ test_unit", stdin: stdin_data, log: true)
+      sa.d_run(unit_name, run_command, stdin: stdin_data)
+    end
+  end
 end


fdr:
Does it make sense to put the last parameter first?

mohi-kalantari:
@fdr I kept the same structure as the V1, specifying the name of the unit as the last parameter. do you have any proposal on where to put it?

mohi-kalantari:
I didn't reuse the functions between the two rhizomes because I wanted to avoid changing the old daemonizer. That can be left as a todo for later

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,72 @@
...
+    state = get_state(name)
+    case state
+    when "InProgress", "Unknown"
+      break
+    when "Succeeded", "Failed"
+      clean(name)
+    end
+
+    dir = Dir.pwd
+    command = command_parts.map { |arg| Shellwords.escape(arg) }.join(" ")
```
fdr, line 59: I think you want `shelljoin`

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,72 @@
...
+  print(get_state(name))
+elsif ARGV[0] == "--clean"
+  name = ARGV[1]
+  clean(name)
+else
+  *command_parts, name = ARGV
+
+  FileUtils.mkdir_p("var/proc")
+
+  File.open("var/proc/#{name}.lock", File::RDWR | File::CREAT) do |f|
```
fdr, line 47: I don't think we need this, it dates back to when we used the djb daemontools daemonizer, but the unit name is adequate interlock.

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,72 @@
...
+      clean(name)
+    end
+
+    dir = Dir.pwd
+    command = command_parts.map { |arg| Shellwords.escape(arg) }.join(" ")
+
+    stdin = $stdin.read.strip
+
+    if !stdin.empty?
+      safe_write_to_file("var/proc/#{name}.stdin", stdin)
```
fdr, line 64: might it be better to use tmpfs in some way? This can interact with PrivateTmp which might be a decent setting to put on the unit. I think that will also neatly handle clean-up as well. You can probably do this by cooking  systemd-run a bit more.

fdr: One danger, though, that may preclude privatetmp or anything like that: it's not good to expose the stdin in any process names, we often use stdin when there can be secrets that it might be undesirable that anything that can run `ps` can get access to. That's not conventionally a privileged operation under unix.

enescakir:
Since we produce a lot of logs to journalctl, I've heard several times that @pykello couldn't investigate SPDK issues because the journalctl logs had already been rotated. Should we change the default configuration of journalctl to increase the rotation window?

fdr:
> @fdr I kept the same structure as the V1, specifying the name of the unit as the last parameter. do you have any proposal on where to put it?

Generally I think such an argument would come first, because it's fixed in its position and length. For example, ip netns exec.

rhizome/common/bin/daemonizer2
mohi-kalantari: done thanks

rhizome/common/bin/daemonizer2
mohi-kalantari: also removed that. became neater

rhizome/common/bin/daemonizer2
mohi-kalantari: I used the tmpfs.

 but for privatetmp, it interfered with our own namespaing and it was breaking our networking commands. for example there were some errors on creating the network namespace saying the peer reference is invalid. since using the privatetmp moves the unit to another namespace, I guess it would require bigger changes across the progs.

mohi-kalantari:
@fdr makes sense. I'll move the unit name to the first parameter

mohi-kalantari:
@enescakir even if the logs are rotated, I guess they must be present is some other file in syslogs. I'll look into the default values and see If I can increase it

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,63 @@
...
+    r "sudo systemctl reset-failed #{name}.service"
+  end
+end
+
+if ARGV[0] == "--check"
+  name = ARGV[1]
+  print(get_state(name))
+elsif ARGV[0] == "--clean"
+  name = ARGV[1]
+  clean(name)
```
fdr, line 41: clean should remove the stdin right? we don't want to leak them, is that what currently happens?

fdr (approved):
with stdin cleanup added, let's ship it.

rhizome/common/bin/daemonizer2
mohi-kalantari: done

model/sshable.rb
```diff
@@ -104,6 +104,18 @@ def cmd(cmd, stdin: nil, log: true)
...
+  def daemonizer_check(unit_name)
+    cmd("common/bin/daemonizer2 check #{unit_name}")
+  end
+
+  def daemonizer_clean(unit_name)
+    cmd("common/bin/daemonizer2 clean #{unit_name}")
+  end
+
+  def daemonizer_run(unit_name, run_command, stdin: nil)
+    cmd("common/bin/daemonizer2 run #{unit_name} #{run_command}", stdin: stdin)
```
enescakir, line 116: What do you think about adding `.shellescape` to them by default instead of leaving it up to the caller? @fdr Do multiple `shellescape` calls have any side effects?

enescakir, line 116: nit:
```suggestion
    cmd("common/bin/daemonizer2 run #{unit_name} #{run_command}", stdin:)
```

mohi-kalantari: that shellescape is taken care of in the daemonizer script. I guess we do not need to add another layer here

mohi-kalantari: done thanks.

jeremyevans: Multiple `shellescape` calls would double escape the result, which is a bad idea.  There are no problems if nothing needs escaping, but if something needs escaping, it definitely results in undesirable behavior.

fdr: yes, it's important to get the right number of escaping levels, so the obligation of the caller needs to be exact. This is why we have the `q_` methods (or variable names) in some other places.

fdr: yes, it's important to get the right number of escaping levels, so the obligation of the caller needs to be exact. This is why we have the `q_` methods (or variable names) in some other places.

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,63 @@
...
+  dir = Dir.pwd
+  command = Shellwords.shelljoin(command_parts)
+
+  stdin = $stdin.read.strip
+  unless stdin.empty?
+    safe_write_to_file(stdin_path, stdin)
+    command = "cat #{stdin_path} | #{command}"
+  end
+
+  r "sudo systemd-run --working-directory #{dir} --unit #{name} --remain-after-exit /bin/bash -c #{Shellwords.escape(command)}"
```
fdr, line 60: looks double escaped? Shelljoin + escape?

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,63 @@
...
+end
+
+def clean(name, stdin_path)
+  state = get_state(name)
+  case state
+  when "InProgress", "Unknown"
+    fail "Cannot clean unless the service is in Succeeded or Failed state!"
+  when "Succeeded"
+    r "sudo systemctl stop #{name}.service"
+    r "sudo rm -f #{stdin_path}"
```
fdr, line 31:
```suggestion
    begin
      File.unlink(stdin_path)
    rescue Errno::ENOENT
    end
```

fdr: I think stdin is written as owned by `rhizome` (though, it seems like we routinely do the wrong thing and install as `ubi` on managed VMs?) so privilege escalation and forking a process unnecessary

model/sshable.rb
```diff
@@ -104,6 +104,18 @@ def cmd(cmd, stdin: nil, log: true)
...
+  end
```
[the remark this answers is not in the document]
mohi-kalantari: changed to d_

thanks for the suggestion

rhizome/common/bin/daemonizer2
mohi-kalantari: fixed it thanks.

rhizome/common/bin/daemonizer2
mohi-kalantari: it actually isn't. without that final escape it breaks. I'll provide more details here.

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,66 @@
...
+
+def get_state(name)
+  state = r("systemctl show -p SubState --value #{name}.service").chomp
+  STATE_MAP[state] || "Unknown"
+end
+
+def clean(name, stdin_path)
+  state = get_state(name)
+  case state
+  when "InProgress", "Unknown"
```
fdr, line 27:
```suggestion
  when "InProgress", "Unknown", "NotStarted"
```

Seems like 713d4ba591b1092d02da0b43ef3098a09b20fe87 applies here, too, right?

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,66 @@
...
+    fail "Cannot clean unless the service is in Succeeded or Failed state!"
```
fdr, line 28: This message would need amendment as well given https://github.com/ubicloud/ubicloud/pull/2756/files#r1989969330

rhizome/common/bin/daemonizer2
```diff
@@ -0,0 +1,66 @@
...
+  when "Succeeded"
+    r "sudo systemctl stop #{name}.service"
+    begin
+      File.unlink(stdin_path)
+    rescue Errno::ENOENT
+    end
```
fdr, line 34: Is this the right place to do the cleanup, thinking about non-successful units, the use of reset-failed, etc? Do we want to retain stdin *forever* for a failed one? May be simpler to always run it either before (first thing) or after (with `ensure`).

rhizome/common/bin/daemonizer2
mohi-kalantari: it would look something like this at the end:

```
ExecStart="/bin/bash" "-c" "cat /dev/shm/prep_vmt5g2st.stdin | sudo host/bin/setup-vm prep prep_vmt5g2st"
```

rhizome/common/bin/daemonizer2
mohi-kalantari: changed accordingly. thanks.

model/sshable.rb
```diff
@@ -112,6 +112,18 @@ def cmd(cmd, stdin: nil, log: true)
...
+  def d_check(unit_name)
+    cmd("common/bin/daemonizer2 check #{unit_name}")
+  end
+
+  def d_clean(unit_name)
+    cmd("common/bin/daemonizer2 clean #{unit_name}")
+  end
+
+  def d_run(unit_name, *run_command, stdin: nil)
+    cmd("common/bin/daemonizer2 run #{unit_name} #{run_command.join(" ")}", stdin:)
```
fdr, line 124: this `join` is kind of suspicious, what if run_command has complicated characters in its elements?

mohi-kalantari: I replaced it with a shellwords.join.

There was a shellwords.join in daemonizer2 which I thought would take care of it but I guess no it wouldn't

PR #2760, opened 2025-02-06
Attempt to recreate VM up to 2 times if it fails during prep, take 2

jeremyevans:
Currently split into two commits, so the fixes can be easily seen.

### DIFF
--- a/spec/prog/vm/nexus_spec.rb
+++ b/spec/prog/vm/nexus_spec.rb
@@ -258,14 +258,40 @@ RSpec.describe Prog::Vm::Nexus do
     end
   end
 
+  describe "#prep_failed" do
+    it "naps" do
+      expect { nx.prep_failed }.to nap(300)
+    end
+  end
+
   describe "#prep" do
-    it "hops to run if prep command is succeeded" do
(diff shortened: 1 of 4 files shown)


spec/prog/vm/nexus_spec.rb
```diff
@@ -976,7 +976,7 @@ def requested_disk_size(st)
         expect(strand).to eq st
         expect(storage_volumes).to eq [{size_gib: 1}]
       end.and_return(true)
-      expect(nx).to receive(:frame).and_return("attempt" => 1, "storage_volumes" => [{"size_gib" => 1}]).at_least(:once)
+      expect(nx).to receive(:frame).and_return("attempt" => 1, "assemble_storage_volumes" => [{"size_gib" => 1}]).at_least(:once)
```
fdr, line 979: I suppose our mistake here can be seen as a problem of using these unschema'd data structures to too great an extent. Normally, expect's verified double behavior checks for method existence, that would have spared us from a bogus mock while also sparing us from the speed (and coupling that can blow up combinatorics) of integration testing.

When dealing with straight up (parsed) json, we're without any support.

fdr (approved):

PR #2765, opened 2025-02-07
Add array casting in additional cases

jeremyevans:
This addresses the issues caused by the merge of #2730 earlier today, and subsequent partial revert in #2763.

Reenable the automatic conversion of string value lists to text arrays only in development and when running frozen tests, and do it for all non-empty arrays, instead of the default of only arrays of 2 or more elements.

I do not feel comfortable enough to turn this on in production, since it will not catch places in the tests where we are only using empty arrays, but where the arrays will be non-empty in production.  I'm not sure we have that case, but since we don't have the tooling to ensure that, enabling this in production seems too risky.  Doing it only in the frozen tests is sufficient for now, since it will still result in the query parameterization log having consistent SQL. I'm also enabling it in development, because after enough time, if there are any issues that are not caught by the tests, they should occur in a developer's development environment, and once reported, they can be easily fixed.

Update the Sequel checkout to pull in support for automatically using correct array type when eager loading associations using join tables, and for supporting automatic conversion for arrays with 1 element.

### DIFF
--- a/db.rb
+++ b/db.rb
@@ -9,7 +9,8 @@ db_ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", "db_ca_bundle.cr
 Util.safe_write_to_file(db_ca_bundle_filename, Config.clover_database_root_certs)
 max_connections = Config.db_pool - 1
 max_connections = 1 if ENV["SHARED_CONNECTION"] == "1"
-DB = Sequel.connect(Config.clover_database_url, max_connections:, pool_timeout: Config.database_timeout).tap do |db|
+pg_auto_parameterize_min_array_size = 1 if Config.test? && ENV["CLOVER_FREEZE"] == "1"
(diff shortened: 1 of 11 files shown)


fdr (approved):

jeremyevans:
@fdr I rebased this to fix the conflicts.  Could you please rereview?

jeremyevans:
@fdr I've rebased this to fix conflicts, again.  Can you please review when you have time?

fdr (approved):
interesting, the two phases, is that for phased deployment or for exposition only?

jeremyevans:
> interesting, the two phases, is that for phased deployment or for exposition only?

No, I just used a separate commit so you could easily see the differences compared to your previous approval.  I'll squash the two commits into a single commit when merging, but I'll need to fix the conflicts first.

PR #2768, opened 2025-02-07
Make sure allocated memory matches vm memory

bsatzger:
In https://github.com/ubicloud/ubicloud/pull/2666, the memory allocated for a vm is dynamically calculated and in some cases may differ from the vm's `memory_gib` property. In vm destroy, we return memory back to the host by consulting `vm.memory_gib`. That means if our calculation differs from `vm.memory_gib` we may leak memory. One instance where that calculation differs (2x higher than `vm.memory_gib`) is allocating a standard family vm on a GEX44 host.

This commit makes sure we allocate memory according `vm.memory_gib` for non-burstables.


bsatzger:
Closing this in favor of https://github.com/ubicloud/ubicloud/pull/2770

PR #2769, opened 2025-02-07
Make neg-ttl dynamic for k8s DnsZone and use k8s DnsZone for creating k8s lbs

mohi-kalantari:
Default neg-ttl is 3600 seconds which causes problems for bootstrapping kubernetes. There is 4 minute timeout for kubeadm init and if the first dns query fails, the bootstrap will fail.

By creating a separate DnsZone for kuberenetes and changing its neg-ttl to 15 seconds, we will only change this field for entities querying the api server domain. so the change is limited.

While creating the k8s loadbalancer, we will pass the k8s DnsZone.id to the loadbalancer assemble.


spec/prog/dns_zone/setup_dns_server_vm_spec.rb
```diff
@@ -165,12 +165,19 @@
...
+      CONF
+
       expect(prog.sshable).to receive(:cmd).with("sudo -u knot tee /var/lib/knot/zone1.domain.io.zone > /dev/null", stdin: f1)
       expect(prog.sshable).to receive(:cmd).with("sudo -u knot tee /var/lib/knot/zone2.domain.io.zone > /dev/null", stdin: f2)
+      expect(prog.sshable).to receive(:cmd).with("sudo -u knot tee /var/lib/knot/k8s.ubicloud.com.zone > /dev/null", stdin: f3)
 
       expect(prog.sshable).to receive(:cmd).with("sudo systemctl restart knot")
       expect(dzs[0]).to receive(:purge_obsolete_records)
       expect(dzs[1]).to receive(:purge_obsolete_records)
+      expect(dzs[2]).to receive(:purge_obsolete_records)
```
aamederen, line 180: Well, I learned that we can do something like `expect(dzs).to all(receive(...))`

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -108,7 +108,24 @@
...
+
+      expect { nx.create_load_balancer }.to hop("bootstrap_control_plane_vms")
+
+      expect(kubernetes_cluster.api_server_lb.name).to eq "k8scluster-apiserver"
+      expect(kubernetes_cluster.api_server_lb.src_port).to eq 443
+      expect(kubernetes_cluster.api_server_lb.dst_port).to eq 6443
+      expect(kubernetes_cluster.api_server_lb.health_check_endpoint).to eq "/healthz"
+      expect(kubernetes_cluster.api_server_lb.health_check_protocol).to eq "tcp"
+      expect(kubernetes_cluster.api_server_lb.stack).to eq LoadBalancer::Stack::IPV4
+      expect(kubernetes_cluster.api_server_lb.private_subnet_id).to eq subnet.id
```
aamederen, line 123: I couldn't see a check that verifies that the LB uses the correct DnsZone

mohi-kalantari: added the check. thanks.

spec/prog/dns_zone/setup_dns_server_vm_spec.rb
mohi-kalantari: changed all the patterns like this. thanks

furkansahin (approved):

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -61,6 +62,7 @@ def before_run
       dst_port: 6443,
       health_check_endpoint: "/healthz",
       health_check_protocol: "tcp",
+      custom_hostname_dns_zone_id: custom_dns_zone_id,
```
aamederen, line 65: you also have to provide `custom_hostname_prefix`. Without it, the LB doesn't use the customer hostname for the DNS record and the DNS zone sync fails.

fdr:
I'll pass this along to @byucesoy but looks okay to me, except one thing: should probably not hard code k8s.*ubicloud.com*. While it's somewhat fanciful at this time that someone external would run the entire stack, coding the domain suffix into it makes it completely impossible. It's also impossible for us to run a staging copy.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -61,6 +62,7 @@ def before_run
       dst_port: 6443,
       health_check_endpoint: "/healthz",
       health_check_protocol: "tcp",
+      custom_hostname_dns_zone_id: custom_hostname_dns_zone_id,
```
enescakir, line 65: nit:
```suggestion
      custom_hostname_dns_zone_id:,
```

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: I tried once and caused an incident, so that's what I thought. Maybe it was me doing something wrong.

fdr (changes requested):
Without any particular endorsement, the other sub-domain systems add configuration:

```
> heroku config -s | fgrep SERVICE_HOSTNAME
LOAD_BALANCER_SERVICE_HOSTNAME=lb.ubicloud.com
POSTGRES_SERVICE_HOSTNAME=postgres.ubicloud.com
```

While perhaps I would have been fine collapsing all under a "dns zone suffix" and saying that (lb,postgres,k8s) are hard-coded prefixes, following the norm seems adequate, too.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -132,8 +132,10 @@ def ds
     nap 5 if ds.dns_zones.any?(&:refresh_dns_servers_set?)
 
     ds.dns_zones.each do |dz|
+      neg_ttl = (dz.name == "k8s.ubicloud.com") ? 15 : 3600
```
byucesoy, line 135: We shouldn't put special branching for an arbitrary DNS zone into DNS Zone setup prog. If neg_ttl needs to be different from zone to zone, we can add this as a field to dns zone model.

byucesoy (commented):
Agreed with @fdr on not hard coding `k8s.ubicloud.com`. Also, we shouldn't put branching to dns zone setup code for an arbitrary dns zone.

prog/dns_zone/setup_dns_server_vm.rb
mohi-kalantari: fixed thanks.

prog/kubernetes/kubernetes_cluster_nexus.rb
mohi-kalantari: fixed thanks.

mohi-kalantari:
@fdr Now config is used for passing the k8s.ubicloud.com

fdr (approved):
I'm okay if @byucesoy is okay with it. I guess @furkansahin should also have a quick look.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -53,6 +53,7 @@ def before_run
   end
 
   label def create_load_balancer
+    custom_hostname_dns_zone_id = Config.development? ? nil : DnsZone[name: Config.kubernetes_service_hostname].id
```
byucesoy, line 56: No need for `Config.development?` check any more. You can decide based on existence of `Config.kubernetes_service_hostname`

byucesoy (approved):

prog/kubernetes/kubernetes_cluster_nexus.rb
mohi-kalantari: done thanks

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -53,6 +53,7 @@ def before_run
   end
 
   label def create_load_balancer
+    custom_hostname_dns_zone_id = Config.kubernetes_service_hostname && DnsZone[name: Config.kubernetes_service_hostname].id
```
aamederen, line 56: nit: `DnsZone[name: Config.kubernetes_service_hostname]&.id`

migrate/20250211_add_neg_ttl_to_dns_zone.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:dns_zone) do
+      add_column :neg_ttl, Integer, default: 3600, null: false
```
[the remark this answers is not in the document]
enescakir: What's the best practice there if we need to add the column with default? 🤔 Is it adding as `null: true` without default, then update all existing, then update column as `null: false, default: val`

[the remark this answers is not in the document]
aamederen: yeah, something like that. Although not null: false would enforce the table scan anyway.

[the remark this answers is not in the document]
mohi-kalantari: Yes I knew that but since the table is small, can we proceed with the current migration?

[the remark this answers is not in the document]
mohi-kalantari: thanks for noting that.

PR #2773, opened 2025-02-07
Use post options for cli vm ssh/sftp/scp commands

jeremyevans:
This changes the commands to use the following formats:
    
```
      vm location/vm-name -4 ssh cmd
      vm location/vm-name -6 sftp -A
      vm location/vm-name -u username scp -r local-path :remote-path
      vm location/vm-name ssh -A -- cmd
      vm location/vm-name ssh -A --
```

The `--` is still needed for ssh with ssh options to separate ssh options from arguments, but otherwise I think this makes the commands simpler and more in line with the standard ssh/sftp/scp behavior.

The first two commits add the necessary support to Rodish, and the third uses the support to update the commands.

### DIFF
--- a/cli-commands/vm.rb
+++ b/cli-commands/vm.rb
@@ -10,6 +10,12 @@ UbiRodish.on("vm") do
 
   args(2...)
 
+  post_options("ubi vm location-name/(vm-name|_vm-ubid) [options] subcommand [...]", key: :vm_ssh) do
+    on("-4", "--ip4", "use IPv4 address")
+    on("-6", "--ip6", "use IPv6 address")
+    on("-u", "--user user", "override username")
+  end
+
   run do |(vm_ref, *argv), opts, command|
     @location, @vm_name, extra = vm_ref.split("/", 3)
     raise Rodish::CommandFailure, "invalid vm reference, should be in location/(vm-name|_vm-ubid) format" if extra
--- a/cli-commands/vm/post/scp.rb
+++ b/cli-commands/vm/post/scp.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 UbiRodish.on("vm").run_on("scp") do
-  options("ubi vm location-name/(vm-name|_vm-ubid) scp [options] (local-path :remote-path|:remote-path local-path) [scp-options]", key: :vm_ssh, &UbiCli::SSHISH_OPTS)
+  skip_option_parsing
 
   args(2...)
 
-  run do |(path1, path2, *argv), opts|
+  run do |(*argv, path1, path2), opts|
     remote_path1 = path1[0] == ":"
     remote_path2 = path2[0] == ":"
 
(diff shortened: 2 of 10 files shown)


fdr:
Huh, why does mixing arguments and ssh options  requiring a double-dash in more cases than using `ssh` alone?

fdr:
I played around with ssh a little bit, it looks like it terminates option parsing the when a second positional argument shows up. I tested with `ssh host echo -A` vs `ssh host -A echo`

jeremyevans:
> Huh, why does mixing arguments and ssh options requiring a double-dash in more cases than using `ssh` alone?

When `ssh` parses options, it knows which options require an argument and which do not.  We don't have the information on how the client's `ssh` program will parse options, so we cannot  do that.  Consider:

```
  ssh -A x y # x is first argument
  ssh -J x y # y is first argument, x is value of -J option
```

fdr:
> > Huh, why does mixing arguments and ssh options requiring a double-dash in more cases than using `ssh` alone?
> 
> When `ssh` parses options, it knows which options require an argument and which do not. We don't have the information on how the client's `ssh` program will parse options, so we cannot do that. Consider:
> 
> ```
>   ssh -A x y # x is first argument
>   ssh -J x y # y is first argument, x is value of -J option
> ```

Maybe I haven't grappled with the internals enough yet, but why wouldn't `ssh` stop option parsing right then and there, and pass along the rest of the argv to the client `ssh` binary however they are specified?, i.e.

`ssh hostname $restofargv`

$restofargv may include options, it may include a positional argument (i.e. a command). It might not be seen as kosher to do options after a positional argument, but I think openssh makes this work just right, ...and perhaps for this exact reason.

jeremyevans:
> > > Huh, why does mixing arguments and ssh options requiring a double-dash in more cases than using `ssh` alone?
> > 
> > 
> > When `ssh` parses options, it knows which options require an argument and which do not. We don't have the information on how the client's `ssh` program will parse options, so we cannot do that. Consider:
> > ```
> >   ssh -A x y # x is first argument
> >   ssh -J x y # y is first argument, x is value of -J option
> > ```
> 
> Maybe I haven't grappled with the internals enough yet, but why wouldn't `ssh` stop option parsing right then and there, and pass along the rest of the argv to the client `ssh` binary however they are specified?, i.e.
> 
> `ssh hostname $restofargv`
> 
> $restofargv may include options, it may include a positional argument (i.e. a command). It might not be seen as kosher to do options after a positional argument, but I think openssh makes this work just right, ...and perhaps for this exact reason.

As part of the security when executing commands, any argv executed by `ubi` must contain `--`, and the argument added by the server (not in the original argv) must come after `--`.  This is to ensure that the server cannot inject new options. 

```
ssh -- hostname -A # does not work
ssh -A -- hostname # works
ssh -J -- hostname x # does not work
ssh -J x -- hostname # works
```

We can switch to your approach, but it allows the server to execute a command with a custom option instead of just a custom argument.  Whether or not that causes an issue depends on the option handling for every program that `ubi` will be able to execute.  I haven't done an analysis of all the options for all the programs, but a review of `ssh`'s options lead me to believe that it is unwise to allow the server that much power.  The server should only be able to specify where to connect (argument, not option).

jeremyevans:
Closing as changes were included in #2788.

PR #2775, opened 2025-02-07
Add cli vm create/destroy/show commands, and list -l option

jeremyevans:
With these changes, the CLI now fully covers what you can do with the API in regards to VMs.

The first three commits are from #2773, as this builds on top of the same support.


jeremyevans:
Closing as changes were included in #2788.

PR #2782, opened 2025-02-10
Make token creation for github repository idempotent

velioglu:
If concurrent cache requests come for the same repository while setting it ip, multiple tokens can be created. It causes tokens to accumulate on E2E tests. Making it idempotent by creating them only if bucket is created before.

### DIFF
--- a/model/github/github_repository.rb
+++ b/model/github/github_repository.rb
@@ -66,6 +66,9 @@ class GithubRepository < Sequel::Model
 
   def setup_blob_storage
     DB.transaction do
+      lock!
+      return if access_key && secret_key
+
       begin
         admin_client.create_bucket({
           bucket: bucket_name,
--- a/spec/model/github/github_repository_spec.rb
+++ b/spec/model/github/github_repository_spec.rb
@@ -63,14 +63,24 @@ RSpec.describe GithubRepository do
       ]
       expect(cloudflare_client).to receive(:create_token).with("gph0hh0ahdbj6ng2cc3rvdecr8-token", expected_policy).and_return(["test-key", "test-secret"])
       expect(github_repository).to receive(:update).with(access_key: "test-key", secret_key: Digest::SHA256.hexdigest("test-secret"))
+      expect(github_repository).to receive(:lock!)
       github_repository.setup_blob_storage
     end
 
-    it "succeeds if the bucket already exists" do
+    it "succeeds if the bucket already exists and access key does not exist" do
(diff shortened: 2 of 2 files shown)


fdr (approved):

model/github/github_repository.rb
```diff
@@ -84,6 +81,7 @@ def setup_blob_storage
 
       token_id, token = CloudflareClient.new(Config.github_cache_blob_storage_api_key).create_token("#{bucket_name}-token", policies)
       update(access_key: token_id, secret_key: Digest::SHA256.hexdigest(token))
+    rescue Aws::S3::Errors::BucketAlreadyOwnedByYou
```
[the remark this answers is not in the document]
enescakir: We should be aware of that before the customer is.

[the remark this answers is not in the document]
velioglu: Ok, I wanted to keep it simple but overlook that case. I've changed the logic completely. I've checked whether access token is created by first locking the repository tuple. Can you check again?

model/github/github_repository.rb
```diff
@@ -74,6 +74,10 @@ def setup_blob_storage
       rescue Aws::S3::Errors::BucketAlreadyOwnedByYou
       end
 
+      lock!
```
enescakir, line 77: According to `lock!`'s documentation, it refresh the record. No need to reload it again.
https://sequel.jeremyevans.net/rdoc/classes/Sequel/Model/InstanceMethods.html#method-method-i-lock-21

I tested it with two prys on the same model, and it looks like it has the new value after `lock!`

```ruby
# left pry
DB.transaction do
  puts "#{Time.now} Before: #{default_branch}"
  lock!
  sleep 10
  update(default_branch: "new-branch-5")
  puts "#{Time.now} After: #{default_branch}"
end



## right pry
DB.transaction do
  puts "#{Time.now} Before: #{default_branch}"
  lock!
  puts "#{Time.now} After: #{default_branch}"
end
```
CleanShot 2025-02-12 at 14 20 19@2x

model/github/github_repository.rb
```diff
@@ -74,6 +74,10 @@ def setup_blob_storage
...
+      reload
+      return if access_key && secret_key
```
enescakir, line 79: What do you think about moving this check to the beginning of the transaction so we don’t send multiple create bucket requests?

model/github/github_repository.rb
velioglu: Removed reload

enescakir (approved):
It looks good to me. Can you run the E2E tests? Then we can check for duplicate tokens.

PR #2783, opened 2025-02-10
Remove openapi entries using /id/ routes

jeremyevans:
The routing tree was changed from an id/ prefix to an _ prefix for ubids in the cases were you can provide either a name or an id, since before the introduction of committee.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -384,122 +384,6 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Firewall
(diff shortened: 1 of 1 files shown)


fdr (approved):

fdr:
These are obsolete entries now, and would permit (they would then 404 I believe, code gone right?) requests that could be rejected earlier (...but with a 404 as well, iirc)

jeremyevans:
> These are obsolete entries now, and would permit (they would then 404 I believe, code gone right?) requests that could be rejected earlier (...but with a 404 as well, iirc)

Correct.

PR #2784, opened 2025-02-10
Fixing issues in allocator when using slices

macieksarnowicz:
**Separating slice and no-slice VMs on hosts**
The allocator decided if the VM should be created with a slice or without a slice based on the host's accept_slices flag. However that logic only applied when the project had `use_slices_for_allocation` flag set to `true` When the project flag was false any host could be picked, and a VM with no slice was created. The result was a mix of VMs in-slices and VMs with no-slices on the same host, if the host had `accept_slices=true` flag. On hosts with `accept_slices=false` no slices were created.

This fix adds an extra constraint to not allow placement of no-slice VMs on hosts set with `accept_slices=true` from projects that do not want slices. It also sets a preference to pick a host that accepts slices for projects that do want that.

Tests added to cover all those cases explicitly.

**Fix for GEX44 memory computation**
Previous commit introduced a couple of problems in this area, for GEX44 hosts:
    - the amount of memory calculated for a VM/slice was based on number of cores and assumed 2:1 ratio of cpus to cores for stanard-gpu VM. It should be 1:1 ratio
    - the memory allocated was computed in the allocator but the memory freed was taken from VmSize structure, resulting in different values on GEX44 hosts.
    
This commit fixes those problems.
    - It computes the amount of memory based on vcpus, just like VmSize
    - The update to the host is based on request.memory_gib value when not creating a slice
    - The update to the host is based on the amount allocated for a slice when using slices
    
Added multiple tests to cover all those cases. Also reverted temporary fixes that Ben did to mitigate this issue.

### DIFF
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -68,16 +68,16 @@ module Scheduling::Allocator
       self.diagnostics ||= false
     end
 
-    def memory_gib_for_cores(cores)
-      memory_gib_ratio = if arch_filter == "arm64"
+    def memory_gib_for_vcpus(cpus)
+      memory_to_cpu_ratio = if arch_filter == "arm64"
         3.2
       elsif family == "standard-gpu"
-        10.68
+        5.34
       else
-        8
+        4
       end
 
-      (cores * memory_gib_ratio).to_i
+      (cpus * memory_to_cpu_ratio).to_i
     end
 
     def cores_for_vcpus(threads_per_core)
@@ -213,10 +213,9 @@ module Scheduling::Allocator
       @candidate_host = candidate_host
       @request = request
       request_cores = request.cores_for_vcpus(candidate_host[:total_cpus] / candidate_host[:total_cores])
-      request_memory = request.memory_gib_for_cores(request_cores)
 
       @vm_host_allocations = [VmHostCpuAllocation.new(:used_cores, candidate_host[:total_cores], candidate_host[:used_cores], request_cores),
-        VmHostAllocation.new(:used_hugepages_1g, candidate_host[:total_hugepages_1g], candidate_host[:used_hugepages_1g], request_memory)]
+        VmHostAllocation.new(:used_hugepages_1g, candidate_host[:total_hugepages_1g], candidate_host[:used_hugepages_1g], request.memory_gib)]
       @device_allocations = [StorageAllocation.new(candidate_host, request)]
       @device_allocations << GpuAllocation.new(candidate_host, request) if request.gpu_count > 0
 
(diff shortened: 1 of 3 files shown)


scheduling/allocator.rb
```diff
@@ -180,7 +180,6 @@ def self.candidate_hosts(request)
       ds = ds.exclude(Sequel[:vm_host][:id] => request.host_exclusion_filter) unless request.host_exclusion_filter.empty?
       ds = ds.where(location: request.location_filter) unless request.location_filter.empty?
       ds = ds.where(allocation_state: request.allocation_state_filter) unless request.allocation_state_filter.empty?
-      ds = ds.exclude(total_cores: 14, total_cpus: 14) unless request.family == "standard-gpu"
```
bsatzger, line 183: it might make sense to not put regular vms onto gex hosts. runners running on gex hosts get a very different performance characteristic. should we keep this commit?

scheduling/allocator.rb
```diff
@@ -262,6 +264,9 @@ def calculate_score
       # penalty for AX161, TODO: remove after migration to AX162
       score += 0.5 if @candidate_host[:total_cores] == 32
 
+      # penalty for hosts that do not accept slices if the request wants to use one
+      score += 1 if @request.use_slices && !@candidate_host[:accepts_slices]
```
bsatzger, line 266: would that still result in a valid placement (putting a request that wants to use slice on a host that doesn't accept slices?

scheduling/allocator.rb
```diff
@@ -343,7 +342,7 @@ def update(vm, vm_host)
           vm_host,
           family: vm.family,
           allowed_cpus: cpus,
-          memory_gib: @request.memory_gib_for_cores(@request.cores_for_vcpus(vm_host.total_cpus / vm_host.total_cores)),
+          memory_gib: @request.memory_gib_for_vcpus(cpus.count),
```
bsatzger, line 345: we now only use this calculation for slices only and use the vm's `memory_gib` value otherwise, correct?

For allocations outside of a slice, we return `vm.memory_gib` memory in Prog::Vm::Nexus:
```used_hugepages_1g: Sequel[:used_hugepages_1g] - vm.memory_gib```
That means, when allocating outside of a slice, we must allocate `vm.memory_gib` of memory, as this is what we return in the destroy path.

macieksarnowicz: yes, my attempt here is to make it more consistent. 

When no slices - we allocate number of cores computed from vcpus and request.memory_gib and we store both values in the VM so when the VM is destroyed those values are used to return resources to the host

When slices - we compute number of cores and cpus and memory from cpus for the slice and remove that from the host. the values are stored in the VmHostSlice and returned to the host when slice is deallocated. When VM is in the slice, the VM values are added / subtracted from the slice.

pykello:
@macieksarnowicz the first commit looks good to me. I'll wait until discussions with @bsatzger are resolved before approving.

scheduling/allocator.rb
bsatzger: We are currently discussing how/if we'll leverage GEX going forward. We could keep this until the discussion has concluded.

bsatzger (approved):

PR #2788, opened 2025-02-11
Add cli pg commands

jeremyevans:
This builds on top of #2775 and adds CLI support for all actions you can take in the API:

* pg list
* pg loc/name create [-f -h -s -S -v]
* pg loc/name show
* pg loc/name destroy
* pg loc/name add-firewall-rule cidr
* pg loc/name delete-firewall-rule ubid
* pg loc/name add-metric-destination user pass url
* pg loc/name delete-metric-destination ubid
* pg loc/name failover
* pg loc/name restore name time
* pg loc/name reset-superuser-password pass

In addition, it also supports commands for using PostgreSQL CLI tools, similar to how the vm ssh/sftp/scp commands are supported:

* pg loc/name [-U -d] psql
* pg loc/name [-U -d] pg_dump
* pg loc/name [-U -d] pg_dumpall

### DIFF
--- a/cli-commands/pg/post/create.rb
+++ b/cli-commands/pg/post/create.rb
@@ -2,7 +2,7 @@
 
 UbiRodish.on("pg").run_on("create") do
   options("ubi pg location/pg_name create [options]", key: :pg_create) do
-    on("-f", "--flavor=type", "flavor (standard, paradedb, latern)")
+    on("-f", "--flavor=type", "flavor (standard, paradedb, lantern)")
     on("-h", "--ha_type=type", "replication type (none, async, sync)")
     on("-s", "--size=size", "server size (standard-{2,4,8,16,30,60})")
     on("-S", "--storage_size=size", "storage size GB (64, 128, 256)")
--- a/cli-commands/pg/post/create.rb
+++ b/cli-commands/pg/post/create.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+UbiRodish.on("pg").run_on("create") do
+  options("ubi pg location/pg_name create [options]", key: :pg_create) do
+    on("-f", "--flavor=type", "flavor (standard, paradedb, latern)")
+    on("-h", "--ha_type=type", "replication type (none, async, sync)")
+    on("-s", "--size=size", "server size (standard-{2,4,8,16,30,60})")
+    on("-S", "--storage_size=size", "storage size GB (64, 128, 256)")
+    on("-v", "--version=version", "PostgreSQL version (16, 17)")
+  end
+
+  run do |opts|
+    params = opts[:pg_create]&.transform_keys(&:to_s) || {}
+    params["size"] ||= "standard-2"
+    post(project_path("location/#{@location}/postgres/#{@pg_name}"), params) do |data|
+      ["PostgreSQL database created with id: #{data["id"]}"]
+    end
+  end
+end
--- a/cli-commands/pg/post/create.rb
+++ b/cli-commands/pg/post/create.rb
@@ -12,7 +12,7 @@ UbiRodish.on("pg").run_on("create") do
   run do |opts|
     params = opts[:pg_create]&.transform_keys(&:to_s) || {}
     params["size"] ||= "standard-2"
-    post(project_path("location/#{@location}/postgres/#{@pg_name}"), params) do |data|
+    post(project_path("location/#{@location}/postgres/#{@name}"), params) do |data|
       ["PostgreSQL database created with id: #{data["id"]}"]
     end
   end
--- a/cli-commands/pg/post/show.rb
+++ b/cli-commands/pg/post/show.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+UbiRodish.on("pg").run_on("show") do
+  fields = %w[id name state location vm_size storage_size_gib version ha_type flavor connection_string primary earliest_restore_time latest_restore_time firewall_rules metric_destinations ca_certificates].freeze.each(&:freeze)
+
+  options("ubi pg location/(vm-name|_vm-ubid) show [options]", key: :pg_show) do
+    on("-f", "--fields=fields", "show specific fields (default: #{fields.join(",")})")
+  end
+
+  run do |opts|
+    get(project_path("location/#{@location}/postgres/#{@pg_name}")) do |data|
+      keys = fields
+
+      if (opts = opts[:pg_show])
+        keys = check_fields(opts[:fields], fields, "pg show -f option")
+      end
+
+      body = []
+
+      keys.each do |key|
+        case key
+        when "firewall_rules"
+          body << "firewall rules:\n"
+          data[key].each_with_index do |rule, i|
+            body << "  " << (i + 1).to_s << ": " << rule["id"] << "  " << rule["cidr"].to_s << "\n"
+          end
+        when "metric_destinations"
+          body << "metric destinations:\n"
+          data[key].each_with_index do |md, i|
+            body << "  " << (i + 1).to_s << ": " << md["id"] << "  " << md["username"].to_s << "  " << md["url"] << "\n"
+          end
+        when "ca_certificates"
+          body << "CA certificates:\n" << data[key].to_s << "\n"
+        else
+          body << key << ": " << data[key].to_s << "\n"
+        end
+      end
+
+      body
+    end
+  end
+end
--- a/cli-commands/pg/post/show.rb
+++ b/cli-commands/pg/post/show.rb
@@ -8,7 +8,7 @@ UbiRodish.on("pg").run_on("show") do
   end
 
   run do |opts|
-    get(project_path("location/#{@location}/postgres/#{@pg_name}")) do |data|
+    get(project_path("location/#{@location}/postgres/#{@name}")) do |data|
       keys = fields
 
       if (opts = opts[:pg_show])
--- a/cli-commands/vm/post/create.rb
+++ b/cli-commands/vm/post/create.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+UbiRodish.on("vm").run_on("create") do
+  options("ubi vm location/vm_name create [options] public_key", key: :vm_create) do
+    on("-6", "--ipv6-only", "do not enable IPv4")
+    on("-b", "--boot_image=image_name", "boot image (ubuntu-noble,ubuntu-jammy,debian-12,almalinux-9)")
+    on("-p", "--private_subnet_id=id", "place VM into specific private subnet")
+    on("-s", "--size=size", "server size (standard-{2,4,8,16,30,60})")
+    on("-S", "--storage_size=size", "storage size (40, 80)")
+    on("-u", "--unix_user=username", "username (default: ubi)")
+  end
+
+  args(1, invalid_args_message: "public_key is required")
+
+  run do |public_key, opts|
+    params = opts[:vm_create]&.transform_keys(&:to_s) || {}
+    unless params.delete("ipv6-only")
+      params["enable_ip4"] = "1"
+    end
+    params["public_key"] = public_key
+    post(project_path("location/#{@location}/vm/#{@vm_name}"), params) do |data|
+      ["VM created with id: #{data["id"]}"]
+    end
+  end
+end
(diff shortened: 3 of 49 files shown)


cli-commands/pg/post/create.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+UbiRodish.on("pg").run_on("create") do
+  options("ubi pg location/pg_name create [options]", key: :pg_create) do
+    on("-f", "--flavor=type", "flavor (standard, paradedb, latern)")
```
fdr, line 5:
```suggestion
    on("-f", "--flavor=type", "flavor (standard, paradedb, lantern)")
```

fdr:
I have a question: is this ending up much like a machine generated variant, where ~everything in the API is exposed via the CLI? Does the openapi schema provide enough to DRY some stuff? On the other hand, there is also very little code besides helpful texts themselves, there may not be much to DRY. Basically, if the path going on is substantially straight enumeration without much editorialization, the more there may be to gain from reflecting on the OpenAPI artifacts.

fdr:
Well if you are making a number of decisions and the code to express it is
short then there’s not much to be gained. It naively looks like an
enumeration, so that’s why I asked

openapi/openapi.yml
```diff
@@ -1116,6 +1116,9 @@ paths:
                 storage_size:
                   description: Requested storage size in GiB
                   type: integer
+                version:
+                  description: PostgreSQL version
+                  type: string
```
fdr, line 1121: Huh so no tests submitted this key? Because they'd break.

lib/ubi_cli.rb
```diff
@@ -66,6 +66,14 @@ def self.list(cmd, label, fields, fragment: cmd)
     end
   end
 
+  def self.destroy(cmd, label, fragment: cmd)
+    UbiRodish.on(cmd).run_is("destroy") do
+      delete(project_path("location/#{@location}/#{fragment}/#{@name}")) do |_, res|
+        ["#{label}, if it exists, is now scheduled for destruction"]
+      end
```
fdr, line 73: for some later time, often, for destructive operations, many CLIs will prompt or ask you to type something.

openapi/openapi.yml
```diff
@@ -860,6 +860,21 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_name}/failover':
```
fdr: I guess we lacked a test for this as well? I wonder how that works given 100%?

fdr (approved):

openapi/openapi.yml
jeremyevans: There is an existing `openapi.yml` entry for `'/project/{project_id}/location/{location}/postgres/_{postgres_database_id}/failover'`, so I'm guessing coverage was handled by that.  Basically, there are four cases:

1. web - name
2. web - ubid
3. api - name
4. api - ubid

You can get 100% branch coverage when testing only three of these four cases.  However, the cli test I wrote uses `api - name`, which was previous untested and didn't work.

This type of issue may well affect other cases.  We could avoid this issue by not having separate entries in `openapi.yml` for the name and ubid cases.

lib/ubi_cli.rb
jeremyevans: I thought about this, but the API doesn't require this, so I don't think the CLI should either.  We could potential added a `-f`/`--force` option, and have the CLI call fail without the option, but I don't like that.  It would be preferable to making the CLI interactive, though.  In general, I think we should assume that CLI users know what they are doing (or at least are OK with taking responsibility for their CLI use).

openapi/openapi.yml
jeremyevans: From the output `git grep version spec/routes/api`, looks like the cli pg create tests were the first to use the parameter.

openapi/openapi.yml
fdr: Do you know why we have two? I presume two distinct patterns for tighter validation or something? Seems like the costs may not justify so much bookkeeping, if this is the one thing causing spurious rejections of requests and doubling up schema entries, cc @geemus

spec/routes/api/cli/vm/show_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      size: standard-2
+      unix_user: ubi
+      storage_size_gib: 0
+      ip6: 128:1234::2
+      ip4_enabled: false
+      ip4: 128.0.0.1
+      private_ipv4: #{@vm.private_ipv4}
+      private_ipv6: #{@vm.private_ipv6}
+      subnet: default-eu-central-h1
+      firewall 1:
```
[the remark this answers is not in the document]
jeremyevans: Looks like VMs can theoretically be in multiple subnets (`Vm.many_to_many :private_subnets`), but the API doesn't show which firewall is associated to which subnets (it only shows the subnet for the first nic, but it includes all related firewalls).  If we actually had a VM with multiple subnets, the CLI could not tell you which firewall is in which subnet .  We would have to change the API to nest firewall information under subnets in order for the CLI to display it properly.

openapi/openapi.yml
jeremyevans: I do not know.  I'm also not sure why having the duplicate route works in this case, but broke tests when I converted all `/id/` routes to `/_` routes (I tried that approach, but since it didn't work, I submitted #2783 to remove the `/id/` routes).

spec/routes/api/cli/vm/show_spec.rb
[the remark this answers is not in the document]
fdr: @furkansahin an algebra over NICs (that is, more than one on a machine) may not be soon in our future, but, there will be some path dependent consequences when it does happen. Do you have thoughts? I guess also @geemus who was looking at the teaching/conveyance issues of this correlation of data.

It's reasonable to streamline the common case (a single NIC) but we also don't want to get cornered.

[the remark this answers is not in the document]
fdr: (regardless, I don't think it's a reason to forestall integrating the rest of the CLI work, so long as the safety is good)

openapi/openapi.yml
jeremyevans: FWIW, I recommend that we accept either name or ubid (without underscore), and have a name validation that does not allow ubid format for the model's ubid type (chance of someone using ubid format by accident is minimal).  We would need to check that nobody is currently doing that, though.

lib/ubi_cli.rb
fdr: I don't think the API-therefore-CLI syllogism holds. An API can't prompt, it's only an interface, a CLI exists to do things an API can't really do, since an API arguably defines itself in being abstract from a client implementation, it's something of a category error.

The problem is, it comes back as desperate support tickets for which varying surgeries are possible to fix the customer. Here are two kinds of errors:

* hacking up a previous command and bouncing on the enter key.
* accidentally recalling a previous dangerous command with ctrl-r, and bouncing on the enter key

jeremyevans: Well, I'm happy to make it work the way you want, you just need to describe how you want it to work.  Making `ubi` interactive (prompting the user for input) will require changes to the `ubi` program, and the main advantage of the thin-CLI approach is you need do not to make changes to the program to add features to the CLI. I'd prefer to make any necessary changes to `ubi` earlier rather than later.

fdr: @geemus , @byucesoy : do you have opinions on re-prompts in the CLI? The former, as a general principle from experience, the latter, because you might be on the hook for "oops" (and yet others we simply have to say "SOL")

IIRC heroku had a `--confirm` flag for non-interactive mode, once you used it, you knew you were living dangerously, so often I wouldn't unless all the identifiers were safe even in unintended contexts/times of execution. But otherwise I believe it re-prompted you for the identifier name, which would stop a wild paste or keyboard bounce. I guess I didn't have a problem with this, but I never gave it much thought, either.

jeremyevans:
I'm going to resolve the conversations and then merge this.  We can address some of these conversation topics in future pull requests.

lib/ubi_cli.rb
enescakir: I was planning to write similar comments in the #cli channel, but it seems like it has already been discussed here. I think it should ask for confirmation by default for destructive operations, and `--force` should bypass that.

I agree that CLI users know what they are doing, but people might accidentally run destructive commands on the wrong resource due to typos or selecting the wrong command from CTRL+R, etc. Once executed, there's no way to cancel or revert it. IMHO, a confirmation feels safer.

In the future, we might add a `production` or `do not allow destruction` toggle in the UI to help prevent accidental deletions of production resources.

cli-commands/vm/post/create.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+UbiRodish.on("vm").run_on("create") do
+  options("ubi vm location/vm_name create [options] public_key", key: :vm_create) do
+    on("-6", "--ipv6-only", "do not enable IPv4")
+    on("-b", "--boot_image=image_name", "boot image (ubuntu-noble,ubuntu-jammy,debian-12,almalinux-9)")
+    on("-p", "--private_subnet_id=id", "place VM into specific private subnet")
+    on("-s", "--size=size", "server size (standard-{2,4,8,16,30,60})")
+    on("-S", "--storage_size=size", "storage size (40, 80)")
+    on("-u", "--unix_user=username", "username (default: ubi)")
```
enescakir, line 10: Should we mix `-` and `_` in parameter names? `--ipv6-only` vs `--boot_image` in vm create?

cli-commands/vm/post/show.rb
```diff
@@ -0,0 +1,54 @@
...
+
+UbiRodish.on("vm").run_on("show") do
+  fields = %w[id name state location size unix_user storage_size_gib ip6 ip4_enabled ip4 private_ipv4 private_ipv6 subnet firewalls].freeze.each(&:freeze)
+  firewall_fields = %w[id name description location path firewall_rules].freeze.each(&:freeze)
+  firewall_rule_fields = %w[id cidr port_range].freeze.each(&:freeze)
+
+  options("ubi vm location/(vm-name|_vm-ubid) show [options]", key: :vm_show) do
+    on("-f", "--fields=fields", "show specific fields (default: #{fields.join(",")})")
+    on("-r", "--firewall_rule_fields=fields", "show specific fields (default: #{firewall_rule_fields.join(",")})")
+    on("-w", "--firewall_fields=fields", "show specific fields (default: #{firewall_fields.join(",")})")
```
enescakir, line 11: What do you think about combining them into `fields` and get firewall fields with a prefix. i.e. `-f id,name,firewalls:name`

cli-commands/pg/post/show.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+UbiRodish.on("pg").run_on("show") do
+  fields = %w[id name state location vm_size storage_size_gib version ha_type flavor connection_string primary earliest_restore_time latest_restore_time firewall_rules metric_destinations ca_certificates].freeze.each(&:freeze)
```
enescakir, line 4: Can we read `fields` from openapi.yml somehow?

Otherwise, when we introduce new fields, we need to remember to add them in multiple places.

jeremyevans: I don't think the additional complexity inherent in that approach is worth the benefit.

cli-commands/vm/post/show.rb
jeremyevans: How would you handle multiple nested fields in that case? `-f id,name,firewalls:name,firewalls:id`?  There aren't many nested fields, so I guess the repetitiveness issue isn't too severe.  However, it adds complexity.  The only reason I can think to do it is if you wanted to use `-r` or `-w` for other things.

cli-commands/vm/post/create.rb
jeremyevans: I chose the current approach only because it was simpler to have the long options match the api parameters, as the option parser uses the long option names as the hash keys by default.  It's simple to convert `-` to `_` though.  Or we could do `--ipv6_only`.  Which do you prefer?

lib/ubi_cli.rb
jeremyevans: I'll work on support for confirmation for `destroy` commands.  Are there any other commands that should require confirmation?

enescakir: I think `destroy` is enough for now. May be resetting super user password 🤔

jeremyevans: I looked at that one, but we don't require confirmation on the web for it, so it doesn't make sense to require confirmation in the cli.

PR #2789, opened 2025-02-11
Sshd configure more

fdr:
@byucesoy kinda an invasive one that will require some care in deployment, wanted to run it by you, and why not @shikharbhardwaj since you work together more closely, and he saw my last patch fixing that typo in 4461be0b541bacb8f6e2d92d28acb11a5813a942.

Also @jeremyevans because this should both obsolete, and be superior to forking the `timeout` program every time we do a command

.
Terminate SSH sessions promptly and their processes

This is done with two configuration changes: one, to use ClientAlive
settings in the SSH server to take action on connections without a
responsive client, and secondly, to use `logind` configuration to
specify that processes that belong to that session are to be
terminated rather than re-parented as orphans.

A downside is that running `ssh` with even a single `-v` (verbose)
flag will spam annoying messages to the screen during the session.  I
have often used the lowest level of verbosity to supervise at what
stage an establishing connection is hanging or slow at, now I cannot
leave that flag on anymore.

Verification is a bit of a pain. First, it's best to have two paths to
the test computer, e.g. allocate two VMs, where one acts as a SSH jump
host to a test VM, as well as connecting to the test VM directly from
your laptop: that way you can cut network access to the jump host and
observe the results.

Secondly, it's important to know that signal propagation in SSH is
different if a `pty` is allocated (e.g. for interactive mode): with a
`pty`, SIGHUP is propagated, and most processes will decide to exit at
that point.  Without a `pty`, there is no signal.  So, when testing
this, it is better to do something like:

    ssh host sleep 3600

To start a session that is non-interactive, and waits.  By omitting
the changes in this patch, you should be able to see `sleep` hanging
around after the SSH server has closed the forked SSH process because
of a non-responsive client.  You can use `pstree -s` to check its
parentage.

From the session with the jump host, start such a hanging command.
Then, on your other connection, drop TCP packets to the jump host's
address, by running `nft -f` on nftable rules like this, substituting
the `saddr` for the jump host address:

    add table inet filter
    flush table inet filter
    table inet filter {
      chain input {
        type filter hook input priority filter; policy accept;
        ip6 saddr 2a01:4f8:2b01:d85:214a::2 tcp dport 22 drop
      }
    }

By using commands like `w` and `loginctl list-sessions`, you should
see the process disappear momentarily.

If you need to re-set the test and allow the jump host to send traffic
again, use `nft -f` again with:

    delete table inet filter;

Sync the end of rhizome bootstrapping

Out of an abundance of caution, block until files and directories have
been flushed, so an unlucky reboot that did not go through graceful
flushing would leave empty or incomplete files or directories that are
connected to the sensitive process of setting up the rhizome user and
secure shell.

Move openssh server configuration to BootstrapRhizome

Previously, it was prep_host, which only affected the bare metal
layer, whereas I think we would be better off with uniform
configuration for managed services (e.g. Postgres) as well.

### DIFF
--- a/prog/bootstrap_rhizome.rb
+++ b/prog/bootstrap_rhizome.rb
@@ -14,6 +14,26 @@ class Prog::BootstrapRhizome < Prog::Base
     hop_setup
   end
 
+  # Taken from https://infosec.mozilla.org/guidelines/openssh
+  SSHD_CONFIG = <<SSHD_CONFIG
+# Supported HostKey algorithms by order of preference.
+HostKey /etc/ssh/ssh_host_ed25519_key
+HostKey /etc/ssh/ssh_host_rsa_key
+HostKey /etc/ssh/ssh_host_ecdsa_key
+
+KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp521,ecdh-sha2-nistp384,ecdh-sha2-nistp256,diffie-hellman-group-exchange-sha256
+
+Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes192-ctr,aes128-ctr
+
+MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com
+
+# Password based logins are disabled - only public key based logins are allowed.
+AuthenticationMethods publickey
+
+# LogLevel VERBOSE logs user's key fingerprint on login. Needed to have a clear audit track of which key was using to log in.
+LogLevel VERBOSE
+SSHD_CONFIG
+
   label def setup
     pop "rhizome user bootstrapped and source installed" if retval&.dig("msg") == "installed rhizome"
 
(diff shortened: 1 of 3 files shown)


prog/bootstrap_rhizome.rb
```diff
@@ -32,8 +32,18 @@ def user
...
+
+# Terminate sessions with clients that cannot return packets rapidly.
+ClientAliveInterval 2
+ClientAliveCountMax 4
 SSHD_CONFIG
 
+  LOGIND_CONFIG = <<LOGIND
+[Login]
+KillOnlyUsers=rhizome
+KillUserProcesses=yes
```
jeremyevans, line 44: `KillUserProcesses` is supposed to default to `yes` for  non-root users, according to https://www.freedesktop.org/software/systemd/man/latest/logind.conf.html . I don't see a problem with including this, but either the that documentation isn't matching behavior, or it is matching and this isn't necessary.

jeremyevans (approved):
This makes sense to me.  Note that we still may want to use #2688 even with this.  If there is a connection loss between the server and client, and the server cannot reach the client, the client can still hang even if the server disconnects.

Alternatively, for all `ssh` commands submitted by the control plane, they should use an `ssh` config that sets `ServerAliveCountMax 4` and `ServerAliveInterval 2` to match these settings.

PR #2792, opened 2025-02-11
Do not report down if the runner is getting destroyed

furkansahin:
Our monitor prints out all of the "down" signals, however, because of the continuous churn of the runner VMs, it's almost impossible to watch the down signals in a reasonable way. This commit simply adds a check. This way, if the runner is getting destroyed, we report it as up.

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -65,11 +65,15 @@ class GithubRunner < Sequel::Model
     rescue
       "down"
     end
-    aggregate_readings(previous_pulse: previous_pulse, reading: reading, data: {available_memory: available_memory})
+    aggregate_readings(previous_pulse: previous_pulse, reading: reading, data: {available_memory: available_memory}) unless is_destroying?
   end
 
   def self.redacted_columns
     super + [:workflow_job]
+  end
+
+  def is_destroying?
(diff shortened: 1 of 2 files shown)


model/github_runner.rb
```diff
@@ -63,14 +63,18 @@ def check_pulse(session:, previous_pulse:)
       available_memory = session[:ssh_session].exec!("awk '/MemAvailable/ {print $2}' /proc/meminfo").chomp
       "up"
     rescue
-      "down"
+      is_destroying? ? "up" : "down"
```
enescakir, line 66: Can't we just ignore it if it's destroying instead of saying it's `up`?

furkansahin: I guess we can just return nil. Still, I don't know its impact on the monitoring. So, @byucesoy to the rescue.

byucesoy: Monitor actually refreshes its model cache to remove resources it doesn't need to monitor, but it does this only every 5 minutes, because we want to limit database access as much as possible.

I guess eventual solution we would converge would be some sort of pub/sub mechanism to inform monitor about model changes. Though, it is a too big of a change at the moment and not a high priority.

PR #2805, opened 2025-02-12
Delete cache entry if it is not committed in 20 mins

velioglu:
Cache entries are saved in 2 steps (i) Reserve (ii) Commitl. If an entry is reserved but not committed, it causes subsequent save requests for that entry to fail. To prevent it, removing non-committed entries after 20 mins. 20 mins is calculated by adding a 1.5 buffer after max upload time for an entry with max size.


prog/github/github_repository_nexus.rb
```diff
@@ -91,6 +91,18 @@ def cleanup_cache
       .limit(200)
       .destroy
 
+    # Destroy cache entries if it is created 20 minutes ago
+    # but couldn't committed yet. 20 minutes decided as during
+    # our performance tests uploading 10GB of data (which is
+    # the max size for a single cache entry) takes ~8 minutes at most.
+    # To be on the safe side, 1.5x buffer is added.
+    twenty_minutes_ago = Time.now - 20 * 60
```
enescakir, line 99: In the worst case, if it uploads at 10 MB/s, it will take 17 minutes. Let's be on the safer side and increase it to 30 minutes.

prog/github/github_repository_nexus.rb
```diff
@@ -91,6 +91,18 @@ def cleanup_cache
...
+    github_repository.cache_entries_dataset
+      .where { created_at < twenty_minutes_ago }
+      .where { committed_at =~ nil }
```
enescakir, line 102: nit:
```suggestion
      .where(committed_at: nil)
```

enescakir (approved):

prog/github/github_repository_nexus.rb
```diff
@@ -91,6 +91,18 @@ def cleanup_cache
       .limit(200)
       .destroy
 
+    # Destroy cache entries if it is created 30 minutes ago
+    # but couldn't committed yet. 30 minutes decided as during
+    # our performance tests uploading 10GB of data (which is
+    # the max size for a single cache entry) takes ~8 minutes at most.
+    # To be on the safe side, ~2x buffer is added.
+    twenty_minutes_ago = Time.now - 30 * 60
```
enescakir, line 99: It's `thirty_minutes_ago` now :D

prog/github/github_repository_nexus.rb
enescakir: Didn't you like it? 🥲

prog/github/github_repository_nexus.rb
enescakir: Since it's used only once, you can make it inline too. `seven_days_ago` is a variable because it's used multiple times.

enescakir (approved):

PR #2806, opened 2025-02-12
Move ProviderLocations to database

furkansahin:
## Create ProviderLocation Entity to move providers.yaml file to DB
This is a preparation commit for creating customer specific regions. We
need providers and locations in the database because the yaml file is
very inflexible to create customer specific regions.

## Remove providers.yml


lib/ubi_cli.rb
```diff
@@ -394,7 +394,7 @@ def _submit_req(env)
   if Config.test? && ENV["CLOVER_FREEZE"] == "1"
     singleton_class.prepend(Module.new do
       def process(argv, env)
-        DB.block_queries do
+        DB.allow_queries do
```
furkansahin, line 397: @jeremyevans , since the paths now need to pull locations from the database, `block_queries` here cause failure. I don't know what's the best way to solve this, can you please help me?

migrate/20250212_add_provider_locations.rb
```diff
@@ -0,0 +1,32 @@
...
+Sequel.migration do
+  up do
+    create_enum(:provider_name, ["hetzner", "leaseweb", "latitude", "global"])
+    create_table(:provider_location) do
+      column :provider_name, :provider_name, null: false
+      column :display_name, String, null: false
+      column :internal_name, String, null: false
+      column :ui_name, String, null: false
+      column :visible, :boolean, null: false
+      column :id, :uuid, primary_key: true
```
aamederen, line 12: nit: curious about the reason for placing the `id` at the end, rather then the top

migrate/20250212_add_provider_locations.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_enum(:provider_name, ["hetzner", "leaseweb", "latitude", "global"])
```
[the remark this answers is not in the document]
furkansahin: Thanks, updated accordingly

migrate/20250212_add_provider_locations.rb
furkansahin: Not a single reason, I just started typing the columns that I need and added the id later :D

migrate/20250212_add_provider_locations.rb
```diff
@@ -0,0 +1,35 @@
...
+    create_table(:provider_location) do
+      column :id, :uuid, primary_key: true
+      column :display_name, String, null: false
+      column :internal_name, String, null: false
+      column :ui_name, String, null: false
+      column :visible, :boolean, null: false
+      foreign_key :provider_name, :provider_name, type: :text, null: false
+    end
+
+    run <<~SQL
```
jeremyevans, line 17: Sequel does not technically support multiple queries in a `run` command, though it does work on some adapters (including the one we are currently using).  You  can use two `run` calls, or you could switch to using `Sequel::Dataset#import` (see `20241215_authorization_refactor` for an example of that).

model/provider_location.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require_relative "../model"
+
+class ProviderLocation < Sequel::Model
+  include ResourceMethods
```
jeremyevans, line 6: If we only plan on adding locations inside migrations, using `plugin :static_cache` would be a good idea.

migrate/20250212_add_provider_locations.rb
jeremyevans: Unless there is a good reason to deviate, for consistency with other tables, `id` should be the first column.

lib/ubi_cli.rb
furkansahin: Unfortunately no, I would like to dynamically be able to create new locations at the runtime. Can you please elaborate on 1#

migrate/20250212_add_provider_locations.rb
```diff
@@ -0,0 +1,35 @@
...
+    create_table(:provider_name) do
+      column :name, :text, null: false, primary_key: true
+    end
+    create_table(:provider_location) do
+      column :id, :uuid, primary_key: true
+      column :display_name, String, null: false
+      column :internal_name, String, null: false
+      column :ui_name, String, null: false
+      column :visible, :boolean, null: false
+      foreign_key :provider_name, :provider_name, type: :text, null: false
```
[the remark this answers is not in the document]
furkansahin: Yeah, we shouldn't have more than 10 anytime soon :)

lib/ubi_cli.rb
```diff
@@ -85,7 +85,7 @@ def self.list(cmd, fields)
       run do |opts|
         opts = opts[key]
         path = if (location = opts[:location])
-          if LocationNameConverter.to_internal_name(location)
+          if ProviderLocation[display_name: location]&.internal_name
             "location/#{location}/#{fragment}"
           else
             raise Rodish::CommandFailure, "invalid location provided in #{cmd} list -l option"
```
jeremyevans, line 91:
```suggestion
          if location.include?("/")
            raise Rodish::CommandFailure, "invalid location provided in #{cmd} list -l option"
          else
            "location/#{location}/#{fragment}"
```

You can modify the test (or tests) that uses an invalid location to have the location include a `/`.  Other invalid locations will still result in an error during the internal API request.

migrate/20250212_add_provider_locations.rb
furkansahin: I moved it

model/provider_location.rb
furkansahin: The idea is to dynamically be able to add locations at runtime, that's why we cannot use `static_cache`

migrate/20250212_add_provider_locations.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
```
byucesoy, line 1: I think we can call these tables just `provider` and `location`.

byucesoy (approved):
I think we can shorten the table names. For example provider_name.name is repetitive.  Also provider_location sounds like join table, but it is actually just location.

Other than these, I think this PR is good to go.

enescakir (approved):
It looks good to me overall. I added a few minor comments, but nothing is a big blocker.

lib/location_name_converter.rb
```diff
@@ -2,10 +2,10 @@
 
 module LocationNameConverter
```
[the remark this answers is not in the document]
byucesoy: Actually here is another idea... Why don't we use display_name all the way. We can start refering hetzner-fsn1 as eu-central-h1 from now on. We can even update vm_host.location fields to refer it as `eu-central-h1`. Then, there wouldn't be any need for conversion. It would also allow us to drop one of the `xxx_name` fields. The only disadvantage I can think of is that we use internal name in our daily lives while communicating about regions, so there will be period of adjusting for ourselves, but it is temporary.

It is not necessarily in the scope of this PR, so no need to do it now. I just want to get opinions. If general feeling towards is positive, we can do it some time.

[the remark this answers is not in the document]
furkansahin: we need to still keep that information somewhere because the provider APIs need those. So, not sure if we should really remove it.

PR #2810, opened 2025-02-12
Fix 2pc problem between `daemonizer --clean` and hop

fdr:
Running `daemonizer --clean` has an immediate side effect on the remote instance, but there's no guarantee that the transaction nesting the side effect, implied in the label, will commit.  This creates a problem on retry, as the daemonized process will have been cleaned, but `prep` would have not recorded it had done so, and then it would try to run the `prep` daemon again as though nothing ever happened.

Reported-By: Enes Cakir <enes@ubicloud.com>

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -259,9 +259,8 @@ class Prog::Vm::Nexus < Prog::Base
   label def prep
     case host.sshable.cmd("common/bin/daemonizer --check prep_#{q_vm}")
     when "Succeeded"
-      host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")
       vm.private_subnets.each(&:incr_add_new_nic)
-      hop_wait_sshable
+      hop_clean_prep
     when "NotStarted", "Failed"
       secrets_json = JSON.generate({
         storage: vm.storage_secrets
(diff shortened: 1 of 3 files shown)


prog/vm/nexus.rb
```diff
@@ -275,6 +274,11 @@ def before_run
     nap 1
   end
 
+  label def clean_prep
+    host.sshable.cmd("common/bin/daemonizer --clean prep_#{q_vm}")
```
enescakir, line 278: Just to be sure, we can clean multiple times, right? It's idempotent?

enescakir (approved):
🚀

prog/vm/nexus.rb
fdr: good that you asked that. No, daemonizer was not designed that way best I can tell, patch forthcoming

fdr: It's in the follow-up patch.

fdr:
@enescakir requesting re-review since I added something. @mohi-kalantari and @byucesoy also have various reasons to be appraised (and perhaps think if I'm introducing a problem, although...we probably have to fix the problem, rather than adjust this, fundamentally, side effects will have to occur before the transaction recording that they happened)

fdr:
think this one would have spared me some pages and some customer delay just now

byucesoy (approved):

mohi-kalantari (approved):

enescakir (approved):

PR #2813, opened 2025-02-12
Require confirmation for cli vm/pg destroy commands

jeremyevans:
This requires changes to the cli at all levels, since the cli was not originally designed to be interactive.

Have the /cli API endpoint ask for confirmation via an ubi-confirm header.  Make bin/ubi recognize the ubi-confirm header and prompt the user for confirmation.  The confirmation value is sent in the top level --confirm option.

Change UbiCli.destroy, used for both vm and pg destory to support an -f/--force option to destroy without asking for confirmation. Unless the -f option is given, it asks for confirmation.  If confirmation is already provided, and it is correct, it runs the destroy command.  If the confirmation is already provided and it is incorrect, an error is returned.

To ease implementation of this in bin/ubi, a 1.times do block is used. If confirmation is required, argv is prepended with the --confirm option and confirmation value, and then redo is used to send another request.  Rubocop doesn't understand this and tries to remove the 1.times do, breaking the code, so disable that cop.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -87,3 +87,6 @@ Sequel/IrreversibleMigration:
 
 Capybara/ClickLinkOrButtonStyle:
   EnforcedStyle: strict
+
+Lint/UselessTimes:
+  Enabled: false
--- a/bin/ubi
+++ b/bin/ubi
@@ -23,80 +23,98 @@ get_prog = lambda do |prog|
 end
 
 uri = URI(url)
-data = {"argv" => ARGV}.to_json
+argv = ARGV
 headers = {
   "authorization" => "Bearer: #{token}",
   "content-type" => "application/json",
   "accept" => "text/plain",
   "connection" => "close"
 }
+confirmation_prompt = false
+confirmation = nil
 
-if ENV["UBI_DEBUG"] == "1"
-  p [:sending, *ARGV]
-end
-response = Net::HTTP.post(uri, data, headers)
+1.times do
+  if ENV["UBI_DEBUG"] == "1"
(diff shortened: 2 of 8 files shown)


geemus:
Looks good overall. One thing to consider might be whether to allow `name` or `_id`, or if it should only allow name. Could probably be added later if you so decided (easier to loosen that  than tighten it). Just came to mind as in other places in the API they are used somewhat interchangeably.

jeremyevans:
> Looks good overall. One thing to consider might be whether to allow `name` or `_id`, or if it should only allow name. Could probably be added later if you so decided (easier to loosen that than tighten it). Just came to mind as in other places in the API they are used somewhat interchangeably.

Maybe the wording should be changed?  It only requires that the confirmation value match what was provided as the name or ubid, so if you provide the ubid when deleting, the confirm value needs to be the ubid (including the underscore, currently).

fdr (approved):
form looks fine, cost reasonable, functionality I think my opinions are not as deep as those furnished by @geemus , @enescakir , and @jeremyevans

jeremyevans:
> > Looks good overall. One thing to consider might be whether to allow `name` or `_id`, or if it should only allow name. Could probably be added later if you so decided (easier to loosen that than tighten it). Just came to mind as in other places in the API they are used somewhat interchangeably.
> 
> Maybe the wording should be changed? It only requires that the confirmation value match what was provided as the name or ubid, so if you provide the ubid when deleting, the confirm value needs to be the ubid (including the underscore, currently).

Pushed a commit to address this.  I'll merge this after CI passes.

PR #2815, opened 2025-02-13
Adding UI support and Postgres support for Burstables 

macieksarnowicz:
**Postgres support**
Allowing Postgres to run in Burstable VMs. This commit adds a new set of PostgresSize options that define Postgres VMs running on Burstable-1 and Burstable-2 instances.
    
In order to differentiate billing counters for PG on Standard vs. PG on Burstable, we are now using a combination of vm_family and pg flavor values to create a resource_family used in billing counters. For the UI the family is for now hardcoded as 'standard', but this will be changed in the next commit.
    
For small instances, the work_mem configuration value for PG instance is now set to a 1MB minimum. This will be used for 2GB and 4GB burstable VMs.

**UI support**
Adding a new selector for `family` to create pages and using it to list VM sizes specific to each family. The visibility of the Burstable family, or any other family that requires slice support in the future, is guarded by the project feature flag.
    
 Also updated the logic in generate_vm_options to fully use what is set in Options::VmSize and Options::PostgresSize structures. This should make it easier to add more Vm sizes in the future.


config/billing_rates.yml
```diff
@@ -55,18 +55,30 @@
...
-- { id: 172d6513-0c1c-4f11-9110-db360ef85a33, resource_type: PostgresStandbyVCpu,     resource_family: paradedb,            location: hetzner-fsn1,   unit_price: 0.0011036706, billed_by: duration, active_from: 2025-01-01T00:00:00Z }
-- { id: a4b9f38d-1999-434f-b4ec-d2f59161be4b, resource_type: PostgresStandbyVCpu,     resource_family: paradedb,            location: leaseweb-wdc02, unit_price: 0.0013392858, billed_by: duration, active_from: 2025-01-01T00:00:00Z }
-- { id: 53339b81-f5e4-4911-bfc3-118781b4a1b2, resource_type: PostgresStandbyVCpu,     resource_family: lantern,             location: hetzner-fsn1,   unit_price: 0.0008277530, billed_by: duration, active_from: 2025-01-01T00:00:00Z }
-- { id: 130c8872-6a01-430e-9128-ec035b817d9e, resource_type: PostgresStandbyVCpu,     resource_family: lantern,             location: leaseweb-wdc02, unit_price: 0.0010044644, billed_by: duration, active_from: 2025-01-01T00:00:00Z }
+- { id: c62a5b1d-0ed8-493c-89a5-d19db075d3c2, resource_type: PostgresVCpu,            resource_family: standard-standard,   location: hetzner-fsn1,   unit_price: 0.0005518353, billed_by: duration, active_from: 2025-01-01T00:00:00Z }
```
macieksarnowicz, line 58: @byucesoy - I'd like to hear your feedback on this commit, while it is still in the DRAFT mode. This is bringing in the PG support changes that we have discussed a while ago, now as an official commit to ubicloud. My main questions here is about backward compatibility - the existing VMs use the billing record ID, so should find the records even though the `resource_family` is now changed. However, I do not know if other parts of the billing pipeline will have a problem with this. 
Question: Is this change OK like this, or should I create a new set of billing counters, with new IDs for the existing PostgresVcpu and PostgresStandbyVCpu counters and mark the current ones as archived?

byucesoy: Few thoughts:
- We should reverse the order in the resource_family (i.e. 'bustable-paradedb' to 'paradedb-bustable'), because in the option hierarchy, flavor comes first.
- We should keep existing billing rates as is with their original ids (but move them to bottom of the list in the deprecated section), and generate new billing rates with new ids. After deploying this, I can replace the billing rates used in existing resources.

macieksarnowicz: thanks for the review. I will make those changes today.

macieksarnowicz: Updated now.

helpers/general.rb
```diff
@@ -102,9 +102,9 @@ def current_account
 
   def validate_request_params(required_keys, allowed_optional_keys = [])
     params = if api?
-      request.params
+      request.params.reject { _1 == "family" }
```
enescakir, line 105: Why did we remove it in the general request validator?

macieksarnowicz: @enescakir - where would that be?

byucesoy: Adding it [here](https://github.com/ubicloud/ubicloud/blob/e4f17b4354327b6a2b2aa8a1a8885c967af4775d/helpers/postgres.rb#L11) would be more appropriate.

macieksarnowicz: I tried that initially, when coding this several weeks ago. However I don't think it should be a required parameter. Even making it an optional parameter seems wrong, as the family is derived from size, so if someone passes in a size and family values that are conflicting, we need to handle that as well. IMHO it should not be a user-visible parameter on the request at all.

config/billing_rates.yml
```diff
@@ -49,18 +49,30 @@
...
+- { id: 58ffe0f0-b7ea-4db6-b957-dc75dea7dae2, resource_type: PostgresVCpu,            resource_family: paradedb-standard,   location: hetzner-fsn1,   unit_price: 0.0011036706, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
+- { id: b9e92fa8-8393-4fce-99af-b9e8ce6d376a, resource_type: PostgresVCpu,            resource_family: paradedb-standard,   location: leaseweb-wdc02, unit_price: 0.0013392858, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
+- { id: 471db859-bf6f-4f55-b38d-cd56f6d88f64, resource_type: PostgresVCpu,            resource_family: lantern-standard,    location: hetzner-fsn1,   unit_price: 0.0008277530, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
+- { id: f2e73126-4bc7-416a-bae0-7badad08ea56, resource_type: PostgresVCpu,            resource_family: lantern-standard,    location: leaseweb-wdc02, unit_price: 0.0010044644, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
+- { id: c5c601b9-4c79-4570-a20c-a665253d891d, resource_type: PostgresVCpu,            resource_family: standard-burstable,  location: hetzner-fsn1,   unit_price: 0.0002759177, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
```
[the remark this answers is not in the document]
macieksarnowicz: Sure thing.

helpers/postgres.rb
```diff
@@ -86,18 +86,26 @@ def generate_postgres_options(flavor: "standard")
     options.add_option(name: "name")
     options.add_option(name: "flavor", values: flavor)
     options.add_option(name: "location", values: Option.postgres_locations.map(&:display_name), parent: "flavor")
-    options.add_option(name: "size", values: [2, 4, 8, 16, 30, 60].map { "standard-#{_1}" }, parent: "location")
+    options.add_option(name: "family", values: Option::PostgresSizes.map { _1.vm_family }.uniq, parent: "location") do |flavor, location, family|
+      available_families = Option.families(use_slices: @project.get_ff_use_slices_for_allocation || false).map { _1.name }
```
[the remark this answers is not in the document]
macieksarnowicz: My understanding is that parameters need to be passed in for all parent options, whether they are used or not. Otherwise the filtering method fails.

helpers/general.rb
macieksarnowicz: In other words - in the UI we are using "family" just as a selector to group the sets of "sizes". When user selects the size, they automatically select the family as well. The request carries only the size selection and the Vm Nexus only expects the size value. Internally, in the VM Nexus/Model the family is derived from the size through the use of VmSize structure. So no user request should be specifying the family.

enescakir: We might add a new argument to `validate_request_params`, like `removed_keys/purged_keys/rejected_keys`, and pass a value to it in these endpoints instead of removing `family` for all requests

helpers/postgres.rb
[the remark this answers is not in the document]
macieksarnowicz: I tried again, and the filtering does not work correctly if I exclude those parameters.

helpers/postgres.rb
```diff
@@ -86,18 +86,26 @@ def generate_postgres_options(flavor: "standard")
     options.add_option(name: "name")
     options.add_option(name: "flavor", values: flavor)
     options.add_option(name: "location", values: Option.postgres_locations.map(&:display_name), parent: "flavor")
-    options.add_option(name: "size", values: [2, 4, 8, 16, 30, 60].map { "standard-#{_1}" }, parent: "location")
+    options.add_option(name: "family", values: Option::PostgresSizes.map { _1.vm_family }.uniq, parent: "location") do |flavor, location, family|
```
enescakir, line 89: nit:
```suggestion
    options.add_option(name: "family", values: Option::PostgresSizes.map(&:vm_family).uniq, parent: "location") do |flavor, location, family|
```

helpers/general.rb
macieksarnowicz: I changed this code, I hope the updated version is more acceptable. @enescakir, @byucesoy - please let me know what you think.

byucesoy (approved):

helpers/general.rb
byucesoy: :+1:

enescakir (approved):

PR #2817, opened 2025-02-13
Use Cloudflare Tunnel instead of ngrok

enescakir:
- **Use Cloudflare Tunnel instead of ngrok**
  We need to make our web server public so that the webhook can reach to
  our controller from the GitHub App.
  
  We were using ngrok, but we've hit its monthly free limit.
  
  Cloudflare Tunnel[^1] offers similar features, and we already use
  Cloudflare extensively.
  
  So, instead of relying on another service like ngrok, it makes more
  sense to use Cloudflare Tunnel for free.
  
  [^1]: https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/
  

- **Build frontend assets in E2E as well**
  This lets us connect to the Ubicloud console while E2E tests are
  running.
  
  We can also investigate issues or manually test the UI during E2E tests.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -56,22 +56,11 @@ jobs:
         CLOVER_DATABASE_URL: postgres://${{ env.DB_USER }}:${{ env.DB_PASSWORD }}@localhost:5432/${{ env.DB_NAME }}
       run: bundle exec rake _test_up
 
-    - name: Cache ngrok binary
-      id: cache-ngrok
-      uses: actions/cache@v4
-      with:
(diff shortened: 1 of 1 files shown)


fdr (approved):
compound products, amirite

enescakir:
It seems the second run failed because of the cache. Let me fix that

enescakir:
It turned out that `/usr/local/bin/cloudflared` was just a symlink, so its cache was invalid. I changed it to `/usr/bin/cloudflared`

velioglu (approved):

PR #2818, opened 2025-02-13
Add billing check and free quota for AI inference (v2)

0xjunhao:
Provide free quota for AI inference. Projects without valid billing info will have requests rejected after the free quota is exhausted.

Free quota is configurable in free_quota.yml, currently set to 500,000 tokens/month/project, based on conversation with Ozgun.

Here's how the frontend would look like for projects with and without valid billing respectively.

Screenshot 2025-02-07 at 6 55 13 PM
Screenshot 2025-02-07 at 6 55 25 PM

The invoice will convert the free tokens to a dollar credit. Here's how the invoice will look like.
Screenshot 2025-02-14 at 5 09 34 PM
Screenshot 2025-02-14 at 5 09 47 PM

### DIFF
--- a/config/free_quotas.yml
+++ b/config/free_quotas.yml
@@ -1,0 +1,1 @@
+- { name: inference-tokens, value: 500000, resource_type: "InferenceTokens" }
--- a/lib/free_quota.rb
+++ b/lib/free_quota.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+require "yaml"
+
+class FreeQuota
+  # :nocov:
+  def self.freeze
+    free_quotas
+    super
+  end
+  # :nocov:
+
+  def self.free_quotas
+    @free_quotas ||= begin
+      quotas = YAML.load_file("config/free_quotas.yml")
+      quotas.each_with_object({}) do |item, hash|
+        item["billing_rate_ids"] = BillingRate.from_resource_type(item["resource_type"]).map { _1["id"] }
+        hash[item["name"]] = item
+      end
+    end
+    @free_quotas
+  end
+
+  def self.remaining_free_quota(name, project_id)
+    free_quota = free_quotas[name]
+    used_amount = BillingRecord
+      .where(project_id:, billing_rate_id: free_quota["billing_rate_ids"])
+      .where { Sequel.pg_range(span).overlaps(Sequel.pg_range(FreeQuota.begin_of_month...Time.now)) }
+      .sum(:amount) || 0
+    [0, free_quota["value"] - used_amount].max
+  end
+
+  def self.get_exhausted_projects(name)
+    free_quota = free_quotas[name]
+    BillingRecord
+      .where(billing_rate_id: free_quota["billing_rate_ids"])
+      .where { Sequel.pg_range(span).overlaps(Sequel.pg_range(FreeQuota.begin_of_month...Time.now)) }
+      .group(:project_id)
+      .having { sum(:amount) >= free_quota["value"] }
+      .select(:project_id)
+  end
+
+  def self.begin_of_month
+    Time.new(Time.now.year, Time.now.month, 1)
+  end
+end
--- a/lib/invoice_generator.rb
+++ b/lib/invoice_generator.rb
@@ -48,6 +48,8 @@ class InvoiceGenerator
             line_item_content[:amount] = li[:amount].to_f
             line_item_content[:duration] = li[:duration]
             line_item_content[:cost] = li[:cost].to_f
+            line_item_content[:begin_time] = li[:begin_time]
+            line_item_content[:unit_price] = li[:unit_price].to_f
 
             resource_content[:line_items].push(line_item_content)
             resource_content[:cost] += line_item_content[:cost]
@@ -81,6 +83,26 @@ class InvoiceGenerator
           project_content[:credit] += project_content[:github_credit]
           project_content[:cost] -= project_content[:github_credit]
         end
+
+        # Each project have some free AI inference tokens every month
+        # Free AI tokens WILL be shown on the portal billing page as a separate credit.
+        free_inference_tokens_remaining = FreeQuota.free_quotas["inference-tokens"]["value"]
+        free_inference_tokens_credit = 0.0
+        project_content[:resources]
+          .flat_map { _1[:line_items] }
+          .select { _1[:resource_type] == "InferenceTokens" }
+          .sort_by { |li| [li[:begin_time].to_date, -li[:unit_price]] }
+          .each do |li|
+            used_amount = [li[:amount], free_inference_tokens_remaining].min
+            free_inference_tokens_remaining -= used_amount
+            free_inference_tokens_credit += used_amount * li[:unit_price]
+          end
+        free_inference_tokens_credit = [free_inference_tokens_credit, project_content[:cost]].min
+        if free_inference_tokens_credit > 0
+          project_content[:free_inference_tokens_credit] = free_inference_tokens_credit
+          project_content[:cost] -= project_content[:free_inference_tokens_credit]
+        end
+
         project_content[:cost] = project_content[:cost].round(3)
 
         if @save_result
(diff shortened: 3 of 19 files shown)


routes/project/inference_api_key.rb
```diff
@@ -5,6 +5,9 @@ class Clover
     r.web do
       r.get true do
         @inference_api_keys = Serializers::InferenceApiKey.serialize(inference_api_key_ds.all)
+        @remaining_free_quota = FreeQuota.remaining_free_quota("inference-tokens", @project.id)
+        @free_quota_unit = "inference tokens"
```
bsatzger, line 9: I don't think we need to dynamically set this here. in the views we could simply change
`free <%= @free_quota_unit %> available` to `free inference tokens available`.

prog/ai/inference_endpoint_replica_nexus.rb
```diff
@@ -193,10 +196,10 @@ def ping_gateway
     resp = vm.sshable.cmd("sudo curl -m 5 -s -H \"Content-Type: application/json\" -X POST --data-binary @- --unix-socket /ie/workdir/inference-gateway.clover.sock http://localhost/control", stdin: body.to_json)
     project_usage = JSON.parse(resp)["projects"]
     Clog.emit("Successfully pinged inference gateway.") { {inference_endpoint: inference_endpoint.ubid, replica: inference_endpoint_replica.ubid, project_usage: project_usage} }
-    update_billing_records(project_usage)
+    update_billing_records(project_usage, free_quota_exhausted_projects_ds)
```
bsatzger, line 199: `free_quota_exhausted_projects_ds` isn't required here anymore

prog/ai/inference_endpoint_replica_nexus.rb
```diff
@@ -193,10 +196,10 @@ def ping_gateway
...
   end
 
-  def update_billing_records(project_usage)
+  def update_billing_records(project_usage, free_quota_exhausted_projects_ds)
```
bsatzger, line 202: `free_quota_exhausted_projects_ds` isn't required here anymore

lib/invoice_generator.rb
```diff
@@ -123,7 +123,15 @@ def active_billing_records
     active_billing_records = BillingRecord.eager(project: [:billing_info, :invoices])
       .where { |br| Sequel.pg_range(br.span).overlaps(Sequel.pg_range(@begin_time...@end_time)) }
     active_billing_records = active_billing_records.where(project_id: @project_ids) unless @project_ids.empty?
-    active_billing_records.all.map do |br|
+    sorted_records = active_billing_records.sort_by { |br| [br.span.begin.to_date, -br.billing_rate["unit_price"]] }
```
bsatzger, line 126: This sorting could be costly. It looks like we only need the inference records sorted. Should we only sort those?

enescakir: > Another thing: I'm not sure if the logic is correct. active_billing_records can be for multiple @project_ids. If I'm not mistaken the logic assumes active_billing_records is always called for a single project.

Yes, it has billing records for all projects, and then we group them by project_id. So we shouldn't calculate them here.

https://github.com/ubicloud/ubicloud/blob/7480075eda2c969f0542f5a2d74a0e187c4f44db/lib/invoice_generator.rb#L18

enescakir: IMHO, `active_billing_records` should include all usages without any deductions for credits or discounts.

We calculate discounts and credits at https://github.com/ubicloud/ubicloud/blob/7480075eda2c969f0542f5a2d74a0e187c4f44db/lib/invoice_generator.rb#L62-L84

When we deduct them at `active_billing_records`, they won't appear as credits on the invoice or current usage page.

enescakir: You may apply credits similar to how GitHub does.

https://github.com/ubicloud/ubicloud/blob/7480075eda2c969f0542f5a2d74a0e187c4f44db/lib/invoice_generator.rb#L75

This way, it shows as a credit on the invoice, and we also have the `project_content[:github_credit]` value in the `invoice.content` record. So, if we want, we can display it as `Free GitHub Minutes: -$0.84` or free token on the invoice.

lib/free_quota.rb
```diff
@@ -0,0 +1,53 @@
...
+    free_quotas
+    super
+  end
+  # :nocov:
+
+  def self.free_quotas
+    @free_quotas ||= begin
+      quotas = YAML.load_file("config/free_quotas.yml")
+      quotas.each_with_object({}) do |item, hash|
+        hash[item["name"]] = item
```
enescakir, line 17: You can also calculate the applicable billing rate IDs here, so it won't cost anything at runtime, and we only need to load them once. Then you can use it as `free_quota["billing_rate_ids"]`

lib/free_quota.rb
```diff
@@ -0,0 +1,53 @@
...
+    free_quota = free_quotas[name]
+    begin_time = get_begin_time
+    tokens_used = BillingRecord.where(
+      project_id: project_id,
+      billing_rate_id: get_applicable_billing_rate_ids(free_quota)
+    ).where {
+      Sequel.pg_range(span).overlaps(Sequel.pg_range(begin_time...Time.now))
+    }.sum(:amount) || 0
+    [0, free_quota["value"] - tokens_used].max
+  end
```
enescakir, line 37: nit:
- not need to `begin_time` since it's used once. we might also rename the method as `begin_time` instead of `get_begin_time`
- single line `where` clauses look more like the `self.get_free_quota_exhausted_projects` method.
- if you apply https://github.com/ubicloud/ubicloud/pull/2818/files#r1956158740
- `tokens_used` -> `used_amount`
- `project_id: project_id` -> `project_id:`

```suggestion
  def self.remaining_free_quota(name, project_id)
    free_quota = free_quotas[name]
    used_amount = BillingRecord
      .where(project_id:, billing_rate_id: free_quota["billing_rate_ids"])
      .where { Sequel.pg_range(span).overlaps(Sequel.pg_range(begin_time...Time.now)) }
      .sum(:amount) || 0
    [0, free_quota["value"] - used_amount].max
  end
```

lib/free_quota.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def self.get_free_quota_exhausted_projects(name)
```
enescakir, line 39: It returns a dataset, and also we call it via `FreeQuota.get_free_quota_exhausted_projects`, no need to say free quota twice.

`FreeQuota.exhausted_projects_ds(name)` looks more natural to me.

```suggestion
  def self.exhausted_projects_ds(name)
```

config/free_quotas.yml
```diff
@@ -0,0 +1,5 @@
+[{
+  name: inference-tokens,
+  value: 500000,
+  resource_type: "InferenceTokens"
+}]
```
enescakir, line 5: We keep our YAML config list as single lines because it makes them easier to maintain when we have more items.

```suggestion
- { name: inference-tokens, value: 500000, resource_type: "InferenceTokens" }
```

routes/project/inference_api_key.rb
0xjunhao: We have to calculate this amount in order to know whether there are free tokens available, so why not just display the amount?

prog/ai/inference_endpoint_replica_nexus.rb
0xjunhao: Done.

prog/ai/inference_endpoint_replica_nexus.rb
0xjunhao: Done.

config/free_quotas.yml
0xjunhao: Done.

lib/free_quota.rb
0xjunhao: Done.

lib/free_quota.rb
0xjunhao: Done.

lib/free_quota.rb
0xjunhao: Done.

0xjunhao:
@fdr

lib/invoice_generator.rb
0xjunhao: Done. Moved the calculation next to where the credits are calculated and will show it as a separate credit.

routes/project/inference_api_key.rb
bsatzger: sure, we need to calculate the amount and display the amount. but the unit will always be "inference tokens". I was wondering why we set the unit dynamically.

bsatzger (approved):

PR #2819, opened 2025-02-13
Improve cli error reporting

jeremyevans:
This makes necessary changes to Rodish to improve error reporting, then uses those new features to improve the error reporting in UbiCli.  This improves command discoverability in bin/ubi, and makes error messages much more helpful.

To further improve discoverability, this adds `ubi help`, which can be used to get help for a specific command, or for all commands at once.

This fixes the usage lines to be consistent across commands.

This sets proper dependencies so that modifying cli-related code in development works correctly. 

The first commit is the same as PR #2813 to avoid the need to resolve conflicts.


spec/routes/api/cli/help_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      Options:
+          -6, --ipv6-only                  do not enable IPv4
+    OUTPUT
+  end
+
+  it "shows usage for all subcommands of command if -ru is given" do
+    expect(cli(%w[help -ru vm])).to include <<~OUTPUT
+      Usage: ubi vm list [options]
+      Usage: ubi vm location/vm_name create [options] public_key
```
geemus, line 53: It might be worth considering if this should be two separate `Usage: ...` lines or one `Usage: ` header, with two newline delineated instances (ie similar to options).

You could also consider whether individual options should be inline, or if they warrant having their own lines as well. Either could be fine, but consistency matters.

spec/routes/api/cli/help_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    OUTPUT
+  end
+
+  it "shows error and help for top level if top level command invalid" do
+    expect(cli(%w[help foo], status: 400)).to eq <<~OUTPUT
+      invalid command: foo
```
geemus, line 59: It might look more consistent if `invalid command` were titlecase, or at least sentence case. Other things before colons are, so it seems a little inconsistent. That might also draw the eye. You could also consider additional treatment for errors or warnings, in my recent usage I've used a exclamation prefix, so something like `! Invalid Command: foo` might stick out more and hopefully draw the eye more as well.

spec/routes/api/cli/help_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      Usage: ubi [options] [subcommand [subcommand-options] ...]
+
+      Options:
+              --version                    show program version
+              --help                       show program help
+              --confirm=confirmation       confirmation value (not for direct use)
```
geemus, line 66: Option descriptions could also potentially be sentence case. I don't have much feeling/preference one was or the other, but another place where consistency matters.

geemus:
@jeremyevans Thanks for bringing my attention, I've added a few style things to consider. I'll share a couple broader points from my experience that might be of interest as well.

1. It can be helpful to have a style guide, which outlines expectations for the appearance of different aspects of the output. So it might outline what should be lowercase vs title or sentence case, when punctuation is expected (or not), and other things of that nature. It can also note things like what errors ought to look like. When it's just ascii this can be a pretty simple/short doc, but it can become even more valuable (and elaborate) if you get into colors/emojis/etc.
2. You may find some value in golden file style testing (I certainly have) for some of the outputs, especially things like help files. Even if they don't change regularly, as they are systematically generated, if you change something about their generation they are ALL going to change. Having to manually update all the related tests can be pretty annoying, so having something that can just record that output and then test against it for regressions can be pretty nice.

fdr:
>     2. You may find some value in golden file style testing (I certainly have) for some of the outputs, especially things like help files. Even if they don't change regularly, as they are systematically generated, if you change something about their generation they are ALL going to change. Having to manually update all the related tests can be pretty annoying, so having something that can just record that output and then test against it for regressions can be pretty nice.

I also kind of like golden files to see what the output is like in a compact way, if the complexity is a reasonable size (whatever that means). `pg_regress`, for all its limitations, did quite a lot well with golden files. A cost is that it may exist in ungratifying duplication with unit tests. 

More so than most of our programs, the precise formatting changes produced by tweaks to the CLI code and how they render literally on the screen is of somewhat greater importance.

spec/routes/api/cli/help_spec.rb
jeremyevans: You get this output for `-ru`, showing all usages for the command and all subcommands.  The example only tests two lines, but there will generally be more (up to 27 lines currently).  Since all lines start with `Usage: `, we could strip that out, but I don't think we should add a header, since the user is explicitly asking only for usages.  Let me know if you think it's better to strip the `Usage: ` out.

I think inlining the options into the usage would result in excessively long lines in some cases, so to be consistent, I think it's best never to inline them.

spec/routes/api/cli/help_spec.rb
jeremyevans: That makes sense.  I'll make this change and add it to this PR.

spec/routes/api/cli/help_spec.rb
jeremyevans: I think we consistently use lower case currently, so unless there is a good argument for changing that, I think we should keep it as-is.

jeremyevans:
I'll look into switching the tests to use a golden file testing approach.

fdr:
> I'll look into switching the tests to use a golden file testing approach.

yeah not to be too pushy, may or may not be solvent, fwiw, pg_regress the workflow was like this (not to say it was optimal in every respect).

a `sql` directory contained inputs.
an `expected` directory contained *expected* outputs, committed to the source
a `results` directory contains the output of each sql file in the run

And it would produce a diff of the difference between these two directories. I suppose that the directory format was driven by the desire to make it easy on the tool `diff`.

To update golden files, it was as simple as: "cp results/* expected". The diff could be reviewed in `git` (although pg_regress dates back before that).

I found many bugs, or at least interesting side-effects, of my patches because they could affect other parts of the system. The postgres case is a bit different than ours, the exact presentation is not as essential to what we're seeking to accomplish. And it has a major weakness, which is it is hard to target particular branches in the code. But for what it did, I found the workflow rather nice for things it could verify.

spec/routes/api/cli/help_spec.rb
jeremyevans: Pushed a commit for this.

spec/routes/api/cli/help_spec.rb
geemus: Sounds good to me, just seemed like something to make an intentional choice about instead of perhaps accidentally falling into something and/or being inconsistent.

geemus:
Another thing that can be nice with golden files is that they can be somewhat "cheaper" to produce. So whereas you might only have unit tests for a subset of things that you feel exercise the edges, it can be easy to have a golden file test for the help output (and variations) on every command if you want. We tend toward that in our usage, because it feels like you might as well.

Also, as you suggest the specific formatting really matters and this can be a good way to reveal that.

Another idea that I've been interested in and not pursued at the intersection of golden files and style guides is that of linting. I suspect you could put together linters against the golden files that could help enforce particular things, like potentially raising if things are not lowercase when they are expected to be (to pull from an example in our discussion on this PR).

spec/routes/api/cli/help_spec.rb
geemus: Ah, yeah that makes sense. I was focused on the output as I saw it in the test (which seemed inconsistent) and not so much the call pattern that resulted in it.

Sorry for the ambiguity, when I mentioned inline individual options, I meant more that in cases where there is only a single option, it would look something like:
```
Option: --version version
```
as opposed to
```
Option:
  --version  version
```

So they would still appear as their own section, it was more a matter of a question like: is inline unique to Usage, or is it a generalized special case for things to display where there is only one (which would otherwise be newline delineated). Does that make more sense?

jeremyevans: The option formatting comes directly from ruby's optparse library.  We can probably override it, but I don't think we should without a good reason.  Single options being on their own line seems fine to me, and it is consistent.

The `help -u` option is special case, where instead of asking the optparse library to return a string version of the parser (`OptionParser#to_s`) , we only take the option parser's banner.

jeremyevans:
I've added support for a golden files approach to this PR.

cli-commands/pg/post/delete-firewall-rule.rb
```diff
@@ -1,11 +1,17 @@
...
-  end
+UbiRodish.on("pg").run_on("delete-firewall-rule") do
+  options("ubi pg location/(pg-name|_pg-ubid) delete-firewall-rule id")
+
+  args 1, invalid_args_message: "rule id is required"
+
+  run do |ubid|
+    if ubid.include?("/")
+      raise Rodish::CommandFailure, "invalid firewall rule id format"
+    end
```
geemus, line 11: Seeing this pattern, and that it is just applied to a subset (ie firewall rule above with cidr didn't have any format checks) makes me wonder if this should be a more universal pattern. Like maybe the `args` command could take a regex or something (as a blunt instrument in terms of what comes immediately to mind). Also in this particular case, there are presumably a lot of invalid values other than including a slash. Though I guess maybe anything that doesn't cause a client side error could potentially be delegated to the server for errors?

cli-commands/vm/post/show.rb
```diff
@@ -1,14 +1,14 @@
...
-  firewall_rule_fields = %w[id cidr port_range].freeze.each(&:freeze)
+  fields = %w[id name state location size unix-user storage-size-gib ip6 ip4-enabled ip4 private-ipv4 private-ipv6 subnet firewalls].freeze.each(&:freeze)
+  firewall_fields = %w[id name description location path firewall-rules].freeze.each(&:freeze)
+  firewall_rule_fields = %w[id cidr port-range].freeze.each(&:freeze)
 
   options("ubi vm location/(vm-name|_vm-ubid) show [options]", key: :vm_show) do
     on("-f", "--fields=fields", "show specific fields (default: #{fields.join(",")})")
-    on("-r", "--firewall_rule_fields=fields", "show specific fields (default: #{firewall_rule_fields.join(",")})")
-    on("-w", "--firewall_fields=fields", "show specific fields (default: #{firewall_fields.join(",")})")
+    on("-r", "--rule-fields=fields", "show specific firewall rule fields (default: #{firewall_rule_fields.join(",")})")
```
geemus, line 10: I think my inclination on something like this would be to maintain the `firewall` prefix to this. There might not be other rules presently, but that might not always be true and regardless I like the clarity/explicitness. That said, it's certainly more verbose, and that's always a tension in these things. I tend to err on the side of correctness/specificity, even at the cost of being more verbose, but I know others differ on where their tastes lie.

spec/routes/api/cli/golden-files/200-help -r vm.txt
```diff
@@ -0,0 +1,57 @@
+Usage: ubi vm subcommand [...]
+
+Subcommands: list
+
+
+Usage: ubi vm location/(vm-name|_vm-ubid) [options] subcommand [...]
```
geemus, line 6: It might be good to consider a description or summary (similar to openapi, or maybe even pulled from there) which provides a description of what the command does.

spec/routes/api/cli/golden-files/200-vm eu-central-h1_test-vm show -f firewalls.txt
```diff
@@ -0,0 +1,8 @@
+firewall 1:
+  id: fw4gj2v4h1fe3q28q0hnf7g8n1
+  name: default-eu-central-h1-default
+  description: Default firewall
+  location: eu-central-h1
+  path: /location/eu-central-h1/firewall/default-eu-central-h1-default
+  rule 1: fra7jvrz94be7qy8qrq8n1jrp1  0.0.0.0/0  0..65535  
+  rule 2: fr7cv7h5bz324maazy93ykn8s4  ::/0  0..65535
```
geemus, line 8: This seemed a bit inconsistent to me, compared to other representations. I guess I expected these to be nested, maybe more like this from above:
```
Current metric destinations:
  1: et2t1bswsqz21m7j7njfjbp901  foo  https://baz.example.com
  2: etqhb3wgwvge5pepd72pcgzyp6  foo  https://md.example.com
```
That would result in two layers of indentation though, which I don't think I've seen elsewhere in the goldenfiles (yet anyway), though I may have forgotten or overlooked it.

spec/routes/api/cli/golden-files/200[command_execute:pg_dump]-pg eu-central-h1_test-pg -U bar pg_dump.txt
Comment (geemus) [inline spec/routes/api/cli/golden-files/200[command_execute:pg_dump]-pg eu-central-h1_test-pg -U bar pg_dump.txt:1]: I'm not sure why github is treating many of these as BIN files, since they seem to be readable text. It makes review more of a nuisance though, so it would be good to make github not do this if possible (I'm not sure why it's happening off hand, so also uncertain how to remedy).

Comment (geemus) [reply spec/routes/api/cli/golden-files/200[command_execute:pg_dump]-pg eu-central-h1_test-pg -U bar pg_dump.txt:1]: Maybe this is intended, but I wasn't sure since some of them are BIN and some aren't (particularly for pg_dumpall it looks like).

spec/routes/api/cli/golden-files/400---confirm foo pg eu-central-h1_test-pg destroy.txt
```diff
@@ -0,0 +1,2 @@
+
+Confirmation of PostgreSQL database name not successful.
```
geemus, line 2: This might deserve the `! ` prefix treatment to indicate an error.

spec/routes/api/cli/golden-files/400-vm eu-central-h1_test-vm -u foo scp :remote-path local-path invalid.txt
```diff
@@ -0,0 +1 @@
+Only one path should be remote (start with ':')
```
geemus, line 1: This might also deserve the `! ` prefix to indicate error.

cli-commands/pg/post/delete-firewall-rule.rb
jeremyevans: Correct.  We are only checking for slash here to ensure that it will hit the same api route.

spec/routes/api/cli/golden_files_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    fwr_uuids = %w[c81cb3d1-81bc-89f8-96c8-6b4e8d4375bd e3fdf2f2-2603-85f8-b0f9-3fc2c28636cd 99785615-0cc7-8df8-a937-4f1d26b620c8 168054f0-e069-89f8-b12a-e4b010cf47b5 6d590d38-5f88-89f8-b6d8-0a079e1c61b6].cycle
+    expect(FirewallRule).to receive(:generate_uuid).and_invoke(-> { fwr_uuids.next }).at_least(:once)
+    expect(PostgresResource).to receive(:generate_uuid).and_return("97eb0a77-7869-86d0-9dcb-a46416ddc5c9").at_least(:once)
+    expect(PostgresFirewallRule).to receive(:generate_uuid).and_return("6d674a31-e1c1-8ecf-b5ac-363abb5b9185").at_least(:once)
+    expect(PostgresMetricDestination).to receive(:generate_uuid).and_return("bc563e43-9b83-89da-b3ac-d38acc87fd63").at_least(:once)
+
+    Dir["#{golden_file_dir}/*.txt"].each do |f|
+      f = File.basename(f)
+      cmd = f.delete_suffix(".txt")
+      status, cmd = cmd.split("-", 2)
```
geemus, line 52: Since you are stubbing everything out above it may not come to it, but I thought I would share a little from our experience just in case.

For us, we explicitly writing out code that calls the tests (rather than figuring it out from filenames). One of the things we do along with that in some cases is supply substitutions, which are applied to the file. We use that to remove sensitive (or unstable) data. I think that your use of stubbing should largely address that in it's own right, but there may be exceptions.

Just thought I would mention it for your consideration moving forward. Some of our challenge is that in the golang context where we are doing things, stubbing would be much harder at best.

spec/routes/api/cli/golden-files/200-vm eu-central-h1_test-vm show -f firewalls.txt
jeremyevans: There are other places with multiple levels of nesting.  I don't have a problem doing that, but it saves only 3 characters per line (`s/rule /  /`), so in most cases, doing so would make the output larger even with a significant number of rules.

cli-commands/vm/post/show.rb
geemus: Ah, that makes sense. Definitely will want to work within those constraints.

spec/routes/api/cli/golden-files/200[command_execute:pg_dump]-pg eu-central-h1_test-pg -U bar pg_dump.txt
Comment (jeremyevans) [reply spec/routes/api/cli/golden-files/200[command_execute:pg_dump]-pg eu-central-h1_test-pg -U bar pg_dump.txt:1]: The ones that execute commands are going to be binary, as they are argv separated by `\0` characters.

spec/routes/api/cli/golden-files/200-help -r vm.txt
jeremyevans: It's possible to add this for each command.  I worry that in many cases, it will be redundant. (e.g. `vm list` -> `Lists VMs`).

The particular line Usage line you linked is for a command with subcommands, so it's not a command usable by itself.  We could potentially filter those out of the recursive usage output, since all usable subcommands should be listed in the recursive mode.

spec/routes/api/cli/golden-files/400---confirm foo pg eu-central-h1_test-pg destroy.txt
jeremyevans: Makes sense, I'll fix this and the scp case.

spec/routes/api/cli/golden_files_spec.rb
jeremyevans: I guess we could switch things to specify the command lines in the tests.  That would probably make adding new tests easier at least.  I can look into doing that in the future.

spec/routes/api/cli/golden-files/200-help -r vm.txt
geemus: Yeah, in the openapi schema many of them are fairly redundant. I'm not sure if that's inherent vs that we just haven't invested enough there though. Like in the openapi case there is a distinction between summary, which will be something like `Lists VMs` and description, which is intended to be a longer and more detailed explanation that provides context. Kind of like how git commits have a short initial line and then a longer message in some cases.

I don't think we should add this just for the sake of adding it, but it does feel like it could be useful to provide context on at least some commands. Mostly it crossed my mind so I thought I would share the idea, it seems like a bridge to cross if/when we reach it. And maybe at that point it could be added just as-needed and not something to worry about on every command (especially if most of them would be redundant).

spec/routes/api/cli/golden_files_spec.rb
geemus: Yeah, I think that offers more flexibility (since you can then have distinct setup/teardown things, whether they are substitution related or not), but it's also not something that was needed yet so makes sense to defer. As you suggest it probably could make adding new ones easier too, and also could avoid the need to do escaping (since the commands could then just be ruby text strings that match what they would otherwise look like, vs filenames), which might also help make things a bit easier to grok, update and maintain.

spec/routes/api/cli/golden-files/400---confirm foo pg eu-central-h1_test-pg destroy.txt
geemus: Another thing I was noticing with this (and some related messages), is that the capitalization feels weird to me for multi-word objects that appear in the middle of messages.

So, for instance, it makes sense for `Postgres database` at the beginning of a sentence. But when it's in the middle, it feels to be like it should either be all lower case or title case.

Maybe Postgres in particular isn't a good example, since it's a proper noun, but I believe the same capitalization appears for other multi-word things like `Firewall rule` or `Private subnet`. Like other things, the actual answer probably doesn't matter too much as long as it's consistent (and the current usage, to me at least, feels slightly off). 

I'd point to those specific examples, but am having trouble finding them again in the list (and am running out of steam for the evening I fear). Hopefully this is clear enough, but let me know otherwise and I can try to find other spots that are more specific.

jeremyevans: I know the cases you are talking about.  I was planning on cleaning them up after finishing the commands, since it seemed minor.  Probably will have to special case `PostgreSQL`, since that should remain capitalized.

spec/routes/api/cli/golden-files/200-help -r vm.txt
jeremyevans: I filtered out the Usage for commands with subcommands in https://github.com/ubicloud/ubicloud/pull/2842/commits/ec7ac799e9938d174e616d120760b584a9675b6b

We can consider adding details for each command at a later point.

spec/routes/api/cli/golden-files/200-vm eu-central-h1_test-vm show -f firewalls.txt
jeremyevans: I made this change in https://github.com/ubicloud/ubicloud/pull/2842/commits/d53974e8b2dc057d229d4228be9c7a5a3fd72da7

spec/routes/api/cli/golden-files/400---confirm foo pg eu-central-h1_test-pg destroy.txt
jeremyevans: I made these changes in https://github.com/ubicloud/ubicloud/pull/2842/commits/35d6fa7c29b1a40f473647e6e5ea225530642e8a and https://github.com/ubicloud/ubicloud/pull/2842/commits/35d6fa7c29b1a40f473647e6e5ea225530642e8a

spec/routes/api/cli/golden_files_spec.rb
jeremyevans: I refactored how golden file commands are handled, so they are pulled from file content and not filenames in https://github.com/ubicloud/ubicloud/pull/2842/commits/9328334f7444dba7ab7fd09eda639331fc5c2bf6

jeremyevans:
Since #2842 includes these commits and all of the changes made to address the comments in this PR, I'm going to close this, and #2825 and #2826 for similar reasons.

PR #2824, opened 2025-02-14
Introduce AI model mistral-small-3

bsatzger:
For more info about the model, see
https://huggingface.co/mistralai/Mistral-Small-24B-Instruct-2501.

Job building the image:
https://github.com/ubicloud/ai-images/actions/runs/13327977005


furkansahin (approved):

config/ai_models.yml
```diff
@@ -15,3 +15,4 @@
...
 - { id: 43a7016a-69b2-47f7-89b5-eb725a1ab540, model_name: ms-phi-4,            enabled: true, locations: [latitude-ai], vm_size: standard-16,    gpu_count: 1, max_requests: 500, max_project_rps: 100, max_project_tps: 10000, storage_volumes: [{encrypted: true, size_gib: 150}, {read_only: true, image: ai-model-ms-phi-4}],            boot_image: ai-ubuntu-2404-nvidia, tags: {capability: "Text Generation", hf_model: "microsoft/phi-4"},                                                                engine: vllm, engine_params: "\nEnvironment=VLLM_USE_V1=1" }
+- { id: 71e448aa-0308-45c8-a613-80401c9bc7ac, model_name: mistral-small-3,     enabled: true, locations: [latitude-ai], vm_size: standard-16,    gpu_count: 1, max_requests: 500, max_project_rps: 100, max_project_tps: 10000, storage_volumes: [{encrypted: true, size_gib: 150}, {read_only: true, image: ai-model-mistral-small-3}],     boot_image: ai-ubuntu-2404-nvidia, tags: {capability: "Text Generation", hf_model: "mistralai/Mistral-Small-24B-Instruct-2501"},                                      engine: vllm, engine_params: "\nEnvironment=VLLM_USE_V1=1" }
```
0xjunhao, line 18: Just a note that when we enable V1, the max_requests setting may need to be adjusted. vLLM currently has a default max_num_seqs of 256 without V1 and 1024 with V1.
https://github.com/vllm-project/vllm/blob/c9e2d644e728e8e93ef2871276ed7a6b39c1d0eb/vllm/engine/arg_utils.py#L220
I think 500 is a good value without V1, meaning that under full load, we can process 256 requests in parallel while the remaining ~250 wait for the next round, chunk by chunk.

0xjunhao (approved):

config/ai_models.yml
bsatzger: Doubling `max_requests` for V1 after discussions with @0xjunhao

PR #2825, opened 2025-02-14
Refactor Rodish

jeremyevans:
This builds on top of #2819, and adds additional improvements:

* Always set option subhashes, even if they are empty, simplifying
  cli commands that use option subhashes.

* Changes remaining underscored cli options to use dashes instead.

* Adds UbiCli#{vm,pg}_path methods to DRY up cli commands.

* Refactors Rodish and merges UbiRodish into UbiCli.

cc @geemus


geemus (approved):
lgtm

PR #2828, opened 2025-02-15
Support building a static executable of bin/ubi via mruby

jeremyevans:
This requires minor changes to bin/ubi, but it's much easier than
porting to a new language. This does require a few mrbgems (mruby's
version of gems), and the mruby-curl library has a different API
that ruby's net/http library, but it's not a big difference.
mruby doesn't support Kernel#warn by default, so switch to
$stderr.puts and update the rubocop config not to complain about
that.

Currently, this only builds a binary for the current platform, but
because the way this works is actually building a .c file, any
C cross compiler should work.


jeremyevans:
Closing this as a go version has been developed.

PR #2829, opened 2025-02-17
Wait up to 60 seconds if can not authorize R2 with token

velioglu:
To access a bucket for a repository, we are creating access token. When it is first created, it might take up to 60 seconds to propagate according to Cloudflare documentation.

https://developers.cloudflare.com/r2/reference/consistency/#operations-and-consistency

To make sure the initial save cache request works, sleeping up to 60 seconds with 1 second intervals.


routes/runtime/github.rb
```diff
@@ -108,8 +108,28 @@ class Clover
...
+          retries = 0
+          begin
+            upload_id = repository.blob_storage_client.create_multipart_upload(bucket: repository.bucket_name, key: entry.blob_key).upload_id
+            entry.update(upload_id: upload_id)
+          rescue Aws::S3::Errors::Unauthorized => ex
+            retries += 1
+            if retries < 3
+              # :nocov:
+              sleep(Config.test? ? 0 : 1)
+              # :nocov:
```
enescakir, line 126: Instead of branching here, we can mock the sleep in the tests like this:

```ruby
expect_any_instance_of(described_class).to receive(:sleep).with(1).twice
```

enescakir: We talked about it privately; since we freeze classes, we can't mock `sleep`.

enescakir (approved):

PR #2837, opened 2025-02-18
Increase the timeout for kubeadm join and init commands

mohi-kalantari:
We rely on DNS to discover API Server and sometimes it takes a bit of time to have the records updated and successfully run some queries. In order to increase the success rate of kubeadm init, timeout is increased to 10 minutes.

### DIFF
--- a/rhizome/kubernetes/bin/init-cluster
+++ b/rhizome/kubernetes/bin/init-cluster
@@ -50,6 +50,9 @@ config = {
     "local" => {
       "dataDir" => "/var/lib/etcd"
     }
+  },
+  "timeouts" => {
+    "controlPlaneComponentHealthCheck" => "10m0s"
   }
 }
 


rhizome/kubernetes/bin/init-cluster
```diff
@@ -50,6 +50,9 @@ config = {
     "local" => {
       "dataDir" => "/var/lib/etcd"
     }
+  },
+  "timeouts" => {
+    "controlPlaneComponentHealthCheck" => "10m"
```
aamederen, line 55: nit: All such values in the config are expressed in the form of "XXmXs", not sure why though.

```
timeouts:
  controlPlaneComponentHealthCheck: 4m0s
  discovery: 5m0s
  etcdAPICall: 2m0s
  kubeletHealthCheck: 4m0s
  kubernetesAPICall: 1m0s
  tlsBootstrap: 5m0s
  upgradeManifests: 5m0s
```

rhizome/kubernetes/bin/init-cluster
```diff
@@ -50,6 +50,9 @@ config = {
     "local" => {
       "dataDir" => "/var/lib/etcd"
     }
+  },
```
aamederen, line 53: This commit changes the timeout only for the init phase, but the commit message says "Increase the timeout for kubeadm **join and init** commands", so either the message or comment should change.

We haven't got the opportunity to see a join command fail yet, so maybe just stick with init failures for now?

aamederen (changes requested):

rhizome/kubernetes/bin/init-cluster
mohi-kalantari: the go parser can parse that time duration, although for consistency, I added the 0s thanks.

rhizome/kubernetes/bin/init-cluster
mohi-kalantari: yes you're right, fixed the commit message

mohi-kalantari:
It seems that timeout config is not respected and the kubeadm init phase does not use it at all. 

I tried to manually fail the init phase and the timeout was still 4 minutes

https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/phases/init/waitcontrolplane.go#L72
you can see the timeout handed over as the second argument is 0 which I assume falls back to the default of 4 minutes

fdr:
dns delays are constantly dogging us...we could program dnsmasq directly to make it work instantly.

PR #2839, opened 2025-02-18
Use public IPs for minio cluster communication

shikharbhardwaj:
## Description
Use public IPs for peer communication in minio cluster, to avoid private networking performance penalty.

## Tests
- [x] Unit tests
- [x] Create a minio cluster with multiple servers

### DIFF
--- a/model/minio/minio_server.rb
+++ b/model/minio/minio_server.rb
@@ -21,12 +21,24 @@ class MinioServer < Sequel::Model
     enc.column :cert_key
   end
 
+  def generate_etc_hosts_entry
+    entries = ["::1 #{hostname}"]
+    entries += cluster.servers.reject { _1.id == id }.map do |server|
+      "#{server.public_ipv6_address} #{server.hostname}"
+    end
+    entries.join("\n")
+  end
+
   def hostname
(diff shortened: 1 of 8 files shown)


model/minio/minio_server.rb
```diff
@@ -21,6 +21,14 @@ class MinioServer < Sequel::Model
     enc.column :cert_key
   end
 
+  def generate_etc_hosts_entry
+    entries = ["127.0.0.1 #{hostname}"]
+    entries += cluster.servers.reject { _1.id == id }.map do |server|
+      "#{server.public_ipv4_address} #{server.hostname}"
```
shikharbhardwaj, line 27: Wondering if we should make this conditional on provider? (not sure if we know if this problem is isolated to Hetzner setups only)

furkansahin: Provider is not a factor in the poor private networking performance. Its impact is the same for all. However, I agree with @fdr , can we try if it works with IPv6?

shikharbhardwaj: Updated to use public IPv6

furkansahin (approved):

PR #2826, opened 2025-02-14
Add cli ps {list,create,show,destroy} commands

jeremyevans:
This builds on top of #2825, and adds cli support for doing everything the api supports in regards to private subnets.

cc @geemus

PR #2842, opened 2025-02-18
Add cli fw {list,create,show,destroy,attach-subnet,detach-subnet,add-rule,delete-rule} commands

jeremyevans:
Builds on top of #2826. As part of this, fix openapi configuration to allow attach/detach subnet with firewall name.

cc @geemus


lib/ubi_cli.rb
```diff
@@ -4,6 +4,7 @@ class UbiCli
   force_autoload = Config.production? || ENV["FORCE_AUTOLOAD"] == "1"
 
   FRAGMENTS = {
+    "fw" => "firewall",
```
geemus, line 7: It might be worth considering re-using these fragments for the `...-fields` params as well. If these are consistently shortened in the commands anyway, it seems like using the same thing in the params makes sense.

This might also address the `firewall-rule-fields` issue as it could become something like `fwr-fields` presumably.

geemus: To be explicit/concrete, this would mean changing things like `firewall-fields` as a param to `fw-fields`.

I'm a bit on the fence about it, to be honest, but seems worth considering.

geemus (approved):
lgtm

lib/ubi_cli.rb
jeremyevans: I'm on the fence about it as well.  We can change this easily later if we need to.

fdr (approved):

PR #2848, opened 2025-02-19
Reduce VM create time

macieksarnowicz:
Remove the `wait_for_slice` label from the Vm Nexus and with that remove the wait time before calling the `prep`. The VM service on the host (systemctl service) has an implicit dependency on the slice by specifying the slice name in the service file. That ensures that the VM is not started until the slice is started and ready. That waiting can happen in `wait_for_sshable`, as we are waiting there anyway for the service to start and be connectable. This allows the slice creation to proceed fully parallel to the Vm creation.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -244,17 +244,6 @@ class Prog::Vm::Nexus < Prog::Base
 
     host.sshable.cmd(command)
 
-    hop_wait_for_slice
-  end
-
-  label def wait_for_slice
-    if vm.vm_host_slice
-      if !vm.vm_host_slice.enabled
-        # Just wait here until the slice creation is completed
-        nap 1
-      end
-    end
-
     hop_prep
   end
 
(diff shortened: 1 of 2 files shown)


enescakir:
The control plane side looks good to me; just a removal 😄. However, I don't have the context for the data plane side, so I'm unsure if it has any necessary dependencies etc.

macieksarnowicz:
> The control plane side looks good to me; just a removal 😄. However, I don't have the context for the data plane side, so I'm unsure if it has any necessary dependencies etc.

No changes are needed to the data plane. In `vm_setup.rb:636` and `vm_setup.rb:685` we already make the VM services dependent on the slice. Per the systemctl documentation, this will make sure that the services are started after the slice is started. Before this change, we are waiting for the slice to be fully running before we start the services creation (in `prep` label). After this change, the prep proceeds in parallel to the slice creation and the VM will wait in `wait_for_sshable` for everything to be running. 

I have run E2E tests on my own host with this change and the VMs and slices are being created correctly. I think Hadi has also started an E2E test run here. On my host, the non-slice VMs consistently take a longer time to create than in-slice VMs, so I do not have a way to say how much improvement this brings, but on paper it should save some time especially for the outlier create times.

pykello (approved):

PR #2849, opened 2025-02-19
Add cli lb commands

jeremyevans:
The following subcommands are supported:

* list
* create
* show
* destroy
* attach-vm
* detach-vm
* update

While testing the commands, I found that load balancer creation in the API failed when the health_check_endpoint param is not provided, and that using an invalid value for an enum resulted in an unhandled exception, so fixes for those are included.

This also includes some general CLI improvements:

* Ensure /cli response bodies end with a newline
* Avoid long option lines by showing supported fields on their own lines, wrapping to 80 characters.

CC @geemus


jeremyevans:
Since I was able to finish it today, I pushed another commit that adds a test suite for bin/ubi.

spec/routes/api/cli/golden-files/fw list invalid.txt
```diff
@@ -3,6 +3,7 @@
 Usage: ubi fw list [options]
 
 Options:
-    -f, --fields=fields              show specific fields (default: location,name,id)
+    -f, --fields=fields              show specific fields (comma separated)
     -l, --location=location          only show firewalls in given location
     -N, --no-headers                 do not show headers
+Fields: location name id
```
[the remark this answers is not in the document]
geemus: Also, it appears the options already appear in alpha order (but maybe this is coincidental and they are just in order defined as well). In any event, might be good to ensure they are similarly consistently ordered.

spec/routes/api/cli/golden-files/help -r.txt
```diff
@@ -77,6 +82,57 @@ Options:
...
 
 
+Usage: ubi lb subcommand [...]
+
+Subcommands: list
+
+
+Usage: ubi lb location/(lb-name|_lb-ubid) subcommand [...]
+
+Subcommands: attach-vm create destroy detach-vm show update
```
geemus, line 92: Similar here, subcommands should have an order (looks like alpha again).

spec/routes/api/cli/golden-files/lb eu-central-h1_test-lb update bad 1234 54321 _up2 vmdzyppz6j166jh5e9t2dwrfas.txt
```diff
@@ -0,0 +1,2 @@
+! Unexpected response status: 400
+Details: algorithm is not in range or set: ["round_robin", "hash_based"]
```
geemus, line 2: Relating to suggestions on the error above. When there is just one line, it seems fine to not have a newline, but when there were two I guess it felt somewhat different. I think with other things the pattern we leaned toward was newline even if there was only one line (ie for the nested serializations).

In a sense, if this is a one item unordered list, it should probably be newline dileneated if that is what we are doing elsewhere.

spec/routes/api/cli/golden-files/help -r.txt
jeremyevans: The alpha order here is enforced.

spec/routes/api/cli/golden-files/fw list invalid.txt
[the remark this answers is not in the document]
jeremyevans: I don't think alpha sorting this is a good idea.  If you alpha sort just the display of fields so it looks sorted, then it doesn't match the order shown with no option given (inconsistency). If you actually change the displayed order of fields, the output will look worse, since you'll be mixing nested outputs and non-nested items (for the `show` commands), and be inconsistent across commands for `list`.

The current order for `show` commands is consistent with the serialization used in the API.  The current order for `list` commands is generally consistent across commands (location (except for load balancer which doesn't expose this, yet), name, id, ...), and there are only 2-3 additional fields per `list` command. Having `location` followed by `name` is very useful, considering that almost all commands other than `list` take `location/name` as the reference (they also take `location/_ubid`, but I can't imagine anyone would prefer to use that).

If you have specific suggestions for default `list`/`show` field ordering for individual commands, please let me know.

spec/routes/api/cli/golden-files/lb eu-central-h1_test-lb update bad 1234 54321 _up2 vmdzyppz6j166jh5e9t2dwrfas.txt
jeremyevans: The part following `Details: ` is from the `error.message`, and is always a single line.  There could be zero or more lines after that, one for each entry in `error.details`.  Those provide details for the message, so if you were going to indent the message, I think the proper way would be this:

```
Details:
  Validation failed for following fields: vm_id, foo
    vm_id: VM is already attached to a load balancer
    foo: some other error message
```

I don't think the additional nesting improves the output, but if you do, please let me know and I'll make that change.

fdr (approved):

PR #2853, opened 2025-02-20
Resolve unavailability pages if resource is destroyed.

pykello:
Otherwise, an orphaned page will remain which won't auto-resolve.


enescakir (approved):

prog/vm/nexus.rb
```diff
@@ -447,6 +447,9 @@ def write_params_json
       hop_prevent_destroy
     end
 
+    # resolve outstanding pages, if any
+    Page.from_tag_parts("VmUnavailable", vm.ubid)&.incr_resolve
```
byucesoy, line 451: Maybe a better solution would be creating deadline pages (with deadline_target: ` wait`) instead of separate VmUnavailable page. Deadlines already get auto resolved in resource deletion. We also populate them few extra information which helps with on call.

pykello: @byucesoy does something like this work? https://github.com/ubicloud/ubicloud/pull/2855

byucesoy: Yes, it would.

PR #2855, opened 2025-02-20
Use deadline pages for VM/Slice unavailability

pykello:
Replaced `VmUnavailable` and `VmHostSliceUnavailable` pages with deadline pages. Deadlines automatically resolve on resource deletion and provide additional context useful for on-call debugging.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -409,11 +409,12 @@ class Prog::Vm::Nexus < Prog::Base
 
     begin
       if available?
-        Page.from_tag_parts("VmUnavailable", vm.ubid)&.incr_resolve
         decr_checkup
         hop_wait
       else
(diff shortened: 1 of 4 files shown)


prog/vm/nexus.rb
```diff
@@ -409,11 +409,12 @@ def write_params_json
...
     begin
       if available?
-        Page.from_tag_parts("VmUnavailable", vm.ubid)&.incr_resolve
         decr_checkup
         hop_wait
       else
-        Prog::PageNexus.assemble("#{vm} is unavailable", ["VmUnavailable", vm.ubid], vm.ubid)
+        # Use deadlines to create a page instead of a custom page, so page
+        # resolution in different states can be handled properly.
+        register_deadline("wait", 0, allow_extension: false)
```
byucesoy, line 417: No need for `allow_extension: false` as it is the default already.

byucesoy (approved):

PR #2857, opened 2025-02-21
Add a go version of bin/ubi

jeremyevans:
This builds on top of #2849, and adds a go version that is much faster than the ruby version and passes the same test suite (after minor test suite changes).

@0xjunhao including you as you mentioned you have some go experience, and it's been over a decade since I last used go. Also, a significant amount of the go code was written by Perplexity.ai.  Only the last commit ("Add a go version of bin/ubi") contains the go code.

### DIFF
--- a/cli/ubi.go
+++ b/cli/ubi.go
@@ -1,0 +1,215 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"slices"
+	"strings"
+)
+
+var allowConfirmation bool = true
+var debugEnabled = os.Getenv("UBI_DEBUG") == "1"
+
+func getToken() string {
+	token := os.Getenv("UBI_TOKEN")
+	if token == "" {
+		fmt.Fprintln(os.Stderr, "! Personal access token must be provided in UBI_TOKEN env variable for use")
+		os.Exit(1)
+	}
+	return token
+}
+
+func baseURL() string {
+	if url := os.Getenv("UBI_URL"); url != "" {
+		return url
+	}
+	return "https://api.ubicloud.com/cli"
+}
+
+type Client struct {
+	http.Client
+	Headers http.Header
+}
+
+func NewClient() *Client {
+	return &Client{
+		Headers: make(http.Header),
+	}
+}
+
+func sendRequest(args []string) {
+	requestBodyHash := make(map[string][]string)
+	requestBodyHash["argv"] = args
+	request_body, err := json.Marshal(requestBodyHash)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


enescakir:
I just changed the base to `jeremy-cli-lb` instead of `main` to make the review easier.

cli/ubi.go
```diff
@@ -0,0 +1,215 @@
...
+	"os/exec"
+	"slices"
+	"strings"
+)
+
+var allowConfirmation bool = true
+var debugEnabled = os.Getenv("UBI_DEBUG") == "1"
+
+func getToken() string {
+	token := os.Getenv("UBI_TOKEN")
```
enescakir, line 20: Should we add support for the `.env` file as well?

jeremyevans: `bin/ubi` loads `.env.rb` for easier configuration in multiple environments. `cli/ubi` is written in go and cannot load ruby code, so using `.env.rb` is definitely out.

I don't think looking for `.env` in the current directory makes sense.  Since `cli/ubi` is designed for production use, and will be runnable by clients from any directory, we have a couple options (we could do both):

1.  Parse `~/.ubi.conf` if available.
2. Make `ubi` accept `-i` as the first argument, specifying a file to load.

My idea with either of these files is parsing would be limited to setting environment variables. For example a file with the following content:

```
token=blah
ssh=putty
  
foo
```

Would set the UBI_TOKEN and UBI_SSH environment variables, skip the empty/blank line, and maybe warn about the `foo` line.

jeremyevans:
I switched the merge base back to main, as the jeremy-cli-lb branch was merged to main.

fdr (approved):
probably should add a "go fmt" step to this stanza in `ci.yml`:

```yaml
    - name: Run erb-formatter
      run: bundle exec rake linter:erb_formatter

    - name: Run rubocop
      run: bundle exec rake linter:rubocop

    - name: Run brakeman
      run: bundle exec rake linter:brakeman

    - name: Run linter:openapi
      run: bundle exec rake linter:openapi
```

cli/ubi.go
fdr: I think we should avoid some slicker way to pick up environment variables until we have a reason to include it. Also the "correct" location of configs is a minor litigation across the platforms  (e.g. on Linux, it's generally `~/.config` these days).

Some sensitive stuff, like the token, we are more likely to be seeking operating system keychain (or even ... in less orthodox fashion ... ssh-agent) integration.

jeremyevans:
> probably should add a "go fmt" step to this stanza in `ci.yml`:

I can do just that, but is that considered more important than program actually building?  Or do we want to run a go build as well, and check that the cli_spec runs with the built version?

It would be nice to avoid a go build/fmt if the go code hasn't changed, since I expect most commits won't change it.  I'm fairly sure there is a way to make it conditional (different workflow file?), but I haven't done it before, so I'd have to do some research.

PR #2858, opened 2025-02-21
Use deadline pages for VM host unavailability

enescakir:
PostgresServer was using deadlines instead of separate unavailability
pages. Hadi recently replaced them with deadline pages for both VM and
VM slice at 57f2d0f7.

The advantage of deadline pages is that they automatically resolve when
the resource is deleted and provide extra context useful for on-call
debugging. Otherwise, operators often forget to resolve them if the
resource is deleted.

### DIFF
--- a/prog/vm/host_nexus.rb
+++ b/prog/vm/host_nexus.rb
@@ -311,12 +311,12 @@ class Prog::Vm::HostNexus < Prog::Base
   end
 
   label def unavailable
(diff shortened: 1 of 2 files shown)


enescakir:
https://github.com/ubicloud/ubicloud/pull/2855

byucesoy (approved):

pykello (approved):

PR #2860, opened 2025-02-21
Make cli lb list command include location

jeremyevans:
This requires changes to the api/serializer so that location is included for load balancers by default.

I think the reason this wasn't included for load balancer but was included for private subnets is that load balancer does not have a location column, it gets the location through the private_subnet it is associated with.

CC @geemus


geemus (approved):
lgtm, I had noticed it as being somewhat anomalous. I think this will be nice in terms of consistency (and not requiring the user to know more about the domain model such that they could track it back to subnet).

fdr (approved):

PR #2863, opened 2025-02-21
Update rubocop and standard versions

enescakir:
```
Offenses:

views/account/multifactor/webauthn_remove.erb:51:25: C: [Correctable] Style/RedundantParentheses: Don't use parentheses around an assignment.
                  <% if (error = rodauth.field_error(rodauth.webauthn_remove_param)) %>
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
views/components/form/cloudflare_turnstile.erb:1:7: C: [Correctable] Style/RedundantParentheses: Don't use parentheses around an assignment.
<% if (@enable_cloudflare_turnstile = Config.cloudflare_turnstile_site_key) %>
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```


fdr (approved):
all about those new automatic rewrites.

jeremyevans:
Rubocop has previously required adding parentheses around assignments in if statements.  I'm good with removing that requirement, but I'm not sure why it's only making changes to the `views` and not the rest of the code.

enescakir:
> Rubocop has previously required adding parentheses around assignments in if statements. I'm good with removing that requirement, but I'm not sure why it's only making changes to the `views` and not the rest of the code.

In this example, `Style/RedundantParentheses` is failing and removes parentheses in views. However, as you mentioned, `Lint/AssignmentInCondition` keeps parentheses around assignments in non-`erb` files. It's interesting how these cops conflict and cause different behaviors.

enescakir:
I added 4 different statements in an `.erb` file and checked them with rubocop. It accepts 1, 2, and 3, but converts 4 to 3. It doesn't seem to work consistently for `.erb` files.

```ruby
#1
<%
  if (e = ENV)
    puts e
  end
%>

#2
<%
  if e = ENV
    puts e
  end
%>

#3
<% if e = ENV %>
<% end %>

#4
<% if (e = ENV) %>
<% end %>
```

enescakir:
I'm not disabling it for now and will keep as in the rubocop format. However, we can investigate further if we find any bugs in rubocop

PR #2864, opened 2025-02-21
Add cli/api support for a few previously web-only actions

jeremyevans:
This adds the following cli commands:

* vm restart
* pg restart
* ps connect
* ps disconnect

All of these were previously web-only, and the cli can only implement what the api exposes, so this adds api support for all of these.

This also changes usage lines to say `id` instead of `ubid`, as ubids are internal things that users consider to be ids (we don't generally expose the uuid directly to users).

CC @geemus

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -832,21 +832,6 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
-  '/project/{project_id}/location/{location}/postgres/{postgres_database_name}/restart':
-    parameters:
-      - $ref: '#/components/parameters/project_id'
-      - $ref: '#/components/parameters/location'
-      - $ref: '#/components/parameters/postgres_database_name'
-    post:
-      operationId: restartPostgresDatabase
-      summary: Restart a specific Postgres Database
-      responses:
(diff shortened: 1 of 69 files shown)


openapi/openapi.yml
```diff
@@ -1087,6 +1102,47 @@ paths:
...
+                  type: string
+              additionalProperties: false
+      responses:
+        '200':
+          $ref: '#/components/responses/PrivateSubnet'
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Private Subnet
+  '/project/{project_id}/location/{location}/private-subnet/{private_subnet_name}/disconnect/{private_subnet_id}':
```
geemus, line 1130: For consistency, I think maybe the private_subnet_id should have a _ prefix?

geemus (approved):
lgtm

openapi/openapi.yml
jeremyevans: I think that would actually break it, since the route doesn't accept the underscore.  The underscore is not necessary for disambiguation as you cannot use a name in this case, only a ubid.

fdr (approved):

PR #2865, opened 2025-02-21
Do not added create_account method to all objects when running route specs

jeremyevans:
This makes the create_account method only usable inside the spec methods.


bsatzger:
"Do not added create_account method to all objects when running route specs" -> "Do not add create_account method to all objects when running route specs"?

bsatzger (approved):

PR #2868, opened 2025-02-21
Fix usage for cli lb show command

jeremyevans:
This was missed in the ubid -> id fix in
9f4280f72267ad6f7699a3408cfbff1954b1f451.

CC @geemus


jeremyevans:
This fix is included as part of #2875.

PR #2875, opened 2025-02-24
Do not allow model ubid as model name, remove underscore requirement, allow ubid without location in cli

jeremyevans:
This is a series of commits that:

* Disallows setting a name for a model that matches the model's ubid format
* Removes the leading underscore requirement for ubids in route segments (no longer needed for disambiguation, underscore allowed for backwards compatibility)
* Adds object-info api endpoint that accepts a ubid and returns type/location/name information for it
* Updates the cli to accept ubid without location, using the object-info api endpoint to get the location

I'm currently basing this on top of #2873 to avoid conflicts, but if that isn't accepted, I can adjust this.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1241,6 +1241,20 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Virtual Machine
+  '/project/{project_id}/object-info/{object_id}':
+    parameters:
+      - $ref: '#/components/parameters/project_id'
+      - $ref: '#/components/parameters/object_id'
+    get:
+      operationId: getObjectInfo
+      summary: 'Return information on object type, location, and name'
+      responses:
+        '200':
+          $ref: '#/components/responses/ObjectInfo'
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Project
   '/project/{project_id}/postgres':
     get:
       operationId: listPostgresDatabases
@@ -1331,6 +1345,15 @@ components:
         type: string
         example: et7ekmf54nae5nya9s6vebg43f
         pattern: '^(et|md)[0-9a-hj-km-np-tv-z]{24}$'
+    object_id:
+      description: ID of a supported object
+      in: path
+      name: object_id
+      required: true
+      schema:
+        type: string
+        example: fwkkmx0f2vke4h36nk9cm8v8q0
+        pattern: '^(fw|1b|pg|ps|vm)[0-9a-hj-km-np-tv-z]{24}$'
     order_column:
       description: Pagination - Order column
       in: query
@@ -1838,6 +1861,27 @@ components:
             required:
               - count
               - items
+    ObjectInfo:
+      description: 'Type, location, and name information for object'
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              type:
+                description: Type of object
+                type: string
+              location:
+                description: Location of object
+                type: string
+              name:
+                description: Name of object
+                type: string
+            additionalProperties: false
+            required:
+              - type
+              - location
+              - name
     PostgresDatabase:
       description: A Postgres Database
       content:
--- a/spec/model/vm_spec.rb
+++ b/spec/model/vm_spec.rb
@@ -246,6 +246,18 @@ RSpec.describe Vm do
     vm.init_health_monitor_session
   end
 
+  it "disallows VM ubid format as name" do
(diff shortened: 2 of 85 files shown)


openapi/openapi.yml
geemus, line 1: Should we drop the _ version of paths from the API as well? It could certainly simplify the schema to do so (and then just have a single path that can take id or name, I think all of the ones in question are already nested under a location, so it would allow for location/name or location/id already).

Taking it even further, I suppose id only (sans location) version things of API actions could be added also.

spec/routes/api/project/location/firewall_spec.rb
geemus, line 1: Are there other specs where similar changes might be desired? I just ask as testing these cases seems worth doing, while doing it specifically and only on firewalls in seems a bit arbitrary. I guess I would expect either that it would be in a more generic spec (where some particular resource would be used, but from context it would be clear that it was generic) or that this would appear more consistently across different resources.

Note: This is probably fine as is, to be clear, just thought I would share that train of thought in case it helped.

geemus (approved):

openapi/openapi.yml
jeremyevans: I think it's a good idea to drop the `_` paths from openapi.yml (is that something you could handle after this?).

Having id only API actions requires significant routing tree changes to add such support.  I'm not sure the benefit is worth the additional complexity.

spec/model/vm_spec.rb
```diff
@@ -246,6 +246,18 @@
     vm.init_health_monitor_session
   end
 
+  it "disallows VM ubid format as name" do
+    vm = described_class.new(name: described_class.generate_ubid.to_s)
+    vm.validate
+    expect(vm.errors[:name]).to eq ["cannot be exactly 26 numbers/lowercase characters starting with vm to avoid overlap with id format"]
+  end
+
+  it "allows postgres server ubid format as name" do
```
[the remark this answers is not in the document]
jeremyevans: Originally, the plan was to disallow as names anything looking like a ubid, regardless of type.  That broke a lot, since we use ubids from one model as names for another model, so I started added workarounds for those cases.  After a few rounds of that, I decided it wasn't a feasible approach, so I decided to only disallow ubids for the specific model.  I had already written these tests, and they passed, so I kept them.

openapi/openapi.yml
geemus: Sounds good. Yes, I'd be happy to work on those updates once this lands.

Yeah, I think there may end up being cases where id-only would be worth it, but it is not apparent to me at this time that it would be worth the trouble.

spec/routes/api/project/location/firewall_spec.rb
geemus: Ah, got it. I'm not too bothered by it, just happened to notice and thought I would ask in case you felt like there was anything that should be done about it.

fdr (approved):
I was mulling over this, although it's an option we can exercise at a later point, we can perhaps exploit the high entropy of of system generated ubids to tweak this into a double-probing model (first check `id` if it's a matching string, then `name`) if we get into trouble with the statistical nature of allowed/disallowed strings from the first two bytes, or by our preference upon consideration. I think we floated such an idea near the beginning, but thought there were less ambiguous alternatives available at a low cost to the API, before making contact with the problem of rapid work with the CLI.

This introduces a heinous kind of shadowing, but given the entropic nature of system generated identifiers, you'd have to try to alias the string on purpose. I don't worry much at all about customers doing this on purpose because of the consciousness required to do so, the main area I thought I was trying to avoid was attacking a system by creative use of shadowing, hence the `_` mode of address sigil. That was intended to be visible, to even make typo-squats (e.g. a ubid minus a single character) visible.

We gave all that (for small definitions of "all that") up because, at the time it was being evaluated, it was in the context of the API alone, i.e. computer programs where sigils are not that big a deal, rather than hacking around the CLI and perhaps by extension the web site. But, after @jeremyevans was writing the CLI, the distributed cost of pushing this computer-program optimized scheme into something intended for people to read in interactive setting seems unappealingly high.

Thinking over the entire evolution, the `_` sigil could have been maintained in some contexts but not others, but then the user has to learn two sets of rules and when they are applied and what they're optimized for (quick work versus robustness against typo-squatting-like confused identity). That counter-factual seems like a cure worse than the disease. So I think what we're winding up with is one more complex rule (alphanumeric, except a certain length and a pair of bytes that changes for every object in the API)  that has a rather satisfying property of allowing quick navigation by copy and paste without postprocessing, and is still tolerable in every other respect.

PR #2879, opened 2025-02-24
Add CI workflow for the cli/ubi go program

jeremyevans:
The `cli/ubi.go` program should not see changes very often, so no need to run the workflow all the time.  The workflow checks three things:

* `go fmt` doesn't make changes (linting)
* `go build` builds successfully
* related cli specs pass when using built `cli/ubi`

The last commit makes it only run if one of the related files is changed.  I split the commit to get a CI run with the workflow (since this PR does not make changes to the related files). That CI run passed: https://github.com/ubicloud/ubicloud/actions/runs/13510171025)

CC @geemus

### DIFF
--- a/.github/workflows/cli-ci.yml
+++ b/.github/workflows/cli-ci.yml
@@ -54,9 +54,6 @@ jobs:
         CLOVER_DATABASE_URL: postgres://${{ env.DB_USER }}:${{ env.DB_PASSWORD }}@localhost:5432/${{ env.DB_NAME }}
       run: bundle exec rake _test_up
 
-    - name: "Run go fmt and check it makes no changes"
-      run: "cd cli && go fmt && git diff --stat --exit-code"
-
     - name: Build cli/ubi
       run: bundle exec rake ubi
 
@@ -67,3 +64,12 @@ jobs:
         CLOVER_COLUMN_ENCRYPTION_KEY: TtlY0+hd4lvedPkNbu5qsj5H7giPKJSRX9KDBrvid7c=
         UBI_CMD: ./cli/ubi
       run: bundle exec rspec spec/cli_spec.rb
+
+    - name: Run go fmt and check it makes no changes
+      run: "cd cli && go fmt && git diff --stat --exit-code"
+
+    - name: Run golangci-lint
+      uses: golangci/golangci-lint-action@v6
+      with:
+        version: v1.64
+        working-directory: cli
--- a/.github/workflows/cli-ci.yml
+++ b/.github/workflows/cli-ci.yml
@@ -1,0 +1,69 @@
+name: cli/ubi CI
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  cli-ci:
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ubicloud, ubicloud-arm]
+    name: cli/ubi CI - ${{matrix.runs-on}}
+    runs-on: ${{matrix.runs-on}}
+
+    env:
+      DB_USER: clover
+      DB_PASSWORD: nonempty
+      DB_NAME: clover_test
+
+    services:
+      postgres:
+        image: postgres:15.4
+        env:
+          POSTGRES_USER: ${{ env.DB_USER }}
+          POSTGRES_PASSWORD: ${{ env.DB_PASSWORD }}
+          POSTGRES_DB: ${{ env.DB_NAME }}
+        ports:
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
+    steps:
+    - name: Perform superuser-only actions, then remove superuser
+      run: |
+        psql "postgres://${{ env.DB_USER }}:${{ env.DB_PASSWORD }}@localhost:5432/${{ env.DB_NAME }}" \
+          -c "CREATE EXTENSION citext; CREATE EXTENSION btree_gist; CREATE ROLE clover_password PASSWORD '${{ env.DB_PASSWORD }}' LOGIN; ALTER ROLE ${{ env.DB_USER }} NOSUPERUSER"
+
+    - name: Check out code
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/cli-ci.yml
```diff
@@ -0,0 +1,89 @@
...
+        if [ ! -f $RUNNER_TOOL_CACHE/Ruby/$RUBY_VERSION/arm64.complete ]; then
+          git clone https://github.com/rbenv/ruby-build.git
+          sudo ./ruby-build/install.sh
+          sudo ruby-build "$RUBY_VERSION" $RUNNER_TOOL_CACHE/Ruby/$RUBY_VERSION/arm64
+          sudo chown -R runner:runner $RUNNER_TOOL_CACHE/Ruby
+          touch $RUNNER_TOOL_CACHE/Ruby/$RUBY_VERSION/arm64.complete
+          rm -rf ruby-build
+        fi
+
+    - name: Set up Ruby
```
geemus, line 66: Similar to this, you will probably want to utilize the `setup-go` action so you can ensure the expected go version is being used, rather than whatever happens to be on the image.

.github/workflows/cli-ci.yml
```diff
@@ -0,0 +1,89 @@
...
+      with:
+        ruby-version: .tool-versions
+        bundler-cache: true
+
+    - name: Apply migrations
+      env:
+        CLOVER_DATABASE_URL: postgres://${{ env.DB_USER }}:${{ env.DB_PASSWORD }}@localhost:5432/${{ env.DB_NAME }}
+      run: bundle exec rake _test_up
+
+    - name: "Run go fmt and check it makes no changes"
```
geemus, line 77: In addition to go fmt, you might consider adding some more linting (ie akin to rubocop or something). We use this: https://github.com/golangci/golangci-lint

geemus (approved):
Huh, I was unaware of the paths directive. We do something similar, but with a different action to control the conditional file check.

Several steps here look duplicative to the non-cli ci workflow (ie like the arm-related ruby setup). You could consider using composite actions to DRY some of that up (though they can have their own challenges).

.github/workflows/cli-ci.yml
jeremyevans: I'm guessing this works now because I'm using the distro-provided go package, and the runner is running the same OS.  I'll work on this.

.github/workflows/cli-ci.yml
jeremyevans: I'll look into this.

fdr (approved):

.github/workflows/cli-ci.yml
```diff
@@ -0,0 +1,89 @@
+name: cli/ubi CI
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'cli/go.mod'
+      - 'cli/ubi.go'
```
enescakir, line 9: You can use a pattern instead of adding each file one by one.
```suggestion
      - 'cli/**'
```

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

.github/workflows/cli-ci.yml
```diff
@@ -0,0 +1,89 @@
...
+    steps:
+    - name: Perform superuser-only actions, then remove superuser
+      run: |
+        psql "postgres://${{ env.DB_USER }}:${{ env.DB_PASSWORD }}@localhost:5432/${{ env.DB_NAME }}" \
+          -c "CREATE EXTENSION citext; CREATE EXTENSION btree_gist; CREATE ROLE clover_password PASSWORD '${{ env.DB_PASSWORD }}' LOGIN; ALTER ROLE ${{ env.DB_USER }} NOSUPERUSER"
+
+    - name: Check out code
+      uses: actions/checkout@v4
+
+    - name: Install ruby for ARM runners if not cached
```
enescakir, line 53: Actually, `ruby/setup-ruby` has added prebuilt artifacts for arm64 as well recently.
https://github.com/ruby/setup-ruby/issues/689#issuecomment-2600265901

We probably don't need it anymore, so we might remove it from our Clover CI too.

.github/workflows/cli-ci.yml
jeremyevans: Updated cli-ci.yml to use setup-go with a matching version.

.github/workflows/cli-ci.yml
jeremyevans: Added a Rake task to do this locally, added it to cli-ci.yml, and fixed all identified issues.

.github/workflows/cli-ci.yml
jeremyevans: I removed it from cli-ci.yml, and CI still passed.  I'll submit a PR for ci.yml in a bit.

.github/workflows/cli-ci.yml
```diff
@@ -0,0 +1,69 @@
...
+        psql "postgres://${{ env.DB_USER }}:${{ env.DB_PASSWORD }}@localhost:5432/${{ env.DB_NAME }}" \
+          -c "CREATE EXTENSION citext; CREATE EXTENSION btree_gist; CREATE ROLE clover_password PASSWORD '${{ env.DB_PASSWORD }}' LOGIN; ALTER ROLE ${{ env.DB_USER }} NOSUPERUSER"
+
+    - name: Check out code
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: '1.22.2'
```
geemus, line 54: You may, at least eventually, want to have this key off the go.mod files (and then manage the version via asdf in .tool-versions). I think that level of indirection is your best bet (rather than referencing .tool-versions directly), but I'm not sure. It appears to be what we do. Here is the relevant readme bit: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

I suspect the DEVELOPERS.md file might be ready for some love to describe some of the new setup and stuff at some point also.

jeremyevans: OK, I switched from `go-version` to `go-version-file`.

I can work on DEVELOPERS.md changes in the near future, when I work on general documentation of the cli.

PR #2887, opened 2025-02-25
Fix NoMethodError when viewing VM create page with burstables and location_latitude_fra feature flags enabled

jeremyevans:
This issue occurred because there is no billing rate for a burstable size for the latitude_fra location.

Add a BillingRate.unit_price_from_resource_properties that returns nil instead of raising a NoMethodError if there is no matching resource property.  Use it to DRY up code in ContentGenerator::{Vm,Postgres}.

Make ContentGenerator::Vm.family not return an array if there are no valid sizes at the requested location for the family.

This issue affects VMs and not PostgreSQL databases because Option.postgres_locations does not include the latitude_fra location.

### DIFF
--- a/lib/content_generator.rb
+++ b/lib/content_generator.rb
@@ -12,13 +12,17 @@ module ContentGenerator
 
     def self.enable_ipv4(location, value)
       location = LocationNameConverter.to_internal_name(location)
-      unit_price = BillingRate.from_resource_properties("IPAddress", "IPv4", location)["unit_price"].to_f
+      unit_price = BillingRate.unit_price_from_resource_properties("IPAddress", "IPv4", location)
 
       "Enable Public IPv4 ($#{"%.2f" % (unit_price * 60 * 672)}/mo)"
     end
 
     def self.family(location, family)
       vm_family = Option::VmFamilies.find { _1.name == family }
+
+      # Do not include a family for a location that has no valid sizes at the location
+      return unless BillingRate.unit_price_from_resource_properties("VmVCpu", family, LocationNameConverter.to_internal_name(location))
+
       [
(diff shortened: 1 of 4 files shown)


lib/content_generator.rb
```diff
@@ -12,13 +12,17 @@ def self.private_subnet(location, private_subnet)
...
+      unit_price = BillingRate.unit_price_from_resource_properties("IPAddress", "IPv4", location)
 
       "Enable Public IPv4 ($#{"%.2f" % (unit_price * 60 * 672)}/mo)"
     end
 
     def self.family(location, family)
       vm_family = Option::VmFamilies.find { _1.name == family }
+
+      # Do not include a family for a location that has no valid sizes at the location
+      return unless BillingRate.unit_price_from_resource_properties("VmVCpu", family, LocationNameConverter.to_internal_name(location))
```
macieksarnowicz, line 24: The filtering of options is done in helpers/vm.rb. How about fixing it like this:

```
    options.add_option(name: "family", values: Option.families(use_slices: @project.get_ff_use_slices_for_allocation || false).map { _1.name }, parent: "location") do |location, family|
      !BillingRate.from_resource_properties("VmVCpu", family, LocationNameConverter.to_internal_name(location)).nil?
    end

```

macieksarnowicz: That is in helpers/vm.rb:100

byucesoy: Agreed. The purpose of ContentGenerator is just deciding how to display given data. Actual valid options are decided in /helpers/*.rb.

jeremyevans:
Closing as @macieksarnowicz is addressing in #2903.

PR #2891, opened 2025-02-26
Fix initialization issue when populating StorageDevices and increase error verbosity

mohi-kalantari:
During the write/read/delete check, we faced many failures in Prod and couldn't dig more to see what the issue is. Verbosity is increased to investigate more.

Also the /dev/ prefix is deleted from the StorageDevice creation logic which caused issues when a new VmHost was added.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -316,14 +316,25 @@ class VmHost < Sequel::Model
     end
 
     all_mount_points.uniq.all? do |mount_point|
-      file_name = (mount_point == "/") ? "/test-file" : Shellwords.escape("#{mount_point}/test-file")
-      write_status = ssh_session.exec!("sudo bash -c \"head -c 1M </dev/zero > #{file_name}\"").exitstatus == 0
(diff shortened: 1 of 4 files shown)


model/vm_host.rb
```diff
@@ -317,13 +317,24 @@ def check_storage_read_write(ssh_session, devices)
 
     all_mount_points.uniq.all? do |mount_point|
       file_name = (mount_point == "/") ? "/test-file" : Shellwords.escape("#{mount_point}/test-file")
```
aamederen, line 319: can mount_point end with "/"? Maybe something like `File.join(mount_point, 'test-file')` can be used and help eliminating the ? condition

aamederen (approved):

model/vm_host.rb
mohi-kalantari: that ? was not primarily there for the joining of the paths. some of our disks are mounted in /var/storage/disk/disk-name and I need to provide the right prefix for test file in such cases.

mohi-kalantari: fixed it. thanks

PR #2900, opened 2025-02-27
Store device id instead of device name for StorageDevices

mohi-kalantari:
Right now StorageDevices store the name of devices which on reboot can change. In order to fix that, we will use udev rules and find the device id in /dev/disk/by-id.

For SSD disks, there is a unique identifier with wwn- prefix and for NVMe disks, there is a unique Identifier with the nvme-eui prefix.

We will store these values in the database and on healthcheck queries find the right device name and proceed with the normal health check routine.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -353,16 +353,25 @@ class VmHost < Sequel::Model
     }
   end
 
-  def disk_devices
-    storage_devices.each { |sd| sd.set_underlying_unix_devices if sd.unix_device_list.nil? || sd.unix_device_list.empty? }
+  def disk_device_ids
+    # we use this next line to migrate data from the old formatting (storing device names) to the new (storing id) so we trigger the convert
+    # whenever an element inside unix_device_list is not a SSD or NVMe id.
+    # SSD and NVMe ids start with wwn or nvme-eui respectively.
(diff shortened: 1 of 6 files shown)


model/vm_host.rb
```diff
@@ -353,16 +353,25 @@ def init_health_monitor_session
     }
   end
 
-  def disk_devices
-    storage_devices.each { |sd| sd.set_underlying_unix_devices if sd.unix_device_list.nil? || sd.unix_device_list.empty? }
+  def disk_device_ids
+    # we use this next line to migrate data from the old formatting (storing device names) to the new (storing id) so we trigger the convert
+    # whenever an element inside unix_device_list is not a SSD or NVMe id.
+    # SSD and NVMe ids start with wwn or nvme-eui respectively.
+    # This next line can be removed in the future after the first run of the code.
```
pykello, line 360: Maybe add a YYY prefix so we can grep this as a thing to do. We have used this prefix in few other places for the same purpose.

```suggestion
    # YYY: This next line can be removed in the future after the first run of the code.
```

pykello (approved):

model/vm_host.rb
mohi-kalantari: done thanks.

PR #2901, opened 2025-02-27
Move all the location references to id based foreign keys

furkansahin:
## Make location references a foreign key, create indexes - Mig 1
This commit adds the location foreign_keys to all of the models that
uses location. This way, we will not be matching the text values but
make use of the database relationship.
We are refilling the location_ids based on the location column of the
entities.

We are also adding a new index on project_id, location_id, name because we
already had an index on project_id, location, name and it will be
dropped once the location column is dropped.

## Update location references with foreign key in the controlplane
We are basically updating the location references everywhere with a
location id and adding the location relationship to the models to be
able to fetch location names when needed.
This also makes the LocationNameConverter model obsolete, so we are
removing it.

## Drop location column and make location_id not null - Mig 2


migrate/20250227_id_based_locations.rb
```diff
@@ -0,0 +1,129 @@
...
+    run "UPDATE vm_pool SET location_id = (SELECT id FROM location WHERE name = vm_pool.location);"
+    run "UPDATE private_subnet SET location_id = (SELECT id FROM location WHERE name = private_subnet.location);"
+    run "UPDATE firewall SET location_id = (SELECT id FROM location WHERE name = firewall.location);"
+    run "UPDATE postgres_resource SET location_id = (SELECT id FROM location WHERE name = postgres_resource.location);"
+    run "UPDATE minio_cluster SET location_id = (SELECT id FROM location WHERE name = minio_cluster.location);"
+    run "UPDATE inference_endpoint SET location_id = (SELECT id FROM location WHERE name = inference_endpoint.location);"
+    run "UPDATE kubernetes_cluster SET location_id = (SELECT id FROM location WHERE name = kubernetes_cluster.location);"
+
+    alter_table(:vm_host) do
+      drop_column :location
```
jeremyevans, line 52: It's too dangerous to drop the location columns in the same commit, because the migration will be run before the code changes that use it are submitted.  This needs to be staged, with the location column dropped in a later migration.  You'll need to make sure that the tests pass at every stage:

* migration that adds location_id
* Commit that switches from using location to using location_id
* migration that drops location

You probably want to mark the `location` columns as allowing `NULL` values with `set_column_allow_null`, so that later commit can stop setting them, and a following migration can drop them.

migrate/20250227_id_based_locations.rb
```diff
@@ -0,0 +1,129 @@
...
+    run "UPDATE vm_host SET location_id = (SELECT id FROM location WHERE name = vm_host.location);"
+    run "UPDATE vm SET location_id = (SELECT id FROM location WHERE name = vm.location);"
+    run "UPDATE vm_pool SET location_id = (SELECT id FROM location WHERE name = vm_pool.location);"
+    run "UPDATE private_subnet SET location_id = (SELECT id FROM location WHERE name = private_subnet.location);"
+    run "UPDATE firewall SET location_id = (SELECT id FROM location WHERE name = firewall.location);"
+    run "UPDATE postgres_resource SET location_id = (SELECT id FROM location WHERE name = postgres_resource.location);"
+    run "UPDATE minio_cluster SET location_id = (SELECT id FROM location WHERE name = minio_cluster.location);"
+    run "UPDATE inference_endpoint SET location_id = (SELECT id FROM location WHERE name = inference_endpoint.location);"
+    run "UPDATE kubernetes_cluster SET location_id = (SELECT id FROM location WHERE name = kubernetes_cluster.location);"
+
```
jeremyevans, line 50: I would create the necessary location_id indexes here.

DEVELOPERS.md
```diff
@@ -505,7 +505,7 @@ We show cloudifying a host from Hetzner, but the principles should work everywhe
     VM_HOST_ID = ""
     default_boot_images = ["ubuntu-noble", "ubuntu-jammy", "debian-12", "almalinux-9"]
 
-    st = Prog::Vm::HostNexus.assemble(VM_HOST_IP, provider_name: "hetzner", location: "hetzner-fsn1", server_identifier: VM_HOST_ID, default_boot_images: default_boot_images)
+    st = Prog::Vm::HostNexus.assemble(VM_HOST_IP, provider_name: "hetzner", location_id: "caa7a807-36c5-8420-a75c-f906839dad71", server_identifier: VM_HOST_ID, default_boot_images: default_boot_images)
```
jeremyevans, line 508: For the most common locations, I recommend we store the uuids in constants in the Location model, and then reference the constant.

bin/regen-screenshots
```diff
@@ -236,7 +236,7 @@ class RegenScreenshots
       net4: "10.0.0.0/8",
       state: "available",
       name: "blog-app1-subnet",
-      location: "hetzner-fsn1"
+      location_id: Location[name: "hetzner-fsn1"]
```
jeremyevans, line 239:
```suggestion
      location_id: Location[name: "hetzner-fsn1"].id
```

Or just reference a constant.

bin/regen-screenshots
```diff
@@ -260,7 +260,7 @@ class RegenScreenshots
         name: "blog-app#{i + 1}",
         unix_user: "a",
         public_key: "a",
-        location: "hetzner-fsn1",
+        location_id: "caa7a807-36c5-8420-a75c-f906839dad71",
```
jeremyevans, line 263: We should probably be consistent and use the same approach as earlier in this script.

cache/schema.cache
jeremyevans, line 1: Each migration commit should have the cache files committed with it.

lib/content_generator.rb
```diff
@@ -25,10 +25,10 @@ def self.family(location, family)
       ]
     end
 
-    def self.size(location, family, size)
-      location = LocationNameConverter.to_internal_name(location)
+    def self.size(location_id, family, size)
+      location = Location[location_id]
```
jeremyevans, line 29: This type will lead to N+1 issues, since you'll have a query for every size.  You would need to refactor this so you could pass in a hash of valid locations and look up the locations in the hash, so it does not cause a query per method call.  Other calls to `Location[]` and similar should be audited to make sure they do not cause N+1 query issues.

model/ai/inference_endpoint.rb
```diff
@@ -18,7 +19,7 @@ class InferenceEndpoint < Sequel::Model
   semaphore :destroy, :maintenance
 
   def display_location
-    LocationNameConverter.to_display_name(location)
+    location.display_name
```
jeremyevans, line 22: This does a query per call if the association is not already loaded. Similarly to the `Location[]` issue, you need to audit all calls to make sure this doesn't cause N+1 query issues in cases where you are retrieving multiple objects and calling the `location` method on each.  You can generally use `eager(:location)` to fix such N+1 issues.

Similar for all models where you are adding a `location` association and defining a similar method.

model/location.rb
```diff
@@ -18,3 +18,11 @@ class Location < Sequel::Model
 #  location_pkey | PRIMARY KEY btree (id)
 # Foreign key constraints:
 #  location_provider_fkey | (provider) REFERENCES provider(name)
+# Referenced By:
```
jeremyevans, line 21: Model annotation update should be included with the related migration commit.  A PR with just the migration commit should be able to pass CI.

prog/ai/inference_endpoint_nexus.rb
```diff
@@ -55,9 +55,9 @@ def self.assemble(project_id:, location:, boot_image:, name:, vm_size:, storage_
       ubid = InferenceEndpoint.generate_ubid
       internal_project = Project[Config.inference_endpoint_service_project_id]
       fail "No project configured for inference endpoints" unless internal_project
-      firewall = internal_project.firewalls_dataset.where(location: location).where(Sequel[:firewall][:name] => "inference-endpoint-firewall").first
-      fail "No firewall named 'inference-endpoint-firewall' configured for inference endpoints in #{location}" unless firewall
-      subnet_s = Prog::Vnet::SubnetNexus.assemble(internal_project.id, name: ubid.to_s, location: location, firewall_id: firewall.id)
+      firewall = internal_project.firewalls_dataset.where(location_id: location_id).where(Sequel[:firewall][:name] => "inference-endpoint-firewall").first
```
jeremyevans, line 58:
```suggestion
      firewall = internal_project.firewalls_dataset.where(location_id:).where(Sequel[:firewall][:name] => "inference-endpoint-firewall").first
```

Could omit hash values here and elsewhere.

prog/vm/vm_pool.rb
```diff
@@ -61,7 +61,7 @@ def before_run
 
   label def wait
     if vm_pool.size - vm_pool.vms.count > 0
-      idle_cpus = VmHost.where(allocation_state: "accepting", arch: vm_pool.arch, location: ["github-runners", "hetzner-hel1", "hetzner-fsn1"]).select_map { sum((total_cores - used_cores) * total_cpus / total_cores) }.first.to_i
+      idle_cpus = VmHost.where(allocation_state: "accepting", arch: vm_pool.arch, location_id: Location.where(name: ["github-runners", "hetzner-hel1", "hetzner-fsn1"]).select_map(:id)).select_map { sum((total_cores - used_cores) * total_cpus / total_cores) }.first.to_i
```
jeremyevans, line 64:
```suggestion
      idle_cpus = VmHost.where(allocation_state: "accepting", arch: vm_pool.arch, location_id: Location.where(name: ["github-runners", "hetzner-hel1", "hetzner-fsn1"]).select(:id)).select_map { sum((total_cores - used_cores) * total_cpus / total_cores) }.first.to_i
```

Better to use a subquery here than a separate query to get 3 ids.  It would be even better to use three constant references and avoid the need for a separate query or a subquery.

prog/vnet/subnet_nexus.rb
```diff
@@ -172,7 +175,7 @@ def self.random_private_ipv6(location, project)
     network_mask = NetAddr::Mask128.new(64)
     selected_addr = NetAddr::IPv6Net.new(network_address, network_mask)
 
-    selected_addr = random_private_ipv6(location, project) if project.private_subnets_dataset[Sequel[:net6] => selected_addr.to_s, :location => location]
+    selected_addr = random_private_ipv6(location, project) if project.private_subnets_dataset[Sequel[:net6] => selected_addr.to_s, :location_id => location.id]
```
jeremyevans, line 178: Realize the existing code was this way, but there's not a reason to use `Sequel[:net6]` instead of `:net6` here:

```suggestion
    selected_addr = random_private_ipv6(location, project) if project.private_subnets_dataset[net6:  selected_addr.to_s, location_id: location.id]
```

Similar for the net4 case below.

routes/project/location.rb
```diff
@@ -3,7 +3,7 @@
 class Clover
   hash_branch(:project_prefix, "location") do |r|
     r.on String do |location_display_name|
-      @location = LocationNameConverter.to_internal_name(location_display_name)
+      @location = Location[display_name: location_display_name]
```
jeremyevans, line 6:
```suggestion
      next unless (@location = Location[display_name: location_display_name])
```

Looks like this was an issue previously, but I assume we don't want to continue processing the request if an invalid location is used.

Note that once you support project-specific locations, you will need to make sure to only allow a location to be selected if it is for all projects or for this specific project.

Other routes should be checked to see we should also block invalid locations of those.

jeremyevans (commented):
I agree in concept with switching to location id references.

migrate/20250227_id_based_locations.rb
furkansahin: Updated accordingly

lib/content_generator.rb
furkansahin: I have solved this by passing the whole location object until here and rendering the id in `resource_creation_form`. What do you think now?

migrate/20250227_id_based_locations.rb
furkansahin: I am planning to run them separately. The main reason is that, I would like to run this file in a transaction. Creation of the indexes will be handled as the very last step after controlplane is also migrated. Do you think that's OK?

lib/content_generator.rb
furkansahin: There is something wrong with this approach and rendering that I cannot seem to solve though. The moment I pick a non default region in the UI, the private subnet list becomes disabled all together. I will try fixing it on Monday but if it persists, I might simply defer this to another PR to reduce the query count. For now, the safest option seems like passing the ID and rendering accordingly.

migrate/20250227_id_based_locations.rb
```diff
@@ -0,0 +1,129 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
```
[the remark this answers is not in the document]
jeremyevans: I didn't know that was allowed.  Doesn't that mean we wouldn't provision new github runners (and other things like VMs/databases) while it is down?

It seems like a risky approach if something goes wrong.  A phased approach that doesn't require stopping respirate seems safer, though it likely requires more engineering work.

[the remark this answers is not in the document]
fdr: I think it would be alright to whack ps:scale around a couple of times, that's pretty low risk, to get the advantage of atomicity .Whether the migration succeeds or fails, the important part is to be timely for the benefit of github actions.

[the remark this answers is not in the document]
fdr: If we were to try to mitigate some kind of risk, I think it's thinking about the revert path carefully, e.g. if we want to go back to config files right away without reverse-migrating, because we busted something

migrate/20250226_project_id_location_indexes_drop.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
+
+  up do
+    %i[postgres_resource firewall private_subnet vm minio_cluster kubernetes_cluster].reverse_each do |table|
+      drop_index table, [:project_id, :location, :name], name: :"#{table}_project_id_location_name_uidx", unique: true, concurrently: true
```
fdr, line 8: note this also drops a constraint, right? that is, the uniqueness constraint? so it could break some concurrent algorithms, if there are any.

fdr (commented):
Probably the relaxation of constraint being so early in this complex process is the biggest risk I can identify immediately. It looks like some re-ordering (especially of constraint/partial unique index relaxation), introduction of phases and bail-outs (usually via double-writing) to relax do-or-die deployment pressure ought to be introduced.

Let me know if you need me to work out the specifics on this one with you, but I suspect you have an idea of what I'm getting at. If such surgery is more expensive than my quick estimate, let me know so we can weigh the risks versus the costs together.

spec/model/project_spec.rb
```diff
@@ -130,19 +130,19 @@
...
-        {allocation_state: "accepting", location: "hetzner-fsn1", total_cores: 10, used_cores: 3},
-        {allocation_state: "accepting", location: "hetzner-fsn1", total_cores: 10, used_cores: 3},
-        {allocation_state: "accepting", location: "hetzner-fsn1", total_cores: 10, used_cores: 1},
-        {allocation_state: "accepting", location: "leaseweb-wdc02", total_cores: 100, used_cores: 99},
-        {allocation_state: "draining", location: "location-4", total_cores: 100, used_cores: 0},
-        {allocation_state: "accepting", location: "github-runners", total_cores: 100, used_cores: 0}
+        {allocation_state: "accepting", location_id: Location::HETZNER_FSN1_ID, total_cores: 10, used_cores: 3},
+        {allocation_state: "accepting", location_id: Location::HETZNER_FSN1_ID, total_cores: 10, used_cores: 3},
+        {allocation_state: "accepting", location_id: Location::HETZNER_FSN1_ID, total_cores: 10, used_cores: 1},
+        {allocation_state: "accepting", location_id: Location[name: "leaseweb-wdc02"].id, total_cores: 100, used_cores: 99},
```
jeremyevans, line 136: I think you have constants for these three locations as well.

views/components/form/resource_creation_form.erb
```diff
@@ -39,6 +39,12 @@ form_elements.each do |element|
     opt_attrs[:disabled] = "disabled" unless parents_selected
     opt_attrs[:checked] = "checked" if checked
 
+    opt_val = if option[name].is_a?(Location)
```
jeremyevans, line 42: No need for an else block, just only do the assignment inside the if block.

jeremyevans (commented):
I think this should be broken into more phases:

* Run add location_id migration
* Code change to set both location and location_id, but still operate on location
* Populate location_id for all rows
* Run migration to make location_id required and location optional
* Code change to switch to only using location_id and stop setting/accessing location
* Run migration to drop location

DEVELOPERS.md
furkansahin: I had actually changed it everywhere but this. Since it's an md file, I missed it. Thanks for catching.

migrate/20250227_id_based_locations.rb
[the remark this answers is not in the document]
furkansahin: Yes, I am thinking around the same lines with @fdr. If something goes wrong, it will just be the work of matching location_id with location. As long as the web and worker scale is set to 0, we shouldn't have any synchronisation issue.

migrate/20250226_project_id_location_indexes_drop.rb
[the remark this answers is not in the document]
furkansahin: I removed the migration, you're right.

migrate/20250227_id_based_locations.rb
```diff
@@ -0,0 +1,29 @@
...
+    %i[vm_host vm vm_pool private_subnet firewall postgres_resource minio_cluster inference_endpoint kubernetes_cluster].each do |table|
+      alter_table(table) do
+        set_column_allow_null :location
```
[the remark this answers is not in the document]
furkansahin: I moved it to the latest migration file.

furkansahin:
```
Run add location_id migration
Code change to set both location and location_id, but still operate on location
Populate location_id for all rows
Run migration to make location_id required and location optional
Code change to switch to only using location_id and stop setting/accessing location
Run migration to drop location
```
Considering the size of the PR, this will require a huge amount of re-work and I believe the impact would be minimal.

My plan is to scale down respirate and web workers to 0 for the first migration, handle the migration and then scale back up. Considering the size of our database - deleted_records, I don't think this causes a big downtime. I feel like the benefits really doesn't worth the effort. So, if this is not a huge blocker for you, I'd like to go ahead.

migrate/20250226_project_id_location_indexes_drop.rb
furkansahin: I removed the migration file. The constraint is removed when we remove the location column anyway.

fdr:
I think for the purposes of deadlock elimination a pause in processing is okay, but we should avoid a situation where we can't operate on any intermediate state of the schema, e.g. if a migration can't complete due to unforeseen constraint issues or whatever.

spec/model/private_subnet_spec.rb
```diff
@@ -96,11 +96,11 @@
     it "includes ubid if id is available" do
       ubid = described_class.generate_ubid
       private_subnet.id = ubid.to_uuid.to_s
-      expect(private_subnet.inspect).to eq "#<PrivateSubnet[\"#{ubid}\"] @values={:net6=>\"fd1b:9793:dcef:cd0a::/64\", :net4=>\"10.9.39.0/26\", :location=>\"hetzner-fsn1\", :state=>\"waiting\", :name=>\"ps\", :project_id=>\"#{private_subnet.project.ubid}\"}>"
+      expect(private_subnet.inspect).to eq "#<PrivateSubnet[\"#{ubid}\"] @values={:net6=>\"fd1b:9793:dcef:cd0a::/64\", :net4=>\"10.9.39.0/26\", :location_id=>\"10saktg1sprp3mxefj1m3kppq2\", :state=>\"waiting\", :name=>\"ps\", :project_id=>\"#{private_subnet.project.ubid}\"}>"
```
byucesoy, line 99: There are few places where you can use `Location::HETZNER_FSN1_ID`

byucesoy (commented):
Looks good IMO. I'll leave giving approval to Jeremy/Daniel since they did more through review.

spec/model/private_subnet_spec.rb
furkansahin: the variable is an id, the location is visualized as ubid. So, I'm leaving this as is.

migrate/20250228_project_id_location_indexes_to_id.rb
```diff
@@ -11,7 +11,7 @@
 
   down do
     %i[postgres_resource firewall private_subnet vm minio_cluster kubernetes_cluster].each do |table|
-      add_index table, [:project_id, :location, :name], name: :"#{table}_project_id_location_id_name_uidx", unique: true, concurrently: true
+      add_index table, [:project_id, :location, :name], name: :"#{table}_project_id_location_name_uidx", unique: true, concurrently: true
```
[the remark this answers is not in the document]
furkansahin: Yes, thanks for catchin this.

views/components/form/resource_creation_form.erb
jeremyevans: This was marked as resolved, but no change was made to implement it.

jeremyevans (commented):
Looks like all of my comments have been addressed, other than one in the resource creation form.  After that is made, no need for me to review again.

byucesoy (approved):

PR #2902, opened 2025-02-27
Increase SPDK iobuf sizes.

pykello:
In https://github.com/ubicloud/ubicloud/commit/ab3933bc93b50f9c8fb4633dceeb31e100677b1f we increased iobuf sizes to have enough resources for our current workload. It reduced the error rate across our fleet from 5 times per week to once in every few weeks. This PR further increases the resources to reduce the chance of error even more.

Since we fixed error handling with a previous release of bdev_ubi-0.3, the expected behavior in case of exceeding limits is an error instead of a crash.

### DIFF
--- a/prog/storage/setup_spdk.rb
+++ b/prog/storage/setup_spdk.rb
@@ -24,19 +24,20 @@ class Prog::Storage::SetupSpdk < Prog::Base
   label def start
     version = frame["version"]
     arch = vm_host.arch
+    spdk_hugepages = 4
 
     fail "Unsupported version: #{version}, #{arch}" unless SUPPORTED_SPDK_VERSIONS.include? [version, arch]
 
     fail "Can't install more than 2 SPDKs on a host" if vm_host.spdk_installations.length > 1
 
-    fail "No available hugepages" if frame["start_service"] && vm_host.used_hugepages_1g > vm_host.total_hugepages_1g - 2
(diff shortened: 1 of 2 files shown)


prog/storage/setup_spdk.rb
```diff
@@ -36,7 +36,7 @@ def self.assemble(vm_host_id, version, start_service: false, allocation_weight:
       allocation_weight: 0,
       vm_host_id: vm_host.id,
       cpu_count: vm_host.spdk_cpu_count,
-      hugepages: 3
```
enescakir, line 39: Yesterday, I noticed that we still check if the host has 2 hugepages, even though we allocated 3. We probably forgot to update it when we increased it to 3. We need to make that update. We could extract it to a variable and use that variable for both checks.

```ruby
    fail "No available hugepages" if frame["start_service"] && vm_host.used_hugepages_1g > vm_host.total_hugepages_1g - 2
```

fdr (approved):
Looks good.

Just dropping this here, so I am less likely to lose it, but, we may want to think hard about buffer usage and queuing next time we consider a major design revision https://people.kernel.org/kuba/nic-memory-reserve. But also, with regard to micro-architectural sizes, which have about 1MB L2 per core and 32MB shared L3 on a core complex (8 cores).

PR #2903, opened 2025-02-27
Prevent showing Burstable family for locations where it is not allowed

macieksarnowicz:
We got NoMethodError when viewing VM create page with burstables and location_latitude_fra feature flags enabled. This issue occurred because there is no billing rate for a burstable size for the latitude_fra location.
    
Add a BillingRate.unit_price_from_resource_properties that returns nil instead of raising a NoMethodError if there is no matching resource property. Use it to DRY up code in ContentGenerator::{Vm,Postgres}.
    
Make helpers/Vm#family and Postgres#family not return an Burstable option if there is no valid billing_rate at the requested location for the family. Also added a check for presence of a billing rate for Nexus.assemble and PostgresResource.assemble, as a back-stop.

Second commit is removing the Burstable billing_rates from regions where we cannot offer them yet, due to lack of servers.


pykello:
Looks good to me, but I'll wait for @byucesoy 's approval before merging

helpers/vm.rb
```diff
@@ -97,7 +97,9 @@ def generate_vm_options
     end
 
     options.add_option(name: "enable_ip4", values: ["1"], parent: "location")
-    options.add_option(name: "family", values: Option.families(use_slices: @project.get_ff_use_slices_for_allocation || false).map { _1.name }, parent: "location")
+    options.add_option(name: "family", values: Option.families(use_slices: @project.get_ff_use_slices_for_allocation || false).map { _1.name }, parent: "location") do |location, family|
+      !BillingRate.from_resource_properties("VmVCpu", family, LocationNameConverter.to_internal_name(location)).nil?
```
enescakir, line 101: nit: 
```suggestion
      !!BillingRate.from_resource_properties("VmVCpu", family, LocationNameConverter.to_internal_name(location))
```

lib/validation.rb
```diff
@@ -246,6 +246,12 @@ def self.validate_vcpu_quota(project, resource_type, requested_vcpu_count)
     end
   end
 
+  def self.validate_billing_rate(resource_type, resource_family, location)
+    if !BillingRate.from_resource_properties(resource_type, resource_family, location)
```
enescakir, line 250: nit:
```suggestion
    unless BillingRate.from_resource_properties(resource_type, resource_family, location)
```

prog/vm/nexus.rb
```diff
@@ -25,6 +25,7 @@ def self.assemble(public_key, project_id, name: nil, size: DEFAULT_SIZE,
     end
     Validation.validate_location(location)
     vm_size = Validation.validate_vm_size(size, arch)
+    Validation.validate_billing_rate("VmVCpu", vm_size.family, location)
```
enescakir, line 28: We don't have `burstable` rates for `github-runners` location. Since our runners are standard and not burstable, it shouldn't be an issue. But just to be sure, I started E2E test https://github.com/ubicloud/ubicloud/actions/runs/13584542704

enescakir (commented):
LGTM

prog/vm/nexus.rb
macieksarnowicz: Thank you! Looks like it has passed.

byucesoy (approved):

PR #2907, opened 2025-02-28
[openapi] remove obvious/explicit _id usage in openapi schema

geemus:
This removes the obvious/explicit _id api routes and makes associated fixes in the main schema.

A couple points of note:

1. The name regex seems overly broad, but I was having a hard time finding in the code what it should be exactly. Happy to change it if someone can tell me what it should be specifically.
2. I was surprised just how few _id paths there were in the schema. I think a bunch of additional stuff needs to be updated to allow name or id (but just never was updated/accurate in the schema to begin with), but I'm certainly willing to be corrected/wrong.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1446,19 +1446,17 @@ components:
         type: string
     postgres_database_reference:
       description: Postgres database ID or name
+      examples:
+        id:
+          value: pgn30gjk1d1e2jj34v9x0dq4rp
+        name:
+          value: postgres-database-name
       in: path
       name: postgres_database_reference
       required: true
       schema:
-        oneOf:
-          - type: string
-            example: pgn30gjk1d1e2jj34v9x0dq4rp
-            pattern: '^pg[0-9a-tv-z]{24}$'
-          - type: string
-            example: postgres-database-name
-            not:
-              pattern: '^pg[0-9a-tv-z]{24}$'
-            pattern: '^[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?$'
+        type: string
+        pattern: '^[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?$'
     postgres_firewall_rule_id:
       description: ID of the postgres firewall rule
       in: path
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -831,44 +831,6 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
-  '/project/{project_id}/location/{location}/postgres/{postgres_database_name}/firewall-rule':
-    parameters:
-      - $ref: '#/components/parameters/location'
-      - $ref: '#/components/parameters/postgres_database_name'
-      - $ref: '#/components/parameters/project_id'
-    get:
-      operationId: listLocationPostgresFirewallRules
-      summary: List location Postgres firewall rules
-      responses:
-        '200':
-          $ref: '#/components/responses/PostgresFirewallRules'
-        default:
-          $ref: '#/components/responses/Error'
-      tags:
-        - Postgres Firewall Rule
-    post:
-      operationId: createLocationPostgresFirewallRule
-      summary: Create a new postgres firewall rule
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              type: object
-              properties:
-                cidr:
-                  description: CIDR of the firewall rule
-                  type: string
-              additionalProperties: false
-              required:
-                - cidr
-      responses:
-        '200':
-          $ref: '#/components/responses/PostgresFirewallRule'
-        default:
-          $ref: '#/components/responses/Error'
-      tags:
-        - Postgres Firewall Rule
   '/project/{project_id}/location/{location}/postgres/{postgres_database_name}/firewall-rule/{firewall_rule_id}':
     parameters:
       - $ref: '#/components/parameters/project_id'
@@ -1009,6 +971,44 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/firewall-rule':
+    parameters:
+      - $ref: '#/components/parameters/location'
+      - $ref: '#/components/parameters/postgres_database_reference'
+      - $ref: '#/components/parameters/project_id'
+    get:
+      operationId: listLocationPostgresFirewallRules
+      summary: List location Postgres firewall rules
+      responses:
+        '200':
+          $ref: '#/components/responses/PostgresFirewallRules'
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Postgres Firewall Rule
+    post:
+      operationId: createLocationPostgresFirewallRule
+      summary: Create a new Postgres firewall rule
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                cidr:
+                  description: CIDR of the firewall rule
+                  type: string
+              additionalProperties: false
+              required:
+                - cidr
+      responses:
+        '200':
+          $ref: '#/components/responses/PostgresFirewallRule'
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Postgres Firewall Rule
   '/project/{project_id}/location/{location}/private-subnet':
     get:
       operationId: listLocationPrivateSubnets
@@ -1444,6 +1444,19 @@ components:
       required: true
       schema:
         type: string
+    postgres_database_reference:
+      description: Postgres database ID or name
+      in: path
+      name: postgres_database_reference
+      required: true
+      schema:
+        oneOf:
+          - type: string
+            example: pgn30gjk1d1e2jj34v9x0dq4rp
+            pattern: '^pg[0-9a-tv-z]{24}$'
+          - type: string
+            example: postgres-database-name
+            pattern: '^[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?$'
     postgres_firewall_rule_id:
       description: ID of the postgres firewall rule
       in: path
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1456,6 +1456,8 @@ components:
             pattern: '^pg[0-9a-tv-z]{24}$'
           - type: string
             example: postgres-database-name
+            not:
+              pattern: '^pg[0-9a-tv-z]{24}$'
             pattern: '^[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?$'
     postgres_firewall_rule_id:
       description: ID of the postgres firewall rule


openapi/openapi.yml
```diff
@@ -1370,15 +1327,19 @@ components:
...
+      name: postgres_database_identifier
       required: true
       schema:
-        type: string
-        example: pgn30gjk1d1e2jj34v9x0dq4rp
-        pattern: '^pg[0-9a-hj-km-np-tv-z]{24}$'
+        oneOf:
+          - type: string
+            example: pgn30gjk1d1e2jj34v9x0dq4rp
+            pattern: '^pg[0-9a-hj-km-np-tv-z]{24}$'
```
jeremyevans, line 1339: For backwards compatibility, you have to allow an optional `_` at the beginning.

Additionally, we should probably accept `[0-9a-tv-z]`.  The ubid code will allow `i`, `l`, and `o`, and turn them into `1`, `1`, and `0`.  I'm unsure why `u` isn't converted to `v`, seems like an oversight to not handle all 26 characters.  Maybe @pykello knows?

openapi/openapi.yml
```diff
@@ -1370,15 +1327,19 @@ components:
...
+          - type: string
+            example: postgres_database_name
+            pattern: '^(?:(?!pg)).{1,26}$'
```
jeremyevans, line 1342: The allowed name pattern is in `Validation::ALLOWED_NAME_PATTERN`.  It does not allow underscores, so the example needs to be fixed.

spec/routes/api/project/location/postgres_spec.rb
```diff
@@ -35,7 +35,7 @@
         [:post, "/project/#{project.ubid}/location/#{pg.display_location}/postgres/_#{pg.ubid}/restore"],
         [:post, "/project/#{project.ubid}/location/#{pg.display_location}/postgres/#{pg.name}/reset-superuser-password"],
         [:post, "/project/#{project.ubid}/location/#{pg.display_location}/postgres/_#{pg.ubid}/reset-superuser-password"],
-        [:post, "/project/#{project.ubid}/location/#{pg.display_location}/postgres/_#{pg.ubid}/failover"],
```
jeremyevans, line 38: This should be left for backwards compatibility, unless @fdr thinks we should drop backwards compatibility.  Same for the similar changes in this file.

spec/routes/api/project/location/postgres_spec.rb
```diff
@@ -11,7 +11,7 @@
     Prog::Postgres::PostgresResourceNexus.assemble(
       project_id: project.id,
       location: "hetzner-fsn1",
-      name: "pg-with-permission",
+      name: "postgres-with-permission",
```
jeremyevans, line 14: This doesn't appear related to the other changes.

openapi/openapi.yml
```diff
@@ -1370,15 +1327,19 @@ components:
       schema:
         type: integer
         default: 10
-    postgres_database_id:
-      description: Postgres database ID
+    postgres_database_identifier:
```
jeremyevans, line 1330: Up to you, but maybe `postgres_database_reference` instead?

spec/routes/api/project/location/postgres_spec.rb
fdr: I think we're pre-governance so we could drop it. I think we're still at the stage we could expunge the `_` methodology altogether.

spec/routes/api/project/location/postgres_spec.rb
geemus: I thought that with the changes, things shouldn't have names that start with the same prefix as their id would. I'm pretty sure this was throwing errors once I made the other changes. So it seemed related, am I missing something in my understanding?

jeremyevans: Ubids cannot have dashes, and this isn't exactly 26 characters, so this isn't ambiguous.  Only names that could be valid ubids for the model are rejected.

geemus: Oof, right. I forgot how narrow the conflict was.

openapi/openapi.yml
geemus: I like that better. I didn't love identifier because it felt to close to id and seemed like it could create confusion, but wasn't coming up with a suitable alternative in the moment. I changed this to reference in https://github.com/ubicloud/ubicloud/pull/2907/commits/15994d3999d9c1d0eaab7eab2afa0dce4764e5ea

spec/routes/api/project/location/postgres_spec.rb
geemus: Resolving as it seems we are going this direction and it's progressed in #2912 (I'll work on o rebase soon).

openapi/openapi.yml
geemus: Thanks. I think we aren't worrying about backwards compatibility any more, but let me know if I'm wrong there. Otherwise, I've updated here: https://github.com/ubicloud/ubicloud/pull/2907/commits/b2e7dcf1004ad4a3609bfc8e9b8a79fde061d845

openapi/openapi.yml
geemus: I adjusted it to match that validation here: https://github.com/ubicloud/ubicloud/pull/2907/commits/fb5946d1332379d1542ba6c48ee2418987b6bdb1

Note, I think this technically is not quite right as it ought to specifically omit the case where it begins with pg and is 26 characters, but my regex-foo is at least presently failing me in terms of how to write that.

geemus:
I'm going to squash these commits to simplify the history (and hopefully then the rebase as well).

geemus:
I think/hope this largely fixes the database stuff here, but I believe it might still be the case that many other things would accept references and it wasn't reflect in the spec. It would be good to validate the approach here first though, and then I can work on making it more universal if we are satisfied.

geemus:
I need to wrap for the night, but will circle back. For reference, I changed the regexes to be more accurate and now it's failing because OneOf is not valid unless one and ONLY one matches. Since the name regex doesn't disallow ubid, they both match and it fails. I attempted to fix this with a not (since that seemed simpler than changing the regex), but had no luck. I'm not sure if it should work that way and committee isn't aware or not (since I was struggling to find docs/examples on not). In any event, I need to crash, will look again in my next work block.

jeremyevans:
> I need to wrap for the night, but will circle back. For reference, I changed the regexes to be more accurate and now it's failing because OneOf is not valid unless one and ONLY one matches. Since the name regex doesn't disallow ubid, they both match and it fails. I attempted to fix this with a not (since that seemed simpler than changing the regex), but had no luck. I'm not sure if it should work that way and committee isn't aware or not (since I was struggling to find docs/examples on not). In any event, I need to crash, will look again in my next work block.

My recommendation is to only use the name regexp for path segments.  The ubid pattern is a subset of the name pattern.  When you are accepting new names, you cannot use a valid ubid, but for path references, either a ubid or a name will work.

I don't see the reason to do a lot of work to get OneOf to work.  Is there a technical advantage to using OneOf combining (subset of pattern) with (pattern except subset of pattern), instead of using a single regexp (pattern)?  I can guess that the OneOf approach is more philosophically pure, but I see no pragmatic advantage for it.

geemus:
@jeremyevans Fair point, I'm fine with just using name.

In terms of the regex, I was trying to use oneof with two regexes, because it seemed simpler to reason about and edit than one big regex that tried to combine the conditions of both. If we don't plan to combine the conditions, then this is a moot point.

In terms of value, with the broader regex, it's possible that some invalid things wouldn't be caught by the validations, but I think the likelihood is quite low. ie I think technically a ubid length and prefixed string containing a `u` would be missed? It might also impact how accurate the mocks would be, but I don't think anyone is actively using/relying on those presently. So in the end I suppose it boils down mostly to philosophical purity.

jeremyevans:
> In terms of value, with the broader regex, it's possible that some invalid things wouldn't be caught by the validations, but I think the likelihood is quite low. ie I think technically a ubid length and prefixed string containing a `u` would be missed? It might also impact how accurate the mocks would be, but I don't think anyone is actively using/relying on those presently. So in the end I suppose it boils down mostly to philosophical purity.

The only places where you are combining the regexes are the path segments that take either name or ubid.  There is no "validation" of such segments.  If it is in ubid format for the related model, it is treated as a ubid.  If it is not in ubid format for the model, it is treated as a name.  From an openapi perspective, there is separate validation for name and ubid (everything matching the name validation regex could potentially be valid) .  I think that separating the regexes would boil down solely to philosophical purity.

geemus:
I think this should be good to go now, anything I'm missing?

openapi/openapi.yml
```diff
@@ -997,6 +944,59 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/failover':
```
byucesoy, line 947: We removed failover endpoints from the codebase with [this PR](https://github.com/ubicloud/ubicloud/pull/2923), FYI

byucesoy, line 947: We removed failover endpoints from the codebase with https://github.com/ubicloud/ubicloud/pull/2923, FYI.

byucesoy (commented):
I didn't understand what is special about `firewall-rule` endpoint, that we are changing it to use `postgres_database_reference` but not other endpoints that uses `postgres_database_name`.

geemus:
@byucesoy Good question.

I was particularly focused on changing it because without this change it still uses the `_id` style, which has been removed altogether. There definitely are a significant number of additional places that should also be changed to the `..._reference` style, but I wanted to make sure we were on the same page/comfortable with this approach before doing it throughout.

So, if this approach seems reasonable, it's a good step in the right direction which I think would be fine to merge even now. And then I can work on updating everything to be consistent.

Thanks for the heads up on failover, I can work on removing that also (I'll likely keep that in it's own PR as well just to keep things cleaner and hopefully less confusing).

Does that help clarify the intent and scope for this particular PR?

geemus:
Just rebased also in order to resolve the conflict.

jeremyevans (changes requested):
This looks good, except that the failover endpoint should be removed, as mentioned by @byucesoy.

geemus:
Updated to remove failover.

As we are otherwise good I think, I can plan to follow up with fleshing out the many other name or id things to match in an upcoming, separate PR (as this should already be a clear improvement that need not further block).

byucesoy:
For the record, the PR I linked already removes the failover endpoint from openapi.yml file. This PR adds it back and then removes it :slightly_smiling_face:, not a big deal, just wanted to clarify.

byucesoy (approved):
For the record, the PR I linked already removes the failover endpoint from openapi.yml file. This PR adds it back and then removes it 🙂, not a big deal, just wanted to clarify. Looks good otherwise.

geemus:
@byucesoy Oops, missed that, sorry.

geemus:
@jeremyevans looks like it will need you to approve still to allow merging.

jeremyevans (approved):

openapi/openapi.yml
geemus: I rebased and combined the removal with the first commit (in effect removing it from this altogether). So hopefully that fixes any history or other issues. Thanks for the heads up.

PR #2909, opened 2025-02-28
Update aws-sdk-s3 from 1.177 to 1.182

enescakir:
Actually, the "aws-sdk-s3" gem releases a new version every week, and generally these updates are harmless, auto-generated packages. However, version "1.177" has different content.

They changed the client's default checksum validation to CRC32, which Cloudflare R2 doesn't support. This change breaks many applications. [^1]

Cloudflare recommends either pinning to version 1.177 or passing the parameters `request_checksum_calculation: "when_required"` and `response_checksum_validation: "when_required"` to the S3 client [^2]. Since we'll need to update the version eventually, it's a good idea to do it now.

[^1]: https://www.cloudflarestatus.com/incidents/t5nrjmpxc1cj
[^2]: https://developers.cloudflare.com/r2/examples/aws/aws-sdk-ruby/

### DIFF
--- a/model/github/github_repository.rb
+++ b/model/github/github_repository.rb
@@ -28,12 +28,7 @@ class GithubRepository < Sequel::Model
   end
 
   def blob_storage_client
-    @blob_storage_client ||= Aws::S3::Client.new(
-      endpoint: Config.github_cache_blob_storage_endpoint,
-      access_key_id: access_key,
-      secret_access_key: secret_key,
-      region: Config.github_cache_blob_storage_region
-    )
+    @blob_storage_client ||= s3_client(access_key, secret_key)
   end
 
   def url_presigner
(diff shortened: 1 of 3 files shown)


fdr (approved):

model/github/github_repository.rb
```diff
@@ -89,6 +79,16 @@ def setup_blob_storage
       update(access_key: token_id, secret_key: Digest::SHA256.hexdigest(token))
     end
   end
+
+  private def s3_client(access_key_id, secret_access_key)
+    Aws::S3::Client.new(
+      endpoint: Config.github_cache_blob_storage_endpoint,
+      access_key_id:, secret_access_key:,
```
jeremyevans, line 86:
```suggestion
      access_key_id:,
      secret_access_key:,
      endpoint: Config.github_cache_blob_storage_endpoint,
```

Up to you, but I think this reads nicer.

fdr (approved):
huh. notch that one up for 'dangers of relying on someone else to write your storage sdk'

PR #2911, opened 2025-02-28
Minor cli improvements

jeremyevans:
* Include version in cli filenames
* Do not build Windows 386 version
* Reduce file size
* Update reported cli version to 1.0.

CC @geemus


fdr (approved):

geemus (approved):

geemus:
@jeremyevans Thanks for the ping. This looks good as it is, but I'll add a few thoughts from my own experience as you continue forward.

First, I recall us having to file an issue somewhere and chase things down relating to windows seeing us as malware, but it's been so long I forget the details. So it may be fixable/recoverable, but I think there was a fair amount of lag.

Second, establishing a set version is a great first step and as you say helps provide better error messages. It will be another governance related thing to sort out. ie would you always try to still operate but warn/suggest they should upgrade? Or do all out-of-date versions simply error and not allow operation until they've been updated? Or perhaps something in between (many it's ok to be out-of-date in terms of patch versions, but minors or perhaps majors would error). Like many other things, any of these could be fine, but it's probably best to be clear and explicit. I think we settled on basically reporting back a minimum-acceptable version. Anything above that gets a nudge/warning, and anything below that fails unless you update first. That has worked well enough so far (we usually don't change the minimum, but have if/when we make known breaking changes). Certainly something you can figure out if/when you get to it, but might be good to have a plan at least.

jeremyevans:
> First, I recall us having to file an issue somewhere and chase things down relating to windows seeing us as malware, but it's been so long I forget the details. So it may be fixable/recoverable, but I think there was a fair amount of lag.

I'm not sure if it's worth the effort to do that, considering it only affects the 32-bit version and not the 64-bit version, and it's really a stretch that someone would be running windows and wouldn't be able to use the 64-bit version.

> Second, establishing a set version is a great first step and as you say helps provide better error messages. It will be another governance related thing to sort out. ie would you always try to still operate but warn/suggest they should upgrade? Or do all out-of-date versions simply error and not allow operation until they've been updated? Or perhaps something in between (many it's ok to be out-of-date in terms of patch versions, but minors or perhaps majors would error).

IMO, we should never error unless we know the client's version is not capable of performing the cli request.  For example, since we whitelist programs before execution, if we later add a new whitelisted program, but the client's version indicates they don't support that, then we should provide an error message telling the client to upgrade, instead of the client receiving an error that the server requested an unsupported program.

The way the client is designed, anything that only requires printing information should work for all client versions.  It would only be new capabilities that would need new versions.  In addition to new whitelisted programs, other things that would require an upgraded program would be:

* If we wanted to support multiple confirmation prompts (or anything where we need to ask users for information multiple times).
* If we wanted to support a file upload/download feature.

geemus:
@jeremyevans sounds good. It might be worth capturing that somewhere for future reference/posterity (though I suppose the comment could be found again).

fdr (approved):

PR #2912, opened 2025-02-28
Drop support for optional leading underscore for ubids in routes

jeremyevans:
Also, do not allow `u` as a character, since ubids cannot include it.

Some of this overlaps with #2907, but this handles the actual route support and updates all related specs.

CC @geemus

### DIFF
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -3,7 +3,7 @@
 class Clover < Roda
   def self.name_or_ubid_for(model)
     # (\z)? to force a nil as first capture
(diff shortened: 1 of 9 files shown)


fdr (approved):
I suppose it might be nice to have the commit give the justification (still pre-governace, can simplify by removing) but this is just a kibitz.

jeremyevans:
I checked the logs and nobody has been using the underscored ubids, so that seems sufficient reason to remove it.  I updated the commit message to mention that.

PR #2913, opened 2025-03-01
Add golang to .tool-versions to compile ubi cli

fdr:
Otherwise, this fails if one has the asdf golang plugin installed.  An alternative would be to tell .tool-versions to use the "system" golang if we think forcing people who don't work with the CLI to get new versions of Go from time to time is too invasive.

On the other hand, bringing golang into asdf's purview means it's very easy for everyone working with the project to instrument or alter the cli.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,4 @@
 ruby 3.2.7
 postgres 15.8
 nodejs 23.6.0
+golang 1.24.0
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -59,13 +59,14 @@ asdf plugin add <name> [<git-url>]      Add a plugin from the plugin repo OR,
 [...]
 ```
 
-We like to have these four plugins (you can paste these commands):
+We like to have these plugins (you can paste these commands):
 
 ```sh
 $ asdf plugin add ruby
 $ asdf plugin add direnv
 $ asdf plugin add postgres
 $ asdf plugin add nodejs
+$ asdf plugin add golang
 ```
 
 Once you have the plugins, you can start to install the software the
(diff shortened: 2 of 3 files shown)


fdr:
I assigned some cross-section of the team, because while I think we ought to either this patch or something like this:

```diff
-golang 1.24.0
+golang system
```

that, if you use asdf for any go project (that is, it is aware of the go plugin at all), without specifying _something_ you get this:


```
> bundle exec rake ubi
cd cli && go build -tags osusergo,netgo
No version is set for command go
Consider adding one of the following versions in your config file at /home/fdr/code/clover/.tool-versions
golang 1.22.5
golang 1.23.2
golang 1.23.3
golang 1.23.4
golang 1.24.0
rake aborted!
Command failed with status (126): [cd cli && go build -tags osusergo,netgo]
/home/fdr/code/clover/Rakefile:283:in `block in <top (required)>'
/home/fdr/.asdf/installs/ruby/3.2.7/bin/bundle:25:in `load'
/home/fdr/.asdf/installs/ruby/3.2.7/bin/bundle:25:in `<main>'
Tasks: TOP => ubi
(See full trace by running task with --trace)
```

Which sucks, because you have to maintain an out-of-tree patch against the code to get asdf to chill out.

.tool-versions
```diff
@@ -1,3 +1,4 @@
 ruby 3.2.7
 postgres 15.8
 nodejs 23.6.0
+golang 1.24.0
```
jeremyevans, line 4: This isn't currently the go version used by the cli (`cli/go.mod` shows 1.22.2).  No objections to updating the go version, but we should make sure they match.

fdr: I wonder why nothing complains. And weirdly best I can tell, `go mod` doesn't have a quick option to insert the current running go version. I wonder if it's an implied >=. Anyway, updated, for now

jeremyevans (approved):

PR #2915, opened 2025-03-03
Customer specific region creation

furkansahin:
## Add Private Locations migration
We are following a design where locations now have project_id and a
LocationCredential entity's ID is a foreign key to a Location entity.

This way, we can have locations that are not attached to a specific
project and if they are, we create a location_credential entity.

## Private Location management experience with UI
We are adding a new tab under Project Details for listing private
locations. We keep it behind a feature flag. The create view simply gets
the AWS region, ubicloud region name, access and secret keys. We list
the locations and give chance to customer to be able to update the name.

## Billing and PG integrations of private locations
Enable the privately created locations in the UI to be listed as one of
the options while creating the PostgresResource. That also requires some
billing integration, validation and option changes.

## Add project_id, display_name index to location - migration 2


migrate/20250303_add_aws_location.rb
```diff
@@ -0,0 +1,71 @@
...
+    create_table(:aws_location_credential) do
+      column :id, :uuid, primary_key: true
+      column :access_key, String, null: false, collate: '"C"'
+      column :secret_key, String, null: false, collate: '"C"'
+      column :region_name, String, null: false
+      foreign_key :project_id, :project, type: :uuid, null: false
+    end
+
+    alter_table(:location) do
+      add_foreign_key :aws_location_credential_id, :aws_location_credential, type: :uuid, null: true
```
jeremyevans, line 14: I recommend a check constraint:

```ruby
constraint :aws_location_credential_id_provider_check,
  {Sequel.~(aws_location_credential_id: nil) => {provider: "aws"}}
# (aws_location_credential_id IS NOT NULL) = (provider = 'aws')
```

Additionally, there should be a unique constraint/index on `aws_location_credential_id`, since you don't want multiple locations pointing to it.

model/project.rb
```diff
@@ -28,10 +28,12 @@ class Project < Sequel::Model
   one_to_many :quotas, class: :ProjectQuota, key: :project_id
   one_to_many :invitations, class: :ProjectInvitation, key: :project_id
   one_to_many :api_keys, key: :owner_id, class: :ApiKey, conditions: {owner_table: "project"}
+  one_to_many :aws_location_credentials, key: :project_id, class: :AwsLocationCredential
```
jeremyevans, line 31:
```suggestion
  one_to_many :aws_location_credentials
```

I think the values provided are what Sequel would chose as defaults.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+class Clover
+  hash_branch(:project_prefix, "region") do |r|
+    r.get true do
+      puts "project: #{@project.path}"
```
jeremyevans, line 6:
```suggestion
```

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+      regions = @project.aws_location_credentials
+
+      if api?
+        Serializers::Region.serialize(regions)
+      else
+        @regions = Serializers::Region.serialize(regions)
```
jeremyevans, line 12: We should avoid the use of serializers as presenters for web views going forward.  Please change all views in this PR to call methods directly on the Region instances, and remove the use of serializers for web views.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+        view "region/index"
+      end
+    end
+
+    r.post true do
+      puts "geldieem project: #{@project.path}"
```
jeremyevans, line 18:
```suggestion
```

routes/project/region.rb
jeremyevans, line 1: This seems specific to AWS, so we should probably use `aws_region` instead of `region` for the naming of the route and serializer.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+      request_body_params = validate_request_params(required_parameters)
+      unless @project
+        if api?
+          fail ValidationError.new("Project not found")
+        else
+          r.redirect "region/create"
+        end
+      end
+
+      if @project.accounts_dataset.where(Sequel[:accounts][:id] => current_account_id).empty?
```
jeremyevans, line 30: Can we reach a point where this is true?  This looks like it duplicates a check in `routes/project.id`?

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      pl = DB.transaction do
+        alc = AwsLocationCredential.create_with_id(
+          access_key: request_body_params["aws_access_key"],
+          secret_key: request_body_params["aws_secret_key"],
+          region_name: request_body_params["aws_region_name"],
+          project_id: @project.id
+        )
+        Location.create_with_id(
+          display_name: request_body_params["name"].downcase.tr(" ", "-"),
```
jeremyevans, line 42: We don't seem to validate this name as a valid name, and I think we should.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+        view "region/index"
+      end
+    end
+
+    r.post true do
+      puts "geldieem project: #{@project.path}"
+      authorize("Region:create", @project.id)
+      required_parameters = ["name", "aws_region_name", "aws_access_key", "aws_secret_key"]
+      request_body_params = validate_request_params(required_parameters)
+      unless @project
```
jeremyevans, line 22: Can you reach this point without `@project` being set?  A review of `routes/project.rb` indicates no.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+      @available_aws_regions = ["us-east-1", "us-west-1"]
+      view "region/create"
+    end
+
+    r.on String do |region_ubid|
+      @region = AwsLocationCredential.from_ubid(region_ubid)
+
+      next(r.delete? ? 204 : 404) unless @region
+
+      if @region.project.accounts_dataset.where(Sequel[:accounts][:id] => current_account_id).empty?
```
jeremyevans, line 69: This code isn't needed if you already check that the region is related to the project.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+        fail Authorization::Unauthorized
+      end
+
+      @project = @region.project
```
jeremyevans, line 73: `@project` should already be set, and if `region.project` is different, you have problems.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    r.get(web?, "create") do
+      authorize("Region:create", @project.id)
+      @available_aws_regions = ["us-east-1", "us-west-1"]
+      view "region/create"
+    end
+
+    r.on String do |region_ubid|
+      @region = AwsLocationCredential.from_ubid(region_ubid)
```
jeremyevans, line 65:
```suggestion
      @region = @project.aws_location_credentials_dataset.first(id: UBID.to_uuid(region_ubid))
```

You should check the region is valid for the current project.

routes/project/region.rb
jeremyevans, line 73:
```suggestion
```

`@project` is already set.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+      end
+    end
+
+    r.get(web?, "create") do
+      authorize("Region:create", @project.id)
+      @available_aws_regions = ["us-east-1", "us-west-1"]
+      view "region/create"
+    end
+
+    r.on String do |region_ubid|
```
jeremyevans, line 64:
```suggestion
    r.is String do |region_ubid|
```

No subpaths are supported here.  So it's better do use `is` here, and remove the `true` argument to `r.get`, `r.delete`, and `r.post` inside the block.

routes/project/region.rb
```diff
@@ -0,0 +1,111 @@
...
+        if @region.has_resources
+          fail DependencyError.new("'#{@region.name}' region has some resources. Delete all related resources first.")
+        end
+
+        @region.location.destroy
+        @region.destroy
+        204
+      end
+
+      if web?
```
jeremyevans, line 99: Any reason to make this web-only and not support updating the name in the API (and therefore CLI)?

migrate/20250303_add_aws_location.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    run "INSERT INTO provider (name) VALUES ('aws');"
+
+    run <<~SQL
+      INSERT INTO
+        action_type (id, name)
+      VALUES
+        ('ffffffff-ff00-830e-87ff-f02b80d85dc0', 'Region:view'),   -- rezzzzzzzz021gzzz0aw0v1ew0
```
jeremyevans, line 23: I would name the permissions `AwsLocationCredential` instead of `Region`, unless we plan to use the same permission for non-AWS regions in the future.

views/region/index.erb
```diff
@@ -0,0 +1,31 @@
+<% @page_title = "Custom Regions" %>
+
+<%== part(
+  "components/page_header",
+  right_items: ((@regions&.count > 0) && has_project_permission("Region:create")) ? [
```
jeremyevans, line 5:
```suggestion
  right_items: ((!@regions.empty?) && has_project_permission("Region:create")) ? [
```

I don't think this page is rendered without `@regions` being set.  Coverage testing should tell you for sure.  Also, it's better to use `!empty?` instead of `count > 0`.

routes/project/region.rb
furkansahin: Changed it

migrate/20250303_add_aws_location.rb
furkansahin: I used `AwsRegion` instead

migrate/20250303_add_aws_location.rb
furkansahin: I have added additional indexes.

lib/validation.rb
```diff
@@ -54,7 +54,7 @@ def self.validate_minio_username(username)
   end
 
   def self.validate_postgres_location(location)
-    available_pg_locs = Option.postgres_locations
+    available_pg_locs = Option.postgres_all_locations
```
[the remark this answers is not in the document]
furkansahin: Right, I have changed that.

helpers/postgres.rb
```diff
@@ -86,19 +86,23 @@ def generate_postgres_options(flavor: "standard")
...
       available_families = Option.families(use_slices: @project.get_ff_use_slices_for_allocation || false).map { _1.name }
       available_families.include?(family) && BillingRate.from_resource_properties("PostgresVCpu", "#{flavor}-#{family}", location.name)
     end
-    options.add_option(name: "size", values: Option::PostgresSizes.map { _1.name }.uniq, parent: "family") do |flavor, location, family, size|
-      pg_size = Option::PostgresSizes.find { _1.name == size && _1.flavor == flavor && _1.location_id == location.id }
+    options.add_option(name: "size", values: (Option::PostgresSizes + customer_postgres_sizes_for_project(@project.id)).map { _1.name }.uniq, parent: "family") do |flavor, location, family, size|
+      puts "size: #{size}"
+      puts "flavor: #{flavor}"
+      puts "location: #{location.id}"
+      puts "all sizes: #{Option::PostgresSizes + customer_postgres_sizes_for_project(@project.id)}"
```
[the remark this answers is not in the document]
furkansahin: Yeap, it's just that the PR I am referring to is still in review, so I was feeling relaxed on this one. I'll be more careful next time.

migrate/20250303_add_aws_location.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_table(:aws_location_credential) do
```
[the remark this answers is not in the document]
furkansahin: Considering we are implementing this as an MVP, I believe it's fine to keep it. I agree with your statement but this is not planned to be massively used by customers and the worst case scenario, we can just rename the table name :D It's easy.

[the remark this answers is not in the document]
furkansahin: I have refactored it.

migrate/20250303_add_aws_location.rb
```diff
@@ -0,0 +1,81 @@
...
+      add_foreign_key :aws_location_credential_id, :aws_location_credential, type: :uuid, null: true
+      add_constraint :aws_location_credential_id_provider_check,
+        {Sequel.~(aws_location_credential_id: nil) => {provider: "aws"}}
+      # where aws_location_credential_id is not null, it must be unique
+      add_unique_constraint :aws_location_credential_id,
+        where: Sequel.~(aws_location_credential_id: nil),
+        name: :aws_location_credential_id_unique
+      add_unique_constraint [:name, :aws_location_credential_id],
+        where: Sequel.~(aws_location_credential_id: nil),
+        name: :aws_location_credential_id_name_unique
```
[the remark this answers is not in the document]
furkansahin: I made a refactoring, can you please look at the indexes again?

migrate/20250303_add_aws_location.rb
```diff
@@ -0,0 +1,81 @@
...
+      unique [:project_id, :region_name]
+    end
+
+    alter_table(:location) do
+      add_foreign_key :aws_location_credential_id, :aws_location_credential, type: :uuid, null: true
+      add_constraint :aws_location_credential_id_provider_check,
+        {Sequel.~(aws_location_credential_id: nil) => {provider: "aws"}}
+      # where aws_location_credential_id is not null, it must be unique
+      add_unique_constraint :aws_location_credential_id,
+        where: Sequel.~(aws_location_credential_id: nil),
```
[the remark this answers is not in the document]
furkansahin: Same here

model/aws_location_credential.rb
```diff
@@ -0,0 +1,41 @@
...
+    enc.column :access_key
+    enc.column :secret_key
+  end
+
+  def path
+    "#{project.path}/aws-region/#{ubid}"
+  end
+
+  def has_resources
+    project.postgres_resources_dataset.where(location_id: location.id).count > 0
```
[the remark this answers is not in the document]
furkansahin: Nope, it's used only once when delete location is called. So, I will keep it as is.

routes/project/aws_region.rb
```diff
@@ -0,0 +1,97 @@
...
+        if api?
+          204
+        else
+          r.redirect "#{@project.path}/aws-region/index"
+        end
+      end
+
+      r.post do
+        authorize("AwsRegion:edit", @project.id)
+        @region.location.update(ui_name: r.params["name"], display_name: r.params["name"])
```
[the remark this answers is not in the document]
furkansahin: The columns are individually used for default locations. For Private locations, I don't think that's necessary. It's like 
hetzner-fsn1 is the location name
eu-central-a1 is the location display name
Germany is the location UI name

We don't have that for private locations.

views/private-location/create.erb
```diff
@@ -0,0 +1,23 @@
+<% @page_title = "Create AWS Region" %>
+
+<%== part(
+  "components/page_header",
+  breadcrumbs: [
+    %w[Projects project],
+    [@project_data[:name], @project_data[:path]],
+    ["AWS Regions", "#{@project_data[:path]}/private-location"],
```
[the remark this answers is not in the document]
furkansahin: Yes, I wanted to make everything on the backend ready for multiple providers. In the UI, I am just introducing this as an AWS specific thing because "Private Locations" term would need some explaining to the customers. 
The vision is, when we decide to support multiple providers, I only make the UI changes to tell the story that we have multiple providers.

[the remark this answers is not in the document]
furkansahin: We'll only need to add the "provider" as a parameter to the UI then.

migrate/20250303_add_private_location.rb
```diff
@@ -0,0 +1,69 @@
...
+    run "INSERT INTO provider (name) VALUES ('aws');"
+
+    run <<~SQL
+      INSERT INTO
+        action_type (id, name)
+      VALUES
+        ('ffffffff-ff00-835a-87ff-f00400d85dc0', 'Location:view'),   -- ttzzzzzzzz021gzzz0100v1ew0
+        ('ffffffff-ff00-835a-87c0-10019872b4e0', 'Location:create'), -- ttzzzzzzzz021gz0aw0create0
+        ('ffffffff-ff00-835a-87ff-f004007343a0', 'Location:edit'),   -- ttzzzzzzzz021gzzz0aw0ed1t1
+        ('ffffffff-ff00-835a-87c0-1001ae0bb4e0', 'Location:delete'); -- ttzzzzzzzz021gz0aw0de1ete1
```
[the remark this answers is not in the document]
furkansahin: The uuids were regenerated but the ubids (comment) were not updated, I have updated them now

lib/billing_rate.rb
```diff
@@ -15,6 +15,14 @@ def self.rates
   end
 
   def self.from_resource_properties(resource_type, resource_family, location, active_at = Time.now)
+    rate = select_rate(resource_type, resource_family, location, active_at)
+    if rate.nil? && (location = Location[name: location]) && location.location_credential
```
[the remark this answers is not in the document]
furkansahin: I removed the logic to pull the entities.

migrate/20250307_add_location_index.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
+
+  up do
+    alter_table(:location) do
+      add_index [:project_id, :display_name], name: :location_project_id_display_name_uidx, unique: true, concurrently: true
```
jeremyevans, line 8: Considering the size of the `location` table, is it really necessary to build the index concurrently?  Can we disable the Rubocop warning and include this in the single migration?

routes/project/private_location.rb
```diff
@@ -0,0 +1,115 @@
...
+      options.add_option(name: "provider_location_name", values: Option::AWS_LOCATIONS.map { |l| {value: l, display_name: l} })
+      options.add_option(name: "access_key")
+      options.add_option(name: "secret_key")
+      @option_tree, @option_parents = options.serialize
+
+      view "private-location/create"
+    end
+
+    r.is String do |name|
+      @location = @project.locations.find { |loc| loc.ui_name == name }
```
jeremyevans, line 72:
```suggestion
      @location = @project.locations_dataset.first(ui_name: name)
```

Are we enforcing unique `ui_name` per project?

jeremyevans (commented):
Changes look good, only a couple additional comments.

routes/project/private_location.rb
furkansahin: Now, we do. Thanks for catching this.

migrate/20250307_add_location_index.rb
furkansahin: Makes sense, I moved it.

jeremyevans (approved):

PR #2917, opened 2025-03-03
Fixes for starting slices after host reboot

pykello:
## Execute the block passed to Sshable.start_fresh_session

In `Prog::Vm::VmHostSliceNexus.available?` we relied on the block passed to `Sshable.start_fresh_session` being executed. But it wasn't executed previously.

This change fixes that by passing it to `Net::SSH.start` which accepts a block:

```
.start(host, user = nil, options = {}, &block)
```

## Start slices after host reboot.

We need to start slices after host reboot. Otherwise, `cpuset.cpus.partition` might not be in the correct state.

### DIFF
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -128,8 +128,8 @@ class Sshable < Sequel::Model
     sess
   end
 
-  def start_fresh_session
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
This makes sense to me.

macieksarnowicz:
Looks good to me.

PR #2919, opened 2025-03-03
Support showing separating reasoning content from vllm

0xjunhao:
vLLM introduced a new feature that adds a reasoning_content field to its outputs, containing the reasoning steps leading to the final conclusion. This PR includes the necessary Ubicloud frontend changes to support this feature.


0xjunhao:
Current vllm backend still works:
Screenshot 2025-03-03 at 7 21 07 PM
Screenshot 2025-03-03 at 7 21 12 PM

When we use content as synthetic reasoning_content, the results are as expected:
Screenshot 2025-03-03 at 7 21 30 PM
Screenshot 2025-03-03 at 7 21 38 PM

PR #2923, opened 2025-03-04
PostgreSQL Failover improvements

byucesoy:
These are several failover related improvements that I was planning to implement for a while.


jeremyevans (approved):
Looks good to me. I think it would be good to also get an approval from an engineer more familiar with this code.

PR #2925, opened 2025-03-04
K8s Billing

aamederen:
We were not dealing with billing records for kubernetes clusters, as
    we were in free early-access stage. But that will eventually change
    and we will want to charge for kubernetes clusters.

This change prepares for that day by incorporating logic to create and
    delete billing records into cluster lifecycle.

Billing records are created right after the cluster is ready to use and
    get finalized right after a customer decides to delete it.

All billing records are handled in KubernetesClusterNexus. Alternatively,
    nodepool nexus could handle the nodepool billing, but that's not a big deal.

Currently, all resources are $0 but that will change once we go public-beta.

We still need to finalize LB management via CCM and start managing resources
    in an internal project. After that, we can fully start billing for k8s.

We only have create and delete steps, and no update to billing records. When
    update logic for a cluster is implemented (add node, change SKU, etc.), we would
    need to remember implementing billing logic for these actions. Also, we should
    remember to update billing records if/when we do manual scaling for a customer with
    support ticket.


config/billing_rates.yml
```diff
@@ -92,6 +92,14 @@
 - { id: a3abe656-6ca2-4f17-a911-169c51a4a77a, resource_type: GitHubRunnerMinutes,     resource_family: standard-60-arm,     location: global,         unit_price: 0.0240000000, billed_by: amount,   active_from: 2023-01-01T00:00:00Z }
 - { id: 27d84309-e39d-8d78-b603-7632dafe374e, resource_type: GitHubRunnerMinutes,     resource_family: standard-gpu-6,      location: global,         unit_price: 0.0320000000, billed_by: amount,   active_from: 2023-01-01T00:00:00Z }
 - { id: c31b3640-92fa-49c3-bae8-c1011cc5f917, resource_type: GitHubRunnerConcurrency, resource_family: standard,            location: global,         unit_price: 0.0000248016, billed_by: duration, active_from: 2023-01-01T00:00:00Z }
+- { id: b1bf0809-d8b4-47b0-9a2a-c50f6037b4ca, resource_type: KubernetesCPVCpu,        resource_family: standard,            location: hetzner-fsn1,   unit_price: 0.0000000000, billed_by: duration, active_from: 2025-03-01T00:00:00Z }
```
enescakir, line 95: IMHO, there's no need to use abbreviations. It makes less readable

```suggestion
- { id: b1bf0809-d8b4-47b0-9a2a-c50f6037b4ca, resource_type: KubernetesControlPlaneVCpu,        resource_family: standard,            location: hetzner-fsn1,   unit_price: 0.0000000000, billed_by: duration, active_from: 2025-03-01T00:00:00Z }
```

fdr:
Doesn't kubernetes still lack a service project, i.e. the customer can break it, use regular VM delete calls, etcetera? If it's all folded into their resources (presumed to be solved at some later time), then wouldn't VM billing pick it up?

aamederen:
> Doesn't kubernetes still lack a service project, i.e. the customer can break it, use regular VM delete calls, etcetera? If it's all folded into their resources (presumed to be solved at some later time), then wouldn't VM billing pick it up?

@fdr you're right, this is why these rates are $0

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -9,6 +9,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
   one_to_many :nodepools, class: :KubernetesNodepool
+  one_to_many :active_billing_records, class: :BillingRecord, key: :resource_id do |ds| ds.active end
```
jeremyevans, line 12: Could use `&:active` instead of the literal block.

config/billing_rates.yml
aamederen: KubernetesControlPlaneVCpu is long, causing table spacing changes in the whole file, so I went with the short version. But since you and Burak liked it, I am making the change.

mohi-kalantari (approved):

PR #2933, opened 2025-03-06
Fix regeneration of pg_auto_constraint_validations.cache

jeremyevans:
Delete the pg_auto_constraint_validations.cache file before regenerating it, so it will pick up changes to existing tables and remove no longer used tables.

This regenerates the cache, but the only changes were removing the constraint metadata for the provider and provider_location tables.  provider_location was never used in production, it was proposed in PR #2806, but changed to location before that was merged.  provider does exist as a table, but not as a model, so metadata is not needed for it.


jeremyevans:
This pointed out an issue in Sequel, which is that the hash order inside the cache/pg_auto_constraint_validations.cache file is non-deterministic.  I'm testing a fix for that, I'll push it after it lands in Sequel.

fdr (approved):

PR #2937, opened 2025-03-06
Simplify current invoice view

enescakir:
We will add VAT details to the invoices. I simplified the current invoice view, so we don't have to apply the same VAT changes in both places.

One of the biggest simplifications for invoices is removing Invoice serializer, but it requires too much effort for now.

- **Add footer to the table card component**
  

- **Show finalized invoices as PDF**
  We used to display both current and finalized invoices in HTML format,
  allowing customers to convert HTML to PDF. Now, we generate PDF invoices
  directly using pure Ruby, so there's no need to show finalized invoices
  in HTML anymore. This change reduces maintenance since we no longer have
  to update both HTML and PDF views for invoice updates. The view now only
  shows current usage so no need to show billing details.
  

- **Use the table component in the current usage view**
  Since it's only for current usage, we can simplify it by using the table
  card component.

### DIFF
--- a/views/project/invoice.erb
+++ b/views/project/invoice.erb
@@ -1,4 +1,4 @@
-<% @page_title = "#{@invoice_data[:name]} Invoice" %>
+<% @page_title = "Current Usage Summary" %>
 
 <%== part(
   "components/page_header",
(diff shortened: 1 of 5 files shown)


views/project/invoice.erb
```diff
@@ -1,88 +1,55 @@
...
+             item[:name],
+             item[:description],
+             [item[:usage], { extra_class: "text-right"}],
+             [item[:cost_humanized], { extra_class: "text-right"}],
+           ],
+           { id: "item-#{item[:name]}" }
+         ]
+      end,
+      empty_state: "No resources",
+      footer: '<div class="px-4 py-5 sm:p-6 flex justify-end"><div class="grid gap-2 text-right">' + footer_content + "</div></div>"
```
jeremyevans, line 52: I recommend a double-quoted string and `#{footer_content}` interpolation here (not a big deal, just a bit more efficient).

jeremyevans (approved):

PR #2938, opened 2025-03-07
Set workflow job permissions explicitly

enescakir:
- **Set workflow job permissions explicitly**
  CodeQL has started scanning GitHub Actions workflows as well. It's not a
  major issue, but it's good to follow best practices.
  
  https://github.com/ubicloud/ubicloud/security/code-scanning/11
  
      Workflow does not contain permissions
  
      If a GitHub Actions job or workflow has no explicit permissions set,
      then the repository permissions are used. Repositories created under
      organizations inherit the organization permissions. The
      organizations or repositories created before February 2023 have the
      default permissions set to read-write. Often these permissions do
      not adhere to the principle of least privilege and can be reduced to
      read-only, leaving the write permission only to a specific types as
      issues: write or pull-requests: write.
  

- **Add workflow_dispatch trigger to the workflow jobs**
  I often had to create draft PRs just to trigger the workflow jobs. To
  avoid this, I added a `workflow_dispatch` trigger, allowing us to
  trigger the workflow jobs directly from the GitHub Actions UI for any
  branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,8 @@
 name: Build
 
+permissions:
+  contents: read
+
 on:
   workflow_dispatch:
     inputs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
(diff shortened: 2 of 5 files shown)


enescakir:
| Before | After |
|--------|-------|
| Before | After |

aamederen (approved):

PR #2943, opened 2025-03-08
Expand destroy guard of VM Nexus to all 3 states

aamederen:
I managed to stuck a few Nexus progs in dev env with double `incr_destroy` calls to vms, then realized that we prevent the double delete cycle for the initial `destroy` label, but not for the following labels.


enescakir (approved):

enescakir:
I have a more general solution, but I need to work on the PR a bit more.
https://github.com/ubicloud/ubicloud/pull/2722

PR #2946, opened 2025-03-10
Use UBIDs in Kubernetes Node Names

aamederen:
Previously, we gave names like "clustername-control-plane-something" to k8s cluster nodes in CP side and let them default to VM inhost names in the DP side.

This commit changes the node names to cluster and nodepool UBID based strings in CP and DP. The current format is something like `{(cluster|nodepool).ubid}-{r4nd0}`.

Alternatively, we can provide VM ubids as names to k8s and use them. However, associating VMs with cluster or nodepools in the names seemed more intuitive to me.

One technicality I couldn't figure out was the node name setting during init phase. I initially tried supplying the node name through `nodeRegistration.name` in the kubeadm-config.yaml, but couldn't make it work.


aamederen:
Here's how they look now
CP:
image
DP:
image

enescakir (approved):

PR #2947, opened 2025-03-10
Remove unnecessary join table for some associations

jeremyevans:
Having the associated model's or current model's table as a join table is almost always a bad idea unless the table contains self-referential foreign keys.

While here, use the :order option instead of an association block, and remove unnecessary association options.


enescakir (approved):

enescakir:
You can ignore the review from Copilot; I'm just curious about how it works

PR #2963, opened 2025-03-11
Attempt to avoid the 2 most common production Sequel unhandled exceptions

jeremyevans:
This attempts to avoid two cases that raise `Sequel::NoExistingObject` in race conditions.  These cases are similar:

1. Retrieve object in route
2. Concurrent deletion of object (presumably by respirate)
3. Attempt to update object in route, raises as object no longer exists

In one case, this switches from Model#update to Dataset#update to implicitly ignore the case where the object was deleted.  In the other case, this uses a transaction with FOR UPDATE to ensure the object is not deleted between when it was retrieved and when it was updated.

### DIFF
--- a/model/github/github_cache_entry.rb
+++ b/model/github/github_cache_entry.rb
@@ -5,12 +5,31 @@ require_relative "../../model"
 require "aws-sdk-s3"
 
 class GithubCacheEntry < Sequel::Model
+  plugin :instance_filters
+
   many_to_one :repository, key: :repository_id, class: :GithubRepository
 
   include ResourceMethods
 
+  dataset_module do
+    def destroy_where(...)
+      all do |entry|
+        entry.destroy_where(...)
+      end
+    end
+  end
+
   def blob_key
     "cache/#{ubid}"
+  end
+
+  def destroy_where(...)
+    instance_filter(...)
+    db.transaction(savepoint: true) do
+      destroy
+    end
+  rescue Sequel::NoExistingObject
+    # DELETE query modified no rows due to filter
   end
 
   def after_destroy
--- a/model/github/github_cache_entry.rb
+++ b/model/github/github_cache_entry.rb
@@ -12,9 +12,11 @@ class GithubCacheEntry < Sequel::Model
   include ResourceMethods
 
   dataset_module do
-    def destroy_where(...)
+    # The same as Dataset#destroy, except that it adds the given condition
+    # to the WHERE clause of the DELETE statement for each destroyed row.
+    def destroy_where(cond)
       all do |entry|
-        entry.destroy_where(...)
+        entry.destroy_where(cond)
       end
     end
   end
@@ -23,13 +25,13 @@ class GithubCacheEntry < Sequel::Model
     "cache/#{ubid}"
   end
 
-  def destroy_where(...)
-    instance_filter(...)
+  def destroy_where(cond)
+    instance_filter(cond)
     db.transaction(savepoint: true) do
       destroy
     end
   rescue Sequel::NoExistingObject
-    # DELETE query modified no rows due to filter
+    # DELETE statement modified no rows due to filter
   end
 
   def after_destroy
--- a/prog/github/github_repository_nexus.rb
+++ b/prog/github/github_repository_nexus.rb
@@ -86,30 +86,30 @@ class Prog::Github::GithubRepositoryNexus < Prog::Base
     # Destroy cache entries not accessed in last 7 days or
     # created more than 7 days ago and not accessed yet.
     seven_days_ago = Time.now - 7 * 24 * 60 * 60
+    cond = Sequel.expr { (last_accessed_at < seven_days_ago) | ((last_accessed_at =~ nil) & (created_at < seven_days_ago)) }
     github_repository.cache_entries_dataset
-      .where { (last_accessed_at < seven_days_ago) | ((last_accessed_at =~ nil) & (created_at < seven_days_ago)) }
+      .where(cond)
       .limit(200)
-      .for_update
-      .destroy
+      .destroy_where(cond)
 
     # Destroy cache entries if it is created 30 minutes ago
     # but couldn't committed yet. 30 minutes decided as during
     # our performance tests uploading 10GB of data (which is
     # the max size for a single cache entry) takes ~8 minutes at most.
     # To be on the safe side, ~2x buffer is added.
+    cond = Sequel.expr(committed_at: nil)
     github_repository.cache_entries_dataset
       .where { created_at < Time.now - 30 * 60 }
-      .where { committed_at =~ nil }
+      .where(cond)
       .limit(200)
-      .for_update
-      .destroy
+      .destroy_where(cond)
 
     # Destroy oldest cache entries if the total usage exceeds the limit.
     dataset = github_repository.cache_entries_dataset.exclude(size: nil)
     total_usage = dataset.sum(:size).to_i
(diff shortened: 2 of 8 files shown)


routes/runtime/github.rb
```diff
@@ -48,7 +48,8 @@ class Clover
 
       fail CloverError.new(204, "NotFound", "No cache entry") if entry.nil?
 
-      entry.update(last_accessed_at: Time.now, last_accessed_by: runner.id)
+      # Entry may no longer exist, use this.update to avoid raising an error in that case
+      entry.this.update(last_accessed_at: Sequel::CURRENT_TIMESTAMP, last_accessed_by: runner.id)
```
enescakir, line 52: If it's already deleted, we return an invalid presigned URL since it silently ignores it. We should return a 204 in this case. 

If `.update` returns the number of updated rows, we can check that to make a decision.

routes/webhook/github.rb
```diff
@@ -79,26 +79,28 @@ def handle_workflow_job(data)
       return error("A workflow_job without runner_id")
     end
 
-    runner = GithubRunner.first(
-      installation_id: installation.id,
-      repository_name: data["repository"]["full_name"],
-      runner_id: runner_id
-    )
+    DB.transaction do
+      runner = GithubRunner.for_update.first(
```
enescakir, line 83: It's one of our busiest endpoints. Since the operation in the transaction is very quick, `for_update` doesn't increase its latency, right?

enescakir: I just wanted to make sure.

routes/runtime/github.rb
jeremyevans: `update` does return the number of updated rows.

I'm not sure if the `invalid presigned URL` is actually an issue. Without this change, the model instance could be deleted before the route returns, after the update.  What is the difference in the presigned URL between cases where the entry is deleted after the update as opposed to before?

If returning 204 fixes it, I'm fine with that approach, but then testing the conditional would be difficult.  We could just add `:nocov:`, but I'd prefer not to do that unless we are sure we should treat the cases differently.

enescakir: Currently, when we encounter this case, we return an HTTP 500 error, and the client prints a warning about the unhandled exception.

With this PR, we no longer have an unhandled exception, but we are returning a presigned URL for a blob that has already been deleted. The client will attempt to download it, fail, and print again a warning about the failure.

The 204 status code is special for GitHub's cache client, and it doesn't print a warning for it. When we encounter this case, the cache entry no longer exists, so a 204 response is more appropriate for the situation.

What do you think about this approach?
```ruby
    if entry.this.update(last_accessed_at: Sequel::CURRENT_TIMESTAMP, last_accessed_by: runner.id) == 0
        fail CloverError.new(204, "NotFound", "No cache entry")
    end
```

OR

```ruby
    begin
        entry.update(last_accessed_at: Time.now, last_accessed_by: runner.id)
    rescue Sequel::NoExistingObject
        fail CloverError.new(204, "NotFound", "No cache entry")
    end
```

routes/webhook/github.rb
enescakir: > Is there a reason to create a semaphore for a strand that doesn't exist?

No. Also, it should be prevented by a foreign key.

https://github.com/ubicloud/ubicloud/blob/ac85e59d9bffc0e5932278c2a0edd87b627f2083/model/semaphore.rb#L25

routes/runtime/github.rb
jeremyevans: That approach makes sense to me.  I originally thought cache entry could be deleted after the UPDATE but before we return a result to the user, but since it updates last_accessed_at, I'm guessing that is not possible.

I'll work on this tomorrow.

routes/webhook/github.rb
jeremyevans: I came to the same conclusion as you.  I submitted a separate PR to address this issue: #2971

routes/runtime/github.rb
jeremyevans: Pushed a couple more commits.  One uses 204 in this case, with a slight restructuring to allow for full branch coverage.  While looking into this, I found what I think is a related race condition in the prog, so the other commit is a fix for that.

prog/github/github_repository_nexus.rb
```diff
@@ -89,6 +89,7 @@ def cleanup_cache
     github_repository.cache_entries_dataset
       .where { (last_accessed_at < seven_days_ago) | ((last_accessed_at =~ nil) & (created_at < seven_days_ago)) }
       .limit(200)
+      .for_update
```
enescakir, line 92: We delete objects from blob storage when we delete them, so this operation takes some time. If we lock the cache entries, it will affect the latencies of the cache endpoint, right?

Since you changed the update operation in the previous commit, it no longer fails with a 'no existing object' exception. Do we still need this lock?

jeremyevans: It can affect the latency, so if this is performance sensitive, we can take a different approach.

I think some approach to handling concurrent updates is needed, though.  Because I believe currently:

* Thread A (prog) SELECT entry with last_access_time over a week ago
* Thread B (runtime route) UPDATE entry set last_access_time = CURRENT_TIMESTAMP
* Thread A (prog) DELETE entry # should not be deleted as last_access_time is not the same as select criteria

If we don't want a lock, then we need a filter on the first DELETE to also ensure we don't delete an entry whose last access time has been updated since the SELECT.  For the second DELETE, we would want a DELETE filter to ensure the entry is still not committed.  For the final DELETE, I guess we should just allow the delete even for concurrent updates.  Does that make sense to you?

model/github/github_cache_entry.rb
```diff
@@ -5,14 +5,33 @@
...
 
   include ResourceMethods
 
+  dataset_module do
+    def destroy_where(...)
+      all do |entry|
+        entry.destroy_where(...)
+      end
+    end
+  end
```
furkansahin, line 20: These 2 code blocks look like dark magic to me. Can you explain what do these do for educational purposes as a comment? I believe it won't be clear to anyone in the team either.

jeremyevans: Sure, I can add a comment.  Basically:

* `dataset_module` defines a module included in the model's dataset (this is used by a number of existing models already).
* `destroy_where(...)` defines a method (a dataset method in this case), that accepts any number of arguments.
* `all do |entry|` retrieves all entries in the dataset and iterates over all of them
* `entry.destroy_where(...)` calls the `destroy_where` method on the model instance, with the same arguments passed to the dataset method.
* Putting everything together, this is the same as the existing `Dataset#destroy` method, but it allows you to add an additional filter on the `DELETE` statement.

jeremyevans: I pushed another commit.  This drops the usage of `...` (generic argument forwarding), since the callers currently only pass a single argument.  I think that was likely the main source of confusion.  I also added a comment describing what the method does.

jeremyevans:
Rebased to fix conflicts.

jeremyevans:
@enescakir Could you please give this another review?

routes/webhook/github.rb
enescakir: I thought we agreed to use `Dataset#update` instead of `for_update` because of the latency penalty.

jeremyevans: We did, sorry I missed that when I removed the three other cases of `for_update`.  I pushed another commit that removes it from the webhook route.

enescakir (approved):
Thanks for these improvements. I triggered E2E tests. We can merge it after it's succeeded

jeremyevans:
E2E tests passed: https://github.com/ubicloud/ubicloud/actions/runs/14114306299

PR #2965, opened 2025-03-12
Disconnect Sequel::Database instance before fork

jeremyevans:
Without this, we are getting exceptions in production shortly after deployment due to multiple processes sharing database connections.

Update bin/production_check to try to use a clustered, preloaded puma, since I assume that is what we must be running in production.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
-web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-development}
+web: bundle exec puma -C puma_config.rb
 respirate: bin/respirate
 monitor: bin/monitor
--- a/puma_config.rb
+++ b/puma_config.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+# :nocov:
+environment ENV["RACK_ENV"] || "development"
+port ENV["PORT"] || "3000"
+threads 5, 5
+
+if @config.options[:workers] > 0
+  silence_single_worker_warning
+  preload_app!
+
+  before_fork do
+    Sequel::DATABASES.each(&:disconnect)
+  end
+end
+# :nocov:
--- a/bin/production_check
+++ b/bin/production_check
@@ -4,12 +4,14 @@
 Dir.chdir(File.dirname(__dir__))
 
 require "open3"
+require "net/http"
+require "json"
 
 works = nil
 error = +""
 queue = Queue.new
 
-Open3.popen3({"RACK_ENV" => "production"}, "bundle", "exec", "puma", "-p", "8081") do |stdin, stdout, stderr, wait_thr|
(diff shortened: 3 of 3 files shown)


Procfile
```diff
@@ -1,3 +1,3 @@
-web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-development}
+web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-development} -C puma_config.rb
```
enescakir, line 1: What do you think about moving all the puma related configurations to the puma_config.rb file? Right now, we have some command line configs, and production has the `WEB_CONCURRENCY` environment variable. They’re a bit scattered everywhere.

```ruby
environment ENV["RACK_ENV"] || "development"
port ENV["PORT"] || "3000"
threads 5, 5
workers ENV["WEB_CONCURRENCY"] || 0
preload_app!
```

jeremyevans: That makes sense to me.

fdr:
My question is: why do we connect to begin with? Maybe connecting, then disconnecting, is most practical...but it seems weird to create connections before doing a single thing.

puma_config.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# :nocov:
+environment ENV["RACK_ENV"] || "development"
+port ENV["PORT"] || "3000"
+threads 5, 5
+
```
enescakir, line 7: What do you think about something like this?

```ruby
workers ENV["RACK_ENV"] == "production" ? 2 : 0
```

Currently, we have 2 puma workers in production. However, this information is hidden from engineers unless they log in to the Heroku console and check the `WEB_CONCURRENCY` environment variable. It's likely that no one has checked it if they didn't know about it before. Since it's not our environment variable and isn't available in the codebase. We can make it more obvious.

jeremyevans: I'm not opposed to it, but I think it would break user exceptions for external users running Ubicloud. "Why is my WEB_CONCURRENCY setting not respected?"

However, we could check whether `WEB_CONCURRENCY` is set, and only call the `workers` method if it is not set.

jeremyevans: After some testing, I've found that if we want to default to 2 workers in production mode, while still respecting WEB_CONCURRENCY:

```ruby
if ENV["RACK_ENV"] == "production" && !ENV["WEB_CONCURRENCY"]
  workers 2
end
```

If you want to avoid respecting WEB_CONCURRENCY environment variable for our production deployment, as it is hidden in Heroku config, we should change the Procfile (`-w` CLI option has precedence over WEB_CONCURRENCY):

```
web: bundle exec puma -w 2 -C puma_config.rb
```

This means that increasing the concurrency requires a code change and not a config variable change.

I don't have strong feelings about this.  The only necessary part of the config file to me is:

```ruby
  before_fork do
    Sequel::DATABASES.each(&:disconnect)
  end
```

@fdr What are your thoughts regarding this?

enescakir (approved):
We can do other improvements in the next PRs too

PR #2966, opened 2025-03-12
Fix reverse-translation issue while using direct nft commands

mohi-kalantari:
~~We were applying nat rules using nft binary right after node bootstrap and in the mean time, kube-proxy was also setting up.~~

~~If we applied our rule first, it was ok but if we did it after kube proxy, since we were not using a standard approach (direct nft rule instead of iptables-nft) it would break the healthcheck of kube proxy (checks the NAT table's POSTROUTING chain and it would fail because iptables working as an interface here, couldn't translate back the current nft rules since they were not in an standard form)~~

~~A new hop is added and we apply our nat rule earlier than bootstrap.~~

~~Also iptables-nft is used instead of nft binary~~

~~kube-proxy relies on the iptables command to make changes to the VM networking.
Based on what config is set for iptables, it would either use legacy or nftables.
Since we are provisioning our VMs on ubuntu 22 and higher, it would choose nftables.~~

On startup, kube-proxy runs this command

iptables -S POSTROUTING -t nat

to make sure everything is correctly setup in ipv4.

We were adding our own NAT rules using the nft binary and it caused issues for
reverse translation during the healthcheck and we would get errors like this one:

iptables v1.8.7 (nf_tables): chain POSTROUTING in table nat is incompatible, use 'nft' tool.

To address that issue, we would use iptables-nft instead to fix the reverse translation

Since kube-proxy does not support direct nft integration and works only using iptables,
we had to change the way we apply our rules.

Also the stage we apply our NAT rules is now before kubeadm init/join, this way we are
sure we've done everything we wanted to do with the node and not cause any race-condition while
kube-proxy is also setting up its rules.

### DIFF
--- a/prog/kubernetes/provision_kubernetes_node.rb
+++ b/prog/kubernetes/provision_kubernetes_node.rb
@@ -92,6 +92,12 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
   sudo systemctl enable --now kubelet
 BASH
 
+    hop_apply_nat_rules
+  end
+
+  label def apply_nat_rules
+    vm.sshable.cmd("sudo iptables-nft -t nat -A POSTROUTING -s #{vm.nics.first.private_ipv4} -o ens3 -j MASQUERADE")
+
     hop_bootstrap_rhizome
   end
 
@@ -207,18 +213,6 @@ BASH
 }
 CONFIG
     vm.sshable.cmd("sudo tee /etc/cni/net.d/ubicni-config.json", stdin: cni_config)
-    vm.sshable.cmd("sudo nft -f -", stdin: <<~BASH)
-      add table ip nat;
-
-      table ip nat {
-        chain POSTROUTING {
-          type nat hook postrouting priority 100;
-          policy accept;
-          ip saddr #{vm.nics.first.private_ipv4} oifname ens3 masquerade;
-        }
-      }
-    BASH
-
     pop vm_id: vm.id
   end
 end
--- a/spec/prog/kubernetes/provision_kubernetes_node_spec.rb
+++ b/spec/prog/kubernetes/provision_kubernetes_node_spec.rb
@@ -150,7 +150,16 @@ RSpec.describe Prog::Kubernetes::ProvisionKubernetesNode do
       expect(prog.vm).to receive(:strand).and_return(st)
       expect(prog.vm).to receive(:sshable).and_return(sshable)
       expect(sshable).to receive(:cmd).with(/sudo apt install -y kubelet kubeadm kubectl/)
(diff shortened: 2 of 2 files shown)


aamederen (approved):

fdr:
> (direct nft rule instead of iptables-nft)

...the standard approach [at least, if you were to ask the people who write nftables], when using nftables, is to use nft, though. It all gets translated anyway. How does this really work?

mohi-kalantari:
> > (direct nft rule instead of iptables-nft)
> 
> ...the standard approach [at least, if you were to ask the people who write nftables], when using nftables, is to use nft, though. It all gets translated anyway. How does this really work?

Kubeproxy relies on iptables which its default engine on ubuntu 22 and higher is nftables. as a check at startup, it runs a `iptables -S POSTROUTING -t nat` to make sure everything is working correctly

Since we use nft directly for adding the nat rule, the rules we apply is not standard enough to get translated back to results iptables expect and we would get this error for that command:

`iptables v1.8.7 (nf_tables): chain POSTROUTING' in table nat is incompatible, use 'nft' tool.`

And it would crash.

Instead by relying on iptables-nft, we get the same result but with a slightly different set of rules that can get translated back to iptables desired schema.

We can find the exact rules iptables-nft rules apply but IMO it's more standard to just use the interface kubeproxy will interact during the life time of that node.

@fdr

fdr (changes requested):
you should, if not electing to use an nft native backend for kube-proxy, change the commit message to describe the actual problem (nft that cannot be reverse-translated causes kube-proxy to crash when using the most common iptables mode). Justifying something as "strandard" is effectively anti-documentation, meaning almost nothing at all unless referring to a specification.

For bonus points, add rationale as to the theory of why we should suffer this degradation instead of using the nftables mode.

mohi-kalantari:
Unfortunately the kube proxy does not support direct nft rules and it would just rely on the old interface, trusting it would do the translation and has the right backend. Otherwise we would have definitely gone that way.

I'll update the commit message and add the right info. @fdr

mohi-kalantari:
commit message is updated.

I also re-read one of your comments and I have to add, at the end, we are still using the nftables, but the API to configure the rules is not nft, instead we have to use iptables. so there is no degradation.

fdr (approved):
much better, although it's a little sad that we have to abide such a restriction.

fdr:
> commit message is updated.
> 
> I also re-read one of your comments and I have to add, at the end, we are still using the nftables, but the API to configure the rules is not nft, instead we have to use iptables. so there is no degradation.

well, I consider being obliged/restricted to iptables notation a degradation (I much prefer `nft`'s syntax), rather than a functional degradation per se. But a small one.

PR #2967, opened 2025-03-12
Introduce LoadBalancer management functionality of CloudControllerManager

mohi-kalantari:
For kubernetes clusters, we will watch the Services with type set to 'LoadBalancer' and create corresponding Ubicloud loadbalancers and also react to port/vm removals or additions.

### DIFF
--- a/lib/kubernetes/client.rb
+++ b/lib/kubernetes/client.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+class Kubernetes::Client
+  def initialize(kubernetes_cluster, session)
+    @session = session
+    @kubernetes_cluster = kubernetes_cluster
+    @load_balancer = LoadBalancer.where(name: kubernetes_cluster.services_load_balancer_name).first
+  end
+
+  def service_deleted?(svc)
+    !!svc.dig("metadata", "deletionTimestamp")
+  end
+
+  # Returns a flat array of [port, nodePort] pairs from all services
+  # Format: [[src_port_0, dst_port_0], [src_port_1, dst_port_1], ...]
+  def lb_desired_ports(svc_list)
+    svc_list.flat_map do |svc|
+      svc.dig("spec", "ports")&.map { |port| [port["port"], port["nodePort"]] } || []
+    end
+  end
+
+  def load_balancer_hostname_missing?(svc)
+    svc.dig("status", "loadBalancer", "ingress")&.first&.dig("hostname").to_s.empty?
+  end
+
+  def kubectl(cmd)
+    @session.exec!("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
+  end
+
+  def set_load_balancer_hostname(svc, hostname)
+    patch_data = JSON.generate({
+      "status" => {
+        "loadBalancer" => {
+          "ingress" => [{"hostname" => hostname}]
+        }
+      }
+    })
+    kubectl("-n #{svc.dig("metadata", "namespace")} patch service #{svc.dig("metadata", "name")} --type=merge -p '#{patch_data}' --subresource=status")
+  end
+
+  def sync_kubernetes_services
+    k8s_svc_raw = kubectl("get service --all-namespaces --field-selector spec.type=LoadBalancer -ojson")
+    svc_list = JSON.parse(k8s_svc_raw)["items"]
+
+    if @load_balancer.nil?
+      raise "services load balancer does not exist."
+    end
+
+    extra_vms, missing_vms = @kubernetes_cluster.vm_diff_for_lb(@load_balancer)
+    missing_vms.each { |missing_vm| @load_balancer.add_vm(missing_vm) }
+    extra_vms.each { |extra_vm| @load_balancer.detach_vm(extra_vm) }
+
+    extra_ports, missing_ports = @kubernetes_cluster.port_diff_for_lb(@load_balancer, lb_desired_ports(svc_list))
+    extra_ports.each { |port| @load_balancer.remove_port(port) }
+    missing_ports.each { |port| @load_balancer.add_port(port[0], port[1]) }
+
+    return unless @load_balancer.strand.label == "wait"
+    svc_list.each { |svc| set_load_balancer_hostname(svc, @load_balancer.hostname) }
+  end
+
+  def any_lb_services_modified?
+    k8s_svc_raw = kubectl("get service --all-namespaces --field-selector spec.type=LoadBalancer -ojson")
+    svc_list = JSON.parse(k8s_svc_raw)["items"]
+
+    return true if svc_list.empty? && !@load_balancer.ports.empty?
+
+    extra_vms, missing_vms = @kubernetes_cluster.vm_diff_for_lb(@load_balancer)
+    return true unless extra_vms.empty? && missing_vms.empty?
+
+    extra_ports, missing_ports = @kubernetes_cluster.port_diff_for_lb(@load_balancer, lb_desired_ports(svc_list))
+    return true unless extra_ports.empty? && missing_ports.empty?
+
+    svc_list.any? { |svc| load_balancer_hostname_missing?(svc) }
+  end
+end
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -16,8 +16,9 @@ class KubernetesCluster < Sequel::Model
 
   include ResourceMethods
   include SemaphoreMethods
+  include HealthMonitorMethods
 
-  semaphore :destroy
+  semaphore :destroy, :sync_kubernetes_services
 
   def validate
     super
@@ -43,8 +44,20 @@ class KubernetesCluster < Sequel::Model
     api_server_lb.hostname
   end
 
+  def client(session: cp_vms.first.sshable.start_fresh_session)
+    Kubernetes::Client.new(self, session)
+  end
+
   def sshable
     cp_vms.first.sshable
+  end
+
+  def services_load_balancer_name
+    "#{ubid}-services"
+  end
+
+  def apiserver_load_balancer_name
+    "#{ubid}-apiserver"
   end
 
   def self.kubeconfig(vm)
@@ -61,6 +74,41 @@ class KubernetesCluster < Sequel::Model
 
   def kubeconfig
     self.class.kubeconfig(cp_vms.first)
+  end
+
+  def vm_diff_for_lb(load_balancer)
+    worker_vms = nodepools.flat_map(&:vms)
+    worker_vm_ids = worker_vms.map(&:id).to_set
+    lb_vms = load_balancer.load_balancers_vms.map(&:vm)
+    lb_vm_ids = lb_vms.map(&:id).to_set
+
+    extra_vms = lb_vms.reject { |vm| worker_vm_ids.include?(vm.id) }
+    missing_vms = worker_vms.reject { |vm| lb_vm_ids.include?(vm.id) }
+    [extra_vms, missing_vms]
+  end
+
+  def port_diff_for_lb(load_balancer, desired_ports)
+    lb_ports_hash = load_balancer.ports.to_h { |p| [[p.src_port, p.dst_port], p.id] }
+    missing_ports = desired_ports - lb_ports_hash.keys
+    extra_ports = (lb_ports_hash.keys - desired_ports).map { |p| LoadBalancerPort[id: lb_ports_hash[p]] }
+
+    [extra_ports, missing_ports]
+  end
+
+  def init_health_monitor_session
+    {
+      ssh_session: sshable.start_fresh_session
+    }
+  end
+
+  def check_pulse(session:, previous_pulse:)
+    reading = begin
+      incr_sync_kubernetes_services if client(session: session[:ssh_session]).any_lb_services_modified?
+      "up"
+    rescue
+      "down"
+    end
+    aggregate_readings(previous_pulse: previous_pulse, reading: reading)
   end
 end
 
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -40,11 +40,30 @@ class LoadBalancer < Sequel::Model
     vm_ports_dataset.where(load_balancer_vm_id: load_balancers_vms_dataset.where(vm_id: vm.id).select(:id), state:)
   end
 
+  def add_port(src_port, dst_port)
+    DB.transaction do
+      port = LoadBalancerPort.create(load_balancer_id: id, src_port:, dst_port:)
+      load_balancers_vms.each do |lb_vm|
+        LoadBalancerVmPort.create(load_balancer_port_id: port.id, load_balancer_vm_id: lb_vm.id)
+      end
+      incr_update_load_balancer
+    end
+  end
+
+  def remove_port(port)
+    DB.transaction do
+      vm_ports_dataset.where(load_balancer_port_id: port.id).destroy
+      port.destroy
+      incr_update_load_balancer
+    end
+  end
+
   def add_vm(vm)
     DB.transaction do
       load_balancer_vm = LoadBalancersVms.create(load_balancer_id: id, vm_id: vm.id)
-      ports.each { |port| LoadBalancerVmPort.create(load_balancer_port_id: port.id, load_balancer_vm_id: load_balancer_vm.id) }
-
+      ports.each { |port|
+        LoadBalancerVmPort.create(load_balancer_port_id: port.id, load_balancer_vm_id: load_balancer_vm.id)
+      }
       Strand.create_with_id(prog: "Vnet::CertServer", label: "put_certificate", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
       incr_rewrite_dns_records
     end
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -58,9 +58,9 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
     custom_hostname_prefix = if custom_hostname_dns_zone_id
       "#{kubernetes_cluster.name}-apiserver-#{kubernetes_cluster.ubid.to_s[-5...]}"
     end
-    load_balancer_st = Prog::Vnet::LoadBalancerNexus.assemble(
+    load_balancer = Prog::Vnet::LoadBalancerNexus.assemble(
       kubernetes_cluster.private_subnet_id,
-      name: "#{kubernetes_cluster.name}-apiserver",
+      name: kubernetes_cluster.apiserver_load_balancer_name,
       algorithm: "hash_based",
       src_port: 443,
       dst_port: 6443,
@@ -69,8 +69,8 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
       custom_hostname_dns_zone_id:,
       custom_hostname_prefix:,
       stack: LoadBalancer::Stack::IPV4
-    )
-    kubernetes_cluster.update(api_server_lb_id: load_balancer_st.id)
+    ).subject
+    kubernetes_cluster.update(api_server_lb_id: load_balancer.id)
 
     hop_bootstrap_control_plane_vms
   end
@@ -120,7 +120,17 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
   end
 
   label def wait
+    when_sync_kubernetes_services_set? do
+      hop_sync_kubernetes_services
+    end
     nap 6 * 60 * 60
+  end
+
+  label def sync_kubernetes_services
+    decr_sync_kubernetes_services
+    # TODO: timeout or other logic to avoid apoptosis should be added
+    kubernetes_cluster.client.sync_kubernetes_services
+    hop_wait
   end
 
   label def destroy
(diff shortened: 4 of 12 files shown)


prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -91,7 +118,11 @@ def before_run
   end
 
   label def wait
-    nap 65536
+    k8s_svc_raw = kubernetes_cluster.sshable.cmd("sudo kubectl --kubeconfig /etc/kubernetes/admin.conf get service --all-namespaces --field-selector spec.type=LoadBalancer -ojson")
```
aamederen, line 121: nit: we can move this to another function for readability and consistency with other progs.

aamederen, line 121: thinking about a bit more, can we make a change such that `monitor` actually monitors the services and increments a semaphore on the cluster nexus when a change is needed? This would be more in-line with the clover design.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -45,6 +48,14 @@ def sshable
     cp_vms.first.sshable
   end
 
+  def kubectl(cmd)
+    cp_vms.first.sshable.cmd("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
+  end
+
+  def kubectl_s(session, cmd)
+    session[:ssh_session].exec!("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
```
aamederen, line 56: It would be nice if we can avoid the duplication. What about something like this:

```ruby
def kubectl(cmd, session: cp_vms.first.sshable.start_fresh_session)
  session.exec!("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
end
```

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
...
+
+  def extra_and_missing_ports_for_lb(lb, desired_ports)
+    lb_ports_hash = lb.ports.to_h { |p| [[p.src_port, p.dst_port], p.id] }
+    missing_ports = desired_ports - lb_ports_hash.keys
+    extra_ports = (lb_ports_hash.keys - desired_ports).map { |p| LoadBalancerPort[id: lb_ports_hash[p]] }
+
+    [extra_ports, missing_ports]
+  end
+
+  def detect_loadbalancer_svc_updates(svc)
```
aamederen, line 120: I didn't like the name. We have both `detect_loadbalancer_svc_updates` and `detect_loadbalancer_svcs_updates` and it's easy to miss the difference. Something like `lb_service_modified?` could be easier to read.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
...
+
+    extra_vms, missing_vms = extra_and_missing_vms_for_lb(lb)
+    extra_ports, missing_ports = extra_and_missing_ports_for_lb(lb, lb_desired_ports(svc))
+
+    hostname_set_on_status = !svc.dig("status", "loadBalancer", "ingress", "hostname").to_s.empty?
+
+    !extra_vms.empty? || !missing_vms.empty? || !extra_ports.empty? || !missing_ports.empty? || !hostname_set_on_status
+  end
+
+  def detect_loadbalancer_svcs_updates(session)
```
aamederen, line 134: Maybe we can give this function a shorter name, ending with a `?`, something like `any_lb_services_modified?`

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
...
+    extra_ports = (lb_ports_hash.keys - desired_ports).map { |p| LoadBalancerPort[id: lb_ports_hash[p]] }
+
+    [extra_ports, missing_ports]
+  end
+
+  def detect_loadbalancer_svc_updates(svc)
+    return true if is_deleted(svc) && svc.dig("metadata", "finalizers")&.include?(SERVICE_FINALIZER)
+
+    lb = LoadBalancer.where(name: lb_name_for_svc).first
+    return true if lb.nil?
```
aamederen, line 124: nit: `return true unless lb`

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
...
+  def extra_and_missing_ports_for_lb(lb, desired_ports)
+    lb_ports_hash = lb.ports.to_h { |p| [[p.src_port, p.dst_port], p.id] }
+    missing_ports = desired_ports - lb_ports_hash.keys
+    extra_ports = (lb_ports_hash.keys - desired_ports).map { |p| LoadBalancerPort[id: lb_ports_hash[p]] }
+
+    [extra_ports, missing_ports]
+  end
+
+  def detect_loadbalancer_svc_updates(svc)
+    return true if is_deleted(svc) && svc.dig("metadata", "finalizers")&.include?(SERVICE_FINALIZER)
```
aamederen, line 121: if we have `is_deleted`, then we should also have `is_finalized` for consistency IMO.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
...
+    ubid
+  end
+
+  # Returning an array in this format:
+  # [[src_port_0, dst_port_0], [src_port_1, dst_port_1],...]
+  def lb_desired_ports(svc)
+    svc["spec"]["ports"].map { |port| [port["port"], port["nodePort"]] }
+  end
+
+  def is_deleted(svc)
```
aamederen, line 97: nit: s/is_deleted/is_service_deleted

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
...
+    svc.dig("metadata", "deletionTimestamp")
+  end
+
+  def extra_and_missing_vms_for_lb(lb)
```
aamederen, line 101: nit: rename to something like `vm_diff_for_lb` (using `diff` instead of `extra and missing`)? This can also apply to `_ports_`.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
...
+
+    lb = LoadBalancer.where(name: lb_name_for_svc).first
+    return true if lb.nil?
+
+    extra_vms, missing_vms = extra_and_missing_vms_for_lb(lb)
+    extra_ports, missing_ports = extra_and_missing_ports_for_lb(lb, lb_desired_ports(svc))
+
+    hostname_set_on_status = !svc.dig("status", "loadBalancer", "ingress", "hostname").to_s.empty?
+
+    !extra_vms.empty? || !missing_vms.empty? || !extra_ports.empty? || !missing_ports.empty? || !hostname_set_on_status
```
aamederen, line 131: instead of chaining clauses, you can exit early at the time of detecting some change.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +71,93 @@ def self.kubeconfig(vm)
   def kubeconfig
     self.class.kubeconfig(cp_vms.first)
   end
+
+  def add_finalizer(svc)
```
aamederen, line 75: nit: rename to `add_svc_finalizer` or something like that.

aamederen (changes requested):
Overall, I really like that, logic-wise, we have a very simple CCM here. I have some suggestions naming and construction wise. Also, I feel bad that we have CCM-specific logic spread across model and nexus. I suggest we collect such logic in a specific class in `lib` folder.

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,123 @@
...
+  def set_load_balancer_host_name(svc, hostname)
+    kubectl("-n #{svc["metadata"]["namespace"]} patch service #{svc["metadata"]["name"]} --type='merge' -p '{\"status\":{\"loadBalancer\":{\"ingress\":[{\"hostname\":\"#{hostname}\"}]}}}' --subresource='status'")
+  end
+
+  def add_svc_finalizer(svc)
+    kubectl("-n #{svc["metadata"]["namespace"]} patch service #{svc["metadata"]["name"]} --type='merge' -p '{\"metadata\":{\"finalizers\":[\"#{SERVICE_FINALIZER}\"]}}'")
+  end
+
+  def remove_svc_finalizer(svc)
+    kubectl("-n #{svc["metadata"]["namespace"]} patch service #{svc["metadata"]["name"]} --type='json' -p='[{\"op\": \"remove\", \"path\": \"/metadata/finalizers\",\"value\":[\"#{SERVICE_FINALIZER}\"]}]'")
```
fdr, line 46: This is a bit "ew", rather than a JSON.generate or a , unless you feel it's much better to have it like this (e.g. to cut-and-paste into the terminal). Seems like it's asking for an ungratifying (although quick to fix) typo.

model/load_balancer.rb
```diff
@@ -40,6 +40,24 @@ def vm_ports_by_vm_and_state(vm, state)
     vm_ports_dataset.where(load_balancer_vm_id: load_balancers_vms_dataset.where(vm_id: vm.id).select(:id), state:)
   end
 
+  def add_port(port_info)
```
enescakir, line 43: Why don't we pass `src_port` and `dst_port` separately?

mohi-kalantari: fixed thanks.

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,136 @@
...
+    extra_ports.each { |port| lb.remove_port(port) }
+    missing_ports.each { |port| lb.add_port(port[0], port[1]) }
+
+    set_load_balancer_host_name(svc, lb.hostname) if lb.strand.label == "wait"
+  end
+
+  def sync_kubernetes_services
+    k8s_svc_raw = kubectl("get service --all-namespaces --field-selector spec.type=LoadBalancer -ojson")
+    JSON.parse(k8s_svc_raw)["items"].each do |svc|
+      reconcile_kubernetes_service(svc)
```
aamederen, line 133: nit: In terms of ordering the functions, it could be more readable to put callers above callees. In this case, sync_kubernetes_services would be on top.

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def reconcile_kubernetes_service_deletion(svc)
+    LoadBalancer.where(name: lb_name_for_svc).first&.incr_destroy
+    remove_svc_finalizer(svc)
+  end
+
+  def reconcile_kubernetes_service(svc)
+    return reconcile_kubernetes_service_deletion(svc) if is_service_deleted(svc)
+    add_svc_finalizer(svc)
```
aamederen, line 109: seems safe, but can there be a side-effect of doing too much redundant patching?

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    name = lb_name_for_svc
+    desired_ports = lb_desired_ports(svc)
+    lb = LoadBalancer.where(name:).first
+    if lb.nil?
```
aamederen, line 114: I think we should error out here if `lb.nil?` and guarantee the LB exists during cluster creation

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,136 @@
...
+class Kubernetes::Client
+  SERVICE_FINALIZER = "k8s.ubicloud.com/load-balancer"
+
+  def initialize(kubernetes_cluster, session)
+    @session = session
+    @kubernetes_cluster = kubernetes_cluster
+  end
+
+  def lb_name_for_svc
+    @kubernetes_cluster.ubid
```
aamederen, line 12: In this model, we have 2 LBs per cluster: one for API Server and one for Apps. I think we should make their names easier to distinguish. Maybe we can follow such naming scheme:
API server: ubid-apiserver
SVC: ubid-services

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+
+class Kubernetes::Client
+  SERVICE_FINALIZER = "k8s.ubicloud.com/load-balancer"
+
+  def initialize(kubernetes_cluster, session)
+    @session = session
+    @kubernetes_cluster = kubernetes_cluster
```
aamederen, line 8: you can also add @lb = LoadBalancer.... here to avoid querying for the LB in the methods

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def any_lb_services_modified?
+    k8s_svc_raw = kubectl("get service --all-namespaces --field-selector spec.type=LoadBalancer -ojson")
+    found = JSON.parse(k8s_svc_raw)["items"].find { |svc| lb_service_modified?(svc) }
+    !!found
+  end
+
+  def reconcile_kubernetes_service_deletion(svc)
+    LoadBalancer.where(name: lb_name_for_svc).first&.incr_destroy
```
aamederen, line 103: I didn't get this part, don't we use a single Lb for multiple services? If so, if the service is dropped, why do we drop the LB?

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -90,6 +96,12 @@ def before_run
     donate
   end
 
+  label def sync_kubernetes_services
+    decr_sync_kubernetes_services
+    kubernetes_cluster.client.sync_kubernetes_services
```
aamederen, line 101: can we put a mechanism here (timeout or number of operations) to limit the execution time so that we avoid possible apoptosis. It's OK to let it go for now but let's put a TODO at least.

aamederen (changes requested):

lib/kubernetes/client.rb
mohi-kalantari: during the cluster creation, the customer does not have access to cluster to create one with type LoadBalancer. this logic only cares about customer created svcs and apiserver is not in included in this logic.

lib/kubernetes/client.rb
mohi-kalantari: no it's ok. the patch takes care of that and won't add extra ones.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -60,6 +65,28 @@ def self.kubeconfig(vm)
...
+
+  def init_health_monitor_session
+    {
+      ssh_session: sshable.start_fresh_session
+    }
+  end
+
+  def check_pulse(session:, previous_pulse:)
+    reading = begin
+      incr_sync_kubernetes_services if strand.label == "wait" && client(session: session[:ssh_session]).any_lb_services_modified?
```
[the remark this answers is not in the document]
mohi-kalantari: IIRC multiple incr_ s would create multiple semaphores. does decr_ ing it once handle all of those?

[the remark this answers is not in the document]
aamederen: yes

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -90,8 +90,18 @@ def before_run
     donate
   end
 
+  label def sync_kubernetes_services
```
aamederen, line 93: nit: this label can be placed below `wait` to match order of reading with order of execution.

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+class Kubernetes::Client
+  def initialize(kubernetes_cluster, session)
+    @session = session
+    @kubernetes_cluster = kubernetes_cluster
+    @load_balancer = LoadBalancer.where(name: kubernetes_cluster.services_load_balancer_name).first
+  end
+
+  def is_service_deleted(svc)
```
aamederen, line 10: nit: `service_deleted?` is a more ruby-like name than `is_service_deleted`

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,103 @@
...
+    lb_ports_hash = @load_balancer.ports.to_h { |p| [[p.src_port, p.dst_port], p.id] }
+    missing_ports = desired_ports - lb_ports_hash.keys
+    extra_ports = (lb_ports_hash.keys - desired_ports).map { |p| LoadBalancerPort[id: lb_ports_hash[p]] }
+
+    [extra_ports, missing_ports]
+  end
+
+  def any_lb_services_modified?
+    k8s_svc_raw = kubectl("get service --all-namespaces --field-selector spec.type=LoadBalancer -ojson")
+    found = JSON.parse(k8s_svc_raw)["items"].find { |svc| lb_service_modified?(svc) }
```
aamederen, line 86: nit: you can use `any?` rather than `find`, so you don't have to use the `found` variable

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,103 @@
...
+    !!found
+  end
+
+  def lb_service_modified?(svc)
+    return true if is_service_deleted(svc) && is_service_finalized(svc)
```
aamederen, line 91: am I missing the `is_service_finalized` function?

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    return true unless @load_balancer
```
aamederen, line 93: this case is a bug now

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,103 @@
...
+    extra_vms.each { |extra_vm| @load_balancer.detach_vm(extra_vm) }
+
+    extra_ports, missing_ports = port_diff_for_lb(desired_ports)
+    extra_ports.each { |port| @load_balancer.remove_port(port) }
+    missing_ports.each { |port| @load_balancer.add_port(port[0], port[1]) }
+
+    set_load_balancer_host_name(svc, @load_balancer.hostname) if @load_balancer.strand.label == "wait"
+  end
+
+  def vm_diff_for_lb
```
aamederen, line 65: this function shouldn't be in the client, as it has nothing to do with Kubernetes itself. We should make sure to handle this in LB control functions.

lib/kubernetes/client.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def load_balancer_hostname_missing?(svc)
+    svc.dig("status", "loadBalancer", "ingress", "hostname").to_s.empty?
+  end
+
+  def kubectl(cmd)
+    @session.exec!("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
+  end
+
+  def set_load_balancer_host_name(svc, hostname)
```
aamederen, line 28: nit: `set_load_balancer_hostname` for consistency

aamederen (changes requested):
A function is missing, that means a re-test is required. Also, note that it's better to merge this PR after the multiple-ports-per-LB (#2718) is merged

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -120,7 +120,17 @@ def before_run
   end
 
   label def wait
-    nap 6 * 60 * 60
+    when_sync_kubernetes_services_set? do
+      hop_sync_kubernetes_services
+    end
+    nap 10
```
aamederen, line 126: no need to change nap duration here

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -26,6 +26,31 @@ def before_run
...
+    custom_hostname_prefix = if custom_hostname_dns_zone_id
+      "#{kubernetes_nodepool.cluster.ubid.to_s[-10...]}-services"
+    end
+    Prog::Vnet::LoadBalancerNexus.assemble(
+      kubernetes_nodepool.cluster.private_subnet_id,
+      name: kubernetes_nodepool.cluster.services_load_balancer_name,
+      algorithm: "hash_based",
+      # The next two fields will be later modified by the sync_kubernetes_services label
+      # These are just set for passing the creation validations
+      src_port: 443,
```
aamederen, line 45: Do we have a TODO for allowing null values here for now?

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -62,6 +75,41 @@ def self.kubeconfig(vm)
   def kubeconfig
     self.class.kubeconfig(cp_vms.first)
   end
+
+  def vm_diff_for_lb(load_balancer)
```
aamederen, line 79: This is a nitpick and maybe a note for future:

Load balancer code should be responsible for maintaining LBVM integrity.
The k8s Nexus codes should be responsible for making sure that every single node added/removed to the cluster is also modified in the LB.

Therefore, this check should not exist, maybe apart from sanity checking, which we don't have much in the code in general.

aamederen (approved):
LGTM, we can revisit some decisions later on based on how the ops/maintenance goes. Let's ship it after a final test.

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: I didn't fully understand. from my perspective, LB code gives us api to add/remove ports/vms and using that we should make sure we have satisified our requirements.

PR #2970, opened 2025-03-12
Install rhizome programs in all Vm based services to /home/rhizome

fdr:
Every managed service was going through the motions of creating a
rhizome user, and then not using it, instead installing into
/home/ubi, which is our default non-root, sudo-capable virtual machine
user.

This is bad for management of `authorized_keys` between machine and
personnel login, a future auditing system (separating administrative
and automated activity), it's bad for parity (on VmHosts, the concerns
were split as designed between `root` [which is bad for other reasons]
and rhizome, and finally, it renders 22f3aa5c, which will terminate
sessions from abandoned connections in a timely manner, but only does
so for the user name `rhizome` (via the configuration
`KillOnlyUsers=rhizome`)

### DIFF
--- a/prog/ai/inference_endpoint_replica_nexus.rb
+++ b/prog/ai/inference_endpoint_replica_nexus.rb
@@ -16,7 +16,7 @@ class Prog::Ai::InferenceEndpointReplicaNexus < Prog::Base
 
       inference_endpoint = InferenceEndpoint[inference_endpoint_id]
       vm_st = Prog::Vm::Nexus.assemble_with_sshable(
-        "ubi",
+        "rhizome",
         Config.inference_endpoint_service_project_id,
         location: inference_endpoint.location,
         name: ubid.to_s,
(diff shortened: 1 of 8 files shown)


aamederen (approved):

fdr:
If someone (ideally everyone, one per managed) could, like, try this out to see if it breaks their stuff in an obvious way...that would be the best, particular for those with more provisioning (e.g. Postgres). It's at least a little bit dangerous.

bsatzger:
> If someone (ideally everyone, one per managed) could, like, try this out to see if it breaks their stuff in an obvious way...that would be the best, particular for those with more provisioning (e.g. Postgres). It's at least a little bit dangerous.

In its current form it breaks inference endpoints. I'm pretty sure it breaks Kubernetes, too (search for `/home/ubi/kubernetes`). Please don't merge as is. I'll provide additional required changes for inference endpoints.

bsatzger:
https://github.com/ubicloud/ubicloud/pull/2979 goes a slightly different route as a suggestion. It works for inference endpoints. I also tested the change proposed here, bootstrapping rhizome fails.

fdr:
Needs refreshing if we're going to do it.

enescakir:
It would require more work when we decide to go this route. We can create the PR again when we're ready to work on it.

PR #2971, opened 2025-03-12
Do not create a Semaphore if there is no related Strand

jeremyevans:
This is designed to handle race conditions, where the strand is deleted between when the object is retrieved and when incr_* is called.

An alternative approach would be raising an exception for this case.  That can potentially alert us to cases that pass invalid strand ids, but it would require more code to handle race conditions.  We couple have separate methods, one that raised by default, and another than can be used in cases where we know the strand id is correct, but want to avoid a race condition.  However, I'm not sure the complexity is worth it.

While here, omit unnecessary hash values, switch from create_with_id to create, and remove explicit dataset call.

### DIFF
--- a/model/semaphore.rb
+++ b/model/semaphore.rb
@@ -7,8 +7,9 @@ class Semaphore < Sequel::Model
 
   def self.incr(strand_id, name)
     DB.transaction do
-      Strand.dataset.where(id: strand_id).update(schedule: Sequel::CURRENT_TIMESTAMP)
-      Semaphore.create_with_id(strand_id: strand_id, name: name)
+      if Strand.where(id: strand_id).update(schedule: Sequel::CURRENT_TIMESTAMP) == 1
+        create(strand_id:, name:)
+      end
     end
   end
 end
--- a/lib/sem_snap.rb
+++ b/lib/sem_snap.rb
@@ -38,7 +38,9 @@ class SemSnap
   end
 
   def incr(name)
-    add_semaphore_instance_to_snapshot(Semaphore.incr(@strand_id, name))
+    if (semaphore = Semaphore.incr(@strand_id, name))
+      add_semaphore_instance_to_snapshot(semaphore)
+    end
   end
 
   private
--- a/spec/lib/sem_snap_spec.rb
+++ b/spec/lib/sem_snap_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe SemSnap do
-  let(:st) { Strand.create_with_id(prog: "Test", label: "start") }
+  let(:st) { Strand.create(prog: "Test", label: "start") }
 
   it "can decrement semaphores" do
     described_class.use(st.id) do |snap|
@@ -38,5 +38,11 @@ RSpec.describe SemSnap do
   it "reads semaphores at initialization" do
     described_class.new(st.id).incr(:test)
     expect(described_class.new(st.id).set?(:test)).to be true
+  end
+
+  it ".incr returns nil if strand no longer exists" do
+    st.destroy
+    snap = described_class.new(st.id)
+    expect(snap.incr(:test)).to be_nil
   end
 end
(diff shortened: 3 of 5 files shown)


fdr:
I think incr has to support this, consider something among its most fundamental use cases, besides intra-process signaling: the user clicking stuff in the UI.

The whole point of `incr` is to allow inputs from anywhere anytime and retirement in an controlled manner, so I think a return value for the rare case it is worthwhile to even know about the situation that the object disappeared between two expressions is the right level of signaling.

fdr (approved):

jeremyevans:
Looking more into this, I don't think race condition would be a silent issue. `Semaphore.create` should raise a foreign key constraint violation if the Strand has been deleted.  I think it's still worth doing this, so we can avoid the unhandled exception.

When you say the code is wrong, maybe you are thinking about the situation where the strand is deleted between the strand UPDATE and semaphore INSERT, but that's not the situation I was trying to protect against.  I was trying to protect against the case where the strand has been deleted before `incr` has been called.  However, I agree that we should also handle the case where the strand is deleted after being updated, before the semaphore insert

I think we could do:

```diff
diff --git a/model/semaphore.rb b/model/semaphore.rb
index 54674029..1c4fc508 100644
--- a/model/semaphore.rb
+++ b/model/semaphore.rb
@@ -7,9 +7,13 @@ class Semaphore < Sequel::Model
 
   def self.incr(strand_id, name)
     DB.transaction do
-      Strand.dataset.where(id: strand_id).update(schedule: Sequel::CURRENT_TIMESTAMP)
-      Semaphore.create_with_id(strand_id: strand_id, name: name)
+      if Strand.where(id: strand_id).update(schedule: Sequel::CURRENT_TIMESTAMP) == 1
+        Semaphore.create(strand_id: , name:)
+      end
     end
+  rescue Sequel::ForeignKeyConstraintViolation
+    # Strand deleted between the strand update and semaphore create,
+    # no need to create the semaphore in this case
   end
 end
 
```

From some searching, I think this needs more work as there is a caller using the result of `Semaphore.incr`, and that needs an update to handle the case it does not return a Semaphore instance:

```diff
diff --git a/lib/sem_snap.rb b/lib/sem_snap.rb
index fceb90ea..6de8c5f0 100644
--- a/lib/sem_snap.rb
+++ b/lib/sem_snap.rb
@@ -38,7 +38,9 @@ class SemSnap
   end
 
   def incr(name)
-    add_semaphore_instance_to_snapshot(Semaphore.incr(@strand_id, name))
+    if (semaphore = Semaphore.incr(@strand_id, name))
+      add_semaphore_instance_to_snapshot(semaphore)
+    end
   end
 
   private
```

Should I update the PR with that?  It passes specs, but the SemSnap change needs a spec for coverage.

jeremyevans:
Pushed a commit to handle the race conditions and `SemSnap.incr` caller.

model/semaphore.rb
```diff
@@ -7,8 +7,15 @@ class Semaphore < Sequel::Model
 
   def self.incr(strand_id, name)
     DB.transaction do
-      Strand.dataset.where(id: strand_id).update(schedule: Sequel::CURRENT_TIMESTAMP)
-      Semaphore.create_with_id(strand_id: strand_id, name: name)
+      if Strand.where(id: strand_id).update(schedule: Sequel::CURRENT_TIMESTAMP) == 1
+        create(strand_id:, name:)
+      end
+    end
```
fdr, line 13: So this one is probably safe even without the rescue, because it's in a transaction and update takes a lock for the duration of the transaction that blocks a delete. The previous code I read in the context of autocommit, where an UPDATE commits, and then the record disappears.

jeremyevans: Good point.  I had a spec that tested this, but only by mocking things so we deleted the strand before the creation inside the same transaction.  Since that cannot occur in production, I agree the rescue is unnecessary.  I've pushed an updated commit to remove it the rescue and related spec.

fdr (approved):
I think we've done everything necessary (and nothing unnecessary) here.

PR #2973, opened 2025-03-12
Test tweaks

fdr:
some opportunistic fixes, though I am unsatisfied with the change to `routes/web/vm_spec.rb`, because it seems like the `let(:vm)` fixture should be better scoped. I am wondering what the latest and greatest in route testing organization is (e.g. norms in how to name/nest `describe` and/or `context` blocks), because I'm not sure vm_spec.rb abides it.

### DIFF
--- a/spec/prog/dns_zone/setup_dns_server_vm_spec.rb
+++ b/spec/prog/dns_zone/setup_dns_server_vm_spec.rb
@@ -38,8 +38,9 @@ RSpec.describe Prog::DnsZone::SetupDnsServerVm do
 
       expect(described_class.assemble(ds.id)).to be_a Strand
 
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
These changes all look good.  In terms of organization/nested contexts, I recommend avoiding nested context unless they need different before/after/let handling.

fdr:
> These changes all look good. In terms of organization/nested contexts, I recommend avoiding nested context unless they need different before/after/let handling.

In this case, perhaps :vm is in too large a scope, which could fall in the "need different `let` handling" bucket. The organization of the file in general seems a bit suspect to me, I am wondering if there's a newer convention held among the routes.

PR #2977, opened 2025-03-13
Start to charge VAT

enescakir:
- **Split invoice content expectations into smaller chunks**
  When we compare all the invoice content in one expectation, it's hard to
  see what's not matching. I think it's better to compare them key by key,
  as this makes it easier to find the differences.
  

- **Calculate VAT for EU countries**
  We will start charging VAT for EU customers next month.
  
  In this commit, we identify which Ubicloud entity will issue the invoice
  and calculate the VAT amount.
  
  I will update the invoice rendering to include VAT information in future
  commits.
  
  We will manually enable the `annual_non_dutch_eu_sales_exceed_threshold`
  flag when needed.
  
  The above rules are used to determine the issuer and VAT amount for
  invoices:
  
  - Invoices are issued by Ubicloud Inc. for non-EU customers without VAT
    applied.
  - Invoices are issued by Ubicloud B.V. for EU customers.
      - If the customer has provided a VAT number from the Netherlands, we
        charge 21% VAT.
      - If the customer has provided a VAT number from another European
        country, we include a reverse charge notice along with 0% VAT.
      - If the customer hasn't provided a VAT number, we charge 21% VAT
        until non-Dutch EU sales exceed annual threshold, than we charge
        local VAT.
  

- **Show the calculated VAT amount on the invoice**
  We calculated the related VAT amount in the previous commit.
  
  We’ve stopped using serializers for web routes. The invoice serializer
  is only used for web routes and adds unnecessary complexity to the
  codebase. We should remove it, but that’s beyond the scope of this PR.
  

- **Show the VAT amount in EUR as well**
  We charge customers in USD, but since we will start charging VAT, we
  also need to show the VAT amount in EUR.
  
  We finalize invoices at the end of each month. While doing this, we will
  check the EUR rates [^1] and provide it as an argument.
  
  [^1]: https://www.belastingdienst.nl/wps/wcm/connect/nl/douane_voor_bedrijven/content/hulpmiddel-wisselkoersen
  

- **Remove unnecessary placeholders in the billing form**
  These inputs already have labels, so we don’t need to display their
  titles as placeholders again.
  

- **Add VAT notice to billing form**
  
- **Remove billing info serializer**

  We've stopped using serializers for web routes, so I removed
  `Serializers::BillingInfo`.

  I moved the Stripe data mapping to the `stripe_data` method, which now
  returns the used hash instead of the raw data.


spec/lib/invoice_generator_spec.rb
```diff
@@ -28,28 +28,43 @@ def generate_billing_record(project, resource, span, amount = 5000)
...
-      "company_name" => nil
+      "name" => project.billing_info.stripe_data["name"],
+      "email" => project.billing_info.stripe_data["email"],
+      "address" => project.billing_info.stripe_data["address"]["line1"],
+      "country" => project.billing_info.stripe_data["address"]["country"],
+      "city" => project.billing_info.stripe_data["address"]["city"],
+      "state" => project.billing_info.stripe_data["address"]["state"],
+      "postal_code" => project.billing_info.stripe_data["address"]["postal_code"],
+      "tax_id" => project.billing_info.stripe_data["metadata"]["tax_id"],
+      "company_name" => project.billing_info.stripe_data["metadata"]["company_name"],
```
jeremyevans, line 66: I recommend we assign `project.billing_info.stripe_data` to a local variable and use that to DRY this up.

spec/lib/invoice_generator_spec.rb
```diff
@@ -77,11 +92,12 @@ def check_invoice_for_single_vm(invoices, project, vm, duration, begin_time)
       ["project_name", project.name],
       ["billing_info", expected_billing_info],
       ["issuer_info", expected_issuer],
+      ["vat_info", expected_vat_info],
       ["resources", expected_resources],
       ["subtotal", expected_cost],
       ["discount", 0],
       ["credit", 0],
-      ["cost", expected_cost]
+      ["cost", (expected_cost + (expected_vat_info&.[]("amount") || 0)).round(3)]
```
jeremyevans, line 100: Up to you, but you may want to consider `expected_vat_info&.fetch("amount", 0)`, since the fallback value is an immediate.

model/invoice.rb
```diff
@@ -164,6 +164,10 @@ def generate_pdf(data)
       pdf.text "#{data[:issuer_address]},"
       pdf.text "#{data[:issuer_city]}, #{data[:issuer_state]} #{data[:issuer_postal_code]},"
       pdf.text data[:issuer_country]
+      # :nocov:
+      pdf.text "#{data[:issuer_in_eu_vat] ? "VAT" : "Tax"} ID: #{data[:issuer_tax_id]}" if data[:issuer_tax_id]
```
jeremyevans, line 168: I realize it was like this already, but what is the reason behind the `:nocov:` here?  Is this difficult to cover?

jeremyevans (approved):
I don't know anything about VAT rates and when it is legal to charge/not charge, but in terms of the Ruby code, this looks good.

spec/lib/invoice_generator_spec.rb
enescakir: I removed `Serializers::BillingInfo` and refactor this part at https://github.com/ubicloud/ubicloud/pull/2977/commits/f847804386553b2c9faa6778e9b5e06095dd230b

Can you take a look at it?

spec/lib/invoice_generator_spec.rb
enescakir: If `expected_vat_info` is nil, `expected_vat_info&.fetch("amount", 0)` returns nil and fails with `nil can't be coerced into Float`. I added `.to_f`

model/invoice.rb
enescakir: I added test for it

jeremyevans (approved):
Changes look good.

PR #2979, opened 2025-03-13
Use rhizome as sshable user and /home/rhizome as script location for service VMs

bsatzger:
- **Use rhizome user for service VM sshables**
  As the final step of BootstrapRhizome, update the sshable unix_user from
  the bootstrap user to rhizome.
  
  Additionally, add operator_ssh_public_keys to autorized keys of the
  bootstrap user.
  
  For VM hosts, the operator/bootstrap user is `root`. For service VMs,
  the user used for bootstrapping and operator actions is configurable. By
  convention, that user is `ubi`.
  

- **Update rhizome script location for inference endpoints**
  

- **Update rhizome script location for Kubernetes**

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -130,7 +130,7 @@ class Prog::Vm::Nexus < Prog::Base
     ssh_key = SshKey.generate
     kwargs[:unix_user] = unix_user
     st = assemble(ssh_key.public_key, *, **kwargs)
-    Sshable.create(unix_user: unix_user, host: "temp_#{st.id}", raw_private_key_1: ssh_key.keypair) {
+    Sshable.create(unix_user: (unix_user == "runneradmin") ? "runneradmin" : "rhizome", host: "temp_#{st.id}", raw_private_key_1: ssh_key.keypair) {
       _1.id = st.id
     }
     st
--- a/prog/bootstrap_rhizome.rb
+++ b/prog/bootstrap_rhizome.rb
@@ -48,6 +48,7 @@ LOGIND
     pop "rhizome user bootstrapped and source installed" if retval&.dig("msg") == "installed rhizome"
 
     key_data = sshable.keys.map(&:private_key)
+    operator_keys = Config.operator_ssh_public_keys || ""
     Util.rootish_ssh(sshable.host, user, key_data, <<SH)
 set -ueo pipefail
 sudo apt update && sudo apt-get -y install ruby-bundler
@@ -60,6 +61,7 @@ sudo mkdir -p /etc/systemd/logind.conf.d
 echo #{LOGIND_CONFIG.shellescape} | sudo tee /etc/systemd/logind.conf.d/rhizome.conf > /dev/null
 echo #{SSHD_CONFIG.shellescape} | sudo tee /etc/ssh/sshd_config.d/10-clover.conf > /dev/null
 echo #{sshable.keys.map(&:public_key).join("\n").shellescape} | sudo tee /home/rhizome/.ssh/authorized_keys > /dev/null
+echo #{operator_keys.shellescape} | sudo tee -a ~/.ssh/authorized_keys > /dev/null
 sync
 SH
 
--- a/prog/kubernetes/provision_kubernetes_node.rb
+++ b/prog/kubernetes/provision_kubernetes_node.rb
@@ -130,7 +130,7 @@ BASH
         private_subnet_cidr6: kubernetes_cluster.private_subnet.net6,
         vm_cidr: vm.nics.first.private_ipv4
       }
-      vm.sshable.cmd("common/bin/daemonizer /home/ubi/kubernetes/bin/init-cluster init_kubernetes_cluster", stdin: JSON.generate(params))
+      vm.sshable.cmd("common/bin/daemonizer /home/rhizome/kubernetes/bin/init-cluster init_kubernetes_cluster", stdin: JSON.generate(params))
       nap 30
     when "InProgress"
       nap 10
--- a/rhizome/inference_endpoint/lib/replica_setup.rb
+++ b/rhizome/inference_endpoint/lib/replica_setup.rb
@@ -122,7 +122,7 @@ CONFIG
 Description=Download loadbalancer SSL cert service
     
 [Service]
-ExecStart=/home/ubi/inference_endpoint/bin/download-lb-cert
+ExecStart=/home/rhizome/inference_endpoint/bin/download-lb-cert
 
 #{common_systemd_settings}
 CERT_DOWNLOAD_SERVICE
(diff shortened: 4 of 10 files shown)


fdr:
Is mutating sshable necessary? A consideration for Util.rootish_ssh was to get around this (simplify a path where caching or other accreted code is not important, and may interact poorly)

prog/vm/nexus.rb
```diff
@@ -130,7 +130,7 @@ def self.assemble_with_sshable(unix_user, *, **kwargs)
     ssh_key = SshKey.generate
     kwargs[:unix_user] = unix_user
     st = assemble(ssh_key.public_key, *, **kwargs)
-    Sshable.create(unix_user: unix_user, host: "temp_#{st.id}", raw_private_key_1: ssh_key.keypair) {
+    Sshable.create(unix_user: (unix_user == "runneradmin") ? "runneradmin" : "rhizome", host: "temp_#{st.id}", raw_private_key_1: ssh_key.keypair) {
```
bsatzger, line 133: That is a little hacky: `(unix_user == "runneradmin")`

I'm not sure if it's feasible to use the rhizome user for Github runners. We could maybe extend `assemble_with_sshable` to accept the sshable user as additional parameter.

bsatzger:
> Is mutating sshable necessary? A consideration for Util.rootish_ssh was to get around this (simplify a code path where caching or other accreted code is not important, and may interact poorly)
> 
> e.g. consider how we never mutate sshable for vmhosts

that's possible. made some adjustments. there are implications to Github runners.

prog/vm/nexus.rb
fdr: naaah let's leave runneradmin, that IIRC is named to be the same as the image and we don't need a drip of bug reports about that.

I think more irksome is that the unix_user is ignored for every other value. I guess the problem is Util.rootish_ssh has no viable way to use a non-default for this? I will analyze this more closely. Thank you for testing & writing the patch versions

bsatzger: we could maybe do something like: `assemble_with_sshable(unix_user, sshable_user, ...`
assemble_with_sshable("ubi", "rhizome", ...)
assemble_with_sshable("runneradmin", "runneradmin", ...)

fdr:
> > Is mutating sshable necessary? A consideration for Util.rootish_ssh was to get around this (simplify a code path where caching or other accreted code is not important, and may interact poorly)
> > e.g. consider how we never mutate sshable for vmhosts
> 
> that's possible. made some adjustments. there are implications to Github runners.

so I'm a bit out of water here since I don't have any of the managed services running *at this very moment* (and I yielded my VM/hypervisor experimentation computer to someone else), but, I don't quite understand why we need to do something interesting here:

1. BootstrapRhizome takes a `user` in the frame. This is nominally/default `root` for bare metal where we're forced to bootstrap that way. It does not interact with `Sshable` at all, which, whether optimal in function or not, is the point of the simplified Util.rootish_ssh path.

2. Every prog that wants to do a managed VM service runs a BootstrapRhizome call itself, e.g. `    bud Prog::BootstrapRhizome, {"target_folder" => "minio", "subject_id" => vm.id, "user" => "ubi"}`, notably this should have it log in as `ubi`, our nominal not-root but root-alike bootstrap user. That this is the boilerplate is maybe not the best (one for every managed service except the micro-optimized Github Actions)

What this suggests to me is that `assemble_with_sshable` can accept `rhizome` as the `unix_user` per normal and each prog, e.g  `MinioServerNexus` can do the right thing (per line copied from above), which is invariably to use the user `ubi`, and does so independently, where `BootstrapRhizome` is unaware of Sshable's fields at all.

Am I missing something?

prog/vm/nexus.rb
fdr: See https://github.com/ubicloud/ubicloud/pull/2979#issuecomment-2731376582, maybe there's a loose connection somewhere, but it seems like there's already two places to feed in users, one in the `bud` of Bootstrap and the other via the `assemble_with_sshable`

fdr:
> > What this suggests to me is that `assemble_with_sshable` can accept `rhizome` as the `unix_user` per normal and each prog, e.g `MinioServerNexus` can do the right thing (per line copied from above), which is invariably to use the user `ubi`, and does so independently, where `BootstrapRhizome` is unaware of Sshable's fields at all.
> 
> Without any modifications, if `assemble_with_sshable` is called with the `rhizome` user, no `ubi` user will be created but only a rhizome user. BootstrapRhizome, attempting to (drop, for idempotency, and then) create that user, likely runs into issues.

Okay, I see, so the problem is here, in `  def self.assemble_with_sshable(unix_user, *, **kwargs)`

```ruby
    kwargs[:unix_user] = unix_user
    st = assemble(ssh_key.public_key, *, **kwargs)
    Sshable.create(unix_user: unix_user, host: "temp_#{st.id}", raw_private_key_1: ssh_key.keypair) {
      _1.id = st.id
    }
```

Namely that Sshable is coupled to the cloud-init user from `assemble(...**kwargs)`. The whole idea of using Sshable with custom users was added in 51ee43807c172f42312773fc10ce9cd6750b3843, for CI use, and it has one other known use to me, which is Github Actions runners which have a particular user name and topology we want to keep exactly on-par (we already had some friction from the two-user vs one-user skew seen in Github's first-party system). I think it's permissible to have these special cases go through a bit more difficulty.

Here's an idea:

```diff
-  def self.assemble_with_sshable(unix_user, *, **kwargs)
+  def self.assemble_with_sshable(*, sshable_unix_user: "rhizome", **kwargs)
     ssh_key = SshKey.generate
-    kwargs[:unix_user] = unix_user
     st = assemble(ssh_key.public_key, *, **kwargs)
-    Sshable.create(unix_user: unix_user, host: "temp_#{st.id}", raw_private_key_1: ssh_key.keypair) {
+    Sshable.create(unix_user: sshable_unix_user, host: "temp_#{st.id}", raw_private_key_1: ssh_key.keypair) {
       _1.id = st.id
     }
     st
```

disambiguate the initial-user (`cloud-init`) from ssh/"rpc" user, by giving it a distinct keyword argument symbol. The default "rpc" user (as it were) is `rhizome`, which is how Sshable was even originally designed (hard-coded). That keeps the verbose symbol name from proliforating through most code paths, and when it does appear, it suggests careful reading, the user considerations will be un-typical, and advanced.

Furthermore, typically, `**kwargs` wouldn't have anything to say about initial (`cloud-init`) unix users except for Github Actions and CI, and it gets passed through in an opaque manner.

fdr:
(I'm cooking up a patch for this, plz hold)

bsatzger:
Closed in favor of https://github.com/ubicloud/ubicloud/pull/3008

PR #2981, opened 2025-03-13
Validate VAT

enescakir:
- **Add a valid_vat column to billing_info**
  VAT validation is expensive and should be done asynchronously. We don't
  want to validate VAT numbers every month, so we cache their status until
  it changes.
  

- **Add helper to validate VAT number**
  European Commission has a REST API to validate VAT numbers.
  
  Example response for Ubicloud B.V.
  https://ec.europa.eu/taxation_customs/vies/rest-api/ms/NL/vat/NL864651442B01
  

- **Remove non-alphanumeric characters from the tax ID**
  Some users add spaces, hyphens, or periods to the tax ID, likely by
  copying from another source.
  
  Keeping the tax ID in a consistent format is helpful for processing.
  

- **Validate VAT async when it's changed**
  VAT validation is a costly operation, and also the European Commission
  services are fragile. So, it's better to validate VAT in the background
  without blocking the customer. If it fails, we can manually send an
  email to the customer to inform them about the failure for now. Once we
  trust the operation, we can automate the email notifications.
  

- **Send an email if the validated VAT is invalid**
  For now, we will send an email notification to ourselves if the VAT is
  invalid. Later, we can send this email to the customer.


routes/project/billing.rb
```diff
@@ -40,10 +42,16 @@ class Clover
...
               metadata: {
-                tax_id: r.params["tax_id"].gsub(/[^a-zA-Z0-9]/, ""),
+                tax_id: new_tax_id,
                 company_name: r.params["company_name"]
               }
             })
+            if new_tax_id != current_tax_id
+              DB.transaction do
+                billing_info.update(valid_vat: nil)
+                Strand.create(prog: "ValidateVat", label: "start", stack: [{subject_id: billing_info.id}])
```
jeremyevans, line 52: If `billing_info.country.in_eu_vat?` doesn't require a costly lookup, and only `billing_info.validate_vat` does, I recommend we only create a strand in the case where we need to do the `validate_vat` lookup.

jeremyevans (approved):
This looks good, though I recommend moving some logic from the prog to the route, assuming I'm correct that it isn't expensive.

PR #2987, opened 2025-03-14
Wait for a github runner job up to 7 minutes, instead of 5 minutes, 10 seconds

jeremyevans:
We've had two cases today where the github runner did not get assigned a job until a little over 6 minutes.


enescakir:
If the job isn't assigned to the runner after 5 minutes, we recycle the runner without charging the customer. However, this runner still uses some of our resources during that time. We don't want to extend the 5 minutes too much to avoid taking up our capacity.

As an alternative, if we encounter this edge case, we can extend the deadline to 2 hours:
https://github.com/ubicloud/ubicloud/pull/2990

PR #2993, opened 2025-03-17
Postgres Scale Up/Down Backend

byucesoy:
This is actually more than just scale up/down. It is E2E convergence logic. With
that, it automatically implements enable/disable HA.

Convergence is the process of converging from an arbitrary physical state of
PostgresResource to its target state, which is combination of target vm size,
target storage size and target server count.

Convergence logic is implemented in 4 steps:
1. Provision enough `fresh` servers to meet the target server count.
2. Wait until new servers are `ready` (i.e. in wait state).
3. Recycle representative_server to match resource's target.
4. Prune extra servers that are possibly left over from pre-convergence time.

At the end of the convergence, we ensure that the representative_server is in
the desired state and there are enough fresh standby servers that are ready to
take over the primary role if it becomes necessary.

Since the convergence is long running operation, it is possible that the target
changes during the convergence, especially during second step as other steps
are relatively quick. To ensure converging to correct target, second step can
decide to hop back to first step for course correction.

### DIFF
--- a/prog/postgres/converge_postgres_resource.rb
+++ b/prog/postgres/converge_postgres_resource.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+require_relative "../../lib/util"
+
+class Prog::Postgres::ConvergePostgresResource < Prog::Base
+  subject_is :postgres_resource
+
+  label def start
+    register_deadline(nil, 2 * 60 * 60)
+    hop_provision_servers
+  end
+
+  label def provision_servers
+    hop_wait_servers_to_be_ready if postgres_resource.has_enough_fresh_servers?
+
+    if postgres_resource.servers.all? { _1.vm.vm_host }
+      exclude_host_ids = (Config.development? || Config.is_e2e) ? [] : postgres_resource.servers.map { _1.vm.vm_host.id }
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids)
+    end
+
+    nap 5
+  end
+
+  label def wait_servers_to_be_ready
+    hop_provision_servers unless postgres_resource.has_enough_fresh_servers?
+    hop_recycle_representative_server if postgres_resource.has_enough_ready_servers?
+
+    nap 60
+  end
+
+  label def recycle_representative_server
+    if (rs = postgres_resource.representative_server)
+      hop_prune_servers unless rs.needs_recycling?
+      rs.trigger_failover
+    end
+
+    nap 60
+  end
+
+  label def prune_servers
+    # Below we only keep servers that does not need recycling. If there are
+    # more such servers than required, we prefer ready and recent servers (in that order)
+    servers_to_keep = postgres_resource.servers
+      .reject { _1.representative_at || _1.needs_recycling? }
+      .sort_by { [(_1.strand.label == "wait") ? 0 : 1, Time.now - _1.created_at] }
+      .take(postgres_resource.target_standby_count) + [postgres_resource.representative_server]
+    (postgres_resource.servers - servers_to_keep).each.each(&:incr_destroy)
+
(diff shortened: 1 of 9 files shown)


prog/postgres/converge_postgres_resource.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  label def wait_servers_to_be_ready
+    hop_provision_servers unless postgres_resource.has_enough_fresh_servers?
+    hop_recycle_representative_server if postgres_resource.has_enough_ready_servers?
+
+    nap 60
+  end
+
+  label def recycle_representative_server
+    if postgres_resource.representative_server
```
aamederen, line 29: How can a postgres resource not have a representative server? During the failover maybe?

prog/postgres/converge_postgres_resource.rb
```diff
@@ -0,0 +1,48 @@
...
+    if postgres_resource.servers.none? { _1.vm.vm_host.nil? }
+      exclude_host_ids = (Config.development? || Config.is_e2e) ? [] : postgres_resource.servers.map { _1.vm.vm_host.id }
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids)
+    end
+
+    nap 5
+  end
+
+  label def wait_servers_to_be_ready
+    hop_provision_servers unless postgres_resource.has_enough_fresh_servers?
```
aamederen, line 22: If I'm not mistaken, in an absurd case of multiple back-to-back repeated updates on the target, this code would jump back and forward to provision many servers, and the excess servers would be pruned at the end only. This can cause an capacity alert. This is a very hypothetical scenario and easy to mitigate once a page occurs, but just wanted to note if the logic makes sense. 

Maybe in order to avoid this and some unforeseen concurrency issues, we can disable patching while an update is going on in the front end PR.

enescakir (approved):
LGTM overall!

prog/postgres/converge_postgres_resource.rb
shikharbhardwaj: Yes, and this label itself might trigger a failover in case the representative server needs to be recycled so we wait here until a new one comes up.

shikharbhardwaj (approved):
Overall LGTM, should we add an E2E scenario for this?

prog/postgres/converge_postgres_resource.rb
byucesoy: Yes, during failover, there is a time window without a representative server.

Actually, while preparing this PR, I realized that there are many places where we assume representative server exist, which is not the case all the time. Thankfully, assuming its existence would not cause any permanent prblem. It only causes error to be thrown and we retry in next iteration (and eventually failover completes and we get a representative server). Still, it would be better to not assume. I put that down as a repair item for me.

prog/postgres/converge_postgres_resource.rb
byucesoy: Yeah, it is a possibility, but a quite rare one. I prefer to keep the logic simple for now and allow users to change target 1-2 times if they change their mind. We can change the logic if people start to abuse this.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "forwardable"
+
+require_relative "../../lib/util"
+
+class Prog::Postgres::ConvergePostgresResource < Prog::Base
+  subject_is :postgres_resource
+
+  label def provision_servers
```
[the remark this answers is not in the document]
byucesoy: Makes sense. Though MMW feature we are adding in upcoming PR makes this a bit complicated because it might increase the wait time for quite a bit. Still, I'm adding 2 hour deadline for this. We can  think about deadlines caused by MMW when they are an actual problem.

byucesoy:
> Overall LGTM, should we add an E2E scenario for this?
Yeah, probably. I'll add one.

PR #3000, opened 2025-03-17
Postgres Scale Up/Down Frontend

byucesoy:
**Add form title to resource creation form**

This form can be embedded to resource details page. In such cases, it would be
better to have a title for the form to indicate what the form is for.

**Add pre-selected options to the resource creation form**
These will be used to display forms with some options pre-selected, primarily
to update configurations of existing resources.

**Display different buttons on the resource creation form based on mode**
We are planning to re-use the resource creation form for both creating and
updating resources. This commit adds a new mode attribute to the form and
uses it to display different buttons based on the mode.

**Add PATCH support to resource creation form**
This form can be used to create a new resource or to update an existing one.
To update an existing resource, we need to send a PATCH request to the server.
This commit adds ability to send PATCH requests to the backend.

HTML forms unfortunately do not support PATCH requests. There are 2 common ways
to work around this limitation:
1. Send regular `POST` request with special hidden input and backend can handle
the request as if it is `PATCH` based on the hidden input field.
2. Use JavaScript to change the request method before submitting the form

I opted for the second approach because it fits to RESTful semantics better.
Also the proposal for extending HTML form methods lists additional drawbacks
regarding use of hidden input fields [1].

**Add PATCH endpoint for updating PostgresResource's target state**

**Add UI for PostgresResource update**

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -3,6 +3,7 @@ $(function () {
   setupDatePicker();
   setupFormOptionUpdates();
   setupPlayground();
+  setupFormsWithPatchMethod()
 });
 
 $(".toggle-mobile-menu").on("click", function (event) {
@@ -427,4 +428,33 @@ function setupPlayground() {
   };
 
   $('#inference_submit').on("click", generate);
+}
+
+function setupFormsWithPatchMethod() {
+  $("#creation-form.PATCH").on("submit", function (event) {
+    event.preventDefault();
+
+    var form = $(this);
+    var jsonData = {};
+    form.serializeArray().forEach(function(item) {
+        jsonData[item.name] = item.value;
+    });
+
+    $.ajax({
+        url: form.attr('action'),
+        type: 'PATCH',
+        dataType: "html",
+        data: jsonData,
+        success: function (response, status, xhr) {
+          var redirectUrl = xhr.getResponseHeader('Location');
+          if (redirectUrl) {
+              window.location.href = redirectUrl;
+          }
+        },
+        error: function (xhr, ajaxOptions, thrownError) {
+          let message = thrownError;
+          alert(`Error: ${message}`);
+        }
+    });
+  });
 }
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -40,6 +40,39 @@ class Clover
         204
       end
 
+      r.patch do
+        authorize("Postgres:edit", pg.id)
+
+        allowed_optional_parameters = ["size", "storage_size", "ha_type"]
+        ignored_parameters = ["family"]
+        request_body_params = validate_request_params([], allowed_optional_parameters, ignored_parameters)
+
+        target_vm_size = Validation.validate_postgres_size(pg.location, request_body_params["size"] || pg.target_vm_size)
+        target_storage_size_gib = Validation.validate_postgres_storage_size(pg.location, target_vm_size.vm_size, request_body_params["storage_size"] || pg.target_storage_size_gib)
+        ha_type = request_body_params["ha_type"] || PostgresResource::HaType::NONE
+        Validation.validate_postgres_ha_type(ha_type)
+
+        if pg.representative_server.nil? || target_storage_size_gib < pg.representative_server.storage_size_gib
+          begin
+            current_disk_usage = pg.representative_server.vm.sshable.cmd("df --output=used /dev/vdb | tail -n 1").strip.to_i / (1024 * 1024)
+          rescue
+            fail CloverError.new(400, "InvalidRequest", "Database is not ready for update", {})
+          end
+
+          if target_storage_size_gib * 0.8 < current_disk_usage
+            fail Validation::ValidationFailed.new({storage_size: "Insufficient storage size is requested. It is only possible to reduce the storage size if the current usage is less than 80% of the requested size."})
+          end
+        end
+
+        current_postgres_vcpu_count = (PostgresResource::TARGET_STANDBY_COUNT_MAP[pg.ha_type] + 1) * pg.representative_server.vm.vcpus
+        requested_postgres_vcpu_count = (PostgresResource::TARGET_STANDBY_COUNT_MAP[ha_type] + 1) * target_vm_size.vcpu
+        Validation.validate_vcpu_quota(@project, "PostgresVCpu", requested_postgres_vcpu_count - current_postgres_vcpu_count)
+
+        pg.update(target_vm_size: target_vm_size.vm_size, target_storage_size_gib:, ha_type:)
+
+        Serializers::Postgres.serialize(pg, {detailed: true})
+      end
+
       r.post "restart" do
         authorize("Postgres:edit", pg.id)
         pg.servers.each do |s|
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -30,6 +30,8 @@ class Clover
           Serializers::Postgres.serialize(pg, {detailed: true})
         else
           @pg = Serializers::Postgres.serialize(pg, {detailed: true, include_path: true})
+          @family = pg.representative_server.vm.family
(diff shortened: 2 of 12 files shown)


routes/project/location/postgres.rb
```diff
@@ -40,6 +40,33 @@ class Clover
         204
       end
 
+      r.patch do
+        authorize("Postgres:edit", pg.id)
```
aamederen, line 46: what do you think about disabling patching during a convergence is going on? It feels kinda dangerous/complex to allow changing the convergence target while trying to converge already, even though I cannot pinpoint an immediate problem at the moment.

views/components/form/resource_creation_form.erb
```diff
@@ -40,9 +42,14 @@ form_elements.each do |element|
     opt_text = content_generator.call(*option.values)
 
     parents = option.reject { |k, v| k == name }
+
     parents_selected = parents.all? { |k, v| selected_options[k] == v }
-    checked = parents_selected && (flash.dig("old", name) == opt_val || (flash.dig("old", name).nil? && selected_options[name].nil?))
+    selected_in_form_submission = flash.dig("old", name) == opt_val
```
enescakir, line 47: Instead of digging it three times, we could set it as a variable.

views/components/form/resource_creation_form.erb
```diff
@@ -104,8 +104,12 @@ form_elements.each do |element|
       </div>
       <div class="px-4 py-5 sm:p-6">
         <div class="flex items-center justify-end gap-x-6">
-          <a href="<%= action %>" class="text-sm font-semibold leading-6 text-gray-900">Cancel</a>
-          <%== part("components/form/submit_button", text: "Create") %>
+          <% if mode == "create" %>
+            <a href="<%= action %>" class="text-sm font-semibold leading-6 text-gray-900">Cancel</a>
```
enescakir, line 108: The update mode also might have a cancel button. Imagine you select different values for some properties but then decide not to apply them. The cancel button can refresh the page and load the current values.

assets/js/app.js
```diff
@@ -428,3 +429,32 @@ function setupPlayground() {
...
+        type: 'PATCH',
+        dataType: "json",
+        data: jsonData,
+        success: function (response, status, xhr) {
+          var redirectUrl = xhr.getResponseHeader('Location');
+          if (redirectUrl) {
+              window.location.href = redirectUrl;
+          }
+        },
+        error: function (xhr, ajaxOptions, thrownError) {
```
enescakir, line 454: When it returns due to validation failures, does it display them as regular validation errors below the input field or in an alert popup?

routes/project/location/postgres.rb
```diff
@@ -40,6 +40,33 @@ class Clover
...
+        request_body_params = validate_request_params([], allowed_optional_parameters, ignored_parameters)
+
+        parsed_target_vm_size = Validation.validate_postgres_size(pg.location, request_body_params["size"] || pg.target_vm_size)
+        target_storage_size_gib = Validation.validate_postgres_storage_size(pg.location, parsed_target_vm_size.vm_size, request_body_params["storage_size"] || pg.target_storage_size_gib)
+        ha_type = request_body_params["ha_type"] || PostgresResource::HaType::NONE
+        Validation.validate_postgres_ha_type(ha_type)
+
+        requested_standby_count = PostgresResource::TARGET_STANDBY_COUNT_MAP[ha_type]
+        requested_postgres_vcpu_count = (requested_standby_count + 1) * parsed_target_vm_size.vcpu
+        Validation.validate_vcpu_quota(@project, "PostgresVCpu", requested_postgres_vcpu_count)
```
enescakir, line 59: Since it already accounts for the used vCPUs on existing servers, should we only check the difference?

routes/project/location/postgres.rb
```diff
@@ -40,6 +40,33 @@ class Clover
...
+
+        DB.transaction do
+          pg.update(
+            target_vm_size: parsed_target_vm_size.vm_size,
+            target_storage_size_gib: target_storage_size_gib,
```
enescakir, line 64: Should we check the disk usage and not allow downgrading below that? If the disk is already full, this operation will fail anyway.

routes/project/location/postgres.rb
```diff
@@ -40,6 +42,39 @@ class Clover
...
+            ha_type: ha_type
+          )
```
enescakir, line 66: nit: if you keep `parsed_target_vm_size.vm_size` in `target_vm_size` variable since it already used twice.
```suggestion
          pg.update(target_vm_size:, target_storage_size_gib:, ha_type:)
```

routes/project/location/postgres.rb
```diff
@@ -40,6 +42,39 @@ class Clover
...
+        parsed_target_vm_size = Validation.validate_postgres_size(pg.location, request_body_params["size"] || pg.target_vm_size)
+        target_storage_size_gib = Validation.validate_postgres_storage_size(pg.location, parsed_target_vm_size.vm_size, request_body_params["storage_size"] || pg.target_storage_size_gib)
+        ha_type = request_body_params["ha_type"] || PostgresResource::HaType::NONE
+        Validation.validate_postgres_ha_type(ha_type)
+
+        requested_standby_count = PostgresResource::TARGET_STANDBY_COUNT_MAP[ha_type]
+        requested_postgres_vcpu_count = (requested_standby_count + 1) * parsed_target_vm_size.vcpu
+        Validation.validate_vcpu_quota(@project, "PostgresVCpu", requested_postgres_vcpu_count)
+
+        DB.transaction do
```
enescakir, line 61: It only includes single `pg.update` query. Do we need this transaction?

enescakir (commented):
Should we show a different display state while it's converging?

assets/js/app.js
byucesoy: It used to display as alert popup. But error under input field is better way to show this, so I changed it to that.

views/components/form/resource_creation_form.erb
```diff
@@ -8,6 +8,16 @@ let option_children = <%==JSON.generate(option_parents.each_with_object(Hash.new
...
+  <div class="md:flex md:items-center md:justify-between pb-2 lg:pb-4">
+    <div class="min-w-0 flex-1">
+      <h3 class="text-2xl font-bold leading-7 text-gray-900 sm:truncate sm:text-2xl sm:tracking-tight">
+        <%= form_title %>
+      </h3>
+    </div>
+  </div>
+<% end %>
+
 <div class="grid gap-6">
```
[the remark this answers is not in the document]
byucesoy: Subsequent PR handles this along with few other similar display issues.

routes/project/location/postgres.rb
byucesoy: Convergence is long operation and I don't want to put users into a situation where they had to wait long time when they change their mind. It felt dangerous to me at the beginning as well but after thinking on it for a quite while, I feel confident that changing the target wouldn't cause any major issues.

routes/project/location/postgres.rb
```diff
@@ -70,7 +72,20 @@ class Clover
...
+        else
+          flash["notice"] = "'#{pg.name}' will be updated according to requested configuration"
+          response["Location"] = "#{@project.path}#{pg.path}"
+          204
+        end
+      rescue CloverError => e
+        raise e if api?
+
+        flash["error"] = e.message
+        flash["errors"] = (flash["errors"] || {}).merge(e.details).transform_keys(&:to_s)
```
enescakir, line 86: We already perform similar rescue:
https://github.com/ubicloud/ubicloud/blob/28dacb968b9222e72bdc829c240ad3d272fb80c8/clover.rb#L226

But AJAX requests has early return

https://github.com/ubicloud/ubicloud/blob/28dacb968b9222e72bdc829c240ad3d272fb80c8/clover.rb#L214

I prefer not to duplicate rescues here. We can allow AJAX requests with a special header or let all PATCH requests continue instead of returning.

enescakir (approved):

PR #3003, opened 2025-03-17
Destroy Child Strands of K8s Provision Prog on Destroy

aamederen:
This specific case happens if the cluster is destroyed when the Provision prog is in wait_bootstrap_rhizome step.

### DIFF
--- a/prog/kubernetes/provision_kubernetes_node.rb
+++ b/prog/kubernetes/provision_kubernetes_node.rb
@@ -21,6 +21,7 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
 
   def before_run
     if kubernetes_cluster.strand.label == "destroy" && strand.label != "destroy"
+      strand.children.each(&:destroy)
       pop "provisioning canceled"
     end
   end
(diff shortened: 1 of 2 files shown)


fdr:
I don't think `BootstrapRhizome` implements destroy, and neither does `InstallRhizome`. Which is probably an oversight. But also represents a challenge for the compositional model used by the entire program (small progs have to do boilerplate to implement destroy). And that one uses `bud` most cases and the other uses `push` is almost an artifact of the earliest days of the program where I was trying out both (these days I think I should have focused on making `bud` more usable and not used `push` at all).

I guess for now we can bulldoze through both of those, by adding uninteresting `before_run` hooks.

fdr (changes requested):
need `before_run` on the composed progs  to make it effective :(

aamederen:
@fdr I think there are a few issues with implementing the `before_run` in Rhizome progs:
1. I was hesitant to touch BootstrapRhizome or InstallRhizome, as they are more commonly used, and wanted to fix this specific case in my limited domain. It's pretty safe to destroy rhizome strands as is **for now**.
1. Rhizome progs have `sshable` as subjects, and those cannot know whether they are destined to be destroyed, so the `before_run` block would even be uglier.
1. For k8s specific case, we don't do `Vm#incr_destroy` until the child processes are cleaned. So that also need to be reworked...

A better fix could revolve around #2722

PR #3004, opened 2025-03-17
Allow VmHost#download_boot_image with Default Version

aamederen:
We have a list of default options in the DownloadBootImage prog, so we can use it to allow default values when initiating download from the console on a host.

We can follow up with making the version in BootImage not null. In prod, we have 6 boot images with version: nil left.

### DIFF
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -10,7 +10,7 @@ class Prog::DownloadBootImage < Prog::Base
   end
 
   def version
-    @version ||= frame.fetch("version") { default_boot_image_version(image_name) }
+    @version ||= frame.fetch("version", nil) || default_boot_image_version(image_name)
   end
 
   def download_from_blob_storage?
@@ -133,13 +133,9 @@ class Prog::DownloadBootImage < Prog::Base
   end
 
   label def start
-    # YYY: we can remove this once we enforce it in the database layer.
-    # Although the default version is used if version is not passed, adding
-    # a sanity check here to make sure version is not passed as nil.
-    fail "Version can not be passed as nil" if version.nil?
-
     fail "Image already exists on host" if vm_host.boot_images_dataset.where(name: image_name, version: version).count > 0
 
+    # YYY: we should enforce version of boot image in the database layer.
     BootImage.create_with_id(
       vm_host_id: vm_host.id,
       name: image_name,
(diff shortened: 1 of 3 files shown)


prog/download_boot_image.rb
```diff
@@ -133,13 +133,9 @@ def blob_storage_client
...
 
   label def start
-    # YYY: we can remove this once we enforce it in the database layer.
-    # Although the default version is used if version is not passed, adding
-    # a sanity check here to make sure version is not passed as nil.
-    fail "Version can not be passed as nil" if version.nil?
-
     fail "Image already exists on host" if vm_host.boot_images_dataset.where(name: image_name, version: version).count > 0
 
+    # YYY: we should enforce version of boot image in the database layer.
```
fdr, line 138: Meh, no need to spawn more `YYY`, and what does it mean to enforce anyway? They were intended to be for stuff that solved in a timely manner and deleted from the ancient era,  but I never did, at least for some.

Funny thing, why `YYY`? Because in Postgres, `XXX` is the upstream string for suggestions to do things later, and when I worked on a fork of Postgres at Truviso, we wanted our own string, so, why not increment the byte in all three places: `YYY`.

That said, I actually don't care for the practice, except to mark things in development that really ought to be deleted, i.e. someone is really going to grep them and do something about them, which can happen in the alpha stage of a project or in complex out-of-tree patchsets. I would just as soon accept a patch to purge all our `YYY` comments. The nominal state should be no `YYY` comments.

prog/download_boot_image.rb
```diff
@@ -133,13 +133,9 @@ def blob_storage_client
   end
 
   label def start
-    # YYY: we can remove this once we enforce it in the database layer.
-    # Although the default version is used if version is not passed, adding
-    # a sanity check here to make sure version is not passed as nil.
-    fail "Version can not be passed as nil" if version.nil?
```
enescakir, line 139: Why are we removing this check? Shouldn't it automatically get the default version if an empty version is passed? So it shouldn't be nil anyway?

prog/download_boot_image.rb
```diff
@@ -10,7 +10,7 @@ def image_name
   end
 
   def version
-    @version ||= frame.fetch("version") { default_boot_image_version(image_name) }
+    @version ||= frame.fetch("version", nil) || default_boot_image_version(image_name)
```
enescakir, line 13:
```suggestion
    @version ||= frame["version"] || default_boot_image_version(image_name)
```

prog/download_boot_image.rb
aamederen: All I agree, except this is not a new `YYY`, but a partial address of an existing one.

PR #3008, opened 2025-03-18
Separate Sshable user and initial unix user in assemble_with_sshable

fdr:
This patch is intended to be a semantic no-op as applied, but once
managed services have been adapted to install their `rhizome` programs
into the `rhizome` home directory and use the `rhizome` user to
execute them (like `VmHost`s), the call sites currently specifying
`sshable_unix_user: "ubi"` can stop specifying `sshable_unix_user` and
rely on the default value `rhizome`.

`Sshable` was originally designed exclusively for the `rhizome` user
to support running its programs. This changed in commit
51ee43807c172f42312773fc10ce9cd6750b3843 to support E2E test cases
requiring standard cached SSH access without needing rhizome
installation (initially introduced in
68f0c7a8595318822e0b2877291dce66cd61ecf5).

It was later adapted for GitHub Actions, which learned to bypass the
`rhizome` upload phase for performance gains and meeting specific unix
user naming requirements for parity with first-party GitHub Action
Runners.  We saw firsthand that minor deviations here could lead to
confusion, including an inconsistency seen in the GitHub Actions
product: see e2329cc1ad27e2cd1e9e9183835fa429d7a52312.

It so happens both of these use cases involve using a single,
bootstrapped user as the only user clover is aware of or needs to be
aware of.  But, it was incorrectly extended to services with full
`rhizome` directories and upload steps.  This was an oversight.
Despite creating a separate `rhizome` user, `rhizome` programs were
loaded into the home directory of `ubi`, and both automation and
personnel used the same unix user to log in and `authorized_keys` file
to log in, which made `authorized_keys` more awkward to update than
the arrangement seen on VM Hosts, which was intended to be the default
arrangement.

Although both Sshable and Vm have parameters and database fields named
`unix_user`, they originated independently for distinct purposes. I
think their shared naming, while typically a useful heuristic for
coupling in Clover's codebase, contributed to this repeated oversight.

### DIFF
--- a/prog/ai/inference_endpoint_replica_nexus.rb
+++ b/prog/ai/inference_endpoint_replica_nexus.rb
@@ -17,8 +17,8 @@ class Prog::Ai::InferenceEndpointReplicaNexus < Prog::Base
 
       inference_endpoint = InferenceEndpoint[inference_endpoint_id]
       vm_st = Prog::Vm::Nexus.assemble_with_sshable(
-        "ubi",
         Config.inference_endpoint_service_project_id,
+        sshable_unix_user: "ubi",
         location: inference_endpoint.location,
         name: ubid.to_s,
         size: inference_endpoint.vm_size,
--- a/prog/dns_zone/setup_dns_server_vm.rb
+++ b/prog/dns_zone/setup_dns_server_vm.rb
@@ -20,8 +20,8 @@ class Prog::DnsZone::SetupDnsServerVm < Prog::Base
 
     DB.transaction do
       vm_st = Prog::Vm::Nexus.assemble_with_sshable(
-        "ubi",
         Config.dns_service_project_id,
+        sshable_unix_user: "ubi",
         location: location,
         name: name,
         size: vm_size,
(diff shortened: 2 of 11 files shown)


jeremyevans (approved):
The changes seem fine to me, but I'm not sure if this separates the arguments.  This looks like it just converts a required positional argument to an optional keyword argument.  Maybe the backstory should be inserted into the commit message?

fdr:
> The changes seem fine to me, but I'm not sure if this separates the arguments. This looks like it just converts a required positional argument to an optional keyword argument. Maybe the backstory should be inserted into the commit message?

there is an aliasing issue, in that both Vm has a `unix_user` (which is passed to `cloud-init`, and defaults to `ubi` on this platform), but so does Sshable, though Sshable began life being hard-coded to `rhizome`, but to the extent it grew the ability to be not-rhizome is due to a couple of particular carve-outs (a case in CI, as related in commit message said): 51ee43807c172f42312773fc10ce9cd6750b3843. So it really began as two unrelated concepts that got merged.

bsatzger (approved):

enescakir (approved):

byucesoy (approved):

furkansahin:
Why is everyone approving a draft pr?

PR #3011, opened 2025-03-19
Add a runner force terminate button to the UI

enescakir:
We destroy the runner when GitHub sends the workflow job completed webhook event.

Recently, the runner script has gotten stuck a few times due to zombie processes.

This might be related to GitHub incidents, but we don't have enough information to confirm it.

https://github.com/actions/runner/issues/3737

Since GitHub servers think the script is still running, they don't stop it.

Customers can force terminate the runner by clicking the button. This destroys the underlying virtual machine, causing GitHub servers to lose connection to the runner and mark the job as failed.

CleanShot 2025-03-19 at 12 36 24@2x

CleanShot 2025-03-19 at 12 36 30@2x

### DIFF
--- a/routes/project/github.rb
+++ b/routes/project/github.rb
@@ -18,10 +18,23 @@ class Clover
         r.redirect "#{@project.path}/github/runner"
       end
 
-      r.get "runner" do
-        @runners = Serializers::GithubRunner.serialize(@project.github_runners_dataset.eager(:vm, :strand).reverse(:created_at).all)
+      r.on "runner" do
+        r.get true do
+          @runners = Serializers::GithubRunner.serialize(@project.github_runners_dataset.eager(:vm, :strand).reverse(:created_at).all)
 
-        view "github/runner"
+          view "github/runner"
+        end
+
+        r.is String do |runner_ubid|
+          next unless (runner = GithubRunner.from_ubid(runner_ubid)) && runner.installation.project_id == @project.id
+
+          r.delete true do
+            runner.incr_skip_deregistration
+            runner.incr_destroy
+            flash["notice"] = "Runner '#{runner.ubid}' forcibly terminated"
+            204
+          end
+        end
       end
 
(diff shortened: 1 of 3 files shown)


routes/project/github.rb
```diff
@@ -18,10 +18,23 @@ class Clover
...
+        end
+
+        r.is String do |runner_ubid|
+          next unless (runner = GithubRunner.from_ubid(runner_ubid)) && runner.installation.project_id == @project.id
 
-        view "github/runner"
+          r.delete true do
+            runner.incr_skip_deregistration
+            runner.incr_destroy
+            flash["notice"] = "Runner '#{runner.ubid}' forced to be terminated"
```
bsatzger, line 34:
```suggestion
            flash["notice"] = "Runner '#{runner.ubid}' forcibly terminated"
```

views/github/runner.erb
```diff
@@ -7,9 +7,10 @@
 <div class="grid gap-6">
   <%== part(
     "components/table_card",
-    headers: ["Runner", "Repository", "Label", "State", "Branch", "Workflow Job"],
+    headers: ["Runner", "Repository", "Label", "State", "Branch", "Workflow Job", ""],
```
bsatzger, line 10: The button has no text and the column has no header. To a user, it's a dangerously looking button but not clear what it does until you push it.

enescakir: Since we already have a lot of columns, it's hard to fit a wide table on the page. Adding text to it makes this worse. I assume this button won't be used much. It's for rare cases.

What do you think about the HTML title attribute? It shows up when you keep the cursor on the element.

CleanShot 2025-03-19 at 15 22 03@2x

bsatzger: sounds good. if we get feedback that it's not obvious enough then we can still make changes.

bsatzger (approved):

PR #3014, opened 2025-03-19
Fix Github Runner user and ssh setup

fdr:
fb31111e65c330b3c030f80e244901827c29ef63 had an oversight that blows back very quickly: I didn't adjust Github Runners to set an initial, non-default-named unix user, I only adjusted its Sshable creation. The result is a cascade of authentication errors to `runneradmin` and lack of progress.

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -47,6 +47,7 @@ class Prog::Vm::GithubRunner < Prog::Base
 
     vm_st = Prog::Vm::Nexus.assemble_with_sshable(
       Config.github_runner_service_project_id,
+      unix_user: "runneradmin",
       sshable_unix_user: "runneradmin",
       name: github_runner.ubid.to_s,
       size: label_data["vm_size"],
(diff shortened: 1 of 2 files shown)


fdr:
If you apply this diff without the fix, it provokes a test failure as expected:

```diff
diff --git a/spec/prog/vm/github_runner_spec.rb b/spec/prog/vm/github_runner_spec.rb
index c2e56383..c372649b 100644
--- a/spec/prog/vm/github_runner_spec.rb
+++ b/spec/prog/vm/github_runner_spec.rb
@@ -78,6 +78,7 @@
       vm = nx.pick_vm
       expect(vm).not_to be_nil
       expect(vm.sshable.unix_user).to eq("runneradmin")
+      expect(vm.unix_user).to eq("runneradmin")
       expect(vm.family).to eq("standard")
       expect(vm.vcpus).to eq(4)
       expect(vm.project_id).to eq(Config.github_runner_service_project_id)
```

jeremyevans (approved):
Seems like it would be good to have one of these default to the other, so you only need to specify both in the case where they are different.

fdr:
> Seems like it would be good to have one of these default to the other, so you only need to specify both in the case where they are different.

They are usually different, is the thing, Github Actions is an exception (both wants initial user and automated user the same)

PR #3016, opened 2025-03-19
PostgreSQL Managed Maintenance Windows

byucesoy:
We are adding self-serve scale up/down ability, but it triggers failover as soon as the standbys are ready, which is probably not a desired behavior for many users. More preferable option would be doing the failover at the low load time (i.e at maintenance window). The problem is that we don't have any notion of maintenance windows, which is what this PR adds.

image


lib/validation.rb
```diff
@@ -99,6 +99,17 @@ def self.validate_postgres_flavor(flavor)
     end
   end
 
+  def self.validate_postgres_maintenance_window_start_at(maintenance_window_start_at)
+    return nil if maintenance_window_start_at.nil?
+
+    parsed_maintenance_window_start_at = maintenance_window_start_at.to_i
+    if parsed_maintenance_window_start_at.to_s != maintenance_window_start_at.to_s || parsed_maintenance_window_start_at < 0 || parsed_maintenance_window_start_at > 23
```
aamederen, line 106: what about something like `(0..23).map{_1.to_s}.include?(maintenance_window_start_at)`

aamederen (approved):

routes/project/location/postgres.rb
```diff
@@ -234,6 +234,28 @@ class Clover
...
       end
 
+      r.post "set-maintenance-window" do
+        authorize("Postgres:edit", pg.id)
+
+        allowed_optional_parameters = ["maintenance_window_start_at"]
+        request_body_params = validate_request_params([], allowed_optional_parameters)
+
+        maintenance_window_start_at = request_body_params["maintenance_window_start_at"]
+        maintenance_window_start_at = nil if web? && maintenance_window_start_at == ""
```
enescakir, line 244: Why only for the web?

enescakir (commented):
I have a high level question. We already return the value in the PostgreSQL resource serializer. Why don't we update it also in the PG patch request instead of separate endpoint?

routes/project/location/postgres.rb
```diff
@@ -234,6 +234,28 @@ class Clover
...
+        maintenance_window_start_at = Validation.validate_postgres_maintenance_window_start_at(maintenance_window_start_at)
+
+        DB.transaction do
+          pg.update(maintenance_window_start_at:)
+        end
```
jeremyevans, line 249: I don't think this transaction is necessary. `Model#update` uses a transaction already, and it is the only call inside the block.

migrate/20250318_add_maintenance_window.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:postgres_resource) do
+      add_column :maintenance_window_start_at, :integer
```
jeremyevans, line 6: I think it might be worthwhile to add a check constraint:

```ruby
constraint :valid_maintenance_windows_start_at, maintenance_window_start_at: 0..23
```

lib/validation.rb
```diff
@@ -99,6 +99,17 @@ def self.validate_postgres_flavor(flavor)
     end
   end
 
+  def self.validate_postgres_maintenance_window_start_at(maintenance_window_start_at)
```
jeremyevans, line 102: I think it's better to define `PostgresResource#validate` instead of defining this method:

```ruby
def validate
  super
  validates_includes(0..23, :maintenance_window_start_at, allow_nil: true, message: "must be between 0 and 23")
end
```

routes/project/location/postgres.rb
jeremyevans: ` maintenance_window_start_at = nil` isn't needed if you move the validation code to the model (as mentioned in the earlier comment).  Sequel's typecasting will automatically convert `""` to `nil` since the column type is integer.

jeremyevans (commented):
I agree that we should add this feature.

fdr:
Love that @will-minded daily and hourly maintenance window.

will:
I am @will

PR #3017, opened 2025-03-19
Check that provided locations in routes are valid visible locations

jeremyevans:
Add LocationNameConverter.to_visible_internal_name to check that a
given location is a valid visible location.

Add Clover#check_visible_location, which calls to_visible_internal_name
and halts route processing if a location is not a valid visible location.
Use this in the routes in all cases where @location was being set
manually.

Change delete specs that use invalid locations to expect 404 instead of
204.  204 is not needed, as the deletion could never have succeeded with
an invalid location.

### DIFF
--- a/lib/location_name_converter.rb
+++ b/lib/location_name_converter.rb
@@ -5,10 +5,6 @@ module LocationNameConverter
     Location.where(display_name:).get(:name)
   end
 
-  def self.to_visible_internal_name(display_name)
-    Location.where(display_name:, visible: true).get(:name)
-  end
-
   def self.to_display_name(internal_name)
     Location.where(name: internal_name).get(:display_name)
   end
--- a/lib/location_name_converter.rb
+++ b/lib/location_name_converter.rb
@@ -5,6 +5,10 @@ module LocationNameConverter
     Location.where(display_name:).get(:name)
   end
 
+  def self.to_visible_internal_name(display_name)
+    Location.where(display_name:, visible: true).get(:name)
+  end
+
   def self.to_display_name(internal_name)
     Location.where(name: internal_name).get(:display_name)
   end
--- a/lib/location_name_converter.rb
+++ b/lib/location_name_converter.rb
@@ -2,10 +2,10 @@
 
 module LocationNameConverter
   def self.to_internal_name(display_name)
-    Location[display_name:]&.name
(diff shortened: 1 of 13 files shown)


lib/location_name_converter.rb
```diff
@@ -5,6 +5,10 @@ def self.to_internal_name(display_name)
     Location.where(display_name:).get(:name)
   end
 
+  def self.to_visible_internal_name(display_name)
```
enescakir, line 8: Since it's used in just one place, do we really need a separate helper instead of using it directly?

To be honest, I haven't liked the `LocationNameConverter` module since day one. Since we have the `Location` model ("We didn't have it when this module was introduced), we might deprecate it and use the `Location` model with its own helpers in the future. This isn't related to this PR.

spec/routes/api/project/location/postgres_spec.rb
```diff
@@ -334,7 +334,7 @@
       it "not exist ubid in location" do
```
enescakir, line 334: Since we centralized this logic with `check_visible_location` helper, we might test `/project/#{project.ubid}/location/foo_location` path once instead of testing it for all resources separately.

lib/location_name_converter.rb
jeremyevans: We definitely don't need a helper method for it.  I'll push a commit that inlines the behavior into the Clover helper method.

spec/routes/api/project/location/postgres_spec.rb
jeremyevans: That makes sense.  I'll push a commit for that.

jeremyevans: Unfortunately, you cannot test `/project/#{project.ubid}/location/foo_location` separately, since by itself, it isn't a valid path.  We can pick one spec that tests the behavior, and delete the other ones that would be redundant as they test the same behavior.  I pushed a commit for that, as well as one that fixes some related specs.

enescakir (approved):

enescakir:
It has a conflict 🔥

jeremyevans:
> It has a conflict 🔥

Yep, that's for the private subnet route fix.  I'll fix that tomorrow.

PR #3020, opened 2025-03-20
Rename ubi-cross rake task to ubi-release

jeremyevans:
Have ubi-release build build tar.gz/zip versions and source tarball.
    
GitHub releases doesn't support directly shipping executables, so they need to be wrapped. This builds wrapped versions that unwrap to ubi/ubi.exe.
    
We should also produce a source tarball that operating system packagers can use to build packages.  I am including a basic Makefile in the source tarball that uses the correct version. I tested the source tarball by using it to successfully build an OpenBSD package.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -293,20 +293,43 @@ task "ubi" do
   sh("cd cli && go build -ldflags '-X main.version=#{cli_version.call}' -tags osusergo,netgo")
 end
 
-desc "Cross compile cli/ubi binaries for common platforms"
-task "ubi-cross" do
+desc "Build release files for cli/ubi"
+task "ubi-release" do
   version = cli_version.call
 
   Dir.chdir("cli") do
+    File.delete("ubi") if File.file?("ubi")
+
     os_list = %w[linux windows darwin]
     arch_list = %w[amd64 arm64 386]
     os_list.each do |os|
       arch_list.each do |arch|
         next if os == "darwin" && arch == "386"
         next if os == "windows" && arch == "386" # Windows Defender falsely flags as Trojan:Win32/Bearfoos.A!ml
-        sh("env GOOS=#{os} GOARCH=#{arch} go build -ldflags '-s -w -X main.version=#{version}' -o ubi-#{os}-#{arch}-#{version}#{".exe" if os == "windows"} -tags osusergo,netgo")
+
+        filename = "ubi-#{os}-#{arch}-#{version}"
+        exe_filename = "ubi#{".exe" if os == "windows"}"
+
+        sh("env GOOS=#{os} GOARCH=#{arch} go build -ldflags '-s -w -X main.version=#{version}' -o #{exe_filename} -tags osusergo,netgo")
+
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -293,20 +293,43 @@ task "ubi" do
...
+
+        filename = "ubi-#{os}-#{arch}-#{version}"
+        exe_filename = "ubi#{".exe" if os == "windows"}"
+
+        sh("env GOOS=#{os} GOARCH=#{arch} go build -ldflags '-s -w -X main.version=#{version}' -o #{exe_filename} -tags osusergo,netgo")
+
+        if os == "windows"
+          sh("zip", "#{filename}.zip", "ubi.exe")
+        else
+          sh("tar", "zcf", "#{filename}.tar.gz", "ubi")
```
fdr, line 318: Is it necessary to tar stuff with a single file as the payload, at least on github releases & unix? I don't think people even gzip stuff necessarily anymore given transport compression. It does make the `curl && chmod a+x` or web browser + `chmod && mv` methodology more difficult.

fdr: while it's common to zip things or whatever, I guess I had cloud-hypervisor in mind which releases bare files. Not saying it's normal, I haven't done a survey, but I did feel it made a couple of things slightly more straightforward.

fdr: ...then again, I guess this idea falls apart if releasing a file of the same name on several platforms, right? then encoding into a `tar` all containing `ubi` might make things easier again.

jeremyevans: It's necessary to wrap in tar/gz/zip, as GitHub releases does not allow uploading files without an extension, or files with an `.exe` extension.  I chose the tar.gz route instead of plain gz so I could keep the executable name `ubi`, so clients wouldn't need to rename to get the expected name.

We could certainly drop the wrapping if we wanted to host the files somewhere other than GitHub releases.

fdr: You sure? i see a lot of extension-less binaries here: https://github.com/cloud-hypervisor/cloud-hypervisor/releases

*but* given it's also somewhat annoying to have to do curl *-o ubi* https:///ubi-darwin-arm (or whatever) I think it's a wash.

jeremyevans: You are correct.  Looks like I was clicking on the wrong button.  The "Paste, drop, or click to add files" button doesn't allow it, but the "Attach binaries ..." button does.  I guess that's the one we should be using.

Do you want to upload the binaries directly, or should I keep the tar.gz/zip wrappers?  I'm fine either way.

fdr (approved):

Rakefile
fdr: I think either way is fine. More binary handling does raise questions about how the whole pipeline is kept clean of a supply chain attack.

I suppose, microscopically, tarballs can be annoying when multiple files are not involved or if being a stickler for reproducible builds (which we're not doing, [but we could ](https://github.com/golang/go/issues/53976)), but it ends up being one command anyway.

PR #3025, opened 2025-03-20
Add vm pool_id index

fdr:
A quick look at pg_stat_statements shows this can be a slow one without an index on this field.


jeremyevans (approved):

fdr:
had to run `rake refresh_sequel_caches` for this one

jeremyevans:
> had to run `rake refresh_sequel_caches` for this one (I thought re-migrating my test database might do it in passing)

I think the reason `rake test_up` didn't do it is because PostgreSQL had not finished creating the index and marking it active by the time `refresh_sequel_caches` was run implicitly (because CONCURRENTLY was used).

jeremyevans:
> That seems weird, that only should happen if you don't let the CREATE CONCURRENTLY finish first. As far as I know there's no asynchrony _after_ it completes. If there was, it wouldn't even be safe to use the finished database as a template immediately thereafter (as the parallel test setup does)

There's nothing in the code that waits for the index build to finish.  You are likely right about the parallel setup code.  For non-unique indexes, this shouldn't cause problems, though.  The cached index metadata for unique indexes is important, as auto_validations uses it to set up uniqueness validations.

Is this something you want me to address?  I could add a `wait_for_concurrent_indexes_to_build` or something similar.

fdr:
Huh. Weird, I don't think it's important enough to address, although I'm still puzzled as to the mechanical cause, because because I (perhaps incorrectly?!) think that so long as you run all the commands of the migrations first, after the last command completes, there should be no `invalid` indexes (which would give lie to my thought that CREATE INDEX CONCURRENTLY flips the catalog entry before it commits its final transaction).

jeremyevans:
`CREATE INDEX CONCURRENTLY` completes immediately, because all it does is schedule an index build (you can think of it as asynchronous).  The index builds in the background, and can take an arbitrary amount of time.  When it finishes, assuming it was successful, it then marks the index as valid.  I don't think PostgreSQL offers a synchronous version of `CREATE INDEX CONCURRENTLY` (which would wait until the index is built successfully before returning).  The only way to get the behavior you want would be to poll until the index shows up as valid (or alternatively, until there are no indexes where `pg_index.indisready` is false).

fdr:
> `CREATE INDEX CONCURRENTLY` completes immediately,

uhhhh not in my experience. Rather, it completes very slowly, doing multiple scans.

indexcmd.c:

```c
	/*
	 * Phase 2 of concurrent index build (see comments for validate_index()
	 * for an overview of how this works)
	 *
	 * Now we must wait until no running transaction could have the table open
	 * with the old list of indexes.  Use ShareLock to consider running
	 * transactions that hold locks that permit writing to the table.  Note we
	 * do not need to worry about xacts that open the table for writing after
	 * this point; they will see the new index when they open it.
	 *
	 * Note: the reason we use actual lock acquisition here, rather than just
	 * checking the ProcArray and sleeping, is that deadlock is possible if
	 * one of the transactions in question is blocked trying to acquire an
	 * exclusive lock on our table.  The lock code will detect deadlock and
	 * error out properly.
	 */
	WaitForLockers(heaplocktag, ShareLock, true);

	/*
	 * At this moment we are sure that there are no transactions with the
	 * table open for write that don't have this new index in their list of
	 * indexes.  We have waited out all the existing transactions and any new
	 * transaction will have the new index in its list, but the index is still
	 * marked as "not-ready-for-inserts".  The index is consulted while
	 * deciding HOT-safety though.  This arrangement ensures that no new HOT
	 * chains can be created where the new tuple and the old tuple in the
	 * chain have different index keys.
	 *
	 * We now take a new snapshot, and build the index using all tuples that
	 * are visible in this snapshot.  We can be sure that any HOT updates to
	 * these tuples will be compatible with the index, since any updates made
	 * by transactions that didn't know about the index are now committed or
	 * rolled back.  Thus, each visible tuple is either the end of its
	 * HOT-chain or the extension of the chain is HOT-safe for this index.
	 */

	/* Set ActiveSnapshot since functions in the indexes may need it */
	PushActiveSnapshot(GetTransactionSnapshot());

	/* Perform concurrent build of index */
	index_concurrently_build(tableId, indexRelationId);

	/* we can do away with our snapshot */
	PopActiveSnapshot();

	/*
	 * Commit this transaction to make the indisready update visible.
	 */
	CommitTransactionCommand();
	StartTransactionCommand();

	/* Tell concurrent index builds to ignore us, if index qualifies */
	if (safe_index)
		set_indexsafe_procflags();

	/*
	 * Phase 3 of concurrent index build
	 *
	 * We once again wait until no transaction can have the table open with
	 * the index marked as read-only for updates.
	 */
	pgstat_progress_update_param(PROGRESS_CREATEIDX_PHASE,
								 PROGRESS_CREATEIDX_PHASE_WAIT_2);
	WaitForLockers(heaplocktag, ShareLock, true);

	/*
	 * Now take the "reference snapshot" that will be used by validate_index()
	 * to filter candidate tuples.  Beware!  There might still be snapshots in
	 * use that treat some transaction as in-progress that our reference
	 * snapshot treats as committed.  If such a recently-committed transaction
	 * deleted tuples in the table, we will not include them in the index; yet
	 * those transactions which see the deleting one as still-in-progress will
	 * expect such tuples to be there once we mark the index as valid.
	 *
	 * We solve this by waiting for all endangered transactions to exit before
	 * we mark the index as valid.
	 *
	 * We also set ActiveSnapshot to this snap, since functions in indexes may
	 * need a snapshot.
	 */
	snapshot = RegisterSnapshot(GetTransactionSnapshot());
	PushActiveSnapshot(snapshot);

	/*
	 * Scan the index and the heap, insert any missing index entries.
	 */
	validate_index(tableId, indexRelationId, snapshot);

	/*
	 * Drop the reference snapshot.  We must do this before waiting out other
	 * snapshot holders, else we will deadlock against other processes also
	 * doing CREATE INDEX CONCURRENTLY, which would see our snapshot as one
	 * they must wait for.  But first, save the snapshot's xmin to use as
	 * limitXmin for GetCurrentVirtualXIDs().
	 */
	limitXmin = snapshot->xmin;

	PopActiveSnapshot();
	UnregisterSnapshot(snapshot);

	/*
	 * The snapshot subsystem could still contain registered snapshots that
	 * are holding back our process's advertised xmin; in particular, if
	 * default_transaction_isolation = serializable, there is a transaction
	 * snapshot that is still active.  The CatalogSnapshot is likewise a
	 * hazard.  To ensure no deadlocks, we must commit and start yet another
	 * transaction, and do our wait before any snapshot has been taken in it.
	 */
	CommitTransactionCommand();
	StartTransactionCommand();

	/* Tell concurrent index builds to ignore us, if index qualifies */
	if (safe_index)
		set_indexsafe_procflags();

	/* We should now definitely not be advertising any xmin. */
	Assert(MyProc->xmin == InvalidTransactionId);

	/*
	 * The index is now valid in the sense that it contains all currently
	 * interesting tuples.  But since it might not contain tuples deleted just
	 * before the reference snap was taken, we have to wait out any
	 * transactions that might have older snapshots.
	 */
	pgstat_progress_update_param(PROGRESS_CREATEIDX_PHASE,
								 PROGRESS_CREATEIDX_PHASE_WAIT_3);
	WaitForOlderSnapshots(limitXmin, true);

	/*
	 * Index can now be marked valid -- update its pg_index entry
	 */
	index_set_state_flags(indexRelationId, INDEX_CREATE_SET_VALID);
```

CREATE INDEX CONCURRENTLY is a rare command that compounds with multiple transactions, but it is synchronous -- there's no built-in mechanism for forking another backend to background anything.

jeremyevans:
> > `CREATE INDEX CONCURRENTLY` completes immediately,
> 
> uhhhh not in my experience. Rather, it completes very slowly, doing multiple scans.

TIL :)

I don't have an explanation for why refresh_sequel_caches needed to be run manually after a test_up, then.

PR #3031, opened 2025-03-21
Nap longer while waiting for the semaphore to increase

enescakir:
- **Nap 6 hours while waiting for the semaphore to increase**
  When we increase the semaphore, we have already scheduled the strand for
  now.
  
  If the label is just waiting for the semaphore to increase, there's no
  need for short naps.
  
  Most of our wait labels are just waiting for the semaphore to increase,
  so I extended their naps to 6 hours.
  
  It will help decrease the load on the respirate on production
  
  [^1]: https://github.com/ubicloud/ubicloud/blob/28dacb968b9222e72bdc829c240ad3d272fb80c8/model/semaphore.rb#L10
  

- **Increase the subnet nap time to 10 minutes**
  The subnet nexus waits for the semaphore to increase or checks if 24
  hours have passed since the last rekeying.
  
  It is already scheduled for now when the semaphore increased.
  
  For the last rekeying, there's no need to check every 30 seconds; 10
  minutes is more than enough.

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -120,7 +120,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
   end
 
   label def wait
-    nap 65536
+    nap 6 * 60 * 60
   end
 
   label def destroy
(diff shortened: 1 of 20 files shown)


jeremyevans (approved):
Ruby wise, the changes look fine to me. I agree it makes sense to nap longer in these cases.

byucesoy (approved):
LGTM

PR #3032, opened 2025-03-21
Add a restarter program and use it for respirate

fdr:
By default, this prevents the supervisor monitoring `restarter` from seeing that the process it is supervises exited more than once per hour by default.

In the case of Heroku, this is of importance, because after a few consecutive crashes (e.g. from a recalcitrant server causing apoptosis from a `select()` loop without a timeout).  The quantitatively exact Heroku backoff behavior is not documented, so some future tuning will have to take place.

When Heroku elects to no longer start the process again, this pages us or degrades `respirate` throughput for a time.

There's no reason we can't restart more or less right away. So, this patch does this.  There are some secondary/fine considerations:

1. The restarting is randomized to a degree, to prevent too many exactly simultaneous startups.

2. It is desirable to exit sometimes, in case there is a network or computer specific dependency to the failure, to request the process supervisor (e.g. Heroku, Kubernetes) to put the entire process somewhere else...but only once in a while.

3. The log format has some regularity to it: all the `restarter` activity is in the top level key `restarter`.  In it, there is always a `start_time`, which is thought to be unique (along with PID, as often conveyed by syslog) to both identify and convey useful information about the `restarter` session.  In addition to `start_time` there is always a "going concern" key (`startup`, `subprocess_exit`, `shutdown`) with information relevant to that phase in the program's execution in particular.

You can test the basic function of the program like this:

    $ RESTART_MINIMUM_TIME_ALIVE=10 bin/restarter bash -c 'sleep 1; echo hi; exit 1'
    {"restarter":{"start_time":"2025-03-21T17:52:53+00:00","startup":{"subprocess_args":["bash","-c","sleep 1; echo hi; exit 1"],"minimum_time_alive":10}}}
    hi
    {"restarter":{"start_time":"2025-03-21T17:52:53+00:00","subprocess_exit":{"command":["bash","-c","sleep 1; echo hi; exit 1"],"exitstatus":1,"pid":938199,"success":false}}}
    hi
    {"restarter":{"start_time":"2025-03-21T17:52:53+00:00","subprocess_exit":{"command":["bash","-c","sleep 1; echo hi; exit 1"],"exitstatus":1,"pid":938212,"success":false}}}
    {"restarter":{"start_time":"2025-03-21T17:52:53+00:00","shutdown":{"start_time":"2025-03-21T17:52:53+00:00"}}}

[1]: We decline to use the Ruby Timeout class that injects an exception at any point in child thread execution, because it's more or less impossible for us or any library we use, including the standard library, to write a correct program that can have its stack unwound at any point that way.

### DIFF
--- a/bin/restarter
+++ b/bin/restarter
@@ -1,0 +1,43 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+minimum_time_alive = Integer(ENV.fetch("RESTART_MINIMUM_TIME_ALIVE", "3600"), 10)
+require "json"
+
+command = ARGV.dup.freeze
+start_time = Time.now
+
+iso_start = start_time.utc.strftime("%Y-%m-%dT%H:%M:%S%:z").freeze
+puts JSON.generate(restarter: {
+  start_time: iso_start,
+  startup: {
+    command:,
+    minimum_time_alive:
+  }
+})
+
+loop do
+  system(*command)
+  status = $?
+
+  puts JSON.generate(restarter: {
+    start_time: iso_start,
+    command_exit: {
+      command:,
+      exitstatus: status.exitstatus,
+      pid: status.pid,
+      success: status.success?
+    }
+  })
+
+  sleep(rand(1..10))
+
+  if Time.now - start_time > minimum_time_alive
+    puts JSON.generate(restarter: {start_time: iso_start,
+                                   shutdown: {
+                                     command:,
+                                     minimum_time_alive:
+                                   }})
+    exit
+  end
+end
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 web: bundle exec puma -C puma_config.rb
-respirate: bin/respirate
+respirate: bin/restarter bin/respirate
 monitor: bin/monitor


fdr:
I don't love the lack of tests here (mea culpa), but I think it can eliminate some hypotheses about possibly sporadically degraded respirate throughput (a single crashed process). I'll port it be more a shim for a `lib` file

bin/restarter
```diff
@@ -0,0 +1,42 @@
...
+command = ARGV.dup.freeze
+start_time = Time.now
+
+iso_start = start_time.utc.strftime("%Y-%m-%dT%H:%M:%S%:z").freeze
+puts JSON.generate(restarter: {
+  start_time: iso_start,
+  startup: {
+    subprocess_args: command,
+    minimum_time_alive: MINIMUM_TIME_ALIVE
+
```
jeremyevans, line 16:
```suggestion
```

bin/restarter
```diff
@@ -0,0 +1,42 @@
...
+  }
+})
+
+loop do
+  pid = Process.spawn(*command, out: :out, err: :err)
+  Process.wait(pid)
```
jeremyevans, line 22:
```suggestion
  system(*command)
```

I think this has the same behavior.

bin/restarter
```diff
@@ -0,0 +1,42 @@
...
+  status = $?
+
+  puts JSON.generate(restarter: {
+    start_time: iso_start,
+    subprocess_exit: {
+      command: command,
```
jeremyevans, line 28:
```suggestion
      command:,
```

bin/restarter
```diff
@@ -0,0 +1,42 @@
...
+      pid: status.pid,
+      success: status.success?
+    }
+  })
+
+  sleep(rand(1..10))
+
+  if Time.now - start_time > MINIMUM_TIME_ALIVE
+    puts JSON.generate(restarter: {start_time: iso_start,
+                                   shutdown: {start_time: iso_start}})
```
jeremyevans, line 39: Is there a reason you want the start_time logged in two separate places in the same message?

bin/restarter
```diff
@@ -0,0 +1,42 @@
...
+    exit
```
jeremyevans, line 40:
```suggestion
    exit 1
```

This indicates a failure condition, I believe.

bin/restarter
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+MINIMUM_TIME_ALIVE = Integer(ENV.fetch("RESTART_MINIMUM_TIME_ALIVE", "3600"), 10)
```
jeremyevans, line 4: This could be a local variable instead of a constant.  That would allow you to omit the hash value below.

jeremyevans (commented):
I think it makes sense to have this so we have better control over the restarts.

bin/restarter
fdr: buffers & captures stdout, so it doesn't, right?

fdr: Oh no that's backticks.

bin/restarter
fdr: oversight from refactoring, thank you

bin/restarter
fdr: seems like a good win for that, even though it is a constant in some larger sense..

bin/restarter
fdr: I suppose, although...isn't every restarter exit a failure condition? that's why it seemed strange to give it an abnormal exit code.

jeremyevans: I guess since is this is the only way to exit other than by signal, it doesn't matter.

bin/restarter
```diff
@@ -0,0 +1,43 @@
...
+require "json"
+
+command = ARGV.dup.freeze
+start_time = Time.now
+
+iso_start = start_time.utc.strftime("%Y-%m-%dT%H:%M:%S%:z").freeze
+puts JSON.generate(restarter: {
+  start_time: iso_start,
+  startup: {
+    subprocess_args: command,
```
jeremyevans, line 14: You could consider `command` instead of `subprocess_args`, both here and below.  You are already using `command` for the other use.

jeremyevans (approved):

bin/restarter
fdr: sure, will do

enescakir (approved):
I thought we would create multiple respirate processes at some point, but I guess that's for later, right?

bin/restarter
```diff
@@ -0,0 +1,43 @@
...
+    minimum_time_alive:
+  }
+})
+
+loop do
+  system(*command)
```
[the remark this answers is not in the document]
fdr: signals are propagated by default.

fdr:
> I thought we would create multiple respirate processes at some point, but I guess that's for later, right?

only if the interpreter lock was a bottleneck...and since our processes are so I/O intensive (that is, many opportunities to release the lock, for an era of CPU time) , that seems...remote?

PR #3040, opened 2025-03-22
Fix division by zero in progress bar calculation

enescakir:
When we set the project quota to zero, the denominator becomes zero,
which breaks the project detail page.


views/components/progress_bar.erb
```diff
@@ -1,5 +1,5 @@
 <%# locals: (numerator:, denominator:, title:) %>
-<% progress = [100, 100 * numerator / denominator].min
+<% progress = [100, denominator.zero? ? 0 : 100 * numerator / denominator].min
```
byucesoy, line 2: I feel like it would be better representation if we assume progress is 100% when the denominator is 0

enescakir: Makes sense. Changed it

byucesoy (approved):

PR #3041, opened 2025-03-22
Accepts multiple SSH keys from the user

enescakir:
When a customer inputs two SSH keys separated by a newline, they expect
both keys to be added to the .ssh/authorized_keys file on the created
VM.

Our dataplane does support multiple SSH keys, but we currently pass them
as a single string. We need to split them into individual keys before
sending them to the dataplane.


byucesoy (approved):

model/vm.rb
```diff
@@ -197,6 +197,7 @@ def params_json(swap_size_bytes)
     topo = cloud_hypervisor_cpu_topology
 
     project_public_keys = project.get_ff_vm_public_ssh_keys || []
+    public_keys = public_key.split("\n").map(&:strip).reject { _1.nil? || _1.empty? }
```
jeremyevans, line 200:
```suggestion
    public_keys = public_key.split("\n").each(&:strip!)
    public_keys.reject!(&:empty?)
    public_keys.concat(project_public_keys)
```

I don't believe `String#strip` can return `nil`.  Since `String#split` returns a newly allocated array of newly allocated strings, `.map(&:strip)` => `.each(&:strip!)` has the same behavior, but does not allocate a new array and new strings.  Likewise, `reject` -> `reject!` saves an array allocation.  And using `concat` instead of `+` also saves an array allocation.

model/vm.rb
```diff
@@ -207,7 +208,7 @@ def params_json(swap_size_bytes)
       "local_ipv4" => local_vetho_ip.to_s.shellescape || "",
       "dns_ipv4" => nics.first.private_subnet.net4.nth(2).to_s,
       "unix_user" => unix_user,
-      "ssh_public_keys" => [public_key] + project_public_keys,
+      "ssh_public_keys" => public_keys + project_public_keys,
```
jeremyevans, line 211:
```suggestion
      "ssh_public_keys" => public_keys,
```

Assuming previous recommended change implemented.

jeremyevans (approved):
Looks good.  I have a couple recommendations, but I leave them up to you.

model/vm.rb
enescakir: I applied changes

PR #3051, opened 2025-03-26
Add Ruby SDK and reimplement CLI processing using it

jeremyevans:
This adds a custom (i.e. not-autogenerated) Ruby SDK, designed specifically
around Ubicloud.  Examples of use:

```ruby
require "ubicloud"

# Setup SDK context
UBI = Ubicloud.new(:net_http, token: "YOUR_API_TOKEN", project_id: "pj...")

# Get list of VMs
UBI.vm.list

# Create a firewall
UBI.firewall.create(location: "eu-central-h1", name: "my-fw")

# Retrieive a load balancer,
lb = UBI["1b345678901234567890123456"]

# then destroy it
lb.destroy

# Schedule a PostgreSQL database restart
UBI.postgres.new("eu-central-h1/my-fw").restart
```

The SDK comes with two adapters, net_http and rack. net_http uses the
net/http standard library to submit HTTP requests.  rack directly calls
rack applications.  Users are expected to use net_http. UbiCli uses the
rack adapter to directly submit internal requests to Clover.

This changes the custom code in UbiCli and the cli-commands files to use
the Ruby SDK for all changes.  This is slightly slower, but allows for
testing Ruby SDK using the existing tests for the CLI.

The only spec change is to an error message, and the error message is now
more helpful.

As the Ruby SDK uses symbol keys instead of string keys, this switches
UbiCli and the cli-commands files to also use symbol keys.

This adds an UbiCli#check_no_slash helper method to DRY up code that
checks for slashes in strings used as path fragments.

CC @geemus

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /tmp/*
 /assets/css/app.css
 /.auto-parallel-tests
+/sdk/ruby/*.gem
 /sql.log
 /sql_query_parameterization_analysis.txt
 /spec/routes/api/cli/spec-output-files
--- a/Rakefile
+++ b/Rakefile
@@ -325,6 +325,11 @@ task "annotate" do
   Sequel::Annotate.annotate(Dir["model/**/*.rb"])
 end
 
+desc "Build sdk gem"
+task "build-sdk-gem" do
+  sh("cd sdk/ruby && gem build ubicloud.gemspec")
+end
+
 desc "Emit assets before deploying"
 task "assets:precompile" do
   sh("npm", "install")
--- a/cli-commands/fw/post/add-rule.rb
+++ b/cli-commands/fw/post/add-rule.rb
@@ -11,11 +11,8 @@ UbiCli.on("fw").run_on("add-rule") do
   args 1
 
   run do |cidr, opts|
-    range = opts[:fw_add_rule].values_at(:"start-port", :"end-port")
-    range[1] ||= range[0] || 65535
-    range[0] ||= 0
-    post(fw_path("/firewall-rule"), "cidr" => cidr, "port_range" => range.join("..")) do |data|
-      ["Added firewall rule with id: #{data["id"]}"]
-    end
+    start_port, end_port = opts[:fw_add_rule].values_at(:"start-port", :"end-port")
+    id = sdk_object.add_rule(cidr, start_port:, end_port:)[:id]
+    response("Added firewall rule with id: #{id}")
   end
 end
--- a/cli-commands/fw/post/attach-subnet.rb
+++ b/cli-commands/fw/post/attach-subnet.rb
@@ -8,8 +8,7 @@ UbiCli.on("fw").run_on("attach-subnet") do
   args 1
 
   run do |subnet_id|
-    post(fw_path("/attach-subnet"), "private_subnet_id" => subnet_id) do |data|
-      ["Attached private subnet with id #{subnet_id} to firewall with id #{data["id"]}"]
-    end
+    id = sdk_object.attach_subnet(subnet_id).id
+    response("Attached private subnet with id #{subnet_id} to firewall with id #{id}")
   end
 end
--- a/cli-commands/fw/post/create.rb
+++ b/cli-commands/fw/post/create.rb
@@ -9,8 +9,7 @@ UbiCli.on("fw").run_on("create") do
 
   run do |opts|
     params = underscore_keys(opts[:fw_create])
-    post(fw_path, params) do |data|
-      ["Firewall created with id: #{data["id"]}"]
-    end
+    id = sdk.firewall.create(location: @location, name: @name, **params).id
+    response("Firewall created with id: #{id}")
   end
 end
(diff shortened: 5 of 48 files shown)


fdr:
I will let @geemus comment in substance...but, if he's okay with how the API looks and can't see a nice mechanical tweak to make it even better, I'm okay with it.

One thing that initially alarmed me was the net size, but breaking down the diffstat, it isn't so bad:

```
 .gitignore                                    |   1 +
 Rakefile                                      |   5 ++
 sdk/ruby/MIT-LICENSE                          |  21 +++++
 sdk/ruby/README.md                            | 432 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 sdk/ruby/lib/ubicloud.rb                      |  48 +++++++++++
 sdk/ruby/lib/ubicloud/adapter.rb              |  62 +++++++++++++++
 sdk/ruby/lib/ubicloud/adapter/net_http.rb     |  55 +++++++++++++
 sdk/ruby/lib/ubicloud/adapter/rack.rb         |  41 ++++++++++
 sdk/ruby/lib/ubicloud/context.rb              |  53 +++++++++++++
 sdk/ruby/lib/ubicloud/model.rb                | 293 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 sdk/ruby/lib/ubicloud/model/firewall.rb       |  60 ++++++++++++++
 sdk/ruby/lib/ubicloud/model/load_balancer.rb  |  53 +++++++++++++
 sdk/ruby/lib/ubicloud/model/postgres.rb       |  73 +++++++++++++++++
 sdk/ruby/lib/ubicloud/model/private_subnet.rb |  29 +++++++
 sdk/ruby/lib/ubicloud/model/vm.rb             |  27 +++++++
 sdk/ruby/lib/ubicloud/model_adapter.rb        |  25 ++++++
 sdk/ruby/ubicloud.gemspec                     |  21 +++++
 spec/ruby_sdk_spec.rb                         | 256 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 spec/thawed_mock.rb                           |   1 +
 19 files changed, 1556 insertions(+)
```

that is, the economy on each new type in the algebra is, 30-70 lines, and I guess since Jeremy is going to rdoc this, it has far more comments than average.  Mostly fixed costs (e.g. model.rb) are of a reasonable size. Plus, there's README.md

I took a look at how correlations work, e.g. posting or referencing to correlated objects, and how those correlations are itemized, and how systematically. I don't have super advanced thoughts or comments, but I guess I see that as one of the more sensitive design areas.

The way that the SDK jacked straight into the CLI but, per Thin CLIent, on the back end is a clever way to get maximum exercise for minimum extra. I wonder if there's some way we could be even more systematic than fragments of strings and little checks here and there (not so often), but maybe not.

Finally, we probably can't (and are unable to) think as hard about SDKs in other languages as this one. I wonder if the Ruby SDK could be reflected upon to generate, say, a Python implementation that was Pretty Okay, and even a Go one? Possibly augmented with OpenAPI to some degree, e.g. for at least nillity information when necessary. That is, it would be an auxiliary source of information that, in practice, is necessary beyond OpenAPI to produce a decent client, along with custom code generation.

geemus (approved):

geemus:
Seems good to me (did a lot of feedback/discussion in Slack already before the PR was ready).

I think the API is looking pretty good overall, though the schema need some love (it's on my todo list). The only thing that stuck out a bit is that I'm finding the location prefixes everywhere seems a bit noisy. I guess I imagine that in many use cases you would mostly be in operating in one location, not changing constantly. So having to constantly pass it vs something maybe more modal seems a bit annoying. That feels pretty minor though and I suspect is best just fixed at the SDK level. (I could argue for a header or something instead of path, but it seems like many folks find that a worse experience). In any event, it doesn't make me itch enough to be convinced it would be worth the trouble to try to change.

jeremyevans:
> I guess I imagine that in many use cases you would mostly be in operating in one location, not changing constantly. So having to constantly pass it vs something maybe more modal seems a bit annoying. That feels pretty minor though and I suspect is best just fixed at the SDK level. (I could argue for a header or something instead of path, but it seems like many folks find that a worse experience). In any event, it doesn't make me itch enough to be convinced it would be worth the trouble to try to change.

In terms of the SDK, this is easy to add later in a backwards compatible manner if needed:

```ruby
UBI = Ubicloud.new(...).location("eu-north-h1")
UBI.vm.new("test-vm") # Assumes eu-north-h1 location
UBI.vm.list # Only list those in specified location
UBI.vm.create(name: "test-vm", public_key: "...") # Create in specified location
```

It raises questions about how we should handle these cases:

```ruby
UBI["vm-ubid"] # Should we allow access if in another location?
UBI.vm.new("eu-central-h1/test-vm") # Should we allow this?
UBI.vm.list(location: "eu-central-h1") # Should we allow this?
```

Basically, should `.location` restrict the location or should it set a default location if not specified?  I expect setting a default location is easier to implement, and gives users more flexibility.

Anyway, I think we should wait for customer requests before implementing this feature.

geemus:
Default vs explicit is tricky. I think default might be a better user experience, except if/when it isn't (in which case it could be surprising in a bad way). Whereas more explicit is a worse experience but less prone to surprise pain.

fdr (approved):
this found fewer subtle improvements to the API than I expected, but I feel much better for the proof that API can interoperate with a real program (that is, the Ruby SDK seen herein)

PR #3052, opened 2025-03-26
Fix issue when two consecutive calls to ip generate resulted in duplcate ips

mohi-kalantari:
For setting up the ipv4 ip of the pod, we needed 2 ips and a function was called twice to get the ips. On rare occasions, since the first available ip was not accounted for, we could end up with duplicate ips. so a new optional field was added to reserve the first ip so we would be sure we won't end up with the same ip twice.

Some improvements were also made to improve the ipv4 ip allocation.

The old behaviour for allocating ip (both for ipv4 and ipv6) was randomly selecting an ip and checking if it is duplicate and repeating the process 100 times until we find an available ip. this works fine on ipv6 because ranges are so huge and the number of pods can never reach that state.

but for ipv4, we could easily run out of available ips. in the new method, we will generate every possible ipv4 ip (possible because ranges are so small, definitely not possible with ipv6) and then rejecting the used ips and randomly select one from them. This way we won't rely on luck and max retries to get the available ip.

### DIFF
--- a/rhizome/kubernetes/lib/ubi_cni.rb
+++ b/rhizome/kubernetes/lib/ubi_cni.rb
@@ -141,7 +141,7 @@ options ndots:5
 
   def setup_ipv4(subnet, cni_netns, inner_ifname, outer_ifname)
     container_ip = find_random_available_ip(IPAddr.new(subnet))
-    gateway_ip = find_random_available_ip(IPAddr.new(subnet))
+    gateway_ip = find_random_available_ip(IPAddr.new(subnet), reserved_ips: [container_ip.to_s])
 
     r "ip addr replace #{gateway_ip}/24 dev #{outer_ifname}"
     r "ip link set #{outer_ifname} mtu #{MTU} up"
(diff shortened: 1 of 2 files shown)


rhizome/kubernetes/lib/ubi_cni.rb
```diff
@@ -219,17 +219,23 @@ def error_exit(message)
     exit 1
   end
 
-  def find_random_available_ip(subnet)
-    allocated_ips = @ipam_store["allocated_ips"].values.flatten
-    max_retries = 100
+  def find_random_available_ip(subnet, reserved_ips: [])
+    allocated_ips = @ipam_store["allocated_ips"].values.flatten + reserved_ips
```
jeremyevans, line 223: Since `flatten` returns a new array, you can use `concat` to avoid an array allocation

```suggestion
    allocated_ips = @ipam_store["allocated_ips"].values.flatten.concat(reserved_ips)
```

rhizome/kubernetes/lib/ubi_cni.rb
```diff
@@ -219,17 +219,23 @@ def error_exit(message)
...
 
-    max_retries.times do
-      ip = generate_random_ip(subnet)
-      unless allocated_ips.include?(ip.to_s)
-        return ip
+    if subnet.ipv4?
+      available_ips = generate_all_usable_ips(subnet).reject { |ip| allocated_ips.include?(ip.to_s) }
```
jeremyevans, line 226: `generate_all_usable_ips` returns a new array, so `reject!` is more efficient (but you cannot chain it):

```suggestion
      available_ips = generate_all_usable_ips(subnet)
      available_ips.reject! { |ip| allocated_ips.include?(ip.to_s) }
```

jeremyevans (approved):
This looks correct to me, though I have a couple small suggestions for Ruby improvements.

rhizome/kubernetes/lib/ubi_cni.rb
mohi-kalantari: thanks

furkansahin (approved):

PR #3058, opened 2025-03-27
Adjust Billing Rates for k8s

aamederen:
We will start charging for k8s clusters, so this commit updates
    the billing rates for the nodes.

We don't offer node selection choices at the moment and default to
    the smallest worker node size standard-2 with public IP and 40GB
    storage. The pricing calculation for a k8s worker node is the same
    with a bare VM with the same specs.

For CP, in order to be competitive, we make the price $16 per node.

We might need to adjust slightly when we offer different sizes for
    nodes. Since we don't have a separate line item for the Public IP
    of a node, the pricing of k8s nodes won't scale the same with VMs.


byucesoy (approved):

aamederen:
Thanks @byucesoy , as discussed after we deploy the change and the time to charge comes, I will manually finalize existing billing records and recreate them to catch up the new rates.

I will hold the PR just for now, in case an unexpected delay comes.

PR #3061, opened 2025-03-27
Fix loadbalancer tests flakiness

mohi-kalantari:
After introducing the multiple ports feature, some tests started to fail randomly

The issue was calling some associations like vm_ports on the load_balancer model for example. we should not rely on the first and last since the result is unordered. Since there was no order the generated rules would differ if compared char-to-char but had the same functionality.

LB methods are modified to generate rules based on the sorted results (sorted whether on port or address) which gives deterministic output

Also the test specs are improved to use the database records instead of mocking for a cleaner code. for example Nics are now created instead of being mocked at every level.

### DIFF
--- a/prog/vnet/update_load_balancer_node.rb
+++ b/prog/vnet/update_load_balancer_node.rb
@@ -31,6 +31,9 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
   end
 
   def generate_lb_based_nat_rules
+    active_vm_ports_uniq_by_port = load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port, load_balancer_vm: {vm: :nics}) }
+      .uniq(&:load_balancer_port_id)
+      .sort_by { |vm_port| vm_port.load_balancer_port.src_port }
     public_ipv4 = vm.ephemeral_net4.to_s
     public_ipv6 = vm.ephemeral_net6.nth(2).to_s
     private_ipv4 = vm.private_ipv4
@@ -47,7 +50,7 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
     end
 
     ipv4_prerouting = if load_balancer.ipv4_enabled?
-      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }.map do |vm_port|
+      active_vm_ports_uniq_by_port.map do |vm_port|
         port = vm_port.load_balancer_port
         ipv4_map_def = generate_lb_map_defs_ipv4(port)
         modulo = ipv4_map_def.count
@@ -60,7 +63,7 @@ ip daddr #{private_ipv4} tcp dport #{port.src_port} ct state established,related
     end
 
     ipv6_prerouting = if load_balancer.ipv6_enabled?
-      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }.map do |vm_port|
+      active_vm_ports_uniq_by_port.map do |vm_port|
         port = vm_port.load_balancer_port
         ipv6_map_def = generate_lb_map_defs_ipv6(port)
         modulo = ipv6_map_def.count
@@ -129,22 +132,30 @@ TEMPLATE
       "elements = {#{neighbor_vms.map(&:private_ipv6).join(", ")}}"]
   end
 
-  def generate_lb_map_defs_ipv4(current_port)
-    load_balancer.active_vm_ports
+  def generate_lb_map_defs(current_port)
+    items = load_balancer.active_vm_ports
       .select { |vm_port| vm_port.load_balancer_port.dst_port == current_port.dst_port }
-      .map
-      .with_index { |vm_port, index| "#{index} : #{vm_port.load_balancer_vm.vm.private_ipv4} . #{(vm_port.vm.id == vm.id) ? vm_port.load_balancer_port.dst_port : vm_port.load_balancer_port.src_port}" }
+      .map do |vm_port|
+        address = yield vm_port
+        port = (vm_port.load_balancer_vm.vm_id == vm.id) ? vm_port.load_balancer_port.dst_port : vm_port.load_balancer_port.src_port
+        [address.to_s, port.to_i]
+      end
+
+    items.sort!.map!.with_index do |(address, port), index|
+      "#{index} : #{address} . #{port}"
+    end
+  end
+
+  def generate_lb_map_defs_ipv4(current_port)
+    generate_lb_map_defs(current_port) do |vm_port|
+      vm_port.load_balancer_vm.vm.private_ipv4
+    end
   end
 
   def generate_lb_map_defs_ipv6(current_port)
-    load_balancer.active_vm_ports
-      .select { |vm_port| vm_port.load_balancer_port.dst_port == current_port.dst_port }
-      .map
-      .with_index do |vm_port, index|
-        port = (vm_port.load_balancer_vm.vm.id == vm.id) ? vm_port.load_balancer_port.dst_port : vm_port.load_balancer_port.src_port
-        address = (vm_port.load_balancer_vm.vm.id == vm.id) ? vm.ephemeral_net6.nth(2) : vm_port.load_balancer_vm.vm.private_ipv6
-        "#{index} : #{address} . #{port}"
-      end
+    generate_lb_map_defs(current_port) do |vm_port|
+      (vm_port.load_balancer_vm.vm_id == vm.id) ? vm.ephemeral_net6.nth(2) : vm_port.load_balancer_vm.vm.private_ipv6
+    end
   end
 
   def generate_nat_rules(current_public_ipv4, current_private_ipv4)
--- a/spec/prog/vnet/update_load_balancer_node_spec.rb
+++ b/spec/prog/vnet/update_load_balancer_node_spec.rb
@@ -6,22 +6,34 @@ RSpec.describe Prog::Vnet::UpdateLoadBalancerNode do
   }
 
   let(:st) {
-    Strand.create_with_id(prog: "Vnet::UpdateLoadBalancerNode", stack: [{"subject_id" => vm.id, "load_balancer_id" => lb.id}], label: "update_load_balancer")
+    Strand.create(prog: "Vnet::UpdateLoadBalancerNode", stack: [{"subject_id" => vm.id, "load_balancer_id" => lb.id}], label: "update_load_balancer")
   }
+  let(:prj) {
+    Project.create(name: "test-prj")
+  }
+  let(:ps) {
+    Prog::Vnet::SubnetNexus.assemble(prj.id, name: "test-ps").subject
+  }
+
   let(:lb) {
-    prj = Project.create_with_id(name: "test-prj")
-    ps = Prog::Vnet::SubnetNexus.assemble(prj.id, name: "test-ps").subject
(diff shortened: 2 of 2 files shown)


prog/vnet/update_load_balancer_node.rb
```diff
@@ -47,7 +47,8 @@ def generate_lb_based_nat_rules
     end
 
     ipv4_prerouting = if load_balancer.ipv4_enabled?
-      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }.map do |vm_port|
+      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }
+        .sort_by { |vm_port| vm_port.load_balancer_port.src_port }.map do |vm_port|
```
jeremyevans, line 51: This looks like it has an N+1 query issue, even before this change, not just here but in the rest of the method.

Near the top of this method, you should add:

```ruby
load_balancer.active_vm_ports{|ds| ds.eager(:load_balancer_port, load_balancer_vm: :vm)}
```

That will make sure that `active_vm_ports` has eagerly loaded the related associations.

In general, for performance reasons, you should prefer direct column access over indirect column access through an associated object.  For example, `load_balancer_port.id` should be changed to `load_balancer_port_id`.  Only go through the association if the data you need is not in the current object (e.g. `src_port` access requires going through the association).

```suggestion
      load_balancer
        .active_vm_ports
        .uniq(&:load_balancer_port_id)
        .sort_by { |vm_port| vm_port.load_balancer_port.src_port }
        .map do |vm_port|
```

similar change below.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -129,22 +131,30 @@ def generate_lb_ip_set_definition(neighbor_vms)
       "elements = {#{neighbor_vms.map(&:private_ipv6).join(", ")}}"]
   end
 
-  def generate_lb_map_defs_ipv4(current_port)
-    load_balancer.active_vm_ports
+  def generate_lb_map_defs(current_port, address_selector)
+    items = load_balancer.active_vm_ports
```
jeremyevans, line 135: Likewise, in this method, you want to eagerly load the associations:

```ruby
load_balancer.active_vm_ports{|ds| ds.eager(:load_balancer_port, load_balancer_vm: :vm)}
```

You may not need the `vm` association eagerly loaded if you change from `vm_port.load_balancer_vm.vm.id` to `vm_port.load_balancer_vm.vm_id`.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -47,7 +47,8 @@ def generate_lb_based_nat_rules
     end
 
     ipv4_prerouting = if load_balancer.ipv4_enabled?
-      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }.map do |vm_port|
+      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }
```
furkansahin, line 50: instead of this, you can modify `active_vm_ports` to return a sorted result set by using `order: Sequel`. There are examples of it in the repo.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -129,22 +131,30 @@ def generate_lb_ip_set_definition(neighbor_vms)
       "elements = {#{neighbor_vms.map(&:private_ipv6).join(", ")}}"]
   end
 
-  def generate_lb_map_defs_ipv4(current_port)
-    load_balancer.active_vm_ports
+  def generate_lb_map_defs(current_port, address_selector)
```
jeremyevans, line 134: `address_selector` appears to be a proc.  In Ruby, you don't generally pass a proc as the final argument to a method.  Instead, you pass a block.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -129,22 +131,30 @@ def generate_lb_ip_set_definition(neighbor_vms)
...
+    items = load_balancer.active_vm_ports
       .select { |vm_port| vm_port.load_balancer_port.dst_port == current_port.dst_port }
-      .map
-      .with_index { |vm_port, index| "#{index} : #{vm_port.load_balancer_vm.vm.private_ipv4} . #{(vm_port.vm.id == vm.id) ? vm_port.load_balancer_port.dst_port : vm_port.load_balancer_port.src_port}" }
+      .map do |vm_port|
+        address = address_selector.call(vm_port)
```
jeremyevans, line 138: When passing a block to the method, you would use:

```suggestion
        address = yield vm_port
```

prog/vnet/update_load_balancer_node.rb
```diff
@@ -129,22 +131,30 @@ def generate_lb_ip_set_definition(neighbor_vms)
...
+
+    items.sort_by { |address, port| [address, port] }.map.with_index do |(address, port), index|
+      "#{index} : #{address} . #{port}"
+    end
+  end
+
+  def generate_lb_map_defs_ipv4(current_port)
+    generate_lb_map_defs(current_port, ->(vm_port) {
+      vm_port.load_balancer_vm.vm.private_ipv4
+    })
```
jeremyevans, line 151: To pass a block:

```suggestion
    generate_lb_map_defs(current_port) do |vm_port|
      vm_port.load_balancer_vm.vm.private_ipv4
    end
```

You'll want to make a similar change for the other call.

prog/vnet/update_load_balancer_node.rb
mohi-kalantari: fix it thanks

prog/vnet/update_load_balancer_node.rb
mohi-kalantari: thanks for the insight I fixed it. I just added the eager for load_balancer_port and removed the other because I was not using the load_balancer_vm or its vm.

prog/vnet/update_load_balancer_node.rb
mohi-kalantari: that ordering requires a field to order on. there was none actually i could rely on its sort output. If I want to go this way I need to add another field like created_at.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -47,7 +47,10 @@ def generate_lb_based_nat_rules
     end
 
     ipv4_prerouting = if load_balancer.ipv4_enabled?
-      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }.map do |vm_port|
+      load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port) }
```
jeremyevans, line 50: You are already calling `active_vm_ports` earlier in the method.  This will result in the association being loaded twice.  Please move the eager loading to the top of the method, as I specified in my previous review.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -47,7 +47,10 @@ def generate_lb_based_nat_rules
...
+        .uniq(&:load_balancer_port_id)
+        .sort_by { |vm_port| vm_port.load_balancer_port.src_port }
+        .map do |vm_port|
         port = vm_port.load_balancer_port
         ipv4_map_def = generate_lb_map_defs_ipv4(port)
         modulo = ipv4_map_def.count
```
jeremyevans, line 56: Please indent this further than the `map`.  Similar change below.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -60,7 +63,10 @@ def generate_lb_based_nat_rules
     end
 
     ipv6_prerouting = if load_balancer.ipv6_enabled?
-      load_balancer.active_vm_ports.uniq { |row| row.load_balancer_port.id }.map do |vm_port|
+      load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port) }
+        .uniq(&:load_balancer_port_id)
+        .sort_by { |vm_port| vm_port.load_balancer_port.src_port }
```
jeremyevans, line 68: This is the same code as in the IPv4 branch.  Please move the shared code above the `if`, and assign it to a local variable.  Use the local variable in both the IPv4 and IPv6 branches.

prog/vnet/update_load_balancer_node.rb
mohi-kalantari: I moved the eager to the first line calling the active_vm_ports which is the line populating the neighbor_vms. hope there is the correct place 😄

prog/vnet/update_load_balancer_node.rb
mohi-kalantari: My ruby language server (Ruby LSP) does not allow me and moves back the indention to its current place

prog/vnet/update_load_balancer_node.rb
```diff
@@ -46,8 +46,12 @@ def generate_lb_based_nat_rules
       fail ArgumentError, "Unsupported load balancer algorithm: #{load_balancer.algorithm}"
     end
 
+    active_vm_ports_uniq_by_port = load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port, load_balancer_vm: :vm) }
```
jeremyevans, line 49: For the third time, please move the `load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port, load_balancer_vm: :vm) }` to near the top of the method, so that you don't have duplicate queries (`load_balancer.active_vm_ports` is already called on line 38, so it should be above that).

Also, it looks like the `generate_lb_map_defs_ipv4/6` needs nic information for the vm to determine `private_ipv4/6`, so you want:

```ruby
load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port, load_balancer_vm: {vm: :nics}) }
```

prog/vnet/update_load_balancer_node.rb
jeremyevans: This looks fine now that the block code is indented further than its caller.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -129,22 +133,30 @@ def generate_lb_ip_set_definition(neighbor_vms)
...
       .select { |vm_port| vm_port.load_balancer_port.dst_port == current_port.dst_port }
-      .map
-      .with_index { |vm_port, index| "#{index} : #{vm_port.load_balancer_vm.vm.private_ipv4} . #{(vm_port.vm.id == vm.id) ? vm_port.load_balancer_port.dst_port : vm_port.load_balancer_port.src_port}" }
+      .map do |vm_port|
+        address = yield vm_port
+        port = (vm_port.load_balancer_vm.vm_id == vm.id) ? vm_port.load_balancer_port.dst_port : vm_port.load_balancer_port.src_port
+        [address.to_s, port.to_i]
+      end
+
+    items.sort_by { |address, port| [address, port] }.map.with_index do |(address, port), index|
```
jeremyevans, line 145: It doesn't look like `sort_by` is needed here.  Since the array was generated by  the `map` above, and it is not shared, you can use `sort!` and `map!`:

```suggestion
    items.sort!.map!.with_index do |(address, port), index|
```

prog/vnet/update_load_balancer_node.rb
mohi-kalantari: thanks for the tip. I fixed it.

fdr:
Note this commit doesn't _only_ change the tests/fix its flakiness, it hits production paths a well (I was about to deploy it because I saw a test flaked on this, but held my fire since it has a non-test side effect, now.)

PR #3063, opened 2025-03-27
Disallow access to pods GUA ips from internet

mohi-kalantari:
So far pods had both ULA and GUA ipv6 ips and the GUA ip was accessible from the internet. with this change, we add nft rules to each vm to only allow connections to be made from inside the pod and disallow all new connections coming to the GUA ip address.

This change is critical to not expose pods to the internet.

### DIFF
--- a/prog/kubernetes/provision_kubernetes_node.rb
+++ b/prog/kubernetes/provision_kubernetes_node.rb
@@ -70,6 +70,27 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
     nap 5 unless vm.strand.label == "wait"
 
     vm.sshable.cmd "sudo iptables-nft -t nat -A POSTROUTING -s #{vm.nics.first.private_ipv4} -o ens3 -j MASQUERADE"
+    vm.sshable.cmd("sudo nft --file -", stdin: <<TEMPLATE)
+table ip6 pod_access;
+delete table ip6 pod_access;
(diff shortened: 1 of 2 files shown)


fdr:
This one I will put to @furkansahin who I think you already discussed this with, and much of the backstory I got from you previously, but his nftables spidey sense is the strongest.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -70,6 +70,27 @@ def before_run
...
+delete table ip6 pod_access;
+table ip6 pod_access {
+  chain ingress_egress_control {
+    type filter hook forward priority filter; policy drop;
+    # allow access to the vm itself in order to not break the normal functionality of Clover and SSH
+    ip6 daddr #{vm.ephemeral_net6.nth(2)} ct state established,related,new counter accept
+    ip6 saddr #{vm.ephemeral_net6.nth(2)} ct state established,related,new counter accept
+
+    # not allow new connections from internet but allow new connections from inside
+    ip6 daddr #{NetAddr::IPv6Net.new(vm.ephemeral_net6.network, NetAddr::Mask128.new(80))} ct state established,related counter accept
```
furkansahin, line 84: are you casting again because of the netmask being 79? You don't need to, this can block the whole /79

furkansahin (approved):

PR #3064, opened 2025-03-27
Add gen_random_ubid_uuid database function for generating UUIDs in UBID format

jeremyevans:
This is passed the UBID type as an integer, which you can determine via `UBID.to_base32_n` (e.g. `UBID.to_base32_n("vm") # => 884`)
    
This could be used as the DEFAULT value for UUID primary keys.

### DIFF
--- a/migrate/20250327_gen_ubid_uuid.rb
+++ b/migrate/20250327_gen_ubid_uuid.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    # This function generates a pure-random (not timestamp-based) UUID,
+    # that is in valid UBID format.  It is passed the UBID type as an integer.
+    # You can get the type integer using `UBID.to_base32_n(prefix)`
+    # (e.g. `UBID.to_base32_n("vm") # => 884`).
+    #
+    # This could be used in the future as the DEFAULT value for uuid primary
+    # keys.
+    run <<~SQL
+      CREATE FUNCTION gen_random_ubid_uuid(ubid_type int) RETURNS uuid AS $$
+      DECLARE
+        r0 bigint;
+        r1 int;
+        r2 int;
+        r3 bigint;
+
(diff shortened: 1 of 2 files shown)


fdr:
I'm going to pass this one to @pykello, not only because he wrote ubids in general, but because his number theory is actually fluent.

migrate/20250327_gen_ubid_uuid.rb
```diff
@@ -0,0 +1,44 @@
...
+        r2 int;
+        r3 bigint;
+
+        p1 text;
+        p2 text;
+      BEGIN
+        -- 48 bit random
+        r0 = floor(281474976710656 * random()::numeric)::bigint;
+        -- 4 bit version + 2 bit random + 10 bit type
+        r1 = ((32 + floor(2 * random()::numeric)::integer) << 10) | ubid_type;
```
pykello, line 26: Here we want 2 bits of random, but doesn't `floor(2 * random()::numeric)::integer` generate 0 or 1 (i.e. 1 bit of random)?

migrate/20250327_gen_ubid_uuid.rb
```diff
@@ -0,0 +1,44 @@
...
+        --  2 bit variant + 2 bit random
+        r2 = 8 + floor(2 * random()::numeric);
```
pykello, line 28: * Same comment as above about number of random bits
* `floor(numeric)`'s return type is `numeric`. Should we cast to `integer`?

migrate/20250327_gen_ubid_uuid.rb
```diff
@@ -0,0 +1,44 @@
...
+        -- 60 bit random
+        r3 = floor(1152921504606846976 * random()::numeric);
```
pykello, line 30: * `floor(numeric)`'s return type is `numeric`. Should we cast to `bigint`?

pykello (approved):

PR #3065, opened 2025-03-27
Remove old login-based API authentication

jeremyevans:
This makes personal access token the only possible authentication for the API.

For API requests without a personal access token, they get a specific error that the credentials are missing.  I chose to use 401 for this, because that seemed more appropriate than 400.

In the specs, `login_api` now takes no arguments.  It just sets to use personal access tokens for requests.  This was chosen over removing the method completely, because there are a significant number of specs that want to test for unauthenticated access.  I'm not sure whether such specs are actually useful, since they all hit the same condition at the top of the main route block.  We should consider changing the specs in the future to provide an access token that is valid, but does not grant access to the objects, in order to test for specific authorization. We already do this in some cases, but not in others.

With this change, the only error that Rodauth can generate for API requests is for invalid credentials, so simplify the json_response_body method, and generate the response body up front so we don't need to generate it per-request.

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -243,54 +243,28 @@ class Clover < Roda
     end
   end
 
-  require "rodauth"
   require_relative "rodauth/features/personal_access_token"
 
   plugin :rodauth, name: :api do
-    enable :argon2, :json, :jwt, :active_sessions, :login, :personal_access_token
+    enable :json, :personal_access_token
 
     only_json? true
-    use_jwt? { !use_pat? }
 
-    # Converting rodauth error response to the common error format of the API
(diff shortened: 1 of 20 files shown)


enescakir:
Should we update our documentation first?

jeremyevans:
> Should we update our documentation first?

Makes sense.  I updated https://github.com/ubicloud/documentation/pull/38 to change the login-based instructions to use the personal access token instructions.

I also updated this to remove `/login` from `openapi.yml`, since it is no longer valid.

enescakir (approved):

PR #3066, opened 2025-03-27
Polish k8s UI

aamederen:
Polish K8s Create Form

  - Add pricing details for CP and DP
  - Describe 1 node vs 3 node CP
  - Make Worker count >= CP count
  - Bump Worker count max to 10
  
Validate that k8s worker count >= 0 in the backend

Pass the `required` attribute to the `select` input in the creation form


aamederen:
resim

mohi-kalantari (approved):

PR #3070, opened 2025-03-28
Don't wait to collect telemetry for the runner when for all SSH errors

enescakir:
- **Group SSH connection errors**
  SSH commands might raise different types of exceptions if there’s
  something wrong with the connection. We catch these exceptions in
  multiple places, so I grouped them together.
  

- **Don't wait to collect telemetry for the runner when for all SSH errors**
  We save runner script logs and serial logs if the job isn't successful
  for further investigation. We also collect DockerHub limits to monitor
  usage.
  
  If the host or VM is unavailable, it raises an exception other than
  Sshable::SshError, causing the runner program to get stuck.
  
  If they are down, we shouldn't wait for the runner to collect telemetry.
  
  I also extracted it to a helper method, which made the tests more
  readable.

### DIFF
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -11,6 +11,14 @@ class Sshable < Sequel::Model
     enc.column :raw_private_key_2
   end
 
+  SSH_CONNECTION_ERRORS = [
+    Net::SSH::Disconnect,
+    Net::SSH::ConnectionTimeout,
+    Errno::ECONNRESET,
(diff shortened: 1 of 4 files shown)


jeremyevans (approved):
Looks good!

fdr:
Hit this one in prod again, mebbe merge it soon since it seems to trigger semi-regularly since authorship circa March 26th

PR #3073, opened 2025-03-28
Rewrite Scheduling::Dispatcher (bin/respirate)

jeremyevans:
Previous, respirate spun up a thread and created 4 IOs per strand run, and used IO.select to handle timeouts.

This changes respirate to use a fixed-size thread pool, fed using a single queue (the strand queue).  The thread pool is made up of pairs of strand/apoptosis threads.  The strand threads pop from the strand queue.  Once they get a strand, the signal their related apoptosis thread via their shared start queue.  The apoptosis thread does a timed pop of the finish queue (timed pop is a new feature added in Ruby 3.2).  If the strand thread doesn't finish the strand run and push to the finish queue before the apoptosis timeout, then the apoptosis thread kills the process.

This changes the database query to get strands to run to use a prepared statement, to save parsing/planning time.  The database query now avoids selecting strands that the thread pool is currently running.


enescakir (commented):
It looks good to me overall, but @fdr should review it since he knows the most about it.

bin/respirate
```diff
@@ -31,9 +31,15 @@ Strand.dataset.insert_conflict.insert(id: "08200dd2-20ce-833a-de82-10fd5a082bff"
...
-loop do
-  d.start_cohort
-  next if d.wait_cohort > 0
-  duration_slept = sleep 1
-  Clog.emit("respirate finished sleep") { {sleep_duration_sec: duration_slept} }
+until d.shutting_down
+  if d.start_cohort && !d.shutting_down
+    # Only sleep if not shutting down and there were no strands in previous scan
+    # Note that this results in an up to a 1 second delay to pick up new strands,
+    # but that is an accept tradeoff, because we do not want to busy loop the
```
bsatzger, line 38:
```suggestion
    # but that is an acceptable tradeoff, because we do not want to busy loop the
```

scheduling/dispatcher.rb
```diff
@@ -1,93 +1,228 @@
...
+    @thread_data = Array.new(pool_size) do
+      start_thread_pair(@strand_queue)
+    end
+
+    # A prepared statement to get the strands to run.  A prepared statement
+    # is used to reduce parsing/planning work in the database.
+    @strand_ps = Strand
+      .where(
+        Sequel.|({lease: nil}, Sequel[:lease] < Sequel::CURRENT_TIMESTAMP) &
+        (Sequel[:schedule] < Sequel::CURRENT_TIMESTAMP) &
```
bsatzger, line 57: A simple approach to deal with the 1 sec sleep delays could be to slightly relax this condition and include strands whose schedule will be due soon (in 1 sec or less). My understanding is that our prog should be ok with being executed slightly earlier than requested. The `order_by` makes sure that strands that are already due have higher priority. But I see that this has relatively far reaching semantic consequences and could introduce unexpected behavior.

scheduling/dispatcher.rb
```diff
@@ -1,93 +1,228 @@
...
 
-  def wait_cohort
-    if shutting_down
-      Clog.emit("Shutting down.") { {unfinished_strand_count: @notifiers.length} }
-      Kernel.exit if @notifiers.empty?
+  # Find strands that need to be run, and push each onto the
+  # strand queue.  This can block if the strand queue is full,
+  # to allow for backoff in the case of a busy thread pool.
+  def start_cohort
+    strand_queue = @strand_queue
```
bsatzger, line 217: Let's say we have this situation (schedule and strand id):
00:00:00 strand 1
NOW
00:00:01 strand 2
00:00:01 strand 3
...
00:00:01 strand 100

scan only returns "strand 1" but within a second many more strands are due. "strand 1" runs for one minute. will strands 2-100 be executed at ~00:00:01  or at ~00:01:00?

jeremyevans: strands 2-100 would be enqueued to the job pool around `~00:00:01` (though enqueing may block if it exceeds the queue buffer size).  When strands start processing depends on thread pool size.  With a thread pool of size 1, strand 2 would start at `~00:01:00`, with others following that.  With a thread pool size of 2 or more, strand 2 would start around `~00:00:01`.  I'm guessing we use a decent number of threads for the production thread pool.

fdr (approved):
I think we've beaten this one up enough, I do have the conflict resolved version in https://github.com/ubicloud/ubicloud/pull/3351

jeremyevans:
Cherry-picked the conflict resolved version and rebased it on master.

PR #3076, opened 2025-03-28
Bump maximum pagination page size to 1000, and make that the default

jeremyevans:
Actually testing over 1000 results in the specs is way too slow, so just test that a page_size lower than the number of objects is respected.  This doesn't test the upper clamp, but it's not necessary to do so for coverage, and I don't think the benefit of such testing is worth the large performance hit.

CC @geemus

### DIFF
--- a/lib/pagination.rb
+++ b/lib/pagination.rb
@@ -3,11 +3,11 @@
 module Pagination
   def paginated_result(start_after: nil, page_size: nil, order_column: nil)
     model = @opts[:model]
-    page_size ||= 10
+    page_size ||= 1000
     order_column_sym = (order_column || "id").to_sym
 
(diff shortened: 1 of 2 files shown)


geemus (approved):

geemus:
I'm not sure where the source for the related documentation you shared lives, but that should be updated as well to note the change to the default.

jeremyevans:
> I'm not sure where the source for the related documentation you shared lives, but that should be updated as well to note the change to the default.

I submitted a pull request for it earlier today: https://github.com/ubicloud/documentation/pull/51

enescakir (approved):
I hope we don’t have any N+1 queries in our listing endpoints.

PR #3079, opened 2025-03-29
Openapi references

geemus:
Complete using references for id or name to match current API. I also identified some bugs/gaps, but will address those separately.

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -183,8 +183,6 @@ class Clover < Roda
       when OpenAPIParser::InvalidPattern
         pattern = e.original_error.instance_variable_get(:@pattern)
         value = e.original_error.instance_variable_get(:@value)
-        code = 404
-        type = "ResourceNotFound"
         message = "Parameter #{value.inspect} does not match pattern #{pattern}"
       when OpenAPIParser::NotExistPropertyDefinition
         keys = e.original_error.instance_variable_get(:@keys)
@@ -208,7 +206,7 @@ class Clover < Roda
       message = "Sorry, we couldn’t process your request because of an unexpected error."
     end
 
-    raise e if Config.test? && e.is_a?(Committee::Error) && code != 404
+    raise e if Config.test? && e.is_a?(Committee::Error) && code != 400
 
     response.status = code
     next if code == 204
--- a/clover.rb
+++ b/clover.rb
@@ -181,17 +181,11 @@ class Clover < Roda
         message = "Validation failed for following fields: body"
         details = {"body" => "Request body isn't a valid JSON object."}
       when OpenAPIParser::InvalidPattern
-        if e.original_error.instance_variable_get(:@reference) == "#/components/schemas/Reference"
-          value = e.original_error.instance_variable_get(:@value)
-          pattern = e.original_error.instance_variable_get(:@pattern)
-          code = 404
-          type = "ResourceNotFound"
-          message = "Parameter #{value.inspect} does not match pattern #{pattern}"
-        else
-          code = 500
-          type = "UnexceptedError"
-          message = "Sorry, we couldn’t process your request because of an unexpected error."
-        end
+        pattern = e.original_error.instance_variable_get(:@pattern)
+        value = e.original_error.instance_variable_get(:@value)
+        code = 404
+        type = "ResourceNotFound"
+        message = "Parameter #{value.inspect} does not match pattern #{pattern}"
       when OpenAPIParser::NotExistPropertyDefinition
         keys = e.original_error.instance_variable_get(:@keys)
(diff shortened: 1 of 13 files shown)


jeremyevans (commented):
This looks good, but the specs apparently need to be fixed before merging.  Maybe we could benefit from support in the error_handler to handle cases like this?  IMO, providing an invalid pattern in a path segment should result in a 404 error, since the requested entity doesn't/can't exist.

geemus:
@jeremyevans thanks for taking a look. I had seen the error, but just hadn't quite figured out how I wanted to fix it and found the time to do it. I agree that something more generic/general seems worth exploring, as opposed to a more whack-a-mole approach to these particular errors. `error_handler` seems like a good possibility for that, and I'll plan to explore that when I next have a bit of time to dig in.

geemus:
Ok, good/bad news I suppose.

Good: I was able to find a way to generically translate cases where an invalid reference is passed in to 404.

Bad: It traded the existing test breakages for a new set...

I'm happy to work on fixing these as well, but wanted to check-in on what the proper fix is because it seems open to some interpretation.

1. Deletes with invalid names currently return 204 (as all deletes return 204 all the time I believe). I think that behavior made sense previously, but it seems more ambiguous now. If you know that there could never be an object with that name, 404 seems reasonable (to me), even if it maintains 204 for a potentially valid object.
1. It gets weird around create with a specified name that is invalid. Historically this returned an error that looked like errors you would see when body params didn't match (not a 404, but an invalid field error, though this isn't actually a field per se). Only one test appears to encounter this particular intersection, but it would be true across the API. Probably having the body-like error is more correct, but I'll need to think about how to properly discriminate this case (I welcome suggestions).
1. There is a test for a not authorized case on subnets which uses an invalid reference to get to an error state. This now translates into a 404 instead (since there could never be such a reference). I think the new behavior is probably the correct one, so probably the fix is changing it to a valid reference (which should still error since it wouldn't exist), but I just wanted to double check I'm not missing anything.

I think all the failures fall into one of those buckets (including one CLI test that fails because of 204 vs 404), but do let me know if I've overlooked any.

geemus:
Just let me know if you want me to rebase/cleanup the history, it was a bit meandering to get to the final version.

geemus:
I think this should be largely ready to go, though it appears I'm getting coverage related breakage. I've not needed to fix that previously, so could use some guidance on how to proceed.

geemus:
Updated with thanks to @fdr for identifying the regex weirdness that led to coverage issues.

I also:
- changed invalid param from 404 to 400 as per @jeremyevans suggestion
- changed testing to rely on a central/shared committee test for invalid params (and removed the scattered one-off tests for this edge case). This brings it more inline with other committee error testing and avoids the one-off treatment for this case that was there previously.

Hoping this should now be good to go (presuming tests are now passing, we'll see soon).

fdr:
alright so I'm going through this, no substantive complaints, my main thinking is if there any risk-reducing deployment phases, or if it needs to all happen in one go.

One minor thing though, I notice you have a special commit to capitalize the regexp, which is what made it not-match in the first place. Was there much thought to that or was it closer to an intended quick-fix? Some of my changes can be dropped by dropping that commit.

I'd also probably squish some of our ruby changes.  The ones I sent to you were micro-fine in the proposal phase, but also, under-justified for recall if someone `git blame`s them.

geemus:
I'm uncertain on how best to de-risk. I think it might be challenging to break apart into pieces that all play nice individually, as you suggest. Happy to discuss though.

The regex case-insensitivity was more of a quick-fix, though I no longer recall the specifics. I thought at the time that the ruby was case-insensitive, and was trying to match. But I can't find such a regex now, so it's fuzzy at best. In any event, I think just dropping that (and any follow on commits that were implied as fixes) should be fine.

For the one-off tests, I changed the underlying code such that it was raising a new/different error (ie committee was giving back the error instead of other mechanisms) and then played whack-a-mole by nuking all the tests that failed because they were looking for a mismatched error (in particular this was pretty clear due to expecting 404 vs 400).

I'm open to squashing/removing some of the error slicing. Originally I had added 404 as an exception to avoid re-reraising to avoid breaking the tests that I described removing in the paragraph prior to this one. Once I had better integrated things so that these errors functioned (and were tested) like other committee errors, it was no longer necessary to have an exception, so I shifted it back to where it started. That likely means it could get revised/squashed to no longer include parts that change it away and then back again if so desired.

fdr:
I think I'll probably do it in one go and erase some intermediate states then today, but keep the cumulative diff at zero.

fdr:
closing as duplicate of #3150, which makes no cumulative changes but removes some intermediate states and elaborates a couple of commit messages.

PR #3083, opened 2025-03-31
Handle case when user attempts to add a user to a project when they already have access

jeremyevans:
This raises Sequel::UniqueConstraintViolation, so rescue that and show an appropriate error message.  As the specs use transactions, you need to use a savepoint in the specs to avoid issues.  This uses the savepoint: :only option, so it will open a savepoint during the specs, but not create a new transaction outside of the specs.

### DIFF
--- a/routes/project/user.rb
+++ b/routes/project/user.rb
@@ -42,11 +42,12 @@ class Clover
           end
 
           if (user = Account.exclude(status_id: 3)[email: email])
-            begin
-              DB.transaction(savepoint: :only) do
-                user.add_project(@project)
-              end
-            rescue Sequel::UniqueConstraintViolation
+            result = DB[:access_tag]
+              .returning(:hyper_tag_id)
+              .insert_conflict
+              .insert(hyper_tag_id: user.id, project_id: @project.id)
+
+            if result.empty?
               flash["error"] = "The requested user already has access to this project"
               r.redirect "#{@project.path}/user"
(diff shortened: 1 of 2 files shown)


routes/project/user.rb
```diff
@@ -42,7 +42,15 @@ class Clover
           end
 
           if (user = Account.exclude(status_id: 3)[email: email])
-            user.add_project(@project)
+            begin
+              DB.transaction(savepoint: :only) do
+                user.add_project(@project)
```
fdr, line 47: could this be done with insert...on conflict? considerably less going on there than a savepoint database-side, and possibly less code as well.

fdr: It might look a bit like this (LLM code incoming, though I had to give it some pointed advice)

```ruby
result = DB[:projects_users]
           .insert_conflict(target: [:user_id, :project_id])
           .insert(user_id: user.id, project_id: @project.id)

if result.nil?
  flash["error"] = "The requested user already has access to this project"
  r.redirect "#{@project.path}/user"
end
```

jeremyevans: It can be handled with insert on conflict.  Note that `savepoint: :only` does not use a savepoint or transaction in production.  It uses a savepoint during testing, because tests run inside a transaction.  I've pushed a change to use insert on conflict.

fdr (approved):

PR #3086, opened 2025-03-31
Add cli-sync rake task to sync cli folder to cli repository

jeremyevans:
The cli repository should be checked out in ../cli.  We'll be using the cli repository only to publish releases, actual development will still occur in the ubicloud repository.

While here, add a basic README for the cli repository, which will also be synced.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -288,9 +288,22 @@ cli_version = lambda do
   File.read("cli/version.txt").chomp
 end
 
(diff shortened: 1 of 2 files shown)


fdr (approved):
think that's fine, wonder if we can use a special branch or something to avoid crawling out of the repository to a sibling directory but maybe given how we manage the `documentation` repository, we should stick with one way (if it's several sibling git repositories)

PR #3091, opened 2025-04-01
Deduplicate service ports by 'port' value, keeping first occurrence

mohi-kalantari:
Previously customers were able to create multiple LoadBalancer services
with duplicate ports which would cause issue for us.

The proper way to do it is installing a webhook inside the clustere and
not allow the creation of the duplicate ports.

As a quick fix for now, we would sort all LoadBalancer services based on
the creationTimestamp and use the first occurance and ignore the other ones.

### DIFF
--- a/lib/kubernetes/client.rb
+++ b/lib/kubernetes/client.rb
@@ -12,11 +12,18 @@ class Kubernetes::Client
   end
 
   # Returns a flat array of [port, nodePort] pairs from all services
+  # Deduplicates based on the 'port' value, keeping only the first occurrence
   # Format: [[src_port_0, dst_port_0], [src_port_1, dst_port_1], ...]
   def lb_desired_ports(svc_list)
-    svc_list.flat_map do |svc|
-      svc.dig("spec", "ports")&.map { |port| [port["port"], port["nodePort"]] } || []
+    seen_ports = {}
+    sorted = svc_list.sort_by { |svc| svc["metadata"]["creationTimestamp"] }
+    sorted.each do |svc|
+      svc.dig("spec", "ports")&.each do |port|
+        seen_ports[port["port"]] ||= port["nodePort"]
+      end
     end
+
+    seen_ports.to_a
   end
 
   def load_balancer_hostname_missing?(svc)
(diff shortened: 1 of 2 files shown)


lib/kubernetes/client.rb
```diff
@@ -12,11 +12,24 @@ def service_deleted?(svc)
   end
 
   # Returns a flat array of [port, nodePort] pairs from all services
+  # Deduplicates based on the 'port' value, keeping only the first occurrence
   # Format: [[src_port_0, dst_port_0], [src_port_1, dst_port_1], ...]
   def lb_desired_ports(svc_list)
-    svc_list.flat_map do |svc|
-      svc.dig("spec", "ports")&.map { |port| [port["port"], port["nodePort"]] } || []
+    seen_ports = {}
+    sorted = svc_list.sort_by do |svc|
```
aamederen, line 19: as discussed, let's sort this by the create time of the service so that the first service gets the assignment all the time.

mohi-kalantari: fixed it. thanks.

lib/kubernetes/client.rb
```diff
@@ -12,11 +12,22 @@ def service_deleted?(svc)
...
+    sorted.flat_map do |svc|
+      svc.dig("spec", "ports")&.map do |port|
+        src = port["port"]
+        dst = port["nodePort"]
+        unless seen_ports[src]
+          seen_ports[src] = true
+          [src, dst]
+        end
+      end || []
+    end.compact
```
jeremyevans, line 30: I haven't tested this, but assuming it works, it seems like a simpler implementation:

```suggestion
    seen_ports = {}
    sorted = svc_list.sort_by { |svc| svc["metadata"]["creationTimestamp"] }

    sorted.each do |svc|
      svc.dig("spec", "ports")&.each do |port|
        seen_ports[port["port"]] ||= port["nodePort"]
      end
    end
    
    seen_ports.to_a
```

mohi-kalantari: It's the same I tested. thanks.

aamederen (approved):

PR #3093, opened 2025-04-01
Use FOR UPDATE SKIPPED LOCKED in Strand#take_lease_and_reload

jeremyevans:
This can prevent the query blocking if the strand is locked. Here is a test showing this:

```ruby
id = "08200dd2-20ce-833a-de82-10fd5a082bff"
q1 = Queue.new
q2 = Queue.new
t = Thread.new do
  DB.transaction(rollback: :always) do
    Strand[id].this.update(parent_id: :parent_id)
    q1.push nil
    q2.pop(timeout: 5)
    sleep 1
  end
end
p Time.now
r = DB.transaction(rollback: :always) do
  q1.pop
  Strand::TAKE_LEASE_AND_RELOAD_PS.call(id:)
end
p Time.now
q2.push nil
p Time.now
t.join
p Time.now
```

```
-- Without FOR UPDATE SKIPPED LOCKED
2025-04-01 14:50:49.856485021 -0700
-- Main thread blocked for 6 seconds until worker thread finishes
2025-04-01 14:50:55.859276833 -0700
2025-04-01 14:50:55.859350615 -0700
2025-04-01 14:50:55.859368655 -0700

-- With FOR UPDATE SKIPPED LOCKED
2025-04-01 14:51:17.518616169 -0700
-- Main thread not blocked
2025-04-01 14:51:17.535050745 -0700
2025-04-01 14:51:17.535089917 -0700
2025-04-01 14:51:18.53660402 -0700
```

Since SKIPPED LOCKED is only supported for SELECT and not UPDATE, have UPDATE use a SELECT subquery.

While here, use a prepared statement for the query.  This is one of the 2 most common respirate queries, and it makes sense to use a prepared statement to optimize it.

Also, have the query return `*` instead of `lease`.  This allows for skipping the `Strand#reload` call, since that does the same thing.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -19,19 +19,34 @@ class Strand < Sequel::Model
     UBID.decode(ubid)
   end
 
+  # :nocov:
+  schedule = if Config.development?
+    Sequel.function(:least, 5, :try)
+  # :nocov:
+  else
+    Sequel.function(:least, Sequel[2]**Sequel.function(:least, :try, 20), 600) * Sequel.function(:random)
+  end
+
+  TAKE_LEASE_AND_RELOAD_PS = DB[:strand]
+    .returning
+    .where(
+      Sequel[id: DB[:strand].select(:id).where(id: :$id).for_update.skip_locked, exitval: nil] &
+        (Sequel[lease: nil] | (Sequel[:lease] < Sequel::CURRENT_TIMESTAMP))
+    )
+    .prepare(:update, :take_lease_and_reload,
+      lease: Sequel::CURRENT_TIMESTAMP + Sequel.cast("120 seconds", :interval),
+      try: Sequel[:try] + 1,
+      schedule: Sequel::CURRENT_TIMESTAMP + (schedule * Sequel.cast("1 second", :interval)))
+
   def take_lease_and_reload
-    affected = DB[<<SQL, id].first
-UPDATE strand
-SET lease = now() + '120 seconds', try = try + 1, schedule = #{SCHEDULE}
-WHERE id = ? AND (lease IS NULL OR lease < now()) AND exitval IS NULL
-RETURNING lease
-SQL
+    affected = TAKE_LEASE_AND_RELOAD_PS.call(id:).first
     return false unless affected
+    # Also operate as reload query
+    @values = affected
     lease_time = affected.fetch(:lease)
     verbose_logging = rand(1000) == 0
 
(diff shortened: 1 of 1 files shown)


model/strand.rb
```diff
@@ -19,19 +19,34 @@ def subject
     UBID.decode(ubid)
   end
 
+  # :nocov:
+  schedule = if Config.development?
```
enescakir, line 23: Should we remove the `SCHEDULE` constant then?

jeremyevans: Yep, forgot about that.  I'll push a commit to remove it.  Probably also should only prepare the statement if it is needed.

jeremyevans: Pushed a commit for that.

model/strand.rb
```diff
@@ -19,19 +19,34 @@ def subject
...
+  TAKE_LEASE_AND_RELOAD_PS = DB[:strand]
+    .returning
+    .where(
+      Sequel[id: DB[:strand].select(:id).where(id: :$id).for_update.skip_locked, exitval: nil] &
+        (Sequel[lease: nil] | (Sequel[:lease] < Sequel::CURRENT_TIMESTAMP))
+    )
+    .prepare(:update, :take_lease_and_reload,
+      lease: Sequel::CURRENT_TIMESTAMP + Sequel.cast("120 seconds", :interval),
+      try: Sequel[:try] + 1,
+      schedule: Sequel::CURRENT_TIMESTAMP + (schedule * Sequel.cast("1 second", :interval)))
```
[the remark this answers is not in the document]
fdr: (Pry features I use: `cd`, `show-source`)

fdr (approved):
we don't exactly have a hard rule against prefixing with `.`, though I do think it raises some risks with pasting code, even outside pry, because if you don't paste the next byte what the prefix, intended to be a subexpression will evaluate.
```
foo = ' '
.strip()
```

To work in irb (binding `foo` to an empty string rather than a space), this relies on convenient but sort of complicated multiline paste stuff. Although I think pry makes the matter much, much worse, I sort of don't love how semantics change depending where the EOF (or similar) goes.

model/strand.rb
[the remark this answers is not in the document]
jeremyevans: My preference is to use trailing `.`, and not leading `.`. If you would prefer trailing `.`, it seems unwise to enforce the `Layout/DotPosition` cop, which forces leading `.`.

fdr (approved):

PR #3103, opened 2025-04-03
Add support for respirate partitioning

jeremyevans:
This adds a partition keyword argument to the dispatcher, which accepts a range of UUID strings.  If provided, the dispatcher runs in partitioning mode.  In partitioning mode:

* The default scan query returns only strands inside the partition.

* There is a scan_old method that can be used to return strands where schedule is more than 5 seconds in the past.  Respirate runs this scan every 5 seconds.  This handles failure of one paritioned respirate process gracefully.  Other respirate processes will pick up the slack, just with an additional 5 seconds of latency.

The advantage of partitioning is that multiple respirate processes no longer try to operate on the same strands, so this reduces contention.

For bin/respirate usage, provide 2 arguments to enable partitioning. The first argument is the number of partitions, and the second is the partition number for each partition (1 based). So:

  partition 4 2

Divides the UUID space into 4 partitions, and operates on the 2nd of 4 ("80000000-0000-0000-0000-000000000000"..."c0000000-0000-0000-0000-000000000000").

For Heroku usage, the second argument is optional if the DYNO environment variable is present, in which case it will try to parse the DYNO value to determine the current partition.

Currently, up to 16 partitions are supported, and it is trivial to go from 16 to 256 if needed.

This should be compatible with an index-only scan if we have an index on `strand(schedule, lease, id) WHERE exitval IS NULL`, once the `lease IS NULL` condition is removed.  The scan_old query should also be able to use the same index for an index-only scan.

While here, add a minor spec fix to shutdown the dispatcher when using custom dispatcher options.

### DIFF
--- a/bin/respirate
+++ b/bin/respirate
@@ -1,9 +1,51 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+num_partitions, partition_num = ARGV
+if num_partitions
+  unless partition_num
+    if ENV["DYNO"]
+      partition_num = ENV["DYNO"].split(".", 2)[1]
+    else
+      warn "must provide number of partitions as second argument"
+      exit 1
+    end
+  end
+
+  num_partitions = Integer(num_partitions)
+  partition_num = Integer(partition_num)
+
+  unless (1..16).cover?(num_partitions)
+    warn "invalid number of partitions: #{num_partitions}"
+    exit 1
+  end
+
+  unless (1..num_partitions).cover?(partition_num)
+    warn "invalid partition number #{partition_num}"
+    exit 1
+  end
+
+  partition_num -= 1
+
+  hex_chars = 1
+  chars = (0...16**hex_chars).map { ("%x" % _1).rjust(hex_chars, "0") }
+  chars_per_partition = chars.size / num_partitions.to_r
+  start_char = chars[(chars_per_partition * partition_num).to_i]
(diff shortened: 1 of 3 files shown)


byucesoy:
Currently, if a respirate process crashes, another one can pick up the work and no strand would starve. If we do partitioning, then we risk not running some strands for a long time. In ideal scenario, respirate crash would be rare, but this is not the case at the moment.

jeremyevans:
> Currently, if a respirate process crashes, another one can pick up the work and no strand would starve. If we do partitioning, then we risk not running some strands for a long time. In ideal scenario, respirate crash would be rare, but this is not the case at the moment.

This has specific code to pick up strands for other partitions if they are old.  The current threshold is over 5 seconds old.  We could reduce it if that is too much time.

fdr (approved):

jeremyevans:
After I rebased this on current main, it passed the E2E tests: https://github.com/ubicloud/ubicloud/actions/runs/15597820182

bin/respirate
```diff
@@ -82,7 +56,7 @@ DB.synchronize do
 
     break if d.shutting_down
 
-    if partition && next_old_strand_scan < Time.now
+    if next_old_strand_scan&.<(Time.now)
```
fdr, line 59: slick.

fdr (approved):

PR #3105, opened 2025-04-03
Fix LoadBalancer cert renewal empty cert bug

furkansahin:
Recently we have hit the issue of certificate being empty in inference endpoints. This happened after a cert renewal. The bug is that, the moment we create a new CertNexus, created_at column is set by default but the cert itself is not finalized just yet. Since we forgot to check that column, we thought the cert is finalized and went into broadcasting the new certificate to the metadata-endpoint. Now, checking if the cert column is properly filled, we essentially wait for the new certificate to be finalized before going into distribution mode.

### DIFF
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -117,11 +117,11 @@ class LoadBalancer < Sequel::Model
   def need_certificates?
     return true if certs_dataset.empty?
 
-    certs_dataset.where { created_at > Time.now - 60 * 60 * 24 * 30 * 2 }.empty?
+    certs_dataset.where { created_at > Time.now - 60 * 60 * 24 * 30 * 2 }.exclude(cert: nil).empty?
   end
 
   def active_cert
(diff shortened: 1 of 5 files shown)


model/load_balancer.rb
```diff
@@ -117,11 +117,11 @@ def dns_zone
   def need_certificates?
     return true if certs_dataset.empty?
 
-    certs_dataset.where { created_at > Time.now - 60 * 60 * 24 * 30 * 2 }.empty?
+    certs_dataset.where { created_at > Time.now - 60 * 60 * 24 * 30 * 2 }.exclude(cert: nil).empty?
```
bsatzger, line 120: is it guaranteed that other required fields are also set (like key) when the cert is set?

bsatzger (approved):

model/load_balancer.rb
furkansahin: Yes, cert is the very last field we get following the acme challenge.

PR #3108, opened 2025-04-03
Fix SELECT option handling in UI

aamederen:
Preserve selected option when parent changes

Drop <option> elements instead of hiding

Remove cp_nodes dependency from worker_nodes in k8s form

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -1,3 +1,4 @@
+let select_options = {}
 $(function () {
   setupAutoRefresh();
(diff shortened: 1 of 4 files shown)


byucesoy (commented):
Adding, removing entities could be slow due to DOM modification. Also I'm not fan of hiding elements in 2 different ways. If we don't have any option, then there is nothing to do, but [this approach](https://stackoverflow.com/a/17158417) seems promising. Does it work?

PR #3110, opened 2025-04-04
Switch firewall creation form to a simpler approach using forme

jeremyevans:
This uses the forme library to build the form. Forme is an HTML form library that has integrations for both Roda and Sequel (it's older than Roda, first release was in 2011).  In addition to simplifying HTML form creation, another benefit is it handles CSRF tags implicitly.

For the javascript showing and hiding of inputs based on selected options, this adds a simpler 10 line implementation to replace the ~60 line current implementation.  The new implementation does not require any inline javascript to set option_tree, option_children, and option_dirty. Instead:

* The form lists what types of dependencies it supports (`dependencies: "location"`)

* The radio button set to choose the parent object marks itself as a provider (`provides: "location"`).

* Any elements that have a dependency on a parent have 2 classes, one of which is the value of the radio button for the related dependency, and the other is `depends-` followed by the dependency name (`class: "depends-location #{_1.location_id}"`).

On a change to a radio button, all elements with the `depends-*` class are hidden, and then the ones with a class matching the value of the radio button are shown.

This approach also does not require the use of the OptionTreeGenerator or the resource_creation_form, and instead of using the `view/components/form` files, uses the forme library for generating the HTML.  Customization of the forms for Ubicloud's needs is done in a single file, `helpers/clover_forme_config.rb`.

Currently, this doesn't use a model-based form.  Using a model-based form would likely make things even simpler, and could be done later, but requires additional refactoring.


byucesoy:
Would this keep existing functionality of option tree generation and content generation? Few things from top of my mind:
- Can we display different prices for different options? standard-2 costs more in US and the price difference affects what we show in other components like HA selection.
- With option tree, the goal was keeping a single source of truth that can be used for both back end and front end. If we start to define dependencies on the front end code, we would need an equivalent of it in back end as well. Firewalls are simple enough that it comes with one dependency. In PostgreSQL form almost every component depends another one.
- Even though we are not using at the moment, my idea with option tree is that it can be used to eliminate majority of validation code. It simply generates all valid options. If we keep dependency relationships on front end code, we would lose that ability.

jeremyevans:
> Would this keep existing functionality of option tree generation and content generation? Few things from top of my mind:

This PR doesn't delete such functionality, but it removes its use for the firewall creation form (it does remove `generate_firewall_options`, as that is no longer used).  However, assuming we prefer this direction, the eventual goal would be remove the option tree generator (possibly not, see below) and content generator.  @furkansahin told me the complexity of the option tree generator and content generator resulted in additional 30% of development time for UBI-on-AWS.
 
> * Can we display different prices for different options? standard-2 costs more in US and the price difference affects what we show in other components like HA selection.

The existing content generator handles this by using a separate input/card for each price, only displaying currently valid options.  The approach used in this PR would work in a similar way, using significantly less code.

> * With option tree, the goal was keeping a single source of truth that can be used for both back end and front end. If we start to define dependencies on the front end code, we would need an equivalent of it in back end as well. Firewalls are simple enough that it comes with one dependency. In PostgreSQL form almost every component depends another one.

This does not change the backend storage or source of truth.  For the firewall creation form (this PR), locations are still taken from `Option.locations` and private subnets still are still taken from `dataset_authorize(@project.private_subnets_dataset, "PrivateSubnet:view")`, just as they were before.

> * Even though we are not using at the moment, my idea with option tree is that it can be used to eliminate majority of validation code. It simply generates all valid options. If we keep dependency relationships on front end code, we would lose that ability.

For cases where generating all valid options is complex in and of itself, I see no problem keeping the option tree generator (or something simpler derived from it) for that, and using it as the source of truth, while still using the approach in this PR for front-end display.

enescakir:
If we go with this approach, I expect to see at least a form for creating a VM/PostgreSQL. The firewall rule creation form is relatively simple. Most of the complexity in content generation comes from the VM/PG creation forms.

jeremyevans:
I've pushed a commit to convert the PostgreSQL creation form to using forme.  See the commit message for details.  This passes specs, but is not fully covered.

Overall, I'm not convinced it reduces the complexity significantly.  I think slightly, but not much, and other developers may consider the complexity increased instead of decreased.  I think that the nature of the problem is complex.

Maybe a good test of the two approaches is to try to add new options for standard-32/standard-64 for the AWS private locations.  @furkansahin  mentioned that this was difficult with the current approach.  I'm not sure it would be any easier when using forme, but it seems like a decent test.

One possible alternative approach, which Furkan mentioned, is to store all valid configurations, likely in the database (could be in YAML files, but the database would likely be better).

Each entry would have

* Location ID
* Family
* Server Size
* Storage Size
* HA Type

This basically does the dependency calculations up front, and makes it easy to add new configurations by adding a database row instead of modifying the logic.  So it would make the OptionTreeGenerator code simpler.  However, you'd still need some logic, as private locations would affect things, and if some options were-flavor specific, you'd need to handle that.  Overall, as complex as the OptionTreeGenerator is, I'm not sure replacing it with a more-fixed list would significantly reduce complexity.

furkansahin:
Hey, I checked the commit now, thanks @jeremyevans . You're right, I also don't feel like this reduces complexity. We might try adding the new instances sizes but I don't think the root of the problem is fixed which, in my opinion, Option.rb and dynamic parent/children relationship calculation for different configurations. 

I see that @enescakir has created this PR, previously https://github.com/ubicloud/ubicloud/pull/2698. To me, having these configurations openly written like that reduces the brain power needed when a new config comes in. The focus then shifts to properly copy pasting and being careful. However, that's already the case for the current code base. I feel like that's our only option, to be honest. 

As long as we have dynamic filtering + selections that depend on multiple other options, this code base will not be welcoming to work in. 

However, the biggest drawback of yaml files is that, we cannot update the location information on the fly which makes ubi-on-aws impossible. Therefore, that information needs to be moved to the database instead of keeping as a yaml file.

byucesoy:
I agree that the complexity comes from the nature of the problem and I also feel similarly with @furkansahin on this approach not reducing the complexity.

I'd be open about explicitly writing down each allowed configuration instead of generating them on the fly. However I have 3 concerns:
- Numbers of configs increase exponentially. For PostgreSQL we would end up more than 100 items. When we need to add/change something (for example when open new region), I wouldn't want to copy 100 items and manually editing them. It is quite error prone. We have a similar explicit list for billing meters and we already had typos while editing it couple of time in the past.
- We would still need to perform filtering based on project at some place, so we wouldn't completely free ourselves from the dynamic nature of the problem. We might even have to keep parent/child relationship (or maybe we wouldn't need, not sure)
- On YAML vs Database. I agree that we need to keep this in database (to enable ubi-on-aws), but seeing & editing database rows are even more difficult compared to editing YAML file. So there is extra layer of difficulty here as well.

PR #3111, opened 2025-04-04
Enforce project location access in routes

jeremyevans:
In the routes, when a location is given, check that is either available for all projects, or that it is tied to the current project.

This adds:

* Location.for_project: Filters to only locations not tied to projects, and those tied to the given project.

* Location:visible_or_for_project: Filters to only visible locations not tied to projects, and those tied to the given project.

* Location#visible_or_for_project?: Checks whether the location is either visible and not tied to a project, or is tied to the given project.


furkansahin (approved):
Thanks for catching this.

PR #3114, opened 2025-04-05
Fixes and improvements on pgbouncer setup

byucesoy:
**Show port 6432 along with 5432 in the PostgreSQL firewall rules**

**Configure pgbouncer in template1 database**
This ensures that new databases will automatically have pgbouncer configuration
such as auth_query and pgbouncer schema. Otherwise, without auth_query, it is
not possible to connect to the new database using pgbouncer.

**Don't check existence of pgbouncer role before creating it**
The check is added for the sake of idempotency, but we already delete entire
database cluster few lines above anyway, so it is not needed.

**Combine database initialization commands**
I'm moving replication and monitoring role creations to the same code block
that configures the pgbouncer. I'm also making small improcements such as
removing the unnecessary transaction block and bringing the GRANT/REVOKE
statements to the same place.

**Use ON_ERROR_STOP=1 with psql**
By default, psql always returns 0, even if the command fails. This prevents us
from detecting errors and acting on it (i.e. retrying usually). This commit
adds ON_ERROR_STOP=1 to the psql command, which will cause it to stop
execution and return a non-zero exit code if any command fails.

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -212,7 +212,7 @@ class PostgresServer < Sequel::Model
   end
 
   def self.run_query(vm, query)
-    vm.sshable.cmd("PGOPTIONS='-c statement_timeout=60s' psql -U postgres -t --csv", stdin: query).chomp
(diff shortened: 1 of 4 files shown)


shikharbhardwaj (approved):
Thanks for fixing this!

PR #3116, opened 2025-04-07
ConvergePostgresResource deadline change for maintenance windows

furkansahin:
With the introduction of maintenance windows, the refresh of a PG instance may simply wait for a while until the maintenance window comes. At the start of the ConvergePostgresResource, we are setting the deadline to 2 hours. This simply triggers a page if the preparation of a server was quick but still we are waiting because we are not yet at the maintenance window. With this commit, we are fixing that.

### DIFF
--- a/prog/postgres/converge_postgres_resource.rb
+++ b/prog/postgres/converge_postgres_resource.rb
@@ -6,7 +6,7 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
   subject_is :postgres_resource
 
   label def start
-    register_deadline(nil, 2 * 60 * 60)
+    register_deadline("recycle_representative_server", 2 * 60 * 60)
     hop_provision_servers
   end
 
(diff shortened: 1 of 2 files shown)


prog/postgres/converge_postgres_resource.rb
```diff
@@ -24,7 +24,10 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
 
   label def wait_servers_to_be_ready
     hop_provision_servers unless postgres_resource.has_enough_fresh_servers?
-    hop_recycle_representative_server if postgres_resource.has_enough_ready_servers?
+    if postgres_resource.has_enough_ready_servers?
+      register_deadline(nil, (24 - PostgresResource::MAINTENANCE_DURATION_IN_HOURS) * 60 * 60)
```
byucesoy, line 28: 22 hours of deadline is pretty long. I think instead, we can set a short deadline just after the line we nap for 10 minutes (in next label)

furkansahin: Nope, that's it. I was kind of worried that we wouldn't be notified if somehow we messed and some servers are infinitely waiting but that's the case for many other operations in the codebase as well. I have updated it accordingly

byucesoy (approved):

PR #3122, opened 2025-04-08
Put PostgreSQL standbys to different data centers

byucesoy:
We don't have different availability zones in none of our bare metal providers, but at least, in Hetzner, we have different data centers. This commit ensures that we put PostgreSQL standbys in different data centers if they are using Hetzner as underlying provider.

For LeaseWeb, we don't even have different data centers, but we can place the servers to different racks. Currently all our hosts are already in different racks, so we don't need to do anything special for them.

### DIFF
--- a/prog/postgres/converge_postgres_resource.rb
+++ b/prog/postgres/converge_postgres_resource.rb
@@ -14,8 +14,11 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
     hop_wait_servers_to_be_ready if postgres_resource.has_enough_fresh_servers?
 
     if postgres_resource.servers.all? { _1.vm.vm_host } || postgres_resource.location.provider == "aws"
-      # vm_host might be nil for AWS VM based PG instances
-      exclude_host_ids = (Config.development? || Config.is_e2e) ? [] : postgres_resource.servers.map { _1.vm.vm_host&.id }.compact
+      exclude_host_ids = []
+      if !(Config.development? || Config.is_e2e) && postgres_resource.location.provider == HostProvider::HETZNER_PROVIDER_NAME
+        used_data_centers = postgres_resource.servers.map { _1.vm.vm_host.data_center }.uniq
+        exclude_host_ids = VmHost.where(data_center: used_data_centers).map(&:id)
+      end
       Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids)
     end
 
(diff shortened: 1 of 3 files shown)


furkansahin (approved):

prog/postgres/converge_postgres_resource.rb
```diff
@@ -14,8 +14,11 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
     hop_wait_servers_to_be_ready if postgres_resource.has_enough_fresh_servers?
 
     if postgres_resource.servers.all? { _1.vm.vm_host } || postgres_resource.location.provider == "aws"
-      # vm_host might be nil for AWS VM based PG instances
-      exclude_host_ids = (Config.development? || Config.is_e2e) ? [] : postgres_resource.servers.map { _1.vm.vm_host&.id }.compact
+      exclude_host_ids = []
+      if !(Config.development? || Config.is_e2e) && postgres_resource.location.provider == HostProvider::HETZNER_PROVIDER_NAME
+        used_data_centers = postgres_resource.servers.map { _1.vm.vm_host.data_center }.uniq
+        exclude_host_ids = VmHost.where(data_center: used_data_centers).map { _1.id }
```
enescakir, line 20: nit
```suggestion
        exclude_host_ids = VmHost.where(data_center: used_data_centers).map(&:id)
```

enescakir, line 20: Do we have data_center information for US hosts?

byucesoy: No we don't, but this code only runs for Hetzner.

In US, there is only one data center for Leaseweb, our isolation possibility there is using different racks, which we do already by placing each of our host to different rack (so we don't need to do anything for PG). I added some explanation to the commit message.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -14,8 +14,11 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
     hop_wait_servers_to_be_ready if postgres_resource.has_enough_fresh_servers?
 
     if postgres_resource.servers.all? { _1.vm.vm_host } || postgres_resource.location.provider == "aws"
-      # vm_host might be nil for AWS VM based PG instances
-      exclude_host_ids = (Config.development? || Config.is_e2e) ? [] : postgres_resource.servers.map { _1.vm.vm_host&.id }.compact
+      exclude_host_ids = []
+      if !(Config.development? || Config.is_e2e) && postgres_resource.location.provider == HostProvider::HETZNER_PROVIDER_NAME
+        used_data_centers = postgres_resource.servers.map { _1.vm.vm_host.data_center }.uniq
```
shikharbhardwaj, line 19: jw: does the allocator do some balancing among different data centres? Or do we have enough spare capacity in 3 different DCs always to not worry about this?

byucesoy: Allocator does not do any balancing, but I think we are big enough (in Hetzner) to not worry about this. This returns many hosts from many different data centers. (I'm not sharing the output as it is not a public information)

` VmHost.all.map{ _1.data_center }.tally`

enescakir (approved):

PR #3125, opened 2025-04-08
Switch period position in multi-line expressions to "trailing"

fdr:
This is particularly dire in `pry` where the idiomatic/default `.` starts a runs a command, e.g. `.ls` runs the `ls` command.  Using trailing dots solves this problem, it's also the common code style in Jeremy Evans software libraries.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -95,3 +95,6 @@ Capybara/ClickLinkOrButtonStyle:
 
 Lint/UselessTimes:
   Enabled: false
+
+Layout/DotPosition:
+  EnforcedStyle: trailing
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ end
 
 group :lint do
   gem "brakeman"
-  gem "erb-formatter", github: "ubicloud/erb-formatter", ref: "886cb4e27b099d6f347b950375aa22e638888f95"
+  gem "erb-formatter", github: "ubicloud/erb-formatter", ref: "df3174476986706828f7baf3e5e6f5ec8ecd849b"
 end
 
 group :test do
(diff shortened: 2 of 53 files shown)


enescakir:
I think you should send an email to eng@ after merging this. Since most of the members used a leading period, they need to change how they write the code.

fdr:
> I think you should send an email to eng@ after merging this. Since most of the members used a leading period, they need to change how they write the code.

sure, at least to not be confused as rubocop rewrites it for them

fdr:
@jeremyevans @enescakir I was playing with this, and although I want to keep getting rid of redundant code formatting in erb-formatter, I think pry added some kind of multi-line paste feature (*or* it was mental bitrot) that prevents Bad Stuff from happening in pasting cases like this:

```ruby
  foo = 'a b'
    . split
```

Where `foo` is `["a", "b"]` as expected.

I'm thinking about leaving the leading `.` because it does close up the problem of slicing off a subexpression and running against `self` rather than some other expression:

```ruby
not_self.
  foo.
  bar
```

Where slicing off the last two lines can result in effectively self.foo.bar

In the case where you accidentally copy/paste the following sub-sequence:

```ruby
.foo
.bar
```

The result is still bad: it tries to do a system command on `foo`. But somehow I think the danger factor flips. What do you think?

fdr:
deciding against this for now: #3133

jeremyevans:
FWIW, the `.foo` running the `foo` command is pry thing, irb doesn't do that (you get a syntax error).

fdr:
> FWIW, the `.foo` running the `foo` command is pry thing, irb doesn't do that (you get a syntax error).

oh yeah, totally aware of that...this is why I mentioned the couple of features that kept me on pry, because my reasons for keeping it are very micro.

PR #3132, opened 2025-04-09
Create multiple pgbouncer instances

shikharbhardwaj:
Create multiple pgbouncer instances

Configure a systemd template to create multiple pgbouncer instances, to
avoid bottlenecks created due to it being single-threaded.

To have pgbouncer connect to PG via a domain socket, we need to specify
a unix_socket_dir in the pgbouncer config. But this config value cannot
be shared between different pgbouncer instances, thus we use systemd
socket activation to have pgbouncer use the same server connection
socket but separate listening sockets. [1]

This change also adds a [peers] section to the pgbouncer config, to
allow processing cancellation requests correctly. Since this needs a
unique `peer_id` to be set for each process, we create separate config
files for each process here. [2]

[1]: https://github.com/pgbouncer/pgbouncer/blob/a980f572d537f43216d4ac2ea4550cbbfa4c31b7/src/pooler.c#L510
[2]: https://www.pgbouncer.org/config.html#section-peers

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -101,7 +101,8 @@ class PostgresServer < Sequel::Model
         }
       },
       identity: resource.identity,
-      hosts: "#{resource.representative_server.vm.private_ipv4} #{resource.identity}"
+      hosts: "#{resource.representative_server.vm.private_ipv4} #{resource.identity}",
+      pgbouncer_instances: (vm.vcpus / 2.0).ceil.clamp(1, 8)
     }
   end
 
--- a/rhizome/postgres/lib/pgbouncer_setup.rb
+++ b/rhizome/postgres/lib/pgbouncer_setup.rb
@@ -1,0 +1,132 @@
+# frozen_string_literal: true
+
+class PgBouncerSetup
+  def initialize(version, max_connections, num_instances)
+    @version = version
+    @max_connections = max_connections
+    @num_instances = num_instances
+  end
+
+  def service_template_name
+    "pgbouncer@"
+  end
+
+  def pgbouncer_service_file_path
+    "/etc/systemd/system/#{service_template_name}.service"
+  end
+
+  def socket_service_file_path
+    "/etc/systemd/system/#{service_template_name}.socket"
+  end
+
+  def create_service_templates
+    File.write(pgbouncer_service_file_path, <<PGBOUNCER_SERVICE
+[Unit]
+Description="connection pooler for PostgreSQL (%i)"
+After=network.target
+Requires=pgbouncer@%i.socket
+
+[Service]
+Type=notify
+User=postgres
+ExecStart=/usr/sbin/pgbouncer /etc/pgbouncer/pgbouncer_%i.ini
+ExecReload=/bin/kill -HUP $MAINPID
+KillSignal=SIGINT
+
+[Install]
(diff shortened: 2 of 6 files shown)


rhizome/postgres/lib/pgbouncer_setup.rb
```diff
@@ -0,0 +1,132 @@
...
+client_tls_sslmode = require
+client_tls_protocols = tlsv1.3
+client_tls_ca_file = /etc/ssl/certs/ca.crt
+client_tls_cert_file = /etc/ssl/certs/server.crt
+client_tls_key_file = /etc/ssl/certs/server.key
+
+user = postgres
+
+max_client_conn = 5000
+max_db_connections = #{@max_connections}
```
byucesoy, line 107: Should we divide this and above to num_instances or do these numbers shared across pgbouncer processes already?

model/postgres/postgres_server.rb
```diff
@@ -101,7 +101,8 @@ def configure_hash
         }
       },
       identity: resource.identity,
-      hosts: "#{resource.representative_server.vm.private_ipv4} #{resource.identity}"
+      hosts: "#{resource.representative_server.vm.private_ipv4} #{resource.identity}",
+      pgbouncer_instances: vm.vcpus / 2
```
byucesoy, line 105: Maybe we should set a fixed maximum number. 30 pgbouncers for standard-60 seems excessive. I'd go with

```
(vm.vcpus / 4).ceil.clamp(1, 8)
```

BTW, I don't have any scientific thought behind this formula. It just feels alright. If you have a better idea, that would work as well.

byucesoy (approved):

model/postgres/postgres_server.rb
shikharbhardwaj: Yeah it makes sense not to create too many of them for the larger instances. I was mostly going by our tests for `standard-8`, where we created 4 instances which was close enough to the original numbers, thinking to keep the factor as 2.

rhizome/postgres/lib/pgbouncer_setup.rb
```diff
@@ -0,0 +1,132 @@
...
+
+client_tls_sslmode = require
+client_tls_protocols = tlsv1.3
+client_tls_ca_file = /etc/ssl/certs/ca.crt
+client_tls_cert_file = /etc/ssl/certs/server.crt
+client_tls_key_file = /etc/ssl/certs/server.key
+
+user = postgres
+
+max_client_conn = 5000
```
byucesoy, line 106: Do we need to divide this as well?

model/postgres/postgres_server.rb
byucesoy: Yeah, but our test was designed specifically to hit pgbouncer's weak spot and make it work as much as possible. I'd expect real world workloads to be less severe. That said I don't have strong arguments for 4 as a factor anyway and more importantly this is not difficult to change if we turned out to be wrong. So keeping the factor as 2 is also fine.

Let's not use more than 8 pgbouncer processes though.

shikharbhardwaj:
Since deploying this PR needs some changes to existing DBs to setup the prerequisites (role and the SECURITY DEFINER function), we've decided to notify customers about these changes. Will merge and deploy this change on Thursday, 9-10 AM UTC.

PR #3137, opened 2025-04-10
Postgres Read Replicas

furkansahin:
## Postgres backend support to provision read replicas
We start supporting read replica provisioning at the backend. Basically,
we need to pass a parent_id and nil restore target to
PostgresResourceNexus.assemble. It handles the pg configuration and
setting up the replication.

## Read replica lag detection and server recycling
Even if we setup replication, it is possible that the replication would
fail and we might need to recycle the server. To detect that, we are
continuously checking the lag between a read replica and its parent. If
the applied wal difference is greater than 80 MB and the replica does
not apply new wal files in the upcoming 15 minutes, we decide to recycle
the server.
We also implement the read replica failover mechanism in this commit
which is essentially refreshing dns records marking the new server as
the representative.

## Postgres read replicas web and API support
With this commit, read replicas are available in the UI. We can
provision a read replica basically by entering a name and clicking to
the create button. We are not allowing customers to pick different spec
for read replicas because it opens up a bunch of options for scale
operations.

----

> [!IMPORTANT]
> Add support for PostgreSQL read replicas, including provisioning, lag detection, failover, and UI/API integration.
> 
>   - **Behavior**:
>     - Add support for PostgreSQL read replicas in `postgres_resource.rb` and `postgres_server.rb`.
>     - Implement read replica lag detection and recycling in `postgres_server.rb` and `postgres_server_nexus.rb`.
>     - Add read replica failover mechanism in `postgres_server.rb`.
>   - **API**:
>     - Add `createPostgresDatabaseReadReplica` endpoint in `openapi.yml`.
>     - Update `routes/project/location/postgres.rb` to handle read replica creation.
>   - **UI**:
>     - Update `views/postgres/index.erb` and `views/postgres/show.erb` to display read replicas.
>   - **Tests**:
>     - Add tests for read replica functionality in `postgres_server_spec.rb` and `postgres_server_nexus_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for e29fb8a2afbac74662ece5cbef1ec8feda72ec8c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/helpers/postgres.rb
+++ b/helpers/postgres.rb
@@ -60,7 +60,11 @@ class Clover
       }
     else
       dataset = dataset.eager(:representative_server, :timeline)
-      @postgres_databases = Serializers::Postgres.serialize(dataset.all, {include_path: true})
+      resources = dataset.all
+        .group_by { |r| r.read_replica? ? r[:parent_id] : r[:id] }
+        .flat_map { |group_id, rs| rs.sort_by { |r| r[:created_at] } }
+
+      @postgres_databases = Serializers::Postgres.serialize(resources, {include_path: true})
       view "postgres/index"
     end
   end
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -14,6 +14,7 @@ class PostgresResource < Sequel::Model
   one_to_many :metric_destinations, class: :PostgresMetricDestination, key: :postgres_resource_id
   many_to_one :private_subnet
   many_to_one :location, key: :location_id, class: :Location
+  one_to_many :read_replicas, class: :PostgresResource, key: :parent_id, conditions: {restore_target: nil}
 
   plugin :association_dependencies, firewall_rules: :destroy, metric_destinations: :destroy
   dataset_module Pagination
@@ -121,6 +122,10 @@ class PostgresResource < Sequel::Model
   def validate
     super
     validates_includes(0..23, :maintenance_window_start_at, allow_nil: true, message: "must be between 0 and 23")
+  end
+
+  def read_replica?
+    parent_id && restore_target.nil?
   end
 
   module HaType
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -17,7 +17,7 @@ class PostgresServer < Sequel::Model
   include HealthMonitorMethods
 
   semaphore :initial_provisioning, :refresh_certificates, :update_superuser_password, :checkup
-  semaphore :restart, :configure, :take_over, :configure_prometheus, :destroy
+  semaphore :restart, :configure, :take_over, :configure_prometheus, :destroy, :recycle
 
   def configure_hash
     configs = {
@@ -107,7 +107,7 @@ class PostgresServer < Sequel::Model
   end
 
   def trigger_failover
-    if primary? && (standby = failover_target)
+    if representative_at && (standby = failover_target)
       standby.incr_take_over
       true
     else
@@ -137,7 +137,7 @@ class PostgresServer < Sequel::Model
   end
 
   def needs_recycling?
-    vm.display_size != resource.target_vm_size || storage_size_gib != resource.target_storage_size_gib
+    recycle_set? || vm.display_size != resource.target_vm_size || storage_size_gib != resource.target_storage_size_gib
   end
 
   def lsn_caught_up
@@ -155,8 +155,9 @@ class PostgresServer < Sequel::Model
 
   def failover_target
     target = resource.servers
-      .select { _1.standby? && _1.strand.label == "wait" && !_1.needs_recycling? }
-      .map { {server: _1, lsn: _1.run_query("SELECT pg_last_wal_receive_lsn()").chomp} }
+      .reject { _1.representative_at }
+      .select { _1.strand.label == "wait" && !_1.needs_recycling? }
+      .map { {server: _1, lsn: _1.current_lsn} }
       .max_by { lsn2int(_1[:lsn]) }
 
     return nil if target.nil?
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -128,12 +128,29 @@ class PostgresServer < Sequel::Model
     !resource.representative_server.primary?
   end
 
+  def read_replica?
+    resource.read_replica?
+  end
+
   def storage_size_gib
     vm.vm_storage_volumes_dataset.first(boot: false)&.size_gib
   end
 
   def needs_recycling?
     vm.display_size != resource.target_vm_size || storage_size_gib != resource.target_storage_size_gib
+  end
+
+  def lsn_caught_up
+    parent_server = if read_replica?
+      resource.parent.representative_server
+    else
+      resource.representative_server
+    end
+    lsn_diff(parent_server.current_lsn, current_lsn) < 80 * 1024 * 1024
+  end
+
+  def current_lsn
+    run_query("SELECT #{lsn_function}").chomp
   end
 
   def failover_target
@@ -152,6 +169,16 @@ class PostgresServer < Sequel::Model
     target[:server]
   end
 
+  def lsn_function
+    if primary?
+      "pg_current_wal_lsn()"
+    elsif standby?
+      "pg_last_wal_receive_lsn()"
+    else
+      "pg_last_wal_replay_lsn()"
+    end
+  end
+
   def init_health_monitor_session
     FileUtils.rm_rf(health_monitor_socket_path)
     FileUtils.mkdir_p(health_monitor_socket_path)
@@ -167,7 +194,6 @@ class PostgresServer < Sequel::Model
   def check_pulse(session:, previous_pulse:)
     reading = begin
       session[:db_connection] ||= Sequel.connect(adapter: "postgres", host: health_monitor_socket_path, user: "postgres", connect_timeout: 4)
-      lsn_function = primary? ? "pg_current_wal_lsn()" : "pg_last_wal_receive_lsn()"
       last_known_lsn = session[:db_connection]["SELECT #{lsn_function} AS lsn"].first[:lsn]
       "up"
     rescue
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -897,6 +897,33 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Metric Destination
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/read-replica':
+    post:
+      operationId: createPostgresDatabaseReadReplica
+      summary: Create a read replica of the Postgres database
+      parameters:
+        - $ref: '#/components/parameters/project_id'
+        - $ref: '#/components/parameters/location'
+        - $ref: '#/components/parameters/postgres_database_reference'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                name:
+                  type: string
+              additionalProperties: false
+              required:
+                - name
+      responses:
+        '200':
+          $ref: '#/components/responses/PostgresDatabase'
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Postgres Database
   '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/reset-superuser-password':
     post:
       operationId: resetSuperuserPassword
@@ -1722,6 +1749,13 @@ components:
           description: Maintenance window start time
           type: integer
           nullable: true
+        read_replica:
+          description: If the database is a read replica or not
+          type: boolean
+        parent:
+          description: Parent Postgres database
+          type: string
+          nullable: true
       additionalProperties: false
       required:
         - flavor
@@ -1734,6 +1768,7 @@ components:
         - version
         - vm_size
         - maintenance_window_start_at
+        - read_replica
     PostgresFirewallRule:
       type: object
       properties:
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -144,7 +144,7 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
     when "Succeeded"
       hop_refresh_certificates
     when "Failed", "NotStarted"
-      backup_label = if postgres_server.standby?
+      backup_label = if postgres_server.standby? || postgres_server.read_replica?
         "LATEST"
       else
         postgres_server.timeline.latest_backup_label_before_target(target: postgres_server.resource.restore_target)
@@ -243,7 +243,7 @@ CONFIG
 
       when_initial_provisioning_set? do
         hop_update_superuser_password if postgres_server.primary?
-        hop_wait_catch_up if postgres_server.standby?
+        hop_wait_catch_up if postgres_server.standby? || postgres_server.read_replica?
         hop_wait_recovery_completion
       end
 
@@ -300,10 +300,9 @@ SQL
   end
 
   label def wait_catch_up
-    query = "SELECT pg_current_wal_lsn() - replay_lsn FROM pg_stat_replication WHERE application_name = '#{postgres_server.ubid}'"
-    lag = postgres_server.resource.representative_server.run_query(query).chomp
+    nap 30 unless postgres_server.lsn_caught_up
 
-    nap 30 if lag.empty? || lag.to_i > 80 * 1024 * 1024 # 80 MB or ~5 WAL files
+    hop_wait if postgres_server.read_replica?
 
     postgres_server.update(synchronization_status: "ready")
     postgres_server.resource.representative_server.incr_configure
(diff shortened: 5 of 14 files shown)


model/postgres/postgres_resource.rb
```diff
@@ -123,6 +124,10 @@ def validate
     validates_includes(0..23, :maintenance_window_start_at, allow_nil: true, message: "must be between 0 and 23")
   end
 
+  def read_replica?
+    !parent_id.nil? && restore_target.nil?
```
byucesoy, line 128: You can just use `parent_id` instead of `!parent_id.nil?`

model/postgres/postgres_resource.rb
```diff
@@ -14,6 +14,7 @@ class PostgresResource < Sequel::Model
   one_to_many :metric_destinations, class: :PostgresMetricDestination, key: :postgres_resource_id
   many_to_one :private_subnet
   many_to_one :location, key: :location_id, class: :Location
+  one_to_many :read_replicas, class: :PostgresResource, key: :parent_id do |ds| ds.where(restore_target: nil) end
```
enescakir, line 17: nit:
```suggestion
  one_to_many :read_replicas, class: :PostgresResource, key: :parent_id, conditions: {restore_target: nil}
```

model/postgres/postgres_server.rb
```diff
@@ -83,10 +83,16 @@ def configure_hash
...
+      if doing_pitr? && !read_replica?
         configs[:recovery_target_time] = "'#{resource.restore_target}'"
       end
 
+      if read_replica?
+        configs[:recovery_target_time] = "''"
+        configs[:recovery_target_timeline] = "latest"
+        configs[:recovery_target_inclusive] = false
+      end
+
```
enescakir, line 95: nit: It might be personal, but it feels easier to read for me
```suggestion
      if read_replica?
        configs[:recovery_target_time] = "''"
        configs[:recovery_target_timeline] = "latest"
        configs[:recovery_target_inclusive] = false
      elsif doing_pitr?
        configs[:recovery_target_time] = "'#{resource.restore_target}'"
      end
```

model/postgres/postgres_server.rb
```diff
@@ -106,6 +112,20 @@ def configure_hash
...
+        update(representative_at: nil)
+        target.update(representative_at: Time.now)
+        resource.incr_refresh_dns_record
+        return true
+      end
+
+      if read_replica?
+        Clog.emit("Failed to trigger read replica failover")
+        return false
+      end
```
enescakir, line 126: nit: a single branch for read replicas might be easier to read
```suggestion
      if read_replica?
        if resource.representative_server.id == id && (target = read_replica_failover_target)
          update(representative_at: nil)
          target.update(representative_at: Time.now)
          resource.incr_refresh_dns_record
          return true
        end
        Clog.emit("Failed to trigger read replica failover")
        return false
      end
```

prog/postgres/postgres_server_nexus.rb
```diff
@@ -204,7 +204,7 @@ def before_run
   scrape_interval: 10s
   external_labels:
     ubicloud_resource_id: #{postgres_server.resource.ubid}
-    ubicloud_resource_role: #{(postgres_server.id == postgres_server.resource.representative_server.id) ? "primary" : "standby"}
+    ubicloud_resource_role: #{(!postgres_server.read_replica? && (postgres_server.id == postgres_server.resource.representative_server.id)) ? "primary" : "standby"}
```
byucesoy, line 207: In this context, shouldn't we send the role as `primary` if the server is representative_server of the resource?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -384,6 +384,20 @@ def before_run
...
 
+    if postgres_server.read_replica?
+      already_lagging = postgres_server.timeline_id_is_lagging_set?
+      currently_lagging = postgres_server.timeline_id_is_lagging
+
+      if already_lagging && currently_lagging
+        postgres_server.incr_recycle
+      elsif currently_lagging
+        postgres_server.incr_timeline_id_is_lagging
+        nap 2 * 60
```
byucesoy, line 395: 2 minutes seems too short. Maybe set it to 15 minutes?

Also, @fdr, what do you think this way of detecting broken replication, which can happen if read replica fetches the update before standby and we perform failover? (Not sure if it can happen in any other way). I believe it is OK, but wanted to get your opinion as well.

model/postgres/postgres_server.rb
```diff
@@ -140,7 +154,16 @@ def storage_size_gib
   end
 
   def needs_recycling?
-    vm.display_size != resource.target_vm_size || storage_size_gib != resource.target_storage_size_gib
+    recycle_set? || vm.display_size != resource.target_vm_size || storage_size_gib != resource.target_storage_size_gib
+  end
+
+  def timeline_id_is_lagging
+    fail "the server is not a read replica" unless read_replica?
```
byucesoy, line 161: I don't think we need this check.

model/postgres/postgres_server.rb
```diff
@@ -159,6 +182,17 @@ def failover_target
     target[:server]
   end
 
+  def read_replica_failover_target
+    target = resource.servers
```
byucesoy, line 186: I'd add an `if/else` branch and embed this into `failover_target` method. My reasoning is that callers wouldn't need to pick one or the other function, `failover_target` would abstract away the complexity. Actually, we might not need branching.

byucesoy:
```ruby
  def failover_target
    target = resource.servers
      .reject { _1.representative_at }
      .select { _1.strand.label == "wait" && !_1.needs_recycling? }
      .map { {server: _1, lsn: _1.run_query("SELECT pg_last_wal_receive_lsn()").chomp} }
      .max_by { lsn2int(_1[:lsn]) }

    return nil if target.nil?

    if resource.ha_type == PostgresResource::HaType::ASYNC
      return nil if lsn_monitor.last_known_lsn.nil?
      return nil if lsn_diff(lsn_monitor.last_known_lsn, target[:lsn]) > 80 * 1024 * 1024 # 80 MB or ~5 WAL files
    end
    target[:server]
  end
```

This seems to cover all cases, but please double check.

model/postgres/postgres_server.rb
```diff
@@ -110,6 +110,20 @@ def configure_hash
   end
 
   def trigger_failover
+    if read_replica?
```
byucesoy, line 113: Can we keep the logic more or less same for both regular cluster and read replicas somehow? For example;

```ruby
  def trigger_failover
    if representative_at && (standby = failover_target)
      standby.incr_take_over
      true
    else
      Clog.emit("Failed to trigger failover")
      false
    end
  end
```

You might need to slightly modify `taking_over` label.

helpers/postgres.rb
```diff
@@ -60,7 +60,25 @@ def postgres_list
...
+      grouped_resources = resources.group_by { |r| r[:parent_id] }
+
+      # Get all parents (resources with parent_id = nil)
+      parents = grouped_resources[nil] || []
+
+      # Create the ordered result
+      ordered_resources = []
+
+      # For each parent, add it followed by its children
+      parents.each do |parent|
```
byucesoy, line 74: Maybe sort parents by their created_at as well? Also same for children, but in their own group?

openapi/openapi.yml
```diff
@@ -899,6 +899,33 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Metric Destination
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_name}/read-replica':
+    post:
+      operationId: readReplicaPostgresDatabase
```
byucesoy, line 904: createPostgresDatabaseReadReplica?

openapi/openapi.yml
```diff
@@ -899,6 +899,33 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Metric Destination
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_name}/read-replica':
```
byucesoy, line 902: We replaces `postgres_database_name` with `postgres_database_reference` in this [PR](https://github.com/ubicloud/ubicloud/pull/3150)

openapi/openapi.yml
```diff
@@ -1750,6 +1777,14 @@ components:
           description: Maintenance window start time
           type: integer
           nullable: true
+        read_replica:
+          description: Read replica of the Postgres database
```
byucesoy, line 1781:
```
is_read_replica:
  description: Whether this is a read replica of the Postgres database
```

Not sure about `is_` prefix? We usually don't use `is_` prefix, but `read_replica` and `read_replicas` fields together seems confusing. @fdr, @geemus any opinions?

routes/project/location/postgres.rb
```diff
@@ -184,6 +187,34 @@ class Clover
         end
       end
 
+      r.on "read-replica" do
+        r.post true do
+          # Create the read-replica
```
byucesoy, line 192: No need for this comment.

routes/project/location/postgres.rb
```diff
@@ -184,6 +187,34 @@ class Clover
...
+            project_id: @project.id,
+            location_id: pg.location_id,
+            name: request_body_params["name"],
+            target_vm_size: pg.target_vm_size,
+            target_storage_size_gib: pg.target_storage_size_gib,
+            ha_type: PostgresResource::HaType::NONE,
+            version: pg.version,
+            flavor: pg.flavor,
+            parent_id: pg.id,
+            restore_target: nil # TODO: remove this when we have a proper restore target
```
byucesoy, line 205: We don't need this `TODO` anymore, right?

serializers/postgres.rb
```diff
@@ -13,7 +13,8 @@ def self.serialize_internal(pg, options = {})
       ha_type: pg.ha_type,
       flavor: pg.flavor,
       ca_certificates: pg.ca_certificates,
-      maintenance_window_start_at: pg.maintenance_window_start_at
+      maintenance_window_start_at: pg.maintenance_window_start_at,
+      read_replica: pg.read_replica? || false
```
byucesoy, line 17: Do we need `|| false` part?

spec/prog/postgres/converge_postgres_resource_spec.rb
```diff
@@ -20,6 +20,7 @@
 
   before do
     allow(nx).to receive(:postgres_resource).and_return(postgres_resource)
+    allow(postgres_resource).to receive(:read_replica?).and_return(false)
```
byucesoy, line 23: We don't make change on converge prog. Do we need this change on the test?

views/postgres/index.erb
```diff
@@ -35,12 +35,16 @@
...
       rows: @postgres_databases.map do |pg|
         brand, = flavors[pg[:flavor]]
         [
           [
             [pg[:name], {link: @project_data[:path] + pg[:path]}],
+            [pg[:vm_size]],
+            [pg[:storage_size_gib]],
+            [pg[:version]],
+            [pg[:read_replica] ? "Read Replica" : "Primary"],
             ["<div class='inline-flex gap-x-1 items-center'><img src='/icons/pg-#{brand.downcase}.png' class='object-cover rounded-lg overflow-hidden w-6 h-6'/><p>#{brand}</p></div>", {escape: false}],
```
byucesoy, line 48: How does it look if we indent read replicas few pixels? If it looks good, we might remove read replica column.

views/postgres/show.erb
```diff
@@ -108,100 +108,171 @@ delete_perm = has_permission?("Postgres:delete", @pg[:id]) %>
...
-              <div class="col-span-12 sm:col-span-2 flex justify-end items-end">
-                <%== part("components/form/submit_button", text: "Fork") %>
+              <div class="grid grid-cols-12 gap-6">
+                <div class="col-span-12 sm:col-span-5">
+                  <%== part(
+                    "components/form/text",
+                    label: "New server name",
+                    name: "name",
+                    attributes: {
+                      placeholder: @pg[:name] + "-restored",
```
byucesoy, line 193: -fork?

views/postgres/show.erb
```diff
@@ -108,100 +108,171 @@ delete_perm = has_permission?("Postgres:delete", @pg[:id]) %>
...
+    <div>
+      <div class="md:flex md:items-center md:justify-between pb-2 lg:pb-4">
+        <div class="min-w-0 flex-1">
+          <h3 class="text-2xl font-bold leading-7 text-gray-900 sm:truncate sm:text-2xl sm:tracking-tight">
+            Read Replicas
+          </h3>
+        </div>
+      </div>
+      <div class="overflow-hidden rounded-lg shadow ring-1 ring-black ring-opacity-5 bg-white divide-y divide-gray-200">
+        <!-- Read Replicas -->
```
byucesoy, line 227: We already have this HTML comment few lines above.

byucesoy (commented):
Thanks for the PR. I think we can merge this soon. Number of comments might seem high but most of them are quite minor.

serializers/postgres.rb
enescakir: If we want to convert `nil` to `false`, `!!pg.read_replica?` does it

model/postgres/postgres_server.rb
furkansahin: Yeap, I have updated it as requested

model/postgres/postgres_server.rb
furkansahin: yeap!

helpers/postgres.rb
furkansahin: I believe the new code handles that

openapi/openapi.yml
furkansahin: considering I have removed read_replicas, this should be OK?

views/postgres/index.erb
furkansahin: I have solved it by changing the column to "parent" and then ordering the replicas under the parent

helpers/postgres.rb
```diff
@@ -60,7 +60,11 @@ def postgres_list
       }
     else
       dataset = dataset.eager(:representative_server, :timeline)
-      @postgres_databases = Serializers::Postgres.serialize(dataset.all, {include_path: true})
+      resources = dataset.all
+        .group_by { |r| r[:parent_id] || r[:id] }
```
byucesoy, line 64: Just realized it. This would put forks next to read replicas. How about this instead;

```
.group_by { |r| r.read_replica? ? r[:parent_id] : r[:id] }
```

model/postgres/postgres_server.rb
```diff
@@ -136,6 +140,14 @@ def needs_recycling?
     vm.display_size != resource.target_vm_size || storage_size_gib != resource.target_storage_size_gib
   end
 
+  def replica_lsn_caught_up
```
byucesoy, line 143: This function can be used for standbys as well, no need to prefix with `replica`.

Actually, I think we should use this in wait_catch_up label for standbys as well. That way we wouldn't need different logic for nearly same functionality.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -410,14 +435,24 @@ def before_run
 
   label def prepare_for_take_over
     decr_take_over
-    hop_taking_over if postgres_server.resource.representative_server.nil?
+
+    hop_taking_over if postgres_server.read_replica? || postgres_server.resource.representative_server.nil?
```
byucesoy, line 439: We should increment destroy for representative_server of read replica cluster here as well, no? Do we need this exception for read replicas?

That would also save you same logic from the next label.

model/postgres/postgres_server.rb
furkansahin: I'm not sure if that's a good idea because I think the logic for standby is better. The reason is, it only queries the primary. For read replicas, the lag is calculated by querying both of the servers. If you think that's a fair trade-off, and want to have a single logic, I can still change it, though. However, I believe it's better to keep it this way.

prog/postgres/postgres_server_nexus.rb
byucesoy: I think we are still risking some data async issues. Two clients can be connected to different read replicas at the same time and would read different data from each.

Also, I think in both cases, there would be some down time. In the current case, the down time would be much shorter (only connection drop). If we combine the logic, there would be 4-5 seconds of down time, which is acceptable I think.

model/postgres/postgres_server.rb
byucesoy: `wait_catch_up` label only gets executed for few hours max and during that time, I think it is OK to run query in both places.

That said, I believe, the recovering server does not accept any queries until some point. Does that affect us here? Though. it would affect both standby and read replicas. IIRC, not needing to deal that case was one of my reasons to only query primary in the past.

byucesoy (approved):

PR #3139, opened 2025-04-10
Rewrite first part of DEVELOPERS.md to use mise

fdr:
Note: We should delete `.envrc` after a transition period. If `asdf`
is uninstalled but `direnv` remains active, it will complain that `use
asdf` doesn't work. This emits an annoying message, but has no other
harmful effects.

Motivation for this rewrite:

`asdf-direnv` has not been updated for the Go-rewritten version of
`asdf`. This creates problems for anyone following these
instructions. I originally used `asdf-direnv` because some programs
(if memory serves, the `pq` gem compilation running `pg_config`) don't
work when shimmed. I also preferred the speed of `$PATH` manipulation
over using shim programs.

The author of `asdf-direnv` stopped maintaining it months ago, since
`mise` offers a similar model to `asdf` but is based on `$PATH` rather
than shims. As a result, he switched to `mise`, and now so are we.

There are generally fewer steps when using `mise`, but I made
`DEVELOPERS.md` a bit longer to include shell activation instructions
and system dependencies, so the instructions have grown slightly
overall. Hopefully, this will speed up the process, since it is less
necessary to visit other manual pages.

Another major difference is that `mise install` tries to download and
install *every* dependency, so it’s necessary to set up system
dependencies first. This required some reordering and restructuring.

Finally, while it was theoretically possible to manage Postgres by
other means, having it in `.tool-versions` made this awkward and often
painful in practice—even though the previous manual implied it should
be easy. This has now been changed to use the `mise.local.toml`
feature as an opt-in.

----

> [!IMPORTANT]
> Rewrites `DEVELOPERS.md` to use `mise` for software version management, adding detailed setup instructions and a `mise.local.toml.template` for Postgres.
> 
>   - **Behavior**:
>     - Rewrites `DEVELOPERS.md` to replace `asdf` with `mise` for managing software versions.
>     - Adds detailed instructions for installing and setting up `mise`, including shell integration and autocompletion.
>     - Introduces `mise.local.toml.template` for Postgres configuration.
>   - **Installation**:
>     - Provides system dependency installation instructions for Ruby and Postgres on macOS and Ubuntu.
>     - Describes `mise install` process to install all required dependencies.
>   - **Environment**:
>     - Details how to check `mise`-set environment variables using `mise env`.
>     - Explains how to manage Postgres with `mise` and the use of `mise.local.toml`.
>   - **Misc**:
>     - Updates Postgres setup instructions to use `mise` instead of `asdf`.
>     - Updates Node.js installation and verification steps to reflect `mise` usage.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 76dd8715fffee7f9dfd31c3d66f8bdfec798137c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -1,111 +1,228 @@
 # Clover
 
-Clover is the codename for Ubicloud's software, which consists of the control plane, data plane, and web console program for managing virtual machines and other applications.
+Clover is the codename for Ubicloud's software. It includes a control
+plane, a data plane, and a web console for managing virtual machines
+and other applications.
 
-It's a Ruby program that connects to Postgres.
+It is a Ruby program that connects to Postgres.
 
-The source code organization is based on the [Roda-Sequel
-Stack](https://github.com/jeremyevans/roda-sequel-stack), though a
-number of development choices have been modified.  As the name
-indicates, this project uses [Roda](https://roda.jeremyevans.net/)
-(for HTTP code) and [Sequel](http://sequel.jeremyevans.net/) (for
-database queries).
+The source code is organized based on the [Roda-Sequel
+Stack](https://github.com/jeremyevans/roda-sequel-stack), though
+several development choices have been modified. As the name indicates,
+this project uses [Roda](https://roda.jeremyevans.net/) (for HTTP
+handling) and [Sequel](http://sequel.jeremyevans.net/) (for database
+queries).
 
 Web authentication is managed with
 [Rodauth](http://rodauth.jeremyevans.net/).
 
-It communicates with servers using SSH, via the library
-[net-ssh](https://github.com/net-ssh/net-ssh).
+Clover communicates with servers using SSH via the
+[net-ssh](https://github.com/net-ssh/net-ssh) library.
 
-The tests are written using [RSpec](https://rspec.info/).
+Tests are written using [RSpec](https://rspec.info/).
 
-Code is automatically linted and formatted using
+Code is automatically linted and formatted with
 [RuboCop](https://rubocop.org/).
 
-Web console is designed with [Tailwind CSS](https://tailwindcss.com)
-based on components from [Tailwind UI](https://tailwindui.com). It uses
-jQuery for interactivity.
+The web console is designed with [Tailwind
+CSS](https://tailwindcss.com), based on components from [Tailwind
+UI](https://tailwindui.com), and uses jQuery for interactivity.
 
 ## Development Environment
 
-We suggest using [asdf-vm](https://github.com/asdf-vm/asdf) to manage
-software versions.  There is a [.tool-versions file](.tool-versions)
-that `asdf-vm` reads, and it is kept up to date.
+We recommend using [mise](https://mise.jdx.dev) to manage software
+versions. `mise` reads the `.tool-versions` file maintained in the
+repository.
 
-In the case of Ruby, obtaining a matching version is most
-important, because it is constrained in the [Gemfile](Gemfile).
+For Ruby, obtaining a matching version is especially important because
+it is constrained in the [Gemfile](Gemfile).
 
-Though, any method of obtaining Ruby and Postgres is adequate.
+### Install mise
 
-### Install asdf-vm and plugins
+If you are using `mise`, follow the instructions in the [Getting
+Started Manual](https://mise.jdx.dev/getting-started.html). There is a
+stand-alone installer, but you may prefer the Homebrew (`brew install
+mise`) or Debian/Ubuntu apt repository options, which are also
+documented on that page.
 
-If using `asdf-vm`, follow the instructions at the [Getting Started
-Manual](https://asdf-vm.com/guide/getting-started.html). There are
-three general steps:
-
-1. Download some common dependencies, `git` and `curl`. You may
-   already have them.
-2. Use `git` to clone `asdf-vm` into your home directory
-3. Source it into your shell automatically
-
-Having done so, typing `asdf` will yield a bunch of help text:
+After installing `mise`, typing `mise` will display help text:
 
 ```sh
-$ asdf
-version: v0.14.1-f00f759
+$ mise
+The front-end to your dev env
 
-MANAGE PLUGINS
-asdf plugin add <name> [<git-url>]      Add a plugin from the plugin repo OR,
+Usage: mise [OPTIONS] [TASK] [COMMAND]
+
+Commands:
 [...]
 ```
 
-We like to have these plugins (you can paste these commands):
+`mise` has [shell integration instructions in its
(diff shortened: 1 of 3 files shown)


DEVELOPERS.md
```diff
@@ -171,58 +109,52 @@ than with Ruby.
 
 Node.js is required to work with frontend tooling such as tailwind-cli.
 
-Install Node.js. `asdf` will consult the `.tool-versions`
+Install Node.js. `mise` will consult the `.tool-versions`
 file to select the version.
 
 ```sh
-$ asdf install nodejs
+$ mise install nodejs
```
enescakir, line 116: Since we have `nodejs` and `postgres` in our `.tool-versions` file, do we need to run `mise install nodejs` separately? Doesn't `mise install` install everything?

fdr: it does, I was going to take another pass to try to make it more idiomatic after getting rid of asdf. Unfortunately the "do it all" does stuff like expose you to _every_ system dependency requirement/compiler/etc.

fdr: Another thing I need to add in is how to hook in into the shell

mohi-kalantari:
I tried to go through the process here. there is one missing step: activating mise which developer must add a line to bashrc or zshrc.

something like this is mentioned in their docs:
```
echo 'eval "$(mise activate zsh)"' >> ~/.zshrc
```
or
```
echo 'eval "$(mise activate bash)"' >> ~/.bashrc
```

DEVELOPERS.md
```diff
@@ -28,138 +30,75 @@ jQuery for interactivity.
...
-```
-
 Then run:
 
 ```sh
-$ asdf install postgres
+$ mise install postgres
 ```
 
 There are many alternative ways to get Postgres, e.g. via system
```
enescakir, line 104: > There are many alternative ways to get Postgres, e.g. via system
package manager, Homebrew, Postgres.app, etc.  They are all acceptable, our version requirements for Postgres are more relaxed than with Ruby.

Since we have `postgres 15.8` in our `.tool-versions` file, running `asdf/mise install` will also install PostgreSQL. If we don't install it, a missing package warning will appear every time you enter the directory. If we really want to allow any PostgreSQL installation, we should remove it from the project's `.tool-versions` file and recommend adding it to the global `.tool-versions` file. Another option is to require the use of asdf/mise PG :D

fdr: what do you think? On the one hand, it's semi-convenient that it just takes care of itself...kind of. If I remember right, though, some issues how the asdf plugin managed the data directory were tedious.

On the downside, it may be rather weird to quite a few people in a way that using `mise` or a similar program for handling ruby, python, go, etc is not.

maybe I will use the `mise.local.toml` feature to do this, and have a version controlled template file. I will probably symlink something like `mise.local.toml.template` to `mise.local.toml` and leave it. This is not quite as nice for people that want to use the mise.local.toml feature with more tools, but I think we can deal with that then.

fdr:
> I tried to go through the process here. there is one missing step: activating mise which developer must add a line to bashrc or zshrc.
> 
> something like this is mentioned in their docs:
> 
> ```
> echo 'eval "$(mise activate zsh)"' >> ~/.zshrc
> ```
> 
> or
> 
> ```
> echo 'eval "$(mise activate bash)"' >> ~/.bashrc
> ```

I added a section for this, which is net-new, but I think a keen addition. Some of `mise`'s simplifications to process (no direnv plugin) freed up a bit of space.

DEVELOPERS.md
fdr: @enescakir done that way for now at least.

DEVELOPERS.md
```diff
@@ -1,228 +1,244 @@
...
+zshrc="$HOME/.zshrc"
+grep -qF "$compdir" "$zshrc" 2>/dev/null || \
+  printf '\nfpath=(%s $fpath)\n' "$compdir" >> "$zshrc"
+grep -qF "compinit" "$zshrc" 2>/dev/null || \
+  printf '\nautoload -U compinit; compinit\n' >> "$zshrc"
 
-```sh
-$ which ruby
-/home/ubicloud/.asdf/installs/ruby/3.2.5/bin/ruby
+echo "Installed mise zsh completion to $compfile and enabled it in $zshrc"
```
enescakir, line 132: I'm not sure if we should add the shell integration part. I think they can follow the instructions on mise doc. mise also shows some shell integration instructions after the installation is complete. More sections mean more things to maintain, and since we don't set up our development environment repeatedly, it's hard to keep everything updated.

enescakir (approved):

DEVELOPERS.md
fdr: I think we should give it a try, I think it'd speed things up for some people. If it breaks too much, let's remove it. In total honesty, it's not super well tested (I only tested for fish...a bit). It's LLM generated, plus a number of post-processing prompts to cross-examine it.

It seems to me that `mise`'s manual is restrained by the fact it supports a lot of shells and has enough users that have advanced shell configuration where they may have opinions about where things should go. The approach I take here is a bit more opinionated, spends more space on the subset of shells the staff has actually encountered and are supported, but is short enough that the scripts can be  inspected to see how the abstract relates to the tangible. I could have glommed *all* the configuration into a big script that attempts to be robust, but that's more opinion than I wanted to have.

enescakir: Makes sense. Let's try

jeremyevans:
Tried these instructions and did not run into any major issues.  I was able to install postgres, restore a backup of my asdf-postgres database to the mise-postgres database, and bundle install.

Minor issues I experienced:

Untrusted warning:

```
mise WARN  Config files in ~/code/ubicloud/mise.local.toml are not trusted.
Trust them with `mise trust`. See https://mise.jdx.dev/cli/trust.html for more information.
```

Fixed by  `mise trust mise.local.toml`

Idiomatic version file warning:

```
mise WARN  deprecated [idiomatic_version_file_enable_tools]: idiomatic version files like ~/code/ubicloud/Gemfile are currently enabled by default however this will change in mise 2026.1.1 to instead default to disabled.
You can remove this warning by explicitly enabling idiomatic version files for ruby with `mise settings add idiomatic_version_file_enable_tools ruby`.
See https://github.com/jdx/mise/discussions/4345 for more information
```

Fixed by creating `~/.config/mise/config.toml` with the following content:

```
[settings]
idiomatic_version_file_enable_tools = []
```

The documentation could be updated from ruby 3.2.8 to ruby 3.4.3.

PR #3144, opened 2025-04-14
Remove handle_get command from CNI and reorder ipv6 ips in handle_add response

mohi-kalantari:
Return the GUA ip ealier than ULA ip in CNI

In order to see the GUA ipv6 ip in pod's yaml status, we will return the GUA ip sooner than ULA since kubelet will store the first ipv6 as the pods ip and GUA ip is more usable to the customers

Remove handle_get function in CNI

This function is not called at all by kubelet or CRI and is only used by some cni tools. This commit removes the implementation and its tests.

----

> [!IMPORTANT]
> Reorder IPv6 addresses in CNI response to prioritize GUA over ULA and remove unused `handle_get` function.
> 
>   - **Behavior**:
>     - Reorder IPv6 addresses in `build_add_response` in `ubi_cni.rb` to return GUA before ULA.
>     - Remove `handle_get` function from `ubi_cni.rb` as it is unused by kubelet or CRI.
>   - **Tests**:
>     - Remove tests for `handle_get` in `ubi_cni_spec.rb`.
>     - Update `run` tests in `ubi_cni_spec.rb` to reflect removal of `handle_get`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 3cbcc40903b5f8b9a8dd46d5b6c98b81a40aa1f5. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/kubernetes/lib/ubi_cni.rb
+++ b/rhizome/kubernetes/lib/ubi_cni.rb
@@ -85,8 +85,8 @@ class UbiCNI
       interfaces: [{name: inner_ifname, mac: inner_mac, sandbox: "/var/run/netns/#{cni_netns}"}],
       ips: [
         {address: "#{ipv4_container_ip}/#{ipv4_container_ip.prefix}", gateway: ipv4_gateway_ip.to_s, interface: 0},
-        {address: "#{container_ula_ipv6}/#{container_ula_ipv6.prefix}", gateway: outer_link_local, interface: 0},
-        {address: "#{container_ipv6}/#{container_ipv6.prefix}", gateway: outer_link_local, interface: 0}
+        {address: "#{container_ipv6}/#{container_ipv6.prefix}", gateway: outer_link_local, interface: 0},
+        {address: "#{container_ula_ipv6}/#{container_ula_ipv6.prefix}", gateway: outer_link_local, interface: 0}
       ],
       routes: [{dst: "0.0.0.0/0"}],
       dns: {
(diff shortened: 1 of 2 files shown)


rhizome/kubernetes/lib/ubi_cni.rb
```diff
@@ -61,30 +48,32 @@ def handle_add
...
 
+    @logger.info "Generating MAC addresses for container #{container_id}"
+    inner_mac = gen_mac
     outer_mac = gen_mac
+    inner_link_local = mac_to_ipv6_link_local(inner_mac)
     outer_link_local = mac_to_ipv6_link_local(outer_mac)
     outer_ifname = "veth_#{container_id[0, 8]}"
 
+    @logger.info "Setting up veth pair for container #{container_id}"
+    r "ip link add #{outer_ifname} addr #{outer_mac} type veth peer name #{inner_ifname} addr #{inner_mac} netns #{cni_netns}"
```
fdr, line 62: this kind of stuff doesn't seem to have anything to do with locking, right? (it's part of the commit that says it's about logging. Ditto the extra logging, which is more than residual)

fdr (commented):
I think some of the commit messages could use copy-editing. I have been experimenting on and off with. At least one of them has a typo. A prompt like this acts as spell check and a bit more:

> Copy edit the following git commit message, with these instructions: do not add appeals to general justification that weren't there before. Don't add a bunch of adverbs. Do not omit technical details that I included. Do not change the wording to make it AI slop. Here is the text to edit:

rhizome/kubernetes/lib/ubi_cni.rb
mohi-kalantari: I forgot to update this branch after force pushing the destination branch, that's why you saw this line. It's irrelevant to the PR and now is not seen on the changes tab.

fdr (approved):

PR #3148, opened 2025-04-15
Postgres Read Replica promote feature

furkansahin:
## Postgres Read Replica promote feature
This simply adds both the backend and UI components to promote a read
replica. We make use of the existing take_over code path.

----

> [!IMPORTANT]
> Add feature to promote Postgres read replicas to primary databases, including backend logic, API routes, UI components, and tests.
> 
>   - **Behavior**:
>     - Adds `promote` semaphore to `PostgresResource` and `PostgresServer` classes.
>     - Implements promotion logic in `prog/postgres/postgres_resource_nexus.rb` and `prog/postgres/postgres_server_nexus.rb`.
>     - Updates `openapi.yml` to include `/promote` endpoint for promoting a read replica.
>   - **Routes**:
>     - Adds POST `/promote` route in `routes/project/location/postgres.rb` to handle promotion requests.
>     - Validates that only read replicas can be promoted.
>   - **UI**:
>     - Updates `views/postgres/show.erb` to include a "Promote" button for read replicas.
>     - Displays a message indicating the promotion process has started.
>   - **Tests**:
>     - Adds tests in `spec/prog/postgres/postgres_resource_nexus_spec.rb` and `spec/prog/postgres/postgres_server_nexus_spec.rb` for promotion logic.
>     - Adds API tests in `spec/routes/api/project/location/postgres_spec.rb` for the promote endpoint.
>     - Adds web tests in `spec/routes/web/project/postgres_spec.rb` for UI interactions related to promotion.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8140d884ae3b03939bbd74402ff06af193c10d3f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -17,7 +17,7 @@ class PostgresServer < Sequel::Model
   include HealthMonitorMethods
 
   semaphore :initial_provisioning, :refresh_certificates, :update_superuser_password, :checkup
-  semaphore :restart, :configure, :take_over, :configure_prometheus, :destroy, :recycle
+  semaphore :restart, :configure, :take_over, :configure_prometheus, :destroy, :recycle, :promote
 
   def configure_hash
     configs = {
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -335,17 +335,21 @@ SQL
     end
 
     if !is_in_recovery
-      timeline_id = Prog::Postgres::PostgresTimelineNexus.assemble(location_id: postgres_server.resource.location_id, parent_id: postgres_server.timeline.id).id
-      postgres_server.timeline_id = timeline_id
-      postgres_server.timeline_access = "push"
-      postgres_server.save_changes
-
-      refresh_walg_credentials
+      switch_to_new_timeline
 
       hop_configure
     end
 
     nap 5
+  end
+
+  def switch_to_new_timeline
+    timeline_id = Prog::Postgres::PostgresTimelineNexus.assemble(location_id: postgres_server.resource.location_id, parent_id: postgres_server.timeline.id).id
+    postgres_server.timeline_id = timeline_id
+    postgres_server.timeline_access = "push"
+    postgres_server.save_changes
+
+    refresh_walg_credentials
   end
 
   label def wait
@@ -380,6 +384,11 @@ SQL
 
     when_restart_set? do
       push self.class, frame, "restart"
+    end
+
+    when_promote_set? do
+      switch_to_new_timeline
+      hop_taking_over
     end
 
     if postgres_server.read_replica?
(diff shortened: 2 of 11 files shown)


prog/postgres/postgres_server_nexus.rb
```diff
@@ -428,6 +437,7 @@ def before_run
   end
 
   label def taking_over
+    decr_promote
```
fdr, line 440: you sure you want to mix this side effect with failure (rollback) -prone I/O? I mean, it may be *okay*-ish but, pointing it out.

routes/project/location/postgres.rb
```diff
@@ -215,6 +215,27 @@ class Clover
...
       end
 
+      r.post "promote" do
+        authorize("Postgres:view", pg.id)
+
+        unless pg.read_replica?
+          if api?
+            fail CloverError.new(400, "InvalidRequest", "Non read replica servers cannot be promoted!")
+          else
+            flash["error"] = "Non read replica servers cannot be promoted!"
```
fdr, line 225: seems like you want these two strings to be coupled (the same) but they aren't. Not sure if we use exclamation marks in our error messages either, although we don't have a style guide per se.

spec/prog/postgres/postgres_resource_nexus_spec.rb
```diff
@@ -364,6 +364,22 @@
       expect(postgres_resource).to receive(:set_firewall_rules)
       expect { nx.wait }.to nap(30)
     end
+
+    it "if read_replica and promote is set, promotes and naps" do
+      expect(nx).to receive(:when_promote_set?).and_yield
+      expect(postgres_resource).to receive(:read_replica?).and_return(true)
+      expect(postgres_resource).to receive(:servers).and_return([])
+      expect(postgres_resource).to receive(:update).with(parent_id: nil)
```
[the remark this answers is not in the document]
furkansahin: I'm not sure, I think I prefer the current state of the things for code readability's sake tbh.

model/postgres/postgres_server.rb
```diff
@@ -87,7 +87,7 @@ def configure_hash
         configs[:recovery_target_time] = "'#{resource.restore_target}'"
       end
 
-      if standby? || doing_pitr?
+      if standby? || doing_pitr? || read_replica?
```
byucesoy, line 90: We already enter this block for read replicas thanks to `doing_pitr?`, right? Do we need this?

Though, maybe `doing_pitr?` is not a good name for that method anymore. How about something like`restoring_from_blob_storage?`

prog/postgres/postgres_server_nexus.rb
```diff
@@ -436,23 +445,25 @@ def before_run
...
-  label def taking_over
-    if postgres_server.read_replica?
-      old_representative = postgres_server.resource.representative_server
-      old_representative.update(representative_at: nil)
-      postgres_server.update(representative_at: Time.now)
-      postgres_server.resource.incr_refresh_dns_record
-      old_representative.incr_destroy
-      hop_configure
-    end
+  label def read_replica_promote
```
byucesoy, line 456: I'll think about how to reduce the branching and specialized code for read replicas vs. regular databases. Let's merge the parent PR first, though.

model/postgres/postgres_server.rb
furkansahin: `doing_pitr?` is actually false for read-replica representative server because the timeline_access is `fetch`

furkansahin: nah, I am wrong here, you're right

byucesoy (approved):
Looks great. :shipit:

prog/postgres/postgres_server_nexus.rb
byucesoy: The new version seems much cleaner. Thanks. :+1:

PR #3150, opened 2025-04-16
OpenAPI: verifying references

fdr:
cc @geemus 

This obsoletes (but, if you look at the cumulative diff, does not change by even one byte) #3079. I mostly did work on the commit messages, patch grouping, and eliminating excess motions for ideas that came in and were backed out to increase retrospective clarity.

Nevertheless, the patch organization is a bit different (that is, less atomic) than usual because a large `openapi.yml` is mechanically tougher to read without commit context. The ruby component of it is handled more like normal though.


byucesoy (approved):
LGTM

PR #3154, opened 2025-04-17
Parallelize bootstrapping worker nodes

mohi-kalantari:
Previously we would wait for the control-plane bootstrapping to finish before starting the worker bootstrap. From now on, workers will start their bootstrap after the finish of the first CP node, all happening in parallel decreasing the total bootstrap time substantially.

----

> [!IMPORTANT]
> Parallelizes worker node bootstrapping in Kubernetes clusters, reducing total bootstrap time by starting after the first control-plane node finishes.
> 
>   - **Behavior**:
>     - Workers start bootstrapping after the first control-plane node finishes in `kubernetes_cluster_nexus.rb`.
>     - Uses semaphore `start_bootstrapping` to coordinate worker node bootstrapping.
>   - **Concurrency**:
>     - Adds `:start_bootstrapping` semaphore to `KubernetesNodepool` in `kubernetes_nodepool.rb`.
>     - In `kubernetes_cluster_nexus.rb`, checks if workers are ready to bootstrap after the first control-plane node.
>   - **Tests**:
>     - Updates in `kubernetes_cluster_nexus_spec.rb` to test new bootstrapping logic for 1-node and 3-node control-plane setups.
>     - Updates in `kubernetes_nodepool_nexus_spec.rb` to test worker node bootstrapping and semaphore logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 6fdb71835c1ac9e17da21674f2255f0b7d9e00d5. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -75,6 +75,14 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
   label def bootstrap_control_plane_vms
     nap 5 unless kubernetes_cluster.endpoint
 
+    # In 1-node control plane setup, we will wait until it's over
+    # In 3-node control plane setup, we start the bootstrapping after
+    # the first CP bootstrap
+    ready_to_bootstrap_workers =
+      kubernetes_cluster.cp_vms.count >= kubernetes_cluster.cp_node_count ||
+      (kubernetes_cluster.cp_node_count == 3 && kubernetes_cluster.cp_vms.count == 1)
+    kubernetes_cluster.nodepools.each(&:incr_start_bootstrapping) if ready_to_bootstrap_workers
+
     hop_wait_nodes if kubernetes_cluster.cp_vms.count >= kubernetes_cluster.cp_node_count
 
     bud Prog::Kubernetes::ProvisionKubernetesNode, {"subject_id" => kubernetes_cluster.id}
--- a/prog/kubernetes/kubernetes_nodepool_nexus.rb
+++ b/prog/kubernetes/kubernetes_nodepool_nexus.rb
@@ -26,9 +26,11 @@ class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
   end
 
   label def start
-    nap 30 unless kubernetes_nodepool.cluster.strand.label == "wait_nodes"
     register_deadline("wait", 120 * 60)
-    hop_create_services_load_balancer
+    when_start_bootstrapping_set? do
+      hop_create_services_load_balancer
+    end
+    nap 10
   end
 
   label def create_services_load_balancer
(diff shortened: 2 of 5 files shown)


prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -25,8 +25,16 @@ def before_run
     end
   end
 
+  def can_start_bootstrapping
+    if kubernetes_nodepool.cluster.cp_node_count == 1
```
aamederen, line 29: what about just checking `kubernetes_nodepool.cluster.cp_vms.first.strand`

or maybe something with `kubernetes_nodepool.cluster.client.kubectl`

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -58,14 +66,15 @@ def before_run
   end
 
   label def bootstrap_worker_vms
-    hop_wait if kubernetes_nodepool.vms.count >= kubernetes_nodepool.node_count
-    bud Prog::Kubernetes::ProvisionKubernetesNode, {"nodepool_id" => kubernetes_nodepool.id, "subject_id" => kubernetes_nodepool.kubernetes_cluster_id}
+    kubernetes_nodepool.node_count.times do
```
aamederen, line 69: I thought about "do we want to put a limit to the concurrency of provisioning", but then, we do a lot of concurrent provisioning of VMs thanks to GH actions, and k8s should probably handle ~10 nodes joining, so it could be OK.

aamederen (approved):

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -25,8 +25,16 @@ def before_run
     end
   end
 
+  def can_start_bootstrapping
+    if kubernetes_nodepool.cluster.cp_node_count == 1
+      kubernetes_nodepool.cluster.strand.label == "wait_nodes"
```
fdr, line 30: this kind of ting is weird and brittle, perhaps you want to use semaphores instead.

mohi-kalantari: changed to semaphores. thanks.

prog/kubernetes/kubernetes_nodepool_nexus.rb
mohi-kalantari: yes but after merging this PR https://github.com/ubicloud/ubicloud/pull/3237 we can be sure that we will have no issue.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -75,6 +75,14 @@ def before_run
   label def bootstrap_control_plane_vms
     nap 5 unless kubernetes_cluster.endpoint
 
+    # In 1-node control plane setup, we will wait until it's over
+    # In 3-node control plane setup, we start the bootstrapping after
+    # the first CP bootstrap
+    ready_to_bootstrap_workers =
+      kubernetes_cluster.cp_vms.count >= kubernetes_cluster.cp_node_count ||
```
aamederen, line 82: what about checking `kubernetes_cluster.cp_vms.first&.strand&.label == 'wait'`?

mohi-kalantari: I guess that doesn't work because the moment vm becomes running, its label would change to wait and first init and other kubeadm joins will happen at the same time.

aamederen (approved):

PR #3164, opened 2025-04-21
Add -W:strict_unused_block when running specs

jeremyevans:
This can find more cases where unused blocks are present.

I'm marking this as a draft, because it currently shows warnings in rspec.  There is a pull request to fix rspec, but it hasn't been merged yet: https://github.com/rspec/rspec/issues/175

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
-ruby 3.2.8
+ruby 3.4.3
 postgres 15.8
 nodejs 23.6.0
 golang 1.24.0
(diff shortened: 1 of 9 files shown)


enescakir:
@jeremyevans Should we rebase and merge if we think it's useful?

jeremyevans:
I believe it causes unnecessary warnings in rspec. rspec 4 fixes the issue, but is not yet released. After it is released, I'll rebase and update this.

PR #3168, opened 2025-04-22
Start listening on ipv6 ip of nodes in control plane nodes

mohi-kalantari:
So far, we only listened on the IPv4 IP of nodes, creating the k8s apiserver load balancer in IPv4 mode. From now on, we will also listen on the IPv6 IP, making the LB dual stack.

----

> [!IMPORTANT]
> Enable dual-stack (IPv4 and IPv6) support for Kubernetes control plane nodes by updating load balancer and node configuration.
> 
>   - **Behavior**:
>     - Update `kubernetes_cluster_nexus.rb` to configure load balancer as `LoadBalancer::Stack::DUAL` instead of `IPV4`.
>     - Modify `init-cluster` and `join-node` scripts to bind API server to IPv6 (`::`) and advertise IPv4 address.
>   - **Tests**:
>     - Update `kubernetes_cluster_nexus_spec.rb` to expect `LoadBalancer::Stack::DUAL` for load balancer stack.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 75a796928f431c1644a857f6847b56391d579f87. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/kubernetes/bin/init-cluster
+++ b/rhizome/kubernetes/bin/init-cluster
@@ -44,7 +44,17 @@ cluster_config = {
   "clusterName" => cluster_name,
   "controlPlaneEndpoint" => "#{lb_hostname}:#{port}",
   "apiServer" => {
-    "certSANs" => [lb_hostname]
+    "certSANs" => [lb_hostname],
+    "extraArgs" => [
+      {
+        "name" => "bind-address",
+        "value" => "::"
+      },
+      {
+        "name" => "advertise-address",
+        "value" => node_ipv4
+      }
+    ]
   },
   "networking" => {
     "podSubnet" => "#{private_subnet_cidr4},#{private_subnet_cidr6}",
--- a/rhizome/kubernetes/bin/join-node
+++ b/rhizome/kubernetes/bin/join-node
@@ -46,7 +46,23 @@ config = {
 
 if is_control_plane
   config["controlPlane"] = {
-    "certificateKey" => certificate_key
+    "certificateKey" => certificate_key,
+    "localAPIEndpoint" => {
+      "advertiseAddress" => node_ipv4,
+      "bindPort" => 6443
+    },
+    "apiServer" => {
+      "extraArgs" => [
+        {
+          "name" => "bind-address",
+          "value" => "::"
+        },
+        {
+          "name" => "advertise-address",
+          "value" => node_ipv4
+        }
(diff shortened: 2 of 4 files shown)


rhizome/kubernetes/bin/init-cluster
```diff
@@ -44,7 +44,17 @@ cluster_config = {
...
-    "certSANs" => [lb_hostname]
+    "certSANs" => [lb_hostname],
+    "extraArgs" => [
+      {
+        "name" => "bind-address",
+        "value" => "::"
+      },
+      {
+        "name" => "advertise-address",
+        "value" => node_ipv4
```
aamederen, line 55: what do you think about advertising the IPv6 address? Would that cause any issues for a client that doesn't have IPv6 for example?

rhizome/kubernetes/bin/join-node
```diff
@@ -46,7 +46,23 @@ config = {
...
+    },
+    "apiServer" => {
+      "extraArgs" => [
+        {
+          "name" => "bind-address",
+          "value" => "::"
+        },
+        {
+          "name" => "advertise-address",
+          "value" => node_ipv4
```
aamederen, line 62: out of curiosity: is this redundant over localAPIEndpoint.advertiseAddress?

aamederen (approved):

rhizome/kubernetes/bin/init-cluster
mohi-kalantari: that wouldn't cause any issues but all the internal communication would be through ipv4. IMO kubernetes has better support and integration with ipv4 that's why I decided to go with the ipv4 instead

rhizome/kubernetes/bin/join-node
mohi-kalantari: LocalAPIEndpoint: represents the endpoint of the API server instance to be deployed on this node.

advertise-address: The IP address on which to advertise the apiserver to members of the cluster. This address must be reachable by the rest of the cluster.

Not exactly redundant since their scope is different. more like explicitly setting the ipv4 in a dual stack vm to avoid unexpected ip discoveries or such.

PR #3170, opened 2025-04-22
Create VictoriaMetrics resource

shikharbhardwaj:
## Description

This is the initial commit for the Victoriametrics resource. This is a basic version which supports creating a single-node Victoriametrics instance backed by a VM, creating and managing appropriate certs and authentication configuration. This version:
- Installs victoriametrics and vmauth binaries on the VM
- Configures the systemd services for these two services
- Sets up the certificates and http basic authentication credentials

This version only supports a single-node version for simplicity, as current read/write benchmarks show VictoriaMetrics it can support upto 10k write targets (PG instnaces, for eg.) and around 2k concurrent read users on a single standard-8 VM.  In the future, we can add support for cluster mode, migrations, backup etc as needed.

Here is an example command to create a VictoriaMetrics resource: st = Prog::VictoriaMetrics::VictoriaMetricsResourceNexus.assemble( Config.victoriametrics_service_project_id,
"victoriametrics-test", Location::HETZNER_FSN1_ID, "vmuser", "pass", 20, 1, "standard-8")

## TODOs
- [x] Setup health_monitor
- [x] Unit tests
- [x] Use the storage_size parameter correctly 
- [x] Switch to daemonizer2

----

> [!IMPORTANT]
> Introduce VictoriaMetrics resource management with single-node setup, systemd configuration, and comprehensive testing.
> 
>   - **Behavior**:
>     - Adds `VictoriaMetricsResource` and `VictoriaMetricsServer` models to manage VictoriaMetrics instances.
>     - Implements `VictoriaMetrics::Client` for interacting with VictoriaMetrics API.
>     - Supports single-node VictoriaMetrics setup with systemd service configuration in `rhizome/victoria_metrics/bin/configure`.
>     - Handles certificate creation and management for secure communication.
>   - **Validation**:
>     - Adds `validate_victoria_metrics_username` and `validate_victoria_metrics_storage_size` in `validation.rb`.
>   - **Configuration**:
>     - Updates `config.rb` to include VictoriaMetrics related configurations.
>   - **Monitoring**:
>     - Updates `bin/monitor` to include `VictoriaMetricsServer` in monitorable resources.
>   - **Testing**:
>     - Adds unit tests for `VictoriaMetrics::Client` in `client_spec.rb`.
>     - Adds tests for `VictoriaMetricsResource` and `VictoriaMetricsServer` models.
>     - Includes tests for `VictoriaMetricsResourceNexus` and `VictoriaMetricsServerNexus` in `prog/victoria_metrics` specs.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c10a3406637cf55eb0f450be24af0b78db0108d1. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


prog/victoria_metrics/victoria_metrics_server_nexus.rb
```diff
@@ -0,0 +1,172 @@
...
+    donate
+  end
+
+  label def restart
+    case victoria_metrics_server.vm.sshable.cmd("common/bin/daemonizer --check restart_victoriametrics")
+    when "Succeeded"
+      victoria_metrics_server.vm.sshable.cmd("common/bin/daemonizer --clean restart_victoriametrics")
+      pop "victoriametrics server is restarted"
+    when "Failed", "NotStarted"
+      victoria_metrics_server.vm.sshable.cmd("common/bin/daemonizer 'systemctl restart victoriametrics && systemctl restart vmauth' restart_victoriametrics")
```
fdr, line 138: consider daemonizer2, 496ebc92e9abfcdc17ecf3b1916221b255ac177e

config.rb
```diff
@@ -120,6 +120,11 @@ def self.test?
   optional :minio_service_project_id, string
   override :minio_version, "minio_20240406052602.0.0_amd64"
 
+  # VictoriaMetrics
+  optional :victoria_metrics_service_project_id, string
+  override :victoria_metrics_host_name, "victoriametrics.ubicloud.com", string
```
byucesoy, line 125: Maybe just `metrics.ubicloud.com`? If we eventually replace victoria metrics with something else, it would be good to do in in non-breaking fashion from the dependents perspective.

rhizome/victoriametrics/bin/configure
```diff
@@ -0,0 +1,95 @@
+#!/bin/env ruby
```
byucesoy, line 1: It seems in some files we use `victoria_metrics` and in some others we use `victoriametrics`. It would be good to be consistent about them.

byucesoy (commented):
I had a brief look. I'll do a more detailed review later on.

config.rb
shikharbhardwaj: Yes that makes sense, updated

prog/victoria_metrics/victoria_metrics_server_nexus.rb
shikharbhardwaj: Nice, updated to use this. Looks shiny!

model/victoria_metrics_server.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require_relative "../model"
+
+class VictoriaMetricsServer < Sequel::Model
+  one_to_one :strand, key: :id
+  one_to_one :vm, key: :id, primary_key: :vm_id
```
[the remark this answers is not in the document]
fdr: 1) may be a lie, in that an additional unique constraint may make 1-to-1 more appropriate
1a) ...is there such a unique constraint? Looks like no, should there be one? 
2) @jeremyevans does many_to_one generate convenient association methods if, via constraint, the relationship is in fact one-to-one?

model/victoria_metrics_server.rb
```diff
@@ -0,0 +1,93 @@
...
+  def public_ipv6_address
+    vm.ephemeral_net6.nth(2).to_s
+  end
+
+  def private_ipv4_address
+    vm.private_ipv4.to_s
+  end
+
+  def ip6_url
+    "https://[#{public_ipv6_address}]:8427"
```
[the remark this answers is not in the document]
fdr: might be unnecessary. I really don't like this rule from ellipsis although it was right exactly once I've seen so far.

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,119 @@
...
+    DB.transaction do
+      ubid = VictoriaMetricsResource.generate_ubid
+      root_cert_1, root_cert_key_1 = Util.create_root_certificate(common_name: "#{ubid} Root Certificate Authority", duration: 60 * 60 * 24 * 365 * 5)
+      root_cert_2, root_cert_key_2 = Util.create_root_certificate(common_name: "#{ubid} Root Certificate Authority", duration: 60 * 60 * 24 * 365 * 5)
+
+      victoria_metrics_resource = VictoriaMetricsResource.create(
+        name: name,
+        location_id: location.id,
+        admin_user: admin_user,
+        admin_password: admin_password,
```
byucesoy, line 32: We can generate admin_password instead of accepting as a parameter, similar to MinIO and Postgres.

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,119 @@
...
+  subject_is :victoria_metrics_resource
+
+  def self.assemble(project_id, name, location_id, admin_user, admin_password, storage_size_gib, node_count, vm_size)
+    project = Project[project_id]
+    fail "No existing project" unless project
+
+    location = Location[location_id]
+    fail "No existing location" unless location
+
+    unless node_count == 1
```
byucesoy, line 15: If we only support 1 node, then no need to get this as parameter at this point.

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,119 @@
...
+      fail "Only single node is supported"
+    end
+
+    Validation.validate_vm_size(vm_size, "x64")
+    Validation.validate_name(name)
+    Validation.validate_minio_username(admin_user)
```
byucesoy, line 21: `validate_victoria_metrics_username` or if you want to re-use something, maybe `validate_name`?

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    DB.transaction do
+      ubid = VictoriaMetricsResource.generate_ubid
+      root_cert_1, root_cert_key_1 = Util.create_root_certificate(common_name: "#{ubid} Root Certificate Authority", duration: 60 * 60 * 24 * 365 * 5)
+      root_cert_2, root_cert_key_2 = Util.create_root_certificate(common_name: "#{ubid} Root Certificate Authority", duration: 60 * 60 * 24 * 365 * 5)
+
+      victoria_metrics_resource = VictoriaMetricsResource.create(
+        name: name,
```
byucesoy, line 29: You can omit repeating most of these variables. This should work fine

```
      victoria_metrics_resource = VictoriaMetricsResource.create(
        name:,
        location_id: location.id,
        admin_user:,
        admin_password:,
        root_cert_1:,
        root_cert_key_1:,
        root_cert_2:,
        root_cert_key_2:,
        target_vm_size: vm_size,
        target_storage_size_gib: storage_size_gib,
        project_id: project.id
      )
```

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,119 @@
...
+    fail "No existing location" unless location
+
+    unless node_count == 1
+      fail "Only single node is supported"
+    end
+
+    Validation.validate_vm_size(vm_size, "x64")
+    Validation.validate_name(name)
+    Validation.validate_minio_username(admin_user)
+
```
byucesoy, line 22: We should have validation for storage size as well.

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+require_relative "../../lib/util"
+
+class Prog::VictoriaMetrics::VictoriaMetricsResourceNexus < Prog::Base
+  subject_is :victoria_metrics_resource
+
+  def self.assemble(project_id, name, location_id, admin_user, admin_password, storage_size_gib, node_count, vm_size)
```
byucesoy, line 8: nit, but I'd put storage_size_gib after vm_size

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,119 @@
...
+  label def wait
+    if victoria_metrics_resource.certificate_last_checked_at < Time.now - 60 * 60 * 24 * 30 # ~1 month
+      hop_refresh_certificates
+    end
+
+    when_reconfigure_set? do
+      hop_reconfigure
+    end
+
+    nap 30
```
byucesoy, line 83: We can have much longer nap, since this label only checks for semaphores and when a semaphore is incremented the strand gets scheduled immediately.

Certificate renewal happens without semaphore, but since it is checked once a month, we can nap for 1 month.

prog/victoria_metrics/victoria_metrics_server_nexus.rb
```diff
@@ -0,0 +1,177 @@
...
+    reap
+    if leaf?
+      hop_wait
+    end
+    donate
+  end
+
+  label def wait
+    when_reconfigure_set? do
+      bud Prog::VictoriaMetrics::VictoriaMetricsSetup, {}, :configure
```
byucesoy, line 107: Why bud and not hop? bud would be slightly faster yes, but I think it makes it more difficult to reason about what the software would do.

For example resource nexus does this;
```
    victoria_metrics_resource.servers.map(&:incr_reconfigure)
    victoria_metrics_resource.servers.map(&:incr_restart)
```

but due to bug being parallel, it is not guaranteed that restart would happen after reconfigure. There might be other unexpected interactions as well, I didn't look into all cases carefully, but that is the point, parallel execution requires careful inspection.

I'd recommend using serial execution unless there is a good reason for parallelism.

prog/victoria_metrics/victoria_metrics_server_nexus.rb
```diff
@@ -0,0 +1,177 @@
...
+      decr_restart
+
+      push self.class, frame, "restart"
+    end
+
+    if victoria_metrics_server.certificate_last_checked_at < Time.now - 60 * 60 * 24 * 30 # ~1 month
+      hop_refresh_certificates
+    end
+
+    nap 10
```
byucesoy, line 121: Similar to other prog, this can nap for long time without loss of functionality.

prog/victoria_metrics/victoria_metrics_setup.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class Prog::VictoriaMetrics::VictoriaMetricsSetup < Prog::Base
```
byucesoy, line 3: This one is personal taste, so feel free to ignore, but I see setup as part of the original prog and usually embed there. One benefit is that I can follow the entire creation without jumping between files.

That said, other people prefer to split logic into multiple files, so it is indeed a matter of taste.

byucesoy (commented):
Looks good IMO. Left bunch of minor comments.

prog/victoria_metrics/victoria_metrics_server_nexus.rb
```diff
@@ -0,0 +1,206 @@
...
+  def_delegators :victoria_metrics_server, :vm
+
+  def self.assemble(victoria_metrics_resource_id, name)
+    vr = VictoriaMetricsResource[victoria_metrics_resource_id]
+    fail "No existing VictoriaMetricsResource" unless vr
+    Validation.validate_name(name)
+
+    DB.transaction do
+      ubid = VictoriaMetricsServer.generate_ubid
+      vm_st = Prog::Vm::Nexus.assemble_with_sshable(
```
enescakir, line 20: Do you want to place all servers of the same VictoriMetricsResource in the same private subnet?

migrate/20250425_victoria_metrics_resource.rb
```diff
@@ -0,0 +1,32 @@
...
+      column :root_cert_key_1, :text, collate: '"C"'
+      column :root_cert_2, :text, collate: '"C"'
+      column :root_cert_key_2, :text, collate: '"C"'
+      column :certificate_last_checked_at, :timestamptz, null: false, default: Sequel.lit("now()")
+      foreign_key :project_id, :project, type: :uuid, null: false
+      foreign_key :location_id, :location, type: :uuid, null: false
+    end
+    create_table(:victoria_metrics_server) do
+      column :id, :uuid, primary_key: true
+      column :name, :text, null: false
```
enescakir, line 23: Do we really need a name for the servers?

shikharbhardwaj: It's just for informational purpose for now, if we move to cluster version we'll have a bit more information here to make it easier to identify different server types

prog/victoria_metrics/victoria_metrics_server_nexus.rb
shikharbhardwaj: I left it as right now this only has single node support. But if we add a functionality to add a new server to existing resource it might be needed. I will add it just in case.

model/victoria_metrics_resource.rb
```diff
@@ -0,0 +1,57 @@
...
+    enc.column :root_cert_key_1
+    enc.column :root_cert_key_2
+  end
+
+  def hostname
+    "#{name}.#{Config.victoria_metrics_host_name}"
+  end
+
+  def root_certs
+    [root_cert_1, root_cert_2.to_s].join("\n") if root_cert_1 && root_cert_2
```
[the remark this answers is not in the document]
jeremyevans: `Array#join` calls `to_s` on all elements, so it isn't needed.

prog/victoria_metrics/victoria_metrics_server_nexus.rb
shikharbhardwaj: I added a private subnet with the resource

prog/victoria_metrics/victoria_metrics_setup.rb
shikharbhardwaj: Ack, I moved these into the same file to make things a bit simpler

rhizome/victoria_metrics/bin/install
```diff
@@ -0,0 +1,14 @@
...
+
+unless (ver = ARGV.shift)
+  fail "No version provided"
+end
+
+r "wget https://github.com/VictoriaMetrics/VictoriaMetrics/releases/download/v1.113.0/victoria-metrics-#{ver}.tar.gz -O victoria-metrics.tar.gz"
+r "tar xvf victoria-metrics.tar.gz -C /usr/local/bin/"
+
+r "wget https://github.com/VictoriaMetrics/VictoriaMetrics/releases/download/v1.113.0/vmutils-#{ver}.tar.gz -O vmutils.tar.gz"
+r "tar xvf vmutils.tar.gz -C /usr/local/bin/"
```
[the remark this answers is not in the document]
shikharbhardwaj: Nice catch, I will update this

prog/victoria_metrics/victoria_metrics_server_nexus.rb
```diff
@@ -0,0 +1,230 @@
...
+    when "Succeeded"
+      vm.sshable.d_clean("configure_victoria_metrics")
+      hop_wait
+    when "Failed", "NotStarted"
+      config_json = JSON.generate({
+        admin_user: resource.admin_user,
+        admin_password: resource.admin_password,
+        cert: victoria_metrics_server.cert,
+        cert_key: victoria_metrics_server.cert_key,
+        ca_bundle: victoria_metrics_server.resource.root_certs
```
enescakir, line 127: nit
```suggestion
        ca_bundle: resource.root_certs
```

prog/victoria_metrics/victoria_metrics_resource_nexus.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      firewall = Firewall.create_with_id(name: "#{victoria_metrics_resource.ubid}-firewall", location_id: location.id, description: "VictoriaMetrics default firewall", project_id: Config.victoria_metrics_service_project_id)
+
+      private_subnet_id = Prog::Vnet::SubnetNexus.assemble(Config.victoria_metrics_service_project_id, name: "#{victoria_metrics_resource.ubid}-subnet", location_id: location.id, firewall_id: firewall.id).id
+      victoria_metrics_resource.update(private_subnet_id: private_subnet_id)
+      victoria_metrics_resource.set_firewall_rules
+
+      Prog::VictoriaMetrics::VictoriaMetricsServerNexus.assemble(
+        victoria_metrics_resource.id,
+        "#{name}-node-01"
```
byucesoy, line 47: I agree with @enescakir on this name being not required. I guess this can be useful when we have multiple servers, but even then it is tricky to maintain. Two reasons:
- Over time, we would recycle some servers, so there would be gaps in the numbers.
- Every time we add a new server, you would need to parse existing VM names and find highest used number to generate correct suffix.

byucesoy (approved):

model/victoria_metrics_server.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def check_pulse(session:, previous_pulse:)
+    reading = begin
+      session[:victoria_metrics_client].health ? "up" : "down"
+    rescue
+      "down"
+    end
+    pulse = aggregate_readings(previous_pulse: previous_pulse, reading: reading)
+
+    if pulse[:reading] == "down" && pulse[:reading_rpt] > 5 && Time.now - pulse[:reading_chg] > 30 && !reload.checkup_set?
```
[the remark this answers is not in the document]
shikharbhardwaj: similar pattern used in `minio/client.rb` and works fine there

PR #3180, opened 2025-04-24
Bump cloud-hypervisor version to 45.0

jeremyevans:
The --disk option changed in cloud-hypervisor 36.0.  cloud-hypervisor now only accepts a single --disk option, though it can have multiple parameters after it.

Add some refactoring of rhizome code to make this simpler.

I tested this with basic standard-2 VM creation.  Could not try with burstables as my test box is in Finland and we do not support burstables there.  Could not try with PostgreSQL is I don't think we support 3rd party downloads of the necessary boot images.

----

> [!IMPORTANT]
> Update cloud-hypervisor to version 45.0, refactor VM setup, and update tests for new `--disk` option handling.
> 
>   - **Behavior**:
>     - Update cloud-hypervisor version to 45.0, modifying `--disk` option handling in `vm_setup.rb` and `nexus.rb`.
>     - Refactor `setup-vm` script to handle new `--disk` option format.
>     - Add support for `hugepages`, `ch_version`, and `firmware_version` in `nexus.rb` and `vm_setup.rb`.
>   - **Testing**:
>     - Replace `COVERAGE=rhizome bundle exec rspec -O /dev/null rhizome` with `bundle exec rake rhizome_spec` in `ci.yml`.
>     - Add `rhizome_spec` task to `Rakefile` for running data plane tests.
>     - Update tests in `nexus_spec.rb` and `update_ipv6_spec.rb` to reflect changes in VM setup and cloud-hypervisor version.
>   - **Misc**:
>     - Refactor `cloud_hypervisor.rb` to manage firmware and version downloads.
>     - Remove `arch_class_spec.rb` as it is no longer needed.
>     - Add `base64` gem to `Gemfile` for encoding support.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b044b6068748150304058b6232fed6630a0b3de4. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


rhizome/host/Gemfile
```diff
@@ -2,6 +2,8 @@
 
 source "https://rubygems.org"
 
+gem "base64"
```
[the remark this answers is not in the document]
jeremyevans: As the related commit message mentions, `base64` was in the standard library in Ruby 3.2, but is a bundled gem in Ruby 3.4, so it's best to specify it in the Gemfile.

enescakir:
I triggered the E2E test for it. Another good way to test dataplane changes is E2E tests. I think they have some burstable tests.

Rakefile
```diff
@@ -247,6 +247,11 @@ ensure
   File.delete(output_file) if File.file?(output_file)
 end
 
+desc "Run rhizome (data plane) tests"
+task "rhizome_spec" do
+  sh "COVERAGE=rhizome bundle exec rspec -O /dev/null rhizome"
```
enescakir, line 252: In the CI, we usually run rake tasks, so we can change this line to `bundle exec rake rhizome_spec` too

https://github.com/ubicloud/ubicloud/blob/ff5c10bf7e013359b4ff9013458aee1eb6ee07ef/.github/workflows/ci.yml#L116

jeremyevans: Makes sense, I pushed a commit for that.

rhizome/host/lib/cert_server_setup.rb
```diff
@@ -48,10 +48,7 @@ def vm_server_path
   end
 
   def package_url
-    Arch.render(
-      x64: "https://github.com/ubicloud/metadata-endpoint/releases/download/v#{server_version}/metadata-endpoint_Linux_x86_64.tar.gz",
-      arm64: "https://github.com/ubicloud/metadata-endpoint/releases/download/v#{server_version}/metadata-endpoint_Linux_arm64.tar.gz"
-    )
+    "https://github.com/ubicloud/metadata-endpoint/releases/download/v#{server_version}/metadata-endpoint_Linux_#{Arch.x64? ? "x86_64" : "arm64"}.tar.gz"
```
fdr, line 51: IMO: I don't like this conversion, where "else" assumes some architecture. `Arch.render` is more reliable to grep and had exhaustiveness checking. I could abide it, but previously, I forbade it.

jeremyevans: It's still easy to grep for `Arch.`.  I can understand the point of exhaustiveness checking, if you expect to move beyond 2 architectures (if not, YAGNI).  Anyway, if you would forbade the approach if someone else suggested it, I should not be special, I'll switch to an alternative approach.

rhizome/host/lib/cloud_hypervisor.rb
```diff
@@ -20,25 +25,46 @@ def path
...
+    v45 = Arch.render(
+      x64: new("45.0", "362d42eb464e2980d7b41109a214f8b1518b4e1f8e7d8c227b67c19d4581c250", "11a050087d279f9b5860ddbf2545fda43edf93f9b266440d0981932ee379c6ec"),
+      arm64: new("45.0", "3a8073379d098817d54f7c0ab25a7734b88b070a98e5d820ab39e244b35b5e5e", "a4b736ce82f5e2fc4a92796a9a443f243ef69f4970dad1e1772bd841c76c3301")
+    )
+    v35 = Arch.render(
+      x64: new("35.1", "e8426b0733248ed559bea64eb04d732ce8a471edc94807b5e2ecfdfc57136ab4", "337bd88183f6886f1c7b533499826587360f23168eac5aabf38e6d6b977c93b0"),
+      arm64: new("35.1", "071a0b4918565ce81671ecd36d65b87351c85ea9ca0fbf73d4a67ec810efe606", "355cdb1e2af7653a15912c66f7c76c922ca788fd33d77f6f75846ff41278e249")
+    )
+
+    if `lsb_release -d`.include?("Ubuntu 24.")
```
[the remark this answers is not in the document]
jeremyevans: Caching is pointless as this is evaluated only once.  Abstraction seems reasonable, but may be premature.

rhizome/host/lib/vm_setup.rb
```diff
@@ -653,11 +653,18 @@ def install_systemd_unit(max_vcpus, cpu_topology, mem_gib, storage_params, nics,
...
-        "--disk path=#{volume.image_path},readonly=on \\"
+        "path=#{volume.image_path},readonly=on"
       else
-        "--disk vhost_user=true,socket=#{volume.vhost_sock},num_queues=1,queue_size=256 \\"
+        "vhost_user=true,socket=#{volume.vhost_sock},num_queues=1,queue_size=256"
       end
     }
+    disk_params << "path=#{vp.cloudinit_img}"
+
+    disk_args = if CloudHypervisor::Version::DEFAULT.version >= "36"
```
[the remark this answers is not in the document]
jeremyevans: This is a fair point, though not an issue until version hits 100 (over 6 years away based on historical cloud-hypervisor releases).  Since only the major version is compared, a simple integer comparison could be used.  `Gem::Version` makes sense once minor versions are also being compared.

prog/vm/nexus.rb
```diff
@@ -8,6 +8,9 @@
 
 class Prog::Vm::Nexus < Prog::Base
   DEFAULT_SIZE = "standard-2"
+  # :nocov:
+  SETUP_VM_HUGEPAGES = Config.development? ? " hugepages=off" : ""
```
fdr: Let's keep this one in the vault until we figure out how to avoid the reboot in prep_host (to make effective the changes in `Prog::SetupHugepages`) and disable most/all of the hugepage allocation. The reboot is slow, and a reboot where we request the system to make a huge fraction of it available in contiguous, empty blocks for hugepages is even slower.

I think for most development, fewer reboots and faster reboots is the winning side of this formula, but when I first asked about this, I forgot about the hugepage requirements of SPDK, which is a fly in the ointment: granted, it only needs a handful, and finding less contiguous memory is much faster, but we're also pretty sure we're going to replace it with a different program that almost certainly can make hugepages optional.

Also, if we go through the trouble of allocating the hugepages, and then don't use them, the VM allocations will come out of the tiny sliver of non-hugepage memory remaining.

rhizome/host/lib/cert_server_setup.rb
fdr: I suppose grepping for all `Arch` is a solution, mostly, I like that grepping for `Arch.render` can identify cases that are intended exhaustive to be exhaustive, rather than `Arch.x64?` and `Arch.arm64?`, which I anticipated -- unrealized, I think -- other forms of use (when needing to run different imperative code around the branch, or when one architecture requires a special action, e.g. `do_this if Arch.arm64?`. But selecting among two objects has been enough for everything, which I did not expect. If I knew that from the outset, maybe I would have done things the normal way.

prog/vm/nexus.rb
fdr: I think that's sufficient, we need to do a bit of follow-up work (that will change with a new storage stack that may not oblige hugepages) to capture the advantage of not slowing down reboots / incurring an extra reboot when trapped in a prep_host iteration loop) in development

fdr:
Okay, so most interesting new macro-ish-features of this patch that I'm still digesting is:

1. a class method that is used in two progs (`setup_vm_options_str`)
2. a new argument protocol held consistent in two rhizome programs
3. A coupling between a elements in a stack frame and a helper method
 
We don't have obvious patterns for these. I have to reread this with these in mind, I found it while disassembling the cumulative diff for #3214

jeremyevans:
Closing in favor of #3221

PR #3183, opened 2025-04-24
Update pry version to 0.15.2

enescakir:
After we upgraded Ruby to version 3.4.3, pry started showing the following warnings:

    ./ruby/3.4.3/lib/ruby/gems/3.4.0/gems/pry-0.14.2/lib/pry/command_state.rb:3: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
    You can add ostruct to your Gemfile or gemspec to silence this warning.
    ./ruby/3.4.3/lib/ruby/3.4.0/readline.rb:4: warning: reline was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
    You can add reline to your Gemfile or gemspec to silence this warning.

pry has already removed ostruct usage in the latest version released in December 2024. It's interesting that Dependabot didn't upgrade it automatically.

For the reline warning, pry has already included reline in the Gemfile
but not in the gemspec. I believe pry requires readline, and readline
needs reline. I added to our Gemfile until pry added it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ gem "puma", ">= 6.2.2"
 gem "rack-unreloader", ">= 1.8"
 gem "rake"
 gem "refrigerator", ">= 1"
+gem "reline" # Remove it when pry adds it as a dependency
 gem "roda", ">= 3.89"
 gem "rodauth", github: "jeremyevans/rodauth", ref: "2a72b30136c512ea2853ad9dbc126c067bef89e9"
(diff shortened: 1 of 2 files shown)


enescakir:
@jeremyevans What do you think about the reline warning? What's a possible solution that doesn't involve adding reline to our Gemfile?

jeremyevans:
We can temporarily add `reline` to our Gemfile to avoid the warning, and remove it when it is added by `pry`.

jeremyevans (approved):

jeremyevans (approved):

fdr (approved):

PR #3184, opened 2025-04-24
Small fixes for read replica promote

furkansahin:
## Promote read replica infinite promotion bug
Since we are cleaning up the parent_id in resource promote code, the
server doesn't see itself as a read replica which in turn causes the
decrement of promote semaphore to be overlooked. In this commit, we are
making sure that the promote semaphore is decremented for sure.


prog/postgres/postgres_server_nexus.rb
```diff
@@ -447,15 +447,15 @@ def switch_to_new_timeline
   end
 
   label def taking_over
```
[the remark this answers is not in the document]
furkansahin: wrong

prog/postgres/postgres_server_nexus.rb
```diff
@@ -447,15 +447,15 @@ def switch_to_new_timeline
...
+    when_promote_set? do
+      postgres_server.update(synchronization_status: "ready")
+      decr_promote
```
byucesoy, line 452: How about we move `decr_promote` to the wait label and `synchronization_status` update to `Succeeded` branch below. Then we can remove `when_promote_set` branch?

byucesoy (approved):

prog/postgres/postgres_server_nexus.rb
furkansahin: Yeah, that makes sense, updated it.

PR #3185, opened 2025-04-24
Prioritize AX102 hosts for premium CPU runner testers

enescakir:
- **Remove the transparent_cache feature flag**
  We've already added a column to the GithubInstallation model, and it's
  no longer in use.
  

- **Prioritize AX102 hosts for premium CPU runner testers**
  We will introduce a new runner tier soon; I'm working on a PR to enable
  it.
  
  We have some AX102 machines in our fleet. Until the new runner tier is
  launched, we want to give higher priority to some customers to test
  them.
  
  We get the runner object in the VM nexus to check if it’s a preferred
  performance CPU. It’s a bit of an antipattern, but making changes to the
  codebase would require a lot of effort. We need to pass the preference
  from runner nexus to VM nexus to allocator. Additionally, it allocates
  the VM at the `start` label, not in `assemble`, so we need to persist
  the preference in the stack. To avoid all of this, I just try to get the
  runner if the location is "github-runners." This code will be removed
  after the new runner tier is launched.
  
  Since we don't have a VM pool for them, we don't pick from the pool if
  it's a preferred performance CPU.
  

----

> [!IMPORTANT]
> Prioritize AX102 hosts for premium CPU runner testers and remove unused `transparent_cache` feature flag, updating allocation logic and tests accordingly.
> 
>   - **Behavior**:
>     - Prioritize AX102 hosts for premium CPU runner testers by adding `prioritize_performance_cpu` flag in `allocator.rb`.
>     - Remove `transparent_cache` feature flag from `project.rb`.
>   - **Allocation Logic**:
>     - Update `allocate()` in `allocator.rb` to include `prioritize_performance_cpu` parameter.
>     - Modify `calculate_score()` in `allocator.rb` to adjust scoring based on `prioritize_performance_cpu`.
>   - **VM Handling**:
>     - Update `pick_vm()` in `github_runner.rb` to consider `performance_runners` feature flag.
>     - Adjust `before_run()` in `nexus.rb` to handle performance prioritization.
>   - **Tests**:
>     - Add tests in `github_runner_spec.rb`, `nexus_spec.rb`, and `allocator_spec.rb` to verify prioritization logic and removal of `transparent_cache`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7eb9ed6750b6bf4f1759f14c01a36fd34a0e23b8. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -12,7 +12,7 @@ module Scheduling::Allocator
     @target_host_utilization ||= Config.allocator_target_host_utilization
   end
 
-  def self.allocate(vm, storage_volumes, distinct_storage_devices: false, gpu_count: 0, allocation_state_filter: ["accepting"], host_filter: [], host_exclusion_filter: [], location_filter: [], location_preference: [])
+  def self.allocate(vm, storage_volumes, distinct_storage_devices: false, gpu_count: 0, allocation_state_filter: ["accepting"], host_filter: [], host_exclusion_filter: [], location_filter: [], location_preference: [], prioritize_performance_cpu: false)
     request = Request.new(
       vm.id,
       vm.vcpus,
@@ -34,7 +34,8 @@ module Scheduling::Allocator
       vm.cpu_percent_limit,
       true, # use slices
       Option::VmFamilies.find { _1.name == vm.family }&.require_shared_slice || false,
-      vm.project.get_ff_allocator_diagnostics || false
+      vm.project.get_ff_allocator_diagnostics || false,
+      prioritize_performance_cpu
     )
     allocation = Allocation.best_allocation(request)
     fail "#{vm} no space left on any eligible host" unless allocation
@@ -64,7 +65,8 @@ module Scheduling::Allocator
     :cpu_percent_limit,
     :use_slices,
     :require_shared_slice,
-    :diagnostics
+    :diagnostics,
+    :prioritize_performance_cpu
   ) do
     def initialize(*args)
       super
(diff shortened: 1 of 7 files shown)


scheduling/allocator.rb
```diff
@@ -316,6 +318,11 @@ def calculate_score
       # penalty for ongoing vm provisionings on the host
       score += @candidate_host[:vm_provisioning_count] * 0.5
 
+      # prioritize AX102 for our premium CPU testers. Since we currently have
+      # a small number of AX102s in the fleet, decrease the score by 5
+      # instead of 0.5 to compensate the concurrent VM penalty.
```
bsatzger, line 323: the concurrent VM penalty is only to avoid concurrent vm provisioning, not concurrently running vms. even for the performance use case, I'd assume we wouldn't want to provision many vms at the same time on a single host. we saw provisioning times spike for such cases.

enescakir: We currently have very few AX102 units in the fleet :facepalm: So, when a customer wants to test a premium CPU and requests runners more than that number in the same workflow, we can't provide premium runners for all jobs due to concurrent VM provisioning penalty. A few extra seconds in provisioning times aren't a big issue for them.

bsatzger: just an idea: if we want to test AX102 with a preference for certain customers, we could do something like that:
`score += 0.5 * (@request.prioritize_performance_cpu ? -1 : 1) if @candidate_host[:total_cores] == 16`

otherwise, it's not unlikely that AX102 are already fully utilized when a tester comes along.

bsatzger (approved):

PR #3195, opened 2025-04-25
Default to setting hugepages=off in development mode

jeremyevans:
When turning off hugepages, you need to turn on shared memory in cloud-hypervisor, or you get the error:

  Error validating configuration: Using vhost-user requires using shared memory or huge pages

The prog changes should be backwards compatible with older rhizome, since older rhizome will ignore extra arguments to setup-vm.


jeremyevans:
I'm using the infrastructure for this to support choosing a cloud-hypervisor version, so I'm closing this, and will be including it in #3180.

PR #3202, opened 2025-04-28
Add bin/metrics-collector to collect and buffer metrics

shikharbhardwaj:
This change adds a systemd-timer triggered script to collect prometheues-style metrics from a configured list of endpoints on the metrics producer side (eg. Postgres server). These will be written to the metrics storage DB (VictoriaMetrics) by bin/monitor. The buffer exists in case certain parts of the stack (either monitor or VictoriaMetrics) are unavailable, we are able to catch up and recover without losing data. With the default buffer size (20MiB), we can tolerage about 30 minutes of downtime for PG metrics.

To add this functionality to other resources, one can configure the systemd-timer in a similar manner as done in this change for PostgresServer and provide the configuration as part of the `metrics_config` method, like this:

```
{
  # Array of endpoints to collect metrics from
  endpoints: [],

  # Maximum size of the pending buffer in MiB
  max_pending_buffer_size_mib: 20,

  # Interval for collecting metrics in seconds or as a time span string
  interval: "15s",

  # Additional label names and values to be added to the collected metrics
  additional_labels: {"foo": "bar"}
}
```


rhizome/postgres/bin/configure
shikharbhardwaj, line 1: Actually on second thought, I will move this to PostgresServerNexus under `configure_promethues` or a different label.

model/postgres/postgres_server.rb
```diff
@@ -246,6 +247,17 @@ def self.run_query(vm, query)
   def run_query(query)
     self.class.run_query(vm, query)
   end
+
+  def metrics_config
+    {
+      endpoints: [
+        "https://localhost:9090/federate?match[]=%7B__name__%3D~%22.%2B%22%7D"
```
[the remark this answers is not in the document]
fdr: Revised this prompt to make it even more specific.

rhizome/postgres/bin/metrics-collector
```diff
@@ -0,0 +1,42 @@
...
+  r "curl --fail --insecure #{Shellwords.escape(endpoint)} >> /tmp/#{filename}.txt"
+end
+
+r "mv /tmp/#{filename}.txt #{metrics_dir}/pending/#{filename}.txt"
+
+# Remove pending files till the size of the pending directory is less than the max_pending_buffer_size
+pending_files = Dir.glob("#{metrics_dir}/pending/*").sort
+
+pending_files.each do |pending_file|
+  break if r("du -sm #{metrics_dir}/pending | cut -f1").to_i < max_pending_buffer_size_mib
```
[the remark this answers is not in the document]
fdr: Ellipsis is right, but per commonly, has made a bad suggestion: `du` is quite expensive, and this seems more complex than necessary, why not keep X files, and trim the most obsolete? If coded right, it should only have to read the directory entries (`getdents`), e.g. via `Dir.children`. If you have named them right, then you can know which to unlink by their name encoding their obsolescence (e.g. a date)

rhizome/postgres/bin/metrics-collector
```diff
@@ -0,0 +1,43 @@
...
+require_relative "../../common/lib/util"
+
+if ARGV.count != 2
+  fail "Wrong number of arguments. Expected 2, Given #{ARGV.count}"
+end
+
+metrics_dir = ARGV[0]
+metrics_config_path = ARGV[1]
+
+r "mkdir -p #{metrics_dir}/pending"
```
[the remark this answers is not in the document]
fdr: .shellescape is more terse, @ellipsis-dev you should suggest that instead.

rhizome/postgres/bin/metrics-collector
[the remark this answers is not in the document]
shikharbhardwaj: I wanted to keep the buffer size as amount of disk instead of a fixed max count, mainly because the file size of metrics fetched in one iteration can vary depending on the target. But I guess it should be fine this way, I was not worried about perf in this case because the number of files here would be small < 20 in this case.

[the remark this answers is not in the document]
fdr: < 150 syscalls vs one or two, many more forks or memory allocations, which can hit OOM on Postgres guests in particular, no need to be proliferate. I am just surprised that the extent that the size occupancy is dynamic is would be to that degree.

rhizome/postgres/bin/metrics-collector
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env ruby
+
+# frozen-string-literal: true
```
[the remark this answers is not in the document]
fdr: I would have thought rubocop would rewrite this automatically (and complain otherwise, it inserts these)

rhizome/postgres/bin/metrics-collector
```diff
@@ -0,0 +1,51 @@
...
+
+# Remove pending files till the number of files is less than max_pending_buffer_size.
+pending_files = Dir.children("#{q_metrics_dir}/pending").sort
+num_files_to_remove = [0, pending_files.count - max_pending_buffer_size].max
+
+if num_files_to_remove.positive?
+  files_to_remove = pending_files[0...(num_files_to_remove - 1)]
+
+  files_to_remove.each do |pending_file|
+    r "mv #{pending_file} #{q_metrics_dir}/done"
```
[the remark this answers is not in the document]
fdr: @ellipsis-dev Avoid more forks and simplify, use `File.rename`

model/postgres/postgres_server.rb
enescakir, line 254: Just out of curiosity, what is `%7B__name__%3D~%22.%2B%22%7D`?

shikharbhardwaj: This is urlencoded from `match[]={__name__=~".+"}`, which means to match all metrics to federate from the target prometheus

prog/postgres/postgres_server_nexus.rb
```diff
@@ -236,6 +236,52 @@ def before_run
     hop_wait
   end
 
+  label def configure_metrics
+    metrics_config = postgres_server.metrics_config
+    metrics_dir = metrics_config[:metrics_dir]
+    vm.sshable.cmd("mkdir -p #{metrics_dir}")
+    vm.sshable.cmd("tee #{metrics_dir}/config.json > /dev/null", stdin: metrics_config.to_json)
+
+    metrics_service = <<SERVICE
```
byucesoy, line 245: Similar to above, systemd config does not seem to have instance related information. It can be moved to image generation.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -236,6 +236,52 @@ def before_run
...
+
+[Install]
+WantedBy=timers.target
+TIMER
+    vm.sshable.cmd("sudo tee /etc/systemd/system/postgres-metrics.timer > /dev/null", stdin: metrics_timer)
+
+    vm.sshable.cmd("sudo systemctl daemon-reload")
+
+    when_initial_provisioning_set? do
+      vm.sshable.cmd("sudo systemctl enable --now postgres-metrics.timer")
```
byucesoy, line 278: The logic (and the area of work) is in close proximity of `configure_prometheus`. How about (after moving the above to image generation) combining both of them under this label `configure_metrics`?

rhizome/postgres/bin/metrics-collector
```diff
@@ -0,0 +1,47 @@
...
+
+require_relative "../../common/lib/util"
+
+if ARGV.count != 2
+  fail "Wrong number of arguments. Expected 2, Given #{ARGV.count}"
+end
+
+metrics_dir = ARGV[0]
+q_metrics_dir = metrics_dir.shellescape
+metrics_config_path = ARGV[1]
```
byucesoy, line 17: I feel like these 2 parameters can be hard-coded.

rhizome/postgres/bin/metrics-collector
```diff
@@ -0,0 +1,47 @@
...
+
+fail "Metrics config file does not exist: #{metrics_config_path}" unless File.exist?(metrics_config_path)
+fail "Metrics directory does not exist: #{metrics_dir}" unless Dir.exist?(metrics_dir)
```
byucesoy, line 20: In which case these wouldn't exist? I guess, maybe caller would send wrong parameter. But if we hard-code them, then no need for these checks.

model/postgres/postgres_server.rb
```diff
@@ -246,6 +247,18 @@ def self.run_query(vm, query)
   def run_query(query)
     self.class.run_query(vm, query)
   end
+
+  def metrics_config
```
byucesoy, line 251: Given that there is no server specific field, how about moving this to image generation? Or they can be hard-coded where we are using them. Any reason for adding these as configs?

shikharbhardwaj: `bin/monitor` would rely on some of these, specifically the `additional_labels` and `metrics_dir` fields, so wanted to keep them in once place where it can be referenced from both places

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Makes sense, I wanted to keep it here for 2 main reasons:
1. In case we need to change the collection interval, it can be easily done vs changing the image and service for each instance
2. Deploying this change is easier vs the image rollout
But I don't have a strong preference overall

rhizome/postgres/bin/metrics-collector
byucesoy: Based on the conversation in other threads, I think it makes sense to keep metrics_dir as parameter. That said, metrics_config_path can still be hard-coded.

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Combined both into a single label

rhizome/postgres/bin/metrics-collector
shikharbhardwaj: These should always exist, just added these checks to catch misconfigs easier

byucesoy:
Just a note:
There are few customers with manually applied prometheus settings. We should be careful while deploying this change as incrementing `configure_metrics` would overwrite those.

PR #3211, opened 2025-04-29
Use golangci-lint 2.15 and golangci-lint-action v7

jeremyevans:
> [!IMPORTANT]
> Update `golangci-lint` to v2.1.5 and `golangci-lint-action` to v7 in `cli-ci.yml`.
> 
>   - **CI Configuration**:
>     - Update `golangci-lint-action` to `v7` in `cli-ci.yml`.
>     - Update `golangci-lint` version to `v2.1.5` in `cli-ci.yml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9fdb8c41e2f788a0e1c1ece2b41289a2dd6a77bd. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/ubi.go
+++ b/cli/ubi.go
@@ -71,9 +71,14 @@ func sendRequest(args []string) {
 		fmt.Fprintf(os.Stderr, "! Error sending http request\n")
 		os.Exit(1)
 	}
-	defer resp.Body.Close()
 
 	processResponse(resp, args)
+
+	err = resp.Body.Close()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "! Error closing response body\n")
+		os.Exit(1)
+	}
 }
 
 func main() {
--- a/.github/workflows/cli-ci.yml
+++ b/.github/workflows/cli-ci.yml
@@ -85,7 +85,7 @@ jobs:
       run: "cd cli && go fmt && git diff --stat --exit-code"
 
     - name: Run golangci-lint
-      uses: golangci/golangci-lint-action@v6
+      uses: golangci/golangci-lint-action@v7
       with:
-        version: v1.64
+        version: v2.1.5
         working-directory: cli


cli/ubi.go
```diff
@@ -71,9 +71,14 @@ func sendRequest(args []string) {
 		fmt.Fprintf(os.Stderr, "! Error sending http request\n")
 		os.Exit(1)
 	}
-	defer resp.Body.Close()
 
 	processResponse(resp, args)
+
+	err = resp.Body.Close()
```
enescakir, line 77: Should we defer to close in case it exits before reaching this point? Something like that

```golang
defer func() {
    err := resp.Body.Close()
    if err != nil {
        fmt.Fprintf(os.Stderr, "! Error closing response body\n")
 		os.Exit(1)
    }
}()
```

jeremyevans: There are no return statements that can cause early exits between the `defer` call and the end of the function, so I don't think the `defer` call is actually necessary.  That's why I moved it to the end of the function.  However, I'm certainly not a go expert.  Is there is a way for `processResponse` to cause an early exit of the function?

enescakir: It has some `os.Exit(1)` calls, but I haven't been writing Go for +5 years, so I'm not sure if it's an issue or not.

enescakir (approved):

cli/ubi.go
jeremyevans: `os.Exit` does not call deferred functions in Go, so it wouldn't have been called anyway: https://github.com/golang/go/issues/38261

Even if it would have been called, it wouldn't matter in this case, because process exit implicitly closes file descriptors.

PR #3217, opened 2025-04-30
Wait for vm scheduling before applying LoadBalancer logics

mohi-kalantari:
Logics like applying the nft rules or putting certificates in the Vm's VmHost were not checking whether the vm was scheduled to a VmHost or not. With this commit, we will wait for the Vm's VmHost before proceeding.

----

> [!IMPORTANT]
> Ensure VM operations in `cert_server.rb` and `update_load_balancer_node.rb` only proceed if VM is scheduled to a host, with corresponding tests added.
> 
>   - **Behavior**:
>     - In `cert_server.rb`, `put_cert_to_vm`, `put_certificate`, and `remove_cert_server` now check `vm.vm_host` before proceeding.
>     - In `update_load_balancer_node.rb`, `remove_load_balancer` checks `vm.vm_host` before executing commands.
>   - **Tests**:
>     - Added tests in `cert_server_spec.rb` for `reshare_certificate`, `put_certificate`, and `remove_cert_server` to ensure they handle unscheduled VMs correctly.
>     - Added tests in `update_load_balancer_node_spec.rb` for `remove_load_balancer` to verify behavior when VM is not scheduled.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for bd6c89caad6e266854874e9c275f5a106ea50903. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/cert_server.rb
+++ b/prog/vnet/cert_server.rb
@@ -12,13 +12,14 @@ class Prog::Vnet::CertServer < Prog::Base
   end
 
   label def reshare_certificate
-    put_cert_to_vm
+    put_cert_to_vm if vm.vm_host
 
     pop "certificate is reshared"
   end
 
   label def put_certificate
     nap 5 unless load_balancer.active_cert&.cert
+    nap 5 unless vm.vm_host
 
     put_cert_to_vm
     hop_start_certificate_server
@@ -30,6 +31,8 @@ class Prog::Vnet::CertServer < Prog::Base
   end
 
   label def remove_cert_server
+    pop "certificate resources and server are removed" unless vm.vm_host
+
     vm.vm_host.sshable.cmd("sudo host/bin/setup-cert-server stop_and_remove #{vm.inhost_name}")
     pop "certificate resources and server are removed"
   end
--- a/prog/vnet/update_load_balancer_node.rb
+++ b/prog/vnet/update_load_balancer_node.rb
@@ -25,6 +25,7 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
   end
 
   label def remove_load_balancer
+    pop "load balancer is removed" unless vm.vm_host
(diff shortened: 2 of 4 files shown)


prog/vnet/cert_server.rb
```diff
@@ -19,6 +20,7 @@ def vm
 
   label def put_certificate
     nap 5 unless load_balancer.active_cert&.cert
+    nap 5 if vm.vm_host.nil?
```
enescakir, line 23: nit
```suggestion
    nap 5 unless vm.vm_host
```

prog/vnet/cert_server.rb
```diff
@@ -30,6 +32,8 @@ def vm
   end
 
   label def remove_cert_server
+    pop "certificate resources and server are removed" if vm.vm_host.nil?
```
enescakir, line 35: nit
```suggestion
    pop "certificate resources and server are removed" unless vm.vm_host
```

prog/vnet/update_load_balancer_node.rb
```diff
@@ -25,6 +25,7 @@ def before_run
   end
 
   label def remove_load_balancer
+    pop "load balancer is removed" if vm.vm_host.nil?
```
enescakir, line 28: nit:
```suggestion
    pop "load balancer is removed" unless vm.vm_host
```

furkansahin (changes requested):
The problem seems like the change is made without thinking there might be VMs without load balancers but making use of certificates (I know, not yet, but it could be in future).
Also, if the vm is detached from a load_balancer, we bud this prog to stop the cert-server, your change makes it simply pop back out.

PR #3221, opened 2025-04-30
Selectable ch second part

fdr:
I just rebased #3180. I think this entire patch progression can be squashed.   But I wanted to save my reconciliation of rebase conflicts to spare others, and comment on the condensed patch.

----

> [!IMPORTANT]
> Refactor VM setup and teardown processes, update cloud hypervisor version handling, and improve test coverage.
> 
>   - **Refactoring**:
>     - Introduced `setup_vm_str(action)` in `nexus.rb` to generate VM setup command strings, replacing repeated command strings.
>     - Replaced `CloudHypervisor::Version::DEFAULT` and `CloudHypervisor::Firmware::DEFAULT` with dynamic version handling in `cloud_hypervisor.rb`.
>     - Updated `VmSetup` initialization to handle `ch_version` and `firmware_version` dynamically in `vm_setup.rb`.
>   - **Behavior Changes**:
>     - VM setup commands in `nexus.rb` now use `setup_vm_str(action)` for consistency.
>     - `prep_host.rb` now downloads cloud hypervisor and firmware using updated methods.
>     - `setup-vm` script now handles missing `prep.json` for delete actions gracefully.
>   - **Testing**:
>     - Updated tests in `cloud_hypervisor_spec.rb` and `vm_setup_spec.rb` to reflect changes in version handling and command generation.
>   - **Misc**:
>     - Removed redundant `DEFAULT` constants in `cloud_hypervisor.rb` and replaced with `SUPPORTED` and `INSTALLED` hashes.
>     - Minor updates to logging and error handling in various scripts.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for ad390ad7f4b2b4d218bcb52719f2fddd0b5f0f76. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -16,8 +16,19 @@ require_relative "storage_volume"
 class VmSetup
   Nic = Struct.new(:net6, :net4, :tap, :mac, :private_ipv4_gateway)
 
-  def initialize(vm_name)
+  def initialize(vm_name, hugepages: true, ch_version: nil, firmware_version: nil)
     @vm_name = vm_name
+    @hugepages = hugepages
+    @ch_version = CloudHypervisor::Version[ch_version] || no_valid_ch_version
+    @firmware_version = CloudHypervisor::Firmware[firmware_version] || no_valid_firmware_version
+  end
+
+  private def no_valid_ch_version
+    raise("no valid cloud hypervisor version")
+  end
+
+  private def no_valid_firmware_version
+    raise("no valid cloud hypervisor firmware version")
   end
 
   def q_vm
@@ -199,12 +210,14 @@ add element inet drop_unused_ip_packets allowed_ipv4_addresses { #{ip_net} }
   end
 
   def unmount_hugepages
+    return unless @hugepages
     r "umount #{vp.q_hugepages}"
   rescue CommandFail => ex
     raise unless /(no mount point specified)|(not mounted)|(No such file or directory)/.match?(ex.stderr)
   end
 
   def hugepages(mem_gib)
+    return unless @hugepages
     FileUtils.mkdir_p vp.hugepages
     FileUtils.chown @vm_name, @vm_name, vp.hugepages
     r "mount -t hugetlbfs -o uid=#{q_vm},size=#{mem_gib}G nodev #{vp.q_hugepages}"
@@ -653,11 +666,18 @@ DNSMASQ_SERVICE
 
     disk_params = storage_volumes.map { |volume|
       if volume.read_only
-        "--disk path=#{volume.image_path},readonly=on \\"
+        "path=#{volume.image_path},readonly=on"
       else
-        "--disk vhost_user=true,socket=#{volume.vhost_sock},num_queues=1,queue_size=256 \\"
+        "vhost_user=true,socket=#{volume.vhost_sock},num_queues=1,queue_size=256"
       end
     }
+    disk_params << "path=#{vp.cloudinit_img}"
+
+    disk_args = if Gem::Version.new(@ch_version.version) >= Gem::Version.new("36")
+      "--disk #{disk_params.join(" ")}"
+    else
+      disk_params.map { |x| "--disk #{x}" }.join(" ")
+    end
 
     spdk_services = storage_volumes.map { |volume| volume.spdk_service }.uniq
     spdk_after = spdk_services.map { |s| "After=#{s}" }.join("\n")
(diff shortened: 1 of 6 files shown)


prog/vm/nexus.rb
```diff
@@ -293,6 +297,18 @@ def before_run
...
 
+  def self.setup_vm_options_str(frame)
+    options = []
+    options << "hugepages=off" if frame["hugepages"] == false
+    options << "ch_version=#{frame["ch_version"]}" if frame["ch_version"]
+    options << "firmware_version=#{frame["firmware_version"]}" if frame["firmware_version"]
+    options.join(",")
+  end
+
+  def setup_vm_str(action)
```
fdr, line 308: I think it's a bit awkward to expect callers to run this special formatting method and then putting it into sshable.cmd, given it always targets the same sshable, perhaps the ssh access as well should be coupled as a helper (or a few helpers that may share an internal formatting routine) that runs sshable.cmd, e.g. `VmHost[ubid].setup_vm(...)` => dials the VM and does SSH stuff on its own.

prog/vm/nexus.rb
```diff
@@ -293,6 +297,18 @@ def before_run
     hop_wait_sshable
   end
 
+  def self.setup_vm_options_str(frame)
+    options = []
+    options << "hugepages=off" if frame["hugepages"] == false
+    options << "ch_version=#{frame["ch_version"]}" if frame["ch_version"]
+    options << "firmware_version=#{frame["firmware_version"]}" if frame["firmware_version"]
+    options.join(",")
+  end
```
fdr, line 306: I think per your commit message, it probably does make sense to put this in the entity record. It's not transient  either, it's the kind of thing we would want to report on.

prog/vm/update_ipv6.rb
```diff
@@ -22,7 +22,7 @@ def vm_host
     )
 
     write_params_json
-    vm_host.sshable.cmd("sudo host/bin/setup-vm reassign-ip6 #{vm.inhost_name}", stdin: JSON.generate({storage: vm.storage_secrets}))
+    vm_host.sshable.cmd("sudo host/bin/setup-vm reassign-ip6 #{vm.inhost_name} #{Prog::Vm::Nexus.setup_vm_options_str(vm.strand.stack.first)}", stdin: JSON.generate({storage: vm.storage_secrets}))
```
fdr, line 25: Maybe while pushing things into a helpers in a third space (such as VmHost, Vm, or some combination), this will de-awkardify this strange method class dependency between two progs.

fdr (commented):
@jeremyevans not sure if these ideas will pan out, I wonder if we can de-weirdify some stuff

fdr:
Another, distinct idea from inventing a new semi-standard protocol (the option-comma-list) with setup-vm: what about `params.json`?

rhizome/host/lib/vm_setup.rb
```diff
@@ -16,8 +16,19 @@
 class VmSetup
   Nic = Struct.new(:net6, :net4, :tap, :mac, :private_ipv4_gateway)
 
-  def initialize(vm_name)
+  def initialize(vm_name, hugepages: true, ch_version: nil, firmware_version: nil)
```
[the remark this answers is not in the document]
jeremyevans: No it doesn't, because Version/Firmware is a hash with a default value being the default version/firmware to use.  It will only raise an exception if there is no installed cloud-hypervisor version or firmware.

spec/prog/vm/update_ipv6_spec.rb
```diff
@@ -46,8 +46,9 @@
   it "rewrites persisted" do
     expect(pr).to receive(:vm_host).and_return(vm_host).at_least(:once)
     expect(vm).to receive(:update).with(ephemeral_net6: "2001::/64")
+    expect(vm).to receive(:strand).and_return(instance_double(Strand, stack: [{}]))
     expect(pr).to receive(:write_params_json)
-    expect(vm_host.sshable).to receive(:cmd).with("sudo host/bin/setup-vm reassign-ip6 test", stdin: JSON.generate({storage: "storage_secrets"}))
+    expect(vm_host.sshable).to receive(:cmd).with("sudo host/bin/setup-vm reassign-ip6 test ", stdin: JSON.generate({storage: "storage_secrets"}))
```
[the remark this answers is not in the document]
fdr: It is, it's in a test, so it's going to fail otherwise.

jeremyevans (commented):
This looks good.  I agree that we should move the hugepages/ch_version/firmware_version into the params json, before merging this. If it is in the params json, I'm not sure it's needed in Vm model storage unless we want to use it for reporting, because the host can always get the information from the stored json file.

We will need to change things so that the params json is always parsed, as VmSetup#purge_without_network at least needs to know whether hugepages are in use.

Putting this information in the params json file avoids the awkwardness of the class method and referencing the strand stack from from UpdateIpv6.

jeremyevans:
@fdr I pushed changes to put the hugepages/ch_version/firmware_version information in the params json.  This removes the awkward class method as well as needing to reach into the frame to get information.  It also fixes a potential issue with deleting params json before rewriting it, by writing new params json to a temp file, merging it with the existing params json file, and then doing an atomic rename, so at all points, the params json file on disk is good.

I tested with these changes and can still create and run both a 35.1 and 45.0 CH VM on the same host.

FWIW, I cannot see in the code where `Prog::Vm::UpdateIpv6` was ever used.  Maybe it was only run in `pry`?  Do we think it will be needed in the future?  If not, we could remove it. CC @furkansahin

fdr:
@jeremyevans refreshed this, you can see I made two edits in addendum, and for sake of familiarity of the cumulative diff I separate a refactoring that I'm not sure about in `Factor out setup-vm command to a method`, that I plan to omit.

prog/vm/nexus.rb
```diff
@@ -304,6 +304,10 @@ def before_run
     hop_wait_sshable
   end
 
+  def setup_vm_str(action)
```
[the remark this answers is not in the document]
fdr: :elaborate series of shrugs:

prog/vm/nexus.rb
jeremyevans: We could just drop this commit.  Now that it has no special behavior and isn't used by UpdateIpv6, there's not really an advantage for having a separate method.

jeremyevans (approved):

PR #3225, opened 2025-05-01
Update pg_hba.conf to perform local connections via scram-sha-256

shikharbhardwaj:
Update pg_hba.conf to allow pgbouncer connections via other users.
Also reorders some more specific rules to appear before the generic catch-all rule for local connections.


rhizome/postgres/bin/configure
```diff
@@ -48,25 +48,27 @@ pg_hba_entries = <<-PG_HBA
 # Database administrative login by Unix domain socket
 local   all             postgres                                peer map=system2postgres
 local   all             pgbouncer                               peer map=system2pgbouncer
+local   replication     all                                     peer
```
byucesoy, line 51: It is bit misleading, but `all` does not cover `replication`. So no need to move this.

Second keyword in each row refers to database, so `all` covers all databases, but there is no database named `replication`. It is seen as pseudo-database.

rhizome/postgres/bin/configure
```diff
@@ -48,25 +48,27 @@ pg_hba_entries = <<-PG_HBA
...
+# Allow connections from localhost with ubi_monitoring OS user as
+# ubi_monitoring database user. This will be used by postgres_exporter
+# to scrape metrics and expose them to prometheus.
+local   all             ubi_monitoring                          peer
 
 # "local" is for Unix domain socket connections only
-local   all             all                                     peer
+# Use SCRAM authentication for all local connections to allow pgbouncer
+# passthrough to work for all non-unix users
+local   all             all                                     scram-sha-256
```
byucesoy, line 61: It would be good to add a rule that allows peer authentication just for postgres user. Otherwise we would lose the ability to do psql directly while we are using postgres user.

byucesoy (approved):

rhizome/postgres/bin/configure
shikharbhardwaj: Would the 1st rule in the file work for that?

byucesoy: Ah, yes it would. All is good then. :+1:

byucesoy (approved):

PR #3230, opened 2025-05-01
Check all requests are appropriately authorized when running non-frozen specs

jeremyevans:
This adds a needs_authorization flag before the main hash branch dispatch.

When running non-frozen specs, if that flag is not cleared before a response
is returned, an exception is raised.  The flag can be cleared by:

* Calling an authorization method
* Calling no_authorization_needed to explicit indicate authorization is not needed

This adds no_authorization_needed to routes I believe do not need additional
authorization.

To find incorrect usage of no_authorization_needed, it also raises if the
request is already authorizated or does not require authorization.

This support found a missing postgres read-replica authorization (which is
fixed in the first commit), and should prevent missing authorization in the future.

This also fixes an issue where Clover was returning 204 responses for DELETE
requests where the path had garbage at the end.  This changes handling so that
204 is only used if the DELETE request is fully routed.

----

> [!IMPORTANT]
> Introduces authorization checks for all requests in `clover.rb` and fixes DELETE request handling for invalid paths.
> 
>   - **Authorization**:
>     - Introduces `@still_need_authorization` flag in `clover.rb` to ensure all requests are authorized or explicitly marked with `no_authorization_needed`.
>     - Raises error if `no_authorization_needed` is called when not needed or already authorized.
>     - Adds `no_authorization_needed` to routes in `clover.rb`, `account.rb`, `cli.rb`, `github.rb`, `project.rb`, and others.
>   - **Bug Fix**:
>     - Fixes issue in `general.rb` where DELETE requests returned 204 for invalid paths by using `check_found_object()`.
>   - **Testing**:
>     - Adds test in `clover_web_spec.rb` to ensure `no_authorization_needed` raises errors when misused.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 4e90227d715f576288e8671b0c06bba623edc69f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -692,7 +692,7 @@ class Clover < Roda
       end
     end
 
-    @need_authorization = true
+    @still_need_authorization = true
     r.hash_branches("")
   end
 
@@ -715,30 +715,30 @@ class Clover < Roda
     # is needed by calling no_authorization_needed
 
     after do |res|
-      if @need_authorization && res && res[0] != 404 && res[0] != 501
+      if @still_need_authorization && res && res[0] != 404 && res[0] != 501
         raise "no authorization check for #{request.request_method} #{request.path_info}"
       end
     end
 
     prepend(Module.new do
       def authorize(actions, object_id)
-        @need_authorization = false
+        @still_need_authorization = false
         super
       end
 
       def has_permission?(actions, object_id)
-        @need_authorization = false
+        @still_need_authorization = false
         super
       end
 
       def dataset_authorize(ds, actions)
-        @need_authorization = false
+        @still_need_authorization = false
         super
       end
 
       def no_authorization_needed
-        raise "called no_authorization_needed when authorization already not needed: #{request.inspect}" unless @need_authorization
-        @need_authorization = false
+        raise "called no_authorization_needed when authorization already not needed: #{request.inspect}" unless @still_need_authorization
+        @still_need_authorization = false
         super
       end
     end)
--- a/clover.rb
+++ b/clover.rb
@@ -572,6 +572,7 @@ class Clover < Roda
 
   hash_branch("after-login") do |r|
     r.get web? do
+      no_authorization_needed
       if (project = current_account.projects_dataset.order(:created_at, :name).first)
         r.redirect "#{project.path}/dashboard"
       else
@@ -588,6 +589,7 @@ class Clover < Roda
     end
 
     hash_branch("clear-last-password-entry") do |r|
+      no_authorization_needed
       session.delete("last_password_entry")
       ""
     end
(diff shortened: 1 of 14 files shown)


fdr (approved):
My one question mark is the naming of `need_authorization`, to more strenuously reference its temporal logic. For example, I could imagine being alarmed by printing the state of the app sometime *after* authorization and being alarmed that the (presumed immutable) `need_authorization` instance variable is false, i.e. maybe I would think, "did we skip all the authorization?!" (quite the opposite)

But I have no proposal and maybe this is just one of those things the reader needs to know.

jeremyevans:
> My one question mark is the naming of `need_authorization`, to more strenuously reference its temporal logic. For example, I could imagine being alarmed by printing the state of the app sometime _after_ authorization and being alarmed that the (presumed immutable) `need_authorization` instance variable is false, i.e. maybe I would think, "did we skip all the authorization?!" (quite the opposite)
> 
> But I have no proposal and maybe this is just one of those things the reader needs to know.

We could easily use two flags, `@need_authorization` and `@authorization_checked`,  set `@authorization_checked` if we did authorize, and unset `@need_authorization` for the `no_authorization_needed` case.  Then we would only error if `@need_authorization` is set and `@authorization_checked` is not.  Do you think it is worth switching to that?

fdr:
> > My one question mark is the naming of `need_authorization`, to more strenuously reference its temporal logic. For example, I could imagine being alarmed by printing the state of the app sometime _after_ authorization and being alarmed that the (presumed immutable) `need_authorization` instance variable is false, i.e. maybe I would think, "did we skip all the authorization?!" (quite the opposite)
> 
> > 
> 
> > But I have no proposal and maybe this is just one of those things the reader needs to know.
> 
> 
> 
> We could easily use two flags, `@need_authorization` and `@authorization_checked`,  set `@authorization_checked` if we did authorize, and unset `@need_authorization` for the `no_authorization_needed` case.  Then we would only error if `@need_authorization` is set and `@authorization_checked` is not.  Do you think it is worth switching to that?

Nah, two fields seems much worse than one. I was wondering if it could be named in some overwrought way since it's infrequently referenced.

jeremyevans:
> Nah, two fields seems much worse than one.

That was my initial feeling as well.

> I was wondering if it could be named in some overwrought way since it's infrequently referenced.

`@still_need_authorization` ?

fdr:
> > Nah, two fields seems much worse than one.
> 
> 
> 
> That was my initial feeling as well.
> 
> 
> 
> > I was wondering if it could be named in some overwrought way since it's infrequently referenced.
> 
> 
> 
> `@still_need_authorization` ?

I think that does the trick.

jeremyevans:
Pushed a commit to rename to `@still_need_authorization`.

fdr (approved):

PR #3232, opened 2025-05-01
Avoid defining Object#login

jeremyevans:
Define the login method only in the RSpec class, since it is only used in the specs.


fdr (approved):

fdr:
woops!

jeremyevans:
Actually, this moving this definition inside the RSpec.configure block does not have an effect.  I added  the following:

```ruby
def Object.method_added(method)
  if Object == self
    p [self, method, caller(1, 3)]
    super
  end
end
```

Using this, I can see this issue affects a number of methods, not just `login`.  I'll look into how to properly handle this, so the methods are only added to RSpec scope and not to Object.

fdr (approved):

PR #3237, opened 2025-05-02
Fix parallel vm creation issue

mohi-kalantari:
An unused semaphore is added to Nic Nexus which allows us to notify the
Subnet to start the rekeying logic through the Nic.

In the former version, Vm would directly call the Subnet Nexus to
start the rekey process but if an unlucky nic didn't get enough CPU
time to reach the wait_setup state, it would miss the start_rekey
semaphore and since when we decr a semaphore, it is set to zero,
Nic strand would get stuck on wait_setup until a manual trigger is done.

Now each nic is used as the bridge between Vm and Subnet and this way,
we make sure Nic strand is in the right state before subnet starts
triggering different semaphores and logics.

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -284,7 +284,7 @@ class Prog::Vm::Nexus < Prog::Base
   label def prep
     case host.sshable.cmd("common/bin/daemonizer --check prep_#{q_vm}")
     when "Succeeded"
-      vm.private_subnets.each(&:incr_add_new_nic)
+      vm.nics.each(&:incr_setup_nic)
       hop_clean_prep
     when "NotStarted", "Failed"
       secrets_json = JSON.generate({
--- a/prog/vnet/nic_nexus.rb
+++ b/prog/vnet/nic_nexus.rb
@@ -53,8 +53,13 @@ class Prog::Vnet::NicNexus < Prog::Base
 
   label def wait_setup
     decr_vm_allocated
+    when_setup_nic_set? do
+      DB.transaction do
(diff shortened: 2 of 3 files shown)


furkansahin:
The problem with that is, now, you don't distinguish the nics that you can start adding (because they are provisioned) or not. Consider that, we provision 3 VMs in parallel, 2 of them were able to get provisioned, 1 is not because of capacity, We essentially need to be able to setup the necessary stuff for the 2 and handle the rest when the 3rd VM comes. With this implementation, we will immediately start working on adding all these 3 Nics, we won't be able to create state in RekeyNicTunnel prog for 1 of the Nics that will block the whole operation for other 2 as well. There is a reason this commit is added :) https://github.com/ubicloud/ubicloud/commit/9ca6575a585e15c02ac44308dbc4d207ca6285de

mohi-kalantari:
you're right I'll revert it 😄 @furkansahin

mohi-kalantari:
I removed that commit and wrote it in another way. could you check please? @furkansahin

furkansahin (commented):
can we make use of the `setup_nic` semaphore?

mohi-kalantari:
There is no semaphore named `setup_nic`. I guess you mean the `add_new_nic` in subnet_nexus. 

Since we want the nic setup logic to go throug nic_nexus, I feel like we need to introduce a new semaphore. @furkansahin

furkansahin (approved):
Make sure this passes https://github.com/ubicloud/ubicloud/actions/runs/14969844791 and then 🚢

PR #3239, opened 2025-05-02
Add hugepage, cloud-hypervisor, and firmware fields to params.json

fdr:
These fields are not yet read by `rhizome`.

The firmware and cloud-hypervisor version fields allow control over the versions used at provisioning time, enabling gradual rollout and testing of new versions and combinations. This is the main motivation. `nil` values for all of these are acceptable, a near-future version of `rhizome` will apply a default in that case.

The hugepage field is for skipping VM allocation using hugepages, to speed up some aspects of development iteration: hugepages cost an extra reboot in `prep_host` and slows down every reboot as Linux searches for contiguous memory.  Alone, this does not complete the objective, but it's also necessary to move towards it.

Commit message by Daniel Farina <daniel@ubicloud.com>

----

> [!IMPORTANT]
> Add `ch_version`, `firmware_version`, and `hugepages` fields to VM parameters and update related methods and tests.
> 
>   - **Behavior**:
>     - Add `ch_version`, `firmware_version`, and `hugepages` fields to `params_json` in `model/vm.rb`.
>     - Update `assemble` method in `prog/vm/nexus.rb` to accept `ch_version`, `firmware_version`, and `hugepages` parameters.
>     - Modify `write_params_json` in `prog/vm/nexus.rb` to include new fields in JSON.
>   - **Tests**:
>     - Update tests in `nexus_spec.rb` to verify `params_json` includes `ch_version`, `firmware_version`, and `hugepages` fields.
>     - Add test cases for `assemble` method to handle new parameters.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 100e66b696b87c7269c4604d365481200015b4a9. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -15,7 +15,8 @@ class Prog::Vm::Nexus < Prog::Base
     unix_user: "ubi", location_id: Location::HETZNER_FSN1_ID, boot_image: Config.default_boot_image_name,
     private_subnet_id: nil, nic_id: nil, storage_volumes: nil, boot_disk_index: 0,
     enable_ip4: false, pool_id: nil, arch: "x64", swap_size_bytes: nil,
-    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [], gpu_count: 0)
+    distinct_storage_devices: false, force_host_id: nil, exclude_host_ids: [], gpu_count: 0,
+    hugepages: true, ch_version: nil, firmware_version: nil)
 
     unless (project = Project[project_id])
       fail "No existing project"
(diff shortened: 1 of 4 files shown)


prog/vm/nexus.rb
```diff
@@ -301,7 +305,8 @@ def before_run
   end
 
   def write_params_json
-    host.sshable.cmd("sudo -u #{q_vm} tee #{params_path.shellescape}", stdin: vm.params_json(frame["swap_size_bytes"]))
+    host.sshable.cmd("sudo -u #{q_vm} tee #{params_path.shellescape}",
+      stdin: vm.params_json(**frame.slice("swap_size_bytes", "hugepages", "ch_version", "firmware_version").transform_keys(&:to_sym)))
```
jeremyevans, line 309: Could use `transform_keys!` to save an allocation.

jeremyevans (approved):
Changes look fine.  Might be good to put it through an E2E test before merging.

prog/vm/nexus.rb
fdr: applied

fdr:
> Changes look fine. Might be good to put it through an E2E test before merging.

also did that, launching shortly....

PR #3246, opened 2025-05-05
Use the Roda plain_hash_response_headers plugin

jeremyevans:
Rack 3 requires that response headers use lower case keys. To support applications designed for Rack < 3, Roda by default uses Rack::Headers for response headers, which will automatically lower case keys. This is suboptimal for performance.  The plain_hash_response_headers plugin allows using a plain hash for response headers, but in order to use it, you must ensure that all response header keys are already lower case.  So change all response header keys to lower case.

Unfortunately, committee is not rack 3 compliant and does not support lower case response header keys.  It indexes into response headers with Content-Type instead of content-type, which breaks on rack 3 unless the headers hash is automatically lower casing keys. Work around this issue by setting the Content-Type header before calling committee's response validation, and deleting it afterward. This can be removed after committee is fixed.

CC @geemus 

----

> [!IMPORTANT]
> Use `plain_hash_response_headers` plugin in Roda, lowercase all response headers, and update `committee` and `rodauth` gem references.
> 
>   - **Behavior**:
>     - Use `plain_hash_response_headers` plugin in `clover.rb` to improve performance by using plain hash for response headers.
>     - Change all response header keys to lowercase in `clover.rb`, `routes/project/billing.rb`, `routes/project/location/kubernetes_cluster.rb`, `routes/project/location/postgres.rb`, and `routes/webhook/github.rb`.
>     - Workaround for `committee` gem's non-compliance with Rack 3 by setting `Content-Type` header before and deleting it after response validation.
>   - **Gemfile**:
>     - Update `committee` to specific commit `fc9329c7e7eebb19c46114e6dcee6e292bb6cf4c`.
>     - Update `rodauth` to specific commit `6b853d8eabfbebe8b6036c55282fe0aae09a90ab`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 583b6a21642186e001b336394ca72448b6188a99. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


enescakir (approved):

jeremyevans:
The committee fix submitted by @geemus seems to work, but I found there are issues in other libraries used (at least Rodauth also needs fixes).   I'll work on that, probably next week.

geemus:
@jeremyevans Thanks for testing/confirming, I can work on a release soon (possibly also next week).

PR #3248, opened 2025-05-05
Update bin/monitor to export metrics to VictoriaMetrics

shikharbhardwaj:
## Description
This commit updates bin/monitor to scan for resources which have metrics
export capability (currently only PG servers) and periodically export
metrics out store them in VictoriaMetrics.

Each resource which exports metrics provides `MetricsTargetMethods`,
which contain the config and a commmon algorithm to fetch scraped data
from a given list of endpoints. Roghly the logic looks like this:

Producer side (PG server):
A systemd-backed timer executes `postgres/bin/metrics-collector`, which
scrapes the list of endpoints and stores them on disk, every 15s
(default interval). This location is the `pending` buffer, which can
grow upto a 120 entries by default, providing recovery for 30
minutes of outages in other parts of the system.

Consumer side (bin/monitor):
Monitor periodically connects to each metrics target resource to fetch
upto 4 scrapes at a time and write them to VictoriaMetrics, and move the
successfully exported scrapes to the `done` buffer, which is cleand up
periodically by the `metrics-collector`.  Currently, we only create a
single global VictoriaMetrics instance for simplicity, although it can
be configured to be location-aware.

----

> [!IMPORTANT]
> Add functionality to export metrics from Postgres servers to VictoriaMetrics, including new classes, methods, and tests.
> 
>   - **Behavior**:
>     - `bin/monitor` updated to export metrics from `PostgresServer` to VictoriaMetrics using `MetricsTargetMethods`.
>     - New `MetricsTargetResource` class handles metric export sessions and manages resource state.
>     - `VictoriaMetrics::Client` updated with `import_prometheus()` to send gzipped data.
>   - **Configuration**:
>     - `config.rb` updated with `max_metrics_export_threads` and `victoria_metrics_service_project_id`.
>   - **Models**:
>     - `PostgresServer` includes `MetricsTargetMethods` for metric configuration and session initiation.
>   - **Tests**:
>     - New tests in `metrics_target_methods_spec.rb`, `metrics_target_resource_spec.rb`, and `client_spec.rb` for metric export functionality.
>     - `postgres_server_spec.rb` updated to test metric session initiation.
>   - **Misc**:
>     - `metrics_target_methods.rb` and `metrics_target_resource.rb` added for metric export logic.
>     - `client.rb` updated to handle gzipped data and authentication for VictoriaMetrics.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 3832f833f772bcdf0ce3889b54094b522035a80a. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bin/monitor
+++ b/bin/monitor
@@ -4,19 +4,32 @@
 require_relative "../loader"
 clover_freeze
 
-resources = {}
-mutex = Mutex.new
-thread_pool_size = (Config.max_monitor_threads - 2).clamp(1, nil)
+health_monitor_resources = {}
+health_monitor_mutex = Mutex.new
+health_monitor_thread_pool_size = (Config.max_health_monitor_threads - 2).clamp(1, nil)
 monitorable_resource_types = [VmHost, PostgresServer, Vm.where(~Sshable.where(id: Sequel[:vm][:id]).exists), MinioServer, GithubRunner, VmHostSlice, LoadBalancerVmPort, KubernetesCluster, VictoriaMetricsServer]
-queue_size = thread_pool_size + (monitorable_resource_types.sum(&:count) * 1.5).round
-queue = SizedQueue.new(queue_size)
+health_monitor_queue_size = health_monitor_thread_pool_size + (monitorable_resource_types.sum(&:count) * 1.5).round
+health_monitor_queue = SizedQueue.new(health_monitor_queue_size)
+
+metrics_target_resources = {}
+metrics_export_mutex = Mutex.new
+metrics_export_thread_pool_size = (Config.max_metrics_export_threads - 2).clamp(1, nil)
+metrics_target_resource_types = [PostgresServer]
+metrics_export_queue_size = metrics_export_thread_pool_size + (metrics_target_resource_types.sum(&:count) * 1.5).round
+metrics_export_queue = SizedQueue.new(metrics_export_queue_size)
 
 resource_scanner = Thread.new do
   loop do
     monitorable_resources = monitorable_resource_types.flat_map(&:all)
-    mutex.synchronize do
+    health_monitor_mutex.synchronize do
       monitorable_resources.each do |r|
-        resources[r.id] ||= MonitorableResource.new(r)
+        health_monitor_resources[r.id] ||= MonitorableResource.new(r)
+      end
+    end
+    metrics_resources = metrics_target_resource_types.flat_map(&:all)
+    metrics_export_mutex.synchronize do
+      metrics_resources.each do |r|
+        metrics_target_resources[r.id] ||= MetricsTargetResource.new(r)
       end
     end
 
@@ -28,9 +41,9 @@ rescue => ex
   Kernel.exit!
 end
 
-thread_pool = Array.new(thread_pool_size) do
+health_monitor_thread_pool = Array.new(health_monitor_thread_pool_size) do
   Thread.new do
-    while (r = queue.pop)
+    while (r = health_monitor_queue.pop)
       r.lock_no_wait do
         r.open_resource_session
         r.process_event_loop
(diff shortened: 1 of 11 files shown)


model/postgres/postgres_server.rb
```diff
@@ -231,6 +243,10 @@ def health_monitor_socket_path
     @health_monitor_socket_path ||= File.join(Dir.pwd, "var", "health_monitor_sockets", "pg_#{vm.ip6}")
   end
 
+  def metrics_export_socket_path
+    @health_monitor_socket_path ||= File.join(Dir.pwd, "var", "metrics_export_sockets", "pg_#{vm.ip6}")
```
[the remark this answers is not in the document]
shikharbhardwaj: Nice catch

rhizome/victoria_metrics/bin/restart
```diff
@@ -0,0 +1,4 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+r "systemctl restart victoria_metrics && systemctl restart vmauth"
```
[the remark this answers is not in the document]
shikharbhardwaj: Good catch, although the suggestion is not in the right direction

rhizome/victoria_metrics/bin/restart
```diff
@@ -0,0 +1,4 @@
+#!/bin/env ruby
```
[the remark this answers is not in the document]
shikharbhardwaj: Looks like `/bin/env` is the general convention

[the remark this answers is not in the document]
jeremyevans: Looks like outside of rhizome, we always use `/usr/bin/env`.  Inside rhizome, we are using a mix or `/bin/env` and `/usr/bin/env`, but we should probably consistently use `/usr/bin/env` there as well.  On Ubuntu, in appears that `/bin/env` and `/usr/bin/env` are hard links to the same thing, and I'm not sure rhizome is currently running on non-Ubuntu, but non-Linux may not have `/bin/env`.  Since `/usr/bin/env` is more universal, it's best to standardize on that.

shikharbhardwaj:
@jeremyevans that does not seem to work here, because I based the other 2 PRs off `main` and rebased this one off both
The other two are ready to merge anyways, so I think I'll avoid rebasing more branches for now

rhizome/victoria_metrics/bin/restart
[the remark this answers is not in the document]
shikharbhardwaj: :ack:
Agreed on making this uniform across the board, I will switch these to use `/usr/bin/env` in respective branches

bin/monitor
shikharbhardwaj, line 1: jw: Should we split the export responsibilities to a different daemon?

rhizome/victoria_metrics/bin/restart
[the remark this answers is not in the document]
shikharbhardwaj: Updated this to use `/usr/bin/env`

model/postgres/postgres_server.rb
```diff
@@ -193,6 +194,16 @@ def init_health_monitor_session
     }
   end
 
+  def init_metrics_export_session
+    FileUtils.rm_rf(metrics_export_socket_path)
```
byucesoy, line 198: Do we ever use these sockets?

lib/metrics_target_methods.rb
```diff
@@ -0,0 +1,65 @@
...
+    scrape_files = session[:ssh_session].exec!("ls #{metrics_dir}/pending | sort | head -n #{MAX_SCRAPE_FETCH_COUNT}").split("\n")
+
+    scrape_files.filter_map do |file|
+      time_str = file.split(".").first
+      time = Time.strptime(time_str, FILENAME_FORMAT)
+      status = {}
+
+      scrape_content = session[:ssh_session].exec!("cat #{metrics_dir}/pending/#{file.shellescape}", status: status)
+
+      VictoriaMetrics::Client::Scrape.new(time: time, samples: scrape_content) if status[:exit_code] == 0
```
byucesoy, line 53: This class doesn't seem to add much value. You can use hash instead.

```
{time:, samples: scrape_content} if status[:exit_code] == 0
```

You can also omit repeating the second field by renaming it to `samples` in previous line.

lib/metrics_target_resource.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
```
byucesoy, line 1: Is there anything we need to be careful about persistence of this? We don't persist anything about this class (as opposed to our models), which is fine, we don't persist anything for `MonitorableResource` as well. Just wanted to be sure that restarts (which happens multiple times a day after each deployment) is OK from the perspective of how this operates.

byucesoy (approved):

lib/metrics_target_resource.rb
shikharbhardwaj: It should be okay, we should not miss out on any data if the restarts happen within ~30 mins.
We can have retransmit for the same data in some cases (eg. the data was written but not moved to `done`). In this case Victoriametrics will dedup samples within a configured `-dedup.minScrapeInterval`, which we can set set to match the scrape interval 15s.

model/postgres/postgres_server.rb
shikharbhardwaj: Yeah these are unused, I'll remove them

lib/metrics_target_methods.rb
shikharbhardwaj: I changed this to use the `Data` core class

lib/metrics_target_resource.rb
shikharbhardwaj: I just noticed that this parameter does not have a default, added it in #3267

PR #3249, opened 2025-05-05
Change root redirect to default project dashboad redirect

jeremyevans:
Without this, if a user is already logged in, it redirects to /login, which redirects to /after-login, which redirects to the same place.


clover.rb
```diff
@@ -573,11 +573,7 @@ def assets
   hash_branch("after-login") do |r|
     r.get web? do
       no_authorization_needed
-      if (project = current_account.projects_dataset.order(:created_at, :name).first)
-        r.redirect "#{project.path}/dashboard"
-      else
-        r.redirect "/project"
-      end
+      default_project_dashboard_redirect
```
enescakir, line 576: It sounds like returning a value.
```suggestion
      redirect_default_project_dashboard
```
OR
```suggestion
      request.redirect default_project_dashboard
```

jeremyevans: Pushed a commit to switch to `redirect_default_project_dashboard`.

enescakir (approved):

PR #3254, opened 2025-05-06
Allow customers to select the Kubernetes version in UI

mohi-kalantari:
New radio_small_cards is added to kubernetes create form to specify the kubernetes version.

Currently we plan to only show v1.32 and v1.33 versions.


aamederen:
We need changes in "download_boot_image.rb", so maybe that change can be a pre-requisite of this PR

mohi-kalantari:
yes those changes will be in another PR after I test the images locally first. @aamederen

lib/option.rb
```diff
@@ -22,6 +22,10 @@ def self.kubernetes_locations
     Location.where(name: ["hetzner-fsn1", "leaseweb-wdc02"]).all
   end
 
+  def self.kubernetes_versions
+    ["v1.32", "v1.33"].freeze
```
aamederen, line 26: this would make v1.32 the default right? If so, what do you think about making v1.33 the default by reversing the order?

aamederen (approved):

lib/option.rb
mohi-kalantari: since it's a required option now in the backend and the customers have to select one, I guess there is no defautling and it doesn't matter which one is the first element.

PR #3255, opened 2025-05-06
Introduce a new "performance" vm family

enescakir:
- **Add migration to add family column to vm host**
  We plan to introduce a new virtual machine family (performance) for
  premium CPU runners. Currently, the allocator uses all hosts without
  distinguishing between families. We have different host types with
  various CPU models. I added a "family" column to allow the selection of
  specific hosts for certain VM families. We also have a "standard-gpu" VM
  family, so we can manually change the family of GPU runner hosts and use
  the family to filter.
  

- **Pass the vm host family while assembling it**
  

- **Introduce a new "performance" vm family**
  We're working on a new runner tier that has higher CPU performance. I've
  introduced a new vm family to use with this tier. The VMs with better
  CPU performance belong to this family.
  
  I added a billing rate only for the "github-runners" location since we
  don't introduce performance VMs publicly that are only used internally
  for premium runners. For the VmVCpu billing rate, I simply doubled the
  standard price. Since we don't create billing records for internal
  projects, the value is not important at all for the runner location. We
  can adjust the value for other locations when we decide to make the new
  family available to other services.
  

- **Add a family filter to the VM host allocator**
  It selects only VM hosts for the specified families. If the premium CPU
  runners are enabled, it tries to allocate for both standard and
  performance families, giving higher priority to the performance family
  if there is capacity. Otherwise, it falls back to the standard family.
  
  Since we use the same hosts for standard and burstable virtual machines,
  we pass it as "standard" for "burstable" instances.


prog/vm/nexus.rb
```diff
@@ -202,25 +202,29 @@ def before_run
...
           end
-          [["accepting"], runner_location_filter, runner_location_preference, []]
+          [["accepting"], runner_location_filter, runner_location_preference, [], runner_family_filter]
         else
-          [["accepting"], [vm.location_id], [], []]
+          [["accepting"], [vm.location_id], [], [], [vm.family]]
         end
-
-      prefer_performance = runner.installation.project.get_ff_performance_runners if runner
+      family_filter = ["standard"] if vm.family == "burstable"
```
bsatzger, line 227: for hosts, we won't have `burstable`?

prog/vm/nexus.rb
```diff
@@ -202,25 +202,29 @@ def before_run
       host_exclusion_filter = frame["exclude_host_ids"] || []
       gpu_count = frame["gpu_count"] || 0
       runner = GithubRunner.first(vm_id: vm.id) if vm.location_id == Location::GITHUB_RUNNERS_ID
-      allocation_state_filter, location_filter, location_preference, host_filter =
+      allocation_state_filter, location_filter, location_preference, host_filter, family_filter =
```
bsatzger, line 205: unrelated to this PR, but I'm wondering if we should rename `host_filter` to something like `id_filter`.

bsatzger (approved):

prog/vm/nexus.rb
enescakir: Makes sense. I might change it at https://github.com/ubicloud/ubicloud/pull/3281. 

It also applies to host id exclusion. May be `id_exclusion_filter`

prog/vm/nexus.rb
enescakir: Since we use the same hosts for `standard` and `burstable` virtual machines with different slicing values, I don't think we will have any hosts reserved for the `burstable` family in near future.

migrate/20250429_add_vm_host_family.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    alter_table(:vm_host) do
+      add_column :family, String, collate: '"C"'
+    end
+
+    run "UPDATE vm_host SET family = 'standard'"
```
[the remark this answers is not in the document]
enescakir: I was torn between these two options, but the safer one won out. Creating as `standard` unknowingly seemed like an easy mistake to make by operators.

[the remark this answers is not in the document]
jeremyevans: I haven't tested but, but it may be simpler to use `default: "standard", null: false` in the `add_column`, and then do `set_column_default :family, nil` in the `same alter_table` block.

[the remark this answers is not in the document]
enescakir:
```ruby
      add_column :family, String, collate: '"C"', default: "standard", null: false
      set_column_default :family, nil
```

This tip worked. I updated it as is. Since it’s irreversible, I kept the `down` block.

PR #3256, opened 2025-05-06
[PremiumCPU #2] Unify GitHub runner allocator logic

enescakir:
- **Add migration for github installation allocator preferences**
  

- **Unify GitHub runner allocator logic**
  When we allocate a VM for a runner, we need to consider different
  scenarios, such as allowing US for standard-60 runners or allowing US
  runners for specific customers. We currently have separate feature flags
  for each scenario, which were added for specific customers. Instead of
  adding feature flags for every situation, we can maintain allocator
  preferences. This approach offers more flexibility and is easier to
  manage, allowing us to tailor it for these specific customers.
  
  For now, we'll skip the VM pool if performance runners are preferred.
  We'll implement the pool for them later.
  
  I have two other improvements in mind, but I will address them later:
  -  To eliminate the allocator branching in Vm::Nexus, I will move the
     GitHub related parts to the GithubRunner prog and the force_id logic
     to the allocator. This will significantly clean up the code.
  -  We also have the same allocator preferences in the VM stack as a
     separate property. With each feature, the number of properties will
     increase. We can integrate allocator preferences as a property of the
     VM stack, making it easier to manage and maintain. Then we can simply
     pass allocator_preferences hash to the Allocator, allowing it to
     override the default preferences. I believe this will also streamline
     the code, especially the tests, since we need to test each branch
     separately.
  

----

> [!IMPORTANT]
> Unifies GitHub runner allocator logic by introducing allocator preferences and renaming `performance` to `premium`.
> 
>   - **Behavior**:
>     - Unifies GitHub runner allocator logic by introducing `allocator_preferences` in `prog/vm/github_runner.rb` and `prog/vm/nexus.rb`.
>     - Adds migration `20250430_github_allocator_preference.rb` to add `allocator_preferences` column to `github_installation` table.
>     - Skips VM pool if `premium` runners are preferred.
>   - **Renames**:
>     - Renames `performance` to `premium` in `lib/option.rb`, `config/billing_rates.yml`, and `scheduling/allocator.rb`.
>   - **Tests**:
>     - Updates tests in `github_runner_spec.rb`, `nexus_spec.rb`, and `allocator_spec.rb` to reflect allocator logic changes and renaming.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c48bd030b8b036d451da57b25bfa9b6ccfc26eec. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


prog/vm/nexus.rb
```diff
@@ -207,20 +207,20 @@ def before_run
...
-          if runner
-            if (preferred_locations = runner.installation.project.get_ff_preferred_runner_locations) && !preferred_locations.empty?
-              # Just be sure your preferred ones are also in the location filter
-              runner_location_filter |= preferred_locations
-              runner_location_preference = preferred_locations
-            end
-            if vm.family == "standard" && runner.installation.project.get_ff_performance_runners
-              runner_family_filter = [vm.family, "performance"]
-            end
+          prefs = if runner
```
bsatzger, line 212: could also be written as:
```
prefs = runner ? runner.installation.allocator_preferences : {}
```

bsatzger (approved):

prog/vm/github_runner.rb
```diff
@@ -24,15 +24,14 @@ def self.assemble(installation, repository_name:, label:, default_branch: nil)
   end
 
   def pick_vm
-    skip_sync = true
-    prefer_performance = github_runner.installation.project.get_ff_performance_runners
+    prefer_performance = (github_runner.installation.allocator_preferences["family_filter"] || []).include?("performance")
```
jeremyevans, line 27:
```suggestion
    prefer_performance = github_runner.installation.allocator_preferences["family_filter"]&.include?("performance")
```

Since this is only used in boolean context, the `nil`/`false` distinction doesn't matter here.

prog/vm/nexus.rb
```diff
@@ -207,20 +207,20 @@ def before_run
...
+            runner.installation.allocator_preferences
+          else
+            {}
           end
```
jeremyevans, line 216:
```suggestion
          prefs = runner&.installation&.allocator_preferences || {}
```

Another possible approach.

jeremyevans (approved):

model/github_installation.rb
```diff
@@ -23,13 +23,14 @@ def total_active_runner_vcpus
...
-#  name              | text    | NOT NULL
-#  type              | text    | NOT NULL
-#  project_id        | uuid    |
-#  cache_enabled     | boolean | NOT NULL DEFAULT true
-#  use_docker_mirror | boolean | NOT NULL DEFAULT false
+#  id                    | uuid    | PRIMARY KEY
+#  installation_id       | bigint  | NOT NULL
+#  name                  | text    | NOT NULL
+#  type                  | text    | NOT NULL
+#  project_id            | uuid    |
```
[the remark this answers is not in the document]
enescakir: As far as I remember, we don't have any specific reason. I think I just forgot to add `null: false` in the initial migration.
https://github.com/ubicloud/ubicloud/pull/555

prog/vm/github_runner.rb
enescakir: I changed it and added a new test for branch coverage as well.

prog/vm/nexus.rb
enescakir: I remembered while rebasing why I chose the `if/else` block. I added more code to the if block in the PR. https://github.com/ubicloud/ubicloud/pull/3268/commits/5a9593bc0016c78daf19daba32f241e0972183d4

prog/vm/nexus.rb
enescakir: I remembered while rebasing why I chose the `if/else` block. I added more code to the if block in the PR. https://github.com/ubicloud/ubicloud/pull/3268/commits/5a9593bc0016c78daf19daba32f241e0972183d4

PR #3257, opened 2025-05-06
Fix required="true" attributes

jeremyevans:
Add a custom validator looking for {required,checked}={true,false} attributes. This found numerous cases.

Add an html_attrs method that handles required/checked attributes properly, skipping them if the value is false, and using them without a value if the value is true.

Change all attribute handling to use html_attrs to fix the issue.

### DIFF
--- a/helpers/web.rb
+++ b/helpers/web.rb
@@ -120,6 +120,24 @@ class Clover < Roda
     end
   end
 
+  def html_attrs(attributes)
+    attributes.map do |key, value|
+      case key
+      when :required, :checked
+        case value
+        when true
+          key.name
+        when false
+          ""
+        else
+          "#{key.name}=\"#{h(value)}\""
+        end
+      else
+        "#{h(key)}=\"#{h(value)}\""
+      end
+    end.join(" ")
+  end
+
   def object_tag_membership_label(obj)
     case obj
     when ObjectTag
--- a/views/components/form/resource_creation_form.erb
+++ b/views/components/form/resource_creation_form.erb
@@ -64,20 +64,19 @@ let option_dirty = <%==JSON.generate(form_elements.map { it[:name] }.each_with_o
 
       opt_attrs = {}
       opt_attrs[:disabled] = "disabled" unless parents_selected
-      opt_attrs[:checked] = "checked" if checked
+      opt_attrs[:checked] = checked
 
       [opt_val, opt_text, opt_classes, opt_attrs]
     end if has_option
 
     if has_option && selected_options[name].nil?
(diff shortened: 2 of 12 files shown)


helpers/web.rb
```diff
@@ -120,6 +120,24 @@ def ace_label(obj)
     end
   end
 
+  def html_attrs(attributes)
+    attributes.map do |key, value|
+      case key
+      when :required, :checked
+        case value
+        when true
+          key.name
```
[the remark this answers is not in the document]
jeremyevans: Symbol#name has been defined since Ruby 3.0.

helpers/web.rb
```diff
@@ -120,6 +120,24 @@ def ace_label(obj)
...
+        when false
+          ""
+        else
```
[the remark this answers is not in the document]
jeremyevans: Not sure we have any cases with a nil value, but this seems reasonable.

views/components/form/resource_creation_form.erb
```diff
@@ -64,20 +64,19 @@ let option_dirty = <%==JSON.generate(form_elements.map { it[:name] }.each_with_o
...
       options.find{ |_, _, _, opt_attrs| opt_attrs[:disabled].nil? }&.tap do |opt_val, _, _, opt_attrs|
-        opt_attrs[:checked] = "checked"
+        opt_attrs[:checked] = true
         selected_options[name] = opt_val
       end
     end
 
-    attributes = {}
-    attributes[:required] = "" if required
+    attributes = {required:}
```
byucesoy, line 79: Do we need to add

```
attributes[:required] = true if required
```

, so that `html_attrs` wouldn't ignore this attribute?

byucesoy, line 79: Do we need to add
```
attributes[:required] = true if required
```

, so that html_attrs wouldn't ignore this attribute?

byucesoy (approved):

views/components/form/resource_creation_form.erb
jeremyevans: `attributes = {required:}` is short for `attributes = {required: required}` (hash value omission).  So this uses the value of the `required` variable.  If `required` is true, this would be `attributes = {required: true}`. Similar if required is `false` or `nil`.  One case where this has different behavior is if `required` was `"true"`.  However, I don't think we have those cases (if we did, the custom HTML validation added would catch it.

byucesoy: Ah yes. I forgot about hash value omission :+1:

PR #3260, opened 2025-05-06
WIP: SPDK Replacement Integration

pykello:


rhizome/host/lib/storage_volume.rb
```diff
@@ -67,9 +83,124 @@ def prep(key_wrapping_secrets)
...
+        After=network.target
+
+        [Service]
+        Slice=#{@slice}
+        Environment=RUST_LOG=info
+        ExecStart=#{vhost_backend.bin_path} --config #{vhost_backend_config_path} #{kek_arg}
+        Restart=always
+        User=#{@vm_name}
+        Group=#{@vm_name}
+
```
fdr: I wonder if we can slim down the system capabilities here (e.g. the ability to set the system clock). Maybe check `systemd-analyze`

migrate/20250506_vhost_block_backend.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:vhost_block_backend) do
+      column :id, :uuid, primary_key: true
+      column :created_at, :timestamptz, null: false, default: Sequel::CURRENT_TIMESTAMP
+      column :version, :text, null: false
+      column :allocation_weight, :Integer, null: false
```
jeremyevans:
```suggestion
      column :allocation_weight, :integer, null: false
```

The current code works fine, but seems weird to capitalize integer.  You could use `Integer` instead of `:Integer`, but since you aren't using generic types for other columns in the migration, it doesn't make sense to use it here.

model/vhost_block_backend.rb
```diff
@@ -0,0 +1,27 @@
...
+
+require_relative "../model"
+
+class VhostBlockBackend < Sequel::Model
+  many_to_one :vm_host
+  one_to_many :vm_storage_volumes
+
+  def self.generate_uuid
+    UBID.generate(UBID::TYPE_ETC).to_uuid
+  end
```
jeremyevans:
```suggestion
  plugin ResourceMethods, etc_type: true
```

model/vm_host.rb
```diff
@@ -13,6 +13,7 @@ class VmHost < Sequel::Model
   one_to_one :provider, key: :id, class: :HostProvider
   one_to_many :assigned_host_addresses, key: :host_id, class: :AssignedHostAddress
   one_to_many :spdk_installations, key: :vm_host_id
+  one_to_many :vhost_block_backends, key: :vm_host_id
```
jeremyevans: Note that `, key: :vm_host_id` is not necessary, as that is the default.  I can see why it was used, though.

prog/storage/setup_vhost_block_backend.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+class Prog::Storage::SetupVhostBlockBackend < Prog::Base
+  subject_is :sshable, :vm_host
+
+  SUPPORTED_VHOST_BLOCK_BACKEND_VERSIONS = [
+    ["v0.1-5", "x64"],
+    ["v0.1-5", "arm64"]
+  ]
```
jeremyevans:
```suggestion
  ].freeze.each(&:freeze)
```

It's best if constants are frozen.

prog/storage/setup_vhost_block_backend.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def self.assemble(vm_host_id, version, allocation_weight: 0)
+    Strand.create_with_id(
```
jeremyevans:
```suggestion
    Strand.create(
```

No need to use `create_with_id` in new code.  We'll eventually convert all remaining `create_with_id` to `create`.

prog/storage/setup_vhost_block_backend.rb
```diff
@@ -0,0 +1,53 @@
...
+  label def start
+    version = frame["version"]
+    arch = vm_host.arch
+    fail "Unsupported version: #{version}, #{arch}" unless SUPPORTED_VHOST_BLOCK_BACKEND_VERSIONS.include? [version, arch]
+
+    VhostBlockBackend.create(
+      version: frame["version"],
+      allocation_weight: 0,
+      vm_host_id: vm_host.id
+    ) { it.id = VhostBlockBackend.generate_uuid }
```
jeremyevans:
```suggestion
    )
```

The ResourceMethods plugin should take care of this.

rhizome/host/spec/storage_volume_spec.rb
```diff
@@ -46,6 +75,7 @@
       vol = described_class.new("test", {"disk_index" => 1, "encrypted" => false})
       expect(File).to receive(:exist?).with("/var/storage").and_return(true)
       expect(FileUtils).to receive(:mkdir_p).with("/var/storage/test/1")
+      expect(FileUtils).to receive(:chown).with("test", "test", "/var/storage/test/1")
```
fdr: while chown/chmod can't be easily run for real and need stubbing out, I wonder if it's worth arranging things so other file system manipulations are completed within a directory and can be run for real on a unix system.

This would impose on the code under test, to prefix its various reads/writes/mkdirs. But, it would reduce brittleness. @jeremyevans, thoughts on balance of benefits there?

prog/storage/setup_vhost_block_backend.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  label def install_vhost_backend
+    q_version = frame["version"].shellescape
+    sshable.cmd("sudo host/bin/setup-vhost-block-backend install #{q_version}")
+
+    hop_update_database
+  end
+
+  label def update_database
+    vhost_block_backend = VhostBlockBackend.where(vm_host_id: vm_host.id, version: frame["version"]).first
```
jeremyevans:
```suggestion
    vhost_block_backend = VhostBlockBackend.first(vm_host_id: vm_host.id, version: frame["version"])
```

prog/storage/setup_vhost_block_backend.rb
```diff
@@ -0,0 +1,53 @@
...
+      allocation_weight: 0,
+      vm_host_id: vm_host.id
+    ) { it.id = VhostBlockBackend.generate_uuid }
+
+    hop_install_vhost_backend
+  end
+
+  label def install_vhost_backend
+    q_version = frame["version"].shellescape
+    sshable.cmd("sudo host/bin/setup-vhost-block-backend install #{q_version}")
```
jeremyevans: Can this command fail?  Does it take a long time?  If it cannot fail and is very fast, I don't see a need for a separate update_database label.  If it can fail, this needs failure handling.  If it can take a long time, there should be a separate `wait_install_vhost_backend` label.

rhizome/host/lib/storage_volume.rb
```diff
@@ -67,9 +83,124 @@ def prep(key_wrapping_secrets)
...
   end
 
+  def prep_vhost_backend(encryption_key, key_wrapping_secrets)
+    vhost_backend_create_config(encryption_key, key_wrapping_secrets)
+    vhost_backend_create_metadata(key_wrapping_secrets) if @image_path
+    vhost_backend_create_service_file
+  end
+
+  def vhost_backend_config_path
+    "/var/storage/#{@vm_name}/#{@disk_index}/vhost-backend.conf"
```
jeremyevans: Might be a good idea to add a `vhost_backend_dir` method that returned the directory name, so you didn't need to repeat it.

rhizome/host/lib/storage_volume.rb
```diff
@@ -67,9 +83,124 @@ def prep(key_wrapping_secrets)
...
+  def vhost_backend_metadata_path
+    "/var/storage/#{@vm_name}/#{@disk_index}/metadata"
+  end
+
+  def vhost_backend_create_config(encryption_key, key_wrapping_secrets)
+    config_path = vhost_backend_config_path
+    config = vhost_backend_config(encryption_key, key_wrapping_secrets)
+    File.write(config_path, config.to_yaml)
+    FileUtils.chown @vm_name, @vm_name, config_path
+    FileUtils.chmod "u=rw,g=,o=", config_path
```
jeremyevans: I believe this is a race condition, since there a point between when the data is written and when the ownership and mode is changed.  Whether the race condition is exploitable depends on the user's umask and maybe what group the file created uses.  I'm not sure how important this is, but if we want to ensure the file is created safely, we can use an approach like this:

```suggestion
    File.open(config_path, "w", 0600, flags: File::CREAT|File::EXCL) do |file|
      FileUtils.chown @vm_name, @vm_name, config_path
      file.write(config.to_yaml)
    end
```

This ensures the file does not already exist, and sets the mode and ownership before writing the data.  Note that if the write fails, this is not retryable, since the partially written file will already exist. If you want to handle that, writing to a file in the same directory (`"#{config_path}.tmp"`), and renaming it after the write to the final file name should work, as long as you delete the `.tmp` file if it exists before starting the process.

rhizome/host/lib/storage_volume.rb
```diff
@@ -67,9 +83,124 @@ def prep(key_wrapping_secrets)
...
+      kek_yaml = vhost_backend_kek(key_wrapping_secrets).to_yaml
+      kek_arg = "--kek /dev/stdin"
+    else
+      kek_yaml = ""
+      kek_arg = ""
+    end
+    FileUtils.touch(metadata_path)
+    File.truncate(metadata_path, 8 * 1024 * 1024)
+    FileUtils.chown @vm_name, @vm_name, metadata_path
+    FileUtils.chmod "u=rw,g=,o=", metadata_path
```
jeremyevans: Similar race condition as above.

rhizome/host/lib/storage_volume.rb
```diff
@@ -67,9 +83,124 @@ def prep(key_wrapping_secrets)
...
+    end
+
+    if @encrypted
+      key_encryption = StorageKeyEncryption.new(key_wrapping_secrets)
+      key1_bytes = [encryption_key[:key]].pack("H*")
+      key2_bytes = [encryption_key[:key2]].pack("H*")
+      key1_wrapped = key_encryption.wrap_key(key1_bytes).join
+      key2_wrapped = key_encryption.wrap_key(key2_bytes).join
+      key1_wrapped_b64 = Base64.strict_encode64(key1_wrapped).strip
+      key2_wrapped_b64 = Base64.strict_encode64(key2_wrapped).strip
```
jeremyevans: Since you do not need the intermediate variables, I recommend creating a method that returns the base64 wrapped values, and calling that method for key1 and key2.

rhizome/host/lib/storage_volume.rb
```diff
@@ -87,7 +218,44 @@ def start(key_wrapping_secrets)
     end
   end
 
+  def vhost_backend_start(key_wrapping_secrets)
+    if !@encrypted
```
jeremyevans:
```suggestion
    unless @encrypted
```

rhizome/host/lib/storage_volume.rb
```diff
@@ -87,7 +218,44 @@ def start(key_wrapping_secrets)
...
+    File.delete(kek_pipe) if File.exist?(kek_pipe)
+    File.mkfifo(kek_pipe, 0o600)
+    FileUtils.chown @vm_name, @vm_name, kek_pipe
+
+    r "systemctl start #{@vm_name}-storage"
+
+    begin
+      Timeout.timeout(5) do
+        kek_yaml = vhost_backend_kek(key_wrapping_secrets).to_yaml
+        File.open(kek_pipe, File::WRONLY) { |f| f.write(kek_yaml) }
```
jeremyevans:
```suggestion
        File.write(kek_pipe, kek_yaml)
```

rhizome/host/lib/storage_volume.rb
```diff
@@ -87,7 +218,44 @@ def start(key_wrapping_secrets)
     end
   end
 
+  def vhost_backend_start(key_wrapping_secrets)
+    if !@encrypted
+      r "systemctl start #{@vm_name}-storage"
+      return
+    end
+
+    File.delete(kek_pipe) if File.exist?(kek_pipe)
```
jeremyevans: Might be worth wrapping the rest of the method in a begin/ensure, and have the ensure delete `kek_pipe`.

rhizome/host/lib/storage_volume.rb
```diff
@@ -87,7 +218,44 @@ def start(key_wrapping_secrets)
...
+    rescue Timeout::Error
+      puts "Timed out while writing KEK to pipe #{kek_pipe}."
+
+      # Stop the service so we can retry
+      r "systemctl stop #{@vm_name}-storage"
+    end
+  end
+
+  def kek_pipe
+    @kek_pipe ||= "/tmp/#{@vm_name}-kek.pipe"
```
jeremyevans: If there is a vm-specific folder that we could use instead of `/tmp`, I think that would be preferable.

rhizome/host/lib/vhost_block_backend.rb
```diff
@@ -0,0 +1,47 @@
...
+      "f34de686dd7d1d7fcc26ba6ab593e81253600195610823ab1c61f603f337c6cc"
+    elsif @version == "v0.1-5" && Arch.sym == :arm64
+      "ee6298fa0024aa15543dfa7fd6f2f6795ab15db3134ea14ae44803c07d3a35ee"
+    else
+      fail "Unsupported version: #{@version}, #{Arch.sym}"
+    end
+  end
+
+  def url
+    "https://github.com/ubicloud/ubiblk/releases/download/#{@version}/vhost-backend-#{Arch.x64? ? "x64" : "arm64"}.tar.gz"
```
jeremyevans:
```suggestion
    "https://github.com/ubicloud/ubiblk/releases/download/#{@version}/vhost-backend-#{Arch.sym}.tar.gz"
```

rhizome/host/lib/vhost_block_backend.rb
```diff
@@ -0,0 +1,47 @@
...
+
+class VhostBlockBackend
+  def initialize(version)
+    @version = version
+  end
+
+  def sha256
+    if @version == "v0.1-5" && Arch.sym == :x64
+      "f34de686dd7d1d7fcc26ba6ab593e81253600195610823ab1c61f603f337c6cc"
+    elsif @version == "v0.1-5" && Arch.sym == :arm64
```
jeremyevans:
```suggestion
    if @version == "v0.1-5" && Arch.x64?
      "f34de686dd7d1d7fcc26ba6ab593e81253600195610823ab1c61f603f337c6cc"
    elsif @version == "v0.1-5" && Arch.arm64?
```

rhizome/host/lib/vhost_block_backend.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def init_metadata_path
+    "#{dir}/init-metadata"
+  end
+
+  def download
+    temp_tarball = "/tmp/vhost-backend-#{@version}.tar.gz"
+    puts "Downloading ubiblk package from #{url}"
+    r "curl -L3 -o #{temp_tarball} #{url}"
```
jeremyevans:
```suggestion
    r "curl -L3 --connect-timeout 5 --max-time 30 -o #{temp_tarball} #{url}"
```

To prevent this from blocking for too long.

scheduling/allocator.rb
```diff
@@ -570,6 +570,19 @@ def self.allocate_spdk_installation(spdk_installations)
       rand_choice.id
     end
 
+    def self.allocate_vhost_block_backend(backends)
+      total_weight = backends.sum(&:allocation_weight)
+      fail "Total weight of all eligible vhost_block_backends shouldn't be zero." if total_weight == 0
+
+      rand_point = rand(0..total_weight - 1)
```
jeremyevans:
```suggestion
      rand_point = rand(0...total_weight)
```

scheduling/allocator.rb
```diff
@@ -570,6 +570,19 @@ def self.allocate_spdk_installation(spdk_installations)
...
+      weight_sum = 0
+      rand_choice = backends.each { |si|
+        weight_sum += si.allocation_weight
+        break si if weight_sum > rand_point
+      }
```
jeremyevans:
```suggestion
      rand_choice = backends.find do |si|
        weight_sum += si.allocation_weight
        weight_sum > rand_point
      end
```

scheduling/allocator.rb
```diff
@@ -597,7 +610,15 @@ def map_volumes_to_devices
 
     def create_storage_volumes(vm, vm_host)
       @request.storage_volumes.each do |disk_index, volume|
-        spdk_installation_id = StorageAllocation.allocate_spdk_installation(vm_host.spdk_installations)
+        if vm_host.vhost_block_backends.sum(&:allocation_weight) == 0
```
jeremyevans: No need to sum at all, just check if there are no positive allocation weights (assuming that negative allocation weights are not supported):

```suggestion
        if vm_host.vhost_block_backends_dataset.exclude(allocation_weight: 0).empty?
```

migrate/20250506_vhost_block_backend.rb
```diff
@@ -0,0 +1,18 @@
...
+      foreign_key :vm_host_id, :vm_host, type: :uuid, null: false
+      unique [:vm_host_id, :version]
```
jeremyevans: Maybe add CHECK constraint for `allocation_weight >= 0`, unless you want to support negative weights?

scheduling/allocator.rb
```diff
@@ -623,12 +644,13 @@ def create_storage_volumes(vm, vm_host)
           vm_id: vm.id,
           boot: volume["boot"],
           size_gib: volume["size_gib"],
-          use_bdev_ubi: SpdkInstallation[spdk_installation_id].supports_bdev_ubi? && volume["boot"],
+          use_bdev_ubi: use_bdev_ubi,
```
jeremyevans: Hash value omission:

```suggestion
          use_bdev_ubi:,
```

scheduling/allocator.rb
```diff
@@ -623,12 +644,13 @@ def create_storage_volumes(vm, vm_host)
...
           boot_image_id: image_id,
           skip_sync: volume["skip_sync"],
           disk_index: disk_index,
           key_encryption_key_1_id: key_encryption_key&.id,
           spdk_installation_id: spdk_installation_id,
+          vhost_block_backend_id: vhost_block_backend_id,
```
jeremyevans:
```suggestion
          vhost_block_backend_id:,
```

rhizome/host/spec/storage_volume_spec.rb
jeremyevans: I don't think it's worth trying to refactor and split things up to allow for less mocking in the test, as long as this is tested in the E2E tests (which I assume it would be).

migrate/20250506_vhost_block_backend.rb
pykello: done in #3432

migrate/20250506_vhost_block_backend.rb
pykello: done in #3432

model/vhost_block_backend.rb
pykello: done in #3432

model/vm_host.rb
pykello: done in #3432

prog/storage/setup_vhost_block_backend.rb
pykello: done in #3432

prog/storage/setup_vhost_block_backend.rb
pykello: done in #3432

prog/storage/setup_vhost_block_backend.rb
pykello: done in #3432

prog/storage/setup_vhost_block_backend.rb
pykello: addressed in #3432

1. For error handling, I added a deadline to get paged if retries don't succeed
2. When I tried, this command took 3.5s. Assuming this might take longer depending on the load, I started using `daemonizer` and used the patter we use in other places.

prog/storage/setup_vhost_block_backend.rb
pykello: done in #3432

rhizome/host/lib/vhost_block_backend.rb
pykello: done in #3432

rhizome/host/lib/vhost_block_backend.rb
pykello: done in #3432

rhizome/host/lib/vhost_block_backend.rb
pykello: done in #3432

rhizome/host/lib/storage_volume.rb
pykello: done

rhizome/host/lib/storage_volume.rb
pykello: done

rhizome/host/lib/storage_volume.rb
pykello: done

rhizome/host/lib/storage_volume.rb
pykello: done

rhizome/host/lib/storage_volume.rb
pykello: done

rhizome/host/lib/storage_volume.rb
pykello: Done.

scheduling/allocator.rb
pykello: Done.

scheduling/allocator.rb
pykello: Done.

scheduling/allocator.rb
pykello: Done.

scheduling/allocator.rb
pykello: Done.

scheduling/allocator.rb
pykello: Done.

rhizome/host/lib/storage_volume.rb
pykello: Addressed by using `sp.` for paths.

rhizome/host/lib/storage_volume.rb
pykello: done

PR #3265, opened 2025-05-07
Add backend API to fetch metrics for Postgres resource

shikharbhardwaj:
## Description

This commit introduces an API to fetch metrics for a given PG resource. The list supported metrics is maintained in lib/metrics.rb. The system supports upto a month of historical data, with the query evaluation interval adjusted to return upto 480 datapoints per time series.

## Tests
- [x] Unit tests
- [x] Local run

----

> [!IMPORTANT]
> Adds an API endpoint to fetch metrics for Postgres resources, supporting various metrics and historical data, with comprehensive validation and error handling.
> 
>   - **API**:
>     - Adds `GET /project/{project_id}/location/{location}/postgres/{postgres_database_reference}/metrics` endpoint in `openapi.yml` to fetch metrics for a Postgres database.
>     - Supports optional query parameters `start`, `end`, and `key` for time range and specific metric selection.
>   - **Metrics**:
>     - Introduces `lib/metrics.rb` to define supported Postgres metrics like `cpu_usage`, `memory_usage`, `disk_io`, etc.
>     - Implements `query_range` in `lib/victoria_metrics/client.rb` to fetch time series data from VictoriaMetrics.
>   - **Validation**:
>     - Adds `validate_rfc3339_datetime_str` in `lib/validation.rb` to validate datetime strings.
>   - **Configuration**:
>     - Updates `metrics_config` in `lib/metrics_target_methods.rb` and `model/postgres/postgres_server.rb` to include `project_id` for metrics storage.
>   - **Error Handling**:
>     - Handles errors for invalid metric names, time ranges, and missing configurations in `routes/project/location/postgres.rb`.
>   - **Tests**:
>     - Adds tests in `spec/routes/api/project/location/postgres_spec.rb` for the new metrics endpoint.
>     - Includes unit tests for `query_range` in `spec/lib/victoria_metrics/client_spec.rb` and validation in `spec/lib/validation_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 75e43d501835cf18d986dbcf74a978d6cb527cfd. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/metrics.rb
+++ b/lib/metrics.rb
@@ -1,0 +1,161 @@
+# frozen_string_literal: true
+
+module Metrics
+  TimeSeries = Data.define(:labels, :query)
+  MetricDefinition = Data.define(:name, :description, :unit, :series)
+
+  POSTGRES_METRICS = {
+    cpu_usage:
+    MetricDefinition.new(
+      name: "CPU Usage",
+      description: "Percentage of CPU used by the system",
+      unit: "percent",
+      series: [
+        TimeSeries.new(
+          labels: {},
+          query: "(1 - sum(avg(rate(node_cpu_seconds_total{mode=\"idle\", ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"}[1m])))) * 100"
+        )
+      ]
+    ),
+    load_average:
+    MetricDefinition.new(
+      name: "Load Average",
+      description: "System load average over different time periods",
+      unit: nil,
+      series: [
+        TimeSeries.new(
+          labels: {name: "1 minute"},
+          query: "sum(node_load1{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"})"
+        ),
+        TimeSeries.new(
+          labels: {name: "5 minutes"},
+          query: "sum(node_load5{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"})"
+        ),
+        TimeSeries.new(
+          labels: {name: "15 minutes"},
+          query: "sum(node_load15{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"})"
+        )
+      ]
+    ),
+    memory_usage:
+    MetricDefinition.new(
+      name: "Memory Usage",
+      description: "Memory usage statistics",
+      unit: "percent",
+      series: [
+        TimeSeries.new(
+          labels: {name: "Used Memory"},
+          query: "(1 - (node_memory_MemAvailable_bytes{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"} / node_memory_MemTotal_bytes{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"})) * 100"
+        ),
+        TimeSeries.new(
+          labels: {name: "Cache & Buffers"},
+          query: "(node_memory_Cached_bytes{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"} + node_memory_Buffers_bytes{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"}) / node_memory_MemTotal_bytes{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"} * 100"
+        )
+      ]
+    ),
+    disk_usage:
+    MetricDefinition.new(
+      name: "Disk Usage",
+      description: "Disk space utilization",
+      unit: "percent",
+      series: [
+        TimeSeries.new(
+          labels: {name: "Used Space"},
+          query: "100 - (sum(node_filesystem_avail_bytes{mountpoint=\"/dat\", ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"} / node_filesystem_size_bytes{mountpoint=\"/dat\", ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"}) * 100)"
+        )
+      ]
+    ),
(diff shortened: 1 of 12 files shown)


lib/metrics.rb
```diff
@@ -0,0 +1,183 @@
...
+  class MetricDefinition
+    attr_accessor :name, :description, :unit, :series
+
+    def initialize(name:, description:, unit:, series:)
+      @name = name
+      @description = description
+      @unit = unit
+      @series = series
+    end
+  end
```
enescakir, line 22: You can use the Data core class. It's similar to a struct but immutable and also allow named arguments.

https://www.shakacode.com/blog/ruby-3-2-adds-a-new-data-class/

shikharbhardwaj: Nice, I was looking for something similar to Python's `dataclasses`, this looks like it

lib/metrics.rb
```diff
@@ -0,0 +1,183 @@
...
+
+  POSTGRES_METRICS = [
+    MetricDefinition.new(
+      name: "CPU Usage",
```
enescakir, line 26: Should we change the metric name to something URL-safe, like `cpu_usage`? We can keep `CPU Usage` as the display name. Sending the `Cache hit ratio` string in the request doesn't feel very programmatic.

lib/metrics.rb
```diff
@@ -0,0 +1,183 @@
...
+
+    def initialize(name:, description:, unit:, series:)
+      @name = name
+      @description = description
+      @unit = unit
+      @series = series
+    end
+  end
+
+  POSTGRES_METRICS = [
```
enescakir, line 24: We often map or find by name. Should we use a hash instead of an array, like `{ name: details }`? This way, we can get the list of metric names with `POSTGRES_METRICS.keys` or find a specific metric with `POSTGRES_METRICS[name]`

lib/metrics.rb
shikharbhardwaj: I was thinking about this, along with the change to use the hash below. I will make this change.

lib/metrics.rb
shikharbhardwaj: Done

lib/metrics.rb
shikharbhardwaj: Done both

routes/project/location/postgres.rb
```diff
@@ -319,6 +319,87 @@ class Clover
...
+
+              # This can be a two cases:
+              # 1. Missing data (e.g. no data for the given time range)
+              # 2. No data for the given query (maybe bad query)
+              if series_query_result.empty?
+                next
+              end
+
+              # Combine labels with configured series labesls.
+              series_query_result.each { it["labels"].merge!(s.labels) }
```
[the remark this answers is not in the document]
jeremyevans: I don't think this suggestion improves the code.

lib/metrics.rb
```diff
@@ -0,0 +1,177 @@
...
+          labels: {name: "Received"},
+          query: "rate(node_network_receive_bytes_total{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"}[1m])"
+        ),
+        TimeSeries.new(
+          labels: {name: "Transmitted"},
+          query: "rate(node_network_transmit_bytes_total{ubicloud_resource_id=\"$ubicloud_resource_id\", ubicloud_resource_role=\"primary\"}[1m])"
+        )
+      ]
+    ),
+    connection_count:
```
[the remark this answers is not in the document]
shikharbhardwaj: I am thinking to remove `Disk I/O`, for some reason the read IOPs always show up as 0, could be bad measurement from the exporter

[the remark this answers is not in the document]
byucesoy: It would be good to show read and write IOPS since we claim we provide superior disk performance. Maybe remove it now, but we can debug why read IOPS show up as 0 and add it back.

routes/project/location/postgres.rb
```diff
@@ -308,6 +308,87 @@ class Clover
         response.headers["Content-Type"] = "application/x-pem-file"
         certs
       end
+
+      r.get "metrics" do
+        authorize("Postgres:view", pg.id)
+
+        start_ts = request.params["start"]&.to_i || Time.now.utc.to_i - 30 * 60
+        end_ts = request.params["end"]&.to_i || Time.now.utc.to_i
```
byucesoy, line 316: Instead of integer, can we get data in some popular datetime format (i.e. something we can pass to Time.parse()) for example? That would be more user friendly I believe.

routes/project/location/postgres.rb
```diff
@@ -308,6 +308,87 @@ class Clover
...
+        metric_key = request.params["key"]&.to_sym
+        single_query = !request.params["key"].nil?
+
+        if single_query && !Metrics::POSTGRES_METRICS.key?(metric_key)
+          raise CloverError.new(400, "InvalidRequest", "Invalid metric name")
+        end
+
+        metric_keys = metric_key ? [metric_key] : Metrics::POSTGRES_METRICS.keys
+
+        vmr = VictoriaMetricsResource.first(project_id: Config.victoria_metrics_service_project_id)
```
byucesoy, line 339: Shouldn't project_id of the VictoriaMetricsResource be `Config.postgres_service_project_id` (i.e. VictoriaMetricsResource belongs to Postgres)?

project_id of the VMs should be `Config.victoria_metrics_service_project_id` (i.e. VMs belong to VictoriaMetrics)

byucesoy (approved):

routes/project/location/postgres.rb
shikharbhardwaj: I think this makes more sense. There are a couple of other parts that assume otherwise, let me update those here.

routes/project/location/postgres.rb
shikharbhardwaj: I'll update this to support both timestamps and rfc3339 datetime strings

PR #3272, opened 2025-05-07
Create billing info as needed for discount code

0xjunhao:
- Allow customers to use a discount code without existing billing info, and create billing info using the current user's name and email when needed.
- Explicitly require a billing address when adding a payment method if the address is missing from the Stripe customer, and then update the customer with the billing address from the payment method.
- Refactor Stripe customer caching.

----

> [!IMPORTANT]
> Enhance discount code handling by creating billing info as needed and refining billing address management in Stripe integration.
> 
>   - **Behavior**:
>     - Allow discount code usage without existing billing info by creating billing info with current user's name and email in `discount_code.rb`.
>     - Require billing address when adding payment method if missing from Stripe customer in `billing.rb`.
>     - Update Stripe customer with billing address from payment method in `billing.rb`.
>   - **Models**:
>     - Add `has_address?` method to `BillingInfo` to check for address presence.
>     - Refactor `stripe_data` in `BillingInfo` to handle missing data gracefully.
>   - **Views**:
>     - Extract discount code form to `components/discount_code.erb`.
>     - Update `billing.erb` to use new discount code component.
>   - **Tests**:
>     - Add tests for discount code application and billing info creation in `billing_spec.rb` and `discount_code_spec.rb`.
>     - Add tests for VAT handling and address updates in `billing_info_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9dc1a68dade75be08adb7a74e082892e03aa0bf2. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/billing_info.rb
+++ b/model/billing_info.rb
@@ -15,7 +15,9 @@ class BillingInfo < Sequel::Model
     if (Stripe.api_key = Config.stripe_secret_key)
       @stripe_data ||= begin
         data = Stripe::Customer.retrieve(stripe_id)
-        address = data["address"]
+        return nil unless data
+        address = data["address"] || {}
+        metadata = data["metadata"] || {}
         {
           "name" => data["name"],
           "email" => data["email"],
@@ -24,11 +26,15 @@ class BillingInfo < Sequel::Model
           "city" => address["city"],
           "state" => address["state"],
           "postal_code" => address["postal_code"],
-          "tax_id" => data["metadata"]["tax_id"],
-          "company_name" => data["metadata"]["company_name"]
+          "tax_id" => metadata["tax_id"],
+          "company_name" => metadata["company_name"]
         }
       end
     end
+  end
+
+  def has_address?
+    !stripe_data&.[]("address").to_s.empty?
   end
 
   def country
--- a/routes/project/billing.rb
+++ b/routes/project/billing.rb
@@ -48,7 +48,7 @@ class Clover
             if new_tax_id != current_tax_id
               DB.transaction do
                 billing_info.update(valid_vat: nil)
-                if new_tax_id && billing_info.country.in_eu_vat?
+                if new_tax_id && billing_info.country&.in_eu_vat?
                   Strand.create(prog: "ValidateVat", label: "start", stack: [{subject_id: billing_info.id}])
                 end
               end
@@ -77,7 +77,8 @@ class Clover
         setup_intent = Stripe::SetupIntent.retrieve(checkout_session["setup_intent"])
 
         stripe_id = setup_intent["payment_method"]
-        card_fingerprint = Stripe::PaymentMethod.retrieve(stripe_id)["card"]["fingerprint"]
+        stripe_payment_method = Stripe::PaymentMethod.retrieve(stripe_id)
+        card_fingerprint = stripe_payment_method["card"]["fingerprint"]
         if PaymentMethod.where(fraud: true).select_map(:card_fingerprint).include?(card_fingerprint)
           flash["error"] = "Payment method you added is labeled as fraud. Please contact support."
           r.redirect @project.path + "/billing"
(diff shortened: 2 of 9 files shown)


routes/project/billing.rb
```diff
@@ -123,6 +124,13 @@ class Clover
           PaymentMethod.create_with_id(billing_info_id: billing_info.id, stripe_id: stripe_id, card_fingerprint: card_fingerprint, preauth_intent_id: payment_intent.id, preauth_amount: preauth_amount)
         end
 
+        if !@project.billing_info.has_address?
+          Stripe::Customer.update(@project.billing_info.stripe_id, {
+            address: stripe_payment_method["billing_details"]["address"].to_hash
+          })
+          @project.billing_info.clear_stripe_customer_cache
```
enescakir, line 131: Do we really need to clear the cache? Since we redirect afterwards, doesn't that reload the object?

routes/project/discount_code.rb
```diff
@@ -33,6 +33,12 @@ class Clover
           flash["notice"] = "Discount code successfully applied."
         end
 
+        unless @project.billing_info
+          stripe_customer = Stripe::Customer.create(name: current_account.name, email: current_account.email)
+          billing_info = BillingInfo.create_with_id(stripe_id: stripe_customer["id"])
+          @project.update(billing_info_id: billing_info.id)
+        end
```
enescakir, line 40: Web routes do not run in a database transaction like progs do, so we need to move them into a transaction for data consistency. Also, do we want to create billing info if the discount code couldn't be applied?

0xjunhao: Done. Have put them in a transaction and only invoke for valid code.

routes/project/billing.rb
0xjunhao: Removed.

routes/project/billing.rb
```diff
@@ -134,6 +141,7 @@ class Clover
             payment_method_types: ["card"],
             mode: "setup",
             customer: billing_info.stripe_id,
+            billing_address_collection: billing_info.has_address? ? "auto" : "required",
```
bsatzger, line 144: could there be any differences creating a customer through `Customer.create` followed by  `Session.create` vs creating a customer through `Session.create` other than billing address?

routes/project/discount_code.rb
```diff
@@ -30,6 +30,13 @@ class Clover
         rescue Sequel::UniqueConstraintViolation
           flash["error"] = "Discount code has already been applied to this project."
         else
+          unless @project.billing_info
```
bsatzger, line 33: everyone entering a discount code would be considered having a valid payment method, right? they'd have billing info and credit > 0. we should at least keep an eye on suspicious activity of projects without payment method but a coupon. 

```ruby
  def has_valid_payment_method?
    return true unless Config.stripe_secret_key
    !!billing_info&.payment_methods&.any? || (!!billing_info && credit > 0)
  end
  ```
  https://github.com/ubicloud/ubicloud/blob/6d568f373bb4a8b0ba75dc5c2ca73fcab1165a7e/model/project.rb#L40

File: model/billing_info.rb [hunk 5]
```diff
@@ -11,24 +11,29 @@ class BillingInfo < Sequel::Model
...
+      "city" => address["city"],
+      "state" => address["state"],
+      "postal_code" => address["postal_code"],
+      "tax_id" => data["metadata"]["tax_id"],
+      "company_name" => data["metadata"]["company_name"]
+    }
+  end
+
+  def has_address?
+    !stripe_customer&.[]("address").nil?
```
enescakir, line 36: Since we removed the cache clearing, I don't think we need the extra `stripe_customer` method.


```suggestion
    !stripe_data&.[]("address").empty?
```

routes/project/billing.rb
0xjunhao: There are some minor differences, but doesn't seem to be affecting our use cases. @enescakir

model/billing_info.rb
0xjunhao: Done.

bsatzger (approved):

enescakir (approved):

PR #3275, opened 2025-05-09
Show different display state if waiting for the maintenance window

byucesoy:
**Add in_maintenance_window? helper to PostgresResource**
Deciding whether a resource is in the maintenance window requires non-trivial
calculation. We will need it in multiple places, so we add a helper method to
the PostgresResource model.

**Show different display state if waiting for the maintenance window**
If the resource is ready for failover but waiting for the maintenance window,
it is better to show that in the UI. Otherwise users would have no idea on why
the resource is not failing over.

----

> [!IMPORTANT]
> Adds `in_maintenance_window?` to `PostgresResource` and updates display state and convergence logic to respect maintenance windows.
> 
>   - **Behavior**:
>     - Adds `in_maintenance_window?` to `PostgresResource` to check if current time is within maintenance window.
>     - Updates `display_state` in `PostgresResource` to show "waiting for maintenance window" if ready for failover but outside maintenance window.
>     - Modifies `recycle_representative_server` in `converge_postgres_resource.rb` to wait if outside maintenance window.
>   - **Tests**:
>     - Adds tests for `in_maintenance_window?` and updated `display_state` logic in `postgres_resource_spec.rb`.
>     - Updates tests in `converge_postgres_resource_spec.rb` to reflect changes in maintenance window handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for e517a33bee9fadf1de74550603cc0895f767ef84. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -43,7 +43,11 @@ class PostgresResource < Sequel::Model
   def display_state
     return "deleting" if destroy_set? || strand.nil? || strand.label == "destroy"
     return "unavailable" if representative_server&.strand&.label == "unavailable"
-    return "converging" if strand.children.any? { it.prog == "Postgres::ConvergePostgresResource" }
+    if strand.children.any? { it.prog == "Postgres::ConvergePostgresResource" }
+      return "converging" if has_enough_ready_servers? && in_maintenance_window?
+      return "waiting for maintenance window" if has_enough_ready_servers? && !in_maintenance_window?
+      return "preparing for convergence"
+    end
     return "running" if ["wait", "refresh_certificates", "refresh_dns_record"].include?(strand.label) && !initial_provisioning_set?
     "creating"
   end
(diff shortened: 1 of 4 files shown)


model/postgres/postgres_resource.rb
```diff
@@ -43,7 +43,11 @@ def path
   def display_state
     return "deleting" if destroy_set? || strand.nil? || strand.label == "destroy"
     return "unavailable" if representative_server&.strand&.label == "unavailable"
-    return "converging" if strand.children.any? { it.prog == "Postgres::ConvergePostgresResource" }
+    if strand.children.any? { it.prog == "Postgres::ConvergePostgresResource" }
+      return "converging" if has_enough_ready_servers? && in_maintenance_window?
+      return "waiting maintenance window" if has_enough_ready_servers? && !in_maintenance_window?
```
shikharbhardwaj, line 48: nit, not sure if the UI would look okay with this though
```suggestion
      return "waiting for maintenance window" if has_enough_ready_servers? && !in_maintenance_window?
```

shikharbhardwaj (approved):

PR #3277, opened 2025-05-09
Support sub sub domains for K8s service URLs

aamederen:
This change adds an additional wildcard DNS record routing anything
    under the K8s services URL to the actual K8s services URL.

This would allow users to deploy different applications on the provided
    domain and configure ingress inside the k8s cluster to route everything
    to the correct app.

One such example is Okteto. When Okteto is deployed:
    - It deploys 4 different apps for the Okteto itself
    (okteto.xxxxx-services, registry.xxxxx-services)
    - It also deploys every development environment with a subdomain too

As a future work, we can also integrate DNS challenge support to our CCM
    with these subdomains to make things even more interesting.

----

> [!IMPORTANT]
> Add support for sub-subdomains in Kubernetes service URLs by managing wildcard DNS records in `kubernetes_nodepool_nexus.rb`.
> 
>   - **Behavior**:
>     - Adds wildcard DNS record for sub-subdomains in `create_services_load_balancer` in `kubernetes_nodepool_nexus.rb`.
>     - Deletes DNS record in `destroy` in `kubernetes_nodepool_nexus.rb` if DNS zone exists.
>   - **Tests**:
>     - Updates `kubernetes_nodepool_nexus_spec.rb` to test creation and deletion of DNS records.
>     - Adds test for DNS record creation with `CNAME` type and correct data.
>     - Adds test for DNS record deletion when DNS zone exists.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0a4b3e02b668cb656e24ee15c43d7617b9316024. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/kubernetes/kubernetes_nodepool_nexus.rb
+++ b/prog/kubernetes/kubernetes_nodepool_nexus.rb
@@ -34,11 +34,13 @@ class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
   label def create_services_load_balancer
     hop_bootstrap_worker_vms if LoadBalancer[name: kubernetes_nodepool.cluster.services_load_balancer_name]
 
-    custom_hostname_dns_zone_id = DnsZone[name: Config.kubernetes_service_hostname]&.id
-    custom_hostname_prefix = if custom_hostname_dns_zone_id
+    dns_zone = DnsZone[name: Config.kubernetes_service_hostname]
+
+    custom_hostname_prefix = if dns_zone
       "#{kubernetes_nodepool.cluster.ubid.to_s[-10...]}-services"
     end
-    Prog::Vnet::LoadBalancerNexus.assemble(
+
+    lb = Prog::Vnet::LoadBalancerNexus.assemble(
       kubernetes_nodepool.cluster.private_subnet_id,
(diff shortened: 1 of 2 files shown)


fdr:
> @furkansahin , @fdr do you have any opinions on this?

I think my main comment is that there's probably a better way to convey what's important about this patch, which is support for wildcard behavior. The two most-accepted names are "wildcard domain" or, maybe my preference though more obscure, "asterisk label." 

Is it our first time using it? To wit, I think so.

I did a little background research, I don't see any reason to be alarmed exercising this feature in KnotDNS. But given the latter is something of a @byucesoy special from antiquity, he should give it a moment's thought, too (or at least integrate the information)

byucesoy:
I think this is fine from the DNS perspective. We need to be careful about certs though. Do we create the certs with `*.#{lb.hostname}....` to the cert as well or with just `#{lb.hostname}....`. If latter, cert wouldn't be valid for wildcard domains.

aamederen:
@byucesoy, thanks. We don't create the certs for the wildcard domain.

PR #3280, opened 2025-05-09
Enable slice-based allocation by default on Ubuntu 24.04.

pykello:
Now that we have tested slice-based allocation for a while, it is reasonable to make it default for future host setups.

----

> [!IMPORTANT]
> Enable slice-based allocation by default for Ubuntu 24.04, removing conditional logic for allowing slices.
> 
>   - **Behavior**:
>     - Enables slice-based allocation by default for Ubuntu 24.04 in `host_nexus.rb`.
>     - Removes `allow_slices` semaphore and related logic in `hetzner_server.rb` and `ci`.
>   - **Functions**:
>     - Updates `os_supports_slices?` in `host_nexus.rb` to return true for Ubuntu 24.04.
>   - **Tests**:
>     - Removes tests related to `allow_slices` in `hetzner_server_spec.rb`.
>     - Adds test for `os_supports_slices?` in `host_nexus_spec.rb` to verify behavior for Ubuntu 24.04.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a8e92799626bc994a3954634fc96f8314ce41ef2. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/host_nexus.rb
+++ b/prog/vm/host_nexus.rb
@@ -90,11 +90,16 @@ class Prog::Vm::HostNexus < Prog::Base
     hop_wait_prep
   end
 
+  def os_supports_slices?(os_version)
+    os_version == "ubuntu-24.04"
+  end
+
   label def wait_prep
     reap.each do |st|
       case st.prog
       when "LearnOs"
-        vm_host.update(os_version: st.exitval.fetch("os_version"))
+        os_version = st.exitval.fetch("os_version")
+        vm_host.update(os_version: os_version, accepts_slices: os_supports_slices?(os_version))
       when "LearnMemory"
         mem_gib = st.exitval.fetch("mem_gib")
         vm_host.update(total_mem_gib: mem_gib)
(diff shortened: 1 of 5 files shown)


prog/vm/host_nexus.rb
```diff
@@ -90,11 +90,16 @@ def before_run
     hop_wait_prep
   end
 
+  def os_supports_slices?(os_version)
+    os_version == "ubuntu-24.04"
```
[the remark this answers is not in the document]
fdr: We'll just change the code. this is, like, among the worst genre of suggestion you have, maybe 10% even worthy of consideration, tops.

enescakir, line 94: Did we test the slices on arm64 hosts?

enescakir (approved):

prog/vm/host_nexus.rb
pykello: Good point.

In dev environment: yes. In prod: not yet

I'll try this on couple of arm64 machines for a while before merging the PR.

PR #3299, opened 2025-05-14
Ignore high-cardinality timeseries when collecting PG metrics

shikharbhardwaj:
Add filters to ignore table-cardinality timeseries when collecting metrics from PG instances, to avoid large scrape sizes leading to backlogs.


model/postgres/postgres_server.rb
```diff
@@ -257,9 +258,19 @@ def run_query(query)
   end
 
   def metrics_config
+    ignored_timeseries_patterns = [
+      "pg_stat_user_tables_.*",
+      "pg_statio_user_tables_.*"
+    ]
+    match_pattern = ignored_timeseries_patterns.join("|")
```
byucesoy, line 265: The variable name is bit confusing, it feels like we are trying to match those pattern, however we are trying to exclude them.

byucesoy (approved):

model/postgres/postgres_server.rb
shikharbhardwaj: makes sense, renamed

PR #3295, opened 2025-05-13
Bump images

fdr:
after puttering around to try to figure out how to encheapn this process (which is overdue), look at my cool makefile (a quick way to get resume and `make -j` parallelism)

```makefile
# Define URLs for all required images
URL_ubuntu-24.04-server-cloudimg-amd64.img := https://cloud-images.ubuntu.com/releases/noble/release-20250502.1/ubuntu-24.04-server-cloudimg-amd64.img
URL_ubuntu-24.04-server-cloudimg-arm64.img := https://cloud-images.ubuntu.com/releases/noble/release-20250502.1/ubuntu-24.04-server-cloudimg-arm64.img
URL_ubuntu-22.04-server-cloudimg-amd64.img := https://cloud-images.ubuntu.com/releases/jammy/release-20250508/ubuntu-22.04-server-cloudimg-amd64.img
URL_ubuntu-22.04-server-cloudimg-arm64.img := https://cloud-images.ubuntu.com/releases/jammy/release-20250508/ubuntu-22.04-server-cloudimg-arm64.img
URL_debian-12-genericcloud-amd64-20250428-2096.raw := https://cloud.debian.org/images/cloud/bookworm/20250428-2096/debian-12-genericcloud-amd64-20250428-2096.raw
URL_debian-12-genericcloud-arm64-20250428-2096.raw := https://cloud.debian.org/images/cloud/bookworm/20250428-2096/debian-12-genericcloud-arm64-20250428-2096.raw
URL_AlmaLinux-9-GenericCloud-9.5-20241120.x86_64.qcow2 := https://repo.almalinux.org/almalinux/9/cloud/x86_64/images/AlmaLinux-9-GenericCloud-9.5-20241120.x86_64.qcow2
URL_AlmaLinux-9-GenericCloud-9.5-20241120.aarch64.qcow2 := https://repo.almalinux.org/almalinux/9/cloud/aarch64/images/AlmaLinux-9-GenericCloud-9.5-20241120.aarch64.qcow2
URL_alpine-3.21.2-nocloud-uefi-x86_64.qcow2 := https://dl-cdn.alpinelinux.org/alpine/v3.21/releases/cloud/nocloud_alpine-3.21.2-x86_64-uefi-cloudinit-r0.qcow2
[...]

PR #3308, opened 2025-05-14
Add new Ubuntu 24.04, 22.04, Debian 12 image snapshots

fdr:
We're overdue to update the base snapshot, so here it is.

There is no new AlmaLinux of the 9.X series version yet.


fdr:
I actually integration-tested these (which means I uploaded them using @enescakir's somewhat obscure but still handy procedure) in #3295 . But this is the first, basically harmless commit, to prime host capacity  with new images.

jeremyevans (approved):

PR #3330, opened 2025-05-19
Upgrade acme-client gem

furkansahin:
It looks like there is a new error type that needs to be handled properly by acme-client "OrderNotReady". That is added to the gem in a later version https://github.com/unixcharles/acme-client/pull/246 which we did not upgrade to. Now, we have a page with this error for certificate cec899231d861h3wrvfe8pbpj0. Therefore, I am upgrading the client.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,7 +45,7 @@ GEM
   remote: https://rubygems.org/
   specs:
     Ascii85 (2.0.1)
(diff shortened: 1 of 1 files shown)


bsatzger (approved):
if we want to be super cautious, we could run e2e to make sure faraday and json library upgrades are not causing any issues.

PR #3332, opened 2025-05-19
CertNexus Cert finalization idempotancy fix

furkansahin:
We were setting the csr_key and finalizing the certificate. The problem with that is, if the cert finalization endpoint errors out but it essentially worked, we loose the csr_key information. To fix that, we first record the key to the database and use that continuously to finalize the certificate.


prog/vnet/cert_nexus.rb
```diff
@@ -63,19 +63,21 @@ def before_run
...
     else
       Clog.emit("DNS validation failed") { {order_status: dns_challenge.status} }
       dns_zone.delete_record(record_name: dns_record_name)
       hop_restart
     end
   end
 
+  label def cert_finalization
+    csr = Acme::Client::CertificateRequest.new(private_key: cert.csr_key, common_name: cert.hostname)
+    acme_order.finalize(csr: csr)
```
enescakir, line 77:
```suggestion
    acme_order.finalize(csr:)
```

enescakir (approved):

PR #3337, opened 2025-05-19
Accept location as either ubid or uuid

jeremyevans:
This is the first step in the process.  When rolling out this change, you could have a GET request to display a page go to a new version of the code, but the POST request when submitted to go to the old version of the code.  This allows for that to work without breaking, using a multi-step commit process:

1) Application only accepts uuid format
2) This commit
3) New versions accept uuid or ubid, old versions accept uuid only,
   uuid is used for submission in all versions
4) After a few minutes, all versions accept uuid or ubid
5) Future commit that switches to submitting ubid
6) All versions accept uuid or ubid, new versions submit ubid,
   old versions submit uuid
7) After a few minutes, all versions accept uuid or ubid, and submit
   ubid
8) Future commit that drops support for uuid
9) All versions submit ubid, new versions only accept ubid, old versions
   accept both
10) After a few minutes, all versions only accept ubid and only submit
   ubid

### DIFF
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -223,7 +223,7 @@ class Clover < Roda
     #
     # If location not previously retrieved, require it be visible or tied to the current project
     # when retrieving it.  This is called when creating resources in the web routes.
-    @location ||= if (id = typecast_params.uuid("location"))
+    @location ||= if (id = typecast_params.ubid_uuid("location") || typecast_params.uuid("location"))
       Location.visible_or_for_project(@project.id).first(id:)
     end
     handle_invalid_location unless @location&.visible_or_for_project?(@project.id)


helpers/general.rb
```diff
@@ -223,7 +223,7 @@ def check_visible_location
     #
     # If location not previously retrieved, require it be visible or tied to the current project
     # when retrieving it.  This is called when creating resources in the web routes.
-    @location ||= if (id = typecast_params.uuid("location"))
+    @location ||= if (id = typecast_params.ubid_uuid("location") || typecast_params.uuid("location"))
```
fdr, line 226: You're the expert, but I presume that instead of erroring these nil out (thus become falsy) if they don't match?

jeremyevans: Correct.  If you wanted to force a match, and raise if there wasn't a match, you could use `ubid_uuid!`.  In this case, we have existing error handling directly below this line, so it's better to use that.

fdr (approved):

PR #3344, opened 2025-05-21
Add node_exporter installation prog to VmHost

mohi-kalantari:
A new prog is added to install the node_exporter on VmHosts.

A flag is also added to make sure the application will serve the metrics on localhost instead of all network interfaces.

For the VmHosts which are already setup and do not have the app installed we can create this strand:

Strand.create(prog: "SetupNodeExporter", label: "start",
	stack: [{subject_id: vm_host.id}])


rhizome/host/bin/setup-node-exporter.rb
```diff
@@ -0,0 +1,45 @@
...
+end
+
+file_name = case r("common/bin/arch").strip
+when "arm64"
+  "node_exporter-#{version}.linux-arm64"
+when "x64"
+  "node_exporter-#{version}.linux-amd64"
+else
+  raise "Unsupported architecture: #{arch}"
+end
```
enescakir, line 18: We have a helper for it, the one @fdr loves the most.


```suggestion
file_name = "node_exporter-#{version}.linux-#{Arch.render(x64: "amd64", arm64: "arm64")}"
```

mohi-kalantari: that's way cleaner thanks!

prog/setup_node_exporter.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Prog::SetupNodeExporter < Prog::Base
+  subject_is :sshable, :vm_host
```
fdr, line 4: why vm_host rather than bare sshables?

fdr: I'm also a bit puzzled if this is related to kubernetes or not, because the host doesn't have much insight into what the guests are doing.

enescakir: > later we would use the same prog for kubernetes vms too

It's probably better to download it in image generation for k8s use case for faster boot up like PG images.

mohi-kalantari: It's pretty tiny but we can put it later you're right. thanks for the suggestion.

rhizome/host/bin/setup-node-exporter
```diff
@@ -0,0 +1,39 @@
...
+unless (version = ARGV.shift)
+  fail "No version provided"
+end
+
+file_name = "node_exporter-#{version}.linux-#{Arch.render(x64: "amd64", arm64: "arm64")}"
+url = "https://github.com/prometheus/node_exporter/releases/download/v#{version}/#{file_name}.tar.gz"
+
+r "wget -q \"#{url}\""
+r "tar xvfz #{file_name}.tar.gz"
+r "sudo mv #{file_name}/node_exporter /usr/local/bin/"
```
aamederen, line 17: what do you think about creating a metrics user and chmod + chowning this to the metrics user?

aamederen (approved):
Overall, LGTM.

rhizome/host/bin/setup-node-exporter
```diff
@@ -0,0 +1,39 @@
...
+[Service]
+Type=simple
+ExecStart=/usr/local/bin/node_exporter --web.listen-address=127.0.0.1:9100
+Restart=always
+User=nobody
+Group=nogroup
+
+[Install]
+WantedBy=multi-user.target
+NODEEXPORTERCONFIG
```
fdr: You should be able to improve on this with `systemd-analyze security`. Not every checklist entry helps (like, RootDirectory/RootImage is overkill), but I think you could easily eliminate quite a lot of access.

Here's some stuff to get started.  It's based on https://gist.github.com/ageis/f5595e59b1cddb1513d1b425a323db04

```
[Service]
# Basic privilege restrictions
NoNewPrivileges=yes
PrivateTmp=yes
PrivateDevices=yes
DevicePolicy=closed

# File system protection
ProtectSystem=strict
ProtectHome=read-only
ProtectControlGroups=yes
ProtectKernelModules=yes
ProtectKernelTunables=yes

# Network and process restrictions
RestrictAddressFamilies=AF_INET
RestrictNamespaces=yes
RestrictRealtime=yes
RestrictSUIDSGID=yes

# Memory and execution protection (may break v8 jit)
MemoryDenyWriteExecute=yes
LockPersonality=yes
```

mohi-kalantari: Since the systemd unit is for node-exporter, it needs access to certain parts of the filesystem for example /dev to read system info.

```
PrivateDevices=yes
DevicePolicy=closed
RestrictNamespaces=yes
ProtectControlGroups=yes
```

I'll try to find a subset of those which doesn't break anything and is safe.

rhizome/host/bin/setup-node-exporter
mohi-kalantari: with Daniel's comment, I guess now the node exporter is way more restricted and we wouldn't gain much from changing the user

rhizome/host/bin/setup-node-exporter
mohi-kalantari: I found a combination which works fine. I had to remove 5 parameters. 4 mentioned above + RestrictAddressFamilies=AF_INET

PR #3348, opened 2025-05-22
Do not create cert related strands when LB doesn't need certs

mohi-kalantari:
Prior to this commit, we would create strands for putting or removing certs for VMs belonging to a LoadBalancer which didn't need to provision certificates.

----

> [!IMPORTANT]
> Ensure certificate strands are created only when load balancer requires certificates by checking `cert_enabled_lb?`.
> 
>   - **Behavior**:
>     - Modify `add_vm`, `detach_vm`, and `evacuate_vm` in `load_balancer.rb` to conditionally create strands for certificate operations only if `cert_enabled_lb?` is true.
>     - Introduce `cert_enabled_lb?` method in `load_balancer.rb` to check if the load balancer requires certificates based on `health_check_protocol`.
>   - **Tests**:
>     - Add tests in `load_balancer_spec.rb` for `detach_vm` to verify strand creation is conditional on `cert_enabled_lb?`.
>     - Add test for `cert_enabled_lb?` in `load_balancer_spec.rb` to ensure it returns false when `health_check_protocol` is not `https`.
>     - Update `load_balancer_nexus_spec.rb` to reflect changes in load balancer assembly with `health_check_protocol`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 2a30ed73e16f81aaab56b18fa1b475fdee1fa0d8. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -64,7 +64,7 @@ class LoadBalancer < Sequel::Model
       ports.each { |port|
         LoadBalancerVmPort.create(load_balancer_port_id: port.id, load_balancer_vm_id: load_balancer_vm.id)
       }
-      Strand.create_with_id(prog: "Vnet::CertServer", label: "put_certificate", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id)
+      Strand.create(prog: "Vnet::CertServer", label: "put_certificate", stack: [{subject_id: id, vm_id: vm.id}], parent_id: id) if cert_enabled_lb?
       incr_rewrite_dns_records
     end
   end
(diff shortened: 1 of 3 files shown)


model/load_balancer.rb
```diff
@@ -114,8 +114,12 @@ def dns_zone
     custom_hostname_dns_zone || DnsZone[project_id: Config.load_balancer_service_project_id, name: Config.load_balancer_service_hostname]
   end
 
+  def need_certificate_provisioning?
+    health_check_protocol == "https"
+  end
```
furkansahin, line 119: rename this to `def cert_enabled_lb?`

PR #3351, opened 2025-05-22
Thread pool

fdr:
I rebased your change, but left to the side, and added my ideas to lib. Needs a bit more testing work, going to stress test it for any deadlocks. also need to check it more carefully for correctness on terminating with deadlines. Overhead is on the order of 1.4us per job.

### DIFF
--- a/lib/deadline_thread_pool.rb
+++ b/lib/deadline_thread_pool.rb
@@ -49,7 +49,7 @@ class DeadlineThreadPool
 
   def check_deadlines
     # This .values.min is the most dubiously expensive thing in this.
-    return nil unless (earliest_deadline = @running_deadlines.values.min)
+    return nil unless (earliest_deadline = @running_deadlines.each_value.min)
     now = Time.now
       
     # Check if any deadline is passed.
--- a/lib/deadline_thread_pool.rb
+++ b/lib/deadline_thread_pool.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+class DeadlineThreadPool
+  DEFAULT_DEADLINE = 60 # seconds
+
+  def initialize(size, job_deadline_seconds = DEFAULT_DEADLINE, max_jobs=1000)
+    @size = size
+    @job_deadline_seconds = job_deadline_seconds
+    @jobs = SizedQueue.new(max_jobs)
+    @mutex = Mutex.new
+    @condition = ConditionVariable.new
+    @running_deadlines = {}
+  end
+
+  def start
+    return false if @jobs.closed?
+
+    # Create worker threads
+    @pool = Array.new(@size) do
+      Thread.new do
+        while (job = @jobs.pop)
+          run_job(job)
+        end
+      end
+    end
+
+    # Start monitor thread
+    @monitor = Thread.new do
+      @mutex.synchronize do
+        while !(@jobs.closed? && @jobs.empty? && @running_deadlines.empty?)
+          wait_time = check_deadlines
+          if wait_time.nil?
+            @condition.wait(@mutex)
+          else
+            @condition.wait(@mutex, wait_time)
+          end
+        end
+      end
+    end
+
+    true
+  end
+
+  def schedule(&block)
+    raise ArgumentError, "Block required" unless block_given?
+    @jobs << block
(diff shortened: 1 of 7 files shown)


lib/deadline_thread_pool.rb
```diff
@@ -0,0 +1,100 @@
...
+    # Start monitor thread
+    @monitor = Thread.new do
+      @mutex.synchronize do
+        while !(@jobs.closed? && @jobs.empty? && @running_deadlines.empty?)
+          wait_time = check_deadlines
+          if wait_time.nil?
+            @condition.wait(@mutex)
+          else
+            @condition.wait(@mutex, wait_time)
+          end
```
jeremyevans, line 36: `wait_time` here is not necessarily maximum amount of time the method will take.  It is the time that the condition variable will wait for the signal.  The condition variable then must reacquire the mutex before returning, and that can block for an arbitrary amount of time.  In this case, probably not long, so this should be safe.

lib/deadline_thread_pool.rb
```diff
@@ -0,0 +1,100 @@
...
+    self
+  end
+
+  def check_deadlines
+    # This .values.min is the most dubiously expensive thing in this.
+    return nil unless (earliest_deadline = @running_deadlines.values.min)
+    now = Time.now
+      
+    # Check if any deadline is passed.
+    Kernel.exit!(1) if earliest_deadline <= now
```
jeremyevans, line 56: I assume we still want to report what strand caused the apoptosis.  How do we do that?

lib/deadline_thread_pool.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  def shutdown
+    return if @jobs.closed?
+
+    # Shut down worker threads, pop will return nil on a closed queue.
+    @jobs.close
+
+    # Wake up monitor thread to shut down, as they check @running.
+    @mutex.synchronize { @condition.signal }
```
jeremyevans, line 69: I recommend creating a `signal` method:

```ruby
def signal
  @mutex.synchronize do
    yield if block_given?
  ensure
    @condition.signal
  end
end
```

And use it here and the 2 calls below.

jeremyevans (commented):
For the most part, this looks OK.  Seems comparable in complexity to the connected threads approach to me.  This implementation does not appear to report which thread caused the apoptosis, which I assume we would want to fix before using it.  I assume performance will be worse for a busy respirate with lots of threads, since every job start and job finish requires the monitor thread to recalculate the then current earliest deadline.  However, as long as the thread pool size isn't too big, I doubt that would cause an issue.  This approach probably uses less memory than the connected threads approach.

lib/deadline_thread_pool.rb
fdr: I suppose min_by to have side effects in the accumulator will do the trick. I also don't set the thread name in this prototype, or integrate with clog. Kinda wanted to get the basic idea there first.

lib/deadline_thread_pool.rb
fdr: yeah ultimately, critical sections have to be short. Not much to be done about that, the same can be said about how queue timed pop has to  protect its data structures. This approach effectively absorbs it.

fdr:
> For the most part, this looks OK. Seems comparable in complexity to the connected threads approach to me. This implementation does not appear to report which thread caused the apoptosis, which I assume we would want to fix before using it. I assume performance will be worse for a busy respirate with lots of threads, since every job start and job finish requires the monitor thread to recalculate the then current earliest deadline. However, as long as the thread pool size isn't too big, I doubt that would cause an issue. This approach probably uses less memory than the connected threads approach.

well, I think I'll finish it then. It would be nice if Ruby had a fast Hash Heap [(rust implementation)](https://docs.rs/hashheap/latest/hashheap/) data structure, that would be ideal, although, a niche use I admit...and really, at less than 1.5us of overhead per job, given our situation, would it really matter?

lib/deadline_thread_pool.rb
fdr: that seems a good idea

fdr:
obsolete

PR #3353, opened 2025-05-23
Change the text for the premium runner toggle

enescakir:
- **Add the ability to change the width in the screenshot script**
  We can already change the window height, but for some screenshots, like
  the dashboard, the width isn't enough, and it makes all the dashboard
  cards look a bit squeezed. It would be better to use a wider window for
  those.
  

- **Change the text for the premium runner toggle**
  

- **Add screenshot for premium runner toggle**
  
| Before | After |
|--------|-------|
| CleanShot 2025-05-23 at 16 26 03@2x | CleanShot 2025-05-23 at 16 25 55@2x |

----

> [!IMPORTANT]
> Update premium runner toggle text and enhance screenshot script for width adjustments.
> 
>   - **Behavior**:
>     - Updated text for premium runner toggle in `views/github/setting.erb`.
>     - Added screenshot for premium runner toggle in `bin/regen-screenshots`.
>   - **Screenshot Script**:
>     - Added width parameter to `resize()` in `bin/regen-screenshots`.
>     - Adjusted window size to 1600x900 for specific screenshots in `bin/regen-screenshots`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for daf82def68b19ff8231cc93c8be8985becaaad60. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bin/regen-screenshots
+++ b/bin/regen-screenshots
@@ -102,6 +102,8 @@ class RegenScreenshots
 
     click_button "Verify Account"
 
+    project = Project.first
+
     resize(900, width: 1600)
     visit "/"
     screenshot "quick-start/managed-services-1-screenshot.png"
@@ -113,7 +115,10 @@ class RegenScreenshots
     click_link "GitHub Runners"
     screenshot "github-actions-integration/quickstart-2-screenshot.png"
 
-    project = Project.first
+    GithubInstallation.create(installation_id: 123, name: "ubicloud-demo", type: "Organization", project_id: project.id, allocator_preferences: {"family_filter" => ["premium", "standard"]})
+    page.refresh
+    click_link "Settings", class: "text-gray-500"
+    screenshot "github-actions-integration/use-premium-runners-1-screenshot.png"
 
     resize(650)
     click_link "Tokens"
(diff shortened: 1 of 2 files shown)


bin/regen-screenshots
```diff
@@ -78,8 +78,8 @@ class RegenScreenshots
     puts "Saved screenshot: #{filename}"
   end
 
-  def resize(height)
-    Capybara.current_session.driver.browser.resize(width: 1200, height:)
+  def resize(height, width = 1200)
```
jeremyevans, line 81: I think it would be easier to understand if `width` was a keyword argument.

jeremyevans (approved):

bin/regen-screenshots
enescakir: What do you think about making both of them keywords?

jeremyevans: I wouldn't be in favor of it, but I'm not opposed if you want to do it.

PR #3357, opened 2025-05-23
Refactor some PrivateSubnet methods

jeremyevans:
Biggest change is using a recursive CTE instead of a recursive ruby method for PrivateSubnet#find_all_connected_nics.

### DIFF
--- a/model/private_subnet.rb
+++ b/model/private_subnet.rb
@@ -28,9 +28,9 @@ class PrivateSubnet < Sequel::Model
   include ObjectTag::Cleanup
 
   def connected_subnets
-    PrivateSubnet.where(
-      id: DB[:connected_subnet].where(subnet_id_1: id).or(subnet_id_2: id).select(Sequel.case({{subnet_id_1: id} => :subnet_id_2}, :subnet_id_1))
-    ).all
+    PrivateSubnet
+      .where(id: DB[:connected_subnet].where(id => [:subnet_id_1, :subnet_id_2]).select(Sequel.case({id => :subnet_id_2}, :subnet_id_1, :subnet_id_1)))
+      .all
   end
 
   def all_nics
--- a/model/private_subnet.rb
+++ b/model/private_subnet.rb
@@ -142,8 +142,15 @@ class PrivateSubnet < Sequel::Model
     end
   end
 
-  def find_all_connected_nics(excluded_private_subnet_ids = [])
-    nics + connected_subnets.select { |subnet| !excluded_private_subnet_ids.include?(subnet.id) }.flat_map { it.find_all_connected_nics(excluded_private_subnet_ids + [id]) }.uniq
+  def find_all_connected_nics
+    Nic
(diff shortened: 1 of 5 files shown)


model/private_subnet.rb
```diff
@@ -142,8 +142,15 @@ def create_tunnels(nics, src_nic)
     end
   end
 
-  def find_all_connected_nics(excluded_private_subnet_ids = [])
-    nics + connected_subnets.select { |subnet| !excluded_private_subnet_ids.include?(subnet.id) }.flat_map { it.find_all_connected_nics(excluded_private_subnet_ids + [id]) }.uniq
```
fdr, line 146: @furkansahin IIRC the general idea we had was that connected subnets would be bilateral only, to avoid tricky questions about routing via a connected subnet. That being the case, why is there a recursive routine in here?

fdr: I did read the blame, which says a lot, but one thing it doesn't quite say (though it says about as much as can be reasonably expected) is any relationship between Nic A and Nic C, which we left as no relationship.

```
    Add connected subnet entity management functions
    
    Here we are adding the methods to be able to find connected_subnets,
    dis/connect 2 subnets, find out all the related nics. The
    find_all_connected_nics function is interesting and needs further
    explanation.
    
    find_all_connected_nics simply find all the nics that are kind of
    related to the subnet we are in. For example, let's assume we have three
    subnets that are connected to each other in this way A<->B<->C. Let's
    also assume that each of these subnets have 1 nic. To be able to
    implement connected subnets, we need to implement IpsecTunnels in
    between nicA<->nicB AND nicB<->nicC. Furthermore, to be able to
    implement tunnel destroys in rekey nic tunnel, we remove old states to
    avoid leaving ip xfrm state objects dangling. Therefore, in a graph, if
    we are performing rekeying for a subset of tunnels, we need to rekey all
    the active tunnels. Otherwise, they will get dropped. For that reason,
    we find all the connected nics by performing depth first search.
    
    Let's also explain what we do to connect_subnet and disconnect_subnet.
    We first create a ConnectedSubnet entity, and then, we create tunnels
    from each nic in the existing subnet to every other nic in the
    connected_subnet. Therefore, if we have 3 nics in the current subnet and
    2 nics in the other, we create 6 tunnels in total. We destroy these
    tunnels at the tume of disconnect_subnet. However, we try to be careful
    not to destroy tunnels that are in between the nics in the same subnet.
```

furkansahin: The recursion is used to resolve the relationship from nic a to nic c. Since the tunnels are created bidirectional, if we rekey the tunnels in between nic a and b, we want to also rekey all of the tunnels touching nic b which include nic c - nic b tunnels. This is needed because the rekey nic tunnel prog drops all of the state objects that are not part of the current rekeying operations. So, if we don't include the tunnel between nic b and nic c into that rekeying batch, that connection is dropped as part of the rekeying of the tunnel nic A-B.

PR #3363, opened 2025-05-25
Use temporary access keys to cache/restore github cache files

velioglu:
**Add migration to add temp access token columns to GitHub repository table**
To add temp access keys in a controlled way, adding a flag as well
to a repository to specify whether the repository will use temp key.

**Use temporary tokens instead of permanent ones to access R2 buckets**
We were using permanent tokens to access R2 buckets. However, we found
that we could only create 1000 permanent API tokens. Given we create
tokens for each customer/bucket, this approach isn't scalable.
Unfortunately, there's no easy way to increase this limit. While an
enterprise account allows for more tokens, upgrading is not practical
for us. We learned from the community that temporary tokens don't have
this restriction [^1] [^2]. Thus, we've decided to switch to temporary
tokens.  Another future benefit is that temporary tokens allow us to set
more specific permissions at the bucket, prefix, and object levels.

We've chosen to use temporary tokens for accessing R2 buckets. Unlike
permanent tokens, this requires sending a session_token with each
request, and we must refresh the token before it expires. I've
implemented refreshing part in the next commit.

[^1]: https://developers.cloudflare.com/r2/api/s3/tokens/#temporary-access-credentials
[^2]: https://developers.cloudflare.com/api/operations/r2-create-temp-access-credentials

**Refresh the repository's blob storage token if has less than 2 days**
We started to create R2 access tokens with a 5-day TTL. It's important
to refresh the token before it expires. 2-day serves as a suitable time
for token refreshment, providing a satisfactory buffer in case of any
issues.

**Create a page if the repository blob token will expire in 1 day**
If the repository nexus is unable to refresh the blob storage token for
any reason, it may fail silently. We generate tokens that last for 5
days and refresh them when there are 2 days left. If the token can't be
refreshed with 1 day remaining, we create a page. The on-call engineer
should investigate the issue before it expires.


enescakir:
Since Cloudflare increased API key limits, we decided to postpone it

PR #3365, opened 2025-05-26
GPU VMs

bsatzger:
image

image

image

---

*Enable creation of GPU-enabled VMs via the UI*

Allows users to create GPU-enabled VMs through the UI. Only available
GPU options are shown - if a specific GPU type is unavailable in a
region, it will not be listed. The option to attach a GPU is visible
only for projects with the `gpu_vm` feature flag enabled.

---

*Enable programmatic creation of GPU-enabled VMs*
    
Allows VMs with GPUs to be created programmatically when the `gpu_vm`
feature flag is enabled. The corresponding UI support will be
implemented in a subsequent PR.

---

*Show gpu field on vm detail card*

When the `gpu_vm` feature flag is set, show the `gpu` field on the
detail card. It shows `-` if no GPU is attached, otherwise it displays
information about the GPU such as `2x NVIDIA A100 80GB PCIe`.

---

*Add GPU configuration to VM serializer*

Add a new field `gpu` to the detailed section of VM. It would contain
`2x NVIDIA A100 80GB PCIe" in case 2 GPUs of device type 20b5
are attached to the VM.

---

*Remove gpu field from Option::VmSize*

The gpu boolean field is unused and lacks expressiveness.
It does not account for:
    * The number of GPUs attached to a VM
    * The specific type of GPU attached

This change removes the field in favor of more precise mechanisms for
representing GPU configurations.

---

*Enable VM assembly with specific gpu_device*
    
The allocator now supports requesting specific GPU devices for a VM.
This commit extends Nexus to include the gpu_device field, allowing it
to assemble VMs with the requested GPU type.

---

*Enhance allocator to recognize GPU device types*
    
Previously, the allocator could not differentiate between GPU types.
With this update, it can now handle requests for specific GPU
models, ensuring that VMs are provisioned with the desired GPU type.

---

*Add gpu_vm feature flag*

Enables GPU-equipped virtual machines for individual projects to support
testing and phased rollout.

---

*Billing rate for A100 GPUs in hetzner-fsn1*

Reference for what `20b5` stands for:
https://github.com/akash-network/provider-configs/blob/98f5bce184a7e45a78c5aa36953e6d813508f435/devices/pcie/gpus.json#L30

----

> [!IMPORTANT]
> Add support for creating and managing GPU-enabled virtual machines with UI and programmatic methods, gated by a feature flag.
> 
>   - **Behavior**:
>     - Enable GPU VM creation via UI and programmatically when `gpu_vm` feature flag is enabled.
>     - Display GPU field on VM detail card if GPU is attached.
>     - Only show available GPU options based on region.
>   - **Validation**:
>     - Add validation for GPU count and type in `validation.rb`.
>     - Ensure GPU is not available for burstable VMs or unsupported types.
>   - **Allocation**:
>     - Update `allocator.rb` to support GPU device requests.
>     - Modify `VmHostAllocation` and `GpuAllocation` to handle GPU resources.
>   - **Serialization**:
>     - Add GPU information to VM serialization in `serializers/vm.rb`.
>   - **UI**:
>     - Update `vm/create.erb` and `vm/show.erb` to include GPU options and display.
>   - **Testing**:
>     - Add tests for GPU VM creation and validation in `vm_spec.rb` and `allocator_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a31fcab1cab4e4de0a3d95dc17066050f4d54444. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


scheduling/allocator.rb
```diff
@@ -233,7 +236,7 @@ def self.candidate_hosts(request)
       ds = ds.where(location_id: request.location_filter) unless request.location_filter.empty?
       ds = ds.where(allocation_state: request.allocation_state_filter) unless request.allocation_state_filter.empty?
       ds = ds.where(Sequel[:vm_host][:family] => request.family_filter) unless request.family_filter.empty?
-      ds = ds.exclude(total_cores: 14, total_cpus: 14) unless request.family == "standard-gpu"
+      ds = ds.exclude(total_cores: 14, total_cpus: 14) unless request.gpu_count > 0
```
enescakir, line 239: Since we have the `family_filter`, we might remove the total_cores:14 filter. No need to maintain it. GPU vm hosts should have `family: standard-gpu` anyway.

prog/vm/nexus.rb
```diff
@@ -115,7 +115,10 @@ def self.assemble(public_key, project_id, name: nil, size: DEFAULT_SIZE,
       ) { it.id = ubid.to_uuid }
       nic.update(vm_id: vm.id)
 
-      gpu_count = 1 if gpu_count == 0 && vm_size.gpu
+      if vm_size.name.start_with?("standard-gpu")
```
enescakir, line 118: We can use `vm_size.family == "standard-gpu"`

openapi/openapi.yml
```diff
@@ -2432,6 +2432,9 @@ components:
                   unix_user:
                     description: Unix user of the VM
                     type: string
+                  gpu:
```
enescakir, line 2438: What if we return an array of objects like this: `gpu: [{"name": "NVIDIA A100 80GB PCIe"}, {"name": "NVIDIA A100 80GB PCIe"}]`?

Can the GPU device provide more information in the future?

views/vm/show.erb
```diff
@@ -20,16 +20,15 @@ viewable_fws =
...
-    data: [
+<% data = [
       ["ID", @vm[:id]],
       ["Name", @vm[:name]],
       ["Location", @vm[:location]],
       ["Size", @vm[:size]],
-      ["Storage", (@vm[:storage_size_gib] > 0) ? "#{@vm[:storage_size_gib]} GB" : nil],
+      ["Storage", (@vm[:storage_size_gib] > 0) ? "#{@vm[:storage_size_gib]} GB" : nil]
+    ]
+  data << ["GPU", @vm[:gpu]] if @project.get_ff_gpu_vm
```
enescakir, line 30: If the VM has any GPU device, IMHO, we can show the GPU info without checking the feature flag.

enescakir (approved):
It looks good to me overall

scheduling/allocator.rb
bsatzger: I need to to make some changes in the allocator anyways and would address this then.

openapi/openapi.yml
bsatzger: I thought about that, too. but right now you can create a vm with up to 8 identical gpus and we'd always return an array with identical items. I agree it's not optimal as is but might require a little more thought altogether.

PR #3379, opened 2025-05-28
Add inline edit functionality for PostgreSQL firewall rules

byucesoy:
**Fix matcher for postgres resource patch endpoint**
Previously, we didn't add `true` keyword, which caused the matcher to match
every patch request, without checking if there are more segments in the path
that needs to be matched. `true` keyword ensures that the matcher matches only
the exact path and not any sub-paths. It didn't cause any problem so far, as we
only had one patch endpoint.

**Reorder firewall description and port range**
In the UI, we used to put static port information before user controlled
description. I moved the description to be before the port range, so all
the user controlled information is grouped together.

**Add edit inline button component**
This commits adds a new button component for inline editing functionality.
When clicked, it converts the text into an editable input field, allowing
users to modify the content directly on the page. Currently it only supportss
text input fields, but can be extended to support other types of inputs such
as select dropdowns or radio buttons in the future.

**Add cancel inline edit button component**
This commit adds a button that allows users to cancel the inline edit operation
they started.

**Add save inline edit button component**
This commit adds a new component for the save the changes made in the inline
edit mode.

**Add ability to inline edit PostgreSQL firewall rules**

chrome-capture-2025-5-28

----

> [!IMPORTANT]
> Adds inline editing functionality for PostgreSQL firewall rules, updates API endpoint matching, and refines UI components and layout.
> 
>   - **Behavior**:
>     - Adds inline editing for PostgreSQL firewall rules in `postgres.rb` and `show.erb`.
>     - Updates `openapi.yml` to include `patchLocationPostgresFirewallRule` for editing firewall rules.
>     - Refines path matcher in `postgres.rb` to use `r.patch true` for exact path matching.
>   - **UI Components**:
>     - Adds `edit_inline_button.erb`, `cancel_inline_button.erb`, and `save_inline_button.erb` for inline editing controls.
>     - Updates `icon.erb` to include `hero-pencil` and `hero-check-circle` icons.
>   - **UI Layout**:
>     - Reorders description and port range in `show.erb` for better user information grouping.
>   - **JavaScript**:
>     - Implements inline edit, cancel, and save functionality in `app.js` for `.edit-inline-btn`, `.cancel-inline-btn`, and `.save-inline-btn`.
>   - **Testing**:
>     - Adds tests for inline editing in `postgres_spec.rb` and `web/project/postgres_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 09cfbd7273770e1c813396d7ea528eb92085eae6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -45,7 +45,7 @@ class Clover
         204
       end
 
-      r.patch do
+      r.patch true do
         authorize("Postgres:edit", pg.id)
 
         target_vm_size = Validation.validate_postgres_size(pg.location, typecast_params.str("size") || pg.target_vm_size, @project.id)
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -135,6 +135,32 @@ class Clover
           end
         end
 
+        r.patch :ubid_uuid do |id|
+          authorize("Postgres:edit", pg.id)
+
+          if (fwr = pg.firewall_rules_dataset[id:])
+            DB.transaction do
+              current_cidr = fwr.cidr.to_s
+              new_cidr = Validation.validate_cidr(typecast_params.nonempty_str!("cidr")).to_s
+
+              fwr.update(
+                cidr: new_cidr,
+                description: typecast_params.str("description")&.strip
+              )
+              pg.incr_update_firewall_rules if current_cidr != new_cidr
+              audit_log(fwr, "update")
+            end
+
+            if api?
+              Serializers::PostgresFirewallRule.serialize(fwr)
+            else
+              204
+            end
+          else
+            404
+          end
+        end
+
         r.delete :ubid_uuid do |id|
           authorize("Postgres:edit", pg.id)
 
--- a/spec/routes/web/project/postgres_spec.rb
+++ b/spec/routes/web/project/postgres_spec.rb
@@ -510,7 +510,7 @@ RSpec.describe Clover, "postgres" do
         pg
         visit "#{project.path}#{pg.path}"
 
-        btn = find "#fwr-delete-#{pg.firewall_rules.first.ubid} .delete-btn"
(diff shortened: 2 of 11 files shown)


routes/project/location/postgres.rb
```diff
@@ -135,6 +135,29 @@ class Clover
...
+        r.patch :ubid_uuid do |id|
+          authorize("Postgres:edit", pg.id)
+
+          if (fwr = pg.firewall_rules_dataset[id:])
+            DB.transaction do
+              fwr.update(
+                cidr: Validation.validate_cidr(typecast_params.nonempty_str!("cidr")).to_s,
+                description: typecast_params.str("description")&.strip
+              )
+              pg.incr_update_firewall_rules
```
furkansahin, line 147: I think we shouldn't update the firewall rules if cidr value didn't change.

spec/routes/web/project/postgres_spec.rb
```diff
@@ -558,6 +558,46 @@
...
+
+        btn = find "#fwr-buttons-#{pg.firewall_rules.first.ubid} .save-inline-btn"
+        url = btn["data-url"]
+        _csrf = btn["data-csrf"]
+        page.driver.submit :patch, url, {cidr: "0.0.0.0/1", description: nil, _csrf:}
+
+        expect(SemSnap.new(pg.id).set?("update_firewall_rules")).to be true
+      end
+
+      it "cannot delete firewall rule if it is not exist" do
```
furkansahin, line 586: if it `doesn't` exist

furkansahin (approved):
I don't know about the javascript part but the rest LGTM. I'd be happy if you could address one comment

PR #3385, opened 2025-05-29
Fix icon.erb ignoring for erb-formatter

enescakir:
- **Fix icon.erb ignoring for erb-formatter**
  
  "views/inference/endpoint/index.erb" was ignored due to this issue. I
  removed it from the new list becasue it has multiline string and
  erb-formatter has bug for them.
  

- **Format mistakenly ignored files with erb-formatter**
  As expected all ERB files starting with the letters 'i', 'c', 'o', 'n'
  are formatted.
  

----

> [!IMPORTANT]
> Fixes erb-formatter file ignoring pattern to only exclude `icon.erb`, ensuring correct formatting of other files.
> 
>   - **Fix icon.erb ignoring for erb-formatter**
>     - Corrects file ignoring pattern in `Rakefile` to only exclude `views/components/icon.erb`.
>     - Previously, all files starting with 'i', 'c', 'o', 'n' were ignored.
>     - Excludes `views/inference/endpoint/index.erb` due to erb-formatter bug with multiline strings.
>   - **Format mistakenly ignored files with erb-formatter**
>     - Formats all ERB files starting with 'i', 'c', 'o', 'n' except those explicitly excluded.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for fb6e22fe7d79649490d949a56f08b23c8cfd7e2d. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/views/account/change_password.erb
+++ b/views/account/change_password.erb
@@ -10,7 +10,8 @@
         <%== render("account/submenu") %>
         <div class="divide-y divide-gray-200 lg:col-span-8 xl:col-span-9 2xl:col-span-10 pb-10">
           <div class="px-4 py-6 sm:p-6 lg:pb-8 space-y-4">
-            <h2 class="text-lg font-medium leading-6 text-gray-900"><%= action %> Password</h2>
+            <h2 class="text-lg font-medium leading-6 text-gray-900"><%= action %>
+              Password</h2>
             <form action="/<%= rodauth.change_password_route %>" role="form" method="POST">
               <%== rodauth.change_password_additional_form_tags %>
               <%== rodauth.csrf_tag("/" + rodauth.change_password_route) %>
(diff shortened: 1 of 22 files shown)


jeremyevans (approved):
Looking at the template changes, I'm not sure that this is an improvement, but unless we want to stop using erb-formatter, we should make this change.

byucesoy:
> Looking at the template changes, I'm not sure that this is an improvement, but unless we want to stop using erb-formatter, we should make this change.

Agreed.

byucesoy (approved):

enescakir:
I agree that they aren't improvements, but it's good that we're consistent across different files.

Unfortunately, `erb-formatter` doesn't have an alternative.
-  `erb_lint` is popular and well-maintained, but it's a linter and doesn't format files automatically. https://github.com/Shopify/erb_lint
-  The prettier plugin is not maintained. https://github.com/adamzapasnik/prettier-plugin-erb

enescakir:
Since we’re unhappy with its results and are planning to replace it with the herb formatter, I’m closing this PR.

PR #3390, opened 2025-05-29
Second phase of route cleanups

jeremyevans:
These contain the more complex changes, most of which are related to N+1 query reduction.

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -8,6 +8,7 @@ class GithubRunner < Sequel::Model
   many_to_one :installation, key: :installation_id, class: :GithubInstallation
   many_to_one :repository, key: :repository_id, class: :GithubRepository
   one_to_one :vm, key: :id, primary_key: :vm_id
+  one_through_one :project, join_table: :github_installation, left_key: :id, left_primary_key: :installation_id, read_only: true
 
   include ResourceMethods
   include SemaphoreMethods
--- a/routes/project/location/private_subnet.rb
+++ b/routes/project/location/private_subnet.rb
@@ -84,7 +84,18 @@ class Clover
 
         r.delete do
           authorize("PrivateSubnet:delete", ps.id)
-          unless ps.vms.all? { it.destroy_set? || it.strand.nil? || it.strand.label == "destroy" }
+
+          vms_dataset = ps.vms_dataset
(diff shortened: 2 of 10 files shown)


routes/project/location/private_subnet.rb
```diff
@@ -84,7 +84,18 @@ class Clover
...
         r.delete do
           authorize("PrivateSubnet:delete", ps.id)
-          unless ps.vms.all? { it.destroy_set? || it.strand.nil? || it.strand.label == "destroy" }
+
+          vms_dataset = ps.vms_dataset
+            .association_join(:strand)
+            .exclude(label: "destroy")
+            .exclude(Sequel[:vm][:id] => Semaphore
+              .where(
+                strand_id: DB[:nic].where(private_subnet_id: ps.id).select(:vm_id),
```
enescakir, line 93: Why don't we use `ps.nics_dataset` directly?

jeremyevans: That makes sense. As there is a PR depending on this one (already approved), I think I'll wait until around merging time to change over.

jeremyevans:
I think before running E2E tests, this needs to be rebased against main.  I know other PRs I had that were newer than this one required rebasing against main before E2E would pass.  I'll rebase these three PRs against main today, then run E2E on them.

model/github_runner.rb
```diff
@@ -8,6 +8,7 @@ class GithubRunner < Sequel::Model
   many_to_one :installation, key: :installation_id, class: :GithubInstallation
   many_to_one :repository, key: :repository_id, class: :GithubRepository
   one_to_one :vm, key: :id, primary_key: :vm_id
+  one_through_one :project, join_table: :github_installation, left_key: :id, left_primary_key: :installation_id, read_only: true
```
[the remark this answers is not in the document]
jeremyevans: `github_installation.installation_id` is an integer for the GitHub ID, not a uuid for the Ubicloud ID, so this makes no sense. 

The relationship is `github_runner.installation_id`<->`github_installation.id` and `github_installation.project_id`<->`project.id`.  As the right keys are Sequel's defaults, neither `right_key` nor `right_primary_key` need to be specified.

furkansahin (approved):

PR #3392, opened 2025-05-29
Add respirate smoke test

jeremyevans:
Also, always use verbose logging in tests for Strand#take_lease_and_reload , and always call yield to   Clog.emit block in the tests.

----

> [!IMPORTANT]
> Add `respirate` smoke test and CI workflow, ensure verbose logging in tests, and modify `Clog.emit` behavior in tests.
> 
>   - **Behavior**:
>     - Add `respirate` smoke test in `spec/respirate_smoke_test.rb` to verify processing of 1000 strands within 30 seconds.
>     - Always use verbose logging in tests for `Strand#take_lease_and_reload` in `model/strand.rb`.
>     - Ensure `Clog.emit` calls original method in tests across multiple specs.
>   - **CI/CD**:
>     - Add `respirate-ci.yml` GitHub Actions workflow to run `respirate` smoke tests on `main` branch and specific paths.
>   - **Rake Tasks**:
>     - Add `respirate_smoke_test` task to `Rakefile` to execute the smoke test script.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 088bfbf77b9e7f97b34c53a70a48e579bbc48caf. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/prog/vnet/cert_nexus_spec.rb
+++ b/spec/prog/vnet/cert_nexus_spec.rb
@@ -114,7 +114,7 @@ RSpec.describe Prog::Vnet::CertNexus do
 
     it "hops to restart if dns_challenge validation fails" do
       expect(challenge).to receive(:status).and_return("failed")
-      expect(Clog).to receive(:emit).with("DNS validation failed")
+      expect(Clog).to receive(:emit).with("DNS validation failed").and_call_original
       expect(dns_zone).to receive(:delete_record).with(record_name: "test-record-name.cert-hostname")
       expect(nx).to receive(:dns_zone).and_return(dns_zone)
       expect { nx.wait_dns_validation }.to hop("restart")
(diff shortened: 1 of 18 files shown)


fdr (approved):

spec/prog/vnet/cert_nexus_spec.rb
```diff
@@ -114,7 +114,7 @@
 
     it "hops to restart if dns_challenge validation fails" do
       expect(challenge).to receive(:status).and_return("failed")
-      expect(Clog).to receive(:emit).with("DNS validation failed")
+      expect(Clog).to receive(:emit).with("DNS validation failed").and_call_original
```
fdr, line 117: I wonder if we can use an rspec extension to somehow make this a do-not-forget type of item (or, I guess, a rubocop autocorrect)

jeremyevans: I would prefer to change `Clog.emit` to not use a block, and instead use a regular argument.  The block approach only makes sense if you were conditionally calling the block, but since `Clog.emit` always calls the block if given, using a block is only cost and no benefit.

PR #3399, opened 2025-05-31
Unify clover workflow environment setup

enescakir:


jeremyevans (approved):
This is awesome. It makes the workflow files so much nicer.

PR #3402, opened 2025-06-01
Prep for Postgres UI Revamp

byucesoy:
Soon, I'll open a pretty big PR for Postgres UI. I extracted bunch of self-contained commits from that work to make that PR smaller and easier to review.

**Use value of existing ha_type if ha_type is not passed for updates**
This used to reset the ha_type to none if it was not passed in the request,
which was not the desired behavior. For updates, we should use the existing
ha_type value if it is not provided in the request.

**Use divs instead of spans in some components**
This is bit of a cosmetic change without any functional impact. span elements
are usually used for simple, inline text styling, while div elements are
typically used for block-level elements that can contain other elements. For
these components, using divs instead of spans makes more sense as they contain
more complex structures and multiple levels of sub-elements.

**Restructure the revealable_content divs for better control over content**
In upcoming PG UI changes, we will need to edit the specific parts of the
revealable_content divs without impacting the show/hide icons. This commit
restructures the divs to allow for more granular control over the content,
making it easier to update connection information without affecting the
entire content.

**Implement a workaround for the incorrect overflow calculations**
For some reason, adding `overflow-wrap: anywhere` style directly to the element
does not trigger layout reflow for handling overflow when the subject is made
visible. However, adding the overflow-wrap dynamically makes the browser
recalculate the layout and apply the style correctly.

**Use pathname instead of entire URL while deciding on metrics url**
Target endpoint for metrics url is now determined by the pathname of the URL
instead of the entire URL. This is because the entire URL can contain query
parameters that are not relevant for the metrics endpoint. Using the entire URL
generates malformed URLs in some cases, such as when the URL contains query
parameters as well.

**Use grid-template-columns for radio small card**
Previously, we used media queries to set the number of columns. However, media
columns are based on entire viewport width. When we start to use more of the
viewport area, for example with a second sidebar for submenus, the grid
columns will not be correct anymore. We can change the numbers we set in the
grid columns to accomodate new remaining space, but that would break the places
where there is no second sidebar.

The solution is to use grid-template-columns instead of media queries, because
they are based on the width of the container.

**Add port and database name to connection string**
This is to be more explicit about the information in the connection string.
With the addition of pgBouncer, we need to specify the port anyway, which
leaves only the database name to be optional. For the sake of consistency,
I decided to include it in the connection string as well.

----

> [!IMPORTANT]
> Refactor Postgres UI and backend with improved `ha_type` handling, connection strings, and UI component structures.
> 
>   - **Behavior**:
>     - Use existing `ha_type` if not provided in update requests in `postgres.rb`.
>     - Add port and database name to connection strings in `postgres_resource.rb`.
>     - Determine metrics URL using pathname in `app.js`.
>   - **UI Components**:
>     - Replace `span` with `div` in `copyable_content.erb` and `revealable_content.erb` for better structure.
>     - Use `grid-template-columns` in `radio_small_cards.erb` for responsive design.
>     - Restructure `revealable_content` for better control over content visibility.
>   - **CSS and Layout**:
>     - Implement dynamic `overflow-wrap` to fix layout reflow issues in `app.js`.
>   - **Tests**:
>     - Update tests in `postgres_resource_spec.rb` and `pg_dumpall_spec.rb` to reflect connection string changes.
>     - Modify golden files to include updated connection strings.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 373a8299458c1a8ca309cb2d41c2ad219ec1ab41. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


enescakir (approved):

enescakir:
https://github.com/ubicloud/ubicloud/pull/3409

PR #3404, opened 2025-06-02
Use cert_store in VictoriaMetrics client

mohi-kalantari:
Previously we were relying on the ssl_ca_file api of Excon which seems that it couldn't parse a chain of certs. In order to fix that we will parse the certs and add them all to a cert_store and then pass that to the client which fixes the issue.

### DIFF
--- a/lib/victoria_metrics/client.rb
+++ b/lib/victoria_metrics/client.rb
@@ -7,14 +7,18 @@ require "digest"
 require "zlib"
 
 class VictoriaMetrics::Client
-  def initialize(endpoint:, ssl_ca_file_data: nil, socket: nil, username: nil, password: nil)
+  def initialize(endpoint:, ssl_ca_data: nil, socket: nil, username: nil, password: nil)
     @endpoint = endpoint
     @username = username
     @password = password
-    @client = if ssl_ca_file_data
-      ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", Digest::SHA256.hexdigest(ssl_ca_file_data) + ".crt")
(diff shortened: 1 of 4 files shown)


lib/victoria_metrics/client.rb
```diff
@@ -7,14 +7,18 @@
...
     @username = username
     @password = password
-    @client = if ssl_ca_file_data
-      ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", Digest::SHA256.hexdigest(ssl_ca_file_data) + ".crt")
-      Util.safe_write_to_file(ca_bundle_filename, ssl_ca_file_data) unless File.exist?(ca_bundle_filename)
-      Excon.new(endpoint, socket: socket, ssl_ca_file: ca_bundle_filename)
+    @client = if ssl_ca_data
+      cert_store = OpenSSL::X509::Store.new
+      certs_pem = ssl_ca_data.scan(/-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----/m)
+      certs_pem.each do |cert_pem|
```
shikharbhardwaj, line 17: Could we do `certs = OpenSSL::X509::Certificate.load_file(ca_bundle_filename)` here instead of parsing the file via regex ourselves?

mohi-kalantari: It didn't feel quite right to me to write a temp file and then `load_file` it because we can keep it all in memory.

shikharbhardwaj: makes sense, mainly want to avoid regex but this looks like a standard pattern

shikharbhardwaj (approved):
Thanks for fixing this!

PR #3405, opened 2025-06-02
Configure hugepages for Postgres

shikharbhardwaj:
This commit introduces a new label to postgres_server_nexus.rb to configure huge pages, via a rhizome lib+bin.

At the moment, it uses the default huge page size (2M), and calculates the number of huge pages using `postgres ... -C
shared_memory_size_in_huge_pages`. Since configuring hugepages needs to restart postgres, its placed in a separate label from `configure`, to allow other configuration changes to go through without restarting postgres.

In a subsequent change, we can consider adding the boot parameters to setup 1G hugepages for larger instances (possibly standard-30+), where they are most useful.

[1]: https://www.postgresql.org/docs/current/kernel-resources.html#LINUX-HUGE-PAGES

----

> [!IMPORTANT]
> Adds hugepages configuration for Postgres with a new setup script and class, including tests for the setup logic.
> 
>   - **Behavior**:
>     - Adds hugepages configuration to `configure` script in `rhizome/postgres/bin/configure`.
>     - New script `configure-hugepages` to handle hugepages setup.
>     - Calculates hugepages based on system memory and Postgres parameters.
>     - Logs success or failure of hugepages allocation.
>   - **Classes**:
>     - New `HugepagesSetup` class in `hugepages_setup.rb` to manage hugepages configuration.
>     - Methods include `get_postgres_param`, `hugepage_info`, `setup_postgres_hugepages`, and `update_postgres_hugepages_conf`.
>   - **Tests**:
>     - Adds `hugepages_setup_spec.rb` to test `HugepagesSetup` class.
>     - Tests hugepages calculation and setup logic, including edge cases with no hugepages configured.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 93dcb961a2e2eb049281f37cb0ecf43050da15a8. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/postgres/bin/configure
+++ b/rhizome/postgres/bin/configure
@@ -98,6 +98,39 @@ standby2replication   #{identity}             ubi_replication
 PG_IDENT
 safe_write_to_file("/etc/postgresql/#{v}/main/pg_ident.conf", pg_ident_entries)
 
+# Configure system hugepages
+meminfo = File.read("/proc/meminfo")
+hugepage_size_kib = Integer(meminfo[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)
+memory_total_kib = Integer(meminfo[/^MemTotal:\s*(\d+)\s*kB/, 1], 10)
+
+target_hugepages_size_kib = memory_total_kib / 4
+target_hugepages = target_hugepages_size_kib / hugepage_size_kib
+
+r "echo 'vm.nr_hugepages = #{target_hugepages}' | sudo tee /etc/sysctl.d/10-hugepages.conf"
+r "sync"
+r "echo 3 | sudo tee /proc/sys/vm/drop_caches"
+r "echo 1 | sudo tee /proc/sys/vm/compact_memory"
+r "sudo sysctl --system"
+
+meminfo = File.read("/proc/meminfo")
+allocated_hugepages = Integer(meminfo[/^HugePages_Total:\s*(\d+)/, 1], 10)
+
+if allocated_hugepages < target_hugepages
+  puts "Failed to allocate #{target_hugepages} hugepages. Only #{allocated_hugepages} were allocated."
+  exit 1
+else
+  puts "Successfully allocated #{allocated_hugepages} hugepages."
+end
+
+# Create drop-in directory for postgresql@.service to override
+# the default systemd service file with a custom ExecStartPre to configure hugepages.
+r "mkdir -p /etc/systemd/system/postgresql@.service.d"
+execstartpre_hugepages_conf = <<-HUGEPAGES_CONF
+[Service]
+ExecStartPre=/home/ubi/postgres/bin/configure-hugepages %i
+HUGEPAGES_CONF
+safe_write_to_file("/etc/systemd/system/postgresql@.service.d/hugepages.conf", execstartpre_hugepages_conf)
+
 # Reload the postmaster to apply changes
 r "pg_ctlcluster #{v} main reload || pg_ctlcluster #{v} main restart"
 
--- a/rhizome/postgres/bin/configure-hugepages
+++ b/rhizome/postgres/bin/configure-hugepages
@@ -1,0 +1,17 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require_relative "../lib/hugepages_setup"
+require "logger"
+
+if ARGV.count != 1
+  fail "Wrong number of arguments. Expected 1, Given #{ARGV.count}"
+end
+
+instance = ARGV[0]
+
+logger = Logger.new($stdout)
+logger.level = Logger::INFO
+
+HugepagesSetup.new(instance, logger).setup
--- a/rhizome/postgres/lib/hugepages_setup.rb
+++ b/rhizome/postgres/lib/hugepages_setup.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require "logger"
+
+class HugepagesSetup
+  def initialize(instance, logger)
+    @version, @cluster = instance.split("-", 2)
+    @logger = logger
+  end
+
+  def get_postgres_param(param)
+    Integer(r("sudo -u postgres /usr/lib/postgresql/#{@version.shellescape}/bin/postgres -D /dat/#{@version.shellescape}/data -c config_file=/etc/postgresql/#{@version.shellescape}/#{@cluster.shellescape}/postgresql.conf -C #{param.shellescape}"), 10)
+  end
+
+  def hugepage_info
+    meminfo = File.read("/proc/meminfo")
+    hugepages_count = Integer(meminfo[/^HugePages_Total:\s*(\d+)/, 1], 10)
+    hugepage_size_kib = Integer(meminfo[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)
+    [hugepages_count, hugepage_size_kib]
+  end
+
+  def stop_postgres_cluster
+    r "sudo pg_ctlcluster stop #{@version} #{@cluster}", expect: [0, 2] # 2 is "not running"
+  end
+
+  def setup_postgres_hugepages
+    hugepages_count, hugepage_size_kib = hugepage_info
+
+    if hugepages_count == 0
+      @logger.warn("No hugepages configured, skipping setup.")
+      return
+    end
+
+    hugepages_kib = hugepages_count * hugepage_size_kib
+    update_postgres_hugepages_conf(hugepages_kib)
+
+    shmem_and_overhead_kib = 1024 * get_postgres_param("shared_memory_size")
+    overhead = shmem_and_overhead_kib - hugepages_kib
+
+    target_kib = hugepages_kib - overhead
+
+    # Floor division to nearest multiple of block_size: (a / b) * b
+    block_size_bytes = get_postgres_param("block_size")
+    block_size_kib = block_size_bytes / 1024
+    final_shared_buffers_kib = (target_kib / block_size_kib) * block_size_kib
+
+    update_postgres_hugepages_conf(final_shared_buffers_kib)
+  end
+
+  def update_postgres_hugepages_conf(shared_buffers_kib)
+    safe_write_to_file("/etc/postgresql/#{@version}/#{@cluster}/conf.d/002-hugepages.conf", <<CONF
+huge_pages = 'on'
+huge_page_size = 0
+shared_buffers = #{shared_buffers_kib}kB
+CONF
+    )
+  end
+
+  def postgres_running?
+    r "sudo pg_ctlcluster status #{@version} #{@cluster}", expect: [3]
+    false
+  rescue CommandFail
+    true
+  end
+
+  def setup
+    unless postgres_running?
+      stop_postgres_cluster
+      setup_postgres_hugepages
+    end
+  end
+end
--- a/rhizome/postgres/spec/hugepages_setup_spec.rb
+++ b/rhizome/postgres/spec/hugepages_setup_spec.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+require_relative "../lib/hugepages_setup"
+
+RSpec.describe HugepagesSetup do
+  let(:logger) { instance_double(Logger, warn: nil) }
+  let(:hugepages_setup) { described_class.new("17-main", logger) }
+
+  describe "#setup_postgres_hugepages" do
+    it "calculates hugepages via overhead subtraction and rounding" do
+      # Mock hugepage info: 512 x 2MB hugepages = 1024MB total hugepage space
+      expect(hugepages_setup).to receive(:hugepage_info).and_return([512, 2048])
+
+      # shared_memory_size increases due to simulated overhead:
+      # 1024MB -> 1061MB (37MB overhead)
+      expect(hugepages_setup).to receive(:get_postgres_param)
+        .with("shared_memory_size").and_return(1061)
+
+      # Block size for rounding shared_buffers
+      expect(hugepages_setup).to receive(:get_postgres_param)
+        .with("block_size").and_return(8192)  # 8KB blocks
+
+      # First call: set shared_buffers to total hugepage space
+      # (1024MB = 1,048,576 KiB)
+      expect(hugepages_setup).to receive(:update_postgres_hugepages_conf)
+        .with(1_048_576)
+
+      # Second call: back off by overhead and round down to block boundary
+      # overhead = (1061 - 1024) * 1024 = 37,888 KiB
+      # target = 1,048,576 - 37,888 = 1,010,688 KiB
+      expected_shared_buffers = 1_010_688
+      expect(hugepages_setup).to receive(:update_postgres_hugepages_conf)
+        .with(expected_shared_buffers)
+      expect(expected_shared_buffers & 0b111).to eq(0)  # Divisible by 8
+
+      expect { hugepages_setup.setup_postgres_hugepages }.not_to raise_error
+    end
+
+    it "skips setup if no hugepages are configured" do
+      expect(hugepages_setup).to receive(:hugepage_info).and_return([0, 2048])
+      expect(hugepages_setup).not_to receive(:get_postgres_param)
+      expect(hugepages_setup).not_to receive(:update_postgres_hugepages_conf)
+      expect(logger).to receive(:warn).with("No hugepages configured, skipping setup.")
+      expect { hugepages_setup.setup_postgres_hugepages }.not_to raise_error
+    end
+  end
+end


model/postgres/postgres_server.rb
```diff
@@ -19,7 +19,7 @@ class PostgresServer < Sequel::Model
   include MetricsTargetMethods
 
   semaphore :initial_provisioning, :refresh_certificates, :update_superuser_password, :checkup
-  semaphore :restart, :configure, :take_over, :configure_prometheus, :configure_metrics, :destroy, :recycle, :promote
+  semaphore :restart, :configure, :take_over, :configure_metrics, :configure_hugepages, :destroy, :recycle, :promote
```
shikharbhardwaj: this is a bug from a previous change, I will push this in a separate PR/commit

prog/postgres/postgres_server_nexus.rb
```diff
@@ -295,6 +293,25 @@ def before_run
     nap 5
   end
 
+  label def configure_hugepages
```
fdr: This is a way, to push so much to clover (also, use daemonizer2 and helper symbols if you are going to do that)

One that may work just as well or better is to compute the shared memory size with every `systemctl restart` with a `ExecStartPre`. This is robust to slight fluctuations that crop up with updating extension versions or tweaking settings, otherwise you have to haul in this whole loop again for what is some basic arithmetic. The result is something like, "I'm doing some simple work and systemctl restart doesn't work and then I gotta haul in some semaphores to make this arithmetic work"

The downside is every restart is slightly slower for running the extra exec of Postgres emitting the hugepage number, even in the event it doesn't change

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,39 @@
...
+require_relative "../../common/lib/util"
+
+class HugepagesSetup
+  def initialize(version)
+    @version = version
+  end
+
+  def get_shared_memory_size_in_hugepages
+    r "sudo pg_ctlcluster stop #{@version} main", expect: [0, 2] # 2 is "not running"
+    huge_pages_count = r "sudo -u postgres /usr/lib/postgresql/#{@version}/bin/postgres -D /dat/#{@version}/data -c config_file=/etc/postgresql/#{@version}/main/postgresql.conf -C shared_memory_size_in_huge_pages"
```
fdr: IMO, do your own rounding from `shared_memory_size` rather than "in huge pages" because it's 1) simple and 2) more robust to 1G hugepages. I mean, who wants to be told "yeah you need 11 hugepage" when the last one is 1% full, and you could have allocated 995MB more shared buffers?. "in huge pages" is basically only useful for smaller hugepage size as a minor convenience. 

For bonus points, you can computed shared_buffers from the remaining round-off space. 1.2 megabytes (or whatever) of shared_buffers is 1.2 megabytes of shared_buffers, after all.

shikharbhardwaj: Do you mean something like this? In this patch we are using the default 2M hugepages, so I guess the last optimization would not help much, but it is pretty meaningful with 1G hugepages for small instances. I was thinking to introduce 1G hugepages for instances larger than a given size (let's say standard-16 and above), wdyt?

```
huge_page_size_mib = 2 # or 1024 for 2G hugepages
shared_memory_size_mib = r "sudo -u postgres /usr/lib/postgresql/#{@version}/bin/postgres -D /dat/#{@version}/data -c config_file=/etc/postgresql/#{@version}/main/postgresql.conf -C shared_memory_size"
huge_pages = (shared_memory_size_mib.to_f / huge_page_size_mib)
huge_pages_full = huge_pages.ceil
additional_shared_buffers = (huge_pages_full - huge_pages) * huge_page_size_mib

# Add `additional_shared_buffers` to `shared_buffers` config, set nr_hugepages to `huge_pages_full`
```

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: I think this should be cleaner, will try this out

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,86 @@
...
+    shared_buffers_blocks * block_size_bytes / 1024
+  end
+
+  def stop_postgres_cluster
+    r "sudo pg_ctlcluster stop #{@version} main", expect: [0, 2] # 2 is "not running"
+  end
+
+  def hugepages_config
+    stop_postgres_cluster
+    shared_memory_size_kib = 1024 * get_postgres_param("shared_memory_size").to_f
```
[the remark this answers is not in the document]
shikharbhardwaj: it's only called once, in either of the if/else branch, so does not make much sense to cache this

rhizome/postgres/lib/hugepages_setup.rb
fdr: What you wrote is basically what I mean, that is: having Postgres do the round-off, if you do it yourself with not-much code, you do get more of the 1G version for free, sparing yourself a ungratifying migration of approach. 

Your plan to use 2MB hugepages on small instances is sound, and has this advantage: it's hard to get so much contiguous memory on a small instance.

That said, I have also used 1GB pages on very small instances in the past, but not out of the interest of being quite so tightfisted with TLB entries, but rather, to have fewer branches in the code. Some background on that:

I wouldn't call my previous decision there, to use 1GB hugepages on small VMs, a slam dunk on simplification: getting, say, 25% of contiguous memory -- let's say 1GB of 4GB -- is tougher than, say, 2GB of 8GB, and then again at 4GB of 16GB: the fixed costs of starting the operating system and their effect on fragmentation are more intrusive in the smaller cases. Your hybrid approach can almost certainly use the more humane sysctl interface, which tries to get the pages pretty early, but not as early as the kernel command line in grub.

The other area that the hybrid approach may be superior is reboot speed. Linux can move some stuff around to get a contiguous page, and may be to do so desperately -- and slowly -- in the 1GB-in-4GB type of case. that is, it barely is able to make it, and spends a lot of time doing so. You can see this phenomena with how slow our VM hosts are to reboot, as they try to get as many hugepages as possible (if you try turning off hugepages, you should notice `systemctl reboot` is much faster).

Anyway, I tell you this because it embeds some heuristics about how to adjust your approach: if you find yourself not getting the hugepages you need, maybe you need to go to grub.conf or use the smaller hugepages, or find some way to allocate the hugepages earlier in boot (not sure how much earlier you can get than sysctl, but maybe there's a way)

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,86 @@
...
+
+require_relative "../../common/lib/util"
+
+class HugepagesSetup
+  def initialize(instance)
+    @version, @cluster = instance.split("-", 2)
+  end
+
+  def hugepage_size_kib
+    (r 'awk \'/Hugepagesize/ {printf "%.2f\n", $2}\' /proc/meminfo').to_f
```
fdr:
```suggestion
    r('awk \'/Hugepagesize/ {printf "%.2f\n", $2}\' /proc/meminfo').to_f
```

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def get_postgres_param(param)
+    r "sudo -u postgres /usr/lib/postgresql/#{@version}/bin/postgres -D /dat/#{@version}/data -c config_file=/etc/postgresql/#{@version}/#{@cluster}/postgresql.conf -C #{param}"
```
fdr:
```suggestion
    r "sudo -u postgres /usr/lib/postgresql/#{@version}/bin/postgres -D /dat/#{@version}/data -c config_file=/etc/postgresql/#{@version}/#{@cluster}/postgresql.conf -C #{param.shellescape}"
```

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,86 @@
...
+    if hugepages_fractional > 0.5
+      additional_hugepages = 1 - hugepages_fractional
+      additional_shared_buffers_kib = (additional_hugepages * hugepage_size_kib).floor
+
+      {
+        hugepages_count: hugepages_count.ceil.to_i,
+        shared_buffers_kib: (shared_buffers_kib + additional_shared_buffers_kib).to_i
+      }
+    else
+      excess_shared_buffers_kib = (hugepages_fractional * hugepage_size_kib).ceil
```
fdr: You shouldn't need to do anything that complicated (50%, whatever). Whatever memory is left in the last hugepage, use it for shared_buffers. You may or may not need a bit of slop for padding, maybe try with no slop.

fdr: I think this algorithm in general, though, can use a little work. I think the main problem is path dependency/determinism in the `shared_memory_size` calculation. I don't see you resetting it (did I miss it?) to do your calculations afresh on every restart:

If memory serves, a large `shared_buffers` can impose some additional shared memory use beyond the number you quoted. The way you can confirm this is by taking a large-ish system, try setting `shared_buffers` to something tiny, assessing `shared_memory_size` and then using subtraction to figure out the overhead. Then, repeat, after setting the shared buffers to something large. If large `shared_memory_size` introduces a larger overhead after subtracting,  you can see the problem: the `shared_buffer` setting is best placed close to its target value to get an accurate estimation, alas, you can't use simple addition alone.

The way I would write the algorithm is:

1. Sum up the bytes in all the hugepages. The mental model is: I have X memory in hugepages, I want to back-calculate a fitting shared_buffers, rather than, "I want to round up"
2. Set `shared_buffers` to _exactly_ the number of bytes in those hugepages. By definition, a server configured this way cannot start, since there is overhead atop `shared_buffers`. But, it's a simple, deterministic source of `shared_buffers` number "close" to the target, and for the reasons mentioned above, you need something "close" to get an accurate estimate. Though the server cannot start, the `-C` option to echo `shared_memory_size` will still work.
3. Get the `shared_memory_size`, which by definition, will be larger than the memory in the hugepages
4. Subtract from `shared_buffers` the excess. This requires no slop factor, because a smaller value for `shared_buffers` will, at worst, require even less memory than the original projection (e.g. if you cross a discontinuity that allows a data structure to be shrunk). This residual memory that maybe _could_ have been `shared_buffers` we make a sacrificial offering to the gods of simplicity.

shikharbhardwaj: Makes sense, I just checked this with a `standard-16` instance and with 128M vs 16G shared buffers, there's a difference of about 289M in addition to the increase in shared_buffers. I'll update this.

rhizome/postgres/bin/configure-hugepages
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
shikharbhardwaj: This is not correct, chmod +x yields no diff for this file

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,80 @@
...
+
+require_relative "../../common/lib/util"
+
+class HugepagesSetup
+  def initialize(instance)
+    @version, @cluster = instance.split("-", 2)
+  end
+
+  def hugepage_size_kib
+    r('awk \'/Hugepagesize/ {printf "%.2f\n", $2}\' /proc/meminfo').to_f
```
fdr: Weird: subprocess? awk? floating point?

```suggestion
    Integer(File.read('/proc/meminfo')[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)
```

fdr: You can test with:

```sh
$ ruby -e "puts Integer(File.read('/proc/meminfo')[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)"
2048
```

On a system with 1GB hugpages:

```sh
$ ruby -e  ruby -e "puts Integer(File.read('/proc/meminfo')[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)"
1048576
```

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def get_postgres_param(param)
+    r "sudo -u postgres /usr/lib/postgresql/#{@version}/bin/postgres -D /dat/#{@version}/data -c config_file=/etc/postgresql/#{@version}/#{@cluster}/postgresql.conf -C #{param.shellescape}"
+  end
+
+  def shared_buffers_kib
+    shared_buffers_blocks = get_postgres_param("shared_buffers").to_f
```
fdr: why rely on floats so much? these are all integral, right?

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,80 @@
...
+    block_size_bytes = get_postgres_param("block_size").to_f
+
+    shared_buffers_blocks * block_size_bytes / 1024
```
fdr: I suppose it's to prevent integer flooring. Unless it's especially pervasive throughout the program, the way to do that is `fdiv` on integers.
```suggestion
    (shared_buffers_blocks * block_size_bytes).fdiv(1024)
```

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def stop_postgres_cluster
+    r "sudo pg_ctlcluster stop #{@version} #{@cluster}", expect: [0, 2] # 2 is "not running"
+  end
+
+  def hugepages_count
+    @hugepages_count ||= begin
+      shared_memory_size_kib = 1024 * get_postgres_param("shared_memory_size").to_f
+      (shared_memory_size_kib / hugepage_size_kib).ceil
```
fdr: This is a trick for integral ceiling division that I have to look up every time. Granted, AI does make it faster to retrieve...but not much faster, given "round up integer division" does the trick via old fashion text search. But I always know to reach for it :)

```suggestion
      (shared_memory_size_kib + hugepage_size_kib - 1) / hugepage_size_kib
```

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+  end
+
+  def setup_postgres_hugepages
+    shared_memory_size_kib = 1024 * get_postgres_param("shared_memory_size").to_f
+    hugepages_kib = hugepages_count * hugepage_size_kib
+    update_postgres_hugepages_conf(hugepages_kib.ceil)
```
fdr: This is puzzling, memory size kib, hugepages kib, they are all integral, so all multiplications are likewise integral, so why the ceil? rounding errors? If so, it's related to over-application of floating point.

fdr:
I submitted as a fixup -- don't forget to use autosquash if you like the changes! -- my proposals to simplify, mostly by taking advantage the fact that shared_memory_size is a MB quantity, thus implying certain alignments and divisibility. In doing so, even fdiv can be eliminated. The algorithm itself gets a tiny bit shorter, but I blow up the test a bit with comments.

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -8,18 +8,11 @@ def initialize(instance)
...
   def hugepage_size_kib
-    r('awk \'/Hugepagesize/ {printf "%.2f\n", $2}\' /proc/meminfo').to_f
+    Integer(File.read("/proc/meminfo")[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)
   end
 
   def get_postgres_param(param)
-    r "sudo -u postgres /usr/lib/postgresql/#{@version}/bin/postgres -D /dat/#{@version}/data -c config_file=/etc/postgresql/#{@version}/#{@cluster}/postgresql.conf -C #{param.shellescape}"
-  end
-
-  def shared_buffers_kib
```
fdr: this is dead code, right?

fdr:
I made one other major change, because the original code doesn't seem to implement https://github.com/ubicloud/ubicloud/pull/3405#discussion_r2132641015. It seems to have some idea of changing the number of hugepages to fit what postgres says rather than changing shared buffers to fit what hugepages are available, and those hugepages have to be chosen a-priori somehow.

So I took out the part where it set up the system hugepages, because you are going to have to have a different code path somewhere that says: "I want 25% of system memory as hugepages". This code wouldn't involve postgres at all, it'd be entirely based on `/proc/meminfo`, e.g. `MemTotal`, or, perhaps knowing from the VM record the `mem_gib` (MemTotal can be a bit fussy, it's not a round number)

The program as I have edited will read in the hugepages available, calculate overhead, and back off to make postgres fit.  Note this is incompatible with the general idea of giving users control over their shared_buffers except in an approximate sense, which is good enough.  It's also the only approach I can think of compatible with gigabyte or half-gigabyte (some ARM models) hugepages that does not involve excessive imprecision in allocation of hugepage memory.

The test has constants chosen to convey this idea: you start with a nice round number, you get some weird overhead like 37MB, and then you back off.

Whether kernel command line or sysctl, your odds of being able to freely manipulate allocate hugepages upwards at run-time is limited. You can improve it by dropping caches (e.g. `echo 3 | sudo tee /proc/sys/vm/drop_caches`, but in general, presume they are not terribly flexible once the system has been up and the memory somewhat fragmented.

In spite of these apparently annoying drawbacks, to the extent people think about shared_buffers at all (not many, not often), more people will appreciate their database performing better -- sometimes, substantially better, with number of backends >> number of processors -- than a lot of run-time flexibility to upwardly-revise `shared_buffers`

shikharbhardwaj:
@fdr I rolled all the change into one commit and added a couple on top: a check to make sure we don't set PG hugepage config for
instances where hugepages are not allocated, and a blurb in `postgres/bin/configure` to setup the system level hugepages.

rhizome/postgres/lib/hugepages_setup.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def hugepage_info
+    meminfo = File.read("/proc/meminfo")
+    hugepages_count = Integer(meminfo[/^HugePages_Total:\s*(\d+)/, 1], 10)
+    hugepage_size_kib = Integer(meminfo[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)
+    [hugepages_count, hugepage_size_kib]
+  end
+
+  def stop_postgres_cluster
+    r "sudo pg_ctlcluster stop #{@version} #{@cluster}", expect: [0, 2] # 2 is "not running"
```
fdr, line 24: I'm a little puzzled that we use pg_ctlcluster commands rather than systemd directly, but, if that's what we do...

rhizome/postgres/bin/configure
```diff
@@ -98,6 +98,36 @@ standby2replication   #{identity}             ubi_replication
...
+# Configure system hugepages
+meminfo = File.read("/proc/meminfo")
+hugepage_size_kib = Integer(meminfo[/^Hugepagesize:\s*(\d+)\s*kB/, 1], 10)
+memory_total_kib = Integer(meminfo[/^MemTotal:\s*(\d+)\s*kB/, 1], 10)
+
+target_hugepages_size_kib = memory_total_kib / 4
+target_hugepages = target_hugepages_size_kib / hugepage_size_kib
+
+r "echo 'vm.nr_hugepages = #{target_hugepages}' | sudo tee /etc/sysctl.d/10-hugepages.conf"
+r "sudo sysctl --system"
```
fdr: Before applying sysctl, you might find the odds of getting the contiguous memory improve if you do this:

```
sync
echo 3 | sudo tee /proc/sys/vm/drop_caches
echo 1 | sudo tee /proc/sys/vm/compact_memory
```

You don't need to do it now, maybe in practice you always have enough contiguous memory. It's hard to say for sure, it all depends on how long after boot this process begins. At least this handles you on reboot.

fdr (approved):

PR #3407, opened 2025-06-02
Collect VmHost node_exporter metrics and store in VictoriaMetrics

mohi-kalantari:
New logic is added to VmHost model to serve metrics config to start collecting node_exporter metrics on the VmHosts.

Also a new label is added to the VmHost prog to install the right scripts to collect the metrics from the node_exporter in the right directory for bin/monitor to collect

----

> [!IMPORTANT]
> Add functionality to collect and store VmHost node_exporter metrics in VictoriaMetrics, including configuration and setup for Grafana.
> 
>   - **Behavior**:
>     - Add `VmHost` to `metrics_target_resource_types` in `bin/monitor` to collect node_exporter metrics.
>     - New method `metrics_config` in `VmHost` to define metrics collection configuration.
>     - New `configure_metrics` label in `host_nexus.rb` to set up metrics collection service and timer.
>   - **Configuration**:
>     - Add `monitoring_service_project_id` to `config.rb` for metrics configuration.
>   - **Setup**:
>     - New `Prog::SetupGrafana` class and `setup-grafana` script to install and configure Grafana.
>   - **Testing**:
>     - Add tests for `metrics_config` in `vm_host_spec.rb`.
>     - Add tests for `SetupGrafana` in `setup_grafana_spec.rb`.
>     - Add tests for `configure_metrics` in `host_nexus_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b2f79d61aad42067cc5ac2734eb157ab03790ebc. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -26,7 +26,8 @@ class VmHost < Sequel::Model
   plugin ResourceMethods
   include SemaphoreMethods
   include HealthMonitorMethods
-  semaphore :checkup, :reboot, :hardware_reset, :destroy, :graceful_reboot
+  include MetricsTargetMethods
+  semaphore :checkup, :reboot, :hardware_reset, :destroy, :graceful_reboot, :configure_metrics
 
   def host_prefix
     net6.netmask.prefix_len
@@ -355,6 +356,12 @@ class VmHost < Sequel::Model
     }
   end
 
+  def init_metrics_export_session
+    {
+      ssh_session: sshable.start_fresh_session
+    }
+  end
+
   def disk_device_ids
     # we use this next line to migrate data from the old formatting (storing device names) to the new (storing id) so we trigger the convert
     # whenever an element inside unix_device_list is not a SSD or NVMe id.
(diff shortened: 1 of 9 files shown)


model/vm_host.rb
```diff
@@ -408,6 +415,19 @@ def render_arch(arm64:, x64:)
...
+  def metrics_config
+    {
+      endpoints: [
+        "http://localhost:9100/metrics"
+      ],
+      max_file_retention: 120,
+      interval: "15s",
+      additional_labels: {},
+      metrics_dir: "/home/rhizome/host/metrics",
+      project_id: Config.monitoring_project_id
```
shikharbhardwaj: This should ideally go to something like a `<service_name>_service_project` in general, but since nothing like that exists in this case creating a new project makes sense. We can make this the default destination for other resources as well which do not have a service project.

model/vm_host.rb
```diff
@@ -408,6 +415,19 @@ def render_arch(arm64:, x64:)
...
   end
+
+  def metrics_config
+    {
+      endpoints: [
+        "http://localhost:9100/metrics"
+      ],
+      max_file_retention: 120,
+      interval: "15s",
+      additional_labels: {},
```
shikharbhardwaj: Although `node_exporter` already adds an `instance` label to all the time series it exports, might be useful to add a standard label called `ubicloud_resource_id` to make queries across all exporters have a standard identifier.

```suggestion
      additional_labels: {"ubicloud_resource_id": ubid},
```

model/vm_host.rb
mohi-kalantari: I saw that pattern first but I thought it's not a "service" we would provide to the customers but minio also had such use cases. I added the _service_ to the name. thanks.

PR #3409, opened 2025-06-02
Replace revealable content logic with CSS

enescakir:
It's better to maintain less JS.

Adding break-all to content results same with adding overflow-wrap via JS. @byucesoy, can you confirm that it's working for you?

----

> [!IMPORTANT]
> Replaces JavaScript logic with CSS for toggling revealable content visibility and text wrapping.
> 
>   - **Behavior**:
>     - Replaces JavaScript logic for toggling visibility of `.revealable-content` with CSS classes in `app.js`.
>     - Uses CSS `break-all` for text wrapping in `revealable_content.erb`.
>   - **JavaScript**:
>     - Removes `.reveal-button` and `.hide-button` click handlers in `app.js`.
>     - Adds `.revealable-button` click handler to toggle `active` class on `.revealable-content`.
>   - **CSS**:
>     - Adds `group-[.active]:hidden` and `group-[.active]:block` to toggle visibility in `revealable_content.erb`.
>     - Uses `break-all` for text wrapping in `.revealed-content`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 373a8299458c1a8ca309cb2d41c2ad219ec1ab41. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/views/components/revealable_content.erb
+++ b/views/components/revealable_content.erb
@@ -1,13 +1,13 @@
 <%# locals: (content:) %>
-<div class="revealable-content">
-  <div class="shadow-content flex items-center gap-2">
+<div class="revealable-content group">
+  <div class="shadow-content items-center gap-2 flex group-[.active]:hidden">
     ●●●●●●
-    <img src="/icons/hero-eye.svg" width="18px" class="reveal-button inline-block cursor-pointer ms1"/>
+    <img src="/icons/hero-eye.svg" width="18px" class="revealable-button inline-block cursor-pointer ms1"/>
   </div>
-  <div class="revealed-content hidden">
+  <div class="revealed-content hidden group-[.active]:block break-all">
     <div class="flex gap-2 items-center">
       <div class="revealed-content-text"><%= content %></div>
-      <img src="/icons/hero-eye-slash.svg" width="18px" class="hide-button inline-block cursor-pointer ms1"/>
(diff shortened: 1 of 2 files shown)


views/components/revealable_content.erb
```diff
@@ -1,13 +1,13 @@
 <%# locals: (content:) %>
-<div class="revealable-content">
-  <div class="shadow-content flex items-center gap-2">
+<div class="revealable-content group">
+  <div class="shadow-content items-center gap-2 flex group-[.active]:hidden">
```
jeremyevans, line 3: `group-[.active]:hidden` is some Tailwind magic I presume?

jeremyevans (approved):
Down with JavaScript, up with CSS!

byucesoy (approved):

views/components/revealable_content.erb
enescakir: Actually, these are new CSS features that let you style based on the parent or siblings. Tailwind has a shorter syntax for this.
https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state

PR #3411, opened 2025-06-03
Use cert_store in MinioClient

mohi-kalantari:
Previously we were relying on the ssl_ca_file api of Excon which seems that it couldn't parse a chain of certs. In order to fix that we will parse the certs and add them all to a cert_store and then pass that to the client which fixes the issue.

----

> [!IMPORTANT]
> Switch `Minio::Client` to use a certificate store for handling multiple certificates, updating initialization and tests accordingly.
> 
>   - **Behavior**:
>     - `Minio::Client` now uses `ssl_cert_store` instead of `ssl_ca_file` in `client.rb`.
>     - Parses `ssl_ca_data` to add multiple certificates to `OpenSSL::X509::Store`.
>   - **Parameters**:
>     - Renames `ssl_ca_file_data` to `ssl_ca_data` in `client.rb`, `minio_server.rb`, `postgres_timeline.rb`, `download_boot_image.rb`, and `postgres_timeline_nexus.rb`.
>   - **Tests**:
>     - Updates `client_spec.rb` to test `ssl_cert_store` usage.
>     - Updates `minio_server_spec.rb`, `postgres_timeline_spec.rb`, and `postgres_timeline_nexus_spec.rb` to reflect parameter rename and new behavior.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for cb8107d5d54bd8d286458869edf980b47922e63b. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/minio/client.rb
+++ b/lib/minio/client.rb
@@ -9,13 +9,16 @@ REGION = "us-east-1"
 ADMIN_URI_PATH = "/minio/admin/v3"
 
 class Minio::Client
-  def initialize(endpoint:, access_key:, secret_key:, ssl_ca_file_data:, socket: nil)
-    ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", Digest::SHA256.hexdigest(ssl_ca_file_data) + ".crt")
-    Util.safe_write_to_file(ca_bundle_filename, ssl_ca_file_data) unless File.exist?(ca_bundle_filename)
-
+  def initialize(endpoint:, access_key:, secret_key:, ssl_ca_data:, socket: nil)
+    cert_store = OpenSSL::X509::Store.new
+    certs_pem = ssl_ca_data.scan(/-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----/m)
+    certs_pem.each do |cert_pem|
+      cert = OpenSSL::X509::Certificate.new(cert_pem)
+      cert_store.add_cert(cert)
+    end
     @creds = {access_key: access_key, secret_key: secret_key}
     @endpoint = endpoint
-    @client = Excon.new(endpoint, socket: socket, ssl_ca_file: ca_bundle_filename)
+    @client = Excon.new(endpoint, socket: socket, ssl_cert_store: cert_store)
     @signer = Minio::HeaderSigner.new
     @crypto = Minio::Crypto.new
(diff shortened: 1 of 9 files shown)


lib/minio/client.rb
```diff
@@ -9,13 +9,16 @@
 ADMIN_URI_PATH = "/minio/admin/v3"
 
 class Minio::Client
-  def initialize(endpoint:, access_key:, secret_key:, ssl_ca_file_data:, socket: nil)
-    ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", Digest::SHA256.hexdigest(ssl_ca_file_data) + ".crt")
-    Util.safe_write_to_file(ca_bundle_filename, ssl_ca_file_data) unless File.exist?(ca_bundle_filename)
-
+  def initialize(endpoint:, access_key:, secret_key:, ssl_ca_data:, socket: nil)
+    cert_store = OpenSSL::X509::Store.new
+    certs_pem = ssl_ca_data.scan(/-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----/m)
```
furkansahin, line 14: maybe instead of scanning, we can simply pass an array of certs since they are already available in the form of individual certs in the caller.

mohi-kalantari: they are not in the form of arrays. for example:

in this file: `prog/download_boot_image.rb:129`

this is how it's passed:
```ssl_ca_data: Config.ubicloud_images_blob_storage_certs``` which is a big string consisted of multiple certs.

furkansahin: Alright, I guess we read that from env variables which should support an array of string, doesn't it?

mohi-kalantari: env variables do not support arrays and we need the parsing anyways, we would need a delimiter there. it's either in the env variable reading or cert store creation which makes more sense to me to not make big changes and just parse during the cert creation.

furkansahin: Alright, thanks then.

mohi-kalantari:
Thanks man @shikharbhardwaj

PR #3412, opened 2025-06-03
Don't try to update dns when order doesn't exist at CertNexus

furkansahin:
> [!IMPORTANT]
> Add safe navigation to handle `nil` `acme_order` in `dns_challenge` and update tests accordingly.
> 
>   - **Behavior**:
>     - Use safe navigation (`&.`) in `dns_challenge` method in `cert_nexus.rb` to handle `nil` `acme_order`.
>     - In `cert_nexus_spec.rb`, add test to ensure DNS record deletion is skipped if `acme_order` is `nil`.
>   - **Tests**:
>     - Add test case in `cert_nexus_spec.rb` to verify behavior when `acme_order` is `nil` during `destroy` method execution.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 5d187f5d0f898cef7073d78403b54554ac2f3733. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/cert_nexus.rb
+++ b/prog/vnet/cert_nexus.rb
@@ -158,7 +158,7 @@ class Prog::Vnet::CertNexus < Prog::Base
   end
 
   def dns_challenge
-    acme_order.authorizations.first.dns
+    acme_order&.authorizations&.first&.dns
   end
 
   def dns_record_name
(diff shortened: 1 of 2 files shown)


prog/vnet/cert_nexus.rb
```diff
@@ -158,7 +158,7 @@ def acme_order
   end
 
   def dns_challenge
-    acme_order.authorizations.first.dns
+    acme_order.authorizations&.first&.dns
```
jeremyevans: Might be better to :

```suggestion
    acme_order&.authorizations&.first&.dns
```

Then you don't need the earlier change.

PR #3414, opened 2025-06-04
Deprecate 2004 runner images

velioglu:


enescakir (approved):
1. Can you also remove all Ubuntu 20.04 boot images? 
2. Some of the hosts have it in the `default_boot_images` in their stack. If we reimage the same host model, it might fail due to the missing image. So, it would be a good idea to remove Ubuntu 20.04 from the stack variable too

PR #3416, opened 2025-06-04
Add grafana installation strand

mohi-kalantari:
This strand will try to install grafana on the provided subject id. The strand won't create the vm or add the DNS records but takes care of all the remaining logics which must be run inside the vm like installing the nginx, certbot, getting the certificate and installing the grafana.

The API looks like this:

vm = Prog::Vm::Nexus.assemble_with_sshable(project_id,
  sshable_unix_user: "ubi", name: "grafana-0", enable_ip4: true).subject

At this step create an A record, resolving your domain to the vm ip.

Strand.create(prog: "SetupGrafana", label: "start", stack:
  [{subject_id: vm.id, domain: "grafana.domain.com",
  cert_email: "youremail@gmail.com"}])

You can then read the password of the grafana using this command: vm.sshable.cmd("sudo cat /etc/grafana/grafana.ini | grep admin_password")

----

> [!IMPORTANT]
> Add strands for Grafana and Node Exporter setup, integrate metrics configuration in VM host lifecycle, and update tests.
> 
>   - **Behavior**:
>     - Adds `SetupGrafana` strand in `prog/setup_grafana.rb` to install Grafana on a VM, handling domain and certificate email.
>     - Adds `SetupNodeExporter` strand in `prog/setup_node_exporter.rb` to install Node Exporter version 1.9.1.
>     - Integrates metrics configuration in `prog/vm/host_nexus.rb` with a new `configure_metrics` label.
>   - **Scripts**:
>     - Adds `setup-grafana` script in `rhizome/host/bin/setup-grafana` for installing Grafana, configuring Nginx, and setting up SSL.
>     - Adds `setup-node-exporter` script in `rhizome/host/bin/setup-node-exporter` for installing Node Exporter and configuring it as a service.
>   - **Models**:
>     - Updates `VmHost` in `model/vm_host.rb` to include metrics configuration and session initialization for metrics export.
>   - **Tests**:
>     - Adds tests for `SetupGrafana` in `spec/prog/setup_grafana_spec.rb`.
>     - Adds tests for `SetupNodeExporter` in `spec/prog/setup_node_exporter_spec.rb`.
>     - Updates `VmHost` tests in `spec/model/vm_host_spec.rb` to cover metrics configuration.
>     - Updates `HostNexus` tests in `spec/prog/vm/host_nexus_spec.rb` to cover new metrics configuration behavior.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 773bdd48952344cb75f472295c8d6fcf5b83669a. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/setup_grafana.rb
+++ b/prog/setup_grafana.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+class Prog::SetupGrafana < Prog::Base
+  subject_is :sshable
+
+  def self.assemble(sshable_id, grafana_domain:, certificate_owner_email:)
+    grafana_domain = grafana_domain.strip
+    certificate_owner_email = certificate_owner_email.strip
+    if grafana_domain.empty? || certificate_owner_email.empty?
+      fail "grafana_domain and certificate_owner_email should not be empty"
+    end
+    unless (sshable = Sshable[sshable_id])
+      fail "Sshable does not exist"
+    end
+    Strand.create(prog: "SetupGrafana", label: "start", stack: [{subject_id: sshable.id, domain: grafana_domain, cert_email: certificate_owner_email}])
+  end
+
+  def domain
+    frame["domain"]
+  end
+
+  def cert_email
+    frame["cert_email"]
+  end
+
+  label def start
+    bud Prog::BootstrapRhizome, {"target_folder" => "host", "subject_id" => sshable.id, "user" => sshable.unix_user}
+    hop_wait_for_rhizome
+  end
+
+  label def wait_for_rhizome
+    reap
+    hop_install_grafana if leaf?
+    donate
+  end
+
+  label def install_grafana
+    case sshable.d_check("install_grafana")
+    when "Succeeded"
+      sshable.d_clean("install_grafana")
+      pop "grafana was setup"
+    when "NotStarted"
+      sshable.d_run("install_grafana", "sudo", "host/bin/setup-grafana", domain, cert_email)
+      nap 10
+    when "InProgress"
+      nap 10
+    else
+      Clog.emit("Install grafana failed")
+      nap 65536
+    end
+  end
+end
--- a/rhizome/host/bin/setup-grafana
+++ b/rhizome/host/bin/setup-grafana
@@ -1,0 +1,65 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require "securerandom"
+
+domain = ARGV.shift.to_s.strip
+cert_email = ARGV.shift.to_s.strip
+if domain.empty? || cert_email.empty?
+  puts "Error: Both domain and cert_email must be provided."
+  exit 1
+end
+admin_password = SecureRandom.hex(16)
+grafana_ini = "/etc/grafana/grafana.ini"
+
+r "sudo apt update"
+r "sudo apt install -y apt-transport-https software-properties-common wget nginx snapd"
+r "sudo snap install certbot --classic"
+
+r "sudo mkdir -p /etc/apt/keyrings/"
+r "wget -q -O - https://apt.grafana.com/gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/grafana.gpg > /dev/null"
+r "echo \"deb [signed-by=/etc/apt/keyrings/grafana.gpg] https://apt.grafana.com stable main\" | sudo tee -a /etc/apt/sources.list.d/grafana.list"
+
+r "sudo apt update"
+r "sudo apt install -y grafana"
+
+r "sudo cp #{grafana_ini} #{grafana_ini}.bak"
+
+r "sudo sed -i \
+  -e 's/;admin_user = admin/admin_user = admin/' \
+  -e 's/;admin_password = admin/admin_password = #{admin_password}/' \
+  #{grafana_ini}"
+
+r "sudo mkdir -p /var/www/html"
+
+r %(
+sudo tee /etc/nginx/sites-available/grafana.conf > /dev/null <<'EOF'
+server {
+    listen 80;
+    server_name {{DOMAIN}};
+
+    location / {
+        proxy_pass http://localhost:3000;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+    }
+}
+EOF
+)
+r "sudo sed -i 's|{{DOMAIN}}|'#{domain.shellescape}'|' /etc/nginx/sites-available/grafana.conf"
+
+r "sudo ln -s /etc/nginx/sites-available/grafana.conf /etc/nginx/sites-enabled/"
+r "sudo systemctl start nginx"
+
+r "sudo certbot --nginx -d #{domain.shellescape} --non-interactive --agree-tos --email #{cert_email.shellescape}"
+r "sudo apt install -y ufw"
+r "sudo ufw allow 22"
+r "sudo ufw allow 80"
+r "sudo ufw allow 443"
+r "sudo ufw enable"
+
+r "sudo systemctl enable grafana-server"
+r "sudo systemctl start grafana-server"
--- a/spec/prog/setup_grafana_spec.rb
+++ b/spec/prog/setup_grafana_spec.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+require_relative "../model/spec_helper"
+
+RSpec.describe Prog::SetupGrafana do
+  subject(:sn) {
+    described_class.new(Strand.new(prog: "SetupGrafana", stack: [{"subject_id" => sshable.id, "cert_email" => "email@gmail.com", "domain" => "grafana.domain.com"}]))
+  }
+
+  let(:sshable) { Sshable.create(host: "test.localhost", raw_private_key_1: SshKey.generate.keypair) }
+
+  before do
+    allow(sn).to receive(:sshable).and_return(sshable)
+  end
+
+  describe "#domain" do
+    it "returns the domain based on the stack" do
+      expect(sn.domain).to eq("grafana.domain.com")
+    end
+  end
+
+  describe "#cert_email" do
+    it "returns the cert email based on the stack" do
+      expect(sn.cert_email).to eq("email@gmail.com")
+    end
+  end
+
+  describe "#assemble" do
+    it "fails if domain is not provided" do
+      expect { described_class.assemble(sshable.id, grafana_domain: "", certificate_owner_email: "cert@gmail.com") }.to raise_error(RuntimeError)
+    end
+
+    it "fails if cert_email is not provided" do
+      expect { described_class.assemble(sshable.id, grafana_domain: "domain.com", certificate_owner_email: "") }.to raise_error(RuntimeError)
+    end
+
+    it "fails if sshable is not provided or does not exist" do
+      expect { described_class.assemble("vm6htsmcfx5t1p60s609v49fbf", grafana_domain: "domain.com", certificate_owner_email: "cert@gmail.com") }.to raise_error(RuntimeError)
+    end
+
+    it "creates an strand with the right input" do
+      st = described_class.assemble(sshable.id, grafana_domain: "domain.com", certificate_owner_email: "cert@gmail.com")
+      st.reload
+      expect(st.label).to eq("start")
+    end
+  end
+
+  describe "#install_rhizome" do
+    it "buds a bootstrap rhizome prog and hops to wait_for_rhizome" do
+      expect(sn).to receive(:bud).with(Prog::BootstrapRhizome, {"target_folder" => "host", "subject_id" => sshable.id, "user" => sshable.unix_user})
+      expect { sn.start }.to hop("wait_for_rhizome")
+    end
+  end
+
+  describe "#wait_for_rhizome" do
+    before { expect(sn).to receive(:reap) }
+
+    it "donates if install_rhizome is not done" do
+      expect(sn).to receive(:leaf?).and_return false
+      expect(sn).to receive(:donate).and_call_original
+
+      expect { sn.wait_for_rhizome }.to nap(1)
+    end
+
+    it "hops to install_grafana when rhizome is done" do
+      expect(sn).to receive(:leaf?).and_return true
+
+      expect { sn.wait_for_rhizome }.to hop("install_grafana")
+    end
+  end
+
+  describe "#install_grafana" do
+    it "starts to install the grafana if its the first time" do
+      expect(sn).to receive(:domain).and_return("grafana.domain.com")
+      expect(sn).to receive(:cert_email).and_return("email@gmail.com")
+      expect(sshable).to receive(:d_check).and_return("NotStarted")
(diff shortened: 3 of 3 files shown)


prog/setup_grafana.rb
```diff
@@ -0,0 +1,48 @@
...
+    frame["domain"].strip
+  end
+
+  def cert_email
+    frame["cert_email"].strip
+  end
+
+  label def start
+    if domain.empty? || cert_email.empty?
+      fail "domain and cert_email should not be empty"
```
jeremyevans: If we fail, we try again, and that will fail again (a repeated failure loop).  The only way to get it to succeed is manually modify the strand's frame.

Probably a better idea to add a `self.assemble` method that strips and checks for valid domain and cert_email and raises if invalid instead of adding the strand.

If we want to keep the conditional, Clog.emit and nap 65536, similar to the failed grafana install, because there is no point in retrying until it is manually fixed.

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require "securerandom"
+
+domain = ARGV.shift
+cert_email = ARGV.shift
+if domain.nil? || domain.strip.empty? || cert_email.nil? || cert_email.strip.empty?
```
jeremyevans:
```suggestion
domain = ARGV.shift.to_s.strip
cert_email = ARGV.shift.to_s.strip
if domain.empty? ||  cert_email.empty?
```

This ensures the local variable usage below does not have excess space.

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+
+r "sudo apt update"
+r "sudo apt install -y grafana"
+
+r "sudo cp #{grafana_ini} #{grafana_ini}.bak"
+
+r "sudo sed -i 's/;admin_user = admin/admin_user = admin/' #{grafana_ini}"
+r "sudo sed -i 's/;admin_password = admin/admin_password = #{admin_password}/' #{grafana_ini}"
+r "sudo sed -i 's/;anonymous_enabled = true/anonymous_enabled = false/' #{grafana_ini}"
+r "sudo sed -i 's|^root_url *=.*|root_url = https://#{domain}/|' #{grafana_ini}"
```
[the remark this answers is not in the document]
jeremyevans: I believe this suggestion is wrong.  Shellwords.escape (or better `String#shellescape`) is designed to be used for unquoted strings, not for strings inside single quotes.   You would need:

```suggestion
r "sudo sed -i 's|^root_url *=.*|root_url = https://'#{domain.shellescape}'/|' #{grafana_ini}"
```

It does seem reasonable to shell escape this.  A potentially compromised grafana from a malicious configuration is not nearly as bad a compromised root user, which is what not escaping would allow in this case in case of malicious input.

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+
+r "sudo mkdir -p /var/www/html"
+
+r %(
+sudo tee /etc/nginx/sites-available/grafana.conf > /dev/null <<EOF
+server {
+    listen 80;
```
jeremyevans, line 39: We are only listening on 80 here.

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+
+r "sudo ln -s /etc/nginx/sites-available/grafana.conf /etc/nginx/sites-enabled/"
+r "sudo systemctl start nginx"
+
+r "sudo certbot --nginx -d #{domain} --non-interactive --agree-tos --email #{cert_email}"
+
+r "sudo apt install -y ufw"
+r "sudo ufw allow 22"
+r "sudo ufw allow 80"
+r "sudo ufw allow 443"
```
jeremyevans, line 61: But we are allowing 443 here.

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+        proxy_set_header X-Forwarded-Proto \\$scheme;
+    }
+}
+EOF
+)
+
+r "sudo ln -s /etc/nginx/sites-available/grafana.conf /etc/nginx/sites-enabled/"
+r "sudo systemctl start nginx"
+
+r "sudo certbot --nginx -d #{domain} --non-interactive --agree-tos --email #{cert_email}"
```
jeremyevans:
```suggestion
r "sudo certbot --nginx -d #{domain.shellescape} --non-interactive --agree-tos --email #{cert_email.shellescape}"
```

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+r "sudo sed -i 's/;anonymous_enabled = true/anonymous_enabled = false/' #{grafana_ini}"
+r "sudo sed -i 's|^root_url *=.*|root_url = https://#{domain}/|' #{grafana_ini}"
+
+r "sudo mkdir -p /var/www/html"
+
+r %(
+sudo tee /etc/nginx/sites-available/grafana.conf > /dev/null <<EOF
+server {
+    listen 80;
+    server_name #{domain};
```
jeremyevans: I don't think you can safely escape this here. Maybe use a sentinel value, and then use `sed` to replace it later, using `shellescape`.

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+
+r "sudo sed -i 's/;admin_user = admin/admin_user = admin/' #{grafana_ini}"
+r "sudo sed -i 's/;admin_password = admin/admin_password = #{admin_password}/' #{grafana_ini}"
+r "sudo sed -i 's/;anonymous_enabled = true/anonymous_enabled = false/' #{grafana_ini}"
+r "sudo sed -i 's|^root_url *=.*|root_url = https://#{domain}/|' #{grafana_ini}"
+
+r "sudo mkdir -p /var/www/html"
+
+r %(
+sudo tee /etc/nginx/sites-available/grafana.conf > /dev/null <<EOF
```
jeremyevans: Might be simpler with a quoted heredoc:

```suggestion
sudo tee /etc/nginx/sites-available/grafana.conf > /dev/null <<'EOF'
```

And remove the `\\` from the proxy lines below.

fdr:
some high level and/or commit message questions. At least some answers are in the diff, but I'd kinda like to know these high level calculations without having to guess. 

1) why do it? there's the obvious reason (display metrics somehow), but is there a pointed justification to do so now? It doesn't need a huge reason, but to have bothered to write this, you must have had a pointed one ("I was working on Kubernetes and saw a need for this while doing some benchmarking", or "I was working on Kubernetes and was think I might need this for support given my previous experiences, so I did this". Nothing fancier than that necessary, unless there is, in fact, a fancier reason.

    1a) "why?" is even a bit more mysterious in light of some of the limitations, like having to mess with DNS records manually. For debugging? for a few installations managed by hand? What's the calculation behind the _semi_ automation?

2) where does it run? On the same computer as victoria metrics? anywhere? In principle anywhere, but in practice on the same computer?

3) What's the reason to use certbot rather than the metadata server & load balancer? I don't think we _need_ to use our own cooking just because it's there, but seems a reflexive question.

prog/setup_grafana.rb
mohi-kalantari: changed to assemble. thanks.

rhizome/host/bin/setup-grafana
mohi-kalantari: cert bot will add the 443 entry in the nginx config and we do not need to add it manually.

mohi-kalantari:
I tried to answer your questions and make the commit more clear. do you think that's enough? @fdr

prog/setup_grafana.rb
```diff
@@ -0,0 +1,51 @@
...
+    Strand.create(prog: "SetupGrafana", label: "start", stack: [{subject_id: sshable.id, domain: grafana_domain, cert_email: certificate_owner_email}])
+  end
+
+  def domain
+    frame["domain"].strip
+  end
+
+  def cert_email
+    frame["cert_email"].strip
+  end
```
jeremyevans: Strip these values in `assemble`, so you don't need to strip them every time these methods are called.

prog/setup_grafana.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+class Prog::SetupGrafana < Prog::Base
+  subject_is :sshable
+
+  def self.assemble(sshable_id, grafana_domain, certificate_owner_email)
```
jeremyevans: Since there is no obvious order, I recommend changing the 2nd and 3rd arguments to keyword arguments.

prog/setup_grafana.rb
```diff
@@ -0,0 +1,51 @@
...
+    if grafana_domain.empty? || certificate_owner_email.empty?
```
jeremyevans, line 9: Strip the values before checking whether they are empty.

prog/setup_grafana.rb
```diff
@@ -0,0 +1,51 @@
...
+    when "NotStarted"
+      sshable.d_run("install_grafana", "sudo", "host/bin/setup-grafana", domain, cert_email)
+      nap 10
+    when "InProgress"
+      nap 10
+    when "Failed"
+      Clog.emit("Install grafana failed")
+      nap 65536
+    end
+    nap 65536
```
jeremyevans:
```suggestion
    else
      nap 65536
    end
```

As all other paths raise exceptions, the only way to reach this is in the else case, and I think it's better to make that explicit.

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+
+r "sudo apt update"
+r "sudo apt install -y grafana"
+
+r "sudo cp #{grafana_ini} #{grafana_ini}.bak"
+
+r "sudo sed -i 's/;admin_user = admin/admin_user = admin/' #{grafana_ini}"
+r "sudo sed -i 's/;admin_password = admin/admin_password = #{admin_password}/' #{grafana_ini}"
+r "sudo sed -i 's/;anonymous_enabled = true/anonymous_enabled = false/' #{grafana_ini}"
+r "sudo sed -i 's|^root_url *=.*|root_url = https://'#{domain.shellescape}'/|' #{grafana_ini}"
```
jeremyevans: `sed` supports multiple expressions separated by semicolons or newlines:

```suggestion
r "sudo sed -i '
s/;admin_user = admin/admin_user = admin/
s/;admin_password = admin/admin_password = #{admin_password}/
s/;anonymous_enabled = true/anonymous_enabled = false/
s|^root_url *=.*|root_url = https://'#{domain.shellescape}'/|' #{grafana_ini}"
```

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+        proxy_pass http://localhost:3000;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+    }
+}
+EOF
+)
+r "sudo sed -i 's|{{DOMAIN}}|#{domain.shellescape}|' /etc/nginx/sites-available/grafana.conf"
```
jeremyevans: As mentioned earlier, `shellescape` does not work correctly inside quotes:

```suggestion
r "sudo sed -i 's|{{DOMAIN}}|'#{domain.shellescape}'|' /etc/nginx/sites-available/grafana.conf"
```

mohi-kalantari:
Thanks for the new round of review. I fixed all. @jeremyevans

prog/setup_grafana.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+class Prog::SetupGrafana < Prog::Base
+  subject_is :sshable
+
+  def self.assemble(sshable_id, grafana_domain:, certificate_owner_email:)
+    if grafana_domain.strip.empty? || certificate_owner_email.strip.empty?
```
jeremyevans: It's better to avoid duplication (currently, `strip` is called twice for each variable):
```suggestion
    grafana_domain = grafana_domain.strip
    certificate_owner_email = certificate_owner_email.strip
    if grafana_domain.empty? || certificate_owner_email.empty?
```

prog/setup_grafana.rb
```diff
@@ -0,0 +1,52 @@
...
+      fail "grafana_domain and certificate_owner_email should not be empty"
+    end
+    unless (sshable = Sshable[sshable_id])
+      fail "Sshable does not exist"
+    end
+    Strand.create(prog: "SetupGrafana", label: "start", stack: [{subject_id: sshable.id, domain: grafana_domain.strip, cert_email: certificate_owner_email.strip}])
```
jeremyevans:
```suggestion
    Strand.create(prog: "SetupGrafana", label: "start", stack: [{subject_id: sshable.id, domain: grafana_domain, cert_email: certificate_owner_email}])
```

prog/setup_grafana.rb
```diff
@@ -0,0 +1,52 @@
...
+    when "NotStarted"
+      sshable.d_run("install_grafana", "sudo", "host/bin/setup-grafana", domain, cert_email)
+      nap 10
+    when "InProgress"
+      nap 10
+    when "Failed"
+      Clog.emit("Install grafana failed")
+      nap 65536
+    else
+      nap 65536
```
jeremyevans, line 49: My understanding is we shouldn't hit this case.

```suggestion
      Clog.emit("Unexpected install_grafana result")
      nap 65536
```

Maybe we should assume a different result is the same as failure, and eliminate the extra conditional?

@fdr is this something we should create a Page for?

rhizome/host/bin/setup-grafana
```diff
@@ -0,0 +1,65 @@
...
+
+r "sudo apt update"
+r "sudo apt install -y grafana"
+
+r "sudo cp #{grafana_ini} #{grafana_ini}.bak"
+
+r "sudo sed -i \
+  -e 's/;admin_user = admin/admin_user = admin/' \
+  -e 's/;admin_password = admin/admin_password = #{admin_password}/' \
+  #{grafana_ini}"
```
jeremyevans, line 32: Did you determine that the `anonymous_enabled` and `root_url` setting changes are no longer necessary?

prog/setup_grafana.rb
mohi-kalantari: I knew that this behavior is not optimal and I looked at the lifecycle of this strand as something that would be run very infrequently and someone would wait and check the result when creating it. That's why I didn't took the extra steps to create pages for it.

rhizome/host/bin/setup-grafana
mohi-kalantari: Yes the default for anonymous is false so we don't need to set it explicitly. for the root_url, nginx takes care of that and we do not need to set it.

prog/setup_grafana.rb
jeremyevans: In that case, maybe a page is not necessary.  However, I think it's still better to combine the Failed and unexpected cases.  I'd move the `Clog.emit` to the `else` branch, and remove the `when "Failed"` branch.

PR #3420, opened 2025-06-05
Simplify Postgres's Options Usage

byucesoy:
**Use lambda instead of block for option filtering**
Return statements in blocks return from not only the block but also from the
method that contains the block. This means returns cannot be used for exiting
early from option filtering blocks. Using lambdas instead allows us to use
return statements to exit early from the filtering logic without affecting
the surrounding method.

**Combine Postgres option generators**
In Postgres, we have 2 option generators: one for creation and one for updates.
They are quite similar to each other. This commit combines them into a single
generator that can handle both cases.

----

> [!IMPORTANT]
> Simplifies option handling by using lambdas for filtering and combining Postgres option generators.
> 
>   - **Behavior**:
>     - Replaces block with lambda for option filtering in `generate_firewall_options` in `helpers/firewall.rb`, `generate_kubernetes_cluster_options` in `helpers/kubernetes_cluster.rb`, and `generate_vm_options` in `helpers/vm.rb`.
>     - Combines `generate_postgres_options` and `generate_postgres_configure_options` into a single function in `helpers/postgres.rb`.
>   - **Functions**:
>     - Updates `add_option` in `lib/option_tree_generator.rb` to accept a `check` lambda instead of a block.
>   - **Routes**:
>     - Updates `routes/project/location/postgres.rb` to use the combined `generate_postgres_options` function.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for e971ca2bcc5420a25ae42e00ec5824776b82f8c4. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/helpers/firewall.rb
+++ b/helpers/firewall.rb
@@ -55,9 +55,9 @@ class Clover
         display_name: it.name
       }
     }
-    options.add_option(name: "private_subnet_id", values: subnets, parent: "location") do |location, private_subnet|
+    options.add_option(name: "private_subnet_id", values: subnets, parent: "location", check: ->(location, private_subnet) {
       private_subnet[:location_id] == location.id
-    end
+    })
     options.serialize
   end
 end
--- a/helpers/kubernetes_cluster.rb
+++ b/helpers/kubernetes_cluster.rb
@@ -43,10 +43,10 @@ class Clover
     options.add_option(name: "location", values: Option.kubernetes_locations)
     options.add_option(name: "version", values: Option.kubernetes_versions)
     options.add_option(name: "cp_nodes", values: Option::KubernetesCPOptions.map(&:cp_node_count), parent: "location")
-    options.add_option(name: "worker_size", values: Option::VmSizes.select { it.visible && it.vcpus <= 16 }.map { it.display_name }, parent: "location") do |location, size|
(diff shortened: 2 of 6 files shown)


jeremyevans (commented):
Where are we using `return` statements in these blocks?  I don't see a current case in this PR.  If this is to support the ability to use `return` in future code, instead of `return`, use `next`, and keep using a regular block instead of switching to this more cumbersome lambda approach.

byucesoy:
Yes, I was planning to use return in subsequent commits/PRs.

Honestly, I don't have any preference over blocks or lambdas. I can keep these as blocks if next provides same functionality. Just to learn, why do you think lambdas are more cumbersome? Is it because of the syntax or is there some other reason to avoid lambdas?

jeremyevans:
In general, passing a lambda as a keyword argument when you could use a block instead is not idiomatic Ruby.  Lambdas should generally be used over procs/blocks when you want to be strict about arity.

We've had many issues in the routes where `return` is used that have only worked by accident.  I think I've fixed all cases to use `next` instead.  In general, if you are inside a block and want an early exit from the block, you should use `next` unless the desired behavior is returning from the containing method.

byucesoy:
Will try a different approach.

PR #3423, opened 2025-06-05
Ubi-on-aws s3 integration for backups/restore

furkansahin:
## Add Location to PostgresTimeline - Migration
This information is required since there can be timelines without a
leader. For AWS S3 integrated timelines, we also don't have a
MinioCluster record in the database, therefor, we should still be able
to somehow refer to the address of the S3 store. For that, we are adding
the reference of location to the timeline. This way, we can always refer
to the S3 account that holds the backups of this timeline.

## Update base AWS images

----

> [!IMPORTANT]
> Add AWS S3 integration for PostgreSQL timelines, update models and configurations, and enhance tests for backup/restore functionality.
> 
>   - **Behavior**:
>     - Adds AWS S3 integration for PostgreSQL timelines in `postgres_timeline.rb`, allowing backup/restore without a MinioCluster.
>     - Updates `assemble()` in `postgres_server_nexus.rb` to select AWS base images based on PostgreSQL version.
>     - Modifies `refresh_walg_credentials()` in `postgres_server_nexus.rb` to skip blob storage CA for AWS timelines.
>   - **Models**:
>     - Adds `location_id` to `PostgresTimeline` in `20250605_add_postgres_timeline_aws.rb`.
>     - Updates `Location` model with `aws?()` method in `location.rb`.
>     - Updates `PostgresTimeline` model to include `location` association and AWS-specific logic in `postgres_timeline.rb`.
>   - **Configuration**:
>     - Updates AWS AMI versions in `config.rb`.
>     - Adds `aws-sdk-iam` to `Gemfile`.
>   - **Tests**:
>     - Adds tests for AWS S3 integration in `postgres_timeline_spec.rb` and `postgres_server_nexus_spec.rb`.
>     - Updates `client_spec.rb` to test Minio client initialization without `ssl_ca_data`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 26700110d3ac2a466e9d1b85f4066bcfa2d65745. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -56,7 +56,7 @@ class Vm < Sequel::Model
   end
 
   def ip6
-    (location.provider == "aws") ? ephemeral_net6.nth(0) : ephemeral_net6&.nth(2)
+    (location.provider == "aws") ? ephemeral_net6&.nth(0) : ephemeral_net6&.nth(2)
   end
 
   def nic
--- a/spec/model/vm_spec.rb
+++ b/spec/model/vm_spec.rb
@@ -239,6 +239,10 @@ RSpec.describe Vm do
       expect(vm).to receive(:location).and_return(instance_double(Location, provider: "aws"))
       expect(vm).to receive(:ephemeral_net6).and_return(NetAddr::IPv6Net.parse("2001:db8::/128"))
       expect(vm.ip6.to_s).to eq("2001:db8::")
+
+      expect(vm).to receive(:location).and_return(instance_double(Location, provider: "aws"))
+      expect(vm).to receive(:ephemeral_net6).and_return(nil)
+      expect(vm.ip6).to be_nil
     end
 
     it "returns the right private_ipv4 based on the netmask" do


furkansahin:
I don't know why but this E2E tests are stuck at "Test::GithubRunner.trigger_test_runs"

migrate/20250605_add_postgres_timeline_aws.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
byucesoy, line 1: If there is a timeline without a leader, it shouldn't (couldn't) archive anything anyway. For all cases where archival is required, we can count on leader being exists.

I think we can continue depending on leader.resource.location if/when we need location.

byucesoy: After reading the code more:

I think I understand the root of the problem. Previously we depend on blob storage to be an internal concept, so we were passing blob storage id and timeline can fetch the necessary information directly from our metadata database using the id.

With the AWS+S3 in picture, blob storage is not an entirely internal concept and abstraction of passing id is not working anymore. I think new and correct abstraction would be passing blob storage endpoint, because there is an endpoint for both internal and external blob storages. With the new logic, PostgresResource would need to do decoding from id to MinioCluster for internal blob storages, not the PostgresTimeline.

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -89,32 +102,50 @@ def before_run
...
-    )
+    return destroy_aws_s3 if postgres_timeline.aws?
 
     admin_client.admin_remove_user(postgres_timeline.access_key)
     admin_client.admin_policy_remove(postgres_timeline.ubid)
   end
 
+  def destroy_aws_s3
+    iam = Aws::IAM::Client.new(access_key_id: postgres_timeline.location.location_credential.access_key, secret_access_key: postgres_timeline.location.location_credential.secret_key)
+    iam.list_attached_user_policies(user_name: postgres_timeline.ubid).attached_policies.each do |it|
```
byucesoy, line 113: I assume we cannot get policy_arn without iterating the all attached policies?

migrate/20250605_add_postgres_timeline_aws.rb
byucesoy: After thinking about this more:

I think it is possible remove the location without changing what we are passing to timeline. However it would be better to change both so that:
- We would use correct abstraction 
- We wouldn't create second source of truth for location information.

Gemfile
```diff
@@ -84,3 +84,5 @@ end
 group :test, :development do
   gem "sequel-annotate"
 end
+
+gem "aws-sdk-iam", "~> 1.123"
```
enescakir: Can you move it to line 11-12? We keep them in order.

enescakir: There's no need to add a version with `~>` in the Gemfile since Dependabot upgrades it anyway.

ref: https://github.com/ubicloud/ubicloud/commit/77b004a3437726360723cfabafe16177a4febd47

migrate/20250605_add_postgres_timeline_aws.rb
furkansahin: I am not sure if I understand what you mean after reading these comments :) We are already passing the location_id to the TimelineNexus in the previous version as well, I just start to store it. I don't want to store blob storage endpoint and walk back to a blob storage object because that feels risky. 
Ideally, we could just resolve whatever endpoint would be used from the postgres resource. But, that's not possible here because the postgres timeline lifecycle surpasses postgres resource's or postgres server's. Therefore, we have to properly store which information we want to store. 
We can do one thing that combines both, though. Instead of storing `blob_storage_id`, we can just store `location_id`. Location object would be responsible from returning whatever information regarding `blob_storage`.

prog/postgres/postgres_timeline_nexus.rb
furkansahin: We can either store it when we first create the policy in an entity in the database or do this. Since we are creating this user only for this timeline, and creating a single policy to attach to it, this doesn't really increase the execution time.

byucesoy: I think what you are doing here is better than storing it.

migrate/20250605_add_postgres_timeline_aws.rb
byucesoy: > We are already passing the location_id to the TimelineNexus in the previous version as well, I just start to store it. 

We need to pass the location, only because timeline is created before server, so the link between timeline and resource is not established at the creation time. We don't need to store it though, because anytime timeline needs to do something with the location, it can get it from the resource (see below for additional thoughts on this)

> Ideally, we could just resolve whatever endpoint would be used from the postgres resource. But, that's not possible here because the postgres timeline lifecycle surpasses postgres resource's or postgres server's.

Yes, but not because of a necessity. In previous iterations, timeline would stay active and delete the entries in the blob storage. So it really needed to surpass the lifecycle of server. However, here, we delete the entries via lifecycle policy. Timeline's only responsibility is taking backups and there is no data to take backup of after server is gone. We can just delete the timeline with the resource if we want. We don't do that right now because at the beginning we were considering using a similar blob storage cleanup logic like previous iterations and didn't had chance to revisit that logic. We should at some point implement that (i.e. deletion of timeline along with the resource). Until that time, we can assume resource/server exists whenever timeline needs to do something non-trivial.

> We can do one thing that combines both, though. Instead of storing blob_storage_id, we can just store location_id. Location object would be responsible from returning whatever information regarding blob_storage.

Location object being responsible for blob storage information makes sense. Though, I would prefer going to location through resource.

furkansahin: > Until that time, we can assume resource/server exists whenever timeline needs to do something non-trivial.

We cannot though. If we make that assumption and access location through resource, we break the `destroy` logic. Because, we probably don't have a `leader` or a `resource` at the time of timeline destroy. Therefore, no way of resolving to the respective blob storage credentials for bucket/user/policy cleanup. So, we either do that refactoring now to start destroying timeline immediately at the time of resource destroy or somehow store the information in there. I'm OK refactoring the destroy `logic`, too but the question we need to answer then is, are we going to keep the backups/user/policy for a week in our records or not? What do we do if customer mistakenly drops their database and asks us to restore it 10 minutes later. We need to be able to somehow restore the timeline entity.

All things considered, I really feel like storing `location_id` in timeline is the most low cost and working solution.

byucesoy: But we don't execute the destroy logic today anyway. Timelines just linger around. We also didn't automate restoring deleted database. That is entirely manual.

Overall, I disagree that we need to store location_id twice, but at the same time I don't think it is a blocker for enabling UI for AWS. So I'm fine with going forward with storing location_id. We can revisit that once we implement timeline deletion and/or restoring deleted databases.

PR #3427, opened 2025-06-05
Refactor runner quota check to avoid project locking

enescakir:
- **Add migration for used_vcpus in the github_installation**
  We started caching the total used vCPUs in the github_installation table
  to eliminate project locking. The next commit has the details.
  

- **Refactor runner quota check to avoid project locking**
  Previously, we locked the project when checking quota for GitHub runners
  to ensure concurrent requests respected the quota limits. However, this
  caused significant performance issues for customers with many concurrent
  runner requests, as they all waited on the same lock.
  
  We removed the lock in commit 86f3d5cf due to these production
  performance issues. While this introduced a risk of temporarily
  exceeding the quota, the degraded performance was a more pressing
  concern.
  
  This commit introduces a new approach. Instead of locking the project,
  we attempt to update the total used vCPUs atomically. If the update
  succeeds, we proceed with runner creation. If it fails, we fall back to
  checking overall utilization to determine whether we can allow the
  request.
  
  We had two options: track remaining quota and decrease as runners are
  created or track used vCPUs and increment as needed.
  
  We chose the second approach for the following reasons:
      - Quotas are stored in the project_quota table, but remaining quota
        is tracked in github_installation. Keeping the two in sync would
        require a more complex update process when we want to update
        concurrency limit, including accounting for active runners’ vCPUs.
      - The remaining quota approach would require a non-negative
        constraint to prevent values dropping below zero. However, we
        occasionally allow overuse when overall utilization is low, so
        this constraint would be too restrictive and skipping it would
        make correctness harder to enforce.
  

- **Populate used_vcpus for github_installations**
  We need to populate the used_vcpus field for github_installations after
  we begin incrementing/decrementing it in the application logic.
  
  While we could run this query during the initial migration, it may
  become out of sync again before the corresponding code changes are
  deployed. Therefore, it’s safer to run this after deploying the updated
  code.
  
  Additionally, constraints are added after populating the values. If
  added beforehand, they could block runner destruction operations as
  runners may attempt to decrement used_vcpus for VMs that haven’t yet
  been counted.
  

- **Make the runner concurrency logs more like Clog style**
  

----

> [!IMPORTANT]
> Refactor runner quota check to update `used_vcpus` atomically, avoiding project locking, with supporting database migrations and updated logging.
> 
>   - **Behavior**:
>     - Refactor runner quota check in `github_runner.rb` to update `used_vcpus` atomically, avoiding project locking.
>     - If atomic update fails, checks overall utilization to decide on runner creation.
>     - Logs concurrency limit events with Clog style.
>   - **Database**:
>     - Add `used_vcpus` column to `github_installation` in `20250605_github_used_vcpus.rb`.
>     - Populate `used_vcpus` in `20250606_populate_github_used_vcpus.rb` using runner data.
>     - Add constraint `used_vcpus_not_negative` to ensure non-negative values.
>   - **Models**:
>     - Update `GithubInstallation` model to include `used_vcpus` and its constraint.
>     - Add `cores_utilization` method to `VmHost` for calculating core usage.
>   - **Tests**:
>     - Update `github_runner_spec.rb` to test new quota logic and logging changes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 91c8d762de3efe0274a764f5eda218890b863c8f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fdr (approved):
Nice changes, thoughtful commit messages.

bsatzger (approved):

furkansahin (approved):

prog/vm/github_runner.rb
```diff
@@ -143,36 +143,31 @@ def before_run
...
-    # requests. There are some remedies, but it would require some refactoring
-    # that I'm not keen to do at the moment. Although it looks weird, passing 0
-    # as requested_additional_usage keeps the existing behavior.
-    github_runner.installation.project.quota_available?("GithubRunnerVCpu", 0)
-  end
+    requested_vcpus = github_runner.label_data["vcpus"]
+    quota_vcpus = github_runner.installation.project.effective_quota_value("GithubRunnerVCpu")
+    updated_rows = github_runner.installation_dataset
+      .where { (Sequel[:used_vcpus] + requested_vcpus) <= quota_vcpus }
+      .update(used_vcpus: Sequel[:used_vcpus] + requested_vcpus)
```
jeremyevans:
```suggestion
    new_used_vcpus = Sequel[:used_vcpus] + requested_vcpus
    updated_rows = github_runner.installation_dataset
      .where(new_used_vcpus <= quota_vcpus)
      .update(used_vcpus: new_used_vcpus)
```

This addresses a couple of issues:

1. repetitive expression
2. unnecessary virtual row block

prog/vm/github_runner.rb
```diff
@@ -143,36 +143,31 @@ def before_run
...
 
-  label def wait_concurrency_limit
-    unless quota_available?
-      # check utilization, if it's high, wait for it to go down
+    # If the customer doesn't have enough quota, check the overall utilization
+    # and allow provisioning if it is low.
+    if updated_rows != 1
       utilization = VmHost.where(allocation_state: "accepting", arch: github_runner.label_data["arch"]).select_map {
         sum(:used_cores) * 100.0 / sum(:total_cores)
       }.first.to_f
```
jeremyevans:
```suggestion
      utilization = VmHost.where(allocation_state: "accepting", arch: github_runner.label_data["arch"]).get {
        sum(:used_cores) * 100.0 / sum(:total_cores)
      }.to_f
```

Not related to your PR, but this saves an array allocation, and better reflects what is desired.

prog/vm/github_runner.rb
```diff
@@ -143,36 +143,31 @@ def before_run
...
 
       unless utilization < 75
-        Clog.emit("Waiting for customer concurrency limit, utilization is high") { [github_runner, {utilization: utilization}] }
+        Clog.emit("not allowed because of high utilization") { {reached_concurrency_limit: {utilization:, label: github_runner.label, repository_name: github_runner.repository_name}} }
         nap rand(5..15)
       end
 
-      Clog.emit("Concurrency limit reached but allocation is allowed because of low utilization") { [github_runner, {utilization: utilization}] }
+      Clog.emit("allowed because of low utilization") { {reached_concurrency_limit: {utilization:, label: github_runner.label, repository_name: github_runner.repository_name}} }
+      github_runner.installation_dataset.update(used_vcpus: Sequel[:used_vcpus] + requested_vcpus)
```
jeremyevans:
```suggestion
      github_runner.installation_dataset.update(used_vcpus: new_used_vcpus)
```

Assuming you use the earlier change.

jeremyevans (approved):
Looks good.  I have a few minor optional suggestions.

prog/vm/github_runner.rb
enescakir: What do you think about https://github.com/ubicloud/ubicloud/pull/3427/commits/0a2ab7c5fe639226dd3835ced40d845188f82713

model/vm_host.rb
```diff
@@ -28,6 +28,10 @@ class VmHost < Sequel::Model
   include HealthMonitorMethods
   semaphore :checkup, :reboot, :hardware_reset, :destroy, :graceful_reboot
 
+  def self.cores_utilization(**)
+    VmHost.where(**).get { sum(:used_cores) * 100.0 / sum(:total_cores) }.to_f.round(2)
+  end
```
jeremyevans: I think it's better for this to be a dataset method:

```suggestion
  dataset_module do
    def cores_utilization
      get { sum(:used_cores) * 100.0 / sum(:total_cores) }.to_f.round(2)
    end
  end
```

prog/vm/github_runner.rb
```diff
@@ -143,36 +143,29 @@ def before_run
...
+    quota_vcpus = github_runner.installation.project.effective_quota_value("GithubRunnerVCpu")
+    new_used_vcpus = Sequel[:used_vcpus] + requested_vcpus
+    updated_rows = github_runner.installation_dataset
+      .where(new_used_vcpus <= quota_vcpus)
+      .update(used_vcpus: new_used_vcpus)
+
+    # If the customer doesn't have enough quota, check the overall utilization
+    # and allow provisioning if it is low.
+    if updated_rows != 1
+      utilization = VmHost.cores_utilization(allocation_state: "accepting", arch: github_runner.label_data["arch"])
```
jeremyevans:
```suggestion
      utilization = VmHost.where(allocation_state: "accepting", arch: github_runner.label_data["arch"]).cores_utilization
```

prog/vm/nexus.rb
```diff
@@ -249,7 +249,7 @@ def before_run
 
       incr_waiting_for_capacity unless vm.waiting_for_capacity_set?
       queued_vms = queued_vms.all
-      utilization = VmHost.where(allocation_state: "accepting", arch: vm.arch).select_map { sum(:used_cores) * 100.0 / sum(:total_cores) }.first.to_f
+      utilization = VmHost.cores_utilization(allocation_state: "accepting", arch: vm.arch)
```
jeremyevans:
```suggestion
      utilization = VmHost.where(allocation_state: "accepting", arch: vm.arch).cores_utilization
```

jeremyevans (approved):
New `cores_utilization` approach looks fine.  I suggest making it a dataset method, at your discretion.

PR #3429, opened 2025-06-06
Add kubernetes cluster list to CLI

mohi-kalantari:
Serializers and OpenAPI definitions are added. Routes are updated to
use the new serializers.

Add missing "state" column label to Kubernetes list panel

Prior to this commit, we only supported web requests but from now on,
in order to support CLI, we will support api requests as well.

New command is added to CLI which allows the customers to list
the KubernetesClusters.

A corresponding KubernetesCluster model is also created in the
Ubicloud ruby SDK.

----

> [!IMPORTANT]
> Adds CLI and API support for managing Kubernetes clusters, including listing and creating clusters, with updated serializers and OpenAPI definitions.
> 
>   - **CLI**:
>     - Adds `kc` command to `cli-commands/kc.rb` for managing Kubernetes clusters.
>     - Implements `list` and `create` subcommands in `cli-commands/kc/list.rb` and `cli-commands/kc/post/create.rb`.
>   - **API**:
>     - Updates `routes/project/kubernetes_cluster.rb` and `routes/project/location/kubernetes_cluster.rb` to support API requests for listing and creating Kubernetes clusters.
>     - Adds OpenAPI definitions for Kubernetes cluster operations in `openapi/openapi.yml`.
>   - **Serializers**:
>     - Adds `Serializers::KubernetesCluster` and `Serializers::KubernetesNodepool` for API responses.
>   - **Models**:
>     - Updates `KubernetesCluster` model in `model/kubernetes/kubernetes_cluster.rb` to include validation for `cp_node_count`.
>   - **SDK**:
>     - Adds `KubernetesCluster` model to Ubicloud Ruby SDK in `sdk/ruby/lib/ubicloud/model/kubernetes_cluster.rb`.
>   - **Tests**:
>     - Adds tests for CLI commands in `spec/routes/api/cli/kc/create_spec.rb`.
>     - Adds tests for API routes in `spec/routes/api/project/location/kubernetes_cluster_spec.rb`.
>     - Adds tests for serializers in `spec/serializers/kubernetes_cluster_spec.rb` and `spec/serializers/kubernetes_nodepool_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for aafdf38609be29e65f10dcd14d8366ad56df0bef. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fdr:
Nice, I thought @jeremyevans might write this, but all the better that you found time to do it, since you are more proximate to the product.

cli-commands/kc/list.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+UbiCli.list("kc", %w[location name id display_state version cp_node_count])
```
aamederen, line 3: shall we return the nodepool node count as well or should we return the nodepools for a future `ubi kn` command?

lib/ubi_cli.rb
```diff
@@ -12,7 +12,8 @@ class UbiCli
     "lb" => "load_balancer",
     "pg" => "postgres",
     "ps" => "private_subnet",
-    "vm" => "vm"
+    "vm" => "vm",
+    "kc" => "kubernetes_cluster"
```
aamederen: I think we keep these lists in alphabetical order

sdk/ruby/lib/ubicloud/context.rb
```diff
@@ -13,6 +13,7 @@ module Ubicloud
   # +postgres+ :: Ubicloud::Postgres
   # +private_subnet+ :: Ubicloud::PrivateSubnet
   # +vm+ :: Ubicloud::Vm
+  # +kubernetes_cluster+ :: Ubicloud::KubernetesCluster
```
aamederen: alphabetical ordering

aamederen:
apart from a few nitpicks, LGTM. @jeremyevans also should review IMO

helpers/kubernetes_cluster.rb
```diff
@@ -31,8 +31,8 @@ def kubernetes_cluster_post(name)
   end
 
   def kubernetes_cluster_list
-    @kcs = dataset_authorize(@project.kubernetes_clusters_dataset, "KubernetesCluster:view").all
-
+    dataset = dataset_authorize(@project.kubernetes_clusters_dataset, "KubernetesCluster:view")
+    @kcs = Serializers::KubernetesCluster.serialize(dataset.all)
```
jeremyevans: We are not changing any views to use serializers or adding any new views using serializers.  We are slowing working on migrating existing views that use serializers to not use serializers.  So all parts of this commit that switch the views to use serializers should be backed out.

routes/project/location/kubernetes_cluster.rb
```diff
@@ -2,42 +2,53 @@
...
-        filter = if kc_name
-          {Sequel[:kubernetes_cluster][:name] => kc_name}
-        else
-          {Sequel[:kubernetes_cluster][:id] => UBID.to_uuid(kc_ubid)}
+    r.get api? do
+      kubernetes_cluster_list
+    end
+
+    r.on KUBERNETES_CLUSTER_NAME_OR_UBID do |kc_name, kc_ubid|
+      filter = if kc_name
```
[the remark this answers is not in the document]
jeremyevans: Ignore the ellipsis comment, this is the standard way other routes are structured.

spec/routes/api/project/location/kubernetes_cluster_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      login_api
+    end
+
+    describe "list" do
+      it "success list" do
+        get "/project/#{project.ubid}/location/#{kc.display_location}/kubernetes-cluster"
+
+        expect(last_response.status).to eq(200)
+        parsed_body = JSON.parse(last_response.body)
+        expect(parsed_body["items"].length).to eq(1)
```
jeremyevans, line 55: There should be at least one spec that tests the returned JSON contains the expected fields and values, for every api route where we are returning a successful response.

cli-commands/kc/list.rb
jeremyevans: If the nodepool node count is useful, it's probably best to return it if the api exposes it.  We can still add a `ubi kn` command later to get more detailed information about nodepools later.

mohi-kalantari: I feel like the best way to see the nodepool info is through its own API `ubi kn`. The current cluster model in DB does not hold the info and nodepool count is inconclusive if it doesn't contain the nodepool node count and sizes.

helpers/kubernetes_cluster.rb
mohi-kalantari: This is fixed. Reverted all changes of using Serializers in views.

spec/routes/api/project/location/kubernetes_cluster_spec.rb
mohi-kalantari: added the missing test.

routes/project/kubernetes_cluster.rb
```diff
@@ -2,13 +2,13 @@
...
+    end
+
     r.web do
       r.is do
-        r.get do
-          kubernetes_cluster_list
-        end
-
-        r.post do
+        r.post true do
```
jeremyevans: `r.is` is for when you have different HTTP verbs handling the same route, and want to do the check that you've routed the entire path only once.  In this case, you only have 1 route, so it isn't needed.  You can remove the `r.is`, and just have `r.post true`.  Note that if you are using `r.get` or `r.post` inside `r.is`, you do not need to provide the `true` argument.

routes/project/location/kubernetes_cluster.rb
```diff
@@ -2,44 +2,54 @@
...
+      filter[:location_id] = @location.id
+      kc = @project.kubernetes_clusters_dataset.first(filter)
 
-        r.is do
-          r.get do
-            authorize("KubernetesCluster:view", kc.id)
+      next 404 unless kc
 
+      r.is do
+        r.get true do
```
jeremyevans: You shouldn't be adding the `true` argument here, you are already inside an `r.is` block, so the path has already been full routed.  Same with the following `r.delete` call.

spec/routes/api/project/location/kubernetes_cluster_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        expect(parsed_body["count"]).to eq(1)
+        expect(parsed_body["items"][0]["name"]).to eq("cluster")
+        expect(parsed_body["items"][0]["version"]).to eq("v1.32")
+      end
+    end
+
+    describe "create" do
+      it "success" do
+        post "/project/#{project.ubid}/location/#{TEST_LOCATION}/kubernetes-cluster/test-cluster", {
```
jeremyevans, line 64: Should actually test that this creates a cluster in the database.

cli-commands/kc.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+UbiCli.base("kc") do
+  banner "ubi kc command [...]"
+  post_banner "ubi kc (location/cluster-name | kc-id) post-command [...]"
```
jeremyevans:
```suggestion
  post_banner "ubi kc (location/kc-name | kc-id) post-command [...]"
```

For consistency with the other commands.

cli-commands/kc/post/create.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+UbiCli.on("kc").run_on("create") do
+  desc "Create a Kubernetes cluster"
+
+  options("ubi kc location/cluster-name create [options]", key: :kc_create) do
```
jeremyevans:
```suggestion
  options("ubi kc location/kc-name create [options]", key: :kc_create) do
```

cli-commands/kc/post/create.rb
```diff
@@ -0,0 +1,22 @@
...
+    on("-v", "--version=version", "Kubernetes version")
+    on("-c", "--cp-node-count=count", "Control plane node count")
+    on("-w", "--worker-node-count=count", "Worker node count")
+    on("-z", "--worker-size=size", "Worker size")
+  end
+  help_option_values("Version:", Option.kubernetes_versions)
+  help_option_values("Control Plane Node Count:", Option::KubernetesCPOptions.map { |x| x.cp_node_count.to_s })
```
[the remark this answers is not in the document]
jeremyevans:
```suggestion
  help_option_values("Control Plane Node Count:", Option::KubernetesCPOptions.map(&:cp_node_count))
```

I don't think the explicit `to_s` is necessary, as rodish calls it.

cli-commands/kc/post/create.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  run do |opts|
+    params = underscore_keys(opts[:kc_create])
+    params[:cp_nodes] = params.delete(:cp_node_count)
```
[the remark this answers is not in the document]
jeremyevans: I don't think this is necessary.  If you are passing a non-integer, the `create` below will fail with a validation error. It is probably worth adding a golden file test for passing a non-integer, and confirm the error output is reasonable (and make appropriate changes if not).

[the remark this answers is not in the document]
jeremyevans: Ellipsis's advice is not necessary, as Sequel will convert values to integer since the column is integer.  

However, it's a bad idea to have required command line options. If the user does not provide a value for the option, choose a reasonable default.  I wouldn't do it here, I would do it in the API handling.  Use whatever defaults are shown in the web interface (1 CP node, standard-2 worker size, 1 worker node, etc.).

It looks like KubernetesCluster#validate does not correctly handle the case where `cp_node_count` is nil or a non-integer value, so that should be fixed as well.

To make sure bad input is handled correctly, add separate golden file tests using a invalid input for each separate option, and ensure the error output is reasonable.

Additionally, I would not set an explicit nil value if no option was provided (same for the following line for worker nodes):

```suggestion
    params[:cp_nodes] = params.delete(:cp_node_count) if params[:cp_node_count]
```

jeremyevans (changes requested):

cli-commands/kc/post/create.rb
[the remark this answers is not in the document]
jeremyevans: I think this was mistakenly marked as resolved, because the change I requested was not made.

jeremyevans (approved):

jeremyevans:
@mohi-kalantari can you please make the change suggested in https://github.com/ubicloud/ubicloud/pull/3429#discussion_r2159339950 before merging?

jeremyevans:
#3513 has the fixes necessary for better error handling during CLI option parsing. Please update this PR to use those features for better error handling in the `kc` CLI (either switch the merge base to it, or wait until it is merged).

mohi-kalantari:
> @mohi-kalantari can you please make the change suggested in [#3429 (comment)](https://github.com/ubicloud/ubicloud/pull/3429#discussion_r2159339950) before merging?

I had fixed it but forgot to force push it. thanks.

PR #3431, opened 2025-06-10
PG cache earliest backup time

shikharbhardwaj:
> [!IMPORTANT]
> Add caching for earliest backup time in `PostgresTimeline` to improve performance.
> 
>   - **Database Schema**:
>     - Add `cached_earliest_backup_at` column to `postgres_timeline` table in `20250610_add_cached_earliest_backup_time.rb`.
>   - **Model Changes**:
>     - Add `BACKUP_BUCKET_EXPIRATION_DAYS` constant in `PostgresTimeline`.
>     - Update `earliest_restore_time` in `PostgresTimeline` to cache earliest backup time and use it if valid.
>   - **Code Adjustments**:
>     - Use `BACKUP_BUCKET_EXPIRATION_DAYS` in `set_lifecycle_policy` in `postgres_timeline_nexus.rb`.
>   - **Tests**:
>     - Add test for `earliest_restore_time` caching behavior in `postgres_timeline_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 82e0c7c715bd4c769fed9edb28831af56cd86de2. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


model/postgres/postgres_timeline.rb
```diff
@@ -64,12 +66,23 @@ def latest_backup_label_before_target(target:)
...
+    # Check if we have cached earliest backup time, if not, calculate it.
+    # The cached time is valid if its within BACKUP_BUCKET_EXPIRATION_DAYS.
+    time_limit = Time.now - BACKUP_BUCKET_EXPIRATION_DAYS * 24 * 60 * 60
+
+    if cached_earliest_backup_at.nil? || cached_earliest_backup_at <= time_limit
+      # Pick the earliest backup that is later than
+      # the time limit.
+      earliest_backup = backups
+        .select { |b| b.last_modified > time_limit }
+        .min_by(&:last_modified)&.last_modified
```
byucesoy: If you do .map(&:last_modified).min (like the original) you can avoid safe navigation operator (also usually (not in this case) the extra test case to cover it). The cost is that you do one extra loop but this array will have max size of 8(9?) anyway.

model/postgres/postgres_timeline.rb
```diff
@@ -64,12 +66,23 @@ def latest_backup_label_before_target(target:)
...
-  # can cache this somehow.
   def earliest_restore_time
-    if (earliest_backup = backups.map(&:last_modified).min)
-      earliest_backup + 5 * 60
+    # Check if we have cached earliest backup time, if not, calculate it.
+    # The cached time is valid if its within BACKUP_BUCKET_EXPIRATION_DAYS.
+    time_limit = Time.now - BACKUP_BUCKET_EXPIRATION_DAYS * 24 * 60 * 60
+
+    if cached_earliest_backup_at.nil? || cached_earliest_backup_at <= time_limit
+      # Pick the earliest backup that is later than
```
byucesoy: This comment seems unnecessary. It explains `what` (which can be understand by reading the code). If it was explaining `why`, then it would be more valuable, such as the previous comment: `The cached time is valid if its within BACKUP_BUCKET_EXPIRATION_DAYS.`

byucesoy (approved):

PR #3437, opened 2025-06-11
Better NetAddr/Sequel integration

jeremyevans:
Previously, the only integration was conversion procs for inet and cidr. This adds:

* conversion procs for inet[] and cidr[] (previously these used IPAddr instead of NetAddr)

* typecasting support

* schema type validation

* NetAddr object literalization and bound variable support

Update the update_firewall_rules prog to handle NetAddr instead of IPAddr instances.

This gives inet/cidr/inet[]/cidr[] columns a schema type when they did not have a schema type previously, which breaks the usage of literal string values for array columns in model objects. Literal strings are expressions and not values, and therefore shouldn't be used as model object values, they only work due to implementation details if a schema type is not present.  Fix cases where literal strings were used as model object values, making all affected code simpler.

----

> [!IMPORTANT]
> Enhance NetAddr and Sequel integration by adding conversion procs, typecasting, and schema validation for inet/cidr types, and update firewall rules to use NetAddr.
> 
>   - **Integration Enhancements**:
>     - Add conversion procs for `inet[]` and `cidr[]` in `db.rb`.
>     - Implement typecasting and schema type validation for `inet` and `cidr` in `lib/net_addr_sequel.rb`.
>     - Register array types for `inet` and `cidr` in `db.rb`.
>   - **Program Updates**:
>     - Update `resolve_globally_blocked_dnsnames.rb` to handle `NetAddr` objects.
>     - Modify `update_firewall_rules.rb` to use `NetAddr` instead of `IPAddr`.
>   - **Testing**:
>     - Add tests for `NetAddr` conversion and typecasting in `db_spec.rb`.
>     - Update `update_firewall_rules_spec.rb` to reflect changes in firewall rule handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c9d17194b7b56321384e17991267695bda71854d. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fdr (approved):
Looking at this, I think maybe I missed a trick somewhere with using a `cidr` where I should have used `inet` to represent something like `2601:645:c57e:6a80:c999:8475:1ac5:3e33/64`, that is, a v6 network with the low bits set.

fdr:
I feel like this is at least one occurrence of my misunderstanding:

```
CREATE TABLE public.vm_host (
    id uuid NOT NULL,
    allocation_state public.allocation_state DEFAULT 'unprepared'::public.allocation_state NOT NULL,
    ip6 inet,
    net6 cidr,
```

ip6 + net6 are redundant if ip6's inet could contain the mask. I think.

prog/vnet/update_firewall_rules.rb
```diff
@@ -258,8 +258,11 @@ def generate_globally_blocked_lists
     GloballyBlockedDnsname.each do |globally_blocked_dnsname|
       ips = globally_blocked_dnsname.ip_list || []
       ips.each do |ip|
-        globally_blocked_ipv4s << "#{ip}/32" if ip.ipv4?
-        globally_blocked_ipv6s << "#{ip}/128" if ip.ipv6?
+        if ip.version == 4
+          globally_blocked_ipv4s << "#{ip}/32"
+        else
+          globally_blocked_ipv6s << "#{ip}/128"
+        end
```
enescakir, line 265: another option
```suggestion
        globally_blocked_ipv4s << ip.version == 4 ? "#{ip}/32" : "#{ip}/128"
```

jeremyevans: I made the same mistake while I was working on this, but it actually appends to two separate arrays: `globally_blocked_ipv4s` and `globally_blocked_ipv6s`

PR #3438, opened 2025-06-11
Optimize VmHost#ip4_random_vm_network

jeremyevans:
This allows VmHost#ip4_random_vm_network to use a single query instead of potentially recursive method calls with N+1 queries.

The first step is populating a table with all valid IPv4 addresses for VMs running on the host. This adds a `ipv4_address` table, and has newly created `Address` instances populate the table if the address is IPv4 and not for a single IP.

With the table populated, VmHost#ip4_random_vm_network only needs to take the universe of available addresses, remove the assigned addresses, and pick a random remaining address.

This currently emits and falls back to the previous implementation if the `ipv4_address` has not been populated for the host.  We can remove the fallback after some time if we do not see any production issues, this is just me being cautious.

This depends on #3437, which adds better NetAddr/Sequel integration.


model/address.rb
```diff
@@ -5,8 +5,44 @@
...
+  end
+
+  def after_create
+    super
+    populate_ipv4_addresses
+  end
+
+  def populate_ipv4_addresses
+    # Do nothing for ipv6 addresses, since VM addresses are chosen randomly from the /64.
+    # Ignore /32 IPv4 addresses, since those would be used by the host itself and not for
```
enescakir: Our MinIO hosts have an additional /32 subnet along with the host's /32. We purchased a /32 instead of a larger subnet because we only provision a single VM on it. (@furkansahin for verification)

enescakir: You might use `vm_host.sshable.host` or `Address[vm_host.id]` to ignore

jeremyevans: Maybe we should check whether the host's sshable ip matches instead of just checking the mask length?

spec/model/vm_host_spec.rb
```diff
@@ -184,6 +186,7 @@ def initialize(str, exitstatus = 0)
   end
 
   it "returns nil if there is no available subnet" do
+    vh.id = described_class.generate_uuid
```
enescakir: Instead of calling `generate_uuid` individually, you might use `new_with_id` for the `:vh` subject.

```ruby
subject(:vh) {
    described_class.new_with_id(

```

jeremyevans: This makes sense to me.  However, it causes errors in 6 other vm host specs.  So I need to work around that.

model/address.rb
jeremyevans: Changed this to use `id != routed_to_host_id` (skipping Address where id matches related host).

spec/model/vm_host_spec.rb
jeremyevans: Made the appropriate spec changes in an additional commit.

jeremyevans:
I noticed this comment in `ip4_random_vm_network`:

```
# we pick a random /31 subnet from the available subnet
```

Which indicates that we don't actually assign the entire IP address range to VMs, we only assign half of the range to VMs, with the other half used as the gateway for VMs.

If this comment is accurate, this approach needs to be modified to only add half of the addresses to `ipv4_address`.  However, I think this comment is inaccurate.  It dates back to da357bda9c7d97aff993e92c3f52961cda800b89, where that appears to be how it worked (`.nth_subnet(31, rand)`).  I think it changed in 6524645b650262d7efd1bdc5a3b411e7361ed356, but the comment was not removed.

Is my understanding correct?

model/address.rb
enescakir: The comment should be updated as well. "Ignore /32 IPv4 addresses" is not correct anymore

model/address.rb
```diff
@@ -5,8 +5,44 @@
...
+      addresses = Array.new(cidr.len) { [cidr.nth(it), cidr.to_s] }
+
+      if vm_host.provider_name == "leaseweb"
+        # Do not use first or last addresses for leaseweb
+        addresses.shift
+        addresses.pop
+      end
+
+      DB[:ipv4_address].import([:ip, :cidr], addresses)
+    end
```
enescakir: It's totally personal (feel free to ignore) but I prefer an early return rather than wrapping the whole function in a block. less indentation

```suggestion
    return unless cidr.is_a?(NetAddr::IPv4Net) && id != routed_to_host_id
    
    addresses = Array.new(cidr.len) { [cidr.nth(it), cidr.to_s] }

    if vm_host.provider_name == "leaseweb"
      # Do not use first or last addresses for leaseweb
      addresses.shift
      addresses.pop
    end

    DB[:ipv4_address].import([:ip, :cidr], addresses)
```

furkansahin:
@jeremyevans your understanding is correct but that comment is outdated. Let me update the comment. We make use of the whole range now.

enescakir:
The PR looks good to me overall, but our network expert, @furkansahin, should be the final approver.

jeremyevans:
Fixed/removed the inaccurate comments and rebased on main to fix the conflict. @furkansahin, waiting on your review and approval.

enescakir:
it looks like E2E tests were failed at `23:03:34 e2e.1       | 54.85m | 01976646-77f0-8b3a-b4fc-8cf54cb332d0 | Test::VmGroup.wait_vms | waiting for exit | Vm.start, Vm.start, Vm.start, Vm.start`

jeremyevans:
Looking at the logs, problem is:

```
12:38:18 respirate.1 | {"exception":{
"message":"undefined method 'id' for an instance of String",
"class":"NoMethodError",
"backtrace":[
"/home/runner/work/ubicloud/ubicloud/scheduling/allocator.rb:273:in 'Scheduling::Allocator::Allocation.update_vm'
```

So the issue is the method should return an `Address` and not an address id.  That's an easy fix.  Surprised there wasn't an existing test for this. I'll add one.

jeremyevans:
Actually, the existing tests do test for `Address` instances being returned.  They just don't hit this code path as they all use heavy mocking and do not test with saved `VmHost`/`Address`/etc. instances.  So it's yet another case of the use of mocking causing problems.

furkansahin (approved):

model/vm_host.rb
```diff
@@ -127,8 +129,31 @@ def ip6_random_vm_network
   end
 
   def ip4_random_vm_network
+    ipv4_ds = DB[:ipv4_address].join(:address, [:cidr]).where(cidr: assigned_subnets_dataset.select(:cidr))
+
+    res = ipv4_ds
+      .exclude(ip: assigned_vm_addresses_dataset.select(:ip))
+      .order { random.function }
+      .first
```
fdr: Maybe use `~Sequel.exists(
  assigned_vm_addresses_dataset.where(ip: Sequel[:ipv4_address][:ip])` to get a NOT EXIST (SELECT 1 ...) construction (right?) because the null semantics of NOT IN are almost never what anyone wants. They also don't usually like the de-optimization either, even if at times they can be spared by a NOT NULL constraint. I basically never use `NOT IN`, doesn't seem worth it.

jeremyevans: Pushed a commit to use NOT EXISTS

PR #3444, opened 2025-06-13
Create worker strands from a list

enescakir:
We have some worker strands that process tasks on a schedule, similar to cron jobs. We create them when starting the respirate if they haven't been created yet. While we create them one by one, we can use a loop to avoid duplicating code.


bin/respirate
```diff
@@ -6,29 +6,20 @@ require_relative "../loader"
...
+  ["stcheckzvsagezalertszzzzza", "CheckUsageAlerts"],
+  ["stexp1repr0ject1nv1tat10na", "ExpireProjectInvitations"],
+  ["st10g0vmh0st0vt111zat10nzz", "LogVmHostUtilizations"],
+  Config.heartbeat_url ? ["stheartbeatheartbheartheaz", "Heartbeat"] : nil,
+  Config.github_app_id ? ["stredelivergith0bfail0reaz", "RedeliverGithubFailures", [{last_check_at: Time.now}]] : nil
+]
+workers.compact!
+
+workers.each do |ubid, prog, stack|
+  Strand.dataset.insert_conflict.insert(id: UBID.parse(ubid).to_uuid, prog:, label: "wait", stack: (stack || [{}]).to_json)
```
jeremyevans: I would use `import` instead of `insert`, and then you can handle everything in one `INSERT` statement.

jeremyevans (approved):

bin/respirate
enescakir: How do we deal with conflicts when using `import`?

jeremyevans: You are already using `insert_conflict`, and that handles it.

enescakir: I changed it

fdr (approved):

PR #3446, opened 2025-06-13
Postgres UI Revamp

byucesoy:
As we add more and more features for PostgreSQL, it become difficult to navigate. So we decided to split the features into multiple pages. This PR does those splitting along with some clean up work. All tests are fixed as a separate commit, which is something I don't want to do usually. I prefer every commit to pass tests, however it was difficult to do that for such big refactoring.

Here are some notable pages:
|Overview    | Connections | Backups
| -------- | ------- |------- |
| Overview | Connection | Backups |

----

> [!IMPORTANT]
> Revamps PostgreSQL UI by splitting features into multiple pages, updating routes, and enhancing components for improved navigation and user experience.
> 
>   - **UI Revamp**:
>     - Splits PostgreSQL features into multiple pages: `overview`, `connection`, `charts`, `networking`, `resize`, `high-availability`, `read-replica`, `backup-restore`, `settings`.
>     - Updates `show.erb` to render pages dynamically based on the selected submenu.
>     - Adds new components like `icon_with_text.erb` for better UI representation.
>   - **Routing and Redirection**:
>     - Updates routes in `postgres.rb` to redirect to specific pages like `/overview`, `/settings`, etc.
>     - Adjusts redirection logic for actions like `delete`, `promote`, `reset-superuser-password`.
>   - **JavaScript and CSS**:
>     - Adds animations in `tailwind.config.js` for UI elements.
>     - Implements JavaScript for dynamic content updates in `app.js`.
>   - **Testing**:
>     - Updates tests in `postgres_spec.rb` to align with new page structure and functionalities.
>     - Ensures tests cover new navigation and feature access scenarios.
>   - **Miscellaneous**:
>     - Changes `channel_binding=require` to `sslmode=require` in connection strings across multiple files.
>     - Refactors components like `copyable_content.erb` and `revealable_content.erb` for enhanced functionality.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8d43478127462bba481c936c5bbdf1ed804e0395. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


routes/project/location/postgres.rb
```diff
@@ -29,9 +29,10 @@ class Clover
         if api?
           Serializers::Postgres.serialize(pg, {detailed: true})
```
enescakir, line 30: We completely remove serializers. As far as I know, we also inlined API responses instead of using serializers. You might want to remove them in future PRs as well.

routes/project/location/postgres.rb
```diff
@@ -29,9 +29,10 @@ class Clover
...
         else
-          @pg = Serializers::Postgres.serialize(pg, {detailed: true, include_path: true})
+          @pg = pg
```
enescakir: Since you made this change in a single commit, I would prefer to see `@pg = @project.postgres_resources_dataset.first(filter)` and the change applied throughout the whole file instead of `@pg = pg`

routes/project/location/postgres.rb
```diff
@@ -22,18 +22,26 @@ class Clover
       pg = @project.postgres_resources_dataset.first(filter)
       check_found_object(pg)
 
-      r.get true do
+      r.get api?, true do
```
jeremyevans: I expect this will result in broken links for users who have an existing page, and click one a link to the postgres show page, as that link will become invalid by this change (also true for users who have bookmarked the page directly).  I recommend you accept web requests for this route, and redirect to the overview page.

views/postgres/overview.erb
```diff
@@ -0,0 +1,71 @@
...
+  %>
+  <%== part("components/icon_with_text", icon:, text:, subtext:) %>
+  <%
+    end
+  %>
+</div>
+
+<% @enable_echarts = true %>
+
+<div id="metrics-container" class="p-6" data-metrics-url=<%= "#{@project_data[:path]}#{@pg.path}" %>>
```
jeremyevans:
```suggestion
<div id="metrics-container" class="p-6" data-metrics-url="<%= "#{@project_data[:path]}#{@pg.path}" %>">
```

Maybe not strictly necessary, as neither path should have spaces or other characters that would cause problems, but it's best to always quote all attribute values.

Same for the other places where `data-metrics-url` is used with an unquoted value.

views/postgres/submenu.erb
```diff
@@ -0,0 +1,30 @@
+<aside class="lg:col-span-4 xl:col-span-3 2xl:col-span-2">
+  <nav class="space-y-1">
+    <% [
+        ["Overview", "overview", "hero-squares-plus"],
+      ].each do |label, page, icon| %>
+      <% if @page == page %>
```
jeremyevans: Up to you, but I think this would be better if you had a separate conditionals around the attributes that may change depending on current page, as opposed to duplicating the entire `a` element.

views/postgres/charts.erb
```diff
@@ -0,0 +1,113 @@
...
+          <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Password</th>
+          <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"></th>
+        </tr>
+      </thead>
+      <tbody class="divide-y divide-gray-200 bg-white">
+        <% @pg.metric_destinations.each do |md| %>
+          <tr>
+            <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6"><%= md.url %></td>
+            <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6"><%= md.username %></td>
+            <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6">●●●●●●</td>
```
jeremyevans: Is there a reason to have a password column if we aren't showing the password?  If we have the column, shouldn't it be copyable and/or revealable?

views/postgres/backup_restore.erb
```diff
@@ -0,0 +1,71 @@
...
+      <table class="min-w-full divide-y divide-gray-300">
+        <thead class="bg-gray-50">
+          <tr>
+            <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">Backup Time</th>
+            <th scope="col" class="py-3.5 text-left text-sm font-semibold text-gray-900">Type</th>
+            <th scope="col" class="py-3.5 text-left text-sm font-semibold text-gray-900">Fork</th>
+          </tr>
+        </thead>
+        <tbody class="divide-y divide-gray-200 bg-white">
+          <% backups.map(&:last_modified).each do |time| %>
```
jeremyevans: Not sure how many backups there are per server on average, but if the number is large, it may be worth skipping the map and instead calling last modified in the two places that need it.

jeremyevans (approved):
This looks really good.  I tested locally in the browser and it is easy to use.

views/postgres/overview.erb
```diff
@@ -0,0 +1,71 @@
+<%
+vm = @pg.representative_server.vm
+%>
+
+<div class="grid grid-cols-1 md:grid-cols-3 gap-6 p-6 database-overview">
+  <%
+    vmr = VictoriaMetricsResource.first(project_id: @pg.representative_server.metrics_config[:project_id])
+    if vmr
+      tsdb_client = vmr.servers.first.client
+      query = "node_filesystem_size_bytes{mountpoint='/dat', ubicloud_resource_id='#{@pg.ubid}', ubicloud_resource_role='primary'} - node_filesystem_avail_bytes{mountpoint='/dat', ubicloud_resource_id='#{@pg.ubid}', ubicloud_resource_role='primary'}"
```
shikharbhardwaj: We can potentially reuse the query from here: https://github.com/ubicloud/ubicloud/blob/c4f0be4ace07d2b9ca5d1e32dca8886bf10839f6/lib/metrics.rb#L64

views/postgres/overview.erb
```diff
@@ -0,0 +1,71 @@
...
+      disk_usage = begin
+        tsdb_client.query_range(query:, start_ts: Time.now.to_i - 60 * 60, end_ts: Time.now.to_i)[0]
+      rescue Excon::Error::Socket => e
+        Clog.emit("Failed to fetch disk usage for #{@pg.ubid} from VictoriaMetrics") { Util.exception_to_hash(e) }
+      end
+
+      disk_usage_subtext = if disk_usage
+        disk_usage_gib = disk_usage["values"].last[1].to_f / 1024 / 1024 / 1024
+        disk_usage_percentage = (100 * disk_usage_gib / @pg.representative_server.storage_size_gib)
+        "#{disk_usage_gib.round(2)} GB is used (#{disk_usage_percentage.round(2)}%)"
```
shikharbhardwaj: This is neat. An idea: Should we also indicate a warning for high disk usage (maybe > 90%) with some formatting like text color?

assets/js/app.js
```diff
@@ -246,6 +246,28 @@ function setupDatePicker() {
...
+  date_picker = flatpickr("#restore_target", {enableTime: true, dateFormat: "Y-m-d H:i"})
+  date_picker.setDate(target_datetime, true);
+
+  $("#restore_target").addClass("animate-flash transition-colors duration-1000");
+  setTimeout(() => {
+    $("#restore_target").removeClass('animate-flash');
+  }, 2000);
+})
+
+$(document).on('change', '.connection-info-format-selector select, .connection-info-format-selector input', function() {
```
enescakir: Why do we add listener to whole document?

```suggestion
$(".connection-info-format-selector select, .connection-info-format-selector input").on('change', function() {
```

views/components/copyable_content.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: (content:, message: "Copied", revealable: false) %>
 
-<div class="copyable-content inline-flex items-center gap-1" data-content="<%= content %>" data-message="<%= message %>">
+<div class="copyable-content justify-between flex items-center gap-1" data-content="<%= content %>" data-message="<%= message %>">
```
enescakir: We use this component in a few places, and these changes make them a little bit uglier.
CleanShot 2025-06-17 at 12 22 17@2x

views/postgres/submenu.erb
```diff
@@ -0,0 +1,39 @@
+<aside class="lg:col-span-4 xl:col-span-3 2xl:col-span-2">
+  <nav class="space-y-1">
+    <% [
+        ["Overview", "overview", "hero-squares-plus"],
+        ["Connection", "connection", "plug-disconnected"],
+        ["Metrics", "charts", "hero-chart-pie"],
+        ["Networking", "networking", "hero-globe-alt"],
+        ["Scale Up/Down", "scale-up-down", "hero-arrow-top-right-on-square"],
```
enescakir: may be? `Resize`, `Resize Database` or `Resize DB`
```suggestion
        ["Resize Database", "scale-up-down", "hero-arrow-top-right-on-square"],
```

enescakir:
You can move the status label next to the title. It's easy to miss it right now.

CleanShot 2025-06-17 at 12 27 53@2x

enescakir:
This padding fuels my obsession.

CleanShot 2025-06-17 at 12 29 18@2x

views/postgres/submenu.erb
enescakir: `/` makes little ugly

enescakir:
It's not related to this PR but it showing empty storage while provisioning.

CleanShot 2025-06-17 at 12 32 20@2x

views/postgres/backup_restore.erb
```diff
@@ -0,0 +1,71 @@
...
+  </div>
+<% else %>
+  <div class="p-6">
+    <div class="text-gray-500">
+      No backups available for this PostgreSQL database.
+    </div>
+  </div>
+<% end %>
+
+<% backups = @pg.timeline.backups %>
```
shikharbhardwaj, line 37: Just a note: for me, clicking on the `Backup/Restore` submenu has some lag compared to others, I guess most likely because of listing of these backups. Could be because my local setup has higher network latency to the backup MinIO.

views/postgres/show.erb
```diff
@@ -1,645 +1,33 @@
...
-            </div>
+<main>
+  <div class="max-w-screen-xl">
+    <div class="flex rounded-lg bg-white shadow divide-x p-8">
+      <div class="flex-none pr-8"><%== render("postgres/submenu") %></div>
+      <div class="flex-auto pl-8">
+        <div class="divide-y">
+          <div class="pb-6">
+            <span class="text-2xl font-bold leading-7 text-gray-900 sm:text-3xl sm:tracking-tight"><%= @pg.name %></span>
+            (id:
```
shikharbhardwaj: A couple of UI nits I could spot on my machine:
1. ~~The DB name appears right beside the submenu, without any margin to its left~~
This got fixed with `npm run prod`
2. The page does not fill up to the viewport

image

views/components/revealable_content.erb
```diff
@@ -1,12 +1,12 @@
 <%# locals: (content:) %>
-<div class="revealable-content group">
-  <div class="shadow-content items-center gap-2 flex group-[.active]:hidden">
+<div class="revealable-content group flex-grow">
+  <div class="shadow-content justify-between items-center gap-2 flex group-[.active]:hidden">
     ●●●●●●
     <img src="/icons/hero-eye.svg" width="18px" class="revealable-button inline-block cursor-pointer ms1"/>
```
shikharbhardwaj, line 5: UI nit: the reveal icon moves after interaction, maybe it should be beside the copy button?

out

shikharbhardwaj (approved):
Looks good, I added a few UI nits I could spot when using it on my dev env

views/components/revealable_content.erb
shikharbhardwaj: This got fixed with an `npm run prod`, ignore

views/postgres/backup_restore.erb
byucesoy: We have maximum of 8 backups usually. Still. I think it would be better to skip the map and call last_modified where necessary.

routes/project/location/postgres.rb
byucesoy: Noted for future a PR.

routes/project/location/postgres.rb
byucesoy: Makes sense. I started doing this, but then ended up with lots of conflicts in the subsequent commits and other PRs that depends on this. I'll make this refactor as another PR.

views/postgres/backup_restore.erb
byucesoy: Hmm, another cache maybe? :slightly_smiling_face: 

Leaving that to different PR though.

byucesoy:
> This padding fuels my obsession.
> 
> CleanShot 2025-06-17 at 12 29 18@2x

This was intentional. It exists for all sub pages. Think of it like an indentation for sub headings. Also I'm not sure if the alternative is better.

|With Padding | Without Padding |
| ------------------- | ----------------- |
| Screenshot from 2025-06-19 04-43-30  | Screenshot from 2025-06-19 04-43-25  |

views/postgres/charts.erb
byucesoy: We need the column, because at the bottom row, we have inputs to add a new metric destination. I could leave this cell empty, but that doesn't look good. That said, I agree that when we put `●●●●●●`, there might be some expectation for a way to revealing it. I changed it to `-`

PR #3449, opened 2025-06-13
Use Ubiblk for volumes if installed on the host.

pykello:
If any Ubiblk installations are enabled on a VmHost, the allocator will choose one at random, weighted by `allocation_weight`. Otherwise, SPDK will be used.

Currently, Ubiblk is not installed by default and must be set up manually by an operator. Therefore, the default behavior in production is to continue using SPDK.

Note: The values for `num_queues` and `queue_size` are provisional and may change based on future performance and resource usage experiments.


scheduling/allocator.rb
```diff
@@ -570,6 +570,19 @@ def self.allocate_spdk_installation(spdk_installations)
       rand_choice.id
     end
 
+    def self.allocate_vhost_block_backend(backends)
+      total_weight = backends.sum(&:allocation_weight)
+      fail "Total weight of all eligible vhost_block_backends shouldn't be zero." if total_weight == 0
+
+      rand_point = rand(0...total_weight)
```
jeremyevans:
```suggestion
      rand_point = rand(total_weight)
```

No need for a range here.

scheduling/allocator.rb
```diff
@@ -597,7 +610,15 @@ def map_volumes_to_devices
 
     def create_storage_volumes(vm, vm_host)
       @request.storage_volumes.each do |disk_index, volume|
-        spdk_installation_id = StorageAllocation.allocate_spdk_installation(vm_host.spdk_installations)
+        if vm_host.vhost_block_backends_dataset.exclude(allocation_weight: 0).empty?
+          spdk_installation_id = StorageAllocation.allocate_spdk_installation(vm_host.spdk_installations)
+          vhost_block_backend_id = nil
```
jeremyevans: If you want to include this line to be explicit, that's fine.  Just note that it isn't required.  It will automatically be nil if this branch is taken, because the other branch attempts to define it:

```
$ ruby -e 'if ARGV.empty?; a = 1; else b = 1; end; p [a,b]'
[1, nil]

ruby -e 'if ARGV.empty?; a = 1; else b = 1; end; p [a,b]' foo
[nil, 1]
```

Same with `spdk_installation_id` in the else block.

jeremyevans (approved):

PR #3462, opened 2025-06-17
Simplify Options Usage for PostgreSQL (SOUP)

byucesoy:
The option.rb file for PostgreSQL (and other products) has become increasingly difficult to maintain over time. We've followed various approaches throughout its evolution, and as more features were added, those approaches began to overlap and resulted in a tangled mess of spaghetti code.

This PR aims to clean things up by:
- Keeping only globally valid options in option.rb and removing all embedded business logic.
- Moving the business logic into generate_postgres_options, where it was originally intended to reside.
- Leveraging the option tree for validations, eliminating the need for separate validation functions for each parameter.

----

> [!IMPORTANT]
> Centralizes PostgreSQL option handling and validation in `generate_postgres_options`, improving maintainability and reducing redundancy.
> 
>   - **Behavior**:
>     - Centralizes PostgreSQL option business logic in `generate_postgres_options`.
>     - Uses option tree for validations, removing redundant validation functions.
>   - **Code Changes**:
>     - Updates `postgres_post` in `postgres.rb` to use centralized option handling.
>     - Refactors `generate_postgres_options` to include all business logic and validations.
>     - Removes specific validation functions from `validation.rb`.
>   - **Misc**:
>     - Updates CLI commands and tests to reflect new option handling.
>     - Renames and restructures option constants in `option.rb` for clarity.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 052f63daad97c877f4babba88bc9b908c8ffa73a. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


lib/option.rb
```diff
@@ -78,42 +70,42 @@ def display_name
...
+  POSTGRES_SIZE_OPTIONS = [
+    ["standard", 2, 8],
+    ["standard", 4, 16],
+    ["standard", 8, 32],
+    ["standard", 16, 64],
+    ["standard", 30, 120],
+    ["standard", 60, 240],
+    ["burstable", 1, 2],
+    ["burstable", 2, 4]
+  ].map { |args| PostgresSizeOption.new("#{args[0]}-#{args[1]}", *args) }.freeze
```
enescakir: I think we can make it a bit clearer. It's hard to keep track of positional arguments over time. It also applies to other new global PG options.

```suggestion
  ].map { |family, vcpus_count, memory_gib| PostgresSizeOption.new("#{family}-#{vcpus_count}", family:, vcpu_count:, memory_gib:) }.freeze
```

lib/option.rb
```diff
@@ -138,6 +138,13 @@ def display_name
 
   POSTGRES_VERSION_OPTIONS = ["17", "16"].freeze
 
+  PostgresHaOption = Data.define(:name, :standby_count, :description)
+  POSTGRES_HA_OPTIONS = [
```
enescakir, line 104: Since we use it with `.find {}` all the time, why don't we keep it as a `name => PostgresHaOption` hash?

lib/option.rb
```diff
@@ -108,12 +108,9 @@ def display_name
...
     PostgresHaOption.new(*it)
   }.freeze
+  # Postgres Global Options
 
-  POSTGRES_VERSION_OPTIONS = {
-    PostgresResource::Flavor::STANDARD => ["17", "16"],
-    PostgresResource::Flavor::PARADEDB => ["17", "16"],
-    PostgresResource::Flavor::LANTERN => ["16"]
-  }
+  POSTGRES_VERSION_OPTIONS = ["17", "16"].freeze
```
jeremyevans, line 101: Are we sure that future flavors will not be version dependent, and all new PostgreSQL versions will be supported by existing flavors?  Or is the plan to just revert this change if that happens?

helpers/postgres.rb
```diff
@@ -95,9 +95,13 @@ def generate_postgres_options(flavor: "standard", location: nil)
...
-      if location.provider == "aws" && (size.split("-").last.to_i > 16 || size.split("-").first == "burstable")
-        false
+    options.add_option(name: "size", values: Option::POSTGRES_SIZE_OPTIONS.map(&:name), parent: "family") do |flavor, location, family, size|
+      family_from_size, vcpu_count = size.split("-")
+
+      next false if family_from_size != family
+      next false if location.provider == "aws" && vcpu_count.to_i > 16
+      true
+    end
       else
```
jeremyevans: This looks like a syntax error.  It is fixed in 4337a1a6534549e9bccce7212e40d23bf0596248 , but it's best to have valid syntax in each commit.

helpers/postgres.rb
```diff
@@ -102,18 +101,22 @@ def generate_postgres_options(flavor: "standard", location: nil)
...
     end
+
+    aws_storage_size_options = ["118", "237", "475", "950"]
+    storage_size_options = Option::POSTGRES_STORAGE_SIZE_OPTIONS + aws_storage_size_options
+    options.add_option(name: "storage_size", values: storage_size_options, parent: "size") do |flavor, location, family, size, storage_size|
+      vcpu_count = size.split("-").last.to_i
+
+      if location.provider == "aws"
+        storage_index = Math.log2(vcpu_count).ceil - 1
+        aws_storage_size_options[storage_index] == storage_size
```
jeremyevans: It ma be simpler to have `aws_storage_size_options` be a hash with `vcpu_count` keys and `storage_size` values.

lib/validation.rb
```diff
@@ -325,4 +325,23 @@ def self.validate_rfc3339_datetime_str(datetime_str, param = "time")
...
+
+      parents.each do |parent|
+        parent_value = params[parent]
+        subtree = subtree[parent][parent_value]
+      end
+
+      if subtree[key][value].nil?
+        display_key = key.tr("_", " ")
+        value = value.is_a?(Location) ? value.display_name : value.to_s
+        available_options = subtree[key].keys.map! { it.is_a?(Location) ? it.display_name : it.to_s }.join(", ")
```
jeremyevans: Not sure how often we do this, but if it is in more than these 2 places, it would be good to have a method for the `display_name` or `to_s` logic to DRY things up.

lib/validation.rb
```diff
@@ -325,4 +325,23 @@ def self.validate_rfc3339_datetime_str(datetime_str, param = "time")
...
+        fail ValidationFailed.new({key.to_sym => "\"#{value}\" is not a valid #{display_key}. Available options: #{available_options}"})
```
jeremyevans: I would avoid using the `value` in the error message.  If you must use it, it's better to use `value.inspect`.

lib/option.rb
jeremyevans: Pretty much every place you are calling `find { it.name == name }` on an array constant, you should probably be using a hash instead. If you really need a separate array constant: `ARRAY_CONSTANT = HASH_CONSTANT.keys.freeze`

jeremyevans (approved):
I think this is a big improvement.  I especially like the use of the option tree for validation.

lib/option.rb
byucesoy: I agree this in principle, but I don't think this makes it any easier to read in this case. You still need to read above on to understand the potential value of each field.

lib/option.rb
byucesoy: Version will likely to depend on flavor in the future. The plan is reverting that at that time.

jeremyevans: Seems like this is likely to cause unnecessary churn later, but I don't object if you feel the benefits exceed the costs.

PR #3464, opened 2025-06-17
Add multiple regions for Ubi-on-AWS and refactor to use S3 sdk for AWS

furkansahin:
## Introduce multiple S3 regions with AMI integration
This commit basically refactors the ami resolution code to make use of
the new PgAwsAmi entity. Since we have multiple regions and multiple
versions of pg, we need to be able to programmatically modify that
logic. If we continued the old way, we would have constantly added new
config variables.

## Use AWS S3 SDK client for bucket operations
We used to use our MinIO client to interract with AWS S3, too. However,
that has a bunch of problems because we manually created that to talk to
our own minio clusters. Now, introducing multiple regions, we have to
either dive deeper to refactor the client or we could simply introduce
the aws sdk. I chose latter considering it's much cheaper for


config/billing_rates.yml
```diff
@@ -163,6 +161,36 @@
...
+- { id: 41866dbe-b7a1-8d78-9c22-d6a6a3ff3abb, resource_type: IPAddress,                   resource_family: IPv4,                       location: us-east-2,      unit_price: 0.0000000000, billed_by: duration, active_from: 2023-06-17T00:00:00Z }
+- { id: 42eb5c58-f2a8-8178-8010-b5883699544b, resource_type: PostgresVCpu,                resource_family: standard-standard,          location: us-west-2,      unit_price: 0.0005518353, billed_by: duration, active_from: 2025-06-17T00:00:00Z }
+- { id: d53861e2-a2ce-8578-9e79-3fae0d3b26c2, resource_type: PostgresVCpu,                resource_family: standard-standard,          location: us-east-2,      unit_price: 0.0006696429, billed_by: duration, active_from: 2025-06-17T00:00:00Z }
+- { id: 924c49d9-99d9-8d78-b819-11769d2f6293, resource_type: PostgresVCpu,                resource_family: paradedb-standard,          location: us-west-2,      unit_price: 0.0011036706, billed_by: duration, active_from: 2025-06-17T00:00:00Z }
```
byucesoy: Can we remove
- paradedb and lantern billing meters
- non us-west-2 billing meters

Because we don't support those configurations right now and our pricing is not finalized for AWS. Thus I don't want to create extra meters that we would deprecate.

config/billing_rates.yml
```diff
@@ -138,14 +138,12 @@
...
 - { id: 6d88457f-4241-8178-9205-274ebea1d19f, resource_type: PostgresVCpu,                resource_family: lantern-standard,           location: us-west-1,      unit_price: 0.0008277530, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
 - { id: 025e5656-3eca-8978-879c-0958a0ae93c2, resource_type: PostgresVCpu,                resource_family: lantern-standard,           location: us-east-1,      unit_price: 0.0010044644, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
 - { id: 2a7fb1ca-02c0-8578-97bb-89df451cffe9, resource_type: PostgresStandbyVCpu,         resource_family: standard-standard,          location: us-west-1,      unit_price: 0.0005518353, billed_by: duration, active_from: 2025-02-17T00:00:00Z }
-- { id: f921f3c5-709d-4692-b31d-88c61cfe66bb, resource_type: PostgresStandbyVCpu,         resource_family: standard-standard,          location: us-west-2,      unit_price: 0.0010075640, billed_by: duration, active_from: 2025-05-10T00:00:00Z }
```
byucesoy, line 148: The rough grouping/sorting in this file is first based on resource_type then resource_family and location. Can we keep that instead of moving meters to the bottom?

config/billing_rates.yml
```diff
@@ -163,6 +161,36 @@
...
+- { id: 0e73d783-3c9e-8178-995a-4ecc2cd84963, resource_type: VmStorage,                   resource_family: standard,                   location: us-east-2,      unit_price: 0.0000000000, billed_by: duration, active_from: 2025-06-17T00:00:00Z }
+- { id: a3e77384-102b-8578-b465-73b7892710c7, resource_type: IPAddress,                   resource_family: IPv4,                       location: us-west-2,      unit_price: 0.0000000000, billed_by: duration, active_from: 2023-06-17T00:00:00Z }
+- { id: 41866dbe-b7a1-8d78-9c22-d6a6a3ff3abb, resource_type: IPAddress,                   resource_family: IPv4,                       location: us-east-2,      unit_price: 0.0000000000, billed_by: duration, active_from: 2023-06-17T00:00:00Z }
+- { id: 42eb5c58-f2a8-8178-8010-b5883699544b, resource_type: PostgresVCpu,                resource_family: standard-standard,          location: us-west-2,      unit_price: 0.0005518353, billed_by: duration, active_from: 2025-06-17T00:00:00Z }
```
byucesoy: It seems we reduced the price of PostgresVCpu and add another meter for PostgresStorage. Is that correct? What is the math behind it?

ubid.rb
```diff
@@ -101,6 +101,7 @@ class UBID
   TYPE_NIC_AWS_RESOURCE = "na"
   TYPE_VICTORIA_METRICS_RESOURCE = "vr"
   TYPE_VICTORIA_METRICS_SERVER = "vn"
+  TYPE_PG_AWS_AMI = "aa"
```
byucesoy: Maybe use TYPE_ETC since this is not a critical type?

model/postgres/postgres_timeline.rb
```diff
@@ -104,17 +104,60 @@ def blob_storage_endpoint
...
+  end
+
+  def set_lifecycle_policy
+    aws? ?
+      blob_storage_client.put_bucket_lifecycle_configuration({
+        bucket: ubid,
+        lifecycle_configuration: {
+          rules: [
+            {
+              id: "DeleteOldBackups",
```
byucesoy, line 149: I believe we need 2 policies for AWS. One for expiring and one for deleting the files. Not 100% sure. Could you please double check?

ubid.rb
jeremyevans: Please only use TYPE_ETC is you are 100% sure it will never be exposed to the user.

byucesoy: I'm pretty sure this won't be exposed to user. 100% is a very high bar though :smile:

model/postgres/postgres_timeline.rb
furkansahin: We're not using versioned buckets that means an expired object is also queued for deletion.

config/billing_rates.yml
furkansahin: That was an error on my end.

ubid.rb
furkansahin: Switched to TYPE_ETC

model/postgres/postgres_timeline.rb
```diff
@@ -104,17 +104,60 @@ def blob_storage_endpoint
...
 
   def blob_storage_client
-    @blob_storage_client ||= Minio::Client.new(
+    @blob_storage_client ||= aws? ? Aws::S3::Client.new(
+      region: location.name,
+      access_key_id: access_key,
+      secret_access_key: secret_key,
+      endpoint: blob_storage_endpoint,
+      force_path_style: true
+    ) : Minio::Client.new(
```
enescakir, line 113: I have a high level question. Should we deprecate `Minio::Client`? We started implementing it to avoid the large AWS SDK, but we couldn't complete the required feature set. We're already using `Aws::S3::Client` for runner caches, and we're starting to use it for Ubi-on-AWS as well. So, we can actually stop using `Minio::Client` and use `Aws::S3::Client` for everything.

furkansahin: no we cannot, administrator related calls cannot be made via aws s3 sdk

prog/vnet/update_firewall_rules.rb
```diff
@@ -187,61 +187,28 @@ def before_run
...
-      perm = {
-        ip_protocol: "tcp",
-        from_port: permission.from_port,
-        to_port: permission.to_port
-      }
-      perm[:ip_ranges] = permission.ip_ranges if permission.ip_ranges.any?
-      perm[:ipv_6_ranges] = permission.ipv_6_ranges if permission.ipv_6_ranges.any?
-      perm
+    security_group.ip_permissions.each do |permission|
+      aws_client.revoke_security_group_ingress({
```
byucesoy: Doesn't removing all permissions and then adding new permissions, means that we are not allowing new connections momentarily, even from the allowed IP addresses? Also this is not atomic right, so if something fails in between that short time can be even longer (until we try it next time).

I think we need to compare the current allowed IPs with the target and decide what to do with each rule (add, remove, skip).

furkansahin: Updated

PR #3465, opened 2025-06-17
Use NUL delimiter for xargs in github actions

fdr:
Some `encoded_jit_config` values can interact badly with `xargs`'s quoting rules, which are not of that of the shell, but of xargs's own, similar design (see the first few paragraphs of the manual page)

In 29cbec28c, I packed the token in `xargs` to avoid it being emitted in our log telemetry, as we don't save stdin, because we customarily use it for secrets.

Because we're only packing a single value for a single invocation of the subcommand, we can use the NUL byte delimitation mode of `xargs` to evade its own interpretation of spaces and quotes, so long as `encoded_jit_config` is not binary. That no NUL byte appears in stdin is fine: the end-of-file condition when emitting the token is enough.

----

> [!IMPORTANT]
> Change `xargs` command in `github_runner.rb` to use NUL byte delimiter for robust handling of `encoded_jit_config`.
> 
>   - **Behavior**:
>     - Change `xargs` command in `github_runner.rb` to use `-0` option for NUL byte delimiter, ensuring robust handling of `encoded_jit_config`.
>   - **Tests**:
>     - Update test in `github_runner_spec.rb` to expect `xargs -0I{}` command format.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7db8eadc8a58a669a421dc50246adea06c604c6f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -283,7 +283,7 @@ class Prog::Vm::GithubRunner < Prog::Base
 
     # We initiate an API call and a SSH connection under the same label to avoid
     # having to store the encoded_jit_config.
-    vm.sshable.cmd("sudo -- xargs -I{} -- systemd-run --uid runner --gid runner " \
+    vm.sshable.cmd("sudo -- xargs -0I{} -- systemd-run --uid runner --gid runner " \
                    "--working-directory '/home/runner' --unit runner-script --remain-after-exit -- " \
                    "/home/runner/actions-runner/run-withenv.sh {}",
       stdin: response[:encoded_jit_config])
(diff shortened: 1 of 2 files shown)


enescakir:
GitHub is currently experiencing an incident, so it can't run the E2E test suite properly.
https://www.githubstatus.com/incidents/y7lb2rg4btd7

enescakir (approved):

PR #3467, opened 2025-06-17
Split VM creation page into GPU/non-GPU if get_ff_gpu_vm feature flag is set

jeremyevans:
This can potentially make it easier to select a GPU instance.

With the feature flag set, instead of one button link to the create page, there are two:
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_vm
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_vm (1)

The "Create Virtual Machine" link does not show any GPU options:

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_vm_create_gpu=f

The "Create GPU Virtual Machine" only shows valid options for a GPU configuration:

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_vm_create_gpu=true

If users have a direct link to the old page, it works the same as before, showing both GPU and non GPU options:

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_vm_create

Ignore the fact that the screenshots show the GPU VM in Finland.  My test host is in Finland, and I had to add billing rates to get things to work in development.

Requested by @umurc

----

> [!IMPORTANT]
> Split VM creation page into GPU and non-GPU options based on feature flag, with UI updates and tests.
> 
>   - **Behavior**:
>     - Split VM creation page into GPU and non-GPU options if `get_ff_gpu_vm` feature flag is set.
>     - `generate_vm_options` in `helpers/vm.rb` now considers `@show_gpu` to filter GPU options.
>     - Redirects to error if no GPU VMs are available when `show_gpu` is true.
>   - **UI Changes**:
>     - `views/vm/index.erb` and `views/vm/create.erb` updated to show separate buttons for "Create Virtual Machine" and "Create GPU Virtual Machine" based on feature flag.
>     - `views/components/empty_state.erb` supports multiple buttons.
>   - **Tests**:
>     - Added tests in `spec/routes/web/vm_spec.rb` to verify GPU and non-GPU VM creation paths and error handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 702ec88424bea2378bcb8139b2a1be2037abccea. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/views/components/empty_state.erb
+++ b/views/components/empty_state.erb
@@ -1,11 +1,15 @@
-<%# locals: (title:, icon:, description:, button_link: nil, button_title: nil) %>
+<%# locals: (title:, icon:, description:, button_link: nil, button_title: nil, buttons: nil) %>
 
 <div class="text-center empty-state">
   <%== part("components/icon", name: icon, classes: "mx-auto h-12 w-12 text-gray-400") %>
   <h3 class="mt-2 text-sm font-semibold text-gray-900"><%= title %></h3>
   <p class="mt-1 text-sm text-gray-500"><%= description %></p>
   <div class="mt-6">
-    <% if button_link %>
+    <% if buttons %>
+      <% buttons.each do |text, link| %>
+        <%== part("components/button", link:, text:) %>
+      <% end %>
+    <% elsif button_link %>
       <%== part("components/button", link: button_link, text: button_title) %>
     <% elsif button_title %>
       <div class="flex justify-center">
--- a/helpers/vm.rb
+++ b/helpers/vm.rb
@@ -132,9 +132,13 @@ class Clover
     end
 
     options.add_option(name: "gpu", values: ["0:"] + gpu_options, parent: "family") do |location, family, gpu|
-      gpu = gpu.split(":")
-      gpu_count = gpu[0].to_i
-      gpu_count == 0 || (family == "standard" && !!BillingRate.from_resource_properties("Gpu", gpu[1], location.name) && gpu_availability[location.name] && gpu_availability[location.name][gpu[1]] && gpu_availability[location.name][gpu[1]] >= gpu_count)
+      gpu_count, device = gpu.split(":", 2)
+      gpu_count = gpu_count.to_i
+      next true if gpu_count == 0
+
+      family == "standard" &&
+        !!BillingRate.from_resource_properties("Gpu", device, location.name) &&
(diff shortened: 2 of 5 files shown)


byucesoy:
> Ignore the fact that the screenshots show the GPU VM in Finland. My test host is in Finland, and I had to add billing rates to get things to work in development.

I used to have development host in Finland. As a workaround, I added it as if it was in Germany while calling VmHostNexus.assemble. The location does not get used in any of our API calls to Hetzner, so using "wrong" location does not break anything. In that way, you wouldn't need to change any code.

> If users have a direct link to the old page, it works the same as before, showing both GPU and non GPU options:

If means less code to maintain, I think it is OK to not have backwards compatibility on this one. But that is my personal opinion.

views/components/empty_state.erb
```diff
@@ -1,11 +1,15 @@
-<%# locals: (title:, icon:, description:, button_link: nil, button_title: nil) %>
+<%# locals: (title:, icon:, description:, button_link: nil, button_title: nil, buttons: nil) %>
```
byucesoy, line 1: Being able to handle array also covers the case for single button. Maybe remove `button_link` and `button_title` and update callers to use `buttons` array?

jeremyevans: That sounds fine, but it would be a separate PR after this is merged.

jeremyevans:
> > Ignore the fact that the screenshots show the GPU VM in Finland. My test host is in Finland, and I had to add billing rates to get things to work in development.
> 
> I used to have development host in Finland. As a workaround, I added it as if it was in Germany while calling VmHostNexus.assemble. The location does not get used in any of our API calls to Hetzner, so using "wrong" location does not break anything. In that way, you wouldn't need to change any code.

Good to know.  I made that change in my development environment.

> > If users have a direct link to the old page, it works the same as before, showing both GPU and non GPU options:
> 
> If means less code to maintain, I think it is OK to not have backwards compatibility on this one. But that is my personal opinion.

It would probably be a bit simpler, but considering I already spent time keeping it backwards compatible, I don't think it makes sense to remove the backwards compatibility.

PR #3471, opened 2025-06-18
More respirate improvements

jeremyevans:
Improvements:

* For strands with expired leases, calculate delay based on lease time instead of schedule time, to prevent the delay numbers showing 120 second spikes when processing a strand with an expired lease.

* Make respirate sleep time dependent on dispatcher state.  This shouldn't do much in production currently, but it is necessary step to prevent starvation when decreasing the queue size ratio, which is something I'd like to try to see if it results in lower overall delay.

----

> [!IMPORTANT]
> Improve `respirate` by handling expired leases and dynamically adjusting sleep duration based on dispatcher state.
> 
>   - **Behavior**:
>     - Calculate delay based on lease time for expired leases in `Strand#respirate_metrics`.
>     - Adjust `respirate` sleep duration based on dispatcher state in `Dispatcher#sleep_duration`.
>   - **Config**:
>     - Change `dispatcher_queue_size_ratio` type from `int` to `float` in `config.rb`.
>   - **Metrics**:
>     - Add `lease_expired_percentage` to metrics in `Dispatcher#metrics_hash`.
>   - **Tests**:
>     - Add tests for expired lease handling in `dispatcher_spec.rb`.
>     - Add tests for dynamic sleep duration in `dispatcher_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 155770cebd12c2881ec9c1739fa8468ad32692c6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bin/respirate
+++ b/bin/respirate
@@ -69,7 +69,7 @@ DB.synchronize do
       # Note that this results in an up to a 1 second delay to pick up new strands,
       # but that is an accept tradeoff, because we do not want to busy loop the
       # database (potentially, LISTEN/NOTIFY could be used to reduce this latency)
-      duration_slept = sleep 1
+      duration_slept = sleep d.sleep_duration
       Clog.emit("respirate finished sleep") { {sleep_duration_sec: duration_slept} }
     end
   end
--- a/config.rb
+++ b/config.rb
@@ -79,7 +79,7 @@ module Config
   override :pretty_json, false, bool
   override :dispatcher_max_threads, 8, int
   override :dispatcher_min_threads, 1, int
-  override :dispatcher_queue_size_ratio, 4, int
+  override :dispatcher_queue_size_ratio, 4, float
   override :puma_max_threads, 16, int
   override :puma_min_threads, 1, int
   override :puma_workers, 3, int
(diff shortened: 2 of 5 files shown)


jeremyevans:
Added a couple more metrics:

* lease_expired_percentage: With the earlier change to better calculate delay, expired leases are not obvious from delay spikes. We should still be tracking them.

* current_strand_delay: Usually the same as p95 total delay, but can be different if the process is processing old strands.

This also allows for fractional queue size ratios.  I plan on dropping from 4 to 2 and see how it goes.  If it goes well, we might go from 2 to 1.5 and see the results of that.  Dropping straight from 2 to 1 is more risk than I would prefer.

fdr (approved):
alternative abstract: adaptive scan polling, I suppose?

jeremyevans:
Rebased and removed the "Do not push to the metrics queue if shutting down", as that was handled a different way in 1fcb10a6a291373d7be7b8344cdc24b365e9a997.

PR #3474, opened 2025-06-19
PG configuration backend

shikharbhardwaj:
## Description
Adds database migration and API to manage PG configuration

----

> [!IMPORTANT]
> Adds PostgreSQL and PgBouncer configuration management with validation, API endpoints, and database schema changes, including comprehensive test coverage.
> 
>   - **Behavior**:
>     - Adds `PostgresConfigValidator` in `postgres_config_validator.rb` for validating PostgreSQL and PgBouncer configurations.
>     - Introduces new API endpoints in `openapi.yml` and `postgres.rb` for managing PostgreSQL configurations.
>     - Updates `configure` script to apply user configurations to PostgreSQL and PgBouncer.
>   - **Database**:
>     - Adds `user_config` and `pgbouncer_user_config` columns to `postgres_resource` table in `20250701_add_pg_configs.rb`.
>   - **Validation**:
>     - Defines configuration schemas in `postgres_config_validator_schema.rb` for PostgreSQL versions 16, 17, and PgBouncer.
>     - Implements validation logic in `PostgresConfigValidator`.
>   - **Testing**:
>     - Adds tests for configuration validation in `postgres_config_validator_spec.rb` and `postgres_config_validator_schema_spec.rb`.
>     - Adds API tests for configuration management in `postgres_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c87506d2758f0e8438d993dfe3c8661a1ac00692. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


routes/project/location/postgres.rb
```diff
@@ -453,6 +453,40 @@ class Clover
...
+          pg_validator = PostgresConfig::Validator.new(pg.version)
+          pg_validator.validate(pg_config)
+
+          pgbouncer_validator = PostgresConfig::Validator.new("pgbouncer")
+          pgbouncer_validator.validate(pgbouncer_config)
+
+          pg.update(user_config: pg_config, pgbouncer_user_config: pgbouncer_config)
+          audit_log(pg, "update")
+
+          pg.servers.each(&:incr_configure)
```
shikharbhardwaj, line 493: Need to check if this can trigger a restart

cache/schema.cache
jeremyevans, line 1: Be aware that this will make the application assume the columns actually exist before the migration is run.  If you want to ensure this works correctly, you need to try this commit in isolation, without running the related migration, and see if anything breaks.

spec/routes/api/project/location/postgres_spec.rb
```diff
@@ -647,5 +647,22 @@
         expect(last_response.status).to eq(204)
       end
     end
+
+    describe "config" do
+      it "read" do
```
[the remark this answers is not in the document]
jeremyevans: I'd use stronger wording. This type of test that submits a request and only checks response status and does not check actual results is harmful, as it will show the code as covered, even though the only real test is that the route doesn't raise an exception.

The get request should be checking the response body matches what is expected. The update request should check the response body, and also check that the expected changes were made to the database.

routes/project/location/postgres.rb
```diff
@@ -453,6 +453,40 @@ class Clover
           metrics: results
         }
       end
+
+      r.on "config" do
+        r.get true do
```
```suggestion
      r.is "config" do
        r.get do
```

`r.on` is used if you want to support additional segments of the request path in this branch. As that does not appear to be the case, it's best to use `r.is`.  With `r.is`, the check for a fully routed path is done once at the branch level, so it doesn't need to be done separately in both the `r.get` and `r.put` calls (that is why the `true` argument is dropped).

routes/project/location/postgres.rb
```diff
@@ -453,6 +453,40 @@ class Clover
...
+          authorize("Postgres:view", pg.id)
+
+          {
+            pg_config: pg.user_config,
+            pgbouncer_config: pg.pgbouncer_user_config
+          }
+        end
+
+        r.put true do
```
```suggestion
        r.put do
```

cache/schema.cache
shikharbhardwaj: Interesting, I thought deploying this commit separately should be safe, if the code does not access the column (the API is added in a separate commit). I'll try this separately to make sure.

Just curious: which parts of the stack could potentially break in such a case (assuming none of the API/web code paths access the new column)?

spec/routes/api/project/location/postgres_spec.rb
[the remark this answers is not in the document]
shikharbhardwaj: Done, added tests to validate reads/updates through to the DB

cache/schema.cache
jeremyevans: I'm not sure that anything would break. However, I'm also not sure that nothing would break.  Better to spend a few minutes testing to get confidence that things won't break.

lib/validation/postgres_config_validator.rb
```diff
@@ -0,0 +1,4717 @@
+# frozen_string_literal: true
+
+module Validation
+  class PostgresConfigValidator
+    PG_17_CONFIG_SCHEMA = {
```
byucesoy: Let's put these into a separate JSON files (one for each version) instead of embedding them into the code.

lib/validation/postgres_config_validator.rb
```diff
@@ -0,0 +1,4717 @@
...
+      },
+      "TimeZone" => {
+        description: "Sets the time zone for displaying and interpreting time stamps.",
+        type: :string,
+        default: "UTC"
+      },
+      "allow_alter_system" => {
+        description: "Allows running the ALTER SYSTEM command.",
+        type: :enum,
+        allowed_values: ["on", "off"],
```
byucesoy: Postgres usually allows other type of boolean indicators such as true, false, 0, 1 etc.

lib/validation/postgres_config_validator.rb
```diff
@@ -0,0 +1,4717 @@
...
+        allowed_values: ["content", "document"],
+        default: "content"
+      },
+      "zero_damaged_pages" => {
+        description: "Continues processing past damaged page headers.",
+        type: :enum,
+        allowed_values: ["on", "off"],
+        default: "off"
+      }
+    }
```
byucesoy: Extensions can also have config variables. They are usually namespaced with the extensions' name like `citus.shard_count`. Maybe we should allow free text configs if they contain dot in the name. Or maybe we should add configs of some of the popular extensions to this list. Or maybe something else.

Any ideas on how to handle extension configs?

lib/validation/postgres_config_validator.rb
byucesoy: How did you generate these JSONs?

routes/project/location/postgres.rb
```diff
@@ -453,6 +453,40 @@ class Clover
...
+          pg_config = typecast_params.Hash("pg_config")
+          pgbouncer_config = typecast_params.Hash("pgbouncer_config")
+
+          pg_validator = Validation::PostgresConfigValidator.new(pg.version)
+          pg_validator.validate(pg_config)
+
+          pgbouncer_validator = Validation::PostgresConfigValidator.new("pgbouncer")
+          pgbouncer_validator.validate(pgbouncer_config)
+
+          pg.update(user_config: pg_config, pgbouncer_user_config: pgbouncer_config)
```
byucesoy, line 490: So we expect user to pass previously changed configs while changing a new config, right? This might be OK for web request as we can populate the form with all the changes. For API, though, it seems like putting a big burden to user.

lib/validation/postgres_config_validator.rb
shikharbhardwaj: I took the list from `SHOW ALL;` and then created a script to fetch the type, min/max, allowed values etc for each param from here: https://postgresqlco.nf/doc/en/param/allow_alter_system/17

Although it wasn't perfect, needed some manual review for a few that were not there on this (for eg. removing show only parameters).

routes/project/location/postgres.rb
shikharbhardwaj: We can add a `PATCH` version for this API, which just adds/modifies the specified keys without updating existing values

cache/schema.cache
shikharbhardwaj: I tried this commit in isolation without the migration, worked okay.

lib/validation/postgres_config_validator.rb
shikharbhardwaj: Open to doing this, however in the past @fdr had suggested keeping things like these within Ruby unless the file was manipulated by external tools. Thoughts?

lib/validation/postgres_config_validator.rb
shikharbhardwaj: Since its possible to define custom properties [1], I think it makes sense to skip validation in case a parameter has 2 parts separated by a dot. PG does not care if the corresponding extension is loaded or not, and it can even be arbitrary user-defined configs.

We can keep expanding the schema for more extensions to do more strict checking for the known parameters.

[1] https://www.postgresql.org/docs/17/runtime-config-custom.html#RUNTIME-CONFIG-CUSTOM

lib/validation/postgres_config_validator.rb
shikharbhardwaj: I added a new type to handle booleans

lib/validation/postgres_config_validator.rb
byucesoy: It is 4500+ lines of file and 4400+ of it is config :slightly_smiling_face:. It just makes it difficult to read interesting part (i.e. code).  @fdr what do you think about this case? Would you be OK moving this to separate config files? I think that would increase readability.

lib/validation/postgres_config_validator.rb
byucesoy: Yeah, I think that makes sense.

routes/project/location/postgres.rb
byucesoy: Yep, sounds good.

shikharbhardwaj: I added another API for patching configs

lib/validation/postgres_config_validator.rb
shikharbhardwaj: I took another route here: I moved the schema into its own file. Let me know if this sounds better.

routes/project/location/postgres.rb
shikharbhardwaj: Looking deeper: this should only trigger a restart if `pg_ctlcluster reload #{v} main` exits with a non-zero code. This happens when:
1. The server is not running
2. There is a critical internal error (missing files/permission errors etc)

So I think its safe to increment this semaphore for a config update. We can give the user heads up that they might need to hit the restart button in case the config they're changing requires it.

lib/validation/postgres_config_validator.rb
```diff
@@ -0,0 +1,105 @@
...
+      "must be one of: #{allowed_values.join(", ")}"
+    end
+
+    def validate_string(value, pattern)
+      return nil if pattern.nil? || value.match?(Regexp.new(pattern))
+      "must match pattern: #{pattern}"
+    end
+
+    def validate_bool(value)
+      return nil if ["on", "of", "off", "t", "tr", "tru", "true", "f", "fa", "fal", "fals", "false", "1", "0"].include?(value.downcase)
```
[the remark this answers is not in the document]
shikharbhardwaj: This is not useful, the list is correct.

lib/validation/postgres_config_validator.rb
byucesoy: I think that is OK.

routes/project/location/postgres.rb
```diff
@@ -465,6 +465,63 @@ class Clover
...
+        r.get do
+          authorize("Postgres:view", pg.id)
+
+          {
+            pg_config: pg.user_config,
+            pgbouncer_config: pg.pgbouncer_user_config
+          }
+        end
+
+        r.post do
```
byucesoy, line 478: Do we ever need POST given that PATCH can do both complete and partial updates?

byucesoy (approved):

routes/project/location/postgres.rb
```diff
@@ -465,6 +465,63 @@ class Clover
...
+          pg_config = typecast_params.Hash("pg_config")
+          pgbouncer_config = typecast_params.Hash("pgbouncer_config")
+
+          pg_validator = Validation::PostgresConfigValidator.new(pg.version)
+          pg_validator.validate(pg_config)
+
+          pgbouncer_validator = Validation::PostgresConfigValidator.new("pgbouncer")
+          pgbouncer_validator.validate(pgbouncer_config)
+
+          pg.update(user_config: pg.user_config.merge(pg_config), pgbouncer_user_config: pg.pgbouncer_user_config.merge(pgbouncer_config))
```
byucesoy: We should call `compact` after `merge` to remove keys with nil values. So that users can remove configuration by sending NULL as value.

So validator should also allow nil. Alternatively, you can merge & compact (but not update) before checking validity.

routes/project/location/postgres.rb
shikharbhardwaj: PATCH cannot delete entries

routes/project/location/postgres.rb
shikharbhardwaj: Makes sense, I moved the merge before the validation to allow `nil`s

lib/validation/postgres_config_validator.rb
```diff
@@ -0,0 +1,105 @@
...
+      @config_schema.key?(key)
+    end
+
+    def validate_config(key, value)
+      config = @config_schema[key]
+      errors = []
+
+      begin
+        value = Integer(value) if config[:type] == :integer
+      rescue
```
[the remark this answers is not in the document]
jeremyevans: You can use `Integer(value, exception: false)` and then handle `nil` values separately.  This is better than using a rescue.

PR #3475, opened 2025-06-19
Serve a list of available IPv4 ranges

enescakir:
- **Serve a list of available IPv4 ranges**
  For security reasons, some of our customers need to restrict access to
  their resources by IP address. We currently share this list with them
  manually. Since it may change over time, we decided to automate the
  process and provide a simple API endpoint that returns the available
  IPv4 ranges.
  
  Instead of sharing exact IP addresses, we provide /16 CIDR ranges of
  IPv4 addresses to avoid exposing our addresses directly, which could
  make them vulnerable to attacks.
  
  Since customers may frequently call this endpoint, we don't want to run
  the query every time. Instead, we cache the result in a file and return
  the file's contents with cache headers. This way, the WAF can cache the
  result and return it to customers without hitting the API each time.
  

- **Repopulate IPv4 cache file every hour**

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -668,6 +668,12 @@ class Clover < Roda
   # :nocov:
 
   route do |r|
+    r.is "ips-v4" do
+      response["content-type"] = "text/plain"
+      response["cache-control"] = "public, max-age=3600"
+      next File.read("var/ips-v4.txt")
+    end
+
     if api?
       unless /\ABearer:?\s+pat-/i.match?(env["HTTP_AUTHORIZATION"].to_s)
         if r.path_info == "/cli"
--- a/lib/util.rb
+++ b/lib/util.rb
@@ -87,4 +87,12 @@ module Util
   def self.send_email(...)
     EmailRenderer.sendmail("/", ...)
   end
+
+  def self.populate_ipv4_txt
+    ips = Address.where { (family(cidr) =~ 4) & (routed_to_host_id !~ id) }
+      .map { NetAddr::IPv4Net.new(it.cidr.network, NetAddr::Mask32.new(16)).to_s }
+    ips.uniq!
+    ips.sort!
+    File.open("var/ips-v4.txt", "w") { it.write(ips.join("\n")) }
+  end
 end
--- a/spec/prog/populate_ipv4_cache_spec.rb
+++ b/spec/prog/populate_ipv4_cache_spec.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative "../model/spec_helper"
+
+RSpec.describe Prog::PopulateIpv4Cache do
+  subject(:pc) {
+    described_class.new(Strand.new(prog: "PopulateIpv4Cache"))
+  }
+
+  describe "#wait" do
+    it "populates file and naps an hour" do
+      expect(Util).to receive(:populate_ipv4_txt).and_call_original
+      expect { pc.wait }.to nap(60 * 60)
+    end
+  end
+end
--- a/spec/routes/web/clover_web_spec.rb
+++ b/spec/routes/web/clover_web_spec.rb
@@ -117,4 +117,18 @@ RSpec.describe Clover do
 
     expect(failures).to be_empty
   end
+
+  it "returns ipv4 list from cached file" do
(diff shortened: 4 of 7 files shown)


clover.rb
```diff
@@ -668,6 +668,12 @@ def assets
   # :nocov:
 
   route do |r|
+    r.is "ips-v4" do
```
jeremyevans, line 671: I don't think we want this for the api/runtime routes.  Can you move it near the `r.public` and `r.assets` calls? Also, it should be `r.get`, not `r.is`, since we should only accept GET requests for this route.

lib/util.rb
```diff
@@ -87,4 +87,12 @@ def self.safe_write_to_file(filename, content)
   def self.send_email(...)
     EmailRenderer.sendmail("/", ...)
   end
+
+  def self.populate_ipv4_txt
+    ips = Address.where { (family(cidr) =~ 4) & (routed_to_host_id !~ id) }
+      .map { NetAddr::IPv4Net.new(it.cidr.network, NetAddr::Mask32.new(16)).to_s }
+    ips.uniq!
+    ips.sort!
+    File.open("var/ips-v4.txt", "w") { it.write(ips.join("\n")) }
```
jeremyevans, line 96:
```suggestion
    File.write("var/ips-v4.txt", ips.join("\n"))
```

This is assuming you wanted to store this in a file, which as I point out in a different comment, will not work with your update strategy.

enescakir:
Refresh logic will not work since we have multiple dynos

lib/util.rb
```diff
@@ -87,4 +87,12 @@ def self.safe_write_to_file(filename, content)
   def self.send_email(...)
     EmailRenderer.sendmail("/", ...)
   end
+
+  def self.populate_ipv4_txt
+    ips = Address.where { (family(cidr) =~ 4) & (routed_to_host_id !~ id) }
+      .map { NetAddr::IPv4Net.new(it.cidr.network, NetAddr::Mask32.new(16)).to_s }
+    ips.uniq!
+    ips.sort!
```
jeremyevans, line 95: This does an alphabetic sort, so `9.9.0.0` comes after `10.10.0.0`.  It may be better to sort by address:

```suggestion
      .map { NetAddr::IPv4Net.new(it.cidr.network, NetAddr::Mask32.new(16)) }
    ips.sort!(&:cmp)
    ips.map!(&:to_s)
    ips.uniq!
```

spec/routes/web/clover_web_spec.rb
```diff
@@ -117,4 +117,18 @@
 
     expect(failures).to be_empty
   end
+
+  it "returns ipv4 list from cached file" do
+    vmh1, vmh2 = Array.new(2) { create_vm_host }
+    Address.create(cidr: "1.1.1.1/32", routed_to_host_id: vmh1.id) { it.id = vmh1.id }
+    Address.create(cidr: "3.3.3.3/27", routed_to_host_id: vmh2.id)
+    Address.create(cidr: "2.2.2.2/27", routed_to_host_id: vmh1.id)
+    Address.create(cidr: "3.3.3.3/28", routed_to_host_id: vmh2.id)
```
jeremyevans, line 126: Probably should have an IP address like `10.10.10.10/28` so you can test the sorting works as expected.

spec/prog/populate_ipv4_cache_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+require_relative "../model/spec_helper"
+
+RSpec.describe Prog::PopulateIpv4Cache do
+  subject(:pc) {
+    described_class.new(Strand.new(prog: "PopulateIpv4Cache"))
+  }
+
+  describe "#wait" do
+    it "populates file and naps an hour" do
```
jeremyevans, line 11: This approach doesn't work.  The prog runs in a single respirate process, and only that process will have the file written.  None of the web servers would have that file created (the file systems are not shared between dynos in production).

You should switch to storing the cached information in the database.

jeremyevans (commented):
As mentioned in the inline comments, using a file for this will not work, as the production processes do not use a shared file system.  You'll need to switch to storing the cache in the database.

PR #3477, opened 2025-06-19
Don't use cached children in `donate`.

pykello:
A strand's child might have changed label between two calls to the same `donate`, so using the cached value might miss the update and cause unexpected behaviour.


jeremyevans:
> I think run always implies a new instance though. See how it relates to unsychronized_run

I don't think it's a new instance.  The `@values` get replaced, but other instance state (such as associations) isn't affected: https://github.com/ubicloud/ubicloud/blob/b8c37ca5eb759f2531e6a63cad1f1430d0b66ebe/model/strand.rb#L92-L93 .

This should pick up the changed prog label, though.

fdr:
> > I think run always implies a new instance though. See how it relates to unsychronized_run
> 
> 
> 
> I don't think it's a new instance.  The `@values` get replaced, but other instance state (such as associations) isn't affected: https://github.com/ubicloud/ubicloud/blob/b8c37ca5eb759f2531e6a63cad1f1430d0b66ebe/model/strand.rb#L92-L93 .
> 
> 
> 
> This should pick up the changed prog label, though.  

That used to be .reload, would that have done the right thing?

fdr:
Like basically, this is also bad in the console, where using run on instances is extremely common

jeremyevans:
> > > I think run always implies a new instance though. See how it relates to unsychronized_run
> > 
> > 
> > I don't think it's a new instance.  The `@values` get replaced, but other instance state (such as associations) isn't affected: https://github.com/ubicloud/ubicloud/blob/b8c37ca5eb759f2531e6a63cad1f1430d0b66ebe/model/strand.rb#L92-L93
> >  .
> > This should pick up the changed prog label, though.
> 
> That used to be .reload, would that have done the right thing?

`reload` clears associations as well, and also clears changed columns, but you are still in the same instance.  prog label is not an association.  Maybe this is related to changed columns not being cleared?  Seems like that would result in columns being saved when they don't need to be, instead of not being saved when they need to be.

Assuming this doesn't fix the issue. I think we should run the E2E tests with a database logger attached in addition to the extra debugging, to see what queries are actually being sent to the database.  So far, this has passed 2 E2E tests, and is current running a third one.  I think the plan is to run 8 E2E tests, and if the problem doesn't occur on any of them, statistically is probably fixed.

fdr:
Yet, something can't be right. .run is meant to always be safe, including for console work. Fixing it at a call site is incomplete.

jeremyevans:
I'm not against adding association/changed columns clearing to `take_lease_and_reload`. We probably also want to clear `@subject` as well, right?

jeremyevans:
> I'm not against adding association/changed columns clearing to `take_lease_and_reload`. We probably also want to clear `@subject` as well, right?

https://github.com/ubicloud/ubicloud/pull/3479

jeremyevans:
@fdr Even with #3479, I still think we should merge this.  Are you OK with that?

fdr:
I think there's nothing wrong with this patch but the commit message. It seems like we don't have any convincing theory why it's correct, right? RE: label, which is not even an association. It's harmless enough to try it without a sound reason and see if it helps, but as-is it seems like we think the commit message can't possibly be true, is that right?

fdr:
If we're like, "we suspect some cached state is involved" and maybe better yet write up the symptom and add it on (that takes more time), I think "worth a try" is better than something that doesn't hold together.

PR #3483, opened 2025-06-20
Fix ipv4_address table population in Address#populate_ipv4_addresses

jeremyevans:
Change

id != routed_to_host_id

to

vm_host.sshable.host != cidr.network.to_s

We do not consistently use the VmHost id as the primary Address id (it is not done at all for Hetzner hosts), so the next simplest way to check is check the sshable address.

Existing installations that previously ran
Address#populate_ipv4_addresses can be fixed using:

DB[:ipv4_address].where(ip: Sshable.where(id: VmHost.select(:id)).select(Sequel.cast(:host, :inet))).delete

### DIFF
--- a/model/address.rb
+++ b/model/address.rb
@@ -29,9 +29,8 @@ class Address < Sequel::Model
 
   def populate_ipv4_addresses
     # Do nothing for ipv6 addresses, since VM addresses are chosen randomly from the /64.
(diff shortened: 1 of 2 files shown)


fdr (approved):
funny thing is sshable uses text because it could even be a dns name, that was the general idea

PR #3489, opened 2025-06-23
Use savepoints when running strands if already inside a transaction

jeremyevans:
Otherwise, use of donate in progs can cause a failure in a subsequent sibling prog to rollback successful completion of current prog.


jeremyevans:
I think I found a way to avoid needing savepoints for these `transaction` calls, so I'm going to close this.

PR #3492, opened 2025-06-23
Change Prog#donate to only nap

jeremyevans:
Running child progs in donate causes issues.  Let's say you have two child progs A and B:

* Prog A runs successfully
* Prog B does not run successfully

The failure of prog B rolls back the changes in prog A. If prog A is not idempotent, this is a permanent failure that requires external intervention.

The current approach is designed to run progs in serial, they are also executed concurrently in parallel by other respirate threads or processes.  For example, while prog A is running in the above example inside donate, prog B could be executing concurrently. By the time prog A finishes, it checks prog B, which has already finished.

The nap 1 approach may result in higher than desired load on the database.  A more efficient approach would be to nap longer in donate, and have exiting children update the schedule for their parent if all of the parent's children are ready to be reaped.

### DIFF
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -181,7 +181,6 @@ end
   end
 
   def donate
-    strand.children_dataset.each(&:run)
     nap 1
   end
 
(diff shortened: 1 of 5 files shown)


jeremyevans:
This prevents calling `run` manually on a parent strand to run child strands.  Since that is undesirable, closing this while I work on a different approach.

PR #3496, opened 2025-06-23
umount at `verify_extra_disks` if disk is already mounted.

pykello:
If `verify_extra_disks` fails after mounting the disk, a retry will always fail because the `sudo mkfs.ext4` command will error out with:

```
mke2fs 1.46.5 (30-Dec-2021)
/path/to/disk is mounted; will not make a filesystem here!
```

This change makes the label retriable by umounting if already mounted.


bsatzger (approved):

prog/test/vm.rb
```diff
@@ -37,10 +37,18 @@ class Prog::Test::Vm < Prog::Test::Base
...
+  def umount_if_mounted(mount_path)
+    sshable.cmd("sudo umount #{mount_path.shellescape}")
+  rescue Sshable::SshError => e
+    raise unless e.stderr.include?("not mounted")
+  end
+
   label def verify_extra_disks
     vm.vm_storage_volumes[1..].each_with_index { |volume, disk_index|
       mount_path = "/home/ubi/mnt#{disk_index}"
       sshable.cmd("mkdir -p #{mount_path}")
```
jeremyevans: We are using `mount_path.shellescape` in `umount_if_mounted`, but not here or later in this method.  It would probably be best to consistently use `mount_path.shellescape` in other commands.

PR #3505, opened 2025-06-25
Remove the max_ios_per_sec storage limit.

pykello:
SPDK supported max ios per sec for read/write combined, so we added support for it. However, cgroup v2 (which we’re migrating to for I/O rate limiting) does not support combined IOPS limits — only separate `IOReadIOPSMax` and `IOWriteIOPSMax`.

This patch removes the unused `max_ios_per_sec` knob. It was never used in production, and our current rate limiting relies solely on read/write bandwidth, which is fully supported by cgroup v2.

----

> [!IMPORTANT]
> Remove unused `max_ios_per_sec` feature and related logic across the codebase.
> 
>   - **Behavior**:
>     - Removes `max_ios_per_sec` from `IoLimits` in `lib/option.rb` and related logic in `lib/validation.rb`.
>     - Drops `max_ios_per_sec` column from `vm_storage_volume` table in `migrate/20250624_remove_ios_per_sec.rb`.
>     - Updates `set_qos_limits` in `rhizome/host/lib/storage_volume.rb` to exclude `rw_ios_per_sec`.
>   - **Tests**:
>     - Removes tests related to `max_ios_per_sec` in `spec/lib/validation_spec.rb`, `spec/model/vm_spec.rb`, and `spec/prog/test/vm_spec.rb`.
>     - Updates `spec/scheduling/allocator_spec.rb` to reflect removal of `max_ios_per_sec`.
>   - **Misc**:
>     - Removes `max_ios_per_sec` from `VmStorageVolume` and related logic in `model/vm.rb` and `model/vm_storage_volume.rb`.
>     - Updates `prog/test/vm.rb` and `prog/test/vm_group.rb` to remove `max_ios_per_sec` references.
>     - Adjusts `prog/vm/nexus.rb` and `scheduling/allocator.rb` to remove `max_ios_per_sec` handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0eef6db2871d261f175a8cc8e12fa1403ce89789. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


lib/option.rb
```diff
@@ -43,8 +43,8 @@ def display_name
     ["burstable", "Shared CPU", true, true]
   ].map { |args| VmFamily.new(*args) }
 
-  IoLimits = Struct.new(:max_ios_per_sec, :max_read_mbytes_per_sec, :max_write_mbytes_per_sec)
-  NO_IO_LIMITS = IoLimits.new(nil, nil, nil).freeze
+  IoLimits = Struct.new(:max_read_mbytes_per_sec, :max_write_mbytes_per_sec)
+  NO_IO_LIMITS = IoLimits.new(nil, nil).freeze
```
jeremyevans:
```suggestion
  NO_IO_LIMITS = IoLimits.new.freeze
```

Values are optional when initializing structs, and default to nil.  If you think IoLimits values should always be frozen, consider changing `Struct.new` to `Data.define`.  You can think of Data like a frozen Struct.  In that case, you'll need to keep the `nil` arguments, because Data subclasses require values for each member.

jeremyevans (approved):

PR #3506, opened 2025-06-25
Add FK Reference from Kubernetes Cluster to Services Load Balancer

aamederen:
**Add FK Reference from Kubernetes Cluster to Services Load Balancer**

This was a technical debt left from the launch of CCM for automatic
service to LB port mapping. We were using name and project queries
to reference the load balancer created for serving user services.

The fields will be backfilled manually upon deployment.

**Create Services LB as part of Kubernetes Cluster creation**

This change moves the services LB creation from kubernetes nodepool
to kubernetes cluster nexus, alongside the apiserver lb creation.
We still don't know how to support multiple nodepools and their
relationship to service LBs, but it's simpler to assume that we will
support a single service LB for the whole cluster.

This change also starts using the FK for the service LB reference,
instead of the name based query.

----

> [!IMPORTANT]
> Add FK from Kubernetes clusters to service load balancers, moving LB creation to cluster level and updating references.
> 
>   - **Behavior**:
>     - Add FK `services_lb_id` to `kubernetes_cluster` table in `20250623_add_services_lb_to_kubernetes_cluster.rb`.
>     - Move service LB creation to `kubernetes_cluster_nexus.rb` from `kubernetes_nodepool_nexus.rb`.
>     - Use FK `services_lb` in `Kubernetes::Client` instead of name-based query.
>   - **Models**:
>     - Add `many_to_one :services_lb` in `KubernetesCluster`.
>     - Update `LoadBalancer` references in `kubernetes_cluster.rb` and `load_balancer.rb`.
>   - **Tests**:
>     - Update tests in `client_spec.rb`, `kubernetes_cluster_nexus_spec.rb`, and `kubernetes_nodepool_nexus_spec.rb` to reflect FK changes.
>   - **Views**:
>     - Update `show.erb` to use `services_lb.hostname` for service URL display.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for af3f6cd9d12602bb576f1509269784016f144be9. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


migrate/20250623_add_services_lb_to_kubernetes_cluster.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:kubernetes_cluster) do
+      add_foreign_key :services_lb_id, :load_balancer, type: :uuid, null: true
```
enescakir, line 6: You should separate the migration and model changes into different commits

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -5,6 +5,7 @@
 class KubernetesCluster < Sequel::Model
   one_to_one :strand, key: :id
   many_to_one :api_server_lb, class: :LoadBalancer
+  many_to_one :services_lb, class: :LoadBalancer
```
enescakir, line 8: Why not just `:service_lb`?

aamederen: This load balancer serves many services the user creates, so we called it services lb all the time

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -134,6 +135,7 @@ def worker_vms
 #  target_node_size             | text                     | NOT NULL
 #  target_node_storage_size_gib | bigint                   |
 #  location_id                  | uuid                     | NOT NULL
+#  services_lb_id               | uuid                     |
```
jeremyevans, line 138: The schema annotation should be included in the migration commit.  Otherwise testing the migration commit by itself would fail.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -65,9 +70,29 @@ def before_run
       health_check_endpoint: "/healthz",
       health_check_protocol: "tcp",
       custom_hostname_dns_zone_id:,
-      custom_hostname_prefix:
+      custom_hostname_prefix: custom_apiserver_hostname_prefix
+    ).subject
+
+    kubernetes_cluster.update(api_server_lb_id: api_server_lb.id)
```
jeremyevans: Move `api_server_lb_id: api_server_lb.id` to the `update` call later in the method, so you are only issuing one UPDATE query and not 2.

model/kubernetes/kubernetes_cluster.rb
aamederen: For clarification, the migration, annotation and cache files should belong to the same commit, right? I initially did so, but then separated the migration file as a separate commit.

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: Good catch, thank you.

model/kubernetes/kubernetes_cluster.rb
jeremyevans: In general, this is correct.  In some cases, you may want to delay committing the cache files to a later commit, if the cache file change without the migration would break functionality.  In such cases, the migration commit will not pass CI in isolation, but it's more important that things work in production.

aamederen: Alright. I will move the annotations to the migrations commit and keep the cache as a separate commit.

PR #3519, opened 2025-06-26
Implement CSI

mohi-kalantari:
This PR introduces a full CSI (Container Storage Interface) implementation in Ruby. It adds the official CSI protobuf spec (v1.11.0) and generates Ruby gRPC definitions for use in the server.

The controller service is implemented with idempotent volume creation and deletion, proper validation, and thread safety. 

The node service manages local-disk-backed volumes, handling staging, publishing, and cleanup, and supports both block and filesystem volumes as requested by the PVC. All services include robust error handling and follow the CSI spec.

----

> [!IMPORTANT]
> This PR implements a full Container Storage Interface (CSI) in Ruby, including controller and node services, Kubernetes integration, and comprehensive testing.
> 
>   - **CSI Implementation**:
>     - Implements full CSI in Ruby with gRPC definitions from CSI protobuf spec v1.11.0.
>     - `ControllerService` and `NodeService` handle volume creation, deletion, staging, publishing, and cleanup.
>     - Supports block and filesystem volumes, idempotent operations, and robust error handling.
>   - **Kubernetes Integration**:
>     - `KubernetesClient` manages PV and PVC operations, node information, and scheduling.
>     - Manifests for CSI driver, node plugin, and provisioner setup in Kubernetes.
>   - **Testing and Configuration**:
>     - RSpec tests for `ControllerService`, `NodeService`, `IdentityService`, and `KubernetesClient`.
>     - SimpleCov for test coverage with 100% threshold.
>     - `ubi-csi.gemspec` defines dependencies and metadata for the CSI gem.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b3eee03b31af323db87c6c5f45b9c7c2162d5cba. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


kubernetes/csi/ubi-csi.gemspec
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+Gem::Specification.new do |spec|
+  spec.name = "ubi-csi"
+  spec.version = "0.1.0"
+  spec.authors = ["UbiCloud"]
```
enescakir: It looks too much like the LLM style of Ubicloud :D

kubernetes/csi/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+COPY Gemfile ubi-csi.gemspec ./
+RUN bundle config set without 'development test' && bundle install
+
+FROM ruby:3.2-slim
+
+RUN apt update && apt install -y apt-transport-https ca-certificates curl gnupg openssh-client
+RUN curl -fsSL https://pkgs.k8s.io/core:/stable:/v1.33/deb/Release.key | gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
+RUN echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.33/deb/ /' | tee /etc/apt/sources.list.d/kubernetes.list && \
+    chmod 644 /etc/apt/sources.list.d/kubernetes.list
+RUN  apt update && apt install -y kubectl
```
aamederen, line 16: nit: combining commands into a single RUN is generally advised. you already do some chaining but there may be more opportunities to reduce the layers.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
```
aamederen, line 2: should we have something like `#!/usr/bin/env ruby` at the top?

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,40 @@
...
+  puts "Registering services..."
+  server.handle(Csi::V1::IdentityService)
+  server.handle(Csi::V1::ControllerService)
+  server.handle(Csi::V1::NodeService)
+
+  puts "Starting UbiCSI server on #{endpoint}..."
+  begin
+    server.run_till_terminated
+  rescue => e
+    puts "Error starting server: #{e.message}"
```
aamederen: would `run_till_terminated` throw an exception only when starting the server?

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,166 @@
...
+        @volume_store = {} # Maps volume name to volume details
+        @mutex = Mutex.new
+      end
+
+      def log_with_id(id, message)
+        LOGGER.info("[req_id=#{id}] #{message}")
+      end
+
+      def run_cmd(*cmd, req_id: nil)
+        log_with_id(req_id, "Running command: #{cmd}") unless req_id.nil?
```
aamederen: should we escape the `cmd`?

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,166 @@
...
+
+        if selected.nil?
+          raise GRPC::FailedPrecondition.new("No suitable worker node topology found", GRPC::Core::StatusCodes::FAILED_PRECONDITION)
+        end
+
+        selected
+      end
+
+      def create_volume(req, _call)
+        req_id = SecureRandom.uuid
```
aamederen: do we use `req_id` for anything other than logging?

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,166 @@
...
+                type: ControllerServiceCapability::RPC::Type::CREATE_DELETE_VOLUME
+              )
+            )
+          ]
+        )
+        log_with_id(req_id, "controller_get_capabilities response: #{resp.inspect}")
+        resp
+      end
+
+      def select_worker_topology(req)
```
aamederen, line 43: some context in the form of comments around these functions could be helpful, like when, why and by whom these functions are called.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,166 @@
...
+
+        resp = nil
+        @mutex.synchronize do
+          if @volume_store[req.name]
+            existing = @volume_store[req.name]
+            if req.accessibility_requirements.requisite.first != existing[:accessible_topology]
+              raise GRPC::FailedPrecondition.new("Existing volume has incompatible topology", GRPC::Core::StatusCodes::FAILED_PRECONDITION)
+            end
+
+            if existing[:capacity_bytes] != req.capacity_range.required_bytes
```
aamederen, line 89: is this used for resizing an existing volume? If so, we can adjust the error message as "not supported"

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,166 @@
...
+        raise GRPC::InvalidArgument.new("Capacity range is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.capacity_range.nil?
+        raise GRPC::InvalidArgument.new("Required bytes must be specified", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.capacity_range.required_bytes.nil?
+        raise GRPC::InvalidArgument.new("Required bytes must be positive", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.capacity_range.required_bytes <= 0
+        raise GRPC::InvalidArgument.new("Volume size exceeds maximum allowed size of 2GB", GRPC::Core::StatusCodes::OUT_OF_RANGE) if req.capacity_range.required_bytes > MAX_VOLUME_SIZE
+        raise GRPC::InvalidArgument.new("Volume capabilities are required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.volume_capabilities.nil? || req.volume_capabilities.empty?
+        raise GRPC::InvalidArgument.new("Topology requirement is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.accessibility_requirements.nil? || req.accessibility_requirements.requisite.empty?
+
+        resp = nil
+        @mutex.synchronize do
+          if @volume_store[req.name]
```
aamederen: you can change this line to `if existing = @volume_store[req.name]`

kubernetes/csi/Dockerfile
mohi-kalantari: I thought about that a bit but the issue is we need the first set of packages to add the kubernetes apt repo and then run apt install kubectl so I couldn't do much

Do you mean combining the 3 RUNs for installing the kubectl?

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: no just logging. there is no unique parameter which is present in all requests and can identify the request. that's why I used the uuid.

kubernetes/csi/bin/ubi-csi-server
mohi-kalantari: yes based on the docs, it seems that it might throw exceptions only on starts

kubernetes/csi/bin/ubi-csi-server
mohi-kalantari: since we directly call it with /usr/local/bin/ruby, it's not necessary.

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: no it's not used for resize and there is another method called expand volume for that.

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: Will fix it

kubernetes/csi/bin/ubi-csi-server
jeremyevans: Even if not strictly necessary, I think it's a good practice to add this.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,40 @@
...
+    raise
+  end
+end
+
+main if __FILE__ == $PROGRAM_NAME
```
jeremyevans: No need for a `main` function (this actually adds `main` as a private instance method on all objects). Just have the function code be directly in the program body.

You only need the `__FILE__ == $PROGRAM_NAME` if you will be requiring the file from another file.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,40 @@
...
+$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+$LOAD_PATH.unshift(File.expand_path("..", __dir__))
+
+require "bundler/setup"
+require "ubi_csi"
+
+def main
+  $stdout.sync = true
+  puts "Starting main function..."
+  endpoint = ENV["CSI_ENDPOINT"].strip
```
jeremyevans: If the environment variable is not set, you get a NoMethodError here.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,40 @@
...
+
+$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+$LOAD_PATH.unshift(File.expand_path("..", __dir__))
+
+require "bundler/setup"
+require "ubi_csi"
+
+def main
+  $stdout.sync = true
+  puts "Starting main function..."
```
jeremyevans: Debugging code should be removed, both here and elsewhere in this method and anywhere else in this PR.   If you want to support debugging output, use `puts "str" if $VERBOSE`. This way it is not verbose by default, but can be verbose if requested.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,40 @@
...
+  endpoint = ENV["CSI_ENDPOINT"].strip
+  raise "CSI_ENDPOINT environment variable is not set" if endpoint.nil? || endpoint.empty?
```
jeremyevans: The `endpoint.nil?` is unnecessary, since if we get to this point, there is no way that `endpoint` can be `nil`, since `nil.strip` is not defined.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+$LOAD_PATH.unshift(File.expand_path("..", __dir__))
```
jeremyevans: The `../lib` I can understand.  Why is the `..` needed?  What are you requiring relative to `..`?

kubernetes/csi/bin/ubi-csi-server
jeremyevans: This rescues all exceptions, so any exception raised by the method will result in this error method. Granted, most of those errors will be during startup, but looking at the RpcServer code, it seems unlikely that is all possible errors.

Also, errors should be printed to stderr and not stdout.

kubernetes/csi/lib/ubi_csi/controller_service.rb
jeremyevans: Prefer `if req_id` to `unless req_id.nil?` (and `unless req_id` to `if req_id.nil?`), unless you specifically want to treat `false` differently from `nil`.  Both here and elsewhere in this PR.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+        log_with_id(req_id, "Running command: #{cmd}") unless req_id.nil?
+        Open3.capture2e(*cmd)
+      end
+
+      def controller_get_capabilities(req, _call)
+        req_id = SecureRandom.uuid
+        log_with_id(req_id, "controller_get_capabilities request: #{req.inspect}")
+        if req.nil?
+          raise GRPC::InvalidArgument.new("Request cannot be nil", GRPC::Core::StatusCodes::INVALID_ARGUMENT)
+        end
```
jeremyevans: Unless there is a good reason to perform logging before checking this, I would move this to the top of the method. Similar issue in other methods.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+          selected_topology = select_worker_topology(req)
+          volume_id = "vol-#{SecureRandom.uuid}"
+          @volume_store[req.name] = {
+            volume_id: volume_id,
+            name: req.name,
+            accessible_topology: selected_topology,
+            capacity_bytes: req.capacity_range.required_bytes,
+            parameters: req.parameters.to_h,
+            capabilities: req.volume_capabilities.map(&:to_h)
+          }
```
jeremyevans: As this is potentially accessed by multiple threads. I recommend freezing this unless you plan on modifying it.  Also consider freezing the values in it as well, unless those will be modified.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+        raise GRPC::InvalidArgument.new("Volume name is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.name.nil? || req.name.empty?
+        raise GRPC::InvalidArgument.new("Capacity range is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.capacity_range.nil?
+        raise GRPC::InvalidArgument.new("Required bytes must be positive", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.capacity_range.required_bytes <= 0
+        raise GRPC::InvalidArgument.new("Volume size exceeds maximum allowed size of 2GB", GRPC::Core::StatusCodes::OUT_OF_RANGE) if req.capacity_range.required_bytes > MAX_VOLUME_SIZE
+        raise GRPC::InvalidArgument.new("Volume capabilities are required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.volume_capabilities.nil? || req.volume_capabilities.empty?
+        raise GRPC::InvalidArgument.new("Topology requirement is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.accessibility_requirements.nil? || req.accessibility_requirements.requisite.empty?
+
+        resp = nil
+        @mutex.synchronize do
```
jeremyevans: The scope of this mutex synchronization is quite large, and it doesn't look like values in the hash are modified. I recommend reducing the scope:

```ruby
existing = new = nil
mutex.synchronize do
  unless existing = @volume_store[req.name]
     # ...
     new = @volume_store[req.name] = ...
  end
end

if existing
  # various checks
else # new
end
```

Only hold the mutex for as long as necessary. If you don't strictly need the mutex to ensure correct behavior, then release it.

This section also has significant redundancy. By restructuring using the above code, you can probably have a a single CreateVolumeResponse call.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+
+            if existing[:capacity_bytes] != req.capacity_range.required_bytes
+              raise GRPC::FailedPrecondition.new("Volume with same name but different size exists", GRPC::Core::StatusCodes::FAILED_PRECONDITION)
+            end
+
+            if existing[:parameters] != req.parameters.to_h
+              raise GRPC::FailedPrecondition.new("Volume with same name but different parameters exists", GRPC::Core::StatusCodes::FAILED_PRECONDITION)
+            end
+
+            existing_capabilities = existing[:capabilities].sort_by { |cap| cap.to_json }
```
jeremyevans: `sort_by(&:to_json)` . Both here, the following line, and other places where you are calling a method without arguments on the block argument.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+      end
+
+      def delete_volume(req, _call)
+        req_id = SecureRandom.uuid
+        log_with_id(req_id, "delete_volume request: #{req.inspect}")
+        raise GRPC::InvalidArgument.new("Request cannot be nil", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.nil?
+        raise GRPC::InvalidArgument.new("Volume ID is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.volume_id.nil? || req.volume_id.empty?
+
+        client = KubernetesClient.new(req_id:)
+        pv_name = @mutex.synchronize { @volume_store.find { |_, d| d[:volume_id] == req.volume_id }&.first }
```
jeremyevans: Unless you are sure `@volume_store` will always be a small hash, create a separate hash keyed by the volume id for the lookup.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+require "json"
+
+module Csi
+  module V1
+    class ControllerService < Controller::Service
+      MAX_VOLUME_SIZE = 2 * 1024 * 1024 * 1024 # 2GB in bytes
+      LOGGER = Logger.new($stdout)
+
+      def initialize
+        @volume_store = {} # Maps volume name to volume details
```
jeremyevans, line 17: This is an in-memory store.  Is it expected that if the machine running this crashes, all information on the stored volumes is lost?

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+
+      def delete_volume(req, _call)
+        req_id = SecureRandom.uuid
+        log_with_id(req_id, "delete_volume request: #{req.inspect}")
+        raise GRPC::InvalidArgument.new("Request cannot be nil", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.nil?
+        raise GRPC::InvalidArgument.new("Volume ID is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.volume_id.nil? || req.volume_id.empty?
+
+        client = KubernetesClient.new(req_id:)
+        pv_name = @mutex.synchronize { @volume_store.find { |_, d| d[:volume_id] == req.volume_id }&.first }
+        pv = pv_name.nil? ? client.find_pv_by_volume_id(req.volume_id) : client.get_pv(pv_name)
```
jeremyevans: If we can't find the `pv`, should we raise a `GRPC::InvalidArgument` here?

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+          raise GRPC::Internal, "Could not delete the PV's backing file"
+        end
+        @mutex.synchronize { @volume_store.delete(pv_name) }
+
+        resp = DeleteVolumeResponse.new
+        log_with_id(req_id, "delete_volume response: #{resp.inspect}")
+        resp
+      rescue GRPC::InvalidArgument => e
+        log_with_id(req_id, "Handled gRPC validation error in delete_volume: #{e.class} - #{e.message}}")
+        raise e
```
jeremyevans: `raise` without an argument reraises the current exception. Generally, you should only provide an argument if it would be different from that.

kubernetes/csi/lib/ubi_csi/identity_service.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require "grpc"
+require "logger"
+require_relative "../csi_services_pb"
+
+module Csi
+  module V1
+    class IdentityService < Identity::Service
+      LOGGER = Logger.new($stdout)
```
jeremyevans: This is a separate logger from the one used by other services. It would probably be better to have the logger passed in, so you could create a single logger in `ubi-csi-server`, and pass it to each created service. Alternatively, define this constant in a module, and include the module in each of th eservices.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -0,0 +1,172 @@
...
+      def log_with_id(id, message)
+        LOGGER.info("[req_id=#{id}] #{message}")
+      end
+
+      def run_cmd(*cmd, req_id: nil)
+        log_with_id(req_id, "Running command: #{cmd}") unless req_id.nil?
+        Open3.capture2e(*cmd)
+      end
+
+      def controller_get_capabilities(req, _call)
```
jeremyevans, line 21: Looking at these methods, I see a general pattern:

```ruby
req_id = SecureRandom.uuid
log_with_id(req_id, "... request: #{req.inspect}")
raise GRPC::InvalidArgument.new("Request cannot be nil", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.nil?
resp = ...
log_with_id(req_id, "... response: #{resp.inspect}")
resp 
```

You should probably have a method that handles this:

```ruby
def log_request_response(req, type)
  raise GRPC::InvalidArgument.new("Request cannot be nil", GRPC::Core::StatusCodes::INVALID_ARGUMENT) unless req
  req_id = SecureRandom.uuid
  log_with_id(req_id, "#{type} request: #{req.inspect}")
  resp = yield(req_id) # caller already has access to req
  log_with_id(req_id, "#{type} response: #{resp.inspect}")
  resp 
end
```

And use that method to DRY up the other code.

kubernetes/csi/lib/ubi_csi/identity_service.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def log_with_id(id, message)
```
jeremyevans: This is the same code as in the other service.  You should move it to a module that is included in both services.

kubernetes/csi/lib/ubi_csi/identity_service.rb
```diff
@@ -0,0 +1,59 @@
...
+        LOGGER.info("[req_id=#{id}] #{message}")
+      end
+
+      def get_plugin_info(request, _call)
+        req_id = SecureRandom.uuid
+        log_with_id(req_id, "get_plugin_info request: #{request.inspect}")
+        resp = GetPluginInfoResponse.new(
+          name: "csi.ubicloud.com",
+          vendor_version: "0.1.0"
```
jeremyevans: Should use `UbiCSI::VERSION` here.

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -0,0 +1,84 @@
...
+
+require "open3"
+require "yaml"
+require_relative "errors"
+
+module Csi
+  class KubernetesClient
+    def initialize(req_id: nil)
+      @req_id = req_id
+      @logger = Logger.new($stdout) if req_id.nil?
```
jeremyevans: Callers of this method always pass non-nil `req_id`, so this appears to be unneeded.  This would also allow you to remove lines calling `@logger&.info`.

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    def run_cmd(*cmd, **options)
```
jeremyevans: Another candidate to move to a shared module.

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -0,0 +1,84 @@
...
+        if output.strip.end_with?("not found")
+          raise ObjectNotFoundError, output
+        end
+        raise "Command failed: #{cmd.join(" ")}\nOutput: #{output}"
+      end
+      output
+    end
+
+    def get_node(name)
+      YAML.safe_load(run_kubectl("get", "node", name, "-oyaml"))
```
jeremyevans: Probably worthwhile to add a private method that for this, and use it to DRY things up:

```ruby
def yaml_load_kubectl(*)
  YAML.safe_load(run_kubectl(*))
end
```

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -0,0 +1,84 @@
...
+    def initialize(req_id: nil)
+      @req_id = req_id
+      @logger = Logger.new($stdout) if req_id.nil?
+    end
+
+    def run_cmd(*cmd, **options)
+      Open3.capture2e(*cmd, **options)
+    end
+
+    def run_kubectl(*args, stdin_data: nil)
```
jeremyevans: I would change this from `stdin_data` to `yaml_data`. and do the `YAML.dump` in this method, to DRY things up.

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -0,0 +1,84 @@
...
+    def update_pvc(yaml_data)
+      run_kubectl("apply", "-f", "-", stdin_data: YAML.dump(yaml_data))
+    end
+
+    def delete_pvc(namespace, name)
+      run_kubectl("-n", namespace, "delete", "pvc", name, "--wait=false", "--ignore-not-found=true")
+    end
+
+    def node_schedulable?(name)
+      !get_node(name)["spec"]&.fetch("unschedulable", false)
```
jeremyevans: `!get_node(name).dig("spec", "unschedulable")` .  The only reason to use `fetch` with `false` is to treat `false` differently from `nil`, but then you are using `!`, which is going to convert both to `true`.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+      VOLUME_BASE_PATH = "/var/lib/ubicsi"
+      LOGGER = Logger.new($stdout)
+      OLD_PV_NAME_ANNOTATION_KEY = "csi.ubicloud.com/old-pv-name"
+      OLD_PVC_OBJECT_ANNOTATION_KEY = "csi.ubicloud.com/old-pvc-object"
+
+      def log_with_id(id, message)
+        LOGGER.info("[req_id=#{id}] [CSI NodeService] #{message}")
+      end
+
+      def node_name
```
jeremyevans: Use a constant for this, since I assume we will not be changing the environment variable while the program is running.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+        req_id = SecureRandom.uuid
+        log_with_id(req_id, "node_stage_volume request: #{req.inspect}")
+        client = KubernetesClient.new(req_id:)
+
+        pvc = fetch_and_migrate_pvc(req_id, client, req)
+        resp = perform_node_stage_volume(req_id, pvc, req, _call)
+        roll_back_reclaim_policy(req_id, client, req, pvc)
+        remove_old_pv_annotation(client, pvc)
+
+        resp
```
jeremyevans: No logging of response here. Is the difference here intentional?

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+        output, _ = run_cmd(*cmd, req_id:)
+        output
+      end
+
+      def is_mounted?(path, req_id: nil)
+        _, status = run_cmd("mountpoint", "-q", path, req_id:)
+        status == 0
+      end
+
+      def find_loop_device(backing_file)
```
jeremyevans: Probably should accept a `req_id` and pass it to `run_cmd`.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+        log_with_id(req_id, "Running command: #{cmd}") unless req_id.nil?
+        Open3.capture2e(*cmd)
+      end
+
+      def run_cmd_output(*cmd, req_id: nil)
+        output, _ = run_cmd(*cmd, req_id:)
+        output
+      end
+
+      def is_mounted?(path, req_id: nil)
```
jeremyevans: Make `req_id` a required keyword argument, and have all callers pass it, unless there is a caller that doesn't have a `req_id`.  Same for other methods where `req_id` is an optional keyword argument.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+        _, status = run_cmd("mountpoint", "-q", path, req_id:)
+        status == 0
+      end
+
+      def find_loop_device(backing_file)
+        output, ok = run_cmd("losetup", "-j", backing_file)
+        if ok && !output.empty?
+          loop_device = output.split(":")[0].strip
```
jeremyevans: `split(":", 2)`, since you only care about a single split.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+        raise GRPC::Internal, e.message
+      rescue => e
+        log_with_id(req_id, "Internal error in node_stage_volume: #{e.class} - #{e.message} - #{e.backtrace}")
+        raise GRPC::Internal, "Unexpected error: #{e.class} - #{e.message}"
+      end
+
+      def perform_node_stage_volume(req_id, pvc, req, _call)
+        volume_id = req.volume_id
+        staging_path = req.staging_target_path
+        size_bytes = req.volume_context["size_bytes"].to_i
```
jeremyevans: If `req.volume_context["size_bytes"]` is a string, prevent treating unexpected input as `0`, use `Integer(req.volume_context["size_bytes"], 10)`

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+        backing_file = NodeService.backing_file_path(volume_id)
+        unless Dir.exist?(VOLUME_BASE_PATH)
```
jeremyevans: Unless there is a reason to be lazy about this, I would move this code to when the file is loaded.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+        if pv.dig("spec", "persistentVolumeReclaimPolicy") == "Retain"
+          pv["spec"]["persistentVolumeReclaimPolicy"] = "Delete"
+          client.update_pv(pv)
+        end
+      rescue => e
+        log_with_id(req_id, "Internal error in node_stage_volume: #{e.class} - #{e.message} - #{e.backtrace}")
+        raise GRPC::Internal, "Unexpected error: #{e.class} - #{e.message}"
+      end
+
+      def is_copied_pvc(pvc)
```
jeremyevans: I would replace this method with: `alias_method :is_copied_pvc?, :pvc_needs_migration?` . Note the `?` on the method name to indicate it is a predicate method. You'd want to update callers to add the `?`.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+            prepare_data_migration(client, req_id, req.volume_id)
+          end
+        rescue => e
+          log_with_id(req_id, "Internal error in node_unstage_volume: #{e.class} - #{e.message} - #{e.backtrace}")
+          raise GRPC::Internal, "Unexpected error: #{e.class} - #{e.message}"
+        end
+        perform_node_unstage_volume(req_id, req, _call)
+      end
+
+      def perform_node_unstage_volume(req_id, req, _call)
```
jeremyevans: This is only called in a single place. I think this should be merged into `node_unstage_volume`, unless there is another reason to keep the methods separate (external callers?).

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+
+      def prepare_data_migration(client, req_id, volume_id)
+        log_with_id(req_id, "Retaining pv with volume_id #{volume_id}")
+        pv = retain_pv(req_id, client, volume_id)
+        log_with_id(req_id, "Recreating pvc with volume_id #{volume_id}")
+        recreate_pvc(req_id, client, pv)
+      end
+
+      def retain_pv(req_id, client, volume_id)
+        pv = client.find_pv_by_volume_id(volume_id) # todo: if pv is nil, react accordingly
```
jeremyevans: Should probably implement this todo.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,402 @@
...
+          pv["spec"]["persistentVolumeReclaimPolicy"] = "Retain"
+          client.update_pv(pv)
+          log_with_id(req_id, "Updated PV to retain")
+        end
+        pv
+      end
+
+      def recreate_pvc(req_id, client, pv)
+        pvc_namespace = pv["spec"]["claimRef"]["namespace"]
+        pvc_name = pv["spec"]["claimRef"]["name"]
```
jeremyevans:
```suggestion
        pvc_namespace, pvc_name = pv["spec"]["claimRef"].values_at("namespace", "name")
```

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: Yes that is expected. main goal is to be idempotent in consecutive calls otherwise it's not important in the long run.

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: what if I plan to just log the exception? is there a better way to do that?

kubernetes/csi/lib/ubi_csi/node_service.rb
mohi-kalantari: logging was happening in the perform_node_stage function, now it's fixed

kubernetes/csi/lib/ubi_csi/controller_service.rb
jeremyevans: It looks like you are already logging the exception in this case. How you are logging currently seems fine. If you are only logging and not calling `raise`, then you aren't passing arguments to `raise`, so I'm not sure how it is related to my comment.

kubernetes/csi/bin/ubi-csi-server
jeremyevans: This was not resolved. Please add the shebang line.

kubernetes/csi/bin/ubi-csi-server
jeremyevans: This is not resolved. You only removed the `if __FILE__ == $PROGRAM_NAME`, you kept the `main` function.  Please remove the `main` function and put the function code directly in the program body.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,43 @@
...
+  $stdout.sync = true
+  logger = Logger.new($stdout)
+  logger.level = $VERBOSE ? Logger::DEBUG : Logger::INFO
+
+  logger.debug("Starting main function")
+
+  endpoint = ENV["CSI_ENDPOINT"]
+  raise "CSI_ENDPOINT environment variable is not set" if endpoint.nil?
+  endpoint = endpoint.strip
+  raise "CSI_ENDPOINT environment variable is empty" if endpoint.empty?
```
jeremyevans:
```suggestion
```

You don't need this check, as an empty environment variable will be caught by the following line.

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,43 @@
...
+def main
+  $stdout.sync = true
+  logger = Logger.new($stdout)
+  logger.level = $VERBOSE ? Logger::DEBUG : Logger::INFO
+
+  logger.debug("Starting main function")
+
+  endpoint = ENV["CSI_ENDPOINT"]
+  raise "CSI_ENDPOINT environment variable is not set" if endpoint.nil?
+  endpoint = endpoint.strip
```
jeremyevans:
```suggestion
  endpoint.strip!
```

kubernetes/csi/bin/ubi-csi-server
```diff
@@ -0,0 +1,43 @@
...
+  logger.debug("Registering services")
+  server.handle(Csi::V1::IdentityService.new(logger:))
+  server.handle(Csi::V1::ControllerService.new(logger:))
+  server.handle(Csi::V1::NodeService.new(logger:))
+
+  logger.debug("Starting UbiCSI server on #{endpoint}")
+  begin
+    server.run_till_terminated
+  rescue => e
+    logger.error("Error starting server: #{e.message}\n#{e.backtrace.join("\n")}")
```
jeremyevans: This still prints the error to stdout and not stderr. We may want to avoid using the `logger` for this, and print directly to stderr.

kubernetes/csi/lib/ubi_csi/service_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+  module ServiceHelper
+    def log_with_id(req_id, message)
+      @logger.debug("[req_id=#{req_id}] #{message}")
+    end
+
+    def log_request_response(req, type)
+      raise GRPC::InvalidArgument.new("Request cannot be nil", GRPC::Core::StatusCodes::INVALID_ARGUMENT) unless req
+      req_id = SecureRandom.uuid
+      log_with_id(req_id, "#{type} request: #{req.inspect}")
+      resp = yield(req, req_id)
```
jeremyevans: Do we need to yield `req` here?  The calling scope already has access to it, since it is passed to the method. Please change this to just yield `req_id`.

kubernetes/csi/lib/ubi_csi/controller_service.rb
jeremyevans: This line wasn't modified. If you are sure that `@volume_store` will always be a small hash, add a comment explaining that.

jeremyevans (commented):
Most requested changes have been made correctly. There are a few that still need additional work.

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: Yes I didn't resolve this one because I wasn't done with it. we would have at most 8 volumes per node so it would be a very small hash. added the comment

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: fixed it thanks. I didn't understand correctly first

kubernetes/csi/bin/ubi-csi-server
mohi-kalantari: fixed

kubernetes/csi/lib/ubi_csi/node_service.rb
mohi-kalantari: I changed the implementation of find_pv_by_volume_id and will throw error when object is not found

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: I'm throwing an ObjectNotFound error when find_pv_by_volume_id does not find the pv

mohi-kalantari: basically it's an internal error but internal from the kubernetes side. will see if InternalError is the right error code to return

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -0,0 +1,359 @@
...
+          client = KubernetesClient.new(req_id:, logger: @logger)
+          pvc = fetch_and_migrate_pvc(req_id, client, req)
+          perform_node_stage_volume(req_id, pvc, req, _call)
+          roll_back_reclaim_policy(req_id, client, req, pvc)
+          remove_old_pv_annotation(client, pvc)
+          NodeStageVolumeResponse.new
+        end
+      end
+
+      def fetch_and_migrate_pvc(req_id, client, req)
```
aamederen, line 103: I believe in-code documentation would be super useful, especially for this file. But it's not a blocker for the PR and can be followed up later.

aamederen (approved):

fdr:
There's still more stuff that can be done, but I think this will do for now. Although I do not subscribe to TDD, this implementation would have benefited from having tests at an earlier point. After working with the node service tests enough to get a sense of how they work, I took these notes:

commit 2baac19a8206b5cc539a4929e434788e6d1a7e91
Author: Daniel Farina <daniel@ubicloud.com>
Date:   Thu Jul 24 14:50:00 2025 -0700

    Experiment with using files for testing the CSI
    
    I don't think use of files is that helpful -- the CSI does very few
    file manipulations directly since it is designed around some of the
    more advanced options of `fallocate` called through the shell -- but
    doing this exposes something interesting about some refactoring that
    would be useful to improve the code's test robustness.  It also makes
    the tests noticeably slower.
    
    I'm not going to do those changes now, because we have some other
    project priorities and at least as-is the tests fumble through all the
    lines and branches of this implementation by brute force, which is
    good enough for now.  But I wanted to describe it for later reference
    and education:
    
    The "unit" for `perform_node_stage_volume` is too monolithic. It
    starts innocently at first:
    
          it "handles fallocate failure" do
            prefix_helper.call do |service, req, backing_file|
              expect(service).to receive(:run_cmd).with("fallocate", "-l", size_bytes.to_s, backing_file, req_id:).and_return(["Error message", false])
    
              expect { service.perform_node_stage_volume(req_id, pvc, req, nil) }.to raise_error(GRPC::ResourceExhausted, /Failed to allocate backing file/)
    
    This is because `fallocate` is the first step in a multi-step process.
    
    However, once we want to test the second step, we need to keep some
    kind of `fallocate` mock to then inject faults in the second step:
    
            prefix_helper.call do |service, req, backing_file|
              expect(service).to receive(:run_cmd).with("fallocate", any_args, req_id:).and_return(["", true]).at_least(:once)
              expect(service).to receive(:run_cmd).with("losetup", "-j", any_args, req_id:).and_return(["a fake found loop device", true])
              expect(service).to receive(:run_cmd).with("losetup", "--find", "--show", backing_file, req_id: req_id).and_return(["Error setting up loop device", false])
    
    This keeps building as we want to proceed deeper into the code.
    
    The general idea to fix this would be to refactor the implementation
    so that each error-prone step can have its faults injected in detail,
    but they fit together into a somewhat simple data flow in the end.
    That data flow has to be simple, for two reasons:
    
    * If it's complex, the interaction of mocks and what the resulting
      mocked code can prove about the implementation is hard to read.
    
    * If it's complex, the tests are more voluminous and greater in
      quantity to handle various branches, and are longer.
    
    Sometimes, completely satisfying solutions are not available for every
    piece of code, but I think we could succeed here, if we had time to.

fdr:
Note I made one important interface change:

```
commit ff984c723af4c7aaf36c69b4072d0c30a4b0c3a2
Author: Daniel Farina <daniel@ubicloud.com>
Date:   Thu Jul 24 15:35:47 2025 -0700

    Outline mkdir_p for CSI directory initialization
    
    It's unusual to have such a side effect upon object initialization.
    
    In production cases, this object is not created more than once per
    startup, so, no big deal...but once testing and mocks get involved, it
    clutters the tests, so abide the more conventional approach of
    outlining the side effect.
    
    In passing, remove the check for existence of a directory before
    `mkdir_p`: it idempotent, it will check if there's nothing to do
    internally anyway.
```

fdr (approved):

PR #3523, opened 2025-06-26
Give free premium runner upgrade for new installations for 7 days

enescakir:
- **Fix free premium runner upgrade logic**
  If a free premium upgrade is enabled for an installation, we bill based
  on the original VM size, not the upgraded one.
  
  Previously, runners were upgraded during provisioning when the free
  upgrade was active. However, if the upgrade expired before the runner
  was destroyed, we ended up charging based on the upgraded VM size, since
  the upgrade was no longer active at billing time.
  
  To fix this, we now determine the billed_vm_size at provisioning time
  and store it. This ensures consistent billing, regardless of the upgrade
  status at the time of destruction.
  
  I didn’t remove the old billing logic yet, because runners already
  provisioned before this change won’t have a billed_vm_size set. If the
  column is empty, we still fall back to the old logic. I’ll remove the
  fallback after confirming all active runners have the new column
  populated.
  

- **Give free premium runner upgrade for new installations for 7 days**
  If a GitHub installation was created within the past 7 days,
  automatically grant a free premium runner upgrade to allow users to try
  out enhanced performance.
  

- **Show badge on UI for free premium upgrades**


jeremyevans (approved):
Combined with #3520, I believe this results in allowing free premium runner upgrades  for all existing github installations for a week after the migration is run, as the migration will set all installations as being created that day.  If we are OK with that, I recommend we deliberately announce it as a benefit to existing customers, to thank them for their support.

With this approach, if anyone wants free premium runners upgrades, they can delete the github installation and re-add it every week. If we care about that, we should probably decide how we want to detect the issue and handle it.

enescakir:
> Combined with #3520, I believe this results in allowing free premium runner upgrades for all existing github installations for a week after the migration is run, as the migration will set all installations as being created that day. If we are OK with that, I recommend we deliberately announce it as a benefit to existing customers, to thank them for their support.

I’ll populate the created_at column for installations using the GitHub API after adding created_at column. https://github.com/ubicloud/ubicloud/pull/3520/commits/883a4c835971a03d9ed10948c8a7b0b50d4f6300

> With this approach, if anyone wants free premium runners upgrades, they can delete the github installation and re-add it every week. If we care about that, we should probably decide how we want to detect the issue and handle it.

Yes, that's true. It's also true for the 1250 free credits. It's per project, so they can delete the project and recreate it as many times as they want. I keep an eye on deleted installations for churned customers, hoping to spot a pattern after a few tries.

jeremyevans (approved):

enescakir:
@jeremyevans I removed the first fix commit because it wasn't working properly. The new location also had incorrect data.

I added a new commit at the end. Instead of moving the decision tree, we check the free upgrade expiration time against the runner creation time, not the current time.

enescakir:
You're faster than I am! :D Thanks for the new approval!

PR #3527, opened 2025-06-26
Have last exiting child schedule parent strand

jeremyevans:
Previously, when reaping child processes, if there were no remaining
reapable children, the parent strand would only nap 1, which puts
unnecessary load on respirate unless at least one child strand
exits in the next second.

Change this approach by having the exiting child strands, after they
release the lease, schedule their parent immediately if the parent
has no non-exited child strands.

When doing this, you need to be careful to make sure there are not
race conditions that would delay the scheduling of the parent.
There are two potential situations you need to handle:

1. Multiple children exiting at the same time
2. Parent currently running while child is exiting

By waiting until after the child strand leases are released, you
still have a race condition with 1, but the race condition is that 
multiple child strands exiting concurrently could both reschedule
the parent strand.  However, that isn't a problem. You want to avoid
the case where neither child strand schedules the parent, which
rescheduling after releasing the lease should do.

To handle 2, inside reap use Model#lock! to lock the parent strand.
This will make exiting child strands block if they UPDATE the parent
strand with a new schedule, until the parent strand's transaction
commits.  However, it's possible that a child strand already UPDATED
the parent. To handle this situation, before calling lock!, store
the cached schedule value in a local variable.  lock! implicitly does
a reload, so compare the schedule value after reload.  If the schedule
has changed, likely a child scheduled the parent for immediate
execution, so nap 0 in that case.

Just in case there are unforeseen race conditions that are not handled,
only nap for 120 seconds if there are active children.  Worst case
scenario, this results in a 2 minute delay for running the parent.
However, this can potentially result in 120x less load from parent
strands polling children.

----

> [!IMPORTANT]
> Optimize parent strand scheduling by having the last exiting child schedule the parent immediately, reducing unnecessary load and potential delays.
> 
>   - **Behavior**:
>     - Exiting child strands now schedule their parent immediately if no active siblings exist, reducing potential 120s delay (`strand.rb`).
>     - Parent strand locks during reap to prevent race conditions (`base.rb`).
>     - Default nap time increased to 120s if active children exist (`base.rb`).
>   - **Tests**:
>     - Updated nap expectations from 1s to 120s in various test files (`inference_endpoint_replica_nexus_spec.rb`, `inference_router_replica_nexus_spec.rb`, `install_dnsmasq_spec.rb`, `kubernetes_cluster_nexus_spec.rb`, `kubernetes_nodepool_nexus_spec.rb`, `provision_kubernetes_node_spec.rb`, `upgrade_kubernetes_node_spec.rb`, `minio_server_nexus_spec.rb`, `postgres_server_nexus_spec.rb`, `setup_grafana_spec.rb`, `vm_group_spec.rb`, `victoria_metrics_server_nexus_spec.rb`, `host_nexus_spec.rb`).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 79843f8dccbb8c3be720517bc9146e355f8413c0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -235,7 +235,23 @@ end
         active_children.each do |child|
           nap 0 if child.run
         end
-        nap 1
+
+        schedule = strand.schedule
+
+        # Lock this parent strand. This is run inside a transaction,
+        # and will make exited child strands attempting to update the
+        # parent's schedule block until the transaction commits.
+        strand.lock!
+
+        # lock! does an implicit reload, so check the new schedule
+        new_schedule = strand.schedule
+
+        # In case the exiting child updated the parent schedule before
+        # the lock, check whether the schedule changed. If the schedule
+        # changed, assume it was set to CURRENT_TIMESTAMP, and nap 0.
+        # Otherwise, nap for 120s and rely on the exiting child strand
+        # scheduling this parent sooner in most cases.
+        nap((schedule != new_schedule) ? 0 : 120)
       end
     end
   end
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -208,7 +208,7 @@ SQL
       update(exitval: ext.exitval, retval: nil)
       if parent_id.nil?
         # No parent Strand to reap here, so self-reap.
-        Semaphore.where(strand_id: id).destroy
+        semaphores_dataset.destroy
         destroy
         @deleted = true
       end
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -105,9 +105,32 @@ SQL
           fail "BUG: strand with @deleted set still exists in the database"
         end
       else
-        unless RELEASE_LEASE_PS.call(id:, lease_time:) == 1
-          Clog.emit("lease violated data") { {lease_clear_debug_snapshot: this.all} }
-          fail "BUG: lease violated"
+        begin
+          unless RELEASE_LEASE_PS.call(id:, lease_time:) == 1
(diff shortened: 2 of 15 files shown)


prog/base.rb
```diff
@@ -235,7 +235,23 @@ def reap(hop = nil, reaper: nil, nap: nil, fallthrough: false)
...
+
+        # lock! does an implicit reload, so check the new schedule
+        new_schedule = strand.schedule
+
+        # In case the exiting child updated the parent schedule before
+        # the lock, check whether the schedule changed. If the schedule
+        # changed, assume it was set to CURRENT_TIMESTAMP, and nap 0.
+        # Otherwise, nap for 120s and rely on the exiting child strand
+        # scheduling this parent sooner in most cases.
+        nap((schedule != new_schedule) ? 0 : 120)
```
fdr, line 254: why nap for 120s rather than forever?

fdr (approved):

prog/base.rb
jeremyevans: As mentioned in the commit message:

```
Just in case there are unforeseen race conditions that are not handled,
only nap for 120 seconds if there are active children.  Worst case
scenario, this results in a 2 minute delay for running the parent.
However, this can potentially result in 120x less load from parent
strands polling children.
```

PR #3533, opened 2025-06-29
[openapi] use shared request body for load balancer creation requests

geemus:
I noted a couple of other discrepancies/inconsistencies related to this that I wanted to call out (they are a bit less easy to fix and I have limited availability, so I thought I would share/highlight to start and then figure out if I should tackle them).

1. Although the API allows creation of load balancers at either the project layer or project+location layer, it only allows them to be edited at the project+location layer. Unless there is a technical reason for this, it might be better to allow editing at both layers. Delete for load balancers is also only available at the project+location layer, fwiw.
2. The create methods have the same request bodies, but update differs (otherwise I would have moved it to the shared requestBody as well). Patch adds a `vms` field that is not available at creation time. From talking briefly to @fdr it sounds like this was likely just oversight. So it would be great to add `vms` to the create method for consistency (and then shift the patch method to also use the same shared requestBody to simplify and avoid drift).

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -288,40 +288,7 @@ paths:
       operationId: createLoadBalancer
       summary: Create a new Load Balancer in a project
       requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              type: object
-              properties:
-                algorithm:
-                  description: Algorithm of the Load Balancer
-                  type: string
-                stack:
-                  description: 'Networking stack of the Load Balancer (ipv4, ipv6, or dual)'
-                  type: string
-                dst_port:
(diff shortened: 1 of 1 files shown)


fdr (commented):
There are conflicts, but I think the intent is clear, so I had claude generate a "rebase" in https://github.com/ubicloud/ubicloud/pull/4561. Sorry for the missed PR here. we have new tooling and new year's resolution to notify on non-draft PRs and at least say something within 48 hours.

fdr:
Closed per https://github.com/ubicloud/ubicloud/pull/4561#pullrequestreview-3666114796

PR #3534, opened 2025-06-29
Make cloudify_server Hetzner-only

aamederen:
As we developed clover, we made more assumptions about the hosting providers and coupled the code with them, specifically Hetzner. As reported by users, things break in the cloudify_server demo script and we don't do a great job at actively maintaining it.

This commit applies some modifications to the cloudify_server script:
- Make the script Hetzner-only
- Ask for Hetzner details to be entered in .env file beforehand
- Allow re-issue of the script on the same host, for retry attempts
- Bump the VmHost init timeout from 10m to 30m
- Update VM image downloaded to the Host to Ubuntu Noble

----

> [!IMPORTANT]
> Make `cloudify_server` script Hetzner-specific, requiring Hetzner credentials in `.env` and allowing retries on the same host.
> 
>   - **Behavior**:
>     - `cloudify_server` script is now Hetzner-only, exiting if a non-Hetzner provider is selected.
>     - Requires `HETZNER_USER`, `HETZNER_PASSWORD`, `HETZNER_SSH_PUBLIC_KEY`, `HETZNER_SSH_PRIVATE_KEY` in `.env` file.
>     - Allows re-running the script on the same host by checking existing cloudified servers.
>     - Increases `VmHost` initialization timeout from 10m to 30m.
>   - **Documentation**:
>     - Updates `README.md` to reflect Hetzner-only support and `.env` file requirements.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 44336c24d1163209a970663870a9ebc46444820c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/demo/cloudify_server
+++ b/demo/cloudify_server
@@ -7,6 +7,7 @@ require_relative "../loader"
 require "timeout"
 
 LOCATIONS = Location.where(visible: true).map { |l| [l, "#{l.provider} #{l.name} (#{l.display_name})"] }.to_h
+REQUIRED_ENV_VARS = %w[HETZNER_USER HETZNER_PASSWORD HETZNER_SSH_PUBLIC_KEY HETZNER_SSH_PRIVATE_KEY].freeze
 
 def get_input(msg, default = nil)
   prompt = "#{msg}#{default.nil? ? "" : " [default: #{default}]"}: "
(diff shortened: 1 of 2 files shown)


demo/cloudify_server
```diff
@@ -46,32 +47,39 @@ def select_option(msg, options, default = nil)
   selected
 end
 
+unless REQUIRED_ENV_VARS.all? { Config.public_send(it.downcase) }
```
enescakir, line 50: Should we add it to our README as well?

aamederen: Makes sense, we should also remember to update the docs.

demo/cloudify_server
```diff
@@ -46,32 +47,39 @@ def select_option(msg, options, default = nil)
...
+if location.provider != HostProvider::HETZNER_PROVIDER_NAME
+  puts "Currently only Hetzner is supported for this demo. Please use a Hetzner server for cloudification."
+  exit 1
 end
-puts "Add following public SSH key to '/root/.ssh/authorized_keys' on your machine\n\n"
-puts ssh_key
 
-print "\n\nPress enter after you add the above SSH key to your machine\n\n"
-gets.chomp
+strand = if (vmh = VmHost.all.find { it.sshable.host == hostname && it.location == location && it.provider.server_identifier == host_id })
```
jeremyevans:
```suggestion
vm_host_ds = VmHost
  .where(id: Sshable.where(host: hostname).select(:id))
  .where(id: HostProvider.where(server_identifier: host_id).select(:id))
  .where(location_id: location.id)
  
strand = if (vmh = vm_host_ds.first)
```

While it's likely that this script is only run on an empty or near empty database, it's probably best to avoid the 3N+1 queries.

Unrelated to this pull request, currently, this only errors if the combination of sshable hostname, location, and server identifier match. If there is a matching sshable host, but the server identifier is different, or vice versa, or the both match but the location doesn't, should we be creating a new strand for it?  Maybe we should, I'm not sure.

jeremyevans (commented):
Ruby-wise, this looks good to me.

demo/cloudify_server
```diff
@@ -46,32 +47,44 @@ def select_option(msg, options, default = nil)
...
+  puts "Found existing cloudified server for '#{hostname}' with ID: #{vmh.id}"
+  vmh.strand
+else
+  # Create a new strand for the cloudified server
+  Prog::Vm::HostNexus.assemble(
+    hostname,
+    provider_name: location.provider,
+    server_identifier: host_id,
+    location_id: location.id,
+    default_boot_images: ["ubuntu-noble"]
```
jeremyevans, line 82: Is there a reason we only use a single boot image here, instead of the 4 we recommend in DEVELOPERS.md?

aamederen: I guess it's to make the process fast for demo purposes.

PR #3548, opened 2025-07-02
add safeguards so that the API calls do not fail

furkansahin:


prog/aws/vpc.rb
```diff
@@ -119,8 +119,11 @@ class Prog::Aws::Vpc < Prog::Base
   end
 
   label def destroy
+    nap 5 if client.describe_subnets({filters: [{name: "subnet-id", values: [private_subnet.private_subnet_aws_resource.subnet_id]}]}).subnets[0].state != "available"
```
enescakir:
```suggestion
    nap 5 if client.describe_subnets({filters: [{name: "subnet-id", values: [private_subnet.private_subnet_aws_resource.subnet_id]}]}).subnets.first&.state != "available"
```

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -48,6 +48,10 @@ def before_run
   end
 
   label def setup_bucket
+    if postgres_timeline.aws? && !aws_access_key_is_available?
+      nap 1
+    end
```
enescakir:
```suggestion
    nap 1 if postgres_timeline.aws? && !aws_access_key_is_available?
```

PR #3554, opened 2025-07-03
Run CI on macOS runners too

enescakir:
We hit macOS specific issues a few times in the past.

Since macOS runners are free for open source projects, there is no reason not to run CI on them.

bcrypt_pbkdf doesn't support Ruby 3.4 yet; only the rc version is supported for now. https://github.com/net-ssh/bcrypt_pbkdf-ruby/issues/25

    Fetching gem metadata from https://rubygems.org/.......... The
    following platform specific gems are getting installed, yet the
    lockfile includes only their generic ruby version: *
    bcrypt_pbkdf-1.1.1-arm64-darwin Please run `bundle lock
    --normalize-platforms` and commit the resulting lockfile.
    Alternatively, you may run `bundle lock --add-platform
    <list-of-platforms-that-you-want-to-support>`
    bcrypt_pbkdf-1.1.1-arm64-darwin requires ruby version < 3.4.dev, >=
    2.7, which is incompatible with the current version, 3.4.4


enescakir:
@jeremyevans I'm not sure about `bcrypt_pbkdf` change since it's rc

enescakir:
It's faster than Ubuntu 🚀 

CleanShot 2025-07-04 at 13 53 40@2x

jeremyevans:
I don't think there is sufficient value for MacOS CI to use an rc version of bcrypt_pbkdf in production. One possible work around is in the MacOS CI, until there is an actual release, run `bundle update bcrypt-pdkdf --pre`. What are your thoughts on that?

enescakir:
It makes sense to run `bundle update bcrypt-pdkdf --pre` for macOS. However, the Ruby setup at `ruby/setup-ruby` is done while installing gems. We need to run the `bundle update` command between these two operations, but they can't be separated. I'm putting this PR on hold for now; I will recreate it when there’s a non-rc release.

PR #3555, opened 2025-07-03
Run control plane tests in parallel

enescakir:
The CI now takes almost 7 minutes, which can be a bit frustrating when you're trying to merge PRs.

Running tests in parallel speeds up the CI by 2-3 times.


jeremyevans (approved):
Faster CI is better! I think 8 CPUs is a good number for this.

PR #3560, opened 2025-07-04
Various Improvement and Fixes for Ubi on AWS

byucesoy:
We found number of edge cases and improvement opportunities while testing Ubi/PG on AWS. This PR fixes the issues we found.


routes/project/location/postgres.rb
```diff
@@ -242,6 +242,11 @@ class Clover
         authorize("Postgres:edit", pg.id)
 
         name = typecast_params.nonempty_str!("name")
+
+        Validation.validate_name(name)
+
+        Validation.validate_vcpu_quota(@project, "PostgresVCpu", Option::POSTGRES_SIZE_OPTIONS[pg.target_vm_size].vcpu_count)
```
furkansahin, line 248: I think we can skip this for AWS since the quota management is different there?

byucesoy: Will do this in separate PR.

PR #3566, opened 2025-07-04
Simply declare caught up if primary is destroyed when rr is being created

furkansahin:
> [!IMPORTANT]
> Update `lsn_caught_up` in `postgres_server.rb` to handle `nil` parent server or representative server, with corresponding tests in `postgres_server_spec.rb`.
> 
>   - **Behavior**:
>     - Update `lsn_caught_up` in `postgres_server.rb` to handle `nil` parent server or representative server.
>     - Returns `true` if `parent_server` is `nil` or its `current_lsn` is `nil`.
>   - **Tests**:
>     - Add tests in `postgres_server_spec.rb` for `lsn_caught_up` to cover cases where parent server or representative server is `nil`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 492144f5421cf411bf2cc66648d3a24da486afc3. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -149,11 +149,12 @@ class PostgresServer < Sequel::Model
 
   def lsn_caught_up
     parent_server = if read_replica?
-      resource.parent.representative_server
+      resource.parent&.representative_server
     else
       resource.representative_server
     end
-    lsn_diff(parent_server.current_lsn, current_lsn) < 80 * 1024 * 1024
+
+    lsn_diff(parent_server&.current_lsn || current_lsn, current_lsn) < 80 * 1024 * 1024
   end
 
   def current_lsn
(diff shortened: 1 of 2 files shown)


model/postgres/postgres_server.rb
```diff
@@ -149,11 +149,12 @@ def needs_recycling?
...
   def lsn_caught_up
     parent_server = if read_replica?
-      resource.parent.representative_server
+      resource.parent&.representative_server
     else
       resource.representative_server
     end
-    lsn_diff(parent_server.current_lsn, current_lsn) < 80 * 1024 * 1024
+
+    lsn_diff(parent_server&.current_lsn || current_lsn, current_lsn) < 80 * 1024 * 1024
```
shikharbhardwaj, line 157: jw: Can condition this be hit in an unfortunate timing case during a genuine failover? For eg. the primary becomes unhealthy and another server starts taking over by removing the primary, and someone creates a read replica around that moment and it get marked as caught up.

I imagine a very low probably case so safe to ignore.

shikharbhardwaj (approved):

byucesoy (approved):

PR #3570, opened 2025-07-05
Pass JIT config via file to runner script

enescakir:
Previously, we passed the JIT config to the runner execution script via
command-line arguments using xargs. However, GitHub recently increased
the size of JIT tokens by over 4KB, which made this approach unreliable
due to command-line length limits.

Now, we pass the JIT config via a file instead. This approach is more
stable for handling large strings. Additionally, we now persist the
runner’s service config rather than generating it ad hoc with
systemd-run.

This change is currently behind a feature flag. Once we confirm its
reliability, we can move the files created in the setup_environment step
into the image generation process.

----

> [!IMPORTANT]
> Switch JIT config to file-based approach in `prog/vm/github_runner.rb`, controlled by a feature flag, with updated tests.
> 
>   - **Behavior**:
>     - Pass JIT config via file instead of command-line in `prog/vm/github_runner.rb` to handle large tokens.
>     - Controlled by `:runner_jit_file` feature flag in `model/project.rb`.
>     - Backward compatible; falls back to old method if feature flag is off.
>   - **Tests**:
>     - Update `spec/prog/vm/github_runner_spec.rb` to test new file-based JIT config behavior.
>     - Add tests for feature flag behavior in `prog/test/github_runner.rb`.
>   - **Misc**:
>     - Add `:runner_jit_file` to feature flags in `model/project.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b3599e2bf73237307fb9159297a147ba927b8867. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -245,6 +245,31 @@ class Prog::Vm::GithubRunner < Prog::Base
       COMMAND
     end
 
+    if github_runner.installation.project.get_ff_runner_jit_file
+      # After testing in production for a while, we can move these files to the image generation process
+      command += <<~COMMAND
+        sudo tee /etc/systemd/system/runner-script.service > /dev/null > /dev/null <<'EOT'
+        [Unit]
+        Description=runner-script
+
+        [Service]
+        RemainAfterExit=yes
+        User=runner
+        Group=runner
+        WorkingDirectory=/home/runner
+        ExecStart=/home/runner/actions-runner/run-withenv.sh
+        EOT
+
+        sudo -u runner tee /home/runner/actions-runner/run-withenv.sh > /dev/null <<'EOT'
+        #!/bin/bash
+        mapfile -t env </etc/environment
+        JIT_CONFIG="$(cat ./actions-runner/.jit_token)"
+        exec env -- "${env[@]}" ./actions-runner/run.sh --jitconfig "$JIT_CONFIG"
+        EOT
+        sudo systemctl daemon-reload
+      COMMAND
+    end
+
     # Remove comments and empty lines before sending them to the machine
     vm.sshable.cmd(command.gsub(/^(\s*# .*)?\n/, ""))
 
@@ -258,12 +283,19 @@ class Prog::Vm::GithubRunner < Prog::Base
     data = {name: github_runner.ubid.to_s, labels: [github_runner.label], runner_group_id: 1, work_folder: "/home/runner/work"}
     response = github_client.post("/repos/#{github_runner.repository_name}/actions/runners/generate-jitconfig", data)
     github_runner.update(runner_id: response[:runner][:id], ready_at: Time.now)
-    # We initiate an API call and a SSH connection under the same label to avoid
-    # having to store the encoded_jit_config.
-    vm.sshable.cmd("sudo -- xargs -0 -- systemd-run --uid runner --gid runner " \
-                   "--working-directory '/home/runner' --unit runner-script --description runner-script --remain-after-exit -- " \
-                   "/home/runner/actions-runner/run-withenv.sh",
-      stdin: response[:encoded_jit_config].gsub("$", "$$"))
+    if github_runner.installation.project.get_ff_runner_jit_file
+      vm.sshable.cmd(<<~COMMAND, stdin: response[:encoded_jit_config])
+      sudo -u runner tee /home/runner/actions-runner/.jit_token > /dev/null
+      sudo systemctl start runner-script.service
+      COMMAND
+    else
+      # We initiate an API call and a SSH connection under the same label to avoid
+      # having to store the encoded_jit_config.
+      vm.sshable.cmd("sudo -- xargs -0 -- systemd-run --uid runner --gid runner " \
+                    "--working-directory '/home/runner' --unit runner-script --description runner-script --remain-after-exit -- " \
+                    "/home/runner/actions-runner/run-withenv.sh",
+        stdin: response[:encoded_jit_config].gsub("$", "$$"))
+    end
     github_runner.log_duration("runner_registered", github_runner.ready_at - github_runner.allocated_at)
 
     hop_wait
--- a/model/project.rb
+++ b/model/project.rb
@@ -170,7 +170,8 @@ class Project < Sequel::Model
     end
   end
 
-  feature_flag :vm_public_ssh_keys, :location_latitude_fra, :access_all_cache_scopes, :allocator_diagnostics, :private_locations, :free_runner_upgrade_until, :gpu_vm, :postgres_lantern
+  feature_flag :vm_public_ssh_keys, :location_latitude_fra, :access_all_cache_scopes, :allocator_diagnostics, :private_locations, :free_runner_upgrade_until, :gpu_vm, :postgres_lantern,
+    :runner_jit_file
 end
 
 # Table: project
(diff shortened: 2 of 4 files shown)


prog/vm/github_runner.rb
```diff
@@ -245,6 +245,31 @@ def setup_info
       COMMAND
     end
 
+    if github_runner.installation.project.get_ff_runner_jit_file
+      # After testing in production for a while, we can move these files to the image generation process
+      command += <<~COMMAND
+        sudo tee /etc/systemd/system/runner-script.service > /dev/null > /dev/null <<'EOT'
```
jeremyevans, line 251:
```suggestion
        sudo tee /etc/systemd/system/runner-script.service > /dev/null 2>&1 <<'EOT'
```

Assuming you are trying to redirect both stdout and stderr to /dev/null.

jeremyevans (approved):

fdr:
> Previously, we passed the JIT config to the runner execution script via
command-line arguments using xargs. However, GitHub recently increased
the size of JIT tokens by over 4KB, which made this approach unreliable
due to command-line length limits.

Were you able to reproduce this?

If so, please place detail. If not, this looks too firm for what is speculation.

enescakir:
> > Previously, we passed the JIT config to the runner execution script via
> > command-line arguments using xargs. However, GitHub recently increased
> > the size of JIT tokens by over 4KB, which made this approach unreliable
> > due to command-line length limits.
> 
> Were you able to reproduce this?
> 
> If so, please place detail. If not, this looks too firm for what is speculation.

I couldn't reproduce the issue locally. It shows the following exception in production: somehow, the JIT token has been corrupted. DP printed a shorter JIT token than the CP log, but it doesn't contain any special characters.

```
2025-07-07 10:06:11  2025-07-07 10:06:11 +0000 grvy2e9mer6a2jgp4af41yyrn6 ssh cmd execution | cat: /run/systemd/transient/runner-script.service: No such file or directory
Jul 07 10:06:10 vmdvtkt4 systemd[1]: /run/systemd/transient/runner-script.service:11: Failed to resolve unit specifiers in <********LONG_JIT_TOKEN********>
Jul 07 10:06:10 vmdvtkt4 systemd[1]: runner-script.service: Unit configuration has fatal error, unit will not be started.
Jul 07 10:06:10 vmdvtkt4 systemd[1]: Failed to start runner-script.
░░ Subject: A start job for unit runner-script.service has failed
░░ Defined-By: systemd
░░ Support: http://www.ubuntu.com/support
░░ 
░░ A start job for unit runner-script.service has finished with a failure.
░░ 
░░ The job identifier is 1075 and the job result is failed.

```

prog/vm/github_runner.rb
enescakir: If something goes wrong, I think we should see the `stderr`.

enescakir: I only redirect stdout because tee also prints the file content.

fdr:
> > > Previously, we passed the JIT config to the runner execution script via
> > > command-line arguments using xargs. However, GitHub recently increased
> > > the size of JIT tokens by over 4KB, which made this approach unreliable
> > > due to command-line length limits.
> > 
> > 
> > Were you able to reproduce this?
> > If so, please place detail. If not, this looks too firm for what is speculation.
> 
> I couldn't reproduce the issue locally. It shows the following exception in production: somehow, the JIT token has been corrupted. DP printed a shorter JIT token than the CP log, but it doesn't contain any special characters.
> 
> ```
> 2025-07-07 10:06:11  2025-07-07 10:06:11 +0000 grvy2e9mer6a2jgp4af41yyrn6 ssh cmd execution | cat: /run/systemd/transient/runner-script.service: No such file or directory
> Jul 07 10:06:10 vmdvtkt4 systemd[1]: /run/systemd/transient/runner-script.service:11: Failed to resolve unit specifiers in <********LONG_JIT_TOKEN********>
> Jul 07 10:06:10 vmdvtkt4 systemd[1]: runner-script.service: Unit configuration has fatal error, unit will not be started.
> Jul 07 10:06:10 vmdvtkt4 systemd[1]: Failed to start runner-script.
> ░░ Subject: A start job for unit runner-script.service has failed
> ░░ Defined-By: systemd
> ░░ Support: http://www.ubuntu.com/support
> ░░ 
> ░░ A start job for unit runner-script.service has finished with a failure.
> ░░ 
> ░░ The job identifier is 1075 and the job result is failed.
> ```

If we can't reproduce it, it seems like there must be more to the story. I don't object to applying this patch, but it makes a pretty unambiguous claim (length is the culprit) that should be sustained by experimentation, and that's not the case: there's something missing.

It can be tweaked to say that: systemd-run operates by templating templating its unit files, placed in `/run/systemd/system/`, and we suspect something is going wrong there. We decided to take more direct control over the unit file to further our understanding.

I don't mind applying patches that are sound but don't necessarily solve the issue, and maybe make at least one thing better...e.g. 00b73fbf9cda01beb0426acea04284e0a0028ba1, though I should have been even more clear that we never spotted a JIT token with a bad character sequence.

enescakir:
> > > > Previously, we passed the JIT config to the runner execution script via
> > > > command-line arguments using xargs. However, GitHub recently increased
> > > > the size of JIT tokens by over 4KB, which made this approach unreliable
> > > > due to command-line length limits.
> > > 
> > > 
> > > Were you able to reproduce this?
> > > If so, please place detail. If not, this looks too firm for what is speculation.
> > 
> > 
> > I couldn't reproduce the issue locally. It shows the following exception in production: somehow, the JIT token has been corrupted. DP printed a shorter JIT token than the CP log, but it doesn't contain any special characters.
> > ```
> > 2025-07-07 10:06:11  2025-07-07 10:06:11 +0000 grvy2e9mer6a2jgp4af41yyrn6 ssh cmd execution | cat: /run/systemd/transient/runner-script.service: No such file or directory
> > Jul 07 10:06:10 vmdvtkt4 systemd[1]: /run/systemd/transient/runner-script.service:11: Failed to resolve unit specifiers in <********LONG_JIT_TOKEN********>
> > Jul 07 10:06:10 vmdvtkt4 systemd[1]: runner-script.service: Unit configuration has fatal error, unit will not be started.
> > Jul 07 10:06:10 vmdvtkt4 systemd[1]: Failed to start runner-script.
> > ░░ Subject: A start job for unit runner-script.service has failed
> > ░░ Defined-By: systemd
> > ░░ Support: http://www.ubuntu.com/support
> > ░░ 
> > ░░ A start job for unit runner-script.service has finished with a failure.
> > ░░ 
> > ░░ The job identifier is 1075 and the job result is failed.
> > ```
> 
> If we can't reproduce it, it seems like there must be more to the story. I don't object to applying this patch, but it makes a pretty unambiguous claim (length is the culprit) that should be sustained by experimentation, and that's not the case: there's something missing.
> 
> It can be tweaked to say that: systemd-run operates by templating templating its unit files, placed in `/run/systemd/system/`, and we suspect something is going wrong there. We decided to take more direct control over the unit file to further our understanding.
> 
> I don't mind applying patches that are sound but don't necessarily solve the issue, and maybe make at least one thing better...e.g. [00b73fb](https://github.com/ubicloud/ubicloud/commit/00b73fbf9cda01beb0426acea04284e0a0028ba1), though I should have been even more clear that we never spotted a JIT token with a bad character sequence.

That makes sense. I updated the commit message.

velioglu (approved):

PR #3579, opened 2025-07-07
Add PG Configuration UI

shikharbhardwaj:
https://github.com/user-attachments/assets/f70d70ad-32a6-491d-87b5-3378e6659eeb


routes/project/location/postgres.rb
```diff
@@ -484,37 +484,17 @@ class Clover
...
         end
 
-        r.post do
+        r.on method: [:post, :patch] do
           authorize("Postgres:edit", pg.id)
 
-          pg_config = typecast_params.Hash("pg_config")
-          pgbouncer_config = typecast_params.Hash("pgbouncer_config")
+          pg_config = typecast_params.Hash("pg_config") || {}
+          pgbouncer_config = typecast_params.Hash("pgbouncer_config") || {}
```
jeremyevans, line 500: If both of these hashes are empty, I recommend an early exit with no changes.

views/components/delete_button.erb
```diff
@@ -1,4 +1,4 @@
-<%# locals: (url: request.path, csrf_url: url, text: "Delete", confirmation: nil, confirmation_message: nil, redirect: request.path, method: "DELETE") %>
+<%# locals: (url: request.path, csrf_url: url, text: "Delete", confirmation: nil, confirmation_message: "Are you sure to delete?", redirect: request.path, method: "DELETE") %>
```
jeremyevans:
```suggestion
<%# locals: (url: request.path, csrf_url: url, text: "Delete", confirmation: nil, confirmation_message: "Are you sure you want to delete?", redirect: request.path, method: "DELETE") %>
```

This reads better to me.

byucesoy:
`+` button implies that it would only add a new row, but it also saves the previous change. Save behavior is more critical than adding row. I think the icon should depict saving. I think we should decouple saving and adding row behavior. Maybe we can have 2 icons one for save and one for adding a row?

shikharbhardwaj:
@byucesoy any changes to existing configs are done by the inline edit for its row, the `+` button does not affect them. Though if a user edits anything inline and then adds a row, they will lose the inline edits they did not confirm. Similarly for a new row that's not added but an inline edit that is confirmed. Maybe we can add safeguards to warn the user to confirm any entries they've modified before adding a new row and vice versa.

I also had another version of this which might be closer to what you are describing here:


https://github.com/user-attachments/assets/251f4dfe-a0bb-44c1-bc10-edbcde6a42cd

byucesoy:
Yes, the new video is closer to what I had my mind.

I think the primary part that concerns me is the meaning of `+` button is not clear. It can mean:

1. **add** this config to user-defined configs. 
2.  **add** new row to UI.

In this PR it means both. We both add the config to user-defined configs and also add a row to the UI.

People expecting only 2 to happen (and assume they would have another way of saving the work), would get confused. So I think we should have separate add a row (`+`) and add to user config (maybe a diskette icon) buttons. Alternatively what you have in your latest video also works.

byucesoy:
Another option could be changing the + icon with diskette icon, which emphasizes the more serious effect (saving the config/doing form submission) instead of emphasizing the minor effect (adding new row to UI). Please let me know when this PR is good for another round of review.

shikharbhardwaj:
Closing this one in favour of https://github.com/ubicloud/ubicloud/pull/3622

PR #3594, opened 2025-07-09
Add support for locking domains to a specific OIDC provider

jeremyevans:
Organizations using OIDC providers will often want the OIDC provider to be used to all authentication.  The simplest way to implement this support is to allow locking domains to an OIDC provider. During login, we check if the account is in a locked domain. If so, we reject the login unless the authentication uses the OIDC provider that the domain is locked to.

----

> [!IMPORTANT]
> Add support for locking domains to specific OIDC providers, affecting authentication and account management processes.
> 
>   - **Behavior**:
>     - Domains can be locked to specific OIDC providers, restricting authentication to those providers.
>     - `before_login`, `before_create_account`, `before_reset_password`, and other hooks in `clover.rb` check for locked domains and redirect if necessary.
>     - `check_locked_domain` function added to handle domain checks and redirections.
>   - **Models**:
>     - New `LockedDomain` model with `domain` and `oidc_provider_id` fields.
>     - `OidcProvider` model updated to include `one_to_many :locked_domains` association.
>   - **Migrations**:
>     - New migration `20250708_locked_domain.rb` to create `locked_domain` table.
>   - **Views**:
>     - `submenu.erb` updated to hide certain options if the domain is locked.
>   - **Tests**:
>     - Extensive tests added in `auth_spec.rb` to cover locked domain scenarios, including login, account creation, and password reset.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8babb5e65840b7988875f9e68f8c20c88509da1c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fdr (approved):
I am not as skilled as seeing the forest for the trees as @enescakir  is in this code, but nothing looks surprising? I suppose the issue with this feature is, it's mostly about thinking about every other feature currently in the system, and besides the-author-of-Rodauth, @enescakir is that person.

PR #3598, opened 2025-07-09
Update to Alma Linux 9.6

jeremyevans:
Alma Linux 9.5 is no longer available for download, so without this, you cannot setup a new host with an almalinux-9 boot image.


fdr:
You do need to pre-warm enough hosts with the image before changing the default though, otherwise you hit zero capacity problems

aamederen:
We have only 1 almalinux VM right now, created a month ago and I couldn't find any almalinux in our archived records of our past week. I think it's safe enough to switch now and follow-up with download afterwards.

jeremyevans:
While @aamederen is probably right that we wouldn't have a problem taking a shortcut in this case, I would like to get in the habit of doing this safely, so I can get experience handling future cases where it would matter.

To that end, I've added a new PR that supports Alma Linux 9.6 while leaving Alma Linux 9.5 the default: #3616 .  After that is merged, we can download Alma Linux 9.6 to hosts.  After Alma Linux 9.6 has been downloaded, we can merge this to switch the default from 9.5 to 9.6.

PR #3600, opened 2025-07-10
Log how many runners can fit into available capacity

enescakir:
- **Fix unreachable expectations in the utilization log test**
  We replaced `:location` with `:location_id` in the VM host table.
  
  We didn't catch this bug earlier because the if condition didn't match,
  so it just skipped the expectation.
  

- **Group utilization metrics with family along with arch**
  Since standard and premium VMs are allocated on their corresponding
  hosts, it's better to log their utilization separately. We might have
  available capacity for standard VMs but not for premium VMs.
  

- **Log how many runners can fit into available capacity**
  We currently log overall utilization, which gives a good high-level
  view, but it’s not very useful for larger runner sizes. For example, we
  might have enough free cores for a standard-30 runner, but if those
  cores are fragmented across multiple hosts, we can’t actually allocate
  one.
  
  This change adds tracking for how many runners of each type can
  realistically fit into the available capacity on the hosts.
  

----

> [!IMPORTANT]
> Enhance VM host utilization logging by adding runner count metrics and fixing test expectations.
> 
>   - **Logging Enhancements**:
>     - Add `family` to grouping in `Prog::LogVmHostUtilizations` to separate standard and premium VM utilization.
>     - Log available runner counts (`available_standard_2_count`, `available_standard_4_count`, etc.) based on core availability in `prog/log_vm_host_utilizations.rb`.
>   - **Test Fixes**:
>     - Replace `:location` with `:location_id` in `spec/prog/log_vm_host_utilizations_spec.rb` to fix unreachable expectations.
>     - Update test cases to include `family` and validate new runner count metrics.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7777931d9f9bb5e6e500a55ded2a4d1ff90c14c0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/log_vm_host_utilizations.rb
+++ b/prog/log_vm_host_utilizations.rb
@@ -11,19 +11,25 @@ class Prog::LogVmHostUtilizations < Prog::Base
         round(sum(:used_cores) * 100.0 / sum(:total_cores), 2).cast(:float).as(:core_utilization),
         sum(:used_hugepages_1g).as(:used_hugepages_1g),
         sum(:total_hugepages_1g).as(:total_hugepages_1g),
-        round(sum(:used_hugepages_1g) * 100.0 / sum(:total_hugepages_1g), 2).cast(:float).as(:hugepage_utilization)
+        round(sum(:used_hugepages_1g) * 100.0 / sum(:total_hugepages_1g), 2).cast(:float).as(:hugepage_utilization),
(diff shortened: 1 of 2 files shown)


aamederen (approved):
Following up with a dashboard on mezmo maybe?

PR #3610, opened 2025-07-11
Add new semaphore for full mesh connectivity between k8s worker nodes

mohi-kalantari:
A new semaphore is added to start the rekey process, which creates an SSH key pair for each worker node and distributes all public keys to all worker VMs so they can connect to each other.


aamederen:
When will this be useful?

mohi-kalantari:
> When will this be useful?

I'm in the last step of CSI which I need to do rsyncs across nodes and copy the data. I need the nodes to be able to ssh to each other

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -185,6 +188,25 @@ def before_run
...
+      {vm: vm, ssh_key: SshKey.generate}
+    end
+
+    public_keys = key_pairs.map { |kp| kp[:ssh_key].public_key }
+    key_pairs.each do |kp|
+      vm = kp[:vm]
+      private_key = kp[:ssh_key].private_key
+      vm.sshable.cmd("echo '#{private_key}' > ~/.ssh/id_rsa && chmod 600 ~/.ssh/id_rsa", log: false)
+      all_keys_str = ([vm.sshable.keys.first.public_key] + public_keys).join("\\n")
+      vm.sshable.cmd("echo -e '#{all_keys_str}' > ~/.ssh/authorized_keys")
```
aamederen: maybe `tee` here?

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -185,6 +188,25 @@ def before_run
...
+    key_pairs = kubernetes_cluster.worker_vms.map do |vm|
+      {vm: vm, ssh_key: SshKey.generate}
+    end
+
+    public_keys = key_pairs.map { |kp| kp[:ssh_key].public_key }
+    key_pairs.each do |kp|
+      vm = kp[:vm]
+      private_key = kp[:ssh_key].private_key
+      vm.sshable.cmd("echo '#{private_key}' > ~/.ssh/id_rsa && chmod 600 ~/.ssh/id_rsa", log: false)
+      all_keys_str = ([vm.sshable.keys.first.public_key] + public_keys).join("\\n")
```
aamederen: nit: instead of doing an array concat + join each time, maybe generating the string of public keys outside the loop could be more efficient

aamederen (approved):

prog/kubernetes/kubernetes_cluster_nexus.rb
jeremyevans: I would recommend `tee` with `stdin`. I don't think the current code has security issues as public/private key values shouldn't contain `'`, but I think it's better to avoid the issue.  Plus, you don't have to use `\\n` and rely on `echo -e` changing them to `\n`.

Similar changes for the private key two lines prior.

mohi-kalantari: fixed thanks

prog/kubernetes/kubernetes_cluster_nexus.rb
mohi-kalantari: since the first element of array changes on every iteration, I guess that doesn't work 🤔

PR #3612, opened 2025-07-11
Regen screenshot updates

shikharbhardwaj:
* Fix regen-screenshots to mock Kubernetes LB ID

* Add more PG features in regen-screenshots

New screenshots used in: https://github.com/ubicloud/documentation/pull/93/

----

> [!IMPORTANT]
> Enhance `regen-screenshots` script with PostgreSQL and Kubernetes features, mock components, and add new screenshots.
> 
>   - **Enhancements in `bin/regen-screenshots`**:
>     - Mock Kubernetes Load Balancer ID to associate with clusters.
>     - Add PostgreSQL features: `PostgresServer`, `PostgresTimeline`, `PostgresFirewallRule`, and `VictoriaMetrics` resources.
>     - Mock VictoriaMetrics client to return sample data for consistent results.
>     - Add screenshots for PostgreSQL and Kubernetes features, including connection, networking, resizing, and configuration.
>     - Update methods in `PostgresResource` and `PostgresServer` to support new features.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 5490afb33300c1c4ce84d7180b3e47502af07eb3. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bin/regen-screenshots
+++ b/bin/regen-screenshots
@@ -134,6 +134,7 @@ class RegenScreenshots
 
     resize(1400)
     click_link "Create PostgreSQL Database"
+    screenshot "managed-postgresql/create-screenshot.png"
     screenshot "quick-start/using-kamal-with-ubicloud-3-screenshot.png"
 
     pg_resource = PostgresResource.create(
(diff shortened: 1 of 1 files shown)


bin/regen-screenshots
```diff
@@ -133,7 +133,10 @@ class RegenScreenshots
     screenshot "managed-postgresql/overview-1-screenshot.png"
 
     resize(1400)
+    # Hide billing rates for PostgreSQL create screenshot
+    BillingRate.define_singleton_method(:unit_price_from_resource_properties) { |*| 0 }
```
jeremyevans: Can you explain the reasoning behind this? It makes the screenshot not reflect reality, so absent a good reason, I think it should be removed.

bin/regen-screenshots
```diff
@@ -152,16 +155,147 @@ class RegenScreenshots
...
+    def mock_client.query_range(query:, start_ts:, end_ts:)
+      # Generate sample time series data based on the query
+      Time.now.to_i
+      step = 60 # 1 minute intervals
+      timestamps = (start_ts..end_ts).step(step).to_a
+
+      case query
+      when /node_filesystem_avail_bytes/
+        # Return disk usage around 45%
+        values = timestamps.map { |ts| [ts, rand(40..49).to_s] }
```
jeremyevans: In general, we want to avoid randomness in regen-screenshots.  2 consecutive runs of regen-screenshots should results in the same screenshots.  You can generate the random numbers using this code, and then replace this code with the generated values (you can keep the code as comments in this file if you want).

bin/regen-screenshots
shikharbhardwaj: In some other (manually captured) screenshots, I saw that the billing rates were removed (for eg: https://ubicloud.com/docs/managed-postgresql/high-availability). I believe this is to avoid putting misleading pricing information in pages where its not required, so I tried to remove them from the automatically generated pages too.
This is not 100% clean either as it shows 0.00 for everything, would have liked to do the blank info here instead if possible.

jeremyevans: My preference is to show current pricing in the screenshots, and regenerate the screenshots when the pricing changes. I don't think that approach results in misleading pricing information, but maybe there are cases I am not considering.

bin/regen-screenshots
shikharbhardwaj: I think setting the RNG seed to a fixed value might be cleaner, wdyt?

jeremyevans: Assuming you use a separate RNG for this test data, that is not used for anything else, that would probably work.

bin/regen-screenshots
shikharbhardwaj: I think that is reasonable, with the autogenerated screenshots we can update the docs on any billing rate changes

jeremyevans (approved):

byucesoy (approved):

PR #3615, opened 2025-07-12
Fix private key format of SshKey generation

fdr:
Previously, this program generated private SSH key binaries that
worked with net-ssh but not with openssh/libcrypto.

You can reproduce pre-patch like so:

    load 'lib/ssh_key.rb'
    File.write('/tmp/test.key', SshKey.generate.private_key, perm: 0600)
    system('ssh-add /tmp/test.key')

It'll write:

    Error loading key "/tmp/test.key": error in libcrypto

And after the patch, it writes:

    Identity added: /tmp/test.key (/tmp/test.key)

There were three deviations where net-ssh has more relaxed
requirements than openssh/libcrypto, all of which needed to be
addressed to load the generated key text with `ssh`:

1. The comment field in private keys is required.
2. Padding bytes must be set to values one through eight *in binary.*
3. The public key data must be encapsulated in another SSH bitstring,
   including the public key type before the payload.

Regarding the third point, the system maintains three representations
of each public key: one in the `public_key` column in ASCII, and two
within the openssh private key blob. The two copies in the private key
blob are mandated by the openssh format. The separate ASCII column
enables SQL-based auditing of non-sensitive public key copies without
requiring application-level decryption, making it easier to track
which keys are intended to be active at any time, by digesting them
and matching them with the same digests reported by OpenSSH in logs.

----

> [!IMPORTANT]
> Fixes SSH private key format in `SshKey` for openssh compatibility and adds tests for verification.
> 
>   - **Behavior**:
>     - Fixes private key format in `SshKey` to be compatible with openssh/libcrypto by adding a required comment field, setting padding bytes to binary values, and encapsulating public key data in an SSH bitstring.
>     - Maintains three representations of each public key for auditing and compatibility.
>   - **Tests**:
>     - Adds test in `ssh_key_spec.rb` to verify generated private keys are compatible with openssh using `ssh-keygen` for signing and verifying data.
>     - Ensures `SshKey` can generate and load ed25519 keys without errors.
>   - **Misc**:
>     - Adds `base64`, `ed25519`, and `net/ssh` dependencies to `ssh_key.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8ed8d86b5209d91bacb7a6036ea4369a4d145542. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/lib/ssh_key_spec.rb
+++ b/spec/lib/ssh_key_spec.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require "open3"
+require "tmpdir"
+
 require "net/ssh"
 
 RSpec.describe SshKey do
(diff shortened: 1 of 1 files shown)


fdr:
> This seems fine to me, but since the point of this is compatibility, I think it may be worth adding tests to check that `ssh-add` will accept the key. This is a bit tricky, but should work if you use `ssh-agent -D -a test-socket` and `SSH_AUTH_SOCK=test-socket ssh-add -` and provide the key data on stdin.

I am kind wary of making openssh a dependency to run tests though I guess everyone has it, so maybe that's silly on my part.

jeremyevans (approved):

fdr:
took a number of flaked runs, but it did blend.

PR #3622, opened 2025-07-15
Add PG Configuration UI (unified submit)

shikharbhardwaj:
https://github.com/user-attachments/assets/e70effd8-fa38-47a1-9558-e82d0dd45c08

----

> [!IMPORTANT]
> Add PostgreSQL configuration management UI with backend support for handling configuration updates and validation.
> 
>   - **UI Enhancements**:
>     - Add `setupPgConfigCard()` in `assets/js/app.js` for handling add/delete config actions.
>     - New `config_card.erb` component for displaying configuration entries.
>     - Add `config.erb` view for PostgreSQL configuration management.
>     - Update `submenu.erb` to include "Configuration" in the navigation.
>   - **Backend Changes**:
>     - Add `config_hash_from_kvs()` in `helpers/web.rb` to convert key-value pairs to hash.
>     - Update `routes/project/location/postgres.rb` to handle GET, POST, and PATCH requests for `/config`.
>     - Modify `PostgresConfigValidator` in `lib/validation/postgres_config_validator.rb` to include empty value validation.
>   - **Testing**:
>     - Add tests in `postgres_spec.rb` for viewing, updating, and error handling in configuration management.
>     - Update `postgres_config_validator_spec.rb` to test empty value validation and other scenarios.
>   - **Miscellaneous**:
>     - Update `tailwind.config.js` to include `border-red-500` in the safelist.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 05d97650735c7aef8c7dc81721587366da8373dc. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -5,6 +5,7 @@ $(function () {
   setupPlayground();
   setupFormsWithPatchMethod();
   setupMetricsCharts();
+  setupPgConfigCard();
 });
 
 $(".toggle-mobile-menu").on("click", function (event) {
@@ -992,4 +993,43 @@ function flexiblePrecision(value, precision) {
   const increasedPrecision = Math.max(1, precision);
 
   return (value < 10) ? value.toFixed(increasedPrecision) : value.toFixed(precision);
+}
+
+function setupPgConfigCard() {
+  $(".delete-config-btn").on("click", function (event) {
+    const configGroup = $(this).closest(".group");
+    configGroup.remove();
+  });
+
+  $(".add-config-btn").on("click", function (event) {
+    event.preventDefault();
+    addConfigRow(event.target);
+  });
+
+  function addConfigRow(addBtn) {
+    const createConfigGroup = $(addBtn).closest(".group");
+    const keyInput = createConfigGroup.find("input").eq(0);
+    const valueInput = createConfigGroup.find("input").eq(1);
+    if (!keyInput[0].reportValidity()) return;
+    
+    const placeHolderGroup = $(addBtn).closest(".group").siblings(".config-placeholder-group");
+    const configId = placeHolderGroup.data("config-id");
+
+    const newConfigId = configId + 1;
+    const newConfigGroup = placeHolderGroup.clone(true);
+    newConfigGroup.data("config-id", newConfigId);
+    newConfigGroup.find("input").prop("disabled", false);
+
+    newConfigGroup.removeClass("hidden");
+    newConfigGroup.removeClass("config-placeholder-group");
+    newConfigGroup.addClass("config-group");
+
+    newConfigGroup.find("input").eq(0).prop("value", keyInput.val());
+    newConfigGroup.find("input").eq(1).prop("value", valueInput.val());
+
+    keyInput.val("");
+    valueInput.val("");
+
+    $(addBtn).closest(".group").before(newConfigGroup);
+  }
 }
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -99,7 +99,7 @@ class Clover
         end
       end
 
-      r.get web?, %w[overview connection charts networking resize high-availability read-replica backup-restore settings] do |page|
+      r.get web?, %w[overview connection charts networking resize high-availability read-replica backup-restore config settings] do |page|
         authorize("Postgres:view", pg.id)
 
         next if pg.read_replica? && %w[resize high-availability read-replica backup-restore].include?(page)
@@ -493,53 +493,55 @@ class Clover
           }
         end
 
-        r.post do
+        r.on method: [:post, :patch] do
           authorize("Postgres:edit", pg.id)
 
-          pg_config = typecast_params.Hash("pg_config")
-          pgbouncer_config = typecast_params.Hash("pgbouncer_config")
+          if web?
+            pg_keys = typecast_params.array(:str, "pg_config_keys") || []
+            pg_values = typecast_params.array(:str, "pg_config_values") || []
+            pg_config = config_hash_from_kvs(pg_keys, pg_values)
+
+            pgbouncer_keys = typecast_params.array(:str, "pgbouncer_config_keys") || []
+            pgbouncer_values = typecast_params.array(:str, "pgbouncer_config_values") || []
+            pgbouncer_config = config_hash_from_kvs(pgbouncer_keys, pgbouncer_values)
+          else
+            pg_config = typecast_params.Hash("pg_config") || {}
+            pgbouncer_config = typecast_params.Hash("pgbouncer_config") || {}
+          end
+
+          # For PATCH requests, merge with existing config
+          if r.patch?
+            pg_config = pg.user_config.merge(pg_config).compact
+            pgbouncer_config = pg.pgbouncer_user_config.merge(pgbouncer_config).compact
+          end
 
           pg_validator = Validation::PostgresConfigValidator.new(pg.version)
-          pg_validator.validate(pg_config)
+          pg_errors = pg_validator.validation_errors(pg_config)
 
           pgbouncer_validator = Validation::PostgresConfigValidator.new("pgbouncer")
-          pgbouncer_validator.validate(pgbouncer_config)
+          pgbouncer_errors = pgbouncer_validator.validation_errors(pgbouncer_config)
+
+          if pg_errors.any? || pgbouncer_errors.any?
+            pg_errors = pg_errors.map { |key, value| ["pg_config.#{key}", value] }.to_h
+            pgbouncer_errors = pgbouncer_errors.map { |key, value| ["pgbouncer_config.#{key}", value] }.to_h
+            raise Validation::ValidationFailed.new(pg_errors.merge(pgbouncer_errors))
+          end
 
           pg.update(user_config: pg_config, pgbouncer_user_config: pgbouncer_config)
-          audit_log(pg, "update")
 
           pg.servers.each(&:incr_configure)
 
-          {
-            pg_config: pg.user_config,
-            pgbouncer_config: pg.pgbouncer_user_config
-          }
-        end
-
-        r.patch do
-          authorize("Postgres:edit", pg.id)
-
-          pg_config = typecast_params.Hash("pg_config")
-          pgbouncer_config = typecast_params.Hash("pgbouncer_config")
-
-          pg_config = pg.user_config.merge(pg_config).compact
-          pgbouncer_config = pg.pgbouncer_user_config.merge(pgbouncer_config).compact
-
-          pg_validator = Validation::PostgresConfigValidator.new(pg.version)
-          pg_validator.validate(pg_config)
-
-          pgbouncer_validator = Validation::PostgresConfigValidator.new("pgbouncer")
-          pgbouncer_validator.validate(pgbouncer_config)
-
-          pg.update(user_config: pg_config, pgbouncer_user_config: pgbouncer_config)
           audit_log(pg, "update")
 
-          pg.servers.each(&:incr_configure)
-
-          {
-            pg_config: pg.user_config,
-            pgbouncer_config: pg.pgbouncer_user_config
-          }
+          if api?
+            {
+              pg_config: pg.user_config,
+              pgbouncer_config: pg.pgbouncer_user_config
+            }
+          else
+            flash["notice"] = "Configuration updated successfully"
+            r.redirect "#{@project.path}#{pg.path}/config"
+          end
         end
       end
     end
--- a/spec/routes/web/project/postgres_spec.rb
+++ b/spec/routes/web/project/postgres_spec.rb
@@ -811,5 +811,82 @@ RSpec.describe Clover, "postgres" do
         expect { find ".delete-btn" }.to raise_error Capybara::ElementNotFound
       end
     end
+
+    describe "config" do
+      it "can view configuration" do
+        pg.update(user_config: {"max_connections" => "120"})
+        visit "#{project.path}#{pg.path}/config"
+
+        expect(page).to have_content "PostgreSQL Configuration"
+        expect(page).to have_field "pg_config_keys[]", with: "max_connections"
+        expect(page).to have_field "pg_config_values[]", with: "120"
+      end
+
+      it "does not show update button when user does not have permissions" do
+        pg_wo_permission.update(user_config: {"max_connections" => "120"})
+        AccessControlEntry.create_with_id(project_id: project_wo_permissions.id, subject_id: user.id, action_id: ActionType::NAME_MAP["Postgres:view"])
+
+        visit "#{project_wo_permissions.path}#{pg_wo_permission.path}/config"
+        expect(page.title).to eq "Ubicloud - pg-without-permission"
+
+        expect { find ".delete-config-btn" }.to raise_error Capybara::ElementNotFound
+        expect { find ".save-config-btn" }.to raise_error Capybara::ElementNotFound
+      end
+
+      it "shows update button when user has permissions" do
+        visit "#{project.path}#{pg.path}/config"
+        expect(page).to have_button "Save"
+
+        expect { find ".pg-config-card .delete-config-btn" }.not_to raise_error
+        expect { find ".save-config-btn" }.not_to raise_error
+      end
+
+      it "can update configuration" do
+        pg
+        pg.update(user_config: {"max_connections" => "120"})
+        visit "#{project.path}#{pg.path}/config"
+
+        within ".pg-config-card .new-config" do
+          fill_in "pg_config_keys[]", with: "max_connections"
+          fill_in "pg_config_values[]", with: "240"
+        end
+        click_button "Save"
+
+        expect(page).to have_field "pg_config_keys[]", with: "max_connections"
+        expect(page).to have_field "pg_config_values[]", with: "240"
+        expect(page).to have_flash_notice "Configuration updated successfully"
+        expect(pg.reload.user_config).to eq({"max_connections" => "240"})
+      end
+
+      it "shows errors when an unknown configuration is provided" do
+        pg.update(user_config: {"max_connections" => "120"})
+        visit "#{project.path}#{pg.path}/config"
+
+        within ".pg-config-card .new-config" do
+          fill_in "pg_config_keys[]", with: "invalid"
+          fill_in "pg_config_values[]", with: "invalid"
+        end
+        click_button "Save"
+
+        expect(page).to have_content "Unknown configuration parameter"
+        expect(page).to have_flash_error "Validation failed for following fields: pg_config.invalid"
+        expect(pg.reload.user_config).to eq({"max_connections" => "120"})
+      end
+
+      it "shows errors when an invalid configuration is provided" do
+        pg.update(user_config: {"max_connections" => "120"})
+        visit "#{project.path}#{pg.path}/config"
+
+        within ".pg-config-card .new-config" do
+          fill_in "pg_config_keys[]", with: "work_mem"
+          fill_in "pg_config_values[]", with: "16iB"
+        end
+        click_button "Save"
+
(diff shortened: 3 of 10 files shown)


assets/js/app.js
```diff
@@ -993,3 +994,129 @@ function flexiblePrecision(value, precision) {
...
+    const configGroup = $(this).closest(".group");
+    configGroup.remove();
+  });
+
+  $(".add-config-btn").on("click", function (event) {
+    event.preventDefault();
+    addConfigRow(event.target);
+  });
+
+  $(".save-config-btn").on("click", async function (event) {
```
jeremyevans: This is a lot of logic to have client-side, and we do not have client-side automated testing. Since this only uses POST, I think you should not use JavaScript for this. The save button should submit the form, and this logic should be server-side.

views/components/page_header.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: (title: @page_title, breadcrumbs: [], left_items: nil, right_items: nil, back: breadcrumbs.dig(-2, 1) || "#") %>
 
-<div class="space-y-1 sticky -mx-10 mb-0.5 px-10 top-[2.9rem] bg-gray-50 z-30">
+<div class="space-y-1 -mx-10 mb-0.5 px-10 top-[2.9rem] bg-gray-50 z-30">
```
jeremyevans: I approve of this as I dislike sticky page headers, but is this related to this PR?

views/components/config_card.erb
```diff
@@ -0,0 +1,87 @@
+<%# locals: (title: nil, config: {}, extra_class: nil, name: nil) %>
```
jeremyevans: Since all users of this template provider values for all keyword arguments, make the keyword arguments required (remove the default values).

views/postgres/config.erb
```diff
@@ -0,0 +1,16 @@
+<% 
+  pg_config = @pg.user_config || {}
+  pgbouncer_config = @pg.pgbouncer_user_config || {}
+%>
+
+<form action="<%= "#{@project_data[:path]}#{@pg.path}/config" %>" role="form" method="POST" class="min-w-2/3 pg-config">
+    <%== csrf_tag("#{@project_data[:path]}#{@pg.path}/config") %>
```
jeremyevans: This looks like 4 space indent instead of the 2 space indent we use in other templates.

views/postgres/config.erb
```diff
@@ -0,0 +1,16 @@
...
+
+    <%== part("components/config_card", title: "PostgreSQL Configuration", config: pg_config, extra_class: "pg-config-card", name: "pg_config") %>
+    <%== part("components/config_card", title: "PgBouncer Configuration", config: pgbouncer_config, extra_class: "pgbouncer-config-card", name: "pgbouncer_config") %>
+    <div class="flex justify-end">
+        <% if @edit_perm %>
+            <%== part("components/form/submit_button", text: "Save", extra_class: "save-config-btn") %>
+        <% end %>
```
jeremyevans:
```suggestion
    <%== part("components/form/submit_button", text: "Save", extra_class: "save-config-btn") if @edit_perm %>
```

views/components/config_card.erb
```diff
@@ -0,0 +1,87 @@
...
+              value="<%= value %>"
+              class="border rounded py-1 focus:ring-blue-500 focus:border-blue-500"
+              name="value"
+            />
+            <span class="error inline-block h-8 w-54 py-2 px-2 max-w-54 text-xs text-red-600 hidden bg-red-100 rounded" aria-live="polite"></span>
+            </div>
+            <div class="min-w-16 flex items-center ml-2">
+              <% if @edit_perm %>
+                <%== part("components/button", type: "danger", text: "", icon: "hero-trash", extra_class: "delete-config-btn") %>
+              <% end %>
```
jeremyevans:
```suggestion
              <%== part("components/button", type: "danger", text: "", icon: "hero-trash", extra_class: "delete-config-btn") if @edit_perm %>
```

views/components/config_card.erb
```diff
@@ -0,0 +1,87 @@
+<%# locals: (title: nil, config: {}, extra_class: nil, name: nil) %>
+
+<div class="bg-white rounded-lg shadow overflow-hidden mb-6 <%= extra_class %>">
+  <div class="px-6 py-4 border-b border-gray-200">
+    <h3 class="text-lg font-medium text-gray-900"><%= title %></h3>
+  </div>
+
+  <div class="divide-y divide-gray-200 w-full max-h-64 overflow-y-auto">
+      <% config.each_with_index do |(key, value), index| %>
```
jeremyevans: This looks like a mix of 2-space and 4-space indent.  Please consistently use 2-space indent.

jeremyevans (changes requested):
In general, I agree with merging the configuration UI. However, unless there is a reason why we must use JavaScript for the form submission, I think we should use a normal form submission, and handle the validation/error reporting logic server-side.

views/components/page_header.erb
shikharbhardwaj: Yes, having `sticky` hides the notification popup which I added to indicate success/failure here.

image

The difference it makes otherwise seems to be minimal, I actually prefer the version without the gap at the top.

With `sticky`             |  Without `sticky`
:-------------------------:|:-------------------------:
  |

assets/js/app.js
shikharbhardwaj: I wanted to minimize the javascript usage too, but there were a couple of challenges that came up:

1. The field names are provided by user input. So the form would need at least some javascript afaik to alter the input field names to match the input values.
2. The UX we are going for here has a WIP -> save flow, with detailed reports on which user inputs failed validation so that they can fix it without having to fill the entire form from scratch again. For eg. imagine a user fills in 10 config values and submits the form, only to be told that 1 of the 10 failed validation and then having to fill everything from scratch again, losing progress. In the current state our validation also reports only field names, which misses detailed feedback on what needs fixing in the input (for example possible enum values.  video).

I had an alternative for this in the previous version of this PR: https://github.com/ubicloud/ubicloud/pull/3579, where we only allowed single key-value changes and used PATCH to submit, although that still used some javascript but the amount of client-side logic was less.

Let me know if you have ideas to achieve this without relying on javascript (or reducing what's here as much as possible).

jeremyevans: Why do the HTML field names need to match the user input?  I would assume both configs can submit the key value pairs as an array, using these HTML input names: `pg_config_keys[]` and `pg_config_values[]`.  You transform it into a hash server-side:

```ruby
  keys = typecast_params.array(:str, "pg_config_keys")
  values = typecast_params.array(:str, "pg_config_values")
  hash = keys.zip(values).to_h
```

Handling errors for form inputs happens on many of our forms. You should be able to get the same error handling you currently have on the clients-side done server-side.  You only need JavaScript error handling if you want to do validation without contacting the server. If you are contacting the server anyway, you should have the server do it.  You can add tests for the case where the 1 out of 10 fields fail validation, and make sure the server is returning the correct error message for the correct field.

shikharbhardwaj: > You should be able to get the same error handling you currently have on the clients-side done server-side.

Do you mean this flash message or something else?
image

I see that with other forms (for eg. the firewall rule form), we do get the errors annotated to fields. Does that use field names to match the errors back? The other form does not seem to do anything special afaics.

jeremyevans: What you are doing to set the `Unknown configuration parameter` and `must be one of: ...` field errors can be done server-side when generating the HTML for the page.  When constructing the HTML, if the user-requested configuration setting has an error, you can use the `border-red-500` error class for the related input, and you can include the error message in the related sibling with the `error` class.

shikharbhardwaj: > field errors can be done server-side when generating the HTML for the page

That is neat. Do you have a small example/pointer on how to access the submitted form/errors from the template? I wasn't able to find similar handling in other forms.

jeremyevans: Assuming you got the parameters as mentioned in https://github.com/ubicloud/ubicloud/pull/3622#discussion_r2208249204, you could have a second hash for errors.  For each input with errors, use the same key, and store the error message in the value. When building the form input in the template, check the error hash using the current key, and if there is an error for the input, use the `border-red-500` class in the input, and populate the sibling with the `error` class with the error message.

shikharbhardwaj: I found a different way which seems to be cleaner, using the errors and old values from the session flash variables. Let me know your thoughts. Right now the form submit uses a minor amount of javascript, in such a way that the form can be submitted without javascript too.

views/components/page_header.erb
shikharbhardwaj: Since the latest version does not need this, I reverted this back to the original

routes/project/location/postgres.rb
```diff
@@ -493,53 +493,53 @@ class Clover
...
         end
 
-        r.post do
+        r.on method: [:post, :patch] do
           authorize("Postgres:edit", pg.id)
 
-          pg_config = typecast_params.Hash("pg_config")
-          pgbouncer_config = typecast_params.Hash("pgbouncer_config")
+          pg_config = typecast_params.Hash("pg_config") || {}
+          pgbouncer_config = typecast_params.Hash("pgbouncer_config") || {}
```
jeremyevans: This can move to an else block of the following `if` statement.

views/postgres/config.erb
```diff
@@ -0,0 +1,27 @@
+<% 
+  pg_config = @pg.user_config || {}
+  pgbouncer_config = @pg.pgbouncer_user_config || {}
+
+  errors = flash.dig("errors") || {}
+  old = flash.dig("old") || {}
```
jeremyevans:
```suggestion
  errors = flash["errors"] || {}
  old = flash["old"] || {}
```

`dig` should be used when you have a multiple arguments.

spec/routes/web/project/postgres_spec.rb
```diff
@@ -811,5 +811,91 @@
...
+        page.driver.submit :post, "#{project.path}#{pg.path}/config",
+          {
+            "pg_config_keys[]" => "max_connections",
+            "pg_config_values[]" => "240",
+            "_csrf" => csrf_token
+          }
+
+        expect(page).to have_field "pg_config_keys[]", with: "max_connections"
+        expect(page).to have_field "pg_config_values[]", with: "240"
+        expect(page).to have_flash_notice "Configuration updated successfully"
```
jeremyevans, line 857: Please add an expectation that the underlying object was updated.

assets/js/app.js
```diff
@@ -993,3 +994,61 @@ function flexiblePrecision(value, precision) {
...
+    event.preventDefault();
```
jeremyevans: We should make it so we don't need a javascript event handler on the Save button.  The Save button should be a standard HTML input, and clicking it should use a standard form submission.  The only javascript needed should be to add/remove rows from the form.

The form should always submit the empty row. When processing the input, you should remove hash entries where the key and value are both `""`.

views/components/config_card.erb
```diff
@@ -0,0 +1,95 @@
...
+    <% end %>
+    <!-- Hidden row for generating new entries -->
+    <div class="px-6 py-2 hover:bg-gray-50 group config-placeholder-group font-mono hidden" data-config-id="">
+      <div class="flex items-center justify-between">
+        <div class="flex flex-wrap items-center config-entry">
+          <input
+            type="text"
+            class="border rounded py-1 focus:ring-blue-500 focus:border-blue-500 mr-2"
+            value=""
+            name="<%= name %>_keys[]"
```
jeremyevans, line 54: For hidden row, I would remove the name from the inputs, and add it via javascript when adding it to the form. This way the hidden row inputs are not submitted with the form.

views/components/config_card.erb
```diff
@@ -0,0 +1,95 @@
...
+    <!-- Empty row for new configuration -->
+    <% if @edit_perm %>
+      <div class="px-6 py-2 hover:bg-gray-50 group new-config font-mono">
+        <div class="flex items-center justify-between">
+          <div class="flex flex-wrap items-center">
+            <input
+              type="text"
+              class="border rounded py-1 focus:ring-blue-500 focus:border-blue-500 mr-2"
+              placeholder="New Key"
+              name="keys_placeholder"
```
jeremyevans: For the empty row, I would use`<%= name %>_keys[]` and  `<%= name %>_values[]` as the input names. This will ensure the empty row is submitted with a normal form submission. This should allow you to eliminate the javascript handler on on the save button.

views/components/config_card.erb
```diff
@@ -0,0 +1,95 @@
+<%# locals: (title:, config:, extra_class:, name:, errors:) %>
+
+<div class="bg-white rounded-lg shadow overflow-hidden mb-6 <%= extra_class %>">
+  <div class="px-6 py-4 border-b border-gray-200">
+    <h3 class="text-lg font-medium text-gray-900"><%= title %></h3>
+  </div>
+
+  <div class="divide-y divide-gray-200 w-full max-h-64 overflow-y-auto">
+    <% config.each_with_index do |(key, value), index| %>
```
jeremyevans, line 9:
```suggestion
    <% config.each_with_index do |(key, value), index| %>
      <% error = errors[key] %>
```

And use `error` instead of `errors[key]` inside the block.

spec/routes/web/project/postgres_spec.rb
```diff
@@ -811,5 +811,91 @@
...
+
+        # Read csrf token from hidden input field
+        csrf_token = find(".pg-config input[name='_csrf']", visible: false).value
+
+        page.driver.submit :post, "#{project.path}#{pg.path}/config",
+          {
+            "pg_config_keys[]" => "max_connections",
+            "pg_config_values[]" => "240",
+            "_csrf" => csrf_token
+          }
```
jeremyevans: Use standard capybara form filling:

```suggestion
        fill_in "pg_config_keys[]", with: "max_connections"
        fill_in "pg_config_values[]", with: "240"
        click_button "Save
```

here and elsewhere. This tests that things work with a normal form submission. If it doesn't, that is an issue that should be fixed. This also tests that the form is usable without JavaScript, even if you can only add one new config value at a time.

byucesoy (commented):
It looks good from Postgres point of view. I'll leave final approve to @jeremyevans since he is already doing a detailed review.

assets/js/app.js
shikharbhardwaj: This makes sense. For the removal, do you propose allowing empty keys? I think we should remove everything with empty keys or error out if the key is empty but the corresponding value is not.

views/components/config_card.erb
shikharbhardwaj: Right now they don't get submitted with the form, as they are `disabled`.

views/components/config_card.erb
shikharbhardwaj: Doing this introduces a side effect in the error handling path: the returned form has an extra empty row. If the user corrects their mistake, they would have to delete this empty row to submit the change. I can remove the `required` attribute from the key input, but it looks off to introduce a third row when the user only submitted two. wdyt?

image

assets/js/app.js
jeremyevans: If the key and value are both empty, we should remove the hash entry server side. If the either is not empty, we should not ignore it.

views/components/config_card.erb
jeremyevans: I would remove `required` from the key input. If they submit with empty key and value, ignore that row.  You should be able to fix the display to not display empty rows ( something like `next if key.empty? && value.empty?` near the top of the loop)

shikharbhardwaj: Sounds good, skipping the row makes it work fine, even with `required` on the existing config rows. I think we should leave it this way to avoid empty key submits.

assets/js/app.js
shikharbhardwaj: I updated the validator to take care of this

image

routes/project/location/postgres.rb
```diff
@@ -493,53 +493,55 @@ class Clover
...
+        r.on method: [:post, :patch] do
           authorize("Postgres:edit", pg.id)
 
-          pg_config = typecast_params.Hash("pg_config")
-          pgbouncer_config = typecast_params.Hash("pgbouncer_config")
+          if web?
+            pg_keys = typecast_params.array(:str, "pg_config_keys") || []
+            pg_values = typecast_params.array(:str, "pg_config_values") || []
+            pg_config = pg_keys.zip(pg_values).to_h.compact
+            pg_config.delete_if { |key, _value| key.empty? && _value.empty? }
```
jeremyevans: If `pg_config_keys` has more elements than `pg_config_values`, the value of the hash will be nil, and this line will result in a NoMethodError.  While that cannot happen in normal use of the form, we try to make all of our routes safe from unexpected input. You can either `pg_config.transform_values!(&:to_s)`, or you can change the `delete_if` to use `_value.nil? || (key.empty? && _value.empty?)`.

I also recommend you use `value` instead of `_value` for the block argument. Only use an underscore prefix if you will be ignoring the parameter.

Same changes needed in `pgbouncer_config`.

You may want to add a helper method for this, so you aren't duplicating the logic.

routes/project/location/postgres.rb
```diff
@@ -493,53 +493,53 @@ class Clover
...
           pg_validator = Validation::PostgresConfigValidator.new(pg.version)
-          pg_validator.validate(pg_config)
+          pg_errors = pg_validator.validation_errors(pg_config)
 
           pgbouncer_validator = Validation::PostgresConfigValidator.new("pgbouncer")
-          pgbouncer_validator.validate(pgbouncer_config)
+          pgbouncer_errors = pgbouncer_validator.validation_errors(pgbouncer_config)
+
+          if pg_errors.any? || pgbouncer_errors.any?
+            raise Validation::ValidationFailed.new(pg_errors.merge(pgbouncer_errors))
```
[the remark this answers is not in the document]
jeremyevans: I'm not sure if there is any overlap between valid keys.  If so, I agree we should address this issue in some way.

routes/project/location/postgres.rb
shikharbhardwaj: Done

routes/project/location/postgres.rb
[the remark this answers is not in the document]
shikharbhardwaj: I checked, and there are a few keys in common. I added the card name as a prefix to disambiguate between them.

routes/project/location/postgres.rb
```diff
@@ -493,55 +493,63 @@ class Clover
...
+            r.redirect "#{@project.path}#{pg.path}/config"
+          end
         end
       end
     end
   end
+
+  def config_hash_from_kvs(keys, values)
+    hash = keys.zip(values).to_h.compact
+    hash.delete_if { |key, value| value.nil? || (key.empty? && value.empty?) }
```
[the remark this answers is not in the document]
jeremyevans: I don't think this is a requirement, but it may be worth adding. If we did want to do it, it would be better to do:

```ruby
hash = keys.map(&:strip).zip(values.map(&:strip)).to_h.compact
```

Also, I missed the `compact` in my earlier review.  That will remove entries from the hash where the value is nil, so you don't need the `value.nil?` check in the `delete_if`.

routes/project/location/postgres.rb
```diff
@@ -493,55 +493,63 @@ class Clover
...
+          else
+            flash["notice"] = "Configuration updated successfully"
+            r.redirect "#{@project.path}#{pg.path}/config"
+          end
         end
       end
     end
   end
+
+  def config_hash_from_kvs(keys, values)
```
jeremyevans: Please move this from the route file to the `helpers/web.rb` file.

jeremyevans (approved):
Looks good. One more change requested, but this looks good to merge after that change.

PR #3628, opened 2025-07-15
Fix `persistent_device_id` for NVMe disks

pykello:
NVMe disk Ids start with "nvme-eui." not "nmve-eui-".

For example:

```
nvme-eui.3634473057c004720025384e00000001
```


jeremyevans (approved):
Looking at my `/dev/disk/by-id` directory, it has `nvme-eui.`, so I assume this is correct.

PR #3632, opened 2025-07-16
Add config value for db pool size for monitor process

velioglu:
Since we want to update pool size for monitor process without updating pool size for other processes, adding a new config value and use it while initiating the monitor process.


db.rb
```diff
@@ -7,7 +7,11 @@
 
 db_ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", "db_ca_bundle.crt")
 Util.safe_write_to_file(db_ca_bundle_filename, Config.clover_database_root_certs)
-max_connections = Config.db_pool - 1
+max_connections = if ENV["DYNO"]&.include?("monitor")
```
jeremyevans, line 10: DYNO probably only works on Heroku, and we want something that also works in other environments. Please have bin/monitor set a constant before it requires anything, and check here if the constant is defined.

PR #3636, opened 2025-07-18
CSI E2E tests

mohi-kalantari:
A very minimal test prog is added to test the cluster bootstrapping

It will attempt to create a cluster, run a kubectl get node command and destroy the cluster.

### DIFF
--- a/config/e2e_test_cases.yml
+++ b/config/e2e_test_cases.yml
@@ -3,3 +3,4 @@
 - { name: github_runner_ubuntu_2204, images: ["github-ubuntu-2204"], details: {repo_name: ubicloud/github-e2e-test-workflows, workflow_name: test_2204.yml, branch_name: main} }
 - { name: postgres_standard,         images: ["postgres17-ubuntu-2204"] }
 - { name: postgres_ha,               images: ["postgres17-ubuntu-2204", "ubuntu-jammy"] }
+- { name: kubernetes,                images: ["kubernetes-v1_33"]}
--- a/prog/test/kubernetes.rb
+++ b/prog/test/kubernetes.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+require_relative "../../lib/util"
+
+class Prog::Test::Kubernetes < Prog::Test::Base
+  semaphore :destroy
+
+  def self.assemble
+    kubernetes_test_project = Project.create(name: "Kubernetes-Test-Project")
+    kubernetes_service_project = Project[Config.kubernetes_service_project_id] ||
+      Project.create(name: "Ubicloud-Kubernetes-Resources") do |project|
+        project.id = Config.kubernetes_service_project_id
+      end
+
+    frame = {
+      "kubernetes_service_project_id" => kubernetes_service_project.id,
+      "kubernetes_test_project_id" => kubernetes_test_project.id
+    }
+
+    Strand.create(
+      prog: "Test::Kubernetes",
+      label: "start",
+      stack: [frame]
+    )
+  end
+
+  label def start
+    st = Prog::Kubernetes::KubernetesClusterNexus.assemble(
+      name: "kubernetes-test-standard",
+      project_id: frame["kubernetes_test_project_id"],
+      location_id: Location::HETZNER_FSN1_ID,
+      version: "v1.33",
+      cp_node_count: 1
+    )
+
+    update_stack({"kubernetes_cluster_id" => st.id})
+    hop_wait_kubernetes_resource
+  end
+
+  label def wait_kubernetes_resource
+    if kubernetes_cluster.strand.label == "wait"
+      hop_test_kubernetes
+    else
+      nap 10
+    end
+  end
+
+  label def test_kubernetes
+    unless kubernetes_cluster.client.kubectl("get nodes")
+      update_stack({"fail_message" => "Failed to run test kubectl command"})
+    end
+
+    hop_destroy_kubernetes
+  end
+
+  label def destroy_kubernetes
+    kubernetes_cluster.incr_destroy
+    hop_destroy
+  end
+
+  label def destroy
+    kubernetes_test_project.destroy
+
+    fail_test(frame["fail_message"]) if frame["fail_message"]
+
+    pop "Kubernetes tests are finished!"
+  end
+
+  def kubernetes_test_project
+    @kubernetes_test_project ||= Project[frame["kubernetes_test_project_id"]]
+  end
+
+  def kubernetes_cluster
+    @kubernetes_cluster ||= KubernetesCluster[frame["kubernetes_cluster_id"]]
+  end
+end
--- a/bin/e2e
+++ b/bin/e2e
@@ -47,6 +47,10 @@ def main(options)
     tests_to_wait << Prog::Test::HaPostgresResource.assemble
   end
 
+  if options[:test_cases].include?("kubernetes")
+    tests_to_wait << Prog::Test::Kubernetes.assemble
+  end
+
   # Although wait_until will be blocked while checking the first one
   # it won't affect the total time as other strands will continue in parallel.
   # No need to make it parallel.


prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -224,7 +224,7 @@ def before_run
       decr_destroy
 
       if (services_lb = kubernetes_cluster.services_lb)
-        services_lb.dns_zone.delete_record(record_name: "*.#{services_lb.hostname}.")
+        services_lb.dns_zone.delete_record(record_name: "*.#{services_lb.hostname}.") if services_lb.dns_zone
```
jeremyevans: Simpler to use the lonely operator:

```suggestion
        services_lb.dns_zone&.delete_record(record_name: "*.#{services_lb.hostname}.")
```

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,94 @@
...
+      }]
+    )
+  end
+
+  label def start
+    st = Prog::Kubernetes::KubernetesClusterNexus.assemble(
+      name: "kubernetes-test-standard",
+      project_id: frame["kubernetes_test_project_id"],
+      location_id: Location::HETZNER_FSN1_ID,
+      version: Option.kubernetes_versions.last,
```
jeremyevans: Looks like we are testing the oldest version instead of the newest? That looks deliberate as we are using 1.33 boot image. Seems like we would typically want to test the newest supported version if we are only testing one version, but not a big deal.

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,94 @@
...
+      cp_node_count: 1
+    )
+    Prog::Kubernetes::KubernetesNodepoolNexus.assemble(
+      name: "kubernetes-test-standard-nodepool",
+      node_count: 1,
+      kubernetes_cluster_id: st.id,
```
jeremyevans: This works because `st` has the same id as its subject, but looks odd.  Consider doing `kc = st.subject` and using the `kc` variable instead of the `st` variable.

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    update_stack({"kubernetes_cluster_id" => st.id})
+    hop_update_loadbalancer_hostname
+  end
+
+  def ensure_hosts_entry(sshable, api_hostname)
+    host_line = "#{kubernetes_cluster.sshable.host} #{api_hostname}"
+    output = sshable.cmd("cat /etc/hosts")
+    unless output.include?(host_line)
+      sshable.cmd("echo '#{host_line}' | sudo tee -a /etc/hosts > /dev/null")
```
jeremyevans: To be safer, I would do:

```suggestion
      sshable.cmd("echo #{host_line.shellescape} | sudo tee -a /etc/hosts > /dev/null")
```

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,94 @@
...
+    hop_wait_kubernetes_resource
+  end
+
+  label def wait_kubernetes_resource
+    if kubernetes_cluster.strand.label == "wait"
+      hop_test_kubernetes
+    else
+      (kubernetes_cluster.cp_vms + kubernetes_cluster.worker_vms).each do |vm|
+        ensure_hosts_entry(vm.sshable, kubernetes_cluster.api_server_lb.hostname)
+      end
```
jeremyevans: Do we potentially want to run this more than once? Doesn't look harmful to run more than once, but once it has been run once successfully, I'm not sure why we would nap and potentially run it again, since there seem no advantages to doing so. Running it more than once is slower.

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,94 @@
...
+      nap 10
+    end
+  end
+
+  label def test_kubernetes
+    unless kubernetes_cluster.client.kubectl("get nodes")
```
jeremyevans: Do we not care about the output in this case, only the return value (I assume the method returns truthy for 0, and falsely for other values)?

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,94 @@
...
+  label def destroy
+    kubernetes_test_project.destroy
+
+    fail_test(frame["fail_message"]) if frame["fail_message"]
+
+    pop "Kubernetes tests are finished!"
+  end
+
+  def kubernetes_test_project
+    @kubernetes_test_project ||= Project[frame["kubernetes_test_project_id"]]
```
jeremyevans: If this should never return `nil`, you can consider changing from `Project.[]` to `Project.with_pk!`. Same for KubernetesCluster a few lines down.

prog/test/kubernetes.rb
mohi-kalantari: I had upgrades in mind and we should use the last image in order to have a version to upgrade to. But that's optimization for later. I'll change it to `first`

prog/test/kubernetes.rb
mohi-kalantari: It raises a runtime error if it has an issue. I'll add a max retry count for better error detection.

config/e2e_test_cases.yml
```diff
@@ -3,3 +3,4 @@
 - { name: github_runner_ubuntu_2204, images: ["github-ubuntu-2204"], details: {repo_name: ubicloud/github-e2e-test-workflows, workflow_name: test_2204.yml, branch_name: main} }
 - { name: postgres_standard,         images: ["postgres17-ubuntu-2204"] }
 - { name: postgres_ha,               images: ["postgres17-ubuntu-2204", "ubuntu-jammy"] }
+- { name: kubernetes,                images: ["kubernetes-v1_33"]}
```
aamederen: let's switch to 1.34

aamederen: but if we are testing upgrades as well, let's keep it with a comment

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,94 @@
...
+      node_count: 1,
+      kubernetes_cluster_id: st.id,
+      target_node_size: "standard-2"
+    )
+
+    update_stack({"kubernetes_cluster_id" => st.id})
+    hop_update_loadbalancer_hostname
+  end
+
+  def ensure_hosts_entry(sshable, api_hostname)
```
aamederen: for code organization, it could be nicer to keep FSM labels and other functions separate

enescakir:
You need to add `kubernetes` to `test_cases` input if you want to run them by default when triggered `/run-e2e`

```yml
    inputs:
      test_cases:
        description: 'Comma-separated list of test cases (vm)'
        required: true
        default: 'vm,github_runner_ubuntu_2404,github_runner_ubuntu_2204,postgres_standard,postgres_ha'
        type: string
```

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,118 @@
...
+    )
+    kc = st.subject
+    Prog::Kubernetes::KubernetesNodepoolNexus.assemble(
+      name: "kubernetes-test-standard-nodepool",
+      node_count: 1,
+      kubernetes_cluster_id: kc.id,
+      target_node_size: "standard-2"
+    )
+
+    update_stack({"kubernetes_cluster_id" => st.id})
```
enescakir: nit:
```suggestion
    kc = Prog::Kubernetes::KubernetesClusterNexus.assemble(
      name: "kubernetes-test-standard",
      project_id: frame["kubernetes_test_project_id"],
      location_id: Location::HETZNER_FSN1_ID,
      version: Option.kubernetes_versions.first,
      cp_node_count: 1
    ).subject
    Prog::Kubernetes::KubernetesNodepoolNexus.assemble(
      name: "kubernetes-test-standard-nodepool",
      node_count: 1,
      kubernetes_cluster_id: kc.id,
      target_node_size: "standard-2"
    )

    update_stack({"kubernetes_cluster_id" => kc.id})
```

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,117 @@
...
+      target_node_size: "standard-2"
+    )
+
+    update_stack({"kubernetes_cluster_id" => kc.id})
+    hop_update_loadbalancer_hostname
+  end
+
+  label def update_loadbalancer_hostname
+    nap 5 unless kubernetes_cluster.api_server_lb
+    kubernetes_cluster.api_server_lb.update(custom_hostname: "k8s-e2e-test.ubicloud.test")
```
aamederen, line 43: why do we need these changes? Is it something related to the lack of DNS server/zone in E2E test suite?

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  label def test_kubernetes
+    begin
+      nodes_output = kubernetes_cluster.client.kubectl("get nodes")
+    rescue RuntimeError => ex
+      update_stack({"fail_message" => "Failed to run test kubectl command: #{ex.message}"})
+      hop_destroy_kubernetes
+    end
+    kubernetes_cluster.all_nodes.each { |node|
+      update_stack({"fail_message" => "node #{node.name} not found in cluster nodes"}) unless nodes_output.include?(node.name)
```
aamederen: minor, but this doesn't look like adding all missing nodes to the stack. Instead, fail_message would have the message for the last missing node.

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,117 @@
...
+    kubernetes_cluster.incr_destroy
+    hop_destroy
+  end
+
+  label def destroy
+    kubernetes_test_project.destroy
+
+    fail_test(frame["fail_message"]) if frame["fail_message"]
+
+    pop "Kubernetes tests are finished!"
```
aamederen, line 92: IMO we should wait for kc destroy to finish as well, in order to catch any issues in the destroy logic.

prog/test/kubernetes.rb
mohi-kalantari: If we don't set it, the lb.hostname returns an invalid dns address which breaks the dns resolve. we need a valid domain for it

aamederen: alright, sounds like a larger issue than this specific test. We can consider it OK for now and revisit the DNS issue later.

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,120 @@
...
+    cp_vm = kubernetes_cluster.cp_vms.first
+    nap 5 unless vm_ready?(cp_vm)
+    ensure_hosts_entry(cp_vm.sshable, kubernetes_cluster.api_server_lb.hostname)
+    hop_update_worker_hosts_entries
+  end
+
+  label def update_worker_hosts_entries
+    vm = kubernetes_cluster.nodepools.first.vms.first
+    nap 5 unless vm_ready?(vm)
+    ensure_hosts_entry(vm.sshable, kubernetes_cluster.api_server_lb.hostname)
```
jeremyevans, line 57: It appears this switched from ensuring the hosts entry exists on all cp and nodepool vms to ensuring it exists on a single cp vm and since nodepool vm. Is it not necessary to ensure the hosts entry is updated on all of them?

prog/test/kubernetes.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  def vm_ready?(vm)
+    return false unless vm
+    begin
+      vm.sshable.cmd("uptime")
+    rescue
+      return false
+    end
+    true
```
jeremyevans: Slightly simpler:

```suggestion
    return false unless vm
    vm.sshable.cmd("uptime")
    true
  rescue
    false
```

prog/test/kubernetes.rb
mohi-kalantari: Right now we are testing in a 2 node cluster and the current program is adding the hosts entries to all the vms so we are fine.

To fully test all the functions we would need to run upgrades for examples in bigger clusters. This is an area for improvement when we are adding those tests.

PR #3648, opened 2025-07-22
Premium runner allocation improvements

enescakir:
- **Do not downgrade premium runner that was explicitly requested**
  If the toggle on the UI is enabled, even a runner requested by the
  customer with a premium label might be downgraded to a standard runner
  because the toggle sets the installation's "family_filter" to
  `["standard", "premium"]`
  

- **Check premium utilization for the premium concurrency check as well**
  Currently, we check the overall utilization to be below 75% to allow
  exceeding the soft concurrency limit. However, this check is global and
  doesn’t differentiate between standard and premium runners.
  Consequently, if the standard utilization is low, but a single customer
  with premium runners requests a large number of vCPUs, they could
  consume all available premium capacity. As a result, other customers
  with premium access might be blocked from getting premium runners.
  
  If both standard and premium utilization are above 75%, we don’t allow
  exceeding the soft concurrency limit.
  
  If premium utilization is high, but standard utilization is low, we
  allow exceeding the soft concurrency limit without a premium upgrade.
  
  To accommodate requested premium customers, I reduced the limit to 50%
  for free upgrades.
  

----

> [!IMPORTANT]
> Improves premium runner allocation by preventing downgrades and refining concurrency checks in `prog/vm/github_runner.rb` and `prog/vm/nexus.rb`.
> 
>   - **Behavior**:
>     - Prevents downgrading explicitly requested premium runners by adjusting `family_filter` in `prog/vm/nexus.rb`.
>     - Adjusts concurrency checks in `wait_concurrency_limit` in `prog/vm/github_runner.rb` to consider both standard and premium utilizations separately.
>     - Reduces limit for free premium upgrades to 50% utilization.
>   - **Code Changes**:
>     - Adds `:not_upgrade_premium` to `SemaphoreMethods` in `model/github_runner.rb`.
>     - Modifies `wait_concurrency_limit` in `prog/vm/github_runner.rb` to handle premium and standard utilization checks.
>     - Updates `before_run` and `start` methods in `prog/vm/nexus.rb` to respect premium runner requests.
>   - **Tests**:
>     - Adds tests in `spec/prog/vm/github_runner_spec.rb` for new concurrency logic and premium runner handling.
>     - Updates `spec/prog/vm/nexus_spec.rb` to test allocation logic respecting premium runner requests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 211aa45f8b5051b0d0ab02a52242f4fca1a471c0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -166,16 +166,26 @@ class Prog::Vm::GithubRunner < Prog::Base
   label def wait_concurrency_limit
     unless quota_available?
       # check utilization, if it's high, wait for it to go down
-      utilization = VmHost.where(allocation_state: "accepting", arch: github_runner.label_data["arch"]).select_map {
-        sum(:used_cores) * 100.0 / sum(:total_cores)
-      }.first.to_f
+      family_utilization = VmHost.where(allocation_state: "accepting", arch: github_runner.label_data["arch"])
+        .select_group(:family)
+        .select_append { round(sum(:used_cores) * 100.0 / sum(:total_cores), 2).cast(:float).as(:utilization) }
+        .to_hash(:family, :utilization)
 
-      unless utilization < 75
-        Clog.emit("Waiting for customer concurrency limit, utilization is high") { [github_runner, {utilization: utilization}] }
+      not_allow = if github_runner.label_data["arch"] == "x64" && github_runner.label_data["family"] == "standard" && github_runner.installation.premium_runner_enabled?
+        family_utilization["premium"] > 75 && family_utilization["standard"] > 75
+      else
+        family_utilization.fetch(github_runner.label_data["family"], 0) > 75
+      end
+
+      if not_allow
+        Clog.emit("not allowed because of high utilization") { {reached_concurrency_limit: {family_utilization:, label: github_runner.label, repository_name: github_runner.repository_name}} }
         nap rand(5..15)
       end
 
-      Clog.emit("Concurrency limit reached but allocation is allowed because of low utilization") { [github_runner, {utilization: utilization}] }
+      if github_runner.label_data["arch"] == "x64" && ((family_utilization["premium"] > 75) || (github_runner.installation.free_runner_upgrade? && family_utilization["premium"] > 50))
+        github_runner.incr_not_upgrade_premium
+      end
+      Clog.emit("allowed because of low utilization") { {exceeded_concurrency_limit: {family_utilization:, label: github_runner.label, repository_name: github_runner.repository_name}} }
     end
     github_runner.log_duration("runner_capacity_waited", Time.now - github_runner.created_at)
     hop_allocate_vm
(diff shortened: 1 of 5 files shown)


prog/vm/github_runner.rb
```diff
@@ -171,16 +171,25 @@ def quota_available?
   label def wait_concurrency_limit
     unless quota_available?
       # check utilization, if it's high, wait for it to go down
-      utilization = VmHost.where(allocation_state: "accepting", arch: github_runner.label_data["arch"]).select_map {
-        sum(:used_cores) * 100.0 / sum(:total_cores)
-      }.first.to_f
+      utils = VmHost.where(allocation_state: "accepting", arch: "x64").select {
```
jeremyevans: Here we are always checking `x64` arch, even if the runner is for `arm64` arch. Does that make sense (maybe it does, or maybe this is only reachable for `x64`, but it looks suspect, so I thought I'd ask)?

Up to you, but you may want to use:

```ruby
select_group(:family).
  select_append { round(...).as(:utilization) }.
  to_hash(:family, :utilization)
```

You probably should at least change the `to_h { ... }` to `to_hash(:family, :utilization)`

enescakir: Good catch! `x64` came from my local tests.

enescakir: Thanks for the suggestions of `select_group` and `select_append`. It’s become more readable now.

prog/vm/github_runner.rb
```diff
@@ -166,16 +166,26 @@ def quota_available?
...
-      }.first.to_f
+      utils = VmHost.where(allocation_state: "accepting", arch: github_runner.label_data["arch"])
+        .select_group(:family)
+        .select_append { round(sum(:used_cores) * 100.0 / sum(:total_cores), 2).cast(:float).as(:utilization) }
+        .to_hash(:family, :utilization)
 
-      unless utilization < 75
-        Clog.emit("Waiting for customer concurrency limit, utilization is high") { [github_runner, {utilization: utilization}] }
+      not_allow = if github_runner.label_data["arch"] == "x64" && github_runner.label_data["family"] == "standard" && github_runner.installation.premium_runner_enabled?
+        utils["premium"] > 75 && utils["standard"] > 75
```
bsatzger: if there is either no `premium` or no `standard` host in `accepting` of the requested `arch`, then this would compare `nil` to 75 and fail.

bsatzger: maybe use `utils` with `fetch` and a default value (0 or 100)?

bsatzger: maybe also rename utils to utilizations or family_utilization. utils sounds more like utilities.

bsatzger (approved):

prog/vm/github_runner.rb
enescakir: We shouldn't have this kind of issue in production since we need some hosts for each available runner label. However, it makes sense to use `.fetch(family, 0)` to be on the safe side. I also renamed it. Thanks for the review.

enescakir: I opted for `0` instead of `100` because if we don’t have that kind of machine, I prefer to stuck on `allocate_vm` instead of `wait_concurrency`.

PR #3656, opened 2025-07-23
Fix old backup and wal files delete policy

furkansahin:
We were only cleaning up the basebackups and forget about the wal files. With this commit, we start cleaning up the wal files, too.

----

> [!IMPORTANT]
> Update lifecycle policy in `PostgresTimeline` to clean up wal files by modifying `set_lifecycle_policy` to use `filter` instead of `prefix`.
> 
>   - **Behavior**:
>     - Updates `set_lifecycle_policy` in `postgres_timeline.rb` to remove `prefix` and add `filter` for lifecycle rules, ensuring wal files are also cleaned up.
>   - **Tests**:
>     - Updates test in `postgres_timeline_spec.rb` to expect lifecycle configuration without `prefix` and with `filter`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 6a91330cf72b207093a20ddf2cb64cc10ffe4edf. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_timeline.rb
+++ b/model/postgres/postgres_timeline.rb
@@ -149,11 +149,11 @@ PGHOST=/var/run/postgresql
           rules: [
             {
               id: "DeleteOldBackups",
-              prefix: "basebackups_005/",
               status: "Enabled",
               expiration: {
                 days: BACKUP_BUCKET_EXPIRATION_DAYS
-              }
+              },
+              filter: {}
             }
           ]
         }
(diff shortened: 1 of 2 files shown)


model/postgres/postgres_timeline.rb
```diff
@@ -154,6 +154,14 @@ def set_lifecycle_policy
               expiration: {
                 days: BACKUP_BUCKET_EXPIRATION_DAYS
               }
+            },
```
byucesoy: Or, let's not add a prefix rule at all. In UI I can add lifecycle configuration that can be applied to the entire bucket without filtering. I assume it is possible via API as well.

furkansahin: makes sense, updated.

byucesoy (approved):

PR #3657, opened 2025-07-23
Support monitor partitioning

jeremyevans:
This adds respirate-style partitioning to monitor, and refactors monitor.

With this, partitioning works the same way for monitor as it does for respirate. There is a repartition thread that listens on a channel for information about active partitions, and notifies about the current process so that other monitor processes are aware of its existance.

As monitor is less time sensitive than respirate, only scanning for new resources once a minute instead of once a second, the repartition thread doesn't notify as often. Repartitioning for new monitor processes still happens immediately, before the next enqueue (within 5 seconds). Removing stale monitor processes happens within 1 minute.

In adding to adding partitioning, this refactors monitor significantly:

* This abstracts the monitor and metric export handling into a single Struct class.  Almost all of the logic was shared, and this removes the duplication.

* This avoids the separation of the scan thread from the main/enqueuing thread. Instead, the main/enqueing thread just scans if it has been more than a minute since the last scan, or if it was repartitioned since the last enqueue.  This removes the need for a mutex to protect the hash of resources.

* Previously, all resources that were initially monitored continued to be monitored until they were deleted. That doesn't work when repartitioning.  If the scan query does not pick up the resource, either because the resource was deleted or is no longer in the current partition, then it is no longer monitored.  It will keep the same resource previously used if it still exists, ensuring that the same SSH session is still used.

* Monitor now traps TERM (and INT), and generally exits cleanly within about 1 second on shutdown unless worker threads are busy. It gives worker threads up to 2 seconds to exit, and if they don't exit within 2 second, the process exits with status 1, signifying failure.  If the process dies due to pulse checking or resource scanning failure, it exits with status 2.

This keeps the behavior of enqueuing locked resources, but I think that behavior is questionable and should probably be removed.  It's better to skip a locked resource in the current loop, and enqueue it in a later loop after it is no longer locked.

In general, the enqueuing approach is questionable. We enqueue every monitored resource every 5 seconds, regardless of the current status.  If it takes 4 seconds to run the job, then it is enqueued again 1 second after finishing. We should probably have each resource keep track of when monitoring finished, and only enqueue 5 seconds after the resource last completed its action. This may result in increased overall time between monitoring in some cases, but it can result in decreased overall time between monitoring in the case where the resource was locked, assuming that we look at existing resources every second, and only enqueue those that are not locked and finished over 5 seconds ago.

The monitor process is now extensively documented. Hopefully, it should be easier to understand.

We currently lack a test suite for monitor, which makes a change like this risky. We should consider adding a test suite. We would need the equivalent of Prog::Test for monitor. Then we could have the equivalent of the respirate_smoke_test for monitor.

----

> [!IMPORTANT]
> Adds partitioning to monitor process, refactors codebase, and introduces CI and smoke tests for monitor functionality.
> 
>   - **Behavior**:
>     - Adds partitioning to monitor similar to respirate, using `MonitorRepartitioner` for partition management.
>     - Monitor process scans for new resources every minute and handles repartitioning within 5 seconds.
>     - Traps TERM and INT signals for clean shutdown, allowing 2 seconds for worker threads to exit.
>   - **Refactoring**:
>     - Abstracts monitor and metric export handling into `MonitorResourceType` to reduce duplication.
>     - Removes separate scan thread; main thread handles scanning and enqueuing.
>     - Updates resource monitoring logic to stop monitoring resources not in the current partition.
>   - **Testing and CI**:
>     - Adds `monitor-ci.yml` for GitHub Actions to run monitor smoke tests.
>     - Introduces `spec/monitor_smoke_test.rb` for testing monitor partitioning and functionality.
>   - **Misc**:
>     - Updates `Gemfile` to use a specific Sequel version from GitHub.
>     - Adds `monitor_smoke_test` task to `Rakefile`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 86952e52df29af5cab927ae96597da52f0b9f5b0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bin/monitor
+++ b/bin/monitor
@@ -17,42 +17,8 @@ if partition_number
   raise "invalid partition_number: #{partition_number}" if partition_number < 1
 end
 
-# Used for NOTIFY, since NOTIFY payload must be a string
-partition_number_string = partition_number.to_s
-
-# Assume when starting that we are the final partition. For cases where we aren't,
-# this will quickly be updated after startup.
-num_partitions = partition_number
-
-# Flag set when we have repartitioned, to ensure we do a scan using the new partition
-# before enqueuing additional resources.
-repartitioned = true
-
-# Flag set when shutting down. All threads check this in their loops, and exit within
-# 1 second after it is set.
-shutdown = false
-
-# Time after which to run the scan query to check for new resources.
-scan_after = Time.now
-
-# Time after which to report the number of active threads and other metric information.
-report_after = Time.now + 5
-
-# Time after which to report the number of active threads.
-check_stuck_pulses_after = Time.now + 5
-
-# The number of seconds until we should run the next scan query. This runs a scan
-# every minute.
-scan_every = 60
-
-# The number of seconds between reporting monitor metrics.
-report_every = 5
-
-# The number of seconds after a monitor job completes before resubmitting it.
-enqueue_every = 5
-
-# The number of seconds between checking for for stuck pulses.
-check_stuck_pulses_every = 5
+require_relative "../loader"
+clover_freeze
 
 # Information (seconds, log message, log key) for stuck pulses for monitor jobs.
 monitor_pulse_info = [120, "Pulse check has stuck.", :pulse_check_stuck].freeze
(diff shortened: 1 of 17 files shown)


bin/monitor
```diff
@@ -3,99 +3,303 @@
...
+partition_number = ARGV[0]
+partition_number ||= if (match = /monitor\.(\d+)\z/.match(ENV["DYNO"] || ENV["PS"]))
+  match[1] # Heroku/Foreman
+end
+
+# For simplicity, monitor always runs partitioned. Even if only running a single
+# process, in which case, the partition is the entire id space.
+partition_number ||= "1"
+
+if partition_number
```
byucesoy, line 15: At this point, partition_number is definitely set to something, right? Do we need the `if` check?

bin/monitor
```diff
@@ -3,99 +3,303 @@
...
-      r.lock_no_wait do
-        r.open_resource_session
-        r.process_event_loop
-        r.check_pulse
-      end
-    end
+# Need to define the MonitorResourceType constant before freezing
+clover_freeze
+
+partition_boundary = lambda do |partition_num, partition_size|
```
byucesoy: It seems that there are lots of common logic between this and Dispatcher's partitioning. Should we move some of the functions/procs to `Util`?

bin/monitor
```diff
@@ -3,99 +3,303 @@
...
+    end
+  end
+
+  # Continuouly LISTENs for notifications on the monitor channel until shutdown.
+  # If notified about a higher partition number than the currently expected
+  # partitioning, repartition the current process to decrease the partition size.
+  DB.listen(:monitor, loop:, after_listen: notify_partition, timeout: listen_timeout) do |_, _, payload|
+    throw :stop if shutdown
+
+    unless (partition_num = Integer(payload, exception: false)) && (partition_num <= 8)
```
byucesoy: This might be personal taste, but in my mind partition_num can refer either number of the current partition (i.e. partition index) or number of total partitions (i.e. partition count). In this context, it is the latter, so maybe we can use something like `new_partition_count`?

bin/monitor
```diff
@@ -3,99 +3,303 @@
...
+  loop = proc do
+    if (max_partition = repartition_check.call(partition_times))&.<(num_partitions)
+      repartition.call(max_partition)
+    end
+  end
+
+  # Continuouly LISTENs for notifications on the monitor channel until shutdown.
+  # If notified about a higher partition number than the currently expected
+  # partitioning, repartition the current process to decrease the partition size.
+  DB.listen(:monitor, loop:, after_listen: notify_partition, timeout: listen_timeout) do |_, _, payload|
```
byucesoy: What is the benefit of using proc loop instead of passing `loop: true` and adding whatever logic we want to do in the loop at the end of the block? In this way, it is slightly more difficult to follow through the code because you need to move back (visually) to check what is going to happen.

byucesoy (approved):
Looks good to me. I added bunch of questions to understand it better.

> In general, the enqueuing approach is questionable. We enqueue every monitored resource every 5 seconds, regardless of the current status. If it takes 4 seconds to run the job, then it is enqueued again 1 second after finishing. We should probably have each resource keep track of when monitoring finished, and only enqueue 5 seconds after the resource last completed its action. This may result in increased overall time between monitoring in some cases, but it can result in decreased overall time between monitoring in the case where the resource was locked, assuming that we look at existing resources every second, and only enqueue those that are not locked and finished over 5 seconds ago.

We introduced monitor to check the health status of Postgres resources (and trigger failover if necessary). Enqueing every 5 seconds regardless of the locking status gives tighter upper bound for unavailability detection. If we change the behavior, it would take longer of us to detect the unavailability.

bin/monitor
jeremyevans: Correct, we don't need the if statement.  This was copied over from respirate, which does not always run partitioned.

bin/monitor
jeremyevans: Sure, I can update the variable name.

bin/monitor
jeremyevans: It's true that much of the logic is duplicated, but it isn't exactly the same. I'm not sure it's worth trying to abstract it, unless we plan to support partitioning for additional programs, but it is something I can work on if you think it is worth it. However, it would probably be best to do so after merging, or to extract it from respirate first, rather than doing it in the same commit.

bin/monitor
jeremyevans: The `listen` block is only called for notifications. `loop` is for running code as part of the event loop even if you haven't received a notification. This is important, because you want to check for stale partitions as well as notify about other partitions on a set frequency, even if you haven't received a notification from another server.

jeremyevans:
> We introduced monitor to check the health status of Postgres resources (and trigger failover if necessary). Enqueing every 5 seconds regardless of the locking status gives tighter upper bound for unavailability detection. If we change the behavior, it would take longer of us to detect the unavailability.

I don't think the minor frequency difference will matter. Currently, we try to enqueue every 5 seconds, and if a resource is already in use, it just wastes time as it needs to report the resource is locked (we enqueue it anyway, even if it is locked, just to skip it again if it is still locked!).

At the very most, the approach I used in #3661 could not result in more than 5 additional seconds of unavailability detection, and that is only if it became unavailable immediately after the last check completed successfully, and the check itself took just under 5 seconds to run (a situation in which the current monitor would be running the same job almost constantly). For jobs that run quickly, there is almost no difference.

Do you believe it is a problem to run a health check 5 seconds after the last health check completed, as opposed to every 5 seconds and skipping if we have not finished the last check?

jeremyevans:
@byucesoy After discussion with @fdr, I'm planning to merge this, and submit a PR to address the issues you brought up, as well as some other issues.

PR #3659, opened 2025-07-24
Use higher target utilization for premium hosts

enescakir:
- **Use Config.allocator_target_host_utilization directly in the allocator**
  I'm not sure why we're using it as a class instance variable since we
  already access the Config class directly in our codebase.
  

- **Use higher target utilization for premium hosts**
  To prevent fragmentation, we currently only fill hosts up to a threshold
  (currently 72%). This is to avoid overloading hosts, especially since
  multiple VMs on the same host share some resources. However, this
  threshold causes inefficiencies, especially for hosts with fewer but
  more powerful cores like AX102. Since AX102 hosts have fewer but
  stronger cores, they are less affected by resource contention. This
  should improve premium runner availability, especially for higher vCPU
  requests like 30 vCPUs jobs, since we pack others more.


bsatzger (approved):

scheduling/allocator.rb
```diff
@@ -24,7 +13,8 @@ def self.allocate(vm, storage_volumes, distinct_storage_devices: false, gpu_coun
       gpu_count,
       gpu_device,
       vm.ip4_enabled,
-      target_host_utilization,
+      Config.allocator_target_host_utilization,
```
bsatzger: what about we don't add `target_premium_host_utilization` but set the `target_host_utilization` to either `Config.allocator_target_host_utilization` or `Config.allocator_target_premium_host_utilization`.

enescakir: If we set the single value to %85, it means that “premium VMs can be allocated on both standard/premium hosts up to 85%,” not that “premium hosts can be filled up to 85%”

PR #3661, opened 2025-07-24
Refactor monitor

jeremyevans:
Previously, monitor started the monitoring processing for every
resource every 5 seconds. If the resource was still being processed
from a previous run, it reported the resource was locked, and still
added it to the queue to process. When it came time for the worker
thread to process it, if it was still locked it was skipped.

This changes the behavior so that monitor will never attempt to
start a monitoring job for a resource currently being monitored.
Worker threads will pop jobs to process off the main queue (now called
the submit queue). When the worker thread is done processing, it will
push the the job onto the finish queue. During each loop iteration,
the main thread will pop jobs off the finish queue, and add them
to the run queue.  Then it will run all jobs in the run queue that
finished more than 5 seconds ago.

So previously, jobs were run every 5 seconds. If a job took 4 seconds
to run, it was run 1 second after it last finished. This changes the
behavior so jobs are run 5 seconds after they finish. So jobs that
take longer to run are run less frequently.

When jobs finish, we record the finish time for the job, which is
used for scheduling the next execution.  Finished jobs are appended
to the run queue.  When processing jobs, we find all jobs that should
be run (jobs are ordered by finish time), and slice off the front of
the run queue, and then submit those jobs.

We no longer emit "Resource is locked." messages, because resources
are never locked. There are no mutexes used anymore. The "owning"
thread of a resource is the last thread that popped it off a queue.

To handle checking for stuck pulses, the main threads runs a check
of all resources every 5 seconds, and logs if there is an active job
for the resource and it has been running for more than than expected
number of seconds. Technically, this accesses the same object from
multiple threads (the main thread and the worker thread) at the same
time. However, it only calls an attr_reader method on the object,
and doing that from multiple threads in Ruby is safe, even if the
object is not frozen.

The MonitorResourceType#scan method now returns an array with the
new resources scanned.  These resources are immediately enqueued,
without going through the run queue. This simplifies things, as the
run queue only has to deal with resources that have a finish time.

The MonitorResourceType#enqueue method now takes a Time agument,
and will enqueue all jobs in the run queue that finished before that
time. It will also move jobs from the finish queue to the run queue.
It returns the finish time for the next job to run in the run queue
(or nil if the run queue is empty).

The main thread will consider the times returned by
MonitorResourceType#enqueue.  It will sleep for whatever the least
amount of time is until it is time to run the next job.  That means the
main thread iteration will run much more often than every 5 seconds,
but it will due much less work each time.  If there are no jobs in the
either run queue, then it will sleep for 5 seconds, as it did before.

The metric information reported by monitor has been expanded. For
both resource types, it now shows:

* Total number of resources
* Submit queue length (will show whether and the extent of a backlog
  of jobs)
* Number of idle worker threads (worker threads waiting on the submit
  queue, potentially indicating too many workers)
* Number of threads waiting for a database connection (this required
  a new feature in Sequel, so I updated to the latest Sequel commit).

Since we are not waiting 5 seconds per iteration, we only emit the
logged information if it has been more than 5 seconds since the last
emit.

This is a pretty extensive change, and since we don't yet have a
test suite for monitor, fairly risky. We should add a test suite
for monitor.  I'll work on that next.

----

> [!IMPORTANT]
> Refactor monitor system to improve job handling and partitioning, add smoke tests, and update CI configuration.
> 
>   - **Behavior**:
>     - Monitor no longer starts jobs for resources currently being monitored, preventing re-queuing of locked resources.
>     - Introduces `submit_queue`, `finish_queue`, and `run_queue` in `MonitorResourceType` to manage job processing.
>     - `MonitorRunner` handles job scheduling and execution, ensuring jobs run 5 seconds after completion.
>     - `MonitorRepartitioner` dynamically adjusts partitioning based on active monitor processes.
>     - Removes "Resource is locked." messages as resources are no longer locked.
>   - **Models**:
>     - Adds `MonitorRepartitioner` and `MonitorResourceStub` for handling partitioning and testing.
>     - Refactors `MonitorResourceType` to use new queuing system.
>   - **Testing**:
>     - Adds `monitor_smoke_test.rb` to test monitor functionality.
>     - Updates Rakefile with `:monitor_smoke_test` task.
>     - Adds CI workflow `monitor-ci.yml` to run smoke tests on changes.
>   - **Misc**:
>     - Updates `Gemfile` to use latest Sequel commit for database connection handling.
>     - Refactors `bin/monitor` to use new `MonitorRunner` and `MonitorRepartitioner` classes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8f9a53ac990b962700cd8329cb3b83a08ad3215b. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/monitor-ci.yml
+++ b/.github/workflows/monitor-ci.yml
@@ -1,0 +1,53 @@
+name: monitor CI
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'bin/monitor'
+      - 'lib/monitor_runner.rb'
+      - 'lib/monitor_repartitioner.rb'
+      - 'lib/monitor_resource_type.rb'
+      - 'lib/monitorable_resource.rb'
+      - 'lib/monitor_resource_stub.rb'
+      - 'lib/metrics_target_resource.rb'
+      - 'spec/monitor_smoke_test.rb'
+      - '.github/workflows/monitor-ci.yml'
+      - '.github/actions/setup-clover/action.yml'
+  pull_request:
+    paths:
+      - 'bin/monitor'
+      - 'lib/monitor_runner.rb'
+      - 'lib/monitor_repartitioner.rb'
+      - 'lib/monitor_resource_type.rb'
+      - 'lib/monitorable_resource.rb'
+      - 'lib/monitor_resource_stub.rb'
+      - 'lib/metrics_target_resource.rb'
+      - 'spec/monitor_smoke_test.rb'
+      - '.github/workflows/monitor-ci.yml'
+      - '.github/actions/setup-clover/action.yml'
+
+jobs:
+  cli-ci:
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ubicloud, ubicloud-arm]
+    name: monitor CI - ${{matrix.runs-on}}
+    runs-on: ${{matrix.runs-on}}
+
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v4
+
+    - name: Set up Clover
+      uses: ./.github/actions/setup-clover
+
+    - name: Run monitor smoke test
+      run: bundle exec rake monitor_smoke_test
+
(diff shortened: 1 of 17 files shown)


fdr (approved):
seeing that this merges into another branch, I'll leave some other comments there.

PR #3673, opened 2025-07-29
Support tags for Postgres Resources

byucesoy:
This allows setting tags for Postgres Resources, so that it can be identified by user. Currently this is only available in API layer. In the future, we can possibly expose it in the UI as well.

----

> [!IMPORTANT]
> Add support for setting and updating tags on Postgres resources, including schema changes, API updates, and tests.
> 
>   - **Behavior**:
>     - Add support for setting and updating `tags` for Postgres resources in `postgres_post()` and `postgres_list()` in `helpers/postgres.rb`.
>     - Update `tags` in `pg.update()` in `routes/project/location/postgres.rb`.
>     - Add `tags` to the OpenAPI specification in `openapi.yml`.
>   - **Database**:
>     - Add `tags` column to `postgres_resource` table in `migrate/20250729_postgres_tags.rb`.
>   - **Models**:
>     - Add `tags` attribute to `PostgresResource` in `postgres_resource.rb`.
>   - **Serialization**:
>     - Include `tags` in `serialize_internal()` in `serializers/postgres.rb`.
>   - **Tests**:
>     - Add test for setting and updating `tags` in `postgres_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9305e2b0937f9b52bb7ea7b834d34781062fb0e1. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


migrate/20250729_postgres_tags.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:postgres_resource) do
+      add_column :tags, :jsonb, null: false, default: "[]"
```
enescakir, line 6: Can we use our object tags for this? cc @jeremyevans

jeremyevans: `ObjectTag`s are designed for use used by the authorization system. It is possible to use them for non-authorization purposes, though.  Not sure if that would be better or worse.

byucesoy: Retrofitting `ObjectTag` to this use case seems like changing their purpose significantly. Also we just need to store a simple KV array for which `jsonb` column feels like a better fit. I'm inclined to keep this as is.

shikharbhardwaj (approved):

PR #3679, opened 2025-07-30
Increase ssh keep_alive interval to 60 seconds

mohi-kalantari:
As our load increased, we scaled up accordingly, and the current keep_alive interval, which is set to 3 seconds, is too aggressive. By increasing the interval, we allow the VmHosts to catch up with the keepalive packets.

Right now, we are getting many SSH exceptions in LoadBalancerVmPort monitor like this:
class IOError, message: closed stream
for loadbalancer healthchecks.

----

> [!IMPORTANT]
> Increase SSH `keepalive_interval` to 60 seconds in `model/sshable.rb` to reduce exceptions and align with OpenSSH defaults.
> 
>   - **Behavior**:
>     - Increase `keepalive_interval` from 3 to 60 seconds in `COMMON_SSH_ARGS` in `model/sshable.rb` to reduce SSH exceptions.
>     - Aligns with OpenSSH's default keepalive interval.
>   - **Rationale**:
>     - Current 3-second interval is too aggressive under increased load, causing SSH exceptions.
>     - Allows VmHosts to better handle keepalive packets.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 26db3483adb491aa9a1337d86f2061d3584204bc. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -133,7 +133,7 @@ class Sshable < Sequel::Model
   COMMON_SSH_ARGS = {non_interactive: true, timeout: 10,
                      user_known_hosts_file: [], global_known_hosts_file: [],
                      verify_host_key: :accept_new, keys: [], key_data: [], use_agent: false,
-                     keepalive: true, keepalive_interval: 3, keepalive_maxcount: 5}.freeze
+                     keepalive: true, keepalive_interval: 60, keepalive_maxcount: 5}.freeze
 
   def connect
     Thread.current[:clover_ssh_cache] ||= {}


jeremyevans (approved):

fdr:
I bet this doesn't help. That's tcp keepalive, which the kernel handles.

fdr:
actually, it might be ssh not-quite-keepalive, (though the packet has the bytes `keepalive@openssh.com`) it's just "ServerAlive", (vs ClientAlive on the server side) but it probably still won't help, since the problem I think we have is the _server_ disconnecting us, not the other way around.

fdr:
let us try #3683 first. Given the information we have, it makes more sense to me that this would be the problem (particularly: as failures seem to relate to load on monitor)

fdr:
Okay, there's enough metric evidence that aa2e0c0749903b863bf5989b1c59f0cf6de7d8bb, 8df8d6c7acd7b4c29070111c3afa22b83054bb23 do the job, and 9772ffd8d0d01e2786b12823b755f502a4e53ac7 might do it better still.

PR #3683, opened 2025-07-31
For load balancers, do one free retry on a stale connection

fdr:
This is some pretty rough work, because this problem affects *all* low level monitoring routines and it's not sensible to go around solving them all one by one like this.  But, we have a timely problem with load balancer flaps from stale connections when the monitor process doesn't check in with its SSH connections often enough, and this can be a model for a solution.

The problem is: in the event that we run pulses less than once per ten seconds, we currently record a flap of availability each and every time from the resulting `IOError` exception, because servers are configured as of 22f3aa5c1dcc7a219476af066dc849faf58d5763 to close connections ten seconds after the last non-ClientAlive packet (which is only sent when the protocol is idle), as seen in `sshd_config`:

    ClientAliveInterval 2
    ClientAliveCountMax 4

This patch make the consequence of some slowness gracefully degrade: it anticipates that the server would have dropped connections that have not had a `check_pulse` run in a while, and offers *one* retry without recording the failed probe as a flap.


fdr:
alas, this is indicative of the interface for check_pulse being a bit anemic (get passed the ssh session, handle exceptions your own way, i.e. something as complex as this has no place just yet to be centralized). It's also a bit complicated with the ssh tunneling for Postgres FEBE protocol traffic, the recover-and-ignore code would be slightly different, there.

fdr:
really gotta eliminate those non-rewriting rubocop rules in one swoop one of these days

mohi-kalantari (approved):

PR #3684, opened 2025-07-31
Pry Audit

byucesoy:
This PR builds on top of @fdr's POC to enable auditing of `pry` commands. It does some refactoring, adds minor functionality and tests to cover every line/branch.

----

> [!IMPORTANT]
> Adds auditing for `pry` commands using `LogDnaBatcher` for logging inputs and outputs, with comprehensive tests.
> 
>   - **Behavior**:
>     - Adds auditing for `pry` commands in `bin/pry` by logging inputs and outputs using `LogDnaBatcher`.
>     - Exits if `INGEST_KEY` is not set in production.
>     - Logs errors to Sentry if evaluation fails.
>   - **LogDnaBatcher**:
>     - New class `LogDnaBatcher` in `lib/log_dna_batcher.rb` for batching and sending logs to LogDNA.
>     - Handles connection management, batch sending, and error handling.
>   - **Tests**:
>     - Adds `log_dna_batcher_spec.rb` to test `LogDnaBatcher` functionality, including connection handling, batch sending, and error scenarios.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for aa37831138f1a31b1378ff7efae155eff309b987. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bin/pry
+++ b/bin/pry
@@ -65,7 +65,7 @@ if ENV["INGEST_KEY"]
       @counter ||= 0
       @counter += 1
 
-      PRY_LOGGER.log("INPUT: #{code.strip}", app: "pry", level: "info", type: "pry_input", counter: @counter)
+      PRY_LOGGER.log("$> #{code.strip}", app: "pry", level: "info", type: "pry_input", counter: @counter)
 
       # Evaluate and capture result
       begin
@@ -74,11 +74,11 @@ if ENV["INGEST_KEY"]
         # Log output
         output = result.inspect
         output = "#{output[0..500]}..." if output.length > 500
-        PRY_LOGGER.log("OUTPUT: #{output}", app: "pry", level: "info", type: "pry_output", counter: @counter)
+        PRY_LOGGER.log(output, app: "pry", level: "info", type: "pry_output", counter: @counter)
 
         result
       rescue => e
-        PRY_LOGGER.log("EVAL_ERROR: #{e.class}: #{e.message}\n#{e.backtrace.join("\n")}", app: "pry", level: "error", type: "eval_error", counter: @counter)
+        PRY_LOGGER.log("#{e.class}: #{e.message}\n#{e.backtrace.join("\n")}", app: "pry", level: "error", type: "eval_error", counter: @counter)
         raise
       end
     end
--- a/bin/pry
+++ b/bin/pry
@@ -62,9 +62,7 @@ if ENV["INGEST_KEY"]
 
   module PryLogger
     def evaluate_ruby(code)
-      PRY_LOGGER.log("INPUT: #{code.strip}",
-        level: "info",
-        type: "pry_input")
+      PRY_LOGGER.log("INPUT: #{code.strip}", app: "pry", level: "info", type: "pry_input")
 
       # Evaluate and capture result
       begin
@@ -73,15 +71,11 @@ if ENV["INGEST_KEY"]
         # Log output
         output = result.inspect
         output = "#{output[0..500]}..." if output.length > 500
-        PRY_LOGGER.log("OUTPUT: #{output}",
-          level: "info",
-          type: "pry_output")
+        PRY_LOGGER.log("OUTPUT: #{output}", app: "pry", level: "info", type: "pry_output")
 
         result
       rescue => e
-        PRY_LOGGER.log("EVAL_ERROR: #{e.class}: #{e.message}",
-          level: "error",
-          type: "eval_error")
+        PRY_LOGGER.log("EVAL_ERROR: #{e.class}: #{e.message}", app: "pry", level: "error", type: "eval_error")
         raise
       end
     end
--- a/bin/pry
+++ b/bin/pry
@@ -52,4 +52,39 @@ if Config.development?
   Pry.prepend(PryReloader)
 end
 
+PRY_LOGGER = LogDnaBatcher.new(ENV["INGEST_KEY"])
+
+module PryLogger
+  def evaluate_ruby(code)
+    PRY_LOGGER.log("INPUT: #{code.strip}",
+      level: "info",
+      type: "pry_input")
+
+    # Evaluate and capture result
+    begin
+      result = super
+
+      # Log output
+      output = result.inspect
+      output = "#{output[0..500]}..." if output.length > 500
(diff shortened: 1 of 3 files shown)


lib/log_dna_batcher.rb
```diff
@@ -0,0 +1,116 @@
...
+    data.merge!(metadata)
+    data[:line] = line
+    data[:timestamp] = timestamp
+    @input_queue.push(data)
+    nil
+  end
+
+  def stop
+    @input_queue.close
+    @processor_thread&.join
```
byucesoy: I added a test case to cover the case where `@processor_thread` is `nil`, however, I'm not sure if we can realistically hit this case.

`initialize` always call `start_processor`, which sets `@processor_thread`. Is there any reason to use safe navigation here (cc @fdr)

jeremyevans: For `@processor_thread` to not be set, you would need to be using `allocate` or manually unsetting it. Neither of those seem likely. I think it is safe to remove the safe navigation.

We may want to provide an argument to `join` for the maximum time we are willing to wait when stopping. Otherwise, this can wait indefinitely.

lib/log_dna_batcher.rb
```diff
@@ -0,0 +1,116 @@
...
+        if input_queue_is_closed || flush_interval_is_exceeded || max_batch_size_is_exceeded
+          send_batch(batch)
+        end
+
+        if batch.empty?
+          last_flush_time = Time.now
+          break if input_queue_is_closed
+        end
+      rescue => e
+        puts "Error in processor: #{e.message}"
```
jeremyevans, line 110: The error message here should be more specific, since it could show up on the user's terminal. I would mention the class and method.

In production, if we hit this, we may have already missed logging something we should have logged. I don't think we should continue processing at that point, because we could continue missing logs. We should probably have the `rescue` at the thread level, and should exit the process after printing the error message.

bin/pry
```diff
@@ -52,4 +52,47 @@ if Config.development?
   Pry.prepend(PryReloader)
 end
 
+if Config.production? || ENV["INGEST_KEY"]
+  PRY_LOGGER = LogDnaBatcher.new(ENV["INGEST_KEY"])
```
jeremyevans, line 61: If ingest key is not set in production, I think we should print an error message and exit.

lib/log_dna_batcher.rb
```diff
@@ -0,0 +1,116 @@
...
+      request["Content-Type"] = "application/json; charset=UTF-8"
+      request["Connection"] = "keep-alive"
+      request.body = {lines: batch}.to_json
+
+      response = http.request(request)
+
+      if response.is_a?(Net::HTTPSuccess)
+        batch.clear
+      else
+        puts "Failed to send logs: #{response.code} #{response.message}"
```
jeremyevans, line 54: We should track the number of failures, and after a number of consecutive failures, we should raise a different error that will result in the process exiting.  Otherwise, we run the risk of continuing to run while things are not getting logged.

jeremyevans (commented):
This looks pretty good.

lib/log_dna_batcher.rb
fdr: nah I just kind of bashed this out, and of course, no tests at the time...so I expect there to be some oversights like this one.

lib/log_dna_batcher.rb
```diff
@@ -0,0 +1,116 @@
...
+    @flush_interval = flush_interval
+    @max_batch_size = max_batch_size
+    @input_queue = Queue.new
+    @http = nil
+    @base_url = URI.parse(base_url)
+    @default_metadata = default_metadata.dup.freeze
+
+    params = @base_url.query ? URI.decode_www_form(@base_url.query).to_h : {}
+    params["hostname"] = Socket.gethostname
+    params["app"] = $0
```
fdr: You sure that actually works, when in the query parameters? It's redundant with `default_metadata`, and I had as it was at least in part because I seem to remember not being able to get "app" into the query string. But, I was hacking this out quickly, could have been in error.

Also, default metadata is passed as a parameter, so, like...caller doesn't have to live with $0 if they don't want to.

spec/lib/log_dna_batcher_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      log_dna_batcher.log("test log")
+      sleep 0.5
+    end
+
+    it "logs error in case of an exception during batch processing" do
+      log_dna_batcher = described_class.new("dummy-key")
+
+      expect(log_dna_batcher.instance_variable_get(:@input_queue)).to receive(:pop).and_raise(StandardError, "Unexpected error").at_least(:once)
+      expect(log_dna_batcher).to receive(:puts).with("Error in processor: Unexpected error").at_least(:once)
+      sleep 0.5
```
fdr: these sleeps are brutal. should be possible to get around them.

fdr (changes requested):

spec/lib/log_dna_batcher_spec.rb
jeremyevans: My typical approach is to use a Queue in the specs for synchronization in these cases. The dispatcher and monitor repartitioner specs both have examples.

lib/log_dna_batcher.rb
byucesoy: Yep, you are right. Removed that `app` as param.

byucesoy: That said, it seems few fields like `app`, `level` are first class citizens on [Mezmo's ingest endpoint](https://docs.mezmo.com/log-analysis-api#tag-ingestion) compared to other metadata fields. So I think it would make sense to make them separate parameters than other metadata fields in `log` method.  So I'm removing it from `default_parameters` as well and passing it explicitly.

furkansahin (approved):

spec/lib/log_dna_batcher_spec.rb
byucesoy: I used the Queue approach Jeremy suggested.

PR #3687, opened 2025-07-31
Add support for stale connection retry on connection reset by peer

fdr:
My change for aa2e0c0749903b863bf5989b1c59f0cf6de7d8bb, with the fix in 8df8d6c7acd7b4c29070111c3afa22b83054bb23, appears effective, but only handles system call failures on write: sometimes the program only receives notification of the connection being closed upon a read, like this:

    /app/vendor/bundle/ruby/3.4.0/gems/net-ssh-7.3.0/lib/net/ssh/buffered_io.rb:64:in 'BasicSocket#recv'
    /app/vendor/bundle/ruby/3.4.0/gems/net-ssh-7.3.0/lib/net/ssh/buffered_io.rb:64:in 'Net::SSH::BufferedIo#fill'
    /app/vendor/bundle/ruby/3.4.0/gems/net-ssh-7.3.0/lib/net/ssh/connection/session.rb:275:in 'block in Net::SSH::Connection::Session#ev_do_handle_events'

The exception is an instance of `Errno::ECONNRESET` with a message like `Connection reset by peer - recvfrom(2)`.

So, handle it much the same way, and refactor the tests to test with either exception.

----

> [!IMPORTANT]
> Add handling for `Errno::ECONNRESET` in `check_probe` for stale connection retries and update tests accordingly.
> 
>   - **Behavior**:
>     - `check_probe` in `load_balancer_vm_port.rb` now handles `Errno::ECONNRESET` exceptions for stale connection retries, similar to `IOError` with "closed stream".
>     - Retries occur if `session[:last_pulse]` is older than 8 seconds.
>   - **Tests**:
>     - Updated `load_balancer_vm_port_spec.rb` to test `check_probe` with both `IOError` and `Errno::ECONNRESET` exceptions.
>     - Ensures retries and correct behavior for stale and fresh connections.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for f3e331ad6d70f64ee078db06c3f4c2385ffe406e. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/model/load_balancer_vm_port_spec.rb
+++ b/spec/model/load_balancer_vm_port_spec.rb
@@ -98,22 +98,26 @@ RSpec.describe LoadBalancerVmPort do
         expect(session[:ssh_session]).to receive(:exec!).with("healthcheck command").and_raise("error").at_least(:once)
         expect(lb_vm_port.health_check(session:)).to eq(["down", "down"])
       end
+    end
 
-      context "when exercising stale-connection retry" do
+    [IOError.new("closed stream"), Errno::ECONNRESET.new("Connection reset by peer - recvfrom(2)")].each do |ex|
+      describe "#check_probe", "stale connection retry behavior with #{ex.class}" do
+        let(:session) { {ssh_session: instance_double(Net::SSH::Connection::Session)} }
+
         it "is down if :last_pulse is not set, even with a qualifying exception" do
-          expect(session[:ssh_session]).to receive(:exec!).and_raise(IOError.new("closed stream"))
+          expect(session[:ssh_session]).to receive(:exec!).and_raise(ex)
           expect(lb_vm_port.check_probe(session, :ipv6)).to eq("down")
         end
 
         it "does not retry if the connection is fresh" do
           session[:last_pulse] = Time.now - 1
-          expect(session[:ssh_session]).to receive(:exec!).and_raise(IOError.new("closed stream"))
+          expect(session[:ssh_session]).to receive(:exec!).and_raise(ex)
           expect(lb_vm_port.check_probe(session, :ipv6)).to eq("down")
         end
 
         it "is up on a retry on a stale connection that works the second time" do
           session[:last_pulse] = Time.now - 10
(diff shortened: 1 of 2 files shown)


spec/model/load_balancer_vm_port_spec.rb
```diff
@@ -123,22 +123,52 @@
...
-          session[:last_pulse] = Time.now - 10
-
-          expect(session[:ssh_session]).to receive(:exec!).and_raise(IOError.new("closed stream"))
-
-          second_ssh = instance_double(Net::SSH::Connection::Session)
-          expect(second_ssh).to receive(:exec!).and_raise(IOError.new("closed stream"))
-          expect(lb_vm_port).to receive(:init_health_monitor_session).and_return(ssh_session: second_ssh)
-
-          expect(lb_vm_port.check_probe(session, :ipv6)).to eq("down")
+        shared_examples "a stale connection retrier" do |exception_class, exception_message|
```
jeremyevans: My personal preference is to avoid shared example and use a loop in the cases where it is simple to do so. In this case, you could iterate over a hash of exception class keys and exception message values.  However, that's up to you.

jeremyevans (approved):

spec/model/load_balancer_vm_port_spec.rb
fdr: I also was thinking of using a loop, but was wondering if rspec would do the right thing when running individual tests. Let me give it a  shot.

fdr:
@jeremyevans alright, I whacked things around a bit to decrease the amount of indentation/nesting and formatting messages and got a loop in there. On balance I also prefer loops if there aren't many rspec gotchas.

fdr:
You may not want to bother reading the fixup diff, it should scan better under cumulative

fdr:
alright autosquashed, will deploy presently

PR #3694, opened 2025-08-04
Enable alien vm runners behing feature flag

enescakir:
- **Use the specified size of the boot image for AWS instances**
  The provided boot image size isn't used. We have hardcoded a value of
  40GB.
  
  Instead, we should use the given boot image size to provision the OS
  disk.
  

- **Add billing rate for m7a series**
  The m7a series features AMD EPYC 9R14 processors, which are the closest
  match to the AMD EPYC 9454P used in the AX162 for our standard runners.
  
  We use eu-central-1, which is Frankfurt, similar to our Hetzner hosts.
  
                       9R14(m7a)      9454P(AX162)
       Clockspeed       3.7 GHz      Up to 3.8 GHz
             Year       Q3 2023         Q3 2023
      Single Core        2920            2988
  

- **Add m7a support**
  In AWS SKUs, "g" stands for Graviton, "a" for AMD, and "i" for Intel.
  
  So, if it contains "g," it is arm64.
  

- **Enable alien VM runners on AWS**
  When we have high demand for runners, we exceed the capacity of our host
  fleet.
  
  To handle this spiky demand, we route the requests to AWS VMs. We use
  "alien" to identify the VM on the external cloud rather than on our own
  hosts.
  
  This PR uses the same logic as Ubi-on-AWS, which is used for PostgreSQL.
  When you provide a private location and AWS vm SKU, the nexus assembler
  provisions it on AWS.
  
  Currently, it's controlled by a feature flag in the project. After we
  test it, we will change the logic to route only the exceeded demand.
  

----

> [!IMPORTANT]
> Enable alien VM runners on AWS with m7a support, controlled by a feature flag, and update billing rates and tests accordingly.
> 
>   - **Behavior**:
>     - Enable alien VM runners on AWS, routing spiky demand to AWS VMs, controlled by a feature flag in `project.rb`.
>     - Use specified boot image size for AWS instances in `instance.rb`.
>     - Add m7a support in `option.rb` and `instance.rb`.
>   - **Billing**:
>     - Add billing rate for m7a series in `billing_rates.yml`.
>   - **Tests**:
>     - Update tests in `instance_spec.rb` and `github_runner_spec.rb` to cover new AWS VM logic and m7a support.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7390a124c612f54a8c8eacc39083f2b363bf89e6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -109,6 +109,7 @@ module Config
   optional :vm_pool_project_id, string
   optional :github_runner_service_project_id, string
   override :enable_github_workflow_poller, true, bool
+  optional :github_runner_aws_location_id, string
 
   # GitHub Cache
   optional :github_cache_blob_storage_endpoint, string
@@ -167,6 +168,7 @@ module Config
   override :github_ubuntu_2404_version, "20250821.1.0", string
   override :github_ubuntu_2204_version, "20250821.1.0", string
   override :github_gpu_ubuntu_2204_version, "20250821.1.0", string
+  override :github_ubuntu_2204_aws_ami_version, "ami-08b71dd98be505433", string
   override :postgres16_ubuntu_2204_version, "20250425.1.1", string
   override :postgres17_ubuntu_2204_version, "20250425.1.1", string
   override :postgres16_paradedb_ubuntu_2204_version, "20250425.1.1", string
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -46,9 +46,19 @@ class Prog::Vm::GithubRunner < Prog::Base
       return picked_vm
     end
 
+    boot_image = label_data["boot_image"]
+    location_id = Location::GITHUB_RUNNERS_ID
+    size = label_data["vm_size"]
+    alien_ratio = github_runner.installation.project.get_ff_aws_alien_runners_ratio || 0
+    if label_data["arch"] == "x64" && boot_image == "github-ubuntu-2204" && rand < alien_ratio
+      boot_image = Config.send(:"#{boot_image.tr("-", "_")}_aws_ami_version")
+      location_id = Config.github_runner_aws_location_id
+      size = Option.aws_instance_type_name("m7a", label_data["vcpus"])
+    end
+
     ps = Prog::Vnet::SubnetNexus.assemble(
       Config.github_runner_service_project_id,
-      location_id: Location::GITHUB_RUNNERS_ID,
+      location_id:,
       allow_only_ssh: true
     ).subject
 
@@ -57,9 +67,9 @@ class Prog::Vm::GithubRunner < Prog::Base
       unix_user: "runneradmin",
       sshable_unix_user: "runneradmin",
       name: github_runner.ubid.to_s,
-      size: label_data["vm_size"],
-      location_id: Location::GITHUB_RUNNERS_ID,
-      boot_image: label_data["boot_image"],
+      size:,
+      location_id:,
+      boot_image:,
       storage_volumes: [{size_gib: label_data["storage_size_gib"], encrypted: true, skip_sync: true}],
       enable_ip4: true,
       arch: label_data["arch"],
@@ -81,6 +91,8 @@ class Prog::Vm::GithubRunner < Prog::Base
       # If we enable free upgrades for the project, we should charge
       # the customer for the label's VM size instead of the effective VM size.
       label_data["vm_size"]
+    elsif github_runner.vm.location.aws?
+      "standard-#{vm.vcpus}"
     else
       "#{vm.family}-#{vm.vcpus}"
     end
@@ -210,6 +222,8 @@ class Prog::Vm::GithubRunner < Prog::Base
   end
 
   def setup_info
+    vmh = vm.vm_host
+    project = github_runner.installation.project
     {
       group: "Ubicloud Managed Runner",
       detail: {
@@ -218,12 +232,12 @@ class Prog::Vm::GithubRunner < Prog::Base
         "VM Family" => vm.family,
         "Arch" => vm.arch,
         "Image" => vm.boot_image,
-        "VM Host" => vm.vm_host.ubid,
+        "VM Host" => vmh&.ubid,
         "VM Pool" => vm.pool_id ? UBID.from_uuidish(vm.pool_id).to_s : nil,
-        "Location" => Location[vm.vm_host.location_id].name,
-        "Datacenter" => vm.vm_host.data_center,
-        "Project" => github_runner.installation.project.ubid,
-        "Console URL" => "#{Config.base_url}#{github_runner.installation.project.path}/github"
+        "Location" => vmh&.location&.name,
+        "Datacenter" => vmh&.data_center,
+        "Project" => project.ubid,
+        "Console URL" => "#{Config.base_url}#{project.path}/github"
       }.map { "#{_1}: #{_2}" }.join("\n")
     }
   end
--- a/spec/prog/vm/github_runner_spec.rb
+++ b/spec/prog/vm/github_runner_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe Prog::Vm::GithubRunner do
   let(:installation) { runner.installation }
   let(:project) { installation.project }
   let(:client) { instance_double(Octokit::Client) }
-  let(:now) { Time.utc(2025, 5, 19, 19, 0) }
+  let(:now) { Time.utc(2025, 8, 1, 19, 0) }
 
   before do
     allow(Config).to receive(:github_runner_service_project_id).and_return(vm.project_id)
@@ -100,6 +100,17 @@ RSpec.describe Prog::Vm::GithubRunner do
       expect(vm.id).to eq(picked_vm.id)
       expect(picked_vm.family).to eq("premium")
     end
+
+    it "uses alien vms if enabled" do
+      project.set_ff_aws_alien_runners_ratio(0.5)
+      expect(nx).to receive(:rand).and_return(0.4)
+      location = Location.create(name: "eu-central-1", provider: "aws", project_id: vm.project_id, display_name: "aws-eu-central-1", ui_name: "AWS Frankfurt", visible: true)
+      expect(Config).to receive(:github_runner_aws_location_id).and_return(location.id)
+      picked_vm = nx.pick_vm
+      expect(picked_vm.family).to eq("m7a")
+      expect(picked_vm.location.aws?).to be(true)
+      expect(picked_vm.boot_image).to eq(Config.github_ubuntu_2204_aws_ami_version)
+    end
   end
 
   describe ".update_billing_record" do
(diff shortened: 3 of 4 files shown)


config/billing_rates.yml
```diff
@@ -354,4 +357,4 @@
 - { id: 765221ad-10cb-8178-9d11-f0268a8be95c, resource_type: PostgresVCpu,                resource_family: standard-m8gd,              location: us-east-2,      unit_price: 0.0017399860, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
 - { id: a2d44646-0f12-48d9-82e7-148ef8963322, resource_type: PostgresStandbyVCpu,         resource_family: standard-m8gd,              location: us-west-2,      unit_price: 0.0017399860, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
 - { id: bd3cc3c3-9b7a-8d78-a318-832d37b8ddd9, resource_type: PostgresStandbyVCpu,         resource_family: standard-m8gd,              location: us-east-1,      unit_price: 0.0017399860, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
-- { id: 3444b23d-4044-8978-8cc5-f80e5620286b, resource_type: PostgresStandbyVCpu,         resource_family: standard-m8gd,              location: us-east-2,      unit_price: 0.0017399860, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
```
furkansahin: why is this line removed?

config.rb
```diff
@@ -167,6 +168,7 @@ def self.test?
   override :github_ubuntu_2404_version, "20250724.1.0", string
   override :github_ubuntu_2204_version, "20250724.1.0", string
   override :github_gpu_ubuntu_2204_version, "20250724.1.0", string
+  override :github_ubuntu_2204_aws_ami_version, "ami-08b71dd98be505433", string
```
furkansahin, line 171: we already have aws_ami table. I think it makes more sense to add this as another row there.

config/billing_rates.yml
enescakir: It seems like it moved the end of the file somehow, possibly due to conflict resolution. Let me fix that.

config.rb
enescakir: The table is `pg_aws_ami`. Is it to be used for all AMIs?

enescakir: I saw that but didn't use it because it was prefixed by "Pg"

lib/option.rb
```diff
@@ -85,7 +85,7 @@ def display_name
     VmSize.new("burstable-#{it}", "burstable", it, it * 50, it * 50, (it * 1.6).to_i, storage_size_options, io_limits, false, "arm64")
   }).concat(AWS_FAMILY_OPTIONS.product([2, 4, 8, 16, 32, 64]).map { |family, vcpu|
     memory_coefficient = (family == "c6gd") ? 2 : 4
-    arch = (family == "m6id") ? "x64" : "arm64"
+    arch = family.include?("g") ? "arm64" : "x64"
```
byucesoy: I think it would be better to explicitly list families here. "g" might be in a different place of the family name for a different reason.

config/billing_rates.yml
```diff
@@ -372,4 +375,4 @@
...
 - { id: bc7e1096-6688-8978-88e7-c947e08e68c5, resource_type: PostgresStandbyStorage,      resource_family: m8gd,                       location: us-east-1,      unit_price: 0.0000000000, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
 - { id: 2befac30-f11a-8178-93c0-c7d4ecad8228, resource_type: PostgresStandbyStorage,      resource_family: m8gd,                       location: us-east-2,      unit_price: 0.0000000000, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
-- { id: 0981f65e-5f1d-40fe-a416-c5e3a544f8be, resource_type: PostgresStandbyStorage,      resource_family: m8gd,                       location: ap-southeast-2, unit_price: 0.0000000000, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
\ No newline at end of file
+- { id: 0981f65e-5f1d-40fe-a416-c5e3a544f8be, resource_type: PostgresStandbyStorage,      resource_family: m8gd,                       location: ap-southeast-2, unit_price: 0.0000000000, billed_by: duration, active_from: 2025-06-29T00:00:00Z }
```
enescakir, line 396: It should have a new line at the end according to our `.editorconfig` file.

CleanShot 2025-08-05 at 09 38 43@2x

config.rb
enescakir: It also has `pg_version` column

prog/aws/instance.rb
```diff
@@ -135,7 +135,7 @@ class Prog::Aws::Instance < Prog::Base
             encrypted: true,
             delete_on_termination: true,
             iops: 3000,
-            volume_size: 40,
+            volume_size: vm.vm_storage_volumes.find { it.boot == true }.size_gib,
```
jeremyevans: This retrieves all rows and all columns, but only uses 1 column of a specific row. I don't see vm_storage_volumes elsewhere in the file, so it doesn't appear already cached. This could be more efficient:

```suggestion
            volume_size: vm.vm_storage_volumes_dataset.get(:size_gib),
```

`vm_storage_volumes` association is already ordered by boot DESC, so the bootable volume comes before the non-bootable ones. Assuming the VM has a bootable volume, the behavior is the same, just more efficient.

If the VM does not have a bootable volume, this will give the size of the non-boot volume instead of a NoMethodError. If the VM doesn't have any volumes, this gives nil instead of a NoMethodError. However, I don't think we care about those cases.

prog/vm/github_runner.rb
```diff
@@ -46,9 +46,14 @@ def pick_vm
       return picked_vm
     end
 
+    boot_image, location_id, size = if label_data["arch"] == "x64" && label_data["boot_image"] == "github-ubuntu-2204" && github_runner.installation.project.get_ff_aws_alien_runners
```
jeremyevans: This looks Ubuntu 22.04 specific.  Maybe that is intentional, but if not, it should be updated, I would assume at least to support Ubuntu 24.04.

I think this would be slightly nicer if you set the three variables separately in both the if and else block, instead of having the if/else return an array. But no objections if you want to keep it this way.

prog/vm/github_runner.rb
```diff
@@ -218,10 +225,10 @@ def setup_info
         "VM Family" => vm.family,
         "Arch" => vm.arch,
         "Image" => vm.boot_image,
-        "VM Host" => vm.vm_host.ubid,
+        "VM Host" => vm.vm_host&.ubid,
         "VM Pool" => vm.pool_id ? UBID.from_uuidish(vm.pool_id).to_s : nil,
-        "Location" => Location[vm.vm_host.location_id].name,
-        "Datacenter" => vm.vm_host.data_center,
+        "Location" => vm.vm_host&.location&.name,
+        "Datacenter" => vm.vm_host&.data_center,
```
jeremyevans: Might be worth using a `vm_host` local variable? Up to you.

spec/prog/vm/github_runner_spec.rb
```diff
@@ -23,7 +23,7 @@
   let(:installation) { runner.installation }
   let(:project) { installation.project }
   let(:client) { instance_double(Octokit::Client) }
-  let(:now) { Time.utc(2025, 5, 19, 19, 0) }
+  let(:now) { Time.utc(2025, 8, 1, 19, 0) }
```
jeremyevans, line 26: Is this change necessary? No objections to it.

enescakir: Otherwise, it can't find the active billing rates. It should be later than the `active_from` date.

config.rb
furkansahin: makes sense, keep it as is for now then but you may consider adding a table for your amis as well, or we can modify to code to make pg_aws_ami table to be a more generic one.

prog/aws/instance.rb
furkansahin: I feel like this would drastically reduce the readability of the code. We assume anyone looking into this line will know `vm_storage_volumes association is already ordered by boot DESC`. I feel like if we are going to make this change, we must definitely add a comment or not do it at all.

enescakir: We might change the order of the `vm_storage_volumes` association and forget to update it here in the future. I prefer to keep it explicit which is more developer friendly.

enescakir: For efficiency, I might change it to `vm.vm_storage_volumes_dataset.where(boot: true).get(:size_gib)`.

prog/vm/github_runner.rb
enescakir: Yeah, it's intentional. Right now I'm testing it with Ubuntu 22. We'll also support Ubuntu 24. But for now, I'm optimizing images. Instead of generating both images for every attempt, I'll add support for 24 once I'm done with the optimization.

enescakir: > I think this would be slightly nicer if you set the three variables separately in both the if and else block, instead of having the if/else return an array. But no objections if you want to keep it this way.

Do you mean something like that?
```ruby
boot_image = label_data["boot_image"]
location_id = Location::GITHUB_RUNNERS_ID
size = label_data["vm_size"]
if label_data["arch"] == "x64" && label_data["boot_image"] == "github-ubuntu-2204" && github_runner.installation.project.get_ff_aws_alien_runners
  boot_image = Config.send(:"#{label_data["boot_image"].tr("-", "_")}_aws_ami_version")
  location_id = Config.github_runner_aws_location_id
  size = Option.aws_instance_type_name("m7a", label_data["vcpus"])
end
```

prog/aws/instance.rb
jeremyevans: Note that you can use `where(:boot)` instead of `where(boot: true)`, as `boot` is a boolean column.

In terms of readability, a more readable approach would be to add `Vm#boot_volume_size_gib` and call that. Not sure if we need the boot volume size in any other case, but if so, that's probably worth doing.

prog/vm/github_runner.rb
jeremyevans: I think that looks better. Plus, you could use the `boot_image` variable to replace the `label_data["boot_image"]` in a couple places.

prog/vm/github_runner.rb
```diff
@@ -46,9 +46,18 @@ def pick_vm
       return picked_vm
     end
 
+    boot_image = label_data["boot_image"]
```
velioglu, line 49: nit: boot_image and location_id can be defined at the top to use them with VmPool.where

prog/vm/github_runner.rb
```diff
@@ -46,9 +46,18 @@ def pick_vm
...
+    location_id = Location::GITHUB_RUNNERS_ID
+    size = label_data["vm_size"]
+    if label_data["arch"] == "x64" && label_data["boot_image"] == "github-ubuntu-2204" && github_runner.installation.project.get_ff_aws_alien_runners
```
velioglu: We’re currently using Alien VMs for projects when the flag is enabled. To verify both correctness and performance, we need to compare job results against historical ones. We might run jobs partially on Hetzner and partially on AWS, which would allow us to directly compare results after enabling the flag for a project.

enescakir: I changed boolean feature flag with ratio. It uses alien vms in given ratio `alien_chance = project.get_ff_aws_alien_runners_ratio && rand < project.get_ff_aws_alien_runners_ratio`

PR #3719, opened 2025-08-06
Rename a few models to use singular names

jeremyevans:
Model names should use the singular form, not the plural form.


enescakir:
Since most of these models/associations were added by @furkansahin, it would be good to get his feedback.

PR #3724, opened 2025-08-07
Use Roda path plugin to DRY up path creation code

jeremyevans:
I found a lot of places couldn't benefit from the path plugin due to serializer usage for views, so this removes a lot of serializer usage, before updating the view code to use the path method.


fdr:
passing this one to @enescakir, looks like we win on expression complexity and lose on lines, which is still a win.

routes/github.rb
```diff
@@ -14,7 +14,7 @@ class Clover
         authorize("Project:github", installation.project.id)
         flash["notice"] = "GitHub runner integration is already enabled for #{installation.project.name} project."
         Clog.emit("GitHub installation already exists") { {installation_failed: {id: installation_id, account_ubid: current_account.ubid}} }
-        r.redirect "#{installation.project.path}/github/#{installation.ubid}/runner"
+        r.redirect "#{path(installation)}/runner"
```
enescakir, line 17: Can we add optional parameters to the `path`? Then we can change it to:

```suggestion
        r.redirect path(installation, "runner")
```

routes/project/location/postgres.rb
```diff
@@ -30,7 +30,7 @@ class Clover
           if api?
             Serializers::Postgres.serialize(pg, {detailed: true})
           else
-            r.redirect "#{@project.path}#{pg.path}/overview"
+            r.redirect "#{path(pg)}/overview"
```
enescakir, line 33: Same here. If we can add an optional argument, it will become more clear.

```suggestion
            r.redirect path(pg, "overview")
```

routes/github.rb
enescakir: Docs have such example

```ruby
path Quux do |quux, path|
  "/quux/#{quux.id}/#{path}"
end

```

jeremyevans: It's certainly possible. I avoided it deliberately, though. For r.redirect, I was planning to use a new plugin that allows:

```ruby
r.redirect installation, "/runner"
```

https://github.com/jeremyevans/roda/commit/30da88b47fadca1487bcbf6103d110a45795b602

enescakir: Even cooler. It will remove lots of code

PR #3728, opened 2025-08-08
Remove more serializer usage in web routes

jeremyevans:
This also removes the serializer usage to set `@project_data`. Switch from accessing `@project_data` to accessing `@project`, or in a few cases, accessing `@project_permissions`.

After this, the only remaining usage of serializers in web routes is the invoice serializer. This PR updates the invoice serializer to use a stricter approach, by switching the return from Hash -> Struct -> Data. Struct is used as the middleman because it works both with `#[]` and named methods, while Data only works with named methods. This found a small bug in the existing invoice serializer, which this also fixes.

----

> [!IMPORTANT]
> Remove serializer usage in web routes, update views to use `@project.path`, and refactor `Invoice` model to use `InvoiceData`.
> 
>   - **Behavior**:
>     - Removes serializer usage in web routes, except for invoice serializer.
>     - Replaces `@project_data` with `@project` or `@project_permissions` in views and routes.
>   - **Helpers**:
>     - Adds `humanize_size` and `humanize_time` methods in `web.rb`.
>   - **Models**:
>     - Updates `Invoice` model in `invoice.rb` to use `InvoiceData` structure.
>   - **Views**:
>     - Updates paths in various views to use `@project.path` instead of `@project_data[:path]`.
>     - Removes `Serializers::GithubCacheEntry` and `Serializers::InferenceApiKey`.
>   - **Misc**:
>     - Deletes unused serializer files: `github_cache_entry.rb`, `inference_api_key.rb`, `inference_endpoint.rb`, `inference_router_model.rb`, `usage_alert.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 623ace37f036964d306380f7354d8115f043816c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


views/inference/endpoint/index.erb
```diff
@@ -24,21 +24,21 @@ MSG
 
 
     curl_snippet = <<-CURL
-curl #{ie[:url]}#{path} \\
+curl #{ie.load_balancer.health_check_protocol}://#{ie.load_balancer.hostname}#{path} \\
```
enescakir, line 27: Should we add a helper function for the load balancer health check URL?

views/inference/endpoint/index.erb
```diff
@@ -48,25 +48,25 @@ curl #{ie[:url]}#{path} \\
...
       </div>
       <div>
         <h3 class="text-sm font-semibold text-gray-900">Context length</h3>
-        <p class="text-gray-500"><%= ie[:tags].fetch("context_length", "Full") %></p>
+        <p class="text-gray-500"><%= ie.tags.fetch("context_length", "Full") %></p>
       </div>
       <div class="sm:col-span-2">
         <h3 class="text-sm font-semibold text-gray-900">URL</h3>
-        <div class="text-gray-500"><%== part("components/copyable_content", content: ie[:url], message: "Copied URL") %></div>
+        <div class="text-gray-500"><%== part("components/copyable_content", content: "#{ie.load_balancer.health_check_protocol}://#{ie.load_balancer.hostname}", message: "Copied URL") %></div>
```
enescakir, line 69: LB health check URL helper

serializers/invoice.rb
```diff
@@ -76,6 +76,8 @@ def self.serialize_internal(inv, options = {})
                    amount: amount_sum,
                    cost: cost_sum,
                    cost_humanized: humanized_cost(cost_sum),
+                   resource_type: line_items.first.resource_type,
+                   resource_family: line_items.first.resource_family,
```
enescakir, line 80: If I recall correctly, when @byucesoy  designed it, line items were meant to support different types of resources by design. But we might not actually have such cases in production.

views/inference/endpoint/index.erb
jeremyevans: I think that's a good idea. I'll work on that in a separate PR after merging.

serializers/invoice.rb
jeremyevans: I don't believe these values are actually used. They just need to have a value because of the switch to using a `Data` subclass. Maybe it is better to leave them as `nil`.

PR #3750, opened 2025-08-18
Synchronise VM removal from load balancer in case of destroy

furkansahin:
When we try to destroy the VM completely before the load balancer
completes the metadata endpoint clean-up, the vm gets stuck on removal.
With this commit, we make sure the related networking cleanup is done
only after load balancer completes its cleanup.

----

> [!IMPORTANT]
> Synchronize VM removal with load balancer cleanup to prevent VM removal issues by adding checks and semaphores in `nexus.rb`, `cert_server.rb`, and `load_balancer_nexus.rb`.
> 
>   - **Behavior**:
>     - Ensure VM network cleanup occurs after load balancer cleanup in `nexus.rb` and `cert_server.rb`.
>     - Add `removed_from_lb` semaphore to `vm.rb` to track load balancer removal status.
>   - **Functions**:
>     - Add `incr_removed_from_lb` in `cert_server.rb` and `load_balancer_nexus.rb` to increment removal status.
>     - Add `removed_from_lb_set?` check in `nexus.rb` to verify load balancer removal completion.
>   - **Tests**:
>     - Add tests in `nexus_spec.rb` for `wait_lb_expiry` to verify behavior when VM is not fully removed from load balancer.
>     - Add tests in `cert_server_spec.rb` for `remove_cert_server` to ensure `incr_removed_from_lb` is called.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a886f37ba86efe485024d74924f84add99f2f038. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -588,6 +588,7 @@ class Prog::Vm::Nexus < Prog::Base
       end
       lb.remove_vm(vm)
     end
+    nap 5 unless vm.removed_from_lb_set?
 
     vm.vm_host.sshable.cmd("sudo host/bin/setup-vm delete_net #{q_vm}")
 
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -167,7 +167,7 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
   label def wait_destroy
     reap(nap: 5) do
       load_balancer.destroy
-
+      load_balancer.vms.map(&:incr_removed_from_lb)
       pop "load balancer deleted"
     end
   end
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -33,7 +33,7 @@ class Vm < Sequel::Model
 
   plugin ResourceMethods, redacted_columns: :public_key
   plugin SemaphoreMethods, :destroy, :start_after_host_reboot, :prevent_destroy, :update_firewall_rules,
-    :checkup, :update_spdk_dependency, :waiting_for_capacity, :lb_expiry_started, :restart, :stop
+    :checkup, :update_spdk_dependency, :waiting_for_capacity, :lb_expiry_started, :restart, :stop, :removed_from_lb
   include HealthMonitorMethods
 
   include ObjectTag::Cleanup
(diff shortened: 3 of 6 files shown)


prog/vnet/load_balancer_nexus.rb
```diff
@@ -167,7 +167,7 @@ def need_to_rewrite_dns_records?
   label def wait_destroy
     reap(nap: 5) do
       load_balancer.destroy
-
+      load_balancer.vms.map(&:incr_removed_from_lb)
```
bsatzger, line 170: can that happen?

1. vm is attached to lb1
2. lb1 is destroyed, removed_from_lb is set for vm
3. vm is attached to lb2, removed_from_lb is still set
4. vm is destroyed -> stuck

should we reset the semaphore when attaching a vm to a lb?

prog/vm/nexus.rb
```diff
@@ -588,6 +588,7 @@ def write_params_json
       end
       lb.remove_vm(vm)
     end
+    nap 5 unless vm.removed_from_lb_set?
```
jeremyevans, line 591: This looks odd to me (maybe it is correct, but it looks odd). Shouldn't we only do this check if the vm has/had a load balancer? If the vm never had a load balancer, doesn't this just nap 5 seconds until the semaphore is set manually?

jeremyevans (commented):
I think it may be better to add a prog for the remove-vm-from-load-balancer logic.  In `wait_lb_expiry`, we start a new prog, and continually nap if the vm still has a load balancer. The new prog waits until all necessary steps have been taken, then calls `lb.remove_vm(vm)`, then pops.  After the vm is removed from the load balancer, the vm prog will automatically delete the net and then hop to destroy_slice.

prog/vm/nexus.rb
bsatzger: Was thinking the same. But we only ever go into this label if the vm has a load balancer.

furkansahin:
@jeremyevans , @bsatzger , would you prefer this ? https://github.com/ubicloud/ubicloud/pull/3879

PR #3756, opened 2025-08-19
Allow application_name in PG configs

shikharbhardwaj:
> [!IMPORTANT]
> Add `application_name` configuration to PostgreSQL schemas for PG 16 and PG 17, allowing session-specific application names.
> 
>   - **Behavior**:
>     - Adds `application_name` configuration to `PG_17_CONFIG_SCHEMA` and `PG_16_CONFIG_SCHEMA` in `postgres_config_validator_schema.rb`.
>     - `application_name` allows setting a session-specific application name, defaulting to an empty string.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 1b721f824ebeb2e3bc76a150d610de71e8a2bd1b. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/validation/postgres_config_validator_schema.rb
+++ b/lib/validation/postgres_config_validator_schema.rb
@@ -34,6 +34,11 @@ module Validation
         type: :bool,
         default: "off"
       },
+      "application_name" => {
+        description: "Sets the application name for the session.",
+        type: :string,
+        default: ""
+      },
       "archive_cleanup_command" => {
         description: "Sets the shell command that will be executed at every restart point.",
         type: :string
@@ -2144,6 +2149,11 @@ module Validation
         description: "Allows modifications of the structure of system tables.",
         type: :bool,
         default: "off"
+      },
+      "application_name" => {
+        description: "Sets the application name for the session.",
+        type: :string,
+        default: ""
       },
       "archive_cleanup_command" => {
         description: "Sets the shell command that will be executed at every restart point.",


shikharbhardwaj:
@jeremyevans these files were semi-generated, the base was captured using the `SHOW ALL` output from the corresponding PG version, then typing/help info was added from sources from the web and some additional point fixes made (for eg. removing some configs which `SHOW ALL` outputs but are not configurable). Given the manual fixes, I think treating them as non-generated files makes sense, since there isn't a single source of truth we generate this information from.

On removing duplication: I think that sounds like a good change, we can keep a base postgres schema and a diff of them based on the version. I will file this as a separate change from this PR.

fdr:
This is kind of weird, since application_name is typically set by clients so that it can be exposed in pg_stat_activity. If it works in a custom setting (~= postgresql.conf) it would be more like setting a default application name (who does this, and why?). Thus, the description is misleading, at least in part because the _postgres_ description has an anticipation on how it'd be used (in a session, not via postgresql.conf)

And looking at the neighboring context: do we _really_ expect or want to encourage customers to mess with cleanup commands upon restartpoint/checkpoint without a clear idea of what someone who wants this flexibility might reasonably try to accomplish?

I think this gets the product management backwards: being completionist and having not an idea how we anticipate a configuration might be used but including it because it's there is going to increase choice bloat, and problems.

PR #3761, opened 2025-08-20
Add KubernetesNode object

mohi-kalantari:
In this PR, we introduce a new model named KubernetesNode which will own a Vm and kubernetes related logics like drain would be implemented in its Prog

Why did I create this PR? I was looking for a place to execute the drain logic for nodes and there were no good place for it. I didn't want to add a new label to Vm prog, didn't want to add a new field to the KubernetesNodepool model to store the list of nodes which we need to drain, there was no way to correctly use semaphores so I thought we are missing a key model in our design.

### DIFF
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -9,6 +9,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :private_subnet
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
+  one_to_many :nodes, class: :KubernetesNode, order: :created_at, conditions: {kubernetes_nodepool_id: nil}
   one_to_many :nodepools, class: :KubernetesNodepool
   one_to_many :active_billing_records, class: :BillingRecord, key: :resource_id, &:active
   many_to_one :location, key: :location_id
--- a/model/kubernetes/kubernetes_nodepool.rb
+++ b/model/kubernetes/kubernetes_nodepool.rb
@@ -6,6 +6,7 @@ class KubernetesNodepool < Sequel::Model
   one_to_one :strand, key: :id
   many_to_one :cluster, key: :kubernetes_cluster_id, class: :KubernetesCluster
   many_to_many :vms, order: :created_at
+  one_to_many :nodes, class: :KubernetesNode, order: :created_at
 
   plugin ResourceMethods
   plugin SemaphoreMethods, :destroy, :start_bootstrapping, :upgrade
--- a/prog/kubernetes/provision_kubernetes_node.rb
+++ b/prog/kubernetes/provision_kubernetes_node.rb
@@ -7,6 +7,10 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
     @vm ||= Vm[frame["vm_id"]]
   end
 
+  def node
+    @node ||= KubernetesNode[frame["node_id"]]
+  end
+
   def kubernetes_nodepool
     @kubernetes_nodepool ||= KubernetesNodepool[frame["nodepool_id"]]
   end
@@ -49,7 +53,7 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
 
     boot_image = "kubernetes-#{kubernetes_cluster.version.tr(".", "_")}"
 
-    vm = Prog::Vm::Nexus.assemble_with_sshable(
+    node = Prog::Kubernetes::KubernetesNodeNexus.assemble(
       Config.kubernetes_service_project_id,
       sshable_unix_user: "ubi",
       name: name,
@@ -58,11 +62,15 @@ class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
       storage_volumes: storage_volumes,
       boot_image: boot_image,
       private_subnet_id: kubernetes_cluster.private_subnet_id,
-      enable_ip4: true
+      enable_ip4: true,
+      kubernetes_cluster_id: kubernetes_cluster.id,
+      kubernetes_nodepool_id: kubernetes_nodepool&.id
     ).subject
+    vm = node.vm
 
     current_frame = strand.stack.first
     current_frame["vm_id"] = vm.id
+    current_frame["node_id"] = node.id
     strand.modified!(:stack)
 
     if kubernetes_nodepool
@@ -219,6 +227,6 @@ TEMPLATE
 }
 CONFIG
     vm.sshable.cmd("sudo tee /etc/cni/net.d/ubicni-config.json", stdin: cni_config)
-    pop vm_id: vm.id
+    pop({vm_id: vm.id, node_id: node.id})
   end
 end
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -247,6 +247,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
 
       kubernetes_cluster.api_server_lb&.incr_destroy
       kubernetes_cluster.cp_vms.each(&:incr_destroy)
+      kubernetes_cluster.nodes.each(&:incr_destroy)
       kubernetes_cluster.remove_all_cp_vms
       kubernetes_cluster.nodepools.each { it.incr_destroy }
       kubernetes_cluster.private_subnet.incr_destroy
--- a/prog/kubernetes/kubernetes_nodepool_nexus.rb
+++ b/prog/kubernetes/kubernetes_nodepool_nexus.rb
@@ -77,8 +77,10 @@ class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
     reap do
       decr_destroy
 
+      kubernetes_nodepool.nodes.each(&:incr_destroy)
       kubernetes_nodepool.vms.each(&:incr_destroy)
       kubernetes_nodepool.remove_all_vms
+      nap 5 unless kubernetes_nodepool.nodes.empty?
       kubernetes_nodepool.destroy
       pop "kubernetes nodepool is deleted"
     end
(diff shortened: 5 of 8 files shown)


aamederen:
I think we need a bit more comment on the necessity of this change (The problems dealing with VMs without a Node wrapper, alternatives, etc.)

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
+  subject_is :kubernetes_node
+
+  def self.assemble(project_id, sshable_unix_user:, name:, location_id:, size:, storage_volumes:, boot_image:, private_subnet_id:, enable_ip4:, kubernetes_cluster_id:, kubernetes_nodepool_id: nil)
+    DB.transaction do
+      ubid = KubernetesNode.generate_ubid
+      id = ubid.to_uuid
```
jeremyevans: You aren't using the ubid, so it's simpler to just generate the uuid:

```suggestion
      id = KubernetesNode.generate_uuid
```

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -0,0 +1,39 @@
...
+  label def start
+    hop_wait
+  end
+
+  label def wait
+    nap 6 * 60 * 60
+  end
+
+  label def destroy
+    decr_destroy
```
jeremyevans: When a strand exits, it automatically deletes all semaphores. So I don't think this is needed.

```suggestion
```

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -9,6 +9,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :private_subnet
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
+  one_to_many :nodes, class: :KubernetesNode, key: :kubernetes_cluster_id, order: :created_at, conditions: {kubernetes_nodepool_id: nil}
```
jeremyevans: I don't think `key` is needed. You may want to consider renaming to `cp_nodes`.

```suggestion
  one_to_many :nodes, class: :KubernetesNode, order: :created_at, conditions: {kubernetes_nodepool_id: nil}
```

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -9,6 +9,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :private_subnet
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
```
jeremyevans, line 11: The `kubernetes_clusters_cp_vms` table should eventually be dropped, since you can use `kubernetes_node` as a join table.

model/kubernetes/kubernetes_nodepool.rb
```diff
@@ -6,6 +6,7 @@ class KubernetesNodepool < Sequel::Model
   one_to_one :strand, key: :id
   many_to_one :cluster, key: :kubernetes_cluster_id, class: :KubernetesCluster
   many_to_many :vms, order: :created_at
```
jeremyevans, line 8: This currently uses a `kubernetes_nodepools_vms` join table. That should eventually be dropped, and it should use `kubernetes_node` as a join table.

model/kubernetes/kubernetes_node.rb
```diff
@@ -0,0 +1,37 @@
...
+
+require_relative "../../model"
+
+class KubernetesNode < Sequel::Model
+  one_to_one :strand, key: :id
+  many_to_one :vm
+  many_to_one :kubernetes_cluster
+  many_to_one :kubernetes_nodepool
+
+  dataset_module Pagination
```
jeremyevans: I don't see where `Pagination` is currently used. Is this needed?

model/kubernetes/kubernetes_node.rb
```diff
@@ -0,0 +1,37 @@
...
+  plugin ResourceMethods
+  plugin SemaphoreMethods, :destroy
+
+  def display_location
+    vm.location.display_name
+  end
+
+  def path
+    "/location/#{display_location}/kubernetes-node/#{vm.name}"
+  end
```
jeremyevans: I assume these are being called somewhere, since this passes coverage testing, but I can't see where. This isn't exposed in the web/api, so it shouldn't need a path or display location AFAIK. Are these required internally for some reason? If so, maybe we should fix that.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -7,6 +7,10 @@ def vm
     @vm ||= Vm[frame["vm_id"]]
   end
 
+  def node
+    @node = KubernetesNode[frame["node_id"]]
```
jeremyevans: It looks like `node` is not called more than once in the same label internally, but it may still be better to cache this:

```suggestion
    @node ||= KubernetesNode[frame["node_id"]]
```

jeremyevans (commented):
I'll leave it to @aamederen to determine whether this is needed. If it is needed, there needs to be a plan about how to handle converting existing clusters to use kubernetes_node.

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: This was actually needed because of the condition, and it would return empty array even when there were KubernetesNodes with kubernetes_nodepool_id set to nil.

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: It will be removed in the next PR which I'm working on.

model/kubernetes/kubernetes_nodepool.rb
mohi-kalantari: It will be removed in the next PR which I'm working on.

model/kubernetes/kubernetes_node.rb
mohi-kalantari: no not needed. removed it thanks.

model/kubernetes/kubernetes_node.rb
mohi-kalantari: It's strange indeed. These were not needed and I removed them but we might need to take a look at why the tests were passed

mohi-kalantari:
@jeremyevans Thanks for the review

I'm working on another PR which is the next step in migrating from Vms to KubernetesNodes.

mohi-kalantari:
I'll finish the next one before merge/deploying this PR

model/kubernetes/kubernetes_cluster.rb
jeremyevans: Are you sure about that? The default for `:key` would be `:kubernetes_cluster_id` even if not specified. I'm not sure how it would be related to the `:conditions` option.  Please test removing the `:key` option as implemented by the suggestion, and if it causes any test failures, please post them here so I can review them.

migrate/20250819_add_kubernetes_node_model.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
```
jeremyevans:
```suggestion
  change do
```

migrate/20250819_add_kubernetes_node_model.rb
```diff
@@ -0,0 +1,17 @@
...
+      column :created_at, :timestamptz, null: false, default: Sequel::CURRENT_TIMESTAMP
+      foreign_key :vm_id, :vm, type: :uuid, null: false
+      foreign_key :kubernetes_cluster_id, :kubernetes_cluster, type: :uuid, null: false
+      foreign_key :kubernetes_nodepool_id, :kubernetes_nodepool, type: :uuid, null: true
+    end
+  end
+
+  down do
+    drop_table(:kubernetes_node)
+  end
```
jeremyevans:
```suggestion
```

jeremyevans (commented):
My comments were addressed, with the exception of the unnecessary `:key` association option.

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: You were right. I tested it and it works fine without the :key. I remember I tested it before submitting the PR and it wasn't working without the key. must have been a mistake on my side. thanks.

model/kubernetes/kubernetes_node.rb
mohi-kalantari: Fix in PR https://github.com/ubicloud/ubicloud/pull/3773

migrate/20250819_add_kubernetes_node_model.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:kubernetes_node) do
+      column :id, :uuid, primary_key: true
+      column :created_at, :timestamptz, null: false, default: Sequel::CURRENT_TIMESTAMP
```
aamederen, line 7: do we also need an updated_at column? In some models we have it (like `postgres_server`) and in some others we don't.

migrate/20250819_add_kubernetes_node_model.rb
```diff
@@ -0,0 +1,13 @@
...
+      foreign_key :vm_id, :vm, type: :uuid, null: false
+      foreign_key :kubernetes_cluster_id, :kubernetes_cluster, type: :uuid, null: false
```
aamederen, line 9: no action needed: not related to this PR, but it feels like if CP nodes would also belong to a nodepool, that could make things more neat in the codebase. Maybe something to do later.

migrate/20250819_add_kubernetes_node_model.rb
aamederen: maybe a question for @jeremyevans ?

jeremyevans: I think we should only add `created_at` and/or `updated_at` if we have a need for them. If you want to order by creation time, that is a reason to have `created_at`, and if you want to order by last update time, that is a reason to have `updated_at`. If you just want to be able to look up when an object was created at or updated at, that's a job for an audit log IMO.

PR #3762, opened 2025-08-20
Setup blob storage only when adding cache entries

velioglu:
In the current logic, blob storage bucket is destroyed if it exists but has no cache entries. Looking at the logs we noticed some buckets repeatedly getting deleted. This happens because some users only request cache entries without ever adding new ones.

To fix this, blob storage bucket will now only be created when adding a new cache entry. Subsequent get requests will still work since the blob storage will already exist once an entry is added.


routes/runtime/github.rb
```diff
@@ -6,13 +6,6 @@ class Clover
...
-    begin
-      repository.setup_blob_storage unless repository.access_key
-    rescue Excon::Error::HTTPStatus => ex
-      Clog.emit("Unable to setup blob storage") { {failed_blob_storage_setup: {ubid: runner.ubid, repository_ubid: repository.ubid, response: ex.response.body}} }
-      fail CloverError.new(400, "InvalidRequest", "unable to setup blob storage")
-    end
-
     # getCacheEntry
     r.get "cache" do
       keys, version = typecast_params.nonempty_str!(%w[keys version])
```
enescakir, line 13: Great finding!
Should we add `next 204 unless repository.access_key` here as early exit instead of running all queries?

enescakir (approved):

routes/runtime/github.rb
```diff
@@ -68,6 +61,13 @@ class Clover
     end
 
     r.on "caches" do
+      begin
```
jeremyevans: Is this the best place for the code?  This affects 3 routes under this branch:

* GET
* POST
* POST /commit

It seems like the GET wouldn't need this, only the two POST routes that access `repository.blob_storage_client`. With the way the routes are currently  structured, the simplest fix is wrapping this code in `if r.post?`.

enescakir: Actually, `POST /commit` doesn't need it either, since it already has to call `POST` first. I moved it there instead.

PR #3771, opened 2025-08-22
Use KubernetesNode instead of Vm in Kubernetes progs

mohi-kalantari:
All the logics related to k8s (provisioning, bootstrapping, upgrades) now all rely on the KuberneteNode object. The former methods for Vm are still kept in place until the code is deployed and in another commit those parts with their DB tables will be removed altogether.

----

> [!IMPORTANT]
> Replace `Vm` with `KubernetesNode` in Kubernetes logic, adding migration and updating models, programs, serializers, tests, and views.
> 
>   - **Behavior**:
>     - Replace `Vm` with `KubernetesNode` in Kubernetes provisioning, bootstrapping, and upgrade logic.
>     - Migration script `20250825_create_kubernetes_node_objects.rb` creates `KubernetesNode` objects from existing VMs.
>     - Old `Vm` methods retained temporarily for backward compatibility.
>   - **Models**:
>     - Add `cp_vms_via_nodes` and `nodes` associations to `KubernetesCluster`.
>     - Add `vms_via_nodes` association to `KubernetesNodepool`.
>     - Update `KubernetesNode` to include `vm` association and methods for SSH and billing.
>   - **Programs**:
>     - Update `kubernetes_cluster_nexus.rb`, `kubernetes_node_nexus.rb`, `kubernetes_nodepool_nexus.rb`, `provision_kubernetes_node.rb`, and `upgrade_kubernetes_node.rb` to use `KubernetesNode`.
>   - **Serializers**:
>     - Update `kubernetes_cluster.rb` and `kubernetes_nodepool.rb` to serialize `KubernetesNode` instead of `Vm`.
>   - **Tests**:
>     - Update tests in `kubernetes_cluster_spec.rb`, `kubernetes_cluster_nexus_spec.rb`, `kubernetes_node_nexus_spec.rb`, `kubernetes_nodepool_nexus_spec.rb`, `provision_kubernetes_node_spec.rb`, `upgrade_kubernetes_node_spec.rb`, `golden_files_spec.rb`, `kubernetes_cluster_spec.rb`, and `kubernetes_nodepool_spec.rb` to reflect changes.
>   - **Views**:
>     - Update `show.erb` to display `KubernetesNode` details instead of `Vm`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 6b2769eac0b927d269bf5ec5a92878e1b636d757. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/migrate/20250825_create_kubernetes_node_objects.rb
+++ b/migrate/20250825_create_kubernetes_node_objects.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    DB[:kubernetes_node].insert_ignore.insert(
+      [:id, :vm_id, :kubernetes_cluster_id],
+      DB[:kubernetes_clusters_cp_vms].select(
+        Sequel.function(:gen_random_ubid_uuid, 621).as(:id),
+        :cp_vm_id,
+        :kubernetes_cluster_id
+      )
+    )
+
+    DB[:kubernetes_node].insert_ignore.insert(
+      [:id, :vm_id, :kubernetes_cluster_id, :kubernetes_nodepool_id],
+      DB[:kubernetes_nodepools_vms]
+        .join(:kubernetes_nodepool, id: :kubernetes_nodepool_id)
+        .select(
+          Sequel.function(:gen_random_ubid_uuid, 621).as(:id),
+          :vm_id,
+          :kubernetes_cluster_id,
+          :kubernetes_nodepool_id
+        )
+    )
+
+    DB[:strand].insert_ignore.insert(
+      [:id, :prog, :label],
+      DB[:kubernetes_node]
+        .select(
+          :id,
+          "Kubernetes::KubernetesNodeNexus",
+          "wait"
+        )
+    )
+  end
+end
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -9,6 +9,7 @@ class KubernetesCluster < Sequel::Model
   many_to_one :private_subnet
   many_to_one :project
   many_to_many :cp_vms, join_table: :kubernetes_clusters_cp_vms, class: :Vm, order: :created_at
+  many_to_many :cp_vms_via_nodes, join_table: :kubernetes_node, right_key: :vm_id, class: :Vm, order: :created_at, conditions: {kubernetes_nodepool_id: nil}
   one_to_many :nodes, class: :KubernetesNode, order: :created_at, conditions: {kubernetes_nodepool_id: nil}
   one_to_many :nodepools, class: :KubernetesNodepool
   one_to_many :active_billing_records, class: :BillingRecord, key: :resource_id, &:active
@@ -45,12 +46,12 @@ class KubernetesCluster < Sequel::Model
     api_server_lb.hostname
   end
 
-  def client(session: cp_vms.first.sshable.connect)
+  def client(session: sshable.connect)
     Kubernetes::Client.new(self, session)
   end
 
   def sshable
-    cp_vms.first.sshable
+    cp_vms_via_nodes.first.sshable
   end
 
   def services_load_balancer_name
@@ -74,11 +75,11 @@ class KubernetesCluster < Sequel::Model
   end
 
   def kubeconfig
-    self.class.kubeconfig(cp_vms.first)
+    self.class.kubeconfig(cp_vms_via_nodes.first)
   end
 
   def vm_diff_for_lb(load_balancer)
-    worker_vms = nodepools.flat_map(&:vms)
+    worker_vms = nodepools.flat_map(&:vms_via_nodes)
     worker_vm_ids = worker_vms.map(&:id).to_set
     lb_vms = load_balancer.load_balancer_vms.map(&:vm)
     lb_vm_ids = lb_vms.map(&:id).to_set
@@ -112,12 +113,12 @@ class KubernetesCluster < Sequel::Model
     aggregate_readings(previous_pulse: previous_pulse, reading: reading)
   end
 
-  def all_vms
-    cp_vms + nodepools.flat_map(&:vms)
+  def all_nodes
+    nodes + nodepools.flat_map(&:nodes)
   end
 
   def worker_vms
-    nodepools.flat_map(&:vms)
+    nodepools.flat_map(&:vms_via_nodes)
   end
 end
 
--- a/model/kubernetes/kubernetes_node.rb
+++ b/model/kubernetes/kubernetes_node.rb
@@ -10,6 +10,35 @@ class KubernetesNode < Sequel::Model
 
   plugin ResourceMethods
   plugin SemaphoreMethods, :destroy
+
+  def sshable
+    vm.sshable
+  end
+
+  def billing_record
+    if kubernetes_nodepool
+      worker_billing_records
+    else
+      control_plane_billing_record
+    end
+  end
+
+  def worker_billing_records
+    [
+      {type: "KubernetesWorkerVCpu", family: vm.family, amount: vm.vcpus},
+      {type: "KubernetesWorkerStorage", family: "standard", amount: vm.storage_size_gib}
+    ]
+  end
+
+  def control_plane_billing_record
+    [
+      {type: "KubernetesControlPlaneVCpu", family: vm.family, amount: vm.vcpus}
+    ]
+  end
+
+  def name
+    vm.name
+  end
 end
 
 # Table: kubernetes_node
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -96,21 +96,21 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
 
     services_lb.dns_zone&.insert_record(record_name: "*.#{services_lb.hostname}.", type: "CNAME", ttl: 3600, data: "#{services_lb.hostname}.")
 
-    hop_bootstrap_control_plane_vms
+    hop_bootstrap_control_plane_nodes
   end
 
-  label def bootstrap_control_plane_vms
+  label def bootstrap_control_plane_nodes
     nap 5 unless kubernetes_cluster.endpoint
 
     # In 1-node control plane setup, we will wait until it's over
     # In 3-node control plane setup, we start the bootstrapping after
     # the first CP bootstrap
     ready_to_bootstrap_workers =
-      kubernetes_cluster.cp_vms.count >= kubernetes_cluster.cp_node_count ||
-      (kubernetes_cluster.cp_node_count == 3 && kubernetes_cluster.cp_vms.count == 1)
+      kubernetes_cluster.nodes.count >= kubernetes_cluster.cp_node_count ||
+      (kubernetes_cluster.cp_node_count == 3 && kubernetes_cluster.nodes.count == 1)
     kubernetes_cluster.nodepools.each(&:incr_start_bootstrapping) if ready_to_bootstrap_workers
 
-    hop_wait_nodes if kubernetes_cluster.cp_vms.count >= kubernetes_cluster.cp_node_count
+    hop_wait_nodes if kubernetes_cluster.nodes.count >= kubernetes_cluster.cp_node_count
 
     bud Prog::Kubernetes::ProvisionKubernetesNode, {"subject_id" => kubernetes_cluster.id}
 
@@ -118,7 +118,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
   end
 
   label def wait_control_plane_node
-    reap(:bootstrap_control_plane_vms)
+    reap(:bootstrap_control_plane_nodes)
   end
 
   label def wait_nodes
@@ -127,14 +127,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
   end
 
   label def create_billing_records
-    records =
-      kubernetes_cluster.cp_vms.map { {type: "KubernetesControlPlaneVCpu", family: it.family, amount: it.vcpus} } +
-      kubernetes_cluster.nodepools.flat_map(&:vms).flat_map {
-        [
-          {type: "KubernetesWorkerVCpu", family: it.family, amount: it.vcpus},
-          {type: "KubernetesWorkerStorage", family: "standard", amount: it.storage_size_gib}
-        ]
-      }
+    records = kubernetes_cluster.all_nodes.flat_map(&:billing_record)
 
     records.each do |record|
       BillingRecord.create(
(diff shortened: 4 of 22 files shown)


model/kubernetes/kubernetes_cluster.rb
```diff
@@ -45,12 +45,12 @@ def endpoint
     api_server_lb.hostname
   end
 
-  def client(session: cp_vms.first.sshable.connect)
+  def client(session: nodes.first.sshable.connect)
```
```suggestion
  def client(session: sshable.connect)
```

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -112,12 +112,12 @@ def check_pulse(session:, previous_pulse:)
     aggregate_readings(previous_pulse: previous_pulse, reading: reading)
   end
 
-  def all_vms
-    cp_vms + nodepools.flat_map(&:vms)
+  def all_nodes
+    nodes + nodepools.flat_map(&:nodes)
```
jeremyevans, line 117:
```suggestion
    nodes + worker_nodes
```

Again, I recommend you rename the association from `nodes` to `cp_nodes`.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -3,10 +3,6 @@
 class Prog::Kubernetes::ProvisionKubernetesNode < Prog::Base
   subject_is :kubernetes_cluster
 
-  def vm
-    @vm ||= Vm[frame["vm_id"]]
-  end
-
```
jeremyevans, line 9: I would keep this method and have it be `@vm ||= node.vm`. That would element the need to switch from `vm` to `node.vm` in other methods.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -178,18 +173,18 @@ def before_run
   label def upgrade
     decr_upgrade
 
-    node_to_upgrade = kubernetes_cluster.cp_vms.find do |vm|
-      vm_version = kubernetes_cluster.client(session: vm.sshable.connect).version
-      vm_minor_version = vm_version.match(/^v\d+\.(\d+)$/)&.captures&.first&.to_i
+    node_to_upgrade = kubernetes_cluster.nodes.find do |node|
+      node_version = kubernetes_cluster.client(session: node.sshable.connect).version
```
jeremyevans, line 170: I think these changes could be avoided by using `vm = node.vm`. Similar for other cases in these progs. In general, I recommend against attempting to push everything through KubernetesNode, when all of the actual operations run on the underlying Vm. Only use KubernetesNode to the extent it is actually needed.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -45,12 +45,12 @@ def endpoint
     api_server_lb.hostname
   end
 
-  def client(session: cp_vms.first.sshable.connect)
+  def client(session: nodes.first.sshable.connect)
     Kubernetes::Client.new(self, session)
   end
 
   def sshable
-    cp_vms.first.sshable
```
jeremyevans, line 53: I recommend defining the `cp_vms` association as a many_to_many with a `kubernetes_node` join table. That may reduce the number of cases where you need to make changes.

mohi-kalantari: I didn't like the many_to_many approach as it seemed to be a workaround of not having a dedicated field for a foreign key to kubernetes cluster/nodepool in Vm model. but having the model, we can have the FK and we wouldn't need two separate tables for the many_to_many assoication.

prog/kubernetes/kubernetes_cluster_nexus.rb
mohi-kalantari: Semantically, node_version makes more sense in terms of k8s terminology as vm_version can be more interpretted as the os version.

jeremyevans: Seems like it's a good idea to change to `node_version` regardless of KubernetesNode changes then. However, unless there is a similar semantic reason to change from `vm` to `node` in other cases, I recommend against it.

model/kubernetes/kubernetes_cluster.rb
jeremyevans: I don't understand this comment. From a database perspective, you can use kubernetes_node as a join table just as you were using kubernetes_clusters_cp_vms table, since both have foreign keys to kubernetes_cluster and vm. If you only need to access vm information and not node information, then there doesn't seem to be a reason to get the nodes and then the vm for each node, as opposed to getting the vms directly.

As far as I can see, you are kubernetes_node for only 2 reasons:

* Cleanup (separate strand with destroy label)
* Billing records

For all other cases, you should operate on the vm node directly, instead of indirectly. The many_to_many association allows you to do that.

mohi-kalantari: Got it. will take another look. I'd also want to avoid two join tables as we have right now. (kubernetes_cluster.cp_vms and node_pool.vms associations)

Introducing KubernetesNode was mainly because of an independent model which would host the node related logics like drain and keeping the state. The PR we see currently don't make that much sense because the main reason for change is not implemented yet.

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+          Sequel[:kubernetes_cluster][:id].as(:kubernetes_cluster_id),
+          Sequel[:kubernetes_nodepool][:id].as(:kubernetes_nodepool_id)
+        )
+    )
+
+    DB[:strand].import(
+      [:id, :prog, :label],
+      DB[:kubernetes_node]
+        .select(
+          Sequel[:kubernetes_node][:id],
```
jeremyevans: Qualification is not needed, there is only one table.

```suggestion
          :id,
```

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+      [:id, :vm_id, :kubernetes_cluster_id, :kubernetes_nodepool_id],
+      DB[:kubernetes_cluster]
+        .join(:kubernetes_nodepool, kubernetes_cluster_id: :id)
+        .join(:kubernetes_nodepools_vms, kubernetes_nodepool_id: :id)
+        .join(:vm, id: :vm_id)
+        .select(
+          Sequel.function(:gen_timestamp_ubid_uuid, 621).as(:id),
+          Sequel[:vm][:id].as(:vm_id),
+          Sequel[:kubernetes_cluster][:id].as(:kubernetes_cluster_id),
+          Sequel[:kubernetes_nodepool][:id].as(:kubernetes_nodepool_id)
```
jeremyevans: This might be simpler:

```suggestion
          :vm_id,
          :kubernetes_cluster_id,
          :kubernetes_nodepool_id
```

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+  up do
+    DB[:kubernetes_node].import(
+      [:id, :vm_id, :kubernetes_cluster_id],
+      DB[:kubernetes_cluster]
+        .join(:kubernetes_clusters_cp_vms, kubernetes_cluster_id: :id)
+        .join(:vm, id: :cp_vm_id)
+        .select(
+          Sequel.function(:gen_timestamp_ubid_uuid, 621).as(:id),
+          Sequel[:vm][:id].as(:vm_id),
+          Sequel[:kubernetes_cluster][:id].as(:kubernetes_cluster_id)
```
jeremyevans:
```suggestion
          :cp_vm_id,
          :kubernetes_cluster_id
```

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+
+Sequel.migration do
+  up do
+    DB[:kubernetes_node].import(
+      [:id, :vm_id, :kubernetes_cluster_id],
+      DB[:kubernetes_cluster]
+        .join(:kubernetes_clusters_cp_vms, kubernetes_cluster_id: :id)
+        .join(:vm, id: :cp_vm_id)
+        .select(
+          Sequel.function(:gen_timestamp_ubid_uuid, 621).as(:id),
```
jeremyevans:
```suggestion
          Sequel.function(:gen_random_ubid_uuid, 621).as(:id),
```

The only classes that use timestamp ubids are Strand and Semaphore, see `UBID.generate` in `ubid.rb`.

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    DB[:kubernetes_node].import(
+      [:id, :vm_id, :kubernetes_cluster_id, :kubernetes_nodepool_id],
+      DB[:kubernetes_cluster]
+        .join(:kubernetes_nodepool, kubernetes_cluster_id: :id)
+        .join(:kubernetes_nodepools_vms, kubernetes_nodepool_id: :id)
+        .join(:vm, id: :vm_id)
+        .select(
+          Sequel.function(:gen_timestamp_ubid_uuid, 621).as(:id),
```
jeremyevans:
```suggestion
          Sequel.function(:gen_random_ubid_uuid, 621).as(:id),
```

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    DB[:kubernetes_node].import(
+      [:id, :vm_id, :kubernetes_cluster_id],
+      DB[:kubernetes_cluster]
+        .join(:kubernetes_clusters_cp_vms, kubernetes_cluster_id: :id)
+        .join(:vm, id: :cp_vm_id)
```
jeremyevans:
```suggestion
      DB[:kubernetes_clusters_cp_vms]
```

You don't need the columns from the other tables.

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+          Sequel[:kubernetes_cluster][:id].as(:kubernetes_cluster_id)
+        )
+    )
+
+    DB[:kubernetes_node].import(
+      [:id, :vm_id, :kubernetes_cluster_id, :kubernetes_nodepool_id],
+      DB[:kubernetes_cluster]
+        .join(:kubernetes_nodepool, kubernetes_cluster_id: :id)
+        .join(:kubernetes_nodepools_vms, kubernetes_nodepool_id: :id)
+        .join(:vm, id: :vm_id)
```
jeremyevans:
```suggestion
      DB[:kubernetes_nodepools_vms]
        .join(:kubernetes_nodepool, id: kubernetes_nodepool_id)
```

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    DB[:kubernetes_node].import(
```
jeremyevans:
```suggestion
    DB[:kubernetes_node].insert(
```

`import` works, but it just calls `insert` if the second argument is a dataset.

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+        .join(:kubernetes_clusters_cp_vms, kubernetes_cluster_id: :id)
+        .join(:vm, id: :cp_vm_id)
+        .select(
+          Sequel.function(:gen_timestamp_ubid_uuid, 621).as(:id),
+          Sequel[:vm][:id].as(:vm_id),
+          Sequel[:kubernetes_cluster][:id].as(:kubernetes_cluster_id)
+        )
+    )
+
+    DB[:kubernetes_node].import(
```
jeremyevans:
```suggestion
    DB[:kubernetes_node].insert(
```

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,41 @@
...
+        .join(:vm, id: :vm_id)
+        .select(
+          Sequel.function(:gen_timestamp_ubid_uuid, 621).as(:id),
+          Sequel[:vm][:id].as(:vm_id),
+          Sequel[:kubernetes_cluster][:id].as(:kubernetes_cluster_id),
+          Sequel[:kubernetes_nodepool][:id].as(:kubernetes_nodepool_id)
+        )
+    )
+
+    DB[:strand].import(
```
jeremyevans:
```suggestion
    DB[:strand].insert(
```

migrate/20250825_create_kubernetes_node_objects.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
```
aamederen, line 4: nit: personally, I would approach this migration this way:
- Change the code so that it handles the mixture of direct VM associations and Node associations
- perform the migration via hand on a pry console with visually reviewing the 
- Change the code to remove old stuff.

however, this is also pretty OK as it's been tested and reviewed.

model/kubernetes/kubernetes_node.rb
```diff
@@ -10,6 +10,25 @@ class KubernetesNode < Sequel::Model
 
   plugin ResourceMethods
   plugin SemaphoreMethods, :destroy
+
+  def sshable
+    vm.sshable
+  end
+
+  def worker_billing_records
```
aamederen, line 26: nit: why not have a single billing_record function that returns CP billing record or worker billing record depending on whether there's a nodepool associated or not?

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -96,29 +96,29 @@ def before_run
...
+  label def bootstrap_control_plane_nodes
     nap 5 unless kubernetes_cluster.endpoint
 
     # In 1-node control plane setup, we will wait until it's over
     # In 3-node control plane setup, we start the bootstrapping after
     # the first CP bootstrap
     ready_to_bootstrap_workers =
-      kubernetes_cluster.cp_vms.count >= kubernetes_cluster.cp_node_count ||
-      (kubernetes_cluster.cp_node_count == 3 && kubernetes_cluster.cp_vms.count == 1)
+      kubernetes_cluster.nodes.count >= kubernetes_cluster.cp_node_count ||
```
aamederen: can't we simplify this condition? If there's a ready CP node, just hit the worker provisioning, withouth checking the cp_node_count

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: Fixed. thanks.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -102,12 +102,7 @@ def before_run
   label def bootstrap_control_plane_nodes
     nap 5 unless kubernetes_cluster.endpoint
 
-    # In 1-node control plane setup, we will wait until it's over
-    # In 3-node control plane setup, we start the bootstrapping after
-    # the first CP bootstrap
-    ready_to_bootstrap_workers =
-      kubernetes_cluster.nodes.count >= kubernetes_cluster.cp_node_count ||
-      (kubernetes_cluster.cp_node_count == 3 && kubernetes_cluster.nodes.count == 1)
+    ready_to_bootstrap_workers = kubernetes_cluster.nodes.count >= 1
```
aamederen, line 105: This might start creating workers immediately, before the node is ready, right?

PR #3773, opened 2025-08-22
Fix coverage filters

jeremyevans:
The preceding slash in the string doesn't work the way you would expect. add_filter will exclude all files that include the string in the path. So "/kubernetes" filtered out most of our kubernetes code.

This exposed missing coverage in the sdk base model, so this includes a commit to fix that, as well as a fix for an sdk issue I found when adding coverage.

----

> [!IMPORTANT]
> Fix coverage filters and improve SDK model handling in `spec/coverage_helper.rb` and `sdk/ruby/lib/ubicloud/model.rb`, with additional tests in `spec/ruby_sdk_spec.rb`.
> 
>   - **Coverage Filters**:
>     - Updated `add_filter` logic in `spec/coverage_helper.rb` to correctly exclude files by removing preceding slashes.
>     - Consolidated filter conditions into a single regex match.
>   - **SDK Model Fixes**:
>     - Fixed missing value handling in `set_columns` method in `sdk/ruby/lib/ubicloud/model.rb`.
>     - Replaced `Dir.open` with `Dir.glob` for requiring model subclasses.
>   - **Tests**:
>     - Added tests in `spec/ruby_sdk_spec.rb` for `Model.new`, `Model.list`, and `Model#id` to cover edge cases and ensure correct error handling.
>     - Added tests for `Firewall#detach_subnet` to handle invalid subnet IDs.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 41349f70e07c0e1b4c665ac9a1abe814ed1da686. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/coverage_helper.rb
+++ b/spec/coverage_helper.rb
@@ -20,17 +20,10 @@ if (suite = ENV.delete("COVERAGE"))
         !LOCKED_FILES.include?(file.filename)
       end
     else
-      add_filter "/rhizome"
-      add_filter "/kubernetes"
-      # No need to check coverage for them
-      add_filter "/migrate/"
-      add_filter "/spec/"
-      add_filter "/db.rb"
-      add_filter "/model.rb"
-      add_filter "/loader.rb"
-      add_filter "/.env.rb"
-      add_filter "/var/"
-      add_filter "/vendor/"
+      add_filter do |file|
+        path = file.filename.delete_prefix(File.dirname(__dir__))
+        path.match?(/\A\/(rhizome|kubernetes|migrate|spec|var|vendor|(db|model|loader|\.env)\.rb)/)
+      end
     end
 
     add_group("Missing") { |src| src.covered_percent < 100 }


fdr (approved):

fdr:
@jeremyevans does the kubernetes csi get 100% as part of CI? I can't quite remember if that got automated, even though we got there. It's a fast test but doesn't have dependencies like the autoloader or postgres, so it can be run independently, and perhaps might work better if it did. But, I don't see it in ci.yml. Thus, hard work improving the test suite can slip, as patches come in and it's hard to tell what revision opened what gaps in the testing.

jeremyevans:
I'm not sure how/whether kubernetes csi is currently CI tested.

If you want to run the kubernetes csi tests independently of the normal CI tests, but still have coverage testing, we would need to keep the filter for the normal CI tests, and use a different filter to the kubernetes csi tests.

fdr:
> I'm not sure how/whether kubernetes csi is currently CI tested.
> 
> If you want to run the kubernetes csi tests independently of the normal CI tests, but still have coverage testing, we would need to keep the filter for the normal CI tests, and use a different filter to the kubernetes csi tests.

I think it could go either way. It's a bit silly when developing to start postgres or get the autoloader and Config, but in principle, it shouldn't hurt anything either.

jeremyevans:
I looked into it and we run the csi specs as part of the normal CI workflow, but in a separate process. One reason for doing this is bundler. csi has a different Gemfile, which we wouldn't want to combine into the main Gemfile. I wouldn't want clover depending on the grpc gem (which I've heard has more lines of code than ruby itself). 

I agree it would be best to coverage test the csi code. I'll work on a PR for that.

jeremyevans:
> I agree it would be best to coverage test the csi code. I'll work on a PR for that.

We already have coverage for that directly in kubernetes/csi/spec/spec_helper.rb. So I don't think anything needs to be done for that.

PR #3777, opened 2025-08-22
Make missing openapi.yml entries raise in tests

jeremyevans:
This greatly improves the debugging experience when dealing with the api if you forget to add the necessary openapi.yml entries. Without this, you just get a 404, and unless you remember that this is caused by committee, you'll likely waste minutes trying to figure out where the problem is.

CC @geemus


clover.rb
```diff
@@ -933,7 +933,12 @@ def locked_domain_for(email)
         @schema_validator = SCHEMA_ROUTER.build_schema_validator(r)
         @schema_validator.request_validate(r)
 
-        next unless @schema_validator.link_exist?
+        unless @schema_validator.link_exist?
+          if Config.test? && !ENV["IGNORE_INVALID_API_PATHS"]
+            raise "request not allowed by committee: #{r.request_method} #{r.path_info}"
```
geemus: Definitely an accurate statement, though I suspect something like this might be more instructive?
```suggestion
            raise "request not found in openapi schema: #{r.request_method} #{r.path_info}"
```

You could certainly note that committee was what couldn't find it too if you wanted. I suspect that this might be helpful in pointing directly to where action probably needs to be taken the vast majority of the time.

geemus (approved):
This definitely seems like a pretty drastic improvement, wish I would have realized/thought of it earlier to avoid the pain for you all. Looks like a good approach overall, just made a minor suggestion about potential tweaks to the error message.

PR #3781, opened 2025-08-25
Download boot images from R2

enescakir:
- **Allow downloading boot images from Cloudflare R2**
  We use our internal MinIO cluster to distribute images to our fleet. It
  has a unique use case where we download a single large file repeatedly,
  which impacts cluster performance and degrades the experience for other
  users.
  
  We decided to move image distribution to Cloudflare R2. This commit
  enables generating presigned URLs for R2.
  

- **Use htcat to download image from presigned URL**
    We previously used curl to download images from presigned URLs. This
    worked well for our internal MinIO cluster, but with R2 presigned URLs,
    curl performs poorly since it downloads with a single connection.

    First, I benchmarked aria2 [^1], which supports multiple connections.
    However, Daniel recommended using a memory-safe tool instead and
    suggested htcat [^2], which his team wrote about 10 years ago. After a
    few small improvements, htcat now works well for our use case and
    delivers performance comparable to aria2.

    If the URL is from R2, we now use htcat to download it with multiple
    connections.

    Since htcat pipes to stdout, we calculate the sha256 hash while
    downloading similar to how we did with curl.

    I will download the htcat binary on existing hosts.

    [^1]: https://aria2.github.io/
    [^2]: https://github.com/htcat/htcat  

- **Download images from Cloudflare R2 for E2E tests**
  

----

> [!IMPORTANT]
> Enable downloading boot images from Cloudflare R2 using `htcat` for improved performance and update E2E tests accordingly.
> 
>   - **Behavior**:
>     - Enable downloading boot images from Cloudflare R2 by generating presigned URLs in `download_boot_image.rb`.
>     - Switch from `curl` to `htcat` for downloading images from R2 in `boot_image.rb` and `download-boot-image`.
>     - Update E2E tests to download images from R2 in `e2e.yml`.
>   - **Configuration**:
>     - Add R2 configuration options to `config.rb`.
>     - Update `vm_host.rb` to include `download_r2` parameter in `download_boot_image()`.
>   - **Implementation**:
>     - Add `r2_signed_url()` method in `download_boot_image.rb` to generate R2 presigned URLs.
>     - Modify `download()` in `boot_image.rb` to use `htcat` when `use_htcat` is true.
>   - **Testing**:
>     - Add tests for R2 URL generation and `htcat` download in `boot_image_spec.rb` and `download_boot_image_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for bcd195e53a2d4135b17350107eb6e5ebdf5e8d39. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


prog/download_boot_image.rb
```diff
@@ -23,6 +24,10 @@ def default_boot_image_version(image_name)
     Config.send(config_name)
   end
 
+  def download_from_r2?
+    frame["download_r2"] || Config.is_e2e
```
jeremyevans, line 28: At some point, we should consider removing the `|| Config.is_e2e`, but it seems fine if it is temporary.

rhizome/host/bin/prep_host.rb
```diff
@@ -92,6 +92,9 @@
 # We need smartmontools, nvme-cli and jq in order to probe the disk status of VmHosts
 r "apt-get -y install smartmontools nvme-cli jq"
 
+# aria2 is able to download a file from signed URLs concurrently
+r "apt-get -y install aria2"
```
jeremyevans:
```suggestion
r "apt-get -y install aria2" if use_aria2
```

Unless there is a reason to install it if it won't be used?

rhizome/host/lib/boot_image.rb
```diff
@@ -77,6 +81,15 @@ def curl_image(url, temp_path, ca_path)
     sha256_sum
   end
 
+  def aria2_image(url, temp_path, sha256sum)
+    sha256_arg = sha256sum ? " --checksum=sha-256=#{sha256sum.shellescape}" : ""
```
jeremyevans: `sha256sum` should always be in hex, and therefore shouldn't need to be escaped. Not a problem if you want to keep the `shellescape` as a failsafe, though.

rhizome/host/lib/boot_image.rb
```diff
@@ -77,6 +81,15 @@ def curl_image(url, temp_path, ca_path)
...
+    dirname = File.dirname(temp_path)
+    filename = File.basename(temp_path)
```
jeremyevans:
```suggestion
    dirname, filename = File.split(temp_path)
```

jeremyevans (approved):

prog/download_boot_image.rb
enescakir: Yeah, we’ll likely make it the default at some point. Without `|| Config.is_e2e`, I need to add some data to host nexus' stack to download from R2 in E2E, which I don’t want to modify just for it.

rhizome/host/bin/prep_host.rb
enescakir: We’ll make downloading from R2 the default after ensuring its reliability. Therefore, it’s better to install `aria2` on all new hosts.

rhizome/host/lib/boot_image.rb
enescakir: That’s a cool. I didn't know about `File.split`.

rhizome/host/bin/prep_host.rb
fdr: I've kinda avoided aria2 because it's kinda complicated and I wonder about the supply chain, but to be fair I had little to gain at the time.

I actually wrote https://github.com/htcat/htcat/ a long time ago and it was one of my points of comparison.  Unlike aria2 of that era, it supported pipelining. Unfortunately the debian package only installs the go library, not the binary.

enescakir: After @fdr done some improvements on htcat, now it has performance as aria2 https://github.com/htcat/htcat/tree/next

enescakir: I forked it and created release. https://github.com/ubicloud/htcat/releases/tag/v2.0.0-ubi1

When the upstream has binaries for the new version, we can switch to it.

enescakir: I updated this PR to use `htcat` instead of `aria2`

enescakir:
> I would like to get rid of the conditionals between aria and curl

It’s still in the evaluation stage, with MinIO remaining the default. However, if we’re satisfied with R2, we can eventually remove the multiple image download options.

rhizome/host/bin/prep_host.rb
```diff
@@ -92,6 +93,10 @@
 # We need smartmontools, nvme-cli and jq in order to probe the disk status of VmHosts
 r "apt-get -y install smartmontools nvme-cli jq"
 
+# htcat is able to download a file from signed URLs concurrently
+r "wget https://github.com/ubicloud/htcat/releases/download/v2.0.0-ubi1/htcat_2.0.0-ubi1_linux_#{Arch.render(x64: "amd64", arm64: "arm64")}.tar.gz -O htcat.tar.gz"
```
velioglu, line 97: nit: We might consider downloading the latest release

enescakir: We don’t plan to release new versions of htcat frequently. I also prefer to keep the version hardcoded so we always know which one is running on the fleet.

PR #3789, opened 2025-08-26
Allow scaling of KubernetesNodepools

mohi-kalantari:
Add migration to add state to KubernetesNode. This will be used in an association to track draining nodes.

New semaphores are added to allow the KubernetesNodeNexus take care of draining and removing nodes from the cluster.

Also right now we can increase the number of nodes easily in a kubernetes cluster

----

> [!IMPORTANT]
> This PR enables scaling of Kubernetes node pools by adding node state management, scaling logic, and node draining/removal processes, with comprehensive tests.
> 
>   - **Behavior**:
>     - Adds `state` column to `kubernetes_node` table with default `active` state in `20250829_add_state_to_kubernetes_node.rb`.
>     - Implements node scaling in `KubernetesNodepoolNexus` by adjusting node count and retiring excess nodes.
>     - Introduces node draining and removal logic in `KubernetesNodeNexus`.
>   - **Models**:
>     - Updates `KubernetesNode` to include `state` and use `SemaphoreMethods` for `retire`.
>     - Adds `functional_nodes` association to `KubernetesNodepool` for active nodes.
>   - **Programs**:
>     - Adds `retire`, `drain`, and `remove_node_from_cluster` labels to `KubernetesNodeNexus`.
>     - Modifies `KubernetesNodepoolNexus` to handle scaling with `scale_worker_count` semaphore.
>     - Updates `UpgradeKubernetesNode` to retire old nodes and wait for drain completion.
>   - **Tests**:
>     - Adds tests for node draining and removal in `kubernetes_node_nexus_spec.rb`.
>     - Adds tests for nodepool scaling logic in `kubernetes_nodepool_nexus_spec.rb`.
>     - Updates `upgrade_kubernetes_node_spec.rb` to test node retirement and drain waiting.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for e082fe64994394180d5bd26ede07b3cdc9bdc2e5. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/kubernetes/kubernetes_node_nexus.rb
+++ b/prog/kubernetes/kubernetes_node_nexus.rb
@@ -21,6 +21,14 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
     end
   end
 
+  def cluster
+    kubernetes_node.kubernetes_cluster
+  end
+
+  def nodepool
+    kubernetes_node.kubernetes_nodepool
+  end
+
   label def start
     hop_wait
   end
@@ -33,8 +41,13 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
   end
 
   label def retire
+    kubernetes_node.update(state: "draining")
+    hop_drain
+  end
+
+  label def drain
     unit_name = "drain_node_#{kubernetes_node.name}"
-    sshable = kubernetes_node.kubernetes_cluster.sshable
+    sshable = cluster.cp_vms.last.sshable
     case sshable.d_check(unit_name)
     when "Succeeded"
       hop_remove_node_from_cluster
@@ -49,12 +62,22 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
       nap 10
     else
       register_deadline("destroy", 0)
-      nap(60 * 60 * 24)
+      nap 3 * 60 * 60
     end
   end
 
   label def remove_node_from_cluster
-    kubernetes_node.kubernetes_cluster.client.delete_node(kubernetes_node.name)
+    # kubeadm reset is necessary for etcd member removal, deleting the node itself
+    # won't remove the node from the etcd cluster, hurting the etcd cluster health
+    kubernetes_node.sshable.cmd("sudo kubeadm reset --force")
+
+    if nodepool
+      cluster.services_lb.detach_vm(kubernetes_node.vm)
+    else
+      cluster.api_server_lb.detach_vm(kubernetes_node.vm)
+    end
+
+    cluster.client(session: cluster.nodes.last.sshable.connect).delete_node(kubernetes_node.name)
     hop_destroy
   end
 
--- a/prog/kubernetes/kubernetes_nodepool_nexus.rb
+++ b/prog/kubernetes/kubernetes_nodepool_nexus.rb
@@ -43,10 +43,7 @@ class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
       end
     elsif current_node_count > desired_node_count
       excess_nodes = kubernetes_nodepool.functional_nodes.first(current_node_count - desired_node_count)
-      excess_nodes.each {
-        it.update(state: "draining")
-        it.incr_retire
-      }
+      excess_nodes.each(&:incr_retire)
     end
     hop_wait_worker_node
   end
--- a/prog/kubernetes/upgrade_kubernetes_node.rb
+++ b/prog/kubernetes/upgrade_kubernetes_node.rb
@@ -7,10 +7,6 @@ class Prog::Kubernetes::UpgradeKubernetesNode < Prog::Base
     @old_node ||= KubernetesNode[frame.fetch("old_node_id")]
   end
 
-  def new_node
-    @new_node ||= KubernetesNode[frame.fetch("new_node_id")]
-  end
-
   def kubernetes_nodepool
     @kubernetes_nodepool ||= KubernetesNodepool[frame.fetch("nodepool_id", nil)]
   end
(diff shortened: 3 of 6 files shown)


migrate/20250826_add_state_to_kubernetes_node.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
```
jeremyevans:
```suggestion
  change do
```

migrate/20250826_add_state_to_kubernetes_node.rb
```diff
@@ -0,0 +1,15 @@
...
+    alter_table(:kubernetes_node) do
+      add_column :state, :text, null: false, default: "active"
+    end
+  end
+
+  down do
+    alter_table(:kubernetes_node) do
+      drop_column :state
+    end
+  end
```
jeremyevans:
```suggestion
```

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -34,8 +34,21 @@ def before_run
...
+    desired_node_count = kubernetes_nodepool.node_count
+
+    if current_node_count < desired_node_count
+      (desired_node_count - current_node_count).times do
+        bud Prog::Kubernetes::ProvisionKubernetesNode, {"nodepool_id" => kubernetes_nodepool.id, "subject_id" => kubernetes_nodepool.kubernetes_cluster_id}
+      end
+    elsif current_node_count > desired_node_count
+      excess_nodes = kubernetes_nodepool.functional_nodes.first(current_node_count - desired_node_count)
+      DB.transaction do
+        excess_nodes.each {
```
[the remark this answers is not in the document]
jeremyevans: Using `do`/`end` blocks for multi-line blocks is preferred, but using `it` is fine for shorter blocks (3 lines or less).

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -26,9 +26,39 @@ def before_run
...
+      nap 10
+    else
+      register_deadline("destroy", 0)
+      nap(60 * 60 * 24)
+    end
+  end
+
+  label def remove_node_from_cluster
+    kubernetes_node.kubernetes_cluster.client.delete_node(kubernetes_node.name)
+    kubernetes_node.kubernetes_nodepool&.remove_vm(kubernetes_node.vm)
```
mohi-kalantari: If this PR https://github.com/ubicloud/ubicloud/pull/3806 is merged first, I'll remove this line

migrate/20250826_add_state_to_kubernetes_node.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:kubernetes_node) do
+      add_column :state, :text, null: false, default: "active"
```
aamederen: Let's make this an enum, similar to VmHost allocation_state.

[not needed for this PR] it will be a good idea to use this new column during provisioning (first, the state is "creating", then "active", then "retiring" and maybe some "unhealthy" when we have better monitors. This would also be a good source for display state.

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -26,9 +26,39 @@ def before_run
...
 
   label def wait
+    when_retire_set? do
+      hop_retire
+    end
     nap 6 * 60 * 60
   end
 
+  label def retire
+    unit_name = "drain_node_#{kubernetes_node.name}"
```
aamederen, line 49: this is a repetition with the upgrade step. We should extract the retire process into another prog and reuse it here and during upgrade OR use this retire method in the upgrade code.

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -34,8 +34,21 @@ def before_run
...
-      bud Prog::Kubernetes::ProvisionKubernetesNode, {"nodepool_id" => kubernetes_nodepool.id, "subject_id" => kubernetes_nodepool.kubernetes_cluster_id}
+    current_node_count = kubernetes_nodepool.functional_nodes.count
+    desired_node_count = kubernetes_nodepool.node_count
+
+    if current_node_count < desired_node_count
+      (desired_node_count - current_node_count).times do
+        bud Prog::Kubernetes::ProvisionKubernetesNode, {"nodepool_id" => kubernetes_nodepool.id, "subject_id" => kubernetes_nodepool.kubernetes_cluster_id}
+      end
+    elsif current_node_count > desired_node_count
+      excess_nodes = kubernetes_nodepool.functional_nodes.first(current_node_count - desired_node_count)
```
aamederen, line 45: if we extract draining into a prog and use it here, then we can also rely on `wait_worker_node` to wait until the drain is finished.

aamederen (changes requested):
I left a few comments that I think should be addressed before merging the PR, but it's OK to postpone/delegate(to me) some of them as well.

prog/kubernetes/kubernetes_node_nexus.rb
mohi-kalantari: Yeah I was thinking about the latter.

migrate/20250826_add_state_to_kubernetes_node.rb
jeremyevans: No more PostgreSQL enums, as it is not possible to remove enum members (results in an irreversible migration). If you want to restrict to only valid states, use a foreign key to another table holding state names, or a check constraint.

prog/kubernetes/upgrade_kubernetes_node.rb
```diff
@@ -52,30 +52,21 @@ def before_run
...
-    when "NotStarted"
-      vm.sshable.d_run("drain_node", "sudo", "kubectl", "--kubeconfig=/etc/kubernetes/admin.conf",
-        "drain", old_node.name, "--ignore-daemonsets", "--delete-emptydir-data")
-      nap 10
-    when "InProgress"
-      nap 10
-    when "Failed"
-      vm.sshable.d_restart("drain_node")
-      nap 10
+    DB.transaction do
```
aamederen: transaction is not needed inside FSM state definitions.

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -34,8 +34,21 @@ def before_run
...
+      DB.transaction do
```
aamederen: similarly, we can remove the transaction here

prog/kubernetes/upgrade_kubernetes_node.rb
```diff
@@ -52,30 +52,21 @@ def before_run
...
+      old_node.update(state: "draining")
```
aamederen: IMO, the state changes should be done in the Kubernetes Node Nexus

model/kubernetes/kubernetes_node.rb
```diff
@@ -9,7 +9,7 @@ class KubernetesNode < Sequel::Model
   many_to_one :kubernetes_nodepool
 
   plugin ResourceMethods
-  plugin SemaphoreMethods, :destroy
+  plugin SemaphoreMethods, :destroy, :retire, :drain, :remove_node_from_cluster
```
aamederen: where do we check the drain semaphore?

Can we simplify this and just handle everything with a single `retire` semaphore? Maybe we can follow drain -> remove -> destroy sequence when retire is set.

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -26,11 +26,49 @@ def before_run
...
+      hop_retire
+    end
+    when_remove_node_from_cluster_set? do
+      hop_remove_node_from_cluster
+    end
     nap 6 * 60 * 60
   end
 
+  label def drain
+    unit_name = "drain_node_#{kubernetes_node.name}"
```
aamederen, line 49: why do we need a special unit_name that contains the node name here, instead of a static name?

model/kubernetes/kubernetes_node.rb
mohi-kalantari: It's not possible like this if we want to reuse the logic since upgrade has some extra steps before removing the node. in order to share the logic we need the semaphores

prog/kubernetes/kubernetes_node_nexus.rb
mohi-kalantari: multiple drains might happen. if a customer sets the node_count from 4 to 2, 2 drain units are needed

aamederen: I see, this is because we are running the drain in the same CP vm, not in this node.

model/kubernetes/kubernetes_node.rb
mohi-kalantari: fixed it based on your recommendation. thanks.

PR #3792, opened 2025-08-26
Print warning if specs leak threads

jeremyevans:
This can be the cause of spec nondeterminism.

----

> [!IMPORTANT]
> Adds thread leak detection in RSpec tests and updates tests to manage threads properly.
> 
>   - **Behavior**:
>     - Adds a check for leaked threads in `spec_helper.rb` using `ObjectSpace::WeakMap` to track threads and print warnings for any new threads after tests.
>     - Sets `@skip_leaked_thread_check` to `true` in `cli_spec.rb` before tests and `false` after tests to control thread leak checks.
>   - **Tests**:
>     - Ensures `log_dna_batcher.stop` is called in `log_dna_batcher_spec.rb` after tests to prevent thread leaks.
>     - Updates `thread_printer_spec.rb` to handle threads with nil backtrace and created_at attributes.
>   - **Misc**:
>     - Minor changes in `cli_spec.rb` to manage server lifecycle and thread checks.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for db28dbef7a85cc9253cba0a34ff860f14dbf2a30. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/lib/thread_printer_spec.rb
+++ b/spec/lib/thread_printer_spec.rb
@@ -20,14 +20,18 @@ RSpec.describe ThreadPrinter do
       # The documentation calls out that the backtrace is an array or
       # nil.
       expect(described_class).to receive(:puts).with(/--BEGIN THREAD DUMP, .*/)
-      expect(described_class).to receive(:puts).with(/Thread: #<InstanceDouble.*>/)
+      expect(described_class).to receive(:puts).with(/Thread: #<Thread:.*>/)
       expect(described_class).to receive(:puts).with(/Created at: .*/)
       expect(described_class).to receive(:puts).with(nil)
       expect(described_class).to receive(:puts).with(/--END THREAD DUMP, .*/)
-      th = instance_double(Thread, backtrace: nil)
+      q = Queue.new
+      th = Thread.new { q.pop }
+      expect(th).to receive(:backtrace).and_return(nil)
       expect(th).to receive(:[]).with(:created_at).and_return(Time.now - 30)
       expect(Thread).to receive(:list).and_return([th])
       described_class.run
+      q.push nil
+      th.join(1)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


spec/lib/thread_printer_spec.rb
```diff
@@ -20,14 +20,19 @@
...
       expect(described_class).to receive(:puts).with(/Created at: .*/)
       expect(described_class).to receive(:puts).with(nil)
       expect(described_class).to receive(:puts).with(/--END THREAD DUMP, .*/)
-      th = instance_double(Thread, backtrace: nil)
+      q = Queue.new
+      th = Thread.new { q.pop }
+      expect(th).to receive(:backtrace).and_return(nil)
       expect(th).to receive(:[]).with(:created_at).and_return(Time.now - 30)
       expect(Thread).to receive(:list).and_return([th])
+      allow(Thread).to receive(:list).and_call_original
```
fdr: why not expect [`.at least(:once)`, I wish there were something shorter]? it's kinda wordy but at least you can delete it if the implementation changes such that it's not needed anymore.

fdr (approved):

spec/lib/thread_printer_spec.rb
jeremyevans: There is no guarantee the method is called twice (notice the line above), so the code breaks if you switch it to that.

fdr: I'm a bit puzzled: what about the thread dumper necessitates non-determinism in listing threads? This so the leaked thread checker can run, right? Doesn't it always run, and always list?

jeremyevans: I can try removing the `allow` line. Maybe it was needed in an earlier version where the leaked thread check was in an `after` hook and not in the `around` hook.

PR #3806, opened 2025-08-28
Remove old vm association in k8s

mohi-kalantari:
Update regen screenshots for Kubernetes to use the new models

Drop all associations and usages of the old model of Kubernetes Vms

Now that we have migrated to the new model and use it for everything, we need to drop the old usages.

Drop old tables for the kubernetes vms many_to_many associations


aamederen (approved):

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -72,10 +72,7 @@ def before_run
     current_frame["node_id"] = node.id
     strand.modified!(:stack)
 
-    if kubernetes_nodepool
-      kubernetes_nodepool.add_vm(vm)
-    else
-      kubernetes_cluster.add_cp_vm(vm)
+    if kubernetes_nodepool.nil?
```
jeremyevans: `unless kubernetes_nodepool` is the idiomatic way to handle this.

migrate/20250828_drop_old_join_tables_of_kubernetes_vms.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    drop_table :kubernetes_clusters_cp_vms
+    drop_table :kubernetes_nodepools_vms
+  end
+
+  down do
```
jeremyevans:
```suggestion
  revert do
```

jeremyevans (approved):

PR #3807, opened 2025-08-28
Add new image for Kubernetes 1.34 and remove 1.32 from version options

mohi-kalantari:
New image is added to support kubernetes 1.34

Update k8s tests so we won't have to update tests on version releases

Before this commit, on every new version, we had to update the tests to use to the new version. With this new approach we would use the first option of the kubernetes versions list.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -184,6 +184,7 @@ module Config
   override :ai_ubuntu_2404_nvidia_version, "20250505.1.0", string
   override :kubernetes_v1_32_version, "20250320.1.0", string
   override :kubernetes_v1_33_version, "20250506.1.0", string
+  override :kubernetes_v1_34_version, "20250828.1.0", string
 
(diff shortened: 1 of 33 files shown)


mohi-kalantari:
Hi @jeremyevans 

Regarding the second commit, do you have anything in mind to automate the spec edits for CLI? I guess we can add rake tasks to populate those files again with CLI but it somehow defeats the testing purpose I guess

jeremyevans:
> Regarding the second commit, do you have anything in mind to automate the spec edits for CLI? I guess we can add rake tasks to populate those files again with CLI but it somehow defeats the testing purpose I guess

After running `rspec spec/routes/api/cli/golden_files_spec.rb`, you run `rake update_golden_files`. You  need to check the diff manually to make sure the changes are expected.

aamederen (approved):

PR #3811, opened 2025-08-28
Update some bundle packages

enescakir:
We use dependabot to update package versions in our project. But the don't update production pacakges for a time. Currently dependabot creates a PR with 29 depencies [^1]. It's litte bit too much for one PR. So I decided to update some packages manually to divide it multiple steps.

In ideal world, we should update packages more often without cluttred pile of packages

[^1]: #3725


Gemfile.lock
```diff
@@ -373,11 +380,11 @@ GEM
       addressable (>= 2.3.5)
       faraday (>= 0.17.3, < 3)
     securerandom (0.4.1)
-    sequel (5.95.0)
+    sequel (5.95.1)
```
jeremyevans: This will conflict with #3797, and the issue fixed doesn't affect us.
```suggestion
    sequel (5.95.0)
```

jeremyevans (approved):

PR #3816, opened 2025-08-28
Add underline to hyperlinks in the web shell output

enescakir:
UBIDs in the web shell output are actually clickable links, which is a very cool feature. However, it’s hard to notice at first glance, I only realized it when I tried to copy one.

Let's add underlines to hyperlinks in the web shell output to make them more visible.

----

> [!IMPORTANT]
> Adds underline to hyperlinks in web shell output in `Clover` class to improve visibility.
> 
>   - **Behavior**:
>     - Adds underline to hyperlinks in web shell output in `Clover` class in `routes/project/cli.rb` to improve visibility.
>     - Specifically modifies the HTML anchor tag to include `underline` class.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for eb448a02592c427be4547f8e81e312e4d8923146. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/routes/project/cli.rb
+++ b/routes/project/cli.rb
@@ -56,7 +56,7 @@ class Clover
             UBID.resolve_map(ubids)
             h(body).gsub(UbiCli::OBJECT_INFO_REGEXP) do
               if (obj = ubids[UBID.to_uuid(it)]) && obj.respond_to?(:path)
-                "<a class=\"text-orange-600\" href=\"#{@project.path}#{obj.path}\">#{it}</a>"
+                "<a class=\"text-orange-600 underline\" href=\"#{@project.path}#{obj.path}\">#{it}</a>"
               else
                 it
               end


jeremyevans:
This is inconsistent from how links are shown on other pages. If we want to do this for the web shell, we should do it for all other pages.  That would be a ~1 line change if we were using plain CSS, but it will be a lot of work due to the use of tailwind.

byucesoy:
> That would be a ~1 line change if we were using plain CSS, but it will be a lot of work due to the use of tailwind.

Maybe I’m a bit old-school here, but honestly I’m not fully convinced about the value of Tailwind. To me, it feels like inline CSS with extra steps. Since it doesn’t cover the full range of CSS features, you often end up needing plain CSS anyway. Plain CSS feels more natural and straightforward to me, but that might just be personal preference.

enescakir:
> This is inconsistent from how links are shown on other pages

I’m closing this one for now.

PR #3819, opened 2025-08-29
Self destruct Postgres Timelines when it's safe enough.

aamederen:
We used to let PG timelines linger around when servers get destroyed and do manual clean-up every once in a while. This change adds some logic to PG timeline to initiate a self-destruct if there's no leader and the timeline is old enough.

----

> [!IMPORTANT]
> Add self-destruct logic for Postgres timelines without a leader or backups older than 10 days.
> 
>   - **Behavior**:
>     - Adds self-destruct logic to `wait` in `postgres_timeline_nexus.rb` for timelines with no leader, no backups, and older than 10 days.
>     - Emits a log message before self-destructing.
>   - **Tests**:
>     - Adds test for self-destruct condition in `wait` method in `postgres_timeline_nexus_spec.rb`.
>     - Verifies no API calls for backups if no leader is present.
>   - **Misc**:
>     - Fixes typo in test description in `postgres_timeline_nexus_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for eefa6eb6fbca8795265cead607a2e9b047562d89. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_timeline_nexus.rb
+++ b/prog/postgres/postgres_timeline_nexus.rb
@@ -65,7 +65,8 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
 
   label def wait
     leader = postgres_timeline.leader
-    if leader.nil? && postgres_timeline.backups.empty? && Time.now - postgres_timeline.created_at > 10 * 24 * 60 * 60
+    backups = leader.nil? ? [] : postgres_timeline.backups
+    if leader.nil? && backups.empty? && Time.now - postgres_timeline.created_at > 10 * 24 * 60 * 60
       Clog.emit("Self-destructing timeline as no leader or backups are present and it is older than 10 days") { postgres_timeline }
       hop_destroy
     end
@@ -75,7 +76,7 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
     # For the purpose of missing backup pages, we act like the very first backup
     # is taken at the creation, which ensures that we would get a page if and only
     # if no backup is taken for 2 days.
-    latest_backup_completed_at = postgres_timeline.backups.map(&:last_modified).max || postgres_timeline.created_at
+    latest_backup_completed_at = backups.map(&:last_modified).max || postgres_timeline.created_at
     if leader && latest_backup_completed_at < Time.now - 2 * 24 * 60 * 60 # 2 days
       Prog::PageNexus.assemble("Missing backup at #{postgres_timeline}!", ["MissingBackup", postgres_timeline.id], postgres_timeline.ubid)
     else
--- a/prog/postgres/postgres_timeline_nexus.rb
+++ b/prog/postgres/postgres_timeline_nexus.rb
@@ -65,7 +65,7 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
 
   label def wait
     leader = postgres_timeline.leader
-    backups = leader.nil? ? [] : postgres_timeline.backups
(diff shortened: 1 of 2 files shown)


prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -64,13 +64,19 @@ def before_run
...
+    end
+
     nap 20 * 60 if postgres_timeline.blob_storage.nil?
 
     # For the purpose of missing backup pages, we act like the very first backup
     # is taken at the creation, which ensures that we would get a page if and only
     # if no backup is taken for 2 days.
     latest_backup_completed_at = postgres_timeline.backups.map(&:last_modified).max || postgres_timeline.created_at
-    if postgres_timeline.leader && latest_backup_completed_at < Time.now - 2 * 24 * 60 * 60 # 2 days
+    if leader && latest_backup_completed_at < Time.now - 2 * 24 * 60 * 60 # 2 days
```
jeremyevans: Not in issue with this PR, but we may want to avoid the `latest_backup_completed_at` logic if there is no `leader`.

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -64,13 +64,19 @@ def before_run
   end
 
   label def wait
+    leader = postgres_timeline.leader
+    if leader.nil? && postgres_timeline.backups.empty? && Time.now - postgres_timeline.created_at > 10 * 24 * 60 * 60
```
jeremyevans: I would save `postgres_timeline.backups` to a local variable and use it here and in the code below, as it requires an API call.

jeremyevans (commented):
Code-wise, this looks good, it's just a question of whether we want to do this.

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -64,13 +64,19 @@ def before_run
   end
 
   label def wait
+    leader = postgres_timeline.leader
```
byucesoy, line 67: These should come after `postgres_timeline.blob_storage.nil?` check.

aamederen: if we move these checks after the postgres_timeline.blob_storage.nil? check, then this logic would not clean the dev environments without minio clusters (such as mine). `postgres_timeline.backups` call also has `postgres_timeline.blob_storage.nil?` check inside.

byucesoy: Makes sense. :+1:

prog/postgres/postgres_timeline_nexus.rb
aamederen: I think I addressed your comment now with the new commit. We still go through the latest_backup_completed_at logic, but don't do an API call if `leader.nil?`

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -64,13 +64,20 @@ def before_run
...
+    backups = leader.nil? ? [] : postgres_timeline.backups
```
jeremyevans:
```suggestion
    backups = leader ? postgres_timeline.backups : []
```

Generally, if you don't care about the `false`/`nil` difference, it's better not to call `nil?`.

jeremyevans (approved):

PR #3822, opened 2025-08-29
Add support for renaming objects on the object's settings page

jeremyevans:
Support for renaming objects was previously added to the api/sdk/cli. This adds support on the web. r.rename now works for both api and web routes, and requires a new template_prefix option used to construct the validation failure template path.

The template code for the new rename support is shared via a new rename_object component.

In order to improve presentation and consistency, this adds a "Danger Zone" heading to the settings pages previously lacking it.

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_kubernetes-cluster_t_settings
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_postgres_t_settings
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_load-balancer_central_settings
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-north-h1_firewall_test-fw4_settings
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-north-h1_private-subnet_default-eu-north-h1_settings
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_vm_t46_settings


helpers/general.rb
```diff
@@ -19,20 +19,26 @@ def accepts_json?
...
-        Validation.validate_name(name)
-
-        DB.transaction do
-          object.update(name:)
-          scope.audit_log(object, "update")
+        scope.instance_exec do
+          authorize(perm, object.id)
+          handle_validation_failure("#{template_prefix}/show") { @page = "settings" }
+          name = typecast_body_params.nonempty_str!("name")
+          Validation.validate_name(name)
```
byucesoy: I missed this in the earlier PR, but it seems k8s uses different validation function to validate the name: `validate_kubernetes_name` and it seems the only difference is the length.

```ruby
  ALLOWED_NAME_PATTERN            = %r{\A[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?\z}
  ALLOWED_KUBERNETES_NAME_PATTERN = %r{\A[a-z0-9](?:[a-z0-9\-]{0,38}[a-z0-9])?\z}
```

Not sure why we needed that and I would prefer consolidating to one name validation method if possible. Just wanted to point out in case this would cause something unexpected. (cc @aamederen , @mohi-kalantari)

routes/project/location/postgres.rb
```diff
@@ -100,7 +100,7 @@ class Clover
         end
       end
 
-      r.rename pg, perm: "Postgres:edit", serializer: Serializers::Postgres
+      r.rename pg, perm: "Postgres:edit", serializer: Serializers::Postgres, template_prefix: "postgres"
```
byucesoy, line 103: nit: I think an empty line after this (and other `r.rename` lines) would be nice, to give visual distinction between Roda branches (is this the correct name for r.xxxx blocks?)

helpers/general.rb
enescakir: They add some suffixes to the name for node names, so I think they want to keep that generated name under 63 characters.

jeremyevans: Good catch. I'll need to update this to handle that. In general, model-level validations like this would be better off as actual model-level validations (in the Sequel models), so that they were checked inside `Model#update`, and therefore you didn't need different code per-model.

routes/project/location/postgres.rb
jeremyevans: I'll add an empty line between these before merging.

PR #3828, opened 2025-08-29
Allow creating a PostgreSQL database without firewall rules in api/sdk/cli

jeremyevans:
The reason for allowing this is it allows you to create a locked-down PostgreSQL database without parsing output of previous commands. Previously, to create a locked-down database, you needed to create the database, then remove the default firewall rules for it, then add your own. Using this, you add the database without rules, and then add the rules you need.

----

> [!IMPORTANT]
> Add `--restrict-by-default` option to create PostgreSQL databases without default firewall rules in CLI, API, and backend logic.
> 
>   - **Behavior**:
>     - Adds `--restrict-by-default` option in `create.rb` to create PostgreSQL databases without default firewall rules.
>     - Updates `postgres_post` in `postgres.rb` to handle `restrict_by_default` parameter, setting `with_firewall_rules` accordingly.
>     - Modifies `assemble` in `postgres_resource_nexus.rb` to conditionally create default firewall rules based on `with_firewall_rules`.
>   - **API**:
>     - Adds `restrict_by_default` boolean to `openapi.yml` for PostgreSQL database creation.
>   - **Tests**:
>     - Updates `create_spec.rb` to test database creation with and without `--restrict-by-default` option.
>     - Updates CLI help text in `help -r.txt` and other golden files to include `--restrict-by-default` option.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for dbd584ab25fa89db58275cafcb026860fe726ce0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli-commands/pg/post/create.rb
+++ b/cli-commands/pg/post/create.rb
@@ -10,6 +10,7 @@ UbiCli.on("pg").run_on("create") do
     on("-S", "--storage-size=size", Option::POSTGRES_STORAGE_SIZE_OPTIONS, "storage size GB")
     on("-v", "--version=version", Option::POSTGRES_VERSION_OPTIONS, "PostgreSQL version")
     on("-t", "--tags=tags", "tags (e.g. key1=value1,key2=value2)")
+    on("-R", "--restrict-by-default", "restrict access by default (add firewall rules to allow access)")
   end
   help_option_values("Flavor:", Option::POSTGRES_FLAVOR_OPTIONS.keys)
   help_option_values("Replication Type:", Option::POSTGRES_HA_OPTIONS.keys)
--- a/helpers/postgres.rb
+++ b/helpers/postgres.rb
@@ -11,6 +11,7 @@ class Clover
     ha_type = typecast_params.nonempty_str("ha_type", PostgresResource::HaType::NONE)
     version = typecast_params.nonempty_str("version", PostgresResource::DEFAULT_VERSION)
     tags = typecast_params.array(:Hash, "tags", [])
+    with_firewall_rules = !typecast_params.bool("restrict_by_default")
 
     postgres_params = {
       "flavor" => flavor,
@@ -39,6 +40,7 @@ class Clover
         target_storage_size_gib: storage_size,
         ha_type:,
         version:,
+        with_firewall_rules:,
         flavor:
       ).subject
       pg.update(tags:)
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -838,6 +838,9 @@ paths:
                 version:
                   description: PostgreSQL version
                   type: string
+                restrict_by_default:
+                  description: 'Whether to restrict access by default (if so, firewall rules must be added to access)'
+                  type: boolean
                 tags:
                   description: Tags for the Postgres Database
                   type: array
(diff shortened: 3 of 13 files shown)


fdr:
Perhaps the API version (as opposed to web) should not add rules by default. It also hurts terraform quite a bit. On the other hand, it would also be surprising if one was trying to translate from click-ops to cli-ops.  Maybe if we make this pattern consistent we could also have a project-wide switch for "production" projects that want no default ingress. What do you think makes sense to square the circle? the issue of default openness for varying use cases doesn't have a pat solution across all levels of formality, and most organizations operate at all levels.

jeremyevans:
> Perhaps the API version (as opposed to web) should not add rules by default. It also hurts terraform quite a bit. On the other hand, it would also be surprising if one was trying to translate from click-ops to cli-ops. Maybe if we make this pattern consistent we could also have a project-wide switch for "production" projects that want no default ingress. What do you think makes sense to square the circle? the issue of default openness for varying use cases doesn't have a pat solution across all levels of formality, and most organizations operate at all levels.

I think we want consistency between web/api/cli, and also across vm/pg/etc.. When you add a vm without a private subnet, it creates a private subnet and firewall, and adds firewall rules to allow access. So I think for now, pg should default to allowing access by default, but supporting the ability to not allow access by default.

I think it would be a good idea to offer a project-wide setting that controls whether permissive rules are added by default when creating a private subnet and postgres database.

fdr:
Suppose we decide to make it consistent: is this the best name for it? It might be, I guess we should inventory the other places we want it (subnets, kubernetes?) to see if it feels right. I don't have an itch to change it, but if we aim to make it a pattern, it seems worth a moment to dwell on naming.

jeremyevans:
By "make it consistent", do you mean "allow a setting change that disables automatic permissive firewall rule addition when creating private subnets and postgresql databases"?  In that case, the "restrict_by_default" argument/option would be a no-op (true whether set or not). If instead we wanted a setting that would allow permissive rule creation even if the project setting was to disable permissiveness by default, then I think we would need to rename the argument/option.

If you think the current behavior is inconsistent, could you explain why? Maybe there is something I missed.

fdr:
Nah, I mean having the same naming, flag, and parameter across several kinds of objects is all.

fdr:
> Nah, I mean having the same naming, flag, and parameter across several kinds of objects is all.

(the most obvious of those objects being subnets, perhaps the only object really, if we manage to unify the data model)

jeremyevans:
For subnets, you can create them with a specific firewall, in which case you don't have this issue (it does not add rules to an existing firewall, only a newly created firewall). If we allowed postgres servers to attach to customer firewalls, that is potentially an alternative way to address this issue.

byucesoy (approved):

byucesoy:
+1 for
- Keeping things open by default
- Keeping things consistent across products.

I’m neutral on adding an option to restrict access at provisioning time. Since you can already issue a follow-up request right after the first request (even before provisioning is completed) to apply restrictions, the added value seems low. The downside is a bit of extra code to maintain, though it’s small enough not to be a concern.

jeremyevans:
> I’m neutral on adding an option to restrict access at provisioning time. Since you can already issue a follow-up request right after the first request (even before provisioning is completed) to apply restrictions, the added value seems low. The downside is a bit of extra code to maintain, though it’s small enough not to be a concern.

The value is mostly for the cli. With this new feature, you can create a new database only allowing specific IP ranges just by running commands, without parsing results of previous commands. Without this, that is not possible, as you must parse output of previous commands in order to determine the PG firewall rules to modify or remove.

Note that VMs already support this, since you can create a firewall (which by default has no rules), create a private subnet with that firewall (does not add rules), and create a virtual machine in that private subnet (which also does not add rules). So this adds for PG what is already possible for VMs.

PR #3830, opened 2025-09-01
Improve k8s upgrades

mohi-kalantari:
Do not silently fail on kubernetes commands

Prior to this commit, we were not checking the existatus code of commands we would run using the client. From now on, a RuntimeError will be raised.

Use the first functional node for KubernetesCluster#sshable

We were using the last vm as the cluster's sshable but by introducing functional_nodes association, we make sure we are not choosing a draining node during upgrade as the cluster's sshable.

This case would have happened if we were upgrading a cluster with 1 control plane node Fix issue in kubernetes upgrade delete node step.

Fix issue in Kubernetes upgrade process

Prior to this commit, we would run the drain/delete command on the last node of the cluster but in a setup where we have 1 control plane node, by the time we are running the delete command, we have already run the kubeadm reset resulting in errors since there are no kubeconfigs there.

With this commit, we make sure we are choosing a functional node and not a draining one.

----

> [!IMPORTANT]
> Improve Kubernetes upgrades by handling command failures and refining node selection logic in `client.rb`, `kubernetes_cluster.rb`, and `kubernetes_node_nexus.rb`.
> 
>   - **Behavior**:
>     - `kubectl` in `client.rb` now raises `RuntimeError` on non-zero exit status to prevent silent failures.
>     - `sshable` method in `kubernetes_cluster.rb` now uses `functional_nodes` instead of `cp_vms` to select nodes.
>     - `drain` method in `kubernetes_node_nexus.rb` now uses `cluster.sshable` for node selection.
>   - **Models**:
>     - Added `functional_nodes` association to `KubernetesCluster` in `kubernetes_cluster.rb`.
>   - **Tests**:
>     - Updated `kubectl` tests in `client_spec.rb` to check for raised errors on command failure.
>     - Modified `kubernetes_node_nexus_spec.rb` to test new `sshable` logic in `drain` method.
>     - Adjusted `provision_kubernetes_node_spec.rb` to use `functional_nodes` for node selection in `join_control_plane` and `join_worker` methods.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for d950079a0e7b3e4ea7503ff7000164d10623768f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/kubernetes/kubernetes_node_nexus.rb
+++ b/prog/kubernetes/kubernetes_node_nexus.rb
@@ -47,7 +47,7 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
 
   label def drain
     unit_name = "drain_node_#{kubernetes_node.name}"
-    sshable = cluster.cp_vms.last.sshable
+    sshable = cluster.sshable
     case sshable.d_check(unit_name)
     when "Succeeded"
       hop_remove_node_from_cluster
@@ -77,7 +77,8 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
       cluster.api_server_lb.detach_vm(kubernetes_node.vm)
     end
 
-    cluster.client(session: cluster.nodes.last.sshable.connect).delete_node(kubernetes_node.name)
+    cluster.client.delete_node(kubernetes_node.name)
+
     hop_destroy
   end
 
--- a/lib/kubernetes/client.rb
+++ b/lib/kubernetes/client.rb
@@ -31,7 +31,9 @@ class Kubernetes::Client
   end
 
   def kubectl(cmd)
-    @session.exec!("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
+    output = @session.exec!("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
+    raise output if output.exitstatus != 0
+    output
   end
 
   def version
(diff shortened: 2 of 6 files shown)


prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -77,7 +77,8 @@ def nodepool
       cluster.api_server_lb.detach_vm(kubernetes_node.vm)
     end
 
-    cluster.client(session: cluster.nodes.last.sshable.connect).delete_node(kubernetes_node.name)
+    cluster.client(session: cluster.alternative_node(kubernetes_node).sshable.connect).delete_node(kubernetes_node.name)
```
aamederen: can we change the client code, so that it connects to the first `node` that's not being retired?

mohi-kalantari: I guess it makes its code complicated. the client was designed to run kubectl related commands with a ssh session. If we want to query and search through the KubernetesCluster associations, I feel like we are leaking too much detail into the client.

aamederen: By client, I meant the place where we called the client. Specifically, I mean changing `KubernetesCluster#sshable` function.

mohi-kalantari: oh got it. Do you have any suggestions for its name? something like `cluster.alternative_sshable(kubernetes_node).connect`? I don't see any clear improvements regarding the code cleanliness. This version seems more readable imo.

aamederen: I think `KubernetesCluster.sshable` method should pick the correct vm and use its sshable, instead of `cp_vms.first.sshable`

mohi-kalantari: makes sense. I'll fix it.

aamederen (approved):

PR #3832, opened 2025-09-01
Retry once when VmHost healthcheck fails because of dropped ssh connections

mohi-kalantari:
There are many cases when monitor is overloaded and VmHosts' keep alive settings drops the connection. With this commit, we add a retry before considering a dead connection "down"


model/vm_host.rb
```diff
@@ -371,9 +371,22 @@ def perform_health_checks(ssh_session)
   end
 
   def check_pulse(session:, previous_pulse:)
+    stale_retry = false
     reading = begin
       perform_health_checks(session[:ssh_session]) ? "up" : "down"
     rescue => e
+      if !stale_retry &&
```
jeremyevans: This is the same logic used in LoadBalancerVmPort. It should be abstracted instead of duplicated.

lib/monitorable_resource.rb
```diff
@@ -41,11 +41,24 @@ def process_event_loop
...
+      if !stale_retry &&
+          (
+            # Seen when sending on a broken connection.
+            ex.is_a?(IOError) && ex.message == "closed stream" ||
+            # Seen when receiving on a broken connection.
+            ex.is_a?(Errno::ECONNRESET) && ex.message.start_with?("Connection reset by peer - recvfrom(2)")
+          ) &&
+          @session[:last_pulse]&.<(Time.now - 8)
+        stale_retry = true
+        @session.merge!(@resource.init_health_monitor_session)
```
jeremyevans, line 57: At any point, `@session` can be set to `nil` by the pulse thread calling `close_resource_session`. If `@session` is `nil` here (very rare, but still possible), this raises a NoMethodError, which causes the related monitor thread to exit.

I think we need this:

```suggestion
        @session&.merge!(@resource.init_health_monitor_session)
```

Probably need to add nocov because I don't think it's possible to hit this case in tests with the current code.

Note that this was in issue in the previous code, but it would be caught by this rescue. As you are inside the block, it will no longer be caught by this rescue.

An alternative approach is to wrap this code in a separate begin/rescue block.

lib/monitorable_resource.rb
```diff
@@ -41,11 +41,24 @@ def process_event_loop
...
       Clog.emit("Got new pulse.") { {got_pulse: {ubid: @resource.ubid, pulse: @pulse}} } if (rpt = @pulse[:reading_rpt]) && (rpt < 6 || rpt % 5 == 1) || @pulse[:reading] != "up"
     rescue => ex
+      if !stale_retry &&
+          (
+            # Seen when sending on a broken connection.
+            ex.is_a?(IOError) && ex.message == "closed stream" ||
+            # Seen when receiving on a broken connection.
+            ex.is_a?(Errno::ECONNRESET) && ex.message.start_with?("Connection reset by peer - recvfrom(2)")
+          ) &&
+          @session[:last_pulse]&.<(Time.now - 8)
```
jeremyevans, line 55:
```suggestion
          @session&.[](:last_pulse)&.<(Time.now - 8)
```

Same issue as in a couple lines down.

fdr:
it's unfortunate we're peppering important branches with nocov. 

One area aa2e0c0749903b863bf5989b1c59f0cf6de7d8bb because I didn't get around to working through the implications of a retry for *all* ssh, particularly with Postgres's connection, which is liable to be persistent. It's not likely in that particular case to have session state, but pretending nothing happened to the connection is not always a good idea. 

Even as-is, we should have a firm idea of the error expected on the postgres side. I suspect it'll crash the next time it tries to send protocol traffic, which means at the minimum this change won't help Postgres. But as @byucesoy might remember, the exactness of layering how errors are propagated is an important part of the implementation.

Alternatively, the net ssh session object may know if it has any channels open to do forwarding, the retry can be disabled in such a case.

jeremyevans:
> it's unfortunate we're peppering important branches with nocov.

Agreed, but as long as we run an unsynchronized pulse thread that can set `@session = nil` at any point, we need to use `@session&.`. Since it's not simple to hit the `@session = nil` case in tests, we either need to introduce a fault injection framework, use `nocov`, or refactor to avoid the unsynchronized pulse thread. I prefer the latter approach.

fdr (approved):
unsynchronized `@session` mangling aside, it's too bad we lose the coverage on the nil last pulse and `<` as well. I suppose some line mangling can solve it, or alternatively, the monitorable interface needs some work. (to synchronize, as you suggest)

jeremyevans:
I submitted #3846, which I think should fix the synchronization issues, and allow for `@session.` instead of `@session&.`

mohi-kalantari:
I'll wait for https://github.com/ubicloud/ubicloud/pull/3846 then update this PR and deploy it.

PR #3838, opened 2025-09-03
Admin site improvements

jeremyevans:
This adds the following:

* Configurable link text on the admin side, instead of always using ubid. For models that support it, name is used. For accounts, email is used.
* Browsing objects by class
* A button for immediate scheduling a strand

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -161,10 +161,18 @@ class CloverAdmin < Roda
         view("objects")
       end
 
-      r.get :ubid do |ubid|
+      r.on :ubid do |ubid|
         next unless (@obj = @klass[ubid])
 
-        view("object")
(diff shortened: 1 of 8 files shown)


enescakir:
What do you think about a table?
https://github.com/ubicloud/ubicloud/compare/jeremy-admin-site-model-labels...enes/admin-objects-table?expand=1

CleanShot 2025-09-03 at 15 06 39@2x

enescakir:
Though it might cause some N+1

jeremyevans:
> Though it might cause some N+1

I don't think we want to enable table display by default for that reason. I can add support for table display on a per-class basis, in a way that avoids N+1. Can you let me know which initial models you would like to use table display?

PR #3849, opened 2025-09-04
Add project-level postgres hostname override

byucesoy:
**Move dns_zone helper to PostgresResource model**
The dns_zone helper was previously a static method in PostgresResourceNexus,
based on the assumption that all PostgresResource instances used the same
global DNS Zone. Since we plan to support project-level hostname overrides,
this assumption no longer holds. Moving the helper to the PostgresResource
model allows it to access the project context and retrieve project-specific
DNS Zone information when needed.

**Add project-level postgres hostname override**
This allows each project to specify its own Postgres hostname suffix, rather
than relying on a global default. This change is necessary to support white
label deployments where branded Postgres hostnames are necessary.

----

> [!IMPORTANT]
> Add project-level Postgres hostname override by moving `dns_zone` to `PostgresResource` and updating related methods and tests.
> 
>   - **Behavior**:
>     - Add project-level Postgres hostname override in `postgres_resource.rb` using `hostname_suffix` method.
>     - Update `dns_zone` method in `postgres_resource.rb` to use project-specific DNS Zone.
>     - Modify `hostname`, `identity`, and `connection_string` methods in `postgres_resource.rb` to use `hostname_suffix`.
>   - **Models**:
>     - Move `dns_zone` helper from `PostgresResourceNexus` to `PostgresResource`.
>     - Add `postgres_hostname_override` feature flag in `project.rb`.
>   - **Tests**:
>     - Update tests in `postgres_resource_spec.rb` and `postgres_resource_nexus_spec.rb` to reflect changes in DNS handling.
>     - Modify CLI tests in `golden_files_spec.rb`, `pg_dumpall_spec.rb`, `psql_pg_dump_spec.rb`, and `show_spec.rb` to use project-specific DNS zones.
>     - Adjust mocking in `thawed_mock.rb` to accommodate changes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for e7602783a092f7c9fa4ea1742b37cefe5c5003ef. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -47,21 +47,25 @@ class PostgresResource < Sequel::Model
     "creating"
   end
 
+  def hostname_suffix
+    project.get_ff_postgres_hostname_override || Config.postgres_service_hostname
+  end
+
   def dns_zone
-    @dns_zone ||= DnsZone[project_id: Config.postgres_service_project_id, name: Config.postgres_service_hostname]
+    @dns_zone ||= DnsZone[project_id: Config.postgres_service_project_id, name: hostname_suffix]
   end
 
   def hostname
     if dns_zone
-      return "#{name}.#{Config.postgres_service_hostname}" if hostname_version == "v1"
-      "#{name}.#{ubid}.#{Config.postgres_service_hostname}"
+      return "#{name}.#{hostname_suffix}" if hostname_version == "v1"
+      "#{name}.#{ubid}.#{hostname_suffix}"
     else
       representative_server&.vm&.ephemeral_net4&.to_s
     end
   end
 
   def identity
-    "#{ubid}.#{Config.postgres_service_hostname}"
+    "#{ubid}.#{hostname_suffix}"
   end
 
   def connection_string
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -47,8 +47,12 @@ class PostgresResource < Sequel::Model
     "creating"
   end
 
+  def dns_zone
+    @dns_zone ||= DnsZone[project_id: Config.postgres_service_project_id, name: Config.postgres_service_hostname]
+  end
+
   def hostname
-    if Prog::Postgres::PostgresResourceNexus.dns_zone
+    if dns_zone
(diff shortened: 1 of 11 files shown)


enescakir (approved):

model/postgres/postgres_resource.rb
```diff
@@ -47,17 +47,25 @@ def display_state
     "creating"
   end
 
+  def postgres_service_hostname
+    project.get_ff_postgres_hostname_override || Config.postgres_service_hostname
+  end
```
enescakir: Since it's not just returning `Config.postgres_service_hostname` and includes some logic, `postgres_hostname` instead of `postgres_service_hostname` might be clearer to indicate that it's not simply returning the `Config` value.

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -103,8 +103,8 @@ def before_run
 
     type, data = postgres_resource.location.aws? ? ["CNAME", representative_server.vm.aws_instance.ipv4_dns_name + "."] : ["A", representative_server.vm.ephemeral_net4.to_s]
 
-    Prog::Postgres::PostgresResourceNexus.dns_zone&.delete_record(record_name: postgres_resource.hostname)
-    Prog::Postgres::PostgresResourceNexus.dns_zone&.insert_record(record_name: postgres_resource.hostname, type:, ttl: 10, data:)
+    postgres_resource.dns_zone&.delete_record(record_name: postgres_resource.hostname)
+    postgres_resource.dns_zone&.insert_record(record_name: postgres_resource.hostname, type:, ttl: 10, data:)
```
jeremyevans: Not an issue in this PR, but in general in situations like this, where you have multiple expressions using `&.` on the same object, it's probably better to use an `if` statement around both:

```suggestion
    if postgres_resource.dns_zone
      postgres_resource.dns_zone.delete_record(record_name: postgres_resource.hostname)
      postgres_resource.dns_zone.insert_record(record_name: postgres_resource.hostname, type:, ttl: 10, data:)
    end
```

jeremyevans (approved):

PR #3720, opened 2025-08-07
Support placing standbys in different AZs

furkansahin:
## Cleanup AWS subnet info from private_subnet_aws_resource - migration

----

> [!IMPORTANT]
> Enhance PostgreSQL standby placement by supporting different AWS AZs, updating models, migrations, and logic for subnet and VPC management.
> 
>   - **Behavior**:
>     - Adds AZ selection functionality to `Prog::Postgres::ConvergePostgresResource` and `Prog::Postgres::PostgresServerNexus`.
>     - Updates `Prog::Aws::Nic` to create subnets with AZ consideration and manage subnet lifecycle.
>     - Modifies `Prog::Aws::Vpc` to handle VPC and route table creation without subnet association.
>   - **Migrations**:
>     - Adds `subnet_id` and `subnet_az` to `nic_aws_resource` in `20250807_add_subnet_info_to_nic_aws_resource.rb`.
>     - Removes `subnet_id` from `private_subnet_aws_resource` in `20250822_drop_subnet_id_on_private_subnet_aws_resource.rb`.
>   - **Models**:
>     - Adds `AWS_PROVIDER_NAME` to `HostProvider`.
>     - Updates `NicAwsResource` to include `subnet_id` and `subnet_az`.
>     - Adds AZ-related logic to `PrivateSubnet`.
>   - **Tests**:
>     - Adds RSpec tests for `Prog::Aws::Nic` and `Prog::Aws::Vpc` to verify subnet and VPC operations.
>     - Tests AZ selection logic in `Prog::Postgres::ConvergePostgresResource`.
>     - Tests `Prog::Vnet::NicNexus` for NIC creation and destruction in AWS.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 15becf71a2e1f7914f34f2c104dfa4613eb5bdf7. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #3854, opened 2025-09-04
Postgres major version upgrades

shikharbhardwaj:
Will raise a separate PR for the API, UI and feature flag changes related to this feature, also there's a TODO for the extension upgrades scripts that I will add in meanwhile.

## Testing
- [x] Unit tests
- [x] Validate upgrades for different environments and setups

Provider | HA |  Flavor | Done
---------|----|--|----
Hetzner | None | default | ✅
Hetzner | async | default | ✅
Hetzner | sync | default | ✅
Hetzner | async | paradedb | ✅
AWS, x86 | None | default | ✅
AWS, x86 | async | default | ✅
AWS, arm | async | default | ✅
AWS, arm | sync | default | 🚧
AWS, arm | none | default | 🚧

----

> [!IMPORTANT]
> Introduces major version upgrade support for Postgres, including version management, upgrade processes, and extensive test coverage.
> 
>   - **Behavior**:
>     - Adds support for Postgres major version upgrades, including version management and upgrade processes.
>     - Introduces `target_version` field in `postgres_resource` and `postgres_server` to manage version upgrades.
>     - Implements upgrade logic in `ConvergePostgresResource` and `PostgresServerNexus`.
>     - Handles upgrade failures and logs errors in `ConvergePostgresResource`.
>   - **Migrations**:
>     - Adds `20250925_add_postgres_server_version.rb` to introduce `version` and `target_version` columns.
>     - Adds `20250926_backfill_postgres_versions.rb` to backfill and set version columns as not null.
>     - Adds `20250927_remove_postgres_resource_version.rb` to remove old `version` column.
>   - **Models**:
>     - Updates `PostgresResource` and `PostgresServer` to include version handling logic.
>     - Adds `upgrade_candidate_server` method to `PostgresResource`.
>   - **Utilities**:
>     - Adds `PostgresSetup` and `PostgresUpgrade` classes for managing setup and upgrade processes.
>     - Implements `refresh_walg_credentials` and `switch_to_new_timeline` in `PostgresServer`.
>   - **Tests**:
>     - Extensive test coverage for new upgrade logic in `spec/prog/postgres` and `spec/model/postgres`.
>     - Tests for `PostgresUpgrade` in `spec/rhizome/postgres/postgres_upgrade_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c069cedd8c2020952f39975e4fda998734df7a01. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -104,7 +104,7 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
   end
 
   label def mount_data_disk
-    case vm.sshable.cmd("common/bin/daemonizer --check format_disk")
+    case vm.sshable.d_check("format_disk")
     when "Succeeded"
       storage_device_paths = postgres_server.storage_device_paths
       device_path = if storage_device_paths.count > 1
@@ -123,10 +123,10 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
     when "Failed", "NotStarted"
       storage_device_paths = postgres_server.storage_device_paths
       if storage_device_paths.count == 1
-        vm.sshable.cmd("common/bin/daemonizer 'sudo mkfs --type ext4 #{storage_device_paths.first}' format_disk")
+        vm.sshable.d_run("format_disk", "sudo", "mkfs", "--type", "ext4", storage_device_paths.first)
       else
         vm.sshable.cmd("sudo mdadm --create --verbose /dev/md0 --level=0 --raid-devices=#{storage_device_paths.count} #{storage_device_paths.join(" ")}")
-        vm.sshable.cmd("common/bin/daemonizer 'sudo mkfs --type ext4 /dev/md0' format_disk")
+        vm.sshable.d_run("format_disk", "sudo", "mkfs", "--type", "ext4", "/dev/md0")
       end
     end
 
@@ -140,18 +140,18 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
   end
 
   label def initialize_empty_database
-    case vm.sshable.cmd("common/bin/daemonizer --check initialize_empty_database")
+    case vm.sshable.d_check("initialize_empty_database")
     when "Succeeded"
       hop_refresh_certificates
     when "Failed", "NotStarted"
-      vm.sshable.cmd("common/bin/daemonizer 'sudo postgres/bin/initialize-empty-database #{postgres_server.version}' initialize_empty_database")
+      vm.sshable.d_run("initialize_empty_database", "sudo", "postgres/bin/initialize-empty-database", postgres_server.version)
     end
 
     nap 5
   end
 
   label def initialize_database_from_backup
-    case vm.sshable.cmd("common/bin/daemonizer --check initialize_database_from_backup")
+    case vm.sshable.d_check("initialize_database_from_backup")
     when "Succeeded"
       hop_refresh_certificates
     when "Failed", "NotStarted"
@@ -160,7 +160,7 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
       else
         postgres_server.timeline.latest_backup_label_before_target(target: postgres_server.resource.restore_target)
       end
-      vm.sshable.cmd("common/bin/daemonizer 'sudo postgres/bin/initialize-database-from-backup #{postgres_server.version} #{backup_label}' initialize_database_from_backup")
+      vm.sshable.d_run("initialize_database_from_backup", "sudo", "postgres/bin/initialize-database-from-backup", postgres_server.version, backup_label)
     end
 
     nap 5
@@ -332,9 +332,9 @@ CONFIG
   end
 
   label def configure
-    case vm.sshable.cmd("common/bin/daemonizer --check configure_postgres")
+    case vm.sshable.d_check("configure_postgres")
     when "Succeeded"
-      vm.sshable.cmd("common/bin/daemonizer --clean configure_postgres")
+      vm.sshable.d_clean("configure_postgres")
 
       when_initial_provisioning_set? do
         hop_update_superuser_password if postgres_server.primary?
@@ -346,7 +346,7 @@ CONFIG
       hop_wait
     when "Failed", "NotStarted"
       configure_hash = postgres_server.configure_hash
-      vm.sshable.cmd("common/bin/daemonizer 'sudo postgres/bin/configure #{postgres_server.version}' configure_postgres", stdin: JSON.generate(configure_hash))
+      vm.sshable.d_run("configure_postgres", "sudo", "postgres/bin/configure", postgres_server.version, stdin: JSON.generate(configure_hash))
     end
 
     nap 5
@@ -614,7 +614,7 @@ SQL
       hop_configure
     end
 
-    case vm.sshable.cmd("common/bin/daemonizer --check promote_postgres")
+    case vm.sshable.d_check("promote_postgres")
     when "Succeeded"
       postgres_server.update(timeline_access: "push", representative_at: Time.now, synchronization_status: "ready")
       postgres_server.resource.incr_refresh_dns_record
@@ -624,7 +624,7 @@ SQL
       postgres_server.resource.servers.reject(&:primary?).each { it.update(synchronization_status: "catching_up") }
       hop_configure
     when "Failed", "NotStarted"
-      vm.sshable.cmd("common/bin/daemonizer 'sudo postgres/bin/promote #{postgres_server.version}' promote_postgres")
+      vm.sshable.d_run("promote_postgres", "sudo", "postgres/bin/promote", postgres_server.version)
       nap 0
     end
 
--- a/spec/prog/postgres/postgres_server_nexus_spec.rb
+++ b/spec/prog/postgres/postgres_server_nexus_spec.rb
@@ -241,23 +241,23 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
   describe "#mount_data_disk" do
     it "formats data disk if format command is not sent yet or failed" do
       expect(postgres_server).to receive(:storage_device_paths).and_return(["/dev/vdb"])
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo mkfs --type ext4 /dev/vdb' format_disk")
+      expect(sshable).to receive(:d_run).with("format_disk", "sudo", "mkfs", "--type", "ext4", "/dev/vdb")
 
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check format_disk").and_return("NotStarted")
+      expect(sshable).to receive(:d_check).with("format_disk").and_return("NotStarted")
       expect { nx.mount_data_disk }.to nap(5)
     end
 
     it "formats data disk correctly when there are multiple storage volumes" do
       expect(postgres_server).to receive(:storage_device_paths).and_return(["/dev/nvme1n1", "/dev/nvme2n1"])
       expect(sshable).to receive(:cmd).with("sudo mdadm --create --verbose /dev/md0 --level=0 --raid-devices=2 /dev/nvme1n1 /dev/nvme2n1")
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo mkfs --type ext4 /dev/md0' format_disk")
+      expect(sshable).to receive(:d_run).with("format_disk", "sudo", "mkfs", "--type", "ext4", "/dev/md0")
 
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check format_disk").and_return("NotStarted")
+      expect(sshable).to receive(:d_check).with("format_disk").and_return("NotStarted")
       expect { nx.mount_data_disk }.to nap(5)
     end
 
     it "mounts data disk if format disk is succeeded and hops to configure_walg_credentials" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check format_disk").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("format_disk").and_return("Succeeded")
       expect(sshable).to receive(:cmd).with("sudo mkdir -p /dat")
       expect(sshable).to receive(:cmd).with("sudo common/bin/add_to_fstab /dev/vdb /dat ext4 defaults 0 0")
       expect(sshable).to receive(:cmd).with("sudo mount /dev/vdb /dat")
@@ -265,7 +265,7 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
     end
 
     it "mounts data disk correctly when there are multiple storage volumes" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check format_disk").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("format_disk").and_return("Succeeded")
       expect(postgres_server).to receive(:storage_device_paths).and_return(["/dev/nvme1n1", "/dev/nvme2n1"])
       expect(sshable).to receive(:cmd).with("sudo mdadm --detail --scan | sudo tee -a /etc/mdadm/mdadm.conf")
       expect(sshable).to receive(:cmd).with("sudo update-initramfs -u")
@@ -276,7 +276,7 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
     end
 
     it "naps if script return unknown status" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check format_disk").and_return("Unknown")
+      expect(sshable).to receive(:d_check).with("format_disk").and_return("Unknown")
       expect { nx.mount_data_disk }.to nap(5)
     end
   end
@@ -298,24 +298,24 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
   describe "#initialize_empty_database" do
     it "triggers initialize_empty_database if initialize_empty_database command is not sent yet or failed" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo postgres/bin/initialize-empty-database 16' initialize_empty_database").twice
+      expect(sshable).to receive(:d_run).with("initialize_empty_database", "sudo", "postgres/bin/initialize-empty-database", "16").twice
 
       # NotStarted
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_empty_database").and_return("NotStarted")
+      expect(sshable).to receive(:d_check).with("initialize_empty_database").and_return("NotStarted")
       expect { nx.initialize_empty_database }.to nap(5)
 
       # Failed
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_empty_database").and_return("Failed")
+      expect(sshable).to receive(:d_check).with("initialize_empty_database").and_return("Failed")
       expect { nx.initialize_empty_database }.to nap(5)
     end
 
     it "hops to refresh_certificates if initialize_empty_database command is succeeded" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_empty_database").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("initialize_empty_database").and_return("Succeeded")
       expect { nx.initialize_empty_database }.to hop("refresh_certificates")
     end
 
     it "naps if script return unknown status" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_empty_database").and_return("Unknown")
+      expect(sshable).to receive(:d_check).with("initialize_empty_database").and_return("Unknown")
       expect { nx.initialize_empty_database }.to nap(5)
     end
   end
@@ -325,31 +325,31 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
       expect(postgres_server.resource).to receive(:restore_target).and_return(Time.now).twice
       expect(postgres_server.timeline).to receive(:latest_backup_label_before_target).and_return("backup-label").twice
       expect(postgres_server).to receive(:standby?).and_return(false).twice
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo postgres/bin/initialize-database-from-backup 16 backup-label' initialize_database_from_backup").twice
+      expect(sshable).to receive(:d_run).with("initialize_database_from_backup", "sudo", "postgres/bin/initialize-database-from-backup", "16", "backup-label").twice
 
       # NotStarted
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_database_from_backup").and_return("NotStarted")
+      expect(sshable).to receive(:d_check).with("initialize_database_from_backup").and_return("NotStarted")
       expect { nx.initialize_database_from_backup }.to nap(5)
 
       # Failed
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_database_from_backup").and_return("Failed")
+      expect(sshable).to receive(:d_check).with("initialize_database_from_backup").and_return("Failed")
       expect { nx.initialize_database_from_backup }.to nap(5)
     end
 
     it "hops to refresh_certificates if initialize_database_from_backup command is succeeded" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_database_from_backup").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("initialize_database_from_backup").and_return("Succeeded")
       expect { nx.initialize_database_from_backup }.to hop("refresh_certificates")
     end
 
     it "naps if script return unknown status" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_database_from_backup").and_return("Unknown")
+      expect(sshable).to receive(:d_check).with("initialize_database_from_backup").and_return("Unknown")
       expect { nx.initialize_database_from_backup }.to nap(5)
     end
 
     it "triggers initialize_database_from_backup with LATEST as backup_label for standbys" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check initialize_database_from_backup").and_return("NotStarted")
+      expect(sshable).to receive(:d_check).with("initialize_database_from_backup").and_return("NotStarted")
       expect(postgres_server).to receive(:standby?).and_return(true)
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo postgres/bin/initialize-database-from-backup 16 LATEST' initialize_database_from_backup")
+      expect(sshable).to receive(:d_run).with("initialize_database_from_backup", "sudo", "postgres/bin/initialize-database-from-backup", "16", "LATEST")
       expect { nx.initialize_database_from_backup }.to nap(5)
     end
   end
@@ -512,29 +512,29 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
   describe "#configure" do
     it "triggers configure if configure command is not sent yet or failed" do
       expect(postgres_server).to receive(:configure_hash).and_return("dummy-configure-hash").twice
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo postgres/bin/configure 16' configure_postgres", stdin: JSON.generate("dummy-configure-hash")).twice
+      expect(sshable).to receive(:d_run).with("configure_postgres", "sudo", "postgres/bin/configure", "16", stdin: JSON.generate("dummy-configure-hash")).twice
 
       # NotStarted
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("NotStarted")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("NotStarted")
       expect { nx.configure }.to nap(5)
 
       # Failed
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Failed")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Failed")
       expect { nx.configure }.to nap(5)
     end
 
     it "hops to update_superuser_password if configure command is succeeded during the initial provisioning and if the server is primary" do
       expect(nx).to receive(:when_initial_provisioning_set?).and_yield
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --clean configure_postgres").and_return("Succeeded")
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_clean).with("configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Succeeded")
       expect(postgres_server).to receive(:primary?).and_return(true)
       expect { nx.configure }.to hop("update_superuser_password")
     end
 
     it "hops to wait_catch_up if configure command is succeeded during the initial provisioning and if the server is standby" do
       expect(nx).to receive(:when_initial_provisioning_set?).and_yield
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --clean configure_postgres").and_return("Succeeded")
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_clean).with("configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Succeeded")
       expect(postgres_server).to receive(:primary?).and_return(false)
       expect(postgres_server).to receive(:standby?).and_return(true)
       expect { nx.configure }.to hop("wait_catch_up")
@@ -542,8 +542,8 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
     it "hops to wait_recovery_completion if configure command is succeeded during the initial provisioning and if the server is doing pitr" do
       expect(nx).to receive(:when_initial_provisioning_set?).and_yield
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --clean configure_postgres").and_return("Succeeded")
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_clean).with("configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Succeeded")
       expect(postgres_server).to receive(:primary?).and_return(false)
       expect(postgres_server).to receive(:standby?).and_return(false)
       expect { nx.configure }.to hop("wait_recovery_completion")
@@ -551,16 +551,16 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
     it "hops to wait for primaries if configure command is succeeded at times other than the initial provisioning" do
       expect(nx).to receive(:when_initial_provisioning_set?)
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --clean configure_postgres").and_return("Succeeded")
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_clean).with("configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Succeeded")
       expect(postgres_server).to receive(:standby?).and_return(false)
       expect { nx.configure }.to hop("wait")
     end
 
     it "hops to wait_catchup for standbys if configure command is succeeded at times other than the initial provisioning" do
       expect(nx).to receive(:when_initial_provisioning_set?)
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --clean configure_postgres").and_return("Succeeded")
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_clean).with("configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Succeeded")
       expect(postgres_server).to receive(:standby?).and_return(true)
       expect(postgres_server).to receive(:synchronization_status).and_return("catching_up")
       expect { nx.configure }.to hop("wait_catch_up")
@@ -568,8 +568,8 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
     it "hops to wait for read replicas if configure command is succeeded" do
       expect(nx).to receive(:when_initial_provisioning_set?).and_yield
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --clean configure_postgres").and_return("Succeeded")
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_clean).with("configure_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Succeeded")
       expect(postgres_server).to receive(:primary?).and_return(false)
       expect(postgres_server).to receive(:standby?).and_return(false)
       expect(postgres_server).to receive(:read_replica?).and_return(true)
@@ -577,7 +577,7 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
     end
 
     it "naps if script return unknown status" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check configure_postgres").and_return("Unknown")
+      expect(sshable).to receive(:d_check).with("configure_postgres").and_return("Unknown")
       expect { nx.configure }.to nap(5)
     end
   end
@@ -935,15 +935,15 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
   describe "#taking_over" do
     it "triggers promote if promote command is not sent yet or failed" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer 'sudo postgres/bin/promote 16' promote_postgres").twice
+      expect(sshable).to receive(:d_run).with("promote_postgres", "sudo", "postgres/bin/promote", "16").twice
 
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check promote_postgres").and_return("NotStarted", "Failed")
+      expect(sshable).to receive(:d_check).with("promote_postgres").and_return("NotStarted", "Failed")
       expect { nx.taking_over }.to nap(0)
       expect { nx.taking_over }.to nap(0)
     end
 
     it "updates the metadata and hops to configure if promote command is succeeded" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check promote_postgres").and_return("Succeeded")
+      expect(sshable).to receive(:d_check).with("promote_postgres").and_return("Succeeded")
 
       expect(postgres_server).to receive(:update).with(timeline_access: "push", representative_at: anything, synchronization_status: "ready")
       expect(postgres_server.resource).to receive(:incr_refresh_dns_record)
@@ -964,7 +964,7 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
     end
 
     it "naps if script return unknown status" do
-      expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check promote_postgres").and_return("Unknown")
+      expect(sshable).to receive(:d_check).with("promote_postgres").and_return("Unknown")
       expect { nx.taking_over }.to nap(5)
     end
 


prog/postgres/postgres_resource_nexus.rb
```diff
@@ -189,7 +189,20 @@ def before_run
...
+      candidate_server_id = Prog::Postgres::PostgresServerNexus.assemble(
+        resource_id: postgres_resource.id,
+        timeline_id: postgres_resource.timeline.id,
+        timeline_access: "fetch"
+      ).id
+      new_timeline_id = Prog::Postgres::PostgresTimelineNexus.assemble(
+        location_id: postgres_resource.location_id
+      ).id
+
+      bud Prog::Postgres::UpgradePostgresResource, {"candidate_server_id" => candidate_server_id, "new_timeline_id" => new_timeline_id}, :start
```
shikharbhardwaj: Is there a cleaner way to bud a Prog from here with these values passed in? I thought of .assemble + updating children but that does not seem like a good pattern.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -21,12 +21,12 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
       else
         flavor_suffix = case postgres_resource.flavor
         when PostgresResource::Flavor::STANDARD then ""
-        when PostgresResource::Flavor::PARADEDB then "-paradedb"
-        when PostgresResource::Flavor::LANTERN then "-lantern"
+        when PostgresResource::Flavor::PARADEDB then "#{postgres_resource.version}-paradedb"
+        when PostgresResource::Flavor::LANTERN then "#{postgres_resource.version}-lantern"
```
shikharbhardwaj: Will update these 2 images as well to add the package lists at least

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -57,9 +57,7 @@ def before_run
   end
 
   label def wait_leader
-    hop_destroy if postgres_timeline.leader.nil?
```
shikharbhardwaj, line 60: Not sure how important this was, but this made the timeline disappear before we could upgrade the candidate and attach it to the new timeline.

model/postgres/postgres_resource.rb
```diff
@@ -137,6 +137,10 @@ def ongoing_failover?
     servers.any? { it.taking_over? }
   end
 
+  def needs_upgrade?
+    !read_replica? && !ongoing_failover? && (representative_server&.version&.to_i&.< desired_version.to_i) || false
```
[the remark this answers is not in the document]
shikharbhardwaj: This is a rubocop rule, to not chain unsafe method calls after the safe navigation operator.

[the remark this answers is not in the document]
jeremyevans: Rubocop is being stupid here, as `to_i` is defined on all objects, including `nil`.

That being said, might be worth adding a `PostgresServer#version_int` method that handles the conversion.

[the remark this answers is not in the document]
shikharbhardwaj: Its complaining about the comparison (`&.<` vs ` < `), which does not work with nils
Makes sense to have a conversion method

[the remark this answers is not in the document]
jeremyevans: Note that `nil.to_i` is `0`, so ellipsis is somewhat correct that `maybe_nil.to_i > x` will not cause a NoMethodError,. However, ellipsis's suggestion is wrong, since it leaves the `&.` before `to_i` and therefore will have problems.  It can be `representative_server&.version.to_i < desired_version.to_i`.

That all being said, I'm not sure if you want it to be that way, since if there is no `representative_server`, I think you want the expression to return false, while if you you use `representative_server&.version.to_i < desired_version.to_i`, the expression will likely return true if `representative_server` is nil. So maybe it's best to continue using the current approach (adding the `version_int` conversion method is probably useful anyway), or checking for a missing representative server earlier and avoiding the use of `&.`

```suggestion
    !read_replica? && !ongoing_failover? && !representative_server.nil? && representative_server.version.to_i < desired_version.to_i
```

migrate/20250901_pg_upgrade_versions.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    alter_table(:postgres_server) do
+      add_column :version, :postgres_version, null: false, default: "16"
```
jeremyevans: Adding a default seems like a bad idea, as it will be wrong for servers/resources set up for PostgreSQL 17. I would add without a default and `null: false`, set the version correctly for all `postgres_server` rows using a lookup on `postgres_resource.version`, then use `set_column_not_null`.

I think we should move away from using the `postgres_version` enum type, and any other enum types that will need to be modified. Adding an entry to an enum type (e.g. for PostgreSQL 18) cannot be reverted, and so results in irreversible migrations. It's possible to implement the same checks that an enum type would add by using a CHECK constraint (which can be modified in later migrations) or using a separate table holding the versions and having the column be a foreign key to that table.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -53,7 +53,8 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
         timeline_access: timeline_access,
         representative_at: representative_at,
         synchronization_status: synchronization_status,
-        vm_id: vm_st.id
+        vm_id: vm_st.id,
+        version: postgres_resource.version
       ) { it.id = ubid.to_uuid }
```
jeremyevans: Unrelated to this PR, but you can switch the previous `create(...)` to `create_with_id(ubid.to_uuid, ...)` and remove this block.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -546,10 +547,21 @@ def switch_to_new_timeline
...
-    postgres_server.vm.sshable.cmd("sudo postgres/bin/lockout #{postgres_server.resource.version}")
-    postgres_server.vm.sshable.cmd("sudo pg_ctlcluster #{postgres_server.resource.version} main stop -m smart")
+    postgres_server.vm.sshable.cmd("sudo postgres/bin/lockout #{postgres_server.version}")
+    postgres_server.vm.sshable.cmd("sudo pg_ctlcluster #{postgres_server.version} main stop -m smart")
 
-    nap 6 * 60 * 60
+    hop_wait_fence
+  end
+
+  label def wait_fence
```
jeremyevans: In regards to naming, there are multiple reasons to fence. Is it possible we could change `fence` to something more descriptive of the purpose of the fencing?

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
...
+    postgres_resource.servers.reject { it.version == postgres_resource.desired_version }.each { it.incr_destroy }
+    hop_finish_upgrade
+  end
+
+  label def finish_upgrade
+    pop "upgrade prog finished"
+  end
+
+  label def upgrade_failed
+    unless candidate_server.nil?
```
jeremyevans:
```suggestion
    if candidate_server
```

Unless you care about the `nil`/`false` distinction, it's better to avoid `nil?` calls if you can use the object directly.

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
...
+      logs = candidate_server.vm.sshable.cmd("sudo journalctl -u upgrade_postgres")
+      logs.split("\n").each { |line| Clog.emit("Postgres resource upgrade failed") { {resource_id: postgres_resource.id, log: line} } }
```
jeremyevans: Is there a reason to emit many lines in separate emit calls, instead of emitting all lines at once? Maybe there is and I'm not aware of it.

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    new_timeline&.incr_destroy
+
+    postgres_resource.representative_server.incr_unfence if postgres_resource.representative_server.strand.label == "wait_fence"
+    nap 6 * 60 * 60
+  end
+
+  def candidate_server
+    PostgresServer[frame["candidate_server_id"]]
```
jeremyevans: These values aren't cached, meaning every call to the method will do a lookup. That seems bad when you are calling the methods multiple times in the same label.

rhizome/postgres/lib/postgres_setup.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require "logger"
+
+class PostgresSetup
+  def initialize(version, logger)
+    @version = version
+    @logger = logger
```
jeremyevans: `@logger` appears to be unused.

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -0,0 +1,118 @@
...
+    pg_setup.create_cluster
+  end
+
+  def stop_new_version
+    r "sudo systemctl stop postgresql@#{@version}-main"
+  end
+
+  def run_check
+    Dir.chdir("/dat/upgrade/#{@version}") do
+      r "sudo -u postgres /usr/lib/postgresql/#{@version}/bin/pg_upgrade --old-bindir /usr/lib/postgresql/#{@prev_version}/bin --old-datadir /etc/postgresql/#{@prev_version}/main/ --new-bindir /usr/lib/postgresql/#{@version}/bin --new-datadir /etc/postgresql/#{@version}/main/ --check"
```
jeremyevans: Move the code to a method that accepts either `"--check"` or `"--link"`, and call that method from `run_check` and `run_pg_upgrade`.

migrate/20250901_pg_upgrade_versions.rb
shikharbhardwaj: Makes sense. JW, what is the convention around this type of (incompatible) schema change? Do I still keep them in separate commits (only migration + schema in one commit and test/app fixes in the other) or combine the minimal set of fixes to make things run without crashing in one commit?

jeremyevans: Ideally, you should ensure compatibility at all points by splitting the task into steps. For example, to replace a column with a new column, you would do:

1. Add new column using migration
2. Change code to populate new column (and continue populating old column), but still read values from old column
3. Run migration/script to populate new column for existing entries where it is not populated
4. Change code to read values from new column and stop populating old column
5. Remove now unused old column using migration

So any incompatible schema change should always require at least two migrations, one to add support for the new approach, and one to remove support for the old approach. If a migration is used to populate the new approach, that would be a third migration.

rhizome/postgres/lib/postgres_setup.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def setup_packages
```
byucesoy: install_packages?

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  def create_upgrade_dir
+    r "sudo mkdir -p /dat/upgrade/#{@version}"
+    r "sudo chown postgres:postgres /dat/upgrade/#{@version}"
+  end
+
+  def disable_archive_mode
+    r "echo 'archive_mode = off' | sudo tee /etc/postgresql/#{@prev_version}/main/conf.d/100-upgrade.conf"
+    r "sudo pg_ctlcluster #{@prev_version} main restart"
```
byucesoy: You can update `archive_command` to no-op, (e.g. `true`), which doesn't require restart. Reload would be enough in that case.

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: I tried to do this, but it failed a few tests with the following
```
  3) Prog::Postgres::PostgresServerNexus.assemble picks correct base image for AWS-pg16
     Failure/Error: obj.save_changes

     Sequel::ValidationFailed:
       vm_id is invalid
     # ./lib/resource_methods.rb:158:in 'ResourceMethods::ClassMethods#create_with_id'
     # ./prog/postgres/postgres_server_nexus.rb:50:in 'block in Prog::Postgres::PostgresServerNexus.assemble'
     # ./prog/postgres/postgres_server_nexus.rb:14:in 'Prog::Postgres::PostgresServerNexus.assemble'
     # ./spec/prog/postgres/postgres_server_nexus_spec.rb:111:in 'block (3 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Sequel::ValidationFailed:
     #   vm_id is invalid
     #   ./lib/resource_methods.rb:158:in 'ResourceMethods::ClassMethods#create_with_id'
```

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Hmm, in the current case, there is really only one reason why we fence the primary, for another server to take over. With the current change, now there are 2 ways in which it can be triggered: one for takeover during scaleup/down, and the other one now is an upgrade.
There are other ways to figure out why the server is fenced (like looking at child progs of the resource nexus), so adding 2 separate labels for these seems odd, wdyt?

prog/postgres/upgrade_postgres_resource.rb
shikharbhardwaj: Mainly for cleaner line-wrapping in the mezmo UI, in case the upgrade logs are long

migrate/20250901_pg_upgrade_versions.rb
shikharbhardwaj: I added the different migrations to do this in various steps, with the 3 migrations

prog/postgres/postgres_server_nexus.rb
jeremyevans: Quite a strange error. If the vm_id was invalid, it should be invalid regardless of whether you call `create(...) { it.id = ubid.to_uuid  }` or `create_with_id(ubid.to_uuid, ...)`.

prog/postgres/postgres_server_nexus.rb
jeremyevans: I guess if we can fence for multiple reasons, and the fencing is identical in every case, it's fine to keep the current naming.

migrate/20250910_pg_upgrade_versions.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    alter_table(:postgres_server) do
+      add_column :version, :postgres_version, null: true
```
jeremyevans: `null: true` is the default, so you can omit this if you want.

Additionally, this still uses the `postgres_version` enum type. As mentioned in a previous comment, I think we should move away from that, before we support upgrades to PostgreSQL 18. Otherwise, adding support for PostgreSQL 18 will require an irreversible migration. How do you plan to address this issue?

migrate/20250910_pg_upgrade_versions.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    alter_table(:postgres_resource) do
+      add_column :desired_version, :postgres_version, null: true
+    end
+
+    run "UPDATE postgres_resource SET desired_version = version"
```
jeremyevans: You have this in a later commit, it doesn't need also be in this commit. Removing this also allows you to use a `change` migration.

migrate/20250911_pg_upgrade_backfill_version.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    run "UPDATE postgres_resource SET desired_version = version"
+    run "UPDATE postgres_server SET version = postgres_resource.desired_version FROM postgres_resource WHERE postgres_resource.id = postgres_server.resource_id"
```
jeremyevans: May want to do this only when the column is NULL. This migration shouldn't be run until after the code is already populating the column for new instances.

migrate/20250911_pg_upgrade_backfill_version.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
```
jeremyevans: This cannot be a `change` migration, as `run` is not reversible.  You should switch to `add` (a missing `down` block is assumed to be a no-op).

migrate/20250912_pg_remove_version.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
```
jeremyevans, line 4: This should be a `revert` migration that adds the column. As-is, this is not a reversible migration, because `drop_column` is not reversible as you do not know the type and other column information.

migrate/20250912_pg_remove_version.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
```
jeremyevans, line 3: Taking these 3 migrations into consideration, it seems like one of the end results is you are renaming `postgres_resource.version` to `postgres_resource.desired_version`. Is the rename that important? It sounds like you could just add `postgres_server.version`, and treat `postgres_resource.version` as the desired version.  No objections to renaming the column if you think it is important to do so.

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -0,0 +1,130 @@
...
+  end
+
+  def stop_new_version
+    r "sudo systemctl stop postgresql@#{@version}-main"
+  end
+
+  def run_check
+    Dir.chdir("/dat/upgrade/#{@version}") do
+      r pg_upgrade_cmdline("--check")
+    end
```
jeremyevans: This `Dir.chdir` is still duplicated between `run_check` and `run_pg_upgrade`. Maybe add something like:

```ruby
  def run_pg_upgrade_cmd(arg)
    Dir.chdir("/dat/upgrade/#{@version}") do
      r pg_upgrade_cmdline(arg)
    end
  end
```

and use that.

migrate/20250912_pg_remove_version.rb
shikharbhardwaj: I wanted to have a distinction between "desired" and "actual" version here, ideally with the `version` field in the API reflecting the running server version.  If we reuse the current field as the desired version, then during the upgrade, the API will show the next version, but the clients will still connect to the older version until it completes. Also if the upgrade fails, the version field still displays the next version which seems misleading.

Another way to do this is to keep the existing version column as the "desired" version but add a new property `current_version` to the model, which refers to the representative server to fetch the actually running version. Its simpler because we won't need any migrations but still somewhat confusing. wdyt?

migrate/20250910_pg_upgrade_versions.rb
shikharbhardwaj: I think the migration away from the enum to a CHECK constraint can be done when we add the new version, which will be some time down the line, wdyt?

migrate/20250912_pg_remove_version.rb
jeremyevans: It makes sense to have separate columns, one for desired version and one for current version, assuming you want to track those two things separately.  But in that case, why are you dropping the column in the third migration? Either the columns track two separate things, and you need both columns, or they track only one thing, in which case this is basically a rename. The approach given would make sense if you only need to track two separate things temporarily, and then going forward only need to track one thing. Is that what is happening? From the issues you are describing (upgrade failures, etc.), I don't get that impression.

migrate/20250910_pg_upgrade_versions.rb
jeremyevans: It can be done later, it doesn't have to be done now. However, that means there will more columns using the enum, so the approach adds technical debt to be dealt with later.

migrate/20250912_pg_remove_version.rb
shikharbhardwaj: The current version is not a property of the resource, but of the primary server attached with the resource, so it is tracked by the version column in `postgres_server`. I think keeping the info in both models is redundant and might cause problems, so keeping it separate seems better.

jeremyevans: I understand the need for the column in postgres_server. When I was discussing the separate columns, I was discussing the two columns in postgres_resource, version and desired_version. Apologies for not making that clear. Can you respond to the previous comment with that in mind?

shikharbhardwaj: The columns don't track separate things, the change is basically a rename. Do you have another way in mind to do a rename more easily?

jeremyevans: Thanks for confirming they don't track separate things, and this is just a rename. As I asked in an earlier comment: "Is the rename that important? It sounds like you could just add postgres_server.version, and treat postgres_resource.version as the desired version. No objections to renaming the column if you think it is important to do so."

shikharbhardwaj: As I mentioned earlier, the main result I want here is for the users to have a way to distinguish between the 2 versions in API responses. Here are examples for the 2 options I described above in more detail for a PG resource that is undergoing an upgrade:

```
GET /.../postgres/example-pg
{
  ...
  "version": "16",
  "desired_version": "17",
  ...
}
```

vs

```
GET /.../postgres/example-pg
{
  ...
  "current_version": "16",
  "version": "17",
  ...
}
```

where the current version is sourced like this:

```ruby
class PostgresResource < Sequel::Model
  ...
  def version
    representative_server&.version || desired_version
  end
  ...
end
```

Personally the 1st one seems clearer to me than the 2nd option. To achieve this we can either do a column rename or just modify the serializer to rename the returned field. Between these two, I slightly prefer keeping the API response fields same as the data model as much as possible. wdyt?

jeremyevans: Updating the serializer is much less invasive, but if you think it is important to rename the column, I'm fine with it. I just wanted to make sure you were aware that the end result you want to achieve from a user perspective the is possible with a one line change to the serializer, as opposed to multiple migrations.

shikharbhardwaj: Makes sense, i think the complexity of deploying this coupled with the future change to make version a constraint instead of enum, I decided to just keep things in the serializer instead of making multiple migrations.

migrate/20250910_pg_upgrade_versions.rb
shikharbhardwaj: I decided to remove these migrations to just keep the rename in the serializer

prog/postgres/postgres_resource_nexus.rb
byucesoy: There is no cleaner way. Passing values using frame is OK. This is how we do it when we need to pass values to budded Prog already.

prog/postgres/postgres_timeline_nexus.rb
byucesoy: Normally, timeline and the leader server is created in the same transaction, which prevents unintentional disappearing. I guess the upgrade path needs to do it in separate transactions.

Removing this line is OK since we have a more general timeline cleaning logic implemented now.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -53,7 +53,8 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
         timeline_access: timeline_access,
         representative_at: representative_at,
         synchronization_status: synchronization_status,
-        vm_id: vm_st.id
+        vm_id: vm_st.id,
+        version: postgres_resource.current_version
```
byucesoy: `current_version` is not defined yet.

rhizome/postgres/bin/initialize-empty-database
```diff
@@ -11,11 +12,10 @@ v = ARGV[0]
 
 r "chown postgres /dat"
```
byucesoy: This line can be removed as well as it is part of `setup_data_directory`.

rhizome/postgres/lib/postgres_setup.rb
```diff
@@ -0,0 +1,33 @@
...
+    r "rm -rf /etc/postgresql/#{@version}"
+
+    r "echo \"data_directory = '/dat/#{@version}/data'\" | sudo tee /etc/postgresql-common/createcluster.d/data-dir.conf"
+  end
+
+  def create_cluster
+    r "pg_createcluster #{@version} main --port=5432 --start --locale=C.UTF8"
+  end
+
+  def teardown
```
byucesoy: This doesn't seem to be used.

model/postgres/postgres_server.rb
```diff
@@ -240,7 +240,7 @@ def check_pulse(session:, previous_pulse:)
         end
       end
 
-      if pulse[:reading] == "down" && pulse[:reading_rpt] > 5 && Time.now - pulse[:reading_chg] > 30 && !reload.checkup_set?
+      if pulse[:reading] == "down" && pulse[:reading_rpt] > 5 && Time.now - pulse[:reading_chg] > 30 && !reload.checkup_set? && !resource.needs_upgrade?
```
byucesoy: Why do we stop incrementing checkup when resource needs upgrade?

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -191,7 +191,20 @@ def before_run
...
     reap(fallthrough: true)
 
-    if postgres_resource.needs_convergence? && strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").empty?
+    if postgres_resource.needs_upgrade? && strand.children_dataset.where(prog: "Postgres::UpgradePostgresResource").empty?
+      candidate_server_id = Prog::Postgres::PostgresServerNexus.assemble(
+        resource_id: postgres_resource.id,
+        timeline_id: postgres_resource.timeline.id,
+        timeline_access: "fetch"
+      ).id
+      new_timeline_id = Prog::Postgres::PostgresTimelineNexus.assemble(
```
byucesoy: It might be better idea to create this timeline just before attaching the server to it. Similar to what we do in `switch_to_new_timeline`. You can even move that method to `postgres_server` if it makes sense to re-use it.

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -191,7 +191,20 @@ def before_run
   label def wait
     reap(fallthrough: true)
 
-    if postgres_resource.needs_convergence? && strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").empty?
+    if postgres_resource.needs_upgrade? && strand.children_dataset.where(prog: "Postgres::UpgradePostgresResource").empty?
+      candidate_server_id = Prog::Postgres::PostgresServerNexus.assemble(
```
byucesoy: If we are doing this during all upgrade operations, why don't we move this to the upgrade prog?

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -191,7 +191,20 @@ def before_run
...
+        timeline_access: "fetch"
+      ).id
+      new_timeline_id = Prog::Postgres::PostgresTimelineNexus.assemble(
+        location_id: postgres_resource.location_id
+      ).id
+
+      bud Prog::Postgres::UpgradePostgresResource, {"candidate_server_id" => candidate_server_id, "new_timeline_id" => new_timeline_id}, :start
+    end
+
+    if !postgres_resource.needs_upgrade? && postgres_resource.needs_convergence? && strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").empty?
```
byucesoy: So my understanding is that we added this check to prevent both progs to be budded at the same time, which is a fair restriction. Though using `postgres_resource.needs_upgrade?` as mutex feels weird.

Also this prevents budding ConvergePostgresResource when UpgradePostgresResource is active, but not the other way around. Maybe we should make the decision based on the child progs.

Or... maybe we can partially combine both progs. I've left another more detailed comment about this.

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+require_relative "../../lib/util"
+
+class Prog::Postgres::UpgradePostgresResource < Prog::Base
```
byucesoy: This seems to have lots of similarities with convergence prog. In a way, this also does a convergence (convergence of versions).

I'd love to see a version of this PR where;
- We bud ConvergePostgresResource prog instead of UpgradePostgresResource prog when upgrade is required.
- ConvergePostgresResource would be responsible for bringing up a candidate server, still in the old version.
- Then if upgrade is required ConvergePostgresResource can push(?) (simplified version of (i.e. without standby provisioning and failover logic)) UpgradePostgresResource prog.
- If everything goes well, ConvergePostgresResource would trigger failover and prune the servers.

I feel like that version would be cleaner and more fitting to general convergence theme.

byucesoy: There is another benefit of letting ConvergePostgresResource to handle standby provisioning. We might have various restrictions regarding the standby server (such as using different hosts or even different AZs #3720). If we create standbys at different places in the code, we would need to duplicate that logic.

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    nap 5
+  end
+
+  label def update_metadata
+    candidate_server.update(version: postgres_resource.version, timeline_id: new_timeline.id)
+
+    postgres_resource.update(
+      version: postgres_resource.version,
+      timeline: new_timeline
```
byucesoy: Postgres resource does not have timeline. It gets it from the representative server using `one_through_one` association.

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    nap 5
+  end
+
+  label def update_metadata
+    candidate_server.update(version: postgres_resource.version, timeline_id: new_timeline.id)
+
+    postgres_resource.update(
+      version: postgres_resource.version,
```
byucesoy: This assignment seems unnecessary.

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
...
+      timeline: new_timeline
+    )
+    candidate_server.incr_refresh_walg_credentials
+    candidate_server.incr_configure
+    candidate_server.incr_restart
+    candidate_server.incr_unplanned_take_over
```
byucesoy: Why unplanned? Unplanned means due to failure. This is a planned take over, no?

prog/postgres/upgrade_postgres_resource.rb
shikharbhardwaj: Makes sense, I will try to sketch a version with both the progs combined.

model/postgres/postgres_server.rb
shikharbhardwaj: Its because the automated recovery causes issues when we perform restarts of different versions on the candidate standby during upgrade. For example, it brings the older version back up when pg_upgrade needs both versions to be down before the upgrade, or tries to start the older cluster up after upgrade.

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Moved this to a later commit

prog/postgres/upgrade_postgres_resource.rb
shikharbhardwaj: Its planned, but we do the fencing of the representative server earlier within the upgrade prog to prevent writes to the primary before pg_upgrade. Because the primary is already fenced, the candidate gets stuck waiting for the semaphore to be unset if we do a planned failover. I added this as a comment to make it clearer.

migrate/20251007_add_postgres_server_version.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    alter_table(:postgres_server) do
+      add_column :version, :postgres_version
```
fdr, line 6: little puzzled how timelines are not with a version, but servers are. Normally I expect the promotion and upgrade process to begin a new timeline, because, it kinda has to: the replay is non-continuous.

fdr: also seems bad in the case where you might leave a timeline record around with no servers: how do you remember the postgres revision to start? This was a classic way of implementing delayed-deletion, as a timeline with no servers is only archives.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -156,7 +157,7 @@ def before_run
       else
         postgres_server.timeline.latest_backup_label_before_target(target: postgres_server.resource.restore_target)
       end
-      vm.sshable.cmd("common/bin/daemonizer 'sudo postgres/bin/initialize-database-from-backup #{postgres_server.resource.version} #{backup_label}' initialize_database_from_backup")
+      vm.sshable.cmd("common/bin/daemonizer 'sudo postgres/bin/initialize-database-from-backup #{postgres_server.version} #{backup_label}' initialize_database_from_backup")
```
fdr: why not use the daemonizer2 helpers mohi added?

migrate/20251007_add_postgres_server_version.rb
shikharbhardwaj: We do switch to a new timeline after the upgrade, but it makes sense to have it in case we want to use the timeline to start the older version.

byucesoy: For promotions (without upgrade) replay is continuous, right? Postgres creates a new timeline, but also creates history files and WAL replay can follow history files. Do I miss anything?

Probably, it was mistake to call the entity handling backups PostgresTimeline, because it is not exactly a timeline. Contrary to previous iterations, this encompasses multiple postgres timelines.

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: I mostly did a find-and-replace for this change, will update these to use daemonizer2 in a separate commit

migrate/20251007_add_postgres_server_version.rb
fdr: > For promotions (without upgrade) replay is continuous, right? Postgres creates a new timeline, but also creates history files and WAL replay can follow history files. Do I miss anything?

Yes, for promotions replay is continuous. It probes the history files which includes the branch points and then decides when to switch timelines (stored in the first eight bytes of the WAL file name) based on the directive there.

I guess if you are stuffing everything into one prefix, it makes it possible for a downstream database to attack the upstream's archives, but closes the follower hole without more programming, which may be an acceptable trade: it is a design pgbackrest shares, for instance. Previous iterations with wal-e always elected for the follower hole on the principle that I'd fix it "someday" but I never did (my last chance was Citus, given Crunchy was pgbackrest based).

I do not know if wal-e/g and every blob store have adequate safety to handle multiple concurrent standby promotions that may select the same incremented timeline integer without a total disaster of multiple servers overwriting to the same prefix. It certainly wasn't safe when wal-e was designed in the era of S3 Classic. It is a hair's breadth from catastrophic though.

Regardless...each of those postgres timelines should be associated with a version.

prog/postgres/upgrade_postgres_resource.rb
shikharbhardwaj: Updated to fold the upgrade steps into the convergence prog

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Fixed the test for this, it was failing because of a mock on the create call

model/postgres/postgres_resource.rb
```diff
@@ -101,11 +101,19 @@ def target_server_count
...
+      !upgrade_candidate_server.nil?
+    else
+      servers.count { !it.needs_recycling? } >= target_server_count
+    end
   end
 
   def has_enough_ready_servers?
-    servers.count { !it.needs_recycling? && it.strand.label == "wait" } >= target_server_count
+    if needs_upgrade?
+      !upgrade_candidate_server.nil? && upgrade_candidate_server.strand.label == "wait" && upgrade_candidate_server.synchronization_status == "ready"
```
byucesoy: Can this ever be in wait state but not synchronized already?

model/postgres/postgres_resource.rb
```diff
@@ -149,6 +157,29 @@ def incr_restart
     Semaphore.incr(servers_dataset.select(:id), "restart")
   end
 
+  def version_int
+    version.to_i
+  end
+
+  def current_version
```
byucesoy: Semantically, version and current_version means same. However we use them to mean different things in our abstraction.

How about renaming version to target_version, inline with target_vm_size and target_storage_size_gib?

model/postgres/postgres_resource.rb
```diff
@@ -149,6 +157,29 @@ def incr_restart
     Semaphore.incr(servers_dataset.select(:id), "restart")
   end
 
+  def version_int
```
byucesoy: AFAICT, `_int` helpers are only used one place. I think we don't need them. We can just inline them in the caller. Also `_int` is not shorter than `.to_i` :slightly_smiling_face:.

That said, I'm open to change to type to int in the database directly as well. One benefit of string was that it would allow us to store the version with the minor part like `17.1`, but if we are using `to_i` in the code, we lose that benefit anyway.

model/postgres/postgres_resource.rb
```diff
@@ -149,6 +157,29 @@ def incr_restart
...
+  def current_version
+    representative_server&.version || version
+  end
+
+  def current_version_int
+    current_version.to_i
+  end
+
+  def needs_upgrade?
+    !read_replica? && !ongoing_failover? && (current_version_int < version_int) || false
```
byucesoy: This reminded me that... we break read replicas with upgrades, right? Technically, read replica would realize it doesn't make progress and create a new server from a fresh backup, but it would use the old version.

I believe, if you update the version for read replica as well (while updating for the primary in API endpoint), it would eventually coverge. Similar to what we do for scale up/down ([here](https://github.com/ubicloud/ubicloud/blob/d618fbfe02311ed9ef86ad14e39b8c5b800e15f7/routes/project/location/postgres.rb#L88-89))

model/postgres/postgres_resource.rb
```diff
@@ -149,6 +157,29 @@ def incr_restart
...
+  end
+
+  def upgrade_candidate_server
+    servers
+      .reject(&:representative_at)
+      .select { |server| server.vm.vm_storage_volumes.any? { |vol| vol.boot_image.version && vol.boot_image.version >= UPGRADE_IMAGE_MIN_VERSIONS[version] } }
```
byucesoy: When `vol.boot_image.version` is nil? Is this for non-boot disks, if so it would be better to filter for boot disks first and avoid `vol.boot_image.version` check.

model/postgres/postgres_server.rb
```diff
@@ -307,6 +307,24 @@ def taking_over?
     unplanned_take_over_set? || planned_take_over_set? || FAILOVER_LABELS.include?(strand.label)
   end
 
+  def switch_to_new_timeline(use_current_parent: true)
```
byucesoy: Instead of use_current_parent, we can simply pass parent_timeline, whose default is timeline.

```ruby
  def switch_to_new_timeline(parent_id: timeline.id)
    update(
      timeline_id: Prog::Postgres::PostgresTimelineNexus.assemble(location_id: resource.location_id, parent_id:).id,
      timeline_access: "push"
    )

    ...
  end
```

prog/postgres/converge_postgres_resource.rb
```diff
@@ -6,7 +6,7 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
   subject_is :postgres_resource
 
   label def start
-    register_deadline("recycle_representative_server", 2 * 60 * 60)
+    register_deadline("prune_servers", 2 * 60 * 60)
```
byucesoy: Why we are changing the target here?

byucesoy: I see you created a new label for `wait_for_maintenance_window`. In that case, shouldn't we use `wait_for_maintenance_window` as target, not `prune_servers`?

prog/postgres/converge_postgres_resource.rb
```diff
@@ -38,19 +38,82 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
...
   label def wait_servers_to_be_ready
     hop_provision_servers unless postgres_resource.has_enough_fresh_servers?
-
-    hop_recycle_representative_server if postgres_resource.has_enough_ready_servers?
+    hop_wait_for_maintenance_window if postgres_resource.has_enough_ready_servers?
 
     nap 60
   end
 
+  label def wait_for_maintenance_window
```
byucesoy, line 46: We need an escape hatch back to provision servers in case the target changes while we are waiting for the maintenance window.

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -199,7 +199,7 @@ def before_run
   label def wait
     reap(fallthrough: true)
 
-    if postgres_resource.needs_convergence? && strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").empty?
+    if (postgres_resource.needs_upgrade? || postgres_resource.needs_convergence?) && strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").empty?
```
byucesoy: You can embed `needs_upgrade?` check into `needs_convergence?`.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -38,19 +38,82 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
...
+    if postgres_resource.needs_upgrade?
+      postgres_resource.representative_server.incr_fence
+      hop_wait_fence_primary
+    end
+
+    hop_recycle_representative_server
+  end
+
+  label def wait_fence_primary
+    return hop_upgrade_standby if postgres_resource.representative_server.strand.label == "wait_fence"
```
byucesoy: `wait_fence` feels like it is waiting to be fenced. Better name could be `wait_in_fence`.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -38,19 +38,82 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
...
+      hop_upgrade_failed
+    when "NotStarted"
+      upgrade_candidate.vm.sshable.d_run("upgrade_postgres", "sudo", "postgres/bin/upgrade", postgres_resource.version)
+    end
+
+    nap 5
+  end
+
+  label def update_metadata
+    DB.transaction do
```
byucesoy: We don't need `DB.transaction` block.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -38,19 +38,82 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
...
+    end
+
+    upgrade_candidate.incr_refresh_walg_credentials
+    upgrade_candidate.incr_configure
+    upgrade_candidate.incr_restart
+
+    # We do an unplanned take over here because the primary is already fenced
+    # above. With a planned take over, the candidate server would be stuck
+    # waiting for the fence semaphore to be unset.
+    upgrade_candidate.incr_unplanned_take_over
```
byucesoy: Instead of using a semaphore created for different purpose, I think we should modify planned take over to handle this case. For example instead of using `postgres_server.resource.representative_server.fence_set?` check in `wait_fencing_of_old_primary`, we can just check for `postgres_resource.representative_server.strand.label == "wait_in_fence"` like we are doing in this prog.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -38,19 +38,82 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
...
+  end
+
+  label def upgrade_failed
+    if upgrade_candidate && !upgrade_candidate.destroy_set?
+      logs = upgrade_candidate.vm.sshable.cmd("sudo journalctl -u upgrade_postgres")
+      logs.split("\n").each { |line| Clog.emit("Postgres resource upgrade failed") { {resource_id: postgres_resource.id, log: line} } }
+      upgrade_candidate.incr_destroy
+    end
+
+    postgres_resource.representative_server.incr_unfence if postgres_resource.representative_server.strand.label == "wait_fence"
```
byucesoy: It would be good to raise a page here.

prog/postgres/converge_postgres_resource.rb
byucesoy: And after we do that, can't we just hop to `recycle_representative_server` and it would use increment planned_take_over semaphore anyway (via `rs.trigger_failover(mode: "planned")` call)

model/postgres/postgres_resource.rb
shikharbhardwaj: I had renamed it to `desired_version` in an earlier version of this PR (https://github.com/ubicloud/ubicloud/pull/3854#discussion_r2346784955), but then thought about keeping the rename in the serializer to keep the change simple. I was slightly leaning towards the rename anyways, will do that.

model/postgres/postgres_resource.rb
shikharbhardwaj: :ack: will update this in the API endpoint

prog/postgres/converge_postgres_resource.rb
shikharbhardwaj: Done, updated to reuse the recycle path

prog/postgres/upgrade_postgres_resource.rb
shikharbhardwaj: Updated to use the planned failover mode

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -11,8 +11,8 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
   def_delegators :postgres_resource, :servers, :representative_server
 
   def self.assemble(project_id:, location_id:, name:, target_vm_size:, target_storage_size_gib:,
-    version: PostgresResource::DEFAULT_VERSION, flavor: PostgresResource::Flavor::STANDARD,
```
byucesoy, line 14: Any reason to remove default version? I believe most of the changes in this commit (especially the spec changes) are because we are removing the default. Do we need to do that?

prog/postgres/converge_postgres_resource.rb
```diff
@@ -38,19 +38,80 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
...
+
+    upgrade_candidate.incr_refresh_walg_credentials
+    upgrade_candidate.incr_configure
+    upgrade_candidate.incr_restart
+
+    hop_wait_upgrade_candidate
+  end
+
+  label def wait_upgrade_candidate
+    nap 5 if upgrade_candidate.restart_set? || upgrade_candidate.strand.label != "wait"
```
byucesoy, line 93: Why do we need to wait for this?

byucesoy (approved):

prog/postgres/postgres_resource_nexus.rb
shikharbhardwaj: I wanted to make sure all callers are updated, and I think version is an important argument that the caller should supply when creating a PG resource, wdyt?

prog/postgres/converge_postgres_resource.rb
shikharbhardwaj: The recycle stage encounters failures as the candidate is waiting to restart. Its not strictly necessary though, as it should eventually succeed but leads to more noise in the logs.

prog/postgres/postgres_resource_nexus.rb
jeremyevans: We don't require the version be specified when creating a postgres resource in the api, so the default needs to be set somewhere, and it makes sense to set the default in a single place in `assemble` as opposed to multiple places that call `assemble`.

shikharbhardwaj:
@byucesoy could you take another pass at this one when you get a chance? I made some minor updates since the last review (fixed the timeline bug and added aws related changes).

PR #3863, opened 2025-09-05
Use herb formatter/linter instead of erb_formatter

jeremyevans:
Herb is a relatively new html+erb library. One of the most useful parts of it for us is a formatter and linter. Currently, the herb formatter and linter have a number of bugs that prevent our switching to it, but I expect the bugs to be fixed. After the bugs are fixed, I think it's worth considering a switch from erb_formatter to herb's formatter and linter.

This gives an example of a switch, though it required manually fixing some things the formatter broke,  and the linter currently results in 4 false positives. I've filed upstream bug reports for all issues.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -108,3 +108,6 @@ RSpec/IndexedLet:
 
 RSpec/NestedGroups:
   Enabled: false
+
+Layout/AssignmentIndentation:
+  Enabled: false
--- a/Gemfile
+++ b/Gemfile
@@ -70,7 +70,6 @@ end
 
 group :lint do
   gem "brakeman"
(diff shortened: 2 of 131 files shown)


enescakir:
rubocop-erb just replaced better_html with herb
https://github.com/r7kamura/rubocop-erb/pull/59

jeremyevans:
> rubocop-erb just replaced better_html with herb [r7kamura/rubocop-erb#59](https://github.com/r7kamura/rubocop-erb/pull/59)

That's a good change, but we'll still need something like this to replace erb_formatter I think (assuming we want functionality similar to erb_formatter). I'm going to let this sit for now until the herb bugs that affect Ubicloud are resolved (3 are currently unresolved).

jeremyevans:
Closing this, will submit a new PR that uses the herb formatter (the herb linter needs more work before we can use it).

PR #3865, opened 2025-09-06
Make number of vring_workers configurable.

pykello:


migrate/20250905_add_storage_volume_vring_workers.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:vm_storage_volume) do
+      add_column :vring_workers, Integer, null: true
```
jeremyevans: `null: true` is the default, so you can do (at your option):

```suggestion
      add_column :vring_workers, Integer
```

spec/scheduling/allocator_spec.rb
```diff
@@ -803,18 +803,20 @@ def create_req(vm, storage_volumes, target_host_utilization: 0.55, distinct_stor
       vmh = VmHost.first
       vhost_backend = VhostBlockBackend.create(vm_host_id: vmh.id, version: "v0.1-5", allocation_weight: 100)
       vm = create_vm
-      described_class.allocate(vm, [{"size_gib" => 5, "use_bdev_ubi" => false, "skip_sync" => false, "encrypted" => true, "boot" => false}])
+      described_class.allocate(vm, [{"size_gib" => 5, "use_bdev_ubi" => false, "skip_sync" => false, "encrypted" => true, "boot" => false, "vring_workers" => 3}])
       expect(vm.vm_storage_volumes.first.vhost_block_backend_id).to eq(vhost_backend.id)
       expect(vm.vm_storage_volumes.first.spdk_installation_id).to be_nil
+      expect(vm.vm_storage_volumes.first.vring_workers).to eq(3)
```
jeremyevans: Maybe DRY this part up?

```suggestion
      volume = vm.vm_storage_volumes.first
      expect(volume.vhost_block_backend_id).to eq(vhost_backend.id)
      expect(volume.spdk_installation_id).to be_nil
      expect(volume.vring_workers).to eq(3)
```

migrate/20250908_backfill_vring_workers_and_add_constraint.rb
```diff
@@ -0,0 +1,25 @@
...
+      UPDATE vm_storage_volume vsv
+      SET vring_workers = GREATEST(1, vm.vcpus / 2)
+      FROM vm
+      WHERE vsv.vm_id = vm.id
+        AND vsv.vring_workers IS NULL
+        AND vsv.vhost_block_backend_id IS NOT NULL;
+    SQL
+
+    alter_table(:vm_storage_volume) do
+      add_constraint(:vring_workers_positive_if_ubiblk,
```
jeremyevans, line 15: May want a similar to constraint `vhost_block_backend_id IS NOT NULL OR vring_workers IS NULL`? I assume that `vring_workers` should only be set if `vhost_block_backend_id` is set.

PR #3875, opened 2025-09-09
Csi improvements

mohi-kalantari:
**Increase CSI disk limit to 10GB**

**Make CSI PublishVolume idempotent**

Check if the target_path is already mounted and if not, proceed
to mount the path.

**Update CSI dockerfile to always install mkfs binaries**

**Do not remove all annotations when trimming PVC during pvc recreation**

Use a better approach to only add/remove the desired annotations
instead of replacing it all so we are sure we are not causing problem
for helm using annotation on resources.

**Remove left over loop device in NodeUnstage**

Previously we would leave the loop device untouched after deleting
a volume. As the CSI spec describes, any action taken in NodeStage
must be undone in NodeUnstage.

**Detect the volume device filesystem using blkid**

Prior to this commit, the logic to detect whether we need to create
filesystem for a device was flawed and it relied on the existence
of loop device or if it was copied across nodes or not.

Now with this commit, we will use blkid which is a reliable tool
to detect the filesystem and act according to it

Also some validation checks are added to make sure we are working
with the filesystems we support.


kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -19,6 +19,7 @@ class NodeService < Node::Service
       VOLUME_BASE_PATH = "/var/lib/ubicsi"
       OLD_PV_NAME_ANNOTATION_KEY = "csi.ubicloud.com/old-pv-name"
       OLD_PVC_OBJECT_ANNOTATION_KEY = "csi.ubicloud.com/old-pvc-object"
+      ACCEPTABLE_FS = ["ext4", "xfs"]
```
jeremyevans: Constants should generally be frozen:

```suggestion
      ACCEPTABLE_FS = ["ext4", "xfs"].freeze
```

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -303,7 +331,9 @@ def recreate_pvc(req_id, client, pv)
       end
 
       def trim_pvc(pvc, pv_name)
-        pvc["metadata"]["annotations"] = {OLD_PV_NAME_ANNOTATION_KEY => pv_name}
+        pvc["metadata"]["annotations"] ||= {}
+        pvc["metadata"]["annotations"].delete(OLD_PVC_OBJECT_ANNOTATION_KEY)
```
jeremyevans, line 302: No need to delete before setting new value:

```suggestion
```

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -79,6 +80,15 @@ def find_loop_device(backing_file, req_id:)
         nil
       end
 
+      def remove_loop_device(backing_file, req_id:)
+        loop_device = find_loop_device(backing_file, req_id:)
+        return unless loop_device
+        output, ok = run_cmd("losetup", "-d", loop_device, req_id:)
+        if !ok
```
jeremyevans: Optional but more idiomatic:

```suggestion
        unless ok
```

Similar situation in other places in this PR.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -158,18 +167,26 @@ def perform_node_stage_volume(req_id, pvc, req, _call)
...
-            output, ok = run_cmd("mkfs.#{fs_type}", loop_device, req_id:)
-            unless ok
-              log_with_id(req_id, "gRPC error in node_stage_volume: failed to format device: #{output}")
-              raise GRPC::Internal, "Failed to format device #{loop_device} with #{fs_type}: #{output}"
+          if req.volume_capability.mount
+            current_fs_type = find_file_system(loop_device, req_id:)
+            if !current_fs_type.empty? && !ACCEPTABLE_FS.include?(current_fs_type)
+              error_message = "Unacceptable file system type for #{loop_device}: #{current_fs_type}"
+              log_with_id(req_id, error_message)
+              raise error_message
```
jeremyevans: Consider adding a method that calls these two methods, and using that to DRY things up, since this pattern shows up in multiple places.

kubernetes/csi/lib/ubi_csi/node_service.rb
mohi-kalantari: They are different keys so I'm not setting a new value

aamederen (approved):

PR #3880, opened 2025-09-10
Refresh DNS record when renaming postgres database

jeremyevans:
Also, add note in UI that renaming the database changes the connection info and requires updating applications.


byucesoy (approved):

jeremyevans:
I added the refresh_certificates semaphore, and used that. I also included some additional changes:

* Don't rename objects if the name did not change. This wasn't an significant issue before, but we don't want to refresh certificates for postgres if the name didn't change.
* Move postgres rename form to the "Danger Zone" in the web UI:

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_postgres_t_settings

jeremyevans:
@byucesoy Can you give this another review to make sure I got the SSL cert changes correct (ignoring the last commit, which is similar but for load balancers)?

byucesoy (approved):
New commits also look good.

PR #3883, opened 2025-09-11
Small Touches to Host Version and Rhizome Bootstrap

aamederen:
While investigating the today's CI issue as the on-call, I made small touches here & there to improve investigation for future.

----

> [!IMPORTANT]
> Update default Ubuntu version to 24.04 in `hetzner_apis.rb` and related tests, and verify `bundle` installation in `bootstrap_rhizome.rb`.
> 
>   - **Behavior**:
>     - Update default distribution to `Ubuntu 24.04 LTS base` in `reimage()` in `hetzner_apis.rb`.
>     - Remove `dist` parameter from `reset()` in `hetzner_apis.rb`.
>   - **Bootstrap**:
>     - Add `sudo which bundle` after installing `ruby-bundler` in `bootstrap_rhizome.rb` to verify installation.
>   - **Tests**:
>     - Update expected OS version to `24.04` in `learn_os_spec.rb` and `host_nexus_spec.rb`.
>     - Update shell command expectations in `bootstrap_rhizome_spec.rb` to match `bootstrap_rhizome.rb` changes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7694c4495c0a4d0874da152930b0f3126c921c6f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/bootstrap_rhizome.rb
+++ b/prog/bootstrap_rhizome.rb
@@ -52,7 +52,7 @@ LOGIND
 set -ueo pipefail
 sudo apt-get update
 sudo apt-get -y install ruby-bundler
-sudo which bundler
+sudo which bundle
 sudo userdel -rf rhizome || true
 sudo adduser --disabled-password --gecos '' rhizome
 echo 'rhizome ALL=(ALL) NOPASSWD:ALL' | sudo tee /etc/sudoers.d/98-rhizome
--- a/spec/prog/bootstrap_rhizome_spec.rb
+++ b/spec/prog/bootstrap_rhizome_spec.rb
@@ -41,7 +41,7 @@ RSpec.describe Prog::BootstrapRhizome do
 set -ueo pipefail
 sudo apt-get update
 sudo apt-get -y install ruby-bundler
-sudo which bundler
+sudo which bundle
 sudo userdel -rf rhizome || true
 sudo adduser --disabled-password --gecos '' rhizome
 echo 'rhizome ALL=(ALL) NOPASSWD:ALL' | sudo tee /etc/sudoers.d/98-rhizome


spec/prog/bootstrap_rhizome_spec.rb
```diff
@@ -39,7 +39,9 @@
       expect(br).to receive(:sshable).and_return(sshable).at_least(:once)
       expect(Util).to receive(:rootish_ssh).with "hostname", "root", ["test private key"], <<'FIXTURE'
 set -ueo pipefail
-sudo apt update && sudo apt-get -y install ruby-bundler
+sudo apt-get update
+sudo apt-get -y install ruby-bundler
+sudo which bundler
```
jeremyevans: Probably want to use `bundle` instead of `bundler` here. Currently, the `bundler` gem does include both programs, but the recommended one is `bundle`, I think `bundler` only exists as a backwards compatible alias.

```suggestion
sudo which bundle
```

prog/bootstrap_rhizome.rb
```diff
@@ -50,7 +50,9 @@ def user
     key_data = sshable.keys.map(&:private_key)
     Util.rootish_ssh(sshable.host, user, key_data, <<SH)
 set -ueo pipefail
-sudo apt update && sudo apt-get -y install ruby-bundler
+sudo apt-get update
+sudo apt-get -y install ruby-bundler
+sudo which bundler
```
jeremyevans:
```suggestion
sudo which bundle
```

PR #3884, opened 2025-09-11
Add no-op version of SSH session locking

fdr:
22f3aa5c1dcc7a219476af066dc849faf58d5763 "Terminate SSH sessions promptly and their processes", by yours truly, introduced many problems. It's useful to be lazy about returning SSH `keepalive@openssh.com` packets (note: not the same as TCP keepalive) while maintaining a session, and that was broken by that patch.

At the same time, lingering sessions without any interlock theory are bad, and I don't want to go back to that.

Back when writing 22f3aa5c1, I thought that tight timing there could be used as a form of lease-style mutual exclusion, but in retrospect, it would either be *too* tightly timed or our retry interval would have to become much longer.

To resolve this, I intend to add mutual exclusion recorded on the server, costing some overhead with every established connection. My experiments with `hyperfine` suggest 3ms of overhead or less. See the bottom of this message for a benchmark script and its output on my laptop.

In this patch, I am taking the first step to introduce opt-in mutual exclusion to sessions via a constant/global called `SSH_SESSION_LOCK_NAME`.

By default, this does nothing until `SSH_SESSION_LOCK_NAME` is defined. This can be set differently for different processes; e.g., `respirate` may bind this symbol but `monitor` may bind it differently, or not at all.

Even with this symbol set, this implementation is nearly a no-op, to test overhead and measure status quo cases of accidental(?) session concurrency where crashing upon mutual exclusion could be disruptive. Instead, it only logs a message when there would have been session lock contention.

Eventually, once we have a handle on everything, the `Clog.emit` can be converted to raise.

A way to test this against a `Sshable`:

    sa = Sshable.first

    # Demonstrate what happens without SSH_SESSION_LOCK_NAME set.
    p ['precondition: no locking', sa.cmd('pgrep -af session || true')]
    sa.invalidate_cache_entry
    p ['postcondition: no locking', sa.cmd('pgrep -af session || true')]
    sa.invalidate_cache_entry

    SSH_SESSION_LOCK_NAME = "test-lock"

    # Should display the session flock holding process, which is
    # automatically started when the session begins.
    p ['precondition: locking', sa.cmd('pgrep -af session')]

    # Invalidating the Sshable cache entry creates a second concurrent
    # session upon the next cmd.
    sa.invalidate_cache_entry

    # This should display the contention logging as a side effect and
    # show the same session flock process pid.
    p ['postcondition: locking', sa.cmd('pgrep -af session')]

At the last part, where you expect to see a message as a side effect, it should look like this:

    ....
    ["precondition: locking", "21256 session-lock-test-lock infinity\n"]
    ....
    {"contended_session_lock":{"exit_code":124,"session_fail_msg":"session lock conflict for test-lock"},"message":"session lock failure","time":"2025-09-08 15:31:37 -0700"}
    ....
    ["postcondition: locking", "21256 session-lock-test-lock infinity\n"]

Note that the pid (21256) doesn't change, and the
`contended_session_lock` key.

You can use `pgrep` to debug the session locking interactively. You might also find the distinctive file descriptor number useful, e.g., `ls /proc/*/fd/999`.

Here's a benchmarking program using `hyperfine` with a closely related version to measure overhead. You can put it into a file (e.g., `bench.bash`) and execute it. It has modifications of exit codes to help ensure the benchmark is doing something meaningful.

```sh
set -uex

cat > prepare.sh << 'EOF'
pkill -f sessionlockasdf 2>/dev/null || true
flock /dev/shm/session_lockfile true
EOF

cat > locking.sh << 'EOF'
exec 999>/dev/shm/session_lockfile || exit 1
flock -xn 999 || { echo "Another session active."; exit 1; }
exec -a sessionlockasdf sleep infinity </dev/null >/dev/null 2>&1 &
disown
EOF

cat > noop.sh << 'EOF'
exit 0
EOF

chmod +x prepare.sh locking.sh noop.sh

hyperfine -N \
    --prepare './prepare.sh' \
    --command-name "Session Lock" './locking.sh' \
    --command-name "Noop" './noop.sh' \
    --cleanup './prepare.sh'

rm -f prepare.sh locking.sh noop.sh
rm -f /dev/shm/session_lockfile 2>/dev/null || true
```

On my laptop, it has output like this:

```
+ cat
+ cat
+ cat
+ chmod +x prepare.sh locking.sh noop.sh
+ hyperfine -N --prepare ./prepare.sh --command-name 'Session Lock' ./locking.sh --command-name Noop ./noop.sh --cleanup ./prepare.sh
Benchmark 1: Session Lock
  Time (mean ± σ):       2.0 ms ±   0.2 ms    [User: 0.7 ms, System: 1.2 ms]
  Range (min … max):     1.5 ms …   2.3 ms    110 runs

Benchmark 2: Noop
  Time (mean ± σ):     928.3 µs ± 126.4 µs    [User: 386.7 µs, System: 445.3 µs]
  Range (min … max):   579.7 µs … 1221.6 µs    154 runs

Summary
  Noop ran
    2.11 ± 0.34 times faster than Session Lock
+ rm -f prepare.sh locking.sh noop.sh
+ rm -f /dev/shm/session_lockfile
```

----

> [!IMPORTANT]
> Introduces a no-op SSH session locking mechanism in `Sshable` to log conflicts and test overhead, with corresponding tests added.
> 
>   - **Behavior**:
>     - Adds `maybe_ssh_session_lock_name` to `Sshable` to check for `SSH_SESSION_LOCK_NAME`.
>     - Modifies `connect` in `Sshable` to attempt session locking if `SSH_SESSION_LOCK_NAME` is set, logging conflicts instead of raising errors.
>   - **Error Handling**:
>     - Logs session lock conflicts and file descriptor errors in `connect`.
>     - Handles specific exit codes (92, 124) with descriptive messages.
>   - **Tests**:
>     - Adds tests for `maybe_ssh_session_lock_name` and session locking behavior in `sshable_spec.rb`.
>     - Verifies error handling for session lock conflicts and file descriptor issues.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 908e471fcef8831d223a74e5a2b862089e5f411e. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -135,6 +135,10 @@ class Sshable < Sequel::Model
                      verify_host_key: :accept_new, keys: [], key_data: [], use_agent: false,
                      keepalive: true, keepalive_interval: 3, keepalive_maxcount: 5}.freeze
 
+  def maybe_ssh_session_lock_name
+    SSH_SESSION_LOCK_NAME if defined?(SSH_SESSION_LOCK_NAME)
+  end
+
   def connect
     Thread.current[:clover_ssh_cache] ||= {}
 
@@ -148,6 +152,33 @@ class Sshable < Sequel::Model
     sess = start_fresh_session
     @connect_duration = Time.now - start
     Thread.current[:clover_ssh_cache][[host, unix_user]] = sess
+
+    if (lock_name = maybe_ssh_session_lock_name&.shellescape)
+      lock_contents = <<LOCK
+exec 999>/dev/shm/session-lock-#{lock_name} || exit 92
+flock -xn 999 || { echo "Another session active: " #{lock_name}; exit 124; }
+exec -a session-lock-#{lock_name} sleep infinity </dev/null >/dev/null 2>&1 &
+disown
+LOCK
+
+      begin
+        cmd(lock_contents, log: false)
+      rescue SshError => ex
+        session_fail_msg = case (exit_code = ex.exit_code)
+        when 92
+          "could not create session lock file for #{lock_name}"
+        when 124
+          "session lock conflict for #{lock_name}"
+        else
+          "unknown SshError"
+        end
+
+        Clog.emit("session lock failure") do
+          {contended_session_lock: {exit_code:, session_fail_msg:}}
+        end
+      end
+    end
+
     sess
   end
 
--- a/spec/model/sshable_spec.rb
+++ b/spec/model/sshable_spec.rb
@@ -23,6 +23,76 @@ RSpec.describe Sshable do
     expect(sa.raw_private_key_1).to eq(key)
   end
 
+  describe "#maybe_ssh_session_lock_name" do
+    it "does not yield if SSH_SESSION_LOCK_NAME is not defined" do
+      expect(sa.maybe_ssh_session_lock_name).to be_nil
+    end
+
+    unless ENV["CLOVER_FREEZE"]
+      it "yields if SSH_SESSION_LOCK_NAME is defined" do
+        stub_const("SSH_SESSION_LOCK_NAME", "testlockname")
+        expect(sa.maybe_ssh_session_lock_name).to eq("testlockname")
+      end
+    end
+  end
+
+  describe "session locking" do
+    lock_script = <<LOCK
+exec 999>/dev/shm/session-lock-testlockname || exit 92
+flock -xn 999 || { echo "Another session active: " testlockname; exit 124; }
+exec -a session-lock-testlockname sleep infinity </dev/null >/dev/null 2>&1 &
+disown
+LOCK
+
+    it "interlocks" do
+      portable_pkill = lambda { system(%q(ps -eo pid,args | awk '$2=="session-lock-testlockname"{print $1}' | xargs -I {} sh -c 'test -n "{}" && kill {}')) }
+      portable_pkill.call
+      q_lock_script = lock_script.shellescape
+      expect([`bash -c #{q_lock_script}`, $?.exitstatus]).to eq(["", 0])
+      expect([`bash -c #{q_lock_script}`, $?.exitstatus]).to eq(["Another session active:  testlockname\n", 124])
+      expect(portable_pkill.call).to be true
+    end
+
+    describe "exit code handling" do
+      before do
+        expect(sa).to receive(:maybe_ssh_session_lock_name).and_return("testlockname")
+        sa.invalidate_cache_entry
+        expect(Net::SSH).to receive(:start) do
+          instance_double(Net::SSH::Connection::Session, close: nil)
+        end
+      end
+
+      it "runs the session lock script if SSH_SESSION_LOCK_NAME is set" do
+        expect(sa).to receive(:cmd).with(lock_script, log: false)
+        sa.connect
+      end
+
+      it "reports a failure to obtain a file descriptor with an obscure exit code" do
+        expect(sa).to receive(:cmd).with(lock_script, log: false).and_raise(Sshable::SshError.new(lock_script, "", "", 92, nil))
+        expect(Clog).to receive(:emit).with("session lock failure").and_wrap_original do |m, a, &b|
+          expect(b.call.dig(:contended_session_lock, :session_fail_msg)).to eq("could not create session lock file for testlockname")
+        end
+        sa.connect
+      end
+
+      it "reports lock conflicts when an obscure exit code is raised" do
+        expect(sa).to receive(:cmd).with(lock_script, log: false).and_raise(Sshable::SshError.new(lock_script, "", "", 124, nil))
+        expect(Clog).to receive(:emit).with("session lock failure").and_wrap_original do |m, a, &b|
+          expect(b.call.dig(:contended_session_lock, :session_fail_msg)).to eq("session lock conflict for testlockname")
+        end
+        sa.connect
+      end
+
+      it "has a generic message for unrecognized errors" do
+        expect(sa).to receive(:cmd).with(lock_script, log: false).and_raise(Sshable::SshError.new(lock_script, "", "", 1, nil))
+        expect(Clog).to receive(:emit).with("session lock failure").and_wrap_original do |m, a, &b|
+          expect(b.call.dig(:contended_session_lock, :session_fail_msg)).to eq("unknown SshError")
+        end
+        sa.connect
+      end
+    end
+  end
+
   describe "caching" do
     # The cache is thread local, so re-set the thread state by boxing
     # each test in a new thread.


model/sshable.rb
```diff
@@ -148,6 +148,33 @@ def connect
...
     @connect_duration = Time.now - start
     Thread.current[:clover_ssh_cache][[host, unix_user]] = sess
+
+    if defined?(SSH_SESSION_LOCK_NAME)
+      lock_contents = <<LOCK
+exec 999>#{("/dev/shm/" + SSH_SESSION_LOCK_NAME).shellescape} || exit 92
+flock -xn 999 || { echo "Another session active: " #{SSH_SESSION_LOCK_NAME.shellescape}; exit 124; }
+exec -a #{("session-lock-" + SSH_SESSION_LOCK_NAME).shellescape} sleep infinity </dev/null >/dev/null 2>&1 &
+disown
+LOCK
```
jeremyevans, line 162: This is not portable shell code, but we can assume that hosts will have bash as the default shell, correct?

model/sshable.rb
```diff
@@ -148,6 +148,33 @@ def connect
     sess = start_fresh_session
     @connect_duration = Time.now - start
     Thread.current[:clover_ssh_cache][[host, unix_user]] = sess
+
+    if defined?(SSH_SESSION_LOCK_NAME)
+      lock_contents = <<LOCK
+exec 999>#{("/dev/shm/" + SSH_SESSION_LOCK_NAME).shellescape} || exit 92
+flock -xn 999 || { echo "Another session active: " #{SSH_SESSION_LOCK_NAME.shellescape}; exit 124; }
+exec -a #{("session-lock-" + SSH_SESSION_LOCK_NAME).shellescape} sleep infinity </dev/null >/dev/null 2>&1 &
```
jeremyevans: May be simpler to shell escape only once, as nothing you are concatenating to also needs shell escaping:

```suggestion
      lock_name = SSH_SESSION_LOCK_NAME.shellescape
      lock_contents = <<LOCK
exec 999>/dev/shm/#{lock_name} || exit 92
flock -xn 999 || { echo "Another session active: " #{lock_name}; exit 124; }
exec -a session-lock-#{lock_name} sleep infinity </dev/null >/dev/null 2>&1 &
```

spec/model/sshable_spec.rb
```diff
@@ -23,6 +23,57 @@
     expect(sa.raw_private_key_1).to eq(key)
   end
 
+  describe "session locking" do
+    before do
+      if ENV["CLOVER_FREEZE"]
+        skip("can only define SSH_SESSION_LOCK_NAME when tests are not frozen")
+      else
+        stub_const("SSH_SESSION_LOCK_NAME", "testlockname")
+      end
```
jeremyevans: We avoid using `skip` in the specs:

```suggestion
unless ENV["CLOVER_FREEZE"] == "1"
  describe "session locking" do
    before do
      stub_const("SSH_SESSION_LOCK_NAME", "testlockname")
```

Alternatively, add a method that returns the value of the constant, use that method in the code, and override that method in the spec, so it can be tested in frozen mode.

model/sshable.rb
fdr: I think to get startup time down, it's basically obligate. If we had to become more portable, we could omit legible process title.

fdr:
> Specs are all mocked, so what you actually care about (session interlock) is not tested by them. To get actual testing of this, there are a couple options:

It tests exit code response only. But "actual" testing at the standard of loginctl behavior will not come easily.

> 
>     * Use it by default in the E2E tests (raising instead of emitting), and see if E2E passes. Ideally, include a test prog that violates the interlock and tests that an expected failure is raised.

That would work, although it'd work more easily once it is converted from a clog to an error I might think.

>     * Implement tests that execute the lock script and check for expected behavior (maybe only possible on Linux)?

I added a test that interlocks by running the unix programs locally. I am not sure if it will run on macintosh.

spec/model/sshable_spec.rb
fdr: I was hoping so much torture of the code under test for something like that, although I suppose we could.

fdr: I am giving it a shot

fdr: I used a block to do it, that way it cleans up some of the internal references to the global const as well.

model/sshable.rb
```diff
@@ -148,6 +154,34 @@ def connect
     sess = start_fresh_session
     @connect_duration = Time.now - start
     Thread.current[:clover_ssh_cache][[host, unix_user]] = sess
+
+    maybe_lock_session do
+      lock_name = it.shellescape
+      lock_contents = <<LOCK
+exec 999>/dev/shm/#{lock_name} || exit 92
+flock -xn 999 || { echo "Another session active: " #{lock_name}; exit 124; }
+exec -a session-lock-#{lock_name} sleep infinity </dev/null >/dev/null 2>&1 &
```
fdr, line 160: fun side fact: I originally was thinking to make testing easier that I'd echo $! (in bash) for the pid here, but it cost about 100us or 5% of 2ms, and thought: eh, how about a one-liner portable(?) `pkill -f` instead in the test, to find the process by its argv

fdr:
@furkansahin @enescakir can one of you macintosh people give the test a try :pray: 

it has some "theoretically portable" elements

fdr:
@jeremyevans maybe give it another look. I added an interlock-exercising test (tbd if macintosh can do it) and out-lined the global symbol dependency to a method that's a block to also encapsulate how the code inside references the lock name.

All that remains is an E2E version, which maybe I'll prep along with a patch that switches Clog for raise, expecting an exception when manipulating the ssh connection should be pretty easy (per my commit message, all that should be necessary is invalidating the cache and logging in again and see how long it takes to work, keyed to the ClientAliveInterval)

model/sshable.rb
```diff
@@ -148,6 +154,34 @@ def connect
...
+disown
+LOCK
+
+      begin
+        cmd(lock_contents)
+      rescue SshError => ex
+        session_fail_msg = case (ec = ex.exit_code)
```
jeremyevans:
```suggestion
        session_fail_msg = case (exit_code = ex.exit_code)
```

model/sshable.rb
```diff
@@ -148,6 +154,34 @@ def connect
...
+        when 92
+          "could not create session lock file for #{lock_name}"
+        when 124
+          "session lock conflict for #{lock_name}"
+        else
+          "unknown SshError"
+        end
+
+        Clog.emit("session lock failure") do
+          {contended_session_lock: {exit_code: ec, session_fail_msg:}}
```
jeremyevans:
```suggestion
          {contended_session_lock: {exit_code:, session_fail_msg:}}
```

fdr:
I simplified away the use of a block for encapsulation, instead using a method and using the `if (var = method)` binding style. I ended up at a block through some path dependent edits, but upon biking home I realized it seemed a little silly.

fdr:
> Looks good to me. Should be a no-op, so safe to deploy assuming it passes on macOS. However, I wouldn't recommend enabling this (defining SSH_SESSION_LOCK_NAME) in production until after a successful E2E test.
> 
> Assuming this is succesful, should we revert [22f3aa5](https://github.com/ubicloud/ubicloud/commit/22f3aa5c1dcc7a219476af066dc849faf58d5763)?

No: once you interlock with this new thing and you have have a lingering session, the program will be stuck for a long time. What we can do is relax the timing of the session destruction.

spec/model/sshable_spec.rb
```diff
@@ -23,6 +23,76 @@
     expect(sa.raw_private_key_1).to eq(key)
   end
 
+  describe "#maybe_lock_session?" do
```
jeremyevans:
```suggestion
  describe "#maybe_ssh_session_lock_name" do
```

enescakir:
> @furkansahin @enescakir can one of you macintosh people give the test a try 🙏
> 
> it has some "theoretically portable" elements

@fdr It works in my dev environment. We might consider disabling its clog log, since I’m not sure it’s useful.

```
[9] clover-development(main)> vmh.sshable.cmd("ls -lah /var/storage/images/");

{"ssh":{"start":"2025-09-15 11:08:59 +0300","finish":"2025-09-15 11:08:59 +0300","cmd":"exec 999>/dev/shm/session-lock-pry || exit 92\nflock -xn 999 || { echo \"Another session active: \" pry; exit 124; }\nexec -a session-lock-pry sleep infinity </dev/null >/dev/null 2>&1 &\ndisown\n","exit_code":0,"exit_signal":null,"ubid":"vhjfs697wr9y2knw0pffwehj4z","duration":0.316004,"channel_duration":0.26782,"connect_duration":0.498913},"message":"ssh cmd execution","time":"2025-09-15 11:08:59 +0300"}

total 154G
drwxr-xr-x 2 root root 4.0K Aug 29 13:10 .
drwxr-xr-x 5 root root 4.0K Aug 29 15:10 ..
-rw-r--r-- 1 root root  75G Aug 27 08:15 github-ubuntu-2204-20250821.1.0.raw
-rw-r--r-- 1 root root  75G Aug 29 13:10 github-ubuntu-2404-20250821.1.0.raw
-rw-r--r-- 1 root root 2.2G Aug 25 11:36 ubuntu-jammy-20250508.raw
-rw-r--r-- 1 root root 3.5G Apr  7 15:44 ubuntu-noble-20240702.raw

{"ssh":{"start":"2025-09-15 11:08:58 +0300","finish":"2025-09-15 11:08:59 +0300","cmd":"ls -lah /var/storage/images/","exit_code":0,"exit_signal":null,"ubid":"vhjfs697wr9y2knw0pffwehj4z","duration":0.987482,"channel_duration":0.939773,"connect_duration":0.498913},"message":"ssh cmd execution","time":"2025-09-15 11:08:59 +0300"}
```

fdr:
> > @furkansahin @enescakir can one of you macintosh people give the test a try 🙏
> > it has some "theoretically portable" elements
> 
> @fdr It works in my dev environment. We might consider disabling its clog log, since I’m not sure it’s useful.
> 
> ```
> [9] clover-development(main)> vmh.sshable.cmd("ls -lah /var/storage/images/");
> 
> {"ssh":{"start":"2025-09-15 11:08:59 +0300","finish":"2025-09-15 11:08:59 +0300","cmd":"exec 999>/dev/shm/session-lock-pry || exit 92\nflock -xn 999 || { echo \"Another session active: \" pry; exit 124; }\nexec -a session-lock-pry sleep infinity </dev/null >/dev/null 2>&1 &\ndisown\n","exit_code":0,"exit_signal":null,"ubid":"vhjfs697wr9y2knw0pffwehj4z","duration":0.316004,"channel_duration":0.26782,"connect_duration":0.498913},"message":"ssh cmd execution","time":"2025-09-15 11:08:59 +0300"}
> 
> total 154G
> drwxr-xr-x 2 root root 4.0K Aug 29 13:10 .
> drwxr-xr-x 5 root root 4.0K Aug 29 15:10 ..
> -rw-r--r-- 1 root root  75G Aug 27 08:15 github-ubuntu-2204-20250821.1.0.raw
> -rw-r--r-- 1 root root  75G Aug 29 13:10 github-ubuntu-2404-20250821.1.0.raw
> -rw-r--r-- 1 root root 2.2G Aug 25 11:36 ubuntu-jammy-20250508.raw
> -rw-r--r-- 1 root root 3.5G Apr  7 15:44 ubuntu-noble-20240702.raw
> 
> {"ssh":{"start":"2025-09-15 11:08:58 +0300","finish":"2025-09-15 11:08:59 +0300","cmd":"ls -lah /var/storage/images/","exit_code":0,"exit_signal":null,"ubid":"vhjfs697wr9y2knw0pffwehj4z","duration":0.987482,"channel_duration":0.939773,"connect_duration":0.498913},"message":"ssh cmd execution","time":"2025-09-15 11:08:59 +0300"}
> ```

good idea, applied

PR #3885, opened 2025-09-11
Bubble Up Exceptions if Check Pulse Thread fails

aamederen:
We used to sit silent if the monitor pulse function doesn't fail properly. We can be more sensitive to exceptions in the pulse threads so that if at any point the thread raises exception, we log a message and incr_checkup on the resource.

----

> [!IMPORTANT]
> Improve exception handling and logging in pulse monitoring threads, ensuring exceptions are raised and resources are checked up if possible.
> 
>   - **Behavior**:
>     - In `monitor_resource_type.rb`, exceptions in monitoring threads are now logged, and `incr_checkup` is called if available.
>     - In `metrics_target_resource.rb` and `monitorable_resource.rb`, exceptions other than `Sequel::NoExistingObject` are now raised.
>   - **Logging**:
>     - Added logging for failed monitoring jobs in `monitor_resource_type.rb`.
>     - Added logging for failed pulse checks in `monitorable_resource.rb`.
>   - **Tests**:
>     - Updated tests in `metrics_target_resource_spec.rb` to expect exceptions to be raised.
>     - Added tests in `monitor_resource_type_spec.rb` for exception handling and logging.
>     - Updated `monitorable_resource_spec.rb` to test new exception handling behavior.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 790f0932fefc05cadff66761c07cc903ce7d3eed. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/metrics_target_resource.rb
+++ b/lib/metrics_target_resource.rb
@@ -20,12 +20,10 @@ class MetricsTargetResource
     return if @session && @last_export_success
 
     @session = @resource.reload.init_metrics_export_session
-  rescue => ex
-    if ex.is_a?(Sequel::NoExistingObject)
-      Clog.emit("Resource is deleted.") { {resource_deleted: {ubid: @resource.ubid}} }
-      @session = nil
-      @deleted = true
-    end
+  rescue Sequel::NoExistingObject
+    Clog.emit("Resource is deleted.") { {resource_deleted: {ubid: @resource.ubid}} }
+    @session = nil
+    @deleted = true
   end
 
   def export_metrics
--- a/spec/lib/metrics_target_resource_spec.rb
+++ b/spec/lib/metrics_target_resource_spec.rb
@@ -65,9 +65,9 @@ RSpec.describe MetricsTargetResource do
       expect(resource.instance_variable_get(:@session)).to be_nil
     end
 
-    it "ignores other exceptions" do
+    it "raises the exception if it is not Sequel::NoExistingObject" do
       expect(postgres_server).to receive(:reload).and_raise(StandardError)
-      expect { resource.open_resource_session }.not_to raise_error
+      expect { resource.open_resource_session }.to raise_error(StandardError)
     end
   end
 


lib/monitorable_resource.rb
```diff
@@ -22,6 +22,9 @@ def open_resource_session
       Clog.emit("Resource is deleted.") { {resource_deleted: {ubid: @resource.ubid}} }
       @session = nil
       @deleted = true
+    else
+      Clog.emit("Opening resource session has failed.") { {open_resource_session_failure: {resource: @resource, exception: Util.exception_to_hash(ex)}} }
+      raise ex
     end
   end
```
jeremyevans: This logs the same information twice if opening a resource section fails. I think that's a bad idea. Only issue with that is the emit message is different, so harder to see the issue is specifically in `open_resource_session`. However, you can determine that by looking at the backtrace.

I would change this to only rescue `Sequel::NoExistingObject`:

```ruby
rescue Sequel::NoExistingObject
  Clog.emit("Resource is deleted.") { {resource_deleted: {ubid: @resource.ubid}} }
  @session = nil
  @deleted = true
end
```

lib/monitor_resource_type.rb
```diff
@@ -31,22 +31,27 @@ def self.create(klass, stuck_pulse_info, num_threads, types)
...
-          finish_queue.push(r)
+          begin
+            # We keep track of started at information to check for stuck pulses.
+            r.monitor_job_started_at = Time.now
+            r.open_resource_session
+
+            # Yield so that monitored resources and metric export resources can be
+            # handled differently.
+            yield r
+          rescue => ex
```
jeremyevans, line 42: I think this is only hit by the `open_resource_session` call. I don't think the yield (`check_pulse` or `export_metrics`) can raise, as those methods appear to rescue all errors. I don't have a problem with including it here, but it seems the odds of this being hit other than by the `open_resource_session` call are very low. If you want the `incr_checkup` called if there are issues in `check_pulse`/`export_metrics`, I think you need to move it to those methods, or change those methods to also not raise.

aamederen: That's a good point but I think this can be a TODO for later, as our check_pulse needs to be fixed first, esp. for VmHost, to be more consistent and reliable. Currently we see lots of "Pulse checking has failed" messages and raising the exception can lead to unexpected issues, but not sure.

lib/metrics_target_resource.rb
```diff
@@ -40,6 +40,7 @@ def export_metrics
       @last_export_success = false
       close_resource_session
       Clog.emit("Metrics export has failed.") { {metrics_export_failure: {ubid: @resource.ubid, exception: Util.exception_to_hash(ex)}} }
+      # TODO: Consider raising the exception here, and let the caller handle it.
```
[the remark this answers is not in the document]
aamederen: That's the point ellipsis, I think we should fix our prod case first, then start raising.

lib/metrics_target_resource.rb
jeremyevans, line 1: You probably want the same `rescue Sequel::NoExistingObject` change `MetricsTargetResource#open_resource_session` that you made in `MonitorableResource#open_resource_session`.

PR #3887, opened 2025-09-11
Fix PostgresServerNexus#destroy to not violate strand leases

jeremyevans:
Previously, PostgresServerNexus#destroy ran:

```ruby
strand.children.each { it.destroy }
```

This is not safe as there could be an existing lease for the strand children.  Child strands are either:

* BootstrapRhizome
* InstallRhizome (via push from Bootstrap Rhizome)
* Postgres::PostgresServerNexus (via restart from #unavailable)

Fix this by adding a destroy semaphore for all child strands, except the Postgres::PostgresServerNexus ones, as we don't want a child restart prog for the same server to also go into destroy.

Note that if a restart has been triggered via unavailable, and then the destroy semaphore has been incremented, the destroy cannot complete until after the restart commands return successfully.

It would probably be better to have restart in a separate prog, that could exit similarly to how Boostrap/InstallRhizome are setup, so that destruction could proceed even if the restart commands will not run successfully.

----

> [!IMPORTANT]
> Fixes `PostgresServerNexus#destroy` to safely manage child strand destruction using a semaphore, with tests verifying the new behavior.
> 
>   - **Behavior**:
>     - `PostgresServerNexus#destroy` now uses a destroy semaphore to safely manage child strand destruction, excluding `PostgresServerNexus` children.
>     - Introduces `wait_children_destroy` and `destroy_vm_and_pg` labels in `postgres_server_nexus.rb` to handle destruction sequence.
>     - `BootstrapRhizome` and `InstallRhizome` exit early if destroy semaphore is set.
>   - **Tests**:
>     - Added tests in `bootstrap_rhizome_spec.rb`, `install_rhizome_spec.rb`, and `postgres_server_nexus_spec.rb` to verify semaphore behavior and destruction sequence.
>     - Mocking updates in `thawed_mock.rb` to support new semaphore logic.
>   - **Misc**:
>     - Minor refactoring in `postgres_server_nexus.rb` to accommodate new destruction logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 773562ebcbb9f0882b3edc2243e27ea72d0d48b2. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -65,7 +65,7 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
       is_destroying = ["destroy", nil].include?(postgres_server.resource&.strand&.label)
 
       if is_destroying || !postgres_server.taking_over?
-        if strand.label != "destroy"
+        if !%w[destroy wait_children_destroy destroy_vm_and_pg].include?(strand.label)
           hop_destroy
         elsif strand.stack.count > 1
           pop "operation is cancelled due to the destruction of the postgres server"
@@ -608,8 +608,15 @@ SQL
 
   label def destroy
     decr_destroy
+    Semaphore.incr(strand.children_dataset.exclude(prog: "Postgres::PostgresServerNexus").select(:id), "destroy")
+    hop_wait_children_destroy
+  end
 
-    strand.children.each { it.destroy }
+  label def wait_children_destroy
+    reap(:destroy_vm_and_pg, nap: 30)
+  end
+
+  label def destroy_vm_and_pg
     vm.incr_destroy
     postgres_server.destroy
 
--- a/prog/bootstrap_rhizome.rb
+++ b/prog/bootstrap_rhizome.rb
@@ -4,9 +4,16 @@ require_relative "../lib/util"
 
 class Prog::BootstrapRhizome < Prog::Base
   subject_is :sshable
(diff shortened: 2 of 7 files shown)


prog/postgres/postgres_server_nexus.rb
```diff
@@ -65,7 +65,7 @@ def before_run
       is_destroying = ["destroy", nil].include?(postgres_server.resource&.strand&.label)
 
       if is_destroying || !postgres_server.taking_over?
-        if strand.label != "destroy"
+        if !strand.label.include?("destroy")
```
fdr: fishing around in the label is definitely smelly (but pre-exists)

jeremyevans: I could switch to `%w[destroy wait_children_destroy destroy_vm_and_pg].include?(strand.label)` if that is preferred?

byucesoy: Yep, that would be preferred.

byucesoy (approved):

prog/postgres/postgres_server_nexus.rb
jeremyevans: Switched to that approach.

PR #3895, opened 2025-09-15
Introducing github customer runner labels

velioglu:
Until now, users could only use predefined labels for their github workflows. However, some users wanted workflow-level limits for finer control. To address this support for custom labels is introduced.

Custom labels are unique per [installation, label] which allows multiple installations to reuse the same custom label. Each custom label is aliased to a predefined label. In other words, users still run on VM types defined by predefined labels, but can introduce workflow-level limits by using custom labels.

With this commit, only concurrency limits are supported. Additional limits can be introduced later by adding new columns to the github_custom_label model.


migrate/20250915_add_github_custom_label.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:github_custom_label) do
+      column :id, :uuid, primary_key: true, default: Sequel.lit("gen_random_uuid()")
```
enescakir: I don’t think we set a random UUID at the DB level, since it should be a valid UBID. Previously, we generated it at the app level. Recently, @jeremyevans added some DB helpers, but I’m not sure if they’re usable here.

```suggestion
      # UBID.to_base32_n("gc") => 524
      column :id, :uuid, primary_key: true, default: Sequel.lit("gen_random_ubid_uuid(524)")
```

migrate/20250915_add_github_custom_label.rb
```diff
@@ -0,0 +1,17 @@
...
+      foreign_key :installation_id, :github_installation, type: :uuid, null: false
+      column :label, :text, null: false
```
enescakir: `custom_label.label` doesn't sound very natural. We already use the `name` property in `config/github_runner_labels.yml`.

```suggestion
      column :name, :text, null: false
```

routes/webhook/github.rb
```diff
@@ -59,9 +59,10 @@ def handle_workflow_job(data)
       return error("No workflow_job in the payload")
     end
 
-    unless (label = job.fetch("labels").find { Github.runner_labels.key?(it) })
-      return error("Unmatched label")
-    end
+    job_labels = job.fetch("labels")
+    label = job_labels.find { Github.predefined_runner_labels.key?(it) } || GithubCustomLabel.where(installation_id: installation.id, label: job_labels).first&.label
+
+    return error("Unmatched label") unless label
```
enescakir, line 73: I suggest adding an additional `actual_label` column to `GithubRunner` model, while keeping the effective value in `runner.label` and storing the customer’s label in `actual_label`. Here’s why:
	- Customers can change the `alias_for` value of their custom labels over time. Without `actual_label`, we lose visibility into the runner size in the `ArchivedRecord` table. Keeping it allows us to know what the effective label was historically. Also they can delete custom label.
	- Many of our operational commands, scripts, and dashboards already rely on `label`. Changing this would break them.
	- We can also use this new column for our own aliases. Currently, several scripts have special logic for these aliases. For example, we could store `label: ubicloud-standard-2-ubuntu-2204` and `actual_label: ubicloud`. That way, even if we change defaults over time, we won’t lose historical records. It also helps when grouping by label.

model/github_runner.rb
```diff
@@ -16,15 +16,18 @@ class GithubRunner < Sequel::Model
 
   dataset_module do
     def total_active_runner_vcpus
+      predefined_runner_labels = Github.predefined_runner_labels
+      custom_label_aliases = GithubCustomLabel.all.to_h { [it.label, it.alias_for] }
```
enescakir: Since `GithubCustomLabel.label` is unique per installation, should we only fetch the labels that belong to the runner’s installation?

prog/vm/github_runner.rb
```diff
@@ -6,7 +6,7 @@ class Prog::Vm::GithubRunner < Prog::Base
   subject_is :github_runner
 
   def self.assemble(installation, repository_name:, label:, default_branch: nil)
-    unless Github.runner_labels[label]
+    unless Github.predefined_runner_labels[label] || GithubCustomLabel.where(label: label, installation_id: installation.id).any?
```
enescakir: If we keep `label` and `actual_label` as separate fields, we can continue using `label` throughout the codebase without having to check whether it comes from a predefined or a custom label each time.

prog/vm/github_runner.rb
```diff
@@ -200,6 +201,23 @@ def quota_available?
...
+    installation_id = github_runner.installation_id
+    custom_label = GithubCustomLabel.where(label:, installation_id:).first
+    hop_allocate_vm unless custom_label
+
+    unless custom_label.concurrency_limit_available?
+      Clog.emit("hit custom label concurrency limit") { {custom_label_concurrency_limit: {label:, installation_id:, repository_name: github_runner.repository_name}} }
+      nap rand(5..15)
+    end
+
+    custom_label.update(allocated_runner_count: custom_label.allocated_runner_count + 1)
```
enescakir: It's not concurrency safe. `custom_label.allocated_runner_count` is from memory. You should use `Sequel[:allocated_runner_count] + 1`

prog/vm/github_runner.rb
```diff
@@ -200,6 +201,23 @@ def quota_available?
...
+
+    hop_apply_custom_label_quota if GithubCustomLabel.where(label: github_runner.label, installation_id: github_runner.installation_id).any?
+    hop_allocate_vm
+  end
+
+  label def apply_custom_label_quota
+    label = github_runner.label
+    installation_id = github_runner.installation_id
+    custom_label = GithubCustomLabel.where(label:, installation_id:).first
+    hop_allocate_vm unless custom_label
```
enescakir: We shouldn’t encounter this case. I think we can assume that `custom_label` exists here. If it doesn’t, we should allow the runner to fail or we destroy it.

prog/vm/github_runner.rb
```diff
@@ -161,6 +161,7 @@ def before_run
   label def start
     pop "Could not provision a runner for inactive project" unless github_runner.installation.project.active?
     hop_wait_concurrency_limit unless quota_available?
+    hop_apply_custom_label_quota if GithubCustomLabel.where(label: github_runner.label, installation_id: github_runner.installation_id).any?
```
enescakir: We use `GithubCustomLabel.where(label: github_runner.label, installation_id: github_runner.installation_id)` in several places. We can add `runner.custom_label` helper as `GithubCustomLabel.where(label:, installation_id:).first` to `GithubRunner` model and use it as `if  github_runner.custom_label`

Better one, we might add it to `GithubRunner` as `many_to_one` association. Copilot says that we can use multiple columns as key, but I didn't test it.

```
  many_to_one :custom_label, key: [:label, :installation_id]
```

prog/vm/github_runner.rb
```diff
@@ -422,6 +440,8 @@ def collect_final_telemetry
       vm.incr_destroy
     end
 
+    custom_label = GithubCustomLabel.where(label: github_runner.label, installation_id: github_runner.installation_id).first
+    custom_label&.update(allocated_runner_count: custom_label.allocated_runner_count - 1)
```
enescakir: Same, not concurrency safe

prog/vm/github_runner.rb
enescakir: You can do this with a single DB query, which is also more concurrency-efficient. Something like:
```suggestion
    new_runner_count = Sequel[:allocated_runner_count] + 1
    updated_rows = GithubCustomLabel
      .where(label:, installation_id:)
      .where(new_runner_count <= concurrent_runner_count_limit)
      .update(allocated_runner_count: new_runner_count)

    if updated_rows != 1
      Clog.emit("hit custom label concurrency limit") { {custom_label_concurrency_limit: {label:, installation_id:, repository_name: github_runner.repository_name}} }
      nap rand(5..15)
    end
```

model/github_runner.rb
```diff
@@ -75,6 +75,10 @@ def check_pulse(session:, previous_pulse:)
     end
     aggregate_readings(previous_pulse: previous_pulse, reading: reading, data: {available_memory: available_memory})
   end
+
+  def custom_label
+    GithubCustomLabel.where(name: label, installation_id: installation_id).first
+  end
```
enescakir, line 81: nit:
```suggestion
  def custom_label
    GithubCustomLabel.first(name: label, installation_id:)
  end
```

model/github_runner.rb
```diff
@@ -75,6 +75,10 @@ def check_pulse(session:, previous_pulse:)
     end
     aggregate_readings(previous_pulse: previous_pulse, reading: reading, data: {available_memory: available_memory})
   end
+
+  def custom_label
+    GithubCustomLabel.where(name: label, installation_id: installation_id).first
```
enescakir: I think `actual_label` has custom label name
```suggestion
    GithubCustomLabel.where(name: actual_label, installation_id: installation_id).first
```

prog/github/github_repository_nexus.rb
```diff
@@ -53,13 +53,17 @@ def check_queued_jobs
...
       jobs.each do |job|
         next if job[:status] != "queued"
-        next unless (label = job[:labels].find { Github.runner_labels.key?(it) })
-        queued_labels[label] += 1
+        if (label = job[:labels].find { Github.runner_labels.key?(it) })
+          queued_labels[[label, label]] += 1 # Actual label is the same as the label for predefined labels
+        else
+          custom_label = GithubCustomLabel.where(installation_id: github_repository.installation.id, name: job[:labels]).first
+          queued_labels[[custom_label.name, custom_label.alias_for]] += 1 if custom_label
+        end
```
enescakir:
```suggestion
        elsif (custom_label = GithubCustomLabel.first(installation_id: github_repository.installation.id, name: job[:labels]))
          queued_labels[[custom_label.name, custom_label.alias_for]] += 1
        end
```

prog/github/github_repository_nexus.rb
```diff
@@ -53,13 +53,17 @@ def check_queued_jobs
...
       end
     end
 
-    queued_labels.each do |label, count|
-      idle_runner_count = github_repository.runners_dataset.where(label: label, workflow_job: nil).count
+    queued_labels.each do |(actual_label, label), count|
+      idle_runner_count = github_repository.runners_dataset.where(actual_label: actual_label, workflow_job: nil).count
```
enescakir:
```suggestion
      idle_runner_count = github_repository.runners_dataset.where(actual_label:, workflow_job: nil).count
```

prog/vm/github_runner.rb
```diff
@@ -206,6 +208,25 @@ def quota_available?
...
+    hop_allocate_vm
+  end
+
+  label def apply_custom_label_quota
+    custom_label = github_runner.custom_label
+    hop_allocate_vm unless custom_label.concurrent_runner_count_limit
+
+    new_allocated_runner_count = Sequel[:allocated_runner_count] + 1
+    updated_rows = GithubCustomLabel
+      .where(name: github_runner.label, installation_id: github_runner.installation_id)
```
enescakir:
```suggestion
      .where(name: github_runner.actual_label, installation_id: github_runner.installation_id)
```

prog/vm/github_runner.rb
```diff
@@ -289,7 +310,7 @@ def setup_info
     # We use generate-jitconfig instead of registration-token because it's
     # recommended by GitHub for security reasons.
     # https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-just-in-time-runners
-    data = {name: github_runner.ubid.to_s, labels: [github_runner.label], runner_group_id: 1, work_folder: "/home/runner/work"}
+    data = {name: github_runner.ubid.to_s, labels: [github_runner.actual_label || github_runner.label], runner_group_id: 1, work_folder: "/home/runner/work"}
```
enescakir, line 341: Did we add it for the runners which doesn't have an `actual_label` when we deploy? What do you think about setting `actual_label` to `label` in the migration?

routes/webhook/github.rb
```diff
@@ -59,15 +59,24 @@ def handle_workflow_job(data)
...
-    unless (label = job.fetch("labels").find { Github.runner_labels.key?(it) })
-      return error("Unmatched label")
+    job_labels = job.fetch("labels")
+
+    if (label = job_labels.find { Github.runner_labels.key?(it) })
+      actual_label = label
+    else
+      custom_label = GithubCustomLabel.where(installation_id: installation.id, name: job_labels).first
+      actual_label = custom_label&.name
+      label = custom_label&.alias_for
```
enescakir:
```suggestion
    elsif (custom_label = GithubCustomLabel.first(installation_id: installation.id, name: job_labels))
      actual_label = custom_label.name
      label = custom_label.alias_for
```

prog/vm/github_runner.rb
```diff
@@ -432,6 +453,18 @@ def collect_final_telemetry
       vm.incr_destroy
     end
 
+    custom_label = github_runner.custom_label
+    if custom_label
```
enescakir:
```suggestion
    if (custom_label = github_runner.custom_label)
```

prog/vm/github_runner.rb
```diff
@@ -432,6 +453,18 @@ def collect_final_telemetry
...
+      new_runner_count = Sequel[:allocated_runner_count] - 1
+      updated_rows = GithubCustomLabel
+        .where(name: github_runner.label, installation_id: github_runner.installation_id)
```
enescakir:
```suggestion
        .where(name: github_runner.actual_label, installation_id: github_runner.installation_id)
```

prog/vm/github_runner.rb
velioglu: I've added migration but it can still be required for the runners provisioned between the update script run and the new code deployed.So, keeping it.

PR #3897, opened 2025-09-15
Update Billing after Kubernetes Nodepool is Scaled

aamederen:
Update billing after kubernetes nodepool is scaled

KubernetesClusterNexus#create_billing_records is replaced with
update_billing_records. This method compares the desired list and
existing list of billing records and performs add/remove to make
the desired list the reality with minimal changes.

The removal of billing records are done on the later-created ones,
in order to create less confusion around the final bill. If a user
creates a 3-node cluster, then scales it to 5 nodes to reduce it back
to 3 nodes, at the end of the month, the user will see 3 long and 2 short
nodes in the billing items.

The billing update is done right at the beginning of node removal and
at the end of node addition process, skewing the bill in favor of the user.
The update semaphore is raised right after the node removal begins, and
the updating logic eliminates the nodes flagged for retirement
from the equation.

PS: Writing specs took horribly long for me, then finally I realized that
`billing_record.finalize` converts `span` to something like `tstzrange(now(), now())` 
and that converts both ends of the range to NULL, which is by definition is
evaluated as an active billing record. Should put a constraint to that.


prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -113,20 +113,37 @@ def before_run
...
     nap 10 unless kubernetes_cluster.nodepools.all? { it.strand.label == "wait" }
-    hop_create_billing_records
+    hop_wait
   end
 
-  label def create_billing_records
-    records = kubernetes_cluster.all_nodes.flat_map(&:billing_record)
+  label def update_billing_records
+    decr_update_billing_records
+    desired_records = kubernetes_cluster.all_nodes.reject(&:retire_set?).flat_map(&:billing_records).tally
```
jeremyevans: I think this reads nicer:

```suggestion
    desired_records = kubernetes_cluster
      .all_nodes
      .reject(&:retire_set?)
      .flat_map(&:billing_records)
      .tally
```

Up to you, but I think doing it here and other places with long lines could improve readability.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -113,20 +113,37 @@ def before_run
...
-    records.each do |record|
-      BillingRecord.create(
-        project_id: kubernetes_cluster.project_id,
-        resource_id: kubernetes_cluster.id,
-        resource_name: kubernetes_cluster.name,
-        billing_rate_id: BillingRate.from_resource_properties(record[:type], record[:family], kubernetes_cluster.location.name)["id"],
-        amount: record[:amount]
-      )
+      br = BillingRate.from_resource_properties(record[:type], record[:family], kubernetes_cluster.location.name)
+      kubernetes_cluster.active_billing_records_dataset.where(billing_rate_id: br["id"], amount: record[:amount]).order_by(Sequel.desc(Sequel.function(:lower, :span))).limit(surplus).each do |r|
```
jeremyevans: `reverse` can be simpler than `order`+`Sequel.desc`. Issuing queries inside a `Dataset#each` block is not supported in Sequel (though it does work with the postgres adapter when not streaming results):

```suggestion
      kubernetes_cluster
        .active_billing_records_dataset
        .where(billing_rate_id: br["id"], amount: record[:amount])
        .reverse(Sequel.function(:lower, :span))
        .limit(surplus)
        .all do |r|
```

aamederen:
@jeremyevans thanks for the review.

> It seems weird to charge for "3 long and 2 short nodes" if the user resizes during the billing period. I would expect a bill of "3 nodes for X hours, 5 nodes for Y hours" or something to that effect.
I'm not knowledgeable about our billing, but if there is a resize, would it be possible to finalize the billing record for the previous size, and then start a new billing record for the new size?

Currently, we don't have a "single bill for 3 nodes", but instead, we have a separete billing records for each node (compute and storage, separately), both in PG and K8s. The idea is to not change the records of the unaffected nodes.

In PG, when you change the SKU (vcpus) of a server, it makes sense to throw all away and create records from scratch. However, if/when you enable/disable HA to add/remove standby nodes, the billing record for the primary node, which was not changed at all, also gets thrown away. This is a less likely event in PG so this behavior is probably OK.

In K8s, we may not even allow changing SKUs but we expect to make add/remove node automatic so this can happen multiple times a day.

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -243,24 +244,84 @@
...
+      end
+
+      expect(kubernetes_cluster.active_billing_records.map(&:id)).to eq records
+    end
+
+    it "creates missing billing records and finalizes surplus billing records" do
+      old_records = kubernetes_cluster.active_billing_records.map(&:id)
+      expect(old_records.length).to eq 4
+
+      older_cp_record, newer_cp_record = kubernetes_cluster.active_billing_records.select { |it| it.billing_rate["resource_type"] == "KubernetesControlPlaneVCpu" }
```
jeremyevans:
```suggestion
      older_cp_record, newer_cp_record = kubernetes_cluster.active_billing_records.select { it.billing_rate["resource_type"] == "KubernetesControlPlaneVCpu" }
```

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -243,24 +244,84 @@
...
+      new_records = kubernetes_cluster.active_billing_records.map(&:id)
+      expect(new_records.length).to eq 6
+      expect(newer_cp_record.reload.span.end).not_to be_nil # the newer record is finalized
+      expect(older_cp_record.reload.span.end).to be_nil
+
+      expect((new_records - old_records).length).to eq 3 # 2 for the new worker node, 1 for the new bigger CP node
+      expect((new_records & old_records).length).to eq 3 # 1 CP node and 2 worker nodes stayed the same
+      expect((old_records - new_records).length).to eq 1 # 1 removed CP node
+      expect(new_records).to include(*(old_records - [newer_cp_record.id]))
+      expect(new_records).not_to include newer_cp_record.id
```
jeremyevans, line 340: In addition to the above, might be worth adding expectations for the expected values of the new records.

jeremyevans (commented):
If we currently have separate billing records for each node, what's the issue with using that same approach even when add/remove node is automatic? I realize it results in more billing records, but I would think that approach would be more accurate and less complex than this one. This approach of updating existing billing records makes me uneasy. However, that could be due to my lack of experience in this area.

byucesoy:
On why we are not doing this for PG:

A similar approach could be applied to PostgreSQL in some cases. However, there are scenarios where it isn’t possible, such as:
- Scaling from burstable to standard (or vice versa)
- Changing the instance type (on AWS)

We could implement a mixed approach; updating where possible and falling back to creating new records where updates aren’t feasible. But I chose to keep the code simple rather than introducing special-case logic. So far, we haven’t received any complaints from customers about invoices being confusing (and I don’t think they are particularly complicated anyway), so we’ve kept it as is.

The main drawback of always creating new records is related to our discount system. We apply a “loyalty discount” when a resource remains active for more than 28 days. Specifically, our billing system treats every month as 28 days and caps charges accordingly, ensuring users pay the same regardless of whether a month has 28, 29, 30, or 31 days. Effectively, the last few days of each month are free for long-running resources.

When scaling operations generate multiple billing records, users no longer benefit automatically from this discount, because instead of one long record, you end up with multiple smaller records that covers the entire month. To address this, I manually review scaling activity each month and apply credits where appropriate. Automating this is technically possible but not trivial, and it has been on my to-do list for a while. That said, the manual review doesn’t take much time, and I want to review scaling operations personally anyway, so I haven’t prioritized automation yet.

mohi-kalantari (approved):

PR #3899, opened 2025-09-15
Enable i8g instances

byucesoy:
We support various instance types on AWS, however they all come with fixed vCPU/storage ratio. For use-cases with more storage needs, we are also introducing i8g instance type.


spec/prog/vm/nexus_spec.rb
```diff
@@ -199,6 +199,17 @@ def requested_disk_size(st)
       expect(st.stack[0]["gpu_count"]).to eq(0)
     end
 
+    it "creates correct number of storage volumes for storage optimized instance types" do
+      loc = Location.create(name: "us-west-2", provider: "aws", project_id: prj.id, display_name: "us-west-2", ui_name: "us-west-2", visible: true)
+      storage_volumes = [
+        {encrypted: true, size_gib: 30},
+        {encrypted: true, size_gib: 7500}
+      ]
+      expect(VmStorageVolume).to receive(:create).exactly(3).times
```
jeremyevans: Probably better to check that the created VmStorageVolume instances have expected values after `assemble`. In general, it's better to avoid mocking like this unless it is necessary (e.g. mocking SSH sessions and external HTTP service responses). I realize we have a lot of (brittle) specs with mocking like this already, but I think we should avoid adding more.

PR #3914, opened 2025-09-16
Make Project#has_valid_payment_methods? return true if discount is 100%

jeremyevans:
With a 100% discount, there will never be a bill, so a valid payment method is not needed. I can only see this causing problems if the 100% discount is temporary and not permanent.

----

> [!IMPORTANT]
> `has_valid_payment_method?` now returns true if discount is 100% in `project.rb`, with corresponding test added in `project_spec.rb`.
> 
>   - **Behavior**:
>     - `has_valid_payment_method?` in `project.rb` now returns true if `discount` is 100.
>   - **Tests**:
>     - Added test in `project_spec.rb` to verify `has_valid_payment_method?` returns true when `discount` is 100.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for acd5abc0de5259d0821957a03fd4914b270e37cd. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -53,6 +53,7 @@ class Project < Sequel::Model
 
   def has_valid_payment_method?
     return true unless Config.stripe_secret_key
+    return true if discount == 100
     !!billing_info&.payment_methods&.any? || (!!billing_info && credit > 0)
   end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
@enescakir To unblock things, and because I think this is low risk, I plan to merge this. If you think it will cause issues, let me know and I can revert.

PR #3917, opened 2025-09-17
Make *_NAME_OR_UBID matchers yield uuid instead of ubid

jeremyevans:
This DRYs up some route code.

Implement this using a symbol matcher that does the ubid to uuid conversion.

This also avoids an unnecessary query to look up a resource when the route segment looks like a ubid, but is not a valid ubid.


fdr (approved):
too bad there's no ubid extension for postgres (and probably not worth it to write and deploy one...at least, for a good while)

PR #3923, opened 2025-09-18
PG Major Upgrades API, UI, CLI and SDK

shikharbhardwaj:
## Description

API and UI changes for PG major version upgrades.

UI Screenshots

Screenshot 2025-10-10 at 2 44 19 PM

Screenshot 2025-10-10 at 2 47 27 PM

Screenshot 2025-10-10 at 2 47 58 PM

Screenshot 2025-10-10 at 2 49 59 PM

----

> [!IMPORTANT]
> Adds major version upgrade functionality for PostgreSQL databases, including CLI commands, API endpoints, UI elements, and validation logic.
> 
>   - **Behavior**:
>     - Adds `upgrade` and `show-upgrade-status` commands in `cli-commands/pg/post/upgrade.rb` and `cli-commands/pg/post/show-upgrade-status.rb` for scheduling and checking PG major version upgrades.
>     - Updates `openapi.yml` to include new API endpoints for upgrading PG databases and checking upgrade status.
>     - Implements upgrade validation in `lib/validation.rb` to prevent upgrades during convergence, failover, or if the database is already at the latest version.
>   - **Models**:
>     - Adds `upgrade_stage` and `upgrade_status` methods to `PostgresResource` in `model/postgres/postgres_resource.rb`.
>     - Introduces `PostgresDatabaseUpgradeStatus` schema in `openapi.yml`.
>   - **UI**:
>     - Adds "Upgrade" tab and upgrade progress tracking in `views/postgres/show.erb` and `views/postgres/upgrade.erb`.
>     - Introduces new icons in `views/components/icon.erb` for upgrade status.
>   - **Tests**:
>     - Adds tests for upgrade functionality in `spec/routes/api/cli/pg/upgrade_spec.rb`, `spec/routes/api/cli/pg/show-upgrade-status_spec.rb`, and `spec/routes/web/project/postgres_spec.rb`.
>     - Updates `spec/serializers/postgres_spec.rb` to include upgrade-related serialization.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8a144678bacd32e44906ba3bcdb578983be4d5a8. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


migrate/20250915_add_postgres_server_version.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
```
jeremyevans:
```suggestion
  change do
```

migrate/20250915_add_postgres_server_version.rb
```diff
@@ -0,0 +1,15 @@
...
+    alter_table(:postgres_server) do
+      add_column :version, :postgres_version
+    end
+  end
+
+  down do
+    alter_table(:postgres_server) do
+      drop_column :version
+    end
+  end
```
jeremyevans, line 12:
```suggestion
```

prog/postgres/postgres_server_nexus.rb
```diff
@@ -53,7 +53,8 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
         timeline_access: timeline_access,
         representative_at: representative_at,
         synchronization_status: synchronization_status,
-        vm_id: vm_st.id
+        vm_id: vm_st.id,
+        version: postgres_resource.current_version
       ) { it.id = ubid.to_uuid }
```
jeremyevans: Not an issue with this PR, but it would be good to switch the method call to use `PostgresServer.create_with_id(ubid.to_uuid, ...)` and drop the block.

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -0,0 +1,133 @@
...
+
+  def wait_for_postgres_to_start
+    Timeout.timeout(60) do
+      loop do
+        r "sudo -u postgres pg_isready"
+        break
+      rescue
+        sleep 1
+      end
+    end
```
jeremyevans: In general, `Timeout.timeout` is a terrible API and should never be used. You already have a loop in this method, just have it reraise the exception if it is beyond the deadline:

```suggestion
    deadline = Time.now + 60
    loop do
      r "sudo -u postgres pg_isready"
      break
    rescue
      raise if deadline < Time.now
      sleep 1
    end
```

If you are worried about the `r` call blocking for indefinite time, maybe use `sudo -u postgres timeout 1s pg_isready`? I don't think that's needed if this is only running on local host against a unix socket, though.

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -0,0 +1,133 @@
...
+  end
+
+  def run_post_upgrade_extension_update
+    databases = r("sudo -u postgres psql -t -c 'SELECT datname FROM pg_database WHERE datistemplate = false;'").split("\n").map(&:strip)
+
+    scripts = EXTENSION_UPGRADE_SCRIPTS[@version]
+    scripts.each do |extension, script|
+      @logger.info("Running post upgrade extension update for #{extension}")
+      databases.each do |database|
+        installed = r("sudo -u postgres psql -d #{database} -t -c \"SELECT 1 FROM pg_extension WHERE extname = '#{extension.shellescape}'\"").strip
```
jeremyevans: This looks dangerous, though I'm not sure whether it is exploitable given PostgreSQL database naming rules. However, considering that `$` is an allowed character in a PostgreSQL database name, I couldn't rule out exploitation. If exploitable, this would result in root remote code execution in the VM. You need to `shellescape` the database.

Note that `extension.shellescape` isn't correct. `shellescape` output should be used unquoted. You are using not only in quotes, but instead nested quotes. See https://docs.ruby-lang.org/en/master/Shellwords.html#method-c-shellescape

If there was a `'` in `extension`, `shellescape` would replace it with `\'`, which would escape it from the shell, but not escape it from the postgres string. You would need to double the `'` inside extension, then call shellescape, then include results outside of the shell quote:

```suggestion
        installed = r("sudo -u postgres psql -d #{database.shellescape} -t -c \"SELECT 1 FROM pg_extension WHERE extname = '\"#{extension.gsub("'", "''").shellescape}\"'\"").strip
```

I'm not 100% that is correct. Please test with various types of input and ensure it doesn't allow either arbitrary shell command execution or SQL injection.

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -0,0 +1,133 @@
...
+        if installed == "1"
+          @logger.info("Running post upgrade extension update for #{extension} on database #{database}")
+          r "sudo -u postgres psql -d #{database} -v 'ON_ERROR_STOP=1' -c #{script.shellescape}"
```
jeremyevans:
```suggestion
          r "sudo -u postgres psql -d #{database.shellescape} -v 'ON_ERROR_STOP=1' -c #{script.shellescape}"
```

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -0,0 +1,133 @@
...
+        end
+      end
+    end
+  end
+
+  def upgrade
+    puts "Creating upgrade directory"
```
jeremyevans: There is a mix of `puts` and `@logger` in this class. Please choose one (no preference as to which) and be consistent. Unless there is a specific reason you wanted the output to be different?

routes/project/location/postgres.rb
```diff
@@ -538,6 +538,50 @@ class Clover
           end
         end
       end
+
+      r.is "upgrade" do
+        r.get do
```
jeremyevans: This uses an api-only response. The web GET case is handled by the show_object call many lines above (line 110 in this review). However, that is only true if non-read-replica databases. Read-replica databases can hit this code. I would add a comment and restrict this to API routes.

```suggestion
        # api-only route, web GET upgrade route handled by r.show_object call earlier in route
        r.get api? do
```

routes/project/location/postgres.rb
```diff
@@ -538,6 +538,50 @@ class Clover
           end
         end
       end
+
+      r.is "upgrade" do
```
jeremyevans, line 541: Probably should have appropriate checks here that apply to both the GET and POST routes. I assume you at least want to check whether the database is a read-replica, since you are not showing the upgrade page in the read-replica case when making the changes above.

Additionally, when adding API features, make sure you update the SDK and CLI at the same time to support the features.

routes/project/location/postgres.rb
```diff
@@ -538,6 +538,50 @@ class Clover
...
+          unless pg.needs_upgrade?
+            raise CloverError.new(400, "InvalidRequest", "Database is not upgrading")
+          end
+
+          {
+            current_version: pg.current_version,
+            desired_version: pg.version,
+            upgrade_status: pg.upgrade_status,
+            upgrade_progress: pg.upgrade_progress
+          }
```
jeremyevans: Probably best to add a simpler serializer and use it both the GET and POST routes.

routes/project/location/postgres.rb
```diff
@@ -538,6 +538,50 @@ class Clover
...
+          Validation.validate_postgres_upgrade(
+            version_int: pg.version_int,
+            needs_convergence: pg.needs_convergence?,
+            ongoing_failover: pg.ongoing_failover?,
+            read_replica: pg.read_replica?,
+            needs_upgrade: pg.needs_upgrade?
+          )
+
+          audit_log(pg, "upgrade")
+          pg.update(version: pg.version_int + 1)
```
jeremyevans: Typically, `audit_log` call comes after action, and is inside the same transaction:

```suggestion
          DB.transaction do
            pg.update(version: pg.version_int + 1)
            audit_log(pg, "upgrade")
          end
```

serializers/postgres.rb
```diff
@@ -12,6 +12,7 @@ def self.serialize_internal(pg, options = {})
       storage_size_gib: pg.storage_size_gib,
       target_storage_size_gib: pg.target_storage_size_gib,
       version: pg.version,
+      current_version: pg.current_version,
```
jeremyevans: When adding a serializer column, update the SDK (`sdk/ruby/lib/ubicloud/model/postgres.rb`) and CLI (`cli-commands/pg/post/show.rb`) as well.

views/postgres/upgrade.erb
```diff
@@ -0,0 +1,77 @@
...
+  <% elsif @pg.needs_upgrade? %>
+    <% if @pg.upgrade_stage == "upgrade_failed" %>
+      <p>Database upgrade failed. We have reverted the database to the previous version, and it is now available again.</p>
+    <% else %>
+      <p>Database upgrade is in progress.</p>
+    <% end %>
+    <% database_available_upgrade_stages = ["start", "wait_for_standby", "wait_for_maintenance_window"]
+    
+    upgrade_status = @pg.upgrade_progress.map do |stage|
+      [
```
jeremyevans, line 56: Indentation starting here looks off. Can we keep a 2-space indentation level here?

prog/postgres/upgrade_postgres_resource.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  label def finish_upgrade
+    pop "upgrade prog finished"
+  end
+
+  label def upgrade_failed
+    if candidate_server
+      logs = candidate_server.vm.sshable.cmd("sudo journalctl -u upgrade_postgres")
+      logs.split("\n").each { |line| Clog.emit("Postgres resource upgrade failed") { {resource_id: postgres_resource.id, log: line} } }
```
jeremyevans: This could potentially be logged multiple times for the same server, assuming something was blocking it for destruction for more than 6 hours. Maybe also check that the destroy semaphore isn't set on the candidate server?

lib/validation.rb
```diff
@@ -304,4 +304,12 @@ def self.validate_from_option_tree(option_tree, option_parents, params)
       end
     end
   end
+
+  def self.validate_postgres_upgrade(version_int:, needs_convergence:, ongoing_failover:, read_replica:, needs_upgrade:)
```
jeremyevans: Have this accept a single `PostgresResource` argument, unless you have callers where you do not have a `PostgresResource` argument (I didn't see any).

rhizome/postgres/lib/postgres_upgrade.rb
shikharbhardwaj: Thanks for catching this. I moved the scripts to `stdin` instead.

rhizome/postgres/lib/postgres_upgrade.rb
shikharbhardwaj: I was initially planning to expose progress via stdin, but that seemed like a Bad Idea™, missed moving these to logger.

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Fixed this in the other PR

serializers/postgres.rb
shikharbhardwaj: I added things to the CLI and SDK in this branch, but I could not get one of the commands to be tested. For me, mocking and test setup in the CLI code is not intuitive, could you please give some pointers on how best to test these changes? I added more details in a commnet on the specific command file.

cli-commands/pg/post/show-upgrade-status.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
shikharbhardwaj, line 1: @jeremyevans I could not test the success case for this file in the CLI goldens, as it needs a PG instance that is "upgrading" to show details. I tried setting up a new PG instance in the test setup code in `golden_files_spec.rb`, but that ran into a bunch of key conflicts due to the UUIDs being mocked. I worked around testing the upgrade command by moving the existing test-pg to version 16, so that it can be upgraded.

What do you recommend as the best way to test this?

routes/project/location/postgres.rb
shikharbhardwaj: The GET route is only intended to be called on currently upgrading databases, so the `needs_upgrade?` check suffices. A read-replica would never have an ongoing upgrade, so would fail that check as well.

prog/postgres/upgrade_postgres_resource.rb
shikharbhardwaj: Done in the other PR

cli-commands/pg/post/show-upgrade-status.rb
jeremyevans: Not everything needs a golden file test. Golden file tests are only capable of testing output, not behavior. You should add a test for the specific command (e.g. `spec/routes/api/cli/pg/upgrade_spec.rb`). Look at the other specific command tests, it should be similar.

lib/validation.rb
```diff
@@ -304,4 +304,12 @@ def self.validate_from_option_tree(option_tree, option_parents, params)
       end
     end
   end
+
+  def self.validate_postgres_upgrade(postgres_resource:)
```
jeremyevans: In general, if a method requires a single argument, it should be a positional argument instead of a keyword argument. An exception would be if you definite plans to add other required arguments in the future.

```suggestion
  def self.validate_postgres_upgrade(postgres_resource)
```

routes/project/location/postgres.rb
```diff
@@ -541,6 +541,37 @@ class Clover
...
+            raise CloverError.new(400, "InvalidRequest", "Database is not upgrading")
+          end
+
+          Serializers::PostgresUpgrade.serialize(pg)
+        end
+
+        r.post do
+          authorize("Postgres:edit", pg.id)
+
+          Validation.validate_postgres_upgrade(postgres_resource: pg)
```
```suggestion
          Validation.validate_postgres_upgrade(pg)
```

jeremyevans (commented):
In some cases, you say this is fixed in the other PR, but I'm not sure which PR that is. Can you provide a link?

Overall, this looks good, but since the migration and rhizome changes have been removed from this, I'm guessing you wouldn't want to merge until after the other PR is merged.

jeremyevans (approved):
Sorry, didn't realize the "other PR" is actually the merge base.

lib/validation.rb
```diff
@@ -334,4 +334,12 @@ def self.validate_from_option_tree(option_tree, option_parents, params)
       end
     end
   end
+
+  def self.validate_postgres_upgrade(postgres_resource)
+    fail ValidationFailed.new({needs_convergence: "Database cluster is waiting for convergence, please wait for it to complete"}) if postgres_resource.needs_convergence?
+    fail ValidationFailed.new({ongoing_failover: "Database cluster is in the middle of a failover, please wait for it to complete"}) if postgres_resource.ongoing_failover?
```
byucesoy: Do we really need to reject in the case of `needs_convergence` and `ongoing_failover`? We can accept the update and the system would eventually converge. Is there is an edge case I'm missing? At least `needs_convergence` seem pretty safe, not sure about `ongoing_failover`.

lib/validation.rb
```diff
@@ -334,4 +334,12 @@ def self.validate_from_option_tree(option_tree, option_parents, params)
...
+    fail ValidationFailed.new({read_replica: "Read replicas cannot be upgraded"}) if postgres_resource.read_replica?
+    fail ValidationFailed.new({needs_upgrade: "Database is already upgrading"}) if postgres_resource.target_version != postgres_resource.version
```
byucesoy: Instead of rejecting this and we can act as it is no-op.

views/postgres/upgrade.erb
```diff
@@ -0,0 +1,80 @@
...
+  <% if @pg.display_state == "creating" %>
+    <p class="text-gray-500 mt-2">Upgrade operation is not available while the database is being created.</p>
+  <% elsif @pg.target_version != @pg.version %>
+    <% if @pg.upgrade_stage == "upgrade_failed" %>
+      <p>Database upgrade failed. We have reverted the database to the previous version, and it is now available again.</p>
+    <% else %>
+      <p>Database upgrade is in progress.</p>
+    <% end %>
+    <% database_available_upgrade_stages = ["start", "wait_servers_to_be_ready", "wait_for_maintenance_window"]
+    hidden_stages = ["recycle_representative_server", "provision_servers"]
```
byucesoy: I think we spend too much effort on generating stages programmatically from the labels, both in frontend and also in backend.

It would be much simpler to;
- Have a hard-coded list of valid upgrade stages here (which we already have at the top of this file)
- Backend would only send the current_stage
- In front end we would have something like
```ruby
status = "pending"
hard_coded_stages.map do |stage|
  if status == "running"
    status = "done"
  elsif stage == current_stage
    status = "running"
  end
  status
end
```

For API, I think it would be enough to report current_stage.

lib/validation.rb
shikharbhardwaj: If we allow this, the user can update target version multiple times while the current upgrade is ongoing. This can lead to the current_version and target_version being non-adjacent, which is an implicit assumption in many places (eg. we call pg_upgrade with target_version instead of current_version + 1). I can survey and update all such cases if you think its important to allow this. I think allowing the update during failovers should be okay.

lib/validation.rb
shikharbhardwaj: `needs_convervence?` already covers this condition, removed this. But users will still not be able to upgrade a currently upgrading database.

PR #3924, opened 2025-09-18
Extend e2e tests

mohi-kalantari:
**Make k8s e2e tests dynamic for different node counts**

In this commit, we rely on the strand stack to save the state of whether we have applied the hosts entries on a node or not.

This way, when adding another node to the cluster, we don't need to add another label to ensure host entries on it

**Add e2e test to check the data hash after data migration**

Number of worker nodes in the e2e test cluster is increased to two so we can test the data migration. During the test, first we write some random data to the pod's mount location, then cordon its node in order to trigger the migration and then wait for the new pod to be running and then finally check the hash after migration.


prog/test/kubernetes.rb
```diff
@@ -193,4 +216,17 @@ def kubernetes_test_project
   def kubernetes_cluster
     @kubernetes_cluster ||= KubernetesCluster.with_pk(frame["kubernetes_cluster_id"])
   end
+
+  def node_host_entries_set?(node_name)
+    frame = strand.stack.first
+    frame["nodes_status"] ||= {}
+    frame["nodes_status"][node_name] == true
```
jeremyevans:
```suggestion
    strand.stack.first.dig("nodes_status", node_name) == true
```

prog/test/kubernetes.rb
```diff
@@ -193,4 +216,17 @@ def kubernetes_test_project
...
+  end
+
+  def set_node_entries_status(node_name)
+    frame = strand.stack.first
+    frame["nodes_status"] ||= {}
+    frame["nodes_status"][node_name] = true
```
jeremyevans, line 254: I don't think it applies to this prog, since it's only used in E2E testing, but if this was used more generally, you would need to remove nodes from the stack when they are removed from the cluster, or otherwise you have unbounded growth possible.

prog/test/kubernetes.rb
```diff
@@ -148,6 +151,26 @@ def self.assemble
...
+      hop_destroy_kubernetes
+    end
+    update_stack({"read_hash" => read_hash})
+    hop_test_pod_data_migration
+  end
+
+  label def test_pod_data_migration
+    pod_node = kubernetes_cluster.client.kubectl("get pods ubuntu-statefulset-0 -ojsonpath={.spec.nodeName}").strip
+    kubernetes_cluster.client.kubectl("cordon #{pod_node}")
+    kubernetes_cluster.client.kubectl("delete pod ubuntu-statefulset-0 --wait=false")
```
jeremyevans: I don't have any K8s knowledge, but it seems weird that you delete the pod here, but still are able to operate on it later. I guess it automatically creates a new pod after one is deleted?

mohi-kalantari: yes exactly. It is recreated after deletion.

prog/test/kubernetes.rb
mohi-kalantari: That's right. we don't have any node deletions in the test. thanks for the tip.

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -75,6 +75,13 @@ def delete_pvc(namespace, name)
       run_kubectl("-n", namespace, "delete", "pvc", name, "--wait=false", "--ignore-not-found=true")
     end
 
+    # This function will first try to get the pvc in order to make sure pvc exists
+    def remove_finalizers(namespace, name)
```
aamederen: nitpick: rename to `remove_pvc_finalizers`

PR #3927, opened 2025-09-19
Prevent Monitoring Exceptions Before a GH runner is ready

aamederen:
There's a race for pulse_checking a GH runner by the monitor and creating of the VM by the respirator. We can reduce the amount of exceptions by gracefully reporting "down" when the VM is not ready.

----

> [!IMPORTANT]
> Modify `init_health_monitor_session` to handle unready VMs and update tests accordingly.
> 
>   - **Behavior**:
>     - Modify `init_health_monitor_session` in `github_runner.rb` to return `nil` for `ssh_session` if VM is not ready (i.e., `vm.strand.label` is not "wait").
>     - Update `check_pulse` in `github_runner.rb` to handle `nil` `ssh_session` by returning "down".
>   - **Tests**:
>     - Add test in `github_runner_spec.rb` to verify `init_health_monitor_session` returns `nil` when VM is not ready.
>     - Update `check_pulse` test in `github_runner_spec.rb` to check behavior with `nil` `ssh_session`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 40322d95181dc4f24c6eb3ea598870a51dd65bb4. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -62,7 +62,7 @@ class GithubRunner < Sequel::Model
 
   def init_health_monitor_session
     {
-      ssh_session: vm.sshable.start_fresh_session
+      ssh_session: (vm && vm.strand.label == "wait") ? vm.sshable.start_fresh_session : nil
     }
   end
 
--- a/spec/model/github_runner_spec.rb
+++ b/spec/model/github_runner_spec.rb
@@ -94,7 +94,17 @@ RSpec.describe GithubRunner do
 
   it "initiates a new health monitor session" do
     expect(github_runner.vm.sshable).to receive(:start_fresh_session)
+    expect(github_runner.vm).to receive(:strand).and_return(instance_double(Strand, label: "wait"))
     github_runner.init_health_monitor_session
+  end
+
+  it "initiates an empty health monitor session if VM is not ready" do
+    allow(github_runner.vm).to receive(:strand).and_return(instance_double(Strand, label: "not-wait"))
+    expect(github_runner.vm.sshable).not_to receive(:start_fresh_session)
+    expect(github_runner.init_health_monitor_session[:ssh_session]).to be_nil
+
+    github_runner.update(vm_id: nil)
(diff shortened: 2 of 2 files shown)


enescakir:
We can also select only ready runners as `GithubRunner.exclude(ready_at: nil)`

https://github.com/ubicloud/ubicloud/blob/main/bin/monitor#L42-L53

model/github_runner.rb
```diff
@@ -62,7 +62,7 @@ def provision_spare_runner
 
   def init_health_monitor_session
     {
-      ssh_session: vm.sshable.start_fresh_session
+      ssh_session: (vm && vm.strand.label == "wait") ? vm.sshable.start_fresh_session : nil
```
jeremyevans, line 65:
```suggestion
      ssh_session: (vm.sshable.start_fresh_session if vm&.strand.label == "wait")
```

spec/model/github_runner_spec.rb
```diff
@@ -94,9 +94,19 @@ def clog_emit_hash
 
   it "initiates a new health monitor session" do
     expect(github_runner.vm.sshable).to receive(:start_fresh_session)
+    expect(github_runner.vm).to receive(:strand).and_return(instance_double(Strand, label: "wait"))
```
jeremyevans, line 97: Avoid mocking other than SSH/HTTP requests. Create a strand and with the appropriate label set, in this spec and the other new spec.

I realize that we mock database access in many other specs, but doing so makes the specs brittle. We should not add new database mocking, and eventually I want to remove the database mocking we currently do (unless such mocking is required to properly mock the SSH/HTTP access).

aamederen:
closing in favor of #3987

PR #3928, opened 2025-09-19
Expose Scaling K8s Nodepools

aamederen:
This PR adds a route and a UI for resizing nodepool(s). SDK & CLI to follow up.

image

----

> [!IMPORTANT]
> This PR adds functionality to resize Kubernetes nodepools via API, CLI, and UI, with validation and tests.
> 
>   - **Behavior**:
>     - Adds `resize-nodepool` route in `routes/project/location/kubernetes_cluster.rb` to resize Kubernetes nodepools.
>     - Validates node count as a positive integer and checks vCPU quota when scaling up.
>     - Updates `openapi.yml` to include new endpoint for resizing nodepools.
>   - **CLI**:
>     - Adds `resize-nodepool` command in `cli-commands/kc/post/resize-nodepool.rb` for resizing nodepools via CLI.
>     - Updates CLI help files to include `resize-nodepool` command.
>   - **UI**:
>     - Updates `views/kubernetes-cluster/settings.erb` to add UI for resizing nodepools.
>   - **Models**:
>     - Adds `resize_nodepool` method to `KubernetesCluster` in `sdk/ruby/lib/ubicloud/model/kubernetes_cluster.rb`.
>   - **Tests**:
>     - Adds tests for resizing nodepools in `kubernetes_cluster_spec.rb` and `kubernetes_cluster_spec.rb` for both API and web interfaces.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 35a842aaa5059af9c51b8128431c3da0115b1834. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli-commands/kc/post/resize-nodepool.rb
+++ b/cli-commands/kc/post/resize-nodepool.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+UbiCli.on("kc").run_on("resize-nodepool") do
+  desc "Resize a Kubernetes cluster nodepool"
+
+  banner "ubi kc (location/kc-name | kc-id) resize-nodepool (np-name | np-id) node-count"
+
+  args 2
+
+  run do |nodepool_ref, node_count, _, cmd|
+    node_count = Integer(node_count, exception: false)
+
+    unless node_count&.>(0)
+      cmd.raise_failure("the node count must be a positive integer")
+    end
+
+    sdk_object.resize_nodepool(nodepool_ref, node_count)
+
+    response("Resized nodepool #{nodepool_ref} to #{node_count} nodes.")
+  end
+end
--- a/clover.rb
+++ b/clover.rb
@@ -110,6 +110,7 @@ class Clover < Roda
 
   path("Project", class_name: true, &:path)
   path("GithubInstallation", class_name: true) { "#{it.project.path}/github/#{it.ubid}" }
+  path("KubernetesNodepool", class_name: true) { "#{it.cluster.project.path}#{it.cluster.path}/nodepool/#{it.ubid}" }
 
   # :nocov:
   if Config.test? && defined?(SimpleCov)
--- a/clover.rb
+++ b/clover.rb
@@ -68,7 +68,7 @@ class Clover < Roda
   symbol_matcher(:ubid_uuid, /([a-tv-z0-9]{26})/) do |s|
     UBID.to_uuid(s)
   end
-  [Firewall, KubernetesCluster, LoadBalancer, PostgresResource, PrivateSubnet, Vm].each do |model|
+  [Firewall, KubernetesCluster, KubernetesNodepool, LoadBalancer, PostgresResource, PrivateSubnet, Vm].each do |model|
     sym = :"#{model.table_name}_ubid_uuid"
     symbol_matcher(sym, /(#{model.ubid_type}[a-tv-z0-9]{24})/) do |ubid|
       if (uuid = UBID.to_uuid(ubid))
--- a/clover.rb
+++ b/clover.rb
@@ -98,6 +98,7 @@ class Clover < Roda
     ApiKey
     Firewall
     KubernetesCluster
+    KubernetesNodepool
     LoadBalancer
     Location
     ObjectTag
@@ -110,7 +111,6 @@ class Clover < Roda
 
   path("Project", class_name: true, &:path)
   path("GithubInstallation", class_name: true) { "#{it.project.path}/github/#{it.ubid}" }
-  path("KubernetesNodepool", class_name: true) { "#{it.cluster.project.path}#{it.cluster.path}/nodepool/#{it.ubid}" }
 
   # :nocov:
   if Config.test? && defined?(SimpleCov)
--- a/routes/project/location/kubernetes_cluster.rb
+++ b/routes/project/location/kubernetes_cluster.rb
@@ -70,6 +70,7 @@ class Clover
 
           r.post "resize" do
             authorize("KubernetesCluster:edit", kc.id)
+            handle_validation_failure("kubernetes-cluster/show") { @page = "settings" }
             node_count = typecast_params.pos_int!("node_count")
             Validation.validate_kubernetes_worker_node_count(node_count)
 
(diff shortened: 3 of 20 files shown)


routes/project/location/kubernetes_cluster.rb
```diff
@@ -54,6 +54,41 @@ class Clover
         response["content-disposition"] = "attachment; filename=\"#{kc.name}-kubeconfig.yaml\""
         kc.kubeconfig
       end
+
+      r.on "nodepool" do
+        r.on KUBERNETES_NODEPOOL_NAME_OR_UBID do |kn_name, kn_ubid|
```
jeremyevans: You only have one route under this. Unless you plan to add new routes later, best to collapse:


```suggestion
      r.post "nodepool", KUBERNETES_NODEPOOL_NAME_OR_UBID, "collapse" do |kn_name, kn_ubid|
```

clover.rb
```diff
@@ -100,6 +100,7 @@ class Clover < Roda
 
   path("Project", class_name: true, &:path)
   path("GithubInstallation", class_name: true) { "#{it.project.path}/github/#{it.ubid}" }
+  path("KubernetesNodepool", class_name: true) { "#{it.cluster.project.path}#{it.cluster.path}/nodepool/#{it.ubid}" }
```
jeremyevans: May be better to add `KubernetesNodepool#path` and instead add KubernetesNodepool to the list above.

routes/project/location/kubernetes_cluster.rb
```diff
@@ -54,6 +54,41 @@ class Clover
...
+            Validation.validate_kubernetes_worker_node_count(node_count)
+
+            DB.transaction do
+              kn.update(node_count:)
+              kn.incr_scale_worker_count
+              audit_log(kn, "update")
+            end
+
+            if api?
+              Serializers::KubernetesCluster.serialize(kc, {detailed: true})
```
jeremyevans: Seems odd to have a nodepool-specific action return information on the cluster instead of the nodepool.

routes/project/location/kubernetes_cluster.rb
aamederen: We have plans but no timeline for supporting more nodepool actions, so I'll collapse.

routes/project/location/kubernetes_cluster.rb
aamederen: This is temporary. It may make more sense to separate cluster and nodepool information when we support more nodepool actions, but for now, nodepool is not a standalone thing yet.

jeremyevans: If this is temporary, I assume that means that in the future, you plan to break backwards compatibility here?

aamederen: That's a possibility, how shall we proceed there? Maybe start treating nodepools as standalone things?

jeremyevans: I think treating nodepools as standalone objects makes sense. @fdr What are your thoughts regarding this and API backwards compatibility in general?

mohi-kalantari: I missed this while reviewing. I think it's better to treat nodepools as standalone models.

Also in the upgrades implementation, nodepools need to get `incr_uprade`d separately.

aamederen: Alright, converted to Nodepools.

We may also need to change the nodepool serializer to return `KubernetesNode`s, instead of `Vm`s.

routes/project/location/kubernetes_cluster.rb
```diff
@@ -54,6 +54,37 @@ class Clover
...
+
+        filter[:kubernetes_cluster_id] = kc.id
+        kn = @kn = kc.nodepools_dataset.first(filter)
+
+        check_found_object(kn)
+
+        authorize("KubernetesCluster:edit", kc.id)
+        handle_validation_failure("kubernetes-cluster/show") { @page = "settings" }
+        node_count = typecast_params.pos_int!("node_count")
+        Validation.validate_kubernetes_worker_node_count(node_count)
```
enescakir, line 73: We need to validate vCPUs quota too

mohi-kalantari:
The pr LGTM but I thought we can improve the action name by changing resize to scale because resize implies changing the VM size more compared to scale which is what we are doing here.

This occurred to me while looking at the example cli commands:
```ubi kc (location/kc-name | kc-id) resize-nodepool (np-name | np-id) node-count```

**Comment by aamederen** [comment]
@mohi-kalantari I think it's up to debate and both usages can be seen in the wild:

- Kubernetes docs uses "scaling" as in "autoscaler": https://kubernetes.io/docs/concepts/cluster-administration/node-autoscaling/
- Amazon EKS docs mention it as "scaling": https://docs.aws.amazon.com/eks/latest/userguide/update-managed-node-group.html
- Azure Kubernetes uses "scale" in CLI but "resize" in docs: https://learn.microsoft.com/en-us/azure/aks/resize-cluster
- Google uses "resize" in CLI and docs: https://cloud.google.com/kubernetes-engine/docs/how-to/resizing-a-cluster
- Digital ocean mentions "resize" in their docs: https://docs.digitalocean.com/products/kubernetes/getting-started/quickstart/#edit-nodes-and-node-pools-for-an-existing-cluster
- Civo uses "scaling": https://www.civo.com/docs/kubernetes/advanced/scaling-nodes

So, if the team has a strong opinion on the wording, I'm up to the change.

File: cli-commands/kc/post/resize-nodepool.rb [hunk 5]
```diff
@@ -0,0 +1,21 @@
...
+  desc "Resize a Kubernetes cluster nodepool"
+
+  banner "ubi kc (location/kc-name | kc-id) resize-nodepool (np-name | np-id) node-count"
+
+  args 2
+
+  run do |nodepool_ref, node_count, _, cmd|
+    node_count = Integer(node_count, exception: false)
+
+    unless node_count.is_a?(Integer) && node_count > 0
```
jeremyevans: `node_count` is guaranteed to be either an integer or `nil` by the previous method call.

```suggestion
    unless node_count&.>(0)
```

jeremyevans (commented):
Looks good from a Ruby perspective. I'll leave approval to @mohi-kalantari as he has actual k8s knowledge.

mohi-kalantari:
We can keep it as is then. Thanks @aamederen

PR #3929, opened 2025-09-19
Remove unused config variables

jeremyevans:
Some of these were used in the past. Some never used.

While here, make clover_column_encryption_key and clover_session_secret mandatory and not optional. You cannot load the models without clover_column_encryption_key, and you cannot load Clover without clover_session_secret.

----

> [!IMPORTANT]
> Remove unused config variables and make `clover_column_encryption_key` and `clover_session_secret` mandatory in `config.rb`.
> 
>   - **Config Changes**:
>     - Remove unused config variables: `app_name`, `versioning_default`, `versioning_app_name`, `github_cache_proxy_repo_uri`, `stripe_public_key`, and others.
>     - Make `clover_column_encryption_key` and `clover_session_secret` mandatory in `config.rb`.
>   - **Behavior**:
>     - Application will raise an exception if `clover_column_encryption_key` or `clover_session_secret` is missing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7fd43e93c3702c9d0b0b40dea81cbe2e72cb02f6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -36,15 +36,12 @@ module Config
 
   # Mandatory -- exception is raised for these variables when missing.
   mandatory :clover_database_url, string, clear: true
+  mandatory :clover_column_encryption_key, base64, clear: true
+  mandatory :clover_session_secret, base64, clear: true
   mandatory :rack_env, string
 
   # Optional -- value is returned or `nil` if it wasn't present.
-  optional :app_name, string
-  optional :versioning_default, string
-  optional :versioning_app_name, string
-  optional :clover_session_secret, base64, clear: true
   optional :clover_runtime_token_secret, base64, clear: true
-  optional :clover_column_encryption_key, base64, clear: true
   optional :heartbeat_url, string
   optional :clover_database_root_certs, string
   override :max_health_monitor_threads, 32, int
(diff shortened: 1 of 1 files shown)


jeremyevans:
I used the following code to help me find the unused variables:

```ruby
module Config
  def self.check_used(sym)
    return sym if sym.end_with?("version")
    output = `git grep 'Config.*#{sym}'`
    if output.strip.empty?
      $stderr.puts "Config.#{sym} unused"
    end
    sym
  end
  def self.optional(sym, ...)
    super(check_used(sym), ...)
  end
  def self.override(sym, ...)
    super(check_used(sym), ...)
  end
end
```

It results in some false positives, due to dynamic method calling, but I kept those after researching them.

jeremyevans:
This doesn't remove `clover_database_root_certs` or `postgres_monitor_database_root_certs`. Those are used in `db.rb`, and are used to write files, but I'm not sure if the files written are actually used. @byucesoy these was added quite a while ago in 014f3a24b6df29c98dab8b6ce4bba2abde371e6b, do you know if support was ever added to use the written files?

PR #3936, opened 2025-09-23
Fix nondeterministic instance double reuse in dispatcher spec

jeremyevans:
Clear the clover_ssh_cache thread variable after every spec.

Should fix this CI failure:

```
  1) Scheduling::Dispatcher#run_strand print exceptions if they are raised
     Failure/Error: expect(di.run_strand(st, start_queue, finish_queue)).to eq ex
       #<InstanceDouble(Net::SSH::Connection::Session) (anonymous)> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.
     # ./scheduling/dispatcher.rb:490:in 'block in Scheduling::Dispatcher#run_strand'
     # ./scheduling/dispatcher.rb:487:in 'Hash#each_value'
     # ./scheduling/dispatcher.rb:487:in 'Scheduling::Dispatcher#run_strand'
     # ./spec/scheduling/dispatcher_spec.rb:432:in 'block (3 levels) in <top (required)>'
     # ./vendor/bundle/ruby/3.4.0/gems/sequel-5.96.0/lib/sequel/database/transactions.rb:257:in 'Sequel::Database#_transaction'
     # ./vendor/bundle/ruby/3.4.0/gems/sequel-5.96.0/lib/sequel/database/transactions.rb:239:in 'block in Sequel::Database#transaction'
     # ./vendor/bundle/ruby/3.4.0/gems/sequel-5.96.0/lib/sequel/connection_pool/timed_queue.rb:90:in 'Sequel::TimedQueueConnectionPool#hold'
     # ./vendor/bundle/ruby/3.4.0/gems/sequel-5.96.0/lib/sequel/database/connecting.rb:283:in 'Sequel::Database#synchronize'
     # ./vendor/bundle/ruby/3.4.0/gems/sequel-5.96.0/lib/sequel/database/transactions.rb:197:in 'Sequel::Database#transaction'
     # ./vendor/bundle/ruby/3.4.0/gems/webmock-3.25.1/lib/webmock/rspec.rb:39:in 'block (2 levels) in <top (required)>'
```

----

> [!IMPORTANT]
> Fix nondeterministic instance double reuse in RSpec by clearing `Thread.current[:clover_ssh_cache]` after each spec in `spec_helper.rb`.
> 
>   - **Behavior**:
>     - Clears `Thread.current[:clover_ssh_cache]` after each spec in `spec_helper.rb` to prevent instance double reuse across tests.
>     - Fixes nondeterministic CI failure related to `InstanceDouble` reuse in RSpec tests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 714680440c24b843870e49eee1dc4fb9a08e01bc. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -59,6 +59,7 @@ RSpec.configure do |config|
     DB.transaction(rollback: :always, auto_savepoint: true) do
       example.run
     end
+    Thread.current[:clover_ssh_cache] = nil
     Mail::TestMailer.deliveries.clear if defined?(Mail)
 
     unless @skip_leaked_thread_check


fdr (approved):
coming back to this...so I was reading it but found the theory of design somewhat hard to penetrate. I feel like a few changed variable names or comments to help me understand the structure of interlock of the test might help.

jeremyevans:
In this case, there is a spec that does:

```
session = instance_double(Net::SSH::Connection::Session)
```

That instance_double ends up getting saved to a thread variable, and if it is later accessed in a different spec, you get the error. 

I don't think a better variable name would help in this case. The root of the problem is not in the specs, it's using thread variables in Sshable. This is just specs trying to clean up after the Sshable usage of thread variables. It is possible there are specs outside of the dispatcher specs that could cause this issue. Maybe the clearing should be moved to a higher level, such as the global after hook. However, I wanted to start with a local hook, and only move it to a higher level if the problem occurs after this change.

That's not to say the dispatcher spec variable names couldn't be improved, they probably could be, I just don't think it's related to this particular issue.

jeremyevans:
After giving this some more thought, I think I should move the clearing to the top level. The reason for  this is I've received 3 reports of this error in the last day, and nothing has changed in the dispatcher or dispatcher specs recently, so this instance double is probably getting set outside of the dispatcher spec, and only hit inside of it.  Maybe by c43d8ccd573ca5ac0d7369aaa82e549b3551e932?

PR #3938, opened 2025-09-24
Do not include server cert when providing CA data to Minio::Client

jeremyevans:
This causes problems in some environments. Only the CA certificates should be provided as CA data.

This fixes a `certificate signature failure` I experienced when setting up a MinioCluster in the staging environment.

### DIFF
--- a/model/minio/minio_server.rb
+++ b/model/minio/minio_server.rb
@@ -91,7 +91,7 @@ class MinioServer < Sequel::Model
       endpoint: server_url,
       access_key: cluster.admin_user,
       secret_key: cluster.admin_password,
-      ssl_ca_data: cluster.root_certs + cert,
+      ssl_ca_data: cluster.root_certs,
       socket: socket
     )
   end


jeremyevans:
I've added a commit to work around a bug I found with server endpoint names returned by minio admin_info returning the cluster name and not the server name, resulting in NoMethodErrors in the related progs. This may be specific to single-server minio clusters, but I'm not sure. With these two changes, I was able to setup a single server minio cluster in the staging environment.

jeremyevans:
I pushed another commit to this PR that adds validations to prevent the assembly of MinioClusters where Minio wouldn't start on the configured servers. I hit this issue when testing in the staging environment, and it took a significant amount of time to debug.

lib/validation.rb
```diff
@@ -56,6 +56,16 @@ def self.validate_minio_username(username)
     fail ValidationFailed.new({username: msg}) unless username&.match(ALLOWED_MINIO_USERNAME_PATTERN)
   end
 
+  def self.validate_minio_setup(storage_size_gib:, pool_count:, server_count:, drive_count:)
+    per_pool_server_count = server_count / pool_count
```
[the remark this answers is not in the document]
jeremyevans: Don't think that is needed. It still results in an exception being raised.

jeremyevans:
I found we were duplicating the Minio ssl_ca_data approach in VictoriaMetrics, so I added a commit that makes the same change to remove the server cert from ssl_ca_data for VictoriaMetrics. I tested in staging and VictoriaMetricsClient#health still worked, and I could still see PG metrics in the web UI.

model/minio/minio_server.rb
```diff
@@ -62,9 +62,18 @@ def init_health_monitor_session
     }
   end
 
+  def server_data(client = self.client)
+    server_data = JSON.parse(client.admin_info.body)["servers"]
+    if cluster.server_count == 1
+      server_data.first
+    else
+      server_data.find { it["endpoint"] == endpoint }
+    end
```
fdr, line 71: Is this branch necessary? It seems the number of comparisons avoided in the fast path (count == 1) is quite tiny. Is it the case that we don't want to validate endpoint equality because of the cardinality as well?

jeremyevans: Yes, it is necessary. For single server Minio clusters, `it["endpoint"] != endpoint`, as `it["endpoint"]` does not use the `index` as a suffix (unlike with multiple server Minio clusters).

fdr: Huh. Maybe it should? But perhaps not for this patch.

PR #3943, opened 2025-09-24
Session lock log prog info

fdr:
commit e4334bc6a848fab8d4e50f2a54f49dec9488aa63

    Log prog information in event of SSH session contention
    
    Although I decline to test the stack-munging code in
    `sshable_spec.rb`, the tests in prog's `base_spec.rb` included with
    that functionality might be good enough.

commit 30ad4fefe46bb135ff3845b26fa4cfaa2464527c

    Add Prog::Base.current_prog to find prog information in the stack
    
    It'd be helpful for my Sshable session conflict assessment to know
    what progs and labels are involved in contention, but as the relevant
    code is down-stack, it's not immediately obvious how to get access to
    the current prog and label information.
    
    Searching the stack is a way to do the job.

----

> [!IMPORTANT]
> Log program information during SSH session contention and add `Prog::Base.current_prog` to retrieve current program from call stack.
> 
>   - **Behavior**:
>     - Log program information during SSH session contention in `connect` method of `sshable.rb`.
>     - Add `Prog::Base.current_prog` to retrieve current program from call stack.
>   - **Tests**:
>     - Update `sshable_spec.rb` to test logging of program information during session lock failures.
>     - Add tests in `base_spec.rb` to verify `Prog::Base.current_prog` functionality.
>   - **Misc**:
>     - Allow mocking of `Prog::Base.current_prog` in `thawed_mock.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 55249b60bec18cca8fbfd3cbd82a0bad237568f3. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -9,6 +9,13 @@ class Prog::Base
     @subject_id = frame.dig("subject_id") || @strand.id
   end
 
+  # Searches the stack for the Prog that caused execution of the code,
+  # which can be useful in logging from nested method calls.
+  def self.current_prog
+    caller_locations.reverse_each { return it.label if it.label.start_with?("Prog::") }
+    nil
+  end
+
   def self.subject_is(*names)
     names.each do |name|
       class_eval %(
--- a/model/sshable.rb
+++ b/model/sshable.rb
@@ -174,7 +174,7 @@ LOCK
         end
 
         Clog.emit("session lock failure") do
-          {contended_session_lock: {exit_code:, session_fail_msg:, sshable_ubid: ubid.to_s}}
+          {contended_session_lock: {exit_code:, session_fail_msg:, sshable_ubid: ubid.to_s, prog: Prog::Base.current_prog}}
         end
       end
     end
--- a/prog/test.rb
+++ b/prog/test.rb
@@ -142,6 +142,15 @@ class Prog::Test < Prog::Base
   label def push_subject_id
     push Prog::Test, {"subject_id" => "70b633b7-1d24-4526-a47f-d2580597d53f"}
   end
+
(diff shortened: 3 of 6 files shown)


prog/base.rb
```diff
@@ -9,6 +9,13 @@ def initialize(strand, snap = nil)
     @subject_id = frame.dig("subject_id") || @strand.id
   end
 
+  # Searches the stack for the Prog that caused execution of the code,
+  # which can be useful in logging from nested method calls.
+  def self.current_prog
+    $~ = nil
+    caller.rindex { it[/Prog::(\w+)#(\w+)/] } && $&
```
jeremyevans: Giving this some more thought, I think this would be less objectionable:


```suggestion
    caller_locations.reverse_each { return it.label if it.label.start_with?("Prog::") }
    nil
```

Parsing entire backtrace lines is brittle and less efficient.  Additionally, the suggested works for progs that have `::` in them (`\w` doesn't match `:`).

fdr:
> I approve this only if it is temporary. If we want to use this long-term, we need to pass the prog information down to Sshable#connect (preferred), or at the very least, store it in a thread variable (still bad, but less bad).

Hmmm. I wasn't planning on it being temporary but I suppose we could make it temporary. One call site, I can't off the top of my head think of other times this would be super useful.

jeremyevans:
With the change to `caller_locations`, this seems less brittle and I guess it's acceptable long term.

PR #3945, opened 2025-09-24
Consider VictoriaMetricsResource in victoria_metrics_service_project_id as fallback

jeremyevans:
If there isn't a VictoriaMetricsResource in postgres_service_project_id (or PostgresServer) or monitoring_service_project_id (for VmHost), then look in victoria_metrics_service_project_id. This allows a single VictoriaMetricsResource to serve both PostgresServer and VmHost, useful for smaller installations.

While here, avoid unnecessary queries if the related project_ids are nil.

### DIFF
--- a/lib/metrics_target_resource.rb
+++ b/lib/metrics_target_resource.rb
@@ -11,9 +11,12 @@ class MetricsTargetResource
     @export_started_at = Time.now
     @deleted = false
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
@shikharbhardwaj rerequesting review because I added a commit. I found there were other places that duplicated the code I fixed in the first commit. I added methods so the logic is in one place and not in three places.

shikharbhardwaj (approved):

PR #3952, opened 2025-09-26
Alien vm improvements

enescakir:
- **Wait for runneradmin to be authenticatable for aliens**
  We set up runneradmin in the user_data script in some environments like
  AWS. So even though the SSH server is up and the VM passes wait_sshable,
  we can’t authenticate yet because cloudinit hasn’t executed the
  user_data script.
  

- **Do not use eu-central-1a az for runners**
  

- **Provision the runner again when encountering AWS insufficient capacity error**
  The AWS instance’s availability zone is determined by its subnet. We
  could recreate the subnet and retry launching the instance.
  
  For now, to keep things simple, I just reprovision the runner and
  destroy the existing one. If we start hitting capacity issues
  frequently, we can refine the logic to reprovision only the subnet
  instead of the entire runner instance.


prog/vm/github_runner.rb
```diff
@@ -269,9 +269,13 @@ def setup_info
       COMMAND
     end
 
-    # Remove comments and empty lines before sending them to the machine
-    vm.sshable.cmd(command.gsub(/^(\s*# .*)?\n/, ""))
-
+    begin
+      # Remove comments and empty lines before sending them to the machine
+      vm.sshable.cmd(command.gsub(/^(\s*# .*)?\n/, ""))
+    rescue Net::SSH::AuthenticationFailed
```
jeremyevans, line 278: This applies to both alien and non-alien runners. I guess it isn't a problem to rescue for both, but in the non-alien case, this is likely not an issue that will fix itself, right?

prog/vm/github_runner.rb
```diff
@@ -49,11 +49,13 @@ def pick_vm
     boot_image = label_data["boot_image"]
     location_id = Location::GITHUB_RUNNERS_ID
     size = label_data["vm_size"]
+    exclude_availability_zones = []
     alien_ratio = github_runner.installation.project.get_ff_aws_alien_runners_ratio || 0
     if label_data["arch"] == "x64" && rand < alien_ratio
       boot_image = Config.send(:"#{boot_image.tr("-", "_")}_aws_ami_version")
       location_id = Config.github_runner_aws_location_id
       size = Option.aws_instance_type_name("m7a", label_data["vcpus"])
+      exclude_availability_zones = ["a"] # eu-central-1a is usually give capacity errors
```
jeremyevans: Since you have an existing array:

```suggestion
      exclude_availability_zones << "a" # eu-central-1a is usually give capacity errors
```

prog/aws/instance.rb
```diff
@@ -170,6 +171,14 @@ class Prog::Aws::Instance < Prog::Base
     rescue Aws::EC2::Errors::InvalidParameterValue => e
       nap 1 if e.message.include?("Invalid IAM Instance Profile name")
       raise
+    rescue Aws::EC2::Errors::InsufficientInstanceCapacity => e
+      if is_runner && (runner = GithubRunner[vm_id: vm.id])
+        Clog.emit("insufficient instance capacity") { {insufficient_instance_capacity: {vm:, message: e.message}} }
+        runner.provision_spare_runner
+        runner.incr_destroy
+        nap 30
```
jeremyevans: Is there a reason to `nap 30` to delay the destroy? There may well be, but it would be good to have a comment here explaining why we are waiting.

lib/util.rb
```diff
@@ -70,8 +70,8 @@ def self.create_certificate(subject:, duration:, extensions: [], issuer_cert: ni
     [cert, key]
   end
 
-  def self.exception_to_hash(ex)
-    {exception: {message: ex.message, class: ex.class.to_s, backtrace: ex.backtrace, cause: ex.cause.inspect}}
+  def self.exception_to_hash(ex, backtrace: true)
+    {exception: {message: ex.message, class: ex.class.to_s, backtrace: backtrace ? ex.backtrace : nil, cause: ex.cause.inspect}}
```
jeremyevans: I think this style is simpler:

```suggestion
  def self.exception_to_hash(ex, backtrace: ex.backtrace)
    {exception: {message: ex.message, class: ex.class.to_s, backtrace:, cause: ex.cause.inspect}}
```

prog/aws/nic.rb
```diff
@@ -154,6 +154,6 @@ def ignore_invalid_nic
     Aws::EC2::Errors::InvalidAllocationIDNotFound,
     Aws::EC2::Errors::InvalidAddressIDNotFound,
     Aws::EC2::Errors::InvalidSubnetIDNotFound => e
-    Clog.emit("ID not found") { Util.exception_to_hash(e) }
+    Clog.emit("ID not found for aws nic") { {ignored_aws_nic_failure: {exception: Util.exception_to_hash(e, backtrace: false)}} }
```
jeremyevans:
```suggestion
    Clog.emit("ID not found for aws nic") { {ignored_aws_nic_failure: {exception: Util.exception_to_hash(e, backtrace: nil)}} }
```

prog/aws/vpc.rb
```diff
@@ -132,7 +132,7 @@ def ignore_invalid_id
     Aws::EC2::Errors::InvalidNetworkInterfaceIDNotFound,
     Aws::EC2::Errors::InvalidInternetGatewayIDNotFound,
     Aws::EC2::Errors::InvalidVpcIDNotFound => e
-    Clog.emit("ID not found") { Util.exception_to_hash(e) }
+    Clog.emit("ID not found for aws vpc") { {ignored_aws_vpc_failure: {exception: Util.exception_to_hash(e, backtrace: false)}} }
```
jeremyevans:
```suggestion
    Clog.emit("ID not found for aws vpc") { {ignored_aws_vpc_failure: {exception: Util.exception_to_hash(e, backtrace: nil)}} }
```

jeremyevans (approved):
Looks good. I have some recommendations, but they are up to you.

prog/vm/github_runner.rb
enescakir: Yes, we have a 10 minute deadline. If it can’t run for any reason, we’ll get a page after 10 minutes.

prog/aws/instance.rb
enescakir: It will be destroyed through runner -> vm destroy before next attempt

enescakir: I will add `destroy` semaphore to `Aws::Instance` in the next PR. We do `strand.children.select { it.prog == "Aws::Instance" }.each { it.destroy }` in the `Vm.destroy` and it causes lease violation bugs.

enescakir: Fix: https://github.com/ubicloud/ubicloud/pull/3955

PR #3953, opened 2025-09-26
Add more models to admin panel

enescakir:
- **Show ubid search form only for authenticated users**
  

- **Add VmHost table to admin panel**
  

- **Add GithubInstallation table to admin panel**
  

- **Add vm model to admin panel**
  

- **Add account model to admin panel**
  

----

> [!IMPORTANT]
> Add `VmHost`, `GithubInstallation`, `Vm`, and `Account` models to admin panel and restrict UBID search to authenticated users.
> 
>   - **Admin Panel Enhancements**:
>     - Add `VmHost`, `GithubInstallation`, `Vm`, and `Account` models to the admin panel in `clover_admin.rb`.
>     - Show UBID search form only for authenticated users in `layout.erb`.
>   - **Model Changes**:
>     - Add `provider_names` method to `Account` in `account.rb`.
>     - Add `admin_label` alias to `Address`, `AssignedHostAddress`, `AssignedVmAddress`, and `BillingRecord`.
>     - Add `sshable_host` method and `name` alias to `VmHost` in `vm_host.rb`.
>   - **Testing**:
>     - Update `admin_spec.rb` to test new models in the admin panel and authenticated UBID search.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b9e61e941e7f81455f3ef970fbd3789af5c1777a. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


enescakir:
@jeremyevans I have some questions about autoforme

1. How can I add `vm_host.sshable.host` to VmHost table?
2. VmHost search fails with HTTP 500 because I added `vm_host.ubid` to table. How can I fix it?
```
2025-09-26 22:03:14 [puma srv tp 001] Exception: ArgumentError - invalid value for Integer(): ""
    <internal:kernel>:290:in 'Kernel#Integer'
    ubicloud/ubid.rb:217:in 'UBID.from_uuidish'
    resource_methods.rb:34:in 'ResourceMethods::InstanceMethods#ubid'
    sequel/plugins/forme.rb:473:in 'Sequel::Plugins::Forme::SequelInput#standard_input'
    sequel/plugins/forme.rb:462:in 'Sequel::Plugins::Forme::SequelInput#input_other'
    sequel/plugins/forme.rb:201:in 'Sequel::Plugins::Forme::SequelInput#input'
    sequel/plugins/forme.rb:504:in 'Sequel::Plugins::Forme::InstanceMethods#forme_input'
    forme/form.rb:157:in 'block in Forme::Form#input'
    forme/form.rb:354:in 'Forme::Form#content_added'
    forme/form.rb:147:in 'Forme::Form#input'
    autoforme/action.rb:465:in 'block (3 levels) in AutoForme::Action#handle_search'
```
3. Can I add `account.identities.map(&:provider)` list to account table?
4. GitHubInstallation has JSONB column. When I search it, it doesn't do string search. How it search?
5. Can I add a column like `github_installation.runners_dataset.count` to the table?

jeremyevans:
> @jeremyevans I have some questions about autoforme

Thanks for testing it out.
 
> 1. How can I add `vm_host.sshable.host` to VmHost table?

Easiest way is is to add `VmHost#host` and use `:host` as the column. Be sure to `eager :sshable` to avoid N+1. However, this should only be a browse column, not a searchable column.  You can handle this via a proc:

```
columns do |type_symbol, request|
  cs = [...]
  cs << :host unless type_symbol == :search_form
  cs
end
```

> 2. VmHost search fails with HTTP 500 because I added `vm_host.ubid` to table. How can I fix it?
> 
> ```
> 2025-09-26 22:03:14 [puma srv tp 001] Exception: ArgumentError - invalid value for Integer(): ""
>     <internal:kernel>:290:in 'Kernel#Integer'
>     ubicloud/ubid.rb:217:in 'UBID.from_uuidish'
>     resource_methods.rb:34:in 'ResourceMethods::InstanceMethods#ubid'
>     sequel/plugins/forme.rb:473:in 'Sequel::Plugins::Forme::SequelInput#standard_input'
>     sequel/plugins/forme.rb:462:in 'Sequel::Plugins::Forme::SequelInput#input_other'
>     sequel/plugins/forme.rb:201:in 'Sequel::Plugins::Forme::SequelInput#input'
>     sequel/plugins/forme.rb:504:in 'Sequel::Plugins::Forme::InstanceMethods#forme_input'
>     forme/form.rb:157:in 'block in Forme::Form#input'
>     forme/form.rb:354:in 'Forme::Form#content_added'
>     forme/form.rb:147:in 'Forme::Form#input'
>     autoforme/action.rb:465:in 'block (3 levels) in AutoForme::Action#handle_search'
> ```

You can't search directly on the ubid, because it isn't a database column. It's not useful anyway as you can use the top level ubid search for that. You could use the same approach of not including it on the search form as mentioned above.  For Firewall, I didn't include the id or ubid, as the name column links to the related object.

> 3. Can I add `account.identities.map(&:provider)` list to account table?

Adding an `Account#provider_names` and using that is possible. You'll want to `eager :identitities` to avoid N+1.

> 4. GitHubInstallation has JSONB column. When I search it, it doesn't do string search. How it search?

Autoforme does not currently have JSONB searching support. JSONB specific support could be added. However, I think it would be better to add generic support for overriding how searching works for specific columns, which would also allow the ubid/sshable host searching to work.

> 5. Can I add a column like `github_installation.runners_dataset.count` to the table?

Adding `GithubInstallation#runner_count` would do it, but would result in N+1 queries. You could probably use a `filter` proc:

```ruby
filter do |ds, type_symbol, request|
  ds.association_left_join(:runners)
    .select_all(:github_installation)
    .select_append{count(runners[:id]).as(:runner_count)}
end
```

`GithubInstallation#runner_count` would use `self[:runner_count]` if it exists, instead of doing `runners_dataset.count`.

If you run into issues with any of the above, please let me know. It's possible I'll need to make some changes to AutoForme (in addition to the searching support discussed above).

enescakir:
Thanks a lot for the response, it was really helpful.

I added `alias_method :name, :ubid` to the VmHost model, which simplified the first commit, and I reverted the link lambda changes.

clover_admin.rb
```diff
@@ -182,22 +182,64 @@ def self.object_action(...)
     form_options(wrapper: :div)
 
     link = lambda do |obj|
+      return "nil" if obj.nil?
```
jeremyevans: Up to you, but my recommendation would be:

```suggestion
      return "" unless obj
```

model/account.rb
```diff
@@ -15,6 +15,10 @@ class Account < Sequel::Model(:accounts)
 
   alias_method :admin_label, :email
 
+  def provider_names
+    identities.map(&:provider)
```
jeremyevans: Up to you, but maybe:

```suggestion
    identities.map(&:provider).join(", ")
```

jeremyevans (approved):
This looks good. I think you should use the updated version of AutoForme (first commit in #3954), and use the new `column_search_filter` feature to ensure that all fields shown on the search form work correctly when filled out (e.g. `Vm#display_state`), or at least ensure those fields are not shown on the search form. This might also allow you to support additional columns on search form that you are currently only showing in the displayed table (e.g.`VmHost#sshable_host`).

However, this can be done later if you prefer, as the changes you already have are a significant improvement, and the admin site is only exposed to admins and not users.

enescakir:
> This looks good. I think you should use the updated version of AutoForme (first commit in #3954), and use the new `column_search_filter` feature to ensure that all fields shown on the search form work correctly when filled out (e.g. `Vm#display_state`), or at least ensure those fields are not shown on the search form. This might also allow you to support additional columns on search form that you are currently only showing in the displayed table (e.g.`VmHost#sshable_host`).
> 
> However, this can be done later if you prefer, as the changes you already have are a significant improvement, and the admin site is only exposed to admins and not users.

Thanks for the quick feature implementation! I cherry-picked it and added some custom search functionality. I cast the dates to text while searching, which allows queries by year (`2025`), month (`2025-09`), or full date (`2025-09-27`).

clover_admin.rb
```diff
@@ -182,22 +182,104 @@ def self.object_action(...)
...
+      column_search_filter do |ds, column, value|
+        if column == :created_at
+          ds.where(Sequel.cast(:created_at, :text).ilike("%#{ds.escape_like(value)}%"))
+        end
+      end
+    end
+
+    model VmHost do
+      order Sequel[:vm_host][:id]
+      eager [:sshable, :location]
```
jeremyevans: You may want to to use `eager_graph` for sshable. That way it is already joined and you don't need a separate join for the column_search_filter:

```suggestion
      eager [:location]
      eager_graph [:sshable]
```

clover_admin.rb
```diff
@@ -182,22 +182,104 @@ def self.object_action(...)
...
+      column_options sshable_host: {label: "Sshable", type: :text, value: ""},
+        allocation_state: {type: "select", options: ["accepting", "draining", "unprepared"], add_blank: true},
+        arch: {type: "select", options: ["x64", "arm64"], add_blank: true},
+        family: {type: "select", options: Option::VmFamilies.map(&:name), add_blank: true},
+        total_cores: {type: "number"},
+        total_hugepages_1g: {type: "number"}
+
+      column_search_filter do |ds, column, value|
+        if column == :sshable_host
+          ds.join(:sshable, id: :id).where(Sequel.ilike(Sequel[:sshable][:host], "%#{ds.escape_like(value)}%"))
```
```suggestion
          ds.where(Sequel.ilike(Sequel[:sshable][:host], "%#{ds.escape_like(value)}%"))
```

clover_admin.rb
```diff
@@ -182,22 +182,104 @@ def self.object_action(...)
...
+      column_search_filter do |ds, column, value|
+        if column == :premium_runner_enabled?
+          ds.where((value == "t") ?
+            Sequel.lit("allocator_preferences->'family_filter' @> ?", Sequel.pg_jsonb(["premium"])) :
+            Sequel.lit("allocator_preferences->'family_filter' IS NULL OR NOT (allocator_preferences->'family_filter' @> ?)", Sequel.pg_jsonb(["premium"])))
+        elsif column == :allocator_preferences
+          ds.where(Sequel.cast(:allocator_preferences, :text).ilike("%#{ds.escape_like(value)}%"))
+        elsif column == :created_at
+          ds.where(Sequel.cast(:created_at, :text).ilike("%#{ds.escape_like(value)}%"))
+        end
```
jeremyevans: DRYer to use a `case` statement and the Sequel pg_json_ops extension support here:

```suggestion
        case column
        when :premium_runner_enabled?
          family_filter = Sequel.pg_jsonb(:allocator_preferences).get('family_filter')
          cond = family_filter.contains(["premium"])
          if value == "t"
            ds.where(cond)
          else
            ds.where(~cond | {family_filter => nil})
          end
        when :allocator_preferences
          ds.where(Sequel.cast(:allocator_preferences, :text).ilike("%#{ds.escape_like(value)}%"))
        when :created_at
          ds.where(Sequel.cast(:created_at, :text).ilike("%#{ds.escape_like(value)}%"))
        end
```

clover_admin.rb
```diff
@@ -220,6 +221,23 @@ def self.object_action(...)
...
+      columns [:name, :display_state, :vm_host, :location, :arch, :boot_image, :family, :vcpus, :created_at]
+      column_options display_state: {type: "select", options: ["running", "creating", "starting", "rebooting", "deleting"], add_blank: true},
+        arch: {type: "select", options: ["x64", "arm64"], add_blank: true},
+        family: {type: "select", options: Option::VmFamilies.map(&:name), add_blank: true},
+        vcpus: {type: "number"},
+        created_at: {type: "text"}
+
+      column_search_filter do |ds, column, value|
+        if column == :created_at
+          ds.where(Sequel.cast(:created_at, :text).ilike("%#{ds.escape_like(value)}%"))
```
jeremyevans: Since this pattern is used multiple times, I'd add a lambda and use it in each case to DRY things up:

```ruby
column_grep = lambda do|ds, column, value|
  ds.where(Sequel.cast(column, :text).ilike("%#{ds.escape_like(value)}%"))
end
```

clover_admin.rb
```diff
@@ -220,6 +221,23 @@ def self.object_action(...)
       end
     end
 
+    model Vm do
+      order Sequel.desc(:created_at)
+      eager [:location, :vm_host]
+      columns [:name, :display_state, :vm_host, :location, :arch, :boot_image, :family, :vcpus, :created_at]
+      column_options display_state: {type: "select", options: ["running", "creating", "starting", "rebooting", "deleting"], add_blank: true},
```
jeremyevans, line 257: I think this needs special support when searching, since `display_state` is not a database column.

clover_admin.rb
```diff
@@ -182,22 +182,104 @@ def self.object_action(...)
...
         link.call(obj)
-      when :project, :location
+      when :project, :location, :vm_host
         link.call(obj.send(column))
       end
     end
 
+    model Account do
+      order Sequel.desc(Sequel[:accounts][:created_at])
+      eager [:identities]
```
jeremyevans: Better to use eager_graph here so the table is already joined:

```suggestion
      eager_graph [:identities]
```

However, note that this will result in search results using provider_names as a filter will only show the related provider_names, and not all provider names. If that is a problem, you may want to keep the current approach of using eager.

clover_admin.rb
```diff
@@ -182,22 +182,104 @@ def self.object_action(...)
...
+      columns [:name, :email, :status_id, :provider_names, :created_at, :suspended_at]
+      column_options email: {type: "text"},
+        status_id: {type: "select", options: {Unverified: 1, Verified: 2, Closed: 3}, add_blank: true},
+        provider_names: {label: "Providers", type: "select", options: ["google", "github"], add_blank: true},
+        created_at: {type: "text"},
+        suspended_at: {label: "Suspended", type: "boolean", value: nil}
+
+      column_search_filter do |ds, column, value|
+        if column == :provider_names
+          ds.join(:account_identities, account_id: :id).where(provider: value)
```
jeremyevans: Assuming you switch to `eager_graph`:

```suggestion
          ds.where(provider: value)
```

clover_admin.rb
```diff
@@ -182,22 +182,104 @@ def self.object_action(...)
...
+        elsif column == :created_at
+          ds.where(Sequel.cast(Sequel[:accounts][:created_at], :text).ilike("%#{ds.escape_like(value)}%"))
+        elsif column == :suspended_at
+          ds.send((value == "t") ? :exclude : :where, suspended_at: nil)
+        end
```
jeremyevans: May want to switch to use `case` here. No suggestion as it may conflict the above suggestion.

clover_admin.rb
enescakir: It's a column for only VM model

enescakir: Previously, I tried to refactor it to be similar to other models, but it required more work than I expected.

clover_admin.rb
enescakir: It cleaned up the code a lot. Thanks!

PR #3954, opened 2025-09-26
Support custom column search filters in AutoForme

jeremyevans:
The first commit updates autoforme to the latest commit to allow for the feature.

The second commit shows an example of using the feature. I don't think you'll want to use this commit, but you'll probably want to include the first commit in #3953 so you can get the custom searching to work. 

After you've included the first commit in #3953, you can close this.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -194,6 +194,17 @@ class CloverAdmin < Roda
       end
     end
 
+    model Account do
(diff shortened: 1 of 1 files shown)


enescakir:
I used the new version at https://github.com/ubicloud/ubicloud/pull/3953.

Thank you

PR #3955, opened 2025-09-27
Fix lease violation issues at AWS resources

enescakir:
- **Fix Vm::Nexus#destroy to not violate strand leases**
  Previously, Vm::Nexus#destroy ran:
  
  ```ruby
  strand.children.select { it.prog == "Aws::Instance" }.each { it.destroy }
  ```
  
  This is not safe as there could be an existing lease for Aws::Instance children.
  
  @jeremy fixed a similar issue at 773562ebc.
  
  Fix this by VM increments the destroy semaphore instead of destroying directly.
  Aws::Instance will pop out if the destroy semaphore is incremented.
  
  I didn’t add semaphore :destroy to Aws::Instance, since one of its
  subjects is :vm, which already has a destroy semaphore.
  

- **Fix Vnet::NicNexus#destroy to not violate strand leases**
  Previously, Vnet::NicNexus#destroy ran:
  
  ```ruby
  strand.children.select { it.prog == "Aws::Nic" }.each { it.destroy }
  ```
  
  This is not safe as there could be an existing lease for Aws::Nic
  children.
  
  @jeremy fixed a similar issue at 773562ebc.
  
  Fix this by NicNexus increments the destroy semaphore instead of destroying
  directly. Aws::Nic will pop out if the destroy semaphore is
  incremented.
  
  I didn’t add semaphore :destroy to Aws::Nic, since one of its
  subjects is :nic, which already has a destroy semaphore.
  

- **Destroy children strands only for AWS resources in Vnet::SubnetNexus#destroy**
  We started destroying children strands of Vnet::SubnetNexus in ecf059371
  to handle `bud Prog::Vnet::UpdateFirewallRules`.
  
  However, the budding of Prog::Vnet::UpdateFirewallRules was moved from
  SubnetNexus to Vm::Nexus in f3836aeab. To stay consistent with other AWS
  resources, children strands should now only be destroyed for AWS, so
  this logic moved into the if block.
  

- **Fix Vnet::SubnetNexus#destroy to not violate strand leases**
  Previously, Vnet::SubnetNexus#destroy ran:
  
  ```ruby
  strand.children.select { it.prog == "Aws::Vpc" }.each { it.destroy }
  ```
  
  This is not safe as there could be an existing lease for Aws::Vpc
  children.
  
  @jeremy fixed a similar issue at 773562ebc.
  
  Fix this by private subnet increments the destroy semaphore instead of
  destroying directly. Aws::Vpc will pop out if the destroy semaphore is
  incremented.
  
  I didn’t add semaphore :destroy to Aws::Vpc, since one of its subjects
  is :private_subnet, which already has a destroy semaphore.
  

----

> [!IMPORTANT]
> Fixes lease violation issues in AWS resources by using semaphores for destruction management in `Vm::Nexus`, `Vnet::NicNexus`, and `Vnet::SubnetNexus`.
> 
>   - **Behavior**:
>     - `Vm::Nexus#destroy`, `Vnet::NicNexus#destroy`, and `Vnet::SubnetNexus#destroy` now use semaphores to manage destruction, preventing lease violations for `Aws::Instance`, `Aws::Nic`, and `Aws::Vpc`.
>     - `Vnet::SubnetNexus#destroy` only destroys children strands for AWS resources.
>   - **Code Changes**:
>     - Added `before_run` hooks in `prog/aws/instance.rb`, `prog/aws/nic.rb`, and `prog/aws/vpc.rb` to exit early if destroy semaphore is set.
>     - Updated `destroy` methods in `prog/vm/nexus.rb`, `prog/vnet/nic_nexus.rb`, and `prog/vnet/subnet_nexus.rb` to increment destroy semaphore instead of direct destruction.
>   - **Tests**:
>     - Added tests for early exit on destroy semaphore in `instance_spec.rb`, `nic_spec.rb`, and `vpc_spec.rb`.
>     - Updated tests in `nexus_spec.rb`, `nic_nexus_spec.rb`, and `subnet_nexus_spec.rb` to verify semaphore behavior and destruction logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 4ed8de4ba12d386e8e42c39fa3489c4f09b3285c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/aws/instance.rb
+++ b/prog/aws/instance.rb
@@ -3,6 +3,12 @@
 class Prog::Aws::Instance < Prog::Base
   subject_is :vm, :aws_instance
 
+  def before_run
+    when_destroy_set? do
+      pop "exiting early due to destroy semaphore"
+    end
+  end
+
   label def start
     # Cloudwatch is not needed for runner instances
     hop_create_instance if is_runner?
@@ -178,7 +184,7 @@ class Prog::Aws::Instance < Prog::Base
         Clog.emit("insufficient instance capacity") { {insufficient_instance_capacity: {vm:, message: e.message}} }
         runner.provision_spare_runner
         runner.incr_destroy
-        nap 30 # It will be destroyed through runner -> vm destroy before next attempt
+        pop "exiting due to insufficient instance capacity"
       end
       raise
     end
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -525,7 +525,7 @@ class Prog::Vm::Nexus < Prog::Base
 
     vm.update(display_state: "deleting")
     if vm.location.aws?
-      strand.children.select { it.prog == "Aws::Instance" }.each { it.destroy }
+      Semaphore.incr(strand.children_dataset.where(prog: "Aws::Instance").select(:id), "destroy")
       bud Prog::Aws::Instance, {"subject_id" => vm.id}, :destroy
       hop_wait_aws_vm_destroyed
     end
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -191,10 +191,10 @@ class Prog::Vnet::SubnetNexus < Prog::Base
     end
 
     decr_destroy
-    strand.children.each { it.destroy }
     if private_subnet.location.aws?
       private_subnet.nics.map(&:incr_destroy)
       private_subnet.firewalls.map(&:destroy)
+      strand.children.each { it.destroy }
       bud Prog::Aws::Vpc, {"subject_id" => private_subnet.id}, :destroy
       hop_wait_aws_vpc_destroyed
     end
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -194,7 +194,7 @@ class Prog::Vnet::SubnetNexus < Prog::Base
     if private_subnet.location.aws?
       private_subnet.nics.map(&:incr_destroy)
       private_subnet.firewalls.map(&:destroy)
-      strand.children.each { it.destroy }
+      Semaphore.incr(strand.children_dataset.where(prog: "Aws::Vpc").select(:id), "destroy")
       bud Prog::Aws::Vpc, {"subject_id" => private_subnet.id}, :destroy
       hop_wait_aws_vpc_destroyed
     end
--- a/prog/aws/nic.rb
+++ b/prog/aws/nic.rb
@@ -4,6 +4,12 @@ require "aws-sdk-ec2"
 class Prog::Aws::Nic < Prog::Base
   subject_is :nic
 
+  def before_run
+    when_destroy_set? do
+      pop "exiting early due to destroy semaphore"
+    end
+  end
+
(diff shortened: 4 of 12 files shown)


prog/aws/instance.rb
```diff
@@ -176,7 +182,8 @@ class Prog::Aws::Instance < Prog::Base
         Clog.emit("insufficient instance capacity") { {insufficient_instance_capacity: {vm:, message: e.message}} }
         runner.provision_spare_runner
         runner.incr_destroy
-        nap 30 # It will be destroyed through runner -> vm destroy before next attempt
+        incr_destroy
+        nap 0
```
furkansahin: why don't we directly pop here?

prog/vm/nexus.rb
```diff
@@ -525,7 +525,7 @@ def write_params_json
 
     vm.update(display_state: "deleting")
     if vm.location.aws?
-      strand.children.select { it.prog == "Aws::Instance" }.each { it.destroy }
+      Semaphore.incr(strand.children_dataset.where(prog: "Aws::Instance").select(:id), "destroy")
```
furkansahin, line 528: We're not sure if these strands are definitely destroyed when we start the Aws::Instance prog's destroy path, right?

prog/aws/instance.rb
enescakir: Normally, we increment destroy for other progs since we need to go through the destroy logic. But yeah, since this one has a different pattern and destroy just pops it, we can do that. Let me make the change.

prog/vm/nexus.rb
enescakir: Yes. But why do we need to be sure that strands are definitely destroyed beforehand? It will just pops out in its next run.

prog/vnet/subnet_nexus.rb
```diff
@@ -191,10 +191,10 @@ def gen_reqid
     end
 
     decr_destroy
-    strand.children.each { it.destroy }
     if private_subnet.location.aws?
       private_subnet.nics.map(&:incr_destroy)
       private_subnet.firewalls.map(&:destroy)
+      strand.children.each { it.destroy }
```
jeremyevans: Are we 100% sure that none of the children could have an active lease at this point? If so, we should document in a comment why this is safe, because on its face, it appears unsafe.

jeremyevans (commented):
Overall, this looks good to me. I just had a question about one change.

prog/vnet/subnet_nexus.rb
enescakir: I changed it to `Semaphore.incr(strand.children_dataset.where(prog: "Aws::Vpc").select(:id), "destroy")` in the next commit

enescakir: I didn’t mix up the commits and kept the move in a separate one

jeremyevans: Sorry, didn't realize that when I was reviewing commit by commit.

PR #3967, opened 2025-09-30
Handle invalid email addresses when using Util.send_email 

jeremyevans:
This requires a layout change. I found that rendering a template other than `error` with the `@error` instance variable set results in missing flash messages and a messed up display.

### DIFF
--- a/spec/routes/web/auth_spec.rb
+++ b/spec/routes/web/auth_spec.rb
@@ -41,6 +41,20 @@ RSpec.describe Clover, "auth" do
     expect(page).to have_content("Name must only contain letters, numbers, spaces, and hyphens and have max length 63.")
   end
 
+  it "can not create new account with invalid email" do
+    visit "/create-account"
+    fill_in "Email Address", with: "\u1234@something.com"
+    fill_in "Full Name", with: "test"
+    fill_in "Password", with: TEST_USER_PASSWORD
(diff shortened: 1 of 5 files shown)


spec/routes/web/auth_spec.rb
```diff
@@ -41,6 +41,20 @@
     expect(page).to have_content("Name must only contain letters, numbers, spaces, and hyphens and have max length 63.")
   end
 
+  it "can not create new account with invalid email" do
+    visit "/create-account"
+    fill_in "Email Address", with: "\u1234@something.com"
+    fill_in "Full Name", with: "test"
+    fill_in "Password", with: TEST_USER_PASSWORD
+    fill_in "Password Confirmation", with: TEST_USER_PASSWORD
+    expect(EmailRenderer).to receive(:sendmail) { raise Net::SMTPSyntaxError, "501 5.1.3 Bad recipient address syntax" }
```
enescakir: We generally use `and_raise` in the codebase. Does it have a different purpose?

```suggestion
    expect(EmailRenderer).to receive(:sendmail).and_raise(Net::SMTPSyntaxError, "501 5.1.3 Bad recipient address syntax")
```

jeremyevans: No, this seems fine.

PR #3971, opened 2025-10-01
Close SSH session on event loop failure

bsatzger:
If the event loop for the pulse check fails, close the SSH session immediately. Otherwise, we run into an issue with Postgres server pulse checks. These pulse checks rely on forwarding a local socket through the SSH connection to establish a database connection. If the SSH connection breaks in certain ways, such as the peer resets the SSH connection, the pulse check that relies on the database connection will hang indefinitely. The critical step is calling `close` on the SSH object. `shutdown!` alone is not sufficient.

----

> [!IMPORTANT]
> `check_pulse` in `monitorable_resource.rb` now closes SSH session immediately on event loop failure, removing `close_resource_session`.
> 
>   - **Behavior**:
>     - In `monitorable_resource.rb`, `check_pulse` now closes the SSH session immediately if the event loop fails, preventing indefinite hanging.
>     - Removes `close_resource_session` method and integrates its functionality into `check_pulse`.
>   - **Tests**:
>     - Updates `monitorable_resource_spec.rb` to test immediate session closure on event loop failure.
>     - Removes tests for `close_resource_session` as it no longer exists.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 3a8fe5f1953aedcdb623f8b7b550568bcf65d7cf. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/monitorable_resource.rb
+++ b/lib/monitorable_resource.rb
@@ -34,6 +34,11 @@ class MonitorableResource
       rescue => ex
         event_loop_failed = true
         Clog.emit("SSH event loop has failed.") { {event_loop_failure: {ubid: @resource.ubid, exception: Util.exception_to_hash(ex)}} }
+        @session[:ssh_session].shutdown!
+        begin
+          @session[:ssh_session].close
+        rescue
+        end
       end
     end
 
@@ -66,17 +71,6 @@ class MonitorableResource
 
     run_event_loop = false
     pulse_thread&.join
-    close_resource_session if event_loop_failed
-  end
-
-  def close_resource_session
-    return if @session.nil?
-
-    @session[:ssh_session].shutdown!
-    begin
-      @session[:ssh_session].close
-    rescue
-    end
-    @session = nil
+    @session = nil if event_loop_failed
   end
 end
--- a/spec/lib/monitorable_resource_spec.rb
+++ b/spec/lib/monitorable_resource_spec.rb
@@ -49,10 +49,11 @@ RSpec.describe MonitorableResource do
     it "swallows exception and logs it if event loop fails" do
       session = {ssh_session: instance_double(Net::SSH::Connection::Session)}
       r_w_event_loop.instance_variable_set(:@session, session)
+      expect(session[:ssh_session]).to receive(:shutdown!)
+      expect(session[:ssh_session]).to receive(:close)
       expect(Thread).to receive(:new).and_call_original
       expect(session[:ssh_session]).to receive(:loop).and_raise(StandardError)
       expect(Clog).to receive(:emit).twice.and_call_original
-      expect(r_w_event_loop).to receive(:close_resource_session)
       r_w_event_loop.check_pulse
     end
 
@@ -169,24 +170,6 @@ RSpec.describe MonitorableResource do
         expect(Clog).to receive(:emit).and_call_original
         expect { r_w_event_loop.check_pulse }.not_to raise_error
       end
-    end
-  end
-
-  describe "#close_resource_session" do
-    it "returns if session is nil" do
-      session = {ssh_session: instance_double(Net::SSH::Connection::Session)}
-      expect(session[:ssh_session]).not_to receive(:shutdown!)
-      expect(session).to receive(:nil?).and_return(true)
-      r_w_event_loop.instance_variable_set(:@session, session)
-      r_w_event_loop.close_resource_session
-    end
-
-    it "shuts down and closes the session" do
-      session = {ssh_session: instance_double(Net::SSH::Connection::Session)}
-      expect(session[:ssh_session]).to receive(:shutdown!)
-      expect(session[:ssh_session]).to receive(:close)
-      r_w_event_loop.instance_variable_set(:@session, session)
-      r_w_event_loop.close_resource_session
     end
   end
 end


model/postgres/postgres_server.rb
```diff
@@ -222,10 +222,13 @@ def init_metrics_export_session
   end
 
   def check_pulse(session:, previous_pulse:)
+    last_known_lsn = nil
     reading = begin
-      session[:db_connection] ||= Sequel.connect(adapter: "postgres", host: health_monitor_socket_path, user: "postgres", connect_timeout: 4, keep_reference: false)
-      last_known_lsn = session[:db_connection]["SELECT #{lsn_function} AS lsn"].first[:lsn]
-      "up"
+      Timeout.timeout(30) do
```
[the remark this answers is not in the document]
jeremyevans: `Timeout.timeout` is in general a footgun.  If we have to do this as a temporary fix, it's OK. Might be best to add the lsn feature to rhizome using psql, and run it from rhizome with a timeout. Alternatively, drop down the the ruby-pg level, and use its async support to exit if a response is not received within the desired time.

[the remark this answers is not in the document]
bsatzger: Makes sense.

lib/monitorable_resource.rb
```diff
@@ -34,6 +34,11 @@ def check_pulse
       rescue => ex
         event_loop_failed = true
         Clog.emit("SSH event loop has failed.") { {event_loop_failure: {ubid: @resource.ubid, exception: Util.exception_to_hash(ex)}} }
+        @session[:ssh_session].shutdown!
+        begin
+          @session[:ssh_session].close
+        rescue
+        end
```
jeremyevans, line 41: Would `@resource.check_pulse` ever work with a closed SSH session? I would guess not. I don't think this causes any major problems, but it likely results in an unnecessary failure in the next `begin` and the stale retry code in the next `rescue` getting executed.

I think we should either:

* `raise` (resulting in "Monitoring job has failed" emit in addition to the "SSH event loop has failed" emit), so we create a new SSH session on the next pass
* `return` (skipping the potentially redundant "Monitoring job has failed" emit), so we create a new SSH session on the next pass
* call `@session.merge!(@resource.init_health_monitor_session)`, so we have a chance of this pass working without a forced stale retry

We should probably keep the `close_resource_session` method and call it from both the stale retry code and this code.

fdr:
I feel like this commit message could use more in it: "could hang," we know the exact cause, presuming you tested it and found the theory we worked out by video correct: `shutdown!` is not adequate to (at least) the unix socket server opened by `ssh_session.forward.local_socket(File.join(health_monitor_socket_path, ".s.PGSQL.5432"), "/var/run/postgresql/.s.PGSQL.5432")`. We need to propagate a problem on the SSH session to the unix socket state (by closing it). 

Does it affect other pulses too?  If it's only an intervention for tunneling, I think the commit leaves too much to imagination. If you identified other kinds of cleanup that are Good Things, that'd be good to note as well.

(isn't it more like "will hang?" We never update the unix listening file descriptor before right, i.e. bad behavior is guaranteed)

bsatzger:
> isn't it more like "will hang?" We never update the unix listening file descriptor before right, i.e. bad behavior is guaranteed

yeah, it will hang. we need the `close`.

> Does it affect other pulses too? If it's only an intervention for tunneling, I think the commit leaves too much to imagination

I think it only affects Postgres.

lib/monitorable_resource.rb
fdr: It hangs: the socket made in net-ssh forward.rb is none the wiser if we use the most minimal cleanup methods (which also are least likely to throw an exception...as they do less):

```ruby
        # Starts listening for connections on the local host, and forwards them
        # to the specified remote socket via the SSH connection. This will
        # (re)create the local socket file. The remote server needs to have the
        # socket file already available.
        #
        #   ssh.forward.local_socket('/tmp/local.sock', '/tmp/remote.sock')
        def local_socket(local_socket_path, remote_socket_path)
          File.delete(local_socket_path) if File.exist?(local_socket_path)
          socket = Socket.unix_server_socket(local_socket_path)

          @local_forwarded_sockets[local_socket_path] = socket

          session.listen_to(socket) do |server|
```

So libpq can rapidly obtain some kind of live connection (presuming the listen/accept loop is running) but nothing-ish ever happens, since no bytes can make it across the channel.

fdr:
It's a fair question if close is guaranteed to terminate in a timely manner, raises exceptions, etc, as it does everything "properly," but I think modulo message this is a substantial improvement from where we are. The more minimal and abrupt version is to rescue the exception and limit ourselves to closing file descriptors, that is, a fraction of the omnibus cleanup.

lib/monitorable_resource.rb
jeremyevans: After discussion, this isn't an issue, but there can be thread safety issues with this (possibly, unsure).

Long term, we should avoid event loops and move related things to rhizome.

fdr:
Digging around...from what I can tell & recall, without intervening in the main event/select loop with some kind of deadline (something @jeremyevans did for some of our other code, to be able to handle very slow commands where the TCP state is still fine to satisfy TCP keepalive, i.e. the OS will not inject an handy file descriptor error for us), `close` can block until some kind of error can propagate from the operating system through the channel abstraction.

We do have some event loop code in here (note the `Thread.new`):

```ruby

    if @resource.needs_event_loop_for_pulse_check?
      run_event_loop = true
      event_loop_failed = false
      pulse_thread = Thread.new do
        @session[:ssh_session].loop(0.01) { run_event_loop }
      rescue => ex
        event_loop_failed = true
        Clog.emit("SSH event loop has failed.") { {event_loop_failure: {ubid: @resource.ubid, exception: Util.exception_to_hash(ex)}} }
      end
    end
```

But `run_event_loop` becomes false, which is mutated finishing each pulse:

```ruby
     begin
      @pulse = @resource.check_pulse(session: @session, previous_pulse: @pulse)
      # ....etc....
     rescue; end;
    run_event_loop = false
    pulse_thread&.join
    close_resource_session if event_loop_failed
```

So there's not, to wit, any time-oriented deadline aspect here.

fdr:
Oh we never merged that, it's part of [jeremy-reduce-apoptosis](https://github.com/ubicloud/ubicloud/tree/jeremy-reduce-apoptosis), aka https://github.com/ubicloud/ubicloud/pull/2688, and...should merge that, we should have the both the server and client protect itself, that patch does the latter.

lib/monitorable_resource.rb
fdr: How can event loops be avoided? that's the only way we exchange protocol traffic after all...

fdr:
fascinating failure here: https://github.com/ubicloud/ubicloud/actions/runs/18173653071/job/51734167105

lib/monitorable_resource.rb
jeremyevans: We don't need an SSH event loop if instead of using port forwarding, we move the related code to rhizome. Then we can use standard rhizome timeout approaches.

PR #3976, opened 2025-10-01
Rely on Sshable#channel_wait for VM Nexus command timeout

fdr:
This change removes the use of the `timeout` program that previously wrapped certain `systemctl` commands, first added in commit 144c3d022bd0e3a77ca05e772aae41fadb5c35e8.

`systemctl` functions as a client to the `systemd` daemon. Terminating `systemctl` after initialization disconnects the client but does not affect ongoing operations on the server. In practice, ending `systemctl` only unblocked net-ssh and limited unwanted `apoptosis` triggers.

Two recent changes render the use of `timeout` unnecessary. On the client side, https://github.com/ubicloud/ubicloud/pull/2688 injects a fault into the net-ssh event loop, enforcing deadlines on command execution and handling network delays.

On the server side, commit 22f3aa5c1dcc7a219476af066dc849faf58d5763 sets `logind` to terminate session processes via
`KillUserProcesses=yes`. This termination uses OpenSSH `ClientAlive` checks to determine client liveness.  If the client does not return traffic promptly, it invokes `logind`'s session termination code.

Together, these client- and server-side changes ensure that slow or unresponsive commands are terminated, preventing accumulation of processes in idle sessions. All commands run under the `rhizome` user now benefit from this, making the earlier `timeout` workaround obsolete.

----

> [!IMPORTANT]
> Remove `timeout` from `systemctl` commands, using SSH channel timeouts in `Sshable` instead.
> 
>   - **Behavior**:
>     - Remove `timeout` program from `systemctl` commands in `prog/vm/nexus.rb`, using SSH channel timeouts instead.
>     - Update `cmd` method in `Sshable` to handle timeouts internally, using a `timeout` parameter.
>     - Set default timeout based on thread-local `apoptosis_at` variable in `scheduling/dispatcher.rb`.
>   - **Error Handling**:
>     - Introduce `SshTimeout` error in `sshable.rb` for handling command timeouts.
>   - **Tests**:
>     - Update tests in `spec/model/sshable_spec.rb` and `spec/prog/vm/nexus_spec.rb` to reflect removal of `timeout` and new timeout handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for fdcad62c1589a47e3206b693e3e8e0ebda839eb7. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/prog/vm/nexus_spec.rb
+++ b/spec/prog/vm/nexus_spec.rb
@@ -1025,7 +1025,7 @@ RSpec.describe Prog::Vm::Nexus do
         expect(sshable).to receive(:cmd).with("sudo systemctl stop #{nx.vm_name}", timeout: 10).and_raise(
           Sshable::SshError.new("stop", "", "Failed to stop #{nx.vm_name} Unit .* not loaded.", 1, nil)
         )
-        expect(sshable).to receive(:cmd).with(/sudo.*systemctl.*stop.*#{nx.vm_name}-dnsmasq/).and_raise(
+        expect(sshable).to receive(:cmd).with("sudo systemctl stop #{nx.vm_name}-dnsmasq").and_raise(
           Sshable::SshError.new("stop", "", "Failed to stop #{nx.vm_name} Unit .* not loaded.", 1, nil)
         )
         expect(sshable).to receive(:cmd).with(/sudo.*bin\/setup-vm delete #{nx.vm_name}/)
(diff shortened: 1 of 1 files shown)


fdr:
this is @jeremyevans's code, but I broke it out from the SSH timeout changes to mitigate risks.

When it's committed, in #2688, I intend to rewrite the commit message to refer to the commit number so I can easily chase the reference without going to web sites.

I did write an elaborate commit message how two measures provide much more rigorous safety handling across all commands, both from the client and the server.

jeremyevans (commented):
This drops the `sudo`. Was it not needed previously? Does `systemctl` work without `sudo`?

fdr:
we do need sudo.

fdr:
well, sometimes we do. I would think so. I'll check the pre-existing code

fdr:
I reprocessed this as a `git revert 144c3d022bd0e3a77ca05e772aae41fadb5c35e8` which should do the job...sending it through e2e

prog/vm/nexus.rb
```diff
@@ -532,7 +532,7 @@ def write_params_json
 
     unless host.nil?
       begin
-        host.sshable.cmd("sudo timeout 10s systemctl stop #{q_vm}")
+        host.sshable.cmd("sudo systemctl stop #{q_vm}")
```
jeremyevans: This drops the 10 second limit and uses the default limit. Is that what we want? If not:

```suggestion
        host.sshable.cmd("sudo systemctl stop #{q_vm}", timeout: 10)
```

fdr:
I reinstated the original patch and fixed the sudo. also threw in one more test-only change I noticed

PR #3980, opened 2025-10-02
Allow VmHost setup with vhost_block_backend

mohi-kalantari:
Prior to this commit, we would always setup a VmHost with SPDK and changing SPDK to VhostBlockBackend required manual creation of its prog and stopping the SPDK services to free up the hugepages.

With this commit, in a backward compatible manner, we can specify vhost_block_backend_version and also set spdk_version to nil to setup vhost instead of SPDK.

Here's a sample assemble example:

Prog::Vm::HostNexus.assemble("1.2.3.4", spdk_version: nil,
vhost_block_backend_version: Config.vhost_block_backend_version)

----

> [!IMPORTANT]
> Add support for `VhostBlockBackend` in `VmHost` setup, allowing `spdk_version` to be `nil` and ensuring backward compatibility.
> 
>   - **Behavior**:
>     - `assemble()` in `host_nexus.rb` now supports `vhost_block_backend_version` and allows `spdk_version` to be `nil` to set up `VhostBlockBackend`.
>     - Raises error if both `spdk_version` and `vhost_block_backend_version` are set.
>     - Updates `setup_storage_backend` to handle both SPDK and VhostBlockBackend setups.
>   - **Tests**:
>     - Added tests in `host_nexus_spec.rb` to verify error when both versions are set and correct setup when one is `nil`.
>     - Updated tests for `setup_storage_backend` to check for both SPDK and VhostBlockBackend setups.
>   - **Misc**:
>     - Renamed `setup_spdk` to `setup_storage_backend` in `host_nexus.rb` and `host_nexus_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for dd4f10c24a2f793f05cc44b1f75e604edadf3916. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/host_nexus.rb
+++ b/prog/vm/host_nexus.rb
@@ -3,10 +3,14 @@
 class Prog::Vm::HostNexus < Prog::Base
   subject_is :sshable, :vm_host
 
-  def self.assemble(sshable_hostname, location_id: Location::HETZNER_FSN1_ID, family: "standard", net6: nil, ndp_needed: false, provider_name: nil, server_identifier: nil, spdk_version: Config.spdk_version, default_boot_images: [])
+  def self.assemble(sshable_hostname, location_id: Location::HETZNER_FSN1_ID, family: "standard", net6: nil, ndp_needed: false, provider_name: nil, server_identifier: nil, spdk_version: Config.spdk_version, vhost_block_backend_version: nil, default_boot_images: [])
     DB.transaction do
       unless Location[location_id]
         raise "No existing Location"
+      end
+
+      if spdk_version && vhost_block_backend_version
+        raise "SPDK and VhostBlockBackend cannot be set simultaneously"
       end
 
       id = VmHost.generate_uuid
@@ -34,7 +38,7 @@ class Prog::Vm::HostNexus < Prog::Base
       Strand.create_with_id(id,
         prog: "Vm::HostNexus",
         label: "start",
-        stack: [{"spdk_version" => spdk_version, "default_boot_images" => default_boot_images}])
+        stack: [{"spdk_version" => spdk_version, "vhost_block_backend_version" => vhost_block_backend_version, "default_boot_images" => default_boot_images}])
     end
   end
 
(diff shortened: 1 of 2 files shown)


prog/vm/host_nexus.rb
```diff
@@ -3,12 +3,16 @@
...
 
-  def self.assemble(sshable_hostname, location_id: Location::HETZNER_FSN1_ID, family: "standard", net6: nil, ndp_needed: false, provider_name: nil, server_identifier: nil, spdk_version: Config.spdk_version, default_boot_images: [])
+  def self.assemble(sshable_hostname, location_id: Location::HETZNER_FSN1_ID, family: "standard", net6: nil, ndp_needed: false, provider_name: nil, server_identifier: nil, spdk_version: Config.spdk_version, vhost_block_backend_version: nil, default_boot_images: [])
     DB.transaction do
       unless Location[location_id]
         raise "No existing Location"
       end
 
+      if spdk_version && vhost_block_backend_version
+        raise "SPDK and VhostBlockBackend cannot be set simultaneously"
```
jeremyevans, line 13: I thought hosts could have some VMs running SPDK and other VMs running ubiblk. If that is correct, I don't think this case should error, and the prog should handle both being set. If incorrect, please ignore.

prog/vm/host_nexus.rb
```diff
@@ -128,25 +132,35 @@ def os_supports_slices?(os_version)
...
+    hop_setup_storage_backend if retval&.dig("msg") == "hugepages installed"
 
     push Prog::SetupHugepages
   end
 
-  label def setup_spdk
-    if retval&.dig("msg") == "SPDK was setup"
+  label def setup_storage_backend
+    return_msg = retval&.dig("msg")
+    if return_msg == "SPDK was setup"
```
jeremyevans:
```suggestion
    case retval&.dig("msg")
    when "SPDK was setup"
```

prog/vm/host_nexus.rb
```diff
@@ -128,25 +132,35 @@ def os_supports_slices?(os_version)
...
       spdk_installation = vm_host.spdk_installations.first
       spdk_cores = (spdk_installation.cpu_count * vm_host.total_cores) / vm_host.total_cpus
       vm_host.update(used_cores: spdk_cores)
 
+      hop_download_boot_images
+    elsif return_msg == "VhostBlockBackend was setup"
```
jeremyevans:
```suggestion
    when "VhostBlockBackend was setup"
```

prog/vm/host_nexus.rb
pykello: If ubiblk is enabled on a host, SPDK will not be used for new VMs. Setting both here is most likely unintentional, so I prefer keeping the error. Users who need both SPDK & ubiblk, can set SPDK up separately.

Once ubiblk is stable, SPDK should be removed anyway.

PR #3984, opened 2025-10-04
Add support for VM init scripts

jeremyevans:
Init scripts are shared across VMs. To allow for per-VM customization,
you can pass per-VM arguments to the init script.

This is supported in the web UI as well as the API/SDK/CLI.

This is based heavily on the support for registered SSH keys.

This requires changes to rhizome, so it will require a rhizome rollout
to have an effect. Old rhizome will still work, but ignore the init
script settings.

### DIFF
--- a/assets/css/tailwind.css
+++ b/assets/css/tailwind.css
@@ -46,4 +46,8 @@
   &:has(select#ssh_public_key option[value^=sk]:checked) #public-key-wrapper {
     display: none;
   }
+
+  &:has(select#init_script option[value=""]:checked) #init-script-args-wrapper {
+    display: none;
+  }
 }
--- a/helpers/vm.rb
+++ b/helpers/vm.rb
@@ -211,6 +211,8 @@ class Clover
     options.add_option(name: "unix_user")
     options.add_option(name: "ssh_public_key", values: @project.ssh_public_keys)
     options.add_option(name: "public_key")
+    options.add_option(name: "init_script", values: @project.vm_init_scripts)
+    options.add_option(name: "init_script_args")
 
     options.serialize
   end
--- a/lib/content_generator.rb
+++ b/lib/content_generator.rb
@@ -9,6 +9,7 @@ module ContentGenerator
     def self.ssh_public_key(ssh_public_key)
       ssh_public_key.name
     end
+    singleton_class.alias_method :init_script, :ssh_public_key
 
     def self.private_subnet(location, private_subnet)
       private_subnet[:display_name]
--- a/spec/routes/web/vm_spec.rb
+++ b/spec/routes/web/vm_spec.rb
@@ -215,6 +215,33 @@ RSpec.describe Clover, "vm" do
         expect(Vm.first.public_key).to eq "a a"
       end
 
+      it "can create new virtual machine using registered init script" do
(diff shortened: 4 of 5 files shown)


rhizome/host/lib/vm_setup.rb
```diff
@@ -599,6 +600,12 @@ def write_user_data(unix_user, public_keys, swap_size_bytes, boot_image)
   - [nft, add, rule, ip6, filter, output, ip6, daddr, 'fd00:0b1c:100d:5AFE::/64', meta, skuid, "!=", 0, tcp, flags, syn, reject, with, tcp, reset]
 NFT_ADD_COMMS
 
+    init_script_cmd = if init_script
+      <<INIT_SCRIPT_CMD
+  - [bash, -c, #{yq(init_script)}, #{init_script_args.to_s.shellsplit.map { |arg| yq(arg) }.unshift("bash").join(", ")}]
```
fdr, line 605: Can this be simplified and brought into line with however cloud-init implements shell execution when passed a string as an array element?

https://cloudinit.readthedocs.io/en/latest/reference/examples.html, see `ls -l /root`

jeremyevans: If we didn't support init script args, I think the simplified version (`  - #{yq(init_script)}`) would work. However, I'm not sure how you would support init script args with that approach. Do you have any ideas?

fdr: Okay, we hashed this out a bit in chat.  I also did a bit of research (mostly, refreshing my old memories on the ancient and wonderfully weird user-data support of AWS)

All things considered, I directed that we should move to a version of this that does not impose management of a separate script entity on the customer in order to use a init script (i.e. runcmd). This is implied by the implementation's choice of a project-wide stored entity, rather than a very short script passed as a literal to the VM upon provision, much as SSH keys traditionally have been. Alas, there must be revision to handle literal/immediate/inlined scripts.

The background:

While I came to understand this implementation's approach bottoms up -- why do we not use `cloud-init`'s shell environment (reason: for parameters), and then why parameters (so scripts can he held on the project and re-used when starting multiple VMs) -- I think those matters were less important than obligating more contact with Ubicloud data models (e.g. creating a script entity before using it with a VM) to use a `runcmd`.

The conceit I have about `runcmd` scripts is that: they are short, as they are a pain to debug, there is no a good error handling channel, and they are used almost exclusively by people building programs communicating via API to provision VMs. 

Because of the limitations of a single file script written in shell, they often are applied to early bootstrapping work, with code you hope to not have to debug or handle complex errors in.

As I allege they are building programs, the script (or template that generates it) tends to exist in version control, so forcing me to synchronize a version of that script as an API entity in Ubicloud feels like a pain with little gain. I can't just commit my change and run it, I have to also remember to mess around one more entity in Ubicloud (and clean up old versions, and check that I am not running an old script version when my program doesn't act right).

While there's nothing about parametric scripts that foreclose _also_ having an inlined/immediate mode per se, it seems that once we accept the storage bloating of inlined/immediate scripts (among the biggest downsides), that adding parametric/reified/organized scripts is closer to a supporting feature than a fundamental one, and can be left out until someone had an inspiration and the will to observe its use and adjust it.

jeremyevans:
Closing in favor of #4079.

PR #3988, opened 2025-10-06
Change local vetho ip4 type

furkansahin:
## Use CIDR type for local_vetho_ip

Fixes UBI-197


model/vm_host.rb
```diff
@@ -163,7 +163,7 @@ def ip4_random_vm_network
   def veth_pair_random_ip4_addr
     ips = vms_dataset.select_map(:local_vetho_ip)
     ip = APIPA_RANGE.nth(2 * SecureRandom.random_number(APIPA_LEN))
-    ip = ip.next.next while ips.include?(ip.to_s)
+    ip = ip.next.next while ips.map { it.network.to_s }.include?(ip.to_s)
```
jeremyevans: Maybe this is correct, but it looks wrong. `ips` should be an array of `NetAddr::IPv4Net`, so it shouldn't include a string:

```ruby
[NetAddr::IPv4Net.parse("169.254.0.0/32")].include?("169.254.0.0/32")
# => false
```

furkansahin: I don't understand. Before this change ips were an array of string, with the schema change, it became an array of IPv4Net. But, `map { it.network.to_s }` turns it to an array of strings, again, so that we can perform the comparison. Still, I updated the variable name and the code a bit. Does it make more sense like this?

jeremyevans: Sorry, I must have read the code wrong. New code looks good, though maybe `ips_string` should be `string_ips` or `ip_strings`.  Kind of sucks that netaddr does not properly implement equality comparisons (ipaddr gets that correct).

jeremyevans (approved):

PR #3989, opened 2025-10-06
Retry on alternative instance type for insufficient capacity errors

enescakir:
- **Add support for m6a and m7i instance types**
  We try to create alien runners with different instance types to increase
  the chances of obtaining spot instances.
  

- **Retry on alternative instance type for insufficient capacity errors**
  Spot instance requests can sometimes fail with an
  “InsufficientInstanceCapacity” error. Best practice is to request
  multiple instance types to increase the chances of obtaining a spot
  instance. Unfortunately, create_instance in E2 does not support multiple
  instance types in a single request.
  
  One recommended solution is to create an EC2 fleet with multiple
  instance types and provision the spot instance from the fleet. However,
  since we don’t know the required instance count beforehand, this would
  require creating a separate fleet for each instance. Additionally, it
  creates divergence with PG on AWS service.
  
  Instead of creating a fleet, we can handle this by retrying the
  create_instance call with alternative instance types until we try all
  whenever an “InsufficientInstanceCapacity” error occurs.


prog/aws/instance.rb
```diff
@@ -208,7 +213,7 @@ def before_run
     public_ipv6 = instance_response.dig(:network_interfaces, 0, :ipv_6_addresses, 0, :ipv_6_address)
     AssignedVmAddress.create(dst_vm_id: vm.id, ip: public_ipv4)
     vm.sshable&.update(host: public_ipv4)
-    vm.update(cores: vm.vcpus / 2, allocated_at: Time.now, ephemeral_net6: public_ipv6)
+    vm.update(cores: vm.vcpus / 2, allocated_at: Time.now, ephemeral_net6: public_ipv6, family: instance_response[:instance_type].split(".").first)
```
furkansahin: you don't need this, the rescue loop already updates the family to the chosen one.

prog/aws/instance.rb
```diff
@@ -281,6 +286,13 @@ def is_runner?
     @is_runner ||= vm.unix_user == "runneradmin"
   end
 
+  def next_family(current_family)
```
furkansahin: you don't need to pass current_family, just use `vm.family` in this method

prog/aws/instance.rb
```diff
@@ -181,7 +181,15 @@ def before_run
...
       if is_runner? && (runner = GithubRunner[vm_id: vm.id])
-        Clog.emit("insufficient instance capacity") { {insufficient_instance_capacity: {vm:, message: e.message}} }
+        families = frame["alternative_families"]
+        next_family = if families && !families.empty? && (index = families.index(vm.family) || -1) < families.length - 1
+          families[index + 1]
+        end
+        Clog.emit("insufficient instance capacity") { {insufficient_instance_capacity: {vm:, next_family:, message: e.message}} }
+        if next_family
+          vm.update(family: next_family)
+          nap 1
```
jeremyevans: Does the Amazon API react negatively to 4 API requests in quick succession? If not, maybe `nap 0` would be better here.

prog/aws/instance.rb
```diff
@@ -181,7 +181,15 @@ def before_run
       raise
     rescue Aws::EC2::Errors::InsufficientInstanceCapacity => e
       if is_runner? && (runner = GithubRunner[vm_id: vm.id])
-        Clog.emit("insufficient instance capacity") { {insufficient_instance_capacity: {vm:, message: e.message}} }
+        families = frame["alternative_families"]
+        next_family = if families && !families.empty? && (index = families.index(vm.family) || -1) < families.length - 1
+          families[index + 1]
```
jeremyevans: This logic seems more complex than is required. No need to optimize for empty families, and you don't need to check the returned index against the length of families, because if the last index is returned, then `families[index + 1]` will return `nil`:

```suggestion
        next_family = if families
          index = families.index(vm.family) || -1
          families[index + 1]
```

jeremyevans (approved):

PR #3991, opened 2025-10-06
refactoring ip4/6, ephemeral_net4/6 usages

furkansahin:


model/vm.rb
```diff
@@ -55,12 +55,8 @@ def path
     "/location/#{display_location}/vm/#{name}"
   end
 
-  def ephemeral_net4
-    assigned_vm_address&.ip&.network
-  end
-
   def ip4
```
jeremyevans, line 58: Might be worth adding `ip4_string` and `ip6_string` methods, since a lot of callers call `to_s` on the results.

scheduling/allocator.rb
```diff
@@ -261,7 +261,7 @@ def self.update_vm(vm_host, vm)
       update_args[:family] = vm_host.family if vm.family != "burstable"
       vm.update(**update_args)
       AssignedVmAddress.create(dst_vm_id: vm.id, ip: ip4.to_s, address_id: address.id) if ip4
-      vm.sshable&.update(host: vm.ephemeral_net4 || NetAddr.parse_net(vm.ephemeral_net6).nth(2))
+      vm.sshable&.update(host: vm.ip4.to_s || vm.ip6.to_s)
```
jeremyevans: This will not work as you expect, because if `vm.ip4` is `nil`, it will update the host with the empty string. You need to check whether you have an IPv4 address:

```suggestion
      vm.sshable&.update(host: vm.ip4 ? vm.ip4.to_s : vm.ip6.to_s)
```

If you add `Vm#ip4_string`, you could have it return `nil` if there is no IPv4 address, and do:


```suggestion
      vm.sshable&.update(host: vm.ip4_string || vm.ip6_string)
```

furkansahin: followed the ip4/6_string method. Thank you

model/postgres/postgres_resource.rb
```diff
@@ -60,7 +60,7 @@ def hostname
       return "#{name}.#{hostname_suffix}" if hostname_version == "v1"
       "#{name}.#{ubid}.#{hostname_suffix}"
     else
-      representative_server&.vm&.ephemeral_net4&.to_s
+      representative_server&.vm&.ip4&.to_s
```
enescakir: nit
```suggestion
      representative_server&.vm&.ip4_string
```

model/vm.rb
```diff
@@ -222,7 +226,7 @@ def params_json(swap_size_bytes: nil, ch_version: nil, firmware_version: nil, hu
     JSON.pretty_generate(
       vm_name: name,
       public_ipv6: project.get_ff_ipv6_disabled ? nic.private_subnet.random_private_ipv6.to_s : ephemeral_net6.to_s,
-      public_ipv4: ip4.to_s || "",
+      public_ipv4: ip4_string || "",
```
enescakir: nit:
You can replace it since `nil.to_s` returns `""` anyway
```suggestion
      public_ipv4: ip4.to_s,
```

prog/minio/minio_server_nexus.rb
```diff
@@ -170,7 +170,7 @@ def cluster
   label def destroy
     register_deadline(nil, 10 * 60)
     decr_destroy
-    minio_server.cluster.dns_zone&.delete_record(record_name: cluster.hostname, type: "A", data: vm.ephemeral_net4&.to_s)
+    minio_server.cluster.dns_zone&.delete_record(record_name: cluster.hostname, type: "A", data: vm.ip4&.to_s)
```
enescakir:
```suggestion
    minio_server.cluster.dns_zone&.delete_record(record_name: cluster.hostname, type: "A", data: vm.ip4_string)
```

model/vm.rb
jeremyevans: There used to be a performance difference between `nil.to_s` versus `nil || ""`, back when `nil.to_s` returned a new string each time (assuming you were using frozen string literals, where `""` would not allocate). However, since Ruby 2.7, `nil.to_s` returns an empty frozen string, so there is no reason to avoid it.

PR #3998, opened 2025-10-08
Handle empty prep.json file gracefully during VM deletion

byucesoy:
It turns out we handle missing prep.json file during delete operations, but we don't handle empty prep.json file. So we try to delete the VM, but it stucks trying to read and parse empty prep.json file. This commit adds handling of empty prep.json file.

----

> [!IMPORTANT]
> Handle empty `prep.json` file during VM deletion to prevent process from getting stuck.
> 
>   - **Behavior**:
>     - Handle empty `prep.json` file during VM deletion in `setup-vm` and `vm_setup.rb`.
>     - Prevents process from getting stuck by checking for empty file and proceeding with deletion.
>   - **Error Handling**:
>     - In `setup-vm`, catch `JSON::ParserError` alongside `Errno::ENOENT` when reading `prep.json`.
>     - In `vm_setup.rb`, add `File.empty?` check in `purge_storage()` to handle empty `prep.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 096df438ee2e894796c5a738838787fc06d5e638. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/host/bin/setup-vm
+++ b/rhizome/host/bin/setup-vm
@@ -16,7 +16,7 @@ end
 
 params = begin
   JSON.parse(File.read(VmPath.new(vm_name).prep_json))
-rescue Errno::ENOENT
+rescue Errno::ENOENT, JSON::ParserError
   if action.start_with?("delete")
     # Params were not historically needed for the delete actions,
     # so handle case where prep file has already been deleted.
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -192,8 +192,8 @@ add element inet drop_unused_ip_packets allowed_ipv4_addresses { #{ip_net} }
   end
 
   def purge_storage
-    # prep.json doesn't exist, nothing more to do
-    return if !File.exist?(vp.prep_json)
+    # prep.json doesn't exist or empty, nothing more to do
+    return if !File.exist?(vp.prep_json) || File.empty?(vp.prep_json)
 
     storage_roots = []
 


rhizome/host/lib/vm_setup.rb
```diff
@@ -192,8 +192,8 @@ def purge_user
   end
 
   def purge_storage
-    # prep.json doesn't exist, nothing more to do
-    return if !File.exist?(vp.prep_json)
+    # prep.json doesn't exist or empty, nothing more to do
+    return if !File.exist?(vp.prep_json) || File.empty?(vp.prep_json)
```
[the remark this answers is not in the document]
jeremyevans: I'm not sure why ellipsis is saying this is non-standard. `File.empty?` is a core method, just as `File.zero?` is (the methods point to the same C function). `File.empty?` certainly makes more sense naming wise than `File.zero?`, so the suggested change makes the code worse, IMO.

rhizome/host/bin/setup-vm
```diff
@@ -16,7 +16,7 @@ end
 
 params = begin
   JSON.parse(File.read(VmPath.new(vm_name).prep_json))
-rescue Errno::ENOENT
+rescue Errno::ENOENT, JSON::ParserError
```
jeremyevans, line 19: This handles general invalid JSON, instead of only empty strings. Are you sure that is what we want? If not:

```suggestion
  prep_json = File.read(VmPath.new(vm_name).prep_json)
  JSON.parse(prep_json)
rescue Errno::ENOENT, JSON::ParserError => e
  raise if e.is_a?(JSON::ParserError) && !prep_json.empty?
```

byucesoy: This part of the code, does not use the params file if the action is delete action. The check below would either continue to ignore this error (if it is delete action) or re-raise it anyway.

Contrary to this part, purge_storage actually uses prep file, and there I specifically check for emptiness. Invalid JSON would raise exception there.

jeremyevans: If the delete action does not use the params file, why are we attempting to read it or parse it in that case? Shouldn't we restructure this so if the action starts with delete, we use an empty params hash instead of trying to read and parse the file?

PR #3999, opened 2025-10-08
Add support for managing GitHub installations/repositories/cache entries to the API/SDK/CLI

jeremyevans:
This adds the following API endpoints:

```
/github
  get # list installations
/github/{installation_id}
  get # show installation details and list repositories
/github/{installation_id}/repository/{repository_id}
  get # show repository details and list cache entries
/github/{installation_id}/repository/{repository_id}/cache/{cache_entry_id}
  get # show cache entry details
  delete # remove cache entry
```

And the following CLI commands:

```
ubi gh installation list
ubi gh installation installation-name list-repositories
ubi gh installation-name/repository-name list-cache-entries
ubi gh installation-name/repository-name remove-cache-entry cache-entry-id
```

----

> [!IMPORTANT]
> Add API endpoints and CLI commands for managing GitHub installations, repositories, and cache entries, with SDK and test updates.
> 
>   - **API Endpoints**:
>     - Adds `/github` to list installations.
>     - Adds `/github/{installation_id}` to show installation details and list repositories.
>     - Adds `/github/{installation_id}/repository/{repository_id}` to show repository details and list cache entries.
>     - Adds `/github/{installation_id}/repository/{repository_id}/cache/{cache_entry_id}` to show and delete cache entry.
>   - **CLI Commands**:
>     - `ubi gh installation list` to list installations.
>     - `ubi gh installation installation-name list-repositories` to list repositories.
>     - `ubi gh installation-name/repository-name list-cache-entries` to list cache entries.
>     - `ubi gh installation-name/repository-name remove-cache-entry cache-entry-id` to remove cache entry.
>   - **SDK Updates**:
>     - Adds `GithubCacheEntry`, `GithubInstallation`, and `GithubRepository` models.
>     - Updates `Context` in `context.rb` to include new models.
>   - **Tests**:
>     - Adds tests for new CLI commands in `remove-cache-entry_spec.rb`.
>     - Updates golden files for CLI command outputs.
>     - Adds tests for API endpoints in `github_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 3c4dc13db534dd3dc84f73f29e730808a538d828. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/routes/api/cli/gh/remove-cache-entry_spec.rb
+++ b/spec/routes/api/cli/gh/remove-cache-entry_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative "../spec_helper"
+
+RSpec.describe Clover, "cli gh remove-cache-entry" do
+  it "removes cache entry" do
+    expect(Config).to receive(:github_app_name).and_return("test-app").at_least(:once)
+    gi = GithubInstallation.create_with_id("6e3ae4a8-5474-8a01-b485-3b02ac649c5f", project_id: @project.id, installation_id: 12345678, name: "test-installation-name", type: "user")
+    gp = GithubRepository.create_with_id("a58006b6-0879-8616-936a-62234e244f2f", installation_id: gi.id, name: "test-installation-name/test-repository-name")
+    ge = GithubCacheEntry.create_with_id("967f7e02-68f8-8a0e-9917-fd13d5f33501", repository_id: gp.id, key: "test-key", version: "test-version", scope: "test-scope", size: 10987654321, created_by: gp.id)
+    client = instance_double(Aws::S3::Client)
+    expect(Aws::S3::Client).to receive(:new).and_return(client)
+    expect(client).to receive(:delete_object).with(bucket: gp.bucket_name, key: ge.blob_key)
+    expect(client).to receive(:abort_multipart_upload).with(bucket: gp.bucket_name, key: ge.blob_key, upload_id: nil)
+
+    cli(%w[gh test-installation-name/test-repository-name remove-cache-entry gejszqw0k8z24k4bzt4ynyctg2])
+    expect(ge.exists?).to be false
+  end
+end
--- a/spec/routes/api/cli/golden-file-commands/success.txt
+++ b/spec/routes/api/cli/golden-file-commands/success.txt
@@ -37,6 +37,10 @@ fw list -l eu-north-h1
 fw list -l eu-north-h1 -f name,id
 fw list -l eu-north-h1 -N -f name,id
 fw list -N
+gh installation list
+gh installation test-installation-name list-repositories
+gh test-installation-name/test-repository-name list-cache-entries
+gh list-cache-entries test-installation-name/test-repository-name
 help
 help help
 help -r
--- a/cli-commands/gh.rb
+++ b/cli-commands/gh.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+UbiCli.on("gh") do
+  desc "Manage GitHub integration"
+
+  banner "ubi gh command [...]"
+  post_banner "ubi gh (installation-name/repository-name) post-command [...]"
+
+  # :nocov:
+  unless Config.production? || ENV["FORCE_AUTOLOAD"] == "1"
+    autoload_subcommand_dir("cli-commands/gh")
+    autoload_post_subcommand_dir("cli-commands/gh/post")
+  end
+  # :nocov:
+
+  args(2...)
+
+  run do |(ref, *argv), opts, command|
+    if command.post_subcommand(ref)
+      # support swapped reference and post command arguments
+      argv.insert(1, ref)
+      ref = argv.shift
+    end
+
+    installation_name, name, extra = ref.split("/", 3)
+
+    if extra || !name
+      raise Rodish::CommandFailure.new("invalid gh reference (#{ref.inspect}), should be in installation-name/repository-name format", command)
+    end
+
+    @repository = sdk.github_repository.new(installation_name:, name:)
+    command.run(self, opts, argv)
+  end
+end
+
+Unreloader.record_dependency(__FILE__, "cli-commands/gh")
--- a/cli-commands/gh/installation.rb
+++ b/cli-commands/gh/installation.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+UbiCli.on("gh", "installation") do
+  desc "Manage GitHub installations"
+
+  banner "ubi gh installation command [...]"
+  post_banner "ubi gh installation installation-name post-command [...]"
+
+  # :nocov:
+  unless Config.production? || ENV["FORCE_AUTOLOAD"] == "1"
+    autoload_subcommand_dir("cli-commands/gh/installation")
+    autoload_post_subcommand_dir("cli-commands/gh/installation/post")
+  end
+  # :nocov:
+
+  args(2...)
+
+  run do |(ref, *argv), opts, command|
+    check_no_slash(ref, "invalid GitHub installation reference (#{ref.inspect}), should not include /", command)
+    @installation = sdk.github_installation.new(ref)
+    command.run(self, opts, argv)
+  end
+end
+
+Unreloader.record_dependency(__FILE__, "cli-commands/gh/installation")
--- a/cli-commands/gh/installation/list.rb
+++ b/cli-commands/gh/installation/list.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+UbiCli.on("gh", "installation", "list") do
+  desc "List GitHub installations"
+
+  key = :github_installations_list
+
+  options("ubi gh installation list [options]", key:) do
+    on("-N", "--no-headers", "do not show headers")
+  end
+
+  run do |opts|
+    response(format_rows(%i[id name], sdk.github_installation.list, headers: opts[key][:"no-headers"] != false))
+  end
+end
--- a/cli-commands/gh/installation/post/list-repositories.rb
+++ b/cli-commands/gh/installation/post/list-repositories.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+UbiCli.on("gh", "installation").run_on("list-repositories") do
+  desc "List GitHub repositories for an installation"
+
+  key = :github_repositories_list
+
+  options("ubi gh installation installation-name list-repositories [options]", key:) do
+    on("-N", "--no-headers", "do not show headers")
+  end
+
+  run do |opts|
+    response(format_rows(%i[id name], @installation.repositories, headers: opts[key][:"no-headers"] != false))
+  end
+end
--- a/cli-commands/gh/post/list-cache-entries.rb
+++ b/cli-commands/gh/post/list-cache-entries.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+UbiCli.on("gh").run_on("list-cache-entries") do
+  desc "List cache entries for a GitHub repository"
+
+  key = :cache_entries_list
+
+  options("ubi gh installation-name/repository-name list-cache-entries [options]", key:) do
+    on("-N", "--no-headers", "do not show headers")
+  end
+
+  run do |opts|
+    cache_entries = @repository.cache_entries.each { it.values[:size] = Clover.humanize_size(it[:size]) }
+    response(format_rows(%i[id size key], cache_entries, headers: opts[key][:"no-headers"] != false))
+  end
+end
--- a/cli-commands/gh/post/remove-cache-entry.rb
+++ b/cli-commands/gh/post/remove-cache-entry.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+UbiCli.on("gh").run_on("remove-cache-entry") do
+  desc "Remove cache entry for a GitHub repository"
+
+  banner "ubi gh installation-name/repository-name remove-cache-entry cache-entry-id"
+
+  args 1
+
+  run do |id, _, cmd|
+    check_no_slash(id, "invalid cache entry id format, should not include /", cmd)
+    @repository.remove_cache_entry(id)
+    response("Cache entry, if it exists, is now scheduled for destruction")
+  end
+end
--- a/helpers/web.rb
+++ b/helpers/web.rb
@@ -205,7 +205,7 @@ class Clover < Roda
     hash
   end
 
-  def humanize_size(bytes)
+  def self.humanize_size(bytes)
     return nil if bytes.nil? || bytes.zero?
 
     units = %w[B KB MB GB]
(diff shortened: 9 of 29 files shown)


sdk/ruby/lib/ubicloud/model/github_repository.rb
```diff
@@ -0,0 +1,65 @@
...
+      {
+        cache_entries: GithubCacheEntry
+      }
+    end
+
+    singleton_class.undef_method(:create)
+    singleton_class.undef_method(:list)
+
+    # Create a new GithubRepository instance. +values+ must be a hash with
+    # :id and :installation_id keys.
```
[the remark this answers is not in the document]
jeremyevans: Fixed this and a couple similar cases.

enescakir:
I have a high-level question. Why don’t we use `/github/{installation_id}/repository` for listing repositories of an installation, instead of showing the repository list in `/github/{installation_id}`? Wouldn’t that be more RESTful?

It’s the same for caches. Why don’t we use `/github/{installation_id}/repository/{repository_id}/cache` for listing caches instead of `/github/{installation_id}/repository/{repository_id}`?

For example, if we also want to add a list of runners for a repository in the future, would that be included in the repository details endpoint as well? Showing two different long lists with pagination in a single response might make the repository detail page more complex.

jeremyevans:
> I have a high-level question. Why don’t we use `/github/{installation_id}/repository` for listing repositories of an installation, instead of showing the repository list in `/github/{installation_id}`? Wouldn’t that be more RESTful?

Perhaps, but most of the API is not designed that way. When you request information on a VM, you get the related firewalls, when you request information on a firewall, you get the related subnets, etc. If there is an objection to doing it here, we would have to change all other parts of the API if we want to be consistent.

Associations in the SDK are designed around the approach the API currently uses, so using a different approach will require adding a different type of association in the SDK.

> For example, if we also want to add a list of runners for a repository in the future, would that be included in the repository details endpoint as well? Showing two different long lists with pagination in a single response might make the repository detail page more complex.

When you request information for a private subnet in the API, you get the full list of firewalls and nics. The CLI currently includes that information in the same command output, but it doesn't have to.

Can you explain what you mean by "repository detail page"? This doesn't change the website, this only changes the API response. And for the CLI, we can have separate commands to list runners (`list-runners`) and caches (`list-cache-entries`).

Note that I'm not necessarily against the approach you are proposing, but I think we should have a consistent API, so if we go with that approach, I think we should change our entire API design.

@fdr @geemus Can you provide your thoughts on this?

enescakir:
Sorry, it should be `repository detail response`, not `page`. 

I agree that we should be consistent across endpoints.

Some repositories have thousands of cache entries and runners, while probably no VM has thousands of firewalls or subnets. I just want to understand whether our current API’s nested resource design was created with very large lists in mind, or if it was primarily intended for shorter lists.

enescakir:
@velioglu also wrote [an API design doc ](https://gist.github.com/velioglu/5787bfcb2b7702def89c748b2ce51e3b) in the early days. I’m not sure how closely we’ve followed those rules in the current API, but he might have some comments on this topic.

geemus:
My thoughts, fwiw.

There can be some big tradeoffs here. On the one hand, including associated objects in a response is often quite helpful and reduces the number of total API calls that might otherwise be needed to piece together what you want (as is the case currently). As mentioned, this usually works great with relatively low count associated resources, with the downside that it can be additional database queries and data over the wire that a user didn't want/need.

The point where it breaks down the most is when the associated resource count grows to the point that you would otherwise paginate if it were an independent list. It seems that it could become quite complicated to handle cursors for potentially several distinct associated resource types to move through the various pages (as that seems like the most consistent approach).

My own approach, historically, has been to show singular associated records, but keep associated lists as sub-resources. That forces lists into a context where there is a known pagination mechanism, should you need it, but admittedly at the cost of requiring users to potentially make several additional API calls to get everything they wanted.

When I started working with you all on this API I believe the current nesting setup was already there and I left it largely as is. I agree that ideally it should be consistent, so it may be painful to rework it from here and I'm not sure that's what you would want to do. Maybe if the argument can be made that this context is sufficiently different, however, it may warrant a different pattern. That said, if count is the only difference I think that may be harder to understand for users (who may not individually have a large count, but are forced to use a different pattern because others do).

fdr:
Tricky. Without thinking too hard about it, it seems like we might need some kind of pattern (that can be identified by the client via some kind of structural typing) to introduce a reference for lists that are large enough to benefit from pagination. Per @geemus's writing, that seems like a pretty good way to think about the dividing line. 

Unless we never have such a list large enough to require pagination, we'll have to confront this problem at some point. 
I don't have an opinion on the project management of this yet.

Here's a sketch of an approach. I'm sure it has precedents.

To simplify, if we have `{cache_entries: [{foo: "bar"}, {baz: "quux"]}`, it may also be valid to have `{cache_entries: {list: "/prefixelements/cache_entries"}}`. Thus, we'd want to reserve some kind of JSON keys/values reliably identify a "list" of those typed elements. The SDK will notice whether it gets an in-line or out-lined list and do "something" appropriate.

I see @geemus had a practice to try to get around the problem of scale-dependence on the inlined-list, by locking inlining to one-to-one relationships. That'd be hard for us to replicate at this point given path dependency, and comes with some downsides: things get ugly again if a one-to-one becomes a one-to-many or you even _think_ it's possible that could happen someday, and you may hedge your bets and out-line a list of one element. It's also a shame to give up inlining for small lists.

So, introducing an API rule -- and SDK mechanism -- that can follow a paginated link when it is encountered but can also handle inlined data in a unified-but-ideally-not-invisible way might be one way to deal with this. Or maybe it's a terrible idea. What do you think @geemus? It's basically optional, dynamic lazy loading that is under control by the *server*, it can decide whether the client gets the lazy version or the eager version.

At least one downside is people might cook up some curl script or otherwise simplified client that works great until we change the inline/outlining rules or they add too many elements. It's unfortunate but I think acceptable.

geemus:
I can imagine a possible middle ground also, which could probably be backwards compatible.

Choose some (smaller) upper bound to how many elements can be inlined, and provide a boolean that indicates whether this is a truncated list or not.

So in a sense you would get a much shorter first page of the pagination and an indication that further pages should be sought, but then you would be expected to use the normal mechanisms for listing the full version (I'm always reticent to have links to follow, maybe too much so).

I think, if you pick the right number, it could keep things manageable and probably avoid breaking this for existing endpoints. 3-5 seems good if I had to just pick numbers, but there might already be data that goes beyond this.

One thing of note though, is that this likely implies adding a number of additional list endpoints as some of them are likely missing as inline was sufficient. At least these can probably be added as-needed (you could document that absence of the boolean and boolean false for truncated results are equivalent).

Overall the goal with something like this is to balance the convenience of inline with the realities of edge cases/etc. It feels like this keeps things relatively simple while hopefully maintaining that balance. Certainly happy to discuss further and bounce more ideas.

fdr:
> I can imagine a possible middle ground also, which could probably be backwards compatible.

Hmm, did we box collections in an object? I'm pretty sure "no", unless we do something interesting (that may torture the openapi specification considerably....or not?) like have a different kind of object as an option sitting somewhere in a JSON array.

Like, given `[{foo: "bar"}, {foo: "baz}]`, I guess the only place to signal this would be like `[{foo: "bar"}, {foo: "baz"}, {more: $hyperlink, pagination_key: "someliteral"}]`. I don't have a strong preference a the moment, it's all a bit finicky, but I suspect it may become more clear once openapi or client implications show up

geemus:
Yeah, that's fair. In other contexts I've sometimes used something like a `relationships` top level key to group things, which can give greater clarity.

I would have to think harder on how to convey this here. It could just be additional keys adjacent to the existing ones, rather than grouped/nested, but that leaves something to be desired in it's own right.

fdr:
@geemus @jeremyevans I think we're at the point where the fundamental problem (pagination for many elements vs. inlining a few elements, something we already do) is clear, and there are only so many options (an inlined JSON array with a special continuation object that has URL & pagination key data, or replacing an array with a JSON object to force dereferencing out-lined data, using type data to tag the switch, or, left undiscussed, breaking backwards compatibility) that the only thing left is to explore what's easiest. I don't have a strong preference for any of them, they seem like they can all be sound approaches. So considerations like how burdensome the OpenAPI definition is or SDK programming is are all I can think of that's left, and I have no opinion there. It just seems like we have to rectify this.

One thing I am squeamish about is deferring this and making a special case. As @geemus said a few times over a long period, exceptions cost more than anyone thinks over time (where some out-of-band knowledge is necessary for particular endpoints, data types, or others). GitHub is special enough that we could consider that, but it's only a tourniquet. This problem will be back, and then we'll also have a special case in the GitHub code.

geemus:
Looking again at what we have today (focusing on VM and it's nested firewalls as an example).

We can see here that there is a `firewalls` property, which is an array of firewall schemad objects. That doesn't inherently leave space for meta data to be attached.

A couple of options here that could work, I think at least roughly in my order of preference:
1. Add something like `truncated_associations` which is an array of one or more nested resources for which truncation occurred. So if the firewall list is sufficiently short, you get it all, otherwise you get the prescribed subset and `truncated_associations` becomes `["firewalls"]`. You could also be more explicit/wordy with this being an object instead, something like `{ "firewalls": false }`, where there would always be a value and the boolean would just flip. I'm not certain about the name of this property, but I think the way it works seems reasonable and repeatable.
2. A similar approach could be taken in more of a case-by-case basis by directly adding something like `firewalls_truncated` as a boolean. Each association could then have a comparable attribute. This doesn't feel like as clean/clear of a pattern to me and seems like it might be a bit harder to implement against.

I'm sure there are many other options, but given the discussion so far, that is where my mind goes, fwiw.

fdr:
> Looking again at what we have today (focusing on VM and it's nested firewalls as an example).
> 
> We can see here that there is a `firewalls` property, which is an array of firewall schemad objects. That doesn't inherently leave space for meta data to be attached.
> 
> A couple of options here that could work, I think at least roughly in my order of preference:
> 
>     1. Add something like `truncated_associations` which is an array of one or more nested resources for which truncation occurred. So if the firewall list is sufficiently short, you get it all, otherwise you get the prescribed subset and `truncated_associations` becomes `["firewalls"]`. You could also be more explicit/wordy with this being an object instead, something like `{ "firewalls": false }`, where there would always be a value and the boolean would just flip. I'm not certain about the name of this property, but I think the way it works seems reasonable and repeatable.

I'm a bit puzzled what this looks like, both in the API response and the OpenAPI specification. I think storing the "this data is incomplete" out of line has pretty high ick factor, what redeems it? Maybe making the `firewalls: []` a union with `firewalls: {}` and using the type change to object to indicate 'more going on here' (such as pagination) would work? I imagine it could drive off-the-shelf SDK generation berserk, but we kinda agree all of those are pretty bad anyway.

geemus:
Fair. If I was doing it from scratch, probably an object that had pagination details and then an array would be higher on my list, but out of line could be less of a breaking change.

fdr:
I guess the ick factor is, suppose you wrote a `jq` type of script against the API: would you be happier with a script that worked, instead crashing because it needs a conditional to handle the object-form? (and in this case, you might grumble and wish the whole thing was an object always, though...do you want to just ignore the part where "object" means "you gotta do stuff", otherwise...wrong results again), or, would you want to operate on an arbitrary subset and do something weird once in a while? I guess I'd rather crash.

fdr:
Come of think of it, the fact that you can write a humble script against things that are inlined "nearly all the time" and it crashes only when you need more advanced API handling could be interpreted as more a feature than bug.

Whereas, if we made things always an object, you could still ignore some fields in it (such as: there are more records) and do weird bad things, just like the out-of-line version, without thinking about it, switching from the safe situation to the unsafe one with no warning.

geemus:
I think it depends a lot on the use case, unfortunately. My preferences for jq and what I would want a programmatic client to see feel different I think. For instance, in jq, I could see it being useful to have a special "continuation" object in the array or something. So you would see say the first 5 firewalls and then a continuation object, which would tip you off that there were more. Dealing with this programmatically across different languages and type systems, however, seems obnoxious. Almost begs for "interactive" vs "noninteractive" modes alla same command line tools, but that's a whole other can of worms. It depends what you want to optimize for I guess.

jeremyevans:
Since the previous discussion, we've analyzed production API usage and determined it has sufficient use that we cannot justify backwards compatibility breakage without a major benefit. Therefore, absent a major problem with existing endpoints, those should stay as they are.

In terms of this feature, it is adding new endpoints, not changing existing endpoints, so we have flexibility. We have a case here (potential numerous associated objects) that tends to be different than existing cases (generally few associated objects), so treating it differently seems reasonable.

I don't think this new case should be considered as an exception, because going forward, we can continue to use this approach (inline for few associated objects, separate endpoint for expected numerous associated objects), or we can switch to an approach where separate endpoints are always used (and the existing endpoints are the exceptions, for backwards compatibility).

Considering that, I will implement @enescakir's suggestion and use a separate endpoint for the associated objects (repositories for installations, cache entries for repositories), instead of inlining the associated objects.

jeremyevans:
@enescakir @fdr Changes have been made, now ready for your review.

spec/routes/api/cli/gh/remove-cache-entry_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require_relative "../spec_helper"
+
+RSpec.describe Clover, "cli gh remove-cache-entry" do
+  it "removes cache entry" do
+    expect(Config).to receive(:github_app_name).and_return("test-app").at_least(:once)
+    gi = GithubInstallation.create_with_id("6e3ae4a8-5474-8a01-b485-3b02ac649c5f", project_id: @project.id, installation_id: 12345678, name: "test-installation-name", type: "user")
+    gp = GithubRepository.create_with_id("a58006b6-0879-8616-936a-62234e244f2f", installation_id: gi.id, name: "test-installation-name/test-repository-name")
+    ge = GithubCacheEntry.create_with_id("967f7e02-68f8-8a0e-9917-fd13d5f33501", repository_id: gp.id, key: "test-key", version: "test-version", scope: "test-scope", size: 10987654321, created_by: gp.id)
+    client = instance_double(Aws::S3::Client)
```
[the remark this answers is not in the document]
jeremyevans: Not really worth it for this spec.

spec/routes/api/cli/golden-file-commands/success.txt
```diff
@@ -37,6 +37,10 @@ fw list -l eu-north-h1
 fw list -l eu-north-h1 -f name,id
 fw list -l eu-north-h1 -N -f name,id
 fw list -N
+gh installation list
+gh installation test-installation-name list-repositories
+gh test-installation-name/test-repository-name list-cache-entries
+gh list-cache-entries test-installation-name/test-repository-name
```
enescakir, line 43: Why don't we have `remove-cache-entry` success file?

enescakir, line 43: Do we have two ways to use `list-cache-entries`? Based on help output, `list-cache-entries` is a post-command.

jeremyevans: Not all commands use golden files. `remove-cache-entry` has a separate spec that tests for output in addition to behavior.

jeremyevans: It is a post command, but just like post commands for the location-based objects, since it is unambiguous (as no post commands will include a slash), we allow reversing the order.

jeremyevans:
Revised the commit message for the fourth commit.

PR #4003, opened 2025-10-09
Spill over alien runners

enescakir:
- **Create strand for github runner prog tests**
  It's better to keep our tests close to real workloads. When the strand
  isn't persisted in the database, `Semaphore.incr` becomes a no-op and
  doesn't behave as expected.
  

- **Add x64? helper to github runner prog**
  
- **Remove runner_capacity_waited log**
  We have already logg the runner_allocated duration in the next label.
  These values are very close to each other, so there is no need to log
  both.

  I will refactor this label in the next commits

- **Decrease indentation at runner wait_concurrency_limit**

- **Refactor runner wait_concurrency_limit**

- **Spill over to alien runners when concurrent limit is reached**
  Previously, we tested the alien runner feature for selected projects
  with a given ratio.
  
  Now, we start spilling over to alien runners when the concurrent limit
  is reached for selected customers.
  
  If the feature flag is enabled, a job gets an alien runner instead of
  waiting in the queue.
  

- **Wait some time before spilling over to alien runners**
  According to last week’s data, waiting 10s before spilling over to alien
  runners decreases the spill rate by 2x, and waiting 30s decreases it by
  4x.
  
  For now, we’ve started with a 30s wait. We plan to experiment with 10s
  in the future. This approach allows us to start with lower load and
  increase it later if needed.
  

- **Add spilled over vCPU capacity limit**

    Currently, we spill over runners to AWS EC2 instances, which naturally
    have a limit on available vCPUs.

    We shouldn't spill over more runners than the available vCPU capacity.

    This check isn't fully concurrency-safe, but it's sufficient for now
    since we don't expect too many runners to spill over simultaneously.

    If we hit the limit frequently, we can consider increasing the vCPU
    capacity on the AWS side.
  

----

> [!IMPORTANT]
> Add spill over to alien runners when concurrency limit is reached, with wait time and vCPU capacity limit, and refactor related logic.
> 
>   - **Behavior**:
>     - Spill over to alien runners when concurrency limit is reached in `github_runner.rb` and `github_runner_spec.rb`.
>     - Introduce a 30s wait before spilling over, configurable via `github_runner_aws_spill_threshold_seconds` in `config.rb`.
>     - Limit spill over by vCPU capacity, configurable via `github_runner_aws_spill_vcpu_capacity` in `config.rb`.
>   - **Refactoring**:
>     - Refactor `wait_concurrency_limit` in `github_runner.rb` to handle spill over logic.
>     - Decrease indentation in `wait_concurrency_limit` in `github_runner.rb`.
>   - **Logging**:
>     - Remove `runner_capacity_waited` log in `github_runner.rb`.
>     - Add logging for spill over events in `github_runner.rb`.
>   - **Helpers**:
>     - Add `x64?` helper method in `github_runner.rb`.
>   - **Testing**:
>     - Update tests in `github_runner_spec.rb` to cover new spill over behavior and configurations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for ff3815cc7dab266a58382ecb467922d91335f1ed. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -104,6 +104,7 @@ module Config
   override :github_runner_aws_spot_instance_enabled, false, bool
   optional :github_runner_aws_spot_instance_max_price_per_vcpu, float
   override :github_runner_aws_spill_threshold_seconds, 30, int
+  override :github_runner_aws_spill_vcpu_capacity, 100, int
 
   # GitHub Cache
   optional :github_cache_blob_storage_endpoint, string
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -212,6 +212,12 @@ class Prog::Vm::GithubRunner < Prog::Base
         Time.now - github_runner.created_at > Config.github_runner_aws_spill_threshold_seconds
 
       if should_spill_over
+        spilled_vcpus = Vm.where(boot_image: [Config.github_ubuntu_2204_aws_ami_version, Config.github_ubuntu_2404_aws_ami_version]).sum(:vcpus) || 0
+        if spilled_vcpus >= Config.github_runner_aws_spill_vcpu_capacity
+          Clog.emit("not allowed because of high utilization and spill capacity exceeded") { {exceeded_spill_capacity: {family_utilization:, spilled_vcpus:, label: github_runner.label, repository_name: github_runner.repository_name}} }
+          nap rand(5..15)
+        end
+
         Clog.emit("spilled over runner") { {spilled_over_runner: {family_utilization:, label: github_runner.label, repository_name: github_runner.repository_name}} }
         github_runner.incr_spill_over
         hop_allocate_vm
--- a/spec/prog/vm/github_runner_spec.rb
+++ b/spec/prog/vm/github_runner_spec.rb
@@ -351,6 +351,17 @@ RSpec.describe Prog::Vm::GithubRunner do
         expect(runner.spill_over_set?).to be(false)
       end
 
+      it "waits if utilization is high, spill over enabled, and waited enough but spill vcpus limit exceeded" do
+        expect(project).to receive(:quota_available?).with("GithubRunnerVCpu", 0).and_return(false)
+        project.set_ff_spill_to_alien_runners(true)
+        runner.update(created_at: now - 40)
+        expect(Config).to receive(:github_runner_aws_spill_vcpu_capacity).and_return(10)
+        create_vm(vcpus: 16, boot_image: Config.github_ubuntu_2204_aws_ami_version)
+
+        expect { nx.wait_concurrency_limit }.to nap
+        expect(runner.spill_over_set?).to be(false)
+      end
+
       it "allocates if utilization is high but spill over enabled and waited enough" do
         expect(project).to receive(:quota_available?).with("GithubRunnerVCpu", 0).and_return(false)
         project.set_ff_spill_to_alien_runners(true)


spec/prog/vm/github_runner_spec.rb
```diff
@@ -6,9 +6,7 @@
 
 RSpec.describe Prog::Vm::GithubRunner do
   subject(:nx) {
-    described_class.new(Strand.new).tap {
```
enescakir, line 9: In the first commit, I initially tried the following:

```ruby
described_class.new(Strand.create_with_id(runner.id, prog: "Vm::GithubRunner", label: "start")
).tap {
  it.instance_variable_set(:@github_runner, runner)
}
```

Even though `Strand.all` showed the strand, `runner.strand` was returning nil.

So I changed it to the current approach. @jeremyevans , do you think the current one is okay, or should I spend more time figuring out why the first approach didn’t work?

jeremyevans: `runner.strand` would return `nil` even if the strand existed, if it was called before the strand existed (it caches the negative lookup). I'm not sure whether that happened, but do you think it is a possibility?

config.rb
```diff
@@ -103,6 +103,8 @@ def self.test?
   optional :github_runner_aws_location_id, string
   override :github_runner_aws_spot_instance_enabled, false, bool
   optional :github_runner_aws_spot_instance_max_price_per_vcpu, float
+  override :github_runner_aws_spill_threshold, 30, int
```
velioglu: At the first glance, it is not easy to understand that threshold is a time threshold in second. Can you please update the name of it?

prog/vm/github_runner.rb
```diff
@@ -189,21 +194,33 @@ def quota_available?
...
-      end
-
-      if github_runner.label_data["arch"] == "x64" && ((family_utilization["premium"] > 75) || (github_runner.installation.free_runner_upgrade? && family_utilization["premium"] > 50))
-        github_runner.incr_not_upgrade_premium
+        if x64? &&
+            github_runner.installation.project.get_ff_spill_to_alien_runners &&
+            Time.now - github_runner.created_at > Config.github_runner_aws_spill_threshold
+          spilled_vcpus = Vm.where(boot_image: [Config.github_ubuntu_2204_aws_ami_version, Config.github_ubuntu_2404_aws_ami_version]).sum(:vcpus) || 0
+          if spilled_vcpus >= Config.github_runner_aws_spill_vcpu_capacity
+            Clog.emit("not allowed because of high utilization and spill capacity exceeded") { {exceeded_spill_capacity: {family_utilization:, spilled_vcpus:, label: github_runner.label, repository_name: github_runner.repository_name}} }
```
velioglu: Shouldn't we need to nap afterwards? Also, I find 4 levels of nested conditional code confusing. Allowing spilled runners via not_allow condition also looks confusing. We can early return if quota is available and also have spilling logic on its own function.

prog/vm/github_runner.rb
```diff
@@ -182,30 +187,48 @@ def quota_available?
...
+    if is_high_util
+      should_spill_over = x64? &&
+        github_runner.installation.project.get_ff_spill_to_alien_runners &&
+        Time.now - github_runner.created_at > Config.github_runner_aws_spill_threshold_seconds
+
+      if should_spill_over
+        spilled_vcpus = Vm.where(boot_image: [Config.github_ubuntu_2204_aws_ami_version, Config.github_ubuntu_2404_aws_ami_version]).sum(:vcpus) || 0
+        if spilled_vcpus >= Config.github_runner_aws_spill_vcpu_capacity
+          Clog.emit("not allowed because of high utilization and spill capacity exceeded") { {exceeded_spill_capacity: {family_utilization:, spilled_vcpus:, label: github_runner.label, repository_name: github_runner.repository_name}} }
+          nap rand(5..15)
```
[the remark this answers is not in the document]
enescakir: nap exits

prog/vm/github_runner.rb
enescakir: I refactored it

spec/prog/vm/github_runner_spec.rb
enescakir: Probably it's the case

prog/vm/github_runner.rb
```diff
@@ -26,13 +26,14 @@ def self.assemble(installation, repository_name:, label:, default_branch: nil)
...
     installation = github_runner.installation
+    skip_pool = installation.project.get_ff_skip_runner_pool || github_runner.spill_over_set?
 
     vm_size = if installation.premium_runner_enabled? || installation.free_runner_upgrade?
       "premium-#{label_data["vcpus"]}"
     else
       label_data["vm_size"]
     end
-    pool = installation.project.get_ff_skip_runner_pool ? nil : VmPool.where(
+    pool = skip_pool ? nil : VmPool.where(
```
jeremyevans: Up to you, but using ternary operator to set `nil` seems unidiomatic. I generally prefer a conditional for this as I think it makes the code clearer:

```ruby
  unless skip_pool
    pool = VmPool.where(...
```

prog/vm/github_runner.rb
```diff
@@ -182,30 +189,48 @@ def quota_available?
...
+      should_spill_over = x64? &&
+        github_runner.installation.project.get_ff_spill_to_alien_runners &&
+        Time.now - github_runner.created_at > Config.github_runner_aws_spill_threshold_seconds
+
+      if should_spill_over
+        spilled_vcpus = Vm.where(boot_image: [Config.github_ubuntu_2204_aws_ami_version, Config.github_ubuntu_2404_aws_ami_version]).sum(:vcpus) || 0
+        if spilled_vcpus >= Config.github_runner_aws_spill_vcpu_capacity
+          Clog.emit("not allowed because of high utilization and spill capacity exceeded") { {exceeded_spill_capacity: {family_utilization:, spilled_vcpus:, label: github_runner.label, repository_name: github_runner.repository_name}} }
+          nap rand(5..15)
+        end
```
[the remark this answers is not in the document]
enescakir: nap exits

PR #4015, opened 2025-10-10
Fix busy wait for parent strand when child strand naps in respirate

jeremyevans:
The parent would nap 0 in this case. The nap 0 makes sense for exit or hop, since that is a state change, so immediately retrying makes sense. However, it does not make sense for nap, since the state did not change.

For a single child, have the parent nap using slightly less time than the child nap, so the parent runs the child.

For multiple children, nap until slightly before the earliest child without an active lease is scheduled.

In all cases where the child naps, do not nap less than 0.1 seconds or more than 120 seconds.

----

> [!IMPORTANT]
> Fixes busy-wait in `Prog::Base` by adjusting parent nap time based on child nap time, with tests verifying behavior.
> 
>   - **Behavior**:
>     - Fixes busy-wait issue in `reap` method of `Prog::Base` when child strands nap.
>     - Parent naps slightly less than child nap time for single child, or until slightly before the earliest scheduled child for multiple children.
>     - Ensures nap time is between 0.1 and 120 seconds.
>   - **Tests**:
>     - Adds tests in `base_spec.rb` to verify parent nap behavior for single and multiple child strands.
>     - Tests cover scenarios for nap times exceeding 120 seconds and multiple children with different schedules.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0a1fb19ab96d0ff54bd2b410b30f222c6a0df500. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -240,7 +240,34 @@ end
         nap(nap)
       else
         active_children.each do |child|
-          nap 0 if child.run
+          if (result = child.run)
+            if result.is_a?(Nap)
+              seconds = if active_children.length == 1
+                # For a single active child napping, parent can nap for as long as the child naps,
+                # since the expectation is there will not be anything to do until then.
+                result.seconds
+              else
+                # For multiple active children, if a single child is napping, it's possible the
+                # other children are immediately runnable. However, you don't want to busy
+                # wait on multiple children. Nap until the time of the earliest scheduled child
+                # that isn't currently running. If all children are running, nap for 120 seconds
+                strand.children_dataset
+                  .where(Sequel[:lease] < Sequel::CURRENT_TIMESTAMP)
+                  .min(Sequel.extract(:epoch, Sequel[:schedule] - Sequel::CURRENT_TIMESTAMP)) || 121
+              end
+
+              # Remove a 10th of a second so it is likely the parent will run the child.
+              seconds -= 0.1
+
+              # Nap for a minimum of 0.1 seconds and a maximum of 120 seconds in any case.
+              # The 0.1 seconds is to avoid busy waiting.
+              nap(seconds.clamp(0.1, 120))
+            else
+              # A non-nap (e.g. Exit or Hop) happened, so the state changed, and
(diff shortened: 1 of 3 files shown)


prog/base.rb
```diff
@@ -240,7 +240,34 @@ def reap(hop = nil, reaper: nil, nap: nil, fallthrough: false)
         nap(nap)
       else
         active_children.each do |child|
-          nap 0 if child.run
+          if (result = child.run)
+            if result.is_a?(Nap)
```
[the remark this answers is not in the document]
jeremyevans: Adding helper methods that are only called in one place does not necessarily improve code clarity, as it reduces locality.

prog/base.rb
```diff
@@ -240,7 +240,34 @@ def reap(hop = nil, reaper: nil, nap: nil, fallthrough: false)
...
+              seconds = if active_children.length == 1
+                # For a single active child napping, parent can nap for as long as the child naps,
+                # since the expectation is there will not be anything to do until then.
+                result.seconds
+              else
+                # For multiple active children, if a single child is napping, it's possible the
+                # other children are immediately runnable. However, you don't want to busy
+                # wait on multiple children. Nap until the time of the earliest scheduled child
+                # that isn't currently running. If all children are running, nap for 120 seconds
```
bsatzger, line 253: Napping for 120 seconds when all children are running seems relatively long. If all children finish quickly, would we have another mechanisms for the parent to wake up or would it finish the 120 sec nap in any case?

jeremyevans: When a child exits, if it is the last active child of the parent, it immediately schedules the parent: https://github.com/ubicloud/ubicloud/blob/37ed38f3d3828393baa56aec8fc3353f91e48c47/model/strand.rb#L121-L138

enescakir (approved):

PR #4027, opened 2025-10-14
Validate string quotes for PG config values

shikharbhardwaj:
> [!IMPORTANT]
> Adds validation for proper quoting of PostgreSQL config strings, ensuring special characters are handled correctly.
> 
>   - **Validation**:
>     - Adds `validate_string_quoting` in `postgres_config_validator.rb` to ensure strings with special characters are properly quoted.
>     - Updates `validate_string` to call `validate_string_quoting` for additional checks.
>   - **Tests**:
>     - Adds tests in `postgres_config_validator_spec.rb` for various string quoting scenarios, including unquoted strings with spaces, commas, and special characters, and quoted strings with unescaped quotes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 1b8d1da73a25b2cd661543ab8213f15455f68ae7. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/validation/postgres_config_validator.rb
+++ b/lib/validation/postgres_config_validator.rb
@@ -100,8 +100,34 @@ module Validation
     end
 
     def validate_string(value, pattern)
-      return nil if pattern.nil? || value.match?(Regexp.new(pattern))
-      "must match pattern: #{pattern}"
+      # Check pattern validation first if pattern is specified
+      if pattern && !value.match?(Regexp.new(pattern))
+        return "must match pattern: #{pattern}"
+      end
+
+      # Check if string value needs quoting and is properly quoted
+      quoting_error = validate_string_quoting(value)
+      return quoting_error if quoting_error
+
+      nil
+    end
+
+    def validate_string_quoting(value)
+      # Ref: https://www.postgresql.org/docs/9.3/config-setting.html#CONFIG-SETTING-CONFIGURATION-FILE
+      is_quoted = value.start_with?("'") && value.end_with?("'")
+
+      if is_quoted
+        inner_value = value[1...-1]
+
+        # Single quotes inside should be escaped as '' or '\
+        escaped_quotes_removed = inner_value.gsub("''", "").gsub("\\'", "")
+
+        if escaped_quotes_removed.include?("'")
+          "contains unescaped single quotes - single quotes must be doubled ('') inside quoted strings"
+        end
+      elsif !value.match?(/\A[a-zA-Z0-9_]+\z/)
+        "must be wrapped in single quotes (e.g., '#{value.gsub("'", "''")}') because it contains special characters"
+      end
     end
 
     def validate_bool(value)
--- a/spec/lib/validation/postgres_config_validator_spec.rb
+++ b/spec/lib/validation/postgres_config_validator_spec.rb
@@ -40,8 +40,33 @@ RSpec.describe Validation::PostgresConfigValidator do
         expect { validator.validate(config) }.not_to raise_error
       end
 
-      it "returns no errors for setting a string with no validation" do
-        config = {"archive_command" => "foo bar"}
+      it "returns no errors for properly quoted string with spaces" do
+        config = {"archive_command" => "'foo bar'"}
+        expect { validator.validate(config) }.not_to raise_error
+      end
+
+      it "returns no errors for simple string without special characters" do
+        config = {"application_name" => "my_app0001"}
+        expect { validator.validate(config) }.not_to raise_error
+      end
+
+      it "returns no errors for string with path characters" do
+        config = {"ssl_ca_file" => "'/etc/ssl/certs/ca.crt'"}
+        expect { validator.validate(config) }.not_to raise_error
+      end
+
+      it "returns no errors for properly quoted string with comma" do
+        config = {"shared_preload_libraries" => "'pg_cron,pg_stat_statements'"}
+        expect { validator.validate(config) }.not_to raise_error
+      end
+
+      it "returns no errors for properly quoted string with two escaped quotes" do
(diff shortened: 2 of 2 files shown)


lib/validation/postgres_config_validator.rb
```diff
@@ -100,8 +100,38 @@ def validate_enum(value, allowed_values)
     end
 
     def validate_string(value, pattern)
-      return nil if pattern.nil? || value.match?(Regexp.new(pattern))
-      "must match pattern: #{pattern}"
+      # Check pattern validation first if pattern is specified
+      if pattern && !value.match?(Regexp.new(pattern))
```
jeremyevans, line 104: Not an issue with this PR, but why do we store patterns as strings? They aren't stored in a Ruby file (postgres_config_validator_schema.rb), not in another format, they should literal regexps in that file. Additionally, I'm guessing use of `^` and `$` in the patterns is wrong, in Ruby, `\A` and `\z` should be used. You should only use `^` and `$` in Ruby if you are absolutely sure you want to allow matching the start or end of any line inside the string. If you want to match the beginning and end of the string, you should always use `\A` and `\z`.

lib/validation/postgres_config_validator.rb
```diff
@@ -100,8 +100,38 @@ def validate_enum(value, allowed_values)
...
+        inner_value = value[1...-1]
+
+        # Single quotes inside should be escaped as '' or '\
+        escaped_quotes_removed = inner_value.gsub("''", "").gsub("\\'", "")
+
+        if escaped_quotes_removed.include?("'")
+          return "contains unescaped single quotes - single quotes must be doubled ('') inside quoted strings"
+        end
+      else
+        unless value.match?(/^[a-zA-Z0-9]+$/)
```
jeremyevans: Should definetely use `\A` and `\z` instead of `^` and `$`.

Probably want to combine `else`/`unless` into `elsif !value.match?(/\A[a-zA-Z0-9]+\z/)`.

Should this include `_` as an allowed character?

lib/validation/postgres_config_validator.rb
```diff
@@ -100,8 +100,38 @@ def validate_enum(value, allowed_values)
...
+
+    def validate_string_quoting(value)
+      # Ref: https://www.postgresql.org/docs/9.3/config-setting.html#CONFIG-SETTING-CONFIGURATION-FILE
+      is_quoted = value.start_with?("'") && value.end_with?("'")
+
+      if is_quoted
+        inner_value = value[1...-1]
+
+        # Single quotes inside should be escaped as '' or '\
+        escaped_quotes_removed = inner_value.gsub("''", "").gsub("\\'", "")
```
jeremyevans, line 123: Is removal of `\'` correct? By default, Postgres does not recognize `\'` as an escaped `'`:

```
$ echo "SELECT '1'''" | psql -U postgres
 ?column? 
----------
 1'
(1 row)

$ echo "SELECT '1\''" | psql -U postgres
ERROR:  unterminated quoted string at or near "'1\''"
LINE 1: SELECT '1\''
```

Does Postgres have different parsing rules for these configuration settings than it does for parsing literal strings in SQL? (maybe it does, I'm not sure)

The `\'` case doesn't appear to be tested. If you are sure it is needed, it should be tested.

lib/validation/postgres_config_validator.rb
```diff
@@ -100,8 +100,38 @@ def validate_enum(value, allowed_values)
...
+        if escaped_quotes_removed.include?("'")
+          return "contains unescaped single quotes - single quotes must be doubled ('') inside quoted strings"
+        end
+      else
+        unless value.match?(/^[a-zA-Z0-9]+$/)
+          return "must be wrapped in single quotes (e.g., '#{value.gsub("'", "''")}') because it contains special characters"
+        end
+      end
+
+      nil
```
jeremyevans: Don't need explicit `nil` here, implicit `nil` should be fine.

```suggestion
```

lib/validation/postgres_config_validator.rb
shikharbhardwaj: It looks to be the case according to the docs. They say two quotes are preferred but backslash quotes are acceptable.

> To embed a single quote in a parameter value, write either two quotes (preferred) or backslash-quote.

I tried the following config and it worked:

```
[76] clover-development(main)> pg.representative_server.vm.sshable.cmd("cat /etc/postgresql/16/main/conf.d/099-user.conf")
cluster_name = 'my\'app'
default_toast_compression = pglz{"ssh":{"ubid":"vmx4rwtgpw621pz36ksw9nhkhk","start":"2025-10-15 00:02:48 +0200","finish":"2025-10-15 00:02:49 +0200","timeout":81,"duration":0.643556,"cmd":"cat /etc/postgresql/16/main/conf.d/099-user.conf","exit_code":0,"exit_signal":null,"channel_duration":0.621379,"connect_duration":0.2972},"message":"ssh cmd execution","time":"2025-10-15 00:02:49 +0200"}
[77] clover-development(main)> pg.representative_server.run_query('SHOW cluster_name');
my'app
```

lib/validation/postgres_config_validator.rb
shikharbhardwaj: :ack: I am working on another change related to the schema, will update the patterns there.

lib/validation/postgres_config_validator.rb
shikharbhardwaj: Added a test for this case

PR #4028, opened 2025-10-14
Postgres/K8s Networking Change - Phase 1

jeremyevans:
This implements the first phase of the Postgres/K8s networking change, which will make Postgres/K8s subnets be customer subnets, and split the firewalls into an internal and customer firewall.

This:

* Prevents deletion of customer subnets with attached internal firewalls (firewalls in other projects). This may not be strictly necessary, since we already check for attached VMs (which would include internal VMs).
* Only deletes the related customer private subnet and customer firewall when deleting a postgres resource if they are not used by other customer assets and haven't been renamed.
* Makes firewall names for kubernetes clusters be in `UBID-firewall` format, matching the format used for postgres resources.
* Adds KubernetesCluster.fix_firewall_names, which will rename existing firewalls.

The first phase needs to stop here, because I want to fix the kubernetes cluster firewall names before adding the same checks relating to kubernetes cluster private subnet deletion that were added to postgres resource private subnet deletion.

----

> [!IMPORTANT]
> Implements phase 1 of Postgres/K8s networking changes, focusing on firewall and subnet management, with updates to models, migrations, and tests.
> 
>   - **Behavior**:
>     - Prevents deletion of customer subnets with attached internal firewalls in `private_subnet.rb`.
>     - Deletes customer private subnet and firewall only if not used by other assets and not renamed in `postgres_resource_nexus.rb`.
>     - Firewall names for Kubernetes clusters now follow `UBID-firewall` format in `kubernetes_cluster.rb`.
>     - Adds `KubernetesCluster.fix_firewall_names` to rename existing firewalls.
>   - **Models**:
>     - Adds `incr_destroy_if_only_used_internally` to `PrivateSubnet` for conditional deletion.
>     - Adds `vm_firewalls` association to `Vm` for VM-specific firewalls.
>     - Updates `firewalls_vms` references in `firewall.rb` and `vm.rb`.
>   - **Migrations**:
>     - Adds `20251015_vm_specific_firewalls.rb` to create `firewalls_vms` table.
>   - **Tests**:
>     - Adds tests for `incr_destroy_if_only_used_internally` in `private_subnet_spec.rb`.
>     - Updates tests in `kubernetes_cluster_nexus_spec.rb` and `postgres_resource_nexus_spec.rb` to reflect new firewall and subnet handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 63e2ee7063592487b12bae7dc163e63bb2fd837f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -21,6 +21,19 @@ class KubernetesCluster < Sequel::Model
   plugin SemaphoreMethods, :destroy, :sync_kubernetes_services, :upgrade, :install_metrics_server, :sync_worker_mesh, :install_csi, :update_billing_records
   include HealthMonitorMethods
 
+  # :nocov:
+  # Remove after firewall names fixed
+  def self.fix_firewall_names
+    DB.transaction do
+      all do |kc|
+        # Will fail if there are multiple firewalls
+        Firewall.where(id: kc.private_subnet.firewalls_dataset.select { firewall[:id] })
+          .update(name: "#{kc.ubid}-firewall")
+      end
+    end
+  end
+  # :nocov:
+
   def validate
     super
     errors.add(:cp_node_count, "must be a positive integer") unless cp_node_count.is_a?(Integer) && cp_node_count > 0


model/private_subnet.rb
```diff
@@ -158,6 +158,27 @@ def old_aws_subnet?
     location.aws? && net4.netmask.prefix_len == DEFAULT_SUBNET_PREFIX_LEN
   end
 
+  def incr_destroy_if_only_used_internally(ubid:, vm_ids:)
+    firewalls_dataset = self.firewalls_dataset
+
+    # Destroy customer firewall if it isn't used in other private subnets
+    # and hasn't been renamed.
+    firewalls_dataset
+      .where(name: "#{ubid}-firewall")
```
furkansahin, line 167: Why are we enforcing a renamed or not rule here? It could as well be just that customer renamed it to better understand what the firewall is and simply that. Maybe we can add a separate field for filtering?

prog/vnet/subnet_nexus.rb
```diff
@@ -33,16 +33,19 @@ def self.assemble(project_id, name: nil, location_id: Location::HETZNER_FSN1_ID,
...
         port_range = allow_only_ssh ? 22..22 : 0..65535
-        fw_name = "#{name[0, 55]}-default"
-        # As is typical when checking before inserting, there is a race condition here with
-        # a user concurrently manually creating a firewall with the same name.  However,
-        # the worst case scenario is a bogus error message, and the user could try creating
-        # the private subnet again.
-        unless firewall_dataset.where(Sequel[:firewall][:name] => fw_name).empty?
-          fw_name = "#{name[0, 47]}-default-#{Array.new(7) { UBID.from_base32(rand(32)) }.join}"
+
+        unless firewall_name
```
furkansahin, line 37: should we error out if both `firewall_name` and `firewall_id` are passed? I am not sure if it's safe to simply omit.

furkansahin (approved):

model/private_subnet.rb
jeremyevans: The idea here is to remove the firewall/subnet if the customer did not modify it or use it for anything else. If the customer renamed it, I don't think we should automatically destroy it. If the customer renamed it, I think it's reasonable to expect the customer to destroy it if it isn't needed.

prog/vnet/subnet_nexus.rb
jeremyevans: I can add that if you think it is needed. I didn't think it was needed. It seems unlikely that an Ubicloud developer would attempt to provide both by accident.

PR #4032, opened 2025-10-15
Add ips-v4 For Broadcasting IP Ranges We Use

aamederen:
Add ips-v4 for broadcasting IP ranges we use

This change adds a new endpoint /ips-v4 that reports the IP ranges we can use.
    This is especially useful for GHA users to allowlist their endpoints
    for requests from Ubicloud runner VMs. The results are calculated initially
    once and reused on subsequent calls.

The format is the same with Cloudflare's: https://www.cloudflare.com/ips-v4/

### DIFF
--- a/lib/util.rb
+++ b/lib/util.rb
@@ -102,11 +102,9 @@ module Util
 
   def self.calculate_ips_v4
     Address
-      .select { set_masklen(:cidr, 16).as(:masked) }
       .where { (family(:cidr) =~ 4) & (masklen(:cidr) !~ 32) }
       .distinct
-      .order(:masked)
-      .map(:masked)
+      .select_order_map { set_masklen(:cidr, 16).as(:masked) }
       .join("\n")
   end
 end


enescakir:
> We should remember to update the file as we create new hosts:

This part is tricky. I'm sure we'll forget it eventually.

enescakir:
You might create this file while booting the web dynos, so it's updated with every deployment at least. Something like

```ruby
  def populate_ipv4_txt
    ips = Address.where { (family(cidr) =~ 4) & (routed_to_host_id !~ id) }
      .map { NetAddr::IPv4Net.new(it.cidr.network, NetAddr::Mask32.new(16)) }
    ips.sort!(&:cmp)
    ips.map!(&:to_s)
    ips.uniq!
    File.write("var/ips-v4.txt", ips.join("\n"))
  end
```

enescakir:
Another thing, since customers may frequently call this endpoint, we can return
the file's contents with cache headers. This way, the WAF can cache the
result and return it to customers without hitting the API each time.

```ruby
  r.get "ips-v4" do
    response["content-type"] = "text/plain"
    response["cache-control"] = "public, max-age=3600"
    next File.read("var/ips-v4.txt")
  end
```

aamederen:
@enescakir I initially thought about this and decided to go with the less LoC approach, thinking that this endpoint may not be used that much and we don't need to update the IPs that frequently. 

For the auto update, I thought we can setup an action to check the freshness of the file every once in a while. 

Alternatively, we can ditch the file entirely, calculate the string at boot and return it through the endpoint.

furkansahin:
I agree that it wouldn't be used that often, and masking the subnets with `/16` probably already gives a much much larger area of coverage. However, it is also bad that someone needs these results, the data is stale and they get a request from an IP that is not in our result set. Considering we make a handful of deployments everyday, I would feel OK to calculate it once per deployment but not OK to use a stale file. Also, ongoing cost would drastically reduce by adding those 5-6 lines of code that we will never touch again.

jeremyevans:
I am also of the opinion that we should calculate this at boot, store it in a string constant, and return that in a route (using the HTTP caching headers). The amount of data is small, so keeping it in memory seems fine. It seems unlikely we'd need to use a different approach in the future, even with substantial growth, but even if so, we can cross that bridge when we come to it.

I think you can optimize the code and do the calculation in the database using the PostgreSQL  `set_masklen` function.

aamederen:
Two people sharing the sentiment is enough and three is plenty! Thanks for the inputs. I will update the code accordingly.

fdr:
I agree, it needs to be adaptive, we don't have the wherewithal for supervision.

aamederen:
@jeremyevans I need some help here. I made the change but couldn't write tests that would pass in the frozen case. Since we calculate and store the results in an instance variable of `Clover`, I think the easiest way for test would be to change that value before my test case with the filled dataset, but since Clover is frozen, I cannot set it so the test fails. Maybe we can put some seed data for the tests, but I'm not sure if there's a more desirable way. What would you suggest?

jeremyevans:
Add a Clover singleton attr_reader (`singleton_class.attr_reader :ips_v4`), call `self.class.ips_v4` in the route, and override `Clover.ips_v4` in the spec (with the appropriate `allow_mocking` change). An added bonus is removing the `instance_variable_get` call.

lib/util.rb
```diff
@@ -99,4 +99,14 @@ def self.aws_tag_specifications(resource_type, name, additional_tags = {})
   def self.monitor_process?
     ENV["MONITOR_PROCESS"] == "1"
   end
+
+  def self.calculate_ips_v4
+    Address
+      .select { set_masklen(:cidr, 16).as(:masked) }
+      .where { (family(:cidr) =~ 4) & (masklen(:cidr) !~ 32) }
+      .distinct
+      .order(:masked)
```
enescakir: I'm not sure how it sorts but jfyi

https://github.com/ubicloud/ubicloud/pull/3475#discussion_r2157332248

aamederen: this sorts by address

```
5.9.0.0/16
46.4.0.0/16
64.131.0.0/16
77.92.0.0/16
108.62.0.0/16
109.94.0.0/16
135.181.0.0/16
136.243.0.0/16
144.76.0.0/16
148.251.0.0/16
178.63.0.0/16
188.40.0.0/16
207.58.0.0/16
216.22.0.0/16
```

public/ips-v4
```diff
@@ -0,0 +1,14 @@
...
+144.76.0.0/16
+148.251.0.0/16
+178.63.0.0/16
+188.40.0.0/16
+207.58.0.0/16
+216.22.0.0/16
+46.4.0.0/16
+5.9.0.0/16
+64.131.0.0/16
+77.92.0.0/16
```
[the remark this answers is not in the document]
aamederen: The example we refer to, https://www.cloudflare.com/ips-v4/, doesn't have a newline. (They also don't sort but that's OK)

aamederen:
Thank you @jeremyevans, I made the changes based on the comments in this PR. In this form, we don't test that the calculation happens only once in the lifetime of the process, since we mock the ips-v4 function.

clover.rb
```diff
@@ -991,6 +983,12 @@ def locked_domain_for(email)
         end
       end
 
+      r.get "ips-v4" do
```
enescakir, line 986: Some other routes call `before_main_hash_branches`. I’m not sure how we use it, but @jeremyevans might know whether we should run it or not.

jeremyevans: This route doesn't use any parameters, so I don't think we need to call it.

lib/util.rb
```diff
@@ -99,4 +99,14 @@ def self.aws_tag_specifications(resource_type, name, additional_tags = {})
...
+      .map(:masked)
```
jeremyevans:
```suggestion
      .where { (family(:cidr) =~ 4) & (masklen(:cidr) !~ 32) }
      .distinct
      .select_order_map { set_masklen(:cidr, 16).as(:masked) }
```

jeremyevans (approved):

jeremyevans:
> might need to cache that some day (setting a header might be enough to let some kind of middleware take care of it), but that day doesn't have to be today.

This does cache it (database query run once at startup), and uses cache headers. Is there some other type of caching you were thinking of?

fdr:
No I’m just not reading carefully enough/forgot the context. Forget what I
said.

PR #4033, opened 2025-10-15
Fix passed frame for load balancer buds

enescakir:
In our codebase, we try to keep our JSON keys as string, since when they fetched from DB, they are always string. When we pass it as a symbol, it causes failure especially in equality check.

I found a place where we are passing symbol instead of string thanks to new version of json package. It prints following warning:

    This will raise an error in json 3.0 unless enabled via `allow_duplicate_key: true`
    .../mise/installs/ruby/3.4.7/lib/ruby/3.4.0/delegate.rb:349: warning: detected duplicate key "subject_id" in {"subject_id" => "9749a267-586c-842b-962c-666f22da28fa", subject_id: "adede52c-c92e-8774-9b28-e33ff0c42d62", load_balancer_id: "9749a267-586c-842b-962c-666f22da28fa"}.

As you see `"subject_id"` and `:subject_id` are different keys. They can't merged in `hash.merge` method.

----

> [!IMPORTANT]
> Fix JSON key handling by changing from symbols to strings in `load_balancer_nexus.rb` and `load_balancer_remove_vm.rb`, and update tests accordingly.
> 
>   - **Behavior**:
>     - Fix JSON key handling by changing from symbols to strings in `load_balancer_nexus.rb` and `load_balancer_remove_vm.rb`.
>     - Specifically, update `bud` method calls to use string keys for `subject_id` and `load_balancer_id`.
>   - **Tests**:
>     - Update `load_balancer_remove_vm_spec.rb` to expect string keys in `bud` method calls.
>     - Ensure tests reflect changes in JSON key handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for ee5296e000bea81400a736c6dd95c64929819b7f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -145,7 +145,7 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
     decr_destroy
     strand.children.map { it.destroy }
     load_balancer.vms.each do |vm|
-      bud Prog::Vnet::LoadBalancerRemoveVm, {subject_id: vm.id, load_balancer_id: load_balancer.id}, :destroy_vm_ports_and_update_node
+      bud Prog::Vnet::LoadBalancerRemoveVm, {"subject_id" => vm.id, "load_balancer_id" => load_balancer.id}, :destroy_vm_ports_and_update_node
     end
     hop_wait_all_vms_removed
   end
--- a/prog/vnet/load_balancer_remove_vm.rb
+++ b/prog/vnet/load_balancer_remove_vm.rb
@@ -13,7 +13,7 @@ class Prog::Vnet::LoadBalancerRemoveVm < Prog::Base
 
   label def destroy_vm_ports_and_update_node
     load_balancer.vm_ports_by_vm(vm).destroy
-    bud Prog::Vnet::UpdateLoadBalancerNode, {subject_id: vm.id, load_balancer_id: load_balancer.id}, :update_load_balancer
+    bud Prog::Vnet::UpdateLoadBalancerNode, {"subject_id" => vm.id, "load_balancer_id" => load_balancer.id}, :update_load_balancer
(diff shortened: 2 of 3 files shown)


jeremyevans (commented):
I propose this simpler and more general change:

```diff
diff --git a/prog/base.rb b/prog/base.rb
index e8a03a466..158f76413 100644
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -180,7 +180,7 @@ end
   end
 
   def bud(prog, new_frame = {}, label = "start")
-    new_frame = {"subject_id" => @subject_id}.merge(new_frame)
+    new_frame = {"subject_id" => @subject_id}.merge(new_frame.transform_keys(&:to_s))
     strand.add_child(
       id: Strand.generate_uuid,
       prog: Strand.prog_verify(prog),
```

enescakir:
> I propose this simpler and more general change:

Makes sense. I kept fixes but also added key transform for both `bud` and `push` at [b691f3c](https://github.com/ubicloud/ubicloud/pull/4033/commits/b691f3c65704a8be9201d0790ee061c1d59f9063)

jeremyevans (commented):
I don't think we should do both changes. If you'd like to change the callers to use strings, then I see no need to transform the keys. If we do run into future duplicate key issues, we can consider those bugs to be fixed in callers. So I'd drop the second commit.

enescakir:
> I don't think we should do both changes. If you'd like to change the callers to use strings, then I see no need to transform the keys. If we do run into future duplicate key issues, we can consider those bugs to be fixed in callers. So I'd drop the second commit.

Since we have only 3 examples of symbol keys, I fixed them to be consistent in the codebase. Removed the last commit

PR #4035, opened 2025-10-15
Configure swap space and memory overcommit for PG instances.

byucesoy:
**Configure swap space for small machines**
Burstable machines come with low memory, which can frequently lead to out of
memory situations. To mitigate this, we configure a swap file to provide extra
memory.

**Configure memory_overcommit**
When OOM killer kills one of the PostgreSQL processes, there is risk of process
leaving the shared memory in an inconsistent state. To prevent this, when one
of its processes is killed, postmaster intentionally crashes itself and then
performs crash recovery on the next startup. This is required to ensure data
consistency.

Still, it is not a good experience to have PostgreSQL crash due to OOM killer,
so we configure memory overcommit to be strict.

----

> [!IMPORTANT]
> Configure swap space and strict memory overcommit for PostgreSQL instances, with updates to initialization scripts and tests.
> 
>   - **Behavior**:
>     - Configure swap space for burstable VMs in `postgres_server_nexus.rb` by setting `swap_size_bytes` to 4 GiB.
>     - Set strict memory overcommit in `configure_memory_overcommit()` in `postgres_setup.rb`.
>   - **Scripts**:
>     - Add `configure_memory_overcommit()` call in `initialize-database-from-backup` and `initialize-empty-database`.
>     - Fix newline issue in `add_to_fstab`.
>   - **Tests**:
>     - Update `assemble` test in `postgres_server_nexus_spec.rb` to use `burstable-2` VM size.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 2fbf821c55d3acfd4c8736843b72b7c11aaa2ce0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -37,7 +37,7 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
         name: ubid.to_s,
         size: postgres_resource.target_vm_size,
         storage_volumes: [
-          {encrypted: true, size_gib: 30, vring_workers: 1},
+          {encrypted: true, size_gib: 16, vring_workers: 1},
           {encrypted: true, size_gib: postgres_resource.target_storage_size_gib, vring_workers: 1}
         ],
         boot_image: boot_image,


rhizome/postgres/lib/postgres_setup.rb
```diff
@@ -17,6 +17,29 @@ def install_packages
     end
   end
 
+  def configure_swap
```
enescakir: cloudinit has swap options. You don't need to setup manually. We use it for runners

enescakir: `Prog::Vm::Nexus.assemble` has `swap_size_bytes` param
https://github.com/ubicloud/ubicloud/blob/7ef3e6abdf10c26253f1972ee64ca26d3a15dd9d/prog/vm/nexus.rb#L17

rhizome/postgres/lib/postgres_setup.rb
```diff
@@ -17,6 +17,29 @@ def install_packages
...
+        r "sudo chmod 600 /swapfile"
+        r "sudo mkswap /swapfile"
+        r "sudo swapon /swapfile"
+      end
+
+      r "sudo common/bin/add_to_fstab /swapfile none swap sw 0 0"
+    end
+  end
+
+  def configure_memory_overcommit
```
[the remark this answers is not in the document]
byucesoy: This only runs once during provisioning. It might get executed multiple times in rare conditions, but even in such cases it is safe to have duplicate entries.

PR #4042, opened 2025-10-15
Add firewalls_vms join table

jeremyevans:
This allows for vm-specific firewalls, which offer two advantages:

1) VMs in the same private subnet can have separate firewall rules.
   Not all VMs in the same subnet run the same services, and there
   could easily be different firewall rules desired.

2) This will make it easier for managed services to use customer
   subnets. Instead of adding an internal firewall to the related
   subnet (which would affect other VMs in the subnet, even if they
   are customer managed and not-Ubicloud managed), the internal
   firewall is added directly to the VM, and only affects that
   specific VM.

----

> [!IMPORTANT]
> Add `firewalls_vms` join table for VM-specific firewall rules, enabling separate rules for VMs in the same subnet and facilitating managed services using customer subnets.
> 
>   - **Behavior**:
>     - Adds `firewalls_vms` join table in `20251015_vm_specific_firewalls.rb` for VM-specific firewall rules.
>     - Enables separate firewall rules for VMs in the same subnet.
>     - Facilitates managed services using customer subnets without affecting other VMs.
>   - **Models**:
>     - Updates `firewall.rb` to include `firewalls_vms` reference.
>     - Updates `vm.rb` to include `firewalls_vms` reference.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for d54090012cca22108a027a835b4237bb95c7f857. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fdr:
I am hoping to avoid this knob until we've exhausted our alternatives and understood very clearly why we need to have it, and ideally somehow signal to the user when they should be using it.

I think it's fine/intended that firewall rules apply to whole subnets worth of VMs, affecting adjacent virtual machines because the general idea is that subnets should contain VMs that are like one another in service exposure.

People will not have this idea when they show up unless gently taught it somehow, so I am seeking to push them in that direction by leaving out a feature like this so they get a hint to reconsider sooner. 

When people start gluing virtual machines together with such a feature without regard for how they sit in contiguous network address space, the implementation is pushed towards more intense micro-management in the data plane, which implies different quotas and limits and data models (e.g. "labels" in GCP).

I have found people wind up still needing to write existential qualifications, but now they're over VMs rather than subnets, and those existential qualifications no longer have anything to do with contiguous address space, e.g. exploding VMs referencing each other in some abstract way into a huge number of below-the-line firewall rules each encoding a single address, since that is ultimately is what is stamped into each packet. When the existential qualification is written against subnets, a neat, contiguous range of integers that is efficient in data plane implementation suffices.

Put another way, I intended to try making Ubicloud opinionated about what subnets were for, correlating their reason to exist with firewalls.

This is a novel design conceit, so I admit perhaps the theory is, or could be, a failure and there's no way to make the many-subnets model usable. That said, I modeled the conceit on a general, more IaaS-like in terms of network feature set version of Heroku "process types," which also de-emphasize individuality of processes, as I thought it worked well and was more convenient for more cases.

jeremyevans:
Note that we don't need to expose this to the user if we don't want to. It's possible to only use this feature in managed VMs. However, I think this will make managed-vms-in-customer-subnets feature much simpler and safer than the alternative of having a customer private subnet with internal firewall in a different project as well as a customer firewall in the customer project.

migrate/20251015_vm_specific_firewalls.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:firewalls_vms) do
+      foreign_key :firewall_id, :firewall, type: :uuid
+      foreign_key :vm_id, :vm, type: :uuid
+      primary_key [:firewall_id, :vm_id]
+      index [:vm_id, :firewall_id]
```
furkansahin, line 9: do we need both of these or just `primary_key` suffice?

furkansahin (approved):

migrate/20251015_vm_specific_firewalls.rb
jeremyevans: Primary key suffices to enforce uniqueness, but if you are looking up by `vm_id`, you need the other index to avoid a sequential scan. You don't necessarily need `:firewall_id` in the second index, but having it allows for a index-only scans, so I think it should be included (could use `:include` for that, I guess, but I didn't think it worth it).

PR #4045, opened 2025-10-15
Postgres/K8s Networking Change - Phase 2

jeremyevans:
This assumes that  KubernetesCluster.fix_firewall_names added in Phase 1 has been run, and removes it as it won't be needed again.

It makes the same change to Prog::Kubernetes::KubernetesClusterNexus#destroy that was made to Prog::Postgres::PostgresResourceNexus, only destroying the firewall and private subnet if they have not been renamed and are not used by other customer objects.

This allows creation of VMs inside a private subnet in another project, which can be used by managed services to create their VMs inside a customer private subnet. This ability is off by default, and requires a keyword argument to enable.

This makes the web UI and API (and therefore SDK/CLI) for private subnets do not show related VMs if the VMs are located in a different project. We could make changes in the future so that it shows postgres resources and kubernetes clusters as being inside the private subnet, but I don't think we should show the underlying VMs, as the customer does not manage those directly. 

----

> [!IMPORTANT]
> Enhances private subnet and VM handling across projects, updates destruction logic, and modifies UI to reflect cross-project subnet usage.
> 
>   - **Behavior**:
>     - Removes `KubernetesCluster.fix_firewall_names` from `kubernetes_cluster.rb` as it is no longer needed.
>     - Updates `KubernetesClusterNexus#destroy` to only destroy firewalls and private subnets if not renamed and not used by other customer objects.
>     - Allows VM creation in private subnets of other projects with `allow_private_subnet_in_other_project` flag in `vm/nexus.rb`.
>     - Updates UI to hide VMs in private subnets from different projects in `private_subnet.rb` and `vms.erb`.
>   - **Tests**:
>     - Updates `kubernetes_cluster_nexus_spec.rb` and `vm/nexus_spec.rb` to test new destruction logic and cross-project subnet usage.
>     - Adds tests in `private_subnet_spec.rb` to verify UI changes for VMs in different projects.
>     - Modifies `golden_files_spec.rb` to reflect changes in VM and subnet handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for fb46ab224c50e27a63212011f89c618ca7b8a832. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/views/networking/private_subnet/vms.erb
+++ b/views/networking/private_subnet/vms.erb
@@ -1,4 +1,8 @@
-<% viewable_vm_map = dataset_authorize(Vm.where(id: @ps.nics_dataset.select(:vm_id)), "Vm:view").select_hash(:id, Sequel.as(true, :v)) %>
+<% vm_ds = @ps.vms_dataset.where(project_id: @ps.project_id)
+viewable_vm_ds = dataset_authorize(vm_ds, "Vm:view")
+displayable_vm_map, linkable_vm_map = [vm_ds, viewable_vm_ds].map! do
+  it.select_hash(Sequel[:vm][:id], Sequel.as(true, :v))
+end %>
 
 <div id="private-subnet-nics" class="p-6">
   <%== part(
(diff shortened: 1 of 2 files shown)


views/networking/private_subnet/vms.erb
```diff
@@ -7,12 +11,12 @@
     headers: ["VM", "Private IPv4", "Private IPv6"],
     empty_state: "No VM attached",
     rows:
-      @ps.nics(eager: :vm).map do |nic|
+      @ps.nics(eager: :vm).select { !it.vm_id || displayable_vm_map[it.vm_id] }.map! do |nic|
```
furkansahin: can we use `it.vm_id.nil?` instead of `!` because there is already a logical `or` operation and I think putting the `!` instead of `.nil?` reduces readability.

furkansahin (approved):

views/networking/private_subnet/vms.erb
jeremyevans: Sure, I can make that change.

PR #4050, opened 2025-10-16
Add extra fields about postgres connection info

byucesoy:
We normally only return the connection string, but then clients might need to parse it if they need individual fields, which is not very straightforward as each field might contain special characters requiring extra attention during parsing. So we are adding username, password and hostname fields to make it easier for clients to consume this information.

----

> [!IMPORTANT]
> Add `username`, `password`, and `hostname` fields to Postgres database response for detailed connection info.
> 
>   - **Behavior**:
>     - Added `username`, `password`, and `hostname` fields to Postgres database response in `openapi.yml` and `serializers/postgres.rb`.
>     - These fields are included when `detailed` option is true.
>   - **OpenAPI**:
>     - Updated `openapi.yml` to include `username`, `password`, and `hostname` in Postgres database schema.
>     - Marked these fields as required in detailed responses.
>   - **Serializers**:
>     - Updated `serialize_internal()` in `serializers/postgres.rb` to add `username`, `password`, and `hostname` when `detailed` is true.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a19bba13eda4249396c1b97efe8fc31a53e16077. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/serializers/postgres.rb
+++ b/serializers/postgres.rb
@@ -25,6 +25,9 @@ class Serializers::Postgres < Serializers::Base
     if options[:detailed]
       base.merge!(
         connection_string: pg.connection_string,
+        username: "postgres",
+        password: pg.superuser_password,
+        hostname: pg.hostname,
         primary: pg.representative_server&.primary?,
         firewall_rules: Serializers::PostgresFirewallRule.serialize(pg.firewall_rules.sort_by { |fwr| fwr.cidr.version && fwr.cidr.to_s }),
         metric_destinations: pg.metric_destinations.map { {id: it.ubid, username: it.username, url: it.url} },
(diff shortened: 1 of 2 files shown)


serializers/postgres.rb
```diff
@@ -25,6 +25,9 @@ def self.serialize_internal(pg, options = {})
     if options[:detailed]
       base.merge!(
         connection_string: pg.connection_string,
+        username: "postgres",
+        password: pg.superuser_password,
+        hostname: pg.hostname,
```
shikharbhardwaj, line 30: Do we need to update openapi.yml as well? I guess committee is fine with extra fields in responses

byucesoy: I guess it is. Still, it makes sense to add them. I added them to openapi.yml now.

PR #4053, opened 2025-10-16
Postgres/K8s Networking Change - Phase 3

jeremyevans:
This switches new postgres resources and kubernetes clusters to use an internal firewall attached directly to the underlying VMs. This is where firewall rules that are necessary to run the managed service are placed.  For postgres, we already had these internal rules in the code.  For kubernetes, the internal firewall rules are new.

For both postgres and kubernetes, a private subnet is created in the customer's project, and a related firewall is also created in the customer's project. These private subnets can be connected to other customer private subnets, and the firewall rules for these firewalls can be modified by the customer for their security needs, without breaking the access Ubicloud needs to manage the VMs in these subnets.

The next phase of this change will convert existing postgres resources and kubernetes clusters before this change to use the same internal firewall/customer firewall approach, and simplify some of these changes (e.g. assume the internal firewall exists instead of checking for it).

----

> [!IMPORTANT]
> This PR introduces internal firewalls for Kubernetes and Postgres resources, updates firewall rules, and modifies tests and CLI golden files accordingly.
> 
>   - **Networking**:
>     - Introduces internal firewalls for Kubernetes clusters in `kubernetes_cluster.rb` and Postgres resources in `postgres_resource.rb`.
>     - Updates `set_firewall_rules` in `postgres_resource.rb` to handle internal firewalls.
>     - Modifies `kubernetes_cluster_nexus.rb` to create and manage internal firewalls for control plane and worker nodes.
>   - **Tests**:
>     - Updates tests in `postgres_resource_spec.rb`, `kubernetes_cluster_nexus_spec.rb`, and `kubernetes_node_nexus_spec.rb` to validate internal firewall creation and management.
>     - Adjusts CLI golden files in `spec/routes/api/cli/golden-files` to reflect new firewall and subnet configurations.
>   - **Misc**:
>     - Adds helper methods for firewall management in `postgres_timeline.rb` and `kubernetes_cluster.rb`.
>     - Updates `thawed_mock.rb` to allow mocking of new methods related to firewalls.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for dc2d4101d09a472632b5ec4bd47b274bb26e3f23. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -300,7 +300,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
 
       kubernetes_cluster.private_subnet.incr_destroy_if_only_used_internally(
         ubid: kubernetes_cluster.ubid,
-        vm_ids: kubernetes_cluster.cp_vms.map(&:id) + kubernetes_cluster.nodes.map(&:vm_id)
+        vm_ids: kubernetes_cluster.all_nodes.map(&:vm_id)
       )
 
       kubernetes_cluster.nodes.each(&:incr_destroy)
--- a/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
+++ b/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
@@ -8,22 +8,22 @@ RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
   let(:st) { Strand.new(id: "8148ebdf-66b8-8ed0-9c2f-8cfe93f5aa77") }
 
   let(:customer_project) { Project.create(name: "default") }
-  let(:subnet) { PrivateSubnet.create(net6: "0::0", net4: "127.0.0.1", name: "x", location_id: Location::HETZNER_FSN1_ID, project_id: customer_project.id) }
+  let(:subnet) { kubernetes_cluster.private_subnet }
 
   let(:kubernetes_cluster) {
-    kc = KubernetesCluster.create(
+    kc = described_class.assemble(
       name: "k8scluster",
       version: Option.kubernetes_versions.first,
       cp_node_count: 3,
-      private_subnet_id: subnet.id,
       location_id: Location::HETZNER_FSN1_ID,
       project_id: customer_project.id,
       target_node_size: "standard-2"
-    )
+    ).subject
     KubernetesNodepool.create(name: "k8stest-np", node_count: 2, kubernetes_cluster_id: kc.id, target_node_size: "standard-2")
 
     dns_zone = DnsZone.create(project_id: Project.first.id, name: "somezone", last_purged_at: Time.now)
 
+    subnet = kc.private_subnet
     apiserver_lb = LoadBalancer.create(private_subnet_id: subnet.id, name: "somelb", health_check_endpoint: "/foo", project_id: Config.kubernetes_service_project_id)
     LoadBalancerPort.create(load_balancer_id: apiserver_lb.id, src_port: 123, dst_port: 456)
     [create_vm, create_vm].each do |vm|
@@ -567,12 +567,28 @@ RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
 
     it "naps until all nodepools are gone" do
       st.update(prog: "Kubernetes::KubernetesClusterNexus", label: "destroy", stack: [{}])
+      kubernetes_nodepool = kubernetes_cluster.nodepools.first
+      Prog::Kubernetes::KubernetesNodeNexus.assemble(
+        Config.kubernetes_service_project_id,
+        sshable_unix_user: "ubi",
+        name: "t3",
+        location_id: kubernetes_cluster.location_id,
+        size: kubernetes_nodepool.target_node_size,
+        storage_volumes: [{encrypted: true, size_gib: kubernetes_nodepool.target_node_storage_size_gib}],
+        boot_image: "kubernetes-#{kubernetes_cluster.version.tr(".", "_")}",
+        private_subnet_id: kubernetes_cluster.private_subnet_id,
+        enable_ip4: true,
+        kubernetes_cluster_id: kubernetes_cluster.id,
+        kubernetes_nodepool_id: kubernetes_nodepool.id
+      ).subject
       expect(kubernetes_cluster.nodes).to all(receive(:incr_destroy))
       expect(kubernetes_cluster.nodepools).to all(receive(:incr_destroy))
 
       expect(kubernetes_cluster).not_to receive(:destroy)
 
+      expect(kubernetes_cluster.private_subnet.semaphores_dataset.select_map(:name)).to eq []
       expect { nx.destroy }.to nap(5)
+      expect(kubernetes_cluster.private_subnet.semaphores_dataset.select_map(:name)).to eq ["update_firewall_rules", "destroy"]
     end
 
     it "naps until all control plane nodes are gone" do
@@ -581,7 +597,23 @@ RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
       expect(kubernetes_cluster.nodes).to all(receive(:incr_destroy))
       expect(kubernetes_cluster.nodepools).to be_empty
 
+      expect(kubernetes_cluster.private_subnet.semaphores_dataset.select_map(:name)).to eq []
       expect { nx.destroy }.to nap(5)
+      expect(kubernetes_cluster.private_subnet.semaphores_dataset.select_map(:name)).to eq ["update_firewall_rules", "destroy"]
+    end
+
+    it "does not incr_destroy private_subnet with other resources" do
+      st.update(prog: "Kubernetes::KubernetesClusterNexus", label: "destroy", stack: [{}])
+      kubernetes_cluster.nodepools_dataset.destroy
+      expect(kubernetes_cluster.nodes).to all(receive(:incr_destroy))
+      expect(kubernetes_cluster.nodepools).to be_empty
+
+      Firewall.create(name: "t", project_id: customer_project.id, location_id: Location::HETZNER_FSN1_ID)
+        .associate_with_private_subnet(kubernetes_cluster.private_subnet, apply_firewalls: false)
+
+      expect(kubernetes_cluster.private_subnet.semaphores_dataset.select_map(:name)).to eq []
+      expect { nx.destroy }.to nap(5)
+      expect(kubernetes_cluster.private_subnet.semaphores_dataset.select_map(:name)).to eq ["update_firewall_rules"]
     end
 
     it "triggers deletion of associated resources and completes destroy when nodepools are gone" do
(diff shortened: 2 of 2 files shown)


jeremyevans:
E2E Failure is:

```
06:30:10 e2e.1       | 54.85m | 0199f0c7-ceb2-873a-98a1-32e005e90b9b | Test::HaPostgresResource.wait_resources_destroyed | waiting for exit | Test::HaPostgresResource.wait_resources_destroyed
```

Looking at the log I see:

```
PG::ForeignKeyViolation: ERROR:  update or delete on table \"firewall\" violates foreign key constraint \"firewalls_vms_firewall_id_fkey\" on table \"firewalls_vms\"\nDETAIL:  Key (id)=(11b44724-0cbd-89fc-8a44-b7cfb56dec0b) is still referenced from table \"firewalls_vms\
```

This comes from `PrivateSubnet#incr_destroy_if_only_used_internally`:

```ruby
    Firewall.with(:firewall_ids, DB[:firewalls_vms].where(vm_id: vm_ids).returning(:firewall_id).with_sql(:delete_sql))
      .where(id: DB[:firewall_ids])
      .destroy
```

I don't think we need this code. We decided to only use vm-specific firewalls for pg/k8s, and the pg/k8s progs directly handling removing of the firewalls after the call to `PrivateSubnet#incr_destroy_if_only_used_internally`. So I think this code can be removed. I'll push a commit for that and rerun E2E. I'll also be testing this in staging today.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -18,17 +18,38 @@ def self.assemble(name:, project_id:, location_id:, version: Option.kubernetes_v
...
+      internal_cp_vm_firewall.replace_firewall_rules([
+        {cidr: "0.0.0.0/0", port_range: Sequel.pg_range(22..22)},
+        {cidr: "::/0", port_range: Sequel.pg_range(22..22)},
+        {cidr: "0.0.0.0/0", port_range: Sequel.pg_range(443..443)},
+        {cidr: "::/0", port_range: Sequel.pg_range(443..443)},
+        {cidr: subnet.net4.to_s, port_range: Sequel.pg_range(10250..10250)},
+        {cidr: subnet.net6.to_s, port_range: Sequel.pg_range(10250..10250)}
+      ])
+
+      # Internal control plane node firewall, will be directly attached to kubernetes worker VMs
```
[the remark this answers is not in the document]
jeremyevans: Change made.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -18,17 +18,38 @@ def self.assemble(name:, project_id:, location_id:, version: Option.kubernetes_v
...
-          Config.kubernetes_service_project_id,
+          project_id,
           name: "#{ubid}-subnet",
           location_id:,
           firewall_name: "#{ubid}-firewall",
           ipv4_range: Prog::Vnet::SubnetNexus.random_private_ipv4(Location[location_id], project, 18).to_s
         ).subject
       end
 
+      # Internal control plane node firewall, will be directly attached to kubernetes control plane VMs
```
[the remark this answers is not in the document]
jeremyevans: This only makes sense if we expect the cp nodes to always use a superset of the worker node firewall rules. I'm unsure whether that is the case, so doing so seems like premature abstraction to me.

jeremyevans:
E2E failed again, this time due to:

```
PG::ForeignKeyViolation: ERROR:  update or delete on table \"project\" violates foreign key constraint \"private_subnet_project_id_fkey\" on table \"private_subnet\"\nDETAIL:  Key (id)=(258ff233-ccd4-86d2-80a8-4676ba1cfb16) is still referenced from table \"private_subnet\"
```

Stemming from:

```
"/home/runner/work/ubicloud/ubicloud/prog/test/kubernetes.rb:208:in 'Prog::Test::Kubernetes#destroy'",
```

I think the issue is an incorrect check for expected vms when destroying the cluster. It was only handling cp nodes and not worker nodes. This wasn't caught by the previous Prog::Kubernetes::KubernetesClusterNexus specs because there was no testing of whether the private subnet was destroyed. Actually testing that was challenging because the specs do not test production cases.  Pushed a separate commit to fix the issue and update the specs to be more like production.

jeremyevans:
E2E still failing, again during project deletion because the subnet still exists, but this time for postgres and not kubernetes:

```
ERROR:  update or delete on table \"project\" violates foreign key constraint \"private_subnet_project_id_fkey\" on table \"private_subnet\"\nDETAIL:  Key (id)=(3c3f70ab-76f0-8ad2-b9be-dc084a26a8d8) is still referenced from table \"private_subnet\"
```

Stemming from:

```
/home/runner/work/ubicloud/ubicloud/prog/test/ha_postgres_resource.rb:119:in 'Prog::Test::HaPostgresResource#destroy'
```

Unfortunately, unlike the k8s issue, where it wasn't checking for all related VMs when determining whether the private subnet can be deleted, the postgres resource code seems to be checking for the correct VMs already:

```ruby
    postgres_resource.private_subnet.incr_destroy_if_only_used_internally(
      ubid: postgres_resource.ubid,
      vm_ids: servers.map(&:vm_id)
    )
```

However, reviewing the progs in more detail, it looks like the Nic `vm_id` is set to `nil` before the VM is destroyed. I pushed a commit that ignores the Nics with `nil` `vm_id` when deciding whether to destroy the subnet.

If this doesn't fix E2E, on Monday, I'll add some debugging code to the E2E specs to attempt to determine why the private subnet isn't being deleted.

jeremyevans:
E2E passed. I'll rebase and add a related spec on Monday.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -18,17 +18,38 @@ def self.assemble(name:, project_id:, location_id:, version: Option.kubernetes_v
...
+        {cidr: "0.0.0.0/0", port_range: Sequel.pg_range(22..22)},
+        {cidr: "::/0", port_range: Sequel.pg_range(22..22)},
+        {cidr: "0.0.0.0/0", port_range: Sequel.pg_range(443..443)},
+        {cidr: "::/0", port_range: Sequel.pg_range(443..443)},
+        {cidr: subnet.net4.to_s, port_range: Sequel.pg_range(10250..10250)},
+        {cidr: subnet.net6.to_s, port_range: Sequel.pg_range(10250..10250)}
+      ])
+
+      # Internal worker node firewall, will be directly attached to kubernetes worker VMs
+      internal_worker_vm_firewall = Firewall.create(name: "#{ubid}-worker-vm-firewall", location_id:, description: "Kubernetes worker node internal firewall", project_id: Config.kubernetes_service_project_id)
```
aamederen, line 45: cc @mohi-kalantari 

Just to understand the UX better, after this change, users would need to open specific ports (most commonly 80 and 443) of their worker vm firewall as an extra step while deploying apps, right? If so, we will need to update our docs and tutorials to reflect this change as well.

aamederen: Btw, AFAIK, we still allow all traffic within the subnet regardless the VM specific FW rules, right? If not, then we might break even the simples non-exposed app deployment flow.

Should we change an K8s E2E test to use the new flow with a customer subnet?

jeremyevans: > Just to understand the UX better, after this change, users would need to open specific ports (most commonly 80 and 443) of their worker vm firewall as an extra step while deploying apps, right? If so, we will need to update our docs and tutorials to reflect this change as well.

No. By default, everything remains as before, with all ports open to all IP addresses. What changes is that users have the ability to remove the default rules and impose their own rules. If users choose to do so, and do so improperly, they risk breaking their use of Kubernetes. However, that's the same as if they modify firewall rules improperly for a regular VM. If users do not modify the firewall rules for the created subnet, there should be no risk of anything breaking network-wise, as all traffic is allowed.

> Btw, AFAIK, we still allow all traffic within the subnet regardless the VM specific FW rules, right? If not, then we might break even the simples non-exposed app deployment flow.

No. By default for VMs, subnet traffic is not allowed unless there is a firewall rule allowing it. You specified that the firewall rules needed for our own management were:

1) Port 22 from all IPs
2) Port 443 from all IPs (for cp nodes, not worker nodes)
3) Port 10250 from the private subnet IPs

So ports 22, 443 (cp nodes), and 10250 from the private subnet are allowed. If the user removes the default firewall rules for the created subnet, those will be the only firewall rules allowed.

If other firewall rules are required for our management of Kubernetes, they should be added to the internal firewall rules. However, if our management of Kubernetes still works, but users only break their own usage of Kubernetes if they modify the firewall rules improperly, then I think we should leave things as they are. If users do mess around with firewall rules and break things, they can always undo the firewall rule changes to fix things.
 
> Should we change an K8s E2E test to use the new flow with a customer subnet?

No need, the K8s E2E test (Prog::Test::Kubernetes) already uses a customer subnet with this change.

mohi-kalantari: Just one comment for the port 10250 @jeremyevans . There are many internal ports which we would need to whitelist in order to include everything. I feel like we can delete the 10250 port and assume that internal communication inside the subnet is allowed. otherwise, I need to provide a more comprehensive list of ports. 

Although I feel like internal firewalling does not make sense in the case of kubernetes because all internal workloads will need to communicate with each other and those ports are dynamic and randomly selected at creation time. CNI and kube proxy will keep forwarding packets to different ports on different nodes.

jeremyevans: @mohi-kalantari Please be specific. Do those ports affect our management of the kubernetes service, or do they affect the customer's use of kubernetes? Changing the number of nodes for a kubernetes cluster still worked with only the internal firewall rules, after the customer firewall rules were removed.

You cannot assume that internal communication inside the subnet will be allowed unless we explicitly allow it through firewall rules. However, again, if the firewall rules do not affect our management of kubernetes, they do not belong on the internal firewall.

mohi-kalantari: Got it. In that case, let's not touch the current rules. Rule 10250 would be later used for autoscaling which we don't have yet.

In customer's subnet firewalls, all internal traffic should be allowed but I guess that's something for later.

mohi-kalantari (approved):

prog/kubernetes/kubernetes_cluster_nexus.rb
jeremyevans: > In customer's subnet firewalls, all internal traffic should be allowed but I guess that's something for later.

In customer subnet firewalls, all traffic (internal and external) is allowed, to match the current behavior.

fdr: For kubernetes, it's fine to have a rule on the subnet that permits 0-65535 TCP + UDP, and both IPv4/IPv6 to the subnet itself, if that's what you want and really nothing more can be done about it. So you'd need four rules. Maybe you want only 1024-65535, I dunno.

This would give the "subnet as a perimeter" behavior that some people assume (and is how it used to be), but I think is generally harmful. But as kubernetes is itself a mixed-workload system (unlike some VMs attached to a load balancer), it's understandable. It is harmful in cases where it would, for example, open up lateral attacks between web VMs in a load balancer sitting in a subnet together...and much of the time, web VMs do not communicate with anything but the outside world and the database.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -8,7 +8,8 @@ def node
   end
 
   def kubernetes_nodepool
-    @kubernetes_nodepool ||= KubernetesNodepool[frame["nodepool_id"]]
+    return @kubernetes_nodepool if defined?(@kubernetes_nodepool)
```
furkansahin, line 11: is there a possibility that this is defined as nil before but gets assigned an object later which we cache a nil value?

jeremyevans: Good question to ask. The answer is no. This operates on the `frame["nodepool_id"]`, and nothing in the prog modifies that. So this should be safe.

PR #4057, opened 2025-10-17
Log spot instance interruptions for alien runners

enescakir:
Spot instances are cost-effective but can be interrupted by AWS when capacity is reclaimed.

When this happens, SSH commands begin to fail. In such cases, we check the termination status via the AWS API.

For now, we simply log these interruptions. In the future, we may stop spilling over to alien runners when the number of interruptions becomes high.

----

> [!IMPORTANT]
> Log AWS spot instance interruptions in `github_runner.rb` and update tests in `github_runner_spec.rb` to verify logging and runner destruction.
> 
>   - **Behavior**:
>     - In `github_runner.rb`, `wait` method now logs AWS spot instance interruptions using `Clog.emit` when SSH fails due to spot instance termination.
>     - If interruption is detected, the runner is marked for destruction with `github_runner.incr_destroy`.
>   - **Tests**:
>     - Added test in `github_runner_spec.rb` to verify logging of spot instance interruptions and runner destruction.
>     - Ensures SSH exceptions are re-raised for non-spot instance terminations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 6205c9613ff807194d4ea818506c86d2bd59e235. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -356,7 +356,22 @@ class Prog::Vm::GithubRunner < Prog::Base
 
   label def wait
     register_deadline(nil, 5 * 24 * 60 * 60)
-    case vm.sshable.cmd("systemctl show -p SubState --value runner-script").chomp
+    substate = begin
+      vm.sshable.cmd("systemctl show -p SubState --value runner-script").chomp
+    rescue *Sshable::SSH_CONNECTION_ERRORS
+      if vm.location.aws? && vm.aws_instance
+        instance_id = vm.aws_instance.instance_id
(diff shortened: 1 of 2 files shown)


enescakir:
> Looks good. Do we provision a new runner for the job in this case?

Since the vm was already interrupted, the job failed. The new runner can't pick up the failed job. We don't provision a new runner.

PR #4063, opened 2025-10-20
Client certificate authentication for Postgres

byucesoy:
This allows setting certain users to authenticate with client certificates instead of passwords. It is not exposed in UI or API yet, so an operator needs to configure this per database. If we see bigger demand for this feature, we can enable it in UI/API as well.

----

> [!IMPORTANT]
> Adds client certificate authentication for Postgres, requiring manual configuration by operators, with schema, model, and configuration updates.
> 
>   - **Behavior**:
>     - Adds client certificate authentication for Postgres, configurable per database by operators.
>     - Not exposed in UI/API yet; manual configuration required.
>   - **Schema**:
>     - Adds `trusted_ca_certs` (text) and `cert_auth_users` (jsonb) columns to `postgres_resource` in `20251020_postgres_cert_auth.rb`.
>   - **Models**:
>     - Updates `PostgresResource` and `PostgresServer` to handle `cert_auth_users` and `trusted_ca_certs`.
>   - **Configuration**:
>     - Modifies `configure` script to include certificate authentication entries in `pg_hba.conf`.
>     - Updates `postgres_server_nexus.rb` to handle certificate files and permissions.
>   - **Tests**:
>     - Updates tests in `postgres_server_spec.rb` and `postgres_server_nexus_spec.rb` to cover new authentication logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c5522498492ffb4a51bf16c1e3acd2b2a8360e2e. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -115,6 +115,7 @@ class PostgresServer < Sequel::Model
           net6: it.net6.to_s
         }
       },
+      cert_auth_users: resource.cert_auth_users,
       identity: resource.identity,
       hosts: "#{resource.representative_server.vm.private_ipv4} #{resource.identity}",
       pgbouncer_instances: (vm.vcpus / 2.0).ceil.clamp(1, 8),
--- a/rhizome/postgres/bin/configure
+++ b/rhizome/postgres/bin/configure
@@ -41,6 +41,10 @@ private_subnets = configure_hash["private_subnets"].flat_map {
   ]
 }.join("\n")
 
+cert_auth_users = configure_hash["cert_auth_users"].map {
+  "hostssl all             #{_1}           all                     cert"
+}.join("\n")
+
 pg_hba_entries = <<-PG_HBA
 # PostgreSQL Client Authentication Configuration File
 # ===================================================
@@ -79,6 +83,9 @@ host    replication     all             ::1/128                 scram-sha-256
 # Allow replication connection using special replication user for
 # HA standbys
 hostssl replication     ubi_replication all                     cert map=standby2replication
+
+# Allow certificate authentication for specific users
+#{cert_auth_users}
 
 # Allow connections from public internet with md5 authentication
 # Note that if the password is encrypted as scram-sha-256, PostgreSQL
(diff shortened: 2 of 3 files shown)


byucesoy:
No, customer owns the client certs. They will generate the CA and client certs and then share CA's public cert with us and operator only needs to set the field.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -173,7 +173,7 @@ def before_run
 
     nap 5 if postgres_server.resource.server_cert.nil?
 
-    ca_bundle = postgres_server.resource.ca_certificates
+    ca_bundle = [postgres_server.resource.ca_certificates, postgres_server.resource.trusted_ca_certs].compact.join("\n")
```
fdr, line 176: I'm a bit puzzled by this bundling. The CA used for authenticating clients and the CA used for TLS *can* be the same can also just as well be different. It seems like you intend to make this more a DIY situation for customer (which has the big advantage that they can control the private keying and never give it to us), which would indicate we're doing two distinct things here, right? ssl_ca_file vs ssl_cert_file.

byucesoy: Yes, the goal is allowing customers to have their own CA and they would be able to generate client certificates out of that to authenticate.

Currently, this CA already serves 2 purposes. For both TLS and standby to primary authentication, we use `server.crt` (i.e. `ssl_cert_file: '/etc/ssl/certs/server.crt'`), which is generated from `postgres_server.resource.ca_certificates`. It is possible to decouple that, but since the CA and cert is already there, it was easy to re-use it. With this PR, we start to add customer generated CA to the `ssl_ca_file`, so it is yet another case of client authentication.

I think what we can do instead is:
- Have one set of certificates (TLS CA and TLS Server Cert), to be used for TLS
- Have another set of certificates (Auth CA and Auth Server Cert), to be used for standby to primary authentication. Customer CA can be bundled with Auth CA.

I think that would be cleaner architecture. If you agree with that, I can implement it at a separate PR. I'd like to keep this one simple.

furkansahin (approved):

prog/postgres/postgres_server_nexus.rb
fdr: Okay, yes, let's decouple them. I wish this justification were in the commit message for ae2c6064e9fdb13032cf9f9d1f89687f1cfd8ee9 given you typed it up to send it to me anyway :(

PR #4068, opened 2025-10-21
Minor Postgres fixes to work better with single VmHost and single MinioCluster

jeremyevans:
First commit falls back to looking for MinioCluster in minio project, similar to how we fallback to the VictoriaMetricsResource in victoria_metrics project

Second commit allows for not enforcing that postgres servers for the same resource must be in separate Hetzner data centers. This breaks cases where there is only one host. FWIW, I don't think this should be enforced except for HA servers.

----

> [!IMPORTANT]
> Adds MinioCluster fallback and config option for Postgres servers in the same data center.
> 
>   - **Behavior**:
>     - Adds fallback to `MinioCluster` in `postgres_timeline_nexus.rb` to search in `minio` project if not found in `postgres` project.
>     - Introduces `POSTGRES_ALLOW_SERVERS_IN_SAME_DATA_CENTER` in `e2e.yml` and `config.rb` to allow Postgres servers in the same Hetzner data center.
>   - **Code Changes**:
>     - Modifies `provision_servers` in `converge_postgres_resource.rb` to use `POSTGRES_ALLOW_SERVERS_IN_SAME_DATA_CENTER` for data center exclusion logic.
>     - Updates `assemble` in `postgres_timeline_nexus.rb` to include fallback logic for `MinioCluster`.
>   - **Tests**:
>     - Adds test cases in `converge_postgres_resource_spec.rb` for provisioning logic with `POSTGRES_ALLOW_SERVERS_IN_SAME_DATA_CENTER`.
>     - Adds test cases in `postgres_timeline_nexus_spec.rb` for `MinioCluster` fallback logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 87667b073b4e53a17800615dbeb7e8d577fa5ec6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/converge_postgres_resource.rb
+++ b/prog/postgres/converge_postgres_resource.rb
@@ -17,7 +17,7 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
       exclude_host_ids = []
       exclude_availability_zones = []
       availability_zone = nil
-      if !(Config.development? || Config.is_e2e || Config.postgres_allow_servers_in_same_data_center) && postgres_resource.location.provider == HostProvider::HETZNER_PROVIDER_NAME
+      if !(Config.development? || Config.postgres_allow_servers_in_same_data_center) && postgres_resource.location.provider == HostProvider::HETZNER_PROVIDER_NAME
         used_data_centers = postgres_resource.servers.map { it.vm.vm_host.data_center }.uniq
         exclude_host_ids = VmHost.where(data_center: used_data_centers).map(&:id)
       end
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -79,6 +79,7 @@ jobs:
         CLOVER_DATABASE_URL: postgres:///clover_test?user=clover
         RACK_ENV: production
         IS_E2E: "true"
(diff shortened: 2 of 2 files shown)


prog/postgres/converge_postgres_resource.rb
```diff
@@ -17,7 +17,7 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
       exclude_host_ids = []
       exclude_availability_zones = []
       availability_zone = nil
-      if !(Config.development? || Config.is_e2e) && postgres_resource.location.provider == HostProvider::HETZNER_PROVIDER_NAME
+      if !(Config.development? || Config.is_e2e || Config.postgres_allow_servers_in_same_data_center) && postgres_resource.location.provider == HostProvider::HETZNER_PROVIDER_NAME
```
byucesoy: I think we can remove is_e2e condition and set the new config in e2e tests.

jeremyevans: Pushed another commit for that.

PR #4069, opened 2025-10-21
Postgres/K8s Networking Change - Phase 4

jeremyevans:
For Postgres resources and k8s clusters created before phase 3, this adds internal firewalls and switches the existing firewalls and subnets to be in the customer project.

----

> [!IMPORTANT]
> Adds internal firewalls and transfers existing firewalls and subnets to customer project for pre-phase 3 Postgres and Kubernetes resources.
> 
>   - **Behavior**:
>     - Adds `add_internal_firewalls` method to `KubernetesCluster` and `PostgresResource` classes to introduce internal firewalls.
>     - Transfers existing firewalls and subnets to the customer project for resources created before phase 3.
>   - **Kubernetes**:
>     - In `kubernetes_cluster.rb`, creates internal firewalls for control plane and worker VMs, updates firewall rules, and moves subnets to customer project.
>   - **Postgres**:
>     - In `postgres_resource.rb`, creates internal firewalls for Postgres resources, updates firewall rules, and moves subnets to customer project.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c86956eba65911f189c273fc827ae8eed5450a4d. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -21,6 +21,68 @@ class KubernetesCluster < Sequel::Model
   plugin SemaphoreMethods, :destroy, :sync_kubernetes_services, :upgrade, :install_metrics_server, :sync_worker_mesh, :install_csi, :update_billing_records
   include HealthMonitorMethods
 
+  # :nocov:
+  def self.add_internal_firewalls
+    # Do not use transaction around the update of all resources, to avoid blocking
+    all do |kc|
+      next if kc.internal_cp_vm_firewall
+
+      ubid = kc.ubid
+      print "Adding internal firewalls for #{ubid}..."
+      subnet = kc.private_subnet
+      location_id = kc.location_id
+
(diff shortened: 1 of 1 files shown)


jeremyevans:
> Do you think not locking is necessary? Seems like we are only working on the postgres/kubernetes resource objects, not strands. So, these objects should be OK to block in terms of performance penalty as long as there is not an active destroy/create operations. In case there is, you may even prefer to lock.

This uses a transaction around each change. That should lock the objects that are being modified while the change is made. If you are talking about a different kind of lock, can you be more specific?

furkansahin:
Oh yeah, since I read the comment `# Do not use transaction around the update of all resources, to avoid blocking`, my mind basically omitted the `DB.transaction` in the later part of the code. This is good to go

PR #4071, opened 2025-10-21
Add support for PG18

shikharbhardwaj:
## Description

Introduces PG18 support, along with a couple of related fixes/updates.

----

> [!IMPORTANT]
> Add support for PostgreSQL version 18, update configurations, and modify validation logic and CLI options accordingly.
> 
>   - **Behavior**:
>     - Adds support for PostgreSQL version 18 in `lib/option.rb` and `lib/validation/postgres_config_validator.rb`.
>     - Updates `validate_postgres_version` in `lib/validation.rb` to include version 18.
>     - Modifies `upgrade_stage` logic in `views/postgres/upgrade.erb` to handle new upgrade stages.
>   - **Migrations**:
>     - Adds migration `20251021_add_pg18.rb` to include version 18 in constraints.
>     - Adds migration `20251022_add_pg_18_amis.rb` to update AMI IDs for version 18.
>   - **CLI**:
>     - Updates CLI help files to include version 18 in `spec/routes/api/cli/golden-files/help -r.txt` and other related files.
>   - **Tests**:
>     - Updates tests in `spec/lib/validation_spec.rb` and `spec/model/postgres/postgres_resource_spec.rb` to validate version 18 support.
>     - Modifies `spec/routes/web/project/postgres_spec.rb` to test UI changes for version 18.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0e9a4888eab348aaaa484a7cf33199c6fea4ffd7. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/views/postgres/upgrade.erb
+++ b/views/postgres/upgrade.erb
@@ -1,13 +1,13 @@
 <% upgrade_stages = [
   # [stage, description, database_available?]
-  ["start", "Start", true],
-  ["provision_servers", "Wait for standby to be ready", true],
-  ["wait_for_maintenance_window", "Wait for maintenance window", true],
-  ["wait_fence_primary", "Wait to fence primary", false],
-  ["upgrade_standby", "Upgrade standby", false],
-  ["update_metadata", "Update metadata", false],
-  ["recycle_representative_server", "Wait for takeover", false],
-  ["prune_servers", "Prune old servers", true]
+  [["start"], "Start", true],
+  [["provision_servers", "wait_servers_to_be_ready"], "Wait for standby to be ready", true],
+  [["wait_for_maintenance_window"], "Wait for maintenance window", true],
+  [["wait_fence_primary"], "Wait to fence primary", false],
+  [["upgrade_standby"], "Upgrade standby", false],
+  [["update_metadata", "wait_upgrade_candidate"], "Update metadata", false],
+  [["recycle_representative_server"], "Wait for takeover", false],
+  [["prune_servers"], "Prune old servers", true]
 ]
 
 current_stage = @pg.upgrade_stage || "start"
@@ -17,7 +17,7 @@ upgrade_progress = upgrade_stages.map do |stage, description, database_available
   if stage_status == "running"
     stage_status = "pending"
   end
-  if stage == current_stage
+  if stage.include?(current_stage)
     stage_status = "running"
   end
 
@@ -37,7 +37,7 @@ upgrade_progress = upgrade_stages.map do |stage, description, database_available
     ""
   end
 
-  [stage, description, database_availability_icon, stage_status_icon]
+  [stage.first, description, database_availability_icon, stage_status_icon]
 end %>
 
 <div class="p-6">


spec/routes/web/project/postgres_spec.rb
```diff
@@ -919,7 +919,7 @@
         click_button "Save"
 
         expect(page).to have_flash_error "Validation failed for following fields: pg_config.work_mem"
-        expect(page).to have_content "must match pattern: ^[0-9]+(kB|MB|GB|TB)?$"
+        expect(page).to have_content "must match pattern: \\A[0-9]+(kB|MB|GB|TB)?\\z"
```
jeremyevans: Questionable whether we should expose ruby regexp specific metacharacters in user error messages. I think it would be best to remove the `\A` and `\z` when displaying the pattern.

lib/validation/postgres_config_validator.rb
```diff
@@ -8,6 +8,8 @@ def initialize(version)
         @config_schema = Validation::PostgresConfigValidatorSchema::PG_17_CONFIG_SCHEMA
       when "16"
         @config_schema = Validation::PostgresConfigValidatorSchema::PG_16_CONFIG_SCHEMA
+      when "18"
```
jeremyevans, line 11: Seems fine, but you may want to consider a hash-lookup approach instead of a case approach to reduce duplication.

migrate/20251017_remove_pg_version_enum.rb
```diff
@@ -0,0 +1,90 @@
...
+      set_column_not_null :target_version
+    end
+
+    alter_table(:postgres_server) do
+      add_constraint(:version_check, Sequel.lit("version IN ('16', '17')"))
+      set_column_not_null :version
+    end
+
+    # Drop the enum type
+    run "DROP TYPE IF EXISTS postgres_version"
```
jeremyevans: `IF EXISTS` should not be used in migrations. You know at this point that the type exists.

jeremyevans (commented):
Looks good. I'll leave approval to @byucesoy.

lib/validation/postgres_config_validator.rb
```diff
@@ -104,8 +104,8 @@ def validate_enum(value, allowed_values)
 
     def validate_string(value, pattern)
       # Check pattern validation first if pattern is specified
-      if pattern && !value.match?(Regexp.new(pattern))
-        return "must match pattern: #{pattern}"
+      unless pattern.nil? || pattern.match?(value)
```
byucesoy: Might be personal preference, but old version seemed more readable. Any reason for this change?

migrate/20251022_add_pg_18_amis.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    run <<~SQL
+    INSERT INTO pg_aws_ami (id, aws_location_name, arch, pg_version, aws_ami_id) VALUES
+      ('853b1491-1f0c-42db-8f38-c916994e32f4', 'us-west-2', 'x64', '18', 'ami-0090a7d48d9b50181'),
```
byucesoy, line 7: Given we have same image for all versions, we should drop version column from PgAwsAmi at some point. No need to change it now.

helpers/postgres.rb
```diff
@@ -121,7 +121,7 @@ def generate_postgres_options(flavor: nil, location: nil)
       end
     end
 
-    options.add_option(name: "version", values: Option::POSTGRES_VERSION_OPTIONS)
+    options.add_option(name: "version", values: Option::POSTGRES_VERSION_OPTIONS[flavor || PostgresResource::Flavor::STANDARD])
```
byucesoy: I think this should be:

```ruby
    options.add_option(name: "version", values: Option::POSTGRES_VERSION_OPTIONS.values.flatten.uniq, parent: "flavor") do |flavor, version|
      Option::POSTGRES_VERSION_OPTIONS[flavor].include?(version)
    end
```

Two important changes:
- We add flavor as parent, otherwise it wouldn't create the tree with correct hierarchy.
- Instead of filtering at the `values`, we do filtering at the block. Otherwise calling `generate_postgres_options` without passing `flavor` would put PG18 for all flavors. This is not a problem, currently because all callers set `flavor` right now.

byucesoy: FYI, after doing this, you would need to update version in ContentGenerator like this;

```ruby
    def self.version(flavor, version)
      "Postgres #{version}"
    end
```

PR #4073, opened 2025-10-22
QEMU hypervisor support for B200 GPU VMs

bsatzger:
- **Update GPU VM boot image to 20251017.1.0**
  

- **Add QEMU hypervisor support to data plane**
  This adds support for the QEMU hypervisor in the data plane (rhizome).
  Cloud Hypervisor remains the default, but QEMU can now be selected
  explicitly.
  

- **Default to QEMU for B200 GPU VMs**
  With QEMU support now available in rhizome, VMs that attach B200 GPUs
  will use QEMU as the hypervisor by default. All other VMs will continue
  to use Cloud Hypervisor unless specified otherwise.
  
  Cloud Hypervisor does not currently provide a way to attach PCI devices
  behind a root port, which appears to be required for NVIDIA B200 GPU
  passthrough. When the device is attached directly to the guest root
  complex (as Cloud Hypervisor does), the GPU fails to initialize properly.
  QEMU allows explicit topology configuration via `pcie-root-port`.
  
  This change also allows explicitly selecting the hypervisor when assembling
  a VM.
  

- **Limit GPU VMs to a maximum of 2 GPUs**
  Previously, VMs could be created with 1, 2, 4, or 8 GPUs. This change
  restricts GPU VM configurations to 1 or 2 GPUs.
  

----

> [!IMPORTANT]
> Adds QEMU hypervisor support for B200 GPU VMs, updates boot image, and limits GPU VMs to 2 GPUs.
> 
>   - **Behavior**:
>     - Default hypervisor for B200 GPU VMs set to QEMU in `vm.rb` and `nexus.rb`.
>     - Limits GPU VMs to 1 or 2 GPUs in `vm.rb`.
>     - Updates GPU VM boot image to `20251017.1.0` in `download_boot_image.rb`.
>   - **Implementation**:
>     - Adds QEMU support in `vm_setup.rb` with `build_qemu_service()`.
>     - Modifies `install_systemd_unit()` in `vm_setup.rb` to choose between QEMU and Cloud Hypervisor.
>     - Updates `params_json()` in `vm.rb` to set hypervisor based on GPU presence.
>   - **Testing**:
>     - Adds tests for QEMU service creation in `vm_setup_spec.rb`.
>     - Updates tests in `vm_spec.rb` and `nexus_spec.rb` to cover new hypervisor logic and GPU limits.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 58a204814ed6f48b909914b1f18790fdd9100eab. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/helpers/vm.rb
+++ b/helpers/vm.rb
@@ -128,7 +128,7 @@ class Clover
         .select_append { max(:count).as(:max_count) }
         .all.filter { !!BillingRate.from_resource_properties("Gpu", it[:device], it[:location_name]) }
 
-      gpu_counts = [1, 2, 4, 8]
+      gpu_counts = [1, 2]
       gpu_options = available_gpus.map { it[:device] }.uniq.flat_map { |x| gpu_counts.map { |i| "#{i}:#{x}" } }
       gpu_availability = available_gpus.each_with_object({}) do |entry, hash|
         hash[entry[:location_name]] ||= {}


rhizome/host/lib/vm_setup.rb
```diff
@@ -700,56 +686,55 @@ def install_systemd_unit(max_vcpus, cpu_topology, mem_gib, storage_params, nics,
...
 
-    net_params = nics.map { "--net mac=#{_1.mac},tap=#{_1.tap},ip=,mask=,num_queues=#{max_vcpus * 2 + 1}" }
-    pci_device_params =
-      if Gem::Version.new(@ch_version.version) >= Gem::Version.new("36")
-        pci_devices.empty? ? "" : " --device #{pci_devices.map { |dev| "path=/sys/bus/pci/devices/0000:#{dev[0]}/" }.join(" ")}"
-      else
-        pci_devices.map { |dev| " --device path=/sys/bus/pci/devices/0000:#{dev[0]}/" }.join
-      end
-    limit_memlock = pci_devices.empty? ? "" : "LimitMEMLOCK=#{mem_gib * 1073741824}"
+    limit_memlock = pci_devices.empty? ? "" : "LimitMEMLOCK=#{mem_gib * 1.25 * 1073741824}"
```
jeremyevans: I don't see a mention of the reason for this 25% increase. Can you add a comment that explains why it is chosen?

rhizome/host/lib/vm_setup.rb
```diff
@@ -649,7 +650,7 @@ def prepare_pci_devices(pci_devices)
   end
 
   def install_systemd_unit(max_vcpus, cpu_topology, mem_gib, storage_params, nics, pci_devices, slice_name, cpu_percent_limit)
-    cpu_setting = "boot=#{max_vcpus},topology=#{cpu_topology}"
+    fail "BUG" if /["'\s]/.match?(cpu_topology)
```
jeremyevans, line 657: Previous `fail` would also catch issues with `max_vcpus` in addition to `cpu_topology`. I would guess that's not an issue, though.

rhizome/host/lib/vm_setup.rb
```diff
@@ -770,4 +755,144 @@ def start_systemd_unit
...
+      end
+
+    net_params = nics.map { "--net mac=#{_1.mac},tap=#{_1.tap},ip=,mask=,num_queues=#{max_vcpus * 2 + 1}" }
+    pci_device_params =
+      if pci_devices.empty?
+        ""
+      elsif Gem::Version.new(@ch_version.version) >= Gem::Version.new("36")
+        "\\\n--device #{pci_devices.map { |dev| "path=/sys/bus/pci/devices/0000:#{dev[0]}/" }.join(" ")}"
+      else
+        "\\\n" + pci_devices.map { |dev| "--device path=/sys/bus/pci/devices/0000:#{dev[0]}/" }.join(" ")
```
jeremyevans: Can you explain the need for `\\\n`? I could understand it if `pci_device_params` were on the same line as `net_params.join(" ")` below, but it isn't.  Seems simpler to have `\\` after the `net_params.join(" ")` below, and not have `\\\n`, but maybe there is a reason it was done this way?

rhizome/host/lib/vm_setup.rb
```diff
@@ -770,4 +755,144 @@ def start_systemd_unit
...
+  ExecStartPre=/usr/bin/rm -f #{vp.ch_api_sock}
+
+  ExecStart=#{@ch_version.bin} -v \\
+  --api-socket path=#{vp.ch_api_sock} \\
+  --kernel #{@firmware_version.path} \\
+  #{disk_args} \\
+  --console off --serial file=#{vp.serial_log} \\
+  --cpus boot=#{max_vcpus},topology=#{cpu_topology} \\
+  --memory size=#{mem_gib}G,#{@hugepages ? "hugepages=on,hugepage_size=1G" : "shared=on"} \\
+  #{net_params.join(" ")}
```
jeremyevans: Without a `\\` at the end of this line, is the next line still treated as part of `ExecStart`?

jeremyevans (commented):
This was a bit challenging to review due to the mix of moved code and new code. Not sure if it would have been possible to split into one commit where you are moving code, and a separate commit where you adding new code/features, but if so, it would have made review easier.

rhizome/host/lib/vm_setup.rb
bsatzger: Without adding that buffer qemu vms failed to start with memory errors. I'm happy to add a comment.

rhizome/host/lib/vm_setup.rb
bsatzger: you are right, but I also don't think it should be an issue in this case. `max_vcpus` should be an integer and I think we'd fail before getting here if it isn't. `cpu_topology` seems more relevant for this check.

rhizome/host/lib/vm_setup.rb
bsatzger: yeah, it was on the same line but accidentally moved to a separate line. `pci_device_params` are empty except for the GPU case, i.e., empty most of the time. This code is pretty brittle and I made a small change here. Thanks!

rhizome/host/lib/vm_setup.rb
bsatzger: See previous comment.

bsatzger:
> This was a bit challenging to review due to the mix of moved code and new code. Not sure if it would have been possible to split into one commit where you are moving code, and a separate commit where you adding new code/features, but if so, it would have made review easier.

That should have been possible. I can still do it if not already too late to help with the review.

jeremyevans:
> That should have been possible. I can still do it if not already too late to help with the review.

I don't think it's worth the effort.

jeremyevans:
Changes look good to me. Assuming E2E passes, I'll approve, but considering the extent of the changes, I recommend a second reviewer. In terms of rollout, I assume we want to first rollout to the B200 machine, but we should also rollout to an AX162 or AX102 github-runners machine and see how that machine does before a broader rollout.

bsatzger:
> Changes look good to me. Assuming E2E passes, I'll approve, but considering the extent of the changes, I recommend a second reviewer. In terms of rollout, I assume we want to first rollout to the B200 machine, but we should also rollout to an AX162 or AX102 github-runners machine and see how that machine does before a broader rollout.

I'd first rollout to the B200 and a single runner machine, then to a few more runner machines (making sure we cover Ubuntu 22/24 and amd64/arm64), before a broader rollout.

fdr:
The commit messages (or at least one of them) don't say why.

fdr:
(I also don't know why, but if you were going bother typing it out to me were to ask, it probably would make good material for a commit message)

bsatzger:
> The commit messages (or at least one of them) don't say why.

Added a more detailed commit message.

PR #4077, opened 2025-10-23
Postgres/K8s Networking Change - Phase 6

jeremyevans:
This removes the use of PostgresFirewallRule, and has the postgres
firewall rule routes operate directly on the FirewallRules for the
customer firewall.

For the web UI, we just link to the customer firewall networking
page if the customer firewall exists, or to the private subnet
networking page otherwise.

For the API/CLI, this isn't 100% backwards compatible, as
PostgresFirewallRules have descriptions and FirewallRules do not,
and PostgreFirewallRule represents a combination of 2 FirewallRules.
Use the description field in the API to show the port information,
and only show FirewallRules that are for only 5432 or for only 6432.

This change avoids an issue where if a customer changes the firewall
rules for the postgres customer firewall, and then changes the
postgres firewall rules, the postgres firewall rules would replace
any changes made directly to the customer firewall.

Long term, we should probably deprecate and remove this API/CLI
support, and point customers to the related firewall if they want to
make networking changes.

----

> [!IMPORTANT]
> Removes `PostgresFirewallRule`, directly using `FirewallRules` for PostgreSQL networking, updating UI, API/CLI, and tests.
> 
>   - **Behavior**:
>     - Removes `PostgresFirewallRule` and uses `FirewallRules` directly for PostgreSQL networking.
>     - Updates web UI to link to customer firewall networking page if available, otherwise to private subnet page.
>     - API/CLI now shows `FirewallRules` for ports 5432 or 6432 only, using description for port info.
>   - **Models**:
>     - Deletes `PostgresFirewallRule` model.
>     - Updates `PostgresResource` to manage `FirewallRules` directly.
>   - **Tests**:
>     - Updates tests in `pg/add-firewall-rule_spec.rb`, `pg/delete-firewall-rule_spec.rb`, `pg/modify-firewall-rule_spec.rb`, and `pg/show_spec.rb` to reflect new firewall rule handling.
>     - Adjusts `project/location/postgres_spec.rb` and `web/project/postgres_spec.rb` for UI changes.
>   - **Misc**:
>     - Removes `hero-check-circle` icon from `icon.erb`.
>     - Deletes `inline_edit_buttons.erb` component.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for facb1c0c1822708a08d6c87719cbb6401a5e2f2d. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


jeremyevans:
I rebased this (and the PRs for phases 3-5) on top of #4801 so we don't lose the descriptions for the firewall rules.

views/postgres/networking.erb
```diff
@@ -1,68 +1,17 @@
...
-      <% unless has_firewall %>
-        <p class="text-gray-500 mt-2">PostgreSQL firewall was deleted, manage firewall rules using an appropriate firewall on the <a class="text-orange-600" href="<%= path(@pg.private_subnet) %>/networking">private subnet</a>.</p>
+      <p class="text-gray-500 mt-2">
+      <% if firewall = @pg.customer_firewall %>
+          You can 
+          <a class="text-orange-600" href="<%= path(firewall) %>/networking">manage firewall rules using the firewall for this PostgreSQL database</a>.
+      <% else %>
+          The firewall related to this PostgreSQL database was deleted or detached from the related private subnet,
+          you can manage firewall rules using an appropriate firewall on the
+          <a class="text-orange-600" href="<%= path(@pg.private_subnet) %>/networking">private subnet</a>.
```
furkansahin, line 12: since we are removing the pg specific firewalls from the UI. For the sake of UX, we might add some predefined port selection in the firewall creation page? Like, the port box stays edittable but we can pick a "Postgres" type that fills 5432 automatically, "pgbouncer", "ssh", "http", or "https". What do you think? It is not really something we have to solve in this PR but it might be a good addition later to increase usability.

jeremyevans: I don't think the benefit of doing that is worth making the UI more complicated. However, I'm not opposed to it if another developer wants to work on that and the UI doesn't get worse. We're basically out of horizontal space in the existing UI at ~1240px width (I generally browse at ~1260px width). You would have to start stacking inputs vertically. Doing it for existing rows means every row will start taking more space, which I don't think we want. Maybe we could do it only for the row to create new rules, though.

PR #4079, opened 2025-10-23
Add support for VM init scripts

jeremyevans:
This is a simpler approach than #3984, not supporting init script registration nor supporting init script args.

This is supported in the web UI as well as the API/SDK/CLI.

This requires changes to rhizome, so it will require a rhizome rollout to have an effect. Old rhizome will still work, but ignore the init script settings.

----

> [!IMPORTANT]
> Add support for VM init scripts in CLI, API, and web UI, with database and model updates for storage and retrieval.
> 
>   - **Behavior**:
>     - Adds support for VM init scripts in CLI (`create.rb`), API (`openapi.yml`), and web UI (`create.erb`).
>     - Init scripts are optional and can be specified during VM creation.
>     - Requires rhizome rollout; old rhizome ignores init scripts.
>   - **Models**:
>     - Adds `VmInitScript` model in `vm_init_script.rb`.
>     - Updates `Vm` model in `vm.rb` to include `init_script` association.
>   - **Database**:
>     - New migration `20251023_vm_init_script.rb` to create `vm_init_script` table.
>   - **Tests**:
>     - Updates tests in `vm_spec.rb`, `create_spec.rb`, and `vm_spec.rb` to cover init script functionality.
>   - **Misc**:
>     - Updates help and usage examples in CLI help files (`help -r vm.txt`, `help -r.txt`).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 93a90668b78c23eb7f95258fb0a8ef5ae2856212. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/helpers/vm.rb
+++ b/helpers/vm.rb
@@ -210,6 +210,7 @@ class Clover
     options.add_option(name: "unix_user")
     options.add_option(name: "ssh_public_key", values: @project.ssh_public_keys)
     options.add_option(name: "public_key")
+    options.add_option(name: "init_script")
 
     options.serialize
   end
--- a/spec/routes/web/vm_spec.rb
+++ b/spec/routes/web/vm_spec.rb
@@ -215,6 +215,28 @@ RSpec.describe Clover, "vm" do
         expect(Vm.first.public_key).to eq "a a"
       end
 
+      it "can create new virtual machine using init script" do
+        visit "#{project.path}/vm/create"
+
+        expect(page.title).to eq("Ubicloud - Create Virtual Machine")
+        name = "dummy-vm"
+        fill_in "Name", with: name
+        fill_in "SSH Public Key", with: "a a"
+        fill_in "Init Script", with: "foo bar"
+        choose option: Location::HETZNER_FSN1_UBID
+        choose option: "ubuntu-jammy"
+        choose option: "standard-2"
+
+        click_button "Create"
+
+        expect(page.title).to eq("Ubicloud - #{name}")
+        expect(page).to have_flash_notice("'#{name}' will be ready in a few minutes")
+        expect(Vm.count).to eq(1)
+        expect(Vm.first.project_id).to eq(project.id)
+        expect(Vm.first.private_subnets.first.id).not_to be_nil
+        expect(Vm.first.init_script.script).to eq "foo bar"
+      end
+
       it "can create new virtual machine with public ipv4" do
         project
 
--- a/views/vm/create.erb
+++ b/views/vm/create.erb
@@ -67,6 +67,7 @@ current_usage = @project.current_resource_usage("VmVCpu")
   end
 
   form_elements.push(public_key_params)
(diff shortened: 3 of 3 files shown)


rhizome/host/bin/setup-vm
```diff
@@ -68,6 +68,7 @@ begin
   slice_name = params.fetch("slice_name", "system.slice")
   cpu_percent_limit = params.fetch("cpu_percent_limit", 0)
   cpu_burst_percent_limit = params.fetch("cpu_burst_percent_limit", 0)
+  init_script = params.fetch("init_script", "")
```
fdr, line 72: In the commit message of `Allow rhizome to handle VM init scripts in setup-vm`, it's not `run_cmd` but `runcmd`.

migrate/20251023_vm_init_script.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:vm_init_script) do
+      foreign_key :id, :vm, type: :uuid, primary_key: true
+      column :script, String, null: false
```
fdr: we could consider setting a limit for this at the database level. AWS sets 16KB, Azure 64KB, GCP 256KB with an out-of-line option using GCS, Oracle likewise 16KB. Personally...I think 4KB is adequate, and AWS's limit comes from the dark ages, and they were never going to revise it downwards. Maybe we should push the envelope and do `size: 4096`.

That said...why 4096? The page-sized quantity actually does nothing for us.

Another valid option would be to limit to `2036` characters (or round down to `2000`): accounting for varlena type overhead in Postgres, this means every value would have a chance to be stored in-line in the heap tuple, avoiding a TOAST out-of-line read, which is a nice bonus because this table is already a special out-of-line storage consideration. because this is denominated in characters, not bytes, it doesn't quite hold for non-ascii text, unless you were to do this as a bytea, which adds some tedium and chance for encoding hijinks late in the program.

Here's a distribution from the clover source directory, knowing that mostly files are way more complicated than what you want in an init script...I think.

```
Files < 2036 bytes: 711
Files < 4096 bytes: 895
Total files: 1162
```

fdr:
```suggestion
      column :script, String, size: 2000, null: false
```

pushing the envelope on smallness...and toastyness.

fdr (approved):

rhizome/host/bin/setup-vm
jeremyevans: Updated the commit message.

migrate/20251023_vm_init_script.rb
jeremyevans: Having a limit makes sense. I highly doubt that the TOAST/non-TOAST different matters at all considering this is only used in a single place and is already separated into its own table. However, 2KB seems like enough for an init script, so I merged your proposed change. We can always increase the limit if we need to.

PR #4080, opened 2025-10-23
Create Vm billing record before prep

jeremyevans:
This creates the billing record between create_unix_user and prep, instead of between wait_sshable and wait.

With support for init scripts, VMs will be able to run arbitrary code during prep, and that usage should be billed. There were historical reasons for creating the billing record late, but those reasons no longer apply.

Move the updating of the VM display state and Clog emit from create_billing_record to wait_sshable.

----

> [!IMPORTANT]
> Move VM billing record creation to occur before the 'prep' stage to ensure all usage is billed, and update related logic and tests.
> 
>   - **Behavior**:
>     - Move billing record creation from `wait_sshable` to `create_unix_user` in `nexus.rb`.
>     - Update VM display state and emit logs in `wait_sshable` instead of `create_billing_record`.
>   - **Logic**:
>     - Ensure billing records are created before `prep` to account for init script usage.
>     - Add checks in `create_billing_record` to handle cases where billing records were not created before `prep`.
>   - **Tests**:
>     - Update tests in `nexus_spec.rb` to reflect changes in billing record creation timing and logic.
>     - Add test cases to ensure correct hopping behavior based on billing record creation status.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 38eaa959b7d85ea81da8042a967699e6c5f201d6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/metal/nexus.rb
+++ b/prog/vm/metal/nexus.rb
@@ -134,7 +134,7 @@ class Prog::Vm::Metal::Nexus < Prog::Base
 
     host.sshable.cmd(command, vm_name:, vm_home:, uid:)
 
-    hop_prep
+    hop_create_billing_record
   end
 
   label def prep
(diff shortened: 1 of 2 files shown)


fdr:
We could probably add the historical reason to the commit message.

Here's a proposal (feel encouraged to edit it):

> Originally designed in 2023 (9eb482b3395e6aa3e429882816e14fd0294231b2), billing records were created only after VM startup because provisioning was unreliable. This avoided charging customers for broken instances and reduced the need to issue manual credits. VM provisioning is now reliable, and the system’s ongoing churn demands that we maintain that reliability continuously.

jeremyevans:
I was about to merge this, but I realized it could cause us to miss creating billing records for VMs currently in prep, clean_prep, or wait_sshable. It could also result in VMs currently in create_billing_record to go through prep, clean_prep, and wait_sshable again, which would be bad as prep is not reentrant. This needs to be broken into phases to avoid those issues.

spec/prog/vm/nexus_spec.rb
```diff
@@ -769,61 +769,67 @@ def requested_disk_size(st)
...
 
-    it "hops to create_billing_record if sshable" do
+    it "hops to wait if sshable and billing record created" do
       expect(vm).to receive(:update_firewall_rules_set?).and_return(true)
       expect(vm).not_to receive(:incr_update_firewall_rules)
       vm_addr = instance_double(AssignedVmAddress, id: "46ca6ded-b056-4723-bd91-612959f52f6f", ip: NetAddr::IPv4Net.parse("10.0.0.1"))
       expect(vm).to receive(:assigned_vm_address).and_return(vm_addr).at_least(:once)
       expect(Socket).to receive(:tcp).with("10.0.0.1", 22, connect_timeout: 1)
-      expect { nx.wait_sshable }.to hop("create_billing_record")
+      now = Time.now
```
enescakir: It has conflicts with my nexus_spec mock removal PR. Sorry for that

enescakir (approved):
When a VM host becomes unavailable, we end up with multiple VMs stuck in `wait_sshable`. This usually happens for runner VMs under high load, and since we don’t create billing records for them, it’s harmless.

It’s very rare for customer VMs, but if it does happen, we should consider giving the customer some credit for the unused time.

jeremyevans:
I had to redo this entire PR manually because of the Metal/AWS split. The new commit only makes this change for Metal VMs. The impetus for this PR is VM init scripts, but I'm not sure if we even support VM init scripts for AWS VMs. If we do, I'm not sure where.

During the AWS/Metal split, the `strand.stack[-1]["create_billing_record_done"] = true` was copied to `create_billing_record` in the AWS prog, but the `strand.stack[-1]["prep_done"] = true` was not copied to the AWS prog. So in terms of AWS, phase 1 of this change was only partially implemented.

Do we want to keep create billing record after wait_sshable in AWS? If so, we can remove the `strand.stack[-1]["create_billing_record_done"] = true`  code. If we want create billing record to be earlier for AWS as well, we need to determine where it should go. Maybe between `create_instance` and `wait_instance_created`? However, unless we support init scripts for AWS VMs, I think we should keep billing record creation where it currently is.

@fdr @furkansahin Let me know what you think I should do here.

jeremyevans:
E2E passed with this, so I plan to deploy this today.

PR #4085, opened 2025-10-24
Disallow changing shared memory configs

shikharbhardwaj:
Given that we setup hugepages based on a fixed ratio of the total memory before PG starts, changing shared buffers can cause problems.

----

> [!IMPORTANT]
> Remove shared memory configuration options from PostgreSQL validator to prevent issues with hugepages setup.
> 
>   - **Behavior**:
>     - Removed `shared_buffers`, `huge_page_size`, and `huge_pages` from `PostgresConfigValidatorSchema` to prevent changes to shared memory configurations.
>     - Updated `restart` method in `postgres_server_nexus.rb` to include a 10-minute wait deadline.
>   - **Tests**:
>     - Removed test for `shared_buffers` in `postgres_config_validator_spec.rb`.
>     - Updated `restart` test in `postgres_server_nexus_spec.rb` to check for deadline registration.
>     - Modified API tests in `postgres_spec.rb` to reflect removal of `huge_pages` and `shared_buffers` configurations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 4297c98e96276e4e1f388d8c8279e86e74a0e007. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -656,6 +656,9 @@ SQL
 
   label def restart
     decr_restart
+
+    register_deadline("wait", 10 * 60)
+
     vm.sshable.cmd("sudo postgres/bin/restart #{postgres_server.version}")
     vm.sshable.cmd("sudo systemctl restart pgbouncer@*.service")
     pop "postgres server is restarted"
--- a/spec/prog/postgres/postgres_server_nexus_spec.rb
+++ b/spec/prog/postgres/postgres_server_nexus_spec.rb
@@ -1032,9 +1032,10 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
   end
 
   describe "#restart" do
-    it "restarts and exits" do
+    it "sets deadline, restarts and exits" do
       expect(sshable).to receive(:cmd).with("sudo postgres/bin/restart 16")
       expect(sshable).to receive(:cmd).with("sudo systemctl restart pgbouncer@*.service")
+      expect(nx).to receive(:register_deadline).with("wait", 10 * 60)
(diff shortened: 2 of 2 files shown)


jeremyevans (commented):
There seem to be a lot of configuration values that affect memory usage. (e.g. `shared_memory_type`, `dynamic_shared_memory_type`, `min_dynamic_shared_memory`). If we are removing access to these three, are we sure that are no remaining config settings that can cause similar problems? 

Basically, I think it we are attempting to remove customer config setting footguns, we should remove all of them, not just the ones we think will be most common.  This would require an analysis of every supported config setting, though.

shikharbhardwaj:
@jeremyevans yes, there is work remaining to make this safer and also allow customers to recover from bad configuration settings on their own which will come as a follow-up. Given that we've seen these specific settings come up a couple of times already, this commit will help avoid repeating the same while the general fix is completed.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -653,6 +653,13 @@ def before_run
 
   label def restart
     decr_restart
+
+    if postgres_server.primary?
```
byucesoy: Why do we use different deadlines for primary and standby?

shikharbhardwaj: I took the same pattern we use for setting deadlines in `bootstrap_rhizome` here

byucesoy: There, we use longer deadline for standbys, because the provisioning takes more time due to recovery from backups + WAL replay. In this case, I think we can use the same duration.

PR #4088, opened 2025-10-24
Add account renaming functionality

enescakir:
# !! It doesn't have tests yet !!

- **Combine account password change and login pages**
  Previously, we had separate pages for changing passwords and logging in,
  which was unnecessary. This change combines them into a single unified
  page for a simpler user experience.
  
  I plan to add account renaming functionality to this page in a future
  commit.
  

- **Add path helper to account model**
  

- **Do not check authorization for renaming if it's passed as nil explicitly**
  We don't need to check authorization for all cases such as renaming
  accounts.
  

- **Skip storing audit logs when no project context is available**
  Some actions, such as renaming accounts, call the audit_log helper
  without a project context. In these cases, there's no relevant project
  information to log, so we now skip storing audit logs when the project
  context is missing.
  

- **Add account renaming functionality**
  Thanks to @jeremyevans's renaming helpers, it was straightforward to
  implement.
  
account_rename

### DIFF
--- a/clover.rb
+++ b/clover.rb
@@ -111,6 +111,7 @@ class Clover < Roda
   ].each { path(it, class_name: true, &under_project_path) }
 
   path("Project", class_name: true, &:path)
+  path("Account", class_name: true) { |_| "/account" }
   path("GithubInstallation", class_name: true) { "#{it.project.path}/github/#{it.ubid}" }
   path("Invoice", class_name: true) { "#{it.project.path}/billing#{it.path}" }
 
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -24,7 +24,11 @@ class Clover < Roda
     def rename(object, perm:, serializer:, template_prefix:)
       post "rename" do
         scope.instance_exec do
-          authorize(perm, object)
+          if perm
+            authorize(perm, object)
+          else
+            no_authorization_needed
+          end
           handle_validation_failure("#{template_prefix}/show") { @page = "settings" }
           name = typecast_body_params.nonempty_str!("name")
 
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -37,6 +37,8 @@ class Clover < Roda
           else
             if object.is_a?(KubernetesCluster)
               Validation.validate_kubernetes_name(name)
+            elsif object.is_a?(Account)
+              Validation.validate_account_name(name)
             else
               Validation.validate_name(name)
             end
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -134,6 +134,9 @@ class Clover < Roda
     return unless LOGGED_ACTIONS.include?(action) || Config.test?
     # :nocov:
 
+    # log only if we have a project context
+    return unless @project
+
     project_id = @project.id
     subject_id = current_account.id
     ubid_type = object.class.ubid_type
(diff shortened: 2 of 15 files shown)


enescakir:
Because of the first commit, authentication tests are failing with the following exception. @jeremyevans, how can I fix this issue? Architecturally, is the first commit incorrect?

```
8) Clover auth authenticated can change password when password entry is not required
     Failure/Error: Clover.app.call(env)

     NoMethodError:
       undefined method 'params' for #<Clover::RodaRequest:0x000000012a912eb0>
     # ./coverage/views/components-rodauth-login_field.erb.rb:14:in 'Tilt::CompiledTemplates#__tilt_2640'
     # ./coverage/views/account-settings.erb.rb:25:in 'block in Tilt::CompiledTemplates#__tilt_2640'
     # ./coverage/views/account-settings.erb.rb:19:in 'Tilt::CompiledTemplates#__tilt_2640'
     # ./routes/account.rb:15:in 'block (3 levels) in <class:Clover>'
     # ./routes/account.rb:11:in 'block (2 levels) in <class:Clover>'
     # ./routes/account.rb:5:in 'block in <class:Clover>'
     # ./clover.rb:1019:in 'block in <class:Clover>'
     # ./spec/routes/spec_helper.rb:64:in 'block (3 levels) in <top (required)>'
     # ./spec/routes/spec_helper.rb:63:in 'block (2 levels) in <top (required)>'
     # ./spec/routes/web/auth_spec.rb:447:in 'block (4 levels) in <top (required)>'
```

jeremyevans:
> Because of the first commit, authentication tests are failing with the following exception. @jeremyevans, how can I fix this issue? Architecturally, is the first commit incorrect?

For catch use of untype-checked parameters in our templates, we undefine `params` during non-frozen testing. This breaks `rodauth.params`. This isn't usually a problem, because we don't undefine it until after `r.rodauth`. However, you are now loading a rodauth-specific template after `r.rodauth`.

To workaround this issue, you can change `rodauth.param('login')` to `typecast_params.str('login')` in the `login_field.erb` template (technically, `rodauth.param` returns an empty string instead of nil if the parameter is not submitted, but that shouldn't matter in this case).

clover.rb
```diff
@@ -111,6 +111,7 @@ class Clover < Roda
   ].each { path(it, class_name: true, &under_project_path) }
 
   path("Project", class_name: true, &:path)
+  path("Account", class_name: true, &:path)
```
jeremyevans: Considering this is static and not dynamic, I would not recommend a model method. Maybe:

```suggestion
  path("Account", class_name: true)  { |_| '/account' } 
```

model/account.rb
```diff
@@ -15,6 +15,10 @@ class Account < Sequel::Model(:accounts)
 
   alias_method :admin_label, :email
 
+  def path
+    "/account"
+  end
+
```
jeremyevans:
```suggestion
```

jeremyevans (commented):
I don't object to the combining of the change login and change login pages, but I don't think it makes things simpler. Changing login and changing password are generally done for different reasons, so putting them on the same page doesn't result in a simpler user experience to me. It also seems unrelated to account renaming. However, it doesn't cause harm.

PR #4099, opened 2025-10-28
Migrate SPDK disks to ubiblk

mohi-kalantari:
Add script for generating vhost backend conf for SPDK vms

-----------

Add script for setting up vhost backend systemd unit

-----------

**Add Prog for migrating SPDK disks to Ubiblk**

This prog first attempts to stop the VM and its spdk controller, creating the required configs for runnning the migration.

The migration happens using a prebuilt binary from the ubiblk-migrate project. It utilizes damonizer2 to run the migration since it takes a long time to convert the disk.

After finishing up the migration, it starts the VM again and sets the Vms strand back to the wait label.

----

> [!IMPORTANT]
> Adds a program to migrate SPDK disks to Ubiblk with configuration scripts and comprehensive tests.
> 
>   - **Behavior**:
>     - Adds `Prog::Storage::MigrateSpdkVmToUbiblk` in `migrate_spdk_vm_to_ubiblk.rb` to migrate SPDK disks to Ubiblk.
>     - Uses a prebuilt binary from `ubiblk-migrate` for disk conversion.
>     - Handles VM state transitions: stops VM, removes SPDK controller, generates vhost config, downloads binaries, performs migration, and restarts VM.
>     - Validates VM and host configurations before migration.
>   - **Scripts**:
>     - Adds `convert-encrypted-dek-to-vhost-backend-conf` for converting DEK to vhost config.
>     - Adds `spdk-migration-helper` for SPDK controller management and service file creation.
>   - **Testing**:
>     - Comprehensive RSpec tests in `migrate_spdk_vm_to_ubiblk_spec.rb` for all migration steps and edge cases.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 857ddc02e1b904d840f07bb5eecb5fd7049cc285. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/storage/migrate_spdk_vm_to_ubiblk.rb
+++ b/prog/storage/migrate_spdk_vm_to_ubiblk.rb
@@ -122,19 +122,16 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
     # We do not want to store the KEK permenantly on the disk since it defeats the purpose,
     # so we write it to a pipe and ubiblk systemd unit will read its content once and the kek is gone.
     write_kek_pipe
-    hop_start_vm
-  end
-
-  label def start_vm
-    vm.vm_host.sshable.cmd("sudo systemctl start #{vm.inhost_name}")
     hop_update_vm_model
   end
 
   label def update_vm_model
-    DB.transaction do
-      vm.strand.update(label: "wait")
-      vm.vm_storage_volumes.first.update(use_bdev_ubi: false, vhost_block_backend_id: vm_host_vhost_block_backend.id, vring_workers: 1, spdk_installation_id: nil)
-    end
+    vm.vm_storage_volumes.first.update(
+      use_bdev_ubi: false,
+      vhost_block_backend_id: vm_host_vhost_block_backend.id,
+      vring_workers: [1, vm.vcpus / 2].max,
+      spdk_installation_id: nil
+    )
 
     hop_update_prep_json_file
   end
@@ -144,6 +141,13 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
     prep_json["storage_volumes"][0]["vhost_block_backend_version"] = Config.vhost_block_backend_version
     prep_json["storage_volumes"][0]["spdk_version"] = nil
     vm.vm_host.sshable.cmd("sudo tee /vm/#{vm.inhost_name}/prep.json >/dev/null", stdin: JSON.pretty_generate(prep_json))
+
+    hop_start_vm
+  end
+
+  label def start_vm
+    vm.vm_host.sshable.cmd("sudo systemctl start #{vm.inhost_name}")
+    vm.strand.update(label: "wait")
 
     pop "Vm successfully migrated to ubiblk"
   end
--- a/spec/prog/storage/migrate_spdk_vm_to_ubiblk_spec.rb
+++ b/spec/prog/storage/migrate_spdk_vm_to_ubiblk_spec.rb
@@ -182,22 +182,12 @@ RSpec.describe Prog::Storage::MigrateSpdkVmToUbiblk do
       unit_name = "#{vm.inhost_name}-0-storage.service"
       expect(vm.vm_host.sshable).to receive(:cmd).with("sudo systemctl start #{unit_name}")
       expect(prog).to receive(:write_kek_pipe)
-      expect { prog.start_ubiblk_systemd_unit }.to hop("start_vm")
-    end
-  end
-
-  describe "#start_vm" do
-    it "starts the vm" do
-      expect(vm.vm_host.sshable).to receive(:cmd).with("sudo systemctl start #{vm.inhost_name}")
-      expect { prog.start_vm }.to hop("update_vm_model")
+      expect { prog.start_ubiblk_systemd_unit }.to hop("update_vm_model")
     end
   end
 
   describe "#update_vm_model" do
     it "updates the vm to be identified as a ubiblk vm" do
-      st = instance_double(Strand)
-      expect(vm).to receive(:strand).and_return(st)
-      expect(st).to receive(:update).with(label: "wait")
       expect(vm.vm_storage_volumes.first).to receive(:update).with(use_bdev_ubi: false, vhost_block_backend_id: vm_host.vhost_block_backends.first.id, vring_workers: 1, spdk_installation_id: nil)
       expect { prog.update_vm_model }.to hop("update_prep_json_file")
     end
@@ -207,7 +197,17 @@ RSpec.describe Prog::Storage::MigrateSpdkVmToUbiblk do
     it "update the prep json file for proper cleanup of the vm later" do
       expect(vm.vm_host.sshable).to receive(:cmd).with("sudo cat /vm/#{vm.inhost_name}/prep.json").and_return({"storage_volumes" => [{"vhost_block_backend_version" => nil, "spdk_version" => "v.0.1.2"}]}.to_json)
       expect(vm.vm_host.sshable).to receive(:cmd).with("sudo tee /vm/#{vm.inhost_name}/prep.json >/dev/null", stdin: JSON.pretty_generate({"storage_volumes" => [{"vhost_block_backend_version" => "v0.2.1", "spdk_version" => nil}]}))
-      expect { prog.update_prep_json_file }.to exit({"msg" => "Vm successfully migrated to ubiblk"})
+      expect { prog.update_prep_json_file }.to hop("start_vm")
+    end
+  end
+
+  describe "#start_vm" do
+    it "starts the vm and exits" do
+      st = instance_double(Strand)
+      expect(vm).to receive(:strand).and_return(st)
+      expect(st).to receive(:update).with(label: "wait")
+      expect(vm.vm_host.sshable).to receive(:cmd).with("sudo systemctl start #{vm.inhost_name}")
+      expect { prog.start_vm }.to exit({"msg" => "Vm successfully migrated to ubiblk"})
     end
   end
 end


prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -0,0 +1,184 @@
...
+      fail "Vm does not exist"
+    end
+
+    unless vm.vm_storage_volumes.count == 1
+      fail "This prog only supports Vms with exactly one disk"
+    end
+
+    unless vm.vm_storage_volumes.first.use_bdev_ubi
+      fail "Vm is already using Ubiblk"
+    end
```
pykello: nit: Checking for `if vm.vm_storage_volumes.first.vhost_block_backend_id` is a more direct way to check if the volume is using ubiblk.

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -0,0 +1,184 @@
...
+      prog: "Storage::MigrateSpdkVmToUbiblk",
+      label: "stop_vm",
+      stack: [{
+        "subject_id" => vm_id
+      }]
+    )
+  end
+
+  label def stop_vm
+    vm.incr_stop
```
pykello, line 34: Maybe also `register_deadline(nil, 60 * 60)` so we get a page and don't forget about this if the migration fails.

rhizome/host/bin/spdk-migration-helper
```diff
@@ -0,0 +1,33 @@
...
+end
+
+if action == "create-vhost-backend-service-file"
+  vm_name = params.fetch("vm_name")
+  StorageVolume.new(vm_name, params).vhost_backend_create_service_file
+
+elsif action == "remove-spdk-controller"
+  vm_name = params.fetch("vm_name")
+  disk_index = params.fetch("disk_index")
+  client_rpc = SpdkRpc.new(SpdkPath.rpc_sock("v23.09-ubi-0.3"))
```
pykello: We shouldn't hard code the SPDK version. Some prod VMs use older versions of SPDK.

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
```diff
@@ -0,0 +1,177 @@
...
+    "path" => "/var/storage/#{vm_name}/0/disk.raw",
+    "num_queues" => 1,
+    "queue_size" => 256,
+    "seg_size_max" => 65536,
+    "seg_count_max" => 4,
+    "copy_on_read" => false,
+    "poll_queue_timeout_us" => 1000,
+    "device_id" => "#{vm_name}_0",
+    "skip_sync" => false,
+    "write_through" => false
```
pykello: To keep flush latency low, I think it's better if calculate `write_through` the same way as we do for new ubiblk volumes.

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
```diff
@@ -0,0 +1,181 @@
...
+def validate_kek_file(data)
+  unless data.is_a?(Hash)
+    raise ArgumentError, "KEK file must be a dictionary"
+  end
+  data.each do |key, value|
+    unless key.is_a?(String)
+      raise ArgumentError, "KEK file keys must be strings"
+    end
+    unless value.is_a?(Hash)
+      raise ArgumentError, "KEK file values must be dictionaries"
```
jeremyevans: This isn't Python :)

```suggestion
      raise ArgumentError, "KEK file values must be hashes"
```

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
```diff
@@ -0,0 +1,177 @@
...
+  decipher.decrypt
+  decipher.key = Base64.decode64(kek_data["key"])
+  decipher.iv = Base64.decode64(kek_data["init_vector"])
+  decipher.auth_data = kek_data["auth_data"]
+
+  auth_tag = encrypted_key[1]
+  decoded_auth_tag = Base64.decode64(auth_tag)
+
+  # We reject if auth_tag length is not within acceptable range for GCM
+  fail "Invalid auth_tag size: #{decoded_auth_tag.bytesize}" unless (12..24).cover(decoded_auth_tag.bytesize)
```
jeremyevans: I see this was fixed in a later commit, but in general it's better to use git rebase so it doesn't stick out like an obvious issue when doing a commit-by-commit review.

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
```diff
@@ -0,0 +1,177 @@
...
+  key1_wrapped_b64 = wrap_key_b64(key1, kek_data)
+  key2_wrapped_b64 = wrap_key_b64(key2, kek_data)
+  config["encryption_key"] = [key1_wrapped_b64, key2_wrapped_b64]
+  config
+end
+
+def write_vhost_conf(output_file, conf)
+  yaml = YAML.dump(conf)
+  # remove quotes around base64 strings in encryption_key
+  yaml.gsub!(/- "(.*?)"$/, '- \1')
```
jeremyevans, line 119: Can you explain why you want to do that?

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
```diff
@@ -0,0 +1,177 @@
...
+    raise ArgumentError, "Data encryption key file does not exist: #{options[:encrypted_dek_file]}"
+  end
+  unless File.exist?(options[:kek_file])
+    raise ArgumentError, "Key encryption key file does not exist: #{options[:kek_file]}"
+  end
+  dek_data = JSON.parse(File.read(options[:encrypted_dek_file]))
+  validate_encrypted_dek_file(dek_data)
+  kek_data = JSON.parse(File.read(options[:kek_file]))
+  validate_kek_file(kek_data)
+  convert(dek_data, kek_data, options[:output_file], options[:vm_name])
```
jeremyevans: Might be better to move this after the begin/rescue, and remove the `rescue => e` code. The default behavior for an unhandled exception would be to print an exception message, backtrace, and exit 1.

rhizome/host/bin/spdk-migration-helper
```diff
@@ -0,0 +1,34 @@
...
+  exit 1
+end
+
+params = JSON.parse($stdin.read)
+begin
+  %w[vm_name device disk_index vhost_block_backend_version slice max_read_mbytes_per_sec max_write_mbytes_per_sec encrypted spdk_version].each { |key| params.fetch(key) }
+rescue KeyError => e
+  puts "Needed #{e.key} in parameters"
+  exit 1
+end
```
jeremyevans:
```suggestion
%w[vm_name device disk_index vhost_block_backend_version slice max_read_mbytes_per_sec max_write_mbytes_per_sec encrypted spdk_version].each { |key| 
  unless params.has_key?(key)
    puts "Needed #{key} in parameters"
    exit 1
  end
end
```

rhizome/host/bin/spdk-migration-helper
```diff
@@ -0,0 +1,34 @@
...
+
+if action == "create-vhost-backend-service-file"
```
jeremyevans:
```suggestion
case action
when "create-vhost-backend-service-file"
```

rhizome/host/bin/spdk-migration-helper
```diff
@@ -0,0 +1,34 @@
...
+  vm_name = params.fetch("vm_name")
+  StorageVolume.new(vm_name, params).vhost_backend_create_service_file
+
+elsif action == "remove-spdk-controller"
```
jeremyevans:
```suggestion
when "remove-spdk-controller"
```

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -0,0 +1,186 @@
...
+    vm.vm_host.sshable.cmd("sudo host/bin/spdk-migration-helper create-vhost-backend-service-file", stdin: migration_script_params)
+    hop_start_ubiblk_systemd_unit
+  end
+
+  label def start_ubiblk_systemd_unit
+    disk_index = 0
+    unit_name = "#{vm.inhost_name}-#{disk_index}-storage.service"
+    vm.vm_host.sshable.cmd("sudo systemctl enable #{unit_name}")
+    vm.vm_host.sshable.cmd("sudo systemctl start #{unit_name}")
+    write_kek_pipe
```
jeremyevans, line 124: If this is needed, please add a comment explaining what will be reading from the kek pipe at this point, since that isn't obvious to me.

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -0,0 +1,186 @@
...
+
+  label def create_prep_json_file
+    vm.vm_host.sshable.cmd("sudo tee /vm/#{vm.inhost_name}/prep.json >/dev/null", stdin: vm.params_json)
+    pop "Vm successfully migrated to ubiblk"
+  end
+
+  def migration_script_params
+    {
+      "slice" => vm.vm_host_slice.name,
+      "device" => vm.vm_storage_volumes.first.storage_device.name,
```
jeremyevans: Create a local variable for `vm.vm_storage_volumes.first` and use it to DRY things up in this method.

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -0,0 +1,186 @@
...
+  def kek_file_path
+    "#{root_dir_path}kek.pipe"
+  end
+
+  def vhost_conf_path
+    "#{root_dir_path}vhost-backend.conf"
+  end
+
+  def write_kek_pipe
+    vm_storage_volume = vm.vm_storage_volumes.first
```
jeremyevans: Probably better to do `kek = vm.vm_storage_volumes.first.key_encryption_key_1`, and use that in the rest of the method.

prog/vm/nexus.rb
```diff
@@ -492,7 +492,9 @@ def write_params_json
       host.sshable.cmd("sudo systemctl stop #{q_vm}")
     end
     decr_stop
-
+    when_start_after_host_reboot_set? do
```
jeremyevans: This seems risky to me. Let's say you start a spdk->ubiblk migration. It moves the VM to stopped (this label). If the host is rebooted before the migration is finished, it will attempt start the VM even though the migration is not finished. Are you sure that's what we want? It's seems unlikely to start the VM correctly if the migration hasn't finished. Maybe you should add a new semaphore to move from stopped to start after host reboot?

Additionally the commit message states that you will `incr_reboot` the VmHost after migration, but that seems undesirable as there could be other VMs running on the host, not just the VM you are migrating. I don't see the `incr_reboot` change in this PR, so maybe the commit message is inaccurate.

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
mohi-kalantari: Sometimes key strings do not start with characters or numbers, as a result when encoded, they are wrapped in double quotes, breaking the migration binary because it expects it to have a certain length.

The migration binary can be improved to address such encoding scenarios better but since it's C code and wanted to keep it minimal, I thought I can tweak it here.

prog/vm/nexus.rb
mohi-kalantari: Fair point. Thanks for pointing it out. I'll think more and come back.

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
jeremyevans: Sounds reasonable. Maybe a comment explaining this is necessary due to limitations in the migration binary?

prog/vm/nexus.rb
mohi-kalantari: I added a new semaphore named resume which we need to manually increment in order to have more control

rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
```diff
@@ -0,0 +1,188 @@
...
+    raise ArgumentError, "DEK file must contain 'key' field as an array with exactly two strings"
+  end
+  unless data.key?("key2") && data["key2"].is_a?(Array) && data["key2"].length == 2 && data["key2"].all?(String)
+    raise ArgumentError, "DEK file must contain 'key2' field as an array with exactly two strings"
+  end
+end
+
+def validate_kek_file(data)
+  unless data.is_a?(Hash)
+    raise ArgumentError, "KEK file must be a dictionary"
```
jeremyevans:
```suggestion
    raise ArgumentError, "KEK file must be a hash"
```

PR #4104, opened 2025-10-29
Support adding a firewall rule using private subnet instead of cidr in the web UI and API/SDK/CLI

jeremyevans:
When adding using private subnet name or id, this creates 2 rules, one
for IPv4 and one for IPv6. This changes the response to include multiple
rules if a private subnet name or id is given. Since it wasn't possible
to use a private subnet name or id before, this is in general backwards
compatible. However, code that passed values through without checking
them (such as the CLI) needs adjustment. I think the amount of backwards
compatibility breakage here is acceptable.

In the web UI, there isn't enough horizontal room to have this in the
same td as the cidr text input. Use a separate tr, and make the border
white so it appears to be part of the same row:
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-north-h1_firewall_test-fw4_networking

----

> [!IMPORTANT]
> Add support for adding firewall rules using private subnet names or IDs in CLI, API, and web UI, with corresponding updates to documentation and tests.
> 
>   - **Behavior**:
>     - Support adding firewall rules using private subnet name or ID, creating two rules for IPv4 and IPv6.
>     - CLI command `add-rule` updated to accept `ps-id` or `ps-name` in `add-rule.rb`.
>     - API response in `openapi.yml` updated to return multiple rules if private subnet is used.
>   - **Web UI**:
>     - Updated `networking.erb` to allow selection of private subnet for firewall rules.
>     - Adjusted layout to accommodate new input fields.
>   - **Helpers**:
>     - `authorized_object` in `general.rb` and `authorized_private_subnet` in `private_subnet.rb` updated to handle name and ID.
>   - **Routes**:
>     - `firewall.rb` updated to process private subnet input and create corresponding rules.
>   - **Tests**:
>     - Added tests in `add-rule_spec.rb` and `firewall_spec.rb` for new functionality.
>     - Updated CLI golden files to reflect new command usage.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 117f026492821493c05560ac0667c2f638783fdc. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/routes/project/location/firewall.rb
+++ b/routes/project/location/firewall.rb
@@ -90,7 +90,8 @@ class Clover
             @page = "networking"
           end
 
-          cidr = typecast_params.str!("cidr")
+          cidr = typecast_params.nonempty_str("fw_rule_private_subnet_id") if web?
+          cidr ||= typecast_params.str!("cidr")
           unless cidr.include?(".") || cidr.include?(":")
             if PrivateSubnet.ubid_format.match?(cidr)
               key = :id
@@ -124,7 +125,11 @@ class Clover
             cidrs = cidrs[0] if cidrs.length == 1
             Serializers::FirewallRule.serialize(cidrs)
           else
-            flash["notice"] = "Firewall rule is created"
+            flash["notice"] = if cidrs.length == 1
+              "Firewall rule is created"
+            else
+              "Firewall rules are created"
+            end
             r.redirect firewall, "/networking"
           end
         end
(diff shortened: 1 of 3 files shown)


furkansahin (approved):
thank you!

PR #4113, opened 2025-10-31
Load balancer health checks and routing logic is split for ipv4/6

furkansahin:
## Add multiple load_balancer_vm_port for each stack option
Since the load balancer health checks are performed through
load_balancer_vm_port entity, and I want to separate the health state of
each stack for the same VM, I am adding a new row per stack.
This way, if the ipv4 or ipv6 (single) endpoint is unhealthy for a dual
stack load balancer, we will be able to individually identify that.

## Use the stack information and fix the loadbalancing by unique port
We start using the stack information while configuring load balancing.
This way, if the ipv4 stack is up on a node while the ipv6 is down, it
keeps getting the traffic for ipv4 but not for ipv6.

More importantly, we are fixing a sneaky bug here. Previously, using
uniq on load_balancer.active_vm_ports collapsed multiple
load_balancer_vm_ports that shared the same load_balancer_port. This
caused traffic to be routed to only one node instead of multiple.
In some cases, it could even select the active VM’s own port, resulting
in an empty neighbor set and a total routing failure. The uniq property
was added to avoid an undeterministic bug which we solve here by sorting
on the stack column.

## Report ipv4/6 health status separately in the UI

----

> [!IMPORTANT]
> Split load balancer health checks and routing logic for IPv4/IPv6, adding stack handling and fixing port uniqueness bug.
> 
>   - **Database**:
>     - Add `stack` column to `load_balancer_vm_port` in `20251031_add_stack_to_lb_vm_port.rb`, allowing separate entries for IPv4 and IPv6.
>     - Create enum `lb_vm_port_stack` with values `ipv4` and `ipv6`.
>     - Update unique index to include `stack`.
>   - **Models**:
>     - Update `LoadBalancer` in `load_balancer.rb` to handle `stack` when adding/removing ports and VMs.
>     - Modify `LoadBalancerVmPort` in `load_balancer_vm_port.rb` to check `stack` in `check_probe()`.
>     - Adjust `Vm` in `vm.rb` to return separate states for IPv4 and IPv6 in `load_balancer_state()`.
>   - **Logic**:
>     - In `update_load_balancer_node.rb`, sort VM ports by `stack` and adjust NAT rules generation to handle separate stacks.
>     - Fix bug in load balancing logic by sorting on `stack` instead of using `uniq` on ports.
>   - **UI**:
>     - Update `vms.erb` to display separate application states for IPv4 and IPv6 when `stack` is `dual`.
>   - **Tests**:
>     - Update tests in `load_balancer_spec.rb`, `load_balancer_vm_port_spec.rb`, and other specs to reflect changes in stack handling and ensure correct behavior.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a9fdfd06402cff8457c01700d26404d3d18c209e. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -51,7 +51,12 @@ class Vm < Sequel::Model
   end
 
   def load_balancer_state
-    load_balancer_vm_ports.first&.state
+    return nil unless load_balancer
+    if load_balancer.stack == "dual"
+      [load_balancer_vm_ports.find { |vm_port| vm_port.stack == "ipv4" }.state, load_balancer_vm_ports.find { |vm_port| vm_port.stack == "ipv6" }.state]
+    else
+      [load_balancer_vm_ports.first.state]
+    end
   end
 
   def path
--- a/spec/routes/api/cli/lb/detach-vm_spec.rb
+++ b/spec/routes/api/cli/lb/detach-vm_spec.rb
@@ -14,14 +14,14 @@ RSpec.describe Clover, "cli lb attach-vm" do
   end
 
   it "detaches VM from load balancer by id" do
-    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["down"]
+    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["down", "down"]
     expect(cli(%W[lb eu-central-h1/test-lb detach-vm #{@vm.ubid}])).to eq "Detached VM #{@vm.ubid} from load balancer with id #{@lb.ubid}\n"
-    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["detaching"]
+    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["detaching", "detaching"]
   end
 
   it "detaches VM from load balancer by name" do
-    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["down"]
+    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["down", "down"]
     expect(cli(%W[lb eu-central-h1/test-lb detach-vm test-vm])).to eq "Detached VM test-vm from load balancer with id #{@lb.ubid}\n"
-    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["detaching"]
+    expect(@lb.vm_ports_dataset.select_map(:state)).to eq ["detaching", "detaching"]
   end
 end
--- a/spec/routes/web/load_balancer_spec.rb
+++ b/spec/routes/web/load_balancer_spec.rb
@@ -281,7 +281,7 @@ RSpec.describe Clover, "load balancer" do
           "SSL Certificate Status", "Available"
         ]
 
-        lb.update(health_check_protocol: "tcp")
+        lb.update(health_check_protocol: "tcp", stack: "ipv4")
         page.refresh
         expect(page.all("dt,dd").map(&:text)).to eq [
           "ID", lb.ubid,
@@ -289,7 +289,7 @@ RSpec.describe Clover, "load balancer" do
           "Connection String", "dummy-lb-3.#{ps.ubid[-5...]}.lb.ubicloud.com",
           "Private Subnet", "dummy-ps-1",
           "Algorithm", "Hash Based",
-          "Stack", "dual",
+          "Stack", "IPv4",
           "Load Balancer Port", "80",
           "Application Port", "8000",
           "Health Check Protocol", "TCP",
--- a/views/networking/load_balancer/vms.erb
+++ b/views/networking/load_balancer/vms.erb
@@ -17,7 +17,13 @@ edit_perm = has_permission?("LoadBalancer:edit", @lb) %>
       <thead class="bg-gray-50">
         <tr>
           <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">VM</th>
-          <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Application State</th>
+          <% if @lb.stack == "dual" %>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Application State (IPv4)</th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Application State (IPv6)</th>
+          <% else %>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Application State</th>
+          <% end %>
+
           <% if edit_perm %>
             <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"></th>
           <% end %>
@@ -30,8 +36,13 @@ edit_perm = has_permission?("LoadBalancer:edit", @lb) %>
               <a href="<%= path(vm) %>" class="text-orange-600 hover:text-orange-700"><%= vm.name %></a>
             </td>
             <td class="px-3 py-3.5 whitespace-nowrap text-sm text-gray-900">
-              <%= vm.load_balancer_state %>
+              <%= vm.load_balancer_state.first %>
             </td>
+            <% if @lb.stack == "dual" %>
+              <td class="px-3 py-3.5 whitespace-nowrap text-sm text-gray-900">
+                <%= vm.load_balancer_state.last %>
+              </td>
+            <% end %>
             <td class="relative whitespace-nowrap py-4 pl-3 pr-4 text-right text-sm font-medium sm:pr-6">
               <% form(action: "#{load_balancer_path}/detach-vm", method: :post) do %>
                 <input type="hidden" name="vm_id" value="<%= vm.ubid %>">
@@ -54,8 +65,8 @@ edit_perm = has_permission?("LoadBalancer:edit", @lb) %>
                 }
               ) %>
             </td>
-            <td class="relative whitespace-nowrap py-4 pl-3 pr-4 text-right text-sm font-medium sm:pr-6">
-            </td>
+            <td class="relative whitespace-nowrap py-4 pl-3 pr-4 text-right text-sm font-medium sm:pr-6"></td>
+            <% if @lb.stack == "dual" %><td class="relative whitespace-nowrap py-4 pl-3 pr-4 text-right text-sm font-medium sm:pr-6"></td><% end %>
             <td class="relative whitespace-nowrap py-4 pl-3 pr-4 text-right text-sm font-medium sm:pr-6">
               <% form(action: "#{load_balancer_path}/attach-vm", method: :post, id: "form-lb-#{@lb.ubid}") do %>
                 <%== part("components/form/submit_button", text: "Attach") %>


furkansahin:
@jeremyevans , I need help here. It looks like we recently enabled an extension to detect N+1 queries. Here, we start to have multiple load_balancer_vm_port elements per load_balancer_vm entity. That's why when we remove a VM using `load_balancer.rb:remove_vm()`, it results in multiple `archive` queries. I think that is expected considering there are 2 distinct entities:
```
[#<LoadBalancerVmPort["1qz191dzr9se4jf5j7950f50cw"] @values={load_balancer_vm_id: "1sg6gv74g4je7hkbeb31rbs36b", load_balancer_port_id: "1p1dsrtghf520pgzpxk4vh4373", state: "evacuating", last_checked_at: "2025-11-03 15:16:13", stack: "ipv4"}>, 
#<LoadBalancerVmPort["1qqr8hvyke9j7gakf9wnemgb68"] @values={load_balancer_vm_id: "1sg6gv74g4je7hkbeb31rbs36b", load_balancer_port_id: "1p1dsrtghf520pgzpxk4vh4373", state: "evacuating", last_checked_at: "2025-11-03 15:16:13", stack: "ipv6"}>]
```

but the test is failing with:
```
     Sequel::Postgres::AutoParameterizeDuplicateQueryDetection::DuplicateQueries:
       duplicate queries detected:

       times:2
       sql:INSERT INTO "archived_record" ("model_name", "model_values") VALUES ($1, $2::jsonb) RETURNING *
       backtrace (filtered):
       /Users/furkansahin/clover/lib/resource_methods.rb:107:in 'ResourceMethods::InstanceMethods#before_destroy'
       /Users/furkansahin/clover/model/load_balancer.rb:147:in 'block in LoadBalancer#remove_vm'
       /Users/furkansahin/clover/model/load_balancer.rb:146:in 'LoadBalancer#remove_vm'
       /Users/furkansahin/clover/routes/project/location/load_balancer.rb:113:in 'block (6 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/project/location/load_balancer.rb:109:in 'Array#each'
       /Users/furkansahin/clover/routes/project/location/load_balancer.rb:109:in 'block (5 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/project/location/load_balancer.rb:87:in 'block (4 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/project/location/load_balancer.rb:79:in 'block (3 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/project/location/load_balancer.rb:60:in 'block (2 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/project/location/load_balancer.rb:9:in 'block in <class:Clover>'
       /Users/furkansahin/clover/routes/project/location.rb:8:in 'block (2 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/project/location.rb:5:in 'block in <class:Clover>'
       /Users/furkansahin/clover/routes/project.rb:98:in 'block (2 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/project.rb:45:in 'block in <class:Clover>'
       /Users/furkansahin/clover/clover.rb:1019:in 'block in <class:Clover>'
       /Users/furkansahin/clover/sdk/ruby/lib/ubicloud/adapter/rack.rb:33:in 'Ubicloud::Adapter::Rack#call'
       /Users/furkansahin/clover/sdk/ruby/lib/ubicloud/adapter.rb:43:in 'Ubicloud::Adapter#patch'
       /Users/furkansahin/clover/sdk/ruby/lib/ubicloud/model/load_balancer.rb:31:in 'Ubicloud::LoadBalancer#update'
       /Users/furkansahin/clover/cli-commands/lb/post/update.rb:18:in 'block (2 levels) in <top (required)>'
       /Users/furkansahin/clover/lib/ubi_cli.rb:128:in 'block (2 levels) in UbiCli.base'
       /Users/furkansahin/clover/lib/ubi_cli.rb:65:in 'UbiCli.process'
       /Users/furkansahin/clover/routes/cli.rb:17:in 'block (2 levels) in <class:Clover>'
       /Users/furkansahin/clover/routes/cli.rb:5:in 'block in <class:Clover>'
       /Users/furkansahin/clover/clover.rb:1019:in 'block in <class:Clover>'
     # ./spec/routes/spec_helper.rb:63:in 'block (2 levels) in <top (required)>'
     # ./spec/routes/api/cli/spec_helper.rb:8:in 'cli'
     # ./spec/routes/api/cli/lb/update_spec.rb:55:in 'block (2 levels) in <top (required)>'
     ```
    Do you have a suggestion for the fix?

**Comment by jeremyevans** [comment]
> Do you have a suggestion for the fix?

You need to wrap the related code in `DB.ignore_duplicate_queries`. For the above error, that could be done in `LoadBalancer#remove_vm`. `LoadBalancer#remove_vm_port` has a similar issue.  However, any place that uses `Dataset#destroy` is likely to cause problems, as that will result in duplicate queries.

Probably a better approach is to add `ResourceMethods::DatasetMethods#destroy` and have it do `DB.ignore_duplicate_queries{super}`. That should avoid the warnings. Doesn't fix the duplicate queries issue, though. If we want to actually fix the duplicate queries, we probably need to switch to a trigger-based approach to adding records to `archived_record`, use `ON UPDATE` on foreign keys (or triggers) instead of using the `association_dependencies` plugin in the model, and change the `destroy` calls to `delete`.

**Comment by furkansahin** [comment]
@jeremyevans , thank you for the help!

File: migrate/20251031_add_stack_to_lb_vm_port.rb [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+    #     load_balancer.ports.each do |port|
+    #       LoadBalancerVmPort.create(load_balancer_vm_id: load_balancer_vm.id, load_balancer_port_id: port.id, stack: "ipv6")
+    #     end
+    #   end
+    # end
+
+    # Update the stack column for load balancer vm ports to ipv6 for ipv6 stack
+    # load balancers because the default stack is added as ipv4 above
+    run <<~SQL
+      UPDATE load_balancer_vm_port SET stack = 'ipv6' WHERE load_balancer_port_id IN (SELECT lb_p.id from load_balancer as lb join load_balancer_port as lb_p on lb.id = lb_p.load_balancer_id where lb.stack = 'ipv6') AND stack = 'ipv4';
```
aamederen, line 25: We should add an additional load_balancer_vm_port entry for dual stack LBs, no?

routes/project/location/load_balancer.rb
```diff
@@ -86,8 +86,10 @@ class Clover
 
           DB.transaction do
             lb.update(algorithm:, health_check_endpoint:)
-            lb.ports.first.update(src_port: Validation.validate_port(:src_port, src_port),
-              dst_port: Validation.validate_port(:dst_port, dst_port))
+            lb.ports.each {
```
aamederen, line 89: nit: might this be done without an `.each` using dataset?

prog/vnet/update_load_balancer_node.rb
```diff
@@ -12,7 +12,7 @@ def before_run
   end
 
   label def update_load_balancer
-    vm.load_balancer_vm_ports.select { |lvp| lvp.state == "detaching" }.each do |load_balancer_vm_port|
+    vm.load_balancer_vm_ports.select { |lvp| lvp.state == "detaching" }.sort_by(&:stack).each do |load_balancer_vm_port|
```
aamederen, line 15: why do we sort here?
(also, feels like an opportunity for converting to a dataset method)

model/vm.rb
```diff
@@ -51,7 +51,12 @@ def display_gpu
   end
 
   def load_balancer_state
-    load_balancer_vm_ports.first&.state
+    return nil unless load_balancer
+    if load_balancer.stack == "dual"
+      [load_balancer_vm_ports.find { |vm_port| vm_port.stack == "ipv4" }.state, load_balancer_vm_ports.find { |vm_port| vm_port.stack == "ipv6" }.state]
```
aamederen, line 56: maybe `sort_by(&:stack).map(&:state)` ?

migrate/20251031_add_stack_to_lb_vm_port.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_enum(:lb_vm_port_stack, %w[ipv4 ipv6])
```
jeremyevans, line 5: We are not adding any new usage of enums (it is not possible to reverse the addition of a member to an enum). You can switch this to a boolean, a foreign key to another table with valid values, or use a text column with a CHECK constraint.

migrate/20251031_add_stack_to_lb_vm_port.rb
```diff
@@ -0,0 +1,39 @@
...
+    alter_table(:load_balancer_vm_port) do
+      add_column :stack, :lb_vm_port_stack, null: false, default: "ipv4"
+      add_index [:load_balancer_port_id, :load_balancer_vm_id, :stack], unique: true, name: :lb_vm_port_stack_unique_index # rubocop:disable Sequel/ConcurrentIndex
+      drop_index [:load_balancer_port_id, :load_balancer_vm_id], name: :lb_vm_port_unique_index # rubocop:disable Sequel/ConcurrentIndex
```
jeremyevans, line 9: Are you sure there will be no issues with creating this index? It takes a lock on the table. In general, it would be better to add the column with a NULL value, but not use it. Create the UNIQUE index concurrently before you start to use it, and drop the old UNIQUE index separately after the new UNIQUE index creation has finished.

migrate/20251031_add_stack_to_lb_vm_port.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_enum(:lb_vm_port_stack, %w[ipv4 ipv6])
+    alter_table(:load_balancer_vm_port) do
+      add_column :stack, :lb_vm_port_stack, null: false, default: "ipv4"
```
jeremyevans, line 7: I think this is trying to do too much at once. Add the column allowing NULL in one migration. Modify the code to start setting the value for the column, treating NULL on the column as it is currently treated, with different behavior if the `stack` column is set. After that, backfill the previously existing fields that are still NULL. Once that is complete, then add the NOT NULL constraint in a separate migration. I don't think you should have a DEFAULT on the column.

migrate/20251031_add_stack_to_lb_vm_port.rb
```diff
@@ -0,0 +1,39 @@
...
+      add_index [:load_balancer_port_id, :load_balancer_vm_id, :stack], unique: true, name: :lb_vm_port_stack_unique_index # rubocop:disable Sequel/ConcurrentIndex
+      drop_index [:load_balancer_port_id, :load_balancer_vm_id], name: :lb_vm_port_unique_index # rubocop:disable Sequel/ConcurrentIndex
+    end
+
+    # need to populate extra load_balancer_vm_port records for dual stack load
+    # balancers
+    # LoadBalancer.select{ it.stack == "dual" }.each do |load_balancer|
```
jeremyevans, line 14: This is just example code, but it doesn't do what you want. `select` here calls `Dataset#select`, not `Enumerable#select`.  Additionally, you should use `all` instead of `each`, as issuing queries inside a `Dataset#each` block is not supported in Sequel (though it does work in most cases with the postgres adapter).

I think you can replace all of this example code with a single `Dataset#insert` call, but you first need to add a default value for the `load_balancer_vm_port.id` column that uses the `gen_random_ubid_uuid` function. Something like:

```
LoadBalancerVmPort.insert([:load_balancer_vm_id, :load_balancer_port_id, :state, :stack], 
  LoadBalancerVmPort.select(:load_balancer_vm_id, :load_balancer_port_id, "down", "ipv6")
  .where(load_balancer_port_id: LoadBalancerPort.select(:id)
    .where(load_balancer_id: LoadBalancer.select(:id)
       .where(stack: "dual"))))
```

migrate/20251031_add_stack_to_lb_vm_port.rb
```diff
@@ -0,0 +1,39 @@
...
+    # end
+
+    # Update the stack column for load balancer vm ports to ipv6 for ipv6 stack
+    # load balancers because the default stack is added as ipv4 above
+    run <<~SQL
+      UPDATE load_balancer_vm_port SET stack = 'ipv6' WHERE load_balancer_port_id IN (SELECT lb_p.id from load_balancer as lb join load_balancer_port as lb_p on lb.id = lb_p.load_balancer_id where lb.stack = 'ipv6') AND stack = 'ipv4';
+    SQL
+  end
+
+  down do
```
jeremyevans, line 29: You missed the `drop_enum` in the down block, so if you migrated up and down, the second attempt to migrate up would fail. However, since you should avoid the use of enums, this won't be an issue. However, ensure that the down block reverses all schema changes made in the up block.

routes/project/location/load_balancer.rb
jeremyevans: Agreed, you should use `lb.ports_dataset.update` here to update both ports in the same query.

prog/vnet/update_load_balancer_node.rb
jeremyevans, line 15: `vm.load_balancer_vm_ports_dataset.where(state: "detaching").order(:stack).all do |load_balancer_vm_port|`

This could theoretically remove multiple ports at once for a dual stack load balancer, so I'm guessing this loop could result in duplicate queries.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -30,16 +30,16 @@ def before_run
     pop "load balancer is removed"
   end
 
+  def load_balancer_ports_to_work_on
+    @load_balancer_ports_to_work_on ||= load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port, load_balancer_vm: {vm: :nics}) }.sort_by { |vm_port| vm_port.load_balancer_port.src_port }
```
jeremyevans, line 34: Maybe do the ordering in the query: 

```ruby
load_balancer.active_vm_ports do |ds|
  ds.eager_graph(:load_balancer_port)
    .eager(load_balancer_vm: {vm: :nics})
    .order(:src_port)
end
```

prog/vnet/update_load_balancer_node.rb
```diff
@@ -50,7 +50,7 @@ def generate_lb_based_nat_rules
     end
 
     ipv4_prerouting = if load_balancer.ipv4_enabled?
-      active_vm_ports_uniq_by_port.map do |vm_port|
+      load_balancer_ports_to_work_on.select { |vm_port| vm_port.stack == "ipv4" }.uniq(&:load_balancer_port_id).map do |vm_port|
```
jeremyevans, line 53: Seems like the `select { |vm_port| vm_port.stack == "ipv4" }.uniq(&:load_balancer_port_id)` would be good to move to a separate method to DRY things up.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -126,16 +126,17 @@ def generate_lb_based_nat_rules
...
 
-  def generate_lb_ip_set_definition(neighbor_vms)
-    return ["", ""] if neighbor_vms.empty?
-
-    ["elements = {#{neighbor_vms.map(&:private_ipv4).join(", ")}}",
-      "elements = {#{neighbor_vms.map(&:private_ipv6).join(", ")}}"]
+  def generate_lb_ip_set_definition(neighbor_vm_ports)
+    return ["", ""] if neighbor_vm_ports.empty?
+    ipv4_ips = neighbor_vm_ports.select { it.stack == "ipv4" }.map { it.vm.private_ipv4 }.uniq.join(", ")
+    ipv6_ips = neighbor_vm_ports.select { it.stack == "ipv6" }.map { it.vm.private_ipv6 }.uniq.join(", ")
```
jeremyevans, line 132: Can probably DRY up slightly using something like:

```ruby
ipv4_ips, ipv6_ips = neighbor_vm_ports.partition { it.stack == "ipv4" }
ipv4_ips = ipv4_ips.map! { it.vm.private_ipv4 }.uniq.join(", ")
ipv6_ips = ipv6_ips.map! { it.vm.private_ipv6 }.uniq.join(", ")
```

migrate/20251031_add_stack_to_lb_vm_port.rb
furkansahin: The problem is, this stack column cannot be populated properly without dropping the old index, because there will be multiple `[load_balancer_port_id, load_balancer_vm_id]` entries for `dual` stack load balancers. I don't know why there would be an issue creating this index. We have 78 `load_balancer_vm_port` entries in the database. It should complete in a blink of an eye.

jeremyevans: Certainly if nothing else is accessing the `load_balancer_vm_port` table concurrently, it would complete quickly. If something else is accessing/updating the table, I believe it could block until the related transaction completes, and that would cascading into blocking all access to the table until related transaction completes (since IIRC the index building takes a full table lock).

furkansahin: It actually only conflicts with `UPDATE, DELETE, INSERT, and MERGE`. During the migration, expected access to the table should only be by the monitor which `SELECT`s. In case there is a state change just at that moment, of course we might be blocked but I see that chance negligible.

jeremyevans: OK. If you believe the chance of blocking is negligible, it's fine to build the indexes non-concurrently.

PR #4124, opened 2025-11-04
Send a slack message when boot_id differs from the one in database

mohi-kalantari:


config.rb
```diff
@@ -138,6 +138,9 @@ def self.test?
   optional :pagerduty_key, string, clear: true
   optional :pagerduty_log_link, string
 
+  # VmHost Monitroing
+  optional :slack_api_token, string
```
enescakir:
```suggestion
  optional :slack_webhook_host_crash_url, string
```

model/vm_host.rb
```diff
@@ -348,6 +348,28 @@ def check_storage_kernel_logs(ssh_session, devices)
...
+      req["Authorization"] = "Bearer #{Config.slack_api_token}"
+      req["Content-Type"] = "application/json"
+      req.body = {
+        channel: "#pager",
+        text: "VmHost #{ubid} has a different last_boot_id than the one recorded in the database."
+      }.to_json
+
+      Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) do |http|
+        http.request(req)
+      end
```
enescakir:
```suggestion
    Excon.post(Config.slack_webhook_host_crash_url, headers: {"Content-Type" => "application/json"}, body: { text: "VmHost #{ubid} has a different last_boot_id than the one recorded in the database."}.to_json, expects: 200)
```

fdr:
1) commit message: why. I mean, I can guess (and it's only a guess), but if you are going to bother to tell someone here, put it in there.
2) I am not that keen with putting in yet another rando dependency to have the system transmit information, but admit it could be practical until we get a handle on this kernel bug business (i.e. we would remove it) or we develop a theory around how to flesh this out
3) If this is not rough-and-ready type of code (that we'd remove), then you probably want to save a record so the admin page can true you up as to what hosts need investigating because of unexpected reboots. Otherwise you won't be able to easily keep track of what investigations seen in slack are complete, and what are not.

enescakir:
Btw, we don’t need to add a Slack dependency just to send messages to a Slack channel. Mezmo can already send Slack messages based on search queries. So if you log crash messages with Clog, we can easily configure Mezmo to send them to Slack automatically.

PR #4126, opened 2025-11-04
Draft: Replace Tailwind with normal CSS

jeremyevans:
This is not intended for commit. It still has many visual issues (see images below), and interactive parts such as the account dropdown and project switcher do not work. This also has a few spec failures.

All changes made by Claude Code. Recent prompts have reduced the number of Tailwind style classes while increasing the number of visual issues, so I think further general work by Claude Code is unlikely to result in net benefits (more focused prompts could fix some of the visual issues, though).

While the results are better than I expected, this still leaves the templates more element-heavy than I think they need to be, and there is still a fair amount of Tailwind-style classes in the app.css file. I think many of these could be removed, though Claude didn't remove them even after prompted to do so multiple times.

@enescakir and @byucesoy, is this a direction you would like to explore?

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_dashboard
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_token
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_user
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_cli
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_kubernetes-cluster
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_inference-endpoint
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_postgres
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_private-subnet
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_vm

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -580,9 +580,6 @@ body {
   margin-block: auto;
 }
 
-.-m-1 {
-  margin: calc(-1 * var(--spacing-1));
-}
 
 .-m-1\.5 {
   margin: calc(-1 * var(--spacing-1-5));
@@ -632,13 +629,7 @@ body {
   margin-top: auto;
 }
 
-.-mt-2 {
-  margin-top: calc(-1 * var(--spacing-2));
-}
 
-.-mt-6 {
-  margin-top: calc(-1 * var(--spacing-6));
-}
 
 .mb-0\.5 {
   margin-bottom: var(--spacing-0-5);
@@ -696,9 +687,6 @@ body {
   margin-left: auto;
 }
 
-.-ml-1 {
-  margin-left: calc(-1 * var(--spacing-1));
-}
 
 .mr-1 {
   margin-right: var(--spacing-1);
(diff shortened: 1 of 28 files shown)


enescakir:
> @enescakir and @byucesoy, is this a direction you would like to explore?

IMHO, It’s not worth it at the moment. I don’t see any short-term benefits that justify the time investment.
- If we ever decide to change our design or the underlying UI kit, maybe then it would make sense. But for the current design, there’s no need to create extra work.
    - Right now, when we need to design a big feature, we generally pick components from [Tailwind UI](https://tailwindcss.com/plus/ui-blocks)￼. If we were to switch to plain CSS, we’d probably need a new UI kit anyway. Designing all new components from scratch isn’t feasible, something like [Tabler](https://github.com/tabler/tabler)￼ could work, but it’s Bootstrap-based.
- If we hire a dedicated frontend/UI/UX engineer, it could be worthwhile. They could design new components and maintain the existing ones.
- Even if something like Claude Code seems successful for refactoring, I’m certain we’d encounter a lot of visual issues. Some might not matter to anyone else but would bother me and take my time to fix. Even now, some current pages have pretty bad visual issues, but I don't have the time to address them. I’m not sure my eyes could handle a new wave of visual issues on top of that.

jeremyevans:
While I think we should move from Tailwind to CSS long term, I agree that while the results are interesting, this PR is not the right approach. So I'm going to close this now.

PR #4130, opened 2025-11-05
Change ubid type of BillingInfo from `et` to `b1`

aamederen:
While working on something else, I realized that the ubid type of BillingInfo was set to ETC_TYPE in #3360. So all BillingInfo's after the change have their ubids start with `et` not `b1` and causing decoding to break.

This change also adds a test that check all types (with exceptions) have the correct ubid_type defined.

----

> [!IMPORTANT]
> Fix `ubid` type of `BillingInfo` from `et` to `b1` and add test for correct `ubid_type` in `ubid_spec.rb`.
> 
>   - **Behavior**:
>     - Change `ubid` type of `BillingInfo` from `et` to `b1` by removing `etc_type: true` from `plugin ResourceMethods` in `billing_info.rb`.
>   - **Testing**:
>     - Add test in `ubid_spec.rb` to verify all types (except `TYPE_ETC` and `TYPE_OBJECT_METATAG`) have correct `ubid_type` defined.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c4444ea1b461881b6d8eef0184f22ca68ecb3da0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/billing_info.rb
+++ b/model/billing_info.rb
@@ -9,7 +9,7 @@ class BillingInfo < Sequel::Model
   one_to_many :payment_methods, order: Sequel.desc(:created_at)
   one_to_one :project
 
-  plugin ResourceMethods, etc_type: true
+  plugin ResourceMethods
 
   def stripe_data
     if (Stripe.api_key = Config.stripe_secret_key)
--- a/spec/ubid_spec.rb
+++ b/spec/ubid_spec.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 RSpec.describe UBID do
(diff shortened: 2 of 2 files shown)


jeremyevans (approved):
Thanks for catching this. It would have been better if we used more bits in the uuid for type id and didn't have a need for ETC_TYPE, but what's done is done.

enescakir:
Should we somehow update existing billing info IDs? Since they fail to decode?

aamederen:
@enescakir not necessarily urgent IMO, since we don't decode them.

PR #4131, opened 2025-11-05
Fix Ubi PG on AWS VPC creation with tcp 22 allow rule

furkansahin:
With a recent change, we moved the tcp 22 rule into an internal firewall which are attached to VMs once they are up and running. However, AWS security groups are created through VPCs which are part of private_subnets. Therefore, not having the tcp 22 rule on the private_subnet firewall breaks the initial VM provisioning in AWS. Because, when we reach wait_sshable, the security group lacks the tcp 22 allow rule since it is part of the vm specific rules.  Those are applied only after we switch to wait state.  This is also necessary because a customer may want to have subnets that are completely locked down and have no allow rules in their firewall. In that case, we cannot verify sshability at the provisioning time if we process the firewalls before.

----

> [!IMPORTANT]
> Adds TCP port 22 ingress rule to AWS VPC creation for SSH access during initial provisioning.
> 
>   - **Behavior**:
>     - Adds TCP port 22 ingress rule from `0.0.0.0/0` in `wait_vpc_created` in `vpc.rb` to ensure SSH access during initial VM provisioning.
>     - Handles `InvalidPermissionDuplicate` error in `allow_ingress()` to avoid duplicate rule errors.
>   - **Refactoring**:
>     - Extracts security group ingress logic into `allow_ingress()` in `vpc.rb`.
>   - **Tests**:
>     - Updates `wait_vpc_created` test in `vpc_spec.rb` to expect TCP port 22 ingress rule.
>     - Adjusts tests to handle duplicate permission errors in `vpc_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b695c9eaf84b0214eb283a9f85656ff4825f26af. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/aws/vpc.rb
+++ b/prog/aws/vpc.rb
@@ -49,20 +49,12 @@ class Prog::Aws::Vpc < Prog::Base
 
       private_subnet.firewalls.flat_map(&:firewall_rules).each do |firewall_rule|
         next if firewall_rule.ip6?
-
-        begin
-          client.authorize_security_group_ingress({
-            group_id: security_group_response.group_id,
-            ip_permissions: [{
-              ip_protocol: "tcp",
-              from_port: firewall_rule.port_range.first,
-              to_port: firewall_rule.port_range.last - 1,
-              ip_ranges: [{cidr_ip: firewall_rule.cidr.to_s}]
-            }]
-          })
-        rescue Aws::EC2::Errors::InvalidPermissionDuplicate
-        end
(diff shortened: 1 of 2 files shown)


jeremyevans:
Do we not need the IPv6 port 22 rule or the 5432 or 6432 rules for the private subnet? Seems like it would be more consistent to allow all of the VM firewall rules for AWS VPC, just as would be done for a normal VM.

PR #4132, opened 2025-11-05
Use SPDK CPUs for VMs on hosts without SPDK and not accepting slices

mohi-kalantari:
**Use SPDK CPUs for VMs on hosts without SPDK and not accepting slices**

On VmHosts not accepting slices and without running SPDK units, SPDK CPUs will be used for vhost block backend units to improve performance of vhost block backend VMs.

**Update specs after introducing cpus to vm.storage_volumes**

After adding the new field, we inspect spdk installations on the vm_host to determine whether to use spdk cpus for ubiblk vms. Some tests did not expect the spdk_installations field in VmHost. This commit fixes it.

----

> [!IMPORTANT]
> Use SPDK CPUs for VMs on hosts running SPDK and not accepting slices, updating `storage_volumes` and related tests.
> 
>   - **Behavior**:
>     - Use SPDK CPUs for VMs on hosts running SPDK and not accepting slices in `storage_volumes` in `vm.rb`.
>     - Update `vhost_backend_config` in `storage_volume.rb` to include `cpus` field.
>   - **Tests**:
>     - Update `vm_spec.rb` to test `storage_volumes` with SPDK CPU allocation.
>     - Update `nexus_spec.rb` to ensure correct behavior with SPDK CPU allocation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for cbb9090553d914f3737897440f0e6f9652f315f5. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/model/vm_spec.rb
+++ b/spec/model/vm_spec.rb
@@ -323,7 +323,9 @@ RSpec.describe Vm do
       total_cores: 12,
       total_dies: 1,
       total_sockets: 1,
-      ndp_needed: false
+      ndp_needed: false,
+      spdk_installations: [],
+      accepts_slices: true
     )).at_least(:once)
     expect(JSON.parse(vm.params_json)["init_script"]).to eq ""
     VmInitScript.create_with_id(vm.id, script: "b")
@@ -465,7 +467,7 @@ RSpec.describe Vm do
             random_private_ipv6: NetAddr::IPv6Net.parse("fd00::/64")
           )
         ),
-        vm_host: instance_double(VmHost, ndp_needed: false)
+        vm_host: instance_double(VmHost, ndp_needed: false, spdk_installations: [], accepts_slices: true)
       )
     end
 
--- a/spec/prog/vm/nexus_spec.rb
+++ b/spec/prog/vm/nexus_spec.rb
@@ -324,7 +324,7 @@ RSpec.describe Prog::Vm::Nexus do
         sshable = instance_double(Sshable)
         expect(sshable).to receive(:cmd).with("common/bin/daemonizer --check prep_#{nx.vm_name}").and_return("NotStarted")
         vmh = instance_double(VmHost, sshable: sshable,
-          total_cpus: 80, total_cores: 80, total_sockets: 10, ndp_needed: false, arch: "arm64")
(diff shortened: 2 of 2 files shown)


rhizome/host/lib/storage_volume.rb
```diff
@@ -39,6 +39,7 @@ def initialize(vm_name, params)
     @queue_size = params.fetch("queue_size", 256)
     @copy_on_read = params.fetch("copy_on_read", false)
     @stripe_sector_count_shift = Integer(params.fetch("stripe_sector_count_shift", 11))
+    @cpus = params.fetch("cpus", nil)
```
jeremyevans: Any reason to use `fetch` here instead of `params["cpus"]`? The only behavioral difference would be if you are using a hash with a default value or proc, which I'm guessing we aren't.

model/vm.rb
```diff
@@ -267,6 +267,9 @@ def params_json(swap_size_bytes: nil, hypervisor: nil, ch_version: nil, firmware
   end
 
   def storage_volumes
+    add_cpus = vm_host.spdk_installations.any? && !vm_host.accepts_slices
+    cpus = (0...vm_host.spdk_installations.first.cpu_count).to_a if add_cpus
+
```
pykello: I was thinking we change the cpus when:

- The host **doesn't** have an SPDK installation (in which case, no one is using cpus  supposed to be used by SPDK)
- ... and `!vm_host.accepts_slice`

Then use some of the `vm_host.cpus.filter(&:spdk).map(&:cpu_number)` for ubiblk.

The goal is to limit the number of cpus IO threads are using to get better performance than not limiting at all. If there's already an SPDK installation, it will do busy loop on these cpus and we might get worse performance.

Out of these SPDK cpus, we can choose `max(spdk_cpus.count, s.num_queues)`  cpus (and adjust `num_queues` to be equal to number of cpus chosen). Initially I thought we can do a round-robin to distribute the load between SPDK cpus, but I think doing just a random subset should be good enough.

Note that we cannot set `num_queues = spdk_cpus.count` since we must have `storage.num_queues <= vm.vcpus`.

Then in prod, we just remove the SPDK installation & let ubiblk use those cpus.

pykello (approved):

PR #4133, opened 2025-11-05
Send invoice to customer when waiting for bank transfer

enescakir:
- **Extract reused values into variables in the invoice generator**
  These variables will be used in upcoming commits.
  

- **Store bank transfer info if no payment method**
  If a project has billing information but no payment method, it means we
  collect payments via bank transfer. In this case, we need to store the
  bank transfer details since we use different accounts for EU and non-EU
  customers, and they might change over time. This commit stores the
  information in the invoice; the next commit will display it in the
  invoice PDF.
  

- **Show bank transfer info on the invoice PDF**
  If the invoice includes bank transfer details, display them on the
  invoice PDF.
  

- **Set due date to 30 days for invoices with bank transfer info**
  Normally, we charge customers immediately via their registered card when
  an invoice is created, so both the invoice date and due date are set to
  the creation date.
  
  Since we’ve started accepting bank transfers from some enterprise
  customers, we now expect them to complete payments via bank transfer
  within 30 days. This commit sets a custom 30-day due date for such
  invoices.
  

- **Use less mocking in Invoice model specs**
  

- **Use invoice status instead of passed boolean in send_success_email**
  Boolean allows to have only two states, and requires pass more arguments
  to the method.
  
  To make it more extensible, we can use invoice status instead.
  

- **Send invoice to customer when waiting for bank transfer**
  Normally, customers can’t create billing information without a payment
  method. For customers who pay via bank transfer, we manually create
  billing info without a payment method from the backend.
  
  Currently, we send invoices to these customers manually. With this
  change, if an invoice includes bank_transfer_info (because it doesn’t
  have a payment method), we’ll automatically send the invoice and mark it
  as waiting_transfer.
  

----

> [!IMPORTANT]
> Add support for handling bank transfer payments in invoices, including storing bank transfer info, setting due dates, and sending invoices automatically.
> 
>   - **Behavior**:
>     - Store bank transfer info in `invoice_generator.rb` if no payment method, using different accounts for EU and non-EU customers.
>     - Display bank transfer info on invoice PDF in `invoice.rb` if present.
>     - Set due date to 30 days for invoices with bank transfer info in `invoice_generator.rb`.
>     - Automatically send invoice and mark as `waiting_transfer` if bank transfer info is present in `invoice.rb`.
>   - **Refactoring**:
>     - Extract reused values into variables in `invoice_generator.rb`.
>     - Use invoice status instead of boolean in `send_success_email` in `invoice.rb`.
>   - **Testing**:
>     - Add tests for bank transfer info in `invoice_generator_spec.rb` and `invoice_spec.rb`.
>     - Update tests for invoice PDF rendering in `billing_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 25b13905d6f2873c62033635f50c5754c2688156. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/model/invoice_spec.rb
+++ b/spec/model/invoice_spec.rb
@@ -62,6 +62,14 @@ RSpec.describe Invoice do
   end
 
   describe ".send_success_email" do
+    it "sends email for waiting transfers" do
+      invoice.update(status: "waiting_transfer")
+      expect(invoice).to receive(:persist)
+      invoice.send_success_email
+      expect(Mail::TestMailer.deliveries.length).to eq 1
+      expect(Mail::TestMailer.deliveries.first.html_part.body).to include("Please follow the bank transfer instructions at the bottom of the invoice to complete the payment.")
+    end
+
     it "does not send the invoice if it has no billing information" do
       update_content(billing_info: nil, resources: [], subtotal: 0.0, credit: 0.0, discount: 0.0, cost: 0.0)
       expect(Clog).to receive(:emit).with("Couldn't send the invoice because it has no billing information").and_call_original
@@ -104,10 +112,12 @@ RSpec.describe Invoice do
       expect(invoice.charge).to be false
     end
 
-    it "not charge if no payment methods" do
-      update_content(billing_info: {"id" => billing_info.id})
-      expect(Clog).to receive(:emit).with("Invoice doesn't have billing info.").and_call_original
-      expect(invoice.charge).to be false
+    it "not charge and wait for transfer if no payment methods" do
+      update_content(bank_transfer_info: {"Beneficiary" => "Ubicloud B.V."}, billing_info: {"id" => billing_info.id, "email" => "customer@example.com"})
+      expect(Clog).to receive(:emit).with("Invoice is waiting for transfer.").and_call_original
+      expect(invoice).to receive(:persist)
+      expect(invoice.charge).to be true
+      expect(invoice.status).to eq("waiting_transfer")
     end
 
     it "not charge if all payment methods fails" do
--- a/model/invoice.rb
+++ b/model/invoice.rb
@@ -87,9 +87,16 @@ class Invoice < Sequel::Model
       return true
     end
 
-    if (billing_info = BillingInfo[content.dig("billing_info", "id")]).nil? || billing_info.payment_methods.empty?
+    unless (billing_info = BillingInfo[content.dig("billing_info", "id")])
       Clog.emit("Invoice doesn't have billing info.") { {invoice_no_billing: {ubid: ubid}} }
       return false
+    end
+
+    if content["bank_transfer_info"]
+      update(status: "waiting_transfer")
+      Clog.emit("Invoice is waiting for transfer.") { {invoice_waiting_transfer: {ubid: ubid, cost: amount}} }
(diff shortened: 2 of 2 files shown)


lib/invoice_generator.rb
```diff
@@ -21,32 +21,47 @@ def run
...
+        country = bi&.country
+        is_eu = country&.in_eu_vat?
+        project_content[:billing_info] = bi&.stripe_data&.merge({
+          "id" => bi.id,
+          "ubid" => bi.ubid,
+          "in_eu_vat" => !!is_eu
         })
 
+        project_content[:bank_transfer_info] = if bi && bi.payment_methods.empty?
+          project_content[:due_date] = (Time.now + 30 * 24 * 60 * 60).to_date.to_s
```
jeremyevans:
```suggestion
          project_content[:due_date] = (Date.today + 30).to_s
```

serializers/invoice.rb
```diff
@@ -18,6 +18,7 @@ def self.serialize_internal(inv, options = {})
       path: inv.path,
       name: inv.name,
       date: inv.created_at.strftime("%B %d, %Y"),
+      due_date: inv.content["due_date"] ? Time.parse(inv.content["due_date"]).strftime("%B %d, %Y") : inv.created_at.strftime("%B %d, %Y"),
```
jeremyevans:
```suggestion
      due_date: inv.content["due_date"] ? Date.parse(inv.content["due_date"]).strftime("%B %d, %Y") : inv.created_at.strftime("%B %d, %Y"),
```

spec/model/invoice_spec.rb
```diff
@@ -3,78 +3,67 @@
...
+  def update_content(new_content)
+    invoice.content.merge!(new_content.transform_keys(&:to_s))
+    invoice.modified!(:content)
+    invoice.save_changes
+  end
+
   describe ".blob_key" do
     it "returns path with below_minimum_threshold group when status is below_minimum_threshold" do
-      invoice.status = "below_minimum_threshold"
+      invoice.update(status: "below_minimum_threshold")
```
jeremyevans: For this and other changes using `update_content`, you don't need to save to the database unless whatever being tested is reading from the database and not just accessing the model's stored data.

jeremyevans (approved):

lib/invoice_generator.rb
enescakir: I didn’t know that integer operations on Date are day based. Good to know! From time to time, I really miss the Duration helpers from ActiveSupport. Things like `1.day` or `30.minutes` make the code so much more readable. Honestly, that might be my favorite part of ActiveSupport.

spec/model/invoice_spec.rb
enescakir: We reload the invoice to get its latest status and avoid double charging. This is necessary for tests that call the `.charge` method.

https://github.com/ubicloud/ubicloud/blob/444b833195933590c68a65049d6bc821109b14ed/model/invoice.rb#L71

PR #4136, opened 2025-11-05
Implement support to remove all cache entries from a project

jeremyevans:
First commit was created by Claude Code. It has some issues, fixed in the second commit.

However, in general the approach will be problematic for repositories with a large number of cache entries, since each destruction needs an S3 query. We need to switch to an approach that uses a prog.

----

> [!IMPORTANT]
> Adds support for removing all cache entries from a GitHub repository via CLI, API, and web interface, with backend logic and tests included.
> 
>   - **Behavior**:
>     - Adds `remove-all-cache-entries` CLI command in `remove-all-cache-entries.rb` to delete all cache entries for a GitHub repository.
>     - Updates `openapi.yml` to include `deleteAllGithubCacheEntries` operation for deleting all cache entries.
>     - Implements `remove_all_cache_entries` method in `github_repository.rb` to call the API endpoint.
>   - **Backend Logic**:
>     - Introduces `Prog::Github::DeleteCacheEntries` in `delete_cache_entries.rb` to handle cache entry deletion.
>     - Adds `delete_all_cache_entries` action to `SUPPORTED_ACTIONS` and `LOGGED_ACTIONS` in `general.rb`.
>     - Updates `github.rb` route to handle cache deletion requests.
>   - **Testing**:
>     - Adds `delete_cache_entries_spec.rb` to test cache entry deletion logic.
>     - Adds `remove-all-cache-entries_spec.rb` to test CLI command functionality.
>     - Updates `github_spec.rb` to test web interface for cache deletion.
>   - **Misc**:
>     - Updates CLI help files to include new command.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 095f857078c7d1f9630d8eb5f237454da8600e23. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -120,7 +120,7 @@ class Clover < Roda
     update_invitation
     upgrade
   ACTIONS
-  LOGGED_ACTIONS = Set.new(%w[create create_replica destroy promote reset_superuser_password restart restore update]).freeze
+  LOGGED_ACTIONS = Set.new(%w[create create_replica delete_all_cache_entries destroy promote reset_superuser_password restart restore update]).freeze
 
   def audit_log(object, action, objects = [])
     raise "unsupported audit_log action: #{action}" unless SUPPORTED_ACTIONS.include?(action)
--- a/prog/github/delete_cache_entries.rb
+++ b/prog/github/delete_cache_entries.rb
@@ -3,34 +3,28 @@
 class Prog::Github::DeleteCacheEntries < Prog::Base
   subject_is :github_repository
 
-  def self.assemble(repository_id, initiated_at = Time.now)
-    DB.transaction do
-      strand = Strand.create_with_id(
-        repository_id,
-        prog: "Github::DeleteCacheEntries",
-        label: "start",
-        stack: [{
-          "subject_id" => repository_id,
-          "initiated_at" => initiated_at.to_s
-        }]
-      )
-      strand
-    end
-  end
-
-  label def start
-    hop_delete_entries
+  def self.assemble(repository_id, initiated_at: Time.now)
+    Strand.create(
+      prog: "Github::DeleteCacheEntries",
+      label: "delete_entries",
+      stack: [{
+        "subject_id" => repository_id,
+        "initiated_at" => initiated_at.to_s
+      }]
+    )
   end
 
   label def delete_entries
-    cache_entry = github_repository.cache_entries_dataset.first
-
-    unless cache_entry
-      pop "all cache entries deleted"
+    if (cache_entry = next_entry)
+      cache_entry.destroy
+      nap 0
     end
 
-    cache_entry.destroy
+    pop "all cache entries deleted"
+  end
 
-    nap 0
+  def next_entry
+    initiated_at = Time.parse(frame["initiated_at"])
+    github_repository.cache_entries_dataset.order(:created_at).first { created_at < initiated_at }
   end
 end
--- a/routes/project/github.rb
+++ b/routes/project/github.rb
@@ -128,15 +128,18 @@ class Clover
             end
 
             r.delete true do
-              DB.transaction do
-                if repository.cache_entries_dataset.empty?
-                  no_audit_log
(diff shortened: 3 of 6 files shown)


enescakir:
We have a similar issue with cache cleanup. If a repository has too many cache entries, it can't be cleaned up in a single strand run and causes apoptosis. Currently, we clean up 200 cache entries per run: https://github.com/ubicloud/ubicloud/blob/f8a0bee4f200d232b7911bedabc3be8f37238b1b/prog/github/github_repository_nexus.rb#L96

One approach is to release the API/CLI first, and then improve it to follow a prog style approach. Though, in the current style, customers can run it multiple times to eventually clean everything. However, because we run it inside a transaction, even if some cache files are destroyed in blob storage before a timeout, the records remain in the control plane. Consecutive runs are faster though.

For the prog-style approach, what do you think would be better: (1) create a strand for each cache entry, or (2) write a prog that destroys them in bulk?

jeremyevans:
> For the prog-style approach, what do you think would be better: (1) create a strand for each cache entry, or (2) write a prog that destroys them in bulk?

I was thinking of using a prog that destroys a the next cache entry in the list and then calls `nap 0`. That will keep destroying cache entries one-by-one, with a transaction for each, in the same `strand.run` call, for STRAND_RUNTIME seconds. Then it would exit. The dispatcher would then pick up for scheduling shortly. That should avoid apoptosis but still be reasonably efficient.

enescakir:
> > For the prog-style approach, what do you think would be better: (1) create a strand for each cache entry, or (2) write a prog that destroys them in bulk?
> 
> I was thinking of using a prog that destroys a the next cache entry in the list and then calls `nap 0`. That will keep destroying cache entries one-by-one, with a transaction for each, in the same `strand.run` call, for STRAND_RUNTIME seconds. Then it would exit. The dispatcher would then pick up for scheduling shortly. That should avoid apoptosis but still be reasonably efficient.

Sounds good :rocket: One small improvement: we previously observed that destroying 1000+ cache entries could take longer than 2 minutes. Also, customers might create new cache entries after triggering a "destroy all" action.

Instead of fetching a fresh cache list from the database for every run, it might be better to persist the list of IDs that existed at the trigger time in the stack. This way, we destroy only those entries. Additionally, having the list allows us to split it into multiple progs for higher concurrency.

Alternatively, instead of storing the list, we could persist the trigger time in the stack and destroy only cache entries with created_at earlier than the trigger time.

jeremyevans:
> Instead of fetching a fresh cache list from the database for every run, it might be better to persist the list of IDs that existed at the trigger time in the stack. This way, we destroy only those entries. Additionally, having the list allows us to split it into multiple progs for higher concurrency.

I was originally planning on storing the list of cache entry ids in the strand's stack (which isn't great in terms of efficiency for large lists), but

> Alternatively, instead of storing the list, we could persist the trigger time in the stack and destroy only cache entries with created_at earlier than the trigger time.

this sounds like a much more efficient approach, so I'll use it.

enescakir:
Yeah, the latter approach (persisting the trigger time) looks more efficient and easier to implement.

For the first approach, if we split the list into multiple strands, say 100 entries each, do you think that would be okay? I'm asking because if we write such a prog with a static list in the stack, we might be able to reuse it in the repository nexus for cleaning up cache more efficiently as well.

jeremyevans:
> Yeah, the latter approach (persisting the trigger time) looks more efficient and easier to implement.
> 
> For the first approach, if we split the list into multiple strands, say 100 entries each, do you think that would be okay? I'm asking because if we write such a prog with a static list in the stack, we might be able to reuse it in the repository nexus for cleaning up cache more efficiently as well.

I think it would be fine to use that approach to implement `cleanup_cache` in the repository nexus.

enescakir (approved):

PR #4138, opened 2025-11-06
Separate IPv4/6 stack health checks in load balancer - Phase 1

furkansahin:
## Add stack column to LoadBalancerVmPort - migration
We are also removing the UNIQUE index on [load_balancer_port_id,
load_balancer_vm_id] because we will have multiple of these rows per
stack for dual stack load balancers. This is why we are adding a new
index on [load_balancer_port_id, load_balancer_vm_id, stack].

## Add one load_balancer_vm_port for each stack option
Since the load balancer health checks are performed through
load_balancer_vm_port entity, and I want to separate the health state of
each stack for the same VM, I am adding a new row per stack.
This way, if the ipv4 or ipv6 (single) endpoint is unhealthy for a dual
stack load balancer, we will be able to individually identify that.

This commit simply populates the rows for new instances

----

> [!IMPORTANT]
> Separate health checks for IPv4 and IPv6 stacks in load balancers with schema, model, and test updates.
> 
>   - **Database Migrations**:
>     - Add `stack` column to `load_balancer_vm_port` with constraint for values 'ipv4' and 'ipv6' in `20251104_add_stack_to_lb_vm_port.rb`.
>     - Add unique index on `load_balancer_port_id`, `load_balancer_vm_id`, and `stack` in `20251105_add_stack_to_lb_vm_port.rb`.
>     - Remove old unique index on `load_balancer_port_id` and `load_balancer_vm_id` in `20251106_add_stack_to_lb_vm_port.rb`.
>   - **Model Changes**:
>     - Update `LoadBalancer` in `load_balancer.rb` to handle dual-stack by adding `add_port_with_stack()`.
>     - Modify `LoadBalancerVmPort` in `load_balancer_vm_port.rb` to include stack-specific health checks.
>   - **Route and Test Updates**:
>     - Update `load_balancer.rb` to handle stack-specific operations.
>     - Adjust tests in `load_balancer_spec.rb` to verify dual-stack behavior.
>     - Update `inference_endpoint_replica_nexus_spec.rb` and `inference_router_replica_nexus_spec.rb` for stack-specific state checks.
>     - Modify `update_load_balancer_node_spec.rb` to test load balancing with stack separation.
>     - Update `detach-vm_spec.rb` to verify VM detachment for both stacks.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for ca387b6f933aed1e5131219e22d125ca360d116b. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/prog/ai/inference_endpoint_replica_nexus_spec.rb
+++ b/spec/prog/ai/inference_endpoint_replica_nexus_spec.rb
@@ -330,7 +330,7 @@ RSpec.describe Prog::Ai::InferenceEndpointReplicaNexus do
     end
 
     it "sets hops to wait when vm is in active set of load balancer" do
-      LoadBalancerVmPort.first.update(state: "up")
+      LoadBalancerVmPort.dataset.update(state: "up")
       expect { nx.wait_endpoint_up }.to hop("wait")
     end
   end
--- a/spec/prog/ai/inference_router_replica_nexus_spec.rb
+++ b/spec/prog/ai/inference_router_replica_nexus_spec.rb
@@ -169,7 +169,7 @@ RSpec.describe Prog::Ai::InferenceRouterReplicaNexus do
     end
 
     it "sets hops to wait when vm is in active set of load balancer" do
-      LoadBalancerVmPort.first.update(state: "up")
+      LoadBalancerVmPort.dataset.update(state: "up")
       expect { nx.wait_router_up }.to hop("wait")
     end
   end
--- a/spec/prog/vnet/update_load_balancer_node_spec.rb
+++ b/spec/prog/vnet/update_load_balancer_node_spec.rb
@@ -65,8 +65,9 @@ RSpec.describe Prog::Vnet::UpdateLoadBalancerNode do
       end
 
       it "removes the VM from load balancer if the VM is detaching" do
-        LoadBalancerVmPort.where(id: lb.vm_ports_dataset.map(&:id)).update(state: "detaching")
-        expect(lb).to receive(:remove_vm_port).with(lb.vm_ports_dataset.first)
+        LoadBalancerVmPort.dataset.update(state: "detaching")
+        expect(lb).to receive(:remove_vm_port).with(lb.vm_ports_dataset.find { |lvp| lvp.stack == "ipv4" })
+        expect(lb).to receive(:remove_vm_port).with(lb.vm_ports_dataset.find { |lvp| lvp.stack == "ipv6" })
         expect { nx.update_load_balancer }.to hop("remove_load_balancer")
       end
     end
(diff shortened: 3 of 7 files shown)


migrate/20251106_drop_index_lbp_lbvm.rb
jeremyevans, line 1: Please split into three migrations, one that sets the column default and adds the column (inside a transaction), and the other two to add the indexes concurrently (in separate migrations, each without a transaction). The reason for this is you want to avoid a situation where a migration is only partially completed. They can be included in the same commit. Name or number the migrations such that the new unique index is added before the old unique index is dropped.

spec/prog/ai/inference_endpoint_replica_nexus_spec.rb
```diff
@@ -330,7 +330,7 @@
     end
 
     it "sets hops to wait when vm is in active set of load balancer" do
-      LoadBalancerVmPort.first.update(state: "up")
+      LoadBalancerVmPort.all.each { |lvp| lvp.update(state: "up") }
```
jeremyevans:
```suggestion
      LoadBalancerVmPort.dataset.update(state: "up")
```

spec/prog/ai/inference_router_replica_nexus_spec.rb
```diff
@@ -169,7 +169,7 @@
     end
 
     it "sets hops to wait when vm is in active set of load balancer" do
-      LoadBalancerVmPort.first.update(state: "up")
+      LoadBalancerVmPort.all.each { |lvp| lvp.update(state: "up") }
```
jeremyevans:
```suggestion
      LoadBalancerVmPort.dataset.update(state: "up")
```

spec/prog/vnet/update_load_balancer_node_spec.rb
```diff
@@ -65,8 +65,9 @@
       end
 
       it "removes the VM from load balancer if the VM is detaching" do
-        LoadBalancerVmPort.where(id: lb.vm_ports_dataset.map(&:id)).update(state: "detaching")
-        expect(lb).to receive(:remove_vm_port).with(lb.vm_ports_dataset.first)
+        LoadBalancerVmPort.all.map { it.update(state: "detaching") }
```
jeremyevans:
```suggestion
        LoadBalancerVmPort.dataset.update(state: "detaching")
```

migrate/20251104_add_stack_to_lb_vm_port.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
```
jeremyevans: We want a transaction for this migration.

```suggestion
```

migrate/20251105_add_index_lbp_lbvm_stack.rb
jeremyevans, line 1: For the index migrations, you may be able to use `change` instead of separate `up`/`down`.

PR #4139, opened 2025-11-06
Use stack for health checks - Phase 2

furkansahin:
## Populate ipv4/6 stack values in load_balancer_vm_port -migration-
Ingest new rows for each stack option

## Start using the stack value in health checks in load_balancer_vm_port

----

> [!IMPORTANT]
> This PR updates load balancer health checks to use stack-specific logic for IPv4 and IPv6, including a migration, model changes, and updated tests.
> 
>   - **Migration**:
>     - Adds migration `20251105_add_load_balancer_vm_port_entries_with_stack.rb` to insert `ipv4` and `ipv6` stack entries into `load_balancer_vm_port`.
>     - Deletes these entries on rollback.
>   - **Model Changes**:
>     - `LoadBalancerVmPort` in `load_balancer_vm_port.rb` now uses `stack` to determine health check type in `check_probe()`.
>     - Raises exceptions if stack type is invalid or not enabled.
>     - Updates `check_pulse()` to use `check_probe()` for stack-specific health checks.
>   - **Monitoring**:
>     - Updates `bin/monitor` to filter `LoadBalancerVmPort` by `stack` values `ipv4` and `ipv6`.
>   - **Testing**:
>     - Updates `load_balancer_vm_port_spec.rb` to test `check_probe()` and `check_pulse()` with stack-specific logic.
>     - Ensures exceptions are raised for invalid or disabled stack types.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for be158a95e797f1975948583d95b910f2db881128. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/load_balancer_vm_port.rb
+++ b/model/load_balancer_vm_port.rb
@@ -22,18 +22,11 @@ class LoadBalancerVmPort < Sequel::Model
     }
   end
 
-  def health_check(session:)
-    [
-      check_probe(session, :ipv4),
-      check_probe(session, :ipv6)
-    ]
-  end
-
   def check_probe(session, type)
     if type == :ipv4
-      return "up" unless load_balancer.ipv4_enabled?
+      raise "This entity should not exist: #{ubid}" unless load_balancer.ipv4_enabled?
     elsif type == :ipv6
-      return "up" unless load_balancer.ipv6_enabled?
+      raise "This entity should not exist: #{ubid}" unless load_balancer.ipv6_enabled?
     else
       raise "Invalid type: #{type}"
     end
@@ -58,9 +51,8 @@ class LoadBalancerVmPort < Sequel::Model
   end
 
   def check_pulse(session:, previous_pulse:)
-    reading_ipv4, reading_ipv6 = health_check(session:)
-    reading = (reading_ipv4 == "up" && reading_ipv6 == "up") ? "up" : "down"
-    pulse = aggregate_readings(previous_pulse:, reading:, data: {ipv4: reading_ipv4, ipv6: reading_ipv6})
+    reading = check_probe(session, stack.to_sym)
+    pulse = aggregate_readings(previous_pulse:, reading:)
 
     time_passed_health_check_interval = Time.now - pulse[:reading_chg] > load_balancer.health_check_interval
 
--- a/prog/vnet/update_load_balancer_node.rb
+++ b/prog/vnet/update_load_balancer_node.rb
@@ -12,7 +12,7 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
   end
 
   label def update_load_balancer
-    vm.load_balancer_vm_ports.select { |lvp| lvp.state == "detaching" }.each do |load_balancer_vm_port|
+    vm.load_balancer_vm_ports.select { |lvp| lvp.state == "detaching" }.sort_by(&:stack).each do |load_balancer_vm_port|
       load_balancer.remove_vm_port(load_balancer_vm_port)
     end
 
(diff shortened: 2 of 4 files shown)


migrate/20251105_add_load_balancer_vm_port_entries_with_stack.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    ["ipv6", "ipv4"].each do |stack|
+      DB[:load_balancer_vm_port].insert([:load_balancer_vm_id, :load_balancer_port_id, :state, :stack],
+        DB[:load_balancer_vm_port].select(:load_balancer_vm_id, :load_balancer_port_id, :state, Sequel.lit("'#{stack}'"))
+          .where(load_balancer_port_id: DB[:load_balancer_port].select(:id)
+            .where(load_balancer_id: DB[:load_balancer].select(:id)
+              .where(stack: "dual").or(stack: stack))))
```
jeremyevans: `or` does work in this case, but it's a dangerous approach, because if you add another condition before the `where`, it will not do what you want. It's simpler to use `IN`:

```suggestion
              .where(stack: ["dual", stack])))
```

model/load_balancer_vm_port.rb
```diff
@@ -58,9 +58,14 @@ def health_check_cmd(type)
   end
 
   def check_pulse(session:, previous_pulse:)
-    reading_ipv4, reading_ipv6 = health_check(session:)
-    reading = (reading_ipv4 == "up" && reading_ipv6 == "up") ? "up" : "down"
-    pulse = aggregate_readings(previous_pulse:, reading:, data: {ipv4: reading_ipv4, ipv6: reading_ipv6})
+    pulse = if stack
```
jeremyevans: The migration leaves us in a temporary state where we have multiple load balancer vm ports for the same load balancer vm and port, one with `NULL` stack and one or two with ipv4 or ipv6 stack.

I think that means `check_pulse` will be duplicating work, at least temporarily. While that may not be a problem, we may want to change callers such that if there is a load balancer vm port with a non-NULL stack, then the load balancer vm port with a NULL stack is ignored.

prog/vnet/update_load_balancer_node.rb
```diff
@@ -12,7 +12,7 @@ def before_run
   end
 
   label def update_load_balancer
-    vm.load_balancer_vm_ports.select { |lvp| lvp.state == "detaching" }.each do |load_balancer_vm_port|
+    vm.load_balancer_vm_ports.select { |lvp| lvp.state == "detaching" }.sort_by(&:stack).each do |load_balancer_vm_port|
```
jeremyevans, line 15: Is the order here only important to ensure consistent behavior in the specs?

furkansahin: yes

model/load_balancer_vm_port.rb
furkansahin: I added the filter to the `./bin/monitor`, this way, once we run the migration and deploy the next commit, we'll perform the health checks only on the `load_blaancer_vm_port`s with `stack` column set to either `ipv4` or `ipv6`.

jeremyevans: I think this is fine if we deploy the next commit right after running the migration in the previous commit. There may be a short period where it is monitoring both, but it's unlikely to cause any significant issues.

PR #4140, opened 2025-11-06
Switch to ipv46 specific lb setup - Phase 3

furkansahin:
## Update load balancer node according to specific stack state
We start using the stack information while configuring load balancing.
This way, if the ipv4 stack is up on a node while the ipv6 is down, it
keeps getting the traffic for ipv4 but not for ipv6.

More importantly, we are fixing a sneaky bug here. Previously, using
uniq on load_balancer.active_vm_ports collapsed multiple
load_balancer_vm_ports that shared the same load_balancer_port. This
caused traffic to be routed to only one node instead of multiple.
In some cases, it could even select the active VM’s own port, resulting
in an empty neighbor set and a total routing failure. The uniq property
was added to avoid an undeterministic bug which we solve here by sorting
on the stack column.

## Remove unnecessary ordering since we removed the ports without stack now

----

> [!IMPORTANT]
> Enhances load balancer functionality by fixing port uniqueness bug, enforcing stack-specific configurations, and improving UI and database integrity.
> 
>   - **Behavior**:
>     - Fixes bug in `update_load_balancer_node.rb` by sorting on `stack` to prevent collapsing of `load_balancer_vm_ports` with the same `load_balancer_port`.
>     - Updates `load_balancer_state` in `vm.rb` to handle dual-stack configurations by reporting separate states for IPv4 and IPv6.
>     - Removes nil `stack` filter in `monitor` and updates `LoadBalancerVmPort` to handle non-null `stack` values.
>   - **Database**:
>     - Adds migration `20251107_drop_load_balancer_vm_port_with_nil_stack.rb` to delete rows with null `stack` and enforce non-null constraint.
>   - **UI**:
>     - Updates `vms.erb` to display separate application states for IPv4 and IPv6 when `stack` is "dual".
>   - **Tests**:
>     - Updates tests in `inference_endpoint_replica_nexus_spec.rb`, `inference_router_replica_nexus_spec.rb`, and `update_load_balancer_node_spec.rb` to reflect changes in load balancer behavior and state handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for d532460b9e7835554cdfa9bd8948f8ee46f0ef05. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -16,7 +16,7 @@ class Vm < Sequel::Model
   one_to_many :pci_devices
   one_through_one :load_balancer
   one_to_one :load_balancer_vm
-  many_to_many :load_balancer_vm_ports, join_table: :load_balancers_vms, right_key: :id, right_primary_key: :load_balancer_vm_id, read_only: true, sort: :stack
+  many_to_many :load_balancer_vm_ports, join_table: :load_balancers_vms, right_key: :id, right_primary_key: :load_balancer_vm_id, read_only: true
   many_to_one :vm_host_slice
   many_to_one :location
   one_to_one :aws_instance, key: :id


prog/vnet/update_load_balancer_node.rb
```diff
@@ -30,16 +30,16 @@ def before_run
     pop "load balancer is removed"
   end
 
+  def load_balancer_ports_to_work_on
+    @load_balancer_ports_to_work_on ||= load_balancer.active_vm_ports { |ds| ds.eager(:load_balancer_port, load_balancer_vm: {vm: :nics}) }.sort_by { |vm_port| vm_port.load_balancer_port.src_port }
```
jeremyevans: Probably better to do the sorting in the database:

```suggestion
    @load_balancer_ports_to_work_on ||= load_balancer.active_vm_ports { |ds| ds.eager_graph(:load_balancer_port).eager(load_balancer_vm: {vm: :nics}).order(:src_port) }
```

model/vm.rb
```diff
@@ -51,7 +51,12 @@ def display_gpu
   end
 
   def load_balancer_state
-    load_balancer_vm_ports.first&.state
+    return nil unless load_balancer
+    if load_balancer.stack == "dual"
+      [load_balancer_vm_ports.find { |vm_port| vm_port.stack == "ipv4" }.state, load_balancer_vm_ports.find { |vm_port| vm_port.stack == "ipv6" }.state]
```
jeremyevans:
```suggestion
      %w[ipv4 ipv6].map! { |stack| load_balancer_vm_ports.find { it.stack == stack }.state }
```

model/vm.rb
```diff
@@ -51,7 +51,12 @@ def display_gpu
...
+    else
+      [load_balancer_vm_ports.first.state]
```
jeremyevans, line 58: I don't think we've deleted the the load balancer vm ports with NULL stack by this point. So isn't there a risk this could return the one with a NULL stack? I suppose you could avoid that with an appropriate order on the `load_balancer_vm_ports` association.

migrate/20251107_drop_load_balancer_vm_port_with_nil_stack.rb
```diff
@@ -0,0 +1,24 @@
...
+    SQL
+
+    alter_table(:load_balancer_vm_port) do
+      set_column_not_null :stack
+    end
+  end
+
+  down do
+    alter_table(:load_balancer_vm_port) do
+      drop_constraint :stack_not_null
```
jeremyevans: I would avoid relying on an auto-generated constraint name:


```suggestion
      set_column_allow_null :stack
```

prog/vnet/update_load_balancer_node.rb
furkansahin: thank you

model/vm.rb
furkansahin: thank you

model/vm.rb
furkansahin: fixed it now

PR #4141, opened 2025-11-06
feat: capability to use IAM auth

serprex:
For sdk calls from control plane & s3 access from data plane

----

> [!IMPORTANT]
> Adds IAM authentication support for AWS RDS and S3, with configuration options and model updates for IAM roles and policies.
> 
>   - **Behavior**:
>     - Adds IAM authentication support for RDS and S3 in `db.rb`, using `pg-aws_rds_iam` and IAM roles.
>     - Introduces `aws-sdk-sts` and `pg-aws_rds_iam` gems in `Gemfile` for IAM token generation.
>     - Adds `assume_role` and `iam_role` columns in `migrate/20251110_aws_iam_auth.rb` for IAM role management.
>   - **Configuration**:
>     - Adds `clover_database_rds_iam_auth_enabled` and `aws_postgres_iam_access` options in `config.rb` for IAM auth toggling.
>   - **Models**:
>     - Updates `LocationCredential` to use `Aws::AssumeRoleCredentials` if `assume_role` is set.
>     - Modifies `PostgresServer` and `PostgresTimeline` to handle IAM roles and policies for S3 access.
>   - **Testing**:
>     - Adds and updates RSpec tests in `spec/model/location_credential_spec.rb`, `spec/model/postgres/postgres_server_spec.rb`, and `spec/model/postgres/postgres_timeline_spec.rb` to cover new IAM functionality.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 49926b6b0b3171b2f259eb1764eb44f0e4c7ca0f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,10 @@ group :aws_rds_iam do
   gem "pg-aws_rds_iam", github: "haines/pg-aws_rds_iam", ref: "fb91b9232837e350aa9c8440b7340346adae845e"
 end
 
+group :aws_sts do
+  gem "aws-sdk-sts", "~> 1.12"
+end
+
 group :development do
   gem "awesome_print"
   gem "by", ">= 1.1.0"
--- a/config.rb
+++ b/config.rb
@@ -80,6 +80,7 @@ module Config
   override :dispatcher_queue_size_ratio, 4, float
   override :recursive_tag_limit, 32, int
   override :root, File.expand_path(__dir__), string
+  override :aws_role_session_name, "ubi", string
   override :clover_database_rds_iam_auth_enabled, false, bool
   optional :hetzner_user, string, clear: true
   optional :hetzner_password, string, clear: true
@@ -148,6 +149,7 @@ module Config
   optional :postgres_monitor_database_root_certs, string
   optional :postgres_paradedb_notification_email, string
   optional :postgres_lantern_notification_email, string
+  override :aws_postgres_iam_access, false, bool
 
   # Logging
   optional :database_logger_level, string
--- a/model/location_credential.rb
+++ b/model/location_credential.rb
@@ -3,18 +3,33 @@
 require_relative "../model"
 require "aws-sdk-ec2"
 require "aws-sdk-iam"
+# :nocov:
+require "aws-sdk-sts" if Config.test? ? ENV["CLOVER_FREEZE"] != "1" : Config.aws_postgres_iam_access
+# :nocov:
 
 class LocationCredential < Sequel::Model
   plugin ResourceMethods, encrypted_columns: [:access_key, :secret_key]
   many_to_one :project
   many_to_one :location, key: :id
 
+  def credentials
+    @credentials ||= if assume_role
+      Aws::AssumeRoleCredentials.new(role_arn: assume_role, role_session_name: Config.aws_role_session_name)
+    else
+      Aws::Credentials.new(access_key, secret_key)
+    end
+  end
+
   def client
-    Aws::EC2::Client.new(access_key_id: access_key, secret_access_key: secret_key, region: location.name)
+    @client ||= Aws::EC2::Client.new(region: location.name, credentials:)
   end
 
   def iam_client
-    Aws::IAM::Client.new(access_key_id: access_key, secret_access_key: secret_key, region: location.name)
+    @iam_client ||= Aws::IAM::Client.new(region: location.name, credentials:)
+  end
+
+  def aws_iam_account_id
+    @account_id ||= Aws::STS::Client.new(region: location.name, credentials:).get_caller_identity.account
   end
 end
 
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -329,6 +329,12 @@ class PostgresServer < Sequel::Model
     vm.sshable.cmd("sudo tee /usr/lib/ssl/certs/blob_storage_ca.crt > /dev/null", stdin: timeline.blob_storage.root_certs) unless timeline.aws?
   end
 
+  def attach_s3_policy_if_needed
+    if Config.aws_postgres_iam_access && timeline.aws? && vm.aws_instance.iam_role
+      timeline.location.location_credential.iam_client.attach_role_policy(role_name: vm.aws_instance.iam_role, policy_arn: timeline.aws_s3_policy_arn)
+    end
+  end
+
   FAILOVER_LABELS = ["prepare_for_unplanned_take_over", "prepare_for_planned_take_over", "wait_fencing_of_old_primary", "taking_over"].freeze
 end
 
--- a/model/postgres/postgres_timeline.rb
+++ b/model/postgres/postgres_timeline.rb
@@ -19,12 +19,17 @@ class PostgresTimeline < Sequel::Model
   end
 
   def generate_walg_config
+    walg_credentials = if access_key
+      <<-WALG_CONF
+AWS_ACCESS_KEY_ID=#{access_key}
+AWS_SECRET_ACCESS_KEY=#{secret_key}
+      WALG_CONF
+    end
     <<-WALG_CONF
 WALG_S3_PREFIX=s3://#{ubid}
 AWS_ENDPOINT=#{blob_storage_endpoint}
-AWS_ACCESS_KEY_ID=#{access_key}
-AWS_SECRET_ACCESS_KEY=#{secret_key}
-AWS_REGION: #{aws? ? location.name : "us-east-1"}
+#{walg_credentials}
+AWS_REGION=#{aws? ? location.name : "us-east-1"}
 AWS_S3_FORCE_PATH_STYLE=true
 PGHOST=/var/run/postgresql
     WALG_CONF
@@ -113,8 +118,7 @@ PGHOST=/var/run/postgresql
   def blob_storage_client
     @blob_storage_client ||= aws? ? Aws::S3::Client.new(
       region: location.name,
-      access_key_id: access_key,
-      secret_access_key: secret_key,
+      credentials: location.location_credential.credentials,
       endpoint: blob_storage_endpoint,
       force_path_style: true
     ) : Minio::Client.new(
@@ -172,6 +176,16 @@ PGHOST=/var/run/postgresql
         }
       })
     : blob_storage_client.set_lifecycle_policy(ubid, ubid, BACKUP_BUCKET_EXPIRATION_DAYS)
+  end
+
+  alias_method :aws_s3_policy_name, :ubid
+
+  def aws_iam_account_id
+    location.location_credential.aws_iam_account_id
+  end
+
+  def aws_s3_policy_arn
+    "arn:aws:iam::#{aws_iam_account_id}:policy/#{aws_s3_policy_name}"
   end
 end
 
--- a/prog/postgres/postgres_timeline_nexus.rb
+++ b/prog/postgres/postgres_timeline_nexus.rb
@@ -48,7 +48,7 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
   end
 
   label def setup_bucket
-    nap 1 if postgres_timeline.aws? && !aws_access_key_is_available?
+    nap 1 if postgres_timeline.aws? && !Config.aws_postgres_iam_access && !aws_access_key_is_available?
 
     # Create bucket for the timeline
     postgres_timeline.create_bucket
@@ -116,15 +116,19 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
   end
 
   def destroy_aws_s3
-    iam_client.list_attached_user_policies(user_name: postgres_timeline.ubid).attached_policies.each do
-      iam_client.detach_user_policy(user_name: postgres_timeline.ubid, policy_arn: it.policy_arn)
-      iam_client.delete_policy(policy_arn: it.policy_arn)
-    end
+    if Config.aws_postgres_iam_access
+      iam_client.delete_policy(policy_arn: postgres_timeline.aws_s3_policy_arn)
+    else
+      iam_client.list_attached_user_policies(user_name: postgres_timeline.ubid).attached_policies.each do
+        iam_client.detach_user_policy(user_name: postgres_timeline.ubid, policy_arn: it.policy_arn)
+        iam_client.delete_policy(policy_arn: it.policy_arn)
+      end
 
-    iam_client.list_access_keys(user_name: postgres_timeline.ubid).access_key_metadata.each do
-      iam_client.delete_access_key(user_name: postgres_timeline.ubid, access_key_id: it.access_key_id)
+      iam_client.list_access_keys(user_name: postgres_timeline.ubid).access_key_metadata.each do
+        iam_client.delete_access_key(user_name: postgres_timeline.ubid, access_key_id: it.access_key_id)
+      end
+      iam_client.delete_user(user_name: postgres_timeline.ubid)
     end
-    iam_client.delete_user(user_name: postgres_timeline.ubid)
   end
 
   def setup_blob_storage
@@ -137,12 +141,14 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
   end
 
   def setup_aws_s3
-    iam_client.create_user(user_name: postgres_timeline.ubid)
-    policy = iam_client.create_policy(policy_name: postgres_timeline.ubid, policy_document: postgres_timeline.blob_storage_policy.to_json)
-    iam_client.attach_user_policy(user_name: postgres_timeline.ubid, policy_arn: policy.policy.arn)
-    response = iam_client.create_access_key(user_name: postgres_timeline.ubid)
-    postgres_timeline.update(access_key: response.access_key.access_key_id, secret_key: response.access_key.secret_access_key)
-    postgres_timeline.leader.incr_refresh_walg_credentials
+    policy = iam_client.create_policy(policy_name: postgres_timeline.aws_s3_policy_name, policy_document: postgres_timeline.blob_storage_policy.to_json)
+    unless Config.aws_postgres_iam_access
+      iam_client.create_user(user_name: postgres_timeline.ubid)
+      iam_client.attach_user_policy(user_name: postgres_timeline.ubid, policy_arn: policy.policy.arn)
+      response = iam_client.create_access_key(user_name: postgres_timeline.ubid)
+      postgres_timeline.update(access_key: response.access_key.access_key_id, secret_key: response.access_key.secret_access_key)
+      postgres_timeline.leader.incr_refresh_walg_credentials
+    end # the policy is later attached at the postgres_server level
   end
 
   def aws_access_key_is_available?
@@ -150,7 +156,7 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
   end
 
   def iam_client
-    @iam_client ||= Aws::IAM::Client.new(access_key_id: postgres_timeline.location.location_credential.access_key, secret_access_key: postgres_timeline.location.location_credential.secret_key, region: postgres_timeline.location.name)
+    postgres_timeline.location.location_credential.iam_client
   end
 
   def admin_client
--- a/prog/vm/aws/nexus.rb
+++ b/prog/vm/aws/nexus.rb
@@ -210,7 +210,7 @@ class Prog::Vm::Aws::Nexus < Prog::Base
     az_id = subnet_response.subnets.first.availability_zone_id
     ipv4_dns_name = instance.public_dns_name
 
-    AwsInstance.create_with_id(vm, instance_id:, az_id:, ipv4_dns_name:)
+    AwsInstance.create_with_id(vm, instance_id:, az_id:, ipv4_dns_name:, iam_role: role_name)
 
     hop_wait_instance_created
   end
@@ -334,6 +334,14 @@ class Prog::Vm::Aws::Nexus < Prog::Base
 
       ignore_invalid_entity do
         iam_client.delete_policy({policy_arn: cloudwatch_policy.arn})
+      end
+    end
+
+    if Config.aws_postgres_iam_access
+      iam_client.list_attached_role_policies(role_name:).attached_policies.each do |policy|
+        ignore_invalid_entity do
+          iam_client.detach_role_policy(role_name:, policy_arn: policy.policy_arn)
+        end
       end
     end
 
--- a/spec/model/location_credential_spec.rb
+++ b/spec/model/location_credential_spec.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative "spec_helper"
+
+RSpec.describe LocationCredential do
+  subject(:location_credential) {
+    described_class.create_with_id(location.id, assume_role: "assume-role")
+  }
+
+  let(:location) {
+    Location.create(
+      name: "test-location",
+      display_name: "display-name",
+      ui_name: "ui-name",
+      visible: true,
+      provider: "aws"
+    )
+  }
+
+  it "uses Aws::AssumeRoleCredentials when assume_role set" do
+    creds = instance_double(Aws::AssumeRoleCredentials)
+    expect(Aws::AssumeRoleCredentials).to receive(:new).with(role_arn: "assume-role", role_session_name: Config.aws_role_session_name).and_return(creds)
+    expect(location_credential.credentials).to be(creds)
+  end
+
+  it "gets account id from sts" do
+    if ENV["CLOVER_FREEZE"] != "1"
+      sts_client = Aws::STS::Client.new(stub_responses: true)
+      expect(Aws::STS::Client).to receive(:new).and_return(sts_client).at_least(:once)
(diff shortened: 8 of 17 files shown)


serprex:

migrate/20251120_aws_iam_auth.rb
jeremyevans, line 1: Migrations should be split into their own commit (generally the first commit in the PR) that doesn't have code changes (it should include the schema annotation and `cache/` changes).

model/account.rb
```diff
@@ -19,8 +19,8 @@ def provider_names
     identities.map(&:provider).join(", ")
   end
 
-  def create_project_with_default_policy(name, default_policy: true)
-    project = Project.create(name: name)
+  def create_project_with_default_policy(name, default_policy: true, project_id: "")
+    project = Project.create(name: name, id: (project_id.nil? || project_id.empty?) ? nil : project_id)
```
jeremyevans: We shouldn't use empty string to signal missing information.

```suggestion
  def create_project_with_default_policy(name, default_policy: true, project_id: nil)
    project = Project.new(name:)
    project.id = project_id if project_id
    project.save_changes
```

migrate/20251106_aws_iam_auth.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:location_credential) do
+      add_column :assume_role, String, null: false, default: ""
```
jeremyevans: It's better to signal missing information with NULL than with empty string.  Same with the `iam_role` column. If you want to avoid ambiguity in terms of handling empty string, add a check constraint that disallows the empty string.

```suggestion
      add_column :assume_role, String
```

model/location_credential.rb
```diff
@@ -9,20 +9,29 @@ class LocationCredential < Sequel::Model
   many_to_one :project
   many_to_one :location, key: :id
 
+  def credentials
+    @credentials ||= if !assume_role.empty?
+      Aws::AssumeRoleCredentials.new(role_arn: assume_role, role_session_name: ENV["HOSTNAME"] || "ubi")
```
jeremyevans: We should probably use a `Config` method instead of direct access to an environment variable here. You can use the `override` feature to provide the default.  See `config.rb`.

model/location_credential.rb
```diff
@@ -9,20 +9,29 @@ class LocationCredential < Sequel::Model
...
+    else
+      Aws::Credentials.new(access_key_id: access_key, secret_access_key: secret_key)
+    end
+  end
+
   def client
-    Aws::EC2::Client.new(access_key_id: access_key, secret_access_key: secret_key, region: location.name)
+    Aws::EC2::Client.new(region: location.name, credentials: credentials)
```
jeremyevans: For new code, we prefer hash value omission (also applies to next method, and maybe additional methods in the PR):

```suggestion
    Aws::EC2::Client.new(region: location.name, credentials:)
```

config.rb
```diff
@@ -40,6 +40,7 @@ def self.test?
 
   # Mandatory -- exception is raised for these variables when missing.
   mandatory :clover_database_url, string, clear: true
+  optional :clover_database_rds_iam_auth_enabled, bool, clear: true
```
jeremyevans: No point in `clear` on a boolean setting. I recommend using `override` and having a default of `false`.

Gemfile
```diff
@@ -28,6 +28,7 @@ gem "octokit"
 gem "omniauth-github"
 gem "omniauth-google-oauth2"
 gem "pagerduty", ">= 4.0"
+gem "pg-aws_rds_iam", github: "haines/pg-aws_rds_iam", ref: "fb91b9232837e350aa9c8440b7340346adae845e"
```
jeremyevans: It would be better if we could set up the Gemfile so that this is only used if needed. However, that applies to other optional features (e.g. pagerduty, omniauth), and not related to this PR, so we can do that later.

fdr:
general feedback:

Numerical limitations on IAM users and roles per account and usable only for small-to-medium sized installations: it's quite easy for an operation to grow past 5000 (the nominal hard limit) on databases, since slicing buckets up with IAM requires a templated policy for each. You also need a read-only policy for standbys, or at least, I customarily would do things that way for even more robustness (at the cost of having to complete a STS call during promotion).

Instead, as it comes to database archives, using STS to write out a file with `{key, secret, token}` with 36 hour token and refreshing them every hour or two (in case something goes wrong, to have that extra time) should let you grow beyond 5000 such policies per account.

The control plane can source its own credential to call STS via IAM roles though, it does not need a jillion little templated policies for its own operation.

model/location_credential.rb
fdr: I'm surprised there's not a rubocop rewrite for this...yet?

iamKunalGupta:

Gemfile
jeremyevans: I did some more testing on this, and while I don't think we are ready to use Gemfile groups for our existing dependencies to enable conditional loading, we should do so for new dependencies that we would not be using in production. In addition to using a group in the Gemfile, the code should be written so that everything works if the related library cannot be required.

The easiest way to ensure that things work both ways is to ensure the `pg-aws_rds_iam` library is loaded during unfrozen testing, but not loaded during frozen testing (the default rake task runs specs in both frozen and unfrozen modes). You can check `ENV["CLOVER_FREEZE"] == "1"` in the specs. If it is set, do not test the `pg-aws_rds_iam` code. In the Rakefile, use `BUNDLER_WITHOUT` to prevent loading of the related Gemfile group when running frozen specs.

serprex: I'd prefer for aws grouping to be a separate PR, could precede this one. I'm not clear on this freeze mechanism. Not clear whether you think this `pg-aws_rds_iam` should be part of a wider aws group, or optional on its own

jeremyevans: A separate PR that precedes this one is fine. `pg-aws_rds_iam` should be part of it's own group. We want to ensure we have a way to exclude loading it in our production environment. That means the application needs to run correctly without requiring `pg-aws_rds_iam` if the support isn't configured.

As to the freeze mechanism, `rake` will run the test suite twice, first in a development-style mode with coverage without freezing all core classes, and second in a production-style mode without coverage with freezing all core classes. The development-style mode tests should require `pg-aws_rds_iam` and test the support. The production-style mode tests should ensure `pg-aws_rds_iam` is not required, not test the support (or test that support is disabled), and still run correctly.

serprex: we're already only requiring `pg-aws_rds_iam` when enabled by config. I'm not sure how your production intends to setup `BUNDLER_WITHOUT`

Split out to a group `aws_rds_iam`, but it seems like what you're outlining you'd want this outside freeze in your production, but we'd want it part of freeze in our production

serprex: split out into #4174 & removed that code from this PR

model/account.rb
```diff
@@ -19,8 +19,10 @@ def provider_names
     identities.map(&:provider).join(", ")
   end
 
-  def create_project_with_default_policy(name, default_policy: true)
-    project = Project.create(name: name)
+  def create_project_with_default_policy(name, default_policy: true, project_id: nil)
+    project = Project.new(name:)
+    project.id = project_id if project_id
+    project.save_changes
```
jeremyevans: Can you explain why you need this? It's not used elsewhere in the code, only in a spec, and the commit message doesn't describe why it was added.

Maybe it would be better to extract  an `add_project_with_default_policy(project)` method. That way you could create the `project` and pass it in. That would be best as a separate commit.

In general, for non-obvious changes, the commit message should discuss the rationale. In this PR, the migration's commit message is quite good, but the second commit message does not describe the reason behind the changes.

config.rb
```diff
@@ -80,6 +80,7 @@ def self.test?
   override :dispatcher_queue_size_ratio, 4, float
   override :recursive_tag_limit, 32, int
   override :root, File.expand_path(__dir__), string
+  override :hostname, "ubi", string
```
jeremyevans: `hostname` seems vague. If this configuration value is specific to the AWS IAM code:

```suggestion
  override :aws_iam_hostname, "ubi", string
```

model/location_credential.rb
```diff
@@ -3,26 +3,44 @@
 require_relative "../model"
 require "aws-sdk-ec2"
 require "aws-sdk-iam"
+require "aws-sdk-sts"
```
jeremyevans: If the `sts_client` method should only be used if the AWS IAM support is configured, this should be guarded:

The reason for checking `CLOVER_FREEZE` in the guard is we want to ensure that everything runs correctly without the library loaded. The easiest way to do that is to load it during unfrozen testing, but not load it during frozen testing. That will likely cause a few of the specs to fail. You can add similar guards to the affected specs so they are not run in frozen testing mode.

```suggestion
require "aws-sdk-sts" if Config.test? ? ENV["CLOVER_FREEZE"] != "1" : Config.aws_postgres_iam_access
```

Gemfile
```diff
@@ -12,6 +12,7 @@ gem "autoforme", ">= 1.14"
 gem "aws-sdk-ec2", "~> 1.512"
 gem "aws-sdk-iam"
 gem "aws-sdk-s3"
+gem "aws-sdk-sts", "~> 1.12"
```
jeremyevans: This should be moved to it's own Gemfile group so users who are not using it do not have to load it. Probably best to use the same Gemfile group you are creating in #4714. To avoid conflicts, you may want to rebase this PR on top of that PR.

model/location_credential.rb
```diff
@@ -3,26 +3,44 @@
...
   def iam_client
-    Aws::IAM::Client.new(access_key_id: access_key, secret_access_key: secret_key, region: location.name)
+    Aws::IAM::Client.new(region: location.name, credentials:)
+  end
+
+  def sts_client
+    Aws::STS::Client.new(region: location.name, credentials:)
+  end
+
+  def get_account_id
```
jeremyevans: Idiomatic Ruby generally avoids `get_*` method naming (other places in the commit should be updated as well). If this is specific to AWS IAM account ids:

```suggestion
  def aws_iam_account_id
```

model/postgres/postgres_server.rb
```diff
@@ -329,6 +329,24 @@ def refresh_walg_credentials
     vm.sshable.cmd("sudo tee /usr/lib/ssl/certs/blob_storage_ca.crt > /dev/null", stdin: timeline.blob_storage.root_certs) unless timeline.aws?
   end
 
+  def needs_s3_policy_attachment?
```
jeremyevans: In general, we avoid creating methods for information only needed in a single place. If you don't have multiple callers, it's best to inline the information where you need it. If it is needed in multiple places later, we can add a method to abstract it at that point.

For `s3_policy_arn` and `get_account_id`, those end up calling the STS code, and since that will not work once the STS require is made conditional if the support is not enabled, it's best to not have unguarded methods that rely on accessing it.

I would combine all 4 methods added into a single `attach_s3_policy_if_needed` method.

model/postgres/postgres_timeline.rb
```diff
@@ -19,12 +19,19 @@ def bucket_name
   end
 
   def generate_walg_config
+    walg_credentials = if access_key
+      <<-WALG_CONF
+AWS_ACCESS_KEY_ID=#{access_key}
+AWS_SECRET_ACCESS_KEY=#{secret_key}
+      WALG_CONF
+    else
+      ""
```
jeremyevans: Unneeded as `nil.to_s` is `""`.

```suggestion
```

model/postgres/postgres_timeline.rb
```diff
@@ -171,6 +177,10 @@ def set_lifecycle_policy
       })
     : blob_storage_client.set_lifecycle_policy(ubid, ubid, BACKUP_BUCKET_EXPIRATION_DAYS)
   end
+
+  def aws_s3_policy_name
+    ubid
+  end
```
jeremyevans:
```suggestion
  alias_method :aws_s3_policy_name, :ubid
```

model/location_credential.rb
```diff
@@ -3,26 +3,44 @@
...
-    Aws::EC2::Client.new(access_key_id: access_key, secret_access_key: secret_key, region: location.name)
+    Aws::EC2::Client.new(region: location.name, credentials:)
   end
 
   def iam_client
-    Aws::IAM::Client.new(access_key_id: access_key, secret_access_key: secret_key, region: location.name)
+    Aws::IAM::Client.new(region: location.name, credentials:)
+  end
+
+  def sts_client
```
jeremyevans: I would combine this into the `account_id` method below, since it isn't called by other places in the code, only in a spec.

spec/model/postgres/postgres_server_spec.rb
```diff
@@ -468,4 +468,21 @@
       expect { postgres_server.refresh_walg_credentials }.not_to raise_error
     end
   end
+
+  describe "#attach_s3_policy_if_needed" do
+    it "calls attach_role_policy when needs_s3_policy_attachment? true" do
+      iam_client = Aws::IAM::Client.new(stub_responses: true)
+      expect(postgres_server).to receive(:timeline).and_return(instance_double(PostgresTimeline, aws?: true, aws_s3_policy_name: "policy-name")).at_least(:once)
```
jeremyevans: We are trying to avoid additional unnecessary mocking/stubbing (and slowly cleaning up the unnecessary mocking and stubbing we already have, see #4188). Mocking/stubbing AWS access such as `Aws::IAM::Client` is fine, as is mocking/stubbing SSH access. However, database access should not be mocked/stubbed. Use a real PostgresTimeline, AwsInstance, and LocationCredential object, not an `instance_double`. 

For new specs added in this commit, similar changes should be made. When modifying existing specs for this commit, if they already use mocked database access, it's fine to continue using the mocking.

model/project_quota.rb
```diff
@@ -13,7 +13,7 @@ def self.freeze
   # :nocov:
 
   def self.default_quotas
-    @default_quotas ||= YAML.load_file("config/default_quotas.yml").each_with_object({}) do |item, hash|
+    @default_quotas ||= YAML.load_file(Config.default_quotas_file_path).each_with_object({}) do |item, hash|
```
jeremyevans: This seems unrelated to the AWS IAM changes and it and the related `config.rb` change should be split into a separate commit.

spec/spec_helper.rb
```diff
@@ -98,6 +98,8 @@ def Object.method_added(method)
     # a real object. This is generally recommended, and will default to
     # `true` in RSpec 4.
     mocks.verify_partial_doubles = true
+    # Prevents expect(nil).to receive(...)
+    mocks.allow_message_expectations_on_nil = false
```
jeremyevans: This should be in a separate commit with it's own justification. Which expectation needs this? Is there an alternative approach that avoids the need to add it?

jeremyevans (commented):
Apologies for the delay in rereviewing this. Please let me know if you have any questions about any of the inline comments.

Gemfile
serprex: I don't think they should be grouped together. `aws-sdk-sts` is what CloudWatch / S3 access wants to rely on eventually, while #4174 is specific for when controlplane & its databases speak IAM

jeremyevans: Understood. A distinct group will be fine in that case.

spec/spec_helper.rb
serprex: I ran into this while writing a buggy tests, & none of your specs use `expect(nil)` already. Split out #4204

model/project_quota.rb
serprex: Removing. Was having issues refreshing mounted file without restarting pod. But doesn't matter anymore

config.rb
serprex: this was selected after code was trying to access `ENV["HOSTNAME"]` which would be tied to HOSTNAME in bash, but that's pretty implementation specific on our end so instead renamed to `aws_role_session_name`

model/account.rb
serprex: I cannot explain, & have reverted this. Will copy/paste/edit out-of-tree

spec/model/postgres/postgres_server_spec.rb
serprex: I think #4188 is the better approach, when much of this test is using mocks against your advice because I was copying code which was using mocks in such a way already. I'd prefer to submit a followup PR like #4188 here. But I guess I'll have to split out a preceding PR that gets away from mocks in this file

jeremyevans: Keeping the copied mocks in this PR and submitting a followup PR like #4188 is OK.

serprex: #4222

serprex:
@jeremyevans took a stab at making sts an optional group, but not really understanding how this is supposed to work

model/location_credential.rb
```diff
@@ -3,7 +3,7 @@
 require_relative "../model"
 require "aws-sdk-ec2"
 require "aws-sdk-iam"
-require "aws-sdk-sts"
+require "aws-sdk-sts" if Config.test? ? ENV["CLOVER_FREEZE"] != "1" : Config.aws_postgres_iam_access
```
jeremyevans, line 7: I noticed the PR is failing the coverage tests, and it is probably because of this line (I didn't check to confirm). You should be able to fix it by adding `# :nocov:` before and after this line. This fails coverage testing because the second branch is never taken (as `Config.test?` is always true during the tests).

Gemfile
```diff
@@ -51,6 +50,10 @@ gem "tilt", ">= 2.6.1"
 gem "warning"
 gem "webauthn"
 
+group :aws_sts do
```
jeremyevans, line 57: This group looks fine.

jeremyevans (commented):
This looks good to me. Can you please rebase and split into separate commits per logical change, with appropriate commit messages? Then this should be ready for merging.

model/location_credential.rb
```diff
@@ -3,26 +3,40 @@
...
 
 # Table: location_credential
 # Columns:
-#  access_key | text | NOT NULL
-#  secret_key | text | NOT NULL
-#  id         | uuid | PRIMARY KEY
+#  access_key  | text | NOT NULL
+#  secret_key  | text | NOT NULL
+#  id          | uuid | PRIMARY KEY
+#  assume_role | text |
```
furkansahin: considering we should either pass `[access_key,secret_key]` tuple OR `assume_role`, maybe we should remove the `NOT NULL` constraint from the access_key,secret_key columns and add a constraint like the following:
```
CHECK (
    (
        access_key IS NOT NULL
        AND secret_key IS NOT NULL
        AND assume_role IS NULL
    )
    OR (
        access_key IS NULL
        AND secret_key IS NULL
        AND assume_role IS NOT NULL
    )
);
```

model/postgres/postgres_server.rb
```diff
@@ -329,6 +329,20 @@ def refresh_walg_credentials
...
+    end
+  end
+
+  def s3_policy_arn
+    "arn:aws:iam::#{aws_iam_account_id}:policy/#{timeline.aws_s3_policy_name}"
+  end
+
+  def aws_iam_account_id
+    vm.location.location_credential.aws_iam_account_id
+  end
```
jeremyevans: I think it might be better to move this code to PostgresTimeline. It would eliminate the duplicate policy_arn string in postgres_timeline_nexus. The timeline location should match the vm location.

Could be done in a later PR, though.

prog/vm/aws/nexus.rb
```diff
@@ -337,6 +337,14 @@ def before_run
       end
     end
 
+    iam_client.list_attached_role_policies(role_name:).attached_policies.tap {
```
jeremyevans: Is this only required if `Config.aws_postgres_iam_access` is set? My guess is if it isn't set, there aren't attached policies, so it's basically a no-op in that case. We could consider making this conditional, but that could be done in a future PR.

spec/model/location_credential_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  it "uses Aws::AssumeRoleCredentials when assume_role set" do
+    creds = instance_double(Aws::AssumeRoleCredentials)
+    expect(Aws::AssumeRoleCredentials).to receive(:new).with(role_arn: "assume-role", role_session_name: Config.aws_role_session_name).and_return(creds)
+    expect(location_credential.credentials).to be(creds)
+  end
+
+  it "gets account id from sts" do
+    skip unless ENV["CLOVER_FREEZE"] != "1"
```
jeremyevans: `skip` makes the spec show up as pending in the output. It's OK for now, but we should use a conditional around the spec, so the spec (or spec class) is never executed. Could be done in a future PR.

serprex: assuming you only mean to wrap test contents with `if` instead of using `skip`, this is done

migrate/20251202_aws_iam_auth.rb
```diff
@@ -0,0 +1,36 @@
...
+          (access_key is not null and secret_key is not null and assume_role is null)
+          or (access_key is null and secret_key is null and assume_role is not null)
+        );
+    SQL
+  end
+
+  down do
+    run <<~SQL
+      alter table location_credential alter column access_key drop not null, alter column secret_key set not null,
+        drop constraint location_credential_single_auth_mechanism;
```
furkansahin: You want to fix the `access_key drop not null`.

Use:
```
ALTER TABLE location_credential
ALTER COLUMN access_key SET NOT NULL,
ALTER COLUMN secret_key SET NOT NULL,
DROP CONSTRAINT location_credential_single_auth_mechanism;
```
Can you beautify the query above as well?

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -137,20 +141,22 @@ def setup_blob_storage
...
+    end # the policy is later attached at the postgres_server level
   end
 
   def aws_access_key_is_available?
     iam_client.list_access_keys(user_name: postgres_timeline.ubid).access_key_metadata.any? { it.access_key_id == postgres_timeline.access_key }
   end
 
   def iam_client
-    @iam_client ||= Aws::IAM::Client.new(access_key_id: postgres_timeline.location.location_credential.access_key, secret_access_key: postgres_timeline.location.location_credential.secret_key, region: postgres_timeline.location.name)
+    postgres_timeline.location.location_credential.iam_client
```
furkansahin, line 159: is there a reason to not cache the result? Can you update this with:
```
@iam_client ||= postgres_timeline.location.location_credential.iam_client
```

serprex: seems like caching should be done at location_credential level, will add there

model/location_credential.rb
```diff
@@ -3,27 +3,45 @@
...
+    @credentials ||= if assume_role
+      Aws::AssumeRoleCredentials.new(role_arn: assume_role, role_session_name: Config.aws_role_session_name)
+    else
+      Aws::Credentials.new(access_key, secret_key)
+    end
+  end
+
   def client
-    Aws::EC2::Client.new(access_key_id: access_key, secret_access_key: secret_key, region: location.name)
+    @client ||= Aws::EC2::Client.new(region: location.name, credentials:)
```
serprex, line 24: @furkansahin I also added caching client here to be consistent with your iam_client feedback, but wanted to point this out in case there's a good reason this wasn't cached previously

furkansahin: Not really, I basically missed it, thanks for adding

spec/model/postgres/postgres_server_spec.rb
```diff
@@ -519,4 +519,27 @@ def create_standby_resource(suffix)
       expect { postgres_server.refresh_walg_credentials }.not_to raise_error
     end
   end
+
+  describe "#attach_s3_policy_if_needed" do
+    before do
+      skip unless ENV["CLOVER_FREEZE"] != "1"
```
jeremyevans: Still uses `skip` instead of wrapping with `if`.

migrate/20251202_aws_iam_auth.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    alter_table(:aws_instance) do
+      add_column :iam_role, String
+    end
+
+    run <<~SQL
+      ALTER TABLE location_credential
+        ALTER COLUMN access_key DROP NOT NULL,
+        ALTER COLUMN secret_key DROP NOT NULL,
+        ADD CONSTRAINT location_credential_single_auth_mechanism CHECK (
```
jeremyevans, line 17: Sequel supports all of this, so you could use the Sequel API and a `change` migration instead of a manual `down` migration. However, since the `down` migration is already written and looks correct, probably not worth it at this point.

spec/model/postgres/postgres_server_spec.rb
serprex: should I be wrapping the entire `describe` here?

jeremyevans: Yes. That avoids a separate pending example warning for every spec example inside the `describe`.

PR #4036, opened 2025-10-15
Add Extra Info for some Models in Admin

aamederen:
This change adds some logic for showing extra information about different models in admin panel. This can be useful for many things such as quick metrics, health results as well as stripe data and invoice download links as implemented in this example.

----

> [!IMPORTANT]
> Enhance admin panel to display extra details for `BillingInfo` and `Invoice`, including download links for invoices.
> 
>   - **Admin Panel Enhancements**:
>     - Adds `OBJECTS_WITH_EXTRAS` in `clover_admin.rb` to track models with extra info views.
>     - Updates `object.erb` to render extra details for models listed in `OBJECTS_WITH_EXTRAS`.
>   - **Invoice Model**:
>     - Adds `generate_download_link` method to `Invoice` for generating presigned S3 URLs.
>     - Aliases `admin_label` to `invoice_number`.
>   - **Views**:
>     - Adds `BillingInfo.erb` to display Stripe data.
>     - Adds `Invoice.erb` to display download link if available.
>   - **Tests**:
>     - Adds tests in `invoice_spec.rb` for `generate_download_link`.
>     - Adds tests in `admin_spec.rb` to verify rendering of extra details and download links.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 88b89dbae7ddea3223311141eaa3990149a24810. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #4145, opened 2025-11-07
Introduce annotation

mohi-kalantari:
**Add migration for annotations**

Annotations are records used as non-urgent pages describing an issue. They appear at the bottom of the admin page and are not integrated with PagerDuty. Simple actions like destroy will be added to the admin page actions.

**Add Annotation model**

**Show annotations at the bottom of the admin page**

Annotations will be rendered and shown in the admin index page if any.

----

> [!IMPORTANT]
> Introduce `Annotation` model to admin interface for non-urgent issue tracking, with display and destroy capabilities.
> 
>   - **Behavior**:
>     - Adds `Annotation` model in `annotation.rb` with fields `id`, `description`, `related_resources`, and `created_at`.
>     - Annotations are displayed at the bottom of the admin page in `index.erb` if any exist.
>     - Adds destroy action for annotations in `clover_admin.rb`.
>   - **Migration**:
>     - Adds migration `20251110_add_annotation_model.rb` to create `annotation` table.
>   - **Tests**:
>     - Adds tests in `admin_spec.rb` to verify listing, linking, and destroying annotations.
>   - **Misc**:
>     - Updates `ubid.rb` to include `TYPE_ANNOTATION`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 2f4b27d22a1b796b3bfd61014403fe443a54c001. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -193,6 +193,9 @@ class CloverAdmin < Roda
       end,
       "reset" => object_action("Hardware Reset", "Hardware reset scheduled for VmHost", &:incr_hardware_reset),
       "reboot" => object_action("Reboot", "Reboot scheduled for VmHost", &:incr_reboot)
+    },
+    "Annotation" => {
+      "destroy" => object_action("Destroy", "Destroyed the annotation", "/", &:destroy)
     }
   }.freeze
   OBJECT_ACTIONS.each_value(&:freeze)
--- a/views/admin/object.erb
+++ b/views/admin/object.erb
@@ -19,6 +19,12 @@
   <% end %>
 </div>
 
+<div id="annotation-info">
+  <% if @obj.is_a?(Annotation) %>
+    <%== form({action: "/model/Annotation/#{@obj.ubid}/destroy", method: :post}, button: "Destroy Annotation") %>
+  <% end %>
+</div>
+
 <% if @klass.associations.include?(:semaphores) && !(semaphores = @obj.semaphores_dataset.select_order_map(:name)).empty? %>
   <p>Semaphores Set: <%= semaphores.join(", ") %></p>
 <% end %>
--- a/spec/routes/web/admin/admin_spec.rb
+++ b/spec/routes/web/admin/admin_spec.rb
@@ -644,4 +644,13 @@ RSpec.describe CloverAdmin do
     created_at_cell = cells[2]
     expect(created_at_cell).to have_content(/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/)
   end
+
+  it "destroys the annotation when clicked on its button" do
+    annotation = Annotation.create(description: "test-annotation")
+    visit "/"
+    click_link annotation.description
+
+    click_button "Destroy Annotation"
+    expect { annotation.reload }.to raise_error(Sequel::NoExistingObject)
+  end
 end


migrate/20251110_add_annotation_model.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:annotation) do
+      # UBID.to_base32_n("an") => 341
+      column :id, :uuid, primary_key: true, default: Sequel.lit("gen_random_ubid_uuid(341)")
+      column :description, :text
```
jeremyevans, line 8: Might be useful to have a jsonb column or uuid[] column that can be used to reference related objects. These could be autolinked in the Admin UI.

clover_admin.rb
```diff
@@ -414,6 +414,7 @@ def self.object_action(...)
         .flatten
         .select { it < ResourceMethods::InstanceMethods }
         .sort_by(&:name)
+      @annotations = Annotation.all
```
jeremyevans: Should use a consistent order (reverse chronological may be best).

views/admin/index.erb
```diff
@@ -52,4 +52,13 @@
     <li><a href="/change-password">Change Password</a></li>
     <li><a href="/multifactor-manage">Manage Multifactor Authentication</a></li>
   </ul>
+
+  <% if @annotations.any? %>
+    <h2>Annotations</h2>
+    <ul id="annotations-list">
```
jeremyevans: Probably better to have a table display than list display, so you can easily see when, and link directly to related objects (if support is added for them).

views/admin/index.erb
```diff
@@ -52,4 +52,13 @@
...
+      <% @annotations.each do |annotation| %>
+        <li><a href="<%="/model/Annotation/#{annotation.ubid}"%>"><%== annotation.description %></a></li>
```
jeremyevans: Even though we control the descriptions, best to avoid the possibility for XSS (use `<%=`, not `<%==`). Also, avoid the need to interpolate the string inside the `<%=` tag by moving the `/model/Annotation/` before the `<%=`.

fdr:
This was an evolved version of the "slack webhook for boot-id is not the expected boot-id" idea, which was intended for people to know that the system had rebooted unexpectedly. Among other things, of course this breaks all block devices, because we keep the key encrypting key ephemeral and removed from the host, and the control plane will have to something to send the key again.


 I suggested that, in addition to adding a new off-the-wall webhook-slack dependency, that ~nobody was going to think to go to an out of band channel on slack to line up the page to this unusual condition.

While in principle the system can automatically resolve an unexpected reboot, they are rare enough that it may be worth keeping someone in the loop to observe anything unusual about them. For now.

The reason I mention it is, because, it'd help a reviewer think about how it sits in the screen and what it's supposed to do.

clover_admin.rb
```diff
@@ -421,6 +424,7 @@ def self.object_action(...)
         .flatten
         .select { it < ResourceMethods::InstanceMethods }
         .sort_by(&:name)
+      @annotations = Annotation.order(Sequel.desc(:created_at)).all
```
jeremyevans, line 428:
```suggestion
      @annotations = Annotation.reverse(:created_at).all
```

views/admin/index.erb
```diff
@@ -52,4 +52,36 @@
...
   </ul>
+
+  <% if @annotations.any? %>
+    <h2>Annotations</h2>
+    <table class="annotation-table">
+      <thead>
+        <tr>
+          <th>Description</th>
+          <th>Related Resources</th>
+          <th>CreatedAt</th>
```
jeremyevans, line 63:
```suggestion
          <th>Created At</th>
```

migrate/20251110_add_annotation_model.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:annotation) do
```
enescakir, line 5: I have a high-level question. The Page model already has the necessary columns: the description works like a summary, and we store related resources in the details column, etc. So why don’t we use the Page model with a lower severity level that doesn’t page the on call, instead of introducing a new model? We could also separate the lists on the admin panel as "Urgent Pages" and "Annotations" (non-urgent ones).

```
#  created_at  | timestamp with time zone | NOT NULL DEFAULT CURRENT_TIMESTAMP
#  resolved_at | timestamp with time zone |
#  summary     | text                     |
#  tag         | text                     | NOT NULL
#  details     | jsonb                    | NOT NULL DEFAULT '{}'::jsonb
#  severity    | page_severity            | NOT NULL DEFAULT 'error'::page_severity

```

views/admin/object.erb
```diff
@@ -19,6 +19,12 @@
   <% end %>
 </div>
 
+<div id="annotation-info">
```
enescakir, line 22: Since you add `ObjectAction` for `Annotation`, it should show the action button already

views/admin/index.erb
```diff
@@ -52,4 +52,36 @@
...
+              <% if annotation.related_resources&.any? %>
+                <% annotation.related_resources.each do |resource_id| %>
+                  <% ubid = UBID.from_uuidish(resource_id) %>
+                  <% ubid_class = UBID.class_for_ubid(ubid.to_s).name %>
+                  <a href="/model/<%= ubid_class %>/<%= ubid %>"><%= ubid %></a><br>
+                <% end %>
+              <% else %>
+                <em>No related resources</em>
+              <% end %>
+            </td>
```
jeremyevans, line 80:
```suggestion
            <td><%== linkify_ubids(annotation.related_resources.map { UBID.from_uuidish(it) }.join(" ")) %></td>
```

clover_admin.rb
```diff
@@ -144,9 +144,9 @@ def linkify_ubids(body)
     end
   end
 
-  ObjectAction = Data.define(:label, :flash, :params, :action) do
-    def self.define(label, flash, params = {}, &action)
-      new(label, flash, params.dup.freeze, action)
+  ObjectAction = Data.define(:label, :flash, :redirect, :params, :action) do
```
enescakir, line 147: nit: `redirect` sounds like a boolean. I had to check the method to see how to use it.

```suggestion
  ObjectAction = Data.define(:label, :flash, :redirect_path, :params, :action) do
```

clover_admin.rb
```diff
@@ -394,7 +397,8 @@ def self.object_action(...)
               params = action.params.map { |k, v| typecast_params.send(v, k.to_s) }
               action.call(@obj, *params)
               flash["notice"] = action.flash
-              r.redirect("/model/#{@obj.class}/#{ubid}")
+              redirect_path = action.redirect.nil? ? "/model/#{@obj.class}/#{ubid}" : action.redirect
+              r.redirect(redirect_path)
```
enescakir, line 401: nit:
```suggestion
              redirect_path = action.redirect.nil? ? "/model/#{@obj.class}/#{ubid}" : action.redirect
              r.redirect(action.redirect || "/model/#{@obj.class}/#{ubid}")
```

clover_admin.rb
```diff
@@ -421,6 +425,7 @@ def self.object_action(...)
         .flatten
         .select { it < ResourceMethods::InstanceMethods }
         .sort_by(&:name)
+      @annotations = Annotation.order(Sequel.desc(:created_at)).all
```
enescakir, line 428: nit:

```suggestion
      @annotations = Annotation.reverse(:created_at).all
```

views/admin/object.erb
jeremyevans, line 22: Add this to `views/extra/Annotation.erb`, using the support in #4036, since it is model-specific.

clover_admin.rb
```diff
@@ -144,9 +144,9 @@ def linkify_ubids(body)
     end
   end
 
-  ObjectAction = Data.define(:label, :flash, :params, :action) do
-    def self.define(label, flash, params = {}, &action)
-      new(label, flash, params.dup.freeze, action)
+  ObjectAction = Data.define(:label, :flash, :redirect, :params, :action) do
+    def self.define(label, flash, redirect = nil, params = {}, &action)
```
jeremyevans, line 148: Let's switch this to keyword arguments. Multiple optional arguments where callers have to use explicit `nil` positional argument is a fairly large smell:

```suggestion
    def self.define(label, flash, redirect: nil, params: {}, &action)
```

clover_admin.rb
jeremyevans, line 401:
```suggestion
              r.redirect(action.redirect || "/model/#{@obj.class}/#{ubid}")
```

PR #4150, opened 2025-11-10
Spread Kubernetes Control Plane VMs Among Hosts

aamederen:
We want to avoid CP VMs landing on the same hosts. This change attempts to do that by utilizing exclude_host_ids parameter of VM assembler.

This could lead to a small problem, when replacing the nodes in an upgrade scenario where we don't necessarily want to avoid the new nodes to land on the same host, but that's a minor inconvenience for now.

This is step #1 of improvements on the node distribution and later we can implement some affinity-like levers in allocator to get better spread with AZs or worker nodes as well.

----

> [!IMPORTANT]
> Spread Kubernetes Control Plane VMs across hosts using `exclude_host_ids` and update related configurations and tests.
> 
>   - **Behavior**:
>     - `KubernetesNodeNexus.assemble` now uses `exclude_host_ids` to spread CP VMs across hosts unless `kubernetes_nodepool_id` is set or `Config.allow_unspread_servers` is true.
>     - `ConvergePostgresResource.provision_servers` updated to use `exclude_host_ids` based on `Config.allow_unspread_servers`.
>   - **Config**:
>     - Replaces `POSTGRES_ALLOW_SERVERS_IN_SAME_DATA_CENTER` with `ALLOW_UNSPREAD_SERVERS` in `e2e.yml` and `config.rb`.
>   - **Tests**:
>     - Added tests in `kubernetes_node_nexus_spec.rb` to verify exclusion of hosts with existing CP VMs.
>     - Updated `converge_postgres_resource_spec.rb` to test new server provisioning logic with `exclude_host_ids`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 2c8e2a8ae746d43760191371e423074058ef2a30. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/kubernetes/kubernetes_node_nexus.rb
+++ b/prog/kubernetes/kubernetes_node_nexus.rb
@@ -8,7 +8,7 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
       id = KubernetesNode.generate_uuid
       cluster = KubernetesCluster[kubernetes_cluster_id]
 
-      exclude_host_ids = if kubernetes_nodepool_id
+      exclude_host_ids = if kubernetes_nodepool_id || Config.allow_unspread_servers
         []
       else
         cluster.cp_vms_dataset
--- a/spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
+++ b/spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
@@ -70,8 +70,13 @@ RSpec.describe Prog::Kubernetes::KubernetesNodeNexus do
       expect(kd.vm.vm_host_id).to be_nil
       existing_hosts = [vm.vm_host_id]
 
-      node = described_class.assemble(Config.kubernetes_service_project_id, sshable_unix_user: "ubi", name: "vm3", location_id: Location::HETZNER_FSN1_ID, size: "standard-2", storage_volumes: [{encrypted: true, size_gib: 40}], boot_image: "kubernetes-v1.33", private_subnet_id: subnet.id, enable_ip4: true, kubernetes_cluster_id: kc.id, kubernetes_nodepool_id: nil).subject
+      expect(Config).to receive(:allow_unspread_servers).and_return(false)
+      node = described_class.assemble(Config.kubernetes_service_project_id, sshable_unix_user: "ubi", name: "node3", location_id: Location::HETZNER_FSN1_ID, size: "standard-2", storage_volumes: [{encrypted: true, size_gib: 40}], boot_image: "kubernetes-v1.33", private_subnet_id: subnet.id, enable_ip4: true, kubernetes_cluster_id: kc.id, kubernetes_nodepool_id: nil).subject
       expect(node.vm.strand.stack[0]["exclude_host_ids"]).to eq existing_hosts
+
+      expect(Config).to receive(:allow_unspread_servers).and_return(true)
+      node = described_class.assemble(Config.kubernetes_service_project_id, sshable_unix_user: "ubi", name: "node4", location_id: Location::HETZNER_FSN1_ID, size: "standard-2", storage_volumes: [{encrypted: true, size_gib: 40}], boot_image: "kubernetes-v1.33", private_subnet_id: subnet.id, enable_ip4: true, kubernetes_cluster_id: kc.id, kubernetes_nodepool_id: nil).subject
+      expect(node.vm.strand.stack[0]["exclude_host_ids"]).to eq []
     end
 
     it "doesn't exclude hosts when creating worker nodes" do


prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -6,9 +6,17 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
   def self.assemble(project_id, sshable_unix_user:, name:, location_id:, size:, storage_volumes:, boot_image:, private_subnet_id:, enable_ip4:, kubernetes_cluster_id:, kubernetes_nodepool_id: nil)
     DB.transaction do
       id = KubernetesNode.generate_uuid
+
+      exclude_host_ids = []
+      unless kubernetes_nodepool_id
+        cluster = KubernetesCluster[kubernetes_cluster_id]
+        exclude_host_ids = cluster.nodes.filter_map { |n| n.vm.vm_host_id }.uniq
```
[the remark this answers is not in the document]
jeremyevans: Or better, yet, do the work in the database since it is uncached.

```suggestion
        exclude_host_ids = cluster.cp_vms_dataset
          .exclude(vm_host_id: nil)
          .distinct
          .select_map(:vm_host_id)
```

Additionally, this code is likely to break single-host configurations (typical developer environment or the staging environment). You could use only do this if `Config.postgres_allow_servers_in_same_data_center` is not true. However, since we'll need the same type of logic in multiple places, maybe it would be better to rename that config value and not make it postgres-specific.

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -6,9 +6,20 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
...
+          .exclude(vm_host_id: nil)
+          .distinct
+          .select_map(:vm_host_id)
+      end
+
       vm = Prog::Vm::Nexus.assemble_with_sshable(project_id, sshable_unix_user:, name:, location_id:,
         size:, storage_volumes:, boot_image:, private_subnet_id:, enable_ip4:,
-        allow_private_subnet_in_other_project: true).subject
+        allow_private_subnet_in_other_project: true,
+        exclude_host_ids: exclude_host_ids).subject
```
enescakir: nit:
```suggestion
        exclude_host_ids:).subject
```

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -6,9 +6,20 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
   def self.assemble(project_id, sshable_unix_user:, name:, location_id:, size:, storage_volumes:, boot_image:, private_subnet_id:, enable_ip4:, kubernetes_cluster_id:, kubernetes_nodepool_id: nil)
     DB.transaction do
       id = KubernetesNode.generate_uuid
+
+      exclude_host_ids = []
+      unless kubernetes_nodepool_id
+        cluster = KubernetesCluster[kubernetes_cluster_id]
```
enescakir: We might move it to outer and remove `cluster = node.kubernetes_cluster` reassignment in the future lines

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -6,12 +6,23 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
   def self.assemble(project_id, sshable_unix_user:, name:, location_id:, size:, storage_volumes:, boot_image:, private_subnet_id:, enable_ip4:, kubernetes_cluster_id:, kubernetes_nodepool_id: nil)
     DB.transaction do
       id = KubernetesNode.generate_uuid
+      cluster = KubernetesCluster[kubernetes_cluster_id]
+      exclude_host_ids = []
+
+      unless kubernetes_nodepool_id || Config.allow_unspread_servers
```
byucesoy: We specifically don't do this in `.assemble` for PG servers, because there is a risk (and not a low one because of allocator's preference) of multiple assemble running at the same time, thus causing multiple VMs to be end up in the same host.

Instead our approach is creating PG servers one by one. Also if there is a PG server created but not assigned to a host, we wait it to be assigned to a host before proceeding provisioning the next one.

This is the check ensuring that all servers are assigned to a host;
```
postgres_resource.servers.all? { it.vm.vm_host } 
```

I don't know K8s provisioning enough to check if this is a real concern, but writing here in case it is.

aamederen: Thanks for the input. In K8s, we bootstrap control plane nodes one by one, so this is OK for now. If/when we do parallel bootstrapping (workers for example), this can become an issue.

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -6,12 +6,23 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
...
+        exclude_host_ids = cluster.cp_vms_dataset
```
jeremyevans: Since there is only a single variable being set:

```suggestion
      exclude_host_ids = if kubernetes_nodepool_id || Config.allow_unspread_servers
        []
      else
        cluster.cp_vms_dataset
```

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -6,12 +6,23 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
...
+          .exclude(vm_host_id: nil)
+          .order(nil)
```
jeremyevans:
```suggestion
          .unordered
```

spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
```diff
@@ -59,6 +59,36 @@
       node = described_class.assemble(Config.kubernetes_service_project_id, sshable_unix_user: "ubi", name: "vm2", location_id: Location::HETZNER_FSN1_ID, size: "standard-2", storage_volumes: [{encrypted: true, size_gib: 40}], boot_image: "kubernetes-v1.33", private_subnet_id: subnet.id, enable_ip4: true, kubernetes_cluster_id: kc.id, kubernetes_nodepool_id: kn.id).subject
       expect(node.vm.vm_firewalls).to eq [kc.internal_worker_vm_firewall]
     end
+
+    it "Excludes hosts that already have other CP VMs" do
```
jeremyevans:
```suggestion
    it "excludes hosts that already have other CP VMs" do
```

PR #4157, opened 2025-11-10
Allow create_with_id to take Sequel::Model object instead of id

jeremyevans:
If a non-String is passed, call id on it to get the id.

DRYs up about 40 callers.

Most changes by Claude Code using the following prompt:

```
ResourceMethods::ClassMethods#create_with_id currently accepts a
id argument, which is expected to be a uuid String. Add support so
that if a non-String argument is passed, it call the id method on
the argument to get the id. Afterward, update all `create_with_id`
calls in the project where the first argument uses the `id` method
call on an object to instead pass the object directly, in order to
simplify the code (e.g. convert `create_with_id(object.id, ...)` to
`create_with_id(object, ...)`).
```

I had to tweak create_with_id to support nil values, as we apparently pass nil values. My prompt to Claude didn't mention nil values, so that's more an issue with the prompt than Claude.

I also had to modify brittle specs that mocked create_with_id.

----

> [!IMPORTANT]
> `create_with_id` now accepts `Sequel::Model` objects, simplifying calls and raising errors for `nil` values.
> 
>   - **Behavior**:
>     - `create_with_id` in `resource_methods.rb` now accepts `Sequel::Model` objects, extracting the ID internally.
>     - Raises an error if `nil` is passed to `create_with_id`.
>   - **Code Simplification**:
>     - Updates 40+ instances of `create_with_id` calls across various files to pass objects directly instead of calling `.id`.
>   - **Testing**:
>     - Modifies tests in `strand_spec.rb`, `github_installation_spec.rb`, and `vm_host_spec.rb` to align with the new `create_with_id` behavior.
>     - Adds a test case to ensure `create_with_id` raises an error when `nil` is passed.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 33b4d1c76fb3e3969e16a08b7ae7640634655ce7. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/resource_methods.rb
+++ b/lib/resource_methods.rb
@@ -153,11 +153,10 @@ module ResourceMethods
       generate_ubid.to_uuid
     end
 
-    def create_with_id(id, **)
-      raise "nil id passed to create_with_id" unless id
+    def create_with_id(id_or_model_object, **)
+      raise "nil id passed to create_with_id" unless id_or_model_object
       obj = new(**)
-      id = id.id unless id.is_a?(String)
-      obj.id = id
+      obj.id = id_or_model_object.is_a?(String) ? id_or_model_object : id_or_model_object.id
       obj.save_changes
     end
 


lib/resource_methods.rb
```diff
@@ -154,7 +154,12 @@ def generate_uuid
 
     def create_with_id(id, **)
```
bsatzger: nit: this now accepts either an id value or an object that has an id. it's a bit more nuanced than the naming suggests.

lib/resource_methods.rb
```diff
@@ -155,7 +154,12 @@ def generate_uuid
...
       obj = new(**)
-      obj.id = id
+      obj.id = case id
+      when String, nil
```
enescakir: Do we ever pass `nil` in production? If this only happens in tests, I'd prefer updating the tests instead of allowing `nil`.

lib/resource_methods.rb
```diff
@@ -155,7 +154,12 @@ def generate_uuid
...
+        id
+      else
+        id.id
+      end
```
enescakir: maybe
```suggestion
      obj.id = if id.respond_to?(:id) 
        id.id
      else
        id
      end
```

jeremyevans: I've been burned with this type of approach before. Maybe not so much an issue with `id`, but definitely with `call`.  I can make it Sequel::Model specific though (`id = id.id if id.is_a?(Sequel::Model)`).

lib/resource_methods.rb
jeremyevans: Not sure if we every pass nil in production. But it definitely happens in tests. I can track down the test issues and fix them.

lib/resource_methods.rb
jeremyevans: The `id = id.id if id.is_a?(Sequel::Model)` approach unfortunately breaks if you are using `instance_double`, and we are. So I decided against that.

lib/resource_methods.rb
jeremyevans: Found and fixed the 3 places where `nil` was being passed unintentionally, and changed the method to raise if `nil` id is passed.

lib/resource_methods.rb
jeremyevans: Pushed a commit to change the argument name.

PR #4162, opened 2025-11-11
Add support for updating firewall descriptions

jeremyevans:
Previously, we didn't offer a way to do this. So if you wanted to change the description after creating the firewall, you were out of luck.

Mostly by Claude Code, using the following prompt:

```
Ubicloud currently supports updating the name of the firewall but
not updating the description of a firewall. Add support for updating
a description of the firewall, similar to the current support for
updating the name for a firewall. Steps:

* Add an appropriate endpoint to routes/project/location/firewall.rb to change the firewall description
  * To allow API requests to the route to work, update the OpenAPI configuration stored in openapi/openapi.yml
* Update views/networking/firewall/settings.erb template to add a form that will submit to the new endpoint.
  * Add appropriate capybara-based specs to test this change to spec/routes/web/firewall_spec.rb
* Update sdk/ruby/lib/ubicloud/model/firewall.rb to add support for calling that route to the SDK
* Add cli-commands/fw/post/update-description.rb to add support to the CLI, using the SDK support
  * Add appropriate specs to test this change to a new file named spec/routes/api/cli/fw/update-description_spec.rb, similar to the existing specs in spec/routes/api/cli/fw/rename_spec.rb
```

Issue fixed manually:

* Fixed the CLI command by including `UbiCli.` before `on`
* Change r`request.redirect` to `r.redirect` in the route
* Add test for case where description is unchanged
* Use different flahs notice when description is unchanged
* Remove unused variable assignment in spec
* Remove redundant self in SDK method.
* Change CLI command description
* Update golden files
* 
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-north-h1_firewall_t-default_settings

----

> [!IMPORTANT]
> Add support for updating firewall descriptions via API, CLI, and web interface, with corresponding tests and documentation updates.
> 
>   - **Behavior**:
>     - Adds endpoint in `firewall.rb` to update firewall descriptions.
>     - Updates OpenAPI spec in `openapi.yml` to include new PATCH operation for firewall descriptions.
>     - Updates `settings.erb` to include a form for updating descriptions.
>   - **CLI**:
>     - Adds `update-description.rb` command to CLI for updating firewall descriptions.
>     - Updates CLI help files to include `update-description` command.
>   - **SDK**:
>     - Adds `update_description` method to `firewall.rb` in SDK.
>   - **Tests**:
>     - Adds tests for CLI command in `update-description_spec.rb`.
>     - Adds web tests for description update in `firewall_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for fafbff76ed8b2921da09f931d42c1a50aa261b35. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli-commands/fw/post/detach-subnet.rb
+++ b/cli-commands/fw/post/detach-subnet.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 UbiCli.on("fw").run_on("detach-subnet") do
-  desc "Detch a private subnet from a firewall"
+  desc "Detach a private subnet from a firewall"
 
   banner "ubi fw (location/fw-name | fw-id) detach-subnet (ps-name | ps-id)"
 
--- a/spec/routes/api/cli/golden-files/help -r.txt
+++ b/spec/routes/api/cli/golden-files/help -r.txt
@@ -90,7 +90,7 @@ Post Commands:
     create                Create a firewall
     delete-rule           Remove a firewall rule
     destroy               Destroy a firewall
-    detach-subnet         Detch a private subnet from a firewall
+    detach-subnet         Detach a private subnet from a firewall
     modify-rule           Modify a firewall rule
     rename                Rename a firewall
     show                  Show details for a firewall
@@ -152,7 +152,7 @@ Options:
     -f, --force                      do not require confirmation
 
 
-Detch a private subnet from a firewall
+Detach a private subnet from a firewall
 
 Usage:
     ubi fw (location/fw-name | fw-id) detach-subnet (ps-name | ps-id)


openapi/openapi.yml
```diff
@@ -494,6 +494,34 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Firewall
+  '/project/{project_id}/location/{location}/firewall/{firewall_reference}/update-description':
```
enescakir: Do we plan to have separate endpoints for each property instead of using PATCH on the resource endpoint to update its properties? It feels a bit like an anti-pattern and not very restful. We're starting to diverge from our general API design. As far as I know, PostgreSQL resources already have their own PATCH endpoint.

jeremyevans: I'm fine changing this into a general patch endpoint in the API. Would we allow changing anything else in such an endpoint?

jeremyevans: I took another look at this. If there were any other columns for firewall that we would change in the same endpoint, I would agree with you. However, looking at firewall's columns, I don't think we would allow changing any of them. We do allow changing name, but renaming has been a separate action, and has effects beyond changing the column value (such as changing what paths are valid for the object).

Since the only column you could reasonably change in this case is the description, I think it's preferable to have a more specific endpoint name. Unless you foresee a change to firewall to add user-modifiable columns, this seems like a better approach to me.

Speaking in terms of general API design (and not this specific situation), I don't think our API follows restful design in many cases, and that's fine. IMO, restful design is not necessarily better than non-restful design (not necessarily worse, it depends on the situation). Blindly following restful design leads to problems, IMO. One example of this is the load balancer patch route. It is certainly a restful design, and using restful design there I think makes the code significantly more complex than it needs to be. The same is true for the postgres patch route, IMO. I think smaller, more focused endpoints are in general easier to maintain and understand.

enescakir: I'm not sure whether we'll add more updateable columns to the firewall resource in the future. However, since PG/LB already use a PATCH route for updates, introducing a new, dedicated single-property endpoint for firewall updates would create an inconsistent API surface and an additional learning curve for users.

Users would have to remember which resources support PATCH on the main resource path and which ones require specialized endpoints. I don't have a strong preference, but I lean toward keeping the API consistent overall so users don't need to learn new patterns when switching between resources.

jeremyevans: @geemus @fdr Could we get your feedback regarding this?

geemus: I would agree that RESTfulness doesn't need to be followed for it's own sake, and also that we should try to to be consistent. It looks like there are at least a couple of precedents in the API for patch. My preference would be toward sticking to that pattern (in the hopes of fewer total patterns and therefore easier learning curve). It may be that this particular one never updates anything but name, but even then I don't think it's significantly worse.

jeremyevans: @geemus Thanks for providing feedback.

@enescakir I've updated the API to use PATCH.

routes/project/location/firewall.rb
```diff
@@ -22,6 +22,30 @@ class Clover
       @firewall = firewall = @project.firewalls_dataset.first(filter)
       check_found_object(firewall)
 
+      r.api_patch_web_post do
+        authorize("Firewall:edit", firewall)
+        handle_validation_failure("networking/firewall/show") { @page = "settings" }
+        description = typecast_body_params.nonempty_str!("description")
+
+        if description == firewall.description
```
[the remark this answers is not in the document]
jeremyevans: I don't think it's worth it. Even if there is a mismatch solely due to whitespace, it uses the requested amount of whitespace, which is arguably more correct.

furkansahin (approved):

jeremyevans:
@enescakir Are you now OK with the approach?

jeremyevans:
Definitely, thanks for catching that.

PR #4166, opened 2025-11-11
Restrict VictoriaMetrics to serve selected endpoints

shikharbhardwaj:
This change updates the vmauth config to serve specific endpoints we need for the client and grafana, to minimize exposure.
With this change, all other endpoints return `remoteAddr: "xx.xx.xx.xx:54042"; requestURI: /debug/pprof/heap?debug=1; missing route for "/debug/pprof/heap?debug=1"`.

----

> [!IMPORTANT]
> Restricts VictoriaMetrics to serve only specific endpoints by updating `url_map` in `configure`, returning errors for all other endpoints.
> 
>   - **Behavior**:
>     - Restricts VictoriaMetrics to serve only specific endpoints by updating `url_map` in `configure`.
>     - Allowed endpoints: `/health`, `/api/v1/import/prometheus`, `/api/v1/query`, `/api/v1/query_range`, `/api/v1/series`, `/api/v1/labels`, `/api/v1/label/.*/values`.
>     - All other endpoints return a missing route error with client IP and request URI.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0564f44abf50523563333516d0ca2e81933b1097. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/victoria_metrics/bin/configure
+++ b/rhizome/victoria_metrics/bin/configure
@@ -52,7 +52,16 @@ auth_config = <<-AUTH_CONFIG
 users:
   - username: "#{user}"
     password: "#{password}"
-    url_prefix: "http://127.0.0.1:8428"
+    url_map:
+      - src_paths:
+          - /health
+          - /api/v1/import/prometheus
+          - /api/v1/query
+          - /api/v1/query_range
+          - /api/v1/series
+          - /api/v1/labels
+          - /api/v1/label/.*/values
+        url_prefix: "http://127.0.0.1:8428"
 AUTH_CONFIG
 
 safe_write_to_file("/etc/victoria_metrics/auth-config.yml", auth_config)


rhizome/victoria_metrics/bin/configure
```diff
@@ -52,7 +52,16 @@ auth_config = <<-AUTH_CONFIG
...
-    url_prefix: "http://127.0.0.1:8428"
+    url_map:
+    - src_paths:
+      - /health
+      - /api/v1/import/prometheus
+      - /api/v1/query
+      - /api/v1/query_range
+      - /api/v1/series
+      - /api/v1/labels
+      - /api/v1/label/.*/values
```
jeremyevans: This syntax is correct, but looks weird. Indenting further makes it more clear that `url_prefix` and `src_paths` are the same level, and these entries are nested under `src_paths`:

```suggestion
        - /health
        - /api/v1/import/prometheus
        - /api/v1/query
        - /api/v1/query_range
        - /api/v1/series
        - /api/v1/labels
        - /api/v1/label/.*/values
```

This further indenting matches the formatting I saw in examples when I searched for this, but I'll leave this to your discretion.

jeremyevans (commented):
Looks good to me with a formatting nit. I'll leave approval for @byucesoy, who I'm guessing is more knowledgeable in this area.

rhizome/victoria_metrics/bin/configure
shikharbhardwaj: I was using this as a reference: https://docs.victoriametrics.com/victoriametrics/vmauth/#generic-http-proxy-for-different-backends. The indentation seems identical to me in both.

shikharbhardwaj: I looked up general YAML formatting examples and they seem to follow the extra indentation for child lists. I updated the file to be consistent.

PR #4174, opened 2025-11-12
Allow IAM auth when clover db and postgres monitor db are RDS

serprex:
split out from #4141

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -80,6 +80,7 @@ module Config
   override :dispatcher_queue_size_ratio, 4, float
   override :recursive_tag_limit, 32, int
   override :root, File.expand_path(__dir__), string
+  override :clover_database_rds_iam_auth_enabled, false, bool
   optional :hetzner_user, string, clear: true
   optional :hetzner_password, string, clear: true
   override :hetzner_connection_string, "https://robot-ws.your-server.de", string
--- a/db.rb
+++ b/db.rb
@@ -5,18 +5,24 @@ require "sequel/core"
 require_relative "config"
 require_relative "lib/util"
 require_relative "lib/clog"
+if Config.clover_database_rds_iam_auth_enabled
+  require "pg/aws_rds_iam"
+  driver_options = {aws_rds_iam_auth_token_generator: "default"}
+else
+  driver_options = {}
+end
 
 db_ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", "db_ca_bundle.crt")
 Util.safe_write_to_file(db_ca_bundle_filename, Config.clover_database_root_certs)
 max_connections = (Util.monitor_process? ? Config.db_pool_monitor : Config.db_pool) - 1
 max_connections = 1 if ENV["SHARED_CONNECTION"] == "1"
 pg_auto_parameterize_min_array_size = 1 if Config.test? && ENV["CLOVER_FREEZE"] == "1"
-DB = Sequel.connect(Config.clover_database_url, max_connections:, pool_timeout: Config.database_timeout, treat_string_list_as_untyped_array: true, pg_auto_parameterize_min_array_size:)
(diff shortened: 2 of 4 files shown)


config.rb
```diff
@@ -143,6 +144,7 @@ def self.test?
   override :postgres_service_hostname, "postgres.ubicloud.com", string
   override :postgres_monitor_database_url, Config.clover_database_url, string
   optional :postgres_monitor_database_root_certs, string
+  override :postgres_monitor_database_rds_iam_auth_enabled, false, bool
```
jeremyevans: It's technically possible for the monitor database to be different and require different auth from the main database, but is that a situation you actually need to support? If not, probably simpler to only have a single option, and let it control both.

db.rb
```diff
@@ -5,20 +5,25 @@
...
 
 postgres_monitor_db_ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", "postgres_monitor_db.crt")
 Util.safe_write_to_file(postgres_monitor_db_ca_bundle_filename, Config.postgres_monitor_database_root_certs)
-begin
-  POSTGRES_MONITOR_DB = Sequel.connect(Config.postgres_monitor_database_url, max_connections: Config.db_pool_monitor, pool_timeout: Config.database_timeout) if Config.postgres_monitor_database_url
-rescue Sequel::DatabaseConnectionError => ex
-  Clog.emit("Failed to connect to Postgres Monitor database") { {database_connection_failed: {exception: Util.exception_to_hash(ex)}} }
+if Config.postgres_monitor_database_url
+  begin
+    driver_options = Config.postgres_monitor_database_rds_iam_auth_enabled ? {aws_rds_iam_auth_token_generator: "default"} : {}
```
jeremyevans: Assuming you go with a single Config option:

```suggestion
```

config.rb
serprex: We don't need the split. Will consolidate

db.rb
```diff
@@ -5,18 +5,20 @@
 require_relative "config"
 require_relative "lib/util"
 require_relative "lib/clog"
+require "pg/aws_rds_iam" if Config.clover_database_rds_iam_auth_enabled
```
jeremyevans:
```suggestion
if Config.clover_database_rds_iam_auth_enabled
  require "pg/aws_rds_iam"
  driver_options = {aws_rds_iam_auth_token_generator: "default"}
else
  driver_options = {}
end
```

db.rb
```diff
@@ -5,18 +5,20 @@
...
 
 db_ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", "db_ca_bundle.crt")
 Util.safe_write_to_file(db_ca_bundle_filename, Config.clover_database_root_certs)
 max_connections = (Util.monitor_process? ? Config.db_pool_monitor : Config.db_pool) - 1
 max_connections = 1 if ENV["SHARED_CONNECTION"] == "1"
 pg_auto_parameterize_min_array_size = 1 if Config.test? && ENV["CLOVER_FREEZE"] == "1"
-DB = Sequel.connect(Config.clover_database_url, max_connections:, pool_timeout: Config.database_timeout, treat_string_list_as_untyped_array: true, pg_auto_parameterize_min_array_size:)
+driver_options = Config.clover_database_rds_iam_auth_enabled ? {aws_rds_iam_auth_token_generator: "default"} : {}
```
jeremyevans:
```suggestion
```

furkansahin:
@serprex can you rewrite the commit message so that the textwidth is 72? As a reference, my `.vimrc` is like the following:
```
autocmd Filetype gitcommit setlocal spell textwidth=72
set number
set ruler
set textwidth=72
set wrap
```

serprex:
Figured I was at the lowest common denominator fitting a terminal on half the screen of a 13 inch screen, but updated accordingly

PR #4177, opened 2025-11-13
NicNexus AWS/Ubicloud separation - Phase 1

furkansahin:
## Nic allow mac to be nil -migration-
We don't need that field for AWS nics, we might drop the column all
together later specifically for that type of nics.

## Add Ubicloud/vnet/nic_nexus.rb and Aws/vnet/nic_nexus.rb
We are adding the new progs as a preparation. These are not used just
yet. We will switch to use them in the next commit.

## Prog migration logic is added to the NicNexus.wait
With the help of this semaphore, we will be able to slowly move nics to
their new progs in the production, manually.

----

> [!IMPORTANT]
> Separate NIC handling logic for AWS and Ubicloud, allowing `mac` to be `nil` for AWS NICs, with corresponding tests.
> 
>   - **Behavior**:
>     - Allow `mac` to be `nil` for AWS NICs in `migrate/20251112_nic_entity_mac_nil_allow.rb`.
>     - Introduce separate NIC handling logic for AWS and Ubicloud in `prog/vnet/aws/nic_nexus.rb` and `prog/vnet/metal/nic_nexus.rb`.
>     - Add migration logic in `Prog::Vnet::NicNexus.wait` to transition NICs to new programs.
>   - **Models**:
>     - Update `Nic` model in `model/nic.rb` to allow `mac` to be `nil`.
>   - **Programs**:
>     - Add `Prog::Vnet::Aws::NicNexus` and `Prog::Vnet::Metal::NicNexus` for specific NIC handling.
>     - Modify `Prog::Vnet::NicNexus` to delegate to AWS or Metal specific logic.
>   - **Tests**:
>     - Add tests for AWS NIC logic in `spec/prog/vnet/aws/nic_nexus_spec.rb`.
>     - Add tests for Metal NIC logic in `spec/prog/vnet/metal/nic_nexus_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7c85ded2f9daa36e681a0f22e28781139b3dfd51. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/nic.rb
+++ b/model/nic.rb
@@ -13,7 +13,8 @@ class Nic < Sequel::Model
 
   plugin ResourceMethods, encrypted_columns: :encryption_key
   plugin SemaphoreMethods, :destroy, :start_rekey, :trigger_outbound_update,
-    :old_state_drop_trigger, :setup_nic, :repopulate, :lock, :vm_allocated
+    :old_state_drop_trigger, :setup_nic, :repopulate, :lock, :vm_allocated,
+    :migrate_to_separate_prog
 
   def self.ubid_to_name(ubid)
     ubid.to_s[0..7]
--- a/prog/vnet/nic_nexus.rb
+++ b/prog/vnet/nic_nexus.rb
@@ -65,6 +65,18 @@ class Prog::Vnet::NicNexus < Prog::Base
   end
 
   label def wait
+    # :nocov:
+    when_migrate_to_separate_prog_set? do
+      if nic.private_subnet.location.aws?
+        strand.update(prog: "Vnet::Aws::NicNexus", label: "wait")
+      else
+        strand.update(prog: "Vnet::Metal::NicNexus", label: "wait")
+      end
+      decr_migrate_to_separate_prog
+      nap 0
+    end
+    # :nocov:
+
     if nic.private_subnet.location.aws?
       nic.semaphores.each(&:destroy)
       nap 60 * 60 * 24 * 365


prog/aws/vnet/nic_nexus.rb
```diff
@@ -0,0 +1,185 @@
...
+  label def start
+    NicAwsResource.create_with_id(nic.id)
+    hop_create_subnet
+  end
+
+  label def create_subnet
+    nap 2 unless nic.private_subnet.strand.label == "wait"
+
+    register_deadline("attach_eip_network_interface", 3 * 60)
+    vpc_response = client.describe_vpcs({filters: [{name: "vpc-id", values: [private_subnet.private_subnet_aws_resource.vpc_id]}]}).vpcs[0]
```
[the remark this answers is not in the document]
jeremyevans: Repeating the call doesn't cause an N+1 query issue, as the association is cached. However, I think you should add a `vpc_id` local variable to DRY things up.

prog/aws/vnet/nic_nexus.rb
```diff
@@ -0,0 +1,185 @@
...
+    end
+  end
+
+  label def start
+    NicAwsResource.create_with_id(nic.id)
+    hop_create_subnet
+  end
+
+  label def create_subnet
+    nap 2 unless nic.private_subnet.strand.label == "wait"
```
jeremyevans:
```suggestion
    nap 2 unless private_subnet.strand.label == "wait"
```

prog/vnet/aws/nic_nexus.rb
```diff
@@ -0,0 +1,185 @@
...
+    NicAwsResource.create_with_id(nic.id)
+    hop_create_subnet
+  end
+
+  label def create_subnet
+    nap 2 unless nic.private_subnet.strand.label == "wait"
+
+    register_deadline("attach_eip_network_interface", 3 * 60)
+    vpc_response = client.describe_vpcs({filters: [{name: "vpc-id", values: [private_subnet.private_subnet_aws_resource.vpc_id]}]}).vpcs[0]
+    ipv_6_cidr_block = NetAddr::IPv6Net.parse(vpc_response.ipv_6_cidr_block_association_set[0].ipv_6_cidr_block).nth_subnet(64, SecureRandom.random_number(2**8))
```
jeremyevans, line 34: Let me preface this comment by saying there may be a good reason for the use of a small random number that I am not aware of. If so, please add a comment explaining why this approach was taken.

That being said, if we are using random numbers here to attempt to avoid conflicts, 2**8 seems much too low. We either need a much larger number (and smaller than /64 subnet mask), or we should use sequential assignment and keep track of what has already been assigned.

If conflicts are not an issue, I would use `0` instead of a random number.

prog/aws/vnet/nic_nexus.rb
```diff
@@ -0,0 +1,185 @@
...
+    nap 1
+  end
+
+  label def create_network_interface
+    network_interface_response = client.create_network_interface({
+      subnet_id: nic.nic_aws_resource.subnet_id,
+      private_ip_address: nic.private_ipv4.network.to_s,
+      ipv_6_prefix_count: 1,
+      groups: [
+        nic.private_subnet.private_subnet_aws_resource.security_group_id
```
jeremyevans:
```suggestion
        private_subnet.private_subnet_aws_resource.security_group_id
```

prog/aws/vnet/nic_nexus.rb
```diff
@@ -0,0 +1,185 @@
...
+      eip_response.addresses[0].allocation_id
+    end
+
+    nic.nic_aws_resource.update(eip_allocation_id:)
+    hop_attach_eip_network_interface
+  end
+
+  label def attach_eip_network_interface
+    eip_response = client.describe_addresses({filters: [{name: "allocation-id", values: [nic.nic_aws_resource.eip_allocation_id]}]})
+    if eip_response.addresses.first.network_interface_id.nil?
```
jeremyevans: Up to you, but unless you care about the false/nil difference, this is more idiomatic:

```suggestion
    unless eip_response.addresses.first.network_interface_id
```

prog/aws/vnet/nic_nexus.rb
```diff
@@ -0,0 +1,185 @@
...
+    ignore_invalid_nic do
+      client.delete_network_interface({network_interface_id: nic.nic_aws_resource.network_interface_id})
+    end
+    hop_release_eip
+  end
+
+  label def release_eip
+    ignore_invalid_nic do
+      allocation_id = nic.nic_aws_resource&.eip_allocation_id
+      client.release_address({allocation_id: allocation_id}) if allocation_id
```
jeremyevans:
```suggestion
      client.release_address({allocation_id:}) if allocation_id
```

prog/aws/vnet/nic_nexus.rb
```diff
@@ -0,0 +1,185 @@
...
+  end
+
+  label def destroy_entities
+    nic&.nic_aws_resource&.destroy
+    nic&.destroy
+    pop "nic deleted"
+  end
+
+  def client
+    @client ||= nic.private_subnet.location.location_credential.client
```
jeremyevans:
```suggestion
    @client ||= private_subnet.location.location_credential.client
```

spec/prog/ubicloud/vnet/nic_nexus_spec.rb
```diff
@@ -0,0 +1,194 @@
+# frozen_string_literal: true
+
+RSpec.describe Prog::Ubicloud::Vnet::NicNexus do
```
jeremyevans: Up to you, but I would rename this to `Prog::VmHost::Vnet::NicNexus`. Both the VmHost nics and AWS nics should be thought of as Ubicloud, IMO.

byucesoy: My read is that we are grouping based on who provides the underlying infrastructure Ubicloud or AWS. From that perspective I think Ubicloud is fine. I don't feel very strong on this though, so both are OK IMO.

However, I think we should change the order, i.e. `Prog::Ubicloud::Vnet::NicNexus` -> `Prog::Vnet::Ubicloud::NicNexus`. My reasoning is that currently prog folder is split based on services. We have postgres, vm, vnet, k8s, ai, dns_zone, victoria_metrics etc. Putting ubicloud and aws folders there breaks the pattern. Also, apart from the foundational services (i.e. vm and vnet), services are agnostic to the underlying provider, so putting provider specific foldering under vm and vnet seems more straightforward to me.

prog/vnet/aws/nic_nexus.rb
furkansahin: When we create a VPC and associate an ipv6 cidr block, AWS uses /56 prefix length by default. At the time of the implementation, we did not have a reason to use another number considering it is big enough. Ubicloud private subnet size for IPv6 is /64, which means, for the sake of consistency, we went for a mask of /64 here as well. This meant that we had 2**8 subnets to pick from. Considering Ubicloud Postgres on AWS provisions up to 4 subnets (1 per VM) at a given time, the possibility of conflict is seen low enough. On top of it, the call fails when there is a conflicting subnet and we can simply run the same state again to reproduce a new subnet range. I don't think it worths to keep track of the used subnets and sequentially assign in this case. I'll add this as a comment to the file. 

Also, if this is still seen as a problem, I can change the VPC ipv6 cidr block size to /44 (lowest allowed). However, I'd prefer to do it in a different PR.

spec/prog/ubicloud/vnet/nic_nexus_spec.rb
furkansahin: I agree with both views. On one hand, Ubicloud provides both but on the other hand, it's not the VmHost entity that provides the nics. We should also not put nic under VmHost because they are not really related. So, I decided to use "Metal" the new paths are `Prog::Vnet::Aws::NicNexus` and `Prog::Vnet::Metal::NicNexus`.

jeremyevans: `Metal` seems like a fine name. It certainly makes sense with the current setup. If we ever have a situation where we can run a VmHost on a non-metal situation (nested VMs), it wouldn't be accurate, but that possibility seems remote. In any case, `Metal` is better than `Ubicloud`.

prog/vnet/aws/nic_nexus.rb
jeremyevans: For 4 subnets, you have around a 1 in 43 ((1+2+3)/256) chance of them conflicting using this random approach. If a duplicate subnet will raise an exception, and therefore the label will be retried, I guess it is fine if you are sure the label is idempotent.

jeremyevans (commented):
All my comments have been addressed. I'll leave approval to @byucesoy.

PR #4179, opened 2025-11-13
Split subnet logic to provider specific progs - Part 1

byucesoy:
**Create provider specific subnet nexus progs**
This is just a dummy commit to add two new progs. These two progs and their
tests are exact copies of Prog::Aws::Vpc and Prog::Vnet::SubnetNexus. They are
not being used at the moment. I'll add actual logic changes in the next commit.
By adding these two files now, I can avoid a huge diff in the next commit, thus
making the review significantly easier.

**Split subnet nexus logic to provider specific progs**
Previously, we would create Vnet::SubnetNexus prog for all subnet related
operations regardless of the provider. If the underlying provider was AWS,
SubnetNexus prog would internally bud Aws::Vpc prog to handle VPC related
operations. Now we are moving to a direction where each provider would have
a standalone prog to handle its VPC/Subnet related operations. This would
allow us to better isolate provider specific logic.

----

> [!IMPORTANT]
> Introduce provider-specific subnet handling classes for AWS and Metal, with corresponding tests.
> 
>   - **Behavior**:
>     - Introduces `Prog::Vnet::Aws::VpcNexus` and `Prog::Vnet::Metal::SubnetNexus` classes for provider-specific subnet handling.
>     - Updates `Prog::Vnet::SubnetNexus` to use provider-specific classes based on location.
>   - **Tests**:
>     - Adds `vpc_nexus_spec.rb` for `Prog::Vnet::Aws::VpcNexus`.
>     - Adds `subnet_nexus_spec.rb` for `Prog::Vnet::Metal::SubnetNexus`.
>   - **Misc**:
>     - Updates `thawed_mock.rb` to allow mocking of `PrivateSubnetAwsResource`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9533dc15da1499b97fac5ab8b3d47541de717881. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/private_subnet.rb
+++ b/model/private_subnet.rb
@@ -63,7 +63,7 @@ class PrivateSubnet < Sequel::Model
     (state == "waiting") ? "available" : state
   end
 
-  plugin SemaphoreMethods, :destroy, :destroying, :refresh_keys, :add_new_nic, :update_firewall_rules
+  plugin SemaphoreMethods, :destroy, :destroying, :refresh_keys, :add_new_nic, :update_firewall_rules, :migrate
 
   def self.random_subnet
     subnet_dict = PRIVATE_SUBNET_RANGES.each_with_object({}) do |subnet, hash|
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -81,6 +81,18 @@ class Prog::Vnet::SubnetNexus < Prog::Base
   end
 
   label def wait
+    # :nocov:
+    when_migrate_set? do
+      if private_subnet.location.aws?
+        strand.update(prog: "Vnet::Aws::SubnetNexus", label: "wait")
+      else
+        strand.update(prog: "Vnet::Metal::SubnetNexus", label: "wait")
+      end
+      decr_migrate
+      nap 0
+    end
+    # :nocov:
+
     if private_subnet.location.aws?
       check_firewall_update
       private_subnet.semaphores.each(&:destroy)


jeremyevans:
Some commits seem to overlap with the already merged #4173. Can you rebase? The test failure should be solvable by updating thawed mock:

```ruby
allow_mocking(PrivateSubnetAwsResource, :create_with_id)
```

byucesoy:
Rebased and fixed the tests.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
+# frozen_string_literal: true
+
+require "aws-sdk-ec2"
+class Prog::Vnet::Aws::VpcNexus < Prog::Base
+  subject_is :private_subnet
+
+  def before_run
+    when_destroy_set? do
+      if strand.label != "destroy"
```
jeremyevans: This pattern is used elsewhere and is fundamentally risky. It also appears to be wrong in this case, since it doesn't include `delete_internet_gateway` and `delete_vpc`. If there is a reason that we should still hop to `destroy` even if we are already at one of those labels, please add a comment explaining why.

I can think of a couple potential alternatives to this pattern:

* Add a `destroying` semaphore, set it during `destroy`, and not hop if that semaphore is set.
* Update the frame to set a `destroying` entry, and not hop if that entry is set.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    private_subnet.private_subnet_aws_resource.update(vpc_id: vpc_id)
+    hop_wait_vpc_created
+  end
+
+  label def wait_vpc_created
+    vpc = client.describe_vpcs({filters: [{name: "vpc-id", values: [private_subnet.private_subnet_aws_resource.vpc_id]}]}).vpcs[0]
+
+    if vpc.state == "available"
```
jeremyevans: Up to you, but

```suggestion
    nap 1 unless vpc.state == "available"
```

and removing the extra indentation from the rest of the method may be simpler.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
...
+          vpc_id: private_subnet.private_subnet_aws_resource.vpc_id,
+          tag_specifications: Util.aws_tag_specifications("security-group", private_subnet.name)
+        })
+      rescue Aws::EC2::Errors::InvalidGroupDuplicate
+        client.describe_security_groups({filters: [{name: "group-name", values: ["aws-#{location.name}-#{private_subnet.ubid}"]}]}).security_groups[0]
+      end
+
+      private_subnet.private_subnet_aws_resource.update(security_group_id: security_group_response.group_id)
+
+      private_subnet.firewalls.flat_map(&:firewall_rules).each do |firewall_rule|
```
jeremyevans: Avoid N+1 query issue:

```suggestion
      private_subnet.firewalls(eager: :firewall_rules).flat_map(&:firewall_rules).each do |firewall_rule|
```

Alternatively, add a `PrivateSubnet.many_through_many :firewall_rules` association and use that.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
...
+
+    vpc_id = if vpc_response.vpcs.empty?
+      client.create_vpc({cidr_block: private_subnet.net4.to_s,
+        amazon_provided_ipv_6_cidr_block: true,
+        tag_specifications: Util.aws_tag_specifications("vpc", private_subnet.name)}).vpc.vpc_id
+    else
+      vpc_response.vpcs.first.vpc_id
+    end
+
+    private_subnet.private_subnet_aws_resource.update(vpc_id: vpc_id)
```
jeremyevans: Considering adding a `private_subnet_aws_resource` method to DRY up the repeated `private_subnet.private_subnet_aws_resource`.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
...
+    hop_wait
+  end
+
+  label def wait
+    when_update_firewall_rules_set? do
+      private_subnet.vms.each(&:incr_update_firewall_rules)
+      decr_update_firewall_rules
+    end
+
+    private_subnet.semaphores.each(&:destroy)
```
jeremyevans: Looking at the available semaphores for PrivateSubnet:

* `destroy` would be already handled by the `before_run`
* `update_firewall_rules` handled by block earlier in this method
* `refresh_keys`, `add_new_nic` are not handled.

I'm guessing they do not apply to AWS VPCs and should be ignored? However, this could be a footgun if we add a semaphore to PrivateSubnet later that could apply to AWS VPCs. I think it would probably be better to only delete specific semaphores, and add a comment documenting why.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
...
+      register_deadline(nil, 10 * 60, allow_extension: true) if private_subnet.nics.any? { |n| n.vm&.prevent_destroy_set? }
+
+      Clog.emit("Cannot destroy subnet with active nics, first clean up the attached resources") { private_subnet }
+
+      nap 5
+    end
+
+    decr_destroy
+    private_subnet.nics.each(&:incr_destroy)
+    private_subnet.firewalls.each(&:destroy)
```
jeremyevans: This looks wrong (maybe it isn't, but it certainly looks like it is). The firewalls could be used by other private subnets. If you want to remove the entries from the join table:


```suggestion
    private_subnet.remove_all_firewalls
```

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
...
+      client.delete_security_group({group_id: private_subnet.private_subnet_aws_resource.security_group_id})
+    end
+    hop_delete_internet_gateway
+  end
+
+  label def delete_internet_gateway
+    ignore_invalid_id do
+      client.detach_internet_gateway({internet_gateway_id: private_subnet.private_subnet_aws_resource.internet_gateway_id, vpc_id: private_subnet.private_subnet_aws_resource.vpc_id})
+      client.delete_internet_gateway({internet_gateway_id: private_subnet.private_subnet_aws_resource.internet_gateway_id})
+    end
```
jeremyevans, line 147: If there is any possibility that the `delete_internet_gateway` could work even if the `detach_internet_gateway` fails, this should be split.

```suggestion
    ignore_invalid_id do
      client.detach_internet_gateway({internet_gateway_id: private_subnet.private_subnet_aws_resource.internet_gateway_id, vpc_id: private_subnet.private_subnet_aws_resource.vpc_id})
    end
    ignore_invalid_id do
      client.delete_internet_gateway({internet_gateway_id: private_subnet.private_subnet_aws_resource.internet_gateway_id})
    end
```

I would guess that it is possible if the `vpc_id` entry is the cause of the invalid id instead of the `internet_gateway_id`, but that's not an educated guess.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,193 @@
...
+    Aws::EC2::Errors::InvalidSubnetIDNotFound,
+    Aws::EC2::Errors::InvalidGroupNotFound,
+    Aws::EC2::Errors::InvalidNetworkInterfaceIDNotFound,
+    Aws::EC2::Errors::InvalidInternetGatewayIDNotFound,
+    Aws::EC2::Errors::InvalidVpcIDNotFound => e
+    Clog.emit("ID not found for aws vpc") { {ignored_aws_vpc_failure: {exception: Util.exception_to_hash(e, backtrace: nil)}} }
+  end
+
+  def location
+    private_subnet.location
```
jeremyevans: Up to you, but you may want to cache this, since the location shouldn't change:

```suggestion
    @location ||= private_subnet.location
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+  def before_run
+    when_destroy_set? do
+      if strand.label != "destroy"
+        register_deadline(nil, 10 * 60)
+        hop_destroy
+      end
+    end
+  end
+
+  label def start
```
jeremyevans, line 15: Personally, I don't like labels that only hop to another label. However, I see that would make the code to dispatch to either this or or the aws prog more complex, so it's probably best to keep this as is.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+  def gen_encryption_key
+    "0x" + SecureRandom.bytes(36).unpack1("H*")
+  end
+
+  def gen_spi
+    "0x" + SecureRandom.bytes(4).unpack1("H*")
+  end
+
+  def gen_reqid
+    SecureRandom.random_number(100000) + 1
```
jeremyevans: If this should only return a non-zero 1-5 digit number, this has a problem if the random number is `99999`. It may be better to do:

```suggestion
    SecureRandom.random_number(1...100000)
```

If this should be allowed to return a non-zero 1-6 digit number, then you should use the above and add a zero.

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  def gen_reqid
+    SecureRandom.random_number(100000) + 1
+  end
+
+  label def add_new_nic
+    register_deadline("wait", 3 * 60)
+    nics_snap = nics_to_rekey
+    nap 10 if nics_snap.any? { |nic| nic.lock_set? }
```
jeremyevans:
```suggestion
    nap 10 if nics_snap.any?(&:lock_set?)
```

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    decr_add_new_nic
+    hop_wait_inbound_setup
+  end
+
+  label def refresh_keys
+    decr_refresh_keys
+    nics = active_nics
+    nap 10 if nics.any? { |nic| nic.lock_set? }
```
jeremyevans:
```suggestion
    nap 10 if nics.any?(&:lock_set?)
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+      nic.incr_lock
+      private_subnet.create_tunnels(nics_snap, nic)
+    end
+
+    decr_add_new_nic
+    hop_wait_inbound_setup
+  end
+
+  label def refresh_keys
+    decr_refresh_keys
```
jeremyevans: In `add_new_nic`, we don't decrement the semaphore until right before hopping. Here we decrement the semaphore at the start of the method. Maybe this causes no problems, but it looks inconsistent. If there is a reason for the difference in behavior, can you add a comment to this call and the `decr_add_new_nic` in `add_new_nic` explaining why we do one early and one late?

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+  def gen_spi
+    "0x" + SecureRandom.bytes(4).unpack1("H*")
+  end
+
+  def gen_reqid
+    SecureRandom.random_number(100000) + 1
+  end
+
+  label def add_new_nic
+    register_deadline("wait", 3 * 60)
```
jeremyevans, line 55: We don't currently register a deadline in `refresh_keys`. Do you think we should do so there?

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+      nic.incr_lock
+    end
+
+    hop_wait_inbound_setup
+  end
+
+  label def wait_inbound_setup
+    nics = rekeying_nics
+    if nics.all? { |nic| nic.strand.label == "wait_rekey_outbound_trigger" }
+      nics.each(&:incr_trigger_outbound_update)
```
jeremyevans: N+1 query issue, can fix with:

```suggestion
      Semaphore.incr(nics.map(&:id), "trigger_outbound_update")
```

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+      hop_wait_outbound_setup
+    end
+
+    nap 5
+  end
+
+  label def wait_outbound_setup
+    nics = rekeying_nics
+    if nics.all? { |nic| nic.strand.label == "wait_rekey_old_state_drop_trigger" }
+      nics.each(&:incr_old_state_drop_trigger)
```
jeremyevans: N+1 query issue. Can fix with:

```suggestion
      Semaphore.incr(nics.map(&:id), "old_state_drop_trigger")
```

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+    nap 5
+  end
+
+  label def wait_old_state_drop
+    nics = rekeying_nics
+    if nics.all? { |nic| nic.strand.label == "wait" }
+      private_subnet.update(state: "waiting", last_rekey_at: Time.now)
+      nics.each do |nic|
+        nic.update(encryption_key: nil, rekey_payload: nil)
+        nic.unlock
```
jeremyevans: 2N+1 query issue. Can fix with:

```suggestion
      all_connected_nics.exclude(rekey_payload: nil).update(encryption_key: nil, rekey_payload: nil)
      Semaphore.where(strand_id: nics.map(&:id), name: "lock").delete(force: true)
```

Might want to add a private method for the `all_connected_nics.exclude(rekey_payload: nil)` code.

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+      hop_wait
+    end
+
+    nap 5
+  end
+
+  label def destroy
+    if private_subnet.nics.any? { |n| !n.vm_id.nil? }
```
jeremyevans: The false/nil difference doesn't matter here, so I would use:

```suggestion
    if private_subnet.nics.any?(&:vm_id)
```

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+      register_deadline(nil, 10 * 60, allow_extension: true) if private_subnet.nics.any? { |n| n.vm&.prevent_destroy_set? }
```
jeremyevans: N+1 query issue. Can be fixed with:

```suggestion
      unless Semaphore.where(strand_id: private_subnet.nics.map(&:vm_id).compact, name: "prevent_destroy").empty?
        register_deadline(nil, 10 * 60, allow_extension: true)
      end
```

prog/vnet/ubicloud/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+
+      Clog.emit("Cannot destroy subnet with active nics, first clean up the attached resources") { private_subnet }
+
+      nap 5
+    end
+
+    decr_destroy
+    private_subnet.firewalls.each { it.disassociate_from_private_subnet(private_subnet, apply_firewalls: false) }
```
jeremyevans: N+1 query issue. Can be fixed with:

```suggestion
    private_subnet.remove_all_firewalls
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -0,0 +1,163 @@
...
+
+      nap 5
+    end
+
+    decr_destroy
+    private_subnet.firewalls.each { it.disassociate_from_private_subnet(private_subnet, apply_firewalls: false) }
+
+    private_subnet.connected_subnets.each do |subnet|
+      private_subnet.disconnect_subnet(subnet)
+    end
```
jeremyevans, line 133: Also an N+1 query issue, but considering the complexity of `PrivateSubnet#disconnect_subnet`, probably best to leave as is.

prog/vnet/aws/vpc_nexus.rb
byucesoy: I think they they can be ignored. I'll remove this line.

prog/vnet/aws/vpc_nexus.rb
byucesoy: This was copy/paste from the earlier prog, but I think you are right. `remove_all_firewalls` makes more sense.

cc @furkansahin to confirm whether this is OK or if we are missing something.

prog/vnet/metal/subnet_nexus.rb
byucesoy: @furkansahin, what do you think about this?

furkansahin: Yeap, you can change it to the suggested code. The important thing is that it doesn't return 0 because it's a special case.

prog/vnet/aws/vpc_nexus.rb
furkansahin: That's alright

prog/vnet/metal/subnet_nexus.rb
furkansahin: We can move decr_refresh_keys to **after** we got the lock

prog/vnet/metal/subnet_nexus.rb
furkansahin: yes, let's do that

spec/prog/vnet/aws/vpc_nexus_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+      client.stub_responses(:describe_subnets, subnets: [{state: "available"}])
+    end
+
+    let(:nic) {
+      instance_double(Nic, vm_id: nil)
+    }
+
+    it "extends deadline if a vm prevents destroy" do
+      vm = Vm.new(family: "standard", cores: 1, name: "dummy-vm", location_id: Location::HETZNER_FSN1_ID).tap {
+        it.id = "788525ed-d6f0-4937-a844-323d4fd91946"
```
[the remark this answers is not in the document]
byucesoy: `it` is automatically defined as the object that is being worked on  in ruby blocks.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -0,0 +1,204 @@
...
+      is_destroying = true
+    end
+
+    unless is_destroying
+      when_destroy_set? do
+        incr_destroying
+        register_deadline(nil, 10 * 60)
+        hop_destroy
+      end
+    end
```
jeremyevans: This may be simpler. Up to you.

```suggestion
    when_destroy_set? do
      when_destroying_set? { return }
      incr_destroying
      register_deadline(nil, 10 * 60)
      hop_destroy
    end
```

This is a case where `when_destroying_unset?` would allow for simpler code.

prog/vnet/metal/subnet_nexus.rb
jeremyevans: This change wasn't made. I',m not sure it is required, if 6 digit numbers are allowed but not required. However, there doesn't seem an advantage to the current code over my suggestion.

jeremyevans (approved):

prog/vnet/metal/subnet_nexus.rb
byucesoy: I did this change, but it seems it is lost during one of the rebases. Weird... I fixed it again.

byucesoy: I mistakenly did it  in the original subnet nexus prog. Fixed it now.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -104,10 +104,10 @@ def gen_reqid
     nics = rekeying_nics
     if nics.all? { |nic| nic.strand.label == "wait" }
       private_subnet.update(state: "waiting", last_rekey_at: Time.now)
-      nics.each do |nic|
-        nic.update(encryption_key: nil, rekey_payload: nil)
-        nic.unlock
-      end
+
+      all_connected_nics.exclude(rekey_payload: nil).update(encryption_key: nil, rekey_payload: nil)
```
furkansahin, line 108: why are you calling `all_connected_nics.exclude` here, you already have `nics`

byucesoy: nics are not Sequel dataset, but materialized sequel models, so we cannot run `.update` on them.

PR #4180, opened 2025-11-13
Attempt to workaround the Linux Kernel race condition. 

pykello:
In production we had several incidents which caused host unavailability after logging the following in syslog:

- A per-cpu kworker processing work on the wrong CPU,
- Internal state of a worker pool has been corrupt: all kworkers in the pool are idle, but number of running kworkers in the pool is non-zero.

We think the state corruption is a consequence of a kworker in the pool having been executed on the wrong cpu.

As a consequence of the worker pool state corruption, some queued deferred works were never processed, which caused:

- IO stalls,
- VM stop timeouts,
- SSH authentication timeouts,
- ...

After reviewing the evidence in logs and Linux 6.8.0-87's source code, our current theory of why this happens is that there is a race condition between kworker creation and cgroup creation/deletion.

The cgroup side in this race condition is `update_tasks_cpumask` in `kernel/cgroup/cpuset.c`, which updates allowed cpumask of tasks after a `root` cpu partition has been created or deleted. In this case, list of cpus available for the system cgroup and other cgroups changes.

In rare cases and under a race condition, `update_tasks_cpumask` might migrate a kworker that is just being created to a wrong cpu.

This commit tries to workaround this race condition by setting `cpuset.cpus.partition` to `member` instead of `root`. In this case, list of available cpus to other cgroups won't change and `update_tasks_cpumask` won't be called, avoiding the described race condition.

The downside of using `member` is that other services might use a bit of VM's cpu, which is acceptable if we get a more stable system.

Note that the described race condition is just our current understanding of the issue, and we might have still missed few things.
Therefore, we might revert this commit and try other workarounds if we still see the issue after this change.

----

> [!IMPORTANT]
> Change `cpuset.cpus.partition` from `root` to `member` to avoid Linux kernel race condition, updating code and tests accordingly.
> 
>   - **Behavior**:
>     - Change `cpuset.cpus.partition` from `root` to `member` in `vm_host_slice.rb`, `slice_setup.rb`, and `slice_setup_e2e_spec.rb` to avoid Linux kernel race condition.
>     - Update `up?()` in `vm_host_slice.rb` to accept both `root` and `member` as valid partition states.
>   - **Tests**:
>     - Update tests in `slice_setup_spec.rb` and `vm_host_slice_nexus_spec.rb` to check for `member` instead of `root`.
>     - Ensure tests validate both `root` and `member` as valid states for `cpuset.cpus.partition`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0fd6f083704d6bef0b613d4f53259cedd8a15ef7. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/host/e2e/slice_setup_e2e_spec.rb
+++ b/rhizome/host/e2e/slice_setup_e2e_spec.rb
@@ -20,8 +20,8 @@ RSpec.describe SliceSetup do
     # check that the slice is started
     expect(r("systemctl show -p ActiveState --value #{slice_name}")).to eq("active\n")
 
-    # check that the cpuset.cpus.partition file contains "root"
-    expect(File.read("/sys/fs/cgroup/#{slice_name}/cpuset.cpus.partition")).to eq("root\n")
+    # check that the cpuset.cpus.partition file contains "member"
+    expect(File.read("/sys/fs/cgroup/#{slice_name}/cpuset.cpus.partition")).to eq("member\n")
 
     # check allowed cpus
     expect(File.read("/sys/fs/cgroup/#{slice_name}/cpuset.cpus")).to eq("#{cpuset}\n")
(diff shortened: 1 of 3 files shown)


fdr (approved):
I love this commit message. So often it's the tiniest changes that justify them. Let's hope your hard work pays off.

jeremyevans (approved):

PR #4181, opened 2025-11-13
Avoid M+N+1 query issue in update_aws_firewall_rules

jeremyevans:
Without this, there are N+1 query issues for both the private_subnet_firewalls and vm_firewalls associations.

----

> [!IMPORTANT]
> Fixes N+1 query issue in `update_aws_firewall_rules` by using eager loading for `firewall_rules`.
> 
>   - **Performance Optimization**:
>     - Fixes N+1 query issue in `update_aws_firewall_rules` by using eager loading for `firewall_rules` in `update_firewall_rules.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for ce68be0f7ea6bb5b85e66ad6b826691b77329ee9. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #4185, opened 2025-11-14
Separate AWS/Metal update_firewall_rules files - Phase 1

furkansahin:
## Add vm method to fetch flattened firewall_rules and combine the ip4/6_rules
This was being referred multiple times from different files

## Start using the separate update_firewall_rules files

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -200,6 +200,12 @@ class Vm < Sequel::Model
     id.to_s[0..7]
   end
 
+  def update_firewall_rules_prog
+    prog = Prog::Vnet
+    prog = location.aws? ? prog::Aws : prog::Metal
+    prog::UpdateFirewallRules
+  end
+
   def inhost_name
     self.class.ubid_to_name(UBID.from_uuidish(id))
   end
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -471,7 +471,7 @@ class Prog::Vm::Nexus < Prog::Base
     end
 
     decr_update_firewall_rules
-    push Prog::Vnet::UpdateFirewallRules, {}, :update_firewall_rules
+    push vm.update_firewall_rules_prog, {}, :update_firewall_rules
   end
 
   label def update_spdk_dependency
--- a/spec/prog/vm/nexus_spec.rb
+++ b/spec/prog/vm/nexus_spec.rb
@@ -886,11 +886,15 @@ RSpec.describe Prog::Vm::Nexus do
   end
 
   describe "#update_firewall_rules" do
-    it "hops to wait_firewall_rules" do
-      vm.incr_update_firewall_rules
-      expect(nx).to receive(:push).with(Prog::Vnet::UpdateFirewallRules, {}, :update_firewall_rules)
-      expect { nx.update_firewall_rules }
-        .to change { vm.reload.update_firewall_rules_set? }.from(true).to(false)
+    ["Aws", "Metal"].each do |provider|
+      it "hops to wait_firewall_rules" do
+        vm.incr_update_firewall_rules
+        expect(vm).to receive(:location).and_return(instance_double(Location, aws?: provider == "Aws"))
+        prog = (provider == "Aws") ? Prog::Vnet::Aws::UpdateFirewallRules : Prog::Vnet::Metal::UpdateFirewallRules
+        expect(nx).to receive(:push).with(prog, {}, :update_firewall_rules)
+        expect { nx.update_firewall_rules }
+          .to change { vm.reload.update_firewall_rules_set? }.from(true).to(false)
+      end
     end
 
     it "hops to wait if firewall rules are applied" do


model/vm.rb
```diff
@@ -196,6 +196,10 @@ def self.ubid_to_name(id)
     id.to_s[0..7]
   end
 
+  def update_firewall_rules_prog
+    Prog.const_get("Prog::Vnet::#{location.aws? ? "Aws" : "Metal"}::UpdateFirewallRules")
```
jeremyevans: This works but looks weird. Generally if you are using `SomeClass.const_get`, you are passing some relative path from `SomeClass`.  You are currently relying on a behavior that Ruby would like to get rid of, but keeps only for backwards compatibility, which is that `const_get` will do the equivalent of a absolute lookup if the constant is found during relative lookup.

I would avoid `const_get` here:

```suggestion
    prog = Prog::Vnet
    prog = location.aws? ? prog::Aws : prog::Metal
    prog::UpdateFirewallRules
```

prog/vnet/aws/update_firewall_rules.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+class Prog::Vnet::Aws::UpdateFirewallRules < Prog::Base
+  subject_is :vm
+
+  FirewallRuleObj = Struct.new(:cidr, :port_range)
+
```
jeremyevans: This appears to be unused.

```suggestion
```

prog/vnet/aws/update_firewall_rules.rb
```diff
@@ -0,0 +1,91 @@
...
+  def before_run
+    pop "firewall rule is added" if vm.destroy_set?
+  end
+
+  label def update_firewall_rules
+    rules = vm.firewalls.flat_map(&:firewall_rules)
```
jeremyevans: This misses the M+N+1 query issue fix in #4181 that is waiting for your review:


```suggestion
    rules = vm.firewalls(eager: :firewall_rules).flat_map(&:firewall_rules)
```

prog/vnet/aws/update_firewall_rules.rb
```diff
@@ -0,0 +1,91 @@
...
+    permissions = rules.select(&:port_range).map! do |rule|
```
jeremyevans: This `select(&:port_range)` looked weird to me, as it would only be necessary if we allowed NULL values for `port_range`. I checked, and surprisingly to me, `firewall_rule.port_range` allows NULL values, as does `firewall_rule.cidr`. So the code seems correct, but it begs the question, why do we allow NULL values for either column? IMO, we should populate the values and mark the columns non-NULL.

It seems simpler to combine this loop and the following loop:

```suggestion
    rules.select(&:port_range).each do |rule|
```

prog/vnet/aws/update_firewall_rules.rb
```diff
@@ -0,0 +1,91 @@
...
+      perm
+    end
+
+    permissions.each do |perm|
+      aws_client.authorize_security_group_ingress({
+        group_id: vm.private_subnets.first.private_subnet_aws_resource.security_group_id,
+        ip_permissions: [perm]
+      })
+    rescue Aws::EC2::Errors::InvalidPermissionDuplicate
+      next
```
jeremyevans:
```suggestion
      
      begin
        aws_client.authorize_security_group_ingress({
          group_id: vm.private_subnets.first.private_subnet_aws_resource.security_group_id,
          ip_permissions: [perm]
        })
      rescue Aws::EC2::Errors::InvalidPermissionDuplicate
        next
      end
```

prog/vnet/aws/update_firewall_rules.rb
```diff
@@ -0,0 +1,91 @@
...
+        ip_protocol: "tcp",
+        from_port: rule.port_range.begin,
+        to_port: rule.port_range.end - 1
+      }
+      if rule.ip6?
+        perm[:ipv_6_ranges] = [{cidr_ipv_6: rule.cidr.to_s}]
+      else
+        perm[:ip_ranges] = [{cidr_ip: rule.cidr.to_s}]
+      end
+      perm
```
jeremyevans: Not necessary if combining loops.

```suggestion
```

prog/vnet/aws/update_firewall_rules.rb
```diff
@@ -0,0 +1,91 @@
...
+      })
+    rescue Aws::EC2::Errors::InvalidPermissionDuplicate
+      next
+    end
+
+    hop_remove_aws_old_rules
+  end
+
+  label def remove_aws_old_rules
+    rules = vm.firewalls.map(&:firewall_rules).flatten
```
jeremyevans:
```suggestion
    rules = vm.firewalls.(eager: :firewall_rules).flat_map(&:firewall_rules)
```

You may want to have a method for this since it is used in multiple places.

prog/vnet/aws/update_firewall_rules.rb
```diff
@@ -0,0 +1,91 @@
...
+    ip4_rules = rules.select { !it.ip6? && it.port_range }
+    ip6_rules = rules.select { it.ip6? && it.port_range }
```
jeremyevans:
```suggestion
    ip6_rules, ip4_rules = rules.select(&:port_range).partition(&:ip6?)
```

prog/vnet/metal/update_firewall_rules.rb
```diff
@@ -0,0 +1,302 @@
...
+  subject_is :vm
+
+  FirewallRuleObj = Struct.new(:cidr, :port_range)
+
+  def before_run
+    pop "firewall rule is added" if vm.destroy_set?
+  end
+
+  label def update_firewall_rules
+    rules = vm.firewalls.map(&:firewall_rules).flatten
```
jeremyevans:
```suggestion
    rules = vm.firewalls(eager: :firewall_rules).flat_map(&:firewall_rules)
```

jeremyevans (commented):
I didn't review most of the metal code, as I think it is the same as the existing `UpdateFirewallRules` prog.

When splitting things in the future, please use the following approach:

* Start with a commit just copies the files to the new location with no changes.
* Then add a commit making changes to those copies.

That makes it possible see what changes are being made to each file using a commit-by-commit review. With the current approach, all of the code looks new and I have no idea which parts are new and which are unmodified.

prog/vnet/aws/update_firewall_rules.rb
furkansahin: That logic is a leftover from a previous refactoring. I'll create a separate PR for that. I don't think that it should be a part of this.

prog/vnet/aws/update_firewall_rules.rb
furkansahin: approved that one

furkansahin:
@jeremyevans , thank you! Please have a look at it again.

model/vm.rb
furkansahin: thank you

model/vm.rb
```diff
@@ -97,6 +97,10 @@ def firewalls(opts = {})
     private_subnet_firewalls(opts) + vm_firewalls(opts)
   end
 
+  def flattened_firewall_rules
```
jeremyevans: This seems simpler (flattening is an implementation detail):

```suggestion
  def firewall_rules
```

prog/vnet/aws/update_firewall_rules.rb
jeremyevans: Since it wasn't critical, I closed that PR to avoid conflicts with this one.

PR #4187, opened 2025-11-14
Inject static node host entries into CoreDNS Corefile after kubernetes block

mohi-kalantari:
Add logic to parse the CoreDNS Corefile from the kube-system/coredns ConfigMap, locate the kubernetes plugin block, and insert a generated hosts block for all K8s nodes immediately after it. The code removes any previous Ubicloud hosts block to keep the operation idempotent, builds a fresh block from current node IP/name pairs, and updates the ConfigMap via `kubectl replace -f -`. Includes error handling for missing kubernetes block start or end.

Call incr_sync_worker_dns_config when nodes are added or removed

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -78,6 +78,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
     incr_install_metrics_server
     incr_sync_worker_mesh
     incr_install_csi
+    incr_sync_internal_dns_config
     hop_create_load_balancers
   end
 
--- a/prog/kubernetes/provision_kubernetes_node.rb
+++ b/prog/kubernetes/provision_kubernetes_node.rb
@@ -228,6 +228,7 @@ CONFIG
 
   label def approve_new_csr
     kubernetes_cluster.sshable.cmd("sudo kubectl --kubeconfig /etc/kubernetes/admin.conf get csr | awk '/Pending/ && /kubelet-serving/ && /'\"#{node.name}\"'/ {print $1}' | xargs -r sudo kubectl --kubeconfig /etc/kubernetes/admin.conf certificate approve")
+    kubernetes_cluster.incr_sync_internal_dns_config
     pop({node_id: node.id})
   end
 end
--- a/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
+++ b/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
@@ -167,6 +167,7 @@ RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
       expect(nx).to receive(:register_deadline)
       expect(nx).to receive(:incr_install_metrics_server)
       expect(nx).to receive(:incr_sync_worker_mesh)
+      expect(nx).to receive(:incr_sync_internal_dns_config)
       expect(nx).to receive(:incr_install_csi)
       expect { nx.start }.to hop("create_load_balancers")
     end
--- a/prog/kubernetes/kubernetes_node_nexus.rb
+++ b/prog/kubernetes/kubernetes_node_nexus.rb
@@ -92,6 +92,7 @@ class Prog::Kubernetes::KubernetesNodeNexus < Prog::Base
   label def destroy
     kubernetes_node.vm.incr_destroy
     kubernetes_node.destroy
+    cluster.incr_sync_internal_dns_config
     pop "kubernetes node is deleted"
   end
 end
--- a/spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
+++ b/spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
@@ -166,6 +166,7 @@ RSpec.describe Prog::Kubernetes::KubernetesNodeNexus do
     it "destroys the vm and itself" do
       expect(kd.vm).to receive(:incr_destroy)
       expect(kd).to receive(:destroy)
+      expect(kd.kubernetes_cluster).to receive(:incr_sync_internal_dns_config)
       expect { nx.destroy }.to exit({"msg" => "kubernetes node is deleted"})
     end
   end
--- a/spec/prog/kubernetes/provision_kubernetes_node_spec.rb
+++ b/spec/prog/kubernetes/provision_kubernetes_node_spec.rb
@@ -342,6 +342,7 @@ table ip6 pod_access {
     it "approves the csr" do
       sshable = instance_double(Sshable)
       expect(kubernetes_cluster.functional_nodes.first).to receive(:sshable).and_return(sshable)
+      expect(kubernetes_cluster).to receive(:incr_sync_internal_dns_config)
       expect(sshable).to receive(:cmd).with("sudo kubectl --kubeconfig /etc/kubernetes/admin.conf get csr | awk '/Pending/ && /kubelet-serving/ && /'\"#{node.name}\"'/ {print $1}' | xargs -r sudo kubectl --kubeconfig /etc/kubernetes/admin.conf certificate approve")
       expect { prog.approve_new_csr }.to exit({node_id: prog.node.id})
     end


prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -294,6 +299,52 @@
...
+    # Remove existing Ubicloud hosts block if present (matches full block with comments)
+    # Example matched block (with ~8-space indent for contents):
+    #     hosts {
+    #         # Ubicloud Hosts
+    #         178.63.152.192 kcg0zevswgsp6hqepg9de83wwq-nz1df
+    #         178.63.152.197 knbvkn6c55gt1mfhab6znfv4j5-ukgtf
+    #         # End Ubicloud Hosts
+    #         fallthrough
+    #     }
+    corefile.gsub!(/^\s*hosts\s*\{\s*\n(?:\s*# Ubicloud Hosts\n)?(?:\s*        .*\n)*?\s*(?:# End Ubicloud Hosts\n)?\s*        fallthrough\s*\n\s*\}\s*$/m, "")
```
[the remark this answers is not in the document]
jeremyevans: Ignore this. You can test and see that in Ruby 3.2+, the regexp runs in linear time:

```ruby
Regexp.linear_time?(/^\s*hosts\s*\{\s*\n(?:\s*# Ubicloud Hosts\n)?(?:\s*        .*\n)*?\s*(?:# End Ubicloud Hosts\n)?\s*        fallthrough\s*\n\s*\}\s*$/m)
# => true
```

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -294,6 +298,52 @@ def billing_rate_for(type, family)
     hop_wait
   end
 
+  # Dynamically injects static hosts DNS records for K8s nodes into CoreDNS Corefile after the kubernetes plugin block,
+  # then updates and replaces the kube-system/coredns ConfigMap to apply the changes.
+  label def sync_worker_dns_config
+    decr_sync_worker_dns_config
+
+    dns_records = kubernetes_cluster.all_functional_nodes.map { [it.vm.ip4.to_s, it.name].join(" ") }
```
jeremyevans: Does K8s not support IPv6?

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -294,6 +299,52 @@ def billing_rate_for(type, family)
...
+    fail "Closing brace not found." unless end_idx
+    kubernetes_block_end = kubernetes_block_start + 1 + end_idx
+
+    # Build new hosts block with comment markers for future idempotency (split records for proper indentation)
+    hosts_lines = ["    hosts {"] +
+      ["        # Ubicloud Hosts"] +
+      dns_records.map { |l| "        #{l}" } +
+      ["        # End of Ubicloud Hosts",
+        "        fallthrough",
+        "    }"]
```
jeremyevans: It is hard to see visually that the correct indentation is used. You may want to consider an approach using `indent = " " * 4`, and doing `indent + "hosts {"`, `indent + indent + l`, and similar.

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -558,6 +564,34 @@
...
+    let(:sshable) { instance_double(Sshable) }
+    let(:node) { KubernetesNode.create(vm_id: create_vm.id, kubernetes_cluster_id: kubernetes_cluster.id) }
+
+    before do
+      expect(kubernetes_cluster).to receive(:client).and_return(client)
+      allow(kubernetes_cluster).to receive(:sshable).and_return(sshable)
+    end
+
+    it "adds the ubicloud block and replaces the configmap" do
+      expect(client).to receive(:kubectl).with("-n kube-system get cm coredns -oyaml").and_return("apiVersion: v1\ndata:\n  Corefile: |-\n    .:53 {\n        errors\n        health {\n           lameduck 5s\n        }\n        ready\n        kubernetes cluster.local in-addr.arpa ip6.arpa {\n           pods insecure\n           fallthrough in-addr.arpa ip6.arpa\n           ttl 30\n        }\n        prometheus :9153\n        forward . /etc/resolv.conf {\n           max_concurrent 1000\n        }\n        cache 30 {\n           disable success cluster.local\n           disable denial cluster.local\n        }\n        loop\n        reload\n        loadbalance\n    }\nkind: ConfigMap\nmetadata:\n  name: coredns\n  namespace: kube-system")
```
jeremyevans: This and the remaining methods would be much better off as a heredoc, so you could visually check that the indentation is correct.

prog/kubernetes/provision_kubernetes_node.rb
```diff
@@ -228,6 +228,7 @@ def before_run
 
   label def approve_new_csr
     kubernetes_cluster.sshable.cmd("sudo kubectl --kubeconfig /etc/kubernetes/admin.conf get csr | awk '/Pending/ && /kubelet-serving/ && /'\"#{node.name}\"'/ {print $1}' | xargs -r sudo kubectl --kubeconfig /etc/kubernetes/admin.conf certificate approve")
+    kubernetes_cluster.incr_sync_worker_dns_config
```
aamederen: this could be unrelated but shall we do incr_sync_worker_mesh here as well as on KubernetedNodeNexus#destroy?

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -294,6 +299,52 @@ def billing_rate_for(type, family)
...
+    #         # Ubicloud Hosts
+    #         178.63.152.192 kcg0zevswgsp6hqepg9de83wwq-nz1df
+    #         178.63.152.197 knbvkn6c55gt1mfhab6znfv4j5-ukgtf
+    #         # End Ubicloud Hosts
+    #         fallthrough
+    #     }
+    corefile.gsub!(/^\s*hosts\s*\{\s*\n(?:\s*# Ubicloud Hosts\n)?(?:\s*        .*\n)*?\s*(?:# End Ubicloud Hosts\n)?\s*        fallthrough\s*\n\s*\}\s*$/m, "")
+    lines = corefile.split("\n")
+
+    # The new block should be right after the kuberentes block, so we aim to find the block and insert
```
aamederen: nit: typo "kuberentes"

prog/kubernetes/kubernetes_cluster_nexus.rb
aamederen: You can also construct such multi-line strings using some heredoc for better readability.

prog/kubernetes/provision_kubernetes_node.rb
mohi-kalantari: Eventually doing one incr_sync_worker_dns_config on KubernetedNodeNexus#destroy must be enough but until the draining is finished, a new node won't be resolved. Doing it once here allows to cover that duration.

prog/kubernetes/kubernetes_cluster_nexus.rb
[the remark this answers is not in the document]
mohi-kalantari: falso positive

prog/kubernetes/kubernetes_cluster_nexus.rb
mohi-kalantari: totally forgot about that. thanks.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -294,6 +299,54 @@ def billing_rate_for(type, family)
...
+    # Build new hosts block with comment markers for future idempotency (split records for proper indentation)
+    hosts_lines = <<~HOSTS
+    hosts {
+        # Ubicloud Hosts
+#{dns_records.map { |l| "        #{l}" }.join("\n")}
+        # End of Ubicloud Hosts
+        fallthrough
+    }
+    HOSTS
+    new_lines = lines[0...kubernetes_block_end + 1] + hosts_lines.split("\n") + lines[kubernetes_block_end + 1..]
```
aamederen, line 342: Maybe

```suggestion
    new_corefile = lines[0...kubernetes_block_end + 1].join("\n") + hosts_lines + lines[kubernetes_block_end + 1..].join("\n")
```

Instead of splitting in one line and joining back in the next

prog/kubernetes/provision_kubernetes_node.rb
mohi-kalantari: Now I understood what you mean. Yes we should add those incr_sync_worker_mesh too.

prog/kubernetes/kubernetes_cluster_nexus.rb
mohi-kalantari: just adding the joins in the first line is not enough because the last line won't get \n. so it would need an extra +"\n"

PR #4188, opened 2025-11-14
Decrease mocking in nexus specs


enescakir:
### Here’s my PR to help me use my 5-hour flight productively while offline. I removed excessive mocking in the nexus specs. This adds more verification, especially for destroy tests, while also removing 78 lines of code.

- **Rename prj to project in Nexus spec**
  It's better to use consistent `let(:)` variable names across specs to
  make moving code around easier. Since most of our specs use project
  instead of prj, I’m updating it here for consistency.
  

- **Create vm related resources instead of mocking in nexus specs**
  Currently, the tests do not persist these models to the database and
  mock some parts of them.
  
  Since we try to avoid mocking database operations, it’s better to create
  real resources.
  
  Because we don’t need them in all tests, I’ve moved their creation to
  the relevant tests in future commits. I also removed mocking in those
  commits.
  

- **Remove project mocks in nexus specs**
  

- **Remove vm host mocks in nexus specs**
  I was surprised by some tests, especially the destroy ones.
  
  After destroying a VM host/slice, the tests didn’t actually check if the
  resources were freed which I think is the most important thing to verify
  in destroy tests. Instead, they were just mocking some data and checking
  that associated objects were called.
  
  I’ve removed the VM host mocks and added real verification that
  resources get freed in destroy tests. While I was at it, I also
  reordered some destroy tests to make them cover specific cases.
  
  Finally, the `#destroy_slice` and `#destroy` tests were mixed together,
  so I separated them for clarity.
  

- **Remove billing record and storage volume mocks in nexus specs**
  Also we don't need to create these resources for each tests.
  

- **Use real strand for nexus tests**
  It helps us to remove semaphore and stack mocks.
  

- **Remove nic mocks in nexus specs**

### DIFF
--- a/spec/prog/vm/nexus_spec.rb
+++ b/spec/prog/vm/nexus_spec.rb
@@ -1092,7 +1092,7 @@ RSpec.describe Prog::Vm::Nexus do
   describe "#destroy_slice" do
     it "#destroy_slice when no slice" do
       expect { nx.destroy_slice }.to exit({"msg" => "vm deleted"})
-      expect(Vm[vm.id]).to be_nil
+      expect(vm.exists?).to be(false)
     end
 
     it "#destroy_slice with a slice" do
@@ -1103,7 +1103,7 @@ RSpec.describe Prog::Vm::Nexus do
       expect { nx.destroy_slice }.to exit({"msg" => "vm deleted"})
       expect(vm_host_slice.reload.enabled).to be(false)
       expect(Semaphore[strand_id: vm_host_slice.id, name: "destroy"]).not_to be_nil
-      expect(Vm[vm.id]).to be_nil
+      expect(vm.exists?).to be(false)
     end
 
     it "skips destroy slice when slice already disabled" do
@@ -1114,7 +1114,7 @@ RSpec.describe Prog::Vm::Nexus do
       expect { nx.destroy_slice }.to exit({"msg" => "vm deleted"})
       expect(vm_host_slice.reload.enabled).to be(false)
       expect(Semaphore[strand_id: vm_host_slice.id, name: "destroy"]).to be_nil
-      expect(Vm[vm.id]).to be_nil
+      expect(vm.exists?).to be(false)
     end
   end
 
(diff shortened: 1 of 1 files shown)


enescakir:
I tried to split this into multiple commits, but they might not be fully isolated

fdr:
Not that tests aren't important to have legible history to a degree, but I think you did the right thing here, for something as shotgun surgery-ish as this: getting it done and reducing conflict drag is more important than our usual standard of manicuring things for implementation changes.

fdr:
(@jeremyevans said he'd review, I was going to glance and approve to ensure you didn't pay conflict costs on this, but as he's the methodological master, I'm not about to pre-empt it)

spec/prog/vm/nexus_spec.rb
```diff
@@ -998,178 +976,174 @@ def requested_disk_size(st)
...
-    it "hops to wait_aws_vm_destroyed if vm is in aws" do
-      location_id = Location.create(name: "us-west-2", provider: "aws", project_id: project.id, display_name: "us-west-2", ui_name: "us-west-2", visible: true).id
-      vm.update(project_id: project.id, location_id:)
-      st.update(prog: "Vm::Nexus", label: "destroy", stack: [{}])
-      child = Strand.create(parent_id: st.id, prog: "Aws::Instance", label: "start", stack: [{}])
-      expect(nx).to receive(:bud).with(Prog::Aws::Instance, {"subject_id" => vm.id}, :destroy)
-      expect { nx.destroy }.to hop("wait_aws_vm_destroyed")
-      expect(Semaphore[strand_id: child.id, name: "destroy"]).not_to be_nil
-      expect(vm.display_state).to eq("deleting")
+      expect(Vm[vm.id]).to be_nil
```
jeremyevans: An alternative would be `expect(vm.exists?).to be false` (maybe there is a better way to do that in rspec).

spec/prog/vm/nexus_spec.rb
```diff
@@ -162,17 +147,17 @@ def requested_disk_size(st)
 
     it "requests as many gpus as specified" do
       st = described_class.assemble("some_ssh key", project.id, size: "standard-2", gpu_count: 2)
-      expect(st.stack[0]["gpu_count"]).to eq(2)
+      expect(st.stack.first["gpu_count"]).to eq(2)
```
jeremyevans, line 150: This seems unrelated to the other improvements. No objections, though.

jeremyevans (approved):
This is a huge improvement, and we would be much better off if the rest of the specs were converted similarly. Thank you very much for working on this.

PR #4191, opened 2025-11-15
postgres create: pg_config & pgbouncer fields

serprex:
default configuration lacking what's necessary for failover slots,
adding pg_config to create request avoids needing to manage immediately updating config after create,
which helps with api usage where config tuning is a common use case

----

> [!IMPORTANT]
> Adds `pg_config` field to Postgres creation, allowing custom configurations, with validation and updated OpenAPI spec.
> 
>   - **Behavior**:
>     - Adds `pg_config` field to Postgres creation in `postgres_post` in `helpers/postgres.rb` and `assemble` in `prog/postgres/postgres_resource_nexus.rb`.
>     - Validates `pg_config` using `PostgresConfigValidator`.
>     - Updates OpenAPI spec in `openapi.yml` to include `pg_config` in Postgres creation request.
>   - **Tests**:
>     - Adds tests in `postgres_spec.rb` for valid and invalid `pg_config` values during Postgres creation.
>     - Verifies `pg_config` is correctly set and retrievable via API.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 5cf5aacd792a15eccd232c4fbcea2e6daa50beff. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli-commands/pg/post/create.rb
+++ b/cli-commands/pg/post/create.rb
@@ -9,6 +9,8 @@ UbiCli.on("pg").run_on("create") do
     on("-s", "--size=size", Option::POSTGRES_SIZE_OPTIONS.keys, "server size")
     on("-S", "--storage-size=size", Option::POSTGRES_STORAGE_SIZE_OPTIONS, "storage size GB")
     on("-v", "--version=version", Option::POSTGRES_VERSION_OPTIONS[PostgresResource::Flavor::STANDARD], "PostgreSQL version")
+    on("-c", "--pg-config=config", "postgres config (e.g key1=value1,key2=value2)")
+    on("-u", "--pgbouncer-config=config", "pgbouncer config (e.g. key1=value1,key2=value2)")
     on("-t", "--tags=tags", "tags (e.g. key1=value1,key2=value2)")
     on("-R", "--restrict-by-default", "restrict access by default (add firewall rules to allow access)")
   end
@@ -21,6 +23,8 @@ UbiCli.on("pg").run_on("create") do
   run do |opts, cmd|
     params = underscore_keys(opts[:pg_create])
     pg_tags_to_hash(params, cmd)
+    params_to_hash(params, :pg_config, "config", cmd)
+    params_to_hash(params, :pgbouncer_config, "pgbouncer config", cmd)
     id = sdk.postgres.create(location: @location, name: @name, **params).id
     response("PostgreSQL database created with id: #{id}")
   end
--- a/helpers/postgres.rb
+++ b/helpers/postgres.rb
@@ -10,6 +10,8 @@ class Clover
     storage_size = typecast_params.pos_int("storage_size")
     ha_type = typecast_params.nonempty_str("ha_type", PostgresResource::HaType::NONE)
     version = typecast_params.nonempty_str("version", PostgresResource::DEFAULT_VERSION)
+    user_config = typecast_params.Hash("pg_config", {})
+    pgbouncer_user_config = typecast_params.Hash("pgbouncer_config", {})
     tags = typecast_params.array(:Hash, "tags", [])
     with_firewall_rules = !typecast_params.bool("restrict_by_default")
 
@@ -30,6 +32,18 @@ class Clover
     requested_postgres_vcpu_count = (requested_standby_count + 1) * parsed_size.vcpu_count
     Validation.validate_vcpu_quota(@project, "PostgresVCpu", requested_postgres_vcpu_count)
 
+    pg_validator = Validation::PostgresConfigValidator.new(version)
+    pg_errors = pg_validator.validation_errors(user_config)
+
+    pgbouncer_validator = Validation::PostgresConfigValidator.new("pgbouncer")
+    pgbouncer_errors = pgbouncer_validator.validation_errors(pgbouncer_user_config)
+
+    if pg_errors.any? || pgbouncer_errors.any?
+      pg_errors = pg_errors.map { |key, value| ["pg_config.#{key}", value] }.to_h
+      pgbouncer_errors = pgbouncer_errors.map { |key, value| ["pgbouncer_config.#{key}", value] }.to_h
+      raise Validation::ValidationFailed.new(pg_errors.merge(pgbouncer_errors))
+    end
+
     pg = nil
     DB.transaction do
       pg = Prog::Postgres::PostgresResourceNexus.assemble(
@@ -41,7 +55,9 @@ class Clover
         target_version: version,
         ha_type:,
         with_firewall_rules:,
-        flavor:
+        flavor:,
+        user_config:,
+        pgbouncer_user_config:
       ).subject
       pg.update(tags:)
       audit_log(pg, "create")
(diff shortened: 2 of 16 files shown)


fdr:
dumb terminology question: do we _really_ want to call it `pg_config`, which as an important  binary by the same name?

```
fdr@albatross ~/c/clover ((5bea460b))> pg_config
BINDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/bin
DOCDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/share/doc
HTMLDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/share/doc
INCLUDEDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/include
PKGINCLUDEDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/include
INCLUDEDIR-SERVER = /home/fdr/.local/share/mise/installs/postgres/17.6/include/server
LIBDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/lib
PKGLIBDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/lib
LOCALEDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/share/locale
MANDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/share/man
SHAREDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/share
SYSCONFDIR = /home/fdr/.local/share/mise/installs/postgres/17.6/etc
PGXS = /home/fdr/.local/share/mise/installs/postgres/17.6/lib/pgxs/src/makefiles/pgxs.mk
CONFIGURE =  '--prefix=/home/fdr/.local/share/mise/installs/postgres/17.6' '--with-uuid=e2fs' '--with-openssl' '--with-zlib' '--with-libraries='''/opt/local/lib:/sw/lib:/usr/local/opt/openssl/lib:/opt/openssl/lib:/lib:/openssl/lib:/libressl/lib:/lib:/usr/local/lib:/usr/lib'''' '--with-includes='''/opt/local/lib:/sw/lib:/usr/local/opt/openssl/include:/opt/openssl/include:/opt/homebrew/include:/include:/openssl/include:/libressl/include:/include:/usr/local/include:/usr/local/lib:/usr/lib''''
CC = gcc
CPPFLAGS = -D_GNU_SOURCE -I/usr/local/include -I/usr/local/lib
CFLAGS = -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wshadow=compatible-local -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-format-truncation -Wno-stringop-truncation -O2
CFLAGS_SL = -fPIC
LDFLAGS = -L/lib -L/lib -L/usr/local/lib -Wl,--as-needed -Wl,-rpath,'/home/fdr/.local/share/mise/installs/postgres/17.6/lib',--enable-new-dtags
LDFLAGS_EX = 
LDFLAGS_SL = 
LIBS = -lpgcommon -lpgport -lssl -lcrypto -lz -lreadline -lm 
VERSION = PostgreSQL 17.6
```

jeremyevans (approved):
It seems inconsistent to do this only for `user_config` and not `pgbouncer_user_config`.

Generally, when adding support to the API, we also add CLI support. We aim for CLI support for all API features.

However, I can work on support for both of those, so this looks good to me.

fdr:
ah nevermind, we committed this sin in the pre-existing endpoint nomenclature. Alas, a small scar for-ever I guess. Well how often are these two going to generate lexical confusion anyway? Probably not that often.

serprex:
There's less urgency here, I'll add pgbouncer later tonight. If you have a PR I can use as reference for CLI support I can take a stab at that too

jeremyevans:
> There's less urgency here, I'll add pgbouncer later tonight. If you have a PR I can use as reference for CLI support I can take a stab at that too

For the cli command, you'll want to update https://github.com/ubicloud/ubicloud/blob/main/cli-commands/pg/post/create.rb to support the config entries. You can probably borrow code from https://github.com/ubicloud/ubicloud/blob/main/cli-commands/pg/post/add-config-entries.rb. Instead of separate arguments per config entry, you'll want a command line option (maybe `-c`) that takes a single string value, and split that string.

serprex:
Went with `-u` for pgbouncer config because it's a sideways `c` & letter is in pgbouncer

serprex:
Squashed. Not much to describe since this PR doesn't really add anything, just plumbs existing config code into create path

fdr (changes requested):
It took me at least a minute searching my recent memory to understand the commit message, and I would not be able to understand it a couple of weeks from now at all. Let's change it. I will make a proposal:

```
Allow Postgres configuration in create API to support failover slots

Currentlhy, Postgres's default configuration prevents creating
failover replication slots, requiring a second API call to update
settings after creation.

Add a configuration field to the create API payload, allowing users
to enable these slots (and other settings) upon creation.

Also expose PgBouncer configuration in the same way for consistency.
```

I used a LLM having related the rough draft to deal with things like converting to the Git-nominal imperative mood.

I had the same LLM write a prompt for what my iterations descriptively accomplished. Here it is. I have found this form of prompt "boosting" pretty effective as a general practice, especially if you perform a deep research to obtain context from the web, e.g. style manuals people have written:

```
**Git Commit Message Refinement Prompt**

Edit this git commit message for conciseness, clarity, and directness. Apply these principles:

1. **Structure:** Problem → Solution. Lead with the "why" (what limitation or bug does this fix?), then explain the "what" (what change enables it?).

2. **Tone & Voice:**
   - Use imperative mood in the body (e.g., "Add a field..." not "This change adds...").
   - Avoid filler words and adverbs (e.g., "previously," "immediately," "instantiation").
   - Prefer active voice over passive.

3. **Vocabulary:**
   - Replace jargon with precise technical terms (e.g., "creation" over "instantiation").
   - Cut buzzwords and abstract language.

4. **Title:**
   - Concise and action-oriented (verb + object).
   - Include high-value context if it clarifies the motivation (e.g., "to support failover slots").

5. **Secondary Changes:**
   - Separate related but non-primary changes into their own sentences or paragraphs.
   - Use "Also" to signal secondary work and justify it briefly (e.g., "for consistency").

6. **Completeness:**
   - Include all systems/components touched, even if secondary (for `git log --grep` searchability).
   - Keep each logical unit concise; avoid running multiple ideas into one sentence.

***

**Original:**
[PASTE MESSAGE]

**Refine it applying the above principles.**

***
```

fdr:
(I had to edit the edit for technical correctness after the editing process, and I think this was due to subtle hair-splitting accuracy in my rough draft, so the LLM was not entirely wrong to make the mistake)

fdr (approved):

furkansahin (approved):

PR #4196, opened 2025-11-15
Admin panel style improvements

enescakir:
- **Do not wrap table headers in admin panels**
  IMHO, Wrapped headers make tables harder to read and scan quickly.
  

- **Add column width to lists in admin panel**
  The CSS `columns` property allows specifying column widths as the first
  parameter [^1]. This prevents columns from overlapping when the window
  is resized.
  
  [^1]: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/columns
  
| Before | After |
|--------|-------|
| CleanShot 2025-11-15 at 17 48 44@2x | CleanShot 2025-11-15 at 17 48 16@2x |

- **Wrap admin panel index with a container**
  Some components in the admin panel are already wrapped in a container,
  while others are not. To ensure consistent layout and styling, the
  entire admin panel index is now wrapped with a container component.
  
| Before | After |
|--------|-------|
| CleanShot 2025-11-15 at 17 48 37@2x | CleanShot 2025-11-15 at 17 48 07@2x |

- **Place inputs for object actions vertically in admin panel**

| Before | After |
|--------|-------|
| CleanShot 2025-11-15 at 18 00 29@2x | CleanShot 2025-11-15 at 18 00 16@2x |

  

----

> [!IMPORTANT]
> Improve admin panel UI by preventing table header wrapping, setting column widths, wrapping index in a container, and aligning inputs vertically.
> 
>   - **CSS Changes**:
>     - Prevent table headers from wrapping by adding `white-space: nowrap;` to `th` in `app.css`.
>     - Set column widths using `columns: 200px 5;` for `#model-class-list` and `columns: 200px 4;` for `#object-list` in `app.css`.
>     - Vertically align inputs in `#action-form` using `flex-direction: column;` in `app.css`.
>   - **HTML Changes**:
>     - Wrap admin panel index content in a `div.container` in `index.erb` for consistent layout.
>     - Change form ID from `action-button` to `action-form` in `object_action.erb` for consistency with CSS.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9ee2633f1ac8410f7764beb91052b596283df6a3. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/public/admin/app.css
+++ b/public/admin/app.css
@@ -138,6 +138,13 @@ code {
   }
 }
 
+#action-form {
+  display: flex;
+  flex-direction: column;
+  align-items: start;
+  gap: 10px;
+}
+
 .rodauth {
   div { margin-bottom: 10px; }
   label { min-width: 100px; }
--- a/views/admin/object_action.erb
+++ b/views/admin/object_action.erb
@@ -1,6 +1,6 @@
 <% @page_title = "#{@klass.name} #{@obj.ubid}" %>
 
-<% form(method: :post, id: "action-button") do |f| %>
+<% form(method: :post, id: "action-form") do |f| %>
   <% @params.keys.each do |name| %>
     <%== f.input("text", key: name, label: name, required: true) %>
   <% end %>


views/admin/index.erb
jeremyevans, line 1: Not wrapping the table was deliberate, as a wide table display can make it easier to visually parse, IMO. However, if you feel differently, I don't object to this change.

PR #4199, opened 2025-11-15
Monitor: only increment checkup when not already set

bsatzger:
Failed monitoring jobs incremented the checkup semaphore whenever the
resource responded to `incr_checkup`. In production we had a case where
a resource accumulated almost 150k checkup semaphores. In order to avoid
runaway accumulation of these semaphores, we now only increment the
checkup semaphore when the resource supports it and has not already set
a checkup .

----

> [!IMPORTANT]
> Modify `incr_checkup` call in `lib/monitor_resource_type.rb` to prevent incrementing when `checkup` is already set, with corresponding test added.
> 
>   - **Behavior**:
>     - In `lib/monitor_resource_type.rb`, modify `incr_checkup` call to check `!r.resource.checkup_set?` before incrementing.
>     - Prevents excessive semaphore increments when `checkup` is already set.
>   - **Tests**:
>     - Add test in `monitor_resource_type_spec.rb` to verify `incr_checkup` is not called if `checkup_set?` returns true.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 32408b54fa17700cebfc0ef9d8af7f535ca9dca7. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/monitor_resource_type.rb
+++ b/lib/monitor_resource_type.rb
@@ -41,7 +41,7 @@ MonitorResourceType = Struct.new(:wrapper_class, :resources, :types, :submit_que
             yield r
           rescue => ex
             Clog.emit("Monitoring job has failed.") { {monitoring_job_failure: {resource: r.resource, exception: Util.exception_to_hash(ex)}} }
-            r.resource.incr_checkup if r.resource.respond_to?(:incr_checkup)
+            r.resource.incr_checkup if r.resource.respond_to?(:incr_checkup) && !r.resource.checkup_set?
           ensure
             # We unset the started at time so we will not check for stuck pulses
             # while this is in the run queue.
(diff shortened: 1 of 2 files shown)


byucesoy (approved):

aamederen:
Would it be overkill to add this check to the `incr` method in `semaphore.rb`? Or, can we make (strand_id, name) a/the primary key and use insert_conflict to handle it with a single query?

bsatzger:
> Would it be overkill to add this check to the `incr` method in `semaphore.rb`? Or, can we make (strand_id, name) a/the primary key and use insert_conflict to handle it with a single query?

I thought about this, too. But the behavior was deliberately chosen (see https://github.com/ubicloud/ubicloud/commit/33a63ea8afe211914b1a260f99c2284a768cd11b). So I was opting for a surgical fix for now, while we discuss if we want to make more fundamental changes.

PR #4201, opened 2025-11-17
Do not create firewalls for subnets in general

fdr:
The default subnet included with a project defaults to all-open to ease the learning curve.

However, all subnets in general follow the same algorithm: if a firewall is not passed, it gins up a new firewall record with all-open rules.  This is dangerous.  As-is the only way to suppress this implied firewall creation is to get the create request on private subnet exactly right, and pass a firewall object from the get-go. Failing (or not knowing) this, you have to read back the subnet to go and delete the all-open firewall rule.  I, personally, totally forgot about how to suppress the default rules until reminded.

----

> [!IMPORTANT]
> Prevents automatic creation of all-open firewalls for subnets, requiring explicit firewall specification.
> 
>   - **Behavior**:
>     - Prevents automatic creation of all-open firewalls for subnets in `project.rb`, `kubernetes_cluster_nexus.rb`, `github_runner.rb`, and `vm_pool.rb`.
>     - Requires explicit firewall specification during subnet creation.
>   - **Functions**:
>     - Updates `assemble()` in `subnet_nexus.rb` to remove default firewall creation logic.
>     - Modifies `default_private_subnet()` in `project.rb` to handle firewall creation explicitly.
>   - **Tests**:
>     - Updates tests in `private_subnet_spec.rb`, `vpc_spec.rb`, `connected_subnets_spec.rb`, `subnet_nexus_spec.rb`, `attach-subnet_spec.rb`, `detach-subnet_spec.rb`, `golden_files_spec.rb`, `create_spec.rb`, and `private_subnet_spec.rb` to align with new firewall handling logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8f6b48f2c0f66fbe7e96516fdc57ddc6e56e7656. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/vm_pool.rb
+++ b/prog/vm/vm_pool.rb
@@ -36,10 +36,24 @@ class Prog::Vm::VmPool < Prog::Base
       encrypted: vm_pool.storage_encrypted,
       skip_sync: vm_pool.storage_skip_sync
     }
+    # Create a firewall with SSH-only access for VM pool
+    project = Project[Config.vm_pool_project_id]
+    location = Location[vm_pool.location_id]
+    firewall_name = "vm-pool-#{location.name}-firewall"
+    firewall = project.firewalls_dataset.first(location_id: vm_pool.location_id, name: firewall_name)
+    unless firewall
+      firewall = Firewall.create(name: firewall_name, location_id: vm_pool.location_id, project_id: Config.vm_pool_project_id)
+      DB.ignore_duplicate_queries do
+        ["0.0.0.0/0", "::/0"].each do |cidr|
+          FirewallRule.create(firewall_id: firewall.id, cidr: cidr, port_range: Sequel.pg_range(22..22))
+        end
+      end
+    end
+
     ps = Prog::Vnet::SubnetNexus.assemble(
       Config.vm_pool_project_id,
       location_id: vm_pool.location_id,
-      allow_only_ssh: true
+      firewall_id: firewall.id
     ).subject
 
     Prog::Vm::Nexus.assemble_with_sshable(
--- a/prog/vm/vm_pool.rb
+++ b/prog/vm/vm_pool.rb
@@ -42,12 +42,7 @@ class Prog::Vm::VmPool < Prog::Base
     firewall_name = "vm-pool-#{location.name}-firewall"
     firewall = project.firewalls_dataset.first(location_id: vm_pool.location_id, name: firewall_name)
     unless firewall
-      firewall = Firewall.create(name: firewall_name, location_id: vm_pool.location_id, project_id: Config.vm_pool_project_id)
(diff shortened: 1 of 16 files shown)


prog/vm/vm_pool.rb
```diff
@@ -36,10 +36,24 @@ def before_run
...
+    firewall_name = "vm-pool-#{location.name}-firewall"
+    firewall = project.firewalls_dataset.first(location_id: vm_pool.location_id, name: firewall_name)
+    unless firewall
+      firewall = Firewall.create(name: firewall_name, location_id: vm_pool.location_id, project_id: Config.vm_pool_project_id)
+      DB.ignore_duplicate_queries do
+        ["0.0.0.0/0", "::/0"].each do |cidr|
+          FirewallRule.create(firewall_id: firewall.id, cidr: cidr, port_range: Sequel.pg_range(22..22))
+        end
+      end
+    end
```
fdr, line 46: this is an iota of duplicated code that perhaps could be de-duplicated, if we wanted to introduce a method.

fdr:
image

Need a test for networking.erb:43, but not sure quite how to get it.

jeremyevans (commented):
There's more than an iota of duplicated code. :) Please add a `Firewall.create_with_open_rules(port_range, ...)` method that calls `create(...)` and adds the 2 firewall rules with the given port range to reduce duplication.

For the spec coverage issue, you need to test with a non-viewable firewall. This was previously handled by the "can show connected subnets" spec. Attach a firewall to the subnet before viewing the page and that should fix the issue. Would be nice to add a check that there is a link to the firewall with full permissions, but not with reduced permissions (there is already a similar check for connected subnet visibility in that spec).

fdr:
> There's more than an iota of duplicated code. :) Please add a `Firewall.create_with_open_rules(port_range, ...)` method that calls `create(...)` and adds the 2 firewall rules with the given port range to reduce duplication.
> 
> For the spec coverage issue, you need to test with a non-viewable firewall. This was previously handled by the "can show connected subnets" spec. Attach a firewall to the subnet before viewing the page and that should fix the issue. Would be nice to add a check that there is a link to the firewall with full permissions, but not with reduced permissions (there is already a similar check for connected subnet visibility in that spec).

see fixup

furkansahin:
let's make sure the tests pass and the conflicts are resolved

fdr:
I'm redraftifying a bunch of my PRs that I realize have kind of accumulated like plaque over the year, I'll pull them out as I process them and follow up.

PR #4214, opened 2025-11-18
Add update_stack helper

enescakir:
We run similar code in multiple places to update the stack. We have already similar helper in E2E tests. We can reuse it here.

### DIFF
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -296,6 +296,12 @@ end
     end
   end
 
+  def update_stack(new_frame)
+    strand.stack.first.merge!(new_frame)
+    strand.modified!(:stack)
+    strand.save_changes
+  end
+
   # A hop is a kind of jump, as in, like a jump instruction.
   private def dynamic_hop(label)
     fail "BUG: #hop only accepts a symbol" unless label.is_a? Symbol
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -686,9 +686,6 @@ SQL
   end
 
   def update_stack_lsn(lsn)
-    current_frame = strand.stack.first
-    current_frame["lsn"] = lsn
-    strand.modified!(:stack)
-    strand.save_changes
+    update_stack({"lsn" => lsn})
   end
 end
--- a/prog/redeliver_github_failures.rb
+++ b/prog/redeliver_github_failures.rb
@@ -2,12 +2,9 @@
 
 class Prog::RedeliverGithubFailures < Prog::Base
   label def wait
-    current_frame = strand.stack.first
-    last_check_time = Time.parse(current_frame["last_check_at"])
+    last_check_time = Time.parse(frame["last_check_at"])
     Github.redeliver_failed_deliveries(last_check_time)
-    current_frame["last_check_at"] = Time.now
-    strand.modified!(:stack)
-    strand.save_changes
+    update_stack({"last_check_at" => Time.now})
 
     nap 2 * 60
   end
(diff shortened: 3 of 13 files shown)


prog/base.rb
```diff
@@ -296,6 +296,12 @@ def reap(hop = nil, reaper: nil, nap: nil, fallthrough: false, strand: self.stra
     end
   end
 
+  def update_stack(new_frame, save: true)
```
jeremyevans: Instead of a keyword argument, it might be simpler to have a `modify_stack` that doesn't save, and an `update_stack` that calls `modify_stack` and then `save_changes`. However, if you prefer the keyword argument approach, it's fine with me.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -684,11 +684,4 @@ def available?
 
     false
   end
-
-  def update_stack_lsn(lsn)
```
jeremyevans, line 688: Might be simpler to keep this method and have it call `update_stack`, but I'm fine either way.

prog/vnet/subnet_nexus.rb
```diff
@@ -270,13 +270,6 @@ def nics_to_rekey
     nics_with_strand_label(%w[wait wait_setup]).all
   end
 
-  def update_stack_locked_nics(locked_nics)
```
jeremyevans, line 273: Similarly, could keep this method and have it call `update_stack`, as it also has multiple callers.

prog/redeliver_github_failures.rb
```diff
@@ -2,12 +2,9 @@
 
 class Prog::RedeliverGithubFailures < Prog::Base
   label def wait
-    current_frame = strand.stack.first
-    last_check_time = Time.parse(current_frame["last_check_at"])
+    last_check_time = Time.parse(strand.stack.first["last_check_at"])
```
jeremyevans: Could use `frame` here, though it's a bit less efficient.

```suggestion
    last_check_time = Time.parse(frame["last_check_at"])
```

prog/base.rb
enescakir: Actually, the only ones that we don't call `.save_changes` are `Prog::Kubernetes::UpgradeKubernetesNode` and `Prog::Kubernetes::ProvisionKubernetesNode`. I think we didn't call in them probably we already save when we run strand. So I think it's okay to run `.save_changes` for them too for consistency. What do you think @aamederen?

enescakir:
@jeremyevans, Since it's called with a single key most of the time, do you think it is worth to change it to `strand.stack.first.merge!(new_frame.transform_keys(&:to_s))` and refactor all callers to `update_stack(new_node_id: node_id)` from `update_stack({"new_node_id" => node_id})`. I think it has better readability.

jeremyevans:
> @jeremyevans, Since it's called with a single key most of the time, do you think it is worth to change it to `strand.stack.first.merge!(new_frame.transform_keys(&:to_s))` and refactor all callers to `update_stack(new_node_id: node_id)` from `update_stack({"new_node_id" => node_id})`. I think it has better readability.

I think it's fine. The database query overhead is much higher than the object allocation overhead.

enescakir:
> > @jeremyevans, Since it's called with a single key most of the time, do you think it is worth to change it to `strand.stack.first.merge!(new_frame.transform_keys(&:to_s))` and refactor all callers to `update_stack(new_node_id: node_id)` from `update_stack({"new_node_id" => node_id})`. I think it has better readability.
> 
> I think it's fine. The database query overhead is much higher than the object allocation overhead.

I tried to change it, but it required more updates than expected since many tests rely on mocks, so I'm not applying the string/symbol change in this PR.

prog/base.rb
aamederen: There's no special care needed for the K8s progs so go ahead.

PR #4216, opened 2025-11-19
Sort PostgreSQL resources alphabetically in list endpoint

byucesoy:
This is more user friendly especially when there are many resources.


jeremyevans (approved):

enescakir:
Is it for API or web? As I know, `paginated_result` helper has also default order column. It might be good to check which one has precedence

PR #4217, opened 2025-11-19
Delay host unavailability pages by 90 sec

bsatzger:
We’ve been sporadically receiving host unavailability alerts that auto-
resolve within seconds. To reduce noise while we make more fundamental
improvements to our monitoring, we extend the delay for unavailability
pages from 45 to 90 seconds.


furkansahin:
thank you

PR #4222, opened 2025-11-19
postgres_server_spec: less mocking

serprex:
removed test `returns true if read replica and the parent is nil` as read replica is false when parent is nil

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -169,7 +169,7 @@ class PostgresServer < Sequel::Model
       resource.representative_server
     end
 
-    lsn_diff(parent_server&.current_lsn || current_lsn, current_lsn) < 80 * 1024 * 1024
+    !parent_server || lsn_diff(parent_server.current_lsn, current_lsn) < 80 * 1024 * 1024
   end
 
   def current_lsn
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -272,6 +272,13 @@ RSpec.configure do |config|
       Vm.create(**args)
     end
 
+    def create_hosted_vm(project, private_subnet, name)
+      Prog::Vm::Nexus.assemble_with_sshable(
+        project.id, name:, private_subnet_id: private_subnet.id,
+        location_id: location.id, unix_user: "ubi"
+      ).subject
+    end
+
     def create_vm_from_size(size, arch, **args)
       vm_size = Validation.validate_vm_size(size, arch)
       args_from_size = {
--- a/spec/model/postgres/postgres_server_spec.rb
+++ b/spec/model/postgres/postgres_server_spec.rb
@@ -4,85 +4,116 @@ require_relative "../spec_helper"
 
 RSpec.describe PostgresServer do
   subject(:postgres_server) {
-    described_class.new { it.id = "c068cac7-ed45-82db-bf38-a003582b36ee" }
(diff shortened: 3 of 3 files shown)


jeremyevans:
Thank you very much for working on this. I'm currently at a conference, but I'll try to review this when I return on Friday.

spec/spec_helper.rb
```diff
@@ -272,6 +272,41 @@ def create_vm(**args)
       Vm.create(**args)
     end
 
+    def create_vm_etc(project, private_subnet, name)
+      vm_host = create_vm_host(used_cores: 2, total_hugepages_1g: 375, used_hugepages_1g: 16)
+      vm = Vm.create(
```
jeremyevans: This may be simpler using `Prog::Vm::Nexus.assemble_with_sshable`, but I didn't look at the details.

spec/spec_helper.rb
```diff
@@ -272,6 +272,41 @@ def create_vm(**args)
       Vm.create(**args)
     end
 
+    def create_vm_etc(project, private_subnet, name)
```
jeremyevans: This method name is a bit vague, though I realize in this case it is hard to get a good descriptive name that isn't long. Maybe `create_hosted_vm`?

jeremyevans (commented):
This looks good and I always like to see a reduction in unnecessary mocking. The CI failure is due to coverage. Let me know if you would like take care of that.  If not, I can look into it.

serprex:
Out until Tuesday, will try figuring out how coverage was lost then

spec/spec_helper.rb
serprex: seems easy to confuse with `create_vm_host` but makes sense

spec/spec_helper.rb
```diff
@@ -272,6 +272,14 @@ def create_vm(**args)
       Vm.create(**args)
     end
 
+    def create_hosted_vm(project, private_subnet, name)
+      st = Prog::Vm::Nexus.assemble_with_sshable(
+        project.id, name:, private_subnet_id: private_subnet.id,
+        location_id: location.id, unix_user: "ubi"
+      )
+      Vm[st.id]
+    end
```
enescakir, line 280: nit:

```suggestion
    def create_hosted_vm(project, private_subnet, name)
      Prog::Vm::Nexus.assemble_with_sshable(
        project.id, name:, private_subnet_id: private_subnet.id,
        location_id: location.id, unix_user: "ubi"
      ).subject
    end
```

model/postgres/postgres_server.rb
```diff
@@ -164,12 +164,12 @@ def needs_recycling?
 
   def lsn_caught_up
     parent_server = if read_replica?
-      resource.parent&.representative_server
+      resource.parent.representative_server
```
furkansahin: Read replica is not false when `parent` is `nil`. Read replica is false when `parent_id` is `nil`. Once we destroy a parent resource, the children aren't automatically promoted since this could cause unexpected behavioural change for the users of read replicas (e.g. accepting writes). Therefore, we expect those to be promoted manually by the customer. Here, that `&` makes this function basically return caught up in case the parent server is destroyed.

serprex: I see, so parent is also `nil` when `parent_id` is a dangling reference. Will add test case setting up that specific scenario

PR #4225, opened 2025-11-20
Redeliver failed GitHub webhook deliveries concurrently

enescakir:
- **Move failed deliveries logic to RedeliverGithubFailures prog**
  This helpers are only used in this prog, so it's better to keep them in
  that prog.
  

- **Remove unnecessary redeliver_failed_deliveries wrapper**
  

- **Redeliver failed GitHub webhook deliveries concurrently**
  GitHub can occasionally fail to deliver webhooks to us for various
  reasons (e.g., network issues, brief downtime). To avoid missing
  important events, we periodically check GitHub’s API for failed
  deliveries and try to redeliver them every 2 minutes.
  
  When GitHub or we have an incident, this process can’t finish within 120
  seconds and ends up triggering apoptosis.
  
  Since the “list deliveries” endpoint doesn’t support filtering, we have
  to paginate through all deliveries until the last check time to find the
  failed ones. There’s no way to make that part more concurrent.
  
  However, once we identify the failed deliveries, we can redeliver them
  concurrently. I refactored the code so that it buds child strands, each
  handling 50 failed deliveries.
  
  Right now it takes ~70s to redeliver 250 failed deliveries. Because the
  listing phase is constant, I expect the new design to drop the
  redelivery time to ~15s for the same workload (since delivering 50
  events currently takes that much).
  

- **Increase max page limit to 100 for fetching failed GitHub deliveries**
  GitHub’s app delivery API doesn’t provide good filtering options and
  only allows fetching up to 100 items per page. We added a max_page
  safeguard earlier to avoid unbounded pagination.
  
  As the total number of delivered webhook events has grown, we now need
  to raise the max_page limit to ensure we can fetch all failed deliveries
  within the desired time range.
  
  In my tests, fetching 200 pages takes about 40 seconds, and 100 pages
  takes about 20 seconds. With the recent concurrency improvements, this
  performance is acceptable.
  

----

> [!IMPORTANT]
> Redeliver failed GitHub webhook deliveries concurrently by refactoring logic into `prog/redeliver_github_failures.rb` and improving efficiency.
> 
>   - **Concurrency**:
>     - Redeliver failed GitHub webhook deliveries concurrently in `prog/redeliver_github_failures.rb` by handling 25 deliveries per child strand.
>     - Improves redelivery time from ~70s to ~15s for 250 deliveries.
>   - **Logic Relocation**:
>     - Move `failed_deliveries` logic from `lib/github.rb` to `prog/redeliver_github_failures.rb`.
>     - Remove `redeliver_failed_deliveries` method from `lib/github.rb`.
>   - **Pagination**:
>     - Increase max page limit to 100 in `prog/redeliver_github_failures.rb` for fetching failed deliveries.
>   - **Testing**:
>     - Update tests in `spec/prog/redeliver_github_failures_spec.rb` to cover new concurrent behavior and logic changes.
>     - Remove tests related to `failed_deliveries` and `redeliver_failed_deliveries` from `spec/lib/github_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 4867abb75ca15fe9229c5aaa97e1a92e5d72322c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/redeliver_github_failures.rb
+++ b/prog/redeliver_github_failures.rb
@@ -29,7 +29,7 @@ class Prog::RedeliverGithubFailures < Prog::Base
     @client ||= Github.app_client
   end
 
-  def failed_deliveries(since, max_page = 50)
+  def failed_deliveries(since, max_page = 100)
     all_deliveries = client.get("/app/hook/deliveries?per_page=100")
     page = 1
     while (next_url = client.last_response.rels[:next]&.href) && (since < all_deliveries.last[:delivered_at])


prog/redeliver_github_failures.rb
```diff
@@ -3,14 +3,23 @@
 class Prog::RedeliverGithubFailures < Prog::Base
   label def wait
     last_check_time = Time.parse(frame["last_check_at"])
-    failed = failed_deliveries(last_check_time).each do |delivery|
-      Clog.emit("redelivering failed delivery") { {delivery: delivery.to_h} }
-      client.post("/app/hook/deliveries/#{delivery[:id]}/attempts")
-    end.count
-    Clog.emit("redelivered failed deliveries") { {deliveries: {failed:}} }
-    update_stack({"last_check_at" => Time.now})
+    nap 10 if Time.now - last_check_time < 2 * 60
```
jeremyevans: Maybe we should calculate the number of seconds until the deadline, and nap that many plus 1? Up to you.

prog/redeliver_github_failures.rb
```diff
@@ -3,14 +3,23 @@
...
+  end
+
+  label def wait_redelivers
+    register_deadline("wait", 10 * 60)
+    reap(:wait)
+  end
 
-    nap 2 * 60
+  label def redeliver
+    frame["delivery_ids"].each { client.post("/app/hook/deliveries/#{it}/attempts") }
```
jeremyevans, line 24: With slower than expected GitHub response times, I think this could still lead to apoptosis, though it would require an average of ~2.4s response time per request. That seems unlikely enough that we probably don't need to worry about it. In the event it does become an issue, we could start checking each iteration how much lease time is remaining, and if it isn't sufficient, modify `frame["delivery_ids"]` and `nap 0`.

prog/redeliver_github_failures.rb
```diff
@@ -3,14 +3,23 @@
...
-    failed = failed_deliveries(last_check_time).each do |delivery|
-      Clog.emit("redelivering failed delivery") { {delivery: delivery.to_h} }
-      client.post("/app/hook/deliveries/#{delivery[:id]}/attempts")
-    end.count
-    Clog.emit("redelivered failed deliveries") { {deliveries: {failed:}} }
-    update_stack({"last_check_at" => Time.now})
+    nap 10 if Time.now - last_check_time < 2 * 60
+    failures = failed_deliveries(last_check_time)
+    failures.each_slice(50) do |deliveries|
+      bud Prog::RedeliverGithubFailures, {"delivery_ids" => deliveries.map { it[:id] }}, "redeliver"
```
jeremyevans, line 12: One thing to note about this approach is because this uses an `st` ubid, which is timestamp based and not fully random, all of these strands end up executing on the first respirate worker. Since each of these strands could take a while to execute, this could block other strands from running on the first respirate worker for a short period time.

I think the best way to address this is to make strand ubids not timestamp based, but that's best left to another pull request, and not a blocker for this one.

prog/redeliver_github_failures.rb
enescakir: Since we've set the timeout to 5 seconds, do you think redelivering in batches of 25 instead of 50 would be a safer approach?

https://github.com/ubicloud/ubicloud/blob/3bdf60d49f62defc266c18695ad5d5ed2b584fa8/lib/github.rb#L11

jeremyevans: It's a safer approach for avoiding apoptosis. The tradeoff is it doubles the number of strands, which makes it more likely tie up the first respirate worker. However, if we change `st` ubids to be fully random and not timestamp based (it least for certain `st` ubids if not for all), then I agree that 25 is better.

@pykello @fdr What are your thoughts on making `st` ubids be random instead of timestamp based? When originally designed, I don't think this was an issue, and using timestamp ubids provided basically free storage for the created at timestamp, but since the switch to partitioned respirate, it has become more of an issue. It doesn't just affect this prog type, but all `st` strands.

enescakir: Most of the time, the number of failures is below 25. But during a GitHub/Ubicloud incident, it can spike up to 300. In those cases, putting a heavy load on the first worker could be an issue.

PR #4226, opened 2025-11-20
fix: add billing check to pg similar to vm

serprex:
https://github.com/ubicloud/ubicloud/blob/3bdf60d49f62defc266c18695ad5d5ed2b584fa8/prog/vm/nexus.rb#L387

> [!IMPORTANT]
> Add billing check to `update_billing_records` in `postgres_resource_nexus.rb` to skip billing for non-billable projects, with corresponding tests.
> 
>   - **Behavior**:
>     - In `update_billing_records` in `postgres_resource_nexus.rb`, added a check to skip billing record updates if `postgres_resource.project.billable` is false.
>     - If the project is not billable, the function hops to "wait" instead of creating billing records.
>   - **Tests**:
>     - Added test in `postgres_resource_nexus_spec.rb` to verify that `update_billing_records` skips billing record creation for non-billable projects.
>     - Ensures that billing records are created when the project is billable.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 271f848d92aa10a130baf3a6eb0312f5993cd4fa. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_resource_nexus.rb
+++ b/prog/postgres/postgres_resource_nexus.rb
@@ -185,27 +185,29 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
   label def update_billing_records
     decr_update_billing_records
 
-    postgres_resource.active_billing_records.each(&:finalize)
+    if postgres_resource.project.billable
+      postgres_resource.active_billing_records.each(&:finalize)
 
-    flavor = postgres_resource.flavor
-    vm_family = representative_server.vm.family
-    vcpu_count = representative_server.vm.vcpus
-    storage_size_gib = representative_server.storage_size_gib
+      flavor = postgres_resource.flavor
+      vm_family = representative_server.vm.family
+      vcpu_count = representative_server.vm.vcpus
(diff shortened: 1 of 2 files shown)


prog/postgres/postgres_resource_nexus.rb
```diff
@@ -183,6 +183,7 @@ def before_run
   end
 
   label def update_billing_records
+    hop_wait unless postgres_resource.project.billable
```
jeremyevans: I looked into this more and I think we should run `decr_update_billing_records` and `decr_initial_provisioning` before hopping.

jeremyevans (changes requested):

prog/postgres/postgres_resource_nexus.rb
serprex: adjusted billing logic to happen within if block

PR #4227, opened 2025-11-20
Separate Vm Nexus files - Phase 2

furkansahin:
## Fix negative lookup caching for is_runner?

----

> [!IMPORTANT]
> Moves AWS-specific VM management to `prog/vm/aws/nexus.rb` with new methods for IAM role and instance profile management, and updates lifecycle management accordingly.
> 
>   - **Behavior**:
>     - Moves AWS-specific VM management from `instance.rb` to `prog/vm/aws/nexus.rb`.
>     - Adds methods for AWS IAM role and instance profile management in `prog/vm/aws/nexus.rb`.
>     - Updates VM lifecycle management to include AWS-specific steps like role creation and instance profile attachment.
>   - **Testing**:
>     - Adds comprehensive RSpec tests in `nexus_spec.rb` for new AWS-specific methods.
>     - Tests cover role creation, policy attachment, instance profile management, and error handling for AWS operations.
>   - **Misc**:
>     - Updates `prog/vm/nexus.rb` to handle AWS VM creation and destruction using the new methods.
>     - Introduces `migrate_to_separate_progs` in `vm.rb` to facilitate the transition to separate AWS handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 01dc41d324abd21daeb7889cc58e706bfce7e2d0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/aws/nexus.rb
+++ b/prog/vm/aws/nexus.rb
@@ -385,7 +385,8 @@ class Prog::Vm::Aws::Nexus < Prog::Base
   end
 
   def is_runner?
-    @is_runner ||= vm.unix_user == "runneradmin"
+    return @is_runner if defined?(@is_runner)
+    @is_runner = vm.unix_user == "runneradmin"
   end
 
   def ignore_invalid_entity


prog/vm/aws/nexus.rb
```diff
@@ -16,14 +15,221 @@ def before_run
...
+            iops: 3000,
+            volume_size: vm.vm_storage_volumes_dataset.where(:boot).get(:size_gib),
+            volume_type: "gp3",
+            throughput: 125
+          }
+        }
+      ],
+      network_interfaces: [
+        {
+          network_interface_id: vm.nics.first.nic_aws_resource.network_interface_id,
```
jeremyevans: I realize this is just copying existing code, but in case you want to simplify it:

```suggestion
          network_interface_id: vm.nic.nic_aws_resource.network_interface_id,
```

Similar changes can be made in other places. Up to you.

prog/vm/aws/nexus.rb
```diff
@@ -109,11 +328,74 @@ def before_run
     end
   end
 
+  label def cleanup_roles
+    ignore_invalid_entity do
```
jeremyevans, line 332: It seems a little inconsistent that during setup, we use a separate label per action, but during cleanup, we do it all in the same label. It's not a problem, though, and maybe there is a reason for it. Anyway, this is just copying the existing implementation, so it's not a problem to keep it as-is.

prog/vm/aws/nexus.rb
```diff
@@ -109,11 +328,74 @@ def before_run
...
+  def role_name
+    vm.name
+  end
+
+  def instance_profile_name
+    "#{vm.name}-instance-profile"
+  end
+
+  def is_runner?
+    @is_runner ||= vm.unix_user == "runneradmin"
```
jeremyevans: You cannot use `||=` for caching when `false` is a valid value. You need to use `defined?`:

```suggestion
    return @is_runner if defined?(@is_runner)
    @is_runner = vm.unix_user == "runneradmin"
```

This was a bug in the existing code. It doesn't cause a problem, but it doesn't cache a negative lookup. It should be fixed, but it's up to you whether to do it in this PR or in a later one.

prog/vm/aws/nexus.rb
furkansahin: thank you!

prog/vm/aws/nexus.rb
furkansahin: I'll keep this as is for now but you're right.

prog/vm/aws/nexus.rb
furkansahin: Makes sense, updated it accordingly. Thank you!

PR #4228, opened 2025-11-20
Prevent deletion of GitHub installations for suspended accounts

enescakir:
When we suspend an account, the user can’t log in to the Ubicloud console, which prevents them from removing the GitHub installation. This helps block fraudulent users from repeatedly reusing the same GitHub organization.

However, if they remove the installation from the GitHub UI, we receive a webhook event and currently delete the installation on our side. This allows them to reuse the same GitHub organization with a new account.

To prevent this, if the installation’s project is inactive, we no longer delete the installation when handling the webhook event.

----

> [!IMPORTANT]
> Prevent deletion of GitHub installations for inactive projects when handling deletion events in `github.rb`.
> 
>   - **Behavior**:
>     - In `handle_installation` in `github.rb`, prevent deletion of GitHub installation if the project is inactive when receiving a "deleted" action.
>   - **Tests**:
>     - Add test in `github_spec.rb` to verify error response when project is inactive during installation deletion event.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for dfcc08d29986511d7d70dfe37df1468a8879adae. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/routes/webhook/github.rb
+++ b/routes/webhook/github.rb
@@ -40,9 +40,8 @@ class Clover
     installation = GithubInstallation[installation_id: data["installation"]["id"]]
     case data["action"]
     when "deleted"
-      unless installation
-        return error("Unregistered installation")
-      end
+      return error("Unregistered installation") unless installation
+      return error("Inactive project") unless installation.project.active?
 
       Prog::Github::DestroyGithubInstallation.assemble(installation)
       return success("GithubInstallation[#{installation.ubid}] deleted")
--- a/spec/routes/web/webhook/github_spec.rb
+++ b/spec/routes/web/webhook/github_spec.rb
@@ -47,6 +47,15 @@ RSpec.describe Clover, "github" do
       expect(page.body).to eq({error: {message: "Unregistered installation"}}.to_json)
     end
 
+    it "fails if the project is inactive" do
+      account = create_account.update(suspended_at: Time.now)
+      account.add_project(installation.project)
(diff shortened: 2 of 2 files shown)


routes/webhook/github.rb
```diff
@@ -40,9 +40,8 @@ def handle_installation(data)
     installation = GithubInstallation[installation_id: data["installation"]["id"]]
     case data["action"]
     when "deleted"
-      unless installation
-        return error("Unregistered installation")
-      end
+      return error("Unregistered installation") unless installation
+      return error("Inactive project") unless installation.project.active?
```
jeremyevans, line 44: This will return an error both for non-`visible` projects as well as projects with suspended accounts. If we want to do this only for projects with suspended accounts, I would add `Project#suspended?` and call that instead of `active?`. However, if you are OK doing it for both, this seems fine.

enescakir: We update `project.visible` when we soft delete a project, and we also destroy all installations during the soft delete. I think it's fine to check for both.

https://github.com/ubicloud/ubicloud/blob/66b717cc8c7a1ed72c2c7bceb1bd4bccb90fb194/model/project.rb#L96-L114

PR #4229, opened 2025-11-21
Decrease mocking at GithubRepositoryNexus specs

enescakir:
- **Decrease mocking at GithubRepositoryNexus specs**
  

- ~**Destroy repository and its runners when project inactivated**
  When we suspend an account, it's project being inactive. So it's good to
  force delete its runners and repository.~

### DIFF
--- a/spec/prog/github/github_repository_nexus_spec.rb
+++ b/spec/prog/github/github_repository_nexus_spec.rb
@@ -5,52 +5,44 @@ require "octokit"
 
 RSpec.describe Prog::Github::GithubRepositoryNexus do
   subject(:nx) {
-    st = Strand.create_with_id(github_repository, prog: "Github::GithubRepositoryNexus", label: "wait")
+    st = Strand.create_with_id(repository, prog: "Github::GithubRepositoryNexus", label: "wait")
     described_class.new(st)
   }
 
   let(:project) { Project.create(name: "test") }
   let(:installation) { GithubInstallation.create(installation_id: 123, project_id: project.id, name: "test-user", type: "User") }
-  let(:github_repository) {
+  let(:repository) {
     GithubRepository.create(name: "ubicloud/ubicloud", last_job_at: Time.now, installation_id: installation.id)
   }
 
+  let(:now) { Time.now.round }
+  let(:client) { instance_double(Octokit::Client) }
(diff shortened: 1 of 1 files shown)


prog/github/github_repository_nexus.rb
```diff
@@ -136,6 +132,14 @@ def before_run
   end
 
   label def wait
+    unless github_repository.installation.project.active?
```
jeremyevans: With the `nap 0`, I think we could and probably would run this block more than once. It could result in multiple semaphores accruing on the runners and repository. Potentially many semaphores if the respirate worker handling the repository is busy.

I think the safest way to handle things is to set a frame variable inside this block. If we reenter the block, `nap 5` or something similar instead of adding more semaphores.

enescakir: Since we run `github_repository.incr_destroy` just before, doesn't the first `nap 0` ends up to hop to `destroy` label?

jeremyevans: I think you are correct, but is there a reason to use `incr_destroy` and `nap 0` instead of using `hop_destroy`? `hop_destroy` makes it more obvious, IMO. I guess that doesn't register the deadline, though. If the deadline is the reason you are using `incr_destroy` and `nap 0`, I think it would be more understandable and more efficient to use `register_deadline` and `hop_destroy`.

enescakir: In the codebase, we try to avoid calling `hop_destroy` outside of `before_run` because it bypasses the `before_run` logic. Right now, `before_run` only has the `register_deadline`, but if we add more logic in the future, it’s easy to forget updating this path as well. So it’s safer to always go through `before_run` in all cases.

jeremyevans: OK. If that is the current practice, this is fine.

FWIW, I think it would be better to add a `start_destroy` method, have `before_run` call that, and allow calling `start_destroy` directly from other labels. That keeps the logic in one place, and makes the calling code more understandable.

enescakir:
I decided to handle the destruction of resources for inactive projects in the suspension method rather than in individual resources. So I'm merging only the demock part for now. I'll create another PR for resource destruction.

PR #4232, opened 2025-11-21
Use forwardable in the GithubRunner prog for easier access to associations

enescakir:
- **Use forwardable in the GithubRunner prog for easier access to associations**
  I didn't know about the 'forwardable' delegator when I first wrote this
  prog. I learned about it from PostgresResourceNexus.
  
  It eliminates the need to write certain property accessors, and reduces
  line width significantly, and enhances code readability.
  

- **Add runners_path helper to GithubRunner prog**
  

----

> [!IMPORTANT]
> Use `forwardable` in `GithubRunner` for method delegation and add `runners_path` for cleaner API path construction.
> 
>   - **Enhancements**:
>     - Use `forwardable` in `GithubRunner` class to delegate methods: `installation`, `vm`, `label_data`, and `project`.
>     - Add `runners_path` method to construct API paths for GitHub runners.
>   - **Code Simplification**:
>     - Replace explicit property accessors with `def_delegators` for `github_runner` and `installation`.
>     - Use `runners_path` in `busy?`, `register_runner`, and `destroy` methods for cleaner API calls.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a0564cd7bf1ead14104ced794dfb7871e54d581e. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -152,12 +152,16 @@ class Prog::Vm::GithubRunner < Prog::Base
   end
 
   def busy?
-    github_client.get("/repos/#{github_runner.repository_name}/actions/runners/#{github_runner.runner_id}")[:busy]
+    github_client.get(runners_path(github_runner.runner_id))[:busy]
   rescue Octokit::NotFound
   end
 
   def x64?
     label_data["arch"] == "x64"
+  end
+
+  def runners_path(suffix = nil)
+    "/repos/#{github_runner.repository_name}/actions/runners#{"/#{suffix}" if suffix}"
   end
 
   def before_run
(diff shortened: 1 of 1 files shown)


prog/vm/github_runner.rb
```diff
@@ -145,21 +147,21 @@ def update_billing_record
     end
   end
 
-  def vm
-    @vm ||= github_runner.vm
```
jeremyevans, line 149: The `def_delegators` approach is slightly slower as it does no caching, but I don't think the performance differences will be significant.

enescakir: Since we delegate `vm` to `github_runner.vm`, doesn’t it get cached on `github_runner.vm` instead of `@vm`? If I'm not missing anything, I tried it, and it only runs a DB query the first time, then uses the cached value directly.

jeremyevans: There is caching in the Sequel association, but it requires more indirection so it will be slower. However, as I mentioned, I don't think the performance differences will be significant, so it's fine to use this approach.

PR #4236, opened 2025-11-21
Allow GPU runners with feature flag

enescakir:
- **Show deprecation warning for GPU runners**
  

- **Allow GPU runners with feature flag**
  We decided to deprecate GPU runners, but existing users should be able
  to continue using them for a while. This change restricts GPU runner
  usage to accounts that have the appropriate feature flag, ensuring only
  current users can access them going forward.
  

----

> [!IMPORTANT]
> Restricts GPU runner usage to accounts with a feature flag and adds a deprecation warning for GPU runners.
> 
>   - **Behavior**:
>     - Restricts GPU runner usage to accounts with `gpu_runner` feature flag in `prog/vm/github_runner.rb`.
>     - Adds deprecation warning for GPU runners in `setup_environment()` in `prog/vm/github_runner.rb`.
>   - **Feature Flags**:
>     - Adds `gpu_runner` feature flag to `Project` model in `model/project.rb`.
>   - **Tests**:
>     - Adds test for GPU runner feature flag in `github_runner_spec.rb`.
>     - Adds test for GPU deprecation warning in `github_runner_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 06c7330e5dfa5f1335724353f4db50531659e6c6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/github_runner.rb
+++ b/prog/vm/github_runner.rb
@@ -177,6 +177,10 @@ class Prog::Vm::GithubRunner < Prog::Base
       github_runner.destroy
       pop "Could not provision a runner for inactive project"
     end
+    if github_runner.label.include?("gpu") && !github_runner.installation.project.get_ff_gpu_runner
+      github_runner.destroy
+      pop "Could not provision a GPU runner for this project"
+    end
     hop_wait_concurrency_limit unless quota_available?
     hop_apply_custom_label_quota if github_runner.custom_label
     hop_allocate_vm
--- a/model/project.rb
+++ b/model/project.rb
@@ -198,6 +198,7 @@ class Project < Sequel::Model
     :enable_m6id,
     :enable_m6gd,
     :free_runner_upgrade_until,
+    :gpu_runner,
     :gpu_vm,
     :ipv6_disabled,
     :postgres_hostname_override,
--- a/spec/prog/vm/github_runner_spec.rb
+++ b/spec/prog/vm/github_runner_spec.rb
@@ -328,6 +328,12 @@ RSpec.describe Prog::Vm::GithubRunner do
       expect { nx.start }.to exit({"msg" => "Could not provision a runner for inactive project"})
       expect(GithubRunner[runner.id]).to be_nil
     end
+
+    it "pops if GPU runners not enabled for the project" do
+      runner.label = "ubicloud-gpu"
+      expect { nx.start }.to exit({"msg" => "Could not provision a GPU runner for this project"})
+      expect(GithubRunner[runner.id]).to be_nil
+    end
   end
 
   describe "#wait_concurrency_limit" do


config/github_runner_labels.yml
```diff
@@ -51,10 +51,10 @@
 - { name: ubicloud-standard-16-arm-ubuntu-2204, family: standard,     vcpus: 16,    arch: arm64, storage_size_gib: 300, boot_image: github-ubuntu-2204 }
 - { name: ubicloud-standard-30-arm-ubuntu-2204, family: standard,     vcpus: 30,    arch: arm64, storage_size_gib: 400, boot_image: github-ubuntu-2204 }
 - { name: ubicloud-standard-60-arm-ubuntu-2204, family: standard,     vcpus: 60,    arch: arm64, storage_size_gib: 600, boot_image: github-ubuntu-2204 }
-- { name: ubicloud-gpu,                         family: standard-gpu, vcpus: 6,     arch: x64,   storage_size_gib: 180, boot_image: github-gpu-ubuntu-2204 }
```
jeremyevans, line 54: No objections to this, but is there a reason to move the line?

prog/vm/github_runner.rb
```diff
@@ -327,7 +327,12 @@ def setup_info
       COMMAND
     end
 
-    if !github_runner.label.include?("ubuntu")
+    if github_runner.label.include?("gpu")
+      message = "The GPU runners will be deprecated on December 31, 2025. All jobs using these runners should be migrated to other runner types."
```
jeremyevans, line 335: How visible is this error message to the user? Does GitHub alert the user about the error, is it visible on the main summary page, or is it only shown when looking at job details/logs?

prog/vm/github_runner.rb
```diff
@@ -177,6 +177,10 @@ def before_run
       github_runner.destroy
       pop "Could not provision a runner for inactive project"
     end
+    if github_runner.label.include?("gpu") && !github_runner.installation.project.get_ff_gpu_runner
```
jeremyevans, line 180: This may be better moved to `assemble`, so we aren't checking it during every strand run. The only reason I can think of to check it here instead of `assemble` would be to catch cases where the `get_ff_gpu_runner` flag was changed after `assemble` was called and the github runner was created, and I don't think we need to worry about that case. However, I'll leave that decision up to you.

config/github_runner_labels.yml
enescakir: We keep deprecated labels at the bottom of the file.
https://github.com/ubicloud/ubicloud/blob/66b717cc8c7a1ed72c2c7bceb1bd4bccb90fb194/config/github_runner_labels.yml#L55-L60

prog/vm/github_runner.rb
enescakir: It's a warning on the job summary page
https://github.com/enescakir/ubicloud-test-dev/actions/runs/19669452635
CleanShot 2025-11-25 at 12 27 13@2x

prog/vm/github_runner.rb
enescakir: When I move this logic into assemble, it could fail or return nil if the condition isn't met, but both options would require changing the code that calls assemble. Since we plan to remove this in a month, I'm trying to avoid touching too many places.

Also, I though it's better to enforce the check in the strand rather than in the web endpoint, since the latter is more synchronous.

PR #4239, opened 2025-11-21
Add dependabot cooldowns

fdr:
I always thought: "that would be a nice feature to have" to mitigate supply chain attacks before they can get noticed,
https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns made me aware that this feature was at hand.


enescakir (approved):
Great 🚀

PR #4248, opened 2025-11-24
Heuristic overcommit for PG instances

shikharbhardwaj:
* Heuristic overcommit for PG instances

    This commit adds a heuristic to calculate the overcommit value for PG
    for PG instances.
    
    Based on observations from allocation failure related pages, we setup a
    y = mx + c type value for overcommit_kbytes, which gives a fixed
    overhead of 256Mi (for an extra bump to the smaller SKUs) and a factor
    of 75% over the actual available pages that can be charged.

* Fix duplicate query in postgres_timeline.rb

  When running tests with both minio_service_project_id and
  postgres_service_project_id set, the tests fail with a duplicate query
  error due to the same query repeating with different parameters.

* Unify PG hugepages and overcommit configuration

  This commit moves the hugepages and overcommit configuration for PG
  instances into a single place which manages the virtual memory
  configuration. In the next commit, we will add a more sophisticated
  heuristic to control the overcommit value

----

> [!IMPORTANT]
> Introduces heuristic overcommit for PostgreSQL, unifies memory configuration, and fixes a duplicate query issue.
> 
>   - **Heuristic Overcommit**:
>     - Adds heuristic for `overcommit_kbytes` in `VirtualMemorySetup` to set 256 MiB + 1.25 * (MemTotal - HugePages).
>     - Removes `configure_memory_overcommit` from `PostgresSetup`.
>   - **Configuration Unification**:
>     - Renames `configure-hugepages` to `configure-vmem`.
>     - Renames `HugepagesSetup` to `VirtualMemorySetup`.
>     - Updates `configure` to use `configure-vmem` for virtual memory settings.
>   - **Bug Fix**:
>     - Fixes duplicate query in `postgres_timeline.rb` by combining `project_id` conditions.
>   - **Testing**:
>     - Updates tests in `virtual_memory_setup_spec.rb` to reflect new overcommit logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 963a08b3168df9f365f6b8f583df5075c5fce48b. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/postgres/lib/virtual_memory_setup.rb
+++ b/rhizome/postgres/lib/virtual_memory_setup.rb
@@ -68,8 +68,16 @@ CONF
     r "sudo sysctl -w vm.overcommit_memory=2"
     r "echo 'vm.overcommit_memory=2' | sudo tee -a /etc/sysctl.conf"
 
-    r "sudo sysctl -w vm.overcommit_ratio=150"
-    r "echo 'vm.overcommit_ratio=150' | sudo tee -a /etc/sysctl.conf"
+    meminfo = File.read("/proc/meminfo")
+    mem_total_kb = Integer(meminfo[/^MemTotal:\s*(\d+)/, 1], 10)
+    hugepages_total, hugepage_size_kb = hugepage_info
+    hugepages_kb = hugepages_total * hugepage_size_kb
+
+    # Calculate overcommit_kbytes as 256 MiB + 1.75 * (MemTotal - HugePages)
+    overcommit_kbytes = (256 * 1024) + (1.75 * (mem_total_kb - hugepages_kb)).round
+
+    r "sudo sysctl -w vm.overcommit_kbytes=#{overcommit_kbytes}"
+    r "echo 'vm.overcommit_kbytes=#{overcommit_kbytes}' | sudo tee -a /etc/sysctl.conf"
   end
 
   def setup
--- a/rhizome/postgres/spec/virtual_memory_setup_spec.rb
+++ b/rhizome/postgres/spec/virtual_memory_setup_spec.rb
@@ -44,4 +44,29 @@ RSpec.describe VirtualMemorySetup do
       expect { virtual_memory_setup.setup_postgres_hugepages }.not_to raise_error
     end
   end
+
+  describe "#configure_memory_overcommit" do
+    before do
+      allow(File).to receive(:read).with("/proc/meminfo").and_return(<<~MEMINFO)
+        MemTotal:        8000000 kB
+        HugePages_Total:     512
+        Hugepagesize:     2048 kB
+        MemFree:         4000000 kB
+      MEMINFO
+    end
+
+    it "calculates and sets overcommit_kbytes correctly" do
+      # MemTotal: 8,000,000 kB
+      # HugePages: 512 * 2048 kB = 1,048,576 kB
+      # Calculation: 256 MiB (262144 kB) + 1.75 * (8,000,000 - 1,048,576) = 12427136 kB
+      expected_kbytes = 12_427_136
+
+      expect(virtual_memory_setup).to receive(:r).with("sudo sysctl -w vm.overcommit_memory=2")
+      expect(virtual_memory_setup).to receive(:r).with("echo 'vm.overcommit_memory=2' | sudo tee -a /etc/sysctl.conf")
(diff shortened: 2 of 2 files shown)


model/postgres/postgres_timeline.rb
```diff
@@ -100,7 +100,7 @@ def blob_storage
     @blob_storage ||= if aws?
       S3BlobStorage.new("https://s3.#{location.name}.amazonaws.com")
     else
-      MinioCluster.first(project_id: Config.postgres_service_project_id, location_id: location.id) || MinioCluster.first(project_id: Config.minio_service_project_id, location_id: location.id)
+      MinioCluster.first(project_id: [Config.postgres_service_project_id, Config.minio_service_project_id], location_id: location.id)
```
jeremyevans: This is incorrect. If you have a MinioCluster in both projects with the location, with this change, it could use the one in minio_service_project_id instead of the one in postgres_service_project_id, when the the one in postgres_service_project_id should be preferred. If you want to use this approach, you must use an order that ensures the postgres_service_project_id one is first.

shikharbhardwaj: Ah yes, that makes sense. Although we should have a spec for this preference somewhere. I think making the query that way is more complicated than it needs to be in this case.

Is there a way to mark this one as okay for the duplicate query check?

jeremyevans: Yes. Wrap in a `DB.ignore_duplicate_queries` block.

shikharbhardwaj: I folded this and another fix into https://github.com/ubicloud/ubicloud/pull/4262

byucesoy:
PR looks fine. I am just not sure if we should allow more overcommit given the latest pages we see even with overcommit_ratio=150

shikharbhardwaj:
@byucesoy I increased the ratio to 175% based on these observations. But I also think that in some of these cases we need to understand the memory usage better.

byucesoy:
@shikharbhardwaj, I approved the PR. I think we can merge & deploy this. I agree on the need for understanding the cases where we see even 200 overcommit_ratio not being enough. Can you take the ownership of getting bottom of those cases and possibly coming up with next version of memory management?

byucesoy:
Let's hold off on this for a while until we understand more on the memory usage. I'll switch all fleet to kernel heuristic for overcommit (memory_overcommit: 0)

fdr:
@shikharbhardwaj if you have copious free time (maybe you don't), did you ever profile what that outlier instance with the crazy amount of virtual memory was all about? it takes some care (but LLMs with information retrieval, e.g. perplexity can help) to subtract out, like, hugepages and ascribe the commit charge to each process, IIRC it's not quite a direct thing, alas

fdr:
In the end, @byucesoy  found it, per his quote from slack:

> Reported [here](https://lists.openwall.net/linux-kernel/2023/09/13/13), bisected [here](https://lists.openwall.net/linux-kernel/2023/09/15/1178) and fixed [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=3cec50490969afd4a76ccee441f747d869ccff77) by no other than Linus himself.

```
vm: fix move_vma() memory accounting being off
Commit 408579cd627a ("mm: Update do_vmi_align_munmap() return
semantics") seems to have updated one of the callers of do_vmi_munmap()
incorrectly: it used to check for the error case (which didn't
change: negative means error).

That commit changed the check to the success case (which did change:
before that commit, 0 was success, and 1 was "success and lock
downgraded".  After the change, it's always 0 for success, and the lock
will have been released if requested).

This didn't change any actual VM behavior _except_ for memory accounting
when 'VM_ACCOUNT' was set on the vma.  Which made the wrong return value
test fairly subtle, since everything continues to work.

Or rather - it continues to work but the "Committed memory" accounting
goes all wonky (Committed_AS value in /proc/meminfo), and depending on
settings that then causes problems much much later as the VM relies on
bogus statistics for its heuristics.

Revert that one line of the change back to the original logic.
```

We began to suspect it when the numbers just weren't adding up and we had claude cook up a statistical regression on kernel version.

fdr:
I think(?) this is obsolete given the kernel finding and 963c115dc86d96a319bf36bd9ccdbd7fa2f1002c. Or aspects that are to be refurbished can be revived separately.

PR #4251, opened 2025-11-24
Ensure destroy flow runs when AWS runners fail to acquire instance

aamederen:
Previously, when an AWS VM Nexus failed due to insufficient instance
capacity, it would `pop` itself directly. This left behind a VM
model without an associated strand, causing the owning GithubRunner
flow to hang and cause page. By `incr_destroy`ing the VM,
we aim proper handling of cleanup and preventing stuck runners.

----

> [!IMPORTANT]
> Ensure AWS VM Nexus lets GithubRunner handle cleanup on insufficient capacity, preventing stuck runners.
> 
>   - **Behavior**:
>     - In `nexus.rb`, when AWS VM Nexus encounters `InsufficientInstanceCapacity`, it no longer pops itself. Instead, it increments `vm.incr_destroy` and naps, allowing the owning `GithubRunner` to handle cleanup.
>     - This prevents the creation of VM models without associated strands, which previously caused the `GithubRunner` flow to hang.
>   - **Tests**:
>     - In `nexus_spec.rb`, updated tests to expect `nap(0)` instead of `exit` when `create_instance` encounters `InsufficientInstanceCapacity`.
>     - Tests cover scenarios with and without `alternative_families` set, ensuring proper handling of runner recreation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for fc8ead27816f516af1e1e0698a75c9d7ee27d65f. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/aws/nexus.rb
+++ b/prog/vm/aws/nexus.rb
@@ -201,7 +201,8 @@ class Prog::Vm::Aws::Nexus < Prog::Base
         end
         runner.provision_spare_runner
         runner.incr_destroy
-        pop "exiting due to insufficient instance capacity"
+        vm.incr_destroy
+        nap 0
       end
       raise
     end
--- a/spec/prog/vm/aws/nexus_spec.rb
+++ b/spec/prog/vm/aws/nexus_spec.rb
@@ -407,13 +407,13 @@ usermod -L ubuntu
       it "recreates runner when alternative_families is not set" do
         expect(nx).to receive(:frame).and_return({}).at_least(:once)
         expect(Prog::Vm::GithubRunner).to receive(:assemble).and_call_original
-        expect { nx.create_instance }.to exit({"msg" => "exiting due to insufficient instance capacity"})
+        expect { nx.create_instance }.to nap(0)
       end
 
       it "recreates runner when alternative_families is empty" do
         expect(nx).to receive(:frame).and_return({"alternative_families" => []}).at_least(:once)
         expect(Prog::Vm::GithubRunner).to receive(:assemble).and_call_original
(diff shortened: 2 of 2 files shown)


prog/vm/aws/nexus.rb
```diff
@@ -201,7 +201,7 @@ def before_run
         end
         runner.provision_spare_runner
         runner.incr_destroy
-        pop "exiting due to insufficient instance capacity"
+        nap 30
```
enescakir: Instead of waiting for the runner to destroy the vm, we can trigger the destroy immediately.


```suggestion
        vm.incr_destroy
        nap 0
```

furkansahin: or we could simply hop_destroy?

enescakir: In the codebase, we try to avoid calling `hop_destroy` outside of `before_run` because it bypasses the `before_run` logic. Right now, `before_run` only has the deadline registering and billing record finalizing, but if we add more logic in the future, it’s easy to forget updating this path as well. So it’s safer to always go through `before_run` in all cases.

aamederen: We have a few alternative ways to inject cleanup:
1\ Don't do anything and let `runner` destroy logic destroy the VM
2\ Trigger destroy of the VM via `vm.incr_destroy`
3\ `hop_destroy` directly
4\ call `final_clean_up` directly

IMO, (1) was the most comprehensive way, but after talking with Enes, (2) seems to bring some speed-up without much risk of future misses, so I changed from (1) to (2).

PR #4255, opened 2025-11-25
Add PostgresResource list by tag endpoint

furkansahin:
> [!IMPORTANT]
> Add endpoint to filter Postgres resources by tags, update OpenAPI spec, and add tests for validation and filtering logic.
> 
>   - **Behavior**:
>     - Adds `tags_param` to `postgres_list()` in `postgres.rb` to filter Postgres resources by tags.
>     - Validates tag format and filters resources using `Sequel.pg_jsonb_op(:tags).contains(tags_param)`.
>     - Updates OpenAPI spec in `openapi.yml` to include `tags` parameter for `/project/{project_id}/postgres` endpoint.
>   - **Routes**:
>     - Updates `routes/project/postgres.rb` to pass `tags_param` to `postgres_list()`.
>   - **Tests**:
>     - Adds tests in `postgres_spec.rb` for filtering by single and multiple tags, empty tags, and invalid tag formats.
>     - Tests ensure correct filtering logic and error handling for invalid formats.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for acfd7c142617a2b82c32f1cf94b534b82fa048e8. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/helpers/postgres.rb
+++ b/helpers/postgres.rb
@@ -56,8 +56,21 @@ class Clover
     end
   end
 
-  def postgres_list
+  def postgres_list(tags_param: nil)
     dataset = dataset_authorize(@project.postgres_resources_dataset.eager(:timeline, representative_server: [:strand, vm: :vm_storage_volumes]), "Postgres:view").eager(:semaphores, :location, strand: :children)
+
+    if tags_param
+      tags_param = tags_param.split(",")
+      tags_param = tags_param.map! { |tag| tag.split(":", 2).map(&:strip) }
+      tags_param = tags_param.map! { |key, value| {key:, value:} }
+      tags_param.each do |tag|
+        unless tag[:value]
+          fail Validation::ValidationFailed.new({tags: "Invalid tag format. Expected format: key:value"})
+        end
+      end
+      dataset = dataset.where(Sequel.pg_jsonb_op(:tags).contains(tags_param))
+    end
+
     if api?
       dataset = dataset.where(location_id: @location.id) if @location
       paginated_result(dataset, Serializers::Postgres)
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1757,6 +1757,7 @@ paths:
       summary: List visible Postgres Databases
       parameters:
         - $ref: '#/components/parameters/project_id'
+        - $ref: '#/components/parameters/tags'
         - $ref: '#/components/parameters/start_after'
         - $ref: '#/components/parameters/page_size'
         - $ref: '#/components/parameters/order_column'
@@ -2187,6 +2188,14 @@ components:
       required: false
       schema:
         type: string
+    tags:
+      description: 'Filter by tags in key:value format (comma-separated for multiple tags)'
+      in: query
+      name: tags
+      required: false
+      schema:
+        type: string
+        example: 'environment:production,team:backend'
     vm_reference:
       description: Virtual machine ID or name
       examples:
--- a/routes/project/postgres.rb
+++ b/routes/project/postgres.rb
@@ -3,7 +3,8 @@
 class Clover
   hash_branch(:project_prefix, "postgres") do |r|
     r.get true do
-      postgres_list
+      tags_param = typecast_params.nonempty_str("tags")
+      postgres_list(tags_param:)
     end
 
     r.web do
--- a/spec/routes/api/project/postgres_spec.rb
+++ b/spec/routes/api/project/postgres_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe Clover, "vm" do
       allow(Config).to receive(:postgres_service_project_id).and_return(postgres_project.id)
     end
 
-    it "success all vms" do
+    it "success all postgres resources" do
       Prog::Postgres::PostgresResourceNexus.assemble(
         project_id: project.id,
         location_id: Location::HETZNER_FSN1_ID,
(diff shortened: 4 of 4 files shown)


helpers/postgres.rb
```diff
@@ -56,8 +56,19 @@ def postgres_post(name)
     end
   end
 
-  def postgres_list
+  def postgres_list(tag_key: nil, tag_value: nil)
```
byucesoy: I think we should accept a list of key/values, so that people can filter by multiple tags. We can accept a single `tags` parameter, which would be an array of key/values. What do you think?

helpers/postgres.rb
```diff
@@ -56,8 +56,19 @@ def postgres_post(name)
...
     dataset = dataset_authorize(@project.postgres_resources_dataset.eager(:timeline, representative_server: [:strand, vm: :vm_storage_volumes]), "Postgres:view").eager(:semaphores, :location, strand: :children)
+
+    if tag_key
+      dataset = if tag_value
+        # Filter by both key and value: tags contains {"key": "tag_key", "value": "tag_value"}
+        dataset.where(Sequel.pg_jsonb_op(:tags).contains([{key: tag_key, value: tag_value}]))
+      else
+        # Filter by key only: tags contains an object with the specified key
+        dataset.where(Sequel.lit("EXISTS (SELECT 1 FROM jsonb_array_elements(tags) AS tag WHERE tag->>'key' = ?)", tag_key))
```
byucesoy: Is being able to fetch records that has specified key (but random value) really useful? I'm not sure on the value of this, but it is up to you.

helpers/postgres.rb
furkansahin: What would be the expected behaviour in that case? Would we filter the resources with `or` or `and` operator? Should we return resources where all of the tags are applicable or only one of them is enough? I would lean towards `AND`

byucesoy: Yep, `AND` would make more sense.

helpers/postgres.rb
furkansahin: I was thinking tags without value but removed it in this iteration

helpers/postgres.rb
furkansahin: updated accordingly

helpers/postgres.rb
```diff
@@ -56,8 +56,20 @@ def postgres_post(name)
...
+
+    if tags_param
+      tags_param = tags_param.split(",").map { |tag| tag.split(":", 2).map(&:strip) }
+      tags_param = tags_param.map { |key, value| {key:, value:} }
+      tags_param.each do |tag|
+        if tag.length != 2 || tag[:key].nil? || tag[:value].nil?
+          fail Validation::ValidationFailed.new({tags: "Invalid tag format. Expected format: key:value"})
+        end
+      end
+      dataset = dataset.where(Sequel.pg_jsonb_op(:tags).contains(tags_param))
```
serprex, line 71: for my understanding: these tags could instead be setup with ACL so result would filter out the rest, tho even in that case it'd be useful to have this as an extra layer in case ACL is misconfigured (or if api user wants to filter tags for a subset of their pg resources ofc)

is that right?

helpers/postgres.rb
```diff
@@ -56,8 +56,20 @@ def postgres_post(name)
     end
   end
 
-  def postgres_list
+  def postgres_list(tags_param: nil)
     dataset = dataset_authorize(@project.postgres_resources_dataset.eager(:timeline, representative_server: [:strand, vm: :vm_storage_volumes]), "Postgres:view").eager(:semaphores, :location, strand: :children)
+
+    if tags_param
+      tags_param = tags_param.split(",").map { |tag| tag.split(":", 2).map(&:strip) }
+      tags_param = tags_param.map { |key, value| {key:, value:} }
```
jeremyevans:
```suggestion
      tags_param = tags_param.split(",")
      tags_param.map! { |tag| tag.split(":", 2).map(&:strip) }
      tags_param.map! { |key, value| {key:, value:} }
```

routes/project/postgres.rb
```diff
@@ -3,7 +3,8 @@
 class Clover
   hash_branch(:project_prefix, "postgres") do |r|
     r.get true do
-      postgres_list
+      tags_param = typecast_params.str("tags")
```
jeremyevans: How do we want to handle an empty `tags` param? If we want to ignore it, I would use:


```suggestion
      tags_param = typecast_params.nonempty_str("tags")
```

spec/routes/api/project/postgres_spec.rb
```diff
@@ -44,5 +44,127 @@
...
+      ).subject
+      pg2.update(tags: [{key: "environment", value: "staging"}, {key: "team", value: "backend"}])
+
+      # Match environment production and team backend
+      get "/project/#{project.ubid}/postgres?tags=environment:production,team:backend"
+
+      expect(last_response.status).to eq(200)
+      body = JSON.parse(last_response.body)
+      expect(body["items"].length).to eq(1)
+      names = body["items"].map { |item| item["name"] }.sort
```
jeremyevans: `sort` not needed when there is only one item.

```suggestion
      names = body["items"].map { |item| item["name"] }
```

openapi/openapi.yml
```diff
@@ -2187,6 +2188,14 @@ components:
       required: false
       schema:
         type: string
+    tags:
+      description: 'Filter by tags in key:value format (comma-separated for multiple tags).'
```
jeremyevans: We don't include `.` at the end of other descriptions.

```suggestion
      description: 'Filter by tags in key:value format (comma-separated for multiple tags)'
```

helpers/postgres.rb
```diff
@@ -56,8 +56,20 @@ def postgres_post(name)
...
+      tags_param.each do |tag|
+        if tag.length != 2 || tag[:key].nil? || tag[:value].nil?
```
jeremyevans: Based on the preceding code, I think it's impossible for `tag.length` to not equal `2`, or for `tag[:key]` to be `nil`.

```suggestion
        unless tag[:value]
```

routes/project/postgres.rb
byucesoy: Instead of getting the tags as string and then trying to parse it, we can just do this (similar to what we do [here](https://github.com/ubicloud/ubicloud/blob/c629c738166c90f251d3e476876f0accea516868/routes/project/location/postgres.rb#L52))

```
tags = typecast_params.array(:Hash, "tags")
```

furkansahin: That params come from the body, which is easy to pass in a dictionary. Here, we are getting it from the path and the whole representation of the dictionary as a string is quite dirty. I like the current implementation

helpers/postgres.rb
furkansahin: Yes and no. We could have tags via ACL but that would not be these tags, it would be `lib/AccessControlModelTag`. Since that tag implementation is not very commonly used and we already have a separate `tags` column here in PostgresResource, we opted to use this one for the time being. In future, if we want to move, we could simply migrate these tags to AccessControlModelTags and start using that as a single source of truth.

helpers/postgres.rb
furkansahin: Good catch, thank you

routes/project/postgres.rb
byucesoy: I didn't realized that these tags come from path. How about getting these tags from the body as well then? It is bit weird that in two different places, we get the same type of parameter in different ways.

PR #4256, opened 2025-11-25
Add PostgresResource GET backups endpoint

furkansahin:


routes/project/location/postgres.rb
```diff
@@ -385,6 +385,22 @@ class Clover
         certs
       end
 
+      r.get api?, "backup" do
```
serprex, line 388:
```suggestion
      r.get api?, "backups" do
```

routes/project/location/postgres.rb
```diff
@@ -385,6 +385,22 @@ class Clover
         certs
       end
 
+      r.get api?, "backups" do
+        authorize("Postgres:view", pg)
+
+        backups = pg.timeline.backups.map do |backup|
```
jeremyevans, line 391: Will we ever have enough backups that it makes sense to paginate this list?

furkansahin: The idea is to introduce incremental backups + a custom backup cadence in future. So, it is possible. But if you think it is unnecessary to keep this at the moment, I can remove it and we'd add it back when the time comes.

routes/project/location/postgres.rb
enescakir: When did we start using plural names for resource listing endpoints? Most of our resources use singular endpoints, so I assumed that was our convention (e.g., `/project`, `/vm`, `/postgres/ID/firewall-rule`, `/postgres/ID/read-replica`).

routes/project/location/postgres.rb
enescakir: Since we return the same interface (`{items:, count:}`) with paginated results, we can add support in future PRs. IMHO, it's not a blocker for this one.


https://github.com/ubicloud/ubicloud/blob/6d734121b3201525badf57b7df72e71692083321/lib/pagination.rb#L40

cli-commands/pg/post/backups.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("backups") do
+  desc "List backups for a PostgreSQL database"
+
+  key = :backups_list
+
+  options("ubi pg (location/pg-name | pg-id) backups [options]", key:) do
```
enescakir: I'm not sure whether we have an overall convention, but Jeremy added `list-repositories` and `list-cache-entries` commands for GitHub installations. `list-backups` would be more consistent with those.

enescakir (approved):

PR #4257, opened 2025-11-25
Add /project/{project_id}/postgres/capabilities

serprex:
API endpoint exposes accessible locations alongside SKU info, which currently lack discoverability

config/instance_availability.yml is generated: bin/generate_instance_availability.rb


helpers/postgres.rb
```diff
@@ -130,6 +130,50 @@ def generate_postgres_options(flavor: nil, location: nil)
     options.serialize
   end
 
+  PostgresLocation = Struct.new(:location, :available_postgres_versions, :available_vm_families)
```
jeremyevans: If instances don't need to be mutable:

```suggestion
  PostgresLocation = Data.define(:location, :available_postgres_versions, :available_vm_families)
```

helpers/postgres.rb
```diff
@@ -130,6 +130,50 @@ def generate_postgres_options(flavor: nil, location: nil)
...
+  def vm_families_for_project(project)
+    locations = Location.visible_or_for_project(@project.id, @project.feature_flags["visible_locations"])
+    locations.map do |location|
+      PostgresLocation.new(
+        location: location,
+        available_postgres_versions: Option::POSTGRES_VERSION_OPTIONS[PostgresResource::Flavor::STANDARD],
```
jeremyevans: Maybe worth adding a local variable? `available_postgres_versions = Option::POSTGRES_VERSION_OPTIONS[PostgresResource::Flavor::STANDARD]`

helpers/postgres.rb
```diff
@@ -130,6 +130,50 @@ def generate_postgres_options(flavor: nil, location: nil)
     options.serialize
   end
 
+  PostgresLocation = Struct.new(:location, :available_postgres_versions, :available_vm_families)
+  def vm_families_for_project(project)
+    locations = Location.visible_or_for_project(@project.id, @project.feature_flags["visible_locations"])
+    locations.map do |location|
+      PostgresLocation.new(
+        location: location,
```
jeremyevans: Hash value omission:

```suggestion
        location:,
```

Can be applied to other cases in this PR.

openapi/openapi.yml
```diff
@@ -1767,6 +1767,40 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
+  '/project/{project_id}/postgres-location':
```
jeremyevans: Seems like we would want this under the postgres branch:

```suggestion
  '/project/{project_id}/postgres/location':
```


@geemus What are your thoughts?

openapi/openapi.yml
```diff
@@ -2970,6 +3004,123 @@ components:
...
+          example: true
+        available_postgres_versions:
+          description: List of supported PostgreSQL versions
+          type: array
+          items:
+            type: string
+          example:
+            - '14'
+            - '15'
+            - '16'
```
jeremyevans: We should avoid example values that are invalid:

```suggestion
            - '18'
            - '17'
            - '16'
```

routes/project/postgres_location.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Clover
+  hash_branch(:project_prefix, "postgres-location") do |r|
+    r.get true do
+      authorize("Postgres:view", @project)
+      {items: vm_families_for_project(@project).map { |l| Serializers::PostgresLocation.serialize_internal(l) }}
```
jeremyevans: I think this should work:

```suggestion
      {items: Serializers::PostgresLocation.serialize(vm_families_for_project(@project))}
```

serializers/postgres_location.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class Serializers::PostgresLocation < Serializers::Base
+  def self.serialize_internal(postgres_location, options = {})
+    {
+      name: postgres_location.location.name,
```
jeremyevans: Probably worth having a `location = postgres_location.location` local variable to DRY things up.

jeremyevans (commented):
Looks good. I have some minor tweaks, but the main change is moving the path to under `postgres` instead of next to it.

openapi/openapi.yml
serprex: also thinking `/locations` instead of `/location`

iamKunalGupta: iirc in the UI we support querying via name, what if customer pg is named `location`?

jeremyevans:
I forgot to mention during my earlier review, but we generally want our CLI to support our entire API, so we should add CLI support for this. However, that's something I can work on after this PR if you don't want to do it as part of this PR.

serprex:
> I forgot to mention during my earlier review, but we generally want our CLI to support our entire API, so we should add CLI support for this. However, that's something I can work on after this PR if you don't want to do it as part of this PR.

this came up in #4191 so I'll be able to add CLI support here, but good to get consensus on API first

openapi/openapi.yml
```diff
@@ -1767,6 +1767,40 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
+  '/project/{project_id}/postgres/locations':
```
byucesoy: I think the name of the endpoint (`locations`) does not reflect well on what this endpoint actually returns. It doesn't just return available locations, but also available VM sizes and PostgreSQL versions.

How about something like `capabilities` or `options`? Latter is bit too generic, but `capabilities` seem to capture the essence of what we are trying to do here.

furkansahin:
Please rewrite the commit message:
```
Feat: Add route for postgres locations
Api endpoint gives list of accessible locations alongside SKU info,
which is otherwise lacking discoverability
```

openapi/openapi.yml
```diff
@@ -2970,6 +3004,123 @@ components:
...
+            $ref: '#/components/schemas/VmFamily'
+          example:
+            - name: standard
+              display_name: Dedicated CPU
+              category: general-purpose
+              sizes:
+                - name: standard-2
+                  vcpu: 2
+                  memory_gib: 8
+                  storage_size_gib: 256
```
byucesoy: The possibilities here are much more complicated that you might be realizing.

Some examples:
- As you are aware, some instance types are not available in some locations. So as you designed, **instance types is a child of location**.
- For this to make sense for non-AWS scenarios, we should be able to return multiple storage sizes per vm size. For example standard-2 can come up with 64, 128 and 256 GB. Even in the AWS case, if we end up adding EBS disks, there could be multiple storage sizes. **So storage sizes should be a child of instance sizes**.
- How about HA options? Currently we allow setting up HA for burstable instances, but I'm getting more and more convinced that it was not a good idea. It gives the impression of that burstable instances are good for production workloads (they are not). **Therefore HA options should be child of instance sizes as well**.

As you look into this more and more, you will realize, this endpoint should return a tree, not a list, because availability of every option depends on which parent option is picked beforehand. This is why option tree is a tree and it has lots of repetitions in it. We initially went with keeping only list of possibilities for each option and then implemented custom logic for each parent-child relationship, but over the time it become very very complicated. Adding/removing an option was time consuming and error prone. We ended up using tree abstraction, because it fits well to this problem (and because, the data we are dealing is a tree, really). Please check [this PR](https://github.com/ubicloud/ubicloud/pull/2235) description for extra context.

I think what this endpoints needs to do is just getting the option tree, serializing it and returning to the user.

byucesoy: BTW, adding this note to clarify my thinking: This is not just a recommendation. I really think that we should return a proper tree of all capabilities from this endpoint. :slightly_smiling_face:

openapi/openapi.yml
serprex: @byucesoy would route `/project/{id}/capabilities/postgres` work?

jeremyevans: I don't think a customer location named `location` is an issue, because this doesn't conflict with an existing route. The postgres routes for a specific location start with `/project/{project_id}/location/{location_id_or_name}/postgres`, not `/project/{project_id}/postgres`

byucesoy: > @byucesoy would route /project/{id}/capabilities/postgres work?

I think `/project/{id}/postgres/capabilities` would be better. I guess you recommended the other way because of possible conflict, but Jeremy's comment clarifies that.

routes/project/postgres.rb
```diff
@@ -7,8 +7,10 @@ class Clover
         authorize("Postgres:view", @project)
         option_tree, = PostgresResource.generate_postgres_options(@project)
```
serprex, line 8: Burak suggested filtering should happen in generate_postgres_options

helpers/postgres.rb
```diff
@@ -108,6 +108,24 @@ def send_notification_mail_to_partners(resource, user_email)
...
 
+  def serialize_option_tree(node)
+    node.each_with_object({}) do |(key, value), result|
+      next if value.nil?
+      serialized_key = key.is_a?(Location) ? key.name : key.to_s
+      result[serialized_key] = serialize_option_tree(value)
+    end
+  end
+
+  def postgres_option_metadata(locations)
```
byucesoy: We shouldn't return all values here. Only values that are valid for given project must be returned.

model/postgres/metal/postgres_resource.rb
```diff
@@ -2,6 +2,20 @@
 
 class PostgresResource < Sequel::Model
   module Metal
+    def self.available_families_and_sizes(_location)
+      nil
```
byucesoy: This should return ["standard", "hobby"]

model/postgres/postgres_resource.rb
```diff
@@ -494,29 +494,24 @@ def self.generate_postgres_options(project, flavor: nil, location: nil)
...
 
+    availability_cache = {}
     options.add_option(name: "family", values: Option::POSTGRES_FAMILY_OPTIONS.keys, parent: "location") do |flavor, location, family|
-      if location.aws?
-        ["m8gd", "i8g"].include?(family) || (Option::AWS_FAMILY_OPTIONS.include?(family) && project.send(:"get_ff_enable_#{family}"))
-      else
-        family == "standard" || family == "hobby"
-      end
+      available = (availability_cache[location.name] ||= available_families_and_sizes(location))
+      next false if available&.none? { |f, _| f == family }
```
byucesoy: After ensuring metal/postgres_resource.rb returns correct values, we don't need `&.` here.

model/postgres/postgres_resource.rb
```diff
@@ -494,29 +494,24 @@ def self.generate_postgres_options(project, flavor: nil, location: nil)
       flavor == PostgresResource.default_flavor || location.provider != "aws"
     end
 
+    availability_cache = {}
```
byucesoy: We can populate availability_cache in the above add_option block, which iterates all related locations. It would simplify some of the code below.

byucesoy: BTW, I'd prefer to use more explicit name, even if it is long.

model/postgres/metal/postgres_resource.rb
```diff
@@ -2,6 +2,20 @@
 
 class PostgresResource < Sequel::Model
   module Metal
+    def self.available_families_and_sizes(_location)
```
byucesoy: Instance methods would fit better here. Because they already have location and project information in them.

serprex: trouble is capabilities endpoint wants class methods since it doesn't have pg instance there, same for validate_postgres_input

serprex: updated to take project as 2nd param, feature flags now handled here

helpers/postgres.rb
```diff
@@ -108,6 +108,37 @@ def send_notification_mail_to_partners(resource, user_email)
...
+    {
+      flavor: Option::POSTGRES_FLAVOR_OPTIONS.slice(*valid["flavor"]).transform_values { |v| {display_name: v.title} },
+      location: (valid["location"] || []).to_h { |l| [l.name, {display_name: l.display_name, ui_name: l.ui_name, provider: l.provider}] },
+      family: Option::POSTGRES_FAMILY_OPTIONS.slice(*valid["family"]).transform_values { |v| {display_name: v.description, category: v.category} },
+      size: Option::POSTGRES_SIZE_OPTIONS.slice(*valid["size"]).transform_values { |v| {vcpu: v.vcpu_count, memory_gib: v.memory_gib} },
+      ha_type: Option::POSTGRES_HA_OPTIONS.slice(*valid["ha_type"]).transform_values { |v| {display_name: v.description, standby_count: v.standby_count} },
+    }
+  end
+
+  def collect_valid_option_keys(tree)
```
byucesoy: This and generate_allowed_options in option_tree_generator.rb seems to very similar things. It generates the allowed options for only one field as opposed for all fields. We can extend it to generate the options for all fields or simply call it multiple times. Slow performance might be acceptable if it isn't too extreme.

helpers/postgres.rb
```diff
@@ -108,6 +108,37 @@ def send_notification_mail_to_partners(resource, user_email)
     end
   end
 
+  def serialize_option_tree(node)
```
byucesoy: I wonder how much of the `serialize` method in option tree generator would handle the requirement of this function. Maybe we can re-use it to some extent?

lib/option_tree_filter.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require "yaml"
+
+class OptionTreeFilter
```
byucesoy, line 5: We can rename OptionTreeGenerator to OptionTree and combine Generating and Filtering together.

model/postgres/metal/postgres_resource.rb
```diff
@@ -2,6 +2,18 @@
 
 class PostgresResource < Sequel::Model
   module Metal
+    def self.available_families_and_sizes(_location, _project)
+      Set.new(
```
byucesoy, line 6: What is the benefit of Set over regular array here (and in other `available_families_and_sizes` function)?

lib/option_tree_filter.rb
byucesoy: Thinking about this more, since this would affect non-PG code as well, I can take over it if you prefer. Up to you.

model/postgres/metal/postgres_resource.rb
jeremyevans: Array search is `O(N)`, set search is `O(1)`. Different constant factors, though. It makes a large difference for a large number of elements. Not sure how large the resulting set is, but I wouldn't advocate for changing it to an array.

helpers/postgres.rb
serprex: `serialize` seems misnamed, as it has Location instances in its output for `PostgresResource.generate_postgres_options`. Is that a bug?
```
{"location" =>
       {#<Location["10saktg1sprp3mxefj1m3kppq2"] @values={display_name: "eu-central-h1", name: "hetzner-fsn1", ui_name: "Germany", visible: false, provider: "hetzner", project_id: nil, dns_suffix: nil, by
oc: false, otel_otlp_export_endpoint: nil}> =>
         {"family" =>
```
thus becoming `Location["10stj19wg33e6k797t27x2f2yh"]` as key in `JSON.generate` instead of location's name which we'd prefer

helpers/postgres.rb
serprex: I think gathering these methods into `option_tree_generator.rb` makes sense, has code live near each other, but the bit of duplication seems negligible for what makes its purpose clear

byucesoy (approved):

PR #4261, opened 2025-11-26
Add firewall rules on AWS to allow udp access to dns server

serprex:
Only tcp was unblocked, but dns should have port 53 open on udp

----

> [!IMPORTANT]
> Add UDP protocol support for DNS server firewall rules on AWS, updating logic and tests to handle protocol in firewall rules.
> 
>   - **Behavior**:
>     - Adds UDP protocol support for DNS server on AWS by opening port 53 for UDP in `setup_dns_server_vm.rb`.
>     - Updates `allow_ingress` in `vpc.rb` to use protocol from firewall rules.
>     - Updates `update_firewall_rules.rb` to handle protocol in firewall rules.
>   - **Database**:
>     - Adds `protocol` column to `firewall_rule` table with default "tcp" in `20251130_add_firewall_protocol.rb`.
>   - **Models**:
>     - Updates `FirewallRule` model to include `protocol` attribute.
>   - **Tests**:
>     - Adds test for UDP protocol handling in `setup_dns_server_vm_spec.rb`.
>     - Updates `update_firewall_rules_spec.rb` to test protocol handling in firewall rules.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b359f52c5889a8d907e691147f044e3d58420321. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/dns_zone/setup_dns_server_vm.rb
+++ b/prog/dns_zone/setup_dns_server_vm.rb
@@ -68,6 +68,19 @@ class Prog::DnsZone::SetupDnsServerVm < Prog::Base
   label def start
     nap 5 unless vm.strand.label == "wait"
     register_deadline(nil, 15 * 60)
+    if vm.location.aws?
+      # Open UDP & TCP port 53 for DNS queries on AWS
+      fw = Firewall[name: "dns", project_id: Config.dns_service_project_id]
+      unless fw
+        fw = Firewall.create(name: "dns", location: vm.location, project_id: Config.dns_service_project_id)
+        fw.add_firewall_rule(cidr: "0.0.0.0/0", port_range: 53..53, protocol: "udp")
+        fw.add_firewall_rule(cidr: "::/0", port_range: 53..53, protocol: "udp")
+        fw.add_firewall_rule(cidr: "0.0.0.0/0", port_range: 53..53, protocol: "tcp")
+        fw.add_firewall_rule(cidr: "::/0", port_range: 53..53, protocol: "tcp")
+      end
+      vm.add_vm_firewall(fw)
+      vm.incr_update_firewall_rules
+    end
     hop_prepare
   end
 
--- a/spec/prog/dns_zone/setup_dns_server_vm_spec.rb
+++ b/spec/prog/dns_zone/setup_dns_server_vm_spec.rb
@@ -111,11 +111,38 @@ RSpec.describe Prog::DnsZone::SetupDnsServerVm do
       expect { prog.start }.to nap(5)
     end
 
-    it "hops to prepare when VM is ready" do
+    it "hops to prepare when VM is ready but not on AWS" do
+      expect(prog.vm.location).to receive(:aws?).and_return(false)
       prog.vm.strand.update(label: "wait")
       expect(prog.strand.stack.first["deadline_at"]).to be_nil
       expect { prog.start }.to hop("prepare")
       expect(prog.strand.stack.first["deadline_at"]).not_to be_nil
+    end
+
+    it "configures firewall when AWS VM is ready" do
+      expect(prog.vm.location).to receive(:aws?).and_return(true).at_least(:once)
+      prog.vm.strand.update(label: "wait")
+      expect(prog.strand.stack.first["deadline_at"]).to be_nil
+      expect { prog.start }.to hop("prepare")
+      expect(prog.vm.firewall_rules.filter { it.protocol == "udp" }.length).to eq(2)
+      expect(prog.strand.stack.first["deadline_at"]).not_to be_nil
+    end
+
+    it "reuses firewall when AWS VM is ready with existing firewall" do
+      fw = Firewall.create(name: "dns", location: prog.vm.location, project_id: Config.dns_service_project_id)
+      fw.add_firewall_rule(cidr: "0.0.0.0/0", port_range: 53..53, protocol: "udp")
+      fw.add_firewall_rule(cidr: "::/0", port_range: 53..53, protocol: "udp")
+      fw.add_firewall_rule(cidr: "0.0.0.0/0", port_range: 53..53, protocol: "tcp")
+      fw.add_firewall_rule(cidr: "::/0", port_range: 53..53, protocol: "tcp")
+
+      expect(prog.vm.location).to receive(:aws?).and_return(true).at_least(:once)
+      prog.vm.strand.update(label: "wait")
+      expect(prog.strand.stack.first["deadline_at"]).to be_nil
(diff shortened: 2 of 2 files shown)


prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -68,6 +68,39 @@ def ds
   label def start
     nap 5 unless vm.strand.label == "wait"
     register_deadline(nil, 15 * 60)
+    hop_configure_firewall
+  end
+
+  label def configure_firewall
+    # Add UDP port 53 for DNS queries on AWS
+    if vm.location.aws?
+      sg_id = vm.nic.private_subnet.private_subnet_aws_resource.security_group_id
```
enescakir: Since whole label in a if block, maybe
```suggestion
    hop_configure_aws_firewall if vm.location.aws?
    hop_prepare
  end

  label def hop_configure_aws_firewall
    # Add UDP port 53 for DNS queries on AWS
   
    sg_id = vm.nic.private_subnet.private_subnet_aws_resource.security_group_id
```

serprex: added as 2nd commit, can squash when this is ready to merge. not sure which is preferable looking forward to more cloud providers

furkansahin: I'd add this firewall rule to the firewall that is attached to the subnet instead of directly calling this endpoint here for each VM. We already have this functionality implemented anyway.

prog/aws/vpc.rb
```diff
@@ -142,11 +142,11 @@ def client
     @client ||= location.location_credential.client
   end
 
-  def allow_ingress(group_id, from_port, to_port, cidr)
+  def allow_ingress(group_id, from_port, to_port, cidr, protocol: "tcp")
```
furkansahin: you're not making use of this function

migrate/20251130_add_firewall_protocol.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:firewall_rule) do
+      add_column :protocol, :text, null: false, default: "tcp"
```
jeremyevans, line 7: Should add a constraint:

```ruby
constraint :valid_protocol, protocol: %w[tcp udp]
```

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -68,6 +69,14 @@ def ds
   label def start
     nap 5 unless vm.strand.label == "wait"
     register_deadline(nil, 15 * 60)
+    if vm.location.aws?
+      # Add UDP port 53 for DNS queries on AWS
+      fw = Firewall.create(name: "#{vm.name}-udp53", location: vm.location, project_id: vm.project_id)
+      fw.add_firewall_rule(cidr: "0.0.0.0/0", port_range: 53..53, protocol: "udp")
+      fw.add_firewall_rule(cidr: "::/0", port_range: 53..53, protocol: "udp")
```
jeremyevans: We should consider adding both `tcp` and `udp` firewall rules as DNS can operate over both, and there are cases where `tcp` needs to be used.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -68,6 +69,14 @@ def ds
   label def start
     nap 5 unless vm.strand.label == "wait"
     register_deadline(nil, 15 * 60)
+    if vm.location.aws?
+      # Add UDP port 53 for DNS queries on AWS
+      fw = Firewall.create(name: "#{vm.name}-udp53", location: vm.location, project_id: vm.project_id)
```
jeremyevans: Since the `Firewall` can be the same for each, I would use a single `Firewall` instance named `aws-dns`, creating it if it doesn't already exist.

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -33,7 +33,8 @@ def self.assemble(dns_server_id, name: nil, vm_size: "standard-2", storage_size_
           {encrypted: true, size_gib: storage_size_gib}
         ],
         boot_image: "ubuntu-jammy",
-        enable_ip4: true
+        enable_ip4: true,
+        new_private_subnet_name: "#{name}-dns-subnet"
```
jeremyevans: Can you explain the change to not use the default private subnet? It doesn't seem problematic, but it isn't mentioned in the commit message and doesn't seem related to the other changes.

serprex: I was thinking this was necessary for having firewall rules not apply to default private subnet, but since I ended up going with vm firewall rules this can be reverted

prog/dns_zone/setup_dns_server_vm.rb
serprex: That seems incorrect when location/project vary between VMs. Not clear what you were thinking those would be for a global firewall record

jeremyevans: While `Firewall` has a `location_id`, it isn't actually location specific, and there is no issue if the firewall is in a different location from the VM. When creating the `Firewall`, you could use the given `location_id`, but you don't need to check for a matching location when looking up the firewall. The project_id in this case is always `Config.dns_service_project_id`

serprex: In that case seems like this firewall could be shared across cloud vendors in future, so going to omit `aws-` from name

furkansahin:
this passed E2E https://github.com/ubicloud/ubicloud/actions/runs/19890663765

migrate/20251130_add_firewall_protocol.rb
```diff
@@ -0,0 +1,21 @@
...
+      drop_constraint :firewall_rule_cidr_port_range_firewall_id_key
+      add_column :protocol, :text, null: false, default: "tcp"
+      add_unique_constraint [:cidr, :port_range, :firewall_id, :protocol], name: :firewall_rule_unique
+      add_constraint :valid_protocol, protocol: %w[tcp udp]
+    end
+  end
+
+  down do
+    alter_table(:firewall_rule) do
+      drop_column :protocol, :text, null: false, default: "tcp"
```
jeremyevans:
```suggestion
      drop_column :protocol
```

jeremyevans (approved):
OK after minor fix to down migration.

PR #4280, opened 2025-12-01
Upgrade read replicas alongside parent

shikharbhardwaj:
Read replicas now get upgraded with the parent, following a different upgrade path for simplicity. Instead of an in-place upgrade via `pg_upgrade`, we create a new server using the latest timeline and version, and failover to it after it is ready.

----

> [!IMPORTANT]
> Read replicas are now upgraded by creating a new server and failing over, simplifying the process and ensuring consistency with parent upgrades.
> 
>   - **Behavior**:
>     - Read replicas are upgraded by creating a new server and failing over, instead of in-place upgrade.
>     - `ready_for_read_replica?` method in `postgres_resource.rb` checks if a parent is ready for read replica creation.
>     - `start` method in `converge_postgres_resource.rb` waits if parent is not ready for read replica.
>   - **Provisioning**:
>     - `assemble` method in `postgres_server_nexus.rb` uses `target_version` for read replicas.
>     - `provision_servers` in `converge_postgres_resource.rb` uses parent's timeline for read replicas.
>   - **Tests**:
>     - Added tests for `ready_for_read_replica?` in `postgres_resource_spec.rb`.
>     - Updated tests in `converge_postgres_resource_spec.rb` to cover new read replica behavior.
>     - Updated API and CLI tests to handle read replica creation and promotion.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 3a163b4549905876761677de7e50cd03fbf35029. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/converge_postgres_resource.rb
+++ b/prog/postgres/converge_postgres_resource.rb
@@ -6,7 +6,10 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
   subject_is :postgres_resource
 
   label def start
+    nap 60 if postgres_resource.read_replica? && !postgres_resource.parent.ready_for_read_replica?
+
     register_deadline("wait_for_maintenance_window", 2 * 60 * 60)
+
     hop_provision_servers
   end
 
@@ -30,7 +33,8 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
         end
       end
 
-      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids, exclude_availability_zones: exclude_availability_zones, availability_zone: availability_zone)
+      timeline_id = postgres_resource.read_replica? ? postgres_resource.parent.timeline.id : postgres_resource.timeline.id
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id:, timeline_access: "fetch", exclude_host_ids: exclude_host_ids, exclude_availability_zones: exclude_availability_zones, availability_zone: availability_zone)
     end
 
     nap 5
@@ -48,7 +52,10 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
 
     hop_provision_servers unless postgres_resource.has_enough_fresh_servers?
 
-    if postgres_resource.version != postgres_resource.target_version
+    # Read replicas skip the in-place upgrade process and directly
+    # recycle servers, which are provisioned at the target version instead of
+    # the current version.
+    if postgres_resource.version != postgres_resource.target_version && !postgres_resource.read_replica?
       postgres_resource.representative_server.incr_fence
       hop_wait_fence_primary
     end
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -16,14 +16,18 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
       ubid = PostgresServer.generate_ubid
 
       postgres_resource = PostgresResource[resource_id]
+      # For read replicas, we upgrade by replacing servers instead of in-place
+      # upgrade.
+      server_version = postgres_resource.read_replica? ? postgres_resource.target_version : postgres_resource.version
+
       arch = Option::VmSizes.find { it.name == postgres_resource.target_vm_size }.arch
       boot_image = if postgres_resource.location.aws?
-        postgres_resource.location.pg_ami(postgres_resource.version, arch)
+        postgres_resource.location.pg_ami(server_version, arch)
       else
         flavor_suffix = case postgres_resource.flavor
         when PostgresResource::Flavor::STANDARD then ""
         when PostgresResource::Flavor::PARADEDB then "-paradedb"
-        when PostgresResource::Flavor::LANTERN then "#{postgres_resource.version}-lantern"
+        when PostgresResource::Flavor::LANTERN then "#{server_version}-lantern"
         else raise "Unknown PostgreSQL flavor: #{postgres_resource.flavor}"
         end
 
@@ -60,7 +64,7 @@ class Prog::Postgres::PostgresServerNexus < Prog::Base
         representative_at: representative_at,
         synchronization_status: synchronization_status,
         vm_id: vm_st.id,
-        version: postgres_resource.version
+        version: server_version
       )
 
       vm_st.subject.add_vm_firewall(postgres_resource.internal_firewall)
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -243,7 +243,7 @@ class Clover
         Validation.validate_name(name)
 
         Validation.validate_vcpu_quota(@project, "PostgresVCpu", Option::POSTGRES_SIZE_OPTIONS[pg.target_vm_size].vcpu_count)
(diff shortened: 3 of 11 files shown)


prog/postgres/postgres_resource_nexus.rb
```diff
@@ -216,7 +216,7 @@ def before_run
   label def wait
     reap(fallthrough: true)
 
-    if postgres_resource.needs_convergence? && strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").empty?
+    if postgres_resource.needs_convergence? && !postgres_resource.parent&.needs_convergence? && strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").empty?
```
byucesoy: I think we also have to wait until the parent gets a new backup, right?

byucesoy: Instead of making this check crowded, we can start ConvergePostgresResource prog and wait at the first label for read replicas until
- parent doesn't need converging
- parent has new backup

shikharbhardwaj: It should implicitly block until the backup is ready (the new server will be stuck retrying in `initialize_database_from_backup`). But I guess that would be noisy in the logs, so makes sense to explicitly wait for it. We don't do it for standbys at the moment though.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -49,8 +56,15 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
     hop_provision_servers unless postgres_resource.has_enough_fresh_servers?
 
     if postgres_resource.version != postgres_resource.target_version
```
jeremyevans: This seems simpler than the separate conditional, though I think you should keep the comment. Up to you:

```suggestion
    if postgres_resource.version != postgres_resource.target_version && !postgres_resource.read_replica?
```

routes/project/location/postgres.rb
```diff
@@ -567,6 +567,7 @@ class Clover
 
           DB.transaction do
             pg.update(target_version: pg.version.to_i + 1)
+            pg.read_replicas_dataset.update(target_version: pg.version.to_i + 1)
```
jeremyevans: Seems simpler to reuse the existing calculation:

```suggestion
            pg.read_replicas_dataset.update(target_version: pg.target_version)
```

prog/postgres/converge_postgres_resource.rb
```diff
@@ -7,6 +7,12 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
 
   label def start
     register_deadline("wait_for_maintenance_window", 2 * 60 * 60)
+
+    if postgres_resource.read_replica?
+      nap 60 if postgres_resource.parent.needs_convergence?
+      nap 60 if PostgresTimeline.earliest_restore_time(postgres_resource.parent.timeline).nil?
```
furkansahin: can we merge this into a single logic and call from the resource?

prog/postgres/converge_postgres_resource.rb
```diff
@@ -30,7 +36,8 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
         end
       end
 
-      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids, exclude_availability_zones: exclude_availability_zones, availability_zone: availability_zone)
+      timeline_id = postgres_resource.read_replica? ? postgres_resource.parent.timeline.id : postgres_resource.timeline.id
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids, exclude_availability_zones: exclude_availability_zones, availability_zone: availability_zone)
```
furkansahin: why do we have this change? If this was a bug, I'd prefer if it was a different commit. Can you explain why do we have this change and how did it work before?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -16,14 +16,18 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
       ubid = PostgresServer.generate_ubid
 
       postgres_resource = PostgresResource[resource_id]
+      # For read replicas, we upgrade by replacing servers instead of in-place
+      # upgrade.
+      server_version = postgres_resource.read_replica? ? postgres_resource.target_version : postgres_resource.version
```
furkansahin, line 21: At this point, is the parent already upgraded? Should we not wait for the parent to be completely upgraded before calling this? If we do that already, isn't the `postgres_resource.version` and `target_version` are the same?

prog/postgres/converge_postgres_resource.rb
shikharbhardwaj: Read replicas were not upgraded before this change, they had to be recreated/upgraded from the backend

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Yes, the parent is already upgraded at this point (we guarantee it by waiting in `start`).
No, the version and target version would not be the same, as the `postgres_resource` here is the read replica, which still needs the upgrade.

For non-read replicas, the candidate standby needs to be on the current version to catch up and run `pg_upgrade`.

prog/postgres/converge_postgres_resource.rb
furkansahin: Sure, but I'm asking why do we have the timeline_id extraction and selection based on parent

shikharbhardwaj: So, at this point for a read replica, the parent is upgraded and is on the timeline for the latest version, we use that because the new candidate is also created on the latest version.
For non-read replicas, the existing timeline is used as the candidate is created on the current version.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -7,6 +7,9 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
 
   label def start
     register_deadline("wait_for_maintenance_window", 2 * 60 * 60)
+
+    nap 60 if postgres_resource.read_replica? && !postgres_resource.parent.ready_for_read_replica?
```
byucesoy, line 9: We should register deadline after parent is ready, otherwise parent taking backup eats from the deadline time.

prog/postgres/converge_postgres_resource.rb
```diff
@@ -30,7 +33,8 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
         end
       end
 
-      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids, exclude_availability_zones: exclude_availability_zones, availability_zone: availability_zone)
+      timeline_id = postgres_resource.read_replica? ? postgres_resource.parent.timeline.id : postgres_resource.timeline.id
+      Prog::Postgres::PostgresServerNexus.assemble(resource_id: postgres_resource.id, timeline_id: timeline_id, timeline_access: "fetch", exclude_host_ids: exclude_host_ids, exclude_availability_zones: exclude_availability_zones, availability_zone: availability_zone)
```
byucesoy: You can omit passing timeline_id explicitly.

PR #4283, opened 2025-12-02
Use more secure approach for building shell commands

jeremyevans:
This is designed to address the potentially problematic usage of
string interpolation to construct SSH command strings. We recently
found a case where an interpolated string was user-modifiable and not
shell escaped. While that is potentially the only case, in general,
the approach is risky.  Instead of an opt-in shell escaping approach,
we should use an opt-out shell escaping approach, in which shell
escaping is always performed unless specifically requested not to.
This is similar to how our HTML escaping works for our templates,
where `<%=` escapes, and you must use `<%==` if you do not want
HTML escaping.

Assuming that `argument` is derived from user input, instead of
using interpolation:

```ruby
sshable.cmd("command with #{argument}")
```

You use `:placeholders` in the command string, with values
provided by keyword arguments:

```ruby
sshable.cmd("command with :argument", argument:)
```

All placeholder subsitutions are shell escaped.

In rare cases, it can be useful to pass an array of values and use
shelljoin, where the arguments of an array are each escaped, but
then joined (with space by default). You can use this by having
your placeholder start with `shelljoin_`:

```ruby
cmd("command with :shelljoin_arguments", shelljoin_arguments: array)
```

In addition to Sshable.cmd, this also works for
::Net::SSH::Connection::Session#exec! and Util.rootish_ssh.

For this to be safe, you cannot mock Sshable#cmd,
Net::SSH::Connection::Session#exec!, or Util.rootish_ssh.
You must instead mock Sshable#_cmd or
Net::SSH::Connection::Session#_exec!. This makes it an error in the
tests to call Sshable#cmd or Net::SSH::Connection::Session#exec!
without mocking _cmd or _exec!. This also adds a linger to check
for mocking of Sshable#cmd, Net::SSH::Connection::Session#exec!,
or Util.rootish_ssh in the specs, and fails if mocking of those
methods is detected.

### DIFF
--- a/lib/net_ssh.rb
+++ b/lib/net_ssh.rb
@@ -3,8 +3,13 @@
 require "net/ssh"
 require "shellwords"
 
-# :nocov:
 module NetSsh
+  class MissingMock < StandardError
+  end
+
+  class PotentialInsecurity < StandardError
+  end
+
   def self.command(command, **)
     WarnUnsafe.convert(command, self, __callee__, **)
   end
@@ -16,7 +21,7 @@ module NetSsh
 
   module WarnUnsafe
     def self.convert(command, klass, method, **kw)
-      raise "invalid type passed to #{klass}##{method}: #{command.inspect}" unless command.is_a?(String)
+      raise TypeError, "invalid type passed to #{klass}##{method}: #{command.inspect}" unless command.is_a?(String)
 
       if command.frozen?
         unless kw.empty?
@@ -31,7 +36,7 @@ module NetSsh
           command = DB.literal(command).freeze
         end
       else
-        Kernel.warn "\npotentially unsafe string passed to #{klass}##{method}: #{command.inspect}", uplevel: 2
+        raise PotentialInsecurity, "Interpolated string passed to #{klass}##{method} at #{caller(2, 1).first}\nReplace interpolation with :placeholders and provide values for placeholders in keyword arguments."
       end
 
       command
@@ -47,26 +52,30 @@ module NetSsh
 
     module SshSession
       if Config.test?
-        def _exec!(_, status: nil)
+        def _exec!(command, status: nil)
+          raise MissingMock, "Net::SSH::Connection::Session#_exec! not mocked. You must add a spec that checks for the expected command. Command: #{command.inspect}"
         end
 
         def exec!(command, **kw)
           pass_kw = WarnUnsafe.extract_keywords(kw, %i[status])
           _exec!(WarnUnsafe.convert(command, self.class, __callee__, **kw), **pass_kw)
         end
+      # :nocov:
       else
         def exec!(command, **kw)
           pass_kw = WarnUnsafe.extract_keywords(kw, %i[status])
           super(WarnUnsafe.convert(command, self.class, __callee__, **kw), **pass_kw)
         end
       end
+      # :nocov:
 
       ::Net::SSH::Connection::Session.prepend self
     end
 
     module Sshable
       if Config.test?
-        def _cmd(_, stdin: nil, log: true, timeout: :default)
+        def _cmd(command, stdin: nil, log: true, timeout: :default)
+          raise MissingMock, "Sshable#_cmd not mocked. You must add a spec that checks for the expected command. Command: #{command.inspect}"
         end
 
         # _skip_command_checking is a an extra keyword argument only for use in the
@@ -77,12 +86,14 @@ module NetSsh
           pass_kw = WarnUnsafe.extract_keywords(kw, %i[stdin log timeout])
           _cmd(WarnUnsafe.convert(cmd, self.class, __callee__, **kw), **pass_kw)
         end
+      # :nocov:
       else
         def cmd(cmd, **kw)
           pass_kw = WarnUnsafe.extract_keywords(kw, %i[stdin log timeout])
           super(WarnUnsafe.convert(cmd, self.class, __callee__, **kw), **pass_kw)
         end
       end
+      # :nocov:
     end
   end
 end
--- a/spec/lib/net_ssh_spec.rb
+++ b/spec/lib/net_ssh_spec.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+RSpec.describe NetSsh do
+  it ".command interpolates variables into string" do
+    expect(described_class.command("a :b :c d", b: 1, c: "e f")).to eq "a 1 e\\ f d"
+  end
+
+  it ".command handles shelljoin interpolation" do
+    expect(described_class.command("a :b :shelljoin_c d", b: 1, shelljoin_c: ["e f", "g"])).to eq "a 1 e\\ f g d"
+  end
+
+  it ".command returns string when not given keyword" do
+    s = "a d"
+    expect(described_class.command(s)).to be s
+  end
+
+  it ".command raises for interpolated strings with keywords" do
+    c = ":c"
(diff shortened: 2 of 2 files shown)


fdr:
What about use of %w{} or %W and moving to something like ssh.cmd(%w{}) as the shortest, obligatory pattern, with a shelljoin implied? that is, it'd take an array as the first argument, rejecting a raw string if we so desire (we can add literal string support to an out of the way, longer method name to signal caution). That won't handle hard cases (real shell directives, like pipes), but should make the easy case easy. I've long thought about doing that, but it is ugly, introduces limitations, and is hard on the fingers, but is more terse than obliging a nested utility procedures?

fdr:
I also thought about writing a rubocop to automatically convert stuff for you, writing %w{} all the time is so sad.

jeremyevans:
We can override `cmd`/`exec!` to accept an array directly and do the conversion, instead of using a new method. In terms of making the change, once we have the API down, I can use Claude to do the large mechanical changes.

One thing I still don't like about this PR is the inability to test that we are only using safe patterns, due to the mocking in the specs. I would like an approach where we can run the tests, and they will fail if a direct interpolation is used. I think overriding `cmd`/`exec!` may make that more difficult than using a new method, but maybe there is a way to get it to work.

fdr:
> We can override `cmd`/`exec!` to accept an array directly and do the conversion, instead of using a new method. In terms of making the change, once we have the API down, I can use Claude to do the large mechanical changes.

I think an array-as-first-argument is an 80/20, a lot of our stuff can go right to %w{} or %W{} I bet.
 
> One thing I still don't like about this PR is the inability to test that we are only using safe patterns, due to the mocking in the specs. I would like an approach where we can run the tests, and they will fail if a direct interpolation is used. I think overriding `cmd`/`exec!` may make that more difficult than using a new method, but maybe there is a way to get it to work.

That seems pretty hard to do, without using source reflection or some pretty invasive value-tagging (which I suppose we could).

jeremyevans:
This is failing E2E because of calls to `Util.rootish_ssh` that were mocked at a higher level. I'll need to check all usage of `rootish_ssh` for these issues, and add checks against mocking it.

spec/prog/vm/host_nexus_spec.rb
```diff
@@ -141,7 +141,9 @@
       expect(Config).to receive(:operator_ssh_public_keys).and_return(nil)
       expect(sshable).to receive(:keys).and_return([vmhost_key])
       expect(sshable).to receive(:host).and_return("127.0.0.1")
-      expect(Util).to receive(:rootish_ssh).with("127.0.0.1", "root", anything, "echo '#{test_public_keys}' > ~/.ssh/authorized_keys")
+      session = Net::SSH::Connection::Session.allocate
+      expect(Net::SSH).to receive(:start).and_yield(session)
+      expect(session).to receive(:_exec!).with("echo #{test_public_keys.gsub(" ", "\\ ")} > ~/.ssh/authorized_keys").and_return(Net::SSH::Connection::Session::StringWithExitstatus.new("", 0))
```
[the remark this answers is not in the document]
jeremyevans: This is only checking for expected spec output, and the input doesn't have backslashes, so it is unnecessary.

spec/prog/vm/host_nexus_spec.rb
```diff
@@ -157,7 +159,9 @@
       expect(Config).to receive(:operator_ssh_public_keys).exactly(2).and_return("#{operational_key_1.public_key}\n#{operational_key_2.public_key}")
       expect(sshable).to receive(:keys).and_return([vmhost_key])
       expect(sshable).to receive(:host).and_return("127.0.0.1")
-      expect(Util).to receive(:rootish_ssh).with("127.0.0.1", "root", anything, "echo '#{test_public_keys}' > ~/.ssh/authorized_keys")
+      session = Net::SSH::Connection::Session.allocate
+      expect(Net::SSH).to receive(:start).and_yield(session)
+      expect(session).to receive(:_exec!).with("echo #{test_public_keys.gsub(" ", "\\ ").gsub("\n", "'\n'")} > ~/.ssh/authorized_keys").and_return(Net::SSH::Connection::Session::StringWithExitstatus.new("", 0))
```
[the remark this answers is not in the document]
jeremyevans: This is only checking for expected spec output, and the input doesn't have backslashes, so it is unnecessary.

jeremyevans:
Another E2E failure, this time due to `kubectl`. This looks specific to E2E tests, but we should disallowing mocking `kubectl` for the same reason we disallow mocking `cmd`/`exec!`.

jeremyevans:
@fdr @byucesoy @enescakir This has passed* E2E (timed out at `67.85m | Test::PostgresResource.wait_postgres_resource | waiting for exit` just like all scheduled E2E runs have today) and is now ready for your review. Considering the importance, large number of changes all over, and high chance for conflicts, I ask that you please prioritize review.

jeremyevans:
E2E now actually passed! :)

prog/github/github_runner_nexus.rb
```diff
@@ -318,30 +318,30 @@ def setup_info
...
+        echo "CUSTOM_ACTIONS_CACHE_URL=http://:private_ipv4:51123/random_token/" | sudo tee -a /etc/environment > /dev/null
       COMMAND
     end
 
     if github_runner.label.include?("gpu")
       message = "The GPU runners will be deprecated on December 31, 2025. All jobs using these runners should be migrated to other runner types."
-      command += <<~COMMAND
-        echo "::warning::#{message}" | sudo -u runner tee /home/runner/actions-runner/.ubicloud_complete_message
+      command << NetSsh.command(<<~COMMAND, message:)
+        echo "::warning::":message | sudo -u runner tee /home/runner/actions-runner/.ubicloud_complete_message
```
enescakir, line 338: I hope it's not replace `:warning` since we don't pass `{warning:.}` as argument

enescakir:
PR description updated, but the first commit has an old description.

CleanShot 2025-12-08 at 15 13 02@2x

prog/github/github_runner_nexus.rb
jeremyevans: Correct. The implementation substitutes based on the provided arguments, it doesn't scan the string and then look for entries. So `":foo :bar", foo: "baz quux"` results in `"baz\ quux :bar"`

jeremyevans:
Good catch on the commit message, I'll update it.

lib/kubernetes/client.rb
```diff
@@ -30,8 +30,9 @@ def load_balancer_hostname_missing?(svc)
     svc.dig("status", "loadBalancer", "ingress")&.first&.dig("hostname").to_s.empty?
   end
 
-  def kubectl(cmd)
-    output = @session.exec!("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf #{cmd}")
+  def kubectl(cmd, **)
+    raise "can only pass frozen string as kubectl cmd argument" unless cmd.frozen?
```
aamederen: is this a redundant check or something we can embed into `exec!`

jeremyevans: It's not currently redundant. We have to check that the string passed to `kubectl` is not created via interpolation. However, I think we can use `NetSsh.combine to do the combination and avoid the explicit check.

jeremyevans: Change made: https://github.com/ubicloud/ubicloud/pull/4283/commits/918f536298d541c9d136ad695b4621f0cc422ff5

fdr (approved):

PR #4284, opened 2025-12-02
Don't show clover-production in staging pry

jeremyevans:
This gives a false sense of urgency for staging. We should reserve this display only for the production environment.  The staging environment uses RACK_ENV=production to be as close as possible to the production. Support a STAGING=true environment variable that the staging environment can also use, and suppress the clover-production display in pry if set. We could potentially use Config.staging in additional cases in the future.

----

> [!IMPORTANT]
> Modify Pry prompt to show `clover-staging` in staging environments by adding a `staging` config option.
> 
>   - **Behavior**:
>     - Modify `Pry.config.prompt_name` in `bin/pry` to show `clover-staging` if `Config.staging` is true, otherwise follow existing logic.
>     - Introduce `staging` configuration in `config.rb` to differentiate between staging and production environments.
>   - **Config**:
>     - Add `override :staging, false, bool` in `config.rb` to support the new staging environment variable.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for fb6ed39669386259c1e9b16302e9ae6ef755c0ff. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bin/pry
+++ b/bin/pry
@@ -36,7 +36,9 @@ def udec(*)
 end
 
 opts = Pry::CLI.parse_options
-Pry.config.prompt_name = if Config.production?
+Pry.config.prompt_name = if Config.staging
+  "clover-staging"
+elsif Config.production?
   "\e[41m⚠️ %s\e[0m" % "clover-#{Config.rack_env}"
 else
   "clover-#{Config.rack_env}"
--- a/config.rb
+++ b/config.rb
@@ -56,6 +56,7 @@ module Config
   optional :hetzner_ssh_private_key, string, clear: true
   optional :hetzner_ssh_private_key_passphrase, string, clear: true
   optional :operator_ssh_public_keys, string
+  override :staging, false, bool
 
   # :nocov:
   override :mail_driver, (production? ? :smtp : :logger), symbol


bin/pry
```diff
@@ -36,7 +36,7 @@ def udec(*)
 end
 
 opts = Pry::CLI.parse_options
-Pry.config.prompt_name = if Config.production?
+Pry.config.prompt_name = if Config.production? && !Config.staging
   "\e[41m⚠️ %s\e[0m" % "clover-#{Config.rack_env}"
 else
   "clover-#{Config.rack_env}"
```
byucesoy, line 44: This only unsets the color, should we also print clover-staging instead clover-production?

byucesoy (approved):

bin/pry
jeremyevans: Yes, that sounds better. I'll get back to this after I finish up #4283.

jeremyevans: Pushed an updated commit for this.

PR #4289, opened 2025-12-02
Add stopped label to postgres server

enescakir:
We need to stop a few postgres resources every month due to fraudulent activity before permanently deleting the data.

Currently, we destroy the resource, since we can restore from backup if needed. However, it consumes some engineering time.

We can manually stop the resource on the dataplane, which causes an unavailability page, and users can't see that it's stopped in the UI.

A proper stop may require releasing CPU and memory resources, but the initial version only stops the PG process without releasing those resources.

This will help with operational tasks.

----

> [!IMPORTANT]
> Add 'stopped' state to Postgres servers, allowing them to be stopped without releasing resources, with corresponding logic and tests.
> 
>   - **Behavior**:
>     - Adds 'stopped' label to `display_state` in `postgres_resource.rb` for when `server_strand_label` is 'stopped'.
>     - Introduces `stop` functionality in `PostgresServer` and `PostgresServerNexus` classes, allowing servers to be stopped without releasing resources.
>   - **Methods**:
>     - Adds `when_stop_set?` and `hop_stopped` logic in `before_run` and `wait` methods in `postgres_server_nexus.rb`.
>     - Implements `stopped` method in `postgres_server_nexus.rb` to stop Postgres and PgBouncer services.
>   - **Tests**:
>     - Adds test for 'stopped' state in `display_state` in `postgres_resource_spec.rb`.
>     - Adds tests for `when_stop_set?` and `stopped` method in `postgres_server_nexus_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for fb8cab96e7d310e3fe7bbcb0f1d4b27e0331d505. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -47,6 +47,7 @@ class PostgresResource < Sequel::Model
     return "restoring_backup" if server_strand_label == "initialize_database_from_backup"
     return "replaying_wal" if ["wait_catch_up", "wait_synchronization"].include?(server_strand_label)
     return "finalizing_restore" if server_strand_label == "wait_recovery_completion"
+    return "stopped" if server_strand_label == "stopped"
     return "running" if ["wait", "refresh_certificates", "refresh_dns_record"].include?(strand.label) && !initial_provisioning_set?
 
     "creating"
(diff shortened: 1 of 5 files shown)


vivek7405:
> A proper stop may require releasing CPU and memory resources, but the initial version only stops the PG process without releasing those resources.

Are there any plans to implement a proper stop feature? Could we also introduce start/stop functionality for VMs with usage-based billing where we only charge for volume storage when a VM is stopped and release the allocated CPU and memory resources?

PR #4290, opened 2025-12-02
Lower quotas for gmail accounts

enescakir:
- **Add migration to add limited project reputation**
  

- **Add a "limited" project reputation level**
  Currently, we have two project reputation levels: verified and new. new
  is the default and has generous quotas compared to verified.
  
  To reduce fraud, we introduce a new limited level with lower quotas than
  new. Projects are assigned limited if their email domain is Gmail. We
  can improve this in the future by adding more generic email services to
  the list or other heuristics. This helps cover a significant portion of
  potential fraud cases.
  
  If the account has any verified projects, the reputation is not
  downgraded.
  

- **Apply hard runner concurrency limits for limited reputation users**
  Our current concurrency limits behave as soft limits: if we have enough
  idle capacity, users are allowed to exceed their configured concurrency.
  This creates a vulnerability where fraudulent users can spin up hundreds
  of large runners during low-load periods (e.g., weekends).
  
  We've introduced a "limited" reputation option for new users with
  generic email domains. For these users, we should enforce hard
  concurrency limits, meaning they cannot exceed their assigned
  concurrency even when spare capacity exists.
  
  Legitimate users can contact us to request a higher concurrency limit.
  

- **Show warning for limited reputation when hit hard concurrency limits**

### DIFF
--- a/spec/routes/web/project/github_spec.rb
+++ b/spec/routes/web/project/github_spec.rb
@@ -271,6 +271,17 @@ RSpec.describe Clover, "github" do
       expect(page.status_code).to eq(200)
       expect(page.title).to eq("Ubicloud - GitHub Runner Settings")
     end
+
+    it "shows concurrency warning for limited access accounts" do
+      installation.project.update(reputation: "limited")
+      Array.new(3).each {
+        runner = GithubRunner.create(installation_id: installation.id, repository_name: repository.name, label: "ubicloud-standard-60")
+        Strand.create_with_id(runner, prog: "Github::GithubRunnerNexus", label: "wait")
+      }
+      visit "#{project.path}/github/#{installation.ubid}/runner"
+      expect(page.status_code).to eq(200)
+      expect(page).to have_content "You've reached your vCPU concurrency limit"
+    end
   end
 
   describe "cache" do
--- a/views/github/runner.erb
+++ b/views/github/runner.erb
@@ -4,6 +4,15 @@
 
 <%== render("github/tabbar") %>
 
+<% if @installation.project.reputation == "limited" && !@installation.project.quota_available?("GithubRunnerVCpu", 0) %>
+  <div class="mb-3 -mt-2 rounded-md bg-red-50 p-4 flex items-center gap-3">
+    <%== part("components/icon", name: "hero-squares-plus", classes: "h-5 w-5 text-red-400") %>
+    <h3 class="text-sm font-medium text-red-800">
+      You've reached your vCPU concurrency limit. If you need higher limits, please contact our support team at support@ubicloud.com
+    </h3>
+  </div>
+<% end %>
+
 <div class="grid gap-6">
   <%
     free_upgrade_badge = if @installation.free_runner_upgrade?


migrate/20251202_add_limited_reputation.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    add_enum_value(:project_reputation, "limited")
```
jeremyevans: This is not a reversible change. We should switch the columns using this enum to text, and add check constraints. If you would like to continue using an enum, `last_irreversible_migration` in the `Rakefile` needs to be updated.

enescakir: I wasn't aware that enum values couldn't be dropped.

I found a few examples where we had already changed enum columns to text in the past, so I migrated the reputation column to text as well.

migrate/20251202_add_limited_reputation.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    # Add temporary VARCHAR column
+    alter_table(:project) do
+      add_column :reputation_new, String, collate: '"C"', default: "new"
```
jeremyevans: Instead of add/drop/rename column, you can change the column type, update the default, and  add the check constraint:

```suggestion
      set_column_type :reputation, "text collate \"C\"", using: Sequel.cast(:reputation, String)
      set_column_default :reputation, "new"
      add_constraint :reputation_check, reputation: %w[new verified limited]
```

Currently, you need the `"text collate \"C\""` because Sequel doesn't recognize the `:collate` option for `set_column_type`. I'll fix that.

model/account.rb
```diff
@@ -19,8 +19,9 @@ def provider_names
     identities.map(&:provider).join(", ")
   end
 
-  def create_project_with_default_policy(name, default_policy: true)
-    project = Project.create(name: name)
+  def create_project_with_default_policy(name, reputation: "new", default_policy: true)
+    reputation = "limited" if email.end_with?("@gmail.com") && projects.none? { it.reputation == "verified" }
```
byucesoy: Can we add other free/disposable email providers here as well? Something like [this](https://github.com/disposable-email-domains/disposable-email-domains)?

enescakir: Yeah, we can improve the logic. Most of the fraud is coming from `gmail.com`, so the current check catches a big chunk of cases. But we can extend the list to include more free email providers. Let me add a few more

Disposable emails are a separate category. IMHO we shouldn't allow accounts to be created with disposable addresses at all. That check should happen at account creation, not project creation. We can handle that in a separate PR.

migrate/20251202_add_limited_reputation.rb
enescakir: I updated the migration

model/account.rb
enescakir: Updated list: `%w[gmail.com outlook.com hotmail.com yahoo.com icloud.com protonmail.com proton.me]`

migrate/20251209_add_limited_reputation.rb
```diff
@@ -0,0 +1,27 @@
...
+      set_column_default :reputation, "new"
+      add_constraint :reputation_check, reputation: %w[new verified limited]
+    end
+    drop_enum :project_reputation
+  end
+
+  down do
+    create_enum(:project_reputation, %w[new verified])
+
+    run "UPDATE project SET reputation = 'new' WHERE reputation = 'limited'"
```
jeremyevans: In general, I think we should reserve raw SQL usage for cases that Sequel's API doesn't support (or are significantly clunkier when using Sequel's API):

```suggestion
    from(:project).where(reputation: "limited").update(reputation: "new")
```

migrate/20251209_add_limited_reputation.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    alter_table(:project) do
+      drop_constraint :reputation_check
+      set_column_default :reputation, Sequel.cast("new", :project_reputation)
```
jeremyevans, line 20: I would move this after the the `set_column_type`. I'm not sure how PostgreSQL handles this, but I believe it adds an implicit type cast to the column's current type, so it's better to change the default after the type instead of before.

enescakir: When I moved, it failed with the following exception:

```
PG::DatatypeMismatch: ERROR:  default for column "reputation" cannot be cast automatically to type project_reputation (PG::DatatypeMismatch)
```

jeremyevans: Interesting. Is this true if you just have `"new"` instead of `Sequel.cast("new", :project_reputation)?` I assume that should work, since the column was originally defined with:

```ruby
add_column :reputation, :project_reputation, default: "new", null: false
```

enescakir: When I remove the `set_column_default` (the default is `new::text` from the up migration) or change it to `set_column_default :reputation, "new"` in the down migration, it fails with the same exception.

jeremyevans: Interesting. I wonder why it works in the initial migration that added the column, but not after the column type is changed back. Not worth further analysis though.

PR #4291, opened 2025-12-02
Move github models to github folder

enescakir:
- **Move github models to github folder**
  We initially placed all models in the models folder. Then we started
  grouping them by functionality. This change moves all remaining
  GitHub-related models to the github folder.

### DIFF
--- a/model/github/github_installation.rb
+++ b/model/github/github_installation.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require_relative "../model"
+require_relative "../../model"
 
 class GithubInstallation < Sequel::Model
   many_to_one :project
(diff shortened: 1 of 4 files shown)


jeremyevans (approved):
This makes sense. My preference would be for not using `model` subdirectories, since we aren't using namespaces, but if we are going to use `model` subdirectories, we should at least be consistent about where models are placed.

enescakir:
I agree that the current approach using directories without namespaces only helps a little with organization.

Since the GitHub service is one of our older services, it doesn’t use model subdirectories, while the newer services do. I made this change to make it consistent with the others.

At some point, we might consider using namespaces more. For example, `Github::Repository` (though might not work with Sequel or table names well) instead of `GithubRepository`, or `Prog::Github::RunnerNexus` instead of `Prog::Github::GithubRunnerNexus`.

PR #4292, opened 2025-12-03
Add Custom/SafeShellCommand RuboCop cop

fdr:
This cop detects nearly-unsafe shell command patterns in `cmd()` calls using Shellwords identity checking. It rewrites certain interpolated strings into a %W[] construction designed to be passed to shelljoin. It has a focus on heuristic autocorrection, to spare our fingers.

A simple one might be:

    cmd("ls #{foo}")

Where the code is rewritten to:

    cmd(%W[ls #{foo}])

With the idea that the array should be shelljoined, thus, all its argv are safely escaped.

And slightly more complex case is:

    cmd("ls directory/#{foo}")

In such a case, it's presumed that `directory/#{foo}` are to be escaped together:

    cmd(%W[ls directory/#{foo}])

An example of a mistaken one is:

   cmd(%W[ls /vm/#{q_vm}])

Where the `q_` prefix means "quoted for shell".  That could be added, or, perhaps with the convenience of this rewrite, simplifying to:

   cmd(%W[ls /vm/#{vm}])

Will become the new way.

Among the more complex (maybe not even worth it) ones were handling multi-line continuations:

    @nic.vm.vm_host.sshable.cmd("sudo -- xargs -I {} -- ip -n #{@namespace} xfrm state add " \
      "src #{src} dst #{dst} proto esp spi #{spi} reqid #{@reqid} mode tunnel " \
      "aead 'rfc4106(gcm(aes))' {} 128 #{"sel src 0.0.0.0/0 dst 0.0.0.0/0" if is_ipv4}", stdin: key)

Preserving the line breaks and noting the spaces at the end of lines to test for safety in newline preservation in %W[]:

    @nic.vm.vm_host.sshable.cmd(%W[sudo -- xargs -I {} -- ip -n #{@namespace} xfrm state add
      src #{src} dst #{dst} proto esp spi #{spi} reqid #{@reqid} mode tunnel
      aead 'rfc4106(gcm(aes))' {} 128 #{"sel src 0.0.0.0/0 dst 0.0.0.0/0" if is_ipv4}], stdin: key)

The tests contain many, many edge cases to reach 100% branch coverage, plus there are some assertions for topologies that appear unreachable for now.

### DIFF
--- a/rubocop/cop/custom/safe_shell_command.rb
+++ b/rubocop/cop/custom/safe_shell_command.rb
@@ -22,7 +22,7 @@ module RuboCop
           "Avoid interpolation in shell commands. This structure (heredoc or mid-word concatenation) " \
           "cannot be safely autocorrected to `%W[]`. Please refactor manually."
 
-        def_node_matcher :cmd_call?, "(send !const :cmd $...)"
+        def_node_matcher :cmd_call?, "(send !const {:cmd :exec!} $...)"
 
         def_node_matcher :wrapped_shell_escape?, <<~PATTERN
           (begin


fdr:
Got somewhat nerdsniped by this. It will rewrite 72% of our `cmd` expressions, some with known problems (like q_ being redundant). It doesn't implement .shelljoin in sshable.cmd and it doesn't have a fix for test methodology to (I presume) post-process the string regardless of frontend before doing the assertion (I assume that's something you want to do in some form @jeremyevans )

But I wanted to make something tangible in case it affected your calculation how the shape of development would end up. If in fact we can get 72% of our stuff through shelljoin, that's pretty good at not hurting too badly. Without stretching your pinky, you can write: `sshable.cmd("foo #{bar} baz")` and get it auto-popped to `sshable.cmd(%W[foo #{bar} baz])`

Among trickier things is heredocs (I don't do anything with them), plus pipelines and redirections (as expected). I could add relaxations to this (e.g. shelljoin except for...) but I figured we'd start with the simpler ones.

fdr:
another thing we should probably do is axe daemonizer version 1, which has an argument order that makes things way, way more complex than necessary. version 2 fixes that and has some support symbols to boot. I think they have a slightly different relationship to the systemd journal, with version 2 intending to subsume 1, but it will require a change in practice.

jeremyevans:
The fact that this will automatically rewrite things but potentially not do so correctly (resulting in double escaping in a case like `q_vm` that already escapes) leads me to be hesitant to recommend its use on an on-going basis. Rubocop doesn't know whether what is being interpolated has already been shell escaped in some manner. I see `wrapped_shell_escape?`, but that appears to only handle direct use of `shellescape`, and not stuff like `q_*`.

The rewriter may be useful to do the heavy lifting during a one-time conversion, as long as the resulting changes are carefully manually checked and all incorrect conversions are fixed. I don't think keeping this as an automatic rewriter is a good idea. That being said, it may be useful to keep the cop without the rewriter, as a secondary check to ensure insecure interpolation is not used.

fdr:
I think it might be okay with test-time validations, though we can tweak to
practical effect: it shouldn’t be wrong too often. I could encode the q_
pattern, but I suspect it’s mostly a move to make things terse, and can be
obsoleted by changing the code.

Ultimately I don’t write cmd(…) as often as many others, they can
adjudicate on ongoing residuals…

The way I think of this: it turns cmd into a macro through the back door
(rewriting the source). We will definitely need an “unsafe_cmd” or whatever
you had in mind, to opt out of the macro and get something else (e.g.
nothing, or a special templating mode, whatever)

fdr:
Incidentally, rubocop has many unsafe rewrites, and I always use them, and
on rare occasion have to change code to get around them (often, the code
that tricks them tends to be a bit weird). I always commit code if its
complex before running rubocop, check the diff, and lean on tests to
identify the occasional problematic rewrite. This last part is the most
incomplete part of the patch, because I want to see that you have in mind
in the test verification area.

(One prominent but not exactly comparable example is frozen string literals, auto-introduced at our option by rubocop)

fdr:
long obsolete

PR #4295, opened 2025-12-03
Improve k8s vm ip allocation

mohi-kalantari:
**Do not use the reserved ips of Vm range for kubernetes nodes**

Kubernetes Vms have bigger subnets and each gets a /26 ipv4 range in order to provision pods using the CNI.

Prior to this commit, we were using the vm_net4.nth(1) for the Vm IP which is wrong and the first 4 ips should not be used.

First ip of the range is the network address Second ip should be presumed as gateway The third and fourth ips are reserved for future proofing.

For the fix, we would call the vm_net4.nth(4) when cidr is not /32.

**Integrate VM IP and Reserved Range into CNI Allocation**

This commit updates the CNI's IP allocation logic to strictly reserve the five initial addresses of the allocated subnet slice,
 ensuring consistency with the host's IP setup and network rules.

The logic now calculates and includes the following IPs in the reserved list before any random IP selection for containers or the gateway:

- Network Address (Offset 0)
- Gateway (Offset 1)
- 2 Reserved IPs (Offsets 2, 3)
- VM's Primary IP (Offset 4), which is deterministically the first usable host address.

By reserving the VM's primary IP explicitly, we prevent the CNI pool from accidentally assigning it to a container, resolving a potential IP conflict. The CNI's random allocation now correctly starts from the 6th IP (Offset 5) onward.

----

> [!IMPORTANT]
> Reserve initial IPs in Kubernetes VM subnets and update CNI allocation to prevent conflicts.
> 
>   - **Behavior**:
>     - Reserve first 5 IPs in Kubernetes VM subnets to prevent conflicts, starting allocation from the 6th IP.
>     - Update CNI allocation logic in `ubi_cni.rb` to respect reserved IPs.
>   - **Code Changes**:
>     - Change `private_ipv4` in `vm.rb` to use `ipv4.nth(4)` instead of `ipv4.nth(1)`.
>     - Modify `random_private_ipv4` in `private_subnet.rb` to account for reserved IPs.
>     - Remove `update_via_routes` calls in `vm_setup.rb` and adjust IP setup logic.
>   - **Tests**:
>     - Update tests in `vm_spec.rb` and `provision_kubernetes_node_spec.rb` to reflect new IP allocation logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 74f0e02562bca49a75736774b222fae98e188abe. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/kubernetes/lib/ubi_cni.rb
+++ b/rhizome/kubernetes/lib/ubi_cni.rb
@@ -121,10 +121,19 @@ options ndots:5
       subnet_ipv6_obj = IPAddr.new(subnet_ipv6)
       subnet_ipv4_obj = IPAddr.new(subnet_ipv4)
 
+      network_int = subnet_ipv4_obj.to_i
+      reserved_ipv4_ips = [
+        (network_int + 0), # Network Address
+        (network_int + 1), # Gateway
+        (network_int + 2), # Reserved 2
+        (network_int + 3), # Reserved 3
+        (network_int + 4)  # Vm's IP
+      ].map { |ip_int| IPAddr.new(ip_int, Socket::AF_INET).to_s }
+
       container_ula_ipv6 = find_random_available_ip(allocated_ips, subnet_ula_ipv6_obj)
       container_ipv6 = find_random_available_ip(allocated_ips, subnet_ipv6_obj)
-      ipv4_container_ip = find_random_available_ip(allocated_ips, subnet_ipv4_obj)
-      ipv4_gateway_ip = find_random_available_ip(allocated_ips, subnet_ipv4_obj, reserved_ips: [ipv4_container_ip.to_s])
+      ipv4_container_ip = find_random_available_ip(allocated_ips, subnet_ipv4_obj, reserved_ips: reserved_ipv4_ips)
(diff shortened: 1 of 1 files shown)


mohi-kalantari:
I worked on a this PR in order to standardize our k8s networking stack by reserving the first ip of the ipv4 range we assign to a vm as the gateway and use it in the dnsmasq. Also propagate the changes to the CNI to account for the newly reserved ips.
To explain more, we would assign a /26 ip range (e.g. 172.27.13.0/26) to a vm, then assign the .nth(1) to the vm (172.27.13.1). It seems wrong at the first glance and might not be backward compatible with some legacy tools.
Our current approach kept bugging me and I wanted to change it since on this planning cycle, we were putting some effort to fix our known issues and I thought it can be done in a small time box.
I talked with Furkan and did some research, there is no must to reserve those ips or have a dnsmasq gateway.
As a result, I'm closing the PR for now and won't proceed with it since keeping it backward compatible might take more time

PR #4297, opened 2025-12-03
Allow column encryption key to be itself encrypted with KMS

serprex:
Avoids leaking key if environment is leaked

KMS decryption is an API call so not scalable to being used for actual SSH key encryption/decryption

----

> [!IMPORTANT]
> Adds AWS KMS decryption for column encryption key in `model.rb` using `aws-sdk-kms` if configured.
> 
>   - **Behavior**:
>     - Adds decryption of `clover_column_encryption_key` using AWS KMS in `model.rb` if `kms_encrypt_clover_column_encryption_key` is true.
>     - Uses `aws-sdk-kms` gem for KMS decryption.
>   - **Configuration**:
>     - Adds `kms_encrypt_clover_column_encryption_key` as a boolean override in `config.rb`.
>   - **Dependencies**:
>     - Adds `aws-sdk-kms` gem under `aws_kms` group in `Gemfile`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c24994641118250c0a85fc33c0d372826efc95a4. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -44,6 +44,7 @@ module Config
   mandatory :rack_env, string
 
   optional :clover_runtime_token_secret, base64, clear: true
+  optional :kms_decrypt_clover_column_encryption_key_with_arn, string
   optional :heartbeat_url, string
   optional :clover_database_root_certs, string
   override :max_health_monitor_threads, 32, int
--- a/model.rb
+++ b/model.rb
@@ -12,7 +12,15 @@ Sequel::Model.plugin :require_valid_schema
 Sequel::Model.plugin :singular_table_names
 Sequel::Model.plugin :subclasses unless ENV["RACK_ENV"] == "development"
 Sequel::Model.plugin :column_encryption do |enc|
-  enc.key 0, Config.clover_column_encryption_key
+  key = Config.clover_column_encryption_key
+  if Config.kms_decrypt_clover_column_encryption_key_with_arn
+    require "aws-sdk-kms"
+    kms_client = Aws::KMS::Client.new
+    ciphertext_blob = Base64.decode64(key)
+    response = kms_client.decrypt(ciphertext_blob:, key_id: Config.kms_decrypt_clover_column_encryption_key_with_arn)
+    key = response.plaintext
+  end
+  enc.key 0, key
 end
 Sequel::Model.plugin :many_through_many
 Sequel::Model.plugin :insert_conflict
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,10 @@ gem "tilt", ">= 2.6.1"
 gem "warning"
 gem "webauthn"
 
+group :aws_kms do
+  gem "aws-sdk-kms"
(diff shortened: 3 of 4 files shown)


config.rb
```diff
@@ -38,14 +38,13 @@ def self.test?
     Config.rack_env == "test"
   end
 
-  # Mandatory -- exception is raised for these variables when missing.
```
jeremyevans, line 41: I don't object to removing these comments, but it is unrelated to the other changes, so it should be a separate commit.

model.rb
```diff
@@ -12,7 +12,15 @@
 Sequel::Model.plugin :singular_table_names
 Sequel::Model.plugin :subclasses unless ENV["RACK_ENV"] == "development"
 Sequel::Model.plugin :column_encryption do |enc|
-  enc.key 0, Config.clover_column_encryption_key
+  key = Config.clover_column_encryption_key
+  unless Config.kms_encrypt_clover_column_encryption_key_region.nil?
```
jeremyevans: `nil`/`false` difference is not important in this case, so it's better to be idiomatic:

```suggestion
  if Config.kms_encrypt_clover_column_encryption_key_region
```

model.rb
```diff
@@ -12,7 +12,15 @@
...
+    require 'aws-sdk-kms'
+    kms_client = Aws::KMS::Client.new(region: Config.kms_encrypt_clover_column_encryption_key_region)
```
jeremyevans: Looking at `Aws::KMS::Client` documentation (https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/KMS/Client.html#initialize-instance_method), this seems to be relying on `:credentials` being passed implicitly via environment variables, files, or `Aws.config`. However, this passes `:region` explicitly (even though `:region` can also be passed implicitly). Is there a reason for this difference?

If the credentials are passed via environment variables, then it seems this wouldn't actually fix the issue, only require an extra step, since an attacker with access to the environment variables would be able to call `Aws::KMS::Client` with the environment variables to get the decryption key.

iamKunalGupta: > However, this passes :region explicitly (even though :region can also be passed implicitly). Is there a reason for this difference?

Yes, we should remove region here

iamKunalGupta: > If the credentials are passed via environment variables, then it seems this wouldn't actually fix the issue, only require an extra step, since an attacker with access to the environment variables would be able to call Aws::KMS::Client with the environment variables to get the decryption key.

In case of environments like kubernetes (and potentially ECS), the environment may be leaked but the aws auth itself can be tied to service accounts which are bound to the execution (like IRSA/Pod Identity) and would require additional exec permissions into the container

PR #4305, opened 2025-12-04
PostgresTimeline related fixes

byucesoy:
**Don't skip backup call when leader is nil**
Even when the leader is nil, there could be backups until the retention period
is over. Therefore, we should not skip the backup call in such cases.

**Check for all dependents not just leader of timeline**
Leader can be nil, but in case of read replicas, there can still be servers
dependent on this timeline. So we should check for existence of any dependent,
not just the leader.

----

> [!IMPORTANT]
> Improve timeline handling by checking for any dependents and ensuring backups are not skipped when the leader is nil in `postgres_timeline_nexus.rb`.
> 
>   - **Behavior**:
>     - In `postgres_timeline_nexus.rb`, change `wait` method to check for any dependents using `PostgresServer[timeline_id: postgres_timeline.id]` instead of just the leader.
>     - Ensure backups are checked even if the leader is nil, preventing premature timeline destruction.
>   - **Tests**:
>     - Update `wait` method tests in `postgres_timeline_nexus_spec.rb` to reflect changes in dependent checking logic.
>     - Remove test case that avoided API calls for backups when there is no leader.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 8ecbf2fec7f69393478e0200a1ea93113497486a. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_timeline_nexus.rb
+++ b/prog/postgres/postgres_timeline_nexus.rb
@@ -62,9 +62,9 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
   end
 
   label def wait
-    leader = postgres_timeline.leader
+    dependent = PostgresServer[timeline_id: postgres_timeline.id]
     backups = postgres_timeline.backups
-    if leader.nil? && backups.empty? && Time.now - postgres_timeline.created_at > 10 * 24 * 60 * 60
+    if dependent.nil? && backups.empty? && Time.now - postgres_timeline.created_at > 10 * 24 * 60 * 60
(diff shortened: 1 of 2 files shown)


serprex (approved):

furkansahin:
you just need to rebase, these tests are fixed in the main

PR #4306, opened 2025-12-04
[destroying #1] Increment destroying semaphore when hopped to destroy label

enescakir:
- **Rename destroy label at DestroyGithubInstallation and Test::HetznerServer**
  It's different than our other destroy flows. It doesn't have destroy
  semaphore and subject. So I renamed it.
  

- **Increment destroying semaphore when hopped to destroy label**
  If the resource is already being destroyed, we don't want to hop to the
  `destroy` label again when the `destroy` semaphore is incremented.
  
  We check the current label to avoid hopping again. Sometimes we forget
  to update the list of labels to check when we add new labels after a
  destroy.
  
  Instead of checking the labels separately, we can increase another
  semaphore, like `destroying`, along with the `destroy` semaphore and
  keep it incremented until the resource is destroyed.
  
  We incremented destroying semaphore autmatically when `hop_destroy`
  called. So if the operator calls it via pry, it will be incremented too.
  

----

> [!IMPORTANT]
> Enhance semaphore handling and rename destroy labels for consistency in destroy processes.
> 
>   - **Behavior**:
>     - Increment `destroying` semaphore when `hop_destroy` is called in `base.rb` to prevent re-hopping to `destroy` label.
>     - Automatically increment `destroying` semaphore when `hop_destroy` is called, even if invoked manually.
>   - **Renames**:
>     - Rename `destroy` label to `wait_resource_destroy` in `destroy_github_installation.rb`.
>     - Rename `destroy` label to `destroy_vm_host` in `hetzner_server.rb`.
>   - **Tests**:
>     - Update tests in `inference_endpoint_nexus_spec.rb`, `inference_endpoint_replica_nexus_spec.rb`, and `inference_router_replica_nexus_spec.rb` to expect `incr_destroying` call.
>     - Update tests in `destroy_github_installation_spec.rb` and `hetzner_server_spec.rb` to reflect label renames.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9ed902720324f0100a0af4b8ed22e9205a7fa5a8. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/private_subnet.rb
+++ b/model/private_subnet.rb
@@ -63,7 +63,7 @@ class PrivateSubnet < Sequel::Model
     (state == "waiting") ? "available" : state
   end
 
-  plugin SemaphoreMethods, :destroy, :destroying, :refresh_keys, :add_new_nic, :update_firewall_rules, :migrate
+  plugin SemaphoreMethods, :destroy, :refresh_keys, :add_new_nic, :update_firewall_rules, :migrate
 
   def self.random_subnet
     subnet_dict = PRIVATE_SUBNET_RANGES.each_with_object({}) do |subnet, hash|
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -32,6 +32,7 @@ end
 
   def self.semaphore(*names)
     names.map!(&:intern)
+    names << :destroying if names.include?(:destroy)
     names.each do |name|
       define_method :"incr_#{name}" do
         @snap.incr(name)
@@ -58,8 +59,15 @@ end
   def self.label(label)
     (@labels ||= []) << label
 
-    define_method :"hop_#{label}" do
-      dynamic_hop label
+    if label == :destroy
+      define_method :"hop_#{label}" do
+        incr_destroying
+        dynamic_hop label
+      end
+    else
+      define_method :"hop_#{label}" do
+        dynamic_hop label
+      end
     end
   end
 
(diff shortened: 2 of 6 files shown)


enescakir:
The previous PR was very outdated and had a bunch of conflicts with the recent refactors, so I created a new one based on the latest main. I'll update the `before_run` logic in smaller follow-up PRs.

prog/github/destroy_github_installation.rb
```diff
@@ -36,10 +36,10 @@ def self.assemble(installation)
       runner.incr_skip_deregistration
       runner.incr_destroy
     end
-    hop_destroy
+    hop_wait_resource_destroy
   end
 
-  label def destroy
+  label def wait_resource_destroy
```
jeremyevans, line 42: This is unsafe if there is already a strand at this label. You should keep the `hop_destroy` -> `hop_wait_resource_destroy`, and add a `destroy` label that does `hop_wait_resource_destroy`. Then later, after you've checked there is no strand using the label, or updated the strands that were using the label, you can remove the `destroy` label. Similar fix needed in the other prog.

enescakir: Yeah, normally removing a label is an unsafe operation and should be done in multiple steps. But this prog is used very rarely (maybe once a week) so the chance of a straggling instance hitting destroy right during deployment is extremely low. It might not be worth the extra effort to do a multi step deployment.

I'll verify after deployment to make sure it's not the case. But if you think that even this small risk isn't acceptable and we should follow the standard procedure, I can update the commit accordingly.

jeremyevans: You have a much better idea of which progs are run when than I do, so if you think the risk is low, it seems fine.

PR #4310, opened 2025-12-04
postgres: walg-daemon-client

serprex:
only install client & run daemon for now,
changing archive_command deferred until fleet can assume daemon

another option would be to use https://github.com/wal-g/walg_archive
configuring archive_library instead, but this library doesn't seem to've been updated since PG15

----

> [!IMPORTANT]
> Add `walg-daemon-client` installation and systemd service setup, and update WAL-G configuration to include version-specific data paths.
> 
>   - **Behavior**:
>     - Install and run `walg-daemon-client` in `install-wal-g` script.
>     - Create and enable `wal-g.service` for systemd to manage the daemon.
>   - **Configuration**:
>     - Modify `generate_walg_config` in `postgres_timeline.rb` to include `PGDATA` path with version.
>     - Update `refresh_walg_credentials` in `postgres_server.rb` to use versioned `walg_config`.
>   - **Tests**:
>     - Update tests in `postgres_timeline_spec.rb` to check versioned `walg_config` generation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 74cf60ce4a149c2f4321928adb0f4299718d9a12. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/rhizome/postgres/bin/install-wal-g
+++ b/rhizome/postgres/bin/install-wal-g
@@ -27,4 +27,24 @@ Dir.chdir("var/wal-g") do
   r "make deps"
   r "make pg_build"
   r "GOBIN=/usr/bin make pg_install"
+
+  # Compile and install walg-daemon-client
+  r "make build_client"
+  r "cp bin/walg-daemon-client /usr/bin/walg-daemon-client"
 end
+
+File.write("/etc/systemd/system/wal-g.service", <<WALG_SERVICE)
+[Unit]
+Description="WAL-G daemon"
+After=network.target
+
+[Service]
+Type=simple
+User=postgres
+ExecStart=/usr/bin/wal-g daemon --config /etc/postgresql/wal-g.env /tmp/wal-g
+
+[Install]
+WantedBy=multi-user.target
+WALG_SERVICE
+
+r "systemctl enable wal-g.service --now"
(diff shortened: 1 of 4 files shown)


serprex:
https://github.com/wal-g/wal-g/blob/6501e763b621931069f305e9f5ebe823abf2091b/internal/databases/postgres/daemon_handler.go#L84
daemon client waits for daemon to respond with success/fail. benefit of daemon is smaller binary to spawn & daemon has long lived connection to s3 shared across client requests

rhizome/postgres/bin/install-wal-g
```diff
@@ -27,4 +27,24 @@ Dir.chdir("var/wal-g") do
   r "make deps"
   r "make pg_build"
   r "GOBIN=/usr/bin make pg_install"
+
+  # Compile and install walg-daemon-client
+  r "make build_client"
+  r "cp bin/walg-daemon-client /usr/bin/walg-daemon-client"
 end
+
+File.write("/etc/systemd/system/wal-g.service", <<WALG_SERVICE)
```
serprex, line 36: I think this rest of this file can be done by AMI instead (which also has to be updated to build/install daemon client)

serprex:
replaced by #4345

PR #4311, opened 2025-12-05
[destroying #2]  Move before_run to the base prog

enescakir:
When the destroy semaphore is increased, we want to hop to the destroy label from any label. If the prog defines a before_run method, we call it just before running each label. Some of these methods have custom logic, but most of them are similar. To achieve 100% line coverage, we test similar functionality repeatedly. I’ve moved this to the base program. So, if the prog doesn’t override before_run and has a destroy label and semaphore, we call this basic flow in all progs.

The custom logics in progs also follow similar patterns. We can gradually move them to the base prog as well.

----

> [!IMPORTANT]
> Centralize `before_run` logic in `Prog::Base`, removing redundant methods across multiple program classes and updating tests accordingly.
> 
>   - **Behavior**:
>     - Move `before_run` logic to `Prog::Base` class, centralizing destroy semaphore handling.
>     - Remove redundant `before_run` methods from `inference_endpoint_nexus.rb`, `inference_router_nexus.rb`, `github_repository_nexus.rb`, `kubernetes_node_nexus.rb`, `kubernetes_nodepool_nexus.rb`, `postgres_timeline_nexus.rb`, `host_nexus.rb`, `vm_host_slice_nexus.rb`, `cert_nexus.rb`, `nic_nexus.rb`, `subnet_nexus.rb`, and `test.rb`.
>   - **Tests**:
>     - Update tests in `inference_endpoint_nexus_spec.rb`, `inference_router_nexus_spec.rb`, `base_spec.rb`, `github_repository_nexus_spec.rb`, `kubernetes_node_nexus_spec.rb`, `kubernetes_nodepool_nexus_spec.rb`, `postgres_timeline_nexus_spec.rb`, `host_nexus_spec.rb`, `vm_host_slice_nexus_spec.rb`, `cert_nexus_spec.rb`, `nic_nexus_spec.rb`, and `subnet_nexus_spec.rb` to reflect the centralized `before_run` logic.
>   - **Misc**:
>     - Ensure all programs with a destroy label and semaphore utilize the base `before_run` logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for ccb810b420062268fb1cb17e57b9e0197d7299d8. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/github/github_repository_nexus.rb
+++ b/prog/github/github_repository_nexus.rb
@@ -126,15 +126,6 @@ class Prog::Github::GithubRepositoryNexus < Prog::Base
     end
   end
 
-  def before_run
-    when_destroy_set? do
-      if strand.label != "destroy"
-        register_deadline(nil, 5 * 60)
-        hop_destroy
-      end
-    end
-  end
-
   label def wait
     cleanup_cache if github_repository.access_key
     nap 15 * 60 if Time.now - github_repository.last_job_at > 6 * 60 * 60
@@ -157,6 +148,7 @@ class Prog::Github::GithubRepositoryNexus < Prog::Base
   label def destroy
     decr_destroy
 
+    register_deadline(nil, 5 * 60)
     unless github_repository.runners.empty?
       Clog.emit("Cannot destroy repository with active runners") { {not_destroyed_repository: {repository_name: github_repository.name}} }
       nap 5 * 60
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -207,7 +207,7 @@ SQL
     DB.transaction do
       SemSnap.use(id) do |snap|
         prg = load(snap)
-        prg.public_send(:before_run) if prg.respond_to?(:before_run)
+        prg.public_send(:before_run)
         prg.public_send(label)
       end
     rescue Prog::Base::Nap => e
--- a/prog/ai/inference_endpoint_nexus.rb
+++ b/prog/ai/inference_endpoint_nexus.rb
@@ -83,14 +83,6 @@ class Prog::Ai::InferenceEndpointNexus < Prog::Base
     end
   end
 
-  def before_run
-    when_destroy_set? do
-      if strand.label != "destroy"
-        hop_destroy
-      end
-    end
-  end
-
   label def start
     reconcile_replicas
     register_deadline("wait", 10 * 60)
--- a/prog/ai/inference_router_nexus.rb
+++ b/prog/ai/inference_router_nexus.rb
@@ -49,14 +49,6 @@ class Prog::Ai::InferenceRouterNexus < Prog::Base
     end
   end
 
-  def before_run
-    when_destroy_set? do
-      if strand.label != "destroy"
-        hop_destroy
-      end
-    end
-  end
-
   label def start
     reconcile_replicas
     register_deadline("wait", 10 * 60)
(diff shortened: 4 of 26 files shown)


prog/github/github_repository_nexus.rb
```diff
@@ -157,6 +148,7 @@ def before_run
   label def destroy
     decr_destroy
 
+    register_deadline(nil, 5 * 60)
```
jeremyevans, line 151: Does this continually push the deadline out if the later `nap` is called? If so, it seems like a behavior change. Maybe it's expected or desired, but moving this from `before_run` should probably be put into it's own commit.

Similar issue in some other progs.

jeremyevans (commented):
This is an impressive reduction in redundant lines.  I worry that this makes it easier to miss making the appropriate updates if you have `destroy` call another label. The strand label checking approach is fundamentally fragile.  The `destroying` semaphore approach fixes that issue.

Maybe we can have `hop` fail if the current `label` is `destroy` and the there isn't already a `destroying` semaphore set? That makes the strand label checking approach much less fragile, and forces the developer to address the situation by adding a `destroying` semaphore. WDYT?

prog/github/github_repository_nexus.rb
enescakir: Since `allow_extension` is false by default, it doesn't push the deadline. Without extension, `register_deadline` is idempotent, doesn't change the current deadlines.

https://github.com/ubicloud/ubicloud/blob/2adb48e4fdfe4e274bb2d79eea2dd2fc0ff94158/prog/base.rb#L322

A known issue with `register_deadline` is when it raises an unexpected exception. Since it doesn't save the changes, it also doesn't register deadlines. So if the operation failed continuously, we don't get deadline pages/

CleanShot 2025-12-05 at 22 49 15@2x

We run `public_send(:before_run)` and `public_send(label)` in the same transaction, so it doesn't change the effect of this issue.

enescakir:
> This is an impressive reduction in redundant lines. I worry that this makes it easier to miss making the appropriate updates if you have `destroy` call another label. The strand label checking approach is fundamentally fragile. The `destroying` semaphore approach fixes that issue.
> 
> Maybe we can have `hop` fail if the current `label` is `destroy` and the there isn't already a `destroying` semaphore set? That makes the strand label checking approach much less fragile, and forces the developer to address the situation by adding a `destroying` semaphore. WDYT?

What do you think about the check I added in the third commit of the next PR?

https://github.com/ubicloud/ubicloud/pull/4316

enescakir:
> This is an impressive reduction in redundant lines

Hold on for the next PRs 🔥 I'll be removing a lot more lines in those.

jeremyevans:
> > This is an impressive reduction in redundant lines. I worry that this makes it easier to miss making the appropriate updates if you have `destroy` call another label. The strand label checking approach is fundamentally fragile. The `destroying` semaphore approach fixes that issue.
> > Maybe we can have `hop` fail if the current `label` is `destroy` and the there isn't already a `destroying` semaphore set? That makes the strand label checking approach much less fragile, and forces the developer to address the situation by adding a `destroying` semaphore. WDYT?
> 
> What do you think about the check I added in the third commit of the next PR?
> 
> #4316

That looks good.

jeremyevans (approved):
This looks good, but we should follow up with #4316 soon.

PR #4312, opened 2025-12-05
Add AWS AZs as a new column

furkansahin:
## Backfill AZs field if it's not already set
AWS doesn't have the same number of AZs in every region. Knowing they
have at least 3, we were making use of that fact and not using the whole
region sometimes. This information can be fetched from the AWS. In this
commit, we are adding the method to basically return the database stored
AZs information if it is filled, if not, we are fetching it from the
AWS.

## Use the AZs field of the location instead of hard coded array

----

> [!IMPORTANT]
> Add AWS Availability Zones as a new column in the database and update code to use it instead of hardcoded values.
> 
>   - **Database Migration**:
>     - Adds `location_aws_az` table in `20251206_add_location_azs.rb` with columns `id`, `location_id`, `az`, and `zone_id`.
>   - **Model Changes**:
>     - Adds `LocationAwsAz` model in `location_aws_az.rb`.
>     - Updates `Location` model in `location.rb` to include `location_aws_azs` association and methods `aws_azs`, `set_aws_azs`, and `get_azs_from_aws`.
>   - **Behavior**:
>     - `aws_azs` method in `Location` fetches AZs from AWS if not present in the database.
>     - Updates `az_to_provision_subnet` in `nic_nexus.rb` to use `aws_azs` instead of hardcoded values.
>   - **Tests**:
>     - Adds tests in `location_spec.rb` for `aws_azs` method behavior.
>     - Updates `nic_nexus_spec.rb` to test `az_to_provision_subnet` with new AZ fetching logic.
>   - **Misc**:
>     - Adds `TYPE_LOCATION_AWS_AZ` to `ubid.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for d3e9888fd3e90b2d9fc44eeec658aafef813cfb9. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/aws/nic_nexus.rb
+++ b/prog/vnet/aws/nic_nexus.rb
@@ -182,7 +182,7 @@ class Prog::Vnet::Aws::NicNexus < Prog::Base
   end
 
   def az_to_provision_subnet
-    frame["availability_zone"] || (["a", "b", "c"] - (frame["exclude_availability_zones"] || [])).sample || "a"
+    frame["availability_zone"] || (private_subnet.location.aws_azs.map(&:az) - (frame["exclude_availability_zones"] || [])).sample || "a"
   end
 
   def get_network_interface
--- a/spec/prog/vnet/aws/nic_nexus_spec.rb
+++ b/spec/prog/vnet/aws/nic_nexus_spec.rb
@@ -304,6 +304,12 @@ RSpec.describe Prog::Vnet::Aws::NicNexus do
   end
 
   describe "#az_to_provision_subnet" do
+    before do
+      nic.private_subnet.location.add_location_aws_az(az: "a", zone_id: "123")
+      nic.private_subnet.location.add_location_aws_az(az: "b", zone_id: "456")
+      nic.private_subnet.location.add_location_aws_az(az: "c", zone_id: "789")
+    end
+
     it "returns the az if set" do
       expect(nx).to receive(:frame).and_return({"availability_zone" => "a"})
       expect(nx.az_to_provision_subnet).to eq("a")
@@ -317,6 +323,13 @@ RSpec.describe Prog::Vnet::Aws::NicNexus do
     it "returns a if nothing is available" do
       expect(nx).to receive(:frame).and_return({"exclude_availability_zones" => ["a", "b", "c"]}).at_least(:once)
       expect(nx.az_to_provision_subnet).to eq("a")
+    end
+
+    it "fetches azs if not present" do
+      nic.private_subnet.location.reload.location_aws_azs.map(&:destroy)
+      client.stub_responses(:describe_availability_zones, availability_zones: [{zone_name: "us-west-2a", zone_id: "123"}, {zone_name: "us-west-2b", zone_id: "456"}, {zone_name: "us-west-2c", zone_id: "789"}])
+      expect(["a", "b", "c"]).to include(nx.az_to_provision_subnet) # rubocop:disable RSpec/ExpectActual
+      expect(LocationAwsAz.count).to eq(3)
     end
   end
 end


model/location.rb
```diff
@@ -52,6 +52,26 @@ def aws?
...
+
+    return super if super.any?
+
+    set_azs
+  end
+
+  private
+
+  def set_azs
+    fetched_azs = location_credential.client.describe_availability_zones.availability_zones.map { it.zone_name.split(name).last }
```
furkansahin: not sure if I should use split or basically fetching the very last character here... Any suggestions?

serprex: Are you looking for [`delete_prefix`](https://ruby-doc.org/core-2.5.7/String.html#method-i-delete_prefix)?

furkansahin: oh yeah, thanks!

migrate/20251206_add_location_azs.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:location) do
+      add_column :azs, :jsonb, null: false, default: "[]"
```
serprex:
```suggestion
      add_column :azs, "text[]", null: false, default: []
```
are there future concerns which would make text[] inappropriate?

model/location.rb
```diff
@@ -52,6 +52,26 @@ def aws?
...
+
+    return super if super.any?
+
+    set_azs
+  end
+
+  private
+
+  def set_azs
+    fetched_azs = location_credential.client.describe_availability_zones.availability_zones.map { it.zone_name.delete_prefix(name) }
```
byucesoy: We should probably keep the physical AZ information as well.

migrate/20251206_add_location_azs.rb
jeremyevans: Is there a reason you need this in an json/array column, instead of a separate table? Considering that this is only used for AWS, I don't like adding it to `location`, since that is used for non-AWS as well.  My recommendation would be to add a separate `aws_location_az` table with `location_id` and `az` columns.

If there is a hard technical requirement for this to be inside `location`, I agree with @serpex that a text array is probably better than jsonb.  However, the `default` needs to be different in that case:

```suggestion
      add_column :azs, :jsonb, null: false, default: "{}"
```

model/location.rb
```diff
@@ -52,6 +52,26 @@ def aws?
...
   end
+
+  # only valid for aws locations
+  def azs
+    unless aws?
+      Clog.emit("azs is only valid for aws locations") { {location_id: id} }
+      return []
+    end
+
+    return super if super.any?
```
jeremyevans: We should avoid calling `super` twice:

```suggestion
    v = super
    return v if v.any?
```

model/location.rb
```diff
@@ -52,6 +52,26 @@ def aws?
   def pg_ami(pg_version, arch)
     PgAwsAmi.find(aws_location_name: name, pg_version:, arch:).aws_ami_id
   end
+
+  # only valid for aws locations
+  def azs
+    unless aws?
+      Clog.emit("azs is only valid for aws locations") { {location_id: id} }
```
jeremyevans: Should this method ever be called for non-AWS?  If not, I think it's better to raise instead of using Clog.emit.

prog/vnet/aws/nic_nexus.rb
```diff
@@ -182,7 +182,7 @@ def vpc_id
   end
 
   def az_to_provision_subnet
-    frame["availability_zone"] || (["a", "b", "c"] - (frame["exclude_availability_zones"] || [])).sample || "a"
+    frame["availability_zone"] || (private_subnet.location.azs - (frame["exclude_availability_zones"] || [])).sample || "a"
```
jeremyevans: What does the `"a"` stand for here? Maybe obvious if you are used to AWS AZs, but I'm not.

furkansahin: picking an availability zone is optional. Excluding some availability zones are also optional. In case operator makes a mistake and we end-up with an empty possibility, we simply pick the az "a". Would you rather prefer if we raised here?

migrate/20251206_add_location_azs.rb
furkansahin: I switched to a separate table

model/location.rb
furkansahin: I switched to raise but doing so, I realized an issue. So, since the relationship I defined and this function had the same signature, admin page is trying to resolve this dependency and it called my function for locations even that are not AWS. So, essentially even when there was not a functional code path that would hit that, admin page was auto hitting it. I worked around that by simly modifying the relationship name and the method name but wanted to let you know

prog/vnet/aws/nic_nexus.rb
jeremyevans: I would only raise if you consider it to be an error. If an `a` availability zone is valid, it seems fine.

migrate/20251206_add_location_azs.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:location_aws_az) do
+      column :id, :uuid, primary_key: true
+      foreign_key :location_id, :location, type: :uuid, null: false, on_delete: :cascade
+      column :az, :text, null: false
```
jeremyevans, line 8: I think we should have a unique constraint/index on these two columns (or `location_id` and `zone_id` if that is more appropriate for AWS).

model/location_aws_az.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative "../model"
+
+class LocationAwsAz < Sequel::Model
+  many_to_one :location, key: :location_id
```
jeremyevans: `key` option is not necessary, as `location_id` would be the default:

```suggestion
  many_to_one :location
```

PR #4315, opened 2025-12-05
Add pkg-config to Postgres linux dependency instructions

fdr:
I was setting up a machine from scratch, and although pkg-config is often found on a development system that has seen enough usage, it is not seen in the default install of Ubuntu.

----

> [!IMPORTANT]
> Add `pkg-config` to Postgres dependencies for Debian/Ubuntu in `DEVELOPERS.md`.
> 
>   - **Dependencies**:
>     - Add `pkg-config` to Postgres dependencies for Debian/Ubuntu in `DEVELOPERS.md` to ensure proper setup on new systems.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for d2df5d2358c7d95992be6391261301e62a3174b2. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -188,7 +188,7 @@ Debian/Ubuntu based:
 apt-get install autoconf patch build-essential rustc libssl-dev libyaml-dev libreadline6-dev zlib1g-dev libgmp-dev libncurses5-dev libffi-dev libgdbm6 libgdbm-dev libdb-dev uuid-dev
 
 # Postgres
-apt-get install build-essential libssl-dev libreadline-dev zlib1g-dev libcurl4-openssl-dev uuid-dev icu-devtools libicu-dev
+apt-get install build-essential libssl-dev libreadline-dev zlib1g-dev libcurl4-openssl-dev uuid-dev icu-devtools libicu-dev pkg-config pkg-config bison flex
 ```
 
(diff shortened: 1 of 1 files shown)


DEVELOPERS.md
```diff
@@ -178,7 +178,7 @@ xcode-select --install
 brew install openssl@3 readline libyaml gmp autoconf
 
 # Postgres
-brew install gcc readline zlib curl ossp-uuid icu4c pkg-config
+brew install gcc readline zlib curl ossp-uuid icu4c pkg-config bison flex
```
enescakir: Previously, I didn't need to install bison and flex on macOS. Did you want to add them for Ubuntu?

We could remove these packages from our documentation and instead link to the official installation guides. That would reduce our maintenance effort.

bison and flex are already documented in the asdf-postgres docs.
https://github.com/smashedtoatoms/asdf-postgres/pull/100

fdr: it was a pure mistake, I should have put it into  Ubuntu only, this is why I never shipped it. my mistkae

PR #4320, opened 2025-12-08
SubnetNexus fix: Use nic.state in SubnetNexus.rb instead of the strand labels

furkansahin:
## Use nic.state in Metal/SubnetNexus.rb
We had previously implemented this, but, when we split SubnetNexus files
into Metal and AWS, we copied the old version of the files in the https://github.com/ubicloud/ubicloud/pull/4179.
This has caused some issues in production since we couldn't synchronize
the operations.

## Completely switch to refresh_keys instead of add_new_nic
Using the nic state, we don't have to follow the strand labels anymore.
Before, we were not doing this because the algorithm was depending on
safely synchronizing the label states and semaphores. Now, using the
entity state, we are sure that we can start working on a nic.

## Remove add_new_nic state

----

> [!IMPORTANT]
> Switch `SubnetNexus` and `NicNexus` to use `nic.state` for synchronization, remove `add_new_nic` logic, and update tests accordingly.
> 
>   - **Behavior**:
>     - Switch from strand labels to `nic.state` in `nic_nexus.rb` and `subnet_nexus.rb` for operation synchronization.
>     - Remove `add_new_nic` state and logic, consolidate into `refresh_keys`.
>   - **Functions**:
>     - Replace `when_add_new_nic_set?` with `when_refresh_keys_set?` in `subnet_nexus.rb`.
>     - Update `refresh_keys` to handle NICs in `active` and `creating` states in `subnet_nexus.rb`.
>   - **Tests**:
>     - Update tests in `nic_nexus_spec.rb` and `subnet_nexus_spec.rb` to reflect removal of `add_new_nic` and use of `nic.state`.
>     - Ensure tests cover new logic paths for `refresh_keys` and state transitions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c67a39c4c8a8df3adf2729bf4b08ff5910780527. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/metal/subnet_nexus.rb
+++ b/prog/vnet/metal/subnet_nexus.rb
@@ -23,11 +23,6 @@ class Prog::Vnet::Metal::SubnetNexus < Prog::Base
       hop_refresh_keys
     end
 
-    when_add_new_nic_set? do
-      private_subnet.update(state: "adding_new_nic")
-      hop_add_new_nic
-    end
-
     when_update_firewall_rules_set? do
       private_subnet.vms.each(&:incr_update_firewall_rules)
       decr_update_firewall_rules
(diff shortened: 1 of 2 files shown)


prog/vnet/metal/subnet_nexus.rb
```diff
@@ -52,24 +47,6 @@ def gen_reqid
     SecureRandom.random_number(1...100000)
   end
 
-  label def add_new_nic
```
bsatzger, line 54: a multi-step deployment would make sense where this commit is deployed when we can be sure no subnet is in this label.

furkansahin: Yeap, that's the plan

PR #4325, opened 2025-12-09
make storage size options numbers instead of strings

serprex:
prog/vm/nexus.rb was ending up trying to divide strings which were leaking through from storage options

convert everything to having storage size be numeric, which was then having code convert back with to_i anyways

----

> [!IMPORTANT]
> Convert storage size handling to numeric values across PostgreSQL-related CLI commands, backend logic, and tests.
> 
>   - **Behavior**:
>     - Convert `storage_size` to numeric in `create.rb` and `modify.rb` for PostgreSQL CLI commands.
>     - Update `storage_size` handling in `postgres.rb` and `location/postgres.rb` to use numeric values.
>   - **Options**:
>     - Change `POSTGRES_STORAGE_SIZE_OPTIONS` in `option.rb` to numeric values.
>     - Update `non_storage_optimized_vm_storage_size_options` and `AWS_STORAGE_SIZE_OPTIONS` to use numeric values.
>   - **Tests**:
>     - Modify test cases in `postgres_spec.rb` to use numeric `storage_size` values.
>   - **Misc**:
>     - Adjust `opt_classes` generation in `resource_creation_form.erb` to handle numeric values.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a9e377151c617dbc03c93b053d38c658ccf76efd. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vm/nexus.rb
+++ b/prog/vm/nexus.rb
@@ -28,10 +28,7 @@ class Prog::Vm::Nexus < Prog::Base
     vm_size = Validation.validate_vm_size(size, arch)
     Validation.validate_billing_rate("VmVCpu", vm_size.family, location.name)
 
-    storage_volumes ||= [{
-      size_gib: vm_size.storage_size_options.first,
-      encrypted: true
-    }]
+    storage_volumes ||= [{}]
 
     # allow missing fields to make testing during development more convenient.
     storage_volumes.each_with_index do |volume, disk_index|


enescakir:
Probably it needs a better commit title

prog/vm/nexus.rb
```diff
@@ -28,10 +28,7 @@ def self.assemble(public_key, project_id, name: nil, size: DEFAULT_SIZE,
     vm_size = Validation.validate_vm_size(size, arch)
     Validation.validate_billing_rate("VmVCpu", vm_size.family, location.name)
 
-    storage_volumes ||= [{
-      size_gib: vm_size.storage_size_options.first,
-      encrypted: true
-    }]
+    storage_volumes ||= [{}]
```
jeremyevans, line 31: This change doesn't look related to the other changes. If it is needed, can you use a separate commit and explain why it is needed in the commit message?

jeremyevans (commented):
Overall, I think this is a good change.

prog/vm/nexus.rb
serprex: Will separate. It isn't needed, but these fields are assigned the same through default value logic like the rest of the fields

PR #4327, opened 2025-12-09
Load balancer ipv6 validationerror

fdr:
maybe? the first commit has the semantic change, the rest are semantically equivalent (right?!) refactors.

----

> [!IMPORTANT]
> Refactor DNS record handling in `load_balancer_nexus.rb` by introducing `vm_dns_records` method and update tests accordingly.
> 
>   - **Refactor DNS Record Handling**:
>     - Introduce `vm_dns_records(vm)` in `load_balancer_nexus.rb` to consolidate DNS record logic.
>     - Replace inline DNS record checks and insertions with `vm_dns_records` in `need_to_rewrite_dns_records?` and `rewrite_dns_records`.
>   - **Tests**:
>     - Update `load_balancer_nexus_spec.rb` to use `expect_dns_records` helper for DNS record expectations.
>     - Add tests for scenarios where DNS records should not be created based on IP availability and stack configuration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9152f195df5c520d16138dcec4e7df7834351f07. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -93,10 +93,10 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
   end
 
   def vm_dns_records(vm)
-    [].tap do
-      it << ["A", vm.ip4_string] if load_balancer.ipv4_enabled? && vm.ip4_string
-      it << ["AAAA", vm.ip6_string] if load_balancer.ipv6_enabled? && vm.ip6_string
-    end
+    ips = []
+    ips << ["A", vm.ip4_string] if load_balancer.ipv4_enabled? && vm.ip4_string
+    ips << ["AAAA", vm.ip6_string] if load_balancer.ipv6_enabled? && vm.ip6_string
+    ips
   end
 
   label def create_new_cert


prog/vnet/load_balancer_nexus.rb
```diff
@@ -84,18 +84,21 @@ def need_to_rewrite_dns_records?
...
     end
 
     false
   end
 
+  def vm_dns_records(vm)
+    [].tap do
+      it << ["A", vm.ip4_string] if load_balancer.ipv4_enabled? && vm.ip4_string
+      it << ["AAAA", vm.ip6_string] if load_balancer.ipv6_enabled? && vm.ip6_string
+    end
```
jeremyevans: Up to you, but it's more efficient and IMO simpler to use a local variable.

```suggestion
    ips = []
    ips << ["A", vm.ip4_string] if load_balancer.ipv4_enabled? && vm.ip4_string
    ips << ["AAAA", vm.ip6_string] if load_balancer.ipv6_enabled? && vm.ip6_string
    ips
```

prog/vnet/load_balancer_nexus.rb
```diff
@@ -170,23 +173,12 @@ def need_to_rewrite_dns_records?
...
-          record_name: load_balancer.hostname,
-          type: "A",
-          ttl: 10,
-          data: vm.ip4_string
-        )
-      end
-
-      # Insert IPv6 record if stack is ipv6 or dual
-      if load_balancer.ipv6_enabled?
+      vm_dns_records(vm).each do |type, data|
```
jeremyevans, line 176: Might be worth surrounding the `each` call inside `if load_balancer.dns_zone`, and drop the `&.`, so you only check it once and not twice. Up to you.

furkansahin:
So, the reason is, vms might be provisioned without an IPv4 address by customer's choice but not without an ipv6 address. Especially if they are part of a load balancer, there is no way a VM can be added and it doesn't have IPv6. Because the whole load balancer stack depends on IPv6. 
On top, this PR makes it so if the VM is not allocated yet and it is added to the loadbalancer, we simply skip adding its address to the dns. And, when do we add it? Maybe we fail the request quick saying that VM is not ready to be added to a load balancer, yet. Or, the initial implementation would retry until it is ready.

fdr:
> So, the reason is, vms might be provisioned without an IPv4 address by customer's choice but not without an ipv6 address. Especially if they are part of a load balancer, there is no way a VM can be added and it doesn't have IPv6. Because the whole load balancer stack depends on IPv6.

Huh, is the address not NULL right in the beginning of the allocation?  I feel like I saw something like that in my ubidon iteration, but I'm not 100% sure

> On top, this PR makes it so if the VM is not allocated yet and it is added to the loadbalancer, we simply skip adding its address to the dns. And, when do we add it? Maybe we fail the request quick saying that VM is not ready to be added to a load balancer, yet. Or, the initial implementation would retry until it is ready.

I am not sure about that either, hence my uncertainty: since Jeremy said it happened "occasionally" I figured it was a data race that self-resolved. Maybe I'm barking up the wrong tree.

fdr:
> I am not sure about that either, hence my uncertainty: since Jeremy said it happened "occasionally" I figured it was a data race that self-resolved. Maybe I'm barking up the wrong tree.

By self-resolved, I presumed a semaphore might get incremented and there'd be re-entrance, rather than, perhaps this code was relying on throwing the error to re-enter and poll. But I made a bad bug here if that's not the case, and perhaps did not understand the cause well either (is IPv6 always set? it can't be, right, or else the ValidationError would never happen)

furkansahin (approved):
re-reading the nexus, this actually works. Thanks! The continuous check `need_to_rewrite_dns_records` in the `wait` label will make sure the ip is added to the dns after all.

fdr:
I think this is obsolete via 33648d2bc and @jeremyevans

PR #4331, opened 2025-12-09
PostgeSQL init script

byucesoy:
This allows users to specify an initialization script that would be executed after mounting data disk. Currently this is behind a feature flag as this can break initial provisioning, so we would only enable this for power users.

### DIFF
--- a/model/postgres/aws/postgres_timeline.rb
+++ b/model/postgres/aws/postgres_timeline.rb
@@ -45,7 +45,11 @@ class PostgresTimeline < Sequel::Model
 
     def aws_create_bucket
       location_constraint = (location.name == "us-east-1") ? nil : {location_constraint: location.name}
-      blob_storage_client.create_bucket(bucket: ubid, create_bucket_configuration: location_constraint)
+      begin
+        blob_storage_client.create_bucket(bucket: ubid, create_bucket_configuration: location_constraint)
+      rescue Aws::S3::Errors::BucketAlreadyOwnedByYou
+        # Ignore if bucket already exists
+      end
     end
 
     def aws_set_lifecycle_policy


prog/postgres/postgres_server_nexus.rb
```diff
@@ -144,6 +140,21 @@ def before_run
...
   end
 
+  label def run_init_script
+    hop_configure_walg_credentials unless postgres_server.resource.init_script
+
+    case vm.sshable.d_check("run_init_script")
+    when "Succeeded"
+      hop_configure_walg_credentials
+    when "Failed", "NotStarted"
+      vm.sshable.cmd("sudo tee init_script.sh > /dev/null", stdin: postgres_server.resource.init_script)
```
shikharbhardwaj: nit: should we keep it within a folder like `postgres/bin` to keep the home directory clean?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -144,6 +140,21 @@ def before_run
...
+      vm.sshable.cmd("sudo chmod +x init_script.sh")
+      vm.sshable.d_run("run_init_script", "sudo", "./init_script.sh")
```
shikharbhardwaj: I'm guessing `sudo` is to allow users to install packages etc? If this is SQL only we should restrict it imo

helpers/postgres.rb
```diff
@@ -14,6 +14,7 @@ def postgres_post(name)
     pgbouncer_user_config = typecast_params.Hash("pgbouncer_config", {})
     tags = typecast_params.array(:Hash, "tags", [])
     with_firewall_rules = !typecast_params.bool("restrict_by_default")
+    init_script = typecast_params.nonempty_str("init_script")
```
shikharbhardwaj, line 18: We should change the openapi definitions as well. Also, I am assuming this works automatically with the CLI and SDK?

migrate/20251209_add_pg_init_script.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:postgres_resource) do
+      add_column :init_script, :text, null: true
```
enescakir: What’s the limit for this column? Since we say `This is useful for setting up initial schemas, tables, or data`, I fear that some customer might try to put a large dataset into it.

migrate/20251209_add_pg_init_script.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:postgres_resource) do
```
enescakir: We store vm init scripts in a separate table.
@jeremyevans do you remember the reasoning behind putting vm init_script in its own table? It might apply here as well, and I’m leaning toward keeping a similar pattern across resources.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -140,6 +140,21 @@ def before_run
...
   end
 
+  label def run_init_script
+    hop_configure_walg_credentials unless postgres_server.resource.init_script
+
+    case vm.sshable.d_check("run_init_script")
+    when "Succeeded"
+      hop_configure_walg_credentials
+    when "Failed", "NotStarted"
+      vm.sshable.cmd("sudo tee init_script.sh > /dev/null", stdin: postgres_server.resource.init_script)
```
enescakir: Quick question. Does this script need a shebang? And if it’s sql, should we use a `.sql` extension?

byucesoy: This file is intended to be .sh file, so that people can install binaries. Users would add shebang themselves as part of the input.

prog/postgres/postgres_server_nexus.rb
byucesoy: Yes, this is to allow users to install packages. It is not SQL only. It is a shell script file, which can technically contain SQL as well.

prog/postgres/postgres_server_nexus.rb
byucesoy: Makes sense. Moved it to a `postgres/bin`

prog/postgres/postgres_server_nexus.rb
byucesoy: Though, thinking about it... `sudo` is unnecessary. Users can add sudo to their own script. I'll remove it.

migrate/20251209_add_pg_init_script.rb
byucesoy: Makes sense. Added 2000 character limit, like we do in vm init scripts.

migrate/20251209_add_pg_init_script.rb
jeremyevans: I would recommend a separate table. We only need this in a single place, we don't want to pull an extra 2KB column per PG resource on every PG resource retrieval.

migrate/20251209_add_pg_init_script.rb
```diff
@@ -0,0 +1,9 @@
...
+      add_column :init_script, String, size: 2000
```
jeremyevans: We should probably be encrypted the column for the same reason we encrypt vm init scripts. When encrypting, we shouldn't have a database column limit, we need to move the max length limit to a model validation, so it's done on the unencrypted value.

views/postgres/create.erb
```diff
@@ -49,6 +49,10 @@
     {name: "ha_type", type: "radio_small_cards", label: "High Availability", required: "required", content_generator: ContentGenerator::Postgres.method(:ha_type)},
   ]
 
+  if @project.get_ff_postgres_init_script
+    form_elements << {name: "init_script", type: "textarea", label: "Initialization Script", required: "", placeholder: "Enter script to run on database creation"}
```
jeremyevans: We should set a maxlength attribute.

```suggestion
    form_elements << {name: "init_script", type: "textarea", label: "Initialization Script", required: "", placeholder: "Enter script to run on database creation", attributes: {maxlength: 2000}}
```

spec/routes/web/project/postgres_spec.rb
```diff
@@ -404,6 +425,23 @@
         expect(page).to have_no_content "Danger Zone"
       end
 
+      it "allows updating initialization script when feature flag is enabled" do
+        project.set_ff_postgres_init_script(true)
+        visit "#{project.path}#{pg.path}/settings"
+        expect(page).to have_content "Initialization Script"
+        fill_in "init_script", with: "sudo whoami"
+
+        # We send PATCH request manually instead of just clicking to button because PATCH action triggered by JavaScript.
```
jeremyevans: We have a lot of these comments, and I don't think they add value. I also have plans to change things so we can use standard Capybara commands instead of `page.driver.submit` (already implemented in a few places, just not everywhere yet).

```suggestion
```

views/postgres/settings.erb
```diff
@@ -36,6 +36,47 @@
...
+      <div class="md:flex md:items-center md:justify-between pb-2 lg:pb-4">
+        <div class="min-w-0 flex-1">
+          <h3 class="text-2xl font-bold leading-7 text-gray-900 sm:truncate sm:text-2xl sm:tracking-tight">
+            Initialization Script
+          </h3>
+        </div>
+      </div>
+      <div class="overflow-hidden rounded-lg shadow ring-1 ring-black ring-opacity-5 bg-white divide-y divide-gray-200">
+        <% form_elements = [
+              {name: "init_script", type: "textarea", required: "", placeholder: "Enter script to run on database creation"}
```
jeremyevans, line 47:
```suggestion
              {name: "init_script", type: "textarea", required: "", placeholder: "Enter script to run on database creation", attributes: {maxlength: 2000}}
```

views/postgres/settings.erb
```diff
@@ -36,6 +36,47 @@
       </div>
     </div>
   </div>
+  <% if @project.get_ff_postgres_init_script %>
```
jeremyevans, line 36: This is used in a commit before the `postgres_init_script` feature flag is added.

I don't think we should use a feature flag for this. We don't use one for VM init scripts. Is there a reason we should do so for Postgres init scripts?

views/postgres/settings.erb
```diff
@@ -36,6 +36,47 @@
...
+  locals: {
+    action: path(@pg),
+    method: "PATCH",
+    form_elements:,
+    pre_selected_values:,
+    option_tree: @option_tree,
+    option_parents: @option_parents,
+    mode: "update"
+  }
+) %>
```
jeremyevans, line 72: This seems way too complicated.  Is there an advantage to this resource creation form approach as opposed to a simple form with a single textarea and button?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -147,7 +147,7 @@ def before_run
     when "Succeeded"
       hop_configure_walg_credentials
     when "Failed", "NotStarted"
-      vm.sshable.cmd("sudo tee postgres/bin/init_script.sh > /dev/null", stdin: postgres_server.resource.init_script)
+      vm.sshable.cmd("sudo tee postgres/bin/init_script.sh > /dev/null", stdin: postgres_server.resource.init_script.gsub("\r\n", "\n"))
```
jeremyevans: Should be squashed into earlier commit.

jeremyevans (commented):
I'm in favor of adding this feature, though I have a few implementation suggestions.

furkansahin:
I don't have a comment on this. Let's make sure the conflicts are resolved and the tests pass

helpers/postgres.rb
byucesoy: We plan to leave this feature to be UI-only for now.

views/postgres/settings.erb
byucesoy: I agree that this is bit complicated for just a single textarea, but it provides standardization over how we render forms. It also handles showing errors.

helpers/postgres.rb
jeremyevans: > We plan to leave this feature to be UI-only for now.

That seems odd, as I would assume an init script feature provides more benefit to a API/CLI workflow than a web UI one. Can you explain the reasoning behind the decision?

PR #4332, opened 2025-12-09
Add disk-full-check on postgres vm to manage default_transaction_read_only

serprex:
Above 7GiB turn off default read only & recreate 1GB buffer file
Below 5G default transaction mode becomes readonly
Below 3G postgres is restarted, killing long running connections

For disks smaller than 64GB this is compressed to 2GB/1GB/512MB

Long running connections can be a problem since they wouldn't be read only,
hence need to restart when disk usage becomes critical

### DIFF
--- a/rhizome/postgres/bin/disk-full-check
+++ b/rhizome/postgres/bin/disk-full-check
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -eu
+
+DAT=${DAT:-/dat}
+
+read name btotal bused bavailable usedp mount_path <<< "$(df -P -B1 "$DAT" | tail -n +2)"
+export PGDATA="$DAT/$1/data"
+
+if test "$btotal" -le 68719476736; then # Smaller thresholds for disks <= 64GB (hobby)
+  recover_threshold=2147483648    # 2GB
+  readonly_threshold=1073741824   # 1GB
+  restart_threshold=536870912     # 512MB
+  human_buffer_size=500M
+else
+  recover_threshold=7516192768    # 7GB
+  readonly_threshold=5368709120   # 5GB
+  restart_threshold=3221225472    # 3GB
+  human_buffer_size=1G
+fi
+
+if test "$bavailable" -gt "$recover_threshold"; then
+  if grep -q "default_transaction_read_only = 'on'" "$PGDATA/postgresql.auto.conf"; then
+    sed -i "/default_transaction_read_only = 'on'/d" "$PGDATA/postgresql.auto.conf"
+    test -f "$DAT/disk-full-read-only-pending-restart-$1" && rm -f "$DAT/disk-full-read-only-pending-restart-$1"
+    pg_ctl reload
+  elif ! test -f "$DAT/disk-full-human-buffer"; then
+    fallocate -l "$human_buffer_size" "$DAT/disk-full-human-buffer"
+  fi
+elif test "$bavailable" -lt "$readonly_threshold"; then
+  if ! grep -q "default_transaction_read_only = 'on'" "$PGDATA/postgresql.auto.conf"; then
+    if test -n "$(tail -c 1 "$PGDATA/postgresql.auto.conf")"; then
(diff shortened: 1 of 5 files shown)


serprex:
on Azure logic started with controlplane & it had reliability issues (you have latency getting info to controlplane, then latency reacting). but maybe controlplane logic just needed the extra step to kill connections as followup if disk usage keeps growing

furkansahin:
We may at least do something hybrid where the state is replicated to the controlplane and all of the action still happens/managed in the dataplane. Once we switch to a not so normal state, controlplane can continuously check if the script is running. Probably the failure to start loop after switching to read only is the biggest possible problem.

We can even read a status file using the monitoring logic and switch to unavailable in case it is not a healthy state.

serprex:
bringing this back because we're still having problems despite #4730 

fact is when someone wants to fill their disk standby can't catch up

added rspec integration for bash script

serprex:
idea: pipe metrics? worry about writing to disk tho, need to make sure metrics are light

rhizome/postgres/bin/disk-full-check
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -eu
+
+DAT=${DAT:-/dat}
+
+read name btotal bused bavailable usedp mount_path <<< "$(df -P -B1 "$DAT" | tail -n +2)"
```
serprex, line 6: verify `usedp` doesn't include root reserved

serprex:
burak suggested taking out % logic

rhizome/postgres/bin/disk-full-check
```diff
@@ -0,0 +1,29 @@
...
+export PGDATA="$DAT/$1/data"
+
+if test "$bavailable" -gt 7516192768; then
+  if grep -q "default_transaction_read_only = 'on'" "$DAT/$1/data/postgresql.auto.conf"; then
+    sed -i "/default_transaction_read_only = 'on'/d" "$DAT/$1/data/postgresql.auto.conf"
+    test -f "$DAT/disk-full-read-only-pending-restart-$1" && rm -f "$DAT/disk-full-read-only-pending-restart-$1"
+    pg_ctl reload
+  elif ! test -f "$DAT/disk-full-human-buffer"; then
```
byucesoy, line 26: This should be an independent if, right?

rhizome/postgres/bin/disk-full-check
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -eu
+
+DAT=${DAT:-/dat}
+
+read name btotal bused bavailable usedp mount_path <<< "$(df -P -B1 "$DAT" | tail -n +2)"
+export PGDATA="$DAT/$1/data"
```
byucesoy, line 7: There are various places where we can use PGDATA instead of DAT below.

rhizome/postgres/bin/disk-full-check
```diff
@@ -0,0 +1,29 @@
...
+
+if test "$bavailable" -gt 7516192768; then
```
byucesoy: Hobby instances can have only 16GB of space. For those we can use lower limits. I'm thinking like if total disk space is less than or equal to 64GB, we use 2GB as limit here, 1GB below and 500MB human buffer.

rhizome/postgres/bin/disk-full-check
serprex: potentially, but I prefer system to move slowly here, one action at a time

rhizome/postgres/lib/postgres_upgrade.rb
```diff
@@ -39,6 +39,7 @@ def promote(version)
 
   def disable_previous_version
     r "sudo systemctl disable --now postgresql@#{@prev_version}-main"
+    r "sudo systemctl disable --now disk-full-check@#{@version}.timer"
```
byucesoy, line 41: We need to create these services manually for existing servers or else upgrade would get broken because this command would fail. (@serprex just reminder for the deployment)

PR #4335, opened 2025-12-09
Use dashes instead of underscores for fields in the CLI output

jeremyevans:
Almost all from Claude Code, using the following prompt:

```
In Ubicloud's CLI input and output, fields should use `-` and
not `_`. Internally, Ubicloud uses `_` to separate words. When
accepting CLI option input, Ubicloud converts `-` to `_`. However,
when outputing fields, Ubicloud does not convert from `_` back
to `-`. Fix this issue by modifying `lib/ubi_cli.rb` files under
`cli-commands` to convert output field names to use `-` instead of
`_`.
```

Only change made was removing an unused branch from the created dash_keys method.

In some cases, this changes from spaces to dashes and uppercase uppercase to lower case
(e.g. `CA certificates -> ca-certificates`).

Note that this may break programs that attempt to parse the current CLI output format.

Best reviewed with
`git diff --color-words --word-diff-regex='\\w+|[^[:space:]]'`

----

> [!IMPORTANT]
> Modify CLI output to use dashes instead of underscores for field names across various commands, updating code and tests accordingly.
> 
>   - **Behavior**:
>     - Modify CLI output to use dashes instead of underscores for field names in `fw`, `kc`, `lb`, `pg`, `ps`, and `vm` commands.
>     - Introduce `each_with_dashed` method in `lib/ubi_cli.rb` to convert underscores to dashes in output field names.
>   - **Code Changes**:
>     - Update `fw/post/show.rb`, `kc/post/show.rb`, `lb/post/show.rb`, `pg/post/show.rb`, `ps/post/show.rb`, and `vm/post/show.rb` to use `each_with_dashed` for field name conversion.
>     - Remove unused branch from `dash_keys` method in `lib/ubi_cli.rb`.
>   - **Tests**:
>     - Update golden files in `spec/routes/api/cli/golden-files` to reflect new field name format with dashes.
>     - Modify `show_spec.rb` for `pg` and `vm` to match updated output format.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a0f9c3b6f4e21d0599f19729a2134ef95ebb5a9b. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli-commands/fw/post/show.rb
+++ b/cli-commands/fw/post/show.rb
@@ -29,7 +29,7 @@ UbiCli.on("fw").run_on("show") do
 
     body = []
 
-    keys.each do |key|
+    each_with_dashed(keys) do |key, display_key|
       case key
       when :firewall_rules
         body << "rules:\n"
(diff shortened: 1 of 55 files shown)


fdr (approved):
I guess that's alright (and we can probably tolerate the change at our phase), but it seems like we're missing an abstraction rather than inlining the tr all the time. seems easy to regress and have more mistakes creep in

jeremyevans:
> I guess that's alright (and we can probably tolerate the change at our phase), but it seems like we're missing an abstraction rather than inlining the tr all the time. seems easy to regress and have more mistakes creep in

Agreed. I wrote an `each_with_dashed` method and had Claude make the appropriate changes to use it. No spec changes were needed compared to the initial submission of this PR, but it's a lot simpler now.

PR #4339, opened 2025-12-10
Use Sequel forbid_lazy_load plugin during non-frozen testing

jeremyevans:
Fix the one N+1 query issue found by it.

I'm honestly quite surprised this didn't find more issues, but I guess the earlier introduction of the pg_auto_parameterize_duplicate_query_detection extension resulted in most of the issues being found and fixed.

Note that there can still be N+1 issues that neither of these catches.

----

> [!IMPORTANT]
> Add Sequel `forbid_lazy_load` plugin for non-frozen testing to prevent lazy loading and fix an N+1 query issue by eagerly loading `location`.
> 
>   - **Behavior**:
>     - Add `forbid_lazy_load` plugin in `model.rb` for non-frozen testing environments to prevent lazy loading.
>     - Fix N+1 query issue in `routes/project/cli.rb` by eagerly loading `location` in `UBID.resolve_map`.
>   - **Configuration**:
>     - Plugin applied when `Config.test?` is true and `ENV["CLOVER_FREEZE"]` is not "1".
>   - **Misc**:
>     - Minor changes in `ubid.rb` to support the eager loading fix.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for e3cd62e33793ab79395d73f14531fe74b0ba0d62. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model.rb
+++ b/model.rb
@@ -30,6 +30,7 @@ Sequel::Model.plugin :pg_auto_constraint_validations, cache_file: "cache/pg_auto
 Sequel::Model.plugin :pg_auto_validate_enums, message: proc { |valid_values| "is not one of the supported values (#{valid_values.sort.join(", ")})" }
 Sequel::Model.plugin :pg_eager_any_typed_array
 Sequel::Model.plugin :association_lazy_eager_option
+Sequel::Model.plugin :forbid_lazy_load if Config.test? && ENV["CLOVER_FREEZE"] != "1"
 
 if (level = Config.database_logger_level) || Config.test?
   require "logger"


model.rb
```diff
@@ -30,6 +30,7 @@
 Sequel::Model.plugin :pg_auto_validate_enums, message: proc { |valid_values| "is not one of the supported values (#{valid_values.sort.join(", ")})" }
 Sequel::Model.plugin :pg_eager_any_typed_array
 Sequel::Model.plugin :association_lazy_eager_option
+Sequel::Model.plugin :forbid_lazy_load if Config.test? && ENV["CLOVER_FREEZE"] != 1
```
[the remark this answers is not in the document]
jeremyevans: Good find.

ubid.rb
```diff
@@ -200,7 +200,8 @@ def self.resolve_map(uuids, assume_et_is_api_key: true)
     end.each do |model, model_uuids|
       next unless model
 
-      model.where(id: model_uuids).each do
+      ds = model.where(id: model_uuids)
+      (block_given? ? yield(ds).all : ds).each do
```
serprex, line 204: why is `.all` necessary? if needed by the one caller, why does it belong here & not in the yielded to block?

jeremyevans: This is currently only used for eager loading, and eager loading requires `all` instead of `each` (because to eager load, you need all values up front).

It's typical for methods that are yielded a dataset to return a dataset from the block instead of an array from the block.

PR #4344, opened 2025-12-10
Add tags to pg /read-replica endpoint

serprex:
Aligns with /create which already accepts tags

----

> [!IMPORTANT]
> Add support for creating Postgres read replicas with tags via API, updating OpenAPI spec, route handling, and tests.
> 
>   - **Behavior**:
>     - Add `tags` property to read replica creation in `openapi.yml`.
>     - Update `createPostgresDatabaseReadReplica` in `postgres.rb` to accept and process `tags`.
>     - Ensure `tags` are applied to replicas in `postgres.rb`.
>   - **Tests**:
>     - Add test for read replica creation with `tags` in `postgres_spec.rb`.
>     - Validate `tags` are correctly set on created replicas in `postgres_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9f33011649052722488c33ffadda271ca28b45ff. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1359,7 +1359,13 @@ paths:
               type: object
               properties:
                 name:
+                  description: Name for the read replica
                   type: string
+                tags:
+                  description: Tags for the read replica
+                  type: array
+                  items:
+                    $ref: '#/components/schemas/PostgresTag'
               additionalProperties: false
               required:
                 - name
--- a/spec/routes/api/project/location/postgres_spec.rb
+++ b/spec/routes/api/project/location/postgres_spec.rb
@@ -316,6 +316,22 @@ RSpec.describe Clover, "postgres" do
         expect(last_response.status).to eq(200)
       end
 
+      it "read-replica with tags" do
+        VmStorageVolume.create(vm_id: pg.representative_server.vm.id, size_gib: pg.target_storage_size_gib, boot: false, disk_index: 0)
+        expect(PostgresTimeline).to receive(:earliest_restore_time).and_return(true)
+
+        tags = [{"key" => "env", "value" => "test"}, {"key" => "team", "value" => "backend"}]
+        post "/project/#{project.ubid}/location/eu-central-h1/postgres/#{pg.name}/read-replica", {
+          name: "my-read-replica-with-tags",
+          tags:
+        }.to_json
+
+        expect(last_response.status).to eq(200)
+
+        created_replica = PostgresResource.first(name: "my-read-replica-with-tags")
+        expect(created_replica.tags).to eq(tags)
+      end
+
       it "fails read-replica creation if the parent is not ready" do
         expect(project).to receive(:postgres_resources_dataset).and_return(instance_double(Sequel::Dataset, first: pg))
         expect(described_class).to receive(:authorized_project).with(user, project.id).and_return(project)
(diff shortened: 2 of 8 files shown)


Amogh-Bharadwaj:

furkansahin (changes requested):
can you please add that functionality to sdk/cli as well?

openapi/openapi.yml
```diff
@@ -1360,6 +1360,11 @@ paths:
               properties:
                 name:
                   type: string
+                tags:
+                  type: array
+                  items:
+                    $ref: '#/components/schemas/PostgresTag'
+                  default: []
```
jeremyevans: As we don't use `default` for other cases where tags are used, I don't think we should do so here.

```suggestion
```

openapi/openapi.yml
```diff
@@ -1360,6 +1360,11 @@ paths:
               properties:
                 name:
                   type: string
+                tags:
+                  type: array
```
jeremyevans, line 1366: We should have a `description` here. I'm not sure why a `description` is missing from the `name` property, we have it on most other properties. I'm quite surprised the openapi.yml linter doesn't flag missing descriptions.

@fdr @geemus do you think we should have the openapi.yml linter check for missing property descriptions?

spec/routes/api/project/location/postgres_spec.rb
```diff
@@ -316,6 +316,22 @@
...
       end
 
+      it "read-replica with tags" do
+        VmStorageVolume.create(vm_id: pg.representative_server.vm.id, size_gib: pg.target_storage_size_gib, boot: false, disk_index: 0)
+        expect(PostgresTimeline).to receive(:earliest_restore_time).and_return(true)
+
+        tags = [{"key" => "env", "value" => "test"}, {"key" => "team", "value" => "backend"}]
+        post "/project/#{project.ubid}/location/eu-central-h1/postgres/#{pg.name}/read-replica", {
+          name: "my-read-replica-with-tags",
+          tags: tags
```
jeremyevans:
```suggestion
          tags:
```

jeremyevans (commented):
This looks mostly good, and I think it's a useful feature.

For each commit, please try to ensure the commit message contains a useful description. It would be good to review our commit history and try to use a similar style. For example, we don't use `feat:` prefixes in commit messages, and the first letter is capitalized. Additionally, we generally include significantly more detail about the commit, especially the purpose behind the commit. Commit messages are important to us.  Applying that to this commit, the first commit should explain that this is already supported when creating a new PostgreSQL database, but wasn't supported when creating read replicas, so this makes things more consistent. For the second commit, you may not need as much detail, but it should at least describe the change (e.g. "Add CLI support for pg create-read-replica to accept tags").

openapi/openapi.yml
serprex: added descriptions to `name` & `tags` based on /create descriptions

leaving conversation unresolved for lint converation

openapi/openapi.yml
```diff
@@ -1359,7 +1359,13 @@ paths:
               type: object
               properties:
                 name:
+                  description: New name for the read replica
```
jeremyevans: "New name" implies the read replica had an existing name, but as we are creating the read-replica, that isn't the case.

```suggestion
                  description: Name for the read replica
```

jeremyevans (approved):

PR #4348, opened 2025-12-10
Observe archival backlog and page if it exceeds threshold

shikharbhardwaj:
> [!IMPORTANT]
> Add archival backlog monitoring and paging to PostgreSQL server with adaptive threshold.
> 
>   - **Behavior**:
>     - Adds `observe_archival_backlog` method in `postgres_server.rb` to monitor archival backlog and page if it exceeds a threshold.
>     - Updates `check_pulse` in `postgres_server.rb` to call `observe_archival_backlog` every 12th pulse.
>     - Threshold is adaptive to storage size, capped at 1000.
>   - **Tests**:
>     - Adds tests for `observe_archival_backlog` and `archival_backlog_threshold` in `postgres_server_spec.rb`.
>     - Updates existing tests in `monitorable_resource_spec.rb` to expect `observe_archival_backlog` call.
>   - **Misc**:
>     - Adjusts logging expectations in `monitorable_resource_spec.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for f38f97c7b6059e9562f284d2e2e158da82c935f0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -250,6 +250,10 @@ class PostgresServer < Sequel::Model
       end
     end
 
+    if pulse[:reading_rpt] % 12 == 1
+      observe_archival_backlog(session:)
+    end
+
     pulse
   end
 
@@ -333,6 +337,35 @@ class PostgresServer < Sequel::Model
     if Config.aws_postgres_iam_access && timeline.aws? && vm.aws_instance.iam_role
       timeline.location.location_credential.iam_client.attach_role_policy(role_name: vm.aws_instance.iam_role, policy_arn: timeline.aws_s3_policy_arn)
     end
+  end
+
+  def observe_archival_backlog(session:)
+    result = session[:ssh_session].exec!(
+      "sudo find /dat/:version/data/pg_wal/archive_status -name '*.ready' | wc -l",
+      version:
+    )
+    archival_backlog = result.strip.to_i
+
+    if archival_backlog > archival_backlog_threshold
+      Prog::PageNexus.assemble("#{ubid} archival backlog high",
+        ["PGArchivalBacklogHigh", id], ubid,
+        severity: "warning", extra_data: {archival_backlog: archival_backlog})
+    else
+      Page.from_tag_parts("PGArchivalBacklogHigh", id)&.incr_resolve
+    end
+  rescue => ex
+    Clog.emit("Failed to observe archival backlog") do
+      {postgres_server_id: id, exception: Util.exception_to_hash(ex)}
+    end
+  end
+
+  def archival_backlog_threshold
+    # To make the threshold adaptive to storage size, we set it as percent of
+    # the storage size as WAL file count based on the allocated storage size,
+    # capped to 1000 to avoid high thresholds on large storage sizes.
+    archival_backlog_threshold_percent = 5
+    archival_backlog_threshold_count = 1000
+    [(storage_size_gib * 1024 / (16 * 100)) * archival_backlog_threshold_percent, archival_backlog_threshold_count].min
   end
 
   FAILOVER_LABELS = ["prepare_for_unplanned_take_over", "prepare_for_planned_take_over", "wait_fencing_of_old_primary", "taking_over"].freeze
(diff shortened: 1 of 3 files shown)


spec/lib/monitorable_resource_spec.rb
```diff
@@ -53,7 +53,7 @@
       expect(session[:ssh_session]).to receive(:close)
       expect(Thread).to receive(:new).and_call_original
       expect(session[:ssh_session]).to receive(:loop).and_raise(StandardError)
-      expect(Clog).to receive(:emit).twice.and_call_original
+      expect(Clog).to receive(:emit).exactly(3).and_call_original
```
jeremyevans: If you really want to test it's three times, there's a method for that.

```suggestion
      expect(Clog).to receive(:emit).thrice.and_call_original
```

However, if you don't care about exactly how many times:

```suggestion
      expect(Clog).to receive(:emit).at_least(:once).and_call_original
```

model/postgres/postgres_server.rb
```diff
@@ -335,7 +339,28 @@ def attach_s3_policy_if_needed
     end
   end
 
+  def observe_archival_backlog(session:)
+    result = session[:ssh_session].exec!(
+      "find /dat/:version/data/pg_wal/archive_status -name '*.ready' | wc -l",
```
byucesoy: I’m not entirely sure, but would `ls dat/:version/data/pg_wal/archive_status/*.ready | wc -l`
be faster? Is there any real benefit to using find here?

Also, I’ve noticed that in some cases `ls ...` returns an error saying a file was not found. I believe this is a race condition: the shell expands the glob, and then `ls` attempts to access the file’s metadata, but the file gets deleted in between, resulting in an error.

Would `find` have a similar issue? In that scenario, do `ls` or `find` return a non-zero exit code? If so, could that cause `observe_archival_backlog` to fail?

Long story short, could you please reproduce this behavior and make sure the function correctly handles this edge case?

model/postgres/postgres_server.rb
```diff
@@ -335,7 +339,28 @@ def attach_s3_policy_if_needed
...
+      Page.from_tag_parts("PGArchivalBacklogHigh", id)&.incr_resolve
+    end
+  rescue => ex
+    Clog.emit("Failed to observe archival backlog") do
+      {postgres_server_id: id, exception: Util.exception_to_hash(ex)}
+    end
+  end
+
   FAILOVER_LABELS = ["prepare_for_unplanned_take_over", "prepare_for_planned_take_over", "wait_fencing_of_old_primary", "taking_over"].freeze
+  ARCHIVAL_BACKLOG_THRESHOLD = 10
```
byucesoy: I think 10 is quite low. How about 1000? Each WAL file is 16MB, so 1000 WAL file is ~16GB, which would be a generous buffer size for busy databases.

byucesoy (approved):

model/postgres/postgres_server.rb
aamederen: We had used something like "1000 or 5%, whatever is the smallest" in previous platforms. 1000 could be too big for small DBs?

model/postgres/postgres_server.rb
shikharbhardwaj: The race condition is exactly the reason to use `find` here.

Another is that `find` streams the matched file names, so has better memory characteristics compared to the shell expansion.

model/postgres/postgres_server.rb
shikharbhardwaj: I was considering this to be time-based, where a low throughput database could lose last 10 mins of data. But it makes sense to vary this as it would be very noisy for high throughput dbs, maybe based on the size?

shikharbhardwaj: I added the rule that @aamederen suggested here

byucesoy (approved):

PR #4349, opened 2025-12-10
List postgres databases api call returns read replica ubids

serprex:
Modify detailed serialization to also only be array of UBIDs

> [!IMPORTANT]
> The PR updates the Postgres list API to include the number of read replicas for each database, with changes in dataset query, serialization, and OpenAPI spec, along with new tests.
> 
>   - **Behavior**:
>     - `postgres_list` in `helpers/postgres.rb` now includes `:read_replicas` in the dataset query to fetch the number of replicas.
>     - `serialize_internal` in `serializers/postgres.rb` adds `num_replicas` to the serialized output, counting `pg.read_replicas`.
>     - OpenAPI spec in `openapi.yml` updated to include `num_replicas` in `PostgresDatabase` schema.
>   - **Tests**:
>     - Added tests in `postgres_spec.rb` to verify `num_replicas` is 0 when no replicas exist and correct count when replicas are present.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0d666911ab6db9a104bd09ee8895e6c43474c0d1. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli-commands/pg/post/show.rb
+++ b/cli-commands/pg/post/show.rb
@@ -40,11 +40,6 @@ UbiCli.on("pg").run_on("show") do
         data[key].each_with_index do |md, i|
           body << "  " << (i + 1).to_s << ": " << md[:id] << "  " << md[:username].to_s << "  " << md[:url] << "\n"
         end
-      when :read_replicas
-        body << "read-replicas:\n"
-        sdk_object.read_replicas.each do |rr|
-          body << "  " << rr[:location] << "/" << rr[:name] << "\n"
-        end
       when :ca_certificates
         body << "ca-certificates:\n" << data[key].to_s << "\n"
       else
--- a/helpers/postgres.rb
+++ b/helpers/postgres.rb
@@ -73,7 +73,7 @@ class Clover
   end
 
   def postgres_list(tags_param: nil)
-    dataset = dataset_authorize(@project.postgres_resources_dataset.eager(:timeline, representative_server: [:strand, vm: :vm_storage_volumes]), "Postgres:view").eager(:semaphores, :location, strand: :children)
+    dataset = dataset_authorize(@project.postgres_resources_dataset.eager(:timeline, :read_replicas, representative_server: [:strand, vm: :vm_storage_volumes]), "Postgres:view").eager(:semaphores, :location, strand: :children)
 
     if tags_param
       tags_param = tags_param.split(",")
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -2728,6 +2728,11 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/PostgresTag'
+        read_replicas:
+          description: Lists ids of read replicas for this database
+          type: array
+          items:
+            type: string
       additionalProperties: false
       required:
         - flavor
@@ -2743,6 +2748,7 @@ components:
         - maintenance_window_start_at
         - read_replica
         - tags
+        - read_replicas
     PostgresDatabaseMetrics:
       type: object
       properties:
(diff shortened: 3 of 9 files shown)


enescakir:
Should we return the number of replicas or maybe the list of replica ubids? 🤔

openapi/openapi.yml
```diff
@@ -2722,6 +2722,9 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/PostgresTag'
+        num_replicas:
```
furkansahin: we would probably want to have this "number_of_replicas" considering it's a public endpoint. 

I also agree with @enescakir 's take. We should probably return names or ubids of the replicas so that it is more useful. Just the number of replicas seems like a too niche of a return type for a very specific use-case.

serprex: replaced with `read_replicas` as array of UBIDs

serializers/postgres.rb
furkansahin: Let's remove this from the detailed view or just use this one and not add it to the basic view.

serprex: removed, read_replicas is always UBIDs

cli-commands/pg/post/show.rb
```diff
@@ -40,11 +40,6 @@
         data[key].each_with_index do |md, i|
           body << "  " << (i + 1).to_s << ": " << md[:id] << "  " << md[:username].to_s << "  " << md[:url] << "\n"
         end
-      when :read_replicas
```
jeremyevans, line 43: We should be showing the read replica ubids here one per line, not using the default behavior of calling `to_s` on the array of ubids.

helpers/postgres.rb
```diff
@@ -73,7 +73,7 @@ def postgres_post(name)
   end
 
   def postgres_list(tags_param: nil)
-    dataset = dataset_authorize(@project.postgres_resources_dataset.eager(:timeline, representative_server: [:strand, vm: :vm_storage_volumes]), "Postgres:view").eager(:semaphores, :location, strand: :children)
+    dataset = dataset_authorize(@project.postgres_resources_dataset.eager(:timeline, :read_replicas, representative_server: [:strand, vm: :vm_storage_volumes]), "Postgres:view").eager(:semaphores, :location, strand: :children)
```
jeremyevans, line 76: Not an issue with this PR, but we should probably have a single `eager` call here instead of two.

```suggestion
    dataset = dataset_authorize(@project.postgres_resources_dataset, "Postgres:view")
      .eager(:timeline, :read_replicas, representative_server: [:strand, vm: :vm_storage_volumes], :semaphores, :location, strand: :children)
```

fdr:
I think we already have too much going on in the list "terse" representation, adding a compound value that executes a scan a multiple records (rather than a one-to-one equijoin for valuable stuff) on top of it all is just Too Much. It also took quite a bit more time to work out the ramifications of what was going on given the (extremely short) commit message, as well as guess (and it's only a guess) the motive.  Please see https://github.com/ubicloud/ubicloud/blob/main/COMMIT_MESSAGES.md

fdr:
With that in mind...what are you trying to accomplish, by loading so much into the list scan, where you cannot/should not pay a de-reference for "detail"-ed representation?

serprex:
> With that in mind...what are you trying to accomplish, by loading so much into the list scan, where you cannot/should not pay a de-reference for "detail"-ed representation?

to be honest this PR has somewhat evolved, it was originally meant to only give a read replica count, usable to know from basic API endpoint whether the instance is a parent to read replicas or not

PR #4350, opened 2025-12-10
Handle empty network interface

enescakir:
We’ve seen cases where the network interface returns nil temporarily and succeeds in subsequent runs.

Instead of raising an exception, it's better to handle this condition gracefully and nap.

    {
      "backtrace": [
        "/app/prog/vnet/aws/nic_nexus.rb:103:in 'Prog::Vnet::Aws::NicNexus#assign_ipv6_address'",
        ...
      ],
      "class": "NoMethodError",
      "message": "undefined method 'ipv_6_addresses' for nil"
    }

----

> [!IMPORTANT]
> Add nap to handle nil network interface in `assign_ipv6_address` in `nic_nexus.rb` and update tests in `nic_nexus_spec.rb`.
> 
>   - **Behavior**:
>     - In `assign_ipv6_address` in `nic_nexus.rb`, add a `nap 1` to handle nil network interface before checking `ipv_6_addresses`.
>   - **Tests**:
>     - Add test in `nic_nexus_spec.rb` to verify `nap` is called when network interface is nil in `assign_ipv6_address`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 98c4bdaaed5b906c4f0785180597202c5fe0b12e. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/aws/nic_nexus.rb
+++ b/prog/vnet/aws/nic_nexus.rb
@@ -100,7 +100,10 @@ class Prog::Vnet::Aws::NicNexus < Prog::Base
   end
 
   label def assign_ipv6_address
-    client.assign_ipv_6_addresses({network_interface_id: nic.nic_aws_resource.network_interface_id, ipv_6_address_count: 1}) if get_network_interface.ipv_6_addresses.empty?
+    nap 1 unless (network_interface = get_network_interface)
+    if network_interface.ipv_6_addresses.empty?
+      client.assign_ipv_6_addresses({network_interface_id: nic.nic_aws_resource.network_interface_id, ipv_6_address_count: 1})
+    end
     hop_wait_network_interface_created
   end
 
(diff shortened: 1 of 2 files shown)


prog/vnet/aws/nic_nexus.rb
```diff
@@ -100,7 +100,10 @@ def before_run
   end
 
   label def assign_ipv6_address
-    client.assign_ipv_6_addresses({network_interface_id: nic.nic_aws_resource.network_interface_id, ipv_6_address_count: 1}) if get_network_interface.ipv_6_addresses.empty?
+    nap 1 unless (network_interface = get_network_interface)
```
jeremyevans, line 103: This will result in an infinite `nap 1` loop with no notification if the the network interface is never assigned. Is that considered sufficiently unlikely that we shouldn't address it?

enescakir: We have a 3 minute deadline

https://github.com/ubicloud/ubicloud/blob/main/prog/vnet/aws/nic_nexus.rb#L15

So we will be getting paged in 3 minutes if it stuck

enescakir:
We just clicked buttons at the same time :D 

CleanShot 2025-12-11 at 09 38 30@2x

furkansahin (approved):
thank you

PR #4354, opened 2025-12-10
Avoid using PATCH requests for web routes

jeremyevans:
Browser forms do not support PATCH, so you need to use javascript. Switching to POST for web routes avoids the need to javascript and makes the specs simpler, since you can use the normal Capybara API.

I previously handled similar cases using `r.api_patch_web_post`, but that is an ugly name, and this behavior will extend to all future web routes, so it makes sense to change the default `r.patch` behavior. This updates the two `r.api_patch_web_post` calls to use `r.patch`.

While removing the javascript code, I noticed I neglected to remove the inline-edit javascript in a6a8b8b5edc63cbcb2e80cba20346b29e9450f71, so this includes a commit to remove that.

Best viewed ignoring whitespace differences.

----

> [!IMPORTANT]
> Switch web routes from PATCH to POST, update related code, and remove unnecessary JavaScript.
> 
>   - **Behavior**:
>     - Change `r.patch` to use POST for web routes in `helpers/general.rb`.
>     - Update `r.api_patch_web_post` calls to `r.patch` in `firewall.rb` and `postgres.rb`.
>     - Remove JavaScript handling for PATCH in `app.js`.
>   - **Views**:
>     - Update form actions to use POST in `high_availability.erb` and `resize.erb`.
>   - **Tests**:
>     - Modify tests in `postgres_spec.rb` to reflect the change from PATCH to POST.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9335e62e10e69623544291ad9b000332afa4a3ba. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -68,12 +68,8 @@ $("#new-ace-btn").on("click", function (event) {
 });
 
 $(".delete-btn").on("click", function (event) {
-  let url = $(this).data("url");
-  let csrf = $(this).data("csrf");
   let confirmation = $(this).data("confirmation");
   let confirmationMessage = $(this).data("confirmation-message") || "Are you sure to delete?";
-  let redirect = $(this).data("redirect");
-  let method = $(this).data("method");
 
   if (confirmation) {
     if (prompt(`${confirmationMessage}\nPlease type "${confirmation}" to confirm deletion`, "") != confirmation) {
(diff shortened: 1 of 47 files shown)


jeremyevans:
@byucesoy Can you please review when you have time?

byucesoy (approved):
Looks good. with this change, method parameter can be dropped from `resource_creation_form.erb`

jeremyevans:
> Looks good. with this change, method parameter can be dropped from `resource_creation_form.erb`

Thanks for that information. I'll handle that in a follow-up PR.

PR #632, opened 2023-09-19
Introduce Clog

fdr:
a.k.a. CLover LOGging.  Also there often can be too many logs, which evokes a the other meaning of the word clog.  Expansion of meaning to the type of shoes is an exercise for the reader.

You would use it like this:

    # Strand ubid gets included in metadata automatically
    Clog.emit("A message")

And like this, when you have metadata:

    Clog.emit("A message") { { my_metadata_name: { field1: ..., field2: ... } } }

I also add logging in a few places.

There are some general ideas that I'm combining from years past with Clog, in no particular order:

1. The message component is almost always static

   It's much easier to search source code and perform log system queries if the log message is a static string, and dynamic information is in metadata.

   Inspiration: exception aggregating systems are more reliable when the Exceptions have messages that do not embed dynamic text.

2. Use thread local variables for automatic log enrichment.

   At first, I used real thread local variables, but changed this to use `Thread.name`, which can improve thread debug rendering as well as being usable to enrich logs.  If we add more metadata to enrich logging output, I suggest use of Thread local variables set by the Dispatcher or Strand framework code.

   A nice consequence of this is that simple log messages that only want to include the UBID of what is being operated upon can obtain this by logging a static message only.

   Inspiration: systemd and syslog both perform automatic enrichment with at least PID, and in the former case, with unit names and other data, without any special considerations by the application emitting output.  And it's pretty useful.

3. Use (mostly) a single hash for metadata, with a unique-ish name.

   Instead of doing something like:

        { a_field: "data", b_field: "more data" }

   Do:

        { going_concern: { a_field: "data", b_field: "more data" } }

   This makes querying a lot easier.

   Inspiration: the defunct logging service "Timber" suggested this style, and it seemed like a pretty good idea after using it for a while.

   There can be aberrations to this, e.g. if there's a standardized representation.  Then you might have something like:

        { well_known_metadata_or_representation_name: { ... },
          going_concern: { a_field: "data", b_field: "more data" } }

4. Use blocks to compute dynamic fields for logs.

   `Clog.emit` takes a block that should yield the metadata included with the log line.  There are two reasons I want to experiment with using blocks to do this:

   (a) It contains the code that exists for logging when it is not as trivial as a single expression in a hash literal.

   (b) In event we do the typical DEBUG/INFO/WARN levels, you can avoid calling the block.  However, that is not implemented here, because I'm not sure if it's necessary.

[...]

PR #4356, opened 2025-12-11
Inference Router Endpoint Monitoring

aamederen:
By using the new `stats` endpoint of inference router, we aim to collect the health status of inference router endpoints and alert when needed.

This is planned to be merged and deployed after all inference router replicas are updated.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -231,7 +231,7 @@ module Config
   optional :huggingface_token, string, clear: true
   override :inference_dns_zone, "ai.ubicloud.com", string
   optional :inference_router_access_token, string, clear: true
-  override :inference_router_release_tag, "v0.1.0", string
+  override :inference_router_release_tag, "v0.1.8", string
 
   # DNS
   optional :dns_service_project_id, uuid
--- a/prog/ai/inference_router_replica_nexus.rb
+++ b/prog/ai/inference_router_replica_nexus.rb
@@ -13,6 +13,17 @@ class Prog::Ai::InferenceRouterReplicaNexus < Prog::Base
 
   def_delegators :inference_router_replica, :vm, :inference_router, :load_balancer_vm_port
 
+  PAGE_TYPES = {
+    unhealthy_replica: {
+      message: "Replica %s of inference router %s is unavailable",
+      tag: "InferenceRouterReplicaUnavailable"
+    },
+    unhealthy_endpoints: {
+      message: "Replica %s of inference router %s has unhealthy endpoints",
+      tag: "InferenceRouterReplicaUnhealthyEndpoints"
+    }
+  }.freeze
+
   def self.assemble(inference_router_id)
     ubid = InferenceRouterReplica.generate_ubid
     DB.transaction do
@@ -202,7 +213,7 @@ class Prog::Ai::InferenceRouterReplicaNexus < Prog::Base
     load_balancer_vm_port.reload.state == "up"
   end
 
-  def create_page
+  def create_page(type = :unhealthy_replica, details = {})
     extra_data = {
       inference_router_ubid: inference_router.ubid,
       inference_router_location: inference_router.location.name,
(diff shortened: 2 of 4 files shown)


model/ai/inference_router_model.rb
```diff
@@ -15,6 +15,10 @@ def load_balancer
   def self.from_model_name(name)
     first(model_name: name)
   end
+
+  def name
+    model_name
+  end
```
enescakir:
```suggestion
  alias_method :name, :model_name
```

config.rb
```diff
@@ -231,7 +231,7 @@ def self.unfrozen_test?
   optional :huggingface_token, string, clear: true
   override :inference_dns_zone, "ai.ubicloud.com", string
   optional :inference_router_access_token, string, clear: true
-  override :inference_router_release_tag, "v0.1.0", string
+  override :inference_router_release_tag, "v0.17.0", string
```
bsatzger: this is overwritten in Heroku, this change currently doesn't have a direct effect. we can change it here as well, though.

prog/ai/inference_router_replica_nexus.rb
```diff
@@ -307,14 +324,26 @@ def update_config
...
+    stats_response = vm.sshable.cmd("curl -k -m 10 --no-progress-meter https://localhost:8080/stats")
+    stats = JSON.parse(stats_response)
+
+    Clog.emit("Successfully pinged inference router.", {inference_router: inference_router.ubid, replica: inference_router_replica.ubid, stats:})
+
+    project_usage = stats["usage"]
     update_billing_records(project_usage, "prompt_billing_resource", "prompt_token_count")
     update_billing_records(project_usage, "completion_billing_resource", "completion_token_count")
+
+    endpoint_healths = stats["health"]
```
bsatzger: since we use endpoint_healths only once (in the next line), we can maybe remove it and combine this and the next line?

config.rb
aamederen: Yes, I think we can even update it now before this PR. Let me look into that. In fact, this needs to be `v0.1.7` not `v0.17.0`.

aamederen: Changed in Heroku and code now

prog/ai/inference_router_replica_nexus.rb
aamederen: Thanks, makes sense, I made the change.

config.rb
```diff
@@ -231,7 +231,7 @@ def self.unfrozen_test?
   optional :huggingface_token, string, clear: true
   override :inference_dns_zone, "ai.ubicloud.com", string
   optional :inference_router_access_token, string, clear: true
-  override :inference_router_release_tag, "v0.1.0", string
+  override :inference_router_release_tag, "v0.1.7", string
```
bsatzger: It would be great if we could release and use v0.1.8.

aamederen: Sure, I'll change the release tag here and wait until v0.1.8 is released.

prog/ai/inference_router_replica_nexus.rb
```diff
@@ -214,15 +225,21 @@ def create_page
...
       vm_host_ubid: vm.vm_host.ubid,
-      vm_host_ip: vm.vm_host.sshable.host
+      vm_host_ip: vm.vm_host.sshable.host,
+      **details
     }
-    Prog::PageNexus.assemble("Replica #{inference_router_replica.ubid.to_s[0..7]} of inference router #{inference_router.name} is unavailable",
-      ["InferenceRouterReplicaUnavailable", inference_router_replica.ubid],
-      inference_router_replica.ubid, severity: "warning", extra_data:)
+
+    page_details = PAGE_TYPES[type] || fail("BUG: unknown page type #{type}")
```
enescakir: IMHO, no need to custom check, and runtime error is enough
```suggestion
    page_details = PAGE_TYPES.fetch(type)
```

prog/ai/inference_router_replica_nexus.rb
```diff
@@ -307,14 +324,25 @@ def update_config
...
+
+    Clog.emit("Successfully pinged inference router.", {inference_router: inference_router.ubid, replica: inference_router_replica.ubid, stats:})
+
+    project_usage = stats["usage"]
     update_billing_records(project_usage, "prompt_billing_resource", "prompt_token_count")
     update_billing_records(project_usage, "completion_billing_resource", "completion_token_count")
+
+    unhealthy_endpoints = stats["health"].reject { it["healthy"] }.map { it["id"] }
+    if unhealthy_endpoints.any?
+      Clog.emit("Some endpoints are unhealthy.", {inference_router: inference_router.ubid, replica: inference_router_replica.ubid, unhealthy_endpoints:})
```
enescakir: maybe
```suggestion
      Clog.emit("Some endpoints are unhealthy", {unhealthy_inference_endpoints: {inference_router: inference_router.ubid, replica: inference_router_replica.ubid, unhealthy_endpoints:}})
```

Clog naming practices: https://github.com/ubicloud/ubicloud/pull/632#issue-1902209682

aamederen: Thanks, that's a good read. I updated the Clogs in the file

aamederen:
> The first commit message is a little messed up.

That was intentional, but apparently confusing. I changed it now

PR #4360, opened 2025-12-12
Timed rotator

fdr:
> [!IMPORTANT]
> Add SSH key rotation feature with `SshKeyRotator` model, migration, and associated logic, including tests.
> 
>   - **Behavior**:
>     - Adds `SshKeyRotator` model in `model/ssh_key_rotator.rb` for managing SSH key rotations.
>     - Introduces `Prog::SshKeyRotator` in `prog/ssh_key_rotator.rb` to handle the rotation process, including creating new keys and updating authorized keys.
>     - Implements `Prog::Test::SshKeyRotation` in `prog/test/ssh_key_rotation.rb` for testing SSH key rotation.
>     - Removes `Prog::RotateSshKey` in `prog/rotate_ssh_key.rb`.
>   - **Database**:
>     - Adds migration `20251212_add_ssh_key_rotator.rb` to create `ssh_key_rotator` table with `id`, `sshable_id`, and `next_rotation_at` columns.
>   - **Models**:
>     - Updates `Sshable` model in `model/sshable.rb` to associate with `SshKeyRotator`.
>   - **Tests**:
>     - Adds tests for `SshKeyRotator` in `spec/model/ssh_key_rotator_spec.rb`.
>     - Adds tests for `Prog::SshKeyRotator` in `spec/prog/ssh_key_rotator_spec.rb`.
>     - Adds tests for `Prog::Test::SshKeyRotation` in `spec/prog/test/ssh_key_rotation_spec.rb`.
>   - **Misc**:
>     - Updates `ubid.rb` to include `TYPE_SSH_KEY_ROTATOR`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 0c972a2a920e6ffa0f4fed83b4bad331fb883e45. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fdr:
@jeremyevans there's still a bit of AI dross in here, it's actually a second design idea I had after the first one to condense things on net: instead of a one-shot rotator, build it into a record, where the one-shot / immediate behavior is via semaphore. However, I have (mostly) purged it of errors, other than the excruciating e2e debugging, which I've been playing darts with for a little while now. I think it's ready for review, though.

fdr:
I do have some question marks about daemonize and nested quoting issues in bash -c, extending from the recent parameterization work, it's a bit suspicious, there (quoting code that ends up spliced into the root shell, but correctly does not splice into the rhizome shell called synchronously, i.e. do we want double-quoting?)

migrate/20251212_add_ssh_key_rotator.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:ssh_key_rotator) do
+      # UBID.to_base32_n("sr") => 824
+      column :id, :uuid, primary_key: true, default: Sequel.lit("gen_random_ubid_uuid(824)")
+      foreign_key :sshable_id, :sshable, type: :uuid, null: false, unique: true, on_delete: :cascade
```
jeremyevans, line 8: Since this is a single row per sshable, I think it makes sense to use the same id as the sshable, and be the primary key.

```suggestion
      foreign_key :id, :sshable, type: :uuid, null: false, primary_key: true, on_delete: :cascade
```

Maybe one reason you avoided this is you want to have a strand per `ssh_key_rotator` and the sshable id is already used by another strand? Do we have that situation currently?

Considering that this is only a single timestamptz column, we could consider adding it to sshable, though I do like the separate table approach.

prog/ssh_key_rotator.rb
```diff
@@ -0,0 +1,154 @@
...
+  label def wait
+    # If no key is set yet, nap far into the future until woken
+    nap FAR_FUTURE unless sshable.raw_private_key_1
+
+    when_rotate_now_set? do
+      decr_rotate_now
+      hop_rotate_start
+    end
+
+    if Time.now >= ssh_key_rotator.next_rotation_at
```
jeremyevans: I prefer napping only as long as is necessary:

```suggestion
    seconds = ssh_key_rotator.next_rotation_at - Time.now 
    if seconds <= 0
```

prog/ssh_key_rotator.rb
```diff
@@ -0,0 +1,154 @@
...
+      hop_rotate_start
+    end
+
+    nap NAP_DURATION
```
jeremyevans:
```suggestion
    nap(seconds + 1)
```

prog/ssh_key_rotator.rb
```diff
@@ -0,0 +1,154 @@
...
+      fail "Unexpected output: #{ret}" unless ret == "target user login successful\n"
+    end
+
+    hop_rotate_finalize
+  end
+
+  label def rotate_finalize
+    changed_records = sshable.this.where(
+      Sequel.~(raw_private_key_2: nil)
+    ).update(raw_private_key_1: Sequel[:raw_private_key_2], raw_private_key_2: nil)
```
jeremyevans:
```suggestion
    changed_records = sshable.this.exclude(raw_private_key_2: nil).
      update(raw_private_key_1: Sequel[:raw_private_key_2], raw_private_key_2: nil)
```

prog/vm/host_nexus.rb
```diff
@@ -48,7 +48,9 @@ def self.assemble(sshable_hostname, location_id: Location::HETZNER_FSN1_ID, fami
 
   label def setup_ssh_keys
     # Generate a new SSH key if one is not set.
-    sshable.update(raw_private_key_1: SshKey.generate.keypair) unless sshable.raw_private_key_1
+    unless sshable.raw_private_key_1
+      sshable.update(raw_private_key_1: SshKey.generate.keypair)
+    end
```
jeremyevans: No objection, but this change seems unrelated to the rest of the commit.

prog/install_rhizome.rb
```diff
@@ -62,7 +62,6 @@ def before_run
 
   label def validate
     sshable.cmd("common/bin/validate")
-
```
jeremyevans, line 73: Same here. Stylistic changes to unrelated code should probably be in separate commits.

jeremyevans (commented):
In general, this looks good. My main comment is on the database design. While the approach in the PR is OK, I think it could be simpler.

migrate/20251212_add_ssh_key_rotator.rb
fdr: yes, that's exactly why I contorted it that way. It's almost always the case sshable does not stand alone: it's almost always on vm and vmhost. I need that separate ubid to do the join with a strand.

prog/vm/host_nexus.rb
fdr: ai dross from the churn, I'll purge it

prog/install_rhizome.rb
fdr: probably dross. thanks for flagging.

fdr:
demoting to draft for now until I can figure out what is going on with how long it takes to run E2E (some processes holding the rotation user linger)

PR #4362, opened 2025-12-12
CheckUsageAlerts optimizations

enescakir:
- **Remove mocking in CheckUsageAlerts prog**
  I was planning to add the eager(:project) improvement in the next
  commit, but when I removed the mocking in the test, it immediately
  raised Sequel::Plugins::ForbidLazyLoad. So it beat me this time :)
  
  It is a great plugin. We should remove all mocking from the tests soon.
  

- **Calculate current usage for each project once in CheckUsageAlerts**
  I noticed a few apoptoses caused by this prog in production, and there
  are some low-hanging optimizations we can apply.
  
  The current usage calculation is an expensive operation. Since a project
  can have multiple usage alerts, we should compute the usage only once
  per project rather than repeating the calculation for each alert.
  
  I checked production data, and around 30% of projects with usage alerts
  have more than one alert configured.
  

- **Do not load all invoices for current usage**
  Invoice records contain very large JSON payloads. Currently, we load all
  invoices just to determine the date of the most recent one.
  
  Fetching these full records from the database is slow and unnecessary.
  Instead, we should query only for the latest invoice date directly,
  avoiding the overhead of loading the full invoice data.
  

- **Pass current cost to usage alerts instead of recalculating it**
  Calculating the current cost is an expensive operation. Since we already
  compute it once, we should pass this value directly to the trigger
  method rather than recalculating it for each usage alert.
  

- **Remove mocking in UsageAlert model**
  

----

> [!IMPORTANT]
> Optimize `CheckUsageAlerts` by reducing redundant calculations and improving database query efficiency, while enhancing test reliability by removing mocks.
> 
>   - **Optimizations in `CheckUsageAlerts`**:
>     - Calculate current usage for each project once instead of per alert in `check_usage_alerts.rb`.
>     - Pass pre-calculated current cost to `UsageAlert#trigger` instead of recalculating.
>   - **Database Query Improvements**:
>     - Query only for the latest invoice date in `Project#current_invoice` instead of loading all invoices.
>   - **Testing Improvements**:
>     - Remove mocking in `usage_alert_spec.rb` and `check_usage_alerts_spec.rb` to use real data and improve test reliability.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 194a930b5aaf28e2ec5ebb1855dcf5df0e1a9b24. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/model/usage_alert_spec.rb
+++ b/spec/model/usage_alert_spec.rb
@@ -4,12 +4,13 @@ require_relative "spec_helper"
 
 RSpec.describe UsageAlert do
   it "trigger sends email and updates last_triggered_at" do
-    alert = described_class.new
-    expect(alert).to receive(:user).and_return(instance_double(Account, name: "dummy-name", email: "dummy-email")).at_least(:once)
-    expect(alert).to receive(:project).and_return(instance_double(Project, name: "dummy-name", ubid: "dummy-ubid", path: "dummy-path", current_invoice: instance_double(Invoice, content: {"cost" => "dummy-cost"}))).at_least(:once)
+    now = Time.now.round
+    expect(Time).to receive(:now).and_return(now).at_least(:once)
+    last_triggered_at = now - 42 * 24 * 60 * 60
+    limit = 100
+    alert = described_class.create(project_id: Project.create(name: "project1").id, user_id: Account.create(email: "user@example.com").id, name: "alert1", limit:, last_triggered_at:)
+
     expect(Util).to receive(:send_email)
-    expect(Time).to receive(:now).and_return("dummy-time")
-    expect(alert).to receive(:update).with(last_triggered_at: "dummy-time")
-    alert.trigger(123)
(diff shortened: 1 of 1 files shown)


model/project.rb
```diff
@@ -118,7 +118,7 @@ def active?
   end
 
   def current_invoice
-    begin_time = invoices.first&.end_time || Time.new(Time.now.year, Time.now.month, 1)
+    begin_time = invoices_dataset.select_map(:end_time).first || Time.new(Time.now.year, Time.now.month, 1)
```
jeremyevans: Sequel has a shortcut for getting a single column value:

```suggestion
    begin_time = invoices_dataset.get(:end_time) || Time.new(Time.now.year, Time.now.month, 1)
```

This is faster as it avoids retrieving the `end_time` for all of the projects invoices.

PR #4363, opened 2025-12-12
Add private-* DNS entries for private IP addresses

jeremyevans:
This allows easier use with private subnets. If you are connecting private subnets, with a VM in one subnet and a PostgreSQL database in a connected subnet, you could already block all public IP address access to the PostgreSQL database, and only allow access from the private IP range of the VM subnet to improve security.

This commit adds the ability for the VM to connect to the PostgreSQL database by DNS name using the private IP address. This is preferable to connecting using the private IP address, as the private IP address can change if you resize, upgrade, or failover the database. Using the DNS name, the VM will connect to the new private IP address after those changes.

This is done for load balancers as well, for the same reason. As kubernetes uses load balancers, I believe this also allows similar support for kubernetes.

----

> [!IMPORTANT]
> Adds 'private-' prefixed DNS records for private IPs, allowing stable DNS-based connections to resources like PostgreSQL databases.
> 
>   - **Behavior**:
>     - Adds 'private-' prefixed DNS records for private IPs in `refresh_dns_record` in `postgres_resource_nexus.rb` and `rewrite_dns_records` in `load_balancer_nexus.rb`.
>     - Ensures VMs can connect to resources using DNS names that reflect private IPs, which are stable across changes.
>   - **Validation**:
>     - Adds validation in `load_balancer.rb` to disallow names starting with 'private-'.
>   - **Tests**:
>     - Updates tests in `load_balancer_spec.rb`, `vm_spec.rb`, `postgres_resource_nexus_spec.rb`, and `load_balancer_nexus_spec.rb` to cover new DNS record functionality and validation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 574240c64ea4a46a11c24b5b04251550bd4873f0. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -191,6 +191,11 @@ class LoadBalancer < Sequel::Model
     IPV6 = "ipv6"
     DUAL = "dual"
   end
+
+  def validate
+    super
+    errors.add(:name, "cannot start with 'private-'") if name&.start_with?("private-")
+  end
 end
 
 # Table: load_balancer
--- a/spec/model/load_balancer_spec.rb
+++ b/spec/model/load_balancer_spec.rb
@@ -20,6 +20,18 @@ RSpec.describe LoadBalancer do
     expect(ps.errors[:name]).to eq ["cannot be exactly 26 numbers/lowercase characters starting with 1b to avoid overlap with id format"]
   end
 
+  it "disallows private- name prefix" do
+    ps = described_class.new(name: "private-a")
+    ps.validate
+    expect(ps.errors[:name]).to eq ["cannot start with 'private-'"]
+  end
+
+  it "only includes single error if name is nil" do
+    ps = described_class.new(name: nil)
+    ps.validate
+    expect(ps.errors[:name]).to eq ["is not present"]
+  end
+
   it "allows inference endpoint ubid format as name" do
     ps = described_class.new(name: InferenceEndpoint.generate_ubid.to_s)
     ps.validate


fdr:
the people demand apps.

jeremyevans:
> I'm not sure if this is the way I'd want to go. I think I would prefer if the private ip addresses are served from the dnsmasq instances in the VMs without hitting the public dns servers at all.

We discussed the pros and cons of that approach compared to this approach, and @fdr decided to use this approach.

jeremyevans:
@byucesoy sorry for reflagging this for review. I had pushed another commit, but I decided to remove it and leave this as you initially approved it. So it doesn't need another review.

jeremyevans:
@byucesoy Sorry to flip flop on this, but I found that we need to disallow load balancers with a name starting with `private-`, to avoid the private DNS name for a load balancer with the name of "foo" having the same DNS entry as the public DNS name for a load balancer with the name of "private-foo". So the last commit does need a review.

fdr:
@furkansahin I would say, cramming them into knotdns is *a* solution that fits the minimum budget, but holding it in a separate DNS server, whether another Knot (this is the alternative they proffer for split horizon, too, there's a forwarding module/extension) or dnsmasq, is probably strictly better...but somewhat more work. What do you think? If you feel strongly about it, it's sort of in your area, you could make the decision to supervise it in relation to managed subnets. 

The upsides are obvious: great robustness, stable speed, hysteresis in event of failure. Downsides are, it's not so good if you have sparse queries (i.e. we waste a lot of time syndicating `hosts` to many VMs rather than to a DNS server with customary caching), plus, another thing to keep de-bugged.

fdr:
@furkansahin This might also impact your calculations (though I'm not sure in which direction): we should account for user services too. For example, my https://github.com/ubicloud/ubidon demo currently hardcodes an IP address for self-hosted Redis. This patch doesn't directly address that issue, but it does reveal a gap: users will reasonably want to assign, and reassign hostnames to their internal resources.

These hostnames create TLS considerations. While TLS isn't required when traversing private addresses, I expect most users will prefer the extra security layer except in extreme performance scenarios. Some might even want public certificate authorities for a middle-ground approach (e.g. libpq's relatively new `sslrootcert=system`). That would require a public DNS record for certificate validation while keeping the listening address private.

furkansahin:
I was thinking this:
> @furkansahin I would say, cramming them into knotdns is a solution that fits the minimum budget, but holding it in a separate DNS server, whether another Knot (this is the alternative they proffer for split horizon, too, there's a forwarding module/extension) or dnsmasq, is probably strictly better...but somewhat more work. What do you think? If you feel strongly about it, it's sort of in your area, you could make the decision to supervise it in relation to managed subnets.

Regarding this:
> We discussed the pros and cons of that approach compared to this approach, and @fdr decided to use this approach.

I re-read the thread and I don't think we discussed the "way" to implement private-* dns there.

I also agree that this is the cheapest way to go there. Considering I am hammered with the other threads, I don't think I have enough slice to supervise a thread on this area. Therefore, I respect the implementer's decision to pick the way to provide the functionality. 

I still wanted to voice my preference to provide it in dnsmasq but it requires another prog implementation that is in the private_subnet perimeter.

fdr:
> Regarding this:
> 
> > We discussed the pros and cons of that approach compared to this approach, and @fdr decided to use this approach.
> 
> I re-read the thread and I don't think we discussed the "way" to implement private-* dns there.

Yeah, I tried to remain studiously neutral on this point. private- enables a cheaper way, but doesn't prescribe it

> I also agree that this is the cheapest way to go there. Considering I am hammered with the other threads, I don't think I have enough slice to supervise a thread on this area. Therefore, I respect the implementer's decision to pick the way to provide the functionality.
> 
> I still wanted to voice my preference to provide it in dnsmasq but it requires another prog implementation that is in the private_subnet perimeter.

hmm. I was thinking of migration cost, might be minor, in the sense that what good is resolving the private addresses from the outside, so we could do a more elaborate job of slimming the public exposure, but I suppose it's a perhaps annoying interface change. What do you think?

PR #4366, opened 2025-12-13
Remove unused association methods

jeremyevans:
Such methods consume unnecessary memory, resulting in longer full garbage collections.

This uses Sequel's unused_associations plugin to detect the unused associations. Unfortunately, there are a number of false positives, due to mocking in the specs. I've flagged those associations so they aren't considered unused, but such flags should be removed after the unnecessary mocking is removed from the specs (and association mocking should always be considered unnecessary).

This runs the unused associations check in CI, so that code will fail CI if it introduces an unused association, or stops using a currently used association without removing it or removing the unused association methods.

----

> [!IMPORTANT]
> Remove unused association methods and add CI check to prevent future unused associations using Sequel's `unused_associations` plugin.
> 
>   - **Behavior**:
>     - Removes unused association methods across various models to optimize memory usage and improve garbage collection.
>     - Introduces a CI check using Sequel's `unused_associations` plugin to prevent future introduction of unused associations.
>     - Flags false positives in tests due to mocking, with a note to remove these flags after unnecessary mocking is addressed.
>   - **Implementation**:
>     - Updates `Gemfile` to use a specific Sequel version from GitHub.
>     - Adds `unused_associations_check` task in `Rakefile` to report removable associations and options.
>     - Modifies `model.rb` to conditionally load the `unused_associations` plugin based on the `UNUSED_ASSOCIATIONS` environment variable.
>   - **Models**:
>     - Adds `read_only: true` and other options to various associations in models like `AccessControlEntry`, `Account`, `Address`, `InferenceEndpoint`, and `InferenceRouter` to mark them as used or unused appropriately.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 9b2fea4debf0df738d24164f0cfd8b8cb3ca2b14. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/vm_host_slice.rb
+++ b/model/vm_host_slice.rb
@@ -4,9 +4,9 @@ require_relative "../model"
 
 class VmHostSlice < Sequel::Model
   one_to_one :strand, key: :id
-  many_to_one :vm_host, read_only: true
+  many_to_one :vm_host
   one_to_many :vms, read_only: true, is_used: true
-  one_to_many :cpus, class: :VmHostCpu, read_only: true
+  one_to_many :cpus, class: :VmHostCpu, adder: nil, remover: nil, is_used: true
 
   plugin ResourceMethods
   plugin SemaphoreMethods, :destroy, :start_after_host_reboot, :checkup
--- a/model/firewall.rb
+++ b/model/firewall.rb
@@ -3,7 +3,7 @@
 require_relative "../model"
 
 class Firewall < Sequel::Model
-  many_to_one :project, read_only: true
+  many_to_one :project
   one_to_many :firewall_rules, order: :cidr, remover: nil, clearer: nil
   many_to_many :private_subnets
   many_to_one :location
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -4,7 +4,7 @@ require_relative "../../model"
 require_relative "../../lib/net_ssh"
 
 class KubernetesCluster < Sequel::Model
-  one_to_one :strand, key: :id, read_only: true
+  one_to_one :strand, key: :id
   many_to_one :api_server_lb, class: :LoadBalancer, read_only: true
   many_to_one :services_lb, class: :LoadBalancer
   many_to_one :private_subnet, read_only: true
--- a/model/minio/minio_server.rb
+++ b/model/minio/minio_server.rb
@@ -5,8 +5,8 @@ require_relative "../../lib/net_ssh"
 
 class MinioServer < Sequel::Model
   one_to_one :strand, key: :id, read_only: true
-  many_to_one :vm, read_only: true
-  many_to_one :pool, key: :minio_pool_id, class: :MinioPool, read_only: true
+  many_to_one :vm
+  many_to_one :pool, key: :minio_pool_id, class: :MinioPool
   one_through_one :cluster, join_table: :minio_pool, left_primary_key: :minio_pool_id, left_key: :id, class: :MinioCluster, read_only: true
 
   plugin ResourceMethods, redacted_columns: :cert, encrypted_columns: :cert_key
--- a/model/postgres/postgres_metric_destination.rb
+++ b/model/postgres/postgres_metric_destination.rb
@@ -3,6 +3,8 @@
 require_relative "../../model"
 
 class PostgresMetricDestination < Sequel::Model
+  many_to_one :postgres_resource, no_association_method: true
+
   plugin ResourceMethods, encrypted_columns: :password
 end
 
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -12,7 +12,7 @@ class Strand < Sequel::Model
   Strand.default_values[:stack] = proc { [{}] }
 
   LEASE_EXPIRATION = 120
-  many_to_one :parent, class: self, read_only: true
+  many_to_one :parent, class: self
   one_to_many :children, key: :parent_id, class: self, remover: nil, clearer: nil
   one_to_many :semaphores
 
--- a/model/victoria_metrics_resource.rb
+++ b/model/victoria_metrics_resource.rb
@@ -4,7 +4,7 @@ require_relative "../model"
 
 class VictoriaMetricsResource < Sequel::Model
   one_to_one :strand, key: :id
-  many_to_one :project, read_only: true
+  many_to_one :project
   many_to_one :location, read_only: true
   one_to_many :servers, class: :VictoriaMetricsServer, is_used: true
   many_to_one :private_subnet, read_only: true
--- a/model/victoria_metrics_server.rb
+++ b/model/victoria_metrics_server.rb
@@ -4,8 +4,8 @@ require_relative "../model"
 
 class VictoriaMetricsServer < Sequel::Model
   one_to_one :strand, key: :id
-  many_to_one :vm, read_only: true
-  many_to_one :resource, class: :VictoriaMetricsResource, key: :victoria_metrics_resource_id, read_only: true
+  many_to_one :vm
+  many_to_one :resource, class: :VictoriaMetricsResource, key: :victoria_metrics_resource_id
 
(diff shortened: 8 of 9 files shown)


jeremyevans:
While running the unused association check in CI works correctly, I think it takes too long:

* amd64: Adds 2:49 to CI
* arm64: Adds 4:02 to CI

In both cases, this more than doubles CI time.

Maybe we could instead run it daily?

enescakir:
Some of these might be useful for operational tasks. I'm not entirely sure which ones, but we could load all of them in pry for oncalls . 🤔

jeremyevans:
I fine with keeping any of the methods deemed helpful for on-call use, as long as specs are added for them so they don't show up as unused.

byucesoy:
> Some of these might be useful for operational tasks. I'm not entirely sure which ones, but we could load all of them in pry for oncalls . 🤔

Agreed with this. From top of my mind, I think `.project` associations are quite useful to find owner(s) of given resource.

jeremyevans:
I've removed this from CI, and added back the `project` associations (as read only unless the setter method is used). I've also switched the `strand` associations to read only, which didn't break any specs. I updated the Rake task to not complain about unused project associations, similar to how strand and semaphore associations were already handled.

model/dns_zone/dns_zone.rb
```diff
@@ -4,10 +4,9 @@
 
 class DnsZone < Sequel::Model
   one_to_one :strand, key: :id
-  many_to_one :project
-  many_to_many :dns_servers
-  one_to_many :records, class: :DnsRecord
-  one_to_one :active_billing_record, class: :BillingRecord, key: :resource_id, &:active
+  many_to_one :project, read_only: true
+  many_to_many :dns_servers, remover: nil, clearer: nil, is_used: true
+  one_to_many :records, class: :DnsRecord, remover: nil, clearer: nil
```
enescakir, line 9: What's `remover` and `clearer`? Can you add them to the commit message? I don't have any idea.

jeremyevans: Sure, I can add the information to the commit message. `remover: nil` causes it not to define the `remove_*` method, and `clearer: nil` causes it not to define the `remove_all_*` method.

jeremyevans: When I rebased, I apparently squashed some commits together without updating the commit message. I rewrote the commit message for the commit that adds association options to describe the options added.

jeremyevans:
Note that leaving these unused associations can cause problems, since you have no idea whether an unused association is even defined correctly: https://github.com/ubicloud/ubicloud/pull/4416#discussion_r2635981387

model/vhost_block_backend.rb
```diff
@@ -3,9 +3,6 @@
 require_relative "../model"
 
 class VhostBlockBackend < Sequel::Model
-  many_to_one :vm_host
-  one_to_many :vm_storage_volumes
```
enescakir, line 7: cc @pykello

model/pci_device.rb
```diff
@@ -3,9 +3,7 @@
 require_relative "../model"
 
 class PciDevice < Sequel::Model
-  many_to_one :vm_host
   many_to_one :vm
-  many_to_many :gpu_partitions, join_table: :gpu_partitions_pci_devices
```
enescakir, line 8: cc @bsatzger

model/assigned_vm_address.rb
```diff
@@ -3,8 +3,6 @@
 require_relative "../model"
 
 class AssignedVmAddress < Sequel::Model
-  many_to_one :vm, key: :dst_vm_id
-  many_to_one :address
```
enescakir, line 7: cc @furkansahin

pykello (approved):
Changes in storage related models look good to me.

jeremyevans:
I've rebased this on main. @bsatzger @furkansahin if you could please review soon, I would appreciate it. Due to the size, it is prone to conflicts. Additionally, I will be submitting another PR building on top of this later today.

model/vm_host_slice.rb
```diff
@@ -5,7 +5,6 @@
 class VmHostSlice < Sequel::Model
   one_to_one :strand, key: :id
   many_to_one :vm_host
-  one_to_many :vms
```
furkansahin, line 8: this is useful for debugging in the console

model/private_subnet_aws_resource.rb
```diff
@@ -3,8 +3,6 @@
 require_relative "../model"
 
 class PrivateSubnetAwsResource < Sequel::Model
-  many_to_one :private_subnet, key: :id
```
furkansahin, line 6: this is useful when debugging

model/nic_aws_resource.rb
```diff
@@ -3,8 +3,6 @@
 require_relative "../model"
 
 class NicAwsResource < Sequel::Model
-  many_to_one :nic, key: :id
```
furkansahin, line 6: this is useful when debugging

model/location_credential.rb
```diff
@@ -5,8 +5,8 @@
 require "aws-sdk-iam"
 
 class LocationCredential < Sequel::Model
-  plugin ResourceMethods, encrypted_columns: [:access_key, :secret_key]
   many_to_one :project
+  plugin ResourceMethods, encrypted_columns: [:access_key, :secret_key]
```
furkansahin, line 8: why is this moved?

model/location_aws_az.rb
```diff
@@ -3,7 +3,6 @@
 require_relative "../model"
 
 class LocationAwsAz < Sequel::Model
-  many_to_one :location
```
furkansahin, line 6: this is useful when debugging

model/load_balancer_vm.rb
```diff
@@ -3,7 +3,6 @@
 require_relative "../model"
 
 class LoadBalancerVm < Sequel::Model(:load_balancers_vms)
-  many_to_one :load_balancer
```
furkansahin, line 6: this is useful when debugging

model/aws_instance.rb
```diff
@@ -3,8 +3,6 @@
 require_relative "../model"
 
 class AwsInstance < Sequel::Model
-  one_to_one :vm, key: :id
```
furkansahin, line 6: this is useful when debugging

model/assigned_vm_address.rb
furkansahin: this is useful when debugging

model/address.rb
```diff
@@ -4,7 +4,6 @@
 
 class Address < Sequel::Model
   one_to_many :assigned_vm_addresses, key: :address_id, class: :AssignedVmAddress
-  one_to_many :assigned_host_addresses, key: :address_id, class: :AssignedHostAddress
```
furkansahin, line 7: this is useful when debugging

model/location_credential.rb
jeremyevans: Likely I removed the `project` association, since it wasn't used, but then added it back, but didn't add it back in the exact same place. I'll fix this.

jeremyevans:
@furkansahin I assume all associations that are useful when debugging only need to be read-only, so I'll add `is_used` and `read_only` association options for them.

model/aws_instance.rb
jeremyevans: This should probably be `many_to_one` instead of `one_to_one`. `one_to_one` should only be used in cases where there is a foreign key in the associated table that points to the current table. FWIW, there isn't a foreign key constraint for `aws_instance.id` referencing `vm.id`. Should there be? Do we want to support cases where we have `aws_instance.id` does not reference a valid `vm`?

jeremyevans:
@bsatzger I merged #4488  (which you already approved) into this. Can you please review the AI-related association changes?

PR #4369, opened 2025-12-15
Add logs when we need to recursively reassign a subnet range

furkansahin:
Recently we had an incident where this recursion caused stack depth issues. Therefore, we'd like to understand what's causing the repeatitive failure.

----

> [!IMPORTANT]
> Add logging to `random_private_ipv4` in `subnet_nexus.rb` to diagnose recursion issues by logging banned or in-use subnets.
> 
>   - **Behavior**:
>     - Adds logging in `random_private_ipv4` in `subnet_nexus.rb` to emit messages when a selected IPv4 subnet is banned or already in use.
>     - Logs the specific reason for failure before retrying subnet selection.
>   - **Misc**:
>     - No changes to existing logic, only added logging for diagnostic purposes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 455b7db3ed378faff37b39157d5d6a6779f3e7ba. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -79,7 +79,16 @@ class Prog::Vnet::SubnetNexus < Prog::Base
       random_private_ipv4(location, project, cidr_size)
     end
 
-    selected_addr = random_private_ipv4(location, project, cidr_size) if PrivateSubnet::BANNED_IPV4_SUBNETS.any? { it.rel(selected_addr) } || project.private_subnets_dataset[net4: selected_addr.to_s, location_id: location.id]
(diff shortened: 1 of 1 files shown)


jeremyevans:
> Should we also pass some kind of depth parameter and only allow it to be called, say, up to 10 times?

Handled that in https://github.com/ubicloud/ubicloud/pull/4371, which builds on top of this. It allows 1000 iterations currently, but I have no problem reducing that.

PR #4371, opened 2025-12-15
Switch from infinite recursion to bounded iteration when selecting random IPs

jeremyevans:
Allow up to 1000 iterations. If a random IP can still not be found, raise an exception.

----

> [!IMPORTANT]
> Switch from infinite recursion to bounded iteration for random IP selection, with a 1000-attempt limit, in `private_subnet.rb` and `subnet_nexus.rb`.
> 
>   - **Behavior**:
>     - Switch from infinite recursion to bounded iteration (1000 attempts) for random IP selection in `random_private_ipv4` and `random_private_ipv6`.
>     - Raise exception if no IP is found after 1000 iterations.
>   - **Functions**:
>     - Add `until_random_ip` in `subnet_nexus.rb` to handle iteration logic.
>     - Refactor `random_private_ipv4` and `random_private_ipv6` to use `until_random_ip`.
>     - Introduce private methods `_random_private_ipv4` and `_random_private_ipv6` for encapsulating IP generation logic.
>   - **Misc**:
>     - Remove recursive calls in `random_private_ipv4` and `random_private_ipv6` in `private_subnet.rb` and `subnet_nexus.rb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for b0f40aa625f72da32199a9072719adf8caa86c7c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -107,5 +107,4 @@ class Prog::Vnet::SubnetNexus < Prog::Base
     end
     raise message
   end
-  private_class_method :until_random_ip
 end
--- a/model/private_subnet.rb
+++ b/model/private_subnet.rb
@@ -108,6 +108,10 @@ class PrivateSubnet < Sequel::Model
   #   - A maximum of 256 IPs (/24) for the largest parent subnet (/16).
   #   - A minimum of 1 IP (/32) for the smallest parent subnet (/26).
   def random_private_ipv4
+    Prog::Vnet::SubnetNexus.until_random_ip("Could not find random IPv4 after 1000 iterations") { _random_private_ipv4 }
+  end
+
+  private def _random_private_ipv4
     cidr_size = [32, (net4.netmask.prefix_len + 8)].min
 
     # If the subnet size is /24 or higher like /26, exclude the first 4 and last 1 IPs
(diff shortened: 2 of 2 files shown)


prog/vnet/subnet_nexus.rb
```diff
@@ -87,9 +92,20 @@ def self.random_private_ipv4(location, project, cidr_size = 26)
...
+
+  def self.until_random_ip(message)
+    1000.times do |i|
+      if (ip = yield)
+        return ip
+      end
+    end
+    raise message
+  end
+  private_class_method :until_random_ip
```
furkansahin: For the purposes of my learning, why do we prefer meta programming here? Can we not simply follow integer passing method? I believe it would be much more easy to read the code and debug

jeremyevans: We could duplicate the logic in both `random_private_ipv4` and `random_private_ipv6`. However, I'm generally against duplicating logic if the reason behind the logic is the same in both places.

FWIW, I wouldn't consider this metaprogramming, this doesn't define classes or methods in a parametric way. This is just calling a private method for the shared logic.

@enescakir @fdr what approach would you recommend here?

jeremyevans: I found we had this same issue in `PrivateSubnet#random_private_ipv4`, so this updates the PR to use the same approach for that. This means that if you don't like the shared logic approach, we would now need to duplicate the logic in 3 places.

fdr: what's the metaprogramming? Unless I glossed over it, it looks like higher order functions. No more exotic than `.each` or `.map` or `.any?`

jeremyevans:
@furkansahin I rebased this and added specs. I'd like to get this in before we have a repeat of the issue that took down respirate.

PR #4372, opened 2025-12-15
Style/HashSyntax: EnforcedShorthandSyntax always

serprex:
Use of this syntax has been requested in past reviews

----

> [!IMPORTANT]
> Enforce shorthand hash syntax in Ruby code by updating RuboCop configuration and modifying various files to comply with this style.
> 
>   - **RuboCop Configuration**:
>     - Enforce shorthand hash syntax by setting `Style/HashSyntax: EnforcedShorthandSyntax: always` in `.rubocop.yml`.
>   - **Code Changes**:
>     - Update hash syntax to shorthand in `node_service_spec.rb`, `util.rb`, and `inference_endpoint.rb`.
>     - Apply shorthand syntax in `minio_server.rb`, `postgres_resource.rb`, and `postgres_server.rb`.
>     - Modify hash syntax in `vm.rb`, `inference_router_replica_nexus.rb`, and `download_boot_image.rb`.
>     - Update hash syntax in `learn_memory.rb`, `rotate_storage_kek.rb`, and `nexus.rb`.
>     - Apply shorthand syntax in `github.rb`, `user.rb`, and `ipsec_tunnel_spec.rb`.
>     - Modify hash syntax in `inference_router_target_nexus_spec.rb`, `postgres_server_nexus_spec.rb`, and `postgres_spec.rb`.
>     - Update hash syntax in `firewall_spec.rb`, `project_spec.rb`, and `spec_helper.rb`.
>     - Apply shorthand syntax in `login_field.erb`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 1d462d283c7449644b5ca2954f9e718accacbd64. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


migrate/20230315_rodauth.rb
```diff
@@ -36,7 +36,7 @@
       if database_type == :mysql
         {null: false}
       else
-        {null: false, default: Sequel.date_add(Sequel::CURRENT_TIMESTAMP, days: days)}
```
jeremyevans, line 39: Modifying applied migrations is generally avoided, but I think we can accept it in this case.

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -112,7 +112,7 @@ def create_volume(req, _call)
           topology = existing ? existing[:accessible_topology] : selected_topology
           CreateVolumeResponse.new(
             volume: Volume.new(
-              volume_id: volume_id,
```
jeremyevans, line 115: @aamederen @mohi-kalantari Is the CSI code supposed to only support Ruby 3.4+, or is it like rhizome and designed to support older Ruby versions? If it is like rhizome and targets Ruby 3.0+, then we should add a `kubernetes/csi/.rubocop.yml` file similar to the `rhizome/.rubocop.yml` file, so that these changes are not applied.

mohi-kalantari: I remember that grpc gem had an issue with the latest ruby version and the fix was not published yet so I had to downgrade ruby. I don't remember the compatibility matrix between versions. I'll check and come back again.

mohi-kalantari: These changes are fine to apply to the CSI code. Since shorthand hash syntax was introduced in Ruby 3.1 and its docker image is built using Ruby 3.2.

I'll later submit a PR to bump the CSI dependencies and use the latest ruby.

serprex: I would suggest still adding a .rubocop.yml to explicitly set CSI as ruby 3.4, so when ubicloud updates to 3.5 CSI stays on 3.4 until it updates

PR #4382, opened 2025-12-15
Demock serializers/postgres_spec.rb

fdr:
The decision to demock tests predates this change. AI made this refactoring affordable.

Replace instance_double mocks with real persisted model instances:
- instance_double(PostgresResource, PostgresTimeline) → real models
- Use create_postgres_resource helper for full model setup

----

> [!IMPORTANT]
> Replace mocked instances with real models in `postgres_spec.rb` and update `serialize_internal` in `postgres.rb`.
> 
>   - **Tests**:
>     - Replace `instance_double` mocks with real model instances in `postgres_spec.rb`.
>     - Use `create_postgres_resource` helper for setting up `PostgresResource`.
>     - Add tests for cases with no server and non-primary representative server.
>   - **Code**:
>     - Remove safe navigation operator in `serialize_internal` in `postgres.rb` for `latest_restore_time`.
>   - **Configuration**:
>     - Disable `RSpec/ExpectInLet` cop in `.rubocop.yml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for c4ee53d5c643d57ffc1a8f4afac388079738d080. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/spec/serializers/postgres_spec.rb
+++ b/spec/serializers/postgres_spec.rb
@@ -3,63 +3,97 @@
 require_relative "../spec_helper"
 
 RSpec.describe Serializers::Postgres do
-  let(:pg) { PostgresResource.new(name: "pg-name", location_id: Location::HETZNER_FSN1_ID).tap { it.id = "69c0f4cd-99c1-8ed0-acfe-7b013ce2fa0b" } }
+  let(:project) { Project.create(name: "pg-test-project") }
+  let(:location_id) { Location::HETZNER_FSN1_ID }
+  let(:location) { Location[location_id] }
+  let(:timeline) { PostgresTimeline.create(location_id: location_id) }
+  let(:private_subnet) {
+    PrivateSubnet.create(
+      name: "pg-subnet", project_id: project.id, location_id: location_id,
+      net4: "172.0.0.0/26", net6: "fdfa:b5aa:14a3:4a3d::/64"
+    )
+  }
+  let(:time) {
+    t = Time.now
+    expect(Time).to receive(:now).and_return(t).at_least(:once)
+    t
+  }
 
-  before do
-    allow(pg).to receive(:project).and_return(instance_double(Project, get_ff_postgres_hostname_override: nil))
+  let(:pg) {
+    PostgresResource.create(
+      name: "pg-name", superuser_password: "dummy-password", ha_type: "none",
+      target_version: "17", location_id: location_id, project_id: project.id,
+      user_config: {}, pgbouncer_user_config: {}, target_vm_size: "standard-2",
+      target_storage_size_gib: 64, private_subnet_id: private_subnet.id
+    )
+  }
+
+  def create_representative_server(primary: true)
+    vm = create_hosted_vm(project, private_subnet, "pg-vm-#{SecureRandom.hex(4)}")
+    server = PostgresServer.create(
+      timeline:, resource_id: pg.id, vm_id: vm.id,
+      representative_at: Time.now,
+      synchronization_status: "ready",
+      timeline_access: primary ? "push" : "fetch",
+      version: "17"
+    )
+    Strand.create_with_id(server, prog: "Postgres::PostgresServerNexus", label: "wait")
+    server
   end
 
-  it "can serialize when no earliest restore time" do
-    time = Time.now
-    expect(pg).to receive(:customer_firewall).and_return(nil)
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, earliest_restore_time: nil, latest_restore_time: time)).exactly(3)
-    expect(pg).to receive(:representative_server).and_return(instance_double(PostgresServer, primary?: true, vm: nil, strand: nil, storage_size_gib: 64, version: "17")).at_least(:once)
+  it "can serialize when no earliest restore time (but latest is always present)" do
+    time
+    create_representative_server(primary: true)
     data = described_class.serialize(pg, {detailed: true})
-    expect(data[:earliest_restore_time]).to be_nil
-    expect(data[:latest_restore_time]).to eq(time.utc.iso8601)
+    expect(data.fetch(:earliest_restore_time)).to be_nil
+    expect(data.fetch(:latest_restore_time)).to eq(time.utc.iso8601)
   end
 
   it "can serialize when earliest_restore_time calculation raises an exception" do
-    time = Time.now
-    expect(pg).to receive(:customer_firewall).and_return(nil)
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, latest_restore_time: time)).exactly(4)
-    expect(pg).to receive(:representative_server).and_return(instance_double(PostgresServer, primary?: true, vm: nil, strand: nil, storage_size_gib: 64, version: "17")).at_least(:once)
+    time
+    create_representative_server(primary: true)
     expect(pg.timeline).to receive(:earliest_restore_time).and_raise("error")
     data = described_class.serialize(pg, {detailed: true})
-    expect(data[:earliest_restore_time]).to be_nil
-    expect(data[:latest_restore_time]).to eq(time.utc.iso8601)
+    expect(data).not_to have_key(:earliest_restore_time)
+    expect(data.fetch(:latest_restore_time)).to eq(time.utc.iso8601)
   end
 
   it "can serialize when have earliest/latest restore times" do
-    time = Time.now
-    expect(pg).to receive(:customer_firewall).and_return(nil)
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, earliest_restore_time: time, latest_restore_time: time)).exactly(3)
-    expect(pg).to receive(:representative_server).and_return(instance_double(PostgresServer, primary?: true, vm: nil, strand: nil, storage_size_gib: 64, version: "17")).at_least(:once)
+    time
+    create_representative_server(primary: true)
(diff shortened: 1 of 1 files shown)


spec/serializers/postgres_spec.rb
```diff
@@ -3,58 +3,92 @@
...
 
   it "can serialize when no earliest/latest restore times" do
-    expect(pg).to receive(:customer_firewall).and_return(nil)
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, earliest_restore_time: nil, latest_restore_time: nil)).exactly(3)
-    expect(pg).to receive(:representative_server).and_return(instance_double(PostgresServer, primary?: true, vm: nil, strand: nil, storage_size_gib: 64, version: "17")).at_least(:once)
+    create_representative_server(primary: true)
+    # Stub pg.timeline to return our timeline let, then stub on that
+    allow(pg).to receive(:timeline).and_return(timeline)
+    allow(timeline).to receive_messages(earliest_restore_time: nil, latest_restore_time: nil)
```
jeremyevans: Could create the PostgresTimeline, and then mock it, as it is a cached association.

```suggestion
    # Create PostgresTimeline
    allow(pg.timeline).to receive_messages(earliest_restore_time: nil, latest_restore_time: nil)
```

Similar in other places in the PR.

spec/serializers/postgres_spec.rb
```diff
@@ -3,58 +3,92 @@
...
   it "can serialize when there is no server" do
-    expect(pg).to receive(:customer_firewall).and_return(nil)
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, earliest_restore_time: nil, latest_restore_time: nil))
-    expect(pg).to receive(:representative_server).and_return(nil).at_least(:once)
+    # No server created, so representative_server is nil and timeline is nil
+    # The serializer checks `pg.timeline && pg.representative_server&.primary?`
+    # so with no server, restore times should be nil
+    data = described_class.serialize(pg, {detailed: true})
+    expect(data[:primary?]).to be_nil
```
jeremyevans: Not testing what you think it is testing:

```suggestion
    expect(data.fetch(:primary)).to be_nil
```

I recommend using `fetch` instead of `[]` in other places where you are using `be_nil` and you expect the key to be present.

fdr:
that last commit was a bit of a sensitive one, changing the application

serializers/postgres.rb
```diff
@@ -40,7 +40,7 @@ def self.serialize_internal(pg, options = {})
         rescue => ex
           Clog.emit("Failed to get earliest restore time") { Util.exception_to_hash(ex) }
         end
-        base[:latest_restore_time] = pg.timeline.latest_restore_time&.utc&.iso8601
+        base[:latest_restore_time] = pg.timeline.latest_restore_time.utc.iso8601
```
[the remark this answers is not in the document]
fdr: @ellipsis-dev: check the source of `latest_restore_time`, looks impossible, what makes you think it is nillable

spec/serializers/postgres_spec.rb
```diff
@@ -3,60 +3,107 @@
...
+    PrivateSubnet.create(
+      name: "pg-subnet", project_id: project.id, location_id: location_id,
+      net4: "172.0.0.0/26", net6: "fdfa:b5aa:14a3:4a3d::/64"
+    )
+  }
 
-  before do
-    allow(pg).to receive(:project).and_return(instance_double(Project, get_ff_postgres_hostname_override: nil))
+  let(:pg) {
+    pr = PostgresResource.create(
```
jeremyevans: Not a big deal, but we typically we use `pg` local variable for `PostgresResource` instances.

spec/serializers/postgres_spec.rb
```diff
@@ -3,60 +3,107 @@
...
   end
 
-  it "can serialize when no earliest/latest restore times" do
-    expect(pg).to receive(:customer_firewall).and_return(nil)
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, earliest_restore_time: nil, latest_restore_time: nil)).exactly(3)
-    expect(pg).to receive(:representative_server).and_return(instance_double(PostgresServer, primary?: true, vm: nil, strand: nil, storage_size_gib: 64, version: "17")).at_least(:once)
+  it "can serialize when no earliest restore time (but latest is always present)" do
+    time = Time.now
+    expect(Time).to receive(:now).and_return(time).at_least(:once)
```
jeremyevans: Could move this to `let(:time)` to DRY things up (including the `now` mocking).

spec/serializers/postgres_spec.rb
```diff
@@ -3,60 +3,107 @@
...
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, earliest_restore_time: nil, latest_restore_time: nil))
-    expect(pg).to receive(:representative_server).and_return(nil).at_least(:once)
+    # No server created, so representative_server is nil and timeline is nil
+    # The serializer checks `pg.timeline && pg.representative_server&.primary?`
+    # so with no server, restore times should be nil
+    data = described_class.serialize(pg, {detailed: true})
+    expect(data.fetch(:primary)).to be_nil
+    expect(data[:earliest_restore_time]).to be_nil
+    expect(data[:latest_restore_time]).to be_nil
```
jeremyevans: If you expect these keys to exist with `nil` value, use `fetch`. If you do not expect them to exist, it's better to use `expect(data.has_key?(:latest_restore_time)).to be false` (maybe there is a more idiomatic rspec way to do that?).

Similar issue in other specs.

fdr:
alright, I think that did the trick, the nil check is something that bit us a couple of times, going to try to bludgeon the context to get that in there

fdr:
also the non-idiomatic naming is AI dross as well I'll try to bludgeon things into, more offensive was "real_sshable" or whatever it was in another pr

fdr:
re-review please

spec/serializers/postgres_spec.rb
```diff
@@ -3,61 +3,93 @@
...
+    expect(data.fetch(:earliest_restore_time)).to be_nil
+    expect(data.fetch(:latest_restore_time)).to eq(time.utc.iso8601)
   end
 
   it "can serialize when earliest_restore_time calculation raises an exception" do
-    expect(pg).to receive(:customer_firewall).and_return(nil)
-    expect(pg).to receive(:strand).and_return(instance_double(Strand, label: "start", children: [])).at_least(:once)
-    expect(pg).to receive(:timeline).and_return(instance_double(PostgresTimeline, latest_restore_time: nil)).exactly(4)
-    expect(pg).to receive(:representative_server).and_return(instance_double(PostgresServer, primary?: true, vm: nil, strand: nil, storage_size_gib: 64, version: "17")).at_least(:once)
+    expect(Time).to receive(:now).and_return(time).at_least(:once)
```
jeremyevans: I was thinking of moving this into the `let(:time)`, since if you are calling the `time` method, you are probably doing so only to mock `Time.now`.

jeremyevans (approved):
Looks good. Up to you whether you want to remove the `Time.now` method expectation into the `let`.

fdr:
@jeremyevans final residue for your OK

spec/serializers/postgres_spec.rb
```diff
@@ -3,61 +3,97 @@
...
+  let(:timeline) { PostgresTimeline.create(location_id: location_id) }
+  let(:private_subnet) {
+    PrivateSubnet.create(
+      name: "pg-subnet", project_id: project.id, location_id: location_id,
+      net4: "172.0.0.0/26", net6: "fdfa:b5aa:14a3:4a3d::/64"
+    )
+  }
+  let(:time) {
+    t = Time.now
+    allow(Time).to receive(:now).and_return(t)
```
jeremyevans: The only 3 specs where this method is called used `expect` , so we should probably use the same here.

```suggestion
    expect(Time).to receive(:now).and_return(t).at_least(:once)
```

jeremyevans (approved):
With the recommended change, OK to merge without another review.

fdr:
@jeremyevans alright, final adjustment in

jeremyevans:
Apparently, rubocop does not like `expect` in `let`. My recommendation would be to disable that cop. Looking at the documentation (https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecexpectinlet), rubocop doesn't even provide an explanation of why this is harmful. The example rubocop gives shows usage with a single spec, instead of usage in multiple specs (our situation, where using `expect` in `let` DRYs things up).

fdr:
Agree, I did research turning off all cops that are not with autocorrects in rubocop and had some success, the main thing that stopped me is it'd be nice if this computed every time it started up, but that's rather slow, at least done naively. It would take more groveling around in its manual to figure out if there's a way to do that without pre-generation of file (which may be acceptable rather than having this back and forth periodically)

fdr:
Okay, cop removed but got rebased to the beginning of the patch series (It'll be unsquashed). The diff is small, so I think you can review the last patch as the last net-add

```
git diff 24babc6ef bb4fda80f 
diff --git a/.rubocop.yml b/.rubocop.yml
index 895c93606..02f96eeff 100644
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -84,6 +84,9 @@ RSpec/SubjectStub:
 RSpec/ExpectInHook:
   Enabled: false
 
+RSpec/ExpectInLet:
+  Enabled: false
+
 RSpec/InstanceVariable:
   Enabled: false
 
```

fdr:
reordered and squashed for history clarity, but:

```
git diff bb4fda80f c4ee53d5c spec/serializers/postgres_spec.rb
<empty>
```

same stuff

PR #4387, opened 2025-12-15
Demock load_balancer_nexus_spec.rb

fdr:
The decision to demock tests predates this change. AI made this refactoring affordable.

Replace instance_double mocks with real persisted model instances:
- Create actual LoadBalancer, VM, and Cert records
- Verify semaphore changes with SemSnap
- Use factory helpers for complex object graphs

### DIFF
--- a/spec/prog/vnet/load_balancer_nexus_spec.rb
+++ b/spec/prog/vnet/load_balancer_nexus_spec.rb
@@ -22,8 +22,22 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
   }
 
   before do
-    allow(nx).to receive_messages(load_balancer: st.subject)
-    allow(Config).to receive(:load_balancer_service_hostname).and_return("lb.ubicloud.com")
+    allow(Config).to receive_messages(load_balancer_service_hostname: "lb.ubicloud.com", load_balancer_service_project_id: ps.project_id)
+  end
+
+  def create_vm_with_ips(name:, private_ipv4:, private_ipv6:, public_ipv4: nil, public_ipv6: nil)
+    nic = Prog::Vnet::NicNexus.assemble(ps.id, name: "#{name}-nic", ipv4_addr: private_ipv4, ipv6_addr: private_ipv6).subject
+    vm = Prog::Vm::Nexus.assemble("pub key", ps.project_id, name:, private_subnet_id: ps.id, nic_id: nic.id).subject
+    vm.update(ephemeral_net6: public_ipv6) if public_ipv6
+    add_ipv4_to_vm(vm, public_ipv4) if public_ipv4
+    vm
+  end
+
+  def add_lb_vm(stack: nil, **vm_args)
+    nx.load_balancer.update(stack:) if stack
+    vm = create_vm_with_ips(**vm_args)
+    nx.load_balancer.add_vm(vm)
+    vm
   end
 
   describe ".assemble" do
@@ -35,7 +49,7 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
 
     it "creates a new load balancer" do
       lb = described_class.assemble(ps.id, name: "test-lb2", src_port: 80, dst_port: 8080).subject
-      expect(LoadBalancer.count).to eq 2
+      expect(LoadBalancer.count).to eq 1
       expect(lb.project).to eq ps.project
       expect(lb.hostname).to eq "test-lb2.#{ps.ubid[-5...]}.lb.ubicloud.com"
     end
@@ -43,7 +57,7 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
     it "creates a new load balancer with custom hostname" do
       dz = DnsZone.create(project_id: ps.project_id, name: "custom.ubicloud.com")
       lb = described_class.assemble(ps.id, name: "test-lb2", src_port: 80, dst_port: 8080, custom_hostname_prefix: "test-custom-hostname", custom_hostname_dns_zone_id: dz.id).subject
-      expect(LoadBalancer.count).to eq 2
+      expect(LoadBalancer.count).to eq 1
       expect(lb.project).to eq ps.project
       expect(lb.hostname).to eq "test-custom-hostname.custom.ubicloud.com"
     end
@@ -80,8 +94,6 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
 
   describe "#create_new_cert" do
     it "creates a new cert" do
-      dns_zone = DnsZone.create(name: "test-dns-zone", project_id: nx.load_balancer.private_subnet.project_id)
-      expect(nx.load_balancer).to receive(:dns_zone).and_return(dns_zone)
       expect { nx.create_new_cert }.to hop("wait_cert_provisioning")
         .and change { Strand.where(prog: "Vnet::CertNexus").count }.from(1).to(2)
         .and change { Strand.where(prog: "Vnet::CertNexus").all.select { it.stack[0]["add_private"] }.count }.from(0).to(1)
@@ -90,8 +102,8 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
     end
 
     it "creates a cert without dns zone in development" do
-      expect(Config).to receive(:development?).and_return(true)
-      expect(nx.load_balancer).to receive(:dns_zone).and_return(nil)
+      expect(Config).to receive(:development?).at_least(:once).and_return(true)
+      expect(Config).to receive(:load_balancer_service_project_id).and_return("00000000-0000-0000-0000-000000000000")
       expect { nx.create_new_cert }.to hop("wait_cert_provisioning")
       expect(Strand.where(prog: "Vnet::CertNexus").count).to eq 2
       expect(nx.load_balancer.certs.count).to eq 2
@@ -128,7 +140,7 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
       expect(nx.load_balancer).to receive(:need_certificates?).and_return(false)
       expect { nx.wait_cert_provisioning }.to hop("wait")
       expect(st.reload.stack[0].fetch("cert")).to be_nil
-      expect(Strand.where(prog: "Vnet::CertServer", label: "reshare_certificate").count).to eq 0
+      expect(Strand.where(prog: "Vnet::CertServer", label: "reshare_certificate").all).to be_empty
     end
   end
 
@@ -157,7 +169,7 @@ RSpec.describe Prog::Vnet::LoadBalancerNexus do
       expect(nx).to receive(:reap).and_yield
 
       expect { nx.wait_cert_broadcast }.to hop("wait")
-      expect(Semaphore[name: "destroy", strand_id: cert_to_remove.id]).not_to be_nil
+        .and change { cert_to_remove.strand.semaphores_dataset.where(name: "destroy").count }.from(0).to(1)
       expect(nx.load_balancer.reload.certs.count).to eq 1
       expect(st.reload.stack[0].fetch("cert")).to be_nil
     end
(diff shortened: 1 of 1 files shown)


spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -21,6 +21,23 @@
...
+    @vm_counter += 1
+    vm = Prog::Vm::Nexus.assemble("pub key", ps.project_id, name: "dns-vm-#{@vm_counter}", private_subnet_id: ps.id).subject
+    if ipv6
+      # Set ephemeral_net6 to produce ip6_string via .nth(2)
+      prefix = format("fd10:9b0b:6b4b:%04x::/79", @vm_counter)
+      vm.update(ephemeral_net6: prefix)
+    end
+    if ipv4
+      # Generate unique IPv4 address
+      ip = format("10.0.%d.%d", @vm_counter / 256, @vm_counter % 256)
```
jeremyevans: This seems over engineered, unless a single spec is actually creating more than 256 VMs. That said, better to a single calculation instead of two.

```suggestion
      ip = format("10.0.%d.%d", *@vm_counter.divmod(256))
```

fdr:
will hold since this kind of "rebasing" is very cheap with the keyboard smashing abilities of llms.

fdr:
okay, it's been modernized.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
-    it "hops to wait if reap is a success" do
-      expect(nx).to receive(:reap).and_yield
-      expect(nx.load_balancer.private_subnet).to receive(:incr_update_firewall_rules)
-      expect(nx.load_balancer).to receive(:dns_zone).and_return(dns_zone)
-      expect(dns_zone).to receive(:delete_record).with(record_name: nx.load_balancer.hostname)
-      expect(nx.load_balancer).to receive(:destroy)
+    it "exits and cleans up if reap is a success" do
+      lb_id = nx.load_balancer.id
       expect { nx.wait_all_vms_removed }.to exit({"msg" => "load balancer deleted"})
+      expect(LoadBalancer[lb_id]).to be_nil
```
jeremyevans:
```suggestion
      expect(nx.load_balancer).not_to exist
```

Similar cases elsewhere.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
+      nic = Prog::Vnet::NicNexus.assemble(ps.id, name: "#{name}-nic", ipv4_addr: private_ipv4, ipv6_addr: private_ipv6).subject
+      vm = Prog::Vm::Nexus.assemble("pub key", ps.project_id, name:, private_subnet_id: ps.id, nic_id: nic.id).subject
+      vm.update(ephemeral_net6: public_ipv6) if public_ipv6
+      add_ipv4_to_vm(vm, public_ipv4) if public_ipv4
+      vm
+    end
+
+    def dns_records_for(hostname)
+      DnsRecord.where(dns_zone_id: dns_zone.id, tombstoned: false)
+        .where(Sequel.like(:name, "%#{hostname}.%") | Sequel.like(:name, "#{hostname}.%")).all
```
jeremyevans: `%` in `LIKE` matches 0 or more characters, so I don't think you need the separate check without the prefix. Note that using interpolation into a LIKE pattern can be risky if the interpolated value contains LIKE metacharaters. I don't believe it does in this case, and this is only for the specs, but in more general cases, `escape_like` can be used. Might be best to use it anyway, so anyone copying this code uses a better pattern:

```suggestion
        .where(Sequel.like(:name, "%#{DB.dataset.escape_like(hostname)}.%")).all
```

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
-      expect(dns_zone).to receive(:insert_record).with(record_name: st.subject.hostname, type: "A", data: "192.168.1.0", ttl: 10)
-      expect(dns_zone).to receive(:insert_record).with(record_name: "private-#{st.subject.hostname}", type: "A", data: "1.2.3.4", ttl: 10)
-      expect(dns_zone).to receive(:insert_record).with(record_name: st.subject.hostname, type: "AAAA", data: "fd10:9b0b:6b4b:8fb0::2", ttl: 10)
-      expect(dns_zone).to receive(:insert_record).with(record_name: "private-#{st.subject.hostname}", type: "AAAA", data: "fd10:9b0b:6b4b:8fb2::2", ttl: 10)
+      nx.load_balancer.update(stack: "ipv4")
+      vm = create_vm_with_ips(
+        name: "dns-vm",
+        private_ipv4: "1.2.3.4/32",
+        private_ipv6: "fd10:9b0b:6b4b:8fb2::/64",
+        public_ipv4: "192.168.1.0/32"
```
jeremyevans: Not sure it matters, but the private and public addresses are reversed here. Similar issue elsewhere.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
-      expect(nx.load_balancer).to receive(:destroy)
+    it "exits without deleting dns record if no dns zone" do
+      expect(Config).to receive(:load_balancer_service_project_id).and_return("00000000-0000-0000-0000-000000000000")
+      lb_id = nx.load_balancer.id
       expect { nx.wait_all_vms_removed }.to exit({"msg" => "load balancer deleted"})
+      expect(LoadBalancer[lb_id]).to be_nil
     end
   end
 
   describe "#rewrite_dns_records" do
```
jeremyevans, line 250: Unless I missed something, there isn't a spec that tests that for dual stack load balancers, both IPv4 and IPv6 addresses are created. Not wrong, and not currently needed for coverage, but seems reasonable to add.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
+      expect { nx.rewrite_dns_records }.to hop("wait")
+
+      hostname = nx.load_balancer.hostname
+      records = dns_records_for(hostname)
+      expect(records.map { [it.name.chomp("."), it.type, it.data] }).to contain_exactly(
+        ["private-#{hostname}", "AAAA", "fd10:9b0b:6b4b:8fb2::2"]
+      )
     end
 
     it "does not create ipv6 dns record if stack is ipv4" do
```
jeremyevans: This now appears to duplicate the `rewrites the dns records` spec.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
+      records = dns_records_for(hostname)
+      expect(records.map { [it.name.chomp("."), it.type, it.data] }).to contain_exactly(
+        [hostname, "A", "192.168.1.0"],
+        ["private-#{hostname}", "A", "1.2.3.4"]
+      )
     end
   end
 
   describe ".need_to_rewrite_dns_records?" do
+    def create_vm_with_ips(name:, private_ipv4:, private_ipv6:, public_ipv4: nil, public_ipv6: nil)
```
jeremyevans: This appears to duplicate the `create_vm_with_ips` helper method in the previous example group. Might be worth moving that up a level and removing this.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
+        name: "ipv4-test-vm",
+        private_ipv4: "192.168.1.0/32",
+        private_ipv6: "fd10:9b0b:6b4b:8fb0::/64",
+        public_ipv4: "192.168.1.0/32"
+      )
+      nx.load_balancer.add_vm(vm)
       expect(nx.need_to_rewrite_dns_records?).to be true
     end
 
     it "returns true if dns record is missing for ipv6" do
```
jeremyevans: This appears to now test the same thing as the previous spec. Both of these specs should probably be updated, one to have the IPv6 DNS records already present, the other to have the IPv4 DNS records already present.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -222,127 +222,198 @@
...
+        name: "ipv6-test-vm",
+        private_ipv4: "192.168.1.0/32",
+        private_ipv6: "fd10:9b0b:6b4b:8fb0::/64",
+        public_ipv6: "fd10:9b0b:6b4b:8fb0::/64"
+      )
+      nx.load_balancer.add_vm(vm)
       expect(nx.need_to_rewrite_dns_records?).to be true
     end
 
     it "returns false if dns record is present for ipv4 and lb is not ipv6 enabled" do
```
jeremyevans: This spec and the next one only check for the public DNS records, they don't check for the private DNS records. That makes sense, as the prog only checks for the public DNS records. During the transition period, that's probably best. However, once we start publishing private DNS for all load balancers, we should probably have the prog check for both the private and public addresses.

I'm not sure it is worth having every load balancer prog check for missing DNS records every 5 seconds is a good tradeoff.  I would assume if DNS should be changed, whatever is causing them to be changed should increment the semaphore for the load balancer. These checks were added in  be727a2183b92ebc743f74898e00cfcf9e08e55a (PR #2220), but no reasoning is given, nor was their discussion on why this approach was considered better. I assume we must have had an issue where DNS records should have been updated and weren't, and instead of (or in addition to) fixing those cases, we decided to add these checks. @furkansahin @bsatzger do you remember the context for this? What are your thoughts?

jeremyevans (approved):

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: yeah that's not good

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: would you want .all not to exist? to see the records? you've given that feedback a few times

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: that is a sound reason, to prevent duping. it's a tricky sub-quoting issue, not as bad as sql injection generally, but not so good.

spec/prog/vnet/load_balancer_nexus_spec.rb
jeremyevans: For a single record, `expect(obj).not_to exist` works fine. For multiple records, `expect(dataset.all).to be_empty`. For cases like this, I prefer `to exist` as opposed to a separate lookup and checking `to be_nil` on that.

spec/prog/vnet/load_balancer_nexus_spec.rb
furkansahin: Yes, a VM can be created and before any IP is assigned, it could be added to a load balancer as well, depending on the race condition, we might not have an IP address assigned yet while creating the dns records.

jeremyevans: If that is the only reason, assuming we don't allow changing IPs after a VM is assigned an IP, wouldn't it be better to nap in `rewrite_dns_records` if a load balancer is IPv4 enabled and the VM doesn't have an IPv4 address assigned (or IPv6 enabled and VM doesn't have an IPv6 address)? That seems more efficient than having the load balancer check DNS records every 5 seconds.

```diff
diff --git a/prog/vnet/load_balancer_nexus.rb b/prog/vnet/load_balancer_nexus.rb
index 79921a1d9..2158cd95b 100644
--- a/prog/vnet/load_balancer_nexus.rb
+++ b/prog/vnet/load_balancer_nexus.rb
@@ -181,13 +181,15 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
       dns_zone.delete_record(record_name: hostname)
       dns_zone.delete_record(record_name: private_hostname)
 
+      ip_info = []
       load_balancer.vms_to_dns.each do |vm|
-        ip_info = []
-
         # Insert IPv4 record if stack is ipv4 or dual, and vm has IPv4
         if load_balancer.ipv4_enabled?
           if vm.ip4_string
             ip_info << [vm.ip4_string, "A", hostname]
+          elsif vm.ip4_enabled
+            # VM will have a public IPv4 address, but it is not assigned yet
+            nap 5
           end
           ip_info << [vm.private_ipv4_string, "A", private_hostname]
         end
@@ -195,18 +197,21 @@ class Prog::Vnet::LoadBalancerNexus < Prog::Base
         if load_balancer.ipv6_enabled?
           if vm.ip6_string
             ip_info << [vm.ip6_string, "AAAA", hostname]
+          else
+            # VM public IPv6 address not assigned yet
+            nap 5
           end
           ip_info << [vm.private_ipv6_string, "AAAA", private_hostname]
         end
+      end
 
-        ip_info.each do |data, type, record_name|
-          dns_zone.insert_record(
-            record_name:,
-            type:,
-            ttl: 10,
-            data:
-          )
-        end
+      ip_info.each do |data, type, record_name|
+        dns_zone.insert_record(
+          record_name:,
+          type:,
+          ttl: 10,
+          data:
+        )
       end
     end
 

```

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: okay, two nested loops, though a bit cryptic, touch much more of the combinatorics space

fdr:
I had to rebase, but the new code is at and after 
[Extract add_lb_vm helper (9 occurrences)](https://github.com/ubicloud/ubicloud/pull/4387/commits/5cd94699f737653208e1b48eebe4e205a16f92c4)

fdr:
looks like the base changed again, somewhat more invasively this time, I dropped some changes.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -157,7 +169,7 @@
       expect(nx).to receive(:reap).and_yield
 
       expect { nx.wait_cert_broadcast }.to hop("wait")
-      expect(Semaphore[name: "destroy", strand_id: cert_to_remove.id]).not_to be_nil
+      expect(cert_to_remove.strand.semaphores_dataset.first(name: "destroy")).to exist
```
jeremyevans: This doesn't handle a failure case correctly. You should use `.to exist` when you already have an known object. If there isn't such an object, `first` returns `nil`, and you end up with `expected nil to exist but it does not respond to either `exist?` or `exists?`

```suggestion
      expect(cert_to_remove.strand.semaphores_dataset.first(name: "destroy")).not_to be_nil
```

or (raise exception if the semaphore is not found):

```suggestion
      cert_to_remove.strand.semaphores_dataset.first!(name: "destroy")
```

Same issue other places.

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -231,119 +232,100 @@
...
-      expect(dns_zone).to receive(:delete_record).with(record_name: st.subject.hostname)
-      expect(dns_zone).to receive(:delete_record).with(record_name: "private.#{st.subject.hostname}")
-      expect(dns_zone).to receive(:insert_record).with(record_name: st.subject.hostname, type: "A", data: "192.168.1.0", ttl: 10)
-      expect(dns_zone).to receive(:insert_record).with(record_name: "private.#{st.subject.hostname}", type: "A", data: "1.2.3.4", ttl: 10)
-      expect(dns_zone).to receive(:insert_record).with(record_name: st.subject.hostname, type: "AAAA", data: "fd10:9b0b:6b4b:8fb0::2", ttl: 10)
-      expect(dns_zone).to receive(:insert_record).with(record_name: "private.#{st.subject.hostname}", type: "AAAA", data: "fd10:9b0b:6b4b:8fb2::2", ttl: 10)
-      expect { nx.rewrite_dns_records }.to hop("wait")
+    def dns_records_for(hostname)
+      DnsRecord.where(dns_zone_id: dns_zone.id, tombstoned: false)
+        .where(Sequel.like(:name, "%#{DB.dataset.escape_like(hostname)}.%") | Sequel.like(:name, "#{DB.dataset.escape_like(hostname)}.%")).all
```
jeremyevans: As mentioned in an earlier comment, `%` matches zero or more characters, so it isn't necessary to have both variants:

```suggestion
        .where(Sequel.like(:name, "%#{DB.dataset.escape_like(hostname)}.%")).all
```

spec/prog/vnet/load_balancer_nexus_spec.rb
```diff
@@ -231,119 +232,100 @@
...
+
+          it "#{stack} stack, pub4=#{has_pub4}, pub6=#{has_pub6}" do
+            vm_args = {name: "vm", private_ipv4: "10.0.0.1/32", private_ipv6: "fd10:9b0b:6b4b:8fb2::/64"}
+            vm_args[:public_ipv4] = "203.0.113.1/32" if has_pub4
+            vm_args[:public_ipv6] = "2001:db8:1::/64" if has_pub6
+            add_lb_vm(stack:, **vm_args)
+
+            if expects_nap
+              expect { nx.rewrite_dns_records }.to nap(5)
+              expect(DnsRecord.where(dns_zone_id: dns_zone.id, tombstoned: false).count).to eq 0
```
jeremyevans:
```suggestion
              expect(DnsRecord.where(dns_zone_id: dns_zone.id, tombstoned: false).all).to be empty
```

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: Interesting that this snuck back in. I'll put in a guard for that.

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: that is quite the bug...I never did know about `first!` after all these years. I can think of many codes over the years that would benefit from that.

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: root cause, simple enough: a regexp screened for these, it didn't find the one without parens.

spec/prog/vnet/load_balancer_nexus_spec.rb
fdr: I decided the best thing to do here was use a change block, which the only reason I haven't been applying this everywhere is it churns up too much. but in this case, it seems fine.

fdr:
git diff 09145ee6e 5ab52628f empty

fdr:
clicked rebase...now waiting ci

PR #4394, opened 2025-12-15
Add boot image version to runner logs

enescakir:
We already include a conclusion field in runner_completed logs, which allows us to analyze success rates by location or VM host. Adding the boot image version to these logs will also let us track success rates per boot image.

This will make it easier to monitor the impact of new boot image rollouts and detect regressions early. Currently, we infer this indirectly via VM hosts while downloading new boot images, which requires manual effort. Logging the boot image version directly provides a cleaner and more reliable signal.

----

> [!IMPORTANT]
> Add boot image version to runner logs in `GithubRunner` to track success rates per boot image.
> 
>   - **Behavior**:
>     - Add `boot_image` version to runner logs in `log_duration` method of `GithubRunner` class.
>     - Handles cases where `vm_storage_volumes` is present or not.
>   - **Tests**:
>     - Update `github_runner_spec.rb` to test logging of `boot_image` version in `log_duration`.
>     - Add test cases for VMs with and without `vm_host`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 32299d3a1234e6074d172e5a50fd92c72998a437. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/github_runner.rb
+++ b/model/github_runner.rb
@@ -46,7 +46,7 @@ class GithubRunner < Sequel::Model
   def log_duration(message, duration)
     values = {ubid:, label:, repository_name:, duration: duration.round(3), conclusion: workflow_job&.dig("conclusion")}
     if vm
-      values.merge!(vm_ubid: vm.ubid, arch: vm.arch, cores: vm.cores, vcpus: vm.vcpus)
+      values.merge!(vm_ubid: vm.ubid, arch: vm.arch, cores: vm.cores, vcpus: vm.vcpus, boot_image: vm.vm_storage_volumes.first&.boot_image&.version || vm.boot_image)
       if vm.vm_host
         values[:vm_host_ubid] = vm.vm_host.ubid
         values[:data_center] = vm.vm_host.data_center
--- a/spec/model/github_runner_spec.rb
+++ b/spec/model/github_runner_spec.rb
@@ -23,6 +23,8 @@ RSpec.describe GithubRunner do
 
   it "can log duration with a vm" do
     vm = github_runner.vm
+    boot_image = BootImage.create(vm_host_id: vm.vm_host_id, name: "github-ubuntu-2204", version: "20251211", size_gib: 75)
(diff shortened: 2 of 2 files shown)


model/github_runner.rb
```diff
@@ -46,7 +46,7 @@ def runner_url
   def log_duration(message, duration)
     values = {ubid:, label:, repository_name:, duration: duration.round(3), conclusion: workflow_job&.dig("conclusion")}
     if vm
-      values.merge!(vm_ubid: vm.ubid, arch: vm.arch, cores: vm.cores, vcpus: vm.vcpus)
+      values.merge!(vm_ubid: vm.ubid, arch: vm.arch, cores: vm.cores, vcpus: vm.vcpus, boot_image: vm.vm_storage_volumes.first&.boot_image&.version || vm.boot_image)
```
jeremyevans, line 49: I determined this is likely correct, even though it looks wrong. That's because `boot_image` is an association for `VmStorageVolume`, but a text column for `Vm`. I assumed there is a reason for this discrepancy, maybe with the  `BootImage` model coming after the `Vm#boot_image` column, as opposed to a deliberate design decision. That turns out to be the case, with the `BootImage` model coming in 33f87d09ddf1d70c23d3e15b54b361250c8af0e2 in April 2024, while the `vm.boot_image` column coming in  adc109d2ea570a170e71e339304c502b7d03c798 in March 2023.

enescakir: Yes, you're right. In the early days, it was `vm.boot_image` string. Later, when we wanted to support multiple versions of the same image, we introduced the `BootImage` model.
We could refactor `vm.boot_image` to return `vm.vm_storage_volumes.find(&:boot).boot_image`, but that would require more work.

PR #4399, opened 2025-12-16
Page on PG promotion failure

shikharbhardwaj:
> [!IMPORTANT]
> Handle PostgreSQL promotion failures by creating a page and retrying in `Prog::Postgres::PostgresServerNexus`.
> 
>   - **Behavior**:
>     - In `Prog::Postgres::PostgresServerNexus`, `taking_over` method now creates a page on promotion failure and retries the promotion.
>     - On successful promotion, resolves any existing failure page.
>   - **Tests**:
>     - Added tests in `postgres_server_nexus_spec.rb` to verify page creation on promotion failure and resolution on success.
>     - Tests cover scenarios for both initial promotion attempts and retries.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a5eb95449738bf3759ac1fa2b3fcd807f58fdb1c. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -599,6 +599,7 @@ SQL
 
   label def prepare_for_unplanned_take_over
     decr_unplanned_take_over
+    register_deadline("wait", 10 * 60)
 
     representative_server = postgres_server.resource.representative_server
 
@@ -614,6 +615,7 @@ SQL
 
   label def prepare_for_planned_take_over
     decr_planned_take_over
+    register_deadline("wait", 10 * 60)
 
     postgres_server.resource.representative_server.incr_fence
     hop_wait_fencing_of_old_primary
--- a/spec/prog/postgres/postgres_server_nexus_spec.rb
+++ b/spec/prog/postgres/postgres_server_nexus_spec.rb
@@ -1099,6 +1099,7 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
     it "stops postgres in representative server and destroys it" do
       standby_nx.incr_unplanned_take_over
+      expect(@standby_nx).to receive(:register_deadline).with("wait", 10 * 60)
       expect(representative_sshable).to receive(:_cmd).with("sudo pg_ctlcluster 16 main stop -m immediate")
       expect { standby_nx.prepare_for_unplanned_take_over }.to hop("taking_over")
       expect(Semaphore.where(strand_id: standby_nx.postgres_server.id, name: "unplanned_take_over").count).to eq(0)
@@ -1107,6 +1108,7 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
     it "handles SSH connection errors gracefully and continues with destroy" do
       standby_nx.incr_unplanned_take_over
+      expect(@standby_nx).to receive(:register_deadline).with("wait", 10 * 60)
       expect(representative_sshable).to receive(:_cmd).with("sudo pg_ctlcluster 16 main stop -m immediate").and_raise(Sshable::SshError.new("", "", "", "", ""))
       expect { standby_nx.prepare_for_unplanned_take_over }.to hop("taking_over")
       expect(Semaphore.where(strand_id: standby_nx.postgres_server.id, name: "unplanned_take_over").count).to eq(0)
@@ -1121,6 +1123,7 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
 
     it "starts fencing on representative server" do
       @standby_nx.incr_planned_take_over
+      expect(@standby_nx).to receive(:register_deadline).with("wait", 10 * 60)
       expect { @standby_nx.prepare_for_planned_take_over }.to hop("wait_fencing_of_old_primary")
       expect(Semaphore.where(strand_id: @standby_nx.postgres_server.id, name: "planned_take_over").count).to eq(0)
       expect(Semaphore.where(strand_id: postgres_server.id, name: "fence").count).to eq(1)


spec/model/postgres/postgres_resource_spec.rb
```diff
@@ -328,4 +328,16 @@
       expect(postgres_resource.ready_for_read_replica?).to be false
     end
   end
+
+  describe "#install_rhizome" do
+    it "installs rhizome on all servers" do
+      server1 = instance_double(PostgresServer, vm: instance_double(Vm, id: "1"))
```
serprex: should avoid `instance_double` over real db objects

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -955,15 +955,50 @@
...
-      expect(sshable).to receive(:d_check).with("promote_postgres").and_return("NotStarted", "Failed")
+      expect(sshable).to receive(:d_check).with("promote_postgres").and_return("NotStarted")
       expect { nx.taking_over }.to nap(0)
+    end
+
+    it "triggers a page and retries if promote command is failed" do
+      expect(Prog::PageNexus).to receive(:assemble).with(
+        "pgubid promotion failed", ["PGPromotionFailed", postgres_server.id],
+        "pgubid"
+      ).and_return(instance_double(Prog::PageNexus, bud: nil))
```
jeremyevans: Instead of mocking the method, check for the expected page at the end of the spec. For new specs, only mock SSH commands or things that result in SSH attempts (like `d_run`), or HTTP access. In general, you shouldn't mock anything that only accesses the database.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -955,15 +955,50 @@
...
+      expect(sshable).to receive(:d_run).with("promote_postgres", "sudo", "postgres/bin/promote", "16")
+
+      expect(sshable).to receive(:d_check).with("promote_postgres").and_return("Failed")
       expect { nx.taking_over }.to nap(0)
     end
 
     it "updates the metadata and hops to configure if promote command is succeeded" do
+      expect(Page).to receive(:from_tag_parts).with("PGPromotionFailed", postgres_server.id).and_return(nil)
```
jeremyevans: Not needed as there shouldn't be an existing page with those tag parts.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -955,15 +955,50 @@
...
+      expect(postgres_server).to receive(:incr_restart)
+
+      standby = instance_double(PostgresServer, primary?: false)
+      expect(standby).to receive(:update).with(synchronization_status: "catching_up")
+      expect(standby).to receive(:incr_configure)
+      expect(standby).to receive(:incr_configure_metrics)
+      expect(standby).to receive(:incr_restart)
+
+      expect(postgres_server.resource).to receive(:servers).at_least(:once).and_return([postgres_server, standby])
+
```
jeremyevans: All of this mocking code should be replaced with database modifications before calling `nx.taking_over`, and checking that the expected modifications were made after calling `nx.taking_over`.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -955,15 +955,50 @@
...
+      expect { nx.taking_over }.to hop("configure")
+    end
+
+    it "resolves existing page, updates the metadata and hops to configure if promote command is succeeded" do
+      page = instance_double(Page)
+      expect(Page).to receive(:from_tag_parts).with("PGPromotionFailed", postgres_server.id).and_return(page)
+      expect(page).to receive(:incr_resolve)
```
jeremyevans: Add a page to the database here, and at the end of the spec, check that the expected resolve semaphore is set.

spec/model/postgres/postgres_resource_spec.rb
```diff
@@ -328,4 +328,16 @@
...
+      server2 = instance_double(PostgresServer, vm: instance_double(Vm, id: "2"))
+      expect(postgres_resource).to receive(:servers).and_return([server1, server2])
+      [server1, server2].each do |server|
+        expect(Strand).to receive(:create).with(prog: "InstallRhizome", label: "start", stack: [{subject_id: server.vm.id, target_folder: "postgres", install_specs: false}])
+      end
```
jeremyevans: Remove all mocking code. Create the PostgresServer and Vm instances before calling `install_rhizome`, and check that the expected strands are created after calling it.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -630,14 +630,20 @@ def before_run
 
     case vm.sshable.d_check("promote_postgres")
     when "Succeeded"
+      Page.from_tag_parts("PGPromotionFailed", postgres_server.id)&.incr_resolve
```
byucesoy, line 641: We should also set up a deadline based page (i.e we set the deadline before starting failover for us to reach back to wait state in timely manner)

This would raise the page quicker, but it would miss problems in other labels. I think we should have both.

spec/prog/postgres/postgres_server_nexus_spec.rb
shikharbhardwaj: IIUC, this needs the entire file to be refactored to setup the database objects instead of using the mocks for `postgres_server` and `resource` etc? Or am I misunderstanding this?

fdr:
I can probably get this keyboard smashed into being democked given my (barely held together) prompt storage methodology, plz hold the phone....

fdr:
Okay, here's the first attempt. We may want to think about how to make the history look (what do you want to see in `blame`?) but it should reduce the despair level in having to keyboard mash existing tests

fdr:
I'm kind of supersizing this to get things out of the way, but I think we can drag it through -- I see the last patch needs a bit of work in a few areas, not least to clean up the commit message, and some of the comments. I am more blase about careful manicuring of PRs and much more careful manicuring  commits.

fdr:
@byucesoy @jeremyevans okay, check out my in-passing demock and/or I think there was one feature suggestion in https://github.com/ubicloud/ubicloud/pull/4399#discussion_r2625669171, didn't want that to get lost...

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -3,58 +3,24 @@
...
-    )
+  let(:postgres_resource) {
+    Prog::Postgres::PostgresResourceNexus.assemble(
+      project_id: user_project.id, location_id: Location::HETZNER_FSN1_ID,
+      name: "test-pg", target_vm_size: "standard-2", target_storage_size_gib: 64
+    ).subject
   }
 
-  let(:sshable) { Sshable.new }
+  let(:server_strand) { postgres_resource.servers.first.strand }
```
jeremyevans: Unless it is different:

```suggestion
  let(:server_strand) { postgres_server.strand }
```

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
     end
 
     it "picks correct base image for Lantern" do
-      expect(PostgresResource).to receive(:[]).and_return(postgres_resource)
-      expect(postgres_resource).to receive(:flavor).and_return(PostgresResource::Flavor::LANTERN).at_least(:once)
-      expect(Prog::Vm::Nexus).to receive(:assemble_with_sshable).with(anything, hash_including(boot_image: "postgres16-lantern-ubuntu-2204")).and_return(instance_double(Strand, id: "62c62ddb-5b5a-4e9e-b534-e73c16f86bcb", subject: instance_double(Vm, add_vm_firewall: nil)))
-      expect(PostgresServer).to receive(:create_with_id).and_return(instance_double(PostgresServer, id: "5c13fd6a-25c2-4fa4-be48-2846f127526a"))
-      described_class.assemble(resource_id: postgres_resource.id, timeline_id: "91588cda-7122-4d6a-b01c-f33c30cb17d8", timeline_access: "push", representative_at: Time.now)
+      postgres_project = Project.create(name: "default")
```
jeremyevans: Handled by the `let`:

```suggestion
```

Same issue in other specs.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
+      expect(Config).to receive(:postgres_service_project_id).and_return(postgres_project.id).at_least(:once)
```
jeremyevans: Handled by the `allow` in `before`:

```suggestion
```

Same issue in other specs.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
-        project_id: user_project.id
+      # Use an existing seed location (us-west-2) that has billing rates
+      aws_location = Location.find(name: "us-west-2")
+      skip "us-west-2 location not available in test seeds" unless aws_location
+
+      postgres_project = Project.create(name: "default")
+      expect(Config).to receive(:postgres_service_project_id).and_return(postgres_project.id).at_least(:once)
+
+      # Ensure AMI mapping exists
+      PgAwsAmi.find_or_create(aws_location_name: "us-west-2", pg_version: "16", arch: "x64") { |a| a.aws_ami_id = "ami-pg16-x64" }
```
jeremyevans: It's odd to use `find_or_create` inside a spec. You should know at this point whether or not the `PgAwsAmi` should exist, and use either `first` or `create`. Same issue in other specs.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
-      expect(postgres_resource).to receive(:version).and_return("16").at_least(:once)
-      expect(Prog::Vm::Nexus).to receive(:assemble_with_sshable).and_return(instance_double(Strand, id: "62c62ddb-5b5a-4e9e-b534-e73c16f86bcb", subject: instance_double(Vm, add_vm_firewall: nil)))
-      expect(PostgresServer).to receive(:create_with_id).and_return(instance_double(PostgresServer, id: "5c13fd6a-25c2-4fa4-be48-2846f127526a"))
-      described_class.assemble(resource_id: postgres_resource.id, timeline_id: "91588cda-7122-4d6a-b01c-f33c30cb17d8", timeline_access: "push", representative_at: Time.now)
+      # Create firewall with correct name
+      Firewall.create(name: "#{aws_resource.ubid}-internal-firewall", location_id: aws_location.id, project_id: postgres_project.id)
+      postgres_timeline = PostgresTimeline.create
+
+      st = described_class.assemble(resource_id: aws_resource.id, timeline_id: postgres_timeline.id, timeline_access: "push", representative_at: Time.now)
+      postgres_server = PostgresServer[st.id]
```
jeremyevans:
```suggestion
      postgres_server = st.subject
```

Same issue in other specs.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
-      loc = Location.create(
-        name: "us-west-2",
-        display_name: "aws-us-west-2",
-        ui_name: "aws-us-west-2",
-        visible: true,
-        provider: "aws",
-        project_id: user_project.id
+      # Use an existing seed location (us-west-2) that has billing rates
+      aws_location = Location.find(name: "us-west-2")
+      skip "us-west-2 location not available in test seeds" unless aws_location
```
jeremyevans: This line should be removed, the `us-west-2` location was added by a migration. Same issue in other specs.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
+        superuser_password: "dummy-password",
+        target_version: "17"
       )
-      expect(postgres_resource).to receive(:version).and_return("17").at_least(:once)
-      expect(postgres_resource).to receive(:location).and_return(loc).at_least(:once)
-      expect(postgres_resource).to receive(:location_id).and_return(loc.id).at_least(:once)
-      expect(Prog::Vm::Nexus).to receive(:assemble_with_sshable).and_return(instance_double(Strand, id: "62c62ddb-5b5a-4e9e-b534-e73c16f86bcb", subject: instance_double(Vm, add_vm_firewall: nil)))
-      expect(PostgresServer).to receive(:create_with_id).and_return(instance_double(PostgresServer, id: "5c13fd6a-25c2-4fa4-be48-2846f127526a"))
-      described_class.assemble(resource_id: postgres_resource.id, timeline_id: "91588cda-7122-4d6a-b01c-f33c30cb17d8", timeline_access: "push", representative_at: Time.now)
+      # Create firewall with correct name
```
jeremyevans: Not sure this type of comment is helpful. Same issue for some other comments added.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
 
-    it "errors out for unknown flavor" do
-      expect(PostgresResource).to receive(:[]).and_return(postgres_resource)
-      expect(postgres_resource).to receive(:flavor).and_return("boring_flavor").at_least(:once)
+      # Error happens when pg_ami returns nil for unsupported version
       expect {
-        described_class.assemble(resource_id: postgres_resource.id, timeline_id: "91588cda-7122-4d6a-b01c-f33c30cb17d8", timeline_access: "push", representative_at: Time.now)
-      }.to raise_error RuntimeError, "Unknown PostgreSQL flavor: boring_flavor"
+        described_class.assemble(resource_id: aws_resource.id, timeline_id: postgres_timeline.id, timeline_access: "push", representative_at: Time.now)
+      }.to raise_error NoMethodError, "undefined method 'aws_ami_id' for nil"
```
jeremyevans: Expecting code to raise a `NoMethodError` is odd. The prog should not be calling methods on an instance that may be nil. We should modify the prog not to do that. The `RuntimeError` raised previously was superior.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
+      expect(Semaphore.where(strand_id: postgres_server.id, name: "destroy").count).to eq(0)
     end
 
     it "pops additional operations from stack" do
-      expect(nx).to receive(:when_destroy_set?).and_yield
-      expect(resource).to receive(:strand).and_return(instance_double(Strand, label: "destroy"))
-      expect(nx.strand).to receive(:label).and_return("destroy").at_least(:once)
-      expect(nx.strand.stack).to receive(:count).and_return(2)
-      expect { nx.before_run }.to exit({"msg" => "operation is cancelled due to the destruction of the postgres server"})
+      nx.incr_destroy
```
jeremyevans: Moving this code later will make things simpler:

```suggestion

```

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
+      server_strand.update(
+        label: "destroy",
+        stack: [
+          {"subject_id" => postgres_server.id, "link" => ["Postgres::PostgresServerNexus", "wait"]},
+          {}
+        ]
+      )
+      # Create fresh nexus so frame memoization picks up the new stack
+      fresh_nx = described_class.new(server_strand.reload)
+      expect { fresh_nx.before_run }.to hop("wait")
```
jeremyevans: You don't need to create a separate nexus object if you want to reference it until after modifying `server_strand`:

```suggestion
      nx.incr_destroy
      expect { nx.before_run }.to hop("wait")
```

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -100,142 +66,194 @@
...
 
   describe "#bootstrap_rhizome" do
-    it "buds a bootstrap rhizome process" do
-      expect(postgres_server).to receive(:primary?).and_return(true)
-      expect(nx).to receive(:bud).with(Prog::BootstrapRhizome, {"target_folder" => "postgres", "subject_id" => postgres_server.vm.id, "user" => "ubi"})
+    it "buds a bootstrap rhizome process for primary" do
+      # Server created by assemble is primary (has representative_at)
       expect { nx.bootstrap_rhizome }.to hop("wait_bootstrap_rhizome")
+      child = Strand.where(prog: "BootstrapRhizome", parent_id: server_strand.id).first
+      expect(child).not_to be_nil
```
jeremyevans: In this particular case, it may be OK, but `not_to be_nil` is generally a bad pattern. In most cases, and this one, you know what it should be:

```suggestion
      expect(child).to be_a Strand
```

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -246,75 +264,91 @@
...
-      expect(postgres_server).to receive(:storage_device_paths).and_return(["/dev/nvme1n1", "/dev/nvme2n1"])
-      expect(sshable).to receive(:_cmd).with("sudo mdadm --create --verbose /dev/md0 --level=0 --raid-devices=2 /dev/nvme1n1 /dev/nvme2n1")
-      expect(sshable).to receive(:d_run).with("format_disk", "sudo", "mkfs", "--type", "ext4", "/dev/md0")
+      # RAID setup only happens on AWS where storage_device_paths returns multiple paths.
+      # storage_device_paths does SSH to lsblk on AWS - stubbing to avoid AWS infra setup.
+      paths = ["/dev/nvme1n1", "/dev/nvme2n1"]
+      allow(postgres_server).to receive(:storage_device_paths).and_return(paths)
 
       expect(sshable).to receive(:d_check).with("format_disk").and_return("NotStarted")
+      expect(sshable).to receive(:cmd).with("sudo mdadm --create --verbose /dev/md0 --level=0 --raid-devices=:count :shelljoin_storage_device_paths",
```
jeremyevans:
```suggestion
      expect(sshable).to receive(:_cmd).with("sudo mdadm --create --verbose /dev/md0 --level=0 --raid-devices=:count :shelljoin_storage_device_paths",
```

Same for any other cases where `cmd` is currently mocked.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -363,22 +400,47 @@
...
+      version = standby.version
+      expect(standby_sshable).to receive(:d_check).with("initialize_database_from_backup").and_return("NotStarted")
+      expect(standby_sshable).to receive(:d_run).with("initialize_database_from_backup", "sudo", "postgres/bin/initialize-database-from-backup", version, "LATEST")
+      expect { standby_nx.initialize_database_from_backup }.to nap(5)
     end
   end
 
   describe "#refresh_certificates" do
+    before do
+      # Set up real certificate data on the resource
```
jeremyevans: I don't believe any certs or keys in the update call are actually valid.

```suggestion
      # Set up fake certificate data on the resource
```

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -431,29 +497,41 @@
...
-      expect(nx).to receive(:when_initial_provisioning_set?).and_yield
-      expect(resource).to receive(:use_old_walg_command_set?).and_return(false)
-      expect(sshable).to receive(:_cmd).with("sudo -u prometheus tee /home/prometheus/web-config.yml > /dev/null", stdin: anything)
-      expect(sshable).to receive(:_cmd).with("sudo -u prometheus tee /home/prometheus/prometheus.yml > /dev/null", stdin: anything)
-      expect(sshable).to receive(:_cmd).with("sudo systemctl enable --now postgres_exporter")
-      expect(sshable).to receive(:_cmd).with("sudo systemctl enable --now node_exporter")
-      expect(sshable).to receive(:_cmd).with("sudo systemctl enable --now prometheus")
-      expect(sshable).to receive(:_cmd).with("sudo systemctl enable --now wal-g")
+      # Set initial_provisioning semaphore (no use_old_walg_command = new style)
+      nx.incr_initial_provisioning
```
jeremyevans: I would move this line after the PostgresTimeline update. Then you shouldn't need the `nx.instance_variable_set(:@postgres_server, nil)`

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -431,29 +497,41 @@
...
+
+      # Update timeline to use an AWS location BEFORE setting up expectations
+      aws_location = Location.where(provider: "aws").first
+      PostgresTimeline.where(id: postgres_server.timeline_id).update(location_id: aws_location.id)
+      # Clear the memoized postgres_server from the nexus to force a fresh load
+      nx.instance_variable_set(:@postgres_server, nil)
+
+      # Now set up expectations on the fresh sshable
+      fresh_sshable = nx.postgres_server.vm.sshable
```
jeremyevans: Better to rename this local variable to `sshable`, as you aren't mixing the use of fresh and stale sshables.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -803,9 +912,11 @@
 
         expect(nx.strand).to receive(:stack).and_return([{"lsn" => "1/A"}]).at_least(:once)
         expect(postgres_server).to receive(:lsn_diff).with("1/A", "1/A").and_return(0)
-        expect(postgres_server).to receive(:recycle_set?).and_return(false)
-        expect(postgres_server).to receive(:incr_recycle)
+        # No recycle semaphore set initially
+        expect(Semaphore.where(strand_id: server_strand.id, name: "recycle").count).to eq(0)
         expect { nx.wait }.to nap(60)
+        # Verify recycle semaphore was incremented
+        expect(Semaphore.where(strand_id: server_strand.id, name: "recycle").count).to eq(1)
```
jeremyevans: Up to you whether to switch this to use `.and change`/`by(1)`. Similar issues elseswhere

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -839,54 +953,54 @@
...
 
   describe "#fence" do
     it "runs checkpoints and perform lockout" do
-      expect(nx).to receive(:decr_fence)
+      nx.incr_fence
       expect(postgres_server).to receive(:run_query).with("CHECKPOINT; CHECKPOINT; CHECKPOINT;")
-      expect(sshable).to receive(:_cmd).with("sudo postgres/bin/lockout 16")
-      expect(sshable).to receive(:_cmd).with("sudo pg_ctlcluster 16 main stop -m smart")
+      expect(sshable).to receive(:_cmd).with("sudo postgres/bin/lockout #{postgres_server.version}")
+      expect(sshable).to receive(:_cmd).with("sudo pg_ctlcluster #{postgres_server.version} main stop -m smart")
```
jeremyevans: I'm not in favor of unnecessary interpolation in expected command strings. we should hard code the expected value if it is known statically (even if we will have to update it when dropping PG 16 support).

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -897,91 +1011,194 @@
...
   describe "#prepare_for_unplanned_take_over" do
+    # Create a cluster with a real primary and standby for failover tests
+    def create_failover_cluster
+      standby_strand = described_class.assemble(
+        resource_id: postgres_resource.id, timeline_id: postgres_resource.timeline.id,
+        timeline_access: "fetch"
+      )
+      standby = PostgresServer[standby_strand.id]
+      primary = postgres_resource.servers.find { it.primary? }
+      [primary, standby, standby_strand]
```
jeremyevans: Probably better to just return

```suggestion
      [primary, standby]
```

and adjust callers to use `standby.strand`.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -897,91 +1011,194 @@
...
+      expect(page.summary).to eq("#{postgres_server.ubid} promotion failed")
     end
 
-    it "updates the metadata and hops to configure if promote command is succeeded" do
-      expect(sshable).to receive(:d_check).with("promote_postgres").and_return("Succeeded")
+    context "when promote succeeds with real models" do
+      def create_postgres_cluster
+        user_project = Project.create(name: "test-project")
+        postgres_project = Project.create(name: "postgres-service")
+        allow(Config).to receive(:postgres_service_project_id).and_return(postgres_project.id)
```
jeremyevans: Duplicates somewhat the `let` and `before` hooks near the top of the file, not sure whether these lines are needed.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -1030,16 +1249,22 @@
...
+      vm_strand = postgres_server.vm.strand
+      pg_server_id = postgres_server.id
+      expect(Semaphore.where(strand_id: vm_strand.id, name: "destroy").count).to eq(0)
 
       expect { nx.destroy_vm_and_pg }.to exit({"msg" => "postgres server is deleted"})
+
+      # Verify vm.incr_destroy was called
+      expect(Semaphore.where(strand_id: vm_strand.id, name: "destroy").count).to eq(1)
+      # Verify postgres_server was destroyed
+      expect(PostgresServer[pg_server_id]).to be_nil
```
jeremyevans: I think this is the simpler rspec syntax:

```suggestion
      expect(postgres_server).not_to exist
```

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -1052,8 +1277,10 @@
     end
 
     it "returns true if the resource is upgrading" do
-      expect(postgres_server.resource).to receive(:target_version).and_return("17")
-      expect(postgres_server.resource).to receive(:upgrade_candidate_server).and_return(postgres_server)
+      # Set target_version different from current version to simulate an upgrade in progress
+      new_version = (postgres_server.version == "17") ? "18" : "17"
```
jeremyevans: We should know at this point what the version is, and hard code the version. If we don't want to do that:


```suggestion
      new_version = postgres_server.version.succ
```

jeremyevans (commented):
The democking is a large improvement, but there are many issues that should be fixed in it.

shikharbhardwaj:
I'll rebase this with the latest democking changes

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: I added a couple of 10-minute deadlines for both types of failover.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -599,6 +599,7 @@ def before_run
 
   label def prepare_for_unplanned_take_over
     decr_unplanned_take_over
+    register_deadline("wait", 10 * 60)
```
jeremyevans, line 602: Should probably have specs added that this deadline is set. Same for deadline added in `prepare_for_planned_take_over`.

PR #4400, opened 2025-12-16
Don't generate keys for timeline when using IAM access

serprex:
These random keys were interfering with wal-g config generation, which expects nil values to skip encoding credentials

----

> [!IMPORTANT]
> Modify `assemble` in `PostgresTimelineNexus` to skip key generation when `aws_postgres_iam_access` is true.
> 
>   - **Behavior**:
>     - In `assemble` method of `PostgresTimelineNexus`, set `access_key` and `secret_key` to `nil` if `Config.aws_postgres_iam_access` is true, otherwise generate random keys.
>   - **Reason**:
>     - Prevents interference with `wal-g` config generation, which expects nil values to skip encoding credentials.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 7115406c32c6e037c1d465dbbd589bfc811818e6. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/prog/postgres/postgres_timeline_nexus.rb
+++ b/prog/postgres/postgres_timeline_nexus.rb
@@ -22,8 +22,8 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
     DB.transaction do
       postgres_timeline = PostgresTimeline.create(
         parent_id: parent_id,
-        access_key: SecureRandom.hex(16),
-        secret_key: SecureRandom.hex(32),
+        access_key: (location.aws? && Config.aws_postgres_iam_access) ? nil : SecureRandom.hex(16),
+        secret_key: (location.aws? && Config.aws_postgres_iam_access) ? nil : SecureRandom.hex(32),
         location_id: location.id
       )
       Strand.create_with_id(postgres_timeline, prog: "Postgres::PostgresTimelineNexus", label: "start")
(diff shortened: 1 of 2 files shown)


prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -22,8 +22,8 @@ def self.assemble(location_id:, parent_id: nil)
     DB.transaction do
       postgres_timeline = PostgresTimeline.create(
         parent_id: parent_id,
-        access_key: SecureRandom.hex(16),
-        secret_key: SecureRandom.hex(32),
+        access_key: Config.aws_postgres_iam_access ? nil : SecureRandom.hex(16),
```
furkansahin: This assumes `aws_postgres_iam_access` flag is only available in an AWS only deployment. It would be better if we checked `(location.aws? && Config.aws_postgres_iam_access)`

serprex: fixed & added test for both cases

enescakir:
Our commit subject/body guideline: https://github.com/ubicloud/ubicloud/blob/main/COMMIT_MESSAGES.md

No need to add prefix to the subject

PR #4404, opened 2025-12-17
[DO NOT REVIEW] Add privatelink support to Ubi Postgres on AWS

furkansahin:


fdr:
oh man. well, it's true, people do ask for this stuff sometimes. the crux is always: what about the port numbering and TLS certificates? It's too unwieldy to have one private link per database.

furkansahin:
I like the implementation in https://github.com/ubicloud/ubicloud/pull/4420 better because that enables us to make this flexible for any ubicloud customer in future. They could essentially have ubicloud VMs serving different services which is connected to their AWS environment via private link. What do you think @fdr ? I will also simply assume for the timebeing that `sslmode=verify-ca` is good enough. The implementation in that new PR should also be flexible enough so that they define different port numbers on the endpoint which maps to 5432 of a different db in future.

fdr:
verify-ca will raise questions, so we'll have to be ready to explain them with a convincing explanation for All Time. It's only adequate with extremely fine-grained CA infrastructure, i.e. where it degenerates to something similar to verify-full. What's the reason?

PR #4417, opened 2025-12-18
Run GitHub Actions tests on Alpine Linux

fdr:
We had at least one platform portability issue when using our docker images, which are built against alpine, and thus busybox and musl libc.  That was the shebang in
4f6396d82c02f6176639c96f26196910dbc6ecd1. While not caught by the tests we have right now: it could be, and the first step to automating catching those regressions is to expand the build matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,7 @@ jobs:
         cat > Dockerfile.alpine-test <<'EOF'
         FROM ruby:3.4.7-alpine3.21
         RUN apk add --no-cache \
+          bash \
           build-base \
           ruby-dev \
           linux-headers \


fdr:
hmm I think I have a trick to get around this

.github/workflows/ci.yml
```diff
@@ -76,6 +76,23 @@ jobs:
...
+          HOST_GID=$(id -g runner)
+
+          docker run --rm                \
+            -u "${HOST_UID}:${HOST_GID}" \
+            --network=host               \
+            -v "$PWD:/app"               \
+            -w /app                      \
+            ruby:3.4.7-alpine3.21        \
+            sh -c '
+              apk add --no-cache build-base ruby-dev linux-headers postgresql-dev
```
fdr:
```suggestion
              sudo apk add --no-cache build-base ruby-dev linux-headers postgresql-dev
```

fdr:
This is probably a good idea but was a bit harder than I had hoped to make it stick.  I think for the sake of cleanliness I'm going to close this.

PR #4419, opened 2025-12-19
Add VM host health check to verify clock source is tsc

enescakir:
tsc is the preferred clock source because it is more reliable and precise. Our hosts use tsc by default, but in some cases they may fall back to hpet if issues are detected with tsc.

When a host switches to hpet, it can cause instability and runtime issues for runners. In particular, this leads to errors like "OS monotonic time stepped backwards!", which we've observed frequently in Elixir-based workflow jobs.

We should monitor the clock source on VM hosts and alert if any host switches from tsc to hpet, so we can take corrective action early.

----

> [!IMPORTANT]
> Add clock source health check to `VmHost` to ensure it is set to `tsc` or `arch_sys_counter` for `arm64`, with logging and tests.
> 
>   - **Behavior**:
>     - Adds `check_clock_source` method in `VmHost` to verify clock source is `tsc` or `arch_sys_counter` for `arm64`.
>     - Integrates `check_clock_source` into `perform_health_checks` in `VmHost`.
>     - Logs error if clock source is not as expected.
>   - **Tests**:
>     - Adds tests for `check_clock_source` in `vm_host_spec.rb` to verify behavior for different clock sources and architectures.
>     - Updates existing tests to include clock source check in `check_pulse` method.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for a7d79fac8e5c9ccc634299302ff88249d3d5aca9. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -348,6 +348,18 @@ class VmHost < Sequel::Model
     error_count.empty?
   end
 
+  def check_clock_source(ssh_session)
+    clock_source = ssh_session.exec!("cat /sys/devices/system/clocksource/clocksource0/available_clocksource").strip
+    clock_status = if arch == "arm64"
+      clock_source == "arch_sys_counter"
+    else
+      clock_source.start_with?("tsc")
+    end
+
+    Clog.emit("unexpected clock source", {unexpected_clock_source: {vm_host_ubid: ubid, clock_source:}}) unless clock_status
+    clock_status
+  end
+
   def check_last_boot_id(ssh_session)
     boot_id = ssh_session.exec!("cat /proc/sys/kernel/random/boot_id")
     fail "Failed to exec on session: #{boot_id}" unless boot_id.exitstatus.zero?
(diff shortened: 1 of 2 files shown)


model/vm_host.rb
```diff
@@ -350,6 +350,19 @@ def check_storage_kernel_logs(ssh_session, devices)
     error_count.empty?
   end
 
+  def check_clock_source(ssh_session)
+    clock_source = ssh_session.exec!("cat /sys/devices/system/clocksource/clocksource0/available_clocksource")
+    return false unless clock_source.exitstatus == 0
```
jeremyevans: Is this an expected condition? If not, shouldn't we treat it similarly to unexpected clock source (including the emit)?

jeremyevans (approved):

fdr (approved):
I guess we should figure out how often this happens. It's not good, but also, setting `tsc=reliable` or putting machines back in tsc without thinking about it at all is also bad.

PR #4420, opened 2025-12-19
[DO NOT REVIEW] Subnet based Privatelink implementation

furkansahin:


furkansahin:
One thing we need to figure out is the subnet association. looks like the elb needs to be assigned subnets from each AZ so that it is accessible from any subnet in customer VPC. Otherwise, they have to make sure the client subnet is in the same AZ as the elb associated subnet.
My plan:
Generate 1 AWS subnet per AZ when we initiate the subnet nexus and reuse these for Nics. That way, we can associate all of these subnets to the ELB

PR #4425, opened 2025-12-20
Implement model hiding in Clover

jeremyevans:
With model hiding, during coverage testing, access to most models is
proxied, and only specific methods are allowed. This is designed to
prevent future insecure object access vulnerabilities. In general,
object access should be based on associations from existing objects, so
that a logged in account only deals with related projects, and once a
project is selected, the displayed objects (e.g. VMs) must be related to
the project, etc. In general, that's how Clover was already setup, but
with this change, tests fail if an insecure pattern is used.

This whitelists the allowed model calls. Only Account and OidcProvider
are allowed direct (full) access, since those are models used before any
authentication is done.

Because setting this up requires access to all models, it is only done
when autoloads are forced (which coverage testing does). It is not done
in frozen testing mode, or in development or production.

In order to get this to work, there are a number of minor changes needed
to avoid the existing direct model accesses. So all commits leading up
to the final one set the ground work for this change.

----

> [!IMPORTANT]
> Implement model hiding in Clover to restrict direct model access during coverage testing, enhancing security by allowing only specific methods.
> 
>   - **Behavior**:
>     - Introduces model hiding in `clover.rb` to restrict direct model access during coverage testing, allowing only specific methods.
>     - Whitelists allowed model calls in `ModelProxy` class, raising `DirectModelAccess` error for unauthorized access.
>     - Only `Account` and `OidcProvider` models have full access due to pre-authentication usage.
>     - Model hiding activated only when autoloads are forced, not in frozen testing, development, or production.
>   - **Refactoring**:
>     - Replaces direct model accesses with authorized methods in `clover.rb`, `helpers/general.rb`, `helpers/inference.rb`, `helpers/load_balancer.rb`, `helpers/postgres.rb`, `helpers/web.rb`, `routes/cli.rb`, `routes/github.rb`, `routes/project/billing.rb`, `routes/project/location/load_balancer.rb`, `routes/project/location/postgres.rb`, `routes/project/location/private_subnet.rb`, `routes/project/token.rb`, `routes/project/user.rb`, `routes/runtime/github.rb`, `routes/webhook/github.rb`, `spec/routes/runtime/github_spec.rb`, `spec/routes/web/github_spec.rb`, `spec/routes/web/project_spec.rb`, `spec/routes/web/webhook/github_spec.rb`, `spec/thawed_mock.rb`, `views/networking/private_subnet/networking.erb`, `views/postgres/create.erb`, `views/postgres/settings.erb`, `views/vm/networking.erb`.
>     - Adds helper methods in models like `Account`, `ApiKey`, `GithubInstallation`, `InferenceEndpoint`, `LoadBalancer`, `Location`, `PaymentMethod`, `PostgresResource`, `PrivateSubnet`, `Project`, `SubjectTag`, `Vm` to support new access patterns.
>   - **Testing**:
>     - Updates tests in `spec/routes/runtime/github_spec.rb`, `spec/routes/web/github_spec.rb`, `spec/routes/web/project_spec.rb`, `spec/routes/web/webhook/github_spec.rb` to align with model hiding changes.
>     - Mocks new methods in `spec/thawed_mock.rb` to facilitate testing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for 90b3b5cd432c09860383991a2ef1c509ddf54258. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/helpers/model_hiding.rb
+++ b/helpers/model_hiding.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+class Clover < Roda
+  if Config.unfrozen_test? && ENV["FORCE_AUTOLOAD"] == "1"
+    class DirectModelAccess < StandardError; end
+
+    module ModelProxyAccess
+      def is_a?(klass)
+        klass = klass.sequel_model if ModelProxy === klass
+        super
+      end
+    end
+    Object.include ModelProxyAccess
+
+    class ModelProxy < BasicObject
+      ALLOWED_CALLS = {
+        ::ActionTag => [:options_for_project],
+        ::ApiKey => [:create_inference_api_key, :create_personal_access_token, :project_id_for_personal_access_token],
+        ::DiscountCode => [:first],
+        ::GithubInstallation => [:with_github_installation_id],
+        ::GithubRepository => [:cache_size_limit],
+        ::InferenceEndpoint => [:is_public],
+        ::InferenceRouterModel => [:where],
+        ::Invoice => [:blob_storage_client],
+        ::LoadBalancer => [:stack_options],
+        ::Location => [:for_project, :postgres_locations, :visible_or_for_project],
+        ::LockedDomain => [:with_pk],
+        ::ObjectTag => [:options_for_project],
+        ::PaymentMethod => [:fraud?],
+        ::PostgresResource => [:available_flavors, :default_flavor, :default_version, :ha_type_none, :maintenance_hour_options, :partner_notification_flavors],
+        ::PostgresServer => [:victoria_metrics_client],
+        ::SubjectTag => [:admin_tag?, :options_for_project, :subject_id_map_for_project_and_accounts],
+        ::Vm => [:from_runtime_jwt_payload]
+      }.freeze
+      ALLOWED_CALLS.each_value(&:freeze)
+
+      def initialize(model)
+        @model = model
+        @allow = [:===, :create, :create_with_id, :new, :new_with_id, :ubid_format, :ubid_type]
+        if (allow = ALLOWED_CALLS[model])
+          @allow.concat(allow)
+        end
+        @allow.freeze
+      end
+
+      def method_missing(m, ...)
+        if @allow.include?(m)
+          @model.send(m, ...)
+        # :nocov:
+        else
+          ::Kernel.raise DirectModelAccess, "Calling #{@model}.#{m} directly in Clover is not allowed"
+        end
+      end
+
+      def respond_to_missing?(m, _include_all)
+        @allow.include?(m)
+      end
+      # :nocov:
+
+      def sequel_model
+        @model
+      end
+    end
+
+    def self.models_loaded
+      skip_models = %w[Account OidcProvider].freeze
+      Sequel::Model.subclasses.each do |model|
+        name = model.name
+        next unless /\A[A-Za-z0-9]+\z/.match?(name) && !skip_models.include?(name)
+        const_set(name, ModelProxy.new(model))
+      end
+    end
+  # :nocov:
+  else
+    def self.models_loaded
+      # nothing
+    end
+  end
+  # :nocov:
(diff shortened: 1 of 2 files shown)


spec/routes/web/project_spec.rb
```diff
@@ -717,10 +717,7 @@
       it "can not have more than 50 pending invitations" do
         visit "#{project.path}/user"
 
-        expect(described_class).to receive(:authorized_project).with(user, project.id).and_return(project)
-        expect(project).to receive(:invitations_dataset).and_return(instance_double(Sequel::Dataset, count: 50))
-        expect(project).to receive(:invitations_dataset).and_call_original
-
+        ProjectInvitation.import([:project_id, :email, :inviter_id, :expires_at], Array.new(50) { [project.id, "a#{it}@example.come", user.id, Time.utc(2100)] })
```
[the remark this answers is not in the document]
jeremyevans: I mean, technically this is correct, but all that matters is the emails are unique, so it is inconsequential. That being said, I'll change it.

helpers/postgres.rb
```diff
@@ -160,7 +160,7 @@ def generate_postgres_options(flavor: nil, location: nil)
   end
 
   def postgres_flavors
-    Option::POSTGRES_FLAVOR_OPTIONS.reject { |k, _| k == PostgresResource::Flavor::LANTERN && !@project.get_ff_postgres_lantern }
+    Option::POSTGRES_FLAVOR_OPTIONS.reject { |k, _| k == "lantern" && !@project.get_ff_postgres_lantern }
```
enescakir: Isn’t the whole point of `Flavor::LANTERN` to avoid using string literals?

model/github/github_repository.rb
```diff
@@ -17,6 +17,9 @@ class GithubRepository < Sequel::Model
   dataset_module Pagination
 
   CACHE_SIZE_LIMIT = 10 * 1024 * 1024 * 1024 # 10GB
+  def self.cache_size_limit
+    CACHE_SIZE_LIMIT
+  end
```
enescakir: Since we only use it here, is there any reason not to return it directly?

```suggestion
  def self.cache_size_limit
    10 * 1024 * 1024 * 1024 # 10GB
  end
```

enescakir (approved):
It affects all developers and how they interact with the codebase. Could you send a note to eng@?

model/github/github_repository.rb
jeremyevans: This calculates it every time instead of only once. Maybe with the JIT it ends up optimized the same, though. My preference would be to use `10_737_418_240 # 10GB` to avoid even the potential for an issue, though.

helpers/postgres.rb
jeremyevans: I think so. I took a shortcut here. One fix here is moving the method implementation to PostgresResource so the constant is accessible. I'll work on that.

jeremyevans:
Addressed those two issues and rebased on main. Waiting for another approval as @enescakir requested one in addition to his own.

model/postgres/postgres_resource.rb
```diff
@@ -213,17 +217,43 @@ module HaType
     SYNC = "sync"
   end
 
+  def self.ha_type_none
+    HaType::NONE
+  end
```
enescakir, line 222: Since we stopped using direct constant references to hide models, we might move those constants to separate files so we can continue to use them directly.

jeremyevans: We could do something like this in Clover before loading the model hiding code: `PostgresResourceHaType = PostgresResource::HaType`. Then you could reference the constant in Clover using `PostgresResourceHaType::NONE`.

PR #4436, opened 2025-12-22
Use the CYCLE clause in authorization CTEs

jeremyevans:
Previously, we relied on a maximum depth as the only limit.  We also take steps to prevent cycles from being introduced.  However, it's best to short stop processing cycles completely. This uses the CYCLE clause added in PostgreSQL 14 to stop recursive CTE processing of rows that represent a cycle.


fdr (approved):
well, TIL, pretty handy. I am leaning on the test suite here, the change is pretty detailed.

PR #4437, opened 2025-12-23
Improve Vm::Nexus location/nic/subnet validation

jeremyevans:
This handles the following validation issues that existed:

* You could pick a project-specific location that was tied to another project.

* You could pick a subnet that is not in the given location.

* If you know the nic id but no other information, you could gather information about it:

  * Whether it has a VM installed
  * Whether it is in the given location
  * Whether it is in the given project

I don't think any of these issues are actual problems, because:

* In routes/project/vm.rb, we call `check_visible_location` to only allow locations tied to the project.

* In helpers/vm.rb in `vm_post`, we check the subnet is in the given location.

* We don't expose creating a Vm with a specific nic to the user.

So these validations are redundant with previous validations, at least for web/api requests. However, there may be other internal cases where this layer of validations could prevent issues. On the flip side, there may be internal cases where these added validations cause problems because the code actually wants to create VMs in a project associated to a location outside the project, or a subnet in a different location.

This allows selecting a project-specific location in a different project if the VM is allowed to choose a private subnet in a different location. This is necessary for a couple PostgreSQL specs to pass. I think it makes sense because the PostgresResource is in a customer project, in potentially a project-specific location, but it will create server VMs in the postgres project.

This also changes the `unless nic` into `else`, because if a nic_id is provided, the code will either skip the `unless nic` branch or raise an exception.

This adds the Project.nics association to allow only finding nics related to the project.

This updates the victoria metrics resource prog spec to crate the resource in the victoria metrics project, instead of a different project. This is confusing as the same local variable name is used in the spec to refer to two separate projects.

### DIFF
--- a/prog/test/github_runner.rb
+++ b/prog/test/github_runner.rb
@@ -14,7 +14,7 @@ class Prog::Test::GithubRunner < Prog::Test::Base
 
     if provider == "aws"
       github_test_project.set_ff_aws_alien_runners_ratio(1)
-      location = Location.create_with_id(Config.github_runner_aws_location_id, name: "eu-central-1", provider: "aws", project_id: github_test_project.id, display_name: "aws-e2e", ui_name: "aws-e2e", visible: true)
+      location = Location.create_with_id(Config.github_runner_aws_location_id, name: "eu-central-1", provider: "aws", project_id: github_service_project.id, display_name: "aws-e2e", ui_name: "aws-e2e", visible: true)
       LocationCredential.create_with_id(location.id, access_key: Config.e2e_aws_access_key, secret_key: Config.e2e_aws_secret_key)
     end
 


furkansahin (approved):

jeremyevans:
This fails E2E with: `No existing location in project` when allocating a GitHub runner VM:

```
22:05:36 respirate.2 | {"exception":{"message":"No existing location in project","class":"RuntimeError","backtrace":["/home/runner/work/ubicloud/ubicloud/prog/vm/nexus.rb:25:in 'Prog::Vm::Nexus.assemble'" ... "message":"exception terminates strand run","time":"2025-12-29 22:05:39 +0000","thread":"grnd63yhcbp63j6mxpa3te5pcv"}
```

 So we either need to relax that constraint for GitHub runners, or just drop it all together and allow any location as we did previously, relying on the existing checks to protect access from Clover. @enescakir which of those options seems preferable?

enescakir:
It succeeded on metal and failed on AWS. This is probably not directly related to GitHub runners, it may be related to the GitHub + AWS combination.

enescakir:
It looks like we pass the customer project when starting GitHub AWS E2E tests, but we create VMs under the GitHub service project. In production, the AWS location belongs to the service project.

So we likely need to use the service project in E2E tests instead of the customer project.
https://github.com/ubicloud/ubicloud/blob/1ba8fb7e5ab8c9fafe1f7eb304e27dfb96f9109a/prog/test/github_runner.rb#L17

enescakir:
I tested it and the E2E tests passed. This makes the E2E setup closer to production.
You can add this commit to your PR.

fix: https://github.com/ubicloud/ubicloud/commit/df71b315ab852b6cc27c5d039f332afa8ef131cf
run: https://github.com/ubicloud/ubicloud/actions/runs/20585030838/job/59119811235

jeremyevans:
Thanks. So this was an E2E test issue and not an issue with the validations themselves? I added that commit, with an updated commit message. I'll run E2E tests after CI passes.

jeremyevans:
E2E still failing, but no errors related to Vm::Nexus. Similar to other E2E errors that have occurred recently. I'll rerun.

jeremyevans:
E2E passed! I'll merge this tomorrow.

PR #4438, opened 2025-12-23
Add retry logic when instance is unsupported in the AZ

furkansahin:
When AWS returns an Unsupported error during instance creation, it typically means the requested instance type is not available in the current availability zone. Previously, this would cause provisioning to fail and get stuck.

This change adds retry logic that:
- Catches Unsupported errors during instance creation
- Excludes the failed availability zone from future attempts
- Recreates the NIC in a different AZ using the exclusion list
- Limits retries to 5 attempts to prevent infinite loops
- Logs retry attempts for observability

The implementation introduces two new state machine labels: wait_old_nic_deleted waits for the old NIC to be fully deleted before creating a new one, preventing resource conflicts. wait_nic_recreated waits for the new NIC to be ready before retrying instance creation.

The retry state (count, excluded AZs, and subnet ID) is preserved in the strand stack to survive across state transitions.

fixes https://linear.app/ubicloud/issue/UBI-252/implement-az-changeretry-in-case-of-capacity-issues

### DIFF
--- a/prog/vm/aws/nexus.rb
+++ b/prog/vm/aws/nexus.rb
@@ -15,7 +15,7 @@ class Prog::Vm::Aws::Nexus < Prog::Base
 
   label def start
     register_deadline("wait", 10 * 60)
-    nap 1 unless vm.nic.strand.label == "wait"
+    nap 1 unless nic.strand.label == "wait"
     # Cloudwatch is not needed for runner instances
     hop_create_instance if is_runner?
 
@@ -163,7 +163,7 @@ class Prog::Vm::Aws::Nexus < Prog::Base
       ],
       network_interfaces: [
         {
-          network_interface_id: vm.nic.nic_aws_resource.network_interface_id,
+          network_interface_id: nic.nic_aws_resource.network_interface_id,
           device_index: 0
         }
       ],
@@ -202,6 +202,16 @@ class Prog::Vm::Aws::Nexus < Prog::Base
         nap 0
       end
       raise
+    rescue Aws::EC2::Errors::Unsupported => e
+      if (retry_count = frame["retry_count"] || 0) >= 5
+        raise e
+      end
+
+      Clog.emit("unsupported instance type") { {unsupported_instance_type: {vm:, message: e.message, retry_count: retry_count + 1}} }
+      azs_excluded = ((nic.strand.stack.first["exclude_availability_zones"] || []) + [nic.nic_aws_resource.subnet_az]).uniq
+      update_stack({"exclude_availability_zones" => azs_excluded, "retry_count" => retry_count + 1})
+      nic.incr_destroy
+      hop_wait_old_nic_deleted
     end
     instance = instance_response.instances.first
     instance_id = instance.instance_id
@@ -213,6 +223,18 @@ class Prog::Vm::Aws::Nexus < Prog::Base
     AwsInstance.create_with_id(vm, instance_id:, az_id:, ipv4_dns_name:, iam_role: role_name)
 
     hop_wait_instance_created
+  end
+
+  label def wait_old_nic_deleted
+    nap 1 if nic&.reload
+    nic = Prog::Vnet::NicNexus.assemble(frame["private_subnet_id"], name: vm.name + "-nic", exclude_availability_zones: frame["exclude_availability_zones"]).subject
+    nic.update(vm_id: vm.id)
+    hop_wait_nic_recreated
+  end
+
+  label def wait_nic_recreated
+    nap 1 unless nic.strand.label == "wait"
+    hop_create_instance
   end
 
   label def wait_instance_created
@@ -354,9 +376,13 @@ class Prog::Vm::Aws::Nexus < Prog::Base
   end
 
   def final_clean_up
-    vm.nic.update(vm_id: nil)
-    vm.nic.incr_destroy
+    nic.update(vm_id: nil)
+    nic.incr_destroy
     vm.destroy
+  end
+
+  def nic
+    @nic ||= vm.nic
   end
 
   def client
--- a/spec/prog/vm/aws/nexus_spec.rb
+++ b/spec/prog/vm/aws/nexus_spec.rb
@@ -453,6 +453,61 @@ usermod -L ubuntu
       expect(vm.nics.first).to receive(:nic_aws_resource).and_return(instance_double(NicAwsResource, network_interface_id: "eni-0123456789abcdefg"))
       expect { nx.create_instance }.to raise_error(Aws::EC2::Errors::InsufficientInstanceCapacity)
     end
+
+    describe "when unsupported instance type error" do
+      let(:nic) { vm.nics.first }
+
+      before do
+        client.stub_responses(:run_instances, Aws::EC2::Errors::Unsupported.new(nil, "Instance type not supported in availability zone"))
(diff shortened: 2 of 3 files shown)


prog/vm/aws/nexus.rb
```diff
@@ -202,6 +202,16 @@ def before_run
         nap 0
       end
       raise
+    rescue Aws::EC2::Errors::Unsupported => e
```
shikharbhardwaj, line 205: Should we also check if the error message matches somewhat to make sure its the instance type being unavailable in the AZ?

```
Your requested instance type (m8gd.large) is not supported in your requested Availability Zone (us-east-1e). Please retry your request by not specifying an Availability Zone or choosing us-east-1a, us-east-1b, us-east-1c, us-east-1d, us-east-1f.
```

prog/vm/aws/nexus.rb
```diff
@@ -202,6 +202,16 @@ def before_run
...
+      if (retry_count = frame["retry_count"] || 0) >= 5
+        raise e
+      end
+
+      Clog.emit("unsupported instance type") { {unsupported_instance_type: {vm:, message: e.message, retry_count: retry_count + 1}} }
+      azs_excluded = ((vm.nic.strand.stack.first["exclude_availability_zones"] || []) + [vm.nic.nic_aws_resource.subnet_az]).uniq
+      update_stack({"exclude_availability_zones" => azs_excluded, "private_subnet_id" => vm.nic.private_subnet_id, "retry_count" => retry_count + 1})
+      vm.nic.incr_destroy
+      hop_wait_old_nic_deleted
```
shikharbhardwaj: jw: With the NIC recreate involved, is there value in keeping a support matrix for instance types vs azs to avoid this path?

There is an aws CLI command to do this ([ref](https://coding-stream-of-consciousness.com/2021/08/27/listing-supported-availability-zones-azs-for-instance-types-in-aws/))

```
% aws ec2 describe-instance-type-offerings --location-type availability-zone --filters="Name=instance-type,Values=r5.2xlarge" --region us-east-1 --output table
-------------------------------------------------------
|            DescribeInstanceTypeOfferings            |
+-----------------------------------------------------+
||               InstanceTypeOfferings               ||
|+--------------+--------------+---------------------+|
|| InstanceType |  Location    |    LocationType     ||
|+--------------+--------------+---------------------+|
||  r5.2xlarge  |  us-east-1f  |  availability-zone  ||
||  r5.2xlarge  |  us-east-1c  |  availability-zone  ||
||  r5.2xlarge  |  us-east-1b  |  availability-zone  ||
||  r5.2xlarge  |  us-east-1d  |  availability-zone  ||
||  r5.2xlarge  |  us-east-1a  |  availability-zone  ||
|+--------------+--------------+---------------------+
```

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -453,6 +453,60 @@
...
+        expect(st.stack.last["exclude_availability_zones"]).to eq(["a"])
+        expect(st.stack.last["private_subnet_id"]).to eq(nic.private_subnet_id)
+        expect(st.stack.last["retry_count"]).to eq(1)
+        expect(nic.reload.destroy_set?).to be true
+      end
+
+      it "increments retry count on subsequent failures" do
+        st.stack.first["retry_count"] = 2
+        nx.instance_variable_set(:@frame, nil) # Clear cached frame
+        nic.strand.stack.first["exclude_availability_zones"] = ["b", "c"]
```
enescakir: You can use our new helper
https://github.com/ubicloud/ubicloud/blob/2a64fd4cdc52e85f61926f72bc8f7607642d54e2/spec/spec_helper.rb#L314

```suggestion
        refresh_frame(nx, new_values: {"retry_count" => 2})
        nic.strand.stack.first["exclude_availability_zones"] = ["b", "c"]
```

prog/vm/aws/nexus.rb
```diff
@@ -202,6 +202,16 @@ def before_run
...
       end
       raise
+    rescue Aws::EC2::Errors::Unsupported => e
+      if (retry_count = frame["retry_count"] || 0) >= 5
+        raise e
+      end
+
+      Clog.emit("unsupported instance type") { {unsupported_instance_type: {vm:, message: e.message, retry_count: retry_count + 1}} }
+      azs_excluded = ((vm.nic.strand.stack.first["exclude_availability_zones"] || []) + [vm.nic.nic_aws_resource.subnet_az]).uniq
+      update_stack({"exclude_availability_zones" => azs_excluded, "private_subnet_id" => vm.nic.private_subnet_id, "retry_count" => retry_count + 1})
```
jeremyevans: This is the same behavior (I think), but does not give the false impression that the private_subnet depends on the nic.

```suggestion
      update_stack({"exclude_availability_zones" => azs_excluded, "private_subnet_id" => frame["private_subnet_id"], "retry_count" => retry_count + 1})
```

prog/vm/aws/nexus.rb
```diff
@@ -202,6 +202,16 @@ def before_run
         nap 0
       end
       raise
+    rescue Aws::EC2::Errors::Unsupported => e
+      if (retry_count = frame["retry_count"] || 0) >= 5
+        raise e
+      end
+
+      Clog.emit("unsupported instance type") { {unsupported_instance_type: {vm:, message: e.message, retry_count: retry_count + 1}} }
+      azs_excluded = ((vm.nic.strand.stack.first["exclude_availability_zones"] || []) + [vm.nic.nic_aws_resource.subnet_az]).uniq
```
jeremyevans: Consider adding a `nic = vm.nic` local variable.

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -453,6 +453,60 @@
...
+
+      it "raises exception after 5 retry attempts" do
+        st.stack.first["retry_count"] = 5
+        nx.instance_variable_set(:@frame, nil) # Clear cached frame
+        expect(Clog).not_to receive(:emit).with("unsupported instance type")
+        expect { nx.create_instance }.to raise_error(Aws::EC2::Errors::Unsupported)
+        expect(nic.reload.destroy_set?).to be false
+      end
+
+      it "logs retry count in emission" do
```
jeremyevans, line 499: This doesn't check the retry count is actually logged.

prog/vm/aws/nexus.rb
furkansahin: No, since we don't have any control on the error message and AWS is not guaranteed to not change a specific part of the message, I'd avoid this. The error type will not change, we know that, though.

prog/vm/aws/nexus.rb
furkansahin: You mean fetching AZs for a given region? I mean, we don't want to make this recreation a regular part of the code. This should behave as a recovery logic. We simply fetch which AZs available for a given region right now, we do not store if a given instance type is offered in a region. We might add that in future and make this reprovisioning logic useful when out of capacity happens in a given region

prog/vm/aws/nexus.rb
furkansahin: The frame did not have private_subnet_id unless we started executing this recreation logic but I have added that to the `vm/nexus.rb` now.

prog/vm/aws/nexus.rb
furkansahin: created a cached method instead.

spec/prog/vm/aws/nexus_spec.rb
furkansahin: fixed it

PR #4439, opened 2025-12-23
Provision new standby when unavailable

furkansahin:
## Extract standby provisioning logic into PostgresResource model
The logic for determining exclusion filters when provisioning a new
standby server was embedded in ConvergePostgresResource. This made it
difficult to reuse when provisioning standbys from other contexts.

This change moves the provisioning logic into the PostgresResource
model by introducing:
- provision_new_standby: High-level method that provisions a standby
  with appropriate exclusion filters
- aws_new_server_exclusion_filters: AWS-specific logic that excludes
  used availability zones or targets a specific AZ
- metal_new_server_exclusion_filters: Metal-specific logic that
  excludes data centers where servers already exist

The refactoring reduces duplication and makes the provisioning logic
accessible from any context that needs to create standby servers.

## Provision new standby when server becomes unavailable
When a Postgres server fails availability checks during checkup, the
cluster may fall below the target standby count. Previously, this
required manual intervention or waiting for the next convergence cycle.

This change automatically provisions a replacement standby when a
server is detected as unavailable:
- The checkup label now triggers ConvergePostgresResource when the
  server is unavailable
- A "recycle" semaphore prevents duplicate provisioning attempts
- The server is restarted to attempt recovery while provisioning the
  replacement

The recycle semaphore is cleared when the server becomes available
again, allowing future unavailability events to trigger provisioning.

This ensures high availability clusters maintain their target standby
count even when individual servers fail.

## Add comprehensive tests for provision_new_standby behavior
Add tests for the provision_new_standby method that was extracted from
ConvergePostgresResource. These tests cover the different server
placement strategies across providers (Hetzner, AWS) and configurations.

Tests moved from converge_postgres_resource_spec to
postgres_resource_spec
because provision_new_standby is now a model method rather than program
logic. The tests verify:

- Host exclusion behavior with Config.allow_unspread_servers
- Data center exclusion for metal providers
- Availability zone exclusion for AWS when use_different_az_set? is true
- AZ placement in same zone when use_different_az_set? is false
- Timeline selection for regular instances vs read replicas

Additionally, add a test for the AWS-specific provisioning path in
ConvergePostgresResource that provisions servers even when not all
existing servers are assigned to vm_hosts yet. This reflects AWS's
ability to provision more aggressively compared to metal providers.

Finally, add missing test coverage for the case when a server becomes
unavailable and recycle_set? returns true. Previously only the false
case was tested.

When a server enters the unavailable state and recycle is already set,
we should not call incr_recycle again or create another
ConvergePostgresResource
strand. However, we still need to bud a restart attempt. This prevents
redundant recycle operations while ensuring recovery is attempted.

The test also fixes an existing test that was missing the decr_recycle
expectation when the server becomes available again.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

## Add data center exclusion filter to the allocator
This way, we will make use of the filter without passing all the
vm_hosts in a datacenter in a later commit.

## Switch from exclude_host_ids to exclude_data_centers for standbys
This way, we filter and pass a much smaller list.

## Replace array-based exclusion filters with named struct
The new_server_exclusion_filters methods previously returned arrays with
positional elements for exclude_data_centers,
exclude_availability_zones, and availability_zone.

### DIFF
--- a/spec/model/postgres/postgres_resource_spec.rb
+++ b/spec/model/postgres/postgres_resource_spec.rb
@@ -170,6 +170,22 @@ RSpec.describe PostgresResource do
       expect(new_server.timeline_id).to eq(parent_timeline.id)
       expect(new_server.vm.strand.stack[0]["exclude_host_ids"]).to eq([])
     end
+
+    it "provisions a new server excluding the representative server's host for leaseweb" do
+      allow(Config).to receive(:allow_unspread_servers).and_return(false)
+      expect(postgres_resource.location).to receive(:provider).and_return(HostProvider::LEASEWEB_PROVIDER_NAME).at_least(:once)
+      ps1
+      vm_host = create_vm_host
+      vm1.update(vm_host_id: vm_host.id)
+      ps1.update(representative_at: Time.now)
+
+      expect(Prog::Postgres::PostgresServerNexus).to receive(:assemble).with(hash_including(exclude_host_ids: [vm_host.id])).and_call_original
+
+      postgres_resource.provision_new_standby
+      expect(postgres_resource.reload.servers.count).to eq(2)
+      new_server = PostgresServer.exclude(id: ps1.id).first
+      expect(new_server.vm.strand.stack[0]["exclude_host_ids"]).to eq([vm_host.id])
+    end
   end
 
   it "returns has_enough_fresh_servers correctly" do


model/postgres/aws/postgres_resource.rb
```diff
@@ -13,5 +13,20 @@ def aws_upgrade_candidate_server
...
+
+    # The first element is the list of vm_hosts to exclude from the new server,
+    # which is empty for aws servers.
+    # The second element is the list of availability zones to exclude from the new server,
+    # which is the list of all used availability zones.
+    # The third element is the availability zone of the representative server,
+    # which is the availability zone of the new server.
+    def aws_new_server_exclusion_filters
+      exclude_availability_zones, availability_zone = if use_different_az_set?
+        [servers.map { it.vm.nic.nic_aws_resource.subnet_az }.uniq, nil]
```
jeremyevans: 3N+1 query issue. Possible fixes are eager loading:

```suggestion
        [servers(eager: {vm: {nic: :nic_aws_resource}}).map { it.vm.nic.nic_aws_resource.subnet_az }.uniq, nil]
```

or better, getting all information in a single query:

```suggestion
        subnet_azs = NicAwsResource
          .where(id: Nic.where(vm_id: servers_dataset.select(:vm_id).select(:id))
          .distinct
          .select_map(:subnet_az)
        [subnet_azs, nil]
```

model/postgres/metal/postgres_resource.rb
```diff
@@ -10,5 +10,17 @@ def metal_upgrade_candidate_server
...
+    # If the server is in leaseweb, we don't have multiple DCs anyway.
+    # The first element is the list of vm_hosts to exclude from the new server,
+    # which means the dc of the existing servers for metal
+    # The second element is the list of availability zones to exclude from the new server,
+    # which is empty for metal.
+    # The third element is the availability zone of the representative server,
+    # which is nil for metal.
+    def metal_new_server_exclusion_filters
+      return [[], [], nil] if Config.allow_unspread_servers || location.provider == HostProvider::LEASEWEB_PROVIDER_NAME
+      [VmHost.where(data_center: servers.map { it.vm.vm_host.data_center }.uniq).map(&:id), [], nil]
```
jeremyevans: 2N+1 query issue. Possible fix:

```suggestion
      hosts_ds = VmHost
        .where(id: Vm.where(id: servers_dataset.select(:vm_id))
      exclude_host_ids = VmHost
        .where(data_center: hosts_ds.select(:data_center)
        .distinct
        .select_map(:id)
      [exclude_host_ids, [], nil]
```

We should probably change the allocator to accept an `exclude_data_center` argument so you could just pass the data centers in use (or a dataset like `hosts_ds.select(:data_center)` as shown above), avoiding the need to retrieve the ids of all hosts to exclude.

model/postgres/postgres_resource.rb
```diff
@@ -104,6 +104,12 @@ def version
     representative_server&.version || target_version
   end
 
+  def provision_new_standby
+    exclude_host_ids, exclude_availability_zones, availability_zone = new_server_exclusion_filters
+    timeline_id = read_replica? ? parent.timeline.id : timeline.id
+    Prog::Postgres::PostgresServerNexus.assemble(resource_id: id, timeline_id:, timeline_access: "fetch", exclude_host_ids: exclude_host_ids, exclude_availability_zones: exclude_availability_zones, availability_zone: availability_zone)
```
jeremyevans: Hash value omission:

```suggestion
    Prog::Postgres::PostgresServerNexus.assemble(resource_id: id, timeline_id:, timeline_access: "fetch", exclude_host_ids:, exclude_availability_zones:, availability_zone:)
```

model/postgres/metal/postgres_resource.rb
jeremyevans, line 1: Commit message should be updated: `s/PostgresResource class context/PostgresResource instance context/`

spec/model/postgres/postgres_resource_spec.rb
```diff
@@ -60,6 +60,83 @@
     expect(s).to include("ubi_replication@#{postgres_resource.ubid}.postgres.ubicloud.com", "application_name=pgubidstandby", "sslcert=/etc/ssl/certs/server.crt")
   end
 
+  describe "#provision_new_standby" do
+    it "provisions a new server without excluding hosts when Config.allow_unspread_servers is true for regular instances" do
+      allow(Config).to receive(:allow_unspread_servers).and_return(true)
+      vm_rep = create_hosted_vm(project, private_subnet, "pg-vm-rep")
+      PostgresServer.create(timeline:, resource_id: postgres_resource.id, vm_id: vm_rep.id,
+        representative_at: Time.now, synchronization_status: "ready", timeline_access: "push", version: "16")
+      expect(Prog::Postgres::PostgresServerNexus).to receive(:assemble).with(hash_including(exclude_host_ids: []))
```
jeremyevans: In general, we should avoid mocking progs and models. In this case, I think it makes sense, because there isn't a good way otherwise to determine which host exclusions were used. However, you should add `.and_call_original` and check that the expected `PostgresServer` was created.  Same issue in other specs.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -852,16 +852,32 @@
...
       expect(postgres_server).to receive(:trigger_failover).and_return(false)
       expect(nx).to receive(:available?).and_return(true)
+      expect(nx).to receive(:decr_recycle)
       expect { nx.unavailable }.to hop("wait")
     end
 
     it "buds restart if the server is not available" do
-      postgres_resource = instance_double(PostgresResource, ongoing_failover?: false)
-      expect(postgres_server).to receive(:resource).and_return(postgres_resource)
+      postgres_resource = instance_double(PostgresResource, ongoing_failover?: false, id: "resource-id")
```
jeremyevans: We should avoid mocking for this. Setup the database state, run the label, and then check that the expected database changes were made.

jeremyevans: This avoids creating the instance double in this particular example, but still uses the instance double used by the whole spec and still mocks everything. However, since that is an existing issue, and not specific to this PR, it's fine. This spec will eventually be democked (or at least have mocking limited to places where it is required).

model/postgres/aws/postgres_resource.rb
```diff
@@ -13,5 +13,26 @@ def aws_upgrade_candidate_server
         .select { |server| PgAwsAmi.where(aws_ami_id: server.vm.boot_image).count > 0 }
         .max_by(&:created_at)
     end
+
+    # The first element is the list of vm_hosts to exclude from the new server,
```
fdr: vm_hosts? in the aws file? I also don't see it referenced here?

model/postgres/aws/postgres_resource.rb
```diff
@@ -13,5 +13,26 @@ def aws_upgrade_candidate_server
...
+          .join(:nic, id: :id)
+          .where(vm_id: servers_dataset.select(:vm_id))
+          .distinct
+          .select_map(:subnet_az)
+
+        [subnet_azs, nil]
+      else
+        [[], representative_server.vm.nic.nic_aws_resource.subnet_az]
+      end
+      [[], exclude_availability_zones, availability_zone]
```
fdr: is a three-tuple (array of three elements) the right thing, or is it worth a Struct or something like that?

model/postgres/metal/postgres_resource.rb
```diff
@@ -10,5 +10,26 @@ def metal_upgrade_candidate_server
...
+
+      exclude_host_ids = VmHost
+        .where(data_center: VmHost
+          .join(:vm, vm_host_id: :id)
+          .where(Sequel[:vm][:id] => servers_dataset.select(:vm_id))
+          .select(:data_center)
+          .distinct)
+        .select_map(:id)
+
+      [exclude_host_ids, [], nil]
```
fdr: I'm wondering about the first element of the array in general: is it passing around quite a lot of state, that can go stale under concurrency (however unlikely that is given the frequency of host churn for us...). Does it make more sense to add exclusions to the allocater? e.g. "do not match `FSN-DC17`" (or however it goes)

fdr:
Looks like I found at least some stuff same as @jeremyevans. probably not a blocker unless you feel like it. I can't think of something right off to endorse a giant exclusion list of VmHosts, but I haven't actually tried to think about the change to the allocater

model/postgres/metal/postgres_resource.rb
furkansahin: This one is added

model/postgres/aws/postgres_resource.rb
furkansahin: This is just a comment. Since the function return types must match with the metal file and vm_hosts were being returned there, I mad it obvious by writing the comment. Though, now this is changed to data_centers.

model/postgres/aws/postgres_resource.rb
furkansahin: switched to struct

prog/postgres/postgres_server_nexus.rb
```diff
@@ -548,9 +548,15 @@ def before_run
 
     if available?
       decr_checkup
+      decr_recycle
       hop_wait
     end
 
+    unless postgres_server.recycle_set?
+      postgres_server.incr_recycle
+      Strand.create(prog: "Postgres::ConvergePostgresResource", label: "start", stack: [{subject_id: postgres_server.resource.id}])
```
shikharbhardwaj: jw: should we check if another convergence is running already before starting one here? The converge prog has some upgrade related operations too which might not work if run simultaneously

model/postgres/aws/postgres_resource.rb
```diff
@@ -32,7 +26,7 @@ def aws_new_server_exclusion_filters
       else
         [[], representative_server.vm.nic.nic_aws_resource.subnet_az]
       end
-      [[], exclude_availability_zones, availability_zone]
+      ServerExclusionFilters.new(exclude_data_centers: [], exclude_availability_zones:, availability_zone:)
```
shikharbhardwaj: Might be a fixup to the 1st commit in the branch

prog/postgres/postgres_server_nexus.rb
```diff
@@ -550,9 +550,15 @@ def before_run
...
```
byucesoy: You don't need to create `ConvergePostgresResource` strand. PostgresResource already does that when any of its servers has `recycle` semaphore set. Though, you can create and increment a dummy semaphore (`wake_up`) to speed up the processing.

model/postgres/metal/postgres_resource.rb
```diff
@@ -21,15 +21,15 @@ def metal_upgrade_candidate_server
     def metal_new_server_exclusion_filters
       return [[], [], nil] if Config.allow_unspread_servers || location.provider == HostProvider::LEASEWEB_PROVIDER_NAME
 
-      exclude_host_ids = VmHost
+      exclude_data_centers = VmHost
```
byucesoy, line 20: Not specific to this PR, but I remembered it seeing this. I'd appreciate if you  can add logic to not use primary's host for leaseweb databases.

Currently we pass empty array if the provider is leaseweb in the line above. We don't have data centers there, but we can at least put the primary and standby to different hosts.

furkansahin: updated

prog/postgres/postgres_server_nexus.rb
furkansahin: updated

PR #4440, opened 2025-12-23
Download boot images from R2 in production by default

enescakir:
We started downloading boot images from R2 in E2E tests several months ago. We’ve also been downloading them from R2 manually in production.

It has been reliable and performed well so far. We can enable R2 by default in production and disable it only when necessary.

### DIFF
--- a/prog/download_boot_image.rb
+++ b/prog/download_boot_image.rb
@@ -26,7 +26,7 @@ class Prog::DownloadBootImage < Prog::Base
   end
 
   def download_from_r2?
-    frame["download_r2"] || Config.is_e2e
+    frame["download_r2"] || Config.production?
   end
 
   def url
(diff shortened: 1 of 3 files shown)


prog/download_boot_image.rb
```diff
@@ -26,7 +26,7 @@ def default_boot_image_version(image_name)
   end
 
   def download_from_r2?
-    frame["download_r2"] || Config.is_e2e
+    frame["download_r2"] || Config.production?
```
jeremyevans, line 29: If we are always defaulting to `download_r2: true` (not sure if we are in all cases), maybe we should eventually switch this to:

```suggestion
    frame["download_r2"] != false
```

We probably don't want to do this right away, though, because of the effect on existing strands.

enescakir: We'll do it eventually. We'll probably also remove MinIO support at some point, but until then, I don’t want to disrupt people’s development environments.

PR #4441, opened 2025-12-24
Make API endpoint descriptions more concise

enescakir:
I was reviewing our API docs to use it and noticed that many endpoint titles are overly wordy, which makes it harder to scan the list quickly.

For example, since tokens are already project-specific, we can drop suffixes like "in the project". Similarly, because delete/show endpoints already require the location parameter, we can remove "in location" from those titles.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -105,7 +105,7 @@ paths:
       - $ref: '#/components/parameters/github_installation_reference'
     get:
       operationId: getGithubInstallation
-      summary: Get a GitHub installation
+      summary: Get GitHub installation details
       responses:
         '200':
           $ref: '#/components/responses/GithubInstallation'
@@ -134,7 +134,7 @@ paths:
       - $ref: '#/components/parameters/github_repository_reference'
     get:
       operationId: getGithubRepository
-      summary: Get a GitHub repository
+      summary: Get GitHub repository details
       responses:
         '200':
           $ref: '#/components/responses/GithubRepository'
@@ -175,7 +175,7 @@ paths:
       - $ref: '#/components/parameters/github_cache_entry_id'
     get:
       operationId: getGithubCacheEntry
-      summary: Get a cache entry for a GitHub repository
+      summary: Get cache entry details
       responses:
         '200':
           $ref: '#/components/responses/GithubCacheEntry'
@@ -185,7 +185,7 @@ paths:
         - GitHub
     delete:
       operationId: deleteGithubCacheEntry
-      summary: Delete a cache entry for a GitHub repository
+      summary: Delete a cache entry
       responses:
         '204':
           $ref: '#/components/responses/Delete'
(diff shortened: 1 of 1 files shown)


enescakir:
It’s a first pass using an LLM. We can iterate on it and improve it further.

openapi/openapi.yml
```diff
@@ -311,7 +311,7 @@ paths:
         - Firewall
     get:
       operationId: getLocationFirewallDetails
-      summary: Get details of a specific firewall
+      summary: Get firewall details
```
jeremyevans, line 314: This is inconsistent with some of the above (e.g. `Get a GitHub repository`). We should probably use a consistent format (e.g. ` Get GitHub repository details` or `Get firewall`) unless we have a reason for deviating.

openapi/openapi.yml
```diff
@@ -927,7 +927,7 @@ paths:
   '/project/{project_id}/location/{location}/postgres':
     get:
       operationId: listLocationPostgresDatabases
-      summary: List Postgres Databases in a specific location of a project
+      summary: List postgres databases in a location
```
jeremyevans: I recommend using `PostgreSQL` instead of `postgres` here and in other descriptions.

```suggestion
      summary: List PostgreSQL databases in a location
```

openapi/openapi.yml
```diff
@@ -1076,7 +1076,7 @@ paths:
       - $ref: '#/components/parameters/postgres_database_reference'
     get:
       operationId: getPostgresCACertificatesByName
-      summary: Download CA certificates for a specific Postgres Database in a location with name
+      summary: Download CA certificates
```
jeremyevans: The description here should mention PostgreSQL.

```suggestion
      summary: Download CA certificates for a PostgreSQL database
```

openapi/openapi.yml
```diff
@@ -1115,7 +1115,7 @@ paths:
         - Postgres Database
     post:
       operationId: updatePostgresDatabaseConfig
-      summary: Update configuration for a specific Postgres Database
+      summary: Update postgres database configuration
```
jeremyevans:
```suggestion
      summary: Replace PostgreSQL database configuration
```

openapi/openapi.yml
```diff
@@ -1135,7 +1135,7 @@ paths:
         - Postgres Database
     patch:
       operationId: patchPostgresDatabaseConfig
-      summary: Partially update configuration for a specific Postgres Database
+      summary: Partially update postgres database configuration
```
jeremyevans:
```suggestion
      summary: Update PostgreSQL database configuration
```

jeremyevans (approved):
This is definitely an improvement on what we have.

openapi/openapi.yml
enescakir: I unified them using the `Get * details` format

openapi/openapi.yml
enescakir: I replaced them

PR #4442, opened 2025-12-24
Don't choke in admin panel over UBID false positive

serprex:
Rendering began failing after creating test-upstream-20251224-arm

### DIFF
--- a/views/admin/object.erb
+++ b/views/admin/object.erb
@@ -44,7 +44,7 @@
     <tr>
       <td><%= k %></td>
       <td>
-        <% if v.is_a?(String) && v.bytesize == 26 && (column_class = UBID.class_for_ubid(v)) && column_class.method_defined?(:ubid) && (obj = UBID.decode(v)) %>
+        <% if v.is_a?(String) && v.bytesize == 26 && (column_class = UBID.class_for_ubid(v)) && column_class.method_defined?(:ubid) && (obj = UBID.decode(v) rescue nil) %>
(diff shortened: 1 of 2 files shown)


enescakir:
Can you add a test for it?

https://github.com/ubicloud/ubicloud/blob/main/spec/routes/web/admin/admin_spec.rb

views/admin/object.erb
```diff
@@ -44,7 +44,7 @@
     <tr>
       <td><%= k %></td>
       <td>
-        <% if v.is_a?(String) && v.bytesize == 26 && (column_class = UBID.class_for_ubid(v)) && column_class.method_defined?(:ubid) && (obj = UBID.decode(v)) %>
+        <% if v.is_a?(String) && v.bytesize == 26 && (column_class = UBID.class_for_ubid(v)) && column_class.method_defined?(:ubid) && (obj = UBID.decode(v) rescue nil) %>
```
jeremyevans, line 47: @pykello what are your thoughts on `UBID.decode` returning `nil` instead of raising for an invalid format? We already return `nil` if there is no entry for the related class. It seems odd to have one failure case return `nil` and another failure case raise. If we want a version that raises, we should probably add `UBID.decode!` that raises for all failure cases.

jeremyevans: I've submitted #4467 to change `UBID.decode`. Merging that will fix this issue.

jeremyevans:
I've merged #4467, which should fix this issue.

PR #4446, opened 2025-12-25
Fix double nested exception logs

enescakir:
`Util.exception_to_hash` already returns a hash with an exception key, so wrapping it again with another exception key is unnecessary and results in double nesting.

### DIFF
--- a/db.rb
+++ b/db.rb
@@ -24,7 +24,7 @@ Util.safe_write_to_file(postgres_monitor_db_ca_bundle_filename, Config.postgres_
 begin
   POSTGRES_MONITOR_DB = Sequel.connect(Config.postgres_monitor_database_url, max_connections: Config.db_pool_monitor, pool_timeout: Config.database_timeout, driver_options:) if Config.postgres_monitor_database_url
 rescue Sequel::DatabaseConnectionError => ex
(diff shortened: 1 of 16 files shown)


enescakir:
> Avoiding the double wrapping is good.
> 
> Future ideas in this area:
> 
> * To avoid unnecessary hash merging, we may want to consider adding either a `no_wrap` or `into` keyword argument to `Util.exception_to_hash` (`no_wrap` removing the outer hash, `into` setting the `exception` key into the given hash).
> * It's weird that we log the `cause` differently (using `inspect`) in `Util.exception_to_hash`. In `respirate`, we emit nested exceptions as separate clogs, so all causes are logged twice, in different manners. Maybe it would be better to instead log an array of exceptions, which would only contain a single exception if the exception did not have a cause, and avoid the loop calling `Clog.emit` separately for each cause in `respirate`?
> 
> Let me know if you think either of those things makes sense, I'm happy to implement either.

@jeremyevans, I think both approaches make sense. The latter has made production debugging harder a few times, as it tends to make the logs noisy.

jeremyevans:
> @jeremyevans, I think both approaches make sense. The latter has made production debugging harder a few times, as it tends to make the logs noisy.

I added #4465, which builds on top of this and includes both changes.

enescakir:
Fixed at https://github.com/ubicloud/ubicloud/pull/4465

PR #4448, opened 2025-12-26
Add set feature flag action to the admin panel

enescakir:
- **Allow different input field types in the admin panel**
  Currently, admin panel action forms only support text input fields.
  
  It's useful to support select input fields when there's a predefined set
  of options to choose from.
  
  If the value of an input is a hash instead of a string, each property of
  the hash is used separately.
  

- **Handle CloverError exceptions in the admin panel**
  This allows us to display custom error messages instead of generic
  internal server errors.
  

- **Add set feature flag action to the admin panel**
  This new action allows feature flags to be set for projects directly
  from the admin panel.
  
  Since feature flags are predefined, a dropdown is used to select the
  flag.
  
  The value is optional, if no value is provided, it is treated as nil.
  
  The value is a JSON field, allowing it to be cast to a boolean, integer,
  float, array, or hash when possible.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -178,6 +178,22 @@ class CloverAdmin < Roda
     "Project" => {
       "add_credit" => object_action("Add credit", "Added credit", {credit: "float!"}) do |obj, credit|
         obj.this.update(credit: Sequel[:credit] + credit)
+      end,
+      "set_feature_flag" => object_action("Set Feature Flag", "Set feature flag", {
+        name: {
+          typecast: :str!,
+          type: "select",
+          add_blank: true,
+          options: Project.instance_methods.filter_map { it.to_s.delete_prefix("set_ff_") if it.start_with?("set_ff_") }
+        },
+        value: {typecast: :nonempty_str, placeholder: "JSON", required: nil}
+      }) do |obj, name, value|
+        begin
+          value = JSON.parse(value) if value
+        rescue JSON::ParserError
+          fail CloverError.new(400, "InvalidRequest", "invalid JSON for feature flag value")
+        end
+        obj.send("set_ff_#{name}", value)
       end
     },
     "Strand" => {
--- a/spec/routes/web/admin/admin_spec.rb
+++ b/spec/routes/web/admin/admin_spec.rb
@@ -644,6 +644,44 @@ RSpec.describe CloverAdmin do
     expect(page.title).to eq "Ubicloud Admin - Project #{p.ubid}"
   end
 
+  it "supports setting feature flags of Project" do
+    p = Project.create(name: "Default")
+
+    fill_in "UBID or UUID", with: p.ubid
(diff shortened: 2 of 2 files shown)


clover_admin.rb
```diff
@@ -178,6 +178,17 @@ def self.object_action(...)
     "Project" => {
       "add_credit" => object_action("Add credit", "Added credit", {credit: "float!"}) do |obj, credit|
         obj.this.update(credit: Sequel[:credit] + credit)
+      end,
+      "set_feature_flag" => object_action("Set Feature Flag", "Set feature flag", {
+        name: {typecast: :str!, type: "select", options: [""] + Project.instance_methods.filter_map { it.to_s.delete_prefix("set_ff_") if it.start_with?("set_ff_") }},
```
aamederen: nit: It's a bit long and difficult to read for me, so maybe:

```suggestion
        name: {
          typecast: :str!, 
          type: "select", 
          options: [""] + Project.instance_methods.filter_map { it.to_s.delete_prefix("set_ff_") if it.start_with?("set_ff_") }
        },
```

aamederen (approved):

views/admin/object_action.erb
```diff
@@ -1,8 +1,18 @@
...
+      type = "text"
+      opts = {key: name, label: name}
+      if value.is_a?(Hash)
+        type = value.fetch(:type, "text")
+        opts[:required] = value.fetch(:required, true)
+        opts[:options] = value[:options]
+        opts[:placeholder] = value[:placeholder]
+      end
+    %>
+    <%== f.input(type, **opts) %>
```
jeremyevans: I don't think we should limit the options we are passing to `f.input`. I think we should pass the entire hash (Forme will ignore options it doesn't recognize):

```suggestion
      if value.is_a?(Hash)
        type = value.fetch(:type, "text")
        value_opts = value
      end
    %>
    <%== f.input(type, key: name, label: name, required: true, **value_opts) %>
```

clover_admin.rb
```diff
@@ -174,6 +178,17 @@ def self.object_action(...)
...
```
jeremyevans: Forme supports an `add_blank` option for an empty `<option>` tag:

```suggestion
        name: {typecast: :str!, type: "select", add_blank: true, options: Project.instance_methods.filter_map { it.to_s.delete_prefix("set_ff_") if it.start_with?("set_ff_") }},
```

jeremyevans (approved):
I'm a little skeptical of accepting JSON via a text input, but without this, you are forced to use pry to do basically the same thing, so I think it makes sense to support this.

views/admin/object_action.erb
enescakir: Makes sense. That made it clearer. Thanks.

clover_admin.rb
enescakir: Cool!

PR #4449, opened 2025-12-26
Make PageNexus.assemble upsert existing Pages instead of ignoring them

aamederen:
Previously, when PageNexus.assemble was called with tag_parts that matched an existing active Page, it would simply return early and ignore the new data. This meant that updates to page summaries, details, or severity would be lost.

Now, when an existing active Page is found by tag, it gets updated with the new summary, details, and severity. This ensures Pages always reflect the most current state.

### DIFF
--- a/prog/page_nexus.rb
+++ b/prog/page_nexus.rb
@@ -5,10 +5,19 @@ class Prog::PageNexus < Prog::Base
 
   def self.assemble(summary, tag_parts, related_resources, severity: "error", extra_data: {})
     DB.transaction do
-      return if Page.from_tag_parts(tag_parts)
+      details = extra_data.merge({"related_resources" => Array(related_resources)})
+      tag = Page.generate_tag(tag_parts)
+      page = Page.new(summary:, details:, tag:, severity:)
+      page.skip_auto_validations(:unique) do
+        page.insert_conflict(
+          target: :tag,
+          conflict_where: {resolved_at: nil},
+          update: {summary: Sequel[:excluded][:summary], details: Sequel[:excluded][:details], severity: Sequel[:excluded][:severity]}
+        ).save_changes
+      end
 
-      pg = Page.create(summary:, details: extra_data.merge({"related_resources" => Array(related_resources)}), tag: Page.generate_tag(tag_parts), severity:)
-      Strand.create_with_id(pg, prog: "PageNexus", label: "start")
+      Strand.new(prog: "PageNexus", label: "start") { it.id = page.id }
+        .insert_conflict(target: :id).save_changes
     end
   end
 
--- a/spec/prog/page_nexus_spec.rb
+++ b/spec/prog/page_nexus_spec.rb
@@ -28,4 +28,57 @@ RSpec.describe Prog::PageNexus do
       expect { pn.wait }.to nap(6 * 60 * 60)
     end
   end
+
+  describe ".assemble" do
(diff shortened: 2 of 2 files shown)


enescakir:
Is it possible to upsert using `insert_conflict` instead of if/else block?

aamederen:
> Is it possible to upsert using insert_conflict instead of if/else block?

I tried, but the resulting code was not simpler nor shorter so I skipped. Maybe, in order to reduce DB back-and-forths, we can try to be smarter and update only if something has changed.

enescakir:
It's something like that, but of course, the final decision is yours.
```
diff --git a/prog/page_nexus.rb b/prog/page_nexus.rb
index 4649431ac..60e9c1982 100644
--- a/prog/page_nexus.rb
+++ b/prog/page_nexus.rb
@@ -5,21 +5,18 @@ class Prog::PageNexus < Prog::Base

   def self.assemble(summary, tag_parts, related_resources, severity: "error", extra_data: {})
     DB.transaction do
-      if (pg = Page.from_tag_parts(tag_parts))
-        pg.update(
-          summary: summary,
-          details: extra_data.merge({"related_resources" => Array(related_resources)}),
-          severity: severity
-        )
-      else
-        pg = Page.create(
-          summary: summary,
-          details: extra_data.merge({"related_resources" => Array(related_resources)}),
-          tag: Page.generate_tag(tag_parts),
-          severity: severity
-        )
-        Strand.create_with_id(pg, prog: "PageNexus", label: "start")
+      details = Sequel.pg_jsonb(extra_data.merge({"related_resources" => Array(related_resources)}))
+      tag = Page.generate_tag(tag_parts)
+      page = Page.new(summary:, details:, tag:, severity:)
+      page.skip_auto_validations(:unique) do
+        page.insert_conflict(
+          target: :tag,
+          conflict_where: {resolved_at: nil},
+          update: {summary:, details:, severity:}
+        ).save_changes
       end
+      Strand.new(prog: "PageNexus", label: "start") { it.id = page.id }
+        .insert_conflict(target: :id).save_changes
     end
   end
```

prog/page_nexus.rb
```diff
@@ -5,10 +5,21 @@ class Prog::PageNexus < Prog::Base
...
+        )
+      else
+        pg = Page.create(
+          summary: summary,
+          details: extra_data.merge({"related_resources" => Array(related_resources)}),
+          tag: Page.generate_tag(tag_parts),
+          severity: severity
+        )
+        Strand.create_with_id(pg, prog: "PageNexus", label: "start")
+      end
```
enescakir: It doesn't allow to add it as a suggestion 🤔 

nit:
```ruby
      details = extra_data.merge({"related_resources" => Array(related_resources)})
      if (pg = Page.from_tag_parts(tag_parts))
        pg.update(summary:, details:, severity:)
      else
        pg = Page.create(tag: Page.generate_tag(tag_parts), summary:, details:, severity:)
        Strand.create_with_id(pg, prog: "PageNexus", label: "start")
      end
```

jeremyevans (commented):
`insert_conflict` may not be simpler nor shorter, but it correctly handles race conditions and the current PR approach does not. For that reason, we should use it. @enescakir's diff looks good.

This needs specs added.

aamederen:
Switched to insert_conflict based on @enescakir's suggestion. Updated the specs a bit too.

jeremyevans (approved):

PR #4450, opened 2025-12-26
Add client helper to GithubInstallation model

enescakir:
Constructing a GitHub API client required passing installation_id to Github.installation_client, which was repetitive. The new GithubInstallation#client method encapsulates this, allowing callers to simply use installation.client.

It's useful for the oncall operations as well.

### DIFF
--- a/prog/github/github_runner_nexus.rb
+++ b/prog/github/github_runner_nexus.rb
@@ -147,12 +147,12 @@ class Prog::Github::GithubRunnerNexus < Prog::Base
     end
   end
 
-  def github_client
-    @github_client ||= Github.installation_client(installation.installation_id, auto_paginate: true)
+  def client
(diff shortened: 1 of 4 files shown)


prog/github/github_runner_nexus.rb
```diff
@@ -147,12 +147,12 @@ def update_billing_record
     end
   end
 
-  def github_client
-    @github_client ||= Github.installation_client(installation.installation_id, auto_paginate: true)
+  def client
+    @client ||= Github.installation_client(installation.installation_id, auto_paginate: true)
```
jeremyevans: Should this be:?

```suggestion
    @client ||= installation.client(auto_paginate: true)
```

enescakir: Yeah, it looks like I overwrote this change at some point.

PR #4455, opened 2025-12-29
Demock sem snap spec

fdr:


fdr:
bonus demock

fdr:
more bonus

spec/lib/sem_snap_spec.rb
```diff
@@ -9,7 +9,7 @@
         snap.incr(:test)
       }.to change { snap.set?(:test) }.from(false).to(true)
 
-      expect(Semaphore.where(strand_id: st.id).empty?).to be false
+      expect(Semaphore.where(strand_id: st.id)).not_to be_empty
```
jeremyevans: This issue with this approach (and the previous approach) is if the dataset is empty, the failure message is unhelpful. `expect(Semaphore.where(strand_id: st.id).all).not_to be_empty` is better, because the failure message shows which semaphores exist.

spec/lib/validation_spec.rb
```diff
@@ -360,14 +360,36 @@
...
+      project.add_quota(quota_id: ProjectQuota.default_quotas["VmVCpu"]["id"], value: 10)
+    end
+
     it "sufficient cpu quota" do
-      p = instance_double(Project, current_resource_usage: 5, effective_quota_value: 10, quota_available?: true)
-      expect { described_class.validate_vcpu_quota(p, "VmVCpu", 2) }.not_to raise_error
+      # In total, one VM occupying five plus a request for two more,
+      # seven total, is less than the fixture total of ten.
+      Vm.create_with_id(
+        Vm.generate_uuid,
```
jeremyevans: No need to use `create_with_id` if you are going to pass it `generate_uuid`:

```suggestion
      Vm.create(
```

Similar issue in other specs.

spec/lib/validation_spec.rb
```diff
@@ -360,14 +360,36 @@
...
+        unix_user: "ubi", public_key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIGWmPgJE", name: "test-vm", family: "standard", cores: 1,
+        vcpus: 5, cpu_percent_limit: 100, cpu_burst_percent_limit: 100, memory_gib: 4,
+        arch: "x64", boot_image: "ubuntu-jammy", display_state: "running", ip4_enabled: true,
+        created_at: Time.now, location_id: Location::HETZNER_FSN1_ID, project_id: project.id
+      )
+      expect { described_class.validate_vcpu_quota(project, "VmVCpu", 2) }.not_to raise_error
```
jeremyevans: `expect{ ... }.not_to raise_error` is in general a bad pattern. You should run `...`, and if it raises an error, then the spec fails. Maybe the only reason this is done is due to rubocop/rspec restrictions that a spec must have an expectation, but IMO, we would be better off disabling that, or adding a helper method like `force_expectation!` that did `expect(true).to be true` or something and calling that.

jeremyevans (approved):

spec/lib/validation_spec.rb
fdr: I'm more neutral on this subject. If I didn't have not_to raise_error, I'd probably feel inclined to add a comment explaining that this dangling line is not a half-finished thought, but the crux of a test.  I follow your reasoning though, that it does nothing, so why have it? (except it exists a dual to the other test).

One way to shave some lines and make it faster, and make it more clear might be to allocate a second VM, then demonstrate the failure, after demonstrating success without remark. Then the two tests will not decorrelate in time.

jeremyevans: I don't object to keeping it as it is (I did approve the change after all). But if you want to refactor it, I'm fine with that as well.

fdr: See what you think of my fixup. It's a more complex test but also couples the two cases together neatly, I think, whereas the two could have drifted otherwise.

jeremyevans: I like this new approach a lot more. :)

jeremyevans (approved):

fdr:
alright rebase-autosquash and waiting for the farm:

```
 git diff c7c6d9297 e5519afd4 --stat
fdr@albatross ~/c/clover ((e5519afd))> 
(no changes)
```

PR #4462, opened 2025-12-30
Add set quota action to the admin panel

enescakir:
Admins can now override per-project quota values for any resource type defined in default_quotas.yml. Leaving the value blank removes the override and reverts to the default quota for the project's reputation.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -186,7 +186,11 @@ class CloverAdmin < Roda
           add_blank: true,
           options: Project.instance_methods.filter_map { it.to_s.delete_prefix("set_ff_") if it.start_with?("set_ff_") }
         },
-        value: {typecast: :nonempty_str, placeholder: "JSON", required: nil}
+        value: {
+          typecast: :nonempty_str,
+          placeholder: "JSON",
+          required: nil
+        }
       }) do |obj, name, value|
         begin
           value = JSON.parse(value) if value
@@ -194,6 +198,31 @@ class CloverAdmin < Roda
           fail CloverError.new(400, "InvalidRequest", "invalid JSON for feature flag value")
         end
         obj.send("set_ff_#{name}", value)
+      end,
+      "set_quota" => object_action("Set Quota", "Set quota", {
+        resource_type: {
+          typecast: :str!,
+          type: "select",
+          add_blank: true,
+          options: ProjectQuota.default_quotas.keys
+        },
+        value: {
+          typecast: :int,
+          type: "number",
+          placeholder: "blank to reset to default",
+          required: nil
+        }
+      }) do |obj, resource_type, value|
+        quota_id = ProjectQuota.default_quotas[resource_type]["id"]
+        if (existing_quota = obj.quotas_dataset.first(quota_id:))
+          if value
+            existing_quota.update(value:)
+          else
+            existing_quota.destroy
+          end
+        elsif value
+          obj.add_quota(quota_id:, value:)
+        end
       end
     },
     "Strand" => {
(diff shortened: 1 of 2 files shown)


clover_admin.rb
```diff
@@ -194,6 +194,26 @@ def self.object_action(...)
...
         obj.send("set_ff_#{name}", value)
+      end,
+      "set_quota" => object_action("Set Quota", "Set quota", {
+        resource_type: {
+          typecast: :str!,
+          type: "select",
+          add_blank: true,
+          options: ProjectQuota.default_quotas.keys
+        },
+        value: {typecast: :int, type: "number", placeholder: "blank to reset to default", required: nil}
```
jeremyevans: Seems weird to format the `value` hash different from the `resource_type` hash. Please use a single line for both, or use a separate line per key for both.

clover_admin.rb
```diff
@@ -194,6 +194,26 @@ def self.object_action(...)
...
+      }) do |obj, resource_type, value|
+        quota_id = ProjectQuota.default_quotas[resource_type]["id"]
+        if (existing_quota = obj.quotas_dataset.first(quota_id:))
+          if value.nil?
+            existing_quota.destroy
+          else
+            existing_quota.update(value:)
+          end
+        else
+          obj.add_quota(quota_id:, value:)
```
jeremyevans: We don't check for a `nil` value here. We should, and not add the quota in this case. Otherwise, you get a `Sequel::ValidationFailed` exception when trying to add a new quota with a `nil` value:

```suggestion
        elsif value
          obj.add_quota(quota_id:, value:)
```

clover_admin.rb
```diff
@@ -194,6 +194,26 @@ def self.object_action(...)
...
+        },
+        value: {typecast: :int, type: "number", placeholder: "blank to reset to default", required: nil}
+      }) do |obj, resource_type, value|
+        quota_id = ProjectQuota.default_quotas[resource_type]["id"]
+        if (existing_quota = obj.quotas_dataset.first(quota_id:))
+          if value.nil?
+            existing_quota.destroy
+          else
+            existing_quota.update(value:)
+          end
```
jeremyevans: In general, only use `nil?` if you care about the difference between `nil` and `false`. Since `false` is not a valid value here:

```suggestion
          if value
            existing_quota.update(value:)
          else
            existing_quota.destroy
          end
```

spec/routes/web/admin/admin_spec.rb
```diff
@@ -682,6 +682,49 @@ def page_data
...
+    expect(page).to have_flash_notice("Set quota")
+    expect(p.effective_quota_value("VmVCpu")).to eq(0)
+
+    # Remove quota when value is blank
+    visit path
+    select "VmVCpu", from: "resource_type"
+    fill_in "value", with: ""
+    click_button "Set Quota"
+    expect(page).to have_flash_notice("Set quota")
+    expect(p.effective_quota_value("VmVCpu")).to eq(32)
```
jeremyevans: Update the specs to test for `nil` value handling when there is no existing quota:

```suggestion
    2.times do
      # Remove quota when value is blank
      # Ensure value being blank doesn't add quota
      visit path
      select "VmVCpu", from: "resource_type"
      fill_in "value", with: ""
      click_button "Set Quota"
      expect(page).to have_flash_notice("Set quota")
      expect(p.effective_quota_value("VmVCpu")).to eq(32)
    end
```

clover_admin.rb
enescakir: Good finding. Thanks

PR #4465, opened 2025-12-30
Update Util.exception_to_hash

jeremyevans:
This makes two changes to Util.exception.to_hash:

* It adds an into keyword argument. This will add the :exception
  key to that hash instead of always using a new hash. For backwards
  compatibility, the default is a new empty hash.

* It uses an array to hold the exception information. If backtraces are
  being logged (the default), it includes all exception causes in the
  array.

This switches callers of the form:

```ruby
{...}.merge(Util.exception_to_hash(ex))
```

to:

```ruby
Util.exception_to_hash(ex, into: {...})
```

Which eliminates 2 hash allocations per call (as merge returns a new
hash).

This also removes the loop in the dispatcher to emit each exception
cause as a separate log entry, since now the single log entry includes
all causes.

This builds on top of #4446.

### DIFF
--- a/helpers/runtime.rb
+++ b/helpers/runtime.rb
@@ -22,7 +22,7 @@ class Clover < Roda
     begin
       jobs = runner.installation.client.workflow_run_jobs(runner.repository_name, run_id)[:jobs]
     rescue Octokit::ClientError, Octokit::ServerError, Faraday::ConnectionFailed, Faraday::TimeoutError => ex
-      log_context[:expection] = Util.exception_to_hash(ex)
+      Util.exception_to_hash(ex, into: log_context)
       Clog.emit("Could not list the jobs of the workflow run ") { {runner_scope_failure: log_context} }
       return
     end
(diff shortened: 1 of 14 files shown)


enescakir:
I noticed something in the exception array logic.

In our log provider, we have several views and graphs that query `exception.message`. If we change `exception` to an array, can we still query it as `exception.0.message`? If that’s not possible, it would be good to check before converting it to an array.

As an alternative, we could keep the root exception at `exception.message` and add an `exception.causes` array to store the additional exceptions. This way also we don't break existing queries and developer experience.

jeremyevans:
> As an alternative, we could keep the root exception at `exception.message` and add an `exception.causes` array to store the additional exceptions. This way also we don't break existing queries and developer experience.

That seems reasonable to me. I can switch this PR to use that approach.

jeremyevans:
@enescakir I rebased this on main, fixed the conflicts, and moved only the causes to an array.

PR #4468, opened 2025-12-30
Enforce hash value omission in Ruby code

jeremyevans:
Basically the same as #4372, but with the Kubernetes CSI issue addressed, and being rebased on main.


kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -112,7 +112,7 @@ def create_volume(req, _call)
           topology = existing ? existing[:accessible_topology] : selected_topology
           CreateVolumeResponse.new(
             volume: Volume.new(
-              volume_id: volume_id,
```
enescakir, line 115: Don't we try to disable value omission for CSI since it uses Ruby 3.2?

jeremyevans: Hash value omission was added in Ruby 3.1: https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/#other-notable-new-features

serprex (approved):

PR #4473, opened 2025-12-31
Avoid NoMethodError in AWS NicNexus

jeremyevans:
This should avoid an exception seen in production:

```
NoMethodError | undefined method 'network_interface_id' for nil
/app/prog/vnet/aws/nic_nexus.rb:132:in 'Prog::Vnet::Aws::NicNexus#attach_eip_network_interface'
```

I'm not sure if this is the correct behavior if there are no addresses, though. If we should wait for an address, this should be changed to nap until there is an address.

### DIFF
--- a/prog/vnet/aws/nic_nexus.rb
+++ b/prog/vnet/aws/nic_nexus.rb
@@ -123,7 +123,8 @@ class Prog::Vnet::Aws::NicNexus < Prog::Base
 
   label def attach_eip_network_interface
     eip_response = client.describe_addresses({filters: [{name: "allocation-id", values: [nic.nic_aws_resource.eip_allocation_id]}]})
-    unless eip_response.addresses.first.network_interface_id
+    nap(1) unless (address = eip_response.addresses.first)
+    unless address.network_interface_id
       client.associate_address({allocation_id: nic.nic_aws_resource.eip_allocation_id, network_interface_id: nic.nic_aws_resource.network_interface_id})
     end
     hop_wait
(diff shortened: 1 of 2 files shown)


prog/vnet/aws/nic_nexus.rb
```diff
@@ -129,7 +129,7 @@ def before_run
 
   label def attach_eip_network_interface
     eip_response = client.describe_addresses({filters: [{name: "allocation-id", values: [nic.nic_aws_resource.eip_allocation_id]}]})
-    unless eip_response.addresses.first.network_interface_id
+    unless eip_response.addresses.first&.network_interface_id
```
enescakir: Since the exception disappeared in the next run, I thought we should nap instead of skip it.

"Not having an address" and "having an address without an attached network" are different scenarios. The first one shouldn't happen; if it does, we need to wait until the Elastic IP address is created.

In the previous label, we explicitly allocate an address if it's empty, so it shouldn't be empty here.

https://github.com/ubicloud/ubicloud/blob/659b3e667e330e4836a6dc235c82b8f482a3fe0e/prog/vnet/aws/nic_nexus.rb#L120-L122

But of course, Furkan knows these states better.

furkansahin: Yes, we should wait for the address to be available. Please change it to nap

jeremyevans:
Switched this to nap if the address hasn't been setup yet.

enescakir (approved):

PR #4474, opened 2025-12-31
Fix display_state method to handle destroying semaphore like destroy semaphore in multiple models

jeremyevans:
Noticed when I deleted a Postgres database in staging and and saw it go from running to creating in the UI before disappearing.

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -117,7 +117,7 @@ class Vm < Sequel::Model
 
   def display_state
     label = strand&.label
-    return "deleting" if destroy_set? || label == "destroy"
+    return "deleting" if destroying_set? || destroy_set?
     return "restarting" if restart_set? || label == "restart"
     return "stopped" if stop_set? || label == "stopped"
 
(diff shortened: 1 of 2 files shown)


model/ai/inference_endpoint.rb
```diff
@@ -29,8 +29,8 @@ def path
 
   def display_state
     label = strand.label
+    return "deleting" if destroy_set? || destroying_set? || label == "destroy"
```
enescakir: Since we already have the following check, there’s probably no need to check the destroy label in display_states.

https://github.com/ubicloud/ubicloud/blob/580b4cc036a4f870cf718c09d3a21198f6ccb844/prog/base.rb#L81

Not a big issue, but destroying_set? is more likely to be true than destroy_set?, so it could be an earlier exit.

```suggestion
    return "deleting" if destroying_set? || destroy_set?
```

jeremyevans: Thank you, that makes things simpler. I've rebased this and used that approach in all modified methods.

PR #4475, opened 2025-12-31
Support private- DNS names in SSL certificates for PG and load balancers

jeremyevans:
This is needed so that customers connecting using the private DNS names get the same security as those connecting using the public DNS names.


jeremyevans:
I've converted this to draft. The CertNexus tests will be broken as they are mocked. @fdr will be providing spec democking, and after the specs are democked, I'll update them.

I've tested this in staging and it does work for both self-signed PG certificates as well as load balancer certificates signed by a public CA.

jeremyevans:
Thanks to @fdr's democking of CertNexus, I was able to get the specs passing. This is now ready for review.

furkansahin (approved):

jeremyevans:
This now includes the changes I suggested in #4562.

PR #4476, opened 2026-01-01
Fdr demock backport 12 31

fdr:

PR #4481, opened 2026-01-02
Remove unnecessary UpdateLoadBalancerNode spec

jeremyevans:
I assume this spec was only added for coverage, since it's impossible for the algorithm value to be least_conn. Remove unnecessary branch in the prog so this spec is not necessary for coverage.

### DIFF
--- a/prog/vnet/update_load_balancer_node.rb
+++ b/prog/vnet/update_load_balancer_node.rb
@@ -51,10 +51,8 @@ class Prog::Vnet::UpdateLoadBalancerNode < Prog::Base
 
     balance_mode_ip4, balance_mode_ip6 = if load_balancer.algorithm == "round_robin"
       ["numgen inc", "numgen inc"]
-    elsif load_balancer.algorithm == "hash_based"
+    else # elsif load_balancer.algorithm == "hash_based"
       ["jhash ip saddr . tcp sport . ip daddr . tcp dport", "jhash ip6 saddr . tcp sport . ip6 daddr . tcp dport"]
-    else
-      fail ArgumentError, "Unsupported load balancer algorithm: #{load_balancer.algorithm}"
     end
 
     ipv4_prerouting = if load_balancer.ipv4_enabled?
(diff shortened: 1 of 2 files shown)


fdr:
funny thing, the llm did the same thing, but I didn't agree with it, because I don't generally believe in having else get "a" behavior unless "any" behavior is the reasonable interpretation of what to get. It's not that robust under change either.

fdr:
I'm open to a more artful spelling though, or a counter-argument for this one. It just tripped my "non-exhaustive switch in waiting" alarm bell.

fdr (commented):
I suppose my review is a question. Anyone have contrary reactions on the value of exhaustiveness checking, vs. undefined behavior we'll silently (rather than noisily, hitting a deadline, probably) encounter later, vs. one awkward test?

jeremyevans:
My argument is: absent a change to add an algorithm, the `else` approach is already exhaustive. I can see your point about switching to `else` may make it more difficult to find places to change if you do want to add an algorithm. So I don't object to closing this.

fdr:
> My argument is: absent a change to add an algorithm, the `else` approach is already exhaustive. I can see your point about switching to `else` may make it more difficult to find places to change if you do want to add an algorithm. So I don't object to closing this.

How likely is it, you think, that we find this branch in test, or in some other noisy way? It seems like it might silently get you a wrong-thing from a moment of inattentiveness, it's not correlated with adding enumeration elements.

jeremyevans:
Generally, whenever you add an enum value (either an actual enum or updating the CHECK constraint to add another allowed value), you should be searching the entire codebase for the column to find places that need to be updated. However, you probably cannot enforce that, and even if you could, there's always a chance it will be missed. So this does allow at least a small amount of risk. I generally accept this risk on my own projects (and often use the approach in this PR), but the risk calculus is different here, so maybe it's not an appropriate trade-off here.

I noticed this issue during the review of the democking PR (#4476), where it stood out because it continued to use mocking for model methods. However, I think there is another way to avoid the mocking, so this is probably not needed (https://github.com/ubicloud/ubicloud/pull/4476/changes#r2658327117). So I'll close this.

PR #4482, opened 2026-01-02
Allow specifying private subnet name when creating PostgreSQL database in the API/CLI

jeremyevans:
This makes things simpler for the API/CLI user. Before, they could rename the created private subnet, but doing so requires knowing the ubid of the created postgres resource, which makes it more challenging to script. This allows for specifying the name up front, so that after creation, you can add firewalls directly to it without parsing the output from other requests/commands.

Mostly written by Claude Code using the following prompt:

```
Add support for accepting a private subnet name when creating a
postgres resource

* Update assemble in prog/postgres/postgres_resource_nexus.rb to
  accept private subnet name
* Update helpers/postgres.rb to pass private subnet name when calling
  subnet name, for api requests and not web resources.
* Update openapi/openapi.yml to accept private subnet name when
  creating a postgres database
* Update cli-commands/pg/post/create.rb to accept an option to specify
  the private subnet name
* Update spec/routes/api/cli/pg/create_spec.rb to pass the option to
  specify the private subnet name, and check that the created private
  subnet has that name.
```

I changed the CLI option from `-n` to `-P`, and did some minor refactoring.

### DIFF
--- a/cli-commands/pg/post/create.rb
+++ b/cli-commands/pg/post/create.rb
@@ -13,6 +13,7 @@ UbiCli.on("pg").run_on("create") do
     on("-u", "--pgbouncer-config=config", "pgbouncer config (e.g. key1=value1,key2=value2)")
     on("-t", "--tags=tags", "tags (e.g. key1=value1,key2=value2)")
     on("-R", "--restrict-by-default", "restrict access by default (add firewall rules to allow access)")
+    on("-P", "--private-subnet-name=name", "override name of created private subnet")
   end
   help_option_values("Flavor:", Option::POSTGRES_FLAVOR_OPTIONS.keys)
   help_option_values("Replication Type:", Option::POSTGRES_HA_OPTIONS.keys)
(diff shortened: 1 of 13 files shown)


byucesoy (approved):

furkansahin (approved):

fdr:
Well everyone else approved. @furkansahin, @byucesoy, main advisory here is that this fundamentally changes (or reckons with the already-changed) model of subnets belonging to the customer, and the managed resources rest within them. 

This is different than what I had in mind from long ago, that the subnet itself would be a delegated entity to the service provider. However, the people writing the code (so @jeremyevans) were pretty convinced this would be simpler, and I am not so convicted as to disagree, the delegation of entire subnets to providers was a reflexive and abstract conservatism that had no regard for how messy things might get under the covers, so I didn't put too much stock in my initial leaning.

In any case, I found that something like this was instrumental in making my [ubidon](https://github.com/ubicloud/ubidon) script work well and feel like I was working with one product, to avoid chasing or duplicating calculation of identifiers.

PR #4483, opened 2026-01-02
No customer firewall for PG creation through api

jeremyevans:
This changes the API/CLI so that creating a PostgreSQL database no longer creates a customer firewall. Previously, a customer firewall was also created, even if the `-R`/`restrict_by_default` option was used to not have any firewall rules created.

This breaks backwards compatibility. In order for PostgreSQL database creation via API/CLI to allow connections, customers must use the API/CLI to also attach a firewall with appropriate rules to the database's private subnet.

This is a trade-off. It makes it easier to use custom firewall rules. Previously, you would have to parse output from previous CLI commands or API responses to get the firewall name to use. However, this requries additional commands for the basic PostgreSQL database setup to work, since by default, no database access will be allowed until a firewall is manually created, attached to the private subnet, and has appropriate rules added to it. It also makes creation via web have different behavior than creation via api.

The previous `-R`/`restrict_by_default` options are now soft-deprecated. No warnings are displayed, but the documentation and CLI help now show the options as deprecated.

This does not add an API or CLI option to get the previous behavior of creating a customer firewall with the private subnet with open 5432/6432 firewall rules.

This makes the Postgres-level firewall rule commands pretty much pointless for new PostgreSQL databases created via the API/CLI, since those commands require a customer firewall to work. So unless the user manually creates a firewall with the ubid-prefixed name, they won't work.

The Prog::Postgres::PostgresResourceNexus.assemble with_firewall_rules argument now controls whether a firewall is added, instead of whether rules are added to the created firewall. The
Prog::Vnet::SubnetNexus.assemble method now accepts a no_firewall argument, which if set, will not create a Firewall.

### DIFF
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -3,7 +3,7 @@
 class Prog::Vnet::SubnetNexus < Prog::Base
   subject_is :private_subnet
 
-  def self.assemble(project_id, name: nil, location_id: Location::HETZNER_FSN1_ID, ipv6_range: nil, ipv4_range: nil, allow_only_ssh: false, firewall_id: nil, firewall_name: nil)
+  def self.assemble(project_id, name: nil, location_id: Location::HETZNER_FSN1_ID, ipv6_range: nil, ipv4_range: nil, allow_only_ssh: false, firewall_id: nil, firewall_name: nil, no_firewall: false)
     unless (project = Project[project_id])
       fail "No existing project"
     end
@@ -27,7 +27,9 @@ class Prog::Vnet::SubnetNexus < Prog::Base
       ps = PrivateSubnet.create_with_id(id, name:, location_id: location.id, net6: ipv6_range, net4: ipv4_range, state: "waiting", project_id:)
       firewall_dataset = project.firewalls_dataset.where(location_id:)
 
-      if firewall_id
+      if no_firewall
+        # nothing
+      elsif firewall_id
         unless (firewall = firewall_dataset.first(Sequel[:firewall][:id] => firewall_id))
           fail "Firewall with id #{firewall_id} and location #{location.name} does not exist"
         end
@@ -50,7 +52,7 @@ class Prog::Vnet::SubnetNexus < Prog::Base
           ["0.0.0.0/0", "::/0"].each { |cidr| FirewallRule.create(firewall_id: firewall.id, cidr:, port_range: Sequel.pg_range(port_range)) }
         end
       end
-      firewall.associate_with_private_subnet(ps, apply_firewalls: false)
+      firewall&.associate_with_private_subnet(ps, apply_firewalls: false)
 
       prog = location.aws? ? "Vnet::Aws::VpcNexus" : "Vnet::Metal::SubnetNexus"
       Strand.create_with_id(id, prog:, label: "start")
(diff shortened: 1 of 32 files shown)


prog/vnet/subnet_nexus.rb
```diff
@@ -50,7 +52,7 @@ def self.assemble(project_id, name: nil, location_id: Location::HETZNER_FSN1_ID,
           ["0.0.0.0/0", "::/0"].each { |cidr| FirewallRule.create(firewall_id: firewall.id, cidr:, port_range: Sequel.pg_range(port_range)) }
         end
       end
-      firewall.associate_with_private_subnet(ps, apply_firewalls: false)
+      firewall&.associate_with_private_subnet(ps, apply_firewalls: false)
```
furkansahin, line 55:
```suggestion
      firewall.associate_with_private_subnet(ps, apply_firewalls: false) unless no_firewall
```

furkansahin (approved):
LGTM

prog/vnet/subnet_nexus.rb
jeremyevans: Is this just a style preference, or is there an actual issue with using `firewall&.`? I'm fine merging purely for style reasons, but I want to understand the reason.

jeremyevans:
@fdr I pushed a commit for the change we discussed at the team meeting, please let me know what you think.

fdr (commented):
leaving it to @furkansahin, my feedback to Jeremy was mostly about making sure https://github.com/ubicloud/ubidon Got Nicer as a result of these changes (whacking it around in staging, not prod)

byucesoy (changes requested):
I’d like to sleep on this for a bit.

- Having different default behaviors between the web and API endpoints is unexpected.
- We could make the web follow the same pattern, but that would optimize the default for production use at the cost of a more complex onboarding and first-use experience. In general, I think we should prefer the opposite trade-off.

I need to think on this for a while. As a quick thought, would not deprecating `--restrict-by-default` and not creating customer firewall if it is passed would satisfy the primary motivation of making `ubidon` and other similar script ops easier?

fdr:
> I’d like to sleep on this for a bit.
> 
>     * Having different default behaviors between the web and API endpoints is unexpected.
> 
>     * We could make the web follow the same pattern, but that would optimize the default for production use at the cost of a more complex onboarding and first-use experience. In general, I think we should prefer the opposite trade-off.
> 
> 
> I need to think on this for a while. As a quick thought, would not deprecating `--restrict-by-default` and not creating customer firewall if it is passed would satisfy the primary motivation of making `ubidon` and other similar script ops easier?

As I was working with setting up my subnets for an app, I was kinda knocked back the platform would "do that to me" when I was reading back the output on inspection, it was only by chance I noticed this rule got slid in. It's not how tying together subnets works, so it felt like quite a transition from one set of features to the other.  Unlike the web, the options for displaying information for read-back are more limited.

 https://github.com/ubicloud/ubidon/blob/02dab05fda1b0dab5821306d40c900a49148916d/ubidon-provision.bash#L97-L125

fdr:
I ran into a time crunch but discussed with Jeremy a bit. I think we need to survey our existing defaults and write a theory of the user and why they are the way they are, and if any should be changed. Please hold.

jeremyevans:
Switched to draft until we determine what we want to do.

jeremyevans:
I'm going to close this. If we decide we do want to change the api behavior, we can reopen and rebase. However, I agree that having web/api behavior be different could be problematic. Potentially a better approach would be defaulting to having new postgresql resources in a single subnet per project named `postgres`. That way, whatever firewall rules you have for that subnet will be used by default for new postgres resources.

PR #4484, opened 2026-01-02
Support postgres private DNS in the web UI

jeremyevans:
If the postgres resource subnet is connected to another subnet, support showing connection info with the private DNS name. It's possible we should support it in all cases, even if not connected to another subnet, since the user can add VMs directly to the postgres resource subnet, but I think we want to encourage using connected subnets instead of adding resources to postgres resource subnets.

Update the specs to correctly test the DNS names.

This is mostly based on one of the commits in #3901, with appropriate updates. While it does not  depend on #4475, it shouldn't be merged until after #4475, because we want the PG SSL certificates  to support the private DNS name before implicitly recommending that customers use them for connecting.


fdr (commented):
I'm going to punt this one to @byucesoy and @furkansahin, since at least they should be familiar with how the gui is changing

byucesoy (approved):

furkansahin (approved):

jeremyevans:
Updated this to use `private.` instead of `private-`.

jeremyevans:
In addition to rebasing this, I've added the following changes:

* Support hostname_version v3
* Include ssl parameters in all connection strings, and use verify-full for publicly signed certificates.
* Update api/cli to also show private connection strings
* Don't attempt to use publicly signed certificates when no matching DNS zone exists.

shikharbhardwaj (approved):

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/28903444011

PR #4486, opened 2026-01-05
Add invoice table to admin panel

enescakir:
- **Allow admin panel links to use columns other than name**
  By default, admin panel links display the name column. However, it’s now
  possible to customize which column is used for link text.
  
  This is useful in cases where showing the ubid or another attribute is
  more informative than the name.
  

- **Add invoice table to admin panel**

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -272,10 +272,12 @@ class CloverAdmin < Roda
 
     show_html do |obj, column|
       case column
-      when :name
-        link.call(obj)
+      when :name, :invoice_number
+        link.call(obj, label: column)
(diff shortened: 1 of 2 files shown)


clover_admin.rb
```diff
@@ -264,10 +264,10 @@ def self.object_action(...)
     supported_actions [:browse, :search]
     form_options(wrapper: :div)
 
-    link = lambda do |obj|
+    link = lambda do |obj, label = "name"|
```
jeremyevans: I think it may be better to use a keyword argument for this, but it's up to you.

PR #4487, opened 2026-01-05
Observe archival backlog and page if it exceeds threshold (try 2)

shikharbhardwaj:
Add an SSH call to count the number of WAL files ready to be archived and raise a Page if it exceeds a threshold. In the previous attempt, this was added to pulse_check, which caused a problem. Since postgres server pulse checks require an event loop, running exec! on the same session is not safe, and can cause state corruption which lead to stuck SSH sessions. In this attempt, we move the backlog check to the metrics export path.

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -278,6 +278,22 @@ class PostgresServer < Sequel::Model
     vm.sshable.cmd("PGOPTIONS='-c statement_timeout=60s' psql -U postgres -t --csv -v 'ON_ERROR_STOP=1'", stdin: query).chomp
   end
 
+  def export_metrics(session:, tsdb_client:)
+    session[:export_count] ||= 0
+    session[:export_count] += 1
+
+    # Check archival backlog every 12 exports (similar to pulse check frequency)
+    # We do this in metrics export rather than pulse check because the metrics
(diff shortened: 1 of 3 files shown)


model/postgres/postgres_server.rb
```diff
@@ -323,6 +339,35 @@ def refresh_walg_credentials
     refresh_walg_blob_storage_credentials
   end
 
+  def observe_archival_backlog(session:)
```
jeremyevans: If a method accepts a single argument, it should generally be positional and not keyword unless you already have future plans to add additional required arguments.

model/postgres/postgres_server.rb
```diff
@@ -323,6 +339,35 @@ def refresh_walg_credentials
...
+    result = session[:ssh_session].exec!(
+      "sudo find /dat/:version/data/pg_wal/archive_status -name '*.ready' | wc -l",
+      version:
+    )
+    archival_backlog = result.strip.to_i
```
jeremyevans: In general, if this should return a single number as a string, it's best to use a strict conversion instead of a loose conversion.

```suggestion
    Integer(archival_backlog = result.strip, 10)
```

byucesoy (approved):

PR #4490, opened 2026-01-05
[destroying #8] Pop extra stack frames in before_run during destruction

enescakir:
When a strand has multiple frames in its stack and is destroyed, the top frame destroys the associated entities and pops. The next frame then continues to work but fails because the entities are already destroyed. This is fixed by first popping all frames but the last one, then switching to the destroy label to handle cleanup.

Several progs (InferenceEndpointReplicaNexus, InferenceRouterReplicaNexus, InferenceRouterTargetNexus, MinioServerNexus) had custom before_run implementations to handle this case. We added these because we hit the issue before, but any prog that pushes other progs could encounter it. Moving this logic into the base before_run handles it universally.

### DIFF
--- a/prog/base.rb
+++ b/prog/base.rb
@@ -77,7 +77,7 @@ end
 
   def before_run
     if defined?(destroy_set?)
-      unless destroying_set?
+      if !destroying_set?
         fail "BUG: destroying semaphore not set on destroy label" if @strand.label == "destroy"
         if destroy_set?
           send(:before_destroy) if respond_to?(:before_destroy)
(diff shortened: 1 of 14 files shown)


prog/base.rb
```diff
@@ -87,6 +87,13 @@ def before_run
             pop "exiting early due to destroy semaphore"
           end
         end
+      elsif destroy_set? && @strand.stack.count > 1
+        # When a strand has multiple frames in its stack, and it is destroyed, the top
+        # frame destroys the associated entities and pops. Then next frame continues to
+        # work, but fails because the entities are already destroyed. This is fixed by
+        # first popping all frames but the last one, and then switching to the destroy
+        # label to handle cleanup.
+        pop "operation is cancelled due to the destruction"
```
jeremyevans:
```suggestion
        pop "operation is cancelled due to explicitly requested destruction"
```

byucesoy (approved):

PR #4494, opened 2026-01-06
Split out AI playground CSP header

serprex:
Previous hard coded *.ubicloud.com string was undesirable,
this is only needed by AI playground, so scope header to that route

### DIFF
--- a/routes/project/inference_playground.rb
+++ b/routes/project/inference_playground.rb
@@ -3,6 +3,8 @@
 class Clover
   hash_branch(:project_prefix, "inference-playground") do |r|
     r.get web? do
+      content_security_policy.add_connect_src "https://*.#{Config.inference_dns_zone}"
+
       DB.ignore_duplicate_queries do
         @inference_models = [inference_router_model_ds.eager(inference_router: {load_balancer: :private_subnet}), inference_endpoint_ds.eager(:location, load_balancer: :private_subnet)].flat_map do |ds|
           ds.where(Sequel.pg_jsonb_op(:tags).get_text("capability") => "Text Generation").all
--- a/clover.rb
+++ b/clover.rb
@@ -173,7 +173,7 @@ class Clover < Roda
     csp.form_action :self, "https://checkout.stripe.com", "https://github.com/login/oauth/authorize", "https://accounts.google.com/o/oauth2/auth"
     csp.script_src :self, "https://cdn.jsdelivr.net/npm/jquery@3.7.0/dist/jquery.min.js", "https://cdn.jsdelivr.net/npm/dompurify@3.0.5/dist/purify.min.js", "https://cdn.jsdelivr.net/npm/flatpickr@4.6.13/dist/flatpickr.min.js", "https://challenges.cloudflare.com/turnstile/v0/api.js", "https://cdn.jsdelivr.net/npm/marked@15.0.5/marked.min.js", "https://cdn.jsdelivr.net/npm/echarts@5.6.0/dist/echarts.min.js"
     csp.frame_src :self, "https://challenges.cloudflare.com"
-    csp.connect_src :self, "https://*.ubicloud.com"
+    csp.connect_src :self
     csp.base_uri :none
     csp.frame_ancestors :none
   end


enescakir:
Why do we need to change default to `*`?

serprex:
Don't need to change default, but there's 1 environment that wants `*.ubicloud.com` & the rest don't. In particular I'm trying to keep local dev setup as simple as possible, & not having random CSP errors in local dev web console would be nice

enescakir:
We try to keep production secure by default, even if an engineer forgets to set a configuration value. In this case, if the operator forgets to set `Config.csp_connect_src`, production becomes insecure because the default value is "*".

If this is mainly a development inconvenience, one option could be `override :csp_connect_src, (production? ? "https://*.ubicloud.com" : "*"), string`.

Another option is to set a default value via the `overwrite_envrb` rake task. That wouldn't change existing environments, but new ones would get `*` default.

serprex:
Updated changed default to not apply to production

jeremyevans (approved):
This change seems reasonable to me.

I looked into the reason for supporting domains in addition to `self`, and apparently this was added for AI playground support in cb90bb8412906ab450308eeb1d7526fe44daae73 . I'm not sure whether it is used for anything else.  It would be best to set it to `self` by default, and only add other connect sources to the branches/routes that actually need them.

@serprex can you describe what you were doing when your browser was complaining about this, since I'm guessing it wasn't for the AI playground?

serprex:
I was just on project view, I hadn't built css so page looked bad, checked web console which showed error:
```
Connecting to 'https://cdn.jsdelivr.net/npm/dompurify@3.0.5/dist/purify.min.js.map' violates the following Content Security Policy directive: "connect-src 'self' https://*.ubicloud.com". The request has been blocked.
```
(fixing this didn't fix page, I needed to run `npm run prod`, but one less warning is always nice)

jeremyevans:
@serprex, thank you for that information. I think we should change the CSP so connect src is only `self` by default, and other values are only added on pages that require it.  We should not have it use `*` ever.  The warning if you haven't generated the CSS is a non-issue IMO, not something we should modify the CSP to avoid.

serprex:
route specific CSP looks nice, but I thought ubicloud would configure their prod with *.ubicloud.com on config (where now incorrect config doesn't cause silent security flaw)

jeremyevans:
> route specific CSP looks nice, but I thought ubicloud would configure their prod with *.ubicloud.com on config (where now incorrect config doesn't cause silent security flaw)

I'm not opposed to a config variable for the AI inference playground CSP connect_src. I'll leave that up to @bsatzger.

bsatzger:
> I like this new approach, but at least the AI playground needs to be updated to add `https://*.ubicloud.com` to connect_src, otherwise I would guess this breaks it.
> 
> Maybe:
> 
> ```diff
> diff --git a/routes/project/inference_playground.rb b/routes/project/inference_playground.rb
> index 0a6403e87..717dd9e83 100644
> --- a/routes/project/inference_playground.rb
> +++ b/routes/project/inference_playground.rb
> @@ -3,6 +3,8 @@
>  class Clover
>    hash_branch(:project_prefix, "inference-playground") do |r|
>      r.get web? do
> +      content_security_policy.add_connect_src "https://*.ubicloud.com"
> +
>        DB.ignore_duplicate_queries do
>          @inference_models = [inference_router_model_ds.eager(inference_router: {load_balancer: :private_subnet}), inference_endpoint_ds.eager(:location, load_balancer: :private_subnet)].flat_map do |ds|
>            ds.where(Sequel.pg_jsonb_op(:tags).get_text("capability") => "Text Generation").all
> ```
> 
> @bsatzger can you test this PR with this diff and let me know whether the AI playground still works? Do you know of any other places where connect_src needs values beyond `self`?

I tested this by removing `https://*.ubicloud.com` from csp.connect_src in clover.rb, which breaks AI playground (as expected). Adding the line above to inference_playground.rb lets the AI playground work properly.

So I'm good with removing `https://*.ubicloud.com` from clover.rb and adding `content_security_policy.add_connect_src "https://*.ubicloud.com"` to inference_playground.rb. This could be either hardcoded or a config variable.

serprex:
should csp be `ai.ubicloud.com` instead of `*.ubicloud.com`? then able to use `Config.inference_dns_zone`

routes/project/inference_playground.rb
```diff
@@ -3,6 +3,8 @@
 class Clover
   hash_branch(:project_prefix, "inference-playground") do |r|
     r.get web? do
+      content_security_policy.add_connect_src Config.inference_dns_zone
```
jeremyevans: This seems wrong. For one, I think there are separate subdomains used, and using just `ai.ubicloud.com` would not allow subdomains. Second, this doesn't enforce https. I think it would at least need to be:

```suggestion
      content_security_policy.add_connect_src "https://*.#{Config.inference_dns_zone}"
```

I'm also not sure if `ai.ubicloud.com` is used directly (seems unlikely, but I don't know). If so, we would need:

```suggestion
      content_security_policy.add_connect_src "https://#{Config.inference_dns_zone}", "https://*.#{Config.inference_dns_zone}"
```

@bsatzger which one of these would be more appropriate?

bsatzger: Our naming schema is currently https://{location}.ai.ubicloud.com, example: `https://us-east-a2.ai.ubicloud.com`. So `"https://*.#{Config.inference_dns_zone}"` seems most appropriate.

PR #4496, opened 2026-01-06
Include created_at field in postgres list API

serprex:
Useful for general observability,
and allows a view with a list of databases sorted by creation time.

### DIFF
--- a/spec/serializers/postgres_spec.rb
+++ b/spec/serializers/postgres_spec.rb
@@ -96,4 +96,11 @@ RSpec.describe Serializers::Postgres do
     expect(data).not_to have_key(:earliest_restore_time)
     expect(data).not_to have_key(:latest_restore_time)
   end
+
+  it "includes created_at in detailed serialization" do
+    create_representative_server(primary: true)
+    data = described_class.serialize(pg)
+    expect(data).to have_key(:created_at)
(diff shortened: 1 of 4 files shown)


furkansahin:
Just reformat the commit message

serprex:
I don't follow, reading over commit message guidelines, mood/length/capitalization is all there, no prefix, body covers purpose

spec/serializers/postgres_spec.rb
```diff
@@ -96,4 +96,11 @@ def create_representative_server(primary: true)
     expect(data).not_to have_key(:earliest_restore_time)
     expect(data).not_to have_key(:latest_restore_time)
   end
+
+  it "includes created_at in detailed serialization" do
+    create_representative_server(primary: true)
+    data = described_class.serialize(pg, {detailed: true})
```
jeremyevans: This PR has it included even in the non-detailed view. So this spec should be updated to reflect that.

jeremyevans (approved):
This should be added to the CLI `pg show` output, though that could be done after it is merged.

furkansahin:
>I don't follow, reading over commit message guidelines, mood/length/capitalization is all there, no prefix, body covers purpose

The first line was cut early. I pushed the new commit.

PR #4497, opened 2026-01-06
Remove unnecessary apt-get update calls

furkansahin:
sudo apt-get update is necessary when we are setting up a new host since the servers do not come with the most up-to date packages, always. However, for any other product where we install rhizome on a VM, it is unnecessary, because the VM images are supposed to come with an up-to-date package set. Therefore, we are making the command run conditional in the BootstrapRhizome prog.

fixes https://linear.app/ubicloud/issue/UBI-262/remove-sudo-apt-get-update-from-rhizome-setup

### DIFF
--- a/prog/bootstrap_rhizome.rb
+++ b/prog/bootstrap_rhizome.rb
@@ -55,10 +55,13 @@ LOGIND
     pop "rhizome user bootstrapped and source installed" if retval&.dig("msg") == "installed rhizome"
 
     key_data = sshable.keys.map(&:private_key)
-    Util.rootish_ssh(sshable.host, user, key_data, <<SH, LOGIND_CONFIG:, SSHD_CONFIG:, public_keys: sshable.keys.map(&:public_key).join("\n"))
+    no_bundler_install = (strand.stack.first["no_bundler_install"] == true).to_s
+    Util.rootish_ssh(sshable.host, user, key_data, <<SH, LOGIND_CONFIG:, SSHD_CONFIG:, public_keys: sshable.keys.map(&:public_key).join("\n"), no_bundler_install:)
 set -ueo pipefail
-sudo apt-get update
-sudo apt-get -y install ruby-bundler
+if [ :no_bundler_install = false ]; then
+  sudo apt-get update
+  sudo apt-get -y install ruby-bundler
+fi
 sudo which bundle
 sudo userdel -rf rhizome || true
 sudo adduser --disabled-password --gecos '' rhizome
--- a/spec/prog/bootstrap_rhizome_spec.rb
+++ b/spec/prog/bootstrap_rhizome_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe Prog::BootstrapRhizome do
     )
   }
   let(:st) {
-    Strand.create_with_id(sshable, prog: "BootstrapRhizome", label: "start")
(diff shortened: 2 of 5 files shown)


spec/prog/bootstrap_rhizome_spec.rb
```diff
@@ -47,7 +47,7 @@
       expect(Net::SSH).to receive(:start).and_yield(session)
       expect(session).to receive(:_exec!).with(<<'FIXTURE').and_return(Net::SSH::Connection::Session::StringWithExitstatus.new("", 0))
 set -ueo pipefail
-sudo apt-get update
+sudo\ apt-get\ update
```
serprex: doesn't this mean your command errors with
```
bash: sudo apt-get update: command not found
```
since it's looking to execute `/bin/sudo apt-get update`

furkansahin: That's correct, I updated it

iamKunalGupta:
> I'm not sure whether VMs boot images should be assumed to always be up to date. Maybe it is a valid assumption, maybe it isn't, but I don't have enough knowledge to be sure. But if that is the assumption, this seems fine.

Another reason for doing this is because of the way VM images are built currently, packages are downloaded at build time, but installed at Instance creation time, so any apt-update can invalidate the package cache.
Additionally, there might be private apt repos which are accessible only during build time, but are not available during instance creation time, so apt-update (and consequently install) will fail for them

spec/prog/bootstrap_rhizome_spec.rb
```diff
@@ -13,7 +13,7 @@
     )
   }
   let(:st) {
-    Strand.create_with_id(sshable, prog: "BootstrapRhizome", label: "start")
+    Strand.create_with_id(sshable, prog: "BootstrapRhizome", stack: [{"target_folder" => "host"}], label: "start")
```
jeremyevans, line 16: Is this change needed now that we don't check `target_folder`, but check `no_apt_update`?

prog/bootstrap_rhizome.rb
```diff
@@ -61,9 +61,12 @@ def before_run
     pop "rhizome user bootstrapped and source installed" if retval&.dig("msg") == "installed rhizome"
 
     key_data = sshable.keys.map(&:private_key)
-    Util.rootish_ssh(sshable.host, user, key_data, <<SH, LOGIND_CONFIG:, SSHD_CONFIG:, public_keys: sshable.keys.map(&:public_key).join("\n"))
+    no_apt_update = (strand.stack.first["no_apt_update"] == "true").to_s
```
enescakir: Since it’s jsonb, it already supports booleans. Is there any reason to use a string instead?

```ruby
[10] clover-development(main)> st.update(stack: [{"boolean_true" => true, "boolean_false" => false }])

[11] clover-development(main)> st.stack
=> [{"boolean_true" => true, "boolean_false" => false}]

[12] clover-development(main)> st.reload.stack
=> [{"boolean_true" => true, "boolean_false" => false}]
```

furkansahin:
I switched this to draft because the removal of apt-get update before the ruby-bundler installation started failing postgres provisioning. We have to install ruby-bundler in the image before starting to not run apt-get update. I am working on our images and will get back to this after that

PR #4498, opened 2026-01-06
Add uuid type to config for earlier validation

serprex:


config.rb
```diff
@@ -122,17 +122,17 @@ def self.unfrozen_test?
   optional :github_cache_blob_storage_region, string
   optional :github_cache_blob_storage_access_key, string, clear: true
   optional :github_cache_blob_storage_secret_key, string, clear: true
-  optional :github_cache_blob_storage_account_id, string
+  optional :github_cache_blob_storage_account_id, uuid
```
enescakir: This one is Cloudflare account id which is not a UUID.

spec/config_spec.rb
```diff
@@ -16,4 +16,27 @@
 
     expect(described_class.test_array).to be_nil
   end
+
+  it "uuid accepts valid uuid" do
+    id = SecureRandom.uuid
+    described_class.class_eval %(
+      override :test_valid_uuid, "#{id}", uuid
+      ), __FILE__, __LINE__ - 2
```
jeremyevans: Better to use block-eval than string-eval for cases where block-eval works:

```suggestion
    described_class.class_eval do
      override :test_valid_uuid, id, uuid
    end
```

Similar for the following two specs.

lib/casting_config_helpers.rb
```diff
@@ -46,6 +46,16 @@ def string
     nil
   end
 
+  def uuid
+    ->(v) do
+      if v.nil? || /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/ =~ v
```
jeremyevans: Simplify using `\h`. Never use `^` and `$` in Ruby regexps unless you are 100% sure you want to match start/end of line inside string instead of start/end of string. Use `match?` instead of `=~` if you only need to test for a match and won't be accessing `MatchData`/`$1`/etc.

```suggestion
      if v.nil? || /\A\h{8}-\h{4}-\h{4}-\h{4}-\h{12}\z/.match?(v)
```

spec/config_spec.rb
serprex: seems like all tests in this file should be updated like that

jeremyevans: Sorry about that. I didn't realize from the context on the PR review page that this was an existing issue in these specs. Thank you for fixing it.

jeremyevans (approved):

PR #4504, opened 2026-01-07
Demock update_firewall_rules_spec

fdr:
Replace instance_double mocks with real persisted Vm, Firewall, Nic, LoadBalancer, and other models, letting Sequel handle all relationships.

Ref: clover-mol-cvd5

### DIFF
--- a/spec/prog/vnet/metal/update_firewall_rules_spec.rb
+++ b/spec/prog/vnet/metal/update_firewall_rules_spec.rb
@@ -2,51 +2,97 @@
 
 RSpec.describe Prog::Vnet::Metal::UpdateFirewallRules do
   subject(:nx) {
-    described_class.new(st)
+    described_class.new(@st)
   }
 
-  let(:st) { Strand.new }
+  let(:project) { Project.create(name: "test-project") }
+  let(:location_id) { Location::HETZNER_FSN1_ID }
+  let(:vmh) { create_vm_host }
   let(:ps) {
-    instance_double(PrivateSubnet)
+    PrivateSubnet.create(
+      name: "ps", location_id:,
+      net6: "fd10:9b0b:6b4b:8fbb::/64",
+      net4: "10.0.0.0/26",
+      state: "waiting",
+      project:
+    )
+  }
+  let(:firewall) {
+    Firewall.create(name: "test-fw", location_id:, project:).tap do |fw|
+      fw.add_private_subnet(ps)
+    end
+  }
+  let(:nic) {
+    Nic.create(
+      private_subnet_id: ps.id,
+      private_ipv6: NetAddr::IPv6Net.parse("fd00::1/128"),
+      private_ipv4: NetAddr::IPv4Net.parse("10.0.0.1/32"),
+      mac: "00:00:00:00:00:01",
+      name: "nic0",
+      state: "active"
+    )
   }
   let(:vm) {
-    vmh = instance_double(VmHost, sshable: Sshable.new)
-    nic = instance_double(Nic, private_ipv4: NetAddr::IPv4Net.parse("10.0.0.0/32"), private_ipv6: NetAddr::IPv6Net.parse("fd00::1/128"), ubid_to_tap_name: "tap0")
-    ephemeral_net6 = NetAddr::IPv6Net.parse("fd00::1/79")
-    instance_double(Vm, project: instance_double(Project, get_ff_ipv6_disabled: false), private_subnets: [ps], vm_host: vmh, inhost_name: "x", nics: [nic], ephemeral_net6:, load_balancer: nil, private_ipv4: NetAddr::IPv4Net.parse("10.0.0.0/32").network, location: Location[Location::HETZNER_FSN1_ID])
+    firewall
+    Vm.create(
+      vm_host: vmh,
+      name: "test-vm",
+      family: "standard",
+      cores: 1,
+      vcpus: 2,
+      cpu_percent_limit: 200,
+      cpu_burst_percent_limit: 0,
+      memory_gib: 4,
+      arch: "x64",
+      boot_image: "ubuntu-jammy",
+      location_id:,
+      project_id: project.id,
+      display_state: "running",
+      ip4_enabled: false,
+      unix_user: "ubi",
+      public_key: "ssh-ed25519 test",
+      ephemeral_net6: NetAddr::IPv6Net.parse("fd00::1/79")
+    ).tap do |v|
+      nic.update(vm_id: v.id)
+    end
   }
+  let(:sshable) { nx.vm.vm_host.sshable }
+
+  before do
+    @st = Strand.create_with_id(vm, prog: "Vnet::Metal::UpdateFirewallRules", label: "update_firewall_rules")
+  end
 
   describe "#before_run" do
     it "pops if vm is to be destroyed" do
-      expect(nx).to receive(:vm).and_return(vm).at_least(:once)
-      expect(vm).to receive(:destroy_set?).and_return(true)
+      vm.incr_destroy
       expect { nx.before_run }.to exit({"msg" => "firewall rule is added"})
     end
 
     it "does not pop if vm is not to be destroyed" do
-      expect(nx).to receive(:vm).and_return(vm).at_least(:once)
-      expect(vm).to receive(:destroy_set?).and_return(false)
       expect { nx.before_run }.not_to exit
     end
   end
 
   describe "update_firewall_rules" do
+    def create_firewall_rules
+      firewall.replace_firewall_rules([
+        {cidr: "0.0.0.0/0", port_range: nil},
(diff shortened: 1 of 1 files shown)


spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -5,473 +5,216 @@
...
+      arch: "x64",
+      boot_image: "ubuntu-jammy",
+      location_id:,
+      project_id: project.id,
+      display_state: "running",
+      ip4_enabled: false,
+      unix_user: "ubi",
+      public_key: "ssh-ed25519 test",
+      ephemeral_net6: NetAddr::IPv6Net.parse("fd00::1/79"),
+      created_at: Time.now
```
jeremyevans: Don't need to set `created_at` explicitly:

```suggestion
      ephemeral_net6: NetAddr::IPv6Net.parse("fd00::1/79")
```

Similar issue later in specs.

spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -5,473 +5,216 @@
...
+    ).tap do |v|
+      nic.update(vm_id: v.id)
+    end
+  }
+  let!(:st) {
```
jeremyevans: In general, `let!` should be avoided. If you really want `let!` behavior, switch to using `before` to make it explicit. Otherwise, use `let`.

spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -5,473 +5,216 @@
...
+    Strand.create_with_id(vm, prog: "Vnet::Metal::UpdateFirewallRules", label: "update_firewall_rules")
   }
 
+  def sshable
```
jeremyevans: Not a problem, but kinda odd that this uses a method and not `let`.

spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -5,473 +5,216 @@
...
+        {cidr: "10.10.10.0/26", port_range: Sequel.pg_range(80..10000)},
+        {cidr: "123.123.123.64/27", port_range: Sequel.pg_range(8080..12000)},
+        {cidr: "123.123.123.64/26", port_range: Sequel.pg_range(9000..16000)},
+        {cidr: "::/0", port_range: nil},
+        {cidr: "fd00::1/128", port_range: Sequel.pg_range(8080..65535)},
+        {cidr: "fd00::1/64", port_range: Sequel.pg_range(0..8081)},
+        {cidr: "fd00::2/64", port_range: Sequel.pg_range(80..10000)}
+      ].each do |rule|
+        FirewallRule.create(firewall_id: firewall.id, cidr: rule[:cidr], port_range: rule[:port_range])
+      end
```
jeremyevans: Probably should switch to using `Firewall#replace_firewall_rules`

spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -5,473 +5,216 @@
...
-    ip6 saddr ::/0 icmpv6 type echo-reply counter accept
-    ip6 daddr ::/0 icmpv6 type echo-reply counter accept
-
-    # Allow load balancer traffic
-
-  }
-}
-ADD_RULES
+      create_firewall_rules
+      expect(sshable).to receive(:_cmd) do |cmd, stdin:|
```
jeremyevans: I prefer the previous approach of checking the entire stdin string here, at least for one of the specs (maybe not for all of them due to the duplication).

spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -5,473 +5,216 @@
...
+        expect(stdin).to include("set private_ipv4_cidrs")
+        expect(stdin).to include("10.0.0.1/32")
+        expect(stdin).to include("set private_ipv6_cidrs")
+        expect(stdin).to include("fd00::1/128")
+        expect(stdin).to include("set globally_blocked_ipv4s")
+        expect(stdin).to include("123.123.123.123/32")
+        expect(stdin).to include("set globally_blocked_ipv6s")
+        expect(stdin).to include("2a00:1450:400e:811::200e/128")
+        expect(stdin).to include("chain forward_ingress")
+        expect(stdin).to include("# An nftables idiom for idempotent re-create")
```
jeremyevans: For cases like this, instead of checking for comments, I think it's better to check for commands.

fdr: I'm a bit puzzled about what you mean by this. the repeated includes are very silly, so thanks for pointing that out, but the whole thing is this is one atomic rule change, that happens to have a few comment strings embedded, not multiple commands.

jeremyevans: I'm just saying I don't think it's necessary to check the output for comment strings. It would be better to check the output for the commands being executed. Currently, it looks to be checking for fragments of commands and comments. However, this isn't a big deal.

fdr: it is kinda necessary because this is stdin fed into the `nft` command, that is, closer to asserting a configuration file's content.

fdr: like, it's not really a command-response language, it's a configuration language, often but not always stored in files, in this case, we happen to not store it in a file and instead use `-` for `/dev/stdin`.

fdr:
looking over these ginormous literal stdins, I think we probably want something like a long *sub*sequence literal that concentrates on relevant part.

fdr:
eh, I think I'm going to bring back at least as many literals as there were previously to not have to tarry too long with what substrings keep the test's power.

fdr:
@jeremyevans this should clean up the diff immensely. notably some changes are in the literal from the democking that make it more realistic...or perhaps document a bug

spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -607,5 +672,18 @@
...
+      firewall.replace_firewall_rules([{cidr: "192.168.1.0/24", port_range: Sequel.pg_range(443..443)}])
+
+      expect(sshable).to receive(:_cmd) do |cmd, stdin:|
+        expect(cmd).to eq("sudo ip netns exec #{vm.inhost_name} nft --file -")
+        expect(stdin).to include("192.168.1.0/24 . 443")
+        expect(stdin).not_to include("192.168.1.0/24 . 443-")
+      end
+
+      expect { nx.update_firewall_rules }.to exit({"msg" => "firewall rule is added"})
+    end
```
fdr: this test is interesting, in that it's net-new, I wonder if it was introduced to make branch coverage work

jeremyevans: What it's testing for seems to be covered in "populates load balancer destination sets and adds related rules" spec. No objection to keeping this, though.

spec/prog/vnet/metal/update_firewall_rules_spec.rb
```diff
@@ -56,7 +102,7 @@
   set allowed_ipv4_port_tuple {
     type ipv4_addr . inet_service;
     flags interval;
-elements = {1.1.1.1/32 . 22,10.10.10.0/26 . 80-9999,123.123.123.64/26 . 9000-15999,123.123.123.64/27 . 8080-8999}
+elements = {1.1.1.1/32 . 22-23,10.10.10.0/26 . 80-10000,123.123.123.64/26 . 9000-16000,123.123.123.64/27 . 8080-8999}
```
jeremyevans: I think the reasoning behind these expected results changing is because PostgreSQL tries to normalize integer ranges from inclusive end to exclusive end, and the code expects to only deal with exclusive ranges, so it subtracts 1 from the end without actually checking whether the range was exclusive or not. Previously, with the `instance_double` approach, inclusive ranges were used because it didn't go through a PostgreSQL normalization, and apparently nobody checked the output carefully to realize it didn't actually reflect the ranges.

jeremyevans (approved):

fdr:
@jeremyevans I cleaned this up, now the port ranges do the right thing and I dropped  the extra test. in the end this reduced the churn quite a bit. I also rebased and had to de-objectify a bunch of correlated objects (to use the `id` form) to get it to pass, mostly from fcab762b3. I think that's fine....

fdr:
```
fdr@albatross ~/c/clover ((2b71357d))> git diff 84bca7a95 2b71357d6
<empty>
```

fdr:
ditto:

```
 git diff 2b71357d6 5f446c739
fdr@albatross ~/c/clover ((5f446c73))> 
```

PR #4505, opened 2026-01-07
Add admin page to find VMs by IPv4 address

enescakir:
Investigating abuse reports requires finding VMs associated with specific IP addresses. Previously, this required using pry for active VMs and direct database queries for archived ones.

This commit adds a custom admin page at /custom/vm_by_ipv4 that searches both active VMs and archived records from the last 15 days in a single query. The page accepts comma-separated IP addresses and displays matching VMs with their metadata including project, boot image, and archive timestamp.

The implementation also introduces a generic custom views system that automatically discovers ERB templates in views/admin/customs/.

CleanShot 2026-01-07 at 14 19 19@2x

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -516,6 +516,44 @@ class CloverAdmin < Roda
       autoforme
     end
 
+    r.get "vm-by-ipv4" do
+      if (@ips_param = typecast_params.nonempty_str("ips"))
+        ips = @ips_param.split(",").filter_map {
+          begin
+            NetAddr.parse_net(it.strip).to_s
+          rescue
+            nil
+          end
+        }
+
+        active_vms = Vm.from_ips(ips).map {
+          {
+            ip: it.assigned_vm_address.ip.to_s,
+            created_at: it.created_at,
+            archived_at: nil,
+            vm_id: it.ubid,
+            vm_name: it.name,
+            boot_image: it.boot_image,
+            project_id: it.project.ubid
+          }
+        }
+        archived_vms = ArchivedRecord.vms_by_ips(ips).map {
+          {
+            ip: it[:ip],
+            created_at: it[:created_at],
+            archived_at: it[:archived_at],
+            vm_id: UBID.from_uuidish(it[:vm_id]).to_s,
+            vm_name: it[:vm_name],
+            boot_image: it[:boot_image],
+            project_id: UBID.from_uuidish(it[:project_id]).to_s
+          }
+        }
+        @vms = (active_vms + archived_vms).sort_by { [it[:ip], -it[:created_at].to_i] }
+      end
+
+      view("vm_by_ipv4")
+    end
+
     r.root do
       if (ubid = typecast_params.ubid("id")) && (klass = UBID.class_for_ubid(ubid))
         r.redirect("/model/#{klass.name}/#{ubid}")
--- a/views/admin/components/table.erb
+++ b/views/admin/components/table.erb
@@ -1,0 +1,25 @@
+<%# locals: (data:, title: nil) %>
+
+<% if data.empty? %>
+  <p>No data available</p>
+<% else %>
+  <table>
+    <caption><%= title %></caption>
+    <thead>
+      <tr>
+        <% data.first.each_key do |k| %>
+          <th><%= k %></th>
+        <% end %>
+      </tr>
+    </thead>
+    <tbody>
+      <% data.each do |row| %>
+        <tr>
+          <% row.each_value do |v| %>
+            <td><%== linkify_ubids(v.to_s) %></td>
+          <% end %>
+        </tr>
+      <% end %>
+    </tbody>
+  </table>
+<% end %>
--- a/views/admin/vm_by_ipv4.erb
+++ b/views/admin/vm_by_ipv4.erb
@@ -1,0 +1,15 @@
+<% @page_title = "Find VM by IPv4 Address" %>
+
+<p>
+  This page allows you to find virtual machines associated with given IPv4 addresses, including archived ones from the
+  last 15 days. It's especially useful when investigating abuse reports.
+</p>
+
+<% form(id: "vm_by_ipv4_form") do |f| %>
+  <%== f.input("textarea", key: "ips", placeholder: "Comma seperated IP list", value: @ips_param) %>
+  <%== f.button("Show Virtual Machines") %>
+<% end %>
+
+<% if @vms %>
+  <%== part("components/table", data: @vms) %>
+<% end %>
--- a/model/archived_record.rb
+++ b/model/archived_record.rb
@@ -4,6 +4,30 @@ require_relative "../model"
 
 class ArchivedRecord < Sequel::Model
   no_primary_key
+
+  def self.vms_by_ips(ips)
+    ip_values = Sequel.pg_jsonb_op(Sequel[:ip][:model_values])
+    vm_values = Sequel.pg_jsonb_op(Sequel[:vm][:model_values])
+    last_15_days = Sequel::CURRENT_TIMESTAMP - Sequel.cast("15 days", :interval)
+    DB.from(Sequel[:archived_record].as(:ip))
+      .join(Sequel[:archived_record].as(:vm), ip_values.get_text("dst_vm_id") => vm_values.get_text("id"))
+      .where(Sequel[:ip][:model_name] => "AssignedVmAddress")
+      .where(Sequel[:vm][:model_name] => "Vm")
+      .where(ip_values.get_text("ip") => ips)
+      .where(Sequel[:ip][:archived_at] > last_15_days)
+      .where(Sequel[:vm][:archived_at] > last_15_days)
+      .select(
+        ip_values.get_text("ip").as(:ip),
+        Sequel[:ip][:archived_at],
+        Sequel.cast(vm_values.get_text("created_at"), :timestamptz).as(:created_at),
+        ip_values.get_text("dst_vm_id").as(:vm_id),
+        vm_values.get_text("name").as(:vm_name),
+        vm_values.get_text("boot_image").as(:boot_image),
+        vm_values.get_text("project_id").as(:project_id)
+      )
+      .reverse(Sequel[:ip][:archived_at])
+      .all
+  end
 end
 
 # Table: archived_record
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -210,6 +210,10 @@ class Vm < Sequel::Model
     id.to_s[0..7]
   end
 
+  def self.from_ips(ips)
+    eager_graph(:assigned_vm_address, :project).where(Sequel[:assigned_vm_address][:ip] => ips).all
+  end
+
   def inhost_name
     self.class.ubid_to_name(UBID.from_uuidish(id))
   end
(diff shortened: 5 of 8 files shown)


enescakir:
I'm planning to add more custom pages to help with oncall duties.

clover_admin.rb
```diff
@@ -516,6 +521,11 @@ def self.object_action(...)
       autoforme
     end
 
+    r.get "custom", String do |custom_name|
+      next unless CUSTOM_VIEWS[custom_name]
+      view("customs/#{custom_name}")
+    end
```
jeremyevans: I don't think we benefit from indirection through CUSTOM_VIEWS.

```suggestion
    r.get "vm-by-ipv4" do
      # Logic to retrieve vms by calling model method
      view("vm_by_ipv4")
    end
```

views/admin/customs/vm_by_ipv4.erb
```diff
@@ -0,0 +1,64 @@
...
+</p>
+
+<% ips_param = request.params["ips"] %>
+
+<% form(id: "vm_by_ipv4_form") do |f| %>
+  <%== f.input("textarea", key: "ips", placeholder: "Comma seperated IP list", value: ips_param) %>
+  <%== f.button("Show Virtual Machines") %>
+<% end %>
+
+<% if ips_param && !ips_param.empty? && (ips = ips_param.split(",").filter_map { (IPAddr.new(it.strip) rescue nil)&.to_s&.concat("/32") }).any? %>
```
jeremyevans: I think we prefer to use NetAddr instead of IPAddr.

views/admin/customs/vm_by_ipv4.erb
```diff
@@ -0,0 +1,64 @@
...
+  <% active_vms = AssignedVmAddress.where(ip: ips).eager(vm: [:project]).all.map {
```
jeremyevans: This logic should be moved to a model method, and the method should be called from the routing tree.

views/admin/customs/vm_by_ipv4.erb
```diff
@@ -0,0 +1,64 @@
+<% @page_title = "Find VM by IPv4 Address" %>
+
+<p>
+  This page allows you to find virtual machines associated with given IPv4 addresses, including archived ones from the
+  last 15 days. It's especially useful when investigating abuse reports.
+</p>
+
+<% ips_param = request.params["ips"] %>
```
jeremyevans: Param access should always be through `typecast_params`, probably using `nonempty_str` in this case.

views/admin/components/table.erb
```diff
@@ -0,0 +1,25 @@
+<%# locals: (data:, title: nil) %>
+
+<% if data.empty? %>
+  <p>No data available</p>
+<% else %>
+  <table>
+    <caption><%= title %></caption>
+    <thead>
+      <tr>
+        <% data.first.keys.each do |k| %>
```
jeremyevans:
```suggestion
        <% data.first.each_key do |k| %>
```

views/admin/components/table.erb
```diff
@@ -0,0 +1,25 @@
...
+          <th><%= k %></th>
+        <% end %>
+      </tr>
+    </thead>
+    <tbody>
+      <% data.each do |row| %>
+        <tr>
+          <% row.values.each do |v| %>
```
jeremyevans:
```suggestion
          <% row.each_value do |v| %>
```

clover_admin.rb
enescakir: Do you think that when we have multiple custom views, we should add them to the routing tree?

views/admin/customs/vm_by_ipv4.erb
enescakir: I couldn't run `typecast_params` in the view. But if we move them to routing tree, I can.

clover_admin.rb
enescakir: I changed it

clover_admin.rb
```diff
@@ -516,6 +516,67 @@ def self.object_action(...)
...
+            {
+              ip: it[:ip],
+              created_at: it[:created_at],
+              archived_at: it[:archived_at],
+              vm_id: UBID.from_uuidish(it[:vm_id]).to_s,
+              vm_name: it[:vm_name],
+              boot_image: it[:boot_image],
+              project_id: UBID.from_uuidish(it[:project_id]).to_s
+            }
+          }
```
jeremyevans: This logic should be moved to a model method.

clover_admin.rb
```diff
@@ -516,6 +516,67 @@ def self.object_action(...)
...
+            ip_values.get_text("ip").as(:ip),
+            Sequel[:ip][:archived_at],
+            vm_values.get_text("created_at").as(:created_at),
+            ip_values.get_text("dst_vm_id").as(:vm_id),
+            vm_values.get_text("name").as(:vm_name),
+            vm_values.get_text("boot_image").as(:boot_image),
+            vm_values.get_text("project_id").as(:project_id)
+          )
+          .order(Sequel.desc(Sequel[:ip][:archived_at]))
+          .all
```
jeremyevans: You can call `map` directly on the dataset.

```suggestion
```

views/admin/vm_by_ipv4.erb
jeremyevans, line 1: Probably best to move this to `views/admin/vm_by_ipv4.erb`.

helpers/model_hiding.rb
```diff
@@ -30,7 +30,7 @@ class ModelProxy < BasicObject
         ::PostgresResource => [:available_flavors, :default_flavor, :default_version, :ha_type_none, :maintenance_hour_options, :partner_notification_flavors],
         ::PostgresServer => [:victoria_metrics_client],
         ::SubjectTag => [:admin_tag?, :options_for_project, :subject_id_map_for_project_and_accounts],
-        ::Vm => [:from_runtime_jwt_payload]
+        ::Vm => [:from_runtime_jwt_payload, :from_ips]
```
jeremyevans: Is this actually needed? The method is called inside CloverAdmin, not Clover.

views/admin/components/table.erb
```diff
@@ -0,0 +1,25 @@
+<%# locals: (data:, title: nil) %>
+
+<% if (data || []).empty? %>
```
jeremyevans: Do we really want to support `nil` data values? I would guess we don't have a current need for it. 

```suggestion
<% if data.empty? %>
```

If we do want to support `nil` data values, no need to allocate an array in that case:

```suggestion
<% if !data || data.empty? %>
```

views/admin/customs/vm_by_ipv4.erb
```diff
@@ -0,0 +1,15 @@
...
+  This page allows you to find virtual machines associated with given IPv4 addresses, including archived ones from the
+  last 15 days. It's especially useful when investigating abuse reports.
+</p>
+
+<% form(id: "vm_by_ipv4_form") do |f| %>
+  <%== f.input("textarea", key: "ips", placeholder: "Comma seperated IP list", value: @ips_param) %>
+  <%== f.button("Show Virtual Machines") %>
+<% end %>
+
+<% if @ips_param %>
```
jeremyevans: I don't this this changes behavior, but I think it makes the code more clear:

```suggestion
<% if @vms %>
```

clover_admin.rb
```diff
@@ -516,6 +516,67 @@ def self.object_action(...)
...
+          }
+        }
+
+        ip_values = Sequel.pg_jsonb_op(Sequel[:ip][:model_values])
+        vm_values = Sequel.pg_jsonb_op(Sequel[:vm][:model_values])
+        last_15_days = Sequel::CURRENT_TIMESTAMP - Sequel.cast("15 days", :interval)
+        archived_vms = DB.from(Sequel[:archived_record].as(:ip))
+          .join(Sequel[:archived_record].as(:vm), ip_values.get_text("dst_vm_id") => vm_values.get_text("id"))
+          .where(Sequel[:ip][:model_name] => AssignedVmAddress.to_s)
+          .where(Sequel[:vm][:model_name] => Vm.to_s)
```
```suggestion
          .where(Sequel[:ip][:model_name] => "AssignedVmAddress")
          .where(Sequel[:vm][:model_name] => "Vm")
```

clover_admin.rb
```diff
@@ -516,6 +516,67 @@ def self.object_action(...)
...
+              created_at: it[:created_at],
+              archived_at: it[:archived_at],
+              vm_id: UBID.from_uuidish(it[:vm_id]).to_s,
+              vm_name: it[:vm_name],
+              boot_image: it[:boot_image],
+              project_id: UBID.from_uuidish(it[:project_id]).to_s
+            }
+          }
+
+        @vms = (active_vms + archived_vms).sort_by { [it[:ip], -it[:created_at].to_i] }
```
jeremyevans, line 551: Is the `to_i` actually what you want? For active VMs, `it[:created_at]` gives a Time object, but I'm guessing for archived records, it is a string (since JSON doesn't natively support time objects). You can probably cast the archived records created at to `timestamptz` when retrieiving it (or parse the string after retrieval and convert it to Time), and drop the `to_i` here.

```suggestion
        @vms = (active_vms + archived_vms).sort_by { [it[:ip], -it[:created_at]] }
```

clover_admin.rb
```diff
@@ -516,6 +516,67 @@ def self.object_action(...)
...
     end
 
+    r.get "vm-by-ipv4" do
+      if (@ips_param = typecast_params.nonempty_str("ips"))
+        ips = @ips_param.split(",").filter_map {
+          begin
+            NetAddr.parse_net(it.strip)
+          rescue
+            nil
+          end&.to_s
```
```suggestion
            NetAddr.parse_net(it.strip).to_s
          rescue
            ""
          end
```

clover_admin.rb
```diff
@@ -516,6 +516,67 @@ def self.object_action(...)
...
+          .select(
+            ip_values.get_text("ip").as(:ip),
+            Sequel[:ip][:archived_at],
+            vm_values.get_text("created_at").as(:created_at),
+            ip_values.get_text("dst_vm_id").as(:vm_id),
+            vm_values.get_text("name").as(:vm_name),
+            vm_values.get_text("boot_image").as(:boot_image),
+            vm_values.get_text("project_id").as(:project_id)
+          )
+          .order(Sequel.desc(Sequel[:ip][:archived_at]))
```
```suggestion
          .reverse(Sequel[:ip][:archived_at])
```

clover_admin.rb
jeremyevans: > Do you think that when we have multiple custom views, we should add them to the routing tree?

Yes. I'm guessing Claude picked this approach because it's similar to `OBJECTS_WITH_EXTRAS`, but the use case there is different. That needs to lookup using the model class name inside the page for that model class name, but that's not the case for `CUSTOM_VIEWS`.

clover_admin.rb
enescakir: Even if I cast to `timestamptz`, it doesn't allow me to prepend `-` to it. How can I reverse the order?

jeremyevans: I guess using `to_i` is fine then. As long as you cast to `timestamptz` so that all values are times (previously, I'm guessing you were comparing unix epoch time integers for active vms with year integers for archived vms).

enescakir: I casted it 👍

jeremyevans (approved):

PR #4512, opened 2026-01-07
Create bin/ssh to open interactive ssh session to vm

serprex:
Particularly useful for debugging system, this has proven consistently useful in past. Pry is not ideal for inspecting single system

### DIFF
--- a/bin/ssh
+++ b/bin/ssh
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S ruby --enable-frozen-string-literal
+#!/usr/bin/env ruby
 # frozen_string_literal: true
 
 require "tempfile"


bin/ssh
```diff
@@ -0,0 +1,30 @@
...
+when Vm
+    vm = model
+end
+
+unless vm
+    puts "could find vm for #{ARGV[0]}", model
+    exit 1
+end
+
+file = Tempfile.new("ubissh")
```
serprex, line 29: ideally there'd be no tempfile, but passing key in as /dev/input runs into issues for interactive session's stdin
ideally tempfile would be cleaned up, but that doesn't play nice with execing process into ssh instead of keeping ruby running on top

bin/ssh
```diff
@@ -0,0 +1,30 @@
...
+file.write(vm.sshable.keys.first.private_key)
+file.close
+exec "ssh", "-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null", "-i", file.path, "ubi@#{vm.ip4_string}"
```
enescakir:
```suggestion
exec "ssh", "-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null", "-i", file.path, "#{sshable.unix_user}@#{sshable.host}"
```

bin/ssh
```diff
@@ -0,0 +1,30 @@
...
+    vm = model.vm
+when Vm
+    vm = model
+end
+
+unless vm
+    puts "could find vm for #{ARGV[0]}", model
+    exit 1
+end
+
```
enescakir, line 23: It should fail if vm doesn't have a sshable. Customer vms don't.

bin/ssh
```diff
@@ -0,0 +1,30 @@
...
+end
+vm = nil
+case model
+when PostgresServer
+    vm = model.vm
+when GithubRunner
+    vm = model.vm
+when Vm
+    vm = model
+end
```
enescakir: nit
```suggestion
vm = case model
when PostgresServer, GithubRunner
    model.vm
when Vm
    model
end
```

bin/ssh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env -S ruby --enable-frozen-string-literal
```
jeremyevans: My guess is this was copied from `bin/pry`, but it shouldn't be needed here:

```suggestion
#!/usr/bin/env ruby
```

bin/ssh
jeremyevans: I'm wary of writing private key information to tempfiles. In general, we use touch-based hardware keys for access to production systems (https://www.ubicloud.com/blog/hardware-touch-stronger-ssh). This wouldn't do that, so it probably needs review from @fdr.

fdr: I can see how this feature might be handy, but I think we'll decline to use it for now. You can get past the temp file issue by starting an ssh-agent and ssh-add from stdin.

The bad things: you will log in as the user for rhizome which, in a really vexing and unintentional break from predecessors, is the same user used for staff log-in...except on vm_host, which does it more as intended, but, you don't use that...

Also doing it this way, you get no useful key fingerprint on the target identifying the staff doing the login, you'd have to correlate back to who requested the teletype for the `ssh` running remote somehow. Not un-doable, but it makes figuring out good faith access a lot more difficult from logs. I think for this reason, I would not apply this.

An upshot you might seize: a teletype recorder can be installed where `bin/ssh` is run from, presuming you wanted no other way to use `ssh`.

I think you could hold onto this one, having been made aware that you can avoid temp file via ssh-agent and that the openssh logs on the target will deteriorate for the use of a shared key, plus, you might find a collision with an eventual move to separate personnel & automated users.

fdr: tl;dr: keep it out of tree, and here's my Free Advice (e.g. key digest, advisory on users), the main thing it could have going for it is concentrating the teletype recording for some (you going to do that?)

serprex: makes sense, I raised this PR to
1. have a link people can get script from
2. query how this could be done better

it sounds like on your own end you have a method of staff access *(some list of SSH keys somewhere which gets put onto development AMI)* which we lack

ideally this can be done in a correct way that has comparable usability, because as is it's always going to break down to people passing some form of this script around & using it when they need eyes on a vm

PR #4517, opened 2026-01-07
Add more tables to admin panel

enescakir:
- **Use admin table component for existing tables**
  

- **Convert billing info Stripe data from raw JSON to table**
  The raw JSON format was difficult to read and inconsistent with the page
  layout. Displaying it as a table improves readability and visual
  alignment.
  
  Since some billing info records may not have Stripe data, the table
  component now supports nil values.
  

- **Display Stripe data for payment methods in the admin panel**
  

- **Remove unnecessary name alias in the admin panel**
  Since admin_label already falls back to ubid when the model doesn’t have
  a name, we can use admin_label directly without adding a separate alias
  for name.
  
  It's a more generic solution for the future models.
  

- **Add billing info table to admin panel**
  

- **Add payment method table to admin panel**

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -274,7 +274,7 @@ class CloverAdmin < Roda
       case column
       when :name, :ubid, :invoice_number
         link.call(obj, label: column)
-      when :project, :location, :vm_host
+      when :project, :location, :vm_host, :billing_info
         link.call(obj.send(column))
       when :subtotal, :cost
         "$%0.02f" % (obj.send(column) || 0)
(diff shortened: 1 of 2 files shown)


views/admin/components/table.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: (data:, table_class:, title: nil, use_admin_label: true) %>
 
-<% if data.empty? %>
+<% if !data || data.empty? %>
   <p>No data available</p>
```
jeremyevans: This looks weird, because you'll show "No data available" without any indication that it is for Stripe data. I think we should include the `title` before `No data available`. WDYT?

jeremyevans (approved):

views/admin/components/table.erb
enescakir: I made the title mandatory and added it to the no data message.

PR #4518, opened 2026-01-07
Isolate primary during unplanned failovers

shikharbhardwaj:
This commit introduces a mechanism to isolate unresponsive primary
postgres servers during unplanned failovers. This attempts to prevent a
split-brain scenario, where a primary becomes unresponsive triggering an
unplanned failover, but later returns back, leading to two primaries
co-existing.

For the lockout mechanism, try 3 different ways at once, and proceed if
any one of them succeeds or all fail:
1. pgctlcluster stop -m immediate.
2. add lockout pg_hba.conf and reload.
3. (Metal only) Bring down the vm virtual network interface.

These are somewhat redundant and might seem unnecessary, but we have
seen failure modes where the VM and network host access behaves
independently.

On AWS, we don't do anything to the EC2 instance yet. There are a few
possible approaches, but each has some downsides:
1. stop-instances --force: slow, as it waits for graceful shutdown
   first.
2. Updating security groups: fast but stateful, so existing connections
   remain active.
3. Network ACLs: fast and stateless, but more complicated and creates
   additional resources which will need cleanup.

### DIFF
--- a/prog/postgres/postgres_lockout.rb
+++ b/prog/postgres/postgres_lockout.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class Prog::Postgres::PostgresLockout < Prog::Base
+  subject_is :postgres_server
+
+  label def start
+    mechanism = strand.stack.first["mechanism"]
+
+    begin
+      send("lockout_with_#{mechanism}")
+      Clog.emit("Fenced unresponsive primary", {fenced_unresponsive_primary: {server_ubid: postgres_server.ubid, mechanism:}})
+      pop "lockout_succeeded"
+    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
+      pop "lockout_failed"
+    end
+  end
+
+  def lockout_with_pg_stop
+    postgres_server.vm.sshable.cmd(
+      "timeout 10 sudo pg_ctlcluster :version main stop -m immediate",
+      version: postgres_server.version,
+      timeout: 15
+    )
+  end
+
+  def lockout_with_hba
+    postgres_server.vm.sshable.cmd(
+      "timeout 10 sudo postgres/bin/lockout-hba :version",
+      version: postgres_server.version,
+      timeout: 15
+    )
+  end
+
+  def lockout_with_host_routing
+    postgres_server.vm.vm_host.sshable.cmd(
+      "timeout 10 sudo ip link set :interface down", interface: "vetho#{postgres_server.vm.inhost_name}",
+      timeout: 15
+    )
+  end
+end
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -469,6 +469,10 @@ SQL
       hop_fence
     end
 
+    when_lockout_set? do
+      hop_lockout
+    end
+
     when_unplanned_take_over_set? do
       register_deadline("wait", 5 * 60)
       hop_prepare_for_unplanned_take_over
@@ -556,6 +560,10 @@ SQL
   end
 
   label def unavailable
+    when_lockout_set? do
+      hop_lockout
+    end
+
     nap 0 if postgres_server.resource.ongoing_failover? || postgres_server.trigger_failover(mode: "unplanned")
 
     when_configure_set? do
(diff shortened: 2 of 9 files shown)


prog/postgres/postgres_lockout.rb
```diff
@@ -0,0 +1,59 @@
...
+    success = case mechanism
+    when "pg_stop"
+      lockout_with_pg_stop
+    when "hba"
+      lockout_with_hba
+    when "host_routing"
+      lockout_with_host_routing
+    else
+      false
+    end
```
enescakir: If you catch SSH errors here, no need to test them separately.

```suggestion
    success = case mechanism
    when "pg_stop", "hba", "host_routing"
      begin
        send("lockout_with_#{mechanism}")
        true
      rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
        false
      end
    else
      false
    end
```

prog/postgres/postgres_lockout.rb
```diff
@@ -0,0 +1,59 @@
...
+    pop success ? "lockout_succeeded" : "lockout_failed"
+  end
+
+  def lockout_with_pg_stop
+    postgres_server.vm.sshable.cmd(
+      "timeout 1 sudo pg_ctlcluster :version main stop -m immediate",
+      version: postgres_server.version,
+      timeout: 2
+    )
+    Clog.emit("Fenced unresponsive primary by stopping PostgreSQL", {ubid: postgres_server.ubid})
```
enescakir: If you convert logs to something like that, you can filter them using `failed_postgres_lockout:*` or `failed_postgres_lockout.mechanism:hba` at the log viewer.

```suggestion
    Clog.emit("Fenced unresponsive primary", {fenced_unresponsive_primary: {server_ubid: postgres_server.ubid, mechanism: "hba"}})
```

Clog naming practices: https://github.com/ubicloud/ubicloud/pull/632#issue-1902209682

prog/postgres/postgres_lockout.rb
byucesoy: We don't even need `case`/`when` block in this case.

prog/postgres/postgres_lockout.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    postgres_server.vm.vm_host.sshable.cmd(
+      "timeout 1 sudo ip link set :interface down", interface: "vetho#{postgres_server.vm.inhost_name}",
+      timeout: 2
+    )
+    Clog.emit("Fenced unresponsive primary by blocking host routing", {ubid: postgres_server.ubid})
+    true
+  rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
+    false
+  end
```
furkansahin, line 39: I feel like there is enough repetition in these 3 methods which you can DRY

prog/postgres/postgres_server_nexus.rb
```diff
@@ -601,17 +609,44 @@ def before_run
...
-      representative_server.vm.sshable.cmd("sudo pg_ctlcluster :version main stop -m immediate", version:)
-    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
-    end
+    hop_wait_representative_lockout
+  end
 
-    representative_server.incr_destroy
+  label def wait_representative_lockout
+    representative_server = postgres_server.resource.representative_server
+    nap 5 if representative_server && ["wait", "unavailable", "lockout", "wait_lockout_attempt"].include?(representative_server.strand.label)
```
furkansahin: shall we move these states to a constant and name?

prog/postgres/postgres_lockout.rb
shikharbhardwaj: I updated this to move the exception handling out to the caller. We can make them more terse, but I think this reads okay. A previous version also had a mechanism for AWS instances (which we may want to add), which made them less uniform.

prog/postgres/postgres_lockout.rb
```diff
@@ -0,0 +1,40 @@
...
+      Clog.emit("Fenced unresponsive primary", {fenced_unresponsive_primary: {server_ubid: postgres_server.ubid, mechanism:}})
+      pop "lockout_succeeded"
+    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
+      pop "lockout_failed"
+    end
+  end
+
+  def lockout_with_pg_stop
+    postgres_server.vm.sshable.cmd(
+      "timeout 1 sudo pg_ctlcluster :version main stop -m immediate",
```
byucesoy: - The timeout of 1 second seems a bit short. I suggest increasing this (and the others) to up to 10 seconds.
- Also, the 1-second difference between the in-VM timeout command and the sshable timeout command seems quite tight. Does the timeout in sshable.cmd include the connection duration as well?

prog/postgres/postgres_lockout.rb
```diff
@@ -0,0 +1,40 @@
...
+      version: postgres_server.version,
+      timeout: 2
+    )
+  end
+
+  def lockout_with_hba
+    postgres_server.vm.sshable.cmd(
+      "timeout 1 sudo postgres/bin/lockout-hba :version",
```
byucesoy: Why do we use the version that doesn't terminate existing connections?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -602,17 +610,44 @@ def before_run
...
-    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
-    end
+    hop_wait_representative_lockout
+  end
 
-    representative_server.incr_destroy
+  label def wait_representative_lockout
+    representative_server = postgres_server.resource.representative_server
+    nap 5 if representative_server&.waiting_for_lockout?
+    representative_server&.incr_destroy
```
byucesoy: We already hop_destroy in `wait_lockout_attempt`. Do we need `incr_destroy` here?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -602,17 +610,44 @@ def before_run
...
-      representative_server.vm.sshable.cmd("sudo pg_ctlcluster :version main stop -m immediate", version:)
-    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
-    end
+    hop_wait_representative_lockout
+  end
 
-    representative_server.incr_destroy
+  label def wait_representative_lockout
+    representative_server = postgres_server.resource.representative_server
+    nap 5 if representative_server&.waiting_for_lockout?
```
byucesoy: What if the representative_server is in another state? (e.g. hopped to `configure` just before unplanned_take_over started). I guess in that case, we trigger failover without fencing.

prog/postgres/postgres_lockout.rb
shikharbhardwaj: The sshable timeout includes connection duration. Increased the command timeout to 10s and sshable.cmd timeout to 15s.

prog/postgres/postgres_lockout.rb
shikharbhardwaj: I wanted to make the commands as minimal as possible to make them less likely to fail. In this case, we already have another command in parallel which is trying to shut down the server so running a psql in parallel did not seem valuable, wdyt?

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Updated this to wait for `destroy`

PR #4524, opened 2026-01-09
Handle lsn_monitor being nil in PostgresServer#failover_target

serprex:
Ran into this locally where lsn monitor wasn't being created because monitor wasn't running

### DIFF
--- a/spec/model/postgres/postgres_server_spec.rb
+++ b/spec/model/postgres/postgres_server_spec.rb
@@ -231,6 +231,11 @@ RSpec.describe PostgresServer do
       expect(postgres_server.failover_target).to be_nil
     end
 
+    it "returns nil if no lsn_monitor for async replication" do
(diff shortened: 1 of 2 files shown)


spec/model/postgres/postgres_server_spec.rb
```diff
@@ -231,6 +231,11 @@ def create_standby_resource(suffix)
       expect(postgres_server.failover_target).to be_nil
     end
 
+    it "returns nil if no lsn_monitor for async replication" do
+      expect(resource).to receive(:ha_type).and_return(PostgresResource::HaType::ASYNC)
```
jeremyevans: Best to avoid unnecessary mocking. I realize this is just following the surrounding code, but that code should be updated as well (not necessarily in this PR):

```suggestion
      resource.update(ha_type: PostgresResource::HaType::ASYNC)
```

jeremyevans (approved):

PR #4525, opened 2026-01-09
Remove support for passing blocks to Clog.emit

jeremyevans:
To avoid the chance someone passes a block that is silently ignored, raise for this case (without this, Ruby would warn, but tests would still pass).

### DIFF
--- a/lib/clog.rb
+++ b/lib/clog.rb
@@ -6,7 +6,12 @@ require "sequel/model"
 class Clog
   MUTEX = Mutex.new
 
-  def self.emit(message, metadata = block_given? ? yield : {})
+  def self.emit(message, metadata = {})
+    # :nocov:
+    # Cannot use spec passing block to cover this, or Ruby produces a warning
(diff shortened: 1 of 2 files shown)


lib/clog.rb
```diff
@@ -6,7 +6,12 @@
 class Clog
   MUTEX = Mutex.new
 
-  def self.emit(message, metadata = block_given? ? yield : {})
+  def self.emit(message, metadata = {})
+    # :nocov:
+    # Cannot use spec passing block to cover this, or Ruby produces a warning
+    raise "Clog.emit no longer takes a block" if block_given?
```
serprex, line 12: if we're going to raise here then clog_spec should be updated to expect this instead of `:nocov:`

jeremyevans: As the comment states, that would result in Ruby issuing a warning, and we want to keep our specs warning free.

fdr (approved):

PR #4526, opened 2026-01-09
Use physical slots to enable failover slots

serprex:
need to override in config:
```
wal_level = logical
hot_standby_feedback = on
sync_replication_slots = on
```

this then handles the part where standbys use a slot once caught up,
synchronized_standby_slots prevents logical replication getting ahead of standby

https://clickhouse.com/blog/clickpipes-postgres-failover-replication

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -101,6 +101,7 @@ class PostgresResource < Sequel::Model
       sslcert: "/etc/ssl/certs/server.crt",
       sslkey: "/etc/ssl/certs/server.key",
       sslmode: "verify-full",
+      dbname: "postgres",
       application_name:
     }.map { |k, v| "#{k}=#{v}" }.join("&")
 
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -17,7 +17,7 @@ class PostgresServer < Sequel::Model
   plugin ProviderDispatcher, __FILE__
   plugin SemaphoreMethods, :initial_provisioning, :refresh_certificates, :update_superuser_password, :checkup,
     :restart, :configure, :fence, :unfence, :planned_take_over, :unplanned_take_over, :configure_metrics,
-    :destroy, :recycle, :promote, :refresh_walg_credentials, :configure_s3_new_timeline, :lockout
+    :destroy, :recycle, :promote, :refresh_walg_credentials, :configure_s3_new_timeline, :lockout, :use_physical_slot
   include HealthMonitorMethods
   include MetricsTargetMethods
 
@@ -83,6 +83,7 @@ class PostgresServer < Sequel::Model
       configs["allow_alter_system"] = "off"
     end
 
+    caught_up_standbys = nil
     if timeline.blob_storage
       configs[:archive_mode] = "on"
       configs[:archive_timeout] = "60"
@@ -93,14 +94,18 @@ class PostgresServer < Sequel::Model
       end
 
       if primary?
+        caught_up_standbys = resource.servers.select { it.standby? && it.synchronization_status == "ready" }
         if resource.ha_type == PostgresResource::HaType::SYNC
-          caught_up_standbys = resource.servers.select { it.standby? && it.synchronization_status == "ready" }
           configs[:synchronous_standby_names] = "'ANY 1 (#{caught_up_standbys.map(&:ubid).join(",")})'" unless caught_up_standbys.empty?
+        end
+        if version.to_i >= 17
+          configs[:synchronized_standby_slots] = "'#{caught_up_standbys.map(&:ubid).join(",")}'"
         end
       end
 
       if standby?
         configs[:primary_conninfo] = "'#{resource.replication_connection_string(application_name: ubid)}'"
+        configs[:primary_slot_name] = "'#{ubid}'" if physical_slot_ready
       end
 
       if doing_pitr?
@@ -118,6 +123,7 @@ class PostgresServer < Sequel::Model
       configs:,
       user_config: resource.user_config,
       pgbouncer_user_config: resource.pgbouncer_user_config,
+      physical_slots: caught_up_standbys&.map(&:ubid),
       private_subnets: vm.private_subnets.map {
         {
           net4: it.net4.to_s,
@@ -192,7 +198,7 @@ class PostgresServer < Sequel::Model
   def failover_target
     target = resource.servers
       .reject { it.representative_at }
-      .select { it.strand.label == "wait" && !it.needs_recycling? }
+      .select { it.strand.label == "wait" && !it.needs_recycling? && it.physical_slot_ready }
       .map { {server: it, lsn: it.current_lsn} }
       .max_by { lsn2int(it[:lsn]) }
 
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -362,8 +362,20 @@ CONFIG
       end
 
       hop_wait_catch_up if postgres_server.standby? && postgres_server.synchronization_status != "ready"
+
+      if postgres_server.primary?
+        postgres_server.resource.servers.select { it.standby? && it.synchronization_status == "ready" && !it.physical_slot_ready }.each do |standby|
+          standby.incr_use_physical_slot
+          standby.incr_configure
+        end
+      end
+
       hop_wait
     when "Failed", "NotStarted"
+      if postgres_server.use_physical_slot_set?
+        postgres_server.update(physical_slot_ready: true)
+        decr_use_physical_slot
+      end
       configure_hash = postgres_server.configure_hash
       vm.sshable.d_run("configure_postgres", "sudo", "postgres/bin/configure", postgres_server.version, stdin: JSON.generate(configure_hash))
     end
--- a/rhizome/postgres/bin/configure
+++ b/rhizome/postgres/bin/configure
@@ -83,6 +83,7 @@ host    replication     all             ::1/128                 scram-sha-256
 # Allow replication connection using special replication user for
 # HA standbys
 hostssl replication     ubi_replication all                     cert map=standby2replication
+hostssl postgres        ubi_replication all                     cert map=standby2replication
 
 # Allow certificate authentication for specific users
 #{cert_auth_users}
@@ -120,6 +121,17 @@ safe_write_to_file("/etc/systemd/system/postgresql@.service.d/hugepages.conf", e
 
 # Reload the postmaster to apply changes
 r "pg_ctlcluster #{v} main reload || pg_ctlcluster #{v} main restart"
+
+if configure_hash["physical_slots"]
+  if configure_hash["physical_slots"].size == 0
+    r "sudo", "-u", "postgres", "psql", "-c", "select pg_catalog.pg_drop_replication_slot(slot_name) from pg_catalog.pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical'"
+  else
+    r "sudo", "-u", "postgres", "psql", "-c", "select pg_catalog.pg_drop_replication_slot(slot_name) from pg_catalog.pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical' and slot_name not in (#{configure_hash["physical_slots"].map { |x| "'#{x}'" }.join(",")})"
+    configure_hash["physical_slots"].each do |ubid|
+      r "sudo", "-u", "postgres", "psql", "-c", "select pg_catalog.pg_create_physical_replication_slot('#{ubid}', false) where not exists (select 1 from pg_catalog.pg_replication_slots where slot_type = 'physical' and slot_name = '#{ubid}')"
+    end
+  end
+end
 
 pgbouncer_setup = PgBouncerSetup.new(v, configure_hash["configs"]["max_connections"], configure_hash["pgbouncer_instances"], configure_hash["pgbouncer_user_config"])
 pgbouncer_setup.setup
--- a/spec/model/postgres/postgres_server_spec.rb
+++ b/spec/model/postgres/postgres_server_spec.rb
@@ -87,6 +87,11 @@ RSpec.describe PostgresServer do
       expect(postgres_server.configure_hash[:configs]).to include(:archive_mode, :archive_timeout, :archive_command)
     end
 
+    it "sets synchronized_standby_slots on Postgres 17" do
+      postgres_server.update(version: "17")
+      expect(postgres_server.configure_hash[:configs]).to include(:synchronized_standby_slots)
+    end
+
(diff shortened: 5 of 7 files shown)


rhizome/postgres/bin/configure
```diff
@@ -83,6 +83,7 @@ host    replication     all             ::1/128                 scram-sha-256
 # Allow replication connection using special replication user for
 # HA standbys
 hostssl replication     ubi_replication all                     cert map=standby2replication
+hostssl postgres        ubi_replication all                     cert map=standby2replication
```
byucesoy, line 86: @furkansahin, assuming you would be one to deploy this change, this is a reminder that we would need to update rhizome in existing databases.

rhizome/postgres/bin/configure
```diff
@@ -121,5 +122,16 @@ safe_write_to_file("/etc/systemd/system/postgresql@.service.d/hugepages.conf", e
 # Reload the postmaster to apply changes
 r "pg_ctlcluster #{v} main reload || pg_ctlcluster #{v} main restart"
 
+if configure_hash["physical_slots"]
+  if configure_hash["physical_slots"].size == 0
+    r "sudo -u postgres psql -c \"select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical'\""
+  else
+    r "sudo -u postgres psql -c \"select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical' and slot_name not in (#{configure_hash["physical_slots"].map { |x| "'#{x}'" }.join(",")})\""
+    configure_hash["physical_slots"].each do |ubid|
+      r "sudo -u postgres psql -c \"select pg_create_physical_replication_slot('#{ubid}', false) where not exists (select 1 from pg_replication_slots where slot_type = 'physical' and slot_name = '#{ubid}')\""
```
byucesoy: Why do we drop the existing replication slots and re-create them? Is there are risk of missing some changes during drop & recreate?

model/postgres/postgres_server.rb
```diff
@@ -89,14 +90,16 @@ def configure_hash
...
         if resource.ha_type == PostgresResource::HaType::SYNC
-          caught_up_standbys = resource.servers.select { it.standby? && it.synchronization_status == "ready" }
           configs[:synchronous_standby_names] = "'ANY 1 (#{caught_up_standbys.map(&:ubid).join(",")})'" unless caught_up_standbys.empty?
         end
+        configs[:synchronized_standby_slots] = "'#{caught_up_standbys.map(&:ubid).join(",")}'"
       end
 
       if standby?
         configs[:primary_conninfo] = "'#{resource.replication_connection_string(application_name: ubid)}'"
+        configs[:primary_slot_name] = "'#{ubid}'" if use_physical_slot
```
byucesoy: So this requires someone to set this flag in the beginning. What would be the one setting this?

Also, shouldn't this flag be on the resource level instead of server?

byucesoy: Are there any harm in setting `primary_slot_name` unconditionally for all clusters with HA?

rhizome/postgres/bin/configure
serprex: we only drop slots which aren't in our list of desired slots, this handles dropping slots for standbys which have been recycled

model/postgres/postgres_server.rb
serprex: each standby gets its own slot. we only set slot name when they're caught up & a slot is available. setting it earlier would result in them attempting to use a slot which doesn't exist. we don't want to create slot until they're caught up because we don't want wal to build up while standby is pulling wal from s3

there's an option to not reserve wal until slot being read, but then that confuses matter of not letting logical replication get ahead of slots with `synchronized_standby_slots` & when we put slot in that setting or not

serprex:
tested that failover slot survived incr_recycle on stable HA setup, & with replacing VM sku

model/postgres/postgres_server.rb
```diff
@@ -184,7 +188,7 @@ def current_lsn
   def failover_target
     target = resource.servers
       .reject { it.representative_at }
-      .select { it.strand.label == "wait" && !it.needs_recycling? }
+      .select { it.strand.label == "wait" && !it.needs_recycling? && it.physical_slot_ready }
```
serprex, line 201: this kind of races, since we set `physical_slot_ready` to true in configure before the slot is actually configured for use, but by time `== "wait"` resolves this ambiguous transition is over

model/postgres/postgres_resource.rb
```diff
@@ -98,11 +98,12 @@ def replication_connection_string(application_name:)
       sslrootcert: "/etc/ssl/certs/ca.crt",
       sslcert: "/etc/ssl/certs/server.crt",
       sslkey: "/etc/ssl/certs/server.key",
-      sslmode: "verify-full",
+      sslmode: dns_zone ? "verify-full" : "require",
```
fdr: huh, why? that seems orthogonal, although of lesser consequence with mtls.

model/postgres/postgres_server.rb
```diff
@@ -89,14 +90,16 @@ def configure_hash
       end
 
       if primary?
+        caught_up_standbys = resource.servers.select { it.standby? && it.synchronization_status == "ready" }
         if resource.ha_type == PostgresResource::HaType::SYNC
-          caught_up_standbys = resource.servers.select { it.standby? && it.synchronization_status == "ready" }
           configs[:synchronous_standby_names] = "'ANY 1 (#{caught_up_standbys.map(&:ubid).join(",")})'" unless caught_up_standbys.empty?
         end
+        configs[:synchronized_standby_slots] = "'#{caught_up_standbys.map(&:ubid).join(",")}'"
```
fdr: only valid in 17+. We have PG16 rolling around.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -362,8 +362,20 @@ def before_run
       end
 
       hop_wait_catch_up if postgres_server.standby? && postgres_server.synchronization_status != "ready"
+
+      if postgres_server.primary?
+        postgres_server.resource.servers.select { it.standby? && it.synchronization_status == "ready" && !it.physical_slot_ready }.each do |standby|
```
fdr, line 367: what makes this safe under concurrent changes of these fields between the read and the semaphore incrementation?

model/postgres/postgres_resource.rb
```diff
@@ -98,11 +98,12 @@ def replication_connection_string(application_name:)
...
+      dbname: "postgres",
       application_name:
     }.map { |k, v| "#{k}=#{v}" }.join("&")
 
-    URI::Generic.build2(scheme: "postgres", userinfo: "ubi_replication", host: identity, query: query_parameters).to_s
+    URI::Generic.build2(scheme: "postgres", userinfo: "ubi_replication", host: dns_zone ? identity : representative_server&.vm&.ip4_string, path: "/replication", query: query_parameters).to_s
```
fdr: @jeremyevans what do you think we should do about this? to me it looks like a development device, but it's intertwined with a downgrade of ssl validation. On the other hand, the tls validation doesn't matter as much as it is via mtls.

fdr: for reference, the branch is also entangled  with https://github.com/ubicloud/ubicloud/pull/4526/files#r2710942012

model/postgres/postgres_server.rb
```diff
@@ -89,14 +90,16 @@ def configure_hash
...
       end
 
       if standby?
         configs[:primary_conninfo] = "'#{resource.replication_connection_string(application_name: ubid)}'"
```
fdr, line 107: this quoting situation -- not this patch's fault -- is pretty jank. most values in postgresql.conf can afford to be quoted, even the integers, a bit like the `unknown` type in the database

model/postgres/postgres_server.rb
fdr: so the monotonic logic is: the main thing this can guarantee is that the `wait` label was touched once. It can't guarantee anything about the label state in what is returned, as expected.

rhizome/postgres/bin/configure
```diff
@@ -121,5 +122,16 @@ safe_write_to_file("/etc/systemd/system/postgresql@.service.d/hugepages.conf", e
 # Reload the postmaster to apply changes
 r "pg_ctlcluster #{v} main reload || pg_ctlcluster #{v} main restart"
 
+if configure_hash["physical_slots"]
+  if configure_hash["physical_slots"].size == 0
+    r "sudo -u postgres psql -c \"select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical'\""
+  else
+    r "sudo -u postgres psql -c \"select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical' and slot_name not in (#{configure_hash["physical_slots"].map { |x| "'#{x}'" }.join(",")})\""
```
fdr: @jeremyevans some Ideas about...quoting sensitivity.

fdr: I wonder if it's worthwhile to have the pg gem around for postgres's rhizome.

fdr (changes requested):
Pretty sure we have a version compatibility problem. need to think about that race condition carefully. Commit message is too cryptic. What are the new operational obligations of this, now, as I don't see a maximum retention configured for these physical slots...so they can be a new source of wal buildup. Plus, the dns_zone change, which looks unrelated and just kinda slides in.

rhizome/postgres/bin/configure
fdr: probably also worth while full-qualifying to `pg_catalog.` and/or being explicit to override SET search_path. This goes for our other utility function calls, now that I think about it.

model/postgres/postgres_resource.rb
serprex: without dns_zone connection is made with IP, which certs can't cover *(this wasn't working at all before, where connection is using identity as domain name unconditionally)*

prog/postgres/postgres_server_nexus.rb
serprex: you can get some unnecessary configures here, harmless

fdr:
I had claude take a crack (with a bit of my help) at writing a commit message. I fed it a perplexity report on postgres 17 enhancements in replication slots, plus some grilling by me, plus fanning out over every hunk to analyze how every line supported or failed to support the commit message, plus COMMIT_MESSAGES.md with a suggestion to read some of the examples for style guiding.

However, it needs your input since you debugged this, to ground truth it. My questions are postfix to each section.

```
  postgres: Add physical replication slots for failover slot support                                                                                                                    
                                                                                                                                                                                        
  PostgreSQL 17 introduced failover slots, which synchronize logical                                                                                                                    
  replication slots to standbys so that after promotion, subscribers                                                                                                                    
  can resume without data loss. This feature requires physical                                                                                                                          
  replication slots between primary and standby.                
```

Front matter; is it true?                                                                                                                        
 
```                                                                                                                                                                                       
  This change creates physical slots on the primary for each caught-up                                                                                                                  
  standby and configures standbys to use them. The synchronized_standby_slots                                                                                                           
  parameter lists these slots; logical replication will not advance                                                                                                                     
  until all listed standbys have received the WAL. 
```

Note/confirm: **all**. I haven't worked through all the ramifications, but I wanted to underline that hooking up new standbys can cause a hiccup in logical replication, its' something we have to be aware of.                                                                                                                                     
   
```                                                                                                                                                                                     
  The primary signals standbys via semaphore after creating their slots;                                                                                                                
  standbys wait for this signal before adding primary_slot_name to their                                                                                                                
  configuration. This ensures the slot exists before the standby                                                                                                                        
  references it.                                                                                                                                                                        
```

Or what? it spins failing to connect, making a bit of log spam? Something worse?

```                                                                                                                                                                                       
  Failover target selection now requires that the standby is using its                                                                                                                  
  physical slot, since a standby without slot-based replication cannot                                                                                                                  
  preserve logical slot state after promotion.    
```

Is that indeed the idea?

model/postgres/postgres_resource.rb
serprex: split out to #4621 since this change is isolated & was made to test with local controlplane

serprex:
That is the idea

rhizome/postgres/bin/configure
serprex: this is a wider issue addressed in #4622

rhizome/postgres/bin/configure
```diff
@@ -121,5 +122,16 @@ safe_write_to_file("/etc/systemd/system/postgresql@.service.d/hugepages.conf", e
 # Reload the postmaster to apply changes
 r "pg_ctlcluster #{v} main reload || pg_ctlcluster #{v} main restart"
 
+if configure_hash["physical_slots"]
+  if configure_hash["physical_slots"].size == 0
+    r "sudo -u postgres psql -c \"select pg_catalog.pg_drop_replication_slot(slot_name) from pg_catalog.pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical'\""
```
jeremyevans: With changes that went in yesterday, `r` now supports separate arguments to avoid the use of a shell. This should be switched to:

```suggestion
    r "sudo", "-u", "postgres", "psql", "-c", "select pg_catalog.pg_drop_replication_slot(slot_name) from pg_catalog.pg_replication_slots where slot_name like 'pv%' and LENGTH(slot_name) = 26 and slot_type = 'physical'"
```

Similar changes to the other cases. This particular example doesn't use interpolation, but the two following cases do. Using separate arguments at least avoids shell quoting issues. You still need to worry about correct PostgreSQL quoting issues.  We should probably use a similar approach as #4283 for that. However, that doesn't need to be done in this commit.  If `physical_slots` is an array of ubids, this particular usage appears safe.

prog/postgres/postgres_server_nexus.rb
fdr: that doesn't answer the question, although it does document an outcome.

model/postgres/postgres_resource.rb
fdr: thank you for that

prog/postgres/postgres_server_nexus.rb
fdr: okay, analyzed it myself. Maybe I'm even right? You could tell me.

it.standby? : presumed monotonic from true to false, but  standby.incr_use_physical_slot is harmless although nonsensical in that case, resulting in a spurious update. configure is familiar and safe at any time we branch into it.

synchronization status: monotonic into "ready" for now, even if there is later fall-behind.

physical_slot_ready: goes from false to true, monotonic, prevents excess incrs (presumed?)

So both the effect of standby flipping to false concurrently is harmless, and it also obsoletes the need to do this, so no necessary work can be missed either.

serprex: Yes

PR #4527, opened 2026-01-10
Add pg_config & pgbouncer_config to endpoints for creating read replicas

serprex:
Also add tags to restore, which was previously added to read replica creation

### DIFF
--- a/cli-commands/pg/post/create-read-replica.rb
+++ b/cli-commands/pg/post/create-read-replica.rb
@@ -3,7 +3,9 @@
 UbiCli.on("pg").run_on("create-read-replica") do
   desc "Create a read replica for a PostgreSQL database"
 
-  options("ubi pg (location/pg-name | pg-id) create-read-replica name [options]", key: :pg_create_read_replica) do
+  options("ubi pg (location/pg-name | pg-id) create-read-replica [options] name", key: :pg_create_read_replica) do
+    on("-c", "--pg-config=config", "postgres config (e.g key1=value1,key2=value2)")
+    on("-u", "--pgbouncer-config=config", "pgbouncer config (e.g. key1=value1,key2=value2)")
     on("-t", "--tags=tags", "tags (e.g. key1=value1,key2=value2)")
   end
 
@@ -12,6 +14,8 @@ UbiCli.on("pg").run_on("create-read-replica") do
   run do |name, opts, cmd|
     params = underscore_keys(opts[:pg_create_read_replica])
     pg_tags_to_hash(params, cmd)
+    params_to_hash(params, :pg_config, "config", cmd)
+    params_to_hash(params, :pgbouncer_config, "pgbouncer config", cmd)
     id = sdk_object.create_read_replica(name, **params).id
     response("Read replica for PostgreSQL database created with id: #{id}")
   end
--- a/cli-commands/pg/post/restore.rb
+++ b/cli-commands/pg/post/restore.rb
@@ -3,12 +3,20 @@
 UbiCli.on("pg").run_on("restore") do
   desc "Restore a PostgreSQL database backup to a new database"
 
-  banner "ubi pg (location/pg-name | pg-id) restore new-db-name restore-time"
+  options("ubi pg (location/pg-name | pg-id) restore [options] new-db-name restore-time", key: :pg_restore) do
+    on("-c", "--pg-config=config", "postgres config (e.g. key1=value1,key2=value2)")
+    on("-u", "--pgbouncer-config=config", "pgbouncer config (e.g. key1=value1,key2=value2)")
+    on("-t", "--tags=tags", "tags (e.g. key1=value1,key2=value2)")
+  end
 
   args 2
 
-  run do |name, restore_target|
-    id = sdk_object.restore(name:, restore_target:).id
+  run do |name, restore_target, opts, cmd|
+    params = underscore_keys(opts[:pg_restore])
+    pg_tags_to_hash(params, cmd)
+    params_to_hash(params, :pg_config, "config", cmd)
+    params_to_hash(params, :pgbouncer_config, "pgbouncer config", cmd)
+    id = sdk_object.restore(name:, restore_target:, **params).id
     response("Restored PostgreSQL database scheduled for creation with id: #{id}")
   end
 end
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -239,6 +239,8 @@ class Clover
         handle_validation_failure("postgres/show") { @page = "read-replica" }
 
         name = typecast_params.nonempty_str!("name")
+        user_config = typecast_params.Hash("pg_config", {})
+        pgbouncer_user_config = typecast_params.Hash("pgbouncer_config", {})
         tags = typecast_params.array(:Hash, "tags", [])
 
         Validation.validate_name(name)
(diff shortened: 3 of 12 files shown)


cli-commands/pg/post/create-read-replica.rb
```diff
@@ -12,6 +12,8 @@
   run do |name, opts, cmd|
     params = underscore_keys(opts[:pg_create_read_replica])
     pg_tags_to_hash(params, cmd)
+    params_to_hash(params, :pg_config, "config", cmd)
+    params_to_hash(params, :pgbouncer_config, "pgbouncer config", cmd)
```
jeremyevans, line 18: We don't accept options for these, so I'm not sure how they can be present in `params`. You need to add  options for them, similar to what you did for the `restore` command. You also need to add a CLI test that uses the options.

routes/project/location/postgres.rb
```diff
@@ -304,11 +320,26 @@ class Clover
...
+        pg_errors = pg_validator.validation_errors(user_config)
+
+        pgbouncer_validator = Validation::PostgresConfigValidator.new("pgbouncer")
+        pgbouncer_errors = pgbouncer_validator.validation_errors(pgbouncer_user_config)
+
+        if pg_errors.any? || pgbouncer_errors.any?
+          pg_errors = pg_errors.transform_keys { |key| "pg_config.#{key}" }
+          pgbouncer_errors = pgbouncer_errors.transform_keys { |key| "pgbouncer_config.#{key}" }
+          raise Validation::ValidationFailed.new(pg_errors.merge(pgbouncer_errors))
+        end
```
jeremyevans: This looks almost identical to the other hunk. Unless there is a reason for these to be different, please add a method to `helpers/postgres.rb` and move the shared logic there, and call the method from both places (and potentially other places using this logic).

cli-commands/pg/post/restore.rb
```diff
@@ -3,12 +3,20 @@
 UbiCli.on("pg").run_on("restore") do
   desc "Restore a PostgreSQL database backup to a new database"
 
-  banner "ubi pg (location/pg-name | pg-id) restore new-db-name restore-time"
+  options("ubi pg (location/pg-name | pg-id) restore new-db-name restore-time [options]", key: :pg_restore) do
+    on("-c", "--pg-config=config", "postgres config (e.g key1=value1,key2=value2)")
+    on("-u", "--pgbouncer-config=config", "pgbouncer config (e.g. key1=value1,key2=value2)")
+    on("-t", "--tags=tags", "tags (e.g. key1=value1,key2=value2)")
```
jeremyevans: There needs to be a CLI test that actually uses these options.

cli-commands/pg/post/restore.rb
```diff
@@ -3,12 +3,20 @@
 UbiCli.on("pg").run_on("restore") do
   desc "Restore a PostgreSQL database backup to a new database"
 
-  banner "ubi pg (location/pg-name | pg-id) restore new-db-name restore-time"
+  options("ubi pg (location/pg-name | pg-id) restore [options] new-db-name restore-time", key: :pg_restore) do
+    on("-c", "--pg-config=config", "postgres config (e.g key1=value1,key2=value2)")
```
jeremyevans: `e.g ` vs `e.g. ` seems like an unnecessary inconsistency. I'm guessing this was copied from the `create` command, where it is also inconsistent. I'll submit a PR for that.

```suggestion
    on("-c", "--pg-config=config", "postgres config (e.g. key1=value1,key2=value2)")
```

Same issue with `create-read-replica` command.

sdk/ruby/lib/ubicloud/model/postgres.rb
```diff
@@ -136,8 +136,8 @@ def reset_superuser_password(password)
     # Schedule a restore of the database at the given restore_target time, to a new
     # database with the given name.  Returns a Postgres instance for the restored
     # database.
-    def restore(name:, restore_target:)
-      Postgres.new(adapter, adapter.post(_path("/restore"), name:, restore_target:))
+    def restore(**params)
+      Postgres.new(adapter, adapter.post(_path("/restore"), **params))
```
jeremyevans: As `name` and `restore_target` are required, I don't think it's wise to switch them to optional arguments. Also, no reason to name the splat argument:

```suggestion
    def restore(name:, restore_target:, **)
      Postgres.new(adapter, adapter.post(_path("/restore"), name:, restore_target:, **))
```

sdk/ruby/lib/ubicloud/model/postgres.rb
```diff
@@ -108,8 +108,8 @@ def update_pgbouncer_config(**values)
     end
 
     # Create a read replica of this database, with the given name.
-    def create_read_replica(name, tags: [])
-      Postgres.new(adapter, adapter.post(_path("/read-replica"), name:, tags:))
+    def create_read_replica(name, **params)
+      Postgres.new(adapter, adapter.post(_path("/read-replica"), name:, **params))
```
jeremyevans, line 117:
```suggestion
    def create_read_replica(name, **)
      Postgres.new(adapter, adapter.post(_path("/read-replica"), name:, **))
```

cli-commands/pg/post/restore.rb
jeremyevans: Submitted #4533 to fix the issue for `create`.

cli-commands/pg/post/create-read-replica.rb
```diff
@@ -12,7 +14,9 @@
   run do |name, opts, cmd|
     params = underscore_keys(opts[:pg_create_read_replica])
     pg_tags_to_hash(params, cmd)
-    id = sdk_object.create_read_replica(name, **params).id
+    params_to_hash(params, :pg_config, "config", cmd)
+    params_to_hash(params, :pgbouncer_config, "pgbouncer config", cmd)
+    id = sdk_object.create_read_replica(@name, name:, **params).id
```
jeremyevans: As mentioned on Slack:

```suggestion
    id = sdk_object.create_read_replica(name, **params).id
```

jeremyevans:
This is failing because the instead of a passing `[{"key": "CONFIG-NAME", "value": "CONFIG-VALUE"}, ...]` to the API, you are passing `{"CONFIG-NAME": "CONFIG-VALUE", ...}`. `committee` (the OpenAPI validation layer) rejects this. You probably need similar code that is used in `create` in the SDK file:

```ruby
if params[:tags]
  params[:tags] = params[:tags].map { |key, value| {key:, value:} }
end
```

sdk/ruby/lib/ubicloud/model/postgres.rb
```diff
@@ -108,8 +108,12 @@ def update_pgbouncer_config(**values)
     end
 
     # Create a read replica of this database, with the given name.
-    def create_read_replica(name, **)
-      Postgres.new(adapter, adapter.post(_path("/read-replica"), name:, **))
+    def create_read_replica(name, **params)
+      if params[:tags]
+        params[:tags] = params[:tags].map { |key, value| {key:, value:} }
+      end
```
jeremyevans: Should add a method that does this, and call it here, in `restore`, and in the `set_create_param_defaults` block.

jeremyevans (approved):
One small change, then it looks good to merge.

Tests are failing due to missing branch coverage.

Please squash before merging.

PR #4528, opened 2026-01-10
DownloadBootImage improvements

enescakir:
- **Add migration to make boot image version column not null**
  The version column historically allowed null because older boot images
  did not have this field. Now that all existing images have a version in
  production, we can enforce the constraint at the database level.
  

- **Remove nil version handling from boot image code**
  The previous commit (4fd55ee73) added a NOT NULL constraint on the
  version column. This commit removes unnecessary nil checks and
  simplifies the code that relied on them.
  

- **Move labels to top of DownloadBootImage prog**
  No functional changes. The file had many helper methods defined before
  the labels, making it hard to see the program flow at a glance. This
  reorganizes the file to place labels first, followed by constants and
  helper methods. The spec file is reordered to match.
  

- **Add assemble method to DownloadBootImage**
  Move parameter validation and resource creation from the start label
  into an assemble class method. This follows our convention for progs
  and provides two benefits:
  
  1. The start label was not idempotent: creating the BootImage there
     caused failures on retry since the record already existed.
  
  2. Validation errors (e.g., unknown image name) were silent failures
     inside the strand. Now they fail immediately at assemble time with
     a clear error.
  
  Also add a destroy label for cleanup. Previously, after 10 failed
  download attempts, the prog would delete the BootImage and continue
  napping. Now it allows operator to clean up by hopping to destroy.
  

- **Require sha256 checksum for boot images in production**
  Previously, images without a known sha256 checksum would download
  without verification.
  
  The sha256sum method now fails in production if couldn't find it. In
  development, unknown images are still allowed to enable testing with
  custom images.


enescakir:
All existing images have a version in production, but some images in development environments may still have a nil version. I’m not sure how to make the migration smooth for those environments.

model/vm_host.rb
```diff
@@ -225,8 +225,8 @@ def install_rhizome(install_specs: false)
   end
 
   # Introduced for downloading a new boot image via REPL.
-  def download_boot_image(image_name, version:, custom_url: nil, download_r2: true)
-    Strand.create(prog: "DownloadBootImage", label: "start", stack: [{subject_id: id, image_name:, custom_url:, version:, download_r2:}])
+  def download_boot_image(*, **)
+    Prog::DownloadBootImage.assemble(self, *, **)
```
jeremyevans: Generally best to use `...` instead of both `*` and `**`, unless the specific intent is to ignore the passed block instead of passing it.

```suggestion
  def download_boot_image(...)
    Prog::DownloadBootImage.assemble(self, ...)
```

prog/download_boot_image.rb
```diff
@@ -6,66 +6,74 @@
 class Prog::DownloadBootImage < Prog::Base
   subject_is :sshable, :vm_host
 
-  def image_name
-    @image_name ||= frame.fetch("image_name")
+  def self.assemble(vm_host, name, version: nil, custom_url: nil, download_r2: true)
+    version ||= default_boot_image_version(name)
```
jeremyevans: Unless you feel strongly that we should handle an explicit `nil` version, better to use a default argument:

```suggestion
  def self.assemble(vm_host, name, version: default_boot_image_version(name), custom_url: nil, download_r2: true)
```

prog/download_boot_image.rb
```diff
@@ -131,10 +139,80 @@ def url
...
+  def daemon_name
+    "download_#{image_name}_#{version}"
+  end
+
   def sha256sum
-    # YYY: In future all images should be checked for sha256 sum, so the nil
-    # default will be removed.
-    BOOT_IMAGE_SHA256.fetch([image_name, vm_host.arch, version], nil)
+    BOOT_IMAGE_SHA256.fetch([image_name, vm_host.arch, version]) do
+      fail "Downloading boot image without sha256 not allowed in production" if Config.production?
```
jeremyevans:
```suggestion
      fail "Downloading a boot image without sha256 is not allowed in production" if Config.production?
```

jeremyevans (approved):
Looks good, just some minor suggestions.

enescakir:
I found an issue. Since we also bud Prog::DownloadBootImage in the host nexus, it bypasses the assemble step. As a result, the version isn’t set to the default and the BootImage resource isn’t created. I need to think about a way to combine assemble and budding.

furkansahin (commented):
I will wait for you to fix that issue for my review.

PR #4529, opened 2026-01-10
Add constraint ensuring used_hugepages_1g >= 0

enescakir:
Over-deallocation bugs have occasionally caused used_hugepages_1g to go negative in production. Adding a database constraint catches these bugs at write time rather than when symptoms appear later. This mirrors the existing used_cores_above_zero constraint.


migrate/20260110_add_hugepage_above_zero_constraint.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:vm_host) do
+      # Catch over-deallocation bugs.
+      add_constraint(:used_hugepages_above_zero) { used_hugepages_1g >= 0 }
```
bsatzger: nit: the name `used_hugepages_above_zero` implies > 0 not >=. since it's consistent with other already existing constraints (e.g., used_cores_above_zero | used_cores >= 0) it shouldn't be a blocker. otherwise I'd have recommended something like:`used_hugepages_non_negative`

PR #4538, opened 2026-01-13
Display current usage of project in the admin panel

enescakir:
- **Add space between object extra in the admin panel**
  

- **Add collapsible table to admin table**
  Adding new extras to object detail pages makes them longer and harder to
  find relevant data. When I added current usage data to the project
  detail page, it became very long. To address this, I added the ability
  to collapse tables so the data is hidden by default and shown only when
  needed.
  
  This is implemented using the HTML details/summary elements, which is a
  pure CSS solution.
  
  I also introduced a new component instead of adding multiple if/else
  blocks to the existing table component.
  
  Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>
  

- **Display current usage of project in the admin panel**
  
  Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

CleanShot 2026-01-13 at 18 02 51@2x

CleanShot 2026-01-13 at 18 02 57@2x

### DIFF
--- a/spec/routes/web/admin/admin_spec.rb
+++ b/spec/routes/web/admin/admin_spec.rb
@@ -410,6 +410,25 @@ RSpec.describe CloverAdmin do
     ]
   end
 
+  it "shows current usage for project as extra" do
+    project = Project.create(name: "test")
+    vm = create_vm(project_id: project.id)
+    BillingRecord.create(
+      project_id: project.id,
+      resource_id: vm.id,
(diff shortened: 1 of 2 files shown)


enescakir:
> I'm OK with this, though I think it would be better for performance to link to a different page with that information, instead of having it hidden by default on the Project page.

I agree with that. I'm planning to add custom pages for objects. When we have that, we can move them to there.

PR #4542, opened 2026-01-13
Return 400 instead of 500 for invalid Kubernetes version

aamederen:
This commit adds proper validation for Kubernetes version parameter that makes the endpoint return a 400 Bad Request with helpful error message instead of throwing a 500.


jeremyevans (commented):
Looks fine from a Ruby perspective.

PR #4543, opened 2026-01-13
Dump Clover database on E2E tests

aamederen:
Add pg_dump step to capture database state and upload as artifact for debugging.
The pg_dump for now includes an allowlisted subset of tables that's relatively safe to export

New sample run: https://github.com/ubicloud/ubicloud/actions/runs/21042941070

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -192,6 +192,51 @@ jobs:
         name: e2e-${{ github.run_id }}-logs
         path: foreman.log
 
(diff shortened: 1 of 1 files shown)


enescakir:
One more thing, we delete the related models at the end of the tests, so the database state may not be very useful. Most tables are empty. Dumping only the ArchivedRecord table would likely cover most of what we need.

enescakir (approved):
We should remember to remove them if we add permanent credentials to any of these tables. jfyi

PR #4544, opened 2026-01-13
Check for PostgresServer#run_query insecurity

jeremyevans:
This changes PostgresServer#run_query to work similarly to Sshable#cmd, and fail if passed a non-frozen string. To handle cases currently using interpolation, instead of the :placeholder usage, allow passing a Sequel::Dataset, and take the SQL from that (without parameterization, since the full SQL is needed). The Sequel::Dataset can use :placeholders, so this allows for very similar code as used for shell command placeholder escaping.

Have run_query call _run_query, and have the specs override that.

Check that run_query itself is never mocked in the specs.

While non of the interpolated strings appear to be vulnerable, this replaces an insecure pattern with a secure pattern, which should be preferable.

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -178,7 +178,7 @@ class PostgresServer < Sequel::Model
   end
 
   def current_lsn
-    run_query("SELECT #{lsn_function}").chomp
+    run_query(DB.select(Sequel.function(lsn_function_name)))
   end
 
   def failover_target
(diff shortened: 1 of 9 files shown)


jeremyevans:
E2E failed due to `Test::PostgresResource.test_postgres | waiting for exit`. I found a `run_query` in `Prog::Test::PostgresResource` called with a frozen string, so I fixed it. Not caught by specs due to mocking in `spec/prog/test/postgres_resource_spec.rb`.

jeremyevans:
E2E passed.

model/postgres/postgres_server.rb
```diff
@@ -198,16 +198,20 @@ def failover_target
...
     elsif standby?
-      "pg_last_wal_receive_lsn()"
+      "pg_last_wal_receive_lsn"
     else
-      "pg_last_wal_replay_lsn()"
+      "pg_last_wal_replay_lsn"
     end
   end
 
+  def lsn_function
```
byucesoy, line 211: In check_pulse, we still use old approach. Should we update it as well. If we do that, then we wouldn't need lsn_function because all callers would use lsn_function_name instead.

jeremyevans: Probably a good idea to switch that as well. It doesn't use `run_query`, which is why it wasn't caught by this.

jeremyevans: Added #4557 for this.

PR #4547, opened 2026-01-14
Demock postgres_resource_spec

fdr:
Instance doubles for PostgresResource and PostgresServer disconnected these tests from real database behavior. Schema or association changes would go undetected since mocks don't validate against actual models.

Replace all Sequel model mocks with real persisted instances:
- PostgresResource with associated Strand via postgres_resource let
- PostgresServer via create_postgres_server() helper with real Vm
- Real semaphore operations (incr_destroy) instead of yield stubs
- Proper FK chain: Project → PrivateSubnet → Vm → PostgresServer

The setup_postgres_resource() helper encapsulates the repeated fixture pattern, and create_postgres_server() handles the 10+ line setup for each server instance.

### DIFF
--- a/spec/prog/test/postgres_resource_spec.rb
+++ b/spec/prog/test/postgres_resource_spec.rb
@@ -7,8 +7,52 @@ RSpec.describe Prog::Test::PostgresResource do
 
   let(:test_project) { Project.create(name: "test-project") }
   let(:service_project) { Project.create(name: "service-project") }
-  let(:working_representative_server) { instance_double(PostgresServer, run_query: "DROP TABLE\nCREATE TABLE\nINSERT 0 10\n4159.90\n415.99\n4.1") }
-  let(:faulty_representative_server) { instance_double(PostgresServer, run_query: "") }
+  let(:location_id) { Location::HETZNER_FSN1_ID }
+
+  let(:private_subnet) {
+    PrivateSubnet.create(
+      name: "pg-subnet", project_id: test_project.id, location_id:,
+      net4: "172.0.0.0/26", net6: "fdfa:b5aa:14a3:4a3d::/64"
+    )
+  }
+
+  let(:timeline) { PostgresTimeline.create(location_id:) }
+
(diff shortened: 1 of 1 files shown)


spec/prog/test/postgres_resource_spec.rb
```diff
@@ -24,83 +68,91 @@
...
+      pgr_test.instance_variable_set(:@postgres_resource, nil)
+      project_id = pgr_test.strand.stack.first["postgres_test_project_id"]
+      PrivateSubnet.create(name: "subnet", project_id:, location_id:, net4: "10.0.0.0/26", net6: "fd00::/64")
       expect { pgr_test.wait_resources_destroyed }.to nap(5)
     end
 
     it "hops to destroy if the postgres resource destroyed" do
-      expect(pgr_test).to receive(:postgres_resource).and_return(nil)
+      pgr_test.instance_variable_set(:@postgres_resource, nil)
       expect { pgr_test.wait_resources_destroyed }.to hop("destroy")
```
fdr, line 139: this fix is weird, seems like it'd be better to delete that record.

spec/prog/test/postgres_resource_spec.rb
```diff
@@ -24,83 +68,91 @@
...
     it "naps if the postgres resource isn't deleted yet" do
-      expect(pgr_test).to receive(:postgres_resource).and_return(instance_double(PostgresResource))
+      setup_postgres_resource(with_server: false)
       expect { pgr_test.wait_resources_destroyed }.to nap(5)
     end
 
     it "naps if the private subnet isn't deleted yet" do
-      expect(pgr_test).to receive(:postgres_resource).and_return(nil)
-      expect(PrivateSubnet).to receive(:[]).and_return(instance_double(PrivateSubnet))
+      pgr_test.instance_variable_set(:@postgres_resource, nil)
```
fdr: this fix is weird, seems like it'd be better to delete (or never have had) that record.

PR #4548, opened 2026-01-14
Add etcd backups

mohi-kalantari:
Introduces the KubernetesEtcdBackup strand and associated logic to
    perform hourly etcd backups and upload them to a MinIO bucket.

    - Model: Add methods to KubernetesEtcdBackup model to check backup
    status, setup_bucket, and retrieve blob storage credentials.
    - Nexus: Create `Prog::Kubernetes::EtcdBackupNexus` to manage the
    backup lifecycle:
        - Sets up blob storage user and policy.
        - Creates the backup bucket.
        - Triggers the backup job periodically.
        - Passes MinIO credentials to the backup script.
    - Rhizome: Add backup-etcd script to:
        - Install `mc` (MinIO Client) idempotently.
        - Clean up old backups to prevent disk accumulation.
        - Capture etcd snapshot using `etcdctl` via `crictl`.
        - Upload the backup to the specified MinIO bucket.
        - Clean up the local backup file after upload.

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -67,6 +67,7 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
 
   label def start
     register_deadline("wait", 120 * 60)
+    Prog::Kubernetes::EtcdBackupNexus.assemble(kubernetes_cluster.id)
     incr_install_metrics_server
     incr_sync_worker_mesh
     incr_install_csi
@@ -346,6 +347,8 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
   label def destroy
     reap do
       decr_destroy
+
+      kubernetes_cluster.kubernetes_etcd_backup&.incr_destroy
 
       kubernetes_cluster.private_subnet.incr_destroy_if_only_used_internally(
         ubid: kubernetes_cluster.ubid,
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -15,6 +15,7 @@ class KubernetesCluster < Sequel::Model
   one_to_many :nodepools, class: :KubernetesNodepool, read_only: true
   one_to_many :active_billing_records, class: :BillingRecord, key: :resource_id, read_only: true, &:active
   many_to_one :location, read_only: true
+  one_to_one :kubernetes_etcd_backup
 
   dataset_module Pagination
 
--- a/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
+++ b/spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
@@ -164,6 +164,7 @@ RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
       expect(nx).to receive(:incr_sync_internal_dns_config)
       expect(nx).to receive(:incr_install_csi)
       expect { nx.start }.to hop("create_load_balancers")
+      expect(KubernetesEtcdBackup.first.kubernetes_cluster_id).to eq(kubernetes_cluster.id)
     end
   end
 
@@ -797,6 +798,7 @@ RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
     end
 
     it "triggers deletion of associated resources and completes destroy when nodepools are gone" do
+      KubernetesEtcdBackup.create(kubernetes_cluster_id: kubernetes_cluster.id, access_key: "access", secret_key: "secret", location_id: Location::HETZNER_FSN1_ID)
       st.update(prog: "Kubernetes::KubernetesClusterNexus", label: "destroy", stack: [{}])
       kubernetes_cluster.nodepools.first.destroy
       kubernetes_cluster.nodes.map(&:destroy)
@@ -806,6 +808,9 @@ RSpec.describe Prog::Kubernetes::KubernetesClusterNexus do
       expect(kubernetes_cluster.services_lb).to receive(:incr_destroy)
       expect(kubernetes_cluster.cp_vms).to all(receive(:incr_destroy))
       expect(kubernetes_cluster.nodes).to all(receive(:incr_destroy))
+
+      expect(kubernetes_cluster.kubernetes_etcd_backup).to receive(:incr_destroy)
+      kubernetes_cluster.kubernetes_etcd_backup.destroy
 
       expect(kubernetes_cluster.nodepools).to be_empty
 


migrate/20260113_add_kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:kubernetes_etcd_backup) do
+      column :id, :uuid, primary_key: true
+      column :access_key, :text, null: false
+      column :secret_key, :text, null: false
+      column :created_at, Time, null: false, default: Sequel::CURRENT_TIMESTAMP
```
jeremyevans: Not a problem, but it is a bit inconsistent to use `Time` for one column, and `timestamptz` for the other (they both result in the `timestamptz` database type). If you really need the column, since the other columns use symbol types, I recommend:

```suggestion
      column :created_at, :timestamptz, null: false, default: Sequel::CURRENT_TIMESTAMP
```

However, I don't see how this column adds any value. Unless you are sure you need it, I would remove it:

```suggestion
```

migrate/20260113_add_kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:kubernetes_etcd_backup) do
```
jeremyevans, line 5: This has no indexes other the the primary key one. Are you sure that is wise? I would think you would at least want one on `kubernetes_cluster_id` or `kubernetes_cluster_id, created_at DESC`.

If there should only be one of these per kubernetes cluster, then the `kubernetes_cluster_id` index should be unique. In that case, I would normally suggest that `id` be a foreign key to `kubernetes_cluster`, instead of having a separate `kubernetes_cluster_id` column, but since this will have it's own strand distinct from the kubernetes cluster strand, I can understand having a separate type for this.

cache/index.cache
jeremyevans, line 1: This and the other cache files should be in the previous commit that added the model.

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+  subject_is :kubernetes_etcd_backup
+
+  def self.assemble(kubernetes_cluster_id)
+    unless (kc = KubernetesCluster[kubernetes_cluster_id])
+      fail "KubernetesCluster does not exist"
+    end
+
+    DB.transaction do
+      id = KubernetesEtcdBackup.generate_uuid
+      KubernetesEtcdBackup.create_with_id(id,
```
jeremyevans: We should generally avoid the pattern where we use `generate_uuid` and then pass the id to `create_with_id` for the same model:

```suggestion
      kb = KubernetesEtcdBackup.create(
```

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+        kubernetes_cluster_id: kc.id,
+        access_key: SecureRandom.hex(16),
+        secret_key: SecureRandom.hex(32),
+        location_id: kc.location.id)
+      Strand.create_with_id(id, prog: "Kubernetes::EtcdBackupNexus", label: "setup_blob_storage")
```
jeremyevans:
```suggestion
      Strand.create_with_id(kb, prog: "Kubernetes::EtcdBackupNexus", label: "setup_blob_storage")
```

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    hop_wait
+  end
+
+  label def wait
+    if kubernetes_etcd_backup.need_backup?
+      hop_run_backup
+    end
+
+    nap 20 * 60
```
jeremyevans: I would prefer an approach where you do something like `nap(Time.now - kubernetes_etcd_backup.next_backup_time + 1).clamp(1, 3601)`, to avoid unnecessary wakeups.

migrate/20260113_add_kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,16 @@
...
+      column :id, :uuid, primary_key: true
```
jeremyevans: For most new uuid primary key types, we add a default value that uses `Sequel.lit("gen_random_ubid_uuid(XXX)")` with the appropriate integer (see other migrations using `gen_random_ubid_uuid`).

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+    job_name = "backup_etcd"
+    sshable = kubernetes_cluster.functional_nodes.first.vm.sshable
+    creds = {
+      "access_key" => kubernetes_etcd_backup.access_key,
+      "secret_key" => kubernetes_etcd_backup.secret_key,
+      "endpoint" => kubernetes_etcd_backup.blob_storage_endpoint,
+      "bucket" => kubernetes_etcd_backup.ubid,
+      "root_certs" => kubernetes_etcd_backup.blob_storage.root_certs
+    }
+    sshable.d_run(job_name, "kubernetes/bin/backup-etcd", stdin: JSON.generate(creds), log: false)
```
jeremyevans:
```suggestion
    sshable.d_run("backup_etcd", "kubernetes/bin/backup-etcd", stdin: JSON.generate(creds), log: false)
```

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    nap 20 * 60
+  end
+
+  label def run_backup
+    nap 20 * 60 unless kubernetes_cluster.strand.label == "wait"
+
+    kubernetes_etcd_backup.update(latest_backup_started_at: Time.now)
+
+    job_name = "backup_etcd"
```
jeremyevans: Probably not worth a local variable for this.

```suggestion
```

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def kubernetes_cluster
+    @kubernetes_cluster ||= kubernetes_etcd_backup.kubernetes_cluster
+  end
+
+  label def setup_blob_storage
+    admin_client.admin_add_user(kubernetes_etcd_backup.access_key, kubernetes_etcd_backup.secret_key)
+    admin_client.admin_policy_add(kubernetes_etcd_backup.ubid, kubernetes_etcd_backup.blob_storage_policy)
+    admin_client.admin_policy_set(kubernetes_etcd_backup.ubid, kubernetes_etcd_backup.access_key)
```
jeremyevans, line 29: Is this idempotent? If `admin_policy_set` results in a temporary error, will the next strand run still work correctly? Similar question for the `destroy` label.

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,68 @@
...
+require_relative "../../common/lib/util"
+require "fileutils"
+require "json"
+
+input = JSON.parse($stdin.read)
+
+MC_PATH = "/usr/local/bin/mc"
+unless File.exist?(MC_PATH)
+  puts "Installing mc..."
+  r "curl -sL https://dl.min.io/client/mc/release/linux-amd64/mc -o #{MC_PATH}"
```
jeremyevans: No hash checking!?

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,68 @@
...
+end
+
+BACKUP_DIR = "/var/lib/etcd/backups"
+FileUtils.mkdir_p(BACKUP_DIR)
+
+puts "Cleaning up old backups."
+r "sudo rm -f #{File.join(BACKUP_DIR, "etcd-snapshot-*.db")}"
+
+backup_file = "etcd-snapshot-#{Time.now.strftime("%Y%m%d%H%M%S")}.db"
+host_backup_path = File.join(BACKUP_DIR, backup_file)
```
jeremyevans: This should be moved later in the program. Ideally, we do not delete a previous backup until we have written a new backup. We should consider whether we need to keep the backup on the local system after uploading to MinIO.

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,68 @@
...
+
+puts "Found etcd container: #{etcd_container_id}"
+puts "Taking snapshot..."
+
+cmd = "ETCDCTL_API=3 etcdctl snapshot save /var/lib/etcd/backups/#{backup_file} " \
+      "--cacert=/etc/kubernetes/pki/etcd/ca.crt " \
+      "--cert=/etc/kubernetes/pki/etcd/server.crt " \
+      "--key=/etc/kubernetes/pki/etcd/server.key"
+
+r "sudo crictl exec #{etcd_container_id} /bin/sh -c '#{cmd}'"
```
jeremyevans: In general, this is an unsafe pattern. I think it happens to be safe in this particular usage because `cmd` does not contain characters that need escaping.

Probably would be best to change `r` to accepting multiple arguments and passing all of them to `Open3.capture3`, then change this to:

```suggestion
r(*%W[sudo crictl exec #{etcd_container_id} /bin/sh -c #{cmd}])
```

This will pass the arguments as separate arguments to `sudo`, instead of executing the command string using a shell.

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,68 @@
...
+
+  env_vars = ""
+  if input["root_certs"]
+    cert_path = "/tmp/minio-root-ca.crt"
+    File.write(cert_path, input["root_certs"])
+    env_vars = "SSL_CERT_FILE=#{cert_path}"
+  end
+
+  alias_name = "minio-target"
+  r "#{env_vars} #{MC_PATH} alias remove #{alias_name} > /dev/null 2>&1 || true"
```
jeremyevans: If we change `r` as recommended above, you can pass an environment hash as the first argument to `r`, instead of passing a string.

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,68 @@
...
+r "sudo chown ubi:ubi #{host_backup_path}"
+
+if File.exist?(host_backup_path)
+  puts "Backup successfully created at #{host_backup_path}"
+  puts "Uploading backup to MinIO..."
+
+  env_vars = ""
+  if input["root_certs"]
+    cert_path = "/tmp/minio-root-ca.crt"
+    File.write(cert_path, input["root_certs"])
```
jeremyevans: Writing to a file with a fixed name at a world-writable location is a security antipattern. Use `Tempfile`.

spec/model/kubernetes/kubernetes_etcd_backup_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      name: "test-cluster",
+      version: Option.kubernetes_versions.first,
+      location_id: location.id,
+      project_id: project.id,
+      private_subnet_id: private_subnet.id,
+      cp_node_count: 1,
+      target_node_size: "standard-2"
+    )
+    vm = Prog::Vm::Nexus.assemble("public key", project.id, name: "cp", private_subnet_id: kc.private_subnet.id).subject
+    Sshable.create_with_id(vm.id)
```
jeremyevans:
```suggestion
    Sshable.create_with_id(vm)
```

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      name: "test",
+      version: Option.kubernetes_versions.first,
+      location_id: location.id,
+      project_id: project.id,
+      private_subnet_id: private_subnet.id,
+      cp_node_count: 1,
+      target_node_size: "standard-2"
+    )
+    vm = Prog::Vm::Nexus.assemble("public key", project.id, name: "cp", private_subnet_id: kc.private_subnet.id).subject
+    Sshable.create_with_id(vm.id)
```
jeremyevans:
```suggestion
    Sshable.create_with_id(vm)
```

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      expect {
+        described_class.assemble(SecureRandom.uuid)
+      }.to raise_error RuntimeError, "KubernetesCluster does not exist"
+    end
+
+    it "creates kubernetes etcd backup and strand" do
+      described_class.assemble(kc.id)
+
+      keb = KubernetesEtcdBackup.last
+      strand = Strand[keb.id]
```
jeremyevans:
```suggestion
      strand = keb.strand
```

model/kubernetes/kubernetes_etcd_backup.rb
```diff
@@ -3,19 +3,64 @@
 require_relative "../../model"
 
 class KubernetesEtcdBackup < Sequel::Model
+  one_to_one :strand, key: :id
+  many_to_one :location
   many_to_one :kubernetes_cluster
```
jeremyevans: Unless you are going to be using setter methods for these, add the `read_only: true` option to each.

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+    it "calls setup_bucket on model and hops" do
+      expect(client).to receive(:create_bucket)
+      expect(client).to receive(:set_lifecycle_policy)
+      expect { nx.setup_bucket }.to hop("wait")
+    end
+  end
+
+  describe "#wait" do
+    it "hops to run_backup if backup needed" do
+      allow(nx.kubernetes_etcd_backup).to receive(:need_backup?).and_return(true)
```
jeremyevans: Mock `_cmd` (or at least `d_check`) on the underlying sshable instead of `need_backup?` (here and the next spec). In general, you should not mock database access, only ssh/http, and mock at the lowest level possible.

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    it "naps if backup not needed" do
+      allow(nx.kubernetes_etcd_backup).to receive(:need_backup?).and_return(false)
+      expect { nx.wait }.to nap(20 * 60)
+    end
+  end
+
+  describe "#run_backup" do
+    it "naps if cluster is not in wait state" do
+      Strand.create_with_id(kc.id, prog: "Kubernetes::KubernetesClusterNexus", label: "starting")
```
jeremyevans:
```suggestion
      Strand.create_with_id(kc, prog: "Kubernetes::KubernetesClusterNexus", label: "starting")
```

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      expect { nx.run_backup }.to nap(20 * 60)
+    end
+
+    it "runs backup command and hops to wait" do
+      Strand.create_with_id(kc.id, prog: "Kubernetes::KubernetesClusterNexus", label: "wait")
```
jeremyevans:
```suggestion
      Strand.create_with_id(kc, prog: "Kubernetes::KubernetesClusterNexus", label: "wait")
```

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+        log: false
+      )
+
+      expect { nx.run_backup }.to hop("wait")
+    end
+
+    it "updates latest_backup_started_at" do
+      Strand.create_with_id(kc.id, prog: "Kubernetes::KubernetesClusterNexus", label: "wait")
+
+      allow(nx.kubernetes_cluster.functional_nodes.first.vm.sshable).to receive(:d_run)
```
jeremyevans: Only use `allow` in a before hook or `let`. In an individual spec, should should know whether or not the method should be called.

```suggestion
      expect(nx.kubernetes_cluster.functional_nodes.first.vm.sshable).to receive(:d_run)
```

aamederen:
My initial feeling:
1. Control plane and data plane flows feel generally OK
2. We don't need a new model and a nexus for managing backups specifically. I'd go like this
    1. Setup the blob part (MinIO, R2, etc.) at the beginning
    2. Spawn a non-nexus K8sBackup prog periodically in KubernetesCluster nexus
    3. Page if a backup prog takes long to finish or fails some validations
    4. Tear down the blob during cluster destroy

A semi unrelated thought, for Postgres, backups are a very key part for all operations. In K8s, it's more of a safety net.

A question: what's the expected maximum size for an etcd backup?

migrate/20260113_add_kubernetes_etcd_backup.rb
mohi-kalantari: Yeah I missed that one. added the indexes

mohi-kalantari:
@aamederen 

We need to store per cluster credentials and I don't like the idea of storing those values on other k8s related models or on the strand stack frames, that's the reason I thought a separate model and strand is needed for the job. even if we store it on the strand stack, the secret key cannot be encrypted and would be stored plain text in database.

regarding the size of clusters we have provisioned so far, probably tens of megabytes.

rhizome/kubernetes/bin/backup-etcd
mohi-kalantari: I changed the logic to keep the last 7 backups.

prog/kubernetes/etcd_backup_nexus.rb
mohi-kalantari: yes both labels are idempotent using put and delete request and their api design in minio is idempotent.

migrate/20260113_add_kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,18 @@
...
+      # UBID.to_base32_n("kb") => 619
+      column :id, :uuid, primary_key: true, default: Sequel.lit("gen_random_ubid_uuid(619)")
+      column :access_key, :text, null: false
+      column :secret_key, :text, null: false
+      column :latest_backup_started_at, :timestamptz
+
+      foreign_key :location_id, :location, type: :uuid, null: false
+      foreign_key :kubernetes_cluster_id, :kubernetes_cluster, type: :uuid, null: false
+
+      index [:location_id, :kubernetes_cluster_id]
```
jeremyevans: Are you sure you want one composite index instead of two normal indexes? If so, can you explain why? If your intent was to create separate indexes per column, you want to use:

```suggestion
      index :location_id
      index :kubernetes_cluster_id
```

model/kubernetes/kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,73 @@
...
+    sshable = kubernetes_cluster.functional_nodes.first.vm.sshable
+    status = sshable.d_check("backup_etcd")
+    return true if ["Failed", "NotStarted"].include?(status)
+    return true if status == "Succeeded" && (latest_backup_started_at.nil? || latest_backup_started_at < Time.now - 60 * 60 * 1)
+
+    false
+  end
+
+  def next_backup_time
+    return Time.now if latest_backup_started_at.nil?
```
jeremyevans: If there is no blob_storage, I think we should use a time further in the future. Otherwise, we'll be checking every second. Also, unless you care about the `nil`/`false` difference, it's preferred to use `unless object` instead of  `if object.nil?`

```suggestion
    return Time.now + 86400 unless blob_storage
    return Time.now unless latest_backup_started_at
```

model/kubernetes/kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,73 @@
...
+  many_to_one :location, read_only: true
+  many_to_one :kubernetes_cluster, read_only: true
+
+  plugin ResourceMethods, encrypted_columns: :secret_key
+  plugin SemaphoreMethods, :destroy
+
+  BACKUP_BUCKET_EXPIRATION_DAYS = 7
+
+  def need_backup?
+    return false if blob_storage.nil?
```
jeremyevans:
```suggestion
    return false unless blob_storage
```

model/kubernetes/kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,73 @@
...
+    return false if kubernetes_cluster.functional_nodes.empty?
+
+    sshable = kubernetes_cluster.functional_nodes.first.vm.sshable
+    status = sshable.d_check("backup_etcd")
+    return true if ["Failed", "NotStarted"].include?(status)
+    return true if status == "Succeeded" && (latest_backup_started_at.nil? || latest_backup_started_at < Time.now - 60 * 60 * 1)
```
jeremyevans: Probably better to use `case` for this. Also, multiplying by `1` has no effect (not sure whether you meant to add `1` or not).

```suggestion
    case sshable.d_check("backup_etcd")
    when "Failed", "NotStarted"
      true
    when "Succeeded"
      latest_backup_started_at.nil? || latest_backup_started_at < Time.now - 60 * 60
    else
      false
    end
```

model/kubernetes/kubernetes_etcd_backup.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def next_backup_time
+    return Time.now if latest_backup_started_at.nil?
+
+    latest_backup_started_at + 60 * 60
+  end
+
+  def blob_storage
+    @blob_storage ||= MinioCluster[project_id: Config.minio_service_project_id, location_id: location.id]
```
jeremyevans: This doesn't cache a negative lookup. You need to use a different approach for that:

```suggestion
    return @blob_storage if defined?(@blob_storage)
    @blob_storage = MinioCluster[project_id: Config.minio_service_project_id, location_id: location.id]
```

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,97 @@
...
+require "fileutils"
+require "json"
+require "digest"
+require "tempfile"
+
+input = JSON.parse($stdin.read)
+
+MC_PATH = "/usr/local/bin/mc"
+MC_RELEASE_URL = "https://dl.min.io/client/mc/release/linux-amd64/mc"
+MC_CHECKSUM_URL = "https://dl.min.io/client/mc/release/linux-amd64/mc.sha256sum"
```
jeremyevans: The purpose of hash checking is to ensure the downloaded file hasn't been switched for something malicious. If someone with access to `dl.min.io` can change the `mc` file, they can change the `mc.sha256sum` file to match. You want to download a versioned release file, where you know the file (and hence hash) will not change.

You need to hardcode the expected hash in the code, and check that it matches that. Please review how we handle pretty much all other downloads, such as https://github.com/ubicloud/ubicloud/blob/main/rhizome/host/lib/vhost_block_backend.rb#L11

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,97 @@
...
+      r "mv #{tmp_mc.path} #{MC_PATH}"
+      r "chmod +x #{MC_PATH}"
+    end
+  end
+end
+
+BACKUP_DIR = "/var/lib/etcd/backups"
+FileUtils.mkdir_p(BACKUP_DIR)
+
+puts "Cleaning up old backups."
```
jeremyevans: I would move cleanup to the end of the program.

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,97 @@
...
+
+cmd = "ETCDCTL_API=3 etcdctl snapshot save /var/lib/etcd/backups/#{backup_file} " \
+      "--cacert=/etc/kubernetes/pki/etcd/ca.crt " \
+      "--cert=/etc/kubernetes/pki/etcd/server.crt " \
+      "--key=/etc/kubernetes/pki/etcd/server.key"
+
+r(*%W[sudo crictl exec #{etcd_container_id} /bin/sh -c #{cmd}])
+
+puts "Changing permission from root to ubi."
+r "sudo chown ubi:ubi #{host_backup_path}"
```
jeremyevans: I recommend using the separate arguments to `r` in every case where you are currently interpolating.

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,97 @@
...
+
+if File.exist?(host_backup_path)
+  puts "Backup successfully created at #{host_backup_path}"
+  puts "Uploading backup to MinIO..."
+
+  env_vars = ""
```
jeremyevans:
```suggestion
  env_vars = {}
```

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,97 @@
...
+  root_cert_file = nil
+  if input["root_certs"]
+    root_cert_file = Tempfile.new("minio-root-ca.crt")
+    root_cert_file.write(input["root_certs"])
+    root_cert_file.close
+    env_vars = "SSL_CERT_FILE=#{root_cert_file.path}"
```
jeremyevans:
```suggestion
    env_vars["SSL_CERT_FILE"] = root_cert_file.path
```

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  begin
+    alias_name = "minio-target"
+    r "#{env_vars} #{MC_PATH} alias remove #{alias_name} > /dev/null 2>&1 || true"
+    r "#{env_vars} #{MC_PATH} alias set #{alias_name} #{input["endpoint"]} #{input["access_key"]} #{input["secret_key"]}"
+    r "#{env_vars} #{MC_PATH} cp #{host_backup_path} #{alias_name}/#{input["bucket"]}/#{backup_file}"
```
jeremyevans: Best to pass `env_vars` as a hash instead of relying on the shell to set the environment variables. For the first entry here, you have to use a shell, or pass the `expect` argument to handle the failing case. 

```suggestion
    r env_vars, "#{MC_PATH} alias remove #{alias_name} > /dev/null 2>&1 || true"
    r env_vars, *%W[#{MC_PATH} alias set #{alias_name} #{input["endpoint"]} #{input["access_key"]} #{input["secret_key"]}]
    r env_vars, *%W[#{MC_PATH} cp #{host_backup_path} #{alias_name}/#{input["bucket"]}/#{backup_file}]
```

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+
+    context "when backup is not needed" do
+      before do
+        expect(kubernetes_etcd_backup.kubernetes_cluster.functional_nodes.first.vm.sshable).to receive(:d_check).with("backup_etcd").and_return("Succeeded")
+        kubernetes_etcd_backup.update(latest_backup_started_at: Time.now)
+      end
+
+      it "naps for the difference between next_backup_time and now + 1" do
+        now = Time.now
+        expect(Time).to receive(:now).and_return(now).at_least(:once)
```
jeremyevans: This is shared in all specs in this context. I would move it to the before (you can use an instance variable or `let` for `now`).

jeremyevans (commented):
All my comments have been addressed. I'll leave approval to Eren.

model/kubernetes/kubernetes_etcd_backup.rb
```diff
@@ -3,9 +3,63 @@
...
   plugin ResourceMethods, encrypted_columns: :secret_key
+  plugin SemaphoreMethods, :destroy
+
+  BACKUP_BUCKET_EXPIRATION_DAYS = 7
+
+  def need_backup?
+    return false unless blob_storage
+    return false if kubernetes_cluster.functional_nodes.empty?
+
+    sshable = kubernetes_cluster.functional_nodes.first.vm.sshable
```
aamederen: can we use `kubernetes_cluster.sshable` here, which already performs `functional_nodes.first.sshable` inside?

rhizome/kubernetes/bin/backup-etcd
```diff
@@ -0,0 +1,93 @@
...
+  begin
+    alias_name = "minio-target"
+    r env_vars, "#{MC_PATH} alias remove #{alias_name} > /dev/null 2>&1 || true"
+    r env_vars, *%W[#{MC_PATH} alias set #{alias_name} #{input["endpoint"]} #{input["access_key"]} #{input["secret_key"]}]
+    r env_vars, *%W[#{MC_PATH} cp #{host_backup_path} #{alias_name}/#{input["bucket"]}/#{backup_file}]
+    puts "Backup uploaded successfully."
+  ensure
+    root_cert_file&.unlink
+  end
+
```
aamederen: nit: extra space before `else`

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+      "bucket" => kubernetes_etcd_backup.ubid,
+      "root_certs" => kubernetes_etcd_backup.blob_storage.root_certs
+    }
+    sshable.d_run("backup_etcd", "kubernetes/bin/backup-etcd", stdin: JSON.generate(creds), log: false)
+
+    hop_wait
+  end
+
+  label def destroy
+    admin_client.admin_remove_user(kubernetes_etcd_backup.access_key)
```
aamederen, line 69: a `decr_destroy` and `deadline_set` could be good here as we do in many nexus programs

spec/prog/kubernetes/etcd_backup_nexus_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+  let(:kubernetes_etcd_backup) {
+    KubernetesEtcdBackup.create(
+      kubernetes_cluster_id: kc.id,
+      access_key: "access",
+      secret_key: "secret",
+      location_id: location.id
+    )
+  }
+
+  let(:cluster) {
```
aamederen: nit: naming-wise this can be ambiguous as we use cluster with both MinIO and Kubernetes. In this case `cluster` referring to minio cluster might be confusing.

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -66,6 +66,7 @@ def billing_rate_for(type, family)
   end
 
   label def start
+    Prog::Kubernetes::EtcdBackupNexus.assemble(kubernetes_cluster.id)
     register_deadline("wait", 120 * 60)
```
aamederen, line 69: any particular reason putting `EtcdBackupNexus.assemble` before `register_deadline`? If not, maybe we can keep register_deadline at the top, to catch any potential exceptions in EtcdBackupNexus.assemble to be caught by the deadline.

prog/kubernetes/etcd_backup_nexus.rb
```diff
@@ -0,0 +1,82 @@
...
+    hop_setup_bucket
+  end
+
+  label def setup_bucket
+    kubernetes_etcd_backup.setup_bucket
+
+    hop_wait
+  end
+
+  label def wait
```
aamederen, line 40: we should implement some kind of pager if a successful backup is not found in the blob storage for some time, similar to the `MissingBackup` page logic in `PostgresTimelineNexus`. This can be done with a subsequent PR.

aamederen (approved):
Thank you!

PR #4550, opened 2026-01-14
Add admin page to find archived records

enescakir:
I often find myself querying the database to find archived records.

This is especially painful when working with short-lived resources such as GitHub runners. Being able to look up archived records in the admin panel using an ID from logs would be extremely helpful and save a lot of time.

ArchivedRecord table has index on `archived_at` and `model_name` columns, so it's important to filter by those columns in any query on this table.

CleanShot 2026-01-14 at 17 09 14@2x

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -121,6 +121,17 @@ class CloverAdmin < Roda
     end
   end
 
+  def available_classes
+    classes = []
+    Sequel::Model.subclasses.each do |c|
+      classes << c if c < ResourceMethods::InstanceMethods
+      c.subclasses.each do |sc|
+        classes << sc if sc < ResourceMethods::InstanceMethods
+      end
+    end
+    classes.sort_by!(&:name)
+  end
+
   plugin :rodauth, route_csrf: true do
     enable :argon2, :login, :logout, :webauthn, :change_password
     accounts_table :admin_account
@@ -560,6 +571,25 @@ class CloverAdmin < Roda
       autoforme
     end
 
+    r.get "archived-record-by-id" do
+      @id = if (uuid = typecast_params.uuid("id"))
+        UBID.from_uuidish(uuid)
+      elsif (ubid = typecast_params.ubid("id"))
+        UBID.parse(ubid)
+      elsif typecast_params.nonempty_str("id")
+        fail CloverError.new(400, "InvalidRequest", "Invalid UBID or UUID provided")
+      end
+      @model_name = typecast_params.nonempty_str("model_name") || UBID.class_for_ubid(@id.to_s)&.name
+      @days = (typecast_params.pos_int("days") || 15).clamp(1, 60)
+      @classes = available_classes
+      @record = if @id
+        fail CloverError.new(400, "InvalidRequest", "Could not determine model name from ID") unless @model_name
+        ArchivedRecord.find_by_id(@id.to_uuid, model_name: @model_name, days: @days)
+      end
+
+      view("archived_record_by_id")
+    end
+
     r.get "vm-by-ipv4" do
       if (@ips_param = typecast_params.nonempty_str("ips"))
         ips = @ips_param.split(",").filter_map {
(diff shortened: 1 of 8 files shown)


clover_admin.rb
```diff
@@ -121,6 +121,15 @@ def linkify_ubids(body)
     end
   end
 
+  def available_classes
+    Sequel::Model
+      .subclasses
+      .map { [it, it.subclasses] }
+      .flatten
+      .select { it < ResourceMethods::InstanceMethods }
+      .sort_by(&:name)
```
jeremyevans: Looks like this is just moving code. However, I think we should consider this implementation, which should result in the same behavior, with significantly fewer array allocations:

```suggestion
    classes = []
    Sequel::Model.subclasses.each do |c|
      classes << c if c < ResourceMethods::InstanceMethods
      c.subclasses.each do |sc|
        classes << sc if sc < ResourceMethods::InstanceMethods
      end
    end
    classes.sort_by!(&:name)
```

views/admin/archived_record_by_id.erb
```diff
@@ -0,0 +1,25 @@
+<% @page_title = "Find Archived Record by ID" %>
+
+<p>
+  This page allows you to find archived records by UBID or UUID. Useful for investigating deleted resources.
+</p>
+
+<% form(id: "archived_record_form", class: "form-vertical") do |f| %>
+  <%== f.input("text", key: "id", label: "ID", placeholder: "Enter UBID or UUID", value: @id) %>
+  <%== f.input("select", key: "model_name", label: "Model Name", options: [["From ID", ""]] + @classes.map { [it.name, it.name] }, value: @model_name) %>
```
jeremyevans: I think this should work. `add_blank` accepts a string, and if option text is the same as value, you should be able to use a string instead of an array where both elements are the same:

```suggestion
  <%== f.input("select", key: "model_name", label: "Model Name", add_blank: "From ID", options: @classes.map(&:name) value: @model_name) %>
```

views/admin/archived_record_by_id.erb
```diff
@@ -0,0 +1,25 @@
...
+  <%== f.input("number", key: "days", label: "Days", placeholder: "1-60", value: @days, min: 1, max: 60) %>
+  Higher day values will increase query duration significantly. Use the smallest value possible.
+  <%== f.button("Find Archived Record") %>
+<% end %>
+
+<% if @id %>
+  <% data = if @record
+      {model_name: @record[:model_name], archived_at: @record[:archived_at]}
+        .merge(@record[:model_values])
```
jeremyevans:
```suggestion
        .merge!(@record[:model_values])
```

views/admin/archived_record_by_id.erb
```diff
@@ -0,0 +1,25 @@
...
+        .map { |k, v| {"Key" => k, "Value" => v} }
+ else
+      {}
```
jeremyevans: This is unneeded, as the table component supports `nil` data:

```suggestion
```

views/admin/archived_record_by_id.erb
```diff
@@ -0,0 +1,25 @@
...
+  <%== f.input("text", key: "id", label: "ID", placeholder: "Enter UBID or UUID", value: @id) %>
+  <%== f.input("select", key: "model_name", label: "Model Name", options: [["From ID", ""]] + @classes.map { [it.name, it.name] }, value: @model_name) %>
+  <%== f.input("number", key: "days", label: "Days", placeholder: "1-60", value: @days, min: 1, max: 60) %>
+  Higher day values will increase query duration significantly. Use the smallest value possible.
+  <%== f.button("Find Archived Record") %>
+<% end %>
+
+<% if @id %>
+  <% data = if @record
+      {model_name: @record[:model_name], archived_at: @record[:archived_at]}
```
jeremyevans: With removal of else block:

```suggestion
  <% if @record
      data = {model_name: @record[:model_name], archived_at: @record[:archived_at]}
```

enescakir:
> We may want to consider an index on `archived_record` for `[model_name, id]` (with the `id` extracted from json). That should avoid the need to specify the number of days, and make for much faster lookups. That slows down model destroys slightly, but if we are doing these kind of lookups on a regular basis, I think it may be worth it.

I use it regularly, but I’m not sure about others. Let’s merge this and see if people use it often and complain about performance. If that happens, we can invest in adding new indexes.

PR #4553, opened 2026-01-14
Add default_pg_config to postgres config endpoint

serprex:
Overview of resource with api lacked computed configuration values, only showing explicit overrides user has made


heavycrystal:

cli-commands/pg/post/show-computed-config.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("show-computed-config") do
```
jeremyevans: This is currently untested. Probably easiest to add a golden file spec. Add `pg eu-central-h1/test-pg show-computed-config` to `spec/routes/api/cli/golden-file-commands/success.txt` (where we are testing `show-config` and `show-pgbouncer-config`), run `rspec spec/routes/api/cli/golden_files_spec.rb`, then run `rake update_golden_files`.

routes/project/location/postgres.rb
```diff
@@ -518,7 +518,8 @@ class Clover
 
           {
             pg_config: pg.user_config,
-            pgbouncer_config: pg.pgbouncer_user_config
+            pgbouncer_config: pg.pgbouncer_user_config,
+            computed_config: pg.representative_server&.configure_hash&.[](:configs)
```
byucesoy: I think `default_pg_config` would be better name.

byucesoy (approved):
Apart from the naming, I don't have any objections from PG side. If it looks good to @jeremyevans from API/CLI/SDK perspective, we can merge this.

PR #4554, opened 2026-01-14
Add Sshable#json and use it

enescakir:
This method implements a common pattern we use for parsing the output of SSH commands.

I was inspired by Jeremy’s Sshable#write_file helper.

### DIFF
--- a/spec/model/sshable_spec.rb
+++ b/spec/model/sshable_spec.rb
@@ -263,6 +263,13 @@ LOCK
     end
   end
 
+  describe "#cmd_json" do
+    it "parses cmd output as JSON" do
+      expect(sa).to receive(:_cmd).with("cat data.json").and_return('{"key": "value"}')
+      expect(sa.cmd_json("cat data.json")).to eq({"key" => "value"})
+    end
+  end
+
   describe "daemonizer methods" do
     let(:unit_name) { "test_unit" }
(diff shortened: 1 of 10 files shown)


spec/model/sshable_spec.rb
```diff
@@ -263,6 +263,13 @@ def simulate(cmd:, exit_status:, exit_signal:, stdout:, stderr:)
     end
   end
 
+  describe "#json" do
+    it "parses cmd output as JSON" do
+      expect(sa).to receive(:cmd).with("echo json", foo: "bar").and_return('{"key": "value"}')
```
jeremyevans: Tests are failing because you are not allowed to override `cmd` in the specs, only `_cmd`, to prevent invalid mocking hiding places where you passing interpolated strings to `cmd`:

```suggestion
      expect(sa).to receive(:_cmd).with("echo json", foo: "bar").and_return('{"key": "value"}')
```

enescakir:
> I like this. I recommend changing the method name from `json` to `json_cmd` or `cmd_json` (or potentially something else), but I'll leave it up to you (you can keep it `json` if you think that is best).

I replaced it with `cmd_json`. I chose this over `json_cmd` because when a developer types `sshable.cm…`, autocomplete suggests both `.cmd` and `.cmd_json`, which feels more natural and discoverable.

PR #4556, opened 2026-01-14
Decrease day limits in archived record search page

enescakir:
The page can fetch archived records from the last 15 days in ~4 seconds.

When the number of days is increased, query time grows much faster than linearly. As a result, the request can exceed Heroku’s 30-second router timeout, causing a timeout error on the UI while the query continues running in the background.


enescakir:
> Clamping it to 15 days max and still allowing lower values for faster results would be more flexible (4s still seems like a really long time). However, I'm fine removing it.

I changed the default to 5 and max to 15

jeremyevans (approved):

PR #4557, opened 2026-01-14
Avoid string interpolation in query in PostgresServer#check_pulse

jeremyevans:


model/postgres/postgres_server.rb
```diff
@@ -198,16 +198,20 @@ def failover_target
...
     elsif standby?
-      "pg_last_wal_receive_lsn()"
+      "pg_last_wal_receive_lsn"
     else
-      "pg_last_wal_replay_lsn()"
+      "pg_last_wal_replay_lsn"
     end
   end
 
+  def lsn_function
```
byucesoy: Does anything use this function anymore? If not, we can remove this.

byucesoy (approved):

model/postgres/postgres_server.rb
jeremyevans: Doesn't look like anything uses it, I'll add a commit to remove it, and rebase to fix the conflict.

PR #4558, opened 2026-01-14
Demock postgres_timeline_nexus_spec

fdr:
Replace instance_double(PostgresTimeline) and association mocks with real persisted models via helper methods for MinioCluster, PostgresResource, PostgresServer, and LocationCredential.

Ref: clover-mol-k7h3

### DIFF
--- a/spec/prog/postgres/postgres_timeline_nexus_spec.rb
+++ b/spec/prog/postgres/postgres_timeline_nexus_spec.rb
@@ -3,32 +3,135 @@
 require_relative "../../model/spec_helper"
 
 RSpec.describe Prog::Postgres::PostgresTimelineNexus do
-  subject(:nx) { described_class.new(Strand.new(id: "8148ebdf-66b8-8ed0-9c2f-8cfe93f5aa77")) }
+  subject(:nx) { described_class.new(st) }
 
-  let(:postgres_timeline) {
-    instance_double(
-      PostgresTimeline,
-      id: "b253669e-1cf5-8ada-9337-5fc319690838",
-      ubid: "ptp99pd7gwyp4jcvnzgrsd443g",
-      aws_s3_policy_name: "ptp99pd7gwyp4jcvnzgrsd443g",
-      blob_storage: instance_double(MinioCluster, url: "https://blob-endpoint", root_certs: "certs"),
-      blob_storage_endpoint: "https://blob-endpoint",
-      blob_storage_client: instance_double(Minio::Client),
-      access_key: "dummy-access-key",
-      secret_key: "dummy-secret-key",
-      blob_storage_policy: {"Version" => "2012-10-17", "Statement" => [{"Action" => ["s3:GetBucketLocation"], "Effect" => "Allow", "Principal" => {"AWS" => ["*"]}, "Resource" => ["arn:aws:s3:::test"], "Sid" => ""}]},
-      aws?: false
+  let(:project) { Project.create(name: "test-project") }
+  let(:postgres_timeline) { create_postgres_timeline }
+  let(:st) { postgres_timeline.strand }
+  let(:service_project) { Project.create(name: "postgres-service-project") }
+  let(:location_id) { Location::HETZNER_FSN1_ID }
+
+  let(:private_subnet) { create_private_subnet }
+
+  def create_minio_cluster(location_id: self.location_id, project_id: service_project.id)
+    mc = MinioCluster.create(
+      location_id:,
+      project_id:,
+      name: "minio-cluster-test",
+      admin_user: "admin",
+      admin_password: "secret",
+      root_cert_1: "certs",
+      private_subnet_id: private_subnet.id
     )
-  }
+    Strand.create_with_id(mc, prog: "Minio::MinioClusterNexus", label: "wait")
+    mc
+  end
+
+  def create_postgres_timeline(location_id: self.location_id)
+    tl = PostgresTimeline.create(
+      location_id:,
+      access_key: "dummy-access-key",
+      secret_key: "dummy-secret-key"
+    )
+    Strand.create_with_id(tl, prog: "Postgres::PostgresTimelineNexus", label: "start")
+    tl
+  end
+
+  def create_postgres_resource(location_id: self.location_id)
+    pr = PostgresResource.create(
+      name: "pg-test-resource",
+      superuser_password: "dummy-password",
+      ha_type: "none",
+      target_version: "16",
+      location_id:,
+      project:,
+      user_config: {},
+      pgbouncer_user_config: {},
+      target_vm_size: "standard-2",
+      target_storage_size_gib: 64,
+      root_cert_1: "root_cert_1",
+      root_cert_2: "root_cert_2",
+      server_cert: "server_cert",
+      server_cert_key: "server_cert_key"
+    )
+    Strand.create_with_id(pr, prog: "Postgres::PostgresResourceNexus", label: "wait")
+    pr
+  end
+
+  def create_postgres_server(resource:, timeline:, timeline_access: "push", representative: true, version: "16", strand_label: "wait", vm: nil, location_id: self.location_id, subnet: private_subnet)
+    vm ||= Prog::Vm::Nexus.assemble_with_sshable(
+      project.id, name: "pg-vm-test", private_subnet_id: subnet.id,
+      location_id:, unix_user: "ubi"
+    ).subject
+    VmStorageVolume.create(vm:, boot: false, size_gib: 64, disk_index: 1)
+    server = PostgresServer.create(
+      timeline:,
+      resource:,
+      vm_id: vm.id,
+      representative_at: representative ? Time.now : nil,
+      synchronization_status: "ready",
+      timeline_access:,
+      version:
+    )
+    Strand.create_with_id(server, prog: "Postgres::PostgresServerNexus", label: strand_label)
+    server
+  end
+
+  def create_private_subnet(
+    name: "pg-subnet",
+    location_id: self.location_id,
+    net4: "172.0.0.0/26",
+    net6: "fdfa:b5aa:14a3:4a3d::/64"
+  )
+    PrivateSubnet.create(name:, project:, location_id:, net4:, net6:)
+  end
+
+  def create_aws_location
+    loc = Location.create(
+      name: "us-west-2",
+      display_name: "AWS US West 2",
+      ui_name: "aws-us-west-2",
+      visible: true,
+      provider: "aws"
+    )
+    LocationCredential.create_with_id(loc,
+      access_key: "access-key-id",
+      secret_key: "secret-access-key")
+    loc
+  end
+
+  def backup_fixture(days_ago:)
+    Struct.new(:key, :last_modified).new(
+      "basebackups_005/base_backup_stop_sentinel.json",
+      Time.now - days_ago * 24 * 60 * 60
+    )
+  end
+
+  def mock_minio_client(methods = {})
+    client = instance_double(Minio::Client, methods)
+    expect(Minio::Client).to receive(:new).and_return(client)
+    client
+  end
+
+  def mock_admin_minio_client(minio_cluster)
+    client = instance_double(Minio::Client)
+    expect(Minio::Client).to receive(:new).with(
+      endpoint: minio_cluster.ip4_urls.first,
+      access_key: minio_cluster.admin_user,
+      secret_key: minio_cluster.admin_password,
+      ssl_ca_data: minio_cluster.root_cert_1
+    ).and_return(client)
+    client
+  end
 
   before do
-    allow(nx).to receive(:postgres_timeline).and_return(postgres_timeline)
+    allow(Config).to receive(:postgres_service_project_id).and_return(service_project.id)
   end
 
   describe ".assemble" do
     it "throws an exception if parent is not found" do
       expect {
-        described_class.assemble(location_id: Location::HETZNER_FSN1_ID, parent_id: "69c0f4cd-99c1-8ed0-acfe-7b013ce2fa0b")
+        described_class.assemble(location_id: Location::HETZNER_FSN1_ID, parent_id: PostgresResource.generate_uuid)
       }.to raise_error RuntimeError, "No existing parent"
     end
 
@@ -41,12 +144,11 @@ RSpec.describe Prog::Postgres::PostgresTimelineNexus do
     it "creates postgres timeline" do
       st = described_class.assemble(location_id: Location::HETZNER_FSN1_ID)
 
-      postgres_timeline = PostgresTimeline[st.id]
-      expect(postgres_timeline).not_to be_nil
+      expect(st.subject).to exist
     end
 
     it "does not generate access_key/secret_key when AWS & Config.aws_postgres_iam_access" do
-      allow(Config).to receive(:aws_postgres_iam_access).and_return(true)
+      expect(Config).to receive(:aws_postgres_iam_access).and_return(true).twice
 
       tl = described_class.assemble(location_id: Location::HETZNER_FSN1_ID).subject
       expect(tl.access_key).not_to be_nil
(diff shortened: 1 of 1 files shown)


spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -75,192 +156,308 @@
...
     end
   end
 
   describe "#setup_bucket" do
     it "hops to wait_leader if bucket is created" do
-      expect(postgres_timeline).to receive(:create_bucket).and_return(true)
-      expect(postgres_timeline).to receive(:set_lifecycle_policy).and_return(true)
+      create_minio_cluster
+      blob_storage_client = instance_double(Minio::Client)
+      allow(Minio::Client).to receive(:new).and_return(blob_storage_client)
```
fdr: this seems it should be expect (possible at least once) otherwise it can be a dead+misleading fixture easily

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -75,192 +156,308 @@
...
-      expect(postgres_timeline).to receive(:created_at).and_return(Time.now - 11 * 24 * 60 * 60)
+      create_minio_cluster
+      resource = create_postgres_resource
+      server = create_postgres_server(resource:, timeline: postgres_timeline, strand_label: "wait")
+      server.destroy
+      postgres_timeline.update(created_at: Time.now - 11 * 24 * 60 * 60)
+
+      blob_storage_client = instance_double(Minio::Client)
+      allow(Minio::Client).to receive(:new).and_return(blob_storage_client)
+      allow(blob_storage_client).to receive(:list_objects).and_return([])
```
fdr: same: if we expect the fixture to be called (e.g. it's not in a before block) then allow is asking for a mistake.

fdr:
fixup was generated by pipeline tweak, so....hopefully less in the future, too

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+  let(:private_subnet) {
+    PrivateSubnet.create(
+      name: "pg-subnet", project:, location_id:,
+      net4: "172.0.0.0/26", net6: "fdfa:b5aa:14a3:4a3d::/64"
     )
   }
 
+  def create_minio_cluster(loc_id: location_id, proj_id: service_project.id)
+    mc_id = MinioCluster.generate_uuid
+    mc = MinioCluster.create_with_id(mc_id,
```
jeremyevans: Claude's behavior for this type of thing is inconsistent. This spec has 3 different ways, and only one of them does what we want.

```suggestion
    mc = MinioCluster.create(
```

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+      location_id: loc_id,
+      project_id: proj_id,
+      name: "minio-cluster-#{SecureRandom.hex(4)}",
+      admin_user: "admin",
+      admin_password: "secret",
+      root_cert_1: "certs",
+      private_subnet_id: private_subnet.id)
+    Strand.create_with_id(mc_id, prog: "Minio::MinioClusterNexus", label: "wait")
```
jeremyevans:
```suggestion
    Strand.create_with_id(mc, prog: "Minio::MinioClusterNexus", label: "wait")
```

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+    mc
+  end
+
+  def create_postgres_timeline(loc_id: location_id)
+    tl_id = PostgresTimeline.generate_uuid
+    tl = PostgresTimeline.create_with_id(tl_id,
```
jeremyevans:
```suggestion
    tl = PostgresTimeline.create(
```

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+      location_id: loc_id,
+      access_key: "dummy-access-key",
+      secret_key: "dummy-secret-key")
+    Strand.create_with_id(tl_id, prog: "Postgres::PostgresTimelineNexus", label: "start")
```
jeremyevans:
```suggestion
    Strand.create_with_id(tl, prog: "Postgres::PostgresTimelineNexus", label: "start")
```

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+    tl
+  end
+
+  def create_postgres_resource(loc_id: location_id)
+    pr = PostgresResource.create(
```
jeremyevans, line 41: Claude handled this case correctly.

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+  end
+
+  def create_postgres_server(resource:, timeline:, timeline_access: "push", representative: true, version: "16", strand_label: "wait", vm: nil, loc_id: location_id, subnet: private_subnet)
+    vm ||= Prog::Vm::Nexus.assemble_with_sshable(
+      project.id, name: "pg-vm-#{SecureRandom.hex(4)}", private_subnet_id: subnet.id,
+      location_id: loc_id, unix_user: "ubi"
+    ).subject
+    VmStorageVolume.create(vm:, boot: false, size_gib: 64, disk_index: 1)
+    server_id = PostgresServer.generate_uuid
+    server = PostgresServer.create_with_id(server_id,
```
jeremyevans:
```suggestion
    server = PostgresServer.create(
```

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+      timeline:,
+      resource:,
+      vm_id: vm.id,
+      representative_at: representative ? Time.now : nil,
+      synchronization_status: "ready",
+      timeline_access:,
+      version:)
+    Strand.create_with_id(server_id, prog: "Postgres::PostgresServerNexus", label: strand_label)
```
jeremyevans:
```suggestion
    Strand.create_with_id(server, prog: "Postgres::PostgresServerNexus", label: strand_label)
```

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+      name: "us-west-2",
+      display_name: "AWS US West 2",
+      ui_name: "aws-us-west-2",
+      visible: true,
+      provider: "aws"
+    )
+    LocationCredential.create(
+      access_key: "access-key-id",
+      secret_key: "secret-access-key"
+    ) { it.id = loc.id }
```
jeremyevans: This is the third way Claude handed this type of change.

```suggestion
    LocationCredential.create_with_id(loc,
      access_key: "access-key-id",
      secret_key: "secret-access-key"
    )
```

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,26 +3,102 @@
...
+      location_id: loc_id,
+      access_key: "dummy-access-key",
+      secret_key: "dummy-secret-key")
+    Strand.create_with_id(tl_id, prog: "Postgres::PostgresTimelineNexus", label: "start")
+    tl
+  end
+
+  def create_postgres_resource(loc_id: location_id)
+    pr = PostgresResource.create(
+      name: "pg-test-#{SecureRandom.hex(4)}",
```
jeremyevans: It seems like a bad idea to use a random name here. We should use a fixed name, maybe settable by keyword argument. While 4 bytes does make collisions unlikely, we should use an approach that makes collisions impossible.  Similar issue elsewhere we are using this pattern.

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -75,192 +156,300 @@
...
+      resource = create_postgres_resource
+      create_postgres_server(resource:, timeline: postgres_timeline, strand_label: "wait")
+
+      blob_storage_client = instance_double(Minio::Client)
+      expect(Minio::Client).to receive(:new).and_return(blob_storage_client)
+
+      backup = Struct.new(:key, :last_modified).new("basebackups_005/base_backup_stop_sentinel.json", Time.now - 3 * 24 * 60 * 60)
+      expect(blob_storage_client).to receive(:list_objects).and_return([backup])
+
+      expect(nx.postgres_timeline.leader.vm.sshable).to receive(:_cmd).and_return("NotStarted")
```
jeremyevans: Probably worth adding `.with(...)` so we are testing which arguments we are passing to `_cmd`. Maybe not in every case (if the arguments are the same in every case), but at least in one place.

jeremyevans (approved):

fdr:
overhauled some context to unify and give it a different idea, which is to reference the ResourceMethods abstraction. We'll see if it holds up.

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -100,7 +100,7 @@ def create_aws_location
   describe ".assemble" do
     it "throws an exception if parent is not found" do
       expect {
-        described_class.assemble(location_id: Location::HETZNER_FSN1_ID, parent_id: "69c0f4cd-99c1-8ed0-acfe-7b013ce2fa0b")
+        described_class.assemble(location_id: Location::HETZNER_FSN1_ID, parent_id: PostgresResource.generate_ubid.to_uuid)
```
jeremyevans: Should use `generate_uuid` instead of `generate_ubid.to_uuid`.

fdr: putting aside ubids are always preferred over uuid in any remotely similar case, when should we use entropy vs when shouldn't we? isn't it fine to use a pregenerated id?

jeremyevans: The behavior is the same in both cases, using `generate_uuid` is just simpler. This is the definition in `ResourceMethods`:

```ruby
    def generate_uuid
      generate_ubid.to_uuid
    end
```

jeremyevans: To answer your other question, I don't object to keeping the hardcoded uuid string, but I do think it's better to generate one.

fdr: why is that?

fdr: (I think  static ubid > generation > static uuid, but not by a whole lot, except that last one, ubids are much better given the type bytes)

jeremyevans: When I see a static string uuid/ubid, I assume it's a magic value that has some sort of meaning. With a `generate_uuid`/`generate_ubid` call, I know that the specific value doesn't matter.

fdr: alright, I'll encode this rationale into the contexts, thank you

fdr: alright, done that way in [a9ccf3f](https://github.com/ubicloud/ubicloud/pull/4558/commits/a9ccf3f06f18d5e1ad33394d6a4b6130cbaa2c30), plus some context engineering (which one day I think can be reusable, once it has a more formed up regression test model) to keep it that way.

fdr:
~~interesting, I ran my full toolchain and it found some more stuff.~~ seems two files got conflated in this one somehow, backed out that change.

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -3,32 +3,107 @@
...
+  let(:location_id) { Location::HETZNER_FSN1_ID }
+
+  let(:private_subnet) {
+    PrivateSubnet.create(
+      name: "pg-subnet", project:, location_id:,
+      net4: "172.0.0.0/26", net6: "fdfa:b5aa:14a3:4a3d::/64"
     )
   }
 
+  def create_minio_cluster(loc_id: location_id, proj_id: service_project.id)
```
jeremyevans: Seems weird to use `loc_id` and `proj_id` instead of `location_id` and `project_id` as the argument names. Maybe done because `location_id` as default value would cause problems?

```suggestion
  def create_minio_cluster(location_id: self.location_id, project_id: service_project.id)
```

Same issue in other helper methods.

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -75,192 +155,300 @@
 
     describe "when blob storage is aws s3" do
       it "creates user and policies and hops" do
+        aws_location = create_aws_location
+        postgres_timeline.update(location_id: aws_location.id)
+        resource = create_postgres_resource(loc_id: aws_location.id)
+        aws_private_subnet = PrivateSubnet.create(
```
jeremyevans: Kind of odd that there isn't a `create_private_subnet` helper method that this and `let(:private_subnet)` calls.

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -75,192 +155,300 @@
...
-      expect(postgres_timeline).to receive(:backups).and_return([instance_double(backup, last_modified: Time.now - 3 * 24 * 60 * 60)])
-      expect(postgres_timeline).to receive(:leader).and_return(instance_double(PostgresServer))
+      create_minio_cluster
+      resource = create_postgres_resource
+      create_postgres_server(resource:, timeline: postgres_timeline, strand_label: "wait")
+
+      blob_storage_client = instance_double(Minio::Client)
+      expect(Minio::Client).to receive(:new).and_return(blob_storage_client)
+
+      backup = Struct.new(:key, :last_modified).new("basebackups_005/base_backup_stop_sentinel.json", Time.now - 3 * 24 * 60 * 60)
```
jeremyevans: Might be worth a helper method that accepts a number of days, and use it in this spec and following specs.

spec/prog/postgres/postgres_timeline_nexus_spec.rb
```diff
@@ -64,8 +138,14 @@
 
     describe "when blob storage is minio" do
       it "creates user and policies and hops" do
-        expect(postgres_timeline).to receive(:blob_storage).and_return(instance_double(MinioCluster, url: "https://blob-endpoint", root_certs: "certs", admin_user: "admin", admin_password: "secret")).at_least(:once)
-        expect(Minio::Client).to receive(:new).with(endpoint: "https://blob-endpoint", access_key: "admin", secret_key: "secret", ssl_ca_data: "certs").and_return(admin_blob_storage_client)
+        minio_cluster = create_minio_cluster
+
+        expect(Minio::Client).to receive(:new).with(
```
jeremyevans: Might be worth a helper method for this, and use it this spec and another one below.

fdr:
@jeremyevans alright, applied. I will have to come up with a new pass, some of these are not in the scope of "just do it" "refinements" as I have a directory of them called, that I can run as a battery. More like "propositions". One thing the LLM proposed was some cross-file helpers could condense things quite a bit, but I've been staying away from that until the work is done.

fdr:
okay, squashed, git diff 02c5de65c 048b5780d empty

fdr:
...and now a rebase so the CI run is most useful...

jeremyevans:
> ...and now a rebase so the CI run is most useful...

I found out recently that the way GitHub Actions works is it tests using a merge of the PR branch to the base branch. The current PR run is testing 4e85a6ddd875a4846938ad5262971dffbf18b143, not 235a178f8ffa1bcc5bf3ce0f61ffb5e10bce06b3: https://github.com/ubicloud/ubicloud/actions/runs/21221095855/job/61055753926?pr=4558#step:3:65

fdr:
> > ...and now a rebase so the CI run is most useful...
> 
> I found out recently that the way GitHub Actions works is it tests using a merge of the PR branch to the base branch. The current PR run is testing [4e85a6d](https://github.com/ubicloud/ubicloud/commit/4e85a6ddd875a4846938ad5262971dffbf18b143), not [235a178](https://github.com/ubicloud/ubicloud/commit/235a178f8ffa1bcc5bf3ce0f61ffb5e10bce06b3): https://github.com/ubicloud/ubicloud/actions/runs/21221095855/job/61055753926?pr=4558#step:3:65

oh weird. I...what?

PR #4561, opened 2026-01-15
Extract shared requestBody for load balancer creation

fdr:
The createLocationLoadBalancer endpoint had an inline requestBody definition. Extract it to components/requestBodies/LoadBalancer for reuse across load balancer endpoints.

Based on https://github.com/ubicloud/ubicloud/pull/3533

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -768,43 +768,7 @@ paths:
       operationId: createLocationLoadBalancer
       summary: Create a new load balancer
       requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              type: object
-              properties:
-                algorithm:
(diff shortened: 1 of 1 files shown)


fdr:
cc @geemus if this was your intent of #3533, it obsoletes it.

enescakir (approved):

geemus:
@fdr this does appear to be correct. That said, the original PR was predicated on that request body actually being reused (once for load balancer in a project and once for load balancer in a location within a project). In this updated PR, there is no longer re-use (I guess underlying stuff must have changed or gone away). Without re-use, it's not obvious to me that there is much value to doing this vs just leaving it inline. So it might make sense to just close the PRs, your call ultimately though.

PR #4562, opened 2026-01-15
Use authorization.domain for ACME DNS challenge records

fdr:
When requesting a certificate with multiple SANs, ACME returns one authorization per domain. Each authorization object contains an identifier field (RFC 8555 Section 7.1.4) specifying which domain it authorizes. The acme-client gem exposes this as authorization.domain.

The previous implementation assumed there was no way to tie an authorization to its domain, so it created DNS challenge records for every authorization across every domain (a cross-product). With 2 domains this created 4 records instead of 2.

This change uses authorization.domain to create exactly one DNS challenge record per authorization, for the correct domain.

Based on https://github.com/ubicloud/ubicloud/tree/jeremy-private-name-ssl-cert

### DIFF
--- a/spec/prog/vnet/cert_nexus_spec.rb
+++ b/spec/prog/vnet/cert_nexus_spec.rb
@@ -16,10 +16,19 @@ RSpec.describe Prog::Vnet::CertNexus do
   let(:client) { instance_double(Acme::Client) }
   let(:account_key) { Clec::Cert.ec_key }
 
-  def setup_order(new_order: false, with_order: true)
+  def setup_order(new_order: false, with_order: true, add_private: false)
     dns_challenge = instance_double(Acme::Client::Resources::Challenges::DNS01, record_name: "test-record-name", record_type: "test-record-type", record_content: "test-record-content")
-    authorization = instance_double(Acme::Client::Resources::Authorization, dns: dns_challenge)
-    order = instance_double(Acme::Client::Resources::Order, authorizations: [authorization], url: "test-order-url") if with_order
+    # Each authorization includes a domain attribute per RFC 8555 Section 7.1.4
+    authorization = instance_double(Acme::Client::Resources::Authorization, dns: dns_challenge, domain: cert.hostname)
+    authorizations = [authorization]
+
+    if add_private
+      private_dns_challenge = instance_double(Acme::Client::Resources::Challenges::DNS01, record_name: "test-record-name", record_type: "test-record-type", record_content: "test-record-content-private")
+      private_authorization = instance_double(Acme::Client::Resources::Authorization, dns: private_dns_challenge, domain: "private-#{cert.hostname}")
+      authorizations << private_authorization
+    end
+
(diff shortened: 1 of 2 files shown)


fdr:
I was suspicious of this, so I had claude research the matter, it elected to read the RFC and the acme client source code. There is a `domain` field and I superficially gave this a shot in staging.  It worked and appeared to not generate the cross product of dns records implied by:

```
      dns_challenge = authorization.dns
      # With multiple authorizations for separate DNS challenge records, there doesn't
      # seem to be a way to tie an authorization to a specific DNS name. So add all
      # authorizations to all names we are requesting.
      each_dns_record_name(dns_challenge) do |record_name|
        dns_zone.insert_record(record_name:, type: dns_challenge.record_type, ttl: 600, data: dns_challenge.record_content)
      end
```

fdr:
"Just an Idea" since you went through the debugging for the original and Saw Things. I did superficially validate it through staging.

fdr:
also the test bloat is inartful I think?

jeremyevans:
I'll do a full review tomorrow, but I'm in favor of the general idea.

spec/prog/vnet/cert_nexus_spec.rb
```diff
@@ -37,6 +46,7 @@ def setup_order(new_order: false, with_order: true)
   def use_add_private(identifiers: [])
     st.stack = [{"restarted" => 0, "add_private" => true}]
     nx.instance_variable_set(:@frame, nil)
+    nx.instance_variable_set(:@acme_order, nil)
```
jeremyevans, line 49: `@frame` is eager initialized, so to refresh it, you need to set the instance variable to `nil`. `@acme_order` is lazy initialized, so I don't think this is needed.

spec/prog/vnet/cert_nexus_spec.rb
```diff
@@ -183,13 +208,29 @@ def use_add_private(identifiers: [])
...
+      dns_challenge = instance_double(Acme::Client::Resources::Challenges::DNS01, record_name: "test-record-name", record_type: "test-record-type", record_content: "test-record-content")
+      authorization = instance_double(Acme::Client::Resources::Authorization, dns: dns_challenge, domain: cert.hostname)
+      private_dns_challenge = instance_double(Acme::Client::Resources::Challenges::DNS01, record_name: "test-record-name", record_type: "test-record-type", record_content: "test-record-content-private")
+      private_authorization = instance_double(Acme::Client::Resources::Authorization, dns: private_dns_challenge, domain: "private-#{cert.hostname}")
+      order = instance_double(Acme::Client::Resources::Order, authorizations: [authorization, private_authorization], url: "test-order-url")
+
+      fresh_client = instance_double(Acme::Client)
+      expect(Acme::Client).to receive(:new).and_return(fresh_client)
+      expect(fresh_client).to receive(:order).with(url: "test-order-url").and_return(order)
+      cert.update(order_url: "test-order-url", account_key: account_key.to_der, kid: "x")
```
jeremyevans, line 227: Seems like we should be using `setup_order` here.

PR #4564, opened 2026-01-15
Cap ext4 root reservation to 50GiB

serprex:
Reserving more than 50GiB needlessly restricts customer disk usage


jeremyevans (commented):
Recalling support incidents I've been involved in, I'm not in favor of this change. If the customer does end up in a low disk space situation (a not-infrequent occurrence), this change would leave a much smaller margin of error.

serprex:
#4332 completely eradicated support cases about disk full in my experience

furkansahin (approved):

PR #4567, opened 2026-01-15
Allow accepting/declining project invitations

jeremyevans:
Previously, project invitations were not invitations, they were force
joins. Even if a user did not want to join the project, they had no
choice. When they created their account (or immediately if they were
already a user), they were force joined to the project.

This adds a Project Invitations table on the Projects page, if the
account has any outstanding invitations. They can choose to either
accept the invitation or decline it.

To make this easier, add a components/submit_button template that wraps
the components/button template in a form, so it can be used like
components/delete_button.

A few minor related changes in their own commits:

* Simplify ExpireProjectInvitations 
* Allow audit_log to accept project_id argument (necessary as `@project`
  is not set in the accept/decline routes).

### DIFF
--- a/spec/routes/web/auth_spec.rb
+++ b/spec/routes/web/auth_spec.rb
@@ -152,6 +152,7 @@ RSpec.describe Clover, "auth" do
 
     visit verify_link
     expect(page.title).to eq("Ubicloud - Verify Account")
+    expect(Account.first.default_project.name).to eq "Default"
 
     click_button "Verify Account"
     expect(page.title).to eq("Ubicloud - Projects")
--- a/views/project/index.erb
+++ b/views/project/index.erb
@@ -12,11 +12,16 @@
 <div class="grid gap-6">
   <%== part(
       "components/table_card",
-      headers: %w[Name Remove\ Access],
+      headers: ["Name", "Set As Default", "Remove Access"],
       rows: @projects.map do |p|
          [
            [
              [p.name, {link: "#{p.path}/dashboard"}],
+             if current_account.default_project == p
+               ["Default"]
+             else
+               [part("components/submit_button", action: "/project/set-default/#{p.ubid}", text: "Set As Default"), {escape: false}]
+             end,
              [
                part("components/delete_button",
                  url: "/project/#{p.ubid}/user/#{current_account.ubid}",
--- a/routes/project.rb
+++ b/routes/project.rb
@@ -35,6 +35,20 @@ class Clover
       end
     end
 
+    r.post web?, "set-default", :ubid_uuid do |project_id|
+      no_authorization_needed
+      no_audit_log
+
+      if (project = current_account.projects_dataset.with_pk(project_id))
+        current_account.default_project = project
+        flash["notice"] = "Default project updated"
+      else
+        flash["error"] = "Invalid default project selected"
+      end
+
+      r.redirect "/project"
+    end
+
     r.on web?, "invitation", :ubid_uuid do |project_id|
       invitation = current_account.invitations_dataset.first(project_id:)
       check_found_object(invitation)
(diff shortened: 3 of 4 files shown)


spec/routes/web/auth_spec.rb
```diff
@@ -94,7 +94,7 @@
     expect(Mail::TestMailer.deliveries.length).to eq 3
   end
 
-  it "can create new account, verify it, and visit project which invited" do
+  it "can create new account, verify it, but without force join to invited project" do
     p = Project.create(name: "Invited-project")
     p.add_invitation(email: TEST_USER_EMAIL, inviter_id: "bd3479c6-5ee3-894c-8694-5190b76f84cf", expires_at: Time.now + 7 * 24 * 60 * 60)
```
enescakir: IMHO, we can remove "but without force join to invited project" and `p.add_invitation` part. It tests non-existent code.

spec/routes/web/auth_spec.rb
```diff
@@ -117,7 +117,7 @@
     expect(page.title).to eq("Ubicloud - Default Dashboard")
 
     visit "#{p.path}/dashboard"
-    expect(page.title).to eq("Ubicloud - Invited-project Dashboard")
+    expect(page.title).to eq("Ubicloud - ResourceNotFound")
```
enescakir: Instead of checking invited project, we can be sure it's redirected to the default project.

spec/routes/web/auth_spec.rb
```diff
@@ -134,7 +134,7 @@
     expect(page).to have_content("Could not create account. Please ensure JavaScript is enabled and access to Cloudflare is not blocked, then try again.")
   end
 
-  it "can create new account, verify it, and visit project which invited with default policy" do
+  it "can create new account, verify it, but not force join to invited project" do
```
enescakir: This test is also not meaningful

helpers/general.rb
```diff
@@ -190,11 +190,6 @@ def before_rodauth_create_account(account, name)
   def after_rodauth_create_account(account_id)
     account = Account[account_id]
     account.create_project_with_default_policy("Default")
-    account.invitations.each do |inv|
```
enescakir, line 193: After an account is created, if the user has any project invitations, we can redirect them to the project invitations page instead of the dashboard to encourage them to accept or reject the invitations.

views/project/index.erb
```diff
@@ -28,3 +30,26 @@
...
+        "components/table_card",
+        title: "Project Invitations",
+        headers: %w[Name ID Invited\ By Accept Decline],
+        rows: @invitations.map do |inv|
+           project_ubid = inv.project.ubid
+           [
+             [
+               inv.project.name,
+               project_ubid,
+               inv.inviter.email,
```
enescakir: If the inviter account is deleted, this might raise an exception.

views/project/index.erb
```diff
@@ -28,3 +30,26 @@
       }
     ) %>
 </div>
+
+<% unless @invitations.empty? %>
```
enescakir, line 48: After an account is logged in, if the user has any project invitations, we might show a notice indicating they have pending invitations to encourage them to accept or reject them.

spec/routes/web/auth_spec.rb
jeremyevans: That's already checked three lines above. So it's probably best to remove this.

views/project/index.erb
jeremyevans: I think redirecting the user to the page showing the invitations after login should be sufficient. If you disagree, can you let me know where we should show this notice?

spec/routes/web/auth_spec.rb
jeremyevans: Turns out it is meaningful and provides coverage testing. It should be renamed, though.

jeremyevans:
Pushed four new commits to address comments.

PR #4570, opened 2026-01-15
Allow users to choose which project is their default project

jeremyevans:
Previously, the default project was always the oldest project the user had access to. This allows the user to choose which project is the default.

This changes behavior for new users, making their "Default" project their default project. So if they are invited to a project created before they created their account, their default project will not switch to the invited project.

This doesn't affect existing accounts, until the accounts choose to set a default project.

### DIFF
--- a/routes/project.rb
+++ b/routes/project.rb
@@ -35,6 +35,20 @@ class Clover
       end
     end
 
+    r.post web?, "set-default", :ubid_uuid do |project_id|
+      no_authorization_needed
+      no_audit_log
+
+      if (project = current_account.projects_dataset.with_pk(project_id))
+        current_account.default_project = project
+        flash["notice"] = "Default project updated"
+      else
+        flash["error"] = "Invalid default project selected"
+      end
+
+      r.redirect "/project"
+    end
+
     r.on web?, "invitation", :ubid_uuid do |project_id|
       invitation = current_account.invitations_dataset.first(project_id:)
       check_found_object(invitation)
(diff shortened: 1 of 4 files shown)


helpers/general.rb
```diff
@@ -189,7 +189,7 @@ def before_rodauth_create_account(account, name)
 
   def after_rodauth_create_account(account_id)
     account = Account[account_id]
-    account.create_project_with_default_policy("Default")
+    account.default_project = account.create_project_with_default_policy("Default")
```
enescakir, line 192: Do we need to call `save_changes` for this?
Should we add a new expectation to test ensure that newly created accounts have this set by default?

migrate/20260115_account_default_project.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:account_default_project) do
```
enescakir, line 5: I was wondering why you preferred using a separate table instead of adding a `default_project_id` column, especially since all new accounts will have this value set.

helpers/general.rb
jeremyevans: > Do we need to call `save_changes` for this?

No. `many_to_one` setter methods only set the column, but `one_to_one` and `one_through_one` setter methods persist immediately.

> Should we add a new expectation to test ensure that newly created accounts have this set by default?

Sure, I can add that.

migrate/20260115_account_default_project.rb
jeremyevans: If you add it to the main table, that means all or almost all account retrievals have to pay the cost for it. This is only used in a single place, so it didn't seem worth it to add it to the account table.

helpers/general.rb
jeremyevans: Added an expectation for that.

PR #4572, opened 2026-01-16
Display workflow job for runner in admin panel

enescakir:
We added `workflow_job` to `redacted_columns` because it contains a lot of data and was cluttering logs and pry output.

However, since the admin panel relies on `inspect_values` to display object details, this also hides `workflow_job` information there.

We can render it as an extra view in the admin panel. This has an added benefit, presenting the data in a structured table makes it much easier to read compared to raw JSON.

CleanShot 2026-01-16 at 15 51 56@2x

### DIFF
--- a/views/admin/extras/GithubRunner.erb
+++ b/views/admin/extras/GithubRunner.erb
@@ -1,0 +1,7 @@
+<%# locals: (obj: nil) %>
+
+<%== part("components/table",
+  title: "Workflow Job",
+  table_class: "workflow-job-table",
+  data: obj.workflow_job&.map { |k, v| {"Column" => k, "Value" => v} }
+) %>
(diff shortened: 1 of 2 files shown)


views/admin/extras/GithubRunner.erb
```diff
@@ -0,0 +1,7 @@
+<%# locals: (obj: nil) %>
```
jeremyevans, line 1: I think this should be required, since if the keywords argument is not passed, you get a `NoMethodError`.

```suggestion
<%# locals: (obj:) %>
```

This looks to be a general problem with all of the admin extra views.

enescakir: I will create a PR to replace all of them

PR #4573, opened 2026-01-16
Stop metrics export when fenced

shikharbhardwaj:
Previously a fenced primary continued exporting metrics with label role=primary. This created duplicate time series and incorrect metrics like disk usage on the overview page.


prog/postgres/postgres_server_nexus.rb
```diff
@@ -683,6 +684,7 @@ def before_run
 
     vm.sshable.cmd("sudo postgres/bin/restart :version", version:)
     vm.sshable.cmd("sudo systemctl restart pgbouncer@*.service")
+    vm.sshable.cmd("sudo systemctl restart postgres-metrics.timer")
```
furkansahin, line 687: why do you need this one?

shikharbhardwaj: We unfence (which causes a restart) after a failed upgrade. Not doing this would stop metrics from working.

PR #4574, opened 2026-01-16
Retry on CIDR collision

serprex:
Provisioning has failed with random CIDR allocation multiple times, /24 in private range isn't sufficiently large enough to rely on prayer

Detects collisions reactively instead of checking catalog in case subnet collision due to resource outside controlplane view, which is the case with external private link setup

### DIFF
--- a/prog/vnet/aws/nic_nexus.rb
+++ b/prog/vnet/aws/nic_nexus.rb
@@ -33,13 +33,27 @@ class Prog::Vnet::Aws::NicNexus < Prog::Base
       [subnet.subnet_id, subnet.availability_zone]
     elsif subnet_response.subnets.empty?
       subnet_az = az_to_provision_subnet
-      subnet_id = client.create_subnet({
-        vpc_id:,
-        cidr_block: NetAddr::IPv4Net.new(nic.private_ipv4.network, NetAddr::Mask32.new(24)).to_s,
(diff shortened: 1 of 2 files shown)


serprex:
NIC destroy also ignores errors, which will leave potential conflicts outside view of controlplane's catalog

furkansahin (changes requested):
The right way to do this is passing cidr_size 24 to `random_private_ipv4` call in NicNexus.assemble for `aws` subnets. After that, `random_private_ipv4` should already handle the conflicts.

serprex:
@furkansahin that won't work when the conflicting subnet is outside controlplane's knowledge

furkansahin:
I see, in your use-case, customers (or you) introduce extra resources to Ubicloud managed VPC? You should probably not do that since ubicloud controlplane may decide to start destroying VPC. In any case, we should combine both of the solutions then, because passing the cidr size will definitely lower the chances of conflicts

iamKunalGupta:
I think we are already doing it in https://github.com/ubicloud/ubicloud/blob/15b8427a0e399c232fd4357dd84450c5b3b7474f/prog/vnet/nic_nexus.rb#L19
and https://github.com/ubicloud/ubicloud/blob/3e509637c5fbd9a0e1e3a67605844c521d44417f/model/private_subnet.rb#L106-L113 should select /24 for the /16 vpc

serprex:
fixed by #4667

PR #4575, opened 2026-01-16
Add some PG Configuration guardrails

shikharbhardwaj:
- Set allow_alter_system-off for PG >= 17

- Disallow modification of certain PG settings

This commit removes a few settings from being overridden by the users,
mainly to restrict disabling key features like TLS or password
authentication setup for pgbouncer.

- Allow recovering from bad configuration settings

### DIFF
--- a/lib/validation/postgres_config_validator_schema.rb
+++ b/lib/validation/postgres_config_validator_schema.rb
@@ -732,6 +732,20 @@ module Validation
         type: :bool,
         default: "off"
       },
+      "huge_page_size" => {
+        description: "The size of huge page that should be requested.",
+        type: :string,
+        pattern: /\A[0-9]+(kB|MB|GB|TB)?\z/,
+        default: "0",
+        deny: true
+      },
+      "huge_pages" => {
+        description: "Use of huge pages on Linux or Windows.",
+        type: :enum,
+        allowed_values: ["try", "on", "off"],
+        default: "on",
+        deny: true
+      },
       "icu_validation_level" => {
         description: "Log level for reporting invalid ICU locale strings.",
         type: :enum,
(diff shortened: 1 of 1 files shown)


serprex (approved):

fdr:
huh, they got around to adding disabling alter system? how about that. some day they'll support two passwords for one role (for gradual rotation)

prog/postgres/postgres_server_nexus.rb
```diff
@@ -677,6 +682,11 @@ def before_run
   end
 
   label def restart
+    when_configure_set? do
```
furkansahin, line 689: So, we push this prog to restart. In case I `incr_configure` and get out of this state, and finally end-up in `wait` we do not ever pop. Is that correct? If so, I think we are introducing a bug here that will go untraced for a long while with a very long stack

lib/validation/postgres_config_validator_schema.rb
```diff
@@ -2015,23 +2016,28 @@ class PostgresConfigValidatorSchema
         type: :integer,
         default: 6432,
         min: 1,
-        max: 65535
+        max: 65535,
```
byucesoy, line 2040: Wouldn’t it effectively behave as “deny” if we simply didn’t include these configs in the schema to begin with? What’s the reason for explicitly setting deny instead? AFAIK, this is how we prevent setting up some other configs in the past.

prog/postgres/postgres_server_nexus.rb
byucesoy: Good catch :+1:

byucesoy (approved):

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: Thanks for the catch! I changed this to just pop from the restart label if configure is set, so that it can be retried later when configure completes.

lib/validation/postgres_config_validator_schema.rb
shikharbhardwaj: This returns a different error back to the user (unknown config vs not allowed), makes the behavior clearer.

byucesoy: Should we use this for other configs we don't allow then?

shikharbhardwaj: Makes sense, I added back the shared memory configs that we removed earlier with `deny: true`

shikharbhardwaj:
I checked for the implications of disabling ALTER SYSTEM for extensions: the standard practice for extension upgrades does not involve ALTER SYSTEM, and none of the extensions we currently install by default get impacted by this.

PR #4576, opened 2026-01-16
Elide location_id to keyword shorthand in assemble calls

fdr:
Ruby 3.1+ allows `key:` as shorthand for `key: value` when a same-named binding exists in scope. This spec has `let(:location_id)` returning `Location::HETZNER_FSN1_ID`, so the 11 call sites passing that constant can use the shorter form.

The shorthand reduces visual noise and clarifies intent: these calls pass through the test fixture's location rather than specifying an independent value.

### DIFF
--- a/spec/prog/postgres/postgres_resource_nexus_spec.rb
+++ b/spec/prog/postgres/postgres_resource_nexus_spec.rb
@@ -94,7 +94,7 @@ RSpec.describe Prog::Postgres::PostgresResourceNexus do
 
     it "validates input" do
       expect {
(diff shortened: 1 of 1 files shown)


jeremyevans (approved):

jeremyevans:
Note that Ruby 3.1+ allows `key:` as shorthand for `key: key`. The same-name binding does not have to exist in the scope. It will call `method_missing` or `const_missing` if it doesn't, just as it would for `key: key` (or `Key: Key` for constant references).

fdr:
I suppose "any scope" but I suppose the issue of `method_missing` et al complicates the issue; is a binding generated on the spot "in scope?" I suppose to a ruby internals hacker, no.

fdr:
I revised the context to be more ruby-correct

PR #4579, opened 2026-01-16
Add basic testing of all models on admin site

jeremyevans:
I used Claude to generate a separate spec file for each model on the admin site. It did most of the work, and I handled the cleanup.

This will allow us to catch breakage for model pages on the admin site before deploying the changes. No existing model pages need to be fixed, but I tested that this would have caught cases that resulted in previous production exceptions.

### DIFF
--- a/spec/routes/web/admin/model/invoice_spec.rb
+++ b/spec/routes/web/admin/model/invoice_spec.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require_relative "spec_helper"
+
+RSpec.describe CloverAdmin, "Invoice" do
+  include AdminModelSpecHelper
+
+  before do
+    @instance = create_invoice
+    admin_account_setup_and_login
+  end
+
+  it "displays the Invoice instance page correctly" do
+    click_link "Invoice"
+    expect(page.status_code).to eq 200
+    expect(page.title).to eq "Ubicloud Admin - Invoice - Browse"
+
+    expect(Aws::S3::Client).to receive(:new).and_return(Aws::S3::Client.new(stub_responses: true))
+    click_link @instance.admin_label
+    expect(page.status_code).to eq 200
+    expect(page.title).to eq "Ubicloud Admin - Invoice #{@instance.ubid}"
+  end
+end
--- a/spec/routes/web/admin/model/access_control_entry_spec.rb
+++ b/spec/routes/web/admin/model/access_control_entry_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require_relative "spec_helper"
+
+RSpec.describe CloverAdmin, "AccessControlEntry" do
+  include AdminModelSpecHelper
+
+  before do
+    @instance = create_access_control_entry
+    admin_account_setup_and_login
+  end
+
+  it "displays the AccessControlEntry instance page correctly" do
+    click_link "AccessControlEntry"
+    expect(page.status_code).to eq 200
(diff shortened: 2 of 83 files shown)


spec/routes/web/admin/model/invoice_spec.rb
```diff
@@ -13,6 +13,7 @@
     expect(page.status_code).to eq 200
     expect(page.title).to eq "Ubicloud Admin - Invoice - Browse"
 
+    expect(Aws::S3::Client).to receive(:new).and_return(instance_double(Aws::S3::Client, build_request: nil))
```
fdr: It is generally be better if we used the stub features (which are very nice)  of the real Aws::S3::Client. We do so inconsistently. Sometimes it's still useful to do an `.and_call_original` on it to get even more test power, but that doesn't apply to this case.

jeremyevans: Switched to use `Aws::S3::Client.new(stub_responses: true)`

fdr:
Oh, I see the to_s were to fix warnings, but similar questions on 1bade6497ef149cb2ff39df4b6ac8a2078b8ede1, it's not obvious what was erroneous about these

jeremyevans:
> Oh, I see the to_s were to fix warnings, but similar questions on [1bade64](https://github.com/ubicloud/ubicloud/commit/1bade6497ef149cb2ff39df4b6ac8a2078b8ede1), it's not obvious what was erroneous about these

All 5 resulted in spec failures (except the `create_account` change, which apparently snuck in). Three of the failures were required columns that Claude didn't add. The Invoice change was because without the mock, it tries to do an http request at runtime, which fails.  The Account change was because the link for account is on the name.

fdr:
interesting, was there a force push? feels hard to believe (though possible) i wouldn't have glanced at a red-x at the time I approved it

jeremyevans:
There was a force push, but the original push had 5 original commits. The only change was to fix the issue you identified:

```diff
--- a/spec/routes/web/admin/model/invoice_spec.rb
+++ b/spec/routes/web/admin/model/invoice_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe CloverAdmin, "Invoice" do
     expect(page.status_code).to eq 200
     expect(page.title).to eq "Ubicloud Admin - Invoice - Browse"
 
-    expect(Aws::S3::Client).to receive(:new).and_return(instance_double(Aws::S3::Client, build_request: nil))
+    expect(Aws::S3::Client).to receive(:new).and_return(Aws::S3::Client.new(stub_responses: true))
     click_link @instance.admin_label
     expect(page.status_code).to eq 200
     expect(page.title).to eq "Ubicloud Admin - Invoice #{@instance.ubid}"
```

This was rebased merged into the commit to fix the errors.

You can confirm this by looking at the CI history (https://github.com/ubicloud/ubicloud/actions?query=branch%3Ajeremy-admin-model-testing). CI tested:

* 3de9cff76a0e5220615154a4270a76082e86aa36 at 2:31pm
* 9b64353c7b5f72a71be50c30dc93e1f84b79b509 at 2:47pm

Both include all 5 commits. Apparently this tests merged commits to the base branch and not the underlying commits (I wasn't aware that GitHub Actions worked that way). The underlying commits were:

* 37a45a6ee12809a46ccd3ee7668113e011155e16 at 2:31pm
* a2b8a83fb2c21ff85ab94f7db8a9cdc6608695e6  at 2:47pm

fdr:
alright well that's actually good, it means I didn't gloss over something broken or jump the gun at the time (I think)

PR #4580, opened 2026-01-16
Demock spec/prog/vm/aws/nexus_spec.rb

fdr:
Replace instance_double and attribute mocking with real database fixtures.

- Remove instance_double(GithubRunner) and instance_double(GithubInstallation)
- Create real GithubInstallation and GithubRunner records
- Replace mocked attribute readers with model.update() calls
- Use Sequel exists? checks instead of nil comparisons
- Scope aws_instance fixture to describe blocks that need it
- Flatten context wrapper that no longer served a purpose
- Use have_attributes matcher for aws_instance verification

### DIFF
--- a/spec/prog/vm/aws/nexus_spec.rb
+++ b/spec/prog/vm/aws/nexus_spec.rb
@@ -1,35 +1,59 @@
 # frozen_string_literal: true
 
+require "aws-sdk-ec2"
+require "aws-sdk-iam"
+
 RSpec.describe Prog::Vm::Aws::Nexus do
   subject(:nx) {
-    described_class.new(vm.strand).tap {
-      it.instance_variable_set(:@vm, vm)
-      it.instance_variable_set(:@aws_instance, aws_instance)
-    }
+    described_class.new(st)
   }
 
   let(:st) {
     vm.strand
   }
 
+  let(:project) { Project.create(name: "test-prj") }
+
+  let(:location) {
+    Location.create(name: "us-west-2", provider: "aws", project_id: project.id,
+      display_name: "aws-us-west-2", ui_name: "AWS US East 1", visible: true)
+  }
+
+  let(:location_credential) {
+    LocationCredential.create_with_id(location, access_key: "test-access-key", secret_key: "test-secret-key")
+  }
+
+  let(:storage_volumes) {
+    [{encrypted: true, size_gib: 30}, {encrypted: true, size_gib: 3800}]
+  }
+
+  let(:vm_params) {
+    {location_id: location.id, unix_user: "test-user-aws", boot_image: "ami-030c060f85668b37d",
+     name: "testvm", size: "m6gd.large", arch: "arm64", storage_volumes:}
+  }
+
   let(:vm) {
-    prj = Project.create(name: "test-prj")
-    loc = Location.create(name: "us-west-2", provider: "aws", project_id: prj.id, display_name: "aws-us-west-2", ui_name: "AWS US East 1", visible: true)
-    LocationCredential.create_with_id(loc, access_key: "test-access-key", secret_key: "test-secret-key")
-    storage_volumes = [
-      {encrypted: true, size_gib: 30},
-      {encrypted: true, size_gib: 3800}
-    ]
-    Prog::Vm::Nexus.assemble("dummy-public key", prj.id, location_id: loc.id, unix_user: "test-user-aws", boot_image: "ami-030c060f85668b37d", name: "testvm", size: "m6gd.large", arch: "arm64", storage_volumes:).subject
+    location_credential  # force creation
+    Prog::Vm::Nexus.assemble_with_sshable(project.id, **vm_params).subject
+  }
+
+  let(:vm_without_sshable) {
+    location_credential
+    Prog::Vm::Nexus.assemble("ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAI", project.id, **vm_params).subject
   }
 
   let(:aws_instance) { AwsInstance.create_with_id(vm, instance_id: "i-0123456789abcdefg") }
+
+  let(:nic_aws_resource) {
+    NicAwsResource.create_with_id(vm.nics.first, network_interface_id: "eni-0123456789abcdefg")
+  }
 
   let(:client) { Aws::EC2::Client.new(stub_responses: true) }
 
   let(:iam_client) { Aws::IAM::Client.new(stub_responses: true) }
 
   let(:user_data) {
+    public_key = vm.sshable.keys.first.public_key.shellescape
     <<~USER_DATA
 #!/bin/bash
 custom_user="#{vm.unix_user}"
@@ -43,20 +67,17 @@ if [ ! -d /home/$custom_user ]; then
   chmod 700 /home/$custom_user/.ssh
   chmod 600 /home/$custom_user/.ssh/authorized_keys
 fi
-echo dummy-public-key > /home/$custom_user/.ssh/authorized_keys
+echo #{public_key} > /home/$custom_user/.ssh/authorized_keys
 usermod -L ubuntu
     USER_DATA
   }
 
   before do
-    allow(nx).to receive_messages(vm:, aws_instance:)
     allow(Aws::EC2::Client).to receive(:new).with(credentials: anything, region: "us-west-2").and_return(client)
     allow(Aws::IAM::Client).to receive(:new).with(credentials: anything, region: "us-west-2").and_return(iam_client)
   end
 
   describe ".assemble" do
-    let(:project) { Project.create(name: "test-prj") }
-
     it "creates correct number of storage volumes for storage optimized instance types" do
       loc = Location.create(name: "us-west-2", provider: "aws", project_id: project.id, display_name: "us-west-2", ui_name: "us-west-2", visible: true)
       storage_volumes = [
@@ -97,18 +118,22 @@ usermod -L ubuntu
   end
 
   describe "#before_destroy" do
-    it "stops billing records" do
+    it "finalizes active billing records" do
       br = BillingRecord.create(
-        project_id: vm.project.id,
+        project_id: project.id,
         resource_id: vm.id,
         resource_name: vm.name,
         billing_rate_id: BillingRate.from_resource_properties("VmVCpu", vm.family, vm.location.name)["id"],
         amount: vm.vcpus
       )
 
-      expect(vm).to receive(:active_billing_records).and_return([br])
-      expect(br).to receive(:finalize)
-      nx.before_destroy
+      expect { nx.before_destroy }
+        .to change { br.reload.span.unbounded_end? }.from(true).to(false)
+    end
+
+    it "completes without billing records" do
+      expect(vm.active_billing_records).to be_empty
+      expect { nx.before_destroy }.not_to change { vm.reload.active_billing_records.count }
     end
   end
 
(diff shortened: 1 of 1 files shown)


spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -638,33 +606,41 @@
...
+  describe "#destroy", "with aws_instance" do
+    before do
+      aws_instance
+    end
+
     it "deletes the instance" do
-      expect(aws_instance).to receive(:destroy)
       expect(client).to receive(:terminate_instances).with({instance_ids: ["i-0123456789abcdefg"]})
       expect { nx.destroy }.to hop("cleanup_roles")
+      expect(aws_instance).not_to exist
```
fdr, line 713: would be better with a change block, it's a bit slick to have to look at the fixture state to make sure this isn't a dead assertion

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -638,33 +606,41 @@
...
     end
 
-    it "pops directly if there is no aws_instance" do
-      expect(nx).to receive(:aws_instance).and_return(nil)
+    it "hops to cleanup_roles if there is no aws_instance" do
+      aws_instance.destroy
+      nx.instance_variable_set(:@aws_instance, nil)
```
fdr: can be simplified by moving it to the sibling describe without the aws_instance to begin with.  Is it even a unique test?

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -492,25 +458,28 @@
...
+    it "updates the sshable host" do
       expect { nx.wait_instance_created }.to hop("wait_sshable")
+      expect(vm.sshable.reload.host).to eq("1.2.3.4")
+    end
+
+    it "handles vm without sshable" do
+      vm.sshable.destroy
+      vm.associations.delete(:sshable) # clear cached association
+      expect { nx.wait_instance_created }.to hop("wait_sshable")
+      expect(vm.reload.cores).to eq(1)
```
fdr: is there a way to create the fixture the right way in the first place rather than creating more than necessary and wasting time hacking it back up to get rid of things?

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -492,25 +458,28 @@
...
     it "updates the vm" do
-      time = Time.now
-      expect(Time).to receive(:now).and_return(time).at_least(:once)
-      expect(client).to receive(:describe_instances).with({filters: [{name: "instance-id", values: ["i-0123456789abcdefg"]}, {name: "tag:Ubicloud", values: ["true"]}]}).and_call_original
-      expect(vm).to receive(:update).with(cores: 1, allocated_at: time, ephemeral_net6: "2a01:4f8:173:1ed3:aa7c::/79")
       expect { nx.wait_instance_created }.to hop("wait_sshable")
+      vm.reload
+      expect(vm.cores).to eq(1)
+      expect(vm.ephemeral_net6.to_s).to eq("2a01:4f8:173:1ed3:aa7c::/79")
+      expect(vm.allocated_at).to be_within(5).of(Time.now)
```
fdr: better to mock time and get it exact than use approximates

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -492,25 +458,28 @@
...
       client.stub_responses(:describe_instances, reservations: [{instances: [{state: {name: "running"}, network_interfaces: [{association: {public_ip: "1.2.3.4"}, ipv_6_addresses: [{ipv_6_address: "2a01:4f8:173:1ed3:aa7c::/79"}]}]}]}])
     end
 
     it "updates the vm" do
-      time = Time.now
-      expect(Time).to receive(:now).and_return(time).at_least(:once)
-      expect(client).to receive(:describe_instances).with({filters: [{name: "instance-id", values: ["i-0123456789abcdefg"]}, {name: "tag:Ubicloud", values: ["true"]}]}).and_call_original
-      expect(vm).to receive(:update).with(cores: 1, allocated_at: time, ephemeral_net6: "2a01:4f8:173:1ed3:aa7c::/79")
       expect { nx.wait_instance_created }.to hop("wait_sshable")
+      vm.reload
```
fdr, line 516: is this reload necessary?

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -295,153 +286,126 @@
...
+      let(:nic_nx) { Prog::Vnet::Aws::NicNexus.new(nic.strand) }
 
       before do
         client.stub_responses(:run_instances, Aws::EC2::Errors::Unsupported.new(nil, "Instance type not supported in availability zone"))
-        expect(vm).to receive(:sshable).and_return(create_mock_sshable(keys: [instance_double(SshKey, public_key: "dummy-public-key")])).at_least(:once)
-        nic.strand.stack.push({"exclude_availability_zones" => []})
-        NicAwsResource.create_with_id(nic, network_interface_id: "eni-0123456789abcdefg", subnet_az: "a")
+        refresh_frame(nic_nx, new_frame: {"exclude_availability_zones" => []})
+        nic.nic_aws_resource.update(subnet_az: "a")
+        nx.instance_variable_set(:@nic, nil)
```
fdr: very suspicious: why? make a nic, but then pretend it isn't there? forcing a reload? there must be a better way.

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -20,36 +20,20 @@
...
-  chown -R $custom_user:$custom_user /home/$custom_user/.ssh
-  chmod 700 /home/$custom_user/.ssh
-  chmod 600 /home/$custom_user/.ssh/authorized_keys
-fi
-echo dummy-public-key > /home/$custom_user/.ssh/authorized_keys
-usermod -L ubuntu
-    USER_DATA
-  }
-
   before do
```
fdr, line 75: the removal of user_data invariants is suspicious. why?

spec/prog/vm/aws/nexus_spec.rb
fdr: yes. it is.

fdr:
One area I think was improved in here and could be improved further with some spec tooling changes I have in reserve is using the aws client test mechanics and eliminating dead stub_responses. Namely, the `change.to` construct doesn't test any kind of precondition and putting in a negation for the precondition as-is is too wordy, but a bit of rspec stuff can make a shorthand. there's [a gem](https://github.com/ashanbrown/rspec-change_to_now) for this, but I think inlining makes more sense.

fdr:
alternatively, we can bring in the gem if you want, it just hasn't been touched in twelve years(!)

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -1,35 +1,59 @@
...
+require "aws-sdk-ec2"
+require "aws-sdk-iam"
+
 RSpec.describe Prog::Vm::Aws::Nexus do
   subject(:nx) {
-    described_class.new(vm.strand).tap {
-      it.instance_variable_set(:@vm, vm)
-      it.instance_variable_set(:@aws_instance, aws_instance)
-    }
+    described_class.new(vm.strand)
```
jeremyevans:
```suggestion
    described_class.new(st)
```

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -75,19 +96,40 @@
...
+
+    it "does not hop to destroy if already in the destroy state" do
+      ["destroy", "cleanup_roles"].each do |label|
+        nx.incr_destroy
+        nx.incr_destroying if label == "destroy"
+        st.label = label
+        expect { nx.before_run }.not_to hop("destroy")
+      end
+    end
+
```
jeremyevans: I think this is tested by the base prog code, and we don't need to add a separate spec here.

```suggestion
```

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -172,101 +211,95 @@
...
       client.stub_responses(:describe_subnets, subnets: [{availability_zone_id: "use1-az1"}])
-      expect(vm).to receive(:vcpus).and_return(2)
-      expect(vm).to receive(:sshable).and_return(create_mock_sshable(keys: [instance_double(SshKey, public_key: "dummy-public-key")]))
+    end
 
-      expect(vm.nics.first).to receive(:nic_aws_resource).and_return(instance_double(NicAwsResource, network_interface_id: "eni-0123456789abcdefg"))
-      expect(client).to receive(:run_instances).with({
+    it "creates an instance" do
+      client.stub_responses(:run_instances, instances: [{instance_id: "i-0123456789abcdefg", network_interfaces: [{subnet_id: "subnet-12345678"}], public_dns_name: "ec2-44-224-119-46.us-west-2.compute.amazonaws.com"}])
+      expect(client).to receive(:run_instances).with(hash_including(
```
jeremyevans: Previously, this didn't use `hash_including`, and now it does. Maybe expected, but seems odd that we could have a full hash match before and only a hash subset match after.

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -295,153 +328,127 @@
...
+        set_alternative_families([])
+        expect { nx.create_instance }.to nap(0).and change(GithubRunner, :count).by(1)
       end
 
       it "creates runner with the first alternative when current_family is the initial family" do
-        expect(nx).to receive(:frame).and_return({"alternative_families" => ["m7i", "m6a"]}).at_least(:once)
+        set_alternative_families(["m7i", "m6a"])
         expect { nx.create_instance }.to nap(0)
-        expect(vm.family).to eq("m7i")
+        expect(vm.reload.family).to eq("m7i")
```
jeremyevans: For this one and the next one, assuming it isn't expected to change the count, maybe worth adding:

```ruby
expect { nx.create_instance }.to nap(0).and not_change(GithubRunner, :count)
```

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -598,15 +635,16 @@
     it "not create billing records when the project is not billable" do
       vm.project.update(billable: false)
       expect { nx.create_billing_record }.to hop("wait")
-      expect(BillingRecord.count).to eq(0)
+      expect(BillingRecord.dataset).to be_empty
```
jeremyevans: Issue here is that on failure, you don't know what billing records existed that you didn't expect to exist. Recommend:

```suggestion
      expect(BillingRecord.all).to be_empty
```

jeremyevans (approved):
Looks good. Still a lot of mocking, but it appears to be AWS library mocking and unavoidable.

spec/prog/vm/aws/nexus_spec.rb
fdr: think this one is revolved (but code shifted around so it's not flagging it as outdated)

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -110,27 +110,6 @@
...
-      nx.incr_destroy
-      expect { nx.before_run }.to hop("destroy")
-      expect(br.reload.span).not_to eq(span_before)
-    end
-
-    it "hops to destroy if billing record is not found" do
-      nx.incr_destroy
-      expect(vm.active_billing_records).to be_empty
-      expect { nx.before_run }.to hop("destroy")
-    end
```
jeremyevans: This removes the two specs that did actually test the prog-specific code, instead of the two specs that tested the base prog code. A later commit removes the two other specs, and a spec after that adds the first spec back. I would still keep the second spec, to check the behavior when there are no billing records.

fdr: alright, it was a bit unclear to me what the original proposal was when the code got de-aligned, sorry about that (I thought it was to delete the code immediately prior)

fdr:
squash:  `git diff e45d21e31 822582639` (empty)

fdr:
....and then the rebase..

PR #4583, opened 2026-01-17
Add VmStorageVolume healthchecks

mohi-kalantari:
This commit implements a health check mechanism for VmStorageVolume models, specifically targeting `ubiblk` volumes that run via vhost-user-backend.

Added VmStorageVolume#check_pulse to verify the status of the associated systemd unit  on the host.

Registered VmStorageVolume in bin/monitor to enable the new health check.

### DIFF
--- a/prog/storage/migrate_spdk_vm_to_ubiblk.rb
+++ b/prog/storage/migrate_spdk_vm_to_ubiblk.rb
@@ -29,6 +29,10 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
     )
   end
 
+  def vm_storage_volume
+    vm.vm_storage_volumes.first
+  end
+
   label def stop_vm
     register_deadline(nil, 60 * 60)
     vm.incr_stop
@@ -55,7 +59,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   label def generate_vhost_backend_conf
-    vm.vm_host.sshable.cmd("sudo host/bin/convert-encrypted-dek-to-vhost-backend-conf --encrypted-dek-file :root_dir_path/data_encryption_key.json --kek-file /dev/stdin --vhost-conf-output-file :vhost_conf_path --vm-name :inhost_name --device :device", inhost_name:, root_dir_path:, vhost_conf_path:, device: vm.vm_storage_volumes.first.storage_device.name, stdin: vm.storage_secrets.to_json)
+    vm.vm_host.sshable.cmd("sudo host/bin/convert-encrypted-dek-to-vhost-backend-conf --encrypted-dek-file :root_dir_path/data_encryption_key.json --kek-file /dev/stdin --vhost-conf-output-file :vhost_conf_path --vm-name :inhost_name --device :device", inhost_name:, root_dir_path:, vhost_conf_path:, device: vm_storage_volume.storage_device.name, stdin: vm.storage_secrets.to_json)
     vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :vhost_conf_path", inhost_name:, vhost_conf_path:)
     hop_ready_migration
   end
@@ -113,9 +117,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   label def start_ubiblk_systemd_unit
-    disk_index = 0
-    unit_name = "#{vm.inhost_name}-#{disk_index}-storage.service"
-    vm.vm_host.sshable.cmd("sudo systemctl start :unit_name", unit_name:)
+    vm.vm_host.sshable.cmd("sudo systemctl start :unit_name", unit_name: vm_storage_volume.vhost_backend_systemd_unit_name)
     # Disk encryption happens using DEKs (Data Encryption Keys). DEKs needs to be presented
     # to the hypervisor to run the Vm. Now in the case of a breach to a VmHost, we don't want
     # to give away the keys easily so we encrypt the DEK with another set of keys: KEKs (key encryption key)
@@ -126,7 +128,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   label def update_vm_model
-    vm.vm_storage_volumes.first.update(
+    vm_storage_volume.update(
       use_bdev_ubi: false,
       vhost_block_backend_id: vm_host_vhost_block_backend.id,
       vring_workers: [1, vm.vcpus / 2].max,
@@ -153,17 +155,16 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   def migration_script_params
-    vsv = vm.vm_storage_volumes.first
     {
       "slice" => vm.vm_host_slice.name,
-      "device" => vsv.storage_device.name,
+      "device" => vm_storage_volume.storage_device.name,
       "vm_name" => vm.inhost_name,
       "encrypted" => true,
       "disk_index" => 0,
       "vhost_block_backend_version" => Config.vhost_block_backend_version,
-      "max_read_mbytes_per_sec" => vsv.max_read_mbytes_per_sec,
-      "max_write_mbytes_per_sec" => vsv.max_write_mbytes_per_sec,
-      "spdk_version" => vsv.spdk_installation.version
+      "max_read_mbytes_per_sec" => vm_storage_volume.max_read_mbytes_per_sec,
+      "max_write_mbytes_per_sec" => vm_storage_volume.max_write_mbytes_per_sec,
+      "spdk_version" => vm_storage_volume.spdk_installation.version
     }.to_json
   end
 
@@ -180,7 +181,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   def write_kek_pipe
-    kek = vm.vm_storage_volumes.first.key_encryption_key_1
+    kek = vm_storage_volume.key_encryption_key_1
     kek_data = {
       "key" => kek.key.strip,
       "init_vector" => kek.init_vector.strip,
@@ -191,7 +192,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   def base_image_path
-    vm.vm_storage_volumes.first.boot_image.path
+    vm_storage_volume.boot_image.path
   end
 
   def vm_host_vhost_block_backend


model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
   plugin ResourceMethods
   plugin ProviderDispatcher, __FILE__
+  include HealthMonitorMethods
 
   def aws?
     vm.location.aws?
   end
 
+  def init_health_monitor_session
+    return {} unless vm&.vm_host
```
jeremyevans: Add a `vm_host` method that does `vm&.vm_host`, and change places using that code to call it.

model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
+    {
+      ssh_session: vm.vm_host.sshable.start_fresh_session
+    }
+  end
+
+  def check_pulse(session:, previous_pulse:)
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless vhost_block_backend
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless vm&.vm_host
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless vm.display_state == "running"
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless session[:ssh_session]
```
jeremyevans: Probably best to DRY this up:

```suggestion
    unless vhost_block_backend && vm_host && vm.display_state == "running" && session[:ssh_session]
      return {reading: "up", reading_rpt: 1, reading_chg: Time.now}
    end
```

model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
+
+    service_name = "#{vm.inhost_name}-#{disk_index}-storage.service"
+    cmd = "systemctl is-active :service_name"
+    kw = {service_name:}
+
+    reading = begin
+      status = session[:ssh_session].exec!(cmd, **kw)
```
jeremyevans: Probably best to avoid these local variables. At least for `kw`, this is better as it saves a hash allocation (you can keep `cmd` if you want):

```suggestion
    reading = begin
      status = session[:ssh_session].exec!("systemctl is-active :service_name", service_name:)
```

model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
+      (status.strip == "active") ? "up" : "down"
+    rescue IOError, Errno::ECONNRESET
+      raise
+    rescue => e
+      Clog.emit("VmStorageVolume #{id} check_pulse ssh error", Util.exception_to_hash(e))
```
jeremyevans: I believe we would want a top level key here. See f2b128f972ce81847956e386d626bd9a94c0106e for the Clog usage guidelines.

```suggestion
      Clog.emit("VmStorageVolume #{id} check_pulse ssh error", vm_storage_volume_check_pulse_error: Util.exception_to_hash(e))
```

model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
+      "down"
+    end
+
+    if reading == "down"
+      Clog.emit("VmStorageVolume #{id} (VM #{vm.id}) systemd unit #{service_name} is not active", {})
```
jeremyevans: Empty hash is the default for the second argument:

```suggestion
      Clog.emit("VmStorageVolume #{id} (VM #{vm.id}) systemd unit #{service_name} is not active")
```

model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
+      status = session[:ssh_session].exec!(cmd, **kw)
+      (status.strip == "active") ? "up" : "down"
+    rescue IOError, Errno::ECONNRESET
+      raise
+    rescue => e
+      Clog.emit("VmStorageVolume #{id} check_pulse ssh error", Util.exception_to_hash(e))
+      "down"
+    end
+
+    if reading == "down"
```
jeremyevans: The log message seems inaccurate if there is an exception raised, in which case, two separate log messages would be emitted. I would only log this message if there was not an exception:

```suggestion
    if e.nil? && reading == "down"
```

Would even be better to move this into the preceding `begin` block.

spec/model/vm_storage_volume_spec.rb
```diff
@@ -68,4 +68,127 @@
...
+    end
+
+    context "when all conditions met" do
+      before do
+        backend = VhostBlockBackend.create(vm_host_id: vm_host.id, version: "test", allocation_weight: 100)
+        volume.update(vhost_block_backend_id: backend.id, vring_workers: 2)
+      end
+
+      it "returns up when service is active" do
+        expect(session[:ssh_session]).to receive(:exec!).with("systemctl is-active :service_name", service_name: "#{vm.inhost_name}-0-storage.service").and_return("active\n")
```
jeremyevans: You are not allowed to mock `exec!`, as it would prevent the interpolation security checks from running. You need to mock `_exec!`. This is what is causing the CI failure. All cases should be fixed.

```suggestion
        expect(session[:ssh_session]).to receive(:_exec!).with("systemctl is-active #{vm.inhost_name}-0-storage.service").and_return("active\n")
```

model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
 
   def aws?
     vm.location.aws?
   end
 
+  def init_health_monitor_session
+    return {} unless vm&.vm_host
+
+    {
+      ssh_session: vm.vm_host.sshable.start_fresh_session
```
enescakir: I'm thinking out loud here. This would double our monitoring load since each VM has its own storage volume. Since we already establish an SSH connection to monitor the VM, could we reuse the same connection to monitor the storage volume as well?
I'd also be curious to hear @byucesoy's thoughts, as we've discussed SSH load of monitoring quite a bit in the past.

model/vm_storage_volume.rb
```diff
@@ -15,11 +15,47 @@ class VmStorageVolume < Sequel::Model
...
+    }
+  end
+
+  def check_pulse(session:, previous_pulse:)
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless vhost_block_backend
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless vm&.vm_host
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless vm.display_state == "running"
+    return {reading: "up", reading_rpt: 1, reading_chg: Time.now} unless session[:ssh_session]
+
+    service_name = "#{vm.inhost_name}-#{disk_index}-storage.service"
```
pykello: nit: I have a slight preference for making formation of this string a method. Right now, this has been repeated here & in `MigrateSpdkVmToUbiblk`, and I'm pretty sure there will be more uses.

model/vm_storage_volume.rb
mohi-kalantari: moved the logic to vms check_pulse and it's way better now. thanks.

model/vm.rb
```diff
@@ -223,14 +224,35 @@ def storage_size_gib
...
     }
   end
 
   def check_pulse(session:, previous_pulse:)
     reading = begin
-      session[:ssh_session].exec!("systemctl is-active :inhost_name :inhost_name-dnsmasq", inhost_name:).split("\n").all?("active") ? "up" : "down"
+      if session[:ssh_session]
+        units = []
+        unless sshable
+          units << inhost_name
```
byucesoy: If we are checking other services anyway, we can check these for sshable vms as well.

bin/monitor
```diff
@@ -43,7 +43,7 @@ else
   monitor_models = [
     VmHost,
     PostgresServer,
-    Vm.where(~Sshable.where(id: Sequel[:vm][:id]).exists),
+    Vm,
```
enescakir: Instead of checking `unless vm_host` and `if session[:ssh_session]` separately, we might filter by `.exclude(vm_host: nil)`

bin/monitor
```diff
@@ -43,7 +43,7 @@ else
...
     MinioServer,
```
enescakir, line 47: We use `vm_host`, `sshable`, `vm_storage_volumes` which generates 3N queries. We should load eagerly.

`.eager(:vm_host, :sshable, :vm_storage_volumes)`

model/vm.rb
```diff
@@ -230,7 +231,13 @@ def init_health_monitor_session
...
   def check_pulse(session:, previous_pulse:)
     reading = begin
-      session[:ssh_session].exec!("systemctl is-active :inhost_name :inhost_name-dnsmasq", inhost_name:).split("\n").all?("active") ? "up" : "down"
+      units = [inhost_name, "#{inhost_name}-dnsmasq"]
+      vm_storage_volumes.each do |vol|
+        units << vol.vhost_backend_systemd_unit_name if vol.vhost_block_backend
+      end
+
+      cmd = "systemctl is-active #{units.map(&:shellescape).join(" ")}".freeze
+      session[:ssh_session].exec!(cmd).split("\n").all?("active") ? "up" : "down"
```
jeremyevans: Never manually call `freeze` on a string passed as a shell command, even if you believe it to be safe.

```suggestion
      session[:ssh_session].exec!("systemctl is-active :shelljoin_units", shelljoin_units: units).split("\n").all?("active") ? "up" : "down"
```

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -29,6 +29,10 @@ def self.assemble(vm_id)
     )
   end
 
+  def vm_storage_volume
+    vm.vm_storage_volumes.first
```
jeremyevans, line 33: Might be worth raising if there is more than a single storage volume. Otherwise, this could silently skip a storage volume. Not sure how many VMs have more than one storage volume, but my guess would be that VMs with more than one storage volume are more likely to have greater importance and greater impact on failure.

jeremyevans (changes requested):

prog/storage/migrate_spdk_vm_to_ubiblk.rb
mohi-kalantari: There is check at the beginning of the prog which makes sure the vm has exactly one disk. It's not designed for vms with multiple disks.

jeremyevans (approved):

PR #4584, opened 2026-01-17
Create pages for unavailable VMs and slices regardless of host status

enescakir:
In the current VM monitoring design, SSH errors during availability checks were silently handled by hopping to unavailable without creating a page, assuming the host being down would be handled separately by HostNexus. This assumption was flawed.

This assumption originates from the initial VM monitoring design (e8edbe60c4), but VMs can have SSH issues while the host is healthy, and vice versa.

VMs and slices should be monitored independently, and pages should be created whenever a resource is unavailable, regardless of the state of underlying resources. This aligns with our prior discussions and ensures consistent behavior across resources similar to PostgresServer.

While this may introduce some additional alerts, pages are already grouped by VM host in the admin UI, which should be help to handle noise.

This change updates both VM Nexus and VM host slice logic, as the slice logic duplicated the same assumption.

### DIFF
--- a/prog/vm/metal/nexus.rb
+++ b/prog/vm/metal/nexus.rb
@@ -261,10 +261,6 @@ class Prog::Vm::Metal::Nexus < Prog::Base
     when_checkup_set? do
       hop_unavailable if !available?
       decr_checkup
-    rescue Sshable::SshError
-      # Host is likely to be down, which will be handled by HostNexus. We still
-      # go to the unavailable state for keeping track of the state.
-      hop_unavailable
     end
 
     nap 6 * 60 * 60
@@ -309,19 +305,12 @@ class Prog::Vm::Metal::Nexus < Prog::Base
       hop_start_after_host_reboot
     end
 
-    begin
-      if available?
-        decr_checkup
-        hop_wait
-      else
-        # Use deadlines to create a page instead of a custom page, so page
-        # resolution in different states can be handled properly.
-        register_deadline("wait", 0)
-      end
-    rescue Sshable::SshError
-      # Host is likely to be down, which will be handled by HostNexus. No need
-      # to create a page for this case.
+    if available?
+      decr_checkup
+      hop_wait
     end
+
+    register_deadline("wait", 0)
 
     nap 30
   end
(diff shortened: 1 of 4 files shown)


enescakir:
> > but VMs can have SSH issues while the host is healthy, and vice versa.
> 
> I don't see how `Prog::Vm::Metal::Nexus.available?` and `Prog::Vm::VmHostSliceNexus.available?` are related to SSHing VMs. Both SSH hosts only. What am I missing here? Could you explain more?
> 
> Maybe a specific example that the current situation failed could help.

Yes, we ssh to the VM host, not the VM, so the commit message is a bit misleading. I'll rewrite it.

`Sshable::SshError` is raised when a command exits with a non-zero status. for the vm example, if the VM service is unloaded for any reason, `systemctl is-active` exits with code 3 for non-existent services, which triggers a `Sshable::SshError`.

However, `systemctl is-active` can return a non-zero exit code for reasons other than the VM host being down. Right now, we treat any such error as if the VM host is unavailable and expect the VM host to be paged, which is an incorrect assumption. VM stuck in the unavailable state without triggering any notification.

prog/vm/metal/nexus.rb
```diff
@@ -309,20 +305,13 @@ def write_params_json
...
-      end
-    rescue Sshable::SshError
-      # Host is likely to be down, which will be handled by HostNexus. No need
-      # to create a page for this case.
+    if available?
+      decr_checkup
+      hop_wait
     end
 
+    register_deadline("wait", 0)
```
byucesoy, line 313: I'd recommend moving registering deadline to `wait` label just before hopping to `unavailable` label, like [this](https://github.com/ubicloud/ubicloud/pull/4601). This ensures that even if a failure happens in `unavailable` label, we would still be able to register deadline.

I think we should start using this pattern to ensure deadlines are getting registered.

byucesoy: This can be in separate PR as well if you prefer.

enescakir: Though `.available?` can still raise an unhandled exception, I fixed that too.

https://github.com/ubicloud/ubicloud/pull/4605

PR #4585, opened 2026-01-18
Demock VmHost model specs

enescakir:
### I needed to add a new test to the VmHost model spec, but the specs were messy. so I cleaned them up first.

- **Organize vm_host_spec into describe blocks by method**
  

- **Rename subject in VmHost model specs**
  

- **Demock VmHost model specs**


fdr (approved):
Looks pretty good, I followed the patches one at a time.

PR #4586, opened 2026-01-18
Do not set up nftables twice when preparing a VM host

enescakir:
We currently call `create_addresses` during host assembly, which starts the `SetupNftables` prog. We also bud the same prog at the `prep` label of the VM host nexus.

The first invocation cannot connect to the VM host until rhizome is bootstrapped, which leads to unnecessary exceptions.

No need to setup nftables twice.

### DIFF
--- a/model/vm_host.rb
+++ b/model/vm_host.rb
@@ -214,7 +214,7 @@ class VmHost < Sequel::Model
       end
     end
 
-    Strand.create(prog: "SetupNftables", label: "start", stack: [{subject_id: id}])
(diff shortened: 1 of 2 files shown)


enescakir:
> We'll still have some errors since even during prep, we run all the strands in parallel including `PrepHost`. I feel like this SetupNftables can be moved to a separate state in prep_host. I'm OK with this implementation, too.

Yes, we see more exceptions during prep, especially around apt install race conditions. It would be better to handle these more gracefully instead of raising exceptions, but I don’t have a good solution yet.

This specific case is about the first nftables setup we trigger during assemble. It logs Net::SSH::AuthenticationFailed until the later steps complete, so we end up creating noise even before the vm host strand run.

PR #4589, opened 2026-01-19
Skip billing record update when unchanged

byucesoy:
Only finalize and recreate billing records when the new records differ from the existing ones.

### DIFF
--- a/prog/postgres/postgres_resource_nexus.rb
+++ b/prog/postgres/postgres_resource_nexus.rb
@@ -203,27 +203,36 @@ class Prog::Postgres::PostgresResourceNexus < Prog::Base
     decr_update_billing_records
 
     if postgres_resource.project.billable
-      postgres_resource.active_billing_records.each(&:finalize)
-
       flavor = postgres_resource.flavor
       vm_family = representative_server.vm.family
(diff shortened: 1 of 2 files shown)


prog/postgres/postgres_resource_nexus.rb
```diff
@@ -216,14 +214,29 @@ def before_run
...
+
+        billing_record_parts.each do |brp|
+          BillingRecord.create(
+            project_id: postgres_resource.project_id,
+            resource_id: postgres_resource.id,
+            resource_name: postgres_resource.name,
+            billing_rate_id: BillingRate.from_resource_properties(brp[:resource_type], brp[:resource_family], postgres_resource.location.name, postgres_resource.location.byoc)["id"],
+            amount: brp[:amount]
+          )
+        end
```
enescakir: Why don't we use `new_billing_records`?

```suggestion
        new_billing_records.each do |br|
          BillingRecord.create(
            project_id: postgres_resource.project_id,
            resource_id: postgres_resource.id,
            resource_name: postgres_resource.name,
            billing_rate_id: br[:billing_rate_id],
            amount: br[:amount]
          )
        end
```

PR #4601, opened 2026-01-20
Reduce failover related deadlines for PG

byucesoy:
Also move deadline registration to before hops, so that if the label itself raises an exception, the deadline is still registered.

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -465,10 +465,12 @@ SQL
     end
 
     when_unplanned_take_over_set? do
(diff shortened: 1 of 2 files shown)


enescakir (approved):

prog/postgres/postgres_server_nexus.rb
```diff
@@ -481,7 +483,11 @@ def before_run
     end
 
     when_checkup_set? do
-      hop_unavailable if !available?
+      if !available?
```
enescakir: nit
```suggestion
      unless available?
```

PR #4609, opened 2026-01-20
Upgrade Stripe SDK to v13+ service-based pattern

enescakir:
Upgrade Stripe SDK to v13+ service-based pattern
Stripe SDK v13 introduced a new service-based call pattern through
StripeClient, replacing the resource-based class method approach. The
old pattern (e.g., Stripe::Customer.retrieve) is deprecated and new API
endpoints will only support the service-based approach. We converted all
Stripe::* class methods to client.v1.* service calls

Migration guide:
https://github.com/stripe/stripe-ruby/wiki/Migration-guide-for-v13

Add centralized StripeClient module that provides a cleaner API for
accessing Stripe services. The module delegates method calls to the v1
services, allowing code like: StripeClient.customers.delete(stripe_id)
instead of:
Stripe::StripeClient.new(Config.stripe_secret_key).v1.customers.delete(stripe_id)

We avoid creating a new Stripe::StripeClient per access. Make
StripeClient be the Stripe::V1Services object. In testing, use a Struct
with columns for all v1 services that are used. In addition to being
faster by avoiding indirection, this also makes the mocking code
simpler.


jeremyevans (approved):
This looks fine, but I recommend adding `lib/stripe_client.rb` with:

```ruby
require "stripe"

StripeClient = Stripe::StripeClient.new(Config.stripe_secret_key).v1
```

That would allow you to use code like:

```ruby
StripeClient.customers.delete(stripe_id)
```

This also allows removing any other `require "stripe"` calls, so that stripe is never required unless `StripeClient` is referenced (and it should only be referenced if `Config.stripe_secret_key` is set).

enescakir:
> ```ruby
> require "stripe"
> 
> StripeClient = Stripe::StripeClient.new(Config.stripe_secret_key).v1
> ```
> 

It didn't work well with frozen tests and rubocop. Claude implemented this version. What do you think? I'll squash before merging.
https://github.com/ubicloud/ubicloud/pull/4609/changes/f62b3c92bc968e83a03775a99eb0b5e25a73b277

jeremyevans:
> > ```ruby
> > require "stripe"
> > 
> > StripeClient = Stripe::StripeClient.new(Config.stripe_secret_key).v1
> > ```
> 
> It didn't work well with frozen tests and rubocop. Claude implemented this version. What do you think? I'll squash before merging. [f62b3c9](https://github.com/ubicloud/ubicloud/commit/f62b3c92bc968e83a03775a99eb0b5e25a73b277)

Rubocop doesn't matter, you can always turn off whatever cops are complaining about it.  What Claude created DRYs things up a bit, but you end up creating a new `Stripe::StripeClient` every place it is used. I'll test this locally and see if I can fix whatever issues are present.

jeremyevans:
I pushed a commit for the changes I recommend.  Note that this results in `StripeClient` (now the `Stripe::V1Services` instance) being frozen in production. I reviewed the `Stripe::V1Services` implementation, and the object is currently read-only, so I don't expect this to cause an issue: https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/services/v1_services.rb

PR #4612, opened 2026-01-20
Create extensions for the paradedb images

furkansahin:


prog/postgres/postgres_server_nexus.rb
```diff
@@ -400,16 +399,14 @@ def before_run
...
-    sudo cp /etc/postgresql-partners/post-installation-script postgres/bin/post-installation-script
-    sudo chown ubi:ubi postgres/bin/post-installation-script
-    sudo chmod +x postgres/bin/post-installation-script
-    postgres/bin/post-installation-script
-    COMMAND
-
-    vm.sshable.cmd(command)
+    if postgres_server.primary? && postgres_server.resource.flavor != PostgresResource::Flavor::STANDARD
+      postgres_server.run_query(<<SQL)
+CREATE EXTENSION IF NOT EXISTS pg_cron;
```
byucesoy, line 408: Can we download, but not install these packages in image build? So that standard PG user's wouldn't be able create them with `CREATE EXTENSION`. In this label we would install the packages for paradedb users just before creating them.

furkansahin: updated

PR #4617, opened 2026-01-20
Retry VM creation on different AZ after InsufficientInstanceCapacity

heavycrystal:
AWS can sometimes have a "temporary" shortage of capacity for a
particular EC2 instance type, returning InsufficientInstanceCapacity.
Treating it like an unavailable AZ and swapping to a different one on
retry should help with faster provisioning. This is currently only for
VMs for PostgresResources.

### DIFF
--- a/prog/vm/aws/nexus.rb
+++ b/prog/vm/aws/nexus.rb
@@ -196,17 +196,10 @@ class Prog::Vm::Aws::Nexus < Prog::Base
         vm.incr_destroy
         nap 0
       end
-      raise
-    rescue Aws::EC2::Errors::Unsupported => e
-      if (retry_count = frame["retry_count"] || 0) >= 5
-        raise e
-      end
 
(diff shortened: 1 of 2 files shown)


prog/vm/aws/nexus.rb
```diff
@@ -196,17 +196,16 @@ def before_destroy
         vm.incr_destroy
         nap 0
       end
-      raise
+
+      if UBID.class_match?(vm.name, PostgresServer)
```
furkansahin: why do you need this check?

heavycrystal: I wanted to restrict it to only Postgres servers because I wasn't sure on semantics for other types of VMs, if it's something that would be appreciated for everything I can remove the conditional

furkansahin: yeah, we can just remove the if check

PR #4619, opened 2026-01-20
Improve blob storage handling etcd backup

mohi-kalantari:
Don't start EtcdBackups until cluster is Ready

A new condition is added to wait for the ready cluster to start capturing the backups.

Wait for available blob storage in EtcdBackupNexus

Previously, e2e tests would fail because EtcdBackupNexus was created before MinioCluster. This caused the model to cache a nil blob storage value, leading to a terminal hang.

This commit introduces a wait loop to ensure blob storage is available before proceeding. Additionally, negative caching has been disabled to prevent disruptions to the e2e test flow.

Specs are also democked

### DIFF
--- a/model/kubernetes/kubernetes_etcd_backup.rb
+++ b/model/kubernetes/kubernetes_etcd_backup.rb
@@ -13,8 +13,9 @@ class KubernetesEtcdBackup < Sequel::Model
   BACKUP_BUCKET_EXPIRATION_DAYS = 7
 
   def need_backup?
-    return false unless blob_storage
-    return false if kubernetes_cluster.functional_nodes.empty?
+    return false unless blob_storage &&
+      kubernetes_cluster.functional_nodes.any? &&
+      kubernetes_cluster.strand.label == "wait"
 
     sshable = kubernetes_cluster.sshable
     case sshable.d_check("backup_etcd")
--- a/spec/model/kubernetes/kubernetes_etcd_backup_spec.rb
+++ b/spec/model/kubernetes/kubernetes_etcd_backup_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe KubernetesEtcdBackup do
   let(:location) { Location[Location::HETZNER_FSN1_ID] }
   let(:private_subnet) { PrivateSubnet.create(project_id: project.id, name: "test", location_id: location.id, net6: "fe80::/64", net4: "192.168.0.0/24") }
   let(:kc) {
-    kc = KubernetesCluster.create(
+    kc = Prog::Kubernetes::KubernetesClusterNexus.assemble(
       name: "test-cluster",
       version: Option.kubernetes_versions.first,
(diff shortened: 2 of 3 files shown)


model/kubernetes/kubernetes_etcd_backup.rb
```diff
@@ -35,8 +36,7 @@ def next_backup_time
   end
 
   def blob_storage
-    return @blob_storage if defined?(@blob_storage)
-    @blob_storage = MinioCluster[project_id: Config.postgres_service_project_id, location_id: location.id]
+    @blob_storage ||= MinioCluster[project_id: Config.postgres_service_project_id, location_id: location.id]
```
jeremyevans: You are sure that you don't want to cache negative lookup? There are cases where you don't, but if so, there should be a commit explaining why negative lookup is not cached.

jeremyevans (commented):
Looks mostly good.

model/kubernetes/kubernetes_etcd_backup.rb
mohi-kalantari: Yep. Most of the times, the MinioCluster exists and there is no need for a negative cache but in the case of e2e tests, a MinioCluster is created at the same time the Backup is created and it would hang on a nil blob storage forever

mohi-kalantari: I updated the commit message

jeremyevans: I don't understand why this would hang on `nil` blob storage forever with negative lookup caching. The caching is per model instance, and each scheduled `respirate` strand run should use a new model instance. Did you test without this specific change (with the other changes in the PR) and confirm that it hangs forever? If so, can you point me to the failing E2E test run? If not, I recommend removing this particular change (and the related model spec change), and running E2E tests. I reviewed the E2E test result in this PR, which failed, and includes this change, so I'm not sure how the conclusion was reached that this change fixed an issue.

I guess we expect the cluster to always exist in production settings, so negative lookup caching is not that important. Removing it results in two separate queries for the `MinioCluster` in `wait` if the cluster doesn't exist, but maybe no effect other than that. Still, using an incorrect caching pattern should generally be avoided, as bad patterns tend to proliferate .

mohi-kalantari: got it thanks for the explanation. I reverted the negative cache.

mohi-kalantari:
It failed because of this test
Test::GithubRunner.check_test_runs

re running the e2e tests

jeremyevans (commented):
Looks good to me.

aamederen:
LGTM as well, just waiting for a successful E2E run involving PG and K8s together

mohi-kalantari:
The e2e pipeline only including k8s and postgres is passed

https://github.com/ubicloud/ubicloud/actions/runs/21221390844

PR #4621, opened 2026-01-21
Allow HA to work without dns_zone configured in development

serprex:
mTLS with sslmode=require is fine; verification impossible connecting via IP

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -96,16 +96,18 @@ class PostgresResource < Sequel::Model
   end
 
   def replication_connection_string(application_name:)
+    return nil unless dns_zone || representative_server
+
     query_parameters = {
       sslrootcert: "/etc/ssl/certs/ca.crt",
       sslcert: "/etc/ssl/certs/server.crt",
       sslkey: "/etc/ssl/certs/server.key",
-      sslmode: "verify-full",
+      sslmode: dns_zone ? "verify-full" : "require",
       dbname: "postgres",
       application_name:
     }.map { |k, v| "#{k}=#{v}" }.join("&")
 
-    URI::Generic.build2(scheme: "postgres", userinfo: "ubi_replication", host: identity, query: query_parameters).to_s
+    URI::Generic.build2(scheme: "postgres", userinfo: "ubi_replication", host: dns_zone ? identity : representative_server.vm.ip4_string, query: query_parameters).to_s
   end
 
   def version
--- a/spec/model/postgres/postgres_resource_spec.rb
+++ b/spec/model/postgres/postgres_resource_spec.rb
@@ -58,9 +58,24 @@ RSpec.describe PostgresResource do
     expect(postgres_resource.connection_string).to be_nil
   end
 
+  it "returns replication connection string as nil if there is no server" do
+    # No server created, no dns_zone
+    expect(postgres_resource.replication_connection_string(application_name: "pgubidstandby")).to be_nil
+  end
+
   it "returns replication_connection_string" do
+    expect(postgres_resource).to receive(:dns_zone).and_return(instance_double(DnsZone)).at_least(:once)
     s = postgres_resource.replication_connection_string(application_name: "pgubidstandby")
     expect(s).to include("ubi_replication@#{postgres_resource.ubid}.postgres.ubicloud.com", "application_name=pgubidstandby", "sslcert=/etc/ssl/certs/server.crt")
+  end
+
+  it "returns replication_connection_string with ip when no dns_zone exists" do
+    vm = create_hosted_vm(project, private_subnet, "pg-vm")
+    PostgresServer.create(timeline:, resource_id: postgres_resource.id, vm_id: vm.id, representative_at: Time.now, synchronization_status: "ready", timeline_access: "push", version: "17")
+    AssignedVmAddress.create(dst_vm_id: vm.id, ip: "1.2.3.4/32")
+    expect(postgres_resource.dns_zone).to be_nil
+    s = postgres_resource.replication_connection_string(application_name: "pgubidstandby")
+    expect(s).to include("ubi_replication@1.2.3.4", "application_name=pgubidstandby", "sslcert=/etc/ssl/certs/server.crt")
   end
 
   describe "#provision_new_standby" do


model/postgres/postgres_resource.rb
```diff
@@ -98,11 +98,11 @@ def replication_connection_string(application_name:)
       sslrootcert: "/etc/ssl/certs/ca.crt",
       sslcert: "/etc/ssl/certs/server.crt",
       sslkey: "/etc/ssl/certs/server.key",
-      sslmode: "verify-full",
+      sslmode: dns_zone ? "verify-full" : "require",
       application_name:
     }.map { |k, v| "#{k}=#{v}" }.join("&")
 
-    URI::Generic.build2(scheme: "postgres", userinfo: "ubi_replication", host: identity, query: query_parameters).to_s
+    URI::Generic.build2(scheme: "postgres", userinfo: "ubi_replication", host: dns_zone ? identity : representative_server&.vm&.ip4_string, query: query_parameters).to_s
```
serprex: bit awkward while representative_server nil, silently giving broken URI probably less desirable compared to raising exception

serprex: updated to return nil

spec/model/postgres/postgres_resource_spec.rb
```diff
@@ -55,11 +55,25 @@
...
     s = postgres_resource.replication_connection_string(application_name: "pgubidstandby")
     expect(s).to include("ubi_replication@#{postgres_resource.ubid}.postgres.ubicloud.com", "application_name=pgubidstandby", "sslcert=/etc/ssl/certs/server.crt")
   end
 
+  it "returns replication_connection_string with ip when no dns_zone exists" do
+    vm = create_hosted_vm(project, private_subnet, "pg-vm")
+    PostgresServer.create(timeline:, resource_id: postgres_resource.id, vm_id: vm.id, representative_at: Time.now, synchronization_status: "ready", timeline_access: "push", version: "17")
+    expect(postgres_resource.dns_zone).to be_nil
+    s = postgres_resource.replication_connection_string(application_name: "pgubidstandby")
+    expect(s).to include("ubi_replication@#{postgres_resource.representative_server&.vm&.ip4_string}", "application_name=pgubidstandby", "sslcert=/etc/ssl/certs/server.crt")
```
jeremyevans: You should know in the spec whether or not the representative server exists. I'm guessing it does (`representative_at: Time.now`) and this should be switched to:

```suggestion
    expect(s).to include("ubi_replication@#{postgres_resource.representative_server.vm.ip4_string}", "application_name=pgubidstandby", "sslcert=/etc/ssl/certs/server.crt")
```

model/postgres/postgres_resource.rb
```diff
@@ -94,15 +94,17 @@ def connection_string
...
 
   def replication_connection_string(application_name:)
+    return nil unless dns_zone || representative_server
+
     query_parameters = {
       sslrootcert: "/etc/ssl/certs/ca.crt",
       sslcert: "/etc/ssl/certs/server.crt",
       sslkey: "/etc/ssl/certs/server.key",
-      sslmode: "verify-full",
+      sslmode: dns_zone ? "verify-full" : "require",
```
jeremyevans, line 105: @fdr This seems fine to me. If you don't have a dns zone setup, I think `sslmode=require` is the best you can do, as the certificates do not include IP address entries.

fdr: I mean, given it's mtls, in principle, we never need verify-full. something not answered in the commit message is...why? is this for development? do we want to guard it with a Config predicate (e.g. !production)? It gets scattered to two or three sites in the code.

serprex: Dev is where I hit this, but any environment without dns zone will need this

fdr: okay. like, and what kind of environment might that be? one without HA support, at a minimum? sounds like probably almost exclusively dev? @jeremyevans is it clear/safe enough as it is (three branches scattered around) or should we bin it with something else? I see already some prior art like that:

```                                                                                                                                                                                       
prog/vnet/cert_nexus.rb                                                                                                                                                                 
12:    unless Config.development? || DnsZone[dns_zone_id]                                                                                                                               
26:    if Config.development? && cert.dns_zone_id.nil?                                                                                                                                  
111:    if Config.development? && cert.dns_zone_id.nil?    
```

jeremyevans: I think it's OK as is. Might be worth adding the `Config.development?` check, so the only environment it's allowed in is development. If we want to consolidate the logic, because this is used in at least 3 different contexts, we could add something like:

```ruby
  def Util.use_dns_zone?(dns_zone)
    dns_zone || !Config.development?
  end
```

fdr (approved):
I'm okay with it as-is, even though I think it's a bit slick. In other places, we don't change the functionality for something as fundamental for DNS availability without guarding behind explicit environment (e.g. `Config.development?`) methods. But if other people are okay with it, I think I can be too. For my taste, it leaves a bit too much to the imagination, that is, implying a production code path that involves a lack of DNS.

model/postgres/postgres_resource.rb
serprex: It's desirable to be able to test HA while running a minimal local dev

no problem adding `Config.development?` checks in, consolidation helps with managing branch growth

furkansahin:
@serprex can you fix the commits? I'll merge it afterwards

serprex:
@furkansahin done, not sure how much detail is necessary in commit message here detailing connecting via IP

model/postgres/postgres_resource.rb
byucesoy: Actually I'm against sprinkling environment checks around the codebase. It creates certain expectation while doing development, which doesn't hold true in production environment. Few checks where alternatives would be too complex to implement is OK but it shouldn't be the norm.

Also we might have DnsServer in development environment, in such cases we would want to have full experience. I, myself, set up dev environment with MinIo and DnsServer time to time.

Instead, we should check for existence of functionalilty and gracefully fail back if it is not available. Hence what is suggested in this PR originally is better where it checks the existence of dns_zone IMO. If we check for the existence of dns_zone, then there is no need for checking for environment anymore.

serprex:
@byucesoy thankfully I could still git revert the commit lost in squash

PR #4633, opened 2026-01-21
Add project FF to default use_different_azs for PostgresResource

heavycrystal:
Gives the capability to have all Postgres instances in AWS locations in a project have their standbys on different AZs. This will safeguard against an AWS AZ outage, but also incur costs due to cross-AZ traffic. Disabled by default.


byucesoy:
You need to add test to cover the new branch. CI enforces 100% line and branch coverage.

Also, could you please squash the commits?

PR #4634, opened 2026-01-21
Enable wal-g verify on backup-push

heavycrystal:
wal-g can check Postgres data checksums during backup-push and log corrupted block numbers to the sentinel JSON, so added --verify to take-backup for that. That information can be used down the line to inform integrity of large backups.

### DIFF
--- a/rhizome/postgres/bin/take-backup
+++ b/rhizome/postgres/bin/take-backup
@@ -9,4 +9,4 @@ end
 
 v = ARGV[0]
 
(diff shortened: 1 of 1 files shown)


byucesoy:
We enable checksum during [image generation](https://github.com/ubicloud/postgres-vm-images/blob/d3452e7ec7e3fa59c5c6c9ac306d1012b6b4d373/common/setup_base.sh#L43). AFAIR, you do the same. Is this not the case?

However, `--verify` for wal-g seems useful.

heavycrystal:
> AFAIR, you do the same. Is this not the case?

It is, I missed that. Removed that code from this PR

PR #4637, opened 2026-01-21
Allow LoadBalancer refresh_cert semaphore to refresh existing cert

jeremyevans:
Previously, this semaphore was set only when there was an existing cert for the load balancer that needed a cert, and could not be used to refresh the cert generally. Refreshing existing certs is needed to add private DNS names to the certs.

### DIFF
--- a/prog/vnet/cert_nexus.rb
+++ b/prog/vnet/cert_nexus.rb
@@ -59,20 +59,28 @@ class Prog::Vnet::CertNexus < Prog::Base
     end
 
     each_authorization { it.dns.request_validation }
+    update_stack("last_dns_validation_request" => Time.now.to_i)
 
     hop_wait_dns_validation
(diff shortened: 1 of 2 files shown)


jeremyevans:
> why is refresh_cert not identical to need_certificates?

`need_certificates?` checks if you have any certificates that are close to expiration (more than 60 days old). This is the first time we need to update an existing cert for a reason other than expiration, I would guess.

jeremyevans:
@fdr Rerequesting your review, as this has changed significantly since your last review. I've tested this in staging an confirmed that it worked for all load balancers with certs in staging.

fdr (approved):
alright, thanks for pulling that one out. I suppose it makes sense for them to push retry out to the client, that knows more about the anticipated visibility of their record.

PR #4641, opened 2026-01-22
Add more data to the VmHostSlice deadline pages

enescakir:
We have already added more data for some other models, and it has been helpful for identifying the VM host from pager notifications.

I didn't check for VM host non-existence because the column is non-nullable and has a foreign key constraint.


model/strand.rb
```diff
@@ -181,6 +181,8 @@ def unsynchronized_run
             {vm_host: sbj.vm_host&.ubid, data_center: sbj.vm_host&.data_center, boot_image: sbj.boot_image, location: sbj.location.display_name, arch: sbj.arch, vcpus: sbj.vcpus, ipv4: sbj.ip4_string}
           when VmHost
             {data_center: sbj.data_center, location: sbj.location.display_name, arch: sbj.arch, ipv4: sbj.sshable.host, total_cores: sbj.total_cores, allocation_state: sbj.allocation_state, os_version: sbj.os_version, vm_count: sbj.vms_dataset.count}
+          when VmHostSlice
+            {vm_host: sbj.vm_host.ubid, data_center: sbj.vm_host.data_center, location: sbj.vm_host.location.display_name, arch: sbj.vm_host.arch}
```
jeremyevans: Up to you:

```suggestion
            vm_host = sbj.vm_host
            {vm_host: vm_host.ubid, data_center: vm_host.data_center, location: vm_host.location.display_name, arch: vm_host.arch}
```

PR #4647, opened 2026-01-22
Observe and page for metrics export backlog

shikharbhardwaj:


model/postgres/postgres_server.rb
```diff
@@ -296,12 +296,14 @@ def export_metrics(session:, tsdb_client:)
...
-    # export session does not use an event loop. Calling exec! on an SSH session
-    # with an active event loop is not thread-safe and leads to stuck sessions.
+    # Check archival and metrics backlog every 12 exports (similar to pulse
+    # check frequency) We do this in metrics export rather than pulse check
+    # because the metrics export session does not use an event loop. Calling
+    # exec! on an SSH session with an active event loop is not thread-safe and
+    # leads to stuck sessions.
     if session[:export_count] % 12 == 1
       observe_archival_backlog(session)
+      observe_metrics_backlog(session)
```
furkansahin: how costly is this? Should we try to separate the backlog observation for archival and metrics to different repetitions? Such as 
```ruby
observe_archival_backlog(session) if session[:export_count] % 12 == 1
observe_metrics_backlog(session)  if session[:export_count] % 12 == 7
```

shikharbhardwaj: It should be fairly quick, but also makes sense to do them separately.

PR #4648, opened 2026-01-22
Assemble the right path based on StorageDevice.name

mohi-kalantari:
During migrations from SPDK to ubiblk, there were vms which had disks in non-default StorageDevices which caused issues since we were assembling the wrong path for the storage dir.

### DIFF
--- a/rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
+++ b/rhizome/host/bin/convert-encrypted-dek-to-vhost-backend-conf
@@ -112,12 +112,13 @@ def write_vhost_conf(output_file, conf)
   yaml = YAML.dump(conf)
 
   # Sometimes key strings do not start with characters or numbers, as a result when
-  # encoded, they are wrapped in double quotes, breaking the migration binary because it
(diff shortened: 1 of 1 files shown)


prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -20,8 +20,10 @@ def self.assemble(vm_id)
       fail "VmHost does not have the right vhost block backend installed"
     end
 
+    storage_device_name = vm.vm_storage_volumes.first.storage_device.name
+    storage_dir = (storage_device_name == "DEFAULT") ? "/var/storage/#{vm.inhost_name}/0" : "/var/storage/devices/#{storage_device_name}/#{vm.inhost_name}/0"
```
pykello, line 24: Why not use the `root_vm_storage_path` function here?

mohi-kalantari: IIUC, we don't have access to that function in assemble and assemble is a static function of that class.

PR #4649, opened 2026-01-22
Handle a couple cases where a missing representative server causes exceptions

jeremyevans:
The overview page resulted in a few production exceptions a couple days ago. I reviewed the postgres route and found another case which is less likely (changing superuser password).

See individual commit messages for details. Best viewed ignoring whitespace.

### DIFF
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -377,7 +377,7 @@ class Clover
 
         DB.transaction do
           pg.update(superuser_password: password)
-          pg.representative_server.incr_update_superuser_password
+          pg.representative_server&.incr_update_superuser_password
           audit_log(pg, "reset_superuser_password")
         end
 
--- a/spec/routes/web/project/postgres_spec.rb
+++ b/spec/routes/web/project/postgres_spec.rb
@@ -574,28 +574,34 @@ RSpec.describe Clover, "postgres" do
         expect(page).to have_flash_error("Non read replica servers cannot be promoted.")
       end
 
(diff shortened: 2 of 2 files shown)


views/postgres/overview.erb
```diff
@@ -1,80 +1,85 @@
-<%
-vm = @pg.representative_server.vm
-%>
+<% if vm = @pg.representative_server&.vm %>
```
byucesoy, line 1: `representative_server` can only be `nil` during failovers. Instead of showing a blank page, we can fall back to target_* values we store in the resource.

routes/project/location/postgres.rb
```diff
@@ -377,7 +377,7 @@ class Clover
 
         DB.transaction do
           pg.update(superuser_password: password)
-          pg.representative_server.incr_update_superuser_password
+          pg.representative_server&.incr_update_superuser_password
```
byucesoy, line 380: This would skip updating the superuser password on the customer database. In such cases, we should return error saying something like "database is not ready for this operation, try again in few minutes". We should also skip updating our metadata database (i.e. the line before this one)

jeremyevans:
> Proper behavior when `representative_server` is missing bit tricky. Let me take a stab at this.

Sounds good. I'll close this, and I can review your PR when you submit it.

jeremyevans:
@byucesoy We had another production exception that db3e86c8f0a1d678d2b8811e6d020119072ba339 would have fixed. Can we merge that until you have time to work on a better fix?

PR #4650, opened 2026-01-22
Use private. instead of private- prefix for private DNS names

jeremyevans:
private- prefix can result in DNS labels longer than 63 characters, violating DNS spec.


jeremyevans:
I've confirmed this works for both load balancers (including ZeroSSL issued certificates with `private.` names) and PostgreSQL databases in staging.

PR #4652, opened 2026-01-22
Add daily CI

jeremyevans:
This CI will run once a weekday, running rake tasks that are too slow to run as part of normal CI.

Update the spec_separate and check_separate_requires rake tasks to correctly exit 1 on failure so CI will fail.  The spec_separate task was failing, so fix a spec file by adding a missing require.

### DIFF
--- a/.github/workflows/daily-ci.yml
+++ b/.github/workflows/daily-ci.yml
@@ -1,0 +1,65 @@
+name: Daily CI
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  '42 9 * * 1-5'
+
+env:
+  CLOVER_DATABASE_URL: postgres:///clover_test?user=clover
+
+jobs:
+  daily-ci:
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


jeremyevans:
The Daily CI here passed: https://github.com/ubicloud/ubicloud/actions/runs/21264309694/job/61199475985?pr=4652

I'm now going to update it so it doesn't run on pull requests, and notifies slack on failure (based on the E2E workflow as an example).

.github/workflows/daily-ci.yml
```diff
@@ -0,0 +1,68 @@
+name: Daily CI
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  '42 9 * * 1-5'
+
```
enescakir, line 10: You can add a global `env:`

```yml
env:
  CLOVER_DATABASE_URL: postgres:///clover_test?user=clover
```

jeremyevans: Makes sense, I made updated the commit to do use that approach.

PR #4655, opened 2026-01-23
Tighten file permissions for SPDK migration

mohi-kalantari:
Previously, several files were globally readable. While data remained secure because DEKs (Data Encryption Keys) were encrypted by KEKs (Key Encryption Keys), this change implements stricter access controls to align with security best practices.

### DIFF
--- a/prog/storage/migrate_spdk_vm_to_ubiblk.rb
+++ b/prog/storage/migrate_spdk_vm_to_ubiblk.rb
@@ -68,7 +68,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
 
   label def generate_vhost_backend_conf
     vm.vm_host.sshable.cmd("sudo host/bin/convert-encrypted-dek-to-vhost-backend-conf --encrypted-dek-file :root_dir_path/data_encryption_key.json --kek-file /dev/stdin --vhost-conf-output-file :vhost_conf_path --vm-name :inhost_name --device :device", inhost_name:, root_dir_path:, vhost_conf_path:, device: storage_device_name, stdin: vm.storage_secrets.to_json)
-    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :vhost_conf_path", inhost_name:, vhost_conf_path:)
+    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :vhost_conf_path && sudo chmod 600 :vhost_conf_path", inhost_name:, vhost_conf_path:)
     hop_ready_migration
   end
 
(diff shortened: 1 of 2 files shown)


prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -119,7 +119,7 @@ def vm_storage_volume
   end
 
   label def create_ubiblk_systemd_unit
-    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :root_dir_path/disk.raw", inhost_name:, root_dir_path:)
+    vm.vm_host.sshable.cmd("sh -c 'sudo chown :inhost_name::inhost_name :root_dir_path/disk.raw && sudo chmod 600 :root_dir_path/disk.raw :root_dir_path/disk.raw.bk && sudo rm -rf :kek_file_path'", inhost_name:, root_dir_path:, kek_file_path:)
```
pykello: I prefer to not change permissions of `disk.raw.bk`. Its permissions are setup so one exceptional user (`spdk`) can also access the file.

If for any reason we want to rollback to SPDK & use `disk.raw.bk`, then we'll need to re-setup its permissions.

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -68,7 +68,7 @@ def vm_storage_volume
 
   label def generate_vhost_backend_conf
     vm.vm_host.sshable.cmd("sudo host/bin/convert-encrypted-dek-to-vhost-backend-conf --encrypted-dek-file :root_dir_path/data_encryption_key.json --kek-file /dev/stdin --vhost-conf-output-file :vhost_conf_path --vm-name :inhost_name --device :device", inhost_name:, root_dir_path:, vhost_conf_path:, device: storage_device_name, stdin: vm.storage_secrets.to_json)
-    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :vhost_conf_path", inhost_name:, vhost_conf_path:)
+    vm.vm_host.sshable.cmd("sh -c 'sudo chown :inhost_name::inhost_name :vhost_conf_path && sudo chmod 600 :vhost_conf_path'", inhost_name:, vhost_conf_path:)
```
jeremyevans: You cannot safely use placeholders inside single or double quotes in a shell command, they are designed to be used outside of any quotes. It's possible that this particular case is not vulnerable because `inhost_name` and `vhost_conf_path` cannot contain `'`, but the general idea is unsafe.

This is also unnecessary, because the command will be run via a shell, you don't need to spawn a separate shell to wrap things.

```suggestion
    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :vhost_conf_path && sudo chmod 600 :vhost_conf_path", inhost_name:, vhost_conf_path:)
```

Same issue in your other change.

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -119,7 +119,7 @@ def vm_storage_volume
   end
 
   label def create_ubiblk_systemd_unit
-    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :root_dir_path/disk.raw", inhost_name:, root_dir_path:)
+    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :root_dir_path/disk.raw && sudo chmod 600 :root_dir_path/disk.raw && sudo rm -rf :kek_file_path", inhost_name:, root_dir_path:, kek_file_path:)
```
jeremyevans: Using `rm -r` when you aren't deleting a directory is unnecessarily risky:

```suggestion
    vm.vm_host.sshable.cmd("sudo chown :inhost_name::inhost_name :root_dir_path/disk.raw && sudo chmod 600 :root_dir_path/disk.raw && sudo rm -f :kek_file_path", inhost_name:, root_dir_path:, kek_file_path:)
```

PR #4664, opened 2026-01-26
Postgres install packages offline and remove apt-get update

furkansahin:
## Remove unnecessary apt-get update for PG
sudo apt-get update is necessary when we are setting up a new host since
the servers do not come with the most up-to date packages, always.
However, for any other product where we install rhizome on a VM, it is
unnecessary, because the VM images are supposed to come with an
up-to-date package set. Therefore, we are making the command run
conditional in the BootstrapRhizome prog.

## Update PG image and remove unnecessary aws ami configs in config.rb
The image version in config.rb is used while downloading the image to
E2E tests. If not updated, the controlplane keeps downloading the old
image.

The AWS specific configs are also not used anymore since the ami logic
is moved into the database.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -185,21 +185,16 @@ module Config
   override :github_gpu_ubuntu_2204_version, "20251208.1.0", string
   override :github_ubuntu_2204_aws_ami_version, "ami-0db39f24fb80dd3d2", string
   override :github_ubuntu_2404_aws_ami_version, "ami-0daff749acdc980cd", string
-  override :postgres_ubuntu_2204_version, "20251218.1.0", string
+  override :postgres_ubuntu_2204_version, "20260123.1.0", string
   override :postgres16_ubuntu_2204_version, "20250425.1.1", string
   override :postgres17_ubuntu_2204_version, "20250425.1.1", string
   override :postgres_paradedb_ubuntu_2204_version, "20260107.1.0", string
   override :postgres16_paradedb_ubuntu_2204_version, "20250901.1.0", string
   override :postgres17_paradedb_ubuntu_2204_version, "20250901.1.0", string
-  override :postgres16_lantern_ubuntu_2204_version, "20250103.1.0", string
-  override :postgres17_lantern_ubuntu_2204_version, "20250103.1.0", string
   override :ai_ubuntu_2404_nvidia_version, "20250505.1.0", string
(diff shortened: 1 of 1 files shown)


prog/bootstrap_rhizome.rb
```diff
@@ -55,9 +55,12 @@ def user
     pop "rhizome user bootstrapped and source installed" if retval&.dig("msg") == "installed rhizome"
 
     key_data = sshable.keys.map(&:private_key)
-    Util.rootish_ssh(sshable.host, user, key_data, <<SH, LOGIND_CONFIG:, SSHD_CONFIG:, public_keys: sshable.keys.map(&:public_key).join("\n"))
+    no_apt_update = (strand.stack.first["no_apt_update"] == true).to_s
+    Util.rootish_ssh(sshable.host, user, key_data, <<SH, LOGIND_CONFIG:, SSHD_CONFIG:, public_keys: sshable.keys.map(&:public_key).join("\n"), no_apt_update:)
 set -ueo pipefail
-sudo apt-get update
+if [ :no_apt_update = false ]; then
```
byucesoy: Instead of passing this variable, we can pass different script that may or maynot have sudo apt get update.

rhizome/postgres/lib/postgres_setup.rb
```diff
@@ -9,11 +9,9 @@ def initialize(version)
...
 
   def install_packages
-    # Check if the package list exist before installing packages, as Lantern
-    # images do not have a packages list and have the packages pre installed.
-    if File.exist?("/usr/local/share/postgresql/packages/#{@version}.txt")
-      r "xargs -a /usr/local/share/postgresql/packages/#{@version}.txt sudo apt-get -y install"
-      r "xargs -a /usr/local/share/postgresql/packages/common.txt sudo apt-get -y install"
+    # Check if the packages exist in the cache, if so, install them.
+    if File.exist?("/var/cache/postgresql-packages/#{@version}")
+      r "sudo install-postgresql-packages #{@version}"
```
serprex, line 14: where is `install-postgresql-packages` from?

furkansahin: the image, https://github.com/ubicloud/postgres-vm-images/pull/6

serprex: @heavycrystal we need to include this

PR #4665, opened 2026-01-26
Allow lsn_caught_up to work when primary is offline

furkansahin:
When a standby is catching up via WAL archive and the primary becomes
unreachable, lsn_caught_up would fail because it could not compare LSNs
with the primary.

Now it falls back to checking whether the standby has replayed all WAL
received from the archive, allowing failover to proceed even when the
primary is completely offline.

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -18,6 +18,8 @@ class PostgresServer < Sequel::Model
   include HealthMonitorMethods
   include MetricsTargetMethods
 
+  LSN_CAUGHT_UP_THRESHOLD = 80 * 1024 * 1024
+
   def self.victoria_metrics_client
     VictoriaMetricsResource.client_for_project(Config.postgres_service_project_id)
   end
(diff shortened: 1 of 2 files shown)


model/postgres/postgres_server.rb
```diff
@@ -188,7 +188,31 @@ def lsn_caught_up
       resource.representative_server
     end
 
-    !parent_server || lsn_diff(parent_server.current_lsn, current_lsn) < 80 * 1024 * 1024
+    return true unless parent_server
+
+    begin
+      parent_lsn = parent_server.current_lsn
+    rescue
```
enescakir: Should we catch these errors only?

https://github.com/ubicloud/ubicloud/blob/0c889cf92039475136b3bc3958cef13131df5f3a/prog/postgres/postgres_lockout.rb#L13

model/postgres/postgres_server.rb
```diff
@@ -188,7 +188,31 @@ def lsn_caught_up
...
+    return false if wal_receiver_status == "streaming"
+
+    # Not streaming - check if replay has caught up to what was received
+    receive_lsn = run_query("SELECT pg_last_wal_receive_lsn()").chomp
+
+    # If receive_lsn is empty (never connected), we're caught up to archive
+    return true if receive_lsn.empty?
+
+    replay_lsn = run_query("SELECT pg_last_wal_replay_lsn()").chomp
+    lsn_diff(receive_lsn, replay_lsn) < 80 * 1024 * 1024
```
enescakir: When we call `current_lsn`, does it return the replay lsn or the receive lsn in this case?

byucesoy (commented):
What if there is archival backlog on the primary side? With this change we basically remove the upper limit on the data loss during failovers.

furkansahin:
if the primary is unavailable, do you want to keep it in place and not perform the failover? The problem is, in case of total loss of the primary, the system cannot recover itself without letting standby going through.

byucesoy:
> If the primary is unavailable, do you want to keep it in place and not perform the failover?

Yes, if the difference is too big (bigger than 5 WAL files).

This is a tradeoff between availability and durability. Historically (both at Ubicloud and in older systems), we were OK with some difference for the sake of availability, but we always capped the max allowed difference.

One thing we can do would be using the last_known_lsn if we cannot reach to primary.

model/postgres/postgres_server.rb
```diff
@@ -190,7 +192,38 @@ def lsn_caught_up
...
+    receive_lsn = run_query("SELECT pg_last_wal_receive_lsn()").chomp
+    last_known_lsn = parent_server.lsn_monitor_ds.get(:last_known_lsn)
+
+    # Collect all available LSNs and use the largest
+    lsns = []
+    lsns << receive_lsn unless receive_lsn.empty?
+    lsns << last_known_lsn if last_known_lsn
+
+    # If no LSN is available, assume caught up
+    return true if lsns.empty?
```
byucesoy, line 222: I don't think we can assume this.

model/postgres/postgres_server.rb
```diff
@@ -190,7 +192,38 @@ def lsn_caught_up
...
     end
 
-    !parent_server || lsn_diff(parent_server.current_lsn, current_lsn) < 80 * 1024 * 1024
+    return true unless parent_server
+
+    begin
+      parent_lsn = parent_server.current_lsn
+    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
+      # Primary is unreachable - check if we've replayed all available WAL
+      return caught_up_to_archive?(parent_server)
```
byucesoy, line 201: Name of this function doesn't make sense anymore.

byucesoy: Also, from design perspective, I think it would be cleaner if we just don't return the result of the `caught_up_to_archive`, but we would have a function that can return `parent_server.current_lsn` or `parent_server.lsn_monitor_ds.get(:last_known_lsn)` or `nil` in this preference order and we would still use `lsn_diff` at the end of this function to return the result.

byucesoy (commented):
Did you use Claude for this PR? I’m asking because the styling and a few of the unexpected patterns felt like Claude's style. Using it is fine, but I think Claude tends to make mistakes regarding how PostgreSQL would behave in certain scenarios because there are not enough online resources about them.

PR #4667, opened 2026-01-26
Pre create aws subnets per az

furkansahin:
## Add aws_subnet table for pre-created AZ subnets
Previously, AWS subnets were created per-NIC at VM provision
time, leading to a proliferation of subnets and hitting AWS
limits. This migration adds an aws_subnet table to track
pre-created subnets, one per availability zone per VPC.

The nic_aws_resource table gains an aws_subnet_id foreign key
to reference which shared subnet each NIC uses.

## Add AwsSubnet model and wire up associations
Introduce the AwsSubnet Sequel model with associations to
PrivateSubnetAwsResource and LocationAwsAz. Update
NicAwsResource and PrivateSubnetAwsResource with the
corresponding reverse associations.

## Pre-create one AWS subnet per AZ at VPC creation time
Instead of creating subnets on-demand per NIC, SubnetNexus
now pre-allocates AwsSubnet records (one per AZ) with IPv4
CIDRs during assembly. VpcNexus then materializes them as
actual AWS subnets once the VPC and its IPv6 block are
available.

The destroy path gains a delete_az_subnets label. On VPC
deletion, DependencyViolation is re-raised for strand retry
rather than looping back explicitly.

## Allocate NICs from pre-created AZ subnets
NicNexus.assemble now selects an AwsSubnet at assembly time
based on the requested availability zone, and allocates an
IPv4 address from that subnet's CIDR (respecting the five
AWS-reserved addresses). The chosen subnet ID is stored in
the strand frame.

Aws::NicNexus no longer creates or manages per-NIC subnets
for new VPCs. Legacy per-NIC subnets are still deleted during
NIC teardown; shared AZ subnets are left for VPC cleanup.

## Add backfill prog for existing VPCs
BackfillAwsSubnets is a one-shot prog that migrates VPCs
created before the per-AZ subnet architecture. It handles
two cases:

- /26 VPCs (legacy): discovers the single existing subnet
  and creates one AwsSubnet record
- /16 VPCs (current): inventories existing per-NIC subnets,
  groups them by AZ, creates AwsSubnet records, then
  provisions new AWS subnets for any uncovered AZs

In both cases it links existing NICs to the appropriate
AwsSubnet via nic_aws_resource.aws_subnet_id.

## Update unrelated specs for SubnetNexus.assemble changes
SubnetNexus.assemble now creates AwsSubnet records during
assembly, which requires a LocationAwsAz to exist for AWS
locations. Add the necessary AZ setup to specs that assemble
AWS subnets indirectly.

## Support dynamic AWS subnet sizing based on VPC range
Pass ipv4_range_size to create_aws_subnet_records so AZ subnets can be
sized relative to the VPC. The subnet prefix is calculated as
min(vpc_size + 8, 28) to respect AWS's /28 minimum subnet size.

Handle edge cases where the VPC is too small:
- If VPC prefix equals subnet prefix, force single-AZ mode and use the
  entire VPC as the subnet
- If VPC is smaller than /28, raise an error since AWS doesn't allow
  subnets smaller than /28
- If VPC can't fit all AZs, log a warning and skip the extra AZs

## Limit GitHub runner VPCs to a single AZ subnet
GitHub runners are ephemeral single-VM workloads that do not
need subnets pre-created in every availability zone. Pass
limit_to_single_az to SubnetNexus.assemble so only one
randomly chosen AZ gets an AwsSubnet record, reducing
unnecessary AWS resource creation.

## Switch from limit_to_single_az to preferred_azs
Instead of passing a flag to only ask for a single az, the caller can
specify their preferred azs. This way, if the ask is a single az, they
can list a single az. This was a better design because the github
runners were excluding the az "a" and preferred a single az. Instead of
introducing multiple flags to the prog for the same job, I decided to
use preferred_azs.

### DIFF
--- a/prog/github/github_runner_nexus.rb
+++ b/prog/github/github_runner_nexus.rb
@@ -57,7 +57,7 @@ class Prog::Github::GithubRunnerNexus < Prog::Base
     boot_image = label_data["boot_image"]
     location_id = Location::GITHUB_RUNNERS_ID
     size = label_data["vm_size"]
-    exclude_availability_zones = []
+    preferred_azs = []
     alternative_families = []
     alien_ratio = project.get_ff_aws_alien_runners_ratio || 0
     if github_runner.spill_over_set? || (x64? && rand < alien_ratio)
@@ -65,7 +65,7 @@ class Prog::Github::GithubRunnerNexus < Prog::Base
       location_id = Config.github_runner_aws_location_id
       size = Option.aws_instance_type_name("m7a", label_data["vcpus"])
       alternative_families << "m7i" << "m6a"
-      exclude_availability_zones << "a" # eu-central-1a is usually give capacity errors
+      preferred_azs << Location[location_id].azs.reject { |az| az == "a" }.sample # eu-central-1a is usually give capacity errors
     end
 
     ps = Prog::Vnet::SubnetNexus.assemble(
@@ -73,7 +73,7 @@ class Prog::Github::GithubRunnerNexus < Prog::Base
       location_id:,
       allow_only_ssh: true,
       ipv4_range_size: 28,
-      limit_to_single_az: true
+      preferred_azs:
     ).subject
 
     vm_st = Prog::Vm::Nexus.assemble_with_sshable(
@@ -89,7 +89,6 @@ class Prog::Github::GithubRunnerNexus < Prog::Base
       arch: label_data["arch"],
       swap_size_bytes: 4294963200, # ~4096MB, the same value with GitHub hosted runners
       private_subnet_id: ps.id,
-      exclude_availability_zones:,
       alternative_families:
     )
 
--- a/prog/vnet/subnet_nexus.rb
+++ b/prog/vnet/subnet_nexus.rb
@@ -3,7 +3,7 @@
 class Prog::Vnet::SubnetNexus < Prog::Base
   subject_is :private_subnet
 
-  def self.assemble(project_id, name: nil, location_id: Location::HETZNER_FSN1_ID, ipv6_range: nil, ipv4_range: nil, allow_only_ssh: false, firewall_id: nil, firewall_name: nil, ipv4_range_size: nil, limit_to_single_az: false)
+  def self.assemble(project_id, name: nil, location_id: Location::HETZNER_FSN1_ID, ipv6_range: nil, ipv4_range: nil, allow_only_ssh: false, firewall_id: nil, firewall_name: nil, ipv4_range_size: nil, preferred_azs: [])
     unless (project = Project[project_id])
       fail "No existing project"
     end
@@ -55,7 +55,7 @@ class Prog::Vnet::SubnetNexus < Prog::Base
       prog = if location.aws?
         # Create PrivateSubnetAwsResource and pre-create AwsSubnet records for each AZ
         ps_aws_resource = PrivateSubnetAwsResource.create_with_id(ps.id)
-        create_aws_subnet_records(ps, ps_aws_resource, location, ipv4_range_size, limit_to_single_az)
+        create_aws_subnet_records(ps, ps_aws_resource, location, ipv4_range_size, preferred_azs:)
         "Vnet::Aws::VpcNexus"
       else
         "Vnet::Metal::SubnetNexus"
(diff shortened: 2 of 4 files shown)


prog/vnet/aws/backfill_aws_subnets.rb
```diff
@@ -0,0 +1,214 @@
+# frozen_string_literal: true
```
furkansahin, line 1: this file will be used during the migration and then we'll remove it in a following pr.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -96,6 +91,49 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
...
+      subnet = if aws_subnet.subnet_id
+        client.describe_subnets({filters: [{name: "subnet-id", values: [aws_subnet.subnet_id]}]}).subnets[0]
+      else
+        ipv6_cidr = vpc_ipv6.nth_subnet(64, idx)
+
+        client.create_subnet({
+          vpc_id: private_subnet_aws_resource.vpc_id,
+          cidr_block: aws_subnet.ipv4_cidr.to_s,
+          ipv_6_cidr_block: ipv6_cidr.to_s,
+          availability_zone: location.name + aws_subnet.location_aws_az.az,
```
enescakir, line 113: Do you need to add any delimiter between them?

prog/vnet/nic_nexus.rb
```diff
@@ -37,4 +39,54 @@ def self.gen_mac
       "%0.2X" % it
     }.join(":").downcase
   end
+
+  def self.select_aws_subnet(subnet, availability_zone, exclude_availability_zones)
+    ps_aws_resource = subnet.private_subnet_aws_resource
+    return nil unless ps_aws_resource
```
enescakir:
```suggestion
    return unless ps_aws_resource
```

prog/vnet/nic_nexus.rb
```diff
@@ -37,4 +39,54 @@ def self.gen_mac
...
+      end
+    end
+
+    # Fallback to any available subnet
+    available_subnets = AwsSubnet
+      .where(private_subnet_aws_resource_id: ps_aws_resource.id)
+      .exclude(location_aws_az_id: excluded_az_ids)
+      .all
+
+    available_subnets.sample
```
enescakir: no need to set to a variable

```suggestion
    AwsSubnet
      .where(private_subnet_aws_resource_id: ps_aws_resource.id)
      .exclude(location_aws_az_id: excluded_az_ids)
      .all
      .sample
```

or you can pick random in DB

```suggestion
    AwsSubnet
      .where(private_subnet_aws_resource_id: ps_aws_resource.id)
      .exclude(location_aws_az_id: excluded_az_ids)
      .order(Sequel.function(:random))
      .first
```

prog/vnet/nic_nexus.rb
```diff
@@ -37,4 +39,54 @@ def self.gen_mac
...
+    excluded_az_ids = if exclude_availability_zones.empty?
+      []
+    else
+      subnet.location.location_aws_azs_dataset.where(az: exclude_availability_zones).select_map(:id)
+    end
+
+    # Try to find subnet for preferred AZ
+    if availability_zone
+      location_aws_az = subnet.location.location_aws_azs_dataset.where(az: availability_zone).first
+      if location_aws_az
```
enescakir: nit: combining where and first
```suggestion
      if (location_aws_az = subnet.location.location_aws_azs_dataset.first(az: availability_zone))
```

or

```suggestion
      if (location_aws_az = subnet.location.location_aws_azs_dataset[az: availability_zone])
```

prog/vnet/nic_nexus.rb
```diff
@@ -37,4 +39,54 @@ def self.gen_mac
...
+        aws_subnet = AwsSubnet.where(
+          private_subnet_aws_resource_id: ps_aws_resource.id,
+          location_aws_az_id: location_aws_az.id
+        ).first
```
enescakir:
```suggestion
        aws_subnet = AwsSubnet.first(
          private_subnet_aws_resource_id: ps_aws_resource.id,
          location_aws_az_id: location_aws_az.id
        )
```

prog/vnet/aws/vpc_nexus.rb
furkansahin: no

PR #4669, opened 2026-01-26
Strip host from cache proxy log lines before tallying

enescakir:
Each runner VM has a unique IP address, which creates a unique log key for every request in the cache proxy logs. This explosion of unique keys causes issues with our log provider. By stripping the host suffix (e.g., " host: 172_27_211_181:51123") before tallying, we consolidate logs by request type rather than by source, significantly reducing cardinality.

### DIFF
--- a/prog/github/github_runner_nexus.rb
+++ b/prog/github/github_runner_nexus.rb
@@ -484,7 +484,7 @@ class Prog::Github::GithubRunnerNexus < Prog::Base
     end
 
     if (cache_proxy_log = vm.sshable.cmd("sudo cat /var/log/cacheproxy.log", log: false))
-      cache_proxy_log_line_counts = cache_proxy_log.lines.each(&:strip!).reject(&:empty?).tally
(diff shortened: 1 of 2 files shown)


prog/github/github_runner_nexus.rb
```diff
@@ -484,7 +484,7 @@ def collect_final_telemetry
     end
 
     if (cache_proxy_log = vm.sshable.cmd("sudo cat /var/log/cacheproxy.log", log: false))
-      cache_proxy_log_line_counts = cache_proxy_log.lines.each(&:strip!).reject(&:empty?).tally
+      cache_proxy_log_line_counts = cache_proxy_log.lines.map { it.strip.gsub(/ host: \S+/, "") }.reject(&:empty?).tally
```
jeremyevans: Seems better to avoid multiple string allocations per line:

```suggestion
      cache_proxy_log_line_counts = cache_proxy_log.lines.each(&:strip!).each { it.gsub!(/ host: \S+/, "") }.reject(&:empty?).tally
```

PR #4673, opened 2026-01-27
Add AWS resource cleanup step to E2E workflow

enescakir:
E2E tests occasionally leak AWS resources (EC2 instances, ENIs, EIPs, VPCs) when they fail or timeout. This adds a cleanup step that runs unconditionally at the end of the workflow to delete any resources created during the test run.

The step extracts instance IDs from the logs, queries AWS for associated resources (network interfaces, elastic IPs, VPCs), then deletes them in the correct order. VPC cleanup handles dependent resources (internet gateways, subnets, security groups, route tables) in parallel before deleting the VPC itself.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -256,3 +256,75 @@ jobs:
                 - title: "Last Line"
                   short: false
                   value: "${{ steps.extract_last.outputs.LAST_LINE }}"
+
+    - name: Cleanup AWS resources
+      if: always()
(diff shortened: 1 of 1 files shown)


aamederen:
After this one, maybe we can also add a step to check if there're any leaked resources after successful runs.

enescakir:
Example run: https://github.com/ubicloud/ubicloud/actions/runs/21395466664/job/61592558343

PR #4677, opened 2026-01-27
Support Stop and Start for VMs

jeremyevans:
In addition to supporting stop and start for VMs, this has a number of other related changes:

* Restart semaphore now has an effect when VM is in unavailable state
* Restart (and start/stop) results in an error for AWS VMs (better than silently accepted and ignoring)
* Restart (and start/stop) results in an error if the metal VM is not in an expected state
* Show unavailable as a display state for VMs
* Do not page when a VM becomes unavailable, as the most likely case is the user shut it down from inside the VM
* Add unavailable VMs page to admin site, so admins can see which VMs are unavailable (now that it doesn't page)

Not paging for unavailable VMs is probably the most controversial change here. I'm not opposed to paging for unavailable VMs if we know for sure that the unavailability is due to a problem that an admin can fix. However, as long as the unavailability is caused by customer action and out of the admin's control, I don't think we should page for it. Blindly restarting the VMs to address the page is generally the wrong solution, because the user may have shut down the VM for security or other reasons. With this change, the user can start the VM if they have shut it down, via the web or the API/CLI.

### DIFF
--- a/prog/vm/metal/nexus.rb
+++ b/prog/vm/metal/nexus.rb
@@ -262,6 +262,7 @@ class Prog::Vm::Metal::Nexus < Prog::Base
 
     when_checkup_set? do
       unless available?
+        update_stack("reason_determined" => false)
         hop_unavailable
       end
       decr_checkup
@@ -347,7 +348,52 @@ class Prog::Vm::Metal::Nexus < Prog::Base
 
     if available?
       decr_checkup
+      Page.from_tag_parts("VmExit", vm.ubid)&.incr_resolve
       hop_wait
+    elsif !frame["reason_determined"]
+      begin
+        result, invocation_id = host.sshable.cmd("systemctl show -p Result -p InvocationID --value :vm_name", vm_name:).strip.split("\n")
+
+        reason = if %w[signal core-dump oom-kill timeout success].include?(result)
+          result
+        else
+          host.sshable.cmd(<<~END, invocation_id:).strip
+            sudo journalctl _SYSTEMD_INVOCATION_ID=:invocation_id -o cat -n 50 --no-pager | \
+              grep -m1 -oF \
+                -e 'ACPI Shutdown signalled' \
+                -e 'vCPU thread panicked' \
+                -e 'VCPU generated error' \
+                -e 'thread panicked'
+          END
+        end
+      rescue Sshable::SshError, *Sshable::SSH_CONNECTION_ERRORS
+        reason = "unknown"
+      end
+
+      case reason
+      when "ACPI Shutdown signalled", "success"
+        msg = if reason == "success"
+          # ExecStop succeeded: someone ran systemctl stop or ch-remote shutdown-vmm.
+          # Deliberate action, not an issue, don't page
+          "VM stopped by operator"
+        else
+          # Customer shutdown their VM from inside the VM, don't page
+          "VM stopped by guest ACPI shutdown"
+        end
+        Clog.emit(msg, {vm_exit: {vm: vm.ubid}})
+        decr_checkup
+        incr_stop
+        hop_stopped
+      else
+        # Unexpected exit: signal, crash, unknown cause.
+        # Page with the reason in the summary.
+        Prog::PageNexus.assemble(
+          "#{vm.ubid} stopped unexpectedly (#{reason})",
+          ["VmExit", vm.ubid], vm.ubid,
+          extra_data: {vm_host: host.ubid, result:, reason:}
+        )
+        update_stack("reason_determined" => true)
+      end
     end
 
     nap 30
--- a/spec/prog/vm/metal/nexus_spec.rb
+++ b/spec/prog/vm/metal/nexus_spec.rb
@@ -808,6 +808,8 @@ RSpec.describe Prog::Vm::Metal::Nexus do
       vm.incr_checkup
       expect(nx).to receive(:available?).and_return(false)
       expect { nx.wait }.to hop("unavailable")
+      frame = st.stack[0]
+      expect(frame["reason_determined"]).to be false
 
       vm.incr_checkup
       expect(nx).to receive(:available?).and_return(true)
@@ -915,14 +917,68 @@ RSpec.describe Prog::Vm::Metal::Nexus do
       expect { nx.unavailable }.to hop("stopped")
     end
 
-    it "naps if vm is still unavailable" do
+    it "naps if vm is still unavailable and the reason has been determined" do
+      refresh_frame(nx, new_values: {"reason_determined" => true})
       expect(nx).to receive(:available?).and_return(false)
       expect { nx.unavailable }.to nap(30)
     end
 
-    it "hops to wait if vm is available" do
+    it "pages and naps if vm is unavailable and systemctl show returns oom-kill" do
+      expect(sshable).to receive(:_cmd).with("systemctl show -p Result -p InvocationID --value #{vm.inhost_name}").and_return("oom-kill\nfoo\n")
+      expect(nx).to receive(:available?).and_return(false)
+      expect { nx.unavailable }.to nap(30)
+        .and change { Page.get(:summary) }.from(nil).to("#{vm.ubid} stopped unexpectedly (oom-kill)")
+      frame = st.stack[0]
+      expect(frame["reason_determined"]).to be true
+    end
+
+    it "pages and naps if vm is unavailable and systemctl command results in an error" do
+      expect(sshable).to receive(:_cmd).with("systemctl show -p Result -p InvocationID --value #{vm.inhost_name}").and_raise(Sshable::SshError.new("", "", "", 1, nil))
+      expect(nx).to receive(:available?).and_return(false)
+      expect { nx.unavailable }.to nap(30)
+        .and change { Page.get(:summary) }.from(nil).to("#{vm.ubid} stopped unexpectedly (unknown)")
+      frame = st.stack[0]
+      expect(frame["reason_determined"]).to be true
+    end
+
+    it "hops to stopped if vm is unavailable and systemctl show returns success" do
+      nx.incr_checkup
+      expect(sshable).to receive(:_cmd).with("systemctl show -p Result -p InvocationID --value #{vm.inhost_name}").and_return("success\nfoo\n")
+      expect(Clog).to receive(:emit).with("VM stopped by operator", Hash)
+      expect(nx).to receive(:available?).and_return(false)
+      expect { nx.unavailable }.to hop("stopped")
+        .and change(nx, :checkup_set?).from(true).to(false)
+        .and change(nx, :stop_set?).from(false).to(true)
+    end
+
+    it "hops to stopped if vm is unavailable and systemctl show returns other value and journalctl shows ACPI Shutdown signalled" do
+      expect(sshable).to receive(:_cmd).with("systemctl show -p Result -p InvocationID --value #{vm.inhost_name}").and_return("other\nfoo\n")
+      expect(sshable).to receive(:_cmd).with(<<~END).and_return("ACPI Shutdown signalled\n")
+        sudo journalctl _SYSTEMD_INVOCATION_ID=foo -o cat -n 50 --no-pager | \
+          grep -m1 -oF \
+            -e 'ACPI Shutdown signalled' \
+            -e 'vCPU thread panicked' \
+            -e 'VCPU generated error' \
+            -e 'thread panicked'
+      END
+      expect(Clog).to receive(:emit).with("VM stopped by guest ACPI shutdown", Hash)
+      expect(nx).to receive(:available?).and_return(false)
+      expect { nx.unavailable }.to hop("stopped")
+        .and change(nx, :stop_set?).from(false).to(true)
+    end
+
+    it "decrements checkup and hops to wait if vm is available" do
+      nx.incr_checkup
       expect(nx).to receive(:available?).and_return(true)
       expect { nx.unavailable }.to hop("wait")
+        .and change(nx, :checkup_set?).from(true).to(false)
+    end
+
+    it "resolves page and hops to wait if vm is available" do
+      page = Prog::PageNexus.assemble("#{vm.ubid} stopped unexpectedly", ["VmExit", vm.ubid], vm.ubid).subject
+      expect(nx).to receive(:available?).and_return(true)
+      expect { nx.unavailable }.to hop("wait")
+        .and change { page.reload.resolve_set? }.from(false).to(true)
     end
   end
 


prog/vm/metal/nexus.rb
```diff
@@ -310,9 +323,10 @@ def write_params_json
       hop_start_after_host_reboot
     end
 
-    when_restart_set? do
-      register_deadline("wait", 5 * 60)
-      hop_restart
+    check_restart
```
furkansahin: If we are exiting unavailable state with a restart action, I think we should get back to `unavailable`. Because we suddenly declare availability and there are other operations which may get triggered in `wait` state before we make sure the availability of the vm.

jeremyevans: Changed this so that if `restart` semaphore is set when the VM is in `unavailable`, it issues the restart command and removes the semaphore, but it does not leave `unavailable` state.

prog/vm/metal/nexus.rb
```diff
@@ -262,7 +262,6 @@ def write_params_json
 
     when_checkup_set? do
       unless available?
-        register_deadline("wait", 2 * 60)
```
enescakir, line 265: I’m not sure if we should disable VM unavailability pages. It’s unlikely that anyone regularly checks the admin panel for unavailable VMs.

Stopping a VM from inside the VM itself is a very rare case and usually tied to a specific customer setup. After we reach out, they generally change this pattern.

I’m leaning toward keeping the pages.

If we want to avoid generating pages for VMs that are intentionally shut down by customers, we might detect that with a similar command @pykello shared earlier

```ruby
vm.vm_host.sshable.cmd("sudo journalctl -u :inhost_name | grep 'ACPI Shutdown'", inhost_name: vm.inhost_name)
```

views/admin/unavailable_vms.erb
```diff
@@ -0,0 +1,3 @@
+<% @page_title = "Unavailable VMs" %>
+
+<%== part("components/table", title: "Unavailable VMs", table_class: "unavailable-vms-table", data: @strands) %>
```
enescakir: We already can filter VMs by display state (we can add new states to the list), so I’m not sure if it’s worth maintaining a separate page.

Image

model/vm.rb
```diff
@@ -130,6 +131,28 @@ def display_state
...
+    display_state == "running"
+  end
+
+  def can_stop?
+    case display_state
+    when "running", "restarting", "unavailable", "rebooting"
+      true
+    else
+      false
+    end
```
enescakir:
```suggestion
    %w[running restarting unavailable rebooting].include?(display_state)
```

model/vm.rb
```diff
@@ -130,6 +131,28 @@ def display_state
...
+  end
+
+  def can_start?
+    case display_state
+    when "unavailable", "stopped"
+      true
+    else
+      false
+    end
```
enescakir:
```suggestion
    %w[unavailable stopped].include?(display_state)
```

prog/vm/metal/nexus.rb
```diff
@@ -293,15 +292,36 @@ def write_params_json
     hop_wait
   end
 
+  label def start_after_stop
+    decr_start
+    host.sshable.cmd("sudo host/bin/setup-vm start :vm_name", vm_name:)
```
enescakir: Does `setup-vm` script support `start` action?

https://github.com/ubicloud/ubicloud/blob/main/rhizome/host/bin/setup-vm

prog/vm/metal/nexus.rb
```diff
@@ -447,6 +482,9 @@ def final_clean_up
     hop_wait
   end
 
+  def check_restart
+  end
+
```
enescakir: Empty method?

views/vm/settings.erb
```diff
@@ -16,24 +16,67 @@
...
+              <div id="vm-restart-<%= @vm.ubid %>" class="mt-5 sm:ml-6 sm:mt-0 sm:flex sm:flex-shrink-0 sm:items-center">
+                <div class="col-span-12 sm:col-span-2 flex justify-end items-end">
+                  <%== part("components/form/submit_button", text: "Stop", extra_class: "stop-btn") %>
+                </div>
+              </div>
+            </div>
+          <% end %>
+        </div>
+      <% end %>
     <% end %>
```
enescakir: These are very similar. Is it possible to loop over an [[action, message]] list?

prog/vm/metal/nexus.rb
jeremyevans: I'm fine keeping the pages as long as we only page for problems that the on-call can and should fix, and not for things caused by the customer. If the customer is causing the issue, I don't think we should page. With the support in this PR, if the customer does something to shutdown their VM, they can use the `start` action to start it by themselves.

Are we sure that `ACPI Shutdown` is the only way for a customer to shut down their own VM from inside the VM? What if they are running a custom kernel module and it crashes their VM? That's not something I think we should page for.

Let's say we do page. What is the on-call supposed to do? The natural reaction is to attempt to start the VM, as that is the only way to clear the page. However, I don't think there is any indication that that is the correct action to take. Unless we can provide instructions to the on-call for how this case should be handled, I think it's better not to page.

Customer's stopping their own VMs is already not "very rare", and certainly as we scale, paging for every case where a customer shuts down their VM seems unsustainable. It's true that customers generally avoid doing this once we inform them about it, but I don't think we can or should rely on that as we scale.

prog/vm/metal/nexus.rb
jeremyevans: No, that's a bug. It should be `sudo systemctl start :vm_name`. Thanks for catching this.

views/vm/settings.erb
jeremyevans: I think so. I can switch to that approach.

prog/vm/metal/nexus.rb
jeremyevans: Yep. Sorry, left over from refactoring for the change Furkan requested.

model/vm.rb
jeremyevans: Sure, I'll make that change and the other one.

views/admin/unavailable_vms.erb
jeremyevans: The page doesn't search VMs, it searches VM strands. However, maybe we don't need a separate template and route for it. I can switch it to `/autoforme/Strand/search/1?prog=Vm%3A%3AMetal%3A%3ANexus&label=unavailable`. However, then we need to add an easy way to get from the Strand page to the page for the Strand's subject.

prog/vm/metal/nexus.rb
enescakir: > Are we sure that ACPI Shutdown is the only way for a customer to shut down their own VM from inside the VM? What if they are running a custom kernel module and it crashes their VM? That's not something I think we should page for.

We can also ask these questions in the opposite order. Are we sure the only reason for an unavailable VM is customer actions? What if the VM is actually unavailable due to a misconfiguration or a systemd service crash?

> Let's say we do page. What is the on-call supposed to do? The natural reaction is to attempt to start the VM, as that is the only way to clear the page. However, I don't think there is any indication that that is the correct action to take.

The oncall can check the serial logs and try to reduce false positives, for example cases where an "ACPI Shutdown" alone isn’t sufficient.

> Customer's stopping their own VMs is already not "very rare", and certainly as we scale, paging for every case where a customer shuts down their VM seems unsustainable

All the recent on-call pain was caused by a single customer.

> Unless we can provide instructions to the on-call for how this case should be handled, I think it's better not to page.

We don't have instructions for most PostgreSQL cases (even though I've asked several times). In practice, I usually search old Slack messages and try to figure things out.

For PostgreSQL, at least we have an owning team and most of the team has prior PG experience. In contrast, VM-related issues don't have a clear owner. Because no one owns it, no one writes instructions or is an obvious point of contact when something goes wrong.

IMHO, we should try to reduce false positives instead of disabling it completely.

prog/vm/metal/nexus.rb
enescakir: `restart` also calls other commands. I'm not sure if we need them for `start` or `systemctl start` is enough. cc @pykello 

https://github.com/ubicloud/ubicloud/blob/27b745a8c0627a8d72468d98814797584b9b5a7c/rhizome/host/lib/vm_setup.rb#L94-L97

enescakir: This could be a good candidate to cover with E2E tests.

jeremyevans:
@enescakir Pushed changes to address all of your comments, other than the one to stop paging for VM unavailability.

prog/vm/metal/nexus.rb
jeremyevans: > > Are we sure that ACPI Shutdown is the only way for a customer to shut down their own VM from inside the VM? What if they are running a custom kernel module and it crashes their VM? That's not something I think we should page for.
> 
> We can also ask these questions in the opposite order. Are we sure the only reason for an unavailable VM is customer actions? What if the VM is actually unavailable due to a misconfiguration or a systemd service crash?

We cannot be sure. I can recall multiple cases where customers shutting down VMs caused a VM unavailability page, and no cases where it was caused by a misconfiguration or systemd service crash. From my discussions with @fdr, misconfiguration/systemd service crash was the more likely cause in the past, but it doesn't seem like the more likely cause any more.

If we can automatically determine that unavailability is due to misconfiguration, systemd service crash, or something similar that requires on-call involvement, I'm fine paging. However, paging when the most likely current cause is customer action seems unwise to me.

> > Let's say we do page. What is the on-call supposed to do? The natural reaction is to attempt to start the VM, as that is the only way to clear the page. However, I don't think there is any indication that that is the correct action to take.
> 
> The oncall can check the serial logs and try to reduce false positives, for example cases where an "ACPI Shutdown" alone isn’t sufficient.

This assumes the on-call has sufficient experience to know what to look for. I expect that in many cases, it may not be apparent the reason the VM service is not running. In that case, what should the on-call do? Starting the VM is in many cases the wrong decision, I think, but that's currently the primary way to clear the page.

> > Customer's stopping their own VMs is already not "very rare", and certainly as we scale, paging for every case where a customer shuts down their VM seems unsustainable
> 
> All the recent on-call pain was caused by a single customer.

However, the issue is this can be caused by any customer. We should consider ourselves lucky this is not a bigger problem already, because as we scale, it will assuredly become a bigger problem if we take no action.

> > Unless we can provide instructions to the on-call for how this case should be handled, I think it's better not to page.
> 
> We don't have instructions for most PostgreSQL cases (even though I've asked several times). In practice, I usually search old Slack messages and try to figure things out.
> 
> For PostgreSQL, at least we have an owning team and most of the team has prior PG experience. In contrast, VM-related issues don't have a clear owner. Because no one owns it, no one writes instructions or is an obvious point of contact when something goes wrong.

I agree with you that we probably should have better on-call documentation on both PG and VMs.

> IMHO, we should try to reduce false positives instead of disabling it completely.

I can agree that reducing false positives by looking for `ACPI Shutdown` seems better than doing nothing. I'll leave to @fdr to decide whether we not to stop paging completely for this, or whether we should continue to page for some cases, but search for `ACPI Shutdown` and not page in that case.

fdr: Let me doing a bit of investigation in the cloud hypervisor code to see if we can catch *any* kind of software exit rather than....something else.

fdr (commented):
I think other people can read lots of this patch better than me, e.g. markup.

 I guess we didn't find a good way to make sure people realize that `stopped` doesn't change billing at all, although the current nomenclature makes sense enough. I think that's something we can live with.

Given `unavailable` didn't seem to cut it for my original vague idea of some way to subtly signal to the customer that "it's not available but you're paying full rate!" -- perhaps it's too alarming -- I think that's something we can deal with, we should just expect the occasional angry email not understanding the billing plus chargebacks (these are not free for our account reputation) even though it's like Azure (the second largest cloud and not AWS (the first largest, where running "halt" or sending the stop api call will make you pay less)

Unless, someone has a better idea...I would really like to find some kind of signaling to reinforce to AWS-transitioning customers that "stopped" does not mean "reduced price" on this platform

My action item, though: figure out a way to discriminate any-kind-kernel-exit from systemd kill, oom kill, etc.

prog/vm/metal/nexus.rb
fdr: here it is. https://github.com/ubicloud/ubicloud/pull/4691. It's an investigation, not something to be merged. So more than a policy decision, here's some decision support, too.

jeremyevans:
I've updated this to incorporate the approach in #4691. I've also tested it in my local development environment with the following code:

```ruby
failures = []

1.times do |i|
  system("cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm stop")
  sleep 60
  failures << [i, :after_stop] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: stopped\n"
  system("cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm start")
  sleep 60
  failures << [i, :after_start_after_stop] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: running\n"
  system("cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm ssh sudo shutdown now")
  sleep 60
  failures << [i, :after_shutdown] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: stopped\n"
  system("cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm start")
  sleep 60
  failures << [i, :after_start_after_shutdown] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: running\n"
  puts "failures: #{failures.count}"
end
```

This tests both the stop action as well as shutting down the VM from inside the VM. No failure. I'm going to run this with 100 iterations and see if there is a failure (should take about 3:40).

jeremyevans:
I had to adjust the iteration code to add more time between commands, because it may take more than a minute for a shutdown inside the VM to reach `stopped` state. Here's what I changed the iteration code to:

```ruby
  system("cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm stop")
  sleep 120
  failures << [i, :after_stop] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: stopped\n"
  system("cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm start")
  sleep 120
  failures << [i, :after_start_after_stop] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: running\n"
  system("timeout 60s cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm ssh sudo shutdown now")
  sleep 120
  failures << [i, :after_shutdown] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: stopped\n"
  system("cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm start")
  sleep 120
  failures << [i, :after_start_after_shutdown] unless `cli/ubi-dev vm vmrsgaam747t6kwd787nn6tjtm show -f state` == "state: running\n"
  puts "#{i} failures: #{failures.count} last: #{failures.last.inspect}"
```

I ran this overnight. It completed 63 iterations with no failures.

jeremyevans:
I stopped the script after 84 iterations. Still no failures.

prog/vm/metal/nexus.rb
```diff
@@ -347,7 +348,52 @@ def write_params_json
...
+          host.sshable.cmd(<<~END, invocation_id:).strip
+            sudo journalctl _SYSTEMD_INVOCATION_ID=:invocation_id -o cat -n 50 --no-pager | \
+              grep -m1 -oF \
+                -e 'ACPI Shutdown signalled' \
+                -e 'vCPU thread panicked' \
+                -e 'VCPU generated error' \
+                -e 'thread panicked'
+          END
+        end
+      rescue Sshable::SshError
```
enescakir: These commands can also raise an `*SSH_CONNECTION_ERRORS` exception if the SSH connection is broken. If that’s the case here, it likely indicates an issue with the VM host itself. I’m not sure whether we should handle this explicitly, but I wanted to highlight that it could get stuck in the unavailable label due to an unhandled exception.

jeremyevans: I think we should handle those exceptions the same way. Let me update the PR for that.

fdr (approved):

PR #4680, opened 2026-01-28
SPDK Migration improvements

mohi-kalantari:
Update Vm's systemd unit after migration

The Vm's systemd unit had After and Require fields which relied on the
SPDK systemd unit and that needs to be changed to its storage systemd
unit.

Do not delete kek fifo file in SPDK migration prog

By deleting this file and calling write_kek_pipe afterwards, we are
writing to a regular file instead of a fifo file, meaning the kek would
be still readable after the vm reads the kek file and boots.

By removing the rm command, after booting the vm, the kek file still
exists but if you try to read it, the cat command will wait for stdin
on the file, meaning the contents are already consumed and there will
be no need for removals.

### DIFF
--- a/prog/storage/migrate_spdk_vm_to_ubiblk.rb
+++ b/prog/storage/migrate_spdk_vm_to_ubiblk.rb
@@ -119,7 +119,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   label def create_ubiblk_systemd_unit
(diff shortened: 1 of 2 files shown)


prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -159,6 +159,11 @@ def vm_storage_volume
     vm.vm_host.sshable.cmd("sudo systemctl start :inhost_name", inhost_name:)
     vm.strand.update(label: "wait")
 
+    hop_clean_up
+  end
+
+  label def clean_up
+    vm.vm_host.sshable.cmd("sudo rm -f :kek_file_path", kek_file_path:)
```
pykello: I suggest using `shred -u` instead to overwrite the contents of files before deletion.

PR #4681, opened 2026-01-28
Use leader version in take_backup

furkansahin:
Use postgres_timeline.leader.version directly instead of going through
the resource. This way, when there is an upgrade, we use the leader
version without waiting the representative server switch

### DIFF
--- a/prog/postgres/postgres_timeline_nexus.rb
+++ b/prog/postgres/postgres_timeline_nexus.rb
@@ -85,7 +85,7 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
     # the state to database. Since backup taking is an expensive operation,
     # we check if backup is truly needed.
     if postgres_timeline.need_backup?
-      d_command = NetSsh.command("sudo postgres/bin/take-backup :version", version: postgres_timeline.leader.resource.version)
+      d_command = NetSsh.command("sudo postgres/bin/take-backup :version", version: postgres_timeline.leader.version)
       postgres_timeline.leader.vm.sshable.cmd("common/bin/daemonizer :d_command take_postgres_backup", d_command:)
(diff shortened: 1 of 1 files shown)


enescakir (commented):
I don't have anything to contribute to this PR. I'll leave it to the PG experts

prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -77,7 +77,8 @@ def self.assemble(location_id:, parent_id: nil)
       hop_take_backup
     end
 
-    nap 20 * 60
+    # if the initial backup is not taken yet, nap for 20 seconds and retry
+    nap (latest_backup_completed_at == postgres_timeline.created_at) ? 20 : 20 * 60
```
byucesoy: We have a logic in few labels above (`wait_leader`) to wait leader to be in `wait` label at first timeline creation. Why that doesn't protect us against this case?

byucesoy: If it doesn't work, and this change protects against all cases, we can remove that one, but it would be good to understand why it didn't work in the first case. I initially added it for this case. Did we introduce regression at some point?

furkansahin: Alright, I dug into this and found out the real issue, once a pg server is up, the timeline detects the leader is ready and switches to `take_backup` state. However, the backup is taken using the `version` function in the `resource` entity. In case of an upgrade, we update the version on the standby server and the representativity happens later. So, the `take-backup` file is called with the older version. I'll switch that to use `postgres_timeline.leader.version`. (cc @shikharbhardwaj )

byucesoy: Good find :+1:

PR #4682, opened 2026-01-29
Allow booting application even if POSTGRES_MONITOR_DB unavailable

jeremyevans:
This changes how POSTGRES_MONITOR_DB is setup so the constant is always available even if the database server is not available during startup.

For this to allow booting, you cannot have models that use the database, because models need to introspect database schema when defined. So remove the PostgresLsnMonitor class, and use plain datasets to access the postgres_lsn_monitor table. This is only accessed in a couple places, so the loss of model functionality is not a significant loss.

### DIFF
--- a/db.rb
+++ b/db.rb
@@ -21,11 +21,7 @@ DB = Sequel.connect(Config.clover_database_url, max_connections:, pool_timeout: 
 
 postgres_monitor_db_ca_bundle_filename = File.join(Dir.pwd, "var", "ca_bundles", "postgres_monitor_db.crt")
 Util.safe_write_to_file(postgres_monitor_db_ca_bundle_filename, Config.postgres_monitor_database_root_certs)
-begin
-  POSTGRES_MONITOR_DB = Sequel.connect(Config.postgres_monitor_database_url, max_connections: Config.db_pool_monitor, pool_timeout: Config.database_timeout, driver_options:) if Config.postgres_monitor_database_url
-rescue Sequel::DatabaseConnectionError => ex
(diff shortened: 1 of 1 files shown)


model/postgres/postgres_server.rb
```diff
@@ -195,6 +197,10 @@ def current_lsn
     run_query(DB.select(Sequel.function(lsn_function_name)))
   end
 
+  def lsn_monitor
```
enescakir: Since it returns a dataset not the actual data.

```suggestion
  def lsn_monitor_ds
```

model/postgres/postgres_server.rb
```diff
@@ -205,8 +211,9 @@ def failover_target
     return nil if target.nil?
 
     if resource.ha_type == PostgresResource::HaType::ASYNC
-      return nil if lsn_monitor&.last_known_lsn.nil?
-      return nil if lsn_diff(lsn_monitor.last_known_lsn, target[:lsn]) > 80 * 1024 * 1024 # 80 MB or ~5 WAL files
+      last_known_lsn = lsn_monitor.get(:last_known_lsn)
+      return unless last_known_lsn
```
enescakir: nit:
```suggestion
      return unless (last_known_lsn = lsn_monitor.get(:last_known_lsn))
```

jeremyevans:
@enescakir Made both of the requested changes.

enescakir:
It looks okay from my perspective, but it might be good to get approval from the code owners who maintain this code.

jeremyevans:
E2E passed. Will be merging later today.

PR #4684, opened 2026-01-29
Destroy child strands when entering pg lockout

furkansahin:
When entering lockout, destroy any existing child strands to prevent
them from blocking the reap in wait_lockout_attempt.

The specific scenario: a restart strand was stuck as a child due to
unavailability, which blocked the reap and caused infinite wait. Since
lockout is about to stop primary, no in-flight child operations
(restart, configure, etc.) should be necessary.

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -585,7 +585,7 @@ SQL
     end
 
     reap(fallthrough: true)
-    nap 5 unless strand.children_dataset.where(prog: "Postgres::PostgresServerNexus", label: "restart").empty? && strand.children_dataset.where(prog: "Postgres::Restart").empty?
+    nap 5 unless strand.children_dataset.where(prog: "Postgres::Restart").empty?
 
     if available?
       decr_checkup
@@ -743,22 +743,6 @@ SQL
     postgres_server.destroy
 
     pop "postgres server is deleted"
-  end
-
-  label def restart
-    when_configure_set? do
-      # Pop so that the parent can handle the configure
-      pop "restart deferred due to pending configure"
-    end
-
-    decr_restart
-
-    register_deadline("wait", 5 * 60)
-
-    vm.sshable.cmd("sudo postgres/bin/restart :version", version:)
-    vm.sshable.cmd("sudo systemctl restart pgbouncer@*.service")
-    vm.sshable.cmd("sudo systemctl restart postgres-metrics.timer")
-    pop "postgres server is restarted"
   end
 
   def available?
--- a/spec/prog/postgres/postgres_server_nexus_spec.rb
+++ b/spec/prog/postgres/postgres_server_nexus_spec.rb
@@ -1346,23 +1346,6 @@ RSpec.describe Prog::Postgres::PostgresServerNexus do
     end
   end
 
-  describe "#restart" do
-    it "pops if configure is set so parent can handle it" do
-      nx.incr_configure
-      expect { nx.restart }.to exit({"msg" => "restart deferred due to pending configure"})
-      expect(nx.configure_set?).to be true
-    end
-
-    it "sets deadline, restarts and exits" do
-      expect(sshable).to receive(:_cmd).with("sudo postgres/bin/restart 16")
-      expect(sshable).to receive(:_cmd).with("sudo systemctl restart pgbouncer@*.service")
-      expect(sshable).to receive(:_cmd).with("sudo systemctl restart postgres-metrics.timer")
-      expect { nx.restart }.to exit({"msg" => "postgres server is restarted"})
-      expect(nx.strand.stack.first["deadline_target"]).to eq("wait")
-      expect(nx.strand.stack.first["deadline_at"]).to be_within(5).of(Time.now + 5 * 60)
-    end
-  end
-
   describe "#available?" do
     before do
       expect(sshable).to receive(:invalidate_cache_entry)


prog/postgres/postgres_server_nexus.rb
```diff
@@ -660,6 +660,11 @@ def before_run
   label def lockout
     decr_lockout
 
+    # Destroy child strands to avoid infinite wait in wait_lockout_attempt during
+    # reap. Since we are locking out, no child strand should be necessary. The
+    # specific scenario: a restart strand stuck due to unavailability was blocking reap.
+    strand.children.each(&:destroy)
```
furkansahin: a note to reviewer, using `Semaphore.incr(strand.children_dataset.select(:id), "destroy")` is not safe here since we push postgres server's restart label as a child and if we `incr_destroy`, we hit line 89:
```ruby
Clog.emit("Postgres server deletion is cancelled, because it is in the process of taking over the primary role")
decr_destroy
```
in `before_run`. So, `strand.children.each(&:destroy)` is the safe thing to do.

furkansahin: added that to the comment.

enescakir: It might be worth checking with @jeremyevans. As far as I know, directly destroying child strands causes a lease violation, and he fixed those cases early on.

jeremyevans: This is never, ever safe, since the strand's children could be currently running. If so, it results in a lease violation. You need to refactor things to avoid that. Simplest change is adding a separate prog for restarting, and bud that instead. Alternatively, when pushing the restart, update the frame to set a flag, and if that flag is set in the frame, have destroy operate differently. I recommend the separate prog approach.

enescakir: What do you think about just pop in `before_run` if the label is `restart` when the destroy incremented?

https://github.com/ubicloud/ubicloud/blob/c35d63b07325bb811f7f61830b0924c9b06229aa/prog/postgres/postgres_server_nexus.rb#L88-L91

```ruby
elsif @strand.label == "restart"
    pop "operation is cancelled due to explicitly requested destruction"
end
```

jeremyevans: Looks like we never `hop_restart`, so that could work. However, since we never `hop_restart`, I still think moving restart to a separate prog would be better (though more work).

enescakir: We can also check parent_id to be sure it's pushed or budded, not hopped

byucesoy: >This is never, ever safe, since the strand's children could be currently running. If so, it results in a lease violation. You need to refactor things to avoid that. Simplest change is adding a separate prog for restarting, and bud that instead. Alternatively, when pushing the restart, update the frame to set a flag, and if that flag is set in the frame, have destroy operate differently. I recommend the separate prog approach.

What is the recommended way of `kill` ing a prog? There are cases where we know we don't care if it operates normally because we are in the process of destroying the resource/server.

furkansahin: added restart prog

prog/postgres/postgres_server_nexus.rb
```diff
@@ -596,7 +596,7 @@ def before_run
       postgres_server.incr_recycle
     end
 
-    bud self.class, {}, :restart
+    bud Prog::Postgres::Restart, {"subject_id" => postgres_server.id}, :start
```
enescakir: nit: `bud` already set subject_id, and default label is `start`

https://github.com/ubicloud/ubicloud/blob/3e244c6c74a7fc27eaf20ef58a2d2c596bc8a291/prog/base.rb#L217-L218


```suggestion
    bud Prog::Postgres::Restart
```

prog/postgres/restart.rb
```diff
@@ -0,0 +1,25 @@
...
+
+class Prog::Postgres::Restart < Prog::Base
+  subject_is :postgres_server
+
+  def vm
+    @vm ||= postgres_server.vm
+  end
+
+  label def start
+    when_configure_set? do
```
enescakir: I think it checks the semaphores of the `Prog::Postgres::Restart` strand, not the postgres_server itself when budded. The ID is same when it pushed but not same when it's budded. Probably the old version was also wrong for bud case

```suggestion
    if postgres_server.configure_set?
```

prog/postgres/postgres_server_nexus.rb
enescakir: We should also push new prog. The interesting part is that tests didn't catch it. cc @jeremyevans

enescakir: It looks like we can push invalid labels.

enescakir: We should also push new prog

enescakir: We should use new prog

prog/postgres/restart.rb
```diff
@@ -0,0 +1,25 @@
...
+      # Pop so that the parent can handle the configure
+      pop "restart deferred due to pending configure"
+    end
+
+    decr_restart
```
enescakir: I think it decrease the semaphore of the `Prog::Postgres::Restart` strand, not the postgres_server itself when budded. The ID is same when it pushed but not same when it's budded. Probably the old version was also wrong for bud case

prog/postgres/postgres_server_nexus.rb
```diff
@@ -412,7 +412,7 @@ def before_run
       if retval&.dig("msg") == "postgres server is restarted"
         hop_run_post_installation_script
       end
-      push self.class, frame, "restart"
+      bud Prog::Postgres::Restart
```
jeremyevans: Is the change from push to bud here safe? Where do we reap this budded prog?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -584,7 +585,7 @@ def before_run
     end
 
     reap(fallthrough: true)
-    nap 5 unless strand.children_dataset.where(prog: "Postgres::PostgresServerNexus", label: "restart").empty?
+    nap 5 unless strand.children_dataset.where(prog: "Postgres::Restart").empty?
```
jeremyevans, line 588: During a transition period, we need to check both Postgres::PostgresServerNexus#restart and Postgres::Restart, because there may be existing child strands in Postgres::PostgresServerNexus#restart

prog/postgres/restart.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  label def start
+    register_deadline(nil, 5 * 60)
+    if postgres_server.configure_set?
+      # Pop so that the parent can handle the configure
+      pop "restart deferred due to pending configure"
+    end
+
+    vm.sshable.cmd("sudo postgres/bin/restart :version", version: postgres_server.version)
+    vm.sshable.cmd("sudo systemctl restart pgbouncer@*.service")
```
jeremyevans: I realize this is just copying code, but maybe this argument should be quoted: `'pgbouncer@*.service'`. Otherwise, the shell attempts to expand this, and if there was anything to expand, it would expand it. I think the expected behavior is to pass it as a literal argument.

spec/prog/postgres/postgres_server_nexus_spec.rb
```diff
@@ -1344,23 +1346,6 @@ def password_update_sql_matcher
     end
   end
 
-  describe "#restart" do
```
jeremyevans, line 1347: This spec removal should be moved to the following commit when the label is removed.

prog/postgres/postgres_server_nexus.rb
jeremyevans: > What is the recommended way of `kill` ing a prog? There are cases where we know we don't care if it operates normally because we are in the process of destroying the resource/server.

You use `incr_destroy` and make sure the prog handles it correctly. Sometimes that means moving labels to a separate prog, as was done here.

Even if you don't care whether the prog operates normally, it is never acceptable to introduce a lease violation.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -523,7 +523,8 @@ def before_run
     end
 
     when_restart_set? do
-      push self.class, frame, "restart"
+      decr_restart
+      push Prog::Postgres::Restart, {"subject_id" => postgres_server.id}, :start
```
enescakir: nit: You can just push. These are default
https://github.com/ubicloud/ubicloud/blob/72db69fc3ffa4bafdf98048e1b202c95d272d475/prog/base.rb#L207-L210

```suggestion
      push Prog::Postgres::Restart
```

PR #4686, opened 2026-01-29
Add tag search UI for Postgres databases

iskakaushik:
Adds tag search and display functionality for Postgres databases, behind
the postgres_tag_search_ui feature flag.

- Add tag search input with filter and clear functionality
- Display tags as clickable pills in the database list table
- Click tag pill to copy "key:value" to clipboard
- Add postgres_tags component for consistent tag rendering
- Fix z-index on notification container for proper visibility

image

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -37,14 +37,6 @@ $(".dropdown").on("click", function (event) {
   $(this).toggleClass("active");
 });
 
-$(document).on("click", ".tag-pill", function (event) {
-  event.stopPropagation();
-  let content = $(this).data("content");
-  navigator.clipboard.writeText(content).then(function() {
-    notification("Copied to clipboard");
-  });
-});
-
 $(".toggle-parent-to-active").on("click", function (event) {
   $(this).parent().toggleClass("active");
 });
--- a/views/components/postgres_tags.erb
+++ b/views/components/postgres_tags.erb
@@ -4,14 +4,7 @@
 <% else %>
   <div class="flex flex-wrap gap-1 max-w-[250px]">
     <% tags.each do |tag| %>
-      <% tag_text = "#{tag["key"]}:#{tag["value"]}" %>
-      <span
-        class="tag-pill inline-block rounded-md bg-gray-100 px-2 py-0.5 text-xs font-medium text-gray-600 hover:bg-gray-200 cursor-pointer max-w-[120px] truncate"
-        data-content="<%= h(tag_text) %>"
-        title="Click to copy: <%= h(tag_text) %>"
-      >
-        <%= h(tag_text) %>
-      </span>
+      <%= tag["key"] %> <%== part("components/copyable_content", content: tag["value"], message: "Copied tag") %>
     <% end %>
   </div>
 <% end %>


views/postgres/index.erb
```diff
@@ -24,10 +24,31 @@
...
+      </form>
+      <p class="mt-1 text-xs text-gray-500">Format: key:value (comma-separated for multiple tags)</p>
+    </div>
+  <% end %>
+
   <div class="grid gap-6">
     <%== part(
       "components/table_card",
-      headers: ["Name", "Compute", "Storage", "Version", "Parent", "Location", "State"],
+      headers: ["Name", "Compute", "Storage", "Version", "Parent", "Location", "Tags", "State"],
```
iskakaushik: need to show only if FF is enabled.

iskakaushik:

views/postgres/index.erb
```diff
@@ -24,23 +24,35 @@
...
+    <%
+      show_tags = @project.get_ff_postgres_tag_search_ui
+      headers = ["Name", "Compute", "Storage", "Version", "Parent", "Location"]
+      headers << "Tags" if show_tags
+      headers << "State"
+    %>
     <%== part(
       "components/table_card",
-      headers: ["Name", "Compute", "Storage", "Version", "Parent", "Location", "State"],
+      headers: headers,
```
furkansahin: you can just say:
```
headers:,
```
This will also fix the tests.

iskakaushik: done!

assets/js/app.js
```diff
@@ -37,6 +37,14 @@ $(".dropdown").on("click", function (event) {
   $(this).toggleClass("active");
 });
 
+$(document).on("click", ".tag-pill", function (event) {
```
byucesoy: We usually try to keep javascript footprint as low as possible. Is this functionality really required?

If the purpose is making it easier to copy/paste to the search box, my preference would be removing the search box entirely and making these clickable and clicking to them would automatically filter the resources.

assets/js/app.js
```diff
@@ -37,6 +37,14 @@ $(".dropdown").on("click", function (event) {
...
+  event.stopPropagation();
```
byucesoy: Why do we need this call?

model/project.rb
```diff
@@ -220,6 +220,7 @@ def self.feature_flag(*flags, into: self)
     :ipv6_disabled,
     :postgres_hostname_override,
     :postgres_lantern,
+    :postgres_tag_search_ui,
```
byucesoy, line 230: I think this feature is independently useful for all. We don't need to hide it behind a feature flag.

At the same time, we don't have any customer using tags because it is not very visible feature. It is only available in API/SDK, so users are not aware of it. Given that, allocating screen space for this feels wasteful.

So all in all, I'm bit conflicted :slightly_smiling_face: and keeping it behind a feature flag is OK. I just wanted share my thought process.

views/components/tag_filter.erb
```diff
@@ -0,0 +1,18 @@
+<%# locals: (action:, name: "tags", value: nil, placeholder: "Filter by tags (e.g., env:prod,team:backend)") %>
+<div class="mb-4">
```
byucesoy, line 2: I added a note to other review as well, but I think better UX would be removing the search box entirely and making tag pills clickable. Clicking to them would automatically redirects the page and filter the resources.

byucesoy: I'm not sure but `typecast_params.nonempty_str("tags")` call in `routes/postgres.rb` work if the tags are passed as URL params or do they have to be part of the body? If they accept URL params, then we don't even need a form, which would simplify the code significantly.

views/components/postgres_tags.erb
```diff
@@ -0,0 +1,15 @@
...
+    <% tags.each do |tag| %>
+      <% tag_text = "#{tag["key"]}:#{tag["value"]}" %>
+      <span class="tag-pill inline-block rounded-md bg-gray-100 px-2 py-0.5 text-xs font-medium text-gray-600 hover:bg-gray-200 cursor-pointer max-w-[120px] truncate"
+            data-content="<%= h(tag_text) %>"
+            title="Click to copy: <%= h(tag_text) %>">
+        <%= h(tag_text) %>
+      </span>
+    <% end %>
+  </div>
+<% end %>
```
byucesoy: I thought our erb_formatter would automatically add a new line to the end of the files, but maybe I'm mistaken. Can you please run `bundle exec rake linter:erb_formatter` just to be sure?

Also for a good measure, `bundle exec rake linter:rubocop`.

erb_formatter is for `.erb` file rubocop is for `.rb` files.

byucesoy (commented):
I think this looks good overall. Left few comments.

model/project.rb
serprex: would it be better to remove this feature flag & expose editing tags in UI?

guess future PR exposing UI could remove FF

assets/js/app.js
serprex: the purpose isn't copy pasting to search box. we put UUIDs in tags, we'd often like to copy/paste UUID to systems where that maps to some other entity. Unfortunately double clicking on UUIDs isn't effective for selecting the entire UUID

enescakir: We already have similar functionality with the `.copyable-content` class. Can’t we reuse the `copyable`-content class in the `tag-pill` components?

```javascript
$(".copyable-content").on("click", ".copy-button", function (event) {
  let parent = $(this).parent();
  let content = parent.data("content");
  let message = parent.data("message");
  navigator.clipboard.writeText(content);

  if (message) {
    notification(message);
  }
})
```

views/components/tag_filter.erb
serprex: The idea is you want to look at resources for a given org id, having to click tags when there's 100 postgres resources to sift through isn't ideal. I think this search is doing it wrong tho, right now validation fails if it isn't in form `key:value` but I just want to paste `value` & have it find any resource with tag holding that value

furkansahin:
Can you rebase this and make sure the CI passes. I'll review it after that.

serprex:
this PR should be moved to draft, when discussed various parts were suggested to be split up, & it's long term value to us is questionable since we'd like to migrate to ABAC tags

serprex:
closing as we're relying on internal tooling here

PR #4689, opened 2026-01-30
PG restart improvements

shikharbhardwaj:


prog/postgres/postgres_server_nexus.rb
```diff
@@ -774,7 +780,8 @@ def available?
 
     # Do not declare unavailability if Postgres is in crash recovery
     begin
-      return true if vm.sshable.cmd("sudo tail -n 5 /dat/:version/data/pg_log/postgresql.log", version:).include?("redo in progress")
+      log_output = vm.sshable.cmd("sudo tail -n 10 /dat/:version/data/pg_log/postgresql.log", version:)
+      return true if log_output.include?("redo in progress") || log_output.include?("Consistent recovery state has not been yet reached")
```
enescakir: Not sure if it makes it more performant, but you can also grep on the dataplane side.

```suggestion
      log_output = vm.sshable.cmd("sudo tail -n 10 /dat/:version/data/pg_log/postgresql.log | grep -e 'redo in progress' -e 'Consistent recovery state has not been yet reached'", version:)
      return true if log_output
```

byucesoy: :+1: on making the grep on dataplane. We can even increase `-n 10` to a bigger value when we do the grep on the data plane.

While doing this, I think it would be good to cover an edge case (that exists even before this change). What if PG process stops and we have `redo in progress` in last few log lines. We wouldn't detect this as unavailability. We should also check that the log lines are recent. There are various ways of doing it (log timestamp, last modified time etc.), I'm not sure which one would be most resilient.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -773,7 +773,15 @@ def available?
 
     # Do not declare unavailability if Postgres is in crash recovery
     begin
-      return true if vm.sshable.cmd("sudo tail -n 5 /dat/:version/data/pg_log/postgresql.log", version:).include?("redo in progress")
+      log_output = vm.sshable.cmd(<<~CMD, version:)
+  sudo awk -v cutoff="$(date -u -d '5 minutes ago' '+%Y-%m-%d %H:%M')" '
```
enescakir: Claude code comment:

```shell
The current version is correct and readable. For a potentially large log file though, it reads the entire file every time. Here are some alternatives:

  Most performant (reads from end, exits early):

  sudo tac /dat/:version/data/pg_log/postgresql.log | awk -v cutoff="$(date -u -d '5 minutes ago' '+%Y-%m-%d %H:%M')" '
    /^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}/ {
      if (substr($0, 1, 16) < cutoff) exit
      if (/redo in progress/ || /Consistent recovery state has not been yet reached/) print
    }
  '
  Uses tac to read backwards and exits as soon as it hits a line older than 5 minutes. Only reads what's needed.
```

prog/postgres/postgres_server_nexus.rb
```diff
@@ -773,7 +773,16 @@ def available?
 
     # Do not declare unavailability if Postgres is in crash recovery
     begin
-      return true if vm.sshable.cmd("sudo tail -n 5 /dat/:version/data/pg_log/postgresql.log", version:).include?("redo in progress")
+      log_output = vm.sshable.cmd(<<~CMD, version:)
+  sudo tac /dat/:version/data/pg_log/postgresql.log | awk -v cutoff="$(date -u -d '5 minutes ago' '+%Y-%m-%d %H:%M')" '
```
byucesoy: Thinking about this more... users can change the log line prefix, which in turn would break this parsing. I think we need another way of understanding that the log lines are recent enough.

shikharbhardwaj: I switched this to use last 50 lines if the file was modified within the last 5 minutes. It could delay failover by 5 minutes in case PG died during a crash recovery and could not come back, but I guess that is okay in some sense to give it some chance to recover.

It might also be helpful to have some way to distinguish between recovery crash looping vs a single long running recovery, as we'd want to delcare unavailability in the crash loop.

PR #4690, opened 2026-01-30
Skip PG upgrade E2E tests

shikharbhardwaj:
Skip PG upgrade E2E tests for the scheduled runs, given that this test has been intermittently failing.


.github/workflows/e2e.yml
enescakir, line 75: It uses all cases from `config/e2e_test_cases.yml` file for schedule.

You might ignore `postgres_upgrade` with grep

shikharbhardwaj: Added a filter in the `yq` command

PR #4691, opened 2026-01-30
sketch: determining expected vs unexpected vs guest-initiated shutdown

fdr:
This is not intended to be merged, but I know of no better way to collate comments on something like this than a PR.

The general idea here is to tell apart when a guest vm exits on its own accord, versus abnormal situations worthy of paging us.

Included are two files: a shell-script-like test file, `test_vm_exit_reason.rb`, that models systemd status outputs and so on, which is used as a context-primer to build the implementation sketch file, `vm_exit_reason.rb`. the latter was not tested in integration, but it's not quite a one-shot either, I did several rounds of correcting design (for example: invocation_id was not something claude found, I thought out the unnecessary pathology of omitting it).

I also had it do significant source diving and offered several meaningful corrections in deriving the `vm_exit_reason.rb` sketch you see within. So it was a bit of an effort, but hopefully better than "yeah I guess we should grep some strings huh?"

### DIFF
--- a/vm_exit_reason.rb
+++ b/vm_exit_reason.rb
@@ -1,0 +1,162 @@
+# frozen_string_literal: true
+
+# Sketch: VM exit reason detection for Metal::Nexus
+#
+# This module provides vm_exit_reason and a replacement unavailable
+# label that uses it. Intended to be integrated into
+# Prog::Vm::Metal::Nexus.
+#
+# ## Experimental results (2026-01-30)
+#
+# Tested on Ubuntu 24.04 with cloud-hypervisor v50, using a
+# persistent systemd unit with ExecStop=ch-remote shutdown-vmm
+# (matching production). Key finding: when the CH main process
+# exits on its own, systemd runs ExecStop, which fails because
+# the API socket is already gone. This makes Result="exit-code"
+# the common case for all self-exits, not Result="success".
+#
+#   | Cause              | systemd Result | Journal grep match         |
+#   |--------------------|----------------|----------------------------|
+#   | Guest ACPI shutdown| exit-code      | "ACPI Shutdown signalled"  |
+#   | SIGTERM (direct)   | exit-code      | (no match)                 |
+#   | SIGKILL / OOM      | signal         | (no CH log output)         |
+#   | systemctl stop     | success        | (ExecStop ran cleanly)     |
+#
+# The original sketch assumed Result="success" was the ambiguous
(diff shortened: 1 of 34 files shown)


fdr:
this is [the crux](https://github.com/ubicloud/ubicloud/pull/4691/files#diff-5105dcceb9ec7fd1676d1fb7befb1d75363e78d2a3b63f44294306cf429f4351R76-R106) of all that bloviation though, though the blovation may be useful if you need it.

vm_exit_reason.rb
```diff
@@ -0,0 +1,162 @@
...
+    reason = begin
+      host.sshable.cmd(
+        "journalctl _SYSTEMD_INVOCATION_ID=:invocation_id " \
+        "-o cat -n 50 --no-pager " \
+        "| grep -m1 -oF " \
+        "-e 'ACPI Shutdown signalled' " \
+        "-e 'vCPU thread panicked' " \
+        "-e 'VCPU generated error' " \
+        "-e 'thread panicked'",
+        vm_name:, invocation_id:
```
jeremyevans, line 98: `:vm_name` placeholder is not used in this command:

```suggestion
        invocation_id:
```

jeremyevans (commented):
This looks good. I'm fine updating #4677 to use the approach instead of turning off paging completely.

fdr:
> This looks good. I'm fine updating #4677 to use the approach instead of turning off paging completely.

Hope it helped with some of the annoying constant strings or such. with a slightly more advanced workbench, I could have made the sketch do validation of a live form as well against a bare metal host. Maybe in the next month or two.

vm_exit_reason.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    # Forcible kill — no CH logs to check, always page-worthy.
+    return {result:, reason: result} if FORCIBLE_RESULTS.include?(result)
+
+    # "exit-code" (or any other unexpected value): CH exited on
+    # its own. Check journal to find out why.
+    reason = begin
+      host.sshable.cmd(
+        "journalctl _SYSTEMD_INVOCATION_ID=:invocation_id " \
+        "-o cat -n 50 --no-pager " \
```
fdr, line 92: something about this doesn't sit right with me...the tail with the limit, even the cat and no pager seems weird in a no-tty environment

fdr: (but it might be okay, i guess, just it's a bit of extra fluff that's not immediately self-justifying in my mind)

PR #4696, opened 2026-01-30
Add web UI for adding AAAA record for older Postgres databases

jeremyevans:
Written by Claude and cleaned up and fixed by me.

See individual commits for details.

### DIFF
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -132,7 +132,7 @@ class Clover < Roda
     update_invitation
     upgrade
   ACTIONS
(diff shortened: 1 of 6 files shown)


fdr (approved):
I'm not the right person to read through markup to see the forest for the trees. I do think we maybe lacking a "how do I test this route" methodology (which can be generating a `psql` or `nc` or "something" against the ipv6 address that checks that it can even get the hello banner

byucesoy (approved):

PR #4698, opened 2026-02-01
Retry azs again after 5 failures

serprex:
Avoids getting locked out after temporary zone availability outage

### DIFF
--- a/prog/vm/aws/nexus.rb
+++ b/prog/vm/aws/nexus.rb
@@ -421,7 +421,9 @@ class Prog::Vm::Aws::Nexus < Prog::Base
 
   def retry_in_different_az(e)
     if (retry_count = frame["retry_count"] || 0) >= 5
-      raise e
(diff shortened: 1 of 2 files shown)


serprex:
aside: I think this is the wrong way about things, but at least it's minimal

what really needs to happen is we go with preference based az selection:
1. sort based on same az / different az standby setting (putting primary in customer preferred AZ)
2. downsort AZs which are currently being excluded, moving the latest failed AZ to the lowest order

ideally with some kind of cache tracking (account, az, instance type, failure) which decays

byucesoy:
I'm supportive of adding preferred az to the endpoints and passing it down through VM provisioning. Though, it requires us to store AZ lists for existing regions. Currently it is hard-coded to `a`, `b` or `c`, which is wrong.

iamKunalGupta:
I think issues like this will also shape the capabilities endpoint. There could be some yaml that tells the full picture of instances: supported regions, azs configuration which can be used by capabilties, az retries etc.

But till we have that, the retry is probably better than nothing

PR #4699, opened 2026-02-01
Add ConnectionCheckoutTelemetry

jeremyevans:
This uses the new Sequel connection_checkout_event_callback extension
to allow gathering of telemetry information about connections. It
provides information such as the percentage of checkout requests that
were handled immediately, as well histogram-like information breaking
down the amount of wait times.
    
Once we can start monitoring this information, we can make educated
decisions regarding connection pool sizes per process type.

To simplify implementation, set a PROCESS_TYPE environment variable in
all processes.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -4,8 +4,18 @@ ENV["PROCESS_TYPE"] = "web"
 
 require_relative "loader"
 
-CONNECTION_CHECKOUT_TELEMETRY = ConnectionCheckoutTelemetry.new
-CONNECTION_CHECKOUT_TELEMETRY.setup
+# We start the thread to monitor checkout events in puma's before_worker_boot
+# hook, but that is not called if puma is not running in forking mode. While
+# there are other ways besides WEB_CONCURRENCY to turn on forking mode, using
+# WEB_CONCURRENCY is how we enable forking in production and staging. Only
+# setup connection checkout telemetry if WEB_CONCURRENCY is used, to avoid a
+# memory leak in development environments that do not set WEB_CONCURRENCY.
+if ENV["WEB_CONCURRENCY"]
+  CONNECTION_CHECKOUT_TELEMETRY = ConnectionCheckoutTelemetry.new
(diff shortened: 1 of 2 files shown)


enescakir:
Something's wrong with the E2E tests

CleanShot 2026-02-01 at 23 29 34@2x

jeremyevans:
Certainly it's wrong for the E2E tests to report success when they failed. Failure is:

```
20:04:05 monitor.1   | /home/runner/work/ubicloud/ubicloud/vendor/bundle/ruby/3.4.0/bundler/gems/sequel-4bce3850a2b0/lib/sequel/database/misc.rb:260:in 'block (2 levels) in Sequel::Database#extension': can't modify frozen Array: [:pg_array, :pg_json, :pg_auto_parameterize, :pg_auto_parameterize_in_array, :pg_timestamptz, :pg_range, :pg_enum, :pg_schema_caching, :index_caching, :date_arithmetic]
```

When testing locally, I didn't test in production mode. Moving:

```ruby
checkout_telemetry = ConnectionCheckoutTelemetry.new
checkout_telemetry.setup_and_run_thread
```

before `clover_freeze` fixes the issue locally. I've pushed a commit for that change and will rerun the E2E tests after CI passes.

jeremyevans:
E2E passed. `monitor_connection_checkout_metrics` appears in the logs, with reasonable looking results (% immediately available from 97.75%-99.50%, pool size of 4). `respirate_connection_checkout_metrics` does not show up, which matches what I've seen locally, since the respirate scan thread and each respirate worker thread checkout a connection from the pool on startup and use it the entire time.

fdr (approved):

jeremyevans:
I tested this in staging and it was not emitting web connection checkout metrics. I had to set `WEB_CONCURRENCY` in staging just as we set it in production, then the web connection checkout metrics showed up. I pushed another commit to only setup the metrics for the web process if `WEB_CONCURRENCY` is set, which avoids a memory leak in development environments that do not set it.

PR #4700, opened 2026-02-02
Run metal and AWS E2E tests in parallel

enescakir:
As we introduce new tests, the overall E2E runtime keeps increasing. To mitigate this, we can run the metal and aws E2E tests in parallel.

The biggest challenge here is runner tests. Since they share the same GitHub App, it’s not possible to run them in parallel within the same repository.

To solve this, I created two separate GitHub Apps (e2e-test-app-metal, e2e-test-app-aws) and two repositories (tahcloud/github-e2e-tests-metal, tahcloud/github-e2e-tests-aws) so the tests can run in parallel.

I added configurations for these into two separate environments in this repository (E2E-metal, E2E-aws). Each job uses its own environment to access the correct secrets.

The changes in the runner test prog are a bit ugly. I have another branch to simplify it.

Previously, when we ran the tests serially, if the metal tests failed, the AWS tests wouldn’t even start. Now, even if the metal tests are broken, we can still see the AWS results.

Since most of the time is spent on the metal tests, they take ~56m, while the AWS tests take ~8m. This doesn’t significantly affect the overall runtime, but it gives us ~48m of free time to add more tests for the AWS provider.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -22,12 +22,33 @@ on:
     - cron:  '0 */2 * * *'
 
 jobs:
-  run-ci:
+  setup:
     if: vars.RUN_E2E == 'true'
+    runs-on: ubuntu-slim
+    outputs:
+      providers: ${{ steps.set-providers.outputs.providers }}
+    steps:
+      - name: Set provider matrix
+        id: set-providers
+        run: |
+          providers="metal,aws"
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" && -n "${{ inputs.providers }}" ]]; then
+            providers="${{ inputs.providers }}"
+          fi
+          # Convert comma-separated list to JSON array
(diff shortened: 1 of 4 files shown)


jeremyevans (approved):
My knowledge of GitHub Actions is fairly shallow, but I'm in favor of the general idea. This looks fine to merge once metal E2E passes with it.

enescakir:
> This looks fine to merge once metal E2E passes with it.

Since I changed the `concurrency` key, it no longer matches the current default. As a result, when I run it, another trigger on the main branch can reset the VM host because it isn't blocked by the `e2e-metal` concurrency key.

jeremyevans:
> > This looks fine to merge once metal E2E passes with it.
> 
> Since I changed the `concurrency` key, it no longer matches the current default. As a result, when I run it, another trigger on the main branch can reset the VM host because it isn't blocked by the `e2e-metal` concurrency key.

So basically, failures are expected on the PR, but would not be expected once merged to the main branch? If you have confidence this will work, I guess it is fine to merge.

PR #4701, opened 2026-02-02
Fix invoice ID lookup

enescakir:
We are passing the `:id` symbol instead of the `id` variable. Since project invoices are ordered by creation date, this caused the query to always return the most recent invoice, regardless of the ID provided.

As a result, customers saw the same invoice when clicking on any invoice in the console.


enescakir:
@jeremyevans it's interesting that when we pass a column symbol to `with_pk` instead of the pk value, it just returns the first row.

PR #4702, opened 2026-02-02
Add cluster level monitoring to UbiCSI node-plugin daemonset

mohi-kalantari:
**Add Mesh Connectivity Monitoring to Kubernetes CSI**
This PR implements mesh network connectivity monitoring for Kubernetes clusters, enabling detection of network isolation between pods and external endpoints.

**Changes**
**Mesh Connectivity Checker**

- Add MeshConnectivityChecker that runs as a background thread in the CSI node plugin
- Periodically verify TCP connectivity to other pods in all other nodes on certain intervals
- Support external endpoint monitoring via EXTERNAL_ENDPOINTS environment variable
- Write connectivity status to /var/lib/ubicsi/mesh_status.json using atomic file writes

**KubernetesNode Health Monitoring**

- Integrate KubernetesNode into the central health monitor system
- Read mesh status file via SSH and report node as down if any peer is unreachable
- Include external endpoint reachability in health checks

**Infrastructure**

- Remove hostNetwork: true from CSI DaemonSet to validate actual CNI network paths
- Add network debugging tools to CSI Docker image (netcat, mtr, traceroute, etc.)
**Refactor** 
- yaml_load_kubectl to automatically append -oyaml

### DIFF
--- a/kubernetes/csi/Dockerfile
+++ b/kubernetes/csi/Dockerfile
@@ -13,7 +13,8 @@ RUN apt update && apt install -y apt-transport-https ca-certificates curl gnupg 
 RUN curl -fsSL https://pkgs.k8s.io/core:/stable:/v1.34/deb/Release.key | gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
 RUN echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.34/deb/ /' | tee /etc/apt/sources.list.d/kubernetes.list && \
     chmod 644 /etc/apt/sources.list.d/kubernetes.list
-RUN  apt update && apt install -y kubectl
+RUN apt update && apt install -y kubectl iputils-ping dnsutils curl telnet \
+    net-tools iproute2 netcat-openbsd mtr-tiny traceroute
 
 COPY --from=builder /usr/local/bundle /usr/local/bundle
 
--- a/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
@@ -22,7 +22,7 @@ spec:
       priorityClassName: system-node-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:0.3.0
+          image: ubicloud/ubicsi:v0.4.0
           imagePullPolicy: IfNotPresent
           env:
             - name: NODE_ID
--- a/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
@@ -34,7 +34,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:0.3.0
+          image: ubicloud/ubicsi:v0.4.0
           imagePullPolicy: IfNotPresent
           env:
             - name: NODE_ID


mohi-kalantari:
Before merging to main, I'll build the official docker image without rc prefix and the update the PR

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -104,5 +104,17 @@ def find_pv_by_volume_id(volume_id)
 
       pv
     end
+
+    def get_worker_nodes
+      nodes_yaml = yaml_load_kubectl("get", "nodes", "-oyaml")
```
aamederen: nitpick: Maybe `yaml_load_kubectl` function adds the `-oyaml` part to each command

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -104,5 +104,17 @@ def find_pv_by_volume_id(volume_id)
...
+      nodes_yaml["items"].reject do |node|
```
aamederen: instead of doing a two-pass (reject + filter_map), you can do just `filter_map` and call `next if name&.start_with?("kc")` inside.

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -104,5 +104,17 @@ def find_pv_by_volume_id(volume_id)
...
+        node.dig("metadata", "name")&.start_with?("kc")
+      end.filter_map do |node|
+        {
+          "name" => node.dig("metadata", "name"),
+          "ip" => node.dig("status", "addresses")&.find { |addr| addr["type"] == "InternalIP" }&.dig("address")
```
aamederen: instead of `{|addr| addr["type"] == "InternalIP"}`, maybe just `{ it["type"] == "InternalIP" }`

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+      threads.each(&:join)
+    end
+
+    def check_node_connectivity(node)
+      ip = node["ip"]
+      name = node["name"]
+      return unless ip
+
+      socket = Socket.tcp(ip, TCP_PORT, connect_timeout: CONNECTION_TIMEOUT)
```
aamederen: As far as I understand, we use Port 22 and internal IP of the other node for connectivity check, is that right?

If so, just asking as an idea: What if we use pod-pod connectivity to check the mesh health? Could this help catching any network issues that impacts pod-pod connection that would not be caught otherwise with pod-node tests?

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -67,11 +80,31 @@ def check_node_connectivity(node)
 
       socket = Socket.tcp(ip, TCP_PORT, connect_timeout: CONNECTION_TIMEOUT)
       socket.close
+      update_node_status(name, ip, true)
```
aamederen: instead of repeating the function call in each block, I think we can set a "success" variable and log its result at the end. Even better, we can store the details of the exception in the status and write it for better telemetry.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -67,11 +80,31 @@ def check_node_connectivity(node)
...
+        @node_status[name] = {
+          ip:,
+          reachable:,
+          last_check: Time.now.utc.iso8601
+        }
+      end
+    end
+
+    def write_status_file
+      File.write(STATUS_FILE_PATH, JSON.pretty_generate(status_response))
```
aamederen: we can use atomic write here (write to `file.tmp` and then `mv file.tmp file`)

kubernetes/csi/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt update && apt install -y apt-transport-https ca-certificates curl gnupg
 RUN curl -fsSL https://pkgs.k8s.io/core:/stable:/v1.34/deb/Release.key | gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
 RUN echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.34/deb/ /' | tee /etc/apt/sources.list.d/kubernetes.list && \
     chmod 644 /etc/apt/sources.list.d/kubernetes.list
-RUN  apt update && apt install -y kubectl
+RUN  apt update && apt install -y kubectl iputils-ping dnsutils curl telnet
```
aamederen: it would be good to double check these packages cover the other possible tools we may need: `nc, ifconfig, netstat, ip, mtr`, maybe even `traceroute`

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -17,9 +17,31 @@ def initialize(logger:, node_id:)
...
       @mutex = Mutex.new
     end
 
+    def parse_external_endpoints(env_value)
+      return [] if env_value.nil? || env_value.strip.empty?
+      env_value.split(",").filter_map do |endpoint|
+        endpoint = endpoint.strip
+        next if endpoint.empty?
+        if endpoint.include?(":")
+          parts = endpoint.rpartition(":")
```
aamederen: nice touch, it took me a bit to understand why `rpartition` is used instead of `split` but it would preserve IPv6 addresses so that's good, maybe even worth leaving a comment on how this function would behave with different inputs (`hostname`, `hostname:port`, `IPv6`, `[IPv6]`, `IPv6:port`, etc.)

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -17,9 +17,31 @@ def initialize(logger:, node_id:)
       @node_id = node_id
       @nodes = []
       @node_status = {}
+      @external_endpoints = parse_external_endpoints(ENV["EXTERNAL_ENDPOINTS"])
```
aamederen, line 22: how do we supply this environment variable and do we need to restart all csi pods after external endpoints change in this case?

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -49,6 +72,7 @@ def spawn_connectivity_check_thread
         sleep(5) # Waiting for the initial nodes lookup
         loop do
           check_all_nodes_connectivity
+          check_all_external_endpoints
```
aamederen, line 72: should we do external endpoint check in parallel with mesh connectivity check?

model/kubernetes/kubernetes_node.rb
```diff
@@ -10,11 +10,50 @@ class KubernetesNode < Sequel::Model
...
+    rescue JSON::ParserError => e
+      Clog.emit("Failed to parse mesh status file", {kubernetes_node_mesh: {ubid:, error: e.message}})
+      "down"
+    rescue IOError, Errno::ECONNRESET
+      raise
+    rescue => e
+      Clog.emit("Exception in KubernetesNode pulse check", Util.exception_to_hash(e, into: {ubid:}))
+      "down"
+    end
+    aggregate_readings(previous_pulse:, reading:)
```
aamederen, line 53: as a consequent PR maybe, we should implement overall node health check,  `checkup` procedure and page creation logic for kubernetes node

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
mohi-kalantari: It shouldn't cause any issues. If it causes any, we can take a look.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
mohi-kalantari: Yes exactly. We should add an env var to the node-plugin DaemonSet. by default it's empty

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
mohi-kalantari: For the first question: yes that's right. 

Since we last talked, do you think kube proxy healthz endpoint is a good candidate?

aamederen: I don't have a strong opinion, but yeah kube proxy healthz sounds good.

kubernetes/csi/spec/csi/v1/node_service_spec.rb
```diff
@@ -5,12 +5,13 @@
...
-  let(:client) { Csi::KubernetesClient.new(logger: Logger.new($stdout), req_id:) }
-  let(:service) { described_class.new(logger: Logger.new($stdout), node_id: "test-node") }
+  let(:logger) { Logger.new(File::NULL) }
+  let(:client) { Csi::KubernetesClient.new(logger:, req_id:) }
+  let(:service) { described_class.new(logger:, node_id: "test-node") }
 
   before do
     allow(SecureRandom).to receive(:uuid).and_return("test-req-id")
-    allow(service).to receive(:log_with_id) # suppress logs during test runs
+    allow(service).to receive(:log_with_id)
```
jeremyevans: With a null logger, you shouldn't need to mock this method to suppress output.

```suggestion
```

kubernetes/csi/spec/ubi_csi/kubernetes_client_spec.rb
```diff
@@ -4,10 +4,11 @@
 require "spec_helper"
 
 RSpec.describe Csi::KubernetesClient do
-  let(:client) { described_class.new(req_id: "test-req-id", logger: Logger.new($stdout)) }
+  let(:logger) { Logger.new(File::NULL) }
+  let(:client) { described_class.new(req_id: "test-req-id", logger:) }
 
   before do
-    allow(client).to receive(:log_with_id) # suppress logs during test runs
+    allow(client).to receive(:log_with_id)
```
jeremyevans:
```suggestion
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+          node_id: @node_id,
+          pods: @pod_status.dup,
+          external_endpoints: @external_status.dup
+        }
+      end
+    end
+
+    def spawn_connectivity_check_thread
+      Thread.new do
+        loop do
```
jeremyevans: In general, spawned threads should be able to shutdown normally. For something like this, I recommend a `@shutdown` instance variable. When the program is shutting down, set the instance variable, then join all threads.

```suggestion
        until @shutdown
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+          check_all_pods_connectivity
+          check_all_external_endpoints
+          write_status_file
+          sleep(CONNECTIVITY_CHECK_INTERVAL)
```
jeremyevans: To avoid blocking at shutdown, I recommend using a queue. On shutdown, close the queue, so the method call returns immediately instead of waiting for the sleep time.

```suggestion
          @queue.pop(timeout: CONNECTIVITY_CHECK_INTERVAL)
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+    STATUS_FILE_PATH = "/var/lib/ubicsi/mesh_status.json"
+
+    def initialize(logger:, node_id:)
+      @logger = logger
+      @node_id = node_id
+      @pod_status = {}
+      @external_endpoints = parse_external_endpoints(ENV["EXTERNAL_ENDPOINTS"])
+      @external_status = {}
+      @mutex = Mutex.new
+    end
```
jeremyevans, line 25: For later suggestions:

```suggestion
      @queue = Queue.new
      @shutdown = false
    end
    
    def shutdown!
      @shutdown = true
      @queue.close
      @thread&.join
    end
```

Update the specs to call `shutdown!` as appropriate to avoid leaking threads. You can copy the leaked thread detector from the main specs to the csi specs: https://github.com/ubicloud/ubicloud/blob/b69e28281e47b83347c3b2fdfe2be4824a33a6af/spec/spec_helper.rb#L57-L75

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def check_all_pods_connectivity
+      client = KubernetesClient.new(req_id: "mesh-check", logger: @logger)
+      pods = client.get_nodeplugin_pods
+      @logger.debug("[MeshConnectivity] Found #{pods.size} nodeplugin pods")
+
+      threads = pods.filter_map do |pod|
+        next if pod["node"] == @node_id
+        Thread.new { check_pod_connectivity(pod) }
```
jeremyevans: Instead of spawning a thread per pod, you may want to consider using `connect_nonblock` and `IO.select`. See example: https://docs.ruby-lang.org/en/3.4/Socket.html#method-i-connect_nonblock

This isn't required, so if you think it's not worth the complexity increase, you can disregard it.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+        end
+        {host:, port:}
+      rescue ArgumentError
+        @logger.warn("[MeshConnectivity] Invalid port in endpoint: #{endpoint}")
+        nil
+      end
+    end
+
+    def start
+      spawn_connectivity_check_thread
```
jeremyevans:
```suggestion
      @thread = spawn_connectivity_check_thread
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,59 @@
...
+      client = KubernetesClient.new(req_id: "mesh-check", logger: @logger)
+      pods = client.get_nodeplugin_pods
+      @logger.debug("[MeshConnectivity] Found #{pods.size} nodeplugin pods")
+
+      threads = pods.filter_map do |pod|
+        next if pod["node"] == @node_id
+        Thread.new { check_pod_connectivity(pod) }
+      end
+      threads.each(&:join)
+    rescue => e
```
jeremyevans, line 175: This is wrapping more code than it probably should. `check_pod_connectivity` already uses a similar approach. Is this to only protect against failures in `client.get_nodeplugin_pods`? If so, only wrap that part in a begin/rescue.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+      @logger.error("[MeshConnectivity] Failed to check pods connectivity: #{e.message}")
+    end
+
+    def check_pod_connectivity(pod)
+      ip = pod["ip"]
+      name = pod["name"]
+      return unless ip
```
jeremyevans: No point in setting `name` if you are going to return early due to lack of `ip`:

```suggestion
      return unless (ip = pod["ip"])
      name = pod["name"]
```

Alternatively, do the `ip` filtering inside the `filter_map` block in `check_all_pods_connectivity` before calling this method.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -360,6 +362,13 @@ def node_unpublish_volume(req, _call)
           log_and_raise(req_id, e)
         end
       end
+
+      private
+
+      def start_mesh_connectivity_checker
+        @mesh_checker = Csi::MeshConnectivityChecker.new(logger: @logger, node_id: @node_id)
+        @mesh_checker.start
```
jeremyevans, line 368: This class should also have a `shutdown!` method that calls `@mesh_checker.shutdown!`.

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,335 @@
+# frozen_string_literal: true
+
+require "logger"
+require "spec_helper"
+
+RSpec.describe Csi::MeshConnectivityChecker do
+  let(:logger) { instance_double(Logger) }
```
jeremyevans: Best to avoid unnecessary mocking:

```suggestion
  let(:logger) { Logger.new(File::NULL) }
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+  let(:node_id) { "worker-1" }
+  let(:checker) { described_class.new(logger:, node_id:) }
+
+  before do
+    allow(logger).to receive(:debug)
+    allow(logger).to receive(:info)
+    allow(logger).to receive(:warn)
+    allow(logger).to receive(:error)
+  end
+
```
jeremyevans: Best to avoid unnecessary mocking.

```suggestion
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+  describe "#initialize" do
+    it "initializes correctly with logger and node_id" do
+      expect(checker.instance_variable_get(:@logger)).to eq(logger)
+      expect(checker.instance_variable_get(:@node_id)).to eq(node_id)
+      expect(checker.instance_variable_get(:@pod_status)).to eq({})
+      expect(checker.instance_variable_get(:@external_endpoints)).to eq([])
+      expect(checker.instance_variable_get(:@external_status)).to eq({})
+    end
+
```
jeremyevans: There doesn't seem to be a good reason to test internal state here.

```suggestion
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+      checker_with_endpoints = described_class.new(logger:, node_id:)
+      endpoints = checker_with_endpoints.instance_variable_get(:@external_endpoints)
+      expect(endpoints).to eq([{host: "10.0.0.1", port: 443}, {host: "api.example.com", port: 8080}])
+    end
+  end
+
+  describe "#start" do
+    it "spawns thread and logs startup" do
+      expect(checker).to receive(:spawn_connectivity_check_thread)
+      expect(logger).to receive(:info).with("[MeshConnectivity] Started mesh connectivity checker for node worker-1")
```
jeremyevans: For all logger calls you are mocking, call the original method for more complete testing. Example:

```suggestion
      expect(logger).to receive(:info).with("[MeshConnectivity] Started mesh connectivity checker for node worker-1").and_call_original
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+
+      call_count = 0
+      expect(checker).to receive(:sleep) do |_seconds|
+        call_count += 1
+        throw :break_loop if call_count >= 1
+      end
+
+      expect(checker).to receive(:loop) do |&block|
+        catch(:break_loop) { loop { block.call } }
+      end
```
jeremyevans: The `shutdown!` approach recommended should make this code unnecessary or at least simpler.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    def update_pod_status(name, ip, reachable, error: nil)
+      @mutex.synchronize do
+        @pod_status[name] = {
+          ip:,
+          reachable:,
+          error:,
+          last_check: Time.now.utc.iso8601
+        }
+      end
```
jeremyevans, line 188: Best practice is to move as much logic as possible outside of the mutex while still keeping thread safety:

```suggestion
        hash = {
          ip:,
          reachable:,
          error:,
          last_check: Time.now.utc.iso8601
        }
        @mutex.synchronize do
          @pod_status[name] = hash
        end
      end
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -47,13 +63,55 @@ def check_pod_connectivity(pod)
...
+        # Use chmod 0644 instead of chown to allow the ubi user on the host to
+        # read the mesh status file. The CSI container runs with its own user
+        # namespace and lacks the ubi user in its /etc/passwd, causing
+        # FileUtils.chown("ubi", "ubi", ...) to fail with "can't find user for
+        # ubi". Since the file only contains non-sensitive connectivity status
+        # data, making it world-readable is a simpler solution than hardcoding
+        # UID/GID values or adding ubi to the container image.
+        FileUtils.chmod(0o644, STATUS_FILE_PATH)
+      end
+      @logger.debug("[MeshConnectivity] Wrote status to #{STATUS_FILE_PATH}")
```
jeremyevans: In general, code like this should in an `else` block after rescue. After all, if `@logger.debug` raises an exception, calling `@logger.error` is unlikely to be what you want.

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -92,49 +107,90 @@
...
+      })
+
+      temp_file = instance_double(Tempfile, path: "/var/lib/ubicsi/mesh_status123")
+      expect(Tempfile).to receive(:create).with("mesh_status", "/var/lib/ubicsi").and_yield(temp_file)
+      expect(temp_file).to receive(:write).with(satisfy { |content| JSON.parse(content)["node_id"] == "worker-1" })
+      expect(temp_file).to receive(:close)
+      expect(File).to receive(:rename).with("/var/lib/ubicsi/mesh_status123", "/var/lib/ubicsi/mesh_status.json")
+      expect(FileUtils).to receive(:chmod).with(0o644, "/var/lib/ubicsi/mesh_status.json")
+      expect(logger).to receive(:debug).with("[MeshConnectivity] Wrote status to /var/lib/ubicsi/mesh_status.json")
+
```
jeremyevans: Replace all of this mocking code. Use a custom directory, call the method, then read the file to ensure it matches what you expect.

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -92,49 +107,90 @@
...
+      checker.write_status_file
+    end
 
-      expect(Socket).not_to receive(:tcp)
+    it "logs error when file write fails" do
+      expect(Tempfile).to receive(:create).and_raise(Errno::EACCES.new("Permission denied"))
```
jeremyevans: Don't mock, use a non-existent directory name and check that an error is logged (maybe not the exact error message, only the prefix, since the underlying message may be system-dependent).

model/kubernetes/kubernetes_node.rb
```diff
@@ -10,11 +10,52 @@ class KubernetesNode < Sequel::Model
...
+  def init_health_monitor_session
+    {
+      ssh_session: sshable.start_fresh_session
+    }
+  end
+
+  def check_pulse(session:, previous_pulse:)
+    reading = begin
+      file_content = session[:ssh_session].exec!("cat :MESH_STATUS_FILE_PATH 2>/dev/null || echo ''", MESH_STATUS_FILE_PATH:)
+      if file_content.strip.empty?
```
jeremyevans: `cat` shouldn't result in extra whitespace, and using `-n` for `echo` should avoid adding a newline, so you don't need to `strip`.

```suggestion
      file_content = session[:ssh_session].exec!("cat :MESH_STATUS_FILE_PATH 2>/dev/null || echo -n", MESH_STATUS_FILE_PATH:)
      if file_content.empty?
```

model/kubernetes/kubernetes_node.rb
```diff
@@ -10,11 +10,52 @@ class KubernetesNode < Sequel::Model
...
+        "up" # File doesn't exist yet - CSI not updated, consider healthy
+      else
+        status = JSON.parse(file_content)
+        pods_status = status["pods"] || {}
```
jeremyevans: Is there a non-error situation where `pods` isn't in the file? It doesn't seem possible looking at the code, since `status_response` always includes it.

```suggestion
        pods_status = status["pods"]
```

model/kubernetes/kubernetes_node.rb
```diff
@@ -10,11 +10,49 @@ class KubernetesNode < Sequel::Model
...
+          errors = unreachable_pods.transform_values { |v| v["error"] }.compact
+          Clog.emit("Mesh connectivity issue detected", {kubernetes_node_mesh: {ubid:, unreachable_pods: unreachable_pods.keys, errors:}})
+          "down"
+        else
+          "up"
+        end
+      end
+    rescue JSON::ParserError => e
+      Clog.emit("Failed to parse mesh status file", {kubernetes_node_mesh: {ubid:, error: e.message}})
+      "down"
```
jeremyevans: Not sure whether this is worth a separate `rescue` block. Up to you, but you may want to consider just having the catch all `rescue` below handle it.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -17,9 +17,37 @@ def initialize(logger:, node_id:)
...
+      env_value.split(",").filter_map do |endpoint|
+        endpoint = endpoint.strip
+        next if endpoint.empty?
+        if endpoint.include?(":")
+          # rpartition splits on the LAST ":" to preserve IPv6 addresses:
+          #   "example.com:443"    -> ["example.com", ":", "443"]
+          #   "10.0.0.1:8080"      -> ["10.0.0.1", ":", "8080"]
+          #   "2001:db8::1:443"    -> ["2001:db8::1", ":", "443"] (IPv6 with port)
+          #   "2001:db8::1"        -> ["2001:db8:", ":", "1"] (bare IPv6 - port=1, likely invalid)
+          # For bare IPv6, users should use bracket notation [2001:db8::1]:443
```
jeremyevans: IPv6 handling is questionable here. Handling bare IPv4 correctly but not bare IPv6 correctly seems like a footgun. Unless you think it is problematic, I would require the port to always be specified.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+
+    def check_pod_connectivity(pod)
+      ip = pod["ip"]
+      name = pod["name"]
+      return unless ip
+
+      reachable = false
+      error = nil
+
```
jeremyevans:
```suggestion
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+      begin
+        socket = Socket.tcp(ip, REGISTRAR_HEALTHZ_PORT, connect_timeout: CONNECTION_TIMEOUT)
+        socket.close
+        reachable = true
+      rescue => e
+        error = "#{e.class}: #{e.message}"
```
jeremyevans:
```suggestion
        reachable = false
        error = "#{e.class}: #{e.message}"
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+
+    def update_pod_status(name, ip, reachable, error: nil)
+      @mutex.synchronize do
+        @pod_status[name] = {
+          ip:,
+          reachable:,
+          error:,
+          last_check: Time.now.utc.iso8601
+        }
```
jeremyevans:
```suggestion
      hash = {
        ip:,
        reachable:,
        error:,
        last_check: Time.now.utc.iso8601
      }
      @mutex.synchronize do
        @pod_status[name] = hash
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+    it "initializes correctly with logger and node_id" do
+      expect(checker.instance_variable_get(:@logger)).to eq(logger)
+      expect(checker.instance_variable_get(:@node_id)).to eq(node_id)
+      expect(checker.instance_variable_get(:@pod_status)).to eq({})
+      expect(checker.instance_variable_get(:@external_endpoints)).to eq([])
+      expect(checker.instance_variable_get(:@external_status)).to eq({})
+    end
+
+    it "parses external endpoints from ENV" do
+      allow(ENV).to receive(:[]).with("EXTERNAL_ENDPOINTS").and_return("10.0.0.1:443,api.example.com:8080")
```
jeremyevans: Do not mock this (only mock what you absolutely have to mock as a general rule):

```suggestion
      ENV["EXTERNAL_ENDPOINTS"] = "10.0.0.1:443,api.example.com:8080"
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+      checker_with_endpoints = described_class.new(logger:, node_id:)
+      endpoints = checker_with_endpoints.instance_variable_get(:@external_endpoints)
+      expect(endpoints).to eq([{host: "10.0.0.1", port: 443}, {host: "api.example.com", port: 8080}])
+    end
```
jeremyevans, line 19:
```suggestion
    ensure
      ENV.delete("EXTERNAL_ENDPOINTS")
    end
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -94,6 +124,49 @@ def update_pod_status(name, ip, reachable, error: nil)
       end
     end
 
+    def check_all_external_endpoints
+      endpoints = @mutex.synchronize { @external_endpoints.dup }
+      threads = endpoints.map do |endpoint|
+        Thread.new { check_external_endpoint(endpoint) }
```
jeremyevans: Consider a `connect_nonblock` approach instead of thread-per-endpoint.

model/kubernetes/kubernetes_node.rb
```diff
@@ -10,11 +10,52 @@ class KubernetesNode < Sequel::Model
...
+  def check_pulse(session:, previous_pulse:)
+    reading = begin
+      file_content = session[:ssh_session].exec!("cat :MESH_STATUS_FILE_PATH 2>/dev/null || echo ''", MESH_STATUS_FILE_PATH:)
+      if file_content.strip.empty?
+        "up" # File doesn't exist yet - CSI not updated, consider healthy
+      else
+        status = JSON.parse(file_content)
+        pods_status = status["pods"] || {}
+        unreachable_pods = pods_status.select { |_, v| v["reachable"] == false }
+        external_status = status["external_endpoints"] || {}
```
jeremyevans: If you are going to be rolling this out at once and not in stages, there shouldn't be a situation where this is not set:

```suggestion
        external_status = status["external_endpoints"]
```

If you are going to roll this out in stages, where `pods` could be set but `external_endpoints` may not be, it is fine to keep this as is temporarily.

kubernetes/csi/lib/ubi_csi/kubernetes_client.rb
```diff
@@ -27,12 +27,12 @@ def run_kubectl(*args, yaml_data: nil)
       output
     end
 
-    def yaml_load_kubectl(*)
-      YAML.safe_load(run_kubectl(*))
+    def yaml_load_kubectl(*args)
+      YAML.safe_load(run_kubectl(*args, "-oyaml"))
```
jeremyevans: Doesn't seem like you need to name the parameter here:

```suggestion
    def yaml_load_kubectl(*)
      YAML.safe_load(run_kubectl(*, "-oyaml"))
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,429 @@
+# frozen_string_literal: true
+
+require "logger"
+require "spec_helper"
+
+RSpec.describe Csi::MeshConnectivityChecker do
+  let(:logger) { Logger.new(File::NULL) }
+  let(:node_id) { "worker-1" }
+  let(:checker) { described_class.new(logger:, node_id:) }
+
```
jeremyevans, line 10: You should add the leaked thread detector here that I mentioned in https://github.com/ubicloud/ubicloud/pull/4702#discussion_r2760903679

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,241 @@
...
+      socket = Socket.new(address_family, :STREAM, 0)
+      # Enable keepalive to detect dead connections
+      socket.setsockopt(Socket::SOL_SOCKET, Socket::SO_KEEPALIVE, true)
+      begin
+        socket.connect_nonblock(sockaddr)
+        socket.close
+        :connected
+      rescue IO::WaitWritable, Errno::EINPROGRESS
+        socket
+      end
```
jeremyevans: This uses exceptions for flow control. It's possible to avoid that:

```suggestion
      if socket.connect_nonblock(sockaddr, exception: false) == :wait_writable
        socket
      else
        socket.close
        :connected
      end
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -0,0 +1,429 @@
...
+      [
+        {"name" => "ubicsi-nodeplugin-abc", "ip" => "10.0.0.1", "node" => "worker-1"},
+        {"name" => "ubicsi-nodeplugin-xyz", "ip" => "10.0.0.2", "node" => "worker-2"}
+      ]
+    }
+
+    it "fetches pods and builds targets, skipping same node" do
+      expect(Csi::KubernetesClient).to receive(:new).and_return(client)
+      expect(client).to receive(:get_nodeplugin_pods).and_return(pods)
+      expect(logger).to receive(:debug).with("[MeshConnectivity] Found 2 nodeplugin pods")
```
jeremyevans: You should add `.and_call_original` to all of these logger mocks. I'm not providing a suggestion this time, because when I provided a suggestion in my initial review, you applied it only to the that case and didn't update other cases.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,241 @@
...
+      env_value.split(",").filter_map do |endpoint|
+        endpoint = endpoint.strip
+        next if endpoint.empty?
+        # Use rpartition to split on the LAST ":" which handles both IPv4 and IPv6:
+        #   "example.com:443"    -> ["example.com", ":", "443"]
+        #   "10.0.0.1:8080"      -> ["10.0.0.1", ":", "8080"]
+        #   "2001:db8::1:443"    -> ["2001:db8::1", ":", "443"] (IPv6 with port)
+        parts = endpoint.rpartition(":")
+        if parts[1].empty?
+          @logger.warn("[MeshConnectivity] Port required in endpoint: #{endpoint}")
```
jeremyevans: Is a warning good enough here? Up to you, but it may be better to raise instead. I leave the decision to you, but if you are going to warn, you should make sure we have the appropriate monitoring to detect this situation and fix it.

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
mohi-kalantari: I added the leaked thread detector to the CSI's spec helper and I guess it is at global level. while you were reviewing I was making lots of changes and it's probable that you hadn't the latest version.

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
mohi-kalantari: You're right. I'll raise the exception so the app crashes and we get notified. The other methods to detect the situation were more complex

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
jeremyevans: Change looks good. Probably best not to request review until you are finished making changes. :)

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -0,0 +1,236 @@
...
+      return [] if env_value.nil? || env_value.strip.empty?
+      env_value.split(",").filter_map do |endpoint|
+        endpoint = endpoint.strip
+        next if endpoint.empty?
+        # Use rpartition to split on the LAST ":" which handles both IPv4 and IPv6:
+        #   "example.com:443"    -> ["example.com", ":", "443"]
+        #   "10.0.0.1:8080"      -> ["10.0.0.1", ":", "8080"]
+        #   "2001:db8::1:443"    -> ["2001:db8::1", ":", "443"] (IPv6 with port)
+        parts = endpoint.rpartition(":")
+        raise "Port required in endpoint: #{endpoint}" if parts[1].empty?
```
jeremyevans, line 43: Up to you, but you may want to consider raising if `parts[0].end_with?(":")`. That won't detect all cases where an IPv6 address is used without a port, but it will catch at least some.

PR #4703, opened 2026-02-02
Add /recycle endpoint for postgres resource

serprex:
Customers often want to be able to test their system's resilience to failovers

### DIFF
--- a/helpers/general.rb
+++ b/helpers/general.rb
@@ -120,6 +120,7 @@ class Clover < Roda
     disassociate
     disconnect
     promote
+    recycle
     remove_account
     remove_member
     reset_superuser_password
@@ -133,7 +134,7 @@ class Clover < Roda
     update_invitation
     upgrade
   ACTIONS
-  LOGGED_ACTIONS = Set.new(%w[create create_replica delete_all_cache_entries destroy promote reset_superuser_password restart restore update cancel_storage_auto_scale]).freeze
+  LOGGED_ACTIONS = Set.new(%w[create create_replica delete_all_cache_entries destroy promote recycle reset_superuser_password restart restore update cancel_storage_auto_scale]).freeze
 
   def audit_log(object, action, objects = [], project_id: @project.id)
     raise "unsupported audit_log action: #{action}" unless SUPPORTED_ACTIONS.include?(action)
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -400,6 +400,26 @@ class Clover
         end
       end
 
+      r.post "recycle" do
+        authorize("Postgres:edit", pg)
+
+        if pg.representative_server.nil?
+          raise CloverError.new(400, "InvalidRequest", "No representative server available to recycle")
+        end
+
+        DB.transaction do
+          pg.representative_server.incr_recycle
+          audit_log(pg, "recycle")
+        end
+
+        if api?
+          Serializers::Postgres.serialize(pg, {detailed: true})
+        else
+          flash["notice"] = "'#{pg.name}' will be recycled soon"
+          r.redirect pg, "/settings"
+        end
+      end
+
       r.post "reset-superuser-password" do
         authorize("Postgres:edit", pg)
         handle_validation_failure("postgres/show") { @page = "settings" }
--- a/spec/routes/web/project/postgres_spec.rb
+++ b/spec/routes/web/project/postgres_spec.rb
@@ -794,6 +794,16 @@ RSpec.describe Clover, "postgres" do
         expect(page).to have_no_content "Cancel automatic storage scale-up"
       end
 
+      it "can failover PostgreSQL database" do
+        visit "#{project.path}#{pg.path}/settings"
+        expect(page).to have_content "Recycle"
+        click_button "Recycle"
+
+        expect(page).to have_flash_notice "'#{pg.name}' will be recycled soon"
+        expect(page.status_code).to eq(200)
+        expect(pg.representative_server.recycle_set?).to be true
+      end
+
       it "doesn't show reset button when does not have permissions" do
         # Give permission to view, so we can see the detail page
         AccessControlEntry.create(project_id: project_wo_permissions.id, subject_id: user.id, action_id: ActionType::NAME_MAP["Postgres:view"])
--- a/cli-commands/pg/post/recycle.rb
+++ b/cli-commands/pg/post/recycle.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("recycle") do
+  desc "Request recycle of primary"
+
+  banner "ubi pg (location/pg-name | pg-id) recycle"
+
+  run do
+    id = sdk_object.recycle.id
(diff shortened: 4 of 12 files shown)


routes/project/location/postgres.rb
```diff
@@ -371,8 +371,23 @@ class Clover
...
+          pg.representative_server.incr_recycle
+          audit_log(pg, "failover")
+        end
+
+        Serializers::Postgres.serialize(pg, {detailed: true})
+      end
+
       r.post "reset-superuser-password" do
-        authorize("Postgres:view", pg)
+        authorize("Postgres:edit", pg)
```
furkansahin, line 424: this is a good change but since it is not related, can you move it to a different commit?

furkansahin (approved):

routes/project/location/postgres.rb
jeremyevans: I'm in favor of this change, but I think I can understand why it was `view` and not `edit` previously. The truth is if the user can view the connection information, they can see the current superuser password, and they can log into the database and then change the password for the database as they are the database superuser. However, since this affects other current servers and future servers, I agree `edit` is more appropriate.

helpers/general.rb
```diff
@@ -117,6 +117,7 @@ def default_headers
     detach_vm
     disassociate
     disconnect
+    failover
```
jeremyevans: Should probably be added to `LOGGED_ACTIONS` as well.

routes/project/location/postgres.rb
```diff
@@ -371,6 +371,21 @@ class Clover
...
+        if pg.representative_server.nil?
+          raise CloverError.new(400, "InvalidRequest", "No representative server available for failover")
+        end
+
+        DB.transaction do
+          pg.representative_server.incr_recycle
+          audit_log(pg, "failover")
+        end
+
+        Serializers::Postgres.serialize(pg, {detailed: true})
```
jeremyevans: This is only valid for an API response, not for a web response. The route doesn't appear to be limited to API requests. Unless there is a good reason not to add a web UI for this, I would add one, and change this route to handle web requests (`handle_validation_failure`, `flash["notice"]`, `r.redirect`).

spec/routes/api/cli/pg/failover_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  before do
+    expect(Config).to receive(:postgres_service_project_id).and_return(@project.id).at_least(:once)
+    cli(%w[pg eu-central-h1/test-pg create -s standard-2 -S 64])
+    @pg = PostgresResource.first
+  end
+
+  it "failover primary of postgres database" do
+    expect(Semaphore.where(strand_id: @pg.servers.first.id, name: "recycle")).to be_empty
+    expect(cli(%w[pg eu-central-h1/test-pg failover])).to eq "Failover requested for PostgreSQL database with id #{@pg.ubid}\n"
+    expect(Semaphore.where(strand_id: @pg.servers.first.id, name: "recycle")).not_to be_empty
```
jeremyevans: For new specs, I think we prefer using `change` for this type of check:

```suggestion
    expect(cli(%w[pg eu-central-h1/test-pg failover])).to eq("Failover requested for PostgreSQL database with id #{@pg.ubid}\n")
      .and change { Semaphore.where(strand_id: @pg.servers.first.id, name: "recycle").count }.from(0).to(1)
```

routes/project/location/postgres.rb
```diff
@@ -371,8 +371,23 @@ class Clover
         end
       end
 
+      r.post "failover" do
+        authorize("Postgres:edit", pg)
+
```
jeremyevans, line 405: We should only allow failover for HA resources.  If the resource is not HA, I'm not sure how failover could work. We should raise an exception in this case:

```suggestion

        if pg.ha_type == PostgresResource.ha_type_none
          raise CloverError.new(400, "InvalidRequest", "Can only request failover for high availability databases")
        end

```

jeremyevans (commented):
I agree that we should add this feature.

routes/project/location/postgres.rb
serprex: Failover on non HA brings up new standby to failover to. This can be used to test resilience to scaling ops, update packages to latest inage, or in future correct az

jeremyevans: Thank you for that information, I wasn't aware of that.

routes/project/location/postgres.rb
serprex: I couldn't tell where tests for this were, grepping for strings in restart's web response didn't bring anything up. Seems to rely on web testing which means adding button

jeremyevans: I looked for this and found our web specs for PostgreSQL restart are inadequate, to put it mildly: https://github.com/ubicloud/ubicloud/blob/b69e28281e47b83347c3b2fdfe2be4824a33a6af/spec/routes/web/project/postgres_spec.rb#L641-L647

No testing that the button even does anything, let alone testing for appropriate flash messages and page title or path after redirect.

serprex: I have used that test as an example for failover button spec

spec/routes/web/project/postgres_spec.rb
```diff
@@ -646,6 +646,14 @@ def create_minio_cluster_for_blob_storage
         expect(page.status_code).to eq(200)
       end
 
+      it "can failover PostgreSQL database" do
+        visit "#{project.path}#{pg.path}/settings"
+        expect(page).to have_content "Failover"
+        click_button "Failover"
+
+        expect(page.status_code).to eq(200)
```
jeremyevans: I know you are just copying the restart spec, but as I mentioned, it's inadequate and should be fixed. You should be testing the expected semaphore is incremented, the flash notice is as expected, and that the expected page is redirected to (either testing page title or path).

spec/routes/api/cli/pg/failover_spec.rb
serprex: Looks like rspec chokes mixing `.to eq(...).and change`

spec/routes/web/project/postgres_spec.rb
serprex: once tests are green I'll insert first commit fixing restart test & update this test accordingly

spec/routes/api/cli/pg/failover_spec.rb
jeremyevans: Sorry that I messed up the suggestion. With change, you need to use `expect {cli(...)}.to eq().and change {...}`

serprex: > You must pass an argument rather than a block to `expect` to use the provided matcher (eq "Failover requested for PostgreSQL database with id pgqp1sddt3my0kavwqz3p1z20k\n" and change `Semaphore.where(strand_id: @pg.servers.first.id, name: "recycle").count` from 0 to 1), or the matcher must implement `supports_block_expectations?`.

Fixed by folding `eq` inside `expect {}`

serprex:
@byucesoy feel like better split would be for endpoint on resource to be /failover & endpoint (which doesn't exist yet) on postgres server being /recycle

will add rename commit

PR #4704, opened 2026-02-02
Eren k8s external connectivity monitor

aamederen:
We have been seeing issues related to Kubernetes clusters
    having connectivity issues, therefore we aim to set periodic
    checks for specific clusters and get alerted for
    timely investigation.

We utilize the connectivity_check_target column in kubernetes_cluster
    by running checks from each node.

A `cluster_health_report` method that verifies external connectivity
    from each node's ubicsi-nodeplugin pod is added.

When configured, the wait loop runs the check every 2 minutes
    and pages on failure.

This check is planned to be moved to ubicsi plugin in the future.

### DIFF
--- a/prog/kubernetes/kubernetes_cluster_nexus.rb
+++ b/prog/kubernetes/kubernetes_cluster_nexus.rb
@@ -65,6 +65,25 @@ class Prog::Kubernetes::KubernetesClusterNexus < Prog::Base
     BillingRate.from_resource_properties(type, family, kubernetes_cluster.location.name)
   end
 
+  def check_external_connectivity
+    retries = 3
+    report = nil
+    begin
+      report = kubernetes_cluster.cluster_health_report
+    rescue => e
+      Clog.emit("Failed to get cluster health report", {kubernetes_cluster_id: kubernetes_cluster.id, exception: Util.exception_to_hash(e)})
+      retries -= 1
+      retry if retries > 0
+    end
+
+    if report&.all? { it[:healthy] }
+      Page.from_tag_parts("K8sExternalConnectivityFailed", kubernetes_cluster.ubid)&.incr_resolve
+    else
+      Prog::PageNexus.assemble("#{kubernetes_cluster.ubid} external connectivity unhealthy",
+        ["K8sExternalConnectivityFailed", kubernetes_cluster.ubid], kubernetes_cluster.ubid, extra_data: {report:})
+    end
+  end
+
   label def start
     register_deadline("wait", 120 * 60)
     Prog::Kubernetes::EtcdBackupNexus.assemble(kubernetes_cluster.id)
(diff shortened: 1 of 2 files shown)


model/kubernetes/kubernetes_cluster.rb
```diff
@@ -108,6 +108,24 @@ def port_diff_for_lb(load_balancer, desired_ports)
...
+
+    nodepools.flat_map(&:nodes).map do |kd|
+      pod_name = client.kubectl(
+        "get pods -n ubicsi --field-selector spec.nodeName=:nodename -o jsonpath='{.items[*].metadata.name}'",
+        nodename: kd.name
+      ).split.find { it.start_with?("ubicsi-nodeplugin-") }
+
+      next unless pod_name
+
+      host, port = connectivity_check_target.split(":", 2)
```
mohi-kalantari, line 122: This works well for our use case but only supports a single target. An array or comma separated list of targets would have been more usable.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -108,6 +108,24 @@ def port_diff_for_lb(load_balancer, desired_ports)
...
+      cmd = NetSsh.command("timeout 3 bash -c \"echo > /dev/tcp/:host/:port\" && echo OK-:nodename || echo FAIL", host:, port:, nodename: kd.name)
```
mohi-kalantari, line 123: is there a reason for concating node name to OK? since the hash we return already contains the kd.ubid

prog/kubernetes/kubernetes_cluster_nexus.rb
```diff
@@ -214,7 +214,23 @@ def billing_rate_for(type, family)
...
+        report = kubernetes_cluster.cluster_health_report
+        if report.all? { it[:healthy] }
+          Page.from_tag_parts("K8sExternalConnectivityFailed", kubernetes_cluster.ubid)&.incr_resolve
+        else
+          Prog::PageNexus.assemble("#{kubernetes_cluster.ubid} external connectivity unhealthy",
+            ["K8sExternalConnectivityFailed", kubernetes_cluster.ubid], kubernetes_cluster.ubid, extra_data: {report:})
+        end
+      rescue => e
+        Clog.emit("Failed to get cluster health report", {kubernetes_cluster_id: kubernetes_cluster.id, exception: Util.exception_to_hash(e)})
+      end
```
mohi-kalantari: nit: it would make the wait label more readable if we put these in a function imo

model/kubernetes/kubernetes_cluster.rb
aamederen: I thought about multiple targets, but given that this is an already expensive and temporary solution, I will omit for this once for simplicity.

model/kubernetes/kubernetes_cluster.rb
aamederen: this is more to make sure we catch the correct value in the output. the "OK" value can be emitted by a random/unrelated output during the execution and we don't want to catch it and think everything is OK by mistake, so we emit something unique and seek for it in the next line.

PR #4705, opened 2026-02-02
Enable PG upgrade E2E test

shikharbhardwaj:
* Update synchronization_status for read replicas when lsn_caught_up

In https://github.com/ubicloud/ubicloud/commit/0c889cf92039475136b3bc3958cef13131df5f3a, the failover_target selection changed to include a check for
synchronization_status in addition to the strand being in wait. This
affected read replica failovers, since the synchronization_status was
never updated for read replicas after the LSN difference was cleared.

* Enable PG Upgrade E2E test

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -10,7 +10,7 @@ on:
       test_cases:
         description: 'Comma-separated list of test cases (vm)'
         required: true
-        default: 'vm,github_runner_ubuntu_2404,github_runner_ubuntu_2204,postgres_standard,postgres_ha,kubernetes'
+        default: 'vm,github_runner_ubuntu_2404,github_runner_ubuntu_2204,postgres_standard,postgres_ha,postgres_upgrade,kubernetes'
         type: string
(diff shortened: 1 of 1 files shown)


byucesoy:
> In https://github.com/ubicloud/ubicloud/commit/0c889cf92039475136b3bc3958cef13131df5f3a, the failover_target selection changed to include a check for synchronization_status in addition to the strand being in wait.

We should revert this change. We already have check for "wait" label. It cannot switch to wait state without first having ready synchronization_status. I put a comment on this but it was after [the PR](https://github.com/ubicloud/ubicloud/pull/4683) got merged.

shikharbhardwaj:
@byucesoy makes sense, but do you think the change in this commit is good to have anyways, because the read replica server remains in `catching_up` forever otherwise?

byucesoy:
I think it should stay in catching_up because by definition it is continuously trying to catch up. We should change it to `ready` if read replica gets promoted though.

byucesoy (approved):

PR #4712, opened 2026-02-04
Introduce configurable subnet size

furkansahin:
## Add ipv4_range_size parameter to SubnetNexus.assemble
This allows callers to specify a custom IPv4 CIDR prefix length
when assembling a subnet, falling back to the existing defaults
for AWS and non-AWS locations when not specified.

## Cap AWS subnet CIDR to match private subnet prefix length
When creating an AWS subnet in NicNexus, use the private subnet's
actual prefix length (capped at /24) instead of always using /24.
This ensures that subnets with smaller ranges (e.g. /28) get
correctly sized AWS subnets.

## Use /28 subnet size for GitHub runner subnets
GitHub runners create a high volume of subnets which causes CIDR
conflicts with the default /16 range. Reduce the subnet size to
/28, the smallest allowed by AWS, to minimize address space
consumption and avoid conflicts.

### DIFF
--- a/bin/regen-screenshots
+++ b/bin/regen-screenshots
@@ -347,7 +347,7 @@ class RegenScreenshots
         project.id,
         name: "#{kc_ubid}-subnet",
         location_id: Location::HETZNER_FSN1_ID,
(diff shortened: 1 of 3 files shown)


enescakir:
By the way, the E2E metal tests failed. Not sure if it’s related.

PR #4715, opened 2026-02-04
Fix 2 N+1 query issues in Vm serializer detailed view

jeremyevans:
In detailed view, the VM serialier includes firewall information, which results in N+1 queries to get the location and firewall rules for each firewall. Eager load those to fix the issue.


serprex:
commit message should replace `serialier` with `serializer`

jeremyevans:
> commit message should replace `serialier` with `serializer`

Thanks for the heads up, fixed.

PR #4722, opened 2026-02-05
Add paging for KubernetesNode mesh connectivity

mohi-kalantari:
Implement the checkup mechanism for KubernetesNode to create pages when mesh connectivity issues persist.

When a node is down for 5+ consecutive readings over 15 minutes, it triggers a checkup.

### DIFF
--- a/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
@@ -22,7 +22,7 @@ spec:
       priorityClassName: system-node-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.4.1
+          image: ubicloud/ubicsi:v0.4.2
           imagePullPolicy: IfNotPresent
           env:
             - name: NODE_ID
--- a/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
@@ -34,7 +34,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.4.1
+          image: ubicloud/ubicsi:v0.4.2
           imagePullPolicy: IfNotPresent
           env:
             - name: NODE_ID


prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -45,12 +45,27 @@ def nodepool
...
   end
 
+  label def unavailable
+    if available?
+      decr_checkup
+      hop_wait
+    end
+
+    Clog.emit("KubernetesNode is unavailable due to mesh connectivity issues", {kubernetes_node_unavailable: {ubid: kubernetes_node.ubid, name: kubernetes_node.name}})
+    register_deadline("wait", 90)
```
jeremyevans: Are you sure 90 seconds is enough? We have `external connectivity unhealthy` pages that auto resolve after ~2 minutes. If this situation is similar to that, I'm not sure 90 seconds is enough. Remember that there is a significant cost every time there is a page. Only page if you are fairly sure operator action is needed to address the situation. It would be good to document what operator action is expected in such cases. If you don't expect the on-call to be able to fix such cases, I'm not sure there is a benefit to paging.

model/kubernetes/kubernetes_node.rb
```diff
@@ -26,31 +26,64 @@ def init_health_monitor_session
...
     rescue IOError, Errno::ECONNRESET
       raise
     rescue => e
       Clog.emit("Exception in KubernetesNode pulse check", Util.exception_to_hash(e, into: {ubid:}))
       "down"
     end
-    aggregate_readings(previous_pulse:, reading:)
+    pulse = aggregate_readings(previous_pulse:, reading:)
+
+    if pulse[:reading] == "down" && pulse[:reading_rpt] > 5 && Time.now - pulse[:reading_chg] > 30 && !reload.checkup_set?
```
jeremyevans: Do we need the `reload` here? Simply `checkup_set?` should be fine to ensure we don't add a semaphore on every run. If we have two semaphores instead of one, I don't think that's a problem.

Note that there is still a race condition even with the `reload`, since the semaphore could be set concurrently between the `reload.checkup_set?` and the `incr_checkup`.

spec/model/kubernetes/kubernetes_node_spec.rb
```diff
@@ -129,5 +135,119 @@
...
+      kn.incr_checkup
+
+      kn.check_pulse(session:, previous_pulse:)
+      expect(kn.reload.strand.semaphores.count { it.name == "checkup" }).to eq(1)
+    end
+  end
+
+  describe "#available?" do
+    it "returns true when check_mesh_availability returns available" do
+      expect(kn).to receive(:check_mesh_availability).and_return({available: true})
```
jeremyevans: Please move the mocking to a lower level, mocking `_exec!` or `_cmd`. Same in the following spec.

spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
```diff
@@ -104,6 +104,33 @@
...
+    it "hops to wait when node becomes available" do
+      nx.incr_checkup
+      expect(nx).to receive(:available?).and_return(true)
+      expect { nx.unavailable }.to hop("wait")
+      expect(kd.reload.checkup_set?).to be false
+    end
+
+    it "logs, registers deadline and naps when still unavailable" do
+      expect(nx).to receive(:available?).and_return(false)
+      expect(nx).to receive(:register_deadline).with("wait", 90)
```
jeremyevans: Check for deadline set in the frame after the `nx.unavailable` call, instead of mocking this method.

spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
```diff
@@ -104,6 +104,33 @@
...
+    it "hops to unavailable when checkup semaphore is set" do
+      nx.incr_checkup
+      expect { nx.wait }.to hop("unavailable")
+    end
+  end
+
+  describe "#unavailable" do
+    it "hops to wait when node becomes available" do
+      nx.incr_checkup
+      expect(nx).to receive(:available?).and_return(true)
```
jeremyevans: Similarly, change the mocking to `_cmd`/`_exec!` here and in the other specs.

mohi-kalantari:
I increased the page unavailability time to 15 minutes. I'll also write a TSG for it for the oncall by the end of tomorrow. thanks for the review @jeremyevans

spec/prog/kubernetes/kubernetes_node_nexus_spec.rb
```diff
@@ -105,6 +105,37 @@
...
+      expect { nx.unavailable }.to nap(15)
+      frame = nx.strand.stack.first
+      expect(frame["deadline_target"]).to eq("wait")
+      expect(Time.parse(frame["deadline_at"].to_s)).to be_within(3).of(Time.now + 15 * 60)
+    end
+  end
+
+  describe "#available?" do
+    it "delegates to kubernetes_node.available?" do
+      expect(nx.kubernetes_node).to receive(:available?).and_return(true)
```
jeremyevans: Switch to mocking at the `_cmd`/`_exec!` level, similar to earlier specs.

jeremyevans (approved):

PR #4723, opened 2026-02-05
Add arm64 alien runner support on AWS 

enescakir:
- **Add arch helper to GitHub runner nexus**
  

- **Add arm64 alien runner support on AWS**
  We already support x64 alien runners on AWS. The main blocker for arm64
  support was the lack of suitable AMI images. We can now build arm64 AMIs
  (see https://github.com/ubicloud/runner-images/pull/54), so there’s no
  longer a reason to hold off on arm64 support.
  
  Adding arm64 support to alien runners on AWS was straightforward. It
  required adding billing rates, updating the spilling logic, and
  including the arm64 AMI IDs.


config.rb
```diff
@@ -185,8 +185,10 @@ def self.unfrozen_test?
   override :github_ubuntu_2404_version, "20260202.1.0", string
   override :github_ubuntu_2204_version, "20260202.1.0", string
   override :github_gpu_ubuntu_2204_version, "20260202.1.0", string
-  override :github_ubuntu_2204_aws_ami_version, "ami-02609928906c79843", string
-  override :github_ubuntu_2404_aws_ami_version, "ami-04046eda554773409", string
+  override :github_ubuntu_2204_x64_aws_ami_version, "ami-02609928906c79843", string
+  override :github_ubuntu_2404_x64_aws_ami_version, "ami-04046eda554773409", string
```
jeremyevans, line 198: The change in the name could break installations that set one of the previously supported environment variables. Does that matter?

spec/prog/github/github_runner_nexus_spec.rb
```diff
@@ -23,7 +23,7 @@
   let(:installation) { runner.installation }
   let(:project) { installation.project }
   let(:client) { instance_double(Octokit::Client) }
-  let(:now) { Time.utc(2025, 8, 1, 19, 0) }
+  let(:now) { Time.utc(2026, 2, 5, 19, 0) }
```
jeremyevans, line 26: Seems an odd change. What was the reason behind this? Will similar changes be needed in the future?

jeremyevans (approved):
In general, this is a good feature to have.

spec/prog/github/github_runner_nexus_spec.rb
enescakir: We only select active billing rates. The `active_from` value for the new rates is today. Since we fixed the `Time.now` value in this file, it couldn’t find the new rates.

https://github.com/ubicloud/ubicloud/blob/329c7dfafdbf1d031d4775f7fb2941aaec3cca99/lib/billing_rate.rb#L19

furkansahin (approved):

config.rb
enescakir: We don't overwrite this config in any of our deployments, so it should be fine. We use it as a constant.

PR #4724, opened 2026-02-05
Fix resize operation for databases with init_script

byucesoy:
We missed one additional indirection of init_script. Because of that in the update call in below, we pass PostgresInitScript object instead of string, which causes the error.

### DIFF
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -60,7 +60,7 @@ class Clover
         target_storage_size_gib = typecast_params.pos_int("storage_size", pg.target_storage_size_gib)
         ha_type = typecast_params.nonempty_str("ha_type", pg.ha_type)
         tags = typecast_params.array(:Hash, "tags", pg.tags)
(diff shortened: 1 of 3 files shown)


routes/project/location/postgres.rb
```diff
@@ -60,7 +60,7 @@ class Clover
         target_storage_size_gib = typecast_params.pos_int("storage_size", pg.target_storage_size_gib)
         ha_type = typecast_params.nonempty_str("ha_type", pg.ha_type)
         tags = typecast_params.array(:Hash, "tags", pg.tags)
-        init_script = typecast_params.nonempty_str("init_script", pg.init_script)
+        init_script = typecast_params.nonempty_str("init_script", pg.init_script&.init_script)
```
jeremyevans: It's better to use this approach, so you don't load the object unless it is needed:

```suggestion
        init_script = typecast_params.nonempty_str("init_script") || pg.init_script&.init_script
```

PR #4727, opened 2026-02-06
X-Request-ID tracking

serprex:
POC tracking requests by tagging semaphores with X-Request-ID HTTP header

Currently initial semaphores are used which pass their request ids downstream to followup semaphores

An /operation endpoint is included which checks if an X-Request-ID has no remaining headers. Currently is returns a list of active semaphore names, but these values should only be used for debugging purposes. Completion is only indicated by returning an empty array

There may be some potential in a mechanism integrating request id with stack frames, similar to deadlines, but for POC I wanted to keep mechanism minimal

For our use case it's intended that one can reuse a request id in order to group multiple api calls as a single operation


migrate/20260330_add_semaphore_request_id.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:semaphore) do
+      add_column :request_ids, "text[]"
```
jeremyevans: What is the use case for more than one request id per semaphore? If the request id is set by the X-Request-ID header on creation, I'm not sure how there can be more than one. I know there is `convert_semaphore`, but that doesn't combine request IDs. If this does need to be an array, I would make it NOT NULL and have a default empty array value.

Using `text` for this seems suboptimal as I don't think we want to support arbitrarily large request id sizes. Maybe we can require the request id be a uuid? If uuid won't work, I would use `varchar(64)` or some similar small limit instead of `text`. Looking at later code, I think it may be better to use a join table instead of an array column.

migrate/20260330_add_semaphore_request_id.rb
```diff
@@ -0,0 +1,10 @@
...
+      add_index :request_ids, concurrently: false, using: "hash"
```
jeremyevans: This takes an exclusive lock on the table. It's generally not acceptable. This should be moved to a separate migration that does not use a transaction, with `concurrently: false` removed. If switching to a new table, then the index can be added inside the `create_table` normally, since there isn't anything to lock.

lib/semaphore_methods.rb
```diff
@@ -10,8 +10,8 @@ def self.configure(model, *semaphore_names)
 
       semaphore_names.each do |sym|
         name = sym.name
-        define_method :"incr_#{name}" do
-          Semaphore.incr(id, sym)
+        define_method :"incr_#{name}" do |request_id = nil|
```
jeremyevans, line 13: Optional positional arguments should generally be avoided as it complicates the addition of future optional arguments. This should be switched to a keyword argument. Same for `Semaphore.incr`.

model/semaphore.rb
```diff
@@ -20,8 +20,8 @@ def self.incr(id, name)
         .where(id:)
         .returning(:id)
         .with_sql(:update_sql, schedule: Sequel::CURRENT_TIMESTAMP))
-      .insert([:id, :strand_id, :name],
-        DB[:updated_strand].select(Sequel[:gen_timestamp_ubid_uuid].function(820), :id, name))
+      .insert([:id, :strand_id, :name, :request_id],
```
jeremyevans, line 23: The column in the migration is `request_ids`. I'm not sure how this worked, but my guess is the migration originally used `request_id` and then it was changed to `request_ids`. In general, if you are going to split a PR into separate commits, you should make sure each commit makes sense in order.

routes/project/github.rb
```diff
@@ -83,8 +83,8 @@ class Clover
           next unless (runner = @installation.runners_dataset.with_pk(id))
 
           DB.transaction do
-            runner.incr_skip_deregistration
-            runner.incr_destroy
+            runner.incr_skip_deregistration(request.get_header("X-RequestID"))
```
jeremyevans: There should be a helper method for this. Just in this PR alone, it would allow you to make the change in one place both times you changed what you were looking for, instead of changing every caller.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -527,12 +528,16 @@ def before_run
     end
 
     when_configure_metrics_set? do
+      request_ids = Semaphore.request_ids(:configure)
```
jeremyevans: `Semaphore.request_ids` is not defined as of this commit.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -528,16 +528,12 @@ def before_run
...
-      incr_reach_wait(request_ids)
+      convert_semaphore(:configure_metrics, :reach_wait)
       hop_configure_metrics
     end
 
     when_configure_set? do
-      request_ids = Semaphore.request_ids(:configure)
-      decr_configure
-      incr_reach_wait(request_ids)
+      convert_semaphore(:configure, :reach_wait)
```
jeremyevans, line 567: `convert_semaphore` is not defined as of this commit.

model/semaphore.rb
```diff
@@ -15,13 +15,17 @@ def self.incr(id, name, request_id = nil)
...
+
     with(:updated_strand,
       Strand
         .where(id:)
         .returning(:id)
         .with_sql(:update_sql, schedule: Sequel::CURRENT_TIMESTAMP))
-      .insert([:id, :strand_id, :name, :request_id],
-        DB[:updated_strand].select(Sequel[:gen_timestamp_ubid_uuid].function(820), :id, name, request_id))
+      .insert([:id, :strand_id, :name, :request_ids],
+        DB[:updated_strand].select(Sequel[:gen_timestamp_ubid_uuid].function(820), :id, name, request_ids))
```
jeremyevans: This probably need to use `Sequel.pg_array` to ensure `request_ids` is treated as a PostgreSQL array.

prog/base.rb
```diff
@@ -30,12 +30,22 @@ def #{name}
     end
   end
 
+  def get_request_ids(name)
+    request_ids = Semaphore.get_request_ids(@strand.id, name)
+  end
+
+  def convert_semaphore(name_from, name_into)
```
jeremyevans: What's the expected behavior regarding the `name_from` semaphore already existing? If it's expected to always exist, we should probably raise an exception here if it doesn't. If it's allowed not to be set, this seems OK, though maybe we should rename the method in that case, as I would not expect `convert_semaphore` to add a semaphore if one did not already exist.

routes/operation.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require "json"
+
+class Clover
+  hash_branch("operation") do |r|
```
jeremyevans: `completed` sounds like a better route name than `operation` if this is supposed to return true/false.

routes/operation.rb
```diff
@@ -0,0 +1,12 @@
...
+    r.post do
```
jeremyevans: Should use `r.post api? do`, as I expect this is only expected to handle API requests. Additionally, without an argument, this is not a terminal match, so it would currently match `.../operation/any/additional/garbage`.

routes/operation.rb
```diff
@@ -0,0 +1,12 @@
...
+      id = typecast_params.str("id")
```
jeremyevans: If we switch to a uuid, probably better to accept uuid or ubid as a route segment instead of a parameter. If you want to keep it as a parameter, should probably use `nonempty_str!` instead of `str`.

routes/operation.rb
```diff
@@ -0,0 +1,12 @@
...
+      Semaphore.where { request_ids.contains(id) }.select(:name).distinct.all.to_json
```
jeremyevans: If you did want to return an array of semaphore names, probably should use `distinct.select_map(:name)`. However, I consider semaphore names as internal state, and I don't think we should leak internal state. I think this should return true/false.

helpers/firewall.rb
```diff
@@ -14,7 +14,7 @@ def firewall_list_api_response(dataset)
     paginated_result(dataset.eager(:firewall_rules, :location), Serializers::Firewall)
   end
 
-  def firewall_post(firewall_name)
+  def firewall_post(firewall_name, request_ids = nil)
```
jeremyevans: Similarly, keyword argument here.

model/semaphore.rb
```diff
@@ -27,6 +27,10 @@ def self.incr(id, name, request_ids = nil)
       .insert([:id, :strand_id, :name, :request_ids],
         DB[:updated_strand].select(Sequel[:gen_timestamp_ubid_uuid].function(820), :id, name, request_ids))
   end
+
+  def self.get_request_ids(id, name)
+    DB.from { Sequel.lit("semaphore, unnest(request_ids)") }.get{array_agg(:unnest).distinct}
```
jeremyevans: The use of `unnest` here strongly indicates that we should have a separate table for semaphore request ids, instead of using an array column.

helpers/postgres.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Clover
-  def postgres_post(name)
+  def postgres_post(name, request_ids: nil)
```
jeremyevans: This is executed in route context. I'm not sure why you would need to pass `request_ids` in as an argument.

model/postgres/aws/postgres_timeline.rb
```diff
@@ -47,7 +47,7 @@ def aws_create_bucket
       location_constraint = (location.name == "us-east-1") ? nil : {location_constraint: location.name}
       begin
         blob_storage_client.create_bucket(bucket: ubid, create_bucket_configuration: location_constraint)
-      rescue Aws::S3::Errors::BucketAlreadyOwnedByYou
+      rescue ::Aws::S3::Errors::BucketAlreadyOwnedByYou
```
jeremyevans, line 50: Seems unrelated and probably should be in its own commit. From searching the codebase, this seems to be the only case where this error was made.

model/postgres/aws/postgres_timeline.rb
```diff
@@ -47,7 +47,7 @@ def aws_create_bucket
...
         # Ignore if bucket already exists
```
jeremyevans, line 51: To avoid similar issues, maybe we should use explicit `nil` values in such `rescue` blocks, to have the coverage tests check that these are actually hit.

prog/vm/aws/nexus.rb
```diff
@@ -241,8 +241,10 @@ def before_destroy
   end
 
   label def wait_sshable
-    unless vm.update_firewall_rules_set?
-      vm.incr_update_firewall_rules
+    request_ids = get_request_ids(:initial_provisioning)
+    unless vm.update_firewall_rules_set? || request_ids
+      vm.incr_update_firewall_rules(request_ids)
```
jeremyevans: This will continually add a semaphore every iteration while napping if there are request ids. We shouldn't add a semaphore if we have an existing semaphore with the same request ids.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -101,7 +101,8 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
 
   label def wait
     when_update_firewall_rules_set? do
-      private_subnet.vms.each(&:incr_update_firewall_rules)
+      request_ids = get_request_ids(:update_firewall_rulses)
```
jeremyevans: Looks like a typo (`rulses`).

jeremyevans (commented):
Overall thoughts:

* Returning semaphore names leaks too much internal state. We should return only true (completed) or false (not completed).

* Basing completion on whether semaphore exists is likely to be a suboptimal approach as it can result in false positives (showing completion before the request is actually completed). Often, semaphores are removed when the process related to the semaphore starts, not necessarily after it is finished. For example, the VM restart semaphore is removed when restart is submitted to the data plane, which is likely before the VM has fully restarted. That calls into question the entire approach. I think the `convert_semaphore` method tries to address this, but I'm not sure whether it handles all possible cases successfully.

* Modifying the semaphore table for this seems too invasive. At the very least, I think we should use a new table for (`semaphore_id`, `request_id`). I'm not sure whether we want to support this across the board, or just for PG-on-AWS. It adds a lot of complexity, and I'm not sure the benefits are worth it. If we only supported it for PG-on-AWS, we could probably use a less invasive approach.

* One possible more robust approach for this is having a separate prog for each action to be tracked, persisting the request_id when creating those progs, having the prog destroy action remove the request_id after the action has been fully completed, and having existing progs bud and reap (or maybe push) those progs. I think overall, that would result in fewer problems than a tracking request ids via semaphores.

migrate/20260330_add_semaphore_request_id.rb
serprex: Limiting to uuid works for us, wanted to keep that detail opaque to start

Multiplicity comes up if multiple ops become correlated to later downstream change. For example, many config updates could end with a single configure call. For POC I haven't pushed semaphore tracking as far as it likely should go, so that isn't to say it's easy to trigger in the current implementation

model/semaphore.rb
serprex: I should've squashed commits before submitting PR, as POC I started by writing code the way I thought it should look, only converging on working code at the end with testing

routes/operation.rb
serprex: I couldn't figure out how route segments work

model/postgres/aws/postgres_timeline.rb
serprex: bit surprised branch coverage misses this regardless

prog/vm/aws/nexus.rb
serprex: this is why I decrement semaphore afterwards, but agreed this felt ugly

routes/operation.rb
jeremyevans: If you want to accept a ubid segment, you could use `r.post api?, :ubid_uuid do |uuid|`, which will convert the ubid to to a uuid and yield the uuid.

To accept a uuid segment, around https://github.com/ubicloud/ubicloud/blob/main/clover.rb#L71, add a `:uuid` symbol matcher, probably something like:

```ruby
symbol_matcher(:uuid, /(\h{8}-\h{4}-\h{4}-\h{4}-\h{12})/)
```

Then you should be able to use `r.post api?, :uuid do |uuid|`.

For general information on how matchers (for matching route segments) work in Roda:

* https://roda.jeremyevans.net/rdoc/files/README_rdoc.html#label-Matchers
* https://roda.jeremyevans.net/rdoc/classes/Roda/RodaPlugins/SymbolMatchers.html

prog/vm/aws/nexus.rb
jeremyevans: Looks like you added a decrement of the `initial_provisioning` semaphore, not a decrement of the `update_firewall_rules` semaphore. Is that what you intended?

serprex: it is, by decrementing `initial_provisioning` future iterations won't pick up any request ids

model/postgres/aws/postgres_timeline.rb
serprex: #4735

prog/postgres/postgres_server_nexus.rb
```diff
@@ -68,6 +69,7 @@ def self.assemble(resource_id:, timeline_id:, timeline_access:, representative_a
         vm_id: vm_st.id,
         version: server_version
       )
+      postgres_server.incr_initial_provisioning(request_ids) if request_ids
```
furkansahin: I guess you can increment it whether or not the request_ids exist

furkansahin:
> Returning semaphore names leaks too much internal state. We should return only true (completed) or false (not completed).

Kind of agree, but this feature could be used to provide a more granular operation status update instead of binary. 


> Basing completion on whether semaphore exists is likely to be a suboptimal approach as it can result in false positives (showing completion before the request is actually completed). Often, semaphores are removed when the process related to the semaphore starts, not necessarily after it is finished. For example, the VM restart semaphore is removed when restart is submitted to the data plane, which is likely before the VM has fully restarted. That calls into question the entire approach. I think the convert_semaphore method tries to address this, but I'm not sure whether it handles all possible cases successfully.

I think it's better if we transformed our semaphore usage in a way that we decrement it only when we are sure the operation is completed. 

> Modifying the semaphore table for this seems too invasive. At the very least, I think we should use a new table for (semaphore_id, request_id). I'm not sure whether we want to support this across the board, or just for PG-on-AWS. It adds a lot of complexity, and I'm not sure the benefits are worth it. If we only supported it for PG-on-AWS, we could probably use a less invasive approach.

I don't agree with this. Especially considering the current semaphore design is put together this specific usecase in mind. I think the requests should be followed using the semaphore table. The previous design handled this just fine. We should also have this integrated into the whole system instead of ubi pg on aws. This was one of the most asked features in previous life and I am sure we'll see it gaining traction for the rest of the products as well.

> One possible more robust approach for this is having a separate prog for each action to be tracked, persisting the request_id when creating those progs, having the prog destroy action remove the request_id after the action has been fully completed, and having existing progs bud and reap (or maybe push) those progs. I think overall, that would result in fewer problems than a tracking request ids via semaphores.

I guess we could switch to using strands but I find this design more complex than the semaphore based one.

migrate/20260330_add_semaphore_request_id.rb
fdr: in this case you wouldn't need an array though, right, since the semaphore system is unary (we blow up the number of rows, so you can put one data element on each row, they clear together). I can see how this would descend from the non-unary variant.

fdr:
hey folks, sorry for the long delay

Good prototype here. It kind of ended up roughly as I expected in the routes, and not as I expected to a degree (which is a normal outcome of a prototype) in the assembly methods. It seems like there's some interaction with tracking prog execution that I think is interesting but we should explore avoiding. Although I chalk it up to prototype status, directly exposing semaphores is likely a bad idea for any kind of external interface. We do need some kind of mapping.

The whole reason we do pay for a row-per-increment (pivoting a normal multibyte integer to one row for each increment, a massive amplification) is exactly for this kind of metadata, and also finer update locking semantics (previous systems, I had to do more careful update deferral to gain and release locks fast enough). Maybe this was a bad idea, but it was the idea. Like @furkansahin, I thought a weakness of the old systems in tracking particular requests even to "some degree" was a worthy enough to experiment with baking it into the foundations of this one.

My general thinking about this was: nearly all progress states (true or false?) could be indicated by the existence-status of one more semaphores, i.e. you could write a routine for anything "interesting" you wanted to track. That adapter would be an interface hard to change, unlike our relatively free management of semaphores.

Another piece of general thinking was that NULL is very cheap in Postgres, being set via bitmap on disk, so internal signaling sharing the same overhead as inputs would be acceptable. Is it true? Perhaps. It does indicate one micro-optimization: it is much more expensive to store the request id in _every_ semaphore resulting from a web request rather than just one that covers what you need. So I don't think that we should feel obligated to paint every semaphore as part of a process with the request ID, only what we need to meet an interface we're willing to support.

Okay, long winding discursive impressions over. I think tangibly:

1. What progress/status reports do we want to establish as an interface?
2. Write one or a few more such adapters to give those status reports a taxonomy and repeatable software engineering model. 
3. Do we need to do array stuff? I really hope not.
4. Do we really need to invade the method signature of assemble? I really hope not.

If the design is compromised, then perhaps we need to think of unary semaphores (vs an integer) as a bloated failure and can consider leaning them back up to look more like the predecessor systems.

migrate/20260330_add_semaphore_request_id.rb
serprex: discussed: need the decr/incr pair to be handled by conversion, 5 input rows become 5 intermediate rows, sounds like `UPDATE`, current array system came about because decr might reduce 5 input rows into 1 intermediate row

fdr:
was talking to @serprex today, and the idea is that the request-id gets carried along to all incrs that cascade from it, i.e. form some kind of transitive closure. A request is complete when no semaphore row has that request ID anymore. This is somewhat elegantly related to the idea of impulse response, but I suppose the tricky bit is how the request IDs get carried along for the closure. This *might* be an interesting, unintended use of `when_foo_set?`, to track any incrs that result from the branch (original, `when` was intended for telemetry, e.g. to document `hop` transitions only, by rescuing the exception.

fdr:
I think I need to find some quality time (or make someone else find some quality time) on how to make this notationally nicer. I'm intrigued by the somewhat exact implementation of the impulse response model (as defined by somehow closing over the semaphores created as a result of the first one), but we need to increase the elegance factor, if we invade that much code to thread an identifier around (and possible miss something/misthread), then we should survey our other, similarly complex options.

prog/base.rb
serprex: convert no longer creates semaphore if one didn't exist beforehand

serprex:
given #5081 it seems like splitting up operations into dozens of progs is a non-starter

PR #4729, opened 2026-02-06
Use dynamic CIDR prefix in CNI IPv4 setup

mohi-kalantari:
Replace hardcoded /24 prefix with dynamic prefix extracted from the subnet configuration. The prefix is now derived from the subnet_ipv4 input parameter, allowing the CNI to work correctly with any subnet size (e.g., /24, /26).

### DIFF
--- a/model/kubernetes/kubernetes_node.rb
+++ b/model/kubernetes/kubernetes_node.rb
@@ -66,6 +66,10 @@ class KubernetesNode < Sequel::Model
     end
   end
 
+  def install_rhizome
+    Strand.create(prog: "InstallRhizome", label: "start", stack: [{subject_id: vm.sshable.id, target_folder: "kubernetes"}])
+  end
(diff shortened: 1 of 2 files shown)


model/kubernetes/kubernetes_node.rb
```diff
@@ -66,6 +66,10 @@ def billing_records
     end
   end
 
+  def install_rhizome
```
furkansahin, line 69: doesn't it make sense to make this part of the node roll-out?

mohi-kalantari: IIUC, in ProvisionKubernetesNode, we create a Strand for installing rhizome, that is done via `bud` which creates a child strand. But this one will be mostly used during maintenance operations and a separate strand is the preferred way. Like what we have for VmHosts.

PR #4730, opened 2026-02-06
PostgreSQL Auto Scale Up

byucesoy:
**Remove unused unlock function**
I found this while adding the decr_ helper methods for models. This is
left over from the old implementation and we stopped using it as part of
avoiding N+1 queries. All its old callers now use Semaphore.where(...)

**Use decr helper instead of deleting semaphores directly**
I found this while adding the decr helper function for models. In this
case, it is used in a prog, so it alrready has its own decr_ function.

**Add decr_ helper methods for models**
We have a similar helper method for incrementing semaphores, but we
don't have a helper method for decrementing them. This commit adds decr_
helper methods for models that use semaphores, allowing us to easily
decrement semaphores.

**Move generate_postgres_options to PostgresResource model**
We need to use `generate_postgres_options` in respirate as well. At the
moment, it is only available in we Clover app. Moving it to the model
allows us to reuse it in respirate and other parts of the system.

**Add automatic storage scaling for PostgreSQL databases**
Automatically scale up PostgreSQL storage when disk usage reaches
critical thresholds. The system monitors disk usage and takes
progressive action:
- At 80% and 85%: sends warning emails to project owners with
  details about the upcoming auto-scale
- At 90%: triggers auto-scaling by updating target storage size
  (and VM size if needed), including read replicas.

**Allow users to cancel in-progress storage auto-scale**
We use advisory locks to prevent race conditions with ongoing converge
operations. Users can cancel from the settings page while the operation
is in early stages (start, provision_servers, wait_servers_to_be_ready).

When canceled, target sizes are reset to current values, a notification
email is sent warning about potential unavailability at 100% usage. We
also create a page to notify on-call.

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -24,7 +24,8 @@ class PostgresResource < Sequel::Model
   plugin ProviderDispatcher, __FILE__
   plugin SemaphoreMethods, :initial_provisioning, :update_firewall_rules, :refresh_dns_record, :update_billing_records,
     :destroy, :promote, :refresh_certificates, :use_different_az, :use_old_walg_command, :check_disk_usage,
-    :storage_auto_scale_action_performed_80, :storage_auto_scale_action_performed_85, :storage_auto_scale_action_performed_90
+    :storage_auto_scale_action_performed_80, :storage_auto_scale_action_performed_85, :storage_auto_scale_action_performed_90,
+    :storage_auto_scale_canceled, :storage_auto_scale_not_cancellable
   include ObjectTag::Cleanup
 
   ServerExclusionFilters = Struct.new(:exclude_host_ids, :exclude_data_centers, :exclude_availability_zones, :availability_zone)
@@ -257,6 +258,7 @@ class PostgresResource < Sequel::Model
       Page.from_tag_parts("PGStorageAutoScaleMaxSize", id)&.incr_resolve
       Page.from_tag_parts("PGStorageAutoScaleQuotaInsufficient", id)&.incr_resolve
       Page.from_tag_parts("PGStorageAutoScaleCanceled", id)&.incr_resolve
+      decr_storage_auto_scale_canceled
     end
 
     return if disk_usage_percent < 80
@@ -271,7 +273,9 @@ class PostgresResource < Sequel::Model
 
     next_option = next_storage_auto_scale_option
 
-    extra_email_content = if next_option.nil?
+    extra_email_content = if storage_auto_scale_canceled_set?
+      :canceled_previously
+    elsif next_option.nil?
       Prog::PageNexus.assemble("#{ubid} high disk usage #{disk_usage_percent}%. No further auto-scaling possible.", ["PGStorageAutoScaleMaxSize", id], ubid, severity: "warning")
       :at_max_size
     elsif !project.quota_available?("PostgresVCpu", vcpu_delta(next_option))
@@ -290,8 +294,10 @@ class PostgresResource < Sequel::Model
         target_vm_size = next_option["size"]
         target_storage_size_gib = next_option["storage_size"]
 
-        update(target_vm_size:, target_storage_size_gib:)
-        read_replicas_dataset.update(target_vm_size:, target_storage_size_gib:)
+        unless storage_auto_scale_canceled_set?
+          update(target_vm_size:, target_storage_size_gib:)
+          read_replicas_dataset.update(target_vm_size:, target_storage_size_gib:)
+        end
       end
 
       send_storage_auto_scale_started_email(disk_usage_percent, next_option, extra_email_content)
@@ -319,15 +325,53 @@ class PostgresResource < Sequel::Model
     vcpu_delta * total_server_count
   end
 
+  def storage_auto_scale_lock_key
+    Digest::SHA2.digest(id).unpack1("q>").abs
+  end
+
+  def can_cancel_storage_auto_scale?
+    return false if storage_auto_scale_canceled_set? || storage_auto_scale_not_cancellable_set? || !storage_auto_scale_action_performed_90_set?
+
+    converge_strand = strand.children_dataset.where(prog: "Postgres::ConvergePostgresResource").first
+    return false unless converge_strand
+
+    ["start", "provision_servers", "wait_servers_to_be_ready", "wait_for_maintenance_window"].include?(converge_strand.label)
+  end
+
+  def cancel_storage_auto_scale
+    DB.transaction do
+      # Try to acquire advisory lock to prevent race with failover
+      unless DB.get(Sequel.function(:pg_try_advisory_xact_lock, storage_auto_scale_lock_key))
+        return false
+      end
+
+      return false unless can_cancel_storage_auto_scale?
+
+      current_storage_size_gib = representative_server.storage_size_gib
+      update(target_vm_size: vm_size, target_storage_size_gib: current_storage_size_gib)
+      read_replicas_dataset.update(target_vm_size: vm_size, target_storage_size_gib: current_storage_size_gib)
+
+      incr_storage_auto_scale_canceled
+
+      Prog::PageNexus.assemble("#{ubid} storage auto-scale canceled by user", ["PGStorageAutoScaleCanceled", id], ubid, severity: "warning")
+
+      send_storage_auto_scale_canceled_email
+
+      true
+    end
+  end
+
   def send_storage_auto_scale_warning_email(usage_percent, next_option, extra_content)
     body = [
       "Your PostgreSQL database '#{name}' (#{ubid}) has reached #{usage_percent}% disk usage.",
       "You are currently using #{storage_size_gib * usage_percent / 100} of #{storage_size_gib} GB of storage."
     ]
 
-    if [:at_max_size, :quota_insufficient].include?(extra_content)
+    if [:canceled_previously, :at_max_size, :quota_insufficient].include?(extra_content)
       body << "Automated disk scaling is normally triggered when disk usage exceeds 90%."
-      body << if extra_content == :at_max_size
+      body << if extra_content == :canceled_previously
+        "However, you previously canceled auto-scaling, so auto-scaling will stay deactivated until disk usage drops below 80%."
+      elsif extra_content == :at_max_size
         "However, your database has already reached the maximum available storage size, so auto-scaling cannot proceed."
       else
         "However, your project does not have sufficient quota, so auto-scaling cannot proceed."
@@ -360,9 +404,11 @@ class PostgresResource < Sequel::Model
       "You are currently using #{storage_size_gib * usage_percent / 100} of #{storage_size_gib} GB of storage."
     ]
 
-    if [:at_max_size, :quota_insufficient].include?(extra_content)
+    if [:canceled_previously, :at_max_size, :quota_insufficient].include?(extra_content)
       body << "Auto-scaling would normally begin at this threshold."
-      body << if extra_content == :at_max_size
+      body << if extra_content == :canceled_previously
+        "However, you previously canceled auto-scaling, so auto-scaling will stay deactivated until disk usage drops below 80%."
+      elsif extra_content == :at_max_size
         "However, your database has already reached the maximum available storage size."
       else
         "However, your project does not have sufficient quota."
@@ -384,6 +430,28 @@ class PostgresResource < Sequel::Model
     Util.send_email(
       accounts_with_access.map(&:email).uniq,
       "PostgreSQL Auto-Scaling: #{name}",
+      cc: Config.mail_from,
+      greeting: "Hello,",
+      body:,
+      button_title: "View Database",
+      button_link: "#{Config.base_url}#{project.path}#{path}"
+    )
+  end
+
+  def send_storage_auto_scale_canceled_email
+    body = [
+      "Auto-scaling for your PostgreSQL database '#{name}' (#{ubid}) has been canceled as requested.",
+      "Automatic scale-up will not be re-triggered until disk usage drops below 80% and rises again.",
+      "The target configuration has been reset to the current values:",
+      "Storage: #{representative_server.storage_size_gib} GB",
+      "Instance size: #{vm_size}",
+      "Please note that if disk usage reaches to 100%, database would become unavailable.",
+      "We recommend freeing up disk space or contacting support to discuss other options."
+    ]
+
+    Util.send_email(
+      accounts_with_access.map(&:email).uniq,
+      "PostgreSQL Auto-Scaling Canceled: #{name}",
       cc: Config.mail_from,
       greeting: "Hello,",
       body:,
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -434,7 +434,7 @@ class PostgresServer < Sequel::Model
   def observe_disk_usage(session)
     disk_usage_percent = session[:ssh_session].exec!("df --output=pcent /dat | tail -n 1").strip.delete("%").to_i
     if reload.primary?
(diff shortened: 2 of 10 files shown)


model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,155 @@ def ready_for_read_replica?
...
   end
 
+  def handle_storage_auto_scale
+    return unless representative_server
+
+    disk_usage_percent = representative_server.vm.sshable.cmd("df --output=pcent /dat | tail -n 1").strip.delete("%").to_i
+
+    # Clear semaphores only when usage drops at least 3% below threshold to avoid spurious emails
+    [90, 85, 80].each {
+      Semaphore.where(strand_id: strand.id, name: "storage_auto_scale_#{it}_percent_action").destroy if disk_usage_percent <= it - 3
```
enescakir: Since the main reasoning behind the model hiding is to avoid using dataset methods directly on the model, you could use something like `strand.semaphores_dataset.where(name: ...).destroy` or you can also add a `destroy_semaphore` helper to the `Semaphore` model and call that instead.

model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,155 @@ def ready_for_read_replica?
...
+      if next_option["size"] != representative_server.vm.display_size
+        body << "Since your current instance size does not support the new storage size, the instance will also be upgraded from #{representative_server.vm.display_size} to #{next_option["size"]}."
+      end
+      body << "Disk scaling requires a cutover to a new server, which will result in a short downtime, typically less than a minute. The scaling operation will be triggered automatically once disk usage reaches 90%."
+      body << "If you wish to scale sooner or have any questions, please contact support."
+      body << "Your database also has read replica(s), which will be scaled alongside the primary instance." if read_replicas.any?
+    end
+
+    Util.send_email(
+      project.accounts.map(&:email),
```
enescakir: The account might not have even view permission for the PostgreSQL resources. I think we should send it to anyone who has any permission on them.

model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,155 @@ def ready_for_read_replica?
...
+
+    next_option = next_storage_auto_scale_option
+
+    extra_email_content = if next_option.nil?
+      Prog::PageNexus.assemble("#{ubid} high disk usage #{disk_usage_percent}%. No further auto-scaling possible.", ["PGStorageAutoScaleMaxSize", id], ubid, severity: "warning")
+      :at_max_size
+    elsif !project.quota_available?("PostgresVCpu", vcpu_delta(next_option))
+      Prog::PageNexus.assemble("#{ubid} high disk usage #{disk_usage_percent}%. Quota insufficient for auto-scale.", ["PGStorageAutoScaleQuotaInsufficient", id], ubid, severity: "warning")
+      :quota_insufficient
+    end
```
enescakir: I hope you'll write a TSG for the new pages.

Also, what are the action items for on-call besides monitoring? If it fills up, we’ll get unavailable page anyway.

model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,155 @@ def ready_for_read_replica?
...
+      body << if extra_content == :at_max_size
+        "However, your database has already reached the maximum available storage size, so auto-scaling cannot proceed."
+      else
+        "However, your project does not have sufficient quota, so auto-scaling cannot proceed."
+      end
+      body << "Please free up disk space or contact support for assistance."
+    else
+      body << "When disk usage reaches 90%, storage will be automatically increased to #{next_option["storage_size"]} GB."
+      if next_option["size"] != representative_server.vm.display_size
+        body << "Since your current instance size does not support the new storage size, the instance will also be upgraded from #{representative_server.vm.display_size} to #{next_option["size"]}."
```
enescakir: We use `representative_server.vm.display_size` lots of times

```ruby
def display_size
    @display_size ||= representative_server.vm.display_size
end
```

model/postgres/postgres_resource.rb
enescakir: Another option, you can add `decr_***` helper to models as well, similar to progs.

prog/postgres/postgres_resource_nexus.rb
```diff
@@ -267,6 +267,10 @@ def before_run
       hop_refresh_dns_record
     end
 
+    when_check_disk_usage_set? do
+      postgres_resource.handle_storage_auto_scale
```
enescakir, line 274: `handle_storage_auto_scale` calls `.sshable.cmd`. If this raises an exception due to an SSH connection issue, we get stuck in the wait state. It also doesn’t check the subsequent semaphores. We might add a error handling for it.

model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,155 @@ def ready_for_read_replica?
     !needs_convergence? && !PostgresTimeline.earliest_restore_time(timeline).nil?
   end
 
+  def handle_storage_auto_scale
+    return unless representative_server
+
+    disk_usage_percent = representative_server.vm.sshable.cmd("df --output=pcent /dat | tail -n 1").strip.delete("%").to_i
+
+    # Clear semaphores only when usage drops at least 3% below threshold to avoid spurious emails
+    [90, 85, 80].each {
```
enescakir, line 253: We use the `[90, 85, 80]` list in several places. Should we extract this into a constant?

model/postgres/postgres_resource.rb
enescakir: Another option, you can add `decr_***` helper to models as well, similar to progs.

model/postgres/postgres_resource.rb
byucesoy: When we get unavailable page, it is already too late. On-call can monitor, reach out to customer proactively, decide to increase the quota, modify reserved file system space, switch to read only mode etc.

I'll add TSGs.

model/postgres/postgres_resource.rb
byucesoy: We only use it in 3 places and now it got dropped to 2. I think it is OK to leave this as is.

model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,164 @@ def ready_for_read_replica?
...
+      Clog.emit("Failed to check disk usage for #{ubid}, skipping storage auto-scale check", representative_server)
+      return
+    end
+
+    # Clear semaphores only when usage drops at least 3% below threshold to avoid spurious emails
+    [90, 85, 80].each {
+      send("decr_storage_auto_scale_#{it}_percent_action") if disk_usage_percent <= it - 3
+    }
+
+    if disk_usage_percent <= 77
```
furkansahin, line 257: should we name these (77, 80, 85, 90) with explanatory names

model/postgres/postgres_resource.rb
```diff
@@ -22,7 +22,9 @@ class PostgresResource < Sequel::Model
   plugin ResourceMethods, redacted_columns: [:root_cert_1, :root_cert_2, :server_cert, :trusted_ca_certs],
     encrypted_columns: [:superuser_password, :root_cert_key_1, :root_cert_key_2, :server_cert_key]
   plugin ProviderDispatcher, __FILE__
-  plugin SemaphoreMethods, :initial_provisioning, :update_firewall_rules, :refresh_dns_record, :update_billing_records, :destroy, :promote, :refresh_certificates, :use_different_az, :use_old_walg_command
+  plugin SemaphoreMethods, :initial_provisioning, :update_firewall_rules, :refresh_dns_record, :update_billing_records,
+    :destroy, :promote, :refresh_certificates, :use_different_az, :use_old_walg_command, :check_disk_usage,
+    :storage_auto_scale_80_percent_action, :storage_auto_scale_85_percent_action, :storage_auto_scale_90_percent_action
```
furkansahin: can you move 80, 85, 90 to the end of the semaphore name and dynamically decr semaphores where you decrement all three?

furkansahin: I think it's also easier to distinguish to the eye when they are at the end

model/postgres/postgres_resource.rb
```diff
@@ -24,7 +24,8 @@ class PostgresResource < Sequel::Model
   plugin ProviderDispatcher, __FILE__
   plugin SemaphoreMethods, :initial_provisioning, :update_firewall_rules, :refresh_dns_record, :update_billing_records,
     :destroy, :promote, :refresh_certificates, :use_different_az, :use_old_walg_command, :check_disk_usage,
-    :storage_auto_scale_80_percent_action, :storage_auto_scale_85_percent_action, :storage_auto_scale_90_percent_action
+    :storage_auto_scale_80_percent_action, :storage_auto_scale_85_percent_action, :storage_auto_scale_90_percent_action,
+    :storage_auto_scale_canceled
```
furkansahin: this is more of a product feedback; I'm not sure if this functionality would give the perception of "opting-out" of auto scaling. Because now, if they cancel auto scale and hit back the 90% limit, the auto scale will start again. A customer having that perspective would not expect us to auto-scale. I wonder if we should not integrate that in the UI but keep it at the backend and let them send a support e-mail.

model/postgres/postgres_server.rb
```diff
@@ -429,6 +431,19 @@ def observe_metrics_backlog(session)
...
+    disk_usage_percent = session[:ssh_session].exec!("df --output=pcent /dat | tail -n 1").strip.delete("%").to_i
+    if disk_usage_percent >= 77
+      if reload.primary?
+        resource.incr_check_disk_usage
+      elsif disk_usage_percent >= 95
+        Prog::PageNexus.assemble("High disk usage on non-primary PG server (#{disk_usage_percent}%)", ["PGDiskUsageHigh", id], ubid, severity: "warning", extra_data: {disk_usage_percent:})
+      end
+    end
+  rescue => ex
+    Clog.emit("Failed to observe disk usage", {postgres_server_id: id, exception: Util.exception_to_hash(ex)})
```
enescakir: `Util.exception_to_hash` already adds `exception` key

- https://github.com/ubicloud/ubicloud/pull/4511
- https://github.com/ubicloud/ubicloud/pull/4465
```suggestion
    Clog.emit("Failed to observe disk usage", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
```

model/postgres/postgres_server.rb
```diff
@@ -429,6 +431,19 @@ def observe_metrics_backlog(session)
     Clog.emit("Failed to observe metrics backlog", {postgres_server_id: id, exception: Util.exception_to_hash(ex)})
```
enescakir: You probably copied the rescue block from there. It's also wrong.

```suggestion
    Clog.emit("Failed to observe metrics backlog", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
```

prog/postgres/converge_postgres_resource.rb
```diff
@@ -100,8 +100,12 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
   label def recycle_representative_server
     if (rs = postgres_resource.representative_server) && !postgres_resource.ongoing_failover?
       hop_prune_servers unless rs.needs_recycling?
+      hop_prune_servers if postgres_resource.storage_auto_scale_canceled_set?
       hop_provision_servers unless postgres_resource.has_enough_ready_servers?
 
+      # Acquire advisory lock to prevent race with cancel_storage_auto_scale
+      nap 5 unless DB.get(Sequel.function(:pg_try_advisory_xact_lock, postgres_resource.storage_auto_scale_lock_key))
```
furkansahin, line 107: so, if we can get the lock and continue with the planned failover, and during that, the customer asks for cancellation, we return success but it's actually too late, no? The locks are there until the end of this transaction, am I wrong?

model/postgres/postgres_resource.rb
byucesoy: I renamed semaphores to put the number at the end. I'm keeping the place where I decremented all three as is though, because both options are 3 lines and explicit version is more readable.

model/postgres/postgres_resource.rb
byucesoy: My intention is not giving an permanent opt-out functionality. The goal was allowing canceling the scale up temporarily if customer is in a position to not accept downtime at that moment.

prog/postgres/converge_postgres_resource.rb
byucesoy: If we are already in `recycle_representative_server` state, customer cannot cancel the scale up anymore.

This lock protects against the following case:
- Converge prog is at `wait_for_maintenance_window`
- Customer triggers cancel, which succeeds because prog is at `wait_for_maintenance_window`. As a part of transaction, it increments cancel semaphore. Transaction is not committed yet.
- Converge prog gets loaded, along with semaphores. Cancel transaction is not committed, so we don't see that cancel semaphore is incremented. We might end up proceeding with failover, but we don't because we need to wait for the lock. Since we cannot get the lock, we nap and in next run, we read the new semaphore snapshot, which includes cancel semaphore as well.

model/postgres/postgres_resource.rb
byucesoy: I think that is just too much scaffolding & indirection. Little bit of hard coded "magic" numbers are fine :slightly_smiling_face:

model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,164 @@ def ready_for_read_replica?
...
+  def vcpu_delta(target_option)
+    current_vcpu = Option::POSTGRES_SIZE_OPTIONS[vm_size].vcpu_count
+    new_vcpu = Option::POSTGRES_SIZE_OPTIONS[target_option["size"]].vcpu_count
+    vcpu_delta = new_vcpu - current_vcpu
+
+    total_server_count = target_server_count + read_replicas.sum(&:target_server_count)
+    vcpu_delta * total_server_count
+  end
+
+  def send_storage_auto_scale_warning_email(usage_percent, next_option, extra_content)
```
shikharbhardwaj, line 364: nit: would `helpers/postgres.rb` be a better place for this (and potentially others)?

model/postgres/postgres_server.rb
```diff
@@ -429,6 +431,19 @@ def observe_metrics_backlog(session)
     Clog.emit("Failed to observe metrics backlog", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
   end
 
+  def observe_disk_usage(session)
+    disk_usage_percent = session[:ssh_session].exec!("df --output=pcent /dat | tail -n 1").strip.delete("%").to_i
+    if disk_usage_percent >= 77
+      if reload.primary?
+        resource.incr_check_disk_usage
+      elsif disk_usage_percent >= 95
+        Prog::PageNexus.assemble("High disk usage on non-primary PG server (#{disk_usage_percent}%)", ["PGDiskUsageHigh", id], ubid, severity: "warning", extra_data: {disk_usage_percent:})
```
shikharbhardwaj: Should we resolve this page when it goes lower?

model/postgres/postgres_resource.rb
```diff
@@ -236,6 +238,164 @@ def ready_for_read_replica?
...
+  end
+
+  def send_storage_auto_scale_started_email(usage_percent, next_option, extra_content)
+    body = [
+      "Your PostgreSQL database '#{name}' (#{ubid}) has reached #{usage_percent}% disk usage.",
+      "You are currently using #{storage_size_gib * usage_percent / 100} of #{storage_size_gib} GB of storage."
+    ]
+
+    if [:at_max_size, :quota_insufficient].include?(extra_content)
+      body << "Auto-scaling would normally begin at this threshold"
```
shikharbhardwaj: nit: new sentence
```suggestion
      body << "Auto-scaling would normally begin at this threshold."
```

shikharbhardwaj (approved):

model/postgres/postgres_resource.rb
byucesoy: `helper/postgres.rb` is in Clover app's namespace. These functions are getting called from the respirate.

PR #4732, opened 2026-02-06
Add knotc telemetry

fdr:
Per comment, stdin is useful as it's faster than forking many knotc commands, in spite of the difficulty aligning and matching error output.

However, per recent incident, it's probably not too voluminous to record.  We could have reacted far faster and learned a lot more with materialized copies of what was done, and when.


fdr:
@jeremyevans good idea, added it as vm_ubid

PR #4734, opened 2026-02-06
Fix nondeterministic dns zone modifiction in Prog::DnsZone::DnsZoneNexus

jeremyevans:
For records with the same created_at, remove any tombstoned records before adding new records. Otherwise, we can add a record and then remove it.

Progs that both delete_record and insert_record in the same label (same transaction) use delete_record first, so the intent is to update the data for the record, not delete the record.

### DIFF
--- a/prog/dns_zone/dns_zone_nexus.rb
+++ b/prog/dns_zone/dns_zone_nexus.rb
@@ -26,7 +26,7 @@ class Prog::DnsZone::DnsZoneNexus < Prog::Base
       records_to_rectify = dns_zone.records_dataset
         .left_join(:seen_dns_records_by_dns_servers, dns_record_id: :id, dns_server_id: dns_server.id)
         .where(Sequel[:seen_dns_records_by_dns_servers][:dns_record_id] => nil)
-        .order(:created_at).all
+        .order(:created_at, Sequel.desc(:tombstoned)).all
 
       next if records_to_rectify.empty?
 
--- a/spec/prog/dns_zone/dns_zone_nexus_spec.rb
+++ b/spec/prog/dns_zone/dns_zone_nexus_spec.rb
@@ -74,7 +74,7 @@ COMMANDS
     end
 
     it "ignores unimportant errors" do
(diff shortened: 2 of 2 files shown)


jeremyevans:
> It is conservative, preferring liveness over deadness, but as @byucesoy analyzed, when the target state is deadness and there is a tie, it is wrong.

My understanding is this can only happen if, inside the same transaction, there is a `insert_record` followed by a `delete_record` (multiple records with same created_at, target state deadness). First, such a thing would not make logical sense if you read the code. Even if it did happen, with deterministic behavior enforced by the order, it's likely it would be caught quickly.

I audited all uses of `delete_record`.

Before `insert_record` in same label:

* Prog::Postgres::PostgresResourceNexus#refresh_dns_record
* Prog::Vnet::LoadBalancerNexus#rewrite_dns_records

During destruction, no `insert_record`:

* Prog::Kubernetes::KubernetesClusterNexus#destroy
* Prog::Minio::MinioServerNexus#destroy
* Prog::Postgres::PostgresResourceNexus#wait_children_destroy
* Prog::Vnet::LoadBalancerNexus#wait_all_vms_removed

Special case:

* Prog::Vnet::CertNexus#cleanup_dns_challenge_records, called by various labels, but `insert_record` is only called in `start`, which doesn't call `cleanup_dns_challenge_records`

Based on these results, I believe this change to be safe. I find it unlikely we would introduce a case that would result in the theoretical problem @byucesoy noted.

byucesoy:
>> It is conservative, preferring liveness over deadness, but as @byucesoy analyzed, when the target state is deadness and there is a tie, it is wrong.
>
> My understanding is this can only happen if, inside the same transaction, there is a insert_record followed by a delete_record (multiple records with same created_at, target state deadness). First, such a thing would not make logical sense if you read the code. Even if it did happen, with deterministic behavior enforced by the order, it's likely it would be caught quickly.

I don’t think anyone would intentionally call insert_record and delete_record back to back. However, it’s possible that different functions are invoked within the same transaction and inadvertently lead to this situation, so I wouldn’t consider it unlikely.

That said, I’m fine moving forward with this approach for now. It’s still an improvement over the current implementation, and we can revisit it later to identify and implement a theoretically correct solution.

PR #4737, opened 2026-02-09
Put ParadeDB behind a feature flag

byucesoy:


enescakir (approved):
The only concern for me is that we list these options in the CLI as valid choices.

If we expect only a small number of users to need this, we might consider removing them from the CLI help altogether in the future.

https://github.com/ubicloud/ubicloud/blob/185b887132d93400181fff10c1f10113f9b48440/cli-commands/pg/post/create.rb#L7

byucesoy:
> The only concern for me is that we list these options in the CLI as valid choices.
> 
> If we expect only a small number of users to need this, we might consider removing them from the CLI help altogether in the future.
> 
> https://github.com/ubicloud/ubicloud/blob/185b887132d93400181fff10c1f10113f9b48440/cli-commands/pg/post/create.rb#L7

Agreed. However currently we list other options that are available to only specific set of customers in help output as well (such as various instance sizes). The good news is that we have the project information at the time we show the help output, so that we can generate valid options and show valid ones. Though, it is bit involved, so I didn't spend time on that yet.

jeremyevans:
> > The only concern for me is that we list these options in the CLI as valid choices.
> > If we expect only a small number of users to need this, we might consider removing them from the CLI help altogether in the future.
> > https://github.com/ubicloud/ubicloud/blob/185b887132d93400181fff10c1f10113f9b48440/cli-commands/pg/post/create.rb#L7
> 
> Agreed. However currently we list other options that are available to only specific set of customers in help output as well (such as various instance sizes). The good news is that we have the project information at the time we show the help output, so that we can generate valid options and show valid ones. Though, it is bit involved, so I didn't spend time on that yet.

Currently, our rodish configuration doesn't support context sensitive help (changing help per-request based on the project). I have a WIP branch for that, so if we want the ability to do that, it can be added.

byucesoy:
> > > The only concern for me is that we list these options in the CLI as valid choices.
> > > If we expect only a small number of users to need this, we might consider removing them from the CLI help altogether in the future.
> > > https://github.com/ubicloud/ubicloud/blob/185b887132d93400181fff10c1f10113f9b48440/cli-commands/pg/post/create.rb#L7
> > 
> > 
> > Agreed. However currently we list other options that are available to only specific set of customers in help output as well (such as various instance sizes). The good news is that we have the project information at the time we show the help output, so that we can generate valid options and show valid ones. Though, it is bit involved, so I didn't spend time on that yet.
> 
> Currently, our rodish configuration doesn't support context sensitive help (changing help per-request based on the project). I have a WIP branch for that, so if we want the ability to do that, it can be added.

Ah, my mistake then. I think context sensitive help would be useful, but not urgent IMO.

furkansahin (approved):

PR #4738, opened 2026-02-09
Rename burstable PostgreSQL instances as hobby

byucesoy:
`hobby` gives the message that these instances are not for production instances better compared to `burstable`. Currently functionality stays same, but I'm planning to remove HA functionality from burstable instances eventually.


byucesoy:
> My only feedback would be that considering this is a breaking change in terms of API/CLI, should we keep both options for a transition period?

We already do.

PR #4740, opened 2026-02-09
Refactor Nic locking for subnets while rekeying

furkansahin:
## Add advisory lock to prevent concurrent NIC rekeying
Multiple subnet strands in a connected mesh can enter refresh_keys
simultaneously. The incr_lock semaphores set on NICs during this
label are not visible to other strands until the transaction commits
at the end of the label, so the existing lock_set? guard does not
prevent the race.

Acquire pg_try_advisory_xact_lock at the top of refresh_keys. The
lock is transaction-scoped and released automatically when the label
finishes. By that point the NIC lock semaphores have been committed,
so subsequent strands will see them and nap via the existing check.

## Add connected_leader_id to PrivateSubnet
In a mesh of transitively connected subnets, the subnet with the
smallest ID is the "connected leader." This will be used to ensure
that only one subnet coordinates rekeying for the entire mesh,
rather than every subnet independently attempting to rekey the same
set of NICs.

The method reuses the same recursive CTE pattern as
find_all_connected_nics. A standalone subnet with no connections is
its own leader.

## Use connected leader to coordinate rekeying across subnets
Every subnet in a connected mesh independently triggers rekeying for
all NICs in the mesh every 24 hours. With N subnets, this means the
same work is attempted N times.

Only the connected leader now checks the 24-hour periodic timer.
Non-leader subnets skip the check entirely. The advisory lock key is
also updated to use the leader's ID, so all subnets in a mesh
contend on the same lock regardless of which one enters refresh_keys.

wait_old_state_drop now updates last_rekey_at on every subnet whose
NICs were rekeyed, not just the coordinator. This prevents a new
leader from immediately re-triggering after a topology change.

## Forward refresh_keys semaphore to the connected leader
When a non-leader subnet receives a refresh_keys semaphore (e.g.
from connect_subnet or disconnect_subnet), it now consumes it and
forwards a new one to the connected leader instead of entering
refresh_keys itself. This completes the picture: a single subnet
is responsible for rekeying the entire mesh, regardless of which
subnet originally received the trigger.

## Defer subnet destroy while NIC rekey is in progress
If a subnet is destroyed mid-rekey, the locked NICs from other
subnets would never get cleaned up, permanently blocking future
rekeying. Skip the destroy hop in before_run while locked NICs
still exist, allowing the rekey to complete first.

### DIFF
--- a/model/metal/private_subnet.rb
+++ b/model/metal/private_subnet.rb
@@ -40,5 +40,18 @@ class PrivateSubnet < Sequel::Model
             .select(Sequel.case({subnet_id_1: :subnet_id_2}, :subnet_id_1, Sequel[:subnet][:id])),
           cycle: {columns: :id})
     end
+
+    def connected_leader_id
+      DB[:subnet]
+        .with_recursive(:subnet,
+          this.select(:id),
+          DB[:connected_subnet]
+            .join(:subnet, {id: [:subnet_id_1, :subnet_id_2]})
+            .select(Sequel.case({subnet_id_1: :subnet_id_2}, :subnet_id_1, Sequel[:subnet][:id])),
+          cycle: {columns: :id})
+        .exclude(:is_cycle)
+        .order(:id)
+        .get(:id)
+    end
   end
 end
--- a/prog/vnet/metal/subnet_nexus.rb
+++ b/prog/vnet/metal/subnet_nexus.rb
@@ -3,6 +3,10 @@
 class Prog::Vnet::Metal::SubnetNexus < Prog::Base
   subject_is :private_subnet
 
+  def before_run
+    super unless destroy_set? && !get_locked_nics_dataset.empty?
+  end
+
   def connected_leader
     @connected_leader ||= PrivateSubnet[private_subnet.connected_leader_id]
   end
--- a/prog/vnet/metal/subnet_nexus.rb
+++ b/prog/vnet/metal/subnet_nexus.rb
@@ -17,8 +17,13 @@ class Prog::Vnet::Metal::SubnetNexus < Prog::Base
 
   label def wait
     when_refresh_keys_set? do
-      private_subnet.update(state: "refreshing_keys")
       decr_refresh_keys
+      unless connected_leader?
+        connected_leader.incr_refresh_keys
+        nap 0
+      end
+
+      private_subnet.update(state: "refreshing_keys")
       hop_refresh_keys
     end
 
--- a/prog/vnet/metal/subnet_nexus.rb
+++ b/prog/vnet/metal/subnet_nexus.rb
@@ -3,6 +3,14 @@
 class Prog::Vnet::Metal::SubnetNexus < Prog::Base
   subject_is :private_subnet
 
+  def connected_leader
+    @connected_leader ||= PrivateSubnet[private_subnet.connected_leader_id]
+  end
+
+  def connected_leader?
+    connected_leader.id == private_subnet.id
+  end
+
   label def start
     hop_wait
   end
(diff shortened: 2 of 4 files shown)


model/metal/private_subnet.rb
```diff
@@ -40,5 +40,19 @@ def find_all_connected_nics
...
+        .with_recursive(:subnet,
+          this.select(:id),
+          DB[:connected_subnet]
+            .join(:subnet, {id: [:subnet_id_1, :subnet_id_2]})
+            .select(Sequel.case({subnet_id_1: :subnet_id_2}, :subnet_id_1, Sequel[:subnet][:id])),
+          cycle: {columns: :id})
+        .exclude(:is_cycle)
+        .select(:id)
+        .order(:id)
+        .single_value
```
jeremyevans:
```suggestion
        .order(:id)
        .get(:id)
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -39,6 +47,11 @@ def gen_reqid
   end
 
   label def refresh_keys
+    unless connected_leader?
+      Clog.emit("Not the connected leader, should not have been called", {not_connected_leader: {private_subnet:, connected_leader:}})
```
jeremyevans: This doesn't necessarily indicate a bug, it could be due to a race condition (subnet connected between hop from `wait` to `refresh_keys`). Maybe reword the log message appropriately (e.g. "No longer the connected leader").

spec/prog/vnet/metal/subnet_nexus_spec.rb
```diff
@@ -89,6 +89,13 @@
       expect(vm.reload.update_firewall_rules_set?).to be true
     end
 
+    it "does not check periodic rekey when not the connected leader" do
+      ps.update(last_rekey_at: Time.now - 60 * 60 * 24 - 1)
+      expect(nx).to receive(:connected_leader?).and_return(false)
```
jeremyevans: We should avoid mocking prog methods like this. Create a new private subnet with a low id and connect it to `ps`. That way, it will be the connected leader. Same for the "hops to wait if not the connected leader" spec.

jeremyevans (approved):
The change makes sense to me. For something this sensitive, probably good to get additional approvals.

fdr:
wow, tricky one, but kudos for carving out time to take a near miss and turn it into a fix. I won't be able to look into this until tomorrow, though.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -3,14 +3,31 @@
 class Prog::Vnet::Metal::SubnetNexus < Prog::Base
   subject_is :private_subnet
 
+  def before_run
+    super unless destroy_set? && get_locked_nics.any?
```
enescakir:
```suggestion
    super unless destroy_set? && !get_locked_nics_dataset.empty?
```

enescakir (approved):

fdr (commented):
I find the topological sort clever to find a leader, though I haven't been able to work through concurrent  modifications in mesh topology: what's the thinking on how this system is robust to that? For examples: disconnecting, or connecting in a way that changes the leader.

furkansahin:
Let's name the action disconnection or connecting: `disruption`.
- If the `disruption` happens in the middle of rekeying, nothing happens, rekeying completes. Because, the new leader wants to start rekeying but the existing nics already have the `lockup` semaphore incremented. 
- If the leader gets an `incr_destroy` in the middle of rekeying, nothing happens, rekeying completes and then it is destroyed.
- If the `disruption` happens before we start rekeying, we do not switch from `wait` to `refresh_keys` for the old leader, even if we did, we get back out immediately from `refresh_keys` to `wait`.
- If the `disruption` happens after a rekeying completes, since this means there is a new connected subnet and we have to go through a full rekeying cycle, new leader handles that.

mohi-kalantari (approved):

fdr (commented):
so there is almost certainly a mutex bug in this but I need to clean up my proof draft branch before @furkansahin can reasonably analyze it. hopefully with the benefit of TLA it'll be among the last times...

PR #4741, opened 2026-02-09
Persist firewall rules on Kubernetes nodes across reboots

mohi-kalantari:
Replaces the previous approach of applying iptables and nft rules at runtime with writing a complete nftables ruleset to /etc/nftables.conf and enabling the nftables systemd service. This ensures NAT masquerading and IPv6 pod access control rules survive node reboots.

The ruleset includes a flush to start clean, an ip nat table for POSTROUTING masquerade, and an ip6 pod_access table for ingress and egress filtering.


fdr (approved):
Watch for abnormalities and be mindful with your deployment given our recent failures to mitigate bugs even in pretty good code, but we gotta do something like this… we cannot stay as we are.

PR #4745, opened 2026-02-09
Update Firewall Rule Web UI

jeremyevans:
Due to limited horizontal space, instead of trying to edit multiple
rows inside a table, switch to a display only table for the rules,
and link to a separate page to add/edit a rule. Make adding/editing a
rule friendly, with drop downs for common ports and IP address ranges
(both overall and for private subnets). Still allow full customization
of both the port ranges and IP address range.

In the able showing the rules, make displayed port ranges nicer,
using names when possible and only showing a single port if the range
only includes a single port.  Similarly, make displayed cidr's nicer,
using names where appropriate.

One behavioral change is when adding a firewall rule for a private
subnet, you pick whether to add the IPv4 or IPv6 range, where
previously, it added both. This change gives the user more control, and
it was necessary to ensure consistency between the add firewall rule
form and the edit firewall rule form (which use the same template).

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_firewall_default-eu-central-h1-default_firewall-rule
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_firewall_default-eu-central-h1-default_firewall-rule_fr78qd3dc95y0n79xma5fsbhfq
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_firewall_default-eu-central-h1-default_firewall-rule_fr33d29s23h65k5d8fh0hz9gw3
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_firewall_default-eu-central-h1-default_firewall-rule_fr7bxg52em967jv88bfnfk5v1m
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_location_eu-central-h1_firewall_default-eu-central-h1-default_networking

### DIFF
--- a/views/networking/firewall/firewall_rule.erb
+++ b/views/networking/firewall/firewall_rule.erb
@@ -105,9 +105,6 @@
               <ul class="ml-6 mb-2 list-disc">
                 <li>To allow access from a single IP address, you can omit the netmask (e.g. use "1.2.3.4" instead of
                   "1.2.3.4/32").</li>
-                <li>To allow access from all IP addresses, use "0.0.0.0/0" for IPv4, or "::/0" for IPv6.</li>
-                <li>To allow access to a single port, you can enter only the port number (e.g. use "22" instead of
-                  "22..22").</li>
                 <li>You can optionally add a description for the rule, to help you remember why the access has been
                   allowed.</li>
               </ul>


enescakir (commented):
I’m not sure adding a long explanatory text actually improves the UX. My feeling is that most users won’t read it, and it makes the UI look cluttered for users who already know what to do. If we really want to include it, IMHO (and I’m not a UI/UX expert), it should at least be collapsible and only shown when the user explicitly clicks to see it.

I tried to improve the UX of this table. Again, I’m not a UI/UX expert, but this felt more intuitive to me:
- The current Submit button is generally perceived as “enter edit mode.” A Save button is more commonly associated with persisting changes.
- Using smaller, horizontally aligned buttons can save space and make the UI feel less cluttered.
- Accepting port ranges in a single input forces us to teach the `...` delimiter. Instead, we can use separate inputs and make the end port optional.
- We can clarify that the source IP accepts both a single IP and CIDR by stating this directly in the placeholder.
- I added a subtle background color to the creation row to visually distinguish it from existing rows.
- Since private subnet selection is an alternative to source IP, I placed them in the same column.

Image

enescakir:
Also, DigitalOcean provides predefined source and port options like `All IPv4`, `All IPv6`, `SSH`, `HTTP`, and `HTTPS`. These make firewall rule creation much easier for users who aren’t familiar with networking. I think adding a few predefined options to our UI could significantly improve usability as well.

CleanShot 2026-02-11 at 14 26 29@2x

jeremyevans:
While I like the look of Enes's example, it only works for large screen widths. At the screen width I typically view at (1263px according to Chrome dev tools), it would be unusable. This is due to our multiple levels of vertical tab navigation, which reduces the total horizontal space for the table to 565px wide in my view.

Unless we want to change our overall visual design, we cannot make the table more complex and have it still be usable at moderate screen widths. Even with the current design, if there is an error submitting the form, the error messages make the form quite hard to use due to the limited horizontal space.

I think we either have to change our overall visual design, or we have to move the firewall rule UI so we display the rules in a table, but to modify them, we link to another page specific to that rule. That page can use some of the recommendations here:

* IP dropdown (with custom)
  * Maybe combine private subnet dropdown into this using optgroup? 
* Port dropdown (with custom)
  * Separate port/start end for custom

This page would use a vertical form, to avoid the horiztonal space issues.

Maybe we can include the explanatory text after the form instead of before, so it doesn't get in the way of people who already understand the form. We could use summary/details to hide it by default if that still seems like too much. We could also eliminate it entirely, though I wouldn't recommend that.

Note that we don't currently support separate TCP/UDP rules or ICMP rules.

jeremyevans:
@enescakir @aamederen I updated the visual design in this PR to incorporate the feedback you provided.  Please let me know what you think.

views/networking/firewall/firewall_rule.erb
```diff
@@ -0,0 +1,122 @@
...
+    <div class="min-w-0 flex-1">
+      <h3 class="text-2xl font-bold leading-7 text-gray-900 sm:truncate sm:text-2xl sm:tracking-tight">
+        <%= page_heading %>
+      </h3>
+    </div>
+  </div>
+  <div>
+    <% form(action:, method: :post, class: "gap-6", id: "firewall-rule-form") do %>
+      <div class="mb-2">
+        <div id="port-type">
```
aamederen: Maybe it's just my assets...

When Port is selected (like SSH) I think 
- we should disable and Start Port and End Port inputs
- or auto-fill them with 22 and auto-change the "Port" depending on the customer input in these fields

Similarly, if the private subnet is selected, customer CIDR input should be disabled

Image

The resulting FW rule, which I think is expected from the code, but could be confusing to users.

```ruby
FirewallRule["frejqkzentg61jkaak3mrddtr2"] {
             :id => "74af3fba-ba81-81f8-94d4-a98e986b7581",
           :cidr => #<NetAddr::IPv4Net:0x00007f975bd29868 @m32=#<NetAddr::Mask32:0x00007f975bd299f8 @prefix_len=16, @mask=4294901760>, @base=#<NetAddr::IPv4:0x00007f975bd29840 @addr=168689664>>,
     :port_range => #<Sequel::Postgres::PGRange [22,23)::int4range>,
    :firewall_id => "f20f0f70-aba1-85fc-b006-8c08cbf80f80",
    :description => "something something",
       :protocol => "tcp"
}
```

model/firewall_rule.rb
```diff
@@ -10,7 +10,85 @@ def ip6?
...
+
+  def <=>(other)
+    (cidr.version <=> other.cidr.version).nonzero? ||
+      (cidr.network.addr <=> other.cidr.network.addr).nonzero? ||
+      (cidr.netmask.mask <=> other.cidr.netmask.mask).nonzero? ||
+      (port_range.begin <=> other.port_range.begin).nonzero? ||
+      port_range.end <=> other.port_range.end
+  end
+
+  DISPLAY_PORT_NAMES = {
```
aamederen, line 42: nit, fwiw, There are multiple constants defined here and it's confusing to get the grasp

```ruby
DISPLAY_PORT_NAMES
PORT_TYPES 
PORT_RANGES 
PORT_OPTIONS 
DISPLAY_SOURCE_TYPE 
SOURCE_TYPES 
SOURCE_CIDRS 
SOURCE_OPTIONS
```

I don't have a suggestion here, but wanted to note anyways. No immediate action required.

assets/css/tailwind.css
```diff
@@ -47,3 +47,18 @@
     display: none;
   }
 }
+
+#firewall-rule-form {
```
aamederen, line 51: I needed to ask AI how these conditional selectors work, but big +1 for handling it without JS

model/firewall_rule.rb
jeremyevans: I agree with you here. I'm not sure I can think up better names, but these should at least have comments explaining what each is used for. I'll work on that.

jeremyevans: I rebased this on main and added comments for all of those constants.

views/networking/firewall/firewall_rule.erb
```diff
@@ -0,0 +1,122 @@
...
+          <div>
+            <div class="mt-2 text-sm text-gray-500">
+              <p class="mb-2">Firewall rules allow connections to specific IP ports from specific IP address ranges (CIDRs). The
+                default behavior for a new connection without an applicable firewall rule is to disallow access.</p>
+              <p class="mb-2">To allow access, create a firewall rule for the source IP address range that will be connecting and the
+                desination port range that will be connected to.</p>
+              <ul class="ml-6 mb-2 list-disc">
+                <li>To allow access from a single IP address, you can omit the netmask (e.g. use "1.2.3.4" instead of
+                  "1.2.3.4/32").</li>
+                <li>To allow access from all IP addresses, use "0.0.0.0/0" for IPv4, or "::/0" for IPv6.</li>
```
enescakir: Do we need that, since they already have predefined sources?

views/networking/firewall/firewall_rule.erb
```diff
@@ -0,0 +1,122 @@
...
+                <li>To allow access to a single port, you can enter only the port number (e.g. use "22" instead of
```
enescakir: Since we get the start and end ports separately, I think this explanation is no longer relevant.

views/networking/firewall/firewall_rule.erb
jeremyevans: Correct, this should be removed after the UI changes. I pushed a commit to remove this line and the line below.

PR #4746, opened 2026-02-10
Make PostgreSQL overview page handle missing representative server

jeremyevans:
Previously, if the overview page was accessed without a representative server, a NoMethodError was raised. We have had at least 2 production exceptions due to this issue. This takes information from the resource if the representative server is not available. Approach based on @byucesoy's comment: https://github.com/ubicloud/ubicloud/pull/4649#discussion_r2721628295

While here, simplify disk_usage_subtext setting (unset local variables in scope default to nil).

Second commit removes some unnecessary lines in the postgres web specs.

### DIFF
--- a/spec/routes/web/project/postgres_spec.rb
+++ b/spec/routes/web/project/postgres_spec.rb
@@ -94,7 +94,6 @@ RSpec.describe Clover, "postgres" do
       end
 
       it "can list PostgreSQL databases with parents" do
-        pg
         pg.update(parent_id: pg_wo_permission.id)
         visit "#{project.path}/postgres"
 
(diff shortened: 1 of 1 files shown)


furkansahin (approved):

byucesoy:
I have a development branch on my local where I ensure that representative_server is never nil, so that we wouldn't need to handle such cases. Can we wait for few more days :slightly_smiling_face: ?

jeremyevans:
Sure, waiting seems fine. If there is a way to ensure `representative_server` is never nil, I agree that is best.

byucesoy:
[Here](https://github.com/ubicloud/ubicloud/pull/4761) is my PR to ensure that `representative_server` is never nil

jeremyevans:
Closing as we made changes so that `representative_server` is not `nil`.

PR #4750, opened 2026-02-10
Allow replication connection via postgres user

byucesoy:
This allows users to replicate data with postgres user.


furkansahin (approved):
Can you add to the commit why this is needed?

PR #4753, opened 2026-02-10
Fix pg_hba.conf rule order for standby cert auth

Amogh-Bharadwaj:
Standby servers connecting to the primary via the ubi_replication role were failing with "password not supplied" errors in environments with private DNS configuration. This occurred because pg_hba.conf processes rules sequentially, using the first match.

The private subnet rules (requiring SCRAM-SHA-256 password auth) were positioned before the HA standby rules (using certificate auth). When standbys connected from within the private subnet, they matched the password-based rule first, even though they were presenting certificates rather than passwords.

Reorder the rules to place certificate-based HA standby authentication before general private subnet password authentication. This ensures standbys can authenticate using their certificates regardless of whether they connect via IP (local) or hostname (DNS-enabled environments). This is backwards compatible

### DIFF
--- a/rhizome/postgres/bin/configure
+++ b/rhizome/postgres/bin/configure
@@ -77,9 +77,6 @@ local   replication     all                                     peer
 host    replication     all             127.0.0.1/32            scram-sha-256
 host    replication     all             ::1/128                 scram-sha-256
 
-# Allow connections from private subnet with SCRAM authentication
-#{private_subnets}
-
 # Allow replication connection using special replication user for
 # HA standbys
 hostssl replication     ubi_replication all                     cert map=standby2replication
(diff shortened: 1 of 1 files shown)


serprex:
missed by #4526 since this doesn't come up when not using dns zone

byucesoy:
I'm OK with this change, with the principle of putting more specific rules towards to the top.

However, I'd like to understand the situation better because I got confused by the explanation. HA standbys would connect to `replication` "database", which is not covered by `all`, so why would they match to private subnet rules?

Also HA doesn't depend on DNS because we configure /etc/hosts on standby servers, so they can resolve the hostnames even if there is no DNS set up.

serprex:
without dns_zone we connect over public ip, avoiding this issue

byucesoy (approved):

PR #4754, opened 2026-02-10
Refactor storage auto scaling methods for easier extension

jeremyevans:
Extract the following private methods:

* handle_storage_auto_scale_action(disk_usage_percent)
* cancel_storage_auto_scale_action(disk_usage_percent)
* perform_storage_auto_scale_action(disk_usage_percent)

These allow extension at different level. To keep the 77/80/85/90 percentage handling, you can override {cancel,perform}_storage_auto_scale_action with custom behavior. To use custom percentages, you can override handle_storage_auto_scale_action.

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -288,7 +288,7 @@ class PostgresResource < Sequel::Model
     }
 
     if disk_usage_percent < 90
-      send_storage_auto_scale_warning_email(disk_usage_percent, next_option, extra_email_content)
+      send_storage_auto_scale_warning_notification(disk_usage_percent, next_option, extra_email_content)
     else
       if next_option
         target_vm_size = next_option["size"]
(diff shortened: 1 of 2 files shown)


model/postgres/postgres_resource.rb
```diff
@@ -249,16 +249,17 @@ def handle_storage_auto_scale
...
     [90, 85, 80].each {
       send("decr_storage_auto_scale_action_performed_#{it}") if disk_usage_percent <= it - 3
     }
 
     if disk_usage_percent <= 77
-      Page.from_tag_parts("PGStorageAutoScaleMaxSize", id)&.incr_resolve
-      Page.from_tag_parts("PGStorageAutoScaleQuotaInsufficient", id)&.incr_resolve
-      Page.from_tag_parts("PGStorageAutoScaleCanceled", id)&.incr_resolve
-      decr_storage_auto_scale_canceled
+      cancel_storage_auto_scale_action(disk_usage_percent)
```
byucesoy: This is not actually cancelling the storage auto scale, it just resolves the pages. 

We have customer initiated cancel, which is already separate function called `cancel_storage_auto_scale`.

I think we should keep this part inlined.

jeremyevans:
Since @serprex appears OK with your proposal of just renaming the existing methods, I'll switch this to do that.

PR #4755, opened 2026-02-10
Wait for 5% drop to decrement auto scale semaphores

byucesoy:
It seems 3% oscilation is not uncommon as I'd expect. For now, I'm increasing the hysteresis to 5% to avoid spurious emails. If this is not sufficient, we can decide to not decrement the semaphores at all.


jeremyevans (approved):
This is an improvement. Let's put it in and see what things look like. We can always take further action later.

PR #4756, opened 2026-02-10
Add Overrider

jeremyevans:
Overrider is designed to make it easier for people to build on top of Ubicloud and avoid merge conflicts.

Without this, typical behavior for people building on top of Ubicloud would be to modify the code they want to change directly. However, that will often result in merge conflicts later if we modify the same code location.

With Overrider, the overrides are placed in files in a separate directory, and are prepended to the classes to override the default behavior (can call super to get the default behavior).

Overrider is designed to handle only classes/modules defined inside the repository. It ignores other classes/modules as well as any anonymous classes/modules. Loading it into a superclass before subclassing will automatically load overrides for subclasses when those subclasses are created. This allows for overrides to work properly when using incremental testing.

Ubicloud doesn't use Overrider itself except during unfrozen testing, where it uses it for all models and progs (though only a single model is overridden with some test methods).

### DIFF
--- a/model.rb
+++ b/model.rb
@@ -3,7 +3,11 @@
 require_relative "db"
 require "sequel/model"
 
-Overrider.setup_overrides(Sequel::Model, "spec/overrider_test") if Config.unfrozen_test?
+if Config.override_dir
+  Overrider.setup_overrides(Sequel::Model, Config.override_dir)
+elsif Config.unfrozen_test?
+  Overrider.setup_overrides(Sequel::Model, "spec/overrider_test")
+end
 
 if ENV["RACK_ENV"] == "development"
   Sequel::Model.cache_associations = false
--- a/config.rb
+++ b/config.rb
@@ -66,6 +66,7 @@ module Config
   optional :operator_ssh_public_keys, string
   override :staging, false, bool
   override :clover_freeze, false, bool
+  optional :override_dir, string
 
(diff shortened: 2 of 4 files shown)


model.rb
```diff
@@ -3,6 +3,8 @@
 require_relative "db"
 require "sequel/model"
 
+Overrider.setup_overrides(Sequel::Model, "spec/overrider_test") if Config.unfrozen_test?
```
serprex: will fork need to add override call? seems like this should include a setup_overrides call to keep out fork modifying model.rb or base.rb

jeremyevans: A fork that wants to use this would add the appropriate `setup_overrides` calls. I suppose we could add an environment variable (`Config` setting) for the override directory, so the code would be present in the main branch, but only run if the environment variable is set. Would you prefer that?

serprex: Yes

jeremyevans: I added that to the PR, please let me know if that will work for you.

jeremyevans:
> I wish there was a bit of example in the commit message that showed the intended exercise of this design (e.g. showing running some `tree` output with an override in place, plus `cat` on that file), but I can kind of get that from the test.
> 
> In context and with a close read we know the suggested usage, but the purely narrative description lends more doubt than necessary.

I added an example to the commit message for the first commit.

PR #4758, opened 2026-02-11
Fix inconsistent utilization scoring between slice and non-slice hosts

furkansahin:
When a host accepts slices, VmHostSliceAllocation replaced the two vm_host_allocations (CPU, Memory) with a single wrapper in the @allocations array. This reduced the array from 3 items to 2, which silently changed how calculate_score weighted resources:
```ruby
  # Regular host: @allocations = [CpuAlloc, MemAlloc, StorageAlloc]
  util = [cpu, mem, storage]
  avg  = (cpu + mem + storage) / 3     # each resource: 33% weight
  imbalance = util.max - util.min      # sees CPU vs Mem vs Storage

  # Slice host: @allocations = [SliceAlloc, StorageAlloc]
  util = [slice_combined, storage]
  avg  = (slice_combined + storage) / 2  # storage: 50% weight
  imbalance = util.max - util.min        # cannot see CPU vs Mem spread
```
Concrete example with cpu=0.3, mem=0.7, storage=0.5, target=0.65:
```
  Regular path (3 items):
    avg       = (0.3 + 0.7 + 0.5) / 3 = 0.50
    deviation = 0.65 - 0.50 = 0.15
    imbalance = 0.7 - 0.3 = 0.4
    score     = 0.15 + 0.4 = 0.55

  Slice path (2 items, combined = (0.3+0.7)/2 = 0.5):
    avg       = (0.5 + 0.5) / 2 = 0.50
    deviation = 0.65 - 0.50 = 0.15
    imbalance = 0.5 - 0.5 = 0.0    ← imbalance hidden
    score     = 0.15 + 0.0 = 0.15
```
Both host types compete for standard VMs (no accepts_slices filter in candidate_hosts for non-burstable requests), so the 0.40 score difference causes slice hosts to be systematically preferred even when they have worse resource balance.

The fix stores VmHostSliceAllocation separately as @slice_allocation instead of replacing @vm_host_allocations. The @allocations array always contains the flat [CpuAlloc, MemAlloc, StorageAlloc, ...] list, ensuring identical scoring formulas. The slice_allocation is used for is_valid and update delegation only.

For the existing-slice case (burstable reuse), calculate_score sets CPU and memory utilization to target_host_utilization so these hosts remain preferred, matching the prior behavior without collapsing the allocation count.

### DIFF
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -286,8 +286,7 @@ module Scheduling::Allocator
       @device_allocations << GpuAllocation.new(candidate_host, request) if request.gpu_count > 0
 
       if request.use_slices && candidate_host[:accepts_slices]
-        # Wrap around and replace the host allocations. That way we can control that logic from the slice POV
-        @vm_host_allocations = [VmHostSliceAllocation.new(candidate_host, request, @vm_host_allocations)]
+        @slice_allocation = VmHostSliceAllocation.new(candidate_host, request, @vm_host_allocations)
       end
 
       @allocations = @vm_host_allocations + @device_allocations
@@ -295,14 +294,24 @@ module Scheduling::Allocator
     end
 
     def is_valid
-      @allocations.all? { it.is_valid }
+      if @slice_allocation&.existing_slice?
+        @device_allocations.all? { it.is_valid }
+      elsif @slice_allocation
+        @slice_allocation.is_valid && @device_allocations.all? { it.is_valid }
+      else
+        @allocations.all? { it.is_valid }
+      end
     end
 
     def update(vm)
       vm_host = VmHost[@candidate_host[:vm_host_id]]
       DB.transaction do
         Allocation.update_vm(vm_host, vm)
-        @vm_host_allocations.each { it.update(vm, vm_host) }
+        if @slice_allocation
+          @slice_allocation.update(vm, vm_host)
+        else
+          @vm_host_allocations.each { it.update(vm, vm_host) }
+        end
         @device_allocations.each { it.update(vm, vm_host) }
       end
     end
@@ -314,7 +323,11 @@ module Scheduling::Allocator
     private
 
     def calculate_score
-      util = @allocations.map { it.utilization }
+      util = if @slice_allocation&.existing_slice?
+        @vm_host_allocations.map { @request.target_host_utilization } + @device_allocations.map { it.utilization }
+      else
+        @allocations.map { it.utilization }
+      end
 
       # utilization score, in range [0, 2]
       score = @request.target_host_utilization - util.sum.fdiv(util.size)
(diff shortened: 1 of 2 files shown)


bsatzger:
The way we define our VM sizes should prevent that CPU utilization is high and memory utilization is low, or the other way around.

I assume we use `VmHostSliceAllocation` in the majority of allocations (not only for burstables), given that we set `use_slices` to true
https://github.com/ubicloud/ubicloud/blob/2c0e82eb1390ec846ad39eaaff31d714f12e17ba/scheduling/allocator.rb#L26 and the majority of our hosts accepts slices.  

A reason why burstable VMs might be packed more densely could be how we calculate utilization for slice allocations at
https://github.com/ubicloud/ubicloud/blob/2c0e82eb1390ec846ad39eaaff31d714f12e17ba/scheduling/allocator.rb#L416

These are the steps for burstables:
For burstables (where `@request.require_shared_slice` is `true`), we try to find an existing slice that can hold the vm (see `select_existing_slice`), and store it in `@existing_slice`.

When asking for `utilization`, if `@existing_slice` is set, we return the desired (optimal) utilization, irrespective of the true utilization.

That means, for burstables we effectively try to find some slice with space and neglect the host's utilization.

The comment in `def utilization` suggests that that was a deliberate decision.

pykello:
I don't have objections to the main idea, but the way I read Ben's comment is that the imbalance factor isn't very effective in our prod environment. Please correct me if I misunderstood.

I'll wait for @furkansahin 's opinion on @bsatzger 's comment before approving.

bsatzger:
My understanding is that the primary motivation for this PR is the observation that burstable VMs are being packed more densely than regular VMs on hosts.

However, this behavior appears to be intentional and originates from the design implemented here:

```rb
    def utilization
      if @existing_slice
        # if we found an existing slice, return the desired utilization
        # to make this a preferred choice
        @request.target_host_utilization
      else
        # otherwise, compute the score based on combined CPU and Memory utilization, as usual
        util = @vm_host_allocations.map(&:utilization)
        util.sum.fdiv(util.size)
      end
    end
```

`@existing_slice` is set only for burstable VMs. When an existing slice is found for a burstable VM, we deliberately return the target host utilization to make that host more attractive, regardless of its current utilization. This effectively encourages denser packing for burstable VMs by design.

Most regular and burstable VMs go through the allocator logic that this PR modifies. As a result, the proposed changes would also alter the behavior for regular VMs - even though we do not currently observe an issue there. Meanwhile, the denser packing of burstable VMs likely stems from the intentional logic shown above.

Given that, it seems this PR isn't addressing the root cause, while also impacting allocation behavior more broadly. It may be worth reassessing whether this PR should be merged.

furkansahin:
this was a rush pr fully left to the AI gods, after a meeting with @bsatzger and now @pykello's comment, I decided to close this.

PR #4760, opened 2026-02-11
Prevent immediate reconvergence due to pruning servers

byucesoy:
At the prune_servers label, we increment destroy semaphore for servers to be destroyed, but we don't wait for them to be actually destroyed. Then PostgresResource detects extra servers and realizes it is not converged (having more servers than desired), and starts reconvergence immediately. This commit adds a new label wait_prune_servers after prune_servers to wait until servers to be destroyed are actually destroyed.

### DIFF
--- a/prog/postgres/converge_postgres_resource.rb
+++ b/prog/postgres/converge_postgres_resource.rb
@@ -124,10 +124,18 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
       .reject { it.representative_at || it.needs_recycling? || it.version != postgres_resource.target_version }
       .sort_by { [(it.strand.label == "wait") ? 0 : 1, Time.now - it.created_at] }
       .take(postgres_resource.target_standby_count) + [postgres_resource.representative_server]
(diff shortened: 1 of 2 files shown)


prog/postgres/converge_postgres_resource.rb
```diff
@@ -124,11 +124,20 @@ class Prog::Postgres::ConvergePostgresResource < Prog::Base
...
 
     servers_to_keep.each(&:incr_configure)
     postgres_resource.incr_update_billing_records
 
+    hop_wait_prune_servers
+  end
+
+  label def wait_prune_servers
+    servers_to_destroy = frame["servers_to_destroy"]
+    nap 30 if servers_to_destroy.any? { PostgresServer[it] }
```
enescakir: You can send single DB request
```suggestion
    nap 30 unless PostgresServer.where(id: frame["servers_to_destroy"]).empty?
```

furkansahin (approved):
nothing to add on top of Enes' suggestion

PR #4761, opened 2026-02-11
Ensure representative_server is never nil

byucesoy:
representative_server of a resource is defined as the server which has
its representative_at value set. The ideal place to be would be having
one and only one representative_server for a resource for the entire
lifetime of the resource. "Only one" part is guaranteed by the UNIQUE
constraint on the database level. "One" part is not guaranteed, because
representative_server can be nil during failovers.

To prevent representative_server from being nil, we need to hand over
being representative_server in a single transaction. In the current form
this is not possible, because destruction of old primary and promotion
of new primary is done by separate strands.

To bring both these operations into the same transaction, we need to
decouple representative_server handover from the destruction of the old
primary. For this purpose, we are adding a new database column called
is_representative. Now during standby promotion, we can both unset the
is_representative of the old primary and set the is_representative of
new primary. Actual destruction of the old primary is done after that.

However, there is one important implication of keeping the old primary
around for a while. During failover, we need to make sure that the old
primary is fenced. This is not a problem for planned failovers, because
we can still access to the old primary and fence it from the inside of
the VM. For unplanned failovers, it is partly OK as well, because for
unplanned failovers, the expectation is that the old primary is already
down, so there is no need to fence. However, if the old primary is not
really down for example, due to a network partition or false positive,
then keeping it around for a while can cause split brain. To prevent
this, we still try to lockout the old primary but it is not guaranteed
to work.

That said, this is not a new defect, because even in the current form,
we don't wait underlying VM to be destroyed, so there is still risk of
split brain if the old primary is not really down. I'm working on a more
general fix for this problem. This change, doesn't change the state of
the defect, but gives us stable invariant that representative_server is
never nil.


migrate/20260216_drop_index_for_representative_at.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
```
jeremyevans, line 4: Issuing multiple statements in a `no_transaction` migration is risky. If any statement other than the first fails, you have a partially applied migration. Best to split into three migrations (you can include them all in a single commit).

migrate/20260210_add_representative_until_to_postgres_server.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  no_transaction
+
+  up do
+    alter_table(:postgres_server) do
+      add_column :representative_until, :timestamptz, null: true, default: nil
+      drop_index :resource_id, concurrently: true
+      add_index :resource_id, unique: true, where: Sequel.~(representative_at: nil) & {representative_until: nil}, concurrently: true
```
jeremyevans: This has a theoretical race condition, since the uniqueness is not enforced between the two statements. In cases like this, it's best to add the new unique index, and after that drop the old unique index. You can avoid duplicate index name issues by using a `:name` option when creating the new index.

FWIW, we don't have a general index on `resource_id`, which means code such as `pg.servers` likely does a sequential scan. Maybe not an issue now, but could become an issue as we scale. Probably a good idea to also add a non-partial index on `resource_id` to avoid that.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -667,10 +667,9 @@ def before_run
   end
 
   label def wait_representative_lockout
-    representative_server = postgres_server.resource.representative_server
-    nap 5 unless representative_server.nil? || representative_server.strand&.label == "destroy"
+    hop_taking_over if postgres_server.resource.representative_server.strand.label == "wait_locked_out"
 
-    hop_taking_over
+    nap 0
```
jeremyevans: `nap 0` can only be used safely if you have changed the state of the object during the label. Otherwise, it can result in spin-locking a respirate thread for 30 seconds. This should be at least `nap 1`, and probably longer depending on the expected timeframe for the current representative server strand to reach `wait_locked_out`.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -706,15 +709,16 @@ def before_run
   end
 
   label def wait_fencing_of_old_primary
-    nap 0 if postgres_server.resource.representative_server.strand.label != "wait_in_fence"
+    hop_taking_over if postgres_server.resource.representative_server.strand.label == "wait_in_fence"
 
-    postgres_server.resource.representative_server.incr_destroy
-    hop_taking_over
+    nap 0
```
jeremyevans: Same issue here, should probably switch to `nap 1`. I realize this was `nap 0` before, but that seems like a problem.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -723,7 +727,9 @@ def before_run
     case vm.sshable.d_check("promote_postgres")
     when "Succeeded"
       Page.from_tag_parts("PGPromotionFailed", postgres_server.id)&.incr_resolve
-      postgres_server.update(timeline_access: "push", representative_at: Time.now, synchronization_status: "ready")
+      postgres_server.resource.representative_server.update(representative_until: current_time)
```
jeremyevans: Can you explain why setting `representative_at: nil` here would not accomplish the same thing (always one representative server)?

There must be something I'm not understanding, because it seems like setting `representative_at: nil` would accomplish the same thing without introducing a new column and changing the definition of the unique index.

byucesoy: We don't want to lose the historical context on which servers were representative at a given time. It is quite useful for debugging and also while creating timeline of events.

It also more semantically correct that if a server was representative at some point, it should have non-nil representative_at.

jeremyevans: I think it would actually be better to have a `is_representative` column than a `representative_until` column. Since you expect any `representative_until` value to match one `representative_at` value, I don't think a separate `representative_until` timestamp column adds useful information over an `is_representative` column. Note that timestamp columns tend to be one of the more expensive column types to retrieve, since they must be parsed and converted to Ruby Time objects.

migrate/20260211_add_column_representative_until.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
```
jeremyevans: I think this can use `change` instead of separate `up`/`down`. Same for the 2nd and 4th migrations.

migrate/20260211_drop_old_postgres_server_representative_index.rb
```diff
@@ -0,0 +1,17 @@
...
+Sequel.migration do
+  no_transaction
+
+  up do
+    alter_table(:postgres_server) do
+      drop_index :resource_id, concurrently: true
+    end
+  end
+
+  down do
```
jeremyevans: I think this can use `reverse` instead of separate `up`/`down`.

jeremyevans (commented):
As mentioned in https://github.com/ubicloud/ubicloud/pull/4761#discussion_r2794577872, if we want to add a column, I don't think it needs to be a timestamp column, we could add a boolean column instead without losing information.

prog/postgres/postgres_server_nexus.rb
byucesoy: Yep, that works too. I'll make that change.

byucesoy: Actually, after making that change, `representative_at` column's purpose reduced only for debugging and keeping a database column just for debugging feels overkill. I think we can still gather the same debugging information from logs, albeit with more effort.

Anyway, I went with only is_representative column and dropped the others.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -692,12 +691,16 @@ def before_run
...
-    reap(:destroy, fallthrough: true, reaper:)
-    hop_destroy if strand.stack.first["lockout_succeeded"]
+    reap(:wait_locked_out, fallthrough: true, reaper:)
+    hop_wait_locked_out if strand.stack.first["lockout_succeeded"]
 
     nap 0.5
   end
 
+  label def wait_locked_out
+    nap 24 * 60 * 60
```
shikharbhardwaj, line 701: There might be some child strands remaining here, as the reap in the previous label has fallthrough enabled. In the previous version, there was a reap in `destroy` that handled them. I guess when removing this server, we have to update the label to destroy mannually?

shikharbhardwaj (approved):

prog/postgres/postgres_server_nexus.rb
byucesoy: When incr_destroy is called for this server by new primary during take over, it would switch to destroy label and there the child strands would get reaped.

PR #4763, opened 2026-02-11
Add Config.clover_admin_development_no_webauthn?

jeremyevans:
If this is set in development mode, do not require webauthn be setup. Note that even in this case, the developer has the option of setting up webauthn.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -132,6 +132,8 @@ class CloverAdmin < Roda
     classes.sort_by!(&:name)
   end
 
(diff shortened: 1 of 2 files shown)


serprex:
Additional config is good, if someone wants to run their dev like a prod environment that should be easily doable

jeremyevans:
> Do we really need an additional config value? Why not just disable it for development?

The general principle is the development environment should mirror the production environment as much as possible in terms of behavior. I would prefer not to introduce this at all. The only reason it is being added is because there are developers working on Ubicloud that do not have webauthn security keys. Those are the only developers that should use this feature.

PR #4764, opened 2026-02-11
Tune Postgres backlog pages

shikharbhardwaj:


model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,70 @@ def refresh_walg_credentials
   end
 
   def observe_archival_backlog(session)
+    result = session[:ssh_session].exec!(
+      "sudo find /dat/:version/data/pg_wal/archive_status -name '*.ready' -printf '%f\\n' | sort | head -1",
+      version:
+    )
+    oldest_pending = result.strip
```
enescakir: nit
```suggestion
    oldest_pending = session[:ssh_session].exec!(
      "sudo find /dat/:version/data/pg_wal/archive_status -name '*.ready' -printf '%f\\n' | sort | head -1",
      version:
    ).strip
```

or


```suggestion
    oldest_pending = session[:ssh_session].exec!(
      "sudo find /dat/:version/data/pg_wal/archive_status -name '*.ready' -printf '%f\\n' | sort | head -1",
      version:
    )
    oldest_pending.strip!
```

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,70 @@ def refresh_walg_credentials
...
   rescue => ex
     Clog.emit("Failed to observe archival backlog", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
   end
 
+  def should_page_for_archival_backlog?(archival_backlog, oldest_pending, previous_oldest, now, previous_time)
+    return true if archival_backlog > archival_backlog_threshold * 2
+    return true unless previous_oldest && oldest_pending && previous_time
+
+    upload_rate = estimate_archival_rate(previous_oldest, oldest_pending, now, previous_time)
+    upload_rate < MIN_ARCHIVAL_RATE_BYTES_PER_SEC
```
enescakir:
```suggestion
    estimate_archival_rate(previous_oldest, oldest_pending, now, previous_time) < MIN_ARCHIVAL_RATE_BYTES_PER_SEC
```

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,70 @@ def refresh_walg_credentials
   end
 
   def observe_archival_backlog(session)
+    result = session[:ssh_session].exec!(
+      "sudo find /dat/:version/data/pg_wal/archive_status -name '*.ready' -printf '%f\\n' | sort | head -1",
```
fdr: perhaps as long as you are forking `sudo` you could throw in a `-u` to not run as root. Marginal, but...

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,70 @@ def refresh_walg_credentials
...
+    bytes_archived / elapsed_seconds.to_f
+  end
+
+  def wal2lsn(filename)
+    # WAL filename format: TTTTTTTTXXXXXXXXYYYYYYYY.ready
+    # TTTTTTTT = timeline (ignored for LSN)
+    # XXXXXXXX = high 32 bits
+    # YYYYYYYY = segment number (multiply by 16MB segment size for byte offset)
+    name = filename.delete_suffix(".ready")
+    return "0/0" unless name.length == 24
```
fdr: huh, do you really want a fake one if it's generically wrong? I can see the appeal of avoiding nil, but what's the thinking there?

fdr (commented):
I think both commit messages could use expansion. That the first one could use expansion  is a bit obvious, the second more minor: there's a lot of constants in both, they presumably got guessed as they are for some reasons (which you may as well write down in there).

But I got the gist of both of them. The thinking behind the constants fits in the "why?"  rubric

model/postgres/postgres_server.rb
shikharbhardwaj: Well, mainly because the rest of the routine would still have value (eventually raise a page) even if the archival rate estimation fails, and moving this to the caller was a bit verbose.

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,66 @@ def refresh_walg_credentials
...
+    oldest_pending = nil if oldest_pending.empty?
+
     result = session[:ssh_session].exec!(
       "sudo find /dat/:version/data/pg_wal/archive_status -name '*.ready' | wc -l",
       version:
     )
     archival_backlog = Integer(result.strip, 10)
 
+    now = Time.now
+    previous_oldest = session[:last_oldest_pending_wal]
```
byucesoy: last as an adjective is ambigious. `previous`or `prev` is better.

`session[:last_oldest_pending_wal]` -> `session[:previous_oldest_pending_wal]`

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,66 @@ def refresh_walg_credentials
...
     elsif archival_backlog < archival_backlog_threshold * 0.8
       Page.from_tag_parts("PGArchivalBacklogHigh", id)&.incr_resolve
     end
   rescue => ex
     Clog.emit("Failed to observe archival backlog", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
   end
 
+  def should_page_for_archival_backlog?(archival_backlog, oldest_pending, previous_oldest, now, previous_time)
+    return true if archival_backlog > archival_backlog_threshold * 2
+    return true unless previous_oldest && oldest_pending && previous_time
```
byucesoy: In general, these variable names can benefit from being bit longer & descriptive. It is not possible to understeand what previous_oldest/oldest_pending mean without looking at caller.

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,68 @@ def refresh_walg_credentials
...
+
+  def estimate_archival_rate(previous_oldest_pending_wal, current_oldest_pending_wal, now, previous_time)
+    bytes_archived = lsn_diff(wal2lsn(current_oldest_pending_wal), wal2lsn(previous_oldest_pending_wal))
+    return 0 if bytes_archived <= 0
+
+    elapsed_seconds = now - previous_time
+    return Float::INFINITY if elapsed_seconds <= 0
+
+    bytes_archived / elapsed_seconds.to_f
+  end
```
enescakir, line 430: I’m not sure if you're extracting `estimate_archival_rate` for easier testing. IMHO, The reader has to reason about what returning 0 or Float::INFINITY means, which adds unnecessary cognitive overhead.


```suggestion
  def should_page_for_archival_backlog?(archival_backlog, current_oldest_pending_wal, previous_oldest_pending_wal, now, previous_time)
    return true if archival_backlog > archival_backlog_threshold * 2
    return true unless previous_oldest_pending_wal && current_oldest_pending_wal && previous_time

    bytes_archived = lsn_diff(wal2lsn(current_oldest_pending_wal), wal2lsn(previous_oldest_pending_wal))
    return true if bytes_archived <= 0
    
    elapsed_seconds = now - previous_time
    return false if elapsed_seconds <= 0

    (bytes_archived / elapsed_seconds.to_f) < MIN_ARCHIVAL_RATE_BYTES_PER_SEC
  end
```

shikharbhardwaj: The estimation was more elaborate in a previous version, makes sense to coalesce it into one here

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,62 @@ def refresh_walg_credentials
...
+    session[:previous_oldest_pending_wal] = oldest_pending
+    session[:previous_archival_check_time] = now
+
     if archival_backlog > archival_backlog_threshold
-      Prog::PageNexus.assemble("#{ubid} archival backlog high",
-        ["PGArchivalBacklogHigh", id], ubid,
-        severity: "warning", extra_data: {archival_backlog:})
-    else
+      if should_page_for_archival_backlog?(archival_backlog, oldest_pending, previous_oldest, now, previous_time)
+        disk_usage_percent = session[:disk_usage_percent] || 0
```
furkansahin, line 408: can this be nil? Seems we call the `observe_disk_usage` before the `archival_backlog` now

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,62 @@ def refresh_walg_credentials
...
+    elsif archival_backlog < archival_backlog_threshold * 0.8
       Page.from_tag_parts("PGArchivalBacklogHigh", id)&.incr_resolve
     end
   rescue => ex
     Clog.emit("Failed to observe archival backlog", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
   end
 
+  def should_page_for_archival_backlog?(archival_backlog, current_oldest_pending_wal, previous_oldest_pending_wal, now, previous_time)
+    return true if archival_backlog > archival_backlog_threshold * 2
+    return true unless previous_oldest_pending_wal && current_oldest_pending_wal && previous_time
```
furkansahin, line 423: can you explain/comment why the existence of these values warrant a page?

shikharbhardwaj: Its a failsafe, to page if the backlog is above threshold but we don't have data from the previous pulse to estimate the upload rate.

model/postgres/postgres_server.rb
shikharbhardwaj: It should not be, but the dependency on `observe_disk_usage` is indirect, hence the fallback

model/postgres/postgres_server.rb
```diff
@@ -405,9 +405,11 @@ def observe_archival_backlog(session)
 
     if archival_backlog > archival_backlog_threshold
       if should_page_for_archival_backlog?(archival_backlog, oldest_pending, previous_oldest, now, previous_time)
+        disk_usage_percent = session[:disk_usage_percent] || 0
+        severity = (disk_usage_percent >= 85) ? "error" : "warning"
         Prog::PageNexus.assemble("#{ubid} archival backlog high",
```
byucesoy, line 410: Even if you change the severity, I think we don't have any mechanism to update the page in the pagerduty. I'm not sure but I suspect if you add a new semaphore which triggers calling `page.trigger` (like we do in start label of `PageNexus`), it might work.

model/postgres/postgres_server.rb
byucesoy, line 450: Just remembered that `storage_size_gib` can be `0` temporarily during provisioning, because Vm might not have storage volumes created yet. No need to change anything right now. Proper fix would be creating the storage volumes eagerly. Just nothing it here in case we hit this case so we wouldn't spend time on debugging why.

model/postgres/postgres_server.rb
```diff
@@ -385,23 +385,60 @@ def refresh_walg_credentials
...
     elsif archival_backlog < archival_backlog_threshold * 0.8
       Page.from_tag_parts("PGArchivalBacklogHigh", id)&.incr_resolve
     end
   rescue => ex
     Clog.emit("Failed to observe archival backlog", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
   end
 
+  def should_page_for_archival_backlog?(archival_backlog, current_oldest_pending_wal, previous_oldest_pending_wal, now, previous_time)
+    return true if archival_backlog > archival_backlog_threshold * 2
+    return true unless previous_oldest_pending_wal && current_oldest_pending_wal && previous_time
```
byucesoy, line 423: Wouldn't this create a page immediatelly after every monitor restart? AFAIR, session values are stored only in memory, they are not persistent.

byucesoy: Another interesting case, what happens if there was no `.ready` file on the filesystem? Would `previous_oldest_pending_wal` be `nil` in that case, thus would we create a page unnecessarily?

In general, I feel like we can be more relaxed here and not create a page. I'm not sure if we need these failsafes here. In next iteration of archival observation, we would have the session values set and we can make an informed decision at that point. Is there a case where these values are never set? I guess that can happen during unavailability, but we would get an unavailable page at that time anyway.

shikharbhardwaj: > Wouldn't this create a page immediatelly after every monitor restart?

It would, if the instance already had a backlog above threshold (this function only gets called if the count is above the threshold).

It won't page in case there's no ready file, as the threshold won't be breached in that case.

model/postgres/postgres_server.rb
shikharbhardwaj: I updated PageNexus to retrigger the existing page if the severity is increased from an existing page.

prog/page_nexus.rb
```diff
@@ -7,6 +7,10 @@ def self.assemble(summary, tag_parts, related_resources, severity: "error", extr
     DB.transaction do
       details = extra_data.merge({"related_resources" => Array(related_resources)})
       tag = Page.generate_tag(tag_parts)
+
+      existing_page = Page.active.where(tag:).first
+      severity_increased = existing_page && Page.severity_order(severity) > Page.severity_order(existing_page.severity)
```
enescakir: Since we're in the same transaction, why not just increase the semaphore here

```suggestion
      if (existing_page = Page.active.first(tag:)) && Page.severity_order(severity) > Page.severity_order(existing_page.severity)
          existing_page.incr_retrigger
      end
```

PR #4766, opened 2026-02-12
Add mtr check to mesh checker

mohi-kalantari:
**Add on-demand MTR diagnostics to mesh connectivity checker**
When intermittent connectivity issues occur between Kubernetes nodes, simple reachability checks tell us a pod is unreachable but not where in the network path the problem lies. This PR adds failure-triggered MTR diagnostics that automatically run when connectivity checks fail, providing hop-by-hop network visibility.

**On-demand MTR**: When a pod connectivity check fails, MTR runs against the failed pod IP. For external endpoints, DNS resolution failures trigger MTR against CoreDNS pod IPs, while connection errors trigger MTR against the endpoint directly. MTR targets are processed through a background queue that blocks until work arrives. Results are cleared automatically when targets recover, and stale entries for removed pods are pruned each cycle.

**Configurable log level**: The mesh checker runs kubectl commands every 30 seconds to discover pods, generating redundant info-level log noise. A log level parameter on the Kubernetes client (defaulting to info) lets the mesh checker demote these to debug.

**API server error reporting**: Previously, if the API server became unreachable, the status file retained the last known healthy state - giving a false impression that everything was fine. The checker now records the API error with backtrace in the status file and clears it once the API recovers. The node availability check treats the API error as unavailable, surfacing it through Clog emissions.

**MTR results in availability checks**: MTR results are included in the availability result hash when connectivity issues are detected, making network path data available in Clog emissions. They are informational and do not affect the availability decision.

### DIFF
--- a/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
@@ -38,7 +38,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.4.2
+          image: ubicloud/ubicsi:v0.5.0
           imagePullPolicy: IfNotPresent
           args: ["controller"]
           env:
--- a/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
@@ -22,7 +22,7 @@ spec:
       priorityClassName: system-node-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.4.2
+          image: ubicloud/ubicsi:v0.5.0
           imagePullPolicy: IfNotPresent
           args: ["node"]
           env:


kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -231,6 +261,47 @@ def write_status_file
...
+          target = @mtr_queue.pop
+          break if target.nil? # queue closed
+          run_mtr_for_target(target)
+          write_status_file
+        end
+      end
+    end
+
+    def enqueue_mtr(name, ip)
+      @mtr_queue << {name:, ip:}
```
jeremyevans, line 276: After `@mtr_queue.close` is called during `shutdown!`, this can start raising exceptions. Unless you are 100% sure that this method cannot be called after `shutdown!` You should probably rescue the exception:

```suggestion
      @mtr_queue << {name:, ip:}
    rescue ClosedQueueError
      nil
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -231,6 +261,47 @@ def write_status_file
...
+    def run_mtr_for_target(target)
+      output, status = run_cmd("timeout", MTR_TIMEOUT.to_s, "mtr", "-c2", "-rw", target[:ip], req_id: "mtr-check")
+      @mutex.synchronize do
+        @mtr_results[target[:name]] = {
+          ip: target[:ip],
+          output: output.strip,
+          exit_status: status.exitstatus,
+          last_check: Time.now.utc.iso8601
+        }
+      end
```
jeremyevans, line 303:
```suggestion
      hash = {
        ip: target[:ip],
        output: output.strip,
        exit_status: status.exitstatus,
        last_check: Time.now.utc.iso8601
      }
      key = target[:name]
      @mutex.synchronize do
        @mtr_results[key] = hash
      end
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -231,6 +261,47 @@ def write_status_file
...
+      @logger.debug("[MeshConnectivity] MTR completed for #{target[:name]} (#{target[:ip]})")
+    rescue => e
+      @logger.error("[MeshConnectivity] MTR failed for #{target[:name]} (#{target[:ip]}): #{e.message}")
```
jeremyevans:
```suggestion
    rescue => e
      @logger.error("[MeshConnectivity] MTR failed for #{target[:name]} (#{target[:ip]}): #{e.message}")
    else
      @logger.debug("[MeshConnectivity] MTR completed for #{target[:name]} (#{target[:ip]})")
```

kubernetes/csi/spec/ubi_csi/kubernetes_client_spec.rb
```diff
@@ -236,4 +236,34 @@
...
+
+    it "returns only running CoreDNS pods" do
+      expect(client).to receive(:run_kubectl).with("-n", "kube-system", "get", "pods", "-l", "k8s-app=kube-dns", "-oyaml").and_return(YAML.dump(pods_list))
+
+      result = client.get_coredns_pods
+
+      expect(result.size).to eq(2)
+      expect(result).to include({"name" => "coredns-abc123", "ip" => "10.96.0.5"})
+      expect(result).to include({"name" => "coredns-xyz789", "ip" => "10.96.0.6"})
+      expect(result).not_to include(hash_including("name" => "coredns-pending"))
```
jeremyevans: In general, it's best to use exact matches, unless there is nondeterministic data in the returned hash. I don't believe there is in this case.

```suggestion
      expect(client.get_coredns_pods).to eq([{"name" => "coredns-abc123", "ip" => "10.96.0.5"}, {"name" => "coredns-xyz789", "ip" => "10.96.0.6"}])
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -87,26 +97,38 @@
...
-      expect(logger).to receive(:debug).with("[MeshConnectivity] Found 2 nodeplugin pods").and_call_original
       expect(checker).to receive(:check_endpoints).with([{host: "10.0.0.2", port: 8080, name: "ubicsi-nodeplugin-xyz"}])
         .and_yield({host: "10.0.0.2", port: 8080, name: "ubicsi-nodeplugin-xyz"}, true, nil)
 
       checker.check_all_pods_connectivity
 
-      status = checker.instance_variable_get(:@pod_status)["ubicsi-nodeplugin-xyz"]
-      expect(status[:reachable]).to be true
-      expect(status[:ip]).to eq("10.0.0.2")
+      expect(checker.instance_variable_get(:@mtr_results)).not_to have_key("ubicsi-nodeplugin-xyz")
```
jeremyevans: This type of spec is prone to false negatives (the hash may contain other keys that shouldn't be there). It's better to compare the actual value:

```suggestion
      expect(checker.instance_variable_get(:@mtr_results).keys).to eq []
```

Similar issue in a later spec cases where you are checking keys.

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -434,19 +468,164 @@
...
+    it "processes targets from queue until shutdown" do
+      target = {name: "ubicsi-nodeplugin-xyz", ip: "10.0.0.2"}
+      expect(checker).to receive(:run_mtr_for_target).with(target)
+      expect(checker).to receive(:write_status_file)
+
+      thread = checker.spawn_mtr_check_thread
+      checker.instance_variable_get(:@mtr_queue) << target
+      sleep(0.1)
+      checker.shutdown!
+      thread.join(1)
```
jeremyevans: The `sleep` approach is inherently either flaky or wasteful (there is no in between).

```suggestion
      expect(checker).to receive(:write_status_file) do
        super()
        shutdown!
      end

      thread = checker.spawn_mtr_check_thread
      checker.instance_variable_get(:@mtr_queue) << target
      thread.join(1)
      expect(thread).not_to be_alive
```

kubernetes/csi/spec/ubi_csi/mesh_connectivity_checker_spec.rb
```diff
@@ -434,19 +468,164 @@
...
+
+      checker.enqueue_mtr_for_coredns
+
+      queue = checker.instance_variable_get(:@mtr_queue)
+      targets = []
+      targets << queue.pop(true) until queue.empty?
+      expect(targets).to contain_exactly(
+        {name: "coredns:coredns-abc", ip: "10.96.0.5"},
+        {name: "coredns:coredns-xyz", ip: "10.96.0.6"}
+      )
```
jeremyevans: `contain_exactly` should be used when you cannot control the order. In this case, I think you control the order:

```suggestion
      expect(targets).to eq([
        {name: "coredns:coredns-abc", ip: "10.96.0.5"},
        {name: "coredns:coredns-xyz", ip: "10.96.0.6"}
      ])
```

kubernetes/csi/lib/ubi_csi/service_helper.rb
```diff
@@ -6,7 +6,8 @@
 module Csi
   module ServiceHelper
     def log_with_id(req_id, message)
-      @logger.info("[req_id=#{req_id}] #{message}")
+      level = defined?(@log_level) ? @log_level : :info
+      @logger.send(level, "[req_id=#{req_id}] #{message}")
```
jeremyevans: A `nil`/`false` `@log_level` would not be valid, so you can simplify this:
```suggestion
      @logger.send(@log_level || :info, "[req_id=#{req_id}] #{message}")
```

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
```diff
@@ -231,6 +261,47 @@ def write_status_file
...
+      pods.each do |pod|
+        next unless pod["ip"]
+        enqueue_mtr("coredns:#{pod["name"]}", pod["ip"])
+      end
+    rescue => e
+      @logger.error("[MeshConnectivity] Failed to get CoreDNS pods for MTR: #{e.message}")
+    end
+
+    def run_mtr_for_target(target)
+      output, status = run_cmd("timeout", MTR_TIMEOUT.to_s, "mtr", "-c2", "-rw", target[:ip], req_id: "mtr-check")
```
fdr: what do you think about `-n` to turn off DNS resolution? DNS resolution is one of the more expensive/slower steps, and can be extremely handy when crossing the internet, but probably(?) doesn't do much for inter-pod testing.

I suppose I will do a more thorough review after Jeremy's stuff is closed out. Though it's unlikely I will find more ruby problems than he. I'm a bit puzzled by how much code there is to add this in general, though it seems to fit into a framework I don't understand yet.

mohi-kalantari: for inter-pod testing, we don't run dns queries and directly run mtr to pod ips.

we run mtr to coredns pods only when we get an error trying to connect to external endpoints.

I feel mtr to coredns pods can be useful if a customer has dns issues. it can remove one probable root cause fast.

mohi-kalantari:
I've done thorough testing already. When I get the approval, I'll do another round to make sure the discussion changes didn't affect anything. @aamederen

aamederen:
and I'm leaving it to @fdr :)

kubernetes/csi/lib/ubi_csi/mesh_connectivity_checker.rb
fdr: it's one thing to check coredns, I'm saying...do you want to bother doing reverse-dns lookups?

mohi-kalantari: you're right I added the -n

mohi-kalantari:
Hey @aamederen 
I added 3 new commits to fix some issues:
- Do not run mesh checker on provisioner pods as they are required to run only on nodeplugin pods (daemonset pods)
- Notifying the customer when disk size is above our limits. Small change. So I included that here
- Use prefer instead of require in provisioner affinity rules to fix the update issue we have in 1 worker node clusters.
Can you have a look at those please?

kubernetes/csi/lib/ubi_csi/controller_service.rb
```diff
@@ -67,7 +67,7 @@ def create_volume(req, _call)
           raise GRPC::InvalidArgument.new("Volume name is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.name.nil? || req.name.empty?
           raise GRPC::InvalidArgument.new("Capacity range is required", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.capacity_range.nil?
           raise GRPC::InvalidArgument.new("Required bytes must be positive", GRPC::Core::StatusCodes::INVALID_ARGUMENT) if req.capacity_range.required_bytes <= 0
-          raise GRPC::InvalidArgument.new("Volume size exceeds maximum allowed size of #{max_volume_size / OneGB}GB", GRPC::Core::StatusCodes::OUT_OF_RANGE) if req.capacity_range.required_bytes > max_volume_size
+          raise GRPC::OutOfRange.new("Volume size exceeds maximum allowed size of #{max_volume_size / OneGB}GB") if req.capacity_range.required_bytes > max_volume_size
```
aamederen: nit: These are very long lines and makes reading a bit harder. For this file, we can switch to explicit `if` blocks, rather than trailing ifs to make it more readable.

aamederen: also, we can mention the requested size in the error message as well.

aamederen: could we have any integer division issues with `#{max_volume_size / OneGB}GB`. If at some point, we allow 4.5 GB, would it be displayed correctly, instead of `4GB` or `4.4999999999999999` ?

rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
```diff
@@ -18,18 +18,20 @@ spec:
...
-                  - key: app
-                    operator: In
-                    values:
-                      - ubicsi
-                  - key: component
-                    operator: In
-                    values:
-                      - provisioner
-              topologyKey: "kubernetes.io/hostname"
+          preferredDuringSchedulingIgnoredDuringExecution:
```
aamederen, line 23: This makes sense. How do we make sure that two ubicsi provisioners are not acting together at all? We should have some code to prevent both provisioners "start working" and also a test that, even though two provisioners are allocated in the same node for some reason, the layout is safe.

mohi-kalantari: That would be a bigger change which we can tackle later. Usually in kubernetes, there is a leader election happening on a `Lease` resource and default go clients support the leader election but it would be trickier on ruby. For now we can assume we will have one provisioner.

kubernetes/csi/lib/ubi_csi/controller_service.rb
mohi-kalantari: fixed all

rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
aamederen: OK, can we add a TODO in the code then for this case?

PR #4767, opened 2026-02-12
Add stop action for vm in admin panel

enescakir:


clover_admin.rb
```diff
@@ -245,7 +245,8 @@ def self.object_action(...)
       end
     },
     "Vm" => {
-      "restart" => object_action("Restart", "Restart scheduled for Vm", &:incr_restart)
+      "restart" => object_action("Restart", "Restart scheduled for Vm", &:incr_restart),
+      "stop" => object_action("Stop", "Stop scheduled for Vm", &:incr_stop)
```
serprex, line 249: seems like this should include start when stopped

enescakir: We don't have the start action. This PR adds it https://github.com/ubicloud/ubicloud/pull/4677

enescakir: We use stop action heavily for fraud activities at the moment

PR #4768, opened 2026-02-12
Fix csi upgrade incompatibility

mohi-kalantari:
**Wait for CSI data copy before removing Kubernetes nodes**
**Problem**
When a Kubernetes node is retired (upgrade or nodepool resize), the CSI driver starts a background rsync to copy persistent volume data from the old node to the new one. Previously, the node nexus proceeded directly from drain to node removal, destroying the VM while the copy was potentially still in progress. Since rsync reads from the old node over SSH, this caused irreversible data loss.

**Solution**
Added a wait_for_copy label in KubernetesNodeNexus between drain and remove_node_from_cluster. It queries all PersistentVolumes for the old-pvc-object annotation with nodeAffinity pointing to the retiring node. While such PVs exist, the nexus naps and rechecks every 15 seconds. The node remains visible in the cluster during the wait for debuggability.

**E2E Test**
Added an end-to-end test that retires a node hosting a StatefulSet pod and verifies:

- The node is not removed while PVs still reference it for data migration
- Data integrity is preserved after the copy completes via sha256sum comparison

**Changes**
- New wait_for_copy label in the node nexus, drain hops to it on success
- Three new e2e test labels: test_node_not_deleted_during_copy, verify_node_not_deleted_during_copy, verify_data_after_drain
- Unit tests for wait_for_copy
- Specs for the new e2e test labels, refactored subject to include kubernetes_cluster_id in the strand stack

### DIFF
--- a/prog/test/kubernetes.rb
+++ b/prog/test/kubernetes.rb
@@ -5,7 +5,7 @@ require_relative "../../lib/util"
 class Prog::Test::Kubernetes < Prog::Test::Base
   semaphore :destroy
 
-  MIGRATION_TRIES = 3
+  MIGRATION_TRIES = 2
 
   def self.assemble
     kubernetes_test_project = Project.create(name: "Kubernetes-Test-Project")
@@ -114,7 +114,7 @@ spec:
     spec:
       accessModes: [ "ReadWriteOnce" ]
       resources:
-        requests: { storage: 1Gi }
+        requests: { storage: 5Gi }
       storageClassName: ubicloud-standard
 STS
     kubernetes_cluster.sshable.cmd("sudo kubectl --kubeconfig /etc/kubernetes/admin.conf apply -f -", stdin: sts)
@@ -138,13 +138,47 @@ STS
   end
 
   label def test_data_write
-    write_hash = kubernetes_cluster.client.kubectl("exec -t ubuntu-statefulset-0 -- sh -c \"head -c 200M /dev/urandom | tee /etc/data/random-data | sha256sum | awk '{print \\$1}'\"").strip
-    read_hash = kubernetes_cluster.client.kubectl("exec -t ubuntu-statefulset-0 -- sh -c \"sha256sum /etc/data/random-data | awk '{print \\$1}'\"").strip
-    if write_hash != read_hash
-      update_stack({"fail_message" => "wrong read hash, expected: #{write_hash}, got: #{read_hash}"})
-      hop_destroy_kubernetes
+    (1..3).each do |i|
+      unit_name = "csi_data_write_#{i}"
+      kubernetes_cluster.sshable.d_run(
+        unit_name,
+        "bash", "-c",
+        "sudo kubectl --kubeconfig /etc/kubernetes/admin.conf exec -t ubuntu-statefulset-0 -- sh -c \"head -c 500M /dev/urandom | tee /etc/data/random-data-#{i} | sha256sum | awk '{print \\$1}'\" > /dev/shm/#{unit_name}.hash"
+      )
     end
-    update_stack({"read_hash" => read_hash})
+    hop_wait_data_write
+  end
+
+  label def wait_data_write
+    (1..3).each do |i|
+      unit_name = "csi_data_write_#{i}"
+      case kubernetes_cluster.sshable.d_check(unit_name)
+      when "InProgress"
+        nap 30
+      when "Failed"
+        update_stack({"fail_message" => "daemonized write for random-data-#{i} failed"})
+        hop_destroy_kubernetes
+      end
+    end
+    hop_verify_data_write
+  end
+
+  label def verify_data_write
+    read_hashes = {}
+    (1..3).each do |i|
+      file = "random-data-#{i}"
+      unit_name = "csi_data_write_#{i}"
+      hash_path = "/dev/shm/#{unit_name}.hash"
+      write_hash = kubernetes_cluster.sshable.cmd("cat :hash_path", hash_path:).strip
+      read_hash = kubernetes_cluster.client.kubectl("exec -t ubuntu-statefulset-0 -- sh -c \"sha256sum /etc/data/:file | awk '{print \\$1}'\"", file:).strip
+      kubernetes_cluster.sshable.d_clean(unit_name)
+      if write_hash != read_hash
+        update_stack({"fail_message" => "wrong read hash for #{file}, expected: #{write_hash}, got: #{read_hash}"})
+        hop_destroy_kubernetes
+      end
+      read_hashes[file] = read_hash
+    end
+    update_stack({"read_hashes" => read_hashes})
     hop_test_pod_data_migration
   end
 
@@ -162,12 +196,7 @@ STS
 
   label def verify_data_after_migration
     nap 5 unless pod_status == "Running"
-    new_hash = kubernetes_cluster.client.kubectl("exec -t ubuntu-statefulset-0 -- sh -c \"sha256sum /etc/data/random-data | awk '{print \\$1}'\"").strip
-    expected_hash = strand.stack.first["read_hash"]
-    if new_hash != expected_hash
-      update_stack({"fail_message" => "data hash changed after migration, expected: #{expected_hash}, got: #{new_hash}"})
-      hop_destroy_kubernetes
-    end
+    verify_data_hashes("migration")
     hop_test_normal_pod_restart if migration_number == MIGRATION_TRIES
     increment_migration_number
     hop_test_pod_data_migration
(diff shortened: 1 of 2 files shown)


prog/test/kubernetes.rb
```diff
@@ -194,6 +194,60 @@ def self.assemble
...
+    pod_node = nodepool.nodes.find { |n| n.name == pod_node_name }
+    pod_node.incr_retire
+
+    hop_verify_node_not_deleted_during_copy
+  end
+
+  label def verify_node_not_deleted_during_copy
+    drain_node_name = strand.stack.first["drain_test_node_name"]
+    drain_node = KubernetesNode.where(kubernetes_cluster_id: kubernetes_cluster.id)
+      .all.find { |n| n.name == drain_node_name }
```
jeremyevans:
```suggestion
      .first(name: drain_node_name)
```

prog/test/kubernetes.rb
```diff
@@ -194,6 +194,60 @@ def self.assemble
...
+
+    # Node record destroyed means the nexus completed the full retire flow
+    hop_verify_data_after_drain unless drain_node
+
+    pvs = JSON.parse(kubernetes_cluster.client.kubectl("get pv -ojson"))["items"]
+    pending_copy = pvs.any? do |pv|
+      pv.dig("metadata", "annotations", "csi.ubicloud.com/old-pvc-object") &&
+        pv.dig("spec", "nodeAffinity", "required", "nodeSelectorTerms", 0,
+          "matchExpressions", 0, "values", 0) == drain_node_name
+    end
```
jeremyevans: This duplicates logic in `KubernetesNodeNexus#wait_for_copy`. Probably should add a method to `KubernetesCluter` and have both places call that.

prog/kubernetes/kubernetes_node_nexus.rb
```diff
@@ -92,6 +92,17 @@ def nodepool
     end
   end
 
+  label def wait_for_copy
+    pvs = JSON.parse(cluster.client.kubectl("get pv -ojson"))["items"]
```
aamederen: nit: We can create a function like `yaml_load_kubectl` in this client as well. Can be a different PR.

aamederen: let's log something about content of `pvs` or `pending`

prog/test/kubernetes.rb
```diff
@@ -194,6 +194,60 @@ def self.assemble
...
     rescue => e
       update_stack({"fail_message" => e.message})
+      hop_destroy_kubernetes
+    end
+    hop_test_node_not_deleted_during_copy
+  end
+
+  label def test_node_not_deleted_during_copy
+    client = kubernetes_cluster.client
+    nodepool.nodes.each { |node|
```
aamederen: nit: the blocks can be converted to use `it` for some simplicity, like `nodepool.nodes.each { client.kubectl("uncordon :name", name: it.name) }`

prog/test/kubernetes.rb
aamederen: I'd say, 200M is a bit low for such a test. Shall we bump this up to about 10T? If 10T is too big for E2E test duration, let's see how big we can get away with.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -160,6 +160,10 @@ def worker_vms
     nodepools.flat_map(&:vms)
   end
 
+  def service_nics_on_subnet
```
aamederen: I did not like this approach. This is breaking some unnecessary abstraction. Kubernetes cluster should not be concerned about the nics of the vms of the nodes.

Would it be simpler to solve this by making KubernetesNodes wait for the destroy of the VMs?

> KubernetesNodeNexus#destroy deletes the KubernetesNode record synchronously but destroys the VM asynchronously via incr_destroy.

aamederen: Or, we can improve `incr_destroy_if_only_used_internally` to filter out the resources that are marked for destroy

prog/test/kubernetes.rb
mohi-kalantari: 10T is really huge! I'm not sure how it helps in the case of e2e tests. We are filling with random data and getting its hash. Increasing the number of files seems more helpful to me.

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: Vms can be deleted faster than nics. I encountered that issue during my tests.

mohi-kalantari: improving `incr_destroy_if_only_used_internally` makes more sense

prog/test/kubernetes.rb
mohi-kalantari: name is not a field in database and that won't work. But I changed to this:
```
    drain_node = kubernetes_cluster.all_nodes.detect { |n| n.name == drain_node_name }

```

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: improved incr_destroy_if_only_used_internally

prog/test/kubernetes.rb
aamederen: 10T is ridiculous, I meant 10G :D Current setting (3x10G) is better.

aamederen:
The latest E2E test run reports success, but it failed actually: https://github.com/ubicloud/ubicloud/actions/runs/22155136452/job/64061027999 

Can you take a look at why it failed, and why it reported success?

> {"contended_session_lock":{"exit_code":124,"session_fail_msg":"session lock conflict for respirate","sshable_ubid":"vmhdgvkadaka1pddvb2k4ctm4e","prog":"Prog::Test::Kubernetes#test_data_write"},"message":"session lock failure","time":"2026-02-18 22:55:36 +0000","thread":"st06e75qd3na5jdxyn5c9b4gpy"}

cc. @jeremyevans

aamederen (changes requested):
Code looks good overall, but we need to make sure that
1. E2E tests are reliable and pass multiple times
2. The "test data" is large enough to take some time, but small enough to not put
3. E2E test Failures are reported correctly

If this would take some time, it's OK to extract the E2E test part and work on it in a separate PR

aamederen:
In the E2E test, `wait_data_write`, `verify_data_after_migration`, `verify_normal_pod_restart` and many other `wait` states in the test prog seem to cause issues with respirate:
```
"contended_session_lock":{"exit_code":124,"session_fail_msg":"session lock conflict for respirate","sshable_ubid":"vmv4h6w0r80p3mh8pwee8skzrb","prog":"Prog::Test::Kubernetes#wait_data_write"},"message":"session lock failure","time":"2026-02-23 15:47:44 +0000","thread":"st06e8p999hy2gqwstab0g3eb3"}
```

These should be fixed, can be a subsequent PR to not block this prod-impacting fix.

The data migration took ~6 minutes. I think it's a good value, but since we perform this operation multiple times, it increased the E2E time significantly (from 45m to 85m). My suggestion is to keep it for now and increase the timeouts for E2E tests.

```
15:48:52 respirate.1 | {"strand_hopped":{"duration":2.095726505,"from":"Test::Kubernetes.test_pod_data_migration","to":"Test::Kubernetes.verify_data_after_migration"},"message":"hopped","time":"2026-02-23 15:48:52 +0000","thread":"st06e8p999hy2gqwstab0g3eb3"}
...
[9](https://github.com/ubicloud/ubicloud/actions/runs/22311308742/job/64543917291#step:15:8070)
15:54:16 respirate.1 | {"strand_hopped":{"duration":324.943191928,"from":"Test::Kubernetes.verify_data_after_migration","to":"Test::Kubernetes.test_pod_data_migration"},"message":"hopped","time":"2026-02-23 15:54:16 +0000","thread":"st06e8p999hy2gqwstab0g3eb3"}
```

aamederen (approved):

mohi-kalantari:
@aamederen  How much should I set as the new deadline? 2 hours can be enough but since we provision a new e2e job every two hours, after merging this PR, it may never converge as people keep submitting e2e jobs manually. Maybe we should change the e2e frequency on main to every 3 hours. 

@enescakir I wanted to ask if you have any comments here.

enescakir:
Increasing the duration from 4m to 85m is significant, almost double. IMHO, it negatively affects developer productivity. During busy times, we already have to wait in the queue for E2E tests, and this will only make things worse.

A few possible workarounds come to mind, but we can also discuss them with the team:
- Do we really need to migrate that much data, or can we reduce the scope or frequency of migrations?
- We could consider not running these tests on every run by default and instead run them on a daily schedule.

aamederen:
We need long transfers, otherwise, it's not easy to capture the "node gone before data transfer" cases. Separating out the "long" tests could help

mohi-kalantari:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22353783893

mohi-kalantari:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22355424908

aamederen:
With the latest changes, the E2E-metal test finished in 56m in the latest run [1]. 

For reference, the last 3 successful scheduled E2E-metal tests against main took 69m, 50m and 48m[2]. I think we can ship in this form and work on the long-haul tests afterwards.

[1] 
- https://github.com/ubicloud/ubicloud/actions/runs/22355424908

[2]
- https://github.com/ubicloud/ubicloud/actions/runs/22339300856
- https://github.com/ubicloud/ubicloud/actions/runs/22342337050
- https://github.com/ubicloud/ubicloud/actions/runs/22354981448

PR #4772, opened 2026-02-13
Show restart requirement for PostgreSQL config parameters

furkansahin:
## Add restart requirement metadata for PostgreSQL config parameters
Some PostgreSQL configuration parameters require a database restart
to take effect (context=postmaster), while others only need a reload.
Add requires_restart: true to each postmaster-context parameter in the
config validator schema and expose restart_required_params/requires_restart?
methods on the validator so consumers can determine which params need
a restart.

## Add tests for restart required config params

CleanShot 2026-02-17 at 12 29 35@2x

### DIFF
--- a/spec/lib/validation/postgres_config_validator_spec.rb
+++ b/spec/lib/validation/postgres_config_validator_spec.rb
@@ -174,4 +174,46 @@ RSpec.describe Validation::PostgresConfigValidator do
       end
     end
   end
+
+  describe "#restart_required_params" do
+    it "returns restart-required params for PG 16" do
+      v = described_class.new("16")
+      params = v.restart_required_params
+      expect(params).to include("max_connections", "shared_buffers", "wal_level", "max_worker_processes")
+      expect(params).to include("archive_mode")
+      expect(params).not_to include("work_mem", "log_statement")
+    end
+
+    it "returns restart-required params for PG 17 with new params" do
+      v = described_class.new("17")
+      params = v.restart_required_params
+      expect(params).to include("max_connections", "shared_buffers", "commit_timestamp_buffers", "archive_mode")
+      expect(params).not_to include("old_snapshot_threshold")
+      expect(params).not_to include("io_combine_limit", "summarize_wal", "allow_alter_system")
+    end
+
+    it "returns restart-required params for PG 18 with new params" do
+      v = described_class.new("18")
+      params = v.restart_required_params
+      expect(params).to include("max_connections", "io_method", "autovacuum_worker_slots", "max_active_replication_origins")
+      expect(params).not_to include("io_workers", "io_combine_limit", "extension_control_path")
+    end
+
+    it "returns empty set for pgbouncer" do
+      v = described_class.new("pgbouncer")
+      expect(v.restart_required_params).to be_empty
+    end
+  end
+
+  describe "#requires_restart?" do
+    it "returns true for restart-required params" do
+      expect(validator.requires_restart?("max_connections")).to be true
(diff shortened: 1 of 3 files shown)


lib/validation/postgres_config_validator_schema.rb
```diff
@@ -2576,5 +2576,100 @@ class PostgresConfigValidatorSchema
       removed.each { |key| schema.delete(key) }
       schema.merge!(added_or_modified)
     end.freeze
+
+    PG_16_RESTART_REQUIRED = Set.new(%w[
```
byucesoy: I think it would be better to add new field to the related configs above. Something like

```ruby
"cron.database_name" => {
  ...
  requires_restart: true
},
```

With a separate set, future readers would need to look into 2 places to understand given config fully.

lib/validation/postgres_config_validator.rb
```diff
@@ -17,6 +17,14 @@ def initialize(version)
       end
     end
 
+    def restart_required_params
+      @restart_required ||= @config_schema.select { |_, v| v[:requires_restart] }.keys.to_set.freeze
```
byucesoy: `@restart_required` -> `@restart_required_params`

routes/project/location/postgres.rb
```diff
@@ -621,19 +621,32 @@ class Clover
...
+            response = {
               pg_config: pg.user_config,
               pgbouncer_config: pg.pgbouncer_user_config
             }
+            response[:message] = "Some changes require a database restart to take effect." if restart_param_changed
+            response
           else
-            flash["notice"] = "Configuration updated successfully"
+            flash["notice"] = if restart_param_changed
+              "Configuration updated successfully. Some changes require a database restart to take effect."
```
byucesoy: Maybe add which changes require database restart to be more specific?

byucesoy: Also it would be good to say something like you can restart the database from settings page.

views/postgres/config.erb
```diff
@@ -1,6 +1,7 @@
-<% 
+<%
   pg_config = @pg.user_config || {}
   pgbouncer_config = @pg.pgbouncer_user_config || {}
+  restart_required_params = Validation::PostgresConfigValidator.new(@pg.version).restart_required_params
```
byucesoy: Instead of generating restart_required_params here and passing to the config card, why don't we create the validator there directly? In that way we can also reuse `requires_restart?` function.

lib/validation/postgres_config_validator_schema.rb
```diff
@@ -22,7 +22,8 @@ class PostgresConfigValidatorSchema
       "allow_in_place_tablespaces" => {
         description: "Allows tablespaces directly inside pg_tblspc, for testing.",
         type: :bool,
-        default: "off"
+        default: "off",
+        requires_restart: true
```
shikharbhardwaj, line 26: Wondering if we should do a similar exercise for pgbouncer configs

routes/project/location/postgres.rb
shikharbhardwaj: +1, we can add a note below the configs to guide the user

views/postgres/config.erb
furkansahin: moved it

lib/validation/postgres_config_validator_schema.rb
furkansahin: there is no pgbouncer config that we allow customers to set requires restart

PR #4776, opened 2026-02-13
Increase Inference Router Replica Page severity to error

aamederen:
By removing the explicit `severity: "warning"` setting when creating a page, we know create inference router replica pages in "error" mode, when inference router itself, or one of the models is unavailable.

severity=error mode triggers notification process in PagerDuty, which is important as we have more customers rely on our inference router in critical workloads.


bsatzger (approved):

enescakir:
Can you also write a TSG for oncall?

PR #4779, opened 2026-02-14
Add admin panel extras for PostgresServer

serprex:
Display metrics like arcival backlog, wal lag, replay timestamp

Need to collect metrics with collect-pg-metrics script as postgres-exporter does not include these

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -153,6 +153,24 @@ class CloverAdmin < Roda
     end
   end
 
+  def format_bytes(bytes)
+    if bytes < 1024
+      "#{bytes.round}B"
+    elsif bytes < 1024**2
+      "#{(bytes / 1024.0).round(1)}KiB"
+    elsif bytes < 1024**3
+      "#{(bytes / 1024.0**2).round(1)}MiB"
+    else
+      "#{(bytes / 1024.0**3).round(1)}GiB"
+    end
+  end
+
+  def format_seconds(s)
+    m, s = s.divmod(60)
+    h, m = m.divmod(60)
+    "%02d:%02d:%02d" % [h, m, s]
+  end
+
   def available_classes
     classes = []
     Sequel::Model.subclasses.each do |c|
--- a/views/admin/components/svg_chart.erb
+++ b/views/admin/components/svg_chart.erb
@@ -1,0 +1,3 @@
+<%# locals: (points:, nonce:, &block) %>
+
+<%== SvgChart.render(points, label_formatter: block, nonce:) %>
--- a/views/admin/extras/PostgresServer.erb
+++ b/views/admin/extras/PostgresServer.erb
@@ -1,0 +1,38 @@
+<%# locals: (obj:) %>
+
+<% query_metric = lambda do |tsdb_client, metric, resource_ubid|
+  next [] unless tsdb_client
+  query = metric.series[0].query.gsub("$ubicloud_resource_id", resource_ubid)
+  end_ts = Time.now.to_i
+  start_ts = end_ts - 6 * 3600
+  result = tsdb_client.query_range(query:, start_ts:, end_ts:)
+  next [] if result.empty?
+  result[0]["values"].map { |v| [v[0].to_f, v[1].to_f] }
+ rescue VictoriaMetrics::ClientError
+  []
+end
+
+tsdb_client = PostgresServer.victoria_metrics_client
+backlog_points = query_metric.call(tsdb_client, Metrics::ADMIN_POSTGRES_METRICS[:archival_backlog], obj.resource.ubid)
+lag_points = query_metric.call(tsdb_client, Metrics::ADMIN_POSTGRES_METRICS[:wal_lag], obj.resource.ubid)
+replay_points = query_metric.call(tsdb_client, Metrics::ADMIN_POSTGRES_METRICS[:replay_lag], obj.resource.ubid) if obj.standby? || obj.read_replica?
+
+nonce = SecureRandom.hex(32)
+content_security_policy.add_img_src [:nonce, nonce] %>
+
+<div class="chart-grid">
+  <div class="chart-cell">
+    <h3>Archival Backlog</h3>
+    <%== part("components/svg_chart", points: backlog_points, nonce:) %>
+  </div>
+  <div class="chart-cell">
+    <h3>WAL Lag</h3>
+    <%== part("components/svg_chart", points: lag_points, nonce:) { format_bytes(it) } %>
+  </div>
+  <% if replay_points %>
+    <div class="chart-cell">
+      <h3>Replay Lag</h3>
+      <%== part("components/svg_chart", points: replay_points, nonce:) { format_seconds(it) } %>
+    </div>
+  <% end %>
+</div>
--- a/lib/metrics.rb
+++ b/lib/metrics.rb
@@ -190,4 +190,43 @@ module Metrics
       ]
     )
   }
+
+  ADMIN_POSTGRES_METRICS = {
+    archival_backlog:
+    MetricDefinition.new(
+      name: "Archival Backlog",
+      description: "Number of WAL files pending archival",
+      unit: "count",
+      series: [
+        TimeSeries.new(
+          labels: {name: "Pending WAL files"},
+          query: "ubicloud_pg_archival_backlog{ubicloud_resource_id=\"$ubicloud_resource_id\"}"
+        )
+      ]
+    ),
+    wal_lag:
+    MetricDefinition.new(
+      name: "WAL Lag",
+      description: "WAL replication lag from primary",
+      unit: "bytes",
+      series: [
+        TimeSeries.new(
+          labels: {name: "Lag"},
+          query: "max(ubicloud_pg_wal_lsn_bytes{ubicloud_resource_id=\"$ubicloud_resource_id\"}) - ubicloud_pg_wal_lsn_bytes{ubicloud_resource_id=\"$ubicloud_resource_id\"}"
+        )
+      ]
+    ),
+    replay_lag:
+    MetricDefinition.new(
+      name: "Replay Lag",
+      description: "Replication replay lag (standbys/replicas)",
+      unit: "seconds",
+      series: [
+        TimeSeries.new(
+          labels: {name: "Lag"},
+          query: "time() - ubicloud_pg_last_xact_replay_timestamp_seconds{ubicloud_resource_id=\"$ubicloud_resource_id\"}"
+        )
+      ]
+    )
+  }
 end
--- a/public/admin/app.css
+++ b/public/admin/app.css
@@ -230,3 +230,18 @@ code {
   td { padding: 3px 5px; }
   td:first-child { min-width: 150px; }
 }
+
+.svg-chart {
+  font: 11px monospace;
+}
+
+.chart-grid {
+  display: flex;
+  gap: 16px;
+  flex-wrap: wrap;
+  margin-bottom: 16px;
+}
+.chart-cell {
+  flex: 1;
+  min-width: 400px;
+}
--- a/spec/routes/web/admin/admin_spec.rb
+++ b/spec/routes/web/admin/admin_spec.rb
@@ -1729,4 +1729,54 @@ RSpec.describe CloverAdmin do
       expect(audit_log_content).to eq []
     end
   end
+
+  describe "format_bytes" do
+    let(:app) { described_class.allocate }
+
+    it "formats bytes" do
+      expect(app.format_bytes(0)).to eq "0B"
+      expect(app.format_bytes(1023)).to eq "1023B"
+    end
+
+    it "formats kibibytes" do
+      expect(app.format_bytes(1024)).to eq "1.0KiB"
+      expect(app.format_bytes(1600)).to eq "1.6KiB"
+      expect(app.format_bytes(1024**2 - 1)).to eq "1024.0KiB"
+    end
+
+    it "formats mebibytes" do
+      expect(app.format_bytes(1024**2)).to eq "1.0MiB"
+      expect(app.format_bytes(1024**2 * 5.55)).to eq "5.6MiB"
+      expect(app.format_bytes(1024**3 - 1)).to eq "1024.0MiB"
+    end
+
+    it "formats gibibytes" do
+      expect(app.format_bytes(1024**3)).to eq "1.0GiB"
+      expect(app.format_bytes(1024**3 * 2.25)).to eq "2.3GiB"
+    end
+  end
+
+  describe "format_seconds" do
+    let(:app) { described_class.allocate }
+
+    it "formats zero" do
+      expect(app.format_seconds(0)).to eq "00:00:00"
+    end
+
+    it "formats seconds only" do
+      expect(app.format_seconds(45)).to eq "00:00:45"
+    end
+
+    it "formats minutes and seconds" do
+      expect(app.format_seconds(125)).to eq "00:02:05"
+    end
+
+    it "formats hours, minutes, and seconds" do
+      expect(app.format_seconds(3661)).to eq "01:01:01"
+    end
+
+    it "format allows for more than 2 digits in hour" do
+      expect(app.format_seconds(999999)).to eq "277:46:39"
+    end
+  end
 end
--- a/spec/routes/web/admin/model/postgres_server_spec.rb
+++ b/spec/routes/web/admin/model/postgres_server_spec.rb
@@ -13,6 +13,8 @@ RSpec.describe CloverAdmin, "PostgresServer" do
   end
 
   it "displays the PostgresServer instance page correctly" do
+    expect(PostgresServer).to receive(:victoria_metrics_client).and_return(nil)
+
     click_link "PostgresServer"
     expect(page.status_code).to eq 200
     expect(page.title).to eq "Ubicloud Admin - PostgresServer - Browse"
@@ -29,5 +31,15 @@ RSpec.describe CloverAdmin, "PostgresServer" do
 
     expect(page).to have_content("Recycle scheduled for PostgresServer")
     expect(Semaphore.where(strand_id: @instance.id, name: "recycle").count).to eq 1
+  end
+
+  it "renders charts when metrics are available" do
+    tsdb_client = instance_double(VictoriaMetrics::Client)
+    expect(PostgresServer).to receive(:victoria_metrics_client).and_return(tsdb_client)
+    expect(tsdb_client).to receive(:query_range).twice.and_return([{"labels" => {}, "values" => [[Time.now.to_i, "5"]]}])
+
+    visit "/model/PostgresServer/#{@instance.ubid}"
+    expect(page.status_code).to eq 200
+    expect(page.body).to include("<svg")
   end
 end


enescakir:
As far as I know, we don't prefer run ssh commands on vms from the web app.

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,14 @@
+<%# locals: (obj: nil) %>
+
+<% current_lsn = obj.current_lsn rescue nil %>
```
jeremyevans: In general, bare `rescue` like this is considered a smell. 3 such calls in the same template is very smelly. For something like this, it's probably better to do the rescuing inside a model method, as that allows for easier testing (you don't appear to be testing any of the `rescue` failure cases in this template). Ideally, you would only rescue the specific exception classes you expect.

views/admin/extras/PostgresTimeline.erb
```diff
@@ -0,0 +1,13 @@
+<%# locals: (obj: nil) %>
+
+<div>
+  Current backup status:
+  <%= obj.leader.vm.sshable.cmd("common/bin/daemonizer --check take_postgres_backup") rescue "Unavailable" %>
```
jeremyevans: The web process should not SSH to resources. We do have at least one other existing case where it does, but we would like to remove such cases. Our long term goal is to remove the ability for the web process to SSH to resources, as allowing it adds risk.

If you'd like to show this information on the admin site, we probably need to track the information in a database table, have the health monitor update it, and then this can just read from the database.

clover_admin.rb
```diff
@@ -194,6 +194,9 @@ def self.object_action(...)
     "PostgresResource" => {
       "restart" => object_action("Restart", "Restart scheduled for PostgresResource", &:incr_restart)
     },
+    "PostgresServer" => {
+      "recycle" => object_action("Recycle", "Recycle PostgresServer", &:incr_recycle)
```
jeremyevans: This doesn't appear to be tested.

views/admin/extras/PostgresServer.erb
```diff
@@ -9,6 +9,7 @@
   data: [{
     "Archive Backlog" => obj.primary? ? (obj.archival_backlog rescue nil) : nil,
     "Current LSN" => current_lsn,
-    "WAL Lag (GiB)" => lsn_lag && (lsn_lag / (1024**3))
+    "WAL Lag (GiB)" => lsn_lag && (lsn_lag / (1024**3)),
+    "Replay timestamep" => obj.standby? || obj.read_replica? ? (obj.pg_last_xact_replay_timestamp rescue nil) : nil
```
jeremyevans: Similar to the SSH case, we should not be adding new code where the web process connects directly to customer postgres databases. If we want to display this information on the admin site, we should have the health monitor record it in a database table, and have the web process read from the database table.

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,14 @@
...
+<% current_lsn = obj.current_lsn rescue nil %>
+<% lsn_lag = obj.lsn_lag(current_lsn) rescue nil %>
+
+<%== part("components/table",
+  title: "WAL Lag",
+  table_class: "postgres-wal-lag",
+  data: [{
+    "Archive Backlog" => obj.primary? ? (obj.archival_backlog rescue nil) : nil,
+    "Current LSN" => current_lsn,
+    "WAL Lag (GiB)" => lsn_lag && (lsn_lag / (1024**3))
```
jeremyevans: It seems odd to only report WAL lag in whole GB increments (rounded down), when we start treating things differently after 80MB of WAL lag.

enescakir:
Technically, `run_query` also is also ssh command 😅 

https://github.com/ubicloud/ubicloud/blob/2daee191b324c1f5acb6dab651a3d2da65f237b1/model/postgres/postgres_server.rb#L309-L312

serprex:
latest version relies on https://github.com/ubicloud/postgres-vm-images/pull/16 to render based on metrics

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
...
+    "#{(bytes / 1024.0**3).round(1)} GiB"
+  end
+end
+
+format_seconds = lambda do |secs|
+  h = (secs / 3600).floor
+  m = ((secs % 3600) / 60).floor
+  s = (secs % 60).floor
+  "%02d:%02d:%02d" % [h, m, s]
+end
```
serprex: would be nice to put these with utilities somewhere, not sure where that'd be

assets/js/app.js
```diff
@@ -849,6 +849,13 @@ function unitFormatter(unit, precision = 0) {
     return bytesFormatter(unit, precision);
   } else if (unit == "IOPS" || unit.startsWith("ops") || unit.startsWith("count") || unit.startsWith("deadlock")) {
     return opsFormatter(unit, precision);
+  } else if (unit == "seconds") {
+    return function (value) {
+      var h = Math.floor(value / 3600);
+      var m = Math.floor((value % 3600) / 60);
+      var s = Math.floor(value % 60);
+      return String(h).padStart(2, '0') + ':' + String(m).padStart(2, '0') + ':' + String(s).padStart(2, '0');
```
jeremyevans: It doesn't appear this function needs to operate as a closure, so I would make it a named function and return it, so you aren't creating a new function object per call.

rhizome/postgres/bin/collect-pg-metrics
```diff
@@ -0,0 +1,57 @@
...
+end
+
+version = ARGV[0]
+output_path = "/var/lib/node_exporter/pg_metrics.prom"
+tmp_path = "#{output_path}.tmp"
+
+lines = []
+
+# Archival backlog: count of WAL files pending archival
+backlog = `sudo find /dat/#{version}/data/pg_wal/archive_status -name '*.ready' 2>/dev/null | wc -l`.strip.to_i
```
jeremyevans: In general, using backtick to execute commands is a poor practice, as a command failure is not caught. We don't use this approach anywhere for that reason. In rhizome, use the `r` method to run external programs.

For any command where you will be interpolating and don't need shell metacharacters, pass separate arguments (treated as argv) instead of a single argument (which will run using a shell). With this currently line, you are using `wc` to count lines, so you need to use a shell. This is especially risky, as you are interpolating and using `sudo` in the same command. So you need to be 100% sure what you are interpolating is valid.

When converting from string to integer, prefer strict conversions (`"bogus".to_i => 0`, `Integer("bogus", 10) # raises`):

```suggestion
backlog = Integer(r "sudo find /dat/#{version}/data/pg_wal/archive_status -name '*.ready' 2>/dev/null | wc -l".strip, 10)
```

All use of backtick to execute commands should be updated to use `r`.

rhizome/postgres/bin/collect-pg-metrics
```diff
@@ -0,0 +1,57 @@
...
+# frozen_string_literal: true
+
+# Collects custom PostgreSQL metrics and writes them as prometheus textfile
+# for node_exporter's textfile collector. Runs periodically via systemd timer.
+
+if ARGV.count != 1
+  fail "Wrong number of arguments. Expected 1 (pg version), given #{ARGV.count}"
+end
+
+version = ARGV[0]
```
jeremyevans: You are interpolating this value into `sudo` commands, it makes sense to be strict about validation:

```suggestion
version = Integer(ARGV[0], 10)
```

rhizome/postgres/bin/collect-pg-metrics
```diff
@@ -0,0 +1,57 @@
...
+  if is_recovery == "t"
+    ts = `sudo -u postgres psql -t -A -c "SELECT extract(epoch FROM pg_last_xact_replay_timestamp())"`.strip
+    unless ts.empty?
+      lines << "# HELP ubicloud_pg_last_xact_replay_timestamp_seconds Last transaction replay timestamp as Unix epoch"
+      lines << "# TYPE ubicloud_pg_last_xact_replay_timestamp_seconds gauge"
+      lines << "ubicloud_pg_last_xact_replay_timestamp_seconds #{ts}"
+    end
+  end
+rescue
+  nil
```
jeremyevans: Ignoring all errors around a large block of code is usually a bad idea. Let's say you make a typo in a variable in a later commit, and it's never caught because you are ignoring all errors. Then days/weeks after deployment, you figure out you are missing metrics.

If you want to ignore a specific type of error, you should only rescue that type of error. You would need to provide a very good justification for ignoring all errors.

views/admin/components/svg_chart.erb
jeremyevans, line 1: This entire file looks undertested. SVG creation should be extracted to a class (in `lib`) and extensively tested to make sure it returns the expected output for a given input.

views/admin/extras/PostgresServer.erb
jeremyevans: An instance method in `CloverAdmin` would work.

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
...
+
+query_metric = lambda do |tsdb_client, metric_key, resource_ubid|
+  next [] unless tsdb_client
+  metric = Metrics::POSTGRES_METRICS[metric_key]
+  next [] unless metric
+
+  query = metric.series[0].query.gsub("$ubicloud_resource_id", resource_ubid)
+  start_ts = Time.now.to_i - 6 * 3600
+  end_ts = Time.now.to_i
```
jeremyevans:
```suggestion
  end_ts = Time.now.to_i
  start_ts = end_time - 6 * 3600
```

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
...
+  result = tsdb_client.query_range(query:, start_ts:, end_ts:)
+  next [] if result.empty?
+  result[0]["values"].map { |v| [v[0].to_f, v[1].to_f] }
+rescue
+  []
```
jeremyevans: Same issue of ignoring all exceptions here. This should be changed to rescue only specific exception classes.

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
...
+end
+
+tsdb_client = PostgresServer.victoria_metrics_client
+resource_ubid = obj.resource.ubid
+backlog_points = query_metric.call(tsdb_client, :archival_backlog, resource_ubid)
+lag_points = query_metric.call(tsdb_client, :wal_lag, resource_ubid)
+replay_points = (obj.standby? || obj.read_replica?) ? query_metric.call(tsdb_client, :replay_lag, resource_ubid) : nil %>
```
jeremyevans:
```suggestion
replay_points = query_metric.call(tsdb_client, :replay_lag, resource_ubid) if obj.standby? || obj.read_replica? %>
```

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
...
+
+<div style="display: flex; gap: 16px; flex-wrap: wrap; margin-bottom: 16px;">
```
jeremyevans: The admin site content security policy doesn't allow inline styles. Use a class attribute, and style the classes appropriate in the admin stylesheet. Use appropriate cascading (ideally, only a single class attribute for this element should be needed).

views/admin/components/svg_chart.erb
```diff
@@ -0,0 +1,35 @@
...
+  span_t = [max_t - min_t, 1].max
+
+  max_v = [(points.map { it[1] }.max || 0) * 1.1, 1].max
+
+  sx = ->(t) { pl + (t - min_t).to_f / span_t * cw }
+  sy = ->(v) { pt + ch - v.to_f / max_v * ch }
+
+  path_d = points.each_with_index.map { |p, i| "#{(i == 0) ? "M" : "L"}#{sx.call(p[0]).round(1)},#{sy.call(p[1]).round(1)}" }.join
+  area_d = "#{path_d}L#{sx.call(points.last[0]).round(1)},#{pt + ch}L#{sx.call(points.first[0]).round(1)},#{pt + ch}Z" %>
+  <svg xmlns="http://www.w3.org/2000/svg" width="<%= w %>" height="<%= h %>" style="font:11px monospace">
```
jeremyevans: The admin site content security policy does not allow inline `svg` tags. You need to use a nonce included in each tag, and use `content_security_policy.add_img_src [:nonce, nonce_value]`.

See `views/components/form/resource_creation_form.erb` for an example of how this is done for inline javascript in Clover.

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
+<%# locals: (obj: nil) %>
```
jeremyevans: I realize this is just copying the existing extra templates, but those also appear to be wrong. The argument for the template should be required.

```suggestion
<%# locals: (obj:) %>
```

Same for PostgresTimeline.

I submitted #5038 to fix the existing templates

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
...
+    "#{(bytes / 1024.0**2).round(1)}MiB"
+  else
+    "#{(bytes / 1024.0**3).round(1)}GiB"
+  end
+end
+
+format_seconds = lambda do |secs|
+  h = (secs / 3600).floor
+  m = ((secs % 3600) / 60).floor
+  s = (secs % 60).floor
```
jeremyevans: Unlike javascript, in Ruby you can use `divmod` to clean this up.

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,58 @@
...
+  "%02d:%02d:%02d" % [h, m, s]
+end
+
+query_metric = lambda do |tsdb_client, metric_key, resource_ubid|
```
jeremyevans: This lambda should probably be a `PostgresResource` instance method.

assets/js/app.js
jeremyevans, line 1: The admin site doesn't use this file, only the console does, so I'm not sure why the file is being modified.

serprex: new metrics appear in metrics dashboard for pg resources. maybe we shouldn't show them

jeremyevans: Ah, I didn't realize that. The commit message did not indicate there were changes to the console display. I would only show the metrics in the console if you think there is a good reason to expose them to the user. If they are just exposed by accident and not deliberately, that's probably something we should avoid.

serprex:
great, looks like general design here is agreeable. splitting up non-metrics dependent parts to separate PRs: #5042 #5043

assets/js/app.js
serprex: Personally as a customer I'd want to see these new metrics. But happy to keep this PR simpler with no JS changes

views/admin/extras/PostgresServer.erb
serprex: that seems like the wrong place. it's a bit specific to interfacing with svg component. maybe I can get svg component processing input well enough that boilerplate is minimal enough to inline

views/admin/components/svg_chart.erb
```diff
@@ -0,0 +1,5 @@
+<%# locals: (points:, &block) %>
+
+<% nonce = SecureRandom.hex(32)
+content_security_policy.add_img_src [:nonce, nonce] %>
+<%== SvgChart.render(points, label_formatter: block, nonce:) %>
```
serprex, line 3: ideally `render` would return some kind of `html_string` but that seems to be a rails thing

rhizome/postgres/bin/collect-pg-metrics
```diff
@@ -0,0 +1,52 @@
...
+  unless ts.empty?
+    lines << "# HELP ubicloud_pg_last_xact_replay_timestamp_seconds Last transaction replay timestamp as Unix epoch\n"
+    lines << "# TYPE ubicloud_pg_last_xact_replay_timestamp_seconds gauge\n"
+    lines << "ubicloud_pg_last_xact_replay_timestamp_seconds #{ts}\n"
+  end
+end
+
+# Atomic write
+File.write(tmp_path, lines.join)
+File.rename(tmp_path, output_path)
```
enescakir: We also have `safe_write_to_file` if you want to use

prog/postgres/postgres_server_nexus.rb
```diff
@@ -317,13 +317,45 @@ def before_run
...
+After=postgresql.service
+
+[Service]
+Type=oneshot
+User=ubi
+ExecStart=/home/ubi/postgres/bin/collect-pg-metrics #{postgres_server.version}
+StandardOutput=journal
+StandardError=journal
+SERVICE
+    vm.sshable.write_file("/etc/systemd/system/pg-collect-metrics.service", pg_metrics_service)
```
enescakir, line 337: Up to you. I generally prefer using `<<~` for better indentation, and there’s no need to use a variable.

```suggestion
    vm.sshable.write_file("/etc/systemd/system/pg-collect-metrics.service", <<~SERVICE)
      [Unit]
      Description=Postgres Metrics Collection
      After=postgresql.service

      [Service]
      Type=oneshot
      User=ubi
      ExecStart=/home/ubi/postgres/bin/collect-pg-metrics #{postgres_server.version}
      StandardOutput=journal
      StandardError=journal
    SERVICE
```

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,33 @@
...
+  next [] unless tsdb_client
+  query = metric.series[0].query.gsub("$ubicloud_resource_id", resource_ubid)
+  end_ts = Time.now.to_i
+  start_ts = end_ts - 6 * 3600
+  result = tsdb_client.query_range(query:, start_ts:, end_ts:)
+  next [] if result.empty?
+  result[0]["values"].map { |v| [v[0].to_f, v[1].to_f] }
+end
+
+tsdb_client = PostgresServer.victoria_metrics_client
```
enescakir, line 15: When we open any PostgresServer page, it’s shown by default. I’m not sure how much latency this adds, but we might consider putting it behind a button.

Also, do we have a timeout limit, or should we rescue errors? If the metric server is down, does this page still load?

Lastly, I don’t have a metric server locally, does that affect this page?

prog/postgres/postgres_server_nexus.rb
serprex: I agree with you, but am being consistent with rest of code

views/admin/extras/PostgresServer.erb
serprex: updated to render "No data available" when excon socket exception thrown, as is the case when victoriametrics isn't running locally

views/admin/components/svg_chart.erb
jeremyevans: This uses a separate nonce per svg. I would use a single nonce for all elements on the page (different nonce per request, of course). The template should require `nonce` as an argument.

views/admin/extras/PostgresServer.erb
serprex: may need to adjust given #5067

rhizome/postgres/bin/collect-pg-metrics
```diff
@@ -0,0 +1,49 @@
...
+output_path = "/var/lib/node_exporter/pg_metrics.prom"
+
+lines = []
+
+# Archival backlog: count of WAL files pending archival
+backlog = r("sudo", "find", "/dat/#{version}/data/pg_wal/archive_status", "-name", "*.ready").count("\n")
+lines << "# HELP ubicloud_pg_archival_backlog Number of WAL files pending archival\n"
+lines << "# TYPE ubicloud_pg_archival_backlog gauge\n"
+lines << "ubicloud_pg_archival_backlog #{backlog}\n"
+
```
shikharbhardwaj, line 20: I was thinking of adding this as another query to postgres_exporter, since we already set custom queries in `/usr/local/share/postgresql/postgres_exporter_queries.yaml`.
But that requires the ubi_monitoring user to have more permissions/or a `SECURITY DEFINER` making rollout painful, so this seems like a better approach.

views/admin/extras/PostgresServer.erb
```diff
@@ -0,0 +1,38 @@
...
+  next [] if result.empty?
+  result[0]["values"].map { |v| [v[0].to_f, v[1].to_f] }
+ rescue VictoriaMetrics::ClientError
+  []
+end
+
+tsdb_client = PostgresServer.victoria_metrics_client
+backlog_points = query_metric.call(tsdb_client, Metrics::ADMIN_POSTGRES_METRICS[:archival_backlog], obj.resource.ubid)
+lag_points = query_metric.call(tsdb_client, Metrics::ADMIN_POSTGRES_METRICS[:wal_lag], obj.resource.ubid)
+replay_points = query_metric.call(tsdb_client, Metrics::ADMIN_POSTGRES_METRICS[:replay_lag], obj.resource.ubid) if obj.standby? || obj.read_replica?
```
shikharbhardwaj, line 18: nit: can this iterate over `ADMIN_POSTGRES_METRICS`, might avoid some repetition

serprex: `:replay_lag` only applying to standbys means it can't be a straight loop without including some logic to skip certain queries in some circumstance. prefer to keep 3 simple lines

PR #4781, opened 2026-02-15
Add cache_proxy_download_url feature flag for GitHub runners

enescakir:
Regenerating VM boot images to test new cache-proxy versions is slow and inconvenient. This feature flag allows per-project override of the cache-proxy binary by providing a download URL per architecture. When set, the runner's setup_environment step stops the service, downloads and extracts the new binary, then restarts the service. This increases provisioning time, but it is acceptable for testing new cache-proxy versions.

The feature flag value is a hash keyed by architecture, e.g.: {"x64": "https://...", "arm64": "https://..."}


aamederen (approved):
lgtm. As a nitpick, `cache_proxy_url` variable can be renamed to `download_url` or the `download_url` parameter can be renamed to `cache_proxy_url` to use keyword argument shorthand.

PR #4784, opened 2026-02-16
Add trailing newline to postgres configs

fdr:
Per customary unix.

Otherwise, simple arrangements like `cat` or a hack using `tee -a` in a scrape are not going to do the right thing.

### DIFF
--- a/rhizome/postgres/bin/configure
+++ b/rhizome/postgres/bin/configure
@@ -27,11 +27,11 @@ safe_write_to_file("/etc/hosts", hosts)
 
 # Update postgresql.conf
 configs = configure_hash["configs"].map { |k, v| "#{k} = #{v}" }.join("\n")
-safe_write_to_file("/etc/postgresql/#{v}/main/conf.d/001-service.conf", configs)
+safe_write_to_file("/etc/postgresql/#{v}/main/conf.d/001-service.conf", configs + "\n")
 
 # Update postgresql.conf with custom settings
 user_configs = configure_hash["user_config"].map { |k, v| "#{k} = #{v}" }.join("\n")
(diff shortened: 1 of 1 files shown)


shikharbhardwaj (approved):

byucesoy (approved):

rhizome/postgres/bin/configure
```diff
@@ -27,11 +27,11 @@ safe_write_to_file("/etc/hosts", hosts)
 
 # Update postgresql.conf
 configs = configure_hash["configs"].map { |k, v| "#{k} = #{v}" }.join("\n")
-safe_write_to_file("/etc/postgresql/#{v}/main/conf.d/001-service.conf", configs)
+safe_write_to_file("/etc/postgresql/#{v}/main/conf.d/001-service.conf", configs + "\n")
 
 # Update postgresql.conf with custom settings
 user_configs = configure_hash["user_config"].map { |k, v| "#{k} = #{v}" }.join("\n")
-safe_write_to_file("/etc/postgresql/#{v}/main/conf.d/099-user.conf", user_configs)
+safe_write_to_file("/etc/postgresql/#{v}/main/conf.d/099-user.conf", user_configs + "\n")
```
enescakir, line 35: another option is that not using join

```ruby
user_configs = configure_hash["user_config"].map { |k, v| "#{k} = #{v}\n" }
safe_write_to_file("/etc/postgresql/#{v}/main/conf.d/099-user.conf", user_configs)
```

fdr: interesting. I suppose it gets to_s called on it? probably need a test to be that slick.

PR #4787, opened 2026-02-17
Skip checks for existence of representative_server

byucesoy:
We guarantee that the representative server exists so we don't need these checks or branchs anymore.


spec/model/postgres/postgres_resource_spec.rb
```diff
@@ -159,6 +149,7 @@
     end
 
     it "provisions a new server with the correct timeline for a regular instance" do
+      ps1
```
shikharbhardwaj, line 152: Why is this required?

byucesoy: The variables defined with let are lazily evaluated, so if you don't access them they are not "exist". This was OK previously because due to the nil checks, we were taking a different code path that doesn't need these variables. With the changes in this PR, we now need these variables, so we need to access them once at the beginning of the test.

PR #4788, opened 2026-02-17
Fix missing DNS records when adding ports to load balancer

mohi-kalantari:
LoadBalancer#add_port and #remove_port were the only mutation methods
that did not call incr_rewrite_dns_records. This caused a race
condition during Kubernetes LoadBalancer service creation:
sync_kubernetes_services calls add_vm (which sets rewrite_dns_records)
before add_port, but since the services LB starts with zero ports,
add_vm creates zero LoadBalancerVmPort entries. If the LB nexus strand
processes rewrite_dns_records before add_port runs, vms_to_dns returns
empty (the join through load_balancer_vm_port finds nothing), so DNS
records are deleted and never recreated — add_port only set
update_load_balancer, not rewrite_dns_records.

Adding incr_rewrite_dns_records to both add_port and remove_port
ensures DNS records are rewritten after port changes, matching the
behavior of add_vm, remove_vm, evacuate_vm, and remove_vm_port.

### DIFF
--- a/model/load_balancer.rb
+++ b/model/load_balancer.rb
@@ -67,6 +67,7 @@ class LoadBalancer < Sequel::Model
         add_port_with_stack(port, lb_vm.id)
       end
       incr_update_load_balancer
+      incr_rewrite_dns_records
     end
   end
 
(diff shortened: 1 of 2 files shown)


model/load_balancer.rb
furkansahin, line 83: do you maybe need to add it here as well?

mohi-kalantari: yeah thanks. I added it

furkansahin (approved):

PR #4790, opened 2026-02-17
Enable GRO on veth interfaces for XFRM ESP tunnel throughput

fdr:
Veth interfaces have GRO disabled by default. Enabling it allows the receive path to coalesce packets before they enter the XFRM stack, significantly improving ESP tunnel throughput between VMs.

Benchmark (iperf3 -P4 -t30, Hetzner bare-metal, Debian 13, kernel 6.18):
  Baseline (no veth GRO):    1.92 Gbps
  With veth GRO:             3.46-3.58 Gbps (~80% improvement)

The ethtool calls are wrapped in begin/rescue so GRO enablement failure does not break VM setup on kernels or configurations that don't support it.

Note: TAP rx-udp-gro-forwarding is intentionally NOT enabled. Testing showed it actually reduces native ESP throughput (2.71 Gbps vs 3.5 Gbps with veth GRO alone). It is only beneficial for ESP-in-UDP encapsulation, which Clover does not use.

### DIFF
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -268,6 +268,12 @@ add element inet drop_unused_ip_packets allowed_ipv4_addresses { #{ip_net} }
     # /sys/class/net/vethi#{q_vm}/address at two points in time.  The
     # result is a race condition that *sometimes* worked.
     r "ip link add vetho#{q_vm} addr #{gen_mac.shellescape} type veth peer name vethi#{q_vm} addr #{gen_mac.shellescape} netns #{q_vm}"
+    # Enable GRO on veth interfaces to improve XFRM ESP tunnel throughput.
+    # GRO is off by default on veth; enabling it allows the receive path to
+    # coalesce packets before they enter the XFRM stack, yielding ~80%
+    # higher throughput on native ESP tunnels (5.14+, all target distros).
+    r "ethtool -K vetho#{q_vm} gro on"
+    r "ip netns exec #{q_vm} ethtool -K vethi#{q_vm} gro on"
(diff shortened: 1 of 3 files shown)


rhizome/host/lib/vm_setup.rb
```diff
@@ -268,6 +268,12 @@ def interfaces(nics, multiqueue)
     # /sys/class/net/vethi#{q_vm}/address at two points in time.  The
     # result is a race condition that *sometimes* worked.
     r "ip link add vetho#{q_vm} addr #{gen_mac.shellescape} type veth peer name vethi#{q_vm} addr #{gen_mac.shellescape} netns #{q_vm}"
+    # Enable GRO on veth interfaces to improve XFRM ESP tunnel throughput.
+    # GRO is off by default on veth; enabling it allows the receive path to
+    # coalesce packets before they enter the XFRM stack, yielding ~80%
+    # higher throughput on native ESP tunnels (5.14+, all target distros).
+    r "ethtool -K vetho#{q_vm} gro on"
```
enescakir: You need to install `ethtool` on the existing host before deploying this change.

fdr: yeah.

fdr: It may be there, but I was testing debian at the time as well (it works) and it has less stuff in it, and I figured, why rely on defaults: we need it

furkansahin:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/32138073133

furkansahin (approved):
Let's do this, I'll handle the tool installation before the deployment.

0xjunhao (approved):

fdr:
oh man people came back to this? alright let me refresh it

PR #4795, opened 2026-02-18
Add assoc table links to admin object view

enescakir:
Some associations have too many entries to browse as a bulleted list in the object view. i.e. I want to see all the runner details for a installation, or all the vm details for a project.

Instead of duplicating the data in the object view, we can just link to a pre-filtered search page for the association.

I don't like the if/else logic in the view, but it was the quickest way to get this working. We can refactor later if we want to.

CleanShot 2026-02-18 at 18 27 55@2x

CleanShot 2026-02-18 at 18 28 10@2x

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -274,6 +274,11 @@ class CloverAdmin < Roda
     .each_with_object({}) { |name, h| h[name] = true }
     .freeze
 
+  OBJECT_ASSOC_TABLE_PARAMS = {
+    ["GithubInstallation", :runners] => "installation",
+    ["Project", :vms] => "project"
+  }.freeze
+
   plugin :autoforme do
     # :nocov:
     register_by_name if Config.development?
@@ -380,8 +385,10 @@ class CloverAdmin < Roda
     model GithubRunner do
       order Sequel.desc(:created_at)
       eager_graph [:strand]
+      eager [:installation]
       columns do |type_symbol, request|
         cs = [:repository_name, :label, :strand_label, :created_at]
+        cs.prepend(:installation) if type_symbol == :search_form
         cs.prepend(:ubid) unless type_symbol == :search_form
         cs
       end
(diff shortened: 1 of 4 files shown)


views/admin/object.erb
```diff
@@ -56,10 +56,20 @@
       Array(@obj.send(assoc))
     end
     next if assoc_objs.empty?
+    table_param = if @klass.name == "GithubInstallation" && associated_class.name == "GithubRunner"
+      "installation"
+    elsif @klass.name == "Project" && associated_class.name == "Vm"
+      "project"
+    end
```
jeremyevans: Instead of using a conditional here, I recommend switching to one of two approaches:

* Hash based approach: Use hash keyed on `[class_name, association_name]` (not `associated_class.name`) with `table_param` values. The hash would be stored as a constant in `CloverAdmin`.
* Add an `:admin_table_param` option to the two associations, and check `reflection[:admin_table_param]`.

clover_admin.rb
```diff
@@ -379,9 +379,10 @@ def self.object_action(...)
 
     model GithubRunner do
       order Sequel.desc(:created_at)
-      eager_graph [:strand]
+      eager_graph [:strand, :installation]
```
jeremyevans: You don't appear to be ordering or filtering on installation columns, so there is no need to use `eager_graph`:

```suggestion
      eager_graph [:strand]
      eager [:installation]
```

views/admin/object.erb
enescakir: I refactored to hash based approach

enescakir: It makes sense a lot! Thanks

clover_admin.rb
```diff
@@ -274,6 +274,11 @@ def self.object_action(...)
     .each_with_object({}) { |name, h| h[name] = true }
     .freeze
 
+  OBJECT_ASSOC_TABLE_PARAMS = {
+    [GithubInstallation, "runners"] => "installation",
+    [Project, "vms"] => "project"
```
jeremyevans: The issue here is this forces autoloading the constants, and it does not work correctly with reloading in development mode. It's better to use class names, as I recommended. I would also use association symbol.

```suggestion
    ["GithubInstallation", :runners] => "installation",
    ["Project", :vms] => "project"
```

views/admin/object.erb
```diff
@@ -59,7 +59,12 @@
     %>
 
     <div class="association">
-      <h3><%= assoc %></h3>
+      <h3>
+        <%= assoc %>
+        <% if table_param = OBJECT_ASSOC_TABLE_PARAMS[[@klass, assoc.name]] %>
```
jeremyevans:
```suggestion
        <% if table_param = OBJECT_ASSOC_TABLE_PARAMS[[@klass.name, assoc]] %>
```

jeremyevans (approved):
Looks good to merge with a couple changes.

PR #4797, opened 2026-02-19
Remove duplicate helper functions in Postgres tests

byucesoy:
While removing mocking, we added 5-6 version of each create_postgres_[timeline|resource|server] functions. They mostly do the same thing. With this PR, we are consolidating them. At the same time, I realized each of these accepts lots of parameters just to satisfy 1-2 callers' needs. I also removed infrequently or unnecessarily (i.e. they are not important for what is being tested) used parameters.

### DIFF
--- a/spec/common/postgres.rb
+++ b/spec/common/postgres.rb
@@ -30,6 +30,28 @@ module PostgresTestHelpers
     Strand.create_with_id(pg, prog: "Postgres::PostgresResourceNexus", label: "start")
     pg
   end
+
+  def create_postgres_server(
+    resource:,
+    timeline: create_postgres_timeline(location_id: resource.location_id),
+    is_representative: true,
+    timeline_access: is_representative ? "push" : "fetch"
+  )
+    vm = Prog::Vm::Nexus.assemble_with_sshable(resource.project_id, location_id: resource.location_id).subject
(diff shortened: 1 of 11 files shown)


spec/model/postgres/postgres_server_spec.rb
```diff
@@ -856,10 +820,13 @@ def create_standby_resource(suffix)
...
         LocationCredential.create(location:, assume_role: "role")
 
-        parent = PostgresTimeline.create(location:)
+        credential = LocationCredential[location:]
+        parent = create_postgres_timeline(location_id: location.id)
         timeline.update(parent:)
-        expect(postgres_server.timeline.location.location_credential).to receive(:aws_iam_account_id).and_return("aws-account-id").at_least(:once)
-        expect(postgres_server.timeline.location.location_credential).to receive(:iam_client).and_return(iam_client)
+        allow(postgres_server.timeline.location).to receive(:location_credential).and_return(credential)
+        allow(postgres_server.timeline.parent.location).to receive(:location_credential).and_return(credential)
```
enescakir: Do we really need to mock them?

byucesoy: Not really. Removed the mocking.

furkansahin (approved):

shikharbhardwaj (approved):

PR #4800, opened 2026-02-19
Search project and installation associations by UBID

enescakir:
### Search association by UUID in addition to UBID in the admin panel

autoforme uses a select input to search associations, which becomes
problematic when there are many resources (e.g., 1,000+ projects). I
replaced the select input with a text input and added search by UBID
already.

It would be useful to also support searching by UUID and extract this
logic so it can be reused across multiple resources.

### Search project and installation associations by UBID

Since we have hundreds of projects and installations in production,
loading the search page with a very large select input is slow. I
replaced the select input with a text input and enabled searching by
UBID.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -276,7 +276,9 @@ class CloverAdmin < Roda
 
   OBJECT_ASSOC_TABLE_PARAMS = {
     ["GithubInstallation", :runners] => "installation",
-    ["Project", :vms] => "project"
+    ["Project", :vms] => "project",
+    ["Project", :postgres_resources] => "project",
+    ["PostgresResource", :servers] => "resource"
   }.freeze
 
   plugin :autoforme do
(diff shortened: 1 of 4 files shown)


clover_admin.rb
```diff
@@ -310,6 +310,19 @@ def self.object_action(...)
       ds.where(Sequel.cast(column, :text).ilike("%#{ds.escape_like(value)}%"))
     end
 
+    ubid_uuid_grep = lambda do |ds, column, value|
+      uuid = if /\A[a-tv-z0-9]{26}\z/.match?(value)
+        UBID.to_uuid(value)
+      else
+        value
```
jeremyevans: We should check for uuid format here. Otherwise it's easy to trigger a error like: `Sequel::DatabaseError: PG::InvalidTextRepresentation: ERROR:  invalid input syntax for type uuid: "foo" (Sequel::DatabaseError)`.

clover_admin.rb
```diff
@@ -310,6 +310,19 @@ def self.object_action(...)
...
+      end
+      ds.where(column => uuid)
+    end
+
+    ubid_input = lambda do |name|
+      {type: "text", placeholder: "#{name} UBID", maxlength: 26, minlength: 26}
```
jeremyevans: If we want to allow this to accept UUID and not just UBID (which I assume we might if `ubid_uuid_grep` is designed to support UUID)

```suggestion
      {type: "text", placeholder: "#{name} UBID/UUID", maxlength: 36, minlength: 26}
```

enescakir: The UUID branch was used from the `(table)` link, but there’s no harm in allowing it in the search form input as well.

PR #4802, opened 2026-02-19
Add PostgresResource and PostgresServer tables to admin panel

enescakir:
### Skip eager loading on Vm when fetched as an association

AutoForme's :eager option runs every time a record is fetched, not
just on the listing page. PostgresServer eagerly loads [:resource,
:vm], and Vm eagerly loads [:location, :vm_host, :project]. Since
PostgresResource also eagerly loads [:project, :location, :parent],
visiting the PostgresServer list causes duplicate project and location
queries through both associations.

AutoForme supports a block form for :eager that receives the fetch
type, similar to Rodauth's per-request configuration. When a Vm is
fetched as an association (type == :association), the display only
needs name or ubid, so skip the eager loads in that case.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -276,7 +276,9 @@ class CloverAdmin < Roda
 
   OBJECT_ASSOC_TABLE_PARAMS = {
     ["GithubInstallation", :runners] => "installation",
-    ["Project", :vms] => "project"
+    ["Project", :vms] => "project",
+    ["Project", :postgres_resources] => "project",
+    ["PostgresResource", :servers] => "resource"
   }.freeze
 
   plugin :autoforme do
@@ -299,8 +301,10 @@ class CloverAdmin < Roda
       case column
       when :name, :ubid, :invoice_number
         link.call(obj, label: column)
-      when :project, :location, :vm_host, :billing_info
+      when :project, :location, :vm_host, :billing_info, :resource, :parent
         link.call(obj.send(column))
+      when :vm
+        link.call(obj.send(column), label: :ubid)
       when :subtotal, :cost
         "$%0.02f" % (obj.send(column) || 0)
       end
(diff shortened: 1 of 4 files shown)


enescakir:
Tests are failing due to a duplicate query exception.

```
     Sequel::Postgres::AutoParameterizeDuplicateQueryDetection::DuplicateQueries:
       duplicate queries detected:

       times:2
       sql:SELECT * FROM "project" WHERE ("project"."id" = ANY($1::uuid[]))
       backtrace (filtered):
       /Users/enescakir/dev/ubicloud/ubicloud/clover_admin.rb:660:in 'block (2 levels) in <class:CloverAdmin>'
       /Users/enescakir/dev/ubicloud/ubicloud/clover_admin.rb:659:in 'block in <class:CloverAdmin>'
       /Users/enescakir/dev/ubicloud/ubicloud/clover.rb:982:in 'block in <class:Clover>'

       times:2
       sql:SELECT * FROM "location" WHERE ("location"."id" = ANY($1::uuid[]))
       backtrace (filtered):
       /Users/enescakir/dev/ubicloud/ubicloud/clover_admin.rb:660:in 'block (2 levels) in <class:CloverAdmin>'
       /Users/enescakir/dev/ubicloud/ubicloud/clover_admin.rb:659:in 'block in <class:CloverAdmin>'
       /Users/enescakir/dev/ubicloud/ubicloud/clover.rb:982:in 'block in <class:Clover>'
```


I found an interesting behavior in AutoForme. I initially thought the `:eager` option would only run on the listing page, but it turns out it runs every time a record is fetched.

The issue is that `PostgresServer` has eager `[:resource, :vm]`. `Vm` has `eager [:location, :vm_host, :project]`, and `PostgresResource` also has `eager [:project, :location, :parent]`.

As a result when we visit `PostgresServer` list, both `PostgresResource` and `Vm` end up fetching project and location, which causes duplicate queries.

One possible solution is to switch from `:eager` to `:eager_graph`.

Claude also recommended the following change:

```ruby
eager do |type, _request|
  [:location, :vm_host, :project] unless type == :association
end
```

@jeremyevans, I’d like to get your opinion on this.

jeremyevans:
This is an interesting case. `eager_graph` would work, but I think the approach recommended by Claude is better. It shows how, like Rodauth, AutoForme allows for per-request behavior. If you know the information displayed by the association will not need the eagerly loaded information, there is no point in eagerly loaded it. In cases where we are showing associated records, the display is generally name or ubid (and not information from the associated tables), so the change looks good to me.

clover_admin.rb
```diff
@@ -466,6 +470,57 @@ def self.object_action(...)
...
 
+    model PostgresResource do
+      order Sequel.desc(:created_at)
+      eager do |type, _request|
+        [:location, :parent, :project] unless type == :association
+      end
+      columns [:name, :project, :location, :flavor, :target_vm_size, :target_storage_size_gib, :ha_type, :target_version, :parent, :created_at]
+      column_options flavor: {type: "select", options: %w[standard paradedb lantern], add_blank: true},
+        ha_type: {type: "select", options: %w[none async sync], add_blank: true},
+        target_version: {type: "select", options: %w[16 17 18], add_blank: true},
```
jeremyevans: These should probably use reference `PostgresResource` or `Option` constants. Otherwise, at least the version stuff is going to need annual changes. Similar for `PostgresServer` below.

enescakir: I changed with `Option::POSTGRES_VERSION_OPTIONS[PostgresResource::Flavor::STANDARD]`

PR #4805, opened 2026-02-19
Refactor cloud-init YAML generation to use YAML.dump

fdr:
Replace heredoc string interpolation with YAML.dump data structure serialization for write_user_data, meta-data, and network-config. Add comprehensive edge-case tests (YAML-special usernames, multi-line init scripts, empty public keys, multi-NIC, swap on/off).

### DIFF
--- a/rhizome/host/lib/vm_path.rb
+++ b/rhizome/host/lib/vm_path.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "shellwords"
+require "yaml"
 
 class VmPath
   def initialize(vm_name)
@@ -89,6 +90,16 @@ class VmPath
 
     define_method write_method_name do |s|
       write(home(file_name), s)
+    end
+
+    # Method serializing data to YAML and writing, e.g. #write_yaml_user_data
+    yaml_write_method_name = "write_yaml_" + method_name
+    fail "BUG" if method_defined?(yaml_write_method_name)
+
+    define_method yaml_write_method_name do |data, prefix: nil|
+      s = YAML.dump(data, line_width: -1)
+      s.sub!(/\A---/, prefix) if prefix
+      send(write_method_name, s)
     end
   end
 end
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -9,6 +9,7 @@ require "json"
 require "openssl"
 require "base64"
 require "uri"
+require "shellwords"
 require_relative "vm_path"
 require_relative "cloud_hypervisor"
 require_relative "storage_volume"
@@ -34,22 +35,6 @@ class VmSetup
 
   def q_vm
     @q_vm ||= @vm_name.shellescape
-  end
-
-  # YAML quoting
-  def yq(s)
-    require "yaml"
-    # I don't see a better way to quote a string meant for embedding
-    # in literal YAML other than to generate a full YAML document and
-    # then stripping out headers and footers.  Consider the special
-    # string "NO" (parses as boolean, unless quoted):
-    #
-    # > YAML.dump('NO')
-    # => "--- 'NO'\n"
-    #
-    # > YAML.dump('NO')[4..-2]
-    # => "'NO'"
-    YAML.dump(s, line_width: -1)[4..-2]
   end
 
   def vp
@@ -490,10 +475,7 @@ add element inet drop_unused_ip_packets allowed_ipv4_addresses { #{ip_net} }
   end
 
   def cloudinit(unix_user, public_keys, gua, nics, swap_size_bytes, boot_image, dns_ipv4, ipv6_disabled:, init_script: nil)
-    vp.write_meta_data(<<EOS)
-instance-id: #{yq(@vm_name)}
-local-hostname: #{yq(@vm_name)}
-EOS
+    vp.write_yaml_meta_data({"instance-id" => @vm_name, "local-hostname" => @vm_name})
 
     guest_network = subdivide_network(NetAddr.parse_net(gua)).first unless ipv6_disabled
     guest_network_dhcp = "dhcp-range=#{guest_network.nth(2)},#{guest_network.nth(2)},#{guest_network.netmask.prefix_len}" unless ipv6_disabled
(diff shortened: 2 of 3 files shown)


fdr:
Claude's role in this:

This was a mostly unsupervised loop via https://github.com/fdr/walk. It had @jeremyevans 's sketch as a seed, though I editorialized it (I find a `write` method that auto-serializes as yaml a bit too slick, but a separate method name working the same is okay). Claude's main ungratifying job was running it through integration tests with a live counter-party hetzner host using restricted API keys to do integration testing, and used that to capture the output into the tests.

This was eighteen permutations of kinds of input  (way more than a human could reasonably justify) with a full re-run even upon making slight adjustments, since, it's basically a background job, requiring little effort from me.

rhizome/host/lib/vm_setup.rb
```diff
@@ -9,6 +9,7 @@
 require "openssl"
 require "base64"
 require "uri"
+require "shellwords"
```
jeremyevans, line 12: What's the reason for requiring this and using `shelljoin`, as opposed to passing an array of arguments (which is what was done by the previous code)? There doesn't seem to be an advantage to this approach.

rhizome/host/lib/vm_path.rb
```diff
@@ -90,5 +91,15 @@ def home(n)
...
     end
+
+    # Method serializing data to YAML and writing, e.g. #write_yaml_user_data
+    yaml_write_method_name = "write_yaml_" + method_name
+    fail "BUG" if method_defined?(yaml_write_method_name)
+
+    define_method yaml_write_method_name do |data, prefix: nil|
+      s = YAML.dump(data, line_width: -1)
+      s.sub!(/\A---/, prefix) if prefix
+      write(home(file_name), s)
```
jeremyevans: Conceptually DRYer, even if 1 additional character:

```suggestion
      send(write_method_name, s)
```

rhizome/host/lib/vm_setup.rb
fdr: Oh man, the way that it renders in yaml is both what you'd expect and insane, like

```
- nft
- add
- {
- table
[..continue for a whole page, one entry per list entry]
```

You could make it the list form [a, b, c] of yaml, that's an option I could use, but...that's why. It hits you in the face when reading the tests with materialized strings

fdr:
> Since this also only handles the YAML files that used `yq` and not the other YAML files, I don't see an advantage over the patch I prepared for this. Were there any issues found with my patch?
> 
> The new tests look good. This rhizome code was definitely undertested before.

No, I just had it slimmed down, but I don't yet have a strong opinion on how much to batch into the roll-out (which has a material cost these days). You want to get more in in one shot? that seems like it could be fine...

rhizome/host/lib/vm_setup.rb
jeremyevans: `YAML.dump` in Ruby doesn't appear to support dumping lists with `[]`. 

I think that for testing, the long form (one line per argument) is actually preferable. I'd allowed me to easily find a bug in my initial implementation (unremoved `,`), which would have been more difficult if all of the options were in one line.

Forcing `shelljoin` on creation and shell parsing when actually running seems like unnecessary work on both ends. I don't think it causing any significant performance issues, though.

One possibility is using `JSON.pretty_generate` instead of `YAML.dump` (all valid JSON is valid YAML 1.2). That supports more customization options.

jeremyevans:
> > Since this also only handles the YAML files that used `yq` and not the other YAML files, I don't see an advantage over the patch I prepared for this. Were there any issues found with my patch?
> > The new tests look good. This rhizome code was definitely undertested before.
> 
> No, I just had it slimmed down, but I don't yet have a strong opinion on how much to batch into the roll-out (which has a material cost these days). You want to get more in in one shot? that seems like it could be fine...

If you are asking whether we should also handle the non-`yq` YAML files at the same time, I don't see a current need for it, since we have no indications that there are issues in that code. That's why my patch didn't change those files.

rhizome/host/lib/vm_setup.rb
fdr: Oh man, no way, it's pages of stuff per single nft command! a very ungratifying problem to debug if you wanted to debug what NFT was doing (wrong)

fdr: we don't need shelljoin though, just a string. that was an oops.

fdr: oh wait no, we do, because the charactres in the original are complex. that's more a caller thing to get the literal string mode.

fdr: here's what someone trying to figure out why their nft expression is a tiny bit off has to look forward to:

```
     bootcmd:                                                                                                                                                                           
     - - nft                                                                                                                                                                            
       - add                                                                                                                                                                            
       - table                                                                                                                                                                          
       - ip6                                                                                                                                                                            
       - filter                                                                                                                                                                         
     - - nft                                                                                                                                                                            
       - add                                                                                                                                                                            
       - chain                                                                                                                                                                          
       - ip6                                                                                                                                                                            
       - filter                                                                                                                                                                         
       - output                                                                                                                                                                         
       - "{"                                                                                                                                                                            
       - type                                                                                                                                                                           
       - filter                                                                                                                                                                         
       - hook                                                                                                                                                                           
       - output                                                                                                                                                                         
       - priority                                                                                                                                                                       
       - '0'                                                                                                                                                                            
       - ";"                                                                                                                                                                            
       - "}"                                                                                                                                                                            
     - - nft                                                                                                                                                                            
       - add                                                                                                                                                                            
       - rule                                                                                                                                                                           
       - ip6                                                                                                                                                                            
       - filter                                                                                                                                                                         
       - output                                                                                                                                                                         
       - ip6                                                                                                                                                                            
       - daddr                                                                                                                                                                          
       - fd00:0b1c:100d:5AFE::/64                                                                                                                                                       
       - meta                                                                                                                                                                           
       - skuid                                                                                                                                                                          
       - "!="                                                                                                                                                                           
       - '0'                                                                                                                                                                            
       - tcp                                                                                                                                                                            
       - flags                                                                                                                                                                          
       - syn                                                                                                                                                                            
       - reject                                                                                                                                                                         
       - with                                                                                                                                                                           
       - tcp                                                                                                                                                                            
       - reset 
   ```

fdr: experimenting with json mode now with a custom formatter to get [a, few, elements]

fdr: eh, I think we lose too much. it makes that better but makes other legibility (e.g. multiline strings) worse.

fdr:
why does my job keep getting cancelled?.. blah

fdr:
this got through vm stuff but died into postgres stuff, so I think e2e is just busted right now/flaky due to timeout, taking well over an hour.

PR #4808, opened 2026-02-19
Fix N+1 issues on Vm/PrivateSubnet index pages

jeremyevans:


routes/project/vm.rb
```diff
@@ -9,7 +9,7 @@ class Clover
         vm_list_api_response(dataset)
       else
         @vms = dataset
-          .eager(:semaphores, :assigned_vm_address, :vm_storage_volumes, :location)
+          .eager(:strand, :semaphores, :assigned_vm_address, :vm_storage_volumes, :location)
```
enescakir: It’s already fixed for the api case. Should we move this logic to a central place?

https://github.com/ubicloud/ubicloud/blob/7fe14dc95c5ad5cf53e319f77bad6339c90e1c46/helpers/vm.rb#L14

jeremyevans: We can do similar hoisting of the `eager` calls here. I wonder if we should move the route logic into the helper for VM, to make it similar to PrivateSubnet.

enescakir: Makes sense. It also helps when we eagerly load new associations for the API or web, seeing them all in one view makes it less likely to miss the other one.

jeremyevans: I refactored to add `vm_list`, using a similar approach to `private_subnet_list`.

PR #4813, opened 2026-02-19
Add Show Job Log action for GithubRepository in admin panel

enescakir:
- **Add type field to admin object actions for different interaction modes**

ObjectAction gains a type field that controls how an action is
presented on the object page and how its request lifecycle works:

- normal (default): renders a link on the object page. GET shows a
  confirmation page with inputs. POST executes the action.
- form: renders a POST form directly on the object page, skipping the
  confirmation step.
- direct: renders a link on the object page. GET executes the action
  and redirects to the returned URL.

Strand "schedule" and GithubRunner "provision" are switched to :form
since they need no confirmation. The action list is also restyled so
that form submit buttons visually match the existing action links.

- **Replace Invoice extras template with a direct object action**

The Invoice "Download PDF" link was rendered via an extras ERB template
that called generate_download_link on every page load. This moves it to
an ObjectAction with direct type, so the S3 presigned URL is only
generated when the user clicks the link. If link generation fails, a
CloverError is raised instead of silently hiding the button.

- **Support HTML responses from admin object actions**

Object actions previously always redirected back to the object page with
a flash notice. Actions can now return an HTML string to render inline
within the admin layout. Actions that return nil or other values retain
the original behavior.

- **Add Show Job Log action for GithubRepository in admin panel**

Fetches a presigned log download URL from the GitHub API via
workflow_run_job_logs and renders it as a clickable link. A POST-based
action with confirmation is used because CSP blocks cross-domain
redirects from POST responses.

- **Use keyword arguments for ObjectAction.define**

The positional flash and params arguments made it easy to pass them
in the wrong order and required a nil flash for actions that return
HTML instead of redirecting. Switching to keyword arguments makes
call sites self-documenting and lets actions omit flash when they
render an inline response.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -658,9 +658,10 @@ class CloverAdmin < Roda
         if (actions = OBJECT_ACTIONS[@obj.class.name])
           r.is actions.keys do |key|
             action = actions[key]
+            action_type = action.type
 
-            r.get do
-              if action.type == :direct
+            r.get(action_type != :form) do
+              if action_type == :direct
                 url = action.call(@obj) || fail(CloverError.new(400, "InvalidRequest", "Action link is not available"))
                 r.redirect url
               end
@@ -669,10 +670,10 @@ class CloverAdmin < Roda
               view("object_action")
             end
 
-            r.post do
+            r.post(action_type != :direct) do
               params = action.params.map { |k, v| typecast_params.send(v.is_a?(Hash) ? v[:typecast] : v, k.to_s) }
               result = action.call(@obj, *params)
-              if action.type == :content
+              if action_type == :content
                 view(content: result)
               else
                 flash["notice"] = action.flash


clover_admin.rb
```diff
@@ -651,9 +651,16 @@ def self.object_action(...)
 
             r.post do
               params = action.params.map { |k, v| typecast_params.send(v.is_a?(Hash) ? v[:typecast] : v, k.to_s) }
-              action.call(@obj, *params)
-              flash["notice"] = action.flash
-              r.redirect("/model/#{@obj.class}/#{ubid}")
+              result = action.call(@obj, *params)
+              case result
+              when /\Ahttps?:\/\//, /\A\//
+                r.redirect(result)
```
jeremyevans: For the `/` case, should we set `flash["notice"]`?

clover_admin.rb
```diff
@@ -167,9 +167,9 @@ def available_classes
     end
   end
 
-  ObjectAction = Data.define(:label, :flash, :params, :action) do
-    def self.define(label, flash, params = {}, &action)
-      new(label, flash, params.dup.freeze, action)
+  ObjectAction = Data.define(:label, :flash, :params, :confirmation, :action) do
+    def self.define(label, flash, params = {}, confirmation: true, &action)
```
jeremyevans: Mixing optional positional and keyword arguments is generally a poor design. I would switch `params` to a keyword argument. You might also switch `flash` to a keyword argument, since you are adding actions that don't have flash notices.

jeremyevans (changes requested):
Making state changes such as strand scheduling or provisioning a spare runner on a GET request is not something I can approve. If you would like a directly executed button on the object page, show a POST form on the object page instead of a link. Using GET for downloading invoice PDFs and job logs seems fine. 

Instead of `confirmation`, maybe it would be better to have a `type` keyword which could default to `:link` (normal behavior of link to separate page with form) and support`:form` (POST form on object page, for strand scheduling and spare runner provisioning), and `:direct` (link that directly navigates to another page, for invoice PDFs and job logs).

enescakir:
@jeremyevans, I refactored it, which makes some parts cleaner. However, I couldn’t use :direct for the workflow job link. We rely on a form because we need the job ID information.

I was able to trigger a redirect when the form is submitted instead of rendering a new button, but the browser blocks redirects to external sites via form submission. We could update the CSP rules to allow it, but I don’t think it’s worth making that change.

PR #4817, opened 2026-02-20
Fix backfill prog

furkansahin:
## Fix selecting a random subnet to backfill
We should at least make sure the subnet is associated to one of our
nics.

### DIFF
--- a/prog/vnet/aws/backfill_aws_subnets.rb
+++ b/prog/vnet/aws/backfill_aws_subnets.rb
@@ -48,7 +48,11 @@ class Prog::Vnet::Aws::BackfillAwsSubnets < Prog::Base
 
     fail "No subnets found in VPC #{private_subnet_aws_resource.vpc_id}" if subnets.empty?
 
-    subnet = subnets.first
+    used_subnet_ids = private_subnet.nics.map { it.nic_aws_resource.subnet_id }.compact.uniq
+    subnets = subnets.select { |subnet| used_subnet_ids.include?(subnet.subnet_id) }
+
+    # this assumes there is only non-ha postgres resources
+    subnet = subnets.min_by(&:cidr_block)
     az_suffix = subnet.availability_zone.delete_prefix(location.name)
     location_aws_az = find_location_aws_az(az_suffix)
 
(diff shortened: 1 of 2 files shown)


prog/vnet/aws/backfill_aws_subnets.rb
```diff
@@ -48,6 +48,10 @@ def self.assemble(private_subnet_id)
 
     fail "No subnets found in VPC #{private_subnet_aws_resource.vpc_id}" if subnets.empty?
 
+    used_subnet_ids = private_subnet.nics.map { it.nic_aws_resource.subnet_id }.compact.uniq
+    subnets = subnets.select { |subnet| used_subnet_ids.include?(subnet.subnet_id) }
+
+    # this assumes there is only non-ha postgres resources
```
bsatzger, line 54: would it be safer to fail if we find more than one resource?

bsatzger: if that is not viable, should we
- log when there is more than a single resource
- pick the subnet from the set of subnets in a more deterministic way, e.g., the oldest/newest/...

prog/vnet/aws/backfill_aws_subnets.rb
```diff
@@ -48,6 +48,10 @@ def self.assemble(private_subnet_id)
 
     fail "No subnets found in VPC #{private_subnet_aws_resource.vpc_id}" if subnets.empty?
 
+    used_subnet_ids = private_subnet.nics.map { it.nic_aws_resource.subnet_id }.compact.uniq
+    subnets = subnets.select { |subnet| used_subnet_ids.include?(subnet.subnet_id) }
```
enescakir, line 52: Should we move this above the `subnets.empty?` check? Could it become empty after the `.select` operation?

furkansahin: That is not viable, if we have the record of a VPC, we must have at least 1 subnet created there.

prog/vnet/aws/backfill_aws_subnets.rb
furkansahin: There cannot be more than a single resource
I ordered it according to the cidr_block because unfortunately, AWS entity doesn't have the `created_at`.

PR #4821, opened 2026-02-20
Simplify GitHub runner e2e tests to use a single workflow run

enescakir:
Previously each test case triggered and tracked a separate workflow
(test_2204.yml, test_2404.yml, etc.), complicating the prog with
per-case iteration in trigger, check, and cancel steps. A single
workflow dispatched with the runner labels as an input can spawn
multiple jobs internally, so we only need to track one run.

The test case config no longer carries per-case workflow details; the
prog derives the required runner labels from the test case names and
passes them to the workflow.

Use the new return_run_details parameter in the workflow dispatch API
to get the run ID directly from the response, eliminating the need to
poll workflow runs and search by creation time and name. This removes
the 30-second sleep and the created_at tracking from the stack frame.

https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/

The branch is pinned to "enes/simply-tests" temporarily because the
workflow file on the target repository is not yet on main. It will
be switched to main once existing branches are rebased.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -165,6 +165,15 @@ jobs:
       if: always()
       run: grep -h -v -E "sleep_duration_sec|monitor.1" foreman.log
 
+    - name: Extract GitHub workflow run
+      if: always()
+      continue-on-error: true
+      run: |
+        html_url="$(grep 'github_workflow_run' foreman.log | cut -d'|' -f2 | jq -r '.github_workflow_run.html_url' | head -1)"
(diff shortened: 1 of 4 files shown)


enescakir:
Simplified workflow: https://github.com/tahcloud/github-e2e-tests-metal/pull/1

enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22459397362

PR #4822, opened 2026-02-20
Fix strand exception logging

jeremyevans:
Previously, if a respirate strand raised an exception, it was not caught by Strand#run, it was caught by
Dispatcher#run_strand. Unfortunately, the strand reported by Dispatcher#run_strand may not be the strand that actually caused the error, due to parent strands running child strands in Prog#reap.

Introduce Strand::RunError for wrapping the exceptions. If a strand raises an exception, it will be logged with the strand that actually caused the exception, and then wrapped in RunError and reraised. Parent strands that run child strands that raise RunError pass it through without rewrapping it. Dispatcher no longer logs RunError exceptions, since they have already been logged.

There is a fair amount of internal exceptions raised in Strand and Prog::Base. Use Strand::InternalError instead of RuntimeError for these, and don't wrap those errors. These errors are programming bugs and should never be hit in production code.

Avoid RunError wrapping in pry, as it could complicate debugging.

### DIFF
--- a/model/strand.rb
+++ b/model/strand.rb
@@ -245,7 +245,7 @@ SQL
       e
     rescue Prog::Base::Hop => hp
       last_changed_at = Time.parse(top_frame["last_label_changed_at"])
-      Clog.emit("hopped", {strand_hopped: {duration: Time.now - last_changed_at, from: prog_label, to: "#{hp.new_prog}.#{hp.new_label}"}})
+      Clog.emit("hopped", {strand_hopped: {strand: ubid, duration: Time.now - last_changed_at, from: prog_label, to: "#{hp.new_prog}.#{hp.new_label}"}})
       top_frame["last_label_changed_at"] = Time.now.to_s
       modified!(:stack)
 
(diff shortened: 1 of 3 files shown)


model/strand.rb
```diff
@@ -275,14 +275,14 @@ def unsynchronized_run
...
-      Clog.emit("exception terminates strand run", [self, Util.exception_to_hash(ex, into: {strand_error: {duration:, from: prog_label}})])
+      Clog.emit("exception terminates strand run", Util.exception_to_hash(ex, into: {strand_error: {strand: ubid, duration:, from: prog_label}}))
       raise RunError.new(self)
     else
       raise InternalError, "BUG: Prog #{prog}##{label} did not provide flow control"
     end
   ensure
     duration = Time.now - start_time
-    Clog.emit("finished strand", [self, {strand_finished: {duration:, prog_label:}}]) if duration > 1
+    Clog.emit("finished strand", {strand_finished: {strand: ubid, duration:, prog_label:}}) if duration > 1
```
enescakir: `strand` also contains more information than just the `ubid`, such as the `try` count. This is useful for identifying jobs with a high number of retries. We can also add it.

jeremyevans:
> re: the second commit, we should probably try to make the representation logged prefer ubids all around, e.g. so the default dump gives you the most usable (defined as: when you double click on things and paste them around, you get the right results) representation.

I've added fixing `Clog.serialize_model` to use ubids to my todo list. I'll tackle that in a separate PR.

model/strand.rb
jeremyevans: I added `try` information.

PR #4824, opened 2026-02-21
Add `View in Clover` links to admin panel

serprex:
Useful for redistributors to easily drill down during investigation


jeremyevans (commented):
This would not make sense for Ubicloud, because administrative users in Ubicloud have no direct access to customer projects in the console.

If it's critical for your usage, then it should be behind a Config setting that is false by default.

Also, if you are adding a new feature, you should be adding tests for the feature. In this case, you would check that the link exists with the expected path (assuming the Config setting is enabled).

PR #4825, opened 2026-02-22
Simplify read replica promotion

byucesoy:
I'm making some refactoring on failover code path to make it more resilient about edge cases. This PR is part of the refactoring work that I can extract and send to review independently.

**Set syncronization_status during timeline switch**
When a Postges server is switched to a different timeline, it starts as
the leader of the new timeline. Therefore, by definition, it is in sync.

**Simplify read replica promotion**
The read replica promotion starts with a semaphore increment on the
resource, which then triggers semaphore increment on the server, and
finally the server goes through taking_over label, which is part of the
failover process and arguably one of the most complex parts of the
codebase. On the flip side, read replica promotion only requires some
metadata update and re-configure on the server.

This commit removes all the indirection and the need for reusing the
failover related code path. We do the metadata update on the endpoint
and increment configure semaphore on the server.

**Rename to read replica promotion endpoint to promote-read-replica**
Promote is a overused word and has a special meaining in the context of
PostgreSQL. Using it alone for read replica promotion is confusing. The
new name is more descriptive and accurate. It is also more consistent
with the naming of the CLI command.

Unfortunatelly, this is a breaking change for the API, but this endpoint
is not used frequently. I think we are safe to make this change now,
before we have more users of this endpoint.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1338,6 +1338,22 @@ paths:
       - $ref: '#/components/parameters/postgres_database_reference'
     post:
       operationId: promotePostgresDatabase
+      summary: 'Promote a PostgreSQL read replica (deprecated, use promote-read-replica)'
+      responses:
+        '200':
+          $ref: '#/components/responses/PostgresDatabase'
+        default:
+          $ref: '#/components/responses/Error'
+      deprecated: true
+      tags:
+        - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/promote-read-replica':
+    parameters:
+      - $ref: '#/components/parameters/project_id'
+      - $ref: '#/components/parameters/location'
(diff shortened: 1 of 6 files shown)


openapi/openapi.yml
```diff
@@ -1331,13 +1331,13 @@ paths:
           $ref: '#/components/responses/Error'
       tags:
         - Postgres Database
-  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/promote':
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/promote-read-replica':
```
serprex: can we have period where both endpoints are accepted? this way we can transition our multiple systems

byucesoy: Sure. I'll add that before merging this.

byucesoy:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22607620134

PR #4826, opened 2026-02-23
Support Ubiblk v0.4.0.

pykello:
Notable changes in ubiblk v0.4.0:

- Archive export & stripe sources
- Remote stripe server & stripe sources
- Security fixes
- ublk backend
- Config v2
- Disk metadata format improvements

Majority of changes in the commit is to support config v2, which is documented at: https://github.com/ubicloud/ubiblk/blob/main/docs/config.md

In config v1, the key encryption key (KEK) path was passed as a command line argument. In v2, it's embedded in the config file instead.

We use the KEK in two places: `init-metadata` and the systemd unit. Previously, `init-metadata` received it via stdin and the systemd unit received it via a named pipe. Since v2 embeds the KEK path in the config file, and we use the same config file for both, we've switched both to use named pipes.

We could have kept stdin for v1's `init-metadata`, but that would mean additional branching in the code. For simplicity, this commit also switches v1 `init-metadata` to named pipes.

Named pipes were chosen over two alternatives:

- stdin: can't easily pass stdin to a systemd unit, and using /dev/shm to materialize it as a file is worse for security (root directory traversal).
- RPC: the current implementation of RPC server starts after device setup, so using it for KEK delivery would require reordering startup, adding complexity for little benefit.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -156,7 +156,7 @@ jobs:
         SMTP_TLS: false
         STRIPE_PUBLIC_KEY: ${{ secrets.STRIPE_PUBLIC_KEY }}
         STRIPE_SECRET_KEY: ${{ secrets.STRIPE_SECRET_KEY }}
-        VHOST_BLOCK_BACKEND_VERSION: v0.3.1
+        VHOST_BLOCK_BACKEND_VERSION: v0.4.0
         OPERATOR_SSH_PUBLIC_KEYS: ${{ secrets.OPERATOR_SSH_PUBLIC_KEYS }}
       run: |
         set -o pipefail
--- a/prog/storage/setup_vhost_block_backend.rb
+++ b/prog/storage/setup_vhost_block_backend.rb
@@ -4,6 +4,8 @@ class Prog::Storage::SetupVhostBlockBackend < Prog::Base
   subject_is :sshable, :vm_host
 
   SUPPORTED_VHOST_BLOCK_BACKEND_VERSIONS = [
+    ["v0.4.0", "x64"],
+    ["v0.4.0", "arm64"],
     ["v0.3.1", "x64"],
     ["v0.3.1", "arm64"],
     ["v0.2.2", "x64"],


rhizome/host/lib/storage_key_encryption.rb
```diff
@@ -3,12 +3,26 @@
...
     @key_encryption_cipher = key_encryption_cipher
   end
 
+  # Wrap a plaintext with AES-256-GCM:
+  # [12-byte nonce || ciphertext || 16-byte tag]
+  def self.aes256gcm_encrypt(key, auth_data, plaintext)
+    cipher = OpenSSL::Cipher.new("aes-256-gcm")
+    cipher.encrypt
+    cipher.key = key
+    nonce = SecureRandom.random_bytes(12)
```
jeremyevans, line 19: 96-bit purely random nonce is probably fine, but personally I prefer a partial time component, such as 32-bit unsigned unix epoch timestamp followed by 64-bit random. Up to you.

rhizome/host/lib/storage_key_encryption.rb
```diff
@@ -3,12 +3,26 @@
...
+    cipher.iv = nonce
+    cipher.auth_data = auth_data
+    ciphertext = cipher.update(plaintext) + cipher.final
+    nonce + ciphertext + cipher.auth_tag
```
jeremyevans: Since `nonce` is mutable, better for performance to do:

```suggestion
    nonce << cipher.update(plaintext) << cipher.final << cipher.auth_tag
```

Makes more difference if `plaintext` is large, but even in small cases avoids 3 string allocations.

rhizome/host/lib/storage_volume.rb
```diff
@@ -99,40 +109,75 @@ def write_new_file(path, user)
...
+      write_config_file(sp.vhost_backend_config, v2_main_toml)
+    else
+      config = vhost_backend_config(encryption_key, key_wrapping_secrets)
+      write_config_file(sp.vhost_backend_config, config.to_yaml)
+    end
+  end
 
-    write_new_file(config_path, @vm_name) do |file|
-      file.write(config.to_yaml)
+  def write_config_file(path, content)
```
jeremyevans, line 125: Not an issue with this PR, but is there a risk of a partially written file causing problems? Would it be better to write a temp file and the rename it to the actual file name?

rhizome/host/lib/storage_volume.rb
```diff
@@ -99,40 +109,75 @@ def write_new_file(path, user)
...
 
     write_new_file(metadata_path, @vm_name) do |file|
       file.truncate(8 * 1024 * 1024)
     end
 
-    r "#{vhost_backend.init_metadata_path.shellescape} -s #{@stripe_sector_count_shift}  --config #{config_path.shellescape} #{kek_arg}", stdin: kek_yaml
+    if @encrypted
+      vhost_backend_create_encrypted_metadata(key_wrapping_secrets)
+    else
+      r "#{vhost_backend.init_metadata_path.shellescape} -s #{@stripe_sector_count_shift} --config #{config_path.shellescape}"
```
jeremyevans: `r` now supports separate arguments, and it is recommended to use separate arguments instead of shell escaping.

```suggestion
      r(vhost_backend.init_metadata_path, "-s", @stripe_sector_count_shift.to_s, "--config", config_path)
```

rhizome/host/lib/storage_volume.rb
```diff
@@ -99,40 +109,75 @@ def write_new_file(path, user)
...
+        "sudo", "-u", @vm_name,
+        vhost_backend.init_metadata_path,
+        "-s", @stripe_sector_count_shift.to_s,
+        "--config", sp.vhost_backend_config
+      ]
+
+      unless use_config_v2?
+        # Unlike v2, where the kek path is explicitly included in the TOML
+        # config, we need to pass the kek pipe path as a CLI arg in v1.
+        cmd += ["--kek", kek_pipe]
```
jeremyevans: `cmd` is mutable, better it just append to it:

```suggestion
        cmd.push("--kek", kek_pipe)
```

rhizome/host/lib/storage_volume.rb
```diff
@@ -274,13 +319,125 @@ def vhost_backend_config(encryption_key, key_wrapping_secrets)
...
   end
 
+  def v2_main_toml
+    sections = []
+    sections << v2_include_section
+    sections << v2_device_section
+    sections << v2_tuning_section
+    sections << v2_encryption_section if @encrypted
+    # unencrypted disks must be explicitly allowed in config v2.
+    sections << v2_danger_zone_section unless @encrypted
```
jeremyevans:
```suggestion
    if @encrypted
      sections << v2_encryption_section
    else
      # unencrypted disks must be explicitly allowed in config v2.
      sections << v2_danger_zone_section
    end
```

rhizome/host/lib/storage_volume.rb
```diff
@@ -274,13 +319,125 @@ def vhost_backend_config(encryption_key, key_wrapping_secrets)
...
+  def v2_device_section
+    lines = ["[device]"]
+    lines << "data_path = #{toml_str(disk_file)}"
+    lines << "metadata_path = #{toml_str(sp.vhost_backend_metadata)}" if @image_path
+    lines << "vhost_socket = #{toml_str(vhost_sock)}"
+    lines << "rpc_socket = #{toml_str(sp.rpc_socket_path)}"
+    lines << "device_id = #{toml_str(@device_id)}"
+    lines.join("\n") + "\n"
+  end
+
```
jeremyevans, line 357: I recommend an approach like this, which I think think is overall simpler:

```suggestion
  def v2_device_section
    hash = {
      "data_path" => disk_file,
      "vhost_socket" =>vhost_sock,
      "rpc_socket" => sp.rpc_socket_path,
      "device_id" => @device_id
    }
    hash["metadata_path"] = sp.vhost_backend_metadata) if @image_path
    toml_section("device", hash)
  end
  
  def toml_section(name, hash)
    "[#{name}]\n#{hash.map{ |k, v| "k = #{v.is_a?(String) ? toml_str(v) : v}" }}.join("\n")}\n"
  end
  
```

Both for this method and for following methods. However, it's up to you.

rhizome/host/lib/storage_volume.rb
```diff
@@ -274,13 +319,125 @@ def vhost_backend_config(encryption_key, key_wrapping_secrets)
...
+  def toml_str(value)
+    # From TOML specs:
+    # > Basic strings are surrounded by quotation marks ("). Any Unicode
+    # > character may be used except those that must be escaped: quotation mark,
+    # > backslash, and the control characters other than tab (U+0000 to U+0008,
+    # > U+000A to U+001F, U+007F).
+    #
+    # See https://toml.io/en/v1.0.0#string
+    escaped = value.gsub(/[\x00-\x08\x0A-\x1F\x7F"\\]/) do |ch|
```
jeremyevans: `gsub` can take a hash of replacements. You can give the hash a default value:

```ruby
  h = {"\b"=>"\\b"}
  h.default_proc = proc{|_,ch| format('\\u%04X', ch.ord)}
  "\b\x1f\b".gsub(/[\x00-\x08\x0A-\x1F\x7F"\\]/, h)
  # => "\\b\\u001F\\b"
```

I think having a hash in a constant for the replacements would make this simpler, but up to you.

rhizome/host/lib/storage_volume.rb
```diff
@@ -321,25 +478,36 @@ def vhost_backend_start(key_wrapping_secrets)
...
     end
 
-    begin
-      kek_pipe = sp.kek_pipe
-      rm_if_exists(kek_pipe)
-      File.mkfifo(kek_pipe, 0o600)
-      FileUtils.chown @vm_name, @vm_name, kek_pipe
-
+    with_kek_pipe do |kek_pipe|
       r "systemctl start #{q_vhost_user_block_service}"
```
jeremyevans: Unrelated, but since we are here.

```suggestion
      r "systemctl", "start", vhost_user_block_service
```

Similar change could be made in other cases calling `systemctl` `start`/`stop`. Up to you.

rhizome/host/lib/storage_volume.rb
```diff
@@ -321,25 +478,36 @@ def vhost_backend_start(key_wrapping_secrets)
...
+  end
 
-      Timeout.timeout(5) do
-        kek_yaml = vhost_backend_kek(key_wrapping_secrets).to_yaml
-        File.write(kek_pipe, kek_yaml)
+  def write_kek_to_pipe(kek_pipe, payload, timeout_sec: KEK_PIPE_WRITE_TIMEOUT_SEC)
+    Timeout.timeout(timeout_sec) do
+      File.open(kek_pipe, File::WRONLY) do |file|
+        file.write(payload)
+        file.flush
```
jeremyevans: Not sure this `flush` is necessary, as the implicit close of the pipe when the block returns should ensure the data is flushed.

rhizome/host/lib/vhost_block_backend.rb
```diff
@@ -4,26 +4,36 @@
...
 class VhostBlockBackend
+  SHA256_BY_VERSION_AND_ARCH = {
+    ["v0.4.0", "x64"] => "c74f1d4bef624d5f46e3763d6d63ac79ca3e8f3fccd1696840266398834d53d0",
+    ["v0.4.0", "arm64"] => "65dd7354b087a07fca7d35bca00014532fd636e7aae838e32d87ee0d37284fad",
+    ["v0.3.1", "x64"] => "3b4a6d3387a8da7c914d85203955c0a879168518aed76679a334070403630262",
+    ["v0.3.1", "arm64"] => "d7cd297468569a0fa197d48eb7d21b64aea9598895d1b5b97da8bec5e307d57b",
+    ["v0.2.2", "x64"] => "f5b7b2b88fa18e5070ff319b15363aed671e496d9f6cccec3bbcc48a6f38a44a",
+    ["v0.2.2", "arm64"] => "7f4a5818fdab4e7524855096352d9ceaa038ff254de2b52c88d491f76a05686f"
+  }.freeze
+
```
jeremyevans, line 16: Not that these would be modified, but it's best to freeze them anyway:

```suggestion
SHA256_BY_VERSION_AND_ARCH.each_key(&:freeze)
```

rhizome/host/lib/vhost_block_backend.rb
```diff
@@ -4,26 +4,36 @@
...
       fail "Unsupported version: #{@version}, #{Arch.sym}"
     end
   end
 
   def url
-    "https://github.com/ubicloud/ubiblk/releases/download/#{@version}/vhost-backend-#{Arch.sym}.tar.gz"
+    if @v0_4_or_later
+      "https://github.com/ubicloud/ubiblk/releases/download/#{@version}/ubiblk-#{Arch.sym}.tar.gz"
+    else
+      "https://github.com/ubicloud/ubiblk/releases/download/#{@version}/vhost-backend-#{Arch.sym}.tar.gz"
```
jeremyevans:
```suggestion
    dir = "https://github.com/ubicloud/ubiblk/releases/download/#{@version}"
    if @v0_4_or_later
      "#{dir}/ubiblk-#{Arch.sym}.tar.gz"
    else
      "#{dir}/vhost-backend-#{Arch.sym}.tar.gz"
```

rhizome/host/spec/storage_volume_spec.rb
```diff
@@ -450,7 +528,7 @@
       expect(unencrypted_vhost_sv).to receive(:write_new_file).with(metadata_path, "test").and_yield(f)
       expect(f).to receive(:truncate).with(8 * 1024 * 1024)
       expect(unencrypted_vhost_sv).to receive(:sync_parent_dir).with(metadata_path)
-      expect(unencrypted_vhost_sv).to receive(:r).with(/.*init-metadata.*--config \/var\/storage\/test\/2\/vhost-backend.conf $/, stdin: "")
+      expect(unencrypted_vhost_sv).to receive(:r).with(/.*init-metadata.*--config \/var\/storage\/test\/2\/vhost-backend.conf$/)
```
jeremyevans: Only use `$` in a Ruby regexp if you are 100% sure you want end-of-line and not end-of-string:

```suggestion
      expect(unencrypted_vhost_sv).to receive(:r).with(/.*init-metadata.*--config \/var\/storage\/test\/2\/vhost-backend.conf\z/)
```

jeremyevans (approved):
Looks good. I left a number of comments for potential improvements.

rhizome/host/lib/storage_key_encryption.rb
pykello: As we discussed internally, we don't reuse the key over time.

PR #4827, opened 2026-02-23
Append triggered E2E workflow run URL to the /run-e2e comment

enescakir:
GitHub's workflow dispatch API now supports returning run details via the return_run_details parameter [1]. Use this to get the run URL directly from the response, then edit the original /run-e2e comment to append it. This gives immediate visibility into the triggered run.

[1] https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/


enescakir:
example: https://github.com/enescakir/ubicloud/pull/21#issuecomment-3944366748

CleanShot 2026-02-23 at 15 08 44@2x

PR #4828, opened 2026-02-23
Flatten mesh availability output for log compatibility

mohi-kalantari:
The integrated log system supports only 3 levels of nesting and a limited number of unique keys. The hash-keyed-by-name structures (pod_errors, external_errors, mtr_results) in check_mesh_availability create unbounded unique keys and deep nesting, which the log system cannot index or query effectively.

Convert these three fields from hashes keyed by target name to arrays with the name stored as a field inside each element. This reduces the nesting depth by one level and eliminates dynamic keys, making the log output compatible with the system's constraints.

### DIFF
--- a/model/kubernetes/kubernetes_node.rb
+++ b/model/kubernetes/kubernetes_node.rb
@@ -70,7 +70,7 @@ class KubernetesNode < Sequel::Model
     status = JSON.parse(file_content)
     pods_status = status["pods"]
     external_status = status["external_endpoints"]
-    mtr_results = status["mtr_results"]
+    mtr_results = status["mtr_results"]&.map { |name, v| v.merge("name" => name) }
     api_error = status["api_error"]
 
     if api_error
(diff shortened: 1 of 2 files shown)


model/kubernetes/kubernetes_node.rb
```diff
@@ -70,7 +70,7 @@ def check_mesh_availability(ssh_session = nil)
     status = JSON.parse(file_content)
     pods_status = status["pods"]
     external_status = status["external_endpoints"]
-    mtr_results = status["mtr_results"]
+    mtr_results = status["mtr_results"]&.map { |name, v| v.merge("name" => name) }
```
aamederen, line 73: I don't have a strong opinion on this, but in this case, stringifying the output (with to_s) feels like a safer option.

mohi-kalantari: It would reduce the readability substantially. In case there is an issue in the report structure, we will get paged as the exceptions would make the resource appear down.

PR #4834, opened 2026-02-23
Handle failed rsync during CSI data migration with retry

mohi-kalantari:
When daemonizer2 reports a failed rsync copy during volume migration, the old behavior raised a hard error immediately, leaving the daemonizer2 unit in a failed state. Subsequent kubelet retries would see the same stale "Failed" status and never recover — the volume would be stuck permanently.

Now on failure, the daemonizer2 unit is cleaned and a CopyNotFinishedError is raised so kubelet can retry the copy from scratch. A retry counter is persisted as a PV annotation (csi.ubicloud.com/migration-retry-count) so it survives pod restarts. After 3 failed attempts, the unit is left intact for log inspection and a hard error is raised directing the user to contact support. The annotation is cleaned up on successful copy completion.

### DIFF
--- a/kubernetes/csi/Gemfile.lock
+++ b/kubernetes/csi/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ubi-csi (0.1.0)
+    ubi-csi (0.6.0)
       base64
       grpc (~> 1.76)
       grpc-tools (~> 1.76)
--- a/kubernetes/csi/lib/ubi_csi.rb
+++ b/kubernetes/csi/lib/ubi_csi.rb
@@ -5,7 +5,7 @@ require "csi_pb"
 require "csi_services_pb"
 
 module Csi
-  VERSION = "0.1.0"
+  VERSION = "0.6.0"
 end
 
 require_relative "ubi_csi/identity_service"
--- a/kubernetes/csi/spec/csi/v1/identity_service_spec.rb
+++ b/kubernetes/csi/spec/csi/v1/identity_service_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe Csi::V1::IdentityService do
       request = Csi::V1::GetPluginInfoRequest.new
       response = service.get_plugin_info(request, call)
       expect(response.name).to eq("csi.ubicloud.com")
-      expect(response.vendor_version).to eq("0.1.0")
+      expect(response.vendor_version).to eq("0.6.0")
     end
   end
 
--- a/kubernetes/csi/ubi-csi.gemspec
+++ b/kubernetes/csi/ubi-csi.gemspec
@@ -2,7 +2,7 @@
 
 Gem::Specification.new do |spec|
   spec.name = "ubi-csi"
-  spec.version = "0.1.0"
+  spec.version = "0.6.0"
   spec.authors = ["Ubicloud"]
   spec.email = ["support@ubicloud.com"]
 
--- a/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
@@ -22,7 +22,7 @@ spec:
       priorityClassName: system-node-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.5.0
+          image: ubicloud/ubicsi:v0.6.0
           imagePullPolicy: IfNotPresent
           args: ["node"]
           env:
--- a/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
@@ -38,7 +38,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.5.0
+          image: ubicloud/ubicsi:v0.6.0
           imagePullPolicy: IfNotPresent
           args: ["controller"]
           env:


kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -220,14 +222,25 @@ def migrate_pvc_data(req_id, client, pvc, req)
...
           copy_command = ["rsync", "-az", "--inplace", "--compress-level=9", "--partial", "--whole-file", "-e", "ssh -T -c aes128-gcm@openssh.com -o Compression=no -x -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /home/ubi/.ssh/id_ed25519", "ubi@#{old_node_ip}:#{old_data_path}", current_data_path]
           run_cmd_output("nsenter", "-t", "1", "-a", "/home/ubi/common/bin/daemonizer2", "run", daemonizer_unit_name, *copy_command, req_id:)
           raise CopyNotFinishedError, "Old PV data is not copied yet"
         when "InProgress"
           raise CopyNotFinishedError, "Old PV data is not copied yet"
         when "Failed"
-          raise "Copy old PV data failed"
+          retry_count = Integer(pv.dig("metadata", "annotations", MIGRATION_RETRY_COUNT_ANNOTATION_KEY) || "0")
+          if retry_count >= MAX_MIGRATION_RETRIES
+            raise "Migration copy failed after #{MAX_MIGRATION_RETRIES} attempts for #{daemonizer_unit_name}, please contact support"
```
aamederen: can we somehow check this in CP and raise a Page when detected?

mohi-kalantari: yes I guess we can rely on the retry counter on pvs for checking the current state. I'll add the checks to the monitor

jeremyevans: This looks like an error we would be showing customers. Is the value of `daemonizer_unit_name` something the customer is expected to understand? If not, maybe we should reword:

Migration copy failed after #{MAX_MIGRATION_RETRIES} attempts, please contact support. Support Details: Failing unit: #{daemonizer_unit_name}.

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -220,14 +222,25 @@ def migrate_pvc_data(req_id, client, pvc, req)
...
+          end
         when "NotStarted"
           copy_command = ["rsync", "-az", "--inplace", "--compress-level=9", "--partial", "--whole-file", "-e", "ssh -T -c aes128-gcm@openssh.com -o Compression=no -x -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /home/ubi/.ssh/id_ed25519", "ubi@#{old_node_ip}:#{old_data_path}", current_data_path]
           run_cmd_output("nsenter", "-t", "1", "-a", "/home/ubi/common/bin/daemonizer2", "run", daemonizer_unit_name, *copy_command, req_id:)
           raise CopyNotFinishedError, "Old PV data is not copied yet"
         when "InProgress"
           raise CopyNotFinishedError, "Old PV data is not copied yet"
         when "Failed"
-          raise "Copy old PV data failed"
+          retry_count = Integer(pv.dig("metadata", "annotations", MIGRATION_RETRY_COUNT_ANNOTATION_KEY) || "0")
```
jeremyevans: While it isn't an issue in this case, it's best to use an explicit base with `Kernel#Integer`:

```suggestion
          retry_count = Integer(pv.dig("metadata", "annotations", MIGRATION_RETRY_COUNT_ANNOTATION_KEY) || "0", 10)
```

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -134,9 +134,25 @@ def init_health_monitor_session
...
   def check_pulse(session:, previous_pulse:)
     reading = begin
-      incr_sync_kubernetes_services if client(session: session[:ssh_session]).any_lb_services_modified?
-      "up"
-    rescue
+      k8s_client = client(session: session[:ssh_session])
+      incr_sync_kubernetes_services if k8s_client.any_lb_services_modified?
+
+      pvs = JSON.parse(k8s_client.kubectl("get pv -ojson"))["items"]
+      stuck_pvs = pvs.select { Integer(it.dig("metadata", "annotations", "csi.ubicloud.com/migration-retry-count") || "0") >= 3 }
```
```suggestion
      stuck_pvs = pvs.select { Integer(it.dig("metadata", "annotations", "csi.ubicloud.com/migration-retry-count") || "0", 10) >= 3 }
```

prog/test/kubernetes.rb
```diff
@@ -225,6 +225,31 @@ def self.assemble
...
+    client.kubectl("delete pod ubuntu-statefulset-0 --wait=false")
+    update_stack({"rsync_retry_source_node" => pod_node_name})
+    hop_kill_rsync_process
+  end
+
+  label def kill_rsync_process
+    source_node_name = strand.stack.first["rsync_retry_source_node"]
+    target_node = nodepool.nodes.find { it.name != source_node_name }
+    output = target_node.vm.sshable.cmd("pgrep -af 'rsync.*ubicsi' || true").strip
+    nap 5 if output.empty?
```
aamederen: `nap 1` could be better here, as we expect a very immediate rsync start, right?

What happens if rsync finishes before we capture it in this line?

prog/test/kubernetes.rb
```diff
@@ -225,6 +225,31 @@ def self.assemble
...
+    target_node.vm.sshable.cmd("sudo pkill -9 -f 'rsync.*ubicsi'")
```
aamederen: let's check if we were able to kill the process properly

aamederen:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22381934015

mohi-kalantari:
the e2e tests won't work here yet. I need to update the docker image first. I'll update it

prog/test/kubernetes.rb
mohi-kalantari: I changed to nap 1 just in case. The test would pass if we miss the kill without checking the retry mechanism. Since 1.5 GB of data is being copied, it's very unlikely to miss it but still I changed it to 1 second.

mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22404615707

mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22405277892

aamederen (approved):

PR #4835, opened 2026-02-23
Version tracking for rhizome installations

bsatzger:
**Record rhizome info upon installation**
    
    Enhance Prog::InstallRhizome to persist a RhizomeInstallation record.
    These data points are persisted:
    
    - folder: target folder of rhizome installation
    - commit: 8-character git commit hash.
    - digest: 24-character hex digest of rhizome (file path+content).
    - installed_at: Timestamp of successful completion.

**Add RhizomeInstallation model**

**Migration: Add rhizome_installation table**
    
    Separate table associated with sshable to keep track of which version of
    rhizome is currently installed on the sshable resource.

**Add GIT_COMMIT_HASH to environment and Config**
    
    Establish a single source of truth for the current git revision across
    environments and process types (web, monitor, respirate).
    
    - loader.rb: Resolve revision from Heroku , Dokku, or local git.
    - config.rb: Expose the value via Config.git_commit_hash.
    - Standardize on an 8-character short hash

### DIFF
--- a/prog/install_rhizome.rb
+++ b/prog/install_rhizome.rb
@@ -37,8 +37,11 @@ class Prog::InstallRhizome < Prog::Base
         end
       end
 
+      hashes_json = JSON.generate(file_hash_map.sort.to_h)
+      digest = Digest::SHA256.hexdigest(hashes_json)[0, 24]
+      update_stack({"rhizome_digest" => digest})
       writer.add_file("hashes.json", 0o100755) do |tf|
-        tf.write file_hash_map.to_json
+        tf.write hashes_json
       end
     end
 
@@ -56,6 +59,13 @@ class Prog::InstallRhizome < Prog::Base
 
   label def validate
     sshable.cmd("common/bin/validate")
+    folder = frame["target_folder"]
+    commit = Config.git_commit_hash
+    digest = frame["rhizome_digest"]
+    RhizomeInstallation.dataset.insert_conflict(
+      target: :id,
+      update: {folder:, commit:, digest:, installed_at: Sequel::CURRENT_TIMESTAMP}
+    ).insert(id: sshable.id, folder:, commit:, digest:)
 
     pop "installed rhizome"
   end
--- a/spec/routes/web/admin/model/rhizome_installation_spec.rb
+++ b/spec/routes/web/admin/model/rhizome_installation_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require_relative "spec_helper"
+
+RSpec.describe CloverAdmin, "RhizomeInstallation" do
+  include AdminModelSpecHelper
+
+  before do
+    @instance = create_rhizome_installation
+    admin_account_setup_and_login
+  end
+
+  it "displays the RhizomeInstallation instance page correctly" do
+    click_link "RhizomeInstallation"
+    expect(page.status_code).to eq 200
+    expect(page.title).to eq "Ubicloud Admin - RhizomeInstallation"
+
+    click_link @instance.admin_label
+    expect(page.status_code).to eq 200
+    expect(page.title).to eq "Ubicloud Admin - RhizomeInstallation #{@instance.ubid}"
+  end
+end


fdr:
I was thinking about this, it has a crucial problem: rhizome stays pretty stable across versions, so dispersion in the git revision is pretty unhelpful, except insomuch as you backward match it to the last  commit to revise rhizome for a particular service (given how we slice directories)

Like I split a roll out across a couple of days...two stored hashes would result, but it's probably(?) the same stuff.

bsatzger:
> I was thinking about this, it has a crucial problem: rhizome stays pretty stable across versions, so dispersion in the git revision is pretty unhelpful, except insomuch as you backward match it to the last commit to revise rhizome for a particular service (given how we slice directories)
> 
> Like I split a roll out across a couple of days...two stored hashes would result, but it's probably(?) the same stuff.

That's true. There are three related pieces of information we could use:
- git revision hash of rhizome (this PR)
- rhizome content hash
- rhizome `updated_at`

Each answer a slightly different question. We could pick one of those, a combination, or all.

bsatzger:
With the current version we add these three fields:
```
#<Sshable..., rhizome_commit: "39e9e969", rhizome_digest: "3e732e4274a222a528f9fb44", rhizome_installed_at: "2026-02-24 13:56:56"}>
```

migrate/20260225_add_rhizome_version_info.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:sshable) do
+      add_column :rhizome_commit, String, collate: '"C"', null: true
+      add_column :rhizome_digest, String, collate: '"C"', null: true
+      add_column :rhizome_installed_at, :timestamptz, null: true
```
jeremyevans: `null: true` is the default, so it doesn't need to be specified.

I would guess this information is accessed far less frequently than other information in sshable. You may want to consider whether it would be better to store it in a separate table with a foreign key to sshable.

bsatzger: Created a separate table.

bsatzger:
> I'm not sure whether `rhizome_digest` is worth adding. I guess it gives a quick indication that two separate commits have the same rhizome installed, but it's possible to get that information using `git diff rhizome_commit1...rhizome_commit2 rhizome`.

Initially I only had commit info for that reason. I added the digest because of feedback I received. And I actually do think it's convenient to be able to run a single query to achieve things like finding all rhizomes of a certain version or verify whether all hosts the exhibit a problem in production are on the same version or not.

model/rhizome_installation.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative "../model"
+
+class RhizomeInstallation < Sequel::Model
+  plugin ResourceMethods, etc_type: true
```
jeremyevans, line 6: I wonder if we should have support in `ResourceMethods` for models whose primary keys are foreign keys to another table. `etc_type: true` looks a bit odd because primary key ubids for this model do not have `et` type. I realize that's the best we can do currently.

migrate/20260226_rhizome_version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:rhizome_installation) do
+      foreign_key :id, :sshable, type: :uuid, primary_key: true
```
jeremyevans: I think it's better to use ON DELETE CASCADE here:

```suggestion
      foreign_key :id, :sshable, type: :uuid, primary_key: true, on_delete: :cascade
```

model/sshable.rb
```diff
@@ -6,6 +6,9 @@
 class Sshable < Sequel::Model
   prepend NetSsh::WarnUnsafe::Sshable
 
+  one_to_one :rhizome_installation, key: :id, read_only: true
+  plugin :association_dependencies, rhizome_installation: :destroy
```
jeremyevans: Not needed if using `ON DELETE CASCADE`:

```suggestion
```

prog/install_rhizome.rb
```diff
@@ -56,6 +59,20 @@ class Prog::InstallRhizome < Prog::Base
...
+        commit: Config.git_commit_hash,
+        digest: frame["rhizome_digest"],
+        installed_at: Time.now
+      }
+    ).insert(
+      id: sshable.id,
+      folder: frame["target_folder"],
+      commit: Config.git_commit_hash,
+      digest: frame["rhizome_digest"]
+    )
```
jeremyevans:
```suggestion
    folder = frame["target_folder"]
    commit = Config.git_commit_hash
    digest = frame["rhizome_digest"]
    RhizomeInstallation.dataset.insert_conflict(
      target: :id,
      update: {folder:, commit:, digest:, installed_at: Sequel::CURRENT_TIMESTAMP}
    ).insert(id: sshable.id, folder:, commit:, digest:)
```

jeremyevans (approved):

furkansahin:
@serprex , we don't know much about your environment, so, if you want to use it, can you make sure the git_sha could be fetched in your environment somehow?

PR #4836, opened 2026-02-23
Make respirate handle database disconnects

jeremyevans:
If a disconnect is detected, have the thread sleep for 100ms and retry. Do this for the main (scan) thread, repartition thread, and all worker threads.

If a disconnect error happens in a worker thread, don't add the respirate metric and don't close the apoptosis thread.

### DIFF
--- a/bin/monitor
+++ b/bin/monitor
@@ -83,7 +83,15 @@ repartitioner = Repartitioner.new(partition_number:, channel: :monitor, listen_t
 
 runner = MonitorRunner.new(monitor_resources:, metric_export_resources:, repartitioner:, **runner_args)
 
-repartition_thread = Thread.new { repartitioner.listen }
+repartition_thread = Thread.new do
+  repartitioner.listen
+rescue Sequel::DatabaseDisconnectError, Sequel::DatabaseConnectionError
+  sleep(1 + rand)
+  retry
+rescue => ex
+  Clog.emit("unexpected monitor repartition thread failure", Util.exception_to_hash(ex))
+  raise
+end
 
 # Only NOTIFY for the first 3-5 seconds, so that by the time we actually start monitoring,
 # all monitor processes know the expected partitioning. The rand is to avoid thundering herd issues.


jeremyevans:
> why? I follow the mechanism, but why not let its exponential backoff push back and reschedule it in the future? I suppose it must make something of consequence faster?

Without this, respirate stops functioning completely if there is a database disconnect, because it checks out a connection per thread. There is no exponential backoff via `try` in this case (you cannot update `try` due to the disconnect).

fdr:
> > why? I follow the mechanism, but why not let its exponential backoff push back and reschedule it in the future? I suppose it must make something of consequence faster?
> 
> Without this, respirate stops functioning completely if there is a database disconnect, because it checks out a connection per thread. There is no exponential backoff via `try` in this case (you cannot update `try` due to the disconnect).

so that seems like a headliner, I would have thought it'd have evicted it out of the pool and shrugged until it needed to get more.

fdr:
though, scan and repartition are special for sure

fdr:
next question: I totally get it  for the scan and repartition, but what about regular strands? what happens then?

jeremyevans:
Typical Sequel usage is for fine grained checkouts, which increases potential concurrency with smaller pool sizes. However, since strands run inside a transaction, there isn't much of an advantage to fine grained checkouts. Checking out a connection for the entire worker thread means it doesn't need to be checked in and out multiple times every run. While we don't necessarily have to use coarse grained checkouts for strand threads, I think the benefits exceed the costs.

fdr (approved):
maybe I'm not understanding this right (which may say something about the commit message, which understates the problem [thing doesn't work right when there's a disconnect in a very fundamental way?]), but it *also*, if I'm reading it right, will convert any failure resulting from a database disconnect to a retry rather than letting strand exponential backoff take over...is that right? Since it'll hit "retry" and go straight to running `strand_thread` again. Unlike other faults, that will do the normal backoff.

IMO, the commit text focuses on "what" not "why" and a pretty big difference(? if I'm right) in how a certain kind of fault interacts with the exponential backoff code, plus the observed defect. Consider this arbitrary example of how a mistake manifested in a system per cross-reference from COMMIT_MESSAGES.md https://github.com/postgres/postgres/commit/b47c50e5667b489bec3affb55ecdf4e9c306ca2d (it could have been better: what were "synchronization failures"?)

jeremyevans:
It's true there is no exponential backoff, but this does sleep, it doesn't go straight to running strand thread again. It is not possible to use the same exponential backoff that strands use via `try`, because if this happens, you cannot write to the database.

Maybe it would be better to sleep for 1s or more instead of 100ms? Database disconnects are quite rare, the only reason we hit this issue at all was for a planned restart to increase max_connections. I guess 100ms for 22 threads ends up being 220 attempts to connect per second per respirate process, and in hindsight that is probably too high if the database is going to be down and not just restarted. My thought here was to reduce downtime during database restarts, but considering we have about a 5 second downtime each deployment, this is probably going too far. Maybe `sleep(1 + rand)` to avoid the thundering herd?

I'll see if I can better explain the issue in the commit message. The main line of the commit message does state the fix, "Make respirate handle database disconnects", which correctly implies that they weren't handled at all before. I didn't think to explicitly answer "why" in the commit message, because I assumed that not handling disconnects is a significant enough problem to be a reason in itself.

fdr:
What is to "not handle" something? "Otherwise the process would exit"?  "would stop repartitioning, or scanning, but otherwise remain"? IMO "handling" things covers a broad range of phenomena and misbehavior, from making error messages or logging more robust to partial crashes (arguably the worst) to process exit. I can read the rescue and retry in this relatively short patch myself, but I can't read the symptom quickly if there are additional layers.

It also has a side effect the retry semantics on a very particular failure class for regular worker stuff, skipping their nominal exponential backoff. This is not "why" but it's a meaningful subtle change, even if we elect to accept it (which is fine, we can). Better still if you checked it out and mentioned that it's quite rare. But I kinda wonder why not rescue, shrug, and wait for the strand work to go through scheduling again...like most other faults, most other times. why is this category special?

jeremyevans:
I'll update the commit message and add some explanatory comments.

> But I kinda wonder why not rescue, shrug, and wait for the strand work to go through scheduling again...like most other faults, most other times. why is this category special?

That's basically what this does. It rescues, shrugs for 100ms, and then retries the loop. Potentially, a strand that was leased at the time of disconnect will remain leased until the lease expires, since it cannot release the lease or update `try` as it has no database connection to do so. Note that it has to rescue the exception at a high enough level that Sequel will actually handle it as a disconnect.  You cannot rescue the disconnect error inside the `DB.synchronize` block, you must rescue outside it. As mentioned earlier, we could drop the coarse grained checkouts for the strand threads (the manual `DB.synchronize` usage), but I don't think that's a good tradeoff. In any case, we'd still need the rescuing and retrying for the scan and repartition threads.

jeremyevans:
I updated the comments and commit message to better describe the problem. I also moved `handle_disconnects` from being a class method to being an instance method.

jeremyevans:
Did some more testing with this today, both in my development environment and in staging. I found that monitor also did not deal well with database disconnects. It still kind of worked if the database was only rebooted, but if it was down for a longer period of time, it died completely. This includes commits to fix those issues, as well as add logging whenever respirate/monitor threads exit due to an exception.

While fixing the issues, I also came across and fixed a logic bug in #4716 that I think is causing the VM flappiness. When looking in Mezmo, there is a dramatic increase in VM flappiness after the deployment of #4716. Take a look: https://github.com/ubicloud/ubicloud/pull/4836/changes/690c96e552ef382d64e7dd953a4721685b37b126#diff-1f48f550efa5bb3f812ae052579b02a9794aa1ef9b8d3ec98597ff2b7a281948R142

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22459329409

fdr (approved):
Some comments on commit messages (the latter ones were short so I don't think they come as a surprise). Perhaps one more fundamental question about robustness can be seen at the very end, like, is our design as-is fundamentally too tolerant of dying singleton Threads of great importance?

Inline comments below

> Only run initial scan logic on the first run

> The initial scan logic should only be run once. Previously, since
> the introduction of the initial scan support, we've been running
> the initial scan logic every iteration, which is not what we want
> to do.

The bug vs. the obvious intent is clear when I load the code, but
seeing as we've been living with it: what symptom or byproduct do we
anticipate from fixing it? I think it'd be good for the commit
message. Perhaps I could piece it together by stringing together
blames on the original designs, but now that it's a deep part of the
status quo, I'm curious what the person who figured it out expects to
see.

>  Handle disconnect/connect errors in monitor worker threads

I pieced together "handle" in this patch more quickly (the "ensure
block" comment was the key hint), but it'd be nice to not have to chase this
reference and guess the symptom it fixes (it looks like this would
prevent sticking?)


>  Handle disconnect/connect errors in monitor repartition thread

I figured it out even more quickly, now that I have this swapped in,
though someone looking through the peephole of `blame` in the future
will not have it swapped in at the time: bin/monitor is an entry point
and there's no indentation, so previously it would crash the whole
thread. It raise a question: ... do we degrade gracefully from *any
other* exception? It seems like repartition thread failure may want to
hit system.exit or log + re-exec, a kind of "PANIC" like situation.

jeremyevans:
> Some comments on commit messages (the latter ones were short so I don't think they come as a surprise). Perhaps one more fundamental question about robustness can be seen at the very end, like, is our design as-is fundamentally too tolerant of dying singleton Threads of great importance?

"Too tolerant" or "too intolerant"? Even with these changes, only disconnect/connect errors are retried. Other errors still cause the thread to exit, though now we at least log the errors via `Clog.emit` (in addition to the Ruby warning prints when the thread exits with an exception, which still happens as we reraise).

In terms of the individual threads dying:

* Worker threads: Concurrency per process drops to some degree. If all worker threads die, that means the process runs, but no work gets done.

* Repartition thread dies: no immediate effect, but cannot respond to changes in number of respirate/monitor processes

* Main (scan) thread dies: process exits

> Inline comments below
> 
> > Only run initial scan logic on the first run
> 
> > The initial scan logic should only be run once. Previously, since
> > the introduction of the initial scan support, we've been running
> > the initial scan logic every iteration, which is not what we want
> > to do.
> 
> The bug vs. the obvious intent is clear when I load the code, but seeing as we've been living with it: what symptom or byproduct do we antipcate from fixing it? I think it'd be good for the commit message. Perhaps I could piece it together by stringing together blames on the original designs, but now that it's a deep part of the status quo, I'm curious what the person who figured it out expects to see.

Good question. The expectation is the flappiness since the introduction of the initial scan support goes away. I agree the commit message should include this.

> > Handle disconnect/connect errors in monitor worker threads
> 
> I pieced together handle in this patch more quickly (the "ensure block" was the key hint), but it'd be nice to not have to chase this reference and guess the symptom it fixes (it looks like this would prevent sticking?)

This patch doesn't have an `ensure` block added or modified. It does have a general `rescue` block added (in addition to the disconnect handling blocks), but all that does is emit the error message and reraise.

> > Handle disconnect/connect errors in monitor repartition thread
> 
> I figured it out even more quickly, now that I have this swapped in: bin/monitor is an entry point and there's no indentation, so previously it would crash the whole thread. It raise a question: ... do we degrade gracefully from _any other_ exception? It seems like repartition thread failure may want to hit system.exit or log + re-exec, a kind of "PANIC" like situation.

As mentioned above, the loss of the repartition thread doesn't effect running of the application unless you are adding/removing processes. Since it's fairly rare that we actually modify the number of monitor (or respirate) processes, this is less critical, I think. With the logging added, we can now search for cases where it did fail, and add appropriate handling for those if there are any.

fdr:
> As mentioned above, the loss of the repartition thread doesn't effect running of the application unless you are adding/removing processes. Since it's fairly rare that we actually modify the number of monitor (or respirate) processes, this is less critical, I think. With the logging added, we can now search for cases where it did fail, and add appropriate handling for those if there are any.

I do find that quite extraordinary. This is a nasty surprise waiting to happen, we should only tolerate it if we absolutely have to degrade in this way. That doesn't block commit, it's status quo, but I find it remarkable, personally. We also never recover from this degradation.

> This patch doesn't have an ensure block added or modified. It does have a general rescue block added (in addition to the disconnect handling blocks), but all that does is emit the error message and reraise.

I thought it said something like this...it doesn't *change* the ensure block, but it waves towards "I rely on this", and the ensure block mostly has to do with messing with some fields that look important to completion status.

```
   rescue Sequel::DatabaseDisconnectError, Sequel::DatabaseConnectionError
            # Reraise so that ensure block runs
            raise
```

jeremyevans:
Pushed edits to a commit message and code comment.

PR #4838, opened 2026-02-23
Guard rhizome user cleanup in bootstrap

fdr:
The old `userdel -rf rhizome || true` suppresses all errors: -f forces removal even with active processes (e.g. an SSH session from a prior bootstrap attempt), and || true swallows everything from EPERM to corrupt passwd.

Replace with an explicit guard: check whether the user exists, terminate any lingering processes, then remove without -f.  If the user doesn't exist, skip entirely.  If userdel fails because processes survived SIGTERM, the control plane retries with backoff until the host is clean.

### DIFF
--- a/prog/bootstrap_rhizome.rb
+++ b/prog/bootstrap_rhizome.rb
@@ -63,7 +63,16 @@ if [ :no_bundler_install = false ]; then
   sudo apt-get -y install ruby-bundler
 fi
 sudo which bundle
-sudo userdel -rf rhizome || true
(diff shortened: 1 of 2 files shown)


fdr:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22327561677

pykello (approved):
I tried every bit of this and looked at the related man pages for exit codes & behaviour. Looks good as far as I can tell.

PR #4839, opened 2026-02-23
Guard VM user/group cleanup in create_unix_user

fdr:
The old `userdel --remove --force || true` and `groupdel -f || true` suppress all errors to achieve idempotency.  This masks real problems: -f silently kills lingering processes or removes groups with other members, and || true swallows permission and I/O errors.

Replace with explicit guards: check `id`/`getent group` before removal, terminate lingering processes with logged output, and drop -f from both commands.  Failures now propagate and the control plane retries with backoff until the host is clean.

### DIFF
--- a/prog/vm/metal/nexus.rb
+++ b/prog/vm/metal/nexus.rb
@@ -120,10 +120,17 @@ class Prog::Vm::Metal::Nexus < Prog::Base
     uid = rand(1100..59999)
     command = <<~COMMAND
       set -ueo pipefail
-      # Make this script idempotent
-      sudo userdel --remove --force :vm_name || true
-      sudo groupdel -f :vm_name || true
-      # Create vm's user and home directory
+      if id :vm_name &>/dev/null; then
+        procs=$(ps -u :vm_name -o pid,comm,args --no-headers) || [ $? -eq 1 ]
+        if [ -n "$procs" ]; then
+          echo "Terminating :vm_name processes:"
+          echo "$procs"
+          sudo pkill -u :vm_name || [ $? -eq 1 ]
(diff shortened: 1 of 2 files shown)


fdr:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22327550553

fdr:
It worked, and I collected timing data, and it did well, but this is in a pretty sensitive path for github actions, we should be monitoring for timely vm provisioning when it goes out.

prog/vm/metal/nexus.rb
```diff
@@ -120,10 +120,17 @@ def before_destroy
     uid = rand(1100..59999)
     command = <<~COMMAND
       set -ueo pipefail
-      # Make this script idempotent
-      sudo userdel --remove --force :vm_name || true
-      sudo groupdel -f :vm_name || true
-      # Create vm's user and home directory
+      if id :vm_name &>/dev/null; then
+        procs=$(ps -u :vm_name -o pid,comm,args --no-headers) || [ $? -eq 1 ]
```
bsatzger, line 124: I'm not positive, but I think if we're not hopping back from a later state to create_unix_user, the only lingering process could be a cert server. before, I assume we effectively sent SIGKILL (as part of force deleting the user) now we send SIGTERM. I assume cert server terminates on SIGTERM.

fdr:
Ah, there’s one thing not true about this: userdel -f I doubt sends any
processes a signal. It will leave them around with uid bits that are
undefined in /etc/passwd or so forth. Thus the process can leak or end up
being part of a user in the database in the future.

PR #4841, opened 2026-02-24
Combine monitoring of Vm and VmHostSlice into VmHost monitoring

jeremyevans:
This allows Vm and VmHostSlice to share the SSH connection that the VmHost uses for monitoring. It also reduces the number of monitor jobs, since the VmHost job also monitors all Vms and VmHostSlices on the VmHost.

This attempts to support this with minimal changes. Vm and VmHostSlice are each individually wrapped in MonitorableResources, and their check_pulse methods remain the same.  There are two main changes

MonitorResourceType#scan when setup for this support will scan for the Vms and VmHostSlices related to hosts in the current partition. This changes behavior from behavior, where the monitor process monitored Vms and VmHostSlices in the current partition.  This is some complexity here due to having to deal with new and removed resources.

MonitorableResource#check_pulse is monitored in two ways. First, if there is an exception, it checks whether the resource still exists, and marks it as deleted if not. Second, after checking the pulse for the current resource, it will check the pulse for attached resources, removing any that no longer exist.

The {Vm,VmHostSlice}#init_health_monitor_session methods are removed as the session is now created by the VmHost.

### DIFF
--- a/lib/monitorable_resource.rb
+++ b/lib/monitorable_resource.rb
@@ -97,8 +97,17 @@ class MonitorableResource
     attached_resources_sync { @attached_resources.values }.each do |resource|
       resource.session = @session
       resource.check_pulse
+      unless @session[:ssh_session]
+        # If checking the pulse for an attached resource drops the SSH session,
+        # remove the resource temporarily. It will be added back on the next scan
+        # (within 60s).  Also emit a log message so we can track this information.
+        # This is done so a problem with a particular attached resource does not
+        # cause monitoring to stop for all attached resources following it.
+        delete_attached_resource_ids << resource.resource.id
+        Clog.emit("monitor VmHost worker SSH connection lost", {monitor_vm_host_ssh_connection_lost: {host: @resource.ubid, resource: resource.resource.ubid}})
+        break
+      end
       delete_attached_resource_ids << resource.resource.id if resource.deleted
-      break unless @session[:ssh_session]
     end
 
     attached_resources_sync do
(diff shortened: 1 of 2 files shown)


jeremyevans:
@furkansahin I was testing this in staging and found it has thread safety issues. I pushed a commit to fix those.

bin/monitor
```diff
@@ -42,20 +42,23 @@ if Config.test?
...
     MinioServer,
     GithubRunner.exclude(ready_at: nil),
-    VmHostSlice.eager(:vm_host),
     LoadBalancerVmPort,
     KubernetesCluster,
     VictoriaMetricsServer,
     KubernetesNode.eager(:vm, :kubernetes_cluster)
   ]
+  host_attached_types = [
+    Vm.exclude(vm_host_id: nil).exclude(unix_user: "runneradmin").eager(:vm_storage_volumes),
```
enescakir: Since we already filter by `vm_host_id`
`ds.where(vm_host_id: id_range)`

```suggestion
    Vm.exclude(unix_user: "runneradmin").eager(:vm_storage_volumes),
```

jeremyevans: Thanks, made this change when I rebased on main.

fdr (approved):
the dynamics are complex but it's fundamentally really sound and can de-amplify ssh handshakes & forks by, like, a factor of a hundred. No small thing. @jeremyevans what's your feeling on how this re-composition sits relative to a clean-sheet design to achieve something similar?

jeremyevans:
For a clean sheet design, we may have wanted to be more efficient and check all VMs and slices in a single command. We can still move in that direction in later PRs, this was deliberately conservative in terms of code changes.

PR #4842, opened 2026-02-24
Require IMDSv2 on AWS

serprex:
Default is optional. IMDSv2 offers better security

Long term this endpoint should be disabled in favor of STS tokens


furkansahin:
I still don't understand how this will work since a user could as well first `PUT` with a token?

PR #4843, opened 2026-02-24
Validate init_script contains only ASCII characters

heavycrystal:
Sequel's column_encryption plugin returns decrypted values with ASCII-8BIT encoding, which causes Encoding::CompatibilityError when rendering in UTF-8 ERB templates. Enforcing ASCII-only scripts avoids this issue entirely.

### DIFF
--- a/model/postgres/postgres_init_script.rb
+++ b/model/postgres/postgres_init_script.rb
@@ -8,6 +8,7 @@ class PostgresInitScript < Sequel::Model
   def validate
     super
     validates_max_length(3000, :init_script)
+    errors.add(:init_script, "must contain only ASCII characters") unless init_script&.ascii_only?
   end
 end
 
--- a/model/vm_init_script.rb
+++ b/model/vm_init_script.rb
@@ -8,6 +8,7 @@ class VmInitScript < Sequel::Model
   def validate
     super
     validates_max_length(2000, :init_script)
+    errors.add(:init_script, "must contain only ASCII characters") unless init_script&.ascii_only?
   end
 end
 


jeremyevans:
I don't think it's reasonable to tell customers they cannot use non-ASCII in init scripts. If displaying the init scripts is causing issues, we should probably replace the invalid bytes when displaying it: `init_script.force_encoding(Encoding::UTF_8).scrub`

byucesoy:
> I don't think it's reasonable to tell customers they cannot use non-ASCII in init scripts. If displaying the init scripts is causing issues, we should probably replace the invalid bytes when displaying it: `init_script.force_encoding(Encoding::UTF_8).scrub`.

Does scrubbing mean characters would get removed? We only display the init script in the input box for users to edit them. If we scrub the invalid characters there, then when user sends the form again, we would break the script.

jeremyevans:
> > I don't think it's reasonable to tell customers they cannot use non-ASCII in init scripts. If displaying the init scripts is causing issues, we should probably replace the invalid bytes when displaying it: `init_script.force_encoding(Encoding::UTF_8).scrub`.
> 
> Does scrubbing mean characters would get removed? We only display the init script in the input box for users to edit them. If we scrub the invalid characters there, then when user sends the form again, we would break the script.

Scrubbing replaces invalid characters with placeholders. I didn't think we supported editing init scripts after they were submitted, but apparently we do for Postgres (unlike VMs).

Note that data submitted via the web via a browser should already be in UTF-8, as the web page is UTF-8. You didn't add support for Postgres init_scripts to the API/CLI, so I'm not sure how a non-UTF-8 init script would be added currently.

Assuming you plan to add support for Postgres init scripts to the API/CLI, a couple of alternative approaches:

* Allow valid UTF-8 (all ASCII is valid UTF-8). Not great as I don't think we should be telling customers what characters are supported in their init scripts, but certainly much better than only allowing ASCII.

* Don't allow editing of non-UTF-8 init scripts on the web. These must have been submitted via the API/CLI. If a customer used the API/CLI when adding a non-UTF-8 init script, they should be able to use the API/CLI to edit it.

PR #4846, opened 2026-02-25
Exclude GitHub runner slices from health monitoring

enescakir:
Runner VMs are already excluded from monitor_models via `.exclude(unix_user: "runneradmin")`, but their associated VmHostSlice records were still being monitored. This is unnecessary since runners have their own monitoring entry via GithubRunner. Filter them out with a subquery on the vm table.


jeremyevans (approved):
This seems fine, but will conflict with #4841. However, the conflict is easy to resolve, so no issue if this is merged first.

mohi-kalantari (approved):

PR #4849, opened 2026-02-25
Wait for kubelet CSR before approving it during node provisioning

mohi-kalantari:
The metrics server requires manually approving the kubelet-serving CSR whenever a new node joins the cluster. The previous code ran the approval immediately after installing the CNI, but sometimes the kubelet had not yet created the CSR by that point, causing the approval to silently do nothing.

The label now checks whether the CSR is already approved first, skipping approval entirely on retries. If not yet approved, it looks for a pending CSR and approves it. If neither exists, it naps to wait for the kubelet to create one.

### DIFF
--- a/lib/kubernetes/client.rb
+++ b/lib/kubernetes/client.rb
@@ -45,6 +45,14 @@ class Kubernetes::Client
     kubectl("delete node :node_name", node_name:)
   end
 
+  def get_csr(node_name, csr_status:)
+    kubectl("get csr --sort-by=.metadata.creationTimestamp | awk '/:csr_status/ && /kubelet-serving/ && /':node_name'/ {print $1}' | tail -1", node_name:, csr_status:).chomp
+  end
+
+  def approve_csr(csr_name)
+    kubectl("certificate approve :csr_name", csr_name:)
+  end
+
   def set_load_balancer_hostname(svc, hostname)
     patch_data = JSON.generate({
       "status" => {
--- a/spec/prog/kubernetes/provision_kubernetes_node_spec.rb
+++ b/spec/prog/kubernetes/provision_kubernetes_node_spec.rb
@@ -11,11 +11,11 @@ RSpec.describe Prog::Kubernetes::ProvisionKubernetesNode do
     Project.create(name: "default")
   }
   let(:subnet) {
-    Prog::Vnet::SubnetNexus.assemble(Config.kubernetes_service_project_id, name: "test", ipv4_range: "172.19.0.0/16", ipv6_range: "fd40:1a0a:8d48:182a::/64").subject
+    Prog::Vnet::SubnetNexus.assemble(project.id, name: "test", ipv4_range: "172.19.0.0/16", ipv6_range: "fd40:1a0a:8d48:182a::/64").subject
   }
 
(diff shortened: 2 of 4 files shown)


lib/kubernetes/client.rb
```diff
@@ -45,6 +45,18 @@ def delete_node(node_name)
     kubectl("delete node :node_name", node_name:)
   end
 
+  def get_pending_csr(node_name)
+    kubectl("get csr --sort-by=.metadata.creationTimestamp | awk '/Pending/ && /kubelet-serving/ && /':node_name'/ {print $1}' | tail -1", node_name:).chomp
+  end
+
+  def get_approved_csr(node_name)
+    kubectl("get csr --sort-by=.metadata.creationTimestamp | awk '/Approved/ && /kubelet-serving/ && /':node_name'/ {print $1}' | tail -1", node_name:).chomp
```
jeremyevans: These commands are the same except for the `Approved`/`Pending` difference. Up to you, but you may want a have a single method where you pass `Approved`/`Pending` as an argument.

spec/prog/kubernetes/provision_kubernetes_node_spec.rb
```diff
@@ -330,12 +330,32 @@
...
+      expect(session).to receive(:_exec!).with("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf get csr --sort-by=.metadata.creationTimestamp | awk '/Pending/ && /kubelet-serving/ && /'test-vm'/ {print $1}' | tail -1").and_return(Net::SSH::Connection::Session::StringWithExitstatus.new("\n", 0))
+      expect { prog.approve_new_csr }.to nap(5)
+    end
+
+    it "skips approve if the csr is already approved" do
+      expect(session).to receive(:_exec!).with("sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf get csr --sort-by=.metadata.creationTimestamp | awk '/Approved/ && /kubelet-serving/ && /'test-vm'/ {print $1}' | tail -1").and_return(Net::SSH::Connection::Session::StringWithExitstatus.new("csr-abc123\n", 0))
+      expect(kubernetes_cluster).to receive(:incr_sync_internal_dns_config)
+      expect(kubernetes_cluster).to receive(:incr_sync_worker_mesh)
+      expect { prog.approve_new_csr }.to exit({node_id: prog.node.id})
```
jeremyevans: This looks like it was just copying existing code, but this should avoid unnecessary mocking by checking for expected semaphore changes:

```suggestion
      expect { prog.approve_new_csr }.to exit({node_id: prog.node.id})
        .and change { kubernetes_cluster.sync_internal_dns_config_set? }.from(false).to(true)
        .and change { kubernetes_cluster.sync_worker_mesh_set? }.from(false).to(true)
```

Similar change to following spec.

jeremyevans (commented):
Looks good from a Ruby perspective. Leaving approval to Eren.

lib/kubernetes/client.rb
aamederen: alternatively, based on the usage pattern, a single get_csr method can return both approved and pending csr

PR #4850, opened 2026-02-25
Add upgrade_stage alongside upgrade_status in api

serprex:
not_running / running / failed lacks fidelity of detail necessary to present a UI similar to Ubicloud's own upgrade status page


serprex:
Considering other options:
* update `upgrade_status` to emit more states. this seemed like too much of a breaking change
* update `upgrade_stage` to limit itself to descriptions used by existing UI. This may be preferable, but wanted feedback first

PR #4854, opened 2026-02-26
Override rollup to fix security vulnerability

enescakir:
@stoplight/spectral-cli pins rollup@~2.79.2 via
@stoplight/spectral-ruleset-bundler, and the spectral repository is not actively maintained. The earliest patched rollup version is 2.80.0, which falls outside that tilde range.

Use npm overrides to force rollup@^2.80.0, resolving the Dependabot alert without waiting for an upstream spectral release.

Fixes https://github.com/ubicloud/ubicloud/security/dependabot/77

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4957,9 +4957,9 @@
       }
     },
     "node_modules/rollup": {
-      "version": "2.79.2",
-      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.79.2.tgz",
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
Seems fine. Does the vulnerability actually affect us?

enescakir:
> Seems fine. Does the vulnerability actually affect us?

No. We only use it to validate our OpenAPI spec in CI.

PR #4856, opened 2026-02-26
Improve CSI migration

mohi-kalantari:
**Add fallback PV-based migration detection in NodeStageVolume**

During CSI data migration, node_unstage_volume recreates the PVC with the old-pv-name annotation so that NodeStageVolume on the destination node knows to rsync data from the old PV. However, recreate_pvc has a non-atomic delete-then-create window: between delete_pvc and create_pvc, the StatefulSet controller can recreate the PVC without the migration annotation. If NodeStageVolume runs before the kubelet retries node_unstage_volume (which would eventually patch the annotation via the else branch), a fresh empty backing file is created and data is lost.

The old-pvc-object annotation on the old PV is set before any PVC deletion, so it is always present when migration is needed. This commit adds a fallback in fetch_and_migrate_pvc: if the PVC lacks the old-pv-name annotation, search for a retained PV whose claimRef matches the PVC and that carries the old-pvc-object annotation. If found, patch the annotation onto the PVC and proceed with migration.

**Handle StatefulSet PVC recreation race in recreate_pvc**

During node drain, recreate_pvc deletes the existing PVC and creates a new one with the old-pv-name migration annotation. Between these two calls, the StatefulSet controller can observe the missing PVC and recreate it first. When this happens, create_pvc fails with a generic RuntimeError, the entire node_unstage_volume call fails, and the annotation is only patched on the next kubelet retry via the else branch. This is correct eventually, but widens the window where the PVC exists without the annotation.

Introduce AlreadyExistsError in run_kubectl, mirroring the existing ObjectNotFoundError pattern, so that recreate_pvc can catch it specifically. On AlreadyExistsError, immediately patch the migration annotation onto the StatefulSet-created PVC instead of letting the unstage fail and relying on a full retry cycle.

### DIFF
--- a/kubernetes/csi/Gemfile.lock
+++ b/kubernetes/csi/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ubi-csi (0.6.0)
+    ubi-csi (0.7.0)
       base64
       grpc (~> 1.76)
       grpc-tools (~> 1.76)
--- a/kubernetes/csi/lib/ubi_csi.rb
+++ b/kubernetes/csi/lib/ubi_csi.rb
@@ -5,7 +5,7 @@ require "csi_pb"
 require "csi_services_pb"
 
 module Csi
-  VERSION = "0.6.0"
+  VERSION = "0.7.0"
 end
 
 require_relative "ubi_csi/identity_service"
--- a/kubernetes/csi/spec/csi/v1/identity_service_spec.rb
+++ b/kubernetes/csi/spec/csi/v1/identity_service_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe Csi::V1::IdentityService do
       request = Csi::V1::GetPluginInfoRequest.new
       response = service.get_plugin_info(request, call)
       expect(response.name).to eq("csi.ubicloud.com")
-      expect(response.vendor_version).to eq("0.6.0")
+      expect(response.vendor_version).to eq("0.7.0")
     end
   end
 
--- a/kubernetes/csi/ubi-csi.gemspec
+++ b/kubernetes/csi/ubi-csi.gemspec
@@ -2,7 +2,7 @@
 
 Gem::Specification.new do |spec|
   spec.name = "ubi-csi"
-  spec.version = "0.6.0"
+  spec.version = "0.7.0"
   spec.authors = ["Ubicloud"]
   spec.email = ["support@ubicloud.com"]
 
--- a/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
@@ -22,7 +22,7 @@ spec:
       priorityClassName: system-node-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.6.0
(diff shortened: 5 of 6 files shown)


mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22491048066

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -313,8 +325,13 @@ def recreate_pvc(req_id, client, pv)
...
             log_with_id(req_id, "Removed PVC finalizers #{pvc_namespace}/#{pvc_name}")
           end
-          client.create_pvc(pvc)
-          log_with_id(req_id, "Recreated PVC with the new spec")
+          begin
+            client.create_pvc(pvc)
+            log_with_id(req_id, "Recreated PVC with the new spec")
+          rescue AlreadyExistsError
+            log_with_id(req_id, "PVC already recreated by StatefulSet controller, patching migration annotation")
+            client.patch_resource("pvc", pvc_name, OLD_PV_NAME_ANNOTATION_KEY, pv_name, namespace: pvc_namespace)
```
jeremyevans: While I'm guessing `log_with_id` cannot raise `AlreadyExistsError`, best practice is to rescue the smallest necessary block:

```suggestion
          rescue AlreadyExistsError
            log_with_id(req_id, "PVC already recreated by StatefulSet controller, patching migration annotation")
            client.patch_resource("pvc", pvc_name, OLD_PV_NAME_ANNOTATION_KEY, pv_name, namespace: pvc_namespace)
          else
            log_with_id(req_id, "Recreated PVC with the new spec")
```

mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22584136139

kubernetes/csi/lib/ubi_csi/node_service.rb
```diff
@@ -126,10 +126,22 @@ def node_stage_volume(req, _call)
...
       def fetch_and_migrate_pvc(req_id, client, req)
-        pvc = client.get_pvc(req.volume_context["csi.storage.k8s.io/pvc/namespace"],
-          req.volume_context["csi.storage.k8s.io/pvc/name"])
+        pvc_namespace = req.volume_context["csi.storage.k8s.io/pvc/namespace"]
+        pvc_name = req.volume_context["csi.storage.k8s.io/pvc/name"]
+        pvc = client.get_pvc(pvc_namespace, pvc_name)
         if pvc_needs_migration?(pvc)
           migrate_pvc_data(req_id, client, pvc, req)
+
+        # Fallback: during node drain, recreate_pvc may race with the StatefulSet controller,
```
aamederen, line 135: can we change `recreate_pvc` so that, instead of delete-then-create, it creates the new PVC (or ensures it exists with the correct annotations) and only then deletes the old pvc?

aamederen (approved):
Overall I think this PR is an improvement, but I'm still not sure if we are covering the case completely and in the simplest way.

kubernetes/csi/lib/ubi_csi/node_service.rb
mohi-kalantari: Since it needs to be recreated with the same name, we have to delete it first.

PR #4858, opened 2026-02-27
Add a general zero-rate inference billing rate

aamederen:
For the cases where we handle billing out-of-band and fast-changing models, add a general zero-rate billing rate to be used within InferenceRouterModels.


config/billing_rates.yml
```diff
@@ -40,12 +40,8 @@
...
-- { id: b40ee941-d226-449e-8df5-cf32fed4e6dc, resource_type: InferenceTokens,             resource_family: qwen3-coder-480b-tom-input, location: global,         unit_price: 0.0000000000, billed_by: amount,   active_from: 2025-10-24T00:00:00Z, byoc: false }
-- { id: dab50985-262a-44a6-9489-b861fa69f8e3, resource_type: InferenceTokens,             resource_family: qwen3-coder-480b-tom-output,location: global,         unit_price: 0.0000000000, billed_by: amount,   active_from: 2025-10-24T00:00:00Z, byoc: false }
+- { id: 29319466-cb59-4741-bdf3-cd28b5f8eacf, resource_type: InferenceTokens,             resource_family: zero-rate-input,            location: global,         unit_price: 0.0000000000, billed_by: amount,   active_from: 2026-02-27T00:00:00Z, byoc: false }
+- { id: 33561edc-0b1e-4162-a1b0-22745b1d6d49, resource_type: InferenceTokens,             resource_family: zero-rate-output,           location: global,         unit_price: 0.0000000000, billed_by: amount,   active_from: 2026-02-27T00:00:00Z, byoc: false }
```
enescakir: Is it okay to show the usage as "zero-rate-output Inference Tokens" in the invoice for these customers?

https://github.com/ubicloud/ubicloud/blob/531c30a5576233cea97f43b190da4d526556a6ff/lib/billing_rate.rb#L66

aamederen: I think it's fine, but I'd be happier if we can find a better word. Something to reflect that "you used this much, but it's charged out of band"

bsatzger: ideas for naming that might look better on invoices:
`complimentary-input`
`preview-input` 

if these customers use multiple models they lose the ability to distinguish which model was used by how much.

aamederen: Thanks! I changed the rate to "preview" and brought the tom rates back.

PR #4859, opened 2026-02-27
Rename destroy label to finish in e2e test progs

enescakir:
The destroy label in test progs is not a true destroy: it cleans up the test project and pops in the happy path, but not for failed case. Its failure path (hop to failed, then nap) did not match the expected terminal semantics of a real destroy label. Naming it "destroy" caused the framework to reserve a destroy semaphore that was never used

When hop_destroy is called, it also increments the destroying semaphore,
but it never decrements it when it fails and naps forever. It blocks to
destroy strand.

Rename it to "finish" to avoid the special-cased destroy semantics and
remove the unnecessary semaphore declarations.

example failure: https://github.com/ubicloud/ubicloud/actions/runs/22472940506/job/65093850325

### DIFF
--- a/prog/test/ha_postgres_resource.rb
+++ b/prog/test/ha_postgres_resource.rb
@@ -3,8 +3,6 @@
 require_relative "../../lib/util"
 
 class Prog::Test::HaPostgresResource < Prog::Test::Base
-  semaphore :destroy
-
   def self.assemble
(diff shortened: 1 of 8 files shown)


enescakir:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22477267124

PR #4860, opened 2026-02-27
Add cache branch protection toggle to installation settings

enescakir:
- **Add cache_scope_protected column to github_installation**
  This column will replace the access_all_cache_scopes feature flag on
  Project, allowing customers to toggle cache branch isolation directly
  from the Ubicloud console instead of requiring admin intervention.
  
  Defaults to true (protected), matching GitHub's default behavior where
  workflow runs can only access caches from the current branch or the
  default branch.
  
- **Replace access_all_cache_scopes feature flag with column**
  The access_all_cache_scopes feature flag on Project required admin
  intervention to toggle. Move to a cache_scope_protected boolean on
  GithubInstallation so customers can control it themselves.
  
  The runtime cache lookup now checks installation.cache_scope_protected
  instead of the project-level feature flag. The logic is inverted:
  protected=true means scoped (the common case), matching the old
  access_all_cache_scopes=false default.
  
- **Add cache branch protection toggle to installation settings**
  Customers previously had to ask an admin to flip a feature flag to
  share caches across branches. Add a self-service toggle on the GitHub
  installation settings page, following the same pattern as the existing
  cache_enabled toggle.
  
  The toggle only appears when cache is enabled. It defaults to on,
  consistent with GitHub's default cache isolation behavior.
  
CleanShot 2026-02-27 at 11 31 37@2x

### DIFF
--- a/routes/project/github.rb
+++ b/routes/project/github.rb
@@ -55,6 +55,10 @@ class Clover
           @installation.cache_enabled = typecast_params.bool("cache_enabled")
         end
 
+        if typecast_params.present?("cache_scope_protected")
+          @installation.cache_scope_protected = typecast_params.bool("cache_scope_protected")
+        end
+
         if typecast_params.present?("premium_runner_enabled")
           @installation.allocator_preferences["family_filter"] = if typecast_params.bool("premium_runner_enabled")
             ["premium", "standard"]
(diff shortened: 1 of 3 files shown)


model/project.rb
```diff
@@ -206,7 +206,6 @@ def self.feature_flag(*flags, into: self)
   end
 
   feature_flag(
-    :access_all_cache_scopes,
```
jeremyevans, line 209: Is the plan to manually update the github installations for the projects with this feature flag to set `cache_scope_protected` to `true` after the migration and before applying this commit?

enescakir: We have several of them. Yes, I will update after the migration.

PR #4862, opened 2026-02-27
Aggregate GitHub usage by installation not project

enescakir:
We aggregate GitHub usage daily to reduce the number of billing records during implementation. Currently, we aggregate them by project. However, since a project can have multiple installations, it's difficult to determine which installation is responsible for the usage. Our UI is also installation-based, so having installation-level usage records makes it straightforward to display per-installation breakdowns. While the billing record already has a project ID column, using the installation ID as the resource ID is the right granularity.

### DIFF
--- a/model/project.rb
+++ b/model/project.rb
@@ -245,6 +245,7 @@ class Project < Sequel::Model
     :enable_r8gd,
     :enable_r8id,
     :free_runner_upgrade_until,
+    :github_billing_by_repository,
     :gpu_vm,
     :ipv6_disabled,
     :postgres_enable_maintenance_window_days,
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
Code wise, looks good.

How does this interact with previous billing records, which were by project and not installation? I guess there is a temporary split, with some records by installation and others by project on the day of the transition. I'm guessing that doesn't cause problems, but is there anything looking at BillingRecord resource_id where that could be an issue?

enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/29579556250

enescakir:
> Code wise, looks good.
> 
> How does this interact with previous billing records, which were by project and not installation? I guess there is a temporary split, with some records by installation and others by project on the day of the transition. I'm guessing that doesn't cause problems, but is there anything looking at BillingRecord resource_id where that could be an issue?

It's okay. We'll keep the old records as they are. The deployment might cause multiple records at the moment, but it won't result in any over or under charging.

Our internal dashboards might use it, I'll check with Umur.

enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/29817856318

PR #4863, opened 2026-02-27
Add incident.io as alternative pager

serprex:
we are decommissioning PagerDuty in favor of incident.io, deadline far off in May *(but already incident.io is what we use via PagerDuty integration)*

tested that this works, but don't think conditional duck typed class is way to go, want to get direction here before adding tests

### DIFF
--- a/model/page.rb
+++ b/model/page.rb
@@ -2,10 +2,72 @@
 
 require_relative "../model"
 
-require "pagerduty"
 require "openssl"
-
 class Page < Sequel::Model
+  class Client
+    def initialize
+      if Config.pagerduty_key
+        require "pagerduty"
+        @client = Pagerduty.build(integration_key: Config.pagerduty_key, api_version: 2)
+      end
+    end
+
+    def trigger(tag, summary:, severity:, details:, links:)
+      if Config.pagerduty_key
+        pagerduty_incident(tag).trigger(summary:, severity:, source: "clover", custom_details: details, links:)
+      elsif Config.incidentio_key
+        metadata = (details || {}).merge({severity:})
+        link, *links = links
+        metadata[:links] = links unless links.empty?
+        body = {
+          deduplication_key: deduplication_key(tag),
+          status: "firing",
+          title: summary,
+          source_url: link&.[](:href),
+          metadata:
+        }.to_json
+        Excon.post(incidentio_uri, body:, headers: {
+          "Authorization" => "Bearer #{Config.incidentio_key}",
+          "Content-Type" => "application/json"
+        })
+      else
+        Clog.emit("page triggered", {page_triggered: {tag:, summary:, severity:, details:, links:}})
+      end
+    end
+
+    def resolve(tag, summary:)
+      if Config.pagerduty_key
+        pagerduty_incident(tag).resolve
+      elsif Config.incidentio_key
+        body = {
+          deduplication_key: deduplication_key(tag),
+          status: "resolved",
+          title: summary
+        }.to_json
+        Excon.post(incidentio_uri, body:, headers: {
+          "Authorization" => "Bearer #{Config.incidentio_key}",
+          "Content-Type" => "application/json"
+        })
+      else
+        Clog.emit("page resolved", {page_resolved: {tag:}})
+      end
+    end
+
+    private
+
+    def pagerduty_incident(tag)
+      @client.incident(deduplication_key(tag))
+    end
+
+    def incidentio_uri
+      "https://api.incident.io/v2/alert_events/http/#{Config.incidentio_alert_source_config_id}"
+    end
+
+    def deduplication_key(tag)
+      OpenSSL::HMAC.hexdigest("SHA256", "ubicloud-page-key", tag)
+    end
+  end
+
   dataset_module do
     def group_by_vm_host
       pages = all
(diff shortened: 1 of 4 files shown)


serprex:
summary as title is kinda funny, but required field: https://docs.incident.io/api-reference/alert-events-v2/createhttp

needed even on resolve

model/page.rb
```diff
@@ -2,11 +2,94 @@
 
 require_relative "../model"
 
-require "pagerduty"
 require "openssl"
 
+if Config.pagerduty_key
+  require "pagerduty"
+
+  class PagerClient
```
jeremyevans: It's a bad idea to add a top-level constant outside of file dedicated to it. For one, it doesn't work well with reloading in development.  This is an implementation detail for `Page`, I would nest it under `Page`, so `Page::Client`

When creating it, I recommend the following structure:

```ruby
  class Page < Sequel::Model
    class Client
      if Config.pagerduty_key
      elsif Config.incidentio_key
      else
      end
    end

    # rest of Page
  end
```

model/page.rb
```diff
@@ -2,11 +2,94 @@
...
+    def initialize
+      @uri = URI("https://api.incident.io/v2/alert_events/http/#{Config.incidentio_alert_source_config_id}")
+    end
+
+    def trigger(tag, summary:, severity:, details:, links:)
+      metadata = (details || {}).merge({severity:})
+      metadata[:links] = links[1..] if links && links.length > 1
+      Net::HTTP.start(@uri.hostname, use_ssl: true) do |http|
```
jeremyevans: We do use `Net::HTTP` in some cases for HTTP requests, but I think the general preference is to use `Excon`. I think it would be better to standardize on `Excon`, and only use `Net::HTTP` in cases that don't load ubicloud's Gemfile, such as rhizome, bin/ubi, and the SDK.

model/page.rb
```diff
@@ -2,11 +2,94 @@
...
+  require "uri"
+
+  class PagerClient
+    def initialize
+      @uri = URI("https://api.incident.io/v2/alert_events/http/#{Config.incidentio_alert_source_config_id}")
+    end
+
+    def trigger(tag, summary:, severity:, details:, links:)
+      metadata = (details || {}).merge({severity:})
+      metadata[:links] = links[1..] if links && links.length > 1
```
jeremyevans:
```suggestion
      link, *links = links
      metadata[:links] = links unless links.empty?
```

model/page.rb
```diff
@@ -2,11 +2,94 @@
...
+      Net::HTTP.start(@uri.hostname, use_ssl: true) do |http|
+        request = Net::HTTP::Post.new(@uri)
+        request["Authorization"] = "Bearer #{Config.incidentio_key}"
+        request["Content-Type"] = "application/json"
+        request.body = {
+          deduplication_key: deduplication_key(tag),
+          status: "firing",
+          title: summary,
+          source_url: links.first&.[](:href),
```
jeremyevans:
```suggestion
          source_url: link&.[](:href),
```

model/page.rb
```diff
@@ -2,11 +2,94 @@
...
+
+    def deduplication_key(tag)
+      OpenSSL::HMAC.hexdigest("SHA256", "ubicloud-page-key", tag)
+    end
+  end
+else
+  class PagerClient
+    def trigger(tag, **kwargs)
+      kwargs[:tag] = tag
+      Clog.emit("page triggered", kwargs)
```
jeremyevans:
```suggestion
      Clog.emit("page triggered", {page_triggered: kwargs})
```

model/page.rb
```diff
@@ -2,11 +2,94 @@
...
+    end
+
+    def resolve(tag)
+      Clog.emit("page resolved", {tag:})
```
jeremyevans:
```suggestion
      Clog.emit("page resolved", {page_resolved: {tag:}})
```

model/page.rb
serprex: even here I wonder how to unit test these different branches. thinking of moving the conditional logic into trigger/resolve

jeremyevans: This approach does not allow for unit testing all different branches. You can move the conditional logic into the `trigger`/`resolve` method, or have a separate classes for each case that are always available, and you have the `client` method pick the appropriate class. I think the `trigger`/`resolve` with conditionals is probably simpler overall. The separate class approach is simpler when there are significantly more methods where conditionals would be needed.

spec/model/page_spec.rb
```diff
@@ -45,37 +45,94 @@
...
 
-  describe "#resolve" do
-    it "resolves the page in Pagerduty if key is present" do
-      expect(Config).to receive(:pagerduty_key).and_return("dummy-key").at_least(:once)
-      stub_request(:post, "https://events.pagerduty.com/v2/enqueue")
-        .to_return(status: 200, body: {dedup_key: "dummy-dedup-key", message: "Event processed", status: "success"}.to_json, headers: {})
+    describe "#trigger" do
+      it "triggers a page if key is present" do
+        expect(p).to receive(:details).and_return({}).at_least(:once)
+        expect(Excon).to receive(:post).with("https://api.incident.io/v2/alert_events/http/source-id", body: {
```
jeremyevans: Existing code uses `Excon.stub` instead of rspec method overriding, and it's probably best to follow that, both here and in the following spec.

model/page.rb
```diff
@@ -2,10 +2,72 @@
...
+          title: summary,
+          source_url: link&.[](:href),
+          metadata:
+        }.to_json
+        Excon.post(incidentio_uri, body:, headers: {
+          "Authorization" => "Bearer #{Config.incidentio_key}",
+          "Content-Type" => "application/json"
+        })
+      else
+        Clog.emit("page triggered", {page_triggered: {tag:, summary:, severity:, details:, links:}})
```
jeremyevans, line 34: Should probably have a spec that tests for this `Clog.emit`. Same for the page resolved `Clog.emit`. I'm guessing you have a spec that covers this code (since coverage testing passed), but it doesn't check what the branch actually does.

PR #4864, opened 2026-02-27
Enable direct I/O on CSI loop devices

mohi-kalantari:
The CSI driver creates a backing .img file on the host's ext4 filesystem, attaches it as a loop device, and formats it with ext4. This ext4-on-ext4 stacking causes the host page cache to buffer data that the inner filesystem is also caching, roughly halving the effective memory available for caching and adding write amplification.

Adding --direct-io=on to losetup causes the loop driver to open the backing file with O_DIRECT, bypassing the host page cache entirely. The inner filesystem's page cache remains the single caching layer, which is the correct behavior. This flag has been stable since Linux 4.10, and ext4 backing filesystems support O_DIRECT. Existing volumes require no migration; the flag takes effect the next time a loop device is attached.


mohi-kalantari:
Holding this off for now

PR #4866, opened 2026-03-02
Decrease e2e timeout by 15 minutes

enescakir:
The average e2e metal runtime is 57m and all runs were under 1h. 70m is high enough to cover outliers, and there is no need to wait 15 minutes in case of stuck tests.


enescakir:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22570249658

shikharbhardwaj (approved):

PR #4868, opened 2026-03-02
Integrate Authorization into Clover

jeremyevans:
This still allows the Authorization class methods to work, but makes
them normal instance methods, so Clover can just include Authorization,
and then override any of the Authorization methods to get
request-specific behavior.

### DIFF
--- a/model/postgres/postgres_resource.rb
+++ b/model/postgres/postgres_resource.rb
@@ -384,7 +384,7 @@ class PostgresResource < Sequel::Model
     end
 
     Util.send_email(
-      accounts_with_access.map(&:email).uniq,
+      accounts_with_access,
       "PostgreSQL Storage Warning: #{name} at #{usage_percent}% capacity",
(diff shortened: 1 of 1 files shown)


enescakir (approved):

jeremyevans:
> It comes out so elegantly with [2425aab](https://github.com/ubicloud/ubicloud/commit/2425aab278bd2b565594cf410c665b88ab16f774) , even though the tiny motions done here do not explain themselves without that as context. I suppose we should let @serprex @iamKunalGupta try it out to see if it meets their concrete need before we commit it?

I think it's a better design, and it's required to fix a couple of N+1 issues and a bug fix (from #4869, which was merged into this PR). So I plan to merge this. We can make additional changes later if @serprex or @iamKunalGupta determine this by itself is insufficient for their needs.

fdr:
Jeremy told me out of band that he thought this was a good change even independent of making the system more flexible, so, no complaints there.

PR #4870, opened 2026-03-03
Throttle Postgres Disk IO during archival backlog pressure

shikharbhardwaj:
### Description

* Add rhizome scripts for PostgreSQL I/O throttling via cgroups v2
Implements disk I/O throttling for PostgreSQL using Linux cgroups v2.
Creates a child cgroup under the postgresql service and moves processes
to throttled/immune cgroups based on their role.

  Immune processes [1]:
  - archiver: its progress resolves the archival backlog
  - logger: throttling affects archiver (logs go to stderr)
  - checkpointer: checkpoints trigger WAL deletion
  - writer (wal writer + background writer): allow checkpoints to complete
  
  Since cgroups v2 requires "no internal processes" to live in a
  group with a controller enabled, we need to create a cgroup hierarchy
  which allows classifying postmaster's child processes into 2 groups.
  This requires systemd to delegate cgroup hierarchy to the user, which
  requires a restart. In the current state, the apply mechanism ignores
  the request if the requested controller cannot be enabled on an
  instance.
  
  [1]: https://github.com/fdr/ratchet

* Implement disk IO throttling based on postgres archival backlog
Adds I/O throttling logic to the control plane that monitors archival
backlog and applies progressive throttling.  The bandwidth thresholds in
this patch are set so that the highest level of throttling limits writes
to be around the average WAL shipping rate on bare metal in our fleet.
Skips throttling for servers using old walg command set (where wal-g
runs as a subprocess of postgres archiver).

### Tests

- [x] Rhizome
- [x] Coverage
- [x] E2E under throttling scenario

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -330,6 +330,7 @@ class PostgresServer < Sequel::Model
     if session[:export_count] % 12 == 1
       observe_disk_usage(session)
       observe_archival_backlog(session)
+      observe_io_throttle(session)
       observe_metrics_backlog(session)
     end
 
@@ -478,6 +479,26 @@ class PostgresServer < Sequel::Model
     end
   rescue => ex
     Clog.emit("Failed to observe metrics backlog", Util.exception_to_hash(ex, into: {postgres_server_id: id}))
+  end
+
+  def observe_io_throttle(session)
+    # Make sure either the IO throttle file does not exist or if it exists, it
+    # has been updated in the last minute.
+    io_throttle_file = "/sys/fs/cgroup/system.slice/system-postgresql.slice/postgresql@#{version}-main.service/throttled/io.max"
+    result = session[:ssh_session].exec!("stat -c '%Y' :io_throttle_file 2>/dev/null || echo missing; date +%s", io_throttle_file:).strip.split("\n")
+    io_max_stat, vm_now_unix = result
+    if io_max_stat != "missing"
+      io_max_mtime = Time.at(Integer(io_max_stat, 10))
+      vm_now = Time.at(Integer(vm_now_unix, 10))
+      if vm_now - io_max_mtime > 120
+        Prog::PageNexus.assemble("#{ubid} I/O throttle stale",
+          ["PGIOThrottleStale", id], ubid,
+          severity: "warning", extra_data: {io_max_mtime:})
+      else
+        Page.from_tag_parts("PGIOThrottleStale", id)&.incr_resolve
+        Clog.emit("I/O throttle applied", {postgres_server_id: id, io_max_mtime:})
+      end
+    end
   end
 
   def observe_disk_usage(session)
--- a/spec/model/postgres/postgres_server_spec.rb
+++ b/spec/model/postgres/postgres_server_spec.rb
@@ -583,6 +583,7 @@ RSpec.describe PostgresServer do
       expect(postgres_server).to receive(:observe_archival_backlog).with(session)
       expect(postgres_server).to receive(:observe_metrics_backlog).with(session)
       expect(postgres_server).to receive(:observe_disk_usage).with(session)
+      expect(postgres_server).to receive(:observe_io_throttle).with(session)
 
       postgres_server.export_metrics(session:, tsdb_client:)
     end
@@ -593,6 +594,7 @@ RSpec.describe PostgresServer do
       expect(postgres_server).not_to receive(:observe_archival_backlog)
       expect(postgres_server).not_to receive(:observe_metrics_backlog)
       expect(postgres_server).not_to receive(:observe_disk_usage)
+      expect(postgres_server).not_to receive(:observe_io_throttle)
 
       postgres_server.export_metrics(session:, tsdb_client:)
     end
@@ -601,6 +603,7 @@ RSpec.describe PostgresServer do
       allow(postgres_server).to receive(:observe_archival_backlog).with(session)
       allow(postgres_server).to receive(:observe_metrics_backlog).with(session)
       allow(postgres_server).to receive(:observe_disk_usage).with(session)
+      allow(postgres_server).to receive(:observe_io_throttle).with(session)
       allow(postgres_server).to receive(:scrape_endpoints).and_return([])
 
       expect(session[:export_count]).to be_nil
@@ -986,6 +989,49 @@ RSpec.describe PostgresServer do
       expect(session[:ssh_session]).to receive(:_exec!).and_raise(Net::SSH::Exception.new("SSH error"))
       expect(Clog).to receive(:emit).with("Failed to observe disk usage", instance_of(Hash)).and_call_original
       postgres_server.observe_disk_usage(session)
+    end
+  end
+
+  describe "#observe_io_throttle" do
+    let(:session) {
+      {ssh_session: Net::SSH::Connection::Session.allocate}
+    }
+    let(:io_throttle_cmd) { "stat -c '%Y' /sys/fs/cgroup/system.slice/system-postgresql.slice/postgresql@16-main.service/throttled/io.max 2>/dev/null || echo missing; date +%s" }
+    let(:vm_now) { Time.now.to_i }
+
+    it "does nothing when io.max file is missing" do
+      expect(session[:ssh_session]).to receive(:_exec!).with(io_throttle_cmd).and_return("missing\n#{vm_now}\n")
+      expect(Prog::PageNexus).not_to receive(:assemble)
+      postgres_server.observe_io_throttle(session)
+    end
+
+    it "creates a stale page when io.max mtime is older than 120 seconds" do
+      old_mtime = vm_now - 121
+      expect(session[:ssh_session]).to receive(:_exec!).with(io_throttle_cmd).and_return("#{old_mtime}\n#{vm_now}\n")
+      expect(Prog::PageNexus).to receive(:assemble).with(
+        "#{postgres_server.ubid} I/O throttle stale",
+        ["PGIOThrottleStale", postgres_server.id],
+        postgres_server.ubid,
+        severity: "warning",
+        extra_data: {io_max_mtime: Time.at(old_mtime)}
+      )
+      postgres_server.observe_io_throttle(session)
+    end
+
+    it "resolves stale page and logs when io.max mtime is recent" do
+      recent_mtime = vm_now - 10
+      expect(session[:ssh_session]).to receive(:_exec!).with(io_throttle_cmd).and_return("#{recent_mtime}\n#{vm_now}\n")
+      page = Prog::PageNexus.assemble("#{postgres_server.ubid} I/O throttle stale",
+        ["PGIOThrottleStale", postgres_server.id], postgres_server.ubid, severity: "warning")
+      postgres_server.observe_io_throttle(session)
+      expect(Semaphore.where(strand_id: page.id, name: "resolve").count).to eq(1)
+    end
+
+    it "does not fail resolving stale page when no existing page" do
+      recent_mtime = vm_now - 10
+      expect(session[:ssh_session]).to receive(:_exec!).with(io_throttle_cmd).and_return("#{recent_mtime}\n#{vm_now}\n")
+      expect(Prog::PageNexus).not_to receive(:assemble)
+      postgres_server.observe_io_throttle(session)
     end
   end
 


rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  def apply_throttle(throttle_mbps)
+    return unless enable_io_controller
+    set_io_limit(throttle_mbps)
+    immune_pids = classify_processes
+    @logger.info("Applied I/O throttle: #{throttle_mbps} MB/s (immune pids: #{immune_pids.join(', ')})")
+  end
+
+  def find_postmaster_pid
+    output = r("ss -tlnp sport = :5432", expect: [0])
```
fdr: would this be more reliable via systemd units (which definitely has the child pid) or even enumeration of the cgroup? It seems weird to be doing so much with cgroups managing processes only to reach around and use a port binding to find postmaster.

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,134 @@
...
+    match = output.match(/pid=(\d+)/)
+    fail "Could not find postmaster PID listening on port 5432" unless match
+    match[1].to_i
+  end
+
+  def find_immune_pids
+    postmaster_pid = find_postmaster_pid
+    child_pids = r("pgrep -P #{postmaster_pid}", expect: [0, 1]).strip.split("\n").map(&:to_i).reject(&:zero?)
```
fdr: for code this detailed, similar to eliminating `ss`, you may find it worthwhile to work directly with `/proc` or other VFS and keep it all uniform (as you do below), e.g.:

```ruby
# my postmaster pid is 504481
File.read('/proc/504481/task/504481/children').split
```

It might eliminate some dross like expect: [0, 1]. You can handle file errors like ENOENT.

As a bonus here's how to find the parent pid:

```ruby
ppid = File.read("/proc/504481/status")[/^PPid:\s+(\d+)/, 1]
```

You may find working with cgroup VFS even more useful, but I'm not as familiar with that one...and your code already has lots of dependencies on it already.

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,134 @@
...
+      if IMMUNE_PATTERNS.any? { |pattern| cmdline.include?(pattern) }
+        immune_pids << pid
+      end
+    end
+    immune_pids
+  end
+
+  def get_cgroup_pids(cgroup_path)
+    procs_file = "#{cgroup_path}/cgroup.procs"
+    return [] unless File.exist?(procs_file)
```
fdr: Almost certainly should be a handler of ENOENT, a bit like how you handled ESRCH below. costs another syscall, and doesn't even solve the problem with a bad interleaving.

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,134 @@
...
+    File.write("#{cgroup_path}/cgroup.procs", pid.to_s)
+  rescue Errno::ESRCH
+    # Process no longer exists
+  end
+
+  private
+
+  def find_device_id(mount_path)
+    data_disk = r("findmnt -n -o SOURCE #{mount_path}").strip
+    data_disk = r("readlink -f #{data_disk}").strip
```
fdr: `File.realpath(data_disk)` spares you a fork. I have no worthwhile tricks for findmnt though

rhizome/postgres/bin/apply-io-throttle
```diff
@@ -0,0 +1,17 @@
...
+
+require_relative "../lib/io_throttle"
+require "logger"
+
+if ARGV.count != 2
+  fail "Usage: apply-io-throttle <instance> <throttle_mbps>"
+end
+
+instance = ARGV[0]
+throttle_mbps = ARGV[1].to_i
```
fdr: to_i bothers me more than just about everyone else because 'oops'.to_i yields "0". At least in this case consider Integer(ARGV[1], 10). You might want to review your other `to_i`s and consider when you would prefer a crash to an unexpected zero. I virtually never use `to_i` though I know other capable programmers are often dance closer to the margin.

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,134 @@
...
+  end
+
+  def remove_throttle
+    io_max_file = "#{@throttled_cgroup}/io.max"
+    unless File.exist?(io_max_file)
+      @logger.info("No throttle to remove")
+      return
+    end
+    File.write(io_max_file, "#{@dev_id} wbps=max")
+    @logger.info("Removed I/O throttle")
```
fdr: rescue ENOENT (see below)

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,134 @@
...
+    Dir.mkdir(@immune_cgroup) unless File.directory?(@immune_cgroup)
+
+    get_cgroup_pids(@service_cgroup).each do |pid|
+      move_pid_to_cgroup(pid, @throttled_cgroup)
+    end
+
+    File.write(subtree_control, "+io")
+    true
+  rescue Errno::ENOENT, Errno::EBUSY, Errno::EPERM, Errno::EACCES => e
+    @logger.warn("Cannot enable I/O controller: #{e.message}. Ensure Delegate=yes is set on the systemd service.")
```
fdr: Quite a bit of code ends up here in a number of error conditions, I wonder if it could be misleading (I guess it depends what you see in `e.message`. Presuming it's as terse as errstr, you may want to do future-someone a favor and print the line, file, syscall, something... that can identify from what expression the exception jumped from. This is splitting the difference of "rescuing and having contextually hyper-accurate messages for every system call" and "not sure what system call caused that..."

model/postgres/postgres_server.rb
```diff
@@ -486,9 +488,37 @@ def observe_disk_usage(session)
...
 
+  def apply_io_throttle(session, archival_backlog)
+    # Only apply I/O throttle when using the WAL-G daemon mode.
+    # With the old command set, wal-g runs as a subprocess of postgres archiver,
+    # so throttling postgresql@.service would also throttle archival itself.
+    return if resource.use_old_walg_command_set?
+
+    throttle_mbps = calculate_io_throttle_mbps(archival_backlog)
+    session[:ssh_session].exec!(
+      "sudo postgres/bin/apply-io-throttle :version :throttle_mbps",
```
byucesoy: Given majority of the logic happening in the VM, I wonder if we should have a process running in the VM to just check the backlog and apply throttling from there. At the moment, control plane gets the backlog size from VM, transforms it to IO throttle MBPS and pass it back to the VM. No need for round-trip, no?

fdr: that seems like a pretty good idea....you could even put it on a systemd timer if you wanted it to fork only occasionally (though this raises the spectre of oom preventing it from happening, while simultaneously leaving more memory for the user).

shikharbhardwaj: I had a version of this branch which used a local systemd timer instead, but decided against it because it might be harder to supervise vs something that goes through the control plane. I'll think on this a bit more.

serprex: this came up in #4332

it'd be nice if we could have these processes running on VMs, with metrics/status read up from controlplane, & in a way where these 2 PRs can share common mechanism

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,154 @@
...
+    "logger",       # Throttling affects archiver (logs go to stderr)
+    "checkpointer", # Checkpoints trigger WAL deletion
+    "writer"        # wal writer + background writer - allow checkpoints to complete
+  ].freeze
+
+  # I/O throttle tiers: [backlog_threshold, throttle_mbps], sorted descending
+  IO_THROTTLE_TIERS = [
+    [1000, 20],  # Critical: 1000+ files → 20 MB/s
+    [500, 50],   # Severe: 500-999 files → 50 MB/s
+    [100, 100]   # Moderate: 100-499 files → 100 MB/s
```
byucesoy: What is our default I/O speed (on AWS, Hetzner and Leaseweb)? Asking because if we are doing 500 MB/s without any throttling, cutting it to 100 MB/s would be too extreme.

I assume the baseline would be different for different providers due to hardware difference. Maybe we should pass a baseline (depending the provider) during initial provisioning and here we would define the tiers as baseline * A, baseline * B, etc. Or if the baseline is close to each other this might not be necessary.

shikharbhardwaj: Here's what we observe currently across providers, from data over the last 30 days. There is variance, so having different behavior across providers makes sense to me. However, I am not sure if some of these numbers (like leaseweb) are because the system there is not being pushed to its limits.

Disk write throughput

  | Top 10 average | Peak
-- | -- | --
Hetzner | 100 MB/s | 815 MB/s
AWS | 119 MB/s | 546 MB/s
Leaseweb | 35 MB/s | 357 MB/s


WAL Archival rate


  | Top 10 average | Peak
-- | -- | --
Hetzner | 0.6 /s | 4/s
AWS | 3 /s | 14/s
Leaseweb | 0.4/s | 8/s

I think the throttling limits have to take both into account. For eg: an instance under archival pressure on AWS can be allowed to write more than what one on other providers.

model/postgres/postgres_server.rb
shikharbhardwaj: In this case, we are trying to prevent 2 type of errors: the throttle is not applied when it should have been, and when it is unnecessarily applied. I believe the 2nd error is more problematic, so I've implemented a Page for that case. In case the throttling is just applied and being constantly updated by the timer, we emit a log for monitoring.

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,157 @@
...
+    dev_stat = File.stat(data_disk)
+    "#{dev_stat.rdev_major}:#{dev_stat.rdev_minor}"
+  end
+
+  def enable_io_controller
+    subtree_control = "#{@service_cgroup}/cgroup.subtree_control"
+    return true if File.read(subtree_control).include?("io")
+
+    Dir.mkdir(@throttled_cgroup) unless File.directory?(@throttled_cgroup)
+    Dir.mkdir(@immune_cgroup) unless File.directory?(@immune_cgroup)
```
fdr: is this FileUtils.mkdir_p or do you find some of mkdir_p (or requiring in FileUtils) undesirable?

model/postgres/postgres_server.rb
```diff
@@ -480,6 +481,24 @@ def observe_metrics_backlog(session)
...
   end
 
+  def observe_io_throttle(session)
+    # Make sure either the IO throttle file does not exist or if it exists, it
+    # has been updated in the last minute.
+    io_throttle_file = "/sys/fs/cgroup/system.slice/system-postgresql.slice/postgresql@#{version}-main.service/throttled/io.max"
+    io_max_stat = session[:ssh_session].exec!("stat -c '%Y' :io_throttle_file || echo 'missing'", io_throttle_file:).strip
+    if io_max_stat != "missing"
+      io_max_mtime = Time.at(Integer(io_max_stat, 10))
+      if Time.now - io_max_mtime > 60
```
byucesoy: I think we can give this a bit more headroom to avoid unnecessary pages. Maybe bump it to 2 minutes?

Also, it would be better to use the VM’s current time instead of `Time.now`. With possibility of clock drift and two different time sources, we could otherwise see some inconsistent behavior.

rhizome/postgres/lib/io_throttle.rb
```diff
@@ -0,0 +1,158 @@
...
+    # Process no longer exists
+  end
+
+  private
+
+  def calculate_throttle(backlog_count)
+    IO_THROTTLE_RATIOS.each do |threshold, ratio|
+      return (@disk_throughput_baseline_mbps * ratio).round if backlog_count >= threshold
+    end
+    0
```
byucesoy: One confusing thing about this ratio is that in `IO_THROTTLE_RATIOS`, lower value means more throttle but when we want to apply no throttle, we return lowest possible ratio: 0. Maybe this should return 1 or `nil` or something else.

shikharbhardwaj: Makes sense, updated to return nil

PR #4874, opened 2026-03-03
Add postgres endpoints for external mTLS

serprex:
cert/create-client-keypair creates keypair with a common name matching user name,
this can then be used to login to users in cert_auth_users list

cert/add-auth-user & cert/remove-auth-user are used to manage cert_auth_users list,
blocking postgres & ubi_replication to avoid unfortunate escalation

Importantly cert auth comes early in pg_hba, blocking password authentication

Certs split into server & client. Server used for authenticating server to clients,
while client certs are used only for authenticating to server. Thus,
replication connections now authenticate with client certs too

### DIFF
--- a/cli-commands/pg/post/add-cert-auth-user.rb
+++ b/cli-commands/pg/post/add-cert-auth-user.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("add-cert-auth-user") do
+  desc "Add user to list of users authenticating with client certificate authentication"
+
+  banner "ubi pg (location/pg-name | pg-id) add-cert-auth-user name"
+
+  args 1
+
+  run do |name|
+    data = sdk_object.add_cert_auth_user(name)
+    body = []
+    body << "Users using certificate authentication:\n"
+    data[:items].each_with_index do |user, i|
+      body << "  " << (i + 1).to_s << ": " << user << "\n"
+    end
+    response(body)
+  end
+end
--- a/cli-commands/pg/post/remove-cert-auth-user.rb
+++ b/cli-commands/pg/post/remove-cert-auth-user.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("remove-cert-auth-user") do
+  desc "Remove user from list of users authenticating with client certificate authentication"
+
+  banner "ubi pg (location/pg-name | pg-id) remove-cert-auth-user name"
+
+  args 1
+
+  run do |name|
+    data = sdk_object.remove_cert_auth_user(name)
+    body = []
+    body << "Users using certificate authentication:\n"
+    if data[:items].empty?
+      body << "No users using certificate authentication.\n"
+    else
+      data[:items].each_with_index do |user, i|
+        body << "  " << (i + 1).to_s << ": " << user << "\n"
+      end
+    end
+    response(body)
+  end
+end
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1095,7 +1095,7 @@ paths:
       summary: Download CA certificates for a PostgreSQL database
       responses:
         '200':
-          description: CA certificate file
+          description: CA certificate file in pem format
           headers:
             Content-Disposition:
               schema:
@@ -1106,6 +1106,87 @@ paths:
               schema:
                 type: string
                 format: binary
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/cert/add-auth-user':
+    parameters:
+      - $ref: '#/components/parameters/project_id'
+      - $ref: '#/components/parameters/location'
+      - $ref: '#/components/parameters/postgres_database_reference'
+    post:
+      operationId: addCertAuthUser
+      summary: Add provided name to list of users with certificate authentication
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CertAuthUserRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/CertAuthUserList'
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/cert/create-client-keypair':
+    parameters:
+      - $ref: '#/components/parameters/project_id'
+      - $ref: '#/components/parameters/location'
+      - $ref: '#/components/parameters/postgres_database_reference'
+    post:
+      operationId: createClientCertKeypair
+      summary: Create and return client certificate keypair with requested common name
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                common_name:
+                  description: CN for cert
+                  type: string
+                duration:
+                  description: 'Duration before expiry, in seconds'
+                  type: integer
+              additionalProperties: false
+      responses:
+        '200':
+          description: certificate keypair file in pem format
+          headers:
+            Content-Disposition:
+              schema:
+                type: string
+                example: attachment; filename="certificate-key.pem"
+          content:
+            application/x-pem-file:
+              schema:
+                type: string
+                format: binary
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/cert/remove-auth-user':
+    parameters:
+      - $ref: '#/components/parameters/project_id'
+      - $ref: '#/components/parameters/location'
+      - $ref: '#/components/parameters/postgres_database_reference'
+    post:
+      operationId: removeCertAuthUser
+      summary: Remove provided name from list of users with certificate authentication
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CertAuthUserRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/CertAuthUserList'
         default:
           $ref: '#/components/responses/Error'
       tags:
@@ -2524,6 +2605,25 @@ components:
           additionalProperties: false
           required:
             - pgbouncer_config
+    CertAuthUserRequest:
+      type: object
+      properties:
+        name:
+          description: name of user
+          type: string
+      additionalProperties: false
+      required:
+        - name
+    CertAuthUserList:
+      type: object
+      properties:
+        items:
+          type: array
+          items:
+            type: string
+      additionalProperties: false
+      required:
+        - items
     PostgresTag:
       type: object
       properties:
@@ -3635,6 +3735,12 @@ components:
             required:
               - count
               - items
+    CertAuthUserList:
+      description: full list of users using client certificate authentication
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/CertAuthUserList'
     PrivateSubnet:
       description: A Private Subnet
       content:
--- a/routes/project/location/postgres.rb
+++ b/routes/project/location/postgres.rb
@@ -488,6 +488,75 @@ class Clover
         certs
       end
 
+      r.on api?, "cert" do
+        authorize("Postgres:edit", pg)
+
+        r.post "create-client-keypair" do
+          common_name = typecast_params.nonempty_str!("common_name")
+          duration = typecast_params.pos_int!("duration")
+
+          if duration > 60 * 60 * 24 * 366 # cap duration to ~1 year
+            raise CloverError.new(400, "InvalidRequest", "Certificate expiry should be less than 367 days.")
+          elsif !common_name.match?(/\A[a-zA-Z0-9_-]{1,64}\z/)
+            raise CloverError.new(400, "InvalidRequest", "Common Name must only contain alphanumeric characters, underscores, and hyphens. It must not exceed 64 characters.")
+          elsif %w[postgres ubi_replication].include?(common_name)
+            raise CloverError.new(400, "InvalidRequest", "Common Name must not be postgres or ubi_replication.")
+          end
+
+          issuer_cert, issuer_key = pg.client_signing_key
+
+          cert_key = Util.create_certificate(
+            subject: "/C=US/O=None/CN=#{common_name}",
+            extensions: ["keyUsage=digitalSignature,keyEncipherment", "subjectKeyIdentifier=hash", "extendedKeyUsage=clientAuth"],
+            duration:,
+            issuer_cert:,
+            issuer_key:
+          ).map(&:to_pem)
+
+          audit_log(pg, "create_cert")
+
+          response.headers["content-disposition"] = "attachment; filename=\"#{common_name}.pem\""
+          response.content_type = :pem
+          cert_key.join
+        end
+
+        r.post "add-auth-user" do
+          name = typecast_params.nonempty_str!("name")
+          cert_auth_users = Sequel.pg_jsonb_op(:cert_auth_users)
+          DB.transaction do
+            n = pg.this
+              .exclude(cert_auth_users.contains([name]))
+              .update(cert_auth_users: cert_auth_users.concat([name]))
+            if n == 1
+              pg.servers.each(&:incr_configure)
+              audit_log(pg, "add_cert_auth_user")
+              pg.refresh
+            else
+              no_audit_log
+            end
+          end
+          {items: pg.cert_auth_users}
+        end
+
+        r.post "remove-auth-user" do
+          name = typecast_params.nonempty_str!("name")
+          cert_auth_users = Sequel.pg_jsonb_op(:cert_auth_users)
+          DB.transaction do
+            n = pg.this
+              .where(cert_auth_users.contains([name]))
+              .update(cert_auth_users: cert_auth_users - name)
+            if n == 1
+              pg.servers.each(&:incr_configure)
+              audit_log(pg, "remove_cert_auth_user")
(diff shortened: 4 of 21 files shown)


routes/project/location/postgres.rb
```diff
@@ -482,6 +482,26 @@ class Clover
...
+          subject: "/C=US/O=None/CN=#{common_name}",
+          extensions: ["keyUsage=digitalSignature,keyEncipherment", "subjectKeyIdentifier=hash", "extendedKeyUsage=serverAuth,clientAuth"],
+          duration:,
+          issuer_cert:,
+          issuer_key:
+        ).map(&:to_pem)
+
+        response.headers["content-disposition"] = "attachment; filename=\"#{pg.name}.pem\""
+        response.content_type = :pem
+        cert_key
```
jeremyevans: This likely isn't what you want. The `create_certificate` method returns an array, and I don't think you want to return an an array from the route.

```suggestion
        cert_key.join
```

routes/project/location/postgres.rb
```diff
@@ -482,6 +482,26 @@ class Clover
         certs
       end
 
+      r.post api?, "sign-cert-key" do
+        authorize("Postgres:edit", pg)
+        common_name = typecast_params.nonempty_str!("common_name")
+        duration = typecast_params.pos_int!("duration")
```
jeremyevans: Probably want to clamp this to something reasonable.

routes/project/location/postgres.rb
```diff
@@ -482,6 +482,26 @@ class Clover
         certs
       end
 
+      r.post api?, "sign-cert-key" do
+        authorize("Postgres:edit", pg)
+        common_name = typecast_params.nonempty_str!("common_name")
```
jeremyevans: Probably want to do some form of validation for this.

openapi/openapi.yml
```diff
@@ -1097,6 +1097,45 @@ paths:
...
+                common_name:
+                  description: CN for cert
+                  type: string
+                duration:
+                  description: Duration before expiry, in seconds
+                  type: integer
+              additionalProperties: false
+      responses:
+        '200':
+          description: certificate key file
```
jeremyevans: This makes it sound like it's just the key and not the certificate.

```suggestion
          description: certificate and key in pem format
```

openapi/openapi.yml
```diff
@@ -1097,6 +1097,45 @@ paths:
...
       tags:
         - Postgres Database
+  '/project/{project_id}/location/{location}/postgres/{postgres_database_reference}/sign-cert-key':
+    parameters:
+      - $ref: '#/components/parameters/project_id'
+      - $ref: '#/components/parameters/location'
+      - $ref: '#/components/parameters/postgres_database_reference'
+    post:
+      operationId: signCertWithName
+      summary: Download certificate key with requested common name
```
jeremyevans:
```suggestion
      summary: Create and return client certificate and key with requested common name
```

routes/project/location/postgres.rb
```diff
@@ -482,6 +482,26 @@ class Clover
...
+        duration = typecast_params.pos_int!("duration")
+
+        issuer_cert, issuer_key = pg.signing_key
+
+        cert_key = Util.create_certificate(
+          subject: "/C=US/O=None/CN=#{common_name}",
+          extensions: ["keyUsage=digitalSignature,keyEncipherment", "subjectKeyIdentifier=hash", "extendedKeyUsage=serverAuth,clientAuth"],
```
jeremyevans: Does a client certificate need `serverAuth` in `extendedKeyUsage`? Not a TLS expert, but that seems wrong.

serprex: Unlikely. Will test without

serprex:
re accepted root, standby already uses this root for mTLS connection to primary

aware of need for cli & tests, but like to get design feedback first on draft PRs. Prefer having some basic code to discuss over, otherwise slack discussions feel too abstract & everyone has a different idea in mind

re terminology, perhaps keypair would be more correct

fdr:
> If this is supposed to be used for mTLS, do we really want to use the same root for both the server and client certificates? That seems unwise to me. The client should verify the server's cert against the resource root, and the server should verify the client's cert against a different client root. Otherwise, it could make it easier to attempt connection as a client using a server cert, or to MITM a client using a different client cert. What are your thoughts, @fdr?

I think that the current arrangement was really for the benefit of replication-only (to avoid password versioning) a restricted use case. We could engage in two roots. The main consideration there is the cost of rotations, but intuitively, I don't see a problem with becoming more orthodox in this respect with more rich client cert features. It shouldn't be too much code.

fdr:
> As a reminder, our goal is that all API features are accessible through the CLI, so please add a CLI command for this.

I take this one as draft territory, but no harm writing it down if we move forward, which I think we will in some form or another.

serprex:
@jeremyevans is it desirable to include a GET endpoint for cert-auth-users? they are not otherwise shown anywhere besides the result of add/remove endpoints

jeremyevans:
> @jeremyevans is it desirable to include a GET endpoint for cert-auth-users? they are not otherwise shown anywhere besides the result of add/remove endpoints

I would only add it now if you have a need for it. It can always be added later.

serprex:
cool. we don't need it, on our end we're only going to be adding 2 cert users at start, tho eventually there's a desire for custom role management which also won't need GET

cli-commands/pg/post/add-cert-auth-user.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("add-cert-auth-user") do
+  desc "Add user to cert_auth_users"
```
jeremyevans: What is a `cert_auth_user`? This should explain what this does. Is it "Authorize a user to authenticate using a certificate", or something else?

sdk/ruby/lib/ubicloud/model/postgres.rb
```diff
@@ -169,5 +169,21 @@ def upgrade_status
...
     end
+
+    # Create a client certificate keypair with the given common name, expiring after duration seconds.
+    # Returns a PEM string containing the certificate and key.
+    def create_client_cert_keypair(common_name:, duration:)
+      adapter.post(_path("/create-client-cert-keypair"), common_name:, duration:)
+    end
+
+    # Add a user to cert_auth_users. Returns a hash with :items key listing all cert auth users.
+    def add_cert_auth_user(name:)
```
jeremyevans: For methods that require a single argument, use a positional argument instead of a keyword argument:

```suggestion
    def add_cert_auth_user(name)
```

sdk/ruby/lib/ubicloud/model/postgres.rb
```diff
@@ -169,5 +169,21 @@ def upgrade_status
...
+      adapter.post(_path("/add-cert-auth-user"), name:)
+    end
+
+    # Remove a user from cert_auth_users. Returns a hash with :items key listing remaining cert auth users.
+    def remove_cert_auth_user(name:)
```
jeremyevans:
```suggestion
    def remove_cert_auth_user(name)
```

openapi/openapi.yml
```diff
@@ -1057,6 +1057,46 @@ paths:
...
+              schema:
+                type: object
+                properties:
+                  items:
+                    type: array
+                    items:
+                      type: string
+                additionalProperties: false
+                required:
+                  - items
```
jeremyevans: This is the same as `remove-cert-auth-user`, so both should use `$ref` and there should be a single description.

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,61 @@ class Clover
...
+          raise CloverError.new(400, "InvalidRequest", "Certificate expiry should be less than 367 days.")
+        elsif !common_name.match?("\\A[a-zA-Z0-9_-]+\\z")
+          raise CloverError.new(400, "InvalidRequest", "Common Name must only contain alphanumeric characters, underscores, and hyphens.")
+        elsif common_name.length > 64
+          raise CloverError.new(400, "InvalidRequest", "Common Name must not exceed 64 characters.")
+        elsif %w[postgres ubi_replication].include?(common_name)
+          raise CloverError.new(400, "InvalidRequest", "Common Name must not be postgres or ubi_replication.")
+        end
+
+        issuer_cert, issuer_key = pg.signing_key
```
jeremyevans: This still uses the same CA root for both client certificates and server certificates. I think that's a mistake, especially since we allow server certificates to be used as client certificates currently due to replication.

@fdr and I discussed this yesterday. There should only be a couple of changes needed to support a separate client CA root distinct from the server CA root:

* Add columns to store client CA roots (for existing databases, these can be the same as the server CA roots)
* For replication certs, create a client cert instead of reusing the server cert.
* Make server certs not support clientAuth.

cli-commands/pg/post/remove-cert-auth-user.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("remove-cert-auth-user") do
+  desc "Remove user to cert_auth_users"
```
jeremyevans: Same description issue here.

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,61 @@ class Clover
         certs
       end
 
+      r.post api?, "create-client-cert-keypair" do
+        authorize("Postgres:edit", pg)
+        audit_log(pg, "create_cert")
```
jeremyevans: Move to near the end of method. Currently, if there is an error in the common name or duration, it will add a `create_cert` audit log when a cert was not actually created.

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,61 @@ class Clover
...
+
+      r.post api?, "add-cert-auth-user" do
+        authorize("Postgres:edit", pg)
+        name = typecast_params.nonempty_str!("name")
+        unless pg.cert_auth_users.include? name
+          pg.update(cert_auth_users: pg.cert_auth_users + [name])
+          pg.save_changes
+          pg.servers.each(&:incr_configure)
+        end
+        audit_log(pg, "update")
```
jeremyevans: Possible race conditions with concurrent updates (removing previously added user or adding the same user twice). Additionally, should use a transaction:

```suggestion
          cert_auth_users = Sequel.pg_jsonb_op(:cert_auth_users)
          DB.transaction do
            n = pg.this
              .exclude(cert_auth_users.contains([name])
              .update(cert_auth_users: cert_auth_users.concat([name]))
            if n == 1
              pg.servers.each(&:incr_configure)
              audit_log(pg, "update") # Consider new audit log action name
              pg.refresh
            else
              no_audit_log
            end
          end
```

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,61 @@ class Clover
...
+      end
+
+      r.post api?, "remove-cert-auth-user" do
+        authorize("Postgres:edit", pg)
+        name = typecast_params.nonempty_str!("name")
+        if pg.cert_auth_users.include? name
+          pg.update(cert_auth_users: pg.cert_auth_users - [name])
+          pg.servers.each(&:incr_configure)
+        end
+        audit_log(pg, "update")
```
jeremyevans: Similar race condition/transaction issue as add-cert-auth-user, use a similar approach to fixing here..

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,61 @@ class Clover
         certs
       end
 
+      r.post api?, "create-client-cert-keypair" do
```
jeremyevans: Maybe better to group these under a `cert` prefix:

* `cert/create-client-keypair`
* `cert/add-auth-user`
* `cert/remove-auth-user`

In addition to grouping related actions, this would allow you to authorize and do `api?` check in one place instead of 3 places.

cli-commands/pg/post/remove-cert-auth-user.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  args 1
+
+  run do |name|
+    data = sdk_object.remove_cert_auth_user(name:)
+    body = []
+    body << "Users using certificate authentication:\n"
+    data[:items].each_with_index do |user, i|
+      body << "  " << (i + 1).to_s << ": " << user << "\n"
+    end
```
jeremyevans, line 20: If no users left are using cert auth, consider returning a result stating that: `"No users using certificate authentication.\n"`.

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,61 @@ class Clover
...
+
+        common_name = typecast_params.nonempty_str!("common_name")
+        duration = typecast_params.pos_int!("duration")
+
+        if duration > 60 * 60 * 24 * 366 # cap duration to ~1 year
+          raise CloverError.new(400, "InvalidRequest", "Certificate expiry should be less than 367 days.")
+        elsif !common_name.match?("\\A[a-zA-Z0-9_-]+\\z")
+          raise CloverError.new(400, "InvalidRequest", "Common Name must only contain alphanumeric characters, underscores, and hyphens.")
+        elsif common_name.length > 64
+          raise CloverError.new(400, "InvalidRequest", "Common Name must not exceed 64 characters.")
```
jeremyevans: You should use a literal regexp here:

```suggestion
        elsif !common_name.match?(/\A[a-zA-Z0-9_-]+\z/)
          raise CloverError.new(400, "InvalidRequest", "Common Name must only contain alphanumeric characters, underscores, and hyphens.")
        elsif common_name.length > 64
          raise CloverError.new(400, "InvalidRequest", "Common Name must not exceed 64 characters.")
```

You could also combine the conditions, but if you want to keep them separate for more precise error messages, that's fine:

```suggestion
        elsif !common_name.match?(/\A[a-zA-Z0-9_-]{1,64}\z/)
          raise CloverError.new(400, "InvalidRequest", "Common Name must only contain alphanumeric characters, underscores, and hyphens and must not exceed 64 characters.")
```

Note that public CAs no longer allow underscores in CN in certificates. I assume we are allowing them here because they are allowed in Postgres user names, and the CN needs to match the user name?

serprex: > Note that public CAs no longer allow underscores in CN in certificates. I assume we are allowing them here because they are allowed in Postgres user names, and the CN needs to match the user name?

yes

routes/project/location/postgres.rb
serprex: Can you describe the practical issue here? I'm not seeing the value

jeremyevans: Using a separate CA root for clients and servers adds an additional level of security. Without it, servers can more easily impersonate clients using server certs and clients can more easily impersonate servers using client certs.

There are a other layers of protection against impersonation that are used regardless of whether the client and server CA roots are the same or separate:

* extendedKeyUsage clientAuth/serverAuth, but that requires the SSL library handle that correctly (probably true for the server, questionable to rely on for the client). Additionally, as server certs currently support clientAuth to allow the use of the server cert as a client cert for replication purposes, it would allow use of server certs as client certs.
* Server CN checking of client certificates (server cert CN unlikely to match database username)
* Client checking of server certificate names (only done with verify-full, not with verify-ca or lesser modes). Note that our recommended configuration parameter for client PG connections is `channel_binding=require` and the recommend configuration parameter for pgBouncer connections is `sslmode=require`, neither of which will check the server certificate name.

Using a shared CA root is not insecure, it just less secure. All else being equal, more layers of security is better. Considering it's simple to switch to separate client/server CA roots, I think the benefits outweigh the costs.

serprex: updated with split happening in latest 2 commits. if this looks good to you I can go ahead & organize commits to be ready for merging

jeremyevans: This looks good, though the second comment was a bit hard to review as it mixes removal of `postgres_server` from a lot of lines.

For the SSL file names that changed (e.g. old servers: `ca.crt`, new servers: `server-ca.crt`), will that cause any problems for existing servers? It looks like `refresh_certificates` is run, it will generate the new file names, it would still use the old file names until the server was reconfigured. If the server was reconfigured before `refresh_certificates` is run, it looks like it would reference files not yet created.

serprex: you'll have problems anyways on standbys when replication string wants `client.key` & `client.crt`. after deployment running `incr_refresh_certificates` should be done to push those files, while relying on client/server aliasing on existing servers to keep everything else working

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,75 @@ class Clover
...
+        r.post "add-auth-user" do
+          name = typecast_params.nonempty_str!("name")
+          cert_auth_users = Sequel.pg_jsonb_op(:cert_auth_users)
+          DB.transaction do
+            n = pg.this
+              .exclude(cert_auth_users.contains([name]))
+              .update(cert_auth_users: cert_auth_users.concat([name]))
+            if n == 1
+              pg.servers.each(&:incr_configure)
+              audit_log(pg, "update")
```
jeremyevans: We are making a change to use more specific audit log action names when possible (#4960).

```suggestion
              audit_log(pg, "add_cert_auth_user")
```

Add `add_cert_auth_user` to the allowed audit actions.

routes/project/location/postgres.rb
```diff
@@ -488,6 +488,75 @@ class Clover
...
+        r.post "remove-auth-user" do
+          name = typecast_params.nonempty_str!("name")
+          cert_auth_users = Sequel.pg_jsonb_op(:cert_auth_users)
+          DB.transaction do
+            n = pg.this
+              .where(cert_auth_users.contains([name]))
+              .update(cert_auth_users: cert_auth_users - name)
+            if n == 1
+              pg.servers.each(&:incr_configure)
+              audit_log(pg, "update")
```
```suggestion
              audit_log(pg, "remove_cert_auth_user")
```

Add `remove_cert_auth_user` to the allowed audit actions.

PR #4876, opened 2026-03-03
Fix chained CSI migration and add E2E coverage

mohi-kalantari:
When a node is cordoned mid-migration, the intermediate PV never finishes staging so node_unstage_volume is never called on it. This leaves the PVC stuck — bound to a PV on an unschedulable node with no recovery path. The StuckVolumeDetector addresses this by periodically scanning for PVCs with the old-pv-name annotation that are bound to PVs on cordoned nodes. When detected, it rolls the intermediate PV's reclaim policy to Delete, resets the retry count on the source PV, and recreates the PVC so migration can resume on a healthy node. It also handles the race where the StatefulSet controller recreates the PVC before the detector does.

The trim_pvc method and shared annotation constants are consolidated into ServiceHelper to eliminate duplication between NodeService and StuckVolumeDetector.

pending_pvs now filters by Retain reclaim policy so intermediate PVs rolled back to Delete no longer block wait_for_copy and prevent node removal.

The PVC RBAC on the ubicsi-provisioner ClusterRole is extended with create, delete, and patch verbs to support the recovery flow.

A new E2E test for chained migration is added: it cordons the source node, waits for rsync to start on a second node, then cordons that node too — verifying the pod eventually runs on a third node with correct data.

Docker images are bumped to v0.8.0.

### DIFF
--- a/kubernetes/csi/Gemfile.lock
+++ b/kubernetes/csi/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ubi-csi (0.7.0)
+    ubi-csi (0.8.0)
       base64
       grpc (~> 1.76)
       grpc-tools (~> 1.76)
--- a/kubernetes/csi/lib/ubi_csi.rb
+++ b/kubernetes/csi/lib/ubi_csi.rb
@@ -5,7 +5,7 @@ require "csi_pb"
 require "csi_services_pb"
 
 module Csi
-  VERSION = "0.7.0"
+  VERSION = "0.8.0"
 end
 
 require_relative "ubi_csi/identity_service"
--- a/kubernetes/csi/spec/csi/v1/identity_service_spec.rb
+++ b/kubernetes/csi/spec/csi/v1/identity_service_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe Csi::V1::IdentityService do
       request = Csi::V1::GetPluginInfoRequest.new
       response = service.get_plugin_info(request, call)
       expect(response.name).to eq("csi.ubicloud.com")
-      expect(response.vendor_version).to eq("0.7.0")
+      expect(response.vendor_version).to eq("0.8.0")
     end
   end
 
--- a/kubernetes/csi/ubi-csi.gemspec
+++ b/kubernetes/csi/ubi-csi.gemspec
@@ -2,7 +2,7 @@
 
 Gem::Specification.new do |spec|
   spec.name = "ubi-csi"
-  spec.version = "0.7.0"
+  spec.version = "0.8.0"
   spec.authors = ["Ubicloud"]
   spec.email = ["support@ubicloud.com"]
 
--- a/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/nodeplugin-daemonset.yaml
@@ -22,7 +22,7 @@ spec:
       priorityClassName: system-node-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.7.0
+          image: ubicloud/ubicsi:v0.8.0
           imagePullPolicy: IfNotPresent
           args: ["node"]
           env:
--- a/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
+++ b/rhizome/kubernetes/manifests/ubicsi/provisioner-deployment.yaml
@@ -38,7 +38,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
         - name: ubicsi
-          image: ubicloud/ubicsi:v0.7.0
+          image: ubicloud/ubicsi:v0.8.0
           imagePullPolicy: IfNotPresent
           args: ["controller"]
           env:


mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22719213632

mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22768036952

kubernetes/csi/spec/ubi_csi/stuck_volume_detector_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+require "spec_helper"
+
+RSpec.describe Csi::StuckVolumeDetector do
+  let(:logger) { Logger.new(File::NULL) }
+  let(:detector) { described_class.new(logger:) }
+
+  describe "#shutdown!" do
+    it "sets shutdown flag, closes queue, and joins thread" do
+      thread = instance_double(Thread)
+      detector.instance_variable_set(:@thread, thread)
```
jeremyevans: Probably better to mock `check_stuck_volumes` to do nothing, then call `start`, then call `shutdown!` and confirm the thread shuts down.

kubernetes/csi/spec/ubi_csi/stuck_volume_detector_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+  end
+
+  describe "#start" do
+    it "spawns thread and logs startup" do
+      expect(detector).to receive(:spawn_check_thread)
+
+      detector.start
+    end
+  end
+
```
jeremyevans: Not needed if you are actually calling `start` in the earlier spec.

kubernetes/csi/spec/ubi_csi/stuck_volume_detector_spec.rb
```diff
@@ -0,0 +1,314 @@
+# frozen_string_literal: true
+
+require "logger"
+require "spec_helper"
+
+RSpec.describe Csi::StuckVolumeDetector do
+  let(:logger) { Logger.new(File::NULL) }
+  let(:detector) { described_class.new(logger:) }
```
jeremyevans, line 8: Should probably have an `after` block that shuts down the detector. Otherwise, if any spec fails, they could potentially leak threads.

kubernetes/csi/spec/ubi_csi/stuck_volume_detector_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+    end
+  end
+
+  describe "#spawn_check_thread" do
+    it "runs check_stuck_volumes in a thread" do
+      sync = Queue.new
+      expect(detector).to receive(:check_stuck_volumes) { sync.push(true) }
+
+      detector.spawn_check_thread
+      sync.pop
```
jeremyevans: `expect(sync.pop(timeout: 1)).to be true`. Otherwise, on failure this can hang.

kubernetes/csi/spec/ubi_csi/stuck_volume_detector_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+      trimmed_pvc = {"metadata" => {"namespace" => "default", "name" => "data-pvc", "annotations" => {"csi.ubicloud.com/old-pv-name" => "pvc-xxx"}}, "spec" => {"accessModes" => ["ReadWriteOnce"]}}
+      old_pvc_object_patch = {"metadata" => {"annotations" => {"csi.ubicloud.com/old-pvc-object" => Base64.strict_encode64(YAML.dump(trimmed_pvc))}}}.to_json
+      expect(Open3).to receive(:capture2e).with("kubectl", "get", "pv", "pvc-xxx", "-oyaml", stdin_data: nil).and_return([YAML.dump(source_pv), success_status])
+      expect(Open3).to receive(:capture2e).with("kubectl", "patch", "pv", "pvc-xxx", "--type=merge", "-p", old_pvc_object_patch, stdin_data: nil).and_return(["patched", success_status])
+      old_pv_patch = {"metadata" => {"annotations" => {"csi.ubicloud.com/old-pv-name" => "pvc-xxx"}}}.to_json
+      expect(Open3).to receive(:capture2e).with("kubectl", "-n", "default", "patch", "pvc", "data-pvc", "--type=merge", "-p", old_pv_patch, stdin_data: nil).and_return(["patched", success_status])
+      expect(Open3).not_to receive(:capture2e).with("kubectl", "-n", "default", "delete", "pvc", "data-pvc", "--wait=false", "--ignore-not-found=true", stdin_data: nil)
```
jeremyevans: `not_to receive` with arguments is prone to false negatives if the arguments change. In general, you want to push `to receive` to the lowest level with arguments, and `not_to receive` to the highest level (omitting arguments if possible).

kubernetes/csi/spec/ubi_csi/stuck_volume_detector_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+      expect(detector).to receive(:recover_stuck_pvc).ordered.and_raise(StandardError.new("recovery failed"))
+      expect(detector).to receive(:recover_stuck_pvc).ordered
+
+      detector.check_stuck_volumes
+    end
+  end
+
+  describe "#recover_stuck_pvc" do
+    let(:success_status) { instance_double(Process::Status, success?: true) }
+    let(:failed_status) { instance_double(Process::Status, success?: false) }
```
jeremyevans: Might want to move these two the top level, since they appear to be used by multiple specs.

mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22848948033

kubernetes/csi/lib/ubi_csi/stuck_volume_detector.rb
```diff
@@ -0,0 +1,121 @@
...
+          @logger.info("[StuckVolumeDetector] Deleted PVC #{pvc_namespace}/#{pvc_name}")
+          client.remove_pvc_finalizers(pvc_namespace, pvc_name)
+          @logger.info("[StuckVolumeDetector] Removed PVC finalizers #{pvc_namespace}/#{pvc_name}")
+        end
+        begin
+          client.create_pvc(trimmed_pvc)
+        rescue => e
+          raise unless e.message.include?("AlreadyExists")
+          # StatefulSet controller recreated PVC before us, just patch the annotation
+          client.patch_resource("pvc", pvc_name, OLD_PV_NAME_ANNOTATION_KEY, source_pv_name, namespace: pvc_namespace)
```
aamederen, line 109: is this some duplicate logic here? I think we have a similar logic in node_service

kubernetes/csi/lib/ubi_csi/stuck_volume_detector.rb
```diff
@@ -0,0 +1,121 @@
...
+        bound_pv = pvs_by_name[bound_pv_name]
+        next unless bound_pv
+
+        pv_node = client.extract_node_from_pv(bound_pv)
+        next unless pv_node
+        next if client.node_schedulable?(pv_node)
+
+        pvc_namespace = pvc.dig("metadata", "namespace")
+        pvc_name = pvc.dig("metadata", "name")
+        @logger.info("[StuckVolumeDetector] Detected stuck PVC #{pvc_namespace}/#{pvc_name} bound to PV #{bound_pv_name} on cordoned node #{pv_node}")
```
aamederen, line 62: High level thought: Instead of checking the code here, can we implement the logic in `migrate_pvc_data` of NodeService? It already contains some migration monitoring logic.

prog/test/kubernetes.rb
```diff
@@ -3,7 +3,7 @@
 require_relative "../../lib/util"
 
 class Prog::Test::Kubernetes < Prog::Test::Base
-  MIGRATION_TRIES = 2
+  MIGRATION_TRIES = 1
```
aamederen, line 6: why do we need to reduce the tries?

kubernetes/csi/lib/ubi_csi/stuck_volume_detector.rb
aamederen: We discussed this with Mohi. The CSI is removed from the target node on drain and we rely on Kubelet putting a pod that consumes the PV to signal the migration to succeed, so there's no way to know, apart from periodic checking.

prog/test/kubernetes.rb
mohi-kalantari: Since a new chained migration scenario is added to the e2e tests, I wanted to keep the total duration of k8s e2e tests the same, that's why I reduced it to one.

kubernetes/csi/lib/ubi_csi/stuck_volume_detector.rb
mohi-kalantari: no not exactly. In that function, we might either see that
- the pvc exists and then patch
- the pvc does not exist, we try to create, fail because in the meantime sts controller created it and then we patch

PR #4878, opened 2026-03-03
Allow modifying pool timeouts per process

jeremyevans:
The reason for this is to give monitor a higher timeout than used for web, since monitor has a much higher tolerance for delays.

This doesn't actually matter for respirate, due to its use of coarse checkouts. The configuration value for respirate is only added to make the db.rb code simpler.


enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22642271872

PR #4880, opened 2026-03-04
Use the Sequel lit_require_frozen extension during unfrozen testing

jeremyevans:
This extension raises errors for unsafe interpolation into literal strings, such as:

```ruby
DB["SELECT * FROM table WHERE column = '#{value}'"]
DB[:table].where(Sequel.lit("column = '#{value}'"))
```

This doesn't detect any violations, which is good. There are violations in the migrate directory, so I turned this off for migrations. Migrations shouldn't be dealing dealing with user-provider input, so the risk is fairly low there.

This requires updating the Sequel checkout.


enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22669467147

fdr (approved):

PR #4882, opened 2026-03-04
Fix backup credential race during postgres upgrade

furkansahin:
During a postgres version upgrade, the timeline nexus and server nexus run as independent strands. The timeline nexus could reach take_backup before the server nexus had written the new timeline's credentials to wal-g.env, causing the backup to push to the wrong bucket.

Use switch_to_new_timeline in the upgrade flow to write credentials synchronously before the timeline strand can trigger a backup.


furkansahin:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22669270756

PR #4307, opened 2025-12-04
Extend i8g sizes, while adding i8ge, i7i, & i7ie SKUs

serprex:
> [!IMPORTANT]
> Adds support for AWS instance types `i8ge`, `i7i`, and `i7ie`, updating options, feature flags, and CLI documentation.
> 
>   - **Behavior**:
>     - Adds support for AWS instance types `i8ge`, `i7i`, and `i7ie` in `helpers/postgres.rb` and `lib/option.rb`.
>     - Updates `generate_postgres_options()` in `helpers/postgres.rb` to include new instance types.
>   - **Options**:
>     - Updates `AWS_FAMILY_OPTIONS` and `AWS_STORAGE_SIZE_OPTIONS` in `lib/option.rb` to include `i8ge`, `i7i`, and `i7ie`.
>     - Adds new size options for `i8ge`, `i7i`, and `i7ie` in `POSTGRES_SIZE_OPTIONS` in `lib/option.rb`.
>   - **Feature Flags**:
>     - Adds feature flags `enable_i7i`, `enable_i7ie`, and `enable_i8ge` in `model/project.rb`.
>   - **CLI Documentation**:
>     - Updates CLI help files in `spec/routes/api/cli/golden-files/` to include new instance types in allowed option values.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ubicloud%2Fubicloud&utm_source=github&utm_medium=referral)<sup> for afed01db695bd1216534c29aad6f07c880475433. You can [customize](https://app.ellipsis.dev/ubicloud/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #4891, opened 2026-03-04
Support context-sensitive help for the CLI

jeremyevans:
This allows help option text in the CLI to be context-sensitive, returning different values depending on the user or project.

This gives an example of how it is used in the vm create command:

```ruby
  help_option_values("Size:") { server_sizes }
```

The block here is evaluated in the context of the UbiCli instance, the same context that the run block is exectued in.

Note that the CLI does not implement any authorization, only authentication. The internal request to the API handles authorization. This is the reason direct database access from the CLI layer is not allowed.

In order to get the above code to correctly show which sizes are available to the current user/project, we would need to expose an API endpoint that lists the allowed sizes (because direct querying is not allowed). So this isn't currently useful by itself, but it is a necessary change if we do want to support context-specific help.

I'm currently leaving this in draft, because I don't think we should merge it until it's actually used to filter allowed sizes for VMs or PostgresResources.

Related to #4257, #4307, #4737

### DIFF
--- a/cli-commands/vm/post/create.rb
+++ b/cli-commands/vm/post/create.rb
@@ -17,7 +17,7 @@ UbiCli.on("vm").run_on("create") do
     on("-u", "--unix-user=username", "username (default: ubi)")
   end
   help_option_values("Boot Image:", Option::BootImages.map(&:name))
-  help_option_values("Size:", server_sizes)
+  help_option_values("Size:") { server_sizes }
   help_option_values("Storage Size:", storage_sizes)
 
   help_example 'ubi vm eu-central-h1/my-vm-name create "$(cat ~/.ssh/id_ed25519.pub)"'
--- a/cli-commands/help.rb
+++ b/cli-commands/help.rb
@@ -29,9 +29,7 @@ UbiCli.on("help") do
               body << banner << "\n"
             end
           else
-            # When we want to switch to context-sensitive help
-            # body << cmd.context_help(self) << "\n\n"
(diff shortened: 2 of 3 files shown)


cli-commands/vm/post/create.rb
```diff
@@ -17,7 +17,7 @@
     on("-u", "--unix-user=username", "username (default: ubi)")
   end
   help_option_values("Boot Image:", Option::BootImages.map(&:name))
-  help_option_values("Size:", server_sizes)
+  help_option_values("Size:") { server_sizes }
```
byucesoy, line 20: So my understanding is that when we pass server_sizes as block, we hit this [line](https://github.com/jeremyevans/rodish/blob/eed47a1fd57aadcc107372d9db5354324064dae7/lib/rodish/plugins/_context_sensitive_help.rb#L28), right?

byucesoy (approved):
Sorry, I missed this PR because it was draft. I think the approach makes sense (if I understand it correctly)

cli-commands/vm/post/create.rb
jeremyevans: Correct. That line  ends up calling [context.instance_exec(&@block)](https://github.com/jeremyevans/rodish/blob/eed47a1fd57aadcc107372d9db5354324064dae7/lib/rodish/plugins/_context_sensitive_help.rb#L18), which evaluates the block in the same context as the `run` blocks.

PR #4892, opened 2026-03-04
Reduce false deadline pages for recycling Postgres servers

serprex:
Use a 30-minute deadline (instead of 5 minutes) when a Postgres server needs recycling, since recycling servers are replaced rather than recovered. Also add resource context (name, role, location, recycling state) to deadline page extra_data for PostgresServer subjects.


byucesoy:
Primary servers can be in state of needing recycling (e.g. during resizing). We shouldn't extend deadlines or we would miss unavailability events.

Am I missing something? @serprex @furkansahin

PR #4894, opened 2026-03-05
Adjust requested hugepage count by a safety ratio

pykello:
On a production host with 256 GB of RAM we observed only ~250 GB of total memory, indicating that usable memory can be roughly up to 1/43 lower than the reported physical memory.

To avoid overcommitting memory when allocating hugepages and leaving too little regular memory available, this change applies a safety ratio of 42/43 to physical memory instead of assuming the entire amount is usable.

This ratio also satisfies our requirement of being able to fit premium-30 VMs on AX 102 hosts.

We use the physical memory with a safety ratio instead of reported total memory to get more predictable number of hugepages.

Note that this applies only to future host setups; existing hosts must be adjusted manually.


fdr:
@furkansahin I think you often (always?!) are involved in leaseweb expansions, so...fyi, a subtle change here that _shouldn't_ matter, but we know how it goes.

PR #4895, opened 2026-03-05
Encrypt swap partition during host setup

pykello:
Set up dm-crypt on the swap partition with an ephemeral /dev/urandom key (regenerated each boot) so swap contents are unreadable after reboot or on physical disk removal. Prevents process memory that may contain secrets from being written to disk as plaintext.

### DIFF
--- a/rhizome/host/bin/encrypt-swap
+++ b/rhizome/host/bin/encrypt-swap
@@ -1,0 +1,6 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "../lib/crypt_swap_setup"
+
+CryptSwapSetup.run
--- a/rhizome/host/bin/prep_host.rb
+++ b/rhizome/host/bin/prep_host.rb
@@ -4,6 +4,7 @@
 require_relative "../../common/lib/arch"
 require_relative "../../common/lib/util"
 require_relative "../lib/cloud_hypervisor"
+require_relative "../lib/crypt_swap_setup"
 require_relative "../lib/spdk_setup"
 require "fileutils"
 require "socket"
@@ -99,6 +100,8 @@ r "wget https://github.com/ubicloud/htcat/releases/download/v2.0.0-ubi1/htcat_2.
 r "tar xvf htcat.tar.gz -C /usr/local/bin/"
 
 SpdkSetup.prep
+
+CryptSwapSetup.run
 
 # cron job to store serial.log files
 FileUtils.mkdir_p("/var/log/ubicloud/serials")
--- a/prog/test/hetzner_server.rb
+++ b/prog/test/hetzner_server.rb
@@ -80,6 +80,13 @@ class Prog::Test::HetznerServer < Prog::Test::Base
     end
     update_stack({"available_storage_gib" => vm_host.available_storage_gib})
 
+    hop_verify_encrypted_swap
+  end
+
+  label def verify_encrypted_swap
+    sshable = vm_host.sshable
+    swap_device = sshable.cmd("swapon --show=NAME --noheadings").strip
+    fail_test "swap is not on a dm-crypt device: #{swap_device}" unless swap_device.start_with?("/dev/dm-")
     hop_install_integration_specs
   end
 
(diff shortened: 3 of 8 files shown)


rhizome/host/bin/encrypt-swap
```diff
@@ -0,0 +1,58 @@
...
+# Ephemeral key via /dev/urandom (regenerated each boot); size=512 means AES-256-XTS (XTS halves the key).
+crypttab_entry = "cryptswap #{by_id} /dev/urandom cipher=aes-xts-plain64,size=512,swap,discard\n"
+if File.exist?("/etc/crypttab")
+  # backup crypttab before modifying in case something goes wrong
+  FileUtils.cp("/etc/crypttab", "/etc/crypttab.bak.#{Time.now.to_i}")
+  crypttab = File.read("/etc/crypttab")
+else
+  crypttab = ""
+end
+safe_write_to_file("/etc/crypttab", crypttab.gsub(/^cryptswap\s.*\n?/, "") + crypttab_entry)
```
jeremyevans:
```suggestion
end
safe_write_to_file("/etc/crypttab", crypttab.gsub(/^cryptswap\s.*\n?/, "") + crypttab_entry.to_s)
```

rhizome/host/bin/encrypt-swap
```diff
@@ -0,0 +1,58 @@
...
+
+# backup fstab before modifying in case something goes wrong
+FileUtils.cp("/etc/fstab", "/etc/fstab.bak.#{Time.now.to_i}")
+safe_write_to_file("/etc/fstab",
+  fstab.sub(swap_line, "/dev/mapper/cryptswap none swap sw 0 0\n"))
+
+# Prevent interactive prompt about existing swap signature on the backing device.
+r "wipefs -a #{swap_real.shellescape}"
```
jeremyevans: Better to avoid a shell instead of using shellescape.
```suggestion
r "wipefs", "-a", swap_real
```

rhizome/host/bin/encrypt-swap
```diff
@@ -0,0 +1,58 @@
...
+swap_real = File.realpath(swap_device)
+by_id = Dir["/dev/disk/by-id/*"]
+  .find { |p| File.identical?(p, swap_real) }
+
+unless by_id
+  warn "No /dev/disk/by-id entry found for #{swap_device} (real=#{swap_real})"
+  exit 1
+end
+
+r "swapoff #{swap_real.shellescape}"
```
jeremyevans:
```suggestion
r "swapoff", swap_real
```

rhizome/host/bin/prep_host.rb
```diff
@@ -100,6 +100,8 @@
 
 SpdkSetup.prep
 
+r File.join(__dir__, "encrypt-swap").to_s
```
jeremyevans: `File.join` should return a string.

```suggestion
r File.join(__dir__, "encrypt-swap")
```

jeremyevans (commented):
Looks good from a Ruby perspective. I'll leave approval to @fdr.

fdr (commented):
why make a bin and then shell out? it's all ruby, it's all rhizome...why not call a method? This might also give you a more natural place to put some tests, although to look at it, it's a fair question how much those tests would prove.

fdr (approved):
nevertheless, encrypted swap is better, and I see this only hits new hosts which is a bit less delicate. a bit odd to shell out to rhizome though.

pykello:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23028396461

pykello:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24734142402

pykello:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24749303969

pykello:
@fdr the flow is the same as the version you already approved. Changes:

- Moved the logic from an executable from `host/bin` to a module in `host/lib` so we don't shell out.
- Reused the persistent device id lookup which we already had in `StorageVolume`

pykello:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/25076141678

PR #4896, opened 2026-03-05
Avoid apoptosis in Prog::Github::GithubRepositoryNexus#check_queued_jobs

jeremyevans:
The apoptosis timeout is 89 seconds, and each
client.workflow_run_attempt_jobs has a 5 second timeout. This raises an exception after 80 seconds to prevent apoptosis.

### DIFF
--- a/prog/github/github_repository_nexus.rb
+++ b/prog/github/github_repository_nexus.rb
@@ -25,11 +25,16 @@ class Prog::Github::GithubRepositoryNexus < Prog::Base
     @polling_interval ||= 5 * 60
   end
 
+  def clock_time
+    Process.clock_gettime(Process::CLOCK_MONOTONIC)
+  end
+
   def check_queued_jobs
     unless github_repository.installation.project.active?
       @polling_interval = 24 * 60 * 60
       return
     end
+    deadline = clock_time + 80
     client = github_repository.installation.client(auto_paginate: true)
     queued_runs = client.repository_workflow_runs(github_repository.name, {status: "queued"})[:workflow_runs]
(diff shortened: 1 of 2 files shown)


prog/github/github_repository_nexus.rb
```diff
@@ -25,11 +25,16 @@ def polling_interval
     @polling_interval ||= 5 * 60
   end
 
+  def clock_time
+    Process.clock_gettime(Process::CLOCK_MONOTONIC)
```
enescakir, line 29: Why don't we just use `Time.now`?

jeremyevans: While it is unlikely to make a significant difference in this particular case, in general, we should prefer `Process.clock_gettime(Process::CLOCK_MONOTONIC)` if the point is to measure the time difference between two points. `Time.now` does not provide a monotonic clock. For cases this large (80 seconds), the difference is unlikely to be material, but when measuring smaller time differences, it can be.

One reason to prefer ` Process.clock_gettime(Process::CLOCK_MONOTONIC)` is because it is over 3 times faster:

```ruby
require 'benchmark'
n = 10000000
puts Benchmark.measure{n.times{Time.now}}
  1.981193   0.001282   1.982475 (  1.982963)
puts Benchmark.measure{n.times{Process.clock_gettime(Process::CLOCK_MONOTONIC)}}
  0.683960   0.000000   0.683960 (  0.684027)
```

I'm not advocating for the removal of existing cases that are using `Time.now` to measure the time difference between two points, but I think we should avoid it for new code.

PR #4899, opened 2026-03-05
Make Vm#load_balancer_state handle missing VM ports

jeremyevans:
We had a couple production exceptions related to this:

```
NoMethodError - undefined method 'state' for nil
/app/model/vm.rb:64:in 'block in Vm#load_balancer_state'
```

Unless there is a way to avoid the situation and ensure that if a VM has a load balancer, that the VM ports must be available, we need to handle this case.

### DIFF
--- a/model/vm.rb
+++ b/model/vm.rb
@@ -61,9 +61,9 @@ class Vm < Sequel::Model
   def load_balancer_state
     return nil unless load_balancer
     if load_balancer.stack == "dual"
-      %w[ipv4 ipv6].map! { |stack| load_balancer_vm_ports.find { it.stack == stack }.state }
+      %w[ipv4 ipv6].map! { |stack| load_balancer_vm_ports.find { it.stack == stack }&.state }
     else
-      [load_balancer_vm_ports.first.state]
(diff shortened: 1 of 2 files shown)


mohi-kalantari:
There are 4 places in code which `LoadBalancerVmPort.create` is called and none of them provide the state but there is a default set in `20250324_add_loadbalancer_ports.rb` file. Updates to the records also provide state all the time.

I don't know how those nil states were set. Do you have any idea how this might have happened @jeremyevans ?

This is the migration line which sets nullable to false
```
column :state, :lb_node_state, null: false, default: "down"
```

jeremyevans:
@mohi-kalantari I think you are misunderstanding the problem. The problem isn't that there are `LoadBalancerVmPort` instances with `nil` states. The problem is there isn't a `LoadBalancerVmPort` instance when you expect there to be one. A `nil` state would not cause the issue, because the production exception is `NoMethodError - undefined method 'state' for nil`.

mohi-kalantari:
Got it thanks for the fix @jeremyevans

PR #4901, opened 2026-03-05
Don't fail VM systemd units on soft stop.

pykello:
When a VM shuts down from inside the guest or a `power-button` command is sent via `ch-remote`, cloud-hypervisor exits and removes the API socket before systemd runs `ExecStop`. The stop command, which tries to send the `shutdown-vmm` command via the API socket, then fails with "Error opening HTTP socket: No such file or directory", causing the service to be marked as failed even though the shutdown was successful.

From `man systemd.service`:

> Also note that the stop operation is always performed if the service
> started successfully, even if the processes in the service terminated
> on their own or were killed. The stop commands must be prepared to
> deal with that case.

This commit fixes the issue by removing `ExecStop`. In this case, if Cloud-Hypervisor hasn't already exited, `systemctl stop` will send `SIGTERM` (followed by `SIGKILL` after the stop timeout, defaulting to 90 seconds). Cloud-Hypervisor's `SIGTERM` handler already causes a call to `vmm_shutdown()`, which is the desired behaviour.

### DIFF
--- a/prog/test/vm.rb
+++ b/prog/test/vm.rb
@@ -124,10 +124,29 @@ class Prog::Test::Vm < Prog::Test::Base
 
   label def check_stopped_by_shutdown_command
     if vm.strand.label == "stopped" && !up?
-      hop_start_semaphore_after_shutdown
+      hop_verify_systemd_unit_status_after_shutdown
     end
 
     nap 5
+  end
+
+  label def verify_systemd_unit_status_after_shutdown
+    vm_state = begin
+      vm.vm_host.sshable.cmd("systemctl is-active :inhost_name", inhost_name: vm.inhost_name).strip
+    rescue Sshable::SshError => ex
+      ex.stdout.strip
+    end
+
+    case vm_state
+    when "active"
+      nap 5
+    when "inactive"
+      # nothing
+    else
+      fail_test "VM should be inactive after shutdown command, but is #{vm_state}"
+    end
+
+    hop_start_semaphore_after_shutdown
   end
 
   label def start_semaphore_after_shutdown
--- a/spec/prog/test/vm_spec.rb
+++ b/spec/prog/test/vm_spec.rb
@@ -363,7 +363,31 @@ RSpec.describe Prog::Test::Vm do
     it "hops if VM is down" do
       expect(vm_test.vm).to receive(:strand).and_return(instance_double(Strand, label: "stopped"))
       expect(vm_test.sshable).to receive(:_cmd).with("true").and_raise(Errno::ECONNREFUSED)
-      expect { vm_test.check_stopped_by_shutdown_command }.to hop("start_semaphore_after_shutdown")
+      expect { vm_test.check_stopped_by_shutdown_command }.to hop("verify_systemd_unit_status_after_shutdown")
+    end
+  end
+
+  describe "#verify_systemd_unit_status_after_shutdown" do
+    before {
+      sshable = Sshable.create
+      vm_host = instance_double(VmHost, sshable:)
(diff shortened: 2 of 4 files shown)


pykello:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22746068779

prog/test/vm.rb
```diff
@@ -124,12 +124,25 @@ def umount_if_mounted(mount_path)
...
 
+  label def verify_systemd_unit_status_after_shutdown
+    vm_state = begin
+      vm.vm_host.sshable.cmd("systemctl is-active :inhost_name", inhost_name: vm.inhost_name).strip
+    rescue Sshable::SshError => ex
+      ex.stdout.strip
+    end
+
+    nap 5 if vm_state == "active"
+    fail_test "VM should be inactive after shutdown command, but is #{vm_state}" unless vm_state == "inactive"
```
jeremyevans: Up to you, but I prefer the single `case` approach here as opposed to two independent conditionals.

```suggestion
    case vm_state
    when "active"
      nap 5
    when "inactive"
      # nothing
    else
      fail_test "VM should be inactive after shutdown command, but is #{vm_state}"
    end
```

spec/prog/test/vm_spec.rb
```diff
@@ -361,7 +361,31 @@
...
+
+    it "naps if VM is still active" do
+      expect(vm_test.vm.vm_host.sshable).to receive(:_cmd).with("systemctl is-active vm123456").and_return("active\n")
+      expect { vm_test.verify_systemd_unit_status_after_shutdown }.to nap(5)
+    end
+
+    it "fails if systemd unit status is unexpected" do
+      expect(vm_test.vm.vm_host.sshable).to receive(:_cmd).with("systemctl is-active vm123456").and_return("unknown\n")
+      expect(vm_test.strand).to receive(:update).with(exitval: {msg: "VM should be inactive after shutdown command, but is unknown"})
+      expect { vm_test.verify_systemd_unit_status_after_shutdown }.to hop("failed")
```
jeremyevans: Better to avoid mocking `update`, and instead check for the expected changes after the method:

```suggestion
      expect { vm_test.verify_systemd_unit_status_after_shutdown }.to hop("failed")
      expect(vm_test.strand.exitval).to eq({msg: "VM should be inactive after shutdown command, but is unknown"})
```

pykello:
One other option is to stop using `ExecStop` and rely on `systemctl stop`'s default behavior, which sends a `SIGTERM` if process hasn't already exited (followed by `SIGKILL` after a timeout specified by `TimeoutStopSec=`).

`SIGTERM` does the same thing as sending the shutdown-vmm api call if I'm reading the code correctly:

- `SIGTERM` handler writes to `exit_evt`: https://github.com/cloud-hypervisor/cloud-hypervisor/blob/01e4053bef19f7dc0166b177415bd70bfeb8b583/vmm/src/lib.rs#L739-L740
- Whenever a write happens to `exit_evt`, `vmm_shutdown()` is called: https://github.com/cloud-hypervisor/cloud-hypervisor/blob/01e4053bef19f7dc0166b177415bd70bfeb8b583/vmm/src/lib.rs#L1539-L1546

pykello:
Removed the review request from Daniel until I evaluate the `ExecStop=` removal option.

pykello:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23031252171

pykello:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23156178143

PR #4905, opened 2026-03-06
MigrateSpdkVmToUbiblk: VM stop/start improvements

pykello:
### MigrateSpdkVmToUbiblk: fix stop condition.

37e9c1d changed the behaviour of stop semaphore to first try a soft stop followed by hard stop after 60s.

`systemctl is-active $vm` returns:

- `inactive` if hard stop succeeds.
- `failed` if soft stop succeeds. This is because in this case, C-H   fails with `Error opening HTTP socket: No such file or directory (os error 2)`. This failure will be fixed in a separate commit for future VMs, but we should handle it when migrating old VMs.

This commit changes the condition in `wait_vm_stop` for waiting to stop to handle both cases.

### MigrateSpdkVmToUbiblk: use incr_start to start the VM. 

8390cb0 re-introduced the `start` semaphore which can be used to start a stopped VM. So, we can use it in `MigrateSpdkVmToUbiblk` to start the VM after migration is done.

### DIFF
--- a/prog/storage/migrate_spdk_vm_to_ubiblk.rb
+++ b/prog/storage/migrate_spdk_vm_to_ubiblk.rb
@@ -162,9 +162,7 @@ class Prog::Storage::MigrateSpdkVmToUbiblk < Prog::Base
   end
 
   label def start_vm
-    vm.vm_host.sshable.cmd("sudo systemctl start :inhost_name", inhost_name:)
-    vm.strand.update(label: "wait")
-
+    vm.incr_start
     pop "Vm successfully migrated to ubiblk"
   end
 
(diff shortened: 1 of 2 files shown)


mohi-kalantari (approved):

prog/storage/migrate_spdk_vm_to_ubiblk.rb
```diff
@@ -49,15 +49,15 @@ def vm_storage_volume
...
+      vm.vm_host.sshable.cmd("systemctl is-active :inhost_name", inhost_name:).strip
     rescue Sshable::SshError => ex
       # systemctl returns exit code 3 when a unit is inactive but writes
       # the result to the stdout. based on the way sshable.cmd work,
       # we need to capture the exception and extract the stdout, otherwise
       # on non-zero exit codes, it wouldn't return the stdout directly.
       ex.stdout.strip
     end
-    nap 5 unless vm_state == "inactive"
+    nap 5 unless ["inactive", "failed"].include?(vm_state)
```
jeremyevans: Up to you, but this is slightly simpler:

```suggestion
    nap 5 unless %w[inactive failed].include?(vm_state)
```

PR #4907, opened 2026-03-06
Destroy custom labels when destroying the installation

enescakir:
Since there is a foreign key constraint, we need to delete the custom labels before deleting the installation; otherwise, it fails with a foreign key constraint error.


enescakir:
> This is fine, but you may want to consider ON DELETE CASCADE on the foreign key as an alternative. I think that is overall simpler, but the tradeoff is that custom labels wouldn't be in `archived_record` after the installation is destroyed. Not sure whether that matters.

It's good to see them in the archived_record, let's keep that way for now

PR #4908, opened 2026-03-06
Check new account email domains against SURBL blacklist

enescakir:
Check new account email domains against SURBL blacklist
Abusers have been creating many accounts with custom domains that
are listed in the SURBL multi blacklist. To counter this, a new
background prog runs after account creation, querying the domain
against multi.surbl.org via DNS.

If the domain is listed, the account is suspended and the hit is
logged via Clog.emit with the matched lists (PH, MW, ABUSE, etc.).
If the query returns 127.0.0.1 (access blocked), it logs and
retries after one hour.

The system default DNS resolver does not recurse into the
multi.surbl.org zone, so a public recursive resolver (Google DNS)
is used instead. Per SURBL guidelines, queries should go through a
caching recursive resolver rather than directly to their
authoritative nameservers (a-j.surbl.org).

The prog is named CheckDomainBlacklist to allow adding other
blacklist databases in the future.

### DIFF
--- a/prog/check_domain_blacklist.rb
+++ b/prog/check_domain_blacklist.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+require "resolv"
+
+class Prog::CheckDomainBlacklist < Prog::Base
+  subject_is :account
+
+  SURBL_LISTS = {
+    "DM" => 4,
+    "PH" => 8,
+    "MW" => 16,
+    "CT" => 32,
+    "ABUSE" => 64,
+    "CR" => 128
+  }.freeze
+
+  # Google Public DNS is used because the system default resolver may not
+  # recurse into the multi.surbl.org zone. Per SURBL guidelines, queries
+  # should go through a caching recursive resolver, not directly to their
+  # authoritative nameservers.
+  SURBL_DNS_NAMESERVER = "8.8.8.8"
+
+  label def start
+    domain = account.email.split("@", 2)[1]
+
+    begin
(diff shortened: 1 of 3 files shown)


prog/check_domain_blacklist.rb
```diff
@@ -0,0 +1,56 @@
...
+  # Google Public DNS is used because the system default resolver may not
+  # recurse into the multi.surbl.org zone. Per SURBL guidelines, queries
+  # should go through a caching recursive resolver, not directly to their
+  # authoritative nameservers.
+  SURBL_DNS_NAMESERVER = "8.8.8.8"
+
+  label def start
+    domain = account.email.split("@", 2)[1]
+
+    if check_surbl(domain)
```
jeremyevans: This seems like a strange API design. The return value makes it seem to be a predicate method, but it can also nap.  Three possible approaches:

1. Inline the method (simplest as it is only called from a single place)

2. Move the `account.suspend` inside the method, and don't have the method return a boolean (have it always return nil).

3. Change the method to not nap, and instead return nil, and do the nap/`account.suspend` in the caller depending on the result.

I recommend 1.

enescakir: It’s probably designed that way so we can add additional database checks. However, I think if we add new database checks in the future, it would be better to perform each check in a separate label, so a failure in one doesn’t trigger retries for all of them.

So, I will change it to #1.

jeremyevans (approved):

PR #4913, opened 2026-03-06
Demock strand.update in test prog specs

pykello:
Replace `expect(*.strand).to receive(:update).with(exitval: ...)` mocks with real strand updates by using Strand.create instead of Strand.new. Assert on strand.reload.exitval after the hop instead.

### DIFF
--- a/spec/prog/test/firewall_rules_spec.rb
+++ b/spec/prog/test/firewall_rules_spec.rb
@@ -5,12 +5,14 @@ require "netaddr"
 
 RSpec.describe Prog::Test::FirewallRules do
   subject(:firewall_test) {
-    described_class.new(Strand.new(prog: "Test::FirewallRules"))
+    described_class.new(Strand.create(prog: "Test::FirewallRules", label: "start"))
   }
 
   let(:sshable) {
     Sshable.new
   }
+
+  let(:strand) { firewall_test.strand }
 
   let(:private_subnet_1) {
     nic = instance_double(Nic, private_ipv6: NetAddr::IPv6Net.parse("fd01:0db8:85a1::/64"), private_ipv4: NetAddr::IPv4Net.parse("192.168.0.1/32"))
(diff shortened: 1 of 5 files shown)


spec/prog/test/firewall_rules_spec.rb
```diff
@@ -5,7 +5,7 @@
 
 RSpec.describe Prog::Test::FirewallRules do
   subject(:firewall_test) {
-    described_class.new(Strand.new(prog: "Test::FirewallRules"))
+    described_class.new(Strand.create(prog: "Test::FirewallRules", label: "start"))
   }
```
enescakir, line 10: If you prefer shorter lines
```ruby
let(:strand) { firewall_test.strand }
```

spec/prog/test/hetzner_server_spec.rb
```diff
@@ -197,7 +198,7 @@
 
     it "fails if /vm is not empty" do
       expect(hs_test.vm_host.sshable).to receive(:_cmd).with("sudo ls -1 /vm").and_return("vm1\nvm2\n")
-      expect(hs_test.strand).to receive(:update).with(exitval: {msg: "VM directory not empty: [\"vm1\", \"vm2\"]"})
+      expect(strand).to receive(:update).with(exitval: {msg: "VM directory not empty: [\"vm1\", \"vm2\"]"})
```
jeremyevans: Not sure if this change and the following two changes in this file should be in this commit, because they do not demock. No objections to them, it's still good to DRY things up.

spec/prog/test/vm_group_spec.rb
```diff
@@ -140,7 +137,7 @@
 
     it "hops to verify_firewall_rules if tests are done" do
       expect(vg_test).to receive(:frame).and_return({"test_slices" => true})
-      expect(vg_test.strand).to receive(:retval).and_return({"msg" => "Verified VM Host Slices!"})
+      expect(strand).to receive(:retval).and_return({"msg" => "Verified VM Host Slices!"})
```
jeremyevans: I think this can be democked:

```suggestion
      strand.retval = {"msg" => "Verified VM Host Slices!"}
```

Similar change later in the file.

jeremyevans (approved):

Issue #4904, opened 2026-03-06
Replication slot deadlock after planned failover during scale-up


iskakaushik:
## Bug

After a planned failover during a Postgres scale-up, the newly promoted primary does not create replication slots for remaining standbys in `wait_catch_up`. This causes a deadlock where standbys can never catch up, and the scale-up cannot complete without manual intervention.

## Sequence of events

1. User triggers a scale-up (e.g., i8g.2xlarge → i8ge.3xlarge)
2. Convergence provisions 3 new standbys (i8ge.3xlarge). They restore from backup and enter `wait_catch_up`
3. Once enough new standbys are ready, convergence triggers a planned failover via `recycle_representative_server`
4. One of the new standbys is promoted to primary (`taking_over` label in `postgres_server_nexus.rb`)
5. In `taking_over` (line 889), all remaining standbys are reset to `synchronization_status: "catching_up"`
6. The new primary runs `configure` → `configure_hash` (`postgres_server.rb:99`) only creates physical replication slots for standbys where `synchronization_status == "ready"`
7. Since all remaining standbys are now `"catching_up"`, **no replication slots are created**
8. Old servers are pruned
9. Remaining standbys try to connect to the new primary: `ERROR: replication slot "<ubid>" specified by "primary_slot_name" does not exist on primary server`
10. `wait_catch_up` (line 606) naps until `lsn_caught_up` — but the standby can't stream WAL without a slot, so `lsn_caught_up` never becomes true
11. `wait_catch_up` only calls `representative_server.incr_configure` **after** `lsn_caught_up` succeeds (line 612), so the primary is never reconfigured to add the missing slots
12. **Deadlock**: standbys need slots to catch up, primary won't create slots until standbys are caught up

## Why this only happens during scale-up

In a normal failover without scale-up, the old standbys already had replication slots on the previous primary. Those slots persist after promotion (the configure script only drops slots not in the current list). During scale-up, the new standbys were still in `wait_catch_up` and never had slots created for them on any primary — so there are no pre-existing slots to inherit.

## Observed impact

Standbys are stuck permanently in `wait_catch_up`. The new primary has 0 streaming replicas. The scale-up cannot complete, and the resource runs without HA until manual intervention.

PR #4915, opened 2026-03-06
Fix replication slot deadlock during scale-up operations

iamKunalGupta:
Attempts to fix https://github.com/ubicloud/ubicloud/issues/4904

### DIFF
--- a/model/postgres/postgres_server.rb
+++ b/model/postgres/postgres_server.rb
@@ -85,7 +85,7 @@ class PostgresServer < Sequel::Model
       configs["allow_alter_system"] = "off"
     end
 
-    caught_up_standbys = nil
+    standbys_needing_slot = nil
     if timeline.blob_storage
       configs[:archive_mode] = "on"
       configs[:archive_timeout] = "60"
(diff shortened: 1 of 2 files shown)


model/postgres/postgres_server.rb
```diff
@@ -96,12 +96,13 @@ def configure_hash
       end
 
       if primary?
-        caught_up_standbys = resource.servers.select { it.standby? && it.synchronization_status == "ready" }
+        # We always need standby slots
+        standbys_needing_slot = resource.servers.select(&:standby?)
```
serprex, line 100: `caught_up_standbys` was still used for `synchronous_standby_names` before physical slots were used. it's important not to list catching up slots early on because:

1. synchronous standby names will cause commit to be blocked until one of the standbys is caught up
2. synchronized standby slots should wait until nearly caught up to avoid creating slots too early, which would cause WAL to build up until standby catches up

First problem isn't so much an issue with `ANY 1` at play. Only trouble is if none of the standbys have ready `synchronization_status`. In that case we should leave synchronous standby names empty

Since we track which standbys expect slots (`physical_slot_ready`) it seems we should _always_ include such standbys' slots to be on primary, as current design doesn't handle turning off `physical_slot_ready`

```suggestion
        caught_up_standbys = resource.servers.select { it.standby? && it.synchronization_status == "ready" }
        standbys_needing_slot = resource.servers.select { it.standby? && (it.synchronization_status == "ready" || it.physical_slot_ready) }
```

byucesoy: Yep, we shouldn't change what `caught_up_standbys` means. It is OK to add a new variable based on what you need.

serprex:
@byucesoy discussed with @iskakaushik, the issue as I outlined it made sense & revealed there may be some misconceptions in #4904

once PR updated we will test a sync standby scaleup

serprex:
Looked at case with @byucesoy of this in prod. Fixed by setting `physical_slot_ready` to false & reloading pg conf with no physical slot name

Better fix: change `physical_slot_ready` to primary server's id. This id represents which server slot exists on. When configuring check if representative server id is same as stored id to determine whether to use slot or not

PR #4917, opened 2026-03-06
Force use of a different env key for admin sessions

jeremyevans:
Without this, if the CloverAdmin and the Clover accessed the session while handling the same request, they could potentially see the session for which application first accessed the session. While I don't think we currently have code that will do that, it's best to avoid the possibility entirely by storing the CloverAdmin session in a different key in the rack env.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -43,6 +43,7 @@ class CloverAdmin < Roda
 
   plugin :sessions,
     key: "_CloverAdmin.session",
+    env_key: "clover.admin.session",
     cookie_options: {secure: !(Config.development? || Config.test?)},
     secret: OpenSSL::HMAC.digest("SHA512", Config.clover_session_secret, "admin-site")
 


enescakir:
It looks like there is a nondeterministic test.

```
  1) Prog::Kubernetes::ProvisionKubernetesNode#start assigns the default storage size if not specified
     Failure/Error: expect(new_vm.strand.stack.first["storage_volumes"].first["size_gib"]).to eq 80

       expected: 80
            got: 40

       (compared using ==)
     # ./spec/prog/kubernetes/provision_kubernetes_node_spec.rb:139:in 'block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:63:in 'block (3 levels) in <top (required)>'

```

enescakir:
I rerun tests

enescakir (approved):

jeremyevans:
Fixed one of those nondeterminism issues in #4879, but it looks like that wasn't the only one.

PR #4918, opened 2026-03-06
Allow closing of admin accounts

jeremyevans:
This allows admins to close their own accounts. It also allows admins to review the list of admin accounts, and to close other admin accounts (in case an offboarded admin forgets to close their own).

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -135,7 +135,7 @@ class CloverAdmin < Roda
   skip_webauthn_requirement = Config.development? && Config.clover_admin_development_no_webauthn?
 
   plugin :rodauth, route_csrf: true do
-    enable :argon2, :login, :logout, :webauthn, :change_password, :close_account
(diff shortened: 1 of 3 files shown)


clover_admin.rb
```diff
@@ -765,6 +766,20 @@ def self.object_action(...)
       view("admin_list")
     end
 
+    r.post "close-admin-account" do
+      login = typecast_params.nonempty_str!("login")
+
+      begin
+        CloverAdmin.rodauth.close_account(account_login: login, session: {closer: rodauth.account_from_session[:login]})
+      rescue Rodauth::InternalRequestError
+        flash["error"] = "Unable to close admin account for #{login.inspect}."
```
enescakir, line 775: Since it's a nonempty_str, why do we call inspect?

jeremyevans: `inspect` will surround the login with quotes, and if there are any weird characters, it escapes them (not that I would expect there to be).

PR #4930, opened 2026-03-09
Add backup_period_hours to adjust period between backups

serprex:
Useful for tuning specific users, currently we have a couple workloads
where multiple backups per day avoid hours of wal replay when scaling

In the past backups were weekly, as they'd take too long to run daily,
& workload may prefer to have wal-g throttled

Not exposed directly for now, meant for operator use


prog/postgres/postgres_timeline_nexus.rb
```diff
@@ -11,7 +11,7 @@ class Prog::Postgres::PostgresTimelineNexus < Prog::Base
   def_delegators :postgres_timeline, :blob_storage_client
 
   def self.assemble(location_id:, parent_id: nil)
-    if parent_id && PostgresTimeline[parent_id].nil?
+    unless !parent_id || (parent = PostgresTimeline[parent_id])
```
byucesoy: `unless` doesn't simplify the condition here. It removes `.nil?` but adds `!`. Also in general `unless` with multiple conditions are more difficult to read. I'd prefer if we just add back `.nil?` check and get rid of `unless`. I usually use `unless` if there is only one condition and it is already negated.

serprex: sure, was being consistent with next block

PR #4934, opened 2026-03-10
UMI: Add `bin/archive-storage-volume`

pykello:
### Refactor TOML helper functions into `toml.rb`
Move TOML helper functions from `storage_volume.rb` to `toml.rb` so they can be reused by other files.

### Refactor KEK pipe helper functions.
Move KEK pipe helper functions from `storage_volume.rb` to the `KekPipe` module so they can be reused by other files.

Also introduce `KekPipe.run_with_kek_pipe`, which runs a command that expects a KEK to be written to a pipe.

### UMI: Add `bin/archive-storage-volume`
This executable archives a VM storage volume to an S3-compatible object store.

It accepts its arguments as a JSON document from stdin. Example:

```
payload = {
  "vm_name": "vm123456",
  "device": "DEFAULT",
  "disk_index": 0,
  "vhost_block_backend_version": "v0.4.0",
  "kek": {"algorithm": "aes-256-gcm", "key": "b64-encoded-32-byte-key"},
  "target_conf": {
     "endpoint": "https://s3-endpoint-url",
     "region": "auto",
     "bucket": "ubicloud-machine-images",
     "prefix": "pj1234567890/mi9876543210",
     "access_key_id": "valid-access-key-id",
     "secret_access_key": "valid-secret-access-key",
     "session_token": "optional-session-token",
     "archive_kek": {
       "algorithm": "aes-256-gcm",
       "key": "b64-encoded-32-byte-key"
     }
  }
}.to_json

sshable.cmd("sudo host/bin/archive-storage-volume", stdin: payload)
```

The input is a storage volume on the host whose data encryption key is encrypted with the `kek` parameter and stored in the volume’s config file on the host.

The output is an archive that is compressed, encrypted, and uploaded to the object store. Its data encryption key is encrypted with `archive_kek` and stored in `metadata.json` in the object store.

### DIFF
--- a/rhizome/common/lib/util.rb
+++ b/rhizome/common/lib/util.rb
@@ -80,3 +80,16 @@ end
 def curl_file(url, path)
   r("bash -c 'curl -f -L3 #{url.shellescape} | tee >(openssl dgst -sha256) > #{path.shellescape}'").split(" ").last
 end
+
+def validate_keys(context, required_keys, optional_keys, hash)
+  all_keys = required_keys + optional_keys
+  missing_keys = required_keys - hash.keys
+  extra_keys = hash.keys - all_keys
+  unless missing_keys.empty?
+    raise ArgumentError, "Missing required keys in #{context}: #{missing_keys.join(", ")}"
+  end
+
+  unless extra_keys.empty?
+    raise ArgumentError, "Unexpected keys in #{context}: #{extra_keys.join(", ")}"
+  end
+end
--- a/rhizome/host/lib/storage_archive.rb
+++ b/rhizome/host/lib/storage_archive.rb
@@ -1,0 +1,96 @@
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require_relative "kek_pipe"
+require_relative "toml"
+require_relative "vhost_block_backend"
+
+class StorageArchive
+  include KekPipe
+  include Toml
+
+  def initialize(disk_config_path, disk_kek_path, disk_kek, target_conf, vhost_block_backend_version)
+    validate_keys(
+      "target_conf",
+      %w[bucket prefix region endpoint access_key_id secret_access_key archive_kek],
+      %w[session_token],
+      target_conf
+    )
+    @backend = VhostBlockBackend.new(vhost_block_backend_version)
+
+    fail "vhost block backend version #{vhost_block_backend_version} does not support archive" unless @backend.supports_archive?
+    StorageArchive.verify_key_encryption_key(disk_kek, "disk_kek")
+    StorageArchive.verify_key_encryption_key(target_conf["archive_kek"], "target_conf archive_kek")
+
+    @disk_config_path = disk_config_path
+    @disk_kek_path = disk_kek_path
+    @disk_kek = disk_kek
+    @target_conf = target_conf
+  end
+
+  def archive
+    cmd = [
+      @backend.archive_path,
+      "--config", @disk_config_path,
+      "--target-config", "/dev/stdin",
+      "--compression", "zstd",
+      "--zstd-level", "3"
+    ]
+    env = {"RUST_LOG" => "info"}
+    target_config = build_target_config
+    run_with_kek_pipe(
+      cmd,
+      kek_pipe: @disk_kek_path,
+      kek_content: @disk_kek["key"],
+      env: env,
+      stdin: target_config
+    )
+  end
+
+  def build_target_config
+    target = {
+      "storage" => "s3",
+      "bucket" => @target_conf["bucket"],
+      "prefix" => @target_conf["prefix"],
+      "region" => @target_conf["region"],
+      "endpoint" => @target_conf["endpoint"],
+      "access_key_id.ref" => "s3-key-id",
+      "secret_access_key.ref" => "s3-secret-key",
+      "archive_kek.ref" => "archive-kek"
+    }
+    target["session_token.ref"] = "s3-session-token" if @target_conf["session_token"]
+
+    parts = [toml_section("target", target)]
+
+    parts << toml_section("secrets.s3-key-id", {
+      "source.inline" => @target_conf["access_key_id"]
+    })
+    parts << toml_section("secrets.s3-secret-key", {
+      "source.inline" => @target_conf["secret_access_key"]
+    })
+    parts << toml_section("secrets.archive-kek", {
+      "source.inline" => @target_conf["archive_kek"]["key"],
+      "encoding" => "base64"
+    })
+
+    if @target_conf["session_token"]
+      parts << toml_section("secrets.s3-session-token", {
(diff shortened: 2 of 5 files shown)


pykello:
/run-e2e metal



E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22916451415

rhizome/host/lib/storage_key_encryption.rb
```diff
@@ -85,3 +89,47 @@ def unwrap_key(encrypted_key)
...
+    rescue => e
+      Process.kill("TERM", pid)
+      Process.waitpid(pid)
+      raise "error writing KEK to pipe: #{e.message}"
+    end
+
+    _, status = Process.wait2(pid)
+
+    fail CommandFail.new("command failed: #{command.join(" ")}", "", "") unless status.success?
+  end
```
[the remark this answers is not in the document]
pykello: not essential for the current change. might do this later if needed.

rhizome/host/lib/storage_key_encryption.rb
```diff
@@ -3,7 +3,11 @@
 require_relative "../../common/lib/util"
 require "openssl"
 require "base64"
+require "fileutils"
 require "securerandom"
+require "timeout"
+
+KEK_PIPE_WRITE_TIMEOUT_SEC = 5
```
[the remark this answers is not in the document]
pykello: not related to the current change

rhizome/host/lib/toml.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+def toml_section(name, hash)
```
jeremyevans: I would add these methods to a module, and include that module in the code that needs these methods. When you add them to the top level, they become private methods for all objects.

rhizome/host/lib/storage_key_encryption.rb
```diff
@@ -85,3 +89,60 @@ def unwrap_key(encrypted_key)
     decipher.update(encrypted_key[0]) + decipher.final
   end
 end
+
+def with_kek_pipe(kek_pipe, owner: nil)
```
jeremyevans: Similar issue here, put methods in module, include module in code that needs it.

rhizome/common/lib/util.rb
```diff
@@ -80,3 +80,16 @@ def safe_write_to_file(filename, content = nil)
 def curl_file(url, path)
   r("bash -c 'curl -f -L3 #{url.shellescape} | tee >(openssl dgst -sha256) > #{path.shellescape}'").split(" ").last
 end
+
+def validate_keys(context, required_keys, optional_keys, hash)
```
jeremyevans, line 84: Ugh, looks like we use this pattern (top-level methods) already in rhizome. We should move away from that eventually, but for now, these are fine.

pykello: I moved the toml & kek pipe functions to modules, but kept this as it is so it is addressed with rest of `util.rb`.

rhizome/host/lib/kek_pipe.rb
```diff
@@ -0,0 +1,75 @@
...
+
+require_relative "../../common/lib/util"
+
+require "fileutils"
+require "timeout"
+
+module KekPipe
+  WRITE_TIMEOUT_SEC = 5
+
+  def self.with_kek_pipe(kek_pipe, owner: nil)
```
jeremyevans: This singleton method approach works, but my recommendation was to include the modules in the classes that use the methods (keeping the methods as instance methods). That way you could keep using `with_kek_pipe` instead of `KekPipe.with_kek_pipe`. If you choose this approach deliberately, I don't object to it, but I do think the normal module approach is simpler.

rhizome/host/lib/storage_archive.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require_relative "../../common/lib/util"
+require_relative "kek_pipe"
+require_relative "toml"
+require_relative "vhost_block_backend"
+
+class StorageArchive
```
jeremyevans, line 8: If using the normal module approach:

```suggestion
class StorageArchive
  include KekPipe
  include Toml
  
```

pykello:
/run-e2e metal



E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/22965961533

rhizome/host/lib/toml.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Toml
```
fdr, line 3: why not pick up a toml gem? this won't be the last rust dependency i suspect. If there is a decent one that can serialize.

fdr: although, hmm, in rhizome. I can see that being a bit unlikable.

fdr: oof the recursive dependencies, even of the pure ruby version...nevermind. I guess we'll do our own toml serialization for now.

pykello:
The E2E failed at `Test::Kubernetes.verify_data_after_drain`, which is not relevant to this change & we have seen few instances of that failure in the main branch in last 2 days. The relevant parts (vm provisioning) passed.

rhizome/common/lib/util.rb
fdr: util mostly exists to make "r" short, but I guess it got "some other stuff" that is not used nearly as often.

rhizome/host/lib/kek_pipe.rb
```diff
@@ -0,0 +1,72 @@
...
+module KekPipe
+  WRITE_TIMEOUT_SEC = 5
+
+  def with_kek_pipe(kek_pipe, owner: nil)
+    rm_if_exists(kek_pipe)
+    File.mkfifo(kek_pipe, 0o600)
+    FileUtils.chown owner, owner, kek_pipe if owner
+    yield kek_pipe
+  ensure
+    FileUtils.rm_f(kek_pipe)
```
fdr: why not `rm_if_exists?` if you went to the trouble of the carefully downscoped "ENOENT only" mode, why ignore, say, EACCESS? that would be a problem, leaving a file around, we'd probably prefer a crash than continuing.

rhizome/host/lib/kek_pipe.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def write_kek_to_pipe(kek_pipe, payload, timeout_sec: WRITE_TIMEOUT_SEC)
+    Timeout.timeout(timeout_sec) do
+      File.open(kek_pipe, File::WRONLY) do |file|
+        file.write(payload)
+      end
```
fdr:
```suggestion
      File.write(kek_pipe, payload, mode: File::WRONLY)
```
?

PR #4935, opened 2026-03-10
Add inference endpoint list API with pricing info

aamederen:
Add API endpoint to list inference endpoints and inference router models with serialized response including filtered tags and per-million-token pricing from billing rates. Update OpenAPI spec with new schemas.

### DIFF
--- a/cli-commands/ai/endpoint/list.rb
+++ b/cli-commands/ai/endpoint/list.rb
@@ -4,6 +4,7 @@ UbiCli.on("ai", "endpoint", "list") do
   desc "List inference endpoints"
 
   fields = %w[name display-name hf-model capability multimodal context-length url input-price output-price].freeze
+  default_fields = %w[name capability multimodal context-length input-price output-price].freeze
 
   key = :endpoint_list
 
@@ -30,7 +31,8 @@ UbiCli.on("ai", "endpoint", "list") do
         output_price: price[:per_million_completion_tokens]
       }
     end
-    keys = underscore_keys(check_fields(opts[:fields], fields, "ai endpoint list -f option", command))
+    given = opts[:fields]
+    keys = underscore_keys(given ? check_fields(given, fields, "ai endpoint list -f option", command) : default_fields)
     response(format_rows(keys, items, headers: opts[:"no-headers"] != false))
   end
 end
--- a/spec/routes/api/cli/ai/endpoint/list_spec.rb
+++ b/spec/routes/api/cli/ai/endpoint/list_spec.rb
@@ -29,8 +29,7 @@ RSpec.describe Clover, "cli ai endpoint list" do
   it "shows list of inference endpoints" do
     ie
 
-    url = ie.load_balancer.health_check_url
-    expect(cli(%w[ai endpoint list -N])).to eq "test-model  Test Model  test-org/test-model  Text Generation      #{url}  0.05  0.05\n"
+    expect(cli(%w[ai endpoint list -N])).to eq "test-model  Text Generation      0.05  0.05\n"
   end
 
   it "-f name option shows only name" do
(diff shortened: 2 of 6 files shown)


openapi/openapi.yml
```diff
@@ -2695,6 +2708,53 @@ components:
       required:
         - id
         - key
+    InferenceEndpointPrice:
+      type: object
+      properties:
+        prompt_per_million_tokens:
```
jeremyevans: Naming for these is strange `per_million_prompt_tokens` and `per_million_completion_tokens` would be simpler to understand.

spec/routes/api/project/inference_endpoint_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+      irt
+
+      get "/project/#{project.ubid}/inference-endpoint"
+
+      expect(last_response.status).to eq(200)
+      body = JSON.parse(last_response.body)
+
+      expect(body["items"].length).to eq(2)
+      expect(body["items"]).to contain_exactly(
+        hash_including(
```
jeremyevans: I would specify an order when retrieving the objects, and use `to eq [{...}, ...]`. Only use `contain_exactly` if you cannot control the order, and only use `hash_including` if you cannot control other contents in the hash.

spec/routes/web/inference_endpoint_spec.rb
```diff
@@ -78,12 +78,12 @@
...
-        .and_return({"unit_price" => 0.0000001})
-      expect(BillingRate).to receive(:from_resource_properties)
-        .with("InferenceTokens", "llama-3-2-1b-it-output", "global")
-        .and_return({"unit_price" => 0.0000002})
+      expect(BillingRate).to receive(:million_token_price)
+        .with("llama-3-2-1b-it-input")
+        .and_return(0.10)
+      expect(BillingRate).to receive(:million_token_price)
+        .with("llama-3-2-1b-it-output")
+        .and_return(0.20)
```
jeremyevans: Is there a need to mock these? Can we use an actual billing rate so that mocking isn't required?

aamederen:
Thanks @jeremyevans. I forgot about the CLI and since the feature request was from a single customer, I didn't think of it, but I think for consistency we can add it.

sdk/ruby/lib/ubicloud/model/inference_endpoint.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Ubicloud
+  class InferenceEndpoint < Model
+    set_prefix "ie"
+
+    set_fragment "inference-endpoint"
+
+    set_columns :id, :name, :display_name, :url, :model_name, :tags, :price
```
jeremyevans, line 9: This appears to be missing some columns from the API response. In general, all columns in the API response should be included here.

aamederen: AFAICS, this list matches the API response (serializer):

```ruby
{
      id: ie.ubid,
      name: ie.model_name,
      display_name: ie.tags["display_name"] || ie.model_name,
      url: ie.load_balancer.health_check_url,
      model_name: ie.model_name,
      tags: ie.tags.slice("hf_model", "capability", "multimodal", "context_length"),
      price: {
        per_million_prompt_tokens: BillingRate.million_token_price(ie.prompt_billing_resource),
        per_million_completion_tokens: BillingRate.million_token_price(ie.completion_billing_resource)
      }
    }
```

However, the CLI output is a simplified version of this (removing `id` while flattening `price` and `tags`):

```
name                         display-name                 hf-model                     capability       multimodal  context-length  url                  input-price  output-price
test-model                   Test Model                   test-org/test-model          Text Generation                              http://ie-lb.dpqe2.  0.05         0.05        
Qwen/Qwen3-VL-235B-Instruct  Qwen/Qwen3-VL-235B-Instruct  Qwen/Qwen3-VL-235B-Instruct  Text Generation  true        64k             http://ie-lb.dpqe2.  0.5          2.5         
```

What do you think?

jeremyevans: You are correct that you want it to match what the API gives you. Sorry, not sure why I thought it didn't match.

Considering the width of the table, are you sure using a table is best for this? If your terminal is <~180 columns, this isn't going to look good due to line wrapping. Actually, considering your example doesn't display full URLs, you would need even more columns. It would probably be better to use an approach like:

```
name: test-model
display-name: Test Model
...
output-price 0.05

name: Qwen/Qwen3-VL-235B-Instruct
...
```

Or maybe limit the default columns to `name capability multimodal context-length input-price output-price`.

aamederen: makes sense, thanks!

PR #4937, opened 2026-03-10
Add direct GitHub page links for installations and repositories

enescakir:
I’ve found myself repeatedly copying the installation name, typing github.com in the address bar, and pasting it there.

Having a direct link to the GitHub page would make this much more convenient.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -196,6 +196,10 @@ class CloverAdmin < Roda
     ObjectAction.define(...)
   end
 
+  github_page_action = object_action("GitHub Page", type: :direct) do |obj|
+    "http://github.com/#{obj.name}"
+  end
+
   OBJECT_ACTIONS = {
     "Account" => {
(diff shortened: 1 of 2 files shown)


clover_admin.rb
```diff
@@ -205,10 +205,18 @@ def self.object_action(...)
...
+        "http://github.com/#{obj.name}"
+      end
+    },
     "GithubRunner" => {
       "provision" => object_action("Provision Spare Runner", flash: "Spare runner provisioned", type: :form, &:provision_spare_runner)
     },
     "GithubRepository" => {
+      "github_page" => object_action("GitHub Page", type: :direct) do |obj|
+        "http://github.com/#{obj.name}"
+      end,
```
jeremyevans: Up to you, but consider setting the result of the `object_action` method to a local variable, and using that as the hash value for both GithubInstallation and GithubRepository, since it is the same in both cases.

PR #4942, opened 2026-03-11
Fix physical slot with new primary

### LINKED ISSUE #4904: Replication slot deadlock after planned failover during scale-up

serprex:
Scaling up a 2 standby setup can get into a scenario where physical slot was created on old primary when standby reused, causing connection to new primary to fail due to slot missing

Fix this by tracking which server standby was created on, when primary doesn't match the slot needs to be recreated & id updated

alternative to #4915
fixes #4904


migrate/20260330_pg_physical_slot_ready_id.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:postgres_server) do
+      add_column :physical_slot_ready_id, :uuid
+    end
+  end
```
byucesoy, line 23: We should also backfill the physical_slot_ready_id, so that we can safely remove physical_slot_ready column (and all the || conditionals that includes it)

serprex: wasn't sure if that migration should happen here or as a followup step, I think there's some potential race, but unlikely, will do

serprex:
I tested this pretty thoroughly: adjusted `main` to force race _(physical slots created for all standbys before failover)_, ported adjustments to this branch, & verified failover happened without errors from previous test

prog/postgres/postgres_server_nexus.rb
```diff
@@ -401,7 +401,7 @@ def before_run
       hop_wait
     when "Failed", "NotStarted"
       if postgres_server.use_physical_slot_set?
-        postgres_server.update(physical_slot_ready: true)
+        postgres_server.update(physical_slot_ready_id: postgres_server.resource.representative_server.id)
```
byucesoy, line 442: This didn't got introduced in this PR, but why do we set `physical_slot_ready` or `physical_slot_ready_id` when configure is failed?

serprex: when primary creates slot it increments use_physical_slot on standby. this update is standby registering that signal for configure_hash to use slot that exists now
would seem tho that there's a potential race here if after creating slot primary does failover, & in meantime new primary came up when standby finally gets to configure

not sure how best to address this, primary shouldn't update standby, but ideal would be for `incr_use_physical_slot` a few lines above being able to take primary's id as parameter

byucesoy: If primary does failover, it would go through configure again and call `incr_use_physical_slot` on standbys with non-matching `physical_slot_ready_id`, so that shouldn't cause any problem? I didn't think about this deeply though...

migrate/20260313_pg_physical_slot_ready_id.rb
```diff
@@ -0,0 +1,29 @@
...
+      SET physical_slot_ready_id = rep.id
+      FROM postgres_server AS rep
+      WHERE ps.physical_slot_ready
+        AND rep.resource_id = ps.resource_id
+        AND rep.is_representative
+    SQL
+
+    alter_table(:postgres_server) do
+      drop_column :physical_slot_ready
+    end
```
furkansahin: Can you please isolate this into a 3rd commit (the very last one). This way, we would have a zero downtime deployment.

PR #4944, opened 2026-03-11
Admin panel misc

enescakir:
- **Remove custom strand logic from the object.erb admin panel**

  Its functionality can be replaced with object actions. There is also no
need to show a separate “Schedule Strand to Run Now” button for strands,
since the “Schedule Strand to Run Immediately” object action already
provides the same functionality.

| Before | After |
|--------|-------|
| CleanShot 2026-03-11 at 14 19 26@2x | CleanShot 2026-03-11 at 14 18 39@2x |

- **Add a fixed footer bar to the admin panel**
  The admin panel is used by engineers across multiple time zones,
  and all timestamps displayed are in UTC. A persistent footer
  showing the current UTC time makes it easier to correlate
  timestamps at a glance, especially for US-based engineers where
  the offset from local time is significant.
  
  The footer also links to the deployed commit for quick reference.
  
CleanShot 2026-03-11 at 14 11 29@2x

- **Show subtotal and cost in admin panel project current usage**
  The admin panel's current usage section lists billing items
  individually but lacks a total, making it hard to see the overall
  spend at a glance. Add subtotal and cost to the collapsible table
  title so they are visible without scanning every line item.
  
CleanShot 2026-03-11 at 14 18 04@2x

- **Add GithubRepository table to admin panel**

- **Add more association table links to admin panel**
Navigating from an object to its associated records in tabular
form is useful for investigation. Add (table) links for
GithubInstallation repositories, GithubRepository runners, and
Project invoices. Also add a repository search filter to
GithubRunner to support the new association table link.

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -317,8 +317,11 @@ class CloverAdmin < Roda
 
   OBJECT_ASSOC_TABLE_PARAMS = {
     ["GithubInstallation", :runners] => "installation",
+    ["GithubInstallation", :repositories] => "installation",
+    ["GithubRepository", :runners] => "repository",
     ["Project", :vms] => "project",
     ["Project", :postgres_resources] => "project",
+    ["Project", :invoices] => "project",
     ["PostgresResource", :servers] => "resource"
   }.freeze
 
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
Timestamp in footer picture was almost perfect!

PR #4947, opened 2026-03-11
Ignore extra Gemfile.lock files during InstallRhizome

serprex:
I'm consistently ignoring "Prog::InstallRhizome#start writes tar" failures on account of ruby-lsp writing Gemfile.lock files which cause an assertion failure in that test, there's also `rhizome/host/Gemfile.lock` which is ignored by .gitignore but causes failure

### DIFF
--- a/prog/install_rhizome.rb
+++ b/prog/install_rhizome.rb
@@ -17,7 +17,7 @@ class Prog::InstallRhizome < Prog::Base
       base = Config.root + "/rhizome"
       Dir.glob(["Gemfile", "Gemfile.lock", "common/**/*", "#{frame["target_folder"]}/**/*"], base:) do |file|
         #:nocov:
-        next if file != "Gemfile.lock" && file.end_with?("Gemfile.lock")
+        next if file == "common/Gemfile.lock" || file == "host/Gemfile.lock"
         #:nocov:
         next if !frame["install_specs"] && file.end_with?("_spec.rb")
 


prog/install_rhizome.rb
```diff
@@ -16,6 +16,9 @@ class Prog::InstallRhizome < Prog::Base
     Gem::Package::TarWriter.new(tar) do |writer|
       base = Config.root + "/rhizome"
       Dir.glob(["Gemfile", "Gemfile.lock", "common/**/*", "#{frame["target_folder"]}/**/*"], base:) do |file|
+        #:nocov:
+        next if file != "Gemfile.lock" && file.end_with?("Gemfile.lock")
```
jeremyevans: I'd rather be specific here, assuming this will work for you:

```suggestion
        next if file == "common/Gemfile.lock" || file == "host/Gemfile.lock"
```

byucesoy:
> I think the root of the problem is we have `host/Gemfile` and `common/Gemfile` that are not designed to be used directly, only via inclusion from `Gemfile`.
> 
> I think we could avoid this issue by renaming `host/Gemfile` and `common/Gemfile` to `host/.gemfile.inc` and `common/.gemfile.inc`, and changing `Gemfile` to do:
> 
> ```ruby
> Dir.glob("*/.gemfile.inc", base: __dir__) do |file|
>   instance_eval File.read(File.join(__dir__, file))
> end
> ```
> 
> @byucesoy What are your thoughts on this approach?

Both approaches looks OK to me. This is falls more into Ruby conventions area. I assume what you suggest is more Ruby way of doing this, so I'm good with doing that.

jeremyevans:
I recommend a different approach to fix this issue: #5030

PR #4948, opened 2026-03-11
Add web/API/CLI interface to the audit log

jeremyevans:
Features:

* Shows logs in reverse chronological order
* Search by action, subject (account email/name/id), and object (id)
* Search by end date, view 3 months at a time (to limit the searched partitions)
* Supports paginated output using an index-friendly approach (not OFFSET)

CLI output is in CSV format, so it can be easily redirected to a file and
opened in a spreadsheet program:

```
$ ubi al search
At,Action,Account,Objects
2026-03-02T15:21:59Z,vm/destroy,Jeremy Evans,vmpkben3fd2e3ng59kk6b6644n 
2026-03-02T15:05:51Z,vm/create,Jeremy Evans,vmpkben3fd2e3ng59kk6b6644n 
...
```

Web interface:

* Allows clicking on action, subject, or object to do a search for that
  value.
* For objects still in the project, adds a View link to take you to the
  page for that object.

localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk_audit-log
localhost_3000_project_pjkvsq7s5e5a7mjtsdjvetweyk

Currently, this allows viewing logs up to 6 months in the past, since
the oldest allowed end date is 3 months ago, and it will show 3 months
before that. We may allow access to older audit logs in the future.

To handle multiple audit records with the same timestamp (created in
same transaction), the pagination key depends both the on the audit
record timestamp and on the id of the next record. We use an approach
where we select one record over the actual limit, and pop if off if
necessary to determine whether there are additional pages.

Claude was used to generate the initial implementation, but there are
numerous changes to add more features and fix issues.

This adds a migration for a Project:auditlog action type, which is added
to the Project:all action tag.

### DIFF
--- a/spec/routes/web/project_spec.rb
+++ b/spec/routes/web/project_spec.rb
@@ -934,7 +934,7 @@ RSpec.describe Clover, "project" do
         insert_audit_log
 
         visit project.path
-        click_link "View/Search Audit Logs"
+        click_link "View Audit Logs"
 
         expect(page.title).to eq("Ubicloud - project-1 - Audit Log")
         expect(audit_log_content).to eq ["vm/create", user.ubid, ""]
--- a/views/project/audit_log.erb
+++ b/views/project/audit_log.erb
@@ -61,7 +61,7 @@
     @audit_logs.map do |entry|
       [
         [
-          [entry[:at].getutc.strftime("%F %T"), {}],
+          entry[:at].getutc.strftime("%F %T"),
           ["<a class=\"text-orange-600\" href=\"?end=#{end_date}&amp;action=#{h entry[:ubid_type]}/#{h entry[:action]}\">#{h entry[:ubid_type]}/#{h entry[:action]}</a>", {escape: false}],
           [entry[:subject], {escape: false}],
           [entry[:objects].join("<br>"), {escape: false}]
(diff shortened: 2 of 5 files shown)


jeremyevans:
I've included an admin audit log in this PR. Here's what it looks like:

admin localhost_3000_audit-log_object=skjydf8ahyj60prdd6sfpgs0c4

It adds a "View Audit Log" link on every object page, so you have easily see any audit logs related to the object. Features are mostly the same as the project audit log page, other than it also includes the ability to search by project.

Almost all of the logic for searching the audit logs is shared between the admin audit log and project audit logs, by putting the logic in a module included in both Clover and CloverAdmin.

views/project/audit_log.erb
```diff
@@ -0,0 +1,63 @@
...
+    empty_state: {
+      icon: "hero-document-text",
+      title: "No audit log entries",
+      description: "No audit log entries found for this project."
+    },
+    rows:
+      @audit_logs.map do |entry|
+        [
+          [
+            [entry[:id], {}],
```
enescakir: If we don't pass any property, passing the value is enough.
```
entry[:id]
```

spec/routes/web/project_spec.rb
```diff
@@ -929,7 +929,8 @@ def insert_audit_log(project_id:, subject_id:, object_ids: [], action: "create",
         id = insert_audit_log(project_id: project.id, subject_id: user.id)
         entry_ubid = UBID.from_uuidish(id).to_s
 
-        visit "#{project.path}/audit-log"
+        visit project.path
+        click_link "View/Search Audit Logs"
```
enescakir: IMHO, no need to say "Search". 
"View Audit Logs"

PR #4951, opened 2026-03-12
Increase the AWS nic nexus start deadline to 5 minutes

enescakir:
The next label already has a 3-minute deadline, so it doesn’t make sense to add another 3-minute deadline at the start to the wait, since that time would also include the next label.

Additionally, we’ve experienced several short-lived pages in production due to rate limits and similar issues.


furkansahin (approved):
somehow I don't believe this will fix the short lived incidents. Whatever number we increase this to, we kept getting those small bursts. Let's still see if it helps

PR #4952, opened 2026-03-12
Handle InvalidIPAddressInUse in AWS NIC creation for idempotency

byucesoy:
It is possible that the API request succeeds but for one reason or another, we couldn't persist the change in our metadata database. In that case, if we retry the same request, we will get InvalidIPAddressInUse error from AWS because the NIC was actually created with the requested private IP address. This change adds handling for this error to make the operation idempotent.

### DIFF
--- a/prog/vnet/aws/nic_nexus.rb
+++ b/prog/vnet/aws/nic_nexus.rb
@@ -37,17 +37,29 @@ class Prog::Vnet::Aws::NicNexus < Prog::Base
   end
 
   label def create_network_interface
-    network_interface_response = client.create_network_interface({
-      subnet_id: nic.nic_aws_resource.subnet_id,
-      private_ip_address: nic.private_ipv4.network.to_s,
-      ipv_6_prefix_count: 1,
-      groups: [
-        private_subnet.private_subnet_aws_resource.security_group_id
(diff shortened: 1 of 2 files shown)


prog/vnet/aws/nic_nexus.rb
```diff
@@ -37,17 +37,29 @@ class Prog::Vnet::Aws::NicNexus < Prog::Base
...
+        ipv_6_prefix_count: 1,
+        groups: [
+          private_subnet.private_subnet_aws_resource.security_group_id
+        ],
+        tag_specifications: Util.aws_tag_specifications("network-interface", nic.name),
+        client_token: nic.id
+      })
+      network_interface_id = network_interface_response.network_interface.network_interface_id
+    rescue Aws::EC2::Errors::InvalidIPAddressInUse
+      network_interfaces = client.describe_network_interfaces({
```
enescakir, line 53: I assumed you tested that this query actually works. Previously I've seen LLMs hallucinate AWS requests and use non-existent params.

byucesoy: Yep. This is how I fixed the page that uncovered this issue.

PR #4953, opened 2026-03-12
Wait for CSI volume detach before checking pending PVs

mohi-kalantari:
After kubectl drain succeeds, the kubelet may still be running the CSI node_unstage_volume call which sets up the data migration annotations on PVs. The wait_for_copy step was checking pending_pvs immediately after drain, finding none, and proceeding to destroy the node before rsync could complete.

Add a wait_for_detach step between drain and wait_for_copy that polls VolumeAttachment objects filtered by the ubicsi CSI driver. This ensures the kubelet has fully completed the volume teardown lifecycle before the node removal proceeds.


mohi-kalantari:
/run-e2e metal kubernetes


E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23011659209

PR #4957, opened 2026-03-12
Add logs to investigate postgres upgrade E2E

shikharbhardwaj:


.github/workflows/e2e.yml
```diff
@@ -297,6 +297,10 @@ jobs:
                   short: false
                   value: "${{ steps.extract_last.outputs.LAST_LINE }}"
 
+    - name: Sleep 10 minutes after failed notification
```
enescakir, line 300: We run test commands with 83m timeout
https://github.com/ubicloud/ubicloud/blob/b6358c2f72429afa8e2b2153c133d812e05b460f/.github/workflows/e2e.yml#L163

And the workflow has 85m timeout
https://github.com/ubicloud/ubicloud/blob/b6358c2f72429afa8e2b2153c133d812e05b460f/.github/workflows/e2e.yml#L50

Even if we sleep 10m here, it will timeout after 2m. So we should increase the workflow timeout to at least 93m.

shikharbhardwaj:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23043361552

.github/workflows/e2e.yml
shikharbhardwaj: Done

PR #4959, opened 2026-03-12
UMI: Data model & a prog to create a version

pykello:
### UMI: Migrations for MachineImage and MachineImageVersion. 

Introduce tables for machine images and their versions.

A machine image can have zero or more versions. Each version stores its data in an S3-compatible object store, identified by `s3_endpoint`, `s3_bucket`, and `s3_prefix`.

The `machine_image.latest_version_id` foreign key points to the most recent version and is used to resolve the `machine-image@latest` specifier.

A VM storage volume can be based on a machine image version via the `vm_storage_volume.machine_image_version_id` foreign key.

### Add CloudflareClient::generate_temp_credentials 

This can be used to generate R2 credentials which expire after a given time period. This is useful when we want to reduce risk by avoiding sending long-term credentials to VM hosts.

### UMI: Add Prog::MachineImage::CreateVersion 

This prog generates a new machine image version from a stopped VM.

Given a stopped VM `vm`, the following example creates a MachineImage record and then creates a version from the VM's disk:

```
mi = MachineImage.create(
    name: "mi01",
    arch: "x64",
    project_id: vm.project.id,
    location_id: vm.location_id)

st = Prog::MachineImage::CreateVersion.assemble(
    mi, "v0.2.0", vm, destroy_source_after: false)
```

If `destroy_source_after` is true, the source VM is scheduled for destruction after the image version is created.

The machine image version is uploaded to `Config.machine_image_r2_bucket` under the "$project_ubid/$machine_image_ubid/$version" prefix. The bucket is expected to exist in both US and EU Cloudflare locations.

R2 connection can be configured using:

- `Config.machine_image_r2_account_id`
- `Config.machine_image_r2_api_token`
- `Config.machine_image_r2_access_key`
- `Config.machine_image_r2_secret_key`

### Add admin model specs for MachineImage and MachineImageVersion


fdr:
the bad-ish news comment from our meeting: we need this to respect the metal vs aws vs gcp split. In practice we always want versioning (that's why umis are designed the way they are) even when superimposed on image references on the hyperscalers we also support.

In practice it's not _so_ bad, only inconvenient: the umi-only fields (like s3 prefix) needs to be decomposed from the others and joined in "metal" join table. I do have claude churning away proving it all works in our ubi app prototype environment.

PR #4962, opened 2026-03-12
Add prefix-based attribute search to admin search bar

enescakir:
The admin search bar only accepted UBIDs and UUIDs, so finding a
resource by name or email required navigating to the model listing
page and using its search form. This was tedious for common lookups.

The search bar now supports a prefix:term syntax that searches a
single model's attributes. Prefixes use the two-character UBID type
codes (e.g. vm:my-server, ac:user@email.com, pg:my-database). When
the input is not a UBID or UUID, it redirects to /search which
parses the prefix, queries the appropriate table, and either
redirects on a single match or displays a results page.

Also switches the CSS alert selectors from IDs to classes so the
.alert-danger class applies to both flash messages and search
warnings.

CleanShot 2026-03-13 at 16 10 28@2x

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -324,6 +324,8 @@ class CloverAdmin < Roda
     "PostgresResource" => [:name],
     "Vm" => [:name]
   }.freeze
+  SEARCH_QUERIES.each_value(&:freeze)
+  SEARCH_PREFIXES = SEARCH_QUERIES.map { "#{Object.const_get(it[0]).ubid_type} (#{it[0]})" }.join(", ").freeze
 
   OBJECTS_WITH_UI = {
     "Vm" => lambda { |vm| "project/#{vm.project.ubid}/location/#{vm.location.display_name}/vm/#{vm.ubid}/overview" },
@@ -850,21 +852,21 @@ class CloverAdmin < Roda
 
       terms = term&.split(",")&.map(&:strip)&.reject(&:empty?)
       if terms.nil? || terms.empty?
-        flash.now["error"] = "Use prefix:term syntax to search (e.g. vm:name). Available prefixes: #{SEARCH_QUERIES.map { "#{Object.const_get(it[0]).ubid_type} (#{it[0]})" }.join(", ")}"
+        flash.now["error"] = "Use prefix:term syntax to search (e.g. vm:name). Available prefixes: #{SEARCH_PREFIXES}"
         next view("search")
       end
 
       klass = UBID.class_for_ubid(prefix)
       columns = klass && SEARCH_QUERIES[klass.name]
       unless columns
-        flash.now["error"] = "Unknown prefix: #{h(prefix)}. Available prefixes: #{SEARCH_QUERIES.map { "#{Object.const_get(it[0]).ubid_type} (#{it[0]})" }.join(", ")}"
+        flash.now["error"] = "Unknown prefix: #{prefix}. Available prefixes: #{SEARCH_PREFIXES}"
         next view("search")
       end
-      condition = terms.map { |t|
-        pattern = "%#{klass.dataset.escape_like(t)}%"
-        columns.map { Sequel.cast(it, :text).ilike(pattern) }.reduce(:|)
-      }.reduce(:|)
-      @search_results = klass.where(condition).limit(10).all
+      patterns = terms.map { "%#{klass.dataset.escape_like(it)}%" }
+      @search_results = klass.grep(columns, patterns).limit(11).all
+      if @search_results.length > 10
+        @truncated = @search_results.pop
+      end
 
       if @search_results.length == 1
         obj = @search_results.first
@@ -879,8 +881,8 @@ class CloverAdmin < Roda
         r.redirect("/model/#{klass.name}/#{ubid}")
       elsif (uuid = typecast_params.uuid("id")) && (ubid = UBID.to_ubid(uuid)) && (klass = UBID.class_for_ubid(ubid))
         r.redirect("/model/#{klass.name}/#{ubid}")
-      elsif typecast_params.nonempty_str("id")
-        r.redirect("/search?q=#{Rack::Utils.escape(typecast_params.nonempty_str("id"))}")
+      elsif (id = typecast_params.nonempty_str("id"))
+        r.redirect("/search?q=#{Rack::Utils.escape(id)}")
       end
 
       @grouped_pages = Page.reverse(:created_at, :summary).exclude(severity: "info").group_by_vm_host
--- a/views/admin/search.erb
+++ b/views/admin/search.erb
@@ -11,8 +11,8 @@
     results for
     <strong><%= @query %></strong>
   </p>
-  <% if @search_results.length >= 10 %>
-    <div class="alert alert-danger" role="alert">Results may be truncated. Try a more specific search term.</div>
+  <% if @truncated %>
+    <div class="alert alert-danger" role="alert">Results are truncated. Try a more specific search term.</div>
   <% end %>
   <ul id="search-results">
     <% @search_results.each do |obj| %>
--- a/spec/routes/web/admin/admin_spec.rb
+++ b/spec/routes/web/admin/admin_spec.rb
@@ -168,13 +168,13 @@ RSpec.describe CloverAdmin do
     expect(page).to have_link(account2.admin_label)
   end
 
-  it "shows truncation warning when a model has 10 or more results" do
-    10.times { |i| GithubInstallation.create(name: "trunctest-#{i}", installation_id: 800 + i, type: "User") }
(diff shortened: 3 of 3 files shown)


fdr:
Three generations ago, we went to full blown tsvector to do stuff like
this. It worked well. At the time we were large enough that we had a whole
separate app and database to manage its performance characteristics, but we
don’t necessarily have to do that this time especially if we can limit
GitHub actions churn.

fdr:
Incidentally, it's not an endorsement of appropriateness for the present time, though it might be worth considering. It's a reasonably scalable model as you can decide what's in the tsvectors and it can handle sub strings and even multiple match ranking okay.

jeremyevans:
We could also do a GIN index using `gin_trgm_ops` opclass from the pg_trgm extension to speed up the `ILIKE '%foo%'` query, instead of switching to tsvector. 

However, my recommendation is orthogonal to these potential performance improvements. In general, if an admin is looking for a specific type, we should only search that type. We should not default to searching all types, even if we allow it for the rare cases where it may be desired.

fdr:
I think it’s nice to put in a string and get “dwim” or even glance at
meaningful correlations. But it’s a fair question what strings we actually
would put in, and whether paying so much for an index is worth it.

For my own part, I think emails, and names of resources. Rarely, but
sometimes, a substring of a resource, but basically never a regexp or more
robust pattern.

We should watch out in presentation how to be robust to social engineering
attacks, e.g. condensing a ubid or noticing multiple, near matches that
span projects with different email domain suffixes. It’s a bit rainbow-y
but a trick is to hash projects by identifier to a color.

jeremyevans:
@fdr Do you have a guess as to how often you would be searching for something on the admin site without already knowing what type of object you are searching for?

fdr:
I'd want to paste a string from customer and see everything related. I found polymorphism of this kind immensely useful in the azure and gcp console, to see all the objects across types sharing a substring. Aws console's relative crappiness in this area is one of my least favorite things about it.

On azure and gcp, I'd even name things a certain way to make that feature work. Not clear the need goes away just because the administrivia is internal, or if we'd opt out of faceted search because we bear the cost of managing indexing on a small team.

fdr:
If we exposed a search bar in the console, I wonder if we could amortize the admin infrastructure, or rather, the other way around? The tricky part is authorization in one but not the other.

jeremyevans:
Exposing a search bar in the console is less of a performance issue, as all of the object types should have an appropriate index on the project_id, so it shouldn't result in sequential scans.

I'm not sure it's actually helpful for users, just as I'm not sure it's helpful for admins. However, your experience indicates it is helpful, so maybe we should add it.

fdr:
Definitely I'd name my stuff in a strategic way to be able to see, for example, the NIC and VM web pages in one search result. Note these were type-ahead search rather than a page transition though. I'm not sure if I want to invest in such a feature in the moment, but it's a fair question as long as we're thinking about it a polymorphic search over objects. Do you have any experience with small-bore typeahead search? Usually over a few thousand objects at most. And more often, fifty.

jeremyevans:
I've setup typeahead search before. It's not complicated, but you want to make sure the queries are indexed properly.  For ~50 records, it's possible to preload all possible results so you don't need to requery as you type.

Note that for many use cases, I think it would be better to have a more detailed overview page that linked to every top-level object you have in your project (i.e. linking to all VMs, PG databases, k8s clusters, private subnets, firewalls, and load balancers).

enescakir:
Let me explain my motivation. For example, during an abuse investigation:
- Click “Account”, click “Search”, enter the email, click “Search”, then click the found account.
- If it’s not found, go back to the home page, click “BillingInfo”, click “Search”, enter the Stripe ID, click “Search”, then click the found billing record.
- If it’s found, then click through to the project page, resources page, etc.
- For another record, I repeat the same process again and again. I constantly have to return to the home page and navigate to related resources.

So it involves a lot of clicks.

In general, I already know what I want to search for, so I’m comfortable using a syntax like `vm:name`. Let me refactor it.

clover_admin.rb
```diff
@@ -833,13 +844,41 @@ def self.object_action(...)
...
+    end
+
     r.root do
       if (ubid = typecast_params.ubid("id")) && (klass = UBID.class_for_ubid(ubid))
         r.redirect("/model/#{klass.name}/#{ubid}")
       elsif (uuid = typecast_params.uuid("id")) && (ubid = UBID.to_ubid(uuid)) && (klass = UBID.class_for_ubid(ubid))
         r.redirect("/model/#{klass.name}/#{ubid}")
       elsif typecast_params.nonempty_str("id")
-        flash.now["error"] = "Invalid ubid/uuid provided"
+        r.redirect("/search?q=#{Rack::Utils.escape(typecast_params.nonempty_str("id"))}")
```
```suggestion
      elsif (id = typecast_params.nonempty_str("id"))
        r.redirect("/search?q=#{Rack::Utils.escape(id)}")
      end
```

clover_admin.rb
```diff
@@ -833,13 +844,41 @@ def self.object_action(...)
...
+      if term.nil? || term.empty?
+        flash.now["error"] = "Use prefix:term syntax to search (e.g. vm:name). Available prefixes: #{SEARCH_QUERIES.map { "#{Object.const_get(it[0]).ubid_type} (#{it[0]})" }.join(", ")}"
+        next view("search")
+      end
+      term.strip!
+
+      klass = UBID.class_for_ubid(prefix)
+      columns = klass && SEARCH_QUERIES[klass.name]
+      unless columns
+        flash.now["error"] = "Unknown prefix: #{h(prefix)}. Available prefixes: #{SEARCH_QUERIES.map { "#{Object.const_get(it[0]).ubid_type} (#{it[0]})" }.join(", ")}"
```
jeremyevans: Flash messages are already HTML escaped, so escaping the prefix results in double escaping.

```suggestion
        flash.now["error"] = "Unknown prefix: #{prefix}. Available prefixes: #{SEARCH_QUERIES.map { "#{Object.const_get(it[0]).ubid_type} (#{it[0]})" }.join(", ")}"
```

I recommend precomputing the available prefixes and storing them in a constant so you don't need to do the same work every time. Especially considering you are duplicating the logic earlier in the route.

views/admin/search.erb
```diff
@@ -0,0 +1,22 @@
...
+  </p>
+<% elsif @search_results %>
+  <p>
+    <%= @search_results.length %>
+    results for
+    <strong><%= @query %></strong>
+  </p>
+  <% if @search_results.length >= 10 %>
+    <div class="alert alert-danger" role="alert">Results may be truncated. Try a more specific search term.</div>
+  <% end %>
```
jeremyevans: In cases like this, I would limit the dataset to 11 records, and if 11 are returned, set a flag and pop the last one off. That way you know for sure whether there are additional results. However, considering this is only for internal use, I'll leave it up to you whether the additional complexity is worth it.

clover_admin.rb
```diff
@@ -314,6 +314,17 @@ def self.object_action(...)
...
+    "Account" => [:email, :name],
+    "BillingInfo" => [:stripe_id],
+    "GithubInstallation" => [:name],
+    "GithubRepository" => [:name],
+    "Invoice" => [:invoice_number],
+    "KubernetesCluster" => [:name],
+    "PostgresResource" => [:name],
+    "Vm" => [:name]
+  }.freeze
+
```
jeremyevans, line 329:
```suggestion
  SEARCH_QUERIES.each_value(&:freeze)

```

clover_admin.rb
```diff
@@ -848,20 +848,22 @@ def self.object_action(...)
...
         flash.now["error"] = "Unknown prefix: #{h(prefix)}. Available prefixes: #{SEARCH_QUERIES.map { "#{Object.const_get(it[0]).ubid_type} (#{it[0]})" }.join(", ")}"
         next view("search")
       end
-      pattern = "%#{klass.dataset.escape_like(term)}%"
-      condition = columns.map { Sequel.cast(it, :text).ilike(pattern) }.reduce(:|)
+      condition = terms.map { |t|
+        pattern = "%#{klass.dataset.escape_like(t)}%"
+        columns.map { Sequel.cast(it, :text).ilike(pattern) }.reduce(:|)
+      }.reduce(:|)
       @search_results = klass.where(condition).limit(10).all
```
jeremyevans: Sequel has a shortcut for this, `Dataset#grep`. The filter used is if any column matches any pattern.

```suggestion
      patterns = terms.map { "%#{klass.dataset.escape_like(it)}%" }
      @search_results = klass.grep(columns, patterns).limit(10).all
```

enescakir:
I applied reviews at https://github.com/ubicloud/ubicloud/pull/4962/changes/5c928c15fec07757203a1ec8170f97df30d6f0f6

jeremyevans (approved):

PR #4964, opened 2026-03-12
Document restore `name` and `restoreTarget`

theory:
Document the `name` and `restoreTarget` properties in the restore API. Update the tests to use proper `rfc3339` syntax.

### DIFF
--- a/openapi/openapi.yml
+++ b/openapi/openapi.yml
@@ -1657,9 +1657,12 @@ paths:
               type: object
               properties:
                 name:
+                  description: Name for the newly restored Postgres database.
                   type: string
                 restore_target:
+                  description: The RFC 3339-formatted timestamp for the point in time at which to recover.
                   type: string
+                  format: date-time
                 pg_config:
                   type: object
                   example:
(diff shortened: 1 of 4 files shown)


theory:

openapi/openapi.yml
```diff
@@ -1509,8 +1509,10 @@ paths:
               properties:
                 name:
                   type: string
+                  description: Name for the newly restored Postgres database.
                 restore_target:
                   type: string
+                  description: The point in time at which to recover, as either date/time, named restore point, or a specific transaction ID.
```
byucesoy: We only accept date/time, not named restore point or transaction id.

theory: Oh, then should it also have `"format": "date-time"`?

byucesoy: Yeah, why not... I'm not sure if it would break anything though.

theory: Well, where is it processed? I can take a look to see how it's parsed.

serprex: should be detected by unit tests anyways, it'd only be a matter of whether the Ruby parsing has a wider acceptance than openapi `date-time`, but being stricter here on what's acceptable format seems fine

theory: Are there tests for the JSONSChema `date-time` format? It's [RFC3339/ISO-8601](https://www.rfc-editor.org/rfc/rfc3339.html#section-5.6) format, basically `2006-01-02T15:04:05.999999999Z07:00`.

theory: Changed in 1cf9d7d3d to:

> The RFC 3339-formatted timestamp for the point in time at which to recover.

PR #4965, opened 2026-03-12
Add delegator to postgres_server.resource in PostgresServerNexus

serprex:
Split out refactoring from external mTLS work

### DIFF
--- a/prog/postgres/postgres_server_nexus.rb
+++ b/prog/postgres/postgres_server_nexus.rb
@@ -777,7 +777,7 @@ SQL
     vm.sshable.invalidate_cache_entry
 
     # Don't declare unavailability if we are upgrading.
-    return true if resource.version != resource.target_version && postgres_server == resource.upgrade_candidate_server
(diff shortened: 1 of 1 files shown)


prog/postgres/postgres_server_nexus.rb
```diff
@@ -777,7 +777,7 @@ def available?
     vm.sshable.invalidate_cache_entry
 
     # Don't declare unavailability if we are upgrading.
-    return true if postgres_server.resource.version != postgres_server.resource.target_version && postgres_server == postgres_server.resource.upgrade_candidate_server
+    return true if resource.version != resource.target_version && postgres_server == resource.upgrade_candidate_server
```
serprex: this seems unsafe, seems like 2nd expression should be comparing ids

jeremyevans: `Sequel::Model#==` is defined, but it compares all values. I agree it would be better to compare ids. Alternatively, you could use `#===` which will do a class and primary key comparison.

serprex:
ended up merging with rest in #4874

PR #4967, opened 2026-03-13
UMI: add `bin/archive-url`

pykello:
This executable downloads a file from the given URL and archives it to an S3-compatible object store.

This will be useful when creating a machine image from publicly distributed images like ubuntu cloud images.

Usage format is:

```
host/bin/archive-url URL SHA256SUM VHOST_BLOCK_BACKEND_FORMAT
```

It accepts target configuration as a JSON document from stdin. Example:

```
payload = {
  "target_conf": {
     "endpoint": "https://s3-endpoint-url",
     "region": "auto",
     "bucket": "ubicloud-machine-images",
     "prefix": "pj1234567890/mi9876543210",
     "access_key_id": "valid-access-key-id",
     "secret_access_key": "valid-secret-access-key",
     "session_token": "optional-session-token",
     "archive_kek": {
       "algorithm": "aes-256-gcm",
       "key": "b64-encoded-32-byte-key"
     }
  }
}.to_json

sshable.cmd(
    "sudo host/bin/archive-url :url :sha256sum :version",
    url: url, sha256sum: sha256sum, version: version,
    stdin: payload)
```

The output is an archive that is compressed, encrypted, and uploaded to the object store. Its data encryption key is encrypted with `archive_kek` and stored in `metadata.json` in the object store.

We could have made encryption optional for such machine images when the URL is public. Always enforcing encryption in this case is mostly to reduce the number of cases we need to handle in the code, given that not encrypting doesn't provide much benefit.

### DIFF
--- a/rhizome/host/bin/archive-url
+++ b/rhizome/host/bin/archive-url
@@ -1,0 +1,29 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require "json"
+require_relative "../lib/storage_archive"
+
+unless (url = ARGV.shift)
+  puts "expected url as argument"
+  exit 1
+end
+
+unless (sha256sum = ARGV.shift)
+  puts "expected sha256sum as argument"
+  exit 1
+end
+
+unless (vhost_block_backend_version = ARGV.shift)
+  puts "expected vhost_block_backend_version as argument"
+  exit 1
+end
+
+params = JSON.parse($stdin.read)
+
+unless (target_conf = params["target_conf"])
+  puts "expected target_conf in params"
+  exit 1
+end
+
+StorageArchive.archive_url(url, sha256sum, target_conf, vhost_block_backend_version)
--- a/rhizome/host/lib/storage_archive.rb
+++ b/rhizome/host/lib/storage_archive.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
+require "tmpdir"
 require_relative "../../common/lib/util"
+require_relative "boot_image"
 require_relative "kek_pipe"
 require_relative "toml"
 require_relative "vhost_block_backend"
@@ -8,6 +10,7 @@ require_relative "vhost_block_backend"
 class StorageArchive
   include KekPipe
   include Toml
+  extend Toml
 
   def initialize(disk_config_path, disk_kek_path, disk_kek, target_conf, vhost_block_backend_version)
     validate_keys(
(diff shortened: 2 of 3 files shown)


rhizome/common/lib/util.rb
```diff
@@ -24,8 +24,8 @@ class FsyncFail < Exception
 end
 # rubocop:enable Lint/InheritException
 
-def r(*command, stdin: "", expect: [0])
-  stdout, stderr, status = Open3.capture3(*command, stdin_data: stdin)
+def r(*command, stdin: "", expect: [0], env: {})
+  stdout, stderr, status = Open3.capture3(env, *command, stdin_data: stdin)
```
jeremyevans: This is unnecessary. You should just provide a env hash as the first argument to `r`.

rhizome/host/bin/archive-url
```diff
@@ -0,0 +1,22 @@
+#!/bin/env ruby
+# frozen_string_literal: true
+
+require "json"
+require_relative "../../common/lib/util"
+require_relative "../lib/storage_archive"
+
+params = JSON.parse($stdin.read)
```
jeremyevans, line 22: Of the arguments, only `target_conf` looks sensitive. I would have the other 3 arguments be command line arguments.

rhizome/host/lib/storage_archive.rb
```diff
@@ -28,6 +30,35 @@ def initialize(disk_config_path, disk_kek_path, disk_kek, target_conf, vhost_blo
...
+
+    Dir.mktmpdir do |dir|
+      # download the image and convert it to raw format.
+      boot_image = BootImage.new("image", nil, image_root: dir)
+      boot_image.download(url: url, sha256sum: sha256sum)
+
+      # setup a disk with the image as the stripe source
+      disk_raw_path = File.join(dir, "disk.raw")
+      image_size = File.size(boot_image.image_path)
+      image_size_mib = (image_size + 1024 * 1024 - 1) / (1024 * 1024)
```
jeremyevans: `image_size_mib = (image_size / 1048576r).ceil`

rhizome/host/lib/storage_archive.rb
```diff
@@ -28,6 +30,35 @@ def initialize(disk_config_path, disk_kek_path, disk_kek, target_conf, vhost_blo
...
+
+      config_path = File.join(dir, "vhost-backend.conf")
+      safe_write_to_file(config_path, [
+        toml_section("device", {"data_path" => disk_raw_path, "metadata_path" => File.join(dir, "metadata")}),
+        toml_section("stripe_source", {"type" => "raw", "image_path" => boot_image.image_path}),
+        toml_section("danger_zone", {"enabled" => true, "allow_unencrypted_disk" => true})
+      ].join("\n"))
+
+      vp = VhostBlockBackend.new(vhost_block_backend_version)
+      r vp.init_metadata_path, "--config", config_path, env: {"RUST_LOG" => "info"}
```
jeremyevans: `r({"RUST_LOG" => "info"}, vp.init_metadata_path, "--config", config_path)`

rhizome/host/lib/storage_archive.rb
```diff
@@ -38,13 +69,17 @@ def archive
...
+    if @disk_kek_path
+      run_with_kek_pipe(
+        cmd,
+        kek_pipe: @disk_kek_path,
+        kek_content: @disk_kek["key"],
+        env: env,
+        stdin: target_config
+      )
+    else
+      r(*cmd, env: env, stdin: target_config)
```
jeremyevans: `r(env, *cmd, stdin: target_config)`

rhizome/host/bin/archive-url
pykello: Done.

We should do the same for `archive-storage-volume`, where it will have two sensitive keys (`target_conf` and `kek`), and stdin params will look like `{"kek": ..., "target_conf": ...}`. So, in this change I kept the stdin format as `{"target_conf": ...}` instead of flattening the target_conf hash, to have consistent formats between two commands.

I'll do the change for `archive-storage-volume` in a separate PR.

jeremyevans (approved):

PR #4970, opened 2026-03-13
Enable premium runners by default

enescakir:
- **Enable premium runners by default**

We decided to make premium runners default


bin/rake-task-runner
```diff
@@ -89,6 +89,7 @@ when "refresh_sequel_caches"
     File.delete(filename) if File.file?(filename)
   end
 
+  DB.instance_variable_set(:@schemas, {})
```
enescakir: @jeremyevans, it didn't update annotation without this change

jeremyevans: Schema cache file size dropped by 38.3KB with this. It's using an empty schema cache file, which is not what we want. I'll look into fixing this.

jeremyevans: Fix: https://github.com/ubicloud/ubicloud/pull/4975

aamederen:
We should make the docs mention that too

PR #4971, opened 2026-03-13
Run postgresql-init scripts during database initialization

serprex:
This can be used to create specific roles for managed service,
or setup extensions we want created by default


serprex:
for our  use case they're all the same, but a new postgres version could introduce new role group. script can get around that with conditional `version()` logic in future tho

I'd rather make it execute shell scripts if we feel checking `version()` in SQL is too ugly, in future we could add glob over `*.sh` for that, but don't see it being necessary

PR #4973, opened 2026-03-13
Use Roda response_attachment plugin

jeremyevans:
This simplifies the code slightly. It could possibly simplify it more if we are OK changing the response content types.

This changes the kubernetes cluster config content type to text/yaml, as it is a yaml file.

This also adds some specs for expected content-disposition/content-type headers.

This requires updating to Roda 3.102.0, so the first commit does that.


enescakir:
Tests are failing.

jeremyevans:
> Tests are failing.

Loosened the regexp on the invoice filename spec.

PR #4975, opened 2026-03-13
Fix refresh_sequel_caches rake task

jeremyevans:
We need to delete the cache files before requiring the loader.


enescakir (approved):
It worked!

https://github.com/ubicloud/ubicloud/pull/4970

PR #5015, opened 2026-03-18
Add resource_type and tags columns to billing records

achudnovskij:
ClickHouse billing integration needs to query billing records by
resource type and attach provider-specific metadata (region, instance
type, HA configuration, etc.) without joining against each resource
table. Today, determining the resource type requires looking up the
billing rate, and there is no structured place to store contextual
tags alongside a billing record.

Add a migration that introduces two columns to billing_record:
- resource_type (text): classifies the record (e.g. VmVCpu,
  PostgresStorage, GitHubRunnerMinutes, InferenceTokens)
- tags (jsonb): stores arbitrary key-value metadata

Populate resource_type at every BillingRecord.create call site across
all nexus programs (VM, Postgres, Kubernetes, GitHub runners, and
inference endpoints).

For Postgres resources, introduce an after_billing_record_created hook
with an override mechanism that attaches provider, region, instance
type, HA type, flavor, server count, storage size, and any
user-defined tags to each billing record.

Add dataset helper methods on BillingRecord (overlapping,
with_resource_types, distinct_resource) to support efficient querying
by the billing API.

### DIFF
--- a/migrate/20260317_billing_record_tags.rb
+++ b/migrate/20260317_billing_record_tags.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:billing_record) do
+      add_column :tags, :jsonb, null: false, default: "[]"
+      add_column :resource_type, :text, collate: '"C"', null: false, default: ""
+    end
+  end
+end
--- a/model/billing_record.rb
+++ b/model/billing_record.rb
@@ -7,6 +7,19 @@ class BillingRecord < Sequel::Model
 
   dataset_module do
     where(:active, Sequel.function(:upper, :span) => nil)
+
+    def overlapping(start_time, end_time)
+      where(Sequel.pg_range(:span).overlaps(start_time...end_time))
+    end
+
+    def with_resource_types(resource_types)
+      where(resource_type: resource_types)
+    end
+
+    def distinct_resource
+      distinct(:resource_id, :resource_type)
+        .order(:resource_id, :resource_type, Sequel.desc(Sequel.function(:lower, :span)))
+    end
   end
 
   plugin ResourceMethods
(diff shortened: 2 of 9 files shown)


migrate/20260317_billing_record_tags.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:billing_record) do
+      add_column :tags, :jsonb, null: false, default: "[]"
+      add_column :resource_type, :text, collate: '"C"', null: false, default: ""
```
byucesoy, line 7: In this table, we already have billing_rate_id, which can be used to look up to resource_type.

In billing_record.rb file, you can access to resource type with `billing_rate["resource_type"]`

migrate/20260317_billing_record_tags.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
```
byucesoy, line 1: Migrating the test database will update the schema.cache file, so please make sure to commit that as well. We also keep migration-related changes (the migration itself, schema.cache updates, and annotation changes at the bottom of model files) in a separate commit before the rest of the changes. This allows us to deploy the migration first and update the database before rolling out code that depends on the new columns.

For column removals, the migration commit should come last.

migrate/20260317_billing_record_tags.rb
achudnovskij: Our goal is to be able to query list of postgres database instances in a single database query with efficient enough joins.
What would be the way to achieve this? the reason I've added resource type here is to be able to index the table by that property and query all postgreSQL resources in one go.
Please suggest alternatives

migrate/20260317_billing_record_tags.rb
achudnovskij: Yes, Ack. This PR is draft now, I will split once we reach agreement

migrate/20260317_billing_record_tags.rb
achudnovskij: Plus, we need to be able to look up up to 90 days back. My understanding billing_rate is loaded from a file with the current values, not historical. 

One more alternative I see is we can move the logic into the extension. Given we're adding a gin index for tags anyway, we will be able to take this logic out of the generic Ubicloud processing and handle it in our override.

PR #5020, opened 2026-03-18
Add custom label management UI

enescakir:
- **Add custom label CRUD routes for GitHub installations**
  GithubCustomLabel records let users create friendly runner label
  aliases (e.g. my-team-large -> ubicloud-standard-16) with optional
  concurrency limits. Until now these could only be managed via the
  admin panel or directly in the database.
  
  Add feature flag custom_runner_labels and a full set of web routes
  (list, create, edit, delete) under /github/:installation/custom-label,
  gated behind the flag. All mutations are audit-logged.
  
- **Add custom labels views, tab, and tests for GitHub installations**
  Add the list page, create/edit form, and a conditional "Custom
  Labels" tab in the GitHub installation tabbar. The tab only appears
  when the custom_runner_labels feature flag is enabled.
  
  The list page shows each label's name, alias, concurrency limit,
  and allocated runner count with edit and delete controls. The form
  provides a dropdown of all available runner labels from the config.
  
  Tests cover feature flag gating, CRUD operations, validation errors,
  and 404 handling.
  
- **Disallow custom labels with names starting with "ubicloud"**
  The "ubicloud" prefix is reserved for built-in runner labels.
  Allowing custom labels with this prefix would create ambiguity
  during runner provisioning. Add a model-level validation so the
  constraint is enforced regardless of the entry point.
  
- **Scroll to custom labels table after create/update/delete operations**
  Add id to the custom labels table_card and append #custom-labels-table
  anchor to redirect URLs so the browser scrolls to the relevant section.
  
  Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

CleanShot 2026-03-19 at 01 11 22@2x

CleanShot 2026-03-19 at 01 12 13@2x

CleanShot 2026-03-19 at 01 12 21@2x

### DIFF
--- a/routes/project/github.rb
+++ b/routes/project/github.rb
@@ -47,6 +47,9 @@ class Clover
       end
 
       r.get web?, "setting" do
+        if @project.get_ff_custom_runner_labels
+          @custom_labels = @installation.custom_labels_dataset.order(:name).all
+        end
         view "github/setting"
       end
 
@@ -92,6 +95,59 @@ class Clover
           no_audit_log
         end
         r.redirect @installation, "/setting"
+      end
+
+      r.on web?, "custom-label" do
+        next unless @project.get_ff_custom_runner_labels
+
+        r.get "create" do
+          view "github/custom_label_form"
+        end
+
+        r.post "create" do
+          handle_validation_failure("github/custom_label_form")
+          label = GithubCustomLabel.new(installation_id: @installation.id)
+          label.name = typecast_params.nonempty_str!("name")
+          label.alias_for = typecast_params.nonempty_str!("alias_for")
+          label.concurrent_runner_count_limit = typecast_params.pos_int("concurrent_runner_count_limit")
+          DB.transaction do
+            label.save_changes
+            audit_log(label, "create")
+          end
+          flash["notice"] = "Custom label '#{label.name}' created"
+          r.redirect @installation, "/setting"
+        end
+
+        r.on :ubid_uuid do |id|
+          @custom_label = @installation.custom_labels_dataset.with_pk(id)
+          check_found_object(@custom_label)
+
+          r.get true do
+            view "github/custom_label_form"
+          end
+
+          r.post "patch" do
+            handle_validation_failure("github/custom_label_form")
+            @custom_label.name = typecast_params.nonempty_str!("name")
+            @custom_label.alias_for = typecast_params.nonempty_str!("alias_for")
+            @custom_label.concurrent_runner_count_limit = typecast_params.pos_int("concurrent_runner_count_limit")
+            DB.transaction do
+              @custom_label.save_changes
+              audit_log(@custom_label, "update")
+            end
+            flash["notice"] = "Custom label '#{@custom_label.name}' updated"
+            r.redirect @installation, "/setting"
+          end
+
+          r.delete true do
+            DB.transaction do
+              @custom_label.destroy
+              audit_log(@custom_label, "destroy")
+            end
+            flash["notice"] = "Custom label '#{@custom_label.name}' deleted"
+            r.redirect @installation, "/setting"
+          end
+        end
       end
 
       r.on web?, "runner" do
--- a/routes/project/github.rb
+++ b/routes/project/github.rb
@@ -115,7 +115,7 @@ class Clover
             audit_log(label, "create")
           end
           flash["notice"] = "Custom label '#{label.name}' created"
-          r.redirect @installation, "/setting"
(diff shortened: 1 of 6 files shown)


routes/project/github.rb
```diff
@@ -73,6 +76,59 @@ class Clover
...
+          label.concurrent_runner_count_limit = typecast_params.pos_int("concurrent_runner_count_limit")
+          DB.transaction do
+            label.save_changes
+            audit_log(label, "create")
+          end
+          flash["notice"] = "Custom label '#{label.name}' created"
+          r.redirect @installation, "/setting#custom-labels-table"
+        end
+
+        r.on :ubid_uuid do |id|
```
jeremyevans, line 121: You have almost the same code for creating new labels and editing existing labels. You should avoid the duplication:


```suggestion
        r.on [:ubid_uuid, "new"] do |id|
```

id will be `new` for a new label, or a uuid for an existing label.

routes/project/github.rb
```diff
@@ -73,6 +76,59 @@ class Clover
...
+          check_found_object(@custom_label)
+
+          r.get true do
+            view "github/custom_label_form"
+          end
+
+          r.post "patch" do
+            handle_validation_failure("github/custom_label_form")
+            @custom_label.name = typecast_params.nonempty_str!("name")
+            @custom_label.alias_for = typecast_params.nonempty_str!("alias_for")
```
jeremyevans, line 132: There should be some validation of these values. At the very least, there should be a size limit. Currently, it looks like a user could upload a very large value for either.

spec/routes/web/project/github_spec.rb
jeremyevans, line 1: The first and second commits should be combined, as the first commit depends on the second commit.

spec/routes/web/project/github_spec.rb
```diff
@@ -333,6 +333,157 @@
...
+    it "creates a custom label" do
+      visit "#{project.path}/github/#{installation.ubid}/custom-label/create"
+
+      fill_in "name", with: "my-label"
+      select "ubicloud-standard-4-ubuntu-2404", from: "alias_for"
+      fill_in "concurrent_runner_count_limit", with: "3"
+      click_button "Add Custom Label"
+
+      expect(page).to have_flash_notice("Custom label 'my-label' created")
+      expect(GithubCustomLabel.first(name: "my-label")).not_to be_nil
```
jeremyevans, line 414: This line is unnecessary as it is implicitly checked in the following line.

```suggestion
```

routes/project/github.rb
```diff
@@ -73,6 +76,59 @@ class Clover
...
+
+        r.on :ubid_uuid do |id|
+          @custom_label = @installation.custom_labels_dataset.with_pk(id)
+          check_found_object(@custom_label)
+
+          r.get true do
+            view "github/custom_label_form"
+          end
+
+          r.post "patch" do
```
jeremyevans, line 129: I would avoid using "patch" here. If this were using `r.patch` for API support, it would be used implicitly. However, as this is currently web-only, it doesn't seem necessary.

```suggestion
          r.post true do
```

views/github/custom_label_form.erb
```diff
@@ -0,0 +1,52 @@
...
+else
+  heading = "Add"
+  action = "#{path(@installation)}/custom-label/create"
+end
+
+@page_title = "#{heading} Custom Label" %>
+
+<%== render("github/tabbar") %>
+
+<% form(action:, method: :post) do %>
```
jeremyevans, line 16: You can avoid the need for an `action` parameter here if the path for GET and POST is the same (which I recommend, as it makes things simpler).

views/github/setting.erb
```diff
@@ -136,4 +136,33 @@
...
+      title: "Custom Labels",
+      right_items: @custom_labels.empty? ? nil : [part("components/button", text: "Add Custom Label", link: "#{installation_path}/custom-label/create")],
+      headers: ["Name", "Alias For", "Concurrency Limit", "Active Runners", "", ""],
+      rows: @custom_labels.map do |label|
+        [
+          [
+            label.name,
+            label.alias_for,
+            (label.concurrent_runner_count_limit || "Unlimited").to_s,
+            label.allocated_runner_count.to_s,
```
jeremyevans, line 153: Are these `to_s` calls actually necessary? I'm not sure, but a brief review of the table_card and value_content components indicates they may not be.

enescakir:
No one is using it at the moment. Let's not complicate things for now.

PR #5030, opened 2026-03-19
Combine rhizome/{host,common}/Gemfile into rhizome/Gemfile

jeremyevans:
To avoid the issues mentioned in #4947, combine the rhizome
host/common Gemfiles into a single Gemfile, so broken tooling that
automatically creates Gemfile.lock files still works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,9 +19,6 @@
 # Used by rake by
 /bin/by
 
-# rhizome/Gemfile is combination of all Gemfiles in subdirectories, so does rhizome/Gemfile.lock
-/rhizome/*/Gemfile.lock
-
 # Frontend
 /node_modules
 
--- a/prog/install_rhizome.rb
+++ b/prog/install_rhizome.rb
@@ -53,7 +53,9 @@ class Prog::InstallRhizome < Prog::Base
   end
 
   label def install_gems
-    sshable.cmd("bundle config set --local path vendor/bundle && bundle install")
+    if frame["target_folder"] == "host"
+      sshable.cmd("bundle config set --local path vendor/bundle && bundle install")
+    end
 
     hop_validate
   end
--- a/rhizome/Gemfile
+++ b/rhizome/Gemfile
@@ -4,7 +4,14 @@
 # `BUNDLE_GEMFILE=rhizome/Gemfile bundle install` works from repository root,
 # as that is what rhizome CI uses.
 
-# Load Gemfiles
-Dir.glob("*/Gemfile", base: __dir__) do |file|
-  instance_eval File.read(File.join(__dir__, file))
+source "https://rubygems.org"
+
+gem "netaddr"
+
+if File.directory?(File.join(__dir__, "host"))
+  gem "base64"
+
+  group :test do
+    gem "rspec"
+  end
 end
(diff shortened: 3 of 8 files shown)


jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23308891953

fdr:
Do we retain control over point releases of dependencies in this mode?

jeremyevans:
> Do we retain control over point releases of dependencies in this mode?

I originally thought we didn't have that control already as I thought `rhizome/Gemfile.lock` was not committed to the repository, but that's apparently not the case (it is committed). So I removed the second commit. So this PR is now only about merging the `rhizome/host/Gemfile` and `rhizome/common/Gemfile` into `rhizome/Gemfile`.

serprex (approved):

fdr:
doesn't this make it start installing netaddr for workloads that don't need them? I think we only(?) need it for the subnet-calculation heavy host slicing part. May be acceptable, but any way around that?

jeremyevans:
As `netaddr` was in the common Gemfile, which was in every rhizome installation, I believe it was installed in all rhizomes already.

Note that it appears `netaddr` only used for the host and inference_endpoint parts, though, so we could check for those directories and only include it if at least one of those exists.

fdr:
I think we should down-scope somehow, its use is not so pervasive.

jeremyevans:
Looking at the code, `rhizome/inference_endpoint/lib/replica_setup.rb` requires `netaddr`, but I cannot see where it is actually using it. @bsatzger Does the inference endpoint code actually use `netaddr`?

jeremyevans:
For that matter, I'm not sure that inference endpoint replica setup is using any of the libraries it is requiring:

```ruby
require "fileutils"
require "netaddr"
require "json"
require "openssl"
require "base64"
require "uri"
```

byucesoy (approved):

jeremyevans:
Pushed a few more commits:

* Remove unnecessary library requires from `rhizome/inference_endpoint/lib/replica_setup.rb`.
* Only `bundle install` in rhizome for hosts, as other rhizome users only use gems that ship with Ruby.
* Remove the host directory check in the rhizome Gemfile.

We could move the rhizome Gemfile to rhizome/host/Gemfile, so it doesn't even get installed on hosts that don't need it, but this doesn't take that step yet.

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23359506586

bsatzger:
> * Remove unnecessary library requires from `rhizome/inference_endpoint/lib/replica_setup.rb`.

thanks!

bsatzger (approved):

fdr (approved):
very nice

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/26842684915

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/26854145968

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/26865531707

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/26890924482

jeremyevans:
E2E passed.

PR #5059, opened 2026-03-24
Guard metrics scraping against missing done directory

achudnovskij:
- On a freshly provisioned VM the metrics agent hasn't run yet, so the `done/` directory does not exist. Without a guard, SSH calls to ls/cat/find that directory raise errors.
- Extract `metrics_done_dir_exists?` into `MetricsTargetMethods` so the check is defined once and shared.
- `scrape_endpoints` guards on it and returns `[]` early; `observe_metrics_backlog` reuses the same helper, removing the duplicate inline check it previously had.

### DIFF
--- a/lib/metrics_target_methods.rb
+++ b/lib/metrics_target_methods.rb
@@ -47,6 +47,8 @@ module MetricsTargetMethods
   end
 
   def scrape_endpoints(session)
+    return [] unless metrics_done_dir_exists?(session)
+
     scrape_files = session[:ssh_session].exec!("ls :metrics_dir/done | sort | head -n :fetch_count", metrics_dir:, fetch_count: MAX_SCRAPE_FETCH_COUNT).split("\n")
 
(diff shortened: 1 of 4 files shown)


lib/metrics_target_methods.rb
```diff
@@ -47,6 +47,8 @@ def export_metrics(session:, tsdb_client:)
   end
 
   def scrape_endpoints(session)
+    return [] unless metrics_done_dir_exists?(session)
```
byucesoy, line 50: This is one extra SSH call for every metrics check only to cover a case that happens once in the server's entire lifetime.

It would be better just to catch the exception and ignore it.

achudnovskij: Let me check more when it's invoked. I did see failures because of that, though agree with your point. Ideally this check is only needed when a new VM is coming up, not any other time in the livecycle

PR #5070, opened 2026-03-25
Expand metric destination with bearer token auth & mTLS support

serprex:
auth_type will be used to support bearer token auth,
mtls is used to require client cert for connection

prometheus designs mtls as connection authorization,
still need to layer authentication on top

CLI support remains backwards compatible, keeping basic auth setup simple

Web only exposes basic auth configuration

Certs for mTLS created with with cert/create-client-keypair endpoint


prog/postgres/postgres_server_nexus.rb
```diff
@@ -252,12 +252,14 @@ def before_run
...
-    username: '#{it.username}'
-    password: '#{it.password}'
-METRIC_DESTINATION
+      config = "- url: '#{it.url}'\n"
+      config << if it.auth_type == "bearer"
+        "  authorization:\n    credentials: '#{it.password}'\n"
+      else
+        "  basic_auth:\n    username: '#{it.username}'\n    password: '#{it.password}'\n"
+      end
+      config << "  tls_config:\n    cert_file: /etc/ssl/certs/client.crt\n    key_file: /etc/ssl/certs/client.key\n" if it.mtls
```
serprex: I've misunderstood this. for mTLS ideally customer would upload a signed cert for us to use. But uploading certs kind of defeats the purpose. Other way is they trust resource's client CA, but then we need to expose downloading certs for that CA

byucesoy: how about they share their public CA and we trust that instead?

PR #5087, opened 2026-03-26
UMI: filter hosts & backends by required vhost block backend version

pykello:
### UMI: filter hosts by min vhost block backend version for UMI source VMs

When `track_written` is enabled, the VM must be allocated only on hosts with ubiblk version `v0.4.0` or later, since archiving support was introduced in `v0.4.0`.

### UMI: respect minimum version in allocate_vhost_block_backend() 

Some features (e.g., archiving) require a minimum ubiblk version. Update `allocate_vhost_block_backend()` to enforce the minimum eligible version when selecting backends installed on the host.

### DIFF
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -650,7 +650,12 @@ module Scheduling::Allocator
       rand_choice.id
     end
 
-    def self.allocate_vhost_block_backend(backends)
+    def self.allocate_vhost_block_backend(backends, min_version: nil)
+      if min_version
+        parsed_min_version = VhostBlockBackend.parse_version(min_version)
+        backends.select! { |b| b.parsed_version >= parsed_min_version }
+      end
+
       total_weight = backends.sum(&:allocation_weight)
       fail "Total weight of all eligible vhost_block_backends shouldn't be zero." if total_weight == 0
 
@@ -696,7 +701,10 @@ module Scheduling::Allocator
           spdk_installation_id = StorageAllocation.allocate_spdk_installation(vm_host.spdk_installations)
           use_bdev_ubi = SpdkInstallation[spdk_installation_id].supports_bdev_ubi? && volume["boot"]
         else
-          vhost_block_backend_id = StorageAllocation.allocate_vhost_block_backend(vm_host.vhost_block_backends)
+          vhost_block_backend_id = StorageAllocation.allocate_vhost_block_backend(
+            vm_host.vhost_block_backends,
+            min_version: @request.minimum_vhost_block_backend_version
+          )
           use_bdev_ubi = false
         end
 
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -26,7 +26,9 @@ module Scheduling::Allocator
       true, # use slices
       Option::VmFamilies.find { it.name == vm.family }&.require_shared_slice || false,
       vm.project.get_ff_allocator_diagnostics || false,
-      family_filter
+      family_filter,
+      os_filter,
+      (storage_volumes.any? { it["track_written"] }) ? "v0.4.0" : nil
     )
     allocation = Allocation.best_allocation(request)
     fail "#{vm} no space left on any eligible host" unless allocation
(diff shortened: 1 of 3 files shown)


scheduling/allocator.rb
```diff
@@ -239,6 +242,27 @@ def self.candidate_hosts(request)
...
+
+        vbb = Sequel[:vhost_block_backend]
+        version_without_v = Sequel.function(:ltrim, vbb[:version], "v")
+
+        subquery = DB[:vhost_block_backend]
+          .where(vbb[:vm_host_id] => Sequel[:vm_host][:id])
+          .exclude(Sequel[vbb][:allocation_weight] => 0)
+          .where(Sequel.lit("? ~ ?", vbb[:version], '^v?\d+\.\d+\.\d+$'))
+          .where do
+            major = Sequel.cast(Sequel.function(:split_part, version_without_v, ".", 1), Integer)
```
bsatzger: up to you: the tree lines to get major, minor, patch could probably be replaced by a single line `using string_to_array`, something along the lines of:
```ruby
version = Sequel.cast(Sequel.function(:string_to_array, version_without_v, "."), Integer[])
```
then it's already an array and we wouldn't do
`[major, minor, patch] >= ...` but `version >= ...`

bsatzger: then we could also do 
```
min_version = VhostBlockBackend.parse_version(request.minimum_vhost_block_backend_version).segments
```
and compare `version >= min_version`
.

bsatzger (approved):

scheduling/allocator.rb
```diff
@@ -239,6 +242,22 @@ def self.candidate_hosts(request)
...
+
+        version = Sequel.cast(Sequel.function(:string_to_array, Sequel.function(:ltrim, vbb[:version], "v"), "."), "integer[]")
+        min_version = VhostBlockBackend.parse_version(request.minimum_vhost_block_backend_version).segments
+
+        subquery = DB[:vhost_block_backend]
+          .where(vbb[:vm_host_id] => Sequel[:vm_host][:id])
+          .exclude(Sequel[vbb][:allocation_weight] => 0)
+          .where { vbb[:version] =~ /^v\d+\.\d+\.\d+$/ }
+          .where(version >= Sequel.pg_array(min_version, :integer))
+          .select(1)
```
jeremyevans, line 256: Parsing a text column into an array of integers to query it seems like a suboptimal design to me. This is doing a correlated subquery, so it would be considering the backends for every matching host. I suppose you could use a functional index to speed things up.

If you've already tested this on the production database, and it does not have a significant negative impact on the VM allocation query, you can ignore this comment.

spec/scheduling/allocator_spec.rb
```diff
@@ -1642,5 +1642,17 @@ def create_req(vm, storage_volumes, target_host_utilization: 0.55, distinct_stor
       vbb_2 = VhostBlockBackend.new(allocation_weight: 100) { it.id = VhostBlockBackend.generate_uuid }
       expect(Al::StorageAllocation.allocate_vhost_block_backend([vbb_1, vbb_2])).to eq(vbb_2.id)
     end
+
+    it "filters backends by min_version" do
+      vbb_1 = VhostBlockBackend.new(allocation_weight: 100, version: "v0.3.0") { it.id = VhostBlockBackend.generate_uuid }
```
jeremyevans: You can use `new_with_id` to make sure a new object has a generated id, instead of using a block.

scheduling/allocator.rb
pykello: I ran candidate selection query with min version constraint & without constraint in the prod database. Without the constraint it took 57ms, with the constraint it took 61ms. I used min version of `[0,1,0]` so it matches all hosts. I should probably do more benchmarks with different params.

Anyway, I guess there's not much reason to do the string parsing here & there. What do you think if I migrate the table, so the version column becomes an `integer[]` & we index it?

This way, we can do the parsing only at `SetupVhostBlockBackend` time & save the version in db as array.

jeremyevans: The performance difference currently seems acceptable. However, I do agree that changing the column storage seems better in the long run.  You could use an integer array, or just an integer: `v1.2.3` -> `10203`

pykello: The version that uses numeric comparison: https://github.com/ubicloud/ubicloud/pull/5113

PR #5091, opened 2026-03-27
Subnet rekey v2

fdr:
This is based on the original furkan patches. I tried to separate, some of these (like conditional update, the strand changes) can be individually deployed and carefully watched. The patch is not exactly easy but not as scary as the sheer bulk there might suggest: a lot of it is TLA, some of it is the otherwise out-of-production TLA proof tool, and github actions stuff.

Even *with* embedded TLA, here's the more crucial diffstat, which I think is not too unreasonable:

```
git diff --stat ef15bf72e99fd3d8e5abba8c844fb3e8d7ab12d6 lib model prog

 lib/semaphore_methods.rb        |   5 +++
 model/metal/private_subnet.rb   |  23 +++++++++++++
 model/nic.rb                    |  31 ++++++++++--------
 model/private_subnet.rb         |   4 ++-
 model/semaphore.rb              |  14 +++++++-
 model/strand.rb                 |  60 ++++++++++++++++++++++++++++++++--
 prog/base.rb                    |   4 +++
 prog/vnet/metal/nic_nexus.rb    |  96 ++++++++++++++++++++++++++++++++++++++++++++++++------
 prog/vnet/metal/subnet_nexus.rb | 259 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++----------------------------
 9 files changed, 418 insertions(+), 78 deletions(-)

```


prog/vnet/metal/subnet_nexus.rb
```diff
@@ -39,6 +47,12 @@ def gen_reqid
   end
 
   label def refresh_keys
+    unless connected_leader?
+      Clog.emit("No longer the connected leader", {not_connected_leader: {private_subnet:, connected_leader:}})
+      private_subnet.incr_refresh_keys
```
jeremyevans, line 118: `incr_refresh_keys` looks wrong at this point. If this subnet isn't the connected leader, this increments `refresh_keys` and hops to `wait`. `wait` will see `refresh_keys` is set, decrement `refresh_keys`, and hop to `refresh_keys`. This looks like an infinite loop until something else breaks the cycle.

Looks like this is fixed in a later commit that adds a `decr_refresh_keys` without a hop to `refresh_keys` if the current subnet is not the connected leader.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -3,14 +3,61 @@
...
+      Clog.emit("SubnetNexus forwarding refresh_keys to leader",
+        {subnet_rekey_forward: {subnet_id: private_subnet.id,
+                                connected_leader_id: private_subnet.connected_leader_id}})
+      # TLA   ∧ refreshNeeded' = [refreshNeeded EXCEPT ![s] = 0, ![ConnectedLeader(s)] = @ + 1]
+      decr_refresh_keys
+      PrivateSubnet[private_subnet.connected_leader_id].incr_refresh_keys
+      # TLA   ∧ UNCHANGED ⟨edges, pc, heldLocks, ops, nicPhase, activeNics⟩
+      # TLA
+      nap 0
+    end
```
jeremyevans, line 40: This should be moved to the top of the `when_refresh_keys_set?` block below, inside an `unless connected_leader?` block.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
-      Semaphore.where(strand_id: nics.map(&:id), name: "lock").delete(force: true)
-      update_stack_locked_nics(nil)
+    nics = locked_nics
+    # TLA   ∧ heldLocks[s] ≠ {}
+    abort_rekey_if_no_nics(nics)
+    bad = nics.reject { |n| %w[outbound old_drop].include?(n.rekey_phase) }
+    fail "BUG: phase monotonicity at phase_old_drop: #{bad.map { "#{it.id}=#{it.rekey_phase}" }}" if bad.any?
+    # TLA   ∧ ∀ n ∈ heldLocks[s] : nicPhase[n] = "old_drop"
+    if nics.all? { |nic| nic.rekey_phase == "old_drop" }
+      PrivateSubnet.where(id: nics.map(&:private_subnet_id).uniq).update(last_rekey_at: Time.now)
```
jeremyevans, line 249:
```suggestion
      PrivateSubnet.where(id: nics.map(&:private_subnet_id).uniq).update(last_rekey_at: Sequel::CURRENT_TIMESTAMP)
```

spec/prog/vnet/metal/subnet_nexus_spec.rb
```diff
@@ -102,6 +109,13 @@
       described_class.new(Strand.create(prog: "Vnet::Metal::SubnetNexus", label: "refresh_keys", id: ps.id))
     }
 
+    it "hops to wait if not the connected leader" do
+      expect(nx).to receive(:connected_leader?).and_return(false)
```
jeremyevans, line 156: Mocking database access is generally something we avoid in new code. Is this done because you don't want to control the `id` columns of the subnets to ensure the current subnet is not the leader? It may be acceptable, but there should probably at least be comments explaining why the choice was made.

spec/prog/vnet/metal/subnet_nexus_spec.rb
```diff
@@ -89,9 +109,39 @@
...
     end
 
+    it "forwards refresh_keys to connected leader when not the leader" do
+      nx # ensure ps has a strand
+      Strand.create(prog: "Vnet::Metal::SubnetNexus", label: "wait", id: ps2.id)
+      ps.connect_subnet(ps2)
+
+      leader_id = [ps.id, ps2.id].min
+      non_leader = (leader_id == ps.id) ? ps2 : ps
+      leader = (leader_id == ps.id) ? ps : ps2
```
jeremyevans, line 119:
```suggestion
      leader, non_leader = [ps, ps2].sort_by(&:id)
```

proof/lib/proof_extract.rb
```diff
@@ -0,0 +1,264 @@
...
+    "\u2264" => "<=",            # ≤ → <=
+    "\u2265" => ">="             # ≥ → >=
+  }.freeze
+
+  # Translate unicode mathematical operators to TLA+ ASCII.
+  def self.unicode_to_tla(line)
+    UNICODE_TO_TLA.each do |unicode, ascii|
+      line = line.gsub(unicode) { ascii }
+    end
+    line
```
jeremyevans, line 41: Let's use a more efficient approach here:
```suggestion
    line.gsub(UNICODE_TO_TLA_REGEXP, UNICODE_TO_TLA)
```

proof/lib/proof_extract.rb
```diff
@@ -0,0 +1,264 @@
...
+    "\u2260" => "#",            # ≠ → #
+    "\u2227" => "/\\",          # ∧ → /\
+    "\u2228" => "\\/",          # ∨ → \/
+    "\u21D2" => "=>",           # ⇒ → =>
+    "\u27E8" => "<<",           # ⟨ → <<
+    "\u27E9" => ">>",           # ⟩ → >>
+    "\u21A6" => "|->",          # ↦ → |->
+    "\u2264" => "<=",            # ≤ → <=
+    "\u2265" => ">="             # ≥ → >=
+  }.freeze
```
jeremyevans, line 34:
```suggestion
  }.freeze
  UNICODE_TO_TLA_REGEXP = Regexp.union(UNICODE_TO_TLA.keys)
```

proof/lib/proof_extract.rb
```diff
@@ -0,0 +1,264 @@
...
+    ["~>", "\u219D"],           # ↝
+    ["~", "\u00AC"]            # ¬
+  ].freeze
+
+  # Translate TLA+ ASCII operators to Unicode for pretty-printing.
+  def self.tla_to_unicode(line)
+    ASCII_TO_UNICODE.each do |ascii, unicode|
+      line = line.gsub(ascii, unicode)
+    end
+    line
```
jeremyevans:
```suggestion
  # Translate TLA+ ASCII operators to Unicode for pretty-printing.
  def self.tla_to_unicode(line)
    line.gsub(TLA_TO_UNICODE_REGEXP, TLA_TO_UNICODE)
```

jeremyevans (commented):
I reviewed the first 5 commits and part of the 6th. I'll continue my review next week.

Placing the TLA comments inline with the source code makes understanding the source code significantly more difficult. I recommend having all TLA syntax in a separate file. Having the source code and TLA syntax in the same file only makes sense to me if the person modifying the source code is expected to modify the TLA syntax, and I highly doubt that is the case. Most likely, an LLM will be modifying the TLA syntax after a person makes changes to the source code. An LLM can probably handle the TLA syntax in a separate file just as easily as having it in the same file.

proof/lib/proof_extract.rb
```diff
@@ -0,0 +1,264 @@
...
+          in_block = false
+          block_start = nil
+        elsif in_scope
+          result << [unicode ? line : unicode_to_tla(line), lineno]
+        end
+      elsif stripped == "# TLA:begin"
+        in_block = true
+        block_start = lineno
+      elsif stripped.start_with?("# TLA")
+        rest = stripped.delete_prefix("# TLA")
```
jeremyevans, line 115:
```suggestion
      elsif (rest = stripped.delete_prefix!("# TLA"))
```

proof/lib/proof_extract.rb
```diff
@@ -0,0 +1,264 @@
...
+        if rest.empty?
+          # Bare "# TLA" → blank line
+          result << ["", lineno] if in_scope
+        elsif rest.start_with?(":")
+          # Metadata line — check for module scope
+          if rest.start_with?(":module")
+            name = rest.delete_prefix(":module").strip
+            in_scope = mod.nil? || name == mod
```
jeremyevans, line 123:
```suggestion
          if rest.delete_prefix!(":module")
            in_scope = mod.nil? || rest.strip == mod
```

proof/spec/proof_extract_spec.rb
```diff
@@ -0,0 +1,376 @@
...
+      result = described_class.unicode_to_tla(input)
+      expect(result).to eq("\\A x \\in S : \\E y \\in S : x \\union y \\intersect z \\subseteq w /\\ ~a \\/ b => c # d <<e>> f |-> g \\notin h")
+    end
+  end
+
+  describe ".tla_to_unicode" do
+    it "translates all ASCII TLA+ operators to unicode" do
+      input = "\\A x \\in S : \\E y \\in S : x \\union y \\intersect z \\subseteq w /\\ ~a \\/ b => c # d <<e>> f |-> g \\notin h"
+      result = described_class.tla_to_unicode(input)
+      expect(result).to eq("\u2200 x \u2208 S : \u2203 y \u2208 S : x \u222A y \u2229 z \u2286 w \u2227 \u00ACa \u2228 b \u21D2 c \u2260 d \u27E8e\u27E9 f \u21A6 g \u2209 h")
```
jeremyevans, line 32: Seems odd to use UTF-8 characters in `.unicode_to_tla` spec, but `\u*" escapes in these specs.

proof/tla
```diff
@@ -0,0 +1,190 @@
...
+def resolve_proof(name)
+  proofs = ProofExtract.load_prooffile(PROOFFILE)
+  proofs[name] || abort("error: proof '#{name}' not found in tla.yaml (have: #{proofs.keys.join(", ")})")
+end
+
+def run_tlc(name, spec, config_name: nil, overrides: {})
+  # Assemble
+  tla_path = File.join(spec[:dir], "#{name}.tla")
+  output_lines, source_map = ProofExtract.assemble(spec[:sources], mod: name)
+  File.write(tla_path, output_lines.join("\n") + "\n")
```
jeremyevans, line 34:
```suggestion
  File.write(tla_path, output_lines.join("\n") << "\n")
```

proof/tla
```diff
@@ -0,0 +1,190 @@
...
+  map_path = "#{tla_path}.map"
+  File.write(map_path, ProofExtract.source_map_to_json(source_map) + "\n")
```
jeremyevans, line 36:
```suggestion
  File.write(map_path, ProofExtract.source_map_to_json(source_map) << "\n")
```

proof/tla
```diff
@@ -0,0 +1,190 @@
...
+  warn "wrote #{tla_path} (#{output_lines.size} lines)"
```
jeremyevans, line 37: Seems odd to warn here. Maybe only in a debug mode? Similar issue elsewhere.

proof/tla
```diff
@@ -0,0 +1,190 @@
...
+  end
+end
+
+args = ARGV.dup
+subcmd = args.shift
+usage unless subcmd
+
+case subcmd
+when "list"
+  proofs = ProofExtract.load_prooffile(PROOFFILE)
```
jeremyevans, line 86:
```suggestion
  usage unless args.empty?
  proofs = ProofExtract.load_prooffile(PROOFFILE)
```

proof/tla
```diff
@@ -0,0 +1,190 @@
...
+  proofs.each do |name, spec|
+    configs = spec[:configs].keys
+    config_info = configs.empty? ? "" : "  configs: #{configs.join(", ")}"
+    puts "#{name}  (#{spec[:sources].size} sources, config: #{spec[:config]})#{config_info}"
+  end
+
+when "pretty"
+  usage if args.empty?
```
jeremyevans, line 94:
```suggestion
  usage unless args.size == 1
```

proof/tla
```diff
@@ -0,0 +1,190 @@
...
+  output_lines, = ProofExtract.assemble(spec[:sources], mod: name, unicode: true, backrefs: true)
+  $stdout.write(output_lines.join("\n") + "\n")
+
+when "check"
+  config_name = nil
+  if (idx = args.index("-c") || args.index("--config"))
+    args.delete_at(idx)
+    config_name = args.delete_at(idx)
+    abort "error: -c requires a config name" unless config_name
+  end
```
jeremyevans, line 106: Maybe worth considering an actual option parser both here and for assemble.

proof/tla
```diff
@@ -0,0 +1,190 @@
...
+    output_path ||= File.join(spec[:dir], "#{name}.tla")
+  end
+
+  output_lines, source_map = ProofExtract.assemble(input_paths, mod:)
+  tla_content = output_lines.join("\n") + "\n"
+
+  if output_path
+    File.write(output_path, tla_content)
+    map_path = "#{output_path}.map"
+    File.write(map_path, ProofExtract.source_map_to_json(source_map) + "\n")
```
jeremyevans, line 181:
```suggestion
    File.write(map_path, ProofExtract.source_map_to_json(source_map) << "\n")
```

proof/strand_schedule/mutation-test
```diff
@@ -0,0 +1,194 @@
...
+  }
+].freeze
+
+def assemble_spec
+  proofs = ProofExtract.load_prooffile(PROOFFILE)
+  spec = proofs[PROOF_NAME]
+  abort "error: proof '#{PROOF_NAME}' not found" unless spec
+
+  output_lines, = ProofExtract.assemble(spec[:sources], mod: PROOF_NAME)
+  [output_lines.join("\n") + "\n"]
```
jeremyevans, line 103:
```suggestion
  output_lines.join("\n") << "\n"
```

proof/subnet_rekey/mutation-test
```diff
@@ -0,0 +1,301 @@
...
+
+def assemble_spec
+  proofs = ProofExtract.load_prooffile(PROOFFILE)
+  spec = proofs[PROOF_NAME]
+  abort "error: proof '#{PROOF_NAME}' not found" unless spec
+
+  output_lines, = ProofExtract.assemble(spec[:sources], mod: PROOF_NAME)
+  cfg_text = ProofExtract.override_cfg(spec[:config], MUTATION_CFG)
+
+  [output_lines.join("\n") + "\n", cfg_text]
```
jeremyevans, line 182:
```suggestion
  [output_lines.join("\n") << "\n", cfg_text]
```

proof/subnet_rekey/mutation-test
```diff
@@ -0,0 +1,301 @@
...
+  end
+end
+
+def run_mutation(mutation)
+  tla_text, cfg_text = assemble_spec
+  mutated_tla, mutated_cfg = apply_mutation(tla_text, cfg_text, mutation)
+
+  # Per-mutation config overrides (e.g., larger MaxOps for specific mutations)
+  mutation[:cfg]&.each do |key, value|
+    mutated_cfg = mutated_cfg.gsub(/^(\s*#{Regexp.escape(key)}\s*=\s*).*$/, "\\1#{value}")
```
jeremyevans, line 212: Probably should have a method for this logic called by both `apply_mutation` and `run_mutation`.

proof/strand_schedule/mutation-test
```diff
@@ -0,0 +1,194 @@
...
+end
+
+def run_mutation(mutation)
+  tla_text, = assemble_spec
```
jeremyevans, line 107:
```suggestion
  tla_text = assemble_spec
```

proof/strand_schedule/mutation-test
jeremyevans, line 1: This has a lot of duplication with the other `mutation-test`.

lib/semaphore_methods.rb
```diff
@@ -21,6 +21,11 @@ def self.configure(model, *semaphore_names)
         define_method :"#{name}_set?" do
           semaphores.any? { it.name == name }
         end
+
+        # Class method for signaling by ID without loading the model.
+        define_singleton_method :"incr_#{name}" do |strand_id|
```
jeremyevans, line 26: I advise against this approach, as there is no guarantee that `strand_id` is an `id` for the same class as the one the method is defined. I would have callers call `Semaphore.incr` directly.  Plus, adding a class method per semaphore is wasteful when most class methods would never be called.

This seems to be used as `PrivateSubnet.incr_nic_phase_done(nic.rekey_coordinator_id)`. I would instead add a `Nic#rekey_coordinator_incr_nic_phase_done` or something similar that handled this case.

prog/vnet/metal/nic_nexus.rb
```diff
@@ -85,6 +108,8 @@ class Prog::Vnet::Metal::NicNexus < Prog::Base
 
     decr_destroy
 
+    # Hard-delete is load-bearing: FK cascade clears rekey_coordinator_id
+    # and rekey_phase, ensuring destroyed NICs release coordinator locks.
```
jeremyevans, line 156: I don't understand this comment. We aren't using `ON DELETE CASCADE` for the foreign key. Even if we were, the cascade is in the other direction (deleting `private subnet` would cascade to `rekey_coordinator_id`, deleting `nic` would not cause a cascade). Also, if we were using `ON DELETE CASCADE`, it would delete the referencing `nic` rows, which is not what we would want. The use of `clears` in the comment seems to be describing `ON DELETE SET NULL` behavior, which seems odd to describe using the word `cascade`. Regardless of the `ON DELETE` setting for `rekey_coordinator_id`, the `rekey_phase` column would not be modified.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -55,65 +62,116 @@ def gen_reqid
     SecureRandom.random_number(1...100000)
   end
 
+  # AbortRekey: all locked NICs destroyed mid-rekey → abort to idle.
```
jeremyevans, line 88: I would prefer keeping comments to ASCII other than those for TLA (similar issue later in commit).

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
+
+    # AllConnectedNics(s) = {} — leader but no NICs in component.
+    # No re-enqueue: nothing to rekey.  CreateNic will signal when a NIC appears.
+    if nics.empty?
+      hop_wait
+    end
+
+    # ∃ n ∈ AllConnectedNics(s) : IsLocked(n) — another coordinator holds these NICs.
+    # Re-enqueue: signal was consumed by ConsumeRefresh but ReadAndLock can't fire.
+    if nics.any?(&:rekey_coordinator_id)
```
jeremyevans, line 130: Maybe this is what we want, but it looks odd to me, since I would think it shouldn't complain if the `rekey_coordinator_id` matches the current `private_subnet`. I would think it should be:

```suggestion
    if nics.any? { it.rekey_coordinator_id &&  it.rekey_coordinator_id != private_subnet.id}
```

However, maybe there is a reason why this check is not needed.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
+    claimed = Nic.where(id: nics.map(&:id), rekey_coordinator_id: nil)
+      .update(rekey_coordinator_id: private_subnet.id)
+    # :nocov:
+    # Structurally unreachable: the lock check above verifies rekey_coordinator_id is nil
+    # on the same set, within the same transaction, under FOR UPDATE row locks.
+    fail "BUG: locked #{claimed}/#{nics.count} NICs" unless claimed == nics.count
+    # :nocov:
+
+    bad_phase = nics.reject { |n| n.rekey_phase == "idle" }
+    fail "BUG: freshly locked NICs should all be idle: #{bad_phase.map { "#{it.id}=#{it.rekey_phase}" }}" if bad_phase.any?
```
jeremyevans, line 158:
```suggestion
    fail "BUG: freshly locked NICs should all be idle: #{bad_phase.map { "#{it.id}=#{it.rekey_phase}" }.join(", ")}" if bad_phase.any?
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -55,65 +62,116 @@ def gen_reqid
...
 
-    nics = nics_to_rekey
-    nap 10 if nics.any?(&:lock_set?)
-    locked_nics = []
     nics.each do |nic|
-      nic.update(encryption_key: gen_encryption_key, rekey_payload: {spi4: gen_spi, spi6: gen_spi, reqid: gen_reqid})
+      nic.update(encryption_key: gen_encryption_key,
+        rekey_payload: {spi4: gen_spi, spi6: gen_spi, reqid: gen_reqid})
       nic.incr_start_rekey
-      nic.incr_lock
```
jeremyevans, line 49: Not a big deal, but the `lock` semaphore was removed prior to this commit.

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
+  # TLA AdvanceInbound(s) ==
+  # TLA   ∧ pc[s] = "phase_inbound"
   label def wait_inbound_setup
-    nics = get_locked_nics
-    if nics.all? { |nic| nic.strand.label == "wait_rekey_outbound_trigger" }
+    nics = locked_nics
+    # TLA   ∧ heldLocks[s] ≠ {}
+    abort_rekey_if_no_nics(nics)
+    bad = nics.reject { |n| %w[idle inbound].include?(n.rekey_phase) }
+    fail "BUG: phase monotonicity at phase_inbound: #{bad.map { "#{it.id}=#{it.rekey_phase}" }}" if bad.any?
```
jeremyevans, line 196:
```suggestion
    fail "BUG: phase monotonicity at phase_inbound: #{bad.map { "#{it.id}=#{it.rekey_phase}" }.join(", ")}" if bad.any?
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
+  # TLA AdvanceOutbound(s) ==
+  # TLA   ∧ pc[s] = "phase_outbound"
   label def wait_outbound_setup
-    nics = get_locked_nics
-    if nics.all? { |nic| nic.strand.label == "wait_rekey_old_state_drop_trigger" }
+    nics = locked_nics
+    # TLA   ∧ heldLocks[s] ≠ {}
+    abort_rekey_if_no_nics(nics)
+    bad = nics.reject { |n| %w[inbound outbound].include?(n.rekey_phase) }
+    fail "BUG: phase monotonicity at phase_outbound: #{bad.map { "#{it.id}=#{it.rekey_phase}" }}" if bad.any?
```
jeremyevans, line 221:
```suggestion
    fail "BUG: phase monotonicity at phase_outbound: #{bad.map { "#{it.id}=#{it.rekey_phase}" }.join(", ")}" if bad.any?
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
-    nics = get_locked_nics
-    if nics.all? { |nic| nic.strand.label == "wait_rekey_outbound_trigger" }
+    nics = locked_nics
+    # TLA   ∧ heldLocks[s] ≠ {}
+    abort_rekey_if_no_nics(nics)
+    bad = nics.reject { |n| %w[idle inbound].include?(n.rekey_phase) }
+    fail "BUG: phase monotonicity at phase_inbound: #{bad.map { "#{it.id}=#{it.rekey_phase}" }}" if bad.any?
+    # TLA   ∧ ∀ n ∈ heldLocks[s] : nicPhase[n] = "inbound"
+    if nics.all? { |nic| nic.rekey_phase == "inbound" }
       nics.each(&:incr_trigger_outbound_update)
```
jeremyevans, line 199: Fix preexisting N+1 query issue:

```suggestion
      Semaphore.incr(nics.map(&:id), "trigger_outbound_update")
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
-    nics = get_locked_nics
-    if nics.all? { |nic| nic.strand.label == "wait_rekey_old_state_drop_trigger" }
+    nics = locked_nics
+    # TLA   ∧ heldLocks[s] ≠ {}
+    abort_rekey_if_no_nics(nics)
+    bad = nics.reject { |n| %w[inbound outbound].include?(n.rekey_phase) }
+    fail "BUG: phase monotonicity at phase_outbound: #{bad.map { "#{it.id}=#{it.rekey_phase}" }}" if bad.any?
+    # TLA   ∧ ∀ n ∈ heldLocks[s] : nicPhase[n] = "outbound"
+    if nics.all? { |nic| nic.rekey_phase == "outbound" }
       nics.each(&:incr_old_state_drop_trigger)
```
jeremyevans, line 224: Fix preexisting N+1 query issue:

```suggestion
      Semaphore.incr(nics.map(&:id), "old_state_drop_trigger")
```

prog/vnet/metal/subnet_nexus.rb
```diff
@@ -38,56 +85,197 @@ def gen_reqid
...
-    if nics.all? { |nic| nic.strand.label == "wait" }
-      private_subnet.update(state: "waiting", last_rekey_at: Time.now)
-      get_locked_nics_dataset.update(encryption_key: nil, rekey_payload: nil)
-      Semaphore.where(strand_id: nics.map(&:id), name: "lock").delete(force: true)
-      update_stack_locked_nics(nil)
+    nics = locked_nics
+    # TLA   ∧ heldLocks[s] ≠ {}
+    abort_rekey_if_no_nics(nics)
+    bad = nics.reject { |n| %w[outbound old_drop].include?(n.rekey_phase) }
+    fail "BUG: phase monotonicity at phase_old_drop: #{bad.map { "#{it.id}=#{it.rekey_phase}" }}" if bad.any?
```
jeremyevans, line 246:
```suggestion
    fail "BUG: phase monotonicity at phase_old_drop: #{bad.map { "#{it.id}=#{it.rekey_phase}" }.join(", ")}" if bad.any?
```

spec/prog/vnet/metal/nic_nexus_spec.rb
```diff
@@ -100,50 +106,107 @@
...
+      nic.update(rekey_coordinator_id: nil)
+      nic_strand = Strand.create_with_id(nic, prog: "Vnet::Metal::NicNexus", label: "start_rekey")
+      nic.incr_start_rekey
+      expect { nic_strand.unsynchronized_run }.to raise_error Strand::RunError
+    end
+
+    it "fails if NIC not locked when advancing to inbound" do
+      nic.update(rekey_coordinator_id: nil)
+      expect(nx).to receive(:retval).and_return({"msg" => "inbound_setup is complete"})
+      expect(nx).to receive(:decr_start_rekey)
```
jeremyevans, line 122: We should avoid unnecessary mocking in new specs. However, considering the spec already appears to use unnecessary mocking, I guess this is allowable and can be cleaned up later.

spec/prog/vnet/metal/subnet_nexus_spec.rb
```diff
@@ -115,81 +170,192 @@
...
     end
+
+    it "consumes nic_phase_done semaphore when barrier not yet met" do
+      nic
+      strand = Strand.create(prog: "Vnet::Metal::SubnetNexus", label: "wait_outbound_setup", id: ps.id)
+      PrivateSubnet.incr_nic_phase_done(ps.id)
+      expect(Semaphore.where(strand_id: ps.id, name: "nic_phase_done").count).to eq(1)
+      fresh_nx = described_class.new(strand)
+      expect { fresh_nx.wait_outbound_setup }.to hibernate
+      expect(Semaphore.where(strand_id: ps.id, name: "nic_phase_done").count).to eq(0)
```
jeremyevans, line 319:
```suggestion
      fresh_nx = described_class.new(strand)
      expect { fresh_nx.wait_outbound_setup }.to hibernate
        .and change { Semaphore.where(strand_id: ps.id, name: "nic_phase_done").count }.from(1).to(0)
```

spec/spec_helper.rb
```diff
@@ -243,6 +243,23 @@ def default_prog(new_prog)
...
 
+  RSpec::Matchers.define :hibernate do
+    supports_block_expectations
+
+    match do |block|
+      block.call
+      false
+    rescue Prog::Base::Nap => nap
+      @nap = nap
+      nap.seconds >= 60 * 60 * 24 * 365
```
jeremyevans, line 254: Is there a reason not to use an exact match here?

```suggestion
      nap.seconds == 60 * 60 * 24 * 365 * 1000
```

spec/spec_helper.rb
```diff
@@ -243,6 +243,23 @@ def default_prog(new_prog)
     end
   end
 
+  RSpec::Matchers.define :hibernate do
```
jeremyevans, line 246: You might be able to replace this with:

```
def hibernate
  nap(60 * 60 * 24 * 365 * 1000)
end
```

jeremyevans (commented):
OK, finished reviewing this. In general, I think the TLA code would be better in separate files rather than interspersed in the main file using comments, but I'm not strongly against this approach.

Many comments are minor and can be ignored if you don't care much about the implementation.

fdr:
@jeremyevans I think we need to make this easier to read rather than separating the TLA. For example, perhaps reducing the interleaving to a top-of-method case and/or using code folding (e.g. `{{{`) If the TLA is not co-located with the source, it's exceedingly hard to check that the proof and the code are aligned. The correlation is important. It also means LLMs won't be quite as stupid at messing around with the concurrency in there, *I hope*. Definitely its intermediate text was much higher quality in that regard when I was reading them.

proof/spec/proof_extract_spec.rb
fdr: that is odd.

furkansahin:
@fdr , I'll review this after you address @jeremyevans 's review. It seems like this can potentially change a lot. Is there a way to move the nic rekey fix separately before the TLA discussion is resolved?

PR #5096, opened 2026-03-27
Inbound PrivateLink support for subnets (and Postgres)

JoshDreamland:
This PR implements PrivateLink support for subnets, and for our team's use case, adds an API call to the Postgres nexus to initialize (and subsequently maintain) that PrivateLink connection for Postgres usage.

### Notes
* **AWS Resource Chain:** Creating a PrivateLink resource also creates an NLB. The NLB is NOT a separate resource; it is intrinsically bound to the PrivateLink, created and destroyed alongside it. (The NLB can't be shared across PrivateLink resources and currently serves no other utility in Ubicloud.)
* **Dual routes for PL:** This changeset assumes that Subnet:PL is 1:1, so you can access the PrivateLink resource through both the `/privatelink` path on the Subnet, or by a separate `/privatelink/<id>` path on the Location.
* **Failover integration:** To emphasize, the PG Nexus change includes not only a convenience call to create and configure PrivateLink for a Postgres instance, but also logic to maintain this link during failover via `update_targets` to point to the new `representative_server`'s VM **private** IP.
* **AWS-Only:** Attempts to create on non-AWS subnets will return 422.

### Screenshots
| Page View | Creating | Available | Deleting |
| :--- | :--- | :--- | :--- |
| **Subnet Page** <br> *(Networking Tab)* | n1-creating | n2-available | *(No visual difference from the "Available" state)* |
| **PrivateLink Resource Page** | pl1-creating | pl2-available | pl3-deleting |


JoshDreamland:

model/privatelink_aws_resource.rb
```diff
@@ -0,0 +1,126 @@
...
+    DB.transaction do
+      # Create VM association
+      pl_vm = PrivatelinkAwsVm.create(
+        privatelink_aws_resource_id: id,
+        vm_id: vm.id
+      )
+
+      # Create vm_port associations for all ports
+      ports.each do |port|
+        PrivatelinkAwsVmPort.create(
```
serprex: should be able to use `add_` methods to create things with right associations without having to pass so much `_id: id` around

model/privatelink_aws_resource.rb
```diff
@@ -0,0 +1,126 @@
...
+    end
+  end
+
+  def remove_vm(vm)
+    DB.transaction do
+      # Mark all vm_ports as deregistering
+      pl_vm = privatelink_aws_vms_dataset[vm_id: vm.id]
+      return unless pl_vm
+
+      pl_vm.vm_ports_dataset.update(state: "deregistering")
```
serprex: would expect this to go through semaphores which update state, avoids races

model/privatelink_aws_resource.rb
```diff
@@ -0,0 +1,126 @@
...
+      port
+    end
+  end
+
+  def remove_port(port)
+    DB.transaction do
+      # Mark all vm_ports for this port as deregistering
+      DB[:privatelink_aws_vm_port]
+        .where(privatelink_aws_port_id: port.id)
+        .update(state: "deregistering")
```
serprex: You can use `PrivatelinkAwsVmPort.where(...)` & again, should use `incr_deregister`

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+      raise CloverError.new(409, "InvalidRequest", "PrivateLink already exists for this subnet")
+    end
+
+    DB.transaction do
+      pl = PrivatelinkAwsResource.create(
+        private_subnet_id: private_subnet_id,
+        description: description || "AWS PrivateLink endpoint for subnet #{ps.name}"
+      )
+
+      # Create ports
```
serprex:
```suggestion
```
comments could use some cleaning up

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+  def before_run
+    when_destroy_set? do
+      unless ["destroy", "wait_nlb_deletion"].include?(strand.label)
+        hop_destroy
+      end
+    end
+  end
+
+  label def start
+    # Validate we have at least one port
```
serprex:
```suggestion
```

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    if privatelink_aws_resource.ports.empty?
+      fail "PrivateLink must have at least one port"
+    end
+
+    # Get subnet info from any VM in the subnet or from the private subnet's resources
+    # We need at least one NIC in the subnet to get the AWS subnet ID
```
serprex: why doesn't `private_subnet.nics.first` always work?

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+      private_subnet.nics.first
+    end
+
+    unless nic&.nic_aws_resource
+      fail "Cannot create PrivateLink: subnet has no NICs with AWS resources. Create a VM in this subnet first."
+    end
+
+    subnet_id = nic.nic_aws_resource.subnet_id
+
+    # Create Network Load Balancer
```
serprex:
```suggestion
```

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    nlb_response = elb_client.create_load_balancer(
+      name: "pl-#{privatelink_aws_resource.ubid.to_s[..16]}",
```
serprex: why truncate? 19 seems like weird limit

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+      type: "network",
+      scheme: "internal",
+      ip_address_type: "ipv4",
+      subnets: [subnet_id],
+      tags: [
+        {key: "resource_id", value: privatelink_aws_resource.id.to_s},
```
serprex: should tag with ubid

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    privatelink_aws_resource.update(nlb_arn: nlb_arn)
+
+    hop_wait_nlb_active
+  end
+
+  label def wait_nlb_active
+    nlb_desc = elb_client.describe_load_balancers(
+      load_balancer_arns: [privatelink_aws_resource.nlb_arn]
+    )
+    state = nlb_desc.load_balancers.first.state.code
```
serprex:
```suggestion
    nlb_state = nlb_desc.load_balancers.first.state.code
```

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+
+    hop_register_initial_targets
+  end
+
+  label def register_initial_targets
+    # Register all VMs to all target groups
+    eager_vm_ports(state: "registering").each do |vm_port|
+      vm = vm_port.vm
+      nic = privatelink_aws_resource.get_vm_nic(vm)
+      next unless nic
```
serprex:
```suggestion
      next unless (nic = privatelink_aws_resource.get_vm_nic(vm))
```

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    privatelink_aws_resource.update(
+      service_name: service_name,
+      service_id: service_id
+    )
+
+    hop_wait
+  end
+
+  label def wait
+    # Handle semaphores for dynamic operations
```
serprex:
```suggestion
```

model/privatelink_aws_resource.rb
serprex: ah, I got a bit mixed up, since in past iterations `label` was named `state`. this should be okay since `incr_update_targets` at end

still a race with `update(state: "registered")`

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    nic = if privatelink_aws_resource.vms.first
+      # Get NIC from a registered VM
+      privatelink_aws_resource.get_vm_nic(privatelink_aws_resource.vms.first)
+    else
+      # Get any NIC from the private subnet
+      private_subnet.nics.first
+    end
+
+    unless nic&.nic_aws_resource
+      fail "Cannot create PrivateLink: subnet has no NICs with AWS resources. Create a VM in this subnet first."
```
serprex: this might be expected when openapi creates DB & immediately creates PL with it

seems like we should be emitting log & napping

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+      ip_address_type: "ipv4",
+      subnets: [subnet_id],
+      tags: [
+        {key: "resource_id", value: privatelink_aws_resource.id.to_s},
+        {key: "subnet_id", value: private_subnet.id.to_s}
+      ]
+    )
+
+    nlb_arn = nlb_response.load_balancers.first.load_balancer_arn
+    privatelink_aws_resource.update(nlb_arn: nlb_arn)
```
serprex:
```suggestion
    privatelink_aws_resource.update(nlb_arn:)
```
use `bundle exec rubocop -a`

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    )
+    state = nlb_desc.load_balancers.first.state.code
+
+    if state == "active"
+      hop_create_target_groups_and_listeners
+    else
+      puts "NLB is not active yet, waiting..."
+      puts "State: #{state}"
+      puts "NLB ARN: #{privatelink_aws_resource.nlb_arn}"
+      puts "NLB Description: #{nlb_desc.load_balancers.first.inspect}"
```
serprex: use `Clog`

can activating nlb terminate in failed state? guess in that case we wait for `incr_destroy`

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+      nap 5
+    end
+  end
+
+  label def create_target_groups_and_listeners
+    # Get VPC ID from the private subnet
```
serprex:
```suggestion
```

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+      private_ipv4 = nic.private_ipv4.network
+      port = vm_port.privatelink_aws_port
+
+      begin
+        elb_client.deregister_targets(
+          target_group_arn: port.target_group_arn,
+          targets: [{id: private_ipv4.to_s}]
+        )
+      rescue Aws::ElasticLoadBalancingV2::Errors::TargetNotRegisteredException, Aws::ElasticLoadBalancingV2::Errors::TargetGroupNotFound
+        # Target already deregistered or target group deleted, that's fine
```
serprex: must say `nil` in empty `rescue` blocks

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+        default_actions: [
+          {
+            type: "forward",
+            target_group_arn: target_group_arn
+          }
+        ]
+      )
+
+      listener_arn = listener_response.listeners.first.listener_arn
+      port.update(listener_arn: listener_arn)
```
serprex: can group updates into single update

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    hop_wait
+  end
+
+  label def remove_port
+    decr_remove_port
+
+    # Find ports marked for deletion (have deregistering vm_ports)
+    ports_to_remove = privatelink_aws_resource.ports.select { |port|
+      port.vm_ports_dataset.where(state: "deregistering").count > 0
+    }
```
serprex: can use `privatelink_aws_resource.ports_dataset` to query this as 1 query

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+
+    ports_to_remove.each do |port|
+      # Deregister all targets first
+      port.vm_ports_dataset.each do |vm_port|
```
serprex:
```suggestion
      port.vm_ports(eager: :vm).each do |vm_port|
```
think it's supposed to be something like this

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    when_add_port_set? do
+      hop_add_port
+    end
+
+    when_remove_port_set? do
+      hop_remove_port
+    end
+
+    when_add_vm_set? do
+      hop_add_vm
```
serprex: this & remove hop to state that just does decr & then hops to update, seems like that could be inlined here

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+        begin
+          elb_client.delete_listener(listener_arn: port.listener_arn)
+        rescue Aws::ElasticLoadBalancingV2::Errors::ListenerNotFound
+          # Already deleted, that's fine
+        end
+      end
+
+      # Deregister all targets from this target group
+      if port.target_group_arn
+        port.vm_ports.each do |vm_port|
```
serprex:
```suggestion
        port.vm_ports(eager: :vm).each do |vm_port|
```

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    # Delete Load Balancer
+    if privatelink_aws_resource.nlb_arn
+      begin
+        elb_client.delete_load_balancer(load_balancer_arn: privatelink_aws_resource.nlb_arn)
+      rescue Aws::ElasticLoadBalancingV2::Errors::LoadBalancerNotFound
+        # Already deleted, that's fine
+      rescue Aws::ElasticLoadBalancingV2::Errors::ResourceInUse
+        # NLB is still associated with VPC Endpoint Service, wait and retry
+        nap 10
+        return
```
serprex:
```suggestion
```

serprex:
does this handle having 2 different private networks that both want to access database? what does someone do with the service endpoint once they've enabled this feature?

model/privatelink_aws_resource.rb
JoshDreamland: Done.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Fix applied (here/throughout).

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Fix applied.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: If we're concerned with NIC turnup time for new clusters with PL at creation time, we'll want to check for the first non-nill subnet NIC. But yes. Done.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Done.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: But then how will people know this code was written by an LLM?

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Because 19 = 16 (a great number) + 3 (another great number that follows more incidentally from this circumstance). Thank you for asking.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Now uses UBIDs for both.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Renamed.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Clog employed. I think in that case we instead loop forever, retrying the strand indefinitely.

(This actually isn't that big a deal, because you can either fix what's causing it to not come up, or just delete the PrivateLink resource from the UI, which is worthless without that NLB anyway.)

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: This one was a load-bearing comment, actually.

Removed... watch for falling rocks.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Applied (and massively deduplicated thanks to the dual-nexus reflow).

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Obsoleted by new VM nexus.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Added (but logic extracted)

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Consolidated.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Refactored. Claude says the new structure (`where(where().select()).each()`) is one query, which is a complete mind-fuck for me, but the bot says "Sequel is magic" and I'm inclined to believe it...

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: You will regret this!

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Our eager fetching ended up looking different; PTAL

model/privatelink_aws_resource.rb
JoshDreamland: Resolved using a new PL VM Nexus.

model/privatelink_aws_resource.rb
JoshDreamland: Restructured.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: (Here too)

JoshDreamland:
> does this handle having 2 different private networks that both want to access database? what does someone do with the service endpoint once they've enabled this feature?

Yes. The user just copies the "Service Name" out of the table in the networking tab (or out of our Control Plane) and into their AWS PrivateLink endpoint in one or more of their VPCs.

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,362 @@
...
+    hop_wait
+  end
+
+  label def remove_port
+    decr_remove_port
+
+    privatelink_aws_resource.ports_dataset
+      .where(id: PrivatelinkAwsVmPort.where(state: "deregistering").select(:privatelink_aws_port_id))
+      .each do |port|
+        port.vm_ports_dataset.eager(privatelink_aws_vm: {vm: :nics}).all.each do |vm_port|
```
serprex: these 2 queries should be able to be merged somehow (eager on outer query, replace `vm_ports_dataset` with `vm_ports`)

model/privatelink_aws_resource.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  def remove_vm(vm)
+    pl_vm = privatelink_aws_vms_dataset[vm_id: vm.id]
+    return unless pl_vm
+
+    if pl_vm.strand
+      pl_vm.incr_destroy
+    else
+      pl_vm.destroy
+    end
```
fdr: this is an odd pattern, can you expand on it?

ubid.rb
```diff
@@ -101,16 +101,17 @@ class UBID
   TYPE_LOCATION_CREDENTIAL = "1c"
   TYPE_PRIVATE_SUBNET_AWS_RESOURCE = "aw"
   TYPE_NIC_AWS_RESOURCE = "na"
+  TYPE_PRIVATELINK_AWS_RESOURCE = "p1"
+  TYPE_PRIVATELINK_AWS_PORT = "pp"
+  TYPE_PRIVATELINK_AWS_VM = "pw"
+  TYPE_PRIVATELINK_AWS_VM_PORT = "pq"
```
fdr, line 107: I'm really regretting using only two bytes for type tag. @furkansahin should we consider trying, to some extent, to group under a single-byte prefix, like "a" for aws? I know it conflicts with "address" as a prefix, but it's an approximate art, after all. It's just "p" conflicts with "postgres" by the same token.

fdr: oh bummer: azure. we might have to choose something vaguely unnatural at points.

JoshDreamland: What's our actual limiting factor on the size?

model/privatelink_aws_resource.rb
JoshDreamland: Glad you raised this; it just inlines the delete where possible, but AFAIK there's no reason we couldn't always let the VM Nexus handle this. Is that your preference?

ubid.rb
fdr: clean-sheet: birthday paradox, since it's embedded into the UUID itself.
real-sheet: that I defined it as two bytes (enough for anyone?!??), so I guess we'd have to make sure new prefixes are disjoint from anything we have generated. 

May be worth it? @pykello was the original designer of the bits and bytes here, but the prefix length was at my direction. Historically, I used _one_ byte, I figured two is luxurious, but in this project we proliferate many, many more record types. probably, TYPE_ETC was also a mistake, initially, it was intended as a sink for many more types.

fdr: if you felt like doing some bit-probing, write a ubid patch so we could define longer prefixes not incompatible with ones already allocated! check the birthday paradox math though.

model/privatelink_aws_resource.rb
JoshDreamland: I see what happened here. I had a shaky understanding of Ubicloud's model and was worried about a case that didn't exist. I think it's flat out safer to always signal here and let the appropriate nexus handle it. In my defense, I was operating from a database in which this resource existed without a corresponding strand to govern it, but that's a nonsensical state for a healthy ubicloud system to enter now. I've pushed a correction.

ubid.rb
JoshDreamland: Do you mean picking new three- or four-letter prefixes for all these such that they only overlap with a two-letter prefix already allocated unless it's the same resource? It's not going to be "satisfying" but it's possible. Problems:

PG means Postgres, so sub-typing that resource means conflits
AC means Account and conflicts with ten other nice resource prefixes...

But we can still do "okay" that way:

```
0p -> OIDP (OIDC_PROVIDER)
10 -> LOCN (LOCATION)
1b -> LDBN (LOAD_BALANCER)
1c -> LCCR (LOCATION_CREDENTIAL)
1e -> IFEP (INFERENCE_ENDPOINT)
1p -> LDBP (LOAD_BALANCER_PORT)
1q -> LDVP (LOAD_BALANCER_VM_PORT)
1r -> IFER (INFERENCE_ENDPOINT_REPLICA)
1s -> LBVM (LOAD_BALANCER_VM)
1v -> INVC (INVOICE)
ac -> ACCT (ACCOUNT)
ad -> ADDR (ADDRESS)
ah -> AHST (ASSIGNED_HOST_ADDRESS)
ak -> APIK (API_KEY)
av -> AVMA (ASSIGNED_VM_ADDRESS)
aw -> AWSB (PRIVATE_SUBNET_AWS_RESOURCE)
b1 -> BILN (BILLING_INFO)
br -> BILR (BILLING_RECORD)
ce -> CERT (CERT)
dc -> DISC (DISCOUNT_CODE)
dr -> DNSR (DNS_RECORD)
ds -> DNSS (DNS_SERVER)
dz -> DNSZ (DNS_ZONE)
fr -> FRWR (FIREWALL_RULE)
fw -> FWAL (FIREWALL)
g1 -> GHIN (GITHUB_INSTALLATION)
ge -> GHCE (GITHUB_CACHE_ENTRY)
gp -> GHRP (GITHUB_REPOSITORY)
gr -> GHAR (GITHUB_RUNNER)
kb -> KUEB (KUBERNETES_ETCD_BACKUP)
kc -> KUCL (KUBERNETES_CLUSTER)
kd -> KUND (KUBERNETES_NODE)
ke -> KUSK (STORAGE_KEY_ENCRYPTION_KEY)
kn -> KUNP (KUBERNETES_NODEPOOL)
mc -> MNIO (MINIO_CLUSTER)
md -> PMET (POSTGRES_METRIC_DESTINATION)
mp -> MINP (MINIO_POOL)
ms -> MINS (MINIO_SERVER)
na -> NICW (NIC_AWS_RESOURCE)
nc -> NICR (NIC)
pa -> PAGE (PAGE)
pc -> AXPC (ACCESS_POLICY)
pd -> PDSC (PROJECT_DISCOUNT_CODE)
pg -> PGDB (POSTGRES_RESOURCE)
pj -> PJCT (PROJECT)
pm -> PYMM (PAYMENT_METHOD)
ps -> PSUB (PRIVATE_SUBNET)
pt -> PTLG (POSTGRES_TIMELINE)  (prefer PGTL)
pv -> PVMG (POSTGRES_SERVER)    (prefer PGVM)
r0 -> IFRO (INFERENCE_ROUTER)
rm -> IFRM (INFERENCE_ROUTER_MODEL)
rr -> IFRR (INFERENCE_ROUTER_REPLICA)
rt -> IFRT (INFERENCE_ROUTER_TARGET)
sh -> SSHA (SSHABLE)
sk -> SSHK (SSH_PUBLIC_KEY)
sm -> SMPH (SEMAPHORE)
st -> STRD (STRAND)
t0 -> OBJT (OBJECT_TAG)
t2 -> OBJM (OBJECT_METATAG)
ta -> TAGA (ACTION_TAG)
te -> AXCE (ACCESS_CONTROL_ENTRY)
tg -> TAGX (ACCESS_TAG)
tn -> IPTN (IPSEC_TUNNEL)
ts -> TAGS (SUBJECT_TAG)
tt -> ATYP (ACTION_TYPE)
v1 -> VSVL (VM_STORAGE_VOLUME)
va -> UGSA (USAGE_ALERT)
vh -> VHST (VM_HOST)
vm -> VMRC (VM)
vn -> VCMS (VICTORIA_METRICS_SERVER)
vp -> VPOO (VM_POOL)   (prefer VMPL) 
vr -> VCMR (VICTORIA_METRICS_RESOURCE)
vs -> VSLC (VM_HOST_SLICE)
```

migrate/20251218_privatelink_subnet_level.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
```
furkansahin, line 1: In the commit, another Furkan is mentioned :) I don't know who Furkan Doğrul is

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+    when_destroy_set? do
+      unless ["destroy", "wait_nlb_deletion"].include?(strand.label)
+        hop_destroy
+      end
+    end
+  end
+
+  label def start
+    # Validate we have at least one port
+    if privatelink_aws_resource.ports.empty?
```
furkansahin: we should make this verification in `assemble` instead of failing the process here

prog/vnet/privatelink_aws_nexus.rb
furkansahin: I feel like we could `Clog.emit` and `nap` here until there is a resource instead of fast fail.

prog/vnet/privatelink_aws_nexus.rb
furkansahin: We had cases in other AWS code where the entity is created but the process died before recording it in the database with `.update`. Can you first `get` the load balancer and if that request fails, rescue and create it?

prog/vnet/privatelink_aws_nexus.rb
furkansahin: We should remove these

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+        default_actions: [
+          {
+            type: "forward",
+            target_group_arn: target_group_arn
+          }
+        ]
+      )
+
+      listener_arn = listener_response.listeners.first.listener_arn
+      port.update(listener_arn: listener_arn)
```
furkansahin: let's split this into its own label

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+        healthy_threshold_count: 3,
+        unhealthy_threshold_count: 3,
+        tags: [
+          {key: "resource_id", value: privatelink_aws_resource.id.to_s},
+          {key: "port_id", value: port.id.to_s}
+        ]
+      )
+
+      target_group_arn = tg_response.target_groups.first.target_group_arn
+      port.update(target_group_arn: target_group_arn)
```
furkansahin: same as above, can we first `get` this entity?

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+        elb_client.register_targets(
+          target_group_arn: port.target_group_arn,
+          targets: [
+            {id: private_ipv4.to_s, port: port.dst_port}
+          ]
+        )
+      rescue Aws::ElasticLoadBalancingV2::Errors::TargetGroupNotFound
+        # Target group deleted, skip registration
+        next
+      end
```
furkansahin: it seems like this part of the code is very similar to what we have in `register_initial_targets`. Can you move this into its own method and DRY. We could even maybe do the same for `deregister`ing

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+            ]
+          )
+        rescue Aws::ElasticLoadBalancingV2::Errors::TargetGroupNotFound
+          # Target group deleted, skip registration
+          next
+        end
+
+        vm_port.update(state: "registered")
+      end
+    end
```
furkansahin, line 196: this is doing a lot in a single label, we should probably split.

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,583 @@
...
+        # Already deleted, that's fine
+      rescue Aws::ElasticLoadBalancingV2::Errors::ResourceInUse
+        # NLB is still associated with VPC Endpoint Service, wait and retry
+        nap 10
+        return
+      end
+    end
+
+    hop_wait_nlb_deletion
+  end
```
furkansahin: again, this is a very big method. Can we split it?

migrate/20251218_privatelink_subnet_level.rb
JoshDreamland: HAHAHAHAHA... I let Claude handle the rebase. I do not know where it got that name either. I guess any old Furkan will do in a pinch. Tell you what—I'll rebase it again after we get through this review so that I don't screw up the review comments. Or we can squash if you prefer.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Good call. Done. I was skeptical at first because of how many destroy states this gives us, but I've prefixed them so the check is simple. I can alternatively publish the set of destroy states in the nexus (or give it a routine to return a nice name for its state).

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Also done.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Stale comment; serp had me replace them with Clog. That acceptable, or do you prefer deletion?

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: @serprex caught the same thing; already done post-nexus-split.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Good call; [done](https://github.com/ubicloud/ubicloud/pull/5096/changes/a3acf0a8db56e4d8896c5faa2b79d0edca9f600e).

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: [Also done](https://github.com/ubicloud/ubicloud/pull/5096/changes/a3acf0a8db56e4d8896c5faa2b79d0edca9f600e).

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: We're doing one per port, so I don't see a huge difference between creating `{port}` resources in one label vs creating `2×{port}` resources (particularly now that we have idempotent helpers). I do not feel strongly, though, if you'd prefer two states with two loops over our ports.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: DRY'd a bunch of shit in the original refactor and a bunch more now.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: [Done](https://github.com/ubicloud/ubicloud/pull/5096/changes/33888ddaf27d151467ef2cc0ab661e4adfdde6b6).

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: [Discussed offline] Split.

routes/project/location/private_subnet.rb
```diff
@@ -30,6 +30,67 @@ class Clover
...
       end
 
+      r.on "privatelink" do
+        r.post true do
+          authorize("PrivateSubnet:edit", ps)
+
+          vm_ids = []
+          if (vm = authorized_vm(location_id: ps.location_id))
+            vm_ids = [vm.id]
+          end
```
furkansahin, line 40: Is this the right thing to do here? If I'm not mistaken, we're simply attaching it to the very first VM according to the database return order in this subnet. What if we have 2 ha setup and this returns one of the standbys?

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -166,6 +166,11 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
 
       nap 5
     end
+    if (pl = private_subnet.privatelink_aws_resource)
+      pl.incr_destroy
```
furkansahin, line 171: please add `register_deadline`

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,376 @@
...
+        name: "pl-#{privatelink_aws_resource.ubid}",
+        type: "network",
+        scheme: "internal",
+        ip_address_type: "ipv4",
+        subnets: [nic.nic_aws_resource.subnet_id],
+        tags: [
+          {key: "ubid", value: privatelink_aws_resource.ubid.to_s},
+          {key: "subnet_ubid", value: private_subnet.ubid.to_s}
+        ]
+      ).load_balancers.first.load_balancer_arn
```
furkansahin: it's almost guaranteed that in the initial run, we won't have the load balancer. Let's swap the order of the operation here.

prog/vnet/privatelink_aws_nexus.rb
```diff
@@ -0,0 +1,376 @@
...
+    end
+
+    begin
+      elb_client.describe_load_balancers(
+        load_balancer_arns: [privatelink_aws_resource.nlb_arn]
+      )
+      nap 10
+    rescue Aws::ElasticLoadBalancingV2::Errors::LoadBalancerNotFound
+      privatelink_aws_resource.destroy
+      pop "PrivateLink deleted"
```
furkansahin: these 2 lines seem to be repeating, let's move them to a method and DRY things up

prog/vnet/privatelink_aws_vm_nexus.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    privatelink_aws_vm.destroy
+    pop "PrivatelinkAwsVm destroyed"
+  end
+
+  private
+
+  def register_registering_vm_ports
+    vm = privatelink_aws_vm.vm
+    return unless vm
```
furkansahin, line 67: is this a valid scenario? I am expecting this strand to have a VM if it is created in the first place.

prog/vnet/privatelink_aws_vm_nexus.rb
```diff
@@ -0,0 +1,108 @@
...
+      next unless port.target_group_arn
+
+      begin
+        elb_client.register_targets(
+          target_group_arn: port.target_group_arn,
+          targets: [{id: private_ipv4.to_s, port: port.dst_port}]
+        )
+      rescue Aws::ElasticLoadBalancingV2::Errors::TargetGroupNotFound
+        next
+      end
```
furkansahin, line 83: if the target group doesn't exist, we're silently skipping the registration. What is the case where the target group doesn't exist? I feel like in the initial run, there is a random chance where we skip this because the PrivatelinkAwsNexus did not create the target group yet?

routes/project/location/private_subnet.rb
JoshDreamland: The `authorized_object` helper extracts the ID from the URL parameters. That said, we're not using this route for Postgres. If you'd like, I can simply delete it.

prog/vnet/privatelink_aws_vm_nexus.rb
JoshDreamland: It would unless the user immediately deleted it. I ran a few tests locally where I deliberately deleted stuff at inopportune times to exercise these transitions. I'll need to do this over again after our changes are done, I believe.

prog/vnet/privatelink_aws_vm_nexus.rb
JoshDreamland: If one of these doesn't exist, it just means the port isn't ready. We can ignore that as we should handle it next time someone calls `incr_add_port`, e.g. on lines [132](https://github.com/ubicloud/ubicloud/pull/5096/changes#diff-cb9f8853c8f4b7910131d946ba238a612457334d58c11dbaffb90ca7b670cbc9R132) and [195](https://github.com/ubicloud/ubicloud/pull/5096/changes#diff-cb9f8853c8f4b7910131d946ba238a612457334d58c11dbaffb90ca7b670cbc9R195).

That said, I agree this distribution is confusing. I think it's an artefact of the switch to a multiple-nexus design. Rather than inlining this logic in `start`, I've had it fire `incr_add_port` on the VM strand as soon as it's created. I think this will make the race a little more likely to run out of order, but makes the logic less confusing in return. Let me know if you preferred it the other way around. Additionally, I changed it so that the "add_port" label will nap and retry rather than advancing to wait if any ports are still registering. Sound good?

prog/vnet/aws/vpc_nexus.rb
JoshDreamland: Ten minutes okay? I really don't have a good intuition for a slow AWS operation.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Assuming you mean "try creation and catch duplicate," done. In fact, two of our three other resources (Target group and Listener) can also work this way. Only Endpoint Services cannot. I've changed the pattern there, too.

prog/vnet/privatelink_aws_nexus.rb
JoshDreamland: Sure; done.

prog/vnet/aws/vpc_nexus.rb
serprex: 10m is okay, bit hard to measure a good time since you want something sufficiently beyond P99 that we don't get paged for every 100th PL setup

my understanding is this timeline includes time to provision an NLB https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1834 suggests you want 5m just for that

views/privatelink/show.erb
```diff
@@ -0,0 +1,38 @@
+<div class="p-6">
```
byucesoy, line 1: Is this supposed to be top level resource or should it be part of private subnet? In the screenshots, there is no menu on the left side like other resources we have, so I didn't quite understand where this is in the hierarchy.

migrate/20251218_privatelink_subnet_level.rb
```diff
@@ -0,0 +1,57 @@
...
+
+Sequel.migration do
+  change do
+    create_table(:privatelink_aws_resource) do
+      column :id, :uuid, primary_key: true
+      column :created_at, :timestamptz, null: false, default: Sequel.lit("now()")
+      column :updated_at, :timestamptz, null: false, default: Sequel.lit("now()")
```
byucesoy, line 8: Do we really want to store this? At one point we added updated_at columns to PG models, but then dropped them.

prog/vnet/aws/vpc_nexus.rb
byucesoy: Why do we want to allow extending deadlines here?

byucesoy: We also have `register_deadline(nil, 10 * 60)` call below. We can just move this addition below that line.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -824,16 +824,19 @@ def before_run
 
   label def taking_over
     if postgres_server.read_replica?
+      old_vm = resource.representative_server.vm
```
byucesoy, line 827: no need to save `resource.representative_server` in a variable. while in this label, `resource.representative_server` would always point to same object.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -905,6 +909,13 @@ def version
     postgres_server.version
   end
 
+  def swap_privatelink_target(old_vm, new_vm)
+    old_vm.privatelink_aws_resources.each do |pl|
```
byucesoy, line 913: Can one server have more than one private links given that there is 1:1 mapping between subnet and private link?

routes/project/location/private_subnet.rb
furkansahin: Neither the path nor the payload has vm_id set, though? Let's remove it.

prog/vnet/privatelink_aws_vm_nexus.rb
furkansahin: nap and retry is good

prog/postgres/postgres_server_nexus.rb
furkansahin: There can be multiple private links per subnet. The relationship should be defined that way.

ubid.rb
fdr: yeah, but...probably we have to just punt on this for a while, I discussed it with @jeremyevans and there's a version bit we can steal. @furkansahin do you have opinions on how to arrange the type bytes to make it as reasonable looking as possible, although perfect reasonableness is probably impossible.

model/privatelink_aws_resource.rb
fdr: yeah, in general, if a semaphore can do the job, you always want to synchronize through it.

JoshDreamland:
Two other points not covered in this PR:

1. Customer's account ID is not passed to the endpoint creation API, and it doesn't allow-all by default, meaning customers can't just set this up with one click. Ideally we'd add a separate API route for managing the allowed account IDs of endpoint users (or allowing all).
2. If you try to delete a resource in Ubicloud that the customer is still pointing to, it'll just fail on loop... customer has to unregister PL ***\*on their end\**** before Ubicloud can delete the resources on its end.

PR #5098, opened 2026-03-29
Add VNC console support for QEMU VMs

bufanoc:
## Summary

- Enable VNC access for VMs running under the QEMU hypervisor by exposing a unix socket at `/vm/<name>/vnc.sock` instead of running headless with `-display none`
- Add `vnc.sock` to the VM path registry in `vm_path.rb`

## Motivation

Currently, QEMU VMs run with `-display none -vga none`, making it impossible to interact with the VM console. This is a problem for:

- **Debugging boot failures** — when a VM fails before SSH comes up, there's no way to see what's happening
- **Non-Linux guest OS images** — VMware ESXi, FreeBSD, Windows, and other operating systems that don't support cloud-init or SSH provisioning need graphical console access for initial setup
- **Self-hosted operators** — anyone running Ubicloud on-prem benefits from console access for troubleshooting

## Changes

| File | Change |
|------|--------|
| `rhizome/host/lib/vm_setup.rb` | Replace `-display none -vga none` with `-vnc unix:<socket> -vga std` in `build_qemu_service` |
| `rhizome/host/lib/vm_path.rb` | Add `vnc.sock` to VM path registry |

## Usage

After a QEMU VM boots, operators can bridge the VNC unix socket to TCP for remote access:

```bash
socat TCP-LISTEN:5900,reuseaddr,fork UNIX-CONNECT:/vm/<vmname>/vnc.sock
```

Then connect with any VNC client (TightVNC, UltraVNC, etc.).

## Test plan

- [ ] Verify QEMU VM starts with VNC socket at `/vm/<name>/vnc.sock`
- [ ] Verify VNC connection works via socat TCP bridge
- [ ] Verify no regression for Cloud Hypervisor VMs (unaffected, separate code path)
- [ ] Verify `-vga std` provides usable graphical output

Tested on a self-hosted Ubicloud deployment with ESXi 6.7 and Ubuntu Noble guests.

### DIFF
--- a/rhizome/host/lib/vm_path.rb
+++ b/rhizome/host/lib/vm_path.rb
@@ -52,6 +52,7 @@ class VmPath
     cloudinit.img
     ch-api.sock
     serial.log
+    vnc.sock
     hugepages
     public_ipv4
     nftables_conf
--- a/rhizome/host/lib/vm_setup.rb
+++ b/rhizome/host/lib/vm_setup.rb
@@ -843,8 +843,8 @@ DNSMASQ_SERVICE
 
     serial_parts = [
       "-serial file:#{vp.serial_log}",
-      "-display none",
-      "-vga none"
+      "-vnc unix:#{vp.vnc_sock}",
+      "-vga std"
     ]
 
     kernel_parts = [


bufanoc:

bufanoc:
recheck

bufanoc:
recheck

bsatzger:
Thank you for this PR! Today, we can only observe VMs via serial.log or rely on SSH access, which we don't have for customer VMs. This change is helpful for debugging and potentially recovering VMs when SSH is broken/misconfigured. However, enabling VNC support by default also introduces:

1. Deviation between hypervisors: QEMU gets an interactive console via VNC. Cloud Hypervisor remains log-only.
2. Expanded attack surface
3. Increases the provider's access to customer VMs

For these reasons, I'd tend to not add VNC access by default but make it optional, and keep it disabled by default. Would that work for your use case?

PR #5105, opened 2026-03-30
Optimize AWS runner firewall rule calls

enescakir:
- **Skip duplicate SSH ingress call in AWS VPC setup**
  When firewall rules already include port 22 from 0.0.0.0/0,
  the explicit SSH allow_ingress call is redundant and causes
  an unnecessary authorize_security_group_ingress API call.
  
  

- **Apply IPv6 firewall rules in AWS VPC setup**
  Previously, IPv6 firewall rules were skipped in wait_vpc_created.
  Now both IPv4 and IPv6 rules are applied, and SSH ingress is
  added for both protocols unless already covered by firewall rules.
  
  

- **Skip firewall rules update for runner VMs in wait_sshable**
  Runner VMs don't need firewall rule updates. Use allocated_at to
  determine the initial nap duration so the optimization is preserved
  for all VMs.

### DIFF
--- a/prog/vm/aws/nexus.rb
+++ b/prog/vm/aws/nexus.rb
@@ -241,13 +241,12 @@ class Prog::Vm::Aws::Nexus < Prog::Base
   end
 
   label def wait_sshable
-    unless vm.update_firewall_rules_set?
-      vm.incr_update_firewall_rules
-      # This is the first time we get into this state and we know that
-      # wait_sshable will take definitely more than 6 seconds. So, we nap here
-      # to reduce the amount of load on the control plane unnecessarily.
-      nap 6
-    end
+    vm.incr_update_firewall_rules unless is_runner? || vm.update_firewall_rules_set?
+
+    # On the first visit, wait_sshable will take definitely more than 6 seconds.
+    # Nap the remaining time to reduce unnecessary load on the control plane.
+    remaining = vm.allocated_at - Time.now + 6
+    nap remaining if remaining > 0
     addr = vm.ip4
     hop_create_billing_record unless addr
 
--- a/prog/vnet/aws/vpc_nexus.rb
+++ b/prog/vnet/aws/vpc_nexus.rb
@@ -43,16 +43,27 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
 
     private_subnet_aws_resource.update(security_group_id: security_group_response.group_id)
 
-    ssh_covered = false
+    ssh_covered_ip4 = false
+    ssh_covered_ip6 = false
     private_subnet.firewalls(eager: :firewall_rules).flat_map(&:firewall_rules).each do |firewall_rule|
-      next if firewall_rule.ip6?
-      ssh_covered = true if firewall_rule.port_range.first == 22 && firewall_rule.port_range.last - 1 == 22 && firewall_rule.cidr.to_s == "0.0.0.0/0"
-      allow_ingress(security_group_response.group_id, firewall_rule.port_range.first, firewall_rule.port_range.last - 1, firewall_rule.cidr.to_s)
+      first_port = firewall_rule.port_range.first
+      last_port = firewall_rule.port_range.last - 1
+      cidr = firewall_rule.cidr.to_s
+      if 22.between?(first_port, last_port)
+        case cidr
+        when "0.0.0.0/0"
+          ssh_covered_ip4 = true
+        when "::/0"
+          ssh_covered_ip6 = true
+        end
+      end
+      allow_ingress(security_group_response.group_id, first_port, last_port, cidr, firewall_rule.ip6?)
     end
 
     # Allow SSH ingress from the internet so that the controlplane can verify
     # that the VM is running.
-    allow_ingress(security_group_response.group_id, 22, 22, "0.0.0.0/0") unless ssh_covered
+    allow_ingress(security_group_response.group_id, 22, 22, "0.0.0.0/0", false) unless ssh_covered_ip4
+    allow_ingress(security_group_response.group_id, 22, 22, "::/0", true) unless ssh_covered_ip6
     hop_create_route_table
 
   end
@@ -253,16 +264,18 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
     @client ||= location.location_credential.client
   end
 
-  def allow_ingress(group_id, from_port, to_port, cidr)
-    client.authorize_security_group_ingress({
-      group_id:,
-      ip_permissions: [{
-        ip_protocol: "tcp",
-        from_port:,
-        to_port:,
-        ip_ranges: [{cidr_ip: cidr}],
-      }],
-    })
+  def allow_ingress(group_id, from_port, to_port, cidr, ip6)
+    perm = {
+      ip_protocol: "tcp",
+      from_port:,
+      to_port:,
+    }
+    if ip6
+      perm[:ipv_6_ranges] = [{cidr_ipv_6: cidr}]
+    else
+      perm[:ip_ranges] = [{cidr_ip: cidr}]
+    end
+    client.authorize_security_group_ingress({group_id:, ip_permissions: [perm]})
   rescue Aws::EC2::Errors::InvalidPermissionDuplicate
   end
 
--- a/prog/vnet/aws/vpc_nexus.rb
+++ b/prog/vnet/aws/vpc_nexus.rb
@@ -277,6 +277,7 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
     end
     client.authorize_security_group_ingress({group_id:, ip_permissions: [perm]})
   rescue Aws::EC2::Errors::InvalidPermissionDuplicate
+    nil
   end
 
   def private_subnet_aws_resource
--- a/prog/vnet/aws/vpc_nexus.rb
+++ b/prog/vnet/aws/vpc_nexus.rb
@@ -43,14 +43,16 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
 
     private_subnet_aws_resource.update(security_group_id: security_group_response.group_id)
 
+    ssh_covered = false
     private_subnet.firewalls(eager: :firewall_rules).flat_map(&:firewall_rules).each do |firewall_rule|
       next if firewall_rule.ip6?
+      ssh_covered = true if firewall_rule.port_range.first == 22 && firewall_rule.port_range.last - 1 == 22 && firewall_rule.cidr.to_s == "0.0.0.0/0"
       allow_ingress(security_group_response.group_id, firewall_rule.port_range.first, firewall_rule.port_range.last - 1, firewall_rule.cidr.to_s)
     end
 
     # Allow SSH ingress from the internet so that the controlplane can verify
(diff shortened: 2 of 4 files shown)


enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23743038261

enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23744255107

enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23749264911

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -43,14 +43,20 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
...
+    ssh_covered_ip4 = false
+    ssh_covered_ip6 = false
     private_subnet.firewalls(eager: :firewall_rules).flat_map(&:firewall_rules).each do |firewall_rule|
-      next if firewall_rule.ip6?
-      allow_ingress(security_group_response.group_id, firewall_rule.port_range.first, firewall_rule.port_range.last - 1, firewall_rule.cidr.to_s)
+      if firewall_rule.port_range.first == 22 && firewall_rule.port_range.last - 1 == 22
+        ssh_covered_ip4 = true if firewall_rule.cidr.to_s == "0.0.0.0/0"
+        ssh_covered_ip6 = true if firewall_rule.cidr.to_s == "::/0"
+      end
+      allow_ingress(security_group_response.group_id, firewall_rule.port_range.first, firewall_rule.port_range.last - 1, firewall_rule.cidr.to_s, firewall_rule.ip6?)
```
jeremyevans: I'd like to DRY this up a bit using local variables:

```suggestion
      first_port = firewall_rule.port_range.first
      last_port = firewall_rule.port_range.last - 1
      cidr = firewall_rule.cidr.to_s
      if first_port == 22 && last_port == 22
        case cidr
        when "0.0.0.0/0"
          ssh_covered_ip4 = true
        when "::/0"
          ssh_covered_ip6 = true
        end
      end
      allow_ingress(security_group_response.group_id, first_port, last_port, cidr, firewall_rule.ip6?)
```

Note that `firewall_rule.port_range.last - 1` makes me a bit uneasy as it only works if the port range uses an exclusive end. It works in this case because this is a direct database retrieval, and PostgreSQL will convert integer ranges with inclusive ends to exclusive ends. We would still probably be better off adding `FirewallRule#end_port` and move some logic from `FirewallRule#display_port_range` into it. However, that could be done later. I think we make the same assumption in other places.

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -251,16 +257,18 @@ def client
...
+      to_port:
+    }
+    if ip6
+      perm[:ipv_6_ranges] = [{cidr_ipv_6: cidr}]
+    else
+      perm[:ip_ranges] = [{cidr_ip: cidr}]
+    end
+    client.authorize_security_group_ingress({group_id:, ip_permissions: [perm]})
   rescue Aws::EC2::Errors::InvalidPermissionDuplicate
   end
```
jeremyevans, line 281: Not an issue in this PR, but we use explicit nils in rescue blocks to ensure that the rescue block is covered:

```suggestion
    nil
  end
```

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -251,16 +257,18 @@ def client
...
+      ip_protocol: "tcp",
+      from_port:,
+      to_port:
+    }
+    if ip6
+      perm[:ipv_6_ranges] = [{cidr_ipv_6: cidr}]
+    else
+      perm[:ip_ranges] = [{cidr_ip: cidr}]
+    end
+    client.authorize_security_group_ingress({group_id:, ip_permissions: [perm]})
```
jeremyevans, line 278: Not specific to this PR, but if `ip_permissions` will accept multiple firewall rules, why do we make separate calls to `authorize_security_group_ingress` per rule instead of a single call to `authorize_security_group_ingress` with all rules?

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -251,16 +257,18 @@ def client
...
-      ip_permissions: [{
-        ip_protocol: "tcp",
-        from_port:,
-        to_port:,
-        ip_ranges: [{cidr_ip: cidr}]
-      }]
-    })
+  def allow_ingress(group_id, from_port, to_port, cidr, ip6)
+    perm = {
+      ip_protocol: "tcp",
```
jeremyevans, line 269: Not related to this PR, but we support non-tcp rules. Shouldn't the protocol be taken from `FirewallRule#protocol` instead of hard coded?

prog/vm/aws/nexus.rb
```diff
@@ -241,13 +241,12 @@ def before_destroy
...
-      # This is the first time we get into this state and we know that
-      # wait_sshable will take definitely more than 6 seconds. So, we nap here
-      # to reduce the amount of load on the control plane unnecessarily.
-      nap 6
-    end
+    vm.incr_update_firewall_rules unless is_runner? || vm.update_firewall_rules_set?
+
+    # On the first visit, wait_sshable will take definitely more than 6 seconds.
+    # Nap the remaining time to reduce unnecessary load on the control plane.
+    remaining = vm.allocated_at + 6 - Time.now
```
jeremyevans: Avoid unnecessary `Time` allocation.

```suggestion
    remaining = vm.allocated_at - Time.now + 6
```

spec/prog/vm/aws/nexus_spec.rb
```diff
@@ -664,26 +664,36 @@ def set_alternative_families(families)
...
-      expect { nx.wait_sshable }.to nap(6)
+    it "naps and increments update_firewall_rules on first run" do
+      vm.update(allocated_at: Time.now)
+      expect { nx.wait_sshable }.to nap
         .and change { vm.reload.update_firewall_rules_set? }.from(false).to(true)
     end
 
+    it "does not increment update_firewall_rules for runner vms" do
+      vm.update(allocated_at: Time.now)
+      expect(nx).to receive(:is_runner?).and_return(true)
```
jeremyevans: We avoid unneccesary mocking in new specs:

```suggestion
      vm.update(allocated_at: Time.now, unix_user: "runneradmin")
```

prog/vnet/aws/vpc_nexus.rb
```diff
@@ -43,14 +43,16 @@ class Prog::Vnet::Aws::VpcNexus < Prog::Base
 
     private_subnet_aws_resource.update(security_group_id: security_group_response.group_id)
 
+    ssh_covered = false
     private_subnet.firewalls(eager: :firewall_rules).flat_map(&:firewall_rules).each do |firewall_rule|
       next if firewall_rule.ip6?
+      ssh_covered = true if firewall_rule.port_range.first == 22 && firewall_rule.port_range.last - 1 == 22 && firewall_rule.cidr.to_s == "0.0.0.0/0"
```
furkansahin:
```suggest
      ssh_covered = true if firewall_rule.port_range.first <= 22 && firewall_rule.port_range.last - 1 >= 22 && firewall_rule.cidr.to_s == "0.0.0.0/0"
```

prog/vnet/aws/vpc_nexus.rb
furkansahin: This function here doesn't get firewall rule, it is introduced purely to handle the very first firewall addition, which is for tcp port 22.

prog/vnet/aws/vpc_nexus.rb
furkansahin: If there is a conflict, something wrong with one of the rules in that case, the whole update fails.

furkansahin (approved):

prog/vnet/aws/vpc_nexus.rb
enescakir: @furkansahin If the main purpose of this logic is port 22, we already allow public port 22 anyway without checking the firewall. We also already apply firewall rules after the VM is provisioned. What do you think about removing the `private_subnet.firewalls` part from here to simplify the logic and just allow port 22 from public? 

https://github.com/ubicloud/ubicloud/blob/85211b174500d01f4d208b70377df61e9aa59408/prog/vnet/aws/vpc_nexus.rb#L46-L54

furkansahin: No, today it is only port 22 but tomorrow we might want to add other rules. Also, the customer may have firewall rules in the subnet that they set, until we apply the VM specific rules all together, we would keep the VM without those applied. It would introduce a small window of time where their firewall rules are not enforced.

enescakir:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/23808529391

prog/vnet/aws/vpc_nexus.rb
jeremyevans: This is because we expect it to fail in at least some cases with `Aws::EC2::Errors::InvalidPermissionDuplicate`? Still seems a bit strange to me. My assumption would be we should know the existing rules and only add the ones that are new and shouldn't be duplicate. However, I can understand that being a larger change.

prog/vnet/aws/vpc_nexus.rb
jeremyevans: > This function here doesn't get firewall rule, it is introduced purely to handle the very first firewall addition, which is for tcp port 22.

The `allow_ingress` method is called with with every firewall rule, though, both before and with this PR. When you say "function" here, are you referring to something else (since it doesn't make sense if you are referring to the `allow_ingress` method)?

jeremyevans (approved):

prog/vnet/aws/vpc_nexus.rb
furkansahin: You're right, `allow_ingress` "method" is the only "method" we use. But the VM UpdateFirewallRules `prog` does not use that method, instead, makes a direct api call. I was also confused to make the separation of subnet rules / vm rules here, which lead to this long thread. You're right, we should get the protocol from the DB, that is the right thing. 

In practice, though, it doesn't change anything because we only get port number and source ip from the customer when specifying the firewall rules and we write only "tcp" as the protocol (in prod database, we do not have a single rule that is not "tcp"). I remember adding it for future proofing but those days did not come in my expected timeframe.

prog/vnet/aws/vpc_nexus.rb
furkansahin: I feel like that is the theme in the whole codebase, no? Reconfigurations, DNS record rewrites, etc. We never assume something is there, if we're performing a task, we assume the worst and rewrite the full dataset. In this case, too, we do not perform bookkeeping for AWS but we try to rewrite it in full whenever we have the chance. Maybe we could do something to ease it here though. If we first fetched the rules and filtered our the ones we already had, and then applied in batch, we would only make 2 API calls.

furkansahin (approved):

enescakir:
code path changed a lot

PR #5122, opened 2026-04-01
Add new Postgres instance families and normalize c/m/r sizing

heavycrystal:
Add c6id, c7gd, c8gd, c8id, m7gd, m8id, r7gd, r8id instance families
for PostgreSQL on AWS. These fill gaps in the fallback chain for
InsufficientInstanceCapacity errors across arm64 and x64.

Add missing 48 vCPU tier to c6gd, m6gd, m6id, m8gd, i8g. Cap
all c/m/r families at 64 vCPU. Beyond that, i-series offers more
storage at lower cost per TB. Fix i7i.12xlarge which was present in
storage and postgres options but missing from VmSizes.

Refactor VmSizes construction into a data-driven AWS_FAMILY_VM_CONFIG
hash, replacing the chain of concat/product blocks with a single
flat_map.


furkansahin (approved):
please rebase

spec/routes/api/cli/golden-files/help -r.txt
```diff
@@ -674,22 +674,33 @@ Allowed Option Values:
...
           hobby-1 hobby-2 c6gd.large c6gd.xlarge c6gd.2xlarge c6gd.4xlarge
-          c6gd.8xlarge c6gd.16xlarge m6id.large m6id.xlarge m6id.2xlarge
-          m6id.4xlarge m6id.8xlarge m6id.16xlarge m6gd.large m6gd.xlarge
-          m6gd.2xlarge m6gd.4xlarge m6gd.8xlarge m6gd.16xlarge m8gd.large
-          m8gd.xlarge m8gd.2xlarge m8gd.4xlarge m8gd.8xlarge m8gd.16xlarge
-          i8g.large i8g.xlarge i8g.2xlarge i8g.4xlarge i8g.8xlarge i8g.16xlarge
-          i8g.24xlarge i8ge.large i8ge.xlarge i8ge.2xlarge i8ge.3xlarge
-          i8ge.6xlarge i8ge.12xlarge i8ge.18xlarge i8ge.24xlarge i7i.large
-          i7i.xlarge i7i.2xlarge i7i.4xlarge i7i.8xlarge i7i.12xlarge
+          c6gd.8xlarge c6gd.12xlarge c6gd.16xlarge c6id.large c6id.xlarge
```
serprex, line 677: @jeremyevans you previously raised concerns about this list getting out of hand, might be time to address this

jeremyevans: I added #4891 to support context sensitive help. That support just needs to be used to filter the list. As mentioned in that PR, we need an API endpoint that exposes the allowed sizes, since the CLI can only use the API, it is not allowed to access the database directly.

lib/option.rb
```diff
@@ -108,21 +116,28 @@ def display_name
     storage_size_options = [it * 10, it * 20]
     io_limits = IoLimits.new(it * 50, it * 50)
     VmSize.new("burstable-#{it}", "burstable", it, it * 50, it * 50, (it * 1.6).to_i, storage_size_options, io_limits, 1, false, "arm64")
-  }).concat(AWS_FAMILY_OPTIONS.reject { |f| ["i8g", "i8ge", "i7i", "i7ie", "r8gd", "r6gd", "r6id"].include?(f) }.product([2, 4, 8, 16, 32, 64]).map { |family, vcpu|
-    memory_coefficient = (family == "c6gd") ? 2 : 4
-    arch = ["m6a", "m6id", "m7a", "m7i"].include?(family) ? "x64" : "arm64"
-    VmSize.new(aws_instance_type_name(family, vcpu), family, vcpu, vcpu * 100, 0, vcpu * memory_coefficient, AWS_STORAGE_SIZE_OPTIONS[family][vcpu], NO_IO_LIMITS, nil, false, arch)
+  }).concat(AWS_FAMILY_OPTIONS.reject { |f| ["c6gd", "c6id", "c7gd", "c8gd", "c8id", "i8g", "i8ge", "i7i", "i7ie", "m6gd", "m6id", "m7gd", "m8gd", "m8id", "r6gd", "r6id", "r7gd", "r8gd", "r8id"].include?(f) }.product([2, 4, 8, 16, 32, 64]).map { |family, vcpu|
```
byucesoy: Can we reorder the families we rejected here in the order of they appearing below?

lib/option.rb
```diff
@@ -108,21 +116,28 @@ def display_name
...
+    arch = ["m6a", "m7a", "m7i"].include?(family) ? "x64" : "arm64"
+    VmSize.new(aws_instance_type_name(family, vcpu), family, vcpu, vcpu * 100, 0, vcpu * 4, AWS_STORAGE_SIZE_OPTIONS[family][vcpu], NO_IO_LIMITS, nil, false, arch)
+  }).concat(["c6gd", "c7gd", "c8gd"].product([2, 4, 8, 16, 32, 48, 64]).map { |family, vcpu|
```
byucesoy: Since the list is quite big now, I think this approach of iterating families here doesn't work and causes unreadable code. Here is a suggestion:

Let's add family to cpu mapping, similar to `AWS_STORAGE_SIZE_OPTIONS` and use it here to DRY some of these.

lib/option.rb
heavycrystal: superseded by the refactor in below comment

lib/option.rb
heavycrystal: added family to `[cpu, mem_ratio and arch]` mapping for a single concat block

lib/option.rb
```diff
@@ -40,25 +40,70 @@ def self.vring_workers(vcpus)
     [1, vcpus / 2].max
   end
 
-  AWS_FAMILY_OPTIONS = ["c6gd", "m6a", "m6id", "m6gd", "m7a", "m7i", "m7g", "m8g", "m8gd", "i8g", "i8ge", "i7i", "i7ie", "r8gd", "r6gd", "r6id"].freeze
+  local_storage_vcpus = [2, 4, 8, 16, 32, 48, 64]
```
byucesoy, line 43: i and ie series are also local storage, so naming is bit misleading. Also `local_storage_with_medium_vcpus` is confusing. What does medium vcpu mean? I think we need better names here. In general I'm OK with long names where necessarry, if this helps with finding better names.

heavycrystal: I meant `medium_vpus` for the families that support `.medium` instance size. This is only for some `r` classes with ARM CPUs.

How about: `generic_local_storage_vcpus` `i_family_local_storage_vcpus` `ie_family_local_storage_vcpus` and `rgd_family_local_storage_vcpus`?

byucesoy (approved):

PR #5145, opened 2026-04-02
Postgres Log Destinations

shikharbhardwaj:
image

image

## Third party provider integration tests

- [x] Mezmo
- [x] OpenObserve
- [x] Honeybadger


rhizome/postgres/lib/otel_log_config.rb
```diff
@@ -0,0 +1,238 @@
...
+            directory: /tmp/otelcol
+            on_start: true
+            on_rebound: true
+            rebound_needed_threshold_mib: 1024
+            rebound_trigger_threshold_mib: 100
+
+      receivers:
+        filelog/pglog:
+          include:
+            - "#{@log_dir}/postgresql-*.json"
```
iamKunalGupta: I haven't visited the upgrade code path, so I am not sure if configure in case something breaks will cause no logs to come up
Additionally, was thinking we should probably move away from %a to prevent rewrites

There was a recent fix upsream on copytruncate too https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/24169d9857dff7adb66ce4a017440a07e68ec7ed

model/postgres/postgres_log_destination.rb
jeremyevans, line 1: Model files must always be in a commit after the commit that adds the migration. Otherwise, if you deploy this commit, it will fail until the related migration is run.

The migration commit should only contain the migration, cache updates, and any annotation updates caused by the migration.

migrate/20260401_add_postgres_log_destination.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:postgres_log_destination) do
+      column :id, :uuid, primary_key: true
```
jeremyevans: We add a ubid uuid default for primary keys in new tables:

```suggestion
      # UBID.to_base32_n("p1") => 705
      column :id, :uuid, primary_key: true, default: Sequel.function(:gen_random_ubid_uuid, 705)`
```

model/postgres/postgres_log_destination.rb
```diff
@@ -0,0 +1,35 @@
...
+    val ? JSON.parse(val) : nil
+  end
+
+  def structured_data=(val)
+    super(if val.nil?
+            nil
+          else
+            val.is_a?(String) ? val : val.to_json
+          end)
+  end
```
jeremyevans: Instead of this, we should probably be passing `format: :json` when setting the column as an encrypted column. That will take a small change to `ResourceMethods.configure`.

rhizome/postgres/bin/configure-logs
```diff
@@ -0,0 +1,58 @@
...
+  server_role: server_role,
+  log_dir: log_dir,
+  log_destinations: destinations,
+)
+
+r "mkdir -p /etc/otelcol-contrib /var/lib/otelcol-contrib/state /tmp/otelcol"
+otelcol_user = r("systemctl show otelcol-contrib -p User --value").strip
+unless otelcol_user.empty?
+  r "chown -R #{otelcol_user}: /var/lib/otelcol-contrib/state /etc/otelcol-contrib /tmp/otelcol"
+  r "usermod -aG systemd-journal #{otelcol_user}"
```
jeremyevans: Avoid interpolation where possible in `r` command lines. You can pass an array of arguments so this does not get executed via the shell:

```suggestion
  r "chown", "-R", otelcol_user, "/var/lib/otelcol-contrib/state", "/etc/otelcol-contrib /tmp/otelcol"
  r "usermod", "-aG", "systemd-journal", otelcol_user
```

I'm not sure if the `:` in the `chown` command was needed. I removed it in the suggestion.

Similar changes needed in the rest of the file.

rhizome/postgres/bin/configure-logs
```diff
@@ -0,0 +1,58 @@
...
+end
+
+unless otelcol_user.empty?
```
jeremyevans: Isn't the same conditional as the previous one?

```suggestion

```

rhizome/postgres/bin/configure-logs
```diff
@@ -0,0 +1,58 @@
...
+  # Grant traverse on /dat and each intermediate directory down to log_dir
+  r "setfacl -m u:#{otelcol_user}:x /dat"
```
jeremyevans: Do we have to use ACLs here? Will normal unix permissions not work for some reason?

rhizome/postgres/lib/otel_log_config.rb
```diff
@@ -0,0 +1,238 @@
...
+      processors:
+        batch:
+      #{indent(transform_processors_yaml, 2)}
+      exporters:
+      #{indent(syslog_exporters_yaml, 2)}
+      service:
+        extensions: [health_check, file_storage/state]
+        pipelines:
+      #{indent(pipelines_yaml, 4)}
+    YAML
```
jeremyevans: We've been burnt badly on this type of interpolating YAML into YAML approach previously. Instead of this, create a normal hash with the data you want, and then use `YAML.dump(hash)`.

If the program consuming this is not YAML 1.2 compliant, we may need to do some post-processing of the resulting YAML, but if it is YAML 1.2 compliant, we shouldn't have to do anything special.

rhizome/postgres/spec/otel_log_config_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+  let(:log_dir) { "/dat/17/data/pg_log" }
+  let(:log_destinations) { [] }
+  let(:config) { described_class.new(instance: instance, server_role: server_role, log_dir: log_dir, log_destinations: log_destinations) }
+
+  describe "#to_config" do
+    subject(:yaml) { config.to_config }
+
+    it "includes health_check and file_storage extensions" do
+      expect(yaml).to include("health_check:")
+      expect(yaml).to include("file_storage/state:")
```
jeremyevans: We should not be using expectations on the serialized YAML. We should be parsing the YAML and have expectations on the resulting data structure.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -780,6 +799,7 @@ def before_run
       resource.representative_server.update(is_representative: false)
       postgres_server.reload.update(is_representative: true, synchronization_status: "ready")
       resource.servers.each(&:incr_configure_metrics)
+      resource.servers.each(&:incr_configure_logs)
```
jeremyevans: I realize this is just copying the existing implementation, but each of these is an N+1 query issue. To fix:

```suggestion
      server_ids = resource.servers.map(&:id)
      Semaphore.incr(server_ids, "configure_metrics")
      Semaphore.incr(server_ids, "configure_logs")
```

Similar issue below.

spec/routes/api/project/location/postgres_spec.rb
jeremyevans, line 1: This seems to be included in the wrong commit. The commit adding the route is the following commit, not this commit.

openapi/openapi.yml
```diff
@@ -1389,6 +1389,66 @@ paths:
...
+                  description: Name for the Log Destination
+                  type: string
+                host:
+                  description: Hostname of the syslog server
+                  type: string
+                port:
+                  description: Port of the syslog server
+                  type: integer
+                structured_data:
+                  description: 'RFC 5424 structured data as a hash of SD-ID to parameter hashes (e.g. for Honeybadger, LogDNA)'
```
jeremyevans: LogDNA was renamed to Mezmo.

```suggestion
                  description: 'RFC 5424 structured data as a hash of SD-ID to parameter hashes (e.g. for Honeybadger, Mezmo)'
```

openapi/openapi.yml
```diff
@@ -2402,6 +2462,15 @@ components:
...
         example: eu-central-h1
+    log_destination_id:
+      description: Postgres Log Destination ID
+      in: path
+      name: log_destination_id
+      required: true
+      schema:
+        type: string
+        example: p1000000000000000000000000
+        pattern: '^(et|p1)[0-9a-hj-km-np-tv-z]{24}$'
```
jeremyevans: I don't think we should allow `et` ubids for this, since the table was just added.

```suggestion
        pattern: '^p1[0-9a-hj-km-np-tv-z]{24}$'
```

routes/project/location/postgres.rb
```diff
@@ -292,6 +292,49 @@ class Clover
...
+          structured_data = typecast_params.Hash("structured_data")
+          if structured_data
+            unless structured_data.values.all? { |v| v.is_a?(Hash) && v.values.all?(String) }
+              fail CloverError.new(400, "InvalidRequest", "structured_data must be a hash of string-to-string hashes")
+            end
+          end
+
+          DB.transaction do
+            ld = PostgresLogDestination.create(postgres_resource_id: pg.id, name:, host:, port:, structured_data:)
+            pg.servers.each(&:incr_configure_logs)
```
jeremyevans: Similar N+1 issue.

routes/project/location/postgres.rb
```diff
@@ -292,6 +292,49 @@ class Clover
...
+          Serializers::Postgres.serialize(pg, {detailed: true})
+        end
+
+        r.delete :ubid_uuid do |id|
+          authorize("Postgres:edit", pg)
+
+          if (ld = pg.log_destinations_dataset[id:])
+            DB.transaction do
+              ld.destroy
+              pg.servers.each(&:incr_configure_logs)
```
jeremyevans: N+1 query issue.

cli-commands/pg/post/add-log-destination.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("add-log-destination") do
+  desc "Add a PostgreSQL log destination"
+
+  banner "ubi pg (location/pg-name | pg-id) add-log-destination name host port [sd_id/key=value [...]]"
```
jeremyevans: We use `-` instead of `_` in CLI commands:

```suggestion
  banner "ubi pg (location/pg-name | pg-id) add-log-destination name host port [sd-id/key=value [...]]"
```

cli-commands/pg/post/add-log-destination.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  args(3..)
+
+  run do |args, _, cmd|
+    name, host, port, *sd_args = args
+    structured_data = structured_data_args_to_hash(sd_args, cmd)
+    data = sdk_object.add_log_destination(name:, host:, port: port.to_i, structured_data:)
```
jeremyevans: This will convert invalid ports to `0`. That probably raises an error later during the API request, since the minimum value is `1`, but the error will not make sense as the user did not provide `0`. 

```suggestion
    data = sdk_object.add_log_destination(name:, host:, port: need_integer_arg(port, "port", cmd), structured_data:)
```

lib/ubi_cli.rb
```diff
@@ -277,6 +277,22 @@ def config_entries_to_hash(args, cmd)
     end
   end
 
+  def structured_data_args_to_hash(args, cmd)
+    result = {}
+    args.each do |arg|
+      unless arg.include?("/") && arg.include?("=")
```
jeremyevans: I don't think you need the `=` check here. It does result in an earlier failure for a certain type of malformed request, but I don't think it's worth it:

```suggestion
      unless arg.include?("/")
```

sdk/ruby/lib/ubicloud/model/postgres.rb
```diff
@@ -71,6 +71,28 @@ def add_metric_destination(username:, password:, url:)
       md
     end
 
+    # Add a log destination for this database with the given name, host, and port.
+    # Optionally pass structured_data as a hash of SD-ID => {param => value} for
+    # providers that require SD-based authentication (e.g. Honeybadger, LogDNA).
```
jeremyevans:
```suggestion
    # providers that require SD-based authentication (e.g. Honeybadger, Mezmo).
```

spec/routes/api/cli/pg/add-log-destination_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    expect(pg.log_destinations_dataset.count).to eq 1
+    ld = pg.log_destinations.first
+    expect(body).to eq <<~END
+      Log destination added to PostgreSQL database.
+      Current log destinations:
+        1: #{ld.ubid}  my-dest  logs.example.com  6514
+    END
+    expect(ld.name).to eq "my-dest"
+    expect(ld.host).to eq "logs.example.com"
+    expect(ld.port).to eq 6514
```
jeremyevans:
```suggestion
    expect(ld.port).to eq 6514
    expect(ld.structured_data).to be_nil
```

spec/routes/api/cli/pg/delete-log-destination_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  it "deletes the specified log destination for the database" do
+    cli(%w[pg eu-central-h1/test-pg create -s standard-2 -S 64])
+    cli(%w[pg eu-central-h1/test-pg add-log-destination my-dest logs.example.com 6514])
+    pg = PostgresResource.first
+    ld = pg.log_destinations.first
+    expect(cli(%w[pg eu-central-h1/test-pg delete-log-destination a/b], status: 400)).to start_with "! Invalid log destination id format\n"
+    expect(pg.log_destinations_dataset).not_to be_empty
+    expect(cli(%W[pg eu-central-h1/test-pg delete-log-destination #{ld.ubid}])).to eq "Log destination, if it exists, has been scheduled for deletion\n"
+    expect(pg.log_destinations_dataset).to be_empty
```
jeremyevans: Good to test that deletion works if the item has already been deleted:

```suggestion
    2.times do
      expect(cli(%W[pg eu-central-h1/test-pg delete-log-destination #{ld.ubid}])).to eq "Log destination, if it exists, has been scheduled for deletion\n"
      expect(pg.log_destinations_dataset).to be_empty
    end
```

spec/routes/api/cli/pg/show_spec.rb
```diff
@@ -19,78 +19,85 @@
...
-        3: #{rules[2].ubid}  ::/0  5432  
-        4: #{rules[3].ubid}  ::/0  6432  
-      metric-destinations:
-        1: #{@pg.metric_destinations[0].ubid}  md-user  https://md.example.com
-      read-replicas:
-      ca-certificates:
-      a
-      b
-    END
+    expected1 = [
```
jeremyevans: Please keep the expectation using a heredoc as it was before. Same with the following spec.

routes/project/location/postgres.rb
```diff
@@ -295,14 +295,28 @@ class Clover
...
           name, host = typecast_params.nonempty_str!(["name", "host"])
           port = typecast_params.pos_int!("port")
           unless (1..65535).cover?(port)
             fail CloverError.new(400, "InvalidRequest", "port must be between 1 and 65535")
           end
 
-          structured_data = typecast_params.Hash("structured_data")
+          structured_data = if web?
+            sd_json = typecast_params.str("structured_data_json")&.strip
+            (sd_json.nil? || sd_json.empty?) ? nil : begin
```
jeremyevans: I'm not generally one to complain about style, but starting a `begin` in a ternary operator is definitely difficult to read. Switch to `if`/`else`.

jeremyevans: I'm not generally one to complain about style, but starting a `begin` in a ternary operator is definitely difficult to read. Switch to `if`/`else`.

routes/project/location/postgres.rb
```diff
@@ -295,14 +295,28 @@ class Clover
...
 
           name, host = typecast_params.nonempty_str!(["name", "host"])
           port = typecast_params.pos_int!("port")
           unless (1..65535).cover?(port)
             fail CloverError.new(400, "InvalidRequest", "port must be between 1 and 65535")
           end
 
-          structured_data = typecast_params.Hash("structured_data")
+          structured_data = if web?
+            sd_json = typecast_params.str("structured_data_json")&.strip
```
jeremyevans: Is accepting JSON in a textarea really what we want to use for this? It seems quite unfriendly to a web user. I recommend accepting a similar format used for the CLI, instead of forcing the user to submit JSON formatted input.

spec/routes/web/project/postgres_spec.rb
```diff
@@ -893,6 +893,110 @@ def create_minio_cluster_for_blob_storage
...
+      it "can create log destination" do
+        visit "#{project.path}#{pg.path}/logs"
+        fill_in "name", with: "graylog"
+        fill_in "host", with: "logs.example.com"
+        fill_in "port", with: "6514"
+        find(".log-destination-create-button").click
+        expect(page.title).to eq "Ubicloud - pg-with-permission"
+        expect(page).to have_flash_notice "Log destination is created"
+        expect(page).to have_content "logs.example.com"
+        expect(pg.reload.log_destinations.count).to eq(1)
```
jeremyevans: This should check that the columns were set as expected.

spec/routes/web/project/postgres_spec.rb
```diff
@@ -893,6 +893,110 @@ def create_minio_cluster_for_blob_storage
...
+
+      it "treats absent structured_data_json parameter as no structured data" do
+        visit "#{project.path}#{pg.path}/logs"
+        _csrf = all("input[name='_csrf']", visible: false).last.value
+        page.driver.post "#{project.path}#{pg.path}/log-destination",
+          {name: "graylog", host: "logs.example.com", port: "6514", _csrf:}
+        expect(page.driver.response.status).to eq(302)
+        expect(pg.reload.log_destinations.first.structured_data).to be_nil
+      end
+
```
jeremyevans: This spec is not needed for the web. The browser will always submit the field, even if it is empty.

```suggestion
```

spec/routes/web/project/postgres_spec.rb
```diff
@@ -893,6 +893,110 @@ def create_minio_cluster_for_blob_storage
...
+          port: 6514,
+        )
+        visit "#{project.path}#{pg.path}/logs"
+
+        find("#ld-delete-#{ld.ubid} .delete-btn").click
+        expect(page).to have_flash_notice("PostgreSQL log destination deleted.")
+        expect(pg.reload.log_destinations.count).to eq(0)
+      end
+
+      it "cannot delete log destination if it does not exist" do
```
jeremyevans: This spec description should be reworded. It makes it sound like attempting to delete a log destination that does not exist will raise an error, when it is silently ignored.

views/postgres/logs.erb
```diff
@@ -0,0 +1,85 @@
...
+        <% @pg.log_destinations.each do |ld| %>
+          <tr>
+            <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6"><%= ld.name %></td>
+            <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6"><%= ld.host %></td>
+            <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:pl-6"><%= ld.port %></td>
+            <td
+              id="ld-delete-<%= ld.ubid %>"
+              class="relative whitespace-nowrap py-4 pl-3 pr-4 text-right text-sm font-medium sm:pr-6"
+            >
+              <%== part("components/delete_button", url: "#{path(@pg)}/log-destination/#{ld.ubid}", text: "") %>
```
jeremyevans: Should only show delete button if `@edit_perm`. Without `@edit_perm`, you shouldn't show the 4th column.

prog/download_boot_image.rb
```diff
@@ -102,9 +102,8 @@ def url
     ["github-ubuntu-2404", "arm64", "20260401.1.0"] => "2bb2b5d40a7bb8b18f2813073e5e1954ba96c6e389f63cf140db47e2f43b6367",
     ["github-ubuntu-2204", "x64", "20260312.1.0"] => "a50f161933a8ced1018b589f49fa723d2638760366b17fb0112b1e98f1d036c5",
     ["github-ubuntu-2204", "x64", "20260401.1.0"] => "a39e9ac4b7b8511cf30b24965fb14d21a456f930e9fc54c0831d67b24e24744e",
-    ["github-ubuntu-2204", "arm64", "20260312.1.0"] => "4291d401ba6d8c0407408f6f4f7c238693c939bdfb2a2d0723dc64151e751974",
```
byucesoy: Probably got deleted unintentionally?

prog/postgres/postgres_server_nexus.rb
```diff
@@ -130,7 +130,7 @@ def before_run
       # ext4 defaults to reserving 5% of disk for root, on 1TiB this is 50GiB. Cap this to 50GiB
       reserved_blocks_per_gb = 13107 # ~5% of 262144 (number of 4KiB blocks per GB)
       reserve_blocks = [postgres_server.storage_size_gib * reserved_blocks_per_gb, 13107200].min
-      vm.sshable.cmd("sudo tune2fs :path -r :reserve_blocks", path: device_path, reserve_blocks:)
+      vm.sshable.cmd("sudo tune2fs :path -r :reserve_blocks -o acl", path: device_path, reserve_blocks:)
```
byucesoy: Why do we need to enable acl here?

routes/project/location/postgres.rb
byucesoy: Do we just need to accept list of key/values or do we need to accept nested data? If only key/values, we can put 2 text boxes similar to postgres configs.

routes/project/location/postgres.rb
shikharbhardwaj: Applying the `Semaphore.incr` change here throws: `Calling Semaphore.incr directly in Clover is not allowed`

rhizome/postgres/lib/otel_log_config.rb
shikharbhardwaj: Upgrades happen on a candidate server that gets destroyed in case the upgrade fails, and the previous primary is bought back as-is, so I don't expect things to break during upgrades. In other cases, we call `incr_configure_logs` at all places which call `incr_configure_metrics` to make sure the config remains in sync with the server state.

On %a: I think we made the change recently to make weekly log rotation work. Do you mean the same log file name being reused is potentially problematic? Would a full timestamp be better?

routes/project/location/postgres.rb
jeremyevans: That's due to model hiding. We probably don't want `Semaphore.incr` usable directly in routes. My recommendation would be to add a `PostgresResource` instance method does the `Semaphore.incr`, and call that.

routes/project/location/postgres.rb
shikharbhardwaj: The format is nested, for eg: `{"provider@1234": {"api_key": "xyz"}}`. I think for api we can keep it as-is, I made the web form more intuitive. A user typically would only configure one SD-ID, but the interface makes the nesting clear.

prog/postgres/postgres_server_nexus.rb
shikharbhardwaj: The otel user is granted permissions to this folder using `setfacl`. I removed this though, on further reading it turns out this is already available on ext4 by default.

rhizome/postgres/bin/configure-logs
shikharbhardwaj: We could do this via unix groups: either add the otel user to the postgres group or make the default permissions for the log files more permissible. Both grant more permissions than strictly necessary for this, and this approach was already in a downstream patch so I went with the same.

rhizome/postgres/bin/configure-logs
```diff
@@ -0,0 +1,51 @@
...
+log_config = OtelLogConfig.new(
+  instance: instance,
+  server_role: server_role,
+  log_dir: log_dir,
+  log_destinations: log_destinations,
+)
+
+r "mkdir", "-p", "/etc/otelcol-contrib", "/var/lib/otelcol-contrib/state", "/tmp/otelcol"
+otelcol_user = r("systemctl", "show", "otelcol-contrib", "-p", "User", "--value").strip
+unless otelcol_user.empty?
```
byucesoy: 1. Is otelcol_user not static?
2. When it would be empty?

rhizome/postgres/bin/configure-logs
```diff
@@ -0,0 +1,51 @@
...
+  r "chown", "-R", otelcol_user, "/var/lib/otelcol-contrib/state", "/etc/otelcol-contrib", "/tmp/otelcol"
+  r "usermod", "-aG", "systemd-journal", otelcol_user
+
+  # Grant traverse on /dat and each intermediate directory down to log_dir
+  r "setfacl", "-m", "u:#{otelcol_user}:x", "/dat"
```
byucesoy: It would be easier to unroll this loop and hardcode this. We are not working with arbitrary paths anyway.

Maybe we can get version from control plane and not the entire log_dir.

migrate/20260414_add_postgres_log_destination.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
```
jeremyevans, line 3: Commit message for first commit should be updated as it describes things no longer in the commit.

rhizome/postgres/bin/configure-logs
jeremyevans: Looks like we already use ACLs in the storage volume rhizome code, so I guess using them here is also acceptable.

prog/postgres/postgres_server_nexus.rb
```diff
@@ -779,7 +798,9 @@ def before_run
     if postgres_server.read_replica?
       resource.representative_server.update(is_representative: false)
       postgres_server.reload.update(is_representative: true, synchronization_status: "ready")
-      resource.servers.each(&:incr_configure_metrics)
+      server_ids = resource.servers.map(&:id)
+      Semaphore.incr(server_ids, "configure_metrics")
+      Semaphore.incr(server_ids, "configure_logs")
```
jeremyevans: Considering this structure is used in multiple places in this prog, I recommend adding a method to this prog:

```ruby
  def server_incr(*semaphores)
    server_ids = resource.servers.map(&:id)
    semaphores.each do
      Semaphore.incr(server_ids, it)
    end
  end
```

This requires 1 query per semaphore. You could avoid that by having `Semaphore.incr` to accept an array of semaphore names, and insert all of them.

openapi/openapi.yml
```diff
@@ -1389,6 +1389,66 @@ paths:
...
+                  description: Port of the syslog server
+                  type: integer
+                structured_data:
+                  description: 'RFC 5424 structured data as a hash of SD-ID to parameter hashes (e.g. for Honeybadger, Mezmo)'
+                  type: object
+                  additionalProperties:
+                    type: object
+                    additionalProperties:
+                      type: string
+                  nullable: true
```
jeremyevans: I don't think we should allow explicit nil values. If the key is present, the value should be an object. It's not required the key is present, since it isn't listed in `required`.

```suggestion
```

migrate/20260408_add_postgres_log_destination.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    create_table(:postgres_log_destination) do
+      # UBID.to_base32_n("p1") => 705
+      column :id, :uuid, primary_key: true, default: Sequel.function(:gen_random_ubid_uuid, 705)
```
jeremyevans: You changed the ubid type prefix but forgot to update this:

```suggestion
      # UBID.to_base32_n("1d") => 45
      column :id, :uuid, primary_key: true, default: Sequel.function(:gen_random_ubid_uuid, 45)
```

spec/routes/api/cli/pg/show_spec.rb
jeremyevans: Please keep the expectation using a heredoc as it was before. Same with the following spec.

views/postgres/logs.erb
```diff
@@ -0,0 +1,98 @@
+<% post = request.POST
+  if post && !post.empty? && post["structured_data_ids"]
+    sd_ids = post["structured_data_ids"] || []
+    sd_keys = post["structured_data_keys"] || []
+    sd_values = post["structured_data_values"] || []
```
jeremyevans: Use `typecast_params` or `typecast_body_params` instead of directly accessing the submitted parameters. Otherwise, users can submit types you don't expect and result in TypeErrors or NoMethodErrors.

views/postgres/logs.erb
```diff
@@ -0,0 +1,98 @@
...
+    form_structured_data = {}
+    sd_ids.zip(sd_keys, sd_values).each do |id, key, val|
+      next if id.to_s.empty? || key.to_s.empty?
+      (form_structured_data[id] ||= {})[key] = val.to_s
+    end
+    form_structured_data = nil if form_structured_data.empty?
+  else
+    form_structured_data = nil
```
jeremyevans: This is unnecessary, as the local variable will automatically have a `nil` value in this case:

```suggestion
```

views/components/structured_data_card.erb
```diff
@@ -0,0 +1,78 @@
+<%# locals: (structured_data:) %>
+<% input_class = "border border-gray-300 rounded py-1 px-2 text-sm focus:ring-blue-500 focus:border-blue-500 font-mono" %>
+
+<div id="sd-id-groups" class="space-y-2">
+
+  <%# Existing SD-ID groups %>
+  <% (structured_data || {}).each do |sd_id, params| %>
```
jeremyevans:
```suggestion
  <% structured_data&.each do |sd_id, params| %>
```

rhizome/postgres/bin/configure-logs
shikharbhardwaj: Made this static

jeremyevans (commented):
All of my comments have been addressed. I'll leave approval to @furkansahin and @byucesoy.

shikharbhardwaj:
@jeremyevans I decided to expand the scope of this PR to also include another type of log destination, to make sure we cover providers that we've seen from customers. Could you take another pass when you get a chance?

Supporting both types on the UI while keeping the interaction simple became a little more javascript heavy than what I would have liked, open to your ideas/suggestions.

rhizome/postgres/lib/otel_log_config.rb
iamKunalGupta: > Would a full timestamp be better?

Yeah, but I think that begs the question who will clean up the old files?

jeremyevans:
@shikharbhardwaj Sure, I can review tomorrow morning. Since you rebased instead of pushing additional commits, can you please be specific about which commits you would like reviewed? I would prefer not to review the entire PR again.

shikharbhardwaj:
@jeremyevans The CLI and UI changes mainly (last 2 commits)

rhizome/postgres/lib/otel_log_config.rb
shikharbhardwaj: I think the current setup should be safe, given that the jsonlog format has the timestamp as the first field in the log message [1]. So the 1000 byte check would always fail on truncate, regardless of the recent upstream fix. I guess, we can keep it simple and avoid adding an external log rotation mechanism.

[1]: https://github.com/postgres/postgres/blob/2fd84e2226e2e1e13f5ad57bec3465488d67e944/src/backend/utils/error/jsonlog.c#L146

cli-commands/pg/post/add-log-destination.rb
```diff
@@ -0,0 +1,32 @@
...
+  desc "Add a PostgreSQL log destination"
+
+  banner "ubi pg (location/pg-name | pg-id) add-log-destination name otlp url [key=value [...]]\n" \
+         "       ubi pg (location/pg-name | pg-id) add-log-destination name syslog tcp://host:port [sd-id/key=value [...]]"
+
+  args(3..)
+
+  run do |args, _, cmd|
+    name, type, url, *extra_args = args
+    unless %w[otlp syslog].include?(type)
```
jeremyevans:
```suggestion
    unless %w[otlp syslog].freeze.include?(type)
```

cli-commands/pg/post/add-log-destination.rb
```diff
@@ -0,0 +1,32 @@
...
+      structured_data = structured_data_args_to_hash(extra_args, cmd)
+      structured_data ? {"structured_data" => structured_data} : nil
+    end
+    data = sdk_object.add_log_destination(name:, type:, url:, options:)
+    body = []
+    body << "Log destination added to PostgreSQL database.\n"
+    body << "Current log destinations:\n"
+    data[:log_destinations].each_with_index do |ld, i|
+      body << "  " << (i + 1).to_s << ": " << ld[:id] << "  " << ld[:name] << "  " << ld[:type] << "  " << ld[:url] << "\n"
+    end
```
jeremyevans: This doesn't tell you which log destination was added. Considering this information is available via `show`, I would just return return a message with the log destination id.

lib/ubi_cli.rb
```diff
@@ -277,6 +277,19 @@ def config_entries_to_hash(args, cmd)
...
   end
 
+  def structured_data_args_to_hash(args, cmd)
+    result = {}
+    args.each do |arg|
+      unless arg.include?("/")
+        raise Rodish::CommandFailure.new("invalid structured_data argument, expected sd-id/key=value format: #{arg.inspect}", cmd)
+      end
+      sd_id, rest = arg.split("/", 2)
+      key, value = rest.split("=", 2)
```
jeremyevans, line 290: Didn't there used to be a check here that `=` was present? Should we allow `value` to be `nil` here? If not, we should check for it.

lib/ubi_cli.rb
```diff
@@ -277,6 +277,19 @@ def config_entries_to_hash(args, cmd)
     end
   end
 
+  def structured_data_args_to_hash(args, cmd)
+    result = {}
+    args.each do |arg|
+      unless arg.include?("/")
+        raise Rodish::CommandFailure.new("invalid structured_data argument, expected sd-id/key=value format: #{arg.inspect}", cmd)
+      end
+      sd_id, rest = arg.split("/", 2)
```
jeremyevans: Better to scan the string once instead of twice:

```suggestion
      sd_id, rest = arg.split("/", 2)
      unless rest
        raise Rodish::CommandFailure.new("invalid structured_data argument, expected sd-id/key=value format: #{arg.inspect}", cmd)
      end
```

sdk/ruby/lib/ubicloud/model/postgres.rb
```diff
@@ -71,6 +71,33 @@ def add_metric_destination(username:, password:, url:)
...
     end
 
+    # Add a log destination for this database.
+    # type must be "otlp" or "syslog".
+    # For otlp: pass an https:// url and optionally options: {"headers" => {...}} for
+    # providers that require header-based authentication (e.g. Grafana, New Relic, Datadog).
+    # For syslog: pass a tcp://host:port url and optionally options: {"structured_data" => {...}} for
+    # providers that require SD-based authentication (e.g. Mezmo, Papertrail).
+    # Returns the serialized postgres resource.
+    def add_log_destination(name:, type:, url:, options: nil)
```
jeremyevans: I recommend making this a private method, and adding `add_otlp_log_destination(name:, url:, headers: {})` and `add_syslog_log_destination(name:, host:, port: 6514, structured_data: {})` that call it.

cli-commands/pg/post/add-log-destination.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+UbiCli.on("pg").run_on("add-log-destination") do
+  desc "Add a PostgreSQL log destination"
+
+  banner "ubi pg (location/pg-name | pg-id) add-log-destination name otlp url [key=value [...]]\n" \
```
jeremyevans: I think it would be more user friendly to have separate `add-otlp-log-destination` and `add-syslog-log-destination` commands. Two separate usage lines makes it pretty obvious that trying to cram both into one command is a bad idea.

migrate/20260414_add_postgres_log_destination.rb
```diff
@@ -0,0 +1,16 @@
...
+  change do
+    create_table(:postgres_log_destination) do
+      # UBID.to_base32_n("1d") => 45
+      column :id, :uuid, primary_key: true, default: Sequel.function(:gen_random_ubid_uuid, 45)
+      foreign_key :postgres_resource_id, :postgres_resource, type: :uuid, null: false
+      column :name, :text, null: false
+      column :type, :text, null: false
+      column :url, :text, null: false
+      column :options, :text
+      constraint(:valid_type, type: %w[otlp syslog])
```
byucesoy, line 13: Ah, so we are supporting syslog as well then?

rhizome/postgres/bin/configure-logs
```diff
@@ -0,0 +1,51 @@
...
+
+# Grant traverse on /dat and each intermediate directory down to the log dir
+r "setfacl", "-m", "u:#{otelcol_user}:x", "/dat"
+r "setfacl", "-m", "u:#{otelcol_user}:x", "/dat/#{version}"
+r "setfacl", "-m", "u:#{otelcol_user}:x", "/dat/#{version}/data"
+
+# Grant read on existing log files, read+traverse on the log dir itself,
+# and a default ACL so new files PostgreSQL creates are automatically readable
+r "setfacl", "-Rm", "u:#{otelcol_user}:r", log_dir
+r "setfacl", "-m", "u:#{otelcol_user}:rx", log_dir
```
byucesoy: Would this cover all the 5 lines above and nothing else or am I missing an edge case?
```
r "setfacl", "-Rm", "u:#{otelcol_user}:rx", log_dir
```

rhizome/postgres/bin/configure-logs
```diff
@@ -0,0 +1,51 @@
...
+r "setfacl", "-d", "-m", "u:#{otelcol_user}:r", log_dir
+
+safe_write_to_file("/etc/otelcol-contrib/config.yaml", log_config.to_config)
```
byucesoy, line 43: We generate the log_config above, but then add bunch of other code between its creation and usage. Let's move this line right after `OtelLogConfig.new` or move `OtelLogConfig.new` to here, so that assignment and usage of the variable would be close to each other.

model/postgres/postgres_server.rb
```diff
@@ -65,6 +67,7 @@ def configure_hash
       "log_truncate_on_rotation" => "true",
       "log_rotation_age" => "1440",
       "logging_collector" => "on",
+      "log_destination" => "'stderr,jsonlog'",
```
byucesoy: This doubles the work and log storage on the disk. I guess the motivation was this:

```
The postgres logs are expected to be in json format, to avoid parsing any custom log format that might be set by the customer.
```

I wonder if we can just simply put log format to the list of settings we don't allow changing. We might have few customers using it though...

rhizome/postgres/lib/otel_log_config.rb
```diff
@@ -0,0 +1,272 @@
+# frozen_string_literal: true
```
byucesoy, line 1: I don't know otel enough to review this file 🙂.

model/postgres/postgres_resource.rb
```diff
@@ -233,6 +233,18 @@ def ongoing_failover?
     servers.any? { it.taking_over? }
   end
 
+  def incr_configure
```
byucesoy: These have risk of collision with actual `incr_` methods. How about we use server_incr method you defined instead of these?

views/postgres/logs.erb
```diff
@@ -0,0 +1,153 @@
...
+      </div>
+    </div>
+    <div class="overflow-hidden rounded-lg shadow ring-1 ring-black ring-opacity-5 bg-white divide-y divide-gray-200">
+      <div class="px-4 py-5 sm:p-6">
+        <% form(action: "#{path(@pg)}/log-destination", method: :post) do %>
+          <input type="hidden" name="type" id="log-destination-type" value="<%= form_type %>">
+          <div class="space-y-4">
+            <div class="grid grid-cols-12 gap-6">
+              <div class="col-span-12 sm:col-span-4">
+                <%== part("components/form/text", name: "name", label: "Name", attributes: { required: true }) %>
```
byucesoy: Let's have providers first, then name.

migrate/20260414_add_postgres_log_destination.rb
shikharbhardwaj: Yes, mainly because there are some providers which are syslog only (for eg. Mezmo has 2 products, and the older "Log analysis" one only supports syslog), so keeping both looked most complete.

model/postgres/postgres_server.rb
shikharbhardwaj: At the moment, there's some divergence in log formatting in general, for eg: AWS uses a different log_line_prefix than metal. We can use the same log_line_prefix for all instances and disallow customers from overriding, as I checked today and only one customer sets this, only to add the client IP/hostname.

With that, keeping the stderr as the only source also has a couple disadvantages:
1. Logs for certain query errors have DETAIL/HINT/STATEMENT fragmentation which get printed as separate stderr log lines and lose correlation.
2. A few fields are absent in the default prefix, like session_id, session_start, txid, vxid, backend_type, remote_host, remote_port, query_id, state_code.

We will need to:
1. Make the aws log line prefix the default for all instances
2. Disallow changing log_line_prefix
3. Add a regex parser to extract these fields from stderr

This still won't give us all the fields that JSON does, but will be consistent, and won't be that brittle unless PG changes the log format definitions. wdyt?

On the flipside: do we need the stderr logs if we have log aggregation?

rhizome/postgres/bin/configure-logs
shikharbhardwaj: I simplified these

model/postgres/postgres_server.rb
shikharbhardwaj: Added the change to make the AWS prefix default and only generate one set of log files

byucesoy (approved):

rhizome/postgres/lib/otel_log_config.rb
```diff
@@ -0,0 +1,275 @@
...
+class OtelLogConfig
+  def initialize(instance:, server_role:, log_dir:, resource_name:, log_destinations:)
+    @instance = instance
+    @server_role = server_role
+    @log_dir = log_dir
+    @resource_name = resource_name
+    @log_destinations = log_destinations
+  end
+
+  def to_config
```
iamKunalGupta, line 43: do you want to add some CI check for otel config validation similar to upstream?

shikharbhardwaj: I tried this, but the CI fails to run locally on macos with the following

```
............................................................................................................................................................................................................................................................................................................................................Error: failed to build pipelines: failed to create "journald" receiver for data type "logs": journald is only supported on linux
telemetry type is not supported
FError: failed to build pipelines: failed to create "journald" receiver for data type "logs": journald is only supported on linux
telemetry type is not supported
FError: failed to build pipelines: failed to create "journald" receiver for data type "logs": journald is only supported on linux
telemetry type is not supported
FError: failed to build pipelines: failed to create "journald" receiver for data type "logs": journald is only supported on linux
telemetry type is not supported
F....................................................
```

I think this will fit better into an E2E test because some of it depends on the data plane itself (the otelcol-contrib version in the image).

PR #5176, opened 2026-04-07
Expose k8s upgrade in UI/CLI

mohi-kalantari:
This PR introduces end-to-end support for Kubernetes cluster upgrades, exposing this capability across our API, CLI, and Web UI. It also refactors how node DNS resolution is handled during tests and provisioning by offloading it to the host dnsmasq, and includes a minor cleanup for CSI migrations.

1. Kubernetes Upgrades (API & Core Logic)
Added a new upgrade POST endpoint to KubernetesCluster to trigger cluster upgrades.
Implemented upgrade semaphore increments for both the Kubernetes cluster and its associated nodepools.
Updated nodepool upgrade logic to ensure worker nodes wait for the control plane to finish its own upgrade operation first.
Added a simple hash-based lookup to determine the next available cluster upgrade version.

2. Kubernetes Upgrades (Web UI & CLI)
Added a new CLI command to allow customers to easily upgrade their Kubernetes clusters from their terminal.
Added an Upgrade section to the Kubernetes settings tab in the Web UI, which dynamically manages 4 distinct states:
Up-to-date
Not ready for upgrade (prevents race conditions if a cluster is bootstrapping or another operation like node resize is actively running)
Upgradable (shows active button)
Upgrading (shows disabled button)
Introduced a new upgrading display state indicating an ongoing upgrade operation on either the cluster or its nodepool.

3. Testing & Infrastructure Improvements
Added comprehensive E2E test suite coverage for Kubernetes upgrades.
Refactored node DNS resolution to leverage the host's dnsmasq. This eliminates the race-prone manual /etc/hosts injection on individual K8s VMs. It delegates DNS resolution to the host infrastructure by injecting the API server host record straight into dnsmasq.conf on the hypervisor, drastically improving stability during cluster bootstrapping and upgrades.
Optimized standard K8s CSI migrations to run a single time.

### DIFF
--- a/cli-commands/kc/post/upgrade.rb
+++ b/cli-commands/kc/post/upgrade.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+UbiCli.on("kc").run_on("upgrade") do
+  desc "Upgrade a Kubernetes cluster to the next supported version"
+
+  banner "ubi kc (location/kc-name | kc-id) upgrade"
+
+  run do |opts, cmd|
+    data = sdk_object.upgrade
+    response("Scheduled version upgrade of Kubneretes cluster with id #{data.id} to version #{data.version}.")
+  end
+end
--- a/lib/option.rb
+++ b/lib/option.rb
@@ -27,6 +27,13 @@ module Option
     MACHINE_IMAGE_SEARCH_LOCATIONS.fetch(location_id, [location_id])
   end
 
+  def self.kubernetes_upgrade_candidate(version)
+    {
+      "v1.33" => "v1.34",
+      "v1.34" => "v1.35",
+    }.freeze[version]
+  end
+
   def self.families
     Option::VmFamilies.select { it.visible }
   end
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -119,6 +119,10 @@ class KubernetesCluster < Sequel::Model
     kubeadm_recorded_version&.[](/^v\d+\.\d+/)
   end
 
+  def available_upgrade_version
+    Option.kubernetes_upgrade_candidate(version)
+  end
+
   def cluster_health_report
     return unless connectivity_check_target
 
--- a/model/kubernetes/kubernetes_cluster.rb
+++ b/model/kubernetes/kubernetes_cluster.rb
@@ -32,6 +32,7 @@ class KubernetesCluster < Sequel::Model
   def display_state
     label = strand.label
     return "deleting" if destroying_set? || destroy_set?
+    return "upgrading" if upgrading?
     return "running" if label == "wait"
 
     "creating"
@@ -194,6 +195,20 @@ class KubernetesCluster < Sequel::Model
 
   def worker_functional_nodes
     nodepools.flat_map(&:functional_nodes)
+  end
+
+  def ready_for_upgrade?
+    !upgrading? && available_upgrade_version && strand.label == "wait" \
+    && !nodepools.empty? && nodepools.first.strand.label == "wait"
+  end
+
+  def upgrading?
+    upgrade_labels = %w[upgrade wait_upgrade].freeze
+    return true if upgrade_labels.include?(strand.label) || upgrade_set?
+
+    return false if nodepools.empty?
+    nodepool = nodepools.first
+    upgrade_labels.include?(nodepool.strand.label) || nodepool.upgrade_set?
   end
 end
 
--- a/prog/kubernetes/kubernetes_nodepool_nexus.rb
+++ b/prog/kubernetes/kubernetes_nodepool_nexus.rb
@@ -3,6 +3,10 @@
 class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
   subject_is :kubernetes_nodepool
 
+  def cluster
+    @cluster ||= kubernetes_nodepool.cluster
+  end
+
   def self.assemble(name:, node_count:, kubernetes_cluster_id:, target_node_size: "standard-2", target_node_storage_size_gib: nil)
     DB.transaction do
       unless KubernetesCluster[kubernetes_cluster_id]
@@ -61,12 +65,22 @@ class Prog::Kubernetes::KubernetesNodepoolNexus < Prog::Base
   label def upgrade
     decr_upgrade
 
+    nap 10 if %w[upgrade wait_upgrade].freeze.include?(cluster.strand.label) || cluster.upgrade_set?
+
     node_to_upgrade = kubernetes_nodepool.nodes.find do |node|
       node_version = kubernetes_nodepool.cluster.client(session: node.sshable.connect).version
       node_minor_version = node_version.match(/^v\d+\.(\d+)$/)&.captures&.first&.to_i
-      cluster_minor_version = kubernetes_nodepool.cluster.version.match(/^v\d+\.(\d+)$/)&.captures&.first&.to_i
+      cluster_minor_version = kubernetes_nodepool.cluster.version.match(/^v\d+\.(\d+)$/).captures.first.to_i
 
-      next false unless node_minor_version && cluster_minor_version
+      unless node_minor_version
+        Prog::PageNexus.assemble(
+          "Invalid version format for #{node.name} of cluster #{kubernetes_nodepool.cluster.ubid}",
+          ["K8sInvalidVersion", kubernetes_nodepool.cluster.ubid, node.name],
+          [kubernetes_nodepool.cluster.ubid, node.ubid],
+          extra_data: {node_version:, cluster_version: kubernetes_nodepool.cluster.version},
+        )
+        next false
+      end
 
       node_minor_version == cluster_minor_version - 1
     end
--- a/prog/test/kubernetes.rb
+++ b/prog/test/kubernetes.rb
@@ -25,7 +25,7 @@ class Prog::Test::Kubernetes < Prog::Test::Base
       name: "kubernetes-test-standard",
       project_id: frame["kubernetes_test_project_id"],
       location_id: Location::HETZNER_FSN1_ID,
-      version: Option.kubernetes_versions.first,
+      version: Option.kubernetes_versions[1],
       cp_node_count: 1,
     ).subject
     Prog::Kubernetes::KubernetesNodepoolNexus.assemble(
@@ -141,7 +141,7 @@ STS
       kubernetes_cluster.sshable.d_run(
         unit_name,
         "bash", "-c",
-        "sudo kubectl --kubeconfig /etc/kubernetes/admin.conf exec -t ubuntu-statefulset-0 -- sh -c \"head -c 500M /dev/urandom | tee /etc/data/random-data-#{i} | sha256sum | awk '{print \\$1}'\" > /dev/shm/#{unit_name}.hash",
+        "sudo kubectl --kubeconfig /etc/kubernetes/admin.conf exec -t ubuntu-statefulset-0 -- sh -c \"head -c 300M /dev/urandom | tee /etc/data/random-data-#{i} | sha256sum | awk '{print \\$1}'\" > /dev/shm/#{unit_name}.hash",
       )
     end
     hop_wait_data_write
@@ -350,6 +350,49 @@ STS
     if pod_access_rules != frame["pod_access_rules_before_reboot"]
       update_stack({"fail_message" => "ip6 pod_access rules changed after reboot"})
     end
+    hop_test_upgrade
+  end
+
+  label def test_upgrade
+    upgrade_candidate = kubernetes_cluster.available_upgrade_version
+    unless upgrade_candidate
+      update_stack({"fail_message" => "No upgrade candidate available"})
+      hop_destroy_kubernetes
+    end
+
+    kubernetes_cluster.update(version: upgrade_candidate)
+    kubernetes_cluster.incr_upgrade
+    nodepool.incr_upgrade
+
+    Clog.emit("waiting for k8s cluster upgrade to #{upgrade_candidate}")
+    hop_wait_for_upgrade
+  end
+
+  label def wait_for_upgrade
+    unless kubernetes_cluster.display_state == "running"
+      kubernetes_cluster.all_nodes.each do |node|
+        unless node_host_entries_set?(node.name)
+          if vm_ready?(node.vm)
+            ensure_hosts_entry(node.sshable, kubernetes_cluster.api_server_lb.hostname)
+            set_node_entries_status(node.name)
+          end
+        end
+      end
+      nap 15
+    end
+
+    nodes = JSON.parse(kubernetes_cluster.client.kubectl("get nodes -o json"))["items"]
+    unless nodes.size == 3 && nodes.all? { |n| n.dig("status", "nodeInfo", "kubeletVersion").start_with?("#{kubernetes_cluster.version}.") }
+      update_stack({"fail_message" => "Not all #{nodes.size} nodes upgraded to #{kubernetes_cluster.version}:\n#{kubernetes_cluster.client.kubectl("get nodes")}"})
+      hop_destroy_kubernetes
+    end
+
+    hop_verify_data_after_upgrade
+  end
+
+  label def verify_data_after_upgrade
+    nap 5 unless pod_status == "Running"
+    verify_data_hashes("upgrade")
     hop_destroy_kubernetes
   end
 
--- a/prog/test/kubernetes.rb
+++ b/prog/test/kubernetes.rb
@@ -42,23 +42,18 @@ class Prog::Test::Kubernetes < Prog::Test::Base
   label def update_loadbalancer_hostname
     nap 5 unless kubernetes_cluster.api_server_lb
     kubernetes_cluster.api_server_lb.update(custom_hostname: "k8s-e2e-test.ubicloud.test")
-    hop_update_all_nodes_hosts_entries
+    hop_update_all_nodes_dns_records
   end
 
-  label def update_all_nodes_hosts_entries
+  label def update_all_nodes_dns_records
     expected_node_count = kubernetes_cluster.cp_node_count + nodepool.node_count
-    current_nodes = kubernetes_cluster.nodes + nodepool.nodes
-    current_node_count = current_nodes.count
 
-    current_nodes.each { |node|
-      unless node_host_entries_set?(node.name)
-        nap 5 unless vm_ready?(node.vm)
-        ensure_hosts_entry(node.sshable, kubernetes_cluster.api_server_lb.hostname)
-        set_node_entries_status(node.name)
-      end
+    kubernetes_cluster.all_nodes.each { |node|
+      nap 5 unless node.vm.vm_host
+      ensure_dns_record(node.vm, kubernetes_cluster.api_server_lb.hostname)
     }
 
-    hop_wait_for_kubernetes_bootstrap if current_node_count == expected_node_count
+    hop_wait_for_kubernetes_bootstrap if kubernetes_cluster.all_nodes.count == expected_node_count
     nap 10
   end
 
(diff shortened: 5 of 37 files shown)


mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24073845975

mohi-kalantari:
/run-e2e metal kubernetes

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24078211793

routes/project/location/kubernetes_cluster.rb
```diff
@@ -101,6 +101,26 @@ class Clover
...
       end
+
+      r.post "upgrade" do
+        authorize("KubernetesCluster:edit", kc)
+        upgrade_candidate = kc.available_upgrade_version
+        DB.transaction do
+          if upgrade_candidate
+            kc.update(version: upgrade_candidate)
+            kc.incr_upgrade
+            kc.nodepools.first.incr_upgrade
```
aamederen, line 112: What do you think about making `KubernetesClusterNexus` to initiate the upgrades on the nodepools? It's not a strong opinion, but it feels easier to follow when if cluster initiates the upgrade on the nexus, instead of endpoint initiating two upgrades.

routes/project/location/kubernetes_cluster.rb
```diff
@@ -101,6 +101,26 @@ class Clover
           r.redirect kc
         end
       end
+
+      r.post "upgrade" do
+        authorize("KubernetesCluster:edit", kc)
+        upgrade_candidate = kc.available_upgrade_version
+        DB.transaction do
```
aamederen: I think we can put the audit log earlier (before the transaction block) and exit early if there's no upgrade_candidate

spec/prog/kubernetes/kubernetes_cluster_nexus_spec.rb
```diff
@@ -471,46 +471,40 @@
     end
 
     it "selects a Node with minor version one less than the cluster's version" do
-      expect(kubernetes_cluster).to receive(:version).and_return("v1.32").twice
-      expect(client).to receive(:version).and_return("v1.32", "v1.31")
+      kubernetes_cluster.update(version: "v1.34")
```
aamederen: +1 for reducing mocks

spec/prog/kubernetes/kubernetes_nodepool_nexus_spec.rb
```diff
@@ -140,58 +142,77 @@
...
-      expect(kn.cluster).to receive(:version).and_return("v1.32").twice
-      expect(client).to receive(:version).and_return("v1.33", "v1.32")
-      expect { nx.upgrade }.to hop("wait")
+      it "does not select a node with minor version more than one less than the cluster's version" do
+        expect(kn.cluster).to receive(:version).and_return("v1.32").twice
+        expect(client).to receive(:version).and_return("v1.30", "v1.32")
+        expect { nx.upgrade }.to hop("wait")
+      end
+
+      it "skips nodes with invalid version formats" do
```
aamederen: shall we fire a page if the version is invalid or something we don't expect?

prog/test/kubernetes.rb
```diff
@@ -350,6 +350,50 @@ def self.assemble
     if pod_access_rules != frame["pod_access_rules_before_reboot"]
       update_stack({"fail_message" => "ip6 pod_access rules changed after reboot"})
     end
+    hop_delete_statefulset
+  end
+
+  label def delete_statefulset
```
aamederen: why do we delete statefulset before upgrade? Let's validate it at the end instead of deleting it.

prog/test/kubernetes.rb
```diff
@@ -15,7 +15,7 @@ def self.assemble
       stack: [{
         "kubernetes_service_project_id" => kubernetes_service_project.id,
         "kubernetes_test_project_id" => kubernetes_test_project.id,
-        "migration_number" => 0,
+        "migration_number" => 1,
```
aamederen: why?

routes/project/location/kubernetes_cluster.rb
mohi-kalantari: I feel like we would need more fine grained control on the clusters when doing operations in production. Doing an increment on the cluster to do the whole upgrade feels a lot. Specially when you want to upgrade the CP first to make sure API is stable then proceed to the workers.

routes/project/location/kubernetes_cluster.rb
mohi-kalantari: audit_log function creates new records in the database and the pattern I saw across all endpoints is the audit_log being included in the same single transaction that's why I wrote it like that. I guess @jeremyevans can help 😅

spec/prog/kubernetes/kubernetes_nodepool_nexus_spec.rb
mohi-kalantari: We don't expect them tbh. It's just a bit of defensive programming imo.

prog/test/kubernetes.rb
mohi-kalantari: Upgrading the whole 3 node cluster takes a good amount of time. Before this PR, we were doing 2 normal migrations now it's one to save some time and fit the whole thing in a 90 minute window

prog/test/kubernetes.rb
mohi-kalantari: It adds another migration which we have tried for 3 or 4 times. I feel like it doesn't add anything and just makes the upgrade a bit unpredictable (it can cause 1 or 2 migrations based on which node is selected first for decomissioning.)

the almost same exact scenario is tested that's why I left it out. wdt? I feel like adding it will be too lengthy for our e2e test time window.

mohi-kalantari:
/run-e2e metal


E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24091064921

lib/option.rb
```diff
@@ -18,6 +18,10 @@ def self.kubernetes_versions
     ["v1.34", "v1.33"].freeze
   end
 
+  def self.kubernetes_upgrade_candidate(version)
+    {"v1.33" => "v1.34"}[version]
```
jeremyevans: Avoids a hash allocation:

```suggestion
    {"v1.33" => "v1.34"}.freeze[version]
```

prog/kubernetes/kubernetes_nodepool_nexus.rb
```diff
@@ -61,6 +65,8 @@ def self.assemble(name:, node_count:, kubernetes_cluster_id:, target_node_size:
   label def upgrade
     decr_upgrade
 
+    nap 10 if %w[upgrade wait_upgrade].include?(cluster.strand.label) || cluster.upgrade_set?
```
jeremyevans:
```suggestion
    nap 10 if %w[upgrade wait_upgrade].freeze.include?(cluster.strand.label) || cluster.upgrade_set?
```

routes/project/location/kubernetes_cluster.rb
jeremyevans: `audit_log` should always be inside the transaction block making the changes. We should never log a change that gets rolled back (risk if `audit_log` before transaction), or miss logging a change (risk if `audit_log` after transaction).

That being said, the `if upgrade_candidate` should be moved outside the transaction, as we should only attempt the transaction if there is an upgrade candidate. If a request is made without an upgrade candidate, the route should return an error. The current code audit logs an upgrade even if an upgrade did not take place because there was no upgrade candidate, which makes no sense.

spec/prog/kubernetes/kubernetes_nodepool_nexus_spec.rb
jeremyevans: Silently ignoring invalid versions doesn't feel like a good defense to me.

spec/prog/kubernetes/kubernetes_nodepool_nexus_spec.rb
```diff
@@ -140,58 +142,77 @@
...
+      end
 
-    it "selects the first node that is one minor version behind" do
-      expect(kn.cluster).to receive(:version).and_return("v1.32")
-      expect(client).to receive(:version).and_return("v1.31")
-      expect(nx).to receive(:bud).with(Prog::Kubernetes::UpgradeKubernetesNode, {"nodepool_id" => kn.id, "old_node_id" => first_node.id, "subject_id" => kn.cluster.id})
-      expect { nx.upgrade }.to hop("wait_upgrade")
-    end
+      it "selects a node with minor version one less than the cluster's version" do
+        expect(kn.cluster).to receive(:version).and_return("v1.32").twice
```
jeremyevans: We should avoid this type of database mocking. Use real model objects, and update their state, don't mock methods on them.

Similar issues elsewhere in these specs.

spec/prog/kubernetes/kubernetes_nodepool_nexus_spec.rb
```diff
@@ -140,58 +142,77 @@
...
+        expect(client).to receive(:version).and_return("v1.32", "v1.31")
+        expect(nx).to receive(:bud).with(Prog::Kubernetes::UpgradeKubernetesNode, {"nodepool_id" => kn.id, "old_node_id" => second_node.id, "subject_id" => kn.cluster.id})
```
jeremyevans: Similarly, do not mock a method on the nexus. Instead, check that the strand was created after the `nx.upgrade`.

Similar issues elsewhere in these specs.

spec/routes/api/project/location/kubernetes_cluster_spec.rb
```diff
@@ -177,5 +178,30 @@
         end
       end
     end
+
+    describe "upgrade" do
+      let!(:kn) { kc.nodepools.first }
```
jeremyevans: You should in general avoid `let!`. I'm not sure this method is even called, but if you need it, use a `before` block so it's explicit that this is called before all specs.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -32,6 +32,10 @@ def validate
   def display_state
     label = strand.label
     return "deleting" if destroying_set? || destroy_set?
+    upgrade_labels = %w[upgrade wait_upgrade]
```
```suggestion
    upgrade_labels = %w[upgrade wait_upgrade].freeze
```

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -32,6 +32,10 @@ def validate
...
+    nodepool = nodepools.first
```
jeremyevans: Do we only care about the first nodepool and not other nodepools?

Note that this will likely cause N+1 issues unless the routes that call this on multiple clusters have the nodepools and semaphores for the nodepools eager loaded.

views/kubernetes-cluster/settings.erb
```diff
@@ -36,6 +36,71 @@
       </div>
     </div>
   </div>
+
+  <% upgrading = %w[upgrade wait_upgrade].include?(@kc.strand.label) \
+    || %w[upgrade wait_upgrade].include?(@kc.nodepools.first.strand.label) %>
```
jeremyevans: Might be worth adding a `KubernetesCluster#upgrading?` method with this logic.

```suggestion
  <% upgrading = %w[upgrade wait_upgrade].freeze.include?(@kc.strand.label) \
    || %w[upgrade wait_upgrade].freeze.include?(@kc.nodepools.first.strand.label) %>
```

views/kubernetes-cluster/settings.erb
```diff
@@ -36,6 +36,71 @@
...
+            </div>
+          </div>
+        <% elsif ready_for_upgrade %>
+          <% form(action: "#{path(@kc)}/upgrade", method: :post) do %>
+            <div class="gap-6 sm:flex sm:items-center sm:justify-between">
+              <div class="flex-grow">
+                <p class="text-sm text-gray-700">
+                  Your cluster is currently on version
+                  <strong><%= @kc.version %></strong>
+                  and will be upgraded to version
```
jeremyevans: `will` sounds like we are going to upgrade whether the customer likes it or not.

```suggestion
                  and can be upgraded to version
```

cli-commands/kc/post/upgrade.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+UbiCli.on("kc").run_on("upgrade") do
+  desc "Upgrade a Kubernetes cluster to the next available minor version"
```
jeremyevans: `available minor` is a bit too specific. We might eventual support an upgrade from `1.x` to `2.0`.

```suggestion
  desc "Upgrade a Kubernetes cluster to the next supported version"
```

cli-commands/kc/post/upgrade.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    fields = %w[id name version display-state].freeze.each(&:freeze)
+    keys = underscore_keys(fields)
+    body = []
+
+    each_with_dashed(keys) do |key, display_key|
+      body << display_key << ": " << data[key].to_s << "\n"
+    end
+
+    response(body)
```
jeremyevans: We should probably make this more similar to the Postgres upgrade output:

```suggestion
    response("Scheduled version upgrade of Kubneretes cluster with id #{data.id} to version #{data.version}.")
```

sdk/ruby/lib/ubicloud/model/kubernetes_cluster.rb
```diff
@@ -16,5 +16,9 @@ def kubeconfig
     def resize_nodepool(nodepool_ref, node_count)
       adapter.post(_path("/nodepool/#{nodepool_ref}/resize"), {node_count:})
     end
+
+    def upgrade
```
jeremyevans, line 21: Please add appropriate rdoc for the method, even if only 1 line. Unlike the rest of Ubicloud, the SDK is designed for usage by other Ruby programmers, and public methods should generally be documented.

prog/test/kubernetes.rb
```diff
@@ -350,6 +345,45 @@ def self.assemble
...
+    unless upgrade_candidate
+      update_stack({"fail_message" => "No upgrade candidate available"})
+      hop_destroy_kubernetes
+    end
+
+    DB.transaction do
+      kubernetes_cluster.update(version: upgrade_candidate)
+      kubernetes_cluster.incr_upgrade
+      nodepool.incr_upgrade
+    end
```
jeremyevans: Labels are run inside transactions.

```suggestion
    kubernetes_cluster.update(version: upgrade_candidate)
    kubernetes_cluster.incr_upgrade
    nodepool.incr_upgrade
```

prog/test/kubernetes.rb
```diff
@@ -350,6 +345,45 @@ def self.assemble
...
+
+    Clog.emit("waiting for k8s cluster upgrade to #{upgrade_candidate}")
+    hop_wait_for_upgrade
+  end
+
+  label def wait_for_upgrade
+    unless kubernetes_cluster.display_state == "running"
+      kubernetes_cluster.all_nodes.each do |node|
+        ensure_dns_record(node.vm, kubernetes_cluster.api_server_lb.hostname) if node.vm.vm_host
```
jeremyevans, line 369: Is it desirable to run this more than once per node? It looks like that is possible currently (even if it may not be likely).

spec/prog/test/kubernetes_spec.rb
```diff
@@ -811,11 +800,134 @@
...
+    it "hops to destroy_kubernetes if all 3 nodes are upgraded correctly" do
+      kubernetes_cluster.strand.update(label: "wait")
+      kubernetes_cluster.nodepools.each { |np| np.strand.update(label: "wait") }
+
+      nodes_json = {
+        "items" => [
+          {"status" => {"nodeInfo" => {"kubeletVersion" => "#{kubernetes_cluster.version}.0"}}},
+          {"status" => {"nodeInfo" => {"kubeletVersion" => "#{kubernetes_cluster.version}.0"}}},
+          {"status" => {"nodeInfo" => {"kubeletVersion" => "#{kubernetes_cluster.version}.0"}}},
+        ],
```
jeremyevans:
```suggestion
        "items" => [{"status" => {"nodeInfo" => {"kubeletVersion" => "#{kubernetes_cluster.version}.0"}}}] * 3,
```

spec/prog/test/kubernetes_spec.rb
```diff
@@ -857,28 +969,43 @@
...
+      address_line = "address=/#{vm_host.sshable.host}/first-api-server-ip"
+
+      expect(vm.vm_host.sshable).to receive(:_cmd).with("sudo cat #{dnsmasq_conf}").and_return("some config")
+      expected_cmd = <<~SH
+        set -ueo pipefail
+        sudo sed -i '/^address=/d' #{dnsmasq_conf}
+        echo #{address_line.shellescape} | sudo tee -a #{dnsmasq_conf} > /dev/null
+        sudo systemctl restart #{vm.inhost_name}-dnsmasq
+      SH
+      expect(vm.vm_host.sshable).to receive(:_cmd).with(expected_cmd)
```
jeremyevans: No objections to using the local variable, but it isn't necessary to do so. Up to you:

```suggestion
      expect(vm.vm_host.sshable).to receive(:_cmd).with(<<~SH)
        set -ueo pipefail
        sudo sed -i '/^address=/d' #{dnsmasq_conf}
        echo #{address_line.shellescape} | sudo tee -a #{dnsmasq_conf} > /dev/null
        sudo systemctl restart #{vm.inhost_name}-dnsmasq
      SH
```

prog/test/kubernetes.rb
mohi-kalantari: Yes. During the upgrades, the control plane node changes and as a result we need to update all the nodes /etc/hosts entries. Optimizing it further overcomplicates it imo.

model/kubernetes/kubernetes_cluster.rb
mohi-kalantari: Currently we only support a single nodepool. As a result we are ok with this for now.

mohi-kalantari:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24139798401

prog/test/kubernetes.rb
aamederen: Maybe we can shrink it so that, we are continuously testing PVC behavior during upgrade without much impact on the speed?

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -186,6 +190,11 @@ def worker_vms
   def worker_functional_nodes
     nodepools.flat_map(&:functional_nodes)
   end
+
+  def upgrading
```
aamederen: let's rename this to `upgrading?`

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -32,6 +32,10 @@ def validate
   def display_state
     label = strand.label
     return "deleting" if destroying_set? || destroy_set?
+    upgrade_labels = %w[upgrade wait_upgrade].freeze
+    nodepool = nodepools.first
+    return "upgrading" if upgrade_labels.include?(label) || upgrade_set? \
```
aamederen: It looks like we can just use the `upgrading?` function here

views/kubernetes-cluster/settings.erb
```diff
@@ -36,6 +36,69 @@
       </div>
     </div>
   </div>
+
+  <% ready_for_upgrade = !@kc.upgrading && @kc.available_upgrade_version \
```
aamederen: shall we wrap this logic into a function of KubernetesCluster model, and use it to guard against upgrades in the endpoint?

model/kubernetes/kubernetes_cluster.rb
aamederen: This may cause an exception during the cluster destroy process, as the nodepools.first may return empty for a very brief time.

model/kubernetes/kubernetes_cluster.rb
```diff
@@ -182,6 +190,11 @@ def worker_vms
   def worker_functional_nodes
     nodepools.flat_map(&:functional_nodes)
   end
+
+  def upgrading
+    %w[upgrade wait_upgrade].freeze.include?(strand.label) \
+    || %w[upgrade wait_upgrade].freeze.include?(nodepools.first.strand.label)
```
aamederen: the nodepools == [] case may apply here as well

mohi-kalantari:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/25159772899

aamederen (approved):

aamederen (approved):

mohi-kalantari:
/run-e2e metal

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/25662062807

PR #5178, opened 2026-04-07
Reorder GithubRunnerNexus methods

enescakir:
Most methods were added at different times and placed randomly. My preferred order is: assemble, before_run, labels, shorter property methods, then longer ones.

Having labels at the top without other methods in between makes it easier to understand the label transitions and the logic behind them.

This commit only reorders methods, no logic changes.

### DIFF
--- a/spec/prog/github/github_runner_nexus_spec.rb
+++ b/spec/prog/github/github_runner_nexus_spec.rb
@@ -55,240 +55,6 @@ RSpec.describe Prog::Github::GithubRunnerNexus do
     end
   end
 
-  describe ".pick_vm" do
-    it "provisions a VM if the pool is not existing" do
-      vm = nx.pick_vm
-      expect(vm.pool_id).to be_nil
-      expect(vm.sshable.unix_user).to eq("runneradmin")
-      expect(vm.unix_user).to eq("runneradmin")
-      expect(vm.family).to eq("standard")
-      expect(vm.vcpus).to eq(4)
-      expect(vm.project_id).to eq(Config.github_runner_service_project_id)
-    end
-
-    it "provisions a new vm if pool is valid but there is no vm" do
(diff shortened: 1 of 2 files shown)


spec/prog/github/github_runner_nexus_spec.rb
```diff
@@ -1141,4 +685,460 @@
...
+    end
+
+    it "returns setup info without vm host" do
+      vm.update(vm_host_id: nil)
+
+      expect(nx.setup_info[:detail]).to eq("Name: #{runner.ubid}\nLabel: ubicloud-standard-4\nVM Family: standard\nArch: x64\nImage: github-ubuntu-2204\nVM Host: \nVM Pool: \nLocation: \nDatacenter: \nProject: #{project.ubid}\nConsole URL: http://localhost:9292/project/#{project.ubid}/github")
+    end
+  end
+
+  describe ".rescue_common_github_api_errors" do
```
jeremyevans, line 705:
```suggestion
  describe "#rescue_common_github_api_errors" do
```

jeremyevans (approved):
While there isn't a logic change, this is not a pure reordering, as you also changed a character in a spec description (`#rescue_common_github_api_errors` => `.rescue_common_github_api_errors`). This appears to be a bug, as `rescue_common_github_api_errors` is an instance method, not a class method. Maybe this was done for consistency with other specs, but if you are going to change spec descriptions, you should fix the spec descriptions that use `.` for instance methods to use `#` (in a separate commit or PR).

FWIW, changing both the prog and the spec in the same commit reduces confidence in the change. In the future, when refactoring both production code and test code (and changing code order is a simple refactoring), it's best to split into two commits, one that only changes the production code, and one that only changes the test code. That allows for easy confirmation that the new production code passes the existing tests.

PR #5181, opened 2026-04-07
Validate VM IPv4 compatibility when attaching to load balancer

furkansahin:
Previously, a VM without IPv4 could be attached to a load balancer with an IPv4 or dual stack, which would silently fail to route traffic. This adds a server-side validation that rejects the attach request when the load balancer requires IPv4 but the VM does not have it enabled.

The view is also updated to filter the attachable VM dropdown so that only IPv4-capable VMs appear for IPv4 and dual-stack load balancers, preventing users from selecting an incompatible VM in the first place.

### DIFF
--- a/views/networking/load_balancer/vms.erb
+++ b/views/networking/load_balancer/vms.erb
@@ -1,6 +1,10 @@
-<% attachable_vms = dataset_authorize(@lb.private_subnet.vms_dataset.eager(:location), "Vm:view")
-  .exclude(Sequel[:vm][:id] => @lb.vms_dataset.select(Sequel[:vm][:id]))
-  .all
+<% ds = @lb.private_subnet.vms_dataset.eager(:location)
+  if @lb.stack == "ipv4" || @lb.stack == "dual"
+    ds = ds.where(Sequel[:vm][:ip4_enabled] => true)
+  end
+  attachable_vms = dataset_authorize(ds, "Vm:view")
+    .exclude(Sequel[:vm][:id] => @lb.vms_dataset.select(Sequel[:vm][:id]))
+    .all
 load_balancer_path = "#{@project.path}#{@lb.path}"
 edit_perm = has_permission?("LoadBalancer:edit", @lb) %>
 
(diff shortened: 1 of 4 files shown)


views/networking/load_balancer/vms.erb
```diff
@@ -1,6 +1,10 @@
-<% attachable_vms = dataset_authorize(@lb.private_subnet.vms_dataset.eager(:location), "Vm:view")
-  .exclude(Sequel[:vm][:id] => @lb.vms_dataset.select(Sequel[:vm][:id]))
-  .all
+<% ds = @lb.private_subnet.vms_dataset.eager(:location)
+  if @lb.stack == "ipv4" || @lb.stack == "dual"
+    ds = ds.where(Sequel[:vm][:ip4_enabled] => true)
+  end
+  attachable_vms = dataset_authorize(ds, "Vm:view")
+    .exclude(Sequel[:vm][:id] => @lb.vms_dataset.select(Sequel[:vm][:id]))
```
jeremyevans, line 6: Not an issue with this PR and doesn't have to be addressed right away, but is this condition what we want? This only excludes VMs already attached to this load balancer. Don't we want to exclude all VMs already attached to any load balancer (that's what the route checks)?

FWIW, I'm not sure why you cannot have a VM attached to multiple load balancers. If multiple load balancers are sending the same VM traffic, why would that be a problem?

jeremyevans (approved):

views/networking/load_balancer/vms.erb
furkansahin: Yes, ideally, we should filter out any VM that is attached to any load balancer. 

Attaching a VM to multiple load balancer is theoretically OK but not tested on the dataplane side. We are also providing a layer 4 load balancer which means that the lb doesn't have any idea if the packet is arriving for a specific lb name, it just has a port number to match. It can be fine but we should be careful. Since it has neither been tested nor asked by any customer, we are blocking it.

jeremyevans: I submitted a PR to implement this change: #5487

PR #5206, opened 2026-04-09
Fix PITR break during Postgres upgrades

shikharbhardwaj:
During upgrades, when a new cluster on the target version is created, it starts with the default postgres configs (even after pg_upgrade is run). Prior to this change, this introduced a window where postgres was run with `archive_mode=off`. This resulted in a WAL archival hole. When reconfigure and restart finished after the first base backup, this would cause standbys to get stuck indefinitely, waiting for missing WAL files to catch up. To fix this, the new cluster is configured to start with `archive_mode=on`, to make sure WAL is retained until the correct archival configuration is setup.

### DIFF
--- a/rhizome/postgres/lib/postgres_upgrade.rb
+++ b/rhizome/postgres/lib/postgres_upgrade.rb
@@ -19,7 +19,7 @@ class PostgresUpgrade
   end
 
   def disable_archiving(version, reload: false)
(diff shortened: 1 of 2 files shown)


shikharbhardwaj:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24212101416

byucesoy (approved):

shikharbhardwaj:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24234062242

PR #5211, opened 2026-04-10
Replicate PG E2E tests to AWS

byucesoy:
It is much faster to run the tests on AWS compared to bare metal. For quick E2E test runs, we can use AWS and for full suite, we can run the tests on both AWS and bare metal.


enescakir (approved):
The e2e test failed, btw

PR #5212, opened 2026-04-10
Clean stale rhizome spec files before install

### LINKED ISSUE #1161: install_rhizome doesn't remove specs files
When you run `install_rhizome(install_specs: false)` for a VM host, even if you `install_rhizome(install_specs: false)

monsonego:
## Summary

Remove stale `*_spec.rb` files before extracting rhizome when `install_specs` is false.

Closes #1161

## Problem

Running `install_rhizome(install_specs: true)` leaves specs on disk. A later `install_rhizome(install_specs: false)` skips shipping spec files in the tarball, but it does not remove the previously installed ones, so stale specs remain on the host.

## Changes

- delete existing `*_spec.rb` files under `common` and the target rhizome folder before extraction when `install_specs` is false
- keep the existing behavior unchanged when `install_specs` is true
- add specs covering the cleanup step and the `install_specs: true` path

## Testing

- added `Prog::InstallRhizome` specs for stale-spec cleanup behavior
- GitHub Actions is running on this PR
- local Ruby tooling is not available in this Windows environment, so I could not run `bundle exec rake` locally here

### DIFF
--- a/prog/install_rhizome.rb
+++ b/prog/install_rhizome.rb
@@ -47,6 +47,9 @@ class Prog::InstallRhizome < Prog::Base
     end
 
     payload = tar.string.freeze
+    unless frame["install_specs"]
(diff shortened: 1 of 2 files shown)


monsonego:
any feedback on this PR when you have a moment? :)

PR #5216, opened 2026-04-10
Match GitHub API errors by message before rate-limit handling

enescakir:
GitHub occasionally returns Octokit::TooManyRequests (HTTP 403) for errors that are not actually rate limits. For example, a request to generate a JIT config for a repository with self-hosted runners disabled raised:

    Octokit::TooManyRequests: POST .../generate-jitconfig: 403 -
    Rate Limit Exceeded
    Error summary: Repository level self-hosted runners are
    disabled on this repository

The previous rescue_common_github_api_errors handled TooManyRequests in a dedicated rescue clause that unconditionally called client.rate_limit and napped until the (non-existent) reset time, masking the real cause and wasting a runner slot until the deadline elapsed.

Collapse the two rescue clauses into a single rescue Octokit::Error. The message-based case/when now runs first so a TooManyRequests with a known page-worthy message pages and destroys the runner like any other Octokit::Error. Only if no page pattern matched and the message actually contains "API rate limit exceeded" do we fall into the rate-limit handling; otherwise the error is re-raised.

### DIFF
--- a/prog/github/github_runner_nexus.rb
+++ b/prog/github/github_runner_nexus.rb
@@ -173,35 +173,6 @@ class Prog::Github::GithubRunnerNexus < Prog::Base
 
   def rescue_common_github_api_errors
     yield
-  rescue Octokit::TooManyRequests
-    rate_limit = client.rate_limit
-    installation_ubid = github_runner.installation.ubid
-    Prog::PageNexus.assemble(
-      "GitHub API rate limit exceeded for installation #{installation_ubid}",
-      ["GithubRateLimitExceeded", installation_ubid],
-      installation_ubid,
-      severity: "warning",
-      extra_data: {
-        remaining: rate_limit.remaining,
-        limit: rate_limit.limit,
-        resets_at: rate_limit.resets_at,
-      },
-    )
-    remaining_seconds = rate_limit.resets_at - Time.now
-    if destroying_set?
-      register_deadline(nil, 15 * 60, allow_extension: true)
-      if remaining_seconds >= 15 * 60
-        github_runner.incr_skip_deregistration
-        nap 0
-      end
-    else
-      register_deadline("wait", 10 * 60, allow_extension: true)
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
I can understand non-rate limit errors are using `Octokit::TooManyRequests`, but unless rate limit errors are using something other than `Octokit::TooManyRequests`, this makes me a bit leery. I guessing this approach was done to avoid having to duplicate the logic in the `Octokit::Error` branch or refactor the code to avoid such duplication?

Since it seems unlikely GitHub will be using "API rate limit exceeded" in the message in other error cases, this is acceptable.

enescakir:
Yes, I didn't add it to the case/when block above because it increased the amount of duplicate code.

Error for disabled self-hosted runners

> Octokit::TooManyRequests: POST https://api.github.com/repos/REPOSITORY/actions/runners/generate-jitconfig: 403 - Rate Limit Exceeded (Octokit::TooManyRequests)
> Error summary:
> Repository level self-hosted runners are disabled on this repository // See: https://docs.github.com/rest/actions/self-hosted-runners#create-configuration-for-a-just-in-time-runner-for-a-repository
from /app/vendor/bundle/ruby/4.0.0/gems/octokit-10.0.0/lib/octokit/response/raise_error.rb:14:in 'Octokit::Response::RaiseError#on_complete'

Error for too many requests
> Octokit::TooManyRequests: GET https://api.github.com/repos/REPOSITORY/actions/runs?per_page=100&status=queued: 403 - API rate limit exceeded for installation ID. If you reach out to GitHub Support for help, please include the request ID REQUEST_ID and timestamp 2026-04-07 04:58:14 UTC. For more on scraping GitHub and how it may affect your rights, please review our Terms of Service (https://docs.github.com/en/site-policy/github-terms/github-terms-of-service) // See: https://docs.github.com/en/rest/using-the-rest-api/getting-started-with-the-rest-api#rate-limiting

PR #5217, opened 2026-04-10
Allow AWS e2e jobs to run in parallel across workflow runs

byucesoy:
Metal jobs remain serialized in a shared concurrency group since they
target a single dedicated host. AWS jobs with GitHub runner tests are
also serialized because we have only one app for AWS environment. That
said, AWS jobs without runner tests can be parallelized. This commit
introduces a new concurrency group for AWS jobs without runner tests,
allowing them to run in parallel across workflow runs.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,6 +27,7 @@ jobs:
     runs-on: ubuntu-slim
     outputs:
       providers: ${{ steps.set-providers.outputs.providers }}
+      has_github_runner: ${{ steps.set-providers.outputs.has_github_runner }}
     steps:
       - name: Set provider matrix
         id: set-providers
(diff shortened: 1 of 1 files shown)


byucesoy:
@enescakir would this cause any problems with GH AWS E2E runs? We have one app for the test environment, right? I wonder if we need to do anything for it.

enescakir (changes requested):
Unfortunately, GitHub tests can’t run concurrently. Each concurrent run requires a separate GitHub app. We have two apps, one for metal and one for AWS, for concurrency.​​​​​​​​​​​​​​​​

We can brainstorm together for new ideas.​​​​​​​​​​​​​​​​

byucesoy:
> Unfortunately, GitHub tests can’t run concurrently. Each concurrent run requires a separate GitHub app. We have two apps, one for metal and one for AWS, for concurrency.​​​​​​​​​​​​​​​​
> 
> We can brainstorm together for new ideas.​​​​​​​​​​​​​​​​

I suspected the same :slightly_frowning_face: . I modified the PR to disable extra concurrency if github runner tests are being executed. I'm open to ideas to allow extra concurrency for github runner tests as well, but I don't have much idea about how they got set up in the first place.

.github/workflows/e2e.yml
enescakir, line 84: We probably can't register multiple tunnels for the same URL. So it's better to not start cloudflared if we're running multiple instances concurrently. We can skip this step if the test cases don't include github_runner.

byucesoy: In my tests, this step just worked. Though it might fail if it actually tries to use the tunnel for real.

I added conditions to these steps to skip them when we don't include github runner.

enescakir (approved):

PR #5222, opened 2026-04-13
Show link to search archived records on admin panel 404 page

serprex:
we often hit 404 on flaky alerts, for postgres server you might want to check on health of postgres resource, this makes next step more discoverable, hadn't realized this tool was in admin panel

### DIFF
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -107,7 +107,11 @@ class CloverAdmin < Roda
     raise "admin route not handled: #{request.path}" if Config.test? && !ENV["DONT_RAISE_ADMIN_ERRORS"]
 
     @page_title = "File Not Found"
-    view(content: "")
+    if (ubid = request.path.split("/").find { UBID_REGEXP.match?(it) })
+      view(content: "<p>Try <a href=\"/archived-record-by-id?id=#{h ubid}\">searching archived records</a></p>")
+    else
+      view(content: "")
+    end
   end
 
   plugin :route_csrf do |token|
--- a/spec/routes/web/admin/admin_spec.rb
+++ b/spec/routes/web/admin/admin_spec.rb
@@ -822,6 +822,19 @@ RSpec.describe CloverAdmin do
     ENV.delete("DONT_RAISE_ADMIN_ERRORS")
   end
 
+  it "links to archived-record-by-id on 404 when path contains ubid" do
+    ENV["DONT_RAISE_ADMIN_ERRORS"] = "1"
+    vm = create_vm(name: "life-after-death")
+    visit "/model/Vm/#{vm.ubid}"
+    expect(page).to have_content("life-after-death")
+    vm.destroy
+    visit "/model/Vm/#{vm.ubid}"
+    click_link "searching archived records"
+    expect(page).to have_content("life-after-death")
+  ensure
+    ENV.delete("DONT_RAISE_ADMIN_ERRORS")
+  end
+
   it "raises errors by default in tests" do
     expect { visit "/error" }.to raise_error(RuntimeError)
   end


heavycrystal (approved):

clover_admin.rb
```diff
@@ -107,7 +107,11 @@ def assets
     raise "admin route not handled: #{request.path}" if Config.test? && !ENV["DONT_RAISE_ADMIN_ERRORS"]
 
     @page_title = "File Not Found"
-    view(content: "")
+    if (ubid = request.path.split("/").find { UBID_REGEXP.match?(it) })
```
enescakir, line 110: We can also match with `UUID_REGEXP` as well.

enescakir (approved):

clover_admin.rb
```diff
@@ -107,7 +107,11 @@ def assets
     raise "admin route not handled: #{request.path}" if Config.test? && !ENV["DONT_RAISE_ADMIN_ERRORS"]
 
     @page_title = "File Not Found"
-    view(content: "")
+    if (ubid = request.path.split("/").find { UBID_REGEXP.match?(it) || UUID_REGEXP.match?(it) })
```
jeremyevans: I don't think we have any UUIDs in paths, do we? We only support a UUID as a general search term.
```suggestion
    if (ubid = request.path.split("/").find { UBID_REGEXP.match?(it) })
```

spec/routes/web/admin/admin_spec.rb
```diff
@@ -822,6 +822,22 @@ def page_data
...
+  end
+
+  it "links to archived-record-by-id on 404 when path contains uuid" do
+    ENV["DONT_RAISE_ADMIN_ERRORS"] = "1"
+    visit "/model/Vm/00000000-0000-0000-0000-000000000000"
+    expect(page).to have_link("searching archived records", href: "/archived-record-by-id?id=00000000-0000-0000-0000-000000000000")
+  ensure
+    ENV.delete("DONT_RAISE_ADMIN_ERRORS")
+  end
+
```
jeremyevans: As far as I know, the admin application does not generate links like this, so this test should be removed. If there are cases where the admin application is using UUIDs instead of UBIDs in links, please let me know so we can fix that.
```suggestion

```

spec/routes/web/admin/admin_spec.rb
```diff
@@ -822,6 +822,22 @@ def page_data
     ENV.delete("DONT_RAISE_ADMIN_ERRORS")
   end
 
+  it "links to archived-record-by-id on 404 when path contains ubid" do
+    ENV["DONT_RAISE_ADMIN_ERRORS"] = "1"
+    visit "/model/Vm/vm000000000000000000000000"
+    expect(page).to have_link("searching archived records", href: "/archived-record-by-id?id=vm000000000000000000000000")
```
jeremyevans: Have the spec:

* create a VM
* go to the VM page
* destroy the VM
* refresh the page
* click the searching archived records link
* confirm the archived record shows up

clover_admin.rb
serprex: jeremy suggests otherwise

clover_admin.rb
enescakir: It's my bad.
https://github.com/ubicloud/ubicloud/pull/5222#discussion_r3071523013

I forgot that we convert UUID to UBID in the path anyway.

clover_admin.rb
```diff
@@ -107,7 +107,11 @@ def assets
     raise "admin route not handled: #{request.path}" if Config.test? && !ENV["DONT_RAISE_ADMIN_ERRORS"]
 
     @page_title = "File Not Found"
-    view(content: "")
+    if (ubid = request.path.split("/").find { UBID_REGEXP.match?(it) })
+      view(content: "<p>Try <a href=\"/archived-record-by-id?id=#{ubid}\">searching archived records</a></p>")
```
jeremyevans: Even though UBIDs do not contain characters that need escaping, as a ubid is not expected to contain html, I think it's best to escape anyway, so if that if this code is copied and reused for something that could need escaping, a security issue is not introduced.

```suggestion
      view(content: "<p>Try <a href=\"/archived-record-by-id?id=#{h ubid}\">searching archived records</a></p>")
```

Alternatively, you could add a template for this, and use `<%= %>`, which will escape by default.

jeremyevans (approved):

clover_admin.rb
serprex: added `h`

PR #5224, opened 2026-04-13
Fix allocator excluding hosts without available IPv4 for IPv6-only VMs

ayakutt:
## Issue
When a host has no available IPv4 addresses (e.g. all public IPs are assigned or the host only has its management IP), creating an IPv6-only VM gets stuck at "waiting for capacity" status even though the host has plenty of CPU, memory, and storage available. This makes it impossible to create IPv6-only VMs on hosts without spare IPv4 addresses.

## Root cause
The `available_ipv4` CTE in the allocator is joined with an INNER JOIN (`allocator.rb:135`), which eliminates hosts from the candidate list entirely if they have zero available IPv4 addresses. The conditional `WHERE ipv4_available` filter at line 239 correctly checks `request.ip4_enabled`, but it never gets a chance to run because the INNER JOIN already removed the host from the results.

## Fix
- Changed the INNER JOIN to a LEFT JOIN so hosts without available IPv4 addresses remain as candidates
- The existing conditional WHERE clause already handles filtering them out when IPv4 is actually requested
- Added a test that creates a host with no IPv4 addresses and verifies it's still returned as a candidate when `ip4_enabled` is false

## Testing
- [x] New test: host with no IPv4 addresses appears as candidate when `ip4_enabled=false` — passes with fix, fails without
- [x] All 107 existing allocator tests pass with the change

### DIFF
--- a/scheduling/allocator.rb
+++ b/scheduling/allocator.rb
@@ -132,7 +132,7 @@ module Scheduling::Allocator
     def self.candidate_hosts(request)
       ds = DB[:vm_host]
         .join(:storage_devices, vm_host_id: Sequel[:vm_host][:id])
-        .join(:available_ipv4, routed_to_host_id: Sequel[:vm_host][:id])
+        .left_join(:available_ipv4, routed_to_host_id: Sequel[:vm_host][:id])
         .left_join(:gpus, vm_host_id: Sequel[:vm_host][:id])
         .left_join(:gpu_partitions, vm_host_id: Sequel[:vm_host][:id])
         .left_join(:vm_provisioning, vm_host_id: Sequel[:vm_host][:id])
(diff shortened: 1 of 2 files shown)


ayakutt:

jeremyevans (approved):
This looks correct to me. I thought it might have been broken by my changes in 0569423301288c551573e47f596d8e07e41ecf26, but apparently it was broken before that.

enescakir:
@ayakutt Can you squash the commits?

scheduling/allocator.rb
bsatzger: Should we coalesce here, to make sure `ipv4_available` is either true or false?  similar to `use_gpu_partition`, i.e.:
`Sequel.function(:coalesce, :ipv4_available, false).as(:ipv4_available),`

ayakutt: Makes sense, done

bsatzger (approved):
thank you!

PR #5228, opened 2026-04-13
Improve DNS Server VM Setup

aamederen:
- Clean up stale zone files and journals before syncing: Removes old .zone files and journal data in /var/lib/knot at the start of sync_zones, ensuring a clean state when re-entering from validate.
- Suppress verbose zone-read logging in vms_in_sync?: The zone-read output can be very large; pass log: false to avoid cluttering logs since it's only used for comparison.
- Place new DNS server VM on a different host: Passes exclude_host_ids from existing DNS server VMs to assemble_with_sshable, improving availability by avoiding co-location on the same host.
- Add test coverage: Tests for exclude_host_ids propagation and assemble parameter forwarding (name, vm_size, storage_size_gib, location_id, boot_image).

### DIFF
--- a/prog/dns_zone/setup_dns_server_vm.rb
+++ b/prog/dns_zone/setup_dns_server_vm.rb
@@ -34,7 +34,7 @@ class Prog::DnsZone::SetupDnsServerVm < Prog::Base
   end
 
   def self.vms_in_sync?(vms)
-    return true if vms.nil? || vms.empty?
+    return true if vms.size <= 1
 
     outputs = vms.map do |vm|
       lines = vm.sshable.cmd("sudo -u knot knotc", stdin: "zone-read --", log: false).split("\n")
--- a/prog/dns_zone/setup_dns_server_vm.rb
+++ b/prog/dns_zone/setup_dns_server_vm.rb
@@ -44,7 +44,7 @@ class Prog::DnsZone::SetupDnsServerVm < Prog::Base
     return true if vms.nil? || vms.empty?
 
     outputs = vms.map do |vm|
-      lines = vm.sshable.cmd("sudo -u knot knotc", stdin: "zone-read --").split("\n")
+      lines = vm.sshable.cmd("sudo -u knot knotc", stdin: "zone-read --", log: false).split("\n")
 
       lines.map do |line|
         parts = line.split
--- a/prog/dns_zone/setup_dns_server_vm.rb
+++ b/prog/dns_zone/setup_dns_server_vm.rb
@@ -148,6 +148,8 @@ zone:
   label def sync_zones
     nap 5 if ds.dns_zones.any?(&:refresh_dns_servers_set?)
 
+    sshable.cmd "sudo rm -f /var/lib/knot/*.zone /var/lib/knot/journal/*"
+
     ds.dns_zones.each do |dz|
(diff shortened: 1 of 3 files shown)


prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -34,6 +34,7 @@ def self.assemble(dns_server_id, name: nil, vm_size: "standard-2", storage_size_
         ],
         boot_image:,
         enable_ip4: true,
+        exclude_host_ids: dns_server.vms.filter_map(&:vm_host_id),
```
enescakir: We can use `Config.allow_unspread_servers`

https://github.com/ubicloud/ubicloud/blob/43a2241e276ed422d19c2eb2ef5479424d37f274/prog/kubernetes/kubernetes_node_nexus.rb#L11

spec/prog/dns_zone/setup_dns_server_vm_spec.rb
```diff
@@ -53,6 +53,22 @@
...
 
+    it "excludes host ids of existing dns server vms" do
+      existing_vm = create_vm_with_sshable
+      vm_host = create_vm_host
+      existing_vm.update(vm_host_id: vm_host.id)
+      ds.add_vm(existing_vm)
+
+      expect(described_class).to receive(:vms_in_sync?).and_return(true)
+
+      expect(Prog::Vm::Nexus).to receive(:assemble_with_sshable).with(
```
enescakir: We can check the stack of the VM strand instead of mocking here.

prog/dns_zone/setup_dns_server_vm.rb
jeremyevans: I agree with Enes.

Also, we only need to exclude hosts in the same location, and we only need the host ids, we don't need to retrieve the entire object:

```suggestion
        exclude_host_ids: (dns_server.vms_dataset.where(location_id:).select_map(:vm_host_id) unless Config.allow_unspread_servers),
```

prog/dns_zone/setup_dns_server_vm.rb
```diff
@@ -44,7 +45,7 @@ def self.vms_in_sync?(vms)
     return true if vms.nil? || vms.empty?
```
jeremyevans: I don't think `vms` can be nil here, and you don't need to check if things are in sync if there is only a single VM:

```suggestion
    return true if vms.size <= 1
```

spec/prog/dns_zone/setup_dns_server_vm_spec.rb
```diff
@@ -53,6 +53,34 @@
       }.to raise_error RuntimeError, "No existing Project"
     end
 
+    it "excludes host ids of existing dns server vms" do
+      existing_vm = create_vm_with_sshable
+      vm_host = create_vm_host
+      existing_vm.update(vm_host_id: vm_host.id)
+      ds.add_vm(existing_vm)
+
+      expect(described_class).to receive(:vms_in_sync?).and_return(true)
```
jeremyevans: I think we can avoid mocking `vms_in_sync?` with one of the changes I am suggesting.

byucesoy (approved):

jeremyevans (approved):

PR #5229, opened 2026-04-13
Remove backward-compat storage volume creation from allocator

pykello:
In 1a128ce, storage volume creation was moved from the allocator to `Nexus.assemble`. A temporary compatibility path remained to handle VMs assembled but not yet allocated during deployment. This change removes that path.

Also remove specs that asserted storage volume creation, add explicit volume creation to specs requiring it, and remove a spec which became duplicate of another spec with this branch removed.

### DIFF
--- a/spec/scheduling/allocator_spec.rb
+++ b/spec/scheduling/allocator_spec.rb
@@ -46,6 +46,10 @@ RSpec.describe Al do
     )
   end
 
+  def create_storage_volumes(vm, storage_volume_params)
+    vm.create_storage_volumes(storage_volume_params.map { it.transform_keys(&:to_sym) })
+  end
+
   describe "allocation_request" do
     let(:storage_volumes) {
       [{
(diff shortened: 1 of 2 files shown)


spec/scheduling/allocator_spec.rb
```diff
@@ -1069,36 +1029,13 @@ def create_req(vm, storage_volumes, target_host_utilization: 0.55, distinct_stor
       expect(vol3.vring_workers).to eq(1)
     end
 
-    it "assigns correct params to each volume when volumes have different sizes & have been prepoulled" do
```
pykello, line 1072: removing this since it is now duplicate of the previous test.

pykello:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24360288194

jeremyevans (approved):

PR #5230, opened 2026-04-13
Allow running Prog::Test::PostgresResource for AWS in development/staging

jeremyevans:
This adds Config.e2e_postgres_test_project_id, which should be set to a valid project id in the environment if trying this. If set, instead of creating a new project, the test will create the Postgres resource in the given project, and it will not destroy the project after the test completes (it will still destroy the postgres resource and the created private subnet).

That project should have a us-west-2 location configured with an appropriate AWS location credential. It will use this location instead of creating a new location.

This may work with metal providers, but I didn't test that as it requires additional setup (minio, images, etc.).

If this approach looks good, I'll expand it to other postgres E2E tests. I also plan:

* An admin interface to monitor in-flight E2E tests
* The ability to start an E2E test via the admin site
* A prog that allows for continuously runs an E2E test (or set of E2E tests) as long as it is successful

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -220,6 +220,9 @@ module Config
   optional :e2e_aws_secret_key, string, clear: true
   optional :e2e_cache_proxy_download_url, string
 
+  # Local e2e
+  optional :local_e2e_postgres_test_project_id, string
+
   # Load Balancer
   optional :load_balancer_service_project_id, uuid
   optional :load_balancer_service_hostname, string
--- a/clover_admin.rb
+++ b/clover_admin.rb
@@ -1067,17 +1067,18 @@ class CloverAdmin < Roda
         r.post do
           provider = typecast_params.nonempty_str("provider")
           raise "invalid local E2E provider" unless LOCAL_E2E_PROVIDERS.include?(provider)
+          progs = typecast_params.array!(:nonempty_str, "progs")
 
-          if typecast_params.bool("local_e2e_loop")
-            prog = "LocalE2eLoop"
-            args = {progs: typecast_params.array!(:nonempty_str, "progs")}
+          sts = if typecast_params.bool("loop")
+            [Prog::Test::LocalE2eLoop.assemble(provider:, progs:)]
           else
-            prog = typecast_params.nonempty_str("prog")
-            raise "invalid local E2E prog" unless LOCAL_E2E_PROGS.include?(prog)
+            progs.map do |prog|
+              Prog::Test::LocalE2eLoop.check_prog(prog)
+              Prog::Test.const_get(prog).assemble(provider:)
+            end
           end
 
-          st = Prog::Test.const_get(prog).assemble(provider:, **args)
-          flash["notice"] = "Started local E2E strand: #{st.ubid}"
+          flash["notice"] = "Started local E2E strand(s): #{sts.map(&:ubid).join(" ")}"
           r.redirect
         end
       end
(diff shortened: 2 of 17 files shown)


byucesoy (approved):

config.rb
```diff
@@ -217,6 +217,7 @@ def self.unfrozen_test?
   optional :e2e_aws_access_key, string, clear: true
   optional :e2e_aws_secret_key, string, clear: true
   optional :e2e_cache_proxy_download_url, string
+  optional :e2e_postgres_test_project_id, string
```
enescakir: My only comment on this PR is that we're adding a config that starts with `e2e_`, but we don't set it for e2e because it's for running e2e locally in development. We might want to indicate somehow that it's for running e2e locally, maybe by adding the word `dev` somewhere.

enescakir (approved):

config.rb
jeremyevans: The goal is to eventually run this in staging and maybe production. How about `:local_e2e_postgres_test_project_id`?

jeremyevans:
This will conflict with #5238. I'll rebase on top of that.

enescakir (approved):

jeremyevans:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24472279560

jeremyevans:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24474665930

jeremyevans:
/run-e2e aws

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/24479088369

jeremyevans:
@byucesoy @shikharbhardwaj Flagging you both for review, as I came across what I think is a bug in the HA Postgres E2E tests. I was able to fix the bug by running the E2E test locally (supported by this PR), and introspecting state in pry while the E2E test was running.

The cause of the failure appears to be that the old primary recovers from the bad configuration and SIGKILL, so `wait_failover` times out. This last commit continually breaks the configuration and kills the old primary on every `wait_failover` iteration, which appears to prevent the issue.

I am not sure why a primary server would recover from a bad config and SIGKILL, but maybe one of you has a idea.

byucesoy:
> The cause of the failure appears to be that the old primary recovers from the bad configuration and SIGKILL, so wait_failover times out. This last commit continually breaks the configuration and kills the old primary on every wait_failover iteration, which appears to prevent the issue.
>
> I am not sure why a primary server would recover from a bad config and SIGKILL, but maybe one of you has a idea.

I also hit this issue and was debugging it. I suspect (not confirmed yet) that it recovers because it enters configure, which fixes the bad config and restart. If this is the case, depending on the time of the configure, the commit you added might not prevent the timeout. I'll check if this is the case and if so I'll think about a different approach to keep primary down for sure.

shikharbhardwaj:
We can add a bad config to the `user_config`, which will keep it in the failure loop

byucesoy (approved):
I fixed the failure on the parent branch with a modified version of what @shikharbhardwaj was suggesting. Adding broken config to `user_config` caused new primary to fail, so instead I directly appended to a file called `999-break.conf`, which is not touched by our code. You can rebase to it.

jeremyevans:
I rebased this on main. I removed the last commit, since the HaPostgresResource test had been fixed in a better way by 69710dd6793e5b9d7b31800de02e82ea26381987.

shikharbhardwaj (approved):

byucesoy (approved):

jeremyevans:
/run-e2e

E2E tests triggered: https://github.com/ubicloud/ubicloud/actions/runs/25119405317

jeremyevans:
E2E passed.

PR #5234, opened 2026-04-14
Add CreateVersionMetalFromUrl prog for archiving images from URLs

pykello:
### Persist archive size stats on archive-url and archive-storage-volume 

Ubiblk v0.4.1 adds `archive --stats <PATH>`, which writes:

```
{
  "physical_size_bytes": 41946957,
  "logical_size_bytes": 1073741824
}
```

Logical size is the input disk size. Physical size is the actual space used in the object store (typically smaller due to compression).

Pass `--stats` to `archive` and persist the results. This addresses:

1. `bin/archive-url` downloads an image, converts it to raw on the Rhizome side, and removes the raw file before returning. Clover still needs the raw image size so it can persist it to the database, so the size must be saved before cleanup.
2. Clover also needs the actual archive size in the object store. Previously this was computed by listing archive objects from Clover, which can require many paginated S3 list-objects requests for large archives (for example, around 40 pages / 40k stripes for a 40 GiB source disk). That can exceed 60 seconds and makes per-label time bounding harder. With `--stats`, the size is computed more efficiently on the Rhizome side, and Clover only polls for the saved result asynchronously.

### CreateVersionMetal: use archive --stats to calculate archive size 

Replace pgainated S3 list objects with reading the `--stats` JSON from `archive-storage-volume`, avoiding many S3 requests for large archives.

A machine image created from a 40G VM could have up to about 40k objects, which results in about 40 list requests. This could potentially take a long time. With the new approach, the archive size calculation is done asynchronously in Rhizome.

### UMI: Add CreateVersionMetalFromUrl for archiving images from URLs

New prog that creates a MachineImageVersionMetal by downloading an image from a URL and archiving it via `bin/archive-url`.

It can be created like the following:

```
st = Prog::MachineImage::CreateVersionMetalFromUrl.assemble(
       machine_image, version, url, sha256sum, store,
       set_as_latest: true)
```

### Update MIN_ARCHIVE_SUPPORT_VERSION to v0.4.1 

In a previous commit we started using the `archive --stats` arg when creating machine images which is available since ubiblk v0.4.1.

### DIFF
--- a/prog/machine_image/create_version_metal_from_url.rb
+++ b/prog/machine_image/create_version_metal_from_url.rb
@@ -1,0 +1,114 @@
+# frozen_string_literal: true
+
+require "json"
+
+class Prog::MachineImage::CreateVersionMetalFromUrl < Prog::Base
+  subject_is :machine_image_version
+
+  def self.assemble(machine_image, version, url, sha256sum, store, set_as_latest: true)
+    vbb = VhostBlockBackend
+      .where(vm_host_id: VmHost.where(location_id: machine_image.location_id).select(:id))
+      .where { version_code >= VhostBlockBackend::MIN_ARCHIVE_SUPPORT_VERSION }
+      .order { random.function }
+      .first
+
+    fail "no vm host with archive support found in location" unless vbb
+
+    DB.transaction do
+      miv = MachineImageVersion.create(
+        machine_image_id: machine_image.id,
+        version:,
+        actual_size_mib: 0,
+      )
+      archive_kek = StorageKeyEncryptionKey.create_random(auth_data: "machine_image_version_#{miv.ubid}_#{version}")
+      MachineImageVersionMetal.create_with_id(miv,
+        enabled: false,
+        archive_kek_id: archive_kek.id,
+        store_id: store.id,
+        store_prefix: "#{machine_image.project.ubid}/#{machine_image.ubid}/#{version}")
+
+      Strand.create_with_id(miv,
+        prog: "MachineImage::CreateVersionMetalFromUrl",
+        label: "archive",
+        stack: [{
+          "url" => url,
+          "sha256sum" => sha256sum,
+          "vm_host_id" => vbb.vm_host_id,
+          "vhost_block_backend_version" => vbb.version,
+          "set_as_latest" => set_as_latest,
+        }])
+    end
+  end
+
+  label def archive
+    register_deadline(nil, 3600)
+
+    unit_name = "archive_#{machine_image_version.ubid}"
+    sshable = vm_host.sshable
+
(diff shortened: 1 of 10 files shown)


rhizome/host/bin/archive-storage-volume
```diff
@@ -37,11 +37,17 @@ unless (kek = params["kek"])
   exit 1
 end
 
+unless (stats_file = ARGV.shift)
+  puts "expected stats_file as argument"
+  exit 1
+end
```
jeremyevans, line 43: This seems risky to add to an existing rhizome installation, at least if rhizome is deployed before the control plane changes. Can this be made backwards compatible temporarily, and then have the backwards compatibility removed in a later commit? Otherwise, I would expect breakage:

* Rhizome before control plane:
  * rhizome breaks because control plane doesn't provide new argument
* Control plane before rhizome:
  * Control plane sends new argument, but it is ignored by older rhizome and no stats are provided

Is backwards compatibility here unnecessary for some reason?

prog/machine_image/create_version_metal_from_url.rb
```diff
@@ -0,0 +1,114 @@
...
+require "json"
+
+class Prog::MachineImage::CreateVersionMetalFromUrl < Prog::Base
+  subject_is :machine_image_version
+
+  def self.assemble(machine_image, version, url, sha256sum, store, set_as_latest: true)
+    vbb = VhostBlockBackend
+      .where(vm_host_id: VmHost.where(location_id: machine_image.location_id).select(:id))
+      .where { version_code >= VhostBlockBackend::MIN_ARCHIVE_SUPPORT_VERSION }
+      .order(Sequel.lit("random()"))
```
jeremyevans: Personally, I prefer to only use `Sequel.lit` when it isn't possible to express what you want in Sequel's DSL. However, I'll leave this up to you:

```suggestion
      .order { random.function }
```

jeremyevans (commented):
The ultimate result looks good, but I'm a bit worried about the transition.

rhizome/host/bin/archive-storage-volume
pykello: > Is backwards compatibility here unnecessary for some reason?

This isn't used in production yet (API, CLI, UI integrations aren't merged yet), so I thought backwards compatibility is unnecessary. The only way to use it right now is manual pry shell.

I can change the transition sequencing to make it backwards compatible.

jeremyevans: No need for backwards compatibility if this isn't used anywhere yet (I just didn't realize that).

jeremyevans (approved):

PR #5238, opened 2026-04-14
Minor improvements in PG E2E tests

byucesoy:
**Extract shared base class for postgres e2e tests**
The three postgres e2e tests (standard, HA, upgrade) each duplicated
the postgres_test_project / postgres_resource / representative_server
accessors, the test_queries_sql / read_queries_sql helpers, and the
body of the finish label. Move them into a shared Prog::Test::PostgresBase
so bug fixes and new helpers land in one place.

**Poll for failover instead of sleeping a flat 3 minutes**
wait_failover previously slept 180 seconds unconditionally and then
optimistically hopped forward, masking slow failovers and adding idle
time to every HA test run. Replace the flat sleep with a poll that
advances as soon as a standby reports timeline_access=push in the wait
state, and fails the test if the failover does not complete within 10
minutes. Drop the now-unused failover_wait_started frame value.

Now that wait_failover guarantees a new primary exists before hopping
to test_postgres_after_failover, drop the dead "no new primary" branch
in that label and its corresponding spec.

### DIFF
--- a/prog/test/ha_postgres_resource.rb
+++ b/prog/test/ha_postgres_resource.rb
@@ -2,7 +2,7 @@
 
 require_relative "../../lib/util"
 
-class Prog::Test::HaPostgresResource < Prog::Test::Base
+class Prog::Test::HaPostgresResource < Prog::Test::PostgresBase
   def self.assemble(provider: "metal")
     postgres_test_project = Project.create(name: "Postgres-HA-Test-Project")
     Project[Config.postgres_service_project_id] ||
@@ -114,34 +114,10 @@ class Prog::Test::HaPostgresResource < Prog::Test::Base
   end
 
   label def finish
-    postgres_test_project.destroy
-
-    fail_test(frame["fail_message"]) if frame["fail_message"]
-
-    pop "Postgres tests are finished!"
+    finish_test("Postgres tests are finished!")
   end
 
   label def failed
     nap 15
-  end
-
-  def postgres_test_project
-    @postgres_test_project ||= Project[frame["postgres_test_project_id"]]
-  end
-
-  def postgres_resource
-    @postgres_resource ||= PostgresResource[frame["postgres_resource_id"]]
-  end
-
-  def representative_server
-    @representative_server ||= postgres_resource.representative_server
-  end
-
-  def test_queries_sql
-    File.read("./prog/test/testdata/order_analytics_queries.sql").freeze
-  end
-
-  def read_queries_sql
-    File.read("./prog/test/testdata/order_analytics_read_queries.sql").freeze
   end
 end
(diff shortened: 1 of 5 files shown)


prog/test/ha_postgres_resource.rb
```diff
@@ -2,22 +2,16 @@
 
 require_relative "../../lib/util"
 
-class Prog::Test::HaPostgresResource < Prog::Test::Base
+class Prog::Test::HaPostgresResource < Prog::Test::PostgresBase
   def self.assemble(provider: "metal")
```
jeremyevans, line 6: The `assemble` methods in the subclasses all look similar. I would have a superclass `assemble` method, and have each subclass `assemble` call `super` with the appropriate arguments.

prog/test/ha_postgres_resource.rb
```diff
@@ -66,27 +60,30 @@ def self.assemble(provider: "metal")
...
   end
 
+  FAILOVER_TIMEOUT_SECONDS = 10 * 60
+
   label def wait_failover
-    # Wait 3 minutes for the failover to finish.
-    failover_wait_started = frame["failover_wait_started"]
-    update_stack({"failover_wait_started" => true})
+    deadline = frame["failover_deadline"] || Time.now.to_i + FAILOVER_TIMEOUT_SECONDS
+    update_stack({"failover_deadline" => deadline})
```
jeremyevans: Don't need to update the stack if `frame["failover_deadline"]` is already set.

```suggestion
    unless (deadline = frame["failover_deadline"])
      deadline = Time.now.to_i + FAILOVER_TIMEOUT_SECONDS
      update_stack({"failover_deadline" => deadline})
    end
```

prog/test/ha_postgres_resource.rb
```diff
@@ -66,27 +60,30 @@ def self.assemble(provider: "metal")
...
+
+    new_primary = postgres_resource.servers.find { |s| s.ubid != frame["primary_ubid"] && s.timeline_access == "push" && s.strand.label == "wait" }
```
jeremyevans: N+1 query issue.

```suggestion
    new_primary = postgres_resource.servers(eager: :strand).find { |s| s.ubid != frame["primary_ubid"] && s.timeline_access == "push" && s.strand.label == "wait" }
```

prog/test/postgres_base.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def test_queries_sql
+    File.read("./prog/test/testdata/order_analytics_queries.sql").freeze
+  end
+
+  def read_queries_sql
+    File.read("./prog/test/testdata/order_analytics_read_queries.sql").freeze
+  end
+
+  def finish_test(success_msg)
```
jeremyevans, line 24: `success_msg` is the same in each caller, not sure it needs to be an argument.

jeremyevans (commented):
Overall, looks good to me.

prog/test/ha_postgres_resource.rb
jeremyevans: I handled this as part of #5230 (which is rebased on this).

prog/test/postgres_base.rb
jeremyevans: I also handled this as part of #5230.

furkansahin (approved):

PR #5241, opened 2026-04-14
Add non-SSH based fencing approach for AWS PG

byucesoy:
**I'm not fully decided on the proper approach to achieve this. Opening this PR as draft to share the idea.**

**Route AWS PG traffic through a secondary ENI**
We need a fence mechanism that cuts networking at the AWS fabric without
any guest cooperation (the AWS equivalent of bare-metal host_routing).
AWS forbids detaching the primary ENI from a running instance, so the
only fenceable via detaching NIC is a secondary one.

Every AWS PG VM now has two ENIs: an untracked placeholder at
primary ENI (required by AWS, no EIP, delete_on_termination) and the
tracked ENI carrying SSH, PG, replication, and the EIP.

Vm::Nexus.assemble accepts a use_secondary_nic kwarg, which
PostgresServerNexus.assemble sets to postgres_resource.location.aws?.
The flag lives in the VM strand's stack and is read by
Vm::Aws::Nexus#create_instance to emit the right network_interfaces
shape to run_instances.

Non-AWS VMs and AWS runners are unchanged. They continue to get a single
NIC.

**Fence AWS PG primary by detaching its tracked ENI**
Bare-metal PG's lockout has three mechanisms: pg_stop, hba, and
host_routing. host_routing is the strongest because it cuts networking
at the hypervisor level and works even when the guest is hung. AWS had
only the two SSH-dependent mechanisms, leaving the old primary unfenced
if a primary became unresponsive.

lockout_with_detach_nic calls DetachNetworkInterface on the tracked
ENI's attachment id. AWS allows detaching any device_index >= 1
instantly and statelessly; the guest kernel sees PCIe removal, the
interface disappears, and every socket on it (SSH, PG, replication)
dies, matching host_routing's total isolation. SSH is severed too,
which is acceptable because bare-metal host_routing has the same
property; post-mortem access is via EC2 Serial Console or through
primary ENI (manual set up for now)

The PostgresServerNexus#lockout label now buds detach_nic for AWS in
place of host_routing. The rescue in PostgresLockout#start is widened
to Aws::EC2::Errors::ServiceError so API failures map cleanly to
lockout_failed.

**Add migration prog for existing AWS PG instances**
Existing AWS PG VMs have their tracked NIC at device_index 0. AWS
forbids detaching device_index 0 from a running instance, so we cannot
rearrange the layout in place by removing the old NIC. The migration
prog instead adds a new ENI at device_index 1, moves all traffic onto
it, and leaves the old NIC attached as an untracked placeholder.

Resumable across crashes; every label stores its state in the strand
stack and is idempotent. listen_addresses stays '*', so PG's wildcard
listener socket automatically accepts connections on the new ENI's IP
the moment the kernel hotplugs it; no PG restart or reload is required
anywhere in the flow.

The drain phase polls pg_stat_activity and waits until all
backend_type = 'client backend' connections from the old private IP
reach zero, with a 48h deadline. No user session is force-terminated.
Walsenders are excluded from the drain, they won't drain on their own,
but get reset as a side effect of disassociating the old EIP, after
which standbys reconnect through the updated /etc/hosts entry. SSH host
swaps only after verifying the new path works, so a mid-migration
failure never leaves the VM unreachable.

### DIFF
--- a/prog/postgres/postgres_lockout.rb
+++ b/prog/postgres/postgres_lockout.rb
@@ -10,7 +10,7 @@ class Prog::Postgres::PostgresLockout < Prog::Base
       send("lockout_with_#{mechanism}")
       Clog.emit("Fenced unresponsive primary", {fenced_unresponsive_primary: {server_ubid: postgres_server.ubid, mechanism:}})
       pop "lockout_succeeded"
-    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError
+    rescue *Sshable::SSH_CONNECTION_ERRORS, Sshable::SshError, Aws::EC2::Errors::ServiceError
       pop "lockout_failed"
     end
   end
@@ -35,6 +35,18 @@ class Prog::Postgres::PostgresLockout < Prog::Base
     postgres_server.vm.vm_host.sshable.cmd(
       "timeout 10 sudo ip link set :interface down", interface: "vetho#{postgres_server.vm.inhost_name}",
       timeout: 15,
+    )
+  end
+
+  def lockout_with_detach_nic
+    nar = postgres_server.vm.nic.nic_aws_resource
+    client = postgres_server.vm.location.location_credential_aws.client
+    network_interface = client.describe_network_interfaces(
+      network_interface_ids: [nar.network_interface_id],
+    ).network_interfaces.first
+    client.detach_network_interface(
+      attachment_id: network_interface.attachment.attachment_id,
+      force: true,
     )
   end
 end
--- a/prog/postgres/postgres_server_aws_nic_migration.rb
+++ b/prog/postgres/postgres_server_aws_nic_migration.rb
@@ -1,0 +1,197 @@
+# frozen_string_literal: true
+
+require "aws-sdk-ec2"
+
+# :nocov:
+class Prog::Postgres::PostgresServerAwsNicMigration < Prog::Base
+  subject_is :postgres_server
+
+  MIGRATION_DEADLINE_SECONDS = 48 * 60 * 60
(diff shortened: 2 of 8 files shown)


fdr:
this is probably a draft thing, but I was thinking you'd use the primary eni as more than a placeholder: that it would do SSH, in particular. Even a "placeholder" eni is going to require egress rules (otherwise, extensions in-database can dial out and do bad stuff even when fenced) and needs an iota of linux configuration to make it low-priority (otherwise, even an unfenced server may try to put traffic into the wrong tube and not work).

fdr:
Note "no eip" still leaves in-subnet behaviors untouched, plus, dual-stack used the normal way will still have internet access, you do need a security group with outbound-drop and inbound-22 [presumed]

fdr (commented):
leaving a comment bit here you can rub off and refresh at any time

Per our brief chat,

1. two-nic is nice and robust for fencing on Amazon
2. two-nic on GCP barely does any fencing at all, but may be super useful for [Shared VPC](https://docs.cloud.google.com/vpc/docs/shared-vpc) setups...someone should really try that out, because on GCP peering is quite a bear.
3. two-nic would be a big, heavy lift on Ubicloud's data model, basically the long-avoided multi-nic project, and I don't think it's actually our only option for fencing, so don't worry about it

That is, none of the cases are particularly symmetric: (1) is a great fencing move, (2) is more for a form of delegation that doesn't even exist on AWS or Azure and (3) has other options. 

The misleading thing here is that the dual-nic-hood may be implemented similarly across GCP and AWS for good reason but for now they solve different problems.

